Issue #715, opened 2015-11-23
en-GB translation strings missing


pmackay:
Mentioned first in #713.

NickWeir63:
Do we need to ask for Aus help on this or will they pick up on the bug label?

pmackay:
@NickWeir63 have created a PR - if Aus can merge to master I can push to staging.

oeoeaio:
Pinging @mkllnk

mkllnk:
@pmackay I'm working on automated commits of the Transifex translations to a separate branch in our repository. But I noticed that Transifex uses `en_GB` instead of `en-GB`. Do you have any concern changing the name of the locale to `en_GB`?

pmackay:
Would changing the file to `en_GB` fix the problem of the strings not being picked up in JS? Isnt Transifex just about creating the translations?

pmackay:
My last comment was before I read https://github.com/openfoodfoundation/openfoodnetwork/pull/716#issuecomment-159748883. So I dont have a problem changing the file to `en_GB` if it doesnt affect anything else. Would that mean the `application.yml` files need changing to use that as the locale?

NickWeir63:
@pmackay 
 is it OK for us to close this now?
thanks for all your work on this

PR #716, opened 2015-11-23
#715: Add missing strings for en-GB locale.

pmackay:
Fix for #715. Tests passed here: https://buildkite.com/open-food-foundation/uk-branches/builds/25.

### DIFF
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -16,9 +16,622 @@ en-GB:
         Invalid email or password.
         Were you a guest last time?  Perhaps you need to create an account or reset your password.
   home: "OFN"
-
+  title: Open Food Network
   welcome_to: 'Welcome to '
   search_by_name: Search by name...
   producers: UK Producers
   producers_join: UK producers are now welcome to join Open Food Network UK.
   charges_sales_tax: Charges sales tax?
+  print: "Print"
+
+  logo: "Logo (640x130)"
+  logo_mobile: "Mobile logo (75x26)"
+  logo_mobile_svg: "Mobile logo (SVG)"
+  home_hero: "Hero image"
+  home_show_stats: "Show statistics"
+  footer_logo: "Logo (220x76)"
+  footer_facebook_url: "Facebook URL"
+  footer_twitter_url: "Twitter URL"
+  footer_instagram_url: "Instagram URL"
+  footer_linkedin_url: "LinkedIn URL"
+  footer_googleplus_url: "Google Plus URL"
+  footer_pinterest_url: "Pinterest URL"
+  footer_email: "Email"
+  footer_links_md: "Links"
+  footer_about_url: "About URL"
+  footer_tos_url: "Terms of Service URL"
+  invoice: "Invoice"
+
+  name: Name
+  first_name: First name
+  last_name: Last name
+  email: Email
+  phone: Phone
+  next: Next
+  address: Address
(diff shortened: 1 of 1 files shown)


pmackay:
@RohanM @oeoeaio please could you check and merge if OK?

pmackay:
Also is the travis-ci stuff being maintained still?

oeoeaio:
pinging @mkllnk

mkllnk:
Cool. Thanks for the PR. But we should fix the underlying problem as well. Javascript doesn't use a fallback language file yet.

pmackay:
So would it be better to have the `en-GB` file only contain overrides? Does Rails find strings in a fallback file already?

Is there any way to know what strings are used in JS and which in Rails?

mkllnk:
> So would it be better to have the `en-GB` file only contain overrides?

I'm not sure. From a developer view, I would think so, yes. But Transifex would always display that the translation is not complete. If you are using that trigger to commit translations automatically, the incomplete translations get never committed.

Rails has several settings to judge missing translations as an error or not.

> Does Rails find strings in a fallback file already?

Yes, it does. It is possible to deactivate that behaviour. But we included only one language file in Javascript so far. It should be easy to add a default language. I planned that at some point, but I don't think that it has a separate issue. You are welcome to open one.

> Is there any way to know what strings are used in JS and which in Rails?

Not that I know of. Of course you could use grep to look for calls of the `t` alias `translate` function or in Javascript the `t` filter (angular).

pmackay:
@mkllnk could you describe a bit more how the translation strings are fed into JS? I've updated UK staging with this merge (https://buildkite.com/open-food-foundation/uk-deployment/builds/40#a4129347-766f-40e1-a0ff-ff4d7fb0548c) but https://staging.openfoodnetwork.org.uk/#/ is still showing `label_login`. Is there something missing that it isnt picking up the `en-GB` strings? 

>But we included only one language file in Javascript so far.
Could it relate to this? What does this mean?

mkllnk:
Interesting. You can follow the [instructions to change the language](https://github.com/openfoodfoundation/ofn_deployment/wiki/Setup#languages). But in your case, I thought the asset pre-compilation during deployment should pick the newer file up. The locale is included as asset `app/assets/javascripts/darkswarm/i18n.js.erb`.

pmackay:
@mkllnk those instructions worked, which is great! However is the file deletion step not in the deploy scripts that Buildkite runs? It would be much better if deployment could handle this reliably.

mkllnk:
> However is the file deletion step not in the deploy scripts that Buildkite runs?

No, it isn't. And the deletion is just a workaround. If you do that during the deploy-step, the site won't have Javascript available for the time of the asset compilation. The pre-compilation should pick-up the language change. But I don't know why it doesn't.

pmackay:
Those instructions delete the JS files and then do asset compilation - what's the difference?

mkllnk:
@pmackay You can customise your deployment script. That is `.git/hooks/post-receive` in your app folder. Just add the line for deletion if you like. We should solve the more general problem with the pre-compilation, but I don't know how long that will take.

> Those instructions delete the JS files and then do asset compilation - what's the difference?

The difference of deleting the file first? That the deleted files are in use by a production instance. We are working in the same folder unlike the Ansible deploy scripts. And the compilation can take some minutes if there is a lot to do. I'm not sure if I got your question right.

Issue #735, opened 2015-12-14
Mailto link incorrect in group show page


edem228:
You can't get the correct email when trying to send a email to group via email icon on group show page.
![email_bug](https://cloud.githubusercontent.com/assets/5853312/11780630/8d58e344-a263-11e5-9f3f-f1ab74b1e285.png)

PR #736, opened 2015-12-14
Fix mailto link on group show/index

edem228:
Fixes: #735

### DIFF
--- a/app/views/groups/_contact.html.haml
+++ b/app/views/groups/_contact.html.haml
@@ -9,7 +9,7 @@
             = @group.phone
       - if @group.email.present?
         %p
-          =link_to_service "", @group.email.reverse, mailto: true do
+          =link_to_service "", @group.email, mailto: true do
             = t :groups_contact_email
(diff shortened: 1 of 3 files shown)


oeoeaio:
Hi @edem228 

Thanks for this contribution. I realise that the email address in the link is not human readable, but that is not really what these links are designed for, they are designed to open mailto links, which seem to be working just fine.

My concern is that the changes you have made mean that our obfuscation of the group's email address will no longer work, which is an important consideration for us as it provides protection against bots trawling our sites for email addresses.

Perhaps a better solution to the problem of human readability would be to create a new AngularJS directive that displays a popover containing the (de-obfuscated) email address on mouseover?

Let me know your thoughts.

edem228:
Hello @oeoeaio 
Ah ok I understand.
I did not have that idea in mind before. I think we don't need to create a new AngularJS directive. .It's better we leave the old functional version of the source code.
So I'll close the issue

Issue #547, opened 2015-05-08
Facebook picks random (rather than relevant) snippet when shopfront or group link is posted


kirstenalarsen:
is there any way we can get it to include information that relates to the enterprise or group that it is linking to?

![image](https://cloud.githubusercontent.com/assets/2911176/7528713/f2072964-f56f-11e4-8100-3c9fabbe501f.png)

pmackay:
@kirstenalarsen any chance of using a label for the hackathon possible issues, rather than another milestone? :)

kirstenalarsen:
i’m not too fussed if you’d rather - zenhub makes life easy for me to organise them all one way or the other :)

pmackay:
How about a `hack idea` label then? Happy to tweak...

pmackay:
Also for anyone who looks at this, might be worth doing at the same time as #551 - could be related.

kirstenalarsen:
maybe check w. @elf-pavlik first as not sure if he’s done / doing some other organising?

welcome back btw - nice to see you pop up :)

elf-pavlik:
links helpful for fixing it
* http://ogp.me/
* https://developers.facebook.com/tools/debug/

pmackay:
@elf-pavlik have you been looking at any other way of organising content for the hackathon? Any thoughts on labelling to filter across the milestones?

ludivinecp:
I'm working on it !

elf-pavlik:
possibly more useful resources
* railscast https://youtu.be/Vg8-fHiLlds
* https://github.com/lassebunk/metamagic
* https://github.com/kpumuk/meta-tags#open-graph

mkllnk:
There are two different things discussed here.

#### Posting shop link on Facebook

That is a problem at the moment. The current shop link goes to the shops list and has a hash selector that is evaluated by Javascript to display the shop's modal. While that's fine for normal browsers, Facebook does something wrong and shows the modal of another shop.

#### Providing a Facebook share button on OFN

With open graph, it is possible to put something on the timeline of the user (after receiving permission to do so). In that case the app puts all relevant information for Facebook into meta tags and we have more control over the display. But to make this work, we need a single resource for a shop, not a list of shops with a modal.

Having a single resource for each shop solves the problem already. The additional share button is therefore a feature on top of solving the problem described of this issue, I think.

Does this make sense?

lin-d-hop:
Closing #gitcull

PR #745, opened 2015-12-16
Opengraph facebook

ludivinecp:
Issues #547 

We create a helper to allow older rails version to run opengraph images (http://api.rubyonrails.org/classes/ActionView/Helpers/AssetUrlHelper.html#method-i-image_url).
For groups and shops we set up specific image, title and description, for everything else we have the default one.

Possible improvements : 
- Better default image quality.
- Enterprise logo upload need to be up to 315 x 315, instead of 100 x 100.
- Default image and description could be set in configuration file


mkllnk:
Hi!

This is great. Facebook is pulling the description from the meta tags you defined. Thank you!

Just the images don't show up in our setup. We use Amazon's S3 to store the images. Instead of the helper `image_url` with e.g. `current_distributor.logo` you should use `current_distributor.logo.url`. That will do the work for you.

And what do you think about `ContentConfig.logo.url` as default image? That is the logo at the upper left which can be uploaded by an admin user in Admin / Configuration / Content.

Em-AK:
> And what do you think about ContentConfig.logo.url as default image?

Good idea. Then if the admin hasn't uploaded any logo, it could fallback to the OFN logo.

mkllnk:
As you probably noticed, we don't use default images at the moment. It would be really nice to add that as well: #748.

app/views/layouts/darkswarm.html.haml
```diff
@@ -2,7 +2,9 @@
   %head
     %meta{charset: 'utf-8'}/
     %meta{name: 'viewport', content: "width=device-width,initial-scale=1.0"}/
-
+    %meta{:content => yield(:title).empty? ? "#{t(:title)}" : yield(:title), :property => "og:title"}/
```
Em-AK: * I would find it more readable to put the `property` key first instead of the `content` one. 
* Then you should adopt this hash syntax: `key: value` instead of `:key => value`

mkllnk:
We just merged the original commit from @ludivinecp. There are some good improvements from @vgty, but I think we still need some more fixes to this.

mkllnk:
I incorporated all our feedback and made a new commit, d562f653.

Issue #280, opened 2014-11-02
An API is available for 3rd party interactions


pmackay:
Providing a full documented API is a long term goal, to support interactions from other websites and mobile apps. 
## Review

There are a number of questions I have about the current state of APIs and use of JSON:
- Is the intent to move everything to AMS instead of RABL?
- Could the use of the `representation` module that uses `.rep` files also get replaced by AMS?
- Why do read-only operations on the `/api` need a product key? Could the key only be needed for write operations?
- Would it be possible to define a full API accessible via an `/api` endpoint and use that instead of routes ending with `.json`? Seems quite mixed currently. What implications would this have?

wvengen:
wrt AMS: is the performance really that bad to to use something else than Spree here? In https://github.com/spree/spree/issues/5971#issuecomment-86541811 some people ponder whether Spree should migrate.

pmackay:
>to to use something else than Spree here?

Can you expand on what you mean by this?

wvengen:
Spree is [using RABL](https://github.com/spree/spree/blob/master/api/app/views/spree/api/v1/products/index.v1.rabl). As a (new) developer I would expect some similarity in approaches between Spree and OFN.

pmackay:
@RohanM might need to chip in. This is an old issue and I started some work on harmonising but didnt get far. Things might have changed anyway since then. I figured it was good to be consistent within the OFN codebase (which was a mix) irrespective of what Spree uses.

wvengen:
:+1: for harmonising within OFN. I might start working on API stuff, so it would be useful for me to know which approach fits best for us.

wvengen:
Related: #287

wvengen:
I discovered the many other AMS files, and can see it makes sense to use that. Especially with Rails 5 preferring AMS as well. Currently I'm experimenting with rendering RABL from within AMS, so that we can use the same JSON format for addresses and images, for example, within our own objects (like an enterprise) - without having to copy that in AMS.

lin-d-hop:
Moved to Discourse wishlist #gitcull2017
https://community.openfoodnetwork.org/t/road-map-for-ofns-api

Issue #624, opened 2015-07-15
Want the full profile to be visible/accessible from the producer listing page.


sstead:
On this page.... www.openfoodnetwork.org.au/producers ...only the 'brief profile' is displayed. There is no indication that a longer profile exists and can be read.
This also carries into the Group producer listing pages.... e.g. www.openfoodnetwork.org.au/groups/mt-alexander-local-produce-network##producers

The only place where full profiles can be seen is the map page (group map pages also show full profiles).

Are either of these suggested solutions doable?
1.  Can we have the full modal open up, rather than the dropdown style, brief profile? Just as occurs in the map view?... this is already implemented on the Group- Hubs listing page... e.g. https://www.openfoodnetwork.org.au/groups/mt-alexander-local-produce-network##hubs
2.  Can we have a link in the brief profile view which says 'view full profile'?

*\* Context- Lucy from the MALPN group raised this issue. This group's full producer profiles all have 'outlets/stockists' at the bottom... she's worried visitors to the site won't see these if they look at producer profiles from the list.

sowfrancis:
https://github.com/wvengen/openfoodnetwork/tree/feature/624-link-to-producer-profile
we have added the link to the producer profile page.
Unfortunately there is no static producer profile page.
So we can create the static show profile page or use the modal already existing.

pmackay:
Personally I think its quite important that producers have a full profile page with a static URL, because it allows them to include that URL in other media. @kirstenalarsen can you comment on if there are reasons why not to do this?

wvengen:
I would agree that having a static url is a plus.

elf-pavlik:
:+1: giving permanent URL to each producer so they can share it easily and once we use WWW as intended, they have a proper - globally unique - identifier

kirstenalarsen:
hi people, couple of options here
- make the link bring up the modal, as it does if you click on a producer link in the hub/shop listing
- create a new producer page with static link

there is quite a difference between the two, as the first one allows for browsing and finding out more without moving and reloading pages, then both the producer modals and 'fat view' (in the listing) have links to places where you can actually go and buy from that producer

We have talked about having an actual producer page and has always been something we'd like to do, but has just never made it to top priority. The tricky bit to consider is whether it is just a profile page (with potential to do some fun stuff like add more photos, stream social media etc) or whether / to what extent is integrated with the shopping e.g.
[...]

PR #747, opened 2015-12-17
OFN API refactoring (WiP)

wvengen:
Please **don't merge yet**, input asked.
Related to #280.

For consistency, it would be good to use the same JSON format in OFN API endpoints as Spree has.
To avoid breaking apps using the current API, I'd like to introduce the `v0` version (to show that it's not finalized yet). When upgraded to Spree 3, we can start using `v1` in the url (just like Spree will do).

Since Spree uses RABL to serialize things like addresses and images, it is useful to embed those from within our serializers. With our preference being AMS, this includes a way to render RABL from AMS.
- `/api/products` etc. - handled by Spree 2 (will become `/api/v1/products` in Spree 3)
- `/api/enterprises` and `/api/order_cycles` - old OFN endpoints (deprecated)
- `/api/v0/enterprises` - new OFN endpoint (more to add)

Filtering products on OFN-specific properties can be done by ransack already. Perhaps it's useful to introduce [ransack scopes](https://github.com/activerecord-hackery/ransack#authorization-whitelistingblacklisting). Where this suffices, it is more flexible than url-scopes, like `products/managed`, in the current api.

I'd like to bring this to a place where Foodsoft can access the data needed for [food cooperatives: _getting information from producers_](http://community.openfoodnetwork.org/t/connecting-food-cooperatives/444), with the approach being ready for future additions.

Would this be a good way forward?

### DIFF
--- a/app/controllers/api/v0/enterprises_controller.rb
+++ b/app/controllers/api/v0/enterprises_controller.rb
@@ -1,0 +1,22 @@
+class Api::V0::EnterprisesController < Api::V0::BaseController
+
+  def index
+    @scope = collection.ransack(params[:q]).result
+    render_collection @scope, each_serializer: Api::V0::EnterpriseSerializer, root: 'enterprises'
+  end
+
+  def show
+    @object = collection.find(params[:id])
+    render json: @object, serializer: Api::V0::EnterpriseSerializer
+  end
+
+  private
+
+  # @see Admin::EnterprisesController#collection
+  def collection
+    OpenFoodNetwork::Permissions.new(current_api_user).
+      visible_enterprises.
+      includes(:address => [:country, :state]).
+      order('is_primary_producer ASC, name')
+  end
+end
--- a/app/serializers/api/v0/enterprise_serializer.rb
+++ b/app/serializers/api/v0/enterprise_serializer.rb
@@ -1,0 +1,10 @@
+class Api::V0::EnterpriseSerializer < ActiveModel::Serializer
+  attributes :id, :url, :name, :email, :website, :category
+  attributes :description, :long_description
+
+  has_one :address
+
+  def url
+    enterprise_url(object)
+  end
+end
--- a/app/controllers/api/v0/base_controller.rb
+++ b/app/controllers/api/v0/base_controller.rb
@@ -1,0 +1,37 @@
+# We need to have Spree's API controller for all the helper methods. But this also
+# means that ActiveModel::Serializer isn't loaded. This class works around that.
+class Api::V0::BaseController < Spree::Api::BaseController
+  include ActionController::Serialization
+
+  # Needed for ActiveModel::Serializer, and to use url helpers
+  def url_options
+    {host: request.host_with_port}
+  end
+
+  private
+
+  def render(*args, **opts)
+    if obj = opts.delete(:json)
+      data = ActiveModel::Serializer.build_json(self, obj, **opts).as_json
+      data.merge!(data.delete(:meta) || {}) unless opts[:meta_key] # meta in root by default
+      super text: data.to_json, content_type: 'application/json'
+    else
+      super *args, **opts
+    end
+  end
+
+  def render_collection(scope, **opts)
+    page = (params[:page] || 1).to_i
+    per_page = (params[:per_page] || 20).to_i # @todo default from model
+    collection = scope.page(page).per(per_page)
+    render({
+        json: collection,
+        meta: {
+          count: collection.count,
+          total_count: scope.count,
+          current_page: page,
+          pages: (scope.count * 1.0 / per_page).ceil
+        }.merge(opts[:meta]||{})
+      }.merge(opts))
+  end
+end
--- a/app/serializers/api/v0/address_serializer.rb
+++ b/app/serializers/api/v0/address_serializer.rb
@@ -1,0 +1,5 @@
+class Api::V0::AddressSerializer < Api::V0::RablSerializer
+  def template
+    "spree/api/addresses/show".freeze
+  end
+end
--- a/app/serializers/api/v0/rabl_serializer.rb
+++ b/app/serializers/api/v0/rabl_serializer.rb
@@ -1,0 +1,16 @@
+# Helper class to allow accessing Spree's RABL templates from AMS
+class Api::V0::RablSerializer
+
+  # We need to find Spree API views
+  VIEW_PATH = Spree::Api::Engine.paths["app/views"].to_a
+
+  # need to define the +template+ method
+
+  def initialize(obj, opts={})
+    @obj = obj
+  end
+
+  def serializable_hash
+    Rabl.render(@obj, template, format: 'hash', view_path: VIEW_PATH)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -129,6 +129,11 @@ Openfoodnetwork::Application.routes.draw do
       get :managed, on: :collection
       get :accessible, on: :collection
     end
+
+    # api version 0 (leading up to v1 with spree 3)
+    namespace :v0 do
+      resources :enterprises
+    end
   end
 
   namespace :open_food_network do


app/serializers/api/v0/enterprise_serializer.rb
```diff
@@ -0,0 +1,10 @@
+class Api::V0::EnterpriseSerializer < ActiveModel::Serializer
+  attributes :id, :url, :name, :email, :website, :category
+  attributes :description, :long_description
+
+  has_one :address
+
+  def url
+    enterprise_url(object)
```
wvengen: This is not a sensible url at the moment. Perhaps we could introduce a helper that makes this a group-url for groups and a producer-url for producers (#624), and a shop-url for a hub.

app/controllers/api/v0/enterprises_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  private
+
+  # @see Admin::EnterprisesController#collection
+  def collection
+    OpenFoodNetwork::Permissions.new(current_api_user).
+      visible_enterprises.
+      includes(:address => [:country, :state]).
+      order('is_primary_producer ASC, name')
```
wvengen: The website allows browsing all enterprises, right? (Or perhaps only those who are _visible_). The API should show the same things (but only allow updating to those the user has access to, of course); i.e. `visible_enterprises` may be too restrictive for reading.

wvengen:
I'm pondering how to handle the products endpoint. On one hand I'd like to use the Spree default. On the other hand there are OFN-specific changes (like the supplier relation). Override the view to add attributes in the json? Re-implement the same serializer in AMS and add things? Don't care about divergence because both could be available? I could use some thoughts on this.

app/serializers/api/v0/enterprise_serializer.rb
mkllnk: Groups are not enterprises. That's a different model. But yes, we need to distinguish between producers and shops here.

mkllnk:
Spree has just one list of products. But in OFN, products belong to a producer. Each product has at least one variant. And variants can be scoped to a shop (variant overrides).

Is the question if you have one of these?
* `/api/v0/:enterprise/variants` (OFN relation)
* `/api/v0/variants/:enterprise` (Spree style scoped)

The output of these two should be the same? If enterprise is a producer, you get all variants of a producer. If enterprise is a shop, you get all variants of that shop including its overrides.

But in some views you can filter by producer, hub and order cycle. How do we put that in a URL? Maybe:

    /api/v0/:producer/variants/:shop/:order_cycle

That would be: show all variants of a producer, sold in a shop in a certain order cycle.

@oeoeaio Does this make sense?

wvengen:
* Are variants the basic objects that are presented to the user in a shop (i.e. products are of secondary importance)?
* I do not readily see the Spree-style scoped url in its [current api documentation](https://guides.spreecommerce.com/api/variants.html).
* Nested scopes are somewhat [discouraged in rails](http://guides.rubyonrails.org/routing.html#limits-to-nesting). I'd rather use ransack searching for that (you can already search OFN relations using ransack, like `q[distributors_id_eq_or_supplier_id_eq]=`). We could add aliases to important parameters like shop and order_cycle.
* Enterprises with a shop can also be a producer. I suppose `/api/v1/:enterprise/variants` would show all products, both their own as wel as from other producers.

mkllnk:
> Are variants the basic objects that are presented to the user in a shop

Variants are the things you order. But they are grouped by product.
screenshot from 2015-12-19 09 06 16

> I do not readily see the Spree-style scoped url in its current api documentation.

That was just an example, how it could work. I thought, if there is `/api/variants` in Spree, we could append the enterprise id to the URL to scope to that enterprise. Enterprises don't exist at all in Spree. That is an OFN extension.

> `/api/v1/enterprises/:enterprise/variants`

I think this is a good idea. But I was wondering if we should distinguish between shops and producers:

* `/api/v1/producers/:enterprise/variants`
* `/api/v1/shops/:enterprise/variants`

That would make it possible to view only the variants of a producer even though it's a shop as well.

elf-pavlik:
when do you need variants of a shop which don't belong to some order cycle?

wvengen:
Thanks for the screenshot, I hadn't seen that in reality yet. I guess products and variants endpoint would be useful then.
I like the distinction between producers and shops. For my understanding: a shop has no products when there is no order cycle, right? (Meaning that you can't view shop-products of an enterprise without an active order cycle.)
@elf-pavlik Not really, I suppose that variants of a shop are usually only brought up-to-date before opening an order cycle. The only reason that might be possibly useful to view products of a closed shop (including variant overrides), is that a food cooperative is setting up their own order cycle before the shop opens. Not so important, I think.

elf-pavlik:
I see the terminology *shop* & *producer* bit confusing. Since each enterprise can **act** as *supplier* and *distributor*. While people who act as *consumer* can only order from enterprises acting as *distributor*, and possibly enterprise acting as *distributor* can only order (without cycles) from enterprises acting as *supplier*. Foodsoft, at its current state, would only need to access *inventory*(?) of enterprises which act as *distributor*. Maybe worth focusing on that use case first, before addressing complexity of a shop which possibly can have multiple open order cycles with same products/variants in them.

I will also need to look up discussions about enterprises acting as *distributor*, keeping products in *stock* (~=*inventory*?) and how it compares to them acting as *suppliers* for their own *shop*. It may also make sense to think of enterprises acting as *processor* or just *aggregation hub* (to optimize long distance transport), which other peers in network may see as *suppliers*, but products they offer will often have additional provenance data. I guess discussion about that should go somewhere on http://community.openfoodnetwork.org/ to avoid going off topic in this PR

wvengen:
A food cooperative (using e.g. Foodsoft) needs to access, in order of relevance: (1) products of a producer, (2) products/variants in a shop (open order cycle), (3) products/variants that will become available in a shop (inventory, variant overrides). The last is something for later.

elf-pavlik:
@mblaney would you like to coordinate ongoing Foodsoft -> OFN integration with your existing LettuceShare -> OFN integration? Your use cases seem similar...

oeoeaio:
Love the look of ransack scopes, I had no idea that ransack included that functionality. If we could start working out patterns for the API that use ransack scopes, I can see lots of places where our internal routing could benefit from the use of ransack scopes as well.

@wvengen:
> "For my understanding: a shop has no products when there is no order cycle, right?"

Correct, for the most part. Order cycles contain all of the information about which shops can offer which products for sale, there is no direct link between a product/variant and a shop. *However* with the introduction of [Inventory Lists](http://community.openfoodnetwork.org/t/upgrade-variant-overrides-phase-2-inventory-lists/165), we will start to associate particular sets of products/variants more directly with shops.

When this functionality is introduced, it may be beneficial for the API to expose the list of variants in a shop's inventory, regardless of the state of any order cycle. This should be fairly simple, particularly if ransack scopes are employed. As I mentioned, this is a future consideration, so not something that needs to be built now, but should be kept in mind.

elf-pavlik:
Regarding structure of URLs, I would like to mention that only server needs to understand it. Client [should consider all URLs as opaque](http://www.w3.org/DesignIssues/Axioms.html#opaque) and follow it's nose using links and understanding meaning of different types of links, for example [link relations](http://www.iana.org/assignments/link-relations/link-relations.xhtml). If client needs to construct URL, API should provide a [URI template](https://tools.ietf.org/html/rfc6570) and machine readable description of all variables it uses. As an example we can look at [triple pattern fragment hypermedia control](http://www.hydra-cg.com/spec/latest/triple-pattern-fragments/#controls). We can discuss all above in more depth, I just wanted to point the difference in how clients and servers use URLs.

mblaney:
hi @elf-pavlik sure that sounds great. OFN did the integration work to allow import into lettuceshare. I'm happy to help from my end and work with anyone else who wants to do cross-platform.

elf-pavlik:
@mblaney awesome! do you have somewhere documented the precise use case which current integration between OFN and Lettuceshare enables? How do you see changes proposed here by @wvengen fitting it?

mkllnk:
@elf-pavlik Under Admin/Reports is a Lettuceshare report that provides a table with the columns needed for Lettuceshare. I don't really know how Lettuceshare works, but I guess it's just a matter of getting that table as CSV from OFN and import it into Lettuceshare. The report contains variants optionally filtered by distributor, producer and order cycle.

mblaney:
yes as @mkllnk has said lettuceshare can import the CSV file generated by OFN. There's some information at https://lettuceshare.org/docs/stock, but the format isn't well documented sorry. (It started as merely formalising the format used by an existing distributor).

@elf-pavlik, @wvengen I would be keen to see where this api discussion ends up, and can try and keep the same terminology in my work. I need to do more on the idea of variants in the new year as it has been requested by some people, though that is more from a supplier perspective. End users tend to purchase products by kg or per item through lettuceshare and the variant information is only used by the admin placing an order with a supplier.

wvengen:
Just giving a small update: somehow this issue became a bit too big for me to finish properly. I was able to connect Foodsoft ('s [sharedlists](https://github.com/foodcoop-adam/sharedlists/tree/feature/ofn-import)) and OFN with this PR in a prototype, and I learned a lot about how OFN handles things, which I hope to put to use some time later. But I need to do some homework first.

So I hope this may be useful to someone as a starting point.
And please feel free to close this PR (perhaps link it somewhere) or leave it open.

daniellemoorhead:
---Doing a PR Cleanup---

@mkllnk @elf-pavlik any thoughts on what to do with this? Worth putting it out to the broader community to see if there's someone new on the team willing to pick it up?

myriamboure:
Maybe this work will start again if we get some expected contracts that would finance that work? Else I think Kalin is working on that, I can't @ him as he doesn't seem to be in the contributors team, sending him the link for him to comment.

djodjoni:
@myriamboure  thx for introducing me to this issue. I've been looking into and experimenting a little with  spree an ofn controllers to spit out some data but didn't do any serious work on that. I will most probably dig into it more soon especially if we have a user in Belgium :)

elf-pavlik:
https://github.com/openfoodfoundation/openfoodnetwork/pull/747#issuecomment-226596811
> And please feel free to close this PR (perhaps link it somewhere) or leave it open.

I think you should close this PR and just keep it as a reference for future work.

daniellemoorhead:
@djodjoni see @elf-pavlik comment above. If you will be digging into it more soon do you want to use this work that has been done previously? If not then please close this and then use as a reference. Cheers!

djodjoni:
@daniellemoorhead I will definitely reuse this work or part of it, however I would rather create separate issues/PRs targeting more specific features as this is too broad for me. I agree with @elf-pavlik and we can close this so please go ahead as I don't have rights to do so :)

PR #750, opened 2015-12-18
Social login

ludivinecp:
issue #9
User can sign in with their facebook account if they already have an Openfood account. Otherwise they are redirect to the sign up page. The sign in is only possible if the facebook and the Openfood email are the same.
Possible improvements :
- Allow users to create an account with their facebook profile
- Create a link between Openfood account and facebook account

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,9 +4,7 @@ ruby "2.1.5"
 gem 'rails', '3.2.21'
 gem 'rails-i18n', '~> 3.0.0'
 gem 'i18n', '~> 0.6.11'
-gem "spree_social", :git => "git://github.com/spree/spree_social.git"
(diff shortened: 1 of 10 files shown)


daniellemoorhead:
---Doing a PR Cleanup---

@elf-pavlik any thoughts on what to do with this PR? Is it worth keeping around and letting someone else build on it, or close it?

myriamboure:
@ludivinecp would you like trying finishing it if you are still around? Or maybe @Em-AK would have an idea as it was done during the hackathon I think? Is there a lot to do to finish it?

lin-d-hop:
Closing as this has been sitting around for a while so is surely quite out of date. It is still referenced from the issue that is currently still open so we can find is again for future reference.
#gitcull2017

PR #751, opened 2015-12-19
Use config to get logo in mails

mkllnk:
Use ContentConfig.footer_logo to get localized logo file instead of
defaut AUS logo.


mkllnk:
It's in production and in tag v1.3.1. @marito59

PR #770, opened 2016-01-07
Fix 2 localisation issues

marito59:
Created an order_pickup_time entry in locale files.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -66,7 +66,8 @@ module Openfoodnetwork
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     config.i18n.default_locale = ENV["LOCALE"]
-
+    I18n.locale = config.i18n.locale = config.i18n.default_locale
+    
     # Setting this to true causes a performance regression in Rails 3.2.17
     # When we're on a version with the fix below, we can set it to true
     # https://github.com/svenfuchs/i18n/issues/230
--- a/app/views/spree/shared/_order_details.html.haml
+++ b/app/views/spree/shared/_order_details.html.haml
@@ -62,7 +62,7 @@
         %strong= order.shipping_method.name
       .pad
         .text-big
-          Ready for collection
+          = t :order_pickup_time
           %strong #{order.order_cycle.pickup_time_for(order.distributor)}
         %p.text-small.text-skinny.pre-line
           %em= order.shipping_method.description.andand.html_safe || ""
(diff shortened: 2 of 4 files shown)


config/application.rb
```diff
@@ -66,7 +66,8 @@ class Application < Rails::Application
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     config.i18n.default_locale = ENV["LOCALE"]
-
+    I18n.locale = config.i18n.locale = config.i18n.default_locale
```
mkllnk: Which problem did you fix with this line?

marito59: The emails sent by delayed_job were not translated. See discussion  with @RohanM here : http://community.openfoodnetwork.org/t/localisation-for-delayed-job/465

mkllnk: Thanks, interesting. Do you know if `I18n.locale` or `config.i18n.locale` must be set? I'm wondering which initialisation is missing in delayed_job so that it doesn't pick up the default locale.

marito59: What I understand is that, for reasons I don't know, delayed_job or something else is not using the variable default_locale that was defined in application.rb but I18n.locale. So it gets initialized with the same value as default_locale which itself is taken from an  export of LOCALE in .bash_rc

At least, it fixed the issue we had  in the french instance.

-----Message d'origine-----
De : "Maikel" <notifications@github.com>
Envoyé : ‎08/‎01/‎2016 08:27
À : "openfoodfoundation/openfoodnetwork" <openfoodnetwork@noreply.github.com>
Cc : "marito59" <maritore59@yahoo.fr>
Objet : Re: [openfoodnetwork] Fix 2 localisation issues (#770)

In config/application.rb:
> @@ -66,7 +66,8 @@ class Application < Rails::Application
>      # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
>      # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
>      config.i18n.default_locale = ENV["LOCALE"]
> -
> +    I18n.locale = config.i18n.locale = config.i18n.default_locale
Thanks, interesting. Do you know if I18n.locale or config.i18n.locale must be set? I'm wondering which initialisation is missing in delayed_job so that it doesn't pick up the default locale.

mkllnk:
@sstead
Okay, we should test two things on our staging server:

* The order summary displays correctly. It displays: "Ready for collection" for pickup.
* Sent emails (e.g. order confirmation) are displaying correctly (all text is there).

Issue #771, opened 2016-01-07
Minor bug with image display on Firefox


myriamboure:
On Firefox 43.0.2, here is what I see when I click on the producer link from the product  list to see who is behind the product:
![firefox-images](https://cloud.githubusercontent.com/assets/8135376/12166753/708698f6-b528-11e5-87fe-d7dc7ce059e8.png)
The products images goes in front of the pop-up producer description page.

I don't have that problem on Chrome:
![chrome-images](https://cloud.githubusercontent.com/assets/8135376/12166770/848234dc-b528-11e5-9753-eb7d8d1c7295.png)

myriamboure:
Great to see you start in the hard code Nicolas :-) It seems I can't ping you... is it because you are not in the "contributors" list on GitHub? For info @RohanM and @mkllnk, Nicolas is a new member of the French team, and as you see he has coding skills :-)

mkllnk:
Ok, I invited him to the *Community* team. That should make him a member and therefore *pingable*.

myriamboure:
Great thank you @mkllnk :-)

blancnic:
Thanks @mkllnk !

monicafrancia:
Tested in staging2 firefox browser- didn't find any problems :+1:  @mkllnk

pmackay:
@blancnic is this a fix that could be merged back and closed?

PR #776, opened 2016-01-11
#771-minor-bug-on-Firefox

blancnic:
Z-index:1 solved problem on Firefox. It works on Chrome too.


mkllnk:
Ah, thanks. That actually annoyed me a couple of times.

PR #781, opened 2016-01-19
Maintenance

askl56:


askl56:
@mkllnk can you review this? I'll fix the merge conflicts, since there are likely to be thousands, the longer it takes to be reviewed.

mkllnk:
Hi!

I really like the idea of cleaning up the code style. But we are touching a lot of files here. We have to be careful. I didn't go through all changes, but I see couple of issues:

* The retab task needs some manual work. I saw a lot of files that need tabs translated to 8 spaces, not 2.
* The tasks touch files from external sources like `app/assets/javascripts/shared/angular-local-storage.js`. We don't want to modify these.

What do you think about a pre-commit hook to check for these things?

pmackay:
@askl56 I might also question if many changes in one go is a good move. I've found these can be a bit problematic for reviewing history. 

Thoughtbot seem to agree (see [guide](https://github.com/thoughtbot/guides#guides):

>Don't rewrite existing code to follow this guide.

I'm not knocking conforming to style, also very keen on this :) I just posted http://community.openfoodnetwork.org/t/using-hound-for-code-style-checking/499/2 thinking Hound might be worth integrating to improve/maintain general style quality over time.

PR #778, opened 2016-01-15
Translation of Weight_per_kg calculator name

marito59:
Translation of the name of the calculator

cf issue  https://github.com/openfoodfoundation/openfoodnetwork/issues/774

PR #825, opened 2016-02-14
Localisation of the Weight Calculator label

marito59:
Replace PR #778

### DIFF
--- a/app/models/open_food_network/calculator/weight.rb
+++ b/app/models/open_food_network/calculator/weight.rb
@@ -4,7 +4,7 @@ module OpenFoodNetwork
     attr_accessible :preferred_per_kg
 
     def self.description
-      "Weight (per kg)"
+      I18n.t('calculator_weight_per_kg')
     end
 
     def compute(object)
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -599,3 +599,5 @@ fr:
   price_graph: "Légende détail du prix"
   included_tax: "Inclut TVA"
   remove_tax: "Retirer TVA"
+  calculator_weight_per_kg: "Poids (par kg)"
+  
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -676,3 +676,5 @@ Please follow the instructions there to make your enterprise visible on the Open
   price_graph: "Price graph"
   included_tax: "Included tax"
   remove_tax: "Remove tax"
+  calculator_weight_per_kg: "Weight (per kg)"
+  


pmackay:
@MikeiLL would merging this PR help with the point you made in https://github.com/openfoodfoundation/openfoodnetwork/pull/1018#issuecomment-221875136 about weight label?

app/models/open_food_network/calculator/weight.rb
```diff
@@ -4,7 +4,7 @@ class Calculator::Weight < Spree::Calculator
     attr_accessible :preferred_per_kg
 
     def self.description
-      "Weight (per kg)"
+      I18n.t('calculator_weight_per_kg')
```
MikeiLL: If it works it would be exactly what I had experimented with doing. I think that the measurement system could be abstracted away from the name of the variable. I used `weight_measurement_unit` in my translation file. Here in the US we don't use the metric system (never heard a reasonable explanation back in the 3rd grade).

config/locales/fr.yml
```diff
@@ -599,3 +599,5 @@ fr:
   price_graph: "Légende détail du prix"
   included_tax: "Inclut TVA"
   remove_tax: "Retirer TVA"
+  calculator_weight_per_kg: "Poids (par kg)"
```
elf-pavlik: shouldn't this go via transifex ?

mkllnk: It's fine to include it here. It just has to be translated again in Transifex.

mkllnk:
Looks good. It would be great to merge this or #1029. They are essentially the same.

elf-pavlik:
I just closed #1029 in favour of this one!

config/locales/fr.yml
MikeiLL: Am I understanding correctly that for the US instance to use Pounds instead of ounces and grams is going to require paying $850 a year (or more) to Transifex?

mkllnk: No. Transifex is free for open source projects. And you can manage the locales yourself as well. But it will be really difficult to do it manually. You need some kind of tool to find out which bits need translating.

MikeiLL:
I would still prefer if the variable name was `weight_measurement_unit`. I'm not clear on whether or not there should be a fallback in the code so that if translation string doesn't exist, code will not break. Possibly we simply want to add the translation string to all of the `localization.yml` files so that the string will exist when changes are pulled in. I can send a new PR with these changes if anyone is interested. There's also a `.coffee` file that we need to localize for the US instance: `/spec/javascripts/unit/admin/services/variant_unit_manager_spec.js.coffee`

mkllnk:
@MikeiLL `weight_measurement_unit` sounds to me like kilogram or pounds. Actually, I'm a bit confused now. Shouldn't it be "Weight (in kg)"? The variable could be `weight_and_unit`.

MikeiLL:
@mkllnk The (stubborn) US doesn't use the metric system. All of our food is measured in pounds, gallons. SO I imagine that at least as far as

`  describe "variantUnitOptions", ->
    it "returns an array of options", ->
      expect(VariantUnitManager.variantUnitOptions()).toEqual [
        ["Weight (g)", "weight_1"],
        ["Weight (kg)", "weight_1000"],
        ["Weight (T)", "weight_1000000"],
        ["Volume (mL)", "volume_0.001"],
        ["Volume (L)", "volume_1"],
        ["Volume (kL)", "volume_1000"],
        ["Items", "items"]
      ]`

There may be some new or alternate calculations required. But I am fairly sure most US citizens won't be comfortable using the metric system.

mkllnk:
Yes, I understand that about the different systems. My point was that "Weight (per kg)" sounds completely wrong. I was wondering if it should be "Weight (in kg)" or "Weight (kg)" in metric English. I have no idea of the implications of applying a non-metric system. Because there might be calculations that are not using `1000` as multiplier. Ideally, the OFN would be agnostic of that.

MikeiLL:
@mkllnk I think simply "Weight (kg)" or in our case "Weight (lb)" and the fact of the matter is, it could be translated as "wdklskowFFHIH (ñš)" if that made sense to people using a particular instance, right?

mkllnk:
Yes. That sounds good.

daniellemoorhead:
---Doing a PR Cleanup---

@MikeiLL @mkllnk where are we at with this? Tis getting quite old so I'm wondering if it's worth continuing and finishing or closing?

myriamboure:
@pacodelaluna did you work on translation on calculator labels also right? Is there any connexion to that PR? I suggest we finish the translation of the weight calculator names here, and open a new issue for non-metric weight system, right? Ping @mkllnk

mkllnk:
@daniellemoorhead I vote for continuing.

daniellemoorhead:
Closing this PR as it has been inactive for a very long time. Feel free to reopen it if it becomes a priority and work recommences. 
#githubcull2018

mkllnk:
I just looked into the code and this has actually been changed already. Done.

Issue #616, opened 2015-07-09
User balances displayed on account screen


NickWeir63:
Can we add the full transaction list to the user screen so that they can see their credit balance?

lin-d-hop:
Exploring more in http://community.openfoodnetwork.org/t/customers-being-able-to-view-their-payments-and-balance-in-account/308

lin-d-hop:
Hi @stveep,

The build on this branch is currently failing:
  1) Api::OrdersByDistributorSerializer serializes orders
     Failure/Error: expect(@serializer.to_json).to match "distributed_orders"
     NameError:
       undefined local variable or method `spree' for #<Api::OrderSerializer:0x007ff0d7611e60>
     # ./app/serializers/api/order_serializer.rb:31:in `path'
     # (eval):13:in `_fast_attributes'
     # ./spec/serializers/orders_by_distributor_serializer_spec.rb:19:in `block (2 levels) in <top (required)>'
  2) Api::OrdersByDistributorSerializer serializes the balance for each distributor
     Failure/Error: expect(@serializer.serializable_array[0].keys).to include :balance
     NameError:
       undefined local variable or method `spree' for #<Api::OrderSerializer:0x007ff0d77c38a8>
     # ./app/serializers/api/order_serializer.rb:31:in `path'
     # (eval):13:in `_fast_attributes'
     # ./spec/serializers/orders_by_distributor_serializer_spec.rb:23:in `block (2 levels) in <top (required)>'

It was passing a couple days ago... would you take a look when you can?

When it was passing I did deploy to staging and the changes looked really great. One comment on:
http://community.openfoodnetwork.org/t/customers-can-view-their-payments-and-balance/308/21

Thanks so much for this awesome work!!

stveep:
This is the method causing the problem - to serialise the path to link to the order page, using the spree url helper:

app/serializers/api/order_serializer.rb

```{ruby}
class Api::OrderSerializer < ActiveModel::Serializer
  attributes :number, :completed_at, :total, :state, :shipment_state, :payment_state, :outstanding_balance, :payments, :path

  has_many :payments, serializer: Api::PaymentSerializer
...

  def path
    spree.order_url(object.number, only_path: true)
  end
end
```

The spec passes locally for me - is there any reason the helpers wouldn't be available when Travis runs the specs? Is there a better way to get the url into the frontend? Needs to go in an ng-repeat so can't render this on the server. @oeoeaio @mkllnk ?

[...]

PR #832, opened 2016-02-21
Uk/account balances

stveep:
See: https://github.com/openfoodfoundation/openfoodnetwork/issues/616
http://community.openfoodnetwork.org/t/customers-can-view-their-payments-and-balance/308/27


app/helpers/injection_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module InjectionHelper
   def inject_enterprises
-    inject_json_ams "enterprises", Enterprise.activated.includes(:address).all, Api::EnterpriseSerializer, enterprise_injection_data
+    inject_json_ams "enterprises", Enterprise.activated.includes(address: :state).all, Api::EnterpriseSerializer, enterprise_injection_data
```
mkllnk: Interesting, does this have a positive impact on performance?

I can't see any difference in the output.

app/helpers/injection_helper.rb
```diff
@@ -51,6 +51,18 @@ def inject_enterprise_attributes
     render partial: "json/injection_ams", locals: {name: 'enterpriseAttributes', json: "#{@enterprise_attributes.to_json}"}
   end
 
+  def inject_orders_by_distributor
+    # Convert ActiveRecord::Relation to array for serialization
+    data_array = spree_current_user.orders_by_distributor.to_a
+    data_array.each do |enterprise|
```
mkllnk: I was wondering if we could do this filtering on database level in `Spree::User.orders_by_distributor`. That would require a left join, I think.

app/helpers/injection_helper.rb
stveep: I think this must have been from when I was using bullet (a gem that flags n+1 queries and unused preloading) while writing my queries and it flagged this as an n+1 where we could preload the association. Sorry it's not really relevant to this feature, should have put it on a different branch.

https://github.com/flyerhzm/bullet

Unfortunately bullet + Spree has some bugs so I stopped using it, maybe it's something to consider if we upgrade Spree though

app/helpers/injection_helper.rb
stveep: OK I'll have a go. By the way, does it make sense to you to have this query in the User model? I wasn't sure what was the best place for it.

app/helpers/injection_helper.rb
stveep: Just to add - If you use `.explain` instead of `.to_sql` in your statements above you should see a difference. Also if you use irb (i.e. `rails c`,  pry suppresses the SQL by default) you can see fewer queries:

```{ruby}
ents = Enterprise.activated.includes(:address).all
ents.map{|e| e.address.state}
# => one query per address

entstate = Enterprise.activated.includes(address: :state).all
entstate.map{|e| e.address.state}
# => no further queries
```

This enterprise injection runs on every page so I guess it's good if we can improve the performance - I didn't test this directly though, just added it for the sake of good practice.

mkllnk: Thanks. That's great to know. And a really good improvement. We have a lot of queries...

app/helpers/injection_helper.rb
mkllnk: Yes, that does make sense to me. And much better than in a serialiser. I think that serialisers shouldn't have any logic except how to *display* things for an API. The only other place could be the view. It could decide to display these payments or not. But since we have only one use case, we don't need to include data we don't need. And data selection / structuring sounds like the kind of thing a model should do.

stveep: Well I can't figure out how to do it in SQL/AR and I think it's very fiddly or not possible - I want to include show all 'completed' orders, but only show payments that were completed, not pending etc., as only completed payments count towards the order balance, which can be done by adding to the where clause. But even if there are no completed payments (no payments at all or only pending, void etc), we still need to show the order (with 'balance due' state). So we need to select orders where there are no payments that satisfy (state = complete), as well as orders and their completed payments for those that do have them. 

Hope that makes sense. My test data (a dump from aus staging) has a lot of orders with no completed payments - but maybe they wouldn't arise in real life? We could also show failed payments (or some types of failed payment) on the page, but would need to ensure that it's clear why they are not counting towards the balance.

Let me know if you have any ideas. In the meantime I moved the filtering code to the model in the latest commit.

mkllnk:
I think I got the right query to filer the payments on database level. Have a look into my last commit and verify that it's all doing what it should do. Once that's fine, I pass it to our staging server.

mkllnk:
@stveep Can you review my last commit and check that it's still doing the right thing? Your tests seem to cover most cases.

stveep:
Hi @mkllnk, thanks for looking at this. Although the query looked ok to me at first, it doesn't work properly. It returns all orders (for all users, but only from enterprises the current user has orders from), and all payments regardless of state. So seems like the where clause is not applied, but I can't figure out why. I tried rewriting the where clause directly in SQL but it gave the same result.

```{ruby}
old.first.distributed_orders.map{|o| o.state}.flatten.uniq
#=> ["complete"]
new.first.distributed_orders.map{|o| o.state}.flatten.uniq
#=> ["cart", "complete"]
old.first.distributed_orders.map{|o| o.user_id}.flatten.uniq
#=> [770]
new.first.distributed_orders.map{|o| o.user_id}.flatten.uniq
#=> [nil, 760, 758, 756, 643, 648, 649, 650, 655, 1, 168, 624, 630, 638, 642, 644, 661, 61, 664, 657, 667, 75, 660, 685, 185, 696, 806, 732, 733, 737, 744, 745, 750, 647, 634, 54, 721, 757, 755, 767, 771, 270, 781, 770, 852, 782, 776, 835, 848, 807, 802, 814, 843, 789, 791, 20, 800, 769, 777, 790, 830, 868, 850]
new.first.distributed_orders.map{|o| o.payments.map{|p| p.state}}.flatten.uniq
#=> ["checkout", "completed", "invalid"]
```

I'm going away for the weekend so won't have any more time until next week, but I've pushed some better specs (including things that shouldn't be returned) in the mean time (so this build should fail). These pass for the old query with the ruby filtering.

Also, the new query issues many more queries (one query per order, and doesn't preload the payments), whereas the old one did everything with a single query. I'm sure there should be a way to do this but it's very frustrating!

stveep:
Thought a bit more about this on discourse: http://community.openfoodnetwork.org/t/customers-can-view-their-payments-and-balance/308/35

NickWeir63:
thanks Steve - I like your suggestions.  I will ask Oliver (Stroudco manager) to comment as he is closer to the Day-2-day stuff than me

mkllnk:
@stveep I think we should go back to your database query including all the payments. Depending on your discussion within the UK team, you decide within the UI if you want to display incomplete payments or not. Would you like to work on that and then update this pull request?

stveep:
OK as discussed with @daniellemoorhead I have reverted to original query/filter and tidied up a bit - this can be merged if you're happy and I can add the ability to view incomplete payments once we have worked out the details.

mkllnk:
@stveep I just read an excellent article about different ways of pre-loading data: http://blog.arkency.com/2013/12/rails4-preloading/

That's good to have in mind for the next time we encounter such an optimisation problem.

Issue #303, opened 2014-11-24
Map can be configured to display a region appropriate to the locale


pmackay:

daniellemoorhead:
Hi @pmackay has this happened because HG Green Cheesemakers producer is displaying in southern africa, so the map has zoomed out to include this marker? I think this is because this producer doesn't have an address on their profile, so OFN puts the marker at 0,0 long/lat.

NickWeir63:
OK i put in an address for HG Green and then another producer popped up off the African coast.  So I changed that one too .....and so on for about 20+ producers (some Stroudco some DFFH) all with either a missing postcode or a missing address line 1.  However my guess is that there are a lot of them and I am flicking backwards and forwards from the map to the enterprise screens (see bleow) because I cant see which ones need editing.  Is there a quicker way to go through the whole enterprise list and do the edits?  If so please point me at it!  @daniellemoorhead please can you point me in the right direction?  thanks
![screenshot 2016-03-02 13 14 51](https://cloud.githubusercontent.com/assets/10517702/13461543/157a32c4-e07a-11e5-88db-204815c1ebc1.png)
![screenshot 2016-03-02 13 18 01](https://cloud.githubusercontent.com/assets/10517702/13461560/25c02850-e07a-11e5-9d81-544ecedfdd00.png)

stveep:
Would be better to exclude producers with no address from being plotted on the map, rather than defaulting to (0,0)

NickWeir63:
Yes please - that would be great.  Is that something that you can do @stveep ?
it seems that producers with incomplete addresses (missing postcode or missing address line 1) are also defaulting to (0,0)

stveep:
Done in PR above

daniellemoorhead:
Nice solution, well done :dancers:

PR #852, opened 2016-03-02
Uk/filter producers for map

stveep:
Fix for:
https://github.com/openfoodfoundation/openfoodnetwork/issues/303

This should work until OFN Ghana opens and starts selling fish!


NickWeir63:
thanks Steve

daniellemoorhead:
Hey @mkllnk can you push this through for @stveep please :)

mkllnk:
Could we actually distinguish between latitude `0` and latitude not set? Could it be `null`? Just for the people in Ghana and because it would be a good practice anyway.

stveep:
You're right - think that should do it.

daniellemoorhead:
@stveep we'll have @sstead do a quick test of this on Friday and then get it merged in.

Issue #814, opened 2016-02-06
Enable UK Offering Different trial length to enterprises


lin-d-hop:
Enterprise trial length is hard coded on the Enterprise model at 30 days. UK would like to be able to offer 6 month trials.
Proposal: To enable setting the enterprise trial length from the SuperAdmin Config -> Business Model.
This would mean 
1) adding Spree::Config[:default_enterprise_trial_length]
2) modifying the Business Model Configuration to allow trial to be set here.
3) Testing enterprise invoicing
Looks like all calculations that need to trial length and expiry refer to Enterprise.SHOP_TRIAL_LENGTH so hopefully this will be reasonably straight forward.

Also, to fit with UK business model, we need to update so that enterprises can have a minimum turnover before being charged enterprise fees. Similarly:
1) adding Spree::Config[:default_enterprise_turnover_without_fees]
2) modifying the Business Model Configuration to allow the allowed turnover before fees to be set here.
3) Testing enterprise invoicing

@oeoeaio Can you think of things that might go wrong in these proposed changes?

oeoeaio:
Nope, that generally sounds fine to me.

I would suggest that you remove "default_" from the beginnings of the suggested global config variable names, because these are not defaults, they are just the names of variables (which have defaults). I would also suggest that shop_trial_length is more indicative of what is happening than enterprise_trial_length. Also worried about the use of 'fees' to describe the billing of enterprises (could get confusing) could we make the name of the second variable something like minimum_billable_turnover or something?

First job should be fairly straight forward. I think substituting the constant on enterprise for the global config variable should work just fine.

Second job may get a bit tricky, but just be rigorous in speccing out all possible combinations of settings (by following the lead of the existing specs) and you should be fine.

myriamboure:
@lin-d-hop I think the instance should be able also to determine if they want to offer a trial period at all. I don't know if that's already in what you wrote ;-)
For example in France, we might have a business model based on "volunteer commission", which means that the platform would basically be free but each hub would define the % they want to give us on a volunteer/consicous based (and we would make it transparent for the end customer). We have not yet defined how we would like that to work, but that means we should also be able to remove the trial period :-)

oeoeaio:
Good idea @myriamboure. Also, I have done some thinking about how to implement individualised "plans" (custom trial periods, customs rates etc) for enterprises, so I'd love it if you could ping me if/when you start a new topic :)

lin-d-hop:
[...]

PR #866, opened 2016-03-16
Uk/trial length

lin-d-hop:
First stage of #814 allowing instances to set the length of their shopfront trials. Also added associated translations.


lin-d-hop:
Hi @oeoeaio CC @daniellemoorhead 
Would you take a look and feedback when you get a chance please? Would be great if these changes could be added as part of v1.6.
In particular could you let me know if you think I should act on the comments of the robot telling me to code better :-)
Merci beaucoup

daniellemoorhead:
Hi @lin-d-hop, @oeoeaio is in the mountains of Tasmania till April 1st which leaves no time for reviewing this and getting it into 1.6. Would @RohanM or @mkllnk be able to help instead?

mkllnk:
@lin-d-hop: I think that Codclimate is giving mostly good feedback. It would be good to tackle the issues. For example, I find the `?:` operator difficult to read.
```ruby
 bill = cap > 0 ? [bill, cap].min : bill
# could be simpler
 bill = [bill, cap].min if cap > 0
```

The other issue with the cyclomatic complexity is questionable, I would say. But there is a bit of code repetition. Maybe a helper method would simplify that.
```ruby
defaults = {
  fixed: :account_invoices_monthly_fixed
  rate: :account_invoices_monthly_rate
  #...
}
defaults.each do |key, config|
  # I'm not sure about the syntax `this[key]`
  this[key] = opts[key] || Spree::Config[config]
```

@daniellemoorhead: There are still a few tasks on my list. But we can do the 'one in, one out' thing. ;-)
Since I don't have my head around this completely, maybe estimate 3 hours.

PR #826, opened 2016-02-16
Bug fixes to Order Cycle Reports

lin-d-hop:
Lots of stupid bugs found in here.... and now lots of little fixes.

Could we queue this up for a production merge please?

PR #867, opened 2016-03-18
Just a few updates to #826

oeoeaio:
@lin-d-hop if you're reading this and wondering what the hell is going on, this is my logic: I had a few changes I wanted to suggest to #826, so I just made them in my local branch. I didn't want to overwrite the original branch that you made pull request #826 from, so I decided to push them to my own fork and then make a second pull request against your branch so that you could review them there and pull them in if you had no issues.

I think as a general workflow it is easier if we do all pull requests from our own forks, as this makes understanding what is going on a lot simpler. What do you reckon?

Other than that - I think this is all looking good. I'll push into production as soon as I hear back from you Re my changes

### DIFF
--- a/app/helpers/spree/reports_helper.rb
+++ b/app/helpers/spree/reports_helper.rb
@@ -11,11 +11,17 @@ module Spree
     end
 
     def report_payment_method_options(orders)
-      orders.map { |o| o.payments.first.payment_method.andand.name }.uniq
+      orders.map do |o|
+        pm = o.payments.first.payment_method
+        [pm.andand.name, pm.andand.id]
+      end.uniq
     end
 
     def report_shipping_method_options(orders)
-      orders.map { |o| o.shipping_method.andand.name  }.uniq
+      orders.map do |o|
+        sm = o.shipping_method
+        [sm.andand.name, sm.andand.id]
+      end.uniq
     end
 
     def xero_report_types
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -69,13 +69,6 @@ Spree::Order.class_eval do
     where("state != ?", state)
   }
 
-  scope :with_payment_method_name, lambda { |payment_method_name|
-    joins(:payments => :payment_method).
-      where('spree_payment_methods.name IN (?)', payment_method_name).
-      select('DISTINCT spree_orders.*')
-  }
-
-
   # -- Methods
   def products_available_from_new_distribution
     # Check that the line_items in the current order are available from a newly selected distribution
--- a/lib/open_food_network/order_cycle_management_report.rb
+++ b/lib/open_food_network/order_cycle_management_report.rb
@@ -17,7 +17,7 @@ module OpenFoodNetwork
     end
 
     def search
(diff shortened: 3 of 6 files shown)


app/helpers/spree/reports_helper.rb
```diff
@@ -11,11 +11,17 @@ def report_order_cycle_options(order_cycles)
     end
 
     def report_payment_method_options(orders)
-      orders.map { |o| o.payments.first.payment_method.andand.name }.uniq
+      orders.map do |o|
```
oeoeaio: So that we can select payment / shipping methods by id rather than name

lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -17,7 +17,7 @@ def header
     end
 
     def search
-      Spree::Order.complete.not_state(:canceled).distributed_by_user(@user).managed_by(@user).search(params[:q])
+      Spree::Order.complete.where("spree_orders.state != ?", :canceled).distributed_by_user(@user).managed_by(@user).search(params[:q])
```
oeoeaio: Bummer...we need this to be explicitly applied to spree_orders, otherwise we end up with this error when we try to scope by payment method: 'PG::Error: ERROR:  column reference "state" is ambiguous'.

Both orders and payments have a 'state' field. :(

lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -75,7 +75,7 @@ def delivery_row(order)
 
     def filter_to_payment_method(orders)
       if params[:payment_method_in].present?
-        orders.with_payment_method_name(params[:payment_method_in])
+        orders.joins(payments: :payment_method).where(spree_payments: { payment_method_id: params[:payment_method_in]})
```
oeoeaio: filtering by payment method id rather than name

lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -83,7 +83,7 @@ def filter_to_payment_method(orders)
 
     def filter_to_shipping_method(orders)
       if params[:shipping_method_in].present?
-        orders.joins(:shipping_method).where("spree_shipping_methods.name = ?", params[:shipping_method_in])
+        orders.joins(:shipping_method).where(shipping_method_id: params[:shipping_method_in])
```
oeoeaio: allows user to select multiple shipping methods to filter by (and filters by id rather than name)

app/models/spree/order_decorator.rb
```diff
@@ -69,13 +69,6 @@
     where("state != ?", state)
   }
 
-  scope :with_payment_method_name, lambda { |payment_method_name|
```
oeoeaio: don't need this anymore

lin-d-hop:
Hiya @oeoeaio 
Thanks, had a look at the changes and they all make perfect sense to me. The original reason for using the 'name' instead of 'id' is that the reports had a drop down to select payment methods etc and in super admin using ID meant there were a load of duplicates in the dropdown. Obviously not an issue now so this solution is much better.

Yep, agreed about making changes in our own forks. I started doing that originally and then switched to this way when it seemed no one minded ;-) Back to the forks.

Thanks Rob, assigning to you.

Issue #849, opened 2016-03-02
Enterprise user can require login for their shopfront and display messages to users not logged in


daniellemoorhead:
http://community.openfoodnetwork.org/t/wholesale-and-tagging/505/14

Estimate = 1 day

mkllnk:
@sstead and @oeoeaio: I need a bit more detail about this. Obviously, the products should not show without a login. But what about these:

- list of order cycles
- order cycles closed message
- custom shop message

Should they be shown?

oeoeaio:
Hi @mkllnk ,

I think we want the following:
  * Order cycles dropdown disabled
  * Show a new message that tells the user the need to login (login_required_message or something)
    * I think this should look exactly like the order cycles closed message in terms of its location on the page, but the login message should always take precedence.
  * this would mean that the custom shop message would never get rendered on the page for shops that require a login unless of course the user is logged in

Does that make sense?

mkllnk:
Okay, so we have the following order:

1. require login message
2. order cycles closed message
3. custom shop message

Sounds good.

mkllnk:
@oeoeaio: I got this ready, but there are two annoying things. The displayed message is this one:

> Please [Login](/login) if you have an account already. Otherwise, [Register as a New User](/register) to become a customer. 

The link *Login* takes you to a new page and after the login you are at the homepage, not at the shop. The Login button in the menu works differently. It's more complex markup with an Angular controller that acts if you click the login-button there. I could replicate the code for the displayed message, but
 - a) the Login link is within i18n text. That requires to squeeze all the HTML code into a variable that is then passed into translation.
 - b) we would end up with duplicate code.

That's why I left the simple link to `/login` there. Should we make a new task to simplify the login code so that you can embed it easily into text? I think it would be nice if we could still use a helper like `link_to(t('login'), login_path)` and an Angular directive would pick up the link and make it act like the current login button in the menu.

PR #871, opened 2016-03-23
Require login

mkllnk:
Solving #849.


sstead:
The 'shopfront requires login' setting is working :) potential rephrase of info tab...
image

If a shopfront requires login, this is the landing message... I think it could be made a bit more friendly (and not red, red is a angry colour :P)? But I assume this is just a placeholder text, and eventually the message will be customisable by the enterprise? In which case this message is fine for now.
image

The only issue I'm seeing is that ANY user can login to see a shopfront currently. It's not yet limited to customers on the enterprise's customer list. I can be a user that's never shopped at that shop before. Or I can be a new user and register with any email and then see the shop (that email isn't confirmed at any point).

mkllnk:
The text can be changed very easily. We can also change the style of the message. But it should be very obvious that it is something new and that you have to read it. If it would look like the normal hub message, people might not read it, because they have read it before and know that there is just some extra information. And then they might wonder why there are no products.

Should people be able to change the message text or layout?

> The only issue I'm seeing is that ANY user can login to see a shopfront currently.

Yes. That was the description. And verifying emails should be on the agenda soon.
Or should we change that to customers of that enterprise only? @oeoeaio Was was the original idea?

mkllnk:
A new version is ready, but not staged yet.

mkllnk:
I found something else. Not only customers should be allowed to see the shopfront. Admins, owners and managers of that enterprise should see it as well.

daniellemoorhead:
@mkllnk so admins, owners and managers of the enterprise can't see the shopfront at the moment, only the customers with correct permissions can?

mkllnk:
That is correct for the deployed version. But I added the rest to the branch already.

Issue #748, opened 2015-12-17
Shipping default logos 


mkllnk:
Currently, a new installation is a bit broken, because a set of logos has to be uploaded by an admin. Especially the missing header logo causes an invisible overlay that blocks clicks on shop items. We should ship a default set of logos that can be changed by an admin.

For example:

Header logo:
![ofn-logo](https://cloud.githubusercontent.com/assets/3524483/11884177/f848ca1a-a56b-11e5-926f-b8bbabffd8cd.png)

I added three images to a new branch: eb4a7c4c93bccaedf1a38d2cbfef475dbd25bd0c

The directory app/assets/images/ probably needs a clean-up as well.

PR #896, opened 2016-04-08
Fix the default logos

bingxie:
Add default logos and home_hero image. A fix for #748 .


bingxie:
@sstead For testing: 
After a new installation, user can have the default logos on the header and footer(including mobile view) and the home_hero (the big background picture).

Or if the admin can remove the uploaded logos from admin console, user will see the default logo.

mkllnk:
@sstead We have to check four images:

- the logo in the menu
- the logo in the menu on a small screen (window)
- the logo in the footer
- the background hero picture

Have a look at them before we stage the new code. The first two logos are configured on staging2 and are displaying. The other two got lost at some point.

After I pushed the new code to staging2, the first two should be the same as before and the last two should magically appear. :-) You can upload other versions of the images and see that it's still working.

That's all, I think.

sstead:
the logo in the menu - 
image

the logo in the menu on a small screen (window)- 
image

the logo in the footer- 
image

the background hero picture- grey

sstead:
After staging2 was loaded with new fix...
The logo in the menu- image
the logo in the menu on a small screen (window) - 
image
the logo in the footer- defaults to... 
image
the background hero picture- defaults to... 
image

Can I change each image in config? Yes.

Issue #870, opened 2016-03-22
Text change on 'listing orders' screen


NickWeir63:
On the ‘listing orders’ screen (see screenshot) the headings ‘payment county’ and ‘shipping county’ should read ‘payment state’ and ‘shipping state’.  
![listing orders](https://cloud.githubusercontent.com/assets/10517702/13955625/4705ea2c-f03d-11e5-83b0-dee8ba5f2909.png)

lin-d-hop:
Due to the fact that we only have 1 translation for the word 'State'. This will obviously be an issue for all languages apart from Australian English :-)

lin-d-hop:
Ok so this was actually from the Spree I18N project.

I've submitted a PR here:
https://github.com/spree-contrib/spree_i18n/pull/668

Will close once that PR has been merged and the changes will flow through to us.

stveep:
In the mean time, this can be overridden locally in `config/locales/en-GB.yml` so we don't need to wait for the spree_i18n PR:

https://github.com/stveep/openfoodnetwork/commit/27fdfb60af61ca2fdd1ed7484df028502b894b69

Should I set up a PR for this into master, or do we maintain our en-GB file somewhere else @pmackay?

lin-d-hop:
Nice one @stveep Thanks for helping out on that.

This will go into master as all is maintained in the main code base.

lin-d-hop:
Hope you don't mind me assigning to you @stveep to get a PR in for your solution :-)

lin-d-hop:
Hi @daniellemoorhead @mkllnk Could this be merged to master for the next release please? thanks :-)

mkllnk:
Should be easy. It's a small change. @daniellemoorhead I'm happy to take this on. Estimate 1 hour.

daniellemoorhead:
Go for it @mkllnk

mkllnk:
@lin-d-hop Merged. :-)

lin-d-hop:
See this on UK production now (side effect of updating translation file for v1.6.0).
Closing.

PR #899, opened 2016-04-11
State translation override

stveep:
Workaround for https://github.com/openfoodfoundation/openfoodnetwork/issues/870


mkllnk:
I think I should copy these changes to `en.yml` as well. Since `en.yml` is our source file in Transifex, people will be able to translate these strings in case there are more missing or strange Spree translation files.

PR #908, opened 2016-04-15
Account balances

stveep:
Showing incomplete payments:

http://community.openfoodnetwork.org/t/customers-can-view-their-payments-and-balance/308/43?u=stveep


sstead:
Testing notes:
Accounts and billing is visible- @oeoeaio I'm not sure this is a bad thing. Some instances might want this to be visible to their enterprise users.
image

The screenshot below shows a successful partial payment ($5) that's been put through manually (edit order > payments). 
It also has some issues with the Invalid text... 
image.

stveep:
Thanks for testing! The text issue looks like a missing translation. I added it to the main en.yml in the 2nd commit though and that fixed it for me locally so I'm not sure why you see it.

Is the spec test case for the accounts distributor appropriate? I don't have one in my test dataset so couldn't check it directly. Let me know if you decide about it and I can look at it more.

mkllnk:
I had to manually rebuild the locale to get the text change displayed correctly. I don't know why rails is not always picking up changes.

I actually don't know about the accounts distributor. Who requested that behaviour?

PR #909, opened 2016-04-15
Updating en-GB file to reflect recent additions

lin-d-hop:


lin-d-hop:
@mkllnk @daniellemoorhead 
Please could this be merged to master before the next release. 

I've put these changes in a branch on the main repository as well so that I could deploy today to UK production (v1.6.0 functionality needed translation updates, something we need to check before every deploy... good to note).

mkllnk:
I'm wondering if we need a better process for this. You want to copy over new strings. You want to delete removed strings. But you don't want to override differently translated strings. Do you know if Transifex supports that?

The other projects don't modify the locale files themselves. They translate in Transifex and we automatically commit translated files to our Transifex branch which is then manually merged into master before the next release.

lin-d-hop:
I agree we need a better process. Thanks for getting this merged so quickly!

I'll take a closer look at Transiflex, sounds like a good solution :-)

PR #915, opened 2016-04-22
Delete customer

mkllnk:
- No flash notification when deleting a customer.
- Revert associating new accounts with existing customers.

To test this:

- Place an order as a guest.
- Sign up with the same email address.
- You should not be able to see the guest order.
- Place an order as a user.
- You should be able to see your order.


sstead:
I'm not able to delete customers 
Your 'to test this' instructions=correct, working as you described.

PR #930, opened 2016-04-29
Openstreetmap tiles

bingxie:
We just used the openstreetmap tiles server and kept using google's API.

The detail discussion is here: http://community.openfoodnetwork.org/t/open-street-map-vs-google-maps/492/4


almereyda:
Please note that you are violating the Google Maps terms of service, if you *don't* display the output from the Google Maps API within the Google Maps JS module.

PR #934, opened 2016-04-29
More intuitive logic for minimum billable turnover

oeoeaio:
Hi @lin-d-hop,

Thanks for this, code looks great, and I think we don't need to worry about the code climate issues too much - we are thinking that we might tweak code climate a bit to give us a bit more room to wriggle anyway. I just a have slight change that i think is not too costly in terms for desired functionality but will make configuration more intuitive.

Scenario that is affected is: when we don't want to apply a fixed charge to shops that sell exactly nothing in a given month, but we do want to charge them a fixed price if they sell anything at all.

I think this is more intuitively handled by setting minimum_billable_turnover to something like $1 is this case and using $0 to levy a fee against everyone regardless of turnover, rather than using $0 for the case when we don't want to charge a fixed price for zero turnover, and -$1 when we do.

I think this is better because:
1) We don't have to specify a non-intuitive figure of -$1 to get the desired functionality.
2) The semantics play out better. ie. minimum_billable_period is always inclusive, you are only exempt if you sell less than the amount specified, including $0.
3) The likelihood of an instance wanting to distinguish between those who sell $0 worth of produce and those who sell $1 worth of produce and to levy a fixed fee against only the latter seems remote.
4) Furthermore, even if this is the desired functionality, using a minimum billable turnover of $1 should be a perfectly acceptable way of handling this, given that the chance of a business selling between $0 and $1 worth of produce in a month is almost impossible, and inconsequential.

Let me know what you think. :)

### DIFF
--- a/app/assets/javascripts/admin/business_model_configuration/controllers/business_model_configuration_controller.js.coffee
+++ b/app/assets/javascripts/admin/business_model_configuration/controllers/business_model_configuration_controller.js.coffee
@@ -10,7 +10,7 @@ angular.module("admin.businessModelConfiguration").controller "BusinessModelConf
     Math.min($scope.bill(), Number($scope.cap))
 
   $scope.finalBill = ->
-    return 0 if Number($scope.turnover) <= Number($scope.min_bill_to)  
+    return 0 if Number($scope.turnover) < Number($scope.minBillableTurnover)
     $scope.cappedBill()
 
   $scope.capReached = ->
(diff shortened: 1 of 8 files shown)


lin-d-hop:
Hi @oeoeaio,

That all sounds fine to me. My thinking originally was based around my thinking that the flat rate was to be charged whether the enterprise made sales or not. Given that this is not the desired functionality I think your changes are an improvement and more intuitive.

Thanks for taking the time to review and update!
:-)

PR #942, opened 2016-05-04
Use save bar

bingxie:
- Extends the save bar to support multiple buttons
- Uses save bar in order cycles editing page


app/views/admin/order_cycles/_form.html.haml
```diff
@@ -52,14 +52,10 @@
 .actions
   - if @order_cycle.new_record?
     = f.submit 'Create', 'ng-click' => "submit('#{main_app.admin_order_cycles_path}')", 'ng-disabled' => '!loaded()'
-  - else
-    = f.submit 'Update', 'ng-click' => "submit(null)", 'ng-disabled' => '!loaded()'
```
RohanM: It looks like we might be losing some functionality here - buttons should be disabled until `loaded()` is true.

Issue #938, opened 2016-05-03
Google Analytics number is hardcoded


pmackay:
Looking at `https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/views/shared/_analytics.html.haml`, it seems to have a single GA `UA` code in it. Should this be changed to insert some Spree code that supports their trackers mechanism? Or make the `UA` number a variable?

@RohanM @mkllnk @oeoeaio please comment :)

cc @lin-d-hop

oeoeaio:
I don't have much experience with GA but I think that is a sensible idea...

mkllnk:
I think GA should be optional. There are good reasons not to use it:

- Not everybody wants to give their user data to Google.
- It can slow down the website.
- There are alternatives.

How does Spree's tracking system work?

pmackay:
See https://spreecommerce.com/blog/ecommerce-tracking-in-spree. Need to do this to get it to work:

> If the Javascript is not present, be sure you are still including the spree/shared/google_analytics partial in your application layout.

Then it would be optional because its configurable via Spree admin. However it seems like it only works for commerce checkouts, not all pages. If so, how important is tracking all pages? How often do you use the tracking you currently have? Would it be ok to just remove the current hardwired tracking code? 

@mkllnk have you used any alternatives before?

mkllnk:
No, I haven't used any alternatives.

I personally would be happy to remove the Google Analytics code. But I think some people would like to keep it.

@daniellemoorhead:

- Are we using Google Analytics?
- Who is interested in it?
- Do we have capacity to analyse the data?
- Do we have capacity to act on the insights?

daniellemoorhead:
@mkllnk it's not an option to remove the analytics code. Definitely not.

- We do use Google Analytics, @sstead reports on stuff each month for the OFN au board.
- Not sure what you mean by who is interested in it...?
- See first point
- We're not using it for product development (is that what you mean?), but we will need insights if we get funding to redo the front end experience.

The bigger question is how can we set this up so that every local instance can have it's own version running? Was this considered when it was added in the first place? @RohanM or @oeoeaio or @serenityhill can you shed any light on this?

pmackay:
@daniellemoorhead could you please add a bit more detail about what info from the analytics you use? Is it important to have stats for all the pages accessed?

Am asking because Spree has some capacity for GA built in, which is configurable, but it may only work when shop checkouts are made. How much of a problem would that be?

[...]

PR #991, opened 2016-05-11
Switch to using the Spree GA tracker code

pmackay:
Fix for #938. Would be good to test this on a staging server and verify analytics are being tracked. Would require setting up a UA code for staging as well (but with Spree trackers, that can now be done!)

Copy created with `uk/` prefix.


pmackay:
This has been tested on UK staging and works for basic page views.

daniellemoorhead:
Hiya @pmackay I'm still a bit concerned about this solution - can you tell us what it will do to the way we access Analytics now? Will there be any change? Given it's the site admins that will be using GA I want to make sure that included in this issue/PR is a description of how it changes the current way we do things (if at all). Are you able to provide this?

pmackay:
I believe it will work the same, tracking basic page views. Once this is merged, you will have to enter the Analytics `UA` number into the "Trackers" config in Spree admin. I did that for UK staging and could see realtime page views being recorded in GA. You can then have a different `UA` number for staging and production if that's of interest (not possible before).

Does that help? If Aus is ok for this to be merged, I'll update the deployment docs with some notes on configuring GA.

pmackay:
Also note that it has a commerce analytics tracking support, feeding into the GA commerce support specifically as well. I've not tested that, but its Spree code so would assume is likely to be ok.

daniellemoorhead:
So what will this mean for the existing data we have, will it change, will there be a blip in the recording? Will it reset?

pmackay:
There would be a slight blip between when the code is updated to the time you add the setting in Spree admin. Existing data should not change I think. I havent tested this, but there is no reason any data should get lost. And it should continue adding data to the same profile if the same `UA` code is used.

daniellemoorhead:
So, this means that we'll be able to look at data on a month on month or year on year basis, we won't have to restart from scratch with this kind of trending data view? If it doesn't impact on that, then go right ahead and merge it :)

pmackay:
Isnt it usually Rohan, Maikel or Rob who merge stuff?

daniellemoorhead:
All of them, but given @mkllnk has been across this it should probably be him.

daniellemoorhead:
Have asked @RohanM to have a look at this in terms of solution, if all is well then we'll push it as part of release 1.8 next week.

Issue #1009, opened 2016-05-19
Producer email contains products for a cancelled order


Rocksj:
I placed an order for some biscuits last week as a demo to someone visiting and then cancelled the order. I then clicked on Notify Producers and received the following email:

Dear Tamar Valley Food Hub,

We now have all the consumer orders for the next food drop.
Orders summary

Here is a summary of the orders for your products:

- Tamar Valley Food Hub - Biscuits - Shortbread - 200g (QTY: 1) @ £1.52 = £1.52
- Tamar Valley Food Hub - Seasonal Veg Bag with Potatoes (5 vegetables) (QTY: 1) @ £4.35 = £4.35

Thanks and best wishes, Tamar Valley Food Hub Tamar Valley Food Hub, Tamar Valley, PL17 7AD 01579 208412 sara@tamarvalleyfoodhubs.org.uk

PR #1011, opened 2016-05-21
Uk/1009 canceled orders in email

stveep:
Fix for issue #1009 .

This was caused by completed orders being defined by a completed_at timestamp, rather than by the state (which can change after completion - e.g. to canceled). Perhaps we should also deal with the 'resumed' state - is resuming a canceled order possible?


oeoeaio:
In master

PR #1018, opened 2016-05-26
en-US

MikeiLL:
This looks like a pretty messy PR. The only thing I'm trying to PR is the `en-US` file so apologies for the mess. Since we (stubbornly) insist on using Pounds and Inches instead of the Metric system here, I added one string: `weight_measurement_unit: "Weight (per lb)"` which I was hoping we can use instead of a hard coded string in `models/open_food_network/calculator/weight.rb`. I experimented briefly, but unsuccessfully.

### DIFF
--- a/config/locales/en-US.yml
+++ b/config/locales/en-US.yml
@@ -1,0 +1,1048 @@
+# English language file
+# ---------------------
+#
+# Originally adapted from source language file maintained by the Australian OFN team.
+# Visit Transifex to translate this file into other languages:
+#
+#   https://www.transifex.com/open-food-foundation/open-food-network/
+#
+# If you translate this file in a text editor, please share your results with us by
+#
+# - uploading the file to Transifex or
+# - opening a pull request at GitHub.
+#
+#
+# See http://community.openfoodnetwork.org/t/localisation-ofn-in-your-language/397
+
+en-US:
+  activerecord:
+    # Overridden here due to a bug in spree i18n (Issue #870)
+    weight_measurement_unit: "Weight (per lb)"
+    attributes:
+      spree/order:
+        payment_state: Payment State
+        shipment_state: Shipment State
+  devise:
+    failure:
+      invalid: |
+        Invalid email or password.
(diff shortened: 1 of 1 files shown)


pmackay:
This cannot be merged in its current state. Is there an issue related to the weights problem? Would it be worth discussing there first the best solution? If it is to have an `en-US` locale, I'd suggest a PR that just has that change. Its not correct to include file mode changes, log files, hardcoded paths for `ofn_america`, removing database.yml and so on.

MikeiLL:
I understand that, which I thought I had made clear in my comment. Will try to send a new PR with only the `en-US` file. Just not sure what that will entail yet. Was hoping you could merge only that new file in.

MikeiLL:
OK. Apologies for the mess. This should do the trick.

MikeiLL:
And yes - there is an issue with weight, because Americans are used to buying produce by the pound. Like everything else international, the metric system scares us. I imagine that the string in `models/open_food_network/calculator/weight.rb` should be i18n anyway rather than hard coded with the word "weight".

daniellemoorhead:
Hiya @MikeiLL we've added time next week in our schedule for @mkllnk to review this, @sstead test it, and then to put it into the dot point release that'll go live by the end of the week. Cheers!

mkllnk:
Hi @MikeiLL!

I can merge the current pull request with just the locale. But I would like to mention that whenever you update to a newer version of OFN, you will need to update the locale as well and translate new strings. Most people use Transifex for that. So if you are happy to maintain the en-US locale, that would be great. You are also very welcome to open a pull request translating strings that are currently hard-coded. The people in France and Norway would be very happy about that as well.

myriamboure:
Ping @pacodelaluna for info

daniellemoorhead:
@mkllnk @MikeiLL can this be closed given Maikel merged the commit to master last year? Or is there more in it that hasn't been finished off yet?

mkllnk:
@daniellemoorhead This is closed. But people can comment on closed pull requests. I don't think any further action is required.

Issue #577, opened 2015-06-01
Software version can be displayed in the UI or Admin interface


pmackay:
It would be very helpful for testers if there was an obvious way to know what version of OFN was running when looking at web pages, rather than having to use git versions to work this out. 

Would best practice here be to add a `VERSION` file in the project root, that is updated on each major version update? 

@lin-d-hop any thoughts?

Em-AK:
@emaestracci and @knopfler81 implemented this by relying on the last existing git tag.

pmackay:
@Em-AK is there a commit for that?

Em-AK:
Yes here for the moment: https://github.com/knopfler81/openfoodnetwork/commit/7aa1a3f892bd56237457436f7c83e96e3408a8b7

Em-AK:
Update here https://github.com/WildCodeSchool/openfoodnetwork/pull/3

pmackay:
@Em-AK what's needed to get the PR for this completed?

bingxie:
I have used this gem(https://github.com/dtaniwaki/rack-dev-mark) to show different things for different envs. And it's very beautiful.

bingxie:
![snip20160601_1](https://cloud.githubusercontent.com/assets/1144652/15695762/f8049c14-27ed-11e6-93db-fd03131d5a64.png)
![snip20160601_2](https://cloud.githubusercontent.com/assets/1144652/15695761/f803a16a-27ed-11e6-897f-933ba9b6cb56.png)

daniellemoorhead:
Hey @bingxie have you put this into staging for @sstead to test?

bingxie:
Need someone to review the PR, if there is no comments, @sstead can stage it and have a test. Thanks

daniellemoorhead:
Work was done in #1021 but a new approach was discussed on that PR that isn't yet implemented. Would be good to have someone pick this up and implement the described approach:

>I really like the idea of using GitHub's webhooks to listen for release events and to then update the version number accordingly.

sauloperez:
Can someone describe that other approach?

> I really like the idea of using GitHub's webhooks to listen for release events and to then update the version number accordingly.

To me it looks overly complex. What's wrong with having a VERSION file in each deploy and reading the number from it? I know it's not as cool as a webhook :trollface:

mkllnk:
> What's wrong with having a VERSION file in each deploy and reading the number from it?

[...]

PR #1021, opened 2016-06-01
Show version

bingxie:
Display the version on two places.
snip20160601_1
snip20160601_2

### DIFF
--- a/config/initializers/version.rb
+++ b/config/initializers/version.rb
@@ -1,0 +1,14 @@
+module GitRepo
+  def self.last_tag
+    begin
+      IO.popen("git describe --tags --abbrev=0 "){ |v| v.readline.strip }
+    rescue
+      I18n.t(:version_not_found)
+    end
+  end
+end
+
+module OpenFoodNetwork
+  VERSION = GitRepo.last_tag
+end
+
--- a/app/assets/stylesheets/admin/navbar_version.css.scss
+++ b/app/assets/stylesheets/admin/navbar_version.css.scss
@@ -1,0 +1,4 @@
+#app-version {
+  float: right;
+  font-size: 80%;
+}
--- a/app/assets/stylesheets/admin/navbar_version.css.scss
+++ b/app/assets/stylesheets/admin/navbar_version.css.scss
@@ -1,4 +1,4 @@
-#app-version {
+#app_version {
   float: right;
   font-size: 80%;
 }
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,6 +16,14 @@ module ApplicationHelper
   end
 
 
+  def app_version
+    begin
+      IO.popen("git describe --tags --abbrev=0 "){|v| v.readline.strip}
+    rescue
+      t(:version_not_found) 
+    end
+  end
+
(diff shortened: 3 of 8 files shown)


config/initializers/version.rb
```diff
@@ -0,0 +1,14 @@
+module GitRepo
+  def self.last_tag
+    begin
+      IO.popen("git describe --tags --abbrev=0 "){ |v| v.readline.strip }
```
mkllnk: I think the `--abbrev=0` could be omitted so that people know if they are running the latest code of a branch instead of a release. But it's okay as is. Definitely an improvement.

mkllnk:
I think this pr can go ahead.

oeoeaio:
Hey @mkllnk and @bingxie, Rohan and I are just looking at this, and we've noticed a bit of an issue. The logic here depends on the existence of a tag for the version number, which is very dependent on the particular workflow being used, and also on the way that tags are used.

For example, the last tag on our production server in Aus will always be the **previous** version, rather than the current one, because the current release is generally only created in GitHub **after** we have pushed to production (which is when tags on the server will be updated). This means that this logic will work quite well on Aus production for intermediate pushes to prod, but not at all for actual releases.

The second consideration is that it restricts the use of tags to releases only - not sure how big an issue this is, but probably not a good idea to assume this will always be the case.

We thought it might be better to use the GitHub API to determine whether the current server commit is a release commit, and if so, to use that version number. This request could be cached, and only invalidated on server initialisation.

Where the server commit does not match a release commit, we can just display the SHA - this will probably only ever happen on Australian production...

How does that sound?

mkllnk:
That's a good point. But since we deploy quite regularly, it's not a big deal.

If we wanted to use Github integration, it could be better to listen to release events and act on that. For example do a `git fetch` and reload the version number. We could use that infrastructure later to inform admins about new releases as well.

bingxie:
Using the github releases API is a very good idea. I went through the API doc, I can only find get a release by tag name API. It still rely on the tag name. I am not sure how can we match a release with server commit.

mkllnk:
Github always uses a tag for a release. You can't have a release without tag. That's why I think that `git describe` is a really good solution. The question is just: When do we update our local data about the release?

Even if we use the Github API and retrieve the information on deploy of our production server, we have the same problem that the release doesn't exist yet. We will have an outdated version number on our production server.

bingxie:
@oeoeaio How should we make decision on this issue?

oeoeaio:
@bingxie @mkllnk @RohanM 

I really like the idea of using GitHub's webhooks to listen for release events and to then update the version number accordingly. How does everyone else feel about that?

bingxie:
👍

mkllnk:
That's a good idea. So we extend this pull request so that we update the tag information based on a webhook call. That's cool.

daniellemoorhead:
---Doing a PR Cleanup---

@oeoeaio @mkllnk do you think this is worth keeping and building on this or should the PR be closed?

mkllnk:
It would be good to do this. Seems really useful.

daniellemoorhead:
Closing this, comment added to #577 issue that can be picked up and implemented via a new PR.

PR #1029, opened 2016-06-06
i18n calculator/weight (fixes #774)

### LINKED ISSUE #774: Weight (per kg) Calculator not translated
The OpenFoodNetwork defined weight calculator string is not translated. 

It seems that the string is hardcoded in `app/models/open_food_network/calculator/weight.rb`

How should we localise that string?

weight not translated

elf-pavlik:
This commit  should address #774 

I guess we need to address support for localized weight units in another issue?

ping @MikeiLL


elf-pavlik:
oops! I just noticed #825 

it didn't reference #774 

i guess we can just close this PR and merge the earlier one

Issue #587, opened 2015-06-04
Customers: Duplicate customer code doesn't save but doesn't tell me why, need message


kirstenalarsen:
On Customer Accounts page admin/customers
![image](https://cloud.githubusercontent.com/assets/2911176/7976952/eb8860c8-0ac6-11e5-8a38-6fa13bad6fa6.png)

oeoeaio:
This is particularly an issue when trying to remove a customer code, because this is saved as an empty string rather than null.

PR #1037, opened 2016-06-15
Duplicate customer codes

mkllnk:
Fixing #587 

* Added javascript to warn about duplicate codes as you type.
* Added before-validation step in the Customer model to save an empty code as `nil`. This avoids duplicate empty strings that are not allowed.
* Removed some random duplicate markup on the dash board and tried to stabilise randomly failing specs.


sstead:
image

Issue #1025, opened 2016-06-02
Configurable 'From' Address for email communication from Enterprises


lin-d-hop:
Currently sends from instance address, which is not very useful for the producers replying or the instance coordination.

Enabling emails to be sent 'from' a different email address needs to be instance configurable as some instances may be using SMTP servers that overwrite the 'from' address.
- Add checkbox in Mail Methods panel in super admin configuration enabling Enterprise communication to be sent from the enterprise email address.
- Update the from address of appropriate emails to use hub 'from' address when this config is set.

lin-d-hop:
Hi @RohanM @stveep 

Just a quick one...
 
We are getting some people replying to the OC notification emails to hello@open.....uk
Looking into this although the 'reply to' email is set to the coordinators email the 'from' email is still the instance.
```
mail(to: @producer.email,
           from: from_address,
           subject: subject,
           reply_to: @coordinator.email,
           cc: @coordinator.email)
```
It seems not all mail clients respect the 'reply_to' and use the 'from' instead.

I was just wondering if there was a reason it was implemented this way? Would you be happy if I just changed the 'from' to the coordinator as well?

Cheers!

daniellemoorhead:
@oeoeaio any thoughts on this?

daniellemoorhead:
Hey @kirstenalarsen are you able to shed some light on the above? Can @lin-d-hop change the from address to be the coordinator as well?

oeoeaio:
I don't have any particular concerns with this change. Sounds good to me?

lin-d-hop:
Hi @stveep 
Thanks for agreeing to look at this when you get the chance.

Fundamentally this change is quite simple. PR #1044 worked fine for our servers.
We just need to add in a UI step in Super Admin -> Configuration -> Mail Methods to allow this to be a feature that is turned on or off at instance level. This is because some instances will be using SMTP servers that override all outgoing emails to the same email address, rendering this feature useless.

I'll communicate via Slack on deployment process to make this easier to test.

Cheers :-)

lin-d-hop:
Closing as this issue blew up in scope and practicalities

PR #1044, opened 2016-06-19
Update from address in notification emails

lin-d-hop:
Quick update to the 'from' email in producer order summary emails: change to from enterprise instead of from instance.

### DIFF
--- a/app/mailers/producer_mailer.rb
+++ b/app/mailers/producer_mailer.rb
@@ -16,7 +16,6 @@ class ProducerMailer < Spree::BaseMailer
       mail(to: @producer.email,
            from: @coordinator.email,
            subject: subject,
-           reply_to: @coordinator.email,
            cc: @coordinator.email)
     end
   end
--- a/spec/mailers/producer_mailer_spec.rb
+++ b/spec/mailers/producer_mailer_spec.rb
@@ -44,8 +44,8 @@ describe ProducerMailer do
     ActionMailer::Base.deliveries.count.should == 1
   end
 
-  it "sets a reply-to of the enterprise email" do
-    mail.reply_to.should == [s1.email]
+  it "sets a from of the enterprise email" do
+    mail.from.should == [s1.email]
   end
 
   it "includes receival instructions" do


lin-d-hop:
@oeoeaio Would you merge if happy?

daniellemoorhead:
@lin-d-hop going to get @mkllnk to look at this rather than @oeoeaio who is busy up to eyeballs with other UK stuff :)

mkllnk:
Is the coordinator triggering that email?

If yes, do we still need the reply-to field?

lin-d-hop:
Good thinking @mkllnk :-)

sstead:
Hi @lin-d-hop 
I couldn't see the desired changes when I tested this today. In my inbox, the 'notify producers' email was still from 'admin@ofw..' instead of from the enterprise (see below)...

image

If I clicked reply to the older email, it would be addresed to the hub's coordinator. If I clicked reply in staging, it was now being addressed to admin@ofw.

Maikel said it could be something at our end that's preventing your change from showing. Or perhaps I don't understand what should be changed?

lin-d-hop:
Hi @sstead 
Totally confused as to how that change hasn't worked.
You understand the change correctly.
Shall I test on our staging server and see what happens on our end?

mkllnk:
@lin-d-hop This confused me a lot as well. But I have two ideas what it might be.

1. Our mail method has a from address. The spree interface says, "Send all mails as from the following address." It could be that spree overrides the from address. But I couldn't find that in the code.

2. We are using Gmail to send our emails. That is like a personal email account. It might be that Gmail is changing the from address to the logged in user to prevent fraud.

If it's Spree overriding the address, that can be an important feature, because some email gateways accept only certain pre-defined email addresses. An email from a random coordinator would be rejected. If this pull request would work, it could actually break sending emails if you use authsmtp.com with restricted sender addresses.

If we want to support sending those emails from user email addresses, we need that as an optional feature that can be turned off. Sorry, I didn't think of all that when I first looked at your pull request. But now I think we can't merge it in the current form.

lin-d-hop:
@mkllnk 
I deployed this to UK staging to test whether there was an unknown place in Spree this is overwritten or if it is gmail.

I know that to use gmail as an smtp server you have to overwrite the from address so suspected this was likely the cause.
Testing on UK staging (we use sendgrid) the email came through fine and dandy with the from address updated.

So I think we go with making this an optional change in the Mail Methods panel, with a note that this might be overriden by some SMTP servers. Will update now in #1025

mkllnk:
Sounds good. Ping me once that's done.

daniellemoorhead:
Removing the Aus Review label on this till you're ready for us to look at it again @lin-d-hop :)

daniellemoorhead:
---Doing a PR Cleanup---

@lin-d-hop thoughts on what to do with this PR? Was it replaced by #1203 as a fix for #1025 and so can be closed?

lin-d-hop:
Closing this PR as this is far too simple an implementation for what we actually need

Issue #1035, opened 2016-06-13
All enterprises loaded into the shopfront on every page load


daniellemoorhead:
As discussed in UK Hangout, Rob has been looking at front-end performance in response to [Lynne's post](http://community.openfoodnetwork.org/t/improving-shopfront-load-speed-for-slow-clients/576) on discourse. 

He has found that enterprise are being serialised and loaded into all checkout (shopfront, cart, checkout, order confirmation) pages, which means a 2.5-3.5 second hit on each of these pages on Australian production. It also likely contributes to poor performance on devices where memory is limited.

The primary reason that this is in place is to ensure the correct data is available for enterprise modals on the various pages.

Strategy to resolve this would be to build async loading of data when a modal is available, rather than loading all possible data into the pages on the off-chance that it is required.

daniellemoorhead:
@rob to run @mkllnk through his thinking on this.

NickWeir63:
Thanks @daniellemoorhead and @rob it would make a big difference to the saleability of OFN in the UK if we could improve the load speed.
Thank you for trying!

pmackay:
@daniellemoorhead am wondering if you mean @oeoeaio above? 

Also has there been previous discussion about enterprises all having permanent links? Do we need to maintain the modals at all? Permanent pages would have a bunch of benefits. Would it make sense to refactor to achieve that rather than changing so modals load on demand?

mkllnk:
@pmackay I agree that it would be great to have individual pages for enterprises. That is something that has been requested a lot of times.

daniellemoorhead:
@pmackay I also agree. Are you able to do a search on github/discourse and see if it's something that has been logged already, and if not create a new item for it? There's no money at present to do this work but it's definitely worthy of inclusion in the backlog (if it isn't already there). Cheers!

mkllnk:
Okay, I'll make some notes here just for myself and for whoever is interested.

Budget: 15h

There are nine pages that load all enterprises. Most of them don't need all enterprises loaded upfront:
```
app/views/checkout/edit.html.haml editing your cart
app/views/enterprises/shop.html.haml the shop front
app/views/groups/index.html.haml listing all groups
app/views/groups/show.html.haml showing a single group
app/views/home/_hubs.html.haml the list of shops
app/views/map/index.html.haml the map with all enterprises
app/views/producers/index.html.haml the list of producers
app/views/spree/orders/edit.html.haml changing an order
[...]

PR #1073, opened 2016-06-29
WIP: Load only needed enterprises for the shop front

mkllnk:
Addressing #1035.

The work is still in progress.

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/enterprise_modal.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/enterprise_modal.js.coffee
@@ -1,9 +1,15 @@
-Darkswarm.directive "enterpriseModal", ($modal)->
+Darkswarm.directive "enterpriseModal", ($modal, Enterprises, EnterpriseResource) ->
   restrict: 'E'
   replace: true
   template: "<a ng-transclude></a>"
   transclude: true
-  link: (scope, elem, attrs, ctrl)->
-    elem.on "click", (ev)=>
+  link: (scope, elem, attrs, ctrl) ->
+    elem.on "click", (ev) =>
       ev.stopPropagation()
+      params =
+        id: scope.enterprise.id
+      EnterpriseResource.relatives params, (data) =>
+        Enterprises.addEnterprises data
+        scope.enterprise = Enterprises.enterprises_by_id[scope.enterprise.id]
+        Enterprises.dereferenceEnterprise scope.enterprise
       scope.modalInstance = $modal.open(controller: ctrl, templateUrl: 'enterprise_modal.html', scope: scope)
--- a/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
@@ -7,10 +7,12 @@ Darkswarm.factory 'Dereferencer', ->
       unreferenced = []
       if source && target
         for object, i in source
+          # skip empty entries in sparse array
+          continue unless source.hasOwnProperty(i)
           key = object?.id
           if data.hasOwnProperty(key)
             target[i] = data[key]
           else
             delete target[i]
-            unreferenced[i] = source[i]
+            unreferenced[i] = object
       unreferenced
--- a/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
@@ -1,6 +1,8 @@
 Darkswarm.factory 'Dereferencer', ->
   new class Dereferencer
-    dereference: (array, data)->
+    dereference: (array, data) ->
       if array
         for object, i in array
-          array[i] = data[object.id]
+          key = undefined
+          key = object.id if object
+          array[i] = data[key]
--- a/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
@@ -13,6 +13,5 @@ Darkswarm.factory 'Dereferencer', ->
           if data.hasOwnProperty(key)
             target[i] = data[key]
           else
-            delete target[i]
             unreferenced[i] = object
       unreferenced
--- a/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
@@ -1,8 +1,16 @@
 Darkswarm.factory 'Dereferencer', ->
   new class Dereferencer
     dereference: (array, data) ->
-      if array
-        for object, i in array
-          key = undefined
-          key = object.id if object
-          array[i] = data[key]
+      @dereference_from(array, array, data)
+
+    dereference_from: (source, target, data) ->
+      unreferenced = []
+      if source && target
+        for object, i in source
+          key = if object then object.id else undefined
+          if data.hasOwnProperty(key)
+            target[i] = data[key]
+          else
+            delete target[i]
+            unreferenced[i] = source[i]
+      unreferenced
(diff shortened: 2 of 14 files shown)


app/assets/javascripts/darkswarm/directives/enterprise_modal.js.coffee
```diff
@@ -1,9 +1,15 @@
...
-    elem.on "click", (ev)=>
+  link: (scope, elem, attrs, ctrl) ->
+    elem.on "click", (ev) =>
       ev.stopPropagation()
+      params =
+        id: scope.enterprise.id
+      EnterpriseResource.relatives params, (data) =>
+        Enterprises.addEnterprises data
+        scope.enterprise = Enterprises.enterprises_by_id[scope.enterprise.id]
+        Enterprises.dereferenceEnterprise scope.enterprise
```
RohanM: If the constructor of the Enterprises service dereferences all enterprises, should `addEnterprises` automatically dereference any new enterprises?

app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
```diff
@@ -1,6 +1,8 @@
 Darkswarm.factory 'Dereferencer', ->
   new class Dereferencer
-    dereference: (array, data)->
+    dereference: (array, data) ->
       if array
         for object, i in array
-          array[i] = data[object.id]
```
RohanM: Perhaps coffeescript's `?.` operator would make this simpler?

app/controllers/enterprises_controller.rb
```diff
@@ -11,19 +13,28 @@ class EnterprisesController < BaseController
 
   respond_to :js, only: :permalink_checker
 
+  def relatives
+    respond_to do |format|
+      format.json do
+        enterprises = Enterprise.find(params[:id]).andand.relatives.activated
```
RohanM: I think you might need another `andand` between `relatives` and `activated` for this to work when enterprise is not found.

app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
```diff
@@ -1,8 +1,16 @@
...
-          key = undefined
-          key = object.id if object
-          array[i] = data[key]
+      @dereference_from(array, array, data)
+
+    dereference_from: (source, target, data) ->
+      unreferenced = []
+      if source && target
+        for object, i in source
+          key = if object then object.id else undefined
```
RohanM: Would `object?.id` work here?

app/assets/javascripts/darkswarm/directives/enterprise_modal.js.coffee
mkllnk: It could, but it's not helpful in this case.
When a user opens the modal of a producer, the relatives of that producer are loaded and added. These enterprises don't need dereferencing in this case. The producer needs dereferencing, so that its shops can be displayed.

app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
mkllnk: I rewrote this in a later commit with if-then-else. Coffeescript doesn't have the `?` operator while javascript has. http://stackoverflow.com/questions/10146080/ternary-operation-in-coffeescript

app/models/enterprise.rb
```diff
@@ -220,6 +220,22 @@ def relatives
     ", self.id, self.id)
   end
 
+  def relatives_and_oc_producers(order_cycles)
+    enterprise_ids = []
+    order_cycles.each do |oc|
+      enterprise_ids += oc.exchanges.incoming.pluck :sender_id
+    end
```
RohanM: In terms of SQL queries, this section is O(n) in the number of order cycles. We could make it O(1) like this:

```ruby
Exchange.in_order_cycle(order_cycles).incoming.pluck(:sender_id)
```

app/models/enterprise.rb
```diff
@@ -220,6 +220,22 @@ def relatives
...
+    Enterprise.where("
+      enterprises.id IN
+        (SELECT child_id FROM enterprise_relationships WHERE enterprise_relationships.parent_id=?)
+      OR enterprises.id IN
+        (SELECT parent_id FROM enterprise_relationships WHERE enterprise_relationships.child_id=?)
+      OR enterprises.id IN
+        (?)
+      OR enterprises.id = ?
+    ", id, id, enterprise_ids, id)
```
RohanM: This should be extracted to a scope on `Enterprise`.

spec/support/request/shop_workflow.rb
```diff
@@ -29,6 +29,15 @@ def toggle_accordion(name)
...
   def add_variant_to_order_cycle(exchange, variant)
+    oc = exchange.order_cycle
+    supplier = variant.product.supplier
+    # An order cycle needs an incoming exchange for a supplier
+    # before having its products. Otherwise the data will be inconsistent and
+    # and not all needed enterprises are loaded into the shop page.
+    if oc.exchanges.from_enterprise(supplier).incoming.empty?
+      create(:exchange, order_cycle: oc, incoming: true,
+                        sender: supplier, receiver: oc.coordinator)
+    end
```
RohanM: Good thinking. Perhaps extract this new block of code to a private method, something along the lines of `ensure_supplier_exchange`.

RohanM:
On all pages other than the shop, are we still injecting all enterprises?

app/assets/javascripts/darkswarm/directives/enterprise_modal.js.coffee
RohanM: Aah, gotcha.

app/controllers/enterprises_controller.rb
mkllnk: You are right. I actually didn't know how `andand` works. Thanks.

app/assets/javascripts/darkswarm/services/dereferencer.js.coffee
mkllnk: Yes, works well. :-)

app/models/enterprise.rb
mkllnk: Okay. It would be so much easier if ActiveRecord would support an or-chaining of scopes.

mkllnk:
There were nine pages which load all enterprises. I changed only one so far. There are three more shop pages that should be changed in this task.

mkllnk:
I measured the current performance on my laptop.

* `inject_enterprises`: ~5 seconds
* `inject_shop_enterprises`: ~0.08 seconds

app/models/enterprise.rb
```diff
@@ -220,6 +230,11 @@ def relatives
     ", self.id, self.id)
   end
 
+  def relatives_and_oc_producers(order_cycles)
+    enterprise_ids = Exchange.in_order_cycle(order_cycles).incoming.pluck :sender_id
```
RohanM: I find that I have to read and understand the query before I understand what `enterprise_ids` is. Maybe we could overcome that by naming the var more descriptively, possibly something like `producer_ids` or `oc_producer_ids`?

RohanM:
60x speedup :D

That's looking much cleaner. Nice work!

mkllnk:
There are three pages that should load faster now:

- [ ] shopfront
- [ ] edit cart
- [ ] checkout

Apart from that, nothing should change. Only one detail: in the shop front. When you click on producers and then open a producer modal, the shops of that producer may appear a bit delayed. There are loaded when you open the modal.

sstead:
image

On edit cart and on checkout, the 'shop now' options aren't all displaying. I'm not able to scroll down to see all shopfronts. The screenshot below shows the shops visible on map modal, compared to cart modal.
image

image

sstead:
- Category buttons are showing correctly (fixed)
- Producers Tab in shopfront is showing correct producers (fixed)... but their 'shop now' links aren't consistently all showing (see below)... there should be 12 shop now links, but only 3 are showing. This is also true if I open the producer modal from a product. Sometimes waiting fixes the problem, but one time I waited 2 minutes and still only 3 were showing. It's not always a problem, sometimes all of the links load correctly.
image

PR #1082, opened 2016-07-01
Map search by url

RohanM:
- Extract the code to use OSM tiles from `MapSearch` into its own directive `mapOsmTiles`.
- Refactor `MapSearch` to remove cruft.
- Allow searching maps by URL.


bingxie:
I reviewed the code, looks very good to me.

PR #1097, opened 2016-07-19
Delete fee on payment method if payment invalid

mkllnk:
PayPalExpress is always creating two payments. The first one is
invalidated and the second one succeeds. Without deleting the old fee on
the invalidated payment, the order lists the fee twice.


sstead:
Hey Maikel, the Paypal fee isn't charging twice anymore, good stuff! I found a separate payment method fee related bug, but it's on production too, so not related to this fix. See bug here - https://github.com/openfoodfoundation/openfoodnetwork/issues/1098

PR #1109, opened 2016-07-28
Remove Comfortable Mexican Sofa

RohanM:
This branch removes the Comfortable Mexican Sofa gem, which before the "new" front end (over a year ago) we used to customise some content. It hasn't been used for a long time and it was complicating another branch I was working on, so this PR removes it.


sstead:
Testing done. All good.

PR #1113, opened 2016-08-03
Rescue spec fails within enqueue_job matcher

mkllnk:
Failing code tested by the enqueue_job matcher made it fail with:

  expected ConfirmOrderJob to be enqueued matching {} (??? others enqueued)

That was not helpful and masking the real failure. That failure is now
passed on. The hidden intermittent failure happened in 5% of runs on
Travis.

### DIFF
--- a/spec/support/delayed_job_helper.rb
+++ b/spec/support/delayed_job_helper.rb
@@ -34,7 +34,12 @@ module OpenFoodNetwork
       match do |event_proc|
         last_job_id_before = Delayed::Job.last.andand.id || 0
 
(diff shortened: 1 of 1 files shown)


RohanM:
I think I found a problem. If the block raises an exception, but the spec passes, the user never sees the exception. Maybe we could solve that by re-raising the exception at the end of the matcher?

mkllnk:
@RohanM I changed it to raise the exception again. What do you think about it now?

mkllnk:
Merged in #1179.

PR #1116, opened 2016-08-03
Thrive changes to home page

RohanM:
Make some changes to the home page (on Australian instance only) - page alert, learn & connect panes, footer Join us section. Also adds sell page, linked to from footer.


RohanM:
Interfering with buildkite, moving to issue.

Issue #312, opened 2014-12-19
Fractional cents in fee calculations can result in price discrepancy


RohanM:
Say I have an item priced $55.55, and that item has a 10% fee on it. When I add two of these to my cart, the item price should be $61.11 and the total cost should be $122.22. Instead, I see a total cost of $122.21.

What is happening here is that the 10% fee is being calculated and stored for the entire line item of quantity two (55.55 * 0.1 * 2 = 11.11). When we want to calculate the item price inclusive of fees, we divide this by two, giving 55.55 + (11.11/2) = 61.105. We round this up to 61.11.

However, when we calculate the total for the line item, we simply add the adjustment. 55.55*2 + 11.11 = 122.21. Because there is no intermediate rounding, we end up one cent lower.

This is not a terrible UX for the customer - they are offered a price, and at checkout it ends up being a bit cheaper. However, it's not fantastic either - they would not expect any discrepancy.

Fixing this is hard. We'd need to change how adjustments are applied to orders. And by the numbers, nothing's wrong - everyone is being charged the correct amount of money.

I'm leaving this as it stands for now, but we might want to consider changing it in the future.

NickWeir63:
I think this bug is what is causing a discrepancy discovered by a Stroudco shopper who ordered 8 of one item and ended up with a 2p (£0.02) disprepancy.  However the discrepancy was not in her favour (whereas the issue Rohan is describing above would be in her favour)
Anyway here is what we see.  The shopper orders 8 x 1 pint goats milk and gets this basket total (note 2p difference between line total and order total)
![screenshot 2016-05-11 16 55 45](https://cloud.githubusercontent.com/assets/10517702/15187549/64570114-1799-11e6-979b-1f64248b35a2.png)
We can see that 

- The producer's master price is 86p
- We add 20% which comes to £1.032
- When multiplied up by 8 pints this comes to £8.256

 
@OliverUK suggested would it be possible to disregard the third decimal place and multiply up £1.03 instead of £1.032......?

Also - why are the line total and the basket/order total different?

Thanks

OliverUK:
@RohanM @NickWeir63 

Looks like the line total is calculated using £1.03 x 8 = £8.24. Does it drop the 2 or round down?
The order total uses £1.032 and then rounds up. 

We'll continue getting strange results until the sales price has any invisible decimals (like here the 2) removed before further calculations, IMHO.

lin-d-hop:
[...]

PR #1127, opened 2016-08-10
Cart fee rounding consistency

RohanM:
In response to #312:

When we have a percentage fee applied to a line item, we see some calculation inconsistencies. For example, say we're applying a 20% fee to a $0.86 item with quantity 8:

- JS cart: `(price * 1.2).round(2) * quantity = 8.24`
- Cart page (line item): same as above
- Order subtotal: `(price * quantity * 1.2).round(2) = 8.26`

This pull request changes how percentage fees are calculated so that the percentage is applied at the item level, rounded and them summed. This removes the inconsistency, so that the customer sees the same price the whole way through.


mkllnk:
That change looks good to me. I was just wondering about the bigger picture. This doesn't look like a change we want to feed back to Spree. There are probably use cases for both, Spree's version and our version. So would it be better to create a new calculator `FlatPercentPerItem` instead of overriding `FlatPercentItemTotal`? The name doesn't suit any more.

RohanM:
That sounds like it would be an improvement. We could make a migration that swapped all the assigned calculators out, and we could disable `FlatPercentItemTotal`, since leaving it in would make for confusion (both in choosing a calculator and the results it produces). I'll look at this next week.

daniellemoorhead:
@RohanM and @mkllnk so what's the plan - do we push this live or look at implementing a bigger picture view?

spec/support/request/shop_workflow.rb
```diff
@@ -16,9 +16,9 @@ def set_order(order)
     ApplicationController.any_instance.stub(:session).and_return({order_id: order.id, access_token: order.token})
   end
 
-  def add_product_to_cart
```
mkllnk: How did this actually work before? How did this method have access to `order` and `product`? Is it because `let` defines a method instead of a variable?

mkllnk:
This looks very good to me now. Easy to read and understand.

spec/support/request/shop_workflow.rb
RohanM: I think that's correct - so let vars are acting kinda like global variables - not exactly transparent or flexible.

Issue #1062, opened 2016-06-22
SO - Store and edit a customer's preferred address


daniellemoorhead:
**As a shop owner I want to store and edit a customer's preferred address**

### Issue 1: Frontend

The logic for auto-loading an address for an order is in CheckoutController#before_address.

Currently this logic is checking the current_user for for a bill_address and ship_address, however since we don't actually have an interface for assigning these, they are always nil. 

What happens instead is that the logic uses the current_order's email and OpenFoodNetwork::LastUsedAddress to find the address on the most recent order associated with that particular email. This is really supposed to be a fallback option, but instead is our only approach to loading address for existing customers.

It would be good to make this configurable by the customer both within the checkout (checkbox: "Save this as my default address") and in the account section.

### Issue 2: Back end

a) Some Enterprise users (Food Connect comes to mind) have expressed a desire to have more control over customer information from the backend, this includes updating account details such as name, contact number and delivery address details. 

b) we need some way of pulling up an appropriate address for customers as part of Standing Orders, and this will allow shops to set up an address for customers where one is known without the customer necessarily having to place an order first.

##### New fields on Customer

- [x] Add new optional fields to customer: bill_address_id and ship_address_id

- [x] Add a new field to customer: name (this is designed to be used in the backend)

##### Admin Customers Index
- [x] Add a new column for name, which is editable in situ

- [x] Add a new column for each address (bill, ship)
- [x] Column displays the start of the address, limited to ~15 characters
- [x] When the address is clicked, it can be edited in a modal (6h)

bingxie:
- [ ] add bill_address_id and ship_address_id index and foreign key

bingxie:
I moved those uncheck taskes to two separated issues #1130 and #1131 
I wanted to make it more easier to review and test.

sstead:
![image](https://cloud.githubusercontent.com/assets/10843056/17916182/4bd3613e-69f4-11e6-8553-ab98b2545e28.png)

sstead:
Regarding create an order... this is working well! yay!... a few nice to haves...
- It would be great if you could search for customers by their billing name, currently you can only search by email. People often have emails like bigapple65@gmail.com, so searching by email isn't always straightforward for the hub.
- Regarding the loading of customer details... Can we make it such that by default, the 'use billing address' box is selected, UNLESS the customer has different default shipping details to their billing details. If they have no default shipping address saved, we can also assume they want to 'use billing address'.
[...]

PR #1132, opened 2016-08-12
Enterprise user can edit the customer's name, billing address and shipping address

bingxie:
The store is here: https://github.com/openfoodfoundation/openfoodnetwork/issues/1062

### DIFF
--- a/app/assets/javascripts/admin/customers/directives/edit_address_dialog.js.coffee
+++ b/app/assets/javascripts/admin/customers/directives/edit_address_dialog.js.coffee
@@ -1,13 +1,21 @@
-angular.module("admin.customers").directive 'editAddressDialog', ($compile, $templateCache, DialogDefaults, CurrentShop, Customers) ->
+angular.module("admin.customers").directive 'editAddressDialog', ($compile, $templateCache, $filter, DialogDefaults, Customers) ->
   restrict: 'A'
   scope: true
   link: (scope, element, attr) ->
-    scope.$watch 'country', (newVal) ->
+    scope.$watch 'address.country_id', (newVal) ->
       if newVal
-        scope.states = newVal.states
+        scope.states = scope.filter_states(newVal)
+
+    scope.updateAddress = ->
+      console.log(scope.edit_address_form)
 
     template = $compile($templateCache.get('admin/edit_address_dialog.html'))(scope)
     template.dialog(DialogDefaults)
 
     element.bind 'click', (e) ->
+      scope.address = scope.customer.bill_address
       template.dialog('open')
+      scope.$apply()
+
+    scope.filter_states = (countryID) ->
+      $filter('filter')(scope.availableCountries, {id: countryID})[0].states
--- a/app/assets/javascripts/admin/customers/directives/edit_address_dialog.js.coffee
+++ b/app/assets/javascripts/admin/customers/directives/edit_address_dialog.js.coffee
@@ -7,13 +7,22 @@ angular.module("admin.customers").directive 'editAddressDialog', ($compile, $tem
         scope.states = scope.filter_states(newVal)
 
     scope.updateAddress = ->
-      console.log(scope.edit_address_form)
+      scope.edit_address_form.$setPristine()
+
+      Customers.update(scope.address, scope.customer, scope.current_address).$promise.then (data) ->
+        scope.customer = data
+        template.dialog('close')
 
     template = $compile($templateCache.get('admin/edit_address_dialog.html'))(scope)
     template.dialog(DialogDefaults)
 
     element.bind 'click', (e) ->
-      scope.address = scope.customer.bill_address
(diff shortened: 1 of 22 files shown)


app/assets/javascripts/admin/customers/directives/edit_address_dialog.js.coffee
```diff
@@ -0,0 +1,36 @@
...
+      if newVal
+        scope.states = scope.filter_states(newVal)
+
+    scope.updateAddress = ->
+      scope.edit_address_form.$setPristine()
+      if scope.edit_address_form.$valid
+        Customers.update(scope.address, scope.customer, scope.current_address).$promise.then (data) ->
+          scope.customer = data
+          template.dialog('close')
+          StatusMessage.display('success', "Address updated successfully.")
```
mkllnk: We should put all language strings visible to a user into `config/locales/en.yml`.

app/assets/javascripts/admin/customers/services/customers.js.coffee
```diff
@@ -25,3 +25,11 @@ angular.module("admin.customers").factory "Customers", ($q, InfoDialog, RequestM
       request = CustomerResource.index(params, (data) => @customers = data)
       RequestMonitor.load(request.$promise)
       request.$promise
+
+    update: (address, customer, current_address) ->
```
mkllnk: Which values does `current_address` have? I guess `bill_address` or `ship_address`. In that case we could name the variable `address_type` or similar. Because when reading `current_address`, I expect it to contain a value like "123 Hight St, Thornbury Victoria".

app/assets/javascripts/templates/admin/edit_address_dialog.html.haml
```diff
@@ -0,0 +1,58 @@
+#edit-address-dialog
+  %h2 {{ current_address === 'bill_address' ?  'Edit Billing Address' : 'Edit Shipping Address'}}
```
mkllnk: Again, we should make use of locales for language parts.

app/models/customer.rb
```diff
@@ -6,8 +6,17 @@ class Customer < ActiveRecord::Base
   has_many :orders, class_name: Spree::Order
   before_destroy :check_for_orders
 
+  belongs_to :bill_address, foreign_key: :bill_address_id, class_name: Spree::Address
+  alias_attribute :billing_address, :bill_address
```
mkllnk: I'm wondering if it will be confusing to have to attribute names pointing to the same attribute. Did you discuss that with Rohan or Rob?

bingxie: It's like a kind of convention from Spree, I learned it from Spree::Order and Spree::User, it's not used much.

app/assets/javascripts/admin/customers/directives/edit_address_dialog.js.coffee
bingxie: yes, will do.

app/assets/javascripts/templates/admin/edit_address_dialog.html.haml
bingxie: will do.

db/migrate/20160713003535_add_bill_address_and_ship_address_to_customer.rb
```diff
@@ -0,0 +1,12 @@
+class AddBillAddressAndShipAddressToCustomer < ActiveRecord::Migration
+  def change
+    add_column :customers, :bill_address_id, :integer
+    add_column :customers, :ship_address_id, :integer
+
+    add_index :customers, :bill_address_id
```
mkllnk: Why do we need an index here?

app/models/customer.rb
mkllnk: Okay. Then it's good to stick to the convention.

db/migrate/20160713003535_add_bill_address_and_ship_address_to_customer.rb
bingxie: I thought it's a good practice to add index on foreign key for better performance. What do you think?

mkllnk: Just thinking through. It's a one-to-one relationship. You normally don't lookup a customer by it's address. That's why I thought we don't need it. But it is possible that we add a feature later to get all customers of a region. In that case the index would be good. I'm not sure what's better. Probably just leave it as is.

bingxie: I think you are right.

app/assets/javascripts/admin/customers/services/customers.js.coffee
bingxie: good!

mkllnk:
Looks all good to me.

Issue #1130, opened 2016-08-12
SO - Customer can set default addresses during checkout


bingxie:
Add a checkbox to the billing and shipping address sections of the checkout page, allowing the customer to set the given address as their default (2h)

When auto-creating a new Customer when an order is placed, auto-assign "name" field using the full name of the billing address given. (1h)

The enterprise user can search/filter the customers(which belongs to current enterprises) and it will auto-complete the address form if the customer has the addresses.

Add a confirm when deleting a customer.

daniellemoorhead:
Hey @bingxie can you let me know if you think you'll get this to Sally for testing this week? We're going to release milestone 1.8.5 on Friday so if it won't be ready by then I'll put it into [v1.8.6](https://github.com/openfoodfoundation/openfoodnetwork/milestone/39).

Cheers!

bingxie:
I will create a PR for review today, then need someone to review it on Friday, I won't work on this Friday, so I think maybe we can put it into v1.8.6

daniellemoorhead:
@sstead can you confirm that this has been added to the user guide? It went through in 1.8.6 but there weren't release notes from you as far as I can see and so I wanted to confirm before I closed this.

sstead:
Hey @daniellemoorhead , yep it was in the [release notes](http://community.openfoodnetwork.org/t/what-s-new-v-1-8-6-162810/720) and covered in the [user guide](https://openfoodnetwork.org/user-guide/advanced-features/customer-accounts-and-tagging/).

PR #1142, opened 2016-09-01
Checkout default address and Customer search filter

bingxie:
The story description is here: #1130 @oeoeaio  or @RohanM if you are free, could you please have a review? Thanks.

@daniellemoorhead


app/assets/javascripts/admin/checkouts/edit_override.js
```diff
@@ -0,0 +1,53 @@
+$(document).ready(function() {
```
oeoeaio: Nice work @bingxie!

Could we turn this into a directive? You can use the `restrict: 'C'` option to bind a directive to an element with a particular class (ie. #customer_search_override could bind to a directive named customerSearchOverride). Then the rest of this function can just go inside the link function on the directive, using `element` instead of `$("#customer_search_override")` to make the `select2` call.

What do you think?

app/assets/javascripts/admin/customers/controllers/customers_controller.js.coffee
```diff
@@ -21,6 +20,10 @@ angular.module("admin.customers").controller "customersCtrl", ($scope, $q, $filt
 
   $scope.shop_id = shops[0].id if shops.length == 1
 
+  $scope.deleteCustomer = (customer) ->
```
oeoeaio: This is great. Good idea.

app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
```diff
@@ -9,9 +9,18 @@ Darkswarm.controller "CheckoutCtrl", ($scope, storage, Checkout, CurrentUser, Cu
   for field in $scope.fieldsToBind
     storage.bind $scope, "Checkout.order.#{field}",
       storeName: "#{prefix}_#{field}"
+
   storage.bind $scope, "Checkout.ship_address_same_as_billing",
     storeName: "#{prefix}_sameasbilling"
-    defaultValue: true
+    defaultValue: 'YES'
```
oeoeaio: What is the reason for switching from true/false to 'YES'/'NO'? I think it's fine, but there must have been a reason for doing it?

bingxie: Yes, there is. I have tried a lot. I found there was a bug to store the false value in the local storage. The false value sometimes can not be retrieved correctly, and can be affected by other form fields. I would like to show you next week with master branch.

app/assets/javascripts/admin/checkouts/edit_override.js
bingxie: That could be a good idea. I will have a try.

app/assets/javascripts/darkswarm/services/current_order.js.coffee
```diff
@@ -1,4 +1,16 @@
 Darkswarm.factory 'CurrentOrder', (currentOrder) ->
   # Populate Currentorder.order from json in page. This is probably redundant now.
   new class CurrentOrder
-    order: currentOrder
+    # order: currentOrder
+
+    getOrder: ->
```
oeoeaio: Hi @bingxie. I don't really like the look of this. What is the problem that this function is solving? Why do we need state_id and country_id to be strings?

app/controllers/spree/admin/search_controller_decorator.rb
```diff
@@ -16,10 +16,17 @@ def known_users
     render :users
   end
 
+  def customers
```
oeoeaio: We will need to scope the list of customers searched to those of enterprises owned by the current user, to prevent a user from find a customer of another enterprise by submitting a different distributor_id in the params. We could just just do a test to check that `spree_current_user.enterprises.pluck(:id).include? params[:distributor_id]` and render an empty array if it doesn't?

app/models/spree/order_decorator.rb
```diff
@@ -289,7 +289,8 @@ def associate_customer
 
   def ensure_customer
     unless associate_customer
-      self.customer = Customer.create(enterprise: distributor, email: email_for_customer, user: user)
+      customer_name = bill_address.andand.full_name
+      self.customer = Customer.create(enterprise: distributor, email: email_for_customer, user: user, name: customer_name, bill_address: bill_address.andand.clone, ship_address: ship_address.andand.clone)
```
oeoeaio: I think that it would be better to set `bill_address` and `ship_address`, and therefore `name` in the `set_default_bill_address` and `set_default_ship_address` methods on `app/controllers/checkout_controller.rb`. Otherwise, the addresses could be set to nil here, because they may not exist yet when ensure_customer is called. If we do it in the controller instead, we always know that we have address to assign to the customer.  

What do you think?

app/serializers/api/admin/customer_serializer.rb
```diff
@@ -8,9 +8,13 @@ def tag_list
     object.tag_list.join(",")
   end
 
+  def name
```
oeoeaio: What about: `object.name || object.bill_address.andand.full_name`?

spec/controllers/spree/admin/search_controller_spec.rb
```diff
@@ -31,5 +31,29 @@
         end
       end
     end
+
```
oeoeaio: Need to add a spec that tests that a user cannot submit an unmanaged distributor_id and return customers of that enterprise.

app/controllers/checkout_controller.rb
```diff
@@ -58,6 +61,23 @@ def update
...
+      @order.customer.update_attribute(:bill_address, new_bill_address) unless @order.customer.bill_address
+    end
+
+  end
+
+  def set_default_ship_address
+    if params[:order][:default_ship_address] == 'YES'
+      new_ship_address = @order.ship_address.clone
+      spree_current_user.set_ship_address(new_ship_address)
+      @order.customer.update_attribute(:ship_address, new_ship_address) unless @order.customer.ship_address
```
oeoeaio: Logic here should be to update customer's bill address if it is already set to something, but otherwise (ie. if it is blank) leave it blank. So we need to change `unless` here to `if`.

app/controllers/checkout_controller.rb
```diff
@@ -58,6 +61,23 @@ def update
 
   private
 
+  def set_default_bill_address
+    if params[:order][:default_bill_address] == 'YES'
+      new_bill_address = @order.bill_address.clone
+      spree_current_user.set_bill_address(new_bill_address)
+      @order.customer.update_attribute(:bill_address, new_bill_address) unless @order.customer.bill_address
```
oeoeaio: Logic here should be to update customer's bill address if it is already set to something, but otherwise (ie. if it is blank) leave it blank. So we need to change unless here to if.

app/controllers/checkout_controller.rb
```diff
@@ -136,7 +156,8 @@ def before_address
     last_used_bill_address = lua.last_used_bill_address.andand.clone
     last_used_ship_address = lua.last_used_ship_address.andand.clone
 
-    preferred_bill_address, preferred_ship_address = spree_current_user.bill_address, spree_current_user.ship_address if spree_current_user.respond_to?(:bill_address) && spree_current_user.respond_to?(:ship_address)
+    preferred_bill_address, preferred_ship_address = spree_current_user.bill_address, spree_current_user.ship_address if spree_current_user
+
```
oeoeaio: We need to add the customers addresses to the front of this chain of addresses. Something like:

    @order.bill_address ||= customer_preferred_bill_address || user_preferred_bill_address || last_used_bill_address || Spree::Address.default
    @order.ship_address ||= customer_preferred_ship_address || user_preferred_ship_address || last_used_ship_address || Spree::Address.default

app/models/spree/user_decorator.rb
```diff
@@ -76,6 +76,22 @@ def orders_by_distributor
...
+
+  def set_ship_address(address)
+    if self.ship_address
+      self.ship_address.update_attributes(address.clone.attributes)
+    else
+      self.update_attribute(:ship_address, address.clone)
+    end
+  end
+
   private
```
oeoeaio: Just interested in your thinking behind these methods, why can't we just call this from the checkout controller?:

  spree_current_user.update_attribute(:ship_address, new_ship_address)

Like you have done for the customer's address?

app/assets/javascripts/darkswarm/services/current_order.js.coffee
bingxie: convert the type in the Serializer all the test passed

app/controllers/spree/admin/search_controller_decorator.rb
bingxie: That's a good idea.

app/models/spree/order_decorator.rb
bingxie: This piece of code is also useful for the admin interface creating a new order for a customer. So I think maybe it is still good to leave it here.

app/serializers/api/admin/customer_serializer.rb
bingxie: yes, good spot.

spec/controllers/spree/admin/search_controller_spec.rb
bingxie: done.

app/controllers/checkout_controller.rb
bingxie: As we discussed before, if the customer was set an address from the admin interface, we won't change the address here. Because here is to set the default one, only when the customer doesn't have one.

app/controllers/checkout_controller.rb
bingxie: got it. done.

app/models/spree/user_decorator.rb
bingxie: The problem is if ship_address is not nil, the update_attribute method will create a new address  record and then we need to delete the old unused address. So that's why I used update_attributes method to only update the attributes of the existing address.

app/controllers/checkout_controller.rb
oeoeaio: Hm, I don't think that is quite right. If the customer's address was set from admin, we still want the customer to be able to change it if they want to. 

If no customer ship address is currently set and default_ship_address is 'YES' here then we can just update the user's address, because next time the logic in `app/controllers/checkout_controller.rb` will load the user's ship address.
However, if the customer ship address is set (address1), then we need to update address1 to whatever the order address is (address2) here when `default_ship_address == 'YES'`. Otherwise, the next time they go through the checkout, the logic will load address1 and not address2 as they would expect.

What do you think?

bingxie: yes, I understand now.

sstead:
This is a display bug...
image

sstead:
I'm finding the name field in the Customer list confusing....
image

sstead:
There is an inconsistency with saving the hub's address as the customer's default shipping address for customers who select a 'pick up' shipping option.... We need to ask if this is what we want in the field? and how is this helpful? Appologies for the long winded explanation below:
image
- I checked out as guest with this new email, for a person who has never been on the OFN before.
- When checking out as a guest there are no 'save this as my default' checkboxes
- At checkout this person selected a 'pickup' shipping method, so no shipping address was entered
- You can see above which details were saved in the enterprises Customer list. A default billing address has been saved. And the hub's address has been saved as the default shipping address.
- When this person checked out as a guest a second time, with different details (name, shipping/billing details), none of the originally saved details were overriden.
- When this customer signed up, checked out as a logged in user and then clicked the delivery shipping method, their shipping address auto-filled with the hub's address (below).
image

My suggestions:
- It's not always correct to say that when a hub has a 'pickup' shipping method that the address is the hubs address. Pick up may be from multiple pickup points such as schools and markets.
- As seen above, for a new customers shopping as a guest, if they select a pick up shipping method, the hub's address is saved in their shipping address field. However, for a logged in customer, when they are checking out, they can't select 'save as my default shipping address' when they choose a pick up shipping method. So it's not clear whether we want 'pick up' shipping addresses saved as the 'shipping address' or not.  
- I would suggest that when a customer select's 'pick up' as their shipping method, that no address is saved in their 'shipping address'. The whole idea of these default addresses is to save time at checkout. If you're a customer who uses a 'pick up' shipping method, it doesn't take much time to click on the pickup point you want.  I don't think we need defaults saved for this. Plus, saving the hub's address as the shipping method doesn't mean your pickup shipping method is automatically selected in checkout. If this is what we want, the system could remember which pick up shipping method this customer prefers, and save this in the shipping address field... something like 'prefers pick up'. This would require a checkbox at checkout saying 'save this as my default shipping method'.

sstead:
I don't think this needs to be fixed now, but could be an nice to have improvement...
image

sstead:
This is working-
image

sstead:
This is a suggested improvement which might stop customers getting confused about which address is being saved... Do others think this would be less confusing?
image

sstead:
Regarding create an order... one bug and some nice to haves...
- 
image

- It would be great if you could search for customers by their billing name, currently you can only search by email. People often have emails like bigapple65@gmail.com, so searching by email isn't always straightforward for the hub.
- Regarding the loading of customer details... Can we make it such that by default, the 'use billing address' box is selected, UNLESS the customer has different default shipping details to their billing details. If they have no default shipping address saved, we can also assume they want to 'use billing address'.
image

PR #1169, opened 2016-09-25
Update sell page mail_to I18n (and UK translations)

lin-d-hop:
Small update to internationalise the mail_to on the sell page to use the instance email rather than a hardcoded aus email.

Translation updates for the sell page for en-GB.yml also in this PR.

@daniellemoorhead would you ask someone to take a look and merge if happy?


daniellemoorhead:
Hiya @lin-d-hop I've assigned this to @mkllnk ...but it's saying that all checks have failed, so does something need to be done to get it to pass before he looks at it?

lin-d-hop:
@daniellemoorhead Tests are passing :-)
@mkllnk Thanks for taking a look!

app/views/home/sell.html.haml
```diff
@@ -44,7 +44,6 @@
 
         %p
           = t :sell_embed
-          %a{href: "hello@openfoodnetwork.org".reverse, target: '_blank', mailto: true}
-            = t :sell_ask_services
+          = mail_to ContentConfig.footer_email, t(:sell_ask_services)
```
mkllnk: Is the `reverse` missing here? We use it to obfuscate the email address. The `mailto` attribute triggers some javascript to un-reverse it again for the user.

lin-d-hop:
Hi @mkllnk 
Made the change to the reverse that you suggested. Seems that using mail_to here the .reverse is not re-reversed (versed?). Testing locally we just end up with a backwards email address and delivery failures. Is it ok to leave the .reverse out?

I think the merge conflict is coming from the fact the config/locales/en_GB.yml is deleted in this PR. This was added by me incorrectly after downloading from transfiex and not changing to en-GB like the existing file. I'm not sure how to resolve this without merging the deleted file on master... how is best for me to resolve this?

lin-d-hop:
Replaced by #1176

lin-d-hop:
@mkllnk Merge conflicts now fixed.

mkllnk:
@lin-d-hop Thank you for solving the conflicts. I added the `mailto` attribute to restore the email address when clicked. We could probably put that into a helper like `mask_mail_to`.

sstead:
Testing notes (for my reference)... 
Initial feedback...
image
This was corrected, now emails of both formats work, in all browsers, written forwards :)

PR #1170, opened 2016-09-25
Terms of Service URL in emails

lin-d-hop:
Updating email footer to read the Terms of Service url from spree-preferences. Field has a default of Terms-of-service.pdf so current functionality unchanged in the case of the field being blank.

Tested on UK staging.

### DIFF
--- a/app/views/layouts/mailer.html.haml
+++ b/app/views/layouts/mailer.html.haml
@@ -42,7 +42,7 @@
               %tr
                 %td{:align => "center"}
                   %p
-                    %a{:href => "#{ URI.join(spree.root_url, "Terms-of-service.pdf").to_s }", :target => "_blank"}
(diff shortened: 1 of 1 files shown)


sstead:
Testing on Staging:
- Email terms of service link goes to the link set in Config>content>Terms of service URL
- If instance has something set here, that URL remains
- If Config>content>Terms of service URL is set to blank, the link in email goes to the homepage
- I suppose for instances that have no Terms of service URL in Config, the link will also go tothe homepage

mkllnk:
@lin-d-hop I merged this commit. But what do you think about the case when the url is set to blank? It seems that ContentConfig is not using the default value then. Should we hide the link in that case? I don't think that the Australian ToS apply for any other instance.

lin-d-hop:
I agree that the Aus ToS is not useful to anyone else. Hiding the link sounds like a good idea. I'll update the issue, but its now a lower priority to us :-/

Issue #1155, opened 2016-09-15
Delivery report - Customer Delivery address incorrect


OliverUK:
@MagsHall @lin-d-hop 

When a customer puts in a separate delivery address for his current order, this doesn't seem to come through in the report. I can't screenshot it without compromising personal data but I'll email you the information shortly. Have opened this issue to keep track and for easy commenting. 

Thanks 
Oliver

lin-d-hop:
@rhodri While you are making changes to the delivery report would you take a look at this as well?

lin-d-hop:
PR #1177

lin-d-hop:
Merged. Thanks crew!

PR #1177, opened 2016-10-04
Update delivery report to use shipping address

lin-d-hop:
From issue #1155 

Quick fix to bug, delivery report now uses delivery address instead of billing address.

@mkllnk @daniellemoorhead 
Any chance of v1.8.6?


oeoeaio:
@mkllnk, no I don't think billing address is required here for now

sstead:
@mkllnk testing all looks good! 
image

Issue #1160, opened 2016-09-16
SO - Opt-in option for enterprises to require customers to login to complete an order


mkllnk:
Enterprises can choose to force their customers to log in (ie. not use guest checkout) to place orders with them. This feature is needed for standing orders, for the use case where a user that has a standing order comes into their shop and goes through the checkout process, logging in after they've added items to their cart. If they have a standing order we need to be able to communicate this to them and give them the option of seeing these standing items in their cart/order (this communication is being built in #1084).

Ideally, #1159 would be done first. But since this is optional, a shop can choose to do this or not and it does not need to be foolproof.

- [x] Create the setting in the enterprise to 'allow guest orders'
- [x] Hide option to checkout as guest if it's not active

sstead:
Could we better stress to Users that if they're using SO that they should also require login? Perhaps when they first click to 'Enable Standing Orders' (AND you have guest checkout allowed) a popup or exclamation should appear saying 'we suggest you also require login before your customers can checkout'.
![image](https://cloud.githubusercontent.com/assets/10843056/21377270/57c1e1e6-c791-11e6-9656-32456cca5b66.png)

Issue #901, opened 2016-04-13
Move 'shopfront requires login' setting to 'shop preferences' tab in Profile Edit


sstead:
Currently this setting is in the Primary Details tab. It's visible to profile only users, which is unnecessary. Makes more sense in shop preferences.

SHOPFRONT REQUIRES LOGIN?
What's this?
 PUBLIC  REQUIRE CUSTOMERS TO LOGIN

PR #1179, opened 2016-10-05
Allow guest orders

mkllnk:
Addressing #1160 

Also moves "shopfront requires login" setting (#901) and includes testing feedback improvements ( #1113)

Please check:

* Enterprise users can disallow guest orders.
* Guests have to login before checkout for those shops.

### DIFF
--- a/app/helpers/checkout_helper.rb
+++ b/app/helpers/checkout_helper.rb
@@ -1,4 +1,8 @@
 module CheckoutHelper
+  def guest_checkout_allowed?
+    current_order.distributor.allow_guest_orders?
+  end
+
   def checkout_adjustments_for(order, opts={})
     adjustments = order.adjustments.eligible
     exclude = opts[:exclude] || {}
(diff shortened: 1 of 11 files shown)


sstead:
Testing notes so far, to be continued on Friday

image

image

image

Questions to address:
- Are the 2 settings obvious? At first glance the two settings seem to do the same thing, how can we make the distinction between guest checkout and private shopfront clearer?

sstead:
github

sstead:
image

sstead:
This is the default setting for a new producer shop or hub shop (CORRECT). I'd suggest having the default option listed first.
image

sstead:
This isn't a priority, but just something small to think about, I can log it as an improvement for the backlog?
image

mkllnk:
Okay, I will look at these two issues:

- re-phrase the options
- look into deactivating the second option depending on the first one

Issue #1180, opened 2016-10-07
Contact tab not opening when shopfront requires login


sstead:
Clicking the contact hyperlink should open the shop's contact details tab. I think the URL of the link should be https://staging2.openfood.com.au/producer-of-honey/shop#/contact instead of https://staging2.openfood.com.au/producer-of-honey/shop##contact

![image](https://cloud.githubusercontent.com/assets/10843056/19175198/72b15258-8c80-11e6-9a8c-01fcff501aaa.png)

lin-d-hop:
This has just been picked up by a UK user too.

lin-d-hop:
Looking at this as it has been raised by a new user. PR #1188

lin-d-hop:
#1210 is updated PR for this issue.

PR #1188, opened 2016-10-14
Broken Contact link in Login Shopfront

lin-d-hop:
Quick fix to mend the broken Contact link specified in #1180 

Feels like a bit of a hack, but it works for me. If I should really be doing this in a different way would you signpost please?

@daniellemoorhead 
@mkllnk  Would you take a look at this fix please? The bug is a problem for new users who are setting up login only shopfronts, it is resulting in new customers walking away before they start :-/

### DIFF
--- a/app/views/shop/_messages.html.haml
+++ b/app/views/shop/_messages.html.haml
@@ -10,7 +10,7 @@
              register: ('<a auth="signup">' + t('.register') + '</a>').html_safe}
         - else
           = t '.require_customer_html',
-            {contact: ('<a href="##contact">' + t('.contact') + '</a>').html_safe,
+            {contact: (link_to t(:contact), "#/contact", method: :get),
              enterprise: current_distributor.name}
 - elsif @order_cycles and @order_cycles.empty?
   - if current_distributor.preferred_shopfront_closed_message.present?
--- a/app/views/shop/_messages.html.haml
+++ b/app/views/shop/_messages.html.haml
@@ -10,7 +10,7 @@
              register: ('<a auth="signup">' + t('.register') + '</a>').html_safe}
         - else
           = t '.require_customer_html',
-            {contact: (link_to t(:contact), "#/contact", method: :get),
+            {contact: (link_to t('.contact'), "#/contact", method: :get),
              enterprise: current_distributor.name}
 - elsif @order_cycles and @order_cycles.empty?
   - if current_distributor.preferred_shopfront_closed_message.present?


daniellemoorhead:
@lin-d-hop a question on priority in terms of @mkllnk's time - he has the multilingual quote and PR #1177 in his queue right now, which is all he can fit this week into his 2 days. Does this new PR trump those 2 items and need to be done this week, or can it wait till next week?

lin-d-hop:
Tests failing :-( next week is fine

lin-d-hop:
@daniellemoorhead would you schedule this for a code review please. Tis a tiny change.

daniellemoorhead:
Hey @lin-d-hop any chance that you could have @stveep or @levent do a code review on it first off, then we can push it through the pipe? Kinda like what we were speaking about during the UK dev catchup, where we share the review load and then it becomes quicker for us to merge them.

levent:
This works but it doesn't toggle the contact tab using js in the same way clicking Contact in the menu bar does.
I'm not sure if there's a clean easy way to do that off the top of my head but might make the UX a little smoother? /cc @stveep

app/views/shop/_messages.html.haml
```diff
@@ -10,7 +10,7 @@
              register: ('<a auth="signup">' + t('.register') + '</a>').html_safe}
         - else
           = t '.require_customer_html',
-            {contact: ('<a href="##contact">' + t('.contact') + '</a>').html_safe,
+            {contact: (link_to t('.contact'), "#/contact", method: :get),
```
levent: ```method: :get``` is superfluous.
Every ```html_safe``` removed is a good thing 👍

lin-d-hop: Thanks @levent. The method: :get was added as a hack to force a page refresh so that the Contact details display. Otherwise the link does nothing, creating big problems from enterprise users.
Would love a better suggestion, less hacky, if you have one?

stveep: Maybe we can use Jquery to fire off the click event for the tab, something like this: 
`$('#tab_contact').find('a').click()`
That works for me in the console at least. Would avoid a page refresh (which is quite slow so I think worth avoiding). The problem I can see is that it would do nothing if it can't find the tab - I think it's ok as the _messages partial is always displayed with the shop details, but it's future-proof (e.g. if this message partial is reused on another page).

I'm not sure it's less hacky than Lynn's method but it is snappier :)

levent: I'm not familiar enough with Angular to know if there's a better way but happy to go with this as is or with @stveep's suggestion as it resolves the main issue which is there being a broken link.

mkllnk: I think the right solution would be a new Angular directive. In html you write something like:

```html
<a href="#contact" navigation-link>...</a>
```

And the directive is activated by the attribute `navigation-link`. It uses the `Navigation` service in `app/assets/javascripts/darkswarm/services/navigation.js.coffee` to trigger opening the contact tab.

lin-d-hop:
Closing now this has been superseded by #1210

Issue #1131, opened 2016-08-12
SO - user can update preferred billing address and shipping address on account page


bingxie:
Add a section to the front-end "Account" section, to allow a customer the manually edit their own billing address or shipping address. (8h)

daniellemoorhead:
Hiya @bingxie can you give me an update on where this issue is at? Almost ready to test and deploy?

sstead:
I'm taking this out of subs, it's related but not specific to subs.

RachL:
@Erioldoesdesign could this be interesting for design contributors? Otherwise I think we should close it. Not enough specified.

PR #1192, opened 2016-10-21
Edit user addresses

bingxie:
User can edit addresses on user account page. #1131

### DIFF
--- a/app/controllers/spree/users_controller_decorator.rb
+++ b/app/controllers/spree/users_controller_decorator.rb
@@ -1,7 +1,7 @@
 Spree::UsersController.class_eval do
   layout 'darkswarm'
 
-  prepend_before_filter :load_object, :only => [:edit_address]
+  prepend_before_filter :load_object, :only => [:edit_address, :show, :edit, :update]
 
   def edit_address
     render :edit
--- a/app/controllers/spree/users_controller_decorator.rb
+++ b/app/controllers/spree/users_controller_decorator.rb
@@ -1,9 +1,21 @@
 Spree::UsersController.class_eval do
   layout 'darkswarm'
 
-  prepend_before_filter :load_object, :only => [:edit_address, :show, :edit, :update]
+  prepend_before_filter :load_object, :only => [:edit_address, :update_address, :show, :edit, :update]
 
   def edit_address
+    unless @user.ship_address
+      @user.ship_address = Spree::Address.new
+      @user.ship_address.country = Spree::Country.find_by_id(Spree::Config[:default_country_id])
+    end
     render :edit_address
+  end
+
+  def update_address
+    if @user.update_attributes(params[:user])
+      redirect_to spree.account_url, :notice => t(:account_updated)
+    else
+      render :edit_address
+    end
   end
 end
--- a/app/controllers/spree/users_controller_decorator.rb
+++ b/app/controllers/spree/users_controller_decorator.rb
@@ -4,10 +4,10 @@ Spree::UsersController.class_eval do
   prepend_before_filter :load_object, :only => [:edit_address, :update_address, :show, :edit, :update]
 
   def edit_address
-    unless @user.ship_address
-      @user.ship_address = Spree::Address.new
-      @user.ship_address.country = Spree::Country.find_by_id(Spree::Config[:default_country_id])
-    end
+    country_id = Spree::Config[:default_country_id]
+    @user.build_ship_address(country_id: country_id) unless @user.ship_address
+    @user.build_bill_address(country_id: country_id) unless @user.bill_address
+
     render :edit_address
   end
 
--- a/app/controllers/spree/users_controller_decorator.rb
+++ b/app/controllers/spree/users_controller_decorator.rb
@@ -1,7 +1,7 @@
 Spree::UsersController.class_eval do
   layout 'darkswarm'
 
-  prepend_before_filter :load_object, :only => [:edit_address, :update_address, :show, :edit, :update]
+  prepend_before_filter :load_object, only: [:edit_address, :update_address, :show, :edit, :update]
 
   def edit_address
     country_id = Spree::Config[:default_country_id]
@@ -13,7 +13,7 @@ Spree::UsersController.class_eval do
 
   def update_address
     if @user.update_attributes(params[:user])
-      redirect_to spree.account_url, :notice => t(:account_updated)
+      redirect_to spree.account_url
     else
       render :edit_address
     end
--- a/app/controllers/spree/users_controller_decorator.rb
+++ b/app/controllers/spree/users_controller_decorator.rb
@@ -4,6 +4,6 @@ Spree::UsersController.class_eval do
   prepend_before_filter :load_object, :only => [:edit_address, :show, :edit, :update]
 
   def edit_address
-    render :edit
+    render :edit_address
   end
 end
--- a/app/controllers/spree/users_controller_decorator.rb
+++ b/app/controllers/spree/users_controller_decorator.rb
@@ -1,3 +1,9 @@
 Spree::UsersController.class_eval do
   layout 'darkswarm'
+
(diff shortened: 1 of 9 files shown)


spec/features/consumer/account_spec.rb
```diff
@@ -60,4 +60,33 @@
...
+      fill_in('user[ship_address_attributes][firstname]', with: 'John')
+      fill_in('user[ship_address_attributes][lastname]', with: 'Doe')
+      fill_in('user[ship_address_attributes][address1]', with: '11 Lovely Street')
+      fill_in('user[ship_address_attributes][phone]', with: '123-456-7890')
+      fill_in('user[ship_address_attributes][city]', with: 'Merbourne')
+      fill_in('user[ship_address_attributes][zipcode]', with: '3127')
+
+      click_button('Update')
+
+      expect(page).to have_content '11 Lovely Street'
```
mkllnk: Do you need to make sure that the data actually saved?

I'm just imagining an "Update" button that doesn't do anything or works very slowly. The next three lines would just test the currently filled out form that is not submitted yet.

app/views/spree/users/show.html.haml
```diff
@@ -7,6 +7,34 @@
       .account-summary{"data-hook" => "account_summary"}
         = @user.email
         (#{link_to t(:edit), spree.edit_account_path})
+      %hr
+      %h3= t(:my_addresses)
```
mkllnk: When I did the internationalisation first, it was easy to write `t(:my_addresses)`. But in most views it's better to use translations in a local scope: `t('.my_addresses')`. The locale then needs an entry like:
```yml
en:
  spree:
    users:
      show:
        my_address: 'My Addresses'
```

This way you avoid any conflicts between views and the locale file is more structured. What do you think?

app/views/spree/users/edit_address.html.haml
```diff
@@ -0,0 +1,15 @@
+.darkswarm
+  .row
+    = render :partial => 'spree/shared/error_messages', :locals => { :target => @user }
```
mkllnk: Can't you use the new ruby syntax here? `render partial: '...', locals: {...}`

bingxie: yes, sure!

app/views/spree/users/show.html.haml
bingxie: very good suggestion!

spec/features/consumer/account_spec.rb
bingxie: good spot, will improve it.

bingxie:
Just updated, could you have a look again @mkllnk ?

app/controllers/spree/users_controller_decorator.rb
```diff
@@ -1,3 +1,21 @@
 Spree::UsersController.class_eval do
   layout 'darkswarm'
+
+  prepend_before_filter :load_object, :only => [:edit_address, :update_address, :show, :edit, :update]
```
mkllnk: Still old syntax here. I use `rubocop` to check my code before committing.

app/controllers/spree/users_controller_decorator.rb
```diff
@@ -1,3 +1,21 @@
...
+    country_id = Spree::Config[:default_country_id]
+    @user.build_ship_address(country_id: country_id) unless @user.ship_address
+    @user.build_bill_address(country_id: country_id) unless @user.bill_address
+
+    render :edit_address
+  end
+
+  def update_address
+    if @user.update_attributes(params[:user])
+      redirect_to spree.account_url, :notice => t(:account_updated)
```
mkllnk: Old Ruby syntax here as well. And did you try the relative translation key `.account_updated` here? I'm not sure how it works with notices. I sometimes use the relative key and then look at the error message in the front end: "translation key ... missing". So I know where Rails tries to find the translation.

spec/features/consumer/account_spec.rb
```diff
@@ -60,4 +60,34 @@
...
+    it 'updates shipping address' do
+      fill_in('user[ship_address_attributes][firstname]', with: 'John')
+      fill_in('user[ship_address_attributes][lastname]', with: 'Doe')
+      fill_in('user[ship_address_attributes][address1]', with: '11 Lovely Street')
+      fill_in('user[ship_address_attributes][phone]', with: '123-456-7890')
+      fill_in('user[ship_address_attributes][city]', with: 'Merbourne')
+      fill_in('user[ship_address_attributes][zipcode]', with: '3127')
+
+      click_button('Update')
+      expect(page).to have_content 'My Addresses'
```
mkllnk: Yes. That should do the trick.

app/controllers/spree/users_controller_decorator.rb
bingxie: Good idea, if you run rubocop through the whole project there so many offenses. How did you check you code?

app/controllers/spree/users_controller_decorator.rb
bingxie: I just removed this notice, seems like on the front end page, there is no place to show it.

app/views/spree/users/edit_address.html.haml
```diff
@@ -0,0 +1,15 @@
+.darkswarm
+  .row
+    = render partial: 'spree/shared/error_messages', locals: { target: @user }
+  .row
+    = form_for Spree::User.new, :as => @user, :url => main_app.account_update_address_path, :method => :put do |f|
```
mkllnk: More old syntax here. :-)

app/views/spree/users/edit_address.html.haml
```diff
@@ -0,0 +1,15 @@
...
+      .small-12.medium-6.columns
+        %h3.text-center= t('.edit_billing_address')
+        = render partial: 'spree/users/address_form', locals: { f: f, address_type: :bill_address }
+
+      .small-12.medium-6.columns
+        %h3.text-center= t('.edit_shipping_address')
+        = render partial: 'spree/users/address_form', locals: { f: f, address_type: :ship_address }
+
+      .medium-12.columns.text-center
+        = f.submit t(:update), :class => 'button primary'
```
mkllnk: And here.

config/routes.rb
```diff
@@ -48,6 +48,9 @@
     end
   end
 
+  get '/account/edit_address', :to => 'spree/users#edit_address'
```
mkllnk: And here.

bingxie: hi @mkllnk, after you think the code is good, then can you help me to deploy it on staging and let @sstead have a look on Wednesday?

sstead:
Address isn't carrying through to the back end....
Default billing/shipping addresses are set on a shop by shop basis. When a customers saves an address in their account, it doesn't overwrite the address they've saved 'as default' during checkout, or that the enterprise has saved... It won't even populate the address details for a customer who's on a shop's customer list with no saved address. What's the point of this function if the address saved by the customer doesn't carry through to their shop's lists?

image
image

Idea...?  I think if a customer updates their address here, it should update as such in all shop customer lists that they are on, and all future shops they shop in. Then we should remove the 'save as default' checkbox in checkout.

I think a 'billing address same as shipping address' tick box would solve this...
image

daniellemoorhead:
---Doing a PR Cleanup---

@oeoeaio @mkllnk what should we do with this PR from Bing? It was related to Standing Orders work, wondering if it's needed/replaced/still relevant?

oeoeaio:
I think still relevant, but perhaps not 100% vital for the initial release. Will revisit once we've release the first cut of standing orders.

myriamboure:
The connected issue #1132 is in the [Subscription project](https://github.com/openfoodfoundation/openfoodnetwork/projects/15) backlog, so I propose to close that PR, and when a dev pick the issue in the backlog to work on it, he will reopen it. @oeoeaio @mkllnk if you have any objection please reopen.

PR #1199, opened 2016-10-26
Mark order cycle form as dirty when removing fees

### LINKED ISSUE #1165: Update button not enabled after deleting a coordinator's fee
When i delete a coordinators fee from an order cycle the update button stays greyed out.   I can only save this change if i make  another change to the order cycle which then enables the update button
screenshot 2016-09-22 13 07 20

levent:
Resolves an issue where removing coordinator/exchange fees
wasn't allowing users to save the OC.

Fixes #1165


oeoeaio:
Hey @levent. Thanks very much for this PR. Will get it through ASAP.

Have been trying to think of a way to catch this kind of issue in a more comprehensive way, but I haven't come up with anything better than an exhaustive feature spec - which is obviously a nightmare to maintain. I think this is great for now. Any thoughts you have much appreciated.

levent:
@oeoeaio cheers. I'll get back to you if/when I can think of something better in the meantime.

PR #1201, opened 2016-10-27
Resources module

oeoeaio:
Moving ngResource oriented services into a separate angular module.


app/assets/javascripts/admin/index_utils/services/dereferencer.js.coffee
```diff
@@ -3,7 +3,8 @@ angular.module("admin.indexUtils").factory 'Dereferencer', ->
     dereference: (array, data)->
       if array
         for object, i in array
-          array[i] = data[object.id]
+          match = data[object.id]
+          array[i] = match if match?
```
mkllnk: I tried something similar in the frontend. I ran into the problem of having a mixed array after that. Some objects were de-referenced and another were not. My solution was to have three arrays:

1. The unreferenced source array
2. The target array containing de-referenced objects. It is potentially sparse, not available objects are just not present. Otherwise other code fails when accessing a missing object.
3. A list of not resolved objects (ids) to resolve them later.

I needed that to lazy-load enterprises in the shopfront. The implementation is a bit more complicated: app/assets/javascripts/darkswarm/services/dereferencer.js.coffee

PR #1150, opened 2016-09-07
Uk/1025 from address

stveep:
See #1025 and previous PR #1044 from @lin-d-hop. This adds a toggle for the 'send confirmation from the enterprise email' behaviour. It's off by default (perhaps should be on?) and can be set in the super admin mail settings.

Issue #1153, opened 2016-09-14
Whitelabelling for Enterprise Emails to Customers


MagsHall:
UK users have requested that order confirmation emails can be personalised to the hub. 

Phase One:
- Hub name in subject line of order confirmation and order cycle total emails
- Hub contact email as 'from' in automated emails

Future:
- Personalisation of the emails to include hub logo, design template, etc. 

Having Hub contact as 'from' email will also solve the issue of customers and producers contacting OFN rather than their hub directly for queries, which sometimes happens and relies on us forwarding messages onto hub co-ordinators.

PR #1203, opened 2016-10-29
Fix from address

stveep:
Fixing the mess from PR #1150 :)

### DIFF
--- a/app/mailers/producer_mailer.rb
+++ b/app/mailers/producer_mailer.rb
@@ -12,9 +12,19 @@ class ProducerMailer < Spree::BaseMailer
 
     subject = "[#{Spree::Config.site_name}] Order cycle report for #{producer.name}"
 
+    # Set from address to the coordinator (optional)
+    if Spree::MailMethod.current
+      if Spree::MailMethod.current.prefers_send_from_enterprise_address?
+        from_email = @coordinator.email
+      else
+        from_email = from_address
+      end
+    end
+    from_email ||= from_address
+
     if has_orders? order_cycle, producer
       mail(to: @producer.email,
-           from: from_address,
+           from: from_email,
            subject: subject,
            cc: @coordinator.email)
     end
@@ -22,6 +32,7 @@ class ProducerMailer < Spree::BaseMailer
 
 
   private
+
 
   def has_orders?(order_cycle, producer)
     line_items_from(order_cycle, producer).any?
--- a/app/mailers/producer_mailer.rb
+++ b/app/mailers/producer_mailer.rb
@@ -16,7 +16,6 @@ class ProducerMailer < Spree::BaseMailer
       mail(to: @producer.email,
            from: from_address,
            subject: subject,
-           reply_to: @coordinator.email,
            cc: @coordinator.email)
     end
   end
--- a/app/mailers/producer_mailer.rb
+++ b/app/mailers/producer_mailer.rb
@@ -26,6 +26,7 @@ class ProducerMailer < Spree::BaseMailer
       mail(to: @producer.email,
            from: from_email,
            subject: subject,
+           reply_to: @coordinator.email,
            cc: @coordinator.email)
     end
   end
--- a/app/models/spree/mail_method_decorator.rb
+++ b/app/models/spree/mail_method_decorator.rb
@@ -1,0 +1,5 @@
+Spree::MailMethod.class_eval do
+  preference :send_from_enterprise_address, :boolean, default: false
+
+  attr_accessible :preferred_send_from_enterprise_address
+end
--- a/app/views/spree/admin/mail_methods/_form.html.haml
+++ b/app/views/spree/admin/mail_methods/_form.html.haml
@@ -15,7 +15,9 @@
           = f.label :preferred_enable_mail_delivery, t(:enable_mail_delivery)
         .field
           = f.check_box :preferred_send_from_enterprise_address, {:style => "vertical-align:middle;"}
-          = f.label :preferred_send_from_enterprise_address, t(:send_from_enterprise_address)
+          = f.label :preferred_send_from_enterprise_address, t('admin.mail_methods.send_from_enterprise_address')
+          %span.info
+            = t('admin.mail_methods.caveat')
         .field
           = f.label :preferred_mails_from, t(:send_mails_as)
           %br/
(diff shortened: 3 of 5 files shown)


app/mailers/producer_mailer.rb
```diff
@@ -12,18 +12,28 @@ def order_cycle_report(producer, order_cycle)
...
+      end
+    end
+    from_email ||= from_address
+
     if has_orders? order_cycle, producer
       mail(to: @producer.email,
-           from: from_address,
+           from: from_email,
            subject: subject,
-           reply_to: @coordinator.email,
```
oeoeaio: @stveep, I don't think we should be removing this, as the reply-to address should be the shop email address, regardless of which address the email is sent from... right?

For example, I regularly get people replying to their order confirmation emails (and I guess producer emails are in much the same boat). If those replies start going to `Spree::Config.emails_sent_from`, I am never going to see them...

stveep:
Some background from @lin-d-hop and @mkllnk is here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/1044

I just added the feature toggle so I can't really answer your question :)

oeoeaio:
Thanks @stveep! @mkllnk, you probably know more about this than me - is it really ok to send emails with a from address we don't own? Even if we can get away with it?

Either way, I think we should definitely keep the reply-to as the coordinators email address, because otherwise we will have a communication breakdown in the case where this new option is not being utilised.

mkllnk:
I think Rob is right here. We still need the reply-to field in the case of a common sender email that is not the enterprise email address.

stveep:
I restored the reply_to. This will be set to the coordinator regardless of where the email comes from, is that ok @lin-d-hop ?

lin-d-hop:
Hi @oeoeaio @stveep @mkllnk, 
The reason that the reply_to was removed is that not all mail clients utilise this field and the result was unexpected behaviour when sometimes responses went to the instance and sometimes the hub manager. This was the problem that prompted this whole issue in the first place. I'm happy for it to stay in but just wanted to flag this up.

WRT sending mail from another account, perhaps in the 'How To' for this feature we could make a note that enabling this feature allows an OFN instance to send mail on your behalf. This feature is important to hubs that wish to whitelabel the tools, but also should explicitly require permission. Is that something we can do @sstead? (On that note what is our process for translating features into the guide?)

mkllnk:
@lin-d-hop That's really interesting. I didn't know that there are mail clients that ignore the reply-to field. But we are not alone:

- http://stackoverflow.com/questions/2421234/gmail-appearing-to-ignore-reply-to
- https://meta.discourse.org/t/reply-keys-reply-email-addresses-not-always-used-mail-client-issues/38787/5
- https://bugzilla.mozilla.org/show_bug.cgi?id=933555

The reply-to field seems to be ignored more often if the from field contains one of your own identities.

sstead:
@lin-d-hop I will usually add details about new features to the user guide after I've tested them, or just before/after the release notes go out.

sstead:
What do you mean by whitelabeling the tools?

oeoeaio:
I honestly can't see a resolution to this where we get both white labelling (ie. no mention of OFN in the email, including the 'from' email address) AND reliable email delivery (ie. not getting stuck in spam filters), UNLESS we get enterprise users who want to white-label to allow us to send emails on their behalf by providing us with all the information we need to authenticate with the relevant SMTP server properly - which sounds like a bad idea to me...

The reply-to thing not working is kind of a separate issue. I wonder if there is a way we can quantify how big a problem this actually is. Surely it's a bug in whatever software it exists in. I know that after several years of using OFN to take thousands of orders, I have never been made aware of anyone having an issue getting through to us by hitting reply, which is something people do every single week.

lin-d-hop:
@oeoeaio You are right that the reply-to and white labelling are two separate issues. But this came about as a single solution to two separate issues:
1) finding 2-3 emails a week in which we've had to manually forward as users mail clients aren't using the reply-to field. This seems unmanageable going forward since we are so small with only two active hubs.
2) repeated requests from users/potential users for white labelling.

It has led me down a path of learning a lot about email and deliverability through! I agree deliverability would be a problem if this was not handled correctly.
One option could be that white labelling becomes an optional paid feature. Sendgrid do allow user to white label domains by providing validation to the DNS.
We have already done this in the UK for our domain, and it worked a treat for a month until spam bots very recently started to flag up some of our transactional emails again. We are continuing to delve into this.
I do think that white labelling will be a valuable feature for OFN in the future and I would like to continue to spec if out. But now that I have learned a lot more about this I agree that we need to do it properly. I'd like to put the release of this issue on hold until we add to the mix a higher level of configurability and solve the existing deliverability issues we are experiencing in the UK.

@stveep Apologies that you worked on this and now it is looking like it might be better to hold off on releasing for the time being! This is a very important feature and a priority for the UK, just more complicated than first thought (as usual).

I'd like to spec it out more here:
http://community.openfoodnetwork.org/t/whitelabelling-for-enterprise-emails-to-customers/730
...before adding issues to the epic #1153

mkllnk:
In the mean time I read more about the Sender Policy Framework (SPF) and agree that it is quite difficult to change the from address to another domain. If you did that for the UK, you probably added an SPF record to your UK DNS to allow sendgrid to send email on your behalf. Maybe you need to update that record? That could be if Sendgrid changed their server configuration. But you probably checked that already.

Here is a summary of solutions:

- Use the food hub name, an OFN email address and reply-to.
  ```
  From: Stroudco <noreply@openfoodnetwork.org.uk>
  Reply-to: Stroudco <stroudco@gmail.com>
   ```
- All enterprises that want white-labelling need their own domain and have to add SPF records for deliverability.
- All enterprises that want white-labelling need to add their own mail method and save their email password in the OFN database. They should probably create a separate account for that which is just forwarding emails so that the damage in case of a revealed password is minimal.
- Setup your own separate email server that sends and receives emails like Github. Emails would be send from `stroudco+email-token@ofn-email-service.org.uk`. A reply goes there and would forward that email to the saved stroudco managers if the token is valid.

I guess the second method sounds the easiest and best. It makes sense that you have your own domain if you are into white-labelling. We just have to find out why you have deliverability problems.

daniellemoorhead:
---Doing a PR Cleanup---

@lin-d-hop where to next with this?

lin-d-hop:
This one is just not going to work in this way. Good learnings for me at least.

PR #1230, opened 2016-11-24
Fix enterprise user auth

RohanM:
When an enterprise user accesses the admin dashboard, they're getting an access denied error. I found that they were being redirected to `spree/admin/analytics#sync` by `Spree::Admin::OverviewController#check_last_jirafe_sync_time`.

I've added a spec for this scenario (which is common on production), and added permission to analytics to all users who had access to the admin dashboard.

### DIFF
--- a/spec/features/admin/overview_spec.rb
+++ b/spec/features/admin/overview_spec.rb
@@ -110,5 +110,45 @@ feature %q{
         end
       end
     end
+
+    context "with the spree dash configured" do
+      let(:d1) { create(:distributor_enterprise) }
+
+      before do
+        stub_jirafe
+        @enterprise_user.enterprise_roles.build(enterprise: d1).save
+      end
+
+      around do |example|
+        with_dash_configured { example.run }
+      end
+
+      it "has permission to sync analytics" do
+        visit '/admin'
+        expect(page).to have_content d1.name
+      end
+    end
+  end
+
+  private
+
+  def stub_jirafe
+    stub_request(:post, "https://api.jirafe.com/v1/applications/abc123/resources?token=").
+      to_return(:status => 200, :body => "", :headers => {})
+  end
+
+  def with_dash_configured(&block)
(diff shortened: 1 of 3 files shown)


spec/features/admin/overview_spec.rb
```diff
@@ -3,31 +3,18 @@
...
-
   context "as an enterprise user" do
-    before :each do
+    before do
       @enterprise_user = create_enterprise_user
       Spree::Admin::OverviewController.any_instance.stub(:spree_current_user).and_return @enterprise_user
       quick_login_as @enterprise_user
     end
 
-    context "with no enterprises" do
```
mkllnk: Is this outdated? Just to test if I got it right:

- People create the first enterprise during the signup process.
- If they haven't done that, the admin panel redirects to /unauthorized.

Correct?

spec/features/admin/overview_spec.rb
```diff
@@ -123,5 +110,45 @@
...
+    context "with the spree dash configured" do
+      let(:d1) { create(:distributor_enterprise) }
+
+      before do
+        stub_jirafe
+        @enterprise_user.enterprise_roles.build(enterprise: d1).save
+      end
+
+      around do |example|
+        with_dash_configured { example.run }
```
mkllnk: Does this mean the error is only on staging, because Dash is configured there, but not on production?

spec/features/admin/overview_spec.rb
RohanM: Yeah, that's correct. If they don't have an enterprise, they're not an enterprise user. That's how they're granted access to the admin panel in the first place.

spec/features/admin/overview_spec.rb
RohanM: Aah, good question. Without knowing the details of which environments were breaking, I believe that's what was going on. Even if we're not using it, I think it'll be good to have this here so that if someone does enable dash, they don't get mysterious errors.

PR #1231, opened 2016-11-24
Correct spelling of further in translations

levent:


levent:
@lin-d-hop / @stveep a tiny pull request.

When we run into stuff like this, do we assign/comment/label/just pr ?

stveep:
Just a PR is good I think - thanks!

@daniellemoorhead I still can't add labels or move stuff on the boards but could you get this merged?

Issue #1228, opened 2016-11-23
Allow 'In Stock' to override 'To Order' checkbox


lin-d-hop:
When a product is checked 'To Order' the 'In Stock' quantity is set to infinity to allow the enterprise to supply this product 'To Order'.

This functionality works fine but causes some user confusion when the 'To order' box is checked, then the user inputs an 'In Stock' quantity. The 'To Order' checkbox overrides the 'In stock' quantity and the value remains at Infinity.

A simple UX improvement would be set the 'To Order' checkbox to unchecked when the user inputs a value into the 'In Stock' field.

Note - Only 1.1 is a change to current functionality.

1. As an enterprise user I input a value into the 'In Stock' field:
 1.1 _If the 'To Order' box is checked -> the box is set to unchecked._
 1.2 If the 'To Order' box is unchecked -> no change
2. As an enterprise user I check the 'To Order' checkbox:
 2.1 If the 'In Stock' field contains a number -> the field is set to Infinity
 2.2 If the 'In Stock' field is set to Infinity -> no change

![instocktoorder](https://cloud.githubusercontent.com/assets/8669848/20571991/86034b18-b1a1-11e6-9daa-e7d62ee43fa8.png)

Matt-Yorkley:
I can implement something for this in the next couple of days. I think a bit of javascipt that updates the form as the user is interacting with it would be good, and quite easy to do. I jotted down some notes for it. 

Maybe as a separate issue, but relating to this one: it might be good in a few places where the UI is a bit vague to implement some sort of tooltip functionality for some of the more confusing parts of the backend. Maybe a question mark with a rollover that gives a short explanation.

lin-d-hop:
@Matt-Yorkley UX is being reviewed. This issue is small but many more to come. Let's keep issues to issues and not blow them up to a whole vague UX scope eh?

Matt-Yorkley:
Fair point, I was just trying to be thorough as the issue was largely about UX and lack of clarity. I've added a "what's this?" help tooltip as well as the UX improvements to the form. Pull request submitted and ready for testing.

lin-d-hop:
@Matt-Yorkley Could you explore adding some specs to capture this behaviour? This seems like a change that should have test coverage.

lin-d-hop:
@Matt-Yorkley Ignore above, was on the wrong issue (on my phone on a train in Germany = hard to navigate). Test coverage less important here me thinks.

[...]

PR #1234, opened 2016-11-25
Product variant form script

Matt-Yorkley:
Addressing issue #1228. Added some javascript to make the product variant form a bit more interactive and show the user what's happening when the "to order" checkbox is selected.


app/overrides/spree/admin/variants/_form/on_demand_tooltip.html.haml.deface
```diff
@@ -0,0 +1,3 @@
+/ insert_bottom "[data-hook='on_demand']"
+%div{'ofn-with-tip' => "Items made to order do not have a set stock level, such as loaves of bread made fresh to order."}
```
lin-d-hop: This should be translatable for i18n.

lin-d-hop:
@rhodri Would you be able to review this one? Doesn't require much rails/angular knowledge so me thinks you're v. qualified :-)

Issue #375, opened 2015-02-06
The appearance of text in About Us editor doesn't perfectly correspond to what's displayed on the site.


kirstenalarsen:
Specifically, if you hit enter once in this text editor after a paragraph, you'll get a actual paragraph break in the site. But within the text editor, you don't get a clean paragraph break, so people are inclined to hit enter twice.

The worst that will happen if this is not fixed is that the about us sections will have 2 line gaps between paragraphs.

OliverUK:
@lin-d-hop 
I'm a bit lost now - remember this is new to me - how all this different issues/commits etc are connected with each other. Do I still need to deploy something to test this? 
Right now I can't get an about us text to display on the staging front end at all.

<img width="413" alt="screen shot 2016-12-10 at 21 36 59" src="https://cloud.githubusercontent.com/assets/17590921/21076449/61bcdc08-bf23-11e6-9b18-1d558a3d5157.png">

<img width="666" alt="screen shot 2016-12-10 at 21 46 12" src="https://cloud.githubusercontent.com/assets/17590921/21076451/69abce1a-bf23-11e6-8f64-f350f706bfa4.png">

lin-d-hop:
Hi @OliverUK 
So I deployed the PR related to this issue to staging. In particular this issue was testing line breaks in the text editors.

![aboutedit](https://cloud.githubusercontent.com/assets/8669848/21140410/c8b61c9a-c12f-11e6-9dcb-1b224800e5b4.png)
![aboutdisplay](https://cloud.githubusercontent.com/assets/8669848/21140409/c8b49320-c12f-11e6-8ee5-a1103166d130.png)


As an aside, it looks like the staging deploy you tested on above broke the About, no? Was this the PR for #1020? Looks like that should be tested on that fix.

daniellemoorhead:
Thanks @lin-d-hop, when something's ready to hand over to the Oz team can you add Sally to assignees and put the aus review label on it. Makes it easier to track then :)

lin-d-hop:
Hi @Matt-Yorkley, 
Given the entanglement between this fix and #1020 do you think it might be an idea to combine them in a new branch and test all of the display issues that popped up in @ssteads testing of #1327? Just a thought, feel free to disagree. The testing of the two branches is feeling more complicated than it needs to be from where I'm sitting.

lin-d-hop:
Line breaks fix merged so closing this issue.

Issue #421, opened 2015-03-14
Create drop-down for easier switching between Enterprises for management


kirstenalarsen:
The Enterprise Index page shows a User all the Enterprises that they manage

![image](https://cloud.githubusercontent.com/assets/2911176/6649729/43658ff8-ca46-11e4-82a3-70f95a983e43.png)

Selecting an Enterprise Profile takes you into the Enterprise Management page
![image](https://cloud.githubusercontent.com/assets/2911176/6649682/21be96ac-ca45-11e4-8433-1b6af7457d0d.png)

To switch between management of different enterprises, the User currently needs to reload the Enterprise Index page, either by clicking on the top menu, or using the 'back to Enterprises list' button. This is annoying - would be much better to be able to change between enterprises directly  

TO DO
- Create a Dropdown selector next to the 'back to Enterprises list' that shows the names of the Enterprises that the User manages (on the right)
- Selecting the Enterprise name takes User directly into the Enterprise management page of that Enterprise
- warning if unsaved changes before going to new page (see https://github.com/openfoodfoundation/openfoodnetwork/issues/420)

Matt-Yorkley:
@kirstenalarsen I'm working on this enterprise switcher dropdown. Do you have a preference for the layout?

Option A:

![screenshot from 2016-11-26 15-40-17](https://cloud.githubusercontent.com/assets/9029026/20641445/ad7d74a2-b3ef-11e6-9f8c-64f6aefc2340.png)

Option B:

![screenshot from 2016-11-26 15-43-08](https://cloud.githubusercontent.com/assets/9029026/20641446/ad9f87d6-b3ef-11e6-8d08-a9047377f336.png)

kirstenalarsen:
Not a strong one, maybe slight inclination to b . . @sstead?

sstead:
Looks great! Yes I agree, B is more similar to existing appearance.
________________________________

lin-d-hop:
Note this PR also fixes #420 so no need to redeploy @OliverUK

OliverUK:
@lin-d-hop 
I have tested this using the super admin login. The drop-down appears when editing an enterprise. I successfully switched between enterprises and mad changes to some and saved them, all without noticing anything untoward.

Please note, Lynne, I've only tested this in super admin. It seems I have to register my normal user account anew every day on staging and today I was once again unable to confirm the account (no confirmation email).

OliverUK:
[...]

Issue #390, opened 2015-02-20
'Users' should be lower down the panels - perhaps at the bottom?


kirstenalarsen:
![image](https://cloud.githubusercontent.com/assets/2911176/6282500/8fbb4168-b92a-11e4-9b8a-a5a6d5e32f7a.png)

OliverUK:
Tested in Safari on Mac and Users now appears on the bottom of the menu. Clicking on it works and their is probably not much else to say. Same result whether I go through the Enterprises menu or the dashboard.

![screen shot 2016-12-14 at 14 05 41](https://cloud.githubusercontent.com/assets/17590921/21185035/fab6f598-c206-11e6-89af-5624f26ef724.png)

sstead:
This one isn't going to make it into release 1.8.7

sstead:
Some minor menu display inconsistencies, but this appears to already be happening on production...

Issue #420, opened 2015-03-14
Warning message if enterprise user leaves enterprise management without update / save


kirstenalarsen:
An enterprise user can make changes within any tab of their enterprise management screen, and switch between tabs. Those changes are saved when they hit the update button (at the bottom)

There is no warning if the User selects 'back to Enterprise list' or clicks elsewhere (e.g. top menu) and has not saved changes.

![image](https://cloud.githubusercontent.com/assets/2911176/6649720/008a3c88-ca46-11e4-93ff-083b2cfc67f8.png)

TODO
- create warning message if User clicks link to leave the page and there are unsaved changes in any tab. Generic / standard warning e.g. "You have unsaved changes, do you want to leave this page or stay"

elf-pavlik:
possibly one could reuse

https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee#L11-L19

OliverUK:
@lin-d-hop @sstead 

I've tested this separately from the enterprise switcher test in a separate issue but I think it's in the same piece of code??

It worked for me once after deploying to staging but after the first time, further testing does not normally produce a warning message even if unsaved changes are present. 

I logged out and logged back in again and tried again.
If I get a warning message and then choose Stay on Page and then save, I can make another change attempt and still get the warning message and so then it works.

If I choose to leave the page without saving, this seems to break the warning message and I don't get any more warnings from then on. This seems to happen even if I'm editing a different enterprise with the exception of the new switcher drop-down, which seems to produce warning messages ok in any case.

I tested this in Safari using spree@example.com.

I also noticed that creating a tag rule throws up a superfluous warning.

![screen shot 2016-12-13 at 15 22 11](https://cloud.githubusercontent.com/assets/17590921/21145969/0df09408-c148-11e6-80a7-e9744a0ce527.png)
![screen shot 2016-12-13 at 15 22 20](https://cloud.githubusercontent.com/assets/17590921/21145970/0dfb2134-c148-11e6-9647-9644e8ba85cc.png)

lin-d-hop:
@Matt-Yorkley Could you look at @OliverUKs testing notes above?

Matt-Yorkley:
@OliverUK Updated and ready for testing.

OliverUK:
Might have to remind me what I need to deploy for re-testing. I tried the latest uk/switcher build but it failed.

[...]

PR #1238, opened 2016-11-27
Various enterprise management adjustments

Matt-Yorkley:
Reordered the Enterprises edit page submenu for #390.

Changed the way text displays in the editor on the About Us tab to be visually consistent with the spacing of paragraphs on the front-end, to resolve issue #375.

Created a dropdown in Enterprise edit page for switching between enterprises for #421, including a popup warning if attempting to switch whilst there is unsaved data in the current page form.


kirstenalarsen:
Awesome :)

Matt-Yorkley:
Now also addresses #420, by displaying a confirmation popup if the user tries to navigate away from the page by other methods (not just the dropdown) whilst the form contains unsaved changes. Refactored a bit and added some comments for clarity.

kirstenalarsen:
It's really exciting to see these annoying things getting sorted out :)

Matt-Yorkley:
That key value was a funny one. I used `'data-url'` instead of `:data-url` as the key because I was getting an error saying that dashes are not allowed in symbols. Just looked it up and the alternatives are `:'data-url'`, or HAML now converts `:data_url` to `'data-url'` when rendering. I'll keep it in mind.

As for pulling the second function out into a separate file, I thought it might be problematic as the first function interacts with the second. May be it's not an issue. Looking at it now though, maybe it would be a good idea to use jquery's event namespacing on the `beforeunload` event so that bind and unbind could be called on the specific handler without interfering with other potential bindings...

Also it needs a translation on the message string...

Matt-Yorkley:
@stveep Am I right in thinking if I pull the function out into a file and re-include it via the Rails asset pipeline, it'll be included site-wide?

mkllnk:
@Matt-Yorkley 

> I used 'data-url' instead of :data-url as the key because I was getting an error saying that dashes are not allowed in symbols.

There is another solution to that: `data: { url: 'value' }`
Thats good if you have multiple `data-*` attributes. Otherwise `"data-url" => "value"` is okay I think. I didn't know that HAML converts underscores. Does our version support it? Do we use that anywhere?

> Am I right in thinking if I pull the function out into a file and re-include it via the Rails asset pipeline, it'll be included site-wide?

Yes, that is correct. All JS assets are bundles into one big file `all.js`.

Matt-Yorkley:
As an angular training excercise, I've dropped the jquery completely and rewritten it as an angular directive and a service. I found an existing angular NavigationCheck factory that only runs on the enterprise page but didn't seem to be working, and fixed it (the callback was passing a function that checked the form's $dirty state, but it was getting run in a separate file that didn't have access to the same scope). That and the new service can now be used on other parts of the site where they're needed.

stveep:
Nice one! Think it's good to do these things in Angular where possible - good practice and makes the project a bit more consistent.

Not sure why the spec is failing, it passes for me locally. Probably just needs to be rebased on master?

lin-d-hop:
@Matt-Yorkley Are all the commits in here from Rohan the result of you updating your branch but not rebasing?
Can you cherry-pick out the relevant commits so that we can test this more easily? And when you do, can you put them in a branch of the pattern uk/* so that they can be tested more easily by oliver?
Thanks!

Matt-Yorkley:
Rebased the branch as suggested, tests passing.

lin-d-hop:
@Matt-Yorkley This PR still shows 86 commits and 148 files changed. Can you clean it up please? This might mean cherry picking onto a new branch (in which case use uk/* pattern so it can be tested easily).

We then need to test it. This would be much easier if there was one issue per PR to test so we can document the testing.

It's really great you've been so helpful with all the little fixes here @Matt-Yorkley... but now we need to test it and demonstrate the fixes in a clear way to the international community this clump of fixes becomes confusing. If you could help us out be splitting out and clarifying what is fixed where that would make the whole process much more manageable.

Matt-Yorkley:
Closing to tidy up and resubmit as multiple smaller PRs that are simpler for uk testing.

Issue #1020, opened 2016-05-31
Enable formatting in product description


myriamboure:
Continuing from https://github.com/openfoodfoundation/openfoodnetwork/issues/953

A request has come from our users to be able to format product description (basic: bold, list, go to the next line, etc.)
Today it's not very user friendly...

![image](https://cloud.githubusercontent.com/assets/8135376/15666602/ece57f60-2711-11e6-8b5b-89d84f5aa890.png)

myriamboure:
Why not using styling with markdown for example? Like it's done in github :-)

myriamboure:
Any solution on that @sstead ? Did you have any request from users to be able to format the product description?

sstead:
Hi @myriamboure, I ran this idea past Rob and Rohan and they said there are a couple of different options. Wizzywig would be the easiest option in the short term (this is what we have for About Us profile description). However this needs to be thought out, because it will mean saving data in HTML, from which point it's hard to go back. Markdown is also appealing, but again needs thought. So essentially, yes there are options for doing this, it's not a huge task but it will need some consideration, and some resources to make happen. We haven't had requests from users for this feature in Australia.

OliverUK:
Thanks for pointing me here @myriamboure! I've added the UK label.

lin-d-hop:
@OliverUK Just wanted to check why you've added this to UK Current? Is it an urgent priority or something Stroudco in particular would like to see?

OliverUK:
@lin-d-hop yes. It's not critical but certainly seems the lack of any formatting results in a rather incomplete package for hubs and users. Here you want to be able to sell your items, write about it, possible add links or a list. Currently it's really only good enough for one or two sentences. It doesn't even do paragraphs.

lin-d-hop:
Ok. I agree it can make a big difference, particularly for producers. Let's get it explored.

@stveep (or @lebreeze)
Would it be possible to have a few options put on the table here? Would you be able to explore options for formatting the product descriptions including enabling HTML and markdown with pros and cons of each and a recommendation of the best course of action. 
It would be good to have flexibility to upgrade or change later if requirements change so please bear this in mind. 
Usability is very important. From a UX perspective we want to make it very easy for users to create attractive product descriptions. 

We interact with a number of examples ourselves, within github and discourse... so we have an idea of what works well :-)

lin-d-hop:
Tagging @levent as I tagged incorrectly above.

[...]

PR #1239, opened 2016-11-27
Product descriptions formatting

Matt-Yorkley:
For issue #1020.

-Added WYSIWYG to product description forms on backed. 

-Modified frontend API to display formatted (properly sanitised) html for product descriptions.

-Made some UI improvements to textAngular implementation, including better visual feedback on selected options.

screenshot from 2016-11-27 14-07-56

New button behaviour utilising the previously unused active css class:
screenshot from 2016-11-27 13-33-54
screenshot from 2016-11-27 13-34-03

### DIFF
--- a/app/assets/javascripts/templates/product_modal.html.haml
+++ b/app/assets/javascripts/templates/product_modal.html.haml
@@ -16,7 +16,7 @@
 
     %div{"ng-if" => "product.description_html"}
       %hr
-      %p.text-small{"ng-bind-html" => "product.description_html"}
+      %p.text-small{"ng-bind-html" => "::product.description_html"}
       %hr
 
   .columns.small-12.large-6
--- a/app/assets/javascripts/templates/product_modal.html.haml
+++ b/app/assets/javascripts/templates/product_modal.html.haml
@@ -14,10 +14,9 @@
     .filter-shopfront.property-selectors.inline-block
       %filter-selector{ 'selector-set' => "productPropertySelectors", objects: "[product] | propertiesWithValuesOf" }
 
-    %div{"ng-if" => "product.description"}
+    %div{"ng-if" => "product.description_html"}
       %hr
-      %p.text-small{"ng-bind" => "::product.description"}
-      {{product.description}}
+      %p.text-small{"ng-bind-html" => "product.description_html"}
       %hr
 
   .columns.small-12.large-6
--- a/app/assets/javascripts/templates/product_modal.html.haml
+++ b/app/assets/javascripts/templates/product_modal.html.haml
@@ -17,6 +17,7 @@
     %div{"ng-if" => "product.description"}
       %hr
       %p.text-small{"ng-bind" => "::product.description"}
+      {{product.description}}
       %hr
 
   .columns.small-12.large-6
--- a/app/overrides/spree/admin/products/_form/add_description_wysiwyg.html.haml.deface
+++ b/app/overrides/spree/admin/products/_form/add_description_wysiwyg.html.haml.deface
@@ -1,2 +1,3 @@
 / replace "[data-hook=admin_product_form_left] code[erb-loud]:contains('f.text_area :description')"
 %text-angular{'id' => 'product_description', 'ng-model' => 'product.description', 'name' => 'product[description]', 'class' => 'text-angular', 'ta-toolbar' => "[['bold','italics','underline','clear']]"}
+  != @product[:description].to_s.html_safe
--- a/app/overrides/spree/admin/products/_form/add_description_wysiwyg.html.haml.deface
+++ b/app/overrides/spree/admin/products/_form/add_description_wysiwyg.html.haml.deface
@@ -1,3 +1,3 @@
 / replace "[data-hook=admin_product_form_left] code[erb-loud]:contains('f.text_area :description')"
-%text-angular{'id' => 'product_description', 'name' => 'product[description]', 'class' => 'text-angular', 'ta-toolbar' => "[['bold','italics','clear']]"}
+%text-angular{'id' => 'product_description', 'name' => 'product[description]', 'class' => 'text-angular', 'textangular-strip' => true, 'ta-paste' => "stripFormatting($html)", 'ta-toolbar' => "[['bold','italics','clear']]"}
   = sanitize(@product.description)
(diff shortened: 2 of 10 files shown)


stveep:
This looks good to me. Underlining doesn't seem to actually do anything - but unless theres a specific case for it, I don't think this should even be available as an option - it would just make stuff look like hyperlinks.

anster1
anster2

Nothing to do with what you've done here, but the layout of the edit page is really ugly and we should probably move the taxons etc. to the left part of the form underneath the descriptions. I just had a quick look and seems a bit of a pain to do with deface.

gapofdoom

Matt-Yorkley:
Removed the underline option. I agree the layout of the edit product page could do with an update but that's a separate issue.

stveep:
Ready for testing.

app/assets/javascripts/templates/product_modal.html.haml
```diff
@@ -14,9 +14,9 @@
     .filter-shopfront.property-selectors.inline-block
       %filter-selector{ 'selector-set' => "productPropertySelectors", objects: "[product] | propertiesWithValuesOf" }
 
-    %div{"ng-if" => "product.description"}
+    %div{"ng-if" => "product.description_html"}
       %hr
-      %p.text-small{"ng-bind" => "::product.description"}
+      %p.text-small{"ng-bind-html" => "product.description_html"}
```
oeoeaio: Do we need to remove the one-time binding?

spec/features/consumer/shopping/products_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    let(:variant) { product.variants.first }
+    let(:order) { create(:order, distributor: distributor) }
+    let(:exchange1) { oc1.exchanges.to_enterprises(distributor).outgoing.first }
+
+    before do
+      set_order order
+    end
+
+    describe "viewing HTML product descriptions" do
+
```
oeoeaio: Just a tiny code style thing: for consistency with the rest of the code base, could I get you to remove these empty lines at the beginning and end of the describe blocks?

app/serializers/api/product_serializer.rb
```diff
@@ -49,10 +49,17 @@ class Api::CachedProductSerializer < ActiveModel::Serializer
...
   has_one :supplier, serializer: Api::IdSerializer
 
+  #return an unformatted descripton
   def description
     strip_tags object.description
   end
 
+  #return a sanitized html description
+  def description_html
+    d = sanitize(object.description, options = {tags: "p, b, strong, em, i"})
```
oeoeaio: Sorry sorry sorry, should have picked this up earlier. 

IMHO we should be sanitising input before it hits the database. What do you think @Matt-Yorkley?

Matt-Yorkley: @oeoeaio The pre-save sanitization remains unchanged, so it's using whatever we're using for all the other textarea fields throughout the admin area. I havn't properly looked into how that's currently done, but I assume there's some basic sanitizing, escaping, etc.

My thinking with this was to leave in the original **description** attribute unchanged, which just uses `strip_tags`. This can still be used to render plain-text descriptions without the HTML tags or styling, which could be useful in other places (like an RSS feed, or a preview snippet of the first 10 words of the description). 

Since the output of the new **description_html** attribute will be rendered directly as html, it needs to be marked with `html_safe` before it gets to the view, so I thought I'd make sure it was definitely safe before doing so.

From my limited knowledge of web security: theoretically if an SQL injection vulnerability was found someone could inject malicious data into the database and pre-save sanitation wouldn't necessarily help. My understanding of the way Rails handles data means that further problems would be fairly limited, unless it happened to be in a field that gets marked with `html_safe` before hitting the browser, in which case very bad things could happen with XSS...

That was my thinking for the use of `html_safe` preceded by very explicit sanitization.

oeoeaio: Ok, awesome. Thanks for that explanation. Totally agree that we need sanitisation post-DB, no arguments there.

My understanding is that the only pre-save sanitisation happening is coming from textAngular, there is no server-side sanitisation whatsoever. This means that a malicious user could theoretically construct a POST request with an unsanitised product description which would be saved straight to the DB. If we had a pre-save callback on the model that sanitised the description on save, that would give me more peace of mind. What do you think?

What I am particularly concerned about is another developer who has not been a part of this conversation taking either allowing a user to set **description_html** value for a product without textAngular (meaning it may not be sanitised), or taking the value in the DB, assuming it is sanitised and calling html_safe on it, when in fact as it currently stands we cannot be sure of that...

Matt-Yorkley: @oeoeaio I agree in principle, but from the standpoint of malicious POST requests there isn't really any difference between this field and every other field in the whole app (at least all the textarea fields), whether they have textAngular or not. Last time I looked I couldn't see any pre-save validation for any fields in any of the models, so I sort of feel like if this field isn't secure enough then everything needs a security review. Or am I missing something?

daniellemoorhead:
Hey there, couldn't put this through in 1.8.8 because of the conflicts that need resolving. Perhaps rebase and solve conflict then put it through again?

Matt-Yorkley:
@daniellemoorhead the merge conflict is resolved now.

daniellemoorhead:
@oeoeaio can you please give the ok and put this on staging for @sstead to give the OK for 1.8.9 release next week.

oeoeaio:
@Matt-Yorkley I'm sorry. I reviewed this is Jan but forgot to submit the review. I'll make the changes that I think need to be made, and resubmit.

app/overrides/spree/admin/products/_form/add_description_wysiwyg.html.haml.deface
```diff
@@ -0,0 +1,3 @@
+/ replace "[data-hook=admin_product_form_left] code[erb-loud]:contains('f.text_area :description')"
+%text-angular{'id' => 'product_description', 'ng-model' => 'product.description', 'name' => 'product[description]', 'class' => 'text-angular', 'ta-toolbar' => "[['bold','italics','clear']]"}
+  != @product[:description].to_s.html_safe
```
Matt-Yorkley: Ah! That's not good... looking at it again I think the sanitization I added was sufficient, but I only added it to the serializer and forgot to add it to the model, which is what's being referenced here.

Matt-Yorkley:
@sstead Sure, I'll take a look at getting it fixed. Seems like a weird textAngular copy/paste thing, I'll see if there's some settings for how pasted content is parsed.

Matt-Yorkley:
@sstead Just fixing this up now... does this apply to the various other textAngular fields in the backend?

Matt-Yorkley:
@sstead Wow, that was diabolical! It should be fixed now. I've made a re-usable solution in case we need it on other fields in the future. 

I decided to strip _all_ formatting on pasted content, as well as those weird html comments, as there were other tags that were not playing well when being saved, and pasting in formatted content seemed to be bypassing textAngular's sanitizing. So this means if a user has pastes a large chunk of text from a website they might have to re-add paragraphs or redo some of the formatting a bit.

oeoeaio:
Thanks @Matt-Yorkley! Please note I've rebased this just now.

Issue #1232, opened 2016-11-24
Include Variant Name in Product Search


lin-d-hop:
Product Variants can have names that hold key information about the product. As an example:
Product: Homemade Icecream
Variant: Chocolate
Variant: Strawberry
Variant: Vanilla
Variant: Chocolate chip

In the current implementation if a customer inputs 'Chocolate' they would not see the Chocolate Icecream. And really they should.

As a customer, I search for products 
If the search term is contained in a product name/producer/taxon
- the product should display with all variants

As a customer, I search for products 
If the search term is contained in a variant name, but not the product name/producer/taxon
- the specific variant (within the product) should display.
- the variants that don't match should not display

1) Search Icecream
Display:
Product: Homemade Icecream
Variant: Chocolate
Variant: Strawberry
Variant: Vanilla
Variant: Chocolate chip

2) Search Chocolate
Product: Homemade Icecream
Variant: Chocolate
Variant: Chocolate chip

We'd like the list of filtered products to contain the results searching upon Product Name AND Producer AND Taxon AND Variant Name.

Edit: Changed the quiche flavours into icecream flavours... hunger does not result in clear issue specification when it comes to food products :-)

OliverUK:
I certainly wouldn't be fussed if a search for "goats cheese" returned all the quiches in this example though of course your suggestions is much more elegant, Lynne. 

Also I think there are some better examples why this is really important. For example I have a product called "speciality cheeses" shown below. A search for "feta", "mozarella, "brie" or "recotta" wouldn't return anything.

![screen shot 2016-11-25 at 12 33 29](https://cloud.githubusercontent.com/assets/17590921/20625464/78e69f96-b30b-11e6-8512-5ea29753b7fa.png)

Equally my listing for "beef" which has as variants all sorts of steaks and other cuts wouldn't come up if you searched for "steak", "sirloin" or "topside".
![screen shot 2016-11-25 at 12 36 10](https://cloud.githubusercontent.com/assets/17590921/20625552/f5a869f6-b30b-11e6-83e2-54bfce8ed8b7.png)

I feel these illustrate a bit better than the quiche that this can make a huge difference.

lin-d-hop:
@OliverUK Thanks for the additional examples.
[...]

PR #1241, opened 2016-11-28
Product search indexes variant names

Matt-Yorkley:
Product searches in the frontend also index variant names, addressing issue #1232.

Suggestions for a more elegant way of writing this are welcome! 

I couldn't get the product filter to match properties inside nested objects (variants are nested inside each product), so I created a new (temporary) variant_names property on the product objects as the matches are being filtered and passed any additional variant names into a concatenated string so it could be indexed by the filter.

### DIFF
--- a/app/assets/javascripts/darkswarm/services/products.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/products.js.coffee
@@ -24,6 +24,10 @@ Darkswarm.factory 'Products', ($resource, Enterprises, Dereferencer, Taxons, Pro
         if product.variants?.length > 0
           prices = (v.price for v in product.variants)
           product.price = Math.min.apply(null, prices)
+          if product.variants.length > 1
+            angular.forEach product.variants, (item, key) ->
+              if product.name != item.name_to_display
+                product.variant_names += item.name_to_display
         product.hasVariants = product.variants?.length > 0
 
         product.primaryImage = product.images[0]?.small_url if product.images
--- a/app/assets/javascripts/darkswarm/services/products.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/products.js.coffee
@@ -19,20 +19,11 @@ Darkswarm.factory 'Products', ($resource, Enterprises, Dereferencer, Taxons, Pro
         @registerVariants()
         @loading = false
 
-    indexVariants: (product) ->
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/darkswarm/services/products.js.coffee
```diff
@@ -24,6 +24,10 @@ Darkswarm.factory 'Products', ($resource, Enterprises, Dereferencer, Taxons, Pro
         if product.variants?.length > 0
           prices = (v.price for v in product.variants)
           product.price = Math.min.apply(null, prices)
+          if product.variants.length > 1
```
levent: I'm not good with js but maybe this can be factored out into a separate method to reduce the if-nesting levels.

levent:
Hi @Matt-Yorkley, I might be missing something but the specs seem to pass on my computer even when I remove the code changes.

Matt-Yorkley:
@levant the test should fail in that instance. Testing as a human with the code as it was previously, when the search term doesn't match product names (but does match a variant display name) the search shows no results. I wonder if rspec/capybara is waiting for angular to filter the results or not?

levant:
@Matt-Yorkley you addressed your comment to the wrong person, me instead of @levent

Matt-Yorkley:
Okay, I've expanded the shopping spec; firstly to test that the functionality of the basic product search filtering is working correctly, and secondly to make the variant name search tests more comprehensive.

Matt-Yorkley:
I've refactored the javascript a bit, as suggested. It still requires the if statements, but now it's in a separate function and looks a bit cleaner. I would have made it slightly more concise, but I've just discovered that coffeescript doesn't allow inline ternary operators...

app/assets/javascripts/darkswarm/services/products.js.coffee
rhodri: alternatively
```
product.variant_names = product.variants?.map(function (v) { return v.name_to_display }).join(' ');
```
which can even be done in `filter_products.js.coffee` to avoid touching more than one file

lin-d-hop:
Hi @levent  Are you happy in your review of this? Is it ready for testing?

levent:
@lin-d-hop @Matt-Yorkley yeah nice work! 👍

Issue #1186, opened 2016-10-12
Change from 3 Letter abbriviation for Enterprise registration county


lin-d-hop:
When enterprises sign up a three letter acronym is displayed when asked for their address.
![ofnenterprisesignup1](https://cloud.githubusercontent.com/assets/8669848/19313970/44edd670-9090-11e6-8de7-53705440b99e.png)

It would be great if the whole name could be displayed, just as it is when a user checks out
![ofnshopfront2](https://cloud.githubusercontent.com/assets/8669848/19314076/b26e5710-9090-11e6-9fed-9c55f4c80dc0.png)

myriamboure:
See case here @lin-d-hop , I think that are overlaps: http://community.openfoodnetwork.org/t/states-management-internationalisation-issues/. If whole name is displayed, we won't have any more problem I think :-)

stveep:
Here's a few thoughts on how to get started on this issue: 

It's probably just a case of figuring out where the checkout page gets its data from, and changing the signup page to use the same source. Here's where I would start

Page with the abbreviated states (i.e. counties for UK):
`app/assets/javascripts/templates/registration/details.html.haml`

This new enterprise signup flow is all in Angular, so the relevant files are in `app/assets/javascripts/controllers`, `/services` etc. The corresponding rails view app/views/registration basically just renders the basic layout and hands over to the angular controller.

Page with the correct (full names) states:
`app/views/checkout/_shipping_ship_address.html.haml`

This is more of a rails view, but still uses some Angular (e,g ShippingCtrl angular controller) - so there may also be some relevant info in the corresponding `app/assets/javascripts` directory.

Tests might be a bit patchy by the look of it so it would be great if you could update these while you're at it.
`spec/javascripts/unit/darkswarm/services/enterprise_registration_spec.js.coffee`
`spec/features/consumer/registration_spec.rb`

I can see that the tests hard code the Australian three-letter abbreviations, which is presumably what they want. So we do need to make sure that we don't break that, and probably leave it as the default behaviour. Perhaps we could add a way to set whether to use abbreviated or full names in the locale configuration.

KeirOsborn:
Hi, 

Awesome, thanks Steve.  

[...]

PR #1242, opened 2016-11-29
Registration state display fix

KeirOsborn:
Fix for issue #1186.

The key change is in app/assets/javascripts/templates/registration/details.html.haml and appears on /register on the front end. Instead of pulling through a state's abbreviation by default, the form now pulls through a state's label. Label is a new method, see app/serializers/api/state_serializer.rb that checks the localization files for the preferred way to display states.

Also made a small CSS change to accommodate the fact that state names are bigger than abbreviations.


stveep:
Hi @KeirOsborn thanks for working on this! It's a clever way to do it but my concern is that it's not a very intuitive way for instance admins to set up the behaviour. I think a better way would be to have a configuration option that can be used to decide what to serlialise as `label`. This could be set by super admin on the admin > configuration > counties/states page. That's a spree view so you'd need to add a deface override targeting this page:

https://github.com/spree/spree/blob/1-3-stable/core/app/views/spree/admin/states/edit.html.erb

Alternatively the option could be set in application.yml - other locale settings are configured there and this is probably not the kind of thing that will need to be changed regularly. Regardless of how you do it, could you also write a spec to cover the new functionality?

Also, I don't think it's necessary to add translations to all locale.yml files - you can do en (and en-GB if you want to test things), then the rest are generated through Transifex....I think.... @mkllnk ?

lin-d-hop:
@stveep You are right about translations, they are handled independently by instances in Transifex. en.yml is the only one that's actually necessary to add to as Transifex uses this as the source for all other translations.

mkllnk:
Yes, I can confirm that `en.yml` is enough.

> I can see that the tests hard code the Australian three-letter abbreviations, which is presumably what they want.

I will check that with the Australian team. It might have just been easier. So if we are okay to switch from `abbr` to `name` then we are all on the same page and the code would be much easier, right?

KeirOsborn:
Hi all,

Ok great sounds good. Thanks for the feedback. Steve - yep agree, I hadn't
really thought of this as something that admins might want to edit but it
makes sense. And yep, will write a spec.

Thanks for the heads up about translations, will remember for the future.

Cool, see some of you tomorrow.

Keir

mkllnk:
Okay, I talked with Rob and we don't see a problem in displaying the full state name. It would still be nice, if there is an option to customise it on an instance level. But in that case we could change the state names in the database. So you can change the test case to select "Victoria" instead of "VIC". And we will have a look on our staging server before it goes through.

lin-d-hop:
Hi @stveep 
I didn't quite follow the above sorry. Is this ready for testing?
Cheers :-)

KeirOsborn:
Hi all,

I've just updated this with a load of changes. I've opted to make how the state is displayed on /register configurable in the admin area. I've added a form field to /admin/general_settings/edit where admins can select how they want states to be displayed. It hooks into Spree's AppConfiguration class. I know this looks like quite an over-engineered solution for such a small issue but it does seem like the most logical way to do it.

lin-d-hop:
@KeirOsborn would you mind moving this to a uk/registration-state-display-fix branch so that it creates an automatic buildkite build? Thanks!

config/locales/en.yml
```diff
@@ -851,6 +851,7 @@ Please follow the instructions there to make your enterprise visible on the Open
   registration_detail_state_error: "State required"
   registration_detail_country: "Country:"
   registration_detail_country_error: "Please select a country"
+  registration_detail_state_label: "abbr"
```
Matt-Yorkley: @KeirOsborn This `registration_detail_state_label` key doesn't seem to be used anywhere now...? Was it for a translation that got removed?

Gemfile.lock
```diff
@@ -732,9 +732,3 @@ DEPENDENCIES
   whenever
   wicked_pdf
   wkhtmltopdf-binary
-
-RUBY VERSION
-   ruby 2.1.5p273
-
-BUNDLED WITH
-   1.12.5
```
Matt-Yorkley: Was there a reason for modifying `Gemfile.lock`? I think it's one of those files that doesn't usually get touched.

KeirOsborn:
Hi Matt,

Yep that is indeed redundant now. Well spotted.

Thanks,

KeirOsborn:
Re Gemfile.lock. No this seemed to happen automatically as part of the PR.
Don't know how to stop it. If you know what I'm doing wrong to make it
happen or what I should do to stop it let me know.

Thanks,

mkllnk:
Nice work. It is indeed a bit over-engineered. But it might be important in some countries to use the abbreviations. Would be great if you can work on the feedback you got. It would also be nice to rebase and squash your branch.

app/serializers/api/state_serializer.rb
```diff
@@ -1,7 +1,12 @@
 class Api::StateSerializer < ActiveModel::Serializer
-  attributes :id, :name, :abbr
+  attributes :id, :name, :abbr, :label
 
   def abbr
     object.abbr.upcase
   end
+
+  def label 
+	eval Spree::Config[:state_display]
```
sauloperez: this should be `public_send` instead. That is the way to execute dynamic method calls in ruby.

lin-d-hop:
UK are very happy **not** to use abbreviations. For us they are confusing as well.

CynthiaReynolds:
We are also very happy **NOT** to be using abbreviations in Norway ;)

myriamboure:
In France we are also happy NOT to use abbreviations but display states names instead :-)

enricostano:
I opened a PR with a proposal to solve this https://github.com/openfoodfoundation/openfoodnetwork/pull/1813

One question though: do we want to make the same change in other parts of the app? In admin perhaps?

sauloperez:
> One question though: do we want to make the same change in other parts of the app? In admin perhaps?

I'd rather be consistent, but show me the code first :trollface:

daniellemoorhead:
@enricostano can we close this PR now that you've created a new one?

mkllnk:
Closing in favour of #1813.

enricostano:
@daniellemoorhead sorry, I didn't see this notification 🙏

PR #1243, opened 2016-11-29
Translation updates

lin-d-hop:
Translation updates. Deployed on UK server.
Please add these updates to the next release. Thanks!


lin-d-hop:
As this is only translation updates I'm going to merge it.
Despite @mkllnk saying this was ok it still feels wrong.... :-/

PR #1245, opened 2016-11-29
Standing orders

oeoeaio:
First go at something to test for Standing Orders


app/models/schedule.rb
```diff
@@ -0,0 +1,14 @@
+class Schedule < ActiveRecord::Base
+  has_and_belongs_to_many :order_cycles, join_table: 'order_cycle_schedules'
+  has_many :coordinators, uniq: true, through: :order_cycles
+
+  attr_accessible :name, :order_cycle_ids
+
+  validates :order_cycles, presence: true
+
+  scope :with_coordinator, lambda { |enterprise| joins(:order_cycles).where('coordinator_id = ?', enterprise.id).select('DISTINCT schedules.*') }
```
RohanM: Might be a little clearer if we specify `order_cycles.coordinator_id` here? Not that it's an ambiguous column, but it took me a little while to figure out where the column was coming from.

app/models/schedule.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def current_or_next_order_cycle
+    order_cycles.where('orders_close_at > (?)', Time.now).order('orders_close_at ASC').first
```
RohanM: I think we probably don't need brackets around the `?` unless using `IN`.

app/models/spree/line_item_decorator.rb
```diff
@@ -4,6 +4,9 @@
   include OpenFoodNetwork::VariantAndLineItemNaming
   has_and_belongs_to_many :option_values, join_table: 'spree_option_values_line_items', class_name: 'Spree::OptionValue'
 
+  # Redefining here to add the inverse_of option
+  belongs_to :order, :class_name => "Spree::Order", inverse_of: :line_items
```
RohanM: I think it'd be nicer to have consistent hash syntax here (ie. `class_name: 'Spree::Order'`).

app/models/spree/order_decorator.rb
```diff
@@ -234,11 +236,17 @@ def account_invoice?
...
-    unless account_invoice?
+    unless account_invoice? || standing_order.present?
       Delayed::Job.enqueue ConfirmOrderJob.new(id)
     end
   end
 
+  # Override or Spree method. Used to prevent payments on standing orders from being processed in the normal way.
+  # ie. they are 'hidden' from processing logic until after the order cycle has closed.
+  def pending_payments
+    return [] if standing_order.present? && order_cycle.orders_close_at.andand > Time.now
```
RohanM: Nifty use of `.andand >`!

app/models/spree/product_decorator.rb
```diff
@@ -115,6 +115,13 @@
     end
   }
 
+  scope :stockable_by, lambda { |enterprise|
+    return where('1=0') unless enterprise.present?
```
RohanM: ActiveRecord in Rails 4 has `.none` that's equivalent to `where('1=0')`. Not yet, but soon...

app/models/standing_line_item.rb
```diff
@@ -0,0 +1,16 @@
...
+  belongs_to :standing_order, inverse_of: :standing_line_items
+  belongs_to :variant, class_name: 'Spree::Variant'
+
+  validates :standing_order, presence: true
+  validates :variant, presence: true
+  validates :quantity, { presence: true, numericality: { only_integer: true } }
+
+  def available_from?(shop, schedule)
+    Spree::Variant.joins(exchanges: { order_cycle: :schedules})
+    .where(id: variant_id, schedules: { id: schedule}, exchanges: { incoming: false, receiver_id: shop })
```
RohanM: Spacing around hashes look a bit unbalanced on these lines. We should probably decide on some conventions for syntax in general and set up Rubocop as part of the build pipeline to enforce it.

app/models/standing_order.rb
```diff
@@ -0,0 +1,82 @@
...
+  def cancel
+    transaction do
+      self.update_column(:canceled_at, Time.zone.now)
+      standing_order_orders.each(&:cancel)
+      true
+    end
+  end
+
+  def state
+    return 'canceled' if canceled_at
```
RohanM: The many exit points from this method make it read nicely, but is considered bad practice because it can make it harder to debug. With a single exit point, you can pry before that point and check the return value etc. But with 4 return statements, you'd need to pry at the top and then manually figure out what was going on.

app/assets/javascripts/admin/index_utils/index_utils.js.coffee
```diff
@@ -1 +1,2 @@
-angular.module("admin.indexUtils", ['admin.resources', 'ngSanitize', 'templates', 'admin.utils']).config ($httpProvider) ->  $httpProvider.defaults.headers.common["X-CSRF-Token"] = $("meta[name=csrf-token]").attr("content");  $httpProvider.defaults.headers.common["Accept"] = "application/json, text/javascript, */*";
\ No newline at end of file
+angular.module("admin.indexUtils", ['admin.resources', 'ngSanitize', 'templates', 'admin.utils']).config ($httpProvider) ->
+  $httpProvider.defaults.headers.common["X-CSRF-Token"] = $("meta[name=csrf-token]").attr("content");  $httpProvider.defaults.headers.common["Accept"] = "application/json, text/javascript, */*";
```
enricostano: is this `^M` escape character in the right place?

app/controllers/admin/order_cycles_controller.rb
```diff
@@ -76,9 +80,14 @@ def update
     def bulk_update
       @order_cycle_set = params[:order_cycle_set] && OrderCycleSet.new(params[:order_cycle_set])
       if @order_cycle_set.andand.save
-        redirect_to main_app.admin_order_cycles_path, :notice => 'Order cycles have been updated.'
+        respond_to do |format|
+          order_cycles = OrderCycle.where(id: params[:order_cycle_set][:collection_attributes].map{ |k,v| v[:id] })
```
enricostano: Maybe you can separate this a bit:
```ruby
def bulk_update
  @order_cycle_set = params[:order_cycle_set] && OrderCycleSet.new(params[:order_cycle_set])
  if @order_cycle_set.andand.save
    respond_to do |format|
      order_cycles = OrderCycle.where(id: order_cycles_ids)
      format.json { render_as_json order_cycles, ams_prefix: 'index', current_user: spree_current_user }
    end
  else
    # ...
  end
end

private

# Now you have more space to describe what that `key` and `value` are
#
# @return [Array<Integer>]
def order_cycle_ids
  params[:order_cycle_set][:collection_attributes].map do |_key, value|
    value[:id]
  end
end
```

lib/open_food_network/address_finder.rb
```diff
@@ -74,9 +89,11 @@ def last_used_ship_address
...
     # a customer or user other than themselves...
     def allow_search_by_email?
-      return false unless email.present? && (user.present? || customer.present?)
-      return false unless email == customer.andand.email || email == user.andand.email
-      true
+      email.present? && email_matches_customer_or_user?
+    end
+
+    def email_matches_customer_or_user?
+      email == customer.andand.email || email == user.andand.email
```
sauloperez: is it really possible for `customer` and `user` records to have an empty email? I would expect them to have them mandatory even at DB level with a `NOT NULL` constraint.

This code shouldn't be concerned about data consistency. This should be provided elsewhere in the code.

lib/open_food_network/address_finder.rb
```diff
@@ -3,20 +3,15 @@
...
       args.each do |arg|
-        case arg
-        when String
-          @email = arg unless @email
-        when Customer
-          @customer = arg unless @customer
-        when Spree::User
-          @user = arg unless @user
-        end
+        type = types[arg.class]
```
sauloperez: It would good to doc this method so that we know what `args` can be.

app/controllers/admin/customers_controller.rb
```diff
@@ -53,6 +55,22 @@ def destroy
...
+      render json: { bill_address: bill_address, ship_address: ship_address }
+    end
+
+    # GET /admin/customers/:id/cards
+    # Used by standing orders form to load details for selected customer
+    def cards
+      cards = Spree::CreditCard.where(user_id: @customer.user_id)
+      render json: ActiveModel::ArraySerializer.new(cards, each_serializer: Api::CreditCardSerializer)
+    end
+
```
sauloperez: have you considered moving these to a separate controller or simply having standing orders as a new REST resource? from my experience, when we diverge from REST in the controllers, things go wild pretty soon. Mainly because the class drops its cohesion.

lib/open_food_network/address_finder.rb
oeoeaio: It's not testing whether they have an empty email. Its testing whether `customer` or `user` are present at all in the local context.

`customer.andand.email` is shorthand for `customer && customer.email`

app/controllers/admin/customers_controller.rb
oeoeaio: Happy to put these two actions in their own controllers. The reason they are here is that they don't really belong in the `AddressesController` or `CreditCardsController`, and I was trying to avoid creating a `CustomerAddressesController` and a `CustomerCreditCardsController`.

sauloperez: Does make sense then, a `StandingOrdersController`? If so, I would just create an issue for it as it would lead to a quite a big refactor.

sauloperez:
You should rebase against master to get the latest Code Climate configuration.

app/controllers/admin/customers_controller.rb
oeoeaio: My [StandingOrdersController](https://github.com/oeoeaio/openfoodnetwork/blob/standing-orders/app/controllers/admin/standing_orders_controller.rb) is already pretty full, and IMHO these actions belong there much less than they belong here in `CustomersController`.

lib/open_food_network/address_finder.rb
sauloperez: Oops, I see.

app/controllers/admin/customers_controller.rb
sauloperez: Ok, I thought it didn't exist yet...

db/migrate/20161210235939_add_order_cycle_id_to_proxy_orders.rb
```diff
@@ -0,0 +1,20 @@
+class AddOrderCycleIdToProxyOrders < ActiveRecord::Migration
+  def up
+    add_column :proxy_orders, :order_cycle_id, :integer
+
+    ProxyOrder.all.each do |proxy_order|
```
sauloperez: please, use [find_each](https://apidock.com/rails/v4.2.7/ActiveRecord/Batches/find_each) instead of `all`. It can have pretty bad consequences.

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -118,7 +118,10 @@ def visible_variants_for_outgoing_exchanges_to(hub)
       return Spree::Variant.where("1=0") unless @order_cycle
 
       if user_manages_coordinator_or(hub)
+        # TODO: Use variants_stockable_by(hub) for this?
+
         # Any variants produced by the coordinator, for outgoing exchanges with itself
+        # TODO: isn't this completely redundant given the assignment of hub_variants below?
```
sauloperez: Can we move these two to issues? here they are not actionable.

lib/open_food_network/proxy_order_syncer.rb
```diff
@@ -0,0 +1,79 @@
...
+      uninitialised_order_cycle_ids.each do |order_cycle_id|
+        proxy_orders << ProxyOrder.new(standing_order: standing_order, order_cycle_id: order_cycle_id)
+      end
+    end
+
+    def create_proxy_orders!
+      return unless not_closed_in_range_order_cycles.any?
+      query = "INSERT INTO proxy_orders (standing_order_id, order_cycle_id, updated_at, created_at)"
+      query << " VALUES #{insert_values}"
+      query << " ON CONFLICT DO NOTHING"
```
sauloperez: why do we need this?

oeoeaio: I wrote this to give a performance improvement. Depending on the data it can save on two or three queries per standing order. From memory it dropped the execution time for 150 StandingOrders x 10 ProxyOrders from 6~7 seconds down to about 0.5 seconds. I'll make a note to have another look at it and see if I think of another way of using ActiveRecord.

db/migrate/20161210235939_add_order_cycle_id_to_proxy_orders.rb
oeoeaio: 👍

lib/open_food_network/proxy_order_syncer.rb
sauloperez: More than that I was wondering about that "ON CONFLICT DO NOTHING"

app/views/admin/enterprises/form/_shop_preferences.html.haml
```diff
@@ -75,3 +76,17 @@
...
+-#     .three.columns.alpha
+-#       %label= t '.enable_standing_orders'
+-#       %div{'ofn-with-tip' => t('.enable_standing_orders_tip')}
+-#         %a= t 'admin.whats_this'
+-#     .three.columns
+-#       = f.radio_button :enable_standing_orders, true
+-#       = f.label :enable_standing_orders, t('.enable_standing_orders_true'), value: :true
+-#     .five.columns.omega
+-#       = f.radio_button :enable_standing_orders, false
+-#       = f.label :enable_standing_orders, t('.enable_standing_orders_false'), value: :false
```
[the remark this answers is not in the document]
oeoeaio: This is just a poor man's feature toggle, without this code users cannot turn standing orders functionality on for themselves. I'll uncomment this in a week or so once we have had the code sitting in production for a while.

lib/open_food_network/proxy_order_syncer.rb
oeoeaio: Each standing order has_many `ProxyOrders` which are just empty placeholders for actual `Order`s. When an `OrderCycle` is added to or removed from a Schedule, the list of ProxyOrders for each `StandingOrder` in the `Schedule` needs to be updated. List gives use a quick way of creating any new `ProxyOrders` that have not been initialised, while leaving existing ones without overwriting them.

oeoeaio:
Closing in favour of #2068

Issue #1031, opened 2016-06-08
basket doesn't save on logout


NickWeir63:
Stroudco shoppers want to be able to leave items in their basket that are not lost when they log out so they can add items during the week and then review and confirm them just before the OC closes

lin-d-hop:
@stveep Would you be willing to look into this issue a little more? Perhaps to scope it out a little more for @lebreeze.

lin-d-hop:
A little more specification from a non-technical perspective:

As a shopper, when I am logged in I would like my cart to persist for the hub I am shopping with. I would like OFN to remember the last cart that I created such then if I log out and log back in again the last cart is persisted.

If I change enterprises, order cycles or the order cycle closes my cart should not persist.

Currently if a user is shopping with one enterprise, then they switch to another enterprise their cart is cleared from the scope. This behaviour is enables multiple shopfronts to exist simultaneously. However the majority of shoppers only shop from a single hub. Therefore there is a clear benefit for a customers cart to persist if they log out and log back in to shop at the same hub.

The spree_order model enables carts to be persisted to a user and spree_line_items are also persisted. So to my understanding this issue is pertaining to reloading the cart to the scope when a user logs back in, with some checks to ensure the order cycle has not closed.

I think this issue will need a couple hours of scoping first to fully understand the implications. @stveep Do you think this might be suitable for @lebreeze to ldelve further into?

@oeoeaio @mkllnk If this triggers any alarm bells to you would you mind noting them here too?

mkllnk:
I think that would be great. There was once a multicart feature before my time. I'm not sure if that actually worked. I definitely find the current cart system confusing and brittle. But it might be a bit difficult to work around Spree's model of only one cart.

oeoeaio:
This would be different to a multi-cart, right? We would simply be tweaking the `current_order` method in `lib/spree/core/controller_helpers/order_decorator.rb` to make it search the database for a relevant incomplete order, in the event that there is no `order_id` in the session.

If this search was restricted to incomplete orders owned by the current user, for the current shop and in the current order cycle, I can't see a problem with it.

kirstenalarsen:
the multi-cart never worked. it was a folly, that was abandoned ass the complexity became too much :)

lin-d-hop:
While multi-cart would be amazing, this enhancement falls a long way short. Thanks for the clear spec @oeoeaio.

Tagging @levent as I tagged him incorrectly above.

[...]

PR #1248, opened 2016-12-01
1031 Restore incomplete cart items on login if cart currently empty

levent:
Addresses #1031 

Expected behaviour:

```
A) If user already has items in their basket on log in,
     do not modify the basket.
B) If user does **not** have items in their basket on login,
     use the old basket if it's the same distributor and order cycle
```


.gitignore
```diff
@@ -40,3 +40,4 @@ coverage
 libpeerconnection.log
 /config/application.yml
 node_modules
+/vendor/bundle
```
stveep: I guess this is not going to cause problems - but does it just to apply to how your dev environment is set up? For me at least I  have a `/vendor` dir but not a `bundle` subdir in the repo. If it's just for you maybe this can go in `.git/info/exclude`, or a separate .gitignore in /vendor to solve your specific git issue.

levent: Yeah it's a fairly standard environment setup convention (see https://github.com/github/gitignore/blob/master/Rails.gitignore#L27) but nothing to do with this functionality so happy to pull it out, if you like?

lin-d-hop:
@stveep Happy with this?

stveep:
Yes I think it's fine.

daniellemoorhead:
---Doing a PR Cleanup---

@lin-d-hop @stveep where are we at with this? Someone taking over from Levant with this?

Issue #1211, opened 2016-11-11
Enterprise Registration - Agree to Terms of Service


lin-d-hop:
Different countries have been developing Terms of Service, but unless we provide a place in which we ask people to our Terms of Service we are going to have trouble enforcing them.

I would like to see an optional 7th step added to the Enterprise Registration process (see image below). It would contain fully translatable text fields. Users will need to check a Checkbox to enable the Finish button that would complete their registration.
The Terms of Service link would link to the file saved in the field:
`ContentConfig.footer_tos_url` 

Some countries will begin using OFN before their terms of service is complete. And different countries have  different legalities around the presentation of Terms of Service. As such this feature should be configurable through the super-admin. This could be done via a checkbox beside the field in the Super-admin UI:
"TERMS OF SERVICE URL" -> Super-admin -> Configuration -> Content (the last field on the page).
Checkbox: "Enable ToS Agreement in Enterprise Registration"

@daniellemoorhead @oeoeaio Any comments on this (fully configurable) enhancement?

![tosagree](https://cloud.githubusercontent.com/assets/8669848/20216590/646e1c9a-a813-11e6-9cb7-104d5597168d.jpg)

oeoeaio:
Sounds good to me!

myriamboure:
That's awesome, definitely needed :-)

NickWeir63:
yes I like this idea.  the UK finalised (for now) [terms of service are here](https://drive.google.com/file/d/0B8GhJcMDO6a1dEI5N3RuS2E2Ym8/view?usp=sharing)

levent:
I will pick this up. @lin-d-hop I can't assign myself or drag cards on the ZenHub board it seems.

lin-d-hop:
@levent... I still can't seem to assign to you. Are you able to do that?

levent:
@lin-d-hop nope. I cannot change anything in the right column (assignees, milestones, labels etc.) or in ZenHub (permissions probably link) and I still don't seem to appear in auto-complete.

levent:
Just a couple of observations / queries:

CSS gurus: As the registration flow breadcrumbs number 7 now they won't fit into the 12 column grid system. Any strict preferences on how you want me to implement this?

@lin-d-hop should this feature be enabled by default?

oeoeaio:
[...]

PR #1249, opened 2016-12-01
Enterprise Terms of Service for #1211

Matt-Yorkley:
Addressing issue #1211.

Added config options for Enterprise Terms of Service, defaulting to disabled (not required).

Reworked frontend registration form to:
- Show terms of service message and checkbox if `Spree::Config[:enterprises_require_tos]` is enabled.
- Hide if not.
- Disable "Lets get started" button if `Spree::Config[:enterprises_require_tos]` is enabled and accept checkbox is not checked. Re-enables when checked.

Added translation entries in en.yml and en-GB.yml.

Default (not required):
screenshot from 2016-12-01 18-21-15

Required, unchecked. Buton is disabled.
screenshot from 2016-12-01 17-52-28

Required, checked.
screenshot from 2016-12-01 17-52-31

Admin config:
screenshot from 2016-12-01 17-57-21


spec/features/consumer/registration_spec.rb
```diff
@@ -94,6 +97,43 @@
...
+
+    context "if accepting Terms of Service is required" do
+      before { Spree::Config.enterprises_require_tos = true }
+
+      it "does not allow registration unless checkbox is checked" do
+        visit registration_path
+
+        page.should have_content('Terms of Service')
+        page.should have_selector("input.button.primary[disabled]")
+
```
Matt-Yorkley: @oeoeaio @RohanM I'm having serious trouble with Spree::Config values having unpredictable values in tests... 

Before the test runs, `Spree::Config.enterprises_require_tos` is set to true. The page displays different content depending on this setting, determined by the view which renders an enclosing element with the attribute `ng-init = "tos_required=#{Spree::Config.enterprises_require_tos}" ` and hides elements based on this value.

The tests here fail though, with strange results, as if the config value has not been changed. After `visit_registration path:`
 If I add in `expect(Spree::Config.enterprises_require_tos).to eq true` it passes.
 If I add in `html.should include('tos_required=true')` it fails.
 If I add in `html.should_not include('tos_required=false')` it fails.

So it's confirming that when the test runs and after the page is loaded, `Spree::Config.enterprises_require_tos` is set to true, but `#{Spree::Config.enterprises_require_tos}` in being rendered as false in the DOM at the same time..?

This behaviour is only present in tests, not in the dev environment. Also, the results seem to change unexpectedly, sometimes passing and sometimes failing. There are 3 tests in this file where the content is altered by the config setting. Sometimes the value seems to be false in all tests, sometimes true in all tests, and sometimes set to false or true as expected (it's explicitly set before each test as true or false). For added weirdness: changing the value in the browser in the dev environment seems to effect the outcome in the tests...?

Am I going mad or missing something really obvious?

Matt-Yorkley: @stveep any ideas?

stveep: Nope, I'm seeing similar weirdness...it failed for me, I fiddled around a bit with different ways of setting the config and then the test above passed...then I rolled back all my changes and they still passed...then they (all) started failing again. I had situations where they would pass with zeus, pass individually with rspec but fail when the complete spec file was run. I'll try to have another look later this week.

Matt-Yorkley: I found a few references online to Spree::Config values responding strangely in tests due to the way they are cached: https://groups.google.com/forum/#!topic/spree-user/93HYDl3ZcPo

Seemed to offer an explanation for the issue, but the suggested fixes didn't help.

Matt-Yorkley: I tried installing the capybara-angular compatibility gem on my local version, but that didn't help either. Glad it's not just me!

oeoeaio:
No sure why this would help, but have you tried using `before { Spree::Config.set(: enterprises_require_tos, true) }`?

Also, given that both `html.should include('tos_required=true')` and `html.should include('tos_required=true')` fail, does that suggest that your ng-init isn't rendering properly at all?

oeoeaio:
Just looking at your code. I think I may know what the problem is, and I think there may be a better solution...

At the moment, we use a gem called `angular-rails-templates` to compile all files in `assets/javascripts/templates` as angular templates, which are then available to angular directives. As you know, this is where all of the registration pages are located. The problem is that we're building this wizard using templates that are compiled as javascript assets, and so in production, they will be unresponsive to changes in any Spree::Config variables until we restart the server.

I think that this may be happening in the feature spec you're having problems with too - I'm not super clear about whether the js is recompiled for every spec, but I would be surprised if it is. I think this may be why you are seeing something that looks like weird caching. The spec is `caching` the value by hard-coding it into the template before running the spec.

I've noticed that you don't actually need `Spree::Config. enterprises_require_tos` to be available to angular, because you're not doing anything dynamic with it, you're just showing/hiding/disabling/enabling elements based on a static variable, which is available on page load, so it could and probably should be done in haml.

We can solve this problem by moving the registration templates out of `assets/javascripts/templates` and instead rendering them as haml partials from `app/views/registration` , by wrapping them in script tags that mark them as angular templates: `<script type="text/ng-template" id="templateId.html">`. See docs [here](https://docs.angularjs.org/api/ng/service/$templateCache). Then you can use normal haml syntax to do whatever you need to do on page load. This has the added advantage of reducing the size of our compiled js on the whole front-end (yes, at the moment all registration templates are being loaded in the js on every single front-end page - not great).

I only recently discovered that this was possible, but I think it is much better than the pre-compiling for anything that is only used on one page (ie. most templates). Hopefully this is a relatively simple fix - the templates should behave in exactly the same way, but you get both of the benefits described.

See if that helps...

Matt-Yorkley:
Thanks Rob, I'll give it a go. 

It was html.should include('tos_required=true') and html.**should_not** include('tos_required=false') that were both failing.

stveep:
I tried using the Spree::Config.set syntax before and it didn't seem to make a difference.

Matt-Yorkley:
@stveep I moved the template file out to the views folder, which fixed the config issue. Now there's some sort of race condition issue with the second part of the test, where the test's actions are triggered before the angular bindings have been evaluated. The tests fail about 50% of the time and when they do, debugging the output shows things like `{{'registration_logo' | t}}` are still present in the DOM at the time the test is running.

This seems to be fixed either by adding `sleep 0.5` after the page loads and before capybara starts interacting with the page elements, or by installing the capybara-angular gem.

I notice without the capybara-angular gem the tests randomly either pass in 11 seconds or fail in 45 seconds, whereas with the gem installed they consistently pass in 15 seconds.

Any problems with adding a new gem to the project?

oeoeaio:
@Matt-Yorkley Can you push your changes up? I doubt there is a need for another gem, and you should almost never need to use sleep in a feature spec. Vanilla Capybara's have_selector and have_content methods implicitly wait for content to render on the page. If you are seeing a race condition, use `expect(page).to have_content blah` where `blah` is some content rendered by angular. This has worked for me on every other feature spec I've ever written.

Matt-Yorkley:
@oeoeaio The test in the above commit fails roughly 50% of the time.

oeoeaio:
Ok, wow. That is super weird. I'll have a crack at replicating and working out what is going on as soon as I can. :)

Matt-Yorkley:
Merged Rob's commits. @lin-d-hop this one's ready for testing.

lin-d-hop:
@Matt-Yorkley Would you just move this to a branch named uk/* so that it builds automatically for @OliverUK. Thanks!

PR #1322, opened 2016-12-09
Only build master branch and PRs

RohanM:
When we push a branch, Travis builds it. Then if we open a pull request for that branch, it builds it again.

Set Travis to build only the master branch and PRs. We should see the master branch build green every time, and a red build is in indication of an intermittent failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rvm:
 branches:
   only:
   - master
+  - merged-pull/*
 
 # Set the timezone for phantomjs with TZ
 # Set the timezone for karma with TIMEZONE
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ bundler_args: --without development
 rvm:
   - "2.1.5"
 
+branches:
+  only:
+  - master
+
 # Set the timezone for phantomjs with TZ
 # Set the timezone for karma with TIMEZONE
 #


.travis.yml
```diff
@@ -5,6 +5,10 @@ bundler_args: --without development
 rvm:
   - "2.1.5"
 
+branches:
+  only:
+  - master
```
mkllnk: We should add:
```yaml
  - merged-pull/*
```

Otherwise CI can't use Travis to check merged pull requests. Alternatively, we can deactivate auto-merging of pull requests and always merge/rebase manually.

mkllnk: One more idea: We give CI it's own repository to build stuff. That would also mean more resources and our main repository is free from `merged-pull/*` branches.

RohanM:
Added build for `merged-pull/*`.

mkllnk:
I just thought of another issue here. `.travis.yml` is shared between repositories. That would mean that our forks would not build any branches either. I can think of three options:

- We configure the branches to build in Travis environment variables and keep `.travis.yml` flexible. This is a bit complex for such an easy task. Maybe better to review the whole process.
- By convention, we just don't push any branches to the main repository.
- We deactivate building pushes in Travis, push the merged-pull branches to a separate repository and optionally build the master branch daily or weekly with Travis' new beta cron feature.

RohanM:
Good points, and it seems to me that the second option (we don't push branches to the main repo) looks like the most appealing. Closing this one for now.

PR #1327, opened 2016-12-09
Uk/ wysiwyg line breaks issue

Matt-Yorkley:
Changed the way text displays in the editor on the "About Us" tab in enterprise management to be visually consistent with the spacing of paragraphs on the front-end, to resolve issue #375.

Also solves line break issues with product descriptions in #1020 and any other sections that use the wysiwyg.

### DIFF
--- a/app/assets/stylesheets/admin/openfoodnetwork.css.scss
+++ b/app/assets/stylesheets/admin/openfoodnetwork.css.scss
@@ -203,6 +203,9 @@ text-angular {
   .ta-scroll-window > .ta-bind {
     max-height: 400px;
     min-height: 100px;
+    p {
+      margin-bottom: 1.5rem;
+    }
   }
   .ta-scroll-window.form-control {
     min-height: 100px;
--- a/app/assets/stylesheets/admin/openfoodnetwork.css.scss
+++ b/app/assets/stylesheets/admin/openfoodnetwork.css.scss
@@ -203,6 +203,7 @@ text-angular {
   .ta-scroll-window > .ta-bind {
     max-height: 400px;
     min-height: 100px;
+    outline: none;
     p {
       margin-bottom: 1.5rem;
     }


sstead:
About Us spacing looks good for new profiles.
image
I'm worried about this existing profile... The profile shows the text, but nothing shows in the editor... This is the only profile I can see the problem for. We don't want this to happen to profiles on Production- will confuse users.
image
The paragraphs display a bit strangely in About US if the text gets big enough to scroll...but this isn't a deal breaker.
image
The fix doesn't carry to product descriptions...
image
Existing profiles are behaving in different ways- most look fine, but it's hard to predict how they'll look on production under the old formatting rules. E.g. the one below is showing existing text that had separate lines in a list. I cannot type another list in this fashion, only existing text behaves like this.
image
Be aware that for profiles who put 2 blank lines in their profile such as below... their gap will remain, and will need to be fixed. This is fine, the alternative would be that everyone who only put one line would need to go through and add another - either way some will still have incorrect formatting.
image

sstead:
Hi @Matt-Yorkley , in regards to this profile... Rohan said it might be 'sanitising' and not showing the markup/HTML.
image
Rohan will send you the database for you to look at.

Matt-Yorkley:
Hi @sstead thanks for the feedback. I'll try to break these down into numbered items.

**1: The description for the "Producers of Stone Fruit" enterprise seems to disappear in the editor.** 
This is quite strange as this specific pull request only changes a single line of CSS, and doesn't effect the sanitizing or how the data is returned form the database at all. I don't think this issue can conceivably be caused by this PR. I'll try to investigate it anyway, but I think it's a separate issue.

**2: paragraphs have a weird border when the content is larger than the box and scroll bars are visible.**
What browser are you testing in? I've been developing in Firefox and the strange borders aren't present. Just had a quick look in Chrome and the borders are there (actually they're grey instead of blue), but the CSS debug seems to indicate they're being added by Chrome itself as opposed to our actual CSS files. I've updated the PR with a fix that should get rid of them. 

**3: The fix doesn't carry across to Product Descriptions**
The product descriptions don't currently have a WYSIWYG editor, but that's being added in a separate PR. The fix will apply to the new Product Descriptions content editor when it gets added. 

**4: Existing entries that have double spaces might need to be adjusted**
There hasn't been any change to the way the text is displayed in the frontend, so any existing enterprise descriptions should look the same as before in the frontend. The layout will look different in the backend but should be easier to update from now on?

Have I missed anything?

Matt-Yorkley:
Looking at your example with the list made with single line breaks, I can't see any way to replicate this either before or after the changes proposed in this PR. Is that example from an entry that was made before the WYSIWYG was introduced? Currently any line breaks get translated into paragraphs in the frontend, and it's not possible to make lists exactly like that. This hasn't been changed by this PR, it's been like that for quite a while...

linebreaks

Matt-Yorkley:
@sstead is it okay to pull the mysterious missing "Producers of Stone Fruit" decription out into a separate issue?

lin-d-hop:
@Matt-Yorkley It would be good to confirm that the PR is not the cause of the missing description. This PR couldn't be pushed if it might possibly break some descriptions, even though unlikely. Hit @RohanM up on slack to chase for the DB for testing locally.

Matt-Yorkley:
@lin-d-hop Okay, I've just loaded the staging data from Rohan and confirmed the "Producers of Stone Fruit" description disappears in the backend regardless of what branch is loaded.

lin-d-hop:
Great! So not ur changes. Might help if u can figure out why its disappearing though? @sstead are u happy with that?

Matt-Yorkley:
Further investigation: the description contains elements that can't be processed (or even created) by the textAngular WYSIWYG, including wrapping `div`s and `span` elements with weird classes. Suggests the data was created before the introduction of textAngular or inserted directly into the database?

sstead:
Looks all good.

**1: The description for the "Producers of Stone Fruit" enterprise seems to disappear in the editor.**
Thanks for investigating, I'm happy that is shouldn't happen to our existing profiles.

**2: paragraphs have a weird border when the content is larger than the box and scroll bars are visible.**
Fixed in Chrome.

Issue #1330, opened 2016-12-09
Limit countries in enterprise sign up to those with states


lin-d-hop:
In the enterprise sign up form State (or County in UK translation) a required field. However not all countries have states attached to them.

It is likely that many countries will have producers from multiple countries. But if a country has no states linked to it a user cannot sign up to this country. Therefore it would make sense to limit the countries displayed in the enterprise sign up drop down to those with states attached.

In fixing this issue it would be great if you could check that 'setting states' is clearly documented in the github wiki.

Note on UK staging default is not set to UK. Would be good to fix this at the same time.

myriamboure:
Lynne do you mean that we need to find an equivalent to "state" for each country, set it up, and then only the countries set up will be proposed?
In France for example we don't need states, but I guess the "search by state" is based on it so we used some other geographical notion not to loose the search functionality. I suspect that can apply to every country so why limiting display to countries with states if we say all countries should have states? 
Or should we make optional the fact of using the "state" field?

Matt-Yorkley:
I've been working on this today. @myriamboure will there not be a list of regions at all for France? In the UK our equivalent for states is counties.

Limiting the list to countries with their states set up means the user doesn't have to scroll through a list of every country in the world, which is quite nice. The list will only show the countries that have been configured on that OFN instance.

@lin-d-hop limiting the list of countries is quite simple but setting the default country might require changing things around quite a bit, and I'm conscious of the fact there are at least two PRs open that both modify the frontend registration form. Might be best to wait until they're both tested and merged?

myriamboure:
Ok, @Matt-Yorkley it seems to me that by default, all countries have the "state required" checkbox checked. So do you mean you would change that, say that by default the checkbox is unchecked and only when checkbox is checked the country is proposed in the list at registration?
I'm just wondering if that wouldn't be more logical to add a checkbox saying "active for the instance" and if checked country is displayed if not not displayed, just thinking if we have more and more people setting up instances it's better if it is logical and not too much tweaked. I think we should leave the instance decide if they want this info to be compulsory or not for the user to fill in (state required or not), which won't be the case anymore. BUT if it's much simpler to do the tweak as long as we know it is fine :-)

Then I would suggest to rename the filled then "active for the instance" and just say in the super admin guidebook that it is compulsory to set up an notion of a regional level that entreprises can associate to their profile and that users can search by.
Instances are then free to have some autofill rule if they don't want to ask the user to fill it.

[...]

PR #1334, opened 2016-12-12
Restrict countries without states, set default country

Matt-Yorkley:
Addressing #1330

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/country_selector.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/country_selector.js.coffee
@@ -1,0 +1,11 @@
+Darkswarm.directive "countrySelector", () ->
+  restrict: "A"
+  require: 'ngModel'
+  link: (scope, elem, attrs)->
+    defaultCountry = elem[0].getAttribute('data-default')
+    options = elem[0].querySelectorAll('option')
+
+    #Set default country on country dropdown
+    angular.forEach options, (value, key) =>
+      if value.label == defaultCountry
+        scope.enterprise.country = scope.countries[key]
--- a/app/assets/javascripts/darkswarm/directives/country_selector.js.erb.coffee
+++ b/app/assets/javascripts/darkswarm/directives/country_selector.js.erb.coffee
@@ -1,8 +1,8 @@
-Darkswarm.directive "countrySelector", () ->
+Darkswarm.directive "countrySelector", (RegistrationService) ->
   restrict: "A"
   require: 'ngModel'
   link: (scope, elem, attrs)->
-    defaultCountry = elem[0].getAttribute('data-default')
+    defaultCountry = "<%= Spree::Country.find_by_id(Spree::Config[:default_country_id]) %>"
     options = elem[0].querySelectorAll('option')
 
     #Set default country on country dropdown
--- a/app/assets/javascripts/darkswarm/controllers/registration/registration_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/registration/registration_controller.js.coffee
@@ -5,7 +5,8 @@ Darkswarm.controller "RegistrationCtrl", ($scope, RegistrationService, Enterpris
 
   $scope.steps = ['details', 'contact', 'type', 'about', 'images', 'social']
 
-  $scope.countries = availableCountries
+  $scope.countries = availableCountries.filter (country) ->
+    country.states.length > 0
 
   $scope.countryHasStates = ->
     $scope.enterprise.country.states.length > 0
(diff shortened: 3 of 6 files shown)


app/assets/javascripts/darkswarm/directives/country_selector.js.coffee
```diff
@@ -0,0 +1,11 @@
+Darkswarm.directive "countrySelector", () ->
```
mkllnk: Maybe we can save some code here if we set the country at the beginning in https://github.com/openfoodfoundation/openfoodnetwork/blob/4b8146dd00239a86b3461fb5c1b203f715abb3f7/app/assets/javascripts/darkswarm/services/enterprise_registration_service.js.coffee#L8. Would that work?

Matt-Yorkley: I think I couldn't get it to work that way. If I recall, the angular script couldn't access the Spree::Config and the indexing of the array was strange.

mkllnk: Ah, yes, it's currently not processed by Ruby. But you can add the `.erb` extension to do that. One example is `app/assets/javascripts/darkswarm/services/geo.js.erb.coffee`. That will be executed during compile time. So if you change the default country, you need to rebuild the assets.

lin-d-hop: @Matt-Yorkley could you take a look at adding Maikels suggestion?

daniellemoorhead:
---Doing a PR Cleanup---

@lin-d-hop where are we at with this? Worth assigning someone to it?

app/assets/javascripts/darkswarm/directives/country_selector.js.coffee
Matt-Yorkley: @mkllnk I've tidied it up a bit, but I still can't get it to work without the directive...

mkllnk:
Thank you @Matt-Yorkley. Can you extend the spec `spec/features/consumer/registration_spec.rb:11`? It needs a before block that creates two more countries. One should be before Australia, e.g. Albania. Instead of selecting Australia, the test case should check that Australia is already chosen, because it is the default country. The other additional country should not have any states. You should check that it is not available in the options. You need to remove these countries in an after block, because countries are not reset between test cases.

Example data from Spree's seeds:
```yml
countries_002:
  name: Albania
  iso3: ALB
  iso: AL
  iso_name: ALBANIA
  id: "2"
  numcode: "8"
countries_039:
  name: Chad
  iso3: TCD
  iso: TD
  iso_name: CHAD
  id: "39"
  numcode: "148"
```

If you use `Spree::Country.create!`, it will filter out most attributes. We either don't care about these attributes for this test case or we insert the countries via SQL.

Matt-Yorkley:
@mklink Specs added.

app/assets/javascripts/darkswarm/directives/country_selector.js.erb.coffee
```diff
@@ -1,8 +1,8 @@
-Darkswarm.directive "countrySelector", () ->
+Darkswarm.directive "countrySelector", (RegistrationService) ->
   restrict: "A"
   require: 'ngModel'
   link: (scope, elem, attrs)->
-    defaultCountry = elem[0].getAttribute('data-default')
```
RohanM: It seems more transparent and less coupled to Spree to pass in a default value via the "data-default" attribute as seems to be the previous behaviour. Is there any reason we're changing to injecting the default directly into the JS rather than passing it as an HTML attribute to the directive?

RohanM:
On the topic of whether we need a new directive here. It seems like what we need is more general than a country selector. We want a select box that is bound to a particular angular model field, and we want to be able to provide a default for it. Is it not possible to do this by using a regular `ngSelect` / `ngOptions` and initialising the value of the model field? If it's difficult to elegantly pass the default value through to angular, I wonder if we could make a much more simple and reusable directive, like `ofnSelectDefault` or `ofnOptionsDefault` that accepted a default value (eg. `ofn-select-default="123"`) and initialised the form in that way. What are your thoughts on that approach?

Matt-Yorkley:
@RohanM I've found setting default values on dropdowns in Angular really annoying at times. It seems to behave inconsistently depending on whether the select was created with ng-repeat or with ng-options, and whether the data being passed to each option is in the form of strings or objects. 

Also, in setting the default, Angular seems to want a reference to the index of the item in the array of options. As these are created by Angular and start from [0], but are obviously variable, it requires looping over the list and checking the values to find the index of the correct object.

I'll have a go at making a general directive...

Matt-Yorkley:
Superceded by #1609

Issue #1254, opened 2016-12-02
When UK user is logged in Connect & Learn are appearing in the menu  


sstead:
Device: Mobile/Tablet
Connect and Learn should show in Au only
![image](https://cloud.githubusercontent.com/assets/10843056/20822799/f77c7eca-b8a1-11e6-99d6-c0efa8e242e3.png)

sstead:
Fixed by the Connect and Learn toggle - https://github.com/openfoodfoundation/openfoodnetwork/pull/1336

PR #1336, opened 2016-12-16
Make feature toggle apply to mobile menu

RohanM:
Addressing #1254.

Add the missing feature toggle to these menu items. I also discovered that when the feature toggle is disabled, the "Groups" and "About" menu items are displayed on the large menu but not the mobile menu. So I've added them to the mobile menu for symmetry.


sstead:
All looks good - just need to check that Notices is showing and hiding correctly. 
image

Issue #1301, opened 2016-12-07
Payment type text wrapping when not needed. 


sstead:
Device: Mobile/Tablet
Where: Shopping cart
Recommendation: Text padding / line height should be reduced
![image](https://cloud.githubusercontent.com/assets/10843056/20952118/471590d6-bc7e-11e6-9cdc-3e3987db0ef6.png)

Matt-Yorkley:
The text folding was in place because a message box is shown to the right if the selected payment option has additional notes attached to it (shown when selected). 

![screenshot from 2016-12-17 20-37-42](https://cloud.githubusercontent.com/assets/9029026/21289640/0d9a48dc-c49a-11e6-8f0e-09f64af6b6d1.png)

I've adjusted the widths for small and medium devices so the text is full-width (instead of half) and the notes are shown underneath.

![screenshot from 2016-12-17 20-38-06](https://cloud.githubusercontent.com/assets/9029026/21289662/9182fe50-c49a-11e6-8ead-3137f2d81aee.png)

The payment options section uses the same sort of layout so I've adjusted it to match, as the issue was replicated there too.

lin-d-hop:
Great! Do you think it is worth also putting a max length on this field @Matt-Yorkley?

Matt-Yorkley:
I think it's ok as it is, it was just an issue with the way lines were wrapping. 50% of the screen is not much space on a 370px screen!

OliverUK:
Tested on my mobile phone (BQ Aquaris X5) using Firefox as browser. The checkout sections in question now look as below.

![screenshot_20170103-171050](https://cloud.githubusercontent.com/assets/17590921/21616082/15044276-d1d8-11e6-9e58-5b4945067e2d.png)
![screenshot_20170103-171133](https://cloud.githubusercontent.com/assets/17590921/21616083/15051ac0-d1d8-11e6-9dc8-5bcf34a884db.png)

OliverUK:
![screenshot_20170103-171821](https://cloud.githubusercontent.com/assets/17590921/21616364/196f7f8c-d1d9-11e6-95cb-d0476601e4ae.png)

PR #1339, opened 2016-12-17
Checkout layout adjustments for mobile view

Matt-Yorkley:
Addressing #1301. Payment options now span the whole width on mobile and tablets instead of half the screen, which caused weird text folding. Also applied the same change to the shipping options as they snap to the same widths in the same way on mobiles.


sstead:
Device: Nokia Lumia and Samsung
Browser: Internet Explorer and Chrome
Looks much better!

wp_ss_20170111_0002
wp_ss_20170111_0003
wp_ss_20170111_0004
wp_ss_20170111_0001

Issue #1343, opened 2016-12-19
Cart disappears when payment options not configured correctly.


Matt-Yorkley:
I've noticed that a hub's payment options have to be set to "testing", "development", "production" or "staging", and if there are no payment methods assigned to a hub with the correct environment set, it assumes the hub has no payment options. Then when I click view cart I get bounced to the home page with an error message about the hub being "temporarily closed", which is not very informative as the hub definitely appears to be open. Also the cart gets emptied without warning. It seems it happens when the hub either has no valid payment options or no valid delivery options. Maybe that check should be made before the user starts interacting with the hub's products rather than after their cart is filled up.

Shopping: 

![screenshot from 2016-12-19 11-15-45](https://cloud.githubusercontent.com/assets/9029026/21310974/d48a45da-c5dc-11e6-913f-0787d2217b3c.png)

After clicking on cart:

![screenshot from 2016-12-19 11-15-35](https://cloud.githubusercontent.com/assets/9029026/21310980/de7e2f48-c5dc-11e6-92d2-59b331f8b257.png)

I'm thinking the shop should not be shown as open if customers aren't actually able to purchase anything (payment or shipping not configured), as opposed to appearing open and only telling the customer at the checkout that the shop is closed (for an unknown reason).

lin-d-hop:
For the most part this happens. If there are no payment or shipping methods hubs with active order cycles do not appear as open. So it seems that this same check does not consider errors in payment or shipping methods.

The solution, then, would be a update the existing check on whether a hub has shipping/payment methods before displaying as open. Not sure where this happens... can dig around if needed :-)

lin-d-hop:
@OliverUK ready to test now :-)

OliverUK:
@lin-d-hop 
Questions for testing purposes. The environment is set in super admin so under what circumstances would a hub have a shipping or payment method which isn't configured properly unless a super admin fiddles with it?
I assume if I sign up on the production site and create an enterprise, my methods are automatically set to production?

OliverUK:
Incidentally I tried a not-properly-configured Paypal payment option on the Stroudco shop on production and was able to shop but got an error message on checkout. It just stayed on the checkout page waiting for me to take an alternative action.

lin-d-hop:
@OliverUK You are correct that this won't generally happen. Matt would have discovered this on playing with test data from another database.
It still seems like a good check to have in place, but it should never come up on production.

sstead:
[...]

Issue #495, opened 2015-04-15
Simple OC doesn't check for or warn about invalid shop - payment / shipping


sstead:
No 'you need to have payment methods to create an order cycle' warning pops up for a simple order cycle.

Instead, when a customer tries to place an order, they get kicked out of the store with this message...""The hub you have selected is temporarily closed for orders. Please try again later."

It looks like perhaps the simple order cycle doesn't quite know what to do when the shop isn't 'valid' - we have built a couple of points where this gets picked up in the full OC, but probably need a couple in simple OC as well. This might be what's causing Bug #

myriamboure:
I tried to create a simple OC (producer own shop), the producer doesn't have a payment method set up. I can create the OC without any issue. BUT when I go to my admin I see this warning message (the shop will remain invisible until you have a payment/shipping method set up)
![screenshot from 2018-03-01 11-31-05](https://user-images.githubusercontent.com/8135376/36840167-5bb20164-1d44-11e8-8d76-c855da5be40e.png)

So the customer can't access the shop if no payment/shipping method.
Closing as seems solved. Feel free to open again if you think the issue is still there @sstead

PR #1347, opened 2016-12-20
Hubs display as closed when not configured for payment or shipping

Matt-Yorkley:
Addressing issue #1343 (and duplicate issue #495 ?)

Hubs now display as closed when browsing shops in the frontend if they are not "ready for checkout", i.e. don't have payment or shipping methods properly configured.

Previously they displayed as open, allowed a user to shop and add things to the basket, but then bounced the user to the homepage with an error message and emptied the cart once the user tried to go to the checkout.

The backend already displays messages to an enterprise user if their enterprise does not have payment or shipping methods properly configured, and correctly warns that customers will not be able to shop at their hub:

screenshot from 2016-12-20 12-37-17

Updated specs to reflect this.

### DIFF
--- a/lib/open_food_network/enterprise_injection_data.rb
+++ b/lib/open_food_network/enterprise_injection_data.rb
@@ -1,7 +1,7 @@
 module OpenFoodNetwork
   class EnterpriseInjectionData
     def active_distributors
-      @active_distributors ||= Enterprise.distributors_with_active_order_cycles
+      @active_distributors ||= Enterprise.distributors_with_active_order_cycles.andand.ready_for_checkout
     end
 
     def earliest_closing_times
--- a/spec/features/consumer/shops_spec.rb
+++ b/spec/features/consumer/shops_spec.rb
@@ -201,4 +201,3 @@ feature 'Shops', js: true do
     end
   end
 end
-
--- a/spec/features/consumer/shops_spec.rb
+++ b/spec/features/consumer/shops_spec.rb
@@ -201,3 +201,4 @@ feature 'Shops', js: true do
     end
   end
 end
+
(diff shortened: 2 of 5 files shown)


lib/open_food_network/enterprise_injection_data.rb
```diff
@@ -1,7 +1,7 @@
 module OpenFoodNetwork
   class EnterpriseInjectionData
     def active_distributors
-      @active_distributors ||= Enterprise.distributors_with_active_order_cycles
+      @active_distributors ||= Enterprise.distributors_with_active_order_cycles.andand.ready_for_checkout
```
mkllnk: Since `distributors_with_active_order_cycles` is a scope, I don't think you need `andand`.

mkllnk: I found three usages of this:

- app/helpers/shared_helper.rb
- app/serializers/api/enterprise_serializer.rb
- app/views/json/partials/_hub.rabl

All of them call the `include?` method on that collection to know if a certain enterprise is active. That sounds like a really poor way of getting that attribute. The computation time will grow linearly with the number of active enterprises.

So this fixes the issue, but it also points to a possible performance problem.

spec/features/consumer/shops_spec.rb
```diff
@@ -201,4 +201,3 @@ def click_link_and_ensure(link_text, check)
     end
   end
 end
-
```
mkllnk: Is this to trigger a Buildkite build or Travis? It seems pretty ugly, but Buildkite has no proper search function for builds. Maybe that is something to ask them about. In Aus we tend to rebase branches.

lin-d-hop: Ok great I'll rebase in future.
Yeah was purely to force a buildkite build. Agreed it's ugly!

daniellemoorhead:
@Matt-Yorkley can we close this one now that it's been rebased as #1516?

Matt-Yorkley:
@daniellemoorhead Ah, I'd already rebased this branch and pushed new commits on top of it. Also I didn't create the other rebased one, so I can't push directly to it. I've cleaned it up and resubmitted as #1522.

PR #1352, opened 2016-12-22
Fix credit card checkout

RohanM:
We were seeing a bug where credit card checkouts (eg. with Pin Payments) would fail at checkout. Investigation revealed that the credit card number was being lost before it was sent to the payment gateway.

After wading through the thrice-sludgy swamps of git-bisect, I found two places where we were reloading the `Payment` or `Order`, and since the credit card is an in-memory attribute only, it was being lost.

We weren't picking this up in test because we were testing with `Spree::Gateway::Bogus`, which uses payment profiles by default, and the payment profile was being saved before the credit card number was lost. Using `Spree::Gateway::BogusSimple` exposed the bug.

To solve the problem, I removed one reload outright, and replaced the other with more targeted data refresh. I haven't confirmed that these don't break anything. Please test that payment method fees and taxes are appearing correctly on first page load.

### DIFF
--- a/app/models/spree/payment_decorator.rb
+++ b/app/models/spree/payment_decorator.rb
@@ -16,7 +16,6 @@ module Spree
         adjustment.save
       else
         payment_method.create_adjustment(adjustment_label, order, self, true)
-        reload
       end
     end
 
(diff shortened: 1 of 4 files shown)


RohanM:
Continued in #1372.

Issue #1346, opened 2016-12-20
Specs use $ for currency


Matt-Yorkley:
When running specs on a non-AUS setup, various tests fail as they are looking for values with "$" as the currency symbol. This isn't a big deal, but it does slow down the dev process when there are lots of errors.

2 possible solutions: internationalise all the tests (seems a bit unnecessary), or set some config options that explicitly set the default currency symbol to "$" before tests are run (easy option). 

Not a critical issue, but it would streamline things for non-AUS devs.

lin-d-hop:
Yeah wavy hands. Would be nice to internationalise this!

Matt-Yorkley:
@lin-d-hop Do you think it would be preferable to have all tests run using localised currency settings? I haven't had a proper look, but it might be an epic faff and could make a lot of specs more messy and complex than they need to be. Setting the default currency to $ whilst the tests run and just having a single test that checks currency functionality might be a lot cleaner...?

Either way it'll make life easier when it's sorted :)

lin-d-hop:
@Matt-Yorkley My vote is setting default currency for simplicity. 
@stveep @sigmundpetersen any thoughts to the contrary?

sigmundpetersen:
No obejctions :)

stveep:
Agree - `I guess in config/environments/test.rb`

myriamboure:
@Matt-Yorkley can I just close this one? Or do you want me to open an icebox item so that we can prioritize later ? If I understand well the focus is to make development easier by improving tests speed and efficiency right?

myriamboure:
In objective to improve efficiency in developing / reviewing / testing / staging, I have created an icebox "developers don't waste precious time while developing" https://community.openfoodnetwork.org/t/developers-dont-waste-precious-time-while-developing/1273 and attach this in it, so closing it for now, we can reopen when we prioritize.

lin-d-hop:
Reclosing. #gitcull

PR #1357, opened 2017-01-03
Set default currency for tests

Matt-Yorkley:
Addressing #1346.

Set default currency in testing environment to be AUD, because the tests are littered with hard-coded values like "$10.32".

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -36,6 +36,9 @@ Openfoodnetwork::Application.configure do
   # Tests assume English text on the site.
   config.i18n.default_locale = "en"
 
+  # Tests assume Aus timezone.
+  config.time_zone = "Australia/Melbourne"
+
   # Use SQL instead of Active Record's schema dumper when creating the test database.
   # This is necessary if your schema can't be completely dumped by the schema dumper,
   # like if you have constraints or database-specific column types
@@ -53,3 +56,7 @@ end
 
 # Allows us to use _url helpers in Rspec
 Rails.application.routes.default_url_options[:host] = 'test.host'
+# Set the timezone for phantomjs with TZ
+ENV['TZ'] = Rails.application.config.time_zone
+# Set the timezone for karma with TIMEZONE
+ENV['TIMEZONE'] = Rails.application.config.time_zone
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -93,6 +93,9 @@ RSpec.configure do |config|
   # Ensure we start with consistent config settings
   config.before(:each) { Spree::Config.products_require_tax_category = false }
 
+  # Tests assume AUD is the default currency.
+  config.before(:each) {Spree::Config.currency = "AUD" }
+
   # Helpers
   config.include Rails.application.routes.url_helpers
   config.include Spree::UrlHelpers


config/environments/test.rb
```diff
@@ -37,7 +37,7 @@
   config.i18n.default_locale = "en"
 
   # Tests assume Aus timezone.
-  config.time_zone = "Melbourne"
+  config.time_zone = "Australia/Melbourne"
```
RohanM: Could you please squash this commit into the previous? Might as well just have one commit with the correct timezone.

RohanM:
With a brief look, I suspect the spec failure might be for an unrelated intermittent failure.

Matt-Yorkley:
@RohanM Squashed commits and resolved conflict. The flaky test passed this time.

daniellemoorhead:
@Matt-Yorkley where to next with this PR?

Matt-Yorkley:
@daniellemoorhead I think the only thing remaining was to squash the last 2 commits together, which is now done. Back to @RohanM ?

daniellemoorhead:
@RohanM can you have a look at this post Easter once the NDD stuff you're working on is done? Ta :-)

enricostano:
Should we coordinate on test strategies?

https://github.com/openfoodfoundation/openfoodnetwork/pull/1521/files#diff-93830fa29d616f7c87903d08b5b1b29aR97

daniellemoorhead:
@Matt-Yorkley and @enricostano @RohanM has approved this change and @sstead will be testing it. Not sure what you mean by test strategies - dev unit tests or Sally doing the "front end trying to break it" testing. So we're going to proceed as we normally do by testing and deploying this one, get in touch if you want us to look at both at the same time 😄

daniellemoorhead:
@Matt-Yorkley @jeronimo @enricostano  can the three of you please connect and discuss what should happen with these two PRs (this one and #1521), and then let us know what you decide and we'll move forward with that. We're not sure what to do in the meantime so will await your instructions and not touch either.

spec/spec_helper.rb
```diff
@@ -93,6 +93,9 @@
   # Ensure we start with consistent config settings
   config.before(:each) { Spree::Config.products_require_tax_category = false }
 
+  # Tests assume AUD is the default currency.
+  config.before(:each) {Spree::Config.currency = "AUD" }
```
enricostano: Is there any way to set this once instead of per each example? `before(:each)` is time and resource consuming.

jeronimo: `before(:all)` could do the job.

Unfortunately, somehow inside current `spree_core` factories `Spree::Calculator::FlatRate` still take currency from `CURRENCY` variable from `application.yml` and this is not the only failing tests which return `no_shipping_methods_available` error because of currencies mismatch.

Examples of failure:
* `rspec ./spec/jobs/finalize_account_invoices_spec.rb:198`
* ` rspec ./spec/features/consumer/shopping/variant_overrides_spec.rb:152`

mkllnk:
@Matt-Yorkley I'm actually not sure if this should be merged now or later. Can you give me your motivation again? It was pointed out that not all tests pass yet, but it would be an improvement if more tests pass. Can you test that? Maybe do two Travis runs with e.g. USD as currency. One run with your changes and the other one without. Then we know how many tests your work fixes.

Matt-Yorkley:
@mklink Initially I made this as a quick solution to speed up my dev process, and at the time I was the only person having issues with failing currency tests. I'll have to take a closer look at those last couple of failing tests, but this has slipped down my to-do list.

daniellemoorhead:
@mkllnk / @Matt-Yorkley where do we go next with this PR?

mkllnk:
Okay @Matt-Yorkley. Ping us again once you it's on top of you todo list again.

daniellemoorhead:
Removed from August dot point release, ping us via aus review label when it's ready to go @Matt-Yorkley :)

myriamboure:
@Matt-Yorkley shall we just close it for now, and open a wishlist item, and when it becomes a priority for you or others we create an icebox item and prioritize it?

lin-d-hop:
Closing and this isn't getting to the top of anyones wishlist too soon :-/

Issue #1027, opened 2016-06-04
sitemap.xml is empty


pmackay:
ComfortableMexicanSofa is generating a `sitemap.xml` but its empty (presumably because in UK, CMS is not being used for anything, is it being used in Aus?).

It would be beneficial for SEO if OFN had a sitemap that contained links to all the basic pages and shops.

@oeoeaio @mkllnk have you considered this before? 

cc @lin-d-hop

mkllnk:
No. I never came across that.

daniellemoorhead:
We'll chat about this during the week when we're in the office @pmackay and let you know.

pmackay:
@mkllnk what do you think about disabling the CMS sitemap generation and adding a controller to create one? I could look at that if you agree.

mkllnk:
Sounds good. @RohanM would know if the CMS is used at all. Maybe to store some content of the site. But we should have our own sitemap. That would be great.

daniellemoorhead:
@RohanM any thoughts on @pmackay's suggestion above of disabling the CMS-generated sitemap and adding a controller to create one?

daniellemoorhead:
@RohanM checking in - thoughts on this one?

RohanM:
I've recently removed CMS, since we haven't been using it for a long time. So now there should be no `sitemap.xml` at all. @pmackay, would you be interested in doing a PR to provide a sitemap?

PR #1366, opened 2017-01-10
Uk/1027 sitemap

pmackay:
Fix for #1027.

### DIFF
--- a/app/views/sitemap/index.xml.haml
+++ b/app/views/sitemap/index.xml.haml
@@ -14,5 +14,4 @@
   - for group in @groups
     %url
       %loc= group_url(group)
-      %lastmod= enterprise.updated_at.strftime('%Y-%m-%d')
       %changefreq yearly
--- a/app/views/sitemap/index.xml.haml
+++ b/app/views/sitemap/index.xml.haml
@@ -1,8 +1,8 @@
 !!! XML
 %urlset{xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9"}
-  - for page in @pages
+  - for page_url in @page_urls
     %url
-      %loc= root_url + page
+      %loc= page_url
       %changefreq monthly
 
   - for enterprise in @enterprises
@@ -14,4 +14,4 @@
   - for group in @groups
     %url
       %loc= group_url(group)
-      %changefreq yearly
+      %changefreq monthly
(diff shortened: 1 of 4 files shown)


app/views/sitemap/index.xml.haml
```diff
@@ -0,0 +1,17 @@
+!!! XML
+%urlset{xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9"}
+  - for page in @pages
+    %url
+      %loc= root_url + page
```
RohanM: It'd be nice to build URLs using the helpers rather than crafting them like this, since if anything changes to these URLs then this will break without us knowing. Perhaps `@pages` could contain `shop_url`, `map_url`, etc.

app/views/sitemap/index.xml.haml
```diff
@@ -0,0 +1,17 @@
...
+  - for enterprise in @enterprises
+    %url
+      %loc= enterprise_shop_url(enterprise)
+      %lastmod= enterprise.updated_at.strftime('%Y-%m-%d')
+      %changefreq weekly
+
+  - for group in @groups
+    %url
+      %loc= group_url(group)
+      %changefreq yearly
```
RohanM: This seems like a long time for groups, especially as we don't have data on how often they change. What do you reckon of using monthly here instead?

app/views/sitemap/index.xml.haml
pmackay: Good call. Have fixed.

app/views/sitemap/index.xml.haml
pmackay: Yes, makes sense.

pmackay:
@mkllnk agreed, would be interesting to see how often it is called by search engines, cache as needed.

mkllnk:
The examples on [Google](https://support.google.com/webmasters/answer/183668?hl=en) and [Sitemaps.org](https://www.sitemaps.org/protocol.html) contain a `xmlns` attribute which we don't have. But still, several online tools rate the sitemap valid without any errors:

- https://www.xml-sitemaps.com/index.php?op=validate-xml-sitemap&go=1&sitemapurl=https%3A%2F%2Fstaging2.openfood.com.au%2Fsitemap.xml&submit=Validate
- http://seositecheckup.com/seo-audit/sitemap-test/staging2.openfood.com.au/sitemap.xml
- http://www.validome.org/google/validate?url=https://staging2.openfood.com.au/sitemap.xml&lang=en&googleTyp=SITEMAP

:+1:

mkllnk:
Our staging server delivers the sitemap in 250ms to 1000ms.

Issue #1195, opened 2016-10-25
Invoice adaptation for French instance


myriamboure:
How can I translate and adapt the invoice format that you can print from the order?
We need to make a French version (I guess every instance will need its own version)

I also need to be able to see the different VAT rate on the invoice.
Even without changing anything in the current set up (VAT is included in prices, so included in shop front, and detail "including xxx VAT" in the confirmation) we would like in the appearance of the invoice and order confirmation to have one aggregating line **for each VAT rate** (2,1%, 5,5%, 10%, 20%). And if no VAT for this rate, the line doesn't appear (ideally).

myriamboure:
Ping @mkllnk or @oeoeaio any idea where /how to translate and adapt the invoice?

oeoeaio:
Hi @myriamboure,

Translation of invoices will need to be added in the same way as any other pages on the site, at the moment it looks like some elements of it are hard-coded so there will be a little bit of initial work (perhaps an hour or so) to add the translated elements. Any developer familiar with rails should be able to do this, so if you have someone there who is willing to add translations to the template, feel free to get them to change `app/views/spree/admin/orders/invoice.html.haml` and submit a pull request and we'll merge it ASAP.

With regard to displaying the different tax rates - this is a little more complicated, and I will need to investigate further before I can give you a good idea of how much work this is. Am I right in understanding that this is the same as point 3 in your comment [here](http://community.openfoodnetwork.org/t/outstanding-tax-requirements-jan-2016/487/17?u=oeoeaio)? If so, I will quote on it as part of the outstanding tax requirements work. 

Hope that helps!

myriamboure:
Thank you @oeoeaio and you're right, this is the point 3, but an urgent point for one of our user and that can be done first, before the whole job on tax refund is done, I guess (I hope). As we have a workaround for the moment by including VAT in prices (tax rate set up as "VAT included in price), so display in shopfront, so we "just" need, in the calculation of the VAT, to isolate the aggregation around each VAT rate and display each aggregation seperately. This is the most urgent for our user and we need to find a way to manage that (and also adapt the sales_tax report accordingly), we have a small budget for that from our user, but maybe that sounds better to raise a bucket on co-budget to do the whole tax job at once... 

Thanks for the translation guidance :-)

myriamboure:
Ok, so given the evolution of discussion on tax overhaul, where is what I propose to answer the French need:

- we aggregate the amount for each "tax category"
[...]

Issue #1194, opened 2016-10-25
Sales_tax report adaptation for French instance


myriamboure:
For France, our users need this report for their accounting / VAT declarations and payments.
But we need in the report for each order to have the total tax for each VAT rate, so total tax 5,5%, total tax 10%, total tax 20%. and then the total tax aggregation those figures.
So that the hub users can transmit this document to their accountant.

myriamboure:
Description of what we want here: https://docs.google.com/document/d/1GhFu8138IDXvs-7vhUBOdhm-2BsZyjTIqh-rC8eeog0/edit?usp=sharing

myriamboure:
Should we change the sales-tax report for everyone that way? Or should do a separate one "sales-tax per VAT rate"? Ping @NickWeir63 or @lin-d-hop tell me what you think, @oeoeaio @sstead I think this report could replace the current one, tell me if that sounds ok or if we do a seperate one.

NickWeir63:
Myriam's google doc proposal looks fine to me thanks.  as long as we can download it as a csv like all the other reports we can always pull out the separate rates if necessary

sstead:
Hi @myriamboure, I'd double check before removing the 'tax on delivery' and 'tax on fees' columns. I assume they were put there for a reason (perhaps a user requested this breakdown, although I'm not sure who, UK maybe?), and at a minimum, this breakdown makes it easy to see where the tax has accumulated from and verify that it's correct. In regards to the process of first selecting your hub and then having tax options become available- this is different to the behavior of most reports. It sounds intuitive, but consistency is also good? I think for Aus users, with only 1 tax zone, this proposed change won't impact them very much (assuming tax on delivery/tax on fees columns remain), so it would be ok to change the existing report. Just my thoughts.

NickWeir63:
good point Sally.  i cant remember why VAT on fees is currently separated out but @myriamboure would it be OK to leave it separated out and hubs that want to add them back in can do that after the csv is downloaded?

myriamboure:
Actually the issue is that with the job we are doing on VAT internationalization, there can be dozens of different tax rates, so that's why I propose to select first the hub, and that the table display according to the VAT categories/rates used by the hub.

[...]

PR #1374, opened 2017-01-11
Internationalization of invoice and tax rate report

ltrls:
This pull request is related to issues #1195 and #1194.
It consists of two things : 
- the invoice style has been adapted to include multiple tax rates, internationalization of texts, and a logo and additional text from the hub (it looke like this: https://framapic.org/FV3jteP7a3fn/UUd2oXZXcRs0.png (FR example with VAT included in price), https://framapic.org/P8XG8hnTVw3T/VWi9mUpV6BZH.png (AUS example without VAT included in price))
- the sales tax report has an option to display all tax rates for each order
I have also changed the i18n production config so that it fallbacks to EN. 
Please note that I am not a seasoned contributor and although I've done as much testing as I thought I'd be happy to have advices on anything that's improvable :) (writing this makes me think that I could write tests for the methods I added in models, I'll see how I can do that)


kirstenalarsen:
Hey @oeoeaio, could you look at this (once @ltrls has the tests passing), and bill/track your time towards France. Thanks :) ping @daniellemoorhead  @myriamboure

oeoeaio:
👍  will have a look as soon as I can

lib/open_food_network/sales_tax_report.rb
```diff
@@ -24,19 +31,39 @@ def orders
...
 
     private
 
+    def relevant_rates
+      queries = [ search.result.joins(:line_items => {:adjustments => :tax_rate}).select('spree_tax_rates.*').uniq,
+                  search.result.joins(:adjustments => :tax_rate).select('spree_tax_rates.*').uniq ]
+      queries.map do |query|
+        ActiveRecord::Base.connection.select_all(query)
+      end.sum.map do
+        |tax_rate| Spree::TaxRate.new(tax_rate, without_protection: true)
```
oeoeaio: Just a style thing for consistency, can we move `|tax_rate|` up to the previous line?

lib/open_food_network/sales_tax_report.rb
```diff
@@ -24,19 +31,39 @@ def orders
...
 
 
     private
 
+    def relevant_rates
+      queries = [ search.result.joins(:line_items => {:adjustments => :tax_rate}).select('spree_tax_rates.*').uniq,
+                  search.result.joins(:adjustments => :tax_rate).select('spree_tax_rates.*').uniq ]
+      queries.map do |query|
+        ActiveRecord::Base.connection.select_all(query)
+      end.sum.map do
```
oeoeaio: Maybe consider splitting some of this out into a separate method to improve readability/testability?

lib/open_food_network/sales_tax_report.rb
```diff
@@ -24,19 +31,39 @@ def orders
...
+           order.bill_address.full_name, order.distributor.andand.name]
+        end
+      else
+        orders.map do |order|
+          [order.number, order.total - order.total_tax] +
+            relevant_rates.map { |rate| order.tax_adjustment_totals.fetch(rate, 0) } +
+            [order.total_tax, order.display_total]
+        end
       end
+
```
oeoeaio: Don't need this empty line.

lib/open_food_network/sales_tax_report.rb
```diff
@@ -9,9 +9,16 @@ def initialize(user, params)
...
-        "Total Tax (#{currency_symbol})", "Customer", "Distributor"]
+      case params[:report_type]
+      when "tax_types"
+        ["Order number", "Date", "Items", "Items total (#{currency_symbol})", "Taxable Items Total (#{currency_symbol})",
+          "Sales Tax (#{currency_symbol})", "Delivery Charge (#{currency_symbol})", "Tax on Delivery (#{currency_symbol})", "Tax on Fees (#{currency_symbol})",
+          "Total Tax (#{currency_symbol})", "Customer", "Distributor"]
+      else
+        ["Order number", "Total excl. VAT"] +
+          relevant_rates.map { |rate| "%s (%.1f%%)" % [rate.name, rate.amount.to_f * 100] } +
+          ["Total VAT", "Total incl. VAT"]
```
oeoeaio: Need to add translations for these headers.

lib/open_food_network/sales_tax_report.rb
```diff
@@ -9,9 +9,16 @@ def initialize(user, params)
...
-        "Sales Tax (#{currency_symbol})", "Delivery Charge (#{currency_symbol})", "Tax on Delivery (#{currency_symbol})", "Tax on Fees (#{currency_symbol})",
-        "Total Tax (#{currency_symbol})", "Customer", "Distributor"]
+      case params[:report_type]
+      when "tax_types"
+        ["Order number", "Date", "Items", "Items total (#{currency_symbol})", "Taxable Items Total (#{currency_symbol})",
+          "Sales Tax (#{currency_symbol})", "Delivery Charge (#{currency_symbol})", "Tax on Delivery (#{currency_symbol})", "Tax on Fees (#{currency_symbol})",
+          "Total Tax (#{currency_symbol})", "Customer", "Distributor"]
+      else
+        ["Order number", "Total excl. VAT"] +
+          relevant_rates.map { |rate| "%s (%.1f%%)" % [rate.name, rate.amount.to_f * 100] } +
```
oeoeaio: This is cool, but I think that displaying it in this way will be slightly odd for some countries. For example: in Australia, we only have one rate of tax, and it either applies or it doesn't, so there will be no difference between this per-rate total column and the `Total VAT` column. Not sure what to do about this - we will test it in the browser and then work out what we can do.

db/migrate/20161215230219_add_invoice_text_and_invoice_logo_to_enterprises.rb
```diff
@@ -0,0 +1,6 @@
+class AddInvoiceTextAndInvoiceLogoToEnterprises < ActiveRecord::Migration
+  def change
+    add_column :enterprises, :invoice_text, :text
+    add_column :enterprises, :invoice_logo, :boolean, default: false
```
oeoeaio: Perhaps `display_invoice_logo` would give a better indication of what this field does? Then rails magic will allow us to do `enterprise.display_invoice_logo?`, which is a bit more readable. I thought this was a new image field when I first read it in the code.

lib/open_food_network/sales_tax_report.rb
ltrls: That's right, perhaps you'd like that I not have the "tax by tax rates" report be the default one and instead have the original one ("tax by tax types")?

lib/open_food_network/sales_tax_report.rb
ltrls: I've now translated all headers of this report

lib/open_food_network/sales_tax_report.rb
ltrls: Yes!

lib/open_food_network/sales_tax_report.rb
ltrls: Yes!

db/migrate/20161215230219_add_invoice_text_and_invoice_logo_to_enterprises.rb
ltrls: Nice idea! As I probably am the only person having applied this migration I changed it directly :)

lib/open_food_network/sales_tax_report.rb
ltrls: It's already kind of in its separate method. If you mean have it in a model/decorator I can't as it builds upon the report search. I agree it's not the prettiest function but it was the only way I found to get relevant rates without making lots of SQL request (actually it perfoms only 2, one for tax adjustments on the order and one for tax adjustments on line_items [1])
[1] for instance : 
```
SELECT DISTINCT spree_tax_rates.* FROM "spree_orders" INNER JOIN "spree_line_items" ON "spree_line_items"."order_id" = "spree_orders"."id" INNER JOIN "spree_adjustments" ON "spree_adjustments"."adjustable_id" = "spree_line_items"."id" AND "spree_adjustments"."adjustable_type" = 'Spree::LineItem' INNER JOIN "spree_tax_rates" ON "spree_tax_rates"."id" = "spree_adjustments"."originator_id" AND spree_adjustments.originator_type = 'Spree::TaxRate' WHERE "spree_orders"."id" IN (4, 3, 9, 10, 7) AND (completed_at IS NOT NULL) AND (state != 'canceled') AND ("spree_orders"."completed_at" > '2015-12-31 23:00:00.000000')
SELECT DISTINCT spree_tax_rates.* FROM "spree_orders" INNER JOIN "spree_adjustments" ON "spree_adjustments"."adjustable_id" = "spree_orders"."id" AND "spree_adjustments"."adjustable_type" = 'Spree::Order' INNER JOIN "spree_tax_rates" ON "spree_tax_rates"."id" = "spree_adjustments"."originator_id" AND spree_adjustments.originator_type = 'Spree::TaxRate' WHERE "spree_orders"."id" IN (4, 3, 9, 10, 7) AND (completed_at IS NOT NULL) AND (state != 'canceled') AND ("spree_orders"."completed_at" > '2015-12-31 23:00:00.000000')
```

ltrls: Actually if you like it more I may simply display all tax rates columns even if they are not in the searched reports :)

myriamboure:
@oeoeaio are the corrections that made @ltrls ok for you? The shop opens in 3 weeks so I feel a bit anxious about this :-o I hope we manage to merge and upgrade by then! Also the translation that Pierre added, I guess I'll be able to change them in Transifex while this is merged? Cause I need to change some of them. Thanks :-)

lib/open_food_network/sales_tax_report.rb
oeoeaio: Ok, so I think I now understand this a bit better. Is there any particular reason that you need new instances of Spree::TaxRate instead of just using the ones from the db? Could we could re-write this function? How about:

```ruby
def relevant_rates
  return @relevant_rates unless @relevant_rates.nil?
  item_rate_ids = search.result.joins(:line_items => {:adjustments => :tax_rate}).select('spree_tax_rates.id').uniq
  order_rate_ids = search.result.joins(:adjustments => :tax_rate).select('spree_tax_rates.id').uniq
  @relevant_rates = Spree::TaxRate.where('id IN (?) OR id in (?)', item_rate_ids, order_rate_ids)
end
```
Sure, it is three queries rather than two, but we could make it one query by using subqueries if you were particularly worried about the number of queries. I think the important thing is that we are not re-running all three queries for every order that we pull up, it just runs once and stores the result for later use.

Perhaps you are right - maybe we should just display all available rates?

lib/open_food_network/sales_tax_report.rb
```diff
@@ -24,19 +42,38 @@ def orders
...
+    def relevant_rates
+      queries = [ search.result.joins(:line_items => {:adjustments => :tax_rate}).select('spree_tax_rates.*').uniq,
+                  search.result.joins(:adjustments => :tax_rate).select('spree_tax_rates.*').uniq ]
+      queries.map do |query|
+        ActiveRecord::Base.connection.select_all(query)
+      end.sum.map do |tax_rate|
+        Spree::TaxRate.new(tax_rate, without_protection: true)
+      end
+    end
+
```
oeoeaio: Ok, so I think I now understand this a bit better. Is there any particular reason that you need new instances of Spree::TaxRate instead of just using the ones from the db? Could we could re-write this function? How about:

```ruby
def relevant_rates
  return @relevant_rates unless @relevant_rates.nil?
  item_rate_ids = search.result.joins(:line_items => {:adjustments => :tax_rate}).select('spree_tax_rates.id').uniq
  order_rate_ids = search.result.joins(:adjustments => :tax_rate).select('spree_tax_rates.id').uniq
  @relevant_rates = Spree::TaxRate.where(id: item_rate_ids | order_rate_ids)
end
```

Sure, it is three queries rather than two, but I think it is more readable and we could make it one query by using subqueries if you were particularly worried about the number of queries. I think the important thing is that we are not re-running all three queries for every order that we pull up, it just runs once and stores the result for later use.

Perhaps you are right - maybe we should just display all available rates?

lib/open_food_network/sales_tax_report.rb
```diff
@@ -9,9 +9,27 @@ def initialize(user, params)
...
+         I18n.t(:report_header_delivery_charge, currency_symbol: currency_symbol),
+         I18n.t(:report_header_tax_on_delivery, currency_symbol: currency_symbol),
+         I18n.t(:report_header_tax_on_fees, currency_symbol: currency_symbol),
+         I18n.t(:report_header_total_tax, currency_symbol: currency_symbol),
+         I18n.t(:report_header_customer),
+         I18n.t(:report_header_distributor)]
+      else
+        [I18n.t(:report_header_order_number),
+         I18n.t(:report_header_total_excl_vat, currency_symbol: currency_symbol)] +
+        relevant_rates.map { |rate| "%s (%.1f%%) (%s)" % [rate.name, rate.amount.to_f * 100, currency_symbol] } +
```
oeoeaio: Just thinking that `relevant_rates` will return a copy of both the tax-inclusive and tax-exclusive versions as separate rates if they both exist, and are both used. This may be a fairly rare case, but just thought I would raise it. @myriamboure, what do you think? Should we be reporting taxes accumulated under tax-inclusive and tax-exclusive versions of the same rate in the same column?

app/models/spree/line_item_decorator.rb
```diff
@@ -56,6 +57,10 @@ def included_tax
     adjustments.included_tax.sum(&:included_tax)
   end
 
+  def tax_rates
+    product.tax_category.tax_rates
```
oeoeaio: Not every product has a tax_category, so we will need to handle the nil case here:

```ruby
product.tax_category.andand.tax_rates || []
```

app/views/spree/admin/orders/_invoice_table.html.haml
```diff
@@ -2,13 +2,15 @@
...
+        %h5= t(:invoice_column_qty)
       %th{:align => "right", :width => "15%"}
-        %h4= @order.total_tax > 0 ? t(:invoice_column_tax) : ""
+        %h5= @order.has_taxes_included ? t(:invoice_column_unit_price_with_taxes) : t(:invoice_column_unit_price_without_taxes)
       %th{:align => "right", :width => "15%"}
-        %h4= t(:invoice_column_price)
+        %h5= @order.has_taxes_included ? t(:invoice_column_price_with_taxes) : t(:invoice_column_price_without_taxes)
+      %th{:align => "right", :width => "15%"}
+        %h5= @order.total_tax > 0 ? t(:invoice_column_tax_rate) : ""
   %tbody
```
oeoeaio: If we do this, we can just hide the tax column altogether for orders without any tax (common in Australia):

```ruby
- if @order.total_tax > 0
  %th{:align => "right", :width => "15%"}
    %h5= t(:invoice_column_tax_rate)
```

Obviously need to do the same thing for the column itself below...

app/views/spree/admin/orders/_invoice_table.html.haml
```diff
@@ -17,9 +19,11 @@
         %td{:align => "right"}
           = item.quantity
         %td{:align => "right"}
-          = item.included_tax > 0 ? item.display_included_tax : ""
+          = item.single_display_amount_with_adjustments
         %td{:align => "right"}
           = item.display_amount_with_adjustments
+        %td{:align => "right"}
+          = display_line_item_tax_rates(item)
```
oeoeaio:
```ruby
- if @order.total_tax > 0
  %td{:align => "right"}
    = display_line_item_tax_rates(item)
```

oeoeaio:
Hey @ltrls, I've submitted a PR against your branch with a few tweaks to the layout on the invoice - let me know what you think.

app/models/spree/line_item_decorator.rb
enricostano: Would be nice to have some sort of documentation explaining this, something like the following:
```ruby
# @return [Array<String>]
def tax_rates
  return [] unless product.tax_category

  product.tax_category.tax_rates
end
```

lib/open_food_network/sales_tax_report.rb
ltrls: Awesome! I hadn't thought of doing it like that, it's much clearer!
I thought of selecting only rates in use because Myriam told there might be OFN instances having taxes for different countries but only relevant for a few hubs

lib/open_food_network/sales_tax_report.rb
myriamboure: @oeoeaio sorry I had not seen you pinged me, @ltrls just pointed it to me. Actually as you say for the moment it is super unlikely that an instance use both as the super admin has to choose if "taxes are included in price or not" for the whole instance. But we are anticipating for the tax overhaul :-) So it might be that the master price includes VAT for some products and that it does not for others. That doesn't matter for the tax display, we want the whole accumulated tax in the same column, we want for the order basically to know what is the total amount of VAT collected. Hope it's clear :) @ltrls let me know if you don't understand also :-)

oeoeaio:
Ok, I think this is looking pretty good. @sstead, ping me when you're ready to test...

oeoeaio:
Hey @ltrls, just FYI, I've just rebased this onto the latest master, and cleaned up the commit history a little.

sstead:
Testing notes...
1) the report is correctly capturing the tax on products. (see below). Note, it is handling base prices and inventory prices correctly.
2) It is correctly reporting the total tax included in the order.
2) However the report is not separating the shipping fee tax or enterprise fee tax into the correct 'tax type' column (see below)
image
3) Shipping fee tax rates are currently setup at the instance level (see below). So they might be difficult to filter into the correct tax type column of this report. 
image
4) Note that currently Payment Method Fees cannot be set to carry tax ([G#1469](https://github.com/openfoodfoundation/openfoodnetwork/issues/1469)). This will need to be corrected, and then this fee added to this report. 
5) The invoice page is not working - Rob says it's not setup to handle enterprise fees- when it tries to work out what the amount is, it doesn't realize it's working with an enterprise fee, not a tax rate, so gets an error.

app/models/spree/order_decorator.rb
```diff
@@ -227,6 +227,22 @@ def total_tax
...
   end
 
+  def tax_adjustments
+    adjustments.with_tax +
+      line_items.includes(:tax_adjustments).map {|li| li.tax_adjustments}.flatten
+  end
+
+  def tax_adjustment_totals
+    Hash[tax_adjustments.group_by(&:label).map do |label, adjustments|
+      [adjustments.first.originator, adjustments.sum(&:amount)]
```
oeoeaio: Ok, so there is a problem here, not your fault, it's because the way tax is stored is not particularly well designed...

`tax_adjustments` are all adjustments with a tax component. The adjustment `amount` may be all tax (eg. `originator_type: Spree::TaxRate`), or only partially tax (eg. `originator_type: EnterpriseFee` or `originator_type: Spree::ShippingMethod`). This method is treating them as though they are all `originator_type: Tax::Rate` adjustments. 

The problem is that even if we know the type of the originator, the way we currently calculate and store the tax component is different for each type of adjustable: EnterpriseFee adjustments *may* have a Spree::TaxRate, but it is not hard linked, because if the adjustment is on a LineItem, it may be inherited from a Product. ShippingMethod adjustments are completely different again, because they don't have a TaxRate at all, the rate is calculated via a system-wide percentage applied to shipments...

This is why we need to upgrade Spree... 😞

myriamboure: Hum... 
- Regarding point 4 in testing, does it mean we should add in the "shipping tax setting" the possibility to choose a tax category instead of putting an amount in the field? Is it something easy? I don't see another way to do actually if we want this to be displayed in this new report...
OR if there is any difficulty with that I'm wondering for our urgent case (the shop opens in a week) as we'll use this report for physical stores that don't have shipping then, can we leave like that and just add a mention "WARNING shipping fees VA not included"? Idea @oeoeaio ?

- Regarding the issue 6. Again, as a short term solution we created a new report and this report can be specifically mentioned as "physical shop tax reports without shipping fee and payment method fee". Because the store who needs it doesn't use any shipping fee or payment method fee, so there will be no adjustment issue. BUT there is still the issue that if I change a product and recalculate fees, the enterprise fees need to be recalculated and the corresponding VAT adjusted... hum... is there a solution / quick workaround here @oeoeaio or @ltrls 
I know we need to find out a solution as you said Rob but just looking for quick workaround ... @ltrls I captured for my own understand the three types of adjustements that can be made in an order below
screenshot from 2017-02-24 08-26-58

- For point 7, do you understand what to do @ltrls ? It's really not clear to me but  maybe it will speak to you...

So just to be clear my issue is that **the hub manage who is paying for having this adjusted open his shop early march**... I see there are here complex issues hard to solve quickly due to spree upgrade issues, but it would be great if we could have a short-term solution... so my question is: **is there short term solution or do I need to tell our userwho is paying for that we are not able to do it and he needs to buy another commercial shop management software?** Which is actually complicated because he wants to use the same system to manage his three pick-up points and his physical store so that he has stock consistency... Rrrrrhhhh...

ltrls:
Oh I had naively thought that taxes on adjustments were handled the same way as taxes on line items... Sorry for a mistake this huge (not hardly spottable at least! I just had to add fees...).
I have fixed it by calculating the taxe rate for adjustements via their amount and tax amount. As for the shipping method tax, the rate is calculated the same way but, (as I don't calculate it for all orders before declaring columns) it might not find itself a column to be in the tax report. To mitigate the problem I've eventually displayed all tax rates as columns, but aggregated by their rate, so that it won't display 2 taxes if there's one inclusive and one exclusive.
Phew! thank you for such nice reviewing anyway!

myriamboure:
Ping @oeoeaio :-) I hope this time will be the good one! Finger crossed :-o Apparently @ltrls did the same tests as @sstead to check that what he has done was ok...

sstead:
Regarding the Sales Tax report [1194](https://github.com/openfoodfoundation/openfoodnetwork/issues/1194)

1) **Shipping Fee tax.** 
- IF the shipping fee tax rate is not the same % as a configured tax rate, there is no column in the report for sales tax to be attributed to. The shipping fee tax is captured in the total, but not a separate column.
image
- IF the shipping fee tax rate is the same as a configured tax rate, it will be attributed to the correct column.
image

Rob suggests that IF the shipping fee tax % is NOT the same as a configured Tax Rate, THEN a new column should be added to the report.

2) **Display**
I'm not sure if this is handled in the instance's translations? It would be good to refer to 'tax' generally, rather than GST, which is only one tax rate?
image

3) **Things working correctly (just for the record)**
- Adjustments added/removed are correctly calculated and put in correct column and in the total. 
- If a new tax is added to the instance, another column is added to the report.
- If an order has zero tax for a certain tax rate, the column shows zero.
- The report works for both tax included in the price, and tax exclusive of the price (instance setting).

sstead:
Regarding the invoice [1195](https://github.com/openfoodfoundation/openfoodnetwork/issues/1195)

image

sstead:
Another thought... 
I can foresee that users will expect their logo and message to show in order confirmation emails... Is this something to address now, or put as a 'nice to have' for the future?
image

myriamboure:
**On tax report:**

- @sstead regarding the tax report, I think it's pretty fair to say that super-admin should have a tax rate set up that is the same to the shipping tax rate. Actually I don't think there can be a case where the shipping tax rate will not be an instance tax rate (I guess countries / states have different rates for different services / products, and transport is in one of those rates). That's why when discussing with @ltrls we said the case can't happen in real life and we said we can precise in the super-admin guide that he needs to be careful about that. Does it sounds ok @sstead and @oeoeaio ?

- If GST is only one tax rate, I guess yes, in French @ltrls the column "total excl GST" will be "Total HT" and the last column "Total TTC", which doesn't refer to any specific tax rate. I guess "Total excl GST ($)" and "Total incl GST ($)" can be strings translated in transifex?

- And there is the small "$" to remove in the last column content, yes. 

**On invoice:**

- I don't think we need to have the shipping address on the invoice the invoice is really just for accounting and legal aspects, so I suggest to just make appear on the invoice the label "Billing address" on top of the address

- regarding "GST", same point as for the tax report

- regarding breaking down taxes on the invoice, we do break it at the bottom on the totals, but the invoice will be impossible to read if we break down each item line for every tax rate. Unless there is any legal requirement in Australia that on an inoice you need to break down for each item each tax rate amount @sstead ? So for me no change on that.

- for the shipping line I guess if we suppose the shipping tax rate corresponds to a tax rate/category, then @ltrls you should be able to display it as @sstead suggested.

- seems ok to use the tax rate name, we will adjust the tax rate name so that they appear ok on the invoice.

- about adding the shipping method, I'm not sure we need it on the invoice... legally the invoice can't be given / sent before the products are delivered... well I guess if you want to add it that wont hurt, but as we are super short in time maybe for this first invoice iteration we can go without, and see if the need is voiced?

- nicely spot @sstead for the GST 10% line in case of "tax not included in price"

**On order confirmation:**

- I suggest we put that as "nice to have" as we are suuuper short in time, and also the customized text for invoice is more a legal stuff and won't be on the order confirmation (ex in France: VAT not applicable given law blablabla) So I think it needs some thinking connected to white labeling of order confirmation ;-)

@ltrls can you make those last adjustments today?

ltrls:
I have added the billing address mention, changed the locale to display "tax" instead of "GST", and fixed the adjustment display (and removed the $ sign in the report).
For the image on the invoice, I tried with a few images and had no problem, could you send me the one you're using ?
For the second screenshot of an invoice that you posted, we made the assumption with Myriam when we were designing the invoice that all taxes in an order (maybe even on a whole instance?) would either be all inclusive or all exclusive. So I considere an order to have inclusive tax if one of its line items have an inclusive tax. Anyway if the checkout has created a tax-exclusive adjustment, I don't think I should hide it and modify the corresponding line item displayed price for display coherence, right?
As for displaying the labels of the taxes, as I might have "manually" calculated the rate if they aren't attached to a tax record, my only option would be to pick for each tax rate the label of the first tax record found in database having the same rate. Should I do that?

myriamboure:
Regarding "inclusive" and "exclusive" tax rates, it's true that today with how the tax works, an instance has to choose if tax is included in price or not, and a hub cannot choose it the shopfront displays price inclusive or exclusive... it has to be the same rule for all the instance (unless all tax rates are duplicated, but anyway, that's not done nowhere as super not practical and has a lot of other issues...). So I think at that stage it's fair to think how @ltrls did... So the invoice show columns "incl. VAT" if the first item have an "included in price" tax type, and "excl. VAT" if the first item has a "tax excl. price", so basically if an instance has set up tax categories as excluded in tax, all items prices are tax excl., shopfronts show prices tax excl. and invoices as well. Else it's the contrary (choices made in France: tax include in price, so the invoice will be "incl. VAT" and then we have the breakdown of tax included at the end)
@sstead @oeoeaio is it a real life case to have a GST "included in price" and another "excluded" in the same invoice/hub/instance given the way tax is currently working? I don't think so but maybe you'll have another view on that...
I would consider this to be something we have anyway to rework on when we move on the tax overhaul, but it's fair enough for me to start with the current reasoning... 
... unless there is another way to code it which would make more sense @oeoeaio ? 

For info about the hub who needs the feature: we have 10 more days to upgrade the production French server so that he can use it... opening of the shop planned mid-march!!! So it would be great if he had time to test before using it for real... hope it's doable to finish all that quickly and merge so that we can deploy :-o

Thanks a lot all of you for the energy you put in this, tax stuff are a bit tricky ;-) @ltrls starting with that you'll be able to work on anything in OFN afterward, it will feel pretty easy I guess!

sstead:
I agree with both of your comments above Myriam. In response to this questions "is it a real life case to have a GST "included in price" and another "excluded" in the same invoice/hub/instance given the way tax is currently working?"... I don't think this will occur, as an instance will be consistent with either tax inclusive or tax exclusive settings.

app/models/spree/order_decorator.rb
```diff
@@ -227,6 +227,26 @@ def total_tax
...
+        tax_rate = adjustment.originator.amount
+      else
+        tax_rate = (adjustment.included_tax / (adjustment.amount - adjustment.included_tax)).round(2)
+      end
+      hash.update({tax_rate => adjustment.included_tax}) { |_tax_rate, amount1, amount2| amount1 + amount2 }
+    end
+  end
+
+  def has_taxes_included
+    not line_items.with_tax.empty?
```
oeoeaio: @ltrls, adjustments can also have tax applied, even if no line items have tax applied.

oeoeaio:
### Issues

> For the second screenshot of an invoice that you posted, we made the assumption with Myriam when we were designing the invoice that all taxes in an order (maybe even on a whole instance?) would either be all inclusive or all exclusive. So I consider an order to have inclusive tax if one of its line items have an inclusive tax.

While I agree that the likelihood of both included and excluded tax rates on a single instance seems unlikely at this stage, this solution makes me quite uncomfortable, because it will mean that our invoices could potentially show completely incorrect data (as Sally has demonstrated by using a mix of included and excluded tax rates). Further, while we can maybe the confident in the rates applied to products, I feel much less confident when we're talking about fees, shipments (and eventually payments). The shipping method tax-rate is currently assumed by the system to always be included in the amount, so as this PR currently stands any order without included tax on products with taxed shipping method will display the shipping amount as tax-exclusive when it is in fact tax-inclusive.

The main constraints we currently have (as I understand them) are as follows:

1. We want to report the amount of tax included at the line-item level (right @myriamboure?).

2. 'Excluded' tax is aggregated into one big lump for each order, so we can't itemise tax for items with an excluded taxes, without a massive amount of horrible hackery involving back-calculating tax from the tax-rate on the product, and taxes on any associated enterprise fees. This is not something that I am ok with.

3. Currently the "Tax Rate" column on the invoice only describes the tax on the product, but if an EnterpriseFee adjustment has been applied to a line item, it will be included in the "Unit Price" and "Total Price" columns. Therefore, reporting the tax as a rate is not really acceptable, unless we break each fee out into it's own line.

### What do we do?

To me, the combination of these three issues means that we have to choose one of the following options:

1. We report tax per-line item, meaning:
   - we have to drop support for excluded tax rates until we upgrade Spree. I don't think I am being drammatic, I can't see another solution... 
   - we should probably opt to report it as a value rather than a rate, because if we report it as a rate, it will inevitably end up being wrong in some cases

2. We don't report tax per line-item (anywhere!), meaning:
   - we can keep support for excluded tax rates.
   - we keep invoices more or less as they were (we can add the reporting of aggregated tax at the bottom though)

myriamboure:
@daniellemoorhead following your message on Slack we reaaaally need that to be merged for the 15th march release. The pull went back and forth for the last month but hopefully now we have something the is ok to be merged... the client who pays for it opens his physical shop in two weeks, and he can't open his shop without a functionnal invoicing / receipt system! I'll ping you on the second pull that we also need absolutely. Thanks :)

app/models/spree/order_decorator.rb
```diff
@@ -227,6 +227,26 @@ def total_tax
...
+  def tax_adjustment_totals
+    tax_adjustments.each_with_object(Hash.new) do |adjustment, hash|
+      if adjustment.originator_type == "Spree::TaxRate"
+        tax_rate = adjustment.originator.amount
+      else
+        tax_rate = (adjustment.included_tax / (adjustment.amount - adjustment.included_tax)).round(2)
+      end
+      hash.update({tax_rate => adjustment.included_tax}) { |_tax_rate, amount1, amount2| amount1 + amount2 }
+    end
+  end
```
enricostano: What about something like this? Maybe more readable and easier to test?
```ruby
def tax_rate_for(adjustment)
  return adjustment.originator.amount if adjustment.originator_type == "Spree::TaxRate"

  (adjustment.included_tax / (adjustment.amount - adjustment.included_tax)).round(2)
end

def tax_adjustment_totals
   tax_adjustments.each_with_object(Hash.new) do |adjustment, hash|
    tax_rate = tax_rate_for(adjustment)
    hash.update({ tax_rate => adjustment.included_tax }) { |_tax_rate, amount1, amount2| amount1 + amount2 }
  end
end
```

myriamboure:
@oeoeaio we need the amount of tax at each item level yes, this is a legal requirement, we have to do it.
About your options:
1- Legally we need in France to display on the line item what is the rate that apply for this line... we only have to tell the value at the invoice level, not by item, but under each item there is in small the VAT rate applying to this item. So we can't do it another way, the client who needs that is an official store, they have to comply with the law.
2- We can't keep the invoice as they are as it is a legal requirement to have all the information we put in it, we wouldn't do it if we were not obliged to do it...

I think we could modify a bit you option 1 by:
- We drop support for excluded tax rate until the Spree upgrade is finished (so if people want to use it they need to contribute to the Spree upgrade) 
and
- if you think it's too risky to display a rate and better to display a value at the item line... then I guess it's pretty easy from the value to recalculate the tax rate and display it, it's just an easy mathematical formula (rate = (item price incl VAT / (item price incl VAT - amount of VAT))-1) if I'm right. So we could display the tax rate as we want :-)

@oeoeaio @kirstenalarsen is it acceptable if we move forward that way?
@ltrls do you have any comment/suggestion to Rob's comment?
Can we finish that before the next release on Wednesday? Else what can I do? I feel super uncomfortable to be honest, and bit powerless... and I have that time pressure on my shoulder as the client paying for that opens his physical store and can't work without an invoicing system that is working... so I'm putting him in a difficult situation if we can't release that in the next week...

myriamboure:
As a complement @oeoeaio:
The invoice can be displayed:
-either with prices "incl. tax" (in the columns), and then at the end we see 1- Total incl tax, 2- breakdown of tax included 3- Total excl tax
-either with prices "excl. tax"(in the columns), and then at the end we see 1- Total excl. tax, 2- breakdown of tax to add 3- Total incl tax

We need that so that both BtoB (who talks "excl tax") and BtoC (who talks "incl tax") can display the invoice they need. This will only be possible when we have a button in the order cycle "display prices excl. or incl. VAT". **But on a shorter term, we need that already so that if an instance has chosen to set up their tax category as tax exclusive (like the US are going to do as they always talk excl. tax, or Canada) can print and send a consistent invoice**.

So that's why, to decide if the invoice is displayed "incl. tax" or "excl tax" @ltrls reads the first product, see if the tax category attached is set up as "VAT included or excluded" and from that deduce if the instance works with VAT categories that include or exclude VAT and display the good version of the invoice. **This is a temporary solution, a workaround we have found so that this invoice can work both for Australia and Canada and France**. Ideally this would be replaced by this checkbox at the OC level that says if the shopfront (and invoice) is displayed excl or incl tax (see tax overhaul)

This being said, this information is only used to choose if displays the "excl tax invoice version" or the "incl tax invoice version". It doesn't have any other impact in my understanding.

We need the VAT detail by product for any BtoB sales. For the first client who needs that in France he works with BtoC so we could potentially temporary delete the last column showing the tax rate for a line item if you make your veto on that (we will need to come back on that soon and add it again...) But that doesn't change anything actually so we don't understand ... we need to find a way to decide if we show the invoice excl or incl, and given the situation of how VAT is handled today, the situation that was tested by @sstead is "absurd" in the sense that it cannot happen in real life... I know ideally we don't want an absurd situation to be possible, but the problem is that the VAT is not properly set up, so we need to find workaround just to be able to simple use the platform today... 

@ltrls will comment more on that.

oeoeaio:
> if you think it's too risky to display a rate and better to display a value at the item line... then I guess it's pretty easy from the value to recalculate the tax rate and display it, it's just an easy mathematical formula (rate = (item price incl VAT / (item price incl VAT - amount of VAT))-1) if I'm right. So we could display the tax rate as we want :-)

Not so much risky as potentially just plain incorrect, I think? How can we display a single rate if the item is taxed at two different rates (eg. Canada)? Or if the item has one rate, but the price includes a fee that is charged at a different rate? 

I suppose the simplest solution would be to list multiple rates if they apply. Is that an acceptable solution for France? I don't think this will work for Australia, apparently we need to show how much tax is applied to each item ([see Sally's Post](https://community.openfoodnetwork.org/t/requirements-for-tax-invoices/315))

Another solution to this that Sally and I thought of is that we could potentially use that same format for the invoice as the `tax_rates` report that @ltrls has created, and display the rates and the correct amount attributable to each rate per line-item....

Something like: Item, Qty, Unit Price (excl tax), Rate 1 (X%), Rate 2 (Y%), Total (incl tax)
 
There are at least two problems with this suggestion (but maybe we can overcome them):
- columns could get a bit squishy if there are more than two different rates that apply to a single invoice, perhaps we could switch to landscape view in this case
- it still doesn't really work for excluded tax rates, because the information we need to display the breakdown isn't actually stored yet. We *could* back-calculate the amount of tax attributable in each column for each item, but this idea is what I called "horrible hackery" above. It makes me **VERY** uncomfortable, because have no way of knowing if the tax category on an item has been changed or removed since it was ordered, the same applies to fees. IMHO this leaves us with an impossible task if we want to create an consistently accurate invoice with excluded tax rates, until we upgrade Spree.

I am not comfortable making the call about what to do unilaterally, I think this needs to be answered by the community. Should we ask these questions of discourse?

### How to display per-item tax on the invoice?
1. Don't show it (not acceptable)
2. Show it as a single rate (incorrect when multiple rates or enterprise fees with different rates apply)
3. Show it as a list of multiple rates (need to work out if this is actually legal)
4. Show a column for each tax rate (IMHO this is what we have to do)

If we choose option 3, we then also need to use @myriamboure and @ltrls's convention that determines whether the order should be reported as tax-inclusive or tax-exclusive. Note that this will involve looking up rates *as they currently stand* when the invoice is generated, and may or may not be a true representation of the rates that applied when the order was processed. We need to decide whether we are happy with this risk (temporarily, seeing as it will be removed when we upgrade Spree).

If we choose option 4, we then need to decide

### How do we report per-item tax for excluded tax rates?
a. We temporarily drop support for excluded tax rates, until we upgrade Spree
b. We use horrible hackery to back-calculate it per item (not OK IMHO)
c. We attempt to back-port or replicate in some way the Spree logic that stores excluded tax in a non-aggregated way so that it can be reported in the same way an included tax rates are now (this is likely to be *very* non-trivial)

Note: If we choose 4a as I am suggesting, the question of whether to create a convention whereby invoices are reported as tax-inclusive or tax-exclusive becomes unnecessary.

oeoeaio:
@myriamboure @ltrls 

Ok, so I just talked to @kirstenalarsen and another solution could potentially be to toggle this new invoice with a feature toggle as a temporary solution, so that your client can launch with the invoice looking however he likes and using whatever conventions you deem fit, and the community can put off any big decisions like dropping support for exclusive tax rates, and just wait for the Spree and Tax overhauls...

kirstenalarsen:
Have just talked through this w Rob and I think 4A is the way to go to get
this through for you by weds. Then we need post on discourse to let
everyone know and find out ASAP if this temp solution is bearable or
actually stuffs someone up (like Theresa?)

If it is a problem, then we can put it behind a feature toggle so instances
can choose whether they have this or not. UNTIL spree upgrade when
everything will be beautiful.

Happy to talk anytime over the weekend if you need.

myriamboure:
Thank you @kirstenalarsen and @oeoeaio :-) 
Now I understand better yes your argument with the Canadian illustration, although it seems they usually communicate only on one rate including both rates, Theresa did validate that invoice layout ;-) 
But in theory it could happen that you have two rates and not show a general rate per item. I'm not talking about fees here as they are on separate lines.

The option 4 will look a bit strange for a French... so **I would vote for the toggle thing**, @ltrls what do you think? Let's talk asap to make sure we are aligned on that :-)

About invoice spec I had done that on community before we start actually developing: https://community.openfoodnetwork.org/t/invoice-internationalisation/753 and I thought we all agreed on that, but maybe it was not precise enough. I suggest we test with the toggle, when Theresa and others try it out, we can see if it works for them or not and adjust accordingly ;-)

Do I understand well the "toggle" notion, it would mean choose for the whole instance what version to use, or is it at the user level where he has to choose? Practically when I click on "print invoice" will I have to select in a list which version I want?

ltrls:
Hi @myriamboure, @kirstenalarsen and @oeoeaio :)

If the toggle thing is possible yeah I think it could allow us to work on these issues without being rushed. We need to work out the details (where do we choose which invoice) and I'll get into it. It seems like a nice temporary solution as those problems will get solved in a short time.

Let me try to recap what's blocking us.
We thought that for the reader of the invoice it's a nice thing to have all prices with or without tax, enforced by seeing ing it written somewhere (like I did in head of the column).
But without further changes we can't do that, as shipping fees (and additional adjustments created from the admin panel) don't increment the already existing "tax rate adjustment" and act with tax included.
In this case it may be better not to explicitly state that all prices are exclusive, even if line items' are.
This problem affects instances with price excluded taxes, and needs a change in shipment and late fee creation.

Additionnally, there's the problem of displaying tax rates.
It's trivial to get the TaxRate for a line item, via the Product relation.
It could be easy for enterprise fee adjustments too, if we didn't packed them in a "Admin & Handling" adjustment (is it possible?), via their EnterpriseFree (orignator) relation.
For shipping fee and late fee adjustments, it's easy now as they have the tax included but if it weren't the case there should be a way too. I guess you could choose a TaxCategory when editing a ShippingMethod (instead of the actual config parameter). For late fees, I don't know, maybe they should originate from one of the existing enterprise fees?

And then there's the problem of displaying taxes for a line item which has itself taxable adjustments.
The way we now display the price with adjustments I guess it's best to only show the product rate as it's probably the majority of the price, but we could also display the price without adjustments and display them with the other enterprise fees (like it's done in the show order admin panel, but grouped?)

myriamboure:
@kirstenalarsen and @oeoeaio we had a quick discussion with @ltrls and here is our proposal for the toggle idea.
This needs to be decided at the instance level as that's where the tax categories are decided, and also we want to avoid one more click for the hub user, it sounds unfriendly to ask him to choose between 2 models of invoices when he wants to print and send an invoice.

SO we propose _in "Configuration" to add anew sub-menu "Invoices"_ and in this sub-menu, offer **two options**:
**Option 1:** use the original invoice model which doesn't display the tax rate per item and which only shows a total "included tax" and a total "excluded tax" but no breakdown of aggregated tax per tax rate.
**Option 2:** use the invoice model that displays the tax rate per item, the total "included tax" and "excluded tax", and a breakdown per tax rate. Note: this model cannot be used by instances who have chosen to enter and show prices excluding tax.

**Depending on the choice, the original model or the new one will be used by the instance.**

**Does it sound ok to everyone?** We will learn from that and see while using it if we need to change something.
And in parallel, it would be great of course after the Spree upgrade to solve the remaining issues so that we can fix the problems mentioned by Rob and Pierre here...

Issue #1304, opened 2016-12-07
Order summary should be placed above 'place order now' button to encourage final review by user.


sstead:
Where: Checkout
Device: Mobile/Tablet
![image](https://cloud.githubusercontent.com/assets/10843056/20952928/feea273a-bc83-11e6-807a-ddc09a676a59.png)

pmackay:
I'm a bit concerned about just moving the whole order summary as-is above the Place order button. It has the Back to cart button which is larger. 

What about moving the Back to cart button to be on the left on the same line as the Place order button? And align Place order to the right? (for all layouts, not just mobile) That would:

* Mean the Place order button is aligned with all the Next buttons in each form section, which seems more consistent.
* The final button layout would follow this [wizard UX recommendation](http://ui-patterns.com/patterns/Wizard).

@sstead @lin-d-hop thoughts?

lin-d-hop:
Hi @pmackay Yep that was what I was envisaging. With both buttons appearing below the Order Summary.
Thanks for clarifying.

myriamboure:
Has this been covered by mobile UX overhaul work @kirstenalarsen ? I'm not sure, it doesn't seem mobile wheckout has already  been worked on, so shall we keep this one and try to find someone to do it by highliting this to hacktoberfest contributors?

Erioldoesdesign:
I don't think this is covered by mobile UX work so far but I disagree with having 50/50 buttons next to each other. Mobile is better if you stack buttons for accessibility and reducing error.

So:

[ Place order]

[< Back to Cart ]

leave at least 50px gap between two buttons so people don't accidentally hit one when they mean the other.

RachL:
I'm closing this one because I think this will be covered and better spec in the checkout split.

PR #1375, opened 2017-01-12
#1304: Rearrange checkout page layout

pmackay:
Fix for #1304 

Both buttons on bottom row.
Buttons below order summary on mobile view.

Here's how mobile looks:

screen shot 2017-01-12 at 12 21 33


lin-d-hop:
Looks great!

pmackay:
@oeoeaio @mkllnk I might need some input on this. How does https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/stylesheets/darkswarm/checkout.css.sass#L22 get used? One of the failing tests here is looking for `edit_order` but not sure how that would get on the page. The way I rearranged the page content to account for mobile layout looks to have had some side effects - my assumptions may not be valid :-/

app/views/checkout/_summary.html.haml
```diff
@@ -1,35 +1,29 @@
 %orderdetails
-  = form_for current_order, url: "#", html: {"ng-submit" => "purchase($event, checkout)"} do |f|
```
mkllnk: I guess this is the line that prints a from tag with class edit_order. I couldn't see that tag anywhere else in this diff. Is this what you were looking for?

pmackay: @mkllnk but does this actually still need to be a form? The main submit control was commented out back in [this commit](https://github.com/openfoodfoundation/openfoodnetwork/commit/d227fb24fe43a06d36343a56f75c6d06b897249e#diff-ef94bd3a5e3be73decff09fb413f1297). (unless I'm not understanding quite how this is working...)

mkllnk: Good question. I guess technically it doesn't need to be a form anymore. Is anything calling the submit action on the form?

So if it's not a form, you need to update the stylesheets and the test cases.

pmackay:
I have refactored my code to move the Login/Anonymous buttons form outside the main checkout form and this fixes the test failures. I think they were interfering. However this means the page layout (desktop) looks like this:

screen shot 2017-01-26 at 09 59 54

I could pull the buttons in to align with the central column of the checkout form, but the sidebar summary will still sit below the buttons. Is that OK?

mkllnk:
@sstead What do you think about the UX?

sstead:
I think Yuko created the initial #1304 to make sure customers reviewed their order before the clicked confirm. I suppose it's less critical on this screen that they review their order, as they'll have another chance in the next screen. So I think it's ok that the summary is off to the side. Below is the current view, note that there is no 'Place Order Now' button. It makes sense that this doesn't show, so I think remove it from the screenshot above. The back to cart at the left seems ok to me. Yuko is the best for UX, but I'm not sure what her availability is.
image

lin-d-hop:
@pmackay would you look at the test fail on this?
Also regarding UX my feeling is the option of being beside looks best on desktop. Does it fall to underneath on small screen views?

pmackay:
@lin-d-hop the problem is:

* To get the sidebar block to insert itself inbetween the checkform sections and final buttons, I had to make it part of the main checkout form.
* To get the sidebar block alongside the Login/Guest form, the Login/Guest form must be part of the main checkout form, which breaks it and causes failures. Hence I tried moving it out, but that results in it being above the checkout form (which includes the sidebar block).

So I'm a bit stuck on how to fix the problem raised in the original issue without causing a different UX issue. As mentioned, I wonder if the Login/Guest form could be a modal popover?

pmackay:
@oeoeaio @mkllnk guys do you have any suggestions how to solve the layout given the points above?

mkllnk:
> To get the sidebar block alongside the Login/Guest form, the Login/Guest form must be part of the main checkout form, which breaks it and causes failures.

Why does it break it? I thought it would just be a matter of updating the selectors in the spec to get it working again. Or is the actual checkout broken?

The discussion was about the layout (UX), the HTML structure and the test cases. So I'm not sure which part needs solving. I guess, if we know how it should look like, it should be possible to display it that way.

daniellemoorhead:
I pinged @yukoosawa on slack yesterday and she's going to have a look after work today. Cheers :)

yukoosawa:
Attaching very basic wireframes!

These don't take into consideration @pmackay's comments around login/guest form either causing failures (inside form) or being presented above the order summary (outside form).

The original issue listed in the UX review was only referring to the mobile view.

From a non-developer perspective I would be interested to know why the login/guest form causes failures inside the form - where as the logged in form (details, billing, shipping, payment) do not, given that from a div/layout perspective they're being injected into the same part of the page??

**Logged out**
desktop-logged out
mobile-logged out

**Logged in**
desktop-logged in
mobile-logged in

pmackay:
@yukoosawa sorry for the delay in replying on this. The issue this is trying to solve is how to position the sidebar Order summary in between the Payment and Back/Place order now buttons (which have been merged together in my recent changes). This is what you've drawn in images 3 and 4 above, but to get that to work I've made the whole page the form. 

Hence when the Login/guest checkout buttons are inserted into the form so they look like this:

screen shot 2017-03-07 at 16 50 07

It results in the checkout form showing this when one of those options is chosen:

screen shot 2017-03-07 at 16 50 19

Probably because its a nested form submit. So that doesnt work IMHO.

@oeoeaio @mkllnk any thoughts on how this might get arranged to achieve the mobile layout? Have you done anything fancier with insertion than just rearranging with Foundation grids?

pmackay:
@mkllnk @oeoeaio please could you review this again? I spent a bit more time reviewing how the forms work and refactored it some more. Thanks!

RohanM:
@pmackay I'm having a look at this and it seems the best way for me to check it out would be to load it up in the browser. There are a few conflicts and a couple of merge commits; would you be able to rebase this branch onto master to fix the conflicts and clean that up?

app/views/checkout/_summary.html.haml
RohanM: At first this seemed odd to me, but I did a bit of research and it seems like inputs outside a `<form>` are okay:

- https://stackoverflow.com/questions/3294572/is-input-well-formed-without-a-form
- https://stackoverflow.com/questions/8568660/is-there-any-danger-to-using-input-fields-outside-without-forms-in-html-javascri

daniellemoorhead:
@pmackay @lin-d-hop this PR has been hanging around for a fair while. I keep moving it to the next dot point release at the end of each  month. Any chance we could get the conflicts resolved and push it through for Oliver to test before next Friday?

pmackay:
@daniellemoorhead I cant this week, will try to update it next week if I can.

daniellemoorhead:
Next release is August 30th @pmackay - any chance of sorting it this month?

daniellemoorhead:
Closing for now as there's been no activity for a long time and we're github gardening. Feel free to reopen and recommence this work when you're back onboard :-)

#githubcull2018

PR #1393, opened 2017-01-20
Add initial stylelint configuration that produces no current warnings

pmackay:
@mkllnk Could you please comment on this as an approach? I've setup the config file with all the standard rules defined, and then commented out those that have issues in the current .scss files. As you suggested we could then gradually turn them on and tidy up the code.

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -1,3 +1,0 @@
-scss_files: 'app/assets/stylesheets/**/*.css.scss'
-
-exclude: 'app/assets/stylesheets/shared/**'
--- a/.stylelintrc.yml
+++ b/.stylelintrc.yml
@@ -1,0 +1,138 @@
+---
+# See https://github.com/stylelint/stylelint-config-standard/blob/master/index.js for these rules
+
+plugins:
+- stylelint-scss
+
+# Turn off with null value
+rules:
+  # at-rule-empty-line-before:
+  # - always
+  # - except:
+  #   - blockless-after-same-name-blockless
+  #   - first-nested
+  #   ignore:
+  #   - after-comment
+  at-rule-name-case: lower
+  at-rule-name-space-after: always-single-line
+  at-rule-semicolon-newline-after: always
+  block-closing-brace-empty-line-before: never
+  block-closing-brace-newline-after: always
+  block-closing-brace-newline-before: always-multi-line
+  block-closing-brace-space-before: always-single-line
+  # block-no-empty: true
+  block-opening-brace-newline-after: always-multi-line
+  block-opening-brace-space-after: always-single-line
+  # block-opening-brace-space-before: always
+  # color-hex-case: lower
+  # color-hex-length: short
+  color-no-invalid-hex: true
+  comment-empty-line-before:
+  - always
+  - except:
+    - first-nested
+    ignore:
+    - stylelint-commands
+  comment-no-empty: true
+  comment-whitespace-inside: always
+  custom-property-empty-line-before:
+  - always
+  - except:
+    - after-custom-property
+    - first-nested
+    ignore:
+    - after-comment
+    - inside-single-line-block
+  declaration-bang-space-after: never
+  declaration-bang-space-before: always
+  # declaration-block-no-duplicate-properties:
+  # - true
+  # - ignore:
+  #   - consecutive-duplicates-with-different-values
+  declaration-block-no-redundant-longhand-properties: true
+  # declaration-block-no-shorthand-property-overrides: true
+  declaration-block-semicolon-newline-after: always-multi-line
+  declaration-block-semicolon-space-after: always-single-line
+  declaration-block-semicolon-space-before: never
+  declaration-block-single-line-max-declarations: 1
+  # declaration-block-trailing-semicolon: always
+  # declaration-colon-newline-after: always-multi-line
+  # declaration-colon-space-after: always-single-line
+  declaration-colon-space-before: never
+  # declaration-empty-line-before:
+  # - always
+  # - except:
+  #   - after-declaration
+  #   - first-nested
+  #   ignore:
+  #   - after-comment
+  #   - inside-single-line-block
+  # font-family-no-duplicate-names: true
+  function-calc-no-unspaced-operator: true
+  function-comma-newline-after: always-multi-line
+  # function-comma-space-after: always-single-line
+  function-comma-space-before: never
+  function-linear-gradient-no-nonstandard-direction: true
+  function-max-empty-lines: 0
+  # function-name-case: lower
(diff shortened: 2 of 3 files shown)


pmackay:
I have the same version as you, 7.7.1. I should add stylelint to the list of dev packages in package.json though. But that wouldnt solve your issue. What version of npm do you have? Do you still get that error if you run stylelint directly? Note that all the rake task does is
```
sh "#{dir}/node_modules/stylelint/bin/stylelint.js #{dir}/app/assets/stylesheets/**/*.scss -f verbose"
```

I was thinking that having rake tasks for each of these tools allows them to be chained with a `rake check` or similar.

mkllnk:
I saw that the called script is for node. So I tried:
```sh
node node_modules/stylelint/bin/stylelint.js app/assets/stylesheets/**/*.scss -f verbose
```

Same error, node version v0.10.29, npm version 3.8.8. Debian Wheezy.

pmackay:
@mkllnk could you please try updating to node v4 or greater and test this again? I'm now on v6.10.0.

enricostano:
Just a quick side note. This development environment dependencies versions hell is why I stress a lot on having a consistent and replicable development environment across developers.

Sorry if I sound pedant 😃  I know that we don't have many resources to work on this kind of things... I'm working on a proposal to make this easier, hopefully. 💪

mkllnk:
Okay, I used node v6.10.1 and ran:

```sh
npm install stylelint-scss
```

Now it works. Just a note: Running `bundle exec rake` is a bit slow and needs the database available. It's much quicker and has fewer dependencies to run:

```sh
node node_modules/stylelint/bin/stylelint.js app/assets/stylesheets/**/*.scss -f verbose
```

pmackay:
@mkllnk great, so, could you please clarify, are there any changes you really want to make to this, or is it ok to merge? As mentioned, for me the idea is that a dev could run `rake lint` and it would run a set of lint tasks (HAML, SCSSS, Rubocop, etc). In CI builds we could call stylelint directly using node as you suggest.

mkllnk:
I'm happy with that approach. Let's ask @oeoeaio and @RohanM about their opinion.

oeoeaio:
@pmackay My only question would be why we wouldn't use CodeClimate for this...?

pmackay:
@oeoeaio an interesting question :) I think a conversation with @mkllnk some time back, he pointed out that CodeClimate checking was removed from the PR review checks (now there is just Travis) because it was annoying to keep getting failures, presumably because the configured checks (I think for Rubocop then) were too numerous. So this PR was a step based on my [improve CSS post](https://community.openfoodnetwork.org/t/some-ideas-for-improving-ofn-css/785) to add an improved linter. However I've approached this following @mkllnk's sugggestion:

>As always, it's good to start with really relaxed rules so that the linter is not annoying. Then we can become stricter over time.

So this config should **pass** and then we could add new linters after resolving all issues that cause failures. The important point (which I should have added on that Discourse post) is to **run the lint tool as an automated check on every PR, and failures must be fixed before its merged**. Otherwise its not worth bothering IMHO. Note that in general, there is a continuous slide in quality reported by CC (bar a curious spike around Feb 2016):

screen shot 2017-03-24 at 08 51 37

Why strive for this? The "Why does style matter?" section on the [HoundCI homepage](https://houndci.com/) captures it well. OFN has an ever growing team, and discussions about [expert reviewers](https://community.openfoodnetwork.org/t/proposal-pipeline-release-management-expert-reviewers-and-expert-testers/863) are very current. Manual reviews could be made shorter and more effective by using more automated reviewing.

So, reflecting on your question, I would agree it would be smart to use a separate tool, cos they are tailored to the job and will run in parallel to Travis. Two options:

* Use CodeClimate. Problem is it already checks for lots of stuff. I'd like to get to a "pass" state and resolve new failures, rather than always monitoring the 6,402 errors its currently reporting. So we could turn off all other tools and introduce stylelint with the config in this PR and the score will shoot up, but it checks less. Would that be ok?

* Use HoundCI for style code checking instead. It has less tool coverage than CC, so would have to use a scss-lint config, but that's not a big deal. 

My preference is probably for the first option but it would lose the measure of quality that CC captures now. However, if we're not actively using it anyway, does that matter much?

oeoeaio:
Cool. I don't want to make the final call, let ask for input from @RohanM @mkllnk @enricostano @stveep @Matt-Yorkley @pacodelaluna, @Em-AK etc.

I agree with you: having so many failures in CodeClimate is not at all useful, because we are so overwhelmed that we tend to just ignore what it is saying, resulting in the gradual slide in code quality you pointed out. It is my understanding that @RohanM was actually contacted by someone from CodeClimate who was interested in helping us get set up properly - is that right Rohan?

My preference would be the first option: to use CodeClimate for PRs, and massively reduce the number of tools being run so that we get a passing run on the current codebase. We could then come up with a roadmap for which tools we want to re-introduce, and we could break these off as individual issues that could be picked up by anyone who was interested in progressing this forward.

Thoughts?

enricostano:
CodeClimate 👍 

One thing we can plan too is to commit to solve existing offending issues and push towards a green / clean state. Nothing that can be done in 1 day of course... but I think that lots of the issues can be solved in batch.

Em-AK:
I like the proposed CodeClimate approach. The checks can also be run locally (in docker) with the codeclimate CLI https://github.com/codeclimate/codeclimate

mkllnk:
I agree that Code Climate would be ideal. But at the moment, we can't use it productively:

- It's too strict and we don't know how to change that (yet).
- Does running the tool locally in Docker mean you need root privileges?

This pull request would enable developers to easily run the style checks on their own machine. And it would enable us to include it in Travis. @pmackay Can you add it to `.travis.yml` so that we see it passing? It would also be interesting by how much it would increase the run time.

I would like to merge this soon. The decrease of code quality is worrying...

And once we get to the point to use Code Climate, we should be able to easily switch and drop this tool again.

enricostano:
> It's too strict and we don't know how to change that (yet).

Isn't this kind of configuration enough? https://docs.codeclimate.com/docs/scss-lint#section-configure-the-engine

oeoeaio:
Hey @pmackay, I've just made PR #1579 which is a tweaking of our Rubocop settings for CodeClimate, that follows a similar approach to the one you have taken here with stylelint. How would you feel about tweaking this so that it works with CodeClimate, rather than using a rake task? I think if people want to run stylelint on their local machine, that is fine, but I don't think we need to wrap that task in a rake task, when we have CodeClimate set up to run it for them. A lot of modern editors implement [linters for stylelint](https://atom.io/packages/linter-stylelint) too.

lin-d-hop:
Closing as we've gone for CodeClimate

Issue #1417, opened 2017-01-25
Numbers do not overwrite when shopping in Explorer


MagsHall:
On UK production, when shopping if I enter a quantity to purchase in the box the default 0 is not overwritten. I have to delete this 0 and enter the quantity I want, or risk purchasing a multiple of 10! 

This is only occurring in Windows Explorer 11 and not on other browsers. 

![numbers not overwriting](https://cloud.githubusercontent.com/assets/19837861/22297310/cb89465c-e314-11e6-9803-152e9c61aefb.png)

stveep:
This seems to be an incompatibility with Angular and IE11. I came up with a fix that just ensures the existing text is highlighted when you click inside the box. It looks a bit uglier but that can be fixed with some CSS targeted at these boxes, as long as this solves the original problem.

Please test that it doesn't break anything on other browsers.

[Aside] I think the original problem only occurs the first time a box is clicked. It might be due to the ng-pristine/untouched attributes in some way but I couldn't figure out exactly why.

MagsHall:
@stveep Are the changes on the staging server? I can't see any difference on production.

lin-d-hop:
@MagsHall If you need something on staging best to give @OliverUK a shout.

stveep:
I have just put it on staging.

MagsHall:
So all looks fine on staging when browsing in Chrome, but products are not loading on the shopfront when I try to browse in IE11

OliverUK:
I'm afraid I don't have IE, so can't help test this one.

enricostano:
Does OFN has a "kind of official" supported browsers list somewhere?

daniellemoorhead:
Reopened this @lin-d-hop as @NickWeir63 has offered to test on IE on the PR #1425. If you'd rather keep it closed then feel free to reclose this plus the PR.

>Does OFN has a "kind of official" supported browsers list somewhere?

@enricostano not as far as I'm aware.

RachL:
The new design has fixed this one 🎉

PR #1425, opened 2017-01-29
Uk/ie11 quantity fix

stveep:
…n incompatibility of IE11 and Angular.

Fix for #1417

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/select_on_focus.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/select_on_focus.js.coffee
@@ -1,0 +1,8 @@
+# This is a workaround for IE11, which doesn't delete placeholder values
+# correctly when an element is focused (when Angular is also used).
+Darkswarm.directive "ofnSelectOnFocus", () ->
+  restrict: 'A'
+  link: (scope, elem, attr) ->
+    elem.bind('focus', ->
+        elem.select()
+      )
--- a/app/assets/javascripts/templates/partials/shop_variant_no_group_buy.html.haml
+++ b/app/assets/javascripts/templates/partials/shop_variant_no_group_buy.html.haml
@@ -5,6 +5,7 @@
   value: nil,
   min: 0,
   placeholder: "0",
+  "ofn-select-on-focus" => true,
   "ofn-disable-scroll" => true,
   "ng-model" => "variant.line_item.quantity",
   "ofn-on-hand" => "{{variant.on_demand && 9999 || variant.count_on_hand }}",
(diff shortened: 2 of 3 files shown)


stveep:
Doing some tidying up - did this slip through the PR review net?

daniellemoorhead:
Sure did @stveep !

Would actually be great if you could get this staged and tested by @OliverUK? @lin-d-hop does that sound like the best next step given @sstead is climbing a mountain somewhere and isn't back until Sept 14?

OliverUK:
I'm afraid I don't have / don't use IE.

NickWeir63:
Hi @OliverUK I have IE set up as an alternative browser and am happy to test if you can tell me what to test and on which server

OliverUK:
@NickWeir63  I believe this is the related issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/1417
However I don't think there is a build to stage yet. If @stveep could make this appear on buildkite for UK dev, we could stage it and you could test it in IE11.

daniellemoorhead:
@sstead has IE @NickWeir63 @OliverUK so we can test it over here if it's easier?

sstead:
Testing notes:
The 0 is disappearing in IE now when you start typing :+1:

oeoeaio:
So @sstead says the functionality it right, but looking at this I am a little bit concerned about performance. We're adding an awful lot of event listeners to the shop page, which is already not the fastest. Any thoughts @stveep? Do you guys have a way of testing this against production data?

sauloperez:
Do we have any analytics data to base our decision on? If the IE11 user base in most of the instances is not significant is not worth optimizing the performance IMHO.

oeoeaio:
@sauloperez as it stands, any performance hit caused by this fix would be experienced by all users. It would not be restricted to only IE users, even though it is solving a UX issue only for them only. So it is a choice between a) ensuring that performance is not affected by this change (which it may not be) or b) not implementing the fix.

sauloperez:
Oops, I didn't notice that this applies to all users. Thanks for the heads-up. Totally agree with you.

daniellemoorhead:
Hey all, it's been a while since this was active. Given the solution will impact on performance and we have big performance issues it feels like we need to go back to the drawing board.

@stveep for the sake of kicking off the new year and trying to get our delivery pipe cleaned up I'm going to close this PR and put the issue back into the backlog. Feel free to reopen and continue work on it if you get the time. And also, happy new year to you and yours :-)

Ping @lin-d-hop for info

#githubcull2018

Issue #1430, opened 2017-02-07
Add Roo gem


lin-d-hop:
Find here: https://github.com/roo-rb/roo

Adds the ability to read in csv and other spreadsheet formats into Rails.

pmackay:
For what purpose? Shouldnt this be part of a feature story?

lin-d-hop:
Patience there cowboy. I'm making an epic as we speak.

lin-d-hop:
@Matt-Yorkley You need to add the Gemfile.lock for the deployment server.

_remote: You are trying to install in deployment mode after changing your Gemfile. Run `bundle install` elsewhere and add the updated Gemfile.lock to version control._

Matt-Yorkley:
Included in main feature.

PR #1448, opened 2017-02-08
Add roo gem

Matt-Yorkley:
Adding roo gem for #1430

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,7 @@ gem 'wkhtmltopdf-binary'
 
 gem 'foreigner'
 gem 'immigrant'
+gem 'roo', '~> 2.7.0'
 
 gem 'whenever', require: false
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -551,6 +551,9 @@ GEM
     roadie-rails (1.0.3)
       rails (>= 3.0, < 4.2)
       roadie (~> 3.0)
+    roo (2.7.1)
+      nokogiri (~> 1)
+      rubyzip (~> 1.1, < 2.0.0)
     rspec (2.14.1)
       rspec-core (~> 2.14.0)
       rspec-expectations (~> 2.14.0)
@@ -569,6 +572,7 @@ GEM
     rspec-retry (0.4.2)
       rspec-core
     ruby-progressbar (1.7.1)
+    rubyzip (1.2.0)
     safe_yaml (0.9.5)
     sass (3.3.14)
     sass-rails (3.2.6)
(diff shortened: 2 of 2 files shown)


pmackay:
Despite this being literally one line, I have a couple of questions:

* Does the version need to be set? Most gems dont have a version and that means they will update if `bundle update` is run. 
* Might it be appropriate to use `require: false` here, and load the gem explicitly if needed? Perhaps the import process might happen separately and could load the gem independently? This import feature is not likely to be used often (compared with daily commerce use).

Matt-Yorkley:
@pmackay I guess setting a version means you don't get unforeseen bugs arising from gem updates. 

I'm not up on dynamically loading gems... how about you take this on and I'll take that basket UI issue you were stuck on? Sound like a good trade? :)

pmackay:
I guess gem updates should get "better" over time, including new security fixes. But any process of running `bundle update` would need careful checking and testing. 

The dynamic loading thing really depends on how the whole feature would get built :-O I'm sure you'll have it whipped up in no time ;-)

enricostano:
Regarding the gem versioning I would even go for fixed version like `gem 'roo', '2.7.0'` since in my experience I've found much more useful to know the exact version I'm using.

I think it's our duty to keep an eye on security updates in used gems. There are even tools to raise an alert and tell developers that there is a new version covering security issues.

Running `bundle update` against a specific gem has always brougth problems in the projects where I've been working on. Running it against the whole `Gemfile` leads to disasters of course.

For me is a 👍 for `gem 'roo', '2.7.0'` or at least `gem 'roo', '~> 2.7.0'`

Sorry for jumping in like this but I just wanted to share my opinion/experience 😀

Matt-Yorkley:
I think having the ability to update it at a scheduled time is probably preferable to having unannounced changes to the codebase whenever the maintainer releases an update?

Matt-Yorkley:
@pmackay It looks like conditionally loaded gems stay loaded until the server is restarted anyway?

"Once a gem is loaded it's loaded until the server is stopped. By [including the gem conditionally], the first time this method is called it would load the gem, [leaving the app] in the same situation as if you were to load it normally. "

https://stackoverflow.com/questions/6086865/how-do-ruby-gems-load-in-rails#comment7058565_6087906

pmackay:
Sure, so it would only benefit if there is a way to run an import as a separate short-lived process. I'm not sure if there is good practice around that. Probably getting too detailed on this anyway....

Matt-Yorkley:
Closing to include this in main feature branch instead

Issue #1444, opened 2017-02-07
Product Import v1


lin-d-hop:
Epic to capture implementation of product import functionality.

Full phase 1 specification captured here: 
https://docs.google.com/document/d/1PB4OBpnhvINPXr4F2qfmUcjl8Ug3nyVh-6v7UokBNiI/edit#

Upon completion users will be able to import products and inventory from a data file for any enterprise they have appropriate permissions. Users will be able to create or update products/inventory from multiple suppliers within a single file, and to multiple inventories on a single import. Fresh imports can be viewed via an overview before import, then edited via Bulk Product Edit page or Inventory page after import. Users can use a range of default values on imports. Clear error messages will be communicated to users. 

Within the scope of this epic, improved image upload functionality will be added to Bulk Product Edit page.

IMPORTANT - Be sure to merge features as they are implemented. This feature should not be enabled in menus until completed. However it can be implemented and accessible via URL such that features can be continuously merged as we go.

Project board to follow the pipeline for that epic: https://github.com/openfoodfoundation/openfoodnetwork/projects/20

Matt-Yorkley:
Very handy Railscast on implementing CSV import with Roo gem: http://railscasts.com/episodes/396-importing-csv-and-excel

Matt-Yorkley:
Weight/volume based fields:

**unit_value** (_int_, example: 500)
**unit_description** (_str_, examples: "g", "kg", "T", "mL", "L")
**variant_unit** (_enum_, "weight"|"volume"|"items")
**variant_unit_scale** (_float_, examples: 0.001 for millimetres. Required if variant_unit = "weight" or "volume")
**display_as** (_str_, example: "500g". Used if variant_unit = "weight" or "volume". Usually null, unless necessary)
**variant_unit_name** (_str_, example: "large bunch". Required if variant_unit = "items")

Example **variant_unit_scale** values: 

grams (g) = 1
kilograms (kg) = 1000
tonnes (T) = 1000000
millilitres (mL) = 0.001
litres (L) = 1
kilolitres = 1000
litres = 1

As far as I can tell, the **unit_description** gets worked out based on **variant_unit_scale** and can be left empty...? And **display_as** can be left empty as well, by default it becomes the calculated **unit_value** multiplied by **unit_scale** plus the **unit_description** string, ie "500g", but it can be overridden if desired.

Matt-Yorkley:
[...]

PR #1451, opened 2017-02-13
Uk/import

Matt-Yorkley:
Product Import *second draft* for #1444. It'll need lots of reviewing, but it covers most of the spec. I'm posting this up to get some feedback and so people can build and play with it on uk testing.

**Covers:** importing various filetypes, handling upload validation, processing spreadsheet entries, pre-save feedback on valid/invalid entries and how many products are to be added/updated, saving all valid entries, import products into multiple enterprises from one spreadsheet (and validate user permissions for each enterprise), extended specs and pretty UI.

**Missing:** import timestamps functionality, translations, inventory/variant overrides functionality.

screenshot from 2017-02-09 11-28-17

screenshot from 2017-03-02 16-40-09
screenshot from 2017-03-02 16-40-18
screenshot from 2017-02-13 11-36-47
screenshot from 2017-02-13 11-43-14

### DIFF
--- a/app/controllers/admin/product_import_controller.rb
+++ b/app/controllers/admin/product_import_controller.rb
@@ -1,0 +1,52 @@
+require 'roo'
+
+class Admin::ProductImportController < Spree::Admin::BaseController
+
+  before_filter :check_upload, except: :index
+
+  def import
+    # Save uploaded file to tmp directory
+    @filepath = save_upload(params[:file])
+    @importer = ProductImporter.new(File.new(@filepath), editable_enterprises)
+
+    if @importer.errors.present?
+      flash[:notice] = @importer.errors.full_messages.to_sentence
+    end
+  end
+
+  def save
+    file = File.new(params[:filepath])
+    @importer = ProductImporter.new(file, editable_enterprises)
+    @importer.save_all_valid
+
+    if @importer.updated_count && @importer.updated_count > 0
+      File.delete(file)
+      flash[:success] = "#{@importer.updated_count} records updated successfully"
+    else
+      flash[:notice] = @importer.errors.full_messages.to_sentence
+    end
+  end
+
+  private
+
+  def check_upload
+    unless params[:file] || (params[:filepath] && File.exist?(params[:filepath]))
+      redirect_to '/admin/product_import', :notice => 'File not found or could not be opened'
+      return
+    end
+  end
+
+  def save_upload(upload)
+    filename = Time.now.strftime('%d-%m-%Y-%H-%M-%S')
+    extension = '.' + upload.original_filename.split('.').last
+    File.open(Rails.root.join('tmp', filename+extension), 'wb') do |f|
+      f.write(upload.read)
+      f.path
+    end
+  end
+
+  # Define custom model class for Cancan permissions
+  def model_class
+    ProductImporter
+  end
+end
--- a/app/controllers/admin/product_import_controller.rb
+++ b/app/controllers/admin/product_import_controller.rb
@@ -2,44 +2,54 @@ require 'roo'
 
 class Admin::ProductImportController < Spree::Admin::BaseController
 
-  before_filter :check_upload, except: :index
+  before_filter :validate_upload_presence, except: :index
 
   def import
     # Save uploaded file to tmp directory
-    @filepath = save_upload(params[:file])
+    @filepath = save_uploaded_file(params[:file])
     @importer = ProductImporter.new(File.new(@filepath), editable_enterprises)
 
-    if @importer.errors.present?
-      flash[:notice] = @importer.errors.full_messages.to_sentence
-    end
+    check_file_errors @importer
+    check_spreadsheet_has_data @importer
+
+    @tax_categories = Spree::TaxCategory.order('is_default DESC, name ASC')
+    @shipping_categories = Spree::ShippingCategory.order('name ASC')
   end
 
   def save
-    file = File.new(params[:filepath])
-    @importer = ProductImporter.new(file, editable_enterprises)
-    @importer.save_all_valid
-
-    if @importer.updated_count && @importer.updated_count > 0
-      File.delete(file)
-      flash[:success] = "#{@importer.updated_count} records updated successfully"
-    else
-      flash[:notice] = @importer.errors.full_messages.to_sentence
-    end
+    @importer = ProductImporter.new(File.new(params[:filepath]), editable_enterprises)
+    @importer.save_all if @importer.has_valid_entries?
   end
 
   private
 
-  def check_upload
+  def validate_upload_presence
     unless params[:file] || (params[:filepath] && File.exist?(params[:filepath]))
-      redirect_to '/admin/product_import', :notice => 'File not found or could not be opened'
+      redirect_to '/admin/product_import', notice: 'File not found or could not be opened'
       return
     end
   end
 
-  def save_upload(upload)
-    filename = Time.now.strftime('%d-%m-%Y-%H-%M-%S')
+  def check_file_errors(importer)
+    if importer.errors.present?
+      redirect_to '/admin/product_import', notice: @importer.errors.full_messages.to_sentence
+      return
+    end
+  end
+
+  def check_spreadsheet_has_data(importer)
+    unless importer.item_count
+      redirect_to '/admin/product_import', notice: 'No data found in spreadsheet'
+      return
+    end
+  end
+
+  def save_uploaded_file(upload)
+    filename = 'import' + Time.now.strftime('%d-%m-%Y-%H-%M-%S')
     extension = '.' + upload.original_filename.split('.').last
-    File.open(Rails.root.join('tmp', filename+extension), 'wb') do |f|
+    directory = 'tmp/product_import'
+    Dir.mkdir(directory) unless File.exists?(directory)
+    File.open(Rails.root.join(directory, filename+extension), 'wb') do |f|
       f.write(upload.read)
       f.path
     end
--- a/app/controllers/admin/product_import_controller.rb
+++ b/app/controllers/admin/product_import_controller.rb
@@ -17,7 +17,7 @@ class Admin::ProductImportController < Spree::Admin::BaseController
   end
 
   def save
-    @importer = ProductImporter.new(File.new(params[:filepath]), editable_enterprises)
+    @importer = ProductImporter.new(File.new(params[:filepath]), editable_enterprises, params[:settings])
     @importer.save_all if @importer.has_valid_entries?
   end
 
--- a/app/models/product_importer.rb
+++ b/app/models/product_importer.rb
@@ -11,6 +11,7 @@ class ProductImporter
     if file.is_a?(File)
       @file = file
       @sheet = open_spreadsheet
+      @entries = []
       @valid_entries = {}
       @invalid_entries = {}
 
@@ -26,12 +27,11 @@ class ProductImporter
       @editable_enterprises = {}
       editable_enterprises.map { |e| @editable_enterprises[e.name] = e.id }
 
-      @non_display_attributes = 'id', 'product_id', 'variant_id', 'supplier_id', 'primary_taxon_id', 'category_id', 'shipping_category_id', 'tax_category_id', 'on_hand_nil'
       @total_supplier_products = 0
       @products_to_reset = {}
       @updated_ids = []
 
-      validate_all if @sheet
+      init_product_importer if @sheet
     else
       self.errors.add(:importer, 'error: no file uploaded')
     end
@@ -78,7 +78,7 @@ class ProductImporter
   end
 
   def suppliers_index
-    index = @suppliers_index || get_suppliers_index
+    index = @suppliers_index || build_suppliers_index
     index.sort_by{ |k,v| v.to_i }.reverse.to_h
   end
 
@@ -87,27 +87,15 @@ class ProductImporter
   end
 
   def invalid_entries
-    entries = {}
-    @invalid_entries.each do |line_number, data|
-      entries[line_number] = {entry: data[:entry].except(*@non_display_attributes), errors: data[:errors]}
-    end
-    entries
+    @invalid_entries
   end
 
   def products_to_create
-    entries = {}
-    @products_to_create.merge(@variants_to_create).each do |line_number, data|
-      entries[line_number] = {entry: data[:entry].except(*@non_display_attributes)}
-    end
-    entries
+    @products_to_create.merge(@variants_to_create)
   end
 
   def products_to_update
-    entries = {}
-    @variants_to_update.each do |line_number, data|
-      entries[line_number] = {entry: data[:entry].except(*@non_display_attributes)}
-    end
-    entries
+    @variants_to_update
   end
 
   def products_created_count
@@ -141,6 +129,13 @@ class ProductImporter
 
   private
 
+  def init_product_importer
+    build_entries
+    build_categories_index
+    build_suppliers_index
+    validate_all
+  end
+
   def open_spreadsheet
     if accepted_mimetype
       Roo::Spreadsheet.open(@file, extension: accepted_mimetype)
@@ -166,25 +161,26 @@ class ProductImporter
     end
   end
 
-  def entries
-    rows.map do |row|
-      Hash[[headers, row].transpose]
+  def build_entries
+    rows.each_with_index do |row, i|
+      row_data = Hash[[headers, row].transpose]
+      entry = SpreadsheetEntry.new(row_data)
+      entry.line_number = i+2
+      @entries.push entry
     end
+    @entries
   end
 
   def validate_all
-    entries.each_with_index do |entry, i|
-      line_number = i+2 # Roo counts "line 2" as the first line of data
+    @entries.each do |entry|
+      supplier_validation(entry)
+      category_validation(entry)
+      set_update_status(entry)
 
-      supplier_validation(line_number, entry)
-      category_validation(line_number, entry)
-      set_update_status(line_number, entry)
-
-      mark_as_valid(line_number, entry) unless entry_invalid?(line_number)
+      mark_as_valid(entry) unless entry_invalid?(entry.line_number)
     end
 
     count_existing_products
-
     delete_uploaded_file if item_count.zero? or valid_count.zero?
   end
 
@@ -212,49 +208,43 @@ class ProductImporter
     !!@invalid_entries[line_number]
   end
 
-  def supplier_validation(line_number, entry)
-    suppliers_index = @suppliers_index || get_suppliers_index
-    supplier_name = entry['supplier']
+  def supplier_validation(entry)
+    supplier_name = entry.supplier
 
     if supplier_name.blank?
-      mark_as_invalid(line_number, entry, "Supplier name field is empty")
-      entry['supplier_id'] = Enterprise.first.id # Removes a duplicate validation message TODO: proper solution
+      mark_as_invalid(entry, attribute: "supplier", error: "can't be blank")
       return
     end
 
     unless supplier_exists?(supplier_name)
-      mark_as_invalid(line_number, entry, "Supplier \"#{supplier_name}\" not found in database")
-      entry['supplier_id'] = Enterprise.first.id # Removes a duplicate validation message TODO: proper solution
+      mark_as_invalid(entry, attribute: "supplier", error: "\"#{supplier_name}\" not found in database")
       return
     end
 
     unless permission_by_name?(supplier_name)
-      mark_as_invalid(line_number, entry, "You do not have permission to manage products for \"#{supplier_name}\"")
-      entry['supplier_id'] = Enterprise.first.id # Removes a duplicate validation message TODO: proper solution
+      mark_as_invalid(entry, attribute: "supplier", error: "\"#{supplier_name}\": you do not have permission to manage products for this enterprise")
       return
     end
 
-    entry['supplier_id'] = suppliers_index[supplier_name]
+    entry.supplier_id = @suppliers_index[supplier_name]
   end
 
   def supplier_exists?(supplier_name)
     @suppliers_index[supplier_name]
   end
 
-  def category_validation(line_number, entry)
-    categories_index = @categories_index || get_categories_index
-    category_name = entry['category']
+  def category_validation(entry)
+    category_name = entry.category
 
     if category_name.blank?
-      mark_as_invalid(line_number, entry, "Category field is empty")
-      entry['primary_taxon_id'] = Spree::Taxon.first.id # Removes a duplicate validation message TODO: proper solution
+      mark_as_invalid(entry, attribute: "category", error: "can't be blank")
       return
     end
 
     if category_exists?(category_name)
-      entry['primary_taxon_id'] = categories_index[category_name]
+      entry.primary_taxon_id = @categories_index[category_name]
     else
-      mark_as_invalid(line_number, entry, "Category \"#{category_name}\" not found in database")
+      mark_as_invalid(entry, attribute: "category", error: "\"#{category_name}\" not found in database")
     end
   end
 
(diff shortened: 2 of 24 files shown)


lin-d-hop:
Great, thanks!! Will have a look and start to give some feedback from a functional perspective :-)

lin-d-hop:
Maybe @stveep would be up for giving some technical feedback?

Matt-Yorkley:
This might come in handy: 
[Example CSV.zip](https://github.com/openfoodfoundation/openfoodnetwork/files/770931/Example.CSV.zip)

app/controllers/admin/product_import_controller.rb
```diff
@@ -0,0 +1,69 @@
+require 'roo'
+
+class Admin::ProductImportController < Spree::Admin::BaseController
+
+  def import
+    unless params[:file]
+      redirect_to '/admin/product_import', :notice => 'File not found or could not be opened'
+      return
+    end
```
enricostano: Maybe we can move this in a `before_filter`, just to keep the controller's action as clean as possible. What do you think?

app/controllers/admin/product_import_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      flash[:notice] = importer.errors.full_messages.to_sentence
+    end
+
+  end
+
+  def save
+    unless File.exist?(params[:filepath])
+      redirect_to '/admin/product_import', :notice => 'File not found or could not be opened'
+      return
+    end
```
enricostano: Same here, move to a `before_filter` maybe

app/controllers/admin/product_import_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    @valid_entries = importer.valid_entries
+    @invalid_entries = importer.invalid_entries
+
+    @valid_count = importer.valid_count
+    @invalid_count = importer.invalid_count
+    @has_valid_entries = importer.has_valid_entries?
+
+    @products_to_create = importer.products_create_count
+    @products_to_update = importer.products_update_count
```
enricostano: Is there any reason we can't just pass `@importer` instance to the view? In that way you don't have to instantiate lots of instance variables to hold `importer`'s state.

app/controllers/admin/product_import_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    filepath = params[:filepath]
+    file = File.new(filepath)
+
+    importer = ProductImporter.new(file, editable_enterprises)
+    importer.save_all_valid
+
+    @products_created = importer.products_created
+    @products_updated = importer.products_updated
+    @updated_count = importer.updated_count
```
enricostano: Same thing here, maybe we can just pass `@importer` to the view.

Matt-Yorkley:
There are a few bits that need thinking about, like currently if you're adding a new product then a product description can be given, but if you're updating an existing product or adding something that's a variant of another product (like "blackberry" as a variant of "fruit jams", where "strawberry" and "raspberry" already exist) then the product description gets ignored. This can easily be changed to allow it to be updated, but it potentially creates a weird situation from the users' perspective, as the description for strawberry jam would overwrite the master description for all fruit jams, but that's not necessarily clear when you have each variant on it's own line in the spreadsheet with it's own description column...

app/controllers/admin/product_import_controller.rb
Matt-Yorkley: I had these in for debugging purposes whilst sorting out the model and controller, but yes, one of the next jobs is to remake all the views with a much better UI.

Matt-Yorkley:
Another thing is using IDs. The main functionality doesn't use ID's for matching things like Category, Supplier and Product. If the user supplies ID's for these things, should they be used, or ignored? Currently they are ignored for consistency. Also, "products" and "variants" have different sets of IDs as they are in different tables, which might make things complicated...?

lin-d-hop:
Category: There is no easy way to match supplier category ID to OFN category ID. These also differ between instances. The best way to handle this might be macros for supplier data files that do the translation before the importer. In the longer term having a tool in OFN that lets suppliers save these translations internally will be even easier.... but not a phase one feature.

Product IDs: A suppliers product ID is likely to map to SKU in OFN?

Supplier IDs: Currently it is an overkill for most enterprises to have a supplier ID themselves for their supppliers. However a similar SKU field on mapping suppliers together might be useful.

Obviously we will be generating our own IDs on table entries dynamically for new entries, and not changing IDs when we look up existing entries. A suppliers ID is not useful as an ID to us but might be in SKU or mapping appropriately. 

Is that helpful?

Matt-Yorkley:
I meant if they were referencing our internal ids, not theirs. Like if a user had exported some product data that included `supplier_id` and `primary_taxon_id` values for supplier and category, would we want to ignore the ids in these fields whilst importing and stick to the current method of expecting a string for `supplier` and `category` and then looking up the correct ids from the respective tables? It's not a pressing question, just needs considering.

I guess it depends on how these fields will be exported, if we're going to have a corresponding CSV/spreadsheet export function. It probably makes sense to not export ids, so that the data is human readable, in which case ignoring ids for these fields on import is the sensible way to go...?

lin-d-hop:
I would agree. Human readable data in exporters and thus no way of using our db ids in importers.

There might be a case as some point to do otherwise, but i think we'll have an easier time of it all if we can avoid using db IDs outside the db.

lin-d-hop:
Just a note, as well, that we want to be able to review successful imports (creates and updates) as well before we press import. Successes might also not be right, if valid data but the spreadsheet has errors.

Matt-Yorkley:
That's already covered, if I catch your drift. 

So after the uploading the spreadsheet there's validation checks before save that include creating relevant Spree::Product and Spree::Variant objects for each item (without saving them) and applying the `.validate? ` method on each object. The various results (including the objects' validation errors) are displayed as part of the initial validation. That should give an accurate picture of which ones are valid and can/will be successfully saved/updated. After that when the the data has been reviewed by the user and they click save and the items are _actually_ saved, there's also feedback on which items _were_ saved/updated.

Matt-Yorkley:
@mkllnk have you got time to do some code review for this PR?

app/models/product_importer.rb
```diff
@@ -0,0 +1,314 @@
...
+    end
+  end
+
+  def entries
+    rows.map do |row|
+      Hash[[headers, row].transpose]
+    end
+  end
+
+  def validate_all
```
pmackay: Just a thought, what about to use https://github.com/theodi/csvlint.rb here to validate the file against a CSV schema? (which we could define, and would allow other tools and so on to be used)

app/views/admin/product_import/import.html.haml
```diff
@@ -0,0 +1,67 @@
+- content_for :page_title do
+  Product Import
+
+= render :partial => 'spree/admin/shared/product_sub_menu'
+
+- if @importer.valid_count && @importer.invalid_count
+  %h5 Products imported from spreadsheet:
+  %br
+  %p{style: "font-size: 1.15em"}
```
pmackay: Why have font styles in HAML?

app/views/admin/product_import/import.html.haml
```diff
@@ -0,0 +1,67 @@
...
+        Item line #{line}:
+      %span= item[:entry]['name']
+      - if item[:entry]['display_name']
+        ( #{item[:entry]['display_name']} )
+    - item[:errors].each do |error|
+      %p{class: "red"}
+        &nbsp;-&nbsp; #{error}
+  %br
+  %h5 Review invalid entries:
+  %div{style: 'width: 100%; overflow: auto;'}
```
pmackay: Suggest all styles should live in .scss files.

app/views/admin/product_import/import.html.haml
```diff
@@ -0,0 +1,67 @@
...
+      - @importer.invalid_entries.each do |line, item|
+        %tr
+          %td= line
+          - item[:entry].each do |key, value|
+            %td{style: 'white-space: nowrap;'}= value
+
+- if @importer.has_valid_entries?
+  %br
+  - if @importer.invalid_count > 0
+    %h5 Invalid entries detected.
```
pmackay: do these and other strings need moving to locale files?

Matt-Yorkley: @pmackay The views are temporary for testing the functionality. Making a nice UI and adding translations are on the to-do list.

oeoeaio:
@Matt-Yorkley looking bloody awesome by the way, nice work!

app/models/product_importer.rb
```diff
@@ -0,0 +1,314 @@
...
+
+      # Fetch/assign and validate supplier id
+      supplier_name = entry['supplier']
+      if supplier_name.blank?
+        invalidate({i+2 => {entry: entry, errors: ["Supplier name field is empty"]}}) #unless entry['supplier_id']
+      else
+        if suppliers_index[supplier_name]
+          entry['supplier_id'] = suppliers_index[supplier_name]
+        else
+          invalidate({i+2 => {entry: entry, errors: ["Supplier \"#{supplier_name}\" not found in database"]}})
```
Matt-Yorkley: Yeah this is something I thought needed tidying up. It's because Roo counts the rows starting from 1, but the first line of actual data is on line 2 (because of the headers), whereas Ruby obviously counts array items from 0.

app/models/product_importer.rb
```diff
@@ -0,0 +1,314 @@
+require 'roo'
+
+class ProductImporter
+  extend ActiveModel::Naming
+  include ActiveModel::Conversion
+  include ActiveModel::Validations
```
Matt-Yorkley: There are a couple of places where they get used. I could maybe strip them out though.

app/models/spree/ability_decorator.rb
```diff
@@ -155,6 +155,8 @@ def add_product_management_abilities(user)
     can [:admin, :index, :read, :search], Spree::Taxon
     can [:admin, :index, :read, :create, :edit], Spree::Classification
 
+    can [:admin, :index, :import, :save], ProductImporter
```
Matt-Yorkley: @oeoeaio I think any users that are able to manage products will be able to use the product import, yes. There are checks by the importer on a per-item basis to see if the user has permission to add/edit products for the enterprise they are trying to import into.

app/models/product_importer.rb
```diff
@@ -0,0 +1,314 @@
...
+    if self.respond_to?(method, include_private=true)
+      @sheet ? self.send(method, *args, &block) : nil
+    else
+      super
+    end
+  end
+
+  private
+
+  def open_spreadsheet
```
Matt-Yorkley: @oeoeaio What do you think about this method_missing section? I couldn't tell if it was a genius metaprogramming solution or some sort of anti-pattern monstrosity. It solves a very specific problem quite well. Thanks for the feedback by the way! :smile:

app/models/spree/ability_decorator.rb
Matt-Yorkley: That's including E2E permissions and updating items for multiple enterprises from one spreadsheet. If they don't have the permissions it doesn't update that product (and explains why in the pre-save feedback, with line number etc).

app/models/product_importer.rb
```diff
@@ -0,0 +1,314 @@
...
+      end
+
+      # Ensure on_hand isn't nil (because Spree::Product and Spree::Variant each validate this differently)
+      entry['on_hand'] = 0 if entry['on_hand'].nil?
+
+      # Check if entry can be updated/saved; assign updatable status
+      set_update_status(entry, i)
+
+      # Add valid entry
+      @valid_entries[i+2] = {entry: entry} unless @invalid_entries[i+2]
```
enricostano: Yes, much more readable. Another option to make it even more readable IMHO would be using "early returns". For instance:
```ruby
def mark_as_valid(row_number, entry)
  return if entry_invalid?(row_number)

  add_valid_entry(row_number, entry)
end
```

app/models/product_importer.rb
```diff
@@ -0,0 +1,314 @@
...
+    if accepted_mimetype
+      @sheet = Roo::Spreadsheet.open(@file, extension: accepted_mimetype)
+    else
+      self.errors.add(:importer, "could not proccess file: invalid filetype")
+      nil
+    end
+  end
+
+  def accepted_mimetype
+    File.extname(@file.path).in?(['.csv', '.xls', '.xlsx', '.ods']) ? @file.path.split('.').last.to_sym : false
```
enricostano: What about using exceptions to manage workflows? You would avoid a lot of `nil`, `false` and `if` IMHO

You could raise a custom `NotSupportedMimetypeError`, rescue that specific error and add the `self.errors.add()`

app/controllers/admin/product_import_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    @filepath = save_upload(params[:file])
+    @importer = ProductImporter.new(File.new(@filepath), editable_enterprises)
+
+    if @importer.errors.present?
+      flash[:notice] = @importer.errors.full_messages.to_sentence
+    end
+  end
+
+  def save
+    file = File.new(params[:filepath])
```
Matt-Yorkley: @oeoeaio Agreed, passing in the filepath is not good. Not sure a new model is needed, but I'll take another look at it. Also, the imported file gets deleted once the data is saved successfully, but if the user imports a spreadsheet then leaves without completing it the file stays on the server. Currently it doesn't get cleaned up or deleted, which could lead to junk spreadsheets sitting around in the filesystem...

oeoeaio: Can always create a cleanup job using `whenever` that runs each night/week/month...

app/models/product_importer.rb
oeoeaio: No, all good - just checking

nanoharsha:
@Matt-Yorkley. Nice work on this development. 
I came across this Blog post "[Import-product-data-spree-commerce](http://vinsol.com/blog/2016/10/20/import-product-data-spree-commerce/)" , and wanted to check with you if any of the work which is put in there could be useful for our development.
The github repo for same is [here](https://github.com/autotelik/datashift_spree).
They also have a ready Heroku demo setup with below details, if you wanna quickly check with details below. 

    - Heroku Demo with Sample CSV file Import
    - Find Product Import tab under Products in Left Sidebar
    - Admin Username: pikender@vinsol.com
    - Admin Password: vinsol

In the same stretch, I came across this [blogpost](http://vinsol.com/spreecommerce-digital-asset-management) where they describe about "Digital-Asset-Management" tool to upload images and other digital assets and map it later to products. I found it a little more streamlined for Textile shopping, as it had fair number of steps etc, but just wanted to share here, so that if we could improvise and use, the code is all opensource and available.

Cheers

app/models/product_importer.rb
oeoeaio: Hm, interesting approach. I can't see anything wrong with it, looks like you've covered off all of the bases.... 👍

oeoeaio: Actually, can you point me to a practical example of where this is used? Just trying to work out if nil is a good value to return, or if an error should really be raised... Also, the more I think about it, the more it seems like a bit of an odd convention - if I come back in 3 years and try and work out how to call these private methods, am I going to work out easily that I should be calling them as public methods, or will I just use `importer.send(:private_method)`?

oeoeaio: It's a clever pattern but I feel like this is much easier to understand:

```ruby
private

def method
  return nil unless @sheet # or raise Error unless @sheet
  # Do the thing
end

Comment (oeoeaio) [inline app/models/product_importer.rb]: Not sure about the use case so this may not be relevant, but I don't think method_missing is hit when a private method is called from the object itself, so the check won't work when private methods are called internally... I could be wrong about that though...

Comment (Matt-Yorkley) [inline app/models/product_importer.rb]: @oeoeaio Basically, the alternative was to put a line like `return nil unless @sheet` above every single method, which seemed a bit messy. All the methods work if the spreadsheet has loaded succesfully, or raise a fatal error if not, and from the controller's point of view it wants to call methods on the object but doesn't know if the spreadsheet is valid until it tries.

Comment (Matt-Yorkley) [inline app/models/product_importer.rb]: It all works ok if the methods are being called internally.

**Comment by Matt-Yorkley** [comment]
Hi @nanoharsha that looks interesting. We've based our import on Roo, which seems to be a bit more flexible with filetypes and CSV delimiters. An automated image importer would be a nice addition to our setup though, I'll take a look when I get some time.

File: app/models/product_importer.rb (hunk 12 above)
Comment (Matt-Yorkley) [inline app/models/product_importer.rb]: Ok, the method_missing stuff is gone now, and it's a lot clearer.

File: app/models/product_importer.rb [hunk 16]
```diff
@@ -177,7 +181,12 @@ def count_existing_products
           AND spree_variants.deleted_at IS NULL', supplier_id).
           count
 
-        @supplier_products[:by_supplier][supplier_id] = {existing_products: products_count}
+        if @supplier_products[:by_supplier][supplier_id]
+          @supplier_products[:by_supplier][supplier_id][:existing_products] = products_count
+        else
+          @supplier_products[:by_supplier][supplier_id] = {existing_products: products_count}
+        end
```
enricostano: You should be able to do the assignement without worring too much:
```ruby
@supplier_products[:by_supplier][supplier_id] = { existing_products: products_count }
```
there is no need for the `if` here, right?

app/models/product_importer.rb
```diff
@@ -0,0 +1,461 @@
...
+  end
+
+  def count_existing_products
+    @suppliers_index.each do |supplier_name, supplier_id|
+      if supplier_id
+        products_count = Spree::Variant.joins(:product).
+          where('spree_products.supplier_id IN (?)
+          AND spree_variants.is_master = false
+          AND spree_variants.deleted_at IS NULL', supplier_id).
+          count
```
enricostano: Moving `products_count` to a method would make this more readable IMHO
```ruby
# @param supplier_id [Integer]
# @return [ActiveRecord::Relation]
def products_for_supplier(supplier_id)
  @products ||= Spree::Variant.joins(:product).
    where('spree_products.supplier_id IN (?)
    AND spree_variants.is_master = false
    AND spree_variants.deleted_at IS NULL', supplier_id)
end

def count_existing_products
  @suppliers_index.each do |supplier_name, supplier_id|
    products = products_for_supplier(supplier_id)
    @supplier_products[:by_supplier][supplier_id] = { existing_products: products.count }
    @supplier_products[:total] += products.count
  end
end
```

app/models/product_importer.rb
Matt-Yorkley: I thought so initially, but there are two keys being set in the same space in the hash at different times, so I found it didn't work without this conditional. I'll have another look and see if it can be refactored a bit now, that it's finished...

Matt-Yorkley: So before this part (during validation) under @supplier_products[:by_supplier][supplier_id] there may or may not be a key set for [:updates_count] that counts the number of updated item per supplier. So here we need to check if the [supplier_id] key has been set already, because if it has then the [:existing_products] key is being set alongside the [:updates_count] key in the same level, and we don't want to overwrite this value.

app/models/product_importer.rb
Matt-Yorkley: In this instance we just want a simple count of the enterprise's products, like: "30".

If we didn't use `.count` as part of the initial query but instead used it afterwards on the returned results (`products.count`), we would end up retrieving large amounts of data from the database that we don't need (every field from every variant in each enterprise). Basically this makes the difference between `SELECT COUNT(*) FROM "spree_variants"...` and `SELECT * FROM "spree_variants"...` in terms of SQL output.

app/models/product_importer.rb
enricostano: I see, maybe storing the whole state of this transaction in a Hash is not the most flexible way and can lead easily to future bugs overriding keys. Imagine that I don't notice that detail and I make a PR removing that `if`...

I generally use ruby objects to represent this kind of state. Another thing that you gain by using ruby objects is that you can encapsulate logic in those objects making this `ProductImporter` much easier to follow, maintain and test.

Of course doing that now would imply dedicating more time on this so don't bother me... but maybe we can just talk about it and see how this feature could be refactored using ruby objects instead of Hash structures. Ping me any time if / when you want to do this. 💪

app/models/product_importer.rb
enricostano: Well, actually since `products_for_supplier()` returns an `ActiveRecord::Relation` it will hit the DB only when finally invoked with `.count`. Making only the `COUNT(*)` query.

What can be improved is to make the query only once, I need to check it in console but the `||=` memoization should get you covered.

app/models/product_importer.rb
Matt-Yorkley: Is there an easy way to create blank/generic objects and assign attributes? Or do you mean creating a whole new class for storing this data? Really it's just a hash containing a few values that are all integers. I could possibly restructure the hash to make it simpler.

One thing I've done this week is write a lot more specs that test the functionality in a wide range of different scenarios. That makes it easier to tell if something isn't working after a change.

enricostano: Yep, I would create a `Product` or `Entry` class and map every row from the spreadsheet into an instance of that class.

What I'm suggesting is that if you use a more flexible object you can encapsulate logic and avoid relying on Hash `keys` presence. It would make all the workflow much more robust and explicit. But again, using Hash to just store numbers is OK. Just make sure that you don't need a better data structure to add logic and meaning to those numbers.

app/views/admin/product_import/_entries_table.html.haml
```diff
@@ -0,0 +1,12 @@
+- if entries && entries.count > 0 #&& entries.values.first && entries.values.first[:entry]
+  %div.table-wrap
```
pmackay: There isnt actually a need for `%div` here and in other places. Its assumed to be the default if just a class or ID is used

app/models/product_importer.rb
Matt-Yorkley: I've refactored it a bit to make it slightly clearer. I split part of the data out into a separate attribute, simplified the hash and renamed the keys.

Matt-Yorkley: I'll have a think about an Entry object though, maybe that would be the way to go.

enricostano: Cool! Please feel free to ping me if you want to talk about it first. I'll be in OFN Slack channels 😃

app/models/product_importer.rb
Matt-Yorkley: Ah I see what you mean. It would still be doing a separate query for each enterprise though, right? So for three enterprises, it would have to do three queries to get a separate count for the number of products in each enterprise.

enricostano: Yes, but that's the same case in the current implementation. Right?

BTW, I would even remove the memoization actually.

Matt-Yorkley:
@oeoeaio Lynn mentioned a merge-as-we-go plan for this in regards to your experience with Standing Orders? Is there any chance we could get this merged before I start adding lots more things on top of this branch? It doesn't have any links to it in the admin section yet, so it's not public-facing.

oeoeaio:
@Matt-Yorkley Yup, let's do it. Going to see if I can grab @sstead for a bit to do a quick run-through to see if she can pick up any issues that might have been missed...

sstead:
Had a play around and found that I couldn't type into these filter lists as I normally can... typing wouldn't narrow the list down.
image

Matt-Yorkley:
@sstead it should be fixed now.

oeoeaio:
Hey @Matt-Yorkley, looking great.

We'd love to have another crack at merging this into master on Friday. The one thing that strikes me to date is that I would love to see a lot more unit testing, particularly the importer itself (`spec/models/product_importer_spec.rb`) and the import controller.

Matt-Yorkley:
@oeoeaio Great, I've got unpushed commits piling up. I've overhauled everything quite drastically and it's got a lot more complex, so maybe we should just wait until the new stuff is ready to go? I'm hoping to get it more or less finished and ready for proper testing this week.

I've got about 14 tests in the feature spec now that basically cover every bit of functionality in every step of the PI flow. What sort of thing would you want form the model and controller specs? Like basic UI checks for the controller and some tests for the kind of data the model returns?

oeoeaio:
Hey @Matt-Yorkley, we've merged in all of this work (via #1479), and it has been released with v1.8.8

I'm closing this PR now, so if you could make a new one, with any additional work based on the current version of master, that would be perfect. :)

oeoeaio:
@Matt-Yorkley in response to your earlier question, I guess my approach would have been different from the beginning: unit tests to cover all of the complexity of the different cases of the importer, and then a feature spec or two to cover the main user flows and check that they are broadly working. This combination results in faster spec run times, and is easier to work with when doing things like the spree upgrade: we tend to make all of the unit specs pass first because they are able to be more explicit about what has gone wrong, and then work on getting feature specs once most of the issues have been resolved.

If you've already put all of the work into the feature specs then there is no point in removing them and starting again - but how do you feel about adopting this approach in future? Does it sit ok with you?

Matt-Yorkley:
@oeoeaio that sounds good. I guess I wanted to make sure everything was working during each step of the import flow and test all the feedback, but I could probably pull a lot of it out into unit tests now. It was really handy whilst developing.

PR #1458, opened 2017-02-19
Finishing Translations keys implemtentation on Admin ERB views

pacodelaluna:
- For accounts_and_billing_settings and business_model_configuration views, I have added the missing keys and updated the old ones to be coherent with the Rails convention (keys built based on view path)
- For accounts views, I haven't done it yet as I am not understanding where this view is called, tell me if it is really used and I will update it. Else I will delete it.


mkllnk:
@pacodelaluna How did you change all the locales? We normally don't bother, because Transifex should do the same with the translation memory.

sstead:
Testing all looks fine.

Issue #1207, opened 2016-11-09
Create a printable receipt and enable printing for POS


myriamboure:
http://community.openfoodnetwork.org/t/print-a-receipt-in-a-pos-use-of-ofn/719

myriamboure:
Proposal of modelisation of receipt: https://docs.google.com/document/d/1oE5unhTI2gG8X467Vl3iR9y4VeUJLo_tyuiB_XSa1co/edit?usp=sharing

NickWeir63:
looks good to me

myriamboure:
Sorry I wrote that in French, I will translate to English!

myriamboure:
I guess we will need to add a "print receipt" button in the "action" list for an order.

myriamboure:
@oeoeaio is what @ltrls propose ok for you? We need to push the commit quickly as the shop is opening in 3 weeks!!!

oeoeaio:
Hey @myriam and @ltrls, sounds promising! I will try and find some time this weekend to have a more detailed look.

oeoeaio:
Hey @ltrls. @RohanM and I have been discussing this and we think it would be best to create a separate repo for the client binaries for now. Seeing as we are not using git LFS for anything else, it seems like a lot of overhead to add to the project for just this feature. How do you feel about storing the libraries in a GitHub repo?

PR #1459, opened 2017-02-20
Add Qz integration

ltrls:
This is a pull request for #1207.
It adds a new admin route and page for printing a ticket using a ESC-P thermic printer via [Qz](www.qz.io).
As suggested in the issue I hosted the related client as releases here: https://github.com/ltrls/qztray-ofn/releases
One thing I couldn't manage to do was opening a pop-up instead of a new page (which would then close itself after printing), as the links I inserted are parts of an Angular dropdown and I didn't find a way to have an `onclick` (or `ng-click`) attribute. My Angular skills are really poor ><


myriamboure:
Ping @oeoeaio, @ltrls is finishing all the job on invoice internationalisation, tax sales report and printing receipt, so you should have everything now or in the coming hours, maybe tomorrow :-) Hope we can manage to merge quickly!

myriamboure:
@oeoeaio we just talked with @ltrls about it, the only impact is that it adds in the menu "actions" at the order level a new line "print receipt"... I don't think that is a problem if all the hubs see that in their menu and can if they want print a receipt?

myriamboure:
Maybe we can add a topic in the user guide on it? @sstead what do you think? And/or in the Github wiki?

sstead:
@Myriam yes this feature should be added to the user guide when it goes to production. I usually do this after testing. In this case, I may not test this feature, because we don't have a printer. In this case, you (or @ltrls ) might like to write the user guide post for this feature, including as much info as you can about how to setup the printer? You can either insert it into the guide yourself (in the two 'View Orders' pages or in advanced features), or send a description to me and I'll add it. How does that sound?

myriamboure:
@sstead that sounds great, I'm happy to update the user guide as soon as the feature is merged, with the help of @ltrls on the technical set up (if needed).
So if we do that is it ok to merge without doing the Toogle thing @oeoeaio ? :-) As it will be available for all instances...
Just ping me when merged and I do the user guide update :-)

myriamboure:
@daniellemoorhead this is the second pull request we need to release in the 15th march release. Thanks in advance! I have started to document the process for the user guide, this can't be test by the Aus team as you don't have a thermal printer, so we are doing it with our client and @ltrls  at the moment. Thanks

app/assets/javascripts/qz/ticket-popup.js
```diff
@@ -0,0 +1,85 @@
+var privateKey = "-----BEGIN RSA PRIVATE KEY-----\n" +
```
ltrls: Yeah I'm totally breaking their "security model". But I have good reasons to do so.
First the "attack surface" is narrow. Someone willing to print a ticket would launch the qz client (the one I compiled for OFN) and, at his first print, would accept the certificate (cert.pem) as trusted. The bad scenario is that a bad guy would trick this person to navigate to a page where he would have used the (in fact public) private key to well, print things via qz.
It's worth noting that this scenario is probably not the only reason for the existence of the cryptosystem, as it's also what they use for their business model: companies would pay for a key pair generated for the cert inside the release of the client that they distribute. They still don't absolutely enforce it as they [allow to bundle you own certificate](https://qz.io/wiki/compiling#ant-command-line).
And having a private key that needs to be both kept private and distributed to anyone willing to deploy OFN is something that seems to me rather hard or at least pretty impratical.
So I believe it's not worth it.

ltrls:
@oeoeaio I've done that! and rebased it on top of #1374 because it was simpler to reuse some stuff

oeoeaio:
@ltrls thanks very much! This is wonderful. I'll have a quick look through and then @sstead and I will get it tested.

app/assets/javascripts/qz/ticket-popup.js
oeoeaio: Ok, but the only benefit to not using an ajax call to retrieve a signed message from an endpoint on the server is your last point, which doesn't really seem overly onerous to me. It's a one-time **optional** configuration. Hardly a difficult hurdle for someone seriously considering deploying the ofn...

In fact, by *not* baking the secret key into the code, we allow greater flexibility: anyone who wants to deploy an instance of the ofn can always compile their own client and use their own secret key.

oeoeaio: The other issue (that @RohanM pointed out to me) is that we are assuming that "print(ing) things via qz" is actually the worst thing that an attacker can do with our private key. I have no knowledge of the capabilities of the Qz client, and so I have no way of knowing whether this is a good assumption or not, especially seeing as we are breaking their "security model"...

oeoeaio: I'm going try and get this through anyway, but I have created a new issue to hold the conversation about this particular point: #1484

Issue #1446, opened 2017-02-08
When manager add an order in back office, inventory prices not taken into account


myriamboure:
A hub manager is using the inventory option for BtoB shop, he uses it to set up different prices for business clients.
After he had closed the order cycle, he receives an email of a client who is late but would like to pass an order. He goes into the back office, "orders", "add new" to capture the order of the client and send him a confirmation. BUT ... the prices that appears are not the one of the inventory, and the wrong prices appear in the order and in the order confirmation that he sends to his client.

I told him as a work around to reopen the OC, pass the order and close it, but that's not ideal, take more time, and when the shop is open it can be confusing if there is another OC at the same time and people connect to buy, they would see both, etc.

Example with the product "pamplemousse":
![screenshot from 2017-02-08 09-33-42](https://cloud.githubusercontent.com/assets/8135376/22729792/870b1494-ede4-11e6-9dc3-25fd9ccaf70b.png)

In inventory:
![screenshot from 2017-02-08 09-54-40](https://cloud.githubusercontent.com/assets/8135376/22729817/a3a1fcee-ede4-11e6-951c-7862f605ff2a.png)

myriamboure:
Did anyone have the same issue @sstead on your side? @lin-d-hop ?

sstead:
@myriamboure  yes this is happening on Aus Production. This is quite an urgent bug, @lin-d-hop can anyone in the UK take a look?

lin-d-hop:
@stveep or @Matt-Yorkley - Do either of you have a moment to look at this? Would be a useful contribution to the global core, though not affecting UK while we aren't making use of inventories much.

sstead:
Potentially related...
This order was put through the front end. The price calculated was correct, but on viewing the completed order the $2 displayed in the 'price' column (for the two grapefruit items) is the Product price,, whereas it should show the Inventory price. The total for this order is correct, it's just the display in this column that is incorrect.
![image](https://cloud.githubusercontent.com/assets/10843056/23289540/7f1104d0-fa9e-11e6-9240-be3e078ccaff.png)

sstead:
The Inventory price is showing correctly now.

There is an issue with On Hand values not reading from Inventory though.
![image](https://cloud.githubusercontent.com/assets/10843056/26238843/26d95044-3cbe-11e7-867a-43edde21dec2.png)

Also, the manager creating an order can add 'hidden' products, when they shouldn't really be able to.

[...]

PR #1460, opened 2017-02-20
Uk/1446 inventory prices

stveep:
Fix for #1446 - this was due to the order not having the distributor set at the point where line items are added, meaning that the variant cannot be scoped to the distributor. I set it up to pass the distributor as an extra param, which can be used if the order distributor is not found.

Please go through the rest of the order process and check that the prices are ok - I think it should be ok once the line item has been added though.

### DIFF
--- a/app/controllers/spree/admin/line_items_controller_decorator.rb
+++ b/app/controllers/spree/admin/line_items_controller_decorator.rb
@@ -19,7 +19,12 @@ Spree::Admin::LineItemsController.class_eval do
 
   def create
     variant = Spree::Variant.find(params[:line_item][:variant_id])
-    OpenFoodNetwork::ScopeVariantToHub.new(@order.distributor).scope(variant)
+    dist = if @order.distributor
+      @order.distributor
+    elsif params[:distributor_id]
+      Enterprise.find(params[:distributor_id])
+    end
+    OpenFoodNetwork::ScopeVariantToHub.new(dist).scope(variant)
 
     @line_item = @order.add_variant(variant, params[:line_item][:quantity].to_i)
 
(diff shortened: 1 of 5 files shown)


app/controllers/spree/admin/line_items_controller_decorator.rb
```diff
@@ -19,7 +19,12 @@ def index
 
   def create
     variant = Spree::Variant.find(params[:line_item][:variant_id])
-    OpenFoodNetwork::ScopeVariantToHub.new(@order.distributor).scope(variant)
+    dist = if @order.distributor
+      @order.distributor
+    elsif params[:distributor_id]
+      Enterprise.find(params[:distributor_id])
+    end
+    OpenFoodNetwork::ScopeVariantToHub.new(dist).scope(variant)
```
enricostano: Maybe you can move `dist` to a private method, something similar to:
```ruby
def distributor
  if @order.distributor
    @order.distributor
  elsif params[:distributor_id]
    Enterprise.find(params[:distributor_id])
  end
end
```
BTW, what happens when `dist` is `nil`?

stveep:
Yes this was a pretty quick and dirty fix. I've put something together the way you suggested:
https://github.com/stveep/openfoodnetwork/tree/uk/1446-2-set-new-order-distributor

I'll probably have to do quite a bit of feature spec rewiring as we now have an extra refresh step, so let me know if this is what you had in mind.

oeoeaio:
Yeah, something like that. Only suggestion would be that maybe we can simplify a bit by rendering a new page `set_distribution` from the controller when `new` is requested but `order.distribution_set?` is `false`. Then you can submit just the order_cycle and distributor and load the normal `new` view, without the need to have any logic about when to show/hide the distribution_fields partial in the view. This way you can remove an override, instead of adding one... Thoughts?

daniellemoorhead:
Hey there, removing the Aus Review label from this as we're waiting on response to @oeoeaio's last comment.

myriamboure:
Hi there! How is this moving forward? Just a question as we have some users here who are confused about the issue... but no pressure of course, that's great you are working on it @stveep thank you so much!

stveep:
Hi @myriamboure - I am picking this up again and hopefully will finish at the weekend. I'm redoing it in the way rob suggested above.

myriamboure:
Awesome, thank you @stveep :-)

stveep:
Eventually superseded and fixed by #1575

Issue #1466, opened 2017-02-23
Stripe: allow logged in customers to store the credit card details they enter at the checkout for later use


oeoeaio:
The StripeConnect gateway I have set up is pretty much the StripeGateway copied from the latest version of Spree, and I have then added a few bits to configure it for use with Connect. 

As it currently stands, I have it working with Stripe tokens requested via Stripe.json the checkout page, which are then used by the StripeConnect gateway to make charge requests on behalf of connected stripe account (shops).

The next phase is to get it to store credit card details for later use. Firstly from the checkout (this issue), and later by explicitly entering card details in a separate form: #1456.

### Storing card details

Most of the data model components required for this to work already exist in Spree, we just need to do a little bit of tweaking to get them to play together properly.

The StripeConnect gateway is already creating instances of `Spree::CreditCard`, and storing card info against these. The problem is that the process of storing the information is a bit broken. For starters, the `Spree::Payment#create_payment_profile` is not really set up for storing cards that are represented by a token (`source.gateway_payment_profile_id`), as is requires `source.number` be present before calling `payment_method.create_profile`. Assuming we override this the main remaining problem is configuring the StripeConnect gateway to conform to Stripes fairly strict process for doing what we want to do: store card details against a customer on the platform's stripe account, and then charge the same card on behalf of a connect account (shop).

In the case where a customer is entering and wanting to store credit card details for the first time, this means we have a fairly complicated process, as we need to ping Stripe in four separate requests:

1. Submit the card details using Stripe.js from the checkout page, get a token back.
2. From `Spree::Gateway::StripeConnect#create_profile`: use the token to store the previously submitted card against a new Stripe customer, get a Stripe `customer_id` and `card_id` back. These are stored against the existing Spree::CreditCard model as `gateway_customer_profile_id` and `gateway_payment_profile_id` respectively. 
3. Request a NEW token, which can be used to make a new charge. We can't reuse the existing token because is has already been 'used', pass this 'new' token needs to be returned from `Spree::Gateway::StripeConnect#options_for_purchase_or_auth`.
4. Submit a charge request on behalf of a connected Stripe Account (ie. that belonging to the enterprise user).

NOTE: If we were not charging on behalf of a connected account, we could just submit a charge request using the customer_id and card_id received in step 2, without doing step 3, but that is not what we are doing.

[...]

Issue #1456, opened 2017-02-15
Stripe: Interface for customers to add their credit card details


daniellemoorhead:
Need a new screen/place on the platform for customers to store their credit card details (within the customer account section). 

This issue should be fairly trivial once #1466 is done.

daniellemoorhead:
@stveep we figure you can get involved / do this story once @oeoeaio has sketched out where the function will live in customer accounts and what it'll look like.

Stay tuned :)

PR #1468, opened 2017-02-23
Stripe connect

oeoeaio:
Work by @stveep to connect Stripe Accounts to the main instance, and by me to implement a new Stripe Connect based payment gateway.

I figure that if this all works, we can pull it in now, and then pull the rest of it in later...(#1466 and #1456)


app/controllers/checkout_controller.rb
```diff
@@ -125,7 +128,9 @@ def advance_order_state(order)
 
 
   def update_failed
+    binding.pry
     clear_ship_address
+    binding.pry
```
enricostano: ⚠️

app/controllers/checkout_controller.rb
```diff
@@ -202,4 +207,28 @@ def redirect_to_paypal_express_form_if_needed
     render json: {path: spree.paypal_express_url(payment_method_id: payment_method.id)}, status: 200
     true
   end
+
+  def construct_saved_card_attributes
+    existing_card_id = params[:order].delete(:existing_card)
+    if existing_card_id.present?
```
enricostano: early return maybe? `return unless existing_card_id.present?` or something like that.

app/controllers/spree/credit_cards_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    def format_credit_card_params(params_hash)
+      { month: params_hash[:exp_month],
+        year: params_hash[:exp_year],
+        last_digits: params_hash[:last4]
+      }
+    end
+
+    def set_user
+      @user = spree_current_user
+    end
```
enricostano: Why this? `spree_current_user` should be memoized already, right? just use it.

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,147 @@
...
+        payment.source.update_attributes!({
+          cc_type: payment.source.cc_type, # side-effect of update_source!
+          gateway_customer_profile_id: response.params['id'],
+          gateway_payment_profile_id: response.params['default_source'] || response.params['default_card']
+        })
+
+      else
+        payment.send(:gateway_error, response.message)
+      end
+    end
```
enricostano: This method seems too long to me, I see the opportunity to split it.

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,147 @@
...
+        if customer = creditcard.gateway_customer_profile_id
+          # Assume the gateway_payment_profile_id is a Stripe card_id
+          # So generate a new token, using the customer_id and card_id
+          return tokenize_instance_customer_card(customer, token_or_card_id)
+        end
+
+        # Assume the gateway_payment_profile_id is a token generated by StripeJS
+        return token_or_card_id
+      end
+    end
```
enricostano: Conditionals with assignments could lead to weird outcomes in ruby. Try to avoid them if possible, furthermore they make the code less readable IMHO.

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    def tokenize_instance_customer_card(customer, card)
+      begin
+        token = Stripe::Token.create({card: card, customer: customer}, { stripe_account: stripe_account_id})
+        token.id
+      rescue Stripe::InvalidRequestError
+        # Not really sure what to do here....
```
enricostano: Maybe at least log it in stdout?

app/controllers/admin/enterprises_controller.rb
```diff
@@ -113,6 +114,51 @@ def for_line_items
...
+        # Get the Enterprise
+        @enterprise = Enterprise.find_by_permalink(state["enterprise_id"])
+        # Get the deets from Stripe
+        response_params = get_stripe_token(params["code"]).params
+
+        # In case of a problem, need to also issue a request to disconnect the account from Stripe
+        if !(spree_current_user.enterprises.include? @enterprise) && !(spree_current_user.admin?)
+          deauthorize_request_for_stripe_id(response_params["stripe_user_id"])
+          redirect_to '/unauthorized' and return
+        end
```
sauloperez: I think that lines from 123 until here are business logic that would be better off in a factory method of `StripeAccount` at least, or even on a different object.

As I see it, we could set up error handling here in the controller to deal with these two exceptions:

```ruby
def stripe_connect_callback
  stripe_account = StripeAccountFactory.new(params)
  if stripe_account.save
    ...
  end
rescue Stripe::Error
  redirect_to '/unauthorized' and return
end
```

app/controllers/enterprises_controller.rb
```diff
@@ -10,14 +10,14 @@ class EnterprisesController < BaseController
   before_filter :check_stock_levels, only: :shop
 
   before_filter :clean_permalink, only: :check_permalink
+  before_filter :set_enterprise, only: :relatives
```
sauloperez: why don't you call the method from the controller action? I believe single-action `before_filter`s have little to no value. It is easier for the reader to see a method call when reading the action than jumping between the `before_filter` and the action.

app/controllers/enterprises_controller.rb
```diff
@@ -37,8 +37,14 @@ def check_permalink
     end
   end
 
+
+
```
sauloperez: :warn

app/controllers/spree/credit_cards_controller.rb
```diff
@@ -0,0 +1,68 @@
+module Spree
+  class CreditCardsController < BaseController
+
+    before_filter :set_credit_card, only: [:destroy]
+    before_filter :destroy_at_stripe, only: [:destroy]
```
sauloperez: related to my comment in enterprises_controller.rb

app/controllers/spree/credit_cards_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        end
+      else
+        render json: "error creating Stripe customer", status: 500
+      end
+    end
+
+    def destroy
+      if @credit_card.destroy
+        flash[:success] = I18n.t(:card_has_been_removed, number: "x-#{@credit_card.last_digits}")
+        redirect_to "/account#/cards"
```
sauloperez: Using url helpers here would make it future-proof. In case someday we need to change the route we don't have to update this line.

app/controllers/spree/users_controller_decorator.rb
```diff
@@ -1,3 +1,20 @@
 Spree::UsersController.class_eval do
   layout 'darkswarm'
+  before_filter :set_credit_card, only: :show
```
sauloperez: See my suggestion in enterprises_controller.rb

app/helpers/admin/stripe_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    private
+    def jwt_encode payload
+      JWT.encode(payload, Openfoodnetwork::Application.config.secret_token, 'HS256')
+    end
+
+    def jwt_decode token
+      JWT.decode(token, Openfoodnetwork::Application.config.secret_token, true, algorithm: 'HS256')[0] # only returns the original payload
+    end
+  end
+end
```
sauloperez: I stronlgy believe this module should actually be the abstraction the controllers are lacking. For my experience, I believe that payment gateways need to be encapsulated so as not to leak details throughout the whole app. I see that these methods could be independent service objects like `Stripe::Authorize`, `Stripe::Deauthorize`, etc.

app/models/enterprise.rb
```diff
@@ -365,6 +366,7 @@ def can_invoice?
     abn.present?
   end
 
+
```
sauloperez: ⚠️

app/models/spree/gateway/stripe_connect.rb
sauloperez: It might be a `CreateProfile` on its own.

sauloperez:
Feel free to dismiss my comments. I know I came late to the party and I don't want to block the PR.

oeoeaio:
@sauloperez, no way. I will definitely be looking at them!

app/controllers/spree/admin/base_controller_decorator.rb
```diff
@@ -34,7 +34,7 @@ def unauthorized
       redirect_to '/unauthorized'
     else
       store_location
-      redirect_to root_path(anchor: "login?after_login=#{ request.env['PATH_INFO'] }")
+      redirect_to root_path(anchor: "login?after_login=#{request.env['PATH_INFO']}")
     end
   end
```
stveep: This reminds me - I had problems with getting signed out of OFN when going through the Stripe Connect process to connect an account. When I was redirected back to OFN I got prompted to sign in, but was not then redirected to the admin page (and the account was not connected on the OFN side as the request was not successful). I only experienced it locally and I'm not sure what the reason for getting signed out was. I remember playing with using some of the other request params here instead (e.g. #fullpath), but was worried I was going to break everything...

https://github.com/stveep/openfoodnetwork/commit/9f3888c68e627a8fa3fbfc7ed13bebf0775e04e5

app/models/spree/gateway/stripe_connect.rb
stveep: What's your preferred pattern to check whether something exists and then use it if it does?

app/helpers/enterprises_helper.rb
```diff
@@ -21,6 +21,7 @@ def available_shipping_methods
   def available_payment_methods
     return [] unless current_distributor.present?
     payment_methods = current_distributor.payment_methods.available(:front_end).all
+    payment_methods.reject!{ |p| p.type.ends_with? "StripeConnect" } unless Spree::Config.stripe_connect_enabled
```
sauloperez: This is the second time I see this, so before duplicating this everytime we show payment methods, I suggest adding a scope/class method on our decorator that filters it out depending on `stripe_connect_enabled`.

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,152 @@
...
+
+        creditcard = card_to_store(payment.source)
+
+        response = provider.store(creditcard, options)
+        if response.success?
+          payment.source.update_attributes!( cc_type: payment.source.cc_type, # side-effect of update_source!
+                                             gateway_customer_profile_id: response.params['id'],
+                                             gateway_payment_profile_id: response.params['default_source'] || response.params['default_card'])
+        else
+          payment.send(:gateway_error, response.message)
```
sauloperez: is there any way to avoid using `send` here? this could break anytime :see_no_evil:

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,152 @@
...
+        # So generate a new token, using the customer_id and card_id
+        tokenize_instance_customer_card(customer, token_or_card_id)
+      end
+
+      def tokenize_instance_customer_card(customer, card)
+        token = Stripe::Token.create({card: card, customer: customer}, stripe_account: stripe_account_id)
+        token.id
+      rescue Stripe::StripeError => e
+        Rails.logger.error("Stripe Error: #{e}")
+        nil
```
sauloperez: Shouldn't we raise when we couldn't create the token? IMO we better show the failure and fix it than finding incosistent data leading to bugs later on.

app/models/spree/payment_decorator.rb
```diff
@@ -73,6 +75,15 @@ def refund!(refund_amount=nil)
       end
     end
 
+    # Import from future Spree
```
sauloperez: please, specify from which version so we know when upgrading.

app/models/spree/payment_method_decorator.rb
```diff
@@ -4,6 +4,7 @@
   acts_as_taggable
 
   has_and_belongs_to_many :distributors, join_table: 'distributors_payment_methods', :class_name => 'Enterprise', association_foreign_key: 'distributor_id'
+  has_many :credit_cards, class_name: "Spree::CreditCard" # from future Spree version
```
sauloperez: please, specify from which version so we know when upgrading.

app/models/stripe_account.rb
```diff
@@ -0,0 +1,16 @@
+class StripeAccount < ActiveRecord::Base
+  belongs_to :enterprise
+  validates :stripe_user_id, :stripe_publishable_key, presence: true
+  validates :enterprise_id, uniqueness: true
```
sauloperez: we must enforce this constrain at db level as well.

app/serializers/api/admin/payment_method_serializer.rb
```diff
@@ -1,11 +1,29 @@
...
+  def method_serializer
+    if object.type == 'Spree::Gateway::StripeConnect'
+      Api::Admin::PaymentMethod::StripeSerializer.new(object)
+    else
+      Api::Admin::PaymentMethod::BaseSerializer.new(object)
+    end
+  end
+end
+
+module Api::Admin::PaymentMethod
```
sauloperez: I would move this to its own file for consistency. Future devs might expect finding the base serializer in the `app/serializers/api/admin/` directory like all the others.

app/serializers/api/admin/payment_method_serializer.rb
```diff
@@ -1,11 +1,29 @@
...
+    end
+
+    def tags
+      object.tag_list.map{ |t| { text: t } }
+    end
   end
 
-  def tags
-    object.tag_list.map{ |t| { text: t } }
+  class StripeSerializer < BaseSerializer
```
sauloperez: here as well

lib/stripe/account_connector.rb
```diff
@@ -0,0 +1,43 @@
...
+        stripe_user_id: oauth_response["stripe_user_id"],
+        stripe_publishable_key: oauth_response["stripe_publishable_key"],
+        enterprise: enterprise
+      )
+    end
+
+    private
+
+    def state
+      OAuth.send(:jwt_decode, params["state"])
```
sauloperez: Why don't we make `jwt_decode` public if we need to use it here?

spec/controllers/admin/enterprises_controller_spec.rb
```diff
@@ -144,6 +144,73 @@ module Admin
...
+          before { allow(controller).to receive(:spree_current_user) { distributor_manager } }
+
+          context "when the connector raises a StripeError" do
+            before do
+              allow(Stripe::AccountConnector).to receive(:new).and_raise Stripe::StripeError, "some error"
+            end
+
+            it "returns a 500 error" do
+              spree_get :stripe_connect_callback, params
+              response.status.should be 500
```
sauloperez: please, we should use the current `expect(...).to eq()` syntax :pray:

app/models/spree/payment_decorator.rb
enricostano: Also a link to the actual code in Spree repo would help, thanks!

app/controllers/admin/stripe_connect_settings_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      return @stripe_account = { status: :empty_api_key } if Stripe.api_key.blank?
+      attrs = %i[id business_name charges_enabled]
+      @obfuscated_secret_key = obfuscated_secret_key
+      @stripe_account = Stripe::Account.retrieve.to_hash.slice(*attrs).merge(status: :ok)
+    rescue Stripe::AuthenticationError
+      @stripe_account = { status: :auth_fail }
+    end
+
+    def update
+      Spree::Config.set(params[:settings])
```
enricostano: Can you please explain and document this?
What do you expect in `params[:settings]`?
Is `Spree::Config.set()` really needed? It will get cached by Rails and affect other requests (and specs).

app/controllers/enterprises_controller.rb
```diff
@@ -39,6 +39,10 @@ def check_permalink
 
   private
 
+  def set_enterprise
+    @enterprise = Enterprise.find(params[:id])
```
enricostano: what if the enterprise does not exist? do we need any authorization ability here?

app/controllers/spree/admin/payment_methods_controller_decorator.rb
```diff
@@ -57,12 +57,19 @@ def load_data
...
 
       def load_hubs
         @hubs = Enterprise.managed_by(spree_current_user).is_distributor.sort_by!{ |d| [(@payment_method.has_distributor? d) ? 0 : 1, d.name] }
       end
+
+      # Show Stripe as an option if enabled, or if the
+      # current payment_method is already a Stripe method
+      def show_stripe?
+        Spree::Config.stripe_connect_enabled || @payment_method.try(:type) == "Spree::Gateway::StripeConnect"
+      end
```
enricostano: ⚠️ DISCLAIMER: buzz words ahead ⚠️ 

I think there is room here to apply some "tell don't ask", instead of asking all the time if we support Stripe we should just get that info from some class defined starting up the app (strategy pattern maybe?). In that way you'll not have to remember to check that "if Stripe" in the future avoiding bugs and letting the code be more clear and easier to test.

BTW, this is not blocking of course. Just something that would be nice to start seeing in the codebase.

app/controllers/spree/credit_cards_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      card = Spree::CreditCard.new(attrs)
+      # Can't mass assign these:
+      card.cc_type = attrs[:cc_type]
+      card.last_digits = attrs[:last_digits]
+      card.user_id = spree_current_user.id
+      card
+    end
+
+    def set_credit_card
+      @credit_card = Spree::CreditCard.find(params[:id])
```
enricostano: What if the card doesn't exist? Any authorization layer needed?

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,152 @@
...
+      #   provider.refund(money, response_code, {})
+      # end
+
+      # def void(response_code, creditcard, gateway_options)
+      #   provider.void(response_code, {})
+      # end
+
+      # def cancel(response_code)
+      #   provider.void(response_code, {})
+      # end
```
enricostano: Can we remove commented out code?

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,152 @@
...
+
+      def create_profile(payment)
+        return unless payment.source.gateway_customer_profile_id.nil?
+        options = {
+          email: payment.order.email,
+          login: Stripe.api_key,
+        }.merge! address_for(payment)
```
enricostano: Not super important, but if you move this hash population to a separate method you'll have more documentation surface and we'll get prepared for the rubocop rules about methods and classes length I hope we'll agree upon some day 😬

app/models/spree/payment_decorator.rb
```diff
@@ -73,6 +73,15 @@ def refund!(refund_amount=nil)
...
+    def build_source
+      return if source_attributes.nil?
+      if payment_method and payment_method.payment_source_class
+        self.source = payment_method.payment_source_class.new(source_attributes)
+        self.source.payment_method_id = payment_method.id
```
oeoeaio: Hey @stveep, I am a bit worried about this line. I don't think it is going to work for us to link instances `Spree::CreditCard` to a single `Spree::PaymentMethod` - this will imply that each card can only be used by a single shop, and we want cards to be used by multiple shops.... Was there a particular reason that we need this association as far as you're aware? I realise this method is coming from Spree, so they must have a reason but I suspect it is more for convenience than actual necessity, because the payment itself stores the payment method, so I don't see why the card needs to store it as well...

enricostano: Maybe @sauloperez can help here too

app/controllers/admin/stripe_connect_settings_controller.rb
oeoeaio: Updating the cached config is exactly what we're trying to achieve here. This controller action is used by super admin users to update the settings the app is using.

app/controllers/spree/admin/payment_methods_controller_decorator.rb
oeoeaio: The value of `Spree::Config.stripe_connect_enabled` is dynamic (can be changed by super admin), so we can't cache the value on app initialisation. Is that what you mean?

app/controllers/admin/stripe_connect_settings_controller.rb
enricostano: Ah OK, could you please specify that in the method documentation? Thanks!

app/controllers/spree/admin/payment_methods_controller_decorator.rb
```diff
@@ -74,7 +74,7 @@ def show_stripe?
 
       def restrict_stripe_account_change
         return unless @payment_method.try(:type) == "Spree::Gateway::StripeConnect"
-        return unless @payment_method.preferred_enterprise_id
+        return unless @payment_method.preferred_enterprise_id.andand > 0
```
sauloperez: What about checking there's an actual `@payment_method` first? this way you can avoid lettering the rest of the method with conditionals like `try` or `andand`. We should even raise an error and show it in the UI accordingly when there is no `@payment_method` on an update, but I don't know precisely how this controller works.

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -127,6 +134,11 @@ def tokenize_instance_customer_card(customer, card)
         Rails.logger.error("Stripe Error: #{e}")
         nil
       end
+
+      def ensure_enterprise_selected
+        return if preferred_enterprise_id.andand > 0
```
sauloperez: I don't quite understand what you check with `andand`. It should be enough with `return if preferred_enterprise_id.present?`

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -3,6 +3,8 @@ class Gateway
     class StripeConnect < Gateway
       preference :enterprise_id, :integer
 
+      validate :ensure_enterprise_selected
```
sauloperez: Isn't this custom validation exactly what  `validates : preferred_enterprise_id, presence: { message: I18n.t(:error_required) }` does?

app/helpers/enterprises_helper.rb
```diff
@@ -99,4 +98,13 @@ def order_changes_allowed?
...
   end
+
+  def apply_stripe_filters_to(payment_methods)
+    stripe_enabled = Spree::Config.stripe_connect_enabled && Stripe.publishable_key
+    if stripe_enabled
+      payment_methods.reject!{ |p| p.type.ends_with?("StripeConnect") && p.preferred_enterprise_id == 0 }
+    else
+      payment_methods.reject!{ |p| p.type.ends_with?("StripeConnect") }
+    end
+  end
```
sauloperez: I'm afraid we might be stuffing too much business logic into a helper.

First, helpers are meant to provide simple pure helper functions to the view. Secondly, as being modules they lack internal state, which makes us scatter such state throughout the methods with many arguments, and lack encapsulation.

IMHO things like this business logic shouldn't be in a helper but in single-purpose plain ruby objects with their own set of unit tests.

app/models/spree/gateway/stripe_connect.rb
oeoeaio: @sauloperez Nope, it's using the preferences gem, which by some magic transforms a nil value into zero, and a value of zero passes the presence validation

app/controllers/spree/admin/payment_methods_controller_decorator.rb
oeoeaio: 👍  sounds good

app/models/spree/gateway/stripe_connect.rb
oeoeaio: As above, it's a function of the way the preference is stored. I've specified it an as integer, and so it is defaulting to 0. `0.present?` is true, and I don't want to return if the value is 0. 

I will investigate whether I can specify a default of `nil`, and avoid all this fluffing around. 🙂

app/helpers/enterprises_helper.rb
oeoeaio: Yep, thanks @sauloperez. I was struggling with that this afternoon. To split or not to split. I'll add it to my list. 🙂

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Resolved

app/models/spree/gateway/stripe_connect.rb
oeoeaio: TODO

app/controllers/spree/credit_cards_controller.rb
oeoeaio: Resolved

app/controllers/checkout_controller.rb
oeoeaio: Resolved

app/controllers/checkout_controller.rb
oeoeaio: Resolved

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Resolved

app/controllers/admin/enterprises_controller.rb
oeoeaio: Resolved

app/controllers/enterprises_controller.rb
oeoeaio: Resolved

app/controllers/enterprises_controller.rb
oeoeaio: Resolved

app/controllers/spree/credit_cards_controller.rb
oeoeaio: Resolved

app/controllers/spree/credit_cards_controller.rb
oeoeaio: Resolved

app/controllers/spree/users_controller_decorator.rb
oeoeaio: Resolved

app/helpers/admin/stripe_helper.rb
oeoeaio: Resolved

app/models/enterprise.rb
oeoeaio: Resolved

app/assets/javascripts/darkswarm/controllers/checkout/payment_controller.js.coffee
```diff
@@ -1,23 +1,15 @@
-Darkswarm.controller "PaymentCtrl", ($scope, $timeout) ->
+Darkswarm.controller "PaymentCtrl", ($scope, $timeout, savedCreditCards, Dates) ->
   angular.extend(this, new FieldsetMixin($scope))
+  defaultCard = [ {id: null, formatted: t("new_credit_card")} ]
+  $scope.savedCreditCards = defaultCard.concat savedCreditCards if savedCreditCards
```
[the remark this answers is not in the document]
oeoeaio: Refactored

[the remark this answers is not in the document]
oeoeaio: Resolved

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Resolved

app/assets/javascripts/darkswarm/services/checkout.js.coffee
```diff
@@ -53,6 +59,20 @@ Darkswarm.factory 'Checkout', (CurrentOrder, ShippingMethods, PaymentMethods, $h
             last_name: @order.bill_address.lastname
         }
 
+      if @paymentMethod()?.method_type == 'stripe'
+        if @secrets.selected_card
```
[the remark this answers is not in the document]
oeoeaio: Sorry @sauloperez, I don't follow. You mean instead of the `?.` syntax?

app/controllers/admin/stripe_connect_settings_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      resource = t('admin.controllers.stripe_connect_settings.resource')
+      flash[:success] = t(:successfully_updated, :resource => resource)
+      redirect_to_edit
+    end
+
+    private
+
+    def load_settings
+      klass = Struct.new(:stripe_connect_enabled)
```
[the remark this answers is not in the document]
oeoeaio: Resolved

app/controllers/checkout_controller.rb
```diff
@@ -201,4 +204,29 @@ def redirect_to_paypal_express_form_if_needed
     render json: {path: spree.paypal_express_url(payment_method_id: payment_method.id)}, status: 200
     true
   end
+
+  def construct_saved_card_attributes
+    existing_card_id = params[:order].delete(:existing_card)
```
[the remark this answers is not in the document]
oeoeaio: Resolved

app/controllers/checkout_controller.rb
```diff
@@ -201,4 +204,29 @@ def redirect_to_paypal_express_form_if_needed
...
+    return if existing_card_id.blank?
+
+    credit_card = Spree::CreditCard.find(existing_card_id)
+    if credit_card.try(:user_id).blank? || credit_card.user_id != spree_current_user.try(:id)
```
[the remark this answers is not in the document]
oeoeaio: @sauloperez Users who check out as a guest can still pay via credit card.

app/helpers/enterprises_helper.rb
oeoeaio: This logic only appears in one place now, and I don't think it is worth adding a dedicated scope. I will however extract all of the logic for finding 'available' payment methods to a service object.

app/models/spree/gateway/stripe_connect.rb
oeoeaio: I'm sure we could do it another way, yes. What is your main concern. This was copy/pasted directly from:
https://github.com/spree/spree_gateway/blob/master/app/models/spree/gateway/stripe_gateway.rb#L82

app/models/spree/payment_decorator.rb
oeoeaio: Resolved

app/serializers/api/admin/payment_method_serializer.rb
oeoeaio: Resolved

app/serializers/api/admin/payment_method_serializer.rb
oeoeaio: Resolved

lib/stripe/account_connector.rb
oeoeaio: Resolved

app/models/spree/gateway/stripe_connect.rb
oeoeaio: TODO

app/controllers/checkout_controller.rb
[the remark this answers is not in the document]
oeoeaio: For reference this was written the way it was because `existing_card` is the key used by [future versions of Spree](https://github.com/spree/spree/blob/2-3-stable/core/app/models/spree/order/checkout.rb#L234).

app/models/stripe_account.rb
oeoeaio: Resolved

spec/controllers/admin/enterprises_controller_spec.rb
oeoeaio: Not in `spec/controllers/admin/stripe_accounts_controller_spec.rb`

Resolved

app/controllers/admin/stripe_connect_settings_controller.rb
oeoeaio: Resolved

app/controllers/enterprises_controller.rb
oeoeaio: @enricostano I have changed this to `Enterprise.find_by_id` so that an error isn't raised when a bad `id` is provided. This method is only used by `EnterprisesController#relatives`, logic there already deals with the case that no enterprise is found.

No authorisation required here IMO, as this is just describing the existence of relationships in the network. I think this is fine to be public knowledge.

app/models/spree/payment_decorator.rb
oeoeaio: Bump @stveep

app/controllers/spree/credit_cards_controller.rb
oeoeaio: Resolved

app/controllers/spree/admin/payment_methods_controller_decorator.rb
oeoeaio: Resolved?

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Resolved

app/models/spree/gateway/stripe_connect.rb
oeoeaio: @sauloperez I have tried but I can get `preferred_enterprise_id` to default to nil. It seems that when a `:integer` is specified as the type the default is always `0`, even if I try `default: nil`. I think that unfortunately I am going to have to keep this use of `andand`. Unless you can think of a better way of doing it?

oeoeaio:
@sauloperez @enricostano I think I have responded to all of your requested changes. Could you please update your reviews when you get a chance?

app/models/spree/gateway/stripe_connect.rb
enricostano: What's the plan on this @oeoeaio ?

oeoeaio: @enricostano Resolved

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Resolved

app/controllers/spree/admin/payment_methods_controller_decorator.rb
oeoeaio: Resolved

app/helpers/enterprises_helper.rb
oeoeaio: Resolved

app/assets/javascripts/darkswarm/services/stripe_elements.js.coffee
```diff
@@ -0,0 +1,47 @@
+Darkswarm.factory 'StripeElements', ($rootScope, Loading, RailsFlashLoader) ->
+  new class StripeElements
+    # TODO: add locale here for translations of error messages etc. from Stripe
```
stveep: Just spotted this - we should pass the locale in the calls to the Stripe API, which will ensure the credit card form and error messages are returned in the correct language. I couldn't figure out how to access it from the front end?

oeoeaio: Documented in #1864

PR #1477, opened 2017-03-08
Restore cart when a user logs back in

lin-d-hop:
This built on Levents changes with some small tweaks.

@OliverUK It needs to be tested thoroughly.
Cases:
- User logs out and logs back in
- User has items in basket when logging in
- User switches hubs from which they are shopping
- Hub closes between logging out and logging back in
- Product goes out of stock between logging out and logging back in

It would be good if you could spend some time and thoroughly document the use cases in a summary. Picture will only be required if you find errors.

Happy to discuss test cases in more detail if you think it would be useful, but I trust you. Take a look at the original issue as well.


OliverUK:
Hi @lin-d-hop   

I had a go at it now as I won't be around much the rest of the week.

User logs out and logs back in
Expected behaviour: cart content still there from last login
Test result **true**

User has items in basket when logging in
Expected behaviour: old cart replaced by current cart
Test result **true**

User left items in basket but logs back in after the old OC closed and a new one opened
Expected behaviour: cart emptied
Test result **true**

User has items in basket when switching to a different OFN shop
Expected behaviour: Warning displayed and cart emptied
Test result **true**
screen shot 2017-03-08 at 15 59 39

User has items in basket and logs back in after someone else bought the last stock of the same items.

Here OFN allows me to check out despite no stock available and enters a negative stock number in BEP for the item in question. It appears the order is not checked against stock levels but this is unrelated to the issue of logging out and back in. On UK staging at least, if I order the last stock before someone else completes their checkout, the other still gets to check out and the stock ends up negative.

lin-d-hop:
Wow you're speedy @OliverUK Thanks!!
Do you know if there is an existing issue for the stock issue you flagged? If not maybe we should create an issue for it?

OliverUK:
@lin-d-hop 
I wasn't aware but it looks like the same happens on production. I only tried it quickly by reducing the stock level and then carrying on checking out all under the same login.

daniellemoorhead:
Hey @oeoeaio @jeronimo @lin-d-hop can this PR be closed as it seems to have been superseded by  #1552 and #1583?

oeoeaio:
@daniellemoorhead looks like different functionality to me, I don't think we can close this...

daniellemoorhead:
@lin-d-hop @sineadfenton have taken this out of the dot point release, will leave it to you to ping us via the aus review label when you're ready for us to look at it

oeoeaio:
For my own reference, this is probably related: https://github.com/oeoeaio/openfoodnetwork/commit/4443229066983f8ff07c6c0dd3b5cbd21859e237

PR #1479, opened 2017-03-10
[TEST] Uk/import

oeoeaio:
IGNORE ME

Same code as @Matt-Yorkley's PR of the same name, I've just rebased it for testing in Aus


RohanM:
It'd be cool if we had an easy way to label PRs like this. I've seen `[WIP]` and `[RFC]` prefixed to the titles for work in progress and requests for comment. Perhaps `[TEST]`?

Issue #833, opened 2016-02-23
Verify emails of users, not enterprises


mkllnk:
Currently, only enterprise emails are validated with a confirmation email. This can lead to enterprise owners with invalid email addresses in the system.

It would be much clearer to verify email addresses on user level as discussed:
- http://community.openfoodnetwork.org/t/confirmation-emails-registration-flow/237
- http://community.openfoodnetwork.org/t/community-forum-for-ofn/450/18

daniellemoorhead:
@lin-d-hop and @stveep maybe this is something worthy of Levent's attention. It's a pretty important thing to have but it's never got to the top of the list of priorities. Designing out how it would work for new users and then for existing unverified users, and then implementing it, shouldn't be too much of a stretch....maybe?

PS. Didn't like Levent to this because not sure of his userID :)

NickWeir63:
Adding Levent https://github.com/lebreeze

NickWeir63:
@levent

myriamboure:
> Currently, only enterprise emails are validated with a confirmation email. This can lead to enterprise owners with invalid email addresses in the system.
@mkllnk I don't think this is true anymore as the owner email can only be chosen among the verified email list. And you cannot add a manager if the email has not been verified. I suggest we close this issue, I'm referencing all the issues in a broader spec document on the user verifications / workflow.

daniellemoorhead:
@sstead are you able to respond to @myriamboure's comment above, is this the case?

sstead:
Yes, I agree with Myriam, enterprise owners must be validated. Shoppers are not validated yet.

mkllnk:
@myriamboure It has been a long time. I can't remember what the loophole was and if that has changed. But we still want to verify the email addresses of customers, right? Maybe we should change the motivation for this change, but we still want to do this.

daniellemoorhead:
Closing this issue as it has been superseded by epic #946 and all accompanying issues within it. Yay!

PR #1495, opened 2017-03-21
Ensure created users are associated with customers

### LINKED ISSUE #1493: Private shop bug 1: if I add a customer before he has an OFN account, he can't access the shop after sign up.
**Case one:**
As a hub manage I add the email myriam[a]ouishare.net in the customer list of the hub named "AMAP Caen 1". Then in the hub menu I check "require customers to login".
In another browser, I login, then clic on the shop, I can access it.

**Case two:**
As a hub manager, I fill in for the same hub the customer list with a new name: olivier_krener[a]hotmail.com, this customer has not yet an account on OFN. So when he clics on the shop link, he sees this message asking him to connect if he has an account, or login.
screenshot from 2017-03-19 22-24-21

He doesn't have an account, so he clicks on "signup" (s'inscrire) and create an account with the same email olivier_krener[a]hotmail.com.
But then it's impossible to access the shop, when I click on the shop page I see that:
screenshot from 2017-03-19 22-24-57

The message saying that the shop is only for members.

I can check in the customer list for the hub, the email is in the list...
screenshot from 2017-03-19 22-23-54

So if I enter an email for someone who has not yet an account in a customer list, there is a bug, he can't access the shop if he creates his account after... 

**You should be able to add a customer to the customer list before they are a user on the OFN, then when they signup, they should get 'customer' privileges.**

Em-AK:
fixes #1493

When a user is created with the same email as existing customers
Then the user is associated with these customers

So that the user can access the private shops where he has been invited
to before signup

### DIFF
--- a/app/models/spree/user_decorator.rb
+++ b/app/models/spree/user_decorator.rb
@@ -19,6 +19,7 @@ Spree.user_class.class_eval do
 
   attr_accessible :enterprise_ids, :enterprise_roles_attributes, :enterprise_limit, :bill_address_attributes, :ship_address_attributes
   after_create :send_signup_confirmation
+  after_create :associate_customers
 
   validate :limit_owned_enterprises
 
@@ -47,6 +48,10 @@ Spree.user_class.class_eval do
 
   def send_signup_confirmation
     Delayed::Job.enqueue ConfirmSignupJob.new(id)
+  end
+
+  def associate_customers
+    self.customers = Customer.where(email: email)
   end
 
   def can_own_more_enterprises?
--- a/spec/models/spree/user_spec.rb
+++ b/spec/models/spree/user_spec.rb
@@ -1,3 +1,5 @@
+require 'spec_helper'
+
 describe Spree.user_class do
   include AuthenticationWorkflow
 
@@ -70,11 +72,22 @@ describe Spree.user_class do
     end
   end
 
-  context "#create" do
+  describe "#create" do
     it "should send a signup email" do
       expect do
         create(:user)
       end.to enqueue_job ConfirmSignupJob
+    end
+
+    context "called with the the same email as existing customers" do
+      let(:email) { Faker::Internet.email }
+      let!(:c1) { create(:customer, user: nil, email: email) }
+      let!(:c2) { create(:customer, user: nil, email: email) }
+      let!(:u) { create(:user, email: email) }
+
+      it "should associate these customers with the created user" do
+        u.customers(:reload).should include c1, c2
+      end
     end
   end
 


myriamboure:
Hum... is there a workaround @mkllnk ? It doesn't seem so easy to fix, no @Em-AK ? So if it's not possible then we should remove from the user guide and features description that we can operate private shops I guess for the moment ;-) For my user in France I will try to see if we have a functional workaround with tags... Thanks

Em-AK:
Thanks @mkllnk to pointing me to these issues. I will look at the details.

mkllnk:
@myriamboure You can have private shops. But if people don't have a user account yet and then sign up for one, they are two separate customers. One is the guest user and the other one is the signed up user. After a user signed up, you have to add that user as a customer. It's just that we can't combine user accounts without verifying the email address.

myriamboure:
@mkllnk this is a real issue, because then there are only two ways for a members only hub to have a private shop:
- either they ask the members to create themselves an account and tell them when it's done, then they need to add them manually in the customer list, and then they can access the shop. it's not at all user friendly, and requires lots of manual work (imagine here, the buying group who wants that has 450 members...)
- or they have to let the shop "public" for the first cycles, check manually that the ones who by are in the members list, tag them, then compare who has not yet bought to manage them one by one so that they create their account...
So I won't say that the private shop feature is operational then ;-)
I guess it's the same with tags... it won't work better if we fill in people before they have created an account, the tag won't apply when they create their account as it will be considered as another one, right @mkllnk ?

The desired behaviour is:
> **As a private hub manager (member only), I want to be able to upload my members list** (ok  if one by one at the begining, as it is now).
> Then **two options**:
- when i save, **if the user is not recognized an email in the name of the hub should be sent to require the user to create an account to access the hubs private shop** (if he has not created it). In the form, the email field is already filled in with the given email. When he has signed-up (or if he has already an account) he receives a message saying he has been added to the hub's private shop buyers list.
- or **the user doesn't receive any message, but when clicking on the shop list the first time, he has the message telling him to log-in or sign-up, and if sign-up, open the form with the email pre-filled and doesn't create a new user** (maybe less user friendly but simpler, very close to what it is today, so I would say let's keep that as a level one target...)

I understand the issue is linked to user verification, it doesn't seem trivial, I'll start to get my head on it too to understand the functional need. Is this improved in the spree upgrade @mkllnk ?

mkllnk:
Hi @myriamboure, sorry for taking so much time to reply. You are right, the current work-flow is not very practical. But I think that the easiest solution is actually verifying users' email addresses (#833). Everything becomes much easier then.

daniellemoorhead:
Hi @Em-AK @myriamboure @mkllnk, this issue has just increased in importance as you can see in @sstead's comment on the parent issue (#1493). We'll need to implement a fix for this bug so that it doesn't impact/delay the standing orders release and we need to have it done in the next 1-2 weeks. 

So, we have a dilemma. We either go with this solution which is not ideal (as you've outlined above @myriamboure). Or we find some money to pay for the work to add an email verification step into the sign up process and get that work happening ASAP. 

Thoughts?

myriamboure:
@daniellemoorhead we have an OFFrance hangout next week to see how to allocate the small money that starts to come in through donation. We might be able to put a few hundreds euros to contribute to it, do you have an idea of the cost ? We do have another priority though in our pipeline which is to solve the separator issue (which is really annoying!) but I'm happy to review the priority if there is a general need for this one... I think we could probably put 500€ on it (we only received 800 for now and need to pay some server cost locally)...this is not a lot I guess... I wanted to propose to allocate for the next three months 800€ to the code review and merge task, but we can switch for that one or we can use the core budget to finance it ! If we get the grants we applied for we'll be able to contribute MUCH more, but we won't know before June.

sineadfenton:
[customer unable to access shop](https://trello.com/c/qC6PXLgB/87-customer-unable-to-access-shop)

sauloperez:
Can we give this a bit more priority? it just happened me again for a new member of the buying group. In this case, it's easy because I can log into production and quickly fix the issue with the following query:

```
UPDATE customers SET user_id = (SELECT id FROM spree_users where email = 'spree@example.com') where email = 'spree@example.com';
```

But for other buying groups this critical use case may cause them to abandon the platform. Without mentioning it's greatly affecting may free time :trollface:

myriamboure:
@Matt-Yorkley can you start with that in the next things to do connected to the email confirmation project? We needed to have email confirmation in place to be able to fix that bug (private shop bug 1)... Personnaly I also need that in priority, as that's because of it that we launched the email confirmation project :-)

Matt-Yorkley:
Is this PR ready now that email confirmation is in place? Maybe it just needs a quick rebase and conflict fix. Should I take it on, or is @Em-AK available?

myriamboure:
Please take it on @Matt-Yorkley !!! @Em-AK is not active anymore on the project. Thanks a LOT !

Matt-Yorkley:
See #2084.

sauloperez:
:heart: @Matt-Yorkley thank so much. I'll spend less time issuing `UPDATE`s in the production's DB during the weekends :trollface:

daniellemoorhead:
Closing this as replaced by #2084

PR #1496, opened 2017-03-23
Some fixes for previous invoice PR

ltrls:
I'm trying to fix the remaining issues in the invoice.
As Rob had warned us, back-calculating the tax rate from the `included_tax` parameter of adjustments was a bad idea. Not only it' a disturbing idead conceptually, but it also gives wrong results with small amounts of tax being rounded to cents...

In the idea of still trying to display those tax rates, there are some stuff we can do that I'd like to discuss.
There are 3 kinds of adjustments that are non-tax adjustments with included tax: enterprise fees, shipping fees and adjustments added via the admin panel.

About enterprise fees, I moved the method responsible for finding tax rates to apply in the adjustment decorator, so that we can also use it afterwards.

About shipping fees, we might want to have a relation between shipping methods and tax categories, like products and enterprise fees have, what do you think?

About adjustments added via the admin panel, only way I thought was that we only allow to add another enterprise fee, but not a custom adjustment. but this removes some flexibility...

To get correct results in the meantime, I've improved the "back-calcutation" by finding the closest tax rate in database.

There are 2 other shortcomings:

Rob mentioned that I might have different tax rates on a line item and its corresponding enterprise fee. According to Myriam this would not usually happen as the fee on a line item is included in the displayed line item price, so it would be a nonsense not to have it "inherit the tax rate".

And there's the thing that these adjustments with included tax can't have excluded tax. Would it be a good thing if their corresponding taxes would only increment the already created (and displayed) tax adjustments if they're supposed to be excluded from price?

### DIFF
--- a/app/helpers/checkout_helper.rb
+++ b/app/helpers/checkout_helper.rb
@@ -54,12 +54,9 @@ module CheckoutHelper
   end
 
   def display_adjustment_tax_rates(adjustment)
-    tax_rate = (adjustment.included_tax / (adjustment.amount - adjustment.included_tax)).round(2)
-    if tax_rate == 0 || tax_rate.infinite?
-      ""
-    else
-      number_to_percentage(tax_rate * 100, :precision => 1)
-    end
+    tax_rates = adjustment.tax_rates
+    return "" if adjustment.amount == adjustment.included_tax
+    tax_rates.map { |tr| number_to_percentage(tr.amount * 100, :precision => 1) }.join(", ")
   end
 
   def display_adjustment_amount(adjustment)
--- a/app/models/spree/adjustment_decorator.rb
+++ b/app/models/spree/adjustment_decorator.rb
@@ -34,9 +34,29 @@ module Spree
       included_tax > 0
     end
 
-    def display_included_tax
-      Spree::Money.new(included_tax, { :currency => currency })
+    def tax_rates
+      case originator
+      when Spree::TaxRate
+        [originator]
+      when EnterpriseFee
+        case source
+        when Spree::LineItem
+          tax_category = originator.inherits_tax_category? ? source.product.tax_category : originator.tax_category
+          return tax_category ? tax_category.tax_rates.match(source.order) : []
+        when Spree::Order
+          return originator.tax_category ? originator.tax_category.tax_rates.match(source) : []
+        end
+      else
+        [find_closest_tax_rate_from_included_tax]
+      end
     end
+
+    def find_closest_tax_rate_from_included_tax
+      approximation = (included_tax / (amount - included_tax))
+      return nil if approximation.infinite? or approximation.zero?
+      Spree::TaxRate.order("ABS(amount - #{approximation})").first
+    end
+
 
     def self.without_callbacks
       skip_callback :save, :after, :update_adjustable
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -234,12 +234,12 @@ Spree::Order.class_eval do
 
   def tax_adjustment_totals
     tax_adjustments.each_with_object(Hash.new) do |adjustment, hash|
-      if adjustment.originator_type == "Spree::TaxRate"
-        tax_rate = adjustment.originator.amount
-      else
-        tax_rate = (adjustment.included_tax / (adjustment.amount - adjustment.included_tax)).round(2)
-      end
-      hash.update({tax_rate => adjustment.included_tax}) { |_tax_rate, amount1, amount2| amount1 + amount2 }
+      tax_rates = adjustment.tax_rates
+      tax_rates_hash = Hash[tax_rates.collect do |tax_rate|
+        tax_amount = tax_rates.one? ? adjustment.included_tax : tax_rate.compute_tax(adjustment.amount)
+        [tax_rate.amount, tax_amount]
+      end]
+      hash.update(tax_rates_hash) { |_tax_rate, amount1, amount2| amount1 + amount2 }
     end
   end
 
--- a/lib/open_food_network/enterprise_fee_applicator.rb
+++ b/lib/open_food_network/enterprise_fee_applicator.rb
@@ -32,20 +32,10 @@ module OpenFoodNetwork
     end
 
     def adjustment_tax(adjustable, adjustment)
-      tax_rates = rates_for(adjustable)
+      tax_rates = adjustment.tax_rates
 
       tax_rates.select(&:included_in_price).sum do |rate|
         rate.compute_tax adjustment.amount
-      end
-    end
-
-    def rates_for(adjustable)
-      case adjustable
-      when Spree::LineItem
-        tax_category = enterprise_fee.inherits_tax_category? ? adjustable.product.tax_category : enterprise_fee.tax_category
-        return tax_category ? tax_category.tax_rates.match(adjustable.order) : []
-      when Spree::Order
-        return enterprise_fee.tax_category ? enterprise_fee.tax_category.tax_rates.match(adjustable) : []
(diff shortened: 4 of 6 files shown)


enricostano:
Maybe we could even have a `TaxRate` model that encapsulates all this logic and split all those `case` branches into single classes that share the same interface.

In any case... it's cool moving business logic near to its business domain 👍

myriamboure:
This PR is a patch as the aggregated tax lines are not working properly.

About what @ltrls said:
"Rob mentioned that I might have different tax rates on a line item and its corresponding enterprise fee. According to Myriam this would not usually happen as the fee on a line item is included in the displayed line item price, so it would be a nonsense not to have it "inherit the tax rate"."
What I meant is that on the item line, like "carrots", a carrot is always sold with a given tax rate, so it there is a fee on it, either it's included in the carrot markup and then "inherit from product" tax applies, either it's considered as a service like packing fee, and then it would be another line on the invoice. 

But anyway, here we only need that to be merged so that we fix the bug on the invoice.

myriamboure:
Just added this is possible to add as it's a patch to something not working on the invoice in aggregated tax calculation @daniellemoorhead

daniellemoorhead:
There is a failing check @myriamboure @ltrls. Best to sort this out before we put it through to @oeoeaio to review. Cheers :)

myriamboure:
Here it's done @daniellemoorhead, @ltrls fixed it :-) Thanks!

daniellemoorhead:
Hey there @myriamboure @ltrls does this PR connect to an issue? If it doesn't that's ok, am just going through the motions of connecting any PRs to issues for 1.8.9 and wasn't sure which one this connected to. Cheers!

myriamboure:
No @daniellemoorhead I had not opened an issue I should have propably ;-) Should I do it now so that you can link it?

daniellemoorhead:
Hmmm. I think we can do with it only if @sstead has enough information within this PR to not need an issue attached to it. I'm assuming that even though this functionality is only being used by France that Sally still includes it in the release notes and so will need to understand what the change is.

myriamboure:
@daniellemoorhead I created an issue corresponding to that PR, it's really a bug fix. https://github.com/openfoodfoundation/openfoodnetwork/issues/1509
Also I would love to review our process with you @daniellemoorhead, maybe we should always deploy on our French staging before pushing anything for review and marge, we had not done it at the begining but now @pacodelaluna has learned how to upgrade the staging and we are doing it from now. But maybe need a quick "restrospective" to smoother our process;-)

daniellemoorhead:
> maybe we should always deploy on our French staging before pushing anything for review and marge

This sounds like the best approach, definitely!

oeoeaio:
@myriamboure @ltrls 
> About shipping fees, we might want to have a relation between shipping methods and tax categories, like products and enterprise fees have, what do you think?

Yep this is coming with the Spree upgrade

> What I meant is that on the item line, like "carrots", a carrot is always sold with a given tax rate, so it there is a fee on it, either it's included in the carrot markup and then "inherit from product" tax applies, either it's considered as a service like packing fee, and then it would be another line on the invoice.

Yes, I can see how that makes sense, and perhaps that the right way to do it, but that is not how it works at the moment. You can add a "service" fee to a product that will appear included in the price in the shop, not as a separate item on the order/invoice. This fee can have a different tax rate to the product.

> And there's the thing that these adjustments with included tax can't have excluded tax. Would it be a good thing if their corresponding taxes would only increment the already created (and displayed) tax adjustments if they're supposed to be excluded from price?

I'm not sure I understand exactly, but taxes included in price are currently treated completely differently from excluded taxes. Excluded taxes are lumped together as one big adjustment on the order, which has not 'included tax', it is a just a big lump of tax that increases the price of the order. The Spree upgrade will change this, but for now that is how it works...

myriamboure:
> Yes, I can see how that makes sense, and perhaps that the right way to do it, but that is not how it works at the moment. You can add a "service" fee to a product that will appear included in the price in the shop, not as a separate item on the order/invoice. This fee can have a different tax rate to the product.

I understand @oeoeaio but as no hub will do it as it doesn't make any sense, shall we take that into account? I just don't see then how to display a tax rate on a product line item... it just becomes purely impossible. Or you need to check "if the fee on the product line is not "inherit from product", then create a new line on the invoice with fee name as title and sum up the amount and display tax rate? But does it need the pain to do that as no one is going to use it basically?
Do you have a concrete real life case? I don't imagine one, so hard for me to visualize...

If needed we can just say,, as we already said this alternative invoice is only for "excluded tax" set up, we can say that entreprise fees applied to products should always be set up as "tax inherited from product tax rate"... just trying to find quick fix until we dive more deeply into the tax overhaul file :-)

oeoeaio:
@ltrls Just noting that I this is with you currently. Test coverage of these sorts of components is important, especially if we are anticipating making changes to Spree, etc.

@myriamboure As I said: "Yes, I can see how that makes sense, and perhaps that is the right way to do it". I don't think this particular issue is worth arguing about because we are setting this invoice up for you, so it should work however you want it to. We are not contemplating solving this problem for the rest of the world until we have completed the Spree upgrade.

My issue is not so much that I am defending a 'real life case', it is that by assuming users will employ conventions that are not actually enforced by the code, we create a broken system. For example: for a layperson who does not understand fully how tax works in Australia, who wanted to make their pricing transparent, it might seem reasonable to add a transport of handing fee to the price of their product (which should include 10% GST if these are 'services'). The product itself would therefore not be taxed, but it's sale price would be inclusive of a fee that does attract tax. As you say: reporting this on an invoice is problematic. I think this is more a problem with the structure of fees than with invoices or this PR, but it is a problem nonetheless.

daniellemoorhead:
Heya @ltrls and @myriamboure, unfortunately @oeoeaio won't have any further time available this week - his focus needs to be on getting the beta standing orders out the door and also stripe payment methods. We're releasing 1.8.9 this afternoon and then making a massive effort to get v1.9 (stripe) and beta 1.10 (standing orders) out the door by friday. Wish us luck!

ltrls:
Added tests. Thanks for nudging me on that, I really need to get used to it!!

daniellemoorhead:
@oeoeaio tests have been added, is this now back to us for review/test?

myriamboure:
@ltrls it seems @oeoeaio requested some changes, can you check that please?

myriamboure:
@daniellemoorhead I see the milestones is 30th April now for this PR that is kind of urgent for us (a user is giving false invoices to his clients... which is annoying and illegal...), from what I understood the next release will be 28th April, right? No release planned before ?

myriamboure:
@oeoeaio it seems @ltrls had done the changes you requested, great if we can merge for the next release :-)

daniellemoorhead:
@myriamboure definitely no release planned before, Australia had a public holiday on tuesday and @oeoeaio needed to work at his food hub on wednesday instead of working on OFN. We'd also considered delaying by a week because @sstead is on leave and not available to test anything.

However, @oeoeaio has said he'll test it so that it can go out this week. Because he's awesome 😄

kirstenalarsen:
one day, when you meet @oeoeaio, you can give him a really really big hug @myriamboure ;)

myriamboure:
Thank you so much @oeoeaio you are awesome, I send you a big virtual hug :-)

oeoeaio:
Haha, no problems. :)

oeoeaio:
@myriamboure I've released this as part of 1.8.10-rc1

We may add a couple more things to the final 1.8.10 release, but you can safely upgrade to 1.8.10-rc1 now if you need this functionality.

myriamboure:
Thanks a lot @oeoeaio you are awesome :-)

Issue #1453, opened 2017-02-14
Customer Enterprise Search Logic


lin-d-hop:
When a customer user searches for an enterprise the existing logic gives some funny results.

When an enterprise exists that matches the input string, no distance results are shown. These are only shown when no name matches exist. 

Given that many, many enterprises have the name of the town in their enterprise name, this means that when searching a town you don't see all of the enterprises in that town unless no enterprise uses the town name.

**_Example:_**
![oxfordsearch](https://cloud.githubusercontent.com/assets/8669848/22925177/a587dcf8-f29f-11e6-9cec-0f3b3f3be133.png)

As 'Cultivate Oxfordshire Ltd' has Oxford in the name, a search for 'Oxford' does not show any of the other shops in Oxford. 

**_Proposal:_** 
Change so that distanceMatches are shown by default.
@oeoeaio @mkllnk @stveep @myriamboure Does this make sense to you? Do you think this be applied for all instances?

Ping @NickWeir63

**_Notes:_** 
app/assets/javascripts/darkswarm/controllers/enterprises_controller.js.coffee 
$scope.updateVisibleMatches
$scope.distanceMatchesShown

There is also some funny logic going on when you make closed enterprises visible as well... tbc...

mkllnk:
Your proposal sounds logical to me. Is there any performance implication? Well, there are probably a lot of other performance issues on that site...

lin-d-hop:
I suspect there are some performance issues. Ill include this in the investigations. Thanks @maikel 

⁣Sent from BlueMail ​

jeronimo:
After a lot of investigation it appears there only one flag needed to be enabled to get results. There are no tests coverage so it is hard to define if attached behaviour is acceptable?

![screen shot 2017-03-24 at 09 41 30](https://cloud.githubusercontent.com/assets/88912/24289356/dadb0618-1078-11e7-9df1-4b753dbdf1eb.jpg)

Note: Performance should be fine with 50+ shops but might be an issue in the future with 300+ or so.

Matt-Yorkley:
@jeronimo is this PR ready for testing/review? If you ping me or @lin-d-hop or someone from the UK team when you've finished a PR and it's ready, we can add it to the testing pipeline to move things along. :)

lin-d-hop:
[...]

PR #1502, opened 2017-03-25
Show distance matches for #1453

jeronimo:


daniellemoorhead:
@oeoeaio I think you and Sally were testing this last week - is it right to get merged now that you've approved it?

oeoeaio:
We had a look at this, and it actually breaks existing functionality. I'm not 100% that the existing functionality is something that we want, but if it is not then we need to clean up the code.

When the user searches for something, the logic matches on name and sends a query to the Geocoder API to see if a match can be found there. If a name match is found locally, the UI calculates distance relative to the matching hub (based on its address). If no name matches are found locally, the UI shows results based on proximity to whatever match is returned by the maps API (though the actual identity of this match is not transparently displayed to the user).

This PR breaks the part of the UI where the user can opt-in to seeing shops proximal to the first shop matched on name. As @sstead pointed out, this option is pretty weird. Surely what we really want is the ability to search on shop name and location simultaneously and to allow users to opt-in to show results based on a transparently displayed location (returned by the API), regardless of whether a name match is found.

I have done a bit of work to refactor the logic for this section, and have switched to using the Places API instead of the Geocoder, because it allows restriction of results to `locality`. The geocoder just matches anything and doesn't allow much filtering, so the results can be quite unpredictable. This allows us to display the name of a town that matches the query entered by the user, and to allow them to see results for that town if they which to.

My work needs a lot of tidying up, it's just a spike at the moment, but the functionality is close. Progress is here: https://github.com/oeoeaio/openfoodnetwork/tree/uk/1453_enterprice-search-logic

daniellemoorhead:
Where are we going with this one @oeoeaio / @jeronimo? Next steps?

oeoeaio:
My last comment is the current status as far as I am concerned. I am happy to work on it if I can have some time (probably another hour or two). IMHO the improvement in usability is significant and worthwhile.

lin-d-hop:
I would agree that it is worthwhile making this improvement. I also think it is worthwhile having @oeoeaios head on the task and ideally UK would pay for this out of our general support bucket. Obviously this is lower priority than SO/Stripe etc but maybe we can add it to the list upon his return?
Ping @daniellemoorhead 
Thanks :-)

daniellemoorhead:
@oeoeaio this is all yours to sort out once you've got all the stripe and standing orders sorted (or as a procrastination tool if you need one in between 😁  )

lin-d-hop:
Closing. Captured a link to Robs spike for reference here:

PR #1505, opened 2017-03-28
Always use english locale in test environment

Em-AK:
https://openfoodnetwork.slack.com/archives/C2GQ45KNU/p1490730191139063

Force test runners to use the english locale. Whatever language is set in `config/application.yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,14 @@ The site is configured to use
 startup time while Rails loads. See the Zeus github page for
 usage instructions.
 
+Once [npm dependencies are
+installed](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Karma), AngularJS tests can be run with:
+
+    ./script/karma run
+
+If you want karma to automatically rerun the tests on file modification, use:
+
+    ./script/karma start
 
 ## Credits
 
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -35,6 +35,7 @@ Openfoodnetwork::Application.configure do
 
   # Tests assume English text on the site.
   config.i18n.default_locale = "en"
+  I18n.locale = config.i18n.locale = config.i18n.default_locale
 
   # Use SQL instead of Active Record's schema dumper when creating the test database.
   # This is necessary if your schema can't be completely dumped by the schema dumper,
--- a/lib/tasks/karma.rake
+++ b/lib/tasks/karma.rake
@@ -1,13 +1,21 @@
 namespace :karma  do
-  task :start => :environment do
+  task :start => :environment do |task|
+    continue_only_in_test_env task
     with_tmp_config :start
   end
 
-  task :run => :environment do
+  task :run => :environment do |task|
+    continue_only_in_test_env task
     with_tmp_config :start, "--single-run"
   end
 
   private
+
+  def continue_only_in_test_env task
+    if Rails.env != 'test'
+      raise "Task must be called in test environment:\n  bundle exec rake #{task.name} RAILS_ENV=test"
+    end
+  end
 
   def with_tmp_config(command, args = nil)
     Tempfile.open('karma_unit.js', Rails.root.join('tmp') ) do |f|
--- a/script/ci/run_js_tests.sh
+++ b/script/ci/run_js_tests.sh
@@ -14,4 +14,4 @@ echo "--- Bundling"
 bundle install
 
 echo "--- Running tests"
-bundle exec rake karma:run
+./script/karma run
--- a/script/karma
+++ b/script/karma
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+task="$1"
+
+if [ "$task" = "run" ] || [ "$task" = "start" ]; then
+  exec bundle exec rake "karma:$task" RAILS_ENV=test
+else
+  echo "Usage:"
+  echo "  $0 run   # to run the tests once"
+  echo "  $0 start # to run the tests on every file modification"
+  exit 1
+fi
+


enricostano:
👏

myriamboure:
@Em-AK is it ready to be reviewed by the Aus tech lead?

Em-AK:
@myriamboure Yes it is !

myriamboure:
@daniellemoorhead do you yourself put the milestones for the release in which the PR should be merged? Or should I ? I imagine I should for the "emergency things" that we really need to merge (like bug fix, and important features that clients are waiting for, for ex. But else for this, or for the PR on "reset pasword mail translation", I guess it's better if you put them in the milestones given the global pipeline of the Aus team, no? Still not clear about the process ;-)

daniellemoorhead:
Hey there @myriamboure! I put them into milestones, what I suggest is that you use the priority labels ('p - critical', 'p - important', 'p - low') if there is something you need urgently and @mention me in a comment. That way I know it needs to be put into the next milestone and it may even trigger off a release that week if required. Cheers 😄

daniellemoorhead:
Spoke with @mkllnk about this and advice was that it needs to be reviewed and dev tested so we won't be able to push it through today as part of release 1.8.9. Am adding it to 1.8.10 which will come through in the next couple of weeks.

config/environments/test.rb
```diff
@@ -35,6 +35,7 @@
 
   # Tests assume English text on the site.
   config.i18n.default_locale = "en"
+  I18n.locale = config.i18n.locale = config.i18n.default_locale
```
Em-AK: This file is loaded after `config/application.rb`, where this this trick happens to properly set the locales in delayed_jobs, see 01bf64e6b03658424b9cd547dc6dc2cccdda669f

I just duplicated the line from `config/application.rb` here.

Em-AK:
I think this PR is related to the enforcement of common defaults in tests environments. Work is also happening in #1357 by @Matt-Yorkley , and #1521 by @jeronimo 
cc @enricostano @daniellemoorhead

Em-AK:
I just refactored the karma rake tasks to force their execution in test environment instead of hardcoding the locale in the tasks.

README.md
```diff
@@ -98,6 +98,14 @@ The site is configured to use
 startup time while Rails loads. See the Zeus github page for
 usage instructions.
 
+Once [npm dependencies are
```
mkllnk: Yay, documentation! :+1:

lib/tasks/karma.rake
```diff
@@ -1,14 +1,22 @@
...
+    invoke_in_test_env task.name, :start, "--single-run"
   end
 
   private
 
+  def invoke_in_test_env(task_name, command, args = nil)
+    if Rails.env == 'test'
+      with_tmp_config command, args
+    else
+      system("bundle exec rake #{task_name} RAILS_ENV=test")
```
mkllnk: Good thought. This would happen if you call rake without the test environment or if you use zeus. We basically start the whole bundle with all initializers a second time which is necessary to load it in the right environment. I think we should output at least a warning that this is happening so that people are aware of their mistake. Next time, they can use the `bundle exec rake karma:run RAILS_ENV=test` command straight away and save the load time of initialising the development environment.

Actually, we could just print that error with the command and exit. That is less fault tolerant, but would simplify the code and make people use the right command. Or are there cases in which we intentionally want this system call?

What do you think?

README.md
```diff
@@ -98,6 +98,14 @@ The site is configured to use
...
+installed](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Karma), AngularJS tests can be run with:
+
+    rake karma:run
```
mkllnk: Shouldn't we call it in the right environment?

    rake karma:run RAILS_ENV=test

That would load twice as fast.

mkllnk: We could also add a simple script `script/karma.sh`:

```sh
#!/bin/sh

task="$1"

if [ -z "$task] ]; then
  echo "Usage:"
  echo "  $0 run  # to run the tests once"
  echo "  $0 start # to run the tests on every file modification"
  exit 1
fi

exec bundle exec rake "karma:$task" RAILS_ENV=test
```

daniellemoorhead:
Do we need to consider this branch in relation to the other two when doing the code review @Em-AK and @mkllnk? Is there any crossover or duplication, will they all work ok together or does some work need to be done getting them aligned?

Em-AK:
@daniellemoorhead this branch is compatible with the work done in the others. There should be no conflict or duplication.

README.md
Em-AK: I think it is better to have a script to take care of `RAILS_ENV`. So we just have to write `karma run` or `karma start`

lib/tasks/karma.rake
Em-AK: OK to raise an error here and document the good usage of the command.

script/ci/run_js_tests.sh
```diff
@@ -14,4 +14,4 @@ echo "--- Bundling"
 bundle install
 
 echo "--- Running tests"
-bundle exec rake karma:run
+./script/karma run
```
Em-AK: This should prevent the CI process to fail

lib/tasks/karma.rake
```diff
@@ -1,14 +1,22 @@
...
+  task :run => :environment do |task|
+    continue_only_in_test_env task
     with_tmp_config :start, "--single-run"
   end
 
   private
 
+  def continue_only_in_test_env task
+    if Rails.env != 'test'
+      raise "Task must be called in test environment:\n  bundle exec rake #{task.name} RAILS_ENV=test"
```
mkllnk: Good. That does the job. Do you think the stack trace could be confusing for people? Devs should be okay, it's just a bit of clutter.

script/karma
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+task="$1"
+
+if [ "$task" = "run" ] || [ "$task" = "start" ]; then
+  exec bundle exec rake "karma:$task" RAILS_ENV=test
+else
+  echo "Usage:"
+  echo "  $0 run  # to run the tests once"
```
mkllnk: If you add another space in front of the hash `#`, the comments should align in the same column.

script/karma
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+task="$1"
+
+if [ "$task" = "run" ] || [ "$task" = "start" ]; then
```
mkllnk: Ah, much better this way. :+1:

lib/tasks/karma.rake
Em-AK: Here is what it looks like. It should only happen when devs execute directly the rake task. I'm open to suggestions on this.
```
$ bundle exec rake karma:run
rake aborted!
Task must be called in test environment:
  bundle exec rake karma:run RAILS_ENV=test
/home/emak/code/openfoodnetwork/lib/tasks/karma.rake:16:in `continue_only_in_test_env'
/home/emak/code/openfoodnetwork/lib/tasks/karma.rake:8:in `block (2 levels) in <top (required)>'
/home/emak/.rbenv/versions/2.1.5/bin/ruby_executable_hooks:15:in `eval'
/home/emak/.rbenv/versions/2.1.5/bin/ruby_executable_hooks:15:in `<main>'
Tasks: TOP => karma:run
```

PR #1514, opened 2017-03-30
Set log level to debug for Staging env

pacodelaluna:
Following a discussion about the log level for Staging env on Slack dev channel, it could be nice to be able to set up the different log level following the use of the staging environment.
My first approach was to hard-code it to debug value but it is an issue with people having a lot of traffic on their Staging servers (requiring then lograte, ...).
So I change the approach and set up a new env variable allowing to define the log level for staging, like this anybody can change it following his need.
Tell me if any issue.

### DIFF
--- a/config/application.yml.example
+++ b/config/application.yml.example
@@ -24,3 +24,5 @@ CURRENCY: AUD
 
 # see: https://developers.google.com/maps/documentation/javascript/get-api-key
 # GOOGLE_MAPS_API_KEY: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+
+STAGING_LOG_LEVEL: 'info'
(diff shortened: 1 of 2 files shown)


daniellemoorhead:
Hey @pacodelaluna can you add a bit of detail into the description for this PR? Also, has it been code reviewed and therefore is ready to be tested? It doesn't say in the details who the reviewer was, whether they passed it or not, etc. Would be great if you can assign yourself to these as well, all meta is good meta 😄

pacodelaluna:
Following a discussion about the log level for Staging env on Slack dev channel, it could be nice to be able to set up the different log level following the use of the staging environment.
My first approach was to hard-code it to debug value but it is an issue with people having a lot of traffic on their Staging servers (requiring then lograte, ...).
So I change the approach and set up a new env variable allowing to define the log level for staging, like this anybody can change it following his need.
Tell me if any issue.

daniellemoorhead:
Hiya @pacodelaluna, I'm closing this PR as it's been a while since it was active. Feel free to reopen and recommence work on this if it becomes a priority. Cheers!

#githubcull2018

PR #1521, opened 2017-04-04
Uk/adapt tests

jeronimo:


enricostano:
@jeronimo did you give a look at https://github.com/openfoodfoundation/openfoodnetwork/pull/1521 ?

jeronimo:
Potentially #1357 hardcoding raises issues like in https://community.openfoodnetwork.org/t/newcomers-issues/876 3) point. 
Meaning that ./spec/features/admin/bulk_order_management_spec.rb:435 test fails with the code in #1357 when Australia moves to next day and local machine stays in the current (at least for me).

I would suggest to have as little hardcoding as possible and to have all settings in one file (in our case in application.yml). And eventually I think we will have to write tests where we will test multiple currencies.

enricostano:
> Australia moves to next day and local machine stays in the current (at least for me)

I think that we should all use the same configuration for tests and not relying on `application.yml`. In theory we all should run the tests in the same TZ, locale, etc.

The reasoning behind that is that we should be able to compare tests results. Right now for instance we have differences between Travis CI and our local machines, and some time tests fail in Travis but not locally.

IMHO same environment for everybody + activating random order execution for tests should be the way to go.

More thoughts?

jeronimo:
The problem for me is that #1357 approach fails one test `/spec/features/admin/bulk_order_management_spec.rb:435` with that code when I run test after 4pm on my local machine with UTC time. That's why I am a little bit against it :) Can we fix that test? I couldn't find quick fix for it

I wouldn't mind having one currency/country/state/timezone setting in tests if everyone knows about it. Though it won't be good if some developers will hardcode `$` by accident instead of `£` or `€` and we won't be able to check that and it will go through builds to UK/EU users.
Travis and other CI can have hardcoded Australian settings but for myself I would like to run tests with `£` or `€` in my local machine before making pull request.

enricostano:
> I wouldn't mind having one currency/country/state/timezone setting in tests if everyone knows about it. Though it won't be good if some developers will hardcode $ by accident instead of £ or € and we won't be able to check that and it will go through builds to UK/EU users.

we can enforce this automating bad practices checks in CI

> Travis and other CI can have hardcoded Australian settings but for myself I would like to run tests with £ or € in my local machine before making pull request.

I think we should have some specific test only testing that all the supported localizations works. But the rest of the tests should be in only one locale / currency. In that way you can ensure that everythings works OK in your currency and you don't break others.

What do you think?

enricostano:
This article can shade some light on how system time, selected time and application time is managed in Rails: https://www.michaelcho.me/article/understanding-time-datetime-and-timezones-in-rails

jeronimo:
#1357 approach has failing test for me. 
#1521 passes by. And it doesn't need to add bad practices checks nor adding additional tests for localisation works. At least for now.

Anyhow, it is just my opinion. Maybe others and elders :) could give their opinion :)

daniellemoorhead:
@Matt-Yorkley @jeronimo @enricostano  can the three of you please connect and discuss what should happen with these two PRs (this one and #1357), and then let us know what you decide and we'll move forward with that. We're not sure what to do in the meantime so will await your instructions and not touch either.

spec/features/admin/order_cycles_spec.rb
```diff
@@ -147,9 +147,8 @@
     page.should have_content 'Your order cycle has been created.'
 
     page.should have_selector 'a', text: 'Plums & Avos'
-
-    page.should have_selector "input[value='2040-11-06 06:00:00 +1100']"
-    page.should have_selector "input[value='2040-11-13 17:00:00 +1100']"
+    page.should have_selector "input[value='#{Time.zone.local(2040, 11, 06, 06, 00, 00)}']"
```
mkllnk: Cool. Since these are repeating a few times, it would be even better to define them in the beginning:
```rb
let(:meaningful_time) { Time.zone.local(2040, 11, 06, 06, 00, 00) }
```

What do you think?

Gemfile.lock
```diff
@@ -754,4 +754,4 @@ RUBY VERSION
    ruby 2.1.5p273
 
 BUNDLED WITH
-   1.14.6
+   1.14.3
```
mkllnk: If you squash that into the commit that changed this line in the first place, then you remove this little detour. But I actually don't mind the bundler version here when other parts of the file change as well.

Gemfile
```diff
@@ -115,7 +115,7 @@ end
 
 group :test do
   gem 'webmock'
-
+  gem 'capybara-screenshot'
```
jeronimo: I was trying to squeeze this gem in master for a while now :) Will create new PR!

spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -32,10 +32,10 @@
       it "rounds fee calculations correctly" do
         # $0.86 + 20% = $1.032
         # Fractional cents should be immediately rounded down and not carried through
-        expect(page).to have_selector '.cart-item-price',         text: '$1.03'
-        expect(page).to have_selector '.cart-item-total',         text: '$8.24'
-        expect(page).to have_selector '.order-total.item-total',  text: '$8.24'
-        expect(page).to have_selector '.order-total.grand-total', text: '$8.24'
+        expect(page).to have_selector '.cart-item-price',         text: "#{@currency_symbol}1.03"
```
jeronimo: I agree... Fast and ugly workaround to skip `Capybara.default_max_wait_time = 30` when tests fail.

Unfortunately I couldn't go with `Spree::Money.new(amount).to_s` helper, because it works as full helper, meaning that it will return "$1,223.00" when 1223 is provided and in the code there and it seems that OFN codebase doesn't use `spree_number_to_currency` helper everywhere.

Matt-Yorkley:
Just a quick note to say we already have capybara screenshot functionality, you can use: `save_and_open_screenshot`.

spec/features/consumer/shopping/cart_spec.rb
mkllnk: I don't understand the connection to the wait time here.

If we don't use that helper, what do we use instead? We should probably unify how we display prices and have a helper for that.

jeronimo: We should use that helper. And probably will have to write one for Angular front-end to unify (if there is none). Basically will have to go through each failing test and check what is happening with each method. Just please give me more time for that :)

As for `money` helper in specs I would suggest it is a bit too short and doesn't say what it does, just something with money, it is like `thing(1.03)` and you still have to look what it does in the code (which is not a bad thing). `display_money` would mean that it gives output, right? Maybe `text: with_currency(1.03)` would be more meaningful? Apologies if github's PR comments is not a good place for discussions like this.

mkllnk: `with_currency` sound good.

jeronimo:
@mkllnk and @RohanM please review again when you'll have time! :)

enricostano:
So, we're going back to dynamic specs? I'm really concerned about this.

jeronimo:
UK CI will test £ and UTC, French CI will test € and UTC +2. Australia will test $ with UTC +9.
Developers running tests in their own machines will test the same.

OR someone needs to write tests now to cover all cases where currency or timezone can change.

enricostano:
We have only 1 CI, right? We should at least.

> OR someone needs to write tests now to cover all cases where currency or timezone can change.
Yes, we should have a small suite checking that.

enricostano:
But again, if I'm the only one concerned... go ahead.

mkllnk:
> So, we're going back to dynamic specs? I'm really concerned about this.

Are you concerned about this as the only approach or do you think that it is a bad idea to have flexible specs?

My thought was that this enables some devs to test certain cases on their machine for now. But I don't think that it solves the problem that the application.yml influences the specs. That still needs to be solved. Specs should run with a default locale. Once we enable users to switch locales, we need some specs to test that as well. Some of them would be explicit like "when uk, expect pound" and others could be flexible like "when aus, expect money_with_currency(5)" and "when uk, expect money_with_currency(5)". These could use shared examples. It depends on the logic you are testing.

I agree that not all specs should be flexible. Otherwise we wouldn't pick up certain errors in which the app and the specs deal with dollars while we want to test pounds.

enricostano:
> I don't think that it solves the problem that the application.yml influences the specs

This is my main concern. The specs should be the same in CI as in every development environment. If not we'll start divergencies and can be difficult to spot issues and help others.

> Once we enable users to switch locales, we need some specs to test that as well. Some of them would be explicit like "when uk, expect pound" and others could be flexible like "when aus, expect money_with_currency(5)" and "when uk, expect money_with_currency(5)". These could use shared examples. It depends on the logic you are testing.

Absolutely, yes. I'm not against specs helpers, shared examples, etc. My concern is only related to `application.yml` and other single OFN instance configuration affecting specs.

mkllnk:
Good, we are on the same page. Thank you for the input. This pull request was maybe a bit more than we would like to have long term. Let's focus on becoming independent of application.yml.

PR #1527, opened 2017-04-09
Set order distribution

stveep:
Updated fix for #1446:

- New dedicated view to set distributor and order cycle for the new order. This then redirects to the usual 'new order' page to add line items. Validation error messages are silenced on this redirect.
- This means that prices are correctly scoped to the hub (distributor) when line items are added.


app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -165,6 +165,12 @@ def order_cycle_expired
 
   private
 
+  def ensure_distributor
```
enricostano: Is this used somewhere?

stveep: Oops no - I added the method to the wrong [non admin] orders controller and forgot to remove it.

oeoeaio:
Superseded by #1575

Issue #1416, opened 2017-01-25
Map does not work in Explorer


MagsHall:
On UK production there are no enterprises showing up on the Map when viewing on Internet Explorer 11. It also freezes when you try to zoom in. 

It is working fine in Chrome and Firefox.

Duende13:
http://stackoverflow.com/questions/27261346/custom-svg-markers-wont-display-in-ie-11

with IE11 and svg images two required bits seem to be scaledSize and optimized.

daniellemoorhead:
Hi @Duende13 when looking for advice on GH probably worth @mentioning people so they know to look at it. There are so many issues on here :)

@oeoeaio and @mkllnk and @RohanM any issues with what Suzanna has outlined above?

myriamboure:
Put s3 in severity as doesn't affect "most users" but probably a few of them. Discovery though is a non-critical feature (but important) that is not working for some users, so could be s2 @lin-d-hop what do you think?

RachL:
I'm closing this one #cleanup. Let's reopen if we still have IE11 users / decide it is a browser we want to support.

PR #1528, opened 2017-04-09
Solved #1416 

Duende13:
ie11 and svg images as markers generate javascript errors and icons don't show. png images have not such problems and for the size and the utility given, this solution solves the problem and it's not generating further problems in other browsers. 
Please, @lin-d-hop could you or someone have a look? Thank you!

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/registration/registration_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/registration/registration_controller.js.coffee
@@ -11,5 +11,14 @@ Darkswarm.controller "RegistrationCtrl", ($scope, RegistrationService, Enterpris
   $scope.countries = availableCountries.filter (country) ->
     country.states.length > 0
 
+  $scope.countriesById = $scope.countries.reduce (obj, country) ->
+    obj[country.id] = country
+    obj
+  , {}
+
+  $scope.setDefaultCountry = (id) ->
+    country = $scope.countriesById[id]
+    $scope.enterprise.country = country if country
+
   $scope.countryHasStates = ->
     $scope.enterprise.country.states.length > 0
--- a/app/views/registration/steps/_details.html.haml
+++ b/app/views/registration/steps/_details.html.haml
@@ -54,7 +54,7 @@
             .small-12.medium-8.large-8.columns
               .field
                 %label{ for: 'enterprise_country' } {{'registration_detail_country' | t}}
-                %select.chunky{ id: 'enterprise_country', name: 'country', required: true, ng: { model: 'enterprise.country', options: 'c as c.name for c in countries' } }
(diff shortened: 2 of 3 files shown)


lin-d-hop:
@jeronimo would you take a look at reviewing this please?

Duende13:
Hi Julius (@jeronimo)!
Thank you for the tip about the meta tag, this one I hadn't try it before. I just did it and nothing happened, just a little bit worst, still markers are not showing, javascript errors are coming back and the cherry on top of the cake, the map freezes when one tries to zoom in, below a screenshot. 
iexplorer

I already tried before the width, height changes, and the scaledSize and optimized solution with not luck, which was my best shot as almost everybody agreed on this. 
I know that png are not the best in comparison with the benefits of svg, but being icons, which are not going to be resized or edited, the file size is smaller, they show in all browsers and OS, and javascript errors vanished from IE11, I just saw on png images the best solution that I could find and not new issues raised.

Duende13:
Sure, Julius. 
I'll keep an eye in case something new come, I've read that with Edge loads of problems are not happening for this reason they stop working in problems from ie11 so I don't think we'll see that coming soon, but I'll keep an eye for any updates. :)

Duende13:
I forgot to mention that I tried another solution without success either. It was to put the path of the svg, which in our case is not a solution, due our svgs are composed of several paths
(path as a list of coordinates and other attributes needed to draw the image), our svgs are several
images one on top of each other, each of them with a path, I tried to combine, group and merge them to obtain a single path but the final image had nothing to be with our current icons.

daniellemoorhead:
@mkllnk if you're happy with this fix then it's good to go to @sstead for testing. Cheers!

enricostano:
I tried to open one of our SVG with https://jakearchibald.github.io/svgomg/ and I get an error related to the format. Could be that the SVG files have some issues?

Here you can find more info about SVG format and compatibility https://css-tricks.com/using-svg/

daniellemoorhead:
@lin-d-hop @Duende13 given @enricostano's error noted above does this need to be resolved before it goes through to @sstead?

Duende13:
@enricostano , what you are saying is that the svg images have problems of compatibility and that's the reason that are not showing on the maps? because my approach was to get these icons with as png images, and I don't thing these give any issue. If you think that the problem with svg on the maps is because of the compatibility I'll have a look but I'm not a designer at all (limited skills there). Just confirm this, please.  Thank you!!!

enricostano:
Yep, I would give a look at the SVG format. It seems a bit broken to me.

Duende13:
Nothing, @enricostano  I changed the svgs, I'm able to open them without errors with the online tool you sent, but I've just tested and the same, all browsers ok except ie11, which doesn't show anything at all.

PR #1529, opened 2017-04-10
Remove unused spec support method

enricostano:
It seems that this method is not used anymore:
```
$ git grep stub_authorization
spec/support/request/authentication_workflow.rb:  def stub_authorization!
```
It was introduced by https://github.com/openfoodfoundation/openfoodnetwork/commit/705e247eb8defe4baa16cc1780ba0b1cf8a8e86e

### DIFF
--- a/spec/support/request/authentication_workflow.rb
+++ b/spec/support/request/authentication_workflow.rb
@@ -19,11 +19,6 @@ module AuthenticationWorkflow
     admin_user
   end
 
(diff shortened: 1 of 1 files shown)


enricostano:
@oeoeaio @RohanM can we merge this? Thanks!

enricostano:
@RohanM I guess we can close this then, right?

Issue #1452, opened 2017-02-13
Improve products UI


Matt-Yorkley:
There are a number of fields relating to products that can currently be edited manually in the **edit product** UI that are actually quite complex in how they relate to each other. 

This complexity is handled really well in the **create product** UI by providing a dropdown that maps things like "kg" to "1000" and "mL" to "0.001" for the `variant_unit_scale` field. There are similar (related) issues with how the `display_as`, `variant_unit_name`, and `unit_value` fields are used conditionally or used in calculations for values.

Currently users can edit these values directly, which I think can break things in some circumstances, or produce confusing results in others.

As most of the functionality is already present in the create product form it should be fairly straightforward to include it in the edit product forms.

![screenshot from 2017-02-13 15-17-01](https://cloud.githubusercontent.com/assets/9029026/22889442/1677b3a6-f201-11e6-9109-3a02042d7293.png)

Values for 250ml item:

![screenshot from 2017-02-13 15-27-27](https://cloud.githubusercontent.com/assets/9029026/22889448/1aecdf7e-f201-11e6-9623-63c6e4e83fc8.png)
![screenshot from 2017-02-13 15-29-53](https://cloud.githubusercontent.com/assets/9029026/22889563/7e5c9d10-f201-11e6-8693-06b93c81d116.png)

Matt-Yorkley:
This ties in to #1444 as well: we should probably do something similar with how these same fields are represented in uploaded Product Import files and how they can be converted into the correct values from "human readable" values.

OliverUK:
I have tested this and it looks great. Back office UX makes so much more sense now. I've played with the settings and haven't found any major bugs and it all displays ok in the front. 

The only thing I've found where it can get confusing (and break) is where the unit is "item".
In the variant edit screen the unit description field is left blank. As long as I don't touch it and only use BEP to change things, it's fine. But if I type something into it, I get the buggy results below. I don't know if it would make sense to disable the field in this case?

![screen shot 2017-04-12 at 17 47 28](https://cloud.githubusercontent.com/assets/17590921/24969414/a787ef44-1fa8-11e7-8dc4-acee6f2595f7.png)
![screen shot 2017-04-12 at 17 47 41](https://cloud.githubusercontent.com/assets/17590921/24969413/a785ca7a-1fa8-11e7-923f-c53a0ece98bb.png)
[...]

PR #1531, opened 2017-04-11
Product Edit UX adjustments

Matt-Yorkley:
Addressing #1452. 

Lots of little changes to the the way weights/volumes are displayed to the user in Product and Variant edit screens, to bring things in line with how the units logic actually works, and to make things clearer and more usable.

**Product edit page:**
- Now has a single dropdown for choosing units which then gets translated into the correct values behind the scenes, so user can select "Weight (kg)" instead of "weight" and having to type in "1000" for the unit scale.
- "Variant Unit Name" field is hidden unless "items" is selected as the unit, because this field actually only applies to products using "items" units. Also added a placeholder description to this field to improve clarity.

**Variant edit page**
- Now displays the currently used unit value (eg Volume, mL)
- Applies the selected scaling to the input box, so if Kg is being used, then a "2" in the box means 2 Kg. Previously "2000" would mean 2 Kg and this was not obvious. Likewise for other scales like mL (0.001), the scaling gets applied so the value the user sees actually makes sense, then this gets converted behind the scenes.
- Hides duplicate weight field when weight is used. Products that use "volume" or "items" still have a weight field, but when "weight" is selected as unit type there was 2 weight fields and one of them didn't actually save if it was changed in the UI... also it often showed a different value. This is now hidden when unit type is "weight".
- Added "Kg" to the second weight field for clarity, as it's always in Kilograms by default.
- Unit Description field is now hidden when unit type is "items", as this field only applies to products using "weight" or "volume".

### DIFF
--- a/app/assets/javascripts/admin/products/controllers/edit_units_controller.js.coffee
+++ b/app/assets/javascripts/admin/products/controllers/edit_units_controller.js.coffee
@@ -1,19 +1,17 @@
 angular.module("admin.products").controller "editUnitsCtrl", ($scope, VariantUnitManager) ->
 
-  $scope.product = {
+  $scope.product =
     variant_unit: angular.element('#variant_unit').val()
     variant_unit_scale: angular.element('#variant_unit_scale').val()
-  }
 
   $scope.variant_unit_options = VariantUnitManager.variantUnitOptions()
 
-  $scope.setDropdown = () ->
-    if $scope.product.variant_unit == 'items'
-      $scope.variant_unit_with_scale = 'items'
-    else
-      $scope.variant_unit_with_scale = $scope.product.variant_unit + '_' + $scope.product.variant_unit_scale
+  if $scope.product.variant_unit == 'items'
+    $scope.variant_unit_with_scale = 'items'
+  else
+    $scope.variant_unit_with_scale = $scope.product.variant_unit + '_' + $scope.product.variant_unit_scale
 
-  $scope.setFields = () ->
+  $scope.setFields = ->
     if $scope.variant_unit_with_scale == 'items'
       variant_unit = 'items'
       variant_unit_scale = null
@@ -24,5 +22,3 @@ angular.module("admin.products").controller "editUnitsCtrl", ($scope, VariantUni
 
     $scope.product.variant_unit = variant_unit
     $scope.product.variant_unit_scale = variant_unit_scale
-
-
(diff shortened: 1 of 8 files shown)


daniellemoorhead:
@lin-d-hop you ok sorting out review/testing for this or ready for AU team to get involved?

jeronimo:
A little bit confusing as it looks differently in different places but probably that is the requirement.
Code looks good and it is covered by feature tests.

👍 

P.S. not important but github makes notifications about "No newline at end of file" in `app/assets/javascripts/admin/products/controllers/variant_units_controller.js.coffee` i.e. Sublime text has `"ensure_newline_at_eof_on_save": true` config.

daniellemoorhead:
@oeoeaio over to you to do the final check. Oliver has tested this and his notes are on the issue (#1452) for @sstead to note and that should make her process a lot easier :)

app/assets/javascripts/admin/products/controllers/edit_units_controller.js.coffee
```diff
@@ -0,0 +1,28 @@
+angular.module("admin.products").controller "editUnitsCtrl", ($scope, VariantUnitManager) ->
+
+  $scope.product = {
+    variant_unit: angular.element('#variant_unit').val()
+    variant_unit_scale: angular.element('#variant_unit_scale').val()
+  }
+
+  $scope.variant_unit_options = VariantUnitManager.variantUnitOptions()
+
+  $scope.setDropdown = () ->
```
oeoeaio: No brackets required when no arguments are passed into a function

app/overrides/spree/admin/variants/_form/add_unit_value_and_description.html.haml.deface
```diff
@@ -1,10 +1,13 @@
 / insert_top "[data-hook='admin_variant_form_fields']"
 
 - if product_has_variant_unit_option_type?(@product)
-  .field{"data-hook" => "unit_value"}
-    = f.label :unit_value, "Unit Value"
-    = f.text_field :unit_value, class: "fullwidth"
+  - if @product.variant_unit != 'items'
+    .field{"data-hook" => "unit_value", 'ng-controller' => 'variantUnitsCtrl', 'ng-init' => 'setInitialValues()'}
```
oeoeaio: This is not really what `ng-init` is for. If you want to run something on initialisation, just run it directly in the controller.

app/assets/javascripts/admin/products/controllers/edit_units_controller.js.coffee
```diff
@@ -0,0 +1,28 @@
+angular.module("admin.products").controller "editUnitsCtrl", ($scope, VariantUnitManager) ->
+
+  $scope.product = {
+    variant_unit: angular.element('#variant_unit').val()
+    variant_unit_scale: angular.element('#variant_unit_scale').val()
+  }
```
oeoeaio: Don't need curly braces for object definition

PR #1206, opened 2016-11-09
Option allow change orders

mkllnk:
This is a follow-up of #1141. Hopefully solving:

- #1083 Display products bought in an order cycle in the shop cart dropdown
- #1084 Display products bought in an order cycle in the edit cart page
- #1085 Allow for bought products to be removed from an order on edit cart page
- #1161 Opt-in option for enterprises to allow their customers to make changes to their current order

PR #1536, opened 2017-04-19
Option allow change orders

oeoeaio:
Adds ability for customers to edit orders after they are placed (until order cycle closes), where this is allowed by the shop.

Supersedes #1206


app/controllers/line_items_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  def bought_items
+    return [] unless current_order_cycle && spree_current_user && current_distributor
+    current_order_cycle.items_bought_by_user(spree_current_user, current_distributor)
+  end
+
+  # Override default which just redirects
+  # See Spree::BaseController and Spree::Core::ControllerHelpers::Auth
+  def unauthorized
+    render text: '', status: 403
+  end
```
enricostano: Maybe we could have a JSON API base controller and put this kind of stuff there.

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -202,4 +223,23 @@ def populate_cart hash
   def wrap_json_infinity(n)
     n == Float::INFINITY ? 2147483647 : n
   end
+
+  def order_to_update
+    order = Spree::Order.complete.find_by_number(params[:id])
+    return order if order.andand.changes_allowed? && can?(:update, order)
+    current_order
+  end
```
enricostano: Shouldn't this logic go into the authorization layer? CanCan abilities for instance? What's the difference between `current_order` and `order` here?

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -202,4 +223,23 @@ def populate_cart hash
...
+
+  def check_at_least_one_line_item
+    order = order_to_update
```
enricostano: why this assignation? can't we use `order_to_update` directly?

app/controllers/spree/orders_controller_decorator.rb
enricostano: In any case we probably need to memoize the result of this method with `||=`, right?

app/models/order_cycle.rb
```diff
@@ -250,6 +250,13 @@ def refresh_products_cache
     OpenFoodNetwork::ProductsCache.order_cycle_changed self
   end
 
+  def items_bought_by_user(user, distributor)
+    # The Spree::Order.complete scope only checks for completed_at date, does not ensure state is "complete"
+    orders = Spree::Order.complete.where(state: "complete", user_id: user, distributor_id: distributor, order_cycle_id: self)
+    items = orders.map(&:line_items).flatten
```
enricostano: Here we're introducing a `N+1` query, right? ActiveRecord can help you to avoid it, instead of using `#map()`.

app/serializers/api/variant_serializer.rb
```diff
@@ -23,4 +22,8 @@ def price_with_fees
   def product_name
     object.product.name
   end
+
+  def tag_list
+    object.tag_list if object.is_a? VariantOverride
+  end
```
enricostano: Just a question here. Is `tag_list` implemented in every "type" of variant or just in `VariantOverride`?

If the latter case we can use duck typing here instead of asking for the "type".

spec/controllers/line_items_controller_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+        end
+
+        context "with an order cycle" do
+          before { item.order.update_attributes(order_cycle_id: order_cycle.id) }
+
+          context "without a distributor" do
+            it "denies deletion" do
+              delete :destroy, params
+              expect(response.status).to eq 403
+              expect { item.reload }.to_not raise_error
```
enricostano: What kind of error we're checking here?

app/controllers/spree/orders_controller_decorator.rb
oeoeaio: Good point

app/controllers/spree/orders_controller_decorator.rb
oeoeaio: It's not really authorisation logic, so I don't think it belongs with CanCan. It's trying to work out which order the user wants to update. `current_order` is defined on the session (it's basically the 'cart', which is always present), but if the user submits a specific order `id` with their update request, this method helps interpret that they want to update that order instead of the default.

You're right though, should probably memoize the result.

app/models/order_cycle.rb
oeoeaio: `items = Spree::LineItem.joins(:order).merge(orders)`?

app/serializers/api/variant_serializer.rb
oeoeaio: You mean get rid of this method altogether? That's probably true actually...

spec/controllers/line_items_controller_spec.rb
oeoeaio: Good question! It seems to be checking that this line `item = Spree::LineItem.find(params[:id])` in the destroy method on the controller doesn't raise an error. Which it should never do, really. I think this should probably be using rescued and return a 404...

oeoeaio:
Thanks for the review @enricostano, I've pushed merged this in (to unblock Maikel from getting some other PRs through), but I will implement all of these changes in a separate PR that I am working on now.

app/serializers/api/variant_serializer.rb
oeoeaio: It's not even being used, so I just deleted it

app/models/order_cycle.rb
enricostano: Here you have almost all the available options http://blog.bigbinary.com/2013/07/01/preload-vs-eager-load-vs-joins-vs-includes.html

But yes, something like that would work.

spec/controllers/line_items_controller_spec.rb
enricostano: In this test we're just testing that you get a `403` trying to delete an `item`. No need to double check if the record is still there IMHO.

app/controllers/line_items_controller.rb
```diff
@@ -0,0 +1,39 @@
+class LineItemsController < BaseController
+  respond_to :json
+
+  def index
+    respond_with bought_items, each_serializer: Api::LineItemSerializer
+  end
+
+  def destroy
+    item = Spree::LineItem.find(params[:id])
```
enricostano: You can create a `before_filter` here:
```ruby
before_filter :load_line_item, only: :destroy

def destroy
  authorize! :destroy, @item
end

def load_line_item
  @item = Spree::LineItem.find_by_id(params[:id])
  raise not_found unless @item
end
```

app/controllers/spree/orders_controller_decorator.rb
enricostano: Could you please add this explaination to the method documentation?

oeoeaio: Done. :)

PR #1540, opened 2017-04-19
cosmetic fixes to the ticket view

ltrls:
This is some view-only fixes on the ticket view.
I also added a `format` method on the Spree::Money decorator so that we proxy the call to the Money object, using the Spree::Money format preferences while allowing to override them.


ltrls:
I added the use of the Rails view helper to escape strings for JS that I just found out (and feel pretty stupid not having searched for it before ><)

myriamboure:
We had an issue with printing the ticket in the last release, apparently it was a small thing that @ltrls just fixed quickly (it seems I had not put the labels to the previous PR here... oups !), if it's super quick to review would be great to add it to this 1.8.10 release @daniellemoorhead as our user now has to manually do a hack to be able to print a ticket :-o Hope that's possible !

myriamboure:
Sorry @daniellemoorhead  I meant this one is preventing the user to print the ticket https://github.com/openfoodfoundation/openfoodnetwork/pull/1551 ! But the present one is making some issue in the ticket style... if longer to review I guess the user can do with a non super nice ticket to start with...

daniellemoorhead:
Hi @myriamboure, when it comes to having @oeoeaio look at this and also #1551 we would like to have these wait until he has launched Standing Orders and Stripe payments. We feel these are more important, especially when you have a work around. Please let me know if you think either of these PRs are more important a priority then getting these bigger features done and launched and we'll see what we can do 😄

myriamboure:
This one seems ok to postpone to next release @daniellemoorhead , it's not so nice but it doesn't prevent printing thanks to the workaround... but great if we can merge in the next release, it's supposed to be a small quick fix as well ;-)

PR #1513, opened 2017-03-30
update wkhtmltopdf-binary and wicked_pdf gems

ltrls:
PR for #1511.
It seems that older versions of wicked_pdf had an issue with downloading images uploaded on aws s3 (wrong SSL version).
By updating the gem it fixes the problem and it seems to break nothing!

PR #1550, opened 2017-04-27
Rspec upgrade

jeronimo:
Following https://github.com/openfoodfoundation/openfoodnetwork/pull/1513. This is just gems and with no braking tests. 
Everyone who touches tests would be nice to contribute extra and replacing to `let` and `expect` syntax :)


oeoeaio:
Thanks @jeronimo! You're a superstar! Still a couple of failing specs to looks at it seems?

jeronimo:
Yep. Looking at those. My guess that those slipped through because my dev machine is faster than Travis or they just randomly fail in Travis.

Gemfile
```diff
@@ -3,7 +3,7 @@ ruby "2.1.5"
 
 gem 'rails', '3.2.21'
 gem 'rails-i18n', '~> 3.0.0'
-gem 'i18n', '~> 0.6.11'
+gem 'i18n'
```
enricostano: Is there any reason why we're removing version info?

spec/controllers/base_controller_spec.rb
```diff
@@ -1,9 +1,9 @@
 require 'spec_helper'
 
 describe BaseController, :type => :controller do
-  let(:oc)    { mock_model(OrderCycle) }
-  let(:hub)   { mock_model(Enterprise, ready_for_checkout?: true) }
-  let(:order) { mock_model(Spree::Order, distributor: hub) }
+  let(:oc)    { double(:order_cycle, id: 1) }
+  let(:hub)   { double(:enterprise, ready_for_checkout?: true) }
+  let(:order) { double(:order, distributor: hub) }
```
enricostano: In this way we loose the method / column check on the stubbed model, did you try `instance_double(OrderCycle)`?
https://relishapp.com/rspec/rspec-rails/v/3-5/docs/model-specs/verified-doubles

jeronimo:
@oeoeaio all looks good now. 

Also fixed Intermittently failing spec in `spec/models/product_distribution_spec.rb` which was caused by `factory :enterprise_fee` which takes sets `enterprise { Enterprise.first || ...` which creates mismatch sometimes.

Seems no more randomly failing tests are appearing.

spec/requests/shop_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 describe "Shop API" do
   include ShopWorkflow
+  include RSpec::Rails::RequestExampleGroup # to have get/post methods in not controller
```
RohanM: Given that we have to bend RSpec to our will here a bit, and it's going against convention, I wonder whether it would make more sense to turn these into request specs. What do you think?

spec/mailers/user_mailer_spec.rb
```diff
@@ -11,6 +11,7 @@
     ActionMailer::Base.delivery_method = :test
     ActionMailer::Base.perform_deliveries = true
     ActionMailer::Base.deliveries = []
+    ActionMailer::Base.default_url_options[:host] = Rails.application.routes.default_url_options[:host]
```
RohanM: I notice that this var is set in `config/environments/test.rb`. I wonder if we should set it globally there?

jeronimo: Yeah. But as far as I remember the problem was that in tests Spree config overrides `action_mailer.default_url_options` so that's why I had to override once again to make tests pass.

spec/features/admin/products_spec.rb
```diff
@@ -54,7 +54,7 @@
       product.tax_category_id.should == tax_category.id
       product.shipping_category.should == shipping_category
       product.description.should == "A description..."
-      product.group_buy.should be false
+      product.group_buy.should be_nil
```
jeronimo: Sure, I have edited Rob's commit.

jeronimo:
@RohanM, I apologise for being heedless. Everything was taken into consideration and force pushed .

daniellemoorhead:
Checks have failed :(

jeronimo:
It feels like someone is sabotaging this pull request :) All fixed again but had to add one workaround.

spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -56,6 +56,7 @@
       let(:variant) { product_tax.variants.first }
 
       before do
+        Spree::Config.set allow_backorders: false
```
RohanM: Would we need to restore this to its original value after this spec has run? I assume that it'll maintain its value after the specs in this block.

RohanM:
I've left one last comment :crossed_fingers: about a config variable change in the specs that might leak out of that spec block. Other than that this PR has my :+1: and I'm happy for it to be merged once that's fixed. Thanks for all your work!

spec/features/consumer/shopping/cart_spec.rb
jeronimo: This is super easy to modify.

But I left it as it is as it feels a bit controversial as it seems that business wants `allow_backorders: false` all the time. And we were facing many issues and wasting a lot of time figuring this out. Also, if you would scan all specs you would find more settings override like this. @oeoeaio, @mkllnk, @lin-d-hop, maybe worth more of discussion?

mkllnk: I think it's good practice if specs leave the scene as they found it. I would create an `around` block, storing the original state, setting the wanted state and resetting it after the test. If we want this in all specs, we can do it in the spec helper. Then it's explicit that you can rely on it.

enricostano: I agree with @mkllnk , actually could be a good idea to mock the needed call to the class instead of changing the state with `#set()`.

jeronimo: My preference was to use `allow(Spree::Config).to receive(:allow_backorders).and_return(false)` but unfortunately one of the tests fails in Travis. My another preference was to use `before` and `after` blocks instead of `around`. Hopefully it won't fail.

enricostano: Yep, let's try to use `#allow()` and see how it goes.

sauloperez:
> Everyone who touches tests would be nice to contribute extra and replacing to let and expect syntax :)

https://github.com/yujinakayama/transpec will be of help here

spec/features/consumer/shopping/cart_spec.rb
jeronimo: @enricostano I tried yesterday and didn't work. You don't see the code because I am rebasing to make commits tidy

spec/spec_helper.rb
```diff
@@ -69,6 +69,10 @@
   # rspec-rails.
   config.infer_base_class_for_anonymous_controllers = false
 
+  # TODO - remove this line after Spree upgrade. Current Spree specs do not have `type: :controller` definitions.
+  # This is a workaround to pass tests in Travis i.e. https://travis-ci.org/openfoodfoundation/openfoodnetwork/jobs/235601640
+  config.infer_spec_type_from_file_location!
+
```
oeoeaio: FYI @jeronimo, you can just add `type: :controller` to `spec/controllers/line_items_controller_spec.rb` and it will be fine, no need to use `infer_spec_type_from_file_location`. :)

spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -56,9 +56,14 @@
...
         add_product_to_cart order, product_tax
       end
 
+      after do
+        Spree::Config.set allow_backorders: true
```
mkllnk: This is probably doing the right thing, but not very clean. It would be correct to save the previous state so that you can restore it at this point.

spec/spec_helper.rb
mkllnk: I tried this in https://github.com/jeronimo/openfoodnetwork/compare/rspec-upgrade...mkllnk:rspec-upgrade. But it actually made a spec fail: https://travis-ci.org/mkllnk/openfoodnetwork/jobs/248648740

It's somehow loading a Spree config instead of the OFN config (change of host in url). Any ideas?

@oeoeaio and @enricostano What don't you like about inferring the spec type from the location? Is it considered bad style?

mkllnk:
I rebased your branch to resolve a conflict. I also started resolving some issues, but got stuck (see comment above). Unfortunately, it won't be included in the June release today. I will move it to July so that we have enough time to solve the outstanding issues. I also noticed that the base controller was touched. So I thought it would be good to let Sally check that it's definitely not braking anything.

Gemfile.lock
```diff
@@ -662,7 +677,7 @@ GEM
     xml-simple (1.1.4)
     xpath (2.0.0)
       nokogiri (~> 1.3)
-    zeus (0.15.4)
+    zeus (0.15.13)
```
mkllnk: This version messes up my terminal: https://github.com/burke/zeus/issues/624
Version 0.15.14 works fine for me.

spec/spec_helper.rb
oeoeaio: Mainly that it is not required. I don't think it is the reason you are seeing the failure you are seeing. We are setting `ActionMailer::Base.default_url_options[:host]` and `Rails.application.routes.default_url_options[:host]` in a bunch of different specs, my guess would be that the spec that is failing has jumped from one thread to another on CI and is preceded by a different set of specs than before and is so being executed in a context where the default host has changed... Another case of tech debt that hasn't been addressed...

spec/features/consumer/shopping/cart_spec.rb
sauloperez: [RSpec around](https://relishapp.com/rspec/rspec-core/v/3-0/docs/hooks/around-hooks#invoke-the-example-using-`run()`) is meant for these cases. You would have:

```ruby
around(:example) do |example|
  Spree::Config.set allow_backorders: false
  example.run
  Spree::Config.set allow_backorders: true
end
```

But I wonder why a simple `allow(Spree::Config).to receive(:set).with(:allow_backorders).and_return(false)` doesn't work in the first place, as @enricostano suggested :thinking:

enricostano: keep in mind that stuff like `Spree::Config.set()` is cached by Rails and affects other tests 💥

spec/spec_helper.rb
jeronimo: @oeoeaio and @enricostano , please read the comments regarding `type: :controller` - this is needed because Spree core internal specs does not have it and those .

Gemfile.lock
enricostano: Is this change in `Gemfile.lock` also reflected in `Gemfile`? If we need a specific version of `zeus` please specify that in `Gemfile`

daniellemoorhead:
@mkllnk where are we at with this PR? Was pushed from June release to July, should I push it out again to August?

mkllnk:
The tests are failing and there some conflicting PRs now as well. I will move it to the next milestone and hope that @jeronimo has some time to sort the tests out.

Gemfile.lock
jeronimo: I am not using zeus so not sure what is the problem. And I've heard that @oeoeaio wants to remove it from `Gemfile` so just upgraded it in `Gemfile.lock`

mkllnk:
@jeronimo Any chance you have time to resolve the conflict?

sauloperez:
Is there something I can help with here? We miss the features that RSpec 3 introduced in other PRs.

jeronimo:
I had only 10 minutes this morning to resolve the conflict which did not work out. Will do that in the evening. And I am charging for solving those as they are not my fault.

jeronimo:
@mkllnk solved!

Gemfile
sauloperez: We should fix it to whatever version works for us, as it was before

spec/features/consumer/shopping/cart_spec.rb
oeoeaio: We should just be able to stub these settings.

````ruby
allow(Spree::Config).to receive(: allow_backorders) { true }
````

File: spec/spec_helper.rb (hunk 7 above)
Comment (oeoeaio) [inline spec/spec_helper.rb]: But we don't run Spree's core internal specs....

Comment (oeoeaio) [inline spec/spec_helper.rb]: For my own recollection: the setting of `default_url_options` in specs is removed in `standing-orders`: 'Stop setting host for url helpers in specs'

Comment (jeronimo) [inline spec/spec_helper.rb]: I don't know what you guys have changed in 3 months time.  But I don't add stuff just for my own amusement.

File: spec/features/consumer/shopping/cart_spec.rb (hunk 8 above)
Comment (oeoeaio) [inline spec/features/consumer/shopping/cart_spec.rb]: Ok, so the reason that stubbing `:allow_backorders` doesn't work is that there are two way of accessing these properties 1) `Spree::Config.allow_backorders` and `Spree::Config[:allow_backorders]`. The second way is not going to be affected by the stub, because it's a different method call (`[]`). The code that this particular spec is testing is in `app/overrides/spree/products/_cart_form/only_show_current_distribution_variants.html.haml.deface` and it uses the hash method to access the setting, so that is why it is not working.

File: spec/spec_helper.rb (hunk 7 above)
Comment (oeoeaio) [inline spec/spec_helper.rb]: Sorry @jeronimo, I'm just trying to document the issues (for my own information as much as anyone else's). I'm definitely not suggesting that you should have been aware of work that I was doing on a totally different branch!! 🙂

**Comment by jeronimo** [comment]
@oeoeaio Is there anything I still need to work on this PR? If yes, could you give me a `fix in one go` list because I am lost in all those conversations. Or may I move to another PR?

**Comment by oeoeaio** [comment]
Hey @jeronimo. I think this should be good to go now. 🤞  We'll try and get it merged this week. Thanks for all your work on it! 🙂

File: config/environments/test.rb [hunk 10]
```diff
@@ -44,7 +44,10 @@
 
   # Print deprecation notices to the stderr
   config.active_support.deprecation = :stderr
-  config.action_mailer.default_url_options = { :host => "test.host" }
+
+  # Configure host for mailers to use for url helpers
```
mkllnk: Should we remove this whole comment then?

oeoeaio: Sure, I'll remove it and rebase.

mkllnk:
I removed some deprecation warnings from failing specs and cleaned up a little bit. But I have no idea why these specs are failing in Travis. They are completely fine on my computer. Usual means of waiting for certain elements to appear hasn't worked.

I think the next step would be to put sleep commands into the affected specs to make sure that we are dealing with race conditions. If that helps, we can find out how to replace the sleep commands with more meaningful commands that wait for certain elements. If it doesn't help, there might be something strange going on depending on the order of specs.

Here is a list of specs that failed within the last 8 Travis runs or so:

```
./spec/features/admin/enterprise_fees_spec.rb:57
./spec/features/admin/enterprise_fees_spec.rb:93
./spec/features/admin/enterprise_fees_spec.rb:110   # x2
./spec/features/admin/enterprise_groups_spec.rb:96
./spec/features/admin/order_cycles_spec.rb:486      # x2
./spec/features/admin/orders_spec.rb:119
./spec/features/admin/orders_spec.rb:133            # x6
./spec/features/admin/orders_spec.rb:160            # x4
./spec/features/admin/variant_overrides_spec.rb:335
./spec/features/consumer/registration_spec.rb:143
```

I'm having a break from this.

enricostano:
@mkllnk try to run the specs using `rm -rf tmp/cache/paralleltests/* && bundle exec rspec spec/path/to/specific/spec`

In some case we pollute Rails cache with our specs making them green locally but red on Travis...

mkllnk:
Interesting. I tried that, but all these tests are still passing on my computer. It could just be that Travis is much quicker than my machine.

Is that cache clearing something we should mention in the README file?

sauloperez:
> Is that cache clearing something we should mention in the README file?

until we fix it complete, I'd say yes.

daniellemoorhead:
Not sure what to do with this @mkllnk, I assume it has to go in the september dot point release? If yes, can you please change the milestone? Thanks :)

enricostano:
there is somethig wrong in this branch and/or Travis CI, The red specs are not in the code 😱

mkllnk:
@enricostano When I last worked on it, I found the specs in the code. I changed a few lines so that some line numbers changed. Another thing to be aware of is that Travis tests the merge-commit of pull requests at the time of pushing.

If you would run Travis now, it uses this commit:

```sh
git fetch origin pull/1550/merge
git checkout FETCH_HEAD
```

enricostano:
@mkllnk I never faced this issue before, and in OFN every time I see a red spec in Travis I can go to the exact line and find the spec. I still suspect that something is wrong.

jeronimo:
I've made all changes towards that it would not fail in Travis, though it was not failing on my local env.
And I have seen failing specs that are in spree core gems which are included in Travis but not in mani `spec`` folder.

Options:
1. Rebase from master into this branch broke something.
2. Extra code after my commits broke something.
3. Some changes in Travis broke something.
4. All above.

oeoeaio:
Closing in favour of #1928

PR #1552, opened 2017-05-02
Uk/1031 insufficient stock in cart

jeronimo:
New features are written to double check https://github.com/openfoodfoundation/openfoodnetwork/issues/1031 if functionality is not broken

### DIFF
--- a/config/initializers/user_class_extensions.rb
+++ b/config/initializers/user_class_extensions.rb
@@ -4,7 +4,7 @@ Spree::Core::Engine.config.to_prepare do
 
       # Override of spree method to ignore orders associated with account_invoices
       def last_incomplete_spree_order_with_ignoring_account_invoices
-        account_invoice_order_ids = account_invoices.pluck(:order_id)
+        account_invoice_order_ids = account_invoices.where('order_id IS NOT NULL').pluck(:order_id)
         return last_incomplete_spree_order_without_ignoring_account_invoices if account_invoice_order_ids.empty?
         spree_orders.incomplete.where("id NOT IN (?)", account_invoice_order_ids).order('created_at DESC').first
       end
--- a/config/initializers/user_class_extensions.rb
+++ b/config/initializers/user_class_extensions.rb
@@ -4,7 +4,7 @@ Spree::Core::Engine.config.to_prepare do
 
       # Override of spree method to ignore orders associated with account_invoices
       def last_incomplete_spree_order_with_ignoring_account_invoices
-        account_invoice_order_ids = account_invoices.map(&:order_id)
+        account_invoice_order_ids = account_invoices.pluck(:order_id)
         return last_incomplete_spree_order_without_ignoring_account_invoices if account_invoice_order_ids.empty?
         spree_orders.incomplete.where("id NOT IN (?)", account_invoice_order_ids).order('created_at DESC').first
       end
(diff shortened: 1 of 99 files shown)


daniellemoorhead:
@lin-d-hop let us know when you want @mkllnk to have another look at this now that @jeronimo has picked up where Levent left off :)

lin-d-hop:
Yeah sure, @mkllnk would you take a look at code review?

config/initializers/user_class_extensions.rb
```diff
@@ -3,9 +3,12 @@
     Spree.user_class.class_eval do
 
       # Override of spree method to ignore orders associated with account_invoices
-      def last_incomplete_spree_order
-        spree_orders.incomplete.where("id NOT IN (?)", account_invoices.map(&:order_id)).order('created_at DESC').first
+      def last_incomplete_spree_order_with_ignoring_account_invoices
+        account_invoice_order_ids = account_invoices.map(&:order_id)
```
enricostano: Using `#map()` you lose the ActiveRecord magic generating a longer SQL query than needed. Compare the current query with the one generated by `account_invoices.pluck(:order_id)` [doc](https://apidock.com/rails/v3.2.1/ActiveRecord/Calculations/pluck).

daniellemoorhead:
There are conflicts @lin-d-hop @jeronimo :(

daniellemoorhead:
Just a note - @mkllnk is off on holiday until early June after today so we won't be doing any reviews and merges...so not sure if he'll get to this before then or if we'll have to add it to the queue for Rob over the coming weeks.

jeronimo:
Fixed merge conflict.

lin-d-hop:
@Matt-Yorkley Would be great to move this through despite reduced capacity in the Aus team. Perhaps you could review?
@OliverUK When you get a chance between all the other things you are testing will you retest that this now works as it should?

OliverUK:
@lin-d-hop Mixed results.
Back orders was enabled again on staging. Once disabled, I was unable to double order. Baskets aren't remembered between logins but I assume that's because we are now testing a separate piece of code.  When over-ordering from different browsers, the second orderer gets an error message and I like that the basket then tells me which items is our of stock and which hasn't sufficient stock but some. 
Problems arose when I adjusted my basked to buy only the available item. The amount didn't seem to update properly. I got slug error pages a few times trying to update the cart and trying to go back. Submitting the order didn't work. It just hung.
Right now my order cycle has disappeared. Maybe someone else is deploying. 


screen shot 2017-05-16 at 14 56 47

lin-d-hop:
Oh sorry oliver! That was me! I'll redeploy and leave you to it.

OliverUK:
Happens. I probably won't get the slug then when I try again but I better check that checking out works ok after adjusting the order to the new stock availability.

OliverUK:
No problems this time. I got the right amount and was able to check out the item that was available. Only thing I noticed twice now on staging is that if I untick "on demand" and put something in the stock field instead, then save, it initially comes out a zero stock. It have to re-enter stock and save a second time before it works. Maybe something that may come up again in the future.

oeoeaio:
@jeronimo, I've made a PR against your branch. Let me know what you think.

oeoeaio:
Superseded by #1583

sstead:
Looks fine to be... but this is still happening... is it not part of this issue?

"If I had an item in my cart, then I logged out, then the item was removed from the OC (either just that item or the supplier), then I log in and go to the shopfront- my cart hangs and I'm unable to proceed through checkout. If I instead login and click straight on the edit my cart dropdown link, I can proceed through checkout and purchase something that's no longer in the OC."

Issue #1494, opened 2017-03-20
Delete a customer from your customer list impossible


myriamboure:
## Description
As a manager of a private shop I try to delete a customer from the customer list to remove access to my shop. The system prevents me to delete it, it says there are orders associated, even if the user has never accessed the shopfront ... and even if they were, as a hub manager I need to be able to remove access to my private shop to a current customer.

## Expected Behavior
As a manager of a "distributor" enterprise setup as "private" (= shop only accessible to customers in the customer list of the enterprise), I should be able to remove access to my private shop to some of my current customers.

## Actual Behavior
I can't remove access to my private shop to a current customer.

## Steps to Reproduce
1. Make sure your enterprise is selling, and is "private" (on shop preference)
2. Check that the shop url leads to a page where it says this shop is only for members (something like that) and you have to login to access the shop. If you are logged in but not in the customer list, you see a message saying you don't have access and need to contact shop manager.
3. Add a customer to your customer list.
4. Login with that customer in another browser, access the shop url : you can access.
5. Try to remove the customer from customer list : you see error below

## Animated Gif/Screenshot
![capture du 2019-01-31 12-26-00](https://user-images.githubusercontent.com/8135376/52051290-72337900-2553-11e9-8585-4eca91a7ba4f.png)

## Context
Reported by various users in Can, Aus, Fr, etc.

## Severity
~bug-s2: a non-critical feature is broken, no workaround~

See below for a more detailed discussion around this. Deleting a customer requires a new feature.

## Possible Fix
**Option 1 -** delete a customer from your customer list regardless of whether they have placed orders in your shop or not (soft delete ? so the data remain but it becomes invisible ?)
**Option 2 -** create an 'inactive' customer - that is marked slightly differently on the list or hidden (but can be displayed if needed), that doesn't have access to the shop

myriamboure:
ref discussion on community https://community.openfoodnetwork.org/t/private-shop-issues/866/2

lin-d-hop:
The solution to #1457 should also cover this...

myriamboure:
[...]

PR #1555, opened 2017-05-03
Allow to remove customer if he has no orders with his enterprise

jeronimo:
This is the tricky one because it also removes foreign key between `spree_orders` and `customer` and leaves only index. Please let me know If there were more reasons than "having properly designed database". And hopefully tests will pass in Travis.

In response of #1494.


daniellemoorhead:
They didn't pass @jeronimo :(

enricostano:
In the OFN dev mumble meeting we talked about keeping the foreign key and pondering how to "disable" a customer. Let's talk about that since it seems a pretty urgent issue.

Issue #1556, opened 2017-05-03
Controller Tweak to enable Embedable Pages


lin-d-hop:
To enable pages to be embedded rails controllers need to be tweaked adding the following line before render action of whichever controller actions:
response.headers.delete "X-Frame-Options"

- [x] An admin checkbox for "enable embedded shopfronts"

- [x] A general helper method that checks the admin setting is enabled, then unblocks iframe use

- [x] The method gets called selectively on the necessary controller actions using :before_render

More background here:
https://jerodsanto.net/2013/12/rails-4-let-specific-actions-be-embedded-as-iframes/

lin-d-hop:
Merged!

PR #1557, opened 2017-05-03
Enable iframes for embedded shopfronts

Matt-Yorkley:
Addressing #1556. Preliminary work towards Shopfront Whitelabelling that allows embedding shops in iframes on external sites.

Edit: this branch has expanded to include #1565 and #1566 (Embedded Shopfront functionality).

Functionality in this PR:

- Restricts iframe use via Rails settings (as opposed to server settings which is what we do currently)
- Allows superadmins to enable embedded shopfronts, and specify a whitelist for alowed domains, seperated by a space
- Allowed domains can use embedded shopfronts in external sites by linking to their shopfront page in an iframe and appending `?embedded_shopfront=true` in the url
- Layout is modified for embedded shopfronts with some minor whitelabelling (logo and navigation hidden, footer content mostly hidden)
- When using embedded shopfronts, logging in or out returns the user to the designated shopfront page instead of the OFN home page.
- Most of this is handled by sessions, but there is also some secondary javascript detection to ensure things are displayed properly. For instance if a user has opened a non-embedded OFN page in a separate tab/window but the session is still set to "embedded mode", the javascript will re-adjust the layout and session data accordingly. 

Testing/Implementing notes:

- For this to work the nginx server configuration for the production/staging instance will need to be modified so that the line referring to `'X-Frame-Options' = 'DENY'` is removed/commented out.
- Embedding shopfronts will only work on sites that use HTTPS.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -52,6 +52,7 @@ class ApplicationController < ActionController::Base
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
     return unless Spree::Config[:enable_embedded_shopfronts] and whitelist.present?
+    return if (request.referer and URI(request.referer).scheme != 'https' and !Rails.env.test?)
 
     response.headers.delete 'X-Frame-Options'
     response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,6 +20,28 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def enable_embedded_styles
+    session[:embedded_shopfront] = true
+    render json: {}, status: 200
+  end
+
+  def disable_embedded_styles
+    session.delete :embedded_shopfront
+    session.delete :shopfront_redirect
+    render json: {}, status: 200
+  end
+
+  protected
+
+  def after_sign_in_path_for(resource_or_scope)
+    return session[:shopfront_redirect] if session[:shopfront_redirect]
+    stored_location_for(resource_or_scope) || signed_in_root_path(resource_or_scope)
+  end
+
+  def after_sign_out_path_for(resource_or_scope)
+    session[:shopfront_redirect] ? session[:shopfront_redirect] : root_path
+  end
+
   private
 
   def restrict_iframes
@@ -28,13 +50,32 @@ class ApplicationController < ActionController::Base
   end
 
   def enable_embedded_shopfront
-    return unless Spree::Config[:enable_embedded_shopfronts]
+    whitelist = Spree::Config[:embedded_shopfronts_whitelist]
+    return unless Spree::Config[:enable_embedded_shopfronts] and whitelist.present?
 
-    @session_data = session
-
-    whitelist = Spree::Config[:embedded_shopfronts_whitelist] || "'none'"
     response.headers.delete 'X-Frame-Options'
     response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
+
+    check_embedded_request
+    set_embedded_layout
+  end
+
+  def check_embedded_request
+    return unless params[:embedded_shopfront]
+
+    # Show embedded shopfront CSS
+    session[:embedded_shopfront] = true
+
+    # Get shopfront slug and set redirect path
+    if params[:controller] == 'enterprises' and params[:action] == 'shop' and params[:id]
+      slug = params[:id]
+      session[:shopfront_redirect] = '/' + slug + '/shop?embedded_shopfront=true'
+    end
+  end
+
+  def set_embedded_layout
+    return unless session[:embedded_shopfront]
+    @shopfront_layout = 'embedded'
   end
 
   def action
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,8 @@ require 'open_food_network/referer_parser'
 class ApplicationController < ActionController::Base
   protect_from_forgery
 
+  prepend_before_filter :restrict_iframes
+
   include EnterprisesHelper
   helper CssSplitter::ApplicationHelper
 
@@ -19,6 +21,21 @@ class ApplicationController < ActionController::Base
   end
 
   private
+
+  def restrict_iframes
+    response.headers['X-Frame-Options'] = 'DENY'
+    response.headers['Content-Security-Policy'] = "frame-ancestors 'none'"
+  end
+
+  def enable_embedded_shopfront
+    return unless Spree::Config[:enable_embedded_shopfronts]
+
+    @session_data = session
+
+    whitelist = Spree::Config[:embedded_shopfronts_whitelist] || "'none'"
+    response.headers.delete 'X-Frame-Options'
+    response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
+  end
 
   def action
     params[:action].to_sym
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -38,7 +38,7 @@ class ApplicationController < ActionController::Base
     stored_location_for(resource_or_scope) || signed_in_root_path(resource_or_scope)
   end
 
-  def after_sign_out_path_for(resource_or_scope)
+  def after_sign_out_path_for(_resource_or_scope)
     session[:shopfront_redirect] ? session[:shopfront_redirect] : root_path
   end
 
@@ -51,8 +51,8 @@ class ApplicationController < ActionController::Base
 
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
-    return unless Spree::Config[:enable_embedded_shopfronts] and whitelist.present?
-    return if (request.referer and URI(request.referer).scheme != 'https' and !Rails.env.test?)
+    return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present?
+    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test?
 
     response.headers.delete 'X-Frame-Options'
     response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
@@ -68,7 +68,7 @@ class ApplicationController < ActionController::Base
     session[:embedded_shopfront] = true
 
     # Get shopfront slug and set redirect path
-    if params[:controller] == 'enterprises' and params[:action] == 'shop' and params[:id]
+    if params[:controller] == 'enterprises' && params[:action] == 'shop' && params[:id]
       slug = params[:id]
       session[:shopfront_redirect] = '/' + slug + '/shop?embedded_shopfront=true'
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,6 +20,11 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def shopfront_session
+    session[:safari_fix] = true
+    render 'shop/shopfront_session', layout: false
+  end
+
   def enable_embedded_styles
     session[:embedded_shopfront] = true
     render json: {}, status: 200
--- a/app/controllers/shop_controller.rb
+++ b/app/controllers/shop_controller.rb
@@ -2,8 +2,7 @@ require 'open_food_network/cached_products_renderer'
 
 class ShopController < BaseController
   layout "darkswarm"
-  before_filter :require_distributor_chosen
-  before_filter :set_order_cycles
+  before_filter :require_distributor_chosen, :set_order_cycles, except: :changeable_orders_alert
   before_filter :enable_embedded_shopfront
 
   def show
--- a/app/controllers/shop_controller.rb
+++ b/app/controllers/shop_controller.rb
@@ -4,6 +4,7 @@ class ShopController < BaseController
   layout "darkswarm"
   before_filter :require_distributor_chosen
   before_filter :set_order_cycles
+  before_filter :enable_embedded_shopfront
 
   def show
     redirect_to main_app.enterprise_shop_path(current_distributor)
--- a/app/controllers/shops_controller.rb
+++ b/app/controllers/shops_controller.rb
@@ -4,6 +4,5 @@ class ShopsController < BaseController
   before_filter :enable_embedded_shopfront
 
   def index
-    #@embeddable = "test"
   end
 end
--- a/app/controllers/shops_controller.rb
+++ b/app/controllers/shops_controller.rb
@@ -1,6 +1,9 @@
 class ShopsController < BaseController
   layout 'darkswarm'
 
+  before_filter :enable_embedded_shopfront
+
   def index
+    #@embeddable = "test"
   end
 end
--- a/app/views/layouts/_shopfront_script.html.haml
+++ b/app/views/layouts/_shopfront_script.html.haml
@@ -1,0 +1,29 @@
+:javascript
+  $(document).ready(function() {
+    var in_iframe = function(){
+      try {
+        return window.self !== window.top;
+      } catch (e) {
+        return true;
+      }
+    };
+
+    var embedded_styles_active = $('body.off-canvas').hasClass('embedded');
+
+    var set_shopfront_styles = function(state) {
+      $.ajax({
+        url: '/embedded_shopfront/'+state,
+        type: 'POST'
+      });
+    };
+
+    if (in_iframe() && !embedded_styles_active){
+      $('body.off-canvas').addClass('embedded');
+      set_shopfront_styles('enable');
+    }
+
+    if (!in_iframe() && embedded_styles_active) {
+      $('body.off-canvas').removeClass('embedded');
+      set_shopfront_styles('disable');
+    }
+  });
--- a/app/views/layouts/darkswarm.html.haml
+++ b/app/views/layouts/darkswarm.html.haml
@@ -21,10 +21,12 @@
     = render "layouts/bugherd_script"
     = csrf_meta_tags
 
-  %body.off-canvas{"ng-app" => "Darkswarm"}
+  %body.off-canvas{class: @shopfront_layout, ng: {app: "Darkswarm"}}
     / [if lte IE 8]
       = render partial: "shared/ie_warning"
       = javascript_include_tag "iehack"
+
+    = render "layouts/shopfront_script" if @shopfront_layout
 
     = inject_current_hub
     = inject_json "user", "current_user"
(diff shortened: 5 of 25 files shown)


lin-d-hop:
@jeronimo would you take a look at reviewing this one please?

app/controllers/shop_controller.rb
```diff
@@ -5,6 +5,8 @@ class ShopController < BaseController
   before_filter :require_distributor_chosen
   before_filter :set_order_cycles
 
+  #after_filter :enable_embedded_shopfront
+
```
jeronimo: Only shops controller can be embedded?

app/helpers/enterprises_helper.rb
```diff
@@ -88,4 +88,11 @@ def shop_trial_expired?(enterprise)
   def remaining_trial_days(enterprise)
     distance_of_time_in_words(Time.zone.now, enterprise.shop_trial_start_date + Spree::Config[:shop_trial_length_days].days)
   end
+
+  # Enable embeddeding within iframes on external sites.
+  # Call with :after_filter on controller actions.
+  def enable_embedded_shopfront
```
jeronimo: As for the big feature I would like this to be covered by controllers tests.

Matt-Yorkley: I'll take a look at maybe a spec to test if the http header has been properly changed, but for the most part I don't think we can test this in a spec. It really needs an iframe in a page in a different domain to load the page remotely, and I don't think we can do that with Rspec?

app/controllers/shop_controller.rb
Matt-Yorkley: Ah, I forgot to uncomment that line after testing.

jeronimo: Yep, that's why I was suggesting test coverage :)

app/helpers/enterprises_helper.rb
jeronimo: We could at least create dummy html with iframe to check if it doesn't render layout.
Other option would be to create dummy app (i.e. on heroku) which would accept iframe source as GET param. I know, going outside testing environment into internet might be a bit costly.

Matt-Yorkley: We can do iframe testing seperately, but I don't think we can do it directly inside of RSpec.

Matt-Yorkley:
@jeronimo I've added a basic spec to check the change in the HTTP headers.

app/controllers/shop_controller.rb
Matt-Yorkley: I've changed this line now, but yes it's just the shop and shops controllers that can be directly accessed in an iframe. The way they work is that once you've accessed one page in an iframe initially, you can then navigate to any other page within the iframe.

app/helpers/enterprises_helper.rb
jeronimo: My suggestion was to create dummy.html inside `spec/support` folder (or more appropriate place) which would have iframe with OFN inside. In a similar way as people create dummy Rails apps to test their gems https://gist.github.com/ramontayag/1045123

enricostano:
Did we investigate security issues? I mean, we should allow only trusted web sites to embed OFN in a iframe. Especially for checkout pages, billing, etc.

lin-d-hop:
@enricostano A more thorough security review is coming before we push the announcement too far. In summary we decided the security vulnerabilities are basically the same vulnerabilities as we are already exposing ourselves to, thus for first MVP release we are not going to go over the top.

enricostano:
OK, but we should plan to list the domain names allowed to embed OFN in the configuration file.

Basically an `[]` of `String` instead of the proposed `Boolean`. The default value for the header should always be `DENY` and in some specific endpoint we should use `ALLOW-FROM` adding the urls from the configuration.

More info about the main vulnerability we're exposed:
https://www.owasp.org/index.php/Clickjacking_Defense_Cheat_Sheet

lin-d-hop:
Thanks for your thoughts @enricostano. I've added them to the SRS doc (which is by no means complete... this project is moving fast due to user need). 
In the interest of keeping info in one place could you add to this doc:
https://docs.google.com/document/d/19WKdXcBCqXYzCNtzGhrVWb3tQPSvW1UU7heUT3UvErA/edit

Particularly if you wanted to develop the Risks section you would be most welcome!

Matt-Yorkley:
@jeronimo do you have any suggestions for making a dummy page with an iframe that's accessible from in RSpec/Capybara?

jeronimo:
1. `Capybara.current_session.driver.visit('spec/dummy/embedded_ofn.html')` might do the job, though `iframe src=` would need to be hardcoded, unless with pass it by `embedded_ofn.html?iframe_source=http://test.host` and extract with javascript (not tested and sounds crazy).
2. Defining dummy controller like in http://ruby-journal.com/create-dummy-controller-for-testing-with-rspec-rails/ and assigning layout to `spec/dummy/layout`

So far I don't see why it should not work but I might be missing something.

Matt-Yorkley:
@enricostano If we're going to append a list of every enterprise that wants a shopfront into the headers, we would need a decent way to allow enterprises to enable this, and then store each enterprise's domain name in a queryable list in the database...? What are your thoughts on that?

Matt-Yorkley:
@enricostano Although, in that instance, we couldn't allow enterprises to enable it themsleves, as anyone could create an enterprise and add a malicious domain. So we would have to reserve it for instance admins, and enterprises would have to request to be added, and an admin would have to manage the list...

enricostano:
@Matt-Yorkley exactly, we should manage that list. There are ways to do that dynamically. For instance look at how payment gateways allow you register external pages to embed their payment iframe.

app/controllers/application_controller.rb
```diff
@@ -20,6 +22,21 @@ def set_checkout_redirect
...
+
+  def enable_embedded_shopfront
+    return unless Spree::Config[:enable_embedded_shopfronts]
+
+    @session_data = session
+
+    whitelist = Spree::Config[:embedded_shopfronts_whitelist] || "'none'"
+    response.headers.delete 'X-Frame-Options'
+    response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
+  end
```
enricostano: 😍

Matt-Yorkley:
Some notes on the iframe security headers: 

'X-Frame-Options' is basically deprecated and doesn't allow whitelisting multiple domains. Also doesn't work at all in Chrome with 'ALLOW-FROM <domain>', but still works in some versions of Explorer.

'Content-Security-Policy' with 'frame-ancestors' is the current standard for controlling iframe content, but is not fully implemented on some versions of Explorer. It uses whitelisting with extendible lists of domains.

Any of these headers could potentially be modified by a malicious user in certain circumstances anyway, so the inclusion of these seems to be more of a "best practice" that helps in most cases, as opposed to an impenetrable defense.

Whitelisted domains can be added in the admin settings in a list separated with spaces.

Matt-Yorkley:
@jeronimo I've added a lot more functionality to this branch and specs that go through the shopping process within an iframe. Could you give it another look before we move on to testing?

@lin-d-hop What's the best way to do testing for this one?

spec/dummy/iframe_test.html
```diff
@@ -0,0 +1,8 @@
+<html>
+<head></head>
+<body>
+
+<iframe src="http://localhost:9999/test_enterprise/shop?embedded_shopfront=true" name="test_iframe" class="test_iframe" id="test_iframe" style="width:100%;min-height:35em"></iframe>
+
+</body>
+</html>
```
jeronimo: No newline at end of file ;)

app/controllers/application_controller.rb
```diff
@@ -18,8 +20,64 @@ def set_checkout_redirect
...
+
+  def enable_embedded_shopfront
+    return unless Spree::Config[:enable_embedded_shopfronts]
+
+    whitelist = Spree::Config[:embedded_shopfronts_whitelist] || "'none'"
+    response.headers.delete 'X-Frame-Options'
+    response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
+
+    check_embedded_request
+    set_embedded_layout
```
jeronimo: This is confusing, if there is no whitelist, but still `check_embedded_request` and `set_embedded_layout` are called.

For example - I can call `/shop?embedded_shopfront=true`, and then session variable is set unnecessary and `embedded` css class is set as as well, right?

app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
```diff
@@ -25,9 +25,9 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
...
-      else
-        Loading.message = t 'going_back_to_home_page'
-        location.hash = ""
-        location.pathname = "/"
+#      if location.pathname in ["/", "/checkout"]
+#        Navigation.navigate "/"
+#      else
+#        Loading.message = t 'going_back_to_home_page'
+#        location.hash = ""
+#        location.pathname = "/"
```
jeronimo: Can we remove comments?

app/views/layouts/darkswarm.html.haml
```diff
@@ -21,11 +21,13 @@
     = render "layouts/bugherd_script"
     = csrf_meta_tags
 
-  %body.off-canvas{"ng-app" => "Darkswarm"}
+  %body.off-canvas{class: "#{@shopfront_layout}", ng: {app: "Darkswarm"}}
     / [if lte IE 8]
       = render partial: "shared/ie_warning"
       = javascript_include_tag "iehack"
 
+    = render "layouts/shopfront_script"
```
jeronimo: I would prefer this to be loaded only when embedded shopfronts are enabled

app/views/layouts/_shopfront_script.html.haml
```diff
@@ -0,0 +1,29 @@
...
+
+    var set_shopfront_styles = function(state) {
+      $.ajax({
+        url: '/embedded_shopfront/'+state,
+        type: 'POST'
+      });
+    };
+
+    if (in_iframe() && !embedded_styles_active){
+      $('body.off-canvas').addClass('embedded');
```
jeronimo: The only thing that stops us is `response.header` but generaly I could call `$('body.off-canvas').addClass('embedded')` from `console` or `window.top `and I would have embedded shopfront which looks vulnerable. But maybe I am wrong.

app/controllers/application_controller.rb
```diff
@@ -18,8 +20,64 @@ def set_checkout_redirect
     end
   end
 
+  def enable_embedded_styles
+    session[:embedded_shopfront] = true
+    render json: {}, status: 200
```
jeronimo: This is weird as well as I  can call `$.ajax({url: '/embedded_shopfront/enable',type: 'POST'});` from browser console without any problem to enable session. There are two ways where we can enable it seems vulnerable to hack as well.

Matt-Yorkley: This bit doesn't effect the security headers, it's only used to switch the alternative CSS layout on and off.

app/views/layouts/_shopfront_script.html.haml
Matt-Yorkley: The security headers and the CSS changes are completely separate. This just alters the CSS.

app/controllers/application_controller.rb
Matt-Yorkley: So if someone figures out they can do an ajax request to a non-public path and change the CSS a little bit it's basically ok...

app/controllers/application_controller.rb
Matt-Yorkley: The first method checks for `?embedded_shopfront=true` and sets the session variable to persist the embedded layout state. The second one sets the variable for the view which tells it to use the different CSS layout.

Matt-Yorkley: I've adjusted the logic to return if there is no whitelist.

Matt-Yorkley:
@enricostano do you have time to take a look?

Matt-Yorkley:
@oeoeaio Passing this one over to AUS after testing and review in the UK. Hoping to get it through quickly for a deadline in the UK. 

It's a tricky one to test, as it needs an iframe on an external site with SSL

daniellemoorhead:
@lin-d-hop same issue of where this fits into the order of priority for @oeoeaio's time this week. Awaiting your response :)

app/views/shared/menu/_large_menu.html.haml
```diff
@@ -37,6 +37,12 @@
               = t 'label_about'
 
     %ul.right
+      %li.powered-by
+        %img{src: '/favicon.ico'}
+        %span
+          = t 'powered_by'
+          %a{href: '/'}
```
mkllnk: I haven't tested this yet. Is the homepage allowed to be embedded? Or should we open a new window here? Would we have a problem with the embedded state if people browse full OFN and then go back to the embedded tab?

app/views/shared/menu/_large_menu.html.haml
```diff
@@ -37,6 +37,12 @@
...
+            = t 'the_open_food_network'
```
mkllnk: What do you think about using `Spree::Config.site_name`?

app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
```diff
@@ -25,9 +25,3 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
     isActive: Navigation.isActive
 
     close: ->
-      if location.pathname in ["/", "/checkout"]
```
Matt-Yorkley: Initially I just commented out this section to see if it could be removed safely. It adds weird behaviour, like redirecting to the root/home page any time a login modal is closed regardless of what page the user is on, which wasn't helpful for embedded shopfronts where the idea is to stay on the shopfront page. 

Actually I don't really understand the purpose of this code block in any context...?

I can remove the function, yes.

app/views/shared/menu/_large_menu.html.haml
Matt-Yorkley: This was a last minute addition at Lynne's request. The homepage can be embedded, yes. It's probably best if users don't switch between embedded and non-embedded pages in a single session, but it should be ok.

app/views/shared/menu/_large_menu.html.haml
Matt-Yorkley: That gives us "OFN UK", which is maybe not what we want. I've just noticed there's another translation key for site title, which has "Open Food Network". Would that be better?

app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
mkllnk: I just asked Rob and he pointed at one use case this change breaks. If you visit `/register`, the page loads an empty layout and opens the modal. That is a weird behaviour, but probably comes from a time when everybody liked modals. When you close the modal there, it needs to redirect somewhere. Otherwise we have a blank page and the user is lost. Can you confirm that?

This behaviour doesn't make sense for the login modal, but they share the same code. Should we resolve this issue in a separate pull request?

app/views/shared/menu/_large_menu.html.haml
mkllnk: Yes, if that fits into your requirements @lin-d-hop?

app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
Matt-Yorkley: Ah yes, confirmed the registration page doesn't redirect. I'll take a look at it. Do you think it needs another PR? The change doesn't make much sense outside of this PR...

mkllnk: It does make sense to me to get rid of the annoying redirects from the login modal.

daniellemoorhead:
@mkllnk back to you for review. @Matt-Yorkley - Maikel is away this week but will be back online next Wednesday :)

Matt-Yorkley:
So we've had some bug reports with this (we've pushed it to UK Production and are using it live). After a lot of investigating it looks like it's an issue with certain versions of Safari and the iOS mobile browser having recently added extremely aggressive cookie blocking features, which for some reason block cookie in iframes (which has apparently broken all kinds of sites and caused a lot of trouble).

I'm investigating workarounds for this, but it looks like any possible fixes might involve some complex scripts on the external site that's embedding the shopfront, so there's nothing to add to this PR at the moment...

mkllnk:
So should I ignore this PR until you it's ready again and you ping me?

Matt-Yorkley:
@mkllnk I've finally finished the Safari/iOS iframes issue. I think it's ready now.

app/controllers/shops_controller.rb
```diff
@@ -1,6 +1,9 @@
 class ShopsController < BaseController
   layout 'darkswarm'
 
+  before_filter :enable_embedded_shopfront
+
   def index
+    #@embeddable = "test"
```
mkllnk: Do we need this line?

db/schema.rb
```diff
@@ -123,8 +123,8 @@
   create_table "distributors_shipping_methods", :force => true do |t|
     t.integer  "distributor_id"
     t.integer  "shipping_method_id"
-    t.datetime "created_at",         :null => false
-    t.datetime "updated_at",         :null => false
+    t.datetime "created_at"
```
mkllnk: I'm not sure why your database doesn't have the `not null` attributes set, but we better not lose them in this pull request.

app/views/layouts/darkswarm.html.haml
```diff
@@ -21,11 +21,13 @@
     = render "layouts/bugherd_script"
     = csrf_meta_tags
 
-  %body.off-canvas{"ng-app" => "Darkswarm"}
+  %body.off-canvas{class: "#{@shopfront_layout}", ng: {app: "Darkswarm"}}
```
mkllnk: `class: @shopfront_layout` would also work.

app/views/shop/_blocked_cookies.html.haml
```diff
@@ -0,0 +1,14 @@
+:javascript
+  $(document).ready(function() {
+    document.cookie = 'cookie_test=true';
```
mkllnk: Are we just overwriting any cookies here? What happens to the session id?

Matt-Yorkley: Session id doesn't get overwritten when it's working correctly.

Matt-Yorkley: The javascript `document.cookie` doesn't overwrite other cookies:

https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie

mkllnk:
@sstead We don't have a setup to test the iframe functionality with our staging server. But UK has tested that. We should just double check that these changes are not breaking anything for us. Have a look at the shop fronts, the header and the footer to see if there are any display issues and everything is working. That's it. It should be fine, UK had this in production already.

Issue #1292, opened 2016-12-07
Order delivery shows a dropdown when there are no other options available


sstead:
Recommendation: Don't show dropdown arrow (this is the case across the site, needs to be changed for all instances).
Devices: All
Where: Shopfront
![image](https://cloud.githubusercontent.com/assets/10843056/20951615/e6427fc4-bc7a-11e6-9db3-e8c1b29414cd.png)

sstead:
This issue is still outstanding on mobile and desktop devises. 
If there is only one open order cycle in the shop, the dropdown can be removed.

myriamboure:
@kirstenalarsen has this been included in the new mobile overhaul project? I think the OC dropdown are impacted by UX overhaul so I'm wondering... In that case please can you close that issue? if it is independant of mobile UX overhaul then let's do it here, hopefully someone from Hacktoberfest will work on it ;-)

lin-d-hop:
This has been considered in mobile redesign. Since this is not in the epic I'll assume outdated. #gitcull

RachL:
@lin-d-hop actually this is not linked to mobile. The dropdown is shown in desktop mode as well. 

In the new designs the dropdown is kept, and I don't see a mention saying it shouldn't appear if only one option is available. Maybe something to add in one of the issue for the redesign?
I believe also that then we are missing a design for it. Like do we keep the back color for "ready for" ? etc

![image](https://user-images.githubusercontent.com/6525576/63880361-e17c2900-c9cd-11e9-9e12-60cd57d73711.png)

ping @kirstenalarsen @daniellemoorhead

sigmundpetersen:
I believe this is fixed in the recent round of mobile improvements.

PR #1558, opened 2017-05-03
#1292 Show span label if only one order cycle is available

jeronimo:
#1292


daniellemoorhead:
Closing this as it hasn't been active for a while. Feel free to reopen and complete when you're back onboard :-)

Issue #1315, opened 2016-12-07
If there are no filters for a search term, don't show empty filter headings (eg. filter by property)


sstead:
If there are no filters for a search term, don't show empty filter headings (eg. filter by property in image below)
Devices: Mobile/tablet
Where: Producers
![image](https://cloud.githubusercontent.com/assets/10843056/20953792/02d4df7e-bc8a-11e6-93b0-f7d75bb70aa5.png)

myriamboure:
Iceboxed https://community.openfoodnetwork.org/t/customers-easily-go-through-the-whole-placing-order-process-on-mobile/1244 will work on it when we open the mobile UX topic
#gitcull2018

PR #1559, opened 2017-05-04
Uk/1315 hide empty filter properties

jeronimo:
Regarding #1315 issue.

This is the "best" solution I came up with which still doesn't feel "Angular way". Please guide me if you have better one.

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/filter_selector.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/filter_selector.js.coffee
@@ -1,5 +1,5 @@
 Darkswarm.directive "filterSelector", ->
-  # Automatically builds activeSelectors for taxons
+  # Automatically builds activeSelectors for taxons or properties
   # Lots of magic here
   restrict: 'E'
   replace: true
@@ -40,6 +40,13 @@ Darkswarm.directive "filterSelector", ->
           selector = selectors_by_id[id] = scope.selectorSet.new
             object: object
           selectors.push selector
+
+      # Hide headlines and labels incase there are no options
+      if selectors.length == 0
+        elem.parents('.filter-block').hide()
+      else
+        elem.parents('.filter-block').show()
+
       selectors
 
     scope.ifDefined = (value, if_undefined) ->
--- a/app/views/producers/_filters.html.haml
+++ b/app/views/producers/_filters.html.haml
@@ -7,17 +7,19 @@
   .small-12.columns
     .row.filter-box
       .small-12.columns
-        %h5.tdhead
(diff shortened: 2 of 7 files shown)


jeronimo:
I was hoping to implement in that way as well. But there are more than one `filter-selector` directive in the view and it is needed to hide parent html element of that directive so I saw it as a namespace conflict so that one would override all. Or `allSelectors` is `nil` because it is outside `filter-selector` scope.  I am not sure anymore, will have to check again :)

jeronimo:
@Matt-Yorkley, @oeoeaio Maybe for suggestions how to make the code better?

oeoeaio:
Hey @jeronimo, I would do this by assigning the result of the filter to a variable on the parent scope (using the `as` syntax), and then showing/hiding the whole block based on whether length is non-zero. Something like:

````ruby
  .filter-block{ ng: { show: "filteredTaxons.length > 0" } }
    // title,etc
    %filter-selector{ ...... objects: 'producers_to_filter | searchEnterprises:query | taxonsOf as filteredTaxons' }
````
That way you can avoid the use of jquery in the `filter-selector` directive.

Thoughts?

jeronimo:
@oeoeaio, if I understand your suggestion correctly this is the same I have tried at the beginning and that is what @Matt-Yorkley suggested as well. But if you would read my comments maybe it will explain the problem. Unless I still do not understand Angular Services yet.

oeoeaio:
@jeronimo ok, so the scoping this is pretty interesting here, I have worked with these directives before and they are pretty complex, I'll give you that. You are right in saying that `allSelectors` is not accessible because it is only present inside the `filter-selector` scope.

I am suggesting something different though, and I think it will work. Each `%filter-selector` element declares it's own `objects` using a filter, the result of which can be assigned to a unique variable on the parent scope. So filtered objects from the taxons filter-selector can be assigned to `filteredTaxons` and the filtered objects from the properties `filter-selector` can be assigned to `filteredProperties`. So there should be no conflict. You can then use the length of `filteredTaxons` or `filteredProperties` with an `ng-show` to show or hide the selectors. Does that make sense?

daniellemoorhead:
Removing this from August dot point release, ping us with the aus review label when you're ready to go @jeronimo :)

myriamboure:
@jeronimo any news on that PR ? @lin-d-hop do we move it forward or close it for now and open an icebox issue to improve mobile UX and bug the issue in it?

myriamboure:
Closing as it's been open for ages and no move. Iceboxed the original issue for when we decide to prioritize mobile UX.

Issue #1562, opened 2017-05-05
Add i18n-js Gem


lin-d-hop:
Add Gem for I18n javascript translations
https://github.com/fnando/i18n-js

This Gem should take care of the following steps:
1) We need to pre-compile a Javascript locale file for each language.
That requires one asset file per language.
We can then compile only those locales defined in the app configuration.

2) The Javascript lacks a fall-back feature to show text of the default locale if the selected locale doesn't have the translated text.
For example, if the British English translation is not complete, the missing parts would be displayed in Australian English.

Time Estimate: 13hrs

lin-d-hop:
Closing as merged

PR #1564, opened 2017-05-08
Uk/1562 add i18n js gem

jeronimo:
Task - #1562 

Comments on steps:

1. Did not create precompiled separate files because:
   - would have to leave default assets pipeline.
   - with `config.i18n.available_locales` translations are not that big anymore.
   - it might allow to change translations dynamically.

2. It falls back to default translation and doesn't break javascript. The problem is still with back-end translations inside `placeholders` for input fields if translation is still missing.

### DIFF
--- a/config/initializers/i18n-js.rb
+++ b/config/initializers/i18n-js.rb
@@ -1,0 +1,30 @@
+# This is coppied from https://github.com/fnando/i18n-js/blob/master/lib/i18n/js.rb
+# As in spree core en.yml there are translations -
+# en:
+#   no: "No"
+#   yes: "Yes"
+# Which become to true and false and those have no #to_sym method
+# TODO - remove this after spree core locales are fixed
+
+module I18n
+  module JS
+    # Filter translations according to the specified scope.
+    def self.filter(translations, scopes)
+      scopes = scopes.split(".") if scopes.is_a?(String)
+      scopes = scopes.clone
+      scope = scopes.shift
+      if scope == "*"
+        results = {}
+        translations.each do |scope, translations|
+          tmp = scopes.empty? ? translations : filter(translations, scopes)
+          scope_symbol = scope.respond_to?(:to_sym) ? scope.to_sym : scope.to_s.to_sym
+          results[scope_symbol] = tmp unless tmp.nil?
+        end
+        return results
+      elsif translations.respond_to?(:key?) && translations.key?(scope.to_sym)
+        return {scope.to_sym => scopes.empty? ? translations[scope.to_sym] : filter(translations[scope.to_sym], scopes)}
+      end
+      nil
+    end
+  end
+end
--- a/config/initializers/i18n-js.rb
+++ b/config/initializers/i18n-js.rb
@@ -1,4 +1,4 @@
-# This is coppied from https://github.com/fnando/i18n-js/blob/master/lib/i18n/js.rb
+# This is copied from https://github.com/fnando/i18n-js/blob/master/lib/i18n/js.rb
 # As in spree core en.yml there are translations -
 # en:
 #   no: "No"
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ ruby "2.1.5"
 gem 'rails', '3.2.21'
 gem 'rails-i18n', '~> 3.0.0'
 gem 'i18n', '~> 0.6.11'
+gem 'i18n-js', '~> 3.0.0'
 
 # Patched version. See http://rubysec.com/advisories/CVE-2015-5312/.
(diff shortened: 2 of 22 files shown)


enricostano:
Nice! 👏

daniellemoorhead:
@enricostano you ok to review this before we push it over to Australia for final test/merge?

daniellemoorhead:
Hey @oeoeaio I reckon this one is going to be a quick push through yourself and Sally to get it merged. Will leave you to it :)

jeronimo:
@RohanM, You are totally right it felt like a spike :) At some point it felt right to leave it with that commit history as it would help to understand what is necessary to use from gem's documentation. But apparently not. Fair enough :)
Instead of creating another branch it was easier to squash current one and the PR stays the same.
Old commits' history is here https://github.com/jeronimo/openfoodnetwork/commits/uk/add-i18n-js-gem-spike

lin-d-hop:
@RohanM Are you happy to sign this off now ready for merge at next opportunity?

RohanM:
Hi @jeronimo,

Unfortunately, squashing the commits puts the entire 22 file, 150 line diff in a single commit - that's too much code for someone to easily absorb, especially when that one commit is accomplishing many distinct things. Having the code in one commit obscures the steps that went into achieving the feature. Additionally, in the future we might discover a problem that involves this code. If it's in a single commit, the we're not able to use `git bisect` to narrow down on the cause of the problem, and it's harder for a developer to understand the code.

Because of those reasons, it's useful to break the code down into smaller commits (both for a reviewer and for any future developers who might work with this code). Preferably each commit introduces one part of the change, comes with its own specs, and those specs pass. That way a developer can read the commit message and glance at the code and confirm that the code does indeed do what the commit message says. And because the specs build green at most if not all commits along the way, we can bisect over it and get useful results.

Finally, a comment on the situation where a branch is broken down into multiple commits but changes one thing and then changes it back later. Although this can be useful in describing the process taken to arrive at a feature, it can result in unnecessary conflicts when rebasing. And when multiple changes back and fourth are combined in a single commit, it can be hard to pick apart what's happening. So I'd see this situation as a trade-off, and unless I really wanted to illustrate a path that didn't work out, I'd tend to the side of simply removing those changes from the code, and writing about it in the PR description (which a future developer could look up from the merge commit).

For those reasons, would you be able to rewrite this PR step by step?

jeronimo:
@RohanM, actually I do not agree. Nearly 60% of added code is test coverage. 20% of code is a workaround until Spree gets upgrade. The rest of the code is used from Rails guides and READMEs and this whole commit is like a small feature, which is great for `git bisect`. If you would notice, original PR had 3 commits only, and that was only because I tried various approaches how to use i18n-js gem until I chose best one (IMHO) for our case.

But, if you still do think it is necessary to split, maybe you could give an example with this code for the future reference we could follow?

jeronimo:
@RohanM, @mkllnk, @oeoeaio please have a look now :)

config/initializers/i18n-js.rb
```diff
@@ -0,0 +1,30 @@
+# This is coppied from https://github.com/fnando/i18n-js/blob/master/lib/i18n/js.rb
+# As in spree core en.yml there are translations -
+# en:
+#   no: "No"
+#   yes: "Yes"
+# Which become to true and false and those have no #to_sym method
+# TODO - remove this after spree core locales are fixed
```
mkllnk: Great. This way it will be super easy to see what needs to be changed after the Spree upgrade.

mkllnk:
I think before we merge this, we need to update our deploy scripts. @jeronimo Have a look at: https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/post-receive.j2

We need to add the line:
```sh
$BUNDLE exec rake tmp:cache:clear
```

Can that happen any time or do we need to do it before the asset compilation?

jeronimo:
@mkllnk It seems it has to be before `assets:precompile`. More details inside PR https://github.com/openfoodfoundation/ofn-install/pull/73.

Note: `tmp:cache:clear` need should disappear after upgrade to rails 4.x which will introduce sprockets.

mkllnk:
@sstead This adds a feature to provide more than one language. Currently, our staging server is configured to show only English. Nothing should change. And we don't have a UI element for language switching yet. That comes later.

sstead:
Looks fine.

mkllnk:
Great work @jeronimo.

I needed to add a fall-back strategy for old `application.yml` files that don't have the locale set correctly. Otherwise that i18n-js gem would crash. But that wasn't your fault. Glad to have this merged.

PR #1568, opened 2017-05-12
Allowing order changes tweaks

oeoeaio:
Tweaking a couple of things as picked up by @enricostano in #1536

### DIFF
--- a/app/controllers/line_items_controller.rb
+++ b/app/controllers/line_items_controller.rb
@@ -1,7 +1,10 @@
 class LineItemsController < BaseController
   respond_to :json
 
-  def index
+  # Taken from Spree::Api::BaseController
+  rescue_from ActiveRecord::RecordNotFound, :with => :not_found
+
+  def bought
     respond_with bought_items, each_serializer: Api::LineItemSerializer
   end
 
@@ -20,10 +23,12 @@ class LineItemsController < BaseController
     current_order_cycle.items_bought_by_user(spree_current_user, current_distributor)
   end
 
-  # Override default which just redirects
-  # See Spree::BaseController and Spree::Core::ControllerHelpers::Auth
   def unauthorized
-    render text: '', status: 403
+    render nothing: true, status: 401 and return
+  end
+
+  def not_found
+    render nothing: true, status: 404 and return
   end
 
   def destroy_with_lock(item)
--- a/app/controllers/line_items_controller.rb
+++ b/app/controllers/line_items_controller.rb
@@ -1,21 +1,24 @@
 class LineItemsController < BaseController
   respond_to :json
 
-  # Taken from Spree::Api::BaseController
-  rescue_from ActiveRecord::RecordNotFound, :with => :not_found
+  before_filter :load_line_item, only: :destroy
 
   def bought
     respond_with bought_items, each_serializer: Api::LineItemSerializer
   end
 
   def destroy
-    item = Spree::LineItem.find(params[:id])
-    authorize! :destroy, item
-    destroy_with_lock item
-    respond_with(item)
+    authorize! :destroy, @line_item
+    destroy_with_lock @line_item
+    respond_with(@line_item)
   end
 
   private
+
+  def load_line_item
+    @line_item = Spree::LineItem.find_by_id(params[:id])
+    not_found unless @line_item
+  end
 
   # List all items the user already ordered in the current order cycle
   def bought_items
@@ -24,7 +27,8 @@ class LineItemsController < BaseController
   end
 
   def unauthorized
-    render nothing: true, status: 401 and return
+    status = spree_current_user ? 403 : 401
+    render nothing: true, status: status and return
   end
 
   def not_found
--- a/app/controllers/spree/orders_controller_decorator.rb
+++ b/app/controllers/spree/orders_controller_decorator.rb
@@ -224,22 +224,27 @@ Spree::OrdersController.class_eval do
     n == Float::INFINITY ? 2147483647 : n
   end
 
+  # If a specific order is mentioned, check that it is COMPLETE, that changes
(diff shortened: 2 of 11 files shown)


app/controllers/line_items_controller.rb
```diff
@@ -20,10 +23,12 @@ def bought_items
...
-  # Override default which just redirects
-  # See Spree::BaseController and Spree::Core::ControllerHelpers::Auth
   def unauthorized
-    render text: '', status: 403
+    render nothing: true, status: 401 and return
+  end
+
+  def not_found
+    render nothing: true, status: 404 and return
   end
```
enricostano: if we are going to use `not_found` and `unouthorized` in others API controllers we should probably move them to some OFN base API controller.

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -224,22 +224,27 @@ def wrap_json_infinity(n)
...
-    order = Spree::Order.complete.find_by_number(params[:id])
-    return order if order.andand.changes_allowed? && can?(:update, order)
-    current_order
+    return @order_to_update if defined? @order_to_update
+    return @order_to_update = current_order unless params[:id]
+    @order_to_update = begin
+      order = Spree::Order.complete.find_by_number(params[:id])
+      order = nil unless order.andand.changes_allowed? && can?(:update, order)
+      order
+    end
```
enricostano: can you move this to another method?

oeoeaio: I can, thought it was on the borderline between readability and excessive breakdown of the method.

app/controllers/line_items_controller.rb
```diff
@@ -1,7 +1,10 @@
 class LineItemsController < BaseController
   respond_to :json
 
-  def index
+  # Taken from Spree::Api::BaseController
+  rescue_from ActiveRecord::RecordNotFound, :with => :not_found
```
enricostano: a `before_filter` would be more explicit IMHO, you probably don't want to rescue all the AR::RNF but only known one

https://github.com/openfoodfoundation/openfoodnetwork/pull/1536#discussion_r116174434

app/controllers/line_items_controller.rb
oeoeaio: See my most recent post on discourse. Trying to think about how we structure an API if we're going to build one. Completely take your point, but I'm reluctant to start building something until we have agreement on direction.

spec/controllers/line_items_controller_spec.rb
```diff
@@ -60,8 +61,8 @@
         context "without an order cycle" do
           it "denies deletion" do
             delete :destroy, params
-            expect(response.status).to eq 403
-            expect { item.reload }.to_not raise_error
+            expect(response.status).to eq 401
```
enricostano: shouldn't this be a `403`? `401` is used for unauthenticated requests. `403` means that even if you're logged in you cannot perform that action.

oeoeaio: Well that is correct in this instance. This action can never be performed if a current_order_cycle is not present.

enricostano:
Thanks! 💪

I added more comments both here and in the original PR.

app/controllers/line_items_controller.rb
oeoeaio: Point taken. As you say, this is a prime example of another convention that should be defined in a lower level controller. Until we make a decision on direction, I am more comfortable implementing something that looks a bit more like a convention than a one-off `before_filter`, but perhaps that is not a very good reason to do it this way in this instance...

enricostano: As you wish, I'm generally used to make clear in the controller what each action need and how that affects the response. Basically: `not_found` implemented in a `Base` controller and `before_filter` in endpoint controller.

spec/controllers/line_items_controller_spec.rb
enricostano: Exactly, the reason of the failure is related to some internal business logic / authorization (`403`) not authentication (`401`).

One thing that could lead to confusion is that in the HTTP status protocol definition the therm "Authorization" is referred to what we refer to "Authentication" in web world 😞 

`CanCan` abilities -> authorization layer `403`
`Devise` and friends -> authentication layer `401`

Or am I missing something regarding the spec? From what I understand a logged in user cannot delete a line item if it is not associated to an order cycle.

I'm sorry if I'm getting this wrong, but at least it's useful to me to understand OFN business logic 🙏

app/controllers/spree/orders_controller_decorator.rb
enricostano: I hope we start introducing tools like rubocop and reek and get to an agreement to code style. For me the nested `begin..end` is a clear clue for code split.

One great opportunity that code splitting offers is more documentation surface. That's another thing that would be nice to see in OFN code too 💪

spec/controllers/line_items_controller_spec.rb
oeoeaio: Yep, you're exactly right. My Friday afternoon brain was backwards. Thanks :)

app/controllers/spree/orders_controller_decorator.rb
oeoeaio: Yep, agreed. I found it hard to name the new method, so I thought that was a good indicator that the logic didn't need to be split. I've called the new method `changeable_order_from_number` which I'm not al that happy with, but I think the code it a little more readable...

app/controllers/line_items_controller.rb
oeoeaio: Ok, so this is the one outstanding question I think. The biggest question for me is how a new hierarchy of json-only controllers fits into our current mix. Do we namespace them separately? If so, what are they called?

Issue #1561, opened 2017-05-05
Replace Open/Closed Icons


lin-d-hop:
The "open" and "closed" icons still need to be replaced.
We need to discuss and decide which icons to use.
And we need to replace these icons in the current symbol font.

From @mkllnk via Slack:
Icons are in a font. And we would like an icon that does not contain text. Maybe a first step would be to replace the icon just with a text label. The internationalisation discussion contains a lot of ideas for a symbol that's easy to understand for everyone: https://community.openfoodnetwork.org/t/internationalisation-project-notes/312/59 pp

Hardcode the language set is a viable first option. It can be an improvement later to have an interface for that.

Time Estimate: 6hrs

daniellemoorhead:
Are visual design aspects such as icons worthy of having some @yukoosawa time on?

mkllnk:
That's an excellent idea. We all just came up with ideas, but don't know what the best solution is. If Yuko could come up with a nice design element that uses text separate from a logo that would be great.

lin-d-hop:
Note from Maikel on slack (for posterity):
These symbols are in the custom ofn font created with icomoon.io. The source file is here: https://gist.github.com/mkllnk/5acd096b7d0da97830012068db7f9b56

daniellemoorhead:
Please see remaining work to be done for this issue on PR #1937. @Matt-Yorkley or @sstead it might be worth listing these things out on here so that it's easier to figure it out (unless you're going to quickly do them and send through a new PR Matt, in that case don't worry about it :) )

daniellemoorhead:
Closing this issue as a) the PR has been merged, and b) it's been a while since I added my comment above and so I assume that there's nothing left to do as there was no action.

Issue #1560, opened 2017-05-05
Complete Placeholder Creation


lin-d-hop:
There remain approximately 83 files that require language string placeholders.

Can these all be replaced and the original translation added to en.yml.

Time Estimate: 8hrs

jeronimo:
There is an issue with Spree Core as it loads model decorators before translations are loaded so validations fail to find translations. There is a fix (https://github.com/spree/spree/pull/7469/files) for Spree 3.0 so it might be solved only after that. 

This happens in situations like `validates :distributors, presence: { message: I18n.t(:spree_distributors_error) }` where translation goes into `Proc`. Temporary solution is to create validator classes or to move validations into methods.

daniellemoorhead:
@lin-d-hop @jeronimo the PR for this issue has been merged so feel free to close this :-)

PR #1569, opened 2017-05-12
Uk/1560 extract text into translations

jeronimo:
For issue #1560.

This task was to extract and replace it. Do not blame me for non-existent structure. As it seems that every dev followed their own way how to add translations.
I would say it is a huge tech debt and we need to fix it somehow. Any thoughts? Link to discourse -
 https://community.openfoodnetwork.org/t/workflow-to-structure-translations/932

Files  in doubt -
This might be ralated with #1561 - renders partial inside with icons `app/views/spree/admin/products/product_distributions.html.haml`
`app/overrides/spree/shared/_product_tabs/add_distributions.html.haml.deface`

Those seems are not used anymore. Can we delete it?
`app/overrides/spree/shared/_header/replace_logo_with_link_text.html.haml.deface`
`app/overrides/spree/shared/_footer/footer.html.haml.deface`

### DIFF
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -15,11 +15,11 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
   $scope.currentView = -> Views.currentView
 
   $scope.views = Views.setViews
-    inventory:    { name: "Inventory Products", visible: true }
-    hidden:       { name: "Hidden Products",    visible: false }
-    new:          { name: "New Products",       visible: false }
+    inventory:    { name: t('js.variant_overrides.inventory_products'), visible: true }
+    hidden:       { name: t('js.variant_overrides.hidden_products'),    visible: false }
+    new:          { name: t('js.variant_overrides.new_products'),       visible: false }
 
-  $scope.bulkActions = [ name: "Reset Stock Levels To Defaults", callback: 'resetStock' ]
+  $scope.bulkActions = [ name: t('js.variant_overrides.reset_stock_levels'), callback: 'resetStock' ]
 
   $scope.columns = Columns.columns
 
@@ -52,22 +52,22 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
 
   $scope.displayDirty = ->
     if DirtyVariantOverrides.count() > 0
-      num = if DirtyVariantOverrides.count() == 1 then "one override" else "#{DirtyVariantOverrides.count()} overrides"
-      StatusMessage.display 'notice', "Changes to #{num} remain unsaved."
+      num = if DirtyVariantOverrides.count() == 1 then t('js.variant_overrides.one_override')  else "#{DirtyVariantOverrides.count()} " + t('js.variant_overrides.overrides')
+      StatusMessage.display 'notice', t('js.variant_overrides.changes_to') + ' ' + num + ' ' + t('js.variant_overrides.remain_unsaved')
     else
       StatusMessage.clear()
 
   $scope.update = ->
     if DirtyVariantOverrides.count() == 0
-      StatusMessage.display 'alert', 'No changes to save.'
+      StatusMessage.display 'alert', t('js.variant_overrides.no_changes_to_save')
     else
-      StatusMessage.display 'progress', 'Saving...'
+      StatusMessage.display 'progress', t('js.saving')
       DirtyVariantOverrides.save()
       .success (updatedVos) ->
         DirtyVariantOverrides.clear()
         VariantOverrides.updateIds updatedVos
         $scope.variant_overrides_form.$setPristine()
-        StatusMessage.display 'success', 'Changes saved.'
+        StatusMessage.display 'success', t('js.changes_saved')
         VariantOverrides.updateData updatedVos # Refresh page data
       .error (data, status) ->
         StatusMessage.display 'failure', $scope.updateError(data, status)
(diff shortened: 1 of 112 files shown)


spec/lib/open_food_network/products_and_inventory_report_spec.rb
```diff
@@ -21,7 +21,7 @@ module OpenFoodNetwork
           "Taxons",
           "Variant Value",
           "Price",
-          "Group Buy Unit Quantity",
+          "Group By Unit Quantity",
```
oeoeaio: This is actually supposed to be 'Buy' not 'By'

app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
```diff
@@ -52,55 +52,55 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
...
-      "I couldn't get authorisation to save those changes, so they remain unsaved."
+      t('js.variant_overrides.no_authorisation')
 
     else if status == 400 && data.errors?
       errors = []
       for field, field_errors of data.errors
         errors = errors.concat field_errors
       errors = errors.join ', '
-      "I had some trouble saving: #{errors}"
+      t('js.variant_overrides.some_trouble') + ": #{errors}"
```
oeoeaio: Just wondering if we should be using interpolation here?

pacodelaluna:
@jeronimo Great job translating the JS part ! About unused views/code, I am totally for removing it to avoid confusion for newcomers (and we can find it back into Git history if needed), but I guess it should be done in an other PR.

config/locales/en.yml
```diff
@@ -1400,13 +1401,17 @@ Please follow the instructions there to make your enterprise visible on the Open
   admin_entreprise_groups_web_twitter: "eg. @the_prof"
   admin_entreprise_groups_web_website_placeholder: "eg. www.truffles.com"
   admin_order_cycles: "Admin Order Cycles"
+  whats_this: What's this?
```
mkllnk: This key already exists in `admin.whats_this`.

spec/features/admin/reports_spec.rb
```diff
@@ -536,12 +538,12 @@ def xero_invoice_account_invoice_row(adjustment, opts={})
     def xero_invoice_row(sku, description, amount, quantity, tax_type, opts={})
       opts.reverse_merge!({customer_name: 'Customer Name', address1: 'customer l1', city: 'customer city', state: 'Victoria', zipcode: '1234', country: country.name, invoice_number: order1.number, order_number: order1.number, invoice_date: '2015-04-26', due_date: '2015-05-26', account_code: 'food sales'})
 
-      [opts[:customer_name], 'customer@email.com', opts[:address1], '', '', '', opts[:city], opts[:state], opts[:zipcode], opts[:country], opts[:invoice_number], opts[:order_number], opts[:invoice_date], opts[:due_date],
+      [opts[:customer_name], 'customer@email.com', opts[:address1], '', opts[:city], opts[:state], opts[:zipcode], opts[:country], opts[:invoice_number], opts[:order_number], opts[:invoice_date], opts[:due_date],
```
mkllnk: I don't think we should alter the spec. Some reports like the Zero report are parsed by other services and they may be very particular about some details.

spec/features/admin/reports_spec.rb
```diff
@@ -536,12 +538,12 @@ def xero_invoice_account_invoice_row(adjustment, opts={})
...
 
        sku,
        description,
        quantity,
-       amount.to_s, '', opts[:account_code], tax_type, '', '', '', '', Spree::Config.currency, '', 'N']
+       amount.to_s, '', opts[:account_code], tax_type, '', '', '', '', Spree::Config.currency, '', 'No']
```
mkllnk: I don't know if Zero can parse `No` or if it parses only `N` correctly.

app/models/content_configuration.rb
```diff
@@ -17,19 +17,19 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
   has_attached_file :home_hero, default_url: "/assets/home/home.jpg"
 
   # Producer sign-up page
-  preference :producer_signup_pricing_table_html, :text, default: "(TODO: Pricing table)"
-  preference :producer_signup_case_studies_html, :text, default: "(TODO: Case studies)"
-  preference :producer_signup_detail_html, :text, default: "(TODO: Detail)"
+  preference :producer_signup_pricing_table_html, :text, default: I18n.t(:content_configuration_pricing_table)
```
mkllnk: Should we introduce `models.content_configuration.pricing_table`? Or just `content_configuration.pricing_table`?

spec/features/admin/reports_spec.rb
jeronimo: Didn't know about Zero report. Definitely will restore it back

app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
jeronimo: I had bad experience with interpolation with colons and it doesn't make sense in other languages. But I will change it anyway

jeronimo:
I have not responded to comments with doubts in it as I don't have opinions about it as well. And have made changes for interpolations though I am skeptic about it as there are too many examples of it's failure. Anyway, we will see how it will go. And please be gentle with this PR, it was hard and not pleasant work, I still don't want to see translations for a while :)

app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
mkllnk: I thought this would be best like this: `t('js.variant_overrides.some_trouble', errors: errors)`

oeoeaio: @jeronimo, isn't that the point of interpolation? You can choose whether use the colon or not in the translation. If we keep it the way it is, the colon is hard-wired, and so it will be impossible to come up with a translation that makes sense in other languages. See @mkllnk's comment above. This would allow you to include a colon in the english translation, but structure it completely differently in another language.

mkllnk: I sometimes get confused when we talk about interpolations in this context. We have to distinguish string interpolation from the variable interpolation in i18n.

- Ruby string interpolation: `"Hello #{name}!"`
- I18n variable interpolation: `t('hello', name: name)` with locale `hello: "Hello %{name}!"`

You know both of these, but I don't always know which one you are talking about. Was there just some misunderstanding?

mkllnk:
Thanks @jeronimo. That is a lot better. Just a little comment on the side. I think you squashed your changes into the commits they belong to. That is exactly how we want it in the end. But for reviewing, it would have been easier for me to see the changes as individual commits first. After roughly reviewing all the affected files, I think this is ready for testing now.

Maybe for another pull request, since Julius is a bit tired of this now, I came across two files that still need translations:

- https://github.com/openfoodfoundation/openfoodnetwork/blob/f84acf44f384227794ccbe78741f4075712805b2/app/models/content_configuration.rb#L48
- https://github.com/openfoodfoundation/openfoodnetwork/blob/f84acf44f384227794ccbe78741f4075712805b2/app/views/producer_mailer/order_cycle_report.text.haml

mkllnk:
I think Sally can test this on Wednesday 12 July. Can you fix the Code Climate issues before then?

jeronimo:
Hi @mkllnk, Code Climate is "green" now.

And it is weird that I have missed translation in those files...  👎

sstead:
@mkllnk I assumed this PR was still in Staging2. The errors I found are shown in the 4 screenshots below...

image

image

image

image

app/views/admin/enterprise_roles/_enterprise_role.html.haml
```diff
@@ -1,5 +1,5 @@
 %td {{ enterprise_role.user_email }}
-%td= t('admin.enterprise_roles.manages')
+%td= t('.manages')
```
sauloperez: first time I see this. Does Rails infer the namespace from the file path? :astonished:

sstead:
@mkllnk sorry found another one, the initial issue are fixed though. image

app/views/admin/enterprise_roles/_enterprise_role.html.haml
oeoeaio: @sauloperez yep! It's called 'Lazy Lookup'
http://guides.rubyonrails.org/i18n.html#looking-up-translations

sstead:
All fixed :)

PR #1574, opened 2017-05-19
Banner notifying user of editable orders updates when OC is changed

oeoeaio:
Banner showing changeable orders updates when order cycle is changed.


sstead:
Looks good @oeoeaio . Changes between OCs correctly, and only shows if Shop Preferences allow changes to orders. 
One thing, if the customer has past orders, but changes to orders is NOT enabled in the shop, when the customer loads the shop, the blue banner blinks on screen for a fraction of a second as the page loads. I can't imagine this matters, bust just thought I'd mention it.

image

Issue #1571, opened 2017-05-12
Development environment - undefined method `tag_list'


jeronimo:
Sometimes I get error `undefined method `tag_list' for #<Spree::Variant:0x007fc2c65bedf8>`
<img width="732" alt="screen shot 2017-05-12 at 15 25 27" src="https://cloud.githubusercontent.com/assets/88912/26002353/47c08e4a-3727-11e7-8ad1-10b5352d1268.png">

Adding `acts_as_taggable` into Spree::Variant decorator seemed to help. Will make pull request later if no one is against it.

daniellemoorhead:
@enricostano? @Em-AK? @RohanM?

enricostano:
I've not been able to reproduce it.

myriamboure:
I propose to close this one as Enrico was not able to reproduce and we just closed the associated PR for the same reason. Will reopen if appear again.

PR #1576, opened 2017-05-19
Add acts_as_taggable to Variant decorator

jeronimo:
PR to #1571

### DIFF
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -9,6 +9,7 @@ Spree::Variant.class_eval do
   remove_method :options_text if instance_methods(false).include? :options_text
   include OpenFoodNetwork::VariantAndLineItemNaming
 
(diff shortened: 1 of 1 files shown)


oeoeaio:
My only problem with this is that it is fixing the symptom, not the cause. We don't actually want or need to tag variants...

lin-d-hop:
@oeoeaio @jeronimo should we close this? 
Agree not worth just masking a symptom and currently its just not a priority to dig deeper (sorry @jeronimo, you could volunteer your time if you really wanted to suss out the root cause).

jeronimo:
Sure, I am volunteering to jump into it the next time it will appear for me. Can we keeps this PR as a reminder for it?

lin-d-hop:
CLosing as we haven't heard from @jeronimo recently. If you reappear mate then please do reopen :-D

PR #1579, opened 2017-05-21
CodeClimate: Rubocop

oeoeaio:
### APPROACH
This PR is the first step towards having an agreed-upon set of standards to which we write code for this project. The idea is to run Rubocop as an engine via CodeClimate, which will run automated tests to help all new PRs to comply with our standards before progressing to code review. This should hopefully give us greater transparency and alleviate some of the burden on reviewers.

I used [relaxed.ruby.style](http://relaxed.ruby.style/) as a starting point for the conventions for the project. I am not particularly attached to any of these rules, but speaking from experience, I think disabling the `Metric` cops is probably a good idea in the short term.

From that point I used Rubocop's `--auto-gen-config` option to generate `.rubocop_todo.yml`, which represents a todo list in the form of specific exceptions required to allow Rubocop to run without errors. Where less that 15 files are found to contain offences, those individual files are listed as exceptions to the particular cop. Where the number of offending files is greater than 15, the generator just completely disables the cop using `Enabled: false`. The workflow from this point basically looks like this:

1. Choose a cop that is mentioned in `.rubocop_todo.yml`
2. Read the docs and understand the options for configuration of the relevant cop
3. Remove the cop from `.rubocop_todo.yml`
4. Either:
  a) Fix all instances of the offence in the codebase
  b) Decide to implement non-default settings for the cop, add these to `.rubocop.yml`
5. Commit code and move on to another offence class / cop

The idea would be that we gradually remove all exceptions from `.rubocop_todo.yml` altogether, instead documenting all of our agreed-upon conventions in `.rubocop.yml`

### NEXT UP
1. I think the highest priority is to remove all instances of any blanket exceptions (`Enabled: false`) that still exist in `.rubocop_todo.yml`. I think there are still about 30 of these remaining, but the vast majority can be resolved automatically by Rubocop (using `--auto-correct`), if that is what we decide to do.
2. I think making decisions about all of the `Rails` cops would be a good next step.
3. I guess move on to the lowest hanging fruit (fewest offences) or highest priority for standard setting (where there is strong agreement within the community).

Comments / suggestions / questions welcome


daniellemoorhead:
You're awesome @oeoeaio

.rubocop.yml
```diff
@@ -1,29 +1,169 @@
...
+# oeoeaio after scrutinising offenses the codebase
+
+# Don't think this is a big issue, mostly picking up RPSEC scope definitions
+# with lamdas and RSpec '.to change{}' blocks
+Lint/AmbiguousBlockAssociation:
+  Enabled: false
+
+# No alternative to this one until we upgrade to Rails 4 and can use #find_by
+Rails/DynamicFindBy:
+  Enabled: false
```
sauloperez: but I think this one would prevent us using any more dynamic finders, otherwise we would make the upgrade to rails 4 even harder.

.rubocop.yml
```diff
@@ -1,29 +1,169 @@
...
+
+# This should be the programmer's discretion, perhaps we should review all of
+# the uses of it an make specific exceptions though.
```
sauloperez: agree. Let's use `# rubocop:disable` for these cases (if any)

.rubocop.yml
```diff
@@ -1,29 +1,169 @@
...
+Rails/SkipsModelValidations:
+  Enabled: false
+
+# This should be enfored, but doing so would require changing the entire
+# file, so should probably do when outstanding PRS are at a minimum
+Style/ClassAndModuleChildren:
```
sauloperez: can we have this one in the `rubocop_todo.yml`? This would mean new classes have to conform to it. I guess that is what we want.

.rubocop.yml
```diff
@@ -1,29 +1,169 @@
...
 
-MethodLength:
+Metrics/MethodLength:
+  Enabled: false
+
+Metrics/ParameterLists:
   Enabled: false
 
-StringLiterals:
+Metrics/PerceivedComplexity:
```
sauloperez: have we considered enabling any of these complexity metrics? Even with a high threshold I believe it can be very benefitial to have a heads-up when the code is getting to complex.

It makes you reconsider things. There's probably a more readable and changeable way to implement what you're doing.

mkllnk:
Awesome! It's good to merge these changes quickly before we get conflicts.

Which version of rubocop do I need? I just ran it locally:
```sh
rubocop -v
# 0.38.0
rubocop
#Warning: unrecognized cop Bundler/OrderedGems found in /home/maikel/github/openfoodnetwork/.rubocop_todo.yml
#Warning: unrecognized cop Lint/InheritException found in /home/maikel/github/openfoodnetwork/.rubocop_todo.yml
#Warning: unrecognized cop Metrics/BlockLength found in /home/maikel/github/openfoodnetwork/.rubocop_todo.yml
# [...]
# 585 files inspected, 118 offenses detected
```

If there are no offences running it locally, everybody can have a pre-commit hook containing `git diff --cached --diff-filter=ACMRTUXB --name-only -- '*.rb' | xargs rubocop`.

sauloperez:
@mkllnk we should all be running the version specified in the Gemfile with `bundle exec rubocop ...`.

That pre-commit would be awesome! Perhaps we can take a look at https://github.com/m4i/rubocop-git. I tried it once without any luck though.

mkllnk:
My current version is not that pretty. I appended that to my pre-commit hook
```sh
# Run Rubocop
NUMBER_BEFORE=`git diff --cached --diff-filter=CMRTUXB --name-only -- '*.rb' | sed 's/^\(.*\)$/HEAD:\1/' | xargs git show | rubocop -s head -f o | tail -n 2 | head -n 1 | cut -d ' ' -f 1`
NUMBER_NOW=`git diff --cached --diff-filter=ACMRTUXB --name-only -- '*.rb' | xargs cat | rubocop -s now -f o | tail -n 2 | head -n 1 | cut -d ' ' -f 1`

if [ "$NUMBER_BEFORE" -lt "$NUMBER_NOW" ]; then
  echo "You have $NUMBER_NOW offenses now. There were $NUMBER_BEFORE before."
  echo "Display all offenses of modified files by running:"
  echo "  git diff --diff-filter=ACMRTUXB --name-only | xargs rubocop"
  exit 1
fi
```

sauloperez:
@oeoeaio can you please add rubocop in the development group of the Gemfile. This way it'll get installed with all the other gems with bundler.

sauloperez:
https://github.com/bbatsov/rubocop/blob/master/manual/integration_with_other_tools.md :point_left: is what I mentioned in the meeting. It's crucial to get information from Rubocop as early as possible.

With that your own editor will warn you about style violations almost as you type ensuring you open the PR adheres to the style guide (rubocop.yml) without having to wait for CodeClimate.

.rubocop.yml
oeoeaio: I'm not sure that this is actually how Rubocop itself works is it? I think that the exceptions listed in `rubocop_todo.yml` are treated in the same way as those in `rubocop.yml`, having the two files just provides a way of separating out existing errors that have not been looked at yet (`rubocop_todo.yml`) from agreed upon configurations (`rubocop.yml`). So if we are using Rubocop linters, and we remove this line, devs will start getting errors about things that they didn't cause. Not sure if that is a problem or not...

From my understanding, CodeClimate adds an additional layer of magic, in that it understands git and only raises Rubocop errors if they are not present on the base branch (ie. `master`). That is great, but as above, doesn't help us with linters.

.rubocop.yml
oeoeaio: What do we use instead? Rails 3.2 has no `find_by` method and I can't seem to find a gem that allows it to be back-ported...

Do you suggest that we use `where(...).first`?

.rubocop.yml
oeoeaio: There are currently 158 exceptions to this cop. 141 of those are in `/spec`. Mostly `update_attribute` but a few uses of `update_all` and `update_column`. How do you feel about just disabling this cop for `/spec` for now, and changing or adding specific allowances for the other exceptions?

.rubocop.yml
mkllnk: Good point. But I would wait for all discussions what to *enable* for after the first pull request is merged. Let's get a minimal version in now and then work on improving it.

mkllnk:
> we should all be running the version specified in the Gemfile with `bundle exec rubocop` ....

That's probably the right way

.rubocop.yml
mkllnk: According to https://github.com/bbatsov/rails-style-guide#find_by both methods are bad and only the Rails 4 `find_by` is good. I would either stick to the old `find_by_attribute` (disable check) or monkey patch ActiveRecord to define `find_by(col, val)` as `where(col => val).first`.

sauloperez: But I think as far as we add the current offenses to the rubocop_todo.yml we won't make things worse.

mkllnk: But what about new code? Do you think `where(name: ...).first` is better than `find_by_name(...)`?

oeoeaio: Also @sauloperez, there are hundreds of uses of dynamic matchers in the codebase at the moment, so we probably wouldn't add them to `rubocop_todo.yml`, cause that would take forever. As I explained below, I don't think Rubocop is quite sophisticated enough to mark some offences as current/todo, and then raise issues on any future offences. CodeClimate does this, but I don't think Rubocop can...

.rubocop.yml
sauloperez: > So if we are using Rubocop linters, and we remove this line, devs will start getting errors about things that they didn't cause.

That's why we need to add them in the rubocop_todo.yml. When running rubocop this will skip the currenr violations but it will raise the new ones not present in the todo. Code Climate will do exactly the same as it runs de rubocop as well.

The point I believe is to agree on a style guide (rubocop.yml) and "hide" all existing violations (rubocop_todo.yml) so that we can apply the guide on any new code without having to clean them up upfront.

oeoeaio: Yeah cool I understand what you mean now. The main problem is the number of offences, we have to add exceptions for 87 (!) files for this one...

.rubocop.yml
sauloperez: See my comment on `ClassAndModuleChildren`. 

I just did a test on my machine. I have a rubocop_todo.yml with the following content:

```ruby
Offense count: 378
RSpec/AnyInstance:
  Exclude:
    - some_random_controller_spec.rb
```

I run the following on my terminal:

```shell
$ rubocop some_random_controller_spec.rb
```

and I get no errors. I remove `some_random_controller_spec.rb` from the `rubocop_todo.yml` and execute rubocop again:

```shell
$ rubocop some_random_controller_spec.rb

some_random_controller_spec.rb:178:15: C: Avoid stubbing using expect_any_instance_of.                        
              expect_any_instance_of(Onboarding::Process)             
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^             
some_random_controller_spec.rb:212:15: C: Avoid stubbing using allow_any_instance_of.                         
              allow_any_instance_of(Onboarding::Process).to receive(:user_errors?).and_return(true)                                         
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^              
some_random_controller_spec.rb:228:15: C: Avoid stubbing using expect_any_instance_of.                        
              expect_any_instance_of(AppLink).to receive(:destroy)    
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                         
some_random_controller_spec.rb:236:15: C: Avoid stubbing using allow_any_instance_of.                         
              allow_any_instance_of(Onboarding::Process).to receive(:app_link_errors?) { true }                                             
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^              
some_random_controller_spec.rb:253:15: C: Avoid stubbing using expect_any_instance_of.                        
              expect_any_instance_of(AppLink).to receive(:destroy)    
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                         

1 file inspected, 7 offenses detected            
```

.rubocop.yml
sauloperez: ok, seems reasonable.

.rubocop.yml
sauloperez: I think that's fine. In the repo I used on my example above there's a cop (the name rubocop gives to linters) with 378 files listed in it :trollface:

.rubocop.yml
oeoeaio: Yeah, I understand all that that, I guess my feeling was that when such a large number of files need to be excluded, the value in retaining the cop on the remaining files is substantially diminished, because it is almost random chance whether future violations will be caught or not. Happy to go with your approach though.

My one remaining issue with that approach for this particular cop is: what do we suggest people do instead? The Ruby style guide [suggests that](https://github.com/bbatsov/rails-style-guide#find_by) a `where(...).first` is just as undesirable as a dynamic `find_by_...`.

Further, I think that forcing developers to use `where(...).first` instead of dynamic matchers is a retrograde step, because switching to `find_by(...)` will probably be easier from dynamic matchers than it will be from `where(...).first`...

.rubocop.yml
oeoeaio: 378 offences, not 378 files. Each file can have multiple offences. I will just remove this cop from `rubocop.yml` and regenerate `rubocop_todo.yml`, and remove the cap on the number of files it will exclude before just disabling the cop altogether. Does that sound good?

.rubocop.yml
enricostano: I would merge this with @sauloperez suggestions and open a PR for each cop we want to discuss about. So we can avoid endless discussions for lots of cops in one PR. What do you think?

sauloperez: Agree. Just not to forget it, this what I found out:

> Note that find(primary_key), find_by..., and find_by...! are not deprecated.

https://github.com/rails/activerecord-deprecated_finders/blob/master/README.md

sauloperez:
Is there something I can help with to get this merged? Let me know if I can be of help @oeoeaio !

oeoeaio:
@sauloperez I think it's good to go. 😊  I made a couple of changes on the weekend as per your suggestions, so if you are happy with it (as a starting point) I will aim to merge it in tomorrow.

sauloperez:
Let's merge it. It's a big step forward to start using Rubocop!

sauloperez:
Yay!! :tada: !

enricostano:
👏 👏 👏

PR #1583, opened 2017-05-26
Uk/1031 insufficient stock in cart

oeoeaio:
Supersedes #1552. I have rebased and cleaned up the commit history.

### DIFF
--- a/spec/support/request/shop_workflow.rb
+++ b/spec/support/request/shop_workflow.rb
@@ -13,7 +13,7 @@ module ShopWorkflow
   end
 
   def set_order(order)
-    ApplicationController.any_instance.stub(:session).and_return({order_id: order.id, access_token: order.token})
+    inject_session(order_id: order.id, access_token: order.token)
   end
 
   def add_product_to_cart(order, product, quantity: 1)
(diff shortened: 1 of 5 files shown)


sstead:
This is problem is still outstanding...
"If I had an item in my cart, then I logged out, then the item was removed from the OC (either just that item or the supplier), then I log in and go to the shopfront- my cart hangs and I'm unable to proceed through checkout. If I instead login and click straight on the edit my cart dropdown link, I can proceed through checkout and purchase something that's no longer in the OC."

oeoeaio:
Hey @jeronimo, from Sally's testing it looks like this is not yet resolved.

Need to handle the case where:
1) a variant is in the user's cart
2) the user logs out
3) the variant is removed completely from the order cycle (not just stock reduced)
4) the user logs back in

spec/support/request/shop_workflow.rb
```diff
@@ -49,4 +49,12 @@ def ensure_supplier_exchange(exchange, supplier)
...
     end
   end
+
+  def inject_session(hash)
+    Warden.on_next_request do |proxy|
+      hash.each do |key, value|
+        proxy.raw_session[key] = value
+      end
+    end
+  end
```
enricostano: Can we add some doc here?

jeronimo:
@lin-d-hop, may I get what notification text message should be when variant gets removed in admin?

daniellemoorhead:
>@lin-d-hop, may I get what notification text message should be when variant gets removed in admin? Or is it ok without any notification?

@lin-d-hop or perhaps @sstead thoughts on what the copy should be for the error message?

oeoeaio:
Note: this appears to be superseded by #1698

daniellemoorhead:
Hey @jeronimo can this PR be closed as it seems to be superseded by #1698?

jeronimo:
@daniellemoorhead, I think so.

PR #1608, opened 2017-06-07
Do not blindly clear the whole Rails cache

enricostano:
This is one of the causes of "random" failing specs in Travis CI that are hard to reproduce locally.

Basically using `Rails.cache.clear` in specs (and any other place IMHO) is harmful since it alters the state of the Rails application beyond the scope of the spec example.


enricostano:
Ping @RohanM @oeoeaio @sauloperez

jeronimo:
👍

daniellemoorhead:
@oeoeaio and @mkllnk do we need to test this one or can we just push it live? Seems really really valuable to get it out ASAP, so if it needs testing then let's get Sally on it now!

PR #1609, opened 2017-06-08
Country selection in enterprise registration

mkllnk:
Addressing #1330 and competing with #1334.

Building on work by @Matt-Yorkley with lots of copy and paste. I created new commits for a better structure. Instead of a new directive proposed by Matt, I used `ng-init` to set the default country instead. I also omitted the new file `app/assets/javascripts/templates/registration/details.html.haml`. Was that used? What do you think about my new version?

@lin-d-hop I volunteered my work on this alternative. I couldn't justify spending so much time on proposing my own solution when you had something working already. :-)

### DIFF
--- a/spec/features/consumer/registration_spec.rb
+++ b/spec/features/consumer/registration_spec.rb
@@ -48,7 +48,7 @@ feature "Registration", js: true do
       fill_in 'enterprise_address', with: '123 Abc Street'
       fill_in 'enterprise_city', with: 'Northcote'
       fill_in 'enterprise_zipcode', with: '3070'
-      expect(page).to have_select('enterprise_country', options: %w(Albania Australia Chad), selected: 'Albania')
+      expect(page).to have_select('enterprise_country', options: %w(Albania Australia), selected: 'Albania')
       select 'Australia', from: 'enterprise_country'
       select 'VIC', from: 'enterprise_state'
       perform_and_ensure(:click_button, "Continue", lambda { page.has_content? 'Who is responsible for managing My Awesome Enterprise?' })
--- a/spec/features/consumer/registration_spec.rb
+++ b/spec/features/consumer/registration_spec.rb
@@ -6,7 +6,20 @@ feature "Registration", js: true do
 
   describe "Registering a Profile" do
     let(:user) { create(:user, password: "password", password_confirmation: "password") }
-    before { Spree::Config.enterprises_require_tos = false }
+
+    before do
+      Spree::Config.enterprises_require_tos = false
+
+      albania = Spree::Country.create!({ name: "Albania", iso3: "ALB", iso: "AL", iso_name: "ALBANIA", numcode: "8" }, without_protection: true)
+      Spree::State.create!({ name: "Berat", abbr: "BRA", country: albania }, without_protection: true)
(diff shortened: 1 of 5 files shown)


spec/features/consumer/registration_spec.rb
```diff
@@ -6,7 +6,20 @@
...
+      albania = Spree::Country.create!({ name: "Albania", iso3: "ALB", iso: "AL", iso_name: "ALBANIA", numcode: "8" }, without_protection: true)
+      Spree::State.create!({ name: "Berat", abbr: "BRA", country: albania }, without_protection: true)
+      Spree::Country.create!({ name: "Chad", iso3: "TCD", iso: "TD", iso_name: "CHAD", numcode: "148" }, without_protection: true)
+    end
+
+    after do
+      Spree::State.where(name: 'Berat').delete_all
+      Spree::Country.where(name: 'Albania').delete_all
+      Spree::Country.where(name: 'Chad').delete_all
+    end
```
sauloperez: Don't we use something like https://github.com/DatabaseCleaner/database_cleaner ? Do we always cleanup the data manually after the test?

mkllnk: We do. The only exception are countries and states.

sauloperez: Is there some reason not to clean the data through database cleaner? I find a bit more readable that way. You assume data gets cleaned eventually and don't care about it.

enricostano: Yep, I'm curious about the reasons behind those exceptions too. Can we safely remove them?

mkllnk: It looks like that code is very old from the very early days: 1288760839f1f1c9f10dec8bd3d8fc92c3b68107

I remember that there was some difficulty with letting the cleaner delete these, because of foreign key constraints. We would also need a method to provide countries and states for the tests that need them.

So yes, if you would like to create a pull request to change it, great. This exception is really not nice. It is very annoying when you did a database migration and then reset the test database. At the moment, you need to quit Zeus and start it again to run the country seeding. Otherwise tests just fail.

enricostano: Cool, this sounds like a nice PR @sauloperez could work on 😉

sauloperez: > It is very annoying when you did a database migration and then reset the test database. At the moment, you need to quit Zeus and start it again to run the country seeding. Otherwise tests just fail.

🙈  Cool @enricostano I will!

Matt-Yorkley:
Looks great. :+1:

mllocs:
Hello 👋 

I wonder if we should also fix the seed data in `master`, removing all those countries without states or adding states to every country. 

I recently forked the repo and I've been struggling to underestand why I wasn't able to select a country without states. https://github.com/openfoodfoundation/openfoodnetwork/issues/1620

mkllnk:
@mllocs 

> I wonder if we should also fix the seed data in master, removing all those countries without states or adding states to every country.

I think future versions of Spree actually import all states as well. It's probably nicer if all data is there and people can choose what to display than every instance creating their own states file. I would wait for the upgrade. If you are impatient, you could back-port that code.

PR #1610, opened 2017-06-08
Add style guide to the source locale

mkllnk:
There has been a lack of rules how to structure our locale en.yml. This
is a start to document some guidelines which can evolve over time.

Rules can be discussed in the community forum or in pull requests. Once
there is a strong agreement, we should add it to this guide.

@jeronimo @enricostano @oeoeaio What do you think about the rules? What do you think about documenting them this way?

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -13,6 +13,45 @@
 #
 #
 # See http://community.openfoodnetwork.org/t/localisation-ofn-in-your-language/397
+#
+# Changing this file
+# ==================
+#
+# You are welcome to fix typos, add missing translations and remove unused ones.
+# Here are some guidelines to make sure that this file is becoming more beautiful
+# with every change we do.
+#
+# * Change text: No problem. Fix the typo. And please enclose the text in quotes
+#   to avoid any accidents.
+#
+#   Example 1: "When you're using double quotes, they look like \"this\""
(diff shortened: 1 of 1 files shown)


enricostano:
Good initiative @mkllnk 👏 

I agree on each point you are proposing. I'm not sure about location of the documentation, maybe a wiki page on GitHub and a link to it and the discussion thread?

We can even give it a try in `en.yml` and then decide if people read it and what to do.

mkllnk:
Yes, I thought about other locations. But I wasn't sure where it should go. It needs one place that holds the document with the agreed wisdom and we need places to discuss about new proposals. That could be en.yml and pull requests. Or on Discourse. Or wiki plus Discourse. ???

pacodelaluna:
Well, I was thinking of keeping the documentation into the Wiki, and we could add links to it inside the code. I think it is good to have a central place to deal with the project documentation, it can help newcomers to have a quick overview on how people are working on the project.
We could use Discourse to share opinions on a subject, and then Wiki to formalize them in order to keep the documentation updated.

pacodelaluna:
Actually, imho, this discussion should happen on Discourse :)

kirstenalarsen:
i agree, discussion on discourse, agreed position ‘latest knowledge’ on the wiki . .?

Issue #1541, opened 2017-04-20
Issue in "change pack" option 2 not displaying


myriamboure:
Open 2 description is the same as option 1 and a text "producer description text 2" appears. On v 1.8.9 deployed on the French staging.
![screenshot from 2017-04-20 15-03-24](https://cloud.githubusercontent.com/assets/8135376/25232188/bf69ca1e-25da-11e7-8a50-fb720392516a.png)

@pacodelaluna is it connected to something you have done with translatability or not?
@sstead do you have the same issue in Australia ? So that we know if it's connected to the translation process...

sstead:
Hey @myriamboure , yes, we are also having this issue...
![image](https://cloud.githubusercontent.com/assets/10843056/25259975/a75207b8-268c-11e7-9635-d6c4a836fa4e.png)

myriamboure:
@pacodelaluna do you think that's something Claire could work on? Is it easy enough for her to start ?

myriamboure:
@sstead @clairezed is fixing that, it was a bad key, can you confirm what text need to display here in the producer shop ?
There are two possible text in Transifex:
> sell_description_text
Sell your products directly to customers through your very own Open Food Network shopfront.
or 
> sell_description_text2
A Producer Shop is for your produce only, if you want to sell produce grown/produced off site, select 'Producer Hub'.

Shall we put both one after the other so that both display ? It sounds logical to us... thanks for confirming !

sstead:
Yes, I believe initially both these sentences were displayed, and the user guide screenshot confirms this....
![image](https://user-images.githubusercontent.com/10843056/26859284-067812d4-4b7a-11e7-9165-0a57b4d880a5.png)
Thanks :)

myriamboure:
Awesome, I think you can push your PR @clairezed :-)

myriamboure:
@clairezed I think your PR was ready on that, am I right ? Sally answered and it confirmed what we had worked out last week. Let me know if any question ;-)

enricostano:
> a text "producer description text 2" appears

This is because the i18n key `producer_description_text2` does not exists. I added more insights in the related PR.

daniellemoorhead:
@myriamboure and @clairezed the PR for this issue was merged so feel free to close this one :-)

PR #1622, opened 2017-06-16
#1541 fixing wrong translation key in producers packs

clairezed:


app/views/admin/enterprises/_change_type_form.html.haml
```diff
@@ -27,10 +27,10 @@
               %monthly-pricing-description{ joiner: "newline" }
 
           %p.description
-            = t('.producer_description_text')
+            = t('.sell_description_text')
             %br
             %br
-            = t('.producer_description_text2')
+            = t('.sell_description_text2')
```
enricostano: If `producer_description_text` and `producer_description_text2` keys are not used anymore, please remove them from locale yml files.

myriamboure: I think they are @enricostano , because it was a duplicate from the first producer pack option, we were seeing twice the same text, so my first bet is they need to remain :-) But maybe Claire will have another input ;)

enricostano: looking for those keys with `git grep` I could find anything, are those keys used by Spree core? If not we can safely remove them.

enricostano: They were introduced by commit https://github.com/clairezed/openfoodnetwork/commit/88dea0f2b8c60d846d97dd591000b504d2e171b6 , actually the key `producer_description_text2` is not even defined anywhere... I think we can safely remove the unused keys.

myriamboure:
@clairezed awesome if you can have a look at @enricostano 's comment before we can merge :-)

clairezed:
I've already replied 2 days ago, but it's written `pending`. Can anyone see my comments ?

"as far as I've seen, `producer_description_text` key exists and is in use.
But I can't find anywhere anymore the key `producer_description_text2` (nothing the local yml files) . So, I don't understand : where should I remove it ?"

myriamboure:
Now I can see your comment @clairezed but not the one from 2 days ago :-o @enricostano any clue for Claire ? Or @pacodelaluna if you understand what it is about ?

oeoeaio:
@clairezed about your 'pending' comment: it sounds like you might have "started a review", which means that all of your comments will be pending until you submit the review. If you click on 'Files Changed' you should see a green box in the top right of the page that says 'Review changes' with a number (the number of pending comments). If you submit the review we should see your comments. In future, if you just want to make a comment, you should choose 'Add single comment' instead of 'Start a review'. Hope that helps!

About the PR: I think this is actually fine. @enricostano was concerned that there were orphaned translation keys being left behind, but I think he then realised that that wasn't the case. As you say, `producer_description_text` is still used, and `producer_description_text2` never existed.

One suggestion I would make however (not your fault at all, but would be good to fix it): I think it would be easier to understand what is going on if the `sell_description_text` and `sell_description_text2` keys were renamed to `producer_shop_description_text` and `producer_shop_description_text2`. That way they follow they would follow the same convention as the other strings on the page.

oeoeaio:
Hey @clairezed, it looks like you may have changed the wrong translation key in b609128cafaf731e8560053e955a4b6944c9cfa9? The commit message says the right thing, but the actual change you made seems to have been to change `producer_description_text` to `sell_description_text`. I think what we actually want is to **keep** `producer_description_text` and to change `sell_description_text` to `producer_shop_description_text `, and `sell_description_text2` to `producer_shop_description_text2 `.

Let me know if I am missing something... 😊 

Also, a picky thing: if you could begin commit messages with a capital letter that would be awesome. 🙏

sauloperez:
Again, this tool can be of great help https://github.com/m1foley/fit-commit

> ## Validations
> Subject Period: Do not end your subject line with a period.

clairezed:
oh gosh, my bad ! Sorry, I clearly had my mind elsewhere. I double checked the keys, added a capital letter (thanks @sauloperez for the commit check tool and guideline, i've bookmarked it !). Hopefully, this time it's ok

daniellemoorhead:
@oeoeaio do we need to have @sstead test this now?

PR #1624, opened 2017-06-17
Sus issue 1604

Duende13:
Variants are displayed Alphabetically and  Ascending unit_value. #1604 
Let me know your thoughts @lin-d-hop

PR #1625, opened 2017-06-19
Issue 1604 - Variant Ordering

lin-d-hop:
Adding order by name_to_display and unit_value to the variants in products

Cleaned up the commit history of @Duende13 PR #1624


enricostano:
Should we add some test covering this?

lin-d-hop:
@enricostano Good idea. Would you be able to give @Duende13 some pointers as I believe this will be a new task for her?

enricostano:
@lin-d-hop @Duende13 

1. For the change in the Variant serializer we can take the opportunity to add the spec, like `spec/serializers/order_serializer_spec.rb`
2. For the products order we can add an example checking for the orders in `spec/features/consumer/shopping/products_spec.rb`

PR #1627, opened 2017-06-19
Remove unused #within_zone? in ShippingMethod decorator

enricostano:
This code is not used anymore 🔥 

### Test
 - Checkout without a shipping address

### DIFF
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -48,14 +48,6 @@ Spree::ShippingMethod.class_eval do
   end
   alias_method_chain :available_to_order?, :distributor_check
 
-  def within_zone?(order)
-    if order.ship_address
-      zone && zone.include?(order.ship_address)
-    else
-      true # Shipping methods are available before we've selected an address
-    end
-  end
-
   def has_distributor?(distributor)
     self.distributors.include?(distributor)
   end
--- a/spec/models/spree/shipping_method_spec.rb
+++ b/spec/models/spree/shipping_method_spec.rb
@@ -55,12 +55,6 @@ module Spree
                   distributor: create(:distributor_enterprise), currency: currency)
         sm.should_not be_available_to_order o
       end
-
-      it "is available to orders with no shipping address" do
-        o = create(:order, ship_address: nil,
-                  distributor: sm.distributors.first, currency: currency)
-        sm.should be_available_to_order o
-      end
     end
 
     describe "finding services offered by all distributors" do


myriamboure:
I'm just wondering @enricostano is there any connexion with the fact of defining a delivery zone ? Cause we would need that for some users who offer home delivery but want to define a delivery zone where they can ship. Just wondering if any connexion, ignore if not :-)

enricostano:
@myriamboure right now is not used and recent Spree provides multiple zones per shipping method out of the box.

app/models/spree/shipping_method_decorator.rb
```diff
@@ -49,14 +49,6 @@ def available_to_order_with_distributor_check?(order, display_on=nil)
   end
   alias_method_chain :available_to_order?, :distributor_check
 
-  def within_zone?(order)
-    if order.ship_address
-      zone && zone.include?(order.ship_address)
-    else
-      true # Shipping methods are available before we've selected an address
-    end
-  end
```
sauloperez: I found out that this method is used internally by spree in `core/app/models/spree/shipping_method.rb`

```ruby
def available_to_order?(order, display_on= nil)
  available?(order, display_on) &&                                    
  within_zone?(order) &&           
  category_match?(order) &&        
  currency_match?(order)                                                                                                                    
end    
```

Although by that comment in line 56, I think we overrode that method so that we actually show the shipping method before the user specifying a shipping adress as by doing so `available_to_order?` will return true.

Turns out that we only introduced that `else` branch as the spree's implementation is exactly the same:

```ruby
def within_zone?(order)    
  zone && zone.include?(order.ship_address)                   
end
```

To me it seems that we implemented the behaviour in the wrong place, changing the meaning of `within_zone?`. IMO it should be done in the view layer since this has nothing to do about business logic but UI. Having less overrides is always better.

enricostano: Yet another case of missing documentation that drives to loosing time... Thanks @sauloperez !

sauloperez: Agree. We should take this as part of the spree upgrade. We might have more like this one.

oeoeaio:
@enricostano is this PR ok to be closed?

sauloperez:
@oeoeaio I think this would break the app as spree still uses this method. It is something we need to refactor for the upgrade to work. Should we add a WIP label?

oeoeaio:
@sauloperez yep, sounds good.

app/models/spree/shipping_method_decorator.rb
enricostano: Talked offline with @sauloperez and @oeoeaio , we agreed that we don't need `#within_zone` customization anymore.

daniellemoorhead:
If it's blocked should it go in the blocked column @sauloperez @oeoeaio ?

luisramos0:
I am closing this one and creating a new one (same same, delete shipping_method.within_zone) against 2-0-stable. See #2652

PR #1639, opened 2017-06-23
Fix #1611 Rephrase error message for people logging in with an existing email 

### LINKED ISSUE #1611: Rephrase error message for people logging in with an existing user email
See suggested rephrase below:
image

diogomatsubara:
Solving #1611 Rephrase error message for people logging in with an existing user email

Overrides default error message for sign up validation to show a more friendly error when the given email is already registered.

Before:
screenshot from 2017-06-28 14 05 30

After:
screenshot from 2017-06-28 14 10 39

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -60,6 +60,12 @@ en:
       spree/order:
         payment_state: Payment State
         shipment_state: Shipment State
+    errors:
+      models:
+        spree/user:
+          attributes:
+            email:
+              taken: "There's already an account for this email. Please login or reset your password."
   devise:
     failure:
       invalid: |
--- a/spec/features/consumer/authentication_spec.rb
+++ b/spec/features/consumer/authentication_spec.rb
@@ -60,6 +60,13 @@ feature "Authentication", js: true, retry: 3 do
             page.should have_content "too short"
           end
 
+          scenario "Failing to sign up because email is already registered" do
+            fill_in "Email", with: user.email
+            fill_in "Choose a password", with: "foobarino"
+            click_signup_button
+            expect(page).to have_content "There's already an account for this email."
+          end
+
           scenario "Signing up successfully" do
             fill_in "Email", with: "test@foo.com"
             fill_in "Choose a password", with: "test12345"


config/locales/en.yml
```diff
@@ -21,6 +21,9 @@ en:
       spree/order:
         payment_state: Payment State
         shipment_state: Shipment State
+    errors:
+      messages:
+        taken: "There's already an account for this email. Please login or reset your password."
```
sauloperez: looks like this would override the generic message refering to the email. I believe we're missing the model and field this should be applied to. A quick google search returns me https://stackoverflow.com/a/18578028

diogomatsubara: @sauloperez Thanks for the review! I added the model and field to the error message.

daniellemoorhead:
@mkllnk this should be a quick one to review and send to @sstead for testing. Sally, would be great to get this in by end end of the week if we can so that's it's in the June release.

spec/features/consumer/authentication_spec.rb
```diff
@@ -60,6 +60,13 @@
             page.should have_content "too short"
           end
 
+          scenario "Failing to sign up because email is already registered" do
+            fill_in "Email", with: user.email
+            fill_in "Choose a password", with: "foobarino"
+            click_signup_button
+            page.should have_content "There's already an account for this email."
```
enricostano: We should start using the "new" Rspec syntax:
`expect(page).to have_content "There's already an account for this email."`

sauloperez: Yes, I wonder what's the best way to do it. Probably by enabling `rubocop-rspec` so that new violations get a warning. I'd say on an another PR other than @oeoeaio 's one.

sstead:
@mkllnk if you put this into staging for me I can test it tonight or first thing tomorrow. Thanks.

oeoeaio:
Just wanted to document for posterity that the vagueness of the error message in this instance was deliberate. Someone at some point thought that providing an error message that could be used to confirm that a particular email address was registered with the system was a privacy/security concern. I think this isn't a significant issue however, and the benefit in terms of UX is worth it.

mkllnk:
I agree with Rob that obscurity is not a benefit here. If someone wanted to run a script to see which addresses are registered, they can anyway. The only secure way would be to always say: "We sent an email, please check." And then the email can say: "You are already registered. Login or reset password."

PR #1640, opened 2017-06-23
Remove checkout flow states logic from Order

sauloperez:
**DISCLAIMER: this PR is just to open a discussion about the checkout flow and test whether it is actually needed for the Spree upgrade  (tests will tell us, hopefully :trollface: )**

These are our findings so far:

1 - When the shipping method does not require ship address (a pickup) we
skip the shipment creation in the payment :if callback. However, a
shipment does get created by spree in the database. Check
https://github.com/spree/spree/blob/1-3-stable/core/app/models/spree/order/checkout.rb#L82

2 - In the same payment :if callback we ensure the ship_address is valid
in order to create the shipment. This doesn't seem to bring any benefit
since ActiveRecord will run the validations before saving the models
anyway.

### DIFF
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -23,6 +23,13 @@ Spree::Order.class_eval do
 
   before_save :update_shipping_fees!, if: :complete?
   before_save :update_payment_fees!, if: :complete?
+  checkout_flow do
+    go_to_state :address
+    go_to_state :delivery
+    go_to_state :confirm, :if => lambda { |order| order.confirmation_required? }
+    go_to_state :complete
+    remove_transition :from => :delivery, :to => :confirm
+  end
 
   # -- Scopes
   scope :managed_by, lambda { |user|
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -24,24 +24,6 @@ Spree::Order.class_eval do
   before_save :update_shipping_fees!, if: :complete?
   before_save :update_payment_fees!, if: :complete?
 
-  checkout_flow do
-    go_to_state :address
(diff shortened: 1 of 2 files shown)


app/models/spree/order_decorator.rb
```diff
@@ -24,24 +24,6 @@
   before_save :update_shipping_fees!, if: :complete?
   before_save :update_payment_fees!, if: :complete?
 
-  checkout_flow do
-    go_to_state :address
-    go_to_state :delivery
-    go_to_state :payment, :if => lambda { |order|
-      # Fix for #2191
-      if order.shipping_method.andand.require_ship_address and
```
sauloperez: the reason behind all this is that in Spree 2.0 we have `order.shipments` which have a `shipping_method`.

app/models/spree/order_decorator.rb
```diff
@@ -24,24 +24,6 @@
...
-        if order.ship_address.andand.valid?
-          order.create_shipment!
```
sauloperez: Also, `create_shipment!` is gone.

sauloperez:
very weird. All specs failing in Travis are passing in local even with the same bundler version (pretty old one btw).

sauloperez:
After long hours investigating this issue we see that it's easier (hopefully) to fix it in the `spree-upgrade-step-7` branch. It's shame that we cannot do it iteratively by merging into `master`.

daniellemoorhead:
>It's shame that we cannot do it iteratively by merging into `master`.

If we can get @enricostano "certified" by @oeoeaio to be merging then this problem should be solved?

sauloperez:
It would help indeed.

kirstenalarsen:
I believe it is very close :)

PR #1641, opened 2017-06-26
Upgrade bundler in ci

sauloperez:
The version used in Travis CI (1.7.6) is way too far from the one specified in our Gemfile (1.14.3). The more similar the versions in CI and our local environments are the more consistent the behaviours will be.

Ideally we should use the exact same version in the Gemfile.lock. We couldn't find a built-in way to do that in Travis CI so will give the [CodeClimate's grep engine](https://docs.codeclimate.com/v1.0/docs/grep) a try so that both bundler versions in `.travis.yml` and `Gemfile.lock` are in sync.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,9 @@ env:
     - CI_NODE_INDEX=3
     - CI_NODE_INDEX=4 KARMA="true" GITHUB_DEPLOY="true"
 
-before_install: gem install bundler
+before_install:
+  - gem install bundler
+  - gem cleanup bundler
 
 before_script:
   - cp config/database.travis.yml config/database.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ env:
     - CI_NODE_INDEX=3
     - CI_NODE_INDEX=4 KARMA="true" GITHUB_DEPLOY="true"
 
+before_install: gem install bundler
+
 before_script:
   - cp config/database.travis.yml config/database.yml
   - cp config/application.yml.example config/application.yml


enricostano:
```
$ bundle --version
Bundler version 1.7.6

$ gem install bundler
Fetching: bundler-1.15.1.gem (100%)
Successfully installed bundler-1.15.1
1 gem installed
```
Which version will use?

sauloperez:
AFAIK it will use the latest but we can fix it running `gem cleanup bundler` right after installing. That will remove all but the latest.

.travis.yml
```diff
@@ -22,6 +22,10 @@ env:
     - CI_NODE_INDEX=3
     - CI_NODE_INDEX=4 KARMA="true" GITHUB_DEPLOY="true"
 
+before_install:
+  - gem install bundler
```
mkllnk: This will install the latest version, right? That means that our builds are not deterministic, but depend on the bundler project releasing new versions. I'm wondering what's worse, automatically upgrading to the latest version or updating the Travis config regularly.

We could do something like:
```
gem install bundler -v "$(tail Gemfile.lock -n 1)"
```
But I'm wondering which bundler versions our production servers are running. Isn't that more important than our dev machines?

Aus production: Bundler version 1.10.6

It's also important to keep backwards compatibility. So having a new version on a dev machine and testing against an older version can be very useful as well.

What are your thoughts on that?

sauloperez: Agree on being deterministic. I wanted to address that with CodeClimate but your solution works perfectly.

Ideally we should have the same bundler version everywhere but I think that while developing what can become a blocker is having too diverse versions in CI and your dev env, potentially leading to different execution results.

Regarding backward compatibility I don't know if it's something we need to care about for bundler. You just need to ensure that it'll work in production.

enricostano: 👍 on using `tail`

Production is important, but out of the scope of this PR. That would be a PR for Ansible in `ofn-admin` repo.

mkllnk: I'm not convinced. If we have different execution results because of an old bundler version, then I would like to know that before we deploy to production. Or do you think that it would affect Travis, but not a production server?

If we want to talk about updating bundler on the production servers, that's another topic, right. But that topic needs to be discussed before we merge this, I think. A discussion about unifying the bundler version would be really good, actually.

What was the motivation for this change? Did you have problems with bundler?

By the way, I wouldn't use my code from above just like that. Not all bundler versions/configuration append their version. And I'm not sure about the whitespace at the beginning of that line.

sauloperez: > What was the motivation for this change? Did you have problems with bundler?

I opened this PR after having red tests in Travis but green in local. The first thing I try when experiencing these issues is to have as much consistent environments as possible in order to reproduce the failures. When I checked I saw very different bundler versions.

The failirues in Travis turned out to be related to bundler. Somehow the `cancan` gem wasn't present in the `bundle/cache` folder which bundler uses in Travis to install the dependencies.

My point is that having the same version would reduce the number of possible culprits when this happens. I can't tell however whether the issue was directly related to the old bundler version.

In any case, I think is also good to have our dependencies up to date, in this case bundler.

enricostano: With the recent Bundler versions `tail` is OK IMHO, and that's what we have in our `Gemfile.lock`

There is no need to deal with it in production first. Let's merge this IMHO I opened an issue in `ofn-admin` to cover this in production https://github.com/openfoodfoundation/ofn-install/issues/74

daniellemoorhead:
Hey @oeoeaio with @mkllnk away do you think you have time to do a quick lookover and merge if you're happy?

enricostano:
Can we merge this @oeoeaio @mkllnk ?

mkllnk:
I worked on a solution to install the Gemfile.lock's bundler version. Give me a bit more time to get it ready.

mkllnk:
Please have a look at https://github.com/openfoodfoundation/openfoodnetwork/pull/1713. What do you think?

sauloperez:
looks good. This one is not needed anymore then.

Issue #928, opened 2016-04-27
Make customer address and phone non compulsory info at checkout


myriamboure:
As a hub manager, I want to be able to choose if the address and phone number fields are compulsory or not at checkout.

And when I select a "delivery" type shipping method, the system should open an address field pre-filled in if I already have filled in an address above, and here it should be compulsory field of course (=delivery). The invoice address can be different from the delivery address, so in any case we need to have another address field for the delivery method, even if the hub manager choose to keep the address compulsory field on the first part.

(original discussion here: http://community.openfoodnetwork.org/t/make-customer-address-and-phone-non-compulsory-info-at-checkout/554/3)

myriamboure:
Ok, we had another discussion with our core beta testeur and this is a priority point. @blancnic do you think it's possible for you to work on that? See the discussion on Discourse.

myriamboure:
Complement to the demand: the hub wants the clients to be able to order easily from a mobile without  having to log in, and entering the min info possible.

daniellemoorhead:
Hi @myriamboure I can't see this being a big piece of work (feel free to correct me on this @oeoeaio or @enricostano or @Matt-Yorkley or @pacodelaluna or anyone else reading this), so I wonder whether Panier Ruse would be willing to contribute to its development?

myriamboure:
To add in the scope : give the possibility for the hub manager to disallow comment on the checkout page (request from a user in France)

myriamboure:
To add in the scope : give the possibility for the hub manager to disallow comment on the checkout page (request from a user in France)

myriamboure:
@daniellemoorhead I wrote a quick spec so that @ltrls can start working on it, can you tell me if that wounds find to you ? It's  super quick, half a page ;-) https://docs.google.com/document/d/1dsimTJdu6XZ4-l16zxbDegSfMyNTxofBjplzA_lSmLw/edit?usp=sharing

myriamboure:
I asked  Panier Rusé if they can chip in a bit, else we will try to cover that from the OFFrance budget as it seems to be pretty low cost / quick thing.

daniellemoorhead:
@myriamboure I don't have a problem but I don't have the same understanding of the system as @sstead and @oeoeaio and @kirstenalarsen so perhaps they could have a look at this and understand if there is any impact?

myriamboure:
Including https://github.com/openfoodfoundation/openfoodnetwork/issues/1197 in the scope

[...]

PR #1642, opened 2017-06-26
Dont assign distributor address to pickups

sauloperez:
This is one of the two parts of the `Order` model that get on the way of the Spree 2.0 upgrade as stated in https://community.openfoodnetwork.org/t/order-model-step7/979. TL;DR; `#shipping_method` is removed from `Order` in that version.

So far when saving an order we assign the distributor's address to the orders that are picked up (not delivered) but setting the name and phone number from the billing address.

By inspecting the ceckout's UI as well the as the rails views, this is not shown anywhere, so there's no need to assign it in the first place.

**Current behaviour**

ofn_master

**With this PR**

ofn_without_validation
 
We also had the need to refactor shipping methos by introducing `Delivery`, `Pickup` and `deliverty?`.

While reading the codebase we constantly find ourselves "computing" `the actual meaning og require_ship_address` in our heads to understand whether we're talking about a delivery or a pickup. This refactor makes the itent behind all this logic a bit more obvious, which makes the spree upgrade a bit easier.

### DIFF
--- a/spec/controllers/checkout_controller_spec.rb
+++ b/spec/controllers/checkout_controller_spec.rb
@@ -79,7 +79,12 @@ describe CheckoutController do
     end
 
     it "only clears the ship address with a pickup shipping method" do
+      shipping_method = build(:shipping_method)
+
+      order.stub(:shipping_method).and_return(shipping_method)
+      shipping_method.stub(:instance_for).with(order).and_return(Pickup.new(order))
       order.stub_chain(:shipping_method, :andand, :delivery?).and_return false
+
       # We render the page populating the last one the user used
       expect(order).to receive(:ship_address=).with(Spree::Address.default).twice
 
--- a/spec/controllers/checkout_controller_spec.rb
+++ b/spec/controllers/checkout_controller_spec.rb
@@ -65,22 +65,25 @@ describe CheckoutController do
     end
 
     it "clears the ship address when re-rendering edit" do
-      controller.should_receive(:clear_ship_address).and_return true
       order.stub(:update_attributes).and_return false
+      expect(order).to receive(:clear_ship_address).and_return(true)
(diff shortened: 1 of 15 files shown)


kirstenalarsen:
not shown in interface but I think it might affect reports and maybe invoices, order interface etc? I’d be a bit cautious about thinking this has no implications anywhere

sauloperez:
By looking at the uses of `ship_address` I can't see any impact in those sections. We seem to always check if the order has a ship address first (which means checking for pickup or delivery).

myriamboure:
@sauloperez there might be some intereference as well with the issue @ltrls is working on https://github.com/openfoodfoundation/openfoodnetwork/issues/928, he will put his PR in progress on github today so that you can understand what he has done and see together if there is an impact.

spec/controllers/checkout_controller_spec.rb
```diff
@@ -65,22 +65,31 @@
...
+
       spree_post :update, order: {}
     end
 
     it "only clears the ship address with a pickup shipping method" do
-      order.stub_chain(:shipping_method, :andand, :require_ship_address).and_return false
-      order.should_receive(:ship_address=)
-      controller.send(:clear_ship_address)
+      shipping_method = build(:shipping_method)
+      shipping_method_instance = Pickup.new(order)
```
enricostano: Do we really need this?

sauloperez: Not really, the var is not used besides as return value of the stub

sauloperez:
@myriamboure there is no interference so far, but since that is still WIP I can't tell for sure.

sauloperez:
We'll address the problems `shipping_address_from_distributor` brings to the Spree upgrade with a smaller change since this didn't inspire much confidence.

Issue #1095, opened 2016-07-10
Untranslated strings in front end interface


myriamboure:
I think this has not been covered actually in the work done on business model configuration.
It seems that when I create a new store and click on the link "access my store", the page is not translated in French.

![screenshot from 2016-07-10 21-47-02](https://cloud.githubusercontent.com/assets/8135376/16715695/f3cfac4c-46e7-11e6-8168-747a24d894e6.png)

I checked in Transifex and I couldnt find the source strings, so I guess it has not been translated, am I wrong? I think the text is hardcoded, am I write? Some strings that went through our radar on the front-end translation job I guess ;-)
@mkllnk, @blancnic, @elf-pavlik  can we add those strings? It's needed for any non English speaking country actually and a front end issue.

mkllnk:
The latest translations have not been copied into the transifex branch. I will have a look why that is. I also just updated the source file. That means that there will be new translations available in Transifex within the next 24h.

mkllnk:
@myriamboure You were right. These strings are still hard-coded in these two files:

- `app/assets/javascripts/templates/admin/panels/enterprise_package.html.haml`
- `app/views/admin/enterprises/_change_type_form.html.haml`

They have been there for a year.

Is someone keen to do the internationalisation work and open a pull request?

myriamboure:
@mkllnk thank you for your answers, and for fixing the transifex branch :-) @blancnic do you think you can do the internationationalisation work for the two files mentionned, so that the strings can be translated in Transifex? @sigmundpetersen that also concerns Norway.

blancnic:
@myriamboure : sorry didn't see your msg. Yes I'm going to take a look

mkllnk:
As part of the quote for multi-lingual instances, I identified all files that still contain untranslatable language strings.
```
app/views/admin/enterprise_roles/_enterprise_role.html.haml
app/views/admin/enterprise_roles/_form.html.haml
app/views/admin/enterprise_roles/index.html.haml
app/views/admin/enterprises/_actions.html.haml
app/views/admin/enterprises/_change_type_form.html.haml
app/views/admin/enterprises/_enterprise_user_index.html.haml
app/views/admin/enterprises/_new_form.html.haml
app/views/admin/enterprises/edit.html.haml
app/views/admin/enterprises/form/_about_us.html.haml
app/views/admin/enterprises/form/_address.html.haml
app/views/admin/enterprises/form/_contact.html.haml
[...]

PR #1643, opened 2017-06-27
Add missing translations on views

pacodelaluna:


pacodelaluna:
@mkllnk Related to the issue https://github.com/openfoodfoundation/openfoodnetwork/issues/1095, please let me know if any issue.

config/locales/fr.yml
```diff
@@ -475,9 +475,9 @@ fr:
         title: Paramètres de facturation
         invoice_style2?: Utiliser le modèle de facture alternatif qui détaille le montant de TVA agrégé par taux et l'information du taux de TVA par produit (pas adapté pour les instances affichant les prix HT)
         enable_receipt_printing?: Afficher les options d'impression de tickets de caisse dans le menu déroulant des commandes?
-  home:
+  shops:
```
pacodelaluna: No worries, I will also update the translations on Transifex once the commit is merged to avoid any regression.

daniellemoorhead:
Hi @pacodelaluna, @mkllnk hasn't been doing OFN things this week, he'll be back next Wednesday though so we should be able to make this a priority then. Cheers!

config/locales/fr.yml
```diff
@@ -516,9 +516,9 @@ fr:
     already_ordered:
       cart: "panier"
       message_html: "Vous avez déjà passé une commande pour ce cycle de vente. Vérifiez votre %{cart} pour voir les produits commandés. Vous pouvez annuler ou modifier votre commande jusqu'à la fermeture du cycle de vente."
-  home:
+  shops:
```
mkllnk: I hope this doesn't cause a conflict when merging changes from Transifex after the next release. Let's see...

enricostano: I was about to comment the same thing, we should only commit `en.yml` file and manage the rest through Transifex process 🙏

sauloperez: We must add this then at the top-level documentation of the `en.yml` that @oeoeaio added. It gives some guidelines when dealing with translations in OFN.

enricostano: This is a good example where the Code Climate engines could come handy, the [grep](https://docs.codeclimate.com/docs/grep) one for instance.

sstead:
@mkllnk I've checked the pages listed and could not see any translation issues. 

Note I saw this again, but it was already spotted in previous PR...
image

This is probably not related but on the shops page only some category icons were showing...
image.

mkllnk:
I think that these two issues are unrelated. I'm merging it.

config/locales/fr.yml
mkllnk: I thought about this a bit more and it is very unlikely to cause a conflict. But Transifex will overwrite `fr.yml` once it's translated. Just when we get new translations from Transifex and have changes in the master branch at the same position of the locale, we can get merge conflicts.

PR #1645, opened 2017-06-28
Optional checkout fields (WIP)

ltrls:
Working on issue #928, I try to make some Address fields optional at checkout. It was not simple as I thought, and I'd like some advice on the way I'm doing it.
I needed to save bill addresses without an address but still with first and last names. I thought I could (conditionaly) bypass validations on the order's bill_address, but I ran into a problem: even if the address itself has already been saved (using `#save(validate: false)`), whenever the order gets saved its associations will have their fields validated, and their errors added to the order's errors (I think this comes from the use of `accept_nested_attributes_for`).
I found that having `belongs_to :bill_address, validate: false` in Spree's Order model worked unfortunately this can't be overriden. The only way I found was to reset `validate` callbacks in the Address decorator [(here)](https://github.com/ltrls/openfoodnetwork/commit/2d3b73dea2a6cf4d1c149f8a6e3304938ca91d1e) but this removes address validations everywhere. I added a "manual" validation if the addresses are required, but I'm wondering if I should add this step each time an address is saved via user input, like when a hub edits its address. Writing this I guess I should, it's still unclear for me if I'm not missing something in all of this. any feedback is really welcome!

### DIFF
--- a/app/models/spree/address_decorator.rb
+++ b/app/models/spree/address_decorator.rb
@@ -2,6 +2,9 @@ Spree::Address.class_eval do
   has_one :enterprise
   belongs_to :country, class_name: "Spree::Country"
 
+  self.reset_callbacks(:validate)
+  validates :firstname, :lastname, :presence => true
+
   after_save :touch_enterprise
 
   geocoded_by :geocode_address
@@ -30,6 +33,12 @@ Spree::Address.class_eval do
     address_part2= [city, zipcode, state.andand.name]
     filtered_address = address_part2.select{ |field| !field.nil? && field != '' }
     filtered_address.compact.join(', ')
+  end
+
+  def validate_required_address
+    [:address1, :city, :zipcode, :country].each do |field|
+      errors.add(field, :blank) if send(field).blank?
+    end
   end
 
   private
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -17,6 +17,7 @@ class CheckoutController < Spree::CheckoutController
   end
 
   def update
+    @order.bill_address.validate_required_address unless @current_distributor.require_bill_address
     if @order.update_attributes(object_params)
       check_order_for_phantom_fees
       fire_event('spree.checkout.update')
--- a/app/views/admin/enterprises/form/_shop_preferences.html.haml
+++ b/app/views/admin/enterprises/form/_shop_preferences.html.haml
@@ -75,3 +75,33 @@
     .five.columns.omega
       = f.radio_button :allow_order_changes, true, "ng-model" => "Enterprise.allow_order_changes", "ng-value" => "true"
       = f.label :allow_order_changes, t('.allow_order_changes_true'), value: :true
+.row
+  .alpha.eleven.columns
+    .three.columns.alpha
+      = f.label :require_phone_number, t('.enterprise_require_phone_number')
+    .eight.columns.omega
+      = f.check_box :require_phone_number
+.row
+  .alpha.eleven.columns
+    .three.columns.alpha
+      = f.label :require_bill_address, t('.enterprise_require_bill_address')
+    .eight.columns.omega
+      = f.check_box :require_bill_address
+.row
+  .alpha.eleven.columns
+    .three.columns.alpha
+      = f.label :hide_comment_box, t('.enterprise_hide_comment_box')
+    .eight.columns.omega
+      = f.check_box :hide_comment_box
+.row
+  .alpha.eleven.columns
+    .three.columns.alpha
+      = f.label :check_the_only_shipping_option, t('.enterprise_check_the_only_shipping_option')
+    .eight.columns.omega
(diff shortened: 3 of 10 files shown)


myriamboure:
@enricostano is there any connexion to the various PR you did with Saulo on this with the Spree upgrade ? Maybe @oeoeaio or @mkllnk will have some inputs for Pierre ?

app/models/spree/address_decorator.rb
```diff
@@ -2,6 +2,9 @@
   has_one :enterprise
   belongs_to :country, class_name: "Spree::Country"
 
+  self.reset_callbacks(:validate)
```
enricostano: 😱 we need to investigate how to avoid this

sauloperez:
First, something that sounds weird to me is:

> I needed to save bill addresses without an address but still with first and last names

It leads me to think that what we are talking about is not a bill address (there's none) but about two different concepts: the name of the person associated to the order and the bill address. Also, having all columns with null values except for few of them is a clear symptom of this IMO.

However, it is Spree who adds `firstname` and `lastname` to the `addresses` table, so splitting this table in two doesn't look like an option to me and will have to live with it.

Secondly, I would reconsider how to store the enterprise settings. There are well known patterns to model such common feature. Take a look at this article http://culttt.com/2015/02/02/storing-user-settings-relational-database/. I think it explains the prons and cons quite well. Replace user with enterprise and we have our exact problem.

I would personally create a new `enterprise_settings` table with a column per setting. The enterprise table is already too big, with too many responsibilities, and this would another one.

We could then add conditionals on the model validations using http://guides.rubyonrails.org/active_record_validations.html#using-a-symbol-with-if-and-unless to start with, although I'm not a big fan of cluttering the model.

ltrls:
> I would personally create a new enterprise_settings table with a column per setting. The enterprise table is already too big, with too many responsibilities, and this would another one.
I'd gladly do that if it's best
> We could then add conditionals on the model validations using http://guides.rubyonrails.org/active_record_validations.html#using-a-symbol-with-if-and-unless to start with, although I'm not a big fan of cluttering the model.
By that you mean always validate name, and validate street, etc conditionaly in the Address model? That's kind of what I did, except I thought I couldn't get the order's hub's parameters from the address instance so I had to "validate from outside". Not sure why I thought that though, I could probably do that. let me see...

myriamboure:
@mkllnk or @oeoeaio any chance one of you can give some advice to @ltrls as well on how you see that ? :-)

enricostano:
@myriamboure I think this is related to the "customize all the things" approach we've been talking lately. I suggest to discuss the side effects on the product before we try to implement it.

mkllnk:
Yes, it looks like this needs a bit more thought about how it should work and how this can be implemented. I would need at least an hour to get my head around how Spree works here and how we would need to change that.

daniellemoorhead:
@mkllnk @enricostano @ltrls where do we go with this next? Do we schedule some time for @mkllnk to get his head into Spree and come up with an alternative solution @myriamboure?

myriamboure:
Hi @daniellemoorhead we are happy for @mkllnk to spend some time on it we would like to move forward on that. @enricostano was wondering if that should be specific to France, or if we need a more global agreement on the ux design for checkout and what should be customizable or not. So yes still a priority as it's required basically by all of our users (I guess users in Uk and Aus and others probably have also similar requests?), sorry if I didn't answer quickly I was really disconnected in the farm :-o

As input just pointing here the original discussion where there had kind of consensus from Canada, Uk, Norway, Australia that this needed to be customizable at the hub level, so I would vote for moving forward in that direction ;-) https://community.openfoodnetwork.org/t/make-customer-address-and-phone-non-compulsory-info-at-checkout/554/3

myriamboure:
Btw if needed as @ltrls has started to work on that and is  happy to finish if @mkllnk that makes sense that you pair with @ltrls on this that's great too :-)

daniellemoorhead:
Perhaps @myriamboure it would be worth having @enricostano and @sauloperez work on this rather than @mkllnk? I say this because they have just as much (if not more now?) knowledge of spree and will be able to come up with a solution, and also they are in the same timezone so can work with @ltrls a lot easier than we can from Australia. Thoughts?

myriamboure:
Sure @daniellemoorhead ! @enricostano @sauloperez do you have time to spend some time with @ltrls to work on it ? Of course the time you spend will be invoiced to OFFrance :-)

myriamboure:
As I closed the connected issues (can't be prioritized before Spee upgrade is complete, so moved to whishlist), I propose to also close that PR (can still be reopened when we work on it again ?) What do you think @sauloperez @enricostano @ltrls ?

#gitcull2017 https://community.openfoodnetwork.org/t/checkout-customization-adapt-fields-required-to-the-hub-model-and-customer-type/1163

sauloperez:
Agree @myriamboure . We can refer back to it whenever we need to. It's closed but not gone.

PR #1650, opened 2017-06-30
Stop using Order#shipment as it will be deprecated

sauloperez:
Spree 2.0 adds a deprecation warning to `Order#shipment` and Spree 2.1 removes it. So we better do not depend on it. Using it would give the wrong message that orders have a single shipment. 

By what I could find out, Spree added support for multiple shipments while keeping the `Order#shipment` method. They update the shipping method of the most recent shipment of the order if there's any. See https://github.com/spree/spree/blob/1-3-stable/core/app/models/spree/order.rb#L346

### DIFF
--- a/spec/controllers/line_items_controller_spec.rb
+++ b/spec/controllers/line_items_controller_spec.rb
@@ -113,7 +113,7 @@ describe LineItemsController do
         item_num = order.line_items.length
         initial_fees = item_num * (shipping_fee + payment_fee)
         expect(order.adjustment_total).to eq initial_fees
-        expect(order.shipment.adjustment.included_tax).to eq 1.2
+        expect(order.shipments.last.adjustment.included_tax).to eq 1.2
 
         # Delete the item
         item = order.line_items.first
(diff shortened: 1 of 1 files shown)


spec/controllers/line_items_controller_spec.rb
```diff
@@ -113,7 +113,7 @@
         item_num = order.line_items.length
         initial_fees = item_num * (shipping_fee + payment_fee)
         expect(order.adjustment_total).to eq initial_fees
-        expect(order.shipment.adjustment.included_tax).to eq 1.2
+        expect(order.shipments.last.adjustment.included_tax).to eq 1.2
```
sauloperez: this is exactly what `#shipment` does internally.

sauloperez:
@oeoeaio this can be safely merged. No need for testing here.

sauloperez:
Is there something else to be done to get this merged?

oeoeaio:
Nope, I'll do it now. :)

PR #1651, opened 2017-06-30
Use order shipments

sauloperez:
This is a work in progress to address what we agreed on in https://community.openfoodnetwork.org/t/shipping-rate-model-step7/985/2

> We have definitely abused order.shipping_method a lot, but when it comes down to it, I think we can pretty much use order.shipment.shipping_method (or equivalent) as a drop in replacement for order.shipping_method.

The goal here is exactly that: not to use `order#shipping_method` anymore and go through `order#shipments` instead.

## TODO

- [ ] Add documentation 
- [ ] Figure out how to abstract `order.shipments.last`

## How is this related to Spree upgrade?

`Order#shipping_method` is gone in Spree 2.0, so any line of code that uses it in OFN will fail. We can provide the same behaviour but being spree-2.0 compatible without having to wait for a specific spree-2.0 branch :metal: 

You can get further details in https://community.openfoodnetwork.org/t/order-model-step7/979?u=sauloperez

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
@@ -7,6 +7,7 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
   prefix = "order_#{Checkout.order.id}#{CurrentUser.id or ""}#{CurrentHub.hub.id}"
 
   for field in $scope.fieldsToBind
+    # SHIPPING_METHOD
     localStorageService.bind $scope, "Checkout.order.#{field}", Checkout.order[field], "#{prefix}_#{field}"
 
   localStorageService.bind $scope, "Checkout.ship_address_same_as_billing", true, "#{prefix}_sameasbilling"
--- a/app/serializers/api/current_order_serializer.rb
+++ b/app/serializers/api/current_order_serializer.rb
@@ -1,4 +1,5 @@
 class Api::CurrentOrderSerializer < ActiveModel::Serializer
+  # SHIPPING_METHOD
   attributes :id, :item_total, :email, :shipping_method_id,
              :display_total, :payment_method_id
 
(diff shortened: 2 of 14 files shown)


app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
```diff
@@ -7,6 +7,7 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
   prefix = "order_#{Checkout.order.id}#{CurrentUser.id or ""}#{CurrentHub.hub.id}"
 
   for field in $scope.fieldsToBind
+    # SHIPPING_METHOD
```
sauloperez: All these tags from the git history once done. I will help us identify the offenses to fix.

app/serializers/api/current_order_serializer.rb
```diff
@@ -15,4 +15,9 @@ def payment_method_id
   def display_total
     object.display_total.money.to_f
   end
+
+  def shipping_method_id
+    return unless object.shipments.any?
+    object.shipments.last.shipping_method_id
+  end
```
enricostano: Maybe we can add some documentation here explaining why we need this and maybe telling that in the future we should move from `order.shipments.last` to something more robust

spec/controllers/spree/admin/shipping_methods_controller_decorator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    context 'when the shipping method is not in use' do
+      it 'allows to destroy the shipping method' do
+        expect { delete :destroy, id: shipping_method.id }
+          .to change(Spree::ShippingMethod, :count)
+      end
+    end
+  end
+end
```
enricostano: 😍

app/serializers/api/current_order_serializer.rb
sauloperez: Agree on the doc adding a TODO. I want to abstract `order.shipments.last` somehow.

sauloperez:
This also needed for the Spree upgrade

sauloperez:
we totally change the way we deal with the spree upgrade since this was open, so no reason to still exist.

Matt-Yorkley:
@sauloperez Is this a good example of where we could test the ideas talked about today in the Spree Upgrade meeting?

sauloperez:
Yep! we could work on this one with @luisramos0

luisramos0:
ok, I have reviewed all commits here and created issues for each part.

Issue #1623, opened 2017-06-16
Bug on customer account display


myriamboure:
URL : https://www.openfoodfrance.org/account#/#hub-demo-open-food-france
![screenshot from 2017-06-16 14-20-53](https://user-images.githubusercontent.com/8135376/27226487-9bb90070-529f-11e7-8ea4-63449928d1b3.png)
Any idea why we have that @pacodelaluna ?

myriamboure:
The "NaN" € instead of the total amount on the top right

myriamboure:
That's strange also, the calculation are wrong on the amount owed...

myriamboure:
![screenshot from 2017-06-27 18-41-15](https://user-images.githubusercontent.com/8135376/27599198-3f8bae84-5b68-11e7-940b-c1ad5c14aabe.png)

myriamboure:
![screenshot from 2017-06-27 18-41-15](https://user-images.githubusercontent.com/8135376/27599261-7e3ef1c2-5b68-11e7-80d1-cbf231ab2000.png)
I have the impression that's a rounding display on column "amount", as when I click on the order it's the right amount.
Do you have the same issue @sigmundpetersen and @sstead ?

enricostano:
`NaN` is a great JS classic :trollface:

myriamboure:
@ltrls awesome if you can look at that, I can't assign you yet but I can mention you at least ;-)

myriamboure:
Seems to be connected also to https://github.com/openfoodfoundation/openfoodnetwork/issues/853

PR #1653, opened 2017-06-30
Proof of concept: float string field

mllocs:
Tries to fix this issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/853

I'm basically using a hidden input field to store the numeric conversion of the input text field. Whenever the user types a value in the visible input, it is transformed and stored in the hidden field as a numeric value. 

http://g.recordit.co/kzHcqGD5qg.gif

This is just a proof of concept, applied to the `price` attribute in the "New product" form. 

I added a partial view that adds the hidden field and the javascript that syncs both fields, and then a Rails helper to make it easier to use that partial. 

---

Note:

With this fix, we accept any kind of format as an input but **we always display those values in the standard numeric way**, which means that it uses the dot for decimals and no thousands separator.

### DIFF
--- a/app/assets/javascripts/shared/util.js.coffee
+++ b/app/assets/javascripts/shared/util.js.coffee
@@ -1,13 +1,16 @@
 window.string_to_float = (string) ->
   # Replace all Currency Symbols, Letters and -- from the string
   string = string.replace(/[^0-9\.,]+/g, '')
+
   # If string ends in a single digit (e.g. ,2) make it ,20 in order for the result to be in "cents"
   if string.match(/[\.,]\d$/)
     string = string + '0'
-  # If does not end in ,00 / .00 then
-  # add trailing 00 to turn it into cents
+
+  # If does not end in ,00 / .00 then add trailing 00 to turn it into cents
   if !string.match(/[\.,]\d\d$/)
     string = string + '00'
+
   # Replace all Currency Symbols, Letters and -- from the string
   string = string.replace(/[\.,]/g, '')
+
   string / 100
--- a/app/assets/javascripts/shared/util.js.coffee
+++ b/app/assets/javascripts/shared/util.js.coffee
@@ -1,0 +1,13 @@
+window.string_to_float = (string) ->
+  # Replace all Currency Symbols, Letters and -- from the string
+  string = string.replace(/[^0-9\.,]+/g, '')
+  # If string ends in a single digit (e.g. ,2) make it ,20 in order for the result to be in "cents"
+  if string.match(/[\.,]\d$/)
+    string = string + '0'
+  # If does not end in ,00 / .00 then
+  # add trailing 00 to turn it into cents
+  if !string.match(/[\.,]\d\d$/)
+    string = string + '00'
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/shared/util.js.coffee
```diff
@@ -0,0 +1,16 @@
...
+
+  # If string ends in a single digit (e.g. ,2) make it ,20 in order for the result to be in "cents"
+  if string.match(/[\.,]\d$/)
+    string = string + '0'
+
+  # If does not end in ,00 / .00 then add trailing 00 to turn it into cents
+  if !string.match(/[\.,]\d\d$/)
+    string = string + '00'
+
+  # Replace all Currency Symbols, Letters and -- from the string
```
sauloperez: I'd say "remove". That's what we do, right?

myriamboure:
Sounds good! I think using a standard display with dot as decimal is really ok, as long as people can fill in either with comas or dots or any other conventionnal way ;-) @ltrls what do you think of this proposal ?

ltrls:
That's smart!
> With this fix, we accept any kind of format as an input but we always display those values in the standard numeric way, which means that it uses the dot for decimals and no thousands separator.

Price amounts are injected as localized strings (using Money#to_s). It's causing the #1623 issue as the [LocalizeCurrency](https://github.com/openfoodfoundation/openfoodnetwork/blob/0760d4cc1f26fd06f60b420293e3671f205d0a07/app/assets/javascripts/darkswarm/filters/localize_currency.js.coffee) filter fails at parsing the string if it's not in English format. I'm thinking of serializing amounts in English format and improving the filter to localize the separator too (as it's written at L11, but probably injecting the locale instead of the separator and using [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)), so that should solves this too :)

oeoeaio:
Others are better placed that me to assess the functionality here, and I realise that this is a spike, so is not fully formed, but just wanted to flag now that I would love to see this written as an angular directive so that we aren't polluting the global namespace. Specs etc. too, ya ya ya.

mkllnk:
Are we doing this in Javascript so that we don't have to deal with Spree? I had a Rails solution in mind. For example in `app/controllers/spree/admin/products_controller_decorator.rb`:
```ruby
  before_filter :parse_price

  def parse_price
    params[:product][:price] = stringToPrice params[:product][:price]
  end
```

There is probably a better way to organise the code and have it in a nice little lib file.
Do you see any downsides of that approach?

daniellemoorhead:
@ltrls @oeoeaio @mkllnk can we close this spike PR now that #1886 is merged?

mkllnk:
@daniellemoorhead #853 is still not solved. #1886 solved #1623. I think that the competing pull request to this one is #1831 which is very far, but not merged yet. I do prefer #1831 and would vote for closing this spike even though it is very interesting.

daniellemoorhead:
Closing in preference for #1831. Feel free to reopen this if there is value in keeping it open for now :)

PR #1655, opened 2017-07-01
Bump eventmachine version to resolve openssl issue on macOS

oeoeaio:
Fixed in `1.0.9` see: https://github.com/eventmachine/eventmachine/issues/602

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -130,4 +130,9 @@ group :development do
   gem 'guard-rspec'
   gem 'parallel_tests'
   gem 'rubocop', '>= 0.49.1'
+
+  # 1.0.9 fixed openssl issues on macOS https://github.com/eventmachine/eventmachine/issues/602
+  # While we don't require this gem directly, no dependents forced the upgrade to a version
+  # greater than 1.0.9, so we just required the latest available version here.
+  gem 'eventmachine', '>= 1.2.3'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -258,7 +258,7 @@ GEM
       eventmachine (>= 0.12.9)
       http_parser.rb (~> 0.5.3)
     erubis (2.7.0)
-    eventmachine (1.0.8)
+    eventmachine (1.2.3)
     excon (0.45.4)
     execjs (2.6.0)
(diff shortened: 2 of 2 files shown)


Gemfile.lock
```diff
@@ -258,7 +258,7 @@ GEM
       eventmachine (>= 0.12.9)
       http_parser.rb (~> 0.5.3)
     erubis (2.7.0)
-    eventmachine (1.0.8)
+    eventmachine (1.2.3)
```
sauloperez: if versions lower than this don't work we must specify so in the Gemfile.

oeoeaio: Yep, sure. It just feels weird because it is not a direct dependency of the project. It's only required by guard-rails. Anyway, I'll specify the version in the Gemfile for now...

sauloperez: Then, we need to find the right version of guard-rails. The one that depends on an eventmachine version that works.

enricostano: Yep, I agree with @sauloperez

oeoeaio: Couldn't force an update to event-machine through its dependents, so I just added a comment to the Gemfile explaining why a version was specified. Look ok?

sauloperez: I would simply add that the dependents of this gem do not require that version so the reader will understand why we did this

Gemfile
```diff
@@ -130,4 +130,8 @@ group :development do
   gem 'guard-rspec'
   gem 'parallel_tests'
   gem 'rubocop', '>= 0.49.1'
+
+  # We don't actually require this gem directly, but 1.0.9 fixed openssl issues
+  # on macOS: https://github.com/eventmachine/eventmachine/issues/602
+  gem 'eventmachine', '>= 1.0.9'
```
enricostano: Sorry to be a PITA, but I'm just curious. Why we can't use `gem 'eventmachine', '>= 1.2.3'`. Is there some dependency deep in the tree that is blocking us?

oeoeaio: No reason. 1.2.3 works fine. I'll do that!

oeoeaio:
@sauloperez @enricostano happy?

PR #1656, opened 2017-07-03
Fix before_filter cop for Rails 3.2

Matt-Yorkley:
Rubocop complains about any use of `before_filter` and `prepend_before_filter`, suggesting it should be `before_action`. I think this was introduced in Rails 4, and doesn't work with 3.2

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -52,6 +52,10 @@ Rails/DynamicFindBy:
 Rails/SkipsModelValidations:
   Enabled: false
 
+# Use before_filter instead of before_action. #TODO: remove when upgraded to Rails 4
+Rails/ActionFilter:
+  EnforcedStyle: filter  
+
 # Relaxed.Ruby.Style SETTINGS
 # These styles are a starting point for the conversation around conventions
 # They should be removed or tweaked and moved above as decisions are made


Matt-Yorkley:
@mkllnk a quick Rubocop adjustment. This is blocking deployment on staging for the shopfronts PR #1557.

oeoeaio:
@Matt-Yorkley have you rebased the branch you are trying to deploy?

The Rubocop config in `master` specifies a target rails version of `3.2`, which should not raise the `ActionFilter ` cop because it is smart enough to know that `before_action` is not available...

See docs here: https://rubocop.readthedocs.io/en/latest/cops_rails/

oeoeaio:
Sorry, I just checked for myself and I can see that it has been rebased...

CodeClimate doesn't seem to be respecting the `TargetRailsVersion` configuration, so I did some digging and it appears that CodeClimate is not using the latest version of RuboCop? It is using `0.46` and we are using `0.49.1`. The `TargetRailsVersion` only appears in `0.48`, so that is our problem...

I guess this PR is a good fix for now. Would be good to update to comment to reflect that fact that this is more a CodeClimate issue than anything else...

oeoeaio:
Did some more digging: we can specify the version of Rubocop that we want CodeClimate to use in our `.codeclimate.yml`. See docs here: https://docs.codeclimate.com/v1.0/docs/rubocop

It looks like `0.48` is the latest version available at the moment, but that should solve this particualr problem.

Let's do that instead?

sauloperez:
> Let's do that instead?

Yep. It would this one and other issues.

Matt-Yorkley:
Superceded by #1661

PR #1659, opened 2017-07-05
Soft delete shipping methods and enable destroys

sauloperez:
This removes the "hack" that forbids destroying shipping methods from /admin and makes use of Spree's soft delete implementation (`deleted_at` column in `spree_shipping_methods`).

This way admins can disable a particular shipping method, so no one uses it anymore, while still showing the shipping method for past orders in reports.

## TODO
- [x] remove N+1
- [ ] Check `/admin/orders` and `/accounts`
- [ ] QA order_cycle_management and customers reports: They must work like now and shipping_methods should be deletable.

## How is this related to Spree upgrade?

You are right to ask. `app/controllers/spree/admin/shipping_methods_controller_decorator.rb` below references the column `shipping_method_id` from the `spree_orders` table and this won't be longer available in Spree 2.0 aka. step 7.

By shipping this simple change into a release before the step 7 we reduce the chance of introducing bugs and more important, we make that actual step 7 release more lightweight, which people testing will thank.

### DIFF
--- a/app/controllers/spree/admin/shipping_methods_controller_decorator.rb
+++ b/app/controllers/spree/admin/shipping_methods_controller_decorator.rb
@@ -1,7 +1,6 @@
 module Spree
   module Admin
     ShippingMethodsController.class_eval do
-      before_filter :do_not_destroy_referenced_shipping_methods, :only => :destroy
       before_filter :load_hubs, only: [:new, :edit, :create, :update]
 
       # Sort shipping methods by distributor name
@@ -29,21 +28,8 @@ module Spree
         collection
       end
 
-      # This method was originally written because ProductDistributions referenced shipping
-      # methods, and deleting a referenced shipping method would break all the reports that
-      # queried it.
-      # This has changed, and now all we're protecting is Orders, which is a spree resource.
-      # Do we really need to protect it ourselves? Does spree do this, or provide some means
-      # of preserving the shipping method information for past orders?
-      def do_not_destroy_referenced_shipping_methods
-        order = Order.where(:shipping_method_id => @object).first
-        if order
-          flash[:error] = "That shipping method cannot be deleted as it is referenced by an order: #{order.number}."
-          redirect_to collection_url and return
-        end
-      end
-
       private
+
       def load_hubs
         @hubs = Enterprise.managed_by(spree_current_user).is_distributor.sort_by!{ |d| [(@shipping_method.has_distributor? d) ? 0 : 1, d.name] }
       end
--- a/lib/open_food_network/order_cycle_management_report.rb
+++ b/lib/open_food_network/order_cycle_management_report.rb
@@ -3,6 +3,7 @@ require 'open_food_network/user_balance_calculator'
 module OpenFoodNetwork
   class OrderCycleManagementReport
     attr_reader :params
+
     def initialize(user, params = {})
       @params = params
       @user = user
@@ -17,7 +18,13 @@ module OpenFoodNetwork
     end
 
     def search
-      Spree::Order.complete.where("spree_orders.state != ?", :canceled).distributed_by_user(@user).managed_by(@user).search(params[:q])
+      Spree::Order
+        .eager_load(shipments: :shipping_method)
+        .complete
+        .where("spree_orders.state != ?", :canceled)
+        .distributed_by_user(@user)
+        .managed_by(@user)
+        .search(params[:q])
     end
 
     def orders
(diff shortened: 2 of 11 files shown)


sauloperez:
@rob @kirstenalarsen @RohanM any idea where we might show shipping methods of past orders? As far as I see is not shown anywhere else in the reports.

app/controllers/spree/admin/shipping_methods_controller_decorator.rb
```diff
@@ -29,21 +28,8 @@ def collection
         collection
       end
 
-      # This method was originally written because ProductDistributions referenced shipping
-      # methods, and deleting a referenced shipping method would break all the reports that
-      # queried it.
-      # This has changed, and now all we're protecting is Orders, which is a spree resource.
-      # Do we really need to protect it ourselves? Does spree do this, or provide some means
-      # of preserving the shipping method information for past orders?
```
sauloperez: Yes, it does by implementing soft deletes: marking the row as deleted through an update instead of actually getting rid of it.

lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -33,41 +33,45 @@ def table_items
...
-       OpenFoodNetwork::UserBalanceCalculator.new(order.email, order.distributor).balance
+      billing_address = order.billing_address
+      [
+        billing_address.firstname,
+        billing_address.lastname,
+        order.distributor.andand.name,
+        customer_code(order.email),
+        order.email,
+        billing_address.phone,
+        Spree::ShippingMethod.unscoped.find(order.shipping_method_id).name,
```
sauloperez: I still need to iterate on this. As it is, introduces a N+1 :warning:

lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -17,7 +17,13 @@ def header
     end
 
     def search
-      Spree::Order.complete.where("spree_orders.state != ?", :canceled).distributed_by_user(@user).managed_by(@user).search(params[:q])
+      Spree::Order
+        .eager_load(:shipping_method)
```
sauloperez: This adds a `LEFT JOIN ON spree_shipping_methods`

lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -17,7 +27,13 @@ def header
     end
 
     def search
-      Spree::Order.complete.where("spree_orders.state != ?", :canceled).distributed_by_user(@user).managed_by(@user).search(params[:q])
+      Spree::Order
+        .eager_load(shipments: :shipping_method)
```
sauloperez: This adds a `LEFT JOIN ON spree_shipping_methods`

lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -33,44 +48,61 @@ def table_items
...
+        order.special_instructions
       ]
     end
 
+    # Returns the appropriate shipping method to display for the given order
+    #
+    # @param order [Spree::Order]
+    # @return [#name]
+    def shipping_method_to_display(order)
+      shipment = order.shipments.last
```
sauloperez: In places like these (reports) some thought on the UI is required. An order with multiple shipments won't be displayed in a table now.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -267,5 +279,18 @@ def total_units(line_items)
...
+    # @param order [Spree::Order]
+    # @return [#name]
+    def shipping_method_to_display(order)
+      shipment = order.shipments.last
+      if shipment
+        shipment.shipping_method
+      else
+        NullShippingMethod.new
+      end
+    end
```
enricostano: This is repeated in 2 places (at least), we should move it at model level. Could you please open an issue to tackle this in another PR?

spec/models/null_shipping_method_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper'
+
+describe NullShippingMethod do
+  describe '#name' do
+    it 'returns nil' do
+      expect(described_class.new.name).to eq(nil)
+    end
+  end
```
enricostano: cover `#require_ship_address` too

lib/open_food_network/order_cycle_management_report.rb
oeoeaio: I think this is ok. I don't really see any alternative.

oeoeaio:
Looks good. Don't know why we didn't do this to begin with.

Do we actually need `NullShippingMethod`? I don't think any reports should be pulling up orders that aren't complete, and all complete orders should have at least one shipment, and shipments require a shipping method.

If we do need it, I'm wondering if we should make the `NullShippingMethod` name configurable using a translation? Would be good to display something rather than an empty string?

The only other place that I can think of where we would be looking up shipping methods on past orders are the order interfaces (`/admin/orders` and `/accounts`).

sauloperez:
> Do we actually need NullShippingMethod? I don't think any reports should be pulling up orders that aren't complete, and all complete orders should have at least one shipment, and shipments require a shipping method.

That's a very good point and I totally agree but are we sure we don't inconsistent data in our databases? if you guys and UK don't see any order without shipment I will remove the `NullShippingMethod` an any other defensive code like `andand` on shipping method.

> The only other place that I can think of where we would be looking up shipping methods on past orders are the order interfaces (/admin/orders and /accounts).

Thanks for the heads-up. That's definitely a place to check.

lib/open_food_network/orders_and_fulfillments_report.rb
sauloperez: We'll need to wait for @oeoeaio 's comment https://github.com/openfoodfoundation/openfoodnetwork/pull/1659#issuecomment-314685297. We might not need it.

daniellemoorhead:
@sauloperez is this PR still valid/active? It's been 6 months since it was worked on...please close as part of #githubcull2018 if it's no longer relevant, or if put it back into an earlier column (all the things?) if you're not currently working on it.

sauloperez:
We need to resume this as part of the Spree upgrade. Could you please move it to the appropriate column? I'm on the go.

enricostano:
@myriamboure should we add this to any Spree upgrade epic?

myriamboure:
For consistency I would rather open an issue, link to this PR, close this PR and put the issue in the epic (we only put stories in an epic until now, no PR). That way we can close this PR without forgetting that we need to work on it, and we don't have old PR waiting. Which problem is this PR solving? From my understanding it is that enterprise users needs to be able to delete a shipping method. Agree @enricostano ?

sauloperez:
I can create an issue and close this PR. We can make these changes in smaller and multiple pull requests. This one was already too big.

daniellemoorhead:
I created the issue so am closing this PR as per comments above. Have also linked it to this to the Spree 2.0 epic so that it's easily found, and put it into the delivery train backlog.

myriamboure:
Not sure @daniellemoorhead we should put it already into the delivery train backlog as @sauloperez said he woud need to split it into small stories, so I suggest we only link the issue you created to the Spree epic and when they incept and decide on all the stories they will include the stories connected to this one and put them in the delivery train backlog... So I'm putting it back to "all the things" for now.

PR #1669, opened 2017-07-07
Add delivery and pickup

sauloperez:
#### What? Why?

While reading the codebase we constantly find ourselves "computing" the actual meaning of `ShippingMethod#require_ship_address` in our heads to understand whether we're talking about a delivery or a pickup.

This refactor makes the intent behind all this logic a bit more obvious, which makes the spree upgrade a bit easier.

#### What should we test?

I'd say having the test suite green is enough. It's a pretty simple change.

This is the code we read everyday working on the upgrade since `Order#shipping_method` is gone in Spree 2.0. The easier it is to understand the better.

### DIFF
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -64,6 +64,10 @@ Spree::ShippingMethod.class_eval do
     'Shipping'
   end
 
+  def delivery?
+    require_ship_address
+  end
+
   private
 
   def touch_distributors
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -64,6 +64,11 @@ Spree::ShippingMethod.class_eval do
     'Shipping'
   end
 
+  # Checks whether the shipping method is of delivery type, meaning that it
+  # requires the user to specify a ship address at checkout. Note this is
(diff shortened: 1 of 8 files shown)


app/models/spree/shipping_method_decorator.rb
```diff
@@ -64,6 +64,10 @@ def adjustment_label
     'Shipping'
   end
 
+  def delivery?
+    require_ship_address
+  end
+
```
enricostano: Can you please add some doc here?

daniellemoorhead:
@sstead can you coordinate with @oeoeaio to get this on staging and tested next week (26th July)?

mkllnk:
Sorry, this slipped through. I'm moving it to the next release.

daniellemoorhead:
Hey @sauloperez is this PR related to an issue or is it an orphan?

sstead:
Quick test, looks fine

daniellemoorhead:
Uh oh, Travis build failed...I think between @sstead testing it and now. @mkllnk any thoughts on why this happened? Strange!

mkllnk:
@daniellemoorhead We changed the pipeline so that Travis can run while Sally is testing. Since the build was fine before, it was likely to be fine now, but it failed. The tests have been very unreliable lately. That is a tech debt to clear up.

sauloperez:
awesome :tada: ! thanks guys !

PR #1671, opened 2017-07-11
Use indented style for multiline method calls

sauloperez:
#### What? Why?

This enables the Rubocop's `Style/MultilineMethodCallIndentation` cop with
indentend enforced style. Which makes you split multiline method calls like:

```ruby
  orders = Spree::Order
    .an_scope
    .another_scope
    .where(id: list_of_ids)
```

The more I touch certain parts of the code such as reports, the more long lines with multiple method call I find myself dealing with. I like this style because you can see the different methods being called at a glance without having to parse a long line while reading.

Thoughts?


oeoeaio:
> The more I touch certain parts of the code such as reports, the more long lines with multiple method call I find myself dealing with.

This doesn't enforce multiline method calls though right? It just describes the style in which they should be indented. I am happy with this style for indentation.

What happens to the files excluded from interrogation by this cop in `.rubocop_toto.yml`? Are they still excluded, or do you get offences for them being raised by now? Should we just autocorrect them with `rubocop -a` and remove this cop from `.rubocop_todo.yml` altogether?

Either way, could you move the config for this cop to the `OFN SETTINGS` section at the top of the file? The section you have added it to contains the styles I pulled from `relaxed.ruby.style` as a starting point. I'd eventually like to remove that section as we accept the configs as our standard or remove them if we don't want them.

sauloperez:
> It just describes the style in which they should be indented

Yep, exactly that. If you happen to split a long line in mutiple ones, you will have to conform to this style.

As I mentioned some other day the biggest benefit is having this feedback integrated directly in your editor. When I split the following line as:

before

I get the visual feedback on the left side and the following rubocop's message at the bottom:

`rubocop: Use 2 (not 12) spaces for indenting an expression spanning multiple lines.`

Wich disappear when I do:

after

You can check your editor in https://github.com/bbatsov/rubocop/blob/master/manual/integration_with_other_tools.md

Now I will autocorrect any violations (I think there are just a couple), update the `rubocop_todo.yml` and move the config to the top of `rubocop.yml` as you suggest.

.rubocop.yml
```diff
@@ -11,6 +11,7 @@ AllCops:
     - 'db/**/*'
     - 'config/**/*'
     - 'script/**/*'
+    - 'vendor/**/*'
```
sauloperez: without this it parses the sourcecode of the gems in your `vendor/bundle` :see_no_evil:

oeoeaio: 👌

enricostano:
@oeoeaio I see that @sauloperez addressed your comments, so... I merged it 😬

Issue #1674, opened 2017-07-11
UK Invoice wording and localisation


Matt-Yorkley:
Invoices currently say "GST" in the UK, instead of "VAT".

Invoices are titled "Tax Invoice", but should probably just be "Invoice". 

Date format is shown as "YYYY-MM-DD" instead of "DD-MM-YYYY".

![screenshot from 2017-07-11 13-41-23](https://user-images.githubusercontent.com/9029026/28068761-cf8a1116-663e-11e7-8e1c-af076fcba81f.png)
![screenshot from 2017-07-11 13-41-29](https://user-images.githubusercontent.com/9029026/28068762-cf9491e0-663e-11e7-892f-42393ea92b39.png)

myriamboure:
@Matt-Yorkley You should check the other invoice format we have done in France, it might  answer your need... see in admin "user alternative invoice" thing ;-)

sineadfenton:
Hey all,

We've had a couple more requests on translating GST into VAT for the UK. I just wanted to follow up in relation to the #1684 ticket and if this was that was currently being actioned?

(excuse my ignorance! just can't quite work it out and it would be good to feedback to the community) 

cheers,

S

Matt-Yorkley:
@sineadfenton it should be going through soon.

sineadfenton:
![](https://github.trello.services/images/mini-trello-icon.png) [GST/VAT Invoice for Shoppers](https://trello.com/c/YWiVRIah/47-gst-vat-invoice-for-shoppers)

lin-d-hop:
Merged so closing

PR #1684, opened 2017-07-13
Invoice date and translation adjustments

Matt-Yorkley:
#### What? Why?

Addressing #1674. Hub users identified issues with text and date localisation UK printable invoices.

#### What should we test?

Printable invoices when using en-GB

### DIFF
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -205,6 +205,12 @@ en-GB:
         max_fulfilled_units: "Max Fulfilled Units"
         order_error: "Some errors must be resolved before you can update orders.\nAny fields with red borders contain errors."
         variants_without_unit_value: "WARNING: Some variants do not have a unit value"
+      invoice:
+        issued_on: Issued on
+        tax_invoice: TAX INVOICE
+        code: Code
+        from: From
+        to: To
     order_cycles:
       edit:
         choose_products_from: 'Choose Products From:'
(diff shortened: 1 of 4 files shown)


config/locales/en-GB.yml
```diff
@@ -238,10 +238,20 @@ en-GB:
...
+  invoice_column_unit_price_with_taxes: "Unit price (Incl. tax)"
+  invoice_column_unit_price_without_taxes: "Unit price (Excl. tax)"
+  invoice_column_price_with_taxes: "Total price (Incl. tax)"
+  invoice_column_price_without_taxes: "Total price (Excl. tax)"
+  invoice_column_tax_rate: "Tax rate"
+  tax_invoice: "INVOICE"
+  tax_total: "Total tax (%{rate}):"
+  total_excl_tax: "Total (Excl. tax):"
+  total_incl_tax: "Total (Incl. tax):"
+  invoice_tax_total: "VAT Total:"
```
enricostano: are these keys in `en.yml` too? AFAIK we should only touch `en.yml` to not break Transifex process.

cc/ @mkllnk

enricostano:
BTW... what's the difference between `en_GB.yml` and `en-GB.yml`? I'm just curious.

config/locales/en-GB.yml
Matt-Yorkley: Yes, they're in both.

Matt-Yorkley:
The duplicate files are a separate Transifex issue that should be fixed soon...

config/locales/en-GB.yml
enricostano: We should use the Transifex process IMHO, if not we'll break it

Matt-Yorkley:
Thanks @mkllnk. I'm not really in charge of the translation or Transifex stuff for the UK, but it's handy to be able to change a few keys like this. I'll take out the `en_GB.yml` changes. Maybe our translation strategy needs a bit of discussion in the UK, but we have it set to pull `en` keys when the `en-GB` keys are missing on production, so I think keeping every key updated isn't a huge issue for us.

Matt-Yorkley:
I've removed the `en_GB.yml` changes now so I think it's ready to go...

lin-d-hop:
@mkllnk Is this ok to be merged for July release as well? Looks like @Matt-Yorkley acted on @enricostano s requested changes :-)

mkllnk:
Yes, I'm planning a final test and merge tomorrow.

sstead:
@mkllnk 
**Testing notes**
- I was told to check that nothing had changed on the invoice in the Aus staging. It's a bit harder to test because we currently have a translation bug in production obscuring the invoice, making comparison difficult.
- From what I can see there are differences in the invoice (see below). From memory, the invoice will be laid out differently for instances that have >1 tax rate than for instances that have just 1. Staging was setup with >1 tax rate and production has 1. So this probably explains the differences in display, but I can't be sure. I tried to delete the second tax rate from staging to confirm this, but it didn't results in a change in the invoice display.
image

As another note, in Aus, we should also have our date displayed as DD-MM-YYYY. Can we implement this on Aus production?

sstead:
Good to go!

Issue #1679, opened 2017-07-11
Datepickers not working


Matt-Yorkley:
- Datepicker widgets have missing icons when viewing individual Order Cycles.

![screenshot from 2017-07-11 15-27-00](https://user-images.githubusercontent.com/9029026/28073678-66c3066e-664e-11e7-84f0-966f0d96ea79.png)

- Datepicker icons on the main Order Cycles page sometimes don't work when clicked?

![screenshot from 2017-07-11 15-26-43](https://user-images.githubusercontent.com/9029026/28073598-2104b7da-664e-11e7-81da-fcf1810c1272.png)

Matt-Yorkley:
Datepicker buttons not working intermittently was a UK localisation issue where saved dates displayed as yyyy-mm-dd but new dates used yyyy/mm/dd format (only on UK) which caused parsing issues and javascript errors, and no datepicker popup.

sineadfenton:
![](https://github.trello.services/images/mini-trello-icon.png) [Datepickers Not Working](https://trello.com/c/jJrOJHRo/65-datepickers-not-working)

PR #1688, opened 2017-07-14
Uk/datepickers

Matt-Yorkley:
#### What? Why?

Addressing #1679. Adjustments to datepickers that were not working very well. See original issue.

#### What should we test?

Datepickers on Order Cycle and Edit Order Cycle pages. Mostly with UK settings.


app/assets/javascripts/admin/util.js.erb
```diff
@@ -1,6 +1,6 @@
 $(document).ready(function() {
   $('.datetimepicker').datetimepicker({
-    dateFormat: Spree.translations.date_picker,
+    dateFormat: "yy-mm-dd",
```
enricostano: Are we using this somewhere else? Shouldn't we modify the Spree setting in that case?

Matt-Yorkley: I thought it would be best to just fix it, as any other formats can break the datepicker. If it's taking translations it could use the wrong format in some situations, no?

Matt-Yorkley: Or did you mean change the default Spree setting as well as this? Maybe both would be preferable...?

Matt-Yorkley: It looks like the default has already been set to "yy-mm-dd" for `en.yml`, which is correct. It's only on other languages where the key isn't explicitly set in the locale file that it defaults to "yy/mm/dd", and then breaks. I think ignoring the locale settings for this one is the best option, as alternate settings will break site the functionality.

enricostano: Mh, OK then, maybe you could add this findings as documentation in the code. will be helpful to understand the change in the future

oeoeaio: That's weird that the translation doesn't fall back to the one defined in `en.yml`...

lin-d-hop:
@Matt-Yorkley can you take a look at @enricostanos request for some test coverage?

Matt-Yorkley:
Spec added.

spec/features/admin/order_cycles_spec.rb
```diff
@@ -58,6 +58,38 @@
     end
   end
 
+  describe 'listing order cycles with other locales' do
+    let!(:oc_de) { create(:simple_order_cycle, name: 'oc', orders_open_at: '2012-01-01 00:00:00') }
+
+    before { I18n.locale = :de }
+    after { I18n.locale = :en }
```
sauloperez: IMO for these cases an [around hook](https://relishapp.com/rspec/rspec-core/v/2-0/docs/hooks/around-hooks)  is better suited

spec/features/admin/order_cycles_spec.rb
```diff
@@ -58,6 +58,38 @@
...
+
+    context 'using datepickers' do
+      it "correctly open the datepicker and changes the date field" do
+        expect(I18n.locale).to eq :de
```
sauloperez: I don't see why we need assertion for this. We just set the locale to `de`. It feels like we don't trust ruby.

daniellemoorhead:
Hey there, have to move this to September release as it hasn't moved anywhere for a few weeks. @Matt-Yorkley what's left to do to get this through to done?

Matt-Yorkley:
@daniellemoorhead I thought it was ready? The requested specs have been added and the code review comments have been addressed.

enricostano:
@sauloperez could you please review this again?

spec/features/admin/order_cycles_spec.rb
```diff
@@ -58,6 +58,39 @@
...
+    let!(:oc_de) { create(:simple_order_cycle, name: 'oc', orders_open_at: '2012-01-01 00:00:00') }
+
+    around(:each) do |spec|
+      I18n.locale = :de
+      spec.run
+      I18n.locale = :en
+    end
+
+    context 'using datepickers' do
+      it "correctly open the datepicker and changes the date field" do
```
sauloperez: should be `opens` right?

spec/features/admin/order_cycles_spec.rb
```diff
@@ -58,6 +58,39 @@
...
+        login_to_admin_section
+        visit admin_order_cycles_path
+
+        within("tr.order-cycle-#{oc_de.id}") do
+          expect(all('input').first.value).to start_with '2012-01-01 00:00:00'
```
sauloperez: please, use `find(".my-selector", match: :first)` instead of `all('input').first`. The former will wait for at least one element matching selector to appear on the page, whereas the latter won't. This is a source of brittle tests.

daniellemoorhead:
Add the aus review label when it's ready for us to push through :-)

spec/features/admin/order_cycles_spec.rb
```diff
@@ -85,7 +85,7 @@
         end
 
         within("tr.order-cycle-#{oc_de.id}") do
-          expect(all('input').first.value).to eq '2012-01-30 00:00'
+          expect(find('input.datetimepicker', match: :first).value).to eq '2012-01-30 00:00'
```
sauloperez: :clap:

daniellemoorhead:
Hey @lin-d-hop do you want to stage and test this in the UK? Given that @sauloperez and @enricostano have looked at this I don't think the AU devs need to get involved, but Sally is going away for 2 weeks so it'll be delayed till she gets back.

daniellemoorhead:
@lin-d-hop still waiting on a response to the question above about testing it in the UK rather than in Aus. Sally is back now so we can do it, just need to know either way.

Am taking the aus review label off, pop it back on if you want us to do it :)

daniellemoorhead:
Have got this on @sstead's list to test. Let's get it done and merged.

lin-d-hop:
Thanks @daniellemoorhead! Sorry about the backlog on the UK side.

myriamboure:
Staged on https://staging1.openfood.com.au

sstead:
Testing notes

- The icons are showing and the date is consistently in the format yyyy-mm-dd 

Good to go 

https://docs.google.com/document/d/16V0EgUei3NsXbLCkEfJhDtwUGHSJ6R2nGIIPGDKOKPc/edit?usp=sharing

PR #1690, opened 2017-07-14
Email confirmation: confirmable user

ltrls:
This is the first part of [email confirmation stuff](https://github.com/openfoodfoundation/openfoodnetwork/projects/8). This is mostly for @Matt-Yorkley to work on top of that and to be reviewed. I used code from enterprise confirmation as much as possible. I'll probably comment on some parts on Slack #email-confirm channel.


app/models/spree/user_decorator.rb
```diff
@@ -18,10 +18,17 @@
...
 
   attr_accessible :enterprise_ids, :enterprise_roles_attributes, :enterprise_limit, :bill_address_attributes, :ship_address_attributes
-  after_create :send_signup_confirmation
 
   validate :limit_owned_enterprises
 
+  # We use the same options as Spree and add :confirmable
+  devise :database_authenticatable, :token_authenticatable, :registerable, :recoverable,
+    :rememberable, :trackable, :validatable, :confirmable, :encryptable,
+    :reconfirmable => true, confirmation_keys: [ :id, :email ], :encryptor => 'authlogic_sha512'
```
enricostano: What if this change in future Spree versions?

enricostano: For instance https://github.com/spree/spree_auth_devise/blob/2-4-stable/app/models/spree/user.rb#L8

maybe we can just add `devise :confirmable`, would that make sense?

config/locales/fr.yml
```diff
@@ -732,11 +732,9 @@ fr:
   email_community_html: "Nous avons aussi un forum de discussion en ligne (en anglais) pour échanger avec la communauté sur des questions liées au logiciel OFN et aux défis de la gestion d'un food hub.  Nous vous invitons à y participer. Nous sommes en constante évolution et vos contributions à ce forum vont façonner les prochaines étapes. %{link}"
   join_community: "Rejoindre la communauté"
   email_help: "En cas de difficulté, consultez notre FAQ, parcourez le forum (en anglais) ou postez un message dans la section 'Support' et quelqu'un viendra vous aider!"
-  email_confirmation_greeting: "Bonjour %{contact}!"
-  email_confirmation_profile_created: "Le profil pour %{name} a été créé avec succès! Pour activer votre Profil nous devons vérifier cette adresse email."
+  email_confirmation_activate_account: "Pour activer votre compte, vous devez confirmer votre adresse email."
```
enricostano: We should only commit locales changes in `en.yml` because this will generate merge conflicts with the Transifex process

enricostano:
Do you have any place where we can look at the architecture / strategy proposed?

app/models/spree/user_decorator.rb
ltrls: I didn't know that was possible! let me try that

config/locales/fr.yml
ltrls: oh yes I mainly wrote those because my dev env was in french, I'll unstage the file

ltrls:
I guess you're looking for [these issues](https://github.com/openfoodfoundation/openfoodnetwork/projects/8)

enricostano:
> I guess you're looking for these issues

@ltrls thanks! would be great to have links to that and to the specific issue (if any) in the PR description.

app/controllers/user_confirmations_controller.rb
```diff
@@ -0,0 +1,37 @@
+class UserConfirmationsController < DeviseController
+  include Spree::Core::ControllerHelpers::Auth # Needed for access to current_ability, so we can authorize! actions
+
+  # GET /resource/confirmation/new
```
sauloperez: Why do we need the `resource` namespacing?

db/migrate/20170710145821_add_confirmable_to_user.rb
```diff
@@ -0,0 +1,13 @@
+class AddConfirmableToUser < ActiveRecord::Migration
+  def up
+    add_column :spree_users, :confirmation_token, :string
+    add_column :spree_users, :confirmed_at, :datetime
+    add_column :spree_users, :confirmation_sent_at, :datetime
```
sauloperez: These `confirm_` look like a code smell. I'd rather create the `confirmations` table and them in there withour the prefix.

ltrls: they're the ones used by Devise! I may have to hack it in order to do that

enricostano: I suggest to stick to the Devise way

sauloperez: :disappointed:

app/controllers/user_confirmations_controller.rb
enricostano: Yeah the expliciter de better 😂 . I would also like to know what kind of customization we need on top of the already customized Spree/Devise

sauloperez: Yes, all this customization pyramid it's rather scary...

config/locales/fr.yml
Matt-Yorkley: @enricostano Can we not add translation keys outside of Transifex? I thought it was ok...?

app/controllers/user_confirmations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  def new
+    build_resource({})
+  end
+
+  # POST /resource/confirmation
```
Matt-Yorkley: I like the annotations with the routes 👍

Matt-Yorkley:
So it looks like the specs are failing on every single test which logs in as admin, which makes sense given the changes. Maybe we'll have to add something to the spec helpers for `#login_as`, `#login_as_admin`, and `#quick_login_as_admin` to set the default test users to "confirmed".

config/locales/fr.yml
enricostano: Is my understanding that we shouldn't since it will generate merge conflict with the `transifex` branch and difficult @mkllnk 's job.

enricostano:
> Maybe we'll have to add something to the spec helpers for #login_as, #login_as_admin, and #quick_login_as_admin to set the default test users to "confirmed".

what about the factory instead of a spec helper?

app/models/spree/user_decorator.rb
Matt-Yorkley: Do we need to use the `:trackable` flag? That's for logging IP addresses and number of sign-ins right?

Matt-Yorkley: We're getting errors form duplicate validation messages on the login form, which is caused by `:validatable` adding form validations on top of our existing ones. 

See here: https://stackoverflow.com/questions/13897844/how-to-remove-duplicate-validation-errors#13897948

Matt-Yorkley:
It's working well in my dev environment, except I've noticed the `confirm password` field on the login form has stopped working, and now accepts an empty field.

app/controllers/user_confirmations_controller.rb
```diff
@@ -13,10 +13,10 @@ def create
 
     self.resource = resource_class.send_confirmation_instructions(resource_params)
 
-    if successfully_sent?(resource)
-      set_flash_message(:success, :confirmation_sent) if is_navigational_format?
-    else
-      set_flash_message(:error, :confirmation_not_sent) if is_navigational_format?
+    if successfully_sent?(resource) && is_navigational_format?
+      set_flash_message(:success, :confirmation_sent)
+    elsif is_navigational_format?
```
sauloperez: I think it would be easier to grasp if you place the `if is_navigation_format?` wrapping it all; both `set_flash_message` calls.

app/controllers/user_confirmations_controller.rb
```diff
@@ -26,10 +26,10 @@ def create
   def show
     self.resource = resource_class.confirm_by_token(params[:confirmation_token])
 
-    if resource.errors.empty?
-      set_flash_message(:success, :confirmed) if is_navigational_format?
-    else
-      set_flash_message(:error, :not_confirmed) if is_navigational_format?
+    if resource.errors.empty? && is_navigational_format?
+      set_flash_message(:success, :confirmed)
+    elsif is_navigational_format?
```
sauloperez: same as above

spec/features/consumer/authentication_spec.rb
```diff
@@ -74,8 +74,8 @@
             expect do
               click_signup_button
               page.should have_content "Welcome! You have signed up successfully"
-            end.to enqueue_job ConfirmSignupJob
-            page.should be_logged_in_as "test@foo.com"
+            end.to enqueue_job Delayed::PerformableMethod
+            Delayed::Job.last.payload_object.method_name.should == :send_on_create_confirmation_instructions_without_delay
```
enricostano: Could you please use the most recent `expect()` syntax?

spec/models/spree/user_spec.rb
```diff
@@ -71,9 +73,18 @@
   end
 
   context "#create" do
-    it "should send a signup email" do
+    it "should send a confirmation email" do
+      expect do
+        create(:user, confirmed_at: nil)
+      end.to enqueue_job Delayed::PerformableMethod
+      Delayed::Job.last.payload_object.method_name.should == :send_on_create_confirmation_instructions_without_delay
```
enricostano: `expect()` syntax?

app/controllers/user_confirmations_controller.rb
enricostano: @Matt-Yorkley @ltrls any info about this? Thanks!

spec/features/consumer/authentication_spec.rb
ltrls: to be honest, considering my lack of recent experience in rspec testing I was taking inspiration from [similar tests](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/spec/models/enterprise_spec.rb#L43). I'll check that out, thanks for the advice!

ltrls: oh you're talking about not using `should`! Easily done! ;)

app/controllers/user_confirmations_controller.rb
ltrls: No credit here, it's from the [EnterpriseConfirmationsController](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/enterprise_confirmations_controller.rb) that I built upon ^.^

spec/features/consumer/authentication_spec.rb
sauloperez: http://www.betterspecs.org/#expect is a good example.

spec/models/spree/user_spec.rb
```diff
@@ -77,7 +77,7 @@
       expect do
         create(:user, confirmed_at: nil)
       end.to enqueue_job Delayed::PerformableMethod
-      Delayed::Job.last.payload_object.method_name.should == :send_on_create_confirmation_instructions_without_delay
+      expect(Delayed::Job.last.payload_object.method_name).to eq(:send_on_create_confirmation_instructions_without_delay)
```
sauloperez: :clap: :clap:

app/controllers/user_confirmations_controller.rb
ltrls: here's [the controller in Devise](https://github.com/plataformatec/devise/blob/beeb48c404ae8b4c3c34f6d0a2325ca2bcc9298e/app/controllers/devise/confirmations_controller.rb), for comparison. Most notable change is the use of `authorize!`. To be honest I didn't know if Cancan would prevent from sending an email without that (that's what it's for, right?), I'll try without it to be sure. I also tried to use the original Devise controller (by having in the routes something like `:controllers => { :confirmations => 'devise/confirmations' }` but I couldn't make it work.

enricostano: Silly question, why do we need `authorize! :resend_confirmation, resource`?

oeoeaio: @ltrls My understanding of Devise (it has been a while) is that the devise controllers are used by default unless you specify `controllers` in your `devise_for` call in `routes.rb`. So if you want vanilla devise behaviour I think you should just be  able to omit the `controllers` key altogether.

app/controllers/user_confirmations_controller.rb
oeoeaio: Yeah, and I just copied and pasted this from Devise's controllers...:)

db/schema.rb
```diff
@@ -997,8 +1002,13 @@
     t.datetime "reset_password_sent_at"
     t.string   "api_key",                :limit => 40
     t.integer  "enterprise_limit",                     :default => 1, :null => false
+    t.string   "confirmation_token"
+    t.datetime "confirmed_at"
+    t.datetime "confirmation_sent_at"
+    t.string   "unconfirmed_email"
   end
 
+  add_index "spree_users", ["confirmation_token"], :name => "index_spree_users_on_confirmation_token", :unique => true
```
sauloperez: It seems reasonable that we query only this column sometimes (I guess from "cofirm your email" link type) but I'd like to see where. Indexes do not come for free, they do bring a cost on writes to the db (they must be updated), so better to have one that fits our needs perfectly.

oeoeaio: @sauloperez this index is [specified by Devise](https://github.com/plataformatec/devise/wiki/How-To:-Add-:confirmable-to-Users), it's used in exactly the context you mention: we need to do a lookup on the token to work out which user is trying to confirm their email address.

db/migrate/20170717105219_copy_confirmed_at_from_enterprises_to_spree_users.rb
```diff
@@ -0,0 +1,5 @@
+class CopyConfirmedAtFromEnterprisesToSpreeUsers < ActiveRecord::Migration
+  def up
+    execute "UPDATE spree_users SET confirmed_at = enterprises.confirmed_at FROM enterprises WHERE spree_users.email = enterprises.email AND enterprises.confirmed_at IS NOT NULL"
+  end
+end
```
Matt-Yorkley: @ltrls I'm thinking maybe we should include this update query at the bottom of the `up` method of the `AddConfirmableToUser` migration instead of independently.

Matt-Yorkley:
Superceded by #1768. Outstanding code review comments have been addressed.

Issue #1672, opened 2017-07-11
Incorrect invoice dates


Matt-Yorkley:
Orders show wrong date in UI. The order here says "03 August" when the date is actually 03-07-2017...? This happens on all orders.

Orders page:
![screenshot from 2017-07-11 13-25-13](https://user-images.githubusercontent.com/9029026/28068403-451ecd2e-663d-11e7-9544-832129255a0c.png)

Order:
![screenshot from 2017-07-11 13-25-06](https://user-images.githubusercontent.com/9029026/28068402-451dfbec-663d-11e7-9db3-780db45785af.png)

oeoeaio:
Is the date being formatted using `:long`? (Sorry on phone, can't check myself). I had a weird issue that looked like this when `:long` was used. Never worked out what was going on but using an explicit format fixed it.

Matt-Yorkley:
@oeoeaio yeah it looks like it's using a method called `pretty_time` which is using `format: :long`.

Matt-Yorkley:
Wow this is really weird... using explicit date formatting the short version `%b` gives me `Aug`, but the long version `%B` gives me `September`. They should be the same month!

oeoeaio:
Yep definitely something odd happening...could you document the steps you took to reproduce if that's easy? I'd like to have a deeper look and see if I can work out what is going on..

Matt-Yorkley:
Ah, I've found the answer: https://github.com/spree/spree/pull/2427 It looks like this might will be fixed in the Spree upgrade...

RachL:
@Matt-Yorkley @lin-d-hop is this fixed now?

RachL:
I'm guessing it is. So closing. Feel free to reopen if not.

PR #1692, opened 2017-07-14
Fixed wrong order completion dates

Matt-Yorkley:
#### What? Why?

Addressing #1672. Wrong order dates shown in edit order page.

#### What should we test?

Order dates display correctly in view/edit order. See screenshots in original issue.

### DIFF
--- a/app/overrides/spree/admin/shared/_order_tabs/replace_completed_date.html.haml.deface
+++ b/app/overrides/spree/admin/shared/_order_tabs/replace_completed_date.html.haml.deface
@@ -1,0 +1,3 @@
+/ replace_contents 'header#order_tab_summary dl.additional-info dd#date_complete'
+
+<%= l(@order.completed? ? @order.completed_at : @order.created_at, format: "%b %d, %Y %H:%M") %>


enricostano:
This is really weird, can we investigate the root problem? I'm just afraid that adding more customization on top of Spree without knowing what's happening will lead to worst problems.

enricostano:
In any case we need a test checking that we're displaying the right thing IMHO

Matt-Yorkley:
@enricostano Sure, I can add a test. It's a really strange one. I can't find any references online to Rails showing the wrong date... 

I'll take another look, it's using the i8ln method for displaying the month, so maybe it's something to do with that.

Matt-Yorkley:
@enricostano I found the source of the issue, it's a Spree 1.2 bug...

https://github.com/spree/spree/pull/2427/commits/5bb262b06013c834b7eb17cec16d8ed371900475

So I'm guessing this will be fixed in the upgrade. I think I'll to hold off on other issues until the upgrade is finished.

lin-d-hop:
Closing. Blocked by Spree Upgrade.
#gitcull2017

Issue #1696, opened 2017-07-15
Email - flaky password confirmation on signup


Matt-Yorkley:
Reference: https://github.com/openfoodfoundation/openfoodnetwork/pull/1697

Steps to replicate this bug:

- Go to Login modal -> Sign Up tab.
- Add an email and password.
- Leave "Confirm password" field blank.
- Click "Sign Up".
- Sign up is completed successfully without confirming password... 😞 

If an incorrect password is typed into the field, the validation works correctly and doesn't allow sign up when the button is clicked, and displays a message. If the text in the field is then deleted so that it is blank, the validation continues to work and shows the correct warning message.

So it seems if the field is left untouched it gets ignored completely, but after it's been touched it works correctly.

![screenshot from 2017-07-15 14-06-59](https://user-images.githubusercontent.com/9029026/28239576-23518eca-6967-11e7-84f2-7e259628b5e3.png)
![screenshot from 2017-07-15 14-07-07](https://user-images.githubusercontent.com/9029026/28239575-235181b4-6967-11e7-8857-8838a7f654c2.png)

sauloperez:
I just encountered this. I think we should raise the priority of this bug. It's not only a bit flaky in terms of UX but also gives a bad image of how we deal with security.

daniellemoorhead:
@sauloperez this should be fixed as part of the work @Matt-Yorkley and @ltrls are doing implementing email confirmation functionality.

Matt-Yorkley:
Actually @daniellemoorhead I think this one is ok to go through as a standalone bugfix. It's a small amount of code and it's not really tied up with the other email branches.

daniellemoorhead:
PR merged so closing issue

PR #1697, opened 2017-07-16
Email - fix password confirmation

Matt-Yorkley:
Addressing #1696. Password confirm field on login modal signup tab was not working when empty.

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/authentication/signup_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/authentication/signup_controller.js.coffee
@@ -1,6 +1,8 @@
 Darkswarm.controller "SignupCtrl", ($scope, $http, $window, $location, Redirections, AuthenticationService) ->
   $scope.path = "/signup"
 
+  $scope.spree_user.password_confirmation = ''
+
   $scope.errors =
     email: null
     password: null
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/darkswarm/controllers/authentication/signup_controller.js.coffee
```diff
@@ -1,9 +1,12 @@
 Darkswarm.controller "SignupCtrl", ($scope, $http, $window, $location, Redirections, AuthenticationService) ->
   $scope.path = "/signup"
 
+  $scope.spree_user.password_confirmation = ''
```
sauloperez: does this line make any difference?

Matt-Yorkley: Ok, I've taken out the other line. I think it wasn't needed.

oeoeaio:
Thanks @Matt-Yorkley, just wondering whether there is a need validate this server side? Do we care if someone can submit a request to create a user without a password confirmation? I'm not sure, but I am very surprised that this isn't handled server-side already TBH...

Matt-Yorkley:
Weirdly, it looks like it _is_ handled server-side. It's working now and I added a spec, so I think it's ok.

enricostano:
Well that's still a feature spec, right? testing the UI. What about the controller? That will tell you the truth about whats happening in the backend.

sstead:
@mkllnk finished testing and this is fixed.

Issue #1491, opened 2017-03-17
Products can be ordered beyond available stock level


OliverUK:
When two customers shop at the same time OFN lets both orders check out even if there isn't enough stock of any item to fill both orders. This then results in negative stock levels.

![screen shot 2017-03-17 at 11 28 37](https://cloud.githubusercontent.com/assets/17590921/24041787/e4cc2cdc-0b06-11e7-83cb-ff11b454a2c7.png)

I tried this on UK production by putting the same product into my basked in two different browsers (one logged in the other guest checkout) and then proceeding through checkout.

lin-d-hop:
Blocks #1031

lin-d-hop:
Notes from Rob:
If you grep for `insufficient_stock_lines` in the `orders_controller_decorator` you should find the logic. If this is not working properly, then it is just a bug, I don't think it represents a new feature.

I'm just looking now and perhaps the problem is that we only check when item quantities are updated (ie shopfront and cart pages) and a final check isn't performed at the checkout phase? Perhaps we need to redirect back to the cart from checkout in `CheckoutController#update` in that same way that we do from `OrdersController#update`

lin-d-hop:
@stveep @Matt-Yorkley might either of you have a moment to take a look at this issue in between tasks?

myriamboure:
Is it the same issue I documented here: https://github.com/openfoodfoundation/openfoodnetwork/issues/1335 ?

lin-d-hop:
Looks like it :-)

lin-d-hop:
Closed - duplicate

Matt-Yorkley:
For reference this is apparently not a duplicate of #1335, and still occurs whilst `allow_backorders` is disabled.

Matt-Yorkley:
Nope, this is definitely fixed by disabling `allow_backorders`.

lin-d-hop:
@sstead can confirm it is not.

lin-d-hop:
Here is the bug spelled out:
If I had an item in my cart, then I logged out, then the item was removed from the OC (either just that item or the supplier), then I log in and go to the shopfront- my cart hangs and I'm unable to proceed through checkout. If I instead login and click straight on the edit my cart dropdown link, I can proceed through checkout and purchase something that's no longer in the OC.

myriamboure:
@lin-d-hop I put s3 as it seems pretty few users are impacted right ?

lin-d-hop:
[...]

PR #1698, opened 2017-07-16
Oeoeaio/uk/1031 insufficient stock in cart

jeronimo:
#### What? Why?

This is extension to PR https://github.com/openfoodfoundation/openfoodnetwork/pull/1583 as issue has appeared:

1. a variant is in the user's cart
2. the user logs out
3. the variant is removed completely from the order cycle (not just stock reduced)
4. the user logs back in

Approach:
Validation method inside order model seemed more appropriate way but there were so many tests failing and fixing that looked like very time consuming thing so this approach has been chosen.

P.S. was not sure about naming on methods so please give better ones if you have any in mind.

#### What should we test?

Checkout process when incoming/outgoing products get unselected in admin order_cycles edit page.

Hopefully it is not related unless later Spree versions has Orders implemented already with double check of this functionality.

#### Dependencies

Extension of https://github.com/openfoodfoundation/openfoodnetwork/pull/1583
And follow up of https://github.com/openfoodfoundation/openfoodnetwork/pull/1477

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -61,6 +61,14 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def check_and_redirect_if_variants_not_available_in_order_cycle
+    flash.keep unless flash[:error].nil? # there are too many redirects between cart and shop, need to keep flash errors
+    unless current_order(true).variants_still_available_in_order_cycle?
+      flash[:error] = current_order.errors.full_messages.join(', ')
+      redirect_to cart_path
+    end
+  end
+
   # All render calls within the block will be performed with the specified format
   # Useful for rendering html within a JSON response, particularly if the specified
   # template or partial then goes on to render further partials without specifying
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -9,6 +9,7 @@ class CheckoutController < Spree::CheckoutController
   prepend_before_filter :require_distributor_chosen
 
   skip_before_filter :check_registration
+  before_filter :check_and_redirect_if_variants_not_available_in_order_cycle, only: %i[edit update]
 
   include OrderCyclesHelper
   include EnterprisesHelper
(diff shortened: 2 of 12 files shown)


sauloperez:
> Validation method inside order model seemed more appropriate way but there were so many tests failing and fixing that looked like very time consuming thing so this approach has been chosen.

How did validation looked like? To me the validation does seem like the way to go rather than having to pollute the application controller. What about ensuring there are no orders still to be completed  that have that variant in their line items when removing it from the order cycle? I'd then tell the admin through a flash error message.

jeronimo:
Cool! @sauloperez, @enricostano, may I assume that you will take this issue out of me, will analyse all requirements and whole situation again and will do it properly?

enricostano:
@jeronimo maybe we can discuss about it quickly on mumble, what do you think? Just to be on the same page and better understand the context. Is that OK for you?

Matt-Yorkley:
For reference, this is the latest PR for #1031.

daniellemoorhead:
>For reference, this is the latest PR for #1031.

@Matt-Yorkley should this PR be closed then?

daniellemoorhead:
Moved to September dot point release.

lin-d-hop:
Hi @enricostano 
Could you just leave a wee update on the latest?
Maybe we could have a chat and we can respec the issue to assign to someone new?
Cheers,
Lynne

oeoeaio:
For my own reference, potentially related to [this](https://github.com/oeoeaio/openfoodnetwork/commit/4443229066983f8ff07c6c0dd3b5cbd21859e237)

enricostano:
@sauloperez can we work on prioritizing this? Maybe we can help in some way.

myriamboure:
@enricostano @lin-d-hop what shall we do with that PR ? Shall we just#1031 into an icebox item, so that it gets specced properly when we prioritize it ? And for the bug #1491 shall we wait for it to be prioritized before we work back on this?

myriamboure:
Do I understand well: this one is blocked by #1767 so we can't review and merge before we have moved forward the other one which is in dev, right?

myriamboure:
Will reopen when #1491 is priorized, incepted and solve, and then we can work on #1031. So closing for now.

Issue #1306, opened 2016-12-07
Grey property tags on producer profile map modals are clickable, but do nothing


sstead:
Device: All
Where: Shops, profiles, map profile modals,
Recommendation: Make not clickable.

![image](https://cloud.githubusercontent.com/assets/10843056/20953200/ccb66a92-bc85-11e6-852f-7f7739447d2f.png)

sauloperez:
Just to let you know, we have a slack team and a dev channel where people working on the codebase chat with each other. It's a great place to get some help as you start working on stuff. Let me know if you'd like us to send through an invitation to join. We also have community message boards where things get discussed [here](https://community.openfoodnetwork.org/).

letakeane:
@sauloperez Thank you so much! I'd love to be added to the team/channel. Looking forward to diving deeper into an open source project!

kirstenalarsen:
hey Leta - can you send your email to hello@openfoodnetwork.org and we’ll add you to slack :)

sstead:
Thanks Leta!

**Testing notes**
- The map profile buttons no longer look clickable (correct)
- Filter buttons are still clickable (correct)
- However, the property and category badges in product description modals, within shops are still clickable when they shouldn't be. See below.
![image](https://user-images.githubusercontent.com/10843056/28607201-155480a6-721e-11e7-8c6a-66f3830c7ced.png)

myriamboure:
Iceboxed here: https://community.openfoodnetwork.org/t/upload-and-display-product-quality-certificates/1145
#gitcull2018

PR #1701, opened 2017-07-17
#1306: Add 'disabled' class to shopfront property selectors buttons ...

letakeane:
… to prevent clickability.

#### What? Why?

This PR closes issue #1306 -
I added a `.disabled` class to the greyed out buttons, similar to the blue buttons, to prevent clickability.

#### What should we test?

This only changes the buttons in the enterprise details section; it should not affect other buttons, but it's worth making sure others are not being "disabled".

### DIFF
--- a/app/assets/javascripts/templates/partials/enterprise_details.html.haml
+++ b/app/assets/javascripts/templates/partials/enterprise_details.html.haml
@@ -11,7 +11,7 @@
         %span.filter-shopfront.property-selectors.pad-top
           %ul.inline-block
             %li{"ng-repeat" => "property in enterprise.supplied_properties"}
-              %a.button.tiny{"ng-bind" => "property.presentation"}
+              %a.button.tiny.disabled{"ng-bind" => "property.presentation"}
 
       .about-container.pad-top
         %img.enterprise-logo{"ng-src" => "{{::enterprise.logo}}", "ng-if" => "::enterprise.logo"}


Gemfile.lock
```diff
@@ -762,7 +762,7 @@ DEPENDENCIES
   wkhtmltopdf-binary
 
 RUBY VERSION
-   ruby 2.1.5p273
+  ruby 2.1.5p273
 
 BUNDLED WITH
-   1.15.1
+  1.15.1
```
enricostano: @letakeane hi! Just one question, is this the resulting `Gemfile.lock` from `bundle install`? Did you modify it yourself? I'm just curious since Bundler generally adds 3 spaces in both `RUBY VERSION` and `BUNDLED WITH` indentation AFAIK.

letakeane: @enricostano I did modify it myself (to restore the file). I'm actually a bit in the dark as to why the file was altered in the first place; as far as I know, I didn't manually edit it. I didn't realize the change had been made until I made my commit.

I'm unfamiliar with Bundler; does it dynamically write the `Gemfile.lock` itself each time Bundler is run?

enricostano: yep when you run `bundle install` it checks for changes in `Gemfile` (where we declare our dependencies) and generates a `Gemfile.lock` to fix exactly which versions we are using.

if you don't need any change in `Gemfile` just run `git checkout Gemfile.lock` and run `bundle install` again. Please ping me if get stuck at some point 💪

letakeane:
Squashed and committed! Sorry for the delay. Looks like all the checks are completed and passed. Ready to go! :)

daniellemoorhead:
@sstead can you coordinate with @mkllnk to get this on staging and tested next week (Wed 26th)?

mkllnk:
@enricostano @Matt-Yorkley: What do you guys think of coordinating testing in the UK and merge this to master? Just if you have time before Wednesday.

mkllnk:
Thank you for your contribution Leta. If you would like to add a bit more, the original issue is not resolved completely. See https://github.com/openfoodfoundation/openfoodnetwork/issues/1306#issuecomment-317959151.

daniellemoorhead:
Hey @mkllnk I looked at the issue and the comments here as well but can't see what is remaining to be done other than maybe the issue @sstead found - is that what still needs to be done? And if this is the case should we re-open the issue?

mkllnk:
Yes, that still needs to be done. So I re-opened the issue. Alternatively, we could open a new issue to make the task clearer for a new-comer.

Issue #1311, opened 2016-12-07
Reduce padding at top of page (above heading) to maximise screen height


sstead:
Devices: Mobile/Tablet
Where: Producers
![image](https://cloud.githubusercontent.com/assets/10843056/20953607/0ac7e916-bc89-11e6-961b-24075de2a027.png)

lauraturk:
Afternoon! I’m a first-time contributor and was hoping to help out with this issue. If there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

daniellemoorhead:
@mkllnk or @enricostano or @Matt-Yorkley could one of you help out @lauraturk with setting up and getting onboard with the OFN codebase?

@lauraturk welcome! It's at a crappy time of the early morning for you but we have a regular dev catchup this Thursday (which makes it 5:30pm Melbourne time and 8:30am UK time). You're very welcome to join us if you're still awake! Also, have you joined our slack team yet?

Matt-Yorkley:
Hi @lauraturk, I don't think anyone's working on this at the moment. If you want to set up a full dev environment there's guides for different OS's in our wiki: https://github.com/openfoodfoundation/openfoodnetwork/wiki

lauraturk:
@Matt-Yorkley and @daniellemoorhead thanks! I'm all set up according to the wiki and Contributing doc. Great documentation, thanks!

There's a small cadre of us diving into our first contributions to open source for a school project. You'll see us popping up the next couple of days to hopefully help with some of the easier things!

kirstenalarsen:
Hey Laura, welcome and if you send your email to hello@openfoodnetwork.org I’ll add you to Slack. Perhaps send me the emails of all the students so i can do it in one go?

sstead:
The padding has been reduced since Yuko first picked up this issue. 
![image](https://user-images.githubusercontent.com/10843056/46839692-e0202d00-ce0a-11e8-963e-324b9b164682.png)
I'm no designer, so I'll leave this for someone else to decide whether the padding could be further improved.

myriamboure:
@kirstenalarsen same here, I'm closing that issue as will definitely be impacted by mobile UX overhaul given Yuko new designs. Reopen if you disagree.

PR #1704, opened 2017-07-18
Add media query to panepadding mixin

lauraturk:
#### What? Why?

Fixes the heading padding at mobile sizes to maximize screen height. Closes issue #1311 

I created a media query for the mixin panepadding, that changes padding-top @ 640px from 100px to 25px.

Producers page
padding-top @ 25px

Also impacts footer:
padding-top @ 25px

#### What should we test?

Visual check that the change to panepadding did not negatively impact other parts of the app.

### DIFF
--- a/app/assets/stylesheets/darkswarm/mixins.scss
+++ b/app/assets/stylesheets/darkswarm/mixins.scss
@@ -15,6 +15,10 @@
 @mixin panepadding {
   padding-top: 100px;
   padding-bottom: 100px;
+
+  @media all and (max-width: 640px) {
+    padding-top: 25px;
+  }
 }
 
 @mixin paneWhiteText {


Gemfile.lock
```diff
@@ -760,9 +760,3 @@ DEPENDENCIES
   whenever
   wicked_pdf
   wkhtmltopdf-binary
-
-RUBY VERSION
-   ruby 2.1.5p273
-
-BUNDLED WITH
-   1.15.1
```
sauloperez: Hi @lauraturk I'm very excited to have you on board!

This is quite similar to what happened to @letakeane in https://github.com/openfoodfoundation/openfoodnetwork/pull/1701#discussion_r127962621.

If you didn't edit the file yourself there's something weird in your setup :thinking: First, I would suggest you do what @enricostano said to @letakeane. Run:

```shell
git reset @^ # Undo last commit leaving unstaged changes
git checkout Gemfile.lock # Remove changes in the Gemfile.lock
git add -p # Stage your changes chunk by chunk ensuring you don't touch the Gemfile.lock
git commit # Well yes, let's commit

# As you modified the last commit you will have to force the push
# (avoid it when collaborating on a branch. More details in
# https://developer.atlassian.com/blog/2015/04/force-with-lease/)
git push --force-with-lease 
```

Then, it would be good to know which versions of bundler and rubygems you have installed. What do the following commands return for you?

```shell
bundler -v
gem --version
```

lauraturk: Wonderful! Thank you for this information! I'm using Bundler version 1.7.6 and gem version 2.4.3,  though I am using ruby 2.1.5. New to Ruby world, so I hope that helps!

lauraturk:
I re-committed with out gemfile.lock (not sure why it changed, I did not open it!) Thanks @sauloperez and @enricostano for the clear git commands!

oeoeaio:
Hey @lauraturk, welcome to the project. :)

Looking at the commit history for this PR (click on the `Commits` tab near the top of this page), it looks like you've push two commits. The first one (63c220ab0e78968c75d095a8e603b759ea7f8aa3) is great and contains the media query change, but the second one (69c300f83aaff300920c07a9a760f1f53b2f3193) still seems to contain the changes to `Gemfile.lock`.

This file gets automatically updated when you run `bundle install` so that is why there are changes that you don't remember making. The problem is this instance is that the version of Bundler you are using is fairly outdated, and it is changing the `Gemfile.lock` in a way that is different to the way other developer's versions of bundler would.

I would suggest doing two things.

1) Remove 69c300f83aaff300920c07a9a760f1f53b2f3193
There are lots of ways to do this, but this is probably the simplest
````bash
git status # Make sure there are no un-committed files and that you are on your master branch
git reset @^ # As above: undo last commit leaving unstaged changes
git checkout . # Removes unstaged changes. ie. the changes to Gemfile.lock
git push --force-with-lease # As per @sauloperez's comment above
````

Or, this will do the same thing:
````bash
git status # Make sure there are no un-committed files and that you are on your master branch
git reset --hard 63c220ab # This resets your commit history for the branch to this commit
git push --force-with-lease # As per @sauloperez's comment above
````

2) Update your version of bundler
`gem install bundler` - Should give you something like version `1.15.2`

Thanks again! Ping any of us if you have any questions.

mkllnk:
@enricostano @Matt-Yorkley: What do you guys think of coordinating testing in the UK and merge this to master? Just if you have time before Wednesday. If you do it, please remove the change to Gemfile.lock and rebase it on the current master before merging.

sstead:
Thanks Laura!
**Test notes:**
- Looks good! It has shaved some excessive space from the producers and shops page, as well as some segments on the homepage, and the footer. Improves the use of space on mobile view.

Issue #946, opened 2016-05-04
User email verification


daniellemoorhead:
At the moment users do not have to verify/confirm their email address to become an OFN user, we want to change this so that they do need to go through a confirmation process to ensure that they are using their own email address.

Check this google doc to see the overarching flows for initial sign up and then for unconfirmed users trying to login: https://docs.google.com/document/d/1S9CDFQZZr8BLeMfzgDQAHM8ZpPGOc_WjC1vgNm-T23M/edit

Question: Should unconfirmed user login issues be done first, followed by the sign up issues?

mkllnk:
This issue is evaluating #833.

- [ ] Create an initial overview and estimates (1)
- [ ] Change the user model to have confirmable email addresses (2)
- [ ] Change user registration process to prompt for email validation (4)
- [ ] Change enterprise user registration to prompt for email validation (3)
- [ ] Change Discourse login to check for confirmed user email address (2)
- [ ] Replace `enterprise.email` by `enterprise.contact` which is a user (3)
- [ ] Prevent guests from using a registered email address (4)
- [ ] Associate guest orders of same email address after validation, #917 and #907 (2)

Estimate: 21 hours

@sstead: Does this change make #948 obsolete?

sstead:
Yes, i think so. I don't think a user who has >1 enterprise permission needs to validate every enterprise they create anymore. Also with the new 'contact only' email field, there's not the problem of disengaged OFN users needing to confirm an email just for it to show on a profile. These were the main reasons why we once wanted to be able to manually verify emails. I think the plan above, of requiring each user to confirm an email is less likely to create situations were people don't verify their email, and superadmin would need to do it for them.

myriamboure:
@daniellemoorhead to move forward on the email validation, I have started to write down how I see that functionnaly given all the inputs and comments in various discussions on community and Github. I put it here on Discourse https://community.openfoodnetwork.org/t/confirmation-emails-registration-flow/237/3?u=myriamboure as we are still on the functional level, I propose facilitate the discussion there first, and when we agree on the process then move back here to specify the tech part (which I might not be super useful for...)
Ping @ltrls if you want to have a look at what we are talking about... when we get the feedback from the community we can see how we can break this down maybe.

daniellemoorhead:
[...]

PR #1705, opened 2017-07-19
Remove enterprise confirmations

Matt-Yorkley:
Removing enterprise confirmations for #946. I think there isn't a specific issue for this, but it's mentioned somewhere in a list of "issues that need to be created".

There were a lot of references to the enterprise confirmations throughout the codebase, so this one might need some fine-tuning. 

Shouldn't be merged before #1690 is finished and merged.


db/schema.rb
```diff
@@ -123,8 +123,8 @@
   create_table "distributors_shipping_methods", :force => true do |t|
     t.integer  "distributor_id"
     t.integer  "shipping_method_id"
-    t.datetime "created_at",         :null => false
-    t.datetime "updated_at",         :null => false
+    t.datetime "created_at"
+    t.datetime "updated_at"
```
enricostano: Why this change?

db/schema.rb
```diff
@@ -997,8 +992,13 @@
     t.datetime "reset_password_sent_at"
     t.string   "api_key",                :limit => 40
     t.integer  "enterprise_limit",                     :default => 1, :null => false
+    t.string   "confirmation_token"
+    t.datetime "confirmed_at"
+    t.datetime "confirmation_sent_at"
+    t.string   "unconfirmed_email"
```
enricostano: Are this coming from another PR?

db/schema.rb
```diff
@@ -1089,6 +1090,7 @@
     t.string   "sku"
     t.boolean  "on_demand"
     t.datetime "permission_revoked_at"
+    t.datetime "import_date"
```
enricostano: ?

db/schema.rb
Matt-Yorkley: That's really weird.  I just added a single migration to remove the enterprise confirmation columns and ran `rake db:migrate`. Maybe I need to fetch from upstream and rebase...

lib/open_food_network/users_and_enterprises_report.rb
```diff
@@ -63,14 +65,25 @@ def users_and_enterprises
     end
 
     def sort(results)
+      pp results
```
enricostano: ⚠️

Matt-Yorkley: Yeah I added that in for debugging... still working on the last few failing specs.

lib/open_food_network/users_and_enterprises_report.rb
```diff
@@ -63,14 +65,25 @@ def users_and_enterprises
...
     end
 
     def to_bool(value)
       ActiveRecord::ConnectionAdapters::Column.value_to_boolean(value)
     end
+
+    def to_local_datetime(string)
+      return "" if string.nil?
+      string.to_datetime.in_time_zone.strftime "%Y-%m-%d %H:%M"
+    end
```
enricostano: Can we use a more descriptive var name than `string`?

Matt-Yorkley: That was what the original method looked like in `master`, I was just putting it back in. I'll change it now...

db/schema.rb
oeoeaio: @Matt-Yorkley If your dev db contained this migration at the time you ran rake `db:migrate`, I think these would have been added to your schema? I could be wrong but I seem to remember this happening to me before. The way schema.rb gets updated is still a bit of a black box to me...

oeoeaio:
I feel like the transfer of responsibility for checking confirmation needs to switch from `enterprise` to `user` simultaneously, but it is impossible to do that until #1690 is merged though right? I'm just concerned that by removing all of the responsibility from `enterprise` in one step (this one) and then trying to remember to add all of that responsibility back to `user` at a later point is a recipe for trouble. I suppose we just have to be super systematic about it...

app/models/enterprise.rb
```diff
@@ -198,10 +188,6 @@ def self.distinct_count
     count(distinct: true)
   end
 
-  def activated?
-    confirmed_at.present? && sells != 'unspecified'
```
oeoeaio: I think we still need to concept of `activated`? It behaves like a check on the enterprise to determine whether the registration process is complete, if anything I think this should just become:

````ruby
owner.confirmed? && sells != 'unspecified'
````

File: app/models/enterprise_relationship.rb [hunk 7]
```diff
@@ -30,7 +30,7 @@ class EnterpriseRelationship < ActiveRecord::Base
   # Load an array of the relatives of each enterprise (ie. any enterprise related to it in
   # either direction). This array is split into distributors and producers, and has the format:
   # {enterprise_id => {distributors: [id, ...], producers: [id, ...]} }
-  def self.relatives(activated_only=false)
+  def self.relatives
```
oeoeaio: Ditto with `activated`

app/models/enterprise.rb
Matt-Yorkley: @oeoeaio I considered that as well, but won't all owners be confirmed if they have to confirm their email before registering an enterprise or logging in to the backend?

Matt-Yorkley: Is it even possible to create an enterprise without specifying the `sells` attribute?

Matt-Yorkley:
I'll move this PR into another branch to make additional changes at the same time.

app/models/enterprise.rb
oeoeaio: @Matt-Yorkley, owners of all *new* enterprises will be confirmed, but we have to handle legacy data unfortunately...

oeoeaio:
For reference: superseded by #1720

PR #1706, opened 2017-07-20
Remove zeus and guard-zeus from the bundle

oeoeaio:
#### What? Why?

Having guard-zeus specified as a dependency is problematic because it also pulls in the zeus gem as a dependency. This makes it impossible to update or switch zeus versions locally without affecting other developers.

The maintainers of zeus say that it is designed to be installed and run outside of the bundle: https://github.com/burke/zeus

#### What should we test?

I suppose it would be good if someone could pull this down and check that is doesn't break anything for them. Also, if anyone is using guard we might need to come up with a different idea.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -126,7 +126,6 @@ group :development do
   gem 'guard-livereload'
   gem 'rack-livereload'
   gem 'guard-rails'
-  gem 'guard-zeus'
   gem 'guard-rspec'
   gem 'parallel_tests'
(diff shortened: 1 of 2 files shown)


oeoeaio:
Cool, I'm using RVM and was able to just do:

````
gem uninstall zeus  # remove all
gem install zeus # installed 0.15.14
zeus start
````

Though I should have done a `bundle clean` first because now I probably have an orphaned copy `guard-zeus`.

PR #1707, opened 2017-07-20
Exclude session data when mirroring data

oeoeaio:
#### What? Why?

What it says on the tin: exclude session data when mirroring. This is particularly useful when there is a lot of session data on the server, and you are on a slow connection.

Probably a better long term solution would be to add a job to clean up old session data on the server, but this is a quick & dirty fix.

#### What should we test?

I have tested it, but feel free to give a try to verify

### DIFF
--- a/script/mirror_db.sh
+++ b/script/mirror_db.sh
@@ -18,12 +18,13 @@ else
   DB_DATABASE='openfoodnetwork'
 fi
 
+DB_OPTIONS='--exclude-table-data=sessions'
 
 # -- Mirror database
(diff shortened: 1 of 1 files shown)


enricostano:
@oeoeaio do you have any documentation about how, when and in which process do you use this script? Thanks!

oeoeaio:
Nope. I will add some documentation to the top of the file. 🙂

oeoeaio:
Bump @enricostano (I know you are away, for when you get back).

Issue #1663, opened 2017-07-06
Variants are ordered illogically (not ordered at all) on purchase order emails


OliverUK:
There doesn't seem to be any logic behind how variants are ordered on emails sent to suppliers. 
You can end up with a very long order email to a supplier where, for example, Apricots 500g appears towards the top and Apricots 250g way down and it's up to the supplier to find them, add them up and make sure they've not missed anything.

sauloperez:
I guess this is again just a matter of applying alphabetic order, in this case in the underlying database query

OliverUK:
Ordering seems to work well now in the linked issue 1604 on UK staging. Could the same be applied to the order emails? Sort alphabetically first and the by qty in ascending order?

Duende13:
I'm working on that @OliverUK :)

Duende13:
@OliverUK , @lin-d-hop , just one question, does it make sense that this order can be seen into the confirmation screen at the end of the check out? The summary of the order screen. It can be applied to both, email and confirmation screen. Is it ok for you? I'm almost there just I have to add some tests and I'll be done :)

OliverUK:
Hi Susana, 
You are saying that the customer orders will also be ordered in the same way as the orders to the suppliers? I can see no problem with this.

lin-d-hop:
![](https://github.trello.services/images/mini-trello-icon.png) [Ordering of Variants and Shipping Methods](https://trello.com/c/l9oQNVZa/52-ordering-of-variants-and-shipping-methods)

lin-d-hop:
@Duende13 
This has now been deployed to UK production but seems to not be working.
Ping @OliverUK to provide more details....

OliverUK:
OK so here is an example of this mornings notify producer emails as copied to me:

![screen shot 2017-10-12 at 10 48 40](https://user-images.githubusercontent.com/17590921/31490175-fbb40c2a-af3a-11e7-97fa-b22c2b24ccd7.png)

OliverUK:
@Duende13 any news on what the problem can be?

Duende13:
Hi @OliverUK ,
I was having a look into that and I realised that the changes that I made were not merged into the current master branch. @lin-d-hop and I will look into that and we'll fix it. Thank you for your understanding :)

Duende13:
Hi @OliverUK ,
[...]

PR #1709, opened 2017-07-20
Sort line items by name and unit_value

Duende13:
for confirmation email and summary screen after checkout. It was done in the cart and now in emails.
Requested in issue #1663 

#### What? Why?

#### What should we test?

### DIFF
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -25,6 +25,13 @@ Spree::LineItem.class_eval do
     end
   }
 
+  scope :sorted_by_name_and_unit_value,     
+      # Find line items that are from order sorted by variant name and unit value
+      joins(:variant=> :product).
+        reorder('spree_products.name asc, spree_variants.unit_value asc').
+        select('spree_line_items.*')
+  
+
   scope :supplied_by, lambda { |enterprise|
     joins(:product).
       where('spree_products.supplier_id = ?', enterprise)
@@ -85,7 +92,12 @@ Spree::LineItem.class_eval do
   end
 
   def display_name
-    variant.display_name
+    if variant.display_name.blank?
+print 'jejeje'
+      return variant.name
+    else
+      return variant.display_name
+    end
   end
 
   def unit_value
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -26,7 +26,7 @@ Spree::LineItem.class_eval do
   }
 
   # Find line items that are from order sorted by variant name and unit value
-  scope :sorted_by_name_and_unit_value, joins(:variant => :product).
+  scope :sorted_by_name_and_unit_value, joins(variant: :product).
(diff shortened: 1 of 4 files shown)


app/models/spree/line_item_decorator.rb
```diff
@@ -25,6 +25,11 @@
     end
   }
 
+  # Find line items that are from order sorted by variant name and unit value
+  scope :sorted_by_name_and_unit_value, joins(:variant => :product).
+    reorder('spree_products.name asc, spree_variants.unit_value asc').
+    select('spree_line_items.*')
+
```
enricostano: Could you please share the generated SQL query? Should be as easy as running the following in the Rails console:
```
@order.line_items.sorted_by_name_and_unit_value.to_sql
```

spec/models/spree/line_item_spec.rb
```diff
@@ -40,6 +52,10 @@ module Spree
           LineItem.without_tax.should == [li2]
         end
       end
+
+      it "finds line items sorted by name and unit_value" do
+        o.line_items.sorted_by_name_and_unit_value.should == [li5,li6,li4,li3]
```
enricostano: Could you please adopt the more recent RSpec syntax using `expect()` instead of `should()`?

`expect(o.line_items.sorted_by_name_and_unit_value).to eq([li5,li6,li4,li3])` or something like that

app/models/spree/line_item_decorator.rb
Duende13: @enricostano , 
 "SELECT spree_line_items.* FROM \"spree_line_items\" INNER JOIN \"spree_variants\" ON \"spree_variants\".\"id\" = \"spree_line_items\".\"variant_id\" INNER JOIN \"spree_products\" ON \"spree_products\".\"id\" = \"spree_variants\".\"product_id\" WHERE \"spree_line_items\".\"order_id\" = 204331 ORDER BY spree_products.name asc, spree_variants.unit_value asc"

spec/models/spree/line_item_spec.rb
Duende13: Done, @enricostano !

app/models/spree/line_item_decorator.rb
enricostano: cc/ @sauloperez

app/models/spree/line_item_decorator.rb
```diff
@@ -25,6 +25,11 @@
     end
   }
 
+  # Find line items that are from order sorted by variant name and unit value
+  scope :sorted_by_name_and_unit_value, joins(:variant => :product).
```
oeoeaio: In general, I would prefer to see this syntax:

````ruby
variant: :product

Duende13: Change done, @oeoeaio  and @enricostano !

daniellemoorhead:
@lin-d-hop will leave it to you to test in the UK and then give us the nod when it's good to go.

PR #1713, opened 2017-07-21
Upgrade bundler to locked version

mkllnk:
#### What? Why?

Developers have different Bundler versions installed and Travis uses a much older version than most of the developers. When @sauloperez encountered different test results locally and in Travis, suggested PR #1641. Instead of updating to the latest version of Bundler, this pull request provides a script to install the version specified in `Gemfile.lock`. This change proposes to use it in Travis. Developers can also benefit from using it to get the right version.

#### What should we test?

Only Travis is affected. No user testing on a staging server required. We could think about best practices for deployment.

Not at all.

### DIFF
--- a/script/upgrade_bundler.sh
+++ b/script/upgrade_bundler.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+# This shell script looks for the last used Bundler version in Gemfile.lock and
+# installs exactly that version, removing other versions.
+
+# Fail if a single command fails.
+set -e
+
+# `grep -m 1`: find the first occurrence of "BUNDLED WITH"
+# `-A`:        print the next line after "BUNDLED WITH" as well
+# `-x -F`:     find exactly that string without interpreting regex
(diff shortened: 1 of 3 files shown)


script/upgrade_bundler.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -e
```
mkllnk: Maybe I should add some comments here to explain the script.

This line makes the script exit whenever a command is not successful.

script/upgrade_bundler.sh
```diff
@@ -0,0 +1,20 @@
...
+
+version="$(grep -m 1 -A 1 -x -F "BUNDLED WITH" Gemfile.lock | tail -n 1  | tr -d '[:space:]')"
```
mkllnk: `-m 1`: find the first occurrence
`-A `: print the next line as well
`-x -F`: find exactly that string without interpreting regex
`tail -n 1`: take the last line
`tr -d`: delete all spaces

PR #1735, opened 2017-07-28
Untranslate report headers

oeoeaio:
#### What? Why?

A couple of our reports are designed to be used to integrate with other pieces of software (Xero and Lettuceshare), so translations of these headers will break the integration.

#### What should we test?

That Xero and LettuceShare integrations still work (in particular that the headers generated by the reports are acceptable).


sstead:
Tested, Xero report is able to be uploaded into Xero again.

PR #1737, opened 2017-07-28
Fix some display errors

mkllnk:
#### What? Why?

Recent i18n updates broke a couple of views.

#### What should we test?

The "+ 3 More" display in shop fronts. The standard view of printed order invoices.

You can switch the invoice display as admin in Settings/Invoice Settings.

A general look around if anything else is broken would be good as well.


sstead:
Both invoice layouts look good!
image

But... they both have an issue on the second page if it's a long order. This is pre-existing. I've created a new issue here - https://github.com/openfoodfoundation/openfoodnetwork/issues/1738
image

The +3more buttons look god!
image

PR #1742, opened 2017-07-29
Add a helper to allow set/reset of Spree::Config settings in specs

oeoeaio:
#### What? Why?

There is a requirement to alter `Spree::Config` settings in the test suite, to allow tests to check the outcome of logic in different contexts. At the moment, we are mostly doing this using something like:

````ruby
before do
  Spree::Config.set some_setting: "some_value"
end
````

The problem with this approach is that the values of such settings appear to be cached, and they are not rolled back with the transaction when the spec has finished running. This results in the test context bleeding out and influencing other specs in the test suite that are completely unrelated. Such behaviour is particularly problematic on Travis-CI, because we run tests in parallel in 5 separate threads, so if a particular spec happens to be moved from one thread to another, this can cause it to be run in a different context, sometimes resulting is seemingly random failures.

The idea to solve this problem using stubbing was raised, but the problem with this idea is that Spree::Config settings are accessed in the codebase using both of these syntaxes (a lot of this is in the Spree repo, so we don't have control of all of it):

````ruby
Spree::Config.setting_name
Spree::Config[:setting_name]
````

Stubbing both of these before each spec looks messy and stubbing the `:[]` method is problematic anyway, because it would depend on specifying a two stubs: one with `.and_call_original` and one with a `with(:setting_name)` clause.

So the approach I have used is to create a helper with a method `with_spree_settings`, which is to be called from an `around` block, and which accepts a hash of settings that will be set and reset to their original values for each relevant spec.

If others are happy with this approach I am happy to take on the work of implementing the change throughout the rest of the test suite. Should only be half an hour or so of work.

#### What should we test?

These changes and any similar future rollout will not touch any functional part of the codebase, they will only touch the specs.

#### Related PRs

There is a fair bit of discussion [on #1550](https://github.com/openfoodfoundation/openfoodnetwork/pull/1550#pullrequestreview-46995367) about this, which prompted this PR.

### DIFF
--- a/spec/support/spree/config_helper.rb
+++ b/spec/support/spree/config_helper.rb
@@ -1,0 +1,18 @@
+module Spree
+  module ConfigHelper
+    def with_spree_config(hash)
+      stored = config_slice(hash.keys)
+      Config.set(hash)
+      yield
+      Config.set(stored)
+    end
+
+    private
+
+    def config_slice(keys)
+      keys.each_with_object({}) do |key, hash|
+        hash[key] = Config[key]
+      end
+    end
+  end
+end
--- a/spec/models/spree/order_populator_spec.rb
+++ b/spec/models/spree/order_populator_spec.rb
@@ -196,7 +196,9 @@ module Spree
       let(:v) { double(:variant, on_hand: 10) }
 
       context "when backorders are not allowed" do
-        before { Spree::Config.allow_backorders = false }
+        around(:each) do |example|
+          with_spree_config({ allow_backorders: false }, &example)
+        end
 
         context "when max_quantity is not provided" do
           it "returns full amount when available" do
@@ -220,10 +222,8 @@ module Spree
       end
 
       context "when backorders are allowed" do
-        around do |example|
-          Spree::Config.allow_backorders = true
-          example.run
-          Spree::Config.allow_backorders = false
+        around(:each) do |example|
+          with_spree_config({ allow_backorders: true }, &example)
         end
 
         it "does not limit quantity" do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -98,6 +98,7 @@ RSpec.configure do |config|
   config.include Spree::UrlHelpers
   config.include Spree::CheckoutHelpers
   config.include Spree::MoneyHelper
+  config.include Spree::ConfigHelper
   config.include Spree::Core::TestingSupport::ControllerRequests, :type => :controller
   config.include Devise::TestHelpers, :type => :controller
   config.extend  Spree::Api::TestingSupport::Setup, :type => :controller

### CODE REVIEW

File: spec/support/spree/config_helper.rb [hunk 1]
```diff
@@ -0,0 +1,18 @@
+module Spree
+  module ConfigHelper
+    def with_spree_config(hash)
+      stored = config_slice(hash.keys)
+      Config.set(hash)
+      yield
+      Config.set(stored)
```
sauloperez: With this line, Would we restore only the value of the keys we modified earlier or does `.set` behave like a merge? It looks like the former to me and this could affect some tests.

oeoeaio: My understanding is that `.set` just changes the value of the settings corresponding to the keys provided. So I think the code already does what you want it to do. At least, that was my intention. Perhaps I need some specs...

oeoeaio: `stored` is using `config_slice` to store only the keys passed into `with_spree_config`. At the end of the spec, it sets just those keys back to their original values.

sauloperez: I just wanted to make sure the following is true:

> My understanding is that .set just changes the value of the settings corresponding to the keys provided.

If instead all the other config keys are empty we have a problem

oeoeaio: Yeah cool. 👍 I am fairly sure that any unmentioned config keys are left untouched, but I'll double check.

enricostano:
What I would do:

1. Replace all the calls in specs and code to `Spree::Config.setting_name`
2. Add a code climate grep rule to avoid future additions of this form
3. Use explicit stubs in specs

I think that this is more explicit, flexible and easy to follow. But if everybody else agree on the proposed approach I'm fine with it.

daniellemoorhead:
@oeoeaio what does this mean in relation to #1550, will that PR still be relevant/needed as a result?

sauloperez:
@daniellemoorhead this is something independent of the RSpec version. Both should PRs should be merged

daniellemoorhead:
Will leave this one in the backlog till @oeoeaio gets back, rather than having Maikel look at it.

Therefore moving this to the September dot point release.

oeoeaio:
@enricostano
>1. Replace all the calls in specs and code to Spree::Config.setting_name

As I mentioned above, a lot of these calls are made using the hash syntax from within Spree itself, and so we cannot replace them using the preferred syntax...

enricostano:
> from within Spree itself

@oeoeaio ooops sorry, I missed that part 😬 Let's try this approach then 👍

spec/support/spree/config_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    private
+
+    def config_slice(keys)
+      keys.each_with_object({}) do |key, hash|
+        hash[key] = Config[key]
```
mkllnk: So if a key doesn't exist, it gets overwritten with `nil`? Isn't that different to a non-existent key?

I'm wondering if we can replace the whole config hash with a copy instead of overwriting values. But I'm not familiar with the Spree config code.

oeoeaio: Good pickup. That's a bug.

Could you explain your 'copy' idea a bit better?

oeoeaio: You mean like mocking the whole Config? I suppose we could try that....

mkllnk: Yes, but I don't really know how Config works internally.

oeoeaio:
Closing in favour of #1904

Issue #1702, opened 2017-07-17
Automated scrolling issues on checkout page when choosing shipping method


sineadfenton:
When users are checking out, the tab after "Billing Info" is "Shipping Info", once the user selects from the options they click "Next", doing this automatically scrolls the page down to the end of page as outlined in the second screenshot.

I imagine the idea here is to assist with navigating users to the next of entering their payment information however a few customers have managed to do this and think that they've placed their order. Can the autoscrolling functionality be removed?

![image](https://user-images.githubusercontent.com/28807449/28288039-bbe0e350-6b35-11e7-9c18-cf0bfa18a19a.png)

![image](https://user-images.githubusercontent.com/28807449/28288152-1fe90828-6b36-11e7-9622-607c9027f16f.png)

sineadfenton:
Hey @stveep,

Just tagging you in relation to my messages on slack. Would be great if you were able to have a gander on this one

Cheers,

Sinead

sineadfenton:
![](https://github.trello.services/images/mini-trello-icon.png) [Checkout Page Jumps to the Bottom](https://trello.com/c/FfnczmVi/60-checkout-page-jumps-to-the-bottom)

sineadfenton:
Duplicate

PR #1747, opened 2017-08-02
Call toScroll after any new panel is shown (not just after validation). 

stveep:
Also added an actual delay as it still scrolled to the 'old' location.. 10 ms was too short to allow the location to update after the old panel closes (on Chrome), 50 ms seems ok. This was causing apparent scrolling to the bottom, when a large panel was closed.

As far as I can see the Accordion is not used anywhere other than the Checkout so hopefully shouldn't break anything else.

Issue: #1702 

https://community.openfoodnetwork.org.uk/t/checkout-page-jumps-to-the-bottom/176?u=lynne.ofn.tech

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
@@ -3,23 +3,22 @@ Darkswarm.controller "AccordionCtrl", ($scope, localStorageService, $timeout, $d
   value = if localStorageService.get(key) then {} else { details: true, billing: false, shipping: false, payment: false }
   localStorageService.bind $scope, "accordion", value, key
   $scope.accordionSections = ["details", "billing", "shipping", "payment"]
+  # Scrolling is confused by our position:fixed top bar - add an offset to scroll
+  # to the correct location, plus 5px buffer
+  offset_height = $("nav.top-bar").height() + 5
 
   $scope.show = (section)->
     $scope.accordion[section] = true
+    # If we call scrollTo() directly after show(), when one of the accordions above the
+    # scroll location is closed by show(), scrollTo() will scroll to the old location of
+    # the element. Putting this in a 50 ms timeout is enough delay for the DOM to
+    # have updated.
+    $timeout (->
+      $document.scrollTo $("##{section}"), offset_height, 500), 50
 
   $scope.$on 'purchaseFormInvalid', (event, form) ->
     # Scroll to first invalid section
     for section in $scope.accordionSections
       if not form[section].$valid
         $scope.show section
-
-        # If we call scrollTo() directly after show(), when one of the accordions above the
-        # scroll location is closed by show(), scrollTo() will scroll to the old location of
-        # the element. Putting this in a zero-length timeout is enough delay for the DOM to
-        # have updated.
-        $timeout ->
-          # Scrolling is confused by our position:fixed top bar - add an offset to scroll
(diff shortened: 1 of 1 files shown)


lin-d-hop:
Thanks for working on this @stveep!!
And thanks for reviewing @sauloperez!
I'll get on testing it and hopefully we can get it out next release. It will make one user (who is experience regular lost sales off the back of it) very happy!!

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
```diff
@@ -3,23 +3,24 @@ Darkswarm.controller "AccordionCtrl", ($scope, localStorageService, $timeout, $d
...
+  offset_height = $("nav.top-bar").height() + 5
 
   $scope.show = (section)->
     $scope.accordion[section] = true
+    # If we call scrollTo() directly after show(), when one of the accordions above the
+    # scroll location is closed by show(), scrollTo() will scroll to the old location of
+    # the element. Putting this in a 100 ms timeout is enough delay for the DOM to
+    # have updated.
+    $timeout (->
+      console.log("Scrolling to ##{section}")
```
enricostano: Do we really need this? I would remove debugging code.

stveep: Oops that shouldn't be there, will fix

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
```diff
@@ -11,16 +11,14 @@ Darkswarm.controller "AccordionCtrl", ($scope, localStorageService, $timeout, $d
     $scope.accordion[section] = true
     # If we call scrollTo() directly after show(), when one of the accordions above the
     # scroll location is closed by show(), scrollTo() will scroll to the old location of
-    # the element. Putting this in a 100 ms timeout is enough delay for the DOM to
+    # the element. Putting this in a 50 ms timeout is enough delay for the DOM to
     # have updated.
     $timeout (->
-      console.log("Scrolling to ##{section}")
       $document.scrollTo $("##{section}"), offset_height, 500), 50
```
mkllnk: You could make it a bit clearer like this:
```coffee
  millis_until_dom_updated = 50
  $timeout (->
    $document.scrollTo $("##{section}"), offset_height, 500),
    millis_until_dom_updated
```

Then you almost don't need the comment and people don't have to remember the function's parameter signature to know where the `50` belongs to.

You derived that value by experimentation, right? Does it depend on the browser speed updating the dom or is there a timeout in the accordion code? If we knew that the accordion code updates in an animation that takes 50 ms or so, then we would know exactly how long to wait.

sstead:
@mkllnk 
Testing Notes:
- The immediate issue (and this use case that's most commonly a problem) has been fixed. If the user clicks 'next' after selecting a shipping method, they're scrolled down to the open payment tab, rather than to the bottom of the page (in any window size).

There are still cases were I get scrolled to the very bottom of the page:
- In any view, if I select a shipping method and then instead of clicking Next, I click 'expand' on the Payment tab, I'm scrolled down to the bottom
- If I'm in mobile view (narrow window) or a small window. I've selected a shipping and payment method (or not) but refresh the page- I can get taken to the bottom (perhaps depending on where I was when I refreshed). Or if I go back to cart (after already selecting my shipping and payment methods) and then come back to the checkout page- taken to the bottom of the page.

I'm not sure if there cases are of concern, they probably occur less frequently.

PR #1758, opened 2017-08-09
Always execute local karma

sauloperez:
#### What? Why?

This ensures the dev will run the version specified in the `package.json`. Besides, it makes the rake task work like all rails tests, allowing you to pass a `RAILS_ENV`.

Doing this, the next time I need to run js unit tests I don't have to do anything other than:

```
npm install
bundle exec rake karma:run
```

- [x] update wiki: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Karma

### DIFF
--- a/lib/tasks/karma.rake
+++ b/lib/tasks/karma.rake
@@ -1,28 +1,22 @@
+ENV["RAILS_ENV"] ||= 'test'
+
 namespace :karma  do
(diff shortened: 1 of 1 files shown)


sauloperez:
This is very weird. Codeclimate reports 2 issues in the PR but when I check the details it says "No new, fixed, or changed issues.".

Issue #1761, opened 2017-08-11
Missing translations in all.js


mkllnk:
@lin-d-hop encountered a broken sign-up form on the UK production server. It seems related to some missing translations in the compiled `all.js` file. We see the same on the Australian production server, but our sign-up form is working. From the file:
```
<span class=\"translation_missing\" title=\"translation missing: en-GB.spree.firstname\">Firstname</span>
```

There are eight `translation missing` errors in that file.

The visible errors on the UK production servers come from missing translations within Javascript.

    [missing "en.welcome_to_ofn" translation]

These are generated by i18n-js which is looking for `en` translations while the `en-GB` locale is included in the all.js asset file.

This could be related to the new i18n-js gem introduced in #1564, but it could also be a configuration issue on the UK server.

daniellemoorhead:
Ping @lin-d-hop @Matt-Yorkley @sineadfenton

lin-d-hop:
Thanks for looking into this @mkllnk. Personally I am still stumped. So is @Matt-Yorkley.

Since this is affecting no other servers I think UK configuration is not playing nice with the new i18n-js gem. Definitely not solved by rebuilding assets, clearing cache etc.

lin-d-hop:
@Matt-Yorkley also tried some things on UK staging in this PR
https://github.com/openfoodfoundation/openfoodnetwork/pull/1757

sineadfenton:
![](https://github.trello.services/images/mini-trello-icon.png) [Translation Text](https://trello.com/c/mJnXNtVt/82-translation-text)

jeronimo:
Ok. So the problem is that registration page uses different layout which does not include 
<img width="466" alt="screen shot 2017-08-12 at 10 05 37" src="https://user-images.githubusercontent.com/88912/29239459-1710fa74-7f47-11e7-938c-0b7a19e2423b.png">
partial `_i18n_script.html.haml`

It exists in `darkswarm` and `admin`
<img width="421" alt="screen shot 2017-08-12 at 10 06 41" src="https://user-images.githubusercontent.com/88912/29239446-d499e5fc-7f46-11e7-98d9-de70dc330c9f.png">

But not the one that is used in registration.
<img width="924" alt="screen shot 2017-08-12 at 10 12 26" src="https://user-images.githubusercontent.com/88912/29239444-c4dc1216-7f46-11e7-9a29-1f69b68e5c16.png">

[...]

PR #1763, opened 2017-08-12
Uk/1761 add missing i18n js partial

jeronimo:
#### What? Why?
Issue - https://github.com/openfoodfoundation/openfoodnetwork/issues/1761
`/register/auth` was missing js I18n config and would fallback to 'en' local by default, but by server settings it had only 'en-GB' translations generated.

#### What should we test?

`/register/auth` should not show missing translations.


sineadfenton:
[customer unable to access shop](https://trello.com/c/qC6PXLgB/87-customer-unable-to-access-shop)

PR #1764, opened 2017-08-12
Uk/1761

lin-d-hop:
#### What? Why?

Issue - #1761
/register/auth was missing js I18n config and would fallback to 'en' local by default, but by server settings it had only 'en-GB' translations generated.

#### What should we test?

/register/auth should not show missing translations.

@jeronimo sorry to be messing about with your PR. A second branch off 1.8.13 as I'd prefer to deploy changes off the release to minimise what we are changing...


enricostano:
> by server settings it had only 'en-GB' translations generated

Just out of curiosity, how is this achieved? Is it some customization in your deploy process?

PR #1767, opened 2017-08-12
Check if items in cart have been removed from OC

Matt-Yorkley:
#### What? Why?

Addressing the bug mentioned at the bottom of #1491 (not the bug detailed in the main description)

#### What should we test?

Checking out with a cart where  some of the items have been removed from the order cycle after the cart was created but before completion of the checkout process. 

Previous behaviour allowed the order to be completed successfully despite the items being missing from the OC, potentially causing stock management issues.

New behaviour should redirect to the cart page with the "missing" items removed from the cart, and an explanatory message.

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -54,6 +54,9 @@ class CheckoutController < Spree::CheckoutController
       else
         update_failed
       end
+    elsif @order.errors.messages[:items_removed].present?
+      flash[:error] = @order.errors.messages[:items_removed].to_sentence
+      raise_insufficient_quantity
     else
       update_failed
     end
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -20,6 +20,7 @@ Spree::Order.class_eval do
   before_validation :shipping_address_from_distributor
   before_validation :associate_customer, unless: :customer_id?
   before_validation :ensure_customer, unless: :customer_is_valid?
+  before_validation :recheck_stock_availability, if: lambda { state == 'cart' }
 
   before_save :update_shipping_fees!, if: :complete?
   before_save :update_payment_fees!, if: :complete?
@@ -77,6 +78,25 @@ Spree::Order.class_eval do
   def products_available_from_new_distribution
     # Check that the line_items in the current order are available from a newly selected distribution
     errors.add(:base, I18n.t(:spree_order_availability_error)) unless DistributionChangeValidator.new(self).can_change_to_distribution?(distributor, order_cycle)
+  end
+
+  def recheck_stock_availability
+    # Adds a validation error if items in the cart have been removed
+    # from the order cycle after a cart was created but before
+    # checkout has been completed.
+    return if self.order_cycle.blank? || self.order_cycle.exchanges.outgoing.blank?
+    outgoing_variants = self.order_cycle.exchanges.outgoing.first.variants.pluck(:id)
+
+    items_available = true
+
+    self.line_items.each do |line_item|
+      unless outgoing_variants.include? line_item.variant_id
+        self.remove_variant(Spree::Variant.find(line_item.variant_id))
+        items_available = false
+      end
+    end
+
+    self.errors.add :items_removed, I18n.t('checkout.items_removed_from_order_cycle') unless items_available
   end
 
   def empty_with_clear_shipping_and_payments!
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -724,6 +724,7 @@ en:
     already_ordered:
       cart: "cart"
       message_html: "You have an order for this order cycle already. Check the %{cart} to see the items you ordered before. You can also cancel items as long as the order cycle is open."
+    items_removed_from_order_cycle: "Items in your cart have been removed from the order cycle. Please review your order before proceeding."
   shops:
     hubs:
       show_closed_shops: "Show closed shops"
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -89,7 +89,7 @@ feature %q{
     puts d.name
     puts @distributor.name
 
-    @order.state = 'cart'; @order.completed_at = nil; @order.save
+    @order.state = 'payment'; @order.completed_at = nil; @order.save
 
     login_to_admin_section
     visit '/admin/orders'
--- a/spec/features/consumer/shopping/checkout_spec.rb
+++ b/spec/features/consumer/shopping/checkout_spec.rb
@@ -16,6 +16,11 @@ feature "As a consumer I want to check out my cart", js: true, retry: 3 do
   let(:variant) { product.variants.first }
   let(:order) { create(:order, order_cycle: order_cycle, distributor: distributor) }
 
+  let(:product2) { create(:taxed_product, supplier: supplier, price: 11, zone: zone, tax_rate_amount: 0.1) }
+  let(:variant2) { product2.variants.first }
+  let!(:order_cycle2) { create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], coordinator: create(:distributor_enterprise), variants: [variant, variant2]) }
+  let(:order2) { create(:order, order_cycle: order_cycle2, distributor: distributor) }
+
   before do
     Spree::Config.shipment_inc_vat = true
     Spree::Config.shipping_tax_rate = 0.25
@@ -61,6 +66,7 @@ feature "As a consumer I want to check out my cart", js: true, retry: 3 do
         page.should have_content "An item in your cart has become unavailable"
       end
     end
+
     context 'login in as user' do
       let(:user) { create(:user) }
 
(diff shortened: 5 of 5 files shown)


app/controllers/checkout_controller.rb
```diff
@@ -22,6 +22,12 @@ def update
       check_order_for_phantom_fees
       fire_event('spree.checkout.update')
       while @order.state != "complete"
+        if @order.order_cycle.present? && !recheck_stock_availability
+          flash[:error] = t('checkout.items_removed_from_oc')
+          raise_insufficient_quantity
+          return
+        end
+
```
enricostano: Shouldn't we being check this kind of stuff at model level instead of directly in the controller? In that way you ensure that we apply this rule every in every possible scenario, not just in this controller.

app/controllers/checkout_controller.rb
```diff
@@ -62,6 +68,22 @@ def update
 
   private
 
+  # Check if items present in cart have been removed from the OC
+  # before checkout has been completed
+  def recheck_stock_availability
```
sauloperez: Agree with @enricostano . This belongs to a model validation that adds ` t('checkout.items_removed_from_oc')` in its model errors.

app/controllers/checkout_controller.rb
```diff
@@ -62,6 +68,22 @@ def update
...
+    outgoing_variants = @order.order_cycle.exchanges.outgoing.first.variants.pluck(:id)
+    items_available = true
+
+    @order.line_items.each do |i|
```
sauloperez: weird that rubocop doesn't complaint about `|i|`. I'd rather name it `line_item` because as it is, it makes you think it is an index and this is not the case.

app/controllers/checkout_controller.rb
```diff
@@ -62,6 +68,22 @@ def update
 
   private
 
+  # Check if items present in cart have been removed from the OC
+  # before checkout has been completed
+  def recheck_stock_availability
+    outgoing_variants = @order.order_cycle.exchanges.outgoing.first.variants.pluck(:id)
+    items_available = true
```
sauloperez: you don't need to initialize any variable if you use [each_with_object](https://apidock.com/rails/Enumerable/each_with_object) instead of each.

config/locales/en.yml
```diff
@@ -724,6 +724,7 @@ en:
     already_ordered:
       cart: "cart"
       message_html: "You have an order for this order cycle already. Check the %{cart} to see the items you ordered before. You can also cancel items as long as the order cycle is open."
+    items_removed_from_oc: "Items in your cart have been removed from the order cycle. Please review your order before proceeding."
```
sauloperez: I'd like not to use the `oc` abbreviation in the codebase. It adds friction for new contributors and it makes it harder for current devs to understand it. I don't think it's a big deal having a longer name if it helps comprehension.

sauloperez:
I'd be great to have a high level and brief description of the approach taken to reach the solution to make it easier for reviewers.

Matt-Yorkley:
@enricostano @sauloperez So the idea is that at the point of completing the checkout for an order we need to double-check whether or not the items in the cart have been removed from the order cycle. This is especially relevant to some of the other PRs that are currently coming through, which add functionality like allowing previous carts to be re-used when a user logs out and logs back in. This will greatly increase the chances that a cart contains items that are no longer in the OC, as longer amounts of time will have passed between the items being added and the checkout being finished. 

Currently this can create issues, as any items that are removed from the OC but are in an active cart are still able to be checkout out and ordered, causing stock problems.

My initial approach was to add this logic in the checkout controller as I think it needs to be triggered specifically as part of the checkout process, when a cart is being updated or finalised. I could move some of this logic into the Order model, but I think it'd still be best to call it from the checkout controller..? What do you think?

sauloperez:
> but I think it'd still be best to call it from the checkout controller

what makes you think that? IMO this increases the coupling in the controller making it hard to change and test.

app/controllers/checkout_controller.rb
```diff
@@ -54,6 +54,9 @@ def update
       else
         update_failed
       end
+    elsif @order.errors.messages[:items_removed].present?
```
sauloperez: this action is going too wild. We'll have to do something about it soon.

app/models/spree/order_decorator.rb
```diff
@@ -79,6 +80,22 @@ def products_available_from_new_distribution
     errors.add(:base, I18n.t(:spree_order_availability_error)) unless DistributionChangeValidator.new(self).can_change_to_distribution?(distributor, order_cycle)
   end
 
+  def recheck_stock_availability
```
enricostano: Could you please add some documentation about this method? Thanks!

app/controllers/checkout_controller.rb
Matt-Yorkley: Yeah it's not great. I couldn't do the redirect to cart page from the model, and simply passing a regular error onto the `update_failed` method at the end made the checkout page freeze up. :/

Matt-Yorkley: It seems like there's inconsistency between what's expected when an order is being updated from the cart edit page, and what's happening when the order is being updated or validated at the checkout page. Both are running through this same action but they are doing different things. And various other things seem to run through here in both json and html formats.

app/models/spree/order_decorator.rb
Matt-Yorkley: I've added a couple of explanatory lines underneath.

daniellemoorhead:
>I'm really concerned about all those conditionals that we keep adding to the checkout controller's actions. We'll need to refactor and simplify them, it's really difficult to follow, understand and debug.

@enricostano worth adding a separate issue in GH for this so it doesn't get lost?

daniellemoorhead:
Hey @mkllnk can you stage this for @sstead to test please :)

enricostano:
@daniellemoorhead I think it's part of some eventual future refactor we'll do some day, hopefully... wouldn't open an issue for now.

sstead:
Testing notes:

**Checking out to buy an items that's no longer in the OC is being prevented, but the sequencing is not quite right.**
Case- I'm a customer who had items in my cart. I navigate away from the shop, and during this time the shop manager removes items from the OC. I then click on the cart dropdown, and click checkout now. I'm taken to checkout and I select all my checkout options. When I click checkout I'm taken back to the Edit Cart page, with a red error banner message (see below). It would be better if when i click the cart dropdown Checkout Now button I'm taken to Edit Cart to resolve stock level issues before I get to checkout.
image
When I reach the Edit Cart page, there's no note next to each item that's no longer in the OC, so I don't know which items are out of stock. I don't take any items out of my cart, because I'm confused, so I click checkout at the bottom of the Edit Cart page. I'm taken to checkout again. I click 'checkout' and am returned to the Edit Cart page.

**If I have items in my cart that have been removed from the OC, and then I return to that shop, I get grumpy cat of the 'you can't do that' variety.**
image
Again, I think we should model behaviour off the scenario where On Hand stock is reduced, or runs out. In this case, if you try to go to the shop, you'll just get directed to Edit Cart, where the error messages tell you which items to fix.

Matt-Yorkley:
@sauloperez @enricostano This check isn't really working as part of the model. I added it into the Order save process, but in order to remove the missing items and update the cart it needs to resave the Order, which makes it loop back through the save process and causes more issues. 

I've tried it a few different ways but it either fails to update properly or gets stuck in a loop. Now I'm thinking to put the method call back in the the checkout controller where it was originally... :/

app/controllers/checkout_controller.rb
sauloperez: >  simply passing a regular error onto the update_failed method at the end made the checkout page freeze up. :/

Did you find out why? I'm afraid there's too much going on in this action that we don't know. Anything we can do to clarify it, no matter how small it is, it will be great.

spec/features/admin/orders_spec.rb
```diff
@@ -89,7 +89,7 @@ def new_order_with_distribution(distributor, order_cycle)
     puts d.name
     puts @distributor.name
 
-    @order.state = 'cart'; @order.completed_at = nil; @order.save
+    @order.state = 'payment'; @order.completed_at = nil; @order.save
```
sauloperez: perhaps an `#update_attributes` it's better here.

spec/features/consumer/shopping/checkout_spec.rb
```diff
@@ -433,5 +419,60 @@
...
+
+        expect(page).to have_content 'Your shopping cart'
+        expect(page).to have_content 'Items in your cart have been removed from the order cycle. Please review your order before proceeding.'
+      end
+    end
+  end
+
+  private
+
+  def fill_in_basic_checkout_details
```
sauloperez: Did you check whether we have something like this or similar (I guess we do)? if so, I'd rather check if we can extend it or directly use it. If not, might be the time to move this new method into a helper module in `spec/support/` and start using it to DRY feature specs.

sauloperez:
> Now I'm thinking to put the method call back in the checkout controller where it was originally... :/

If there's no other way for now, could you consider using a service object to avoid having an even more bloated controller? It would only make things worse.

myriamboure:
So what should we do about that PR @enricostano @Matt-Yorkley @lin-d-hop ? It is blocking the #1698 if I undertand well, which is in code review. From what I read it seems it needs some inception to find how to do this properly. Shall we just put on hold (close for now) and do the proper inception when we decide to prioritize bug #1491 in the backlog?
Then I guess we can remove #1698 from code review and put it back to all the things meanwhile with issue #1031 ... ?

myriamboure:
Ok so after discussion with @enricostano and @oeoeaio we are closing it for now as it needs further inception. We'll reopen when the issue is prioritized and dev can have an inception session on it :-)

PR #1720, opened 2017-07-25
Email/change enterprise confirmations

Matt-Yorkley:
Removing enterprise confirmations for #946. I think there isn't a specific issue for this, but it's mentioned somewhere in a list of "issues that need to be created".

Shouldn't be merged before #1768 is finished and merged.

Note: The implementation for how the user is set to receive emails was changed slightly from the original spec, but the end result is more or less the same. See discussions below about the `receives_notifications` flag.

**Testing**

This PR removes the need to confirm enterprises by email, but also includes changes to the way the user that receives emails for an enterprise is set, as this was previously done via the confirmable `Enterprise.email` field. 

Now the "notification" field (for the user that will receive emails) is a dropdown list of the enterprises' current managers, which sets a flag on that manager. Only managers with confirmed emails can be selected to receive emails. I made some adjustments to the UI to make this more clear, and to show which managers have confirmed their emails.

screenshot from 2017-07-27 13-12-29

The way managers are added to an enterprise is unchanged. A separate PR will add functionality for inviting an unregistered user to be a manager from within the managers UI.

I also adjusted the managers UI so that the current owner and/or current user that receives emails cannot be deleted from the list of managers, as this can cause all kinds of problems and could previously be done by mistake if the user wasn't careful. The delete button next to the owner or contact user is greyed out and disabled. They become delete-able again once a different user is selected for these roles. I readjusted the UI to show which users are the owner/contact. 

screenshot from 2017-07-28 17-13-17

PR #1768, opened 2017-08-15
Email/confirmable users

Matt-Yorkley:
#### What? Why?

Adds email confirmations to the user model as part of the #946 epic. Supercedes #1690. 

#### What should we test?

User signs up, confirmation email gets sent. 
Clicking the link in confirmation email confirms the user's account. 
User can then log in. 
If the user tries to log in before email is confirmed an appropriate message will show in the login modal.

#### Release notes

Users will now have to confirm their email after sign-up before they can log in.

#### Discourse thread

https://community.openfoodnetwork.org/t/confirmation-emails-registration-flow/237

#### Screenshots

After user signup with login modal:
screenshot from 2017-08-15 14-10-19

Attempting to log in:
screenshot from 2017-08-15 14-09-59

After clicking resend confirmation:
screenshot from 2017-08-15 14-05-53

### DIFF
--- a/app/controllers/user_confirmations_controller.rb
+++ b/app/controllers/user_confirmations_controller.rb
@@ -13,10 +13,10 @@ class UserConfirmationsController < DeviseController
 
     self.resource = resource_class.send_confirmation_instructions(resource_params)
 
-    if successfully_sent?(resource)
-      set_flash_message(:success, :confirmation_sent) if is_navigational_format?
-    else
-      set_flash_message(:error, :confirmation_not_sent) if is_navigational_format?
+    if successfully_sent?(resource) && is_navigational_format?
+      set_flash_message(:success, :confirmation_sent)
+    elsif is_navigational_format?
+      set_flash_message(:error, :confirmation_not_sent)
     end
 
     respond_with_navigational(resource){ redirect_to spree.admin_path }
@@ -26,10 +26,10 @@ class UserConfirmationsController < DeviseController
   def show
     self.resource = resource_class.confirm_by_token(params[:confirmation_token])
 
-    if resource.errors.empty?
-      set_flash_message(:success, :confirmed) if is_navigational_format?
-    else
-      set_flash_message(:error, :not_confirmed) if is_navigational_format?
+    if resource.errors.empty? && is_navigational_format?
+      set_flash_message(:success, :confirmed)
+    elsif is_navigational_format?
+      set_flash_message(:error, :not_confirmed)
     end
 
     respond_with_navigational(resource){ redirect_to login_path }
--- a/app/controllers/user_confirmations_controller.rb
+++ b/app/controllers/user_confirmations_controller.rb
@@ -8,9 +8,6 @@ class UserConfirmationsController < DeviseController
 
   # POST /resource/confirmation
   def create
-    self.resource = resource_class.find_by_unconfirmed_email_with_errors(resource_params)
-    authorize! :resend_confirmation, resource
-
     self.resource = resource_class.send_confirmation_instructions(resource_params)
 
     if successfully_sent?(resource) && is_navigational_format?
@@ -19,7 +16,7 @@ class UserConfirmationsController < DeviseController
       set_flash_message(:error, :confirmation_not_sent)
     end
 
-    respond_with_navigational(resource){ redirect_to spree.admin_path }
+    respond_with_navigational(resource){ redirect_to login_path }
   end
 
   # GET /resource/confirmation?confirmation_token=abcdef
--- a/app/controllers/user_confirmations_controller.rb
+++ b/app/controllers/user_confirmations_controller.rb
@@ -10,10 +10,12 @@ class UserConfirmationsController < DeviseController
   def create
     self.resource = resource_class.send_confirmation_instructions(resource_params)
 
-    if successfully_sent?(resource) && is_navigational_format?
-      set_flash_message(:success, :confirmation_sent)
-    elsif is_navigational_format?
-      set_flash_message(:error, :confirmation_not_sent)
+    if is_navigational_format?
+      if successfully_sent?(resource)
+        set_flash_message(:success, :confirmation_sent)
+      else
+        set_flash_message(:error, :confirmation_not_sent)
+      end
     end
 
     respond_with_navigational(resource){ redirect_to login_path }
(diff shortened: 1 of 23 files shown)


app/models/spree/user_decorator.rb
```diff
@@ -18,10 +18,16 @@
...
   attr_accessible :enterprise_ids, :enterprise_roles_attributes, :enterprise_limit, :bill_address_attributes, :ship_address_attributes
-  after_create :send_signup_confirmation
 
   validate :limit_owned_enterprises
 
+  # We use the same options as Spree and add :confirmable
+  devise :confirmable, reconfirmable: true
+  handle_asynchronously :send_confirmation_instructions
+  handle_asynchronously :send_on_create_confirmation_instructions
+  # TODO: Later versions of devise have a dedicated after_confirmation callback, so use that
```
enricostano: Any link to the code of future Devise? Just add it here in the TODO comment please, thanks!

app/controllers/user_confirmations_controller.rb
```diff
@@ -13,10 +13,10 @@ def create
 
     self.resource = resource_class.send_confirmation_instructions(resource_params)
 
-    if successfully_sent?(resource)
-      set_flash_message(:success, :confirmation_sent) if is_navigational_format?
-    else
-      set_flash_message(:error, :confirmation_not_sent) if is_navigational_format?
+    if successfully_sent?(resource) && is_navigational_format?
```
mkllnk: Looks like this could have been:
```ruby
    if is_navigational_format?
      if successfully_sent?(resource)
        set_flash_message(:success, :confirmation_sent)
      else
        set_flash_message(:error, :confirmation_not_sent)
      end
    end
```

sstead:
- [ ] User signs up, confirmation email gets sent.
- [ ] Clicking the link in confirmation email confirms the user's account.
- [ ] User can then log in.
- [ ] If the user tries to log in before email is confirmed an appropriate message will show in the login modal.

Matt-Yorkley:
@sstead I've pushed an update with your suggested changes:

- Changed the email template to adjust confirmation message and fix spelling mistake. 
- Changed the login modal to display signup message in the modal instead of reloading and showing in a flash message.
- This also fixes the reload/missing message issue with signup feedback in the /registration page.

I think the functionality for returning the user to the same page they signed up from via the email link will be a separate PR, after this one has gone through.

mkllnk:
@Matt-Yorkley Is there any way we could feature-toggle the fact that people can't log in without confirming their email address? We are a bit worried about all the existing users and it would be nice to have a transition period.

sauloperez:
> Is there any way we could feature-toggle the fact that people can't log in without confirming their email address?

I would simply run a data migration that confirms all existing users (we assume everyone is confirmed until now, right?) after we run the migration and before we deploy a new release. This way users shouldn't notice any behaviour change and we avoid adding more complexity.

kirstenalarsen:
yes please!!

Matt-Yorkley:
So do we want an on/off toggle, or do we set current users to confirmed?

Matt-Yorkley:
I imagine we won't be turning it "off" again, and all new users will have to confirm emails from now on, so maybe @sauloperez's idea is the best option?

enricostano:
@Matt-Yorkley yes I would follow @sauloperez 's suggestion. The data migration will do the trick.

daniellemoorhead:
>So do we want an on/off toggle, or do we set current users to confirmed?

I say yes to pre-existing users being confirmed. Tis a sensible approach. Let's do it  @Matt-Yorkley :)

mkllnk:
@Matt-Yorkley I'm really sorry, but I accidentally overwrote your commits. Can you upload them again? Does anyone else have the commits ab03fc76b13855abc5ad3010cd343ec5b895d983 and 000eba5ee023b147e8b2ffb3d00ae58edee0c82e?

Matt-Yorkley:
@mkllnk I've restored those two commits, plus 0289f50 with the adjusted migration which also got deleted.

sstead:
- [ ] Changed the email template to adjust confirmation message and fix spelling mistake.
- [ ] Changed the login modal to display signup message in the modal instead of reloading and showing in a flash message. YES
- [ ] This also fixes the reload/missing message issue with signup feedback in the /registration page.

daniellemoorhead:
@Matt-Yorkley do you think you'd be able to have this fixed and ready to be staged for @sstead to test by Friday morning AU team (Thursday night your time)?

mkllnk:
I can confirm that our default locale `en.yml` is missing some entries which are present in other locales though. It looks like Pierre pre-empted the absence of enterprise confirmation emails. @Matt-Yorkley Is it right that this pull request doesn't make sense without the removal of enterprise confirmations? Have we been to quick thinking we could merge this?

Matt-Yorkley:
@mkllnk This initial PR just adds confirmation to the users, yes. I guess there will be a small amount of overlap with user and enterprise confirmations until the PR to remove enterprise confirmations is merged as well. I had hoped we could merge this first, but maybe we'll have to combine them both into a huge PR? The other one (for enterprise confirmations) is already quite big and complex.

mkllnk:
@sstead @daniellemoorhead What would you say about user confirmations and enterprise confirmations for a while? Is that okay until we merged the next step? All the existing users don't need to confirm their current account, but they would still need to confirm an enterprise.

@Matt-Yorkley I see how merging this is still appealing. Depending on how the others see the point above, I can imagine to merge it. But I think we should add the missing translations again, otherwise we will end up with a buggy production site.

sstead:
@mkllnk I think it's ok to merge this, and have users confirming their enterprises for a while, until the next piece of work is ready. But we should make sure the email is translated correctly before we push it to production.

mkllnk:
Just looking at the state of this, I see three outstanding issues:

- merge conflict
- https://github.com/openfoodfoundation/openfoodnetwork/pull/1720/files#r148705378
- https://github.com/openfoodfoundation/openfoodnetwork/pull/1720/files#r148710469

@Matt-Yorkley If you could reply to these, we can let Sally test this soon. It looks like it's almost there. :-)

Matt-Yorkley:
This PR has been combined with / merged into #1720.

enricostano:
> This PR has been combined with / merged into #1720.

I hope one day we'll stop doing this 😬 🙏 😘

PR #1775, opened 2017-08-23
Transifex

mkllnk:
#### What? Why?

Merging translations in every release cycle.

#### What should we test?

Nothing.

#### Release notes

Not necessary.

Not at all.

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -2,7 +2,7 @@ es:
   activerecord:
     attributes:
       spree/order:
-        payment_state: Estado del pago
+        payment_state: Estado del pagos
         shipment_state: Estado del envío
     errors:
       models:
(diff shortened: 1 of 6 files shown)


daniellemoorhead:
@mkllnk who do you think is best to code review this?

mkllnk:
@daniellemoorhead No code review needed. We just merge it without testing. It has just not been automated yet, so I have to do it manually.

sigmundpetersen:
@mkllnk @daniellemoorhead should we have a nice label for this? maybe `skip code review`? does this imply "not in release notes" or should we have a separate label for that as well?

mkllnk:
I added that label. But I'm not sure if it implies "not in the release notes". For example, we could improve the documentation like the README file and include that in the release notes. That could be reviewed though.

daniellemoorhead:
Sally usually adds a list at the bottom of release notes of all the non-user centric stuff that goes live in the release, and I'd categorise this as one of them. Worth knowing it's in the release, not worth knowing much more than that.

config/locales/es.yml
```diff
@@ -2,7 +2,7 @@ es:
   activerecord:
     attributes:
       spree/order:
-        payment_state: Estado del pago
+        payment_state: Estado del pagos
```
enricostano: cc/ @sseerrggii should we fix this in Transifex?

sseerrggii: :+1:

daniellemoorhead:
@mkllnk will this be part of the release today or should it be moved to September?

PR #1776, opened 2017-08-23
Stripe fixes and gem upgrade

stveep:
This is Rob's Stripe Connect branch (https://github.com/openfoodfoundation/openfoodnetwork/pull/1468) with a fix for some specs that broke when the credit card saving was refactored. I also bumped the version of the Stripe gem we used.

I have also addressed a couple of the PR comments but didn't have time to go through them all.

There are still a couple of flaky iframe-related specs on Travis that pass for me locally, so I don't know what is wrong with these. Maybe they will pass in the PR build....
https://travis-ci.org/stveep/openfoodnetwork/jobs/267727702


Gemfile.lock
```diff
@@ -133,8 +134,8 @@ GEM
       sprockets (~> 2.2.1)
     active_model_serializers (0.8.3)
       activemodel (>= 3.0)
-    activemerchant (1.50.0)
-      activesupport (>= 3.2.14, < 5.0.0)
+    activemerchant (1.63.0)
```
enricostano: which change in `Gemfile` generated this specific upgrade?

app/assets/stylesheets/admin/components/jquery_dialog.scss
```diff
@@ -42,21 +42,25 @@ light: #ccc
...
-    color: #545454;
-  }
-}
+// .ui-dialog  {
+//   .ui-state-hover, .ui-state-focus{
+//     border: none;
+//     background: none;
+//     color: #545454;
+//   }
+// }
```
enricostano: Is this WIP?

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,133 @@
...
+            options[:address] = {
+              address1: address.address1,
+              address2: address.address2,
+              city: address.city,
+              zip: address.zipcode
+            }
+
+            if country = address.country
+              options[:address][:country] = country.name
+            end
```
enricostano: Why the `country` variable?
```ruby
if address.country
  options[:address][:country] = address.country
end
```

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,133 @@
...
+
+            if state = address.state
+              options[:address].merge!(state: state.name)
+            end
```
enricostano: same here with `state` var

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -0,0 +1,133 @@
...
+      def card_to_store(source)
+        source = update_source!(source)
+        if source.number.blank? && source.gateway_payment_profile_id.present?
+          # StripeJS Token
+          source.gateway_payment_profile_id
+        else
+          # Spree::CreditCard object
+          source
+        end
+      end
```
enricostano: Probably in the future we'll need some kind of `CreditCard` object that could incapsulate all this logic.

mkllnk:
I did a few more tests. Two specs seem to always fail on Travis, but pass locally:

- spec/features/consumer/shopping/embedded_shopfronts_spec.rb:17
- spec/features/consumer/shopping/embedded_shopfronts_spec.rb:24

I ran the tests with a couple of `sleep 1` commands and they still failed on Travis. So it's probably not a race condition.

I ran all tests of that Travis build to test any influences of other specs. They still all passed on my machine.

```sh
export CI_NODE_TOTAL=5
export CI_NODE_INDEX=3
bundle exec rake 'knapsack:rspec[--tag ~performance]'
```

There must be another difference in the Travis environment. We could try to run the tests in a local Docker image provided by Travis: https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image

But I don't have time for this today. :-(

Gemfile.lock
stveep: I think upgrading the Stripe gem. It's not strictly necessary to upgrade it now, but I think it will be useful for the (near) future Stripe development - it has some features to make testing Stripe integration easier (e.g. it includes a mock server you can run locally to respond to Stripe API requests).

stveep:
I've investigated a bit more. I used the Travis ruby docker image (travisci/ci-garnet:packer-1499451976) and ran the build command: `bundle exec rake 'knapsack:rspec[--tag ~performance]'`
 
Only the first of those specs failed `spec/features/consumer/shopping/embedded_shopfronts_spec.rb:19`.

When I ran `bundle exec rspec spec/features/consumer/shopping/embedded_shopfronts_spec.rb`  everything passes. Maybe it's something to do with knapsack? I don't really know anything about knapsack though...

Also pinging @Matt-Yorkley in case this might be something specifically to do with the iframe specs?

Gemfile.lock
enricostano: I was just asking in case we manually upgraded it, if we need a specific version we should add it explicitly to `Gemfile` (even if it's an indirect dependency).
In the case instead of being ugraded while upgrading another gem, it's OK to leave it as it is. Thanks!

oeoeaio: @enricostano it was manually upgraded, we need at least `1.63.0`. `activemerchant` version is specified by the Spree gem, so the change was made there (in our fork). I will make sure we specify the `activemerchant` requirement in our Gemfile. I guess we will have to deal with this again when merging with Spree upgrade work.

app/assets/stylesheets/admin/components/jquery_dialog.scss
oeoeaio: Nope, looks like it wasn't cleaned up. Added to my list.

oeoeaio:
Incorporated into #1468

Issue #1782, opened 2017-08-29
Enterprises in the same address get overlapped in map


sauloperez:
When multiple enterprises have the same address, it's impossible to find them all in the map. Their markers get totally overlapped hiding all enterprises but the last one that was created.

At maximum zoom other markers don't show up:
![mapa](https://user-images.githubusercontent.com/762088/29828838-247324cc-8cde-11e7-8f8b-4e0572ab99bb.png)

So, can only click the last one of them:
![mapa_capfoguer](https://user-images.githubusercontent.com/762088/29828845-27dac840-8cde-11e7-96aa-10bce68f8812.png)

And the angular's enterprises collection correctly lists them with the same coordinates:

```js
{
  "name":"Cervesa Capfoguer, SCCL",
  "id":31,
  "description":"Cooperativa de producci\u00f3 i distribuci\u00f3 de cervesa artesana",
  "latitude":41.4461847,
  "longitude":2.2087712,
  ...
},
{
  "name":"Cooperativa de Consum El Cab\u00e0s (Ca la Sisqueta)",
  "id":30,
  "description":"",
  "latitude":41.4461847,
  "longitude":2.2087712,
  "long_description":"Cooperativa de consum de fruita i verdura ecol\u00f2gics. Ens trobem un cop per setmana a Ca la Sisqueta per recollir les nostres comandes.",
  ...
}
```

We need to add some spacing between them so that all markers in the same coordinates are clickable.

enricostano:
https://stackoverflow.com/questions/14248365/how-to-render-multiple-markers-at-the-exact-same-coordinates-in-google-maps-api

maybe?

sauloperez:
Perhaps it's better to allow users to introduce latitude and longitude manually to fix these cases but that might involve more work than a pure JS solution.

Morantron:
![](https://media.giphy.com/media/3o7btV5GAWaGkwjTrO/giphy.gif)

Morantron:
Looks like google maps has some clustering options thingies

![image](https://user-images.githubusercontent.com/3304507/29839106-652457ba-8cfd-11e7-86f2-fb6fc179fc64.png)

( still need to center the image properly :sweat_smile: )

Maybe the modal when clicking on a marker can be expanded to showing more than one ( in the case of having many in the _exact same location_ )

sauloperez:
Awesome having you onboard @Morantron 💪 !

[...]

PR #1786, opened 2017-08-30
Handle markers on same location

Morantron:
Addressing https://github.com/openfoodfoundation/openfoodnetwork/issues/1782

#### What? Why?

Map section cannot handle properly markers on same location.

#### What should we test?

Try map section, see if marker clustering feels good.

#### Release notes

Enterprises that share location now properly show in the map.


daniellemoorhead:
Hey @Morantron do you think you could fill in the PR template with details? This helps a lot in the code review and testing process. Cheers!

daniellemoorhead:
@sstead it would be great if we could get you to test this from a user's perspective before we push it live. And @mkllnk could you please code review this.

sauloperez:
@Morantron anything left for us to start reviewing? can we remove the WIP already?

Morantron:
Hi! I've pushed latest changes, this is not WIP anymore.

Attaching an screencast showing the new behavior.

[Screencast](https://imgur.com/a/OVkIQ)

app/assets/javascripts/darkswarm/services/enterprise_modal.js.coffee
```diff
@@ -1,8 +1,12 @@
...
-    open: (enterprise)->
+    open: (enterprises)->
       scope = $rootScope.$new(true) # Spawn an isolate to contain the enterprise
+      scope.enterprises = if angular.isArray(enterprises)
+        enterprises
+      else
+        [enterprises]
 
-      scope.enterprise = enterprise
+      console.log(scope.enterprises)
```
sauloperez: :warning:

config/database.yml
```diff
@@ -22,13 +22,13 @@ production:
...
+  username: ofn_user
   password: f00d
 
 staging:
   adapter: postgresql
   encoding: unicode
   database: open_food_network_prod
   pool: 5
-  username: ofn
+  username: ofn_user
```
sauloperez: :warning:

sauloperez:
I'm very sorry @Morantron but there's a spec failing in https://travis-ci.org/openfoodfoundation/openfoodnetwork/jobs/283757904

sstead:
This is a good first visualisation of what this could look like, but might need some more thought...

1) When a user opens the map, we want the level of zoom to be just enough to encompass all enterprises on the map. A comparison between how the map looks when first opened on Production and Staging is shown below for Australia... On staging it's zoomed in on a random point, which is not ideal.
image

2) Further, when I zoom out, there appears far fewer enterprises than before they were clustered. This makes the map look a bit sparse. I'm not sure what the best solution is, because as you can see above our production looks very cluttered, but we also don't want the map to look empty.
image

3) Clicking on this marker does nothing. The system needs to zoom in another layer, when this 3 splits into 1 and 2 markers. It's not zooming upon clicking- I need to manually zoom to see the profiles.
image
Sometimes this also happens at a more zoomed out level...
image

4) When >1 profiles are at the same address their profiles open as one modal, and you can scroll through each profile. This is a problem for some of our users such as farmer's markets of hubs, who create many (30+) profiles at the same address.
image

5) We're losing the red/green categories. For example in the pic below, 3 of the profiles are producers and one is a hub, yet the marker is red. The user can't tell if they are producer or not at a glance.
image
And when you look at the map broadly there is also few green markers..
image

6) This feature hasn't been extended to the group maps. Eventually it makes sense that both maps are consistent.
image

sauloperez:
Thanks @sstead this clearly needs a second thought. We thought this could fix #1782 but it has more implications than expected. Let's close until we come up with a better solution that considers all your findings.

sigmundpetersen:
Should it be taken out of Oct milestone @sauloperez if it's not going to be merged?

sauloperez:
Oops, sorry. Thanks for the heads-up @sigmundpetersen . We won't be touching this again as we have other and more prioritary issues.

PR #1794, opened 2017-09-04
Uk/1663variant ordering emails

lin-d-hop:
_This was originally approved in [Duende's branch](https://github.com/openfoodfoundation/openfoodnetwork/pull/1709). The PR was created before adjustments to Buildkite that solved our issue of some PRs not triggering builds..._ 

for confirmation email and summary screen after checkout. It was done in the cart and now in emails.
Requested in issue #1663

What? Why?

What should we test?

### DIFF
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -25,6 +25,11 @@ Spree::LineItem.class_eval do
     end
   }
 
+  # Find line items that are from order sorted by variant name and unit value
+  scope :sorted_by_name_and_unit_value, joins(variant: :product).
+    reorder('spree_products.name asc, spree_variants.unit_value asc').
+    select('spree_line_items.*')
+
   scope :supplied_by, lambda { |enterprise|
     joins(:product).
       where('spree_products.supplier_id = ?', enterprise)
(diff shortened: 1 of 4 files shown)


lin-d-hop:
@OliverUK Try this one for testing on staging (when the tests have finished). Thanks!

OliverUK:
@lin-d-hop is there anything I can do to progress this?  I was unable to test purchase order emails because no emails are sent from UK staging.

lin-d-hop:
@OliverUK Only asking if it can be staged at some point on another staging server. I'll get to fixing our staging emails after uni exams are over in 1.5 weeks so if we can wait until then then that's also fine.

OliverUK:
@lin-d-hop if it only delays things by 1.5 weeks then no problem.

lin-d-hop:
This is the PR @OliverUK so not merged. Do you have time to test on staging today? We might be able to get it in tomorrows release!

lin-d-hop:
actually my mistake. #1709 was merged.
Ok now we need to look into that.

OliverUK:
OK @lin-d-hop  leaving it with you for now in that case. Too technical for me to know about merged PRs etc. But I'm around much of today if you need me for testing.

Issue #1801, opened 2017-09-05
Missing translations in "new product" page


myriamboure:
comment : on the page title, the problem is probably not a translation one, just that the title remains the same as when you are on the main "products" page, but it should be different like "new product" so I guess you need to create a new string and add it somewhere ?

![illustration3](https://user-images.githubusercontent.com/8135376/30078406-56471176-927d-11e7-9c76-f295dc3fbea7.jpeg)

sigmundpetersen:
#1784 duplicate?

myriamboure:
Thanks @sigmundpetersen ! Seems there are 2 things that were not pointed in the previous issue : the page title and the image selection, as the other one is closed, I leave this one open, if you want to finish this off @leandroalemao that would be awesome :-) Nice to e-meet you btw !

sauloperez:
I believe the image selection depends on the language setting of your machine as it is an operative-system-level UI element.

myriamboure:
Ok I don't manage to test, not Ubuntu savy enough, but I'll try that tonight with my boyfriend :-o Still the title key needs to be added ;-)

myriamboure:
@RachL Maybe you can test if you have a French language based system ? Do you see the image upload thing in French ? Thanks ;-)

sauloperez:
> Still the title key needs to be added ;-)

yep, I'll take a look

RachL:
@myriamboure yes it is in french ! 
![image](https://user-images.githubusercontent.com/6525576/30116077-d8a65768-931c-11e7-8509-3cb6f302f479.png)

However, the button "new product" does not work on the latest version of Firefox. I had to login on chrome to see this. Do you know if there is already an issue open for that ?

myriamboure:
I don't think so @RachL but you can check by typing in the search bock on the issues page... and then please feel free to open your first OFN Github issue :-) Is it 55.0.2 ? It have this version and it seems to work on my computer.

sauloperez:
Solved in https://github.com/openfoodfoundation/openfoodnetwork/pull/1814 @myriamboure

leandroalemao:
Hi @sauloperez @myriamboure @enricostano @sigmundpetersen I've just opened the PR #1815 to fix the missing translations on edit products page.
I hope all good.
Tks
Leandro

myriamboure:
Ok, we are happy to test when it's deployed somewhere if needed ;-)

enricostano:
[...]

PR #1814, opened 2017-09-06
Change title when navigating to new product page

sauloperez:
#### What? Why?

I'm trying to solve the missing title translation from https://github.com/openfoodfoundation/openfoodnetwork/issues/1801. It turned out to be not so simple :sweat: .

We navigate to the new product page by replacing the current page contents through an ajax call, so we also need change the section title with JS.

#### What should we test?

I'd be good to check the edit and new product pages.

#### Dependencies

https://github.com/openfoodfoundation/openfoodnetwork/issues/1801

### DIFF
--- a/app/views/spree/admin/products/new.js.erb
+++ b/app/views/spree/admin/products/new.js.erb
@@ -1,0 +1,12 @@
+<%# This chunk is just a copy of Spree's core/app/views/spree/admin/products/new.js.erb %>
+$("#new_product").html('<%= escape_javascript(render :template => "spree/admin/products/new", :formats => [:html], :handlers => [:erb]) %>');
+handle_date_picker_fields();
+<% unless Rails.env.test? %>
+  $('.select2').select2();
+<% end %>
+$("#table-filter").hide();
+$("#admin_new_product").parent().hide();
+
+<%# We need to replace the page's title as well. We're navigating to a new page
+  although through ajax %>
+$('#content-header .page-title').html('<%= t('.title') %>');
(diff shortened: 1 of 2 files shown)


app/views/spree/admin/products/new.js.erb
```diff
@@ -0,0 +1,12 @@
+<%# This chunk is just a copy of Spree's core/app/views/spree/admin/products/new.js.erb %>
+$("#new_product").html('<%= escape_javascript(render :template => "spree/admin/products/new", :formats => [:html], :handlers => [:erb]) %>');
+handle_date_picker_fields();
+<% unless Rails.env.test? %>
+  $('.select2').select2();
+<% end %>
+$("#table-filter").hide();
+$("#admin_new_product").parent().hide();
```
sauloperez: does somebody know how we could extend this Spree view without having to copy its contents?

sauloperez: I tried using ruyb's `load` and `Deface` and I couldn't make any of them work. Not worth spending any more time on it.

enricostano: Yeah it seems an old `deface` issue 😞 https://github.com/spree/deface/issues/103

leandroalemao:
Hi @sauloperez I had a look here and seems to be ok for me. the title has been updated when I clicked on the New Product button and also the code looks very fine for me. 👍 😄  Tks. Leandro

sstead:
All good @mkllnk 
image

Issue #1791, opened 2017-08-31
/register allows submitting multiple times while the enterprise is still being created


sauloperez:
## What?

From the registration wizard in `/register` users are allowed to submit the `create profile` form multiple times while the first request didn't finish yet. This causes an alert dialog to appear with a somewhat clunky error message, totally unexpected for the user.

I managed to reproduce the issue by adding a breakpoint in https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/api/enterprises_controller.rb#L17

See the screenshot below. I successfully made 4 requests while I didn't get a first response.

![create_profile](https://user-images.githubusercontent.com/762088/29929022-b455d692-8e6a-11e7-9271-0d068edf1c84.png)

The first one creates the enterprise but the others raise a validation error as expected. This however is a bit confusing for the user in terms of UX, especially given the error message in the alert above.

It's also not clear what the state of the enterprise is (in the database) after the error shows up. In a real scenario with an actual user, as we weren't sure, we started the registration from scratch.

To avoid this situation, I suggest we disable the `Create Profile` button until we receive the response.

https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/darkswarm/services/enterprise_registration_service.js.coffee#L14 is apparently the function that implements this behavior.

daniellemoorhead:
@Matt-Yorkley is it worth waiting until the email confirm functionality is live before fixing this?

sauloperez:
@daniellemoorhead I believe it's a 5min change, so unless email confirmation is ready I'd fix it.

Matt-Yorkley:
@daniellemoorhead I think this quick change to the button behaviour won't be any problem for the email confirmation branches.

PR #1816, opened 2017-09-07
Disable create profile on submit

sauloperez:
#### What? Why?

This prevents people re-submitting the form multiple times. Although the backend validates it, we show an ugly alert message that is hard for users to understand. It's better in terms of UX/UI if we avoid that situation in the first place.

This fixes https://github.com/openfoodfoundation/openfoodnetwork/issues/1791

#### What should we test?

The whole registration form. It should work as always but if there was some lag while creating the enterprise, you shouldn't be able to resubmit the form.

#### Dependencies

https://github.com/openfoodfoundation/openfoodnetwork/issues/1791

### DIFF
--- a/app/assets/javascripts/darkswarm/services/enterprise_registration_service.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/enterprise_registration_service.js.coffee
@@ -11,6 +11,10 @@ Darkswarm.factory "EnterpriseRegistrationService", ($http, RegistrationService, 
       for key, value of enterpriseAttributes
         @enterprise[key] = value
 
+    # Creates the enterprise and redirects to the about step on success.
+    #
+    # @param callback [Function] executed at the end of the operation both in
+    #   case of success or failure.
     create: (callback) =>
       Loading.message = t('creating') + " " + @enterprise.name
       $http(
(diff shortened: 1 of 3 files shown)


andresgutgon:
👍

app/assets/javascripts/darkswarm/services/enterprise_registration_service.js.coffee
```diff
@@ -11,7 +11,11 @@ Darkswarm.factory "EnterpriseRegistrationService", ($http, RegistrationService,
       for key, value of enterpriseAttributes
         @enterprise[key] = value
 
-    create: =>
+    # Creates the enterprise and redirects to the about step on success.
+    #
+    # @param callback [Function] executed at the end of the operation both in
+    #   case of success or failure.
```
sauloperez: I couldn't find any example of function documentation in darkwswarm so I used the first doc format I found on google for coffee script 👉  https://github.com/coffeedoc/codo

daniellemoorhead:
Hey @enricostano did this get tested before it was merged? Did it need to be?

sauloperez:
Yes, it was tested in our staging.katuma.org. Is that ok?

daniellemoorhead:
Should be fine @sauloperez. The only reason I asked it because usually most things are tested by Sally in Aus or Oliver in the UK, and now Simone in NL. If you have people that can test in ES that's great news, it will be good to have extra test help. :-)

Ping @oeoeaio and @mkllnk and @kirstenalarsen to add their thoughts on this if they have anything to add.

Issue #1118, opened 2016-08-03
Make "on demand" check box a default column in the bulk edit product table


myriamboure:
I have noticed in support with new users that they don't find intuitively how to say that the product is available on demand when they create products from the bulk edit page. (=click on "columns" and then "on-demand")
So some of them just put a false arbitrary quantity (see example below), but that pollute the base then and can be annoying if various hubs use the same supplier.
As "on-demand" is very often use for lots of products, I propose to make it visible as a default column on top of the existing ones.

![screenshot from 2016-08-03 12-36-51](https://cloud.githubusercontent.com/assets/8135376/17362848/5860913e-5977-11e6-936a-a5db66c8232a.png)

myriamboure:
Any objection @sstead , what do you think about it, as you are in contact with users often in support?

myriamboure:
@sstead is it ok if we do that? It is confusing for users in France not to have the "on demand" column by default, they don't understand and put a big amount instead :-o

sstead:
Yes I think this is very logical!

myriamboure:
Same, seems pretty easy for Claire @pacodelaluna ? On all the issues where I propose Claire I can put the label "ruby for newbies" or "angular for newbies" ... but I don't know which one to use... or we can assign Claire to the tasks we identify for her ? What do you think ?

oeoeaio:
See [my comment](https://github.com/openfoodfoundation/openfoodnetwork/pull/1817#issuecomment-328395075) on @ltrls's PR. I have reservations but happy to be overruled if others think the interface remains usable.

PR #1817, opened 2017-09-07
Make "on demand" columns visible by default on "bulk product edit" and "inventory" (variant overrides) tables

ltrls:
This is the PR for #1118, it should be quick to review!


daniellemoorhead:
@enricostano @sauloperez @pacodelaluna feel free to also code review this if you have time :)

sauloperez:
If I understood correctly the issue this PR is related to, what we want is to add `on demand` column on the bulk edit table with its default value to `true`. In that case, we're missing the change on the template to actually add the column.

oeoeaio:
Was there a discussion about this? I think enabling it be default will make the interface too cluttered. Individual users can make it appear by default for both of these interfaces by setting their column preferences. My vote would be against this as a blanket rule.

myriamboure:
Yes @oeoeaio we have discussed it and agreed with @sstead basically all our users look for it and make mistakes and but stock to 9999 because they don't see the column, so it's important in term of UX to make it visible by default, the table is still  very readable and not heavier... thanks :-)

sstead:
Testing looks good!
On demand shows by default, and the user can then set their column preferences in both Products and Inventory.
image

Issue #1107, opened 2016-07-25
Need a manual hack for a front end translation...


myriamboure:
From the begining in theFrench instance, we have useda manual hack to display the translation on the "end date" of an order cycle. At each upgrade we needto do it again, so we would like to solve the issue more sustainably.
Any idea how we can display the translated string? @mkllnk or @sigmundpetersen maybe?

![pb_traductionfr](https://cloud.githubusercontent.com/assets/8135376/17096112/1e291fb6-5258-11e6-9bf1-99fdad3417af.png)

myriamboure:
The problem doens't happen on our  staging server, so it's link to internationalisation configuration apparently, but we don't find yet the problem...

mkllnk:
That translation is done by a plugin called `momentjs-rails`. Maybe check for differences in your application configuration `application.yml`. Or read the documentation of that plugin. There might be hints why it thinks it should display English.

myriamboure:
Ping @gnollet :-)

blancnic:
I think we need to add //= require moment/fr.js
https://github.com/derekprior/momentjs-rails

blancnic:
@mkllnk : don't you thing we could include this in a configuration file or define a all.js.coffee by country? Actually, we have to modify the js file each time we upgrade....

myriamboure:
Why is it working in staging then?
Is there the same issue in Norway @sigmundpetersen ? You only have a production instance, o you have to modify the js file each time you upgrade as well?

mkllnk:
all.js.coffee has already locale specific data, the I18n object with translations. If we could load other locale-dependent js-libraries in a similar way, that would be good. I don't think we want to include every locale in every instance. The all.js file would become too big and would negatively impact the user experience.

sigmundpetersen:
@myriamboure yes we have the same issue on the Norwegian instance

gnollet:
Hi,
I add the line //= require moment/fr.js on all.js.coffee and rebuild the assets.

That's fix the issue but we must think to this on each time we want to update platform.

mkllnk:
Are you deploying via `git push`?

You can put that into your deploy script `.git/hooks/post-receive` on the server. Something like:

```sh
echo '//= require moment/fr.js' >> "$CURRENT_PATH/app/assets/javascripts/darkswarm/all.js.coffee"
```

Just before the assets are rebuild. I'm not sure about the path. You need to check that. But it would do it automatically every time you deploy.

[...]

PR #1819, opened 2017-09-08
Switch momentjs internationalisation according to I18n locale

leandroalemao:
#### What? Why?

Hi 
This PR is linked to the #1107 
The issue is about momentjs internationalisation.
Always glad to help.
I hope all good.
Tks.
Leandro.

#### What should we test?

MomentJs stuff translated according to I18n locale.
#### Release notes

For now, I've just added internationalisation files for Spanish and French. If the PR is ok I can add for the other languages.

#### Dependencies

#1107


app/views/layouts/_i18n_script.html.haml
```diff
@@ -1,4 +1,15 @@
...
   I18n.default_locale = "#{I18n.default_locale}";
   I18n.locale = "#{I18n.locale}";
   I18n.fallbacks = true;
+
+  switch(I18n.locale) {
+    case "es":
+      moment.lang('es');
+      break;
+    case "fr":
+      moment.lang('fr');
```
sauloperez: can't we do `moment.lang(I18n.locale);`? AFAIK we both (moment and OFN) use the standard acronym for languages. As it is, it would require adding a bunch of case statements more. I can think of Portugal at least cc @mllocs.

app/assets/javascripts/admin/all.js
```diff
@@ -50,6 +50,8 @@
 //= require textAngular.min.js
 //= require i18n/translations
 //= require darkswarm/i18n.translate.js
-
+//= require moment
+//= require moment/es.js
+//= require moment/fr.js
```
sauloperez: Doesn't `require_tree` do the trick? If I'm not mistaken (I'm not very familiar with the Rails asset pipeline), it'd require all the files in `moment/`

enricostano:
Hi @leandroalemao many thanks for your precious time! I was just wondering if there is any way to avoid copy/pasting all that external code into OFN project and add it as a kind of dependency. What do you think? I'm not sure our `package.json` would be ready for that though.

cc/ @oeoeaio @mkllnk

mkllnk:
Our current package.json is just to install the dependencies for the karma tests. Shouldn't the assets come with the momentjs-rails gem? They are listed in the repository: https://github.com/derekprior/momentjs-rails/tree/v2.5.1/vendor/assets/javascripts/moment

I'm worried that `//= require moment/.` will bloat up the all.js file. Maybe we can use `config.i18n.available_locales`?

sauloperez:
Yep, if there's `momentjs-rails` gem that is the way.

mkllnk:
Yes, there is and we are using it already. https://github.com/openfoodfoundation/openfoodnetwork/blob/72d264fab9301435ac4ed9b3efa0d09fd96451f0/Gemfile#L84

daniellemoorhead:
Where are we at with this, gentlemen? It's marked as critical so would be good to get it done. Does it need further changes/code review/testing? Do you need anything from the AU team to get it through?

myriamboure:
@leandroalemao @sauloperez do you have time to finish that in the coming days ? Yes it's kind of critical as frontend... Would be awesome if we could finish this off :-p

leandroalemao:
Hi @all what do I need to include more in this PR? We're already using `momentjs-rails` gem and I just added a small javascript code to actually switch momentjs locate globally.. just following the doc http://momentjs.com/docs/#/i18n/. The other files are jus the localisation files (fr.js es.js.. etc..) .. which latest versions of, weren't working.. so I put a previous version of them. I guess because the momentjs version is not the latest as well. I hope all good. Tks. Leandro

sauloperez:
What I don't quite understand @leandroalemao is if we use the gem, aren't these localisation files included in it already? Do we need to upgrade its version? if so, let's do it.

sauloperez:
@leandroalemao do you need any help with this?

leandroalemao:
Hi @sauloperez Till yesterday I was working on the PR #1836. I'll have a look at this today. Tks. Leandro.

sauloperez:
Awesome @leandroalemao no pressure! I just thought we all forgot about this (myself included).

sauloperez:
I take this over.

According to https://github.com/derekprior/momentjs-rails in order to include a localization file `//= require moment/<locale>.js` needs to be added.

mllocs:
Could you diff the size of the js bundle where this is added? 🐘

sauloperez:
We go from 3.0MB to 3.1MB. That's already quite a lot.

sauloperez:
Watch out, I had to rebase to get the tests fixed.

daniellemoorhead:
@oeoeaio can you have a look at this one as a priority please. It's marked as critical.

myriamboure:
Yes @daniellemoorhead @oeoeaio I know I said I won't ask another issue to be reviewed in priority but when testing v1.10 I realized this was still not fixed, this is a front end untranslated thing on every French shopfront, so pretty annoying for our users... would be great if we can prioritize as well.

PR #1820, opened 2017-09-08
Missing translations in Enterprise Registration

saimonmoore:
#### What? Why?

The Enterprise Registration wizard was not entirely translated. This PR ensures that all the i18n keys are present (There will be some new keys to translate in Transiflex).

I also took the opportunity, to introduce decent i18n key namespacing for the Details, Contact and Type steps. There are some redundant keys, which are marked for removal after the new keys have been translated. 

#### What should we test?

- [x] Test Enterprise Registration Wizard

#### Release notes

Registration wizard is now completely translatable.

n/a

#### Discourse thread

n/a

#### Dependencies

n/a

### DIFF
--- a/app/views/registration/steps/_contact.html.haml
+++ b/app/views/registration/steps/_contact.html.haml
@@ -12,18 +12,18 @@
         .small-12.medium-12.large-7.columns
           .row
             .small-12.columns.field
-              %label{ for: 'enterprise_contact' } {{'enterprise_contact' | t}}:
-              %input.chunky.small-12.columns{ id: 'enterprise_contact', name: 'contact', required: true, placeholder: t(:enterprise_contact_placeholder), ng: { model: 'enterprise.contact' } }
+              %label{ for: 'enterprise_contact' } {{'enterprise.registration.modal.steps.contact.contact_field' | t}}:
+              %input.chunky.small-12.columns{ id: 'enterprise_contact', name: 'contact', required: true, placeholder: "{{'enterprise.registration.modal.steps.contact.contact_field_placeholder' | t}}", ng: { model: 'enterprise.contact' } }
               %span.error.small-12.columns{ ng: { show: "contact.contact.$error.required && submitted" } }
-                {{'enterprise_contact_required' | t}}
+                {{'enterprise.registration.modal.steps.contact.contact_field_required' | t}}
           .row
             .small-12.columns.field
-              %label{ for: 'enterprise_email_address' } {{'enterprise_email_address' | t}}:
-              %input.chunky.small-12.columns{ id: 'enterprise_email_address', name: 'email_address', type: 'email', placeholder: t(:enterprise_email_placeholder), ng: { model: 'enterprise.email_address' } }
+              %label{ for: 'enterprise_email_address' } {{'enterprise.registration.modal.steps.contact.email_field' | t}}:
+              %input.chunky.small-12.columns{ id: 'enterprise_email_address', name: 'email_address', type: 'email', placeholder: "{{'enterprise.registration.modal.steps.contact.email_field_placeholder' | t}}", ng: { model: 'enterprise.email_address' } }
           .row
             .small-12.columns.field
-              %label{ for: 'enterprise_phone' } {{'enterprise_phone' | t}}:
-              %input.chunky.small-12.columns{ id: 'enterprise_phone', name: 'phone', placeholder: t(:enterprise_phone_placeholder), ng: { model: 'enterprise.phone' } }
+              %label{ for: 'enterprise_phone' } {{'enterprise.registration.modal.steps.contact.phone_field' | t}}:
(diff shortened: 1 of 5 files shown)


app/views/registration/steps/_contact.html.haml
```diff
@@ -12,18 +12,18 @@
         .small-12.medium-12.large-7.columns
           .row
             .small-12.columns.field
-              %label{ for: 'enterprise_contact' } {{'enterprise_contact' | t}}:
-              %input.chunky.small-12.columns{ id: 'enterprise_contact', name: 'contact', required: true, placeholder: "Contact Name", ng: { model: 'enterprise.contact' } }
+              %label{ for: 'enterprise_contact' } {{'enterprise.registration.modal.steps.contact.contact_field' | t}}:
```
sauloperez: I understand this are Angular translations but did you consider using the Rails namespace convention here (`registration.steps`)? I know the key wouldn't be associated to enterprise but that would be consistent with other namespaces and would make it easy to reuse them in Rails (if we ever need to).

enricostano:
@saimonmoore many thanks! maybe you can create an issue to remember that we need to remove those keys in the future.

app/views/registration/steps/_contact.html.haml
saimonmoore: I did but I valued having a namespace that is easier to understand over that possibility. I also think it's unfortunate that /registration is the route. I would have preferred a more restful approach e.g. `enterprise.new.` and then use rails convention. So, tl dr; yes but no. :)

saimonmoore:
Yep.

saimonmoore:
https://github.com/openfoodfoundation/openfoodnetwork/issues/1821

btw...re the additional "steps: I thought were redundant. They're not. They are used as substeps so nothing needed to be removed.

daniellemoorhead:
@enricostano @sauloperez @oeoeaio can this be merged to master or does it need testing?

sauloperez:
Successfully tested in staging.katuma.org :tada: Can you please merge it @enricostano ?

lin-d-hop:
These changes appear to have broken the UK registration process. Only the lines changed here are broken. I have updated transifex. I have rebuilt assets. I have totally deleted and rebuilt assets. I have tried reverting and rebuilding assets (suggesting that something else is going on).

Any thoughts appreciated...

registration

lin-d-hop:
@saimonmoore @enricostano @oeoeiao

lin-d-hop:
Is it possible we need to add something like this...?
https://github.com/openfoodfoundation/openfoodnetwork/pull/1763/files

I'd test is myself but I just got a new laptop and haven't got OFN local dev env set up...

oeoeaio:
Suspect this is the same issue as #1858. Possibly an issue with `i18n-js` gem? We'll look into it.

lin-d-hop:
Thanks @oeoeaio and everyone for looking into this! :-D

saimonmoore:
Hi. I've been looking at this and the reason you are seeing this is because for some reason (still investigating) the fallback system of I18n-js is not working properly. Will update with for info.

lin-d-hop:
The fallback is a secondary thing. Yes, you are right it's not working and it would be awesome to solve!!

saimonmoore:
Ok I tracked it down to the fact that only the current locale's translations is being passed on to JS so despite having fallbacks enabled, there are no translations for the default locale and this is why the translations appear as missing. Looking into getting that fixed.

saimonmoore:
Ok so there's an underlying issue about fallbacks here in ofn (that I think basically makes them useless right now) that I won't get into but the basic issue here is that although the translations for en-GB are in transifex that haven't been updated in the repo so until that happens I'm afraid you will see this issue. Maybe there has been an issue with the CI process that updates the master branch from transifex?

My current knowledge about this process is very limited but as far as I can see this is the immediate issue.

lin-d-hop:
They have been. I deployed the transifex commit. Like I said, this is a side issue. And one it would be amazing to solve! If you could give the details of why defaults don't work that would be great!

⁣Sent from BlueMail ​

saimonmoore:
I'm preparing a separate PR for fallbacks which I'll post here too. I have just pulled master and although I see that the new copies have been translated to en-GB they are not present in master as of this post. Not sure what happened there (I understand that it's becoming a more pressing issue to automate the transifex/github process).

lin-d-hop:
We deploy the transifex branch, not the master branch. So there is some automation that works well.

saimonmoore:
Hi @lin-d-hop I see you just updated the en-GB locale with the translations. If you deploy that now it should work. As I mentioned I am working on setting up fallback to en to avoid this kind of issue again (so at least you see an english version until the translations are in place).

lin-d-hop:
Yep i just relealised that transifex updates one file and our instances uses another. Transifex updated en_GB. So that's a separate issue. Thanks for your help on this :-)

saimonmoore:
Here's the PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/1915

lin-d-hop:
Note there is still an issue for us :-( I'm going to delete some of these chatty comments clogging up the PR

PR #1831, opened 2017-09-14
Fix for decimal separator in prices

ltrls:
PR for #853 
It appeared that the problem was 2 problems: improve the way Spree handles separator in prices and make it work for more than prices.
Enrico found that the latter part was [fixed](https://github.com/spree/spree/commit/b817dab4a95c1beda3f18e42f0efcd2b7e64312a) in Spree 2.3 by having the behaviour (which was in Variant _and_ Price classes) in a LocalizedNumber class, so they could also use it with variants' weights.
Apparently what's important is that it works for prices, so I simply overrode the Spree methods, with a link to the commit in Spree and tests adapted from those in Spree.
When Spree dependency gets updated we'll be able to remove both methods by overriding the one in LocalizedNumber, and using it with things like fee amounts will be way easier.

### DIFF
--- a/app/assets/javascripts/admin/services/bulk_products.js.coffee
+++ b/app/assets/javascripts/admin/services/bulk_products.js.coffee
@@ -16,9 +16,9 @@ angular.module("ofn.admin").factory "BulkProducts", (PagedFetcher, dataFetcher, 
           @unpackProduct newProduct
           @insertProductAfter(product, newProduct)
 
-    updateVariantLists: (serverProducts, productsWithUnsavedVariants) ->
-      for product in productsWithUnsavedVariants
-        server_product = @findProductInList(product.id, serverProducts)
+    updateVariantLists: (serverProducts) ->
+      for server_product in serverProducts
+        product = @findProductInList(server_product.id, @products)
         product.variants = server_product.variants
         @loadVariantUnitValues product
 
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -3,6 +3,7 @@ require 'open_food_network/variant_and_line_item_naming'
 require 'open_food_network/products_cache'
 
 Spree::Variant.class_eval do
+  extend Spree::LocalizedNumber
   # Remove method From Spree, so method from the naming module is used instead
   # This file may be double-loaded in delayed job environment, so we check before
   # removing the Spree method to prevent error.
@@ -54,6 +55,8 @@ Spree::Variant.class_eval do
     joins("LEFT OUTER JOIN (SELECT * from inventory_items WHERE enterprise_id = #{sanitize enterprise.andand.id}) AS o_inventory_items ON o_inventory_items.variant_id = spree_variants.id")
       .where("o_inventory_items.id IS NULL OR o_inventory_items.visible = (?)", true)
   }
+
+  localize_number :price, :cost_price, :weight
 
   # Define sope as class method to allow chaining with other scopes filtering id.
   # In Rails 3, merging two scopes on the same column will consider only the last scope.
--- a/lib/spree/localized_number.rb
+++ b/lib/spree/localized_number.rb
@@ -1,0 +1,64 @@
+module Spree
+  module LocalizedNumber
+    # This method overwrites the attribute setters of a model
+    # to make them use the LocalizedNumber parsing method.
+    # It works with ActiveRecord "normal" attributes
+    # and Preference attributes.
+    # It also adds a validation on the input format.
+    # It accepts as arguments a variable number of attribute as symbols
+    def localize_number(*attributes)
+      validate :validate_localizable_number
+
+      attributes.each do |attribute|
+        setter = "#{attribute}="
+        old_setter = instance_method(setter) if table_exists? && !column_names.include?(attribute.to_s)
+
+        define_method(setter) do |number|
+          if Spree::LocalizedNumber.valid_localizable_number?(number)
+            number = Spree::LocalizedNumber.parse(number)
+          else
+            @invalid_localized_number ||= []
+            @invalid_localized_number << attribute
+            number = nil
+          end
+
+          if has_attribute?(attribute)
+            self[attribute] = number
+          else
+            old_setter.bind(self).call(number)
+          end
+        end
+
+        define_method(:validate_localizable_number) do
+          @invalid_localized_number.andand.each do |error_attribute|
+            errors.set(error_attribute, [I18n.t('spree.localized_number.invalid_format')])
+          end
+        end
+      end
+    end
+
+    def self.valid_localizable_number?(number)
+      return true unless number.is_a?(String) || number.respond_to?(:to_d)
+      return false if number =~ /[\.,]\d{2}[\.,]/
+      true
+    end
+
+    def self.parse(number)
+      return nil if number.blank?
+      return number.to_d unless number.is_a?(String)
+
+      number = number.gsub(/[^\d.,-]/, '') # Replace all Currency Symbols, Letters and -- from the string
+
+      if number =~ /^.*[\.,]\d{1}$/        # If string ends in a single digit (e.g. ,2)
+        number += "0"                      # make it ,20 in order for the result to be in "cents"
+      end
+
+      unless number =~ /^.*[\.,]\d{2}$/    # If does not end in ,00 / .00 then
+        number += "00"                     # add trailing 00 to turn it into cents
+      end
+
+      number = number.gsub(/[\.,]/, '')    # Replace all (.) and (,) so the string result becomes in "cents"
+      number.to_d / 100                    # Let to_decimal do the rest
+    end
+  end
+end
--- a/lib/spree/localized_number.rb
+++ b/lib/spree/localized_number.rb
@@ -9,19 +9,22 @@ module Spree
     def localize_number(*attributes)
       validate :validate_localizable_number
 
+      define_setter_methods(attributes)
+      define_validate_method
+    end
+
+    def define_setter_methods(attributes)
       attributes.each do |attribute|
         setter = "#{attribute}="
         old_setter = instance_method(setter) if table_exists? && !column_names.include?(attribute.to_s)
 
         define_method(setter) do |number|
-          if Spree::Config.enable_localized_number?
-            if Spree::LocalizedNumber.valid_localizable_number?(number)
-              number = Spree::LocalizedNumber.parse(number)
-            else
-              @invalid_localized_number ||= []
-              @invalid_localized_number << attribute
-              number = nil
-            end
+          if Spree::LocalizedNumber.valid_localizable_number?(number)
+            number = Spree::LocalizedNumber.parse(number)
+          else
+            @invalid_localized_number ||= []
+            @invalid_localized_number << attribute
+            number = nil
           end
 
           if has_attribute?(attribute)
@@ -30,13 +33,14 @@ module Spree
             old_setter.bind(self).call(number)
           end
         end
+      end
+    end
 
-        define_method(:validate_localizable_number) do
-          return unless Spree::Config.enable_localized_number?
-
-          @invalid_localized_number.andand.each do |error_attribute|
-            errors.set(error_attribute, [I18n.t('spree.localized_number.invalid_format')])
-          end
+    def define_validate_method
+      define_method(:validate_localizable_number) do
+        return unless Spree::Config.enable_localized_number?
+        @invalid_localized_number.andand.each do |error_attribute|
+          errors.set(error_attribute, [I18n.t('spree.localized_number.invalid_format')])
         end
       end
     end
@@ -53,16 +57,18 @@ module Spree
 
       number = number.gsub(/[^\d.,-]/, '') # Replace all Currency Symbols, Letters and -- from the string
 
-      if number =~ /^.*[\.,]\d{1}$/        # If string ends in a single digit (e.g. ,2)
-        number += "0"                      # make it ,20 in order for the result to be in "cents"
-      end
-
-      unless number =~ /^.*[\.,]\d{2}$/    # If does not end in ,00 / .00 then
-        number += "00"                     # add trailing 00 to turn it into cents
-      end
+      add_trailing_zeros(number)
 
       number = number.gsub(/[\.,]/, '')    # Replace all (.) and (,) so the string result becomes in "cents"
       number.to_d / 100                    # Let to_decimal do the rest
+    end
+
+    def self.add_trailing_zeros(number)
+      # If string ends in a single digit (e.g. ,2), make it ,20 in order for the result to be in "cents"
+      number << "0" if number =~ /^.*[\.,]\d{1}$/
+
+      # If does not end in ,00 / .00 then add trailing 00 to turn it into cents
+      number << "00" unless number =~ /^.*[\.,]\d{2}$/
     end
   end
 end
--- a/lib/spree/localized_number.rb
+++ b/lib/spree/localized_number.rb
@@ -14,12 +14,14 @@ module Spree
         old_setter = instance_method(setter) if table_exists? && !column_names.include?(attribute.to_s)
 
         define_method(setter) do |number|
-          if Spree::LocalizedNumber.valid_localizable_number?(number)
-            number = Spree::LocalizedNumber.parse(number)
-          else
-            @invalid_localized_number ||= []
-            @invalid_localized_number << attribute
-            number = nil
+          if Spree::Config.enable_localized_number?
+            if Spree::LocalizedNumber.valid_localizable_number?(number)
+              number = Spree::LocalizedNumber.parse(number)
+            else
+              @invalid_localized_number ||= []
+              @invalid_localized_number << attribute
+              number = nil
+            end
           end
 
           if has_attribute?(attribute)
@@ -30,6 +32,8 @@ module Spree
         end
 
         define_method(:validate_localizable_number) do
+          return unless Spree::Config.enable_localized_number?
+
           @invalid_localized_number.andand.each do |error_attribute|
             errors.set(error_attribute, [I18n.t('spree.localized_number.invalid_format')])
           end
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -32,7 +32,6 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     $scope.filteredProducts = []
     $scope.currentFilters = []
     $scope.limit = 15
-    $scope.productsWithUnsavedVariants = []
     $scope.query = ""
     $scope.DisplayProperties = DisplayProperties
 
@@ -114,7 +113,6 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
         display_name: null
         on_hand: null
         price: null
-      $scope.productsWithUnsavedVariants.push product
       DisplayProperties.setShowVariants product.id, true
 
 
@@ -196,7 +194,7 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
           filters: $scope.currentFilters
       ).success((data) ->
         DirtyProducts.clear()
-        BulkProducts.updateVariantLists(data.products, $scope.productsWithUnsavedVariants)
+        BulkProducts.updateVariantLists(data.products || [])
         $timeout -> $scope.displaySuccess()
       ).error (data, status) ->
         if status == 400 && data.errors? && data.errors.length > 0
--- a/app/models/calculator/flat_percent_per_item.rb
+++ b/app/models/calculator/flat_percent_per_item.rb
@@ -1,4 +1,5 @@
 require_dependency 'spree/calculator'
+require 'spree/localized_number'
 
 class Calculator::FlatPercentPerItem < Spree::Calculator
   # Spree's FlatPercentItemTotal calculator sums all amounts, and then calculates a percentage
@@ -6,10 +7,13 @@ class Calculator::FlatPercentPerItem < Spree::Calculator
   # In the cart, we display line item individual amounts rounded, so to have consistent
   # calculations we do the same internally. Here, we round adjustments at the individual
   # item level first, then multiply by the item quantity.
+  extend Spree::LocalizedNumber
 
   preference :flat_percent, :decimal, :default => 0
 
   attr_accessible :preferred_flat_percent
+
+  localize_number :preferred_flat_percent
 
   def self.description
     I18n.t(:flat_percent_per_item)
--- a/app/models/spree/adjustment_decorator.rb
+++ b/app/models/spree/adjustment_decorator.rb
@@ -1,5 +1,9 @@
+require 'spree/localized_number'
+
 module Spree
   Adjustment.class_eval do
+    extend Spree::LocalizedNumber
+
     # Deletion of metadata is handled in the database.
     # So we don't need the option `dependent: :destroy` as long as
     # AdjustmentMetadata has no destroy logic itself.
@@ -16,6 +20,8 @@ module Spree
     scope :payment_fee,     where(originator_type: 'Spree::PaymentMethod')
 
     attr_accessible :included_tax
+
+    localize_number :amount
 
     def set_included_tax!(rate)
       tax = amount - (amount / (1 + rate))
@@ -73,6 +79,5 @@ module Spree
 
       result
     end
-
   end
 end
--- a/app/models/spree/app_configuration_decorator.rb
+++ b/app/models/spree/app_configuration_decorator.rb
@@ -42,4 +42,7 @@ Spree::AppConfiguration.class_eval do
 
   # Stripe Connect
   preference :stripe_connect_enabled, :boolean, default: false
+
+  # Number localization
+  preference :enable_localized_number?, :boolean, default: false
 end
--- a/app/models/spree/calculator/flat_percent_item_total_decorator.rb
+++ b/app/models/spree/calculator/flat_percent_item_total_decorator.rb
@@ -1,5 +1,11 @@
+require 'spree/localized_number'
+
 module Spree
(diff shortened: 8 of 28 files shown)


myriamboure:
I see that some tests are not passing @ltrls but @enricostano approved the changes, is it ready for Aus Review or not yet ?

ltrls:
Should I follow the Codeclimate advices? They don't seem very relevant

Matt-Yorkley:
@ltrls With the current setup, PRs can't be merged or tested until all Codeclimate issues are resolved.

ltrls:
Oh OK! I thought it was just indicative.

app/models/spree/variant_decorator.rb
```diff
@@ -128,4 +128,21 @@ def destruction
...
+    price = price.gsub(/[^\d.,]/, '')  # Replace all Currency Symbols, Letters and -- from the string
+    if price =~ /^.*[\.,]\d{1}$/       # If string ends in a single digit (e.g. ,2)
+      price += "0"                     # make it ,20 in order for the result to be in "cents"
+    end
+    unless price =~ /^.*[\.,]\d{2}$/   # If does not end in ,00 / .00 then
+      price += "00"                    # add trailing 00 to turn it into cents
+    end
+    price = price.gsub(/[\.,]/, '')    # Replace all (.) and (,) so the string result becomes in "cents"
+    price.to_d / 100                   # Let to_decimal do the rest
+  end
```
sauloperez: I'd rather copy the whole `Spree::LocalizedNumber` class instead of copying the method two times. The problem I see is that we will eventually fix something in `variant_decorator.rb` but we'll forgot to also apply it in the `price_decorator` and viceversa.

ltrls: I thought of that, but I wasn't sure if it was a great idea to get ahead of the Spree upgrade. When doing the upgrade we'll probably want to remove what we added that become unnecessary (calls to `Spree::LocalizedNumber.parse`, use of `require`). Do you think it would be simpler?

sauloperez: > we'll probably want to remove what we added that become unnecessary

ltrls: Fix'd

app/models/spree/price_decorator.rb
```diff
@@ -1,7 +1,14 @@
+require 'spree/localized_number'
+
 module Spree
   Price.class_eval do
     after_save :refresh_products_cache
 
+    # this method override can be removed when updating to Spree 2.3
```
sauloperez: Can you use the `# TODO:` tag? This way it'll be easier to find when the time comes

app/models/spree/variant_decorator.rb
```diff
@@ -103,6 +104,16 @@ def refresh_products_cache
     end
   end
 
+  # this method override can be removed when updating to Spree 2.3
+  def cost_price=(price)
+    self[:cost_price] = Spree::LocalizedNumber.parse(price) if price.present?
+  end
+
+  # this method override can be removed when updating to Spree 2.3
```
sauloperez: Same for these

app/models/spree/price_decorator.rb
ltrls: sure! there it is

daniellemoorhead:
@mkllnk not sure how to test this, is it something that @sstead or @simoneluijckx can do?

myriamboure:
I think so @daniellemoorhead they can try to enter amounts with , and . indifferently and wth one or two decimals, all should work properly :-) We tested already on french staging so for us it seems fine. The only thing we need to be aware of is that if you type three decimals (which  never happens when you type a price) then the systems interprets the . or , as a thousand separator... but as we said prices are never showed with three decimals so we think that's an acceptable compromise.

daniellemoorhead:
@sstead can you get somebody to stage this for you so you can test it?

sstead:
This PR is difficult to test because there are not very clear testing instructions. 
- There's no description of desired behaviour?
- Does this fix relate to Bulk Product Edit and/or Inventory?
- Does this fix depend on the instance's currency? Is that something that needs testing?

Testing notes...
**In Bulk Product Edit**
The screenshot below shows how an entered price is converted. I've highlighted the ones that appear correct in green.
The cases in orange should have an error message in Australia, because the arrangement of '.' and ',' do not make sense. Perhaps people from other countries can comment whether the orange formats and conversions make sense, or if an error message would be better? 
image

**In Inventory**
- The treatment of decimals is different in Inventory relative to Products... I'd suggest we should fix Inventory as well so there is consistency.
image

Other notes:
- In Inventory, the moment the user inputs a price, the system reformats it if required. This is helpful because the user is more likely to see the conversion, and they can check that that is the format they intended. In Bulk Product edit, you need to refresh the page to see the conversion, which means it's less likely the user will see the reformatted price and be able to check that it's what they want.

- Handling .1c has changed...
image

-On the Australia Staging instance all ',' get either removed, or replaced with a '.' This is ok in Australia because we only use ',' in thousands, but will this also occur in European countries where they use ',' to split dollars and cents? Will these instances prefer that ',' is used?

myriamboure:
Hi @sstead and thanks for testing.
I'm not sure I understand all what you tested, for me we are supposed to test only "logical user behavior", as we said for instance one of the assumption is that prices are entered with max 2 decimals (is the .9c = 0.009$a real life case ?)
The thing is that if we don't assume that, we have no way to find a solution that will fit all as we have no way ffor the system to understand is . separates thousands or decimals in that case for instance. So we can't know if we have to round because it's a .9 case, or if it's a thousand separator...

In the bulk edit cases you tested, I think only the top right two ones (2.000,0 and 2.000,00) are real life ones not working (in Indian context for instance they use . for thousands separator and , for decimals). For the other orange ones I don't see any real life case attach, do you confirm ?
If we want to get those cases as an error we shoud have some complementary rule like : if two "." in a price display "unvalid format" message, or if price finishes with a "."
Would that be a solution ?

I agree with you that it would be great to refresh real time the amount like in inventory.
And of course the behavior should be the same in inventory... @ltrls can you check that please ?

In Europe the "," separate euros and cents, so we can't at all remove it. It is acceptable for us that it is changed in a . that people understand as well as a euros / cents separator. But that won't work in an Indian context as . are used for thousands separators.

myriamboure:
Happy next time to write testing notes, sorry I had never do it but I get the point of course ;-) Is it the product owner / spec people to do it ? Or the dev ? I guess the first one, right ?

myriamboure:
CORRECTION: the rules I thought about won't work as 2.000.000 can be used to talk about 2 millions... so there can be two "." in a price ...  any other idea how to handle that ? @enricostano @sauloperez @augnustin ?

myriamboure:
@sstead one thing is very unclear to me in your tests actually...
why are those corrects:
- 2,00.00 it shouldn't be 200 ? For me this shoud be an invalid format, not ? you cannot have 2 digits between two . or between a , and a . right ? 
- 200,30. it shouldn't be 20030 ? Either the , is a thousand separator, either a decimal separator, but in that case also this is invalid format ?

For me the rules are:
- if there is only one or two digit between either two . or one , and one . = invalid format

I hope there is no incompatible format between Aus and India for instance, I'm a bit afraid when I see that you report 2.000,00 = 2000.0 as false...

sstead:
Sorry I've not done a very good job of testing this, because I was confused about desired behaviour... I hope I haven't confused you also. I've just put my 'Australian' hat on, to see if it makes sense here, not really knowing what others need.

**"We said that we had to make the assumption that prices are entered with max 2 decimals (is the .9c = 0.009$ a real life case ?)"**
 It's unlikely that a user will type $5.009 but, in Australia it is not logical that this becomes five thousand ($5,009.00), it would be logical to become $5.01. We can compromise on this though, because it's rare.

**"In the bulk edit cases you tested, I think only the top right two ones (2.000,0 and 2.000,00) are real life ones, and they seem fine to me, why did you put in orange ?"**
In Australia 2.000,0 and 2.000,00 are meaningless- I'd expect to see an error. It's very rare that someone would type this so we can compromise, because in India this is not an error.

"For the other orange ones I don't see any real life case attach, do you confirm ? If we want to get those cases as an error we should have some complementary rule like : if two "." in a price display "unvalid format" pop-in message + empty field, or if price finishes with a ".""
Yes, the orange ones are user mistakes in Australia - they don't mean anything, so it would be best to have an error message. Yes I agree, 'if 2 '.' in a price display error'. But if price finishes with a "." it's not an error, just add '00' after.

**I agree with you that it would be great to refresh real time the amount like in inventory** This would make the compromises above much better, because the user would see the transformation and be able to amend it.

**2,00.00 it shouldn't be 200 ? For me this should be an invalid format, not ? you cannot have 2 digits between two . or between a , and a . right ?**
Yes you're right, it's invalid because you can't assume the user means $200 or $2000. An error message would be better for us. 

**200,30. it shouldn't be 20030 ? Either the , is a thousand separator, either a decimal separator, but in that case also this is invalid format ?**
Yes, it would be ok in Australia if this gets an error message, because technically it's not meaningful. But it would also be acceptable to just remove the , (because it's in the wrong place) and make it $20030.00.

**if there is only one or two digit between either two . or one , and one . = invalid format**
yes, this makes sense in Australia.

**I hope there is no incompatible format between Aus and India for instance, I'm a bit afraid when I see that you report 2.000,00 = 2000.0 as false...**
2.000,00 - this is meaningless in Australia. Nobody would type this, so I think it's ok for it to become $2000

myriamboure:
Ok, thank you @sstead it seems we are on the same page then ;-) @ltrls just need to find a way to get those "invalid format" error messages  working.
So @ltrls to sum up:
- dynamic update in product page : I think it's done from what I tested yesterday
- format apply in inventory : I think this is aso OK
- we tested yesterday the adjustments and shipping/payment/entreprises fees : there is still work to be done for that formatting to work there
- and we need to add the following rule : if an amount show one or two digits between two "." or between a "," and a "." show "unvalid format" popin. Leave the wrong amount like it is but just, don't save and ideally, color the field borders in red so that the user knows what field to correct.

Hope this is clear ;)

ltrls:
This has changed quite a bit, I'll try to explain why it became more complex:
The simple step of making the setters use the `parse` method, which I
originally implemented [like
this](https://github.com/openfoodfoundation/openfoodnetwork/pull/1831/commits/15976eb6a3ceaa8781857e192246105e840b4e57#diff-e09324be17b9368e6138d062a552e0cfR7)
had to work with setters which don't set an ActiveRecord attribute but a
[Preference attribute](https://github.com/openfoodfoundation/spree/blob/master/core/app/models/spree/preferences/preferable_class_methods.rb#L24) ([like
here](https://github.com/openfoodfoundation/spree/blob/master/core/app/models/spree/calculator/per_item.rb#L5)).
When refactoring this I found that I also had to handle the case where the
setter and the attribute had a different name ([like
here](https://github.com/openfoodfoundation/openfoodnetwork/pull/1831/commits/15976eb6a3ceaa8781857e192246105e840b4e57#diff-e09324be17b9368e6138d062a552e0cfR7)).
I started refactoring and taking everything that I added, out of the models
when I was asked to implement errors. We can't add errors when setting
attributes so I had to add a validation and I thought it was too much code to
add everywhere we would use our custom setter (so far it's in 9
models!). I found that I could do it in a classic ActiveRecord way. I'm not a
big fan of these metaprogramming things (it really is harder to test, and I
need to dig the "hygiene" of these macros, do my variables leak and should I
namespace them, prefix with underscores?) but I couldn't find any other way.
I had a hard time making the errors work in some places, but we thought with
Myriam that it could be best to have this merged and work on displaying the
possible error everywhere later. Here are places where errors on model
instances wouldn't work out of the box:
- When updating multiple records at once using ModelSet (in products bulk edit,
  variant overrides, enterprise fees), if the only modified attribute of the
  records is one which is supposed to get localized but is wrongly formatted
  (the attribute is set to nil and an error is added), the record would get
  discarded because the @delete_if callback
  [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/models/model_set.rb#L43-L53)
  checks if the attributes are nil so the records are discarded even if they
  have errors. I found this as a workaround but I don't find it very pretty:
```patch
diff --git a/app/models/model_set.rb b/app/models/model_set.rb
index cb75f929a..b1863ab75 100644
--- a/app/models/model_set.rb
+++ b/app/models/model_set.rb
@@ -44,12 +45,12 @@ class ModelSet
     # Remove all elements to be deleted from collection and return them
     # Allows us to render @model_set.collection without deleted elements
     deleted = []
-    collection.delete_if { |e| deleted << e if @delete_if.andand.call(e.attributes) }
+    collection.delete_if { |e| deleted << e if @delete_if.andand.call(e.attributes) && e.valid? }
     deleted
   end
 
   def collection_to_keep
-    collection.reject { |e| @delete_if.andand.call(e.attributes) }
+    collection.reject { |e| @delete_if.andand.call(e.attributes) && e.valid? }
   end
 
   def persisted?
```
- [Here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/admin/enterprise_fees_controller.rb#L41)
  if the ModelSet doesn't save, the controller won't show the errors (and will
  try to render a view instead of redirecting to another action, the view
  will need variables set in the `before_filter` of its action so it will
  fail)
- Last but not least, Variants use the spree custom `delegate_belongs_to`
  thing
  ([here](https://github.com/openfoodfoundation/spree/blob/master/core/app/models/spree/variant.rb#L31))
  to set their price in another model. Unlucky for us, the errors in the
  delegated model aren't fed through to the main one (`variant.price =
  "1,23.4"` => variant is `valid?` although its price object is not) (may be
  related to [this
  comment](https://github.com/openfoodfoundation/spree/blob/master/core/lib/spree/core/delegate_belongs_to.rb#L5)).
  It seems like in Solidus they stopped using `delegate_belongs_to` and used
  Rails own `delegate` instead.

myriamboure:
**Testing notes:**

**- Bulk edit products:**
testing notes

**- Inventory:**
testing notes 1

**- Shipping method fee AND payment method fee (same results):**
testing notes 2

**- Enterprise fees:**
testing notes 3
> This needs to be fixed before it can be merged (if complicate we will create separate issue in order not to block merging that improves already the user experience)

**- Adjustments:**
testing notes 4

**- Payments:**
testing notes 6
> This needs to be fixed before it can be merged (if complicate we will create separate issue in order not to block merging that improves already the user experience)

myriamboure:
For info the two issues opened:
https://github.com/openfoodfoundation/openfoodnetwork/issues/1970
and
https://github.com/openfoodfoundation/openfoodnetwork/issues/1971

daniellemoorhead:
Hey @ltrls and @myriamboure given this comment above:

>This has changed quite a bit, I'll try to explain why it became more complex:

do you think that we should do a final code review before merging? @oeoeaio thoughts from you on this?

oeoeaio:
Hi @myriamboure. @sauloperez and @enricostano and @mkllnk are all probably better placed to have a look at this than I am, but they are all out of action for at least another week. If this is high priority for you, I can prioritise reviewing this and getting it through if you would like. Let me know what you'd like to do.

ltrls:
> Hi @myriamboure. @sauloperez and @enricostano and @mkllnk are all probably better placed to have a look at this than I am, but they are all out of action for at least another week. If this is high priority for you, I can prioritise reviewing this and getting it through if you would like. Let me know what you'd like to do.

From what I understood talking with @myriamboure it seems to be the case! I've been making the tests cleaner using rspec's shared examples, it's pretty closed to work, I should push it soon (like in a few minutes if it goes well!).
I know I've been using metaprogramming stuff that we're usually warned about, but I think that I couldn't do otherwise and that I used it "softly" (I think I've seen this sort of things in a few libraries). I hope you'll find it nice.
and it will be nicely tested!

ltrls:
OK it simplified the tests a lot, doing this I found that the variant and price thing could be solved by simply extending the variant for its (delegated) price attribute instead of extending the price model, and it simplified things also in the LocalizedNumber module as I didn't have to handle the price `amount` attribute with its `price=` setter.
anyway I really thought it would take a few minutes..!

myriamboure:
@oeoeaio @daniellemoorhead yes this reaaaaally needs to go through for next release, we can't wait one more month, it's been more than a year we tell our users we gonna solve it ! We got so many feedback from user, it is a very basic use of the platform that is very buggy (fill in a price) and end up with products with wrong prices in some shopfronts... much more important for us than any new feature.
I have tested carefully as you can see in the testing notes so I don't thing it needs more testing, I guess if you can code review Rob to check that on the back side all is good, then it's good to merge.

myriamboure:
Ok I retested again @oeoeaio @daniellemoorhead everything after Pierre's last modifications so it's all good on testing side, ready to merge !
Comment for @ltrls : the error messages still don't work on bulk edit and inventory but I opened 2 separate issues so that we can address them on a second step and that don't block the problem to be solved.

daniellemoorhead:
>I have tested carefully as you can see in the testing notes so I don't thing it needs more testing, I guess if you can code review Rob to check that on the back side all is good, then it's good to merge.

Definitely don't feel the need to test this again @myriamboure - you're an amazing tester, your attention to detail is superb 👌 

@oeoeaio would be great if we could get this reviewed and merged before the next release (which is slated for next week).

app/assets/javascripts/admin/services/bulk_products.js.coffee
```diff
@@ -16,9 +16,9 @@ angular.module("ofn.admin").factory "BulkProducts", (PagedFetcher, dataFetcher,
           @unpackProduct newProduct
           @insertProductAfter(product, newProduct)
 
-    updateVariantLists: (serverProducts, productsWithUnsavedVariants) ->
-      for product in productsWithUnsavedVariants
-        server_product = @findProductInList(product.id, serverProducts)
+    updateVariantLists: (serverProducts) ->
+      for server_product in serverProducts
+        product = @findProductInList(server_product.id, @products)
```
oeoeaio: Hi @ltrls, this means that we will be iterating through every single product, rather than just those that have been changed. Is there any particular reason why this is necessary?

sstead:
@oeoeaio have I missed any number formats?
image

oeoeaio:
Hi @ltrls @myriam.

Unfortunately there are a couple of issues that Sally, Kirsten and I think are blockers to this being merged.

The main problems relate to entering an additional decimal place after the `.` we use as a decimal separator in Aus. For example:
a) `1.255` is interpreted as `1255`, it should be `1.26`.
b) `.023` is interpreted as `23`, it should be `0.02`

We can probably live with everything else, but this is something that really doesn't work for us. Obviously our requirements directly contradict yours in France. Therefore, I think our only option is to make this configurable, and allow instances to opt-in to this behaviour, rather than applying it universally.

@ltrls, what do you think about that idea? I think your logic is really well designed. I like the approach, and it seems to me that it wouldn't take a lot of work to add a feature toggle that preserved the existing behaviour where that is what was desired.

myriamboure:
@oeoeaio @kirstenalarsen @sstead on the 6 october I explained "We said that we had to make the assumption that prices are entered with max 2 decimals (is the .9c = 0.009$ a real life case ?)
The thing is that if we don't assume that, we have no way to find a solution that will fit all as we have no way for the system to understand if . separates thousands or decimals in that case for instance. So we can't know if we have to round because it's a .9 case, or if it's a thousand separator...
Is this convention acceptable to you ?"
And on the 13th october @sstead answered "It's unlikely that a user will type $5.009 but, in Australia it is not logical that this becomes five thousand ($5,009.00), it would be logical to become $5.01. We can compromise on this though, because it's rare."

**So again, is this ever happening ? Does anyone use prices with 3 digits in Aus ? Have we ever seen a price with 3 decimals in a shop ? Do we talk of hypothetical logic or real life situations ?**

I'll discuss with @ltrls today to see what he thinks and if this opt-in module can be added and tested by tomorrow. I hope we find a way to have this merged before Friday, it's been monthes that our users are fighting with prices displaying 350€ for 1 kg of carrots, instead of 3,50€...

oeoeaio:
Hey @myriamboure,

I'm sorry that I am not across the detail of the conversation you've highlighted. This is the first time I have looked at this work, and I have no knowledge of previous discussions.

The issue isn't that there is a use-case for three digit prices. It's a user-experience issue.  The intention of someone who accidentally enters three decimals is very clear, and in this case we should just drop or round out the last digit. Converting `3.500` to `$3,500` instead of `$3.50` would be very confusing to most people in Australia. At the moment, the input is rounded to the nearest cent. Therefore, I feel that putting this change through as it stands would represent a backwards step.

I completely 100% understand you need to get this through as soon as possible. We have every intention of releasing this on Friday, and I will do anything I can to help make sure that happens.

oeoeaio:
@myriamboure One more thing. Do you need people to be able to enter prices in both formats? If not, are you aware of anyone who does? It seems to me that it would be far better to allow the instance manage to set things up to work in one way OR the other. Trying to get both to work at the same time is always going to be very challenging.

myriamboure:
I know you are doing an awesome job @oeoeaio and thank you for your clear explanation, I understand better now why this is an issue for you, it's more "ho you treat a mistake enter by a user that is logical in the Aus logic", which I understand. Thank you for your understanding :-)
 On your last point, I disagree, I alway fight with google spreadsheet for instance as I naturally tend to type in decimals with a "," and it is not recognized. In Spree today, if I type a "," it is displayed then with a "." so we naturally tend to type in decimals with a "." even if our native behaviour is a ","... so I think actually we need to allow both behaviours, in terms of UX. And ... imagine multilingual / multicountries instances ;-)

oeoeaio:
👍  ok, no problems. Is there an example of some software that you use that handles input in both formats really well? Would be good to make sure that we exactly mimic some good behaviour if you know of some. (I'm not suggesting that we do this as part of this PR, just that it would be good to do eventually 😉 ).

ltrls:
Agreed for the instance switch, it's true that this change makes a lot more sense if the regional convention and the international (English) one aren't the same.
(this CodeClimate thing is driving me crazy, it feels like I'm making my code less readable for humans because of how a machine considers code to be readable ><)

myriamboure:
@oeoeaio @sstead @kirstenalarsen : @ltrls is a bit fighting with Code climate, it took him 1hour to do the opt-in fix, but 3 hours trying to cope with the code climate things, as his code is apparently lisible by humans but he has to make it less clear so that it can meet Code Climate standards... I don't know anything about that but it's a bit frustrating to loose so much time on that, and I hope he can manage to finish that today so that we merge that for the release tomorrow... 

Just a comment as well : when we tested the opt-in thing, I realize how the current master is also buggy in some situations, even for Australian standards. For instance on inventory, if you type "2,500.50" and save it becomes "2.0"
screenshot from 2017-11-30 08-57-11
screenshot from 2017-11-30 08-57-27
There are also issues when filling in an adjustement with thousands.

Anyway, I hope Pierre can fin a way to manage with code climate today... it's a bit frustrating to spend 1h on the feature and 5 on making it enter the box... "quick fixes" are often not so quick unfortunately...
Else I would happily suggest that we merge it without the opt-in box tonight so that we can at still have the fix and merge the opt-in option for the next release...

myriamboure:
@oeoeaio it seem @ltrls spent some hours again ttoday rying to find ways to make code climate accept the complexity of his commit but he couldn't make it ... i don't know how this can be solved, but if you have an idea that would be awesome ...

oeoeaio:
Hey @myriam, I'll have a look at it ASAP. We can always just mark individual issues as #WONTFIX, which means CodeClimate will ignore the issue. In future if a lot of time is being spent fighting codeclimate and ending up with less readable code, we can just do this instead. Not worth wasting precious time on it.

@ltrls, the best way to solve complexity offences is generally to split logic out into separate smaller methods.

lib/spree/localized_number.rb
```diff
@@ -30,6 +32,8 @@ def localize_number(*attributes)
         end
 
         define_method(:validate_localizable_number) do
+          return unless Spree::Config.enable_localized_number?
+
```
oeoeaio: Hey @ltrls, thanks for adding this. I am just wondering, it there any reason that we can't return at the top of the `localize_number` method if `Spree::Config.enable_localized_number?` is `false`? I would prefer to not override any attribute setters unless it is necessary...

ltrls: yeah but the `localize_number` method is called when model classes are loaded by Ruby, so I think the Config switch would be effective only when the application is restarted, right? I mean, if I overwrite the setters and in the same runtime I decide to deactive the separator thing, then they wouldn't be un-overwritten before the next restart I think. It's not a big deal as it's not meant to be switched every now and then though...

oeoeaio: Yeah, I think this is a "set and forget" kind of configuration. I don't think we need to handle toggling it on and off frequently. 

This is related: how do you feel about configuring this via `application.yml` rather than through the admin interface? Pau and I think that maybe the interface is overkill, given that it won't need to be changed often, and it is probably a sysadmin task rather than a instance manager task? I am happy to do the work, but I wanted to check with you first.

oeoeaio: @ltrls ☝️

ltrls: yes that would be nice. That would avoid the confusion of an administrator (not sysadmin) configuring something that says "you need to restart the Rails application"..!

lib/spree/localized_number.rb
```diff
@@ -30,13 +33,14 @@ def localize_number(*attributes)
...
+    end
 
-        define_method(:validate_localizable_number) do
-          return unless Spree::Config.enable_localized_number?
-
-          @invalid_localized_number.andand.each do |error_attribute|
-            errors.set(error_attribute, [I18n.t('spree.localized_number.invalid_format')])
-          end
+    def define_validate_method
+      define_method(:validate_localizable_number) do
```
sauloperez: I wonder if we're redefining the method every time this `LocalizedNumber` module is extended, that would be a pointless overhead.

lib/spree/localized_number.rb
```diff
@@ -30,13 +33,14 @@ def localize_number(*attributes)
...
+        return unless Spree::Config.enable_localized_number?
```
sauloperez: Rather than early-returning shouldn't we skip all this module's behavior when number localization is not enabled?

oeoeaio:
Merged via #1982. Closing.

Issue #1718, opened 2017-07-22
The 'show profiles' filter in the Shops directory is not working


sstead:
Expected functionality: By default the Shops page only shows shops that have a currently open order cycle. Profile-only hubs don't show unless the user clicks 'show profiles'. On Aus production this isn't happening, clicking the checkbox does nothing.

![image](https://user-images.githubusercontent.com/10843056/28487690-9cc8ee8e-6ed9-11e7-809d-5db0fc5e60c9.png)

daniellemoorhead:
@kirstenalarsen this is a pretty bad experience for our users. Should we get it up the to do list and paid for out of AU funds?

kirstenalarsen:
do we have a milestone or project or something that is ‘aus priorities’? If so, can we put it there and discuss whether there are any ‘aussie funds’ w serenity next week?

PR #1835, opened 2017-09-15
1718 show profiles

mkllnk:
#### What? Why?

The filter for closed shops was filtering profiles as well. That made it
impossible to show profiles without showing closed shops as well.
    
Replacing the filter for closed shops fixes the "show profiles" feature.
    
Fixes https://github.com/openfoodfoundation/openfoodnetwork/issues/1718

#### What should we test?

The shops page, the producers page and the group pages.

#### Release notes

The "show profiles" feature now also works without showing closed shops.


sstead:
Testing didn't reveal any problems

Issue #1783, opened 2017-08-30
Embedding group pages


sstead:
Users want to be able to embed group pages into their own websites. This feature should be based on the existing work around [embedding shopfronts](https://github.com/openfoodfoundation/openfoodnetwork/pull/1557), with some modifications.

@mkllnk I understand the plan is that you should spend 2 hours on this, and see how far we can get. I've listed some 'nice to have' elements, but it's ok if they don't get done in the first version.

**Group page functionality to keep the same:** (essential)
- Existing map search and zoom
- Existing search and filters on the profiles and hubs tabs
- Expando collapso of the mini profiles in Our Producers and Our Hubs

**Display of profile modals** (not essential)
When a visitor clicks on a profile icon on the map, the modal should open. Ideally the background of the webpage (not just the embedded section) should be greyed (as pictured below). The webpage scroll bar is also disabled, so the user can just scroll in the modal.
![image](https://user-images.githubusercontent.com/10843056/29848884-b5f2c5d2-8d66-11e7-8a32-11cbf5e0701d.png)
If this is not possible, just the embedded section should be greyed, as happens if you open the product modal in an embedded shop - see [here ](https://openfoodnetwork.org/user-guide/advanced-features/demo-embedded-shop/)

**Embedded header** (not essential)
- I'd suggest that the Login (or gear symbol if already logged in) and Cart links could be removed from the header- I imagine the directory will primarily be for the purpose of browsing profiles, not shopping. Therefore we don't need these functions to be at the top, where they'll potentially confuse people.
- Could we move the 'powered by open food network' to somewhere lower on the page? It takes up a whole row of header space, which is probably not neccessary.
![image](https://user-images.githubusercontent.com/10843056/29849036-a8c9e920-8d67-11e7-9306-352fdb32173b.png)

**Contact details** (not essential)
If possible it would make sense to remove the contact details 
![image](https://user-images.githubusercontent.com/10843056/29849252-e194cc92-8d68-11e7-9eae-04d5610b4880.png)

**Remove the logo, banner and Description text fields.** (not essential)
Because the group page is being embedded, the group owner can add whatever images and text they want above their embedded group. Therefore if we remove these fields, it gives the group owner more flexibility to decide what sits above the directory.
![image](https://user-images.githubusercontent.com/10843056/29849362-95a01700-8d69-11e7-846b-7e8289555d9f.png)

**Links to Shops.** (not essential)
[...]

PR #1841, opened 2017-09-20
1783 embedded groups

mkllnk:
#### What? Why?

First iteration of #1783.

It allows group pages to be shown in an iframe. There is a test page to preview groups at `/embedded-group-preview.html?group-name`. I have not added additional specs.

List of Sally's suggestions and what's done:

- [x] the Login (or gear symbol if already logged in) and Cart links could be removed from the header
- [ ] move the 'powered by open food network' to somewhere lower on the page (*the element got removed*)
- [ ] remove the contact details
- [ ] Remove the logo, banner and Description text fields
- [ ] Links to shops open a new window

#### What should we test?

Check that all pages display normally when not embedded.
Check that the embedded shops are displayed correctly.
Check that groups can now be embedded. It would be best to copy the iframe code into another website to test, for example: `<iframe src="/groups/flavour-crusader?embedded_shopfront=true" style="width:100%;min-height:35em"></iframe>`

#### Release notes

Group pages can be embedded on authorised websites.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -57,7 +57,7 @@ class ApplicationController < ActionController::Base
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
     return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present?
-    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test?
+    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
 
     response.headers.delete 'X-Frame-Options'
     response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
--- a/public/embedded-group-preview.html
+++ b/public/embedded-group-preview.html
@@ -1,0 +1,20 @@
+<html>
+  <head><title>Embedded Group</title></head>
+<body>
+
+  <p>
+  This is a preview page for embedded groups.
+  Choose a group to display by copying its permalink id into the URL after the question mark.
+  Example: <code>embedded-group-preview.html?flavour-crusader</code>
+  </p>
+
+  <iframe style="width:100%;min-height:35em"></iframe>
+  <script>
+    var unsafeGroupName = window.location.search.substr(1);
+    var group = unsafeGroupName.replace(/[^a-z0-9-]/g, '');
+    var iframe = document.querySelector("iframe");
+    iframe.src = "/groups/" + group + "?embedded_shopfront=true";
+  </script>
+
+</body>
+</html>
--- a/public/embedded-group-preview.html
+++ b/public/embedded-group-preview.html
@@ -1,20 +1,20 @@
 <html>
   <head><title>Embedded Group</title></head>
-<body>
+  <body>
 
-  <p>
-  This is a preview page for embedded groups.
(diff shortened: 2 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -57,7 +57,7 @@ def restrict_iframes
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
     return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present?
-    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test?
+    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
```
sauloperez: shouldn't we use `force_ssl` in production config? Isn't the whole site using SSL? This way we don't need to manually check any env. That would be set in its config file.

Matt-Yorkley: This is here because of issues with non-SSL sites embedding OFN pages, to make sure they are using SSL.

sauloperez: what kind of issues?

public/embedded-group-preview.html
```diff
@@ -0,0 +1,20 @@
...
+  <iframe style="width:100%;min-height:35em"></iframe>
+  <script>
+    var unsafeGroupName = window.location.search.substr(1);
+    var group = unsafeGroupName.replace(/[^a-z0-9-]/g, '');
+    var iframe = document.querySelector("iframe");
+    iframe.src = "/groups/" + group + "?embedded_shopfront=true";
+  </script>
+
+</body>
+</html>
```
enricostano: Can we fix the indentation?

app/controllers/application_controller.rb
mkllnk: This line checks the referer (if present) to see if the embedding website uses https. Otherwise a sneaky man-in-the-middle wifi router could inject some Javascript code into the embedding http website. That code could then access contents in the iframe and mess around with payment methods or create orders for the attacker.

My change relaxes the condition so that I can test in development mode without https.

public/embedded-group-preview.html
mkllnk: Done.

app/controllers/application_controller.rb
sauloperez: My point is that why do we need to check only here whether the request is HTTPS? Isn't the whole site using SSL? Then we don't need to make this controller aware of SSL but just the app config (as I guess it is now).

Then, the for test or dev env would be useless as that is configured in each env config file for the whole app.

Feel free to dismiss my comment, I might be missing something.

mkllnk: > Isn't the whole site using SSL?

All staging and production OFN instances are using SSL/TLS, yes. But the `referer` here is the URL of the embedding website, a third party we don't control, not OFN.  But it is important that that other website uses SSL/TLS as well.

sstead:
Testing notes:
- the Login (or gear symbol if already logged in) and Cart links could be removed from the header.
Yes this isn't showing in the group, 
- Check that all pages display normally when not embedded.
Yes, they seem fine
- Check that the embedded shops are displayed correctly.
Yes, seems to display and function normally. See here https://openfoodnetwork.org/user-guide/advanced-features/staging-embedded-shop/
- Check that groups can now be embedded.
Yes, see demo here https://openfoodnetwork.org/user-guide/advanced-features/demo-embedded-group/

Issue #1837, opened 2017-09-18
Paypal completing payment but returning to checkout not order conf, order not complete in OFN


kirstenalarsen:
User reported: My store was open Thursday - Sunday yesterday and I took like 8 orders however there is no evidence or confirmation in OFN. Only PayPal receipts with orders in there. 

The orders are in OFN but are not completed, still in State 'checkout'

![image](https://user-images.githubusercontent.com/2911176/30531167-ada96d4e-9c8f-11e7-8d6c-96b9ff708ecf.png)

I have confirmed this problem when I shop through their shop (the mandarin bend) but that our test paypal account on prod is fine

kirstenalarsen:
Have now confirmed that when I put a payment fee on it causes this problem, so very likely connected to #1074 

However, I think this behaviour in response to that problem is new? @oeoeaio @mkllnk @sstead

oeoeaio:
Probably not directly related, but great information to have!

Additional information (from server logs):
````
Redirected by /home/ubuntu/.gem/ruby/2.1.0/bundler/gems/better_spree_paypal_express-8d95f4544c68/app/controllers/spree/paypal_controller.rb:62:in `confirm'
````

This is the method:
````ruby
# app/controllers/spree/paypal_controller.rb
def confirm
  order = current_order || raise(ActiveRecord::RecordNotFound)
  order.payments.create!({
    :source => Spree::PaypalExpressCheckout.create({
      :token => params[:token],
      :payer_id => params[:PayerID]
    }, :without_protection => true),
    :amount => order.total,
    :payment_method => payment_method
  }, :without_protection => true)
  order.next
  if order.complete?
    flash.notice = Spree.t(:order_processed_successfully)
    flash[:commerce_tracking] = "nothing special"
    session[:order_id] = nil
    redirect_to completion_route(order)
  else
    redirect_to checkout_state_path(order.state) # This is line 62
  end
end
````

So now all we need to do is to work out is why adding a payment fee causes the order to not be complete at this point...

oeoeaio:
Ok, so maybe it is more related to #1074 than I thought: `order.next` is failing to transition here because of [this callback](https://github.com/coopdevs/spree/blob/spree-upgrade-step-6/core/app/models/spree/order/checkout.rb#L66) for the state machine on `order`

[...]

Issue #1074, opened 2016-06-29
Paypal % charged twice


Rocksj:
I have added a 4% charge onto our paypal method of payment and its being added onto the customer's invoice twice as below, which leaves a Balance Due on their account.
However, they are only charged this once when they get charged by Paypal.
![screen shot 2016-06-29 at 15 36 35](https://cloud.githubusercontent.com/assets/15102362/16456451/ce1bb3be-3e0f-11e6-875c-784e8ad0b74c.png)

daniellemoorhead:
@mkllnk does this have something to do with the fees on payment types piece of work you did?

mkllnk:
Yes. Somehow it got added twice here. It would be good to know how to reproduce that bug, since we didn't discover that while testing.

Rocksj:
Come shop with us, happens all the time?! Maybe I've set it up wrong?

mkllnk:
@sstead Can you reproduce this?

sstead:
Hey Maikel, I couldn't reproduce this without a properly setup Paypal account. The error wasn't happening for other payment methods with fees. I had a look at the UK server and interestingly in the last month this error seemed to occur on 4 of the 5 orders that were paid  with Paypal. I couldn't glean much else. Would you like me to contact Kirsten/Brenda to get details of the eaterprise paypal account for more thorough testing?

Rocksj:
Sally, when I noticed this was happening I deleted one of the duplicate transactions to make the invoice correct, it was on Order #R757442004. So it may have happened to all of them but doesn't show on the invoice now? I'll leave the rest alone.

Rocksj:
OK, this may be a clue? This was the payment screen which related to the order on my first post where I noticed this

![screen shot 2016-07-01 at 13 57 43](https://cloud.githubusercontent.com/assets/15102362/16522174/e9c841e2-3f93-11e6-9854-625183cd1942.png)

mkllnk:
@sstead Why did the sandbox not work? Or did the error not happen with the sandbox?

The last screenshot looks like someone tried to pay twice and only the second one worked. Maybe that's where the fee is processed twice.

Rocksj:
This is another one with the same transaction twice:
![screen shot 2016-07-02 at 10 12 33](https://cloud.githubusercontent.com/assets/15102362/16539517/9b1222ac-403d-11e6-941e-3e88906af3e8.png)

[...]

PR #1842, opened 2017-09-20
Remove orphaned PayPal payments 

oeoeaio:
#### What? Why?

Fix for both #1074 and #1837. Lots of detail on both of those issues.

Basically, there is no nice way to fix this (as far as I can see). So I went with the simplest and least ugly solution that I could come up with. If you have any better ideas, let me know.

I also tried preventing the orphaned payment from being created in the first place, and shifting responsibility for the application of the transaction fee to the CheckoutController, but I didn't like the outcome. You have [have a look here](https://github.com/oeoeaio/openfoodnetwork/commit/5c72ed7c7baf794e80994b30bb3957cd4f7f6dfd) if you are interested in that approach.

#### What should we test?

That Paypal payments with transaction fees can go through without redirecting to Checkout. Would be good do confirm that cancelling the order when you hit Paypal does not lead to duplicate fees or payments being added to the order.

#### Release notes

Bug causing duplicate transaction fees for PayPal payments has been fixed

This is a bug fix for an issue that was made worse by the Spree Upgrade.

### DIFF
--- a/spec/models/spree/payment_spec.rb
+++ b/spec/models/spree/payment_spec.rb
@@ -124,5 +124,93 @@ module Spree
         payment.refund!
       end
     end
+
+    describe "applying transaction fees" do
+      let!(:order) { create(:order) }
+      let!(:line_item) { create(:line_item, order: order, quantity: 3, price: 5.00) }
+
+      # Mimicing call from PayPalController#confirm in spree_paypal_express
+      def create_new_paypal_express_payment(order, payment_method)
+        order.payments.create!({
+          :source => Spree::PaypalExpressCheckout.create({
+            :token => "123",
+            :payer_id => "456"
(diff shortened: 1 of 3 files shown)


spec/models/spree/payment_spec.rb
```diff
@@ -124,5 +124,86 @@ module Spree
...
+
+            before do
+              existing_payment.update_attributes(payment_method_id: another_payment_method.id)
+            end
+
+            it "does not remove the existing payment" do
+              create_new_paypal_express_payment(order, payment_method)
+
+              order.reload
+              expect(order.payments.count).to eq 2
```
sauloperez: IMO it's better to check the actual payment objects so we avoid false positives.

Issue #1826, opened 2017-09-13
ActiveRecord::RecordInvalid in spree/admin/orders#update


mkllnk:
## Error in OpenFoodNetwork

**ActiveRecord::RecordInvalid** in **spree/admin/orders#update**
Validation failed: State is invalid

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/59b9ab57bf20b40018c56d0d?event_id=59b9abdd93bd8700171ce04b&i=gh&m=ci)

## Stacktrace

    app/models/spree/adjustment_decorator.rb:26 - set_absolute_included_tax!
    app/models/spree/adjustment_decorator.rb:22 - set_included_tax!
    app/models/spree/shipment_decorator.rb:7 - ensure_correct_adjustment_with_included_tax
    app/models/spree/order_decorator.rb:34 - block (3 levels) in <top (required)>

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/59b9ab57bf20b40018c56d0d?event_id=59b9abdd93bd8700171ce04b&i=gh&m=ci)

*Created by Maikel via Bugsnag*

mkllnk:
@oeoeaio @enricostano This is new on Aus production since yesterday, happened four times so far, same user. Could this be connected to to the Spree upgrade?

oeoeaio:
@mkllnk Yep, I think so. I'm investigating now.

PR #1845, opened 2017-09-21
Add spree adjustment state migrations

oeoeaio:
#### What? Why?

This is a better fix for #1826. The original fix (#1827) was a quick hack to prevent users from seeing errors. This fix uses Spree migrations to solve to problem, and reverts the original solution.

#### What should we test?

Make sure that older adjustments on orders can be updated without crashing.

#### Release notes

No release notes, as the bug that caused #1826 should never make it into a release without this fix.

The issue was caused by the spree-upgrade, so this is one of the final fixes needed to allow us to release the upgrade.

### DIFF
--- a/db/migrate/20170921065259_update_adjustment_states.spree.rb
+++ b/db/migrate/20170921065259_update_adjustment_states.spree.rb
@@ -1,0 +1,17 @@
+# This migration comes from spree (originally 20130417120035)
+class UpdateAdjustmentStates < ActiveRecord::Migration
+  def up
+    Spree::Order.complete.find_each do |order|
+      order.adjustments.update_all(:state => 'closed')
+    end
+
+    Spree::Shipment.shipped.includes(:adjustment).find_each do |shipment|
+      shipment.adjustment.update_column(:state, 'finalized') if shipment.adjustment
+    end
+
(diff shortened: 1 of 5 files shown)


db/migrate/20170921065259_update_adjustment_states.spree.rb
```diff
@@ -0,0 +1,17 @@
...
+    Spree::Shipment.shipped.includes(:adjustment).find_each do |shipment|
+      shipment.adjustment.update_column(:state, 'finalized') if shipment.adjustment
+    end
+
+    Spree::Adjustment.where(:state => nil).update_all(:state => 'open')
+  end
+
+  def down
+  end
+end
```
sauloperez: damn! we shouldn't have missed this :see_no_evil:

oeoeaio: It's amazing that you only missed one! There is so much code to look through - impossible to catch everything! @enricostano found this - I looked for it after the bug appeared and I couldn't find it even though I knew what I was looking for!!! ❤️

sauloperez: Note to myself. DDL migrations very very likely come with a data migration :sweat_smile:

Issue #1802, opened 2017-09-05
Missing translations in main "products" page


myriamboure:
comment : on the page title, it seems the key is missing in transifex, so it seems like an automatic translation from spree but which is not at all meaningful.
comment 2 : on the column header "nom" ("name") it looks like it is translated but I wanted to change the translation in transifex by "product/variant" to clarify following users feedbacks, and I realize there is also no key for that header, so I guess it was also an automatic translation from Spree...
![illustration](https://user-images.githubusercontent.com/8135376/30079102-58ebab42-927f-11e7-9cf5-3c1833ff94ef.jpeg)

ltrls:
The title should appear in Transifex, if you can look for it by key it is: `spree.admin.products.bulk_edit.header.title`

myriamboure:
Thanks, and yes the title is translating ok, I guess it was just a transifex branch not update on our server ;-)

Issue #1800, opened 2017-09-05
Missing translations in "orders" page


myriamboure:
comment : "nombre" seems to be like a bad automatic translation from spree, it looks like it's translated but I checked in transifex and it's not.
![illustration](https://user-images.githubusercontent.com/8135376/30077409-10668f48-927c-11e7-8a84-0f9ef479461e.jpeg)

Issue #1798, opened 2017-09-05
Missing translations in the "nofify producers" email


myriamboure:
![illustration2](https://user-images.githubusercontent.com/8135376/30073687-ffda8c9a-926e-11e7-8928-d2cf4dd108f7.jpeg)

ltrls:
The email subject is translated in my development environment, I can't see why it wouldn't work in production oO

myriamboure:
I have re-tested and it's still the same for me... where can the issue come from ? Any idea @sauloperez or @enricostano do you have the same issue in Spain ?

myriamboure:
Did you manage to understand the issue @ltrls ?

myriamboure:
Seems like the key is in transifex since last May but it's not working...
![screenshot from 2017-10-04 17-23-28](https://user-images.githubusercontent.com/8135376/31184129-d9a1b7f8-a928-11e7-87b3-fe545aee7cac.png)

myriamboure:
Title fixed with the upgrade on French production.

Issue #1634, opened 2017-06-21
Missing translation in order confirmation email


myriamboure:
![screenshot from 2017-06-21 10-07-57](https://user-images.githubusercontent.com/8135376/27373742-c0e4b6f0-5669-11e7-9bab-add9e84f1ae3.png)

PR #1847, opened 2017-09-21
Fix some translations

ltrls:
This fixes some missing translations reported in #1634, #1798, #1800 and #1802.

### Testing

In addition to and maybe including the issues mentioned above, we should check:

- bulk products edit
- the main menu
- order emails

### DIFF
--- a/app/views/spree/admin/products/bulk_edit/_products_head.html.haml
+++ b/app/views/spree/admin/products/bulk_edit/_products_head.html.haml
@@ -23,16 +23,16 @@
         = t(:expand_all)
     %th.producer{ 'ng-show' => 'columns.producer.visible' }=t('admin.producer')
     %th.sku{ 'ng-show' => 'columns.sku.visible' }=t('admin.sku')
-    %th.name{ 'ng-show' => 'columns.name.visible' }=t('admin.products.bulk_edit.name')
-    %th.unit{ 'ng-show' => 'columns.unit.visible' }=t('admin.products.bulk_edit.unit')
-    %th.display_as{ 'ng-show' => 'columns.unit.visible' }=t('admin.products.bulk_edit.display_as')
+    %th.name{ 'ng-show' => 'columns.name.visible' }=t('.name')
+    %th.unit{ 'ng-show' => 'columns.unit.visible' }=t('.unit')
+    %th.display_as{ 'ng-show' => 'columns.unit.visible' }=t('.display_as')
     %th.price{ 'ng-show' => 'columns.price.visible' }=t('admin.price')
     %th.on_hand{ 'ng-show' => 'columns.on_hand.visible' }=t('admin.on_hand')
     %th.on_demand{ 'ng-show' => 'columns.on_demand.visible' }=t('admin.on_demand?')
-    %th.category{ 'ng-show' => 'columns.category.visible' }=t('admin.products.bulk_edit.category')
-    %th.tax_category{ 'ng-show' => 'columns.tax_category.visible' }=t('admin.products.bulk_edit.tax_category')
-    %th.inherits_properties{ 'ng-show' => 'columns.inherits_properties.visible' }=t('admin.products.bulk_edit.inherits_proerties?')
-    %th.available_on{ 'ng-show' => 'columns.available_on.visible' }=t('admin.products.bulk_edit.av_on')
+    %th.category{ 'ng-show' => 'columns.category.visible' }=t('.category')
+    %th.tax_category{ 'ng-show' => 'columns.tax_category.visible' }=t('.tax_category')
+    %th.inherits_properties{ 'ng-show' => 'columns.inherits_properties.visible' }=t('.inherits_properties?')
+    %th.available_on{ 'ng-show' => 'columns.available_on.visible' }=t('.av_on')
     %th.actions
     %th.actions
     %th.actions
--- a/app/views/spree/admin/products/bulk_edit/_products_head.html.haml
+++ b/app/views/spree/admin/products/bulk_edit/_products_head.html.haml
@@ -23,7 +23,7 @@
         = t(:expand_all)
     %th.producer{ 'ng-show' => 'columns.producer.visible' }=t('admin.producer')
     %th.sku{ 'ng-show' => 'columns.sku.visible' }=t('admin.sku')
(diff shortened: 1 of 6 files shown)


app/views/spree/admin/products/bulk_edit/_products_head.html.haml
```diff
@@ -23,7 +23,7 @@
         = t(:expand_all)
     %th.producer{ 'ng-show' => 'columns.producer.visible' }=t('admin.producer')
     %th.sku{ 'ng-show' => 'columns.sku.visible' }=t('admin.sku')
-    %th.name{ 'ng-show' => 'columns.name.visible' }=t('admin.name')
+    %th.name{ 'ng-show' => 'columns.name.visible' }=t('admin.products.bulk_edit.name')
```
sauloperez: did you try using `.name` as per Rails lazy lookup? See 4.1.4 in http://guides.rubyonrails.org/i18n.html. If possible, use the namespace in the other copies below :pray:

config/locales/en.yml
```diff
@@ -54,18 +54,24 @@
...
+  # Overridden here due to a bug in spree i18n (Issue #870, and issue #1800)
   activerecord:
-    # Overridden here due to a bug in spree i18n (Issue #870)
     attributes:
       spree/order:
         payment_state: Payment State
         shipment_state: Shipment State
+        completed_at: Completed At
+        number: Number
+        email: Customer E-Mail
```
sauloperez: where are these 3 translations used?

ltrls: In Admin/Orders list (see #1800). As you can see [here](https://github.com/spree-contrib/spree_i18n/blob/1-3-stable/config/locales/fr.yml#L53) they're not translated in spree_i18n (at least for French), it seems that adding them in `en.yml` would make them appear in Transifex and make it possible to translate them here.

sauloperez: Given that we use our own fork, I'd rather fix it there https://github.com/openfoodfoundation/spree_i18n/blob/1-3-stable/config/locales/fr.yml#L53

I see it correctly translated in the `master` branch. Can you check whether it got fix in any upcoming commit? I'm afraid we might need to move a hundred commits forward as we did in the step 6 for the spree gem.

ltrls: It got fixed in a commit which is not on the 1-3-stable branch but there are commits  on the 1-3-stable branch which are not in other branches... do you think I should just PR on the 1-3-stable branch?

ltrls:
The failing test doesn't fail in my dev environment, could it be something with Travis?

sauloperez:
Some tests (mostly integration ones) are not reliable at all and fail sometimes. Until we don't fix them the only we thing we can do is to restart the failing job.

config/locales/en.yml
sauloperez: > do you think I should just PR on the 1-3-stable branch?

Yes, please

ltrls: After talking with @myriamboure it seems like 2 of these translations (`number` and `email`) won't be translated the way we'd like them to be. That would be nice if I add them in OFN repo, but I can still PR for `completed_at` and `show_only_unfulfilled_orders` which will be rightly translated so don't need to be added in OFN repo.

app/views/spree/admin/products/bulk_edit/_products_head.html.haml
ltrls: hop! there it is

ltrls:
Related PR in spree_i18n is [here](https://github.com/openfoodfoundation/spree_i18n/pull/1). I guess we should wait for it to be merged before this one so I can point Gemfile.lock to the last revision and bundle doesn't fail

myriamboure:
@daniellemoorhead who can merge the spree_i18n as mentionned by @ltrls ? We wait for that before tests can pass and we can merge this one ;-)

enricostano:
@ltrls https://github.com/openfoodfoundation/spree_i18n/pull/1 got merged, should we update our Gemfile to include the new changes?

ltrls:
I've already done that, that's what makes the test fail! I'm gonna resolve the conflict so when they run again it should be all green :)

ltrls:
I must have done it wrong, isn't the `revision` that I need to update, and get it from the commit hash?

daniellemoorhead:
I would assume that @mkllnk would merge any i18n PRs @myriamboure, though perhaps @enricostano has been merging these as well?

sauloperez:
AFAIK there's no responsible for I18n PRs only. @enricostano has been merging them as well.

Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'nokogiri', '>= 1.6.7.1'
 
 gem 'pg'
 gem 'spree', github: 'openfoodfoundation/spree', branch: 'step-6a', ref: '5a76d45'
-gem 'spree_i18n', github: 'spree/spree_i18n', branch: '1-3-stable'
+gem 'spree_i18n', github: 'openfoodfoundation/spree_i18n', branch: '1-3-stable'
```
oeoeaio: I didn't realise that we weren't already using our own fork of `spree_i18n`. Is it not possible to just add the translations submitted in https://github.com/openfoodfoundation/spree_i18n/pull/1 to our own `en.yml`? I would prefer that option if possible, rather than forking all the things.

myriamboure:
@sauloperez regarding our mumble discussion, there are also all the "fee calculator options" (for entreprise fees, shipping fees and payment method fees) that are badly translated, I think this comes from Spree ? I don't know really what is Spree and what is not, let us know if you agree with @oeoeaio or not. @ltrls you are also welcome to share if you agree with what Rob suggested.

Gemfile
ltrls: OK I put the changes back to en.yml :)

myriamboure: @sauloperez for info, please let us no if any objection, else we push again for Aus Review and hope it can be merged before next release so that we see better what translations are still missing ;-)

sauloperez: Fine for me. Let's move forward :+1:

sstead:
Testing notes
Order confirmation emails - Yes, look fine
Notify producers - Yes looks fine
Orders page - Yes, looks fine
BOM - yes, looks fine
Products page - Yes looks fine

PR #1857, opened 2017-09-25
Add pointer cursor to EXPAND ALL link in products bulk edit

### LINKED ISSUE #1117: When the arrow of the mouse pass over "expand all" on the product list, the arrow transforms in a vertical writting  tab instead of a hand
Connected to #612, a very little thing which has a bad impact on user experience for the hub manager, that I have noticed in supporting new hub user.
On the bulk edit products page, when I approach the arrow of my mouse to the "expand all" red text, the arrow transform in a writting vertical bar, not in the usual small hand that indicates that the link can be clicked.
I saw with users sharing their screen with me that they were passing the expand all text without noticing they could click on it. Ok it could seem like obvious but not all users are very comfortable with platforms and web interfaces...

ltrls:
This fixes #1117.

### DIFF
--- a/app/assets/stylesheets/admin/products.css.scss
+++ b/app/assets/stylesheets/admin/products.css.scss
@@ -69,4 +69,10 @@ table#listing_products.bulk {
       margin-bottom: 0.5em;
     }
   }
+
+  td.left-actions {
+    a.view-variants, a.add-variant {
+      cursor: pointer;
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


daniellemoorhead:
@mkllnk do you think we could get this onto staging for @sstead or @simoneluijckx to test in the next couple of days?

sstead:
Testing notes:
Looks good!

Issue #1773, opened 2017-08-19
Desindex unvisible shops in google search


myriamboure:
When I search by the name an entreprise that has an unvisible shop in Google, it appears as first link... if they are unvisible, this is because they don't want to be search... can we desindex the unvisible shops in Google/other search tools ?

myriamboure:
@ltrls is it something that is specific to each instance or is it needed to discuss and reach a consent by the community ? It sounds to me like more instance based...
I discussed with the invisible hubs and they all don't want to appear in google searches, so we need to move forward.
One question : if they are desindexed through OFN but embed their shopfront in their own website, I guess as the url will be different that won't have any impact, they can stay indexed from their own website ?
Ping @daniellemoorhead how have you done in Australia ? And @lin-d-hop in UK ?

sauloperez:
This is the way to do so https://support.google.com/webmasters/answer/93710?hl=en

daniellemoorhead:
I think @oeoeaio and @mkllnk can probably tell you better than I what the AU team have done @myriamboure...or maybe @kirstenalarsen?

mkllnk:
That should be a very easy code change to include the mentioned meta tag.

myriamboure:
@pacodelaluna do you have time to work on that in the coming days ? We just tested with @RachL and the french staging server is also still visible on google when you type "staging open food france"... I can create another issue but if you work on that one maybe you can desindex staging at the same time ? Please let me know if you can't work on it, I will try to see with @ltrls find someone else to do it as we really need to move forward... I know you are super busy at the moment, so just let me know ;-)

PR #1885, opened 2017-09-27
Add noindex meta tag for invisible shops and staging environment

ltrls:
PR for #1773.
It adds the noindex tag in the layout for the whole app if the environment isn't production and for shops that have chosen to be "invisible", to prevent from being indexed by search engine bots.

### DIFF
--- a/app/views/layouts/darkswarm.html.haml
+++ b/app/views/layouts/darkswarm.html.haml
@@ -5,6 +5,8 @@
     %meta{property: "og:title", content: content_for?(:title) ? yield(:title) : t(:title)}
     %meta{property: "og:description", content: content_for?(:description) ? yield(:description) : t(:site_meta_description)}
     %meta{property: "og:image", content: content_for?(:image) ? yield(:image) : ContentConfig.logo.url}
+    - if !Rails.env.production? || @noindex_meta_tag
+      %meta{name: "robots", content: "noindex"}
     %title= content_for?(:title) ? "#{yield(:title)} - #{t(:title)}".html_safe : "#{t(:welcome_to)} #{t(:title)}"
     - if Rails.env.production?
       = favicon_link_tag
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -7,7 +7,7 @@ class EnterprisesController < BaseController
 
   # These prepended filters are in the reverse order of execution
   prepend_before_filter :set_order_cycles, :require_distributor_chosen, :reset_order, only: :shop
-  before_filter :check_stock_levels, only: :shop
+  before_filter :check_stock_levels, :set_noindex_meta_tag, only: :shop
 
   before_filter :clean_permalink, only: :check_permalink
   before_filter :enable_embedded_shopfront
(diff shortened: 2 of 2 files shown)


app/views/layouts/darkswarm.html.haml
```diff
@@ -5,6 +5,8 @@
     %meta{property: "og:title", content: content_for?(:title) ? yield(:title) : t(:title)}
     %meta{property: "og:description", content: content_for?(:description) ? yield(:description) : t(:site_meta_description)}
     %meta{property: "og:image", content: content_for?(:image) ? yield(:image) : ContentConfig.logo.url}
+    - if !Rails.env.production? || @noindex_meta_tag
```
sauloperez: I'd rather have a method that assigns a hash like `{ name: 'robots', content: 'noindex' }` depending on this boolean logic above and assigns it in an ivar. Then the view is as simple (and decoupled) as:

```ruby
%meta{name: @meta_tag.name, content: @meta_tag.content} if @meta_tag
```

What do you think?

ltrls: I'm not sure that it wouldn't be premature refactoring. Sure it's more decoupled this way, but we also lose some explicitness.

sauloperez: Could be. The think I dislike the most is start coupling views with environments though.

oeoeaio: Ok, so for now can we just move the environment check into a guard clause in the `set_noindex_meta_tag` method? Then we can just have:



````ruby
# controller
def set_noindex_meta_tag
  return if Rails.env.production? && current_distributor.visible?
  @noindex_meta_tag = true
end

...

#view
%meta{name: "robots", content: "noindex"} if @noindex_meta_tag 
````

ltrls: Yes but it changes the condition. From `(not production env OR (shop action
AND invisible shop))` it becomes `(shop action AND (not production env AND
invisible shop))`. What I intended to do was un-index the whole site if it's
not in production.

oeoeaio: Oh, right! We could add the setter to the base controller and put the env check there, then call super from the EnterprisesController? Getting messy... Maybe it’s ok the way it is?

ltrls: I really think it's the best way because we clearly understand when the tag is "displayed"

myriamboure:
So @oeoeaio is it ok if @ltrls leave it ike that ? @sauloperez no further objection ? If needs be don't hesitate to discuss that together quicky on Slack maybe to avoid the back and forth time ;-)

sigmundpetersen:
@myriamboure @daniellemoorhead should this be added to Oct milestone?

daniellemoorhead:
>should this be added to Oct milestone?

Potentially @sigmundpetersen. I've been letting the devs who merge the PR put the milestone on when they do it. So let's follow that process for now. Many of the PRs that went through on Oct dot point are likely to go in with the stripe release 1.10 (hopefully this week), so we'll be moving things around as a result. Best wait for now, we'll put the correct milestone on once we know which one it will go in :-)

daniellemoorhead:
Over to you @oeoeaio :-)

sauloperez:
No need to further delay this. It's fine as it is. We can always change it if we need to.

sigmundpetersen:
@oeoeaio #nomilestone ?

daniellemoorhead:
@sigmundpetersen #niceuseofhashtags #goodforagiggle

PR #1886, opened 2017-09-27
Fix js currency

ltrls:
#### What? Why?

This PR fixes both problems raised in #1623. The issue was that serializers sent amounts of money as localized strings (`to_money.to_s`) so if the locale used a comma as decimal separator, parsing them in JS to perform operations (in the `/account` page) would fail leading either to a wrong amount (parsing stopped at the comma stripping decimals) or a NaN (not a number) error.
Here I made serializers send the amounts as unlocalized strings, and improved the `localizeCurrency` filter to use i18n-js `toCurrency` method, which as a nice "side effect" also localizes the decimal separator and spacer character.
I18n-js's `toCurrency` method uses translation values in `number.currency.format`, which are set in the rails-i18n gem.

#### What should we test?

The localizeCurrency filter is used in many places but the most notable page is the page at `/account` (see images in related issue) which now presents no error when using a locale with comma as decimal separator. In other places, amounts should also be nicely localized.

### DIFF
--- a/app/serializers/api/order_serializer.rb
+++ b/app/serializers/api/order_serializer.rb
@@ -28,10 +28,6 @@ module Api
       I18n.l(object.order_cycle.andand.orders_close_at, format: "%b %d, %Y %H:%M")
     end
 
-    def total
-      object.total.to_money.to_s
-    end
-
     def shipment_state
       object.shipment_state ? object.shipment_state : nil
     end
--- a/app/assets/javascripts/darkswarm/filters/localize_currency.js.coffee
+++ b/app/assets/javascripts/darkswarm/filters/localize_currency.js.coffee
@@ -3,12 +3,12 @@ Darkswarm.filter "localizeCurrency", (currencyConfig)->
   (amount) ->
     # Set country code (eg. "US").
     currency_code = if currencyConfig.display_currency then " " + currencyConfig.currency else ""
-    # Set decimal points,  2 or 0 if hide_cents.
+    # Set decimal points, 2 or 0 if hide_cents.
     decimals = if currencyConfig.hide_cents == "true" then 0 else 2
-    # Set wether the currency symbol appears first
-    sign_first =  currencyConfig.symbol_position == 'before'
+    # Set format if the currency symbol should come after the number, otherwise (default) use the locale setting.
+    format = if currencyConfig.symbol_position == "after" then "%n %u" else undefined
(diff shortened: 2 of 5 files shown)


app/serializers/api/order_serializer.rb
```diff
@@ -23,10 +23,6 @@ def changes_allowed_until
       I18n.l(object.order_cycle.andand.orders_close_at, format: "%b %d, %Y %H:%M")
     end
 
-    def total
-      object.total.to_money.to_s
-    end
-
```
sauloperez: we don't use it anymore?

ltrls: Yes we do but it would still appear as it's declared in the [`attributes` method](https://github.com/ltrls/openfoodnetwork/blob/c43e40a4c460fb5f7ae352f44a5d5fbc3aec8238/app/serializers/api/order_serializer.rb#L3). What we want is that it's like `def total; object.total; end` and there's no need to have this as it's like not writing the method (customizing the field) at all.

sstead:
**Testing Notes**
- The /Account/Transactions shop 'balance due' numbers are correct, and the tally is calculating correctly.

Some currency display errors we saw...
image

myriamboure:
Are the currency display issues connected to that PR ? Any idea @ltrls ? It seems it is being merged so I suspect it is not related, or should we open another issue ? @sstead this is not clear to me :-o

daniellemoorhead:
@oeoeaio see @myriamboure's comment above. Were the currency display errors fixed by you before you merged it, or where they put into another issue to be done separately?

oeoeaio:
Sorry, I thought I responded to this on Friday. Currency display errors were not fixed, but I've just realised that the reason was probably our user error. The currency components affected by this PR were displaying correctly, but the components being rendered by Rails were not displaying correctly. I now realise that this was probably because we didn't change the configuration of `currency_symbol_position` in the backend. I didn't really think about it (since we don't use it) and just assumed that the locale would change this automatically.

So, in short: @myriamboure, can you confirm that this looks ok for you in France the next time you deploy? There problem is probably just a fairly non-intuitive configuration process, rather than an issue with the logic.

myriamboure:
@oeoeaio I can confirm we just deployed the master version and it's all good on the currency display, the € symbol is after the price on all locations, shopfront, basket, payment, order management, etc. :+1:

oeoeaio:
Awesome, thanks!

PR #1890, opened 2017-09-29
Track node version in .node-version

sauloperez:
#### What? Why?

This is used by https://github.com/nodenv/nodenv. As well as we do with Ruby, we need to track the node version required. Remember we need node to run Angular tests with karma.

thoughts?


daniellemoorhead:
Hey @enricostano #whereisthemilestone #pleaseaddit

PR #1892, opened 2017-10-06
Stripe connect

oeoeaio:
#### What? Why?

A couple of final fixes to Stripe work
- Add some logic to CheckoutController to strip incomplete shipments and payments from orders after a failed attempt to check out. This prevents incorrect shipping and payment fees from displaying in Checkout summary. Also allows correct fees to be charged if items are added or removed from cart after an initial attempt to process an order fails.

#### What should we test?

- Correct fees charged displayed and charged after > 1 failed attempts to process an order (eg. Stripe card is rejected, PayPal checkout is cancelled by the user, etc.)
- Especially test that when items are added to cart after an initial failed attempt to process an order, correct fees are still displayed and ultimately charged.

#### Release notes

N/A, part of Stripe work which is yet to be released.

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -15,6 +15,10 @@ class CheckoutController < Spree::CheckoutController
   include EnterprisesHelper
 
   def edit
+    # This is only required because of spree_paypal_express. If we implement
+    # a version of paypal that uses this controller, and more specifically
+    # the #update_failed method, then we can remove this call
+    restart_checkout
   end
 
   def update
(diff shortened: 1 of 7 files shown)


sstead:
Testing notes: 
If the payment didn't go through because of 1) total was <50c 2) wrong CSV 3) customer backed out of Paypal, the 'total' line at checkout could display an error AFTER the order was adjusted and shipping/payment fees recalculated. 
I believe the error lies in the shipping method fee reculculating, because the error didn't arise when there was only a fee on the payment method.

Despite the 'total' display error, the amount charged to customers was correct. The error seems to just be on this one line at checkout.

image

sstead:
This morning's bug is fixed. 

I've spotted 2 things, but I think they're not related to this PR...

I've made an issue for this one...https://github.com/openfoodfoundation/openfoodnetwork/issues/1901
image

and

image

Issue #1738, opened 2017-07-28
When i use 'print invoice' for a long order it has display errors on second page


sstead:
If an order is long enough to cover 2 pages the 'print invoice' invoice has display issues. This occurs for both invoice types.

![image](https://user-images.githubusercontent.com/10843056/28700435-1fd0629a-7393-11e7-9d59-2c7399fae232.png)

![image](https://user-images.githubusercontent.com/10843056/28700448-2f22ae7e-7393-11e7-9657-16211fa8a429.png)

*Steps to reproduce*

* Create an order with at least 35 items.
* Go to admin / orders / edit order: choose "print invoice" from the actions dropdown.

sineadfenton:
![](https://github.trello.services/images/mini-trello-icon.png) [Invoices & Second Page](https://trello.com/c/jYZOFZNa/83-invoices-second-page)

sstead:
@daniellemoorhead we've had a user in Australia request that this is fixed, how can we speed it along? Should we shift it to Aus Current? Or @sineadfenton any chance you guys can tackle it soonish?

daniellemoorhead:
@kirstenalarsen are we able to get @mkllnk to fix this bug for AU?

kirstenalarsen:
sure . .

daniellemoorhead:
@mkllnk can you also add this to your list :)

sstead:
@mkllnk josh from Synchronicity farm has asked about this a couple of times. Any way to speed it up?

PR #1893, opened 2017-10-06
Add CSS workaround to display repeated table head

mkllnk:
Fixes https://github.com/openfoodfoundation/openfoodnetwork/issues/1738

It adds some CSS as a workaround. We are already using the newest versions of wicked_pdf and wkhtmltopdf.

#### What should we test?

That invoice PDFs are displaying correctly with both invoice styles (see super admin setting).

#### Release notes

Long invoices are displayed correctly.


sstead:
Testing notes
Looks good.
image

PR #1895, opened 2017-10-07
Relax rubocop Style/SymbolArray rule

Matt-Yorkley:
#### What? Why?

Relaxing the symbol arrays style cop to allow symbol arrays like this: `[:one, :two]` and not strictly enforce this style: `%i[one two]`. 

Current rule completely prohibits lines like this:
`before_filter :strip_new_properties, only: [:create, :update]`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -143,6 +143,10 @@ Style/WordArray:
   Enabled: false
   StyleGuide: http://relaxed.ruby.style/#stylewordarray
 
+Style/SymbolArray:
+  Enabled: false
+  StyleGuide: https://rubocop.readthedocs.io/en/latest/cops_style/#stylesymbolarray
+
 Lint/AmbiguousRegexpLiteral:
   Enabled: false
   StyleGuide: http://relaxed.ruby.style/#lintambiguousregexpliteral


oeoeaio:
I don't have a strong preference either way, but we've had a discussion about this before and I think @sauloperez and @enricostano have a preference for keeping this cop?

We could use the `MinSize` setting to enforce the ` %i()` syntax for larger arrays?

sauloperez:
> We could use the MinSize setting to enforce the %i() syntax for larger arrays?

That's a good tradeoff in my opinion. I believe that for larger arrays `%i` it does improve readability.

mkllnk:
The problem with MinSize are patches like this:
```diff
-   before_filter :strip_new_properties, only: [:create, :update]
+   before_filter :strip_new_properties, only: %i[create update destroy]
```

You have to change every entry in the array when you cross the limit. It's also not that consistent while consistent style use is more important than the choice of style.

I'm okay with either keeping the old style, because we are used to it and Spree uses it, or, enforcing the new style and being consistent.

That said, I think that we don't need a consensus on these decisions, only a majority. I'm happy to be overruled here. ;-)

sauloperez:
> That said, I think that we don't need a consensus on these decisions, only a majority. I'm happy to be overruled here. ;-)

Agree. As my beloved Sandi Metz recently said in https://www.sandimetz.com/blog/2017/6/1/why-we-argue-style:

> The truth is, many stylistic choices are arbitrary, and purely a matter of personal preference. Choosing a style guide means building agreements in areas where we have strong differences of opinion about issues of little significance. It's not style that matters, but sameness of style.

So it seems like per @mkllnk arguments it is easier to stick to `[:foo, :bar]` style. Agree on removing the rule then.

PR #1896, opened 2017-10-09
Disable allow_backorders by default in test environment

Matt-Yorkley:
#### What? Why?

The `allow_backorders` admin preference is currently disabled on instances but is not disabled by default in the test environment and has created a number of issues with enabling/disabling it in various specs.

#### What should we test?

Nothing, it just applies to specs.

#### Discourse thread

https://community.openfoodnetwork.org/t/backorders/1080

### DIFF
--- a/config/initializers/spree.rb
+++ b/config/initializers/spree.rb
@@ -18,6 +18,9 @@ Spree.config do |config|
   config.shipping_instructions = true
   config.address_requires_state = true
 
+  # Set allow_backorders to false by default in test environment
+  config.allow_backorders = false if Rails.env == 'test'
+
   # -- spree_paypal_express
   # Auto-capture payments. Without this option, payments must be manually captured in the paypal interface.
   config.auto_capture = true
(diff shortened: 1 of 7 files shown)


daniellemoorhead:
@Matt-Yorkley best bet to get the AU team's attention is to add the aus review label to your PRs :)

config/initializers/spree.rb
```diff
@@ -18,6 +18,9 @@
   config.shipping_instructions = true
   config.address_requires_state = true
 
+  # Set allow_backorders to false by default in test environment
+  config.allow_backorders = false if Rails.env == 'test'
```
oeoeaio: As you suggest @Matt-Yorkley, depending on the result of the discourse discussion, we could potentially just set this to `false` for all environments. My understanding is that the OFN is not really set up to handle backorders properly anyway.

oeoeaio:
Hey @Matt-Yorkley, I've tweaked this commit a bit and included it in #1904. Are you ok if I close this?

Matt-Yorkley:
Incorporated into #1904.

PR #1898, opened 2017-10-11
Fix PayPal spec failure

oeoeaio:
#### What? Why?

The user unit spec was improperly using the `quick_login_as_admin` helper method to create an admin user. I am assuming that this was setting the `user_id` in the session, which was not being torn down, because that shouldn't be necessary for unit specs. This was having an affect of subsequent request specs which thought that a user which didn't exist was logged in.

#### What should we test?

Nothing to test

#### Release notes

None


daniellemoorhead:
@oeoeaio #addthemilestone

PR #1899, opened 2017-10-11
Stripe webhook

oeoeaio:
#### What? Why?

I've worked out that having a single endpoint to handle webhooks is going to make it much easier to handle authentication. Stripe provides a secret for each webhook endpoint provided, so having just the one endpoint will mean we only need to manage one secret.

I've added a new service object to handle the actual logic of what to do with the request, and made the controller itself very simple and generic. There may be a need to delegate to more specialised service objects in the future if the existing one gets too bloated, but I figure we can do that as the need arises.

I've also implemented [Stripe's recommended process](https://stripe.com/docs/webhooks#verify-official-libraries) for verifying webhook authenticity.

Next step is to add logic for handling refunds issues through Stripe's own UI.

#### What should we test?

That Stripe-initiated deauthorizations still work. 

#### Release notes

Stripe has not yet been released, so no release notes required.


config/routes.rb
```diff
@@ -53,6 +53,10 @@
     end
   end
 
+  resource :stripe, only: [] do
+    post :webhook
```
sauloperez: I'd rather follow the Rails convention and stick to REST like:

```
 # POST /stripe/webhooks
  namespace :stripe do
    resource :webhooks, only: [:create]
  end
```

oeoeaio:
Sorry @sauloperez I think I accidentally deleted your comment! Anyway, I've done what you suggested.

app/controllers/stripe/webhooks_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    # and will stop sending events if too many of either of these are returned.
+    def status_mappings
+      {
+        success: 200, # The event was handled successfully
+        unknown: 202, # The event was of an unknown type
+        ignored: 204  # No action was taken in response to the event
+      }
+    end
+  end
+end
```
sauloperez: very neat :ok_hand:

daniellemoorhead:
@oeoeaio #addthemilestoneonmergeplease

sauloperez:
LOL

PR #1903, opened 2017-10-13
Stub out Spree::Config to prevent caching

oeoeaio:
Solving the same problem as #1742, but using a different approach.

#### Release notes

None, only specs affected

### DIFF
--- a/config/initializers/spree.rb
+++ b/config/initializers/spree.rb
@@ -9,6 +9,7 @@
 require 'spree/product_filters'
 require 'spree/core/calculated_adjustments_decorator'
 
+require "#{Rails.root}/app/models/spree/app_configuration_decorator"
 require "#{Rails.root}/app/models/spree/payment_method_decorator"
 require "#{Rails.root}/app/models/spree/gateway_decorator"
 
@@ -17,6 +18,13 @@ Spree::Api::Config[:requires_authentication] = true
 Spree.config do |config|
   config.shipping_instructions = true
   config.address_requires_state = true
+
+  # Set defaults for testing
+  if Rails.env.test?
+    config.stripe_connect_enabled = false
+    config.enable_embedded_shopfronts = false
+    config.products_require_tax_category = false
+  end
 
   # -- spree_paypal_express
   # Auto-capture payments. Without this option, payments must be manually captured in the paypal interface.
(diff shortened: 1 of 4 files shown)


config/initializers/spree.rb
```diff
@@ -18,6 +19,13 @@
   config.shipping_instructions = true
   config.address_requires_state = true
 
+  # Set defaults for testing
+  if Rails.env.test?
+    config.stripe_connect_enabled = false
+    config.enable_embedded_shopfronts = false
+    config.products_require_tax_category = false
+  end
```
sauloperez: Shouldn't this go in `config/environments/test.rb`? No need for an if there and that's the place where I'd expect all this kind of settings to be found.

spec/spec_helper.rb
```diff
@@ -50,6 +51,12 @@
 # Override setting in Spree engine: Spree::Core::MailSettings
 ActionMailer::Base.default_url_options[:host] = 'test.host'
 
+Spree::AppConfiguration.class_eval do
+  prepend Spree::TestingSupport::ConfigStub
+end
+
+Spree::Config.setup_config_stub!
```
sauloperez: Just for me to understand. We don't do the regular  `allow().to receive()` due to the amount of calls `Config` receives in integration tests?

In such case, I like your idea but I'd rather implement an object with the exact same API as `Config` (duck type) and replace the original with the mock in here.

Reopening a class and then changing the behaviour of `Config` feels very risky. We could be messing up the original behaviour without even noticing.

config/initializers/spree.rb
oeoeaio: Yeah, that’s fine. I wasn’t sure so I went with this, but happy to to test.rb instead.

spec/spec_helper.rb
oeoeaio: The reason we can’t use stubbing is that `Config` settings are read in three different ways, and it is impractical/impossible to stub all three.

The only problem with mimicing the existing config API in a mock object is that it would mean we need to come up with a way of providing all of the defaults.

I couldn’t think of a smart way of swapping out the the real config for a mock config, though I suppose I could just reassign `Spree::Config`, I didn’t even try that.

sauloperez: >  I suppose I could just reassign Spree::Config, I didn’t even try that.

I believe so.

oeoeaio:
Closing in favour of #1904

PR #1904, opened 2017-10-14
Use Spree's approach for preventing config caching when testing

oeoeaio:
#### What? Why?

TAKE 3
Solving the same problem as #1903 and #1742. I've also included Matt's work from #1896 using the same approach.

After doing some more investigating, it turns out Spree already has its own [built-in mechanism](https://github.com/openfoodfoundation/spree/blob/step-6a/core/lib/spree/testing_support/preferences.rb) for preventing persistence of config settings in the test environment. I have no idea how we missed it until now, but there you go.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -100,11 +100,13 @@ RSpec.configure do |config|
   # Ensure we start with consistent config settings
   config.before(:each) do
     reset_spree_preferences do |spree_config|
+      # These are all settings that differ from Spree's defaults
       spree_config.default_country_id = default_country_id
       spree_config.checkout_zone = checkout_zone
       spree_config.currency = currency
       spree_config.shipping_instructions = true
       spree_config.auto_capture = true
+      spree_config.allow_backorders = false
     end
 
     Spree::Api::Config[:requires_authentication] = true
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -29,6 +29,7 @@ require 'spree/api/testing_support/setup'
 require 'spree/api/testing_support/helpers'
 require 'spree/api/testing_support/helpers_decorator'
 require 'spree/testing_support/authorization_helpers'
+require 'spree/testing_support/preferences'
 
 # Capybara config
 require 'capybara/poltergeist'
@@ -93,8 +94,21 @@ RSpec.configure do |config|
   # Geocoding
   config.before(:each) { Spree::Address.any_instance.stub(:geocode).and_return([1,1]) }
 
+  default_country_id = Spree::Config[:default_country_id]
+  checkout_zone  = Spree::Config[:checkout_zone]
+  currency = Spree::Config[:currency]
   # Ensure we start with consistent config settings
-  config.before(:each) { Spree::Config.products_require_tax_category = false }
+  config.before(:each) do
+    reset_spree_preferences do |spree_config|
+      spree_config.default_country_id = default_country_id
+      spree_config.checkout_zone = checkout_zone
+      spree_config.currency = currency
+      spree_config.shipping_instructions = true
+      spree_config.auto_capture = true
+    end
+
+    Spree::Api::Config[:requires_authentication] = true
+  end
 
   # Helpers
   config.include Rails.application.routes.url_helpers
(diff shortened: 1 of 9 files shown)


config/initializers/spree.rb
```diff
@@ -18,6 +19,13 @@
   config.shipping_instructions = true
   config.address_requires_state = true
 
+  # Set Spree::Config defaults for testing
+  if Rails.env.test?
+    config.stripe_connect_enabled = false
+    config.enable_embedded_shopfronts = false
+    config.products_require_tax_category = false
+  end
```
oeoeaio: @sauloperez this stuff can't go in `test.rb` because `Spree::Config` is set in an initializer, so it doesn't exist yet. I will think about a better place to put them before this is merged.

spec/spec_helper.rb
```diff
@@ -93,15 +94,31 @@
...
+  config.before(:each) do
+    reset_spree_preferences do |spree_config|
+      # These are all settings that differ from Spree's defaults
+      spree_config.default_country_id = default_country_id
+      spree_config.checkout_zone = checkout_zone
+      spree_config.currency = currency
+      spree_config.shipping_instructions = true
+      spree_config.auto_capture = true
+      spree_config.allow_backorders = false
+    end
```
sauloperez: does it really need to happen for every single test example? this is going to add a considerable overhead to all the test suite.

oeoeaio: I agree that it is not ideal, but it does not seem to affect total spec runtime significantly. Spree's approach clears the preference cache before each spec, so there is no easy way to use customised default values without a) resetting them for each spec, or b) overriding the defaults set in Spree's `core/app/models/app_configuration.rb`.

I will try think of another way of achieving the same outcome.

sauloperez: > so there is no easy way to use customised default values without a) resetting them for each spec, or b) overriding the defaults set in Spree's core/app/models/app_configuration.rb.

Then there's not much else we can do here. We need to solve this problem anyway so I'd move forward even with any performance impact.

I just raise the issue because this is what I'd check for if I were to speed up the test suite. Normally, you can apply the particular settings only to the examples that require it by means of https://relishapp.com/rspec/rspec-core/docs/metadata/user-defined-metadata.

spec/spec_helper.rb
```diff
@@ -93,15 +94,31 @@
   # Geocoding
   config.before(:each) { Spree::Address.any_instance.stub(:geocode).and_return([1,1]) }
 
+  default_country_id = Spree::Config[:default_country_id]
+  checkout_zone  = Spree::Config[:checkout_zone]
+  currency = Spree::Config[:currency]
   # Ensure we start with consistent config settings
-  config.before(:each) { Spree::Config.products_require_tax_category = false }
+  config.before(:each) do
```
mkllnk: Is this how Spree does it as well? Can you add a reference?

spec/spec_helper.rb
oeoeaio: Yeah, we could try and work out which examples actually need which these 'defaults', but then we have the problem of polluting our test suite with settings that we should be able to assume are in a particular state. They are all set during app initialisation so allowing to default to something else for tests is confusing. I'd prefer to work out a way of setting these six preferences in a universal way, and only adding deviations from the assumed 'default' settings for a given example.

spec/spec_helper.rb
oeoeaio: https://github.com/openfoodfoundation/spree/blob/step-6a/core/spec/spec_helper.rb#L59

daniellemoorhead:
@oeoeaio does this need to be tested or can it be merged straight to master?

oeoeaio:
I need to fixed the conflicts, but after that it can be merged straight in.

oeoeaio:
Total test runtime on TravisCI seems to be really variable, so I tested this on SemaphoreCI (which seems to be far more consistent). The most consistent I have got it is on the latest iteration of the `rspec-upgrade` branch after @andypalmer's miracle-working there. So I used the `rspec-upgrade` branch as the benchmark and then merged this PR into it to compare spec runtimes, this is what I got:

`rspec-upgrade` branch (Total runtime: 26:36):
screen shot 2017-11-02 at 4 46 20 pm

`rspec-upgrade` branch + this branch (Total runtime: 26:42):
screen shot 2017-11-02 at 5 39 50 pm

So I would argue that if there is a performance hit caused by this change, it is not significant enough for us to be that worried about it at the moment.

sauloperez:
I would give it at least 3 tries and call it statistics :trollface: Jokes aside if we could see whether it consistently slows it down 10secs would be good.

daniellemoorhead:
@oeoeaio will merge this once the Rspec changes have been in for a couple of days and seem ok.

PR #1910, opened 2017-10-18
Fix #1902: ensure correct payment state when failed payments are present

### LINKED ISSUE #1902: Orders which had a failed payment attempt during checkout should not remain in 'failed' state
This is occurring when an checkout is attempted, but fails with Stripe, and then a successful checkout is achieved with a manual payment option (cash). The state is still showing as failed, which is not accurate.

image

oeoeaio:
#### What? Why?

Fixes #1902 

Problem: order payments didn't seem to be loaded in any particular order, and `OrderUpdater#update_payment_state` was relying on payment order to determine payment state.

Strategy: Adapt a version of this method from a future version of Spree. I tried to select a version where I would have to make the absolute minimum number of changes to get it to work (selected 2.4). See comments in code for justifications of the changes that I did make.

#### What should we test?

That payment states for completed orders make sense.

#### Release notes

Fixed a bug in the display of payment state for orders with failed payments

This PR uses a snippet of code from a future version of Spree (v2.4) to fix the issue. We should just be able to remove all code added in this PR completely when we get to that point.

### DIFF
--- a/app/models/spree/order_updater_decorator.rb
+++ b/app/models/spree/order_updater_decorator.rb
@@ -1,0 +1,41 @@
+module Spree
+  OrderUpdater.class_eval do
+    # TODO: This logic adapted from Spree 2.4, remove when we get there
+    # Handles state updating in a much more logical way than < 2.4
+    # Specifically, doesn't depend on payments.last to determine payment state
+    # Also swapped: == 0 for .zero?, .size == 0 for empty? and .size > 0 for !empty?
+    # See:
+    # https://github.com/spree/spree/commit/38b8456183d11fc1e00e395e7c9154c76ef65b85
+    # https://github.com/spree/spree/commit/7b264acff7824f5b3dc6651c106631d8f30b147a
+    def update_payment_state
+      last_state = order.payment_state
+      if payments.present? && payments.valid.empty?
+        order.payment_state = 'failed'
+      elsif order.state == 'canceled' && order.payment_total.zero?
+        order.payment_state = 'void'
+      else
+        # This part added so that we don't need to override order.outstanding_balance
+        balance = order.outstanding_balance
+        balance = -1 * order.payment_total if canceled_and_paid_for?
+        order.payment_state = 'balance_due' if balance > 0
+        order.payment_state = 'credit_owed' if balance < 0
+        order.payment_state = 'paid' if balance.zero?
+
+        # Original logic
+        # order.payment_state = 'balance_due' if order.outstanding_balance > 0
+        # order.payment_state = 'credit_owed' if order.outstanding_balance < 0
+        # order.payment_state = 'paid' if !order.outstanding_balance?
+      end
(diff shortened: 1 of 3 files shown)


app/models/spree/order_updater_decorator.rb
```diff
@@ -0,0 +1,40 @@
+module Spree
+  OrderUpdater.class_eval do
+    # This logic adapted from Spree 2.4, remove when we get there
```
mkllnk: What do you think about the TODO tag?
```diff
-# This logic adapted from Spree 2.4, remove when we get there
+# TODO: This logic adapted from Spree 2.4, remove when we get there.
```

sauloperez: Yes, please. Better to use TODOs

oeoeaio:
### Testing notes

#### Case 1: 🆗 
- Attempt to place < $0.50 order using Stripe payment method (fails)
- Use paypal instead (succeeds)
- Verify payment state is 'Paid' (yes)

Order state after 3rd step:
screen shot 2017-10-19 at 12 57 32 pm

Payments on order after 3rd step:
screen shot 2017-10-19 at 1 00 45 pm

#### Case 2: 🆗 
- Attempt to place < $0.50 order using Stripe payment method (fails)
- Use 'Cash' payment method instead (succeeds)
- Verify payment state is 'Balance Due' (yes)
- Capture payment (succeeds)
- Verify payment state is 'Paid' (yes)
- Void the payment
- Verify payment state is 'Balance Due' (yes)

Order state after 3rd step:
screen shot 2017-10-19 at 1 08 24 pm

Payments on order after 3rd step:
screen shot 2017-10-19 at 1 09 04 pm

Order state after 5th step:
screen shot 2017-10-19 at 1 10 09 pm

Payments on order after 5th step:
screen shot 2017-10-19 at 1 09 35 pm

Order state after 7th step:
screen shot 2017-10-19 at 1 11 15 pm

Payments on order after 7th step:
screen shot 2017-10-19 at 1 10 40 pm

#### Case 3: ❌ 
- Attempt to place < $0.50 order using Stripe payment method (fails)
- Add an item to push order total > $0.50
- Try Stripe payment method again (succeeds)
- Verify payment state is 'Paid' (NO)
- Remove an item
- Verify payment state is 'Credit Owed'
- Void the payment
- Verify payment state is 'Balance Due'

Order state after 4th step:
screen shot 2017-10-19 at 1 17 05 pm

Payments on order after 4th step:
screen shot 2017-10-19 at 1 18 30 pm

Looks like payment state isn't being updated properly for Stripe?

#### Case 4: need to get case 3 working first
- Place > $0.50 order using Stripe payment method (succeeds)
- Verify payment state is 'Paid'
- Cancel the order
- Verify payment state is 'Credit Owed'

oeoeaio:
Ok, so the issue I came across while testing this is a separate issue not caused by this change. I've documented and fixed it in #1914. This PR is blocked until that one goes through.

PR #1914, opened 2017-10-19
Do not update finalized fee adjustments

oeoeaio:
#### What? Why?

As I was testing #1910, I came across an issue whereby transaction fees associated with failed payments, which had been marked as `eligible: false` and `finalized` seemed to be re-opened at the end of the checkout process, resulting in an additional fee 'appearing' the next time the completed order was updated (adding/removing items etc).

The culprit is a set of `after_save` callbacks on `Spree::Order`, which are designed to update shipping and payment fees in the event that a completed order has items added/removed. The way this works is that all such fees are placed in the `open` state, and then updated. This process was opening the `finalized` failed payment adjustment and reseting the `eligible` flag to `true`, meaning that subsequent updates would incorrectly consider it relevant to the calculation of the order total. 

The fix I have provided simply adds a guard clause to the updater method, allowing it to return early if the adjustment is in the `finalized` state.

I have added additional `TODO` comments suggesting that this logic may be better placed in a controller rather than an `after_save` callback...

#### What should we test?

Completed orders with failed payments should be able to updated any number of times without additional `Transaction fees` from failed payments appearing and affecting the order total.

Shipments marked as `shipped` should not be updated when changes are made to the order that would otherwise cause them to be adjusted.

#### Release notes

Fixed a bug that caused failed payments to add a transaction fee to an order when it was updated by an enterprise user.

### DIFF
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -144,6 +144,8 @@ Spree::Order.class_eval do
   end
 
   # After changing line items of a completed order
+  # TODO: perhaps this should be triggered from a controller
+  # rather than an after_save callback?
   def update_shipping_fees!
     shipments.each do |shipment|
       next if shipment.shipped?
@@ -153,6 +155,8 @@ Spree::Order.class_eval do
   end
 
   # After changing line items of a completed order
+  # TODO: perhaps this should be triggered from a controller
+  # rather than an after_save callback?
   def update_payment_fees!
     payments.each do |payment|
       next if payment.completed?
@@ -367,6 +371,7 @@ Spree::Order.class_eval do
   end
 
   def update_adjustment!(adjustment)
+    return if adjustment.finalized?
     state = adjustment.state
     adjustment.state = 'open'
     adjustment.update!(self)
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -687,6 +687,26 @@ describe Spree::Order do
         expect(order.adjustment_total).to eq expected_fees - shipping_fee - payment_fee
         expect(order.shipment.adjustment.included_tax).to eq 0.6
       end
+
+      context "when finalized fee adjustments exist on the order" do
+        let(:payment_fee_adjustment) { order.adjustments.payment_fee.first }
+        let(:shipping_fee_adjustment) { order.adjustments.shipping.first }
+
+        before do
+          payment_fee_adjustment.finalize!
+          shipping_fee_adjustment.finalize!
+          order.reload
+        end
+
+        it "does not attempt to update such adjustments" do
+          order.update_attributes(line_items_attributes: [{id: order.line_items.first.id, quantity: 0}])
+
+          # Check if fees got updated
+          order.reload
+          expect(order.adjustment_total).to eq expected_fees
+          expect(order.shipment.adjustment.included_tax).to eq 1.2
+        end
+      end
     end
 
     context "changing the shipping method to one without fees" do


app/models/spree/order_decorator.rb
```diff
@@ -144,6 +144,8 @@ def add_variant(variant, quantity = 1, max_quantity = nil, currency = nil)
   end
 
   # After changing line items of a completed order
+  # TODO: perhaps this should be triggered from a controller
```
mkllnk: I think I had a similar thought when I last looked at this code.

app/models/spree/order_decorator.rb
```diff
@@ -367,6 +371,7 @@ def ensure_customer
   end
 
   def update_adjustment!(adjustment)
+    return if adjustment.finalized?
```
mkllnk: This actually makes sense for shipments as well. If something has been shipped, the shipping costs occurred and should not be changed.

The only case that may complicate things:
- Alice orders a, b and c from a shop and pays shipping fees depending on the number of items.
- The shop sends a box to Alice.
- Alice discovers that item c is missing and complains.
- The shop takes item c out of the original order, shipping fees are not updated.
- The shop has to manually add a credit or refund.
- The previous code would have adjusted the fee and the customer's balance would change.

app/models/spree/order_decorator.rb
sauloperez: I don't like to put it neither in the model as a callback nor the controller. Big classes are unmaintenable. Let's extract this into a new abstraction like a service object.

oeoeaio: @sauloperez I’m always happy to abstract detail down into a service object. 👍 My comment was just a suggestion that a controller action is perhaps a more appropriate hook.

sauloperez: as I hook, yes. I'm just troubled with so many nasty callbacks :see_no_evil:

oeoeaio: Me too. 😔 Unfortunately we inherit a lot of them from Spree, but there are plenty of our own that we can work on removing! 💪

app/models/spree/order_decorator.rb
oeoeaio: Good point. Do you think we need to preserve this particular case, or is it rare enough that we don't need to support it?

The fact that this case would "work" as it currently stands is an artefact of these `after_save` callbacks which we added, right? I'm not sure that handling this case was part of the design. As far as I know, doing the same thing in an unaltered deployment of Spree would not cause the shipping fee to be updated.

mkllnk: I think it's rare enough. I don't think it was the initial intent of this code, just a by-product. So I'm happy for this change to got through.

oeoeaio: Just tested this. This case isn't even possible, because removing shipped items from an order is not allowed:

screen shot 2017-10-20 at 12 49 26 pm

oeoeaio:
### Testing notes

Removing 'shipped' items from an order is not possible (see my comment above), but I tested adding items to the order after shipping the existing set of items and the shipment total was not updated. This is the desired behaviour, because items added to an order after a shipment has been shipped should not affect that shipment's fee, they should be added to a new shipment and charged separately.

### Test 1: Failed payments 🆗 

On an order with a failed payment that would have attracted a transaction fee:
screen shot 2017-10-20 at 1 40 39 pm

Adding items to the order in the backend doesn't cause a transaction fee for the failed payment to be added to the order:
screen shot 2017-10-20 at 1 42 44 pm

All good 👍 

#### Test 2: Shipped shipments and complete payments 🆗 

Using 10% fees for both the shipping and payment fees, initial state (un-captured payment and unshipped shipment):
screen shot 2017-10-20 at 1 25 15 pm

Adding an item causes both fees to be updated:
screen shot 2017-10-20 at 1 25 47 pm

I removed the item (which reset the fees) and then captured the payment, allowing the shipment to be `ready`:
screen shot 2017-10-20 at 1 27 14 pm

Adding an item now causes the payment fee to remain intact (since the payment has been finalised) and the shipping fee to be updated (note that the shipment is not no-longer `ready`):
screen shot 2017-10-20 at 1 28 54 pm

I removed the item again (to get the shipment back to `ready`), and then shipped the shipment:
screen shot 2017-10-20 at 1 33 53 pm

Adding an item now updates neither the shipment or payment fees:
screen shot 2017-10-20 at 1 34 57 pm

I think this is good to go 👍

mkllnk:
Your testing notes make sense to me. This is a really good improvement of the current behaviour. :-)

PR #1915, opened 2017-10-19
Feature/enable i18n fallback to en

saimonmoore:
#### What? Why?

Right now in ofn, the process of getting i18n keys translated and into code is a manual process and prone to errors. Due to this, you may have cases that despite the translations have been done in transifex, these do not get into the code when they should and you get  `missing translation` messages in the UI.

This is a very ugly experience for the user so this PR attempts to ensure that, we always fallback to the `en` locale if the key is not found in the current locale (or any available locales).

This PR modifies the `config.i18n.available_locales` list to ensure that both the `I18n.default_locale` and the `en` locale are always *available* and as such they are available both for backend translations and frontend translations (i18n-js gem uses `config.i18n.available_locales` to filter which locales it exports to the frontend).

Additionally, we ensure that the list of I18n fallbacks always includes `en` in ALL environments. This means that even if the default locale of any particular instance of ofn is a non-english based locale (the i18n gem will automatically derive `en` from english based locales), it will still at least use `en` as a fallback if there are no translations found for that locale.

#### What should we test?

- [ ] Missing translations fallback to `en` in both backend and frontend, irrespective of the locale and the environment.

#### Release notes

- Now, i18n translations will always fallback to `en` in both backend and frontend, irrespective of the locale and the environment.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -98,7 +98,8 @@ module Openfoodnetwork
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     config.i18n.default_locale = ENV["LOCALE"] || ENV["I18N_LOCALE"] || "en"
-    config.i18n.available_locales = ENV["AVAILABLE_LOCALES"].andand.split(',').andand.map(&:strip) || [config.i18n.default_locale]
+    config.i18n.available_locales = ENV["AVAILABLE_LOCALES"].andand.split(/[\s,]/).andand.map(&:strip) || []
+    config.i18n.available_locales = (config.i18n.available_locales + [config.i18n.default_locale, 'en']).uniq
     I18n.locale = config.i18n.locale = config.i18n.default_locale
 
     # Setting this to true causes a performance regression in Rails 3.2.17
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -56,7 +56,7 @@ Openfoodnetwork::Application.configure do
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation can not be found)
-  config.i18n.fallbacks = true
+  config.i18n.fallbacks = [:en]
 
   # Send deprecation notices to registered listeners
   config.active_support.deprecation = :notify
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -56,7 +56,7 @@ Openfoodnetwork::Application.configure do
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation can not be found)
-  config.i18n.fallbacks = [:en]
+  config.i18n.fallbacks = true
 
   # Send deprecation notices to registered listeners
   config.active_support.deprecation = :notify
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -29,6 +29,8 @@ Openfoodnetwork::Application.configure do
   # Expands the lines which load the assets
   config.assets.debug = false
 
+  config.i18n.fallbacks = [:en]
+
   # Show emails using Letter Opener
   config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.default_url_options = { host: "0.0.0.0:3000" }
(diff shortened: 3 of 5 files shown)


config/environments/production.rb
```diff
@@ -56,7 +56,7 @@
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation can not be found)
-  config.i18n.fallbacks = [:en]
+  config.i18n.fallbacks = true
```
oeoeaio: Won't this break the desired behaviour of falling back to `:en` when `default_locale` is missing a translation? In other words, shouldn't we keep this as `[:en]`? Or am I misunderstanding something?

If we can keep it as `[:en]`, can we then move it to `application.rb` and remove it from all of the individual environment config files?

oeoeaio: Bump @saimonmoore

saimonmoore: shit...that change shouldn't have made it into the PR. Good catch @oeoeaio. Not sure about KISSing it though. I think it's better to keep it in the inidividual files as I understand ofn may start to support multiple locales.

oeoeaio: 👍  sounds good.

config/application.rb
```diff
@@ -98,7 +98,8 @@ class Application < Rails::Application
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     config.i18n.default_locale = ENV["LOCALE"] || ENV["I18N_LOCALE"] || "en"
-    config.i18n.available_locales = ENV["AVAILABLE_LOCALES"].andand.split(',').andand.map(&:strip) || [config.i18n.default_locale]
+    config.i18n.available_locales = ENV["AVAILABLE_LOCALES"].andand.split(/[\s,]/).andand.map(&:strip) || []
+    config.i18n.available_locales = (config.i18n.available_locales + [config.i18n.default_locale, 'en']).uniq
```
oeoeaio: `config.i18n.available_locales |= [config.i18n.default_locale, 'en']`?

daniellemoorhead:
Gentleman, do we need this tested or can it go directly to merge?

oeoeaio:
Definitely needs to be tested.

oeoeaio:
### Testing Notes

I started off by deploying this to our staging server. I then opened an ssh session, switched the locale to `fr` in `application.yml` and added the following to the shops index template:

````haml
-# app/views/shops/index.html.haml
...
= t('this_should_be_translated')
{{ 'this_should_be_translated_too' | t }}
...
````
After restarting unicorn, I loaded up the shop index and this is what I got:
screen shot 2017-11-01 at 12 13 28 pm
That is pretty much expected. The fact that the JS is not converting the key to a string in the way that rails does is a separate issue.

I then added two new strings to `en.yml`:

````yaml
# config/locales/en.yml
en:
  this_should_be_translated: "This should be translated, and it is yay!"
  this_should_be_translated_too: "This should be translated too, and it is yay!"
````

After doing `bundle exec rake assets:procompile` and restarting unicorn I get:
screen shot 2017-11-01 at 12 14 58 pm

So pure Rails `I18n.t` translation is working great, and falling back to the `en` translation. 👍 
Looks like the JS translation is not being compiled properly (I suspect this is a separate issue). 👎 

Switching the locale to `en` and doing `bundle exec rake assets:precompile` did not fix the problem, so I deleted `darkswarm.all-xxxxxxxxx.js` from `public/assets` and tried again.

This time it worked:
screen shot 2017-11-01 at 12 34 00 pm
 
And the fallback worked after switching the locale back to `fr`:
screen shot 2017-11-01 at 12 36 02 pm

So the issue we're having is that `i18n-js` doesn't seem to be very good at picking up changes to `en.yml` and forcing the JS translations to be re-compiled. I've noticed this is development too.

I think this is probably ok to go through, but we should fix compilation issue ASAP. Anyone else have any thoughts?

saimonmoore:
Yeah, I think you need to run `rake tmp:cache:clear` after each eployment
for the js to pick up the changes correctly.

El dc., 1 de nov. 2017 a les 3:27, Rob H (<notifications@github.com>) va
escriure:

oeoeaio:
@saimonmoore our deploy scripts do that already, though I must admit I didn't try it between precompile runs in the test above. Maybe I should try. Doesn't seem to help on my local machine...

saimonmoore:
Strange.

El dj., 2 de nov. 2017 a les 8:51, Rob H (<notifications@github.com>) va
escriure:

PR #1918, opened 2017-10-20
Transifex

oeoeaio:
#### What? Why?

Merging new Transifex translations for release of v1.10


config/locales/en-GB.yml
```diff
@@ -1,4 +1,4 @@
-en-GB:
+en_GB:
```
oeoeaio: @lin-d-hop is this definitely what you want to do?

Issue #1799, opened 2017-09-05
Missing translations in the "new shipping method" page


myriamboure:
![illustration](https://user-images.githubusercontent.com/8135376/30076322-c3bbeaf6-9278-11e7-81d3-12896385ff1d.jpeg)

myriamboure:
Hi @sauloperez I created a few more missing translation issues following your model should I add them to the September milestone as well ? Please do it if that seems reasonable to you ;-)

daniellemoorhead:
Hey, if you're adding things to the september release then just the PRs, not the issues. Thanks :)

sauloperez:
Being done in #1799

g33kyshivam:
Hey @sauloperez . I would like to work on this issue !

sauloperez:
Hi @g33kyshivam! Unfortunately @yasirazgar already worked on it in #1799 but you can check out https://github.com/openfoodfoundation/openfoodnetwork/labels/good%20first%20issue. There's plenty of work you can help with

sauloperez:
DOne in #1921

PR #1921, opened 2017-10-21
1799 - Missing translations in new shipping methods page

yasirazgar:
Issue - #1799 
#### What? Why?

Translations were missing in new shipping methods page (admin/shipping_methods/new).
Fixed this by removing hardcoded strings and added missing translations in locale file.

#### What should we test?

Test whether translations are available for all the strings in new shipping methods page (admin/shipping_methods/new).

### DIFF
--- a/app/assets/javascripts/templates/admin/tags_input.html.haml
+++ b/app/assets/javascripts/templates/admin/tags_input.html.haml
@@ -1,4 +1,5 @@
 %tags-input{ template: 'admin/tag.html',
+  "placeholder" => t('admin.order_cycles.form.add_a_tag'),
   ng: { model: 'object[tagsAttr]', class: "{'limit-reached': limitReached}"},
   on: { tag: { added: 'tagAdded()', removed:'tagRemoved()' } } }
   %auto-complete{ ng: { if: "findTags" }, source: "findTags({query: $query})",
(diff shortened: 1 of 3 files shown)


yasirazgar:
@sauloperez, please let me know if I am missing something.
I have updated only en.yml file, need to update other locale files, so I need translations for 
```
pick_up: Pick up
add_a_tag: Add a tag
```
Also I have one questions, I can't find translation for some of the strings in the locale file, 
`ex I18n.t(:zone)`, but they are available in the system, is it pulling from some external api ?

sauloperez:
Awesome @yasirazgar and @mohammedyasir! What about the two other highlighted copies?

yasirazgar:
@sauloperez, I checked both translations are already available for 
```
I18n.t('back_to_shipping_methods_list')
I18n.t(:category)
```
for `category` it was hardcoded, I changed it to use for the locale, 
and I grepped for `back_to_shipping_methods_list`, can't find out, but I believe its using from backend/app/views/spree/admin/shipping_methods/new.html.erb, where the translation is correctly used. Please correct me if I am missing something.

And sorry for two github accounts, I have amended the commit which came from the wrong user.

yasirazgar:
@sauloperez, there are various parts in this app, I am not sure how it communicates and works, is there any docs where I can learn the architecture and other stuff about this app.
Thanks!

sauloperez:
I can confirm it works :v: 

qa

sauloperez:
> @sauloperez, there are various parts in this app, I am not sure how it communicates and works, is there any docs where I can learn the architecture and other technical stuffs about this app.
Thanks!

Sure. You can find more details in https://github.com/openfoodfoundation/openfoodnetwork/wiki and https://github.com/openfoodfoundation/ofn-install/wiki

That not might be exactly what you want but don't hesitate to ask. Perhaps we can work on better architectural documentation.

sstead:
Looking fine on Staging1 @oeoeaio 
image

PR #1922, opened 2017-10-23
Uk/1663variant ordering emails

Duende13:
This order by variant name and unit value in the items of an order are applied to confirmation email and summary screen after checkout for clients (not producers). It was done in the cart and now in emails.
Requested in issue #1663 
NOTE : All these changes were already approved and reviewed on PR #1709 , just the mentioned PR was merged to a branch instead to the master.

What? Why?

What should we test?

### DIFF
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -26,7 +26,7 @@ Spree::LineItem.class_eval do
   }
 
   # Find line items that are from order sorted by variant name and unit value
-  scope :sorted_by_name_and_unit_value, joins(:variant => :product).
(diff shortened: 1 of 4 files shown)


daniellemoorhead:
@lin-d-hop do you want the AU team to stage and test this or will you handle that on your end?

lin-d-hop:
@OliverUK Can you give this a test when you get the chance?

OliverUK:
Yes, @lin-d-hop, once there is a build for it.

daniellemoorhead:
Might be worth getting Travis back to green before/whilst Oliver is testing this @lin-d-hop @Duende13 🙂

OliverUK:
Happy to test this once a build is available in buildkite @lin-d-hop

Issue #1075, opened 2016-07-01
Clicking on product image in checkout takes me to the homepage


sstead:
![image](https://cloud.githubusercontent.com/assets/10843056/16508167/c3129d2e-3f73-11e6-8fb4-69a51d803365.png)

Product images are clickable. I'd expect the product details window to open, but instead it takes me to the homepage.

sineadfenton:
Hey @sstead 

Just wondering on this one (sorry know its old). But are you referring products essentially having their own URLs?

We've had a request in from a user about products having the own URLs so when you share a product directly for marketing purposes but I could see this tying into this too where clicking on the image would open a new window or iframe with the product and not the homepage?

S

sstead:
Hi @sineadfenton, I suppose most immediately we just need to remove the confusing homepage redirect. Either by making the image not clickable (this might be easiest), or by having it open the product modal that you see in the shop (e.g. below). 
![image](https://user-images.githubusercontent.com/10843056/31258705-ea21e728-aa8c-11e7-9514-c156fa472d4f.png)

Giving products their own URL would be a next step. I haven't heard requests for this in Australia, but I can see how it would be helpful for marketing. If we did have unique product URLs, for sure it would make sense that clicking on the product image would open that URL in a new tab.

sineadfenton:
Hey @sstead agreed, think focusing on the redirect in the first instance is a good first step.

I'll open up a discussion on Discourse around the unique URLs and have that discussion separately so as to keep things on topic here!

S

daniellemoorhead:
Hey @sstead can this be closed? #1923 is merged and you created #1925 to cover off the rest? If it's done then please close this, else let us know what is remaining to be done.

daniellemoorhead:
Closed as this is done.

PR #1923, opened 2017-10-24
Disabled product links in cart

andypalmer:
#### What? Why?

This is an interim solution to #1075 which is due to product urls being explicitly rerouted to the homepage in routes.rb
Depending on priority, I suggest raising a further issue to make the behaviour of the products in the cart match the experience of the products in the shop (modal with product description)

#### What should we test?

The shopping cart spec has been updated to explicitly check for links under the product image. Manual testing is just making sure that the images are now non-clickable

I specifically chose to use the Deface functionality to implement this, rather than inserting it into the overridden _line_item template. Using Deface rather than overriding the templates makes it easier to upgrade the Spree version. The current changes to this template could (and maybe ought to) be migrated from the override to Deface.

### DIFF
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -27,6 +27,14 @@ feature "full-page cart", js: true do
       Spree::Config.allow_backorders = allow_backorders
     end
 
+    describe "product description" do
+      it "should not try to link to the product page" do
+        add_product_to_cart order, product_fee, quantity: 2
+        visit spree.cart_path
+        expect(page).to_not have_selector '.item-thumb-image a'
+      end
+    end
+
     describe "fees" do
       let(:percentage_fee) { create(:enterprise_fee, calculator: Calculator::FlatPercentPerItem.new(preferred_flat_percent: 20)) }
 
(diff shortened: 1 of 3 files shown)


andypalmer:
close and re-open to retrigger build

andypalmer:
There are two tests that seem to be failing consistently. The order variant one passed when I inserted a byebug into the flow, so it might be a timing issue. Investigating

andypalmer:
The problem is that, for some reason, the Deface of the partial causes string interpolation to fail. So where we're expecting a row with a class of `variant-307` we're actually getting a class of `variant-#variant.id`

spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -27,6 +27,14 @@
       Spree::Config.allow_backorders = allow_backorders
     end
 
+    describe "product description" do
+      it "should not try to link to the product page" do
```
sauloperez: We prefer the imperative style, so I'd type "does not link..."

sauloperez:
This is your first PR, right? In that case, welcome to OFN! 🎉

sstead:
Testing notes:
Looks good. The product images at Edit Cart are no longer clickable. Everything else in the checkout is behaving normally. 
image
Staging1 was not displaying old images correctly, but newly uploaded images are fine. This isn't related to this PR.

Have created a new issue for further, lower priority improvement here - https://github.com/openfoodfoundation/openfoodnetwork/issues/1925 This would have the images be clickable, opening the product modal.

PR #1928, opened 2017-10-26
Rspec upgrade

oeoeaio:
#### What? Why?

Rebase of #1550, there were quite a few things altered during the rebase, so suggest we close that one and restart the conversation here.

I have a couple of concerns:
- The runtime for the suite seems to have gotten quite a lot longer
- There are still some intermittent spec failures

#### What should we test?

That the test suite works and is stable.

#### Release notes

Tech: upgrade of test suite runner

#### Dependencies

#1550 (not dependency but relevant)
#1904 (possibly?)


enricostano:
@oeoeaio can we close https://github.com/openfoodfoundation/openfoodnetwork/pull/1550 then? Would be nice to have 1 PR solving an issue. Thoughts?

Thanks for moving this forward!!!

spec/spec_helper.rb
```diff
@@ -90,6 +90,14 @@
     DatabaseCleaner.clean
   end
 
+  def restart_phantomjs
+    Capybara.send('session_pool').values
+      .select { |s| s.driver.is_a?(Capybara::Poltergeist::Driver) }
+      .each { |s| s.driver.restart}
+  end
+
+  config.before(:all) { restart_phantomjs }
```
sauloperez: Why do we need this?

oeoeaio: tl; dr: I don't think we need it.

@andypalmer if you could weigh in on this discussion that would be great. Please correct me if I'm misrepresenting your work/thinking.

@sauloperez this is a solution (provided by Andy) to the problem of not being able to run the entire test suite from start to finish without a memory leak in Poltergeist/PhantomJS causing a crash. ie. this is a fix for the issue we discussed at the last dev hangout.

As a test, I removed the two commits that added this solution and this is what I got when running the whole suite in Semaphore:
screen shot 2017-11-03 at 10 30 06 am
https://semaphoreci.com/openfoodfoundation/openfoodnetwork/branches/remove-restart/builds/1

The same build passed perfectly in TravisCI (across 5 workers):
https://travis-ci.org/oeoeaio/openfoodnetwork/builds/296528878

However, I noticed that there errors in Semaphore often mentioned `Capybara.reset_sessions!`, which @mkllnk added to `spec_helper.rb` a few years ago in 2488411b943682f8842851eeff05c160e02c1c47 when trying to prevent deadlocks involving the database cleaner. So I removed the `Caprbara.reset_session!` call from `spec_helper.rb` as an experiment, and that built with only two spec failures in Semaphore:
screen shot 2017-11-03 at 11 28 29 am

https://semaphoreci.com/openfoodfoundation/openfoodnetwork/branches/remove-restart/builds/2

And built perfectly in TravisCI (quite fast too):
https://travis-ci.org/oeoeaio/openfoodnetwork/builds/296540525

Both of the specs that failed in the Semaphore build are on my list of problematic specs for this PR. I have seen them both at least twice before with Andy's original solution (this one), so I doubt that they are caused by the removal of `Capybara.reset_sessions!`

Problematic specs (I've observed failures for each of these at least twice on builds of 4796978):
`spec/features/admin/bulk_order_management_spec.rb:440`
`spec/features/consumer/shopping/embedded_shopfronts_spec.rb:64`
`spec/controllers/spree/users_controller_spec.rb:30`

andypalmer: @oeoeaio  @sauloperez `restart_phantomjs` is there to address a memory leak with V8, which propagates to phantomjs, to poltergeist and then to capybara, leading to dying clients (and failing specs). It's mainly for running the whole suite in one go. The issue is mitigated by spreading the suite across multiple workers, but given that we have a way to make the tests reliable under all circumstances, it makes sense to keep it in.

The `bulk_order_management_spec` issue is separate, and is caused by the particular selector used. There was nothing _wrong_ with it per se, but it was way more specific than it needed to be (`div.select2-container#s2id_supplier_filter`). As the HTML spec (both 4.01 and 5) states that an ID should be unique within a document, it's possible (and fixes the problem) to reduce the selector to the ID only.

andypalmer: both of these issues could probably be resolved by switching to something other than poltergeist, but that's not going to be a trivial change. capybara-webkit, for example, requires setting up a virtual framebuffer to run in true headless mode.

oeoeaio: Thanks @andypalmer. At the risk of getting off topic, I don't think the issue with `bulk_order_management_spec ` has been resolved by d40e55f094c9c9796d894bc2f567506fab4c22fe, because I'm still seeing this:

`Failure/Error: expect(page).to have_field "start_date_filter", with: one_week_ago`

andypalmer: Ok. I'll see if I can prompt that failure at my end.
That's a symptom of us testing implementation rather than intent. I've got some ideas to resolve that, but that's a longer term plan

oeoeaio: Update to my earlier comment: I re-ran the specs for the second case (with both `restart_phantomjs ` and `Capybara.reset_sessions!` removed) and I get lots of failures again. So I think that `restart_phantomjs` is definitely required to prevent memory leaks. Still unsure of the effect of `Capybara.reset_sessions!`.

sauloperez:
Yay! :tada: thanks everyone involved! so much needed

PR #1931, opened 2017-10-30
Fix 1664: product tag rules not working

### LINKED ISSUE #1664: Tag rules are not working for show/hide inventory variants
This came up in Spree testing. I checked on Staging2 and it's occuring there as well, so probably in production too..

Problem described below:
image

The above scenario shows the 'by default' tag rules. Note that tag rules applied to product variants are also not working. E.g if by default salmon was visible, but I made a tag rule that hid it from certain customers, it would not hide.

Other tag rule (relating to shipping methods, payment methods and OCs) are working.

oeoeaio:
#### What? Why?

Fixes #1664: product tag rules are not working

Caused by a combination of: fe7bd5e2cd35122245ba5ed93003a315e61c7ff2 and 38d3b446cc171a34b9d939fd3402f810a7c8477b

Have added/tweaked specs to prevent this from happening again

#### What should we test?

Tag rules should work properly to show and hide products in the shopfront

#### Release notes

- Fixed a regression bug that prevented tag rules from filtering products properly in the shopfront

### DIFF
--- a/spec/lib/open_food_network/products_renderer_spec.rb
+++ b/spec/lib/open_food_network/products_renderer_spec.rb
@@ -73,6 +73,11 @@ module OpenFoodNetwork
         Spree::Product.any_instance.stub(:primary_taxon).and_return taxon
         pr.products_json.should include taxon.name
       end
+
+      it "loads tag_list for variants" do
+        VariantOverride.create(variant: variant, hub: distributor, tag_list: 'lalala')
+        expect(pr.products_json).to include "[\"lalala\"]"
+      end
     end
 
     describe "loading variants" do
@@ -96,7 +101,6 @@ module OpenFoodNetwork
         expect(variants[p.id]).to include v1, v3
         expect(variants[p.id]).to_not include v4
       end
-
 
       context "when hub opts to only see variants in its inventory" do
         before do
--- a/app/serializers/api/variant_serializer.rb
+++ b/app/serializers/api/variant_serializer.rb
@@ -1,6 +1,7 @@
 class Api::VariantSerializer < ActiveModel::Serializer
   attributes :id, :is_master, :count_on_hand, :name_to_display, :unit_to_display, :unit_value
   attributes :options_text, :on_demand, :price, :fees, :price_with_fees, :product_name
+  attributes :tag_list
 
   def price
     object.price
@@ -21,5 +22,11 @@ class Api::VariantSerializer < ActiveModel::Serializer
 
   def product_name
     object.product.name
+  end
+
+  # Used for showing/hiding variants in shopfront based on tag rules
+  def tag_list
+    return [] unless object.respond_to?(:tag_list)
+    object.tag_list
   end
 end
--- a/spec/serializers/variant_serializer_spec.rb
+++ b/spec/serializers/variant_serializer_spec.rb
@@ -1,27 +1,26 @@
 require 'spec_helper'
 
 describe Api::VariantSerializer do
-  
   subject { Api::VariantSerializer.new variant }
   let(:variant) { create(:variant) }
 
-  it "includes the expected attributes " do 
+  it "includes the expected attributes" do
     expect(subject.attributes.keys).
     to include(
         :id,
-        :name_to_display, 
-        :is_master, 
-        :count_on_hand, 
-        :name_to_display, 
-        :unit_to_display, 
+        :name_to_display,
+        :is_master,
+        :count_on_hand,
+        :name_to_display,
+        :unit_to_display,
         :unit_value,
-        :options_text, 
-        :on_demand, 
-        :price, 
-        :fees, 
-        :price_with_fees, 
-        :product_name
+        :options_text,
+        :on_demand,
+        :price,
+        :fees,
+        :price_with_fees,
+        :product_name,
+        :tag_list # Used to apply tag rules
       )
   end
-
 end


spec/lib/open_food_network/products_renderer_spec.rb
```diff
@@ -73,6 +73,11 @@ module OpenFoodNetwork
         Spree::Product.any_instance.stub(:primary_taxon).and_return taxon
         pr.products_json.should include taxon.name
       end
+
+      it "loads tag_list for variants" do
+        VariantOverride.create(variant: variant, hub: distributor, tag_list: 'lalala')
+        pr.products_json.should include "[\"lalala\"]"
```
sauloperez: Please, use the `expect` syntax :pray:

oeoeaio: 😱 Copy/paste/brain fart

sauloperez: LOL

lin-d-hop:
Very happy to see this PR. UK are going to test with priority as soon as we can deploy again.... :-/

sineadfenton:
@lin-d-hop  give us the heads up when ready to test

sineadfenton:
[Restricting Customer Product Views](https://trello.com/c/9jS4eWUy/109-restricting-customer-product-views)

lin-d-hop:
Oliver tried and had trouble with order cycles not loading. This is likely because there are now higher memory overheads and staging uses a smaller server. So I am going to upgrade the staging server to enable testing of this.

This also puts more optimisations higher on the agenda me thinks.

OliverUK:
Please let me know when I can test again, @lin-d-hop

lin-d-hop:
@OliverUK I just successfully opened an order cycle on staging. Try again?

OliverUK:
@lin-d-hop not working for me in Safari or Chrome.
screen shot 2017-11-04 at 15 56 05

lin-d-hop:
conversations in slack eh?

sineadfenton:
These are the steps I took during testing, unfortunately the variant still shows (unless I've missed a step)

1.  Create a tag on the OFN enterprise called wholesale (all lowercase) and set to be NOT VISIBLE. I also tried this process using a different word - hiding - and the same occured
image

2. Go to my Products > Inventory and then add the tag wholesale to the 15kg rice
image

3. I then go to my customer list and change Sinead under the OFN shop to have the wholesale tag
image

4. Create two order cycles. 
 **Test 1 - No Hiding has no tag rules**
 image

**Test 3 - Wholesale Hiding has the tag rule whole on it**
image

5. I log in as Sinead and view both cycles and see the 15kg rice in both
  **Test 1 (no hiding)**
image

**Test 3 (wholesale hiding)**
image

NickWeir63:
Hi @sineadfenton i wonder if you might have missed a step......  I  think that in your top screenshot you need to put the word 'wholesale' in the box that says 'Add a tag'.  Let me know if you want me to make this change in the UK demo hub

sineadfenton:
Thanks @NickWeir63! Sorry @lin-d-hop !

Okay, so the reverse seems to happen now - progress! Where the order cycle Testing 1 No Hiding hides the rice whereas the one setup to do hide, show its.

Test 1 - No Hiding
image

Test 3 - Hiding
image

OliverUK:
@NickWeir63 the screen shows that variants tagged wholesale should not be visible. Even if no customer is tagged it should not appear. The tagging of a customer would be used to make the product visible for some. 

@sineadfenton I think it confuses things that you also put a tag in the order cycle. Wouldn't it be best to test the tagging of variants independent of this? I used the feature for hiding items and showing them only to some members and I never did anything in the OC for this to work. Who knows how the two interact when you do both? Better to test them independently and then test what happens if you do both.

sineadfenton:
@OliverUK  yep, I actually got confused with a very very similar issue I was looking at which was around order cycles and hiding variants by cycles using the above method.

So taking a step back, ignoring order cycles **the variants are hidden** when we look at the images in my most recent comment - with the **wholesale** tag setup (ignoring order cycle)

image

OliverUK:
**TESTING NOTES**

Shop with one hidden order cycle and one visible one. Products in each OC that are visible and not visible.

screen shot 2017-11-10 at 09 33 23
screen shot 2017-11-10 at 09 20 36
screen shot 2017-11-10 at 09 33 48


Customer A tagged oliver
Customer B tagged member
Customer C tagged hello

screen shot 2017-11-10 at 09 34 11


A is allowed to see the hidden OC and hidden items
B is allowed to see hidden OC but not hidden items 
C is allowed to see hidden items but not hidden OCs

screen shot 2017-11-10 at 09 50 37
screen shot 2017-11-10 at 09 50 44
screen shot 2017-11-10 at 09 50 53


**RESULT** 

A - can see bidden OC and hidden items within both hidden and visible OC- **working**
B - can see hidden OC but not hidden items - **working**
C - can see hidden items but not hidden OCs - **working**
visitor - can see only visible OCs and only visible items within it - **working**

Note 1: it would be nice if someone how can see only one order cycle wasn’t asked to choose it from a dropdown that only contains on item
Note 2: tags sometimes don’t save; it’s best to always go back to check that they saved and it may be necessary to tab out of the tag field before saving

**HIDE SHIPPING AND PAYMENT METHODS**

In addition to the above customer A was allowed to see hidden payment and shipping methods. In each case one method of two was visible to all and one hidden. The feature worked as expected.

sineadfenton:
[Hidden Order Cycles Still Visible - Tagging not working](https://trello.com/c/bepZ00lN/108-hidden-order-cycles-still-visible-tagging-not-working)

sstead:
@oeoeaio looks good to me. One little thing I noticed below, but this is perhaps something to do with Staging1?

image

PR #1934, opened 2017-10-30
Stripe admin rebase

stveep:
#### What? Why?
Adds Stripe Connect to the backend so that payments can be manually added by admins using Stripe.

#### What should we test?

- Credit card payments can be processed in the backend (orders -> select order -> payments -> add new payment).
- Manual payment by other methods is not affected.
- Checkout flow in the frontend is not affected.
- Cards used do not appear in the admin user's list of saved cards (in /account). Note that the card tokens are still saved in the database even though the `save_requested_by_customer` option is not passed as a param - I can look into disabling this if it's a problem?

#### Release notes

Stripe can now be used in the admin backend to manually add credit card payments to an order.

#### Discourse thread

https://community.openfoodnetwork.org/t/integrating-stripe-into-ofn/664


daniellemoorhead:
@oeoeaio can you please review this once you've got subs sorted. It's been hanging around for an age and should be pushed through if it's given the nod by you.

daniellemoorhead:
@oeoeaio time for this to be pushed through?

myriamboure:
It seems this branch has conflicts. @oeoeaio can you review in the coming days? I don't want to ask @stveep to rebase if he has to redo it again if we don't review and test quickly enough ;-) Would be great to finish it, it's been here for a long time already...

daniellemoorhead:
>@oeoeaio can you review in the coming days?

Sorry @myriamboure, all the Stripe PRs are on hold until Rob has finished Subs. Please don't close this, Rob will definitely rebase and get it sorted once he's got a toggled off Subs into a release :)

stveep:
I can rebase this if you want to get on with testing the functionality @myriamboure ? It still needs rob or someone to look at it eventually though as there is some significant refactoring behind the scenes

myriamboure:
@stveep I suggest we wait for Rob to finish subs and get back on this as it seems you had some questions also he needs to look at, so better to do all the thing at once :-)

app/assets/javascripts/admin/all.js
```diff
@@ -19,6 +19,7 @@
 //= require admin/spree_paypal_express
 //= require ../shared/ng-infinite-scroll.min.js
 //= require ../shared/ng-tags-input.min.js
+//= require ../shared/ofn-stripe.js
```
mkllnk: I can't see this file in this commit.

app/assets/javascripts/admin/payments/services/payments.js.coffee
```diff
@@ -0,0 +1,11 @@
+# Darkswarm.factory 'Payments', ()->
```
mkllnk: This looks like a temporary commit. @stveep Can you rebase your commit history so that each commit makes sense on its own? That makes reviewing a lot easier.

stveep:
I have squashed a couple of commits and made sure new files are added to the commit they are referenced in...hope it makes a bit more sense now.

daniellemoorhead:
Appreciated Steve. The never ending tunnel that is Subs will soon be in
beta testing land (how soon is soon? Hopefully sooner than later ;) ) and
Rob will at last be freed up to get this through.

daniellemoorhead:
@oeoeaio ☝️ 🤣

app/assets/javascripts/admin/payments/controllers/payment.js.coffee
```diff
@@ -0,0 +1,16 @@
+angular.module("admin.payments").controller "PaymentCtrl", ($scope, $timeout, Payment, PaymentMethods, Loading) ->
```
oeoeaio: Doesn't look like $timeout and PaymentMethods are actually required?

app/assets/javascripts/shared/ofn-stripe.js.coffee
```diff
@@ -0,0 +1,84 @@
...
+            displayError.textContent = event.error.message
+          else
+            displayError.textContent = ''
+          return
+
+        StripeElements.stripe = stripe
+        StripeElements.card = card
+
+  .factory 'StripeElements', ($rootScope, Loading, RailsFlashLoader) ->
+    new class StripeElements
```
oeoeaio: I would prefer to see services and directives placed in separate folders as per the rest of the codebase. It becomes very difficult to find things when there is no structure to follow.

app/assets/javascripts/darkswarm/directives/loading.js.coffee
```diff
@@ -1,9 +0,0 @@
-Darkswarm.directive "loading", (Loading)->
-  # Triggers a screen-wide "loading" thing when Ajaxy stuff is happening
```
oeoeaio: I appreciate that a lot of work has gone into making this DRY, but I think sometimes it is possible to take that idea too far. In this case, it feels to me as though some code duplication would have been preferable to bringing the Loading, Navigation and RailsFlashLoader classes into the shared module. Going by my testing, the Loading overlay and the flash are not being used/rendered properly in the admin section anyway, so it feels like a muddying of the waters having this code here. What do you think @stveep?

app/assets/javascripts/shared/ofn-loading.js.coffee
```diff
@@ -0,0 +1,15 @@
...
+    templateUrl: 'loading.html'
+    controller: ($scope)->
+      $scope.Loading = Loading
+      $scope.show = ->
+        $scope.Loading.message?
+  .factory "Loading", ->
+    new class Loading
+      message: null
+      clear: =>
+        @message = null
```
oeoeaio: I think the convention is generally to name the file after the service/directive contained within, so there is no need for the `ofn-` prefix in the filename here.

Also, same deal here with the folder structure, directives in `directives`, factories in `services`. 🙏

app/assets/javascripts/shared/rails-flash-loader.js.coffee
```diff
@@ -0,0 +1,19 @@
+angular.module('RailsFlashLoader', ['angular-flash.service'])
```
oeoeaio: I am all for breaking big modules up into small ones, but having a whole module for single directives/factories seems like a little bit too much granularity. IMO it doesn't improve the hierarchy because we will end up with almost as many folders as files. Module are intended as organisational units whereby the components often common dependencies. This does not mean that ALL components of a module that is inherited by another must be used by the depending module...

daniellemoorhead:
Back over to you @stveep...hopefully you don't have to reach too far back in the memory bank. 

tenor

app/assets/javascripts/darkswarm/directives/loading.js.coffee
stveep: I can duplicate it if you prefer but I don't really see the problem with extracting the Navigation module; this could probably find other uses in the backend too. 

With Loading and FlashLoader, the problem is that these are required in the front end implementation so we still need to satisfy the requirement for them when the Stripe stuff is extracted to be shared with the backend. That was really the only reason for moving these. Do you think there's a better structure for how the dependencies are handled that we could use to avoid having to require these in the backend? As you say, it doesn't really work properly and it would be better to have the notifications/status updates more consistent with the rest of the admin interface - using statusMessage for example - but then we will have the opposite problem in the front end. I could add some logic for this to change how it's handled depending on whether we are in checkout or admin, but would still end up needing to include all the different methods.

oeoeaio: Thanks @stveep 

>this could probably find other uses in the backend too

To me, adding the `Navigation` service to the admin section on the off-chance that we might use it is a case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Also, IMHO, the `Navigation` service is not actually all that useful. It's just a very thin wrapper for the $location service that ships with AngularJS. For example, I wrote the `tabsetCtrl` directive which uses paths and doesn't depends on the `Navigation` service.

My preference would be that we duplicate the relevant parts of the `stripeElements` directive and `StripeElements` service from the frontend to admin, and rewrite them to use the equivalent admin services as required. Then we can keep the `Navigation`, `Loading` and `RailsFlashLoader` services in the frontend, thus removing the need for any shared modules.

I think the only real alternative is to do more work to refactor the Stripe directive and service so that they don't directly depend on any other services (using `$injector` you can check if services are present before using them), so that they can be cleanly used in either context. This still feels pretty messy to me, and will complicate the Stripe logic with lots of conditionals, but we can investigate this more if you'd prefer to go down that path.

PR #1935, opened 2017-10-31
Redefined github source to remove insecure protocol warnings

andypalmer:
#### What? Why?

This redefines the `github` source for rubygems to deal with the insecure protocol warnings
e.g. `The git source 'git://github.com/openfoodfoundation/spree.git' uses the 'git' protocol, which transmits data without encryption. Disable this warning with 'bundle config git.allow_insecure true', or switch to
 the 'https' protocol to keep your data secure.`

#### What should we test?

Test that the `bundle install` works successfully. Github dependencies are updated from `git://` to `https://`
The `warden` gem is upgraded to `1.2.7` via transient dependency from the `devise` gem, so check that authentication is working as expected

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -669,7 +669,7 @@ GEM
       actionpack (>= 3.0)
       activesupport (>= 3.0)
       railties (>= 3.0)
-    warden (1.2.7)
+    warden (1.2.6)
       rack (>= 1.0)
     webmock (1.8.11)
       addressable (>= 2.2.7)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,5 +1,5 @@
 GIT
-  remote: git://github.com/RohanM/simple_form.git
+  remote: https://github.com/RohanM/simple_form.git
   revision: 45f08a213b40f3d4bda5f5398db841137587160a
   specs:
     simple_form (2.0.2)
@@ -7,13 +7,13 @@ GIT
       activemodel (~> 3.0)
 
 GIT
-  remote: git://github.com/jeremydurham/custom-err-msg.git
+  remote: https://github.com/jeremydurham/custom-err-msg.git
   revision: 3a8ec9dddc7a5b0aab7c69a6060596de300c68f4
   specs:
     custom_error_message (1.1.1)
 
 GIT
-  remote: git://github.com/openfoodfoundation/better_spree_paypal_express.git
+  remote: https://github.com/openfoodfoundation/better_spree_paypal_express.git
   revision: 8d95f4544c682634812becaf50999fba0cd04df0
   branch: spree-upgrade-intermediate
   specs:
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -669,7 +669,7 @@ GEM
       actionpack (>= 3.0)
       activesupport (>= 3.0)
       railties (>= 3.0)
-    warden (1.2.6)
+    warden (1.2.7)
```
sauloperez: I wouldn't upgrade this so we can merge this PR without further ado

sauloperez:
Thanks for this PR! this was very needed.

Gemfile.lock
andypalmer: I'm really reluctant to hack Gemfile.lock files manually. On the other hand, release notes for warden 1.2.7 mentioned an upgrade to Ruby 2.3.
I'll make the change, but I'm expressing my overall unhappiness with the situation (both the really old version of Ruby and having to hack the lock file)

oeoeaio: @sauloperez I replicate this when testing on my local machine, and I tend to agree with @andypalmer that hacking the lock file is something we should avoid. A quick look at the [ruby 2.3 PR](https://github.com/hassox/warden/pull/139) that is mentioned in the release notes suggests to me that it is designed to add compatibility for 2.3, rather than deprecate the use of older versions.

Gemfile.lock
```diff
@@ -123,7 +123,7 @@ GIT
       spree_core (>= 1.1)
 
 GIT
-  remote: git://github.com/willrjmarshall/foundation_rails_helper.git
+  remote: https://github.com/willrjmarshall/foundation_rails_helper.git
   revision: 4d5d53fdc4b1fb71e66524d298c5c635de82cfbb
   branch: rails3
   specs:
```
oeoeaio: @sauloperez reposting this just in case you did not see it. I can replicate the upgrade of `warden` to `1.2.7` when testing on my local machine, and I tend to agree with @andypalmer that hacking the lock file is something we should avoid. A quick look at the ruby 2.3 PR that is mentioned in the release notes suggests to me that it is designed to add compatibility for 2.3, rather than deprecate the use of older versions. Are you ok if we include the `warden` upgrade?

sauloperez: Yes, the change is very simple. We'll notice pretty soon if the authentication fails.

oeoeaio: @sauloperez 👍  thanks

Sorry for the stuff around @andypalmer, could I get you to drop that last commit?

enricostano: @andypalmer are you still working on this? Thanks!

andypalmer: @enricostano given the length of time, it's probably better to close this PR and re-raise with the Gemfile change. Otherwise this PR will be 4 commits, of which 3 are noise

andypalmer:
Resubmitted as #1986

PR #1937, opened 2017-10-31
Adjust use of non-translatable open/closed icons

Matt-Yorkley:
#### What? Why?

Changing use of icons that have non-translatable English text for issue #1561. 

#### What should we test?

Icons on /shops page and on shop where order cycles are closed.

#### Screenshots

BEFORE:
screenshot from 2017-10-31 12-14-00

screenshot from 2017-10-31 12-13-20

AFTER:
screenshot from 2017-10-31 12-34-09

screenshot from 2017-10-31 12-54-25

### DIFF
--- a/app/assets/stylesheets/darkswarm/hub_node.css.scss
+++ b/app/assets/stylesheets/darkswarm/hub_node.css.scss
@@ -8,6 +8,14 @@
       white-space: nowrap;
       overflow-x: hidden;
       overflow-y: visible;
+
+      i {
+        margin-right: 0.5rem;
+      }
+    }
+
+    .active_table_row {
+      line-height: 1rem;
     }
 
     //Generic text link style
@@ -30,7 +38,7 @@
     //Closed & Open column
     .open_closed {
       i {
-        font-size: 2rem;
+        font-size: 1.75rem;
         float: right;
         margin-left: 0.5rem;
       }
--- a/app/assets/javascripts/templates/partials/hub_details.html.haml
+++ b/app/assets/javascripts/templates/partials/hub_details.html.haml
@@ -17,8 +17,8 @@
         %a.cta-hub{"ng-href" => "{{::enterprise.path}}",
         "ng-class" => "{primary: enterprise.active, secondary: !enterprise.active}",
         "ofn-change-hub" => "enterprise"}
-          %i.ofn-i_033-open-sign{"ng-if" => "::enterprise.active"}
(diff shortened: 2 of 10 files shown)


lin-d-hop:
Awesome @Matt-Yorkley

sstead:
Testing notes @oeoeaio 
The icons are changed on the directory and in the shopfront, but not on the map view of profiles (viewed from main map, or from group map).
image

oeoeaio:
Over to you @Matt-Yorkley, chuck the `pr-test-ready` label back on when you're good to go.

Matt-Yorkley:
Replaced the map view icons with text, as it made more sense visually. Closed shops are greyed out as well:

screenshot from 2017-11-02 13-10-55
screenshot from 2017-11-02 13-13-05

Matt-Yorkley:
Back to you, @sstead

daniellemoorhead:
@sstead can you get Rob to stage this for you next time you're in so you can test again 😄

sstead:
@oeoeaio one bug spotted. I checked and this isn't happening on production..

image

daniellemoorhead:
Back to you @Matt-Yorkley, see the one bug that Sally found outlined above.

Matt-Yorkley:
@sstead The shops in map view should be working now if you want to take another look.

app/assets/stylesheets/darkswarm/hub_node.css.scss
```diff
@@ -8,6 +8,14 @@
       white-space: nowrap;
       overflow-x: hidden;
       overflow-y: visible;
+
+      i {
+        margin-right: 0.5rem;
+      }
+    }
+
+    .active_table_row {
```
mkllnk: I was wondering if this doesn't allow the rows to grow in case they need two lines of text, but the current design doesn't allow for that either, it just cuts the shop name. So this isn't doing any harm, I think. This may be something to address somewhere else.

Before:
screenshot from 2018-01-24 15-44-04

After (different data):
screenshot from 2018-01-24 15-59-41

sstead:
**Testing notes:** @mkllnk 
- The open/close icons are correct
- The issue I identified last time is fixed (where open/close status was incorrect). 
- The only small note is that the appearance of the 'shop now' links are different on the Producer listing page, compared to when the profile is viewed in the map, or from the shopfront. (see image below showing difference)... I'm not sure if this is reason enough not to push.
image

mkllnk:
Okay, I pushed this. It would be nice to fix the two outstanding issues in another pull request. What do you think @Matt-Yorkley?

PR #1939, opened 2017-11-02
Bootstrap angular module in new_variant form

### LINKED ISSUE #1932: New variant form persists wrong weight
## What

When creating a new variant it is not possible to set a fraction of the product's weight. Neither `0.5`, nor `0,5` nor `500` show the right value in the variants list. See gif below:

variant_bugs

When editing however, typing `0.5` does show up as `500g` in the variants list.

First, we need to fix the weird label copy and then, fix the input field so that it works like the edit form.

## Acceptance criteria

As an admin, when I navigate to the new variant form, I see `PESO (kg)` rather than `PESO ({{UNITNAME(1000.0, 'WEIGHT')}})`.

Also, when I create a new variant with weight `0.5` it shows as up as `500g` in the variants list.

mllocs:
#### What? Why?

Fixes #1932 

In the `new variant` form page, the form is dynamically loaded without rerunning the angular module (re-bootstrapping). The `edit variant` form works because the full page is loaded and so, the angular module is also executed. 

#### What should we test?

Adding, updating or removing variants. 

#### Release notes

Fixed new variant form, not capturing the weight value.

### DIFF
--- a/app/overrides/spree/admin/variants/_form/add_unit_value_and_description.html.haml.deface
+++ b/app/overrides/spree/admin/variants/_form/add_unit_value_and_description.html.haml.deface
@@ -11,3 +11,8 @@
     .field{"data-hook" => "unit_description"}
       = f.label :unit_description, t(:spree_admin_unit_description)
       = f.text_field :unit_description, class: "fullwidth", placeholder: t('admin.products.unit_name_placeholder')
+
+    :javascript
+      angular.element(document.getElementById("new_variant")).ready(function() {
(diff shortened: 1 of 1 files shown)


sauloperez:
Awesome!! this will have very positive impact on our users! :muscle:

Matt-Yorkley:
I restarted that failed Travis spec and it passed. All green :)

sauloperez:
Ups, I didn't notice this was already 🍏 :trollface:

daniellemoorhead:
@sauloperez travis is broken for this PR :(

Did you want to stage and test this, and add test your test results?

PR #1940, opened 2017-11-02
Improve readability of shop's code

sauloperez:
#### What? Why?

While investigating how a shop is rendered I found a number of things that ought to be cleaned up and be a bit more readable.

Also, you will see I removed a Bugsnag notification. In the related commit message you will find the reason why.

### DIFF
--- a/app/helpers/injection_helper.rb
+++ b/app/helpers/injection_helper.rb
@@ -87,10 +87,12 @@ module InjectionHelper
 
   def inject_json_ams(name, data, serializer, opts = {})
     if data.is_a?(Array)
-      json = ActiveModel::ArraySerializer.new(data, {each_serializer: serializer}.merge(opts)).to_json
-    else
-      json = serializer.new(data, opts).to_json
+      opts = { each_serializer: serializer }.merge(opts)
+      serializer = ActiveModel::ArraySerializer
     end
+
+    serializer_instance = serializer.new(data, opts)
+    json = serializer_instance.to_json
     render partial: "json/injection_ams", locals: {name: name, json: json}
   end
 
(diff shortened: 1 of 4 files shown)


sauloperez:
No need for testing here. What label should I add then?

oeoeaio:
Thanks @sauloperez. Seeing as there has been a change to a functional part of the code I would like to test this if that is ok?

I tend to agree with your assertion that error tracking should not be used for monitoring cache misses, but I am reluctant to just burn our existing approach until we have a replacement solution in place.

sauloperez:
> Seeing as there has been a change to a functional part of the code I would like to test this if that is ok?

As you wish, I just see that the change is fully covered by lots of tests.

> but I am reluctant to just burn our existing approach until we have a replacement solution in place.

We need to understand whether you guys actually use that information and how

sauloperez:
I opened an issue to address the cache monitoring https://github.com/openfoodfoundation/openfoodnetwork/issues/1949 so that we reach a consensus here.

daniellemoorhead:
>We need to understand whether you guys actually use that information and how

We get notifications from Bugsnag, and there are also GitHub issues created for some of them (not sure what triggers that because it's not for every event, @mkllnk or @oeoeaio would know). Each time we get a message we are sad because we don't have the time to focus on solving them.

sauloperez:
@oeoeaio do you still want to test this? IMO being the injection helper we would have noticed in the integration tests if it was broken, but otherwise we can take a look ourselves.

sauloperez:
Tested and working. The refactored method is used in https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/helpers/injection_helper.rb#L68 which renders the orders successfully as you see below.

injector

daniellemoorhead:
@oeoeaio this has been tested so ready to merge if you're happy to approve code review.

sauloperez:
cc @oeoeaio

PR #1942, opened 2017-11-03
Add spree user as admin of Enterprise 2 in seed

sauloperez:
#### What? Why?

Just what the subject says. It makes it easier to start using OFN in development.


mllocs:
Yay! 👍

PR #1952, opened 2017-11-09
Remove upgrade_bundler script

oeoeaio:
#### What? Why?

This script was used by travis to ensure that we were using a recent version of bundler. A recent update of `rubygems` (2.7.1-ish) seems to have simultaneously broken the script on travis, and made it redundant. Recent version of bundler now seems to be installed by `rubygems`. 

#### What should we test?

That the test suite runs on travis.

#### Release notes

None. Bug fix relevant to active devs only.


lin-d-hop:
Thanks @oeoeaio!!

Issue #1953, opened 2017-11-13
Embedded Shopfronts - All allowed URLs in header


lin-d-hop:
There is a strange error happening in embedded shopfronts in which there appears to be some kind of hierarchical dependency within the list of Allowed URLs recorded in super-admin.

We have the following URLs on our list:
`dev.landworkersalliance.org.uk
www.foodmadegood.org
www.blackmountainfoodhub.co.uk
www.rootsfruitsandleaves.co.uk`

When I add a new shopfront to the list I get the following error in Chrome console when I load the page with the embedded shop. In this case I loaded Black Mountain Food Hub. But the embedded frame is failing apparently due to errors in Food Made Good (different group site).
![lowerlevelembed](https://user-images.githubusercontent.com/8669848/32722107-b7a41dce-c860-11e7-9bfd-531efba09978.png)
Regardless of whether of not there are errors in that site, surely loading www.blackmountainfoodhub.org.uk should know absolutely nothing about www.foodmadegood.org

Note the order of the URLs is important. The top URL in the list will load the embedded shopfront but will display the following errors in Chrome console on page load
![toplevelembed](https://user-images.githubusercontent.com/8669848/32722368-9b4f7820-c861-11e7-8682-4ffd298e468c.png)

This issue is very high priority for UK as it is currently causing shopfronts to fail to load. In particular, playing around trying to understand this issue currently our established embedded shopfront is not loading for the user = big problem for them.

Unfortunately everyone that can help with this issue is on holiday! x-/

lin-d-hop:
Ok, so I spent a while looking into this.
The issue was caused due to very-easy-to-repeat user error by me. But I do think there is grounds to change something here....

So when I entered the Embeddable URLs into production I did so by entering each URL on a new line.
This became, in the page headers something like:
frame-ancestors = www.url1.com /rwww.url2.com /rwww.url3.com

This resulted in the first url being recognisable but all future URLs being deemed as illegal.

So the solution to actually have pages embed was to ensure that a **space** exists so that page headers actually contain:
frame-ancestors = www.url1.com www.url2.com www.url3.com

This did get me wondering, however, if it is a good idea to embed the URL of every that we have enabled shopfronts for in every header? Maybe I'm being over zealous, but it felt kinda intrusive to be able to see from one website the list of all other websites using OFN embeddable shopfronts.

[...]

PR #1954, opened 2017-11-13
Limit frame-ancestors to requesting URI for embedded shops

lin-d-hop:
#### What? Why?

#1953 
Proposing an alternative to listing every allowed parent in every header. When debugging iframe issues it felt inappropriate to be able to see the list of all URIs that can embed OFN shopfronts from a clients website.

#### What should we test?

Not ready for testing. WIP

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -57,10 +57,10 @@ class ApplicationController < ActionController::Base
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
     return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present?
-    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
+    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development? && !whitelist.include?(URI(request.referer).host)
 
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -57,10 +57,10 @@ def restrict_iframes
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
     return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present?
-    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
+    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development? && !whitelist.include?(URI(request.referer).host)
```
sauloperez: Did we consider extracting this logic from the application controller? This line is getting hairy over time (the last two commits touch it) and thus becoming harder to change. Besides, the embedded stuff adds an extra responsibility the controller. We should extract it from here as per the single responsibility principle.

daniellemoorhead:
@lin-d-hop Maikel is on holiday until early December. Did you want to wait for him to get back to review this or plan B is Rob looking at it. Your choice 🙂

lin-d-hop:
@daniellemoorhead not in a hurry. I'm happy to follow up with @matt-yorkley when he's back

lin-d-hop:
Replaced by #2026

PR #1960, opened 2017-11-15
Fix payment_method_decorator boot-time warning

sauloperez:
#### What? Why?

Removes the annoying message `warning: already initialized constant Spree::PaymentMethod::DISPLAY` that appears 4 times when booting the app.

We are declaring said constant exactly as our Spree version does so  there's no point on repeating work.

Now when booting the app you will only see:

```shell
$ b rails s
=> Booting Unicorn
=> Rails 3.2.21 application starting in development on http://0.0.0.0:3000
=> Call with -d to detach
=> Ctrl-C to shutdown server
(eval):1: warning: encountered \r in middle of line, treated as a mere space
listening on addr=0.0.0.0:3000 fd=15
worker=0 spawning...
master process ready
worker=0 spawned pid=9169
worker=0 ready
```

#### What should we test?

Enough with all tests passing

### DIFF
--- a/app/models/spree/payment_method_decorator.rb
+++ b/app/models/spree/payment_method_decorator.rb
@@ -1,8 +1,6 @@
 Spree::PaymentMethod.class_eval do
   include Spree::Core::CalculatedAdjustments
 
-  Spree::PaymentMethod::DISPLAY = [:both, :front_end, :back_end]
-
   acts_as_taggable
 
   has_and_belongs_to_many :distributors, join_table: 'distributors_payment_methods', :class_name => 'Enterprise', association_foreign_key: 'distributor_id'


daniellemoorhead:
@oeoeaio should be a quick one to review and merge

oeoeaio:
Sorry @sauloperez, I'd like to give this one a really quick test if that is ok. Just want to be 100% sure that it's not going to have a weird affect.

Note to self: 7d79fffa33426dfba1f7775a3abaf968333f5bef

app/models/spree/payment_method_decorator.rb
```diff
@@ -1,8 +1,6 @@
 Spree::PaymentMethod.class_eval do
   include Spree::Core::CalculatedAdjustments
 
-  Spree::PaymentMethod::DISPLAY = [:both, :front_end, :back_end]
```
sauloperez: In the latest version of 2.0 this is still defined in Spree. See by yourself: https://github.com/openfoodfoundation/spree/blob/2-0-stable/core/app/models/spree/payment_method.rb#L4

sauloperez:
tested in our staging and working:

payment_method

otherwise, when visiting this page the app would crash. I believe we're good to merge. cc @oeoeaio @mkllnk

sauloperez:
Can we merge guys? take a look at my test notes above @mkllnk @Matt-Yorkley

sauloperez:
would you agree on merging @mkllnk ? honestly, testers are a scarce resource to be spent on this.

mkllnk:
Rob will have a quick look at it.

daniellemoorhead:
@enricostano @sauloperez please go ahead and merge it. There have been enough eyeballs on this PR, let's just get it done.

giphy

sauloperez:
Lol

PR #1975, opened 2017-11-22
Fix before_filter for Spree::Admin::OrdersController#print_ticket

ltrls:
The load_order in Spree::Admin::OrdersController seems to be mandatory since recent Spree upgrades. If it doesn't include the #print_ticket method, it prevents the `ticket.html` template from accessing the relevant order.
I'll write feature tests on the ticket thing as soon as possible to prevent this from happening again!

### DIFF
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -4,7 +4,7 @@ Spree::Admin::OrdersController.class_eval do
   include OpenFoodNetwork::SpreeApiKeyLoader
   helper CheckoutHelper
   before_filter :load_spree_api_key, :only => :bulk_management
-  before_filter :load_order, only: %i[show edit update fire resend invoice print]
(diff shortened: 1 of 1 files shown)


myriamboure:
@oeoeaio @daniellemoorhead we can't deploy any new version since spree upgrade until we merge that, so great if you can merge, it only impacts France as we are the only one using the print receipts option. Pierre will write the tests but this will not be done before the relase so thanks for merging :-)

oeoeaio:
@myriamboure @ltrls I've merged this but it would be great to create an issue for adding a request/feature spec to cover this scenario so that we don't forget.

myriamboure:
Thanks @oeoeaio I created the issue and plan that in French backlog

sauloperez:
Just as a side comment, we should follow the best practice of adding first a test case to reproduce the bug and then implement the fix. I guess we're all on the same page on this. IMHO splitting up tests from implementations into different PRs can have very bad consequences.

Issue #1756, opened 2017-08-04
Navigation around Enterprise Settings and Shipping/Payment methods


sstead:
Navigation around the Enterprise Settings is not always intuitive. This is one of the first places a user goes, so we should make it less confusing.

Use case: If I'm in Edit Enterprise > Contact details and I make some changes and click update, the page reloads, but I'm back on Primary Details. I would expect to be taken to the contact details page refreshed.

Use case: I'm setting up shipping and payment methods. I navigate to the New shipping method page, via my enterprise settings. When I'm done I'm redirected to a list page of my shipping methods, not back to the enterprise edit page (on the shipping method tab) as I'd expect. Same for payment methods. This is confusing, and I'm not sure where these three screens are or how they interconnect.

![image](https://user-images.githubusercontent.com/10843056/28954875-8eb7dd40-7925-11e7-8a6a-dafee090c66c.png)

adamgunther1:
Hi there! First-time contributors and we're hoping to help out with this issue as an assignment for school. I noticed nobody was assigned to it, but if there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

NickWeir63:
Hi @adamgunther1 welcome to OFN !  If you let me have your email address I will join you up to our Slack channel to help with communication.  My email is nick@openfoodnetwork.org.uk  This issue is not ladbelled as 'Rubies for Newbies' but I will ask @lin-d-hop to let you know if it is suitable for a first-time contribution.  

Thanks very much for getting involved.  Hope to meet you soon.

adamgunther1:
Thank you, please let me know. Three of us are working on the issue and do have solid experience developing full stack rails apps and have been working with Ruby for close to a year.

sauloperez:
Welcome @adamgunther1! Always happy to see new contributors! Your experience will come in handy.

adamgunther1:
Hi, due to the timeframe for our assignment, this issue will require a lot of back and forth on UI/UX decisions, so we are going to offer to contribute on another issue. Thanks for the warm welcome!

NickWeir63:
Thanks @adamgunther1 you might like to look at the issues labelled 'rubies for newbies' unless anyone @daniellemoorhead ? can suggest a quick issue for you to look at.  Thanks again

audez:
Can i create a specific issue "[Admin] Don't jump to primary details after saving a change in Enterprise settings"?
[...]

PR #1979, opened 2017-11-27
#1517 - Payment and shipping methods checkbox issue

### LINKED ISSUE #1517: Deselecting shipping or payment methods in Manage enterprise list doesn't work
I'm Managing my shop enterprise that has shipping and payment methods already setup. I unclick the tickbox next to a shipping/payment method in the 'enable' column to turn 'off' the shipping method. I click save. I get the banner saying that my enterprise has been updated, but in Shipping/Payment methods I still see the method I just deactivated is still active. If I want to deactivate it I need to 'edit shipping/payment method' and remove the enterprise there- which is difficult because if the method is only applied to one enterprise, you can't save the method with no enterprise selected. Thus you'd need to delete the method.

image

The opposite works fine - If I'm selecting to enable a shipping/payment method that is currently deactivated.

yasirazgar:
#### What? Why?
This issue occurs when all the selections are unchecked in payment/shipping methods in manage enterprises.

Closes #1517 

When all the checkboxes are unchecked, the form ignores the whole attribute, so all the unchecked payments/shipments will not get updated.
To fix this, added a hidden_tag with the attribute name, with value as empty array, which updates shipping/payment methods.

#### What should we test?

Need to test whether all the shipping/payment methods in manage enterprises is getting updated correctly on check/uncheck especially unchecking all the checkboxes.
http://localhost:3001/admin/enterprises/{id}/edit#/shipping_methods
http://localhost:3001/admin/enterprises/{id}/edit#/payment_methods

#### Dependencies
 https://github.com/openfoodfoundation/openfoodnetwork/pull/2039 - Shipping/payment methods disappears for manager on uncheck

### DIFF
--- a/app/views/admin/enterprises/form/_payment_methods.html.haml
+++ b/app/views/admin/enterprises/form/_payment_methods.html.haml
@@ -9,6 +9,8 @@
         %th= t('.applies')
         %th
     %tbody
+      %tr 
+        = f.hidden_field :payment_method_ids, value: nil, multiple: true
       - @payment_methods.each do |payment_method|
         %tr{ ng: { controller: 'paymentMethodsCtrl', init: "findPaymentMethodByID(#{payment_method.id})" } }
           %td= payment_method.name
--- a/app/views/admin/enterprises/form/_shipping_methods.html.haml
+++ b/app/views/admin/enterprises/form/_shipping_methods.html.haml
@@ -6,6 +6,8 @@
         %th= t('.applies')
         %th
     %tbody
+      %tr 
+        = f.hidden_field :shipping_method_ids, value: nil, multiple: true
       - @shipping_methods.each do |shipping_method|
         %tr{ ng: { controller: 'shippingMethodsCtrl', init: "findShippingMethodByID(#{shipping_method.id})" } }
           %td= shipping_method.name


daniellemoorhead:
Hey @oeoeaio, @sauloperez, @pacodelaluna, @ltrls, could one of you have a look at this PR and provide a review. Thanks 😄

ltrls:
Code looks fine

daniellemoorhead:
Hey @mkllnk can you please stage this for @sstead to test. Thanks!

sstead:
@mkllnk some enterprises don't have shipping/payment methods visible in their Enterprise Settings menu...
image

mkllnk:
I can reproduce the not-loading of the new order cycle form. In the network console I can see that the page fails to load `https://staging1.openfood.com.au/admin/schedules.json`, status 404. I don't know where that call comes from. Maybe the assets didn't get updated properly? @oeoeaio 

I don't think that this is an issue with this pull request. I think it is an existing issue on the server or in the production code.

mkllnk:
Okay, I rebuilt all assets and that solved the problem. So somehow our assets don't get rebuild properly. This is unrelated to this pull request, but a problem of our staging server.

sstead:
Here's the behaviour of shipping methods...
Essentially, when a user deselects 'applied' the shipping method is removed entirely from their view (appears deleted). This is the case UNLESS, there are multiple enterprises using the shipping methods (this is the vast minority of shipping methods).
image

sstead:
The behaviour of payment methods is more complex...
Payment methods have 2 levels of control
1) There's the 'applied?' tickbox in the Enterprise settings. If the method is 'applied' in an enterprise, it will show in the shop, unless it's not active (see below).
2) There's the 'active?' tickbox in Edit Payment Method. If the method is not active it will not show in a shop, even if it's 'applied'.

The screenshot below details behavior of the 'applied' tickbox:
image

The screenshot below details the behaviour of the 'active' tickbox:
image

Another question...
The only place to delete a shipping/payment method is in the 'manage shipping/payment method' list. Should you be able to delete from the main tab in enterprise settings? Or from the edit shipping/payment method page?

Another related question... 
Do we need the 'manage shipping/payment methods' page. This gives the user 3 places to interact with their shipping/payment methods. I'd suggest they only need the two places. See issue https://github.com/openfoodfoundation/openfoodnetwork/issues/1756

sstead:
My suggestion
1) A shipping/payment method is always applied to an enterprise. You can't save a method if it is not applied to an enterprise. A method can be applied to multiple enterprises. 
2) You can turn a method on/off (applies/not applies) on an enterprise by enterprise basis- If on it will show in the shop. If off it will show in the enterprise's settings, but not in the shop.
3) You can delete a payment method from all interfaces (currently you can only delete from the 'manage s/p method' interface). 
a) If a method is applied to multiple enterprises- If one enterprise tries to delete it, it will only be deleted from their enterprise (they'll be taken of the list of enterprises using the method).
b) If just one enterprise is using the method and they delete it, it will be removed entirely from the system.

mkllnk:
Thank you @sstead, great catch. This needs to go back to development. Is it possible to run into this bug in current master? In that case, we should open a new issue for it. This pull request should then wait until this bug is fixed.

sstead:
@maikel in production this complication shows in a different way. In production you can't 'un-apply' a shipping method if you're the only user with it (this is what the inital bug was about https://github.com/openfoodfoundation/openfoodnetwork/issues/1517). I suspect this is because the system is trying to avoid s/p methods ending up with no enterprise assigned to them- so if you try to disable it just doesn't save it.

mkllnk:
Oh, I thought you can check another one so that it saves. But okay, at least it means that people are "protected" from this error and have to ask us, before they loose their method. ;-)

myriamboure:
@sstead I'm sorry but I don't understand in your testing notes... after @mkllnk rebuilt the asset, do you still have the problem of shops not seeing their shipping/payment method menus ? And the OC issue ?
Does this PR solve the original issue in an acceptable way to start with, or not @mkllnk @sstead ? I don't see if you tested the "uncheck all methods" thing... 
- If not, to solve it, do we need to create an epic and run a proper inception (taking into account your suggestions) and open various stories to overhaul the behavior as a whole ? In that case we will wait until the issue is prioritized in the Delivery Train Backlog to do that.
- If yes, great, let's merge and maybe @sstead you can open a whishlist item (with link to this PR where all your explanations are) for improving the behavior in a later stage.
That would be great to be able to give a feedback soon to @yasirazgar :-)

mkllnk:
@myriamboure I resolved the issue with the missing menu items. But Sally discovered another bug. If you actually uncheck a method now, it doesn't just deactivate the method, but it completely disappears unless other enterprises are using the same method. Only super-admin can bring that method back. I think we need to address that issue first.

myriamboure:
Ok so I created this other issue https://github.com/openfoodfoundation/openfoodnetwork/issues/2023 to cover what you say @mkllnk and added a dependency on this one, so basically we have to fix the other one before we can merge this PR. Correct me if I'm wrong ;-)

So @yasirazgar it seems that when testing your PR we discovered this other bug (#2023) that we need to fix first, or the user experience might be really bad if we merge yours before that. If you want to work on it of course we would be really grateful for that ! Feel free to ask any question :-)

yasirazgar:
Yes @myriamboure, I would like to work on it. Will look into it and get back to you, if have any questions.

myriamboure:
I'm closing this for now as the other one blocking this one is pretty tricky and will need a whole project/new feature. Kept track of this original issue and this connected PR in the icebox item: https://community.openfoodnetwork.org/t/hubs-manages-without-friction-shipping-and-payment-methods-and-can-grant-permissions-on-them/1269

Issue #1971, opened 2017-11-20
Improve invalid number format treatment and UX on inventory page


myriamboure:
**Problem description:**
On inventory page, if I fill in a number that has not a number format (ex: 2,50,50 or 2.50,2 or 2.50. or...) and try saving, the amount is saved with the invalid format and I get no error message. So if I don't pay attention, I don't realize that the amount I tried to enter is false.

![testing notes 1](https://user-images.githubusercontent.com/8135376/33020315-2efcf68c-cdfe-11e7-8f84-9637c715f3c5.jpg)

**Acceptance criteria:**
When I fill in a number that has not a number format and try saving:
- the system prevent saving
- the wrong amount I filled in is still there and appears in red 
- I see an error message "xxx. has an invalid format. Please enter a number."

sauloperez:
See my comment https://github.com/openfoodfoundation/openfoodnetwork/issues/1970#issuecomment-348405112 which is also related to this.

sauloperez:
We need to check first whether the backend returns already a validation error for that particular item so we can display it. If that's the case, it should be easy.

sauloperez:
@kirstenalarsen mention that might related to EU number format. UK and AU haven't heard of this issue.

kirstenalarsen:
have confirmed that this also happens in Aus. If I put a comma in the number or an extra full stop, it just doesn't save and doesn't give me any warning / notification of the problem

Issue #1970, opened 2017-11-20
Improve unvalid number format treatment and UX on product bulk edit page


myriamboure:
**Problem description:**
On product edit page, if I fill in a number that has not a number format (ex: 2,50,50 or 2.50,2 or 2.50. or...) and try saving, the amount is replaced by "0" and I get no error message. So if I don't pay attention, I don't realize that the amount I tried to enter was not saved properly and replaced by 0.

![testing notes](https://user-images.githubusercontent.com/8135376/33020250-d8e3df4a-cdfd-11e7-85e9-87733b584cbd.jpg)

**Acceptance criteria:**
When I fill in a number that has not a number format and try saving:
- the system prevent saving
- the wrong amount I filled in is still there and appears in red 
- I see an error message "xxx. has an invalid format. Please enter a number."

sauloperez:
If we want this to happen before saving, it's a matter of frontend work. What I'd like to check though is if this behavior is consistent throughout the app. If I'm not mistaken in other parts of the app, the new product page for instance, the error message shows up after saving, meaning that the validation happens on the backend.

Erioldoesdesign:
We created a general inline error validation design task here: https://www.notion.so/openfoodnetwork/How-might-we-improve-form-data-entry-from-users-by-improving-how-we-handle-errors-both-inline-on-UI--83738b75b41c44ada4c5e8f43cadf686

Generally this needs work and there was some discussion that happened on one of the stripe payment tickets about more inline error validation.

PR #1983, opened 2017-12-01
Refactor of LocalizedNumber

oeoeaio:
#### What? Why?

I did a bit of a refactor of the LocalizedNumber module created by @ltrls to allow more flexibility in the use of decimal separators for numbers entered by users (#1831).

There are a few outstanding issues from #1831 that still need to be addressed, and so they can be added here as well.

#### What should we test?

Functionality should not change, this is just a bit of a tidy up of the code.

#### Release notes

No release notes necessary.

### DIFF
--- a/lib/spree/localized_number.rb
+++ b/lib/spree/localized_number.rb
@@ -1,32 +1,24 @@
 module Spree
-  module LocalizedNumber
-    # This method overwrites the attribute setters of a model
-    # to make them use the LocalizedNumber parsing method.
+  class LocalizedNumber
+    # This class overwrites the attribute setters of a model
+    # to make them use the LocalizedNumber#parse method.
     # It works with ActiveRecord "normal" attributes
     # and Preference attributes.
     # It also adds a validation on the input format.
-    # It accepts as arguments a variable number of attribute as symbols
-    def localize_number(*attributes)
+
+    attr_reader :model_class
+    delegate :validate, :instance_method, :define_method, :table_exists?, :column_names, to: :model_class
+
+    def initialize(model_class, attributes)
+      @model_class = model_class
+      @attributes = attributes
+    end
+
+    def setup
       validate :validate_localizable_number
 
-      attributes.each do |attribute|
-        setter = "#{attribute}="
-        old_setter = instance_method(setter) if table_exists? && !column_names.include?(attribute.to_s)
-
-        define_method(setter) do |number|
-          if Spree::Config.enable_localized_number? && Spree::LocalizedNumber.valid_localizable_number?(number)
-            number = Spree::LocalizedNumber.parse(number)
-          elsif Spree::Config.enable_localized_number?
-            @invalid_localized_number ||= []
-            @invalid_localized_number << attribute
-            number = nil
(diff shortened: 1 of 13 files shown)


lib/spree/localized_number.rb
```diff
@@ -37,6 +29,31 @@ def localize_number(*attributes)
...
+          self[attribute] = validated_number
+        else
+          old_setter.bind(self).call(validated_number)
+        end
+      end
+    end
+
+    def self.validated_number(number)
+      return parse(number) if valid_localizable_number?(number)
+      nil
```
sauloperez: I believe this nil would be implicitly returned by Ruby

lib/spree/localized_number.rb
```diff
@@ -37,6 +29,31 @@ def localize_number(*attributes)
       end
     end
 
+    def override_setter_for(attribute)
+      setter = "#{attribute}="
+      old_setter = instance_method(setter) if table_exists? && !column_names.include?(attribute.to_s)
```
sauloperez: I fail to see what's this case. An AR attribute not baked by a database table column?

lib/spree/localized_number.rb
oeoeaio: 👍  that's true

lib/spree/localized_number.rb
oeoeaio: I think it is for preferences? Spree's calculators all share the same table, but they have preferences which are stored in a separate table, using the old setter means that we still set the preference properly. @ltrls should be able to confirm...

daniellemoorhead:
@oeoeaio is this something that is still on your radar? Have moved it to In Dev as it's not ready for review again, please close if it's no longer relevant.

#githubcull2018

daniellemoorhead:
Hey @oeoeaio what's the plan of attack with this PR? Can it be handed over to someone else to complete, or closed off as no longer needed given #853 is now closed?

myriamboure:
Note: the two remaining issues that we opened after we merge the fix for decimals are:
- https://github.com/openfoodfoundation/openfoodnetwork/issues/1970
- https://github.com/openfoodfoundation/openfoodnetwork/issues/1971
It's more UX issues but introduced by this decimal work so I guess that's what you talk about @oeoeaio 
What do we want to do with that ? Shall we open an icebox and put that there, as now users are not so much annoyed by errors when they put , instead of . ;-) So I think it can wait. Opinion welcome @sauloperez

daniellemoorhead:
This isn't being worked on at all right now so I'm pushing this back to the backlog. @oeoeaio feel free to pick it up once you're free from subs stuff.

daniellemoorhead:
@oeoeaio I've unassigned you from this PR as I assume you won't be doing it anytime soon and someone else can pick it up to finish it.

Feel free to reassign yourself if you want to finish it off.

daniellemoorhead:
Actually, I'm going to close it for now.

@mkllnk @Matt-Yorkley @luisramos0 @sauloperez if you believe this should be completed and merged then feel free to reopen.

PR #1984, opened 2017-12-01
Disable api auth as there is no Spree api key set

### LINKED ISSUE #1900: Taxons not showing in Edit Product
Taxons aren't loading in Product Edit. I believe that by default the product category should show here, and you should be able to select a series of product categories in this field e.g. below
image

image

### LINKED ISSUE #1965: Taxonomies / Product Categories broken
I discovered this in Aus production the other day but hadn't got to creating issue yet - we can no longer edit or add product categories (was trying to add a new one for 'fibre')

image

This is not a major problem for existing instances, but it is a MAJOR problem for new instances as without product categories I suspect we can't create products e.g. USA

image

I have a suspicion that this could be Spree Upgrade related, as we have no idea how long since it broke but was obviously (apparently?) ok when India was set up?
@sauloperez @oeoeaio @enricostano

sauloperez:
#### What? Why?

Closes #1900 
Closes #1965

What we'll do from now on is not to authentication using the api key, we'll check the user session instead.

This amends the change done in a87c89c83db9e54c58ea43e5b273d2906669c624, which introduced the bug. As there is no Spree api key set the auth fails when getting taxons. Spree's issue https://github.com/spree/spree/pull/4446 and its related issues contain the background. 

What we found out so far: They added api auth by default in Oct 2012 but they realized that it wasn't working (they weren't sending the api key to the backend), so they turned the api auth false by default and so bypassing the auth in Nov 2012. 

It's later in January 2014 that they enable it back, experiencing the bug later solved in March 2014 in https://github.com/spree/spree/pull/4446.

#### What should we test?

We should be able to create taxons as well as fetch them. This can be done by searching them within the edit product page.

#### Release notes

Fixes taxons not being shown on the edit product page.

This was introduced in https://github.com/openfoodfoundation/openfoodnetwork/commit/a87c89c83db9e54c58ea43e5b273d2906669c624

### DIFF
--- a/config/initializers/spree.rb
+++ b/config/initializers/spree.rb
@@ -12,7 +12,7 @@ require 'spree/core/calculated_adjustments_decorator'
 require "#{Rails.root}/app/models/spree/payment_method_decorator"
 require "#{Rails.root}/app/models/spree/gateway_decorator"
 
-Spree::Api::Config[:requires_authentication] = true
+Spree::Api::Config[:requires_authentication] = false
 
 Spree.config do |config|
   config.shipping_instructions = true
--- a/config/initializers/spree.rb
+++ b/config/initializers/spree.rb
@@ -12,8 +12,6 @@ require 'spree/core/calculated_adjustments_decorator'
 require "#{Rails.root}/app/models/spree/payment_method_decorator"
 require "#{Rails.root}/app/models/spree/gateway_decorator"
 
-Spree::Api::Config[:requires_authentication] = false
-
 Spree.config do |config|
   config.shipping_instructions = true
   config.address_requires_state = true
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem 'i18n-js', '~> 3.0.0'
 gem 'nokogiri', '>= 1.6.7.1'
 
 gem 'pg'
-gem 'spree', github: 'openfoodfoundation/spree', branch: 'step-6a', ref: 'b1b33c7ec682e042bc939aac39dfa1f2de446227'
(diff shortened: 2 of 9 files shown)


config/initializers/spree.rb
```diff
@@ -12,8 +12,6 @@
 require "#{Rails.root}/app/models/spree/payment_method_decorator"
 require "#{Rails.root}/app/models/spree/gateway_decorator"
 
-Spree::Api::Config[:requires_authentication] = true
-
```
oeoeaio: @sauloperez can you give a little explanation of what this is doing?

sauloperez: Sure! this is the exact line that was forcing the api_key authentication. Without it we rely on the session auth instead. This is what I implemented in https://github.com/coopdevs/spree/commit/4e658f5f08f1a0e5e315cae9fdc52d08b1708082.

Does that make sense?

sauloperez:
#1965 fixed! :point_down: 

nested_taxons

sauloperez:
And #1900 as well :point_down: 

product_taxonomies

sauloperez:
cc @kirstenalarsen

daniellemoorhead:
@oeoeaio this has been tested by @sauloperez so all this needs is for you to approve the code review and then merge it.

daniellemoorhead:
PS. @sauloperez  #foo #bar 😆

enricostano:
@sauloperez https://github.com/openfoodfoundation/spree/pull/4 got merged, please add a commit here including it and test again `#justincase`

sauloperez:
Done @enricostano still working as shown in the gifs above :ok_hand:

PR #1986, opened 2017-12-04
Change github definition to remove insecure protocol warning.

andypalmer:
#### What? Why?

This change removes a warning about an insecure protocol by redefining the github source to use https rather than git

This is a resubmission of #1935 to avoid a messy commit history


enricostano:
Thanks @andypalmer 👏

PR #1994, opened 2017-12-14
Transifex

mkllnk:
New translations of the last release. :-)

@daniellemoorhead Do you think it's okay to make an exception for this sort of pull request and put it straight into the current milestone? It should be merged before the next release and is unconditional. There is no reason to move it to the next release, but we often forget to do it.


daniellemoorhead:
No worries @mkllnk, do it!

PR #1996, opened 2017-12-19
Create issue template

sauloperez:
#### What? Why?

This the first proposal of issue template, according to the discussions that we had so far in AUS.

It assumes most of the times issues will be reported to keep track of bugs. We'll link to a feature candidate template later on.


myriamboure:
Sounds good to me !

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,41 @@
+<!--- If what you want to file is not a bug, please go to [other_issues] -->
```
enricostano: Should we create at least a placeholder entry in the wiki and put the link here?

PR #1998, opened 2017-12-19
Template I18n changes

### LINKED ISSUE #1991: Some text does not switch languages
https://github.com/openfoodfoundation/openfoodnetwork/pull/1793 introduces a language switcher. This works for most text, but not text in Javascript templates that has been pre-compiled. Examples:

```
> git grep -l '= t[ (]' -- app/assets/javascripts/templates/
app/assets/javascripts/templates/admin/edit_address_dialog.html.haml
app/assets/javascripts/templates/admin/info_dialog.html.haml
app/assets/javascripts/templates/admin/links_dropdown.html.haml
app/assets/javascripts/templates/admin/modals/tag_rule_help.html.haml
app/assets/javascripts/templates/admin/new_customer_dialog.html.haml
app/assets/javascripts/templates/admin/new_tag_rule_dialog.html.haml
app/assets/javascripts/templates/admin/panels/enterprise_package.html.haml
app/assets/javascripts/templates/admin/panels/enterprise_producer.html.haml
app/assets/javascripts/templates/admin/panels/enterprise_status.html.haml
app/assets/javascripts/templates/admin/panels/exchange_distributed_products.html.haml
app/assets/javascripts/templates/admin/panels/exchange_supplied_products.html.haml
app/assets/javascripts/templates/admin/panels/exchange_tags.html.haml
app/assets/javascripts/templates/admin/tag_autocomplete.html.haml
app/assets/javascripts/templates/admin/tag_rules/filter_payment_methods_input.html.haml
app/assets/javascripts/templates/admin/tag_rules/filter_products_input.html.haml
app/assets/javascripts/templates/admin/tag_rules/filter_shipping_methods_input.html.haml
app/assets/javascripts/templates/out_of_stock.html.haml
app/assets/javascripts/templates/single_line_selectors.html.haml

> git grep -l '=t[ (]' -- app/assets/javascripts/templates/
app/assets/javascripts/templates/admin/tag_rules/filter_order_cycles_input.html.haml

```

This files contain a ruby call to the translate function `t`. It's executed during asset compilation. We need to replace these with the Javascript equivalent like this:

```diff
-  %span{'ng-if' => "v.count_on_hand == 0"}
-    = t('js.out_of_stock.now_out_of_stock')
+  %span{'ng-if' => "v.count_on_hand == 0"}
+    {{ t('js.out_of_stock.now_out_of_stock') }}
```

This bug is not visible in development mode, only with pre-compiled assets.

Affected screens found by Sally:

- https://openfoodnetwork.slack.com/files/U0MLU4Z52/F8ABU5KHB/image.png
- https://openfoodnetwork.slack.com/files/U0MLU4Z52/F8A7KNRS6/image.png
- https://openfoodnetwork.slack.com/files/U0MLU4Z52/F8A5EFHBM/image.png

Matt-Yorkley:
#### What? Why?

Closes #1991 and unblocks multilingual UI testing.

JS template string switched to Angular format to address I18n compilation issues.

#### What should we test?

Compiled template strings in Staging. To see a difference, you need to switch languages. You can choose the Spanish language by appending `?locale=es` to the URL in the browser. For example visit: https://staging1.openfood.com.au/admin/order_cycles/110/edit?locale=es

### DIFF
--- a/app/assets/javascripts/admin/dropdown/dropdown.js.coffee
+++ b/app/assets/javascripts/admin/dropdown/dropdown.js.coffee
@@ -1,1 +1,1 @@
-angular.module("admin.dropdown", ['templates'])
+angular.module("admin.dropdown", ['admin.utils'])
--- a/app/assets/javascripts/templates/admin/edit_address_dialog.html.haml
+++ b/app/assets/javascripts/templates/admin/edit_address_dialog.html.haml
@@ -2,7 +2,7 @@
   %h2 {{ addressType === 'bill_address' ?  "#{t('admin.customers.index.edit_bill_address')}" : "#{t('admin.customers.index.edit_ship_address')}" }}
   %form{ name: 'edit_address_form', novalidate: true, ng: { submit: 'updateAddress()'}}
     .row
-      = t('admin.customers.index.required_fileds')
+      {{ 'admin.customers.index.required_fileds' | t }}
       (
       %span.required *
       )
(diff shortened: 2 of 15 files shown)


Matt-Yorkley:
@mkllnk Angular braces are not working in the templates at all. Do we need something like https://github.com/pitr/angular-rails-templates ?

sauloperez:
> Angular braces are not working in the templates at all

Do we know why?

Matt-Yorkley:
The translations just return blank, or the angular doesn't get evaluated. Am I missing something really obvious?

sauloperez:
> The translations just return blank, or the angular doesn't get evaluated

I was wondering whether we know exactly if it's one or the other and why that is happening. I want to make sure we understand what's the actual problem that we're facing before jumping into a new dependency.

I haven't done any effort on trying to understand it myself, that's why I'm asking :trollface:

mkllnk:
We are already using `angular-rails-templates`, Gemfile line 91. So I took some time to understand what is going on here. The issue comes with narrow angular scopes. For example `app/assets/javascripts/templates/admin/panels/exchange_distributed_products.html.haml` cannot access the `t` function, because `app/views/admin/order_cycles/_exchange_form.html.haml` declares the list of local variables, not including the global `t` function. Adding that, solves the problem:
```diff
  %tr.panel-row{ object: "exchange",
    panels: "{products: 'exchange_distributed_products', tags: 'exchange_tags'}",
-    locals: "$index,order_cycle,exchange,supplied_products,setExchangeVariants,incomingExchangeVariantsFor,productSuppliedToOrderCycle,variantSuppliedToOrderCycle",
+    locals: "t,$index,order_cycle,exchange,supplied_products,setExchangeVariants,incomingExchangeVariantsFor,productSuppliedToOrderCycle,variantSuppliedToOrderCycle",
    colspan: 5 }
```

I guess we have to find all places where these templates are used and provide the `t` function as "parameter" (local variable).

Matt-Yorkley:
> I guess we have to find all places where these templates are used and provide the t function as "parameter" (local variable).

Yikes. That seems really messy. Surely there must be a cleaner way to allow template files access to the translation method? I'll do some searching...

Matt-Yorkley:
@mkllnk passing in `t` to templates via `locals` didn't work, but after looking at the Darkswarm (frontend) implementation, I found that single-line translated strings wrapped in angular braces were all using the translate filter instead, i.e. `{{ 'some.string' | t }}`. I switched to this format for the other admin templates and made sure the angular modules had sufficient access to the `t` filter.

mkllnk:
I had a very quick look at our staging server and this seems to do the trick. Sally is going to test a bit more thoroughly this weekend.

Matt-Yorkley:
Great, this should unblock #1793 and get things moving again with the multilingual feature.

PR #2022, opened 2017-12-21
Add feature test for the Spree::Admin::OrdersController print_ticket action

### LINKED ISSUE #1977: Add test for print receipt
As precised in the PR https://github.com/openfoodfoundation/openfoodnetwork/pull/1975 we need to write tests to prevent the issue from happening again.

ltrls:
#### What? Why?

Closes #1977

During a part of the Spree upgrade the ticket printing stopped working, and the Spree upgrade team didn't notice because I hadn't make any test for the feature, so here it is. Apart than testing that rendering the view doesn't cause any exception it checks for elements that are supposed to be in the ticket. As always feedback is more than welcome!

#### What should we test?

Nothing, Travis should be enough!

### DIFF
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -6,6 +6,7 @@ feature %q{
 }, js: true do
   include AuthenticationWorkflow
   include WebHelper
+  include CheckoutHelper
 
   background do
     @user = create(:user)
@@ -175,7 +176,6 @@ feature %q{
     expect(page).to have_current_path spree.admin_orders_path
   end
 
-
   context "as an enterprise manager" do
     let(:coordinator1) { create(:distributor_enterprise) }
     let(:coordinator2) { create(:distributor_enterprise) }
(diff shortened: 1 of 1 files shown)


spec/features/admin/orders_spec.rb
```diff
@@ -196,20 +196,64 @@ def new_order_with_distribution(distributor, order_cycle)
...
-          expect(page).to have_link "Resend Confirmation", href: spree.resend_admin_order_path(order)
-          expect(page).to have_link "Send Invoice", href: spree.invoice_admin_order_path(order)
-          expect(page).to have_link "Print Invoice", href: spree.print_admin_order_path(order)
-          # expect(page).to have_link "Ship Order", href: spree.fire_admin_order_path(order, :e => 'ship')
-          expect(page).to have_link "Cancel Order", href: spree.fire_admin_order_path(order, :e => 'cancel')
+          expect(page).to have_link "Edit", href: spree.edit_admin_order_path(@order)
+          expect(page).to have_link "Resend Confirmation", href: spree.resend_admin_order_path(@order)
+          expect(page).to have_link "Send Invoice", href: spree.invoice_admin_order_path(@order)
+          expect(page).to have_link "Print Invoice", href: spree.print_admin_order_path(@order)
+          # expect(page).to have_link "Ship Order", href: spree.fire_admin_order_path(@order, :e => 'ship')
```
enricostano: Do we need this check?

spec/features/admin/orders_spec.rb
```diff
@@ -196,20 +196,64 @@ def new_order_with_distribution(distributor, order_cycle)
       login_to_admin_as @enterprise_user
     end
 
-    context "viewing the edit page" do
-      it "shows the dropdown menu" do
+    feature "viewing the edit page" do
+      background do
+        Spree::Config[:enable_receipt_printing?] = true
```
enricostano: can we mock this? otherwise config will get cached and can affect other tests, especially in CI where we use parallel tests

spec/features/admin/orders_spec.rb
enricostano: Time will tell us... not blocking for now.

Issue #102, opened 2013-09-22
OC time gets confused if OC shuts while I have items in my cart .. it starts counting back the other way . .

Can we have a message that pops up on any action from the page that says "Sorry orders for this order cycle closed XX minutes / hours ago! Please


kirstenalarsen:
http://www.bugherd.com/projects/21919/tasks/85

PR #2026, opened 2017-12-23
Only include one host in embedded shopfront headers

### LINKED ISSUE #1953: Embedded Shopfronts - All allowed URLs in header

Matt-Yorkley:
#### What? Why?

Closes #1953. 

Adjustments to the headers that are returned with embedded shopfronts.

#### What should we test?

Embedded shopfronts are working and returning a single URL in the 'frame-ancestors' response header. 

#### Testing notes:

Needs to be checked in an external site with an iframe. The PR needs to be staged first, and the correct URL added to the domain whitelist in the admin settings on the staging server, then an iframe on another site needs to be used to check it, pointing to the correct staging URL.

Checklist: 
- Embedded shopfronts needs to be enabled in instance admin config settings
- The domain of the embedding site needs to be in the whitelist
- The embedding site needs to have HTTPS
- The embedding site's iframe should point to: `https://staging-server-url/your-enterprise/shop?embedded_shopfront=true`

In the headers there should only be one domain name present under `frame-ancestors`. As a side-effect of this PR, the whitelist should work whether it has newlines, spaces, commas or whatever else separating the domains in the list.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -56,14 +56,20 @@ class ApplicationController < ActionController::Base
 
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
-    return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present?
+    domain = embedded_shopfront_referer
+    return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present? && domain.present? && whitelist.include?(domain)
     return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
 
     response.headers.delete 'X-Frame-Options'
-    response.headers['Content-Security-Policy'] = "frame-ancestors #{whitelist}"
+    response.headers['Content-Security-Policy'] = "frame-ancestors #{domain}"
 
     check_embedded_request
     set_embedded_layout
+  end
+
+  def embedded_shopfront_referer
+    return if request.referer.blank?
+    URI(request.referer).host.sub!(/^www./, '')
   end
 
   def check_embedded_request
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -67,7 +67,8 @@ class ApplicationController < ActionController::Base
 
   def embedded_shopfront_referer
     return if request.referer.blank?
-    URI(request.referer).host.sub!(/^www./, '')
+    domain = URI(request.referer).host.downcase
+    domain.start_with?('www.') ? domain[4..-1] : domain
   end
 
   def embeddable?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -55,13 +55,11 @@ class ApplicationController < ActionController::Base
   end
 
   def enable_embedded_shopfront
-    whitelist = Spree::Config[:embedded_shopfronts_whitelist]
-    domain = embedded_shopfront_referer
-    return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present? && domain.present? && whitelist.include?(domain)
-    return if request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
+    return unless embeddable?
+    return if embedding_without_https?
 
     response.headers.delete 'X-Frame-Options'
-    response.headers['Content-Security-Policy'] = "frame-ancestors #{domain}"
+    response.headers['Content-Security-Policy'] = "frame-ancestors #{embedded_shopfront_referer}"
 
     check_embedded_request
     set_embedded_layout
@@ -70,6 +68,16 @@ class ApplicationController < ActionController::Base
   def embedded_shopfront_referer
     return if request.referer.blank?
     URI(request.referer).host.sub!(/^www./, '')
+  end
+
+  def embeddable?
+    whitelist = Spree::Config[:embedded_shopfronts_whitelist]
+    domain = embedded_shopfront_referer
+    Spree::Config[:enable_embedded_shopfronts] && whitelist.present? && domain.present? && whitelist.include?(domain)
+  end
+
+  def embedding_without_https?
+    request.referer && URI(request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
   end
 
   def check_embedded_request
--- a/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
+++ b/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
@@ -22,10 +22,10 @@ feature "Using embedded shopfront functionality", js: true do
       add_variant_to_order_cycle(exchange, variant)
 
       Spree::Config[:enable_embedded_shopfronts] = true
-      Spree::Config[:embedded_shopfronts_whitelist] = 'localhost'
+      Spree::Config[:embedded_shopfronts_whitelist] = 'test.com'
 
       page.driver.browser.js_errors = false
-      allow_any_instance_of(ApplicationController).to receive(:embedded_shopfront_referer).and_return('localhost')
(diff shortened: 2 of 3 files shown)


Matt-Yorkley:
@lin-d-hop This one's ready for testing. I've updated the specs to make sure the headers only contain one domain, so I think we just need a quick check to make sure embedded shopfronts look ok.

app/controllers/application_controller.rb
```diff
@@ -56,16 +56,22 @@ def restrict_iframes
 
   def enable_embedded_shopfront
     whitelist = Spree::Config[:embedded_shopfronts_whitelist]
-    return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present?
+    domain = embedded_shopfront_referer
+    return unless Spree::Config[:enable_embedded_shopfronts] && whitelist.present? && domain.present? && whitelist.include?(domain)
```
sauloperez: IMO it's now worth extracting this now long boolean logic into a more abstract method such as `embeddable?` or something alike. It will make things easier to understand to the next dev.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
```diff
@@ -25,6 +25,7 @@
       Spree::Config[:embedded_shopfronts_whitelist] = 'localhost'
 
       page.driver.browser.js_errors = false
+      allow_any_instance_of(ApplicationController).to receive(:embedded_shopfront_referer).and_return('localhost')
```
sauloperez: It's generally a bad practice to stub the class under test, because then the test does not execute the code that you just added. Have you tried stubbing the request instead?

spec/requests/embedded_shopfronts_headers_spec.rb
```diff
@@ -43,16 +43,19 @@
 
     context "with a valid whitelist" do
       before do
-        Spree::Config[:embedded_shopfronts_whitelist] = "test.com"
+        Spree::Config[:embedded_shopfronts_whitelist] = "example.com external-site.com"
+        allow_any_instance_of(ActionDispatch::Request).to receive(:referer).and_return('http://www.external-site.com/shop?embedded_shopfront=true')
```
sauloperez: You can have a reference to the actual request object from `controller.request`. That's more concise and less verbose IMO.

app/controllers/application_controller.rb
enricostano: I agree, moving it to a method would also increase "documentation surface" (i.e. method documentation, tests, etc)

spec/requests/embedded_shopfronts_headers_spec.rb
Matt-Yorkley: I had real trouble getting these extra tests to work. In this instance, the `controller.request` object is not available, as we are viewing the page "through" an iframe. We can see what's in the iframe, but not access the controller directly. I know it's a weird spec, but I couldn't see any other way to test how an iframe loads content from another page.

sauloperez: :+1:

app/controllers/application_controller.rb
Matt-Yorkley: Done. It looks much nicer now.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
Matt-Yorkley: I've stubbed the request instead and improved the test.

Matt-Yorkley:
@sauloperez Could you re-check this if you have time? The requested changes have been made.

daniellemoorhead:
@lin-d-hop @Matt-Yorkley you guys good to test this or should we add it to Sally's test queue?

myriamboure:
@daniellemoorhead  I'm happy to test if someone can stage this.

myriamboure:
@Matt-Yorkley I tested this branch this morning, just realizing that some tests are not passing though... anyway the fix is not working, I tested in two different contexts that I describe in my testing notes here: https://docs.google.com/document/d/162UsCB28c5WhD2wwRp00IPpDG-TYpR754_6YvUWKi9k/edit?usp=sharing
@sstead I took the freedom to use the demo embed on the user guide as I had access to the wordpress so could try with embedding another url. But I put all back in place and it's working as before

The first part of the error message is different in both cases so I let you have a look, but the second one is common, something like 
'https://staging1.openfood.com.au/caringo-buying-group/shop?embedded_shopfront=true' i a frame because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'none'"

There is even a regression as now even when I put the domain first position in the whitelist the embed doesn't work.

I give that back to you @Matt-Yorkley

Matt-Yorkley:
I think this is related to the header settings for the various servers. I have a PR in ofn-install to fix this, but I'm not sure about updating the AUS staging server. 

Maybe @enricostano has some advice?

mkllnk:
@Matt-Yorkley Our staging1 does not set X-Frame-Options. I left your code in our staging server. Maybe you can test if it works?

myriamboure:
I didn't understand @mkllnk what you did (too technical to me). What I understood is that you just left this PR on staging (not the OFN install fix) but how can we test then the OFN-install fix? I tried to go on Aus staging to see if I could test again but got the snail :-( So how can we move forward to test if Matt's fix is working or not? @Matt-Yorkley maybe you understood? We really need to get that fixed, Kirsten shared another feedback from an Aus user today on this. Need some guidance on how to move that forward.

myriamboure:
Ok so I think I understood, #102 in OFN-install is not yet merged, so we wait for @Matt-Yorkley to do the changes requested there, then merge there, and then we need to deploy this branch using OFN-intall to test if the fix work. Did I undertand well @Matt-Yorkley ?

myriamboure:
Deployed on https://staging.openfoodfrance.fr/ with OFN-install branch https://github.com/openfoodfoundation/ofn-install/pull/121#issuecomment-368571081

myriamboure:
It works! Seems the fix on branch OFN-install 121 did the job: https://docs.google.com/document/d/1EL_Cdwcx3U7LxrtLa4mwmpo6hGSiT4gzOqO-TyBfrao/edit?usp=sharing
All good, ready to go!

myriamboure:
Needs to be merged alongside with https://github.com/openfoodfoundation/ofn-install/pull/121

Matt-Yorkley:
Yay!

lin-d-hop:
On deploy to UK this appears to have broken embedded shopfronts. @Matt-Yorkley can you take a look?

myriamboure:
We have deployed the version 1.12 where this was deployed and it fixed the issue (I had issues though when using trackers blockers, it prevents the shopfront to open within any website). But I don't think this broke embedded shopfronts as it fixed it in France @lin-d-hop so if embedded shopfronts are broken there must be another reason.

PR #2028, opened 2017-12-24
Add powered by to embedded groups

KeirOsborn:
Part of task #1783

"move the 'powered by open food network' to somewhere lower on the page (the element got removed)"

This pull enacts the above. I.e. it just adds a small bit of text to the group template saying "Powered by open food network". This only appears when a group page is embedded. Also contains an initial test for embedded group functionality as none existed.

What should we test?
Tests in spec/features/consumer/shopping/embedded_groups_spec.rb

Also functionality can be manually tested using the embedded-group-preview.html file. i.e. by creating a group and then navigating to: http://localhost:3000/embedded-group-preview.html?your-group-name.

Release notes
First of four subtasks. Does not solve issue #1783 by itself.

### DIFF
--- a/app/assets/stylesheets/darkswarm/embedded_shopfront.css.scss
+++ b/app/assets/stylesheets/darkswarm/embedded_shopfront.css.scss
@@ -51,6 +51,15 @@ nav.top-bar ul.right li.powered-by {
   }
 }
 
+.powered-by-embedded {
+   opacity: 0.6;
+   font-family: "Oswald", sans-serif;
+   font-size: 1rem;
+   font-weight: 300;
+   color: #555;
+   padding: 0 !important;
+ }
+
 .blocked-cookies {
   text-align: center;
   margin-bottom: 0 !important;
--- a/spec/features/consumer/shopping/embedded_groups_spec.rb
+++ b/spec/features/consumer/shopping/embedded_groups_spec.rb
@@ -5,30 +5,45 @@ feature "Using embedded shopfront functionality", js: true do
   Capybara.server_port = 9999
 
   describe 'embedded groups' do
-	let(:enterprise) { create(:distributor_enterprise) }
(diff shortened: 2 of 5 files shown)


app/assets/stylesheets/darkswarm/embedded_shopfront.css.scss
```diff
@@ -51,6 +51,15 @@ nav.top-bar ul.right li.powered-by {
   }
 }
 
+.powered-by-embedded {
+   opacity: 0.6;
+   font-family: "Oswald", sans-serif;
```
sauloperez: I haven't looked at any scss file in OFN so far, so I wonder if we have a var for the font-family already.

spec/features/consumer/shopping/embedded_groups_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      Spree::Config[:enable_embedded_shopfronts] = true
+      Spree::Config[:embedded_shopfronts_whitelist] = 'localhost'
+
+      page.driver.browser.js_errors = false
+      Capybara.current_session.driver.visit('spec/support/views/group_iframe_test.html')
+    end
+
+    after do
+      Spree::Config[:enable_embedded_shopfronts] = false
+    end
```
sauloperez: RSpec's `around` is better suited than pairs of `before` and `after`. I'd use it here. See https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/around-hooks. What do you think?

spec/features/consumer/shopping/embedded_groups_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+feature "Using embedded shopfront functionality", js: true do
+
+  Capybara.server_port = 9999
```
sauloperez: Why is this needed? don't we have this configured already?

PR #246, opened 2014-07-03
INSTALL.sh

mkllnk:
I wrote an INSTALL.sh to automate a local installation. It is basically executing all commands of README.markdown plus setting up the PostgreSQL database.

PR #2033, opened 2018-01-02
Add setup script

sauloperez:
#### What? Why?

It aims to aid new open source contributors on setting up their dev env by means of a single command with a meaningful output.

Although ofn-install ansible scripts also work for development we don't want to add too much burden to those single-time beginner contributions.

#### Some example output

```shell
➜  openfoodnetwork add-setup-script ✗ script/setup
The Gemfile's dependencies are satisfied
CREATE ROLE
-- create_table("account_invoices", {:force=>true})
   -> 0.5013s
...
[db:seed] Seeding Spree
...
Create the admin user (press enter for defaults).
Email [spree@example.com]: Password [spree123]: stty: stty: stty:
Done!
[db:seed] Seeding states

[openfoodnetwork:dev:load_sample_data] Seeding taxonomies
[openfoodnetwork:dev:load_sample_data] Seeding addresses
[openfoodnetwork:dev:load_sample_data] Seeding enterprises
[openfoodnetwork:dev:load_sample_data] Seeding products

WELCOME TO OPEN FOOD NETWORK!

To login as Spree default user, use:

  email: spree@example.com
  password: spree123
```

#### Missing

- [x] Provide password when creating `ofn` pg user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,41 +45,15 @@ You can download the source with the command:
 
 ### Get it running
 
-For those new to Rails, the following tutorial will help get you up to speed with configuring a [Rails environment](http://guides.rubyonrails.org/getting_started.html).  If you run into any issues getting your local environment up and running please consult [the wiki](https://github.com/openfoodfoundation/openfoodnetwork/wiki).
+For those new to Rails, the following tutorial will help get you up to speed with configuring a [Rails environment](http://guides.rubyonrails.org/getting_started.html).
 
+When ready, run `script/setup`. If the script succeeds you're ready to start developing. If not, take a look at the output as it should be informative enough to help you troubleshoot.
 
-First, check your dependencies: Ensure that you have Ruby 2.1.5 installed:
+If you run into any other issues getting your local environment up and running please consult [the wiki](https://github.com/openfoodfoundation/openfoodnetwork/wiki).
 
-    ruby --version
+If still you get stuck do not hesitate to open an issue reporting the full output of the script.
 
-Install the project's gem dependencies:
-
-    cd openfoodnetwork
-    bundle install
-    
-Configure the site:
-
-    cp config/application.yml.example config/application.yml
-    edit config/application.yml
-
-Create a PostgreSQL user:
-
-* Login as your system postrgresql priviledged user: `sudo -i -u postgres` (this may vary on your OS). Now your prompt looks like: `[postgres@your_host ~]$`
-* Create the `ofn` database superuser and give it the password `f00d`:
-
-```
-createuser -s -P ofn
-```
-
-Create the development and test databases, using the settings specified in `config/database.yml`, and populate them with a schema and seed data:
-```
-bundle exec rake db:setup
-```
-Load some default data for your environment:
-```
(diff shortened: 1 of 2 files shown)


sauloperez:
:hankey: I forgot to install frontend dependencies

sauloperez:
Hey @oeoeaio I'd like to know what you think about this

mkllnk:
Cool, I had the same idea when I started with OFN: https://github.com/openfoodfoundation/openfoodnetwork/pull/246

At that time, Rohan wasn't really interested in it, because such a script usually needs a lot of maintenance and it harder read than the readme when it doesn't work. It's less flexible. I'm open to give this one a go and see if it works.

It would be nice to give some hints how to resolve dependencies. For example:
```
Open Food Network requires ruby 2.1.5, have a look at:
  https://github.com/rbenv/rbenv
```

sauloperez:
Great @mkllnk ! that will make it even more useful for newcomers.

mkllnk:
Hey @sauloperez. What was the reasoning behind setting Postgresql 9.5 as required version? As far as I know it works well with 9.1. My Debian Stretch doesn't have official packages for 9.5, just 9.6.

sauloperez:
I thought it was the minimum version among the ones that we run either in dev, staging and prod.

This is because we don't enforce it as we do with Ruby or node recently. And I believe we must.

enricostano:
@mkllnk https://github.com/openfoodfoundation/ofn-install/issues/99

Matt-Yorkley:
The supported version seems to change quite regularly, and when it does the previous versions become unavailable as packages. If the main purpose of this setup script is to get new contributors set up with the minimum of hassle, I think using the latest/current packages would be the best option. Especially if one of the main drawbacks of a setup script is that they tend to become outdated over time...

enricostano:
IMO we should develop using the same version of dependencies that we use in staging and production. New contributors should do the same. Maybe it's easy to start with the whatever version that they have on their machine but we can face worst and unclear issues later.

sauloperez:
Agree @Matt-Yorkley but more than that we want them to use the exact versions of the software we use so that the app works as expected. Just imagine all the problems new contributors would experience setting up versions we haven't tried.

As an example, in another Rails app I worked on using MySQL 5.6 and Rails 3, we found out that people that installed MySQL 5.7 couldn't run migrations. It turns out Rails and that version of MySQL are not compatible.

mkllnk:
I understand that we want to avoid any incompatibilities, but I would also like to throw in that too specific version requirements can lead to the dependency hell we are seeing with some of our gems. OFN is not used as gem (yet), but if we know that OFN works with Postgres 9.1 <=> 9.6 for example, it would be nice to allow for that. People have different reasons to install packages from different sources.

I, for example, always install my distribution's packages, because

- I receive security updates automatically,
- they are well maintained,
- they are signed (no untrusted code on my machine).

Let's see which feedback we get from people coming to the project.

sauloperez:
> Let's see which feedback we get from people coming to the project.

👍

PR #2035, opened 2018-01-04
Update rubocop.yml

Matt-Yorkley:
#### What? Why?

Rubocop issues are popping up suggesting refactoring with the #find_by method, which doesn't exist before Rails 4.

#### What should we test?

Nothing to test.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,6 +53,10 @@ Lint/UselessAssignment:
 Rails/DynamicFindBy:
   Enabled: false
 
+# Same as above, #find_by is not available until Rails 4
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
It looks like my initial idea didn't fix anything... 😢  I've added a specific rubocop change for the Rails/FindBy rule instead.

It looks like the other issues I enountered are due to new codeclimate settings that enforce rules almost identical to pre-existing rubocop rules that we have previously disabled?

Matt-Yorkley:
I've modified the original PR description and yml changes. I'll open a separate issue for the other rubocop problems.

PR #2038, opened 2018-01-08
Update README.md

boveus:
#### What? Why?

Hi there, this is my first contribution to this project, but myself and a few people I was working with had an issue getting the initial environment set up.  Hopefully adding this to the readme will help future contributors.  There isn't an issue related to this that I could find.

I encountered an issue with the libv8 gem when getting the project configured locally.  Having some sort of troubleshooting step for that error may help future contributors get up and running faster.  @erose357 @amhursh encountered the same issue and helped with the troubleshooting.

#### What should we test?

This affects the readme and not any of the core code.

#### Release notes

More details of this solution can be found here:
https://stackoverflow.com/questions/19673714/error-installing-libv8-error-failed-to-build-gem-native-extension

This should not affect the core code, since it only changes the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ You can download the source with the command:
 
 ### Get it running
 
-For those new to Rails, the following tutorial will help get you up to speed with configuring a Rails environment: http://guides.rubyonrails.org/getting_started.html .
+For those new to Rails, the following tutorial will help get you up to speed with configuring a Rails environment: http://guides.rubyonrails.org/getting_started.html.  If you run into any issues getting your local environment up and running, please consult the wiki for this repository.
 
 First, check your dependencies: Ensure that you have Ruby 2.1.5 installed:
 
@@ -56,12 +56,6 @@ Install the project's gem dependencies:
     cd openfoodnetwork
     bundle install
     
-If you run into an issue installing libv8 when you bundle, try running the following commands:
-
-    gem install libv8 -v ‘3.16.14.11’ -- --with-system-v8 
-    or
-    gem install libv8 -v 3.16.14.11 -- --with-system-v8
-
 Configure the site:
 
     cp config/application.yml.example config/application.yml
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Thanks @boveus. I think this might be better placed in the wiki than the readme? There have been a few dev setup bugs like this in the past, but they seem to be very system-specific. Could you post the details like OS and version?

boveus:
@Matt-Yorkley, I would be happy to add something about it to the wiki.  Three of us experienced this issue using various flavors of OSX (I am on Sierra) using Macbooks or Macbook pros.  It looks like there is a section here related to getting the environment set up for Sierra: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Development-Environment-Setup%3A-macOS-%28Sierra%29 . Should I go ahead and add that info in there somewhere?

I am not sure if this is an issue related to Sierra, since it seems to be a problem with the libV8 gem itself, but I haven't tried to get that up and running in a different environment. 

I also think it might be helpful to tell folks to refer to the Wiki for specific problems in the Readme.  I wasn't aware of the Wiki when I initially went through the environment setup.

Thanks for the quick response.  I look forward to contributing to this project.

sauloperez:
Agree with @Matt-Yorkley .

>  also think it might be helpful to tell folks to refer to the Wiki for specific problems in the Readme.

Yep @boveus I'd do exactly this.

boveus:
@Matt-Yorkley @sauloperez  I just changed the PR to include a reference to the repo's wiki and will be editing the wiki with information about the libv8 issue.  Does this look okay?   Thanks for the warm welcome and feedback.

boveus:
Hey @sauloperez, I went ahead and changed that to be a hyperlink instead.  I was trying to stay consistent with the style of the link in the rest of the paragraph, so I changed the other link as well.

PR #2039, opened 2018-01-09
2023 - Shipping/payment methods disappears for manager on uncheck

### LINKED ISSUE #2023: Uncheck shipping/payment method make the method disappear when only user
## Description
When a shipping/payment method only apply to one entreprise, if that entreprise wants to unapply that method, the method disappear from the entreprise menu and he can't put it back if needs be.

## Expected Behavior
In the entreprise menu, if I uncheck "applied?" checkbox, the method should just appear as unapplied, but still be there.

## Actual Behavior
If you actually uncheck a method now, it doesn't just deactivate the method, but it completely disappears unless other enterprises are using the same method. Only super-admin can bring that method back.

## Steps to Reproduce
1. In the entreprise menu click on "shipping method". 
2. Ensure you have two methods created, and both are used only by your entreprise.
3. Uncheck the "apply?" checkox for one of them. And save.
4. The method has disappeared.

## Animated Gif/Screenshot
https://www.useloom.com/share/121339b8217a48f2ba02e481aa85c675

## Context
We noticed that issue when working on this other bug https://github.com/openfoodfoundation/openfoodnetwork/pull/1979#issuecomment-353486384

## Severity
3 (you can ask a super admin to restore the method, or create it again)

## Your Environment
* Version used: 1.10.1

yasirazgar:
#### What? Why?
When manager unchecks shipping/payment method it disappears.

The reason for this is that if the user is not admin, only the payment/shipping methods that are applicable to the current enterprise/ the enterprise which the logged in manager manages is displayed.
Refer :managed_by scope in both payment/shipping method.

Closes #2023 

Instead of using the managed_by scope, used all the shipping/payment methods.

#### What should we test?

Login as a manager and need to test shipping/payment methods of the enterprise which is managed by the manager.

### DIFF
--- a/app/controllers/admin/enterprises_controller.rb
+++ b/app/controllers/admin/enterprises_controller.rb
@@ -172,8 +172,8 @@ module Admin
     end
 
     def load_methods_and_fees
-      @payment_methods = Spree::PaymentMethod.managed_by(spree_current_user).sort_by!{ |pm| [(@enterprise.payment_methods.include? pm) ? 0 : 1, pm.name] }
-      @shipping_methods = Spree::ShippingMethod.managed_by(spree_current_user).sort_by!{ |sm| [(@enterprise.shipping_methods.include? sm) ? 0 : 1, sm.name] }
+      @payment_methods = Spree::PaymentMethod.all.sort_by!{ |pm| [(@enterprise.payment_methods.include? pm) ? 0 : 1, pm.name] }
+      @shipping_methods = Spree::ShippingMethod.all.sort_by!{ |sm| [(@enterprise.shipping_methods.include? sm) ? 0 : 1, sm.name] }
       @enterprise_fees = EnterpriseFee.managed_by(spree_current_user).for_enterprise(@enterprise).order(:fee_type, :name).all
     end
 
--- a/spec/controllers/admin/enterprises_controller_spec.rb
+++ b/spec/controllers/admin/enterprises_controller_spec.rb
@@ -107,23 +107,23 @@ module Admin
       let(:enterprise1) { create(:distributor_enterprise) }
       let(:enterprise2) { create(:distributor_enterprise) }
       let(:manager) { create(:user, enterprise_limit: 10, enterprises: [enterprise2]) }
-      let(:time_zone) { Spree::Zone.create(:name => 'South zone')}
+      let(:time_zone) { Spree::Zone.create(name: 'South zone')}
       let(:shipping_method1) { create(:shipping_method, name: 'Pickup', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise1.id]) }
       let(:shipping_method2) { create(:shipping_method, name: 'Delivery', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise2.id]) }
-      let(:payment_method1) { create(:payment_method, :name => 'Online', distributor_ids: [enterprise1.id]) }
-      let(:payment_method2) { create(:payment_method, :name => 'COD', distributor_ids: [enterprise2.id]) }
-      let(:enterprise_fee1) { create(:enterprise_fee, :enterprise_id => enterprise1.id) }
-      let(:enterprise_fee2) { create(:enterprise_fee, :enterprise_id => enterprise2.id) }
+      let(:payment_method1) { create(:payment_method, name: 'Online', distributor_ids: [enterprise1.id]) }
+      let(:payment_method2) { create(:payment_method, name: 'COD', distributor_ids: [enterprise2.id]) }
+      let(:enterprise_fee1) { create(:enterprise_fee, enterprise_id: enterprise1.id) }
+      let(:enterprise_fee2) { create(:enterprise_fee, enterprise_id: enterprise2.id) }
 
       context "as admin" do
         it "sets required variables" do          
-          allow(controller).to receive_message_chain(:spree_current_user).and_return(admin_user)            
+          allow(controller).to receive(:spree_current_user).and_return(admin_user)            
 
           payment_methods = [payment_method1, payment_method2]
           shipping_methods = [shipping_method1, shipping_method2]
           enterprise_fees = [enterprise_fee1]
 
-          spree_get :edit, {:id => enterprise1.permalink} 
+          spree_get :edit, {id: enterprise1.permalink} 
 
           expect(assigns(:shipping_methods)).to eq shipping_methods   
           expect(assigns(:payment_methods)).to eq payment_methods   
@@ -133,13 +133,13 @@ module Admin
 
       context "as manager" do
         it "sets required variables" do                         
-          allow(controller).to receive_message_chain(:spree_current_user).and_return(manager)        
+          allow(controller).to receive(:spree_current_user).and_return(manager)        
 
           payment_methods = [payment_method2, payment_method1]
           shipping_methods = [shipping_method2, shipping_method1]
           enterprise_fees = [enterprise_fee2]          
 
-          spree_get :edit, {:id => enterprise2.permalink} 
+          spree_get :edit, {id: enterprise2.permalink} 
 
           expect(assigns(:shipping_methods)).to eq shipping_methods
           expect(assigns(:payment_methods)).to eq payment_methods
--- a/spec/controllers/admin/enterprises_controller_spec.rb
+++ b/spec/controllers/admin/enterprises_controller_spec.rb
@@ -108,10 +108,10 @@ module Admin
       let(:enterprise2) { create(:distributor_enterprise) }
       let(:manager) { create(:user, enterprise_limit: 10, enterprises: [enterprise2]) }
       let(:time_zone) { Spree::Zone.create(:name => 'South zone')}
-      let(:shipping_method1) { Spree::ShippingMethod.create(name: 'Pickup', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise1.id]) }
-      let(:shipping_method2) { Spree::ShippingMethod.create(name: 'Delivery', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise2.id]) }
-      let(:payment_method1) { Spree::PaymentMethod.create(:name => 'Online', distributor_ids: [enterprise1.id]) }
-      let(:payment_method2) { Spree::PaymentMethod.create(:name => 'COD', distributor_ids: [enterprise2.id]) }
+      let(:shipping_method1) { create(:shipping_method, name: 'Pickup', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise1.id]) }
+      let(:shipping_method2) { create(:shipping_method, name: 'Delivery', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise2.id]) }
+      let(:payment_method1) { create(:payment_method, :name => 'Online', distributor_ids: [enterprise1.id]) }
+      let(:payment_method2) { create(:payment_method, :name => 'COD', distributor_ids: [enterprise2.id]) }
       let(:enterprise_fee1) { create(:enterprise_fee, :enterprise_id => enterprise1.id) }
       let(:enterprise_fee2) { create(:enterprise_fee, :enterprise_id => enterprise2.id) }
 
--- a/spec/controllers/admin/enterprises_controller_spec.rb
+++ b/spec/controllers/admin/enterprises_controller_spec.rb
@@ -102,6 +102,52 @@ module Admin
       end
     end
 
+    describe "edit" do
+      let(:admin_user) { create(:admin_user) }      
+      let(:enterprise1) { create(:distributor_enterprise) }
+      let(:enterprise2) { create(:distributor_enterprise) }
+      let(:manager) { create(:user, enterprise_limit: 10, enterprises: [enterprise2]) }
+      let(:time_zone) { Spree::Zone.create(:name => 'South zone')}
+      let(:shipping_method1) { Spree::ShippingMethod.create(name: 'Pickup', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise1.id]) }
+      let(:shipping_method2) { Spree::ShippingMethod.create(name: 'Delivery', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise2.id]) }
+      let(:payment_method1) { Spree::PaymentMethod.create(:name => 'Online', distributor_ids: [enterprise1.id]) }
+      let(:payment_method2) { Spree::PaymentMethod.create(:name => 'COD', distributor_ids: [enterprise2.id]) }
+      let(:enterprise_fee1) { create(:enterprise_fee, :enterprise_id => enterprise1.id) }
+      let(:enterprise_fee2) { create(:enterprise_fee, :enterprise_id => enterprise2.id) }
+
+      context "as admin" do
+        it "sets required variables" do          
+          allow(controller).to receive_message_chain(:spree_current_user).and_return(admin_user)            
+
+          payment_methods = [payment_method1, payment_method2]
+          shipping_methods = [shipping_method1, shipping_method2]
+          enterprise_fees = [enterprise_fee1]
+
+          spree_get :edit, {:id => enterprise1.permalink} 
+
+          expect(assigns(:shipping_methods)).to eq shipping_methods   
+          expect(assigns(:payment_methods)).to eq payment_methods   
+          expect(assigns(:enterprise_fees)).to eq enterprise_fees             
+        end        
+      end
+
+      context "as manager" do
+        it "sets required variables" do                         
+          allow(controller).to receive_message_chain(:spree_current_user).and_return(manager)        
+
+          payment_methods = [payment_method2, payment_method1]
+          shipping_methods = [shipping_method2, shipping_method1]
+          enterprise_fees = [enterprise_fee2]          
+
+          spree_get :edit, {:id => enterprise2.permalink} 
+
+          expect(assigns(:shipping_methods)).to eq shipping_methods
+          expect(assigns(:payment_methods)).to eq payment_methods
+          expect(assigns(:enterprise_fees)).to eq enterprise_fees
+        end        
+      end
+    end
+
     describe "updating an enterprise" do
       let(:profile_enterprise) { create(:enterprise, sells: 'none') }
 
--- a/app/models/distributor_payment_method.rb
+++ b/app/models/distributor_payment_method.rb
@@ -1,0 +1,6 @@
+class DistributorPaymentMethod < ActiveRecord::Base
+  self.table_name = "distributors_payment_methods" 
+  belongs_to :payment_method, class_name: Spree::PaymentMethod
+  belongs_to :distributor, class_name: Enterprise, touch: true
+  belongs_to :creator, class_name: Spree::User
+end
--- a/app/models/distributor_shipping_method.rb
+++ b/app/models/distributor_shipping_method.rb
@@ -2,4 +2,5 @@ class DistributorShippingMethod < ActiveRecord::Base
   self.table_name = "distributors_shipping_methods" 
   belongs_to :shipping_method, class_name: Spree::ShippingMethod
   belongs_to :distributor, class_name: Enterprise, touch: true
+  belongs_to :creator, class_name: Spree::User
 end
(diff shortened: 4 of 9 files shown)


app/controllers/admin/enterprises_controller.rb
```diff
@@ -171,8 +171,8 @@ def collection_actions
     end
 
     def load_methods_and_fees
-      @payment_methods = Spree::PaymentMethod.managed_by(spree_current_user).sort_by!{ |pm| [(@enterprise.payment_methods.include? pm) ? 0 : 1, pm.name] }
-      @shipping_methods = Spree::ShippingMethod.managed_by(spree_current_user).sort_by!{ |sm| [(@enterprise.shipping_methods.include? sm) ? 0 : 1, sm.name] }
+      @payment_methods = Spree::PaymentMethod.all.sort_by!{ |pm| [(@enterprise.payment_methods.include? pm) ? 0 : 1, pm.name] }
+      @shipping_methods = Spree::ShippingMethod.all.sort_by!{ |sm| [(@enterprise.shipping_methods.include? sm) ? 0 : 1, sm.name] }
```
sauloperez: This will load all shipping methods available in the database blowing up the server. We should query the data instead, by means of an INNER JOIN.

spec/controllers/admin/enterprises_controller_spec.rb
```diff
@@ -112,6 +112,52 @@ module Admin
...
+    describe "edit" do
+      let(:admin_user) { create(:admin_user) }      
+      let(:enterprise1) { create(:distributor_enterprise) }
+      let(:enterprise2) { create(:distributor_enterprise) }
+      let(:manager) { create(:user, enterprise_limit: 10, enterprises: [enterprise2]) }
+      let(:time_zone) { Spree::Zone.create(:name => 'South zone')}
+      let(:shipping_method1) { Spree::ShippingMethod.create(name: 'Pickup', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise1.id]) }
+      let(:shipping_method2) { Spree::ShippingMethod.create(name: 'Delivery', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise2.id]) }
+      let(:payment_method1) { Spree::PaymentMethod.create(:name => 'Online', distributor_ids: [enterprise1.id]) }
+      let(:payment_method2) { Spree::PaymentMethod.create(:name => 'COD', distributor_ids: [enterprise2.id]) }
```
sauloperez: Aren't there factories for these models already? Spree should have them.

app/controllers/admin/enterprises_controller.rb
yasirazgar: @sauloperez, for admin by default we are showing all the shipping methods, 
```
scope :managed_by, lambda { |user|
      scoped
    else
      joins(:distributors).
        where('distributors_shipping_methods.distributor_id IN (?)', user.enterprises).
        select('DISTINCT spree_shipping_methods.*')
    end
  }
```
for manager also we should show all the shipping methods, that was the issue. So I removed the managed_by scope.

I cant understand how can I fix the issue by using inner join, can you please guide on fixing the issue with more details.

Thank you.

sauloperez: I see. What about the `else` branch's behavior? In the case of non-admin users, we want to show the shipping methods that the enterprise uses. This behavior will be gone if we do not call `:managed_by`. 

Regarding the INNER JOIN, nevermind. We should create another issue to solve it, which will probably mean changing the behavior a bit. Let's stick to the bug fix here.

spec/controllers/admin/enterprises_controller_spec.rb
```diff
@@ -112,6 +112,52 @@ module Admin
...
+      let(:shipping_method1) { create(:shipping_method, name: 'Pickup', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise1.id]) }
+      let(:shipping_method2) { create(:shipping_method, name: 'Delivery', zone_id: time_zone.id, require_ship_address: true, calculator_type: 'OpenFoodNetwork::Calculator::Weight', distributor_ids: [enterprise2.id]) }
+      let(:payment_method1) { create(:payment_method, :name => 'Online', distributor_ids: [enterprise1.id]) }
+      let(:payment_method2) { create(:payment_method, :name => 'COD', distributor_ids: [enterprise2.id]) }
+      let(:enterprise_fee1) { create(:enterprise_fee, :enterprise_id => enterprise1.id) }
+      let(:enterprise_fee2) { create(:enterprise_fee, :enterprise_id => enterprise2.id) }
+
+      context "as admin" do
+        it "sets required variables" do          
+          allow(controller).to receive_message_chain(:spree_current_user).and_return(admin_user)
```
sauloperez: IMO we should use `receive` instead of `receive_message_chain` here. The latter has a different purpose. Check the docs: https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/message-chains

spec/controllers/admin/enterprises_controller_spec.rb
```diff
@@ -112,6 +112,52 @@ module Admin
...
+
+          expect(assigns(:shipping_methods)).to eq shipping_methods   
+          expect(assigns(:payment_methods)).to eq payment_methods   
+          expect(assigns(:enterprise_fees)).to eq enterprise_fees             
+        end        
+      end
+
+      context "as manager" do
+        it "sets required variables" do                         
+          allow(controller).to receive_message_chain(:spree_current_user).and_return(manager)
```
sauloperez: here too

spec/controllers/admin/enterprises_controller_spec.rb
```diff
@@ -112,6 +112,52 @@ module Admin
...
+
+      context "as admin" do
+        it "sets required variables" do          
+          allow(controller).to receive_message_chain(:spree_current_user).and_return(admin_user)            
+
+          payment_methods = [payment_method1, payment_method2]
+          shipping_methods = [shipping_method1, shipping_method2]
+          enterprise_fees = [enterprise_fee1]
+
+          spree_get :edit, {:id => enterprise1.permalink}
```
sauloperez: It's been long since Ruby introduced the Hash symbol syntax and we prefere that one, instead of hash rockets.

sauloperez: @yasirazgar I created a PR to automate this style-related review in the future. Make sure you take a look https://github.com/openfoodfoundation/openfoodnetwork/pull/2045

app/controllers/admin/enterprises_controller.rb
mohammedyasir: Yes @sauloperez, For non-admin users If we use managed_by scope it will only display the shipping/payment methods that are used by the enterprise, the method `load_methods_and_fees` in admin/enterprise-controller will be used by only admins & managers, for managers also we need to display all the shipping/payment methods instead of only the items used by the enterprise. As the managers are non-admin, for them this method will display only the shipping/payment methods used by the enterprise. Also due to this when a manager uncheck an shipping/payment methods they will disappear from the list.

So, instead of using the managed_by scope i used all.

Correct me if I am wrong and please let me know if there is any other way of fixing this.

yasirazgar: Yes @sauloperez, For non-admin users If we use managed_by scope it will only display the shipping/payment methods that are used by the enterprise, the method load_methods_and_fees in admin/enterprise-controller will be used by only admins & managers, for managers also we need to display all the shipping/payment methods instead of only the items used by the enterprise. As the managers are non-admin, for them this method will display only the shipping/payment methods used by the enterprise. Also due to this when a manager uncheck an shipping/payment methods they will disappear from the list.

So, instead of using the managed_by scope i used all.

Correct me if I am wrong and please let me know if there is any other way of fixing this.

sauloperez: Thanks! Now I get it all. I don't think there's any other quick way to fix this for now.

myriamboure:
@sauloperez do we need another reviewer before we test? Also @yasirazgar it seems some travis tests are not passing.

sauloperez:
> @sauloperez do we need another reviewer before we test?

I think it's fine. I'd move to test.

yasirazgar:
@myriamboure fixed the broken tests

myriamboure:
@RachL if you have finished with email confirm on French staging, can you see with @pacodelaluna if we can stage this, and test? Please document your testing notes in a google doc, and just share here if some things are still not working + the link to the doc :) That would be awesome ! And ask if anything is not clear, normally the issue is pretty well described as follow the new template, but don't hesitate.

RachL:
@myriamboure No problem, I'm on it :)

pacodelaluna:
@RachL I have deployed the branch on french staging, got some issues on data side (confirmed_at field was missing on 'enterprises' table, fixed it by adding it manually), related to too many differences between branches I guess.

myriamboure:
Hum... i  don't understand @pacodelaluna why you should add such things manually... is it normal @enricostano @yasirazgar that we need to do that when deploying that branch? Just worried that we test something that is not working properly on the data side if you had to do that when deploying...?

sauloperez:
that's weird. That must be due to the branch that was deployed there before. In any case, we should run `bundle exec rake db:reset openfoodnetwork:dev:load_sample_data` always. This ensure we have the same database schema as in the branch plus sample data to test. I don't know if you do this already @pacodelaluna

myriamboure:
Can we still test @sauloperez or do you think it's better to redeploy without getting any error before testing?

sauloperez:
Sure, you can test. @pacodelaluna he fixed it, so nothing to worry about.

myriamboure:
Ping @RachL gogogo!

myriamboure:
@yasirazgar I tested the PR and it seems there are still some issues:
With a "non admin" account (here producer shop account): 
- when I uncheck a shipping method or check an unchecked one, and save, I get the error snail page
- as a hub  manager I can see all the unchecked shipping method, even if I didn't created them or they were never applied to my hub. I should  only see the one I created / that were applied to that hub before.

Here are the testing notes: https://docs.google.com/document/d/1sGPvX8FpbtZQw6UlCH2C3EVSDU5RiKUGOzDS2tWrKbY/edit#

yasirazgar:
@myriamboure, ok so, a hub manager should see only the shipping/payment methods created by him.
I will check and fix it.

yasirazgar:
@sauloperez, I checked, but cannot find a way to track shipping/payment methods added by the hub-manger or one that was added to the enterprise before by the admin.
I think we need to add a new column/feature to track them.
Can you please guide me on fixing this, let me know If missing something.

myriamboure:
@yasirazgar I would say a hub manager should see only the shipping/payment methods created by him for a given entreprise, or that were once attributed to that entreprise.
That includes the  case where  an admin create a shipping method for a hub, then the hub use it, but some day wants to unapply itbut they have been using it, so even if he didn't create it, it should remain in the list.
Or  other case: a hub have two managers, one create the shipping method, but the other sholud see them even if unapplied.

I guess tracking the user id who created the method should be doable, but given what I said above, isn't it more logical to track: has a method ever been applied or not to a hub? If yes then will remain visible for all managers of the hub until someone delete the method. If not doesn't appear in the hub list.
I think by default when you create a shipping method it appears as "applied" (you have to select the hub you want the method to apply to when you create it) so that should work. What do you think @sauloperez ?

myriamboure:
@yasirazgar we redeployed your branch as I had some doubts about the snail (if that was something from our side) and it was a deployment issue, so when I uncheck a shipping method only use by my hub, I can uncheck it and still see it in my list, so this is working.

BUT there is an issue when I check one method, regardless if it is used by other entreprises or not, if that's the only shipping method I have checked, I can't uncheck it. So basically it's not possible to have no shipping method selected.
I like this behavior actually because a shop can't be open until there is a shipping method set up. So I like to kind of "force" that behavior as soon as one shipping method has been selected once, so I would vote to keep it like that. 
https://www.useloom.com/share/87d2fb23f500488fb75b43cee46eae65
Ping @sstead for confirmation.

So the only remaining things to fix is to hide shipping methods that never has been applied to a hub.

yasirazgar:
@myriamboure @sauloperez, I checked the shipping_method model and its associations, currently there is no way to track the users by whom the shipping method was created or enterprise/distributor for which the shipping_method was applied at least once.

I investigated a bit,
For shipping_method there is a join_table model `distributors_shipping_methods`, we can add additional columns to this table
1. `creator_id` - to track the user who created the shipping_method.
2. `enabled` - of type boolean, for tracking whether the shipping_method was/is added to an enterprise. Currently on enabling a shipping_method for a enterprise we are adding a record with shipping_method_id and enterprise_id to this table, and removing it on unchecking, instead we can retain the record when it is added for the first time, and use this `enabled` field on check/uncheck.

However for payment_methods, there is only a join table (`distributors_payment_methods`), it is not a model, we can convert it into a model like `distributors_shipping_methods`.

Or leaving all together we can add a new table common to both shipping/payment methods, for tracking the details with columns `distributor_id`, `item_id`, `creator_id`,`item_type`. But it duplicates the data as we are already tracking shipping/payment methods against distributors.

In either case we need to handle the existing data.

Can I proceed with the former one, and please let me know if there is an better for handling this.

myriamboure:
@sauloperez any technical input/opinion on @yasirazgar suggestion? So that he can move forward on implementing a solution to solve that :-)

sauloperez:
I would go with adding the required columns to `distributors_*` tables as you @yasirazgar suggest. Let's not forget to add the necessary indexes once we know the kind of queries we use.

yasirazgar:
@sauloperez I need to update `Spree::Admin::ShippingMethodsController` and related views which are available in spree gem, I am not sure how to do this, can you please guide me on this.

sauloperez:
Sure @yasirazgar ! Spree's views are overridden using https://github.com/spree/deface, which is their own mechanism for customizing Spree. So, whatever you need to do will be in [app/overrides/spree/admin/shipping_methods](https://github.com/openfoodfoundation/openfoodnetwork/tree/master/app/overrides/spree/admin/shipping_methods).

For other stuff related to business logic, you might need to take a look at our [Spree fork](https://github.com/openfoodfoundation/spree) to see the things we've changed. Keep in mind though, that we want to move away from such practices as they make the maintenance of the project extremely hard. I hope you don't need to touch anything there.

mkllnk:
To my knowledge, there is no such documentation. Sorry. A developers introduction guide would be nice.

sauloperez:
Well, that's better than nothing

myriamboure:
Any progress on this @yasirazgar ? Did you get the answeres you needed? It seems like a bit tricky one... tell us if it's a bit overcomplex as a first time contributor and we can see if you can find you some help.

yasirazgar:
@myriamboure I have got all the answers, I will try to complete by this weekend.

yasirazgar:
@myriamboure @sauloperez 
This is getting complicated in every step,
I have overridden the behavior of the join table and I have to manually create the join table record and track the two additional columns in it. Its inefficient and hits db a lot.
Example on updating an enterprise, i am using this callback, for payment/shipping methods.
```
def update_ship_and_pay_methods
      if (ship_ids = params[:enterprise].delete(:shipping_method_ids)) && (ship_ids.map(&:to_i).sort != @object.shipping_methods.pluck(:id).sort)
        create_distributor_shipping_methods_if_new(pay_ids)
        rel = DistributorShippingMethod.where(:distributor_id => @object.id)
        if ship_ids.present?
          rel.where(:shipping_method_id => ship_ids).update_all(:enabled => true)
          rel = rel.where("shipping_method_id != (?)", ship_ids)
        end
        rel.update_all(:enabled => false)
      end

      if (pay_ids = params[:enterprise].delete(:payment_method_ids)) && (pay_ids.map(&:to_i).sort != @object.payment_methods.pluck(:id).sort)
        create_distributor_payment_methods_if_new(pay_ids)
        rel = DistributorPaymentMethod.where(:distributor_id => @object.id)        
        if pay_ids.present?
          rel.where(:payment_method_id => pay_ids).update_all(:enabled => true)
          rel = rel.where("payment_method_id != (?)", pay_ids)
        end          
        rel.update_all(:enabled => false)
      end
    end
```
Similarly one creating or updating a shipping/payment method I need to write a complex logic, 

I was thinking maybe we can just add those two columns to shipping/payment method models
 1. `enabled_for_enterprises ` -  Array - Track the enterprises to which it is added.
 2. `creator_id`. - User who created it.
Though the array_column may look bad, I think its better than the previous solution.

What do you guys think, please let me know your thoughts.

sauloperez:
I still have to go through all we discussed here again. It's been too long for my little brain :sweat: .

What I can see is that things are getting too complex and tedious for an issue that hasn't been even prioritized. I don't want to waste our efforts.

Regarding `enabled_for_enterprises` I have no experience with arrays in Postgres but storing associations in arrays in relational databases is generally a bad idea as you make plain joins not possible, which is the basic operation these databases are meant to do.

It clearly feels like a new table that associates both, shipping and payment methods and enterprises, so again. I'm not sure it's the right issue to tackle now.

oeoeaio:
Sorry that I am late so this. I am not necessarily advocating that this happens right now, but I think it is important to look at the ideal solution and plan a roadmap to get there.

IMHO the 'OFN-way' to do this would be to make each shipping and payment method owned by a single enterprise. This would allow us to use our existing permissions conventions to allow users to access/modify resources based on the enterprises they own/manage, and would provide clarity around who has permission to alter and delete each shipping/payment method.

Using a new ownership-based model, the methods that are visible to a particular user for selection would be based on:
1) the enterprises that they manage, and the methods belonging to those enterprises
2) the methods their enterprises have been granted to their enterprises via permissions

I am thinking that ownership should be modelled via a `enterprise_id` or `owner_id` field on the shipping/payment method itself. The nice thing about this is that the modelling of which enterprise is offering which method for use by customers does not need to change, the join tables should be sufficient if we can add the ownership element to the method tables. The creating of records in the join table is limited by the existence of permissions in much the same way as variant overrides.

The hard parts are:
- We would need to add a new E2E permission: `use_shipping_and_payment_methods`
- We need a migration that allocates each existing shipping and payment method to an enterprise and grants appropriate permissions without breaking the existing use of methods.

mkllnk:
I agree with Rob that we ideally have payment methods associated to enterprises (shops). Do we have any use case to share a payment method?

My idea would be:

- Introduce a relation (table) to associate one enterprise to several payment methods. It needs a restriction that the payment method id is unique and can't be associated to multiple enterprises. Instead of creating a new table, we could introduce an enterprise_id on the payment table, but I would like to avoid touching Spree tables.
- The payment methods that are currently shared between several shops need to be duplicated, one for each shop.
- The additional permission is nice to have, but probably not essential for the first step. Someone who has full permission to an enterprise can change its payment method.
- When payment methods appear in the enterprise edit pages, we don't need any other way of changing payment methods and can abandon Spree's interface (even for super admin for simplicity).

oeoeaio:
> Do we have any use case to share a payment method?
Yes, absolutely. Hubs that sell through multiple pickup locations which they have opted to model as separate enterprises. In this case they want to have a single method (probably payment in this case) that they can alter once. That is what the new permission is for. The permission will determine which enterprises are available to select on the right-hand side when you are editing a payment method.

> Introduce a relation (table) to associate one enterprise to several payment methods
The only reason I prefer adding a field to the user tables instead of adding a new join table to model ownership is that then we end up with two join tables between enterprise and each of the payment/shipping method tables. It just seems more complicated than it needs to be. I get you point about about modifying sprees tables though, so happy to go with the multiple join tables if that is preferred.

> The payment methods that are currently shared between several shops need to be duplicated, one for each shop.
I think we definitely don't want to do this. It breaks the existing functionality described in the use-case above. The advantage of doing the work to create new permissions is that the relationships become transparent, and the owners of methods can see who is able to use them, and remove them if they want to. If we just duplicate them we will probably end up with lots of orphaned methods, which I suppose it fine, but the permissions path seems tidier to me.

> When payment methods appear in the enterprise edit pages, we don't need any other way of changing payment methods and can abandon Spree's interface (even for super admin for simplicity).
You mean that we don't need the index pages? We obviously still need the edit interfaces. I think we probably won't need Spree index pages, but can just be convenient to see all payment methods available to all enterprises managed by a user. Not sure whether this needs to be in scope...

mkllnk:
> Hubs that sell through multiple pickup locations which they have opted to model as separate enterprises. In this case they want to have a single method
Ah, what a weird modelling. Okay, so payment methods belong to a user then. Since they are quite confidential, this could actually be a good idea. I'm wondering if we should then base the whole permission model on users as well.

- permission to use a payment method
- permission to edit a payment method

If a manager of an enterprise has no permission on the payment method, they just see the name, but can't do anything with it. What do you think?

I'm just trying to avoid this situation where a person gives an enterprise permission and then you have to look at the permissions of the enterprise to know if another person can edit the payment method. It seems like mixing concepts.

myriamboure:
Ok so let's sum up what happened here:
- @yasirazgar has spend some time investigating how to solve the issue and hit this massive problem of a bad design on our data model
- @sauloperez realized it's a much bigger job than we thought and not just a bug fix but a whole new feature and wonder if it's a priority right now
- @oeoeaio and @mkllnk have started already to incept how we could modelize that

Brilliant everyone :) And here is my suggestion:
- As we have so many things to do and we implemented a process to prioritize in which order we do them, and this is a whole new feature, [**I opened an icebox item** here](https://community.openfoodnetwork.org/t/hubs-manages-easily-shipping-and-payment-methods-and-can-grant-permissions-on-them/1269), and listed it in the spreadsheet we use to prioritize, so that **we can decide to prioritize it a bit later**.
- I feel really sorry @yasirazgar because you spent so much time on it but it seems that you touched some deep and pretty complex problem that was hidden behind that bug, so **thank you so much for helping us pointing it out!** If you want to keep on contributing I'm happy to suggest you some well specified issues that hopefully will not be as tricky as this one, you were really unlucky! I can suggest for instance: 2105 / 698 / 1250. 
- @oeoeaio @mkllnk @sauloperez you'll **come back later on inception when we decide to prioritize**. And at that point if @yasirazgar is still around we'll invite him to play with us of course :-)

How does that plan sound to all of you?

yasirazgar:
@myriamboure, thats ok, I learned many things while working on this issue, I will look into the issues suggested by you. And yes please let me know when reopening this issue, happy to work on this again.

myriamboure:
closing this one until we prioritize this big piece of work then

PR #2031, opened 2017-12-28
Remove logo, banner, description and title from embedded groups

KeirOsborn:
## What? Why?

Part of task #1783

"Remove the logo, banner and Description text fields"

This pull enacts the above. I.e. It removed the whole header element from from the group page template if it's being viewed as an embedded page. This also includes the group name so that has also been removed (as instructed). 

## What should we test?
Tests in spec/features/consumer/shopping/embedded_groups_spec.rb

Also functionality can be manually tested using the embedded-group-preview.html file. i.e. by creating a group and then navigating to: http://localhost:3000/embedded-group-preview.html?your-group-name.

## Release notes
Third of four subtasks. Does not solve issue #1783 by itself.

PR #2047, opened 2018-01-16
Embedded groups

### LINKED ISSUE #1783: Embedding group pages

KeirOsborn:
Closes #1783 

Six commits. Two for initial tests and setup and four for each of the remaining sub-issues within this issue. The sub issues in question are:

1. move the 'powered by open food network' to somewhere lower on the page (the element got removed)
- It just adds a small bit of text to the group template saying "Powered by open food network". This only appears when a group page is embedded.

2. remove the contact details.
- It removes the contact details from a group page when that page is being viewed in an embedded setting.

3. Remove the logo, banner and Description text fields.
- It removes the whole header element from from the group page template if it's being viewed as an embedded page. This also includes the group name so that has also been removed (as instructed).

4. Links to shops open a new window.
- If a group page is being viewed as an embedded page, all links to the groups corresponding shops are served with the target = '_blank' attribute so that they open in a new browser window/tab.

All of these are solved by this PR. 

#### What should we test?

Tests in spec/features/consumer/shopping/embedded_groups_spec.rb

Also functionality can be manually tested using the embedded-group-preview.html file. i.e. by creating a group and then navigating to: http://localhost:3000/embedded-group-preview.html?your-group-name.

_Screenshots for comparison:_

Standard show view:
screen shot 2018-02-01 at 12 48 20

Embedded view:
screen shot 2018-02-01 at 12 48 00

#### Discourse thread

Discussion here:
https://github.com/openfoodfoundation/openfoodnetwork/issues/1783
and here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2031

### DIFF
--- a/app/assets/stylesheets/darkswarm/embedded_shopfront.css.scss
+++ b/app/assets/stylesheets/darkswarm/embedded_shopfront.css.scss
@@ -51,6 +51,16 @@ nav.top-bar ul.right li.powered-by {
   }
 }
 
+.powered-by-embedded {
+   opacity: 0.6;
+   font-family: "Oswald", sans-serif;
+   font-size: 1rem;
+   font-weight: 300;
+   color: #555;
+   padding: 0 !important;
+  }
+
+
 .blocked-cookies {
   text-align: center;
   margin-bottom: 0 !important;
--- a/spec/features/consumer/shopping/embedded_groups_spec.rb
+++ b/spec/features/consumer/shopping/embedded_groups_spec.rb
@@ -1,0 +1,33 @@
+require 'spec_helper'
+
+feature "Using embedded shopfront functionality", js: true do
+
+  Capybara.server_port = 9999
+
+  describe 'embedded groups' do
+	let(:enterprise) { create(:distributor_enterprise) }
+  	let!(:group) { create(:enterprise_group, enterprises: [enterprise], permalink: 'group1', on_front_page: true) }
+
+
+  	before do
+  	  Spree::Config[:enable_embedded_shopfronts] = true
+      Spree::Config[:embedded_shopfronts_whitelist] = 'localhost'
+
+      page.driver.browser.js_errors = false
+	  Capybara.current_session.driver.visit('spec/support/views/group_iframe_test.html')
+
+  	end
+
+	it "displays in an iframe" do
+	  expect(page).to have_selector 'iframe#group_test_iframe'
+
+	  within_frame 'group_test_iframe' do
+	  	within 'div#group-page' do
+	  	  expect(page).to have_content 'About Us'
+	  	end
+	  end	  
+	end
+  
+  end
+
+end
--- a/spec/features/consumer/shopping/embedded_groups_spec.rb
+++ b/spec/features/consumer/shopping/embedded_groups_spec.rb
@@ -51,6 +51,18 @@ feature "Using embedded shopfront functionality", js: true do
 	  	end
 	  end
 	end
+
+	it "does not display the header when embedded" do
+	  expect(page).to have_selector 'iframe#group_test_iframe'
+
+	  within_frame 'group_test_iframe' do
+	  	within 'div#group-page' do
+	  		expect(page).to have_no_selector 'header'
+	  	    expect(page).to have_no_selector 'img.group-logo'
+	  	    expect(page).to have_no_selector 'h2.group-name'
+	  	end
+	  end
+	end
   
   end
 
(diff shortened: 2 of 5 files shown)


enricostano:
Cool, thanks @KeirOsborn 👏 

Two questions:
 - do you mind 2 screenshots of the `embedded` and `show` views added to the description?
 - is there anything we can re-use via partials between `embedded` and `show` views?

app/assets/stylesheets/darkswarm/embedded_shopfront.css.scss
```diff
@@ -51,6 +51,16 @@ nav.top-bar ul.right li.powered-by {
   }
 }
 
+.powered-by-embedded {
+   opacity: 0.6;
+   font-family: "Oswald", sans-serif;
```
sauloperez: don't we inherit this already? If I'm not mistaken we use this typeface in other places. If that wasn't the case I'd go with a sass variable. thoughts?

spec/features/consumer/shopping/embedded_groups_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'spec_helper'
+
+feature "Using embedded shopfront functionality", js: true do
+
+  Capybara.server_port = 9999
```
sauloperez: Can you check if this is already configured in `spec/spec_helper.rb`? To me, it's too unrelated to this test file to have it here and it looks like a global `Capybara` config.

KeirOsborn:
@enricostano 

- do you mind adding 2 screenshots of the embedded and show views to the description?

Sure, have added. 

- Is there anything we can re-use via partials between embedded and show views?

I don't think so. The problem is that the embedded version needs to have target="_blank" each time there is a link back to the shop. These are scattered throughout the partials so most of them have to be re-written. Unless anyone knows a way to set target="_blank" for a specific link at the controller level. I don't.

enricostano:
@KeirOsborn 

> Unless anyone knows a way to set target="_blank" for a specific link at the controller level

IF that's (almost) the only difference maybe we can pass a `local` variable to the view like `target` and set it in the controller:

`target: @target` instead of `target: "_blank"`

It can also be a view helper probably. Is that the main difference?

app/assets/stylesheets/darkswarm/embedded_shopfront.css.scss
KeirOsborn: @sauloperez 

Cool. Agreed. This does exist as a sass variable already so i've used that instead. Have updated in latest version which I will push shortly.

KeirOsborn:
@enricostano 

Ah yes. Thanks for the advice. Not sure why I didn't think of that. Ok I've done that now. See latest version of the pull request: #2066.

Thanks

PR #2058, opened 2018-01-29
Improve code style and docs of job

sauloperez:
#### What? Why?

Given that I was investigating the "notify producers" button that shows up in the order cycle page, I thought I could take the chance to improve a bit the code while I was reading it. That's all.

#### What should we test?

Nothing. There's no behavior change at all.

### DIFF
--- a/app/jobs/order_cycle_notification_job.rb
+++ b/app/jobs/order_cycle_notification_job.rb
@@ -1,4 +1,4 @@
-# Deliers an email with a report of the order cycle to each of its suppliers
+# Delivers an email with a report of the order cycle to each of its suppliers
 OrderCycleNotificationJob = Struct.new(:order_cycle_id) do
   def perform
     order_cycle = OrderCycle.find(order_cycle_id)
--- a/app/jobs/order_cycle_notification_job.rb
+++ b/app/jobs/order_cycle_notification_job.rb
@@ -1,6 +1,9 @@
+# Deliers an email with a report of the order cycle to each of its suppliers
 OrderCycleNotificationJob = Struct.new(:order_cycle_id) do
   def perform
-    order_cycle = OrderCycle.find order_cycle_id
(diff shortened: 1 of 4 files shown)


app/mailers/producer_mailer.rb
```diff
@@ -12,12 +12,14 @@ def order_cycle_report(producer, order_cycle)
 
     subject = "[#{Spree::Config.site_name}] #{I18n.t('producer_mailer.order_cycle.subject', producer: producer.name)}"
 
-    if has_orders? order_cycle, producer
-      mail(to: @producer.email,
-           from: from_address,
-           subject: subject,
-           reply_to: @coordinator.email,
-           cc: @coordinator.email)
+    if has_orders?(order_cycle, producer)
```
sauloperez: I think it's a bit too late to check whether the order cycle has orders. It's pointless to send a notification if there are none. IMO we should check before enqueuing the `OrderCycleNotificationJob`. I'll create an issue for it.

app/jobs/order_cycle_notification_job.rb
```diff
@@ -1,6 +1,9 @@
+# Deliers an email with a report of the order cycle to each of its suppliers
```
enricostano: `s/Deliers/Delivers/`

app/mailers/producer_mailer.rb
sauloperez: done

daniellemoorhead:
Hey there lovely Katuma people. 

Would suggest staging this and checking that it does indeed continue to work as expected. If it doesn't break anything then go ahead and merge it (don't forget test notes!).

#myfirstgifof2018
giphy

kirstenalarsen:
made me laugh out loud

sauloperez:
The `OrderCycleNotificationJob` is successfully triggered in both staging.katuma.org and development. Working!

Issue #1999, opened 2017-12-19
Translations missing on "product edit" page


myriamboure:
## Description
Hub managers don't all understand English, and are confused when they see English strings (here only in back office so doesn't affect customers)

## Expected Behavior
All text is translated in local language

## Actual Behavior
Some strings are not translated

## Steps to Reproduce
1. In menu "products", clic on the "edit" button of any product

## Gif
![screenshot from 2017-12-20 07-08-15](https://user-images.githubusercontent.com/8135376/34176671-ee84dcd2-e554-11e7-8ddb-c1202a2722bd.png)

## Severity
4

## Your Environment
* Version used: 1.10.1
* Browser name and version: Chrome 53.0.2785.116 (64-bit)
* Operating System and version (desktop or mobile): Ubuntu 16.04 LTS

PR #2059, opened 2018-01-29
translation missing on product edit page

mjal:
#### What? Why?

Partially address #1999 (The translation for "None" in the select box is not covered)

This adds i18n keys for the left menu in /admin/products/*/edit
This only adds translation in en.yml and other should be added via transifex (i'm not sure about that).

#### What should we test?

The english version remains unchanged.
After translation are provided on transifex, in in /admin/products/carrots/edit, the left menu and the "back to products list" button are translated.

### DIFF
--- a/app/overrides/spree/admin/products/edit/return_to_bulk_product_edit.html.haml.deface
+++ b/app/overrides/spree/admin/products/edit/return_to_bulk_product_edit.html.haml.deface
@@ -1,3 +1,3 @@
 / replace "code[erb-loud]:contains('button_link_to t(:back_to_products_list)')"
 
-= button_link_to t(:back_to_products_list), bulk_edit_admin_products_path, :icon => 'icon-arrow-left'
+= button_link_to t('admin.products.back_to_products_list'), bulk_edit_admin_products_path, :icon => 'icon-arrow-left'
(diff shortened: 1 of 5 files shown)


myriamboure:
@sauloperez @mkllnk can one of you quickly check if @mjal way to handle missing translations is ok? He is ready to do more :-) Thanks!

sauloperez:
Thanks a lot for your contribution @mjal !

myriamboure:
How can I test a missing translation PR ? I guess it needs to be merged and translated in Transifex before it can be tested? Thanks for your tips ;-)

mkllnk:
> How can I test a missing translation PR ?

There are two things to test, but we can do only the first at this stage.

1. Check that the change didn't break anything. With the default English locale, we have to check that the text that got replaced by translatable text is still the same. Look for typos and missing text.
2. Checking translations into another language is not possible at this stage, because the translations are missing. But if you tested the first thing, then the code is not broken. The worst case is that we missed something and there are still some untranslatable text bits.

I don't have an easy solution for not being able to test the second case. But since these untranslatable text bits are finite, we don't need a long-term solution.

mkllnk:
https://staging1.openfood.com.au/
https://buildkite.com/open-food-foundation/open-food-network-staging1/builds/1756

sstead:
Testing notes

- English view looks the same, as intended. Good to go.

https://docs.google.com/document/d/1QgmZJInKdKOot4E8r2rYAkFSYLTQJMkRpZzUzXZxThw/edit?usp=sharing

PR #2069, opened 2018-02-06
Loads all order cycles for current user whether he is producer, distributor or coordinator

### LINKED ISSUE #1889: Bug in "order cycles" filter on customer admin reports
## Description
As a producer, I can't filter my sales by order cycle for the hubs I provide, so I can't get easily the information on what I'm suppose to deliver to a hub for a given order cycle.

## Expected Behavior
As a producer, when I generate a report, I should be able to filter the results by order cycle for all the order cycle I am a supplier for.

## Actual Behavior
As a producer, when I generate a report, I can only filter the results by order cycle for the order cycle I created.

## Steps to Reproduce
1- Connect to admin section with a **producer** account, make sure you create at least one product for the producer.
2- In another brower, connect with a hub account that has the permission to add to order cycle the producer's products (permissions/entreprise relationships, in entreprise menu), and open an order cycle with the producer's products.
3- With the producer account, go to "reports", "order cycle reports", and select for instance "order cycle total by producer".
4- Try to filter by order cycle, the hub's order cycle doesn't appear and can't be selected.

## Animated Gif/Screenshot

1- By default, I see the total quantity ordered for my products from the start date automatically selected, this seems to work ok.
screenshot from 2017-09-29 15-55-40

2- If I select a specific hub "hub demo open food france" here, I see the total corresponding to this hub this seems to work ok as well
screenshot from 2017-09-29 15-56-00

3- **BUT** If I try to select an order cycle to filter a bit more, I am proposed only the producer shop OC, I can't see the hub OC, so I can't filter by hub OC. We need to fix that so that the producer can filter by order cycle to know in real time what in needs to prepare for a given distribution so that he can anticipate a bit before the order cycle is closed.
screenshot from 2017-09-29 15-56-07

## Context
Various producers complained when trying to get the information about what they were supposed to deliver to the hub they are supplying (a buying group in this case).

## Severity
I put s2 as there is no workaround and it's pretty annoying, it's already hard for many producers to manage themselves their sales... let's not make it harder.

## Your Environment
* Version used: 1.10.1

## Comment
If I select a hub first, only the OC connected to that hub should be proposed.
If I select an OC first, I shouldn't be able then to add a filter on the hub.
This is to avoid conflicts.

HugsDaniel:
#### What? Why?

Closes #1889 

Added loading of all active or complete order cycles concerned by the current user to the reports form collection. User will see a collection of all order cycles he is concerned by as producer, distributor or coordinator.

#### What should we test?

The collection of order cycles loaded in the reports form contains all order cycles in which the current user has a role, and there are no duplicates.

### DIFF
--- a/app/controllers/spree/admin/reports_controller_decorator.rb
+++ b/app/controllers/spree/admin/reports_controller_decorator.rb
@@ -338,10 +338,9 @@ Spree::Admin::ReportsController.class_eval do
 
     # Avoids duplicates
     OrderCycle.active_or_complete.involving_managed_producers_of(spree_current_user).each do |order_cycle|
-        @order_cycles << order_cycle unless @order_cycles.include?(order_cycle)
+      @order_cycles << order_cycle unless @order_cycles.include?(order_cycle)
     end
 
-    @order_cycles.sort_by! {|order_cycle| order_cycle.orders_close_at}
-    return @order_cycles
+    @order_cycles.sort_by!(&:orders_close_at)
   end
 end
--- a/app/controllers/spree/admin/reports_controller_decorator.rb
+++ b/app/controllers/spree/admin/reports_controller_decorator.rb
@@ -208,7 +208,7 @@ Spree::Admin::ReportsController.class_eval do
     # My suppliers and any suppliers supplying products I distribute
     @suppliers = permissions.visible_enterprises_for_order_reports.is_primary_producer
 
-    @order_cycles = load_order_cycles
+    @order_cycles = OrderCycle.active_or_complete.accessible_by(spree_current_user).order('orders_close_at DESC')
 
     @report_types = report_types[:orders_and_fulfillment]
     @report_type = params[:report_type]
@@ -330,17 +330,5 @@ Spree::Admin::ReportsController.class_eval do
 
   def timestamp
     Time.zone.now.strftime("%Y%m%d")
-  end
-
-  # Loads all order cycles concerned by current user, may he be producer, distributor or coordinator
-  def load_order_cycles
-    @order_cycles = OrderCycle.active_or_complete.involving_managed_distributors_of(spree_current_user)
-
-    # Avoids duplicates
-    OrderCycle.active_or_complete.involving_managed_producers_of(spree_current_user).each do |order_cycle|
-      @order_cycles << order_cycle unless @order_cycles.include?(order_cycle)
-    end
-
-    @order_cycles.sort_by!(&:orders_close_at)
   end
 end
--- a/app/controllers/spree/admin/reports_controller_decorator.rb
+++ b/app/controllers/spree/admin/reports_controller_decorator.rb
@@ -208,8 +208,7 @@ Spree::Admin::ReportsController.class_eval do
     # My suppliers and any suppliers supplying products I distribute
     @suppliers = permissions.visible_enterprises_for_order_reports.is_primary_producer
 
-    @order_cycles = OrderCycle.active_or_complete.
-      involving_managed_distributors_of(spree_current_user).order('orders_close_at DESC')
+    @order_cycles = load_order_cycles
 
     @report_types = report_types[:orders_and_fulfillment]
     @report_type = params[:report_type]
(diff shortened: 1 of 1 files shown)


HugsDaniel:
I was able to display all the order cycles for the logged user. It seemed to be a choice of the team to only display the order cycles in which the user is a **distributor**, so the test won't pass until rewritten.

HugsDaniel:
The dynamic filtering doesn't seem to be implemented yet though. This was suggested in the issue : 
"If I select a hub first, only the OC connected to that hub should be proposed.
If I select an OC first, I shouldn't be able then to add a filter on the hub.
This is to avoid conflicts."
Could it be done via Angular ?

app/controllers/spree/admin/reports_controller_decorator.rb
```diff
@@ -332,4 +331,16 @@ def authorized_reports
...
+  def load_order_cycles
+    @order_cycles = OrderCycle.active_or_complete.involving_managed_distributors_of(spree_current_user)
+
+    # Avoids duplicates
+    OrderCycle.active_or_complete.involving_managed_producers_of(spree_current_user).each do |order_cycle|
+      @order_cycles << order_cycle unless @order_cycles.include?(order_cycle)
+    end
+
+    @order_cycles.sort_by!(&:orders_close_at)
+  end
```
sauloperez: we should solve this as a single query. As it is, this is much less performant and harder to maintaing than asking the DB to cook the data for us (that's what the DB is for). That fact that we're removing duplicates of a result set in Ruby is a smell.

Unless there's something I haven't foreseen, we should add a new scope in the `OrderCycle` model that fetches the order cycles concerned by the current user. "Concerned" meaning that the user is allowed to see the order cycle.

I'm pretty sure @mkllnk @Matt-Yorkley or @oeoeaio can point us to where this "concerned" logic is already implemented. I guess though, that this is essentially the order cycles of the enterprises that the user has either an enterprise relationship or an enterprise role with.

sauloperez:
To me the dynamic filtering it'd fall out of the scope of the bug fix. If it's not implemented it's a new feature and so it has to be addressed in another issue and PR. thoughts @myriamboure ?

app/controllers/spree/admin/reports_controller_decorator.rb
HugsDaniel: Good point. I just saw that now, there is actually a method returning all order cycles concerned by our user. I change that right now.

HugsDaniel: It's indeed way cleaner this way !

myriamboure:
> To me the dynamic filtering it'd fall out of the scope of the bug fix. If it's not implemented it's a new feature and so it has to be addressed in another issue and PR. thoughts @myriamboure ?
Yep @sauloperez let's fix the access to OC problem and then see if it's enough for now, I guess if there is a conflict it will just send 0 answer, should be fine.
Can it goes to test-ready or do you think another dev should review and approve? Doesn't seem so complex, right?
Awesome @HugsDaniel !

sauloperez:
Yes @myriamboure . Not a big deal.

sauloperez:
Deployed in https://staging.katuma.org/. Go go go!

myriamboure:
I free it up for Aus people to stage it as Katuma staging didn't work :-( So putting it back to dev ready and removing your comment @sauloperez feel free to repost if you happen to fix that before tonight, else let's the Aus stage and test :-)

pacodelaluna:
@myriamboure I have just deployed this PR on https://staging.openfoodfrance.fr/. Please let me know when it is validated or not.

myriamboure:
It seems it is taged on French staging for testing: https://staging.openfoodfrance.fr/ and will be tested today (I'll take care of it).

myriamboure:
Yepppa! It all looks perfect to me, testing notes here: https://docs.google.com/document/d/1bnWr7lyshQyF9CkkrmPiOGOljb31VA_0kwg6UV3OXn8/edit?usp=sharing

I tested lots of cases, also checked permissions potential impacts.
Ready to go for anyone who can merge that :-)

PR #2073, opened 2018-02-07
Disable guest checkout for registered users

### LINKED ISSUE #1159: Email confirmation | Prevent guest checkout using registered email addresses
**Who:** Any user who has signed up on the OFN
**Trigger:** They attempt to checkout as a guest instead of logging in.
**Goal:** To force a user to login if they try to checkout as a guest with an email that is already a registered (confirmed, or not). This can prevent identity theft, duplicate orders and typos.

Note: This issue is connected to #1587 and #1617, where users who sign up/login via the checkout process must confirm their email addresses. Also, this issue will resolve #907.

-------
**Registered user attempts guest checkout**

1. The user is at the checkout page and types in the already registered email:

image

2. When they go to click next the login modal appears over the page, saying the email address for a registered user so they need to log in or use a different email address. The email address typed into the field is displayed in the login email input field.
image

a) The user successfully logs in with the original email. The modal closes and they continue with checkout. 

b) The user closes the modal. They remain on the 'Your Details' section of the checkout. If they click Next, the same login modal opens again. If they change the email to a non-registered email, they can checkout as a guest.

Matt-Yorkley:
#### What? Why?

Closes #1159. 

Stops registered users using the guest checkout option.

#### What should we test?

Checkout as a guest using an email address of a registered user.

#### Release notes

Registered users will not be able to use the guest checkout from now on.

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
@@ -1,4 +1,4 @@
-Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, CurrentUser, CurrentHub) ->
+Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, CurrentUser, CurrentHub, AuthenticationService, SpreeUser, $http) ->
   $scope.Checkout = Checkout
   $scope.submitted = false
 
@@ -21,7 +21,26 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
   $scope.purchase = (event, form) ->
     event.preventDefault()
     $scope.submitted = true
+
+    if CurrentUser.id
+      $scope.validateForm(form)
+    else
+      $scope.confirmGuest(true)
+
+  $scope.validateForm = (form) ->
     if form.$valid
       $scope.Checkout.purchase()
     else
       $scope.$broadcast 'purchaseFormInvalid', form
+
+  $scope.confirmGuest = ->
+    $http.post("/user/registered_email", {email: $scope.order.email}).success (data)->
+      if data.registered == true
+        $scope.promptLogin()
+      else
+        $scope.validateForm() if $scope.submitted
+
+  $scope.promptLogin = ->
+    SpreeUser.spree_user.email = $scope.order.email
+    AuthenticationService.pushMessage t('devise.failure.already_registered')
+    AuthenticationService.open '/login'
--- a/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
@@ -20,6 +20,7 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
 
   $scope.purchase = (event, form) ->
     event.preventDefault()
+    $scope.formdata = form
     $scope.submitted = true
 
     if CurrentUser.id
@@ -27,18 +28,19 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
     else
       $scope.ensureUserIsGuest()
 
-  $scope.validateForm = (form) ->
-    if form.$valid
+  $scope.validateForm = ->
+    if $scope.formdata.$valid
       $scope.Checkout.purchase()
     else
-      $scope.$broadcast 'purchaseFormInvalid', form
+      $scope.$broadcast 'purchaseFormInvalid', $scope.formdata
 
-  $scope.ensureUserIsGuest = ->
+  $scope.ensureUserIsGuest = (callback = null) ->
     $http.post("/user/registered_email", {email: $scope.order.email}).success (data)->
       if data.registered == true
         $scope.promptLogin()
       else
         $scope.validateForm() if $scope.submitted
+        callback() if callback
 
   $scope.promptLogin = ->
     SpreeUser.spree_user.email = $scope.order.email
--- a/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
@@ -25,7 +25,7 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
     if CurrentUser.id
       $scope.validateForm(form)
     else
-      $scope.confirmGuest(true)
+      $scope.ensureUserIsGuest()
 
   $scope.validateForm = (form) ->
     if form.$valid
@@ -33,7 +33,7 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
     else
       $scope.$broadcast 'purchaseFormInvalid', form
 
-  $scope.confirmGuest = ->
+  $scope.ensureUserIsGuest = ->
     $http.post("/user/registered_email", {email: $scope.order.email}).success (data)->
       if data.registered == true
         $scope.promptLogin()
--- a/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
@@ -1,3 +1,8 @@
+# This class deals with displaying things in the login modal. It chooses
+# the modal tab templates and deals with switching tabs and passing data
+# between the tabs. It has direct access to the instance of the login modal,
+# and provides that access to other controllers as a service.
+
 Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redirections, Loading)->
 
   new class AuthenticationService
--- a/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
@@ -2,6 +2,7 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
 
   new class AuthenticationService
     selectedPath: "/login"
+    modalMessage: null
 
     constructor: ->
       if $location.path() in ["/login", "/signup", "/forgot"] || location.pathname is '/register/auth'
@@ -32,6 +33,8 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
         'registration_authentication.html'
       else
         'authentication.html'
+    pushMessage: (message) ->
+      @modalMessage = String(message)
 
     select: (path)=>
       @selectedPath = path
--- a/app/controllers/spree/users_controller_decorator.rb
+++ b/app/controllers/spree/users_controller_decorator.rb
@@ -15,4 +15,10 @@ Spree::UsersController.class_eval do
 
     @orders = @orders.where('distributor_id != ?', Spree::Config.accounts_distributor_id)
   end
+
+  # Endpoint for queries to check if a user is already registered
+  def registered_email
+    user = Spree.user_class.find_by_email params[:email]
+    render json: { registered: user.present? }
+  end
 end
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -17,6 +17,7 @@ Spree::Order.class_eval do
 
   validates :customer, presence: true, if: :require_customer?
   validate :products_available_from_new_distribution, :if => lambda { distributor_id_changed? || order_cycle_id_changed? }
+  validate :disallow_guest_order, if: lambda { using_guest_checkout? && registered_email? }
   attr_accessible :order_cycle_id, :distributor_id, :customer_id
 
   before_validation :shipping_address_from_distributor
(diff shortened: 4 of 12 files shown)


app/controllers/spree/users_controller_decorator.rb
```diff
@@ -15,4 +15,10 @@ def show
 
     @orders = @orders.where('distributor_id != ?', Spree::Config.accounts_distributor_id)
   end
+
+  # Endpoint for queries to check if a user is already registered
+  def registered_email
+    user = Spree.user_class.find_by_email params[:email]
+    render json: { registered: user.present? }
+  end
```
sauloperez: I'm missing at least a controller test for this.

daniellemoorhead:
@myriamboure @RachL @Matt-Yorkley is this the next PR to be completed for email confirm?

Matt-Yorkley:
@sauloperez I've added a controller spec here as well.

cat-gif

sauloperez:
I'm assigning them myself to take a look tomorrow

Gemfile
```diff
@@ -64,6 +64,7 @@ gem 'blockenspiel'
 gem 'acts-as-taggable-on', '~> 3.4'
 gem 'paper_trail', '~> 3.0.8'
 gem 'diffy'
+gem 'skylight', '< 2.0'
```
mkllnk: I think this commit is in the wrong pull request.

spec/features/consumer/shopping/checkout_auth_spec.rb
```diff
@@ -7,54 +7,76 @@
...
+      add_product_to_cart order, product
     end
-  end
 
-  it "renders the login buttons when logged out" do
-    visit checkout_path
-    within "section[role='main']" do
-      page.should have_content "Login"
-      click_button "Login"
+    it "does not not render the login form when logged in" do
```
sauloperez: one "not" is enough :trollface:

spec/features/consumer/shopping/checkout_auth_spec.rb
```diff
@@ -7,54 +7,76 @@
...
-    page.should have_field 'First Name', with: 'Foo'
-    page.should have_field 'Last Name', with: 'Bar'
-  end
+    it "renders the login buttons when logged out" do
+      visit checkout_path
+      within "section[role='main']" do
+        page.should have_content "Login"
+        click_button "Login"
+      end
+      page.should have_login_modal
```
sauloperez: please, use `expect` syntax

app/controllers/spree/users_controller_decorator.rb
Matt-Yorkley: Spec added.

Matt-Yorkley:
@mkllnk It performs the check within the angular checkout controller, before the order data is submitted. The order can't be submitted successfully as a guest using a registered email. I can add a fallback on the model as well if you like?

mkllnk:
@Matt-Yorkley Sorry for my vague reply. I meant that you can circumvent all the UI and Angular and send a POST request to server to create the order. This sounds like someone putting some effort into "hacking" the OFN. While this is unlikely, we have to consider "malicious" users. Additionally, these kind of requests are not necessarily intentional. They can also happen accidentally, because the UI has a bug or retrieved some cached data or we run into race conditions. To have a robust system and really prevent orders being made with a registered email address by an unauthorised user, we have to check for that in the Rails backend. Or do we do that already?

Matt-Yorkley:
@mkllnk I've added a check in the model and an extra model spec.

app/models/spree/order_decorator.rb
```diff
@@ -17,6 +17,7 @@
 
   validates :customer, presence: true, if: :require_customer?
   validate :products_available_from_new_distribution, :if => lambda { distributor_id_changed? || order_cycle_id_changed? }
+  validate :disallow_guest_order, if: lambda { using_guest_checkout? && registered_email? }
```
mkllnk: I would move the logic into `disallow_guest_order`. Much easier to read and everything is in one place. Still nice to keep the separate methods below though.

app/models/spree/order_decorator.rb
```diff
@@ -82,6 +83,18 @@ def products_available_from_new_distribution
     errors.add(:base, I18n.t(:spree_order_availability_error)) unless DistributionChangeValidator.new(self).can_change_to_distribution?(distributor, order_cycle)
   end
 
+  def using_guest_checkout?
+    require_email && !user.andand.id
```
mkllnk: I had to look into the Spree code to find `require_email`. At first I thought it might be something Active Record provides. It's good to use that method here, but we need to keep that in mind during the Spree upgrade. It may change.

mkllnk:
@sauloperez Are you happy with the new version of this pull request?

app/models/spree/order_decorator.rb
sauloperez: Can you do a quick check to see if that's the case @mkllnk 🙏 ?

app/models/spree/order_decorator.rb
```diff
@@ -82,6 +83,20 @@ def products_available_from_new_distribution
...
+  end
+
+  def registered_email?
+    Spree.user_class.find_by_email(email).present?
```
sauloperez: [exists?](https://apidock.com/rails/v3.2.13/ActiveRecord/FinderMethods/exists%3F) is a bit more lightweight and exactly meant for things like this.

app/models/spree/order_decorator.rb
mkllnk: If what is the case? So `require_email` is defined on Spree::Order to check if we are in a checkout step that requires an email. There are earlier steps that don't require an email, for example when you add stuff to your cart, an order is created and doesn't need an email.

app/models/spree/order_decorator.rb
mkllnk: Good point, I will quickly change that.

app/models/spree/order_decorator.rb
sauloperez: I meant if you could check the "it may change" part so that we know in advance, sorry

mkllnk: Ah, the code changes slightly, but should be the same meaning. Compared to spree master:
```diff
-  return true unless new_record? or state == 'cart'
+  true unless new_record? || ['cart', 'address'].include?(state)
```

Same in Solidus master. All good. :-)

Matt-Yorkley: @mkllnk is the plan to move to Solidus?

sauloperez: There's a loooong way before we can even consider that decision :trollface:

mkllnk: We don't have a plan for that stage yet. I just thought it would be interesting to know if a code change like this could affect a future decision like that. Just something to keep in mind. I think we will do the decision which fork to use when we updated to the latest spree version before the Solidus fork.

myriamboure:
@Matt-Yorkley it seems there is a conflict on that branch :-o

myriamboure:
:point_up: @Matt-Yorkley ?

myriamboure:
@Matt-Yorkley :point_right: 
giphy-downsized 1
:-)

mkllnk:

myriamboure:
Testing notes: https://docs.google.com/document/d/1bd-3ccmHuyzWhWMj7DdTi4erQ4-6NIhcDdkJ9P_hke8/edit#

There are 3 remaining things to fix before we can merge (I prefer not to merge as two of them are pretty annoying):
1- Login modal is to high on the page, not fully displayed and not possible to close using the "top right cross".
**Expected behavior:** display login modal at same “heigh” as other login modals and make sure it’s possible to close it.
2- If I click outside of the modal it closes the modal and I can continue filling in checkout page. When I reach the bottom of the checkout page and click “place order” I got the same login modal telling me to login so it’s not that bad, I still can’t checkout. But not really good in UX, as we are letting the user filling in information that they will lose anyway…
**Expected behavior:** if click outside of login modal, modal should close BUT should then display the checkout page back to the state before the button “next” is hit, so basically the user shouldn’t be able to move to another section of the checkout page until he has either changed the email, either logged in.
3- If I try to buy as a guest with an unconfirmed user, I'm asked to login, resend confirmation link. BUT when I click on the new link I am sent back to main page.
**Expected behavior:** I should be sent back to checkout page.

Back to you then @Matt-Yorkley to fix those 3 things.

[side note: I noticed some issue with the "reset password" feature used with unconfirmed user. Opened a dedicated issue as it's not connected to that PR (I tested on French production and it's the same there)]

Matt-Yorkley:
@myriamboure I havn't changed the height of the login modal in this PR. Have you noticed that before using Chrome?

Matt-Yorkley:
@myriamboure does the same thing happen for you on other pages with the login modal when one of the form errors is displayed?

Matt-Yorkley:
It should be fixed now. After a bit of investigating it turned out the script that opens checkout tabs was scrolling the page down after the login modal had opened.

myriamboure:
Is everything fixed or only point 1? Can we move on the pipe again?

Matt-Yorkley:
The first 2 should be fixed, as well as the missing translation when resetting the password from your notes. Regarding point 3: I think it redirects back to the page you were on when you first signed up, so resending the verification email from the checkout might not change it.

myriamboure:
Hum... point 3 is a bit annoying in term of UX as a user I was buying, I log in and when done I'm sent back to homepage.  Can't the validation link redirect back to the page you were on when you **last** requested a validation link (= clicked on resend validation link)? Just wondering... @daniellemoorhead @sstead that would make more sense no? What do you think?

myriamboure:
:point_up: @Matt-Yorkley is that possible then: "Can't the validation link redirect back to the page you were on when you last requested a validation link (= clicked on resend validation link)? " cf last comments.

myriamboure:
@Matt-Yorkley it's been quite a  long time since that hasn't move. Can you answer the point 3 last comment please? We have two options here:
- either we move forward with what you already fixed and I open a new issue to fix the redirection of the confirmation link to " the page you were on when you last requested a validation link"
- either you fix it in this PR as it has been reported here
What do you prefer?

Matt-Yorkley:
I've made issue #2237 to look at that last point, so this PR should be ready to go.

myriamboure:
Awesome @Matt-Yorkley ! @sauloperez @HugsDaniel if you are around can you have a look at Matt's last fixes and approve if all right? We can stage this afternoon then and test :-)

app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
```diff
@@ -20,8 +20,28 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
...
   $scope.purchase = (event, form) ->
     event.preventDefault()
+    $scope.formdata = form
     $scope.submitted = true
-    if form.$valid
+
+    if CurrentUser.id
+      $scope.validateForm()
+    else
+      $scope.confirmGuest(true)
```
sauloperez: I don't see this argument defined in the method

app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
```diff
@@ -20,8 +20,28 @@ Darkswarm.controller "CheckoutCtrl", ($scope, localStorageService, Checkout, Cur
...
+
+  $scope.validateForm = ->
+    if $scope.formdata.$valid
       $scope.Checkout.purchase()
     else
-      $scope.$broadcast 'purchaseFormInvalid', form
+      $scope.$broadcast 'purchaseFormInvalid', $scope.form
+
+  $scope.confirmGuest = ->
```
sauloperez: I don't find this method name particularly meaningful since it does nothing to the user. Would it perhaps be better if we just checked by means the HTTP request whether the user exists?

app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
```diff
@@ -32,6 +33,8 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
         'registration_authentication.html'
       else
         'authentication.html'
+    pushMessage: (message) ->
+      @modalMessage = String(message)
```
sauloperez: it's a bit odd for me to violate the encapsulation of this class. I haven't dug deeper into what this service does but having to pass an error message to a service to be displayed it's a bit strange. It feels likes this service might too many responsibilities. Does it deal with the details of the authentication or simply encapsulates a modal or some other UI component :thinking: ?

Matt-Yorkley: In this case, the checkout controller needed to open the login modal on a specific tab (which this service allows), but also display a specific message related to the checkout process inside the modal at the same time. The message is passed from the checkout controller via this service, to the individual controller for the specific tab.

sauloperez: It's clearly not the scope of this PR but according to what you said this class has way too many responsibilities and we would better off with a better name for the class itself.

sstead:
**Testing notes**

- There's one issue - if a user is checking out as a guest with an unregistered email, the Next button after entering their basic details at checkout doesn't click. Also they get a login modal popping up just after their order is confirmed. @oeoeaio I think these should be fixed first- I think some users would get thrown by the 'next' button not working, they may not think to scroll down.

https://docs.google.com/document/d/1gHTMDbIhQOkVX4IwQv-cUr0mIRiQMs0eeu4MB8XYor8/edit?usp=sharing

oeoeaio:
Does that sound like an easy fix @Matt-Yorkley? Putting back to In Dev for now.

myriamboure:
:point_up: @Matt-Yorkley any chance you can fix that small thing? Thanks!

Matt-Yorkley:
@sstead Should be fixed now.

myriamboure:
Can anyone stage this so that we can test? @mkllnk @sauloperez @enricostano ?

myriamboure:
Sorry I misunderstood, there are things already staged in Aus we will  start with this first, will come back to you then.

sstead:
**Testing notes**

Outstanding issue:
- The login modal shows on the order confirmation page for users who have tried to checkout as a guest with a registered email, but then changed to a non-registered email and checked out. This isn't a massive issue, we can probably leave it?

All else working well

https://docs.google.com/document/d/1gHTMDbIhQOkVX4IwQv-cUr0mIRiQMs0eeu4MB8XYor8/edit?usp=sharing

sstead:
Issue for tiny remaining modal bug https://github.com/openfoodfoundation/openfoodnetwork/issues/2277

PR #1955, opened 2017-11-13
closes cart page shouldn't show 'admin & handling' if it is zero #424

jtruong2:
#### What? Why?

The cart page should hide the table row 'admin & handling' if adjustments equal '$0.00'

#### What should we test?
Test should include 'shows admin & handling row if adjustments exist' and 'hides admin & handling if adjustments do not exist'

#### Release notes

Admin & Handling fee is only displayed in the cart if it is non-zero

PR #2079, opened 2018-02-12
Cart page shouldnt show admin handling if it is zero

### LINKED ISSUE #424: cart page shouldn't show 'admin & handling' if it is zero
image

mjal:
#### What? Why?

Closes #424 

In /cart, there is always a row called "Admin & Handling". The purpose of this PR is to hide it if those fees are 0.
This is based on the work done previously by @jtruong2 and on PR #1955 

#### What should we test?

After adding some article in the cart, if the admin & handling fees are 0, then there is no line "Admin & Handling"

### DIFF
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -143,5 +143,26 @@ feature "full-page cart", js: true do
         expect(page).to have_content item2.variant.name
       end
     end
+
+    context "Admin & Handling" do
+      it "hides row if order includes no adjustments" do
+        add_product_to_cart order, product_fee, quantity: 1
+        visit spree.cart_path
+
+        expect(page).to have_selector('#cart-detail')
+        expect(page).to_not have_content('Admin & Handling')
+      end
+
+      it "shows row if order includes adjustments" do
+        coordinator_fee = create(:enterprise_fee, enterprise: order_cycle.coordinator, fee_type: 'admin', calculator: Spree::Calculator::FlatRate.new(preferred_amount: 1))
+        order_cycle.coordinator_fees << coordinator_fee
+
+        add_product_to_cart order, product_fee, quantity: 1
+        visit spree.cart_path
+
+        expect(page).to have_selector('#cart-detail')
+        expect(page).to have_content('Admin & Handling')
+      end
+    end
   end
 end
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -28,7 +28,7 @@ feature "full-page cart", js: true do
       end
     end
 
-    describe "fees" do
+    describe "percentage fees" do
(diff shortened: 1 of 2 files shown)


myriamboure:
Note to self: to test it we need to create entreprise fees type "2€ per order" that applies not to products but to an order. That's what triggers admin&handling fees. So to test = if such fee on the OC : appears, else doesn't.

myriamboure:
Anyone to quickly review this @sauloperez or others ? It's been here for a month with no review :-(

spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -143,5 +143,26 @@
...
+        coordinator_fee = create(:enterprise_fee, enterprise: order_cycle.coordinator, fee_type: 'admin', calculator: Spree::Calculator::FlatRate.new(preferred_amount: 1))
+        order_cycle.coordinator_fees << coordinator_fee
+
+        add_product_to_cart order, product_fee, quantity: 1
+        visit spree.cart_path
+
+        expect(page).to have_selector('#cart-detail')
+        expect(page).to have_content('Admin & Handling')
+      end
+    end
```
enricostano: You can use `context` and `before` blocks to reuse some code and make the specs more readable: 
```ruby
before do
  add_product_to_cart order, product_fee, quantity: 1
end

context 'when there are no fees' do
  it 'hides admin and handlings row' do
    visit spree.cart_path

    expect(page).to have_selector('#cart-detail')
    expect(page).to_not have_content('Admin & Handling')
  end
end

context 'when there are fees' do
  before do
    coordinator_fee = create(:enterprise_fee, enterprise: order_cycle.coordinator, fee_type: 'admin', calculator: Spree::Calculator::FlatRate.new(preferred_amount: 1))
    order_cycle.coordinator_fees << coordinator_fee
  end

  it 'shows admin and handlings row' do
    visit spree.cart_path

    expect(page).to have_selector('#cart-detail')
    expect(page).to have_content('Admin & Handling')
    # Maybe we can also check the amount since we're here
  end
end

myriamboure:
@mjal sorry it took us so long to get back to you, as you see we are setting up lots of new processes and sometimes we miss things :-( Do you think you can finish that work you started? That would be very valuable !

myriamboure:
@enricostano let's see if that's better for you?

myriamboure:
Or so this is working fine and is ready to go (testing notes: https://docs.google.com/document/d/1mcyYRgS_xooQZyfO5l0Iekn72et1ep2ax5Ei535K0Bk/edit) BUT while testing I realized that "flexible rate" and "price sack" calculators are not working. This is not connected to that PR (I test on French production and error happening there as well) so this PR can be merged and I open straight away two issues for those bugs.
This is ready to go then.

myriamboure:
Ok I created issues https://github.com/openfoodfoundation/openfoodnetwork/issues/2172 and https://github.com/openfoodfoundation/openfoodnetwork/issues/2173 for the fee calculation issues I noticed while testing this, but nothing to do with that PR so let's merge!

PR #2083, opened 2018-02-15
Email/confirmation links

### LINKED ISSUE #1586: Email confirmation | User sign up (modal)
**Who:** Any user wanting to register on the OFN.
**How they access the sign up process:** via the sign up tab which is accessible via the 'Login' link in the top banner (opens login modal).

-------
**Step 1(a): User signs up**
The new user accesses the sign up tab and submits their email and password.
After submitting the modal will display this text:

image

And the following conditions are in place:
-  User remains logged out (new functionality). 
-  They can close the modal and display the original page underneath (existing functionality).
-  The green success banner does not show (existing functionality).

**Step 1(b): User resends confirmation email**
If the user needs the confirmation email to be resent, and clicks on the link in the model (screenshot above), this triggers:
1. The same confirm email to be sent.
2. A message to display below the 'resend email confirmation' saying 'We've resent the confirmation to email address, check your inbox'.

image

-------
**Step 2: System requests email confirmation**
The completion of step 1 will trigger the system to do 2 things:
1. System creates a user record, flagged as unconfirmed.
2. System sends a confirmation email to the user that includes a unique confirm link for that user.

image

-------
**Step 3: User confirms email**
The user clicking on the confirmation link in the email will trigger the following:
1. System flags the user record as confirmed.
2. System sends the user a 'Welcome' email.

image

3. Browser window opens and displays the page that they were on when they signed up, with the login tab open with a 'thanks for confirming message' displayed in the login tab. We need to think about the possibility that somebody tries to abuse this feature for unwanted redirects.

image

-------
**Step 4: User logs in**
Successful login from this screen will then close the modal and allow the user to continue browsing (existing functionality).

-------

EXCEPTIONS

- If the user attempts to login without clicking on the confirmation link in the email, they will not be able to login and will be directed to confirm (https://github.com/openfoodfoundation/openfoodnetwork/issues/1616)

- If the user tries to create an account using an email that is already signed up on OFN the existing functionality will apply: 
image

- If the user is a previous guest checkout, their customer record will be connected (https://github.com/openfoodfoundation/openfoodnetwork/issues/1592).

Matt-Yorkley:
#### What? Why?

Closes #1586 

Redirects user to previous url when clicking email confirmation link. Displays message in login modal to show user has been confirmed and can now log in.

#### What should we test?

User signup confirmation process.

### DIFF
--- a/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
@@ -4,13 +4,8 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
     selectedPath: "/login"
 
     constructor: ->
-      if $location.path() in ["/login", "/signup", "/forgot"] && location.pathname isnt '/register/auth'
-        @open $location.path()
-      else if location.pathname is '/register/auth'
-        if angular.isDefined($location.search()['validation'])
-          @open '/login', 'registration_authentication.html'
-        else
-          @open '/signup', 'registration_authentication.html'
+      if $location.path() in ["/login", "/signup", "/forgot"] || location.pathname is '/register/auth'
+        @open @initialTab(), @initialTemplate()
 
     open: (path = false, template = 'authentication.html') =>
       @modalInstance = $modal.open
@@ -20,6 +15,19 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
       @selectedPath = path || @selectedPath
       Navigation.navigate @selectedPath
 
+    initialTab: ->
+      if angular.isDefined($location.search()['validation'])
+        '/login'
+      else if location.pathname is '/register/auth'
+        '/signup'
+      else
+        $location.path()
+
+    initialTemplate: ->
+      if location.pathname is '/register/auth'
+        'registration_authentication.html'
+      else
+        'authentication.html'
 
     select: (path)=>
       @selectedPath = path
--- a/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
@@ -15,6 +15,9 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
       @selectedPath = path || @selectedPath
       Navigation.navigate @selectedPath
 
+    # Opens the /login tab if returning from email confirmation,
+    # the /signup tab if opened from the enterprise registration page,
+    # otherwise opens whichever tab is selected in the URL params ('/login', '/signup', or '/forgot')
     initialTab: ->
       if angular.isDefined($location.search()['validation'])
         '/login'
@@ -23,6 +26,7 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
       else
         $location.path()
 
+    # Loads the registration page modal when needed, otherwise the default modal
     initialTemplate: ->
       if location.pathname is '/register/auth'
         'registration_authentication.html'
--- a/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
@@ -7,7 +7,10 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
       if $location.path() in ["/login", "/signup", "/forgot"] && location.pathname isnt '/register/auth'
         @open $location.path()
       else if location.pathname is '/register/auth'
-        @open '/signup', 'registration_authentication.html'
+        if angular.isDefined($location.search()['validation'])
+          @open '/login', 'registration_authentication.html'
+        else
+          @open '/signup', 'registration_authentication.html'
 
     open: (path = false, template = 'authentication.html') =>
       @modalInstance = $modal.open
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -107,12 +109,6 @@ en:
         Invalid email or password.
         Were you a guest last time?  Perhaps you need to create an account or reset your password.
       unconfirmed: "You have to confirm your account before continuing."
-    enterprise_confirmations:
```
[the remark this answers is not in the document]
Matt-Yorkley: I just noticed these keys can all be removed, as they are for enterprise email confirmation messages that don't exist anymore in the codebase.

myriamboure:
@mkllnk this is also for you to review as part of the email confirm project :-)

app/controllers/user_confirmations_controller.rb
```diff
@@ -33,6 +33,12 @@ def show
       end
     end
 
-    respond_with_navigational(resource){ redirect_to login_path }
+    respond_with_navigational(resource){ redirect_to after_confirmation_path_for(resource) }
+  end
+
+  protected
+
+  def after_confirmation_path_for(resource)
```
mkllnk: `resource` seems to be unused here.

app/controllers/user_confirmations_controller.rb
```diff
@@ -25,20 +25,20 @@ def create
...
     respond_with_navigational(resource){ redirect_to after_confirmation_path_for(resource) }
   end
 
   protected
 
-  def after_confirmation_path_for(resource)
-    session[:confirmation_return_url] || login_path
+  def after_confirmation_path_for(_resource)
+    result =
+      if resource.errors.empty?
```
mkllnk: `resource` is now in use, but the variable above is declared as not used. I guess that's right if we just use the `resource` method of the controller. But why are we passing an unused variable then?

Matt-Yorkley: Yeah sorry, it should be `resource`, not `_resource`.

mkllnk:
It's back in staging1.

sstead:
**Testing Notes**

- If I sign up from **/register** or from the **cover page of a private shopfront** (where visitors are prompted to login or sign up)... when I confirm and am redirected I go to the correct place, BUT there’s no ‘thanks for confirming’ message and the Sign Up tab is open, not Login.

- Also just for consistency, can we change the wording from 'Sign In' to 'Login'

All the other redirects and messaging were correct.

https://docs.google.com/document/d/1wkEZzInsKFKeAHz9HT9ytxlS4HBDMvSJ2C_2mAk_ouk/edit#

myriamboure:
Back to you @Matt-Yorkley for those small last tweeks on this one :)

myriamboure:
@sauloperez @enricostano this is ready to review again :-) Email confirm on its way!

app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
```diff
@@ -17,6 +15,19 @@ Darkswarm.factory "AuthenticationService", (Navigation, $modal, $location, Redir
       @selectedPath = path || @selectedPath
       Navigation.navigate @selectedPath
 
+    initialTab: ->
```
enricostano: Some documentation would help understanding the code, thanks!

app/controllers/user_registrations_controller.rb
```diff
@@ -6,6 +6,7 @@ def create
     @user = build_resource(params[:spree_user])
     if resource.save
       session[:spree_user_signup] = true
+      session[:confirmation_return_url] = params[:return_url]
```
enricostano: I guess we can't pass this info as parameters in the URL right? I'm always nervous using session for this kind of things.

Matt-Yorkley: We can capture the return URL here, but then the user model later enqueues a job to create the email via `UserConfirmationsController`, but I think we can't directly pass variables to it at this point. Do you have any suggestions?

enricostano: Can't we pass URL parameters to `after_sign_in_path_for(@user)`?

myriamboure:
Any news on that one @Matt-Yorkley ? :-)

app/controllers/user_registrations_controller.rb
Matt-Yorkley: this instance of `after_sign_up_path_for` directs to a path when the user initially signs up, not when the email link is clicked. At this point we can capture the URL we want to redirect to, but we can't pass it directly to the model, as this will be triggered later, once the link is followed. It looks like Devise uses sessions by default to store these redirect URLs between events.

app/assets/javascripts/darkswarm/services/authentication_service.js.coffee
Matt-Yorkley: I refactored this a bit because the logic was getting complex. I've added some comments in the code now to clarify. These methods just choose which template will be used for the modal, and which tab will be selected when it opens (the layout and tab selection is different in /registration pages).

Matt-Yorkley:
@enricostano I've added code comments in the login modal js file as requested, and I think we do need to use a session variable to store the return URL (see above).

myriamboure:
@mkllnk only Australia have a staging where confirmation email work properly so it would be awesome if you could stage and test with Sally tomorrow (if you still both work on Wed.) Rachel will have tested the PR  currently on your staging by tonight so we will free your staging :-)

sstead:
**Testing notes**

- Those 2 cases have been fixed up ( from /register and from a private shop)
- It now reads Log In

This is ready to be pushed.

https://docs.google.com/document/d/144qeabLlxhw-vBwPs8eRWLzsyzW1-HrPt-We7v2JMqU/edit#

PR #2087, opened 2018-02-20
Subs: A dedicated review page for wizard

oeoeaio:
#### What? Why?
The wizard that is used by enterprise users to create new subscriptions has a `Review` step, that currently just shows all of the previous form elements on a single page, making for a fairly average UX. This PR replaces that view with a dedicated review view, making for a much nicer summary at the end of the wizard.

#### What should we test?

The subscription wizard feature should still make sense when creating and editing subscriptions. In particular the buttons that allow navigation backwards and forwards through the wizard should be logical and intuitive.

#### Release notes

This feature has not been released yet and so this will just be a part of the !!!SUBSCRIPTIONS!!! release notes.

### DIFF
--- a/app/assets/javascripts/admin/subscriptions/services/subscription_actions.js.coffee
+++ b/app/assets/javascripts/admin/subscriptions/services/subscription_actions.js.coffee
@@ -1,4 +1,7 @@
-angular.module("admin.subscriptions").factory 'SubscriptionPrototype', ($http, $injector, $q, InfoDialog, ConfirmDialog) ->
+# Wrapper for actions provided by ngResource
+# Used to extend the prototype of the subscription resource created by SubscriptionResource
+
+angular.module("admin.subscriptions").factory 'SubscriptionActions', ($http, $injector, $q, InfoDialog, ConfirmDialog) ->
   buildItem: (item) ->
     return false unless item.variant_id > 0
     return false unless item.quantity > 0
--- a/app/assets/javascripts/admin/subscriptions/services/subscription_resource.js.coffee
+++ b/app/assets/javascripts/admin/subscriptions/services/subscription_resource.js.coffee
@@ -1,4 +1,4 @@
-angular.module("admin.subscriptions").factory 'SubscriptionResource', ($resource, SubscriptionPrototype) ->
+angular.module("admin.subscriptions").factory 'SubscriptionResource', ($resource, SubscriptionActions) ->
   resource = $resource('/admin/subscriptions/:id/:action.json', {}, {
     'index':
       method: 'GET'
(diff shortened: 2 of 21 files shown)


app/assets/javascripts/admin/subscriptions/services/subscription_resource.js.coffee
```diff
@@ -27,5 +27,6 @@ angular.module("admin.subscriptions").factory 'SubscriptionResource', ($resource
   })
 
   angular.extend(resource.prototype, SubscriptionActions)
+  angular.extend(resource.prototype, SubscriptionFunctions)
```
mkllnk: So we put the functions into a different class, but then extend the resource prototype. Does that mean that *all* resources have these functions now? Why don't we declare these functions in the SubscriptionResource?

oeoeaio: Not quite my understanding but I could be wrong. As far as I understand it, the call to `$resource` above returns a `Resource` constructor, which is then used to build new `subscriptions` objects which have access the nice resource api elements ($create, $update, etc). I think I can safely extend the prototype of just this constructor independently of other resource constructors, because constructors are not singletons or global in scope? Here I am extending the prototype of the constructor, to give all the `subscriptions` I instantiate access to the functions defined by `SubscriptionActions` and `SubscriptionFunctions`. I kind of made this up as I went along, so if there is a better way of doing it I am happy to use that instead.

oeoeaio: By doing it this way I am kind of defining the functions in the SubscriptionResource, but I'm just including them like you would a module in ruby, instead of having them all jumbled into one factory.

app/assets/javascripts/admin/subscriptions/services/subscription_actions.js.coffee
```diff
@@ -1,4 +1,7 @@
-angular.module("admin.subscriptions").factory 'SubscriptionPrototype', ($http, $injector, $q, InfoDialog, ConfirmDialog) ->
+# Wrapper for actions provided by ngResource
+# Used to extend the extend the protype of the subscription resource created by SubscriptionResource
```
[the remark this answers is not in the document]
oeoeaio: Typo, will fix it, thanks!

sauloperez:
I'd great to have a screenshot for this cases to see how it will look like

oeoeaio:
@sauloperez 

screen shot 2018-02-22 at 9 21 07 am

sstead:
Testing notes
All good!

https://docs.google.com/document/d/1Jcj8uWHnXpwqvQsuZC2Tit5NWuMHAsoHGjXqP__P55s/edit?usp=sharing

Issue #1061, opened 2016-06-22
SO - View all my standing orders


daniellemoorhead:
**As a shop owner I want to view all my standing orders**

Go into admin section. 
Go to the orders section.
Click on the "Standing Orders" tab.
Result: Display the new standing orders page

Then:
- Filter by customer or by schedule (and possibly by item, but may be a bit tricky, we'll see).
- [x] Create Standing Order model (1 hr)
- [x] Build basic version of interface (11 hr)

sstead:
Yes- page exists and you can filter by coordinator.
Yes- Is angular - with save column defaults.
Still some things to be done...
![image](https://cloud.githubusercontent.com/assets/10843056/20738348/89571226-b706-11e6-99f0-4093e6cb53e6.png)

sstead:
![image](https://cloud.githubusercontent.com/assets/10843056/21171016/49650ebc-c21d-11e6-9c37-57b6acac773c.png)

sstead:
![image](https://cloud.githubusercontent.com/assets/10843056/21171511/99b9aa64-c220-11e6-9da2-d6373cca61a1.png)

sstead:
- Pricing are now showing in the Item column (fixed). They also calculate correctly.
- The 'closing in X days' message is now counting the correct number of days. (fixed)

Bugs:

Translations:
![image](https://user-images.githubusercontent.com/10843056/33110624-3cea8958-cf9d-11e7-82a7-dd9092e1df91.png)

This is outstanding:
![image](https://user-images.githubusercontent.com/10843056/33110641-5eb581b4-cf9d-11e7-9978-cc9130da432a.png)

PR #2089, opened 2018-02-20
Subs: Only count future orders in subs index order count

oeoeaio:
#### What? Why?

Resolves an issue picked up by @sstead and documented in #1061. The Subscriptions listing page shows the total number of orders associated with a subscription, rather than the number or future orders. This PR resolves that.

#### What should we test?

That the number of orders reported on the subscriptions listing page matches the number of visible orders in the tab when clicked.

#### Release notes


sstead:
Testing notes

The orders count is correctly corresponding to the orders shown below in the expando area.

However, some issues found:
- The number of orders count is sometimes including orders that fall outside the customer's subscription period. This occurs when I change and OC's dates.
- There is a text display error... 'closed a few seconds ago' shows for an OC that's not open yet.
- The value of an order is sometimes not including the shipping fee.

https://docs.google.com/document/d/14YCILqbeKGEo1jD4E8MBA8ORCILk8iJ59cukzMBYxMk/edit?usp=sharing

oeoeaio:
👍 Will document these as separate issues.

PR #2090, opened 2018-02-20
Subs: Make checklist more legible

oeoeaio:
#### What? Why?

There is a checklist that lists the steps that must be completed before users can create a subscription. The styling of this checklist was pretty scary, so this is just a bump in readability.

#### What should we test?

Visual check of the subscriptions checklist (visible by going to Orders > Subscriptions in the admin section). Note that the checklist will only be visible if one of the following is true for a particular enterprise: No customers, no shipping methods, no payment methods, no schedules containing coordinated OCs.

#### Release notes

### DIFF
--- a/app/assets/stylesheets/admin/components/todo.scss
+++ b/app/assets/stylesheets/admin/components/todo.scss
@@ -3,8 +3,20 @@
 .todolist{
   .todo {
     &.done {
-      .title, .steps {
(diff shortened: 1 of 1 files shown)


sauloperez:
can you share a screenshot?

oeoeaio:
@sauloperez 

subs-styling

PR #2092, opened 2018-02-20
Fix incorrect tax on Admin Order Show

### LINKED ISSUE #1941: VAT discrepency in order management view
**Bug description:**
I make an order, total amount inc. tax = 17,02€, VAT = 0,88€, excl. tax = 16,14€

These are good on confirmation and invoice:
screenshot from 2017-11-03 12-03-21
screenshot from 2017-11-03 12-02-54

BUT VAT is wrong on the admin "view order" page:
screenshot from 2017-11-03 12-03-09

I found the issue, on view order the VAT displayed is only calculated on the products and doesn't include the VAT on entreprise fees...

**Acceptance criteria:**
VAT amount displayed on "view order" is the same than the one in order confirmation and invoice.

ltrls:
#### What? Why?

Closes #1941 

The Admin Order Show action used Order#price_adjustment_totals to display the amount of taxes. As a Spree method it didn't show taxes on enterprise fees or additional adjustment. I made the Order#tax_adjustment_totals method (which we previously needed for the tax rate report and the invoice2 view) more generic and overrode Order#price_adjustment_totals to make use of it.
I didn't write tests for the new Order#price_adjustment_totals because it's really simple, but wrote more thorough tests for Order#tax_adjustment_totals and feature tests for Admin Order Show view.
I also fixed some old Rspec syntax in this page but I'd understand if you'd prefer it to be in a separate PR (?)

#### What should we test?

What Myriam showed in the issue.

#### Release notes

"Fix VAT decrepency in order management view" ?

### DIFF
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -276,10 +276,17 @@ Spree::Order.class_eval do
       tax_rates = adjustment.tax_rates
       tax_rates_hash = Hash[tax_rates.collect do |tax_rate|
         tax_amount = tax_rates.one? ? adjustment.included_tax : tax_rate.compute_tax(adjustment.amount)
-        [tax_rate.amount, tax_amount]
+        [tax_rate, tax_amount]
       end]
       hash.update(tax_rates_hash) { |_tax_rate, amount1, amount2| amount1 + amount2 }
     end
+  end
+
+  def price_adjustment_totals
+    Hash[tax_adjustment_totals.map do |tax_rate, tax_amount|
+      [tax_rate.name,
+       Spree::Money.new(tax_amount, currency: currency)]
+    end]
   end
 
   def has_taxes_included
--- a/lib/open_food_network/sales_tax_report.rb
+++ b/lib/open_food_network/sales_tax_report.rb
@@ -13,7 +13,7 @@ module OpenFoodNetwork
       when "tax_rates"
         [I18n.t(:report_header_order_number),
          I18n.t(:report_header_total_excl_vat, currency_symbol: currency_symbol)] +
-        relevant_rates.map { |rate| "%.1f%% (%s)" % [rate.to_f * 100, currency_symbol] } +
+        relevant_rates.map { |rate| "%.1f%% (%s)" % [rate.amount.to_f * 100, currency_symbol] } +
         [I18n.t(:report_header_total_tax, currency_symbol: currency_symbol),
          I18n.t(:report_header_total_incl_vat, currency_symbol: currency_symbol)]
        else
@@ -66,7 +66,7 @@ module OpenFoodNetwork
 
     def relevant_rates
       return @relevant_rates unless @relevant_rates.nil?
-      @relevant_rates = Spree::TaxRate.pluck(:amount).uniq
+      @relevant_rates = Spree::TaxRate.uniq
     end
 
     def totals_of(line_items)
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -215,7 +215,11 @@ feature %q{
         Spree::Config[:enable_receipt_printing?] = true
 
         distributor1.update_attribute(:abn, '12345678')
-        @order = create(:order_with_taxes, distributor: distributor1)
+        @order = create(:order_with_taxes,
+                        distributor: distributor1,
+                        product_price: 110,
+                        tax_rate_amount: 0.1,
+                        tax_rate_name: "Tax 1")
         Spree::TaxRate.adjust(@order)
 
         visit spree.admin_order_path(@order)
@@ -256,10 +260,8 @@ feature %q{
 
       scenario "shows the order taxes" do
         within('table.index tbody#price-adjustments') do
-          @order.price_adjustment_totals.each do |label, total|
-            expect(page).to have_selector "td", match: :first, text: label
-            expect(page).to have_selector "td.total", text: total
-          end
+          expect(page).to have_selector "td", match: :first, text: "Tax 1"
+          expect(page).to have_selector "td.total", text: Spree::Money.new(10)
         end
       end
 
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -1,4 +1,5 @@
 require "spec_helper"
+include ActionView::Helpers::NumberHelper
 
 feature %q{
     As an administrator
@@ -214,9 +215,52 @@ feature %q{
         Spree::Config[:enable_receipt_printing?] = true
 
         distributor1.update_attribute(:abn, '12345678')
-        @order = create(:completed_order_with_totals, distributor: distributor1)
+        @order = create(:order_with_taxes, distributor: distributor1)
+        Spree::TaxRate.adjust(@order)
 
         visit spree.admin_order_path(@order)
+      end
+
+      scenario "shows a list of line_items" do
+        within('table.index tbody', match: :first) do
+          @order.line_items.each do |item|
+            expect(page).to have_selector "td", match: :first, text: item.full_name
+            expect(page).to have_selector "td.price", text: item.single_display_amount
+            expect(page).to have_selector "td.qty", text: item.quantity
+            expect(page).to have_selector "td.total", text: item.display_amount
+          end
+        end
+      end
+
+      scenario "shows the order subtotal" do
+        within('table.index tbody#subtotal') do
+          expect(page).to have_selector "td.total", text: @order.display_item_total
+        end
+      end
+
+      scenario "shows the order charges (non-tax adjustments)" do
+        within('table.index tbody#order-charges') do
+          @order.adjustments.eligible.each do |adjustment|
+            next if (adjustment.originator_type == 'Spree::TaxRate') && (adjustment.amount == 0)
+            expect(page).to have_selector "td", match: :first, text: adjustment.label
+            expect(page).to have_selector "td.total", text: adjustment.display_amount
+          end
+        end
+      end
+
+      scenario "shows the order total" do
+        within('table.index tbody#order-total') do
+          expect(page).to have_selector "td.total", text: @order.display_total
+        end
+      end
+
+      scenario "shows the order taxes" do
+        within('table.index tbody#price-adjustments') do
+          @order.price_adjustment_totals.each do |label, total|
+            expect(page).to have_selector "td", match: :first, text: label
+            expect(page).to have_selector "td.total", text: total
+          end
+        end
       end
 
       scenario "shows the dropdown menu" do
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -308,15 +308,15 @@ describe Spree::Order do
     end
 
     it "contains tax on line_item" do
-      expect(order.tax_adjustment_totals[tax_rate10.amount]).to eq(4.0)
+      expect(order.tax_adjustment_totals[tax_rate10]).to eq(4.0)
     end
 
     it "contains tax on shipping_fee" do
-      expect(order.tax_adjustment_totals[tax_rate15.amount]).to eq(6.0)
+      expect(order.tax_adjustment_totals[tax_rate15]).to eq(6.0)
     end
 
     it "contains tax on enterprise_fee" do
-      expect(order.tax_adjustment_totals[tax_rate20.amount]).to eq(8.0)
+      expect(order.tax_adjustment_totals[tax_rate20]).to eq(8.0)
     end
   end
 
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -323,7 +323,7 @@ describe Spree::Order do
     end
 
     it "contains tax on order adjustment" do
-      order.tax_adjustment_totals[tax_rate25].should == 10.0
+      expect(order.tax_adjustment_totals[tax_rate25]).to eq(10.0)
     end
   end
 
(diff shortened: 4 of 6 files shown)


spec/models/spree/order_spec.rb
```diff
@@ -303,20 +306,24 @@
...
+      order.tax_adjustment_totals[tax_rate15].should == 6.0
     end
 
     it "contains tax on enterprise_fee" do
-      order.tax_adjustment_totals[tax_rate20.amount].should == 8.0
+      order.tax_adjustment_totals[tax_rate20].should == 8.0
+    end
+
+    it "contains tax on order adjustment" do
+      order.tax_adjustment_totals[tax_rate25].should == 10.0
```
sauloperez: use the `expect` syntax here as well, please

spec/features/admin/orders_spec.rb
```diff
@@ -201,11 +202,54 @@ def new_order_with_distribution(distributor, order_cycle)
...
+
+      scenario "shows the order total" do
+        within('table.index tbody#order-total') do
+          expect(page).to have_selector "td.total", text: @order.display_total
+        end
+      end
+
+      scenario "shows the order taxes" do
+        within('table.index tbody#price-adjustments') do
+          @order.price_adjustment_totals.each do |label, total|
```
sauloperez: I think we should rather calculate the appropriate totals without having to depend on `price_adjustment_totals`. If we happen to refactor this logic by removing this method, this spec should keep working.

app/models/spree/order_decorator.rb
```diff
@@ -276,12 +276,19 @@ def tax_adjustment_totals
...
       tax_rates_hash = Hash[tax_rates.collect do |tax_rate|
         tax_amount = tax_rates.one? ? adjustment.included_tax : tax_rate.compute_tax(adjustment.amount)
-        [tax_rate.amount, tax_amount]
+        [tax_rate, tax_amount]
       end]
       hash.update(tax_rates_hash) { |_tax_rate, amount1, amount2| amount1 + amount2 }
     end
   end
 
+  def price_adjustment_totals
```
sauloperez: So, is this method called from the view? I'm not fully grasp the fix.

sauloperez:
> I also fixed some old Rspec syntax in this page but I'd understand if you'd prefer it to be in a separate PR (?)

Although it is not a big deal as the changes can still be reviewed, I'd say yes. It's a good practice all of us should get used to, as discussed in the last international meeting.

spec/models/spree/order_spec.rb
ltrls: oh yes! pretty weird that I forgot it

spec/features/admin/orders_spec.rb
ltrls: It's a feature test, we test that the view uses the method, not how the method works, right? if we did that we would simulate the method by copying all of it... or do you mean that I should use `tax_adjustment_totals` instead?

app/models/spree/order_decorator.rb
ltrls: the method is called from the invoice2.html view and the tax_rate report. for those views returning a Hash of `tax_rate_amount (decimal) => tax_aggregate_amount (decimal)` was all that was needed, but the `price_adjustment_totals` method called from the admin order show view expects a Hash of `tax_name (string) => tax_aggregate_amount (Money)`, so I made this method return a Hash of `tax_rate (TaxRate) => tax_aggregate_amount (decimal)` so we can call `TaxRate#name` in `price_adjustment_totals` and `TaxRate#amount` from the other places
(we return the whole object and delegate the call to its attribute to places where the method is called)

ltrls:
> Although it is not a big deal as the changes can still be reviewed, I'd say yes. It's a good practice all of us should get used to, as discussed in the last international meeting. What do you think?

I don't know. In a way it's disturbing to "bundle" unrelated changes in a single PR, and as you said it could lead to more difficult reviews, on the other hand the most important thing (for rollback and cherry-picking of changes, maybe for reviewing too?) is that it's in separate commits, and these kinds of small cosmetic/refactoring fixes ("oh I just saw that while I was doing something else") might not be/feel worth the friction of you know, creating a new branch, submitting a pull request... I'm not sure there's an easy answer to this, maybe it should be assessed on a case-by-case basis?

spec/features/admin/orders_spec.rb
sauloperez: > we test that the view uses the method

This is what views tests are for. Feature tests are just the idiomatic way to call integration tests in Rails land, I believe. As such you're executing the whole tech stack, including the browser. That's why we should not depend on a particular method and assert the behavior instead.

sauloperez:
> might not be/feel worth the friction of you know, creating a new branch, submitting a pull request
I do things like this quite often and it takes 5min at most. As a result, that cosmetic PR can be shipped pretty fast, without having to wait for the bigger PR they stem from. The more you do it, the faster you become.

> maybe it should be assessed on a case-by-case basis

That's the thing I'd like to avoid. We should agree on a procedure and stick to it, and that is what we said last year we will do from now on.

When things like these are left open for everyone to decide, some PRs become wormholes and last years. That's why I believe we need to put an effort to get this habit, which is making things better already.

myriamboure:
@ltrls any chance you can finish that quickly? Thanks :)

ltrls:
here it is :)

myriamboure:
This is the fix @ltrls provided @sauloperez please check if you are happy with it https://github.com/openfoodfoundation/openfoodnetwork/pull/2114 :-)

lib/open_food_network/sales_tax_report.rb
```diff
@@ -66,7 +66,7 @@ def table
 
     def relevant_rates
       return @relevant_rates unless @relevant_rates.nil?
-      @relevant_rates = Spree::TaxRate.pluck(:amount).uniq
+      @relevant_rates = Spree::TaxRate.uniq
```
sauloperez: This is quite dangerous due to the query that generates:

```
irb(main):001:0> Spree::TaxRate.uniq.to_sql
=> "SELECT DISTINCT \"spree_tax_rates\".* FROM \"spree_tax_rates\" "
```

If you take a look at the explain:

```sql
openfoodnetwork=# EXPLAIN SELECT DISTINCT "spree_tax_rates".* FROM "spree_tax_rates";
                                                       QUERY PLAN
------------------------------------------------------------------------------------------------------------------------
 HashAggregate  (cost=14.23..15.53 rows=130 width=560)
   Group Key: id, amount, zone_id, tax_category_id, created_at, updated_at, included_in_price, name, show_rate_in_label
   ->  Seq Scan on spree_tax_rates  (cost=0.00..11.30 rows=130 width=560)
(3 rows)
```

It performs a sequential scan on the `spree_tax_rates` table, which can lead to 💥 (or serious performance hit). As I presume that table does not have too many rows I suggest we create an issue for it. The point is to limit the data we fetch from DB instead of blindly getting everything on that table

ltrls: Getting only the relevant tax rates was what I wanted to do in the first place, but it appeared that it was both costly and mostly useless.
Costly because the report gets the orders that are searched for and gets the corresponding tax rates and amounts via the previously discussed method (L47-50). To build the table we would need those orders/tax rates/amounts beforehand. But maybe we could memorize it (in `#search`) like we do in `#relevant_rates`?
By the way the use of `distinct` seems to lower downs the postgres "cost" and I guess it's useless as there wouldn't be any duplicate rows if there's no joint, right?
It also seemed mostly useless because the `spree_tax_rates` wouldn't be very large, typically there wouldn't be more 2 or 3 different taxes, don't you think? The sequential scan also seems pretty normal as we're doing a `SELECT *` so PG is browsing the whole table.
Happy to do what you think is best, I guess memorizing the search results to use it in `#header` and `#table` might be nice but only if Rails made the queries for `spree_tax_rates` only once for both times we would call `orders.tax_adjustment_totals` (or `orders.map(&:tax_adjustments).flatten.map(&:tax_rates).flatten.uniq` for what it's worth but that's not really pretty although we could create `Order#tax_rates` for this) ; but I think it would run the queries twice.

myriamboure:
@sauloperez I think it's your turn to answer @ltrls argument in fact ;-)

lib/open_food_network/sales_tax_report.rb
sauloperez: I was just asking to create an issue so that we can work on it some other day. To answer your questions:

> To build the table we would need those orders/tax rates/amounts beforehand. But maybe we could memorize it

Exactly, and we can fetch the ones we need as we do with the orders in `#search`.

> By the way the use of distinct seems to lower downs the postgres "cost" and I guess it's useless as there wouldn't be any duplicate rows if there's no joint, right?

If there were no duplicates then there would be no need for the call to `#uniq` in line 69. However, that's not enforced at DB level and so it could happen. In general terms, `DISTINCT` cuts down the cost because fewer rows need to be fetched from disk and sent over the network.

> It also seemed mostly useless because the spree_tax_rates wouldn't be very large, typically there wouldn't be more 2 or 3 different taxes, don't you think?

That's what I guess but nothing stops anyone from creating lots of them. Besides, I'm not sure if that is something created by the instance admin and shared among enterprises or if it's something that each enterprise manager creates. In the latter case, it'd much easier to end up having lots of taxes on DB table.

> The sequential scan also seems pretty normal as we're doing a SELECT * so PG is browsing the whole table.

What makes PG choose a particular [Query Plan](http://www.postgresguide.com/performance/explain.html) are the constraints specified in the query's `WHERE` clause. If you specify a constraint on a column that has an index, chances are that it'll pick it up instead of a sequential scan. That is, the scan will happen on the index structure rather than on the table, which is way faster.

What `SELECT` can help with, when selecting a particular column/s, is reducing the cost of fetching the actual data from disk, the cost of sending it over the network and the number of Ruby objects instantiated in memory (which normally has a big impact on performance).

> I guess memorizing the search results to use it in #header and #table might be nice but only if Rails made the queries for spree_tax_rates only once for both times we would call orders.tax_adjustment_totals (or orders.map(&:tax_adjustments).flatten.map(&:tax_rates).flatten.uniq for what it's worth but that's not really pretty although we could create Order#tax_rates for this) ; but I think it would run the queries twice.

If the query stays the same for both `#header` and `#table` we can just save its result in a var. In any case, I would discourage doing `orders.map(&:tax_adjustments).flatten.map(&:tax_rates).flatten.uniq` performance wise. That's exactly 5 times going through `orders`, while we can do the same leveraging Postgres performance and the declarative nature of SQL. We would just ask for the data we need without loops. 

Also, things like these have become a big problem in reports. `5 x 10` orders it's not a big deal, but when fetching 1000 orders, for instance, performance starts to be affected.

myriamboure:
Ouille, seems that there are branch conflicts :-( @ltrls can you rebase please so that we can test ?

lib/open_food_network/sales_tax_report.rb
ltrls: > I was just asking to create an issue so that we can work on it some other day.

Oh sorry I misunderstood.

> Besides, I'm not sure if that is something created by the instance admin and shared among enterprises or if it's something that each enterprise manager creates.

Yes they're created by the admin, and enterprise managers only choose which one applies.

> What makes PG choose a particular Query Plan are the constraints specified in the query's WHERE clause.

Yeah sorry I meant "SELECT without WHERE" ^.^

>  I would discourage doing orders.map(&:tax_adjustments).flatten.map(&:tax_rates).flatten.uniq performance wise. That's exactly 5 times going through orders, while we can do the same leveraging Postgres performance and the declarative nature of SQL.

Actually I just remembered that I had the same worry concerning fetching the tax rates and I wrote [something like what you say](https://github.com/openfoodfoundation/openfoodnetwork/blob/d22f5678bea35e8ef607f0ee7c3e30770190c695/lib/open_food_network/sales_tax_report.rb#L69-L71) (it made just one query and that was nice!).
... but I removed it afterwards because it didn't work with taxes applied on shipping fees and adjustments created from Spree::Admin::AdjustmentsController#create (see [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/spree/admin/adjustments_controller_decorator.rb#L31-L40)), because they don't "link" to the corresponding tax rate and only have the applied tax amount in their `included_tax` (that's why I had to write [this method](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/models/spree/adjustment_decorator.rb#L61-L68). I had no clue how to make this fit with the "relevant tax rates" query so I resigned to use `TaxRate.all` ><

ltrls:
> Ouille, seems that there are branch conflicts :-( @ltrls can you rebase please so that we can test ?

Voilà ! I also answered in our discussion with @sauloperez :)

myriamboure:
@sauloperez fine for you to move to test ready? Do it if you agree :-)

myriamboure:
I'm not able to do that, would be great if @ltrls you can create those issues for identified tech debt (I don't understand those tech discussions ;-)) But we can still move to test this :-)

pacodelaluna:
Deployed on https://staging.openfoodfrance.fr/

myriamboure:
Ok this seems fine to me, testing notes https://docs.google.com/document/d/1zJ3zXuZTUbKr_ooeWKZ-iKhTwb0ZbKJtaG0-wg0TDqk/edit#
I tried also to play a bit with adjustments to try to break it but seems to behave consistently. @sstead if you want to have a look as well as VAT can sometimes be a bit tricky I'm happy if you can have a look at testing notes and see if you want to test something else before we merge... so would love some second review on this. If ok for you I let you put it to "ready to go".

myriamboure:
I think we can move this to ready to go, didn't have feedback from @sstead but I don't really see how that could break anything given the tests I did, I was just a bit worried I think as VAT is always a bit tricky depending on countries, but this is only a fix to an amount display in back office and was not consistent with the other ones, tests show that now it is consistent.

Issue #1078, opened 2016-07-01
SO - Send an email to the customer when a new standing order opens


daniellemoorhead:

mkllnk:
- [ ] Create or extend mailer (1h)
- [ ] Create email templates in text and html (1h)
- [ ] List items that are unavailable (3h)
- [ ] Highlight changed prices (2h)

sstead:
@oeoeaio, this standing order isn't seeing overridden prices in Inventory, seeing product prices instead... ![image](https://cloud.githubusercontent.com/assets/10843056/21378132/1694365e-c798-11e6-9b64-714e8e88fca7.png)

sstead:
The above is still an issue.

PR #2093, opened 2018-02-21
Subs: take account of overrides

oeoeaio:
#### What? Why?

@sstead realised that subscriptions were not taking account of variant overrides when determining stock availability and price in some circumstances. See #1078 for more detail.

#### What should we test?

All line items in orders created from by a subscription should respect stock limits and prices of any variant overrides for the relevant shop. Places to check:
- [ ] The subscription wizard
- [ ] The edit order interface in the admin section
- [ ] The placement and confirmation emails for the order
- [ ] The customers account page

#### Release notes

### DIFF
--- a/app/models/proxy_order.rb
+++ b/app/models/proxy_order.rb
@@ -1,3 +1,7 @@
+# Each Subscription has many ProxyOrders, one for each OrderCycle to which the Subscription applies
+# Proxy pattern allows for deferral of initialization until absolutely required
+# This reduces the need to keep Orders in sync with their parent Subscriptions
+
 class ProxyOrder < ActiveRecord::Base
   belongs_to :order, class_name: 'Spree::Order', dependent: :destroy
   belongs_to :subscription
--- a/app/models/proxy_order.rb
+++ b/app/models/proxy_order.rb
@@ -42,21 +42,8 @@ class ProxyOrder < ActiveRecord::Base
 
   def initialise_order!
     return order if order.present?
-    create_order!(
-      customer_id: subscription.customer_id,
-      email: subscription.customer.email,
-      order_cycle_id: order_cycle_id,
-      distributor_id: subscription.shop_id,
-      shipping_method_id: subscription.shipping_method_id
-    )
-    order.update_attribute(:user, subscription.customer.user)
-    subscription.subscription_line_items.each do |sli|
-      order.line_items.build(variant_id: sli.variant_id, quantity: sli.quantity, skip_stock_check: true)
-    end
-    order.update_attributes(bill_address: subscription.bill_address.dup, ship_address: subscription.ship_address.dup)
-    order.update_distribution_charge!
-    order.payments.create(payment_method_id: subscription.payment_method_id, amount: order.reload.total)
-
+    factory = OrderFactory.new(order_attrs, skip_stock_check: true)
+    self.order = factory.create
     save!
     order
   end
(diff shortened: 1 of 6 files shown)


app/services/order_factory.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def create_attrs
+    create_attrs = attrs.slice(:customer_id, :order_cycle_id, :distributor_id, :shipping_method_id)
+    create_attrs[:email] = customer.email
+    create_attrs
+  end
+
+  def build_line_items
+    attrs[:line_items].each do |li|
+      next unless variant = Spree::Variant.find_by_id(li[:variant_id])
```
mkllnk: Is this something we should warn about or is that handled in a later step?

app/models/proxy_order.rb
```diff
@@ -42,21 +42,8 @@ def resume
...
-    order.update_attribute(:user, subscription.customer.user)
-    subscription.subscription_line_items.each do |sli|
-      order.line_items.build(variant_id: sli.variant_id, quantity: sli.quantity, skip_stock_check: true)
-    end
-    order.update_attributes(bill_address: subscription.bill_address.dup, ship_address: subscription.ship_address.dup)
-    order.update_distribution_charge!
-    order.payments.create(payment_method_id: subscription.payment_method_id, amount: order.reload.total)
-
+    factory = OrderFactory.new(order_attrs, skip_stock_check: true)
+    self.order = factory.create
```
mkllnk: Maybe a stupid question, but what is a ProxyOrder? Some class level documentation could help to understand this.

app/services/order_factory.rb
oeoeaio: Maybe I should just remove the `next unless`? A missing variant here is almost impossible as the relationship between the two tables is described by a foreign key.

app/models/proxy_order.rb
oeoeaio: I'll add a comment, to the top of the class, but for posterity I'll also describe it here.

A ProxyOrder is a placeholder for an order that allows us to lazily create real orders. Each subscription has many ProxyOrders, one for each OrderCycle to which the Subscription applies. Each ProxyOrder knows how to instantiate a real order when it needs to (ie, when a user wants to alter a specific items for a given order cycle, or when the order is actually placed when the order cycle opens. ProxyOrders can also be cancelled (without having to instantiate an actual Order and the cancel it).

Probably the biggest advantage of using the proxy pattern is that changes can be made to the underlying subscription (eg. payment or shipping methods changed, line items added) without having to propagate these changes to a bunch of instantiated orders (though of course this logic is required because orders CAN be instantiated, but use of proxy orders just reduces complexity as much as possible while maintaining flexibility).

oeoeaio:
@mkllnk I agree with regard to the scoping, it's really inefficient. We need a passive system rather than one which need to be actively applied every time we want to use an override. I suspect that something in the conversation we had about networked products at the end of last year will help us solve it, but for now I think you're right, we just have to get this done and then we can work on the next thing.

sstead:
Testing notes

If there were inventory price overrides they were correctly showing everywhere EXCEPT in the backend when I look at an existing standing order (see below)

image

Google doc - https://docs.google.com/document/d/1olKGdvqlj4ejx3kZnbwXPZA3uvsQ7_68R747yWNxMAo/edit?usp=sharing

oeoeaio:
👍 I know what this is, I will handle it in a separate PR

PR #2095, opened 2018-02-23
Added documentation for SubscriptionLineItemSerializer.price_estimate

bradarv90:
#### What? Why?

Just a simple addition of documentation for SubscriptionLineItemSerializer.price_estimate. 

#### What should we test?

Is the documentation clear?

### DIFF
--- a/app/serializers/api/admin/subscription_line_item_serializer.rb
+++ b/app/serializers/api/admin/subscription_line_item_serializer.rb
@@ -7,6 +7,9 @@ module Api
         "#{object.variant.product.name} - #{object.variant.full_name}"
       end
 
+      # Returns price estimate based on either the estimated price 
+      # for that item or the expected fees for that item.
+      # Otherwise returns "?"
       def price_estimate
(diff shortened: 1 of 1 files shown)


oeoeaio:
Hey @bradarv90, thanks very much for your interest and contribution!

I'm actually still working on this code, and I've just made some changes which probably make this comment obsolete. I'll be submitting a PR shortly.

This illustrates what IMHO is part of the problem with making comments on individual methods: they quickly become obsolete and add to confusion rather than alleviating it. If a method looks like it requires commenting, it is almost always preferable to refactor the method so that what is doing is more obvious. The only exception to this is when a method that is 1) very stable, 2) very very difficult to refactor, and 3) contains confusing logic that would be aided by a comment.

Would you mind if I close this PR for now? If you think is important that it be merged, please let me know and we can discuss further.

Thanks!

bradarv90:
I can totally understand closing this PR. I'll close it for you :)

PR #2096, opened 2018-02-23
Subs:  Improve estimates of item prices and fees

oeoeaio:
#### What? Why?

Fixes two main issues: 
1) The outstanding issue of subs UI not taking account of overrides document by @sstead while testing #2093
2) The lack of an estimate for shipping and payment fees for subscriptions, documented by @sstead while testing #2089

The first couple of commits are a refactor of the existing estimation logic into a dedicated class, I can move these commits into a separate PR if it is too confusing to have them in here. 

#### What should we test?

Price estimates for subscriptions line items should now take variant overrides into account, throughout the platform, especially important to test the subscriptions wizard and index page

Fee estimates should be included in the displayed total for subscriptions once they have been saved

#### Release notes

### DIFF
--- a/app/assets/javascripts/admin/subscriptions/services/subscription_functions.js.coffee
+++ b/app/assets/javascripts/admin/subscriptions/services/subscription_functions.js.coffee
@@ -8,8 +8,11 @@ angular.module("admin.subscriptions").factory 'SubscriptionFunctions', ($injecto
       subtotal += item.price_estimate * item.quantity
     , 0
 
+  estimatedFees: ->
+    @shipping_fee_estimate + @payment_fee_estimate
+
   estimatedTotal: ->
-    @estimatedSubtotal()
+    @estimatedSubtotal() + @estimatedFees()
 
   customer: ->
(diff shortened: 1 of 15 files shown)


sstead:
**Testing notes**

Outstanding:
- there's no fee breakdown in subs backend
- Enterprise fees aren't shown (this is probably correct though)
- The single subscription order view doesn’t include fees until it’s in CART state.
- Tax isn’t correctly calculating on the shipping fee in the first email the goes to the customer when the OC opens.

Otherwise inventory and fees relating to subs looks good!

https://docs.google.com/document/d/1XSuv9phXKcfPJTevJIbY1a28OTsuJTTIgHwrTKVn5Sg/edit#

oeoeaio:
Nothing broken, will deal with outstanding issues in another pull request

myriamboure:
@oeoeaio just disconnected from Subscription v1 issue to not pollute zenboard hub, if you needed to keep the connexion for any reason please reconnect :)

PR #2107, opened 2018-03-02
Subs: Add message to indicate OC has no close date

oeoeaio:
#### What? Why?

Fixes an issue [picked up](https://github.com/openfoodfoundation/openfoodnetwork/pull/2089#issuecomment-367561211) by @sstead while testing #2089: when an order cycle does not have a close date, the corresponding order within the subscription interface said it "closed a few seconds ago". 

#### What should we test?

I've added a message which indicates that the order cycle in question has no close date.

#### Release notes


sstead:
**Testing notes**

All good.

https://docs.google.com/document/d/1MTYRw0xhSR-HtguxIyUgZpyhIvyyYwqAIQzLxIhCC6k/edit#

PR #2108, opened 2018-03-02
Subs neg stock

oeoeaio:
#### What? Why?

Fixes an issue related to capping the quantity of a line item when the stock for the relevant variant is negative. Stock was being capped to zero as expected, but this would then cause an error because it would fail the `sufficient_stock?` validation.

#### What should we test?

This issue particular affected subscriptions which require that all items are added in order to send notification emails, regardless of whether stock is available or not. If a subscription requests variants with negative stock, this should no longer cause an issue when the notification emails are sent.

#### Release notes


sstead:
**Testing notes**

The presence of negative stock levels isn't causing any probems that I can see. 

https://docs.google.com/document/d/1U0eiLb6N6oUF8zVFm4ns8kXZ0DmJZlZj-hbyPTsLXUQ/edit#

PR #2110, opened 2018-03-02
Subs: Fix summary emails for shop owners

oeoeaio:
#### What? Why?

Subscription placement summary emails (when an OC opens) were failing to send because of a small bug caused by an array being used instead of a hash.

#### What should we test?

Summary emails should now work consistently, it is especially important to check the case where an issue occurs with one or more subscriptions, and that the summary email is sent successfully in this case.

#### Release notes

### DIFF
--- a/lib/open_food_network/subscription_summary.rb
+++ b/lib/open_food_network/subscription_summary.rb
@@ -18,7 +18,7 @@ module OpenFoodNetwork
     end
 
     def record_issue(type, order, message)
-      issues[type] ||= []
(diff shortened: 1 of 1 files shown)


sstead:
Testing notes

- yes, the email went out correctly despite the standing order having a 'out of stock' issue

https://docs.google.com/document/d/1plxe0t-Qf39Ov_GaCFh6W9Ris4IXTuCYXkUPW_Qk0oY/edit?usp=sharing

PR #2114, opened 2018-03-06
Fix old rspec syntax

ltrls:
#### What? Why?

This fixes old rspec syntax in files that I've been recently modifying.


sauloperez:
Great @ltrls ! keep in mind though that will be using http://yujinakayama.me/transpec/ soon so do not put too much effort on this beyond fixing the ones you see on your way through the codebase.

PR #2115, opened 2018-03-06
Localize sample data

### LINKED ISSUE #2071: Make load_sample_data not dependent on Australia
## Description

When running the rake task `dev:load_sample_data` from an instance where `DEFAULT_COUNTRY_CODE` is other than `AUS` (specified in https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/application.yml.example#L9) an error is raised and thus no sample data can be created.

## Expected Behavior

`dev:load_sample_data` should work in any instance and any stage (development, staging, production) regardless of the values specified in `config/application.yml`.

## Actual Behavior

When `dev:load_sample_data` is run with a `config/application.yml` containing the following:

```
# Add application configuration variables here, as shown below.
#
# Change this, it has serious security implications.
# Minimum 30 but usually 128 characters. To obtain run 'rake secret', or faster, 'openssl rand -hex 128'
SECRET_TOKEN: <classified>

TIMEZONE: Madrid
# Default country for dropdowns etc. See for codes: http://en.wikipedia.org/wiki/ISO_3166-1
DEFAULT_COUNTRY_CODE: ES
# Locale for translation.
LOCALE: es

CURRENCY: EUR
# Spree zone.
CHECKOUT_ZONE: Europe

GOOGLE_MAPS_API_KEY: <classified>
S3_ACCESS_KEY: 
S3_SECRET: 

STRIPE_CLIENT_ID: 
STRIPE_INSTANCE_SECRET_KEY: 
```

The error below is raised:

```
openfoodnetwork@staging-katuma:~/apps/openfoodnetwork/current$ bundle exec rake openfoodnetwork:dev:load_sample_data
rake aborted!
ActiveRecord::RecordInvalid: La validación falló: Provincia ó Estado no puede estar en blanco
/home/openfoodnetwork/.gem/ruby/2.1.0/gems/activerecord-3.2.21/lib/active_record/validations.rb:56:in `save!'
/home/openfoodnetwork/.gem/ruby/2.1.0/gems/activerecord-3.2.21/lib/active_record/attribute_methods/dirty.rb:33:in `save!'
```

## Steps to Reproduce

1. Log into the staging/development machine you want to use
2. Edit the contents of `config/application.yml` to have `DEFAULT_COUNTRY_CODE: ES` and `LOCALE: es`
2. Run `yes '' | bundle exec rake db:reset openfoodnetwork:dev:load_sample_data` from the app's root folder.

## Animated Gif/Screenshot

failure

## Context

This is blocking any testing in stagings other than the Australian one as all these sample data has to be added by hand and we can't ensure its replicability across deploys. This leads to slightly different data in DB and so we're not testing the exact same scenarios.

## Severity
Putting s2 as preventing any other dev outside Australia to provision staging servers, so blocking the process.

## Your Environment

It does not apply in this case.

## Possible Fix

Go failure by failure checking what data it relies upon and abstracting the explicit conditionals so that they don't rely on Australia only.

sauloperez:
#### What? Why?

Closes #2071 

It turns out that the `openfoodnetwork:dev:load_sample_data` rake task was failing due to the state passed to the addresses we create not belonging to the specified country.

Now the country these addresses are created into is the one defined in the OFN instance through `DEFAULT_COUNTRY_CODE`. This however just makes the already defined Australian addresses belong to the said country, so we will have `25 Myrtle Street, 3153, Bayswater` located in `Michigan, Spain`, for instance.

It's the smallest change needed for us to have sample data in staging and so unblock testing. We will iterate from here. Remember, *small is good* :tm: (:point_left: outcome of last Australian international gathering).

#### What should we test?

We should be able to execute:

```shell
$ bundle exec rake openfoodnetwork:dev:load_sample_data
```

from any staging environment and it should work flawlessly.

#### Release notes

Fixed `openfoodnetwork:dev:load_sample_data` rake task to work on all OFN instances.

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -10,6 +10,7 @@ namespace :openfoodnetwork do
       task_name = "openfoodnetwork:dev:load_sample_data"
 
       spree_user = Spree::User.find_by_email('spree@example.com')
+      country = Spree::Country.find_by_iso(ENV.fetch('DEFAULT_COUNTRY_CODE'))
 
       Spree::MailMethod.create!(
         environment: Rails.env,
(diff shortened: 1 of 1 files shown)


lib/tasks/dev.rake
```diff
@@ -9,6 +9,8 @@ namespace :openfoodnetwork do
       require_relative '../../spec/support/spree/init'
       task_name = "openfoodnetwork:dev:load_sample_data"
 
+      country = Spree::Country.find_by_iso(ENV['DEFAULT_COUNTRY_CODE'])
```
enricostano: Use `fetch()` so it will raise an exception.

PR #2117, opened 2018-03-06
Automatically confirm the spree@example.com user

sauloperez:
#### What? Why?

This allows you to log in right away with `spree@example.com` after you run `bundle exec openfoodnetwork:dev:load_sample_data` rake task. Otherwise, you see an error message asking you to confirm the email first.

#### What should we test?

Exactly what the paragraph above claims.

#### Release notes

The `spree@example.com` user is now confirmed by default when running `bundle exec openfoodnetwork:dev:load_sample_data`.

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -9,9 +9,11 @@ namespace :openfoodnetwork do
       require_relative '../../spec/support/spree/init'
       task_name = "openfoodnetwork:dev:load_sample_data"
 
+      spree_user = Spree::User.find_by_email('spree@example.com')
+
       Spree::MailMethod.create!(
         environment: Rails.env,
-        preferred_mails_from: 'spree@example.com'
+        preferred_mails_from: spree_user.email
       )
 
       # -- Shipping / payment information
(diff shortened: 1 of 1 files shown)


lib/tasks/dev.rake
```diff
@@ -193,6 +195,14 @@ namespace :openfoodnetwork do
       CreateOrderCycle.new(enterprise2, variants).call
 
       EnterpriseRole.create!(user: Spree::User.first, enterprise: enterprise2)
+
+      confirm_user(spree_user)
```
oeoeaio: The Devise `Confirmable` module provides a [confirm!](https://github.com/plataformatec/devise/blob/beeb48c404ae8b4c3c34f6d0a2325ca2bcc9298e/lib/devise/models/confirmable.rb#L58) method, so perhaps you could just do`spree_user.confirm!`.

sauloperez:
Thanks for the tip @oeoeaio !

daniellemoorhead:
Assuming this is tested by the fact that it's been code reviewed @sauloperez - ok to merge this directly without testing?

sauloperez:
Yep, I gave it a try.

PR #2118, opened 2018-03-07
Show payment summary in confirmation email regardless of payment method

oeoeaio:
#### What? Why?

While testing subscriptions, @sstead noticed that confirmation emails for orders with a non-cash payment method did not include a payment summary. After a little bit of investigation I realised that this had been the case since I first [committed](https://github.com/openfoodfoundation/openfoodnetwork/commit/fe0bb49bafcdc046ca6e06eb365c8a6a241e7149) the current version of the confirmation email templates. I have no idea why it is like this so I though I would delete the logic and see what happens.

#### What should we test?

All confirmation emails should include a payment summary, regardless of the payment method on the order.

#### Release notes

Fixed a bug that caused missing payment summaries in order confirmation emails when non-cash payment methods were used.

### DIFF
--- a/app/views/spree/order_mailer/_payment.html.haml
+++ b/app/views/spree/order_mailer/_payment.html.haml
@@ -1,15 +1,14 @@
-- if @order.payments.first.andand.payment_method.andand.type == "Spree::PaymentMethod::Check" and @order.payments.first.andand.payment_method.andand.description
-  %p.callout
-    %span{:style => "float:right;"}
-      - if @order.paid?
-        = t :email_payment_paid
-      - else
-        = t :email_payment_not_paid
-    %strong
-      = t :email_payment_summary
-  %h4
-    = t :email_payment_method
-    %strong=  @order.payments.first.andand.payment_method.andand.name.andand.html_safe
-  %p
-    %em= @order.payments.first.andand.payment_method.andand.description.andand.html_safe
-  %p &nbsp;
+%p.callout
+  %span{:style => "float:right;"}
+    - if @order.paid?
+      = t :email_payment_paid
+    - else
+      = t :email_payment_not_paid
+  %strong
+    = t :email_payment_summary
+%h4
+  = t :email_payment_method
+  %strong=  @order.payments.first.andand.payment_method.andand.name.andand.html_safe
+%p
+  %em= @order.payments.first.andand.payment_method.andand.description.andand.html_safe
+%p &nbsp;


mkllnk:

daniellemoorhead:
@myriamboure did you want to test this as @sstead won't be back in till later next week? Else we can wait till she's back seeing as she found the original bug.

myriamboure:
Testing notes here: https://docs.google.com/document/d/1LpjTCrrnhQKECeoR5rUiCLwWv9KHpKcP2orA5GVuuaM/edit?usp=sharing

The issue above is fixed so this can be merged so I put to ready to go, BUT I identified some remaining issues:
- When I created my first subscription **I didn't know which format to use for "start date"**. I had to check in the user guide to find the info.
**Expected behavior:** let's be consistent and use the same datepicker as everywhere?
- I created an OC with French time so was already closed in Australia, then changed to fit to Aus time, but in the subscription dashboard **the OC still appears as "closed x hours ago" AND there is a translation missing to fix**.
**Expected behavior:** "closed" is translated + if the OC close date is changed the info in subs. dashboard should also be updated.
- I created a second OC at the exact time where the first one was opening. I don't know if a conflict happened but in subs. dashboard **the new OC appears with no OC name and just "Close date not set"** which is pretty confusing.
- And last but not least, I think there is a **UX issue with subscriptions for shops who are set up to "don't allow order modification"**. As a user I receive a confirmation email when the OC opens saying that an order has been made for me. And another one that looks like a duplicate when the OC closes... These two emails make sense if you give the possibility to the customer to modify the order.
Expected behavior: either we say only shops who allow order modification can create a subscription, either we only send one confirmation email to the user then the OC opens (or closes), just as if we was making his order himself.

Shall I open issues for each of those 4 things and put them in the project backlog? I went quickly threw the existing issues and didn't find anything on those... the date picker is the most annoying for me at this stage, and maybe also the non-update of end date in subs. dashboard which is confusing, so the first two points. For the strange bug I think it won't happen to a regular user, and the last one we can just tell the hubs to set up on "allow modification" to start with... Do you agree?

daniellemoorhead:
@myriamboure it would be worth taking this higher level Subscription changes you've noted above and putting them into a separate issue - this PR will be merged and then they may be lost.

Given you've now had a "sneak peak" at Subscriptions (yay!) what you're providing is requests for tweaks and they deserve to be kept somewhere separate. So, perhaps each of these deserves it's own issue? Or at least a single issue that lists them all out as a placeholder for Rob to review and respond?

myriamboure:
I'll create one for each and put them in the project backlog so we can discuss what to do then. Doing it now.

myriamboure:
Ok done they are all in the project backlog will ping Rob on the two most important for me.

oeoeaio:
Ah, now I understand where all of those issues came from! I'm sorry @myriamboure, my summary at the top was very misleading. This issue isn't actually related to subscriptions, that is just what caused Sally to find it. It is totally fine to use regular orders to test this PR. I'm sorry that you encountered all of the not-quite-ready parts of subscriptions! It looks like the code is working either way, so that is good!

PR #2119, opened 2018-03-07
Subs: Fix Order Cycle index crashes

oeoeaio:
#### What? Why?

The Order Cycle index sometimes crashes. This is caused by the complex interaction of the scopes used by the OrderCyclesController#collection method, which can cause multiple copies of a single order cycle to be loaded into the page, which breaks AngularJS indexing.  As far as I can tell, the problem lies with order cycles with a start date after the end date, and those with only an open date and no close date. 

I have done two things to resolve this issue:
1. Added validation to the OrderCycle model requiring that close date must be after open date
2. Return the union of the scopes used by the `#collection` method, rather than concatenating them.

I have also refactored the use of OrderCycleSet by the OrderCyclesController, which should have the happy side effect of significantly speeding up the save time on the order cycles index.

#### What should we test?

The Order Cycles index should load properly regardless of the presence or otherwise of dates on order cycles.

An error should be returned when the open date is after the close date. Test this when:
1. Creating a new OC
2. Updating an existing OC from the edit page
3. Updating multiple OCs from the index page

### DIFF
--- a/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee
@@ -151,23 +151,28 @@ angular.module('admin.orderCycles').factory 'OrderCycle', ($resource, $window, S
       return unless @confirmNoDistributors()
       oc = new OrderCycleResource({order_cycle: this.dataForSubmit()})
       oc.$create (data) ->
-        if data['success']
-          $window.location = destination
+        $window.location = destination
+      , (response) ->
+        if response.data.errors?
+          StatusMessage.display('failure', response.data.errors[0])
         else
-          console.log('Failed to create order cycle')
+          StatusMessage.display('failure', 'Failed to create order cycle')
 
     update: (destination, form) ->
       return unless @confirmNoDistributors()
       oc = new OrderCycleResource({order_cycle: this.dataForSubmit()})
       oc.$update {order_cycle_id: this.order_cycle.id, reloading: (if destination? then 1 else 0)}, (data) =>
-        if data['success']
-          form.$setPristine() if form
-          if destination?
-            $window.location = destination
-          else
-            StatusMessage.display 'success', t('js.order_cycles.update_success')
+        form.$setPristine() if form
+        if destination?
+          $window.location = destination
         else
-          console.log('Failed to update order cycle')
+          StatusMessage.display 'success', t('js.order_cycles.update_success')
+      , (response) ->
+        if response.data.errors?
+          StatusMessage.display('failure', response.data.errors[0])
+        else
+          StatusMessage.display('failure', 'Failed to create order cycle')
+
 
     confirmNoDistributors: ->
       if @order_cycle.outgoing_exchanges.length == 0
--- a/app/controllers/admin/order_cycles_controller.rb
+++ b/app/controllers/admin/order_cycles_controller.rb
@@ -9,7 +9,6 @@ module Admin
     prepend_before_filter :load_data_for_index, :only => :index
     before_filter :require_coordinator, only: :new
     before_filter :remove_protected_attrs, only: [:update]
-    before_filter :remove_unauthorized_bulk_attrs, only: [:bulk_update]
     before_filter :check_editable_schedule_ids, only: [:create, :update]
     around_filter :protect_invalid_destroy, only: :destroy
     create.after :sync_subscriptions
@@ -79,11 +78,9 @@ module Admin
     end
 
     def bulk_update
-      @order_cycle_set = params[:order_cycle_set] && OrderCycleSet.new(params[:order_cycle_set])
-      if @order_cycle_set.andand.save
+      if order_cycle_set.andand.save
         respond_to do |format|
-          order_cycles = OrderCycle.where(id: params[:order_cycle_set][:collection_attributes].map{ |k,v| v[:id] })
-          format.json { render_as_json order_cycles, ams_prefix: 'index', current_user: spree_current_user }
+          format.json { render_as_json @order_cycles, ams_prefix: 'index', current_user: spree_current_user }
         end
       else
         respond_to do |format|
@@ -109,6 +106,7 @@ module Admin
     protected
     def collection
       return Enterprise.where("1=0") unless json_request?
+      return order_cycles_from_set if params[:order_cycle_set]
       ocs = if params[:as] == "distributor"
         OrderCycle.preload(:schedules).ransack(params[:q]).result.
           involving_managed_distributors_of(spree_current_user).order('updated_at DESC')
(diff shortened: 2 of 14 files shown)


app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee
```diff
@@ -151,23 +151,28 @@ angular.module('admin.orderCycles').factory 'OrderCycle', ($resource, $window, S
...
+        if destination?
+          $window.location = destination
         else
-          console.log('Failed to update order cycle')
+          StatusMessage.display 'success', t('js.order_cycles.update_success')
+      , (response) ->
+        if response.data.errors?
+          StatusMessage.display('failure', response.data.errors[0])
+        else
+          StatusMessage.display('failure', 'Failed to create order cycle')
```
sauloperez: Can we create an issue for these untranslated strings?

app/controllers/admin/order_cycles_controller.rb
```diff
@@ -216,6 +214,20 @@ def sync_subscriptions
...
+      OrderCycle.where(id: params[:order_cycle_set][:collection_attributes].map{ |k,v| v[:id] })
+    end
+
+    def order_cycle_set
+      @order_cycle_set ||= OrderCycleSet.new(@order_cycles, params[:order_cycle_set])
+    end
+
+    def require_order_cycle_set_params
+      return if params[:order_cycle_set]
+      render :json => { errors: t('admin.order_cycles.bulk_update.no_data') }, status: :unprocessable_entity
```
sauloperez: Now that you are at it, let's please unhasrocketize the `:json` key too.

app/models/order_cycle_set.rb
```diff
@@ -1,5 +1,5 @@
 class OrderCycleSet < ModelSet
-  def initialize(attributes={})
-    super(OrderCycle, OrderCycle.all, attributes)
+  def initialize(collection, attributes={})
+    super(OrderCycle, collection, attributes)
   end
```
sauloperez: Thanks!

oeoeaio:
> Why did it actually fail when having multiple instances of the same order cycle?

@sauloperez I've altered the comment at the top, but basically the problem was that the `ng-repeat` directive used to render the rows of the table is tracking objects using `track by orderCycle.id` (this helps speed up rendering), so it is expecting unique ids for the order cycle objects, and when it encounters a second object with the same `id` it becomes sad. 😢

oeoeaio:
@sauloperez please approve unless you would like to see some more changes

daniellemoorhead:
Hey @sauloperez, @oeoeaio be like:

giphy

mkllnk:

myriamboure:
Testing notes: https://docs.google.com/document/d/1rqlqm7ZXtMZkSCW5GYmJ8x_HUaFA2HlHFdcWT-owVIU/edit#

**I see one thing to fix:**
- I found one case not covered when editing multiple OCs in bulk edit: when you have one OC with an end date but no start date. You come and do two modifications: you add a start date to that OC + you change the end date of another OC and make it prior to start date > in that case error message displays but start date has not been saved so if user leaves he looses this modification
- This doesn't prevent from merging for me, as this PR is already solving the main issue, so I'll open straight away an issue for that and link it here.

**I see one improvement in term of UX:**
- The error message is different from other error messages in the back office and not really "obvious" to users, it's pretty discrete. I'm afraid users will not see it and just believe their changes have been saved and go to some other page. Nothing prevents them to do so. I would suggest either to change the error message for a pop-in error message that displays in the middle of the page, or I don't know but something more visible and obvious. Doesn't prevent from merging.

What do you think @oeoeaio ?

So in order to merge the first and more urgent fix I would put that as ready to go, open an issue for the small remaining fix and we can discuss UX here to decide if we do something else or not for now.

myriamboure:
Oups, seems tests are not passing @oeoeaio :-(

myriamboure:
As tests are not passing and @oeoeaio needs anyway to have a look at it before we can merge I suggest you fix the small thing Rob at the same time, so I don't open another issue for now on the last remaining bit.

oeoeaio:
Thanks @myriamboure!

I can change the error message on the index page to a pop-up, that is easy. In terms of the first issue, would it be ok to add "Your changes have not been saved" to the end of the error message, to make it clear that they will need to save again to preserve their changes? Saving some changes but not others will require a significant rewrite, and I'm not sure it is worth it in this instance...

myriamboure:
@oeoeaio:
- regarding how error message is displayed yes let's for a pop-up then, unless anyone believes it's not wise (I'm not UX expert but just putting myself in the shoes of users... and I believe pop-up will prevent misundertanding.
- regarding bug 1 I don't understand what you say, because when I tested I noticed that even when I get the error message, when I do multiple updates at the same time with one error on one end date being prior to start date, other changes ARE already saved... if you look at the screenshots in testing notes you will see it yourself. BUT if you don't know why is that and find it easier not to fix the only case that is not saved and add the message "Your changes have not been saved" in the pop-up, I have no ojection, that works for me too... even if in fact most of them have been made ;-)

myriamboure:
And @oeoeaio Feel free to merge now this one to start with if you have no time to work on this remaining bits now and will do the pop-up thing later so that we fix asap the index crash ;-) If can open an issue so that we don't forget about it in that case.

oeoeaio:
> when I do multiple updates at the same time with one error on one end date being prior to start date, other changes ARE already saved

Ok, great! That makes things easier.

> so I think we should either fix the only small change which is not saved (start date when only end date was there) and/or say in the pop-up "Close date must be after open date. Some of your changes couldn't be saved."

Note sure what you mean by the first part (could you explain a bit more?), but I'm happy with second part ("some of your changes could not be saved").

myriamboure:
@oeoeaio sorry if I'm not clear.
> when I do multiple updates at the same time with one error on one end date being prior to start date, other changes ARE already saved
This is true for almost all cases apart one that I identified while testing (I put it in pink in the document) so to sum up, when I have one error due to end date < start date and have done other modifications at the same time, if the other modification is:
- I had an OC with correct end/start date. I change the start and/or end date but keep good logic > this is saved.
- I had an OC with no dates. I change it and add a start + end date in good logic > this is saved.
-  I had an OC with no dates. I change it and add a start date > this is saved.
- I had an OC with no dates. I change it and add a end date > this is saved.
-  I had an OC with only a start date. I change it and add an end date in good logic > this is saved.
-  I had an OC with only an end date. I change it and add a start date in good logic > this is **NOT** saved.

What I say is that if we don't fix this and the user get the error message and leave without fixing, he will now know which of the changes he made are saved and which are not. That's why for a good UX I suggested:
1- To consistently make sure that all  other changes apart end date < start date are saved (the last case is not)
2- If we manage to do so, then it means the only record not saved is the end date < start date one, so message can be "Close date must be after open date. Some of your changes couldn't be saved." or even more precisely "Close date must be after open date. Corresponding records couldn't be saved." or something like that.
3- To help user fix the problem make very visible which record is concerned by displaying the start and end date fields in red (if possible would be great).

This is "best UX" option IMO but then depending on what is easy or not to do, you can adjust of course.
Is all that clearer?

oeoeaio:
Yes, very clear, thanks @myriamboure! I will see if I can replicate and work out what is happening in that case.

oeoeaio:
Ok, so what is happening is actually a little more obscure.

The order cycles are updated in order of database id (~ the date they were created). When an order cycle with an error is updated, all order cycles with a lower id are updated successfully, and all subsequent order cycles are not updated.

I think the best way to solve this issue is just to have dynamic validation of start and end dates in the page so that the user is notified of any issues before they even attempt to submit, I think that will provide the best UX.

myriamboure:
Yes @oeoeaio I definitely agree, that would be great in terms of UX! 
1- When update a line if start date > end date error message straight away (so don't let you type that in fact)
2- When I validate I won't have any issue blocking validation (normally). So we keep the standard/consistent behavior which is to still save with the save button and the message "you have unsaved changes" in the low banner. 

Shall I open an issue "Fix UX issue on OC dates validation" and assign it to you then? Has this been merged actually? Because it seems Semaphor checks didn't pass...

PR #2120, opened 2018-03-07
Subs: messaging around unpausing a subscription

oeoeaio:
#### What? Why?

It was not clear what happened to the orders for open order cycles when a subscription is unpaused. I have added some messaging to communicate the need to manually resume any orders for open order cycles if that is what is desired.

#### What should we test?

Check that appropriate messaging is displayed when:
a) A subscription is paused and then unpaused within the same order cycle.
b) A subscription is paused in once order cycle and then unpaused in a subsequent order cycle.

#### Release notes


sstead:
**Testing notes**

Things that need attention:
- Can we put an alert that lets people know that if they resume a paused order while the OC is open that a new order will be created. This is what happens, so it would be good to let the shop owner know, so there's no confusion.
- There's some issues with resuming an order that's been cancelled. Payment state shows as VOID and the customer gets no email at close of the OC.

Working well:
- The messaging that the manager sees when a subscription is paused (and existing order cancelled) and then resumed within the same OC, is clear. 

https://docs.google.com/document/d/1bZs92CakzknJhaI2xd8ycyf_x-1YVBb80wTCiChK-n8/edit#

Issue #1059, opened 2016-06-22
SO - Edit or delete an order cycle schedule


daniellemoorhead:
**As a shop owner I want to edit or delete an order cycle schedule**

Go to the order cycles screen.
Click on the Schedule name in the table (if it's not visible then you filter/search for it).
Result: the schedule modal displays in edit mode (same as the add schedule modal).

**EDIT**
- User can edit the name and the order cycles within the schedule, and then click on update (changed label).
- Schedule is updated in database.

**DELETE**
- User can also delete the schedule by clicking on the delete button.
- If no standing orders are in place for the schedule then it is deleted once the user confirms the action.
- If there are standing orders in place for the schedule then an error message displays saying you can't delete this because there are standing orders attached. The user then hits OK.

![img_0565](https://cloud.githubusercontent.com/assets/2067859/16256969/ded80468-3898-11e6-9736-fb53c803601c.JPG)

nanoharsha:
@daniellemoorhead @oeoeaio does the above issue and work being done to fix it, also address the below issue #768  raised by @myriamboure .

Cheers.

myriamboure:
I don't think so @nanoharsha I answered in the other thread :-) I was talking about any order cycle, here is a specific feature for standing orders. The other issue was that if you create a test order cycle and just go on it on the front end and add an item to your basket to test, the system create an uncomplete order with the status "cart", and then you can't delete the order cycle... I think it's different.

sstead:
Search for schedule not yet working...
![image](https://cloud.githubusercontent.com/assets/10843056/20735910/6327e5a8-b6f7-11e6-9b3f-59a1bf1b8856.png)

sstead:
Yes - can change name and this saves in all places
Yes - Can delete if no standing orders, after clicking confirm
Delete - If has associated standing order can't delete. When you try to delete I get a 'are you sure' message box. If you click yes, nothing happens and the Schedule hasn't been deleted.

sstead:
This is outstanding, all else is working.
Delete - If has associated standing order can't delete. When you try to delete I get a 'are you sure' message box. If you click yes, nothing happens and the Schedule hasn't been deleted. Would be good to get a dialogue message saying 'sorry you can't delete this schedule because it has associated standing orders'.

sstead:
This has been fixed, there's clear messaging saying that you can't delete a schedule with subs assigned to it.

[...]

PR #2121, opened 2018-03-07
Subs: Add error message for case when schedule cannot be deleted

oeoeaio:
#### What? Why?

Sally picked up [this issue](https://github.com/openfoodfoundation/openfoodnetwork/issues/1059#issuecomment-346242475) while testing #1059. So I added some messaging explaining that schedule cannot be deleted while there are subscriptions that depend on them.

#### What should we test?

Attempting to delete a schedule while there are associated subscriptions should display an error message at the bottom of the modal.

#### Release notes

### DIFF
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -4,6 +4,7 @@ require 'open_food_network/proxy_order_syncer'
 module Admin
   class SchedulesController < ResourceController
     before_filter :check_editable_order_cycle_ids, only: [:create, :update]
+    before_filter :check_dependent_subscriptions, only: [:destroy]
     create.after :sync_subscriptions
     update.after :sync_subscriptions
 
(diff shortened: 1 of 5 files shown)


app/controllers/admin/schedules_controller.rb
```diff
@@ -4,6 +4,7 @@
 module Admin
   class SchedulesController < ResourceController
     before_filter :check_editable_order_cycle_ids, only: [:create, :update]
+    before_filter :check_dependent_subscriptions, only: [:destroy]
```
sauloperez: In the case of a single action, as in here, I think a filter doesn't bring any benefit. The indirection actually makes things a litlle bit harder to read.

oeoeaio: Hey @sauloperez, I know you hate single-use before filters 😉, but in the case where you want to return early they actually have a special property. When you `render` or `redirect_to` from a before filter, the action is never hit. So it is useful for rendering an error when a certain condition is not met, for example. This is *possible* but more difficult to achieve using a simple method call from within the action: you need to a) add and extra render to the action method, b) return from the filter method but return a value which the action method understands, or c) check `performed?`. All of these option add unnecessary clutter the action method IMHO.

sauloperez: :+1:

sstead:
**Testing notes**

Working as intended...
- There's now an error message to the user when they try to delete a schedule that can't be deleted.

Possible future improvement...
- I don't think this is a show stopper- but if I delete all the associated subscription orders in a schedule I still can't delete the schedule. This would only be a pain to a user if they setup a testing schedule, with some subscriptions, and later wanted to 'clear the slate'. That schedule would linger, but it wouldn't stop them from doing anything, it's just clutter.

https://docs.google.com/document/d/13Dl4HuoD6wY6E6buURu9O1McADHiAsvrZGpHE7tqHZ4/edit?usp=sharing

PR #2122, opened 2018-03-07
Subs: prevent fees from displaying in email when subscription order cannot be fulfilled 

oeoeaio:
#### What? Why?

When none of the items requested in a subscription are available when an order cycle opens, the customer is sent a notification email to let them know that an order could not be placed. Despite the fact that the email says that the order was not processed, some fees were still showing in the order summary shown in the email. This PR should resolve that issue.

#### What should we test?

Notification emails for customers indicating a failure to place an order due to insufficient stock should not show fees of any sort.

#### Release notes

### DIFF
--- a/spec/jobs/subscription_placement_job_spec.rb
+++ b/spec/jobs/subscription_placement_job_spec.rb
@@ -149,7 +149,8 @@ describe SubscriptionPlacementJob do
 
         it "does not place the order, clears, all adjustments, and sends an empty_order email" do
           expect{ job.send(:process, order) }.to_not change{ order.reload.completed_at }.from(nil)
-          expect(order.reload.adjustments).to be_empty
+          expect(order.adjustments).to be_empty
+          expect(order.total).to eq 0
           expect(job).to_not have_received(:send_placement_email)
           expect(job).to have_received(:send_empty_email)
         end
--- a/spec/jobs/subscription_placement_job_spec.rb
+++ b/spec/jobs/subscription_placement_job_spec.rb
@@ -120,7 +120,12 @@ describe SubscriptionPlacementJob do
 
   describe "processing a subscription order" do
     let(:subscription) { create(:subscription, with_items: true) }
+    let(:shop) { subscription.shop }
     let(:proxy_order) { create(:proxy_order, subscription: subscription) }
+    let(:oc) { proxy_order.order_cycle }
+    let(:ex) { oc.exchanges.outgoing.find_by_sender_id_and_receiver_id(shop.id, shop.id) }
+    let(:fee) { create(:enterprise_fee, enterprise: shop, fee_type: 'sales', amount: 10) }
+    let!(:exchange_fee) { ExchangeFee.create!(exchange: ex, enterprise_fee: fee) }
     let!(:order) { proxy_order.initialise_order! }
 
     before do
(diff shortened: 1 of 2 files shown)


spec/jobs/subscription_placement_job_spec.rb
```diff
@@ -147,8 +147,10 @@
           allow(job).to receive(:unavailable_stock_lines_for) { order.line_items }
         end
 
-        it "does not place the order, sends an empty_order email" do
+        it "does not place the order, clears, all adjustments, and sends an empty_order email" do
+          expect(order.adjustments).to_not be_empty
```
sauloperez: I wonder if we really need to add this one. Our only expectation is the one you added below. What happens before is not that relevant IMO.

oeoeaio: Sure, I'm happy to remove it.

mkllnk:

mkllnk:
@myriamboure I'm surprised this didn't get picked up for testing last night. Do I need to notify people? I'm working Tuesday to Friday now and the next two Mondays as well. I hope to stage something every working day. But do we have testers every day? I'm happy to continue this conversation in Slack.

myriamboure:
Oups @mkllnk actually @RachL and I are in Barcelona working with the Coopdevs team and spent the whole day on product chain inception, we will finish that tomorrow. We can try to do some testing in the day though to take advantage of you being able to stage every working day this week, we'll see but this week is a bit special because of colocating French and Catalunya team working on bigger picture inceptions first deep dive ;)

sstead:
**Testing Notes** 

- Fees aren't being listed, but the subtotal and total are not $0, which is what I'd expect. The total is including the value of enterprise fees.

https://docs.google.com/document/d/10AjV6FRGhsjLo4OtR3vQ8qKJrJPHnpTDVYtXi2ZSmbQ/edit#

myriamboure:
Put that back in code review as it seems @oeoeaio already fixed the issue after Sally tested.

mkllnk:
Ah, Rob had actually staged it already for testing. I just tried to put it back into testing after reviewing it, but my Zenhub stopped working. :-(

mkllnk:
@myriamboure This is staged on https://staging1.openfood.com.au/ again.

myriamboure:
Testing notes: https://docs.google.com/document/d/10AjV6FRGhsjLo4OtR3vQ8qKJrJPHnpTDVYtXi2ZSmbQ/edit?usp=sharing

It seems the bug is solved BUT I have the impression that a new one has been introduced, I'm not sure if that's connected to this one or not. When I create a new subscription, in the "subscription dashboard" "items" column, the total is not consistent with subtotal and I'm not able to understand how it is calculated (see red lines in testing notes.) Can you have a look @oeoeaio or @sstead and tell if it's connected to that PR or if we need to open a new issue?

oeoeaio:
HI @myriamboure,

Not sure about the issue you found. Definitely troubling that the different between the subtotal and the total is not itemised. I will attempt to replicate and drill down into the issue.

sauloperez:
@myriamboure do you agree on merging this as per @oeoeaio's comment?

myriamboure:
I guess so @sauloperez I didn't retest but if the bug I found was not connected and is addressed somewhere else anyway this should be good to go!

PR #2128, opened 2018-03-09
2113 Display only selected languages in switcher

### LINKED ISSUE #2113: Multilingual language listing
With multilingual, the displayed languages can't be controlled, and the default language gets appended to the available languages listed in the `AVAILABLE_LANGUAGES` list. 

screenshot from 2018-03-05 16-20-13

## Description
`en` is being added to the languages list when it shouldn't be.

## Expected Behavior
A list of available languages can be defined in `application.yml`, and the languages shown should not be different from that list.

## Actual Behavior
Extra languages get added in, possibly just the default/fallback language.

## Steps to Reproduce
screenshot from 2018-03-05 17-22-48

## Context
Instances using multilingual.

## Severity
S2

## Possible Fixes
1. Add another env variable, which lists the languages to be shown?
2. Using the order in exisitng `available_locales` e.g. `[en, en_GB]` means that `en` is the fallback

mkllnk:
#### What? Why?

Closes #2113.

#### What should we test?

The display and behaviour of the language switcher with different locale configurations.

#### Release notes

Fixed bug that always displayed an additional "English" in the language switcher if the server didn't have `en` configured as language.

Not at all.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,7 @@
 require_relative 'boot'
 
 require 'rails/all'
+require_relative "../lib/open_food_network/i18n_config"
 
 if defined?(Bundler)
   # If you precompile assets before deploying to production, use this line
@@ -104,9 +105,8 @@ module Openfoodnetwork
 
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
-    config.i18n.default_locale = ENV["LOCALE"] || ENV["I18N_LOCALE"] || "en"
-    config.i18n.available_locales = ENV["AVAILABLE_LOCALES"].andand.split(/[\s,]/).andand.map(&:strip) || []
-    config.i18n.available_locales = (config.i18n.available_locales + [config.i18n.default_locale, 'en']).uniq
+    config.i18n.default_locale = OpenFoodNetwork::I18nConfig.default_locale
+    config.i18n.available_locales = OpenFoodNetwork::I18nConfig.available_locales
     I18n.locale = config.i18n.locale = config.i18n.default_locale
 
     # Setting this to true causes a performance regression in Rails 3.2.17
--- a/lib/open_food_network/i18n_config.rb
+++ b/lib/open_food_network/i18n_config.rb
@@ -1,0 +1,22 @@
+module OpenFoodNetwork
+  # Provides access to the language settings.
+  # Currently, language settings are read from the environment.
+  # See: config/application.yml
+  class I18nConfig
+    def self.selectable_locales
+      ENV["AVAILABLE_LOCALES"].andand.split(/[\s,]/).andand.map(&:strip) || []
+    end
+
+    def self.available_locales
+      (selectable_locales + [default_locale, 'en']).uniq
+    end
+
+    def self.default_locale
+      ENV["LOCALE"] || ENV["I18N_LOCALE"] || source_locale
+    end
+
+    def self.source_locale
+      "en"
+    end
+  end
+end
--- a/lib/open_food_network/i18n_config.rb
+++ b/lib/open_food_network/i18n_config.rb
@@ -10,7 +10,7 @@ module OpenFoodNetwork
 
     # All locales that can be accessed by the application, including fallbacks.
     def self.available_locales
-      (selectable_locales + [default_locale, 'en']).uniq
+      (selectable_locales + [default_locale, source_locale]).uniq
     end
 
     # The default locale that is used when the user doesn't have a preference.
--- a/lib/open_food_network/i18n_config.rb
+++ b/lib/open_food_network/i18n_config.rb
@@ -4,7 +4,7 @@ module OpenFoodNetwork
   # See: config/application.yml
   class I18nConfig
     def self.selectable_locales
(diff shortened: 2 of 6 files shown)


lin-d-hop:
Sweeeet! @myriamboure @kirstenalarsen (product managers this round) and @daniellemoorhead 
Note this is a bug that affects all non-default-en instances so it would be good to patch this on the last release.

myriamboure:
@sauloperez if you can quickly review we can hopefully stage that maybe today or tomorrow somewhere?

lib/open_food_network/i18n_config.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    def self.default_locale
+      ENV["LOCALE"] || ENV["I18N_LOCALE"] || source_locale
+    end
+
+    def self.source_locale
+      "en"
+    end
+  end
```
sauloperez: can we doc what's the purpose of each these methods? It takes some time to guess.

spec/features/consumer/multilingual_spec.rb
```diff
@@ -71,11 +71,10 @@
 
   describe "using the language switcher UI" do
     context "when there is only one language available" do
-      around do |example|
-        available_locales = I18n.available_locales
-        I18n.available_locales = ['en']
-        example.run
-        I18n.available_locales = available_locales
+      before do
+        allow(ENV).to receive(:[]).and_call_original
```
sauloperez: why this stub?

config/application.rb
```diff
@@ -1,6 +1,7 @@
 require_relative 'boot'
 
 require 'rails/all'
+require_relative "../lib/open_food_network/i18n_config"
```
enricostano: this shouldn't be necessary, Rails should load code in `lib/`

sauloperez: AFAIK `lib/` is not loaded by deafult only code within `app/`

mkllnk: I found that this was necessary.

mkllnk: We could auto-load lib, but is that the cleanest way to do it? It would mean that we load e.g. rake tasks into our main application. I would like a second opinion before doing this.

sauloperez: AFAIK auto-loading `lib/` is not a good practice. I don't see it a big deal to add a require when we need a lib.

enricostano: `require_relative` is a bit of an anti-pattern in Rails, but we can live with it

mkllnk: A tried to use `Rails.root`, but that was empty when loading a rake task.

spec/features/consumer/multilingual_spec.rb
```diff
@@ -84,21 +83,29 @@
       end
     end
 
-    it "allows switching language via the main navigation" do
-      visit root_path
+    context "when there are multiple languages available" do
+      before do
+        allow(ENV).to receive(:[]).and_call_original
```
sauloperez: I'm wondering why we need this line. AFAIK we are just stubbing a particular env var in the line below and not affecting the rest of ENV.

sauloperez: I don't see this done in other specs.

config/application.rb
enricostano: we should always try to use `require` but don't worry, let's move on. we'll see if it cause any problem in the future

:+1:

spec/features/consumer/multilingual_spec.rb
mkllnk: We stub the call of `[]` on ENV. But in a feature spec, a lot more variables are accessed than the ones we stub here. I got an error that call of something like `ENV['TMP_DIR']` was not expected. The test doesn't want to care about all the other environment variables and therefore I put this catch all in. Is there a better way?

pacodelaluna:
Just deployed it on https://staging.openfoodfrance.fr. @myriamboure Seems good, can't see locale select anymore.

myriamboure:
Cool :-) So yes in use case "nothing is done, only default language is there" this work, no language witcher.
screenshot from 2018-03-22 23-50-18
I guess @pacodelaluna  to fully test we want maybe to enable multi-language and add two locals that are not english (like spanish and french) and check that English is not added. But this needs to be done on server side so can't do it myself. @mkllnk if you see that before we have time to do that on French server (it's late already here) please go ahead to try with this other case... before merging :-)

pacodelaluna:
@myriamboure I have enabled ES and FR locales, the logic is working fine but there is still a translation issue.

myriamboure:
Ok so quick image yes we see language switcher BUT we see only English - English
screenshot from 2018-03-23 13-54-07
Why that ?
I checked in Transifex but the translation at least for French is there:
screenshot from 2018-03-23 13-53-19

Also I'm wondering: where should be the translation for "Spanish" if I want to enable French and Spanish on French instance? I only have one `language_name` key in Transifex. @mkllnk can you explain how we can do that? And why in this case one of the two English is not Français as language_name key is properly translated in Transifex?

pacodelaluna:
@myriamboure Ok, my bad, translations are not pulled properly with ofn-install, had to pull them manually, but then it is working properly I guess.

myriamboure:
Ok I deleted my precedent comment to avoid confusion. So now I see the language switcher working properly:
screenshot from 2018-03-23 14-18-51
I notice that language `es` here is translated as "Castellano" but I guess it's fine if every language appear in the corresponding language, will be easier for people speaking those languages to find them an switch languages.
So it seems all good for me, ready to go!

PR #2133, opened 2018-03-12
Fix shipping method creation in sample data

### LINKED ISSUE #2071: Make load_sample_data not dependent on Australia

sauloperez:
#### What? Why?

Closes #2071

Now that we got the fixes in `master` I deployed to staging, tried again and still the shipping method creation failed. I don't understand why but this line wasn't hit before. If not, it would have failed.

Below you have the relevant pieces of the stack trace. 
```
rake aborted!                                                                                                                               
ActiveRecord::RecordInvalid: La validación falló: Provincia ó Estado no puede estar en blanco                                               
/home/openfoodnetwork/.gem/ruby/2.1.0/gems/activerecord-3.2.21/lib/active_record/validations.rb:56:in `save!'                               
(...)
/home/openfoodnetwork/apps/openfoodnetwork/current/spec/factories.rb:209:in `block (3 levels) in <top (required)>'
/home/openfoodnetwork/apps/openfoodnetwork/current/spec/factories.rb:459:in `block (3 levels) in <top (required)>'
(...)
/home/openfoodnetwork/.gem/ruby/2.1.0/gems/factory_girl-3.3.0/lib/factory_girl/strategy_syntax_method_registrar.rb:19:in `block in define_si
ngular_strategy_method'
/home/openfoodnetwork/apps/openfoodnetwork/current/lib/tasks/dev.rake:29:in `block (3 levels) in <top (required)>'
```
That led me to:

https://github.com/openfoodfoundation/openfoodnetwork/blob/6f5080923fe07313f5773ca205cd9def227b213e/spec/factories.rb#L458-L461

which depends on:

https://github.com/openfoodfoundation/openfoodnetwork/blob/6f5080923fe07313f5773ca205cd9def227b213e/spec/factories.rb#L209

so it's not that `shipping_method` requires a country and a state, that's because `Enterprise` needs them, which is created as part of the `shipping_method` factory as a `distributor` (we associate the shipping method with an enterprise).

#### What should we test?

The command `bundle exec rake openfoodnetwork:dev:load_sample_data` should work flawlessly.

#### Release notes

Not worth it.

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -11,6 +11,7 @@ namespace :openfoodnetwork do
 
       spree_user = Spree::User.find_by_email('spree@example.com')
       country = Spree::Country.find_by_iso(ENV.fetch('DEFAULT_COUNTRY_CODE'))
+      state = country.states.first
 
       Spree::MailMethod.create!(
         environment: Rails.env,
@@ -20,10 +21,20 @@ namespace :openfoodnetwork do
       # -- Shipping / payment information
       unless Spree::Zone.find_by_name 'Australia'
         puts "[#{task_name}] Seeding shipping / payment information"
+
         zone = FactoryGirl.create(:zone, name: 'Australia', zone_members: [])
-        country = Spree::Country.find_by_name('Australia')
         Spree::ZoneMember.create(zone: zone, zoneable: country)
-        FactoryGirl.create(:shipping_method, zone: zone)
+        address = FactoryGirl.create(
+          :address,
+          address1: "15/1 Ballantyne Street",
+          zipcode: "3153",
+          city: "Thornbury",
+          country: country,
+          state: state
+        )
+        enterprise = FactoryGirl.create(:enterprise, address: address)
+
+        FactoryGirl.create(:shipping_method, zone: zone, distributors: [enterprise])
       end
 
       # -- Taxonomies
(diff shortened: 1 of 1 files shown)


lib/tasks/dev.rake
```diff
@@ -21,9 +22,8 @@ namespace :openfoodnetwork do
       unless Spree::Zone.find_by_name 'Australia'
         puts "[#{task_name}] Seeding shipping / payment information"
         zone = FactoryGirl.create(:zone, name: 'Australia', zone_members: [])
-        country = Spree::Country.find_by_name('Australia')
```
sauloperez: In Katuma `country.states` was `nil` as it's not our configured country.

lib/tasks/dev.rake
```diff
@@ -20,10 +21,20 @@ namespace :openfoodnetwork do
...
+          :address,
+          address1: "15/1 Ballantyne Street",
+          zipcode: "3153",
+          city: "Thornbury",
+          country: country,
+          state: state
+        )
+        enterprise = FactoryGirl.create(:enterprise, address: address)
+
+        FactoryGirl.create(:shipping_method, zone: zone, distributors: [enterprise])
```
sauloperez: the enterprise didn't have a valid address specified. That was the reason of the failure.

sauloperez:
working both on dev and staging.

lib/tasks/dev.rake
```diff
@@ -20,10 +21,20 @@ namespace :openfoodnetwork do
...
       unless Spree::Zone.find_by_name 'Australia'
         puts "[#{task_name}] Seeding shipping / payment information"
+
         zone = FactoryGirl.create(:zone, name: 'Australia', zone_members: [])
-        country = Spree::Country.find_by_name('Australia')
         Spree::ZoneMember.create(zone: zone, zoneable: country)
-        FactoryGirl.create(:shipping_method, zone: zone)
+        address = FactoryGirl.create(
+          :address,
+          address1: "15/1 Ballantyne Street",
```
enricostano: :heart_eyes:

PR #1981, opened 2017-11-30
Added 'meta_keywords' to search of products

Duende13:
Changed labels on SEO tab and SEO form (edit product Screen) and added in en.yml file.

#### What? Why?

Closes #1936 

It was proposed as solution to be able to look for 'meta_keywords' to avoid that users have to add long names to describe their products and variants, they can add more details into the meta_keywords field, renamed  'Product Search Keywords' to achieve that.

#### What should we test?

To search for products where the meta_keywords or new  'Product Search Keywords' field has been used and applies the filter to find them.

PR #2134, opened 2018-03-12
Added 'meta_keywords' to search of products

### LINKED ISSUE #1936: Using Meta Keywords to Improve Search
The Problem: Currently the Product search box allows searching in the Category, Product Title and Variant Name. This is important as descriptions often contain a lot of additional text we don't want to search for. However, this means that products can end up with long names if we want to be able to search for them but don't want to create a specific category for the product. An example from the UK is craft items like Greeting Cards.

Proposed Solution: To use the SEO fields in the Edit Product screen to do this. I would propose that we use the 'Meta Keyword' field. Thus terms that appear in the 'Meta Keyword' field will be searchable in the Product Search and thus be displayed to the user at that point.

Some fields will need to be renamed for UX:
Tab Name SEO could become 'Search'
'Meta Keyword' could become 'Product Search Keywords'
'Meta Description' could become 'SEO Keywords'.

Fixing the SEO keywords is outside the scope of this issue.
It would be useful to have a tooltip to explain what this is doing.

image

Discussion: https://community.openfoodnetwork.org/t/enabling-meta-keywords-to-help-with-product-search/1102

HugsDaniel:
Copy of #1981 (just added tooltips), **all credits go to @Duende13.**

#### What? Why?

Closes #1936 

It was proposed as solution to be able to look for 'meta_keywords' to avoid that users have to add long names to describe their products and variants, they can add more details into the meta_keywords field, renamed 'Product Search Keywords' to achieve that.

#### What should we test?

To search for products where the meta_keywords or new 'Product Search Keywords' field has been used and applies the filter to find them.

### DIFF
--- a/app/assets/javascripts/darkswarm/filters/filter_products.js.coffee
+++ b/app/assets/javascripts/darkswarm/filters/filter_products.js.coffee
@@ -4,5 +4,5 @@ Darkswarm.filter 'products', (Matcher) ->
     text ?= ""
     return products if text == ""
     products.filter (product) =>
-      propertiesToMatch = [product.name, product.variant_names, product.supplier.name, product.primary_taxon.name]
+      propertiesToMatch = [product.name, product.variant_names, product.supplier.name, product.primary_taxon.name, product.meta_keywords] 
       Matcher.matchBeginning propertiesToMatch, text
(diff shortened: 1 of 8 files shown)


sauloperez:
What should we do stick to #1981 or move to this one?

myriamboure:
It seems @mkllnk added on #1981 @HugsDaniel commits so I guess we will finish there. Hopefully Hugo will be able to finish the work there! So I close this one for now and we can reopen if any issue on finishing the other one. Sorry Hugo I hope you didn't spend too much time on copying files, please test the solution Maikel proposed on the other discussion and we can document that for the next devs facing the same issues as you :-)

PR #2135, opened 2018-03-13
Replace Rohan's email address by optional config

mkllnk:
Rohan's email address is hard-coded, but he doesn't want to receive
these emails any more.

#### What? Why?

Moving to a new server, Rohan was suddenly receiving all these notifications. Our old server didn't have a proper email setup so that these messages were not delivered. The new server does have a working email setup and was sending him lots of notifications every day.

#### What should we test?

Dev's can test that `bundle exec whenever --update-crontab` works correctly. After staging, we should check with `crontab -l` if everything is correct. There is no visible change to the user.

#### Release notes

Admins can now define an email address to notify in case cron jobs have output.

Not.

### DIFF
--- a/config/application.yml.example
+++ b/config/application.yml.example
@@ -16,6 +16,12 @@ CHECKOUT_ZONE: Australia
 # Find currency codes at http://en.wikipedia.org/wiki/ISO_4217.
 CURRENCY: AUD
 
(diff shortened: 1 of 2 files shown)


sstead:
Just did a quick sanity check- the only thing funny is that the product and profile images aren't showing. I'm guessing this won't carry to production? @mkllnk

PR #2148, opened 2018-03-15
Setup mail method automatically

### LINKED ISSUE #2141: Automatically configure mail from seed data
## Description

It's the thing we spend most of our time (and our mood...) on when deploying to staging and when setting up a new instance. Hence, having that done automatically from `db:seed` would be a huge gain.

It's just about doing what is done from the UI on `/admin/mail_methods/new` but without human intervention, which hopefully will also remove the human error.

To do that, we need to provide the domain and the API key as env var and then use that data to populate the spree config as Spree does internally when executing that controller action.

sauloperez:
#### What? Why?

Closes #2141

This ensures that when either resetting your development database, deploying to staging or even setting up a new OFN instance, the mail will be properly configured given some SMTP and other mail settings.

This aims to cut down the enourmous amounts of time we spend setting up environments.

#### What should we test?

When deploying to staging a working mail method should be present in DB. You can try it out by first running `bundle exec rake openfoodnetwork:dev:load_sample_data` and then navigating to `/admin/mail_methods/1/edit`

#### Release notes

New instances OFN will have the mail automatically configured.

#### Dependencies

Depends on https://github.com/openfoodfoundation/ofn-install/pull/134, which should be provisioned before deploying this.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -29,3 +29,24 @@ states.each do |state|
     )
   end
 end
+
+create_mail_method
+
+def create_mail_method
+  Spree::MailMethod.destroy_all
+
+  CreateMailMethod.new(
+    environment: Rails.env,
+    preferred_enable_mail_delivery: true,
+    preferred_mail_host: ENV['MAIL_HOST'],
+    preferred_mail_domain: ENV['MAIL_DOMAIN'],
+    preferred_mail_port: ENV['MAIL_PORT'],
+    preferred_mail_auth_type: 'login',
+    preferred_smtp_username: ENV['SMTP_USERNAME'],
+    preferred_smtp_password: ENV['SMTP_PASSWORD'],
+    preferred_secure_connection_type: 'None',
+    preferred_mails_from: "no-reply@#{ENV['MAIL_DOMAIN']}",
+    preferred_mail_bcc: '',
+    preferred_intercept_email: ''
+  ).call
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -30,23 +30,23 @@ states.each do |state|
   end
 end
 
-create_mail_method
-
 def create_mail_method
   Spree::MailMethod.destroy_all
 
   CreateMailMethod.new(
     environment: Rails.env,
     preferred_enable_mail_delivery: true,
-    preferred_mail_host: ENV['MAIL_HOST'],
-    preferred_mail_domain: ENV['MAIL_DOMAIN'],
-    preferred_mail_port: ENV['MAIL_PORT'],
+    preferred_mail_host: ENV.fetch('MAIL_HOST'),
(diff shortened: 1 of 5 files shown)


daniellemoorhead:
You going to fill in the template @sauloperez? I assumed it was WIP because it wasn't filled in...

sauloperez:
Yes, it was too late at night to find the right words and there's one change missing :trollface:

myriamboure:
Ping @pacodelaluna

db/seeds.rb
```diff
@@ -29,3 +29,24 @@
...
+  CreateMailMethod.new(
+    environment: Rails.env,
+    preferred_enable_mail_delivery: true,
+    preferred_mail_host: ENV.fetch('MAIL_HOST'),
+    preferred_mail_domain: ENV.fetch('MAIL_DOMAIN'),
+    preferred_mail_port: ENV.fetch('MAIL_PORT'),
+    preferred_mail_auth_type: 'login',
+    preferred_smtp_username: ENV.fetch('SMTP_USERNAME'),
+    preferred_smtp_password: ENV.fetch('SMTP_PASSWORD'),
+    preferred_secure_connection_type: 'None',
```
pacodelaluna: Shouldn't we allow a secure option here? I am using TLS actually with my Gmail account. 
Which is the mail system advised by OFN? Couldn't find information from the wiki, I could add it if necessary.

sauloperez: All these settings come from https://github.com/openfoodfoundation/ofn-install/wiki/Sending-emails-from-OFN. Specifically, that depends on the mail provider and it looks like mandrill and others require this.

lib/tasks/dev.rake
```diff
@@ -210,6 +205,5 @@ namespace :openfoodnetwork do
 
       spree_user.confirm!
     end
-
```
pacodelaluna: 👍

db/seeds.rb
sauloperez: If we ever need something different, easy. PR adding another env var.

pacodelaluna: Ok, great, I was looking to the wrong wiki. Then it is all good for me.

spec/services/create_mail_method_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      expect(Spree::MailMethod)
+        .to have_received(:create).with(environment: 'test') { mail_method }
+    end
+
+    it 'updates the MailMethod' do
+      expect(mail_method)
+        .to(receive(:update_attributes)).with(attributes) { mail_method }
+
+      described_class.new(attributes).call
```
danypr92: Why this test structure is different from the others?

sauloperez: that's because this one is actually hitting the DB. I'll try to make it look like the others.

sauloperez:
I believe now the tests are easier to follow @danypr92

PR #2149, opened 2018-03-16
Subs: Remove obsolete subscription fields from order cycles table

oeoeaio:
#### What? Why?

No issue, removes a couple of database fields which are no longer required.

#### What should we test?

Nothing to test really, if the test suite passes we should be fine.

#### Release notes


myriamboure:
Seems the tests are not passing thought :-(
giphy-downsized
Hihihi those GIF are making me happy :-)

Issue #2151, opened 2018-03-17
Variant prices reverting to original value


mkllnk:
A user in Australia reported price changes that they didn't do. The user is the only manager. They said it looks like the price they originally entered for the product when they started using the OFN.

I'm still gathering information about this bug and will fill in this issue as I go.

## Description

Affected enterprise id: 287 - producer
Affected product id: 1395, last updated 2018-03-17 00:36:38 UTC which is 11:36 AEDT.

I downloaded a database backup from this morning and the variant had indeed a price of $8 instead of $19.50. The product had last been changed at 2018-02-19 01:55:59 UTC.

The user reports that order on 1 March were correct. Orders after that date had the wrong price. Pull requests merged around that time:

- https://github.com/openfoodfoundation/openfoodnetwork/pull/2087
- https://github.com/openfoodfoundation/openfoodnetwork/pull/2089
- https://github.com/openfoodfoundation/openfoodnetwork/pull/2088
- https://github.com/openfoodfoundation/openfoodnetwork/pull/2086
- https://github.com/openfoodfoundation/openfoodnetwork/pull/2069

We moved production servers around that time with the final database switch on Saturday 3 March.

Looking at the old database, I found that the affected variant has two prices in the database:
```
openfoodweb_production=> select * from spree_prices where variant_id = 6385;
  id  | variant_id | amount | currency 
------+------------+--------+----------
 8747 |       6385 |  19.50 | AUD
 8746 |       6385 |   8.00 | AUD
(2 rows)
```

This is true for all variants. After the user updated the price to correct it, both prices were the same.

The query generated by ActiveRecord to get the price of a variant is:
```sql
SELECT "spree_prices".* FROM "spree_prices" WHERE "spree_prices"."variant_id" = 6385 AND "spree_prices"."currency" = 'AUD' LIMIT 1
```

That tells me that there should be only one price and it's random which one is retrieved from the database. Random meaning, it depends in which order the database stored the information. This could have changed by dumping and re-importing the database or it could depend on the Postgresql version.

My theory is that moving to a new server, caused this random selection of the two prices to change.

Duplicate price rows have been on the Aus production server in December already. That means that our server change did not cause the duplication, but made it visible.

Affected databases:

- Aus production
- UK production

Unaffected databases:

- Aus staging1
- Canada production
- Katuma

## Expected Behavior

Price should not have changed.

## Actual Behavior

Price reverted to original.

## Steps to Reproduce

[...]

PR #2152, opened 2018-03-18
[WIP] Remove duplicate and conflicting prices

mkllnk:
Fixes https://github.com/openfoodfoundation/openfoodnetwork/issues/2151.

Old OFN servers have two prices per variant with the same currency. But
there should be only one.

For conflicting prices, it chooses the higher price and deletes the
lower price.

I think we should add another migration putting a constraint on `spree_prices` so that this can't happen.

This pull request doesn't touch variant overrides at all.

#### What should we test?

That prices are correct.

#### Release notes

Removes conflicting pricing information from the database. If your database has conflicting information, it's random which information is displayed with this release, it could change. Please verify all prices. So far, we only know of Australia and the UK who have conflicting price information in the database.

It creates two files with backup information with all deleted (duplicate) prices. You may want to delete them at some point.

Should we put them in `tmp` or `log`?

It may help if a future version of Spree enforces unique pricing information per variant and currency.


mkllnk:
@Matt-Yorkley We decided to not include this code into the OFN master branch. You find the code we used to manually repair the database here, but your problem is probably a lot easier to fix.

You can query all affected prices with this SQL:
```sql
SELECT min(id) id1, max(id) id2, variant_id, currency FROM spree_prices GROUP BY variant_id, currency HAVING COUNT(variant_id) > 1
```
If you compare the prices with the prices Rails is currently using, you will probably find that it always selects the lower id or always selects the higher id of the two conflicting prices. You can then delete the other.

oeoeaio:
See comment on #2151

Matt-Yorkley:
Closing as we resolved this nearly a year ago. I just re-checked and the production data is still fine. :+1:

PR #2160, opened 2018-03-20
Fix enterprise filter in customers

sauloperez:
#### What? Why?

When selecting an enterprise from the customers page, the actual value being selected wasn't the one the user picked up. It happened to me that when picking `Enterprise 2` the filter got `Fooderers`, which have the id 2 and 12 respectively.

As a result, when I created a customer, it got associated with `Fooderers` while I wanted `Enterprise 2`.

That was because, as explained in [Angular's filter documentation](https://docs.angularjs.org/api/ng/filter/filter), when providing an Object as the filter expression:

> Object: A pattern object can be used to filter specific properties on objects contained by array. For example {name:"M", phone:"1"} predicate will return an array of items which have property name containing "M" and property phone containing "1".

In my case, the filter was returning:

angular

What we need here instead is an exact match on the attribute `id`.

#### What should we test?

When selecting an enterprise from the customers page, only the customers for that enterprise should be shown, and not others. Also, when creating a new customer, it should be listed when refreshing and selecting her enterprise.

#### Release notes

Fix customers being associated with the wrong enterprise when multiple enterprises are available.

### DIFF
--- a/app/assets/javascripts/admin/customers/controllers/customers_controller.js.coffee
+++ b/app/assets/javascripts/admin/customers/controllers/customers_controller.js.coffee
@@ -12,7 +12,7 @@ angular.module("admin.customers").controller "customersCtrl", ($scope, $q, $filt
 
   $scope.$watch "shop_id", ->
     if $scope.shop_id?
-      CurrentShop.shop = $filter('filter')($scope.shops, {id: $scope.shop_id})[0]
+      CurrentShop.shop = $filter('filter')($scope.shops, {id: parseInt($scope.shop_id)}, true)[0]
       Customers.index({enterprise_id: $scope.shop_id}).then (data) ->
         pendingChanges.removeAll()
         $scope.customers_form.$setPristine()
--- a/spec/javascripts/unit/admin/customers/controllers/customers_controller_spec.js.coffee
+++ b/spec/javascripts/unit/admin/customers/controllers/customers_controller_spec.js.coffee
@@ -10,9 +10,10 @@ describe "CustomersCtrl", ->
       null
 
     shops = [
-      { name: "Shop 1", id: 1 }
-      { name: "Shop 2", id: 2 }
-      { name: "Shop 3", id: 3 }
+      { name: "Shop 1", id: 1 },
+      { name: "Shop 2", id: 12 },
+      { name: "Shop 3", id: 2 },
+      { name: "Shop 4", id: 3 }
     ]
 
     availableCountries = [
@@ -39,9 +40,9 @@ describe "CustomersCtrl", ->
     beforeEach inject (pendingChanges) ->
       spyOn(pendingChanges, "removeAll")
       scope.customers_form = jasmine.createSpyObj('customers_form', ['$setPristine'])
-      http.expectGET('/admin/customers.json?enterprise_id=3').respond 200, customers
+      http.expectGET('/admin/customers.json?enterprise_id=2').respond 200, customers
       scope.$apply ->
-        scope.shop_id = 3
+        scope.shop_id = "2"
       http.flush()
 
     it "sets the CurrentShop", inject (CurrentShop) ->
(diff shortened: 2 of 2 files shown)


myriamboure:
Is there an issue open for that @sauloperez ? This is strange I never saw any problem of customers being visible by another enterprise than the one he is a customer to. If I understand well, it is just that from "customers" you wanted to select an enterprise but finally selected another one by mistake ? Because you filter with the name, not the is so I don't understand how you could hit that issue. Anyway if it's useless just ignore my comment.

myriamboure:
Ah ok didn't read your other message @sauloperez sorry I delete my message :-) ignore!

miquelexposito:
Hi sauloperez,
Have you tried adding a true in the filter condition? Something like that:
$scope.result = $filter('filter')($scope.shops, { id: $scope.shope_id}, true);

sauloperez:
Hi @miquelexposito ! I haven't but I'll give it a try

sauloperez:
I had to read the doc three times but I got it @miquelexposito . Good catch!

sauloperez:
Can you take a look @myriamboure to see if this fixes the issue you encountered?

myriamboure:
Is it staged somewhere @sauloperez ? I can't see if that fixes my issue until I can test it ;-)

sauloperez:
staged on https://staging.katuma.org

myriamboure:
It doesn't seem I'm receiving the confirmation email :-( I've been waiting for some minuntes now, any idea @sauloperez ? Trying to test but blocked.

myriamboure:
I tried to create an account for user myriam[at]ouishare.net

sauloperez:
We ended up buying a new staging. Can you stage this in the meantime @pacodelaluna?

pacodelaluna:
Deployed right now on staging.openfoodfrance.fr

myriamboure:
I'm not able to test as I'm still not receiving confirmation emails :-( @sauloperez @pacodelaluna I think Pau have worked on the email issue, feel free to to talk to understand together what's the problem :-(

sauloperez:
I'm currently working on an issue I spotted with the emails. I hope that's is the issue you're experiencing.

pacodelaluna:
@myriamboure I have fixed emails on french staging, tell me if it is better now.

pacodelaluna:
It was a configuration issue, multiple configuration issues actually. DelayedJob was not getting the good Spree::MailMethod.

myriamboure:
Yes @pacodelaluna ! I received the confirmation email, it was sent from openfoodfrance@gmail.com though, I guess that's another tweek to do ;-) But at least I can test, thank you.

myriamboure:
Oups, not, in fact I receive the email but when I click on confirmation link :-(
screenshot from 2018-03-22 17-56-24

sauloperez:
Is that the right URL? it should be `staging.openfoodfrance.fr` instead of `demo.spreecommerce.com`. What happens if you change it manually?

myriamboure:
Ok so @sauloperez after validating the email the new user still doesn't access the shop BUT his name is still in the hub customer list.
AND I have found another clue: in “orders” I see that for the two people I invited as customers before they register (I tried twice), they have a “cart” status basket as if they had managed to access the shop :-)
SO it’s a bit like if the system believes that they see the shop but the user doesn’t, can’t it be something related only to “view” like Angular stuff? @Matt-Yorkley might have some idea as well...
Testing notes: https://docs.google.com/document/d/1H3J5SuLodfyqoATrYEyj_GtsNwbY3IPAnqWnW91f38A/edit

sauloperez:
Just to be clear, this only fixes the issue where the created customers don't appear in the list.

As for the "cart" status I believe that's the default state regardless if they confirmed or not.

myriamboure:
Ok so I guess this fixes the issue then @sauloperez and is ready to go :-)

PR #2163, opened 2018-03-21
Subs: make sure that a datepicker is available for subscriptions module

### LINKED ISSUE #2155: Subs: clarify format for date and pref. use same datepicker as elsewhere
## Description
When I created my first subscription I didn't know which format to use for "start date". I had to check in the user guide to find the info as it was not written anyway and there were no datepicker as there is elsewhere in the app.

## Expected Behavior
Let's be consistent and use the same datepicker as everywhere.
OR at least specify the format to use like write in grey yyyy-mm-dd

## Actual Behavior
Nothing is written I have no clue in which format I should fill in the date.
screenshot from 2018-03-20 09-07-58

## Steps to Reproduce
1. In "orders" click on "subscriptions" and create a new subscription (make sure you have a schedule created already)
2. You will see that "date" has no indication on format to use

## Context
I was testing subscriptions for the first time and got stuck when filling in the date...

## Severity
NA, new feature being developped.

## Your Environment
- Version used: 1.11

## Possible Fix
- Use same datepicker as elsewhere
OR
- Add a suggestion date in grey like "yyyy-mm-dd" + a tooltip

oeoeaio:
#### What? Why?

Fixes #2155: the datepicker on the subscriptions form was unavailable because the datepicker directive from the (obsolete) `ofn.admin` module is longer available to `admin.subscriptions`. This fixes the issue by including the datepicker module in the `admin.utils` instead.

#### What should we test?

Date elements within the new/edit subscriptions form should show a datepicker to facilitate the selection of a date.

#### Release notes

### DIFF
--- a/app/assets/javascripts/admin/utils/directives/date_picker.js.coffee
+++ b/app/assets/javascripts/admin/utils/directives/date_picker.js.coffee
@@ -1,0 +1,9 @@
+angular.module("admin.utils").directive "datepicker", ->
+  require: "ngModel"
+  link: (scope, element, attrs, ngModel) ->
+    element.datepicker
+      dateFormat: "yy-mm-dd"
+      onSelect: (dateText, inst) ->
+        scope.$apply (scope) ->
+          # Fires ngModel.$parsers
+          ngModel.$setViewValue dateText
--- a/spec/features/admin/subscriptions_spec.rb
+++ b/spec/features/admin/subscriptions_spec.rb
@@ -159,7 +159,10 @@ feature 'Subscriptions' do
         click_button('Next')
         expect(page).to have_content 'can\'t be blank', count: 2
         expect(page).to have_content 'Oops! Please fill in all of the required fields...'
-        fill_in 'begins_at', with: Time.zone.today.strftime('%F')
+        find_field('begins_at').click
+        within(".ui-datepicker-calendar") do
+          find('.ui-datepicker-today').click
+        end
         select2_select card2_option, from: 'credit_card_id'
 
         click_button('Next')


oeoeaio:
@sauloperez We can, but my preferred approach would be to remove the `ofn.admin` module altogether, as it goes against the convention used throughout the rest of the backend, instead of adding it as dependency which we just have to untangle later. It's a case of choosing between the lesser of two evils or a big refactor. I thought duplication was the smallest evil.

sauloperez:
What's that convention? We're talking about frontend here, right?

oeoeaio:
> What's that convention? We're talking about frontend here, right?

Yep, javascript for the admin section. The convention for the admin section is to have all modules built around a particular resource (`admin.products`, `admin.lineItems`, `admin.variantOverrides`, etc). These modules contain domain-specific code which does not need to be shard. All generic directives and services are contained in a few modules that are shared between the domain-specific modules as dependencies (`admin.utils`, `admin.resources`).`ofn.admin` was our original module when we thought it was a good idea to just have a single module for the entire admin section (like we did for darkswarm), but that quickly became impossible to manage because of the sheer amount of code flying around. I have been slowly chipping away at moving things out of it into their own dedicated modules, but the stuff that is left is not touched very often, so I will probably need to actively decide out to refactor it.

myriamboure:
Staged on https://staging.openfoodfrance.fr/

myriamboure:
Ok so I put that back in test ready as either we need to wait for your fix, either someone needs to deploy and enable manually in database subscriptions so that we can test properly.

myriamboure:
@mkllnk can this be the next one staged on Aus staging? With the little tweeks that will enable us to test it properly :-) Thanks!

oeoeaio:
Oops, have now @myriamboure #2213

sstead:
Testing notes @mkllnk ,

- The datepicker is now showing in edit/new subscription. yay!
- I'm noticing some issues with editing a subscription when there's not future OCs lined up for the schedule. It's stopping you from making edits because it sees products that aren't available.
- There seems to be a bug with the Honey subscription. I'm not sure, but maybe it's because that enterprise isn't toggled to allow subscriptions.

mkllnk:
I'm pretty sure these bugs are unrelated to the datepicker. Can you open new issues for that?

PR #2174, opened 2018-03-25
Include option text in dropdown cart even if the variant and product …

### LINKED ISSUE #2169: In dropdown cart product list doesn't display QtyUnity so two variants can look like duplicate
## Description
On dropdown cart if variants have the same name but have different quantities, you can't know which one is which variant in the list (see in the image carrots (first line is 1kg second is 10kg) Fennel (first line is 1 bulb second is 2 bulbs)Squash (first line is 200g second is 400g)
screenshot from 2018-03-23 10-08-18

## Expected Behavior
In dropdown cart all names are displayed with Product name - Variant name (Quantity Unity)

## Actual Behavior
In dropdown cart all only products which have different variant names are displayed with Product name - Variant name (Quantity Unity)

## Steps to Reproduce
1. Make sur you have products with multiple variants, some with various variants names and quantity differences, some with same names and only quantity differences
2. Create an OC
3. As a customer add those items in your cart, and click on the top right corner "cart dropdown quick button"

## Context
I hit that issue while testing PR2030

## Severity
s3 pretty confusing but there is a workaround by changing variant names... but not super intuitive so I suggest we fix quickly nnow.

## Your Environment
- Version used: branch PR2030 tested on Aus staging

## Possible Fix
Apply same name logic that in other listings Product name - variant name (QtyUnty)

stveep:
…have the same name

#### What? Why?

Closes #2169 

The "options text" which shows the format of a variant (e.g. "1 bottle") is only added for cases where the variant and product name differ. This can lead to confusing carts where variants in different formats are listed with the same name.

#### What should we test?

Check that options text appears in brackets after each variant in the dropdown cart. If no options text is provided, nothing should appear.

A potential improvement is to only add this if the cart has multiple variants with the same display name - i.e. only if they need distinguishing. Let me know if this would be useful. It's possible that having the bracketed text after every item isn't necessary or looks really ugly. 

#### Release notes

Variants now show their options text (indicating the format) in the cart dropdown even if they have the same name as the product they are a variant of.

### DIFF
--- a/app/assets/javascripts/darkswarm/services/variants.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/variants.js.coffee
@@ -17,11 +17,11 @@ Darkswarm.factory 'Variants', ->
 
     extendedVariantName: (variant) =>
       if variant.product_name == variant.name_to_display
-        variant.product_name
+        name = variant.product_name
       else
         name =  "#{variant.product_name} - #{variant.name_to_display}"
-        name += " (#{variant.options_text})" if variant.options_text
-        name
+      name += " (#{variant.options_text})" if variant.options_text
+      name
 
     lineItemFor: (variant) ->
       variant: variant
--- a/spec/javascripts/unit/darkswarm/services/variants_spec.js.coffee
+++ b/spec/javascripts/unit/darkswarm/services/variants_spec.js.coffee
@@ -49,6 +49,14 @@ describe 'Variants service', ->
       variant = {product_name: 'product_name', name_to_display: 'product_name'}
       expect(Variants.extendedVariantName(variant)).toEqual "product_name"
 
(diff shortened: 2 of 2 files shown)


myriamboure:
I think it's good to have the option text into bracket whatever variants are in cart, let's keep it simple @stveep :-) Also your tests are not passing :-(

myriamboure:
I'm not "tech" but more "PO" @stveep so I can't really review what you have done but I'm sure others will have a quick look :-)

sstead:
**Testing notes**

Looks good!

- You can now see the option text after every item in the cart and I don' think it looked cluttered.

https://docs.google.com/document/d/1QIVAs_fEHHlHsSurO36S3a_Jgjvt1NCHu9pYVcxbUmA/edit#

PR #2178, opened 2018-03-26
Fixing flexible rate miscalculation

### LINKED ISSUE #2172: "Flexible rate" enterprise fee is not calculated properly
## Description
If you set up an enterprise fee using "flexible rate" calculator, calculations are false.
**Example 1**:
screenshot from 2018-03-23 16-26-22
screenshot from 2018-03-23 16-26-35

**Example 2**:
screenshot from 2018-03-23 16-27-04
screenshot from 2018-03-23 16-27-17

## Expected Behavior
Calculation of "admin&handling" should be : 
- if nb items ordered > max items then cost = (max items - 1) x additionnal item cost + first item cost
- if nb items ordered < or = max items then cost = (nb items ordered - 1) x additionnal item cost + first item cost

## Actual Behavior
I have no idea which calculation is done but it is wrong

## Steps to Reproduce
1. Set up a enterprise fee with "flexible rate" calculator
2. Create an OC and apply that fee in outgoing
3. Make an order

## Context
I hit that while testing PR2079

## Severity
s3 (there is no workaround but no one complained so I guess no one is using it? Don't know... if it is used it would be s2, not sure so putting 2)

## Your Environment
- Version used: branch PR2079 deployed on UK staging

## Possible Fix
Review calculation formula.

HugsDaniel:
#### What? Why?

Closes #2172 

Changed the calculation with what @myriamboure gave in #2172.

#### What should we test?

Calculation is made correctly for both cases (above and below max).

### DIFF
--- a/spec/models/spree/calculator/flexi_rate_spec.rb
+++ b/spec/models/spree/calculator/flexi_rate_spec.rb
@@ -1,18 +1,18 @@
 require 'spec_helper'
 
 describe Spree::Calculator::FlexiRate do
-  let(:calculator) { Spree::Calculator::FlexiRate.new(preferred_first_item: 2, preferred_additional_item: 1) }
+  let(:calculator_first) { Spree::Calculator::FlexiRate.new(preferred_first_item: 2, preferred_additional_item: 1, preferred_max_items: 3.0) }
+  let(:calculator_second) { Spree::Calculator::FlexiRate.new(preferred_first_item: 2, preferred_additional_item: 1, preferred_max_items: 5.0) }
+
   let(:line_item) { instance_double(Spree::LineItem, amount: 10, quantity: 4) }
 
   describe "computing for a single line item" do
-    it "returns the first item rate when above max" do
-      calculator.stub preferred_max_items: 3.0
-      expect(calculator.compute(line_item).round(2)).to eq(4.0)
+    it "returns the first item rate when quantity is above max" do
+      expect(calculator_first.compute(line_item).round(2)).to eq(4.0)
     end
 
-    it "returns the first item rate when below max" do
-      calculator.stub preferred_max_items: 5.0
-      expect(calculator.compute(line_item).round(2)).to eq(5.0)
+    it "returns the first item rate when quantity is below max" do
+      expect(calculator_second.compute(line_item).round(2)).to eq(5.0)
     end
   end
 
--- a/spec/models/spree/calculator/flexi_rate_spec.rb
+++ b/spec/models/spree/calculator/flexi_rate_spec.rb
@@ -1,19 +1,29 @@
 require 'spec_helper'
 
 describe Spree::Calculator::FlexiRate do
-  let(:calculator_first) { Spree::Calculator::FlexiRate.new(preferred_first_item: 2, preferred_additional_item: 1, preferred_max_items: 3.0) }
-  let(:calculator_second) { Spree::Calculator::FlexiRate.new(preferred_first_item: 2, preferred_additional_item: 1, preferred_max_items: 5.0) }
+  let(:line_item) { instance_double(Spree::LineItem, amount: 10, quantity: quantity) }
+  let(:calculator) do
+    Spree::Calculator::FlexiRate.new(
+      preferred_first_item: 2,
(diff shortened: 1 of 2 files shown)


spec/models/spree/calculator/flexi_rate_spec.rb
```diff
@@ -1,13 +1,18 @@
...
-  let(:calculator) { Spree::Calculator::FlexiRate.new }
+  let(:calculator) { Spree::Calculator::FlexiRate.new(preferred_first_item: 2, preferred_additional_item: 1) }
   let(:line_item) { instance_double(Spree::LineItem, amount: 10, quantity: 4) }
 
   describe "computing for a single line item" do
-    it "returns the first item rate" do
-      calculator.stub preferred_first_item: 1.0
-      calculator.compute(line_item).round(2).should == 1.0
+    it "returns the first item rate when above max" do
+      calculator.stub preferred_max_items: 3.0
```
sauloperez: I just noticed this. We shouldn't stub the class under test :red_circle: . Can't we simply build a new calculator instance with this atttribute set to 3.0?

spec/models/spree/calculator/flexi_rate_spec.rb
```diff
@@ -1,13 +1,18 @@
...
+      expect(calculator.compute(line_item).round(2)).to eq(4.0)
+    end
+
+    it "returns the first item rate when below max" do
+      calculator.stub preferred_max_items: 5.0
```
sauloperez: here as well

spec/models/spree/calculator/flexi_rate_spec.rb
HugsDaniel: Sure

spec/models/spree/calculator/flexi_rate_spec.rb
```diff
@@ -1,13 +1,18 @@
...
   describe "computing for a single line item" do
-    it "returns the first item rate" do
-      calculator.stub preferred_first_item: 1.0
-      calculator.compute(line_item).round(2).should == 1.0
+    it "returns the first item rate when quantity is above max" do
+      expect(calculator_first.compute(line_item).round(2)).to eq(4.0)
+    end
+
+    it "returns the first item rate when quantity is below max" do
+      expect(calculator_second.compute(line_item).round(2)).to eq(5.0)
```
enricostano: Just for the sake of playing with rspec DSL I would suggest using `context`:
```ruby
describe Spree::Calculator::FlexiRate do
  let(:line_item) { instance_double(Spree::LineItem, amount: 10, quantity: 4) }
  let(:calculator) do
    Spree::Calculator::FlexiRate.new(
      preferred_first_item: 2,
      preferred_additional_item: 1,
      preferred_max_items: preferred_max_items
    )
  end

  context 'when the calculator has preferred max items set to 3.0' do
   let(:preferred_max_items) { 3.0 }

   it "returns the first item rate when quantity is above max" do
     expect(calculator.compute(line_item).round(2)).to eq(4.0)
   end
  end

  context 'when the calculator has preferred max items set to 4.0' do
   let(:preferred_max_items) { 4.0 }

   it "returns the first item rate when quantity is below max" do
     expect(calculator.compute(line_item).round(2)).to eq(5.0)
   end
  end
end
```
This will give you more "surface" to better describe what's different in each context. Using "first" and "second" doesn't tell much. Actually I would even use a more clear description for the `context` blocks compared to what I suggest, but you'll have a better understanding of the domain.

HugsDaniel:
I slightly modified what I did before and what you suggested @enricostano to fit better the user story : instead of making variations on the `preferred_max_items`, I do it on the `LineItem`'s `quantity`. All green on my side.

sstead:
**Testing notes**

- the flexible rate fees are calculating correctly, across enterprise fees, and shipping fees, and tax is correct.

Good to go

https://docs.google.com/document/d/1tLeFcZH7EHKNzLwrl-YLkkdPA1KLBY91YRYmfTA9LyE/edit?usp=sharing

mkllnk:
First time I saw semaphoreci fail while Travis passed. It's not perfect either (well, our tests are not perfect and the CI server can't fix that).

Issue #2013, opened 2017-12-21
[Spree Upgrade] Refactor all uses of spree_products.count_on_hand


oeoeaio:
Several instances of dead code that use these attributes have been identified as per: #2004 

We need to work out an approach for finding and dealing with other uses of this method.

Likely to be related to #2014

sauloperez:
`spree_products.count_on_hand` is just the aggregation of the `count_on_hand` of the variants of the product:

https://github.com/openfoodfoundation/spree/blob/86bf87f1b1e1b299edc8cd10a2486e44ba0a3987/core/app/models/spree/product.rb#L253-L260

sauloperez:
I just spotted a use of `spree_products.count_on_hand` in:

https://github.com/openfoodfoundation/openfoodnetwork/blob/46d2d6cb8ac9425485cc9635cc5dbc515583151f/app/views/spree/admin/overview/_enterprises_producers_tab.html.haml#L28-L35

That method is defined as follows:
https://github.com/openfoodfoundation/openfoodnetwork/blob/46d2d6cb8ac9425485cc9635cc5dbc515583151f/app/models/enterprise.rb#L206-L208

I wonder if that color change is worth the pain since the method is only used there.

sauloperez:
We overrode the dashboard's controller `Spree::Admin::OverviewController` by means of a `class_eval` in [app/controllers/spree/admin/overview_controller_decorator.rb](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/spree/admin/overview_controller_decorator.rb). There we have custom logic to either render `single_enterprise_dashboard` or `multi_enterprise_dashboard`.

sauloperez:
Given all the problems aforementioned I would simply remove that tab. It's something easy to do but I'd like to check somehow (using analytics?) if someone uses it.

myriamboure:
Hum... @sauloperez I'm wondering one thing, for a producer who only manages their products catalog (as it is used by hubs on OFN) but doesn't manage any hub, that would mean they wouldn't have access to their enterprise from dashboard, am I right? That would be an issue I guess... isn't it possible to just have one tab "my enterprises" and put them  all in if the point is to remove the producer tab? Or is the problem to bring info on active products / products in OC? Cause in that case we can also make the dashboard simpler, I'm not sure those info are so relevant also as you don't know which products are in which OC so for me it brings more confusion than useful info.

sauloperez:
[...]

PR #2183, opened 2018-03-28
Identify uses of Product#count_on_hand by removing the column

sauloperez:
This is just an investigation of https://github.com/openfoodfoundation/openfoodnetwork/issues/2013.

I'm trying to see if removing `spree_products.count_on_hand` the test suite still passes. Depends on https://github.com/coopdevs/spree/tree/remove-products-count-on-hand to do that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem 'i18n-js', '~> 3.0.0'
 gem 'nokogiri', '>= 1.6.7.1'
 
 gem 'pg'
-gem 'spree', github: 'openfoodfoundation/spree', branch: 'step-6a', ref: '86bf87f1b1e1b299edc8cd10a2486e44ba0a3987'
+gem 'spree', github: 'coopdevs/spree', branch: 'remove-products-count-on-hand'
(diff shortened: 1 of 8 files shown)


sauloperez:
The latest Spree migration we have in OFN is `20180405114604_drop_products_count_on_hand.spree.rb`.

sauloperez:
`spree_products.count_on_hand` seems to be used only in the following two enterprise scopes:

https://github.com/openfoodfoundation/openfoodnetwork/blob/21ebb5849c0ce771cd82c7ba2d8b15ef7e630d4f/app/views/spree/admin/overview/_enterprises_producers_tab.html.haml#L28-L44

sauloperez:
Also used in https://github.com/openfoodfoundation/openfoodnetwork/blob/21ebb5849c0ce771cd82c7ba2d8b15ef7e630d4f/spec/controllers/cart_controller_spec.rb#L9-L13

sauloperez:
That seems to be all. No further issues detected by the test suite.

PR #2188, opened 2018-03-28
Calculate price sack fees

### LINKED ISSUE #2173: "Price sack" enterprise fee is not working (not being calculated at all)
## Description
If you set up an enterprise fee using "price sack" calculator, calculations are note happening at all and no fee is charged.
**Example 1:**
screenshot from 2018-03-23 16-45-32
screenshot from 2018-03-23 16-45-47

**Example 2:**
screenshot from 2018-03-23 16-45-58
screenshot from 2018-03-23 16-46-16

## Expected Behavior
Fee should be calculated and displayed in "admin & handling" section.
Formula is:
- if order amount < min amount then cost = normal amount
- if order amount > or = min amount then cost = discount amount

## Actual Behavior
Fee is not calculated and displayed.

## Steps to Reproduce
1. Set up a enterprise fee with "price sack" calculator
2. Create an OC and apply that fee in outgoing
3. Make an order

## Context
I hit that while testing PR2079

## Severity
s3 (there is no workaround but no one complained so I guess no one is using it? Don't know... if it is used it would be s2, not sure so putting 2)

## Your Environment
- Version used: branch PR2079 deployed on UK staging

## Possible Fix
Make sure calculation happens and fees are displayed correctly in cart and checkout.

HugsDaniel:
#### What? Why?

Closes #2173 
Price Sack fees were not calculated.

Much like #2178 the `compute` method returns the right fee depending on the global order amount. I also added `PriceSack` calculator to the `per_order` list, so the calculation is made on the global order amount and not on each item.

#### What should we test?

When an order is created, the right price sack fee is given.

#### Release notes

Price Sack fee is now calculated depending on the order amount and displayed in the cart when an order is made.

### DIFF
--- a/spec/models/enterprise_fee_spec.rb
+++ b/spec/models/enterprise_fee_spec.rb
@@ -22,7 +22,7 @@ describe EnterpriseFee do
       oc = create(:simple_order_cycle, coordinator_fees: [ef])
 
       ef.destroy
-      oc.reload.coordinator_fee_ids.should be_empty
+      expect(oc.reload.coordinator_fee_ids).to be_empty
     end
 
     it "removes itself from order cycle exchange fees when destroyed" do
@@ -30,7 +30,7 @@ describe EnterpriseFee do
       ex = create(:exchange, order_cycle: oc, enterprise_fees: [ef])
 
       ef.destroy
-      ex.reload.exchange_fee_ids.should be_empty
+      expect(ex.reload.exchange_fee_ids).to be_empty
     end
 
     describe "for tax_category" do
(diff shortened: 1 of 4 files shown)


daniellemoorhead:
@sauloperez now you know how @sstead felt trying to do release notes each month 🤣

HugsDaniel:
Tests fail because in tests `Spree::Calculator::PriceSack` is included in `per_item` calculators. @myriamboure isn't it a calculation made on the global order amount ? If it's the case it would seem logical that it is known as a `per_order` calculator.

myriamboure:
Quoted from user guide "Price Sack – This fee is used to charge a discount on fees for orders which exceed a certain dollar amount." so this is definitely a per order calculator @HugsDaniel .

myriamboure:
Seems that now tests are passing, great @HugsDaniel :-) @sauloperez Are you happy with Hugo's modification? Does it answer your requests and is all code good for you to go to test?

spec/models/enterprise_fee_spec.rb
```diff
@@ -69,35 +69,35 @@
       it "does not return fees with FlatRate and FlexiRate calculators" do
```
sauloperez: ⚠️ we need to update this sentence

spec/models/spree/calculator/price_sack_spec.rb
```diff
@@ -9,10 +9,20 @@
...
+  context 'when the order amount is below preferred minimal' do
+    let(:price) { 2 }
+    it "computes with a line item object" do
+      expect(calculator.compute(line_item)).to eq(10)
+    end
+  end
+
+  context 'when the order amount is above preferred minimal' do
+    let(:price) { 6 }
+    it "computes with a line item object" do
```
sauloperez: is this what is actually doing? Isn't applying a discount? `when the order amount is above preferred minimal` > `uses the discount amount`, right?

sstead:
**Testing notes**
- Yes the price sack calculator is working as the user guide described.
- Yes the GST component is applied correctly if required.

https://docs.google.com/document/d/115SwPgh7prjPFCYGEP8mk8T0yG-j0V0n7pbQ0sPzvZI/edit#heading=h.tyc39i58gjyj

PR #2192, opened 2018-04-02
Implementation for USA measuring units

spkprav:
#### What? Why?

Implementation of the USA measuring units.

#### What should we test?

All the units related changes.

### DIFF
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -2,13 +2,12 @@ angular.module("admin.products").factory "VariantUnitManager", ->
   class VariantUnitManager
     @unitNames:
       'weight':
-        1.0: 'g'
-        1000.0: 'kg'
-        1000000.0: 'T'
+        1.0: 'oz'
+        16.0: 'lb'
       'volume':
-        0.001: 'mL'
-        1.0: 'L'
-        1000.0: 'kL'
+        1.0: 'fl oz'
+        16.0: 'pt'
+        128.0: 'gal'
 
     @variantUnitOptions: ->
       options = for unit_type, scale_with_name of @unitNames
--- a/lib/open_food_network/option_value_namer.rb
+++ b/lib/open_food_network/option_value_namer.rb
@@ -61,8 +61,10 @@ module OpenFoodNetwork
     end
 
     def scale_for_unit_value
-      units = {'weight' => {1.0 => 'g', 1000.0 => 'kg', 1000000.0 => 'T'},
-               'volume' => {0.001 => 'mL', 1.0 => 'L',  1000.0 => 'kL'}}
+      units = {
+        'weight' => {1.0 => 'oz', 16.0 => 'lb'},
+        'volume' => {1 => 'fl oz', 16.0 => 'pt',  128.0 => 'gal'}
+      }
 
       # Find the largest available unit where unit_value comes to >= 1 when expressed in it.
       # If there is none available where this is true, use the smallest available unit.


myriamboure:
Hi @spkprav thanks for your contribution! If I'm not misleading I think you work with Jim am I right? It seems your tests are not passing though, can you fix that? If you have any issue in understanding how tests work please ask here and the community will answer you.

@NickWeir63 is it the piece of code that Jim had developed for his project and proposed to contribute to the community with? So this is supposed to fix the issue we have with USA measuring units which are not yet well handled by the OFN. I don't think there an issue open on that so can't connect that PR.

sauloperez:
As long as all measurements in the US instance consistently use imperial units (is that their name?) throughout the app there should be no problem. There won't be any conversions issues since they run their own instance with their own database, or am I missing something?

myriamboure:
I guess that's ok to start with but in a vision of single instance that should be discussed further I guess @sauloperez ... I guess that PR fix a quick issue for US instance but it will need to be overhauled at some point I guess.

sauloperez:
Yes, obviously we'll have many problems to move to a single instance and this one will be one of them. In any case, this will have to be dealt with as we do with dates normally; We store data agnostic of the settings (UTC) and then that data is localized when displayed, according to the user settings.

kirstenalarsen:
i absolutely know nothing about this - but do oz’s and lb’s convert to their higher units by orders of 10? or will they stuff up the automatic g/kg ml/L conversions in the product naming / specifying section? @oeoeaio?

NickWeir63:
Thanks everyone.  Sorry to be late to the party - I have been away for a week.  

Yes this is the PR that the OFN USA team have been waiting for.   

Yes there are 12 ounces to the pound and 8 pints to the gallon.  Please let me know if/how I can help towards getting this merged and deployed - the USA users will be very happy to see it.  

Thank you.

sauloperez:
We need to make these strings translatable by adding new keys in `config/locales/en.yml` and then adding the proper translation for each of the supported languages, including `en_US.yml`.

rbarreca:
I am new to the OFN codebase, but adding new units of measure (UOM) as translations of other UOMs seems problematic to me. USA markets sometimes sell products by the liter and gram as well.  Also, conflating units would create problems quantifying things in the future. 

To be forward looking, a slightly more resilient (but still small) step forward could be to add lb, oz, gal, etc. (instead of replace) and then have a map where we could filter which UOMs a particular market decides to make visible. 

Note, there are 16 oz in a lb, not 12.

NickWeir63:
Sorry I said earlier that there were 12 ounces to the pound.  This was wrong - there are 16 ounces to the pound.  Now corrected above.  My bad

myriamboure:
So what shall we do about this PR? The need is a new feature to be built (enable new measuring units). It seems to me that we need a proper inception to agree on how to tackle the need for US measuring units in a more comprehensive and consistent way. But this need has not been prioritized in the last curation meeting for period April - June. What I would suggest is open an icebox item, start discussing feature candidates and hopefully agreeing on one (maybe the one @rbarreca suggested), close that PR as it's not answering consistently the need, and then decide to prioritize the icebox item sooner rather than later? Any other suggestion @daniellemoorhead @kirstenalarsen ? Let's just keep in mind that the Spree upgrade is our top one priority... maybe that could be a first issue @rbarreca will be willing to work on and a good way to enter OFN curation process :-)

NickWeir63:
Thanks @myriamboure yes that sounds like a good plan.  I know that the USA team are keen to see this done asap but I appreciate that there is a queue.  Let me know if I can do anything to help

rbarreca:
I think the @myriamboure plan to close this PR and put an issue in the queue is the right one. As @sauloperez said, this will change all other OFN instances to USA's measuring units. But, I think doing unit translations will be problematic. So really, I think we need each instance to be able to enable/disable the units it wants to use.

myriamboure:
Ok so closing that PR for now, I created the following icebox: https://community.openfoodnetwork.org/t/hubs-managers-can-choose-the-adapted-weight-and-measure-units-for-their-shops-given-their-own-local-situation/1289 we can decide to prioritize in next curation meeting (or discuss of a plan if we want to prioritize before, we don't have exception process at the moment to include new things in the roadmap. Can be discussed at next global hangout).

PR #2200, opened 2018-04-09
Ofn 985

### LINKED ISSUE #985: When shopfront displays multiple property filters please replace '&' with 'OR'
On the screenshot below I can see products that are produced within 15 miles and I can see products that are certified biodynamic.  However most of these products do not fit BOTH these properties.  It would be great if the message top left could be changed from
Showing: Certified Biodynamic & Produced within 15 miles
to
Showing: Certified Biodynamic **OR** Produced within 15 miles
Ideally with the '**OR**' in bold, capitals, underlined, whatever to emphasise it
Thanks
screenshot 2016-05-07 15 33 06

yorkfrog:
#### What? Why?
Temporary PR to help diagnose failing tests - see Slack https://openfoodnetwork.slack.com/archives/C2GQ45KNU/p1523281248000592

Closes #985 

[Explain why is this change needed and the solution you propose. Provide
context for others to understand it]

#### What should we test?

[List which features should be tested and how]

#### Release notes

[In case this should be present in the release notes, please write them or
remove this section otherwise]

[Any known conflicts with the Spree Upgrade? explain them or remove this section
otherwise]

#### Discourse thread

[Is there a discussion about this in Discourse? add the link if so or remove
this section otherwise]

#### Dependencies

[Does this PR depend on another one? add the link of so or remove this section
otherwise]


yorkfrog:
issue resolved locally

Issue #2205, opened 2018-04-11
Refactor custom `available_to_order?` logic


sauloperez:
## Description

It turns out that `ShippingMethod#available_to_order?` is gone in Spree 2.0 and we rely on it to add our custom logic.

### Current status 

On top of the Spree's checks to know whether a shipping method is eligible to be used for an order, we also check that the order's distributor is one of the enterprises that can use the shipping method.

In other words, while Spree defined it like:

```ruby
def available_to_order?(order)
  available?(order) &&
  within_zone?(order) &&
  category_match?(order) &&
  currency_match?(order)
end
```

We extended it with:

https://github.com/openfoodfoundation/openfoodnetwork/blob/236ba9328d3acc99e26f265f97441439149d3252/app/models/spree/shipping_method_decorator.rb#L45-L49

### How does it look in Spree 2.0?

In https://github.com/spree/spree/commit/e1ca57421ce2efb8ee4715bda448158af5b7e90f `available_to_order?`'s logic is moved to [Spree::StockEstimator#shipping_methods(package)](https://github.com/spree/spree/commit/e1ca57421ce2efb8ee4715bda448158af5b7e90f#diff-c917ee5e7ac233c8f9f9d36c44854e15R29), so we need that method to also take our logic into account.

As long as we have investigated some needs have been identified:

* We'll need to a use a default `ShippingCategory` at instance level since [shipping methods are now fetched through the categories](https://github.com/spree/spree/blob/2-0-stable/core/app/models/spree/stock/package.rb#L84-L86) and [products now must specify a shipping category](https://github.com/spree/spree/blob/2-0-stable/core/app/models/spree/product.rb#L72). So far, we haven't used categories at all because [Spree allowed them to be nil](https://github.com/spree/spree/blob/1-3-stable/core/app/models/spree/shipping_method.rb#L44).

HugsDaniel:
Somewhat related https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/overrides/spree/admin/shipping_methods/_form/replace_form_fields.html.haml.deface#L37

enricostano:
Also maybe related https://github.com/openfoodfoundation/openfoodnetwork/issues/2126

RachL:
Also maybe https://github.com/openfoodfoundation/openfoodnetwork/pull/2039#issuecomment-375137267

sauloperez:
[...]

PR #2209, opened 2018-04-12
Spree upgrade

HugsDaniel:
#### What? Why?

First attempt to boot the app with Spree 2.0.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,17 +11,17 @@ gem 'i18n-js', '~> 3.0.0'
 gem 'nokogiri', '>= 1.6.7.1'
 
 gem 'pg'
-gem 'spree', github: 'openfoodfoundation/spree', branch: 'step-6a', ref: '86bf87f1b1e1b299edc8cd10a2486e44ba0a3987'
+gem 'spree', '2.0.13'
 gem 'spree_i18n', github: 'spree/spree_i18n', branch: '1-3-stable'
-gem 'spree_auth_devise', github: 'openfoodfoundation/spree_auth_devise', branch: 'spree-upgrade-intermediate'
+gem 'spree_auth_devise', github: 'spree/spree_auth_devise', branch: '2-0-stable'
 
 # Our branch contains two changes
 # - Pass customer email and phone number to PayPal (merged to upstream master)
 # - Change type of password from string to password to hide it in the form
-gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "spree-upgrade-intermediate"
-#gem 'spree_paypal_express', github: "spree-contrib/better_spree_paypal_express", branch: "1-3-stable"
+# gem 'spree_paypal_express'
+gem 'spree_paypal_express', github: "spree-contrib/better_spree_paypal_express", branch: "2-0-stable"
 gem 'stripe', '~> 3.3.1'
-gem 'activemerchant', '~> 1.71.0'
+gem 'activemerchant'
 
 gem 'oauth2', '~> 1.2.0' # Used for Stripe Connect
 gem 'jwt', '~> 1.5'
@@ -49,8 +49,8 @@ gem 'representative_view'
 gem 'rabl'
 gem "active_model_serializers"
 gem 'oj'
-gem 'deface', github: 'spree/deface', ref: '1110a13'
-gem 'paperclip'
+gem 'deface', '1.0.0'
+gem 'paperclip', '~> 3.4.1'
 gem 'dalli'
 gem 'geocoder'
 gem 'gmaps4rails'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -11,17 +11,17 @@ gem 'i18n-js', '~> 3.0.0'
...
 gem 'spree_i18n', github: 'spree/spree_i18n', branch: '1-3-stable'
-gem 'spree_auth_devise', github: 'openfoodfoundation/spree_auth_devise', branch: 'spree-upgrade-intermediate'
+gem 'spree_auth_devise', github: 'spree/spree_auth_devise', branch: '2-0-stable'
 
 # Our branch contains two changes
 # - Pass customer email and phone number to PayPal (merged to upstream master)
 # - Change type of password from string to password to hide it in the form
-gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "spree-upgrade-intermediate"
-#gem 'spree_paypal_express', github: "spree-contrib/better_spree_paypal_express", branch: "1-3-stable"
+# gem 'spree_paypal_express'
```
sauloperez: I'd remove this line. If it's commented out it's because we don't need it. If we ever do, Git will have in its history.

app/controllers/base_controller.rb
```diff
@@ -37,7 +37,7 @@ def set_order_cycles
 
     # And default to the only order cycle if there's only the one
     if @order_cycles.count == 1
-      current_order(true).set_order_cycle! @order_cycles.first
+      current_order( create_order_if_necessary: true ).set_order_cycle! @order_cycles.first
```
sauloperez: no need for this extra spacing inside the parenthesis.

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -18,7 +18,7 @@
 
   # Patching to redirect to shop if order is empty
   def edit
-    @order = current_order(true)
+    @order = current_order( create_order_if_necessary: true )
```
sauloperez: here as well

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -78,7 +78,7 @@ def populate
     # callbacks on Spree::Adjustment and then manually invoke Spree::Order#update! on success.
 
     Spree::Adjustment.without_callbacks do
-      populator = Spree::OrderPopulator.new(current_order(true), current_currency)
+      populator = Spree::OrderPopulator.new(current_order( create_order_if_necessary: true ), current_currency)
```
sauloperez: here too

enricostano:
@mkllnk yes, during the Spree upgrade (but not only) we constantly struggle to find the less disruptive strategy and the "wrapper" thing is something we're taking in account all the time. The idea is to take advantage of the Spree upgrade to define our own OFN business logic abstraction and then internally we can plug into Spree or make our own thing.

In any case, yes. This PR could have included that OFN abstraction but in this specific case we didn't find a better reason than possible merge conflicts. But we'll need to question ourselves in any PR and ask other devs how to achieve it.

sauloperez:
Sorry @mkllnk, I don't understand either what you mean with a wrapper nor the explanation about the two parts. 

I'm spending a bit of time while working on https://github.com/openfoodfoundation/openfoodnetwork/issues/2205 to find out a way to make this whole dependency on Spree a bit easier to manage.

mkllnk:
@sauloperez I meant that we are changing a lot of calls to Spree's `current_order` here. My idea was to create our own version of `current_order` and name it slightly differently. Then we call our own method everywhere. We can do this change in the master branch to separate our code from Spree's code a bit more. In the Spree upgrade, we just need to change our implementation of our `current_order` method in one place without changing all the other files. That would prevent merge conflicts. With this way of encapsulation it's easier to deal with API changes in Spree.

kirstenalarsen:
separate our code - YEAH!!!

sauloperez:
that's an interesting idea @mkllnk. That's essentially implementing an [adapter](https://en.wikipedia.org/wiki/Adapter_pattern). I haven't considered the nice side effect of cutting down the possibility of merge conflict, which I believe it's a major win.

This might not be the place to have an in-depth discussion about OOP design but I've been considering the possibility of building our own `extendible_spree` gem (hosted in rubygems) to implement well-designed hooks that allow extending it in a clean way, like we need in OFN, as opposed to the hackish and unmaintainable `class_eval` approach they favour. 

I feel like we need a series of design/architecture session to better address the spree upgrade.

daniellemoorhead:
@sauloperez I say set up a call and invite all the people (aka @oeoeaio @mkllnk alongside of you guys working on Spree already). I think it's super important, and worth running potentially fortnightly catchups with our timezone to talk through things. IMO 😀

sauloperez:
totally agree @daniellemoorhead . I want to prepare first a code proposal to base our discussions on.

PR #2138, opened 2018-03-13
Email/manager invite

Matt-Yorkley:
#### What? Why?

Closes #1593. 

Adds the ability to invite unregistered users to become managers of an enterprise from the enterprise users page.

#### What should we test?

Invite a new (unregistered) enterprise user from the enterprise users page. Check the emails that get sent to the new user, and make sure the invited user gets directed to set a password.

#### Release notes

Unregistered users can now be invited to become enterprise managers.

#### Screenshots

screenshot from 2018-03-13 16-57-38

screenshot from 2018-03-13 19-20-03

screenshot from 2018-03-13 19-18-47

PR #2220, opened 2018-04-14
Add cancan permissions for Admin::ManagerInvitationsController

### LINKED ISSUE #2197: Invite button not clickable when Inviting new Manager
## Description
When I use the Invite Manager field in Aus production, the box opens, I type the email of the manager I want to invite, but when I click the Invite button nothing happens. It works for super-admin users.

## Expected Behavior
The invite button should become clickable after I enter an email, and after I click it I should see a success/error message.

## Steps to Reproduce

1. Log in as enterprise user (not super admin).
1. Go to the Users tab of a profile you own. 
2. Click the invite manager button and enter an email, click Invite.

## Animated Gif/Screenshot
https://www.useloom.com/share/28d04bfcb06641079936c2f41790cb4c

## Context
I can't add managers.

## Severity
S3

## Your Environment
PC, Chrome.

## Possible fix

The POST request to invite a manager causes a 403 forbidden response. We probably need to tweak authorisations. See `app/models/spree/ability_decorator.rb`. This never worked. The issue was already there in the original feature introduced in 82457d2b632923118ac4aed171272a628ff3ed83.

Matt-Yorkley:
#### What? Why?

Closes #2197

Minor amendment to previous PR, the implementation was changed a bit during code review, but the cancan permissions hadn't been updated to reflect that, which meant the manager invite wasn't working for non-superadmins.

#### What should we test?

Manager invites should now work correctly for non-superadmins.

#### Release notes

Manager invitation bugfix for #2197.

### DIFF
--- a/app/controllers/admin/manager_invitations_controller.rb
+++ b/app/controllers/admin/manager_invitations_controller.rb
@@ -1,5 +1,7 @@
 module Admin
   class ManagerInvitationsController < Spree::Admin::BaseController
+    authorize_resource class: false
+
     def create
       @email = params[:email]
       @enterprise = Enterprise.find(params[:enterprise_id])
--- a/app/models/spree/ability_decorator.rb
+++ b/app/models/spree/ability_decorator.rb
@@ -127,6 +127,8 @@ class AbilityDecorator
     can [:admin, :connect, :status, :destroy], StripeAccount do |stripe_account|
       user.enterprises.include? stripe_account.enterprise
     end
+
+    can [:admin, :create], :manager_invitation
   end
 
   def add_product_management_abilities(user)
--- a/spec/controllers/admin/manager_invitations_controller_spec.rb
+++ b/spec/controllers/admin/manager_invitations_controller_spec.rb
@@ -2,8 +2,11 @@ require 'spec_helper'
 
 module Admin
   describe ManagerInvitationsController, type: :controller do
+    let!(:enterprise_owner) { create(:user) }
+    let!(:other_enterprise_user) { create(:user) }
     let!(:existing_user) { create(:user) }
-    let!(:enterprise) { create(:enterprise) }
+    let!(:enterprise) { create(:enterprise, owner: enterprise_owner ) }
+    let!(:enterprise2) { create(:enterprise, owner: other_enterprise_user ) }
     let(:admin) { create(:admin_user) }
 
     describe "#create" do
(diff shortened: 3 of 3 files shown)


myriamboure:
Can someone else quickly review that please so that we can test and merge? :point_up: @mkllnk or @HugsDaniel maybe?

app/controllers/admin/manager_invitations_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Admin
   class ManagerInvitationsController < Spree::Admin::BaseController
+    authorize_resource class: false
```
mkllnk: This is the first time I see a line like this. What does it mean? I found https://github.com/ryanb/cancan/wiki/authorizing-controller-actions#authorize_resource explaining that it adds a before filter like `authorize!(:create, @manager_invitation || false)`. Why did you add `class: false`?

And why do we need two authorisations? One in this filter and the other one in the create method.

enricostano:
> the implementation was changed a bit during code review, but the cancan permissions hadn't been updated to reflect that, which meant the manager invite wasn't working for non-superadmins.

@Matt-Yorkley do we have tests covering that behavior? Could you please add them? And in any case to the controller, right? Thanks!

app/controllers/admin/manager_invitations_controller.rb
Matt-Yorkley: `authorize_class: false` seems to be the recommended way to use cancan in controllers that don't have a corresponding model? It tells cancan to look for symbol keys in the permissions list instead of a named class, and I think it skips some of the default behaviour like trying to instantiate a corresponding object on create actions, amongst others.

The additional permissions check inside the create action makes sure the user has permission to make changes to the specific enterprise in question (after we've pulled up that enterprise record) and not just general permission to access the create manager invites functions.

Matt-Yorkley: https://github.com/ryanb/cancan/wiki/Non-RESTful-Controllers#alternative-authorize_resource

Matt-Yorkley:
@enricostano There's specs in the original PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/2138. 

This PR doesn't change the functionality, it just fixes the missing permissions for non-superadmins.

Matt-Yorkley:
Or do you want an extra spec to check the permissions? Maybe that would be useful.

enricostano:
> Or do you want an extra spec to check the permissions? Maybe that would be useful.

@Matt-Yorkley that would be great

mkllnk:

sstead:
Testing notes @mkllnk 
- Both superadmin and enterprise owners can now invite new users to be managers
- There's one bug - when the new users confirms their email, they're taken to set a pw, and they get an error - 'password reset toek expired' . this is already on production, so not sure if it's reason to not push this PR.

https://docs.google.com/document/d/1P5nop3zhJgy0NTlrb9-CSr5pYdDcazNucr8GBGqx1iE/edit?usp=sharing

sstead:
I've created this new issue for the 'reset password token expired' problem https://github.com/openfoodfoundation/openfoodnetwork/issues/2254

sauloperez:
> Release notes

> Manager invitation bugfix for #2197.

They should be useful for humans :trollface: Consider describing the nature of the fix, please.

PR #2231, opened 2018-04-17
Add image to be used in Spree upgrade wiki page

sauloperez:
#### What? Why?

It turns out that Github's wiki pages do not allow uploading images but they do render image URLs. See https://help.github.com/articles/adding-images-to-wikis/

I believe `docs/` it's the best place for them.


daniellemoorhead:
This lack of image upload in the wiki has been a pain in the butt a few times for me @sauloperez - can you tell me how us non-devs would be able to use what you've done here?

kirstenalarsen:
Could run a gitbook for the wiki? just link across from the github wiki? Devs can update via github easily, and others can also use via gitbook, and then images will be fine?

Would be a job to transfer it though . .

sauloperez:
There's no upload but you can still link to images :point_down: 

toolbar
insert

So it's just a matter of hosting these images somewhere publicly linkable.

> Could run a gitbook for the wiki? just link across from the github wiki? Devs can update via github easily, and others can also use via gitbook, and then images will be fine?

I think this just does the job and I prefer to keep these dev docs as close to the codebase as possible. Also because the wiki is kind of the default place where most devs look at.

PR #2232, opened 2018-04-17
Disable totally unreliable feature test

sauloperez:
#### What? Why?

We need to investigate why it fails so many times, fix it and then enable it back. As it is, it brings cons than pros preventing even PRs that don't touch code from being merged.

This happened all of a sudden. Most of the PRs I've opened lately take few builds even on Semaphore CI to pass.

https://github.com/openfoodfoundation/openfoodnetwork/issues/2233 has been created to keep track of it and eventually solve the underlying issue.

### DIFF
--- a/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
+++ b/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
@@ -47,7 +47,7 @@ feature "Using embedded shopfront functionality", js: true do
       end
     end
 
-    it "allows shopping and checkout" do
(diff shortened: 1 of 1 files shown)


sauloperez:
This will unblock some pull requests blocked by this test.

Matt-Yorkley:
If it's slowing down development by being consistently flaky, I agree we should set it to `pending` for now and take a look at it. In development it always passes when I try it.

sauloperez:
Could you take a look at it if you know anything about it?

Issue #2219, opened 2018-04-13
Upgrade dependencies to make OFN work with Spree 2.0


sauloperez:
## Description

We need to upgrade some dependencies required by Spree to make the app work with Spree 2.0. This involves investigating these dependencies and changing the Gemfile so that Bundler can resolve them all.

This is the first step towards being able to boot the app with Spree 2.0.

sauloperez:
Solved by #2209

Issue #1957, opened 2017-11-14
Remove references to MailMethod when in Spree 2.0


sauloperez:
## What

In Spree 2.0 the `MailMethod` model no longer exists. See https://guides.spreecommerce.org/release_notes/spree_2_0_0.html#mailmethod-model-no-longer-exists.

We should remove all references to it, most of which are in the specs. This is what a `git grep` returns now:

```
app/mailers/spree/base_mailer_decorator.rb:    Spree::MailMethod.current.andand.preferred_mails_from ||
db/migrate/20120327000613_create_mail_methods.rb:class CreateMailMethods < ActiveRecord::Migration
lib/tasks/dev.rake:      Spree::MailMethod.create!(
spec/controllers/spree/admin/orders_controller_spec.rb:            Spree::MailMethod.create!(
spec/controllers/spree/orders_controller_spec.rb:          Spree::MailMethod.create!(
spec/features/consumer/shopping/orders_spec.rb:        Spree::MailMethod.create!(
spec/mailers/order_mailer_spec.rb:    Spree::MailMethod.create!(
spec/mailers/producer_mailer_spec.rb:    Spree::MailMethod.create!(
spec/models/order_cycle_spec.rb:      Spree::MailMethod.create!(
spec/models/spree/order_spec.rb:        Spree::MailMethod.create!(
```

daniellemoorhead:
Can this be closed now that the PR is merged @sauloperez

oeoeaio:
@daniellemoorhead no, this issue is in place of the code that was removed in the PR. This issue is blocked until we reach Spree 2.0.

sauloperez:
Moved to a Trello card for future reference: https://trello.com/c/cSVtpsns/14-remove-references-to-mailmethod

sauloperez:
![](https://media.giphy.com/media/w2fjYsGvjDF3a/giphy.gif)

sauloperez:
Bingo! there's a migration removing the `spree_mail_methods` table in https://github.com/spree/spree/commit/fbb95af890

HugsDaniel:
![](https://github.trello.services/images/mini-trello-icon.png) [Issue](https://trello.com/c/AbUpeztg/1-issue)

luisramos0:
I see the table is only dropped on 2.0.5. But mail_method model is done in our 2-0-4 branch.
This means we can implement this now, right?

luisramos0:
[...]

PR #2235, opened 2018-04-18
Upgrade Spree to 2.0.4 instead of 2.0.13

sauloperez:
#### What? Why?

Related to #2219 

Although we don't know how big the fixes introduced between 2.0.4 and 2.0.13 are, jumping from roughly 1.3.99 to 2.0.13 seems too big of a change to do in one go.

Unless more issues like this one come up that force us to upgrade to a newer version I believe it's best to be conservative. You'll find an in-depth explanation in https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-2.0-upgrade#getting-to-version-204-

A previous effort was done to upgrade to v2.0.13 in https://github.com/openfoodfoundation/openfoodnetwork/pull/2209


sauloperez:
Does it make sense @HugsDaniel @enricostano @mkllnk ?

mkllnk:
Makes sense, thank you. The smaller the steps, the better.

enricostano:
@sauloperez this is you talking about `2.0.5`... https://github.com/openfoodfoundation/openfoodnetwork/issues/1957#issuecomment-374953178

:joy:

sauloperez:
We can live with that. That comment is talking about the removal of the `spree_mail_methods` table. We can keep it but unused and remove it later in the next upgrade.

PR #2238, opened 2018-04-18
Disable password reset for unconfirmed users

### LINKED ISSUE #2186: If unconfirmed user request new password, they can proceed and unfriendly UX happens.
## Description
You shouldn’t be able to reset your password until you have confirmed your email.

## Expected Behavior
If you try to reset password and email is not confirmed system should say “We can’t reset your password until you have validated your email. Resend confirmation link.” and when click on confirmation link should be sent back to the “forgot password” modal where he was before clicking on "resend confirmation link" (as he still doesn’t remember his password I guess…).

## Actual Behavior
If you try to reset password and email is not confirmed, the system let you do so, send you a reset password link, and you can set up a new password even if you don't have confirmed your email. So then when user tries to login with new password he got again an error message saying "you need to confirm your email first. Resend email confirmation".

[note: we could do it the other way around, let the user change his password even if he has not confirmed his email but it seems really unfriendly on a UX prospective, I shouldn't be able to do any action until I have confirmed my email.]

## Steps to Reproduce
1. Sign up BUT don't click on confirmation link.
2. Then go to "reset password" (in login modal) and ask to reset password with the same user email.
3. You receive a "reset password" email, click on the link, set up new password.
4. Then try to login, you get the error message asking you to confirm your email.

## Animated Gif/Screenshot
https://www.useloom.com/share/0cc1125d639a43e28b63bf7115c2abaa

Error message that displays after new password is set up:
screenshot from 2018-03-28 15-48-19

## Context
I hit that while testing PR2073

## Severity
s3 UX unfriendly but part of "email confirmation" epic so will go directly to delivery backlog.

## Your Environment
- Version used: branch PR2073 deployed on Aus staging

## Possible Fix
See expected behavior.

### LINKED ISSUE #2237: Update redirect URL when resending email confirmation
## Description

When sending email confirmations, the page the user is on gets stored so they can be redirected to the page they came from after clicking the validation link. This does not get updated when a confirmation email is re-sent.

## Expected Behavior

Resending email confirmations should update the return URL.

## Actual Behaviour

The return URL stays the same as it was when the user initially signed up.

## Steps to Reproduce
1. Sign up and receive confirmation email
2. Click resend email from a different page

## Context

Clicking validation links form re-sent confirmation emails.

Matt-Yorkley:
#### What? Why?

Closes #2186. User who have not confirmed their email cannot reset their password. If they try to use the password reset, a warning message is shown, including a resend confirmation link.

Also closes #2237. When re-sending email confirmations, the `:return_url` value is updated, so that the redirect after clicking the validation link will return the user to the most recent page from which they requested the validation email, instead of the page they were on when they first signed up.

#### What should we test?

1. Resetting passwords with unconfirmed users. 
2. Resending confirmation emails from a different page then the initial signup.

#### Release notes

Users cannot use password reset until their email address is confirmed. Confirmation emails will redirect to the most recent page from which the email was requested when re-sending confirmation emails.

### DIFF
--- a/app/controllers/user_passwords_controller.rb
+++ b/app/controllers/user_passwords_controller.rb
@@ -4,7 +4,7 @@ class UserPasswordsController < Spree::UserPasswordsController
   before_filter :set_admin_redirect, only: :edit
 
   def create
-    return if user_unconfirmed?
+    render_unconfirmed_response && return if user_unconfirmed?
 
     self.resource = resource_class.send_reset_password_instructions(params[resource_name])
 
@@ -29,12 +29,12 @@ class UserPasswordsController < Spree::UserPasswordsController
     session["spree_user_return_to"] = params[:return_to] if params[:return_to]
   end
 
+  def render_unconfirmed_response
+    render json: { error: t('email_unconfirmed') }, status: :unauthorized
+  end
+
   def user_unconfirmed?
     user = Spree::User.find_by_email(params[:spree_user][:email])
-    if user && !user.confirmed?
-      render json: { error: t('email_unconfirmed') }, status: :unauthorized
-    end
-
     user && !user.confirmed?
   end
 end
--- a/app/controllers/user_passwords_controller.rb
+++ b/app/controllers/user_passwords_controller.rb
@@ -4,6 +4,8 @@ class UserPasswordsController < Spree::UserPasswordsController
   before_filter :set_admin_redirect, only: :edit
 
   def create
+    return if user_unconfirmed?
+
     self.resource = resource_class.send_reset_password_instructions(params[resource_name])
 
     if resource.errors.empty?
@@ -15,7 +17,7 @@ class UserPasswordsController < Spree::UserPasswordsController
           respond_with_navigational(resource) { render :new }
         end
         format.js do
-          render json: resource.errors, status: :unauthorized
+          render json: { error: t('email_not_found') }, status: :not_found
         end
       end
     end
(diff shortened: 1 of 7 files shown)


app/controllers/user_passwords_controller.rb
```diff
@@ -26,4 +28,13 @@ def create
...
   end
+
+  def user_unconfirmed?
+    user = Spree::User.find_by_email(params[:spree_user][:email])
+    if user && !user.confirmed?
+      render json: { error: t('email_unconfirmed') }, status: :unauthorized
+    end
+
+    user && !user.confirmed?
+  end
```
sauloperez: I find this method a bit confusing due to the fact that is named as a boolean method but it also renders a JSON, which is a bit unexpected. What about extracting the render out of this method?

app/controllers/user_passwords_controller.rb
```diff
@@ -15,7 +17,7 @@ def create
           respond_with_navigational(resource) { render :new }
         end
         format.js do
-          render json: resource.errors, status: :unauthorized
+          render json: { error: t('email_not_found') }, status: :not_found
```
sauloperez: here we're implicitly giving for granted that the only error state is not finding the email, is actually that the only case?

Matt-Yorkley: Yes, the original code was only responding to that one error, and the views for this controller were hard-coded to respond with the "not found" error message regardless of the error response, but that wasn't previously clear in the controller.

oeoeaio:
I tend to agree with @mkllnk, I think that would be better UX, but let's get this through and then we can think about a future iteration if we think it is necessary.

oeoeaio:
*REBASED* and staged on: https://staging1.openfood.com.au

sstead:
**Testing notes**

-Yes, unconfirmed users are prevented from resetting their password, in all cases.

https://docs.google.com/document/d/1GlXv42WssEeqDUPZx93mYlrvvCRmCL2NU0DLKi5vsds/edit?usp=sharing

oeoeaio:
Thanks @Matt-Yorkley!

PR #2246, opened 2018-04-23
Autofix comment indentation

sauloperez:
#### What? Why?

Autocorrect Rubocop's Layout/CommentIndentation cop. I also took the chance to update the `rubocop_todo.yml`.

Something I found concerning is that most of the updated counts in there are increased. Something is not working properly at least in Code Climate. We should see a red check for the PRs that introduce these violations. Besides, these violations should be fixed by devs themselves as soon as they appear. I urge you to integrate Rubocop with your editor.


Guardfile
```diff
@@ -5,45 +5,7 @@ guard 'livereload' do
...
-  #watch('config/routes.rb')                           { "spec/routing" }
-  #watch('app/controllers/application_controller.rb')  { "spec/controllers" }
-
-  ## Capybara features specs
-  #watch(%r{^app/views/(.+)/.*\.(erb|haml|slim)$})     { |m| "spec/features/#{m[1]}_spec.rb" }
-
-  ## Turnip features and steps
-  #watch(%r{^spec/acceptance/(.+)\.feature$})
-  #watch(%r{^spec/acceptance/steps/(.+)_steps\.rb$})   { |m| Dir[File.join("**/#{m[1]}.feature")][0] || 'spec/acceptance' }
-#end
```
sauloperez: It was me who removed all those commented out lines. Git keeps them for us.

lib/open_food_network/packing_report.rb
```diff
@@ -84,9 +81,6 @@ def rules
             sort_by: proc { |full_name| full_name } }
         ]
       else
-#        supplier_rows orders
-#        table_items = supplier_rows orders
-#
```
sauloperez: here as well

PR #2248, opened 2018-04-24
Update rubyzip to fix security issue

sauloperez:
#### What? Why?

Github reported us about CVE-2017-5946 which is a high severity issue.

dep

This gem is used by Roo which already supports the Rubyzip version that contains the fix (version 1.2.1). Check https://github.com/roo-rb/roo/commit/872bb3a0b67fbecf7dd4bc23ff03b7c2764462b0
for further details.

Rubyzip's changelog for the version 1.2.1 can be found in https://github.com/rubyzip/rubyzip/blob/master/Changelog.md#121. You can also check the [introduced changes](https://github.com/rubyzip/rubyzip/compare/v1.2.0...v1.2.1).

#### What should we test?

Not worth of any testing, given the nature of the changes and the reach of this dependency.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -628,7 +628,7 @@ GEM
       ruby-progressbar (~> 1.7)
       unicode-display_width (~> 1.0, >= 1.0.1)
     ruby-progressbar (1.8.1)
-    rubyzip (1.2.0)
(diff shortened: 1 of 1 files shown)


sauloperez:
I'll give it a try :+1:

PR #2249, opened 2018-04-26
Upgrade Rubocop config to be compatible with 0.55

mkllnk:
#### What? Why?

We use Code Climate which runs rubocop to watch our coding standards. I just installed the latest version of rubocop and it failed to run with our current config. I updated the config to make rubocop run again.

#### What should we test?

That code climate passes.


sauloperez:
The main problem I see not using Gemfile's Rubocop is, which version are you running? Do I have to update mine?

sauloperez:
what should we do with this? both @enricostano and @oeoeaio said in Slack that they prefer running `bundle exec`.

daniellemoorhead:
Will you get this resolved and merged @mkllnk?

Gemfile.lock
```diff
@@ -565,10 +565,9 @@ GEM
       rake (>= 0.8.7)
       rdoc (~> 3.4)
       thor (>= 0.14.6, < 2.0)
-    rainbow (2.2.2)
-      rake
+    rainbow (3.0.0)
     raindrops (0.13.0)
-    rake (10.5.0)
+    rake (12.3.1)
```
sauloperez: watch out. AFAIK this version of Rake won't work with our version of RSpec Rails. It may cause an error with `#last_comment`.

mkllnk: Ah thanks. I now discovered the `--conservative` flag of `bundle update`. That's much better for our needs. On the other hand, if we know that that version doesn't work, we should blacklist it in the Gemfile, shouldn't we?

See my additional commits. Seems all fine now.

daniellemoorhead:
Back to you @sauloperez 

fat-controller

PR #2258, opened 2018-05-01
Fix initial password setting

### LINKED ISSUE #2254: Error facing new users who've been invited to manager an enterprise - 'password reset token has expired'
## Description
New users who have been invited to help manage an account (via the new Add New User as a Manager button) are redirected to a 'set your password' page when they confirm their email. This is correct functionality (the users have never set a password), but unfortunately when the new user types in a password and saves it they get the 'password reset token has expired' error message. The good thing is that the user has confirmed their account, so they'll be able to ask for a 'reset password' email (PR2238 prevents unconfirmed users from resetting passwords), however it's still confusing for new users- they have to work out where to go to resent a confirmation link.

This bug was found while testing https://github.com/openfoodfoundation/openfoodnetwork/pull/2220

## Expected Behavior
When the new user is take to the page where they create a password they should be able to set a password and then be logged in.

## Actual Behavior
The page won't let them set a password and they can't login. They need to navigate theselves to reset password in order to create a password.

## Steps to Reproduce
1. Login with an enterprise owner. 
2. Go to Enterprise Setting > Users and 'Add and unregistered user'
3. Next go to the email of the invited manager, and click the confirmation link.
4. You'll be redirected to a page to create a pw. Enter a valid pw. click save.- error appears

## Animated Gif/Screenshot
image

## Context
We've just released this new mechanism for inviting new managers - but it has a fairly confusing bug in it, that all users of the feature will run into.

## Severity
s2

## Your Environment
* Chrome and Firefox
* PC, laptop

## Possible Fix
Make the 'create password' form work without errors.

mkllnk:
#### What? Why?

Closes #2254.

Setting an expiry date for the password reset token makes the token valid.

#### What should we test?

Adding a manager who is not a user yet. Nothing else is affected.

#### Release notes

Fixed an issue with new users setting their password as invited manager.

### DIFF
--- a/app/controllers/admin/manager_invitations_controller.rb
+++ b/app/controllers/admin/manager_invitations_controller.rb
@@ -30,6 +30,8 @@ module Admin
       password = Devise.friendly_token
       new_user = Spree::User.create(email: @email, unconfirmed_email: @email, password: password)
       new_user.reset_password_token = Devise.friendly_token
+      # Same time as used in Devise's lib/devise/models/recoverable.rb.
+      new_user.reset_password_sent_at = Time.now.utc
       new_user.save!
 
       @enterprise.users << new_user
--- a/spec/features/consumer/confirm_invitation_spec.rb
+++ b/spec/features/consumer/confirm_invitation_spec.rb
@@ -1,0 +1,35 @@
+require 'spec_helper'
+
+feature "Confirm invitation" do
+  include UIComponentHelper # for be_logged_in_as
+
+  describe "confirm email" do
+    let(:email) { "test@example.org" }
+    let(:user) { Spree::User.create(email: email, unconfirmed_email: email, password: "secret") }
+
+    before do
+      user.reset_password_token = Devise.friendly_token
+      user.reset_password_sent_at = Time.now.utc
+      user.save!
+    end
+
+    it "confirms the email address" do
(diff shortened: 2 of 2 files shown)


mkllnk:
Thanks @oeoeaio. I was wondering about that. I never quite know which principal is more important:

- Test only one thing at a time.
- Speed up tests.

Since we have a really big, slow test suit, I think we should try to optimise feature specs while all others can be clean and separate.

oeoeaio:
Yep, totally agree. I have a preference for the "smoke test" pattern for feature specs, where you try and test as much basic functionality as possible in one or two tests, without trying to cover all possibilities. Really important features can be tested in their own dedicated feature specs, but I think we should try and minimise these as much as possible.

Obviously unit tests are different and the more coverage we can provide of edge cases and specific features the better.

myriamboure:
@mkllnk does it need to be tested? Is Sally testing or should I? I can do it today.

mkllnk:
Ah, yes, this needs testing. Sally will be working on Friday.

sstead:
**Testing notes** @mkllnk 

- The new manager can now set their password after they confirm their email, excellent.
- No issues found

https://docs.google.com/document/d/1siMe-4y-02V0MBy5ecxjvFd45rGXMfkN6Z6W_cGeXtA/edit#

PR #2260, opened 2018-05-02
Improve order_and_distributor report's readability

sauloperez:
#### What? Why?

I had to go through the `order_and_distributor` report's code because a user asked us about a particular column in it. As it was, the report's code wasn't particularly readable and I couldn't tell what was that column at a glance. I believe this is possible now.

#### What should we test?

The `order_and_distributor` report

#### Release notes

Improved reabaility of  `order_and_distributor` report's code

### DIFF
--- a/lib/open_food_network/order_and_distributor_report.rb
+++ b/lib/open_food_network/order_and_distributor_report.rb
@@ -32,15 +32,43 @@ module OpenFoodNetwork
 
       @orders.each do |order|
         order.line_items.each do |line_item|
-          order_and_distributor_details << [order.created_at, order.id,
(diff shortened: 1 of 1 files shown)


kirstenalarsen:
I think this might have been the first custom report we ever did . . perhaps my housemate hacking on a weekend to extract our new concept of ‘distributor’ into a report. Ah, nostalgia ;) Glad it is now readable!

myriamboure:
I checked the report and compared to the same report on French production, I can't see any difference on user side, report is unchanged so good to go for me.
https://docs.google.com/document/d/1SxAMtijUH7TyK7nrqMG6VmhF2ICtMHdcePA5ZkjGrBo/edit#

PR #2263, opened 2018-05-03
2211 do not load unfiltered reports

### LINKED ISSUE #2211: Reports: Don't load any data
## Description

**- On page:** /admin/reports when clicking on any link from that page
screenshot from 2018-04-12 19-35-30

**- I want to be able to do:** access the page with no data loaded (pre-defined existing filters options are still there)

## Acceptance criteria
1- When the user click on any of the links from the /admin/reports, the page opens with the filter and instead of the data a message is shown saying "Please select filters in order to access your data".
untitled presentation

2- Once the user applies filters and click on "search", the page reloads, but does not open a new window (it is already the case).

mkllnk:
#### What? Why?

Closes #2211

Most reports are slow and some even fail to render within the request time. As a first step, we don't load the content of a report until the user has entered some search criteria and submitted the form.

It is still possible to display full reports if the search form is left empty and you click on "Search".

The code also got groomed a little, but I had to stop myself re-factoring to get the job done.

#### What should we test?

All reports. At least, we need to check that every report is not displaying data to start with, but displays data when searching. In addition, the Orders And Distributors report had a few more changes and it would be good to double check that it's still displaying the right thing.

#### Release notes

Reports don't load any data before clicking on "Search" to speed up report access.

We are overriding Spree's reports controller. If there is work on the reports controller within the Spree upgrade branch, there could be merge conflicts. But no change here should make the upgrade more difficult, it should just become easier.


app/controllers/spree/admin/reports_controller_decorator.rb
```diff
@@ -53,17 +53,17 @@ def report_types
...
 
-  # Overide spree reports list.
+  # Override spree reports list.
   def index
     @reports = authorized_reports
     respond_with(@reports)
   end
 
-  # This action is short because we refactored it like bosses
+  # This action is short because we re-factored it like bosses
```
sauloperez: I've always seen `refactor` and never `re-factor` in the computer science literature :thinking:

lib/open_food_network/customers_report.rb
```diff
@@ -1,9 +1,10 @@
 module OpenFoodNetwork
   class CustomersReport
     attr_reader :params
-    def initialize(user, params = {})
+    def initialize(user, params = {}, compile_table = false)
       @params = params
       @user = user
+      @compile_table = compile_table
```
sauloperez: while do you refer to `compile` and not `render`. What's the difference with the BulkCoopReport above?

spec/controllers/spree/admin/reports_controller_spec.rb
```diff
@@ -68,10 +68,13 @@
 
     describe 'Orders & Fulfillment' do
       it "shows all orders in order cycles I coordinate" do
-        spree_get :orders_and_fulfillment
+        # create test objects
+        [orderA1, orderA2, orderB1, orderB2]
```
sauloperez: why not using `let!` instead of this expression? seems a bit odd to me. The fact that you need a comment feels like a symptom to me.

spec/controllers/spree/admin/reports_controller_spec.rb
```diff
@@ -199,74 +211,104 @@
     before { login_as_admin }
 
     it "should build distributors for the current user" do
+      [c1, c2, s1, d1, d2, d3]
```
sauloperez: I would rather create these c1, c2, etc within this "Products & Inventory" context. And if we could provide more meaningful names... it's really hard to understand what this vars are about :pray:

spec/features/admin/reports_spec.rb
```diff
@@ -204,28 +204,28 @@
...
 
       # And the totals and sales tax should be correct
-      page.should     have_content "1512.99" # items total
-      page.should     have_content "1500.45" # taxable items total
-      page.should     have_content "250.08" # sales tax
-      page.should     have_content "20.0" # enterprise fee tax
+      expect(page).to     have_content "1512.99" # items total
+      expect(page).to     have_content "1500.45" # taxable items total
+      expect(page).to     have_content "250.08" # sales tax
+      expect(page).to     have_content "20.0" # enterprise fee tax
```
sauloperez: Now that the tabulation is not aligned with the paragraphs above and below I'd simply replace it with a single space.

app/controllers/spree/admin/reports_controller_decorator.rb
mkllnk: I've seen both and thought that "re-factor" would be grammatically correct. It was probably the influence of my outdated spell checker that made me do this change. I'm happy to remove it.

lib/open_food_network/customers_report.rb
mkllnk: "render" is used to draw something on the screen. If you you have two `render` statements in a controller, it will raise an error. The intent here is not just to avoid the rendering of data, it's to avoid all the data loading and sorting in the first place.

spec/controllers/spree/admin/reports_controller_spec.rb
mkllnk: The thing about the comment is true. I don't know of any pattern for this. If I used `let!`, would need to add a whole block around every single example so that the `let!` happens only for that one example. Or can you use it within `it`?

I would like to create these objects on a case by case basis, individually for each `it`. I could assign them to a variable just for naming purposes, but then we have an unused variable: `test_context = [orderA1, orderA2, orderB1, orderB2]`

How would you do it?

app/controllers/spree/admin/reports_controller_decorator.rb
mkllnk: Looking at this again, now that we have made more of these methods short, we can probably delete the whole comment.

spec/controllers/spree/admin/reports_controller_spec.rb
mkllnk: > I would rather create these c1, c2, etc within this "Products & Inventory" context.

At the moment, there are several tests in this context that don't need these objects at all. I can create more contexts. Do you think that would improve readability?

> if we could provide more meaningful names...

Oh yes, totally agree. That would be great. I put that on my list.

app/controllers/spree/admin/reports_controller_decorator.rb
```diff
@@ -250,23 +215,29 @@ def xero_invoices
...
+    @report = OpenFoodNetwork::XeroInvoicesReport.new spree_current_user, params, render_content?
     render_report(@report.header, @report.table, params[:csv], "xero_invoices_#{timestamp}.csv")
   end
 
+  private
+
+  # We don't want to render data unless search params are supplied.
+  # Compiling data can take a long time.
+  def render_content?
+    request.post?
```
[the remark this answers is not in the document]
mkllnk: Yeah, I was wondering about that. My first draft looked for common search parameters. The question is: how difficult is it to maintain this list? Do we look for the presence or absence of search parameters?

One difficulty is that some actions change the `params` object. They set default values. So I need to determine the search state before that. Or I change the code to not change the `params` object, because I think it's bad practice. But Spree's own implementation does that:

https://github.com/spree/spree/blob/baadc8f95cbd25ff8b1741742dbaed75045cbc25/backend/app/controllers/spree/admin/reports_controller.rb#L29

mkllnk:
Thank you @sauloperez and @enricostano for your review. Really good feedback. I think I addressed all your concerns in new commits. Let me know what you think.

lib/open_food_network/bulk_coop_report.rb
```diff
@@ -44,25 +46,12 @@ def header
     end
 
     def search
-      permissions.visible_orders.complete.not_state(:canceled).search(params[:q])
+      OpenFoodNetwork::Reports::LineItems.search_orders(permissions, params)
```
sauloperez: No need to specify `OpenFoodNetwork` here. This code is already within that namespace.

lib/open_food_network/order_and_distributor_report.rb
```diff
@@ -1,8 +1,12 @@
...
   class OrderAndDistributorReport
 
-    def initialize orders
-      @orders = orders
+    def initialize(user, params = {}, render_table = false)
+      @params = params
+      @user = user
+      @render_table = render_table
+
+      @permissions = OpenFoodNetwork::Permissions.new(user)
```
sauloperez: here as well

sauloperez:
looks pretty good! Thanks for the effort 👏. This leaves us in a much better position for upcoming refactors/performance improvements.

lib/open_food_network/bulk_coop_report.rb
mkllnk: You are right, thanks. I changed it.

spec/lib/open_food_network/order_and_distributor_report_spec.rb
```diff
@@ -34,9 +34,9 @@ module OpenFoodNetwork
       end
 
       it "should denormalise order and distributor details for display as csv" do
-        subject = OrderAndDistributorReport.new [@order]
+	subject = OrderAndDistributorReport.new create(:admin_user), {}, true
 
-        table = subject.table
+        table = subject.send(:line_item_details, [@order])
```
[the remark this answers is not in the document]
mkllnk: This was just the result of refactoring and I'm not sure how to resolve that. I don't think we should change the method to public just so that we can test it. Strictly speaking, we just want to test the public methods, but I would have needed to rewrite the test for that (more effort). What would you do?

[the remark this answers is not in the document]
sauloperez: I don't want to make it public either. I wonder if we can still rely on the `#table` method to compare the expected output against.

mkllnk:
Okay @sauloperez, I refactored that spec to use the public `table` method. I had to change the created order for that and took the opportunity to clean it up. What do you think now?

I never thought this pull request would grow that big.

sauloperez:
Way better now!

daniellemoorhead:
@enricostano can you review the changes made by @mkllnk and approve if you're happy please 😄

app/controllers/spree/admin/reports_controller_decorator.rb
```diff
@@ -250,23 +215,53 @@ def xero_invoices
...
+  def cache_search_state
+    search_keys = [
+      # search parameter for ransack
+      :q,
+      # common in all reports, only set for CSV rendering
+      :csv,
+      # `button` is included in all forms. It's not important for searching,
+      # but the Users & Enterprises report doesn't have any other parameter
+      # for an empty search. So we use this one to display data.
+      :button,
```
oeoeaio: Could we just use the presence of a `:button` param as a proxy for a search request for all reports? That would seem to negate the need to check all of these other params as well...?

app/controllers/spree/admin/reports_controller_decorator.rb
```diff
@@ -289,42 +284,42 @@ def load_data
...
+      I18n.t!(:description, scope: [:admin, :reports, report])
+    rescue I18n::MissingTranslationData
+      render_to_string(
+        partial: "#{report}_description",
+        layout: false,
+        locals: { report_types: report_types[report] }
+      ).html_safe
+    end
+    { name: name, description: description }
   end
```
oeoeaio: Nice. 👌

sstead:
**Testing note:**
- yes reports initially show no data, then the data loads when you hit Search.
- The same is true in superadmin- except for one report (sales totals) but I don't think this is a problem.

- Regarding the 'orders and distributors' report - it's behaving the same as on production. But there seems to be an exisitng issue with one column (shipping cost) not displaying anything.

https://docs.google.com/document/d/1lwevBhGMqh0WC9waDysGG4ZuaVd4efCzt_ItL4JWSSI/edit#

PR #2268, opened 2018-05-07
Set response headers to disable cache - to avoid back button emptying…

### LINKED ISSUE #1213: Back from view cart empties cart
## Description
When shopping **as a guest**, my cart get emptied if I hit the "backward" browser button.

## Expected Behavior
If I am logged in, or if I am shopping as a guest, if I put products in my cart and go to the basket page, when hitting the browser "backward" button, my cart should remain with the products I put in it, and I should just be sent back to the shopfront page.

## Actual Behavior
When I shop as a guest, I add products and then go to cart page, if I hit the browser "backward" button, I'm sent back to shopfront page BUT my cart is emptied.

## Steps to Reproduce
1. Do not log in, and open any shopfront. Add products to your cart and click on "view cart" button (either from bottom of product lists, or small right corner "cart" button)
2. When on cart page, hit the browser "backward" button and see what happens.

## Animated Gif/Screenshot
https://www.useloom.com/share/1e4de54bbbf04944ac8147e2d15232ce

## Context
Reported originally from Rohan when testing previous spree upgrade, re-tested and updated by Myriam.

## Severity
S2, all guest users are affected, and placing order is a big feature, and pretty big UX impact that has no really workaround?

## Your Environment
- Version used: 1.10.1

## Possible Fix
?

stveep:
… cart (#1213)

#### What? Why?

Closes #1213 

Prevents use of cache which was resulting in loss of items in cart when pressing back button

#### What should we test?

- Add items to cart
- Click Edit your cart
- Press browser back button
- Items should still be there
- Check general performance when navigating around site

#### Release notes

Pressing the browser back button from the cart page no longer results in newly-added items being removed.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery
 
   prepend_before_filter :restrict_iframes
+  before_filter :set_cache_headers # Issue #1213, prevent cart emptying via cache when using back button
 
   include EnterprisesHelper
   helper CssSplitter::ApplicationHelper
@@ -150,6 +151,12 @@ class ApplicationController < ActionController::Base
     block.call
     self.formats = old_formats
     nil
+  end
+
+  def set_cache_headers # https://jacopretorius.net/2014/01/force-page-to-reload-on-browser-back-in-rails.html
+    response.headers["Cache-Control"] = "no-cache, no-store, max-age=0, must-revalidate"
+    response.headers["Pragma"] = "no-cache"
+    response.headers["Expires"] = "Fri, 01 Jan 1990 00:00:00 GMT"
   end
 
 end


app/controllers/application_controller.rb
```diff
@@ -4,6 +4,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery
 
   prepend_before_filter :restrict_iframes
+  before_filter :set_cache_headers # Issue #1213, prevent cart emptying via cache when using back button
```
Matt-Yorkley: So this will remove all browser caching on all pages?

app/controllers/application_controller.rb
```diff
@@ -152,4 +153,10 @@ def with_format(format, &block)
     nil
   end
 
+  def set_cache_headers # https://jacopretorius.net/2014/01/force-page-to-reload-on-browser-back-in-rails.html
```
Matt-Yorkley: It feels like this method could have a more descriptive name, like `disable_browser_caching`. `set_cache_headers` doesn't really convey what it does.

app/controllers/application_controller.rb
Matt-Yorkley: What sort of impact will that have?

Matt-Yorkley: I'm thinking if this is an issue with Angular not picking up the changes after the back button is used, maybe there's a way to make Angular trigger a cart refresh, and leave the browser caching alone?

myriamboure:
In term of performance, maybe that worth testing instead of assuming that the user experience is affected @mkllnk ? @stveep warned about that risk in the issue, but I would be in favor to give it a try if we don't find any other better way to solve the problem?

mkllnk:
I think the performance issue is mainly when people hit the back button. At the moment, the page is not reloaded and displays immediately. That also means that you don't lose your scroll position. For example, the list of producers takes 4 seconds to load on our Australian instance. This change would mean at least 4 additional seconds when you hit back to go to the list of producers.

Testing is always good. We can do that. One thing we can't test on a staging server is the impact on the performance of a live site. Page reloads don't just mean that it takes longer for you to see the page, it also means that our server is a lot more busy and that slows everything down.

In the big picture, I don't know how often people use the back button, but it seems to be often enough to report this issue.

There are definitely better solutions, but they also take longer to implement. :-(

stveep:
OK thanks for looking at this. Maybe I will look into catching back button press somehow and forcing a cart refresh in the Angular service instead. Any other ideas? Please post on the issue #1213 

@myriamboure it would be useful if you could quickly test that this does fix the issue just so we know that this is the cause - even if it's not the optimal solution.

myriamboure:
Ok I moved that to test ready for a first round of tests on this.

mkllnk:
Staged on https://staging1.openfood.com.au/ for your convenience, @myriamboure. ;-)

mkllnk:
@sauloperez 

> can't we do that just in the affected controller?

The cart is displayed on every page. So to update the cart properly we can't cache any page or we need to have a script that updates the cart information separately on any navigation event. From your comment https://github.com/openfoodfoundation/openfoodnetwork/issues/1213#issuecomment-389053603 I take that I can go ahead and merge this even though your review requested changes.

I would be happy about pull requests with a better solution, but for now it seems the best we have.

sstead:
**Testing notes**
Good to go.

- Cart doesn't empty for guests or users in Firefox or Chrome. 

https://docs.google.com/document/d/1IYtQxTBj3ewphiD4oirqniRUfruRhaBrKPYB5b2lU4k/edit#

PR #2270, opened 2018-05-08
Update activemerchant to v1.78 with new root cert

### LINKED ISSUE #2265: ActiveMerchant fails to connect to servers with new TLS certificates
The Pin Payments gateway just stopped working, because the server got new TLS/SSL certificates. The new certificates are issued with a root cert that is not in the ActiveMerchant bundle.

## Expected Behavior
All payment gateways are working.

## Actual Behavior
Pin Payments stopped working. If any other services renew their certificates that are not supported, the same happens to other gateways as well.

## Steps to Reproduce

1. `bundle exec rails console`
2. `ActiveMerchant::Connection.new("https://test-api.pin.net.au").request(:get, nil)`

Bad response:
```
ActiveMerchant::ConnectionError: The SSL connection to the remote server could not be established
```
Good response:
```
=> #<Net::HTTPNotFound 404 Not Found readbody=true>
```

## Context

We got notified by Pin Payments about their server updates and tested if payments still go through.

## Severity

I'll give it an s3, because it only affects Australia at the moment and I have a manual patch. But it's not a lasting solution.

## Possible Fix

Hot fix on the server:
```sh
curl -fsS https://raw.githubusercontent.com/activemerchant/active_merchant/7ebdc250a7f802d64731f826135b5b8ff86c663d/lib/certs/cacert.pem > "$(bundle show activemerchant | head -n 1)/lib/certs/cacert.pem"
```

Long term fixes:

- Create own branch of ActiveMerchant and patch it (sorry I ran out of time today).
- Get the Spree upgrade done and upgrade ActiveMerchant to the latest version.

mkllnk:
#### What? Why?

Closes #2265

Update ActiveMerchant including a new root certificate needed for at least Pin Payments.

#### What should we test?

Most changes are in gateways we don't use, I believe. There has been a
change in Stripe, but we use another implementation, I guess.

To verify my guesses and beliefs, we need to check the used payment gateways.

- Pin Payments
- Stripe
- What else do we use?

#### Release notes

Fixed the Pin Payments payment gateway which broke 4 May 2018 due to certificate updates.

Newer versions of Spree may declare different dependencies on ActiveMerchant, but there should be no conflicts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,9 @@ gem 'spree_auth_devise', github: 'openfoodfoundation/spree_auth_devise', branch:
 gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "spree-upgrade-intermediate"
 #gem 'spree_paypal_express', github: "spree-contrib/better_spree_paypal_express", branch: "1-3-stable"
 gem 'stripe', '~> 3.3.1'
-gem 'activemerchant', '~> 1.71.0'
+# We need at least this version to have Digicert's root certificate
+# which is needed for Pin Payments (and possibly others).
+gem 'activemerchant', '~> 1.78'
 
 gem 'oauth2', '~> 1.2.0' # Used for Stripe Connect
 gem 'jwt', '~> 1.5'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -152,7 +152,7 @@ GEM
       sprockets (~> 2.2.1)
     active_model_serializers (0.8.3)
       activemodel (>= 3.0)
-    activemerchant (1.71.0)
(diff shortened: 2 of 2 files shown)


sauloperez:
Spree 2.0 declares `s.add_dependency 'activemerchant', '~> 1.34'` so we're good.

sstead:
**Testing notes**
 These payment gateways were tested, and are functioning:
- Stripe
- Pin
- Paypal
- Cash/eft etc

Not tested (we don't have a test method)
- MIGS

https://docs.google.com/document/d/16fYxSixnomWfkk7CSs-UneM9Q_fhqQg3WK6PDoztNCo/edit#

PR #2272, opened 2018-05-09
changed error message in en.yml line 2207

### LINKED ISSUE #620: Duplicate enterprise name error message tweek.
If it's not too hard can we please change the error message from...

"Name has already been taken. If this is your enterprise and you would like to claim ownership, please contact the current manager of this profile at sstead@student.unimelb.edu.au."

to 

"Name has already been taken. If this is your enterprise and you would like to claim ownership, or if you would like to trade with this enterprise please contact the current manager of this profile at sstead@student.unimelb.edu.au"

This is just to capture the user case of....'I'm a hub, I'm adding profiles for my supplying producers, I already checked to see if they had live profiles at .org.au, but maybe I couldn't see one because it is 'invisible'... so I try to make a profile for it... but can't.... WHY (they might get confused).

rtrunkhill:
Closes #620

modified error message in en.yml line 2207

### DIFF
--- a/spec/models/enterprise_spec.rb
+++ b/spec/models/enterprise_spec.rb
@@ -129,17 +129,15 @@ describe Enterprise do
 
       it "prevents duplicate names for new records" do
         e = Enterprise.new name: enterprise.name
-        e.should_not be_valid
-        e.errors[:name].first.should ==
-          "has already been taken. If this is your enterprise and you would like to claim ownership, please contact the current manager of this profile at owner@example.com."
+        expect(e).to_not be_valid
+        expect(e.errors[:name].first).to include I18n.t('enterprise_name_error', email: owner.email)
       end
 
       it "prevents duplicate names for existing records" do
         e = create(:enterprise, name: 'foo')
         e.name = enterprise.name
-        e.should_not be_valid
-        e.errors[:name].first.should ==
-          "has already been taken. If this is your enterprise and you would like to claim ownership, please contact the current manager of this profile at owner@example.com."
+        expect(e).to_not be_valid
+        expect(e.errors[:name].first).to include I18n.t('enterprise_name_error', email: owner.email)
       end
 
       it "does not prohibit the saving of an enterprise with no name clash" do
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2208,7 +2208,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   content_configuration_pricing_table: "(TODO: Pricing table)"
   content_configuration_case_studies: "(TODO: Case studies)"
   content_configuration_detail: "(TODO: Detail)"
-  enterprise_name_error: "has already been taken. If this is your enterprise and you would like to claim ownership, please contact the current manager of this profile at %{email}."
+  enterprise_name_error: "has already been taken. If this is your enterprise and you would like to claim ownership, or if you would like to trade with this enterprise please contact the current manager of this profile at %{email}."
   enterprise_owner_error: "^%{email} is not permitted to own any more enterprises (limit is %{enterprise_limit})."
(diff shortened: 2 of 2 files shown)


rtrunkhill:
okay, made the rspec changes and sent through the pr

spec/models/enterprise_spec.rb
```diff
@@ -131,15 +131,15 @@
         e = Enterprise.new name: enterprise.name
         e.should_not be_valid
         e.errors[:name].first.should ==
-          "has already been taken. If this is your enterprise and you would like to claim ownership, please contact the current manager of this profile at owner@example.com."
+          "has already been taken. If this is your enterprise and you would like to claim ownership, or if you would like to trade with this enterprise please contact the current manager of this profile at %{email}."
```
sauloperez: It's a bit more future proof if you use i18n here so that we use the translation key rather than the actual copy.

spec/models/enterprise_spec.rb
```diff
@@ -131,15 +131,15 @@
...
       end
 
       it "prevents duplicate names for existing records" do
         e = create(:enterprise, name: 'foo')
         e.name = enterprise.name
         e.should_not be_valid
         e.errors[:name].first.should ==
-          "has already been taken. If this is your enterprise and you would like to claim ownership, please contact the current manager of this profile at owner@example.com."
+          "has already been taken. If this is your enterprise and you would like to claim ownership, or if you would like to trade with this enterprise please contact the current manager of this profile at %{email}."
```
sauloperez: Here too.

rtrunkhill:
im not familiar with i18n.  on my branch those were both changed.   Could
you maybe explain more what i need to do here?

sauloperez:
Sorry, I mean something like it was done in https://github.com/openfoodfoundation/openfoodnetwork/blob/f2e07419a812499f2de568a05872bde46edaf4f8/spec/controllers/admin/schedules_controller_spec.rb.

As you see below, we refer to the translation key instead of the actual text. This way, when the translation changes we don't need to update the test, which is tiresome and pointless.

https://github.com/openfoodfoundation/openfoodnetwork/blob/f2e07419a812499f2de568a05872bde46edaf4f8/spec/controllers/admin/schedules_controller_spec.rb#L215.

Also, please make sure you use [RSpec's expect syntax](http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/) instead of `should`.

rtrunkhill:
how does that look?

frank-west-iii:
Hey there @rtrunkhill, there are a few minor adjustments that might help get the tests passing here. I know that the test run can take a bit, however running this command will scope the test run down to just your changed specs.

`bundle exec rspec spec/models/enterprise_spec.rb:130:136`

There are some minor differences between the two `expect` syntaxes in `rspec`.

One can be stated as you have it: `expect { foo }.to raise_error`
The other one is: `expect(foo).to eq('bar')`

There is some discussion about the differences here: https://stackoverflow.com/questions/19960831/rspec-expect-vs-expect-with-block-whats-the-difference

There are also some lint failures from `rubocop`. Running `bundle exec rubocop spec/models/enterprise_spec.rb` will help guide you through those fixes as well.

Once `rubocop` and `rspec` go green you should be in a good state 👍. Feel free to ping me if you have any questions.

rtrunkhill:
thank you so much for all the help, I've never run into issues like that
for spec before.

sent the PR through with passing tests and 0 issues w/ rubocop

sstead:
**Testing notes**

Hooray, looks good! The message is updated in all the right places.

https://docs.google.com/document/d/1ukEaOk8JaQrc42KKAEI87zu5sQBeNrXjzONKK6odiac/edit#heading=h.gyoezt55hcmx

PR #2273, opened 2018-05-10
Allow credit owed on payments made via stripe to be refunded

oeoeaio:
#### What? Why?

I forgot to implement the ability to partially refund payments made via Stripe, so I've done it here.

This is required in the case that the order total changes after the payment has been made (eg. an item is not available), and the customer needs to be refunded for the difference between what they paid and the new order total.

#### What should we test?

That orders with credit owed which have already been paid for via Stripe can have the credit owed amount refunded to the customers credit card. To do this you need to head to the Payments section of an order in the admin interface and you should see a button with a tick as the icon. Clicking this button submit a request for a refund to Stripe, equal to the amount of credit on the order.

screen shot 2018-05-10 at 7 26 32 pm

This should be fine to test using test cards - Stripe is very well designed in that if something works in test mode it will almost certainly work in production mode. Would be good to do at least one test in a production environment once deployed somewhere though.

#### Release notes

Order which have credit owed can now have the credit owed amount refunded to the customer's card.

### DIFF
--- a/app/assets/stylesheets/admin/icons.css.scss
+++ b/app/assets/stylesheets/admin/icons.css.scss
@@ -1,3 +1,4 @@
 @import 'plugins/font-awesome';
 
 .icon-refund:before { @extend .icon-ok:before }
+.icon-credit:before { @extend .icon-ok:before }
--- a/app/models/spree/gateway/stripe_connect.rb
+++ b/app/models/spree/gateway/stripe_connect.rb
@@ -31,6 +31,7 @@ module Spree
         StripeAccount.find_by_enterprise_id(preferred_enterprise_id).andand.stripe_user_id
       end
 
+      # NOTE: the name of this method is determined by Spree::Payment::Processing
       def purchase(money, creditcard, gateway_options)
         provider.purchase(*options_for_purchase_or_auth(money, creditcard, gateway_options))
       rescue Stripe::StripeError => e
@@ -38,11 +39,13 @@ module Spree
         failed_activemerchant_billing_response(e.message)
       end
 
+      # NOTE: the name of this method is determined by Spree::Payment::Processing
       def void(response_code, _creditcard, gateway_options)
         gateway_options[:stripe_account] = stripe_account_id
         provider.void(response_code, gateway_options)
       end
 
+      # NOTE: the name of this method is determined by Spree::Payment::Processing
       def credit(money, _creditcard, response_code, gateway_options)
         gateway_options[:stripe_account] = stripe_account_id
         provider.refund(money, response_code, gateway_options)
--- a/app/models/spree/gateway/stripe_connect.rb
+++ b/app/models/spree/gateway/stripe_connect.rb
@@ -43,6 +43,11 @@ module Spree
         provider.void(response_code, gateway_options)
       end
 
+      def credit(money, _creditcard, response_code, gateway_options)
+        gateway_options[:stripe_account] = stripe_account_id
+        provider.refund(money, response_code, gateway_options)
+      end
+
       def create_profile(payment)
         return unless payment.source.gateway_customer_profile_id.nil?
 
--- a/spec/controllers/spree/admin/payments_controller_spec.rb
+++ b/spec/controllers/spree/admin/payments_controller_spec.rb
@@ -66,5 +66,59 @@ describe Spree::Admin::PaymentsController, type: :controller do
         end
       end
     end
+
+    context "requesting a partial credit on a payment" do
+      let(:params) { { id: payment.id, order_id: order.number, e: :credit } }
+
+      # Required for the respond override in the controller decorator to work
+      before { @request.env['HTTP_REFERER'] = spree.admin_order_payments_url(payment) }
+
+      context "that was processed by stripe" do
+        let!(:payment_method) { create(:stripe_payment_method, distributors: [shop], preferred_enterprise_id: shop.id) }
+        let!(:payment) { create(:payment, order: order, state: 'completed', payment_method: payment_method, response_code: 'ch_1a2b3c', amount: order.total + 5) }
+
+
+        before do
+          allow(Stripe).to receive(:api_key) { "sk_test_12345" }
+        end
+
+        context "where the request succeeds" do
+          before do
+            stub_request(:post, "https://sk_test_12345:@api.stripe.com/v1/charges/ch_1a2b3c/refunds").
+              to_return(:status => 200, :body => JSON.generate(id: 're_123', object: 'refund', status: 'succeeded') )
+          end
+
+          it "partially refunds the payment" do
+            order.reload
+            expect(order.payment_total).to eq order.total + 5
+            expect(order.outstanding_balance).to eq(-5)
+            spree_put :fire, params
+            expect(payment.reload.state).to eq 'completed'
+            order.reload
+            expect(order.payment_total).to eq order.total
+            expect(order.outstanding_balance).to eq 0
+          end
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/admin/icons.css.scss
```diff
@@ -1,3 +1,4 @@
 @import 'plugins/font-awesome';
 
 .icon-refund:before { @extend .icon-ok:before }
```
sauloperez: what about `.icon-refund:before, .icon-credit:before { ... }` ?

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -43,6 +43,11 @@ def void(response_code, _creditcard, gateway_options)
         provider.void(response_code, gateway_options)
       end
 
+      def credit(money, _creditcard, response_code, gateway_options)
+        gateway_options[:stripe_account] = stripe_account_id
+        provider.refund(money, response_code, gateway_options)
+      end
+
```
sauloperez: I'm missing unit tests for this method.

spec/controllers/spree/admin/payments_controller_spec.rb
```diff
@@ -66,5 +66,59 @@
...
+          before do
+            stub_request(:post, "https://sk_test_12345:@api.stripe.com/v1/charges/ch_1a2b3c/refunds").
+              to_return(:status => 200, :body => JSON.generate(id: 're_123', object: 'refund', status: 'succeeded') )
+          end
+
+          it "partially refunds the payment" do
+            order.reload
+            expect(order.payment_total).to eq order.total + 5
+            expect(order.outstanding_balance).to eq(-5)
+            spree_put :fire, params
```
sauloperez: is `fire` the name of an action? 😲

spec/controllers/spree/admin/payments_controller_spec.rb
```diff
@@ -66,5 +66,59 @@
...
+            expect(payment.reload.state).to eq 'completed'
+            order.reload
+            expect(order.payment_total).to eq order.total
+            expect(order.outstanding_balance).to eq 0
+          end
```
sauloperez: I find confusing the fact that there are two `expect` blocks (verification phase) in this test. I don't know whether the subject of the test is `spree_put :fire, params` or something else I don't see. IMO tests are easier to read when using the [four-phase approach](https://robots.thoughtbot.com/four-phase-test): setup, exercise, verify and teardown.

spec/controllers/spree/admin/payments_controller_spec.rb
```diff
@@ -66,5 +66,59 @@
...
+            order.reload
+            expect(order.payment_total).to eq order.total + 5
+            expect(order.outstanding_balance).to eq(-5)
+            spree_put :fire, params
+            expect(payment.reload.state).to eq 'completed'
+            order.reload
+            expect(order.payment_total).to eq order.total + 5
+            expect(order.outstanding_balance).to eq -5
+            expect(flash[:error]).to eq "Bup-bow!"
+          end
```
sauloperez: Here as well. It feels more like a procedure than an actual unit test to me.

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -43,6 +43,11 @@ def void(response_code, _creditcard, gateway_options)
         provider.void(response_code, gateway_options)
       end
 
+      def credit(money, _creditcard, response_code, gateway_options)
```
sauloperez: I would have expected to name it `refund`. Is this part of some existing interface?

spec/controllers/spree/admin/payments_controller_spec.rb
oeoeaio: Yep, comes from spree

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Yep, I have no control over the method naming, this whole gateway class is basically just a thin wrapper of `ActiveMerchant::Billing::StripeGateway` used by `Spree::Payment::Processing`. There is a `refund` method in `StripeGateway`, but I think it does something different (maybe refunds the entire order?)

app/assets/stylesheets/admin/icons.css.scss
oeoeaio: Higher maintenance cost? If I want to change credit to a different icon I have to split them out again. :trollface:

spec/controllers/spree/admin/payments_controller_spec.rb
oeoeaio: Are you suggesting that we shouldn't make more than one assertion per spec? Would you be happier if I made this a request spec?

app/models/spree/gateway/stripe_connect.rb
sauloperez: It would be good to explain this as method documentation so the next dev doesn't wonder the same.

daniellemoorhead:
@oeoeaio this has been sitting here for a couple of weeks now with no change. Should we put it back into dev ready till you've got time to finish it?

time

oeoeaio:
This shouldn't take me long to tweak. Let's keep it here and I'll get around to it as soon as I can.

spec/controllers/spree/admin/payments_controller_spec.rb
oeoeaio: @sauloperez I want to improve this, but I'm not really sure how. I want to test that all of these things happen, as they are all affects that I am expecting to see. This is very much an integration test, so it is checking that a bunch of different things are working properly together. Would you prefer if I had each of these expectations in their own example?

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Done

app/models/spree/gateway/stripe_connect.rb
oeoeaio: Done.

app/assets/stylesheets/admin/icons.css.scss
oeoeaio: Happy to keep as is? I will change it if you insist.

spec/controllers/spree/admin/payments_controller_spec.rb
sauloperez: > Would you prefer if I had each of these expectations in their own example?

Yes. You're actually asserting different things such as the fact that partially refunding the payment resets the outstanding balance, for instance.

app/assets/stylesheets/admin/icons.css.scss
sauloperez: no need to worry.

mkllnk:

sstead:
Testing notes
Things that work:
- The shop manager can refund credit on a stripe payment (only tested in Test mode).
- Worked in superadmin
- The credit refunded line is recorded and shown at the customers's /account page, transaction history tab.

Things to be improved 
- The layout of the 'credit' button doesn't make it immediately apparent that I'm refunding credit to the customer. Nor is the amount I'm refunding clear.
- There's a void button next to refund payment lines - if I click the button I get a nonsense error message

Wider 'in credit' order related issues
- The way a customers sees their credit owed orders isn't clear or consistent. new issue here https://github.com/openfoodfoundation/openfoodnetwork/issues/2423
- From brief testing of 2 reports, they aren't handling 'in credit' orders or refund transactions very clearly.

PR #2274, opened 2018-05-10
Subs: Default Card

### LINKED ISSUE #2085: Customers authorise the use of credit cards by shops
## Overview

## Expected outcome 

oeoeaio:
#### What? Why?

Part 1 of 3 PRs to close #2085 

To be able to process payments for subscriptions, shops need to be able to to charge a credit card saved against customer's account. The approach I decided to take is to allow the customer to set one of their cards as the default and to allow them to authorise shops to charge their default card for services. We COULD do something more sophisticated in the future where specific cards are nominated for specific services, but I think we should start with this and we can work on that later if required.

#### What should we test?

This part is just the part that allows customers to specify a default card. The other stuff comes later.

Basically:
- [ ] as a regular user, I should be able to easily understand which of my saved credit card is the 'default' from the Account page
- [ ] when a new card is added to an account it should become the default if none is already set
- [ ] adding subsequent cards should not change the default card
- [ ] I should be able to update the default card using the interface on the account page

#### Release notes

Customers can now set a default credit card, which can be used to pay OFN shops for services (such as automatic subscription payments)

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/checkout/payment_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/payment_controller.js.coffee
@@ -9,6 +9,10 @@ Darkswarm.controller "PaymentCtrl", ($scope, $timeout, savedCreditCards, Dates) 
   $scope.secrets.card_month = "1"
   $scope.secrets.card_year = moment().year()
 
+  for card in (savedCreditCards || []) when card.is_default
+    $scope.secrets.selected_card = card.id
+    break
+
   $scope.summary = ->
     [$scope.Checkout.paymentMethod()?.name]
 
--- a/app/assets/javascripts/darkswarm/controllers/checkout/payment_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/payment_controller.js.coffee
@@ -9,7 +9,7 @@ Darkswarm.controller "PaymentCtrl", ($scope, $timeout, savedCreditCards, Dates) 
   $scope.secrets.card_month = "1"
   $scope.secrets.card_year = moment().year()
 
-  for card in (savedCreditCards || []) when card.is_default
+  for card in savedCreditCards when card.is_default
     $scope.secrets.selected_card = card.id
     break
 
--- a/app/models/spree/credit_card_decorator.rb
+++ b/app/models/spree/credit_card_decorator.rb
@@ -14,8 +14,8 @@ Spree::CreditCard.class_eval do
 
   belongs_to :user
 
-  after_create :ensure_default
-  after_save :ensure_default, if: :is_default_changed?
+  after_create :ensure_single_default_card
+  after_save :ensure_single_default_card, if: :is_default_changed?
 
   # Allows us to use a gateway_payment_profile_id to store Stripe Tokens
   # Should be able to remove once we reach Spree v2.2.0
@@ -35,7 +35,7 @@ Spree::CreditCard.class_eval do
     !user.credit_cards.exists?(is_default: true)
   end
 
-  def ensure_default
+  def ensure_single_default_card
     return unless user
     return unless is_default? || default_missing?
     user.credit_cards.update_all(['is_default=(id=?)', id])
--- a/app/models/spree/credit_card_decorator.rb
+++ b/app/models/spree/credit_card_decorator.rb
@@ -5,7 +5,7 @@ Spree::CreditCard.class_eval do
   attr_accessible :cc_type, :last_digits
 
   # For holding customer preference in memory
-  attr_accessible :save_requested_by_customer
+  attr_accessible :save_requested_by_customer, :is_default
   attr_writer :save_requested_by_customer
 
   # Should be able to remove once we reach Spree v2.2.0
@@ -13,6 +13,9 @@ Spree::CreditCard.class_eval do
   belongs_to :payment_method
 
   belongs_to :user
+
+  after_create :ensure_default
+  after_save :ensure_default, if: :is_default_changed?
 
   # Allows us to use a gateway_payment_profile_id to store Stripe Tokens
   # Should be able to remove once we reach Spree v2.2.0
@@ -24,5 +27,18 @@ Spree::CreditCard.class_eval do
 
   def save_requested_by_customer?
     !!@save_requested_by_customer
+  end
+
+  private
+
+  def default_missing?
+    user.credit_cards.where(is_default: true).none?
+  end
+
+  def ensure_default
+    return unless user
+    return unless is_default? || default_missing?
+    user.credit_cards.update_all(['is_default=(id=?)', id])
+    self.is_default = true
   end
 end
--- a/app/models/spree/credit_card_decorator.rb
+++ b/app/models/spree/credit_card_decorator.rb
@@ -32,7 +32,7 @@ Spree::CreditCard.class_eval do
   private
 
   def default_missing?
-    user.credit_cards.where(is_default: true).none?
+    !user.credit_cards.exists?(is_default: true)
   end
 
   def ensure_default
--- a/app/views/spree/users/_saved_cards.html.haml
+++ b/app/views/spree/users/_saved_cards.html.haml
@@ -3,11 +3,14 @@
     %th= t(:card_type)
     %th= t(:card_number)
     %th= t(:card_expiry_date)
+    %th= t(:default?)
     %th= t(:delete?)
   %tr.card{ id: "card{{ card.id }}", ng: { repeat: "card in savedCreditCards" } }
     %td.brand{ ng: { bind: '::card.brand' } }
     %td.number{ ng: { bind: '::card.number' } }
     %td.expiry{ ng: { bind: '::card.expiry' } }
+    %td.is-default
+      %input{ type: 'radio', name: 'default_card', ng: { model: 'card.is_default', change: 'setDefault(card)', value: "true"} }
     %td.actions
       %a{"rel" => "nofollow", "data-method" => "delete", "ng-href" => "{{card.delete_link}}" }
         = t(:delete)
--- a/app/views/spree/users/_saved_cards.html.haml
+++ b/app/views/spree/users/_saved_cards.html.haml
@@ -3,8 +3,8 @@
     %th= t(:card_type)
     %th= t(:card_number)
     %th= t(:card_expiry_date)
-    %th= t(:default?)
-    %th= t(:delete?)
+    %th= t('.default?')
+    %th= t('.delete?')
   %tr.card{ id: "card{{ card.id }}", ng: { repeat: "card in savedCreditCards" } }
     %td.brand{ ng: { bind: '::card.brand' } }
     %td.number{ ng: { bind: '::card.number' } }
--- a/spec/controllers/spree/credit_cards_controller_spec.rb
+++ b/spec/controllers/spree/credit_cards_controller_spec.rb
@@ -69,6 +69,53 @@ describe Spree::CreditCardsController, type: :controller do
     end
   end
 
+  describe "#update" do
+    let(:params) { { format: :json, credit_card: { is_default: true } } }
+    context "when the specified credit card is not found" do
+      before { params[:id] = 123 }
+
+      it "renders a flash error" do
+        put :update, params
+        json_response = JSON.parse(response.body)
+        expect(json_response['flash']['error']).to eq I18n.t(:card_could_not_be_updated)
+      end
+    end
+
+    context "when the specified credit card is found" do
+      let!(:card) { create(:credit_card, gateway_customer_profile_id: 'cus_AZNMJ') }
+      before { params[:id] = card.id }
+
+      context "but the card is not owned by the user" do
+        it "redirects to unauthorized" do
+          put :update, params
+          expect(response).to redirect_to spree.unauthorized_path
+        end
+      end
+
+      context "and the card is owned by the user" do
+        before { card.update_attribute(:user_id, user.id) }
+
+        context "when the update completes successfully" do
+          it "renders a serialized copy of the updated card" do
+            expect{ put :update, params }.to change { card.reload.is_default }.to(true)
+            json_response = JSON.parse(response.body)
+            expect(json_response['id']).to eq card.id
+            expect(json_response['is_default']).to eq true
+          end
+        end
+
+        context "when the update fails" do
+          before { params[:credit_card][:month] = 'some illegal month' }
+          it "renders an error" do
+            put :update, params
+            json_response = JSON.parse(response.body)
+            expect(json_response['flash']['error']).to eq I18n.t(:card_could_not_be_updated)
+          end
+        end
+      end
+    end
+  end
+
   describe "#destroy" do
     context "when the specified credit card is not found" do
       let(:params) { { id: 123 } }
--- a/spec/features/consumer/account/cards_spec.rb
+++ b/spec/features/consumer/account/cards_spec.rb
@@ -4,7 +4,8 @@ feature "Credit Cards", js: true do
   include AuthenticationWorkflow
   describe "as a logged in user" do
     let(:user) { create(:user) }
-    let!(:card) { create(:credit_card, user_id: user.id, gateway_customer_profile_id: 'cus_AZNMJ') }
+    let!(:card) { create(:credit_card, user_id: user.id, gateway_customer_profile_id: 'cus_AZNMJ', is_default: true) }
+    let!(:card2) { create(:credit_card, user_id: user.id, gateway_customer_profile_id: 'cus_FDTG') }
 
     before do
       quick_login_as user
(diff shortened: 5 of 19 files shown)


app/models/spree/credit_card_decorator.rb
```diff
@@ -25,4 +28,16 @@ def has_payment_profile? # rubocop:disable Style/PredicateName
...
   end
+
+  private
+
+  def default_missing?
+    user.credit_cards.where(is_default: true).none?
+  end
+
+  def ensure_default
+    return unless user
```
sauloperez: isn't `user` mandatory for a credit card? I feel like this guard close shouldn't be needed.

app/models/spree/credit_card_decorator.rb
```diff
@@ -14,6 +14,9 @@
 
   belongs_to :user
 
+  after_create :ensure_default
```
sauloperez: isn't this callback then redundant? `after_save` will also be called on creation.

app/models/spree/credit_card_decorator.rb
```diff
@@ -14,6 +14,9 @@
...
+  after_save :ensure_default, if: :is_default_changed?
```
sauloperez: Is `:is_default_changed?` something AR adds??

app/models/spree/credit_card_decorator.rb
```diff
@@ -25,4 +28,16 @@ def has_payment_profile? # rubocop:disable Style/PredicateName
   def save_requested_by_customer?
     !!@save_requested_by_customer
   end
+
+  private
+
+  def default_missing?
+    user.credit_cards.where(is_default: true).none?
```
sauloperez: ⚠️ watch out! this will trigger a DB query that will return the whole credit card data from DB, to later check just if it exists with Enumerable's `none?`. So all the data processing and networking cost will be in vain. Whereas, if you use [exists?](https://apidock.com/rails/ActiveRecord/Base/exists%3f/class) you'll get only the boolean you care about here.

See both queries

**none?**

```ruby
irb(main):004:0> user.credit_cards.where(is_default: true).to_sql
=> "SELECT \"spree_credit_cards\".*
FROM \"spree_credit_cards\" 
WHERE \"spree_credit_cards\".\"user_id\" = 1
  AND \"spree_credit_cards\".\"is_default\" = 't'"
```

**exists?**

```ruby
irb(main):005:0> user.credit_cards.exists?(is_default: true).to_sql
SELECT 1 AS one
FROM "spree_credit_cards"
WHERE "spree_credit_cards"."user_id" = 1
  AND "spree_credit_cards"."is_default" = 't' LIMIT 1
```

app/views/spree/users/_saved_cards.html.haml
```diff
@@ -3,11 +3,14 @@
...
     %th= t(:card_number)
     %th= t(:card_expiry_date)
+    %th= t(:default?)
     %th= t(:delete?)
   %tr.card{ id: "card{{ card.id }}", ng: { repeat: "card in savedCreditCards" } }
     %td.brand{ ng: { bind: '::card.brand' } }
     %td.number{ ng: { bind: '::card.number' } }
     %td.expiry{ ng: { bind: '::card.expiry' } }
+    %td.is-default
+      %input{ type: 'radio', name: 'default_card', ng: { model: 'card.is_default', change: 'setDefault(card)', value: "true"} }
```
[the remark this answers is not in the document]
sauloperez: Or perhaps just a label like `Use this card for all upcoming transactions` rather than just `Default?`

app/models/spree/credit_card_decorator.rb
```diff
@@ -25,4 +28,16 @@ def has_payment_profile? # rubocop:disable Style/PredicateName
...
+  end
+
+  def ensure_default
```
sauloperez: I had to read the method's implementation to actually understand what it is about. Is there a better name? What about `ensure_single_default_card`?

app/models/spree/credit_card_decorator.rb
```diff
@@ -25,4 +28,16 @@ def has_payment_profile? # rubocop:disable Style/PredicateName
...
+    return unless user
+    return unless is_default? || default_missing?
+    user.credit_cards.update_all(['is_default=(id=?)', id])
+  end
```
sauloperez: Unless we don't add the constraint at DB level nothing stops us from having two cards with `is_default = true` due to race conditions. Adding a unique index on `is_default`for true values will do it. Check https://stackoverflow.com/a/28167225

sauloperez:
Although I commented some details, looks good overall 👍

app/assets/javascripts/darkswarm/controllers/checkout/payment_controller.js.coffee
```diff
@@ -9,6 +9,10 @@ Darkswarm.controller "PaymentCtrl", ($scope, $timeout, savedCreditCards, Dates)
   $scope.secrets.card_month = "1"
   $scope.secrets.card_year = moment().year()
 
+  for card in (savedCreditCards || []) when card.is_default
```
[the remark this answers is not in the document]
oeoeaio: :+1

app/models/spree/credit_card_decorator.rb
oeoeaio: The trick is is that `is_default_changed?` doesn't seem to return true on create, so this was the best way I could think of to achieve what I needed: always run on create, and then any time the value is changed.

app/models/spree/credit_card_decorator.rb
oeoeaio: Yep, the name of the attribute followed by `_changed?` tells you whether it is being updated.
http://api.rubyonrails.org/v3.2.19/classes/ActiveModel/Dirty.html

app/models/spree/credit_card_decorator.rb
oeoeaio: Nice pickup, thanks! Should have actually looked at the ActiveRecord API...

app/models/spree/credit_card_decorator.rb
oeoeaio: No, `user` it not mandatory on `credit_card`. A user can checkout using a credit card without opting to store it against their account. In this case a new card object is created (as the source for the payment) but it is not directly linked to the user's account.

app/views/spree/users/_saved_cards.html.haml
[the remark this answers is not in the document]
oeoeaio: I have got a tool-tip set up in the next PR, that seemed like a better place for it. I will add a check item to #2275 to ask testers to ensure that UX of understanding the use case of default card is clear.

spec/controllers/spree/credit_cards_controller_spec.rb
```diff
@@ -69,6 +69,53 @@
...
+          expect(response).to redirect_to spree.unauthorized_path
+        end
+      end
+
+      context "and the card is owned by the user" do
+        before { card.update_attribute(:user_id, user.id) }
+
+        context "when the update completes successfully" do
+          it "renders a serialized copy of the updated card" do
+            expect{ put :update, params }.to change { card.reload.is_default }.to(true)
```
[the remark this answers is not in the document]
oeoeaio: Just a sanity check while I was working on the code. I can delete it you think it is messy, but I don't think it is likely to cause us any problems...

spec/features/consumer/account/cards_spec.rb
```diff
@@ -20,28 +21,52 @@
...
+        expect(find_field('default_card')).to be_checked
       end
 
+      within(".card#card#{card2.id}") do
+        expect(page).to have_content card2.cc_type.capitalize
+        expect(page).to have_content card2.last_digits
+        expect(find_field('default_card')).to_not be_checked
+      end
+
+      # Allows switching of default card
```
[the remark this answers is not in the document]
oeoeaio: I prefer this approach with integration specs, particularly when the full UI is being loaded and interacted with using Capybara. It's just a more pragmatic approach to testing the things that need to be tested. In an ideal world with infinite resources, yes, we would load up the entire app to test every small thing in a feature spec, but if we did that our test suite would take 8 hours to run. I prefer to use a "smoke test" to check that a bunch of stuff is generally working on a page, and then have a couple of dedicated specs for really vital/important features. The use of the comments is designed to still allow other developers to understand what is being tested within each section, so that they can understand what is broken. IMHO this is not really different to the description of a test example, except that you don't get the message in your test output, you have to go to the file find it. I think this is an acceptable tradeoff for the reduced run time of the test suite. 

Obviously it's a totally different with unit specs, and even integration specs which don't need to load the entire app in order to run (controller specs/request specs). In those cases absolutely we should have dedicated examples which test one thing in a structured way.

spec/models/spree/credit_card_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+          context "without specifying it as the default" do
+            it "keeps the existing default" do
+              expect(card1.reload.is_default).to be true
+              expect(card2.reload.is_default).to be false
+            end
+          end
+
+          context "and I specify it as the default" do
+            let(:attrs) { { user: user, is_default: true } }
```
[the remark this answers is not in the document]
oeoeaio: The card needs to have a user or the :ensure_default, callback won't run. I don't think the credit card factory will create a user by default, and it comes from spree, so we would have to alter it to get it to do that. Do you think that is worth it?

oeoeaio:
Thanks so much for your review @sauloperez! So speedy!

spec/features/consumer/account/cards_spec.rb
```diff
@@ -4,7 +4,8 @@
   include AuthenticationWorkflow
   describe "as a logged in user" do
     let(:user) { create(:user) }
-    let!(:card) { create(:credit_card, user_id: user.id, gateway_customer_profile_id: 'cus_AZNMJ') }
+    let!(:card) { create(:credit_card, user_id: user.id, gateway_customer_profile_id: 'cus_AZNMJ', is_default: true) }
+    let!(:card2) { create(:credit_card, user_id: user.id, gateway_customer_profile_id: 'cus_FDTG') }
```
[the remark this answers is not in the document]
oeoeaio: Only problem with this is that the spec tests the ability to switch the default card, so that by the end of the spec the `non_default_card` is in fact the default card... I will make the change anyway.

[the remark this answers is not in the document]
oeoeaio: Done

app/models/spree/credit_card_decorator.rb
oeoeaio: Done

app/models/spree/credit_card_decorator.rb
oeoeaio: Trying to think of a better way of achieving this. If I add a unique constraint on `user_id` and `is_default` then I can only have a maximum of two cards per user, because a third card (ie. the second with a `is_default=false` for the user) will break the constraint, right? Or does `false` behave like `null` in unique indexes, in that they don't break the constraint? I will check...

oeoeaio: Looks like it is not possible to create partial indexes until rails `v4.0`...there is no support for the `where` option in `v3.2.22`

I'm trying to understand the mechanism by which we would end up with multiple card using this logic. There is only one query. Are you suggesting that if two queries are submitted to postgres at the same time by two separate workers, that they may write to the same set of rows at the same time, leaving them in a mess? I don't understand the full details of postgres' row locking, but I would have thought that this was protected against. If not, I think the best solution is probably a table lock? That way we can be sure that only one query is being executed by postgres at a time.

mkllnk: I think it's better to avoid any issues like that with a better database design. So one user can have one default card. The standard way is to store the default card id as a column on the user table. But if we did that with everything, the user table become huge and we don't always need to know the user's default card when we load a user. So the alternative is to create a new table with the columns `user_id` and `card_id`. In this case you can make both of them unique via an index, because a user should have only one default and a card belongs to only one user. This approach also avoids changing a Spree table. It also means that the lookup of the default card of a user is completely indexed. Otherwise the database has to look at all cards of a user and filter by the `is_default` column.

oeoeaio: I like that approach, but I'm not sure that Rails is smart enough to magically manage a join table as a `has_one` association in both directions. We might need to add a model for the join table so that we can just use `has_one, through`. `UserDefaultCard`?

oeoeaio: Hm, I'm playing around with this idea and it seems to be resulting in much more pollution of the user model. I'll persist and see if I can pull it back a bit.

spec/controllers/spree/credit_cards_controller_spec.rb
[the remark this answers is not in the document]
sauloperez: To me it's more about the maintenance cost this lines will have just because we weren't 100% confident when writing the test. I'd like to hear other opinions though.

spec/models/spree/credit_card_spec.rb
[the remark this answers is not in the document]
sauloperez: Agree. Better to avoid that alteration.

sauloperez:
Glad you liked it. Sometimes I feel like I give too much feedback

app/models/spree/credit_card_decorator.rb
oeoeaio: @sauloperez @mkllnk

Ok, I've had a really good crack at implementing Maikel's idea, but I am really fighting rails pretty hard. Doing a join table for a `has_one` just doesn't seem to be working for me. This is where I've ended up: override the setter method on a `has_one through` association.😱 I got here because the default implementation of `default_card=`doesn't update the record in the join table, it just creates a new one and nulls the `user_id` on the old record. 

````ruby
class UsersDefaultCard < ActiveRecord::Base
  belongs_to :user, class_name: 'Spree::User'
  belongs_to :default_card, class_name: 'Spree::CreditCard'
end

Spree.user_class.class_eval do
 ...
  has_one :users_default_card
  has_one :default_card, through: :users_default_card
 ...
  def default_card=(card)
    default = if users_default_card
      users_default_card
    else
      build_users_default_card
    end
    default.update_attributes(default_card_id: card.id)
    self[:users_default_card] = default
    self[:default_card] = card
  end
...
end
````

This doesn't feel like an improvement to me, any suggestions?

sauloperez: I wouldn't go that way then. Let's start with what you implemented.

> I'm trying to understand the mechanism by which we would end up with multiple default cards using this logic.

If the DB doesn't enforce the data integrity chances are that the data will end up in a bad state due to [race conditions](https://en.wikipedia.org/wiki/Race_condition).

oeoeaio: I know what a race condition is, I am just trying to understand how we would actually end up with one, so that I can try to prevent it further up the chain. IMO db constraints should really be a last resort protection against race conditions, not the primary way they are dealt with.

sauloperez: I cannot think of any other way other than DB constraints. You can't control how and when the requests are processed by the unicorn workers.

oeoeaio: @mkllnk I had a go at your suggestion but I really couldn't get it to work easily. Mainly I couldn't work out how to set a has_one association using `update_attributes`.

Are you both ok if we go with the existing version, given that there doesn't seem to be an obvious way of adding an appropriate db-level constraint?

app/assets/javascripts/darkswarm/controllers/checkout/payment_controller.js.coffee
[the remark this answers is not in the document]
oeoeaio: Done

app/models/spree/credit_card_decorator.rb
mkllnk: Yes, that's okay. As Kirsten said the other day: perfect is the enemy of done. It's very unlikely that this causes a problem. I mean it's unlikely that somebody tries to set two different cards as default within seconds and then complains about the wrong card being charged.

oeoeaio:
Ok, moving this to test ready unless, let me know if any objections.

oeoeaio:
FYI I disconnected this from #2085 in ZenHub because it was moving #2275 and #2276 as well. I have kept #2276 connected as it is the last of the three PRs required to close #2085.

sstead:
Testing notes

Basically:
- [ ] as a regular user, I should be able to easily understand which of my saved credit card is the 'default' from the Account page
- [ ] when a new card is added to an account it should become the default if none is already set
- [ ] adding subsequent cards should not change the default card
- [ ] I should be able to update the default card using the interface on the account page

https://docs.google.com/document/d/1CKnjczs693kzpkO-fwvJtQKrleardAC8iMD0Wp4zpVM/edit#

sstead:
@oeoeaio  Testing notes have been updated. All four items are now working.
https://docs.google.com/document/d/1CKnjczs693kzpkO-fwvJtQKrleardAC8iMD0Wp4zpVM/edit

PR #2275, opened 2018-05-10
Subs: Allow Charges

### LINKED ISSUE #2085: Customers authorise the use of credit cards by shops

oeoeaio:
#### What? Why?

Part 2 of 3 PRs to close #2085
This PR just deals with the permissions layer of the change. It adds a field to `Customer` called `allow_charges` which determines whether a shop will be able to charge the customers default card for services (see part 3).

#### What should we test?

No services actually implement use of this permission layer yet, so all we can really test is that the customer is able to alter the set of shop who are (theoretically) able to charge their card.

- [ ] As a regular user, I should be able to see a list of the shops which are currently able to charge my default credit card for services by visiting my account page
- [ ] I should be able to understand the implications of granting charge permissions
- [ ] I should be able to grant or revoke permission to charge my default card from the interface

#### Release notes

Customers can now authorise individual shops to charge their default credit card for services provide through the OFN (such as automatic subscription payments)

#### Dependencies

This PR depends on #2274, so don't worry about testing it until that has gone through, just putting this here for preliminary code review if anyone gets a chance.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -1,0 +1,13 @@
+# Base controller for OFN's API
+# Includes the minimum machinery required by ActiveModelSerializers
+module Api
+  class BaseController < Spree::Api::BaseController
+    # Need to include these because Spree::Api::BaseContoller inherits
+    # from ActionController::Metal rather than ActionController::Base
+    # and they are required by ActiveModelSerializers
+    include ActionController::Serialization
+    include ActionController::UrlFor
+    include Rails.application.routes.url_helpers
+    use_renderers :json
+  end
+end
--- a/app/controllers/api/customers_controller.rb
+++ b/app/controllers/api/customers_controller.rb
@@ -1,0 +1,16 @@
+module Api
+  class CustomersController < Spree::Api::BaseController
+    respond_to :json
+
+    def update
+      @customer = Customer.find(params[:id])
+      authorize! :update, @customer
+
+      if @customer.update_attributes(params[:customer])
+        render text: @customer.id, :status => 200
+      else
+        invalid_resource!(@customer)
+      end
+    end
+  end
+end
--- a/app/controllers/api/customers_controller.rb
+++ b/app/controllers/api/customers_controller.rb
@@ -1,13 +1,16 @@
 module Api
-  class CustomersController < Spree::Api::BaseController
-    respond_to :json
+  class CustomersController < BaseController
+    def index
+      @customers = current_api_user.customers
+      render json: @customers, each_serializer: CustomerSerializer
+    end
 
     def update
       @customer = Customer.find(params[:id])
       authorize! :update, @customer
 
       if @customer.update_attributes(params[:customer])
-        render text: @customer.id, :status => 200
+        render json: @customer, serializer: CustomerSerializer, status: 200
       else
         invalid_resource!(@customer)
       end
--- a/app/controllers/api/customers_controller.rb
+++ b/app/controllers/api/customers_controller.rb
@@ -1,7 +1,7 @@
 module Api
   class CustomersController < BaseController
     def index
-      @customers = current_api_user.customers
+      @customers = current_api_user.customers.of_regular_shops
       render json: @customers, each_serializer: CustomerSerializer
     end
 
--- a/app/serializers/api/customer_serializer.rb
+++ b/app/serializers/api/customer_serializer.rb
@@ -1,0 +1,5 @@
+module Api
+  class CustomerSerializer < ActiveModel::Serializer
+    attributes :id, :enterprise_id, :name, :code, :email, :allow_charges
+  end
+end
--- a/app/views/shared/components/_show_profiles.html.haml
+++ b/app/views/shared/components/_show_profiles.html.haml
@@ -1,6 +1,6 @@
 .small-12.medium-6.columns.text-right
   .profile-checkbox
-    %button.button.secondary.tiny.help-btn.ng-scope{:popover => t(:components_profiles_popover, sitename: Spree::Config[:site_name]), "popover-placement" => "left"}><
+    %button.button.secondary.tiny.right.help-btn.ng-scope{:popover => t(:components_profiles_popover, sitename: Spree::Config[:site_name]), "popover-placement" => "left"}><
       %i.ofn-i_013-help
     %label
       %input{"ng-model" => "show_profiles", type: "checkbox", name: "profile"}
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2740,5 +2740,8 @@ See the %{link} to find out more about %{sitename}'s features and to start using
       saved_cards:
         default?: Default?
         delete?: Delete?
+      cards:
+        authorised_shops: Authorised Shops
+        authorised_shops_popover: This is a list of shops which are permitted to charge your default credit card for OFN services (eg. subscriptions). Your card details will be kept secure and will not be shared with shop owners.
     localized_number:
       invalid_format: has an invalid format. Please enter a number.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2743,5 +2743,6 @@ See the %{link} to find out more about %{sitename}'s features and to start using
       cards:
         authorised_shops: Authorised Shops
         authorised_shops_popover: This is the list of shops which are permitted to charge your default credit card for any subscriptions (ie. repeating orders) you may have. Your card details will be kept secure and will not be shared with shop owners. You will always be notified when you are charged.
+        saved_cards_popover: This is the list of cards you have opted to save for later use. Your 'default' will be selected automatically when you checkout an order, and can be charged by any shops you have allowed to do so (see right).
     localized_number:
       invalid_format: has an invalid format. Please enter a number.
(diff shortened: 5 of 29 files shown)


sauloperez:
Is this something we really need in a first iteration? Might be too late to ask but it's probably something we could have done after seeing the feature in action.

oeoeaio:
@sauloperez Without this permissions layer a manager of a shop can effectively charge any saved card in the system, simply by adding the user as a customer of their shop, and then setting up a subscription for that customer. We need this to ensure that only customers that have explicitly authorised a shop to charge their card can do so.

sauloperez:
> We need this to ensure that only customers that have explicitly authorized a shop to charge their card can be charged.

That is what I'm questioning here. It is nothing the feature can't work without and perhaps we could invest the resource to do other more critical things.

oeoeaio:
@sauloperez Ok, I think that allowing any shop to charge any customer for any arbitrary amount is a really bad idea. Remember than anyone can set up a shop. Particularly since we are having this conversation in public, anyone could stumble on this thread and realise that all they have to do is go to their local OFN instance, set up a shop and charge any user with a stored card as much as they like. This seems like something we can't allow to happen. I realise that in reality Stripe would pick these up as dodgy transactions and would withhold the funds, but I still don't think it's a good look.

Also, the work is done. Let's just merge it?

sauloperez:
> Also, the work is done

Of course, that's why I would prefer to have such discussions beforehand although it probably happened and I wasn't aware of it.

config/locales/en.yml
```diff
@@ -1148,6 +1148,9 @@ en:
   saving_credit_card: Saving credit card...
   card_has_been_removed: "Your card has been removed (number: %{number})"
   card_could_not_be_removed: Sorry, the card could not be removed
+  default?: Default?
+  authorised_shops: Authorised Shops
```
mkllnk: It would be better to move these into their scope. It would look like `spree.users.cards.authorised_shops`, but I'm not 100% sure. And in the view: `t('.authorised_shops')`

spec/features/consumer/account/cards_spec.rb
```diff
@@ -41,7 +43,19 @@
       click_link I18n.t(:delete)
 
       expect(page).to have_content I18n.t(:card_has_been_removed, number: "x-#{card.last_digits}")
+<<<<<<< HEAD
```
mkllnk: Alert! Unresolved conflict! ;-)

app/views/shared/components/_show_profiles.html.haml
```diff
@@ -1,6 +1,6 @@
 .small-12.medium-6.columns.text-right
   .profile-checkbox
-    %button.button.secondary.tiny.help-btn.ng-scope{:popover => t(:components_profiles_popover, sitename: Spree::Config[:site_name]), "popover-placement" => "left"}><
+    %button.button.secondary.tiny.right.help-btn.ng-scope{:popover => t(:components_profiles_popover, sitename: Spree::Config[:site_name]), "popover-placement" => "left"}><
```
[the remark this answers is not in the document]
oeoeaio: Sorry, I'm not sure what you mean. The other instances of `.tiny` should not be affected by this PR?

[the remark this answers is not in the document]
mkllnk: Github is showing this diff for this commit:

screenshot from 2018-05-30 14-36-26

Doesn't that mean that `tiny` doesn't float right any more? In this line you add `right` to the button to make it float right. But all other elements of class `tiny` stop floating, right?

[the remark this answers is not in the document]
oeoeaio: Yeah, you’re right. What am I doing?!?!

oeoeaio:
FYI I disconnected this from #2085 in ZenHub because it was moving #2275 and #2276 as well. I have kept #2276 connected as it is the last of the three PRs required to close #2085.

mkllnk:
@sauloperez I'm a bit confused that you approve a pull request that has a few unresolved issues. Are you saying that you approve of the code apart from the issues I pointed out already?

Since this got moved into code review accidentally (with the associated issue), I will move it back into dev.

sauloperez:
> Are you saying that you approve of the code apart from the issues I pointed out already?

Yes that is exactly it.

config/locales/en.yml
```diff
@@ -1148,6 +1148,9 @@ en:
   saving_credit_card: Saving credit card...
   card_has_been_removed: "Your card has been removed (number: %{number})"
   card_could_not_be_removed: Sorry, the card could not be removed
+  default?: Default?
```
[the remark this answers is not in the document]
oeoeaio: Fixed

config/locales/en.yml
oeoeaio: Done

spec/features/consumer/account/cards_spec.rb
oeoeaio: Fixed

app/views/shared/components/_show_profiles.html.haml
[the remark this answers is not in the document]
oeoeaio: I've worked out what is happening with this. The `.tiny` selector you can see in that image is actually within a `.button.help-btn` selector.  So none of the other uses of `.tiny` (as documented in your original comment above) are affected by this change. So I think this can be marked as resolved?

oeoeaio:
@mkllnk I think everything should now be resolved with this, if you could have one more quick check and approve when you get a chance, that would be awesome.

app/controllers/api/customers_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  class CustomersController < Spree::Api::BaseController
+    respond_to :json
+
+    def update
+      @customer = Customer.find(params[:id])
+      authorize! :update, @customer
+
+      if @customer.update_attributes(params[:customer])
+        render text: @customer.id, :status => 200
```
mkllnk: Why not `status: 200`? Do you even need the status? Isn't `200` the default?

spec/controllers/api/customers_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    render_views
+
+    let(:user) { create(:user) }
+    let(:customer) { create(:customer, user: user) }
+    let(:params) { { format: :json, id: customer.id, customer: { code: '123' } } }
+
+    describe "#update" do
+      context "as a user who is not associated with the customer" do
+        before do
+          allow(controller).to receive(:spree_current_user) { create(:user) }
```
mkllnk: Why do you create a new user here instead of using `user` like below? Ah, you want another user to test the permission. I would probably add `let(:other_user)`.

app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,13 @@
+# Base controller for OFN's API
```
mkllnk: Very nice. One more layer between us and Spree. :-)

spec/controllers/api/customers_controller_spec.rb
```diff
@@ -3,13 +3,32 @@
...
     let(:user) { create(:user) }
-    let(:customer) { create(:customer, user: user) }
-    let(:params) { { format: :json, id: customer.id, customer: { code: '123' } } }
+
+    describe "index" do
+      let!(:customer1) { create(:customer) }
+      let!(:customer2) { create(:customer) }
+
+      before do
+        user.customers << customer1
```
mkllnk: This modelling reads so weird: "A user has customers." And it doesn't mean that the user is an enterprise. ;-) Maybe a better name for customer would be *customer relationship*, but that's not as quick to write and read... It's probably fine as it is. But I do find it confusing to see a variable called *customer* and it not being a user. It's beyond the scope of this pull request anyway.

app/serializers/api/customer_serializer.rb
```diff
@@ -0,0 +1,3 @@
+class Api::CustomerSerializer < ActiveModel::Serializer
```
mkllnk: CodeClimate picked this one up.

> Use nested module/class definitions instead of compact style.

oeoeaio:
I have fixed the CodeClimate issue.

The spec failures will be fixed when I rebase this after #2274 goes through.

sstead:
**Testing notes**

Issues
- The ? text is sometimes too far to the right to see it fully
- When a shop is added to a customer's list by the way of the shop adding the customer to their list manually, the customer can't see that shop's name - the field shows as blank to them

Possible improvement
- Put more specific details in the ? box. 
- Maybe add a ? box to the default column.

Things working
- allowing and removing permission seems to work smoothly, with 'changes saved' messages
- The interface can switch to 'add a card' and then back again to the 'allow charges' table.
- the shops list is updated according to which shop's customer list that customer is on.
- All in all it seems quite logical.

https://docs.google.com/document/d/1IVUKVE5w1cG9hvuoevJpk8Cp_2sEgtXMEPHWhIgIUIQ/edit#

oeoeaio:
I broke some tests...will fix asap...

sstead:
**Testing Notes**
- Looks good, same as above
https://docs.google.com/document/d/1IVUKVE5w1cG9hvuoevJpk8Cp_2sEgtXMEPHWhIgIUIQ/edit?pli=1

PR #2276, opened 2018-05-10
Subs: Use Default Card for Subscriptions

### LINKED ISSUE #2085: Customers authorise the use of credit cards by shops

oeoeaio:
#### What? Why?

Part 3 of 3 PRs to close #2085 
This PR contains the implementation of the tools added in the previous two PRs. It replaces the existing logic for charging customers for subscriptions and for validating the payment method selection on a subscription. The last two commits contain cleanup code, removing actions, routes and attributes that are no longer required. 🔥 

#### What should we test?

Main things to test are:

- [ ] As an enterprise use who wants to set up or edit a subscription that uses Stripe as the payment method, I am notified when it is not possible to charge the customer (for whatever reason, be it that they have no default card or that they have not granted permission to charge to the shop)
- [ ] Basic testing that subscriptions which use a Stripe payment method are processed successfully when the order cycle closes
- [ ] Need to test that something sensible happens when charging the customer fails when the order cycle closes too (for example, when the customer revokes permission to charges while the order cycle is open)

#### Release notes

No real release notes for this, as it is part of basic Subscriptions functionality, which hasn't been released yet (STILL!)


app/assets/javascripts/admin/subscriptions/controllers/details_controller.js.coffee
```diff
@@ -1,38 +1,43 @@
...
   $scope.$watch "subscription.payment_method_id", (newValue, oldValue) ->
     return if !newValue?
     paymentMethod = ($scope.paymentMethods.filter (pm) -> pm.id == newValue)[0]
     return unless paymentMethod?
-    if paymentMethod.type == "Spree::Gateway::StripeConnect"
-      $scope.cardRequired = true
-    else
-      $scope.cardRequired = false
-      $scope.subscription.credit_card_id = null
+    $scope.cardRequired = (paymentMethod.type == "Spree::Gateway::StripeConnect")
```
sauloperez: I feel this `type` property is way too coupled to the underlying implementation. Is this Spree imposes?

app/assets/javascripts/admin/subscriptions/controllers/details_controller.js.coffee
```diff
@@ -1,38 +1,43 @@
...
-      delete response.bill_address.id
-      delete response.ship_address.id
-      angular.extend($scope.subscription.bill_address, response.bill_address)
-      angular.extend($scope.subscription.ship_address, response.ship_address)
-      $scope.shipAddressFromBilling() unless response.ship_address.address1?
+  $scope.loadCustomer = ->
+    params = { id: $scope.subscription.customer_id }
+    params.ams_prefix = 'subscription' unless $scope.subscription.id
+    $scope.customer = CustomerResource.get params, (response) ->
+      for address in ['bill_address','ship_address']
```
sauloperez: I believe [forEach](https://docs.angularjs.org/api/ng/function/angular.forEach) is better suited for this. It feels more natural in a functional language like JS.

app/controllers/admin/customers_controller.rb
```diff
@@ -87,5 +75,9 @@ def user_can_create_customer?
       spree_current_user.admin? ||
         spree_current_user.enterprises.include?(@customer.enterprise)
     end
+
+    def ams_prefix_whitelist
```
sauloperez: where's this whitelist used?

app/serializers/api/admin/customer_serializer.rb
```diff
@@ -1,5 +1,6 @@
 class Api::Admin::CustomerSerializer < ActiveModel::Serializer
   attributes :id, :email, :enterprise_id, :user_id, :code, :tags, :tag_list, :name
+  attributes :allow_charges, :default_card_present?
```
sauloperez: `:allow_charges` should go in #2275 , right?

app/serializers/api/admin/customer_serializer.rb
```diff
@@ -18,4 +19,9 @@ def tags
       tag_rule_map || { text: tag, rules: nil }
     end
   end
+
+  def default_card_present?
+    return unless object.user
```
sauloperez: shouldn't always a customer be associated to a user?

EDIT now I recall having to manually update customer records to associated them to the appropriate user id. So it is not enforced.

```ruby
  create_table "customers", :force => true do |t|
    t.string   "email",           :null => false
    t.integer  "enterprise_id",   :null => false
    t.string   "code"
    t.integer  "user_id"
    ...
  end
```

☝️ `user_id` is nullable 🔴

app/services/subscription_validator.rb
```diff
@@ -76,10 +76,11 @@ def schedule_allowed?
...
 
   def credit_card_ok?
-    return unless payment_method.andand.type == "Spree::Gateway::StripeConnect"
-    return errors.add(:credit_card, :blank) unless credit_card_id
-    return if customer.andand.user.andand.credit_card_ids.andand.include? credit_card_id
-    errors.add(:credit_card, :not_available)
+    return unless customer && payment_method
+    return unless payment_method.type == "Spree::Gateway::StripeConnect"
+    return errors.add(:credit_card, :charges_not_allowed) unless customer.allow_charges
+    return if customer.user.andand.default_card.present?
```
sauloperez: We should add the constraint on customers to have a user always. Having this conditional everytime we use a customer object adds unnecessary complexity.

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -139,99 +139,35 @@ def create_customer(enterprise)
...
+      context "where I manage the customer's enterprise" do
+        render_views
 
-      context "when the customer is not associated with a user" do
-        it "returns with an empty array" do
-          spree_get :cards, params
-          json_response = JSON.parse(response.body)
-          expect(json_response).to eq []
+        before do
+          controller.stub spree_current_user: enterprise.owner
```
sauloperez: please, use RSpec's 3 `allow` instead of `stub`.

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -139,99 +139,35 @@ def create_customer(enterprise)
...
-
-        it "returns with serialized cards for the customer" do
-          spree_get :cards, params
-          json_response = JSON.parse(response.body)
-          expect(json_response).to be_an Array
-          expect(json_response.length).to be 1
-          expect(json_response.first["id"]).to eq credit_card1.id
+        it "renders the customer as json" do
+          spree_get :show, format: :json, id: customer.id
+          expect(JSON.parse(response.body)["id"]).to eq customer.id
```
sauloperez: We have [json_spec](https://github.com/collectiveidea/json_spec) in our Gemfile but we never use it. Do you mind if I remove it?

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -139,99 +139,35 @@ def create_customer(enterprise)
...
-        spree_get :cards, params
-        expect(response).to redirect_to spree.unauthorized_path
-      end
-    end
-
-    context "when no customer with a matching id exists" do
-      before { params.merge!(customer_id: 1) }
+      context "where I don't manage the customer's enterprise" do
+        before do
+          controller.stub spree_current_user: another_enterprise.owner
```
sauloperez: here as well

spec/models/spree/user_spec.rb
```diff
@@ -126,4 +126,32 @@
...
+    context "when the user has no credit cards" do
+      it "returns nil" do
+        expect(user.default_card).to be nil
+      end
+    end
+
+    context "when the user has one credit card" do
+      let!(:card) { create(:credit_card, user: user) }
+
+      it "it should be assigned as the default and be returned" do
```
sauloperez: ⚠️ redundant `it`

spec/models/spree/user_spec.rb
```diff
@@ -126,4 +126,32 @@
...
+        expect(card.reload.is_default).to be true
+        expect(user.default_card.id).to be card.id
+      end
+    end
+
+    context "when the user has more than one card" do
+      let!(:card1) { create(:credit_card, user: user) }
+      let!(:card2) { create(:credit_card, user: user, is_default: true) }
```
sauloperez: IMO naming it `default_card` instead of `card2` will make it easier to understand for the next dev.

spec/serializers/admin/subscription_customer_serializer_spec.rb
```diff
@@ -0,0 +1,22 @@
+describe Api::Admin::SubscriptionCustomerSerializer do
+  let(:distributor) { create(:distributor_enterprise) }
+  let(:delivery) { create(:shipping_method, require_ship_address: true, distributors: [distributor]) }
+  let(:email) { 'some@email.com' }
+  let(:address) { create(:address) }
+  let(:customer) { create(:customer, email: email) }
+  let(:order) { create(:completed_order_with_totals, email: email, user: nil, customer: nil, distributor: distributor) }
```
sauloperez: do we actually need all these models to be persisted in DB for the serializer to work? can we use FactoryBot's `build_stubbed` method instead?

 If we avoid hitting DB we would save precious time (and our test suite is way too slow already) in this single test as the number of objects being used here is considerable.

spec/services/subscription_validator_spec.rb
```diff
@@ -357,19 +356,19 @@ def stub_validations(validator, methods)
           end
 
           context "and the customer is associated with a user" do
-            let(:customer) { instance_double(Customer, user: user) }
+            before { expect(customer).to receive(:user).once { user } }
```
sauloperez: this is should be an `allow` ⚠️

app/serializers/api/admin/subscription_customer_serializer.rb
```diff
@@ -0,0 +1,20 @@
...
+      def ship_address
+        finder.ship_address
+      end
+
+      def finder
+        @finder ||= OpenFoodNetwork::AddressFinder.new(object, object.email)
+      end
+    end
+  end
+end
```
sauloperez: Is this related to the discussion you opened in the community? (I totally forgot about it)

oeoeaio: Yep, I'm still not really sure what to do about it, but this is the current convention, so I just stuck with it. I would quite like to decide whether we are going to stick with this approach or whether we are going to take a different path...

app/assets/javascripts/admin/subscriptions/controllers/details_controller.js.coffee
oeoeaio: Yes, `Spree::Gateway` is implemented using STI, which I don't like, but we are stuck with it unfortunately.

app/controllers/admin/customers_controller.rb
oeoeaio: Used by a method called `render_as_json` to on `Admin::BaseController` which basically sets up a convention whereby a prefix can be submitted in a request and then it is used to find a matching serializer to render the data based on the prefix and the model of the controller. I thought it prudent to maintain a whitelist of serializer prefixes to prevent users from being able to request that data be rendered using some arbitrary serializer.

app/serializers/api/admin/customer_serializer.rb
oeoeaio: Oh, yes, good catch, I wonder how that got there...

app/serializers/api/admin/customer_serializer.rb
oeoeaio: See comment on #2274, we need user_id to be nullable, this is from Spree, not something that we set up

spec/serializers/admin/subscription_customer_serializer_spec.rb
oeoeaio: Yeah, you're right. I am testing stuff that is already covered by the `AddressFinder` spec, I've stubbed out the address finder in the spec, 2.5sec > 0.8 sec.

spec/services/subscription_validator_spec.rb
oeoeaio: Can you explain why using an `expect` is a problem? Is it just because we are testing the same thing for each example?

spec/models/spree/user_spec.rb
oeoeaio: Done

spec/serializers/admin/subscription_customer_serializer_spec.rb
oeoeaio: Done

app/services/subscription_validator.rb
oeoeaio: I tend to agree, but I think that is outside the scope of this work.

spec/controllers/admin/customers_controller_spec.rb
oeoeaio: Sorry, copy paste.

oeoeaio: Done

spec/controllers/admin/customers_controller_spec.rb
oeoeaio: Huh, I didn't even know that was there. I don't mind, I just added a helper method called `json_response` (I think in #2275?), so I am happy to use that, or switch to using `json_spec` if that is a better approach. What do you think?

spec/controllers/admin/customers_controller_spec.rb
oeoeaio: Done

app/assets/javascripts/admin/subscriptions/controllers/details_controller.js.coffee
sauloperez: Oh God... STI...

app/assets/javascripts/admin/subscriptions/controllers/details_controller.js.coffee
sauloperez: Isn't there forEach as well? Goddamn hipsters... :trollface:

app/serializers/api/admin/customer_serializer.rb
sauloperez: :sob:

spec/controllers/admin/customers_controller_spec.rb
sauloperez: I would stick to `json_response` and get rid of a dependency.

spec/services/subscription_validator_spec.rb
sauloperez: It is just that you're stubbing a method here and not asserting anything, like `expect` is meant for.

app/serializers/api/admin/subscription_customer_serializer.rb
sauloperez: Let's take a decision in the community thread then

spec/services/subscription_validator_spec.rb
oeoeaio: What makes you think I'm not asserting something? This expect will still cause the spec to fail if the assertion is not met.

app/assets/javascripts/admin/subscriptions/controllers/details_controller.js.coffee
mkllnk: Lol, I just googled STI: Sexually Transmissible Infections

But you probably mean Single Table Inheritance.

mkllnk:
Hm, all tests fail. I better put this back into dev. @oeoeaio Can you put it into review again once you are happy with it?

oeoeaio:
@mkllnk this is dependent on #2274 and #2275, so the test will fail until they go through and I rebase this branch.

app/assets/javascripts/admin/subscriptions/controllers/details_controller.js.coffee
sauloperez: LOL @mkllnk

sstead:
**Testing notes**
Due to lack of time my test was only 80% as deep as I maybe would have liked, but I spotted a few issues that need work, and I can do a total test when it's next staged (or someone else can take a look)

Issues:
- The systems isn't noticing when a customer has revoked permission of the shop to bill them after the sub is made
- in the 2nd confirmation email to the customer where it should say 'Paid' in the shipping method section, it's saying 'not paid' - this is for customers who have been successfully billed with Stripe.

https://docs.google.com/document/d/1O6WsN2Oa3mRqzNW5XiIsAgZkHYMLjzopYBp6Ya3x_0A/edit#

oeoeaio:
Ok, I've fixed those two issues, hopefully this is ready to test again.

app/controllers/admin/customers_controller.rb
mkllnk: Good call to implement a whitelist.

I could imagine that the ams-prefix is a symptom of a bad API design, but I haven't looked into it much.

mkllnk:

sstead:
**Testing notes:**
- In the backend the shop manager can see that a customer doesn't have a billable credit card/hasn't given permission when they go to edit their subscription .. This is a good start, I'd suggest maybe an error icon so they can see that the order can't be billed _without_ needing to go into edit the subscription
- Subs that use stripe payment methods get processed correctly!... I made a suggestion to make the 'your credit card has been billed' messaging more clear, to make the billing seem more legitimate.
- The messaging for scenarios where a subscription can't be charged to stripe (because customer doesn't have a default card selected or they have revoked permission to the shop) works, but there's room for improvement- such as letting the shop owner and custoomer know about the impending issue in the first email rather than only in the second email.

https://docs.google.com/document/d/1O6WsN2Oa3mRqzNW5XiIsAgZkHYMLjzopYBp6Ya3x_0A/edit#

daniellemoorhead:
@mkllnk @sstead @kirstenalarsen in terms of getting this PR to done:

1. The improvements Sally suggested above can be considered part of incremental improvements we need to put in beyond kicking off beta testing. Sally, can you please add these to the improvements topic on discourse - https://community.openfoodnetwork.org/t/subscriptions-improvements/1349

2. Both Travis and CodeClimate are failing - Maikel, can you look into why this is the case and fix if possible (or if not, explain why as a comment on here). **Note: This is a priority over Email Confirm final PRs.**

easyoptimalhuemul-max-1mb

app/services/subscription_validator.rb
mkllnk: @sauloperez I watched the Rails videos you shared on Slack and one interesting idea there was to return a `Guest` user class when there is no registered user. It would respond to the same methods. In this case, `Guest#default_card` would always return `nil`.

Issue #2018, opened 2017-12-21
[Spree Upgrade] Stock Management


oeoeaio:
Spree 2.0.0 introduces a much more sophisticated model for managing stock levels. Rather than a simple `on_hand` count for each variant, stock is modelled by `StockItems`, which can be held at multiple `StockLocations`.

We need to develop an approach to integrating this more sophisticated model into the OFN.

sauloperez:
https://github.com/spree/spree/commit/2c48bfbaad98b8af35a5e95ba268a057c6973b47 is the migration we're talking about.

sauloperez:
See the discussion :point_right: https://github.com/openfoodfoundation/openfoodnetwork/issues/2012#issuecomment-374969502

luisramos0:
closing, wow!!!

PR #2291, opened 2018-05-17
Remove producers tab from dashboard

### LINKED ISSUE #2289: Remove producers tab from dashboard
## Description

As discussed in https://github.com/openfoodfoundation/openfoodnetwork/issues/2013 the [multienterprise dashboard](https://github.com/openfoodfoundation/openfoodnetwork/blob/46d2d6cb8ac9425485cc9635cc5dbc515583151f/app/views/spree/admin/overview/multi_enterprise_dashboard.html.haml) contains references to `spree_products.count_on_hand` which no longer work with Spree 2.0. That column is gone.

As that dashboard tab is rarely used we decided to remove it completely so that not only makes the upgrade easier but also improves performance and reduces complexity.

This is part of https://github.com/openfoodfoundation/openfoodnetwork/issues/2013

## Expected Behavior

"My Enterprises" lists all enterprises, hubs and producers mixed.

sauloperez:
#### What? Why?

Closes #2289 

captura de pantalla 2018-05-17 a les 13 37 22

## Spree compatibility
I removed the producers tab from the overview view, which allowed us to get rid of `Enterprise#active_products_in_order_cycles` and `Enterprise#supplied_and_active_products_on_hand`. They rely on a database column that is gone in the next Spree version (See #2289 for details).

## Performance

The removal of `enterprises_producers_tab` is also a big performance improvement here. The `overview` view is the one rendered when accessing `/admin` and when being superamin, the peformance degrades so much to the extend of making the server crash in some instances. It's among the worst performant endpoints we have.

captura de pantalla 2018-05-17 a les 13 47 40

As you can see below, said partial was taking more than 50% of the time it takes to render the whole view.

captura de pantalla 2018-05-17 a les 13 43 05
*Data from Katuma's production*

## Refactor

I also took the chance to refactor the `OverviewController` for two reasons: readability and performance. I wasn't easy to understand all the cases hidden within that if/else although the specs helped. At the same time, I optimized the DB queries gain a bit more performance even though 

#### What should we test?

`/admin` needs to be checked with a superadmin and a manager who manages either one or more enterprises. Things should keep working as they do now, except for the producers tab, showing all enterprises (including producers) in a single list.

#### Release notes

Removed the producers tab from the admin dashboard. Now all enterprises are shown in a single list.

Part of the #2018 epic.


app/controllers/spree/admin/overview_controller_decorator.rb
```diff
@@ -1,26 +1,73 @@
...
+
+  # Checks whether the request comes from another admin page or not
+  #
+  # @return [Boolean]
+  def outside_referral
+    !URI(request.referer.to_s).path.match(%r{/admin})
+  end
+
+  # Checks if any of the enterprises owned by the current user doesn't sell to
+  # anyone known yet.
```
oeoeaio: This could perhaps more appropriately be called `incomplete_enterprise_registration?`.

And the comment could be something like: `Checks that all of the enterprises owned by the current user have a 'sells' property specified, which indicates that the registration process has been completed`

app/controllers/spree/admin/overview_controller_decorator.rb
```diff
@@ -1,26 +1,73 @@
...
+  #
+  # @return [Boolean]
+  def unspecified_sales?
+    @unspecified_sales ||= spree_current_user
+      .owned_enterprises
+      .where(sells: 'unspecified')
+      .exists?
+  end
```
oeoeaio: This could perhaps more appropriately be called `incomplete_enterprise_registration?`.

And the comment could be something like: `Checks that all of the enterprises owned by the current user have a 'sells' property specified, which indicates that the registration process has been completed`

app/controllers/spree/admin/overview_controller_decorator.rb
```diff
@@ -1,26 +1,73 @@
...
+      :multi_enterprise_dashboard
     end
   end
+
+  # Returns the list of enterprises the current user is manager of
+  #
+  # @return [ActiveRecord::Relation<Enterprise>]
+  def managed_enterprises
+    spree_current_user.enterprises
+  end
```
oeoeaio: `OverviewController` is great, much easier to understand.

app/views/spree/admin/overview/_enterprises.html.haml
```diff
@@ -1,11 +1,64 @@
...
+              - else
+                %span.icon-warning-sign{ 'ofn-with-tip' => t('.has_no_enterprise_fees', enterprise: enterprise.name) }
+            - else
+              &nbsp;
+          %span.two.columns.omega.right
+            %span.icon-arrow-right
+
+    %a.sixteen.columns.alpha.button.bottom.blue{ href: "#{main_app.admin_enterprises_path}" }
+      = t "spree_admin_overview_enterprises_footer"
+      %span.icon-arrow-right
```
oeoeaio: I'm finding this a bit hard to follow without actually looking at it in the view, but it looks as though we're now not rendering any producers in the dashboard? I guess that is fine, just wanted to make sure that intention was to remove the completely.

spec/controllers/spree/admin/overview_controller_spec.rb
```diff
@@ -2,18 +2,18 @@
...
+      allow(controller).to receive(:spree_current_user).and_return(user)
     end
 
     context "when user owns only one enterprise" do
+      let(:user) { create_enterprise_user }
       let!(:enterprise) { create(:distributor_enterprise, owner: user) }
 
       context "when the referer is not an admin page" do
-        before { @request.env['HTTP_REFERER'] = 'http://test.com/some_other_path' }
+        before { @request.env['HTTP_REFERER'] = 'http://test.com/other_path' }
```
oeoeaio: `not_admin_path` might be clearer?

app/views/spree/admin/overview/_enterprises.html.haml
sauloperez: Oops that's not he idea. I'll check again.

app/views/spree/admin/overview/_enterprises.html.haml
```diff
@@ -25,7 +25,7 @@
         %span.centered.three.columns
           = t "spree_admin_enterprises_fees"
     %div.sixteen.columns.alpha.list
-      - @enterprises.is_distributor.each do |enterprise|
+      - @enterprises.each do |enterprise|
```
oeoeaio: I suppose this is ok, but this view is really designed for shops (distributors) only, just adding non-shops to the view doesn't actually provide any meaningful information for those enterprises that don't sell anything. Sorry I wasn't clearer before, but I think it might be best to either a) only show shops (as you had it before), or b) change the view so that it shows information relevant to all kinds of enterprises. I can't think what this information might be (that is why there were two tabs, showing different information based on the enterprise type), so maybe we should just not show any non-shops.

I don't have a strong feeling about it though to happy with this as is if we should just get this through. I don't think it is used enough for us to stress too much about it.

sauloperez:
>  I don't think it is used enough for us to stress too much about it.

That's exactly it. I'll see if there's something we can do to improve it.

sauloperez:
There isn't much we can do here I think. I see two options:

1 - Removing all icons in producer enterprises rows
2 - show a greyed out :heavy_check_mark: 

I think 1 is the less confusing. I wouldn't go too much in-depth with this now. The performance impact of this page clearly outweighs anything we could do here. Check the latest example below:

performance

myriamboure:
My non expert UX view on this is that I would add some "non applicable" meaning icon on the producer lines to avoid them believe they are supposed to have some payment method, etc. Something like the greyed thing you were thinking about maybe? But I'm happy if we start with this one and just iterate then later on if we realize it is confusing. #agile

oeoeaio:
I think we should just move with this as it is.

sauloperez:
staged on https://staging.katuma.org/

myriamboure:
All good ! Ready to go !
Testing notes: https://docs.google.com/document/d/1kuP-08xRs6nHAJZ_oNdiQLvqd1py2ggPd1-EXr2HpRo/edit#
Sorry I assigned me and didn't do it straight away, I should have assigned me only when starting to et other people test if they can meanwhile. Will do that next time ;-)

PR #2292, opened 2018-05-18
Subs: Allow super admin users to enable subscriptions for enterprises

### LINKED ISSUE #2213: Useless 'Subscriptions' menu item for super-admin
## Description

The subscriptions menu item appears for super admin, even when subscriptions is not enabled for any enterprises. This results in the user being able to see the subscription page but never being able to actually create a subscription. This is because super admin gets access to all menu items, but it is very confusing because the check list does not articulate the solution to the problem (enabling subscriptions for an enterprise), and even if it did there is no way to do this through the ui as subscriptions can only be enabled for an enterprise by manually editing the record in the database.

## Expected Behavior
Either the menu item should not appear for super admin, and they will instead need a developer to manually enable subscription as per other users
OR
We just allow super admin users to enable subscriptions for any enterprise from the ui and add an item to the subscriptions checklist telling them that subscriptions needs to be enabled on an enterprise first.

## Actual Behaviour
'Subscriptions' menu item is visible but it is impossible to create a subscription 

## Steps to Reproduce
1. Log in as super admin on an instance where no enterprises are subscriptions-enabled
2. Click Orders
3. Click Subscriptions
4. The checklist appears but is impossible to satisfy

## Context
This is only being encountered by testers at the moment, as the feature has not been released, but it is something that is easy to fix so we should probably to so.

## Possible Fix
- Allow super-admin to toggle subscriptions functionality from the enterprise edit interface (while subs is still in beta), AND
- Add an item to the subs checklist for super-admin, telling them to first enable subscriptions for the enterprise

oeoeaio:
#### What? Why?

Closes #2213

The subscriptions tab is visible for super admin, but there was no possible way to enable subscriptions for an enterprise from the super admin interface. This has lead to confusion for testers.

This PR adds an element to the admin enterprise form for super admins, allowing them to turn subscriptions on/off for a given enterprise.

#### What should we test?

- [ ] Super admin should be able to enable subscriptions for any enterprise from the Shop Preferences section of the Enterprise form
- [ ] Regular enterprise users should not be able to enable subscriptions

#### Release notes

NO RELEASE NOTES
This feature is yet to be release and so does not need separate notes.


app/views/admin/enterprises/form/_shop_preferences.html.haml
```diff
@@ -77,16 +77,16 @@
...
--#       %label= t '.enable_subscriptions'
--#       %div{'ofn-with-tip' => t('.enable_subscriptions_tip')}
--#         %a= t 'admin.whats_this'
--#     .three.columns
--#       = f.radio_button :enable_subscriptions, true
--#       = f.label :enable_subscriptions, t('.enable_subscriptions_true'), value: :true
--#     .five.columns.omega
--#       = f.radio_button :enable_subscriptions, false
--#       = f.label :enable_subscriptions, t('.enable_subscriptions_false'), value: :false
+- if spree_current_user.admin?
```
sauloperez: Is it too much to ask a `superadmin?` alias method? Please :pray:. It will make it much simpler to understand next time I read it.

oeoeaio: Sure, there are LOTS of uses of this method, if we are going to add a helper method, I feel like I should do them all. In which case it probably becomes a separate PR.

sstead:
**Testing notes**
All good!

https://docs.google.com/document/d/1LbSbenPkOpsog5yrOlAzIY6GhNRzMfX0efD63yxaViY/edit?usp=sharing

PR #2293, opened 2018-05-18
Subs: retain open orders

oeoeaio:
#### What? Why?

This one makes my brain hurt. 

A `ProxyOrder` is a placeholder for an actual order, and belongs to both a subscription and an order cycle. When an order cycle opens, an actual order is instantiated for each proxy order that belongs to that order cycle. A subscription has many proxy orders, which represent the relationship between the subscription and a set of order cycles (as determined by the schedule). When an order cycle or schedule is updated, we call `ProxyOrderSyncer#sync` to make sure that the set of proxy orders on each subscription reflects the set of order cycles described by the schedule of that subscription. Any missing proxy orders are created and any orphaned proxy orders are deleted.

Previously, this method was automatically deleting proxy orders for order cycles which were currently open, or which had already closed. This is confusing and pointless, particularly for open orders, as the customer has already received an email saying that their order had been placed. This PR changes the logic of `ProxyOrderSyncer#sync` so that it only applies to proxy orders for future order cycles.

#### What should we test?

- [ ] Altering an open order cycle should not remove subscription orders which have already been placed, even if the new state of the order cycle means that technically the order should no longer apply (ie. OC close date is changed to after subscription end date, or OC is removed from subscription schedule)
- [ ] Such orders should be confirmed normally when the order cycle closes

#### Release notes

### DIFF
--- a/lib/open_food_network/proxy_order_syncer.rb
+++ b/lib/open_food_network/proxy_order_syncer.rb
@@ -56,9 +56,9 @@ module OpenFoodNetwork
     end
 
     def orphaned_proxy_orders
-      in_range_order_cycle_ids = in_range_order_cycles.pluck(:id)
-      return proxy_orders unless in_range_order_cycle_ids.any?
-      proxy_orders.where('order_cycle_id NOT IN (?)', in_range_order_cycle_ids)
+      order_cycle_ids = active_or_complete_or_in_range_order_cycle_ids
+      return proxy_orders unless order_cycle_ids.any?
+      proxy_orders.where('order_cycle_id NOT IN (?)', order_cycle_ids)
     end
 
     def insert_values
@@ -70,6 +70,10 @@ module OpenFoodNetwork
 
     def not_closed_in_range_order_cycles
       in_range_order_cycles.merge(OrderCycle.not_closed)
+    end
+
+    def active_or_complete_or_in_range_order_cycle_ids
+      in_range_order_cycles.pluck(:id) | order_cycles.active_or_complete.pluck(:id)
     end
 
     def in_range_order_cycles
--- a/lib/open_food_network/proxy_order_syncer.rb
+++ b/lib/open_food_network/proxy_order_syncer.rb
@@ -56,9 +56,10 @@ module OpenFoodNetwork
     end
 
     def orphaned_proxy_orders
-      order_cycle_ids = active_or_complete_or_in_range_order_cycle_ids
-      return proxy_orders unless order_cycle_ids.any?
-      proxy_orders.where('order_cycle_id NOT IN (?)', order_cycle_ids)
+      orphaned = proxy_orders.where(placed_at: nil)
+      order_cycle_ids = in_range_order_cycles.pluck(:id)
+      return orphaned unless order_cycle_ids.any?
+      orphaned.where('order_cycle_id NOT IN (?)', order_cycle_ids)
     end
 
     def insert_values
@@ -70,10 +71,6 @@ module OpenFoodNetwork
 
     def not_closed_in_range_order_cycles
       in_range_order_cycles.merge(OrderCycle.not_closed)
-    end
-
-    def active_or_complete_or_in_range_order_cycle_ids
-      in_range_order_cycles.pluck(:id) | order_cycles.active_or_complete.pluck(:id)
     end
 
     def in_range_order_cycles
--- a/lib/open_food_network/proxy_order_syncer.rb
+++ b/lib/open_food_network/proxy_order_syncer.rb
@@ -19,7 +19,7 @@ module OpenFoodNetwork
       return sync_all! if @subscriptions
       return initialise_proxy_orders! unless @subscription.id
       create_proxy_orders!
-      remove_obsolete_proxy_orders!
+      remove_orphaned_proxy_orders!
     end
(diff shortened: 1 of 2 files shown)


sauloperez:
Ok, I went through your description of the data model a few times (yes my brain does hurt too :troll:) but I haven't checked your changes yet. Does it sound feasible to go find the subscriptions whose schedules are related to a particular order cycle, when that one opens?

So instead of deciding which orders need to be created at the time order cycles and subscriptions are updated, we do it at the time the order cycles are opened. We replace writting placeholders in DB (there must be a better term for that concept) with reading from DB later on.

This saves us from the headache of having to keep proxies consistent and lots of code I pressume. We will need an observer to do that though.

sauloperez:
Feel free to tell me I'm drunk and nothing makes sense :joy:

mkllnk:
When I think of standing orders, I often think of scheduled bank transfers, because that is a model that has been around for a while and is well tested. So:

> I know that I will be going on holidays in two weeks time, and I want to cancel my order for that week.

If we want to stop one particular transfer, I usually need to wait until the previous transfer has been triggered. I then edit the scheduled transfer and change the "Next transfer on" date to a week later. If I don't want to wait, I have to create some manual transfers for the ones that would be skipped by editing the date. I'm not sure if this means dropping features.

> I decide to alter my order to include more vegetables for that particular week well in advance

If I want to transfer more money, I just create another transfer. Within the Open Food Network we have the problem, that a new order can also mean more fees. We don't model *deliveries* and don't know when we could combine two orders for the same date into one. That is something that Ceres Fair Food has solved by combining all orders for the same date and destination.

This is just to play with ideas in our heads. I'm not saying we have to change everything now.

daniellemoorhead:
>Feel free to tell me I'm drunk and nothing makes sense

Don't worry @sauloperez, there's a book for that 👌 

image

oeoeaio:
I am a bit confused about what to do with this one. @sstead the result that you saw was actually the expected result according to the specs. It is definitely possible to implement this that way you thought it should work, but I am wondering if a better approach is just to prevent this from ever happening. There is too much potential for chaos, and very confusing results, so I think it would be better to just tell people "No", at least until we have some indication that this is something that people really need to be able to do. What do you think?

mkllnk:
> you can't remove an open order cycle from a schedule

Can you close an order cycle in a schedule?

sstead:
That could work... If we do not allow users to remove an open OC from a schedule, they can always go into subs and delete all the orders. I'm assuming the main reason why you want to remove an OC from a schedule is to prevent sub orders being made.
Alternatively, we can tell them 'You're removing an open OC with open subscriptions from a schedule, please note that any open sub orders will be kept'- this way they know what the repercussions will be, and can figure out a work around (go in and delete sub orders individually).

Either option seems ok to me.

Yes @mkllnk you can close an OC in a schedule, this triggers confirmation emails to go to all customers with a sub order.

oeoeaio:
Ok, I think I've made it so that once an sub order has been, it will never be automatically cancelled/removed, regardless of what happens to the order cycle or schedule. I think this is more intuitive: the only way to cancel an order once it has been placed is to explicitly cancel it from the subs screen.

lib/open_food_network/proxy_order_syncer.rb
```diff
@@ -56,9 +56,10 @@ def remove_orphaned_proxy_orders!
     end
 
     def orphaned_proxy_orders
-      order_cycle_ids = active_or_complete_or_in_range_order_cycle_ids
-      return proxy_orders unless order_cycle_ids.any?
-      proxy_orders.where('order_cycle_id NOT IN (?)', order_cycle_ids)
+      orphaned = proxy_orders.where(placed_at: nil)
+      order_cycle_ids = in_range_order_cycles.pluck(:id)
+      return orphaned unless order_cycle_ids.any?
```
mkllnk: I was just wondering if we could skip this line, but SQL seems to be stupid enough to not accept an empty collection: `order_cycle_id NOT IN ()` -> `PG::Error: ERROR:  syntax error at or near ")"`. And ActiveRecord deals with it by generating `order_cycle_id NOT IN (NULL)` which doesn't return anything. Any operation with `NULL` always returns `NULL`.

One way to make SQL deal with an empty set is the empty result from a sub query:

```ruby
order_cycle_ids = in_range_order_cycles.select(:id)
orphaned.where("order_cycle_id NOT IN (#{order_cycle_ids.to_sql})")
```

This is saving a query from time to time, but I'm not sure if it's easier to read. I think I'm micro-optimising here. But it would make sense if we used Rails 4 already (saving even another line):

```ruby
orphaned.where.not(order_cycle_id: in_range_order_cycles)
```
Can't wait to upgrade.

sauloperez: Well, I think you will have to @mkllnk :trollface:

sstead:
**Testing notes**
- looks good, all orders are kept, and this is clear in the subs dashboard.
- The only thing maybe missing is a message to the manager 'you're removing an open OC from a schedule, all open subscription orders will be kept active'. This is not essential.

https://docs.google.com/document/d/1upiADU2OamQvJfAmkJ68J1j0UxRNs0JOKfwi9zfCHCk/edit#

PR #2294, opened 2018-05-18
Subs: Confirm date change for Order Cycles with open subscription orders

oeoeaio:
#### What? Why?

#2293 deals with the logic of what happens when an open order cycle with subscription orders is altered in some way. This PR seeks to provide the user with additional information about what they are doing, and to warn them against making unnecessary changes to an open order cycle.

#### What should we test?

- [ ] If a user attempt to alter the dates of an open order cycle with subscription orders, they should be warned that they should avoid doing so if possible.

#### Release notes

Feature is yet to be released, so no separate release notes required.

### DIFF
--- a/app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
@@ -9,10 +9,14 @@ angular.module("admin.orderCycles").directive "changeWarning", (ConfirmDialog) -
     msg = 'admin.order_cycles.date_warning.msg'
     options = { cancel: t(cancel), confirm: t(proceed) }
 
+    isOpen = (orderCycle) ->
+      moment(orderCycle.orders_open_at, "YYYY-MM-DD HH:mm:SS Z").isBefore() &&
+      moment(orderCycle.orders_close_at, "YYYY-MM-DD HH:mm:SS Z").isAfter()
+
     element.focus ->
       count = scope.orderCycle.subscriptions_count
       return if acknowledged
-      return if moment(scope.orderCycle.orders_close_at).isBefore()
+      return unless isOpen(scope.orderCycle)
       return if count < 1
       ConfirmDialog.open('info', t(msg, n: count), options).then ->
         acknowledged = true
--- a/app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
@@ -1,0 +1,19 @@
+angular.module("admin.orderCycles").directive "changeWarning", (ConfirmDialog) ->
+  restrict: "A"
+  scope:
+    orderCycle: '=changeWarning'
+  link: (scope, element, attrs) ->
+    acknowledged = false
+    count = scope.orderCycle.subscriptions_count
+    cancel = 'admin.order_cycles.date_warning.cancel'
+    proceed = 'admin.order_cycles.date_warning.proceed'
+    msg = t('admin.order_cycles.date_warning.msg', n: count)
+    options = { cancel: t(cancel), confirm: t(proceed) }
+
+    element.focus ->
+      return if acknowledged
+      return if moment(scope.orderCycle.orders_close_at).isBefore()
+      return if count < 1
+      ConfirmDialog.open('info', msg, options).then ->
+        acknowledged = true
+        element.siblings('img').trigger('click')
--- a/app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
@@ -4,16 +4,16 @@ angular.module("admin.orderCycles").directive "changeWarning", (ConfirmDialog) -
     orderCycle: '=changeWarning'
   link: (scope, element, attrs) ->
     acknowledged = false
-    count = scope.orderCycle.subscriptions_count
     cancel = 'admin.order_cycles.date_warning.cancel'
     proceed = 'admin.order_cycles.date_warning.proceed'
-    msg = t('admin.order_cycles.date_warning.msg', n: count)
+    msg = 'admin.order_cycles.date_warning.msg'
     options = { cancel: t(cancel), confirm: t(proceed) }
 
     element.focus ->
+      count = scope.orderCycle.subscriptions_count
       return if acknowledged
       return if moment(scope.orderCycle.orders_close_at).isBefore()
       return if count < 1
-      ConfirmDialog.open('info', msg, options).then ->
+      ConfirmDialog.open('info', t(msg, n: count), options).then ->
         acknowledged = true
         element.siblings('img').trigger('click')
--- a/app/controllers/admin/order_cycles_controller.rb
+++ b/app/controllers/admin/order_cycles_controller.rb
@@ -19,7 +19,7 @@ module Admin
       respond_to do |format|
         format.html
         format.json do
-          render_as_json @collection, ams_prefix: params[:ams_prefix], current_user: spree_current_user
+          render_as_json @collection, ams_prefix: params[:ams_prefix], current_user: spree_current_user, subscriptions_counts: subscriptions_counts
         end
       end
     end
@@ -81,7 +81,7 @@ module Admin
     def bulk_update
       if order_cycle_set.andand.save
         respond_to do |format|
-          format.json { render_as_json @order_cycles, ams_prefix: 'index', current_user: spree_current_user }
+          format.json { render_as_json @order_cycles, ams_prefix: 'index', current_user: spree_current_user, subscriptions_counts: subscriptions_counts }
         end
       else
         respond_to do |format|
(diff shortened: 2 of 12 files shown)


app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
```diff
@@ -0,0 +1,19 @@
...
+    cancel = 'admin.order_cycles.date_warning.cancel'
+    proceed = 'admin.order_cycles.date_warning.proceed'
+    msg = 'admin.order_cycles.date_warning.msg'
+    options = { cancel: t(cancel), confirm: t(proceed) }
+
+    element.focus ->
+      count = scope.orderCycle.subscriptions_count
+      return if acknowledged
+      return if moment(scope.orderCycle.orders_close_at).isBefore()
+      return if count < 1
```
sauloperez: doesn't it make sense to collapse these conditionals into a single if statement?

app/serializers/api/admin/index_order_cycle_serializer.rb
```diff
@@ -61,6 +61,10 @@ def delete_path
         admin_order_cycle_path(object)
       end
 
+      def subscriptions_count
+        ProxyOrder.not_canceled.where(order_cycle_id: object.id).count
```
sauloperez: have you checked this doesn't add an N+1?

app/assets/javascripts/admin/order_cycles/directives/change-warning.js.coffee
mkllnk: Then you get:
```ruby
return if acknowledged ||
         moment(scope.orderCycle.orders_close_at).isBefore() ||
         count < 1
```
I actually find Rob's version easier to read, because you can look at each line separately without thinking about logic operators. :man_shrugging: Both valid, I would say.

app/serializers/api/admin/index_order_cycle_serializer.rb
mkllnk: This `count` is on database level. As long as we use indexes for cancelled and order cycle, this should be really fast. I just checked and we don't have an index for `cancelled_at`. For this purpose, we don't need to index all date values. We just want to know if it's `null` or set. I guess we can limit the index with `length: 4` which includes the year.

mkllnk: If someone had 100,000 standing orders with 12 weeks of proxy orders ahead for one order cycle, the records exceed a million and this would become really important.

app/serializers/api/admin/order_cycle_serializer.rb
```diff
@@ -20,6 +20,10 @@ def viewing_as_coordinator
     Enterprise.managed_by(options[:current_user]).include? object.coordinator
   end
 
+  def subscriptions_count
+    ProxyOrder.not_canceled.where(order_cycle_id: object.id).count
```
mkllnk: Second time seeing this code. Maybe we should create `ProxyOrder.count_active(order_cycle)`.

app/serializers/api/admin/index_order_cycle_serializer.rb
oeoeaio: Maybe the issue is more that this query will generate a new query for each order cycle in the collection when the index page is rendered?

Maybe if I defined these associations on `OrderCycle`:
````ruby
has_many :active_proxy_orders, class_name: 'ProxyOrder', conditions: { where('proxy_orders.canceled_at IS NULL') }
has_many :subscriptions, through: :active_proxy_orders
````
Then I could preload the subscriptions associated with all of the order cycles in the collection, and reduce the number of queries required to count subscriptions from N to 1?

That would make this method become:
````ruby
def subscriptions_count
    object.subscriptions.count
end
````
Which would hopefully be an array count, rather than a db count.

Thoughts?

File: app/serializers/api/admin/order_cycle_serializer.rb (hunk 3 above)
Comment (oeoeaio) [reply app/serializers/api/admin/order_cycle_serializer.rb]: Yep, or maybe I should just an a `subscriptions` association to order cycle as per my comment above?

File: app/serializers/api/admin/index_order_cycle_serializer.rb (hunk 2 above)
Comment (mkllnk) [reply app/serializers/api/admin/index_order_cycle_serializer.rb]: I was thinking that db counts are much quicker than array counts. Loading all the subscriptions into objects just to count them sounds very inefficient. If you want to load the count with the query, you probably have to create a scope in which you modify the `select` with your own join and group_by. I'm not sure if there is an ActiveRecords technique for that.

Comment (oeoeaio) [reply app/serializers/api/admin/index_order_cycle_serializer.rb]: > I was thinking that db counts are much quicker than array counts.

Yes definitely when only one order cycle is being serialized, but I think if you serializing 30 order cycles then preloading the subscriptions and counting the ones that apply to each order cycle in ruby is going to be faster than running 30 separate queries (one for each OC)? It's probably not a problem at this stage. I think we should just stick with it the way it is.

Comment (mkllnk) [reply app/serializers/api/admin/index_order_cycle_serializer.rb]: I don't know the implementation of ActiveRecord well enough. It could be that ActiveRecord loads everything from one query, but still creates independent objects. I don't know.

We can either investigate this and test what's actually faster, or we can leave it as is and find out when we have a performance issue.

@sauloperez What's your view? Should we invest more work to find out or can this go through as is?

Comment (sauloperez) [reply app/serializers/api/admin/index_order_cycle_serializer.rb]: It's hard for me to accept the fact that we introduce a clear N+1 query buuuut I agree with 

> we can leave it as is and find out when we have a performance issue.

We have way too many things on our plate, so let's be pragmatic and fix it when it becomes a bottleneck.

Comment (oeoeaio) [reply app/serializers/api/admin/index_order_cycle_serializer.rb]: @sauloperez where in the N+1. This `count` is uses a db `COUNT`, we're not loading the objects into ruby. Am I missing something? I'm really happy to fix it if I know what to do...

Comment (sauloperez) [reply app/serializers/api/admin/index_order_cycle_serializer.rb]: So, if I guessed correctly by the name of this serializer (`IndexOrderCycleSerializer`), we use it to render a list of order cycles. To do that, this serializer gets instantiated for each order cycle we fetched from DB, issuing a query each time `#subscriptions_count` is called.

There we have it; 1 query to fetch N order cycles and N queries to fetch the count of subscriptions. Am I right or did I totally misunderstood how this serializer works?

Comment (oeoeaio) [reply app/serializers/api/admin/index_order_cycle_serializer.rb]: Note for posterity: we decided to resolve this by doing the counts in a single query in the controller and then pass them to the serializer.

**Comment by oeoeaio** [comment]
@sauloperez @mkllnk if you had a moment to look at dcee5bf and confirm that it addresses the concerns you had, that would be great!

File: app/serializers/api/admin/index_order_cycle_serializer.rb [hunk 4]
```diff
@@ -61,6 +61,10 @@ def delete_path
         admin_order_cycle_path(object)
       end
 
+      def subscriptions_count
+        options[:subscriptions_counts][object.id]
```
sauloperez: wouldn't this raise an exception when `@collection.blank?`?

spec/controllers/admin/order_cycles_controller_spec.rb
```diff
@@ -337,5 +337,31 @@ module Admin
...
+          expect(controller.send(:subscriptions_counts)).to eq []
+        end
+      end
+
+      context "when the collection has been set" do
+        let!(:po1) { create(:proxy_order, order_cycle: oc1) }
+        let!(:po2) { create(:proxy_order, order_cycle: oc1) }
+        let!(:po3) { create(:proxy_order, order_cycle: oc2) }
+
+        before { controller.instance_variable_set(:@collection, collection) }
```
sauloperez: IMO we can't afford to do this. As it violates the controller's encapsulation it will break sooner than later. 

this, the need to use `#send` and the fact that we add yet another method that is not directly related to the controller's responsibilities leads me to think that we should rather extract `subscriptions_count` as a `SubscriptionsCount` plain ruby object. This would act as a query object. https://robots.thoughtbot.com/active-record-eager-loading-with-query-objects-and-decorators is good article on the topic.

app/controllers/admin/order_cycles_controller.rb
```diff
@@ -230,6 +230,11 @@ def require_order_cycle_set_params
       render json: { errors: t('admin.order_cycles.bulk_update.no_data') }, status: :unprocessable_entity
     end
 
+    def subscriptions_counts
+      return [] if @collection.blank?
+      ProxyOrder.not_canceled.group(:order_cycle_id).where(order_cycle_id: @collection).count
```
sauloperez: :ok_hand: neat removal of the N+1

app/serializers/api/admin/index_order_cycle_serializer.rb
mkllnk: No, I think `options[:subscriptions_counts]` will be an empty array and this will return `nil`. And maybe it wouldn't even be called, because there is no order cycle to retrieve subscriptions for. But it would be better to return `0` instead of `nil` just in case. The method may be used differently in the future.

spec/controllers/admin/order_cycles_controller_spec.rb
oeoeaio: @sauloperez Ok, see latest commit.

sstead:
**Testing notes**

There's just one thing that could be improved. 
-Currently the warning modal pops up if you try to change the dates of a future OC (in the yellow state, not green) that has subs associated with it. This isn't so necessary because there aren't any open orders attached to future OCs.

Otherwise it works well.

https://docs.google.com/document/d/1AwW8lprPXhAOfEIPe8wS-p28Izjy8eWrIEdBptHOt3E/edit#

sstead:
**Testing notes**
The issue above is fixed. Good to go.

PR #2295, opened 2018-05-18
Subs: Rely on proxy order state to determine whether to confirm or not

oeoeaio:
#### What? Why?

When a user cancels or pauses a subscription while an associated order cycle is open, they are asked whether they would like to keep or cancel the orders. A bug existed whereby even if they selected to keep the order, it would not be confirmed properly at the close of the order cycle, because the logic in the subscription confirm job would ignore any proxy orders associated with a paused or cancelled subscription.

This PR changes that logic to allow each order to determine whether it should be confirmed or not, independent of the state of the underlying subscription. 

#### What should we test?

- [ ] It should be possible to opt to cancel a subscription with an open order, then opt to keep that order. The order should be confirmed at the end of the order cycle.

#### Release notes

### DIFF
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -25,7 +25,6 @@ class SubscriptionConfirmJob
   def proxy_orders
     ProxyOrder.not_canceled.where('confirmed_at IS NULL AND placed_at IS NOT NULL')
       .joins(:order_cycle).merge(recently_closed_order_cycles)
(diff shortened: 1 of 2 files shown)


sstead:
Testing notes
Yay, it works. If the user opts to keep the open order, it will actually get kept. 
https://docs.google.com/document/d/1ykBHTq59X0ADD5V8PmDSwQ2k37NYjyQ24-2julhFUz0/edit?usp=sharing

PR #2296, opened 2018-05-18
Subs: OrderCyclesController refactor

oeoeaio:
#### What? Why?

Just a refactor of the `Admin::OrderCyclesController`. I started this work because I thought I needed it to solve another problem, but I ended up taking a different route. It still seems like a valuable change to make though, so I thought I would submit it and see if there is any appetite for pushing it through. If it all seems to hard we can just bin it.

#### What should we test?

The most annoying thing: all admin order cycle pages should behave exactly as before.

#### Release Notes

### DIFF
--- a/app/controllers/admin/order_cycles_controller.rb
+++ b/app/controllers/admin/order_cycles_controller.rb
@@ -1,5 +1,3 @@
-require 'open_food_network/order_cycle_form_applicator'
-
 module Admin
   class OrderCyclesController < ResourceController
     include OrderCyclesHelper
@@ -41,7 +39,6 @@ module Admin
       @order_cycle_form = OrderCycleForm.new(@order_cycle, params, spree_current_user)
 
       if @order_cycle_form.save
-        OpenFoodNetwork::OrderCycleFormApplicator.new(@order_cycle, spree_current_user).go!
         flash[:notice] = I18n.t(:order_cycles_create_notice)
         render json: { success: true }
       else
@@ -53,10 +50,6 @@ module Admin
       @order_cycle_form = OrderCycleForm.new(@order_cycle, params, spree_current_user)
 
       if @order_cycle_form.save
-        unless params[:order_cycle][:incoming_exchanges].nil? && params[:order_cycle][:outgoing_exchanges].nil?
-          # Only update apply exchange information if it is actually submmitted
-          OpenFoodNetwork::OrderCycleFormApplicator.new(@order_cycle, spree_current_user).go!
-        end
         flash[:notice] = I18n.t(:order_cycles_update_notice) if params[:reloading] == '1'
         render json: { :success => true }
       else
--- a/app/controllers/admin/order_cycles_controller.rb
+++ b/app/controllers/admin/order_cycles_controller.rb
@@ -1,6 +1,4 @@
-require 'open_food_network/permissions'
 require 'open_food_network/order_cycle_form_applicator'
-require 'open_food_network/proxy_order_syncer'
 
 module Admin
   class OrderCyclesController < ResourceController
@@ -9,11 +7,8 @@ module Admin
     prepend_before_filter :load_data_for_index, :only => :index
     before_filter :require_coordinator, only: :new
     before_filter :remove_protected_attrs, only: [:update]
-    before_filter :check_editable_schedule_ids, only: [:create, :update]
     before_filter :require_order_cycle_set_params, only: [:bulk_update]
     around_filter :protect_invalid_destroy, only: :destroy
-    create.after :sync_subscriptions
-    update.after :sync_subscriptions
 
     def index
       respond_to do |format|
@@ -43,11 +38,10 @@ module Admin
     end
 
     def create
-      @order_cycle_form = OrderCycleForm.new(@order_cycle, params)
+      @order_cycle_form = OrderCycleForm.new(@order_cycle, params, spree_current_user)
 
       if @order_cycle_form.save
         OpenFoodNetwork::OrderCycleFormApplicator.new(@order_cycle, spree_current_user).go!
-        invoke_callbacks(:create, :after)
         flash[:notice] = I18n.t(:order_cycles_create_notice)
         render json: { success: true }
       else
@@ -56,14 +50,13 @@ module Admin
     end
 
     def update
-      @order_cycle_form = OrderCycleForm.new(@order_cycle, params)
+      @order_cycle_form = OrderCycleForm.new(@order_cycle, params, spree_current_user)
 
       if @order_cycle_form.save
         unless params[:order_cycle][:incoming_exchanges].nil? && params[:order_cycle][:outgoing_exchanges].nil?
           # Only update apply exchange information if it is actually submmitted
           OpenFoodNetwork::OrderCycleFormApplicator.new(@order_cycle, spree_current_user).go!
         end
-        invoke_callbacks(:update, :after)
         flash[:notice] = I18n.t(:order_cycles_update_notice) if params[:reloading] == '1'
         render json: { :success => true }
       else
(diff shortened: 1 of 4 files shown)


app/controllers/admin/order_cycles_controller.rb
```diff
@@ -43,51 +36,33 @@ def new
...
-      @order_cycle = OrderCycle.find params[:id]
+      @order_cycle_form = OrderCycleForm.new(@order_cycle, params, spree_current_user)
 
-      respond_to do |format|
-        if @order_cycle.update_attributes(params[:order_cycle])
-          unless params[:order_cycle][:incoming_exchanges].nil? && params[:order_cycle][:outgoing_exchanges].nil?
-            # Only update apply exchange information if it is actually submmitted
-            OpenFoodNetwork::OrderCycleFormApplicator.new(@order_cycle, spree_current_user).go!
-          end
-          invoke_callbacks(:update, :after)
```
sauloperez: what about these callbacks? they are not needed anymore?

app/controllers/admin/order_cycles_controller.rb
```diff
@@ -45,49 +45,38 @@ def new
...
 
-      respond_to do |format|
-        if @order_cycle.save
-          OpenFoodNetwork::OrderCycleFormApplicator.new(@order_cycle, spree_current_user).go!
-          invoke_callbacks(:create, :after)
-          flash[:notice] = I18n.t(:order_cycles_create_notice)
-          format.html { redirect_to admin_order_cycles_path }
-          format.json { render :json => { success: true } }
-        else
-          format.html
```
mkllnk: Can you explain why HTML is unnecessary? We never used it? Or can you reference to a commit when we stopped using it, but didn't change this method?

app/services/order_cycle_form.rb
```diff
@@ -1,13 +1,50 @@
-class OrderCycleForm
-  attr_accessor :order_cycle, :params
+require 'open_food_network/permissions'
+require 'open_food_network/proxy_order_syncer'
 
-  def initialize(order_cycle, params)
+class OrderCycleForm
```
mkllnk: I'm happy to extract the logic. I'm just wondering about the name `OrderCycleForm` now. It's not really representing a form, it's containing modelling logic.

app/services/order_cycle_form.rb
```diff
@@ -25,26 +25,50 @@ def save
...
-  def check_editable_schedule_ids
-    return unless params[:order_cycle][:schedule_ids]
-    requested = params[:order_cycle][:schedule_ids].map(&:to_i)
-    @existing_schedule_ids = @order_cycle.persisted? ? @order_cycle.schedule_ids : []
-    permitted = Schedule.where(id: requested | @existing_schedule_ids).merge(permissions.editable_schedules).pluck(:id)
-    result = @existing_schedule_ids
-    result |= (requested & permitted) # add any requested & permitted ids
-    result -= ((result & permitted) - requested) # remove any existing and permitted ids that were not specifically requested
+  def schedule_ids?
+    !!params[:order_cycle][:schedule_ids]
```
mkllnk: Why not `params[:order_cycle][:schedule_ids].present?`?

app/controllers/admin/order_cycles_controller.rb
oeoeaio: As far as I know we never used it.

Rohan's original commit that added this action used json from the very beginning: aacc36ea440d7b84807d5a3b6e6b27a0d3c116d5

spec/controllers/admin/order_cycles_controller_spec.rb
```diff
@@ -96,123 +96,118 @@ module Admin
...
 
-        context "when updating without explicitly submitting exchanges" do
-          let(:form_applicator_mock) { double(:form_applicator) }
+          context "when the page is reloading" do
+            before { params[:reloading] = '1' }
 
-          before do
-            allow(OpenFoodNetwork::OrderCycleFormApplicator).to receive(:new) { form_applicator_mock }
-            allow(form_applicator_mock).to receive(:go!) { nil }
+            it "sets flash message when page is reloading" do
```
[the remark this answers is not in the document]
oeoeaio: 👍 Thanks, nice pickup.

app/controllers/admin/order_cycles_controller.rb
oeoeaio: No, no longer required. They were added by me at the start of the subscriptions work. c2b68ed6b5ba2f033ee734c17aad2c669cce268a

These have been replaced by a method call in `OrderCycleForm#save`.

app/services/order_cycle_form.rb
oeoeaio: Fair enough. I'll admit that it is really only one half of a standard form object (in that it is not being used to render the form itself), but it is still encapsulates the logic required for creating/updating the underlying model, and is very much coupled to the data provided by the submission of the form.

app/services/order_cycle_form.rb
oeoeaio: Is this just a stylistic preference? I don't particularly mind either way.

mkllnk: Yes, I find double negative difficult to read/understand. Your version is pure Ruby though.

spec/controllers/admin/order_cycles_controller_spec.rb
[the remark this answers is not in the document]
oeoeaio: Fixed

sstead:
**Testing notes**

Everything was fine, except a few things...
- There is a error when you use the Advanced settings in edit OC
- Some product import elements are showing - please just check these won't go to production prematurely

https://docs.google.com/document/d/1D3JLjaO3-GL_Y0hRvIqGkmabm5ipvJUzjJeIoY3xlX8/edit?usp=sharing

myriamboure:
Ok @sstead I created an issue with the product import thing you noticed https://github.com/openfoodfoundation/openfoodnetwork/issues/2365 @Matt-Yorkley or @oeoeaio  do you think you can do that in priority? It is already merged so will go in production at next release and users might be disturbed... We can discuss options on the other issue about what to do, I shared two options.

oeoeaio:
Ok, I think I've fixed the first issue, and the product import issue is not related as is being dealt with separately.

app/controllers/admin/order_cycles_controller.rb
```diff
@@ -50,8 +50,11 @@ def update
       @order_cycle_form = OrderCycleForm.new(@order_cycle, params, spree_current_user)
 
       if @order_cycle_form.save
-        flash[:notice] = I18n.t(:order_cycles_update_notice) if params[:reloading] == '1'
-        render json: { :success => true }
+        respond_to do |format|
+          flash[:notice] = I18n.t(:order_cycles_update_notice) if params[:reloading] == '1'
+          format.html { redirect_to main_app.edit_admin_order_cycle_path(@order_cycle) }
```
mkllnk: Aren't we still missing an HTML response in case it can't be saved?

oeoeaio: Probably, but our current `master` doesn't have one either, so this is not a regression.

The only time a html response is require is when the advanced settings form is used, and it currently only allows a single boolean attribute to be altered, so I think the prospect of an error message is pretty remote...

app/services/order_cycle_form.rb
```diff
@@ -0,0 +1,13 @@
...
+  def initialize(order_cycle, params)
+    @order_cycle = order_cycle
+    @params = params
+  end
+
+  def save
+    order_cycle.assign_attributes(params[:order_cycle])
+    order_cycle.save
+  end
+end
```
sauloperez: do we really need this? I don't see the value this brings now. It's possibly the only controller action that's still simple, except for its success logic, which could deserve being extracted.

sstead:
**Testing notes**

- Everything was fine. The advanced settings issue is fixed.
- Import date filter is still showing in Inventory, but this is being fixed elsewhere. 

https://docs.google.com/document/d/1D3JLjaO3-GL_Y0hRvIqGkmabm5ipvJUzjJeIoY3xlX8/edit?usp=sharing

PR #2305, opened 2018-05-21
Ensure domain in SSL header matches request with or without www prefix

### LINKED ISSUE #2302: Embedded shops not working (blank screen) - 1.15.1? or was it earlier?
## Description
I was checking the embedded shop at Black Mountain Food Hub (UK) and noticed it wasn't working. I then went to set one up on staging1.openfood.com.au and it is also not working (blank screen). 

Both shops seem to be working fine in OFN itself

## Actual Behavior
The external website should show an embedded OFN shop

## Steps to Reproduce
Create an open shop front
Follow [instructions](https://ofn-user-guide.gitbook.io/ofn-user-guide-master/advanced-features/shop-setup/embedded-shopfront) to embed it in a wordpress site or other
Ensure the domain is listed in config of the host instance

## Severity
I think it's at least S2 - not many are using this feature, but those that are it's pretty critical. Black Mountain order cycle is open at the moment so they are likely having customers arrive at  a blank page

## Your Environment
Mac
Chrome

## Possible Fix
Could have been in last release? I am not sure who is in best position to confirm when they were last working?

Matt-Yorkley:
#### What? Why?

Closes #2302 

Headers need to include or exclude the www prefix when listing the domain in the `Content-Security-Policy` headers to match the requesting site, instead of just specifying the domain.

#### What should we test?

Embedded shopfronts work on sites with or without "www" in the URL.

#### Release notes

Fixes issue with embedded shopfronts not working on sites with "www" in the URL.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -60,7 +60,7 @@ class ApplicationController < ActionController::Base
     return if embedding_without_https?
 
     response.headers.delete 'X-Frame-Options'
-    response.headers['Content-Security-Policy'] = "frame-ancestors #{embedded_shopfront_referer}"
+    response.headers['Content-Security-Policy'] = "frame-ancestors #{URI(request.referer).host.downcase}"
 
     check_embedded_request
     set_embedded_layout
--- a/spec/requests/embedded_shopfronts_headers_spec.rb
+++ b/spec/requests/embedded_shopfronts_headers_spec.rb
@@ -44,7 +44,7 @@ describe "setting response headers for embedded shopfronts", type: :request do
     context "with a valid whitelist" do
       before do
         Spree::Config[:embedded_shopfronts_whitelist] = "example.com external-site.com"
-        allow_any_instance_of(ActionDispatch::Request).to receive(:referer).and_return('http://www.external-site.com/shop?embedded_shopfront=true')
+        allow_any_instance_of(ActionDispatch::Request).to receive(:referer).and_return('http://external-site.com/shop?embedded_shopfront=true')
       end
 
       it "allows iframes on certain pages when enabled in configuration" do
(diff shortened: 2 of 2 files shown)


myriamboure:
It would be great to review and merge quickly as it is an s2 bug and we want to deploy ASAP. UK already deployed in production if I understand well as it was more like an s1 bug to them as users where trading only through their embedded shopfront so couldn't trade. Ping @sauloperez @mkllnk I'm ready to test and have a user ready to test in a real case whenever this is reviewed and staged.

spec/requests/embedded_shopfronts_headers_spec.rb
```diff
@@ -61,5 +61,20 @@
         expect(response.headers['Content-Security-Policy']).to eq "frame-ancestors 'none'"
       end
     end
+
+    context "with www prefix" do
+      before do
+        Spree::Config[:embedded_shopfronts_whitelist] = "example.com external-site.com"
+        allow_any_instance_of(ActionDispatch::Request).to receive(:referer).and_return('http://www.external-site.com/shop?embedded_shopfront=true')
```
sauloperez: you have access to `request` here so there's no need to use neither `allow_any_instance` nor refer to `ActionDispatch::Request`.

app/controllers/application_controller.rb
```diff
@@ -60,7 +60,7 @@ def enable_embedded_shopfront
     return if embedding_without_https?
 
     response.headers.delete 'X-Frame-Options'
-    response.headers['Content-Security-Policy'] = "frame-ancestors #{embedded_shopfront_referer}"
+    response.headers['Content-Security-Policy'] = "frame-ancestors #{URI(request.referer).host.downcase}"
```
mkllnk: This will work, because the referer got tested for existence before. But it would be nice to have a proper method here.

```ruby
  def csp_frame_ancestors
    return "none" if request.referer.blank?
    URI(request.referer).host.downcase
  end
```

myriamboure:
Testing notes: https://docs.google.com/document/d/1cKZXLyq3ADGB5nylH01M7uz1IQu2c1Sz8RIOPmBgFG4/edit# this is working in both www and without.
We knew it already as UK already deployed in production and didn't reported issue.
Ready to go @sauloperez .

PR #2317, opened 2018-05-22
Added missing translation key for no results in admin orders page

### LINKED ISSUE #2308: Translations missing in "orders" page
## Description
When filtering orders the "no order founds" result isn't transated
namhpcfbljehhfgp

luisramos0:
#### What? Why?

Closes #2308

The key no_orders_found used in orders admin page was missing from the translations file.

#### What should we test?

As far as I understand, this key will be picked up by transifex and we will be able to translate it to other languages. I have manually added this key and a translation to other language files and the text is translated correctly.

#### Release notes

Fixed missing translation in admin orders page.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2243,6 +2243,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   order_cycles_no_permission_to_coordinate_error: "None of your enterprises have permission to coordinate an order cycle"
   order_cycles_no_permission_to_create_error: "You don't have permission to create an order cycle coordinated by that enterprise"
   back_to_orders_list: "Back to order list"
+  no_orders_found: "No Orders Found"
   order_information: "Order Information"
   date_completed: "Date Completed"
   amount: "Amount"


mkllnk:
I'm just wondering how we prevent this translation from disappearing in the future. Since the translation key `no_orders_found` is not in our code base (only in Spree), somebody may think that it's not used any more.

I guess, long term we want to replace Spree's view and then we'll have that key in our code base. Short term, we could organise `en.yml` in a way that we have a section for all the "foreign" translation keys.

sauloperez:
I'm not entirely sure about that section. I guess it depends on how often we commit this mistake.

luisramos0:
Thanks guys.
So, I assume at some point someone will merge this into master (I assume I'll be able to see that here).
Then a build will run and trigger some interactions with transifex or actually transifex will download the new en.yml file (is there a way I can see that happening? is there a read only access to the build system?) and then this new entry will show up on transifex to be translated. Correct?
Thank you! Luís

sigmundpetersen:
Hey @luisramos0 there is some description of the process here https://github.com/openfoodfoundation/openfoodnetwork/wiki/i18n

luisramos0:
Hi Sigmund, thanks, yes, I read that.
Is there anything I can do on this one or should I just wait for someone with privs to follow the rest of the process?

sauloperez:
Yep, I just moved it to "Test Ready" for some tester to pick it up.

daniellemoorhead:
@mkllnk reckon you could stage this for @myriamboure to test?

myriamboure:
In Aus staging thetext still appear correctly in English:
screenshot from 2018-05-28 16-34-48
We will have to test when we deploy that transifex handle well that new key.
Sorry no testing notes as was just so quick... :smiley_cat:

PR #2318, opened 2018-05-23
Fixed broken translation keys in "new order cycle" screen

### LINKED ISSUE #2307: Translations missing in order cycle page
screenshot from 2018-05-22 05-58-38

## Steps to Reproduce
1. Click on order cycle, new order cycle, select a coordinator and see

## Context
Reported by a French user.

## Severity
S3

## Your Environment
* Version used: master branch betwee v1.15 and v1.15.1 on French staging.

luisramos0:
#### What? Why?

Closes #2307 

"Schedules" and "Orders close" translation keys were not being picked up from translations file. 

"Schedules" - changed the lookup path to pick up the right translation key (already available on translation files)
"Orders close" - renamed key on translation file. I decided to rename instead of just adding "orders_close" as I didn't find any occurrence of the key "order_closes" in both ofn and spree codebase.

#### What should we test?

The translations of the labels of the filters on the "New Order Cycle" screen:
- "Orders close" will have to go through transifex
- "Schedules" is not a new key so it will be immediately fixed.

### DIFF
--- a/app/views/admin/order_cycles/_name_and_timing_form.html.haml
+++ b/app/views/admin/order_cycles/_name_and_timing_form.html.haml
@@ -31,7 +31,7 @@
 - if subscriptions_enabled?
   .row
     .two.columns.alpha
-      = f.label :schedule_ids, t('admin.order_cycles.schedules')
+      = f.label :schedule_ids, t('admin.order_cycles.index.schedules')
     .twelve.columns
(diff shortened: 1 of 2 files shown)


myriamboure:
Just a quick screenshot to show that keys are not changed in the English Aus version, so ready to go!
screenshot from 2018-05-29 09-10-30

PR #2330, opened 2018-05-24
Detect back button (alternative fix for #1213)

stveep:
#### What? Why?
Alternative fix for #1213. 

I came up with an alternative approach in case the other hack preventing use of the cache site-wide does result in poor performance. This has three steps: 1) catch the back button press 2) suppress updating of the cart and 3) refresh to get the cart data from the server again. This last full refresh step is a bit ugly, maybe there's a better way? 

Something like this is necessary because of the way the currentOrder data is injected into the page - it's hard to update that dynamically. Stopping the orders#populate method prevents the new items being deleted on the server when the old (cached?) cart is re-synced, but the old currentOrder data is still in the page when back is pressed so the cart can't just be updated with that, hence the reload is needed. If we rewrite the service to use an ajax request (or ngResource) instead, we could get around it.

#### What should we test?

Add items to cart, click edit cart. Pressing back forces refresh of the page [which has the wrong (empty) cart at first] and restores the correct cart contents.

The extra refresh step should not happen on presses of the back button in other contexts (it only happens when the update method gets called in the cart service).

### DIFF
--- a/app/assets/javascripts/darkswarm/services/cart.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/cart.js.coffee
@@ -40,6 +40,11 @@ Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $roo
 
     update: =>
       @update_running = true
+      # This suppresses updating the cart if the back button was pressed.
+      # Items don't get deleted, but it still needs to be refreshed from the server
+      # as currentOrder in the page will be out of date
+      if (window.performance && window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD)
+        return # Need to refresh data from the server here, just reloading page doesn't work as this then triggers populate method below.
 
       $http.post('/orders/populate', @data()).success (data, status)=>
         @saved()
--- a/app/assets/javascripts/darkswarm/services/cart.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/cart.js.coffee
@@ -4,6 +4,7 @@ Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $roo
     dirty: false
     update_running: false
     update_enqueued: false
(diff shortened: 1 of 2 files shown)


Matt-Yorkley:
And does this CurrentOrder class need to exist? https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/darkswarm/services/current_order.js.coffee

oeoeaio:
> we add a quick route with the same output as injection_helper#inject_current_order

Bingo, if that works it would be my preference. The only issue that I forsee is that some things will be expecting `CurrentOrder.order` to exist as soon as the page loads, so we may need to deal with synchronicity....

> And does this CurrentOrder class need to exist? 

Yep, but I propose that we rewrite it, remove the method to inject `currentOrder` completely and instead load it via AJAX on page load. Something like (untested):

```coffee
Darkswarm.factory 'CurrentOrder', ($http) ->
  new class CurrentOrder
    order: {}

    constructor: ->
      $http.get '/current_order.json', (data) ->
        angular.extend(@order, data)
```

stveep:
Yes I was hesitating about going down this route because I don't know if other stuff depends on having currentOrder in the page. Also if we switch everything to an API endpoint like this we will need to know either the order ID or also pass user/session info - what's the best way to get access to that in Angular? 

Another halfway house fix would be to use the injected order ID for an AJAX request after page load to check whether anything has changed. That would allow us to keep the injected details in the page for now in case anything else depends on them. If not, we can modify to inject just the current_order ID and move to an API request for the contents of the cart.

mkllnk:
Okay @oeoeaio and @sauloperez, what do we do with this pull request now? Do you think it's an improvement we should merge or would you like to dismiss it and wait for the next solution?

oeoeaio:
I don't have an alternative that is this quick and easy, but it also feels a little on the hacky side and it sounds like the UX isn't great. I am happy to go with it if we think the issue is high enough priority that we need to just solve it. I don't think we should dedicate much more time or energy to this now.

PR #2341, opened 2018-05-28
fixed country_state selectors on checkout

### LINKED ISSUE #2105: Can't place an order if my address is not from main instance country
## Description
If a customer living in Spain for instance want to order on OFFrance to a hub and try to put a Spanish address, he can't place his order because states don't update when he switches country (and can't place order, just not going through, not possible to make the states list update...)

## Expected Behavior
If I switch country the corresponding list of states is proposed dynamically.

## Actual Behavior
If I switch country the list of states proposed doesn't change, still the ones from the previous country.

## Steps to Reproduce
1. Make sure you use an instance that has states for more than one country.
2. Try to place an order, at checkout change country, you will see that states don't update.

## Animated Gif/Screenshot
https://www.useloom.com/share/f4209daa2baa43dd9ea0a1edc3f39f6b

## Context
I hit that issue while testing bug fix in customer edit address PR2074

## Severity
S3, fundamental feature (placing order) and no workaround, but it's not affecting any users (hopefully) for now...

## Your Environment
- Version used: Aus staging with branch PR2074

## Possible Fix
Update state dynamically.
If state was selected before country is changes, empty dynamically while switching states list.

luisramos0:
#### What? Why?

Closes #2105 

Functional:
- In the checkout screen, the list of states was not being updated when the country was changed.

Technical Solution:
- I injected country/state data into the browser and created simple country controller to handle the behaviour (inspired by what's done on the enterprise registration process code).

Technical question for review:
- is the div with the controller on the right place on the html or is there a better place?
          %div{ "ng-controller" => "CountryCtrl" }

Issue:
- I run into an issue when doing checkout in the US. Checkout errors out with an message related to adjustments:
- RuntimeError (can't modify frozen Hash):
          app/models/spree/order_decorator.rb:413:in `update_adjustment!'
- git blame around this line of code points to Maikel, Rob and Enrico. Do you guys understand what this error is?
- Do you think we need to fix this now? It only happens in US on my local instance.
Thank you!

#### What should we test?

Checkout process and country state selection on both shipping and billing addresses.
I have tested this quite a bit now. I think it's working correctly. I can checkout in both Portugal and Spain. 

#### Release notes

Fixed country and state selectors on checkout. User can checkout with addresses in multiple countries on the same OFN instance.

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/checkout/country_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/checkout/country_controller.js.coffee
@@ -1,0 +1,8 @@
+Darkswarm.controller "CountryCtrl", ($scope, availableCountries) ->
+
+  $scope.countries = availableCountries
+
+  $scope.countriesById = $scope.countries.reduce (obj, country) ->
+    obj[country.id] = country
+    obj
+  , {}
--- a/app/helpers/checkout_helper.rb
+++ b/app/helpers/checkout_helper.rb
@@ -66,16 +66,6 @@ module CheckoutHelper
     Spree::Money.new order.total - order.total_tax, currency: order.currency
   end
 
-  def checkout_state_options(source_address)
-    if source_address == :billing
-      address = @order.billing_address
-    elsif source_address == :shipping
-      address = @order.shipping_address
-    end
-
-    [[]] + address.country.states.map { |c| [c.name, c.id] }
-  end
-
   def validated_input(name, path, args = {})
     attributes = {
       required: true,
--- a/app/helpers/checkout_helper.rb
+++ b/app/helpers/checkout_helper.rb
@@ -76,10 +76,6 @@ module CheckoutHelper
     [[]] + address.country.states.map { |c| [c.name, c.id] }
   end
 
-  def checkout_country_options
-    available_countries.map { |c| [c.name, c.id] }
-  end
-
   def validated_input(name, path, args = {})
     attributes = {
       required: true,
--- a/app/views/checkout/_billing.html.haml
+++ b/app/views/checkout/_billing.html.haml
@@ -33,7 +33,7 @@
 
             .small-6.columns
               %label{ for: 'order.bill_address.state_id' } {{ "state" | t }}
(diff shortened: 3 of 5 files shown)


sauloperez:
Looks good to me! regarding that error, can you reproduce it with the default data from the `openfoodnetwork:dev:load_sample_data` rake task? Can you share a gif to see what happens?

sauloperez:
ok @luisramos0 no need to worry then.

mkllnk:
@luisramos0 In this case, yes, one commit would have been enough. With bigger pull requests, multiple commits are good, but each commit should make sense on its own and lead to the solution without detours.

RachL:
Hi @sauloperez do you have room to stage this one today ? :) Thanks

mkllnk:

RachL:
@myriamboure I will have little internet access today and tomorrow so I cannot guaranty to test it today sorry :(

myriamboure:
@mkllnk @oeoeaio I try to test but I can only see "Australia" in the proposed list of country when I checkout. I added manually some states to the "France" country as I think by default only countries that have states set up are proposed for selection, but it doesn't work either. Any idea what I should do?
screenshot from 2018-06-18 10-39-59
I think there is something I am not doing properly here...

myriamboure:
I tried to add another shipping method with the EU_VAT tax as thought that might have a impact, but it doesn't see it does... no change.

myriamboure:
Ok, found a way to test by adding "france" country to the Australia VAT zone... so here are my testing notes: https://docs.google.com/document/d/1KwcoHu45gr6BbYEUiLM8BbLlR2WTOe2hSLFM0NqF7ag/edit# and it is working all good!
UX is pretty good.
Tried with both billing and shipping address and checked that the addresses were well recorded as well for registered users. Seems all good to me!
As you said @luisramos0 the VAT zone/fee concerns the shipping method, not really the billing/shipping address but this is out of scope here and will come in the big question of VAT overhaul at some point. We will need to think how we make sure that people address match the location conditions for shipping method chosen... especially when going towards a single instance strategy. But again out of scope.
So this is ready to go for me!

luisramos0:
hurray! :-)

PR #2343, opened 2018-05-29
2212 remove report default filters

### LINKED ISSUE #2212: Reports: Empty filters in reports pages
## Description

**- On page:** /admin/reports when clicking on any link from that page
screenshot from 2018-04-12 19-35-30

**- I want to be able to do:** access the page (no data loaded already done) with no pre-defined filters options (= empty filters)

## Acceptance criteria
1- When the user click on any of the links from the /admin/reports, the page as described in the precedent issue #2211 opens, but now **without** any filter.
untitled presentation 1

2- Once the user applies filters and click on "search", the page reloads, but does not open a new window (it is already the case).

mkllnk:
#### What? Why?

Closes #2212.

When you load up a report, it fills in some default dates. Some people forget to change the date to what they need and are then confused when they get the wrong report data. This pull request removes the defaults so that people have to always set dates.

#### What should we test?

All report filters are empty (I think the select boxes can have defaults though. Nobody was confused about them.)

#### Release notes

Reports don't pre-fill date filters any more since the "wrong" defaults are more confusing then no defaults.

If newer versions of Spree add more reports, they may have defaults. But as far as I know we only display our own reports.


spec/lib/open_food_network/xero_invoices_report_spec.rb
```diff
@@ -98,15 +99,15 @@ module OpenFoodNetwork
...
       describe "when an initial invoice number is given" do
-        subject { XeroInvoicesReport.new user, {initial_invoice_number: '123'} }
+        subject { XeroInvoicesReport.new user, initial_invoice_number: '123' }
 
         it "increments the number by the index" do
-          subject.send(:invoice_number_for, order, 456).should == 579
+          expect(subject.send(:invoice_number_for, order, 456)).to eq(579)
         end
       end
     end
```
oeoeaio: So nothing in this spec actually changed right? You've just updated the syntax?

mkllnk: Right. I wanted to run the spec on its own and had to modify it. And since I was at it, I just gave it the automated groom. I didn't have to change any logic though.

myriamboure:
Testing notes: https://docs.google.com/document/d/1CCSmngTPT5i-dBpA3fUk1UigjAQye15Ti51w5b2hfxw/edit?usp=sharing

Ready to go.
There is one thing though @mkllnk that was described in the mockups in both #2212 and #2211 and was not done: it's displaying the message "select filters and click on search in order to access your data". We added that when working on UX with Rachel because as a user, when you arrive on order report page for instance and see "no data" it can be kind of frightening "What!? But I had orderes, where are theyyyy!!! And panicking - remember we have lots of not-at-all tech people. I propose we merge that and I open now a new PR for that attached to the same epic. If you think it doesn't worth to add it we can discuss on the issue.

daniellemoorhead:
>There is one thing though @mkllnk that was described in the mockups in both #2212 and #2211 and was not done: it's displaying the message "select filters and click on search in order to access your data". We added that when working on UX with Rachel because as a user, when you arrive on order report page for instance and see "no data" it can be kind of frightening "What!? But I had orderes, where are theyyyy!!! And panicking - remember we have lots of not-at-all tech people. I propose we merge that and I open now a new PR for that attached to the same epic. If you think it doesn't worth to add it we can discuss on the issue.

@mkllnk @myriamboure I didn't see this discussed on the issue and resolved. Did it get added before the PR was merged, or does it need to be included in another PR or written up as a good first issue?

myriamboure:
See issue #2356 and PR #2361 @daniellemoorhead so all good :+1:

PR #2357, opened 2018-06-05
changed the country and state selectors to angular on enterprises admin

### LINKED ISSUE #1538: States don't update dynamically when you change country
## Description
If I create an entreprise in the back office / modify a profile and want to use an address which is not in the main instance country, the list of states doesn't update. 

## Expected Behavior
When I change the country, the states list update automatically to those corresponding to that country.

## Actual Behavior
When I change the country, the states list available remains the one of the previous country selected.

## Steps to Reproduce
1. Make sure you use an instance where you have localization files for at least two countries, with associated states.
2. Create an entreprise, you will see by default the country and state proposed are the one of country 1 (the country by default of the instance I guess)
3. Try to change the country to country 2, you will see the state doesn't uptade.
4. Save: you get the error message but now the states have been updated and you can select the right state.
5. Save again, now that works, no error message.

## Animated Gif/Screenshot
https://www.useloom.com/share/c036bde11ccc4e59ba47b2e5f53c805c

screenshot from 2017-03-30 17-41-15

If I save, I get an error message:
screenshot from 2017-04-19 09-49-45
But then I realize the states list have been updated and I can set up the right state.

## Context
In France we have a user running a hub based in Spain with Spanish producers, he hit this issue each time he creates a new entreprise (new producer for instance)

## Severity
S4

## Your Environment
* Version used: 1.10.1

## Possible Fix
RAS

luisramos0:
...(new and edit screens) so that states are dynamically updated when country is changed

This PR is ready for review now.

#### What? Why?

Closes #1538 

Changed the country and state selectors on new and edit enterprises to angular so that states becomes dynamic based on country selection.

#### What should we test?

New Enterprise and Edit Enterprise Address should now allow for selection of states of the selected country.

#### Release notes

Create Enterprise and Edit Enterprise Address now allow for selection of states of the selected country.

### DIFF
--- a/app/assets/javascripts/admin/enterprises/controllers/country_controller.js.coffee
+++ b/app/assets/javascripts/admin/enterprises/controllers/country_controller.js.coffee
@@ -1,5 +1,5 @@
+# Used in enterprise new and edit forms to reset the state when the country is changed
 angular.module("admin.enterprises").controller 'countryCtrl', ($scope, availableCountries) ->
-
   $scope.countries = availableCountries
 
   $scope.countriesById = $scope.countries.reduce (obj, country) ->
@@ -7,19 +7,12 @@ angular.module("admin.enterprises").controller 'countryCtrl', ($scope, available
     obj
   , {}
 
-  $scope.countryOnChange = (stateSelectElemId) ->
+  $scope.$watch 'Enterprise.address.country_id', (newID, oldID) ->
     $scope.clearState() unless $scope.addressStateMatchesCountry()
-    $scope.refreshStateSelector(stateSelectElemId)
 
   $scope.clearState = ->
-    $scope.enterprise_address_attributes_state = {}
+    $scope.Enterprise.address.state_id = null
 
   $scope.addressStateMatchesCountry = ->
-    $scope.countriesById[$scope.enterprise_address_attributes_country.id].states.some (state) -> state.id == $scope.enterprise_address_attributes_state?.id
-
-  $scope.refreshStateSelector = (stateSelectElemId) ->
-  	# workaround select2 (using jQuery and setTimeout) to force a refresh of the selected value
-    setTimeout ->
-      selectedState = jQuery('#' + stateSelectElemId)
-      jQuery('#' + stateSelectElemId).select2("val", selectedState)
-    , 500
+    $scope.countriesById[$scope.Enterprise.address.country_id].states.some (state) ->
+      state.id == $scope.Enterprise.address.state_id
--- a/app/assets/javascripts/admin/enterprises/controllers/country_controller.js.coffee
+++ b/app/assets/javascripts/admin/enterprises/controllers/country_controller.js.coffee
@@ -1,0 +1,25 @@
+angular.module("admin.enterprises").controller 'countryCtrl', ($scope, availableCountries) ->
+
+  $scope.countries = availableCountries
+
+  $scope.countriesById = $scope.countries.reduce (obj, country) ->
+    obj[country.id] = country
+    obj
+  , {}
+
+  $scope.countryOnChange = (stateSelectElemId) ->
+    $scope.clearState() unless $scope.addressStateMatchesCountry()
+    $scope.refreshStateSelector(stateSelectElemId)
+
+  $scope.clearState = ->
+    $scope.enterprise_address_attributes_state = {}
+
+  $scope.addressStateMatchesCountry = ->
+    $scope.countriesById[$scope.enterprise_address_attributes_country.id].states.some (state) -> state.id == $scope.enterprise_address_attributes_state?.id
+
+  $scope.refreshStateSelector = (stateSelectElemId) ->
+  	# workaround select2 (using jQuery and setTimeout) to force a refresh of the selected value
+    setTimeout ->
+      selectedState = jQuery('#' + stateSelectElemId)
+      jQuery('#' + stateSelectElemId).select2("val", selectedState)
+    , 500
--- a/spec/features/admin/enterprises_spec.rb
+++ b/spec/features/admin/enterprises_spec.rb
@@ -53,7 +53,7 @@ feature %q{
     fill_in 'enterprise_address_attributes_address1', :with => '35 Ballantyne St'
     fill_in 'enterprise_address_attributes_city', :with => 'Thornbury'
     fill_in 'enterprise_address_attributes_zipcode', :with => '3072'
-    select2_search 'Australia', :from => 'Country'
+    # default country (Australia in this test) should be selected by default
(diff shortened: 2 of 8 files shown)


oeoeaio:
Hey @luisramos0 ping me when you're ready for a review and I'll remove the `pr-wip` label.

oeoeaio:
Sorry just saw your comment on #1538.

> How/where do I change the model of Enterprise to include the address country and state as needed?

Are you still having trouble with this? If you have a look at the `admin_inject_enterprise` method in `admin/injection_helper.rb` you will see that it serializes the enterprise in question using `Api::Admin::EnterpriseSerializer`, which is defined in `app/serializers/api/admin/enterprise_serializer.rb`. You should just be able to add the country and state ids to the list of serialized attributes by adding a new line: 
```ruby
attributes :country_id, :state_id
```

As an aside, I am a bit concerned about the presence of the `cleanup_address_attributes` method, is the plan to keep this, or is it just a temporary helper method for you to use while developing this fix?

luisramos0:
Hey @oeoeaio , thanks a lot for your help! That's exactly what I needed. I missed the Admin version of the serializer and I couldn't understand the other one :-)
I added attributes :address to it. It works.

I only added that cleanup_address_attributes because I couldnt fix it with angular. I just spent a bit more time with it and managed to understand and fix it. As far as I understand:
- this form is not an angular form so the data sent to the server is the data on the DOM, not the model
- ng-options puts typed values on dom like nubmer:175
- for some reason I couldnt make "track by id" work the other day, today all is good, it works, I removed the server side workaround

luisramos0:
This PR is ready for review now.

I have two comments/questions:
- logs - do we have some kind of code conventions and specifically logging approach on the server side? I put some log messages as example, maybe some data could be printed as well? (log level "error" is wrong in this case, Ill fix that with any other review feedback). Or do we have a no logs convention for some reason?
- the second commit related to clearing state select when country changes is quite bad (jquery), this was the only way I managed to make it work... including the timeout, without which, it doesn't work. for some reason the ng-model is not reflected on the state select2. If we don't find another alternative, maybe I can move this jquery magic to a specific directive (to make the problem explicit for the future)?

Thanks!
Luís

sauloperez:
regarding the logs, we never had a particular discussion about the topic. On my experience these logs add little to no value. They are hardly ever grepped from the logs, they make log files grow which makes it harder to process when they become big and with [Bugnsag](https://www.bugsnag.com/) we have a better mean to track errors. What's the purpose you want them to serve @luisramos0 ?

app/assets/javascripts/admin/enterprises/controllers/country_controller.js.coffee
```diff
@@ -0,0 +1,24 @@
+angular.module("admin.enterprises").controller 'countryCtrl', ($scope, availableCountries) ->
+
+  $scope.countries = availableCountries
+
+  $scope.countriesById = $scope.countries.reduce (obj, country) ->
+    obj[country.id] = country
+    obj
+  , {}
+
+  $scope.countryOnChange = (stateSelectElemId) ->
```
sauloperez: I would add a comment explaining why we had to go with `setTimeout`. It'll be valuable in the future.

sauloperez:
It's not the first time select2 gets on our way. Isn't it time to replace it with something that fits better our needs? something more angular-friendly?

luisramos0:
Hi @sauloperez , thanks for the review, I'll make the change.

re logs, I think I say this because I have been working with larger (micro)SOA systems with thousands of simultaneous users where logging app activity to a log central is a fundamental part of service monitoring (for example ELK). In this project, I think it would be good to start with some basic logging, for example, "Error updating enterprise." (and include some more info about the user for example)  would give us info about how many, if any users, are having trouble saving enterprises (and we could already get some info about the possible reason for that). I am not sure what exactly you get with bugnsag...

re grepping, with a good logging framework you can organize your logs by user sessions for example and after a little logging is in place you start to have very clear picture of what is going on. And that's very useful when you get users reporting errors live that you can't replicate.

I don't think this is a major problem for this project right now, I just added the log messages to be able to debug and then decided to leave them there and ask this question here. I am happy to remove them and leave this discussion for later.

Thanks,
Luís

luisramos0:
ok @sauloperez , thanks for explaining.
I have removed logs and added a comment about select2, jquery and the timeout. Anything else?
Thanks.

spec/features/admin/enterprises_spec.rb
```diff
@@ -342,6 +342,8 @@
         fill_in 'enterprise_address_attributes_address1', with: 'z'
         fill_in 'enterprise_address_attributes_city', with: 'z'
         fill_in 'enterprise_address_attributes_zipcode', with: 'z'
+        select2_select 'Australia', from: 'enterprise_address_attributes_country_id'
```
mkllnk: Does that mean people always have to select Australia even when it's the only country in the database? It would be nice to use `Spree::Config[:default_country_id]` as a default like before.

luisramos0: good point. thanks. done.

oeoeaio:
> It's not the first time select2 gets on our way. Isn't it time to replace it with something that fits better our needs? something more angular-friendly?

We have directive called `ofn-select2` which should be used for all select boxes in the admin section. It is actually been quite stable, and does lots of nice things like sanitising options text. We are a bit stuck with select2 until we completely rewrite our backend views, because select2 comes with Spree.

@luisramos0 I have created a PR against your branch. I changed the code so that it uses the `ofn-select2` directive, which improves the readability of the select elements IMHO, and means that we can do away with the whole refreshStateSelector method, including the timeout which solves an issue that has been dealt with in the `ofn-select2` directive already.

I also refactored a couple of other things:
1. Used the existing ng model `Enterprise.address` for the select2 elements, rather than using the duplicate `enterprise_address_attributes*` variables.
2. Improved the serialisation of the address using the Active Model Serializer `has_one` method. This prevents the nesting of the `address` inside `Enterprise.address`.
3. Removed the `ng-app => 'admin.users'` declaration, which I think was actually illegal because my understanding is that only one ng-app should be declared per page?
4. Created a very basic `NewEnterpriseController` which at the moment is only responsible for instantiating `$scope.Enterprise` for the new enterprise form.
5. Injected the value of the default country id using our standard injection conventions, rather than relying on `ng-init`, which is not really intended for this purpose, although it works fine.

luisramos0:
merged and tested. looks good! thanks @oeoeaio !

mkllnk:

myriamboure:
Looking perfect!
https://docs.google.com/document/d/16nQTC8JnEd6-ivVTCTLOo7i4pAWx4P73w4KFvDA32wE/edit#
Ready to go :-)

PR #2360, opened 2018-06-06
Rewrite user stat query for improved performance of homepage

oeoeaio:
#### What? Why?

We just set up Skylight and it immediately became clear that some of the lowest hanging fruit for performance improvement is the count of active shoppers on the homepage. This PR rewrites the query so that it no longer uses an unnecessary join, which should mean a ~6-8x speed improvement (according to my local testing). If we're lucky this may bring our homepage load time down to <100ms for Aus production.

#### What should we test?

- [ ] The `FOOD SHOPPERS` count on the homepage should continue to work, and should show the same number as it did prior to this change being made

#### Release notes

Made a small technical change to stat calculation to improve the load time of the homepage

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -7,7 +7,7 @@ class HomeController < BaseController
     if ContentConfig.home_show_stats
       @num_distributors = Enterprise.is_distributor.activated.visible.count
       @num_producers = Enterprise.is_primary_producer.activated.visible.count
-      @num_users = Spree::User.joins(:orders).merge(Spree::Order.complete).count('DISTINCT spree_users.*')
+      @num_users = Spree::Order.complete.count('DISTINCT user_id')
       @num_orders = Spree::Order.complete.count
     end
   end


oeoeaio:
> Don't take the results on your localhost too seriously. There are many factors at play.

Haha, don't worry, I haven't got my hopes up too high.

> We have also Skylight for staging to try these things out.

Would be great to see the impact there to determine whether this is worth putting through. 👍

RachL:
My apologies for taking so long to test, but I had some problems to open a shop and then to fulfill an order properly. So note to myself, if that happens I shall remember :
- to check the delivery method 
- to check the paiment method (test both staging and production)
- to remember that if the payment method is set to Spain, I cannot order from another country

That being said, here is my testing notes : https://docs.google.com/document/d/1YvPKaHszO4H9A6_qgvXuE6UM-vvxL-h3cx-TTriCqAc/edit#

The issue is ready to go.

oeoeaio:
Not bad @sauloperez @mkllnk:

screen shot 2018-06-27 at 3 20 29 pm

https://oss.skylight.io/app/applications/cy1vxPpf0oaO/1529992260/4h11m/endpoints

Issue #2210, opened 2018-04-12
Enterprise managers can access the page of the report they want to generate


myriamboure:
## What is the problem we are solving 
Some enterprise users, when clicking on “reports” and then clicking on a specific report they would like to access, don’t even manage to reach the given report page as they experience a timeout (too many data loaded).
This is because all data are loaded with some default filtering options (like start date = first day of the month) but for some users even with those pre-defined filters that makes too many data so they experience timeout.

## Success factors = expected outcome 
Enterprise users when clicking on "reports" then clicking on a given report can access the page of this given report.  

## Metric
People who are on a broadband network can click on any report from "report" menu and access the given report page without experiencing timeout.

## Useful information for inception
**Actual behavior:** Enterprise users when clicking on "reports" then clicking on a given report are redirected to report page but by default there is some pre-defined filters and the system automatically tries to load the corresponding data. As a consequence, some users who have many data even with those default filtering options experience timeout and so the page doesn't even open.

## Link to the "Icebox" item in Discourse
https://community.openfoodnetwork.org/t/make-sure-all-enterprise-managers-can-access-reports-filtering-menu/1282

## First feature candidate to be implemented
"remove default filtering and don't load the report before filtering options are set"
WARNING: this will not solved the performance issue experienced by users who try to generate a report with one year data for instance. Should only solve first bit: they manage to go on the given report page.

myriamboure:
@mkllnk this epic is for you :-) There are two stories, and we suggested that @RachL has a quick UX work on it (should be little things, but  like for instance display a message inviting the user to use filters, and telling that if long period might take more time to load... things like that) before you take the stories and implement them. @RachL can you plan that in the coming days?

myriamboure:
Ok so it seems that epic is now done as al the three issues in it have been done! @daniellemoorhead in our processes are we supposed now to have the last master staged to retest the whole feature and review if the original problem have been solved before we can consider the feature as done? Or as long as all the issues in the epic are done we just consider it as done?

daniellemoorhead:
>in our processes are we supposed now to have the last master staged to retest the whole feature and review if the original problem have been solved before we can consider the feature as done? Or as long as all the issues in the epic are done we just consider it as done?

@myriamboure have we achieved this?

![screen shot 2018-06-22 at 1 40 39 pm](https://user-images.githubusercontent.com/10949052/41756835-e87b2b9e-7621-11e8-9f2e-f4dac8bb8b31.png)

[...]

PR #2361, opened 2018-06-06
2356 report search hint

### LINKED ISSUE #2356: Add guiding message in "report" section to prevent users panicking when they see no data.
## Description
Now that we have removed default loading of data and default filtering, users can be disturbed (panicking even) when the first thing they see is "no data". When working on UX with Rachel we thought it would be good to add a little reassuring message:
image2

## Expected Behavior
The reassuring and guiding message is displayed.

## Actual Behavior
It is not.

## Steps to Reproduce
1. Click on reports and see

mkllnk:
#### What? Why?

Closes #2356.

We changed the reports logic to not load any data before searching (#2210). But now some enterprise customers could panic if they look at reports and see that there are none, for example no sales. :scream: 

So we show a friendly message to explain what to do.

#### What should we test?

Clicking on any report should show a friendly message to click on search. Only the Sales Total report has not been modified since it comes from Spree (and is probably not used by enterprise users?).

#### Release notes

Nothing to mention here. The release notes of other pull requests related to reports should be enough.

Not.

### DIFF
--- a/app/views/spree/admin/reports/_table.html.haml
+++ b/app/views/spree/admin/reports/_table.html.haml
@@ -1,21 +1,25 @@
 - column_partials ||= {}
-%br
-%br
-%table{id: id}
-  %thead
-    %tr
-      - @header.each do |heading|
-        %th= heading
-  %tbody
-    - @table.each do |row|
+- if render_content?
+  %br
+  %br
+  %table{id: id}
+    %thead
       %tr
-        - row.each_with_index do |cell_value, column_index|
-          %td
-            - partial = column_partials[column_index]
-            - if partial
-              = render partial, value: cell_value
-            - else
-              = cell_value
-    - if @table.empty?
-      %tr
-        %td{colspan: @header.count}= t(:none)
+        - @header.each do |heading|
+          %th= heading
+    %tbody
+      - @table.each do |row|
+        %tr
+          - row.each_with_index do |cell_value, column_index|
+            %td
+              - partial = column_partials[column_index]
+              - if partial
+                = render partial, value: cell_value
+              - else
+                = cell_value
+      - if @table.empty?
+        %tr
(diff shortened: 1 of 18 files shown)


mkllnk:
These changes could have been in pr https://github.com/openfoodfoundation/openfoodnetwork/pull/2343 which has release notes. I don't think this change makes much sense without the others. But if you really want something for the release notes here:

> Reports display the instruction to click the search button if the user hasn't done it yet.

app/views/spree/admin/reports/_table.html.haml
```diff
@@ -0,0 +1,23 @@
...
+              - if partial
+                = render partial, value: cell_value
+              - else
+                = cell_value
+      - if @table.empty?
+        %tr
+          %td{colspan: @header.count}= t(:none)
+- else
+  %p.report__message
+    = t(".select_and_search")
```
oeoeaio: Do we have a strategy for implementing BEM? Should we just be updating things as we touch them? Should we write up a proposal about how to reimplement our existing styling so that devs know how to approach this when they have to deal with it?

mkllnk: At the moment, we don't have a naming convention. So using BEM is as good as any other naming. Paul suggested using BEM years ago and I think it's a really good idea. Should we discuss that on our next delivery train catchup? If we decide to go with it, we should document it, yes.

sauloperez: Yes, it'd be great to discuss about it but perhaps in a new dev catchup as we agreed on the delivery train catchup?

mkllnk:

RachL:
Here are my testing notes, ready to go !

https://docs.google.com/document/d/1k_KrmGM8foMA5pv6k1bFcIpZnk4tnyLV1_HEcZQbPQg/edit#

PR #2364, opened 2018-06-08
Ask devs to tell whether we should test on mobile

sauloperez:
#### What? Why?

Answering @sstead 's wishes expressed on Slack

> Could someone please put a prompt in the PR template under How to test, asking devs to indicate whether we should test on mobile. This will help me remember. I tried to find the PR template but had not luck. Thanks!

Testers need to know whether mobile devices should be considered.


sauloperez:
I take that as an approval @sstead

PR #2366, opened 2018-06-08
Refactor ship address from distributor

### LINKED ISSUE #2287: Refactor #shipping_address_from_distributor
## Description

We currently set the distributor's address as the shipping address of the order when the shipping method does not require ship address. I wasn't around when this was introduced but it looks like the order model might have a presence validation for the `ship_address` attribute.

Besides, this relies on `Order#shipping_method` which is gone in Spree 2.0. Therefore, this logic needs to be refactored to work on that Spree version.

Lots of tests fail due to `#shipping_method` not being available. Like https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-2285/builds/2:

```
 311) OrderSyncer changing the ship address when a ship address is not required does not change the ship address
       Failure/Error: if shipping_method.andand.require_ship_address == false
       
       NameError:
         undefined local variable or method `shipping_method' for #<Spree::Order:0x0055fa2c8364f0>
       # ./app/models/spree/order_decorator.rb:345:in `shipping_address_from_distributor'
       # ./app/services/order_factory.rb:33:in `create_order'
       # ./app/services/order_factory.rb:12:in `create'
       # ./app/models/proxy_order.rb:52:in `initialise_order!'
       # ./spec/services/order_syncer_spec.rb:218:in `block (3 levels) in <top (required)>'
```

sauloperez:
#### What? Why?

Closes #2287. It moves the logic of https://github.com/openfoodfoundation/openfoodnetwork/blob/f2e07419a812499f2de568a05872bde46edaf4f8/app/models/spree/order_decorator.rb#L354-L364

to the `OrderUpdater#before_save_hook` which, once https://github.com/openfoodfoundation/spree/pull/6 is merged, will get executed when the order is being completed. This way we only set this custom shipping address at the end of the checkout process. 

It comes without saying that while moving the logic to the new location it's been refactored to work with Spree's 2.0 multiple shipments. 

Reading the commits individually and in order will provide a bit more context of the changes (I hope).

#### Dependencies

https://github.com/openfoodfoundation/spree/pull/6


sauloperez:
AFAIK there is only one and in Spree but I haven't checked if that's the case 🙈

oeoeaio:
I'm a bit confused about where the hook is to actually make this work, don't we still need to wrap our `OrderUpdater` around [this constructor call](https://github.com/openfoodfoundation/spree/pull/6/files#diff-434de9976fbb9c7d9358f1cb3d401019R226) somehow?

sauloperez:
you're right... Adding WIP because I clearly need to review this a bit more.

sauloperez:
That's it! Thank's a lot for your feedback. now I'm going to try and manually test this, somehow :see_no_evil:

The changes were only done in https://github.com/openfoodfoundation/spree/pull/6

To answer your question @mkllnk the one below is the only occurrence and the one we decorate

https://github.com/openfoodfoundation/spree/blob/fe0a1311abb097bf3ac8001a24246c6cff5ecdbb/core/app/models/spree/order.rb#L218-L220

sauloperez:
[Close WIP PRs](https://trello.com/c/micYXEgs/15-close-wip-prs)

app/models/order_updater.rb
```diff
@@ -0,0 +1,54 @@
...
+      # Original logic
+      # order.payment_state = 'balance_due' if order.outstanding_balance > 0
+      # order.payment_state = 'credit_owed' if order.outstanding_balance < 0
+      # order.payment_state = 'paid' if !order.outstanding_balance?
+    end
+    order.state_changed('payment') if last_state != order.payment_state
+    order.payment_state
+  end
+
+  def before_save_hook
```
sauloperez: This is my only addition. All the rest was the previous `order_updater_decorator.rb`

sauloperez:
:warning: this is not working as it is.

In spree, there's the following transition

```ruby
after_transition :to => :complete, :do => :finalize!
```

which calls the method below

```ruby
    # Finalizes an in progress order after checkout is complete.
    # Called after transition to complete state when payments will have been processed
    def finalize!
      touch :completed_at

      # lock all adjustments (coupon promotions, etc.)
      adjustments.each { |adjustment| adjustment.update_column('state', "closed") }

      # update payment and shipment(s) states, and save
      updater.update_payment_state
      shipments.each do |shipment|
        shipment.update!(self)
        shipment.finalize!
      end

      updater.update_shipment_state
      updater.before_save_hook
      save
      updater.run_hooks

      deliver_order_confirmation_email

      self.state_changes.create({
        previous_state: 'cart',
        next_state:     'complete',
        name:           'order' ,
        user_id:        self.user_id
      }, without_protection: true)
    end
```

Adding a that transition to our checkout flow could become a pandora box, surely clashing with our own logic. I'll try to investigate this a bit more.

**TL;DR**

Due to our completely overriden `checkout_flow` the appropriate state that ends up calling `updater.before_save_hook` is never executed :see_no_evil:

mkllnk:
Would https://github.com/openfoodfoundation/openfoodnetwork/pull/2431 fix it, because it's not overriding the whole checkout flow any more?

sauloperez:
I believe so (and hope!). But I still wonder if that will execute a whole lot of new logic that wasn't until now and we might end up with Frankenstein made of our checkout controller plus Spree's logic...

sauloperez:
What I'll do now is investigate https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/checkout_controller.rb#L24-L73 and understand what's the exact logic that we're adding on top of the checkout flow, now that we almost got rid of any override there.

PR #2367, opened 2018-06-08
Move product bulk edit to product index route

### LINKED ISSUE #2297: From Product Edit if I cancel I'm taken to unfamiliar Listing Products page.
## Description
I was introducing a new user the the products page, and showed them the edit product area. We clicked cancel from the product edit page and were redirected to the 'Listing Products' page. This page is otherwise inaccessible, and it can disorientate the user. They expect that cancelling will take them to where they were before (Bulk product Edit).

## Expected Behavior
If I click cancel from Edit Product, i should be redirected to 'Bulk Product Edit' - the page I was on before going to edit the produt. 

## Actual Behavior
I'm redirected to Listing Products which is an unfamiliar view, and as far as I know, it's not intended to be used as part of OFN. It's a Spree remnant. 

## Steps to Reproduce
1. Login with an enterprise that has products
2. Click on edit a product
3. Then click cancel - you'll be redirected to Listing Products page.

## Animated Gif/Screenshot
https://www.useloom.com/share/5651af4a95814b7f8c1785795d19bade

## Context
It makes the website harder to navigate and it feels buggy to be taken to a phantom page.

## Severity
S4

## Your Environment
PC, Aus production, Chrome

## Possible Fix
When in the product edit page, if I hit cancel, i should be directed to Bulk Product Edit page.

frank-west-iii:
#### What? Why?

Closes #2297

When a user hit cancel while editing a product it took them to the spree
products index page instead of the bulk edit page. The button was part
of a shared view for all resources so changing it's actions were not
readily available.

It was suggested that instead of carrying our own separate controller
action we could just override the index action of the products
controller with the bulk edit functionality instead. This has the
advantage of removing some overrides and allows us to not add
additional overrides in the future.

#### What should we test?

Bulk edit functionality, redirects from product edit functionality, I18n of bulk edit screen, permissions related to bulk editing.


config/initializers/rack_rewrite.rb
```diff
@@ -0,0 +1,7 @@
+module Openfoodnetwork
+  class Application < Rails::Application
+    config.middleware.insert_before(Rack::Lock, Rack::Rewrite) do
+      r301   '/admin/products/bulk_edit',  '/admin/products'
```
mkllnk: Is this just a temporary measure to ease the transition? Or do we expect requests to bulk_edit in the future? Maybe we should add a TODO comment to remove this after a while.

frank-west-iii: @mkllnk, I am not sure if there are bookmarks, documentation, GH issues, or any other links tied to this that would continue to request this old route. That's usually why I set these up.

I'll let your team decide on the length of time this should be here. If you decide it should be temporary I can put a TODO in here.

mkllnk: Yes, these redirects are a really good practice we have been missing so far. Thank you for that. I do think a TODO comment with the current date would be nice so that a dev can remove it with more confidence in the future (probably years ;-)).

frank-west-iii: Perfect. I'll add that comment.

Gemfile.lock
```diff
@@ -782,8 +782,8 @@ DEPENDENCIES
   representative_view
   roadie-rails (~> 1.0.3)
   roo (~> 2.7.0)
-  rspec-rails (>= 3.5.2)
   roo-xls (~> 1.1.0)
+  rspec-rails (>= 3.5.2)
```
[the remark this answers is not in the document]
frank-west-iii: I just checked out master and bundled. I am on `bundler 16.1` and it wants gems in alphabetical order. This change was committed back in 2017, but recently merged last month. Maybe bundler was updated since then to fix an ordering issue.

myriamboure:
This code two green code reviews but tests are not passing, shall we fix something or can it go to test ready?

frank-west-iii:
@myriamboure, I pushed up an updated version that "should" have passing specs. I don't believe the specs were failing because of this change, but now that we have a green master any spec failures would be the fault of this change.

mkllnk:

myriamboure:
Testing notes:
https://docs.google.com/document/d/1oIEyDEyKl9Ex1CtQkA3AHlV6ea6p62kNcRMme0cK_yM/edit#

Issue is well fixed, I did some sanity check on bulk edit and inventories, seems all fine.

mkllnk:
I just resolved a merge conflict on one of the locales. Sorry to not have mentioned it before, but we usually don't touch any locales other the `en.yml`, because the translations are done through Transifex. Once this is merged, Transifex will download only en.yml and will not know about the changes in the other locales. It will then detect that the content of the translations is still the same and copy that into the other locales. So it would have done the work for you. But then there is a delay until someone translated the languages to 100% before these updates are then committed to a pull request on Github. That means you helped to bridge the gap until translations are available, but you also made merge conflicts more likely and your changes will be overridden by Transifex anyway.

I'm happy for this to be merged as is, but in the future you don't need to worry about locales other than `en.yml`.

frank-west-iii:
@mkllnk, awesome thanks for letting me know. I saw a lot of conversations about Transifex but didn't know anything about the process. 👍

PR #2368, opened 2018-06-08
Uk/admin orders

### LINKED ISSUE #2354: New Order in Admin Pages - Customers do not appear in customer list
Creating new orders and adding payment details via Admin results in error page.

Edit for clarity: there are two issues here;
- **Issue One:** Snail (error) page showing when adding payments on an order created via admin due to a validation error, as the order was sometimes being set to "guest checkout" whilst being placed by an admin on behalf a registered user. This issue is addressed in PR #2368.
- **Issue Two:** Customer search dropdown when adding customer details to an order placed via admin on behalf of another user is not returning results as expected, possibly due to enterprise permissions. 

## Description
When working through the new order section as enterprise administrator you receive the error page when adding payment details for a registered user. Since https://github.com/openfoodfoundation/openfoodnetwork/pull/2073

On the initial screen you are unable to search of an existing user
When creating an order you then receive the slug page when adding the payment.

## Expected Behavior
You can complete a new order in Admin without having an error message for both registered and non register customers. 
If a registered user they should come up in the search fields.
You should be able to define whether the user is a guest or not
There should be better error reporting as the slug page is not helpful in understanding the cause of the issue.

## Actual Behavior
1. Search field is unable to find any email addresses that are entered
2. You cannot set the email address you type as NOT being a guest i.e setting No
3. When adding a payment the page crashes with error message in the logs:
StateMachine::InvalidTransition (Cannot transition state via :next from :cart (Reason(s): This email address is already registered. Please log in to continue, or go back and use another email address.)):

## Steps to Reproduce
1. Search for a registered user - if they are register they should show up, if not, I should then manually enter the email address below
2. I can select whether or not they are a guest using the toggle buttons
3. fill in details and go to the next page - shipping information
4. continue through shipping page and select payments
5. on the payment page select payment method and then continue
6. proceed as normal and register the new payment and order

This process works for a non registered user

## Animated Gif/Screenshot
http://recordit.co/3Uk476JjRb
(slow loading but bare with it - the error can be seen at the end)

## Context
Some enterprises are adding orders on the behalf of customers - for instance they may take them over the phone. This is stopping them from doing that. 

## Severity
s2 - This has rendered OFN useless for one UK user. They cannot run their enterprise currently.

## Your Environment

* Version used: 1.15.1
* Operating System and version (desktop or mobile): windows desktop

Matt-Yorkley:
#### What? Why?

Fixes part of issue #2354 

Creating orders via admin was throwing errors because #2073 introduced a new validation rule to stop registered customers using the guest checkout, but the old form for manually creating orders hadn't been updated. The snail page was showing when adding a payment during admin order creation due to this validation condition not always being met.

The radio button to manually set a new order to "guest checkout" has been removed, and is now set automatically depending on whether or not the customer is registered.

#### What should we test?

Creating orders as enterprise manager via admin, with registered and unregistered users.

This PR shouldn't need additional testing on mobile.

#### Release notes

Creating orders via admin will now automatically be set to "guest checkout" or not, if the customer is registered or unregistered.

### DIFF
--- a/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
+++ b/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
@@ -1,0 +1,65 @@
+require 'spec_helper'
+
+describe Spree::Admin::Orders::CustomerDetailsController, type: :controller do
+  include AuthenticationWorkflow
+
+  describe "#update" do
+    context "adding customer details via newly created admin order" do
+      let!(:user) { create(:user) }
+      let(:address) { create(:address) }
+      let!(:distributor) { create(:distributor_enterprise) }
+      let!(:shipping_method) { create(:shipping_method) }
+      let!(:order) {
+        create(
+            :order_with_totals_and_distribution,
+            state: 'cart',
+            shipping_method: shipping_method,
+            distributor: distributor,
+            user: nil,
+            email: nil,
+            bill_address: nil,
+            ship_address: nil,
+        )
+      }
+      let(:address_params) {
+        {
+          firstname: address.firstname,
+          lastname: address.lastname,
+          address1: address.address1,
+          address2: address.address2,
+          city: address.city,
+          zipcode: address.zipcode,
+          country_id: address.country_id,
+          state_id: address.state_id,
+          phone: address.phone
+        }
+      }
+
+      before do
+        login_as_enterprise_user [order.distributor]
+      end
+
+      it "accepts registered users" do
+        spree_post :update, order: { email: user.email, bill_address_attributes: address_params, ship_address_attributes: address_params }, order_id: order.number
+
+        order.reload
+
+        expect(response).to redirect_to spree.edit_admin_order_shipment_path(order, order.shipment)
+        expect(order.email).to eq user.email
+        expect(order.user_id).to eq user.id
+        expect(order.ship_address).to_not be_nil
+      end
+
+      it "accepts unregistered users" do
+        spree_post :update, order: { email: 'unregistered@email.com', bill_address_attributes: address_params, ship_address_attributes: address_params }, order_id: order.number
+
+        order.reload
+
+        expect(response).to redirect_to spree.edit_admin_order_shipment_path(order, order.shipment)
+        expect(order.email).to eq 'unregistered@email.com'
+        expect(order.user_id).to be_nil
+        expect(order.ship_address).to_not be_nil
+      end
+    end
+  end
+end
--- a/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
+++ b/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
@@ -39,26 +39,24 @@ describe Spree::Admin::Orders::CustomerDetailsController, type: :controller do
         login_as_enterprise_user [order.distributor]
       end
 
-      it "accepts registered users" do
-        spree_post :update, order: { email: user.email, bill_address_attributes: address_params, ship_address_attributes: address_params }, order_id: order.number
+      context "when adding details of a registered user" do
+        it "redirects to shipments on success" do
+          spree_post :update, order: { email: user.email, bill_address_attributes: address_params, ship_address_attributes: address_params }, order_id: order.number
 
-        order.reload
+          order.reload
 
-        expect(response).to redirect_to spree.edit_admin_order_shipment_path(order, order.shipment)
-        expect(order.email).to eq user.email
-        expect(order.user_id).to eq user.id
-        expect(order.ship_address).to_not be_nil
+          expect(response).to redirect_to spree.edit_admin_order_shipment_path(order, order.shipment)
+        end
       end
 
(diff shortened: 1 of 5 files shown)


spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+          state_id: address.state_id,
+          phone: address.phone
+        }
+      }
+
+      before do
+        login_as_enterprise_user [order.distributor]
+      end
+
+      it "accepts registered users" do
```
sauloperez: IMHO this is to vague and doesn't convey any meaning to the reader. What should we expect `adding customer details via newly created admin order`? I would add a `context 'when adding details of a registered user'` and assert our expectations within it.

spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        spree_post :update, order: { email: user.email, bill_address_attributes: address_params, ship_address_attributes: address_params }, order_id: order.number
+
+        order.reload
+
+        expect(response).to redirect_to spree.edit_admin_order_shipment_path(order, order.shipment)
+        expect(order.email).to eq user.email
+        expect(order.user_id).to eq user.id
+        expect(order.ship_address).to_not be_nil
```
sauloperez: these 4 expectations feel redundant and IMO this leads to brittle and slow tests. Shouldn't any of the 4 suffice to know the action passed through?

spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      it "accepts unregistered users" do
+        spree_post :update, order: { email: 'unregistered@email.com', bill_address_attributes: address_params, ship_address_attributes: address_params }, order_id: order.number
+
+        order.reload
+
+        expect(response).to redirect_to spree.edit_admin_order_shipment_path(order, order.shipment)
+        expect(order.email).to eq 'unregistered@email.com'
+        expect(order.user_id).to be_nil
+        expect(order.ship_address).to_not be_nil
```
sauloperez: here as well.

myriamboure:
I think we need this one asap as it's preventing users to make an order from admin. @Matt-Yorkley can you fix quickly the thing? So we can test and hopefully merge today and it can go into the release?

Matt-Yorkley:
@myriamboure It needs a second review, but maybe you could test it today, when the Travis build is finished?

myriamboure:
@RachL or I can test tomorrow if someone stage it. Can you stage in UK @Matt-Yorkley ? I know François is doing some work on French staging but you can still ask Rachel otherwise. Or ask Pau when you have finished to test the think on the spanish server.

Matt-Yorkley:
@myriamboure Deployed to https://staging.openfoodnetwork.org.uk

myriamboure:
Testing notes: https://docs.google.com/document/d/1ZZBjxmWj6KGvdobWkWsEY-zrRM1hIQhvGVu0NrtMXqE/edit#
@Matt-Yorkley I didn't receive confirmation emails, I guess that's because they are not working on UK staging and nothing to do with your PR right?
I saw another small UX bug while testing but not related to that PR so I opened another issue #2393.

For me if you confirm email confirmation are not connected it's good to go, great UX simplification!

myriamboure:
Actually I just realised that in the dropdown menu were also shown previous guest customers apprently, checking and coming back

myriamboure:
Ok so what I realized is that in dropdown menu not only registered users but all previous customers including guest ones were shown. It is good I think, and retested and all working fine, but there is one case I couldn't test as confirmation email are not sent from uk staging: If I make an order as guest and then create an account ad make an order I guess the dropdown will show two entries for the same email address... we should make sure if there is a registered user and a previous guest one with the same address that only the registered user is shown... I guess as the info are pasted below it shouldn't have an impact as email will be recognized as registered one, but would be great to make sure ux is ok. ot blocking though, it can go as it is and we can check that later on if we think UX needs to be improved.

sauloperez:
That problem with email makes me uneasy; it is too critical so I staged this again on https://staging.katuma.org/. Can you make sure it's all good @myriamboure ?

myriamboure:
Ok I retested and it is all good, order confirmations are properly sent, etc.
https://docs.google.com/document/d/1ZZBjxmWj6KGvdobWkWsEY-zrRM1hIQhvGVu0NrtMXqE/edit#
So ready to go.

I just had one question about guest/registered records in customer list but not connected to that PR, I posted it https://github.com/openfoodfoundation/openfoodnetwork/pull/2379#issuecomment-398325516. We need to be careful in that other PR as this might impact the logic in the present fix so we'll need to test carefully.

sigmundpetersen:
@Matt-Yorkley @myriamboure it says in the intro that this "Fixes part of issue #2354". Is there still something remaining to be done in another PR or can we close the connected issue?

PR #2373, opened 2018-06-13
Reimplement import date filter as an actual Filter

### LINKED ISSUE #2372: Some products not visible because of import date filter
## Description

An Australian user noticed that some of their products were no longer visible in the bulk product edit interface. I did some investigating and found that the issue was caused by the addition of [this filter expression](https://github.com/openfoodfoundation/openfoodnetwork/commit/e362c2d8679219ad498e57369d1fd773c95b59ce#diff-acbfa509f78bd58e62ea0733f1075a32R11), which was merged with #1492.  

The expression does not seem to meet the requirements of the [AngularJS API](https://docs.angularjs.org/api/ng/filter/filter). So I suspect that is the issue.

## Expected Behavior
All managed products should be displayed in the bulk edit interface.

## Actual Behavior
Some products seem to be missing from the list for some reason.

## Steps to Reproduce
I haven't worked out exactly what the issue is, so all I have is a GIF demonstrating that the issue is fixed by removal of the import date filter.

## Animated Gif/Screenshot
Shows that no products matching 'sugar' are available at first, but then I remove the import date filter and reload the page and some 'sugar' products are found.
importdatefilter

## Context
The user in question is unable to update prices for products that are available to purchase from her shop, so this is a serious issue.

## Severity
Severity 3 (possibly 2, not sure how many users are affected)

## Your Environment
* Version used: Edge
* Browser name and version: Chrome 67.0
* Operating System and version (desktop or mobile): Mac OS, Sierra

## Possible Fix
Replace the filter with an actual filter rather than an expression, since I'm not sure how Angular evaluates expressions in this context.

oeoeaio:
#### What? Why?

Fixes #2372: Import date filter was causing some products to not be displayed on the bulk product update page. @Matt-Yorkley could you run your eyes over this and check that I haven't completely trashed something? 😬 

I suspect that the filter expression being used:
 ```
... | filter: (importDateFilter != 0) && {import_date: importDateFilter} | ...
```
may not have been working as we thought it should. So I rewrote it as a AngularJS filter service, and it seems to have fixed the issue I was seeing.

#### What should we test?

- [ ] The product import filter should work as per the original spec (see #1492)

#### Release notes

This has not been released yet, but if the #1492 is released before this bug is fixed then maybe use the following:

Fixed a bug which caused some products to be hidden from the bulk product update view in the backend.

### DIFF
--- a/app/assets/javascripts/admin/filters/import_date_filter.js.coffee
+++ b/app/assets/javascripts/admin/filters/import_date_filter.js.coffee
@@ -1,0 +1,4 @@
+angular.module("ofn.admin").filter "importDate", ($filter) ->
+  return (products, importDate) ->
+    return products if importDate == "0"
+    $filter('filter')( products, { import_date: importDate } )
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -28,7 +28,7 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     $scope.filterTaxons = [{id: "0", name: ""}].concat $scope.taxons
     $scope.producerFilter = "0"
     $scope.categoryFilter = "0"
-    $scope.importDateFilter = ""
+    $scope.importDateFilter = "0"
     $scope.products = BulkProducts.products
     $scope.filteredProducts = []
     $scope.currentFilters = []
--- a/app/views/spree/admin/products/bulk_edit/_products.html.haml
+++ b/app/views/spree/admin/products/bulk_edit/_products.html.haml
@@ -8,7 +8,7 @@
 
       = render 'spree/admin/products/bulk_edit/products_head'
 
-      %tbody{ 'ng-repeat' => 'product in filteredProducts = ( products | filter:query | producer: producerFilter | category: categoryFilter | filter: (importDateFilter != 0) && {import_date: importDateFilter} | limitTo:limit )', 'ng-class-even' => "'even'", 'ng-class-odd' => "'odd'" }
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by myriamboure** [comment]
I guess we need that to be merged before the release as well @sauloperez then to avoid create the product bug... happy to test asap when it's staged.

**Comment by Matt-Yorkley** [comment]
Looks good to me. I think we should merge it, then I can rebase #1492 PR and run the specs again.

**Comment by Matt-Yorkley** [comment]
Ah, I just noticed Travis has already answered that question with a failed spec on viewing imported products... I'll have to take a look at it with #1492 and see if it can be fixed up.

File: app/assets/javascripts/admin/filters/import_date_filter.js.coffee [hunk 1]
```diff
@@ -0,0 +1,4 @@
+angular.module("ofn.admin").filter "importDate", ($filter) ->
+  return (products, importDate) ->
+    return products if importDate == "0"
+    $filter('filter')( products, { import_date: importDate }, true )
```
Matt-Yorkley: I've been looking at why the specs fail on this, and it seems you need to remove the `true` flag from the end of the filter. The inputs being filtered in this case are date strings, which apparently get interpreted differently by angular with this setting. Removing it gets both the failing specs passing again. I also tested manually to double check and it is working as intended.

sstead:
**Testing notes**
- Looks fine. I imported a product and then the Import Date filter worked for that product. 

https://docs.google.com/document/d/1upiADU2OamQvJfAmkJ68J1j0UxRNs0JOKfwi9zfCHCk/edit?ouid=107866005736810707026&usp=docs_home&ths=true

PR #2378, opened 2018-06-14
Include admin users as managers on new enterprises

### LINKED ISSUE #2257: Error on updating enterprise users when not initialised
There are some rare cases in which a user is unable to save changes to a new enterprise after fiddling around with the users.

## Description

The `receives_notifications` field is not initialised properly on a new enterprise. I observed the default value `? string:1 ?` which looks like an Angular glitch. When trying to submit the form, a server error is generated.

## Expected Behavior

I should always be able to submit the enterprise edit form.

## Actual Behavior
```
 ActiveRecord::StatementInvalid in Admin::EnterprisesController#update

PG::Error: ERROR:  invalid input syntax for integer: "? string:1 ?"
LINE 1: ...rprise_roles" SET receives_notifications=(user_id='? string:...
                                                             ^
: UPDATE "enterprise_roles" SET receives_notifications=(user_id='? string:1 ?') WHERE "enterprise_roles"."enterprise_id" = 2

Parameters:

{"utf8"=>"✓",
 "_method"=>"put",
 "enterprise"=>{"name"=>"...",
 ...
 "receives_notifications"=>"? string:1 ?",
 ...}
```

## Steps to Reproduce

1. Log in as super admin.
2. Create an enterprise.
3. Go to the users tab of that enterprise.
4. Without submitting, add a manager to the enterprise that is not the owner.
5. Remove that manager again.
6. The update button is now clickable, click it.
7. You get a server error.

## Context

I was testing https://github.com/openfoodfoundation/openfoodnetwork/issues/2254. This is part of email confirmation.

## Severity
S3 - there is a workaround.

You can first add the owner as manager. It send appears in the notifications dropdown. Select them there as well and the form should work.

## Your Environment

Dev environment.

## Possible Fix

Two ideas:

- Initialise the mangers/notifications data properly.
- Upgrade Angular.

Matt-Yorkley:
#### What? Why?

Closes #2257 .

When a superadmin user created a new enterprise via admin and added themselves as the owner, the user was not being explicitly included in the list of managers (superadmins can usually "manage" all enterprises anyway). This meant subsequent setting of the contact user for the enterprise was not happening properly, as the EnterpriseRole record was not there, which in turn meant this was missing and causing a weird Angular glitch in the enterprise users form after creation.

#### What should we test?

Creating an enterprise via admin with a superadmin account, and a superadmin as the enterprise owner. After saving, go to the users tab on the enterprise page and the "Notifications" field should be set to the same user and not be blank. 

screenshot from 2018-06-14 13-41-42

Changing other properties on the enterprise and then saving should be successful and not show an error.

#### Release notes

Bugfix for #2257 for superadmin accounts where enterprises owned by a superadmin user where not setting manager roles and contact user correctly.

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -374,7 +374,7 @@ class Enterprise < ActiveRecord::Base
   end
 
   def ensure_owner_is_manager
-    users << owner unless users.include?(owner) || owner.admin?
+    users << owner unless users.include?(owner)
   end
 
   def enforce_ownership_limit


app/models/enterprise.rb
```diff
@@ -374,7 +374,7 @@ def geocode_address
   end
 
   def ensure_owner_is_manager
-    users << owner unless users.include?(owner) || owner.admin?
+    users << owner unless users.include?(owner)
```
sauloperez: This looks good to me but I think now it's the time improve it and not rely on the callback. So, instead of having to check every single time an action is performed on the object, which triggers the `before_validation` callback I would rather do assign the owner as user of the enterprise at the same point in time where the enteprise is created.

for `app/controllers/api/enterprises_controller.rb` I see this happens in

https://github.com/openfoodfoundation/openfoodnetwork/blob/3399eba4281a4919d5de644e467b57b46c23f647/app/controllers/api/enterprises_controller.rb#L52

but I fail to see if we do that as well on other places such as `app/controllers/enterprises_controller.rb`. All this may possibly mean extracting some sort of `CreateEnterpriseService` and I really believe it would be a great move for the future.

Would you agree on creating and checking this out on a separate issue? this way we can benefit from this fix but improve right away.

sstead:
**Testing notes**

- Superadmin users who create a profile that they own, will be made owner and notifications

https://docs.google.com/document/d/1HlVhgqN9j1C5PLaIzC-95bFbex6WuVBFNsSVm8YZ4Lk/edit#

PR #2379, opened 2018-06-14
Add guest label to admin customers index

### LINKED ISSUE #2338: Email confirmation | Enabling hub manager to identify guest customers lines
## Description
Following conversation https://github.com/openfoodfoundation/openfoodnetwork/issues/1592#issuecomment-392276084
The problem is that when a guest customer creates a verified accounts, they appear twice in the customer list of the shop. One entry is the recort corresponding to guest previous orders, the other corresponds to the new account.
To avoid confusion from hub manager we want to make clear which entry correspond to the guest one.

## Expected Behavior
Hub manager cleary see which entry is the guest customer through a label displayed on the customer line.
screenshot from 2018-05-27 18-30-42

## Actual Behavior
No identification.

## Context
Finishing email confirmation project.

## Possible Fix
Add a "guest" label in the UI. The suggestion above is just a suggestion, if you have a better idea please submit it.

Matt-Yorkley:
#### What? Why?

Closes #2338 

Minor addition to the admin customers UI to show a label on customer records that used guest accounts. The purpose was to clarify in cases where there are two records for registered and guest customers with the same email, which previously just looked like a bug or duplicate. 

screenshot from 2018-06-14 15-45-38

#### What should we test?

Customers have a guest label in the admin customers index page if they were added as a guest.
Test the UI: what happens if a guest becomes registered, how does the hub manager see that? Is it user friendly enough?

#### Release notes

A label has been added in the admin customers page to show which customers used guest checkout.

### DIFF
--- a/app/assets/stylesheets/admin/customers.css.scss
+++ b/app/assets/stylesheets/admin/customers.css.scss
@@ -1,3 +1,14 @@
 .tag-with-rules {
   color: black;
 }
+
+table#customers.index {
+
+  tr.customer {
+
+    .guest-label {
+      color: #999;
+      display: block;
+    }
+  }
+}
--- a/app/assets/stylesheets/admin/customers.css.scss
+++ b/app/assets/stylesheets/admin/customers.css.scss
@@ -7,7 +7,7 @@ table#customers.index {
   tr.customer {
 
     .guest-label {
-      color: #999;
+      color: $medium-grey;
       display: block;
       font-size: 0.85em;
       margin-top: 0.15em;
--- a/app/assets/stylesheets/admin/customers.css.scss
+++ b/app/assets/stylesheets/admin/customers.css.scss
@@ -9,6 +9,8 @@ table#customers.index {
     .guest-label {
       color: #999;
       display: block;
+      font-size: 0.85em;
+      margin-top: 0.15em;
     }
   }
 }
--- a/app/assets/stylesheets/admin/customers.css.scss
+++ b/app/assets/stylesheets/admin/customers.css.scss
@@ -1,3 +1,5 @@
+@import "variables";
+
 .tag-with-rules {
   color: black;
 }
--- a/app/assets/stylesheets/admin/account.css.scss
+++ b/app/assets/stylesheets/admin/account.css.scss
@@ -12,6 +12,6 @@ table.invoice_summary {
 
 .invoice_title {
   .balance {
-    color: #9fc820;
+    color: $spree-green;
   }
 }
--- a/app/assets/stylesheets/admin/account.css.scss
+++ b/app/assets/stylesheets/admin/account.css.scss
@@ -1,3 +1,5 @@
+@import "variables";
+
 .row.invoice_title {
   margin-bottom: 0px;
 }
--- a/app/assets/stylesheets/admin/advanced_settings.css.scss
+++ b/app/assets/stylesheets/admin/advanced_settings.css.scss
@@ -1,3 +1,5 @@
+@import "variables";
+
 #advanced_settings {
   background-color: $spree-light-blue;
   border: 1px solid $pale-blue;
--- a/app/assets/stylesheets/admin/advanced_settings.css.scss
+++ b/app/assets/stylesheets/admin/advanced_settings.css.scss
@@ -1,6 +1,6 @@
 #advanced_settings {
-  background-color: #eff5fc;
-  border: 1px solid #cee1f4;
+  background-color: $spree-light-blue;
+  border: 1px solid $pale-blue;
   margin-bottom: 20px;
 
   .row{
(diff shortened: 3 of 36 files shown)


app/assets/stylesheets/admin/customers.css.scss
```diff
@@ -1,3 +1,14 @@
 .tag-with-rules {
   color: black;
 }
+
+table#customers.index {
+
+  tr.customer {
+
+    .guest-label {
+      color: #999;
```
sauloperez: Is this color already used somewhere else? besides, it should be stored in a var for better reusability.

app/assets/stylesheets/admin/customers.css.scss
```diff
@@ -1,3 +1,14 @@
...
+      display: block;
```
sauloperez: I'd also use a smaller font size as it is extra information at a different hierarchy level as the label.

Matt-Yorkley: Done.

app/assets/stylesheets/admin/customers.css.scss
Matt-Yorkley: It looks like none of our admin CSS uses variables for colours... I could add a new colours file and start adding things, but none of it's organised like that for the back end.

myriamboure:
Ok for you @sauloperez ?

myriamboure:
One this which is not clear to me though. Until now when a user checkout as guest, then creates an account, on the customer list there is not a second line created for that user, only one record.
I just saw it while testing #2368
So with this PR, this logic will change right? If a user checkout as guest and then becomes a registered user there will be two records for the user? Guest orders will be associated to guest record and then allnew orders to registered user. That might impact the logic you implemented in 2368 @Matt-Yorkley I think as you will see two records available for the same customer email. In term of UX probaby we would need to add some filter to display/hide guest records as it might become quickly hard to read for hub managers if there are "duplicates" ... I don't know yet but we'll need to test ux.

myriamboure:
Sometimes I think all in all that not authorizing guest checkout would simplify a lot our app...

app/assets/stylesheets/admin/customers.css.scss
sauloperez: I think it'd make sense to start doing so hoping to reach 100% one day. At least this will make it much easier to reuse colors and have a sort-of style guide. What do others think?

mkllnk:
I think it would be great to define our colour palette in a file. But I would leave it up to @Matt-Yorkley if he is keen or not.

myriamboure:
@Matt-Yorkley any update on that one?

Matt-Yorkley:
I've implemented a new SCSS colour palette for the backend. 

I had a quick search around for some of the most-used colours and added them to the palette, and replaced the references to those colours throughout the admin stylesheets. There's still a lot of colours in different files that are unique or are used once or twice, but reviewing/changing/adding all of them would take a long time!

mkllnk:
Hm, our CI services our consistently failing on this one. Travis fails to run the Javascript specs and Semaphore fails on a brittle feature test, failing to click an element. I restarted both several times, but I think we have to investigate more to fix it.

Matt-Yorkley:
Investigated and fixed. I had to tweak the scss `@import` path, as it was missing the `/components` subfolder, which caused the UI issues in those failing specs.

Matt-Yorkley:
Deployed on: https://staging.openfoodnetwork.org.uk
Ah... images are currently broken on UK staging. That's unrelated to this PR... I'll take a look at it.

sauloperez:
do not forget the `pr-staged-*` label :warning:

RachL:
@sauloperez as we have to wait until UK staging's server is fixed. Shall we stage this PR to katuma's staging in the meantime ?

Matt-Yorkley:
All the images are broken on UK staging, I'll try to sort that out ASAP, but I think we should wait until then.

Matt-Yorkley:
Yes, that's what I was doing before and it created errors when precompiling assets. I notice in `/darkswarm` every file that uses it's variables explicitly imports them at the top of the file. It might be related to the weird way we split our compiled CSS files into 2.

Matt-Yorkley:
It's deploying fine now. No errors during asset precompile.

RachL:
Hey @Matt-Yorkley are emails working on UK staging ? Because I'm trying to confirm my email address for the test (I've done an order as guest, and now I would like to signup), and I don't receive the confirmation email...

Matt-Yorkley:
I've just fixed emails on UK staging so it should be okay now.

RachL:
@myriamboure for your info, there is still just one line on customer list. When the customer did a guest checkout, there is a label, and once he/she registers, the label is removed.
For me it is working and ok :)
My testing notes : 
https://docs.google.com/document/d/1fSlPo9inLWNZ07BAGTf8K4FV1L83p6oWGxSIlUtWm1w/edit#

PR #2380, opened 2018-06-15
built a custom stripe payment form in order to control mobile rendering

### LINKED ISSUE #2352: Stripe: Mobile Checkout layout unusable
When a user tries to checkout using Stripe on mobile the screen layout is such that they cannot input card details or checkout.

## Description

The layout of the checkout screen when paying by card is unusable with fields rendering on top of each other.

While a huge amount of work needs to be done on mobile in general this is a bug affecting existing users that are switching to use Stripe and losing customers. Thus for UK this is a bug.

## Expected Behavior

All fields should render in such a way that all can be seen and users can input their details.

## Actual Behavior

The layout of the checkout screen when paying by card is unusable with fields rendering on top of each other.

## Steps to Reproduce

1. Go to cart and select Stripe payment method
2. Attempt to enter payment details

## Animated Gif/Screenshot

mobilecardpayment

## Context

Stopping checkout for users on mobile.

## Severity

S3 - Stopping checkouts for existing users on some hubs. Workaround is use a PC, which is not a functional workaround for some users.

## Your Environment

* Version used: 1.15.1
* Browser name and version:  I was using the Chrome browser (version 66.0.3359.158) 
* Operating System and version (desktop or mobile): Samsung Galaxy S7, with Android 7.0.

luisramos0:
#### What? Why?

First step to fix #2352

Separated default stripe checkout form into 3 separate/custom parts to control mobile rendering, basically, render card number, expire date and cvc in different lines.

#### What should we test?

Checkout process with stripe payment methods.
Must be tested on mobile.

#### Release notes

New stripe checkout form to input card details, with correct rendering on mobile.

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/stripe_elements.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/stripe_elements.js.coffee
@@ -1,29 +1,49 @@
 Darkswarm.directive "stripeElements", ($injector, StripeElements) ->
   restrict: 'E'
-  template: "<label for='card-element'>\
-             <div id='card-element'></div>\
-             <div id='card-errors' class='error'></div>\
+  template: "<div class='card-element'>\
+             <label for='card-number-element'>\
+             <span>Card number</span>\
+             <div id='card-number-element' class='field'></div>\
+             </label>\
+             <label for='card-expiry-element'>\
+             <span>Expiry date</span>\
+             <div id='card-expiry-element' class='field'></div>\
+             </label>\
+             <label for='card-cvc-element'>\
+             <span>CVC</span>\
+             <div id='card-cvc-element' class='field'></div>\
+             <div id='card-errors' class='error'></div>\              
              </label>"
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/darkswarm/directives/stripe_elements.js.coffee
```diff
@@ -1,29 +1,49 @@
...
+             </label>\
+             <label for='card-expiry-element'>\
+             <span>Expiry date</span>\
+             <div id='card-expiry-element' class='field'></div>\
+             </label>\
+             <label for='card-cvc-element'>\
+             <span>CVC</span>\
+             <div id='card-cvc-element' class='field'></div>\
+             <div id='card-errors' class='error'></div>\              
              </label>"
```
stveep: Since this is getting more complicated perhaps it should be a separate haml file in the templates/ folder.

luisramos0:
Maybe we build a form like this in the future, for now we will just hide postcode field. Closing.

PR #2377, opened 2018-06-14
Pass on redirect path instead of URL

mkllnk:
#### What? Why?

Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/2376

The checkout doesn't deal with absolute URLs since
fc2cc09ea5e6c123243dac27b0f5dff601ac518b.

#### What should we test?

Checking out using Paypal. Nothing else is affected.

#### Release notes

Fixed checkout with Paypal.

Issue #2376, opened 2018-06-14
Using Paypal redirects to the wrong URL


mkllnk:
Paypal checkouts are broken, because they redirect to the wrong URL.

## Description

When you go to the checkout and try to pay via Paypal, it redirects you to an invalid URL and you just see an error message in your browser. This bug got introduced by fc2cc09ea5e6c123243dac27b0f5dff601ac518b.

## Expected Behavior

A Paypal form should appear.

## Actual Behavior

Error screen: "Server not found"

## Steps to Reproduce

1. Add items to your cart.
2. Go to checkout.
3. Select Paypal.
4. Click "Place order now"

## Animated Gif/Screenshot
![peek 2018-06-14 14-43](https://user-images.githubusercontent.com/3524483/41392006-634b28de-6fe1-11e8-89e6-80e62f961e1a.gif)

## Context

The Australian shop Guildford Foodhub reported that issue.

## Severity

s1: Can't checkout. All shops using Paypal are affected.

## Your Environment

* Version used: c72a2fda91002d3691f7b4666ca23b2feb0d32e7

## Possible Fix

We either consider absolute URLs when redirecting or we return a relative URL (path) during checkout.

PR #2337, opened 2018-05-26
Fix guest checkout login modal

Matt-Yorkley:
#### What? Why?

Closes #2277

After using the guest checkout the login modal was appearing on the order completed page.

#### What should we test?
Guest checkout works without the login modal opening afterwards.

#### Release notes

Minor fix for #2277, login modal now doesn't open unexpectedly after using guest checkout.

Issue #2277, opened 2018-05-11
Login modal shows after checkout for users who have previously tried to checkout as a guest as registered user.


sstead:
## Description
This bug was found while testing https://github.com/openfoodfoundation/openfoodnetwork/pull/2073. If a user tries to checkout as a guest with a registered users's email, then that user gets told 'you can't do that' and they then use another email to checkout as a guest, when they see the order confirmation screen, the login modal pops up. 

## Expected Behavior
The login modal shouldn't pop up.

## Actual Behavior
A login modal pops up over the order confirmation page.

## Steps to Reproduce
1. Checkout as a guest in any shop
2. Try to use the email of a registered user - you'll be told you can't do that, you must either login, or remain a guest but use another email. Close the login modal with the error message.
3. Use a different un-registered email. 
4. checkout- on the confirmation page you'll see the login modal pop up unneccessarily. 

## Animated Gif/Screenshot
See the modal at the end of this video... https://www.useloom.com/share/4d1180870f06415b83ec80c68c6dd99b

## Context
It's a very minor bug, a user would probably just close the modal.

## Severity
S5

## Your Environment
PC, chrome, Aus production

## Possible Fix
Stop the modal opening.

myriamboure:
@Matt-Yorkley as you have worked on this issue it might be quicker for you to have a look and fix it. Can you take this one? Still part of email confirm work as was introduced by it, but last small bits!

PR #2381, opened 2018-06-15
2376 improve checkout redirects

mkllnk:
#### What? Why?

Addresses #2376 which is already closed. This is another improvement that would have fixed it as well.

A small change in #2337 broke Paypal. But the problem was not with that change, but that other parts of the code didn't behave as expected. We got an absolute URL as `path` from the checkout controller and `Navigation.go` was preserving hash fragments.

While #2377 changed the checkout controller, this pull request makes the Navigation service more robust to prevent bugs like this from happening in the future.

#### What should we test?

The changed code is used during registration, checkout and changing credit card details. We need to check that every click is bringing us to the right page and that modals open correctly. Especially the redirect on checkout should be tested with different payment methods.

The checkout should also be tested on mobile to verify that mobile browser are doing the same thing.

#### Release notes

Code improvements to prevent bugs.

### DIFF
--- a/app/assets/javascripts/darkswarm/services/checkout.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/checkout.js.coffee
@@ -13,7 +13,7 @@ Darkswarm.factory 'Checkout', ($injector, CurrentOrder, ShippingMethods, StripeE
     submit: =>
       Loading.message = t 'submitting_order'
       $http.put('/checkout.json', {order: @preprocess()}).success (data, status)=>
-        Navigation.goWithoutHashFragments data.path
+        Navigation.go data.path
       .error (response, status)=>
         if response.path
           Navigation.go response.path
--- a/app/assets/javascripts/darkswarm/services/navigation.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/navigation.js.coffee
@@ -21,10 +21,9 @@ Darkswarm.factory 'Navigation', ($location, $window) ->
       $window.location.href = $window.location.origin + path
 
     go: (path)->
-      if path.match /^http/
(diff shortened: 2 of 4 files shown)


Matt-Yorkley:
It looks much nicer. :+1: 

I think we'll have to retest the bug in issue #2277 after this and if the bug is back just reopen it and I'll try to find a better way to strip the `/login`, `/signup` and `forgot` path variables from the URL redirection whilst leaving the others intact?

Matt-Yorkley:
I'm surprised there wasn't a spec covering successful Paypal payments. The specs all came back green when I made the initital change in #2337, so I thought it was all ok.

mkllnk:
@Matt-Yorkley said:

> I'll try to find a better way to strip the /login, /signup and /forgot path variables from the URL redirection whilst leaving the others intact?

Is this still a problem? My solution navigates to a new href as if the user had clicked on a link. It doesn't preserve the hash fragments of the previous URL like when just setting the pathname. If you add hash fragments to the path, it will still work. But did I forget cases in which we were relying on keeping the hash fragments?

Matt-Yorkley:
I'm not really sure what cases there might be with the hash fragments. I guess we can just stage and test it.

mkllnk:

sstead:
**Testing notes**

Working fine
- Redirects while registering a new profile
- Redirects when checking out across all payment methods (except MIGS)
- Redirects when in Customer Account and changing saved credit cards
- Tested on mobile and PC, all fine.

https://docs.google.com/document/d/1-wvw0uCrU05ChlhZ4bfhJIPIiTyJsDJ9huZojgbKow0/edit

PR #2383, opened 2018-06-15
Fix NoMethodError in order cycles index

### LINKED ISSUE #2271: NoMethodError in admin/order_cycles#index
## Description
Katuma's Bugsnag reported the following error a week ago.

captura de pantalla 2018-05-09 a les 15 19 36

## Expected Behavior
At the point the `OrderCycle` model is used `user` should always be defined.

## Actual Behavior
The app fails with the mentioned error when issuing the following request: 

```
curl --request GET \
  --header 'Accept: application/json, text/javascript, */*' \
  --header 'X-Csrf-Token: 7xGaO9wLnV3Ey6PjKvrt9KY/4u+VXVpFEYZFAiERzaI=' \
  --header 'User-Agent: Mozilla/5.0 (Linux; Android 7.0; SM-J710F Build/NRD90M) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/66.0.3359.126 Mobile Safari/537.36' \
  --header 'Referer: https://alpha.katuma.org/admin/order_cycles' \
  --header 'Accept-Encoding: gzip, deflate, br' \
  --header 'Accept-Language: es-ES,es;q=0.9' \
  --header 'Cookie: [FILTERED]' \
  'https://alpha.katuma.org/admin/order_cycles.json?ams_prefix=index&q%5Borders_close_at_gt%5D=Sun+Apr+01+2018+00:00:00+GMT%2B0200+(CEST)'
```

## Steps to Reproduce

1. Log in as admin
1. Navigate to Order Cycles
1. Clear session in browser manually
1. Click "Show 30 more days"
1. 💥 

## Severity
The reach of this bug is still unknown but doesn't seem to affect too many users, so severity 3.

## Your Environment
* Version used: v1.14
* Browser name and version: Possibly Firefox
* Operating System and version (desktop or mobile): Android

frank-west-iii:
#### What? Why?

Closes #2271 

When a user's session has timed out and they try to load new data on the
order cycles page by changing filters, the application throws a
`NoMethodError` because we are prepending the load data method before
checking the user's session.

We can fix this by removing the prepend on this action.

#### What should we test?

Loading and filtering the order cycles page. (Show 30 more days or Show 90 more days)

### DIFF
--- a/app/controllers/admin/order_cycles_controller.rb
+++ b/app/controllers/admin/order_cycles_controller.rb
@@ -6,7 +6,7 @@ module Admin
   class OrderCyclesController < ResourceController
     include OrderCyclesHelper
 
-    prepend_before_filter :load_data_for_index, :only => :index
+    before_filter :load_data_for_index, only: :index
     before_filter :require_coordinator, only: :new
     before_filter :remove_protected_attrs, only: [:update]
     before_filter :check_editable_schedule_ids, only: [:create, :update]


oeoeaio:
Hm, I can't remember exactly, and the specs are not that helpful, but I suspect it was something to do with altering the ransack query params in the `load_data_for_index` method, before the default collection setting logic from `Spree::Admin::ResourceController` was hit.

It looks like 2a5f598fb0 rewrote most of this stuff anyway, and if the current version of the controller specs aren't broken by this change, then I think we just have to trust that everything is ok. We should still do rigorous testing around loading more order cycles using the button on the index though.

mkllnk:
Thank you for the explanation @oeoeaio.

sauloperez:
Fingers crossed...

mkllnk:

RachL:
@mkllnk I've tried to login as spree@example on staging but it's not working. What kind of other account can I use to test this ? Thanks :)

myriamboure:
@RachL we use another user to test on Aus staging I got the credential I give them to you... I know it's still super messy we really need to have those common seeds data implemented please!!!

RachL:
Thanks for the login info @myriamboure ! This is ready to go. However is there already somewhere an issue open in order to display a message in that case?
I'm afraid users are more likely to refresh their page when they see an error rather than an infinite spinner... but that's just my opinion.
Testing notes : 
https://docs.google.com/document/d/12vNkQfzsBoKkXholzkCAPpHXzGQf4BC3ODAjPeFe86U/edit#

mkllnk:
I just tested this myself, because the infinite spinner doesn't tell if the server encountered an error or not.

Previously, when you logged out in another window and then tried to load more order cycles, the server would throw an error. Now it recognises correctly, that the user is not logged in and redirects to the login. Since this happens in a JSON request, the Javascript doesn't know what to do with the result and just shows that infinite spinner. A user is most likely to refresh the page which will then prompt for login.

Pass. This is much better then before.

As Rachel said, we need a new issue for the infinite spinner though. In JSON requests, the redirect should be handled differently. It should be a valid JSON reply telling the browser to show the login modal (or redirect there).

PR #2385, opened 2018-06-15
Remove taxon when primary taxon is changed

### LINKED ISSUE #1250: Changing a product's category should REPLACE the pre-existing taxon, not ADD an additional one
image

## Note from Myriam March 2018: 
I just retested and the problem is still there and we have another problem on top: as we delete the "product taxons" to simplify product edit screen, there is no workaround anymore to delete the old unappropriated taxon :-(
So this is pretty annoying. don't know how many users are affected though.

## Expected behavior: 
if you change category don't add the new category on top (new taxon) but just delete the old one and replace by new one.

## Severity
s3 as not preventing from using the platform so not major feature but pretty annoying.

frank-west-iii:
#### What? Why?

Closes #1250 

We are adding taxons to the product as you change the primary taxon.
However we never remove the previous primary taxon so it forces the user
to update the taxons manually. This can be a big problem if you are bulk
updating products.

We now remove the taxon that matches the previously set primary taxon.

#### What should we test?

Updating a product's primary taxon (Category) either through bulk updating or normal update adds the new taxon to the product's taxon list and removes the previous primary taxon from the product's taxon list.

#### Release notes

When changing a product's primary taxon the system now removes the previously set primary taxon from the product's taxon list automatically.

### DIFF
--- a/app/models/spree/product_decorator.rb
+++ b/app/models/spree/product_decorator.rb
@@ -38,6 +38,7 @@ Spree::Product.class_eval do
   before_validation :sanitize_permalink
   before_save :add_primary_taxon_to_taxons
   after_touch :touch_distributors
+  after_save :remove_previous_primary_taxon_from_taxons
   after_save :ensure_standard_variant
   after_save :update_units
   after_save :refresh_products_cache
@@ -243,6 +244,11 @@ Spree::Product.class_eval do
 
   def add_primary_taxon_to_taxons
     taxons << primary_taxon unless taxons.include? primary_taxon
+  end
+
+  def remove_previous_primary_taxon_from_taxons
+    return unless primary_taxon_id_changed? && primary_taxon_id_was
+    taxons.destroy(primary_taxon_id_was)
   end
 
   def self.all_variant_unit_option_types
--- a/spec/models/spree/product_spec.rb
+++ b/spec/models/spree/product_spec.rb
@@ -193,6 +193,22 @@ module Spree
           expect { product.delete }.to change { distributor.reload.updated_at }
         end
       end
+
+      it "adds the primary taxon to the product's taxon list" do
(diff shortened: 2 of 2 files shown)


frank-west-iii:
@mkllnk, yes I think your solution is much better in the fact that it removes 3 callbacks from the system.

I can look at this proposed change, but since I am largely unfamiliar with the system as a whole, ensuring that the places where we query for or against taxons would be very important in review.

I will look this morning at this.

mkllnk:
Thank you for investigating @frank-west-iii. I should have thought of the entanglement with Spree. Let's go with your current pull request.

mkllnk:

sstead:
**Testing notes**

Working
- The main issue has been fixed - when I change the primary category the original one is removed and replaced in the taxon field.
- I can still add multiple taxons in the Edit Product page, ,and these will correctly carry through to shops/profile/reports

Needs attention
- On the edit product page I can't see the Category dropdown. This causes some confusion when trying to replace the primary taxon from the Edit Product page- because the field isn't showing I can't see how to replace the primary taxon.

https://docs.google.com/document/d/1-TPczXpprjNdNWRANtMX7at0lpMjKTSssW3M3vRMWCY/edit#

mkllnk:
Okay, that is an unrelated problem. Do we need a new issue for that?

sstead:
I can make a new issue for it @mkllnk . Are you sure it's not this PR though? On production the product category field was still visible. I just want to double check.

mkllnk:
@sstead I can't imagine how this code could break it. This change is on production now. You can double check there.

myriamboure:
@sstead any news on this double check? If you need to open an issue to solve that UX problem please do it ;-)

sstead:
Confirming, its an issue on staging, not related to this PR :)

PR #2386, opened 2018-06-15
Set the SSL protocol for secure connections

frank-west-iii:
#### What? Why?

There are ssl errors when using stripe through phantom js. This allows
other SSL protocols now.


mkllnk:
Since this is not affected production code (only CI test runs), it doesn't need further testing. Both CI services agree that this works. So I will merge this straight away, because it blocks all other pull requests.

PR #2387, opened 2018-06-16
Embedded response headers

### LINKED ISSUE #2382: Embedded shops aren't working when you go to checkout
## Description
I'm able to embed my shop into my own website. However, when i try to place an order in the embedded shop I can't get to the 'edit my cart' or 'checkout' pages. 

## Expected Behavior
I should be able to checkout in the embedded shop

## Actual Behavior
I see a blank screen when i try to checkout in an embedded shop. 

## Steps to Reproduce
1. Go to an embedded shop. here's an example - https://openfoodnetwork.org/user-guide/advanced-features/demo-embedded-shop/
2. Put an item in your cart
3. Click edit your cart or checkout
4. You'll see a blank screen- unable to checkout.

## Animated Gif/Screenshot
https://www.useloom.com/share/77734df7480e41f19708df787942bfcc

## Context
I'm trying to have a functional embedded shop. 

## Severity
S2

## Your Environment
* Version used: 1.16
* Browser name and version: Chrome, PC.

Matt-Yorkley:
#### What? Why?

Closes #2382.

Issues with checking out whilst using embedded shopfronts. The initial page shows correctly but then the response headers cause the page to be blocked when moving to cart of checkout. Possibly due to not including the entire whitelist in the response headers since #2026.

#### What should we test?

Embedded  shopfronts work and can go to checkout.

#### Release notes

Issues with moving to the checkout page when using embedded shopfronts should be fixed. #2382

### DIFF
--- a/app/services/embedded_page_service.rb
+++ b/app/services/embedded_page_service.rb
@@ -1,0 +1,90 @@
+# Processes requests for pages embedded in iframes
+
+class EmbeddedPageService
+  attr_reader :use_embedded_layout
+
+  def initialize(params, session, request, response)
+    @params = params
+    @session = session
+    @request = request
+    @response = response
+
+    @embedding_domain = @session[:embedding_domain]
+    @use_embedded_layout = false
+  end
+
+  def embed!
+    return unless embeddable?
+    return if embedding_without_https?
+
+    process_embedded_request
+    set_response_headers
+    set_embedded_layout
+  end
+
+  private
+
+  def embeddable?
+    return true if current_referer == @request.host
+
+    domain = current_referer_without_www
+    whitelist = Spree::Config[:embedded_shopfronts_whitelist]
+
+    embedding_enabled? && whitelist.present? && domain.present? && whitelist.include?(domain)
+  end
+
+  def embedding_without_https?
+    @request.referer && URI(@request.referer).scheme != 'https' && !Rails.env.test? && !Rails.env.development?
+  end
+
+  def process_embedded_request
+    return unless @params[:embedded_shopfront]
+
+    set_embedding_domain
+
+    @session[:embedded_shopfront] = true
+    set_logout_redirect
+  end
+
+  def set_response_headers
+    @response.headers.delete 'X-Frame-Options'
+    @response.headers['Content-Security-Policy'] = "frame-ancestors 'self' #{@embedding_domain}"
+  end
+
+  def set_embedding_domain
+    return unless @params[:embedded_shopfront]
+    return if current_referer == @request.host
+
+    @embedding_domain = current_referer
+    @session[:embedding_domain] = current_referer
+  end
+
+  def set_logout_redirect
+    return unless enterprise_slug
+    @session[:shopfront_redirect] = '/' + enterprise_slug + '/shop?embedded_shopfront=true'
+  end
+
+  def enterprise_slug
+    return false unless @params[:controller] == 'enterprises' && @params[:action] == 'shop' && @params[:id]
+    @params[:id]
+  end
+
+  def current_referer
+    return if @request.referer.blank?
+    URI(@request.referer).host.downcase
+  end
+
+  def current_referer_without_www
+    return unless current_referer
+    current_referer.start_with?('www.') ? current_referer[4..-1] : current_referer
+  end
+
+  def set_embedded_layout
+    return unless @session[:embedded_shopfront]
+    @use_embedded_layout = true
+  end
+
+  def embedding_enabled?
+    Spree::Config[:enable_embedded_shopfronts]
+  end
+end
--- a/app/services/embedded_page_service.rb
+++ b/app/services/embedded_page_service.rb
@@ -1,8 +1,6 @@
 # Processes requests for pages embedded in iframes
 
 class EmbeddedPageService
-  attr_reader :use_embedded_layout
-
   def initialize(params, session, request, response)
     @params = params
     @session = session
@@ -20,6 +18,10 @@ class EmbeddedPageService
     process_embedded_request
     set_response_headers
     set_embedded_layout
+  end
+
+  def use_embedded_layout?
+    @use_embedded_layout
   end
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -58,7 +58,7 @@ class ApplicationController < ActionController::Base
   def enable_embedded_shopfront
     embed_service = EmbeddedPageService.new(params, session, request, response)
     embed_service.embed!
-    @shopfront_layout = 'embedded' if embed_service.use_embedded_layout
+    @shopfront_layout = 'embedded' if embed_service.use_embedded_layout?
   end
 
   def action
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -56,50 +56,9 @@ class ApplicationController < ActionController::Base
   end
 
   def enable_embedded_shopfront
-    return unless embeddable?
-    return if embedding_without_https?
-
-    response.headers.delete 'X-Frame-Options'
-    response.headers['Content-Security-Policy'] = "frame-ancestors 'self' #{URI(request.referer).host.downcase}"
-
-    check_embedded_request
-    set_embedded_layout
-  end
-
-  def embedded_shopfront_referer
-    return if request.referer.blank?
-    domain = URI(request.referer).host.downcase
-    domain.start_with?('www.') ? domain[4..-1] : domain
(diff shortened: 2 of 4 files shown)


myriamboure:
Ok this one then needs to be merged before we release or we'll have issues with all existing embedded shops, right? Ping @sauloperez if you review quickly and stage today we can test and that's the last thing I think we need to include before we can release...

app/services/embedded_page_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def set_embedded_layout
+    return unless @session[:embedded_shopfront]
+    @use_embedded_layout = true
+  end
+
+  def embedding_enabled?
+    Spree::Config[:enable_embedded_shopfronts]
+  end
+end
```
sauloperez: I'm missing unit tests for this whole class 🙈

app/services/embedded_page_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    process_embedded_request
+    set_response_headers
+    set_embedded_layout
+  end
+
+  private
+
+  def embeddable?
+    return true if current_referer == @request.host
```
sauloperez: things are more maintainable and easier to change in the future if we use `attr_reader` for the ivars.

Matt-Yorkley: I'm not sure what you want in `attr_reader`? 

The request object is passed in to the class and altered in various places, but doesn't need to be explicitly read from outside of the class. It needs to be available to the various methods internally but it doesn't need to be readable...

app/services/embedded_page_service.rb
Matt-Yorkley: Ah, in my haste to fix the urgent s2 I forgot to spec the new service object and incurred the wrath of the monkey-face emoji... 

I've added specs now. :+1:

myriamboure:
Cool that you staged and tested @Matt-Yorkley ! Do you have some testing notes just to know which cases you tested? We have some processes around testing to have some peer review on them if needed (see wiki) :-) Is it still staged on UK? If yes @RachL can have a look as well if needed (you will need access to global wordpress to test with the demo shop Rachel we talk in PM). If you think you tested all the cases and it works and we don't need to stage again then tell us :)

app/services/embedded_page_service.rb
sauloperez: My point is to use private reader methods instead of the raw ivars. That brings the usual benefits of encapsulation.

Matt-Yorkley:
It should still be up on https://staging.openfodnetwork.org.uk. I just thought thought it would be good to test it myself as it's not always straightforward to do, and I wanted to assess the request and response headers in different cases, which isn't a normal part of testing.

app/services/embedded_page_service.rb
Matt-Yorkley: I'm not sure I understand what you mean? Adding `attr_reader` and `attr_accessor` lines under `private` will just create private methods for accessing those instance variables, right? But they can't be accessed outside of the instance anyway, and all but one of the methods are private to begin with. 

The class needs to be able to modify the objects referenced by `@resonse` and `@session`. Actually, that's basically it's primary job. It looks like it only needs to read the other two objects, so I guess we could use:

```
private
attr_reader :request, :params
attr_accessor :session, :response
```
But I don't really understand what would be gained from this? There's already no getters or setters for those variables, so this doesn't limits access in any way. Is the intention to make it more explicit that those variables are only accessible from within an instance of this class?

mkllnk: I think the intention here is to standardise access to always use methods instead of direct access to the ivars. The advantage is that it's easier to change behaviour later or move parts into other classes without changing the code accessing these variables. The disadvantage is that you can't see immediately if a call to `request` has side effects or if it is private. You need to look up the accessor/reader first. That's why I'm not clear on this recommendation.

app/services/embedded_page_service.rb
```diff
@@ -0,0 +1,90 @@
+# Processes requests for pages embedded in iframes
+
+class EmbeddedPageService
+  attr_reader :use_embedded_layout
```
mkllnk: Now here is a disadvantage of the attr_reader. This method should be called `use_embedded_layout?`, right?

sauloperez: let's make that a regular method if we want the question mark then. It's the only way if we want it.

app/services/embedded_page_service.rb
sauloperez: Yes, that's my point. 

Regarding

> The disadvantage is that you can't see immediately if a call to request has side effects or if it is private. You need to look up the accessor/reader first. 

I haven't experienced problems with this unless the class is huge. Once you open the class, from the constructor you see what vars compose its internal state and what is all about. However, this encapsulation brings much bigger benefits to changeability. 

It's relatively easier to move private methods to the public API without having to change much, or create a new subclass and rely on those reader methods or even to turn simple `attr_reader`s to regular methods when their logic grows, while keeping the consumers untouched. Not to mention the fact that a typo on a method call raises an error, while a typo on an ivar's name does not; that's really useful!

Having said that, I don't want to be picky. You guys decide.

Matt-Yorkley: It looks like I would also have to convert lines like this:
`@embedding_domain = current_referer`

to use this syntax:
`instance_variable_set '@embedding_domain', current_referer`

because this won't work for assigning to the private ivars via methods:
`embedding_domain = current_referer`

mkllnk: I'm not sure I know what you mean. You last line doesn't work even if you have a private `attr_writer`? Anyway, I'm moving this to test ready and keep all the comments in mind. I will probably make more use of accessor in the future.

sauloperez: I don't follow it either. Look at https://github.com/codetriage/codetriage/blob/master/app/services/git_hub_authenticator.rb. This is a clear example of what I mean.

myriamboure:
@Matt-Yorkley it seems all reviews are green, wouldn't you mind staging on UK staging? And maybe testing if you have an easy way to test? Else let @RachL and me know when it is staged and we can organize with the global wordpress and ask a user who we know pretty well for websites with www. (we had some issues with testing only in wordpress earier...) We are waiting this to be merge for releasing!!! Let's do it!!!

app/services/embedded_page_service.rb
Matt-Yorkley: It works nicely when reading from ivars via private attr_reader methods, but writing to those attributes needs some odd-looking rewrites using `#instance_variable_set`:
```ruby
private
attr_reader :private_ivar_one
attr_accessor :private_ivar_two

def example
example_local_variable = private_ivar_one # reading works fine and is clear/concise

private_ivar_two = 'new_value' # doesn't write to ivar, creates new local variable
instance_variable_set @private_ivar_two, 'new_value' # writes to the ivar, but isn't pretty
end
```

sauloperez: Now I understand what you mean. That's what `self` is for. Check out [this](http://rubyfiddle.com/riddles/34ec5) example.

Matt-Yorkley: Ah that's really obvious now, I don't know why I was only thinking of more complicated solutions like `send(private_ivar='new_value')` and everything except the simple answer... :see_no_evil:

Matt-Yorkley:
@myriamboure Sorry for intervening in the usual testing process. This fix was a bit more urgent for the UK, so we were planning on pushing it to production before the next release. I tested it on UK staging using another site to embed a shopfront, and I wanted to be completely confident that pushing it to production early wouldn't cause any problems (and would fix the issue).

I checked to see if things were working generally, but mostly I was looking at the how the response headers behaved when moving to the checkout and cart, as that was the cause of the issue (and what was changed in the PR). Addition of the `'self'` keyword in the `Content-Security-Policy` header, and persistence of the embedding site's domain name on subsequent requests was my primary criteria:

screenshot from 2018-06-22 14-12-44

And here's a screenshot of the checkout page successfully showing in the embedded shop (basic pass criteria for the issue): 

screenshot from 2018-06-22 14-13-08

We've pushed this to UK production now, so you can see it working at: https://blackmountainfoodhub.co.uk/en/shop/ (which is live).

myriamboure:
Yes @Matt-Yorkley no issue with me, you changed things after @sauloperez feedback so we thought we needed to test again. But if you pushed to production and it's all working good live I guess it's the best test we can have so @sauloperez I think we can merge!

Matt-Yorkley:
@sauloperez do you want me to refactor to the private ivars first? And retest?

PR #2389, opened 2018-06-17
admin page fetcher is now synchronously fetching all pages before reso…

### LINKED ISSUE #436: Bulk product and orders pages should show 'loading' when still loading, rather than 'none found'
## Description
Confusing UX while filters are being applied on products, while info corresponding to filters are loading, "none found" message is displayed.

## Expected Behavior
When filter is applied and data are loading, a message "loading data" should be shown, like https://www.useloom.com/share/b36dec611e1c488c9872c6c8be2491ed

## Actual Behavior
See exemple from Product Import tests: https://www.useloom.com/share/41d9581879234356ad8b990d8ee29ef1
Only happen in some circumstances, I guess when lots of things to calculate...

## Steps to Reproduce
1. Go to "products" and select a shop for instance with many product. 
2. OR when product import available try to import a file and at the last step, click on "check product",  you are redirected to product page and it takes 10 sec before the products appear, and meanwhile you see "no product found"... if you are not patient enough you just  think it is broken.

## Animated Gif/Screenshot
image

## Context
Reported by Kirsten, Myriam hit again that issue while testing product upload, that was VERY confusing as a user, and moste user using product import will hit it, so will set up a high priority.

## Severity
S2 - most users will be concerned using product imports (and not only, basically all users), very confusing in term of UX and no workaround on UX! The feature can be used but it appears it's not working, so might appear it is broken.

## Your Environment
- Version used: 1.10.1

luisramos0:
…lving its promise (so that results are not inacurate due to filters)

#### What? Why?

Closes #436

Makes page fetch synchronous so that results are always accurate.

#### What should we test?

Products bulk edit list and filtering.
Inventory (variant overrides).

#### Release notes

Bulk product edit page will now show results only after fetching all results from server.

### DIFF
--- a/app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
+++ b/app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
@@ -3,14 +3,18 @@ angular.module("admin.indexUtils").factory "PagedFetcher", (dataFetcher) ->
     # Given a URL like http://example.com/foo?page=::page::&per_page=20
     # And the response includes an attribute pages with the number of pages to fetch
     # Fetch each page async, and call the processData callback with the resulting data
-    fetch: (url, processData) ->
+    fetch: (url, processData, onLastPageComplete) ->
       dataFetcher(@urlForPage(url, 1)).then (data) =>
         processData data
 
         if data.pages > 1
           for page in [2..data.pages]
-            dataFetcher(@urlForPage(url, page)).then (data) ->
+            lastPromise = dataFetcher(@urlForPage(url, page)).then (data) ->
               processData data
+          onLastPageComplete && lastPromise.then onLastPageComplete
+          return
+        else
+          onLastPageComplete && onLastPageComplete()
 
     urlForPage: (url, page) ->
       url.replace("::page::", page)
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -1,5 +1,6 @@
 angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout, $http, $window, BulkProducts, DisplayProperties, dataFetcher, DirtyProducts, VariantUnitManager, StatusMessage, producers, Taxons, SpreeApiAuth, Columns, tax_categories) ->
     $scope.loading = true
+    $scope.loadingAllPages = true
 
     $scope.StatusMessage = StatusMessage
 
@@ -49,7 +50,10 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
 
     $scope.fetchProducts = ->
       $scope.loading = true
-      BulkProducts.fetch($scope.currentFilters).then ->
+      $scope.loadingAllPages = true
+      BulkProducts.fetch($scope.currentFilters, ->
+        $scope.loadingAllPages = false
+      ).then ->
         $scope.resetProducts()
         $scope.loading = false
 
--- a/app/assets/javascripts/admin/services/bulk_products.js.coffee
+++ b/app/assets/javascripts/admin/services/bulk_products.js.coffee
@@ -8,7 +8,8 @@ angular.module("ofn.admin").factory "BulkProducts", (PagedFetcher, dataFetcher, 
       , ""
 
       url = "/api/products/bulk_products?page=::page::;per_page=20;#{queryString}"
-      PagedFetcher.fetch url, (data) => @addProducts data.products
+      processData = (data) => @addProducts data.products
+      PagedFetcher.fetch url, processData, onComplete
 
     cloneProduct: (product) ->
       $http.post("/api/products/" + product.id + "/clone").success (data) =>
--- a/app/views/spree/admin/products/index/_indicators.html.haml
+++ b/app/views/spree/admin/products/index/_indicators.html.haml
@@ -5,10 +5,10 @@
   %img.spinner{ src: "/assets/spinning-circles.svg" }
   %h1= t('.title')
 
-%div.sixteen.columns.alpha{ 'ng-show' => '!loading && filteredProducts.length == 0 && query.length==0' }
+%div.sixteen.columns.alpha{ 'ng-show' => '!loadingAllPages && filteredProducts.length == 0 && query.length==0' }
   %h1#no_results= t('.no_products')
 
-%div.sixteen.columns.alpha{ 'ng-show' => '!loading && filteredProducts.length == 0 && query.length!=0' }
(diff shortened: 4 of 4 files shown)


app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
```diff
@@ -9,8 +9,9 @@ angular.module("admin.indexUtils").factory "PagedFetcher", (dataFetcher) ->
 
         if data.pages > 1
           for page in [2..data.pages]
-            dataFetcher(@urlForPage(url, page)).then (data) ->
+            lastPagePromise = dataFetcher(@urlForPage(url, page)).then (data) ->
               processData data
+          return lastPagePromise
```
luisramos0: I just realised that this synchronicity may be the original objective of the code but coffee script confused the author. Maybe... 
The original code in javascript actually returns an _array_ of "results" which is an array of promises. The new code returns just the last promise and that, according to my tests, makes the calls synchronous.
My explanation for this is that an array of promises is taken to be a value that resolves the original promise (and not taken as being a promise), while returning a promise does not resolve the original promise until this returned promise is resolved itself (the last page).

The js code generated with the original code:

      PagedFetcher.prototype.fetch = function(url, processData) {
        return dataFetcher(this.urlForPage(url, 1)).then((function(_this) {
          return function(data) {
            var i, page, ref, results;
            processData(data);
            if (data.pages > 1) {
              results = [];
              for (page = i = 2, ref = data.pages; 2 <= ref ? i <= ref : i >= ref; page = 2 <= ref ? ++i : --i) {
                results.push(dataFetcher(_this.urlForPage(url, page)).then(function(data) {
                  return processData(data);
                }));
              }
              return results;
            }
          };
        })(this));
      };

and the new code :-)


PagedFetcher.prototype.fetch = function(url, processData) {
        return dataFetcher(this.urlForPage(url, 1)).then((function(_this) {
          return function(data) {
            var i, lastPagePromise, page, ref;
            processData(data);
            if (data.pages > 1) {
              for (page = i = 2, ref = data.pages; 2 <= ref ? i <= ref : i >= ref; page = 2 <= ref ? ++i : --i) {
                lastPagePromise = dataFetcher(_this.urlForPage(url, page)).then(function(data) {
                  return processData(data);
                });
              }
              return lastPagePromise;
            }
          };
        })(this));
      };

sauloperez:
> Other pages that use Page Fetcher.

Can you please find these so we test thoroughly?

luisramos0:
This is used only in product bulk edit and in inventory (variant_overrides).
(meanwhile, I have added a unit test for this change)

mkllnk:
Okay @luisramos0. I just updated the PR description so that the tester will see these two areas to test.

oeoeaio:
This page is poorly written (I should know, it was the first thing I ever wrote in AngularJS).

Does this change mean that no products are displayed util all pages have been fetched? We have some users who are loading > 1000 products on this page, so forcing them to wait until all products are loaded before they can use the page is not really an option. I would consider this a fairly big UX regression.

I would much prefer to see this tackled as part of a rewrite where we stop attempting to load and paginate all products and instead request a new list of products from the server each time a new query is made using the filter elements.

daniellemoorhead:
Moving this back to code review status until conversation with @oeoeaio to address his question above has been answered.

Ping @luisramos0 @sauloperez 

traffic

Matt-Yorkley:
I think the idea is the products will still load, but the "No results found" message doesn't show until all products are loaded?

luisramos0:
@Matt-Yorkley that would be better but it's not true. In this solution, products will not load until all pages are fetched from server.
If you think that's not acceptable we need to build something more robust like @oeoeaio describes, server side filtering.

mkllnk:
I'm trusting Rob's judgement about the UX regression here. We need faster page loads. Moving back to dev.

luisramos0:
Yes, I understand. I'll have a look at this soon. At least to define how much work is involved.

luisramos0:
ok guys,
I think the rewrite Rob suggests is the way to go. I just don't want to get into that as I'm sure it would take me a while to get it done at this point.

So... I suggest a new fix to the problem (see commit above): this time, the fetching process takes a onLastPageComplete function that will only execute after the last page is loaded.
With this solution, the problem is solved because the no results message (or no products) will only show after all pages are fetched (and no results are found). The only little problem is that the "Loading" wheel will disappear after the first page is loaded. So, if there are no results, between the end of the loading of the first page and the load of the last page, nothing will be displayed in the list. Not perfect but it fixes the issue... I hope you like it :-)

luisramos0:
ok, thanks @mkllnk it feels hacky because it is hacky :'-(
and now with a green build :-)

sstead:
**Testing notes**
One issue:
- I saw a bit of a glitch when a superadmin switched between shops in the shop fitler, while on the New Products tab of Inventory. If a shop showed that had no New Products, the table would not refresh, showing incorrect products. I could then (seemingly) add these to Inventory.
- The above issue didn't impact non super admin users
- There didn't appear to be any other issues in Inventory or Products

https://docs.google.com/document/d/19RMibbP1wpmnyhaDWsBZfpIMQAU3FWs9Hpf46sjEP78/edit

myriamboure:
@sstead are we sure that issue is linked to that PR? You cannot reproduce in production? I'm just wondering if it is connected to that PR, sounds to me that it could not be. But maybe you already checked and it is linked to that PR ;-) Maybe @luisramos0 when looking at this you will know straightaway if it is connected to your hack or not ;-)

sstead:
I'm not certain. I did check Aus production and the problem wasn't occurring there, but this doesn't necessarily mean that it's this PR.

myriamboure:
Ok so back to you @luisramos0 :-)

luisramos0:
#### TEST
Hi @sstead , can you please reproduce the issue and check the console in chrome's dev tools? (in chrome that would be menu view  > developer > dev tools and then click Console). You will probably see an error there.
That would enable us to confirm you are having the issue I can see on my local instance. I can see a javascript error in this screen and I can replicate on the master branch. It's not related to this PR, it's something to do with hubPermissions.
Let me know what do you see on the console.

#### TECH
For another developer to check, if the issue is what I am seeing, there are no hubPermissions for some hubs and so the fix would be on variant_overides/filters/hub_permissions_filter.js to do this:
return $filter('filter')(products, ((product) -> hubPermissions[hub_id] && hubPermissions[hub_id].indexOf(product.producer_id) > -1), true)

sstead:
Does this help @luisramos0 ?
image

luisramos0:
Yes, it does. So, I don't think this is a problem with this PR.
Let's see if another dev will know what this issue with hubPermissions is.
Thanks.

mkllnk:
Okay, I created https://github.com/openfoodfoundation/openfoodnetwork/issues/2502 for this issue. This is ready for merging.

luisramos0:
Thanks a lot @mkllnk

PR #2390, opened 2018-06-17
fixed number rounding issue in product bulk edit

### LINKED ISSUE #2333: Volume quantity 350mL gets transformed to 349.999999 on some products
## Description
A producer sells a beverage that is 350mL volume. He enters 350 in "value" and select mL as unit, but the value gets transformed to 349.999999.

## Expected Behavior
It remains what is entered (350 here)

## Actual Behavior
It doesn't

## Steps to Reproduce
1.  Create a new product, select mL as unit and enter 350. 
2. See in bulk edit view.

## Animated Gif/Screenshot
https://www.useloom.com/share/1775169424c844438e880ec5c32a28c2

## Context
Reported by a producer on French instance.

## Severity
S3, there is a workaround using the dispay as field.

## Your Environment
* Version used: v1.12
* Browser name and version: Chrome Version 64.0.3282.167 (Official Build) (64-bit)

luisramos0:
#### What? Why?

Closes #2333

#### What should we test?

Unit quantities in product bulk edit page.

#### Release notes

Product bulk edit page now shows correct quantities.


app/assets/javascripts/admin/services/bulk_products.js.coffee
```diff
@@ -66,8 +66,12 @@ angular.module("ofn.admin").factory "BulkProducts", (PagedFetcher, dataFetcher,
...
-          variant.unit_value / product.variant_unit_scale
+          @divideAsInteger variant.unit_value, product.variant_unit_scale
         else
           variant.unit_value
       else
         null
+
+    # forces integer division to avoid javascript floating point imprecision
+    divideAsInteger: (a, b) ->
+      (a * 1000000) / (b * 1000000)
```
sauloperez: and I still wonder why some devs still love JS :trollface:

mkllnk: These values look pretty random to me. This fails again for really small unit scales: `(0.00000035 * 1000000) / (0.000000001 * 1000000) =  349.99999999999994`. Why did you decide against `Math.round`?

Even that's a minefield in Javascript. I found this one to round it to two decimals:

```js
value = a / b;
result = Math.round((value + Number.EPSILON) * 100) / 100;
```

luisramos0: hey @mkllnk I took it from here (first answer with 3 options):
https://stackoverflow.com/questions/1458633/how-to-deal-with-floating-point-number-precision-in-javascript
I choose the 3rd option (working with integers, as you usually do with money) to avoid using external library (option one) and because I am not sure rounding will always work. In this integer division solution, we know it will always work up to a given precision. I don't think we will have a scale of 0.000000001, and if we do, we can change the number from 1000000 to 1000000000. Overall, for this case, the integer division looks better and cleaner to me. But I am totally ok with switching it to round expression you shared! Let me know what do you prefer.

Matt-Yorkley: Won't wrapping the values in the parseFloat() function deal with the math issue? Also, there's an angular filter for displaying the result to specific decimal places if needed: https://docs.angularjs.org/api/ng/filter/number

Matt-Yorkley: e.g: `{{value_to_round | number: 2}}`

luisramos0: Hi @Matt-Yorkley parseFloat converts string to float, I don't see how it could be useful here.
That angular filter should be the same as Math.round. We could also use it.

mkllnk: I quite like Matt's rounding suggestion: `{{variantUnitValue | number: 2}}`. It makes clear that we have an imperfect number and we round it for display. We are not pretending to do the maths right which would require a library.

But I'm also happy with the current implementation if you add a comment about the chosen values and their limits. In your example we have `0.35 kg` displayed as `350 g` with `variant_unit_scale = 0.001`. Your method still works for a display as `350,000 mg` with `variant_unit_scale = 0.000001`. It would fail for anything smaller than mg or bigger than kg. ;-) Fair enough, let's document that. It makes sure that people dealing with tons (instead of kg) of cocaine, selling them as mg, have a problem with the OFN. I'm okay with that. ;-)

luisramos0: ok, thanks @mkllnk 
I was actually about to move to numberFilter but then realized it would change precision of the numbers to always include 2 decimal places (tests would need to be changed) and, also, the view receives a concatenation of value and description (unit_value_with_description), so would have to use the filter on the controller...
Well, I kept it as it is. I changed the multiplier from 1 million to 1 billion (it works for numbers up to 9 decimal places) and added a comment explaining that.
Thanks for the help!

mkllnk:

RachL:
Here are my testing notes : 
https://docs.google.com/document/d/1M2L2mYTzyBT6oM-_Cwsh0sq3aqhzJHglIIqE9hvVnoA/edit#
This is ready to go :)

PR #2391, opened 2018-06-18
Hide postcode - not necessary as passed from billing address and wrec…

stveep:
…ks UI on mobile #2352

#### What? Why?

Simple fix to test for:
https://github.com/openfoodfoundation/openfoodnetwork/issues/2352

The word "Postcode" is too long and overlaps the card number field on small screens. We can just hide the field it as it's not necessary (see issue discussion).

#### What should we test?

Check form is visible and usable

Check Stripe payments still go through ok.

#### Release notes

Removed postcode (ZIP) field from Stripe card form so it now works on small screen mobile devices.


daniellemoorhead:
Travis and semaphore both broken ☹️

Matt-Yorkley:
Looks like it needs a rebase to pick up the Stripe error fixes.

luisramos0:
I have created a rebased version of this PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/2403

Should fix the build.

Issue #1589, opened 2017-05-28
Email confirmation | Super admin adds new user


daniellemoorhead:
**Who:** Super Admin wanting to register a user on the OFN.
**How they access the sign up process:** via the backend admin area that only the super admin can access.

-------
**Step 1(a): Super Admin signs up the user**
The Super Admin accesses the add a user function in the backend and submits that user's email address and sets a password (existing functionality).
![image](https://user-images.githubusercontent.com/10843056/27939405-bb17fa0c-6307-11e7-82a6-4c76277df92e.png)

Note: we know that setting a password for a user is bad, and we've created #1612 to change this functionality so that the user sets it themselves.

## [THIS REMAINS TO BE DONE >>>]
After submitting, a 'pending verification' message displays in the interface (new functionality):
![image](https://user-images.githubusercontent.com/10843056/27939456-1c4e6a54-6308-11e7-8ed2-a13308886143.png)

**Step 1(b): Super admin resends confirmation email to user**
If the Super Admin needs to resend the confirmation email, they navigate to that user in the User section and click on the 'resend' link in the interface.
![image](https://user-images.githubusercontent.com/10843056/27939467-30827768-6308-11e7-8a9a-dc95526b3aba.png)

This triggers:
1. The same confirm email to be sent.
2. Success indicator - page reloads and the success banner shows
![image](https://user-images.githubusercontent.com/10843056/27939476-411d09da-6308-11e7-988f-d57d8c0f7061.png)

-------
**Step 2: System requests email confirmation**
The completion of step 1 will trigger the system to do 2 things:
1. System creates a user record, flagged as unconfirmed.
2. System sends a confirmation email to the new user that includes a unique confirm link for that user.

![image](https://user-images.githubusercontent.com/10843056/26963863-17060700-4d32-11e7-9133-24c5f7eb45ca.png)

-------
**Step 3: User confirms email**
The user clicking on the confirmation link in the email will trigger the following:
1. System flags the user record as confirmed.
2. System sends the user a 'Welcome' email (generic email as in #1585)

![image](https://user-images.githubusercontent.com/10843056/26963776-a89844ae-4d31-11e7-8908-1cdf3b83287e.png)

![image](https://user-images.githubusercontent.com/10843056/26963730-7316e3b2-4d31-11e7-9179-83764400da3f.png)

-------
**Step 4: User logs in**
Successful login from this screen will then close the modal and allow the user to continue browsing (existing functionality).

[...]

Issue #1612, opened 2017-06-09
When SuperAdmin creates a user they should not set their password


sstead:
This is bad practice...
- Instead the new user should get an email confirmation link, and when they click that link they should be prompted to set a password.
- The SuperAdmin user should not be able to see, change or reset the user's password. If they need to reset their password they can do it themselves.

This image shows the SuperAdmin view of creating a new user:
![image](https://user-images.githubusercontent.com/10843056/26957045-a0661b52-4d05-11e7-92bc-e4997b75246b.png)

Matt-Yorkley:
Is this the same as #1589 ?

sauloperez:
It would be worth taking a look at the spree release notes to see if this got ever fixed

myriamboure:
No @Matt-Yorkley it is precised in #1589 "Note: we know that setting a password for a user is bad, and we've created #1612 to change this functionality so that the user sets it themselves." I'll sum up in the other issue what is exactly missing and @ you. If you want to have a look at spree release to check if this one have been fixed that would be great :-)

sauloperez:
The only trace I found in Spree is that in the commit `8fb746c12` the `Spree::Admin::UsersController` has been moved to the api engine and apparently it just returns JSON.

However, when accessing this URL in the app there's an ajax request to `/admin/users/new` that returns the whole HTML markup, including the password confirm field. I haven't found how we manage to return that and whether it's Spree or OFN (I bet it's us)

myriamboure:
I know this is not totally in scope but do we want to include that in email confirmation link? As in term of stories it's on the same UX "user creation" flow, so as we are reviewing it with email confirm... maybe good to do it at the same time? What do you think @enricostano @sauloperez @Matt-Yorkley 

Also I have a question: **why is it such a bad practice?** I mean admins won't touch it, and anyway admins don't see the password and in case of any big issue, that might be useful for an admin to be able to change the user password and do tests on their account, or... I don't know, just wondering why it's bad and we need to change that.

mkllnk:
> why is it such a bad practice?

Setting a password as admin always comes with the problem how to communicate that password to the user. In most cases it's sent via email which is not a secure medium. The password is likely to end up on several computers (the sender, the mail servers, the receiver, backups) and is stored in plain text.

The admin can set a temporary password that has to be changed by the user, but most users don't do it.

myriamboure:
[...]

PR #2394, opened 2018-06-19
1589 admin email confirm dialog

mkllnk:
#### What? Why?

Addresses part of #1589.

screenshot from 2018-06-20 11-01-44

When a super-admin creates a new user, they should see that a confirmation email has been sent. They should also see if a confirmation is still pending and be able to resend it.

These changes do not alter the text of the confirmation message as specified in #1589. I found this would add a lot of complexity to the code and discussed with Danielle and Myriam that we leave it as is for now.

As discussed in the issue, I imported the connected Spree views and consolidated our overrides in our versions of the views. That simplifies the code makes it easier to see what will be rendered.

#### What should we test?

- Creating a user as super admin and resending the confirmation email.
- Signing up as normal user and resending the confirmation email.
- Sanity check that the users tab of enterprises and enterprise groups still works.

#### Release notes

Super-admins can now see if users need their email address confirmed. They can also send the confirmation again.

Since we imported the admin user views, we need to check if we need to update our views with the upgrade.


app/overrides/spree/admin/users/_form/add_enterprise_limit_form_element.html.haml.deface
```diff
@@ -1,5 +0,0 @@
-/ insert_bottom "div[data-hook='admin_user_form_fields'] div.alpha"
```
Matt-Yorkley: Some screenshots in the PR description/comments would be nice to go with the various UI changes.

app/views/spree/admin/users/_form.html.haml
```diff
@@ -0,0 +1,26 @@
...
+          %li
+            = check_box_tag "user[spree_role_ids][]", role.id, @user.spree_roles.include?(role), id: "user_spree_role_#{role.name}"
+            = label_tag role.name
+      = hidden_field_tag "user[spree_role_ids][]", ""
+    = f.field_container :enterprise_limit do
+      = f.label :enterprise_limit, t(:enterprise_limit)
+      = f.text_field :enterprise_limit, class: "fullwidth"
+  .omega.five.columns
+    = f.field_container :password do
+      = f.label :password, Spree.t(:password)
```
Matt-Yorkley: Is the admin user setting the user's password for them here? How are they going to tell the user what their password is? I think I ran into an issue like this before with emails and there was a way to set a random secure password, and flag the account as requiring a password reset. When the user followed the sign up link via their email they were taken straight to a "set your password" page by default via devise...

app/assets/javascripts/admin/users/directives/resend_user_email_confirmation.js.coffee
```diff
@@ -0,0 +1,19 @@
...
+
+    element.bind "click", ->
+      return if sent
+      element.text(text + sending)
+      $http.post("/user/spree_user/confirmation", {spree_user: {email: scope.email}}).success (data) ->
+        sent = true
+        element.addClass "action--disabled"
+        element.text text + done
+      .error (data) ->
+        element.text text + failed
```
oeoeaio: Angular is good at this kind of thing! If we specify a template for the directive, we can bind variables on the isolate scope to the template, rather than using jQuery to update it manually. Something like:

```coffee
angular.module("admin.users").directive "resendUserEmailConfirmation", ($http) ->
  template: "<span>{{ resend + ' ' + status }}</span>"
  scope:
    email: "@resendUserEmailConfirmation"
  link: (scope, element, attrs) ->
    scope.resend = t "js.admin.resend_user_email_confirmation.resend"
    scope.status = ""
    sent = false
    ....
    # Then when we want to change state we can just do:
    scope.status = sending
```

and the element itself would just be:

```haml
%a{"resend-user-email-confirmation" => @user.email}
```

app/views/spree/admin/users/_form.html.haml
mkllnk: Excellent question. Dannielle actually covered that:
> ##### Step 1(a): Super Admin signs up the user
> The Super Admin accesses the add a user function in the backend and submits that user's email address and sets a password (existing functionality).
>
> Note: we know that setting a password for a user is bad, and we've created #1612 to change this functionality so that the user sets it themselves.

So it's existing functionality that we are planning to tackle another time.

app/overrides/spree/admin/users/_form/add_enterprise_limit_form_element.html.haml.deface
mkllnk: Most of these changes don't change the UI at all. I was just importing from Spree and integrating our deface overrides into our view. But you are right, a screenshot would have been nice. I added one.

app/assets/javascripts/admin/users/directives/resend_user_email_confirmation.js.coffee
mkllnk: Thank you Rob, I added a commit making this much easier and easier to translate as well. Great feedback.

mkllnk:
Yes, I copied them, I converted them to HAML and applied our customisations. I think we agreed that on the issue discussion. Does it make sense to you?

spec/features/admin/users_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    describe "resending confirmation email", js: true do
+      let(:user) { create :user, confirmed_at: nil }
+
+      it "displays success" do
+        visit spree.edit_admin_user_path user
+
+        # The `a` element doesn't have an href, so we can't use click_link.
+        find("a", text: "Resend").click
```
oeoeaio: You could add an `href: 'javascript:void(0)'` to the `a` element if you want to fix this

spec/features/admin/users_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  context "as super-admin" do
+    before { quick_login_as_admin }
+
+    describe "creating a user" do
+      it "shows no confirmation message to start with" do
+        visit spree.new_admin_user_path
+        expect(page).to have_no_text "Email confirmation is pending"
+      end
+
+      it "works" do
```
oeoeaio: Unless there are too many things to mention, it might be nice to have a more descriptive title here?

How about "confirms successful creation with a flash message"?

spec/features/admin/users_spec.rb
mkllnk: I think that code should not contain workarounds to make specs work. But yes, that would work.

spec/features/admin/users_spec.rb
mkllnk: Thank you Rob, I was struggling with naming it. I didn't want to write "creating a user creates a user". But "creating a user confirms successful creation" is better than "creating a user works". I added a commit for that.

mkllnk:

sstead:
**Testing notes**
- All good!
https://docs.google.com/document/d/15F7vcNeN9nb16bU6sNx5K2Fn2QFzCOJVl_0T0h-bFpE/edit#

PR #2403, opened 2018-06-22
Hide postcode - not necessary as passed from billing address and wrec…

luisramos0:
…ks mobile UX

#### What? Why?

Simple fix to test for:
#2352

The word "Postcode" is too long and overlaps the card number field on small screens. We can just hide the field it as it's not necessary (see issue discussion).

#### What should we test?

Check form is visible and usable

Check Stripe payments still go through ok.

#### Release notes

Removed postcode (ZIP) field from Stripe card form so it now works on small screen mobile devices.

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/stripe_elements.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/stripe_elements.js.coffee
@@ -10,7 +10,7 @@ Darkswarm.directive "stripeElements", ($injector, StripeElements) ->
       stripe = $injector.get('stripeObject')
 
       card = stripe.elements().create 'card',
-        hidePostalCode: false
+        hidePostalCode: true
         style:
           base:
             fontFamily: "Roboto, Arial, sans-serif"


myriamboure:
@Matt-Yorkley or @lin-d-hop could you stage that so that we can test? And @Matt-Yorkley feel free to review as well so that we have a second review ;-)

mkllnk:

myriamboure:
@RachL do you have capacity to test?

RachL:
@myriamboure yes I'm on it

RachL:
I've tested on a Android device and it works fine. If anyone can test from an IOS device it could be a double check, although I don't think there should be any difference given we're hiding a field.

https://docs.google.com/document/d/18V9i5Q8-yQKLfCO8ujNOK42P1oILmPZtAmE-kRpmJlc/edit#

myriamboure:
On IOS the postcode is not seen anymore and not blocking purchase.
Althought there is still a small unfriendly UX when you type the last 4 digits of the card number, that overlaps with date fields. But it's not blocking that PR, which is ready to go then. @RachL feel free to open another issue if you think it deserve it.
screenshot from 2018-06-28 01-07-56

RachL:
@myriamboure I don't have the issue on android. Do you remember if you've zoomed in ?

In any case, I'm wondering about having to put all the info in the same field. I think dedicated fields (card number, date, CVC) like here : https://stripe.com/checkout would be better.
If you think it is a good idea, I can open an issue to split the fields. That would avoid any mobile problems in the future. (I guess)
image

PR #2405, opened 2018-06-23
added cookies policy link to footer and the cookies policy modal

### LINKED ISSUE #2401: [Cookies] Create cookie policy page
## Description
Today users are not informed about which cookies are set on their browsers and for what. With GDPR we need to inform them.

- As a: visitor/user

- I want to be able to: understand which cookies are set on my computer and for what (later one: I want to be able to customize and disable optional cookies > see other stories)

## Acceptance criteria
- Create the cookie policy page, content is here: https://docs.google.com/document/d/1sWqXeDZBerXP263w1NWSpsgCuBGCZf35368VKlgPdq0/edit#
- The cookie policy page is always applicable to all instances. 
- Must be translatable. 
-  If instance config is "Matomo" then Matomo cookie section is visible and Google Analytics cookies section is not / if instance config is "GA" then GA section displays and matomo section doesn't 
- The url in "set-by" column vary depending on instance url when entity setting up cookie is instance.
- The cookies policy page is a modal page like the login modal here and accessible through a similar URL, for example /policies/cookies
- It needs to be readable on mobile as well.

See slide 5 https://docs.google.com/presentation/d/1uMWiW6lQhqexAe_uFmjp5QNAM1zLfnIe1bF3-_Nb5xo/edit?ts=5b2be17d#slide=id.g3841ad4459_0_19

luisramos0:

### DIFF
--- a/app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
+++ b/app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
@@ -1,0 +1,89 @@
+%h2
+  {{ 'legal.cookies_policy.header' | t}}
+%p
+  {{ 'legal.cookies_policy.desc_part_1' | t}}
+%p
+  {{ 'legal.cookies_policy.desc_part_2' | t}}
+%p
+  {{ 'legal.cookies_policy.desc_part_3' | t}}
+
+%h2
+  {{ 'legal.cookies_policy.essential_cookies' | t}}
+%p
+  {{ 'legal.cookies_policy.essential_cookies_desc' | t}}
+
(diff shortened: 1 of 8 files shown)


myriamboure:
Wahou, that's a fast first bit! I like that :-) Great reactivity @luisramos0 !

myriamboure:
Ok normally we say one PR for one issue, but maybe I should have opened only one issue for the two things...

luisramos0:
I have decided to close this PR and move these 2 commits to PR #2411 
@sauloperez these 2 commits here are also there:
- Changed modals max height from 120% to 100%.
and
- Added new cookies policy page and added a link to it in the footer 
The other PR 2411 has 4 more commits:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits

PR #2408, opened 2018-06-26
Remove impossible route

mkllnk:
#### What? Why?

I randomly came across this impossible entry in our routes: `get "/#/login"`

Hash fragments are not part of the http request and not sent to the
server. So the routes will never see a hash fragment which means that
this route is never used.

#### What should we test?

Logging in from a few different screens. Make sure that nothing changed.

#### Release notes

Removed unused code.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -8,7 +8,6 @@ Openfoodnetwork::Application.routes.draw do
   get "/t/products/:id", to: redirect("/")
   get "/about_us", to: redirect(ContentConfig.footer_about_url)
 
-  get "/#/login", to: "home#index", as: :spree_login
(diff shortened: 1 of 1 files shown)


mkllnk:

myriamboure:
@RachL Logging in I would interpret like click on "login" so login as registered user either:
- from login button
- from uber banner
- from checkout when you checkout as guest but put a non guest email
- from checkout when you are invited to choose to checkout as guest or non guest
But I don't understand the issue so maybe @mkllnk can be more precise about what we should test and why ;-)

mkllnk:
The issue was just dead code that didn't do anything. I removed it and that shouldn't make any difference to the user. It just helps developers understanding the code.

PR #2409, opened 2018-06-27
2013 Remove dead code for Spree upgrade

### LINKED ISSUE #2013: [Spree Upgrade] Refactor all uses of spree_products.count_on_hand

mkllnk:
#### What? Why?

Simplifies #2013. But there is more to do for that issue.

Working on the Spree upgrade, @sauloperez identified the code that is using the soon obsolete column `spree_products.count_on_hand`. Investigating refactoring this code, I found that it can be simply removed. :fire: :fire: :fire: 

#### What should we test?

- Check the list of shops to see that all enterprises are shown correctly as open or closed.

#### Release notes

Removed old, unused code.

The removed code was using `spree_products.count_on_hand` which will be removed in Spree.

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -129,12 +129,6 @@ class Enterprise < ActiveRecord::Base
     joins('LEFT OUTER JOIN exchange_variants ON (exchange_variants.exchange_id = exchanges.id)').
     joins('LEFT OUTER JOIN spree_variants ON (spree_variants.id = exchange_variants.variant_id)')
 
-  scope :active_distributors, lambda {
-    with_distributed_products_outer.with_order_cycles_as_distributor_outer.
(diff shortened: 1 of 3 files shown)


mkllnk:
@sauloperez The referenced serialiser doesn't access `Enterprise.active_distributors`, it accesses the method of injection data: https://github.com/openfoodfoundation/openfoodnetwork/blob/0fd57ee7bd2deb89602bf0fc98ea3f5762f58064/lib/open_food_network/enterprise_injection_data.rb#L3

Regarding your other find, I think that directive is used in several places, for example: https://github.com/openfoodfoundation/openfoodnetwork/blob/0fd57ee7bd2deb89602bf0fc98ea3f5762f58064/app/views/shops/_fat.html.haml#L33

sauloperez:
cool! I figured there was some angular involved I wasn't familiar with. Staging on katuma then.

sauloperez:

RachL:
Seems good to me ! Here are my testing notes : 
https://docs.google.com/document/d/1xHLtE0Lep34h5osqdoHxpbGhciN19QmUOKRc1Cr39zk/edit#

Issue #2407, opened 2018-06-26
[Cookies] Build coookie banner and add a toggle to power it on-off


myriamboure:
## Description
Today users are not informed about the cookies set up on their computer and don't give their consent. With GDPR we need to inform them and request their consent. The content of the cookie policy page and set up options is still pending as we need some more info to decide but we can move forward with implementing the consent banner (it can remain toggled off until the feature is ready to be released).

As a: visitor/user

I want to be able to: be notified that cookies are used and be requested my consent, and check cookie policy if I want to understand.

## Acceptance criteria

- A banner is displayed on desktop and mobile with following text, link to cookie policy, and button to "accept cookies"
![screenshot from 2018-06-26 12-50-46](https://user-images.githubusercontent.com/8135376/41907135-3e521548-7940-11e8-95f1-29cc83b1a694.png)
![screenshot from 2018-06-26 12-50-32](https://user-images.githubusercontent.com/8135376/41907139-4162e56e-7940-11e8-81f8-3d383c8f6c10.png)

- This banner is by default toggled off but can be toggled on from admin > configuration (See slide 10 https://docs.google.com/presentation/d/1uMWiW6lQhqexAe_uFmjp5QNAM1zLfnIe1bF3-_Nb5xo/edit?ts=5b2be17d#slide=id.g3c59fe849e_0_6). Updated: Even if the banner is toggled off the cookies policy link on the footer of the website should still be shown.

- Technically all essential cookies are set up as soon as user either click on "accept cookies" or start navigating on the website (so if clicks on any other menu).

NOTE: Analytics cookies are out of scope for now. Missing informations to decide on them still.

myriamboure:
I put a little bit more text @luisramos0 so that users are incited to click on "accept cookies" and know that if they don't click on "accept" but navigate they do consent. All clear? @RachL does it looks good to you as well?

RachL:
@myriamboure perfect !

luisramos0:
This is what I have so far:
<img width="1075" alt="screen shot 2018-06-27 at 17 26 38" src="https://user-images.githubusercontent.com/1640378/41986849-5116841c-7a2f-11e8-86d9-911331139b3f.png">
and
<img width="266" alt="screen shot 2018-06-27 at 17 34 36" src="https://user-images.githubusercontent.com/1640378/41987210-6459830c-7a30-11e8-9124-eb759fd8f885.png">

myriamboure:
[...]

Issue #2406, opened 2018-06-26
[Cookies] Add privacy policy URL and display in footer


myriamboure:
## Description
Today users are not informed about the privacy policy whose cookie policy is a part of. With GDPR we need to inform them. So we need to allow access to privacy policy as well as cookie policy.

As a: visitor/user

I want to be able to: understand the privacy policy of OFN more generally, into which cookie policy happens.

## Acceptance criteria

- Add in configuration panel the possibility to link a “privacy policy” PDF setup at instance level (optional)
- Add in the footer link to access the privacy policy (if not url set up for the instance it should not appear)

See slide 4 and 10 https://docs.google.com/presentation/d/1uMWiW6lQhqexAe_uFmjp5QNAM1zLfnIe1bF3-_Nb5xo/edit?ts=5b2be17d#slide=id.g3c59fe849e_0_6

luisramos0:
Hi @myriamboure 
I was looking at the configuration page, theres a entry for "TOS required", section "ENTERPRISE TERMS OF SERVICE". Shall we merge this one entry with cookies toggle and privacy URL and call the section "Legal Settings"? It would simplify the screen a little bit.

I suggest:
- Rename Section "ENTERPRISE TERMS OF SERVICE" to "LEGAL SETTINGS"
- Add in this section to the existing "ENTERPRISES MUST ACCEPT TERMS OF SERVICE", "Cookies Toogle" and "Private URL" field.

luisramos0:
I've just found the "terms of service URL" field in the content configuration page...
We should either move the "privacy URL" to the content configuration page or move the "terms of service URL" to the global configuration page. You guys choose? :-)

myriamboure:
I like what you are proposing @luisramos0 ! Let's go for it put it under "legal settings"
Good idea to bring the terms of services URL to legal settings as well I would say (IMO), I propose to put that under "general configuration" page for now, maybe if it starts to get busier that we can move some part in new sub menu but we start to have quite a lot already so let's keep it in general settings for now.

luisramos0:
This is the current legal settings screen:
<img width="384" alt="screen shot 2018-06-29 at 14 25 56" src="https://user-images.githubusercontent.com/1640378/42094665-64ce7ca6-7ba8-11e8-8c69-663955a4421a.png">

luisramos0:
[...]

Issue #2402, opened 2018-06-22
[cookies] Add footer link to access the cookie policy page


myriamboure:
## Description
Today users are not informed about which cookies are set on their browsers and for what. With GDPR we need to inform them. So now that the cookie policy page is created we need to give users access to it.

- As a: visitor/user

- I want to be able to: understand which cookies are set on my computer and for what (later one: I want to be able to customize and disable optional cookies > see other stories)

## Acceptance criteria
- In footer on all app pages I can access the cookie policy via a dedicated link.
- When click on link, opens page in same tab. User can just click on any menu again to access other sections.

See slide 4 https://docs.google.com/presentation/d/1uMWiW6lQhqexAe_uFmjp5QNAM1zLfnIe1bF3-_Nb5xo/edit?ts=5b2be17d#slide=id.g3c59fe849e_0_6

luisramos0:
Marked as blocked, it's dev complete but we will wait for next week (gathering) to make a decision about analytics.

PR #2411, opened 2018-06-27
New Cookies banner

luisramos0:
#### What? Why?

Closes:
- #2401 and #2402 (cookies policy page and link in footer)
- #2407 (cookies banner)
- #2406 (privacy URL)

Added footer link and cookies policy page.
New cookies banner toggle and new cookies banner with accept cookies button working.
Privacy URL BO setting and respective link on footer also added as part of this PR: issue 2406.
New darkswarm folder cookie_policy with directive, controller, service, template and styling.

##### Cookies Policy Modal
The footer link to the cookies policy page should always open the modal.
The modal should describe cookies and show the correct instance domain name.
On clicking the X or pressing Escape key in the cookies policy modal, the modal should close and user should be back to the page she was.
Also,
- URL http://<instance_domain>/#/policies/cookies should load homepage with open cookies policy modal, if modal closed here, the user should be left on the homepage
- URL http://<instance_domain>/shops/#/policies/cookies should load shops page with open cookies policy modal, on modal close, if modal closed here, user should be left on the shops page

Same behaviour on mobile, all cookie policy parts should be readable.

##### Cookies Banner
New cookie banner with "accept cookies" button and link to cookie policy page.
New footer entry with link  to privacy policy URL.
New legal settings in BO with cookie banner toggle and privacy URL.

This new banner also needs to be tested on mobile.

##### Modals

This change affects the max height of all modals. The registration process should be tested while changing the browser window size to see if it looks okay with different sizes.
#### Release notes

New cookies policy link in footer and new cookies policy modal with details about the cookies used on the website.
New cookie banner with "accept cookies" button and link to cookie policy page.
New footer entry with link  to privacy policy URL.
New legal settings in BO with cookie banner toggle and privacy URL.


app/assets/stylesheets/darkswarm/modals.css.scss
```diff
@@ -28,7 +28,7 @@ dialog, .reveal-modal {
   // Medium and up - when modal IS NOT full screen
   @media only screen and (min-width: 641px) {
     top: 10%;
-    max-height: 120%;
+    max-height: 100%;
```
mkllnk: Can you explain why you are doing this change? And does it open #1766 fixed in #2055 again?

It's good practice to explain the why in the commit message.

app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
```diff
@@ -0,0 +1,89 @@
+%h2
```
mkllnk: When you introduce new user interfaces, it's helpful to add some screenshots to your pull request.

config/locales/en.yml
```diff
@@ -1270,8 +1270,33 @@ en:
   footer_legal_tos: "Terms and conditions"
   footer_legal_visit: "Find us on"
   footer_legal_text_html: "Open Food Network is a free and open source software platform. Our content is licensed with %{content_license} and our code with %{code_license}."
-
+  footer_legal_data_text: "We take good care of your data." 
+  footer_legal_data_call:  "See our"
```
mkllnk: I see why you split the text, but it makes it hard for translators to get the context. Depending on the grammar of the language, the link may also not be at the end of the sentence. I would do it like `footer_legal_text_html`:

```yml
footer_legal_data_statement_html: "We take good care of your data. See our %{cookies_policy}."
```

mkllnk: It would also be nice to refactor these translations to use lazy lookup. I know, you went with the existing structure and that's fine. But it's even better when we can do a bit of refactoring on the way and improve the code that touching the new things we want to build.

config/locales/en.yml
```diff
@@ -1270,8 +1270,33 @@ en:
...
+      cookie_stripe_desc: "Data collected by our payment processor Stripe for fraud detection https://stripe.com/cookies-policy/legal. Not all shops use Stripe as a payment method but it is a good practice to prevent fraud to apply it to all pages. Stripe probably build a picture of which of our pages usually interact with their API and then flag anything unusual. So setting the Stripe cookie has a broader function than simply the provision of a payment method to a user. Removing it could affect the security of the service itself. You can learn more about Stripe and read its privacy policy at https://stripe.com/privacy."
+      disabling_cookies_header: "Warning on disabling cookies"
+      disabling_cookies_desc: "As a user you can always allow, block or delete Open Food Network’s or any other website cookies whenever you want to through your browser’s setting control. Each browser has a different operative. Here are the links:"
+      disabling_cookies_firefox_link: "http://support.mozilla.org/es/kb/habilitar-y-deshabilitar-cookies-que-los-sitios-we"
```
mkllnk: These are Spanish links in an English translation file?

app/views/shared/_footer.html.haml
```diff
@@ -143,7 +143,11 @@
           %div
             = t :footer_legal_data_text
             = t :footer_legal_data_call
-            %a{'cookies-policy-modal'=> true, 'cookies-banner'=> @show_cookies_banner}
+            - if "#{Spree::Config.privacy_policy_url}" != ""
```
mkllnk: Why not `- if Spree::Config.privacy_policy_url.present?`?

app/views/shared/_footer.html.haml
```diff
@@ -143,7 +143,11 @@
...
+              %a{href: "#{Spree::Config.privacy_policy_url}"}
```
mkllnk: Same here: `%a{href: Spree::Config.privacy_policy_url}` should work.

app/views/shared/_footer.html.haml
```diff
@@ -143,7 +143,11 @@
...
+                = t :footer_legal_data_privacy_policy
+              = t :footer_legal_data_call_and
+            %a{'cookies-policy-modal'=> true, 'cookies-banner'=> @show_cookies_banner && ("true" == "#{Spree::Config.cookies_consent_banner_toggle}")}
```
mkllnk: Minor style note: missing spaces before `=>`.

app/overrides/spree/admin/general_settings/edit/legal_settings.html.haml.deface
```diff
@@ -0,0 +1,13 @@
...
+%fieldset.legal.no-border-bottom
+  %legend{:align => "center"}= t(:legal_settings)
+  .field
+    = preference_field_tag(:enterprises_require_tos, Spree::Config[:enterprises_require_tos], :type => Spree::Config.preference_type(:enterprises_require_tos))
+    = label_tag(:enterprises_require_tos, t(:enterprises_require_tos)) + tag(:br)  
+  .field
+    = preference_field_tag(:cookies_consent_banner_toggle, Spree::Config[:cookies_consent_banner_toggle], :type => Spree::Config.preference_type(:cookies_consent_banner_toggle))
+    = label_tag(:cookies_consent_banner_toggle, t(:cookies_consent_banner_toggle)) + tag(:br)
+  .field
+    = label_tag(:privacy_policy_url, t(':privacy_policy_url')) + tag(:br)
```
mkllnk: Copy and paste error in `t(':privacy_policy_url')`, additional `:`.

app/assets/javascripts/darkswarm/cookies_banner/cookies_banner.html.haml
```diff
@@ -0,0 +1,15 @@
+.row
+  .large-9.columns
+    %p
+      {{ 'legal.cookies_banner.cookies_usage' | t}}
```
mkllnk: These translations are actually under `legal.cookies_policy`. It would be better to use the lazy lookup and move the translations to the right place.

luisramos0:
Please see screenshots [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/2407#issuecomment-400744870), [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/2406#issuecomment-406655954) and [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/2401#issuecomment-406655044).

app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
luisramos0: Yes, I have now added links to the screenshots as a comment to this PR.

config/locales/en.yml
luisramos0: I did it this way because, on the same sentence, there is a conditional. We have to display privacy URL if it is configured in the BO.
See footer_legal_data_privacy_policy  here: https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits/1e3b67d5da3ed2aaf7b80b3910aa5418c9961d24
With this in mind, I am not sure I can do what you suggest (in the requirements the privacy policy URL comes before the cookies policy in the sentence). Do you understand?

luisramos0: In terms of lazy lookup, do you mean to move all footer translations (the ones from line 1223 on en.yml to line 1254) into the "shared" subfolder on en.yml (line 1085)? I think that's what you mean. Please confirm I'll do it.

config/locales/en.yml
luisramos0: Copied from spec. I have changes the urls to english ones now.

app/views/shared/_footer.html.haml
luisramos0: Yes, fixed.

app/views/shared/_footer.html.haml
luisramos0: Fixed.

app/assets/javascripts/darkswarm/cookies_banner/cookies_banner.html.haml
luisramos0: These translations are not in legal.cookies_policy, they are in legal.cookies_banner, which is different. Or I may be missing something.
These are javascript translations in angular templates where I don't think we can use lazy lookup. Or I may be missing something... (if this is correct, we can't use lazy lookup in angular templates, I can add this detail to the i18n page).

app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
luisramos0: yes, that makes sense. I have added links to all the screenshots below as a comment to this PR.

app/assets/stylesheets/darkswarm/modals.css.scss
luisramos0: oh, sorry, I forgot to explain this one!

I changed it to 100% so that the modal is never outside the viewport. Allowing a height over 100% makes content go outside the viewport without scrollbars...

Thanks for the reference to #2055! I have just tested the registration process with this new value and it looks good according to my tests.

@HugsDaniel does this make sense to you? I see your comment here: https://github.com/openfoodfoundation/openfoodnetwork/pull/2055#issuecomment-366676755
Why would you ever let max-height be bigger than 100%? That means you remove scroll bars and the content will not show. I think 100% is always the correct value, so that if the content doesn't fit, you get the scroll bars. That's how all modals should work. Is there an example where you think this 100% value will not work?
Thanks!

app/views/shared/_footer.html.haml
luisramos0: Yes, fixed.

luisramos0:
Hey @mkllnk , thanks a lot for your review! I have fixed or commented on every point you mentioned. Please have a look. Thanks!

config/locales/en.yml
Matt-Yorkley: Maybe you could have both sentences in full as separate keys, and show either one based on the conditional instead of a split sentence.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
 
   prepend_before_filter :restrict_iframes
   before_filter :set_cache_headers # Issue #1213, prevent cart emptying via cache when using back button
+  before_filter :set_show_cookie_banner
```
mkllnk: Can we move this into a helper instead of adding code to the application controller? I don't think you need to persist the state and introduce a new variable.

app/views/shared/_footer.html.haml
```diff
@@ -143,8 +143,11 @@
           %div
             = t :footer_legal_data_text
             = t :footer_legal_data_call
-            %a{'cookies-policy-modal'=> true}
+            %a{'cookies-policy-modal'=> true, 'cookies-banner'=> @show_cookies_banner}
               = t :footer_legal_data_cookies_policy
+        %p.text-small
+          %a{ng: { controller:'CookiesBannerCtrl', click: 'deleteCookiesConsent()' }}
+            = 'delete consent cookie (for manual testing only)'
```
mkllnk: This doesn't look like something that should be in production code. Did you forget to remove it?

app/overrides/spree/admin/general_settings/edit/legal_settings.html.haml.deface
```diff
@@ -0,0 +1,13 @@
+/ insert_after "fieldset.security"
+
+%fieldset.legal.no-border-bottom
+  %legend{:align => "center"}= t(:legal_settings)
```
mkllnk: Have you tried to use lazy lookup in deface files? It would be good to know if that works. I would just do `t('.legal_settings')` and then check the error message to see where it expects the key to be. We can then move all these strings into the right scope.

app/assets/javascripts/darkswarm/cookies_banner/cookies_banner.html.haml
mkllnk: You are right. I just checked and this is correct. And since this is an Angular template, it needs to be translated through the `t` filter. Somehow my computer is not re-compiling the locale for Javascript and I get the missing translations. But that's not your fault. ;-)

mkllnk:
I found that the layout is not ideal for every window size. It is possible that the accept button is not visible and I can't scroll.
screenshot from 2018-07-23 15-57-58

config/locales/en.yml
luisramos0: thanks for your feedback!
Agreed, I have made it 2 versions with the full sentence. Please have a look at the new version.

app/assets/javascripts/darkswarm/cookies_banner/cookies_banner.html.haml
luisramos0: 👍

app/views/shared/_footer.html.haml
```diff
@@ -139,6 +139,12 @@
           %a{href:"https://github.com/openfoodfoundation/openfoodnetwork", target: "_blank"} GitHub
         %p.text-small
           = t :footer_legal_text_html, {content_license: link_to('CC BY-SA 3.0', 'https://creativecommons.org/licenses/by-sa/3.0/'), code_license: link_to('AGPL 3', 'https://tldrlegal.com/license/gnu-affero-general-public-license-v3-(agpl-3.0)')}
+        %p.text-small
+          %div
+            = t :footer_legal_data_text
```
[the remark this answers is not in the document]
luisramos0: 👍 I have moved all translations in footer to shared/footer. please have a look at the new version. It feels good to clean up!

app/overrides/spree/admin/general_settings/edit/legal_settings.html.haml.deface
luisramos0: yes, it works. Thanks for the hint!
I have added it to "add translations' here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)#development

I am now using lazy loading for these translations of the legal settings 👍

app/views/shared/_footer.html.haml
```diff
@@ -139,6 +139,12 @@
...
+            = t :footer_legal_data_call
+            %a{'cookies-policy-modal'=> true}
```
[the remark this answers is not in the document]
luisramos0: 👍 fixed in the new version.

app/views/shared/_footer.html.haml
luisramos0: (I am pretty sure I have replied to this comment already somewhere else, sorry if it's duplicated)
I left it on purpose so that this can be manually tested. Otherwise I dont know how testers will be able to accept cookies more than once to validate everything is fine. Do you know how we can do that? Thanks!

luisramos0: I didn't forget about this. I just thought we would need this for manual testing. So that manual testers could delete the consent_cookie. BUT I just realised it's extremelly easy to delete a cookie in chromes dev tools on the "Application" tab. I have deleted this code now.

app/assets/stylesheets/darkswarm/modals.css.scss
luisramos0: I see. Well, I have retested the registration process and it looks good with max-height 100%: it makes the modal scroll, not the screen. I think this is fine. But lets wait for someone else's feedback.
Thanks!

luisramos0:
Hi @mkllnk thanks again for you re-review :-D

I have fixed the layout issue by improving the media queries.

app/controllers/application_controller.rb
mkllnk: Sorry for not being clear. I meant to remove all your additions from the ApplicationController and add a helper like this:
```ruby
def show_cookie_banner?
  request.cookies['cookies_consent'].nil?
end
```

I find that a lot simpler than callbacks and an instance variable. But I'm also not the biggest rails pro. Are there other opinions, @Matt-Yorkley @sauloperez?

app/views/shared/_footer.html.haml
mkllnk: I also find it extremely useful to open a private window, test something and then close it again. All cookies of the private window are gone after closing it. You can even do that on mobile.

app/controllers/api/cookies_consent_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  class CookiesConsentController < BaseController
+    respond_to :json
+
+    def initialize
+      super
+      @cookie_name = 'cookies_consent'
+    end
+
+    def show
+      render json: { cookie_consent: !cookies[@cookie_name].nil? }
```
sauloperez: the [key?](http://ruby-doc.org/core-2.1.5/Hash.html#method-i-key-3F) method might be a bit easier to understand than `nil?` plus a negation.

app/controllers/api/cookies_consent_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Api
+  class CookiesConsentController < BaseController
+    respond_to :json
+
+    def initialize
+      super
+      @cookie_name = 'cookies_consent'
+    end
```
sauloperez: I wouldn't override the constructor to avoid coupling with Rails internals. Keep in mind it belongs to the framework and it might change in future versions and receive any kind of low-level optimizations that could break this controller.

Also, I think what you're trying to achieve is better suited as a constant. Like

```ruby
COOKIE_NAME = 'cookies_consent'.freeze
```

app/controllers/application_controller.rb
sauloperez: Personally, I'm not a fan of callbacks either and I dislike having ivars in views. Both lead to a global state that is very hard to manage in the long run.

I would go with @mkllnk suggestion and call that `show_cookie_banner?` straight from the view.

app/controllers/api/cookies_consent_controller.rb
luisramos0: Yes, thanks. Done. #learning_ruby

app/controllers/api/cookies_consent_controller.rb
luisramos0: Yes, thanks. Done. #learning_ruby

app/controllers/application_controller.rb
luisramos0: awesome, thanks for the advise guys. I have removed everything from applicationcontroller and also removed a few things from api/cookies_consent_controller and added the cookies logic to the new and beautiful cookies_consent_helper.

It feels this helper is ok like this BUT it's quite obvious this should not be a helper, it should be a Service. Where would our service layer live? In this specific case, for example, where would a service like CookiesConsentService be in terms of folders?
Thanks!

app/controllers/api/cookies_consent_controller.rb
sauloperez: Of course! and Ruby has plenty of methods like this one in its standard library.

luisramos0:
So... in conversation with Pau, I thought that this would be a wonderful opportunity to introduce the Rails Engines strategy to modularize OFN (see [here](https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377)).

In the context of this PR, that means we would see the birth of the first OFN engine (or OFN module) called Web (see discourse above for names of the modules) that would contain, for now, some of the cookies logic required for this feature.
The current branch includes the new Rails Engine Web. I can require this new Engine but I am not able to use it in the controller on the main app.
If you look [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits/1276a99d91a07730a8b2cb32b4ee0bc0d4baa0cb#diff-ce74c1e0473b18f16e942a702b449164), I am currently using the main CookiesConsentHelper.
I can require the new engine with '''require 'ofn_web'''' but I can't make the include work '''include Web::CookiesConsentHelper'''. It throws an error:
ActionController::RoutingError (uninitialized constant Web::CookiesConsentHelper):
  app/controllers/api/cookies_consent_controller.rb:7:in `<class:CookiesConsentController>'

@mkllnk @Matt-Yorkley @sauloperez @enricostano @HugsDaniel 
I think this would be super awesome progress in the right direction. If this works, I could move most of the code of this PR to the new module already, including views, controllers and javascript! But I need someone's help to make it work :-D

Thanks in advance for your help!

Gemfile
```diff
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 ruby "2.1.5"
 git_source(:github) { |repo_name| "https://github.com/#{repo_name}.git" }
 
+gem 'ofn_web', path: 'apps/web'
```
sauloperez: I would call it engines to keep the convention. I guess it'll make it easier for new Rails devs to join.

apps/web/ofn_web.gemspec
```diff
@@ -0,0 +1,17 @@
...
+  s.authors     = ["ofn"]
+  s.email       = ["hello@openfoodnetwork.org"]
+  s.homepage    = "http://www.openfoodnetwork.org/"
+  s.summary     = "OFN frontoffice web app"
+  s.description = "Consumer facing OFN frontoffoce web application"
+
+  s.files = Dir["{app,config,lib}/**/*"] + ["README.rdoc"]
+
+  s.add_dependency "rails", "~> 3.2.22.5"
+end
```
sauloperez: the gem should be named after its folder. Besides, I don't see the point about the redundancy of the `ofn_` prefix. This is OFN's repo so I don't there's need for it; it can be annoying to have a bunch of engines all of them with this prefix.

app/assets/stylesheets/darkswarm/cookies_banner.css.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.cookies-banner button {
+  background-color: green;
+}
+
+.cookies-banner p {
+  color: white;
+  font-size: 0.75rem;
+}
```
sauloperez: Given that we use scss, I'd rahter nest these last two rules in the `.cookies-banner` block.

app/controllers/api/cookies_consent_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+    include CookiesConsentHelper
+    # include Web::CookiesConsentHelper
+
+    def show
+      render json: { cookie_consent: CookiesConsent.new(cookies, request.host).exists? }
+    end
+
+    def create
+      CookiesConsent.new(cookies, request.host).set
+      show
```
sauloperez: I'm not sure the small DRY is worth here. We are creating a new instance for the call to `render`. Won't this be troublesome if the class grows in the future and this goes unnoticed?

app/controllers/api/cookies_consent_controller.rb
```diff
@@ -0,0 +1,23 @@
+# require 'ofn_web'
+
+module Api
+  class CookiesConsentController < BaseController
```
sauloperez: IMO this belongs to the engine and therefore it should in its folder. Also, if we do that, this controller should be namespaced as part of the engine under the `Web` module.

apps/web/Gemfile
```diff
@@ -0,0 +1 @@
+source "https://rubygems.org"
```
sauloperez: Let's make Rubocop ignore this file, please.

app/services/cookies_consent.rb
```diff
@@ -0,0 +1,28 @@
+class CookiesConsent
```
sauloperez: IMO this should also be inside the engine and under its namespace. Then, it should be available  elsewhere as `Web::CookiesConsent`

sauloperez:
Let's not forget tests

Gemfile
luisramos0: done 👍

apps/web/ofn_web.gemspec
luisramos0: done 👍

lib/ofn_web.rb
```diff
@@ -0,0 +1 @@
+require 'web'
```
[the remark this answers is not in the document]
luisramos0: This was me trying to get the engines code included in the main app. With no success. I will delete this file when we have the engine being used correctly in the main app.

app/controllers/application_controller.rb
mkllnk: There is currently the choice between `app/services` and `lib`. We haven't agreed yet which approach is better. And while you are right that application logic should go into a service object, I find it okay to leave a one-liner in the helpers. You need the helper anyway to call the service object and pass on the `request` variable.

mkllnk:
Sorry @luisramos0, you really confused me now. I like the discussion about modularisation, but I would prefer to see a pull request that is just refactoring to demonstrates that. The pull requests will be smaller, easier to review and we can form an opinion on each concern, not on both together. This pull request was almost test ready and now it's in a big dev process and you say you need help. Can we separate concerns? I would like to see your cookie work in the old structure and then see a pull request demonstrating the module structure where it makes most sense.

Btw, why did you start calling things "cookys"?

app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
```diff
@@ -0,0 +1,177 @@
...
+    %a{ href: "{{ 'legal.cookies_policy.disabling_cookies_ie_link' | t}}", target: "_blank" }
+      Internet Explorer
+  %li
+    %a{ href: "{{ 'legal.cookies_policy.disabling_cookies_safari_link' | t}}", target: "_blank" }
+      Safari
+%p
+  {{ 'legal.cookies_policy.disabling_cookies_note' | t}}
+
+%a.close-reveal-modal{"ng-click" => "$close()"}
+  %i.ofn-i_009-close
```
sauloperez: sorry I forgot, if we continue with the engine, all assets should also be in its folder.

app/assets/stylesheets/darkswarm/cookies_banner.css.scss
luisramos0: thanks. done.

app/controllers/api/cookies_consent_controller.rb
luisramos0: Agreed, I've moved all engine stuff to a different branch, Ill do that there. Thanks.

app/controllers/api/cookies_consent_controller.rb
luisramos0: sorry, I dont think I understand your point @sauloperez

app/services/cookies_consent.rb
luisramos0: agreed.

config/routes.rb
```diff
@@ -214,6 +214,10 @@
 
     resources :customers, only: [:index, :update]
 
+    scope '/cookies' do
+      resource :consent, only: [:show, :create, :destroy], :controller => "CookiesConsent"
+    end
```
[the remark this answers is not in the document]
luisramos0: yes, agreed. Ill do that in the engine branch!

luisramos0:
@sauloperez yes, I just have to clean up the auto tests I have and add then to the branch (I got carried away with the rails engine stuff and left that for later...).

@mkllnk I am sorry! you are absolutely right. I just got really excited about the rails engines thing and mixed everything in the same pot... sorry about that!
I follow your advise, Ive reverted this branch to the original structure and I'll create a new PR/branch for the rails engine stuff (tomorrow).
("cookys" was everything in the old structure and "cookies" was the new rails engine stuff)

Regarding this branch @mkllnk , I think all that is pending your review is what @sauloperez guided me to do with the CookiesConsent.

Thanks!
Luís

app/views/shared/_footer.html.haml
```diff
@@ -131,21 +130,22 @@
...
           %div
-            = t :footer_legal_data_text
-            = t :footer_legal_data_call
-            %a{'cookies-policy-modal'=> true, 'cookies-banner'=> !CookiesConsent.new(cookies, request.host).exists? }
-              = t :footer_legal_data_cookies_policy
-
+            - cookies_policy_link = link_to( t( '.footer_data_cookies_policy' ), '', 'cookies-policy-modal' => true, 'cookies-banner' => !CookiesConsent.new(cookies, request.host).exists? && Spree::Config.cookies_consent_banner_toggle)
+            - privacy_policy_link = link_to( t( '.footer_data_privacy_policy' ), Spree::Config.privacy_policy_url, :target => '_blank' )
+            - if Spree::Config.privacy_policy_url.present?
+              = t '.footer_data_text_with_privacy_policy_html', {cookies_policy: raw( "#{cookies_policy_link}" ), privacy_policy: raw( "#{privacy_policy_link}" )}
```
mkllnk: `html_safe` seems to be the better option instead of `raw` here:
```ruby
cookies_policy_link.html_safe
```

config/locales/en.yml
```diff
@@ -1082,6 +1082,33 @@ en:
     register_call:
       selling_on_ofn: "Interested in getting on the Open Food Network?"
       register: "Register here"
+    footer:
+      footer_global_headline: "OFN Global"
```
mkllnk: You could have removed the `footer_` prefix from all these variables, but that doesn't really matter.

app/controllers/api/cookies_consent_controller.rb
sauloperez: Sorry, I meant the DRY we get by calling `show` instead of directly calling `render`. It does create a new instance of `CookiesConsent`.

config/locales/en.yml
sauloperez: ouch, that's right

app/assets/stylesheets/darkswarm/modals.css.scss
myriamboure: We'll see when we test but my experience of cookies banner is that you can scroll the page behind it but it doesn't scroll the modal. But for the registration process on modal you will definitely need to scroll as some modal pages have too many info to fit on small mobile screens... anyway I don't get the technical side so happy to see when we test what it looks like :-)

config/locales/en.yml
luisramos0: I thought about that BUT I didnt do it because I am not sure transifex will pick up the translations if we change the translation key :-)

app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
luisramos0: thanks!

app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml
luisramos0: 👍

app/views/shared/_footer.html.haml
luisramos0: interesting, apparently raw = to_s.html_safe.
Ive changed it to html_safe. Much better! Thanks.

app/controllers/api/cookies_consent_controller.rb
luisramos0: oh, I didnt see it! now I see it.
I fixed it now.

config/locales/en.yml
mkllnk: Transifex should be smart enough.

mkllnk:

RachL:
@luisramos0 @myriamboure Guys, in order to properly test this I need some feedback:

1. Issue 2401: Where can I add the tracker code ? The analytics tracker menu in configurations only mentions an ID.
2. Issue 2401: It is mentioned it has to work on mobile. I'm sorry to see this only now but this is not working if we say it always opens in a modal. The good pratice is : modal on desktop, new tab on mobile. Here when I've finished scrolling down, I can't get out of the modal, I have to scroll up again:
image
So if you are ok with this I suggest we leave mobile aside for the first version and make some other issues for mobile UX. This could be the opportunity to fix all our modal behaviors on mobile at once. What do you think ?
3. Content of issue 2402 is already in 2401, am I correct? or am I missing something to test?
4. Issue 2406: how does the field for the privacy policy work ? Do I need a dev to upload the file on the server? Just a comment on this issue as well: it's a good example where we could have updated the issue description with the comments decision ;)

myriamboure:
@RachL do you have some testing notes to share? 
About 2402 it was about the footer mention "we take good care of your data" thing, so it was slightly different and complementary to 2401 which was about just building the page ;-) What you need to test is that if there is a privacy policy url (you can put a PDF on a drive and put the link for now it should work, later one we can put the file on the server and put the link to replace it) the link is displayed in the footer message, else the privacy policy is not mentioned.
Have you tested also the correct behavior for each configuration set up possible? Make sure as well that the content of the policy page adapt depending on which set up is chosen.
I'm happy to merge without mobile UX but open a new issue and include it in the epic to make sure we solve it before considering the feature as done. And great if we can take that opportunity to agree and fix other modals behavior on mobile I'm sure @kirstenalarsen will be happy :-)

RachL:
@myriamboure no I don't as I'm not finished testing. I didn't managed to find where to set up the banner. Thanks for the other info I will test them tomorrow!

mkllnk:
This pull request touches 23 files which is a lot to review at once. A separation would have made it easier. The downside is that some of these issues depend on each other. So Luis would have needed to wait for one pull request to be merged before opening the next one. I understand Luis' decision very well, because multiple pull requests can slow the delivery down. On the other hand, they make it easier for everybody and therefore quicker to merge.

So it's a difficult decision, but I lean towards smaller pull requests. It lowers the risk of bugs and, more importantly, trains the developer to create independent code changes and independent code modules separating concerns. That means that individual parts of the code are more encapsuled which makes later changes easier. Smaller pull request will lower the maintenance cost of the code over time.

myriamboure:
@RachL it's in configuration, general setting, legal stuff if I remember well, you have a checkbox to disable/enable the banner normally.

RachL:
Alright so yesterday I had some computer cache pb, that's why I couldn't see the banner, today it has worked fine. There is one thing that is not working I think: it's the GA or Matomo part in cookie policy. Is that part of another PR ?
Also I have two questions:
1. Is it ok that the cookie policy page always stays, even if the cookie banner is off ?
2. Each time I've saved the general settings, I've got the cookie banner again even if I've accepted it previously. It seems to me a bit weird... are we really deleting cookies each time an admin changes the general settings ?
Testing notes:
https://docs.google.com/document/d/1Bz9nnK0raTLi-51O6IgrKkNT8nPInPpMexv62xov2Qc/edit#
Thanks :)

RachL:
@luisramos0 thank you for your feedback :)
For my first comment: forget the tracker code, I mixed up the other issue as well which is not staged yet :) For the mobile part, a X would do the trick, but opening a new tab on mobile would be cleaner. Is it a lot more work ? Otherwise I can make another issue and we don't have to deal with that here.
For my second comment:
A. Ok so this needs to be tested again. How long do you think is the server cache ? Awesome for automated tests :)
B. No problem I can do that. However Aus staging has another PR right now and I'm not sure I've tested it properly so I need some help. Shall we stage this PR elsewhere in the meantime ? @Matt-Yorkley is UK staging free ?

Thank you all :)

mkllnk:
Hey Luis, it's great to see you being excited about this. :-)

First of all, seeing the length of this thread and the length of you reply to everything, it would be a really good idea to split this pull request. I would suggest to close this one and open three new ones:

1. Cookies policy page and footer link.
2. Cookie banner.
3. Privacy policy.

Would that be okay with you?

I need your public ssh key to give you access to staging.

luisramos0:
Hi @mkllnk as we were speaking about this today on the spree call, I didn't realize you had made this comment to split this PR.
What we have talked about is that all these 3 points you mention depend on each other: the banner has a link to the policy page and the privacy policy link in the footer is in the same sentence as the link to the cookies policy page... this means more than 3 dependant PRs (I think it would basically be one PR for each of the 6 commits in this PR).
I think it'd be too much trouble to do it right now. So I'd leave it as it is.
I believe you accepted this decision on the call, but I am not sure. Do let me know if you think I should still separate this PR and I'll do it next week.
Luís

myriamboure:
@mkllnk I also think that at that point it doesn't make sense to resplit, it has been reviewed approved and tested already, so we would loose a lot of time. I agree with @RachL about displaying the cookie policy in a new tab as modal UX on mobile is pretty poor (lesson learn for next time), actually it's totally fine with me if it display in a new tab also on desktop and not only on mobile if that's easier... what do you think @RachL ? I guess it's harder to say that it's a modal on desktop and a new tab on mobile... am I guessing right @luisramos0 ? If it's not too much work let's do it now before we restage for testing taking into account the caching behavior and debugging the looping behavior!

luisramos0:
I submit an updated version with the following changes:
- added tests to validate the functionality ([code change here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits/a18daba60c7625b55db2f3c14d6e68dccd84348d))
- added an instruction to cookies controller to fix tests - this may as well fix @RachL issue with the banner not working ([code change here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits/637ece8172cd3f5aac7d4d4c7d63dfb5172f3470#diff-ce74c1e0473b18f16e942a702b449164R3))
- added a X to the bottom of the cookies policy page modal if on mobile to improve UX #quickfix (making it a new tab would be a lot more work) - [code change here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits/8a6e90294e154a7b54f0bf8e12aea680b34aa6e2)

Let me know if this makes sense or if I need to change something.

myriamboure:
If @mkllnk is happy with your changes let's restage and test, we will open a new issue for the new tab thing if your solution is not satisfying enough in term of UX, but that shouldn't prevent merging.

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,146 @@
+require 'spec_helper'
+
+feature "Cookies", js: true, retry: 3 do
```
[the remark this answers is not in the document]
luisramos0: I've added this tip on retry here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Testing-and-Rspec-Tips

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      end
+    end
+  end
+
+  # for some reason, this test only works if not inside a describe block
+  scenario "banner does not show after cookies are accepted" do
+    Spree::Config.cookies_consent_banner_toggle = true
+    visit root_path
+    click_button I18n.t('legal.cookies_banner.cookies_accept_button')
+    sleep 2 # waits for cookies to be set
```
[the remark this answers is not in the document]
luisramos0: I've added this tip on sleep here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Testing-and-Rspec-Tips

spec/features/consumer/cookies_spec.rb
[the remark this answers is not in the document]
luisramos0: I don't need the retry here, I have removed it.

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    scenario "opens cookies policy page" do
+      visit root_path
+      click_link "cookies policy"
+      expect_visible_cookies_policy_page
+    end
+  end
+
+  describe "related privacy policy link in the footer" do
+    scenario "not showing empty privacy policy link" do
+      Spree::Config[:privacy_policy_url] = nil
```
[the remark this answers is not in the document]
luisramos0: I've added this tip here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Testing-and-Rspec-Tips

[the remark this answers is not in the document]
luisramos0: Interesting point!
I have used delete_matched on this cookies_spec. Is this what you were thinking about?

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      visit '/#/policies/cookies'
+      expect(page).to have_content('_session_id')
+        .and have_content('127.0.0.1')
+    end
+
+    describe "with Matomo section configured" do
+      scenario "shows Matomo cookies details" do
+        Spree::Config[:cookies_policy_matomo_section] = true
+        visit '/#/policies/cookies'
+        # before { skip("test not working, BO config is not picked up") }
```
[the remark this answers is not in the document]
luisramos0: I couldn't find the problem here. I don't think it's cache of the config setting. It could be page cache but I cant find anything on my /tmp/cache folder that indicates that.

I have removed BO from here.
BO is standard nomenclature for pretty much every software solution I've worked in the past, including e-commerce. FO means customer facing and BO means management UI. In OFN's case it's FO for customers and BO for enterprises. The super admin side can be included in the BO but it's not the central part.
I agree we should agree on nomenclature, I purpose a division that includes FO and BO here: https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377

spec/features/consumer/cookies_spec.rb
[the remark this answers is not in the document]
luisramos0: I've added these sleep statements as I was trying (with all possible changes) to make these tests work inside a describe block. I can see now that the tests work well without the sleep statements. I've removed them. Thanks!

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    Spree::Config.cookies_consent_banner_toggle = true
+    visit root_path
+    click_button I18n.t('legal.cookies_banner.cookies_accept_button')
+    sleep 2 # waits for cookies to be set
+    visit root_path
+    sleep 2 # waits for policy page to pop up
+    expect_not_visible_cookies_banner
+  end
+
+  # for some reason, this test only works if not inside a describe block
```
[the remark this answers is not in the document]
luisramos0: I have managed to make the tests work inside the describe block by extracting the footer tests to a separate spec file. The main issue of having tests outside a describe block is fixed.

But I still do not understand why the tests fail if I add more tests.
I think I have spent enough time trying to fix this. I'll leave it like this for now.

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+        visit root_path       
+      end
+
+      scenario "does not show after cookies are accepted" do
+        click_button I18n.t('legal.cookies_banner.cookies_accept_button')
+        expect_not_visible_cookies_banner
+
+        visit root_path
+        expect_not_visible_cookies_banner
+        delete_spree_config_from_cache :cookies_consent_banner_toggle
```
mkllnk: Now you do this multiple times. Once in `after` should be enough.

Did you actually test if you need this? I just tried it without resetting in one spec and it seems like the preference is not persisted between the scenarios. So it looks like I have been wrong and you don't need to delete it from the cache. I'm not sure why. It has been a problem in other specs and that's why there is a reset in the spec helper: https://github.com/openfoodfoundation/openfoodnetwork/blob/32846af23019e4f00d0f7b5fdb1d9fb797641912/spec/spec_helper.rb#L108-L118

So I would say, let's try it without the delete again and see if it causes any problems. Or did you verify that already?

luisramos0: yes, there was an unnecessary duplication of the delete.

I thought I have tested it! But well, I have retested it and the configs are all reset for each test without the delete. So, in this particular case, the delete cache is not necessary. I have removed the cache_helper and the delete calls from this test.

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      visit '/#/policies/cookies'
+      expect(page).to have_content('_session_id')
+        .and have_content('127.0.0.1')
+    end
+
+    describe "with Matomo section configured" do
+      scenario "shows Matomo cookies details" do
+        Spree::Config[:cookies_policy_matomo_section] = true
+        visit '/#/policies/cookies'
+        # before { skip("test not working, settings are not picked up") }
```
mkllnk: I finally understand what's happening here. The cookies page is an Angular template pre-compiled as Javascript asset. When you change the configuration, it doesn't change the template `app/assets/javascripts/darkswarm/cookies_policy/cookies_policy.html.haml`. So Rails won't re-compile it. And even if Rails would know, it would not re-compile assets on it's own. That happens only during deploy.

With the current implementation, it's very difficult to change the cookies page. After changing the setting in the admin panel, you need to clear the cache and re-compile all assets. Even though this doesn't happen frequently, I'm sure it will cause confusion and the whole advantage of having a switch in the admin panel is gone.

It also means that testers can't test this feature in staging without the help of a developer.

I see three options:
* Leave it as is. Testing has to happen with a tester and a developer and will probably take some time. Also changing the option will always require the help of a developer. You can't activate the Matomo section just through the admin panel.
* Inject these options into the page or expose them through an API and render the cookies page dynamically.
* Create a Rails view page to display the cookies policy. If you want a modal, you can dynamically load the content of that page into the modal when it's opened.

My preferred option would be the last one:

* It keeps our assets smaller.
* It loads the page only when needed.
* It is very easy technology to understand and change.
* You have the option to load a modal or go to the page.

luisramos0: great stuff, well spotted @mkllnk !!!  I just realized how bad this templates in assets is! On the first page load we are loading ALL the assets that include ALL html templates. That is totally wrong :-)
Just to be clear: "Angular directives and templates" are a very easy technology to understand and change :-)
"Angular templates inside rails assets pipeline" is not... and I think we just need to fix that.

The problem is what we are doing with angular-rails-templates, the very thing that caused me trouble in rails engine POC: https://github.com/openfoodfoundation/openfoodnetwork/pull/2521

Actually, if you look at my "workaround" in the POC, you will see that it fixes this problem because I moved the template from assets to rails views :-)
So, I propose a little variation on your 3rd option which is to make the move from assets/templates to rails views a generic solution and making my "workaround" on the POC, a standard approach to this problem.
This means we create a new generic templates route and this templates_controller:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/438174b489361106ffa57fbb8f509a41067842c2#diff-61811c80c8207c2e6004afc35f428fa4
And this enables us on the angular side to either use the existing assets/templates OR use templates that are in rails views.
This enables us to move templates for example from app/assets/javascripts/darkswarm/cookies_banner/cookies_banner.html.haml to app/views/templates/cookies_banner.html.haml
and use it on angular services and directives like here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/438174b489361106ffa57fbb8f509a41067842c2#diff-335c2c8d75c330b90106e70912143adfL7

Is there a way we can do this (use templates in rails views on angular services/directives) without needing this templates controller? Maybe you will know how to do that.

I hope I was clear explaining the approach. Let me know.

imo, let's not move this PR forward until we fix this.

spec/features/consumer/cookies_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'spec_helper'
+
+feature "Cookies", js: true do
+  describe "banner", js: true do
```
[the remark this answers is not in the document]
luisramos0: copy paste error, thanks, I will remove.

spec/features/consumer/cookies_spec.rb
mkllnk: I like that option as well. I believe that Angular supports loading templates on demand very well.

I don't know any way to serve dynamic templates without a template controller, but we need only one controller for all templates. I don't see that as a problem. 

I'm not sure about the name "template". Angular uses that term, so good from that point of view. For Rails it's just a page without layout. What do you think about this:

- Route: `/angular-templates/:id`
- `AngularTemplatesController`
- action `show` renders the template as HTML

luisramos0:
Hey @mkllnk 
I have gone the angular-templates way with the naming you suggested.
The policy page is now working properly for the matomo/ga toggle and the respective auto test is enabled and passing.
This was done in a new commit [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits/84a00fc482b0e8c8680e9bc6f091c438f543acfa).

I rebased this PR (to resolve a conflict on routes.rb, that got me into a new scss linter (with originally 20+ issues). I have fixed all issues in another new commit [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2411/commits/dfd8e2c8e838d069aa4ec76798e42260cff7668e).

The reminder commits were just rebased and not unchanged.

Let me know if you see any other issues here.

spec/features/consumer/cookies_spec.rb
luisramos0: ok, done.

spec/features/consumer/cookies_spec.rb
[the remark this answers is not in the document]
luisramos0: done.

mkllnk:

app/views/angular_templates/cookies_policy.html.haml
```diff
@@ -0,0 +1,180 @@
+%h2
+  {{ 'legal.cookies_policy.header' | t}}
```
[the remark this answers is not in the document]
luisramos0: Good point! We could switch to rails t and lazy lookups on en.yml/angular-templates...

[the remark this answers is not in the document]
mkllnk: We could, but it could cause problems if Angular caches the template. It's probably best to leave it.

myriamboure:
Ok, so I couldn't prevent myself from testing that even if I'm on holidays ;-)
https://docs.google.com/document/d/1Bz9nnK0raTLi-51O6IgrKkNT8nPInPpMexv62xov2Qc/edit#

I saw **4 things to improve before we can consider the feature as done, but none of them for me prevents merging**, so I propose to merge and open issues for the 4 following issues:
1- **On desktop, the size of the cookie banner text is pretty small, doesn't seem to adapt to the size of the screen.** It should take more space in the grey banner. (to review and agree with UX person before moving forward).
2- More annoying, **the mobile experience has a bug: when I accept cookie policy by clicking on "accept cookies", then click on cookie policy modal page, when I close the page the banner displays again and I have to accept again.** This happens only on mobile.
3- Another issue on mobile experience: **the cookie policy modal page is both hard to scroll** (strange feeling of having to scroll twice for the scrolling to be reactive, and scrolling happens only in half screen, all that is really strange), **and the table inside is unreadable.** We need to replace the modal page by a real page that opens in a new tab to make the experience good, AND we need to review the table inside, present it differently so that it can be responsive and readable on mobile.
4- **The GA/Matomo conditional text on cookie policy page needs to be updated.** GA text has not been updated to latest version, and for Matomo we had agreed first to put it in Essential cookies. No objection to put it in Statistics cookies section, but we need to review the text, say that data are anonymized, etc. I will describe precisely what is to changed when I open the issue to improve that.

I propose to merge this, and when I'm back from holidays I will good deeper in each things to improve and open the issues so that we finish the feature. Meanwhile @luisramos0 now that the Matomo setup and this are going to be merged you can add the opt-out option from Matomo ;-)

**Can we make sure that default behavior is that checkbox "display cookie banner" is disabled before merging?** I wouldn't want the Australian to have a banner popping up that they don't want... @mkllnk @luisramos0 @Matt-Yorkley  I had a doublt while testing as by default I could see the banner first time I landed on staging. Please check before merging :pray:

luisramos0:
Awesome, thanks a lot @myriamboure 

1- yes, we can increase font size
2- I can see it on desktop as well. my bug. It's a very easy fix (Ive fixed it locally now).
3- I see the modal is not taking up all screen height. I can make it full height, it's very easy to do it, I just need to test the other modals: registration and sign up. I am not sure how a normal page will make it better in terms of ux apart from making it full screen.
re the table, I didnt want to change it too much because it was in the spec. In mobile this should be a paragraph for each cookie, isnt it? Anyway, not the best solution for sure but the table is horizontally scrollable :-/ btw, making the modal a normal page will not fix this table issue, the screen will still be the same size.
4- yes, I think I wrote the Matomo text. Sorry, I should have asked for details before.  Yes, we should move the conditional Matomo text to the essential section. Please let me know how you want it and I will change.

re default value, good point :-D
I checked the code and default is false - disabled. Maikel must have changed it on the BO after staging. Right @mkllnk ?

app/controllers/api/cookies_consent_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    include ActionController::Cookies
+    respond_to :json
+
+    def show
+      render json: { cookies_consent: cookies_consent.exists? }
+    end
+
+    def create
```
mkllnk: @luisramos0 Why do we request the server to set a cookie? Can't we just do it via Javascript? https://docs.angularjs.org/api/ngCookies/service/$cookies That could be a lot simpler and faster.

mkllnk:
> Maikel must have changed it on the BO after staging. 

I didn't, but someone must have done it. I just re-staged which is resetting the database and I can't see a banner. Should be all good!

sigmundpetersen:
So can we close off all of #2401 #2402 #2406 #2407 then? Or are there more to any of them left undone? @luisramos0 @mkllnk

luisramos0:
Hurray, live!

@sigmundpetersen That's it. They should all be done and closed (apart from the points mentioned by Myriam that will be handled on separate new issues). I'd let Myriam confirm and close those issues when she is back.

app/controllers/api/cookies_consent_controller.rb
luisramos0: An e-commerce website should only use httponly cookies that cannot be set by the browser. It's a simple security measure to avoid a million types of XSS attacks. That's why you should always go around to the server like here.
BUT in the end, I forgot to actually set it to httponly, which is as simple as adding a line like this to the cookie setting statement:
:secure => true

Maybe I should create a new issue to make this a secure cookie? What do you think?

mkllnk: Wow, thank you for that insight. Even though it's not crucial for this cookie, we should definitely use HttpOnly cookies everywhere. Go ahead and create that issue.

luisramos0: Issue #2554  with PR #2555 in code review.

app/assets/stylesheets/darkswarm/modals.css.scss
kristinalim: Referencing PR #2544.

In summary, I discovered that there were issues with the JS modal library (`angular-bootstrap`) and the styles we are using:

1. The 10px space at the top is actually intended to be 10% in the styles. The JS library is just treating it as px, I don't know if by design.
2. There is a race condition between the -25% animation of the modal and `focus()` in the JS library which causes the page to scroll up (if still possible).
3. Offset from the top plus the height of the modal (on any screen size) should not exceed 100%, or else some content could be hidden. The change of @luisramos0 decreasing `max-height` for large screens from 120% to 100% is a big improvement. #2544 also addresses the lack of `max-height` for small screens.

I see that this is already merged. Maybe we can move discussion of the modal dimensions to #2544? I've just rebased it upon the updated master. I'll see how it affects the cookies policy modal, and also check #1766.

app/assets/stylesheets/darkswarm/cookies_banner.css.scss
```diff
@@ -0,0 +1,30 @@
+@import 'branding';
+
+.cookies-banner {
+  background: $dark-grey;
+  bottom: 0;
+  position: fixed;
+  top: 20vh !important;
```
kristinalim: What browser versions do we aim to support? I happened to check this recently: https://caniuse.com/#feat=viewport-units

mkllnk: I don't think we have a set of supported browsers. It would be good to create that list though. I'll post something in Slack.

myriamboure:
@luisramos0 

> 2- I can see it on desktop as well. my bug. It's a very easy fix (Ive fixed it locally now).

Did you fix it already? I was not sure while reading it. Let me know I'll open an issue if not fixed yet.

myriamboure:
Ok I opened the remaining issues, just didn't table modal as it seems pretty ok actually when I retested, l et's see after Kristina has finished the work she is doing on modal how it looks like for the cookie policy page ;-)

myriamboure:
It is merged already, so let's discuss on epic now yes ;-) The list you made is on the epic btw, if you look at all issues attached below main thread ;-)

luisramos0:
yes, zenhub is nice :-)  I like to do these summaries for clarity.

Issue #2014, opened 2017-12-21
Refactor uses of spree_variants.count_on_hand


oeoeaio:
`VariantOverride` is particularly coupled to this column, but we should be able to override the new `variant.total_on_hand` method instead of `variant.count_on_hand` to achieve the same result. This is a bandaid solution but a longer term solution to this issue will likely appear out of Networked Products discussions.

sauloperez:
As per the investigation done in https://github.com/openfoodfoundation/openfoodnetwork/pull/2183, we can't get rid of `spree_products.count_on_hand` yet as removing it requires removing `spree_variants.count_on_hand` as well. They depend on each other.

As shown in https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/models/spree/product_decorator.rb#L257-L271, `#save_master` triggers Spree's `#recalculate_product_on_hand`, which indeed updates said column on the `spree_products` table.

mkllnk:
`Spree::Variant` decorator

- `scope :in_stock`: Not on Spree's model. Plenty of use of method `in_stock?`, but no use of the scope `in_stock`.

`VariantOverride`

- It has its own `count_on_hand` value. If this value is present, it uses it for stock management. Otherwise it expects that it's not used for stock. `increment_stock!` will intentionally raise an error instead of using the variant's `count_on_hand` value.

More uses of `count_on_hand`:

- https://github.com/openfoodfoundation/openfoodnetwork/blob/eb9064f5be3433ed72f708c32bccbaa635a2988d/app/models/product_import/entry_processor.rb#L69-L74
- https://github.com/openfoodfoundation/openfoodnetwork/blob/eb9064f5be3433ed72f708c32bccbaa635a2988d/app/models/product_import/entry_processor.rb#L196-L198
- https://github.com/openfoodfoundation/openfoodnetwork/blob/eb9064f5be3433ed72f708c32bccbaa635a2988d/app/models/product_import/entry_validator.rb#L265-L271
- https://github.com/openfoodfoundation/openfoodnetwork/blob/eb9064f5be3433ed72f708c32bccbaa635a2988d/app/serializers/api/variant_serializer.rb#L2
- https://github.com/openfoodfoundation/openfoodnetwork/blob/eb9064f5be3433ed72f708c32bccbaa635a2988d/app/services/order_factory.rb#L46
[...]

PR #2413, opened 2018-06-28
2014 refactor variant count on hand

mkllnk:
#### What? Why?

Part of #2014.

This prepares the inventory report and maybe more for the Spree upgrade. It also fixes a bug in filtering stocked variants and removes some unused code.

#### What should we test?

- Check that the inventory report displays the right list of variants.
- Also check shops for correct display of item availability.
- And sanity check stock management.

#### Release notes

Fixed bug in inventory report which ignored the distributor or on-demand setting for stock calculation. This may slow down the inventory report generation.

This refactors some code to be compatible with Spree 2.0.


mkllnk:
@sauloperez I just noticed that I repeated your commit 0cbd967007d05dee77d440e5f2f406ef70eaef2d. @Matt-Yorkley added that line in a503b5e2d0b1a703f1bc5527a45241d7bb7776e4. I guess that was a merge conflict mistake.

lib/open_food_network/scope_variant_to_hub.rb
```diff
@@ -21,6 +20,14 @@ def price_in(currency)
         Spree::Price.new(amount: price, currency: currency)
       end
 
+      # Old Spree has the same logic as here and doesn't need this override.
+      # But we need this to use VariantOverrides with Spree 2.0.
+      def in_stock?
+        return true unless Spree::Config[:track_inventory_levels]
+
+        on_demand || (count_on_hand > 0)
```
sauloperez: I'm not familiar with variant overrides but I guess `count_on_hand` is a column on a database table added by OFN, right?

Matt-Yorkley: It's a spree database column on both spree_product and spree_variant.

Matt-Yorkley: Just a heads-up @mkllnk: `on_demand` is being removed as well in 2.0.

mkllnk: @Matt-Yorkley Thanks for the heads up. I wasn't sure if I considered that. But this method is only active if the variant has a VariantOverride. The VariantOverride does have `on_demand` and `count_on_hand`. The vanilla Spree::Variant still has an `in_stock?` method in both Spree versions. So we should be safe to call it.

mkllnk:
@sauloperez Yes, I did test that just now. But it should also be covered by our specs.

What do you mean with "too early"? This is production code, going into master.

mkllnk:

RachL:
@myriamboure or @sstead I'm afraid I don't know very well the inventory feature to test this PR. Can one of you have a look? I will be interested in ready the testing notes afterwards.
I've started some notes here, but let's be honest you can just use the document to save yourself the time of creating it, other than that nothing much interesting:
https://docs.google.com/document/d/1-2soDLPWbvUQNpubrDmUE6B1lrK1VlNJWNHrrpt8UM8/edit

sstead:
@mkllnk can you clarify, are we testing all 3 reports within the Products and Inventory report?

I don't have anymore time to test today but here's where I'm up to so far...

Issues:
- In some cases, when I filter by distributor only, I'm seeing products from suppliers who have not granted that shop P-OC E2E. E.g. Sally2's Hub and Producer of Wine.
- In some cases, products are not showing for unknown reasons
- When filtering by OC, the report is filtering by the 'incoming' section of the OC, not outgoing.
- Some inventory settings aren't applying in the report - e.g. I would expect 'hidden' inventry items to be hidden on the report - not the case.
- There seems to be a bug with the Advanced settings for OCs, which makes it hard to test the impact of this setting on the OC filter in the report.

I think there's some design flaws with these reports:
- We shouldn't use the work 'inventory' unless we're specifically refering to the Inventory page. In the case of the Inventory (on hand) report, if you're filtering by a supplier only, it's got nothing to do with the Inventory page, it's referring to On Hand only. I suggest we take the word Inventory out of this report.
- It's confusing the way this report switches between looking at the BEP page (in the case of filtering by supplier only) and then switching to look at both the BEP page and the Inventory page (in the case of filtering by a distributor and a supplier). 
- In the case where I select a distributor filter, I would prefer if the other filters were then restricted to allowable options - e.g. only OCs which that distributor is in. Or only suppliers who have granted P-OC to.

https://docs.google.com/document/d/1-2soDLPWbvUQNpubrDmUE6B1lrK1VlNJWNHrrpt8UM8/edit#

kirstenalarsen:
@sstead how sure are you that these problems are all new / due to this change? I am not at all across this, just wondering if your excellent issue detection skills have ever been applied to these reports before?

sstead:
Yes there’s a very good chance these are prexisting, this isn’t a report I test often. Should we log the other issues for later PRs? I can test more narrowly if that’s what’s needed.

________________________________

daniellemoorhead:
@sstead best way to tell if these are existing is to test on production, yes?

mkllnk:
I would be very surprised if those issues are caused by my code change. I only changed the `on_hand` filter. That already fixed a bug that the `on_demand` property was ignored.

This report is much older than the new inventory feature. It probably does make sense to rename it now. Maybe stock report?

Anyway, to proceed with the Spree upgrade, I just want the report to perform as well as before. Changing the report and fixing it are other issues.

sstead:
**Testing notes**

- Yes, the Inventory on hand report and the lettuceshare report no longer show items with <1 on hand value in Inventory.
- The shop seems to be behaving as normal
- The stock management is working as normal (pre-existing bug only)

Broader issues spotted while testing are logged in this issue - https://github.com/openfoodfoundation/openfoodnetwork/issues/2529

https://docs.google.com/document/d/1-2soDLPWbvUQNpubrDmUE6B1lrK1VlNJWNHrrpt8UM8/edit#

PR #2417, opened 2018-06-29
[WIP] Import count_on_hand methods for Spree 2

mkllnk:
The added methods will be gone in Spree 2. I imported them and added
logic to work with new and old Spree.

#### What? Why?

Helps with #2014.

This is a proposal for discussion, possibly at the next Spree update call. It adds logic to deal with current and new Spree code and is therefore a new approach to the Spree upgrade.

When thinking about refactoring our uses of `count_on_hand`, I saw a lot of work. But all this work could be obsolete when we come to the network feature and get rid of VariantOverride. That's why I propose to add compatibility code instead of refactoring everything now.

I think that this a very quick and easy solution to Spree moving the stock logic into another model. It also allows us to merge this change now. That should avoid merge conflicts. More importantly, it makes every developer working with the current code aware of changes in this area and that they should not use `count_on_hand` in SQL queries or as direct attribute assignment. After the Spree upgrade, we can and should simplify that code. That's the downside. It also needs testing with the new Spree version.

#### What should we test?

Everywhere where variants are used and their stock is shown. Examples:

- Shop front
- Product import
- Reports
- Viewing and editing orders

#### Release notes

Added more compatibility to Spree 2.0.

This is to avoid conflicts with the Spree upgrade.

### DIFF
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -84,6 +84,54 @@ Spree::Variant.class_eval do
     ]
   end
 
+  # Overriding `Spree::Variant.on_hand` in old Spree versions.
+  # It doesn't exist in newer Spree versions.
+  def on_hand
+    if respond_to? :total_on_hand
+      # This is Spree 2.0
+      total_on_hand
+    elsif Spree::Config[:track_inventory_levels] && !on_demand
+      count_on_hand
+    else
+      Float::INFINITY
+    end
+  end
+
+  # Overriding `Spree::Variant.on_hand=` in old Spree versions.
+  # It doesn't exist in newer Spree versions.
+  def on_hand=(new_level)
+    error = 'Cannot set on_hand value when Spree::Config[:track_inventory_levels] is false'
+    raise error unless Spree::Config[:track_inventory_levels]
+
+    self.count_on_hand = new_level unless on_demand
+  end
+
+  # Overriding Spree::Variant.count_on_hand in old Spree versions.
+  # It doesn't exist in newer Spree versions.
+  def count_on_hand
+    if respond_to? :total_on_hand
+      # This is Spree 2.0
+      total_on_hand
+    else
+      # We assume old Spree and call ActiveRecord's method.
+      # We can't call the same method on Variant, because we are overwriting it.
+      super
+    end
+  end
+
+  # Overriding `Spree::Variant.count_on_hand=` in old Spree versions.
+  # It doesn't exist in newer Spree versions.
+  def count_on_hand=(new_level)
+    if respond_to? :total_on_hand
+      # This is Spree 2.0
+      overwrite_stock_levels new_level
+    else
+      # We assume old Spree and call ActiveRecord's method.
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
I really like this approach. It could potentially allow the Spree upgrade to get finished a lot quicker, whilst also giving more breathing room in terms of decoupling existing code from Spree.

Matt-Yorkley:
If we could use similar techniques to get us up to the point where Solidus was forked we might be able to avoid reinventing the wheel with our own set of hooks throughout the Spree codebase.

Matt-Yorkley:
@sauloperez What's you opinion on this solution? ^

app/models/spree/variant_decorator.rb
```diff
@@ -84,6 +84,54 @@ def self.indexed
...
+    end
+  end
+
+  # Overriding `Spree::Variant.on_hand=` in old Spree versions.
+  # It doesn't exist in newer Spree versions.
+  def on_hand=(new_level)
+    error = 'Cannot set on_hand value when Spree::Config[:track_inventory_levels] is false'
+    raise error unless Spree::Config[:track_inventory_levels]
+
+    self.count_on_hand = new_level unless on_demand
```
sauloperez: :warning: this is directly affected by https://github.com/openfoodfoundation/spree/pull/8

sauloperez:
I think this is worth discussing in the next Spree upgrade catch up. See my comment in https://openfoodnetwork.slack.com/archives/C4NDJT3FY/p1531749287000116

app/models/spree/variant_decorator.rb
mkllnk: Ah, that's where I saw this code before! Thank you. So this pull request would override Frank's change.

app/models/spree/variant_decorator.rb
```diff
@@ -84,6 +84,54 @@ def self.indexed
...
+    else
+      # We assume old Spree and call ActiveRecord's method.
+      # We can't call the same method on Variant, because we are overwriting it.
+      super
+    end
+  end
+
+  # Overriding `Spree::Variant.count_on_hand=` in old Spree versions.
+  # It doesn't exist in newer Spree versions.
+  def count_on_hand=(new_level)
```
Matt-Yorkley: Are these methods necessary? `Spree::Variant` doesn't have `count_on_hand` or `count_on_hand=`methods, it uses the `on_hand` methods as getters and setters.

Matt-Yorkley: Or is this related to Variant Overrides?

sauloperez:
Also related to https://github.com/openfoodfoundation/openfoodnetwork/issues/2450

app/models/spree/variant_decorator.rb
mkllnk: In Spree 1.3 the `variants` table has a `count_on_hand` column. So ActiveRecord adds these methods automatically and we have a few calls to these methods in our code base. In order to keep the old API alive, we need these methods.

PR #2419, opened 2018-06-29
Remove spree product navigation override

### LINKED ISSUE #2412: Duplicate "Products" entry in admin menu
Pull request #2367 moved our bulk products edit screen into the place Spree's product edit screen had been. But now we have two menu entries in the submenu for products: "PRODUCTS" and "SPREE PRODUCTS".

## Expected Behavior

Only one menu entry to edit products, no menu entry called "SPREE PRODUCTS".

## Actual Behavior

screenshot from 2018-06-28 09-41-32

## Steps to Reproduce

1. Log in as admin.
2. Click on Products.

## Severity

s4. It's affecting all enterprise users and can lead to confusion, but it doesn't stop anyone from using it.

## Possible Fix

Somewhere we added an override to add the new menu item. That one can be removed. @frank-west-iii Are you up for it?

frank-west-iii:
#### What? Why?

Closes #2412 

This was here to override the original spree products navigation link,
but since moving back to the index action this had the side effect of
adding an additional navigation item instead. We can simply remove this
override to get rid of the additional menu item.

#### What should we test?

Ensure when in admin product pages that the additional "Spree Products" sub-menu item is no longer present, but the "Products" sub-menu item still exists.

### DIFF
--- a/app/overrides/spree/admin/shared/_product_sub_menu/add_products_tab.html.haml.deface
+++ b/app/overrides/spree/admin/shared/_product_sub_menu/add_products_tab.html.haml.deface
@@ -1,4 +1,0 @@
-/ insert_bottom "[data-hook='admin_product_sub_tabs']"
-
-- if spree_current_user.admin?
-  = tab :spree_products, url: admin_products_path, :match_path => '/products'


mkllnk:

RachL:
Ready to go :)

Here are the testing notes : 
https://docs.google.com/document/d/1Wct84n0HnhrTMUGmpC7QBdDQ8IWDnw0dwQnybDOMLOk/edit#

PR #2420, opened 2018-06-29
Stripe admin 2

stveep:
#### What? Why?
Adds Stripe.js (Elements) to the backend so that payments can be manually added by admins using Stripe.

*This PR replaces #1934, modified based on @oeoeaio 's comments*

#### What should we test?

- Credit card payments can be processed in the backend (orders -> select order -> payments -> add new payment).
- Manual payment by other methods is not affected.
- Checkout flow in the frontend is not affected.
- Cards used do not appear in the admin user's list of saved cards (in /account). Note that the card tokens are still saved in the database even though the `save_requested_by_customer` option is not passed as a param - I can look into disabling this if it's a problem?

#### Release notes

Stripe can now be used in the admin backend to manually add credit card payments to an order.

#### Discourse thread

https://community.openfoodnetwork.org/t/integrating-stripe-into-ofn/664

### DIFF
--- a/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
+++ b/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
@@ -4,7 +4,7 @@ angular.module("admin.payments").controller "PaymentCtrl", ($scope, Payment, Sta
   $scope.StatusMessage = StatusMessage
 
   $scope.submitPayment = () ->
-    return false if $scope.submitted == true
+    return false if $scope.submitted
     $scope.submitted = true
     StatusMessage.display 'progress', t("spree.admin.payments.source_forms.stripe.submitting_payment")
     Payment.purchase()
--- a/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
+++ b/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
@@ -6,5 +6,5 @@ angular.module("admin.payments").controller "PaymentCtrl", ($scope, Payment, Sta
   $scope.submitPayment = () ->
     return false if $scope.submitted == true
     $scope.submitted = true
-    StatusMessage.display 'progress', t("submitting_payment")
+    StatusMessage.display 'progress', t("spree.admin.payments.source_forms.stripe.submitting_payment")
     Payment.purchase()
--- a/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
+++ b/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
@@ -1,0 +1,9 @@
+angular.module("admin.payments").controller "PaymentCtrl", ($scope, Payment, Loading) ->
+  $scope.form_data = Payment.form_data
+  $scope.submitted = false
+
+  $scope.submitPayment = () ->
+    return false if $scope.submitted == true
+    $scope.submitted = true
+    Loading.message = t("submitting_payment")
+    Payment.purchase()
--- a/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
+++ b/app/assets/javascripts/admin/payments/controllers/payment.js.coffee
@@ -1,9 +1,10 @@
-angular.module("admin.payments").controller "PaymentCtrl", ($scope, Payment, Loading) ->
+angular.module("admin.payments").controller "PaymentCtrl", ($scope, Payment, StatusMessage) ->
   $scope.form_data = Payment.form_data
   $scope.submitted = false
+  $scope.StatusMessage = StatusMessage
 
   $scope.submitPayment = () ->
     return false if $scope.submitted == true
     $scope.submitted = true
-    Loading.message = t("submitting_payment")
+    StatusMessage.display 'progress', t("submitting_payment")
     Payment.purchase()
--- a/app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
+++ b/app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
@@ -10,7 +10,7 @@ angular.module('admin.payments').directive "stripeElements", ($injector, AdminSt
       stripe = $injector.get('stripeObject')
 
       card = stripe.elements().create 'card',
-        hidePostalCode: true
+        hidePostalCode: false
         style:
           base:
             fontFamily: "Roboto, Arial, sans-serif"
--- a/app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
+++ b/app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
@@ -1,4 +1,4 @@
-angular.module("admin.payments").directive "stripeElements", ($injector, StripeElements) ->
+angular.module('admin.payments').directive "stripeElements", ($injector, AdminStripeElements) ->
   restrict: 'E'
   template: "<label for='card-element'>\
              <div id='card-element'></div>\
@@ -31,5 +31,5 @@ angular.module("admin.payments").directive "stripeElements", ($injector, StripeE
           displayError.textContent = ''
         return
 
-      StripeElements.stripe = stripe
-      StripeElements.card = card
+      AdminStripeElements.stripe = stripe
+      AdminStripeElements.card = card
--- a/app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
+++ b/app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
@@ -1,0 +1,35 @@
+angular.module("admin.payments").directive "stripeElements", ($injector, StripeElements) ->
+  restrict: 'E'
+  template: "<label for='card-element'>\
+             <div id='card-element'></div>\
+             <div id='card-errors' class='error'></div>\
+             </label>"
+
+  link: (scope, elem, attr)->
+    if $injector.has('stripeObject')
+      stripe = $injector.get('stripeObject')
+
+      card = stripe.elements().create 'card',
+        hidePostalCode: false
+        style:
+          base:
+            fontFamily: "Roboto, Arial, sans-serif"
+            fontSize: '16px'
+            color: '#5c5c5c'
+            '::placeholder':
+              color: '#6c6c6c'
+      card.mount('#card-element')
+
+      # Elements validates user input as it is typed. To help your customers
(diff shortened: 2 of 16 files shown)


app/overrides/spree/admin/payments/new/add_angular_to_form.html.erb.deface
```diff
@@ -0,0 +1,4 @@
+<!-- replace 'code[erb-loud]:contains("form_for @payment")' -->
```
mkllnk: Why erb instead of haml?

app/views/spree/admin/payments/_form.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="alpha three columns">
     <div class="field">
       <%= f.label :amount, t(:amount) %>
-      <%= f.text_field :amount, :value => @order.outstanding_balance, :class => 'fullwidth' %>
+      <%= f.text_field :amount, :value => @order.outstanding_balance, :class => 'fullwidth', "watch-value-as" => 'form_data.amount', "ng-init" => "form_data.amount=#{@order.outstanding_balance}" %>
```
mkllnk: Why `watch-value-as` instead of `ng-model`?

app/assets/javascripts/admin/payments/controllers/payment.js.coffee
```diff
@@ -0,0 +1,10 @@
+angular.module("admin.payments").controller "PaymentCtrl", ($scope, Payment, StatusMessage) ->
+  $scope.form_data = Payment.form_data
+  $scope.submitted = false
+  $scope.StatusMessage = StatusMessage
+
+  $scope.submitPayment = () ->
+    return false if $scope.submitted == true
```
sauloperez: `if $scope.submitted` should be enough

app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
```diff
@@ -0,0 +1,35 @@
...
+      # Elements validates user input as it is typed. To help your customers
+      # catch mistakes, you should listen to change events on the card Element
+      # and display any errors:
+      card.addEventListener 'change', (event) ->
+        displayError = document.getElementById('card-errors')
+        if event.error
+          displayError.textContent = event.error.message
+        else
+          displayError.textContent = ''
+        return
```
sauloperez: weird indentation. It looks like it was closing the conditional block

app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
```diff
@@ -0,0 +1,44 @@
...
+      else if ccType == 'Diners Club'
+        return 'diners_club'
+      return
+
+    # It doesn't matter if any of these are nil, all are optional.
+    makeCardData: (secrets) ->
+      {'name': secrets.name,
+      'address1': secrets.address1,
+      'city': secrets.city,
+      'zipcode': secrets.zipcode}
```
sauloperez: isn't this the same as returning `secrets` as it is? or has it more attributes than these?

config/locales/en_GB.yml
```diff
@@ -2101,7 +2101,9 @@ en_GB:
       payments:
         source_forms:
           stripe:
-            no_payment_via_admin_backend: Creating Stripe-based payments from the admin backend is not possible at this time
+            error_saving_payment: Error saving payment
```
sauloperez: :warning: please, do not update other than `config/locales/en.yml`

app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
stveep: It does have some other attributes (e.g. payment method ID) - it wouldn't necessarily be a problem to pass these to Stripe but there is no reason to do it.

app/views/spree/admin/payments/_form.html.erb
stveep: This was rob's suggestion from the previous PR, as the data flow is only one way here.

app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
```diff
@@ -0,0 +1,44 @@
...
+          StatusMessage.display 'error', response.error.message
+        else
+          secrets.token = response.token.id
+          secrets.cc_type = @mapCC(response.token.card.brand)
+          secrets.card = response.token.card
+          submit()
+
+    # Maps the brand returned by Stripe to that required by activemerchant
+    mapCC: (ccType) ->
+      if ccType == 'MasterCard'
```
mkllnk: What about this version?
```coffee
switch ccType
  when 'MasterCard' then return 'master'
  when 'Visa' then return 'visa'
  # ..
```

Or:

```coffee
brand = 'MasterCard'
types = {
  MasterCard: 'master',
  Visa: 'visa',
  # ...
}
return types[brand]
```

app/assets/javascripts/admin/payments/directives/stripe_elements.js.coffee
sauloperez: can we a line in between?

app/views/spree/admin/payments/_form.html.erb
oeoeaio: Do you need the `ng-init`? I was hoping the `watch-value-as` directive would take care of the initial assignment...

RachL:
HI @stveep Is there something special to do to connect the staging server to a stripe account ? 

I've started to test but I have some errors when I'm trying to pay:

https://docs.google.com/document/d/1Da9Ue2alfr3hxAUbiaUQhTqcbKxLADZig5su56ff49s/edit#heading=h.7tt4jhsodd8d

Maybe @myriamboure or @sstead you do have inputs as well ? Thanks

sstead:
@RachL that Stripe gateway is a testing one, so you need to use testing credit card details. full list here https://stripe.com/docs/testing . That might have been your issue?

**Testing notes**
Working:
- yes I can apply Stripe payments in the backend
- Other methods unaffected - yes, but 2 preexisting bugs found.
- Checkout flow in the front end is fine
- Cards used in the backed don't save in /account

To be improved:
- Only issue is one weird error message seen in the backend for some orders
- Suggestion to make the Stripe card fields more obvious

https://docs.google.com/document/d/1Da9Ue2alfr3hxAUbiaUQhTqcbKxLADZig5su56ff49s/edit#

mkllnk:
Back to you @stveep. There is a Javascript error that breaks the page. I have the same error on my dev machine. It's about injecting `currentOrderNumber`, I think.

RachL:
@sstead ah yes thank you! I didn't know about these testing card, good to know!

stveep:
It would be helpful if you could give some more details about the JS error...by breaking the page are you talking about the  {{ StatusMessage... || "" }} in Sally's notes? I don't get it on my machine (seems fine on Rachel's too so could be a config issue as Sally says), can you post the console log?

sstead:
Sorry @stveep that PR is no longer in staging and the order with the error no longer exists so I can't get the console info for you. Does this need to be staged again?

mkllnk:
@stveep The error message was in Sally's testing notes as well:
js-error

But now that I'm checking it again, I actually get a different error at a different time. I tested Firefox and Chrome. I loaded the dev sample data and placed one order. Then I go to the create payment form: http://0.0.0.0:3000/admin/orders/R771072247/payments/new
```
Error: [$injector:unpr] Unknown provider: currentOrderNumberProvider <- currentOrderNumber <- Payment
http://errors.angularjs.org/1.5.5/$injector/unpr?p0=currentOrderNumberProvider%20%3C-%20currentOrderNumber%20%3C-%20Payment
    at all.js:26859
    at all.js:31249
    at Object.getService [as get] (all.js:31402)
    at all.js:31254
    at getService (all.js:31402)
    at injectionArgs (all.js:31426)
    at Object.invoke (all.js:31448)
    at Object.enforcedReturnValue [as $get] (all.js:31295)
    at Object.invoke (all.js:31456)
    at all.js:31255
```

stveep:
OK I found the source of the JS error - it was happening for enterprises that didn't have Stripe set up, as the angular app depended on some stuff that was injected in the Stripe partial that wasn't being rendered in these cases.

I added a label for the card field. I can split it into separate fields if you want but just wanted to check as we already had a similar discussion about this for the front end and decided not to split into separate fields...

daniellemoorhead:
>I added a label for the card field. I can split it into separate fields if you want but just wanted to check as we already had a similar discussion about this for the front end and decided not to split into separate fields...

@stveep what would the value of splitting it into two fields provide, that would make the original decision worth changing? 

@myriamboure I think it was you that worked with @stveep on the previous decision to not split the fields, yes? So I think for consistency sake it's worth you making the decision this time too :)

myriamboure:
https://github.com/openfoodfoundation/openfoodnetwork/issues/2352 that was the issue that triggered the discussion about splitting fields. Two PR were open, this one https://github.com/openfoodfoundation/openfoodnetwork/pull/2403  was moved forward although @RachL and I recommended the second option on UX prospective... I don't really get why it would make sense on this PR to split or not to split, but the case was this one. Is it related? If yes and would be better to split, maybe it's good to do it as we found it better on mobile as well...

daniellemoorhead:
Over to you @stveep - if you're happy with it as it is then we can get it staged and tested. If you believe the split should be done to improve it then put it back to in dev and go for it 😃

mkllnk:
Staged on https://staging1.openfood.com.au/. The staged branch also contains https://github.com/openfoodfoundation/openfoodnetwork/pull/2580 so that it can be tested at the same time. If the test instructions for this PR are fine, then #2580 can be merged as well.

RachL:
Hi guys, while re-testing this I've stumbled upon another issue:
If I try to submit while having an empty Postal code an error is shown, which is good.
However when I’m filling in the postal code, the error still stays when I’m trying to process the payment (not the case with other type of errors). I have to reload the page to make it work.
If I remember correctly we do not need the postal code, I think we deleted this field on other part of the platform right? Can we do the same here? ping @myriamboure @sstead @stveep 

https://docs.google.com/document/d/1Da9Ue2alfr3hxAUbiaUQhTqcbKxLADZig5su56ff49s/edit#

RachL:
@mkllnk I guess #2580 can go through however?

PR #2431, opened 2018-07-05
2007 Skip shipment creation for payment step

### LINKED ISSUE #2007: Remove call to #create_shipment!
As described in `go_to_state :payment` in
https://community.openfoodnetwork.org/t/order-model-step7/979, Spree now creates
shipment objects out of packages in memory, which then presents in a form that
are going to be persisted when the user submits the form. Check it out in
[283654d4a330924b171441427b26ce89469b4b04](https://github.com/spree/spree/commit/283654d4a330924b171441427b26ce89469b4b04).

As a result, we can get rid of https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/models/spree/order_decorator.rb#L34

EDIT

Turns out that due to https://trello.com/c/jQuW8D5A and https://trello.com/c/prYaoPlB we can get rid of the whole custom `checkout_flow`

### LINKED ISSUE #2432: [Spree 2 Upgrade] Refactor Order#shipping_method in checkout flow
## Description

The customization we added in the payment step of the checkout flow relies on `Order#shipping_method` which is gone in Spree 2.0. Therefore, this logic needs to be refactored to work on that Spree version.

https://github.com/openfoodfoundation/openfoodnetwork/blob/09534b41e9a806daf91c5633d0b03e009e297b30/app/models/spree/order_decorator.rb#L33-L42

mkllnk:
#### What? Why?

Closes #2007 and closes #2432.

Spree 2.0.4 creates in-memory shipments for the payment step. That means
we don't need to create a shipment in that transition. This change
does not override the whole `checkout_flow` from the Spree code, but still removes the confirm state.

#### What should we test?

I don't think we can test this yet. But once we can, we should test the checkout with fees on shipping methods. Some methods should require a shipping address and others not.

#### Release notes

Spree upgrade.

It is part of it.

#### Discourse thread

https://community.openfoodnetwork.org/t/order-model-step7/979

### DIFF
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -30,20 +30,14 @@ Spree::Order.class_eval do
   checkout_flow do
     go_to_state :address
     go_to_state :delivery
-    go_to_state :payment, :if => lambda { |order|
-      # Fix for #2191
-      if order.shipping_method.andand.delivery?
-        if order.ship_address.andand.valid?
-          order.create_shipment!
-          order.update_totals
-        end
-      end
+    go_to_state :payment, if: ->(order) {
+      order.update_totals
       order.payment_required?
     }
     # NOTE: :confirm step was removed because we were not actually using it
     # go_to_state :confirm, :if => lambda { |order| order.confirmation_required? }
     go_to_state :complete
-    remove_transition :from => :delivery, :to => :confirm
+    remove_transition from: :delivery, to: :confirm
   end
 
   # -- Scopes
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -27,18 +27,10 @@ Spree::Order.class_eval do
   before_save :update_shipping_fees!, if: :complete?
   before_save :update_payment_fees!, if: :complete?
 
-  checkout_flow do
-    go_to_state :address
-    go_to_state :delivery
-    go_to_state :payment, if: ->(order) {
-      order.update_totals
-      order.payment_required?
-    }
-    # NOTE: :confirm step was removed because we were not actually using it
-    # go_to_state :confirm, :if => lambda { |order| order.confirmation_required? }
-    go_to_state :complete
-    remove_transition from: :delivery, to: :confirm
-  end
+  # Orders are confirmed with their payment, we don't use the confirm step.
+  # Here we remove that step from Spree's checkout state machine.
+  # See: https://guides.spreecommerce.org/developer/checkout.html#modifying-the-checkout-flow
+  remove_checkout_step :confirm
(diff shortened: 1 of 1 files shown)


app/models/spree/order_decorator.rb
```diff
@@ -30,20 +30,14 @@
...
-      end
+    go_to_state :payment, if: ->(order) {
+      order.update_totals
       order.payment_required?
     }
     # NOTE: :confirm step was removed because we were not actually using it
     # go_to_state :confirm, :if => lambda { |order| order.confirmation_required? }
     go_to_state :complete
-    remove_transition :from => :delivery, :to => :confirm
+    remove_transition from: :delivery, to: :confirm
```
sauloperez: if that is the only thing we modify, isn't there a way from Spree to just remove this transition to avoid having to duplicate all the other steps? Having that single line will make it immediately obvious that that is the only change.

sauloperez:
As it is, this also closes https://github.com/openfoodfoundation/openfoodnetwork/issues/2432

app/models/spree/order_decorator.rb
mkllnk: Good point. I had another read about the Spree state machine and added another commit.

app/models/spree/order_decorator.rb
```diff
@@ -27,24 +27,10 @@
...
-    }
-    # NOTE: :confirm step was removed because we were not actually using it
-    # go_to_state :confirm, :if => lambda { |order| order.confirmation_required? }
-    go_to_state :complete
-    remove_transition :from => :delivery, :to => :confirm
-  end
+  # Orders are confirmed with their payment, we don't use the confirm step.
+  # Here we remove that step from Spree's checkout state machine.
+  # See: https://guides.spreecommerce.org/developer/checkout.html#modifying-the-checkout-flow
+  remove_checkout_step :confirm
```
sauloperez: this is soooo much better now!

sauloperez:
have you tried somehow to place an order to be sure this keeps working (hopefully)? It could simply be creating an order from the rails console.

mkllnk:
@sauloperez The missing method `shipping_method` prevents the shop from working, but I just ran our order spec:

Before: 73 examples, 62 failures
After: 73 examples, 51 failures

So I think we are fixing more than we break. ;-)

I also created an order in the console, but didn't try to emulate a whole checkout.

sauloperez:
Let's first clarify https://github.com/openfoodfoundation/openfoodnetwork/pull/2366#issuecomment-407345408 before merging this.

sauloperez:
All good here. This doesn't directly affect #2366 as the checkout flow was almost identical.

PR #2436, opened 2018-07-05
Add diagram for new multiple shipments

sauloperez:
#### What? Why?

We want to start documenting the decisions around each of the main topics of the Spree upgrade in the wiki. This is just the first diagram to help explain it.

This will allow us to store the image in GH's servers and reference it from https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade-Themes


sauloperez:
I'll give that a try.

sauloperez:
[Close WIP PRs](https://trello.com/c/micYXEgs/15-close-wip-prs)

mkllnk:
Fair enough to keep images as documentation with the code. There is just the review process everything goes through. The wiki is quicker to edit.

sauloperez:
It works! https://github.com/openfoodfoundation/openfoodnetwork/wiki/Migration-to-multiple-shipments

I added the picture at the root of the repo, as all the pages, and used https://github.com/openfoodfoundation/openfoodnetwork/wiki/split_shipments_diagram.jpg as its URL. Now it depends on what we prefer either keeping the diagrams together with the code or leave that up to the wiki only.

I have to say that I find more comfortable and productive to edit wiki pages on my machine instead of GH's text editor, which is rather limited. For instance, now I can publish the meeting's minutes straight away by simply pushing upstream. Thoughts?

mkllnk:
I think the wiki is easier to edit. And if there is text referencing and describing the diagrams, it's nice to have them together. On the other hand, a diagram describes code. So if the code changes and we have to update the diagram, then it would be nice to include both in one pull request. The question is: Are we likely to update the wiki as well? If we don't need to update the wiki when a diagram changes, we should keep the diagram with the code. But if we need to update the wiki anyway, let's keep the documentation complete in the wiki with the diagrams and update it after merging the code change.

Another idea to tackle the problem of keeping code and documentation in sync: Git submodules. The wiki could be a submodule in the OFN repo. That would enable us to do the code change and the documentation change in one go. But it's still complicated and people need to learn about submodules. So I would actually advice against submodules, but wanted to bring it up as idea for creativity here. ;-)

sauloperez:
I'll keep the diagram in the wiki then as it is for spree upgrade purposes only.

PR #2441, opened 2018-07-09
Fix front-end sorting in "Customers" index and "Bulk Order Management" page

### LINKED ISSUE #589: Customers: columns need to be sortable - nothing happening on click
image

### LINKED ISSUE #2440: Bulk Order Management: Sorting is not working
Front-end sorting in the "Bulk Order Management" page is not working.

## Description

This is of the same nature as the bug described in #589.

## Expected Behavior

Clicking the header for a sortable column such as "Order No.", "Name", and "Order Date" should sort the line items accordingly. Clicking the header again should sort the line items in reverse order.

## Actual Behavior

Nothing happens.

## Steps to Reproduce

Before doing the following, there should already be a few recently completed orders in your database.

1. Go to the "Bulk Order Management" page.
2. Click on the column header "Name". The line items should be sorted by customer name.
3. Click on the column header "Name" again. The line items should be sorted by customer name in reverse order.

## Severity

4

## Possible Fix

This should be fixed in the same way as #589.

@oeoeaio describes an effective solution [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/589#issuecomment-288942315).

kristinalim:
#### What? Why?

Closes #589 
Closes #2440 

Front-end sorting in the "Customers" index and the "Bulk Order Management" page is not working. This is mostly because of a scope-related issue. @oeoeaio describes the history of the bug [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/589#issuecomment-288942315).

The scope for the page controllers did not have access to the sorting preferences stored in the nested `ColumnsCtrl` scope. To address this, the page controllers have been changed to use a new set of sorting preferences declared in the page controller scope itself.

Also, these sorting preferences are now stored in an object. This enables the parent scope to see changes to the sorting preferences which are done via the nested `ColumnsCtrl` scope,

#### What should we test?

For the "Customers" index:

1. Go to the "Customers" index.
2. Click on the column header "Email". The customers should be sorted by customer email.
3. Click on the column header "Email" again. The customers should be sorted by customer email in reverse order.
4. Repeat 2 and 3 for the other sortable columns.

For the "Bulk Order Management" page:

1. Go to the "Bulk Order Management" page.
2. Click on the column header "Name". The line items should be sorted by customer name.
3. Click on the column header "Name" again. The line items should be sorted by customer name in reverse order.
4. Repeat 2 and 3 for the other sortable columns.

#### Release notes

- Fix front-end sorting in "Customers" index.
- Fix front-end sorting in "Bulk Order Management" page.

### DIFF
--- a/app/views/admin/customers/index.html.haml
+++ b/app/views/admin/customers/index.html.haml
@@ -72,27 +72,28 @@
             %th.actions
               Ask?&nbsp;
               %input{ :type => 'checkbox', 'ng-model' => "confirmDelete" }
-        %tr.customer{ 'ng-repeat' => "customer in filteredCustomers = ( customers | filter:quickSearch | orderBy:predicate:reverse ) | limitTo:customerLimit track by customer.id", 'ng-class-even' => "'even'", 'ng-class-odd' => "'odd'", :id => "c_{{customer.id}}" }
-          -# %td.bulk
-            -# %input{ :type => "checkbox", :name => 'bulk', 'ng-model' => 'customer.checked' }
-          %td.email{ 'ng-show' => 'columns.email.visible'}
-            %span{ 'ng-bind' => '::customer.email' }
-            %span.guest-label{ 'ng-show' => 'customer.user_id == null' }= t('.guest_label')
-          %td.name{ 'ng-show' => 'columns.name.visible'}
-            %input{ type: 'text', name: 'name', ng: { model: 'customer.name' }, 'obj-for-update' => 'customer', 'attr-for-update' => 'name'}
-          %td.code{ 'ng-show' => 'columns.code.visible' }
-            %input{ type: 'text', name: 'code', ng: {model: 'customer.code', change: 'checkForDuplicateCodes()'}, "obj-for-update" => "customer", "attr-for-update" => "code" }
-            %i.icon-warning-sign{ ng: {if: 'duplicate'} }
-              = t('.duplicate_code')
-          %td.tags{ 'ng-show' => 'columns.tags.visible' }
-            .tag_watcher{ 'obj-for-update' => "customer", "attr-for-update" => "tag_list"}
-              %tags_with_translation{ object: 'customer', 'find-tags' => 'findTags(query)' }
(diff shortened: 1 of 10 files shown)


sauloperez:
Thanks for your contribution @kristinalim !

app/views/admin/customers/index.html.haml
```diff
@@ -61,36 +61,37 @@
             -# %th.bulk
               -# %input{ :type => "checkbox", :name => 'toggle_bulk', 'ng-click' => 'toggleAllCheckboxes()', 'ng-checked' => "allBoxesChecked()" }
             %th.email{ 'ng-show' => 'columns.email.visible' }
-              %a{ :href => '', 'ng-click' => "predicate = 'customer.email'; reverse = !reverse" }=t('admin.email')
+              %a{ :href => '', 'ng-click' => "sorting.predicate = 'customer.email'; sorting.reverse = !sorting.reverse" }=t('admin.email')
```
mkllnk: Setting `reverse` to `true` initially is a lot better than `false`. But it's still a bit weird. If I click on Email, it's sorting normally. If I then click on another column, it's reversed, right? I'm wondering if we should add the method `sorting.sort(column)` with code like:
```coffee
reverse = (predicate == column) && !reverse
predicate = column
```

kristinalim: @mkllnk Yes, you're right - it should reset the reverse setting when switching to a new column. I will make a follow-up change for that.

kristinalim:
I appended commits that address the following issues:

* As pointed out by @mkllnk, clicking on a second column could sometimes result in records being sorted by that column in reverse order. Ideal behaviour is that clicking on a new column should reset the "reverse" setting.
* The records in the customer index were not actually being sorted correctly. The sort predicates in the customer index were wrong - they weren't supposed to have `customer.`.

kristinalim:
I rebased this branch, and the checks pass. :slightly_smiling_face: This PR can be reviewed again. Thank you!

mkllnk:

PR #2437, opened 2018-07-05
Modify English tool tip text in yml file

Ecksi:
#### What? Why?

Closes #1848

Before:
before img

After:
after img

Attempted to edit the styling properties of PowerTip but current value used for positioning is `smartPlacement: true` which is supposed to keep the tooltip within the viewport. We believe the text of the tooltip being a long continual string is causing the issue. We decided to shorten the verbiage of the tooltip to allow PowerTip to better handle the placement. Styling still isn't ideal but we believe this is a decent fix without rework of the PowerTip implementation.

As our only available language in our group is English we only applied the shortened tooltip text to the English yml files. We did not want to provide incorrect translations, but we recommend the other language tooltips be changed as well.

#### What should we test?

These changes only modify display text, the test suite we ran had no broken tests.

PR #2447, opened 2018-07-11
Move i18n locale doc to wiki and reference that

mkllnk:
#### What? Why?

Helps to prevent unnecessary work as in https://github.com/openfoodfoundation/openfoodnetwork/pull/2437#issuecomment-403774344.

We regularly receive pull requests with changes to translation files. This change moves the documentation into the wiki and makes clear that you shouldn't work on other files than `en.yml`.

#### What should we test?

No code change, no testing needed.

#### Release notes

Improved documentation for contributors.

Not.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,58 +1,25 @@
 # English language file
 # ---------------------
 #
-# This is the source language file maintained by the Australian OFN team.
+# This is the source language file maintained by the global OFN team and used
+# by the Australian OFN instance.
+#
 # Visit Transifex to translate this file into other languages:
 #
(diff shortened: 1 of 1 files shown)


kristinalim:
As I mentioned [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2437#issuecomment-404372547), I think these would also help:

* In `CONTRIBUTING.md`, mention that translations other than "en" are managed on Transifex and provide a link to the wiki page.
* In the wiki, move the "i18n" from "Specific features" to "Contributing".
* In the wiki, change the sidebar link to the "i18n" page to "Internationalization (i18n)".

What do you think? These might make the documentation easier to come across.

PR #2467, opened 2018-07-13
Use progress formatter on Travis to avoid timeout

mkllnk:
#### What? Why?

We had a few Travis builds lately that timed out, because no output was
received. Even though our standard formatter Fuubar should detect Travis
and print the right output, nothing was visible in our builds.
https://github.com/travis-ci/travis-ci/issues/1337

In this patch, I tell Travis to use Rspec's default formatter to print
dots for every passed test.

#### What should we test?

No manual testing required.

#### Release notes

Improved test output and stability.

Changelog Category: Changed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ before_script:
 
 script:
   - 'if [ "$KARMA" = "true" ]; then bundle exec rake karma:run; else echo "Skipping karma run"; fi'
-  - "bundle exec rake 'knapsack:rspec[--tag ~performance]'"
+  - "bundle exec rake 'knapsack:rspec[--format progress --tag ~performance]'"
 
 after_success:
   - >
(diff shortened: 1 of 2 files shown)


sauloperez:
I think we should decide whether we continue using Travis or go with Semaphore only. I vote for the latter.

mkllnk:
I'm not sure how we make that decision. We don't really have an online polling tool at the moment. I did just make the decision by deactivating Travis for our main repository. I do think that Semaphore is the better option at the moment. But I would like to keep the Travis configuration so that we have the option to run it on Travis. Other contributors may prefer Travis for their forks as they are used to it already and don't want to sign up for yet another service.

sauloperez:
Ok, so unless someone disagrees, we can remove Travis CI integration from Github making us rely on Semaphore CI only, while keeping the `.travis.yml` so people can use it if they want to.

Issue #2233, opened 2018-04-17
Fix unreliable features test


sauloperez:
## Description

As exposed in https://github.com/openfoodfoundation/openfoodnetwork/pull/2232 recent changes may have dropped the following tests reliability: 
https://github.com/openfoodfoundation/openfoodnetwork/blob/ac37b0fc7e7203477ad59f61081b393c69fdfc26/spec/features/consumer/shopping/embedded_shopfronts_spec.rb#L50-L97

It fails way more than it passes in CI.

## Expected Behavior
The test should pass either like this or refactored.

## Actual Behavior
It fails most of the times on CI, not so sure about development environment.

## Steps to Reproduce
Open a PR, no matter what you do, and check it out

## Severity

Not having this test may have an impact on our test suite's coverage so this needs to be done sooner rather than later, so assigning a severity 3.

## Possible Fix

We need to find out what happened to it, fix it and then enable it back. The fact of being a so long feature test greatly increases the chances of running into strange failures.

sauloperez:
We don't normally use severity levels for these issues so please, @myriamboure let me know if this feels right to you.

myriamboure:
Sounds ok to me...I guess we can use severity labels to know the emergency of a tech debt to be solved...? It's a bit hacky, but let's review that when we go. Maybe @daniellemoorhead will have ideas ;)

sauloperez:
> I guess we can use severity labels to know the emergency of a tech debt to be solved...?

That's my point. Let's what the authority, @daniellemoorhead , has to say.

myriamboure:
Let's prioritize it @sauloperez we want to have some things in the pipe occasional devs can jump on. I propose we put this one in, so moving to delivery backlog.

mkllnk:
Rob has made some great progress using headless Chrome as browser for testing: https://github.com/openfoodfoundation/openfoodnetwork/compare/master...oeoeaio:headless-chrome

Since he won't be working on this for another month, I'm willing to take over. This has been really annoying and it costs us a lot of time reviewing and staging pull requests.

daniellemoorhead:
Hey there, so I have removed the bug label...because it's not a bug in the _traditional_ sense of the word. However, I do agree that we need to understand whether tech debt items are a priority and need to be done, so that they are added to the delivery pipe.

I think in the same way we sort bugs we also need to manage a tech debt backlog, and ensure that things like this are prioritised alongside of bugs and alongside of features.

[...]

PR #2468, opened 2018-07-13
Improve testing embedded pages

mkllnk:
#### What? Why?

Working on #2233 I looked at the specs around embedded shops and embedded groups again. They had several flaws which I fixed here. The shopping spec is still deactivated, because we haven't done anything about its reliability yet.

#### What should we test?

No manual testing needed.

#### Release notes

Improved automated tests.

Changelog Category: Changed

### DIFF
--- a/spec/features/consumer/shopping/embedded_groups_spec.rb
+++ b/spec/features/consumer/shopping/embedded_groups_spec.rb
@@ -53,12 +53,17 @@ feature "Using embedded shopfront functionality", js: true do
       end
     end
 
-    it 'opens links to shops in a new window' do
+    it "opens links to shops in a new window" do
       on_embedded_page do
         within 'div#group-page' do
-          enterprise_links = page.all(:xpath, "//*[contains(@href, 'enterprise-5/shop')]", :visible => :false).count
-          enterprise_links_with_target_blank = page.all(:xpath, "//*[contains(@href, 'enterprise-5/shop') and @target = '_blank']", :visible => :false).count
-          expect(enterprise_links).to equal(enterprise_links_with_target_blank)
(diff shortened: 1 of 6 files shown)


spec/features/consumer/shopping/embedded_groups_spec.rb
```diff
@@ -63,16 +53,29 @@
...
+          end
         end
       end
     end
   end
+  
+  private
+  
+  def on_embedded_page
+      expect(page).to have_selector "iframe"
```
sauloperez: do we really need this expectation here? the tests that use this method already check other assertions so I'm not sure it's worth the cost of this extra `expect`.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
```diff
@@ -110,6 +105,16 @@
...
 
+  private
+  
+  def on_embedded_page
+      expect(page).to have_selector "iframe"
+
+      within_frame :frame do
+        yield
+      end
+  end
```
sauloperez: I'd DRY it by moving its definition to `spec/support/`

sauloperez: Now that I see, this might only be used in this file. Then 👍

mkllnk: This was used in two files, so I moved it to a helper to take the refactoring another step further. I also removed the expectation which I think isn't needed. I'm not sure why @Matt-Yorkley did it in https://github.com/mkllnk/openfoodnetwork/commit/113f6565be71cf2b2cf2ca0c3ee622f576f8e1fc#diff-2f114975ce4c238323ecb0c23daab4b7R47, but it could be, because he used the web driver directly to load a local file and wanted to assert that it worked.

PR #2472, opened 2018-07-15
Logo sizing to css

### LINKED ISSUE #2470: [MobUX] Remove Hard Coded Height and Width in Banner Logos
Banner logo height and width is hard coded. This means that it cannot be overwritten in CSS. As a result the logo does not scale on mobile.

## Expected Behavior
Banner logo height and width should be set within CSS, rather than in code.

## Actual Behavior
Banner logo height and width is hard coded.

## Context
This issue is blocking some quick win tweaks to mobile UX.

lilwillifo:
#### What? Why?

Closes #2470

The logo height in width were in some cases set directly in the HTML rather than in CSS. This prevented the logo from scaling on mobile, as style could not be written over by CSS. I moved the formatting over to CSS.

#### What should we test?

Confirm home page logos (center and in footer) are now optimized for mobile. 

#### Release notes

Changelog Category: Fixed

### DIFF
--- a/app/assets/stylesheets/darkswarm/footer.scss
+++ b/app/assets/stylesheets/darkswarm/footer.scss
@@ -36,6 +36,7 @@ footer {
 
       img {
         margin-top: 36px;
+        width: 120px;
       }
     }
 
--- a/app/assets/stylesheets/darkswarm/home_tagline.css.scss
+++ b/app/assets/stylesheets/darkswarm/home_tagline.css.scss
@@ -34,7 +34,7 @@
       max-width: 45%;
 
       @media all and (min-height: 500px) {
-        max-width: 80%;
+        max-width: 250px;
       }
 
       margin-bottom: 2rem;
--- a/app/views/home/_tagline.html.haml
+++ b/app/views/home/_tagline.html.haml
@@ -3,7 +3,7 @@
     .small-12.text-center.columns
       %h1
         / TODO: Rohan - logo asset & width is content manageable:
-        %img{src: "/assets/logo-white-notext.png", width: "250", title: Spree::Config.site_name}
+        %img{src: "/assets/logo-white-notext.png", title: Spree::Config.site_name}
         %br/
         %a.button.transparent{href: "/shops"}
           = t :home_shop
--- a/app/views/shared/_footer.html.haml
+++ b/app/views/shared/_footer.html.haml
@@ -3,7 +3,7 @@
     .row
       .small-12.columns.text-center
         .logo
-          %img{src: "/assets/logo-white-notext.png", width: "120px"}
+          %img{src: "/assets/logo-white-notext.png"}
     .row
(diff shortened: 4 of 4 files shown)


app/assets/stylesheets/darkswarm/home_tagline.css.scss
```diff
@@ -34,7 +34,7 @@
       max-width: 45%;
 
       @media all and (min-height: 500px) {
-        max-width: 80%;
```
Matt-Yorkley: Hi @lilwillifo, thanks for the contribution :slightly_smiling_face: 
I wonder if the max-width in this media query for < 500px screens is set to `max-width: 80%` for a reason... did you experiment with this at different screen sizes?

Matt-Yorkley:
Looking at the current site on UK production it looks like there is another view in play that affects mobiles, and it also needs a similar adjustment: `app/views/shared/menu/_mobile_menu.html.haml`. I think it only appears once the screen is a small enough size.

app/assets/stylesheets/darkswarm/home_tagline.css.scss
Matt-Yorkley: There's a bit of complexity in displaying these logos, in that each country has logo files that are slightly different widths, so we have to keep that in mind.

app/assets/stylesheets/darkswarm/footer.scss
```diff
@@ -36,6 +36,7 @@ footer {
 
       img {
         margin-top: 36px;
+        width: 120px;
```
sauloperez: Isn't `img` a too broad selector? if the logo is all we want to change, I'd rather use `.logo` instead.

mkllnk: This is nested and will be `h1 img`. There is only one `h1` with only one `img`. It doesn't have a logo class. We would need to add something like a `header__logo` class. But I think this pull request is fine as it is, not invasive.

sauloperez: Now I see this rule is actually nested within `.logo` so nevermind.

lin-d-hop:
Is this good to merge now?

mkllnk:
> Is this good to merge now?

Not yet. It needs testing first.

mkllnk:

RachL:
@mkllnk @lin-d-hop I'm not sure of something. When you speak about "banner logo", for the top banner, which of these two are you speaking of : 
image

I'm not seeing much difference before / after the PR, so I'm not sure I'm testing the right thing...

RachL:
Thanks @Matt-Yorkley then everything is ok :) 
Some screenshots here : 
https://docs.google.com/document/d/1PWPEVeiEesgK7Qdvd71L7M3mRU7aIwhtznYugcdkRcM/edit#

mkllnk:
Congrats, @lilwillifo. First pull request merged within eight days. That's really good.

PR #2473, opened 2018-07-15
PI: Don't allow save if invalid entries are present

### LINKED ISSUE #2456: Product import: if error is detected, proceed to import is not possible
## Description
Today if an error is detected in the import file, it doesn't prevent fro importing the spreadsheet and saving the data. We believe this might lead to errors and confusion for users who will not know what was imported or not afterwards. So as a first release we will not enable user to proceed if there are errors in the spreadsheet, and guide them so that they clearly understand the error and fix it before proceeding.

## Acceptance criteria
capture du 2018-07-12 09-34-10
capture du 2018-07-12 09-34-38
- If no error, you directly see the button "save" (instead of proceed which is unclear) + cancel button
- If error, you only have one "back to file upload" button

## Test cases:
- We need to test all possible errors we coudl think of and review if error message is precise enough to help the user to fix it.
- If try to import product into a non producer enterprise, shouldn’t work and error should be displayed
- ...

Matt-Yorkley:
#### What? Why?

Closes #2456.

Changes the Product Import logic so that saving is not possible unless all entries in the spreadsheet are valid. The previous logic allowed the user to save the valid entries and ignore the invalid ones. They will now have to correct their spreadsheet first, and re-import it.

#### What should we test?

Product Import does not allow save unless all entries are valid.

#### Release notes

Product Import: Valid products can not be saved if there are validation errors on other products.

Changelog Category: Changed

### DIFF
--- a/app/views/admin/product_import/import.html.haml
+++ b/app/views/admin/product_import/import.html.haml
@@ -7,16 +7,16 @@
 
   - if @importer.item_count == 0 #and @importer.invalid_count
     %h5
-      = t('admin.product_import.import.no_valid_entries')
+      = t('.no_valid_entries')
     %p
-      = t('admin.product_import.import.none_to_save')
+      = t('.none_to_save')
     %br
   - else
     .settings-section{ng: {show: 'step == "settings"'}}
       = render 'import_options' if @importer.table_headings
       %br
       %a.button.proceed{href: '', ng: {click: 'confirmSettings()'}}
-        = t('admin.product_import.import.proceed')
+        = t('.proceed')
       %a.button{href: main_app.admin_product_import_path} #{t('admin.cancel')}
 
   .progress-interface{ng: {show: 'step == "import"'}}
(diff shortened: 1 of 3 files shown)


app/views/admin/product_import/import.html.haml
```diff
@@ -41,7 +41,7 @@
       %div{ng: {if: 'count((entries | entriesFilterValid:"invalid")) > 0'}}
         %br
         %h5= t('admin.product_import.import.some_invalid_entries')
-        %p= t('admin.product_import.import.save_valid?')
+        %p= t('admin.product_import.import.fix_before_import')
```
sauloperez: doesn't the Rails i18n namespacing work here?

sauloperez:

Matt-Yorkley:
The user cannot proceed to _saving_ if there are any errors in the file. 

There is another issue open for changing the flow between the different pages and how these pages look, which I think will include completely removing that second page that lists the enterprises. That will come in another PR for that issue once some of these others are merged.

Matt-Yorkley:
A validation check to see if the enterprise is enabled as a producer in their enterprise settings is not something I tested for before or added code for, so I think another issue to cover that specific case would be the best option.

RachL:
@Matt-Yorkley alright, the new issue is here : #2490 I'm moving this PR to ready to go then ;)

Issue #2463, opened 2018-07-13
Cloning an OC sets Schedule to 'none' which is not editable from OC overview screen


kirstenalarsen:
Trying to get OC rapid creation more workable . . following from #2430 and #2439

When an OC is cloned, the schedule is set to 'none'

![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/f3e409f6-87c9-4a07-a3f0-87bf39469215)

This 'None' is NOT clickable to pull up the Schedule / OC selector, so the User does need to go into the OC to select the Schedule. Annoying.

## Expected Behavior
Complete #2439
and
make Schedule = 'None' also clickable to bring up the Schedule Updater 

![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/52e510d1-186b-4669-912c-e5a486f323f8)

## Actual Behavior
A user needs to go into the OC to set Schedule every time they create / clone a new OC. Once they have set it once, they can open the Schedule Updater by clicking on the Schedule name from the OC Overview screen

![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/a3d47ae7-e2fe-4f29-9a50-6f147b05e82f)

![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/be058900-1762-4dd5-bd77-a4d76fb146b9)

anon0mys:
Hi @kirstenalarsen, I am a new dev and am interested in contributing! I noticed that this issue was referenced in a PR for another issue, but I am not sure if that PR fixes the issue. Do you still need someone to work on this? Thanks!

oeoeaio:
@kirstenalarsen I'm not sure this is needed? I can see two cases where an OC does not belong to a schedule and so has 'None' in the schedules column:

1. The schedule that you want to add the unassigned OC to already exists. In this case, there must be other order cycles assigned to it, as you cannot have an empty schedule. Therefore, you can just click on another reference to it elsewhere in the page, and then add the unassigned OC to that schedule.
2. The schedule that you want to add the unassigned OC to does not already exist. In this case, you should just create a new schedule at the top right of the screen and add the unassigned OC to it.

The screen that comes up when you click on a schedule name is an edit form for a particular schedule. If we add the ability to click 'None', what would you like to see? We can't show the normal screen that is shown because we have not specified a schedule to render a form for...

oeoeaio:
[...]

PR #2476, opened 2018-07-15
Clone oc copy schedule #2439

### LINKED ISSUE #2439: Subscriptions: cloning OC's should copy schedule
## Description

Enterprise users need to set up lots of basically identical order cycles in a schedule before they can start using subscriptions.

The current clone feature (available from the Order Cycle index page) does not clone key information - most importantly the schedule itself which is left blank

This means that the user must open each order cycle to set the schedule - infuriating. This combined with [G2430](https://github.com/openfoodfoundation/openfoodnetwork/issues/2430) makes this initial essential step very annoying and likely off-putting

## Expected Behavior

When an Order Cycle is cloned from the Order Cycle index page, the schedule information should be copied

## Actual Behavior

The schedule information is left blank

## Steps to Reproduce
1. Go to Order Cycles
2. Select an Order Cycle that is in a Schedule (or put one in one - which will require creating a schedule)
3. Clone it - using button on the right of the of
image
4. notice that cloned OC has no schedule

## Severity
I'm putting this as severity 3. In conjunction with #2530 it makes OC creation for setting up a schedule very very painful and likely off-putting to early adopters

## NB - Relationship to previously reported issue #1052 
I think this is a better and simpler solution and so #1052 should be closed
My (light) investigations suggest that products fees and tags are being correctly captured in cloning so it is just the schedule that needs to be included

ssciolist:
#### What? Why?

Closes #2439: Cloning button within admin/order cycles created a clone that did not copy orders open and orders close values. This issue is referenced in #2463. 

I believe this issue stems from the way that clone! was defined at the OrderCycle model. Previous behavior was to set both the orders_open_at and orders_closed_at to nil.

#### What should we test?

From path admin/order_cycles, confirm that the clone button now creates a copy where the open and close values match the original.

Changelog Category:  Changed

No known relation

### DIFF
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -122,7 +122,8 @@ class OrderCycle < ActiveRecord::Base
   def clone!
     oc = self.dup
     oc.name = "COPY OF #{oc.name}"
-    oc.orders_open_at = oc.orders_close_at = nil
+    oc.orders_open_at = self.orders_open_at
+    oc.orders_close_at = self.orders_close_at
     oc.coordinator_fee_ids = self.coordinator_fee_ids
     oc.preferred_product_selection_from_coordinator_inventory_only = self.preferred_product_selection_from_coordinator_inventory_only
     oc.save!
--- a/spec/features/admin/order_cycles_spec.rb
+++ b/spec/features/admin/order_cycles_spec.rb
@@ -591,6 +591,17 @@ feature %q{
     # Then I should have clone of the order cycle
     occ = OrderCycle.last
     expect(occ.name).to eq "COPY OF #{oc.name}"
+    expect(occ.products).to eq oc.products
+    expect(occ.schedules).to eq oc.schedules
+
+    # On the page, expect the clone to have the same values as the original
+    within "tr.order-cycle-#{occ.id}" do
+      page.should have_input "oc#{occ.id}[orders_open_at]", value: oc.orders_open_at
+      page.should have_input "oc#{occ.id}[orders_close_at]", value: oc.orders_close_at
+      page.should have_selector 'td.schedules', text: 'None'
+      page.should have_selector 'td.coordinator', text: "#{oc.coordinator.name}"
+      page.should have_selector 'td.products', text: '0 variants'
+    end
   end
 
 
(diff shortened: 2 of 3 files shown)


kirstenalarsen:
Hi @ssciolist - thanks so much for your first PR!

I am really sorry I think I didn't specify this clearly enough for someone new to OFN - assumed knowledge is our enemy. My apologies . . 2 follow up things below

The issue was not intended to be the open and close dates - it is/was correct that these should not be copied so that someone with a cloned order cycle is clear that they need to check and set them. 

1. The field that we want / need to have cloned is the one called Schedules - shown in image below. This is part of a major new feature called 'Subscriptions' that is in Beta and requires 'turning on' for specific shops - so you won't see it in your test data until you turn it on. User guide is here - https://guide.openfoodnetwork.org/advanced-features#subscriptions

image

As you have already worked out the cloning thing it should be (hopefully) pretty minor to pick up and include cloning of the 'schedule.

2. As what we are trying to do here is improve UX for someone cloning multiple order cycles to set up a Schedule, I wonder if what you have done here might support that - it most likely means a lot less clicking for someone setting consecutive dates if the dates from previous one are already there (rather than default opening at current date and needing to click back or forward multiple months). There could be an argument for accepting this change even though it is not what I meant. However it would increase possibility of error if someone doesn't look closely enough and actually check all the dates .. the implications of someone not checking their oc dates properly are very annoying so perhaps introducing that risk / likely user mistake would outweigh slight convenience improvement of cloning open/close dates?
 - @myriamboure @sstead do you have a view on this? 
Should OC open / close dates be cloned or not?

ssciolist:
Ah, I see @kirstenalarsen. That makes a lot of sense, thank you for clarifying. Unfortunately, I’ve had to drop this issue from my to-do list. Sorry about that, and hopefully someone else will be able to pick this up!

sstead:
I would suggest we avoid cloning the start/end dates to avoid one potential outcome which would be very annoying... At the moment, if someone cloned an open order cycle which was in a schedule, the moment the clone was made it would be an open OC, which would trigger subscription orders to be created and emails to go out to customers.

myriamboure:
Agree with @sstead I'm closing that PR then also because we said cloning schedules might not be such a good idea either (cf Rob's feedback)... let's understand the need and specify clearly before putting things to dev ready ;-) We can use in the project boards or bug management boad a column "define and design" before dev ready so whenever there is some consent needed at least from one or two people we can take the time to get it. For most bugs it's not necessary, but here we talk of a new feature which is "improve OC quick creation UX", so it needs discussions.

PR #2482, opened 2018-07-18
Allow preferred_mails_from and preferred_mail_bcc to be set from configs

Matt-Yorkley:
#### What? Why?

Needed for issue: https://github.com/openfoodfoundation/ofn-install/issues/202

Twinned with PR: https://github.com/openfoodfoundation/ofn-install/pull/203

Mail settings are being wiped out on each deploy and have to be re-set by instance managers. I've added the use of "defaults" as the second argument in the two `ENV.fetch()` calls that I've changed, so there shouldn't be any change for instances where the new ENV vars are not present.

#### What should we test?

Nothing to test for now.

Emails should work and mail settings should be added correctly, but this can't really be tested without some additional ofn-install work and re-deploying.

#### Release notes

Mail Settings: don't delete from_address and bcc settings from mail methods on deployment

Changelog Category: Fixed

### DIFF
--- a/config/application.yml.example
+++ b/config/application.yml.example
@@ -29,6 +29,10 @@ MAIL_PORT: 25
 SMTP_USERNAME: 'ofn'
 SMTP_PASSWORD: 'f00d'
 
+# Optional mail settings
+# MAILS_FROM: hello@example.com
+# MAIL_BCC: manager@example.com
+
 # SingleSignOn login for Discourse
 #
 # DISCOURSE_SSO_SECRET should be a random string. It must be the same as provided to your Discourse instance.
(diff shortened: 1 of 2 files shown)


config/application.yml.example
```diff
@@ -28,6 +28,8 @@ MAIL_DOMAIN: 'example.com'
 MAIL_PORT: 25
 SMTP_USERNAME: 'ofn'
 SMTP_PASSWORD: 'f00d'
+# MAILS_FROM: hello@example.com
+# MAIL_BCC: manager@example.com
```
sauloperez: I guess you commented this out because they are not mandatory :thinking: . If that is the case, I'd explain why is that in a comment block preceding these lines. I'm afraid it might not be clear for new devs otherwise.

PR #2487, opened 2018-07-20
Make CodeClimate config more strict

sauloperez:
#### What? Why?

There's no point of using Code Climate if most of its checks are
disabled. Now it scores our codebase with an A and reports just 2 code
smells, 64 duplications and 14 other issues. That couldn't farther from
reality. If we care about the architecture and maintainability of our
app, this is a must.

To transition towards enabling all default checks, I started with
scss-lint, duplication and method-complexity. This doesn't mean will
have to fix them all straight away as we'll add the current violations
to the `.rubocop_todo.yml` but it'll point us the parts where we should
focus our efforts.

#### What should we test?

This is just a 3rd party service's configuration. No test required.

#### Release notes

Make Code Climate configuration stricter

Changelog Category: Changed


sauloperez:
Totally agree but also, I think there are two different topics here: style guide and complexity checks. While for the former it's just about preference and picking one, we should run all checks of the latter. It's not about styling but well-established programming best practices that do have a reason. Also `.rubocop_todo.yml` allows us to enable them and set a threshold, so win-win.

PR #2488, opened 2018-07-20
[Spree Upgrade] Remove dependency to Dash and Jirafe

### LINKED ISSUE #2434: Spree Upgrade - Remove dependency to Dash in Spree
## Description
Dash has been removed from Spree in 2.0
See what we are using it for and remove dependency.

See spec/features/admin/Overview_spec:116 with_dash_configured

Interesting reference: https://www.quora.com/What-happened-to-Jirafe-analytics

## Expected Behavior
Pass the referenced test.

## Actual Behavior
Test fails.

## Steps to Reproduce
Run build.

## Your Environment
spree branch

## Possible Fix
Maybe remove it. Is it necessary?

HugsDaniel:
#### What? Why?

Closes #2434 

As explained in the issue, Dash has been removed in Spree 2.0. Since it is not needed, I removed it from the specs.

#### What should we test?

The build should have more passing tests as references to Dash were causing fails.

#### Release notes

Dash and Jirafe related tests have been removed with Spree Upgrade to 2.0.

Changelog Category: Removed

Dash dependencies have been removed in Spree 2.0.

### DIFF
--- a/app/models/spree/ability_decorator.rb
+++ b/app/models/spree/ability_decorator.rb
@@ -73,7 +73,6 @@ class AbilityDecorator
 
   def add_group_management_abilities(user)
     can [:admin, :index], :overview
-    can [:admin, :sync], :analytic
     can [:admin, :index], EnterpriseGroup
     can [:read, :edit, :update], EnterpriseGroup do |group|
       user.owned_groups.include? group
@@ -86,7 +85,6 @@ class AbilityDecorator
     can [:create, :search], nil
 
     can [:admin, :index], :overview
-    can [:admin, :sync], :analytic
 
     can [:admin, :index, :read, :create, :edit, :update_positions, :destroy], ProducerProperty
 
--- a/spec/features/admin/overview_spec.rb
+++ b/spec/features/admin/overview_spec.rb
@@ -103,44 +103,5 @@ feature %q{
         end
       end
     end
-
-    context "with the spree dash configured" do
-      let(:d1) { create(:distributor_enterprise) }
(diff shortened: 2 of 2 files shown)


HugsDaniel:
I can't find any migration related to Spree Dash config in spree-2-0-stable. Also I can't find where or how exactly Dash has been removed from Spree. @luisramos0 @sauloperez did you go through Spree and found out back then ?

sauloperez:
@HugsDaniel you can dig deeper in Spree using `git log -S <string_to_match>`, so matching something like `Dash` or `jirafe` would do the trick. I believe the engine was completely removed so `git log -- dash/` should return you something as well.

mkllnk:
Oh actually, the dash configuration is stored in Spree preferences: https://github.com/spree/spree/commit/31070896a4b6919255bb9dc76b8f3bef64308bc1#diff-3291f4fc33a72e8a78e481f6e066070eL2

If we have these configurations in our databases, they are now orphaned. It's probably not a big deal, but could lead to confusion later. It would be a problem if somebody introduced preferences with the same key names. Can you add a migration that removes these configurations?

sauloperez:
Are we really sure preferences are a stored in the DB? I thought they were in the cache. In any case, if we had to, what about doing that in a different PR so that we can benefit from these changes soon?

mkllnk:
I'm not 100% sure, but I think the preference store does store in the database if the table exists. Anyway. These values don't hurt at the moment and I don't even know if anybody set it up.

PR #2492, opened 2018-07-21
Added basic integration with Matomo

### LINKED ISSUE #2404: [Cookies] Investigate what it takes to replace google analytics by Matomo
## Description

To implement the cookie regulation and be legally compliant we have two options:
- if we stay with google analytics we need to do more work on ability to enable/disable cookies, and let instance decide about default opt-in/opt-out, etc.
- if we switch to matomo (piwik) we can leave it as an essential cookie as it doesn't collect any personal data so it simplifies a lot the implementation of the cookie feature. As we said we anyway want to switch to Matomo maybe it worth spending max 2 hours on investigating what it takes to implement Matomo and if worth doing it now to gain time on cookie feature.

luisramos0:
#### What? Why?

Closes #2404 

Integration with Matomo. The OFN admin user can provide In the global settings the matomo url and site id. That enables the Matomo tracking script.

#### What should we test?

Update matomo URL and site Id, make some interactions with the OFN website under test and check Matomo console to see that visits are being the global settings

Mobile visits should also be recorded.

#### Release notes

New Matomo integration.
Changelog Category: Added

### DIFF
--- a/app/overrides/spree/admin/shared/_configuration_menu/add_invoice_settings.html.haml.deface
+++ b/app/overrides/spree/admin/shared/_configuration_menu/add_invoice_settings.html.haml.deface
@@ -1,4 +1,4 @@
 // insert_bottom "[data-hook='admin_configurations_sidebar_menu']"
 
 %li
-  = link_to t('admin.invoice_settings.edit.invoice_settings'), main_app.edit_admin_invoice_settings_path
+  = link_to t('admin.invoice_settings.edit.title'), main_app.edit_admin_invoice_settings_path
--- a/app/overrides/spree/admin/shared/_configuration_menu/add_stripe_connect_settings.html.haml.deface
+++ b/app/overrides/spree/admin/shared/_configuration_menu/add_stripe_connect_settings.html.haml.deface
@@ -1,4 +1,4 @@
 // insert_bottom "[data-hook='admin_configurations_sidebar_menu']"
 
 %li
-  = link_to t('.stripe_connect'), main_app.edit_admin_stripe_connect_settings_path
+  = link_to t('admin.stripe_connect_settings.edit.title'), main_app.edit_admin_stripe_connect_settings_path
--- a/app/views/layouts/darkswarm.html.haml
+++ b/app/views/layouts/darkswarm.html.haml
@@ -22,6 +22,7 @@
 
     = render "layouts/i18n_script"
     = render "layouts/bugherd_script"
+    = render "layouts/matomo_tag"
     = csrf_meta_tags
 
   %body{class: body_classes, ng: {app: "Darkswarm"}}
--- a/app/controllers/admin/matomo_settings_controller.rb
+++ b/app/controllers/admin/matomo_settings_controller.rb
@@ -1,11 +1,11 @@
 module Admin
-  class InvoiceSettingsController < Spree::Admin::BaseController
+  class MatomoSettingsController < Spree::Admin::BaseController
     def update
       Spree::Config.set(params[:preferences])
 
       respond_to do |format|
         format.html {
-          redirect_to main_app.edit_admin_invoice_settings_path
+          redirect_to main_app.edit_admin_matomo_settings_path
         }
       end
     end
--- a/app/models/spree/app_configuration_decorator.rb
+++ b/app/models/spree/app_configuration_decorator.rb
@@ -35,6 +35,8 @@ Spree::AppConfiguration.class_eval do
 
   # External services
   preference :bugherd_api_key, :string, default: nil
+  preference :matomo_url, :string, default: nil
+  preference :matomo_site_id, :string, default: nil
 
   # Invoices & Receipts
   preference :invoice_style2?, :boolean, default: false
--- a/app/overrides/spree/admin/shared/_configuration_menu/add_matomo_settings.html.haml.deface
+++ b/app/overrides/spree/admin/shared/_configuration_menu/add_matomo_settings.html.haml.deface
@@ -1,0 +1,4 @@
+// insert_bottom "[data-hook='admin_configurations_sidebar_menu']"
+
+%li
+  = link_to t('admin.matomo_settings.edit.title'), main_app.edit_admin_matomo_settings_path
--- a/app/views/admin/matomo_settings/edit.html.haml
+++ b/app/views/admin/matomo_settings/edit.html.haml
@@ -1,0 +1,23 @@
+= render :partial => 'spree/admin/shared/configuration_menu'
+
+- content_for :page_title do
+  = t('.title')
+
+= form_tag main_app.admin_matomo_settings_path, :method => :put do
+
+  .div
+    %legend= t('.info_html')
+    = tag(:br)
+    = t('.config_instructions_html')
+
+  .field
+    = label_tag(:matomo_url, t('.matomo_url')) + tag(:br)
+    = preference_field_tag("preferences[#{:matomo_url}]", Spree::Config[:matomo_url], type: Spree::Config.preference_type(:matomo_url))
+
+  .field
+    = label_tag(:matomo_site_id, t('.matomo_site_id')) + tag(:br)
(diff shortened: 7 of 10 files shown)


luisramos0:
I think I should move the two Matomo config fields out of the general settings, but I can't find a good place. Any ideas?

screen shot 2018-07-21 at 16 09 42

luisramos0:
For testing, we need to setup an account in https://www.innocraft.cloud/ and put the URL on the matomo URL field.
I did setup a trial account for my local server and it's very very easy.

luisramos0:
I only need to add the matomo opt out iframe to the cookies policy page.

app/views/layouts/darkswarm.html.haml
```diff
@@ -22,6 +22,7 @@
 
     = render "layouts/i18n_script"
     = render "layouts/bugherd_script"
+    = render "layouts/matomo_tag"
```
sauloperez: Would it be possible to only render the tag if matomo is configured?

luisramos0:
Thanks @stveep for the suggestions!
Any one else?
1- keep the 2 entries on the general settings
2- incorporate the 2 entries on the Analytics Tracker config page that is already there on the sidebar (this is a Spree view that would need to be defaced...)
3- add own config page similar to Stripe Connect

Matt-Yorkley:
I'd vote for a section in the existing analytics tracker page.

sauloperez:
> If you are thinking about defacing a Spree view, I think it's better to copy the whole view into our code and modify it. We would like to separate ourselves from Spree more.

Yep, we didn't write it down yet but it's something we all agree with.

luisramos0:
I agree we should put this on the analytics config page! It's the place it makes the most sense.
I understand we should move the spree view to our code base!
BUT this trackers code (analytics) is to be removed, in the mid-term we will not be using google analytics at all.
So, I dont want to change/move this code to ofn and have to retest it. It will be removed sooner or later...
I'd prefer a quicker fix for this.
So, I'd vote for either:
- deface spree trackers view adding the 2 matomo fields
- OR
- create a new tab just for matomo config
Votes?

sauloperez:
the second option seems the best one to me as well

mkllnk:
Yes, do a new tab. That will make it easier to get rid of the Spree tab in the future.

myriamboure:
> Now I only need to add the matomo opt out iframe to the cookies policy page.

Are you keeping track of that somewhere? Shall we open another issue not to forget about it? As I guess the first PRs need to be accepted before you can do that, right @luisramos0 ?

I see two Ri devs has approved the PR so should go to test ready as soon as you fixed the new Matomo tab thing, just move it yourself to test ready when it is done :-)

luisramos0:
oh, ok. I put the fields in a new tab and we move this one to test.
I will create a new issue for the matomo optout iframe 👍

mkllnk:
Okay, the tester will need a Matomo account. Where do they get it from? Ask @luisramos0 or @myriamboure?

myriamboure:
I created an account for OFN global and have added Rachel and I am as well as our Matomo account users/admin, I can add Sally as well if needs be for testing, or Rachel or I can test. So do we have what we need @mkllnk ? You are also admin by the way, you should have received an email.

mkllnk:
@luisramos0 Maybe you can create a new tab for Matomo tracking. That would give you some space to write a couple of helpful sentences to explain what Matomo is and where to set up an account.

luisramos0:
New Matomo settings screen:
screen shot 2018-08-06 at 21 00 41

I am happy to change the description text.

I'd like this PR to be reviewed before moving it to test ready.

app/controllers/admin/invoice_settings_controller.rb
```diff
@@ -1,13 +1,13 @@
-class Admin::InvoiceSettingsController < Spree::Admin::BaseController
+module Admin
```
luisramos0: The changes on this file were all related to CodeClimate's feedback.

app/overrides/spree/admin/shared/_configuration_menu/add_invoice_settings.html.haml.deface
```diff
@@ -1,4 +1,4 @@
 // insert_bottom "[data-hook='admin_configurations_sidebar_menu']"
 
 %li
-  = link_to t('admin.invoice_settings.edit.invoice_settings'), main_app.edit_admin_invoice_settings_path
```
luisramos0: this was broken.

app/overrides/spree/admin/shared/_configuration_menu/add_stripe_connect_settings.html.haml.deface
```diff
@@ -1,4 +1,4 @@
 // insert_bottom "[data-hook='admin_configurations_sidebar_menu']"
 
 %li
-  = link_to t('.stripe_connect'), main_app.edit_admin_stripe_connect_settings_path
+  = link_to t('admin.stripe_connect_settings.edit.title'), main_app.edit_admin_stripe_connect_settings_path
```
luisramos0: All the translations of settings are using lazy lookup from admin._settings_.edit, the only absolute reference is on this menu deface file (and only for this stripe connect settings). If we don't do this, the relative reference would leave an isolated entry on the en.yml file, which would just add noise to it (see removed line 2599 in en.yml below).

myriamboure:
@luisramos0 @sstead was talking about a "tracking code" but it seems Matomo doesn't work the same way as GA, there is not just one tracking code you can copy/paste (I checked in the tracking code section in Matomo). Does it sound understandable for you @sstead ? 
I looked for the Matomo site ID in https://openfoodnetwork.innocraft.cloud/ and it can be found on Innocraft in Administration/Websites/Manage, maybe that worth adding it, I had to search a bit to find it.
capture du 2018-08-07 09-53-59

mkllnk:
@luisramos0 The point @sstead and I wanted to make is that there is a tracking code that is sent via email when you setup a project. There is also a tracking code in the Innocraft admin panel for copy and paste. While it's easy for a developer to spot the "Matomo URL" in there, it just looks like a cryptic blob of code to others. It would be much easier for the user to copy the whole code into the OFN admin panel instead of only the URL. It would also mean that we use the term "Tracking code" everybody is familiar with and which is used by Innocraft. Shopify does it this way with Google Analytics, too.

Is there a reason why you want to store only the URL? Or is there a reason to keep URL and site id separate?

luisramos0:
Yes, there is a reason for this. These two values are what the matomo tag script takes as parameters. If we take the "tracking code" we would need to parse it to create these two fields... I wouldnt do it. I'd rather write a simple user guide.

luisramos0:
The user guide would be:
- For the URL: go to your matomo console and copy the URL you have on the browser window, only the domain. For example:
     https://openfoodnetwork.innocraft.cloud/index.php?module=SitesManager&action=index&idSite=1&period=day&date=yesterday
take only
   https://openfoodnetwork.innocraft.cloud
- For the Site ID, go to the Admin console (click on the gear wheel icon on the top right corner) and go to the menu Webpages > Manage. The ID fields are the Site IDs you can use in the OFN integration config.

daniellemoorhead:
Three approved reviews for this PR - @mkllnk @luisramos0 @sstead is this ready for testing again? Has Luis answered your question satisfactorily Maikel, Sally?

mkllnk:
@luisramos0 I don't understand why you can't just insert the tracking code into the page. Isn't that the simplest solution?

myriamboure:
@mkllnk I went into Matomo and clicked on tracking code, you can't have an easy tracking code to copy paste apparently, it doesn't work like Google Analytics... as a user I find it much easier to put the url and site id personnaly so I support that option. Please login Matomo interface as a user and check yourself, if you find an easy way to get a tracking code then explain me ;-)

mkllnk:
Okay, if you are happy with those instructions, let's go with it. It's true that the tracking code is a bit hidden in the settings, but I did find it. Anyway, we can change it later if it's a problem.

sstead:
Testing notes

- All looks fine

https://docs.google.com/document/d/1CIMgwXx4mi36sPG5wSEtNILDujI7LmlCl-65AKOql3c/edit#

PR #2493, opened 2018-07-22
Reimplement Spree's simple_product in OFN factory

### LINKED ISSUE #2433: Adapt tests to change from simple_product to base_product in spree
## Description
enterprise caching Spec is using simple product that has now ben transformed into an extension of BaseProduct with stock location.
See here:
https://github.com/spree/spree/commit/14f911e91

## Expected Behavior
Pass the test: 
  context "key-based caching invalidation" do
    describe "is touched when a(n)" do
      describe "with a supplied product" do

## Actual Behavior
Broken test.

## Steps to Reproduce
Run the spree branch build.

## Severity
spree upgrade

## Your Environment
spree branch

## Possible Fix
replace simple_product with base_product

HugsDaniel:
#### What? Why?

Closes #2433 

As explained in the PR, Spree's product factory got rid of `simple_product` in favor of `base_product`, making all of our tests using `simple_product` fail.

I reimplemented the Spree<2.0 `simple_product` factory inside OFN `spec/factories.rb`, inheriting from OFN's override of Spree's `base_product` factory.
As before, a `simple_product` is an extension of a `base_product` with `on_hand 5`.

#### What should we test?

All tests using Spree's `product_factory` don't fail on `simple_product`.

#### Release notes

Spree Upgrade - Reimplementation in OFN of Spree's `simple_product`, inheriting from `base_product`, as a test helper.

Changelog Category: Changed

Spree 2.0 removed `simple_product` in their `product_factory`, so it has been reimplemented in OFN factories to enable its use in test suite.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -530,5 +530,4 @@ FactoryBot.modify do
       user.spree_roles << Spree::Role.find_or_create_by_name!('admin')
     end
   end
-
 end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -453,6 +453,10 @@ FactoryBot.define do
       Spree::Image.create(attachment: image, viewable_id: product.master.id, viewable_type: 'Spree::Variant')
     end
   end
+
+  factory :simple_product, parent: :base_product do
+    on_hand 5
+  end
 end
 
 
@@ -526,4 +530,5 @@ FactoryBot.modify do
       user.spree_roles << Spree::Role.find_or_create_by_name!('admin')
     end
   end
+
 end


HugsDaniel:
On `2-0-stable` I have 2692 failing tests, and on `2433-change-from-simple-product-to-base-product` I have 2684 failing tests.

HugsDaniel:
A lot of tests still fail because of `on_hand` attribute, or `shipping_method` which is called later in the test, etc...

HugsDaniel:
Ok so I should instead create our own product factory, is that right ?

sauloperez:
and call it `simple_product` so no tests need to be changed.

HugsDaniel:
Should I take into account https://github.com/openfoodfoundation/openfoodnetwork/pull/2417/files and keep a `on_hand` or `total_on_hand` attribute for `simple_product` ?

luisramos0:
@HugsDaniel I think you should keep what's in there. on_hand? I believe we will still have on_hand available after the upgrade. even if only an adapter.

mkllnk:
Yes @HugsDaniel. You can assume that there will be methods called `on_hand` and `count_on_hand`. We decided to keep the old API.

spec/factories.rb
```diff
@@ -526,4 +530,5 @@
       user.spree_roles << Spree::Role.find_or_create_by_name!('admin')
     end
   end
+
```
mkllnk: I guess this slipped in accidentally?

HugsDaniel: Yes it did, sorry about that

PR #2495, opened 2018-07-23
Remove unused gem newrelic-rpm

### LINKED ISSUE #2486: Remove New Relic
Several admins voiced that they don't like New Relic being in our code. A poll on Slack showed that nobody is using it any more:
https://openfoodnetwork.slack.com/archives/C0DNLAZC7/p1531802988000065

## Description

Removing New Relic make our dependencies smaller and avoids some confusing log output of the application. Skylight is the new New Relic.

## Expected Behavior

New Relic is not there.

## Actual Behavior

New Relic is installed with the bundle, it is present in log output and there are some unused configuration files.

## Steps to Reproduce

1. Gemfile.lock contains newrelic

## Severity

Tech debt s4.

mkllnk:
#### What? Why?

Closes #2486.

Nobody uses New Relic any more. So let's remove it.

#### What should we test?

Just a basic sanity check that the checkout is still working. Nothing should have changed.

#### Release notes

Removed unused support for New Relic. It got replaced by Sky Light.

Changelog Category: Removed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ public/stylesheets
 public/images
 public/spree
 config/abr.yml
-config/newrelic.yml
 config/initializers/feature_toggle.rb
 config/initializers/db2fog.rb
 NERD_tree*
--- a/Gemfile
+++ b/Gemfile
@@ -38,7 +38,6 @@ gem 'simple_form', github: 'RohanM/simple_form'
 gem 'unicorn'
 gem 'angularjs-rails', '1.5.5'
 gem 'bugsnag'
-gem 'newrelic_rpm'
 gem 'haml'
 gem 'sass', "~> 3.3"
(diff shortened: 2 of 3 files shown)


mkllnk:
You merged without testing? We would usually still deploy this to a staging server and perform some sanity tests like a checkout just to make sure that it's all good.

daniellemoorhead:
@mkllnk @Matt-Yorkley @sauloperez @luisramos0 in terms of Maikel's final comment above:

Should this have been tested before it was merged? Is it good practice to do a sanity check before merging things like this, even if it's quickly staged and looked at by a dev? Would be good to have your input here as to how they should be handled moving forward...

Matt-Yorkley:
I think it should be ok in this case. It shouldn't really interfere with the rest of the codebase, as it was unused anyway.

PR #2500, opened 2018-07-24
Fix link syntax errors in REAME.md and GETTING_STARTED.md

oeoeaio:
#### What? Why?

I stuffed up the link syntax in my changes to README.md and GETTING_STARTED.md, so I've fixed them.

#### What should we test?

No features to test, but should probably look at all the links in README.md, GETTING_STARTED.md and CONTRIBUTING.md and make sure they are all actually pointing to the right place.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -73,9 +73,9 @@ Then the tests can be run with:
 
     bundle exec rspec spec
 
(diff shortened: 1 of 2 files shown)


oeoeaio:
Ah, ok. I'll get @mkllnk to cherry pick this commit.

PR #2512, opened 2018-07-30
Support removal of enterprise logo and promo image

### LINKED ISSUE #423: cannot remove images from enterprise profile
the enterprise user is able to replace images, but has no way to actually just remove them. sometimes would be really useful just to be able to set back to 'no image'

Create 'remove image' for logo and enterprise banner 

image

kristinalim:
#### What? Why?

Closes #423

It has been possible to set or replace the logo and promo image for an enterprise, but there was no way to remove a previously uploaded logo or promo image.

#### What should we test?

Below should be done in the "Images" section of the enterprise settings page, for both the logo and the promo image:

1. **Adding image** - Select image file, and click "Update". This would bring you back to the main section of the enterprise settings page. Go to "Images" again, and check that the image has indeed been uploaded.
2. **Replacing image** - Repeat Number 1, but using another image. The first image should be replaced with the second image.
3. **Confirmation dialog for removal of image** - Click "Remove Image". A confirmation dialog would pop up, warning you that the image would immediately be removed. Select "Cancel". The image should not disappear from the page. Reload the page to check that the image was indeed not removed.
4. **Removing image** - Click "Remove Image" again. This time, in the confirmation dialog, select "OK". You should see a success notice "Logo removed successfully" at the bottom of the page, and the image should disappear from the page. Reload the page to check that the image has indeed been removed.

#### Release notes

- Add support for removal of enterprise logo and promo image

Changelog Category: Added

#### Dependencies

For the Code Climate checks to pass, the following changes are necessary:

* #2516 
* #2517

### DIFF
--- a/app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
+++ b/app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
@@ -69,6 +69,8 @@ angular.module("admin.enterprises")
       $scope.newUser = $scope.invite_errors = $scope.invite_success = null
 
     $scope.removeLogo = ->
+      return unless confirm(t("admin.enterprises.remove_logo.immediate_removal_warning"))
+
       Enterprises.removeLogo($scope.Enterprise).then (data) ->
         $scope.Enterprise = angular.copy(data)
         $scope.$emit("enterprise:updated", $scope.Enterprise)
@@ -79,6 +81,8 @@ angular.module("admin.enterprises")
           StatusMessage.display("failure", response.data.error)
 
     $scope.removePromoImage = ->
+      return unless confirm(t("admin.enterprises.remove_promo_image.immediate_removal_warning"))
+
       Enterprises.removePromoImage($scope.Enterprise).then (data) ->
         $scope.Enterprise = angular.copy(data)
         $scope.$emit("enterprise:updated", $scope.Enterprise)
--- a/app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
+++ b/app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
@@ -1,5 +1,5 @@
 angular.module("admin.enterprises")
-  .controller "enterpriseCtrl", ($scope, $http, $window, NavigationCheck, enterprise, EnterprisePaymentMethods, EnterpriseShippingMethods, SideMenu, StatusMessage) ->
+  .controller "enterpriseCtrl", ($scope, $http, $window, NavigationCheck, enterprise, Enterprises, EnterprisePaymentMethods, EnterpriseShippingMethods, SideMenu, StatusMessage) ->
     $scope.Enterprise = enterprise
     $scope.PaymentMethods = EnterprisePaymentMethods.paymentMethods
     $scope.ShippingMethods = EnterpriseShippingMethods.shippingMethods
@@ -67,3 +67,23 @@ angular.module("admin.enterprises")
 
     $scope.resetModal = ->
       $scope.newUser = $scope.invite_errors = $scope.invite_success = null
+
+    $scope.removeLogo = ->
+      Enterprises.removeLogo($scope.Enterprise).then (data) ->
+        $scope.Enterprise = angular.copy(data)
+        $scope.$emit("enterprise:updated", $scope.Enterprise)
+
+        StatusMessage.display("success", t("admin.enterprises.remove_logo.removed_successfully"))
+      , (response) ->
+        if response.data.error?
+          StatusMessage.display("failure", response.data.error)
+
+    $scope.removePromoImage = ->
+      Enterprises.removePromoImage($scope.Enterprise).then (data) ->
+        $scope.Enterprise = angular.copy(data)
+        $scope.$emit("enterprise:updated", $scope.Enterprise)
+
+        StatusMessage.display("success", t("admin.enterprises.remove_promo_image.removed_successfully"))
+      , (response) ->
+        if response.data.error?
+          StatusMessage.display("failure", response.data.error)
--- a/app/assets/javascripts/admin/resources/services/enterprises.js.coffee
+++ b/app/assets/javascripts/admin/resources/services/enterprises.js.coffee
@@ -38,3 +38,17 @@ angular.module("admin.resources").factory 'Enterprises', ($q, EnterpriseResource
 
     resetAttribute: (enterprise, attribute) ->
       enterprise[attribute] = @pristineByID[enterprise.id][attribute]
+
+    performActionOnEnterpriseResource = (resourceAction) ->
+      (enterprise) ->
+        deferred = $q.defer()
+        resourceAction({id: enterprise.permalink}, ((data) =>
+          @pristineByID[enterprise.id] = angular.copy(data)
+          deferred.resolve(data)
+        ), ((response) ->
+          deferred.reject(response)
+        ))
+        deferred.promise
+
+    removeLogo: performActionOnEnterpriseResource(EnterpriseResource.removeLogo)
+    removePromoImage: performActionOnEnterpriseResource(EnterpriseResource.removePromoImage)
--- a/app/controllers/api/enterprise_attachment_controller.rb
+++ b/app/controllers/api/enterprise_attachment_controller.rb
@@ -1,0 +1,37 @@
+module Api
+  class EnterpriseAttachmentController < BaseController
+    class MissingImplementationError < StandardError; end
+    class UnknownEnterpriseAuthorizationActionError < StandardError; end
+
+    before_filter :load_enterprise
+
+    respond_to :json
+
+    def destroy
+      return respond_with_conflict(error: destroy_attachment_does_not_exist_error_message) unless @enterprise.public_send("#{attachment_name}?")
+
+      @enterprise.update_attributes!(attachment_name => nil)
+      render json: @enterprise, serializer: Admin::EnterpriseSerializer, spree_current_user: spree_current_user
+    end
+
+    protected
+
+    def attachment_name
+      raise MissingImplementationError, "Method attachment_name should be defined"
+    end
+
+    def enterprise_authorize_action
+      raise MissingImplementationError, "Method enterprise_authorize_action should be defined"
+    end
+
+    def load_enterprise
+      @enterprise = Enterprise.find_by_permalink(params[:enterprise_id].to_s)
+      raise UnknownEnterpriseAuthorizationActionError if enterprise_authorize_action.blank?
+      authorize!(enterprise_authorize_action, @enterprise)
+    end
+
+    def destroy_attachment_does_not_exist_error_message
+      I18n.t("api.enterprise_#{attachment_name}.destroy_attachment_does_not_exist")
+    end
+  end
+end
--- a/app/serializers/api/admin/enterprise_serializer.rb
+++ b/app/serializers/api/admin/enterprise_serializer.rb
@@ -5,10 +5,19 @@ class Api::Admin::EnterpriseSerializer < ActiveModel::Serializer
   attributes :preferred_product_selection_from_inventory_only
   attributes :owner, :contact, :users, :tag_groups, :default_tag_group
   attributes :require_login, :allow_guest_orders, :allow_order_changes
+  attributes :logo, :promo_image
 
   has_one :owner, serializer: Api::Admin::UserSerializer
   has_many :users, serializer: Api::Admin::UserSerializer
   has_one :address, serializer: Api::AddressSerializer
+
+  def logo
+    attachment_urls(object.logo, [:thumb, :small, :medium])
+  end
+
+  def promo_image
+    attachment_urls(object.promo_image, [:thumb, :medium, :large])
+  end
 
   def tag_groups
     object.tag_rules.prioritised.reject(&:is_default).each_with_object([]) do |tag_rule, tag_groups|
@@ -32,5 +41,25 @@ class Api::Admin::EnterpriseSerializer < ActiveModel::Serializer
       return tag_group if tag_group[:tags].length == tags.length && (tag_group[:tags] & tags) == tag_group[:tags]
     end
     return { tags: tags, rules: [] }
+  end
+
+  private
+
+  # Returns a hash of URLs for specified versions of an attachment.
+  #
+  # Example:
+  #
+  #   attachment_urls(object.logo, [:thumb, :small, :medium])
+  #   # {
+  #   #   thumb: LOGO_THUMB_URL,
+  #   #   small: LOGO_SMALL_URL,
+  #   #   medium: LOGO_MEDIUM_URL
+  #   # }
+  def attachment_urls(attachment, versions)
+    return unless attachment.exists?
+
+    versions.each_with_object({}) do |version, urls|
+      urls[version] = attachment.url(version)
+    end
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -95,7 +95,11 @@ Openfoodnetwork::Application.routes.draw do
       post :update_image, on: :member
       get :managed, on: :collection
       get :accessible, on: :collection
+
+      resource :logo, only: [:destroy]
+      resource :promo_image, only: [:destroy]
     end
+
     resources :order_cycles do
       get :managed, on: :collection
       get :accessible, on: :collection
--- a/spec/serializers/admin/enterprise_serializer_spec.rb
+++ b/spec/serializers/admin/enterprise_serializer_spec.rb
@@ -1,9 +1,61 @@
-require 'spec_helper'
+require "spec_helper"
 
 describe Api::Admin::EnterpriseSerializer do
   let(:enterprise) { create(:distributor_enterprise) }
   it "serializes an enterprise" do
     serializer = Api::Admin::EnterpriseSerializer.new enterprise 
     serializer.to_json.should match enterprise.name  
+  end
+
+  context "for logo" do
+    let(:enterprise) { create(:distributor_enterprise, logo: image) }
+
+    context "when there is a logo" do
+      let(:image) do
+        image_path = File.open(Rails.root.join("app", "assets", "images", "logo-black.png"))
+        Rack::Test::UploadedFile.new(image_path, "image/png")
+      end
+
+      it "includes URLs of image versions" do
+        serializer = Api::Admin::EnterpriseSerializer.new(enterprise)
+        expect(serializer.as_json[:logo]).to_not be_blank
+        expect(serializer.as_json[:logo][:medium]).to match(/logo-black.png/)
+      end
+    end
+
+    context "when there is no logo" do
+      let(:image) { nil }
+
+      it "includes URLs of image versions" do
+        serializer = Api::Admin::EnterpriseSerializer.new(enterprise)
+        expect(serializer.as_json[:logo]).to be_blank
+      end
+    end
+  end
+
+  context "for promo image" do
+    let(:enterprise) { create(:distributor_enterprise, promo_image: image) }
+
+    context "when there is a promo image" do
+      let(:image) do
+        image_path = File.open(Rails.root.join("app", "assets", "images", "logo-black.png"))
+        Rack::Test::UploadedFile.new(image_path, "image/png")
+      end
+
+      it "includes URLs of image versions" do
+        serializer = Api::Admin::EnterpriseSerializer.new(enterprise)
+        expect(serializer.as_json[:promo_image]).to_not be_blank
+        expect(serializer.as_json[:promo_image][:medium]).to match(/logo-black.jpg/)
+      end
+    end
+
+    context "when there is no promo image" do
+      let(:image) { nil }
+
+      it "includes URLs of image versions" do
+        serializer = Api::Admin::EnterpriseSerializer.new(enterprise)
+        expect(serializer.as_json[:promo_image]).to be_nil
+      end
+    end
   end
 end
--- a/app/assets/javascripts/admin/resources/resources/enterprise_resource.js.coffee
+++ b/app/assets/javascripts/admin/resources/resources/enterprise_resource.js.coffee
@@ -8,4 +8,14 @@ angular.module("admin.resources").factory 'EnterpriseResource', ($resource) ->
       isArray: true
     'update':
       method: 'PUT'
+    'removeLogo':
+      url: '/admin/enterprises/:id/images/:action.json'
+      method: 'DELETE'
+      params:
+        action: 'remove_logo'
+    'removePromoImage':
+      url: '/admin/enterprises/:id/images/:action.json'
+      method: 'DELETE'
+      params:
+        action: 'remove_promo_image'
   })
--- a/app/assets/javascripts/admin/resources/resources/enterprise_resource.js.coffee
+++ b/app/assets/javascripts/admin/resources/resources/enterprise_resource.js.coffee
@@ -9,13 +9,9 @@ angular.module("admin.resources").factory 'EnterpriseResource', ($resource) ->
     'update':
       method: 'PUT'
     'removeLogo':
-      url: '/admin/enterprises/:id/images/:action.json'
+      url: '/api/enterprises/:id/logo.json'
       method: 'DELETE'
-      params:
-        action: 'remove_logo'
     'removePromoImage':
-      url: '/admin/enterprises/:id/images/:action.json'
+      url: '/api/enterprises/:id/promo_image.json'
       method: 'DELETE'
-      params:
-        action: 'remove_promo_image'
   })
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -9,5 +9,9 @@ module Api
     include ActionController::UrlFor
     include Rails.application.routes.url_helpers
     use_renderers :json
+
+    def respond_with_conflict(json_hash)
+      render json: json_hash, status: :conflict
+    end
   end
 end
--- a/app/controllers/api/logos_controller.rb
+++ b/app/controllers/api/logos_controller.rb
@@ -1,0 +1,16 @@
+module Api
+  class LogosController < EnterpriseAttachmentController
+    private
+
+    def attachment_name
+      :logo
+    end
+
+    def enterprise_authorize_action
+      case action_name.to_sym
+      when :destroy
+        :remove_logo
+      end
+    end
+  end
+end
--- a/app/controllers/api/promo_images_controller.rb
+++ b/app/controllers/api/promo_images_controller.rb
@@ -1,0 +1,16 @@
+module Api
+  class PromoImagesController < EnterpriseAttachmentController
+    private
+
+    def attachment_name
+      :promo_image
+    end
+
+    def enterprise_authorize_action
+      case action_name.to_sym
+      when :destroy
+        :remove_promo_image
+      end
+    end
+  end
+end
--- a/app/controllers/spree/admin/base_controller_decorator.rb
+++ b/app/controllers/spree/admin/base_controller_decorator.rb
@@ -28,7 +28,11 @@ Spree::Admin::BaseController.class_eval do
       record = self.class.to_s.sub("Controller", "").underscore.split('/').last.singularize.to_sym
     end
     authorize! :admin, record
-    authorize! action, record
+    authorize! resource_authorize_action, record
+  end
+
+  def resource_authorize_action
+    action
   end
 
   # This is in Spree::Core::ControllerHelpers::Auth
--- a/app/models/spree/ability_decorator.rb
+++ b/app/models/spree/ability_decorator.rb
@@ -99,7 +99,7 @@ class AbilityDecorator
     end
 
     can [:admin, :index, :create], Enterprise
-    can [:read, :edit, :update, :bulk_update, :resend_confirmation], Enterprise do |enterprise|
+    can [:read, :edit, :update, :remove_logo, :remove_promo_image, :bulk_update, :resend_confirmation], Enterprise do |enterprise|
       OpenFoodNetwork::Permissions.new(user).editable_enterprises.include? enterprise
     end
     can [:welcome, :register], Enterprise do |enterprise|
--- a/app/views/admin/enterprises/form/_images.html.haml
+++ b/app/views/admin/enterprises/form/_images.html.haml
@@ -1,12 +1,13 @@
-.row
+.row.page-admin-enterprises-form__logo-field-group.image-field-group
   .alpha.three.columns
     = f.label :logo
     %br
     100 x 100 pixels
   .omega.eight.columns
-    = image_tag @object.logo(:medium) if @object.logo.present?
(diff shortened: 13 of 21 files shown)


myriamboure:
What a good test descriptions @kristinalim ! You rock :-)

kristinalim:
@myriamboure I'm trying. :grinning: Thanks.

app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
```diff
@@ -22,24 +22,26 @@ angular.module("admin.enterprises")
       $window.location = destination
 
     $scope.removeLogo = ->
-      Enterprises.removeLogo($scope.Enterprise).then (data) ->
-        $scope.Enterprise = angular.copy(data)
-        $scope.$emit('enterprise:updated', $scope.Enterprise)
+      if confirm(t('admin.enterprises.remove_logo.immediate_removal_warning'))
```
luisramos0: you can say I am crazy and ignore but I always prefer less indentation:
if !confirm return
do_what_the_method_is_supposed_to_do_with_no_indentation

app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
```diff
@@ -21,6 +21,26 @@ angular.module("admin.enterprises")
...
+        $scope.$emit('enterprise:updated', $scope.Enterprise)
+
+        StatusMessage.display('success', t('admin.enterprises.remove_logo.removed_logo_successfully'))
+      , (response) ->
+        if response.data.error?
+          StatusMessage.display('failure', response.data.error)
+
+    $scope.removePromoImage = ->
+      Enterprises.removePromoImage($scope.Enterprise).then (data) ->
+        $scope.Enterprise = angular.copy(data)
```
luisramos0: I am curious, why do you new this copy?

app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
```diff
@@ -21,6 +21,26 @@ angular.module("admin.enterprises")
     $scope.cancel = (destination) ->
       $window.location = destination
 
+    $scope.removeLogo = ->
```
luisramos0: These methods are not the most important on this file so I'd move them further down on the file, I'd put them at the end of the file actually.

app/controllers/admin/enterprise_images_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Admin
```
luisramos0: I think this should be an API endpoint, not an admin endpoint. This controller should be on /api. Actually, I think it should be on /api/enterprises folder.

app/assets/javascripts/admin/resources/services/enterprises.js.coffee
```diff
@@ -38,3 +38,23 @@ angular.module("admin.resources").factory 'Enterprises', ($q, EnterpriseResource
 
     resetAttribute: (enterprise, attribute) ->
       enterprise[attribute] = @pristineByID[enterprise.id][attribute]
+
+    removeLogo: (enterprise) ->
```
luisramos0: this is copy pasted twice from save, you can remove that with something like this:
callFunctionOnEnterpriseResource( EnterpriseResource.$update, enterprise );
callFunctionOnEnterpriseResource( EnterpriseResource.removeLogo, enterprise );
callFunctionOnEnterpriseResource( EnterpriseResource.removePromoImage, enterprise );

and then something like:

callFunctionOnEnterpriseResource: (function, enterprise) ->
      deferred = $q.defer()
      function.apply( null, ($update({id: enterprise.permalink})
      .then( (data) =>
        @pristineByID[enterprise.id] = angular.copy(enterprise)
        deferred.resolve(data)
      ).catch (response) ->
        deferred.reject(response))
      deferred.promise

Do you want to give it a try?

app/controllers/admin/enterprise_images_controller.rb
mkllnk: @luisramos0 Do you mean `/admin/api` or just `/api`? I never thought about the mix of admin and customer code in the API before. I guess it would make sense to have *one* `/api` for everything that is really consistent. Would that be its own engine later on?

kristinalim: I checked just now, and it seems that some admin actions for products already use `/api`:

```bash
$ git grep /api app/assets/javascripts/admin/
app/assets/javascripts/admin/bulk_product_update.js.coffee:137:          url: "/api/products/" + product.id + "/soft_delete"
app/assets/javascripts/admin/bulk_product_update.js.coffee:152:              url: "/api/products/" + product.permalink_live + "/variants/" + variant.id + "/soft_delete"
app/assets/javascripts/admin/index_utils/services/resources.js.coffee:2:  LineItem = $resource '/api/orders/:order_number/line_items/:line_item_id.json',
app/assets/javascripts/admin/index_utils/services/spree_api_auth.js.coffee:6:      $http.get("/api/users/authorise_api?token=" + SpreeApiKey)
app/assets/javascripts/admin/products/services/product_image_service.js.coffee:11:      @imageUploader.url = "/api/product_images/#{product.id}"
app/assets/javascripts/admin/services/bulk_products.js.coffee:10:      url = "/api/products/bulk_products?page=::page::;per_page=20;#{queryString}"
app/assets/javascripts/admin/services/bulk_products.js.coffee:15:      $http.post("/api/products/" + product.id + "/clone").success (data) =>
app/assets/javascripts/admin/services/bulk_products.js.coffee:16:        dataFetcher("/api/products/" + data.id + "?template=bulk_show").then (newProduct) =>
app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee:46:    url = "/api/products/overridable?page=::page::;per_page=100"
```

Should we already be building actions within `/api` as much as possible? I think that would be ideal, to take more advantage of using Angular and having an API.

kristinalim:
Thanks for the feedback, @luisramos0 and @mkllnk! :slightly_smiling_face: I'll rebase and make follow-up changes.

app/controllers/admin/enterprise_images_controller.rb
luisramos0: I think /api is better at this stage.
In terms of engines, this is a good example where admin is not appropriate because this is called from the frontoffice. Anyway, this code would go into the "Entities" domain  (Entity Management/Enterprises).

My proposal (in line with Enrico's suggestion) is to have the API code inside each engine. So each engine will have it's own models, controllers, views, api and angular UI, including some rails views and it's html controllers as well as api controllers and angular code.

In terms of api endpoints and engines we can have two options:
- option 1 - engine name on the api endpoint - /entities/api/enterprises or /web/api/content/ or /shop/api/cart, entities, web and shop being the domains.
- option 2 - no engine name on the api endpoint - all engines make endpoints at /api like /api/enterprises /api/content /api/cart

imo, I think we should aim for the first option with domain name on api endpoints but I don't think we should go for it right now. So, I think /entities/api/enterprises is better but for now I'd put it as /api/enterprises.

I'll put this conversation on the domains thread.

mkllnk: Okay, let's go with `/api/`. We still have the option between `/api/enterprise_image` and `/api/enterprise/:id/logo`. The latter seems better structured to me. To really follow Rails' RESTful approach, we would have one controller for the logo and one for the promo image.

:thinking: 

Looking at the current logic, we could create a new model `EnterpriseAttachment` which contains the logic that is currently in protected methods (load enterprise and delete attachment). With such a model, it could be better to have an API like `/api/enterprise/:id/attachment/:name`. That leaves us with only one controller that can handle logos and promo images and other future attachments.

What do you think @kristinalim? I'm not sure if we need the additional model. It could just help to keep the controller even smaller.

@luisramos0 Regarding the implications of the new engines approach: This is an interesting example. My approach would probably be to first think of the best API. If it makes sense to have domains like `shop` and `web` in the API then it would make sense to structure our code accordingly. But in the end, the API should be something stable that is independent from our internal code organisation. So our code structure should not define the API.

You mentioned that engines would have their own models. That was one of my questions about this structure. Wouldn't that lead to a lot of code duplication? I thought the idea of models is that they represent the data to manipulate which is independent of the perspective from which we look at them. It's probably best to continue this discussion on Discourse.

kristinalim: @mkllnk Thank you! I prefer your first suggestion.

About using `/api/enterprise/:id/attachment/:name`, I have hesitations about abstracting the logo and promo image this way, which almost puts them on the same level of importance as each other and implying that they would always behave similarly.

I prefer think of them as two different resources which just happen to have the same type of storage and the same available actions. So, I'm more interested in your first suggestion, where we would have separate singular resource controllers for the logo and for the promo image, and then we can just have these inherit from an `EnterpriseAttachmentController` which would have most of the shared logic.

app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
kristinalim: @luisramos0 I did this to be safe and make sure changes to `$scope.Enterprise` do not affect other callbacks, considering `$scope.Enterprise` is bound to the view and we are caching a pristine copy of the enterprise in `Enterprises`.

```ruby
var deferred = $q.defer();

deferred.promise.then(function(data) {
  $scope.first = data;
});

deferred.promise.then(function(data) {
  data.value = "updated value";
  $scope.second = data;
});

deferred.resolve({value: "original value"});

// $scope.first.value => "updated value"
// $scope.second.value => "updated value"
```

app/assets/javascripts/admin/resources/services/enterprises.js.coffee
kristinalim: Nice suggestion, @luisramos0. I did something patterned after this for the two methods, and it works great.

Regarding `Enterprises.save`, I didn't refactor this yet because it uses `angular.copy(enterprise)` (as opposed to `angular.copy(data)`) which means that it would cache whatever has been loaded/changed in the JS controller:

```js
save: (enterprise) ->
  deferred = $q.defer()
  enterprise.$update({id: enterprise.permalink})
  .then( (data) =>
    @pristineByID[enterprise.id] = angular.copy(enterprise)
    deferred.resolve(data)
  ).catch (response) ->
    deferred.reject(response)
  deferred.promise
```

I suspect this is because `admin/enterprises#update` uses `Api::Admin::IndexEnterpriseSerializer` not `Api::Admin::EnterpriseSerializer`, so the returned data is incomplete.

I'm not yet comfortable enough with the code, but it seems to me that it is appropriate to have the server-side action use the non-index serializer. Do you see possible issues from doing this, aside from the difference in performance?..

It would also make sure that the object stored in `@pristineByID` is exactly what is persisted in the server, after any server-side validation or changes.

kristinalim:
Really sorry about the delay! @mkllnk @luisramos0 I've already made the changes, and this is now ready for another round of reviews. Thank you!

app/controllers/api/enterprise_attachment_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    before_filter :load_enterprise
+
+    respond_to :json
+
+    def destroy
+      return respond_with_conflict(error: destroy_attachment_does_not_exist_error_message) unless @enterprise.send("#{attachment_name}?")
+
+      @enterprise.update_attributes!(attachment_name => nil)
+      render json: @enterprise, serializer: Api::Admin::EnterpriseSerializer, spree_current_user: spree_current_user
```
sauloperez: being picky here, but the `Api` namespace is redundant here since this code is defined within `module Api`.

app/controllers/api/enterprise_attachment_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  class EnterpriseAttachmentController < BaseController
+    class MissingImplementationError < StandardError; end
+    class UnknownEnterpriseAuthorizationActionError < StandardError; end
+
+    before_filter :load_enterprise
+
+    respond_to :json
+
+    def destroy
+      return respond_with_conflict(error: destroy_attachment_does_not_exist_error_message) unless @enterprise.send("#{attachment_name}?")
```
sauloperez: we should favor `public_send` over `send` as the former respects private methods while the latter does not.

app/serializers/api/admin/enterprise_serializer.rb
```diff
@@ -33,4 +42,14 @@ def find_match(tag_groups, tags)
     end
     return { tags: tags, rules: [] }
   end
+
+  private
+
+  def attachment_urls(attachment, versions)
```
sauloperez: I'd be great to have some doc to better understand what this method returns.

config/routes.rb
```diff
@@ -95,7 +95,11 @@
       post :update_image, on: :member
       get :managed, on: :collection
       get :accessible, on: :collection
+
+      resource :logo, controller: "enterprise_logo", only: [:destroy]
+      resource :promo_image, controller: "enterprise_promo_image", only: [:destroy]
```
sauloperez: ❤️ restful routes! why don't you follow the Rails convention regarding the controller name though?

spec/controllers/api/enterprise_logo_controller_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      before do
+        allow(controller).to receive(:spree_current_user) { current_user }
+      end
+
+      context "as manager" do
+        let(:current_user) { enterprise_manager }
+
+        it "removes logo" do
+          expect(enterprise.logo?).to be true
```
sauloperez: I'm not sure this check is worth the added test suite execution time. We should trust the state we set up in line 19 and if that fails, we will notice way before seeing this test fail, IMO.

spec/controllers/api/enterprise_promo_image_controller_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      before do
+        allow(controller).to receive(:spree_current_user) { current_user }
+      end
+
+      context "as manager" do
+        let(:current_user) { enterprise_manager }
+
+        it "removes promo image" do
+          expect(enterprise.promo_image?).to be true
```
sauloperez: same concern here. All these checks do add up on our already 🐌 test suite.

spec/serializers/admin/enterprise_serializer_spec.rb
```diff
@@ -1,9 +1,61 @@
...
+
+    context "when there is a logo" do
+      let(:image) do
+        image_path = File.open(Rails.root.join("app", "assets", "images", "logo-black.png"))
+        Rack::Test::UploadedFile.new(image_path, "image/png")
+      end
+
+      it "includes URLs of image versions" do
+        serializer = Api::Admin::EnterpriseSerializer.new(enterprise)
+        expect(serializer.as_json[:logo]).to_not be_blank
```
sauloperez: this line seems redundant to me, given the line below.

spec/serializers/admin/enterprise_serializer_spec.rb
```diff
@@ -1,9 +1,61 @@
...
+
+    context "when there is a promo image" do
+      let(:image) do
+        image_path = File.open(Rails.root.join("app", "assets", "images", "logo-black.png"))
+        Rack::Test::UploadedFile.new(image_path, "image/png")
+      end
+
+      it "includes URLs of image versions" do
+        serializer = Api::Admin::EnterpriseSerializer.new(enterprise)
+        expect(serializer.as_json[:promo_image]).to_not be_blank
```
sauloperez: this one as well

spec/controllers/api/enterprise_logo_controller_spec.rb
kristinalim: Hmm... You're right, I agree.

spec/serializers/admin/enterprise_serializer_spec.rb
kristinalim: Its main purpose is to test that the JSON provides logo information, before testing that it includes the thumbnail URLs. But this also makes sure that the example fails with "`serializer.as_json[:logo]` is not blank" if it is `nil`, not a missing method error in the next assertion.

Sounds reasonable enough?

config/routes.rb
kristinalim: @sauloperez With that, do you mean that this should use the plural form? I'll make that change.

Or did you also mean that the controller name should be "logo" ("logos" pending change :point_up:) not "enterprise_logo"?

app/controllers/api/enterprise_attachment_controller.rb
luisramos0: Maybe this can be a convention to be added here? https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code-Conventions

app/assets/javascripts/admin/resources/services/enterprises.js.coffee
```diff
@@ -38,3 +38,17 @@ angular.module("admin.resources").factory 'Enterprises', ($q, EnterpriseResource
...
+        deferred = $q.defer()
+        resourceAction({id: enterprise.permalink}, ((data) =>
+          @pristineByID[enterprise.id] = angular.copy(data)
+          deferred.resolve(data)
+        ), ((response) ->
+          deferred.reject(response)
+        ))
+        deferred.promise
+
+    removeLogo: functionToPerformActionOnEnterpriseResource(EnterpriseResource.removeLogo)
```
luisramos0: yes 👍  I'd call it performActionOnEnterpriseResource

and then something like
save: performActionOnEnterpriseResource(EnterpriseResource.$update) ?

app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
```diff
@@ -69,6 +69,8 @@ angular.module("admin.enterprises")
       $scope.newUser = $scope.invite_errors = $scope.invite_success = null
 
     $scope.removeLogo = ->
+      return unless confirm(t("admin.enterprises.remove_logo.immediate_removal_warning"))
```
luisramos0: :+1:

config/routes.rb
sauloperez: I would simply remove the `controller` argument and 💯 with whatever might be the Rails convention. This makes things simpler in the long run.

app/controllers/api/enterprise_attachment_controller.rb
sauloperez: nice one @luisramos0 ! There must be a Rubocop's cop for that though...

sauloperez: 😎 https://github.com/rubocop-hq/rubocop/pull/2081 Rubocop never let's you down

kristinalim: Noting that, @sauloperez. I've never used `public_send` before, actually.

config/routes.rb
kristinalim: I renamed the controllers to `LogosController` and `PromoImagesController`, but kept the subclass `EnterpriseAttachmentController`.

Before, I was concerned that `Logos` and `PromoImages` would be vague, but maybe not. If there would be a naming clash in the future, we can rename them then.

app/assets/javascripts/admin/resources/services/enterprises.js.coffee
kristinalim: Haha! That function name does sound awkward. I wanted to distinguish between:

```javascript
return function() {
  performAction()
}
```

And:

```javascript
performAction()
```

I can update it.

About `EnterpriseResource.$update`, the existing `EnterpriseResource.$update` differs from `removeLogo` and `removePromoImage` in that the existing code doesn't actually use the enterprise object provided by the API endpoint. [A bit more about it here.](https://github.com/openfoodfoundation/openfoodnetwork/pull/2512#discussion_r212593094)

I'll look into it a bit more to see if it's safe to have `admin/enterprises#update` to return `Api::Admin::EnterpriseSerializer` instead of `Api::Admin::IndexEnterpriseSerializer`. Would you happen to know if there are possible authorization related issues?

app/assets/javascripts/admin/resources/services/enterprises.js.coffee
luisramos0: The JS code is exactly the same even if then the things going through it are different If you see a risk in changing it, it's fine! Lets leave it as it is. I'd not change server side just because of this. I was just suggesting possible improvements. Also, function names are my own preference/suggestion, feel free to ignore.

kristinalim: @luisramos0 The enterprise settings page loads data with `EnterpriseSerializer`.

`Enterprises.save` caches `angular.copy(enterprise)` (UI-updated copy) and disregards the returned value from the API, which seems well because `admin/enterprises#update` uses `IndexEnterpriseSerializer` not `EnterpriseSerializer`.

The `remove_*` endpoints uses `EnterpriseSerializer`, and update the cached enterprise with `angular.copy(data)` (fresh API copy).

Thanks. :slightly_smiling_face: Happy if you're okay with leaving this as is for now.

The function name I renamed as you suggested, because I don't remember ever seeing the format I used being used - I was just nitpicking when I wrote the code.

kristinalim:
I've made changes. Pinging @sauloperez, in case you have further comments. Otherwise, I think this is good to go. :slightly_smiling_face: Thanks!

app/assets/javascripts/admin/resources/services/enterprises.js.coffee
luisramos0: awesome, thanks for the detailed explanation!

mkllnk:
I think there is an asset reference missing. Trying to edit an enterprise in staging environment raises this one:
```
gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:142:in `digest_for':  isn't precompiled (ActionView::Template::Error)
    from gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:151:in `rewrite_asset_path'
    from gems/actionpack-3.2.22.5/lib/action_view/asset_paths.rb:27:in `compute_public_path'
    from gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:56:in `asset_path'
    from gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:62:in `image_path'
    from gems/actionpack-3.2.22.5/lib/action_view/helpers/asset_tag_helper.rb:359:in `image_tag'
    from app/views/admin/enterprises/form/_images.html.haml:7:in `_f44f45e5449ed80044497b9b0fc95e6b'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:145:in `block in render'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:161:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:143:in `render'
    from bundler/gems/deface-1110a1336252/lib/deface/action_view_extensions.rb:41:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:265:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:238:in `block in render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications/instrumenter.rb:20:in `instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:237:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/renderer.rb:48:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/helpers/rendering_helper.rb:29:in `render'
    from gems/haml-4.0.4/lib/haml/helpers/action_view_mods.rb:10:in `block in render_with_haml'
    from gems/haml-4.0.4/lib/haml/helpers.rb:89:in `non_haml'
    from gems/haml-4.0.4/lib/haml/helpers/action_view_mods.rb:10:in `render_with_haml'
    from app/views/admin/enterprises/_form.html.haml:32:in `_e80070c97af2dbdd58c381414dc05574'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:145:in `block in render'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:161:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:143:in `render'
    from bundler/gems/deface-1110a1336252/lib/deface/action_view_extensions.rb:41:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:265:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:238:in `block in render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications/instrumenter.rb:20:in `instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:237:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/renderer.rb:48:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/helpers/rendering_helper.rb:29:in `render'
    from gems/haml-4.0.4/lib/haml/helpers/action_view_mods.rb:10:in `block in render_with_haml'

Caused by: gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:142:in `digest_for':  isn't precompiled (Sprockets::Helpers::RailsHelper::AssetPaths::AssetNotPrecompiledError)
    from gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:151:in `rewrite_asset_path'
    from gems/actionpack-3.2.22.5/lib/action_view/asset_paths.rb:27:in `compute_public_path'
    from gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:56:in `asset_path'
    from gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:62:in `image_path'
    from gems/actionpack-3.2.22.5/lib/action_view/helpers/asset_tag_helper.rb:359:in `image_tag'
    from app/views/admin/enterprises/form/_images.html.haml:7:in `_f44f45e5449ed80044497b9b0fc95e6b'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:145:in `block in render'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:161:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:143:in `render'
    from bundler/gems/deface-1110a1336252/lib/deface/action_view_extensions.rb:41:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:265:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:238:in `block in render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications/instrumenter.rb:20:in `instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:237:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/renderer.rb:48:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/helpers/rendering_helper.rb:29:in `render'
    from gems/haml-4.0.4/lib/haml/helpers/action_view_mods.rb:10:in `block in render_with_haml'
    from gems/haml-4.0.4/lib/haml/helpers.rb:89:in `non_haml'
    from gems/haml-4.0.4/lib/haml/helpers/action_view_mods.rb:10:in `render_with_haml'
    from app/views/admin/enterprises/_form.html.haml:32:in `_e80070c97af2dbdd58c381414dc05574'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:145:in `block in render'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:161:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/template.rb:143:in `render'
    from bundler/gems/deface-1110a1336252/lib/deface/action_view_extensions.rb:41:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:265:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:238:in `block in render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `block in instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications/instrumenter.rb:20:in `instrument'
    from gems/skylight-1.6.1/lib/skylight/vendor/active_support/notifications.rb:159:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:38:in `instrument'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/partial_renderer.rb:237:in `render'
    from gems/actionpack-3.2.22.5/lib/action_view/renderer/renderer.rb:48:in `render_partial'
    from gems/actionpack-3.2.22.5/lib/action_view/helpers/rendering_helper.rb:29:in `render'
    from gems/haml-4.0.4/lib/haml/helpers/action_view_mods.rb:10:in `block in render_with_haml'
```

kristinalim:
Yikes! Thanks, @mkllnk. I appended a commit for easier review. This now inserts the preview image HTML directly instead of using `image_tag ''`, which is complaining that the image file `''` does not exist.

kristinalim:
I'm moving this to "Code Review" again.

sstead:
Testing notes

Yay people can now remove their images :) No issues spotted

https://docs.google.com/document/d/1jM2utpzSt-9EeZWlp2oZxOPP1fVbih_bykMp0YXJPQM/edit#

PR #2516, opened 2018-08-06
Separate admin and Spree routes from main file

kristinalim:
WARNING: This could break other pull requests that change `config/routes.rb`.

#### What? Why?

In #2512, Code Climate has started [warning us](https://codeclimate.com/github/openfoodfoundation/openfoodnetwork/pull/2512) about the length of `config/routes.rb`.

To make routes management more manageable, I separated the admin and Spree engine routes and moved them to `config/routes/`.

#### What should we test?

Testing that the routes are loaded correctly should already be covered by feature specs.

But this potentially messes up reloading of routes in `development`. To check, I did the following: (Note: Only developers would be able to test this.)

1. Add logging to the start of each route file, for example:
    ```ruby
    Rails.logger.info "-- Loading config/routes.rb"
    ```
2. Load home page to load all route files. You should see the following in the log:
    ```ruby
    -- Loading config/routes.rb
    -- Loading config/routes/admin.rb
    -- Loading config/routes/spree.rb
    ```
3. In one of the three route files, add a route, for example:
    ```ruby
    get :testmain, to: "home#index"
    ```
4. Test the newly added route, e.g. `/testmain`. This should load the assigned page/action.
5. Check the log. The application should have reloaded all route files in the correct order, so you should be able to see the following again:
    ```ruby
    -- Loading config/routes.rb
    -- Loading config/routes/admin.rb
    -- Loading config/routes/spree.rb
    ```
6. Repeat Items 3, 4, and 5 for the two other route files.

#### Release notes

- Split routes into three separate route files.

Changelog Category: Changed

If the Spree upgrade includes a major upgrade of Rails, there might be a need to update the way that the other route files are loaded.


daniellemoorhead:
I'm assuming this PR doesn't change the UI @kristinalim @mkllnk @HugsDaniel @luisramos0 and therefore won't be tested in the usual way? How do we go about getting this tested and merged?

mkllnk:

There should be no UI changes. This just requires a sanity check. Sally usually tests that the checkout works. It would also be good to check that you can log into the admin panel and that the dashboard looks the same as usual. That's all.

myriamboure:
I've not been testing for a while, I take my turn on this one ;-)

myriamboure:
Testing notes: https://docs.google.com/document/d/1M5OGr5MHc0wZjfMzvs9OvmhZePzI-QqcGvYPsFSSFBw/edit?usp=sharing

Only minor thing I saw was that in product page some images were not displayed properly, but I don’t think it is related to that PR, @maikel you can check before merging.
All seems good, but I didn’t test stripe checkout though, @sstead I didn't know which enterprise to use to test that, I have no idea on which one you created a stripe test account. And it's really time consuming to recreate those kind of things each time we want to do a sanity check, so I didn't do it. That's why it would be great to have common seeds data and enrich them step by step with our need... will argue on that direction in the relevant discussion.
If you think there is any risk that it could have impacted Stripe payment @mkllnk don't merge and we'll see with Sally so she can tell us wich enterprise to use for those tests.

PR #2521, opened 2018-08-07
[OFN Domains] Breaking OFN into domains - POC cookies inside an engine

luisramos0:
#### What? Why?

POC (Proof of Concept) of the breaking OFN down in domains using Rails Engines to implement the cookies feature.

Some links for context:
- Context discussion of breaking OFN in smaller parts: [here](https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377)
- What are Rails Engines? you can check the [v3.2 guide](https://guides.rubyonrails.org/v3.2.13/engines.html).
- The cookies feature is implemented (without engines) in [this (too) big PR](https://github.com/openfoodfoundation/openfoodnetwork/pull/2411).

#### What should we test?
Tests should be the same as in PR #2411 

#### Release notes
Technical change: new rails Engine called Web representing the Web domain (to contain the Web experience part of OFN) that implements, for now, the cookies feature: cookies policy page, cookies banner and related back office configurations.

Changelog Category: Changed

#### Discourse thread
https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377

#### Dependencies
(depended on PR #2411, which is now merged)


luisramos0:
I am really excited about this PR! Rails Engines.... it works!

The folder engines/web will contain all the stuff for the Web domain (see [here](https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377) for a list of possible domains). For now, it contains the cookies stuff. Other engines could be checkout, catalog, orders (order management), entities (entity management), etc.

A note about dependencies between main app and engine:
- the main app will depend on the engine, that's fine. In this case, we have the footer.html with a dependency to cookies-banner directive, which is fine and a dependency to the Web::CookiesConsent service (also on the web engine) which is also ok imo. So, these dependencies are ok. At some point the footer itself can be moved to the web engine.

imo, it looks great!!! The separate routes, the separate controllers and the separate assets... nice.

I have only two concerns that you may be able to help:
-   how to have footer.html call Web::CookiesConsent without having to put the require statement on the BaseController? (the footer is shown in quite a few contexts)
- I have not managed to make angular-rails-templates work inside the engine.
    - I have decided to implement a workaround, the /templates route with a TemplatesController that enables angular directives (app/assets/javascripts) to use templates stored in the rails views folder (app/views).
     - Any ideas how to make angular-rails-templates work? 
Including it on the all.js file like on the main app:
//= require_tree ../templates
//= require angular-rails-templates
This inserts the html directly on the all.js file and breaks it completely.

app/views/layouts/darkswarm.html.haml
```diff
@@ -19,6 +19,8 @@
     %script{src: "//maps.googleapis.com/maps/api/js?libraries=places,geometry#{ ENV['GOOGLE_MAPS_API_KEY'] ? '&key=' + ENV['GOOGLE_MAPS_API_KEY'] : ''} "}
     = split_stylesheet_link_tag "darkswarm/all"
     = javascript_include_tag "darkswarm/all"
+    = split_stylesheet_link_tag "web/all"
```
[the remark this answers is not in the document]
luisramos0: Yes. If we break the frontoffice in 3 domains/engines. We would have 3 includes here. We need to see how we manage common stuff but to keep it simple I'd say for now we keep darkswarm for code that is used in multiple engines.

app/controllers/base_controller.rb
```diff
@@ -1,5 +1,6 @@
 require 'spree/core/controller_helpers/respond_with_decorator'
 require 'open_food_network/tag_rule_applicator'
+require 'web/cookies_consent'
```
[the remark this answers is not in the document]
luisramos0: In rails 3 all helpers are available to all views, SO, yes, I can remove from base_controller and create an cookies_consent_helper that just has that require.

app/views/shared/_footer.html.haml
```diff
@@ -140,7 +140,7 @@
           = t '.footer_legal_text_html', {content_license: link_to('CC BY-SA 3.0', 'https://creativecommons.org/licenses/by-sa/3.0/'), code_license: link_to('AGPL 3', 'https://tldrlegal.com/license/gnu-affero-general-public-license-v3-(agpl-3.0)' )}
         %p.text-small
           %div
-            - cookies_policy_link = link_to( t( '.footer_data_cookies_policy' ), '', 'cookies-policy-modal' => true, 'cookies-banner' => !CookiesConsent.new(cookies, request.host).exists? && Spree::Config.cookies_consent_banner_toggle)
+            - cookies_policy_link = link_to( t( '.footer_data_cookies_policy' ), '', 'cookies-policy-modal' => true, 'cookies-banner' => !Web::CookiesConsent.new(cookies, request.host).exists? && Spree::Config.cookies_consent_banner_toggle)
```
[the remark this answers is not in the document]
luisramos0: yes, I have moved this into an helper, the same helper where I require CookieConsent. I think it's much better now.
We could require Spree in the engine, for now, because the helper is for the footer that is in the main app, I'd leave the helper in the main app. I think it's fine. When we move the footer to the web engine, we can move the helper as well.

luisramos0:
I have rebased this PR with the now merged cookies work. So now it's only one commit. We can work from this commit now.

engines/web/app/assets/stylesheets/web/cookies_banner.css.scss
```diff
@@ -1,4 +1,4 @@
-@import 'branding';
+@import '../../../../../../app/assets/stylesheets/darkswarm/branding';
```
luisramos0: the question here is: how do we manage common css (as here) and common JS?
Is there an easy way (and not as ugly way as above) to make engines use css and js in the main app?
Or do we need to extract common css/js code to a new lib to be imported by the engines?

mkllnk: Copied to the community forum: https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377/16

luisramos0:
I have rebased this PR to fetch the cookies_spec fix and make this build green.

engines/web/app/assets/stylesheets/web/cookies_banner.css.scss
sauloperez: I'd say this is good enough at least for now. We'll see what these common dependencies are.

engines/web/test/test_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+
+Rails.backtrace_cleaner.remove_silencers!
+
+# Load support files
+Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].each { |f| require f }
+
+# Load fixtures from the engine
+if ActiveSupport::TestCase.method_defined?(:fixture_path=)
+  ActiveSupport::TestCase.fixture_path = File.expand_path("fixtures", __dir__)
+end
```
sauloperez: If I'm not mistaken this refers to Rails testing fixtures and we don't use them, so we should be able to remove the block.

engines/web/test/web_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class WebTest < ActiveSupport::TestCase
+  test "truth" do
+    assert_kind_of Module, Web
+  end
+end
```
sauloperez: 🔥 I guess this is just an example of a test autogenerated by Rails.

engines/web/web.gemspec
```diff
@@ -0,0 +1,17 @@
...
+  s.version     = Web::VERSION
+  s.authors     = ["developers@ofn"]
+  s.summary     = "Web domain of the OFN solution."
+
+  s.files = Dir["{app,config,db,lib}/**/*"] + ["LICENSE.txt", "Rakefile", "README.rdoc"]
+  s.test_files = Dir["test/**/*"]
+
+  s.add_dependency "rails", "~> 3.2.22.5"
+
+  s.add_development_dependency "sqlite3"
```
sauloperez: IMO this should be `pg` instead.

Gemfile
```diff
@@ -10,6 +10,8 @@ gem 'i18n-js', '~> 3.0.0'
 # Patched version. See http://rubysec.com/advisories/CVE-2015-5312/.
 gem 'nokogiri', '>= 1.6.7.1'
 
+gem 'web', :path => './engines/web'
```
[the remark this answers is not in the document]
luisramos0: well spotted! rubocup didnt get this one probably because its in the gemfile.

engines/web/Gemfile
```diff
@@ -0,0 +1,11 @@
...
+
+# Declare your gem's dependencies in web.gemspec.
+# Bundler will treat runtime dependencies like base dependencies, and
+# development dependencies will be added by default to the :development group.
+gemspec
+
+# Declare any dependencies that are still in development here instead of in
+# your gemspec. These might include edge Rails or gems from your path or
+# Git. Remember to move these dependencies to your gemspec before releasing
+# your gem to rubygems.org.
```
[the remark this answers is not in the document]
luisramos0: I wasn't sure if we could do that. Looks like we can. Ive removed both Gemfile and Gemfile.lock from the engine.

engines/web/test/test_helper.rb
luisramos0: :+1: done

engines/web/test/web_test.rb
luisramos0: yep, deleted.

engines/web/web.gemspec
luisramos0: I have removed this entry and the rails one as well. This was auto generated. I dont think we need this.

luisramos0:
@sauloperez @mkllnk 
Can we move forward with this?
After this is merged, I am looking forward preparing the next PR with some AR models 🎉

mkllnk:
I'm still not 100% sold on this. It's an experiment for OFN. I see the chances for better structured code, but I can also see possible pitfalls. So I would like the majority of OFN devs being onboard with this. Everybody has to understand what we are doing here, otherwise it will just be chaos. A state of incomplete refactoring is confusing.

luisramos0:
I understand your point @mkllnk. That was discussed in the gathering.
This is mostly @enricostano and @sauloperez idea, I am just implementing it.
I don't think this is an experiment, this is the first step of a big (and very good and much needed) change in OFN.

I think you mention 2 very important points:
- "incomplete refactoring is confusing" - yes, when you start a big change like this one it's is confusing and the solution is simple: that confusion needs to be addressed - what exactly are the doubts you have? We need of address each of those. We have the discourse page for that and maybe we can start with a wiki page to clearly explain how to do things. I dont think we need to close/define everything up-front, just enough.
- all devs need to be onboard - I totally agree! @Matt-Yorkley, @HugsDaniel and @kristinalim  can you guys please review, understand and ask questions if you have about this change? Also, @enricostano would you be able to provide feedback here? (Thanks!)

I propose 2 actions from this:
- every dev should review this PR and clarify questions
- we should create a wiki page with description of each question/challenge and proposed approach

Also, what do you think if I prepare another PR with a more advanced case?

app/views/shared/_footer.html.haml
```diff
@@ -140,8 +140,6 @@
           = t '.footer_legal_text_html', {content_license: link_to('CC BY-SA 3.0', 'https://creativecommons.org/licenses/by-sa/3.0/'), code_license: link_to('AGPL 3', 'https://tldrlegal.com/license/gnu-affero-general-public-license-v3-(agpl-3.0)' )}
         %p.text-small
           %div
-            - cookies_policy_link = link_to( t( '.footer_data_cookies_policy' ), '', 'cookies-policy-modal' => true, 'cookies-banner' => !CookiesConsent.new(cookies, request.host).exists? && Spree::Config.cookies_consent_banner_toggle)
-            - privacy_policy_link = link_to( t( '.footer_data_privacy_policy' ), Spree::Config.privacy_policy_url, :target => '_blank' )
```
Matt-Yorkley: I see you've deleted these two lines and added methods for these in the `FooterLinksHelper` module, but won't you still need to call those methods from here instead of removing them entirely?

engines/web/app/assets/stylesheets/web/cookies_banner.css.scss
Matt-Yorkley: Surely there must be a nicer way to do this with Rails engines?

engines/web/app/controllers/web/application_controller.rb
```diff
@@ -0,0 +1,5 @@
+module Web
+  class ApplicationController < ActionController::Base
+    protect_from_forgery with: :exception
```
Matt-Yorkley: Why does this need to be overridden instead of using the regular ApplicationController value?

engines/web/app/controllers/web/application_controller.rb
```diff
@@ -0,0 +1,5 @@
+module Web
+  class ApplicationController < ActionController::Base
```
Matt-Yorkley: So is the purpose of this to allow access from within the engine to the helpers attached to the main ApplicationController?

Matt-Yorkley:
So there's no changes to the specs? Is that right?

engines/web/app/controllers/web/application_controller.rb
Matt-Yorkley: Ok, I guess I'm a bit confused about how this and the main ApplicationController will work together.

Matt-Yorkley:
I agree with Maikel that it will be messy unless the whole app is reworked to use engines in a similar way.

Matt-Yorkley:
Are there still issues with angular templates that need to be resolved?

app/views/shared/_footer.html.haml
luisramos0: if I understand your doubt, yes, they are used in the lines below, right?

engines/web/app/controllers/web/application_controller.rb
luisramos0: the easiest way to think about an engine is to think of it as a separate app (from the guide "Engines can be considered miniature applications").
we can get to the engine code through requiring its modules and classes but the main way to get to engine will be through its routes (there's a new entry on the main app routes to include all the routes of this engine), so this is a different app and so we extend ActionController::Base.

Actually, this is a good example of where we will get the opportunity to better organize our code. If you look at the main ApplicationController you see it's contains code relate to  embedded_shopfront code, this code can be moved to this Web module ApplicationController and removed from the main app and thus, cleaning up this code from the other spaces, like the unrelated backoffice code.

the reference for engines is this: https://guides.rubyonrails.org/engines.html

engines/web/app/controllers/web/application_controller.rb
luisramos0: yeah, this is a separate app, I dont think we will inherit from the main app app_controller. We want this app to be a different and separate app. We will need to see what we do what common code, but I dont think it should be through extending the main app_controller.

luisramos0:
Hey @Matt-Yorkley thanks a lot for your quick feedback!

>> So there's no changes to the specs? Is that right? Or would you need to restructure those as well?

There was much restructuring of code here, I just moved the files from main app to this engine. Mostly, that's what this approach will require, moving files to their domain, and in all cases possible, include the new domain namespace.
The same for specs, we will move them to the specific domains. For generic specs that cover several parts of the app like the cookies one, I'd keep them in the main app for now.

>> Are there still issues with angular templates that need to be resolved?

If we keep the approach of having views/angular-templates, there's nothing to be done. btw, the issue is not related to this change, it's related to a problem I introduced by using Spree::Config stuff on angular templates, inside assets...

>> .  I agree with Maikel that it will be messy unless the whole app is reworked to use engines in a similar way.

Yes, that is the objective, this is the first step in that direction. We will agree on how to move things and start to gradually migrate everything to engines. I think the beginning is the most confusing, because we all need to understand how it will work.

Thanks again for feedback!

engines/web/app/assets/stylesheets/web/cookies_banner.css.scss
kristinalim: With SASS, we should be able to do something this:

**darkswarm/all.scss**:
```sass
@include "../web/all.scss";
```

**branding.scss**:
```sass
$ofn-web-cookies-banner-background-color: $dark-grey;
```

**web/base/colors.scss**:
```sass
$ofn-web-cookies-banner-background-color: #fff !default;
```

**web/all.scss**:
```sass
@import 'web/base/colors';
@import 'web/cookies_banner.scss';
```

AFAIK, this is how UI frameworks like Foundation manage to get applications to customize variables outside their SASS directory tree.

We don't need to share CSS among engines much though, right? For CSS, as far as I understand, almost all would be under the `web` engine.

kristinalim: Also, the JS ideally would only help provide the UI - all/most of the business logic would belong to the other engines. So, I'm with the opinion that all of the JS should belong to the `web` engine too.

Sorry if I missed any discussion about this, or obvious reasons why this should not be the case.

app/views/layouts/darkswarm.html.haml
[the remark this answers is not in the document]
kristinalim: I wrote a related comment [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2521#discussion_r216161437). I agree with what @mkllnk said: "everything is in the web engine" So not 3 includes but 1 include, which would be for `web`.

engines/web/app/assets/stylesheets/web/cookies_banner.css.scss
luisramos0: Hey @kristinalim
Thanks a lot for your quick feedback!
re scss suggestion, looks great! I think I know what you mean, I'll give it a try and let you guys know here if I can make it work.

re "the JS should belong to the web engine too". The domains are not a tech division of the app, the domains are a business division of the app. They are small standalone applications. "Web" stands for Content, it's the static part of the frontoffice (front page, menu, footer, etc), it does not include all the shop and checkout pages, neither all the backoffice. Certainly it can include frontoffice common stuff, the domains shop and checkout can depend on Web for this (is this what you meant?), or maybe easier, the common JS can remain in the main app for now and then be extracted to a library/lib/engine.

luisramos0:
Good, thanks @mkllnk for going through this!

>>> Circular dependencies

Do we agree this is not about models, controllers and views (these will be in each domain), this is about other common code in ruby, js and css used in multiple parts that don't fit any domain.
There are 3 approaches to this:
1 . For example, as discussed above, for CSS code, we can decide we keep it in the Web domain. It becomes the domain responsible for the frontoffice style and the other frontoffice domains (shop and checkout) depend on it for common CSS.
2. For JS and ruby, we should probably create separate libraries/engines, for example, to keep common code like I18nHelper or AngularFormHelper.
3. But to start with, I think the easiest way is to make these circular dependencies and keep them in the main app. I think this is acceptable as we start, it's simple, you don't have to do anything. I'd vote for leaving common stuff in the main app (for now) and later (we can decide when it's the best time to do it) extract these common parts to separate libraries/engines.

>>> (Code) duplication

I don't think there will much more than this engine scaffolding done on this PR and a few more things. Are you worried about any particular part?

>>> Another question came up when I read your replies to Kristina: In the end, would the main app still have code or would everything be in engines?

I think that we should extract all logic to domains and common code to libraries, I think the end game should be no-main-app apart from the scaffolding (GemFile, routes, etc).

engines/web/README.rdoc
```diff
@@ -0,0 +1,3 @@
+= Web
+
+This is the rails engine for the WEB OFN domain.
```
mkllnk: This would be a good place to a wiki page explaining why we use engines and how they are organised. Or maybe a readme in engines?

mkllnk:
> Circular dependencies

Your approaches seem reasonable. It's still not great, but a trade off we have to make if we want engines.

> duplication

I had nothing in particular on my mind (except the scaffolding). Can we reduce the scaffolding? If the license is the same, we can just have one file in the main app, right?

engines/web/README.rdoc
luisramos0: maybe the readme pointing to the wiki.
I'll do that.

engines/web/app/assets/stylesheets/web/cookies_banner.css.scss
kristinalim: @luisramos0 Let me know if you've worked on the SCSS changes already. I can create a PR with this branch as base - I already did a little POC so have some messy changes which I can clean up.

luisramos0: hey, thanks @kristinalim I accept your offer! I am currently "entangled" on the spree upgrade...

kristinalim: @luisramos0 I submitted a PR to your repository: https://github.com/luisramos0/openfoodnetwork/pull/2 :slightly_smiling_face:

luisramos0: Amazing, thanks so much!!!
It is a much better solution! I'll merge.
I just have one question there. I think we have an option/decision to make. See [here](https://github.com/luisramos0/openfoodnetwork/pull/2/commits/cdabeda2d9bcb02b331bef3c58c1d049f764f109#r217545997).

luisramos0:
Ok guys, thanks for your feedback!
I have rebased and resolved conflicts and I have added a link to the wiki in the engine README (as suggested by Maikel), this is the [new wiki page](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Tech-Doc:-How-OFN-is-organized-in-Domains-using-Rails-Engines).

I have added a few new commits:
- [this commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/dc5eb6448e5d76ea579130d8584f55770aaefce5) uses @kritinalim ’s wisdom from [this PR](https://github.com/luisramos0/openfoodnetwork/pull/2) and fixes the bad looking @import '../../../../../../app/assets/stylesheets/darkswarm/branding’
     - I have adapted [this commit](https://github.com/luisramos0/openfoodnetwork/pull/2/commits/34562c5a5ac91f52ddd2891a6639e73389a3c4ea) and [this commit](https://github.com/luisramos0/openfoodnetwork/pull/2/commits/66a7c089c558063be79aa5e091e943ab116d4ce6) from @kristinalim 's PR
    - Thanks Kristina, I think it looks really good like this!
    - (in this PR Kristina also suggested a new way of managing the colour palete, I’ll create a new PR in master to make this happen)
- [this commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/c22ac0086b99e706890ad40f439e9b5bae0cf65e) moves cookies_policy_helper and its spec to the engine
- [this commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/bded530137c2cb7ec85bdea06e17bc737721560c) moved api endpoint (cookies/consent into from /api to /content/api . This way each domain will have it’s own API namespace.

This is ready for another round of reviews.

luisramos0:
sauloperez approved these changes an hour ago :tada: :tada: :tada:

mkllnk:
I'm still reluctant, but if everybody else is on board then let's try it. We need to modularise the OFN somehow. Worst case: We find that it creates more work, doesn't solve our problem of too many responsibilities in one app and copy everything into the same folders again. Amalgamating engines sounds like an easy task.

luisramos0:
yeah, that's true, it's easy to revert.

"too many responsibilities in one app" makes me think that this PR in itself is already solving that by having cookies code 95% out of main app.

Thanks for your advise, questions and consent @mkllnk 
I am really looking forward the next steps :-)

luisramos0:
all right, thanks @Matt-Yorkley 
Moving to test ready.

luisramos0:
So, I am not able to make this work in staging...

I fixed one issue with precompilation (it's solved) with [this commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/b1c7e6c0916f604572a2fb3712bb469f3f78292d) and while trying to fix my main issue (see below) I improved two things by [isolating namespace](https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/a23b1b980da5cf2eedd1b3fcab270537f972b63a) and [moving angular-templates into the engine](https://github.com/openfoodfoundation/openfoodnetwork/pull/2521/commits/921105301ce51aaab18a0693a9087468d33b57db).

The main issue is that the engine routes are just not working in staging. The exact same code works well locally. So I have created 2 dummy routes to test this /test_main_app and /test_engine similar routes, one in the main app and one inside the engine. The engines one doesnt work in staging.

Help Needed: can someone please fetch this branch and try to run it locally and in some other staging server?
After deploying this branch you can:
- browse to <your domain>/test_main_app
- browse to <your domain>/test_engine
-  open the main page and the click on cookies policy link on the footer (watch dev tools console)

Thanks!

mkllnk:
I just deployed this to https://staging.openfoodnetwork.org.au/ and it seems to work. I didn't use ofn-install though.

mkllnk:
Okay, I just tried it with ofn-install as well and it worked. I'm not sure what's wrong with the French staging server.

luisramos0:
wow, nice. I have no idea what could cause this behaviour... maybe some webserver config in staging FR...

Anyway, moving forward. I have removed the commit with the test routes and fixed a bug I found on slow page loads.
@mkllnk can you please stage this last version so it can be tested?
Thank you!

luisramos0:
@sauloperez @Matt-Yorkley this can also be staged in uk or es if you have time and a free staging server. thanks!

Matt-Yorkley:
Staged on: https://staging.openfoodnetwork.org.uk @luisramos0

luisramos0:
Thanks Matt!

I tested it in detail. It's all working fine!
I just found a typo on the matomo optout iframe URL. This one letter change commit fixes it. The PR needs to be restaged. Thanks!

Matt-Yorkley:
Restaged on: https://staging.openfoodnetwork.org.uk

luisramos0:
Thanks again @Matt-Yorkley 
I tested it now, all good. Definitely ready for testing :-)
https://staging.openfoodnetwork.org.uk

PR #2528, opened 2018-08-08
The main menu is now translatable and configurable (per instance)

### LINKED ISSUE #2410: Enable instances to add menus in the main header (with redirects)
## Description
Given conversation in Slack https://openfoodnetwork.slack.com/archives/C0DNLAZC7/p1529691706000289 there is a lazy consent to do a proper implementation of "adding more menu headers with redirects" feature to avoid a hack more and more instance are doing.

## Expected Behavior
Instance can add more menus headers in the top menu of an OFN website menu

## Actual Behavior
Hack done for example in UK (but also Aus): See here for the code changes commit in the UK
https://github.com/openfoodfoundation/openfoodnetwork/commit/90d5a04368e229fc849a386b91d3284dd893702d
And then you need to create config/initializers/feature_toggle.rb and add the appropriate code to enable it. UK example is: 

require 'open_food_network/feature_toggle'

module OpenFoodNetwork
 FeatureToggle.class_eval do
   private

   def self.features
     {order_cycles: true,
      about_blog_homepage: true}
   end
 end
end

## Context
Now France is willing to do the same so again the hack need is happening, as it's a quick fix we propose to do it properly.

## Possible Fix
Add some backoffice fields that, if you fill in with URL, you get the menu on your instance.

UPDATE Luis 8/8/2018:
The feature implemented includes 7 new translation pairs (title and url) for each menu entry and then a checkbox on the Content configuration to enable/disable these menu entries. The mobile menu icons can also be defined in the content configuration (see screenshots below).

What needs to be tested:
- we need to set a few variation son the translations files and then activate/deactivate different menu entries and see if both large menu and mobile menu work properly with different type of links.

luisramos0:
Added 7 menu entries to the translations file (each with text and url) and 7 content configuration points to enable/disable these. The mobile menu icons can also be configured.

#### What? Why?

Closes #2410 

#### What should we test?
See #2410 for screenshots.
We need to set a few variations on the translations files and then activate/deactivate different menu entries and see if both large menu and mobile menu work properly with different type of links.

#### Release notes
The main menu is now translatable and configurable (per instance).

Changelog Category: Added

### DIFF
--- a/app/models/content_configuration.rb
+++ b/app/models/content_configuration.rb
@@ -33,19 +33,19 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
 
   # Main URLs
   preference :menu_1, :boolean, default: true
-  preference :menu_1_icon_name, :string, default: ""
+  preference :menu_1_icon_name, :string, default: "ofn-i_019-map-pin"
   preference :menu_2, :boolean, default: true
-  preference :menu_2_icon_name, :string, default: ""
+  preference :menu_2_icon_name, :string, default: "ofn-i_037-map"
   preference :menu_3, :boolean, default: true
-  preference :menu_3_icon_name, :string, default: ""
+  preference :menu_3_icon_name, :string, default: "ofn-i_036-producers"
   preference :menu_4, :boolean, default: true
-  preference :menu_4_icon_name, :string, default: ""
+  preference :menu_4_icon_name, :string, default: "ofn-i_035-groups"
   preference :menu_5, :boolean, default: true
-  preference :menu_5_icon_name, :string, default: ""
-  preference :menu_6, :boolean, default: true
-  preference :menu_6_icon_name, :string, default: ""
-  preference :menu_7, :boolean, default: true
-  preference :menu_7_icon_name, :string, default: ""
+  preference :menu_5_icon_name, :string, default: "ofn-i_013-help"
+  preference :menu_6, :boolean, default: false
+  preference :menu_6_icon_name, :string, default: "ofn-i_035-groups"
+  preference :menu_7, :boolean, default: false
+  preference :menu_7_icon_name, :string, default: "ofn-i_013-help"
 
   # Footer
   preference :footer_logo, :file
--- a/app/models/content_configuration.rb
+++ b/app/models/content_configuration.rb
@@ -31,6 +31,22 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
   preference :group_signup_case_studies_html, :text, default: I18n.t(:content_configuration_case_studies)
   preference :group_signup_detail_html, :text, default: I18n.t(:content_configuration_detail)
 
+  # Main URLs
+  preference :menu_1, :boolean, default: true
+  preference :menu_1_icon_name, :string, default: ""
+  preference :menu_2, :boolean, default: true
+  preference :menu_2_icon_name, :string, default: ""
+  preference :menu_3, :boolean, default: true
+  preference :menu_3_icon_name, :string, default: ""
+  preference :menu_4, :boolean, default: true
+  preference :menu_4_icon_name, :string, default: ""
+  preference :menu_5, :boolean, default: true
+  preference :menu_5_icon_name, :string, default: ""
+  preference :menu_6, :boolean, default: true
+  preference :menu_6_icon_name, :string, default: ""
+  preference :menu_7, :boolean, default: true
+  preference :menu_7_icon_name, :string, default: ""
+
   # Footer
   preference :footer_logo, :file
   has_attached_file :footer_logo, default_url: "/assets/ofn-logo-footer.png"
--- a/app/views/admin/contents/_fieldset.html.haml
+++ b/app/views/admin/contents/_fieldset.html.haml
@@ -2,7 +2,11 @@
   %legend{align: "center"}= name
   - preferences.each do |key|
     - type = ContentConfig.preference_type(key)
+    - if name == t('admin.contents.edit.main_links') && type == :boolean
+      - text = t("#{key}_title") + " - " + t("#{key}_url")
+    - else
+      - text = t(key)
     .field
-      = label_tag(key, t(key) + ': ') + tag(:br) if type != :boolean
+      = label_tag(key, text + ': ') + tag(:br) if type != :boolean
       = preference_field_tag(key, ContentConfig[key], :type => type)
-      = label_tag(key, t(key)) + tag(:br) if type == :boolean
+      = label_tag(key, text) + tag(:br) if type == :boolean
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1175,14 +1175,14 @@ en:
   menu_2_url: "/map"
   menu_3_title: "Producers"
   menu_3_url: "/producers"
-  menu_4_title: "Connect"
-  menu_4_url: "https://openfoodnetwork.org/au/connect/"
-  menu_5_title: "Learn"
-  menu_5_url: "https://openfoodnetwork.org/au/learn/"
-  menu_6_title: "Blog"
-  menu_6_url: ""
-  menu_7_title: "Groups"
-  menu_7_url: ""
+  menu_4_title: "Groups"
+  menu_4_url: "/groups"
+  menu_5_title: "About"
+  menu_5_url: "http://www.openfoodnetwork.org/"
+  menu_6_title: "Connect"
+  menu_6_url: "https://openfoodnetwork.org/au/connect/"
+  menu_7_title: "Learn"
+  menu_7_url: "https://openfoodnetwork.org/au/learn/"
 
   logo: "Logo (640x130)" #FIXME
   logo_mobile: "Mobile logo (75x26)" #FIXME
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -428,6 +428,7 @@ en:
         producer_signup_page: Producer signup page
         hub_signup_page: Hub signup page
         group_signup_page: Group signup page
+        main_links: Main Menu Links
         footer_and_external_links: Footer and External Links
         your_content: Your content
 
@@ -1167,6 +1168,21 @@ en:
   ticket_column_item: "Item"
   ticket_column_unit_price: "Unit Price"
   ticket_column_total_price: "Total Price"
+
+  menu_1_title: "Shops"
+  menu_1_url: "/shops"
+  menu_2_title: "Map"
+  menu_2_url: "/map"
+  menu_3_title: "Producers"
+  menu_3_url: "/producers"
+  menu_4_title: "Connect"
+  menu_4_url: "https://openfoodnetwork.org/au/connect/"
+  menu_5_title: "Learn"
+  menu_5_url: "https://openfoodnetwork.org/au/learn/"
+  menu_6_title: "Blog"
+  menu_6_url: ""
+  menu_7_title: "Groups"
+  menu_7_url: ""
 
   logo: "Logo (640x130)" #FIXME
   logo_mobile: "Mobile logo (75x26)" #FIXME
--- a/app/controllers/admin/contents_controller.rb
+++ b/app/controllers/admin/contents_controller.rb
@@ -6,6 +6,7 @@ module Admin
                               {name: I18n.t('admin.contents.edit.producer_signup_page'), preferences: [:producer_signup_pricing_table_html, :producer_signup_case_studies_html, :producer_signup_detail_html]},
                               {name: I18n.t('admin.contents.edit.hub_signup_page'), preferences: [:hub_signup_pricing_table_html, :hub_signup_case_studies_html, :hub_signup_detail_html]},
                               {name: I18n.t('admin.contents.edit.group_signup_page'), preferences: [:group_signup_pricing_table_html, :group_signup_case_studies_html, :group_signup_detail_html]},
+                              {name: I18n.t('admin.contents.edit.main_links'), preferences: [:menu_1, :menu_1_icon_name, :menu_2, :menu_2_icon_name, :menu_3, :menu_3_icon_name, :menu_4, :menu_4_icon_name, :menu_5, :menu_5_icon_name, :menu_6, :menu_6_icon_name, :menu_7, :menu_7_icon_name]},
                               {name: I18n.t('admin.contents.edit.footer_and_external_links'), preferences: [:footer_logo,
                                                              :footer_facebook_url, :footer_twitter_url, :footer_instagram_url, :footer_linkedin_url, :footer_googleplus_url, :footer_pinterest_url,
                                                              :footer_email, :community_forum_url, :footer_links_md, :footer_about_url]}]
--- a/app/views/shared/menu/_large_menu.html.haml
+++ b/app/views/shared/menu/_large_menu.html.haml
@@ -11,54 +11,13 @@
           %a{href: '/'}
             = t 'title'
     %ul.center
-      %li
-        %a{href: main_app.shops_path}
-          %span.nav-primary
-            = t 'label_shops'
-      %li
-        %a{href: main_app.map_path}
-          %span.nav-primary
-            = t 'label_map'
-      %li
-        %a{href: main_app.producers_path}
-          %span.nav-primary
-            = t 'label_producers'
-      - if feature? :connect_learn_homepage
-        %li
-          %a{href: "https://openfoodnetwork.org/au/connect/"}
-            %span.nav-primary
-              = t 'label_connect'
-        %li
-          %a{href: "https://openfoodnetwork.org/au/learn/"}
-            %span.nav-primary
-              = t 'label_learn'
-      - elsif feature? :about_blog_homepage
-        %li
-          %a{href: main_app.groups_path}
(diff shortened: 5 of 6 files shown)


app/views/admin/contents/_fieldset.html.haml
```diff
@@ -2,7 +2,11 @@
   %legend{align: "center"}= name
   - preferences.each do |key|
     - type = ContentConfig.preference_type(key)
+    - if name == t('admin.contents.edit.main_links') && type == :boolean
```
mkllnk: Having a branch in a loop is a bit ugly. I'm also struggling to understand it. So the translation here has to match the translation in the ContentsController. And then you have a different label depending on the type and if it's a special menu item checkbox.

It does look really nice on your screenshot. I'm just wondering if we can simplify this. But that would probably involve having our own model for preferences which decorates the Spree preferences. Probably out of scope of this pull request.

app/models/content_configuration.rb
```diff
@@ -31,6 +31,22 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
   preference :group_signup_case_studies_html, :text, default: I18n.t(:content_configuration_case_studies)
   preference :group_signup_detail_html, :text, default: I18n.t(:content_configuration_detail)
 
+  # Main URLs
+  preference :menu_1, :boolean, default: true
+  preference :menu_1_icon_name, :string, default: ""
```
mkllnk: Do we need some default here? I guess in most instances, the fallback locale will be `en` which always contains values. Just if somebody configured a different fallback locale and hasn't completed the translation yet, this could be a problem. That sounds very unlikely. So I guess this can stay as is.

app/views/admin/contents/_fieldset.html.haml
luisramos0: Yes! This isn't nice but it's the best I could do without creating a new configuration screen, new preferences, etc...
this IF reads: "if it's a checkbox in the content main links, make the checkbox text be the title and the url merged"

luisramos0:
Thanks a lot for reviewing @mkllnk

config/locales/en.yml
```diff
@@ -1132,6 +1133,21 @@ en:
...
+  menu_1_title: "Shops"
+  menu_1_url: "/shops"
+  menu_2_title: "Map"
+  menu_2_url: "/map"
+  menu_3_title: "Producers"
+  menu_3_url: "/producers"
+  menu_4_title: "Connect"
+  menu_4_url: "https://openfoodnetwork.org/au/connect/"
+  menu_5_title: "Learn"
+  menu_5_url: "https://openfoodnetwork.org/au/learn/"
```
Matt-Yorkley: I don't think these Aus links should be in the default settings for the base `en` locale. Hmmm... is the issue that Aus uses `en` without a separate language file?

app/models/content_configuration.rb
```diff
@@ -31,6 +31,22 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
...
+  preference :menu_2_icon_name, :string, default: ""
+  preference :menu_3, :boolean, default: true
+  preference :menu_3_icon_name, :string, default: ""
+  preference :menu_4, :boolean, default: true
+  preference :menu_4_icon_name, :string, default: ""
+  preference :menu_5, :boolean, default: true
+  preference :menu_5_icon_name, :string, default: ""
+  preference :menu_6, :boolean, default: true
+  preference :menu_6_icon_name, :string, default: ""
+  preference :menu_7, :boolean, default: true
```
Matt-Yorkley: From the other file it looks like items 4 and 5 are Aus-specific links, and items 6 and 7 have blank URLs by default. Should all of these menu item preferences default to being enabled in that case?

config/locales/en.yml
```diff
@@ -1132,6 +1133,21 @@ en:
...
+  menu_6_title: "Blog"
+  menu_6_url: ""
+  menu_7_title: "Groups"
+  menu_7_url: ""
```
Matt-Yorkley: I've just looked through again and I think Groups does have a default URL across all instances. It should be: `menu_7_url: "/groups"`.

app/models/content_configuration.rb
Matt-Yorkley: So I think menu items 1, 2, 3, 4 and 7 are default for all instances, 4 and 5 are just for Aus, and 6 is different for each instance but optional.

Matt-Yorkley: Also, the default menu items do have default icons already, so those can be added here under the `*_icon_name` preferences.

config/locales/en.yml
luisramos0: yes, as far as I understand, yes, this is the AUS config file :-)

app/models/content_configuration.rb
luisramos0: thanks @Matt-Yorkley . I can do that, but I wouldn't do it.
I left the Aus stuff in en.yml because I don't think we have other option, but in terms of preferences defaults in the code, I'd leave everything empty/true. It's much cleaner, each instance can do what they want with the menu with just a few clicks afterwards.

If you think we should do it, can you please clarify exactly what defaults would you leave?
Did you suggest having menus 1, 2, 3, 4 and 7 to true, 5 and 6 to false and icons 1, 2, 3, 4 and 7 set to the respective icons of Shops, Map, Producers and Groups?

config/locales/en.yml
luisramos0: the footer link Groups is not related to this change. the footer still has the hard coded link Groups to /groups. This is just for the main menu where for example AUS, doesn't have Groups.

luisramos0: We havent discussed the footer Navigate section that stays untouched with this PR: Shops /shops, Maps /map, Producers /producers, Groups /groups and About /<content_config_footer_about_url>.
We could do the same later with the Navigate footer links as we just did for the main menu.

config/locales/en.yml
mkllnk: Yes, one downside of this approach is that whenever Aus wants to change the menu, we need to change `en.yml` and the changes will be available for translation again. That will probably be confusing for translators.

config/locales/en.yml
Matt-Yorkley: The "Groups" menu item does have a default route, unlike "Blog". Whether or not it's enabled is handled in the other file, no?

app/models/content_configuration.rb
Matt-Yorkley: It just makes sense to me that the defaults are set for more or less "universal" settings, and any extras need to be enabled. So for items 5 and 6 which only show Aus content by default and will not be used by most instances, it seems sensible not to enable them on all instances by default.

Matt-Yorkley: I'm mostly thinking of minimising additional extra work that will have to be done on each instance here, so instance managers have an easier job. 

I would definitely add the default logos for standard menu items from the current codebase into the defaults here, otherwise every instance will need to go through the codebase to find each icon name for each of those icons and re-write it into the correct place, checking each time to see they got the numbers matched up to the correct menu item. And it would need to be re-done every time a new staging/production server is set up.

Matt-Yorkley: Lets imagine a case where this gets merged and an instance manager forgets to manually update all these items... it would be good if the menus are not broken.

luisramos0: Hi Matt, I think I understand you.
Please have a look at the new commit.
Now default menu is 5 elements (1 to 5) with default icons and default URLs as well (I set /groups as Groups default and About default as openfoodnetwork.org).
Let me know if this was you meant.
Thanks, Luis

app/models/content_configuration.rb
```diff
@@ -33,19 +33,19 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
...
+  preference :menu_1_icon_name, :string, default: "ofn-i_019-map-pin"
   preference :menu_2, :boolean, default: true
-  preference :menu_2_icon_name, :string, default: ""
+  preference :menu_2_icon_name, :string, default: "ofn-i_037-map"
   preference :menu_3, :boolean, default: true
-  preference :menu_3_icon_name, :string, default: ""
+  preference :menu_3_icon_name, :string, default: "ofn-i_036-producers"
   preference :menu_4, :boolean, default: true
-  preference :menu_4_icon_name, :string, default: ""
+  preference :menu_4_icon_name, :string, default: "ofn-i_035-groups"
```
Matt-Yorkley: Great! The one last thing I think I'd add is some small comments as a reference in this file to give anyone looking at the code a bit of help understanding which is which, so they don't have to look through the rest of the codebase for hints, eg:

``` ruby
preference :menu_1, :boolean, default: true # Shops
preference :menu_1_icon_name, :string, default: "ofn-i_019-map-pin"
preference :menu_2, :boolean, default: true # Maps
preference :menu_2_icon_name, :string, default: "ofn-i_037-map"
preference :menu_3, :boolean, default: true # Producers
```

Matt-Yorkley:
Staged on: https://staging.katuma.org :tada:

app/models/content_configuration.rb
luisramos0: I can do that but tbh I dont think it's the best approach.
They will change the en.yml entries for AUS at some point and they will not need to change this file. So I'd not link the two things as they may become outdated.
Also, on the other side, a dev changing this needs to know that they correspond to the en.yml file entries, right? :-)

luisramos0:
Great, thanks for reviewing!

luisramos0:
Hi @RachL yes, it is there with spree@.
It's in Configuration / Content / Main Menu Links.
On the admin UI we need to test the checkboxes and the icons on both desktop and mobile.

I think we should also change the en.yml file to something else to see if it is correctly reflected on the admin UI. For this, we will need to do another PR or maybe easier, we can just change it manually on the server for testing.
This is the current en.yml file: https://github.com/openfoodfoundation/openfoodnetwork/pull/2528/files/1853dc9b7a75cbc4554af9f2c902bbee2b7b976b#diff-37c9b835927e3c363588f172de9bb6faR1136

RachL:
Hi @luisramos0 thank you, I'm sorry I didn't scrolled enough on the content page yesterday.

Selecting and de-selecting the menu works fine. We just need to check the translation and we can move it to done. Can you add a portugues version of the menu with random new links on katuma staging? Do you have access?
https://docs.google.com/document/d/1xtLU6Zpugz9VLPzIwE3IQBJaxVXgZzqtZ0oM75C30r8/edit#heading=h.9dmf1ag3agsu

luisramos0:
ah, thanks for testing!

sorry, no, I don't have access to katuma staging...

myriamboure:
I have a question to @RachL and @luisramos0 does the url name also adapt? Like if I check "connect" but translate on transifex for French to "support", and do the same for "learn" with "blog", will the url still be the original aus one? Cause that remains pretty confusing to me for users... I'm a bit afraid of the answer ...

myriamboure:
Pau only comes back in 2 weeks... why do we need to test by changing English language? Anyway we won't be able to test translations before this is merged, and then multilingual switch can't be tested until we can translate the new strings... I don't get the point @luisramos0 ? We never do it for anything so why would we test translations before merging for this one?

luisramos0:
ok, I can clarify:
- "does the url name also adapt? Like if I check "connect" but translate on transifex for French to "support", and do the same for "learn" with "blog", will the url still be the original aus one?"

No, you will be able to "translate" the URL.

- "We never do it for anything so why would we test translations before merging for this one?"

For the same reason, if we change the "translation" of the URL in staging we will be able to test that both the translated menu names and the "translated" URLs are working fine.

I think @Matt-Yorkley has access to katuma staging, right?
If so, I'd go there and "translate" the menu manually. It would be a useful test I think.
The test I would do:
- go to katuma stating
- open config/locales/es.yml
- change some menu_*_title and menu_*_urls entries
- go to backoffice and configure the menu with the new entries
- verify the changes are reflected correctly on the website

to clarify:
- the backoffice config you do, will apply to all languages on your instance
- the translations will only apply to each specific language in all instances

the question is then: do we have any language used in multiple instances? (I dont think so and thus this is not a problem)
The constraint here is that each language can only have one menu list of titles/urls.
I hope this makes sense.

luisramos0:
to clarify: both @mkllnk and @sauloperez understood this from the start, that's why they said it's not perfect. It is not perfect. But it is done and you get lost of new flexibility in the menu with this.
And, it can be improved in the future :-)

myriamboure:
Ok thank you now I understand! So it's good, no language is used in multiple instance, we usually copy it and make a local version, as some things like business model content anyway vary from instance to instance. So it's all green and good, @Matt-Yorkley if you can do the server test and it works then it will be ready to go :-)

myriamboure:
Ok thank you now I understand! So it's good, no language is used in multiple instance, we usually copy it and make a local version, as some things like business model content anyway vary from instance to instance. So it's all green and good, @Matt-Yorkley if you can do the server test and it works then it will be ready to go :-)

daniellemoorhead:
@Matt-Yorkley this has been sitting in the Test Ready column a week now, can you please confirm that you can help @RachL with the Katuma staging server? @sauloperez is back at the end of this week but it would be great to move it across sooner than later.... 🙏

luisramos0:
All right. I have managed to stage this one on fr staging and run the test myself.

I gave it a try and created my first testing notes here:
https://docs.google.com/document/d/1Cz6oi4EgI8Ar3stuuRZJyvSxX3jUkd2z08b5SzGNlpM/edit#

All looks good to me. Ready to go I believe.

daniellemoorhead:
@luisramos0 @myriamboure @RachL are there any implementation super admin requirements for this when it goes into production? I ask because all PRs merged get pushed to australian production before the release occurs, so if we need to do something at this point in time then we need to know now.

Ping @mkllnk @sstead FYI

luisramos0:
Good question, each instance manager needs to validate and configure the menu on their instance after this is released.

This is the default menu that will be installed on every instance upon release:
  menu_1_title: "Shops"
  menu_1_url: "/shops"
  menu_2_title: "Map"
  menu_2_url: "/map"
  menu_3_title: "Producers"
  menu_3_url: "/producers"
  menu_4_title: "Groups"
  menu_4_url: "/groups"
  menu_5_title: "About"
  menu_5_url: "http://www.openfoodnetwork.org/"
  menu_6_title: "Connect"
  menu_6_url: "https://openfoodnetwork.org/au/connect/"
  menu_7_title: "Learn"
  menu_7_url: "https://openfoodnetwork.org/au/learn/"

And the default menu activates only the 5 first items.

## For AUS
It's the easiest because it uses en.yml, no transifex work needed.
So, after the release AUS instance admin needs to go to backoffice content configuration and uncheck menu 4 and 5 (Groups and About) and check menu 6 and 7 (Connect and Learn). That 's it.

## For other instances, for each specific language
(I am not a master os transifex but I believe I know how it works, please correct me if I am wrong)
- First, we need to merge this into master, this will make the menu entries available in transifex
- Then, someone needs to go to transifex and translate these menu entries to each language
- Then we need the transifex job to make these translations into master again (it's a transifex PR)
- And then the feature is ready to go live in that specific language

RachL:
just a small comment @luisramos0 @daniellemoorhead : I'm not sure it is a good practice that a developper test his/her own code. You might be a little biased ;) Anyway here no problem I've check as well everything looks good. Really exited that this will be soon live!

luisramos0:
oh, I am sure testing my own change is not the best practice :-)
We can hold this until another dev runs the test I did on fr staging.
Anyway, we need another dev to validate the release/transifex process I describe above.

luisramos0:
all right, @sauloperez just confirmed on slack the process I described is correct.
So, next step is a merge please :-)

sstead:
**Testing notes**

- The superadmin settings correctly change the visibility of heading links
- Heading links display correctly on mobile and desktop

One issue, the footer 'navigation' menu doesn't update to match the header menu.

https://docs.google.com/document/d/1NH6pczPglSFEUQB_0K1jyfxhS2yI_JjHyZFUpb_LC98/edit?pli=1#

daniellemoorhead:
@mkllnk the footer bug that @sstead found seems to be a problem that means this shouldn't have been merged?

Did you have a look at this?

myriamboure:
Hum... @daniellemoorhead @mkllnk I think it doesn't prevent merging as the links behind them remind valid, if I'm right. I'll check with Luis on the issue and create another issue if we want to fix it. I'm not sure we need to have again another same navigation in footer and header, what's the point?

luisramos0:
Hey, the footer is exactly as it was before this PR, it does not adapt to the menu, it was not adapting to the menu with the previous feature toggle stuff.

Is there anything else pending I can help with?

myriamboure:
No I think it's good, i don't think we should go further on this as it's just working as before with the instances hacks. it's satisfying as is.

luisramos0:
A little help with how to translate the new menu in tx:

see the filter key:menu_ and on the bottom right the context and key for the entry you are translating
screen shot 2018-08-31 at 17 39 33

daniellemoorhead:
>I'm not sure we need to have again another same navigation in footer and header, what's the point?

It's not always the same @myriamboure. For example, we only have the about link in the footer as it didn't need to be in the header.

Issue #2523, opened 2018-08-08
[Spree Upgrade] Refactor OFN shop list to use API for add line items to cart


luisramos0:
## Description
Spree 2 has changed order.add_variant and a few more things in Order and OrderPopulator. See the [checkout customizations wiki page](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Checkout-customisations) for context.

We propose to move away from extending Spree code order_decorator and order_populator_decorator and re-implement this logic in OFN.
So, there should be a new API endpoint for /cart that will take (to be confrmed): order_id, line_item_id and quantity. This will be implemented in Api::CartController. This controller should depend on new OFN service class CartService. This service will implement what's currently in order and order_populator and will depend, but not extend, Spree classes. Basically, what's in  orderpopulator.populate and order.add_variant needs to be moved and adapted.

This change must be done on OFN master branch (so that we can validate and test easily) and then merged back to spree branch.

Additionally, and to support the upgrade to spree 2.1, we should move away from order.add_variant and use add_contents instead. This can be done later. If we make the code depend on order.add_variant we can then simply move from add_variant to add_contents.

Additionally, if not too difficult we would change the frontend to stop making bulk calls with groups of line items and make one call per line item. We can keep current API with bulk to make it easier for now.

## Testing

It's a technical change, there should be no UX change in the add to cart process.

mkllnk:
Excellent, @luisramos0, that's exactly what we discussed. Looking at it now, I'm just wondering if it's too much in one step. Maybe we can break it into smaller pull requests.

1. Create `app/controllers/cart_controller.rb` that is basically doing the same as the current code. The frontend code just needs to change the URL to post updates to. As at the moment, it would always modify the current cart (Spree's current order).
2. Create an API endpoint `app/controllers/api/carts_controller.rb`. Here we expect a cart id (order id) and modify one line item at a time. Actually, we can call it `LineItemsController`. Uh, in fact, we already have a `LineItemsController` in `app/controllers`. It's just for deleting from a finalised order. We can probably build on that.

What do you think?

luisramos0:
Yes, agree, excellent.

Before I start changing things I am documenting my study of the code here:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Tech-Doc:-Customer-shopping-flow-(Orders,-Line-Items-and-Checkout)

The controllers structure is quite amusing!!! 
For example, CartController.add_variant was written in 2013 :-D 
[...]

PR #2539, opened 2018-08-11
removed Cart route, controller, model and specs: dead code

luisramos0:
#### What? Why?

Related to #2523 

This beautiful part of OFN is 5 years old, this was used in the store frontend in the past.
I suspect it's just dead code.
(we may soon get back some of this to re-implement Order and OrderPopulator outside of Spree).

#### What should we test?
Shopping process, add to cart and checkout. Nothing should change.

#### Release notes
Removed Cart related old dead code.

Changelog Category: Removed

This is related to the refactoring of the OrderPopulator out of Spree. #2523 is part of the Spree Upgrade process.


luisramos0:
313 lines of code deleted, the build is green. Looking good. I wonder what do reviewers think? :-D

My only concern is: do we have external systems/users of these /cart endpoints? I don't see any other possible problem. I may be wrong.

app/models/cart.rb
```diff
@@ -1,28 +0,0 @@
-class Cart < ActiveRecord::Base
```
mkllnk: I think removing the model should also go with a migration to remove the table. It also needs to drop the relationship between order and cart.

I just checked our production server for carts and our table is empty. So this should be safe.

I found several uses of `order.cart?`, but I suspect that it's testing whether the order is in the `cart` state and not if it has a cart. Can you double check that?

luisramos0: I doubled checked, yes, cart is one of the states of the order state machine.
All I have found on my double checking was more dead code, please see the new commit.

Matt-Yorkley:

daniellemoorhead:
@myriamboure @RachL is one of you able to test this on UK staging when you're back please? 😃

RachL:
Everything seems fine.

Here are my testing notes:
https://docs.google.com/document/d/1WXC89l-h8DUDbNBUUayb4Ox2JhrakatkeASU_lEC7_M/edit#

luisramos0:
oh man, I love this one! first thing I change on the OFN database is a drop table!!!

PR #2546, opened 2018-08-16
Remove empty SKU values if empty

### LINKED ISSUE #2501: [Product Import] Add SKU check to validation
## Description
When doing a Product Import the template includes a field for SKU
If this field is left blank the import fails
Apparently if the column is removed it will pass (I haven't checked this) but there isn't a way for user to know

## Expected Behavior
If import file contains SKU column and cells are blank, the User should be informed at validation stage and advised to either enter SKUs or remove that column from the import file

## Actual Behavior
The validation passes and shows no errors, but the import freezes at the actual import
image

Using browser tools the error can be identified as NULL in SKU values
image

When values are filled in to the import file the import works

## Steps to Reproduce
1. Create a CSV import file and leave SKU cells empty
2. Proceed through product import

## Animated Gif/Screenshot
http://recordit.co/AwTfjWdkXe

## Severity
S4 - if adequate info in user guide people can work around it

## Possible Fix
- allow null values in sku - some people might want skus on some products and not others, so keeping column but allowing empty cells probably ideal
- add to validation
- remove column from template, but then people would have to know to add it to use skus which would be annoying for some

Matt-Yorkley:
#### What? Why?

Closes #2501

Fixes a bug where validation of the SKU field when saving a product would cause import to fail if some of the entries in the spreadsheet were present and some were `nil`.

#### What should we test?

Product import works when the spreadsheet has an `sku` column present, and some values in the column are empty. 

#### Release notes

Product Import: fixed an issue with empty SKU fields causing products to not be saved. 

Changelog Category: Fixed

### DIFF
--- a/app/models/product_import/spreadsheet_entry.rb
+++ b/app/models/product_import/spreadsheet_entry.rb
@@ -14,6 +14,7 @@ module ProductImport
 
     def initialize(attrs)
       @validates_as = ''
+      remove_empty_skus attrs
       assign_units attrs
     end
 
@@ -56,6 +57,10 @@ module ProductImport
     end
 
     private
+
+    def remove_empty_skus(attrs)
+      attrs.delete('sku') if attrs.key?('sku') && attrs['sku'].blank?
+    end
 
     def assign_units(attrs)
       units = UnitConverter.new(attrs)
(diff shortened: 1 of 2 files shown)


mkllnk:
@luisramos0 If you review something and it has enough approvals to go to test-ready, you are very welcome to move the pull request there. :-)

luisramos0:
ok, thanks. I thought "Test Ready" was after the issue was staged.

mkllnk:

PR #2547, opened 2018-08-16
Pi/producer enabled

### LINKED ISSUE #2490: [Product import] validation check to see if the enterprise is enabled as a producer
## Description
Issue #2456 made it possible to not save info while doing an import and having errors. This issue aims at tackling the error of importing product to a non-producer enterprise.
Two main tasks : 
1. Check if enterprise in product file is a producer
2. If not, display an error

## Acceptance criteria
capture du 2018-07-12 09-34-10
capture du 2018-07-12 09-34-38
- If no error, you directly see the button "save" (instead of proceed which is unclear) + cancel button
- If error, you only have one "back to file upload" button

## Test cases:
- If we try to import product into a non producer enterprise, it shouldn’t work and error should be displayed

Matt-Yorkley:
#### What? Why?

Closes #2490

Adds an additional validation check to see if an enterprise is enabled as a producer with the `is_primary_producer` property, and doesn't allow importing products for that enterprise if not.

#### What should we test?

Product cannot be imported for an enterprise that is not set to `is_primary_producer`.

#### Release notes

Product Import: Added a validation check to make sure enterprise is enabled as a producer.

Changelog Category: Added

#### Documentation updates

This should be noted in the product import user guide.

### DIFF
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -62,6 +62,11 @@ module ProductImport
         return
       end
 
+      unless Enterprise.find_by_name(supplier_name).is_primary_producer?
+        mark_as_invalid(entry, attribute: "supplier", error: I18n.t(:error_not_primary_producer, name: supplier_name))
+        return
+      end
+
       entry.supplier_id = @spreadsheet_data.suppliers_index[supplier_name]
     end
 
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -17,7 +17,7 @@ module ProductImport
 
         next if entry.supplier_id.blank?
 
-        if import_into_inventory?(entry)
+        if import_into_inventory?
           producer_validation(entry)
           inventory_validation(entry)
         else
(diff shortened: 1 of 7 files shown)


app/models/product_import/entry_validator.rb
```diff
@@ -62,6 +62,11 @@ def supplier_validation(entry)
         return
       end
 
+      unless Enterprise.find_by_name(supplier_name).is_primary_producer?
```
mkllnk: Do you really need to load the enterprise from the database or is it in `@editable_enterprises` already? Ah, it's detour, you figured that yourself. :-)

myriamboure:
@Matt-Yorkley can you share what error message you displayed? It would be great to explain what is the problem ("the entreprise is not a producer so products can't be attached to them. Check the producer's profile setup!")
For instance... but maybe you already thought about that... There are so many possible types of error, but it would be great to help the user understand the error so that they can easily correct it.

mkllnk:
I got more detail of the error. The import script reports:
```
PG::Error: ERROR:  null value in column "sku" violates not-null constraint
DETAIL:  Failing row contains (31, null, 1.00, null, null, null, null, f, 19, 100, null, null, 0, f, AUD, 1000, , null, null, null).
: INSERT INTO "spree_variants" ("cost_currency", "cost_price", "count_on_hand", "deleted_at", "depth", "display_as", "display_name", "height", "import_date", "is_master", "lock_version", "on_demand", "position", "product_id", "sku", "unit_description", "unit_value", "weight", "width") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19) RETURNING "id"
```

sstead:
Testing notes

Looks good . I got an error trying to import products under an enterprise that's not a producer.

https://docs.google.com/document/d/1mkGBcmASLrWyzakECPAVcC1wMN_rOR4d26ggEDBOS8E/edit

PR #2548, opened 2018-08-16
Make OC name editable from the OC index

### LINKED ISSUE #2430: Make OC name field editable from OC index
When a user comes along to create a series of OCs to set up their subscription schedule, they need to basically replicate the same thing over and over again, but with different dates and OC name.

A tiny change would make this infinitely easier - **make the name editable from the OC Index / overview page.** Then I can set up one, hit clone as many times as I like, and update the names and open / close times without having to open / edit each order cycle itself (very time consuming) 

See also #2463 #2439 for related issues 

I think that this, combined with #2439 is a more elegant and simple solution than #1052 as it means user doesn't need to open each order cycle

kristinalim:
#### What? Why?

Closes #2430 

When creating a series of order cycles (OC), the user normally duplicates the first OC and then edits the OC name and dates of these subsequent OCs.

The OC dates can already be edited in bulk from the OC index, but the name cannot.

Before this change, the user has to go to the edit page of each OC and change the name. To save users time, the OC name has been made editable from the OC index as well. 

#### What should we test?

**As admin:**

1. Go to the OC index.
2. Duplicate an order cycle.
3. Still in the OC index, find the newly created order cycle.
4. Change the name, and open and close dates of this OC, and click "Save Changes".
5. Check that you received the/ following success alert: "Order cycles have been updated."
6. Without reloading the page yet, check that the name of the OC is still the updated name.
7. Reload the page.
8. Check that the name of the OC is still the updated name.
9. Repeat Items 2 to 8, but only updating the OC name, not the open and close dates.

**As enterprise manager:**

Repeat steps for admin above.

**As user with particular limited permissions for the enterprise:**
(User account that has an "Enterprise Permission" for the enterprise, as long as _add to order cycle_ was not granted)

1. Go to the OC index.
2. Check that the name, open date, and close date fields for the OC are not editable.

#### Release notes

- Add ability to bulk update order cycle name from the order cycle index

Changelog Category: Added

### DIFF
--- a/app/assets/javascripts/admin/resources/services/order_cycles.js.coffee
+++ b/app/assets/javascripts/admin/resources/services/order_cycles.js.coffee
@@ -66,7 +66,7 @@ angular.module("admin.resources").factory 'OrderCycles', ($q, $injector, OrderCy
       changes
 
     attrsToSave: ->
-      ['orders_open_at','orders_close_at']
+      ['name', 'orders_open_at','orders_close_at']
 
     resetAttribute: (order_cycle, attribute) ->
       order_cycle[attribute] = @pristineByID[order_cycle.id][attribute]
--- a/app/views/admin/order_cycles/_row.html.haml
+++ b/app/views/admin/order_cycles/_row.html.haml
@@ -1,7 +1,6 @@
 %tr{ class: "order-cycle-{{orderCycle.id}} {{orderCycle.status}}", ng: { repeat: 'orderCycle in orderCycles | schedule:scheduleFilter | involving:involvingFilter | filter:{name: query} track by orderCycle.id' } }
   %td.name{ ng: { show: 'columns.name.visible' } }
-    %a{ ng: { href: '{{orderCycle.edit_path}}' } }
-      {{ orderCycle.name }}
+    %input{ id: 'oc{{::orderCycle.id}}_name', name: 'oc{{::orderCycle.id}}[name]', type: 'text', ng: { model: 'orderCycle.name', disabled: '!orderCycle.viewing_as_coordinator' } }
   %td.schedules{ ng: { show: 'columns.schedules.visible' } }
     %span{ ng: { repeat: 'schedule in orderCycle.schedules'} }
       %a{ 'schedule-dialog' => true, 'schedule-id' => '{{schedule.id}}' }
--- a/spec/controllers/admin/order_cycles_controller_spec.rb
+++ b/spec/controllers/admin/order_cycles_controller_spec.rb
@@ -220,6 +220,7 @@ module Admin
         let(:params) do
           { format: :json, order_cycle_set: { collection_attributes: { '0' => {
             id: oc.id,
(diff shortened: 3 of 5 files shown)


mkllnk:

RachL:
hey @kristinalim thank you so much for your detailed testing process. I know it has already been said, however it really save me time which is cool :)

Here are my testing notes:
https://docs.google.com/document/d/1ll1nbZzc6mwmtSPB6Xtry4zbGpSFGGhyuvEERT_kjkc/edit#
All looks fine except I'm not sure how to test the last case: enterprise with limited permissions. As "add to order cycle" is not permitted, I don't know how to enable a user to see an OC that he/she cannot edit. Can you or @myriamboure or @sstead share some light on that? Thanks :)

RachL:
@myriamboure ok thanks. I've retested by granting "add to order cycle". Now I can indeed see the OC from enterprise 1 in enterprise 2 profile. But enterprise 2 can change date and name.... So I will wait for @kristinalim 's feedback : is this a good test - and in that case a bug - or is the PR description right (you can't edit name and date as long as "add to order cycle" is unchecked).

sstead:
Testing notes:

I just checked the 'As user with particular limited permissions for the enterprise' scenario. It was behaving correctly, with fields uneditable for a user who is involved in an OC through permissions but has no ownership of the coordinator enterprise.

Notes at the bottom of this doc...
https://docs.google.com/document/d/1ll1nbZzc6mwmtSPB6Xtry4zbGpSFGGhyuvEERT_kjkc/edit#

kristinalim:
@RachL Sorry about this! It seems I got confused with the code and then set up a test scenario that only by coincidence "confirmed" my wrong understanding.

The scenario which @myriamboure described is indeed supposed to show the non-editable OC name and date fields. Basically, if you are a coordinator, producer, or distributor for an OC, you should be able to see it under "Order Cycles", but only the coordinator can edit the name and dates.

(I applied the same permission check to the OC name as what was being used for the open and close dates.)

The following should test this for an existing producer and food hub:

1. Check that the food hub user which you will use is not already an authorized user for the producer, and vice versa.
2. In "Enterprise Permissions", check that there are no special permissions between the producer and the food hub yet.
3. Grant "add to order cycle" permission for the producer to the food hub.
4. Sign in as the food hub user.
5. Set up an OC where the food hub is the coordinator. Add an incoming exchange for a product of the producer.
6. Sign in as the producer user.
7. Go to the list of order cycles. Check that the OC name and open/close dates are not editable.

Please let me know if you encounter issues.

Hmm... On this page, as producer, "Loading..." actually does not disappear for me. I'm not sure if it's because I have invalid data in my local DB instance or not, but this also happens on the "master" branch. Could you let me know if this is also the case in staging? I will also check with a clean DB and create an issue if yes.

20180824093214-supplier-order-cycles-loading

kristinalim:
@sstead Saw your notes. :slightly_smiling_face: Awesome! Thank you.

PR #2556, opened 2018-08-17
Make cookies tests more stable

### LINKED ISSUE #2553: Intermittently failing cookies spec
## Description

A spec recently introduced in #2411 seems to be unreliable. I've seen this error a few times now:

```
  1) Cookies banner in the homepage does not show after cookies are accepted
     Failure/Error: expect(page).to_not have_css("button", :text => accept_cookies_button_text, :visible => true)
       expected not to find visible css "button" with text "Accept Cookies", found 1 match: "Accept Cookies"
     # ./spec/features/consumer/cookies_spec.rb:86:in `expect_not_visible_cookies_banner'
     # ./spec/features/consumer/cookies_spec.rb:13:in `block (4 levels) in <top (required)>'
```

 Example: https://travis-ci.org/mkllnk/openfoodnetwork/jobs/417116813

## Expected Behavior

Automated test always pass.

## Actual Behavior

The mentioned spec fails from time to time.

## Steps to Reproduce

1. `bundle exec rspec ./spec/features/consumer/cookies_spec.rb:11`

## Context

Testing other pull requests often results in failed tests and I need to restart the build.

## Severity

tech-debt, but quite high

## Possible Fix

I'm wondering if the accept button is clicked before the action is associated to it.

luisramos0:
#### What? Why?

Attempt to close #2553

Fixing cookies spec intermittent fails.
I think I just need a 1 sec sleep before the policy page opens.

#### What should we test?
Build doesn't intermittently fail with cookies_spec errors.

### DIFF
--- a/spec/features/consumer/cookies_spec.rb
+++ b/spec/features/consumer/cookies_spec.rb
@@ -9,7 +9,9 @@ feature "Cookies", js: true do
       end
 
       scenario "does not show after cookies are accepted" do
+        sleep 1
         click_button I18n.t('legal.cookies_banner.cookies_accept_button')
+        sleep 2
         expect_not_visible_cookies_banner
 
         visit root_path
@@ -18,6 +20,7 @@ feature "Cookies", js: true do
 
       scenario "banner contains cookies policy link that opens coookies policy page and closes banner" do
         find("p.ng-binding > a", :text => "cookies policy").click
+        sleep 1
         expect_visible_cookies_policy_page
         expect_not_visible_cookies_banner
 
--- a/spec/features/consumer/cookies_spec.rb
+++ b/spec/features/consumer/cookies_spec.rb
@@ -15,12 +15,14 @@ feature "Cookies", js: true do
         expect_not_visible_cookies_banner
 
         visit root_path
+        sleep 1
         expect_not_visible_cookies_banner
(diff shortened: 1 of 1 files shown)


mkllnk:
Nope, that doesn't fix it: https://travis-ci.org/mkllnk/openfoodnetwork/builds/418013043
I created a wip branch with Travis executing that particular spec 25 times on each of the four machines. They all failed with the same error.

Your sleep is just before testing for visibility. Usually Capybara is smart enough to wait for elements to become visible. That's why I'm thinking that the click doesn't trigger the action and that could be because the click happens before the button is initialised properly.

luisramos0:
ahah, I fixed the test that was failing for me as well :-)
We may need a few more sleeps.
I suggest we keep my sleep and add your two sleeps as well. We can then run your repeated build and merge if all is green. Ok?

luisramos0:
I have rerun this build a couple of times in this PR and it has been green.

luisramos0:
yeah, I am sorry about that! I did put 2 or 3 hours on trying to get around this, with no success...

mkllnk:
As this is a no-test change and Luis and I have been working on the solution for a while and it is confusing a lot of other pull requests, I'm merging this now.

Issue #2559, opened 2018-08-19
[Spree Upgrade] Scope Variants to VariantOverrides when order is changed


luisramos0:
Original title: InventoryUnit.increase is gone in Spree 2.0 (related to variant scoping)

OFN 2-0-STABLE build errors, for example:

```
bundle exec rspec ./spec/controllers/admin/bulk_line_items_controller_spec.rb:14

`Admin::BulkLineItemsController#index` as an administrator when no ransack params are passed in retrieves a list of `line_items` with appropriate attributes, including line items with appropriate attributes

Failure/Error: Spree::InventoryUnit.increase(order, variant, quantity)
NoMethodError: undefined method `increase' for #<Class:0x007f913c1d4488>
```

The `InventoryUnit.increase` and `decrease` methods were managing "both variant.count_on_hand and inventory unit creation". This was moved to stock locations and shipments in spree 2.
This is related to Spree 2 Multiple Shipments epic #2011.

There are two different calls to this increase method. Both for scoping of variant to hub/shop when managing stock in inventory.

### /app/models/spree/inventory_unit_decorator.rb
In InventoryUnit.class_eval - assign_opening_inventory
When finalizing an order we were extending InventoryUnit to _scope_ each variant before managing stock. 
Added statements to the original Spree code:
    scoper = OpenFoodNetwork::ScopeVariantToHub.new(order.distributor)
    scoper.scope(line_item.variant)

This assign_opening_inventory was removed and the stock movement is now handled in a new shipment.finalize where stock locations are updated (manifest_unstock).

Possible solution is to:
- drop inventory_unit_decorator
- insert the variant scoping somewhere in the shipment.finalize method

Shipment.class_eval do
    def finalize!
      --- do the scoping of variants in inventory_units here
      InventoryUnit.finalize_units!(inventory_units)
      manifest.each { |item| manifest_unstock(item) }
    end

### /app/models/spree/line_item_decorator.rb
On update_inventory and remove_inventory, the spree code is copied and the variant is scoped.

Possible solution:
- remove monkey patching line_item_decorator.rb
- patch for scoping on new class order_inventory

In both cases, the first step is to remove the existing patching with the scoping.

luisramos0:
Right now I am not sure how/where to do the variant scoping. I think it might need further group discussion. I'll raise this one in the next catch up.
Hopefully we find a better way than .class_eval...ing the spree models...

luisramos0:
[...]

PR #2560, opened 2018-08-19
Removed monkeypatchs to scope variants on inventory_unit and line_ite…

luisramos0:
…m decorators. We need to make the scoping somewhere else in spree2, see #2559

#### What? Why?

First step of #2559 

Removed existing code. Tests still failing but we can now see errors on the new code in OrderInventory.

#### What should we test?
New error in spree2 code: OrderIventory needs a default shipment defined.

This is part of the spree upgrade.

#### Discourse thread
See https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade%3A-Variant%27s-count_on_hand-migration


luisramos0:
Build error now is:
144) Admin::BulkLineItemsController#index as a normal user should deny me access to the index action
       Failure/Error: let!(:line_item1) { FactoryBot.create(:line_item, order: order1) }
       
       NoMethodError:
         undefined method `stock_location' for nil:NilClass
       # ./spec/controllers/admin/bulk_line_items_controller_spec.rb:14:in `block (3 levels) in <top (required)>'

This is happens in LineItem creation:
      def update_inventory
        Spree::OrderInventory.new(self.order).verify(self, target_shipment)
where target_shipment is empty, it needs to be populated.

PR #2561, opened 2018-08-20
Enable embedded shopping spec

### LINKED ISSUE #2233: Fix unreliable features test

mkllnk:
#### What? Why?

Closes #2233.

The spec was disabled in
https://github.com/openfoodfoundation/openfoodnetwork/pull/2232, because
it was very unreliable. But it seems stable again since
https://github.com/openfoodfoundation/openfoodnetwork/pull/2468.

#### What should we test?

This affects automated test only. No manual testing needed.

#### Release notes

Improved reliability of automated tests.

Changelog Category: Fixed

### DIFF
--- a/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
+++ b/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
@@ -43,7 +43,7 @@ feature "Using embedded shopfront functionality", js: true do
       end
     end
 
-    xit "allows shopping and checkout" do
+    it "allows shopping and checkout" do
       on_embedded_page do
         fill_in "variants[#{variant.id}]", with: 1
         wait_until_enabled 'input.add_to_cart'


luisramos0:
cookies_spec is breaking the build. a rerun should fix it.
otherwise, ready to go.

daniellemoorhead:
Semaphore still broken @mkllnk ☹️

mkllnk:
@daniellemoorhead That is confusing, yes. When you open a PR it goes into the review column automatically. I probably should have moved it back to in-dev until all automated tests passed.

sigmundpetersen:
I think the default pipeline is configurable? I don't have access to check it out. Would be lovely if Zenhub could automate pipeline change from In Dev to Code Review when checks are succesful :) Do you know if this is anything near possible @mkllnk ?

daniellemoorhead:
Unfortunately @sigmundpetersen @mkllnk I don't know if this is possible and a brief google search didn't turn up anything helpful. I'll have a look next week and see what I can find.

PR #2562, opened 2018-08-20
[Product Import] Product attributes

### LINKED ISSUE #2458: [Product Import] Display error if attempt to modify product based fields
## Description
When a product is created, technically the first variant of the product is created, and some products have multiple variants.
Some fields are set up at the product level and apply to all variants. They can't be different from one variant to another.
With product import feature, there are issues if the user tries to change product based info on a variant line that is contradicting the product info in the system:
- if there are more variants and product info are changed, those change will apply to all variants without the user understanding it
- if there is only one variant then the change could apply to the product info as no conflict possible

The easier way to handle that is to say that product based fields (category, unit_type, product_item-name, tax_category, shipping_category) can be created for new products, but if the product already exist those fields can't be modified.

It is likely you don't want to change those info, they are set for once, so we could make it user friendly by explaining something like: If attempt to modify product based product, write error message “You cannot update fields X, X, X, X in a product from product import feature”

## Acceptance criteria
Among the various possible errors, if a user try to update in an existing product one of the following fields: category, unit_type, variant_unit-name, tax_category, shipping_category then the following message error is displayed for the product: "You cannot update [field] for an existing product from product import feature"

Matt-Yorkley:
#### What? Why?

Closes #2458

Adds error messages when updating product-based fields on variants of existing products.

#### What should we test?

When attempting to update the **category, description, unit_type, variant_unit_name, tax_category,** or **shipping_category** fields on an existing product, the user is shown an error message that the field cannot be updated.

#### Release notes

Added erorrs when user is attempting to update non-updatable fields during product import.

Changelog Category: Changed

### DIFF
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -178,7 +178,7 @@ module ProductImport
       return if category.blank?
 
       if index.key? category
-        entry.send("#{type}_category_id=", index[category])
+        entry.public_send("#{type}_category_id=", index[category])
       else
         mark_as_invalid(entry, attribute: "#{type}_category", error: I18n.t('admin.product_import.model.not_found'))
       end
@@ -239,11 +239,11 @@ module ProductImport
     end
 
     def attributes_match?(attribute, existing_product, entry)
-      existing_product.send(attribute) == entry.send(attribute)
+      existing_product.public_send(attribute) == entry.public_send(attribute)
     end
 
     def attributes_blank?(attribute, existing_product, entry)
-      existing_product.send(attribute).blank? && entry.send(attribute).blank?
+      existing_product.public_send(attribute).blank? && entry.public_send(attribute).blank?
     end
 
     def permission_by_name?(supplier_name)
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -14,6 +14,16 @@ module ProductImport
       @import_settings = import_settings
     end
 
+    def self.non_updatable_fields
+      {
+        category: :primary_taxon_id,
+        unit_type: :variant_unit_scale,
+        variant_unit_name: :variant_unit_name,
+        tax_category: :tax_category_id,
+        shipping_category: :shipping_category_id
+      }
+    end
+
     def validate_all(entries)
       entries.each do |entry|
         supplier_validation(entry)
@@ -182,6 +192,8 @@ module ProductImport
         return
       end
 
+      products.each { |product| product_field_errors(entry, product) }
+
       products.flat_map(&:variants).each do |existing_variant|
         if entry_matches_existing_variant?(entry, existing_variant) && existing_variant.deleted_at.nil?
           return mark_as_existing_variant(entry, existing_variant)
@@ -213,6 +225,25 @@ module ProductImport
       else
         mark_as_invalid(entry, product_validations: existing_variant.errors)
       end
+    end
+
+    def product_field_errors(entry, existing_product)
+      non_updatable_fields.each do |display_name, attribute|
+        next if attributes_match?(attribute, existing_product, entry) || attributes_blank?(attribute, existing_product, entry)
+        mark_as_invalid(entry, attribute: display_name, error: I18n.t('admin.product_import.model.not_updatable'))
+      end
+    end
+
+    def non_updatable_fields
+      EntryValidator.non_updatable_fields
+    end
+
+    def attributes_match?(attribute, existing_product, entry)
+      existing_product.send(attribute) == entry.send(attribute)
+    end
+
+    def attributes_blank?(attribute, existing_product, entry)
+      existing_product.send(attribute).blank? && entry.send(attribute).blank?
     end
 
     def permission_by_name?(supplier_name)
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -17,6 +17,7 @@ module ProductImport
     def self.non_updatable_fields
       {
         category: :primary_taxon_id,
+        description: :description,
         unit_type: :variant_unit_scale,
         variant_unit_name: :variant_unit_name,
         tax_category: :tax_category_id,
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -228,14 +228,10 @@ module ProductImport
     end
 
     def product_field_errors(entry, existing_product)
-      non_updatable_fields.each do |display_name, attribute|
+      EntryValidator.non_updatable_fields.each do |display_name, attribute|
         next if attributes_match?(attribute, existing_product, entry) || attributes_blank?(attribute, existing_product, entry)
         mark_as_invalid(entry, attribute: display_name, error: I18n.t('admin.product_import.model.not_updatable'))
       end
-    end
-
-    def non_updatable_fields
-      EntryValidator.non_updatable_fields
     end
 
     def attributes_match?(attribute, existing_product, entry)
--- a/app/models/product_import/product_importer.rb
+++ b/app/models/product_import/product_importer.rb
@@ -57,6 +57,13 @@ module ProductImport
       @sheet ? @sheet.last_row - 1 : 0
     end
 
+    def product_field_errors?
+      @entries.each do |entry|
+        return true if entry.errors.messages.values.include? [I18n.t('admin.product_import.model.not_updatable')]
+      end
+      false
+    end
+
     def reset_counts
       # Return indexed data about existing product count, reset count, and updates count per supplier
       @reset_counts.each do |supplier_id, values|
--- a/spec/models/product_importer_spec.rb
+++ b/spec/models/product_importer_spec.rb
@@ -22,17 +22,17 @@ describe ProductImport::ProductImporter do
   let!(:tax_category2) { create(:tax_category) }
   let!(:shipping_category) { create(:shipping_category) }
 
-  let!(:product) { create(:simple_product, supplier: enterprise2, name: 'Hypothetical Cake', primary_taxon_id: category2.id) }
+  let!(:product) { create(:simple_product, supplier: enterprise2, name: 'Hypothetical Cake', description: nil, primary_taxon_id: category2.id) }
   let!(:variant) { create(:variant, product_id: product.id, price: '8.50', on_hand: '100', unit_value: '500', display_name: 'Preexisting Banana') }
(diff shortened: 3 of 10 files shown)


myriamboure:
@Matt-Yorkley we said that we wouldn't save any import file until there are errors, as it is pretty confusing to enable saving if some products contain errors, we want to make sure the user is conscious about what she imports... and safe that she is not making mistakes or messing up her catalog. So shouldn't we stick to the same bevahior? 
= Display those as error lines and let them make the corrections, either by changing the category on the platform for the given product (if they realize they did a mistake previously) or change it in their file if they realize they did a mistake in their file.
@RachL do you agree?
I probably should have been much more details in expected UX in specifying the issue, sorry!
For the upper warning message, I would move it to the main product import page so users know that before starting to import a file. And only display as errors in this second step if they didn't follow the rule and prevent saving.
capture du 2018-08-21 15-37-53

And about the small "warning icon" on the item line, I would keep it but as the line would be in error section users will have to fix before being able to save. So it explains them why it is not saving and what they have to fix. Here also it's not clear if the value you see which is going to be imported is the one you want to import or the original one? If it's in error line it is clear that the value displaying the error message is the one you want to import and which is wrong.

Do you agree?

myriamboure:
I put it back to in dev so that we can discuss before it is reviewed ;-)

Matt-Yorkley:
@myriamboure I thought these would be ok as warnings, as they don't actually make any of the products invalid. 

The message at the top about product fields only shows when there are changed product fields in the file, so most of the time it will not be there (only when relevant). The information feels like something we should have in the docs as opposed to the first upload page.

myriamboure:
@RachL we need some UX feedback on this ;-) As it is important info I would leave it as a warning in the app as well... but happy to just take it out and leave it only in user guide if you all prefer that.

But on the pass/block saving question, @Matt-Yorkley I insist that if there is a risk of confusion in hub manager mind, it's not desired UX. And if I upload a file of 100 products, the system tells me the warning, but I have no clue how many and which products are concerned, so I have to go through the whole list and see where are the warnings to see if there is a problem or not, etc. It's not good UX to me. If you allow import and user doesn't know clearly what she imports, we need to review UX.

RachL:
@myriamboure if I understand correctly, the message is shown when the user add those column in the file, right ? So it basically tells him that those columns won't be taken into account. I guess that it could be convenient to not block the import for that, so that the user has not to do another step of changing his file.
Moreover, in which case will a user really add those columns as they are not on the list of available columns in template file?

myriamboure:
No it's not that @RachL. In a same import, a use can have new products for which those columns will be taken into account AND existing one for which they won't. And he won't know which one are taken into account or not.

myriamboure:
@Matt-Yorkley we just discussed with @RachL and we'll talk quickly about it in our catchup today, but when discussing we agreed with Rachel on UX side that it would be better to :
- Display error and block saving, so that we have a consistent behavior and avoid confusion in users mind about what they changed or didn't change. As any other error, with error message like "this product has conflicting info on main product attribute already in the catalog. Please check and resolve conflicts before submitting again." You can leave some warning icon on the field concerned by the conflict. If user want to import data with conflicting info it means she makes a mistake, either in the file or there is a mistake in the platform. So we want to help her fix her mistakes.
- Don’t add notice on main product import page. We agree that it has to be in user guide. If we find lots of users are having issues with this present case we can see how to explain better.
- For later, on iteration 2, it would be good to open some "conflict resolution modal (display two info and ask which one to choose)" so that we don't block saving and help resolve conflicts directly.

app/models/product_import/entry_validator.rb
```diff
@@ -220,6 +232,25 @@ def mark_as_existing_variant(entry, existing_variant)
...
+    def non_updatable_fields
+      EntryValidator.non_updatable_fields
+    end
+
+    def attributes_match?(attribute, existing_product, entry)
+      existing_product.send(attribute) == entry.send(attribute)
+    end
+
+    def attributes_blank?(attribute, existing_product, entry)
+      existing_product.send(attribute).blank? && entry.send(attribute).blank?
```
sauloperez: please, use `public_send` over `send`. The latter does not respect `private` methods.

app/models/product_import/product_importer.rb
```diff
@@ -57,6 +57,13 @@ def item_count
       @sheet ? @sheet.last_row - 1 : 0
     end
 
+    def product_field_errors?
+      @entries.each do |entry|
+        return true if entry.errors.messages.values.include? [I18n.t('admin.product_import.model.not_updatable')]
+      end
+      false
+    end
```
sauloperez: Using [any?](https://ruby-doc.org/core-2.2.0/Enumerable.html#method-i-any-3F) would be a bit more readable IMO.

spec/models/product_importer_spec.rb
```diff
@@ -317,6 +319,33 @@
...
 
+  describe "updating non-updatable fields on existing products" do
+    before do
+      csv_data = CSV.generate do |csv|
+        csv << ["name", "supplier", "category", "on_hand", "price", "units", "unit_type"]
+        csv << ["Beetroot", "And Another Enterprise", "Meat", "5", "3.50", "500", "g"]
+        csv << ["Tomato", "And Another Enterprise", "Vegetables", "6", "5.50", "500", "Kg"]
+      end
+      File.write('/tmp/test-m.csv', csv_data)
+      file = File.new('/tmp/test-m.csv')
```
sauloperez: maiking use of Rails' tmp/ folder seems more appropriate. If something happens and you spot this file in tmp/ you know it's something related to the app.

Matt-Yorkley: Product import uses the rails tmp folder for it's destination of uploaded files, but this is purely for the specs to create a CSV file first on the disk, and then "upload" or use it during the test. I thought the system /tmp folder would be better, as it will get cleared automatically regardless of what machine the specs are running on.

app/models/product_import/entry_validator.rb
```diff
@@ -220,6 +232,25 @@ def mark_as_existing_variant(entry, existing_variant)
       end
     end
 
+    def product_field_errors(entry, existing_product)
+      non_updatable_fields.each do |display_name, attribute|
+        next if attributes_match?(attribute, existing_product, entry) || attributes_blank?(attribute, existing_product, entry)
+        mark_as_invalid(entry, attribute: display_name, error: I18n.t('admin.product_import.model.not_updatable'))
+      end
+    end
```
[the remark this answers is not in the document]
Matt-Yorkley: There are some comparisons that need to be checked here that depend on truthyness as opposed to truth. Comparing `1 == 1.0`, for instance. If I use `#slice` here the comparison will be strict, and won't give the correct result.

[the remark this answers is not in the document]
Matt-Yorkley: I also need to compare across the two values to see if they are both `#blank`, and I need it to be true if one is `nil` and the other is an empty string, which gives me the same problem with using `#slice` here.

luisramos0:
I have staged this in https://staging.openfoodfrance.org/
For now, if any error, first place to look should be the staging server.

RachL:
@luisramos0 I get the snail once I hit the import button even with a very simple file like here:
https://docs.google.com/document/d/1y_A2w0-yDz28Z4-RH6LfUZJ05zr3o4eBkXJRrFBUTIA/edit#
So, is my test crappy or do we have a pb with the server? Thanks :)

RachL:
@luisramos0 FYI I've used the same file on https://staging1.openfood.com.au and there it works perfectly.

luisramos0:
I'm having a look right now.

luisramos0:
I just restarted the server (unicorn) and it's now working, at least the first step.

RachL:
thx @luisramos0 working good :)

I'm not sure I've tested this right. To sum up:
- yes users are warned and cannot proceed with import as they need to change file first (although the sentence used is not the one proposed by Myriam, I guess we can change this later ;) ).
- I'm confused of the term "attempting to update" : does it mean having the product in the file and that's it? Or if the product has to have the exact info than in DB we are good ? If it's the latter, there is a problem somewhere has the field "description" is not accepted for existing products, even if the content is the same as in DB.
If any import featuring existing products is considered an update, why are fields like category or shipping_category not mentioned in the errors ? Are there because they are mandatory fields?
Thanks for feedback and ping @myriamboure for testing review, maybe I've just misunderstood how it works here :)

Testing notes:
https://docs.google.com/document/d/1y_A2w0-yDz28Z4-RH6LfUZJ05zr3o4eBkXJRrFBUTIA/edit#

myriamboure:
Ok so we just checked again and talked with @RachL and it seems the feature is working as expected for all fields expect Description.
It seems @Matt-Yorkley that the description field is not able to recognize if it is the same or not as in DB, and always send an error if you have a description field in the file for an existing product, even if description is same. So it will prevent imports for any existing product if the column description is used in the file, even is no product-based field is touched.
So we need to fix that before merging or we will just be unable to use any file with a description column and existing products if we merge. Back to in dev @Matt-Yorkley sorry ;-)

myriamboure:
@Matt-Yorkley there has been no activity on this since 11 days. It would  be good to prioritize finishing product import stuff so that we can release, over starting bulk invoice printing stuff ;-) Thanks!

app/models/product_import/product_importer.rb
sauloperez: I meant using any? with `@entries`. If I understood correctly, we're checking if any of the entries has an error regardless of the number of errored ones, right?

Matt-Yorkley:

RachL:
Yes, now it works :) https://docs.google.com/document/d/1y_A2w0-yDz28Z4-RH6LfUZJ05zr3o4eBkXJRrFBUTIA/edit#

@Matt-Yorkley did you install whole production database on UK staging ? Using filter on product list is sometimes very long (or with cache problems maybe....) and there is no loading signal after filtering, so you don't know if the platform is looking for products or just did not find them. IMO We should have a loading signal after filtering as well, and a sentence when no products are found. What do you think ?

Matt-Yorkley:
@RachL yes, we have lots of data on UK staging. Maybe it's worth opening an issue to review the user feedback during loading on that page.

myriamboure:
I have the same experience in France, before we had "no product" now we have nothing but the UX is not better... I experience same issue in France as super admin, that would be worth testing with regular user to see if it's also soooo long to charge.

myriamboure:
Ok so let's merge and we'll open a new issue for loading time and UX, let's merggggge !

PR #2567, opened 2018-08-22
Switch feature matcher style from `to_not have` to `to have_no`

greysteil:
I noticed you're using `to_not have` expectations on some feature tests. I'm pretty sure that you'd be better switching these to Capybara's negative selector, as you'll avoid Capybara waiting the full length of time until it times out for each of these tests.

Check out the Capybara finders section of [this blog post](https://blog.codeship.com/faster-rails-tests/) for more details.

### DIFF
--- a/spec/features/admin/bulk_order_management_spec.rb
+++ b/spec/features/admin/bulk_order_management_spec.rb
@@ -132,7 +132,7 @@ feature %q{
       end
 
       it "adds the class 'ng-dirty' to input elements when value is altered" do
-        expect(page).to_not have_css "input[name='quantity'].ng-dirty"
+        expect(page).to have_no_css "input[name='quantity'].ng-dirty"
         fill_in "quantity", :with => 2
         expect(page).to have_css "input[name='quantity'].ng-dirty"
       end
--- a/spec/features/admin/bulk_product_update_spec.rb
+++ b/spec/features/admin/bulk_product_update_spec.rb
@@ -770,10 +770,10 @@ feature %q{
 
         # Shows spinner whilst loading
         expect(page).to have_css "img.spinner", visible: true
-        expect(page).to_not have_css "img.spinner", visible: true
+        expect(page).to have_no_css "img.spinner", visible: true
       end
 
-      expect(page).to_not have_selector "div.reveal-modal.product-image-upload"
+      expect(page).to have_no_selector "div.reveal-modal.product-image-upload"
 
       within "table#listing_products tr#p_#{product.id}" do
         # New thumbnail is shown in image column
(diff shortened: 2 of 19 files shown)


greysteil:
Ah, interesting - I hadn't seen that. Perhaps this doesn't make the improvements I thought, then!

Sadly no cop available at the moment.

mkllnk:
@kristinalim Can you explain a bit more how this change would cause your described bug not to be detected? It seems to me that if the page isn't rendered yet that either version would fail. So we would need to first test that the page is rendered and then check that only the desired content is there.

There is also this often quoted hint that says that the two versions are the same:
https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends:

> Capybara's RSpec matchers, however, are smart enough to handle either form. The two following statements are functionally equivalent:
> 
> ```ruby
> expect(page).not_to have_xpath('a')
> expect(page).to have_no_xpath('a')
> ```

kristinalim:
It appears I've been outdated on changes in Capybara. As far as I know (I could be wrong), `to_not have_` used to behave like `refute page.has_`, which would wait until Capybara max wait time for the content to appear, which it shouldn't. @greysteil I think this related to the optimization you planned to address in this PR.

@mkllnk Aside from the quote you posted, I did some quick tests and I'm convinced that `to_not have_` and `to have_no_` are indeed functionally the same now.

Marking this PR approved now too.

#### Possible race condition

The race condition that could miss a bug using the `to have_no_` form is also possible in the pre-PR `to_not have_` form, so it's beyond the scope of this PR. (`refute page.has_` could catch it.)

Here's a patch that demonstrates it, if you are interested: [explore_to_have_no.txt](https://github.com/openfoodfoundation/openfoodnetwork/files/2326649/explore_to_have_no.txt)
 This introduces a bug where filtering to an enterprise lists customers for all enterprises.

```shell
git apply explore_to_have_no.txt
bundle exec rspec --tag focus
```

The `to_not have_` tests still pass, but later in the examples, the inverses pass.

Result:

```shell
  1) Customers as an enterprise user using the customers index behaviour with 'to_not have' vs to have_no_' fails on first customer check when using 'refute page.has_selector?'
     Failure/Error: refute page.has_selector?("tr#c_#{customer1.id}"), "customer1 should not appear in the page."
     
     MiniTest::Assertion:
       customer1 should not appear in the page.
     # ./spec/features/admin/customers_spec.rb:31:in `block (5 levels) in <top (required)>'

 3/3 |========================================================================================================================= 100 ==========================================================================================================================>| Time: 00:01:40 

Finished in 1 minute 40.51 seconds (files took 12.72 seconds to load)
```

mkllnk:
Thank you for all the detail @kristinalim.

PR #2583, opened 2018-08-24
Configure mail TLS or SSL connection when seeding

mkllnk:
#### What? Why?

First part of https://github.com/openfoodfoundation/ofn-install/issues/239.

This allows to seed the secure connection type of the default mail
method. Most servers should use transport encryption when sending emails. For example, the new Australian staging server couldn't send email out of the box, because Gmail doesn't except insecure connections.

#### What should we test?

No user testing required.

#### Release notes

Simplified initial mail configuration.

Changelog Category: Added

### DIFF
--- a/config/application.yml.example
+++ b/config/application.yml.example
@@ -30,6 +30,7 @@ SMTP_USERNAME: 'ofn'
 SMTP_PASSWORD: 'f00d'
 
 # Optional mail settings
+# MAIL_SECURE_CONNECTION: 'TLS' # 'None' (default), 'SSL' or 'TLS'
 # MAILS_FROM: hello@example.com
 # MAIL_BCC: manager@example.com
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -42,7 +42,7 @@ def create_mail_method
     preferred_mail_auth_type: 'login',
     preferred_smtp_username: ENV.fetch('SMTP_USERNAME'),
     preferred_smtp_password: ENV.fetch('SMTP_PASSWORD'),
-    preferred_secure_connection_type: 'None',
+    preferred_secure_connection_type: ENV.fetch('MAIL_SECURE_CONNECTION', 'None'),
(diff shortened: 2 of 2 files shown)


mkllnk:
@Matt-Yorkley @sauloperez I noticed that we are setting environment variables in `application.yml` here. But ofn-install uses `/etc/defaults/openfoodnetwork` to set these environment variables. I found that confusing. Is there a reason why we are storing this config in yet another location?

config/application.yml.example
```diff
@@ -30,6 +30,7 @@ SMTP_USERNAME: 'ofn'
 SMTP_PASSWORD: 'f00d'
 
 # Optional mail settings
+# MAIL_SECURE_CONNECTION: 'TLS' # 'None' (default), 'SSL' or 'TLS'
```
luisramos0: I don't think there's ever a reason to use SSL. Should be TLS only.

Matt-Yorkley:
@mkllnk It's a legacy code thing we need to look at. ofn-install overwrites `application.yml` completely, I guess it's there now partly as a reference in the main repo for what should be there, and partly for anyone not using ofn-install for their setup (maybe some contributors in their dev envrionment). 

I think there's an issue open in ofn-install to properly review this.

config/application.yml.example
mkllnk: I'm not sure about the naming here. TLS could mean STARTTLS on port 587. That is the configuration we use at the moment. And SSL could mean TLS on port 465. Both are secure connections as long as the underlying TLS implementation is secure. https://www.fastmail.com/help/technical/ssltlsstarttls.html

And it looks like TLS over port 465 was nominated the winner this year: https://tools.ietf.org/html/rfc8314#section-3

Maybe we should derive the connection method from the port? It's very unlikely that somebody wants to use the wrong encryption on the wrong port.

luisramos0: I am not sure but after reading a few things including your links, I think you may be right and "SSL could mean TLS on port 465".

PR #2585, opened 2018-08-24
Bump active_model_serializer to 0.8.4, and pin it

greysteil:
Also bump `therubyracer` from `0.12.0` to `0.12.3`, because I was finding `0.12.0` impossible to install!

See https://github.com/rails-api/active_model_serializers#status-of-ams for details of what's happening with AMS. At some point (after a Rails update) it may make sense to update to the `0.10.x` branch, but it *doesn't* make sense to go to `0.9.x` first. When the time comes for that there's a guide [here](https://github.com/rails-api/active_model_serializers/blob/v0.10.6/docs/howto/upgrade_from_0_8_to_0_10.md).

Changelog for this release is at https://github.com/rails-api/active_model_serializers/blob/v0.8.4/CHANGELOG.md.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,10 @@ gem 'andand'
 gem 'truncate_html'
 gem 'representative_view'
 gem 'rabl'
-gem "active_model_serializers"
+
+# AMS is pinned to 0.8.4 because 0.9.x is a complete re-write, as is 0.10.x
+# Once Rails is updated to 5.x we should bump directly to 0.10.x
+gem "active_model_serializers", "0.8.4"
 gem 'oj'
 gem 'deface', github: 'spree/deface', ref: '1110a13'
 gem 'paperclip'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -150,7 +150,7 @@ GEM
       rack-cache (~> 1.2)
       rack-test (~> 0.6.1)
       sprockets (~> 2.2.1)
-    active_model_serializers (0.8.3)
+    active_model_serializers (0.8.4)
       activemodel (>= 3.0)
     activemerchant (1.78.0)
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -48,7 +48,7 @@ gem 'andand'
 gem 'truncate_html'
 gem 'representative_view'
 gem 'rabl'
-gem "active_model_serializers"
+gem "active_model_serializers", "0.8.4"
```
mkllnk: It would be nice to add a comment here why we chose this version and a hint when to upgrade (Rails version).

Gemfile.lock
```diff
@@ -660,8 +660,8 @@ GEM
     stringex (1.3.3)
     stripe (3.3.1)
       faraday (~> 0.9)
-    therubyracer (0.12.0)
-      libv8 (~> 3.16.14.0)
+    therubyracer (0.12.3)
```
mkllnk: One of our staging servers is really old and failed to install mini_racer. We should test deployment on Australian staging1 to make sure this passes (or finally replace that server).

greysteil:
@mkllnk - updated. Good to go once you've tested the therubyracer update on that staging server.

mkllnk:
Staged on https://staging1.openfood.com.au/. For testing:

Just do a sanity check. The updated dependency is used on almost every page. So if you can do a checkout, everything should be fine.

greysteil:
Had a click around and everything looks fine.

RachL:
Looking good!

Testing notes : https://docs.google.com/document/d/1Uqmlek2piBx_5IoPEvqNrXfqD2FLNddeFAy_maFLPMk/edit#

sauloperez:
Nice! thanks @greysteil !

PR #2574, opened 2018-08-22
Spree2 - Bulk line items controller spec fix

luisramos0:
#### What? Why?

Closes #2573 
This means 21 freshly fixed tests in the spree branch build.
See issue for details. 

#### What should we test?
The spec should be green.  

#### Release notes
No release notes.

#### How is this related to the Spree upgrade?
This is part of the spree upgrade, see issue.

Issue #2020, opened 2017-12-21
[Spree 2 Upgrade] Fix references to MailMethod in specs


sauloperez:
## What

We need to get rid of the creation of `Spree::MailMethod` in all specs and use `Spree::Config[:mails_from]` instead. 

So in https://github.com/openfoodfoundation/openfoodnetwork/blob/master/spec/controllers/spree/orders_controller_spec.rb#L401 we would turn:

```ruby
Spree::MailMethod.create!(
  environment: Rails.env,
  preferred_mails_from: 'spree@example.com'
)
```

into:

```ruby
Spree::Config[:mails_from] = "spree@example.com"
```

sauloperez:
This is not yet doable as it depends on https://github.com/spree/spree/pull/2643/files#diff-a428b7a87ba7179c65fcb97529aa2452R10 from Spree. Until we don't reach that point our efforts are pointless.

luisramos0:
I just looked at spree 2 branch and I believe we do have this already and this is now doable and it is not blocked.

luisramos0:
Specs referencing MailMethod in spree 2-0-4-stable branch:
spec/controllers/admin/subscriptions_controller_spec.rb
spec/controllers/spree/admin/orders_controller_spec.rb
spec/controllers/spree/orders_controller_spec.rb
spec/controllers/user_passwords_controller_spec.rb
spec/features/consumer/shopping/orders_spec.rb
spec/jobs/subscription_confirm_job_spec.rb
spec/mailers/enterprise_mailer_spec.rb
spec/mailers/producer_mailer_spec.rb
spec/mailers/user_mailer_spec.rb
spec/models/order_cycle_spec.rb
spec/models/spree/order_spec.rb
spec/services/create_mail_method_spec.rb

PR #2587, opened 2018-08-25
Spree 2 Upgrade - MailMethod - Fixed bug on order_mailer_decorator (only visible in spree 2)

luisramos0:
When cancel_email is called with order id in paramater order, order.number will fail. This is only seen in spree 2 because only spree 2 calls this method with order id. Bug introduced here: https://github.com/openfoodfoundation/openfoodnetwork/commit/32d2adc8a24f4352e8c6a0b179fe9ef351295f6d#diff-aa82d768109073ea8ad7858146630be4R9. Spree change introduced here: https://github.com/spree/spree/commit/8d90ae15f807467ffbcc20e779c9ffead577df2a#diff-1e2ba31309f1f1abd2a4b626036d316fR94

We dont need to put this in master because this is not a problem in spree 1 because cancel_email is always called with order (not order id) in spree 1.

#### What? Why?

Related to #2020 

Detected while fixing mailmethod related spec errors.
This error appears in spec/controllers/spree/orders_controller_spec.rb:407 (but only seen after PR #2574 fixes are applied as well).

#### What should we test?
After PR #2574 is applied, this PR gets spec/controllers/spree/orders_controller_spec.rb:404 to green.

This is part of getting the spree2 branch build to green.

#### Dependencies
This is related to PR #2574 but can go first.

### DIFF
--- a/app/mailers/spree/order_mailer_decorator.rb
+++ b/app/mailers/spree/order_mailer_decorator.rb
@@ -3,43 +3,42 @@ Spree::OrderMailer.class_eval do
   helper CheckoutHelper
   helper SpreeCurrencyHelper
 
-  def cancel_email(order, resend = false)
-    @order = find_order(order)
-    subject = (resend ? "[#{t(:resend).upcase}] " : '')
-    subject += "#{Spree::Config[:site_name]} #{t('order_mailer.cancel_email.subject')} ##{@order.number}"
-    mail(to: @order.email, from: from_address, subject: subject)
-  end
-
-  def confirm_email_for_customer(order, resend = false)
-    find_order(order) # Finds an order instance from an id
-    subject = (resend ? "[#{t(:resend).upcase}] " : '')
-    subject += "#{Spree::Config[:site_name]} #{t('order_mailer.confirm_email.subject')} ##{@order.number}"
+  def confirm_email_for_customer(order_or_order_id, resend = false)
+    @order = find_order(order_or_order_id)
+    subject = build_subject(t('order_mailer.confirm_email.subject'), resend)
     mail(:to => @order.email,
          :from => from_address,
          :subject => subject,
          :reply_to => @order.distributor.contact.email)
   end
 
-  def confirm_email_for_shop(order, resend = false)
-    find_order(order) # Finds an order instance from an id
-    subject = (resend ? "[#{t(:resend).upcase}] " : '')
-    subject += "#{Spree::Config[:site_name]} #{t('order_mailer.confirm_email.subject')} ##{@order.number}"
+  def confirm_email_for_shop(order_or_order_id, resend = false)
+    @order = find_order(order_or_order_id)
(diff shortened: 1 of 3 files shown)


app/mailers/spree/order_mailer_decorator.rb
```diff
@@ -6,8 +6,8 @@
   def cancel_email(order, resend = false)
     @order = find_order(order)
```
sauloperez: but by creating an `order` ivar we could be possibly shadowing any ivar already set in the `OrderMailer` definition (very likely given the name of the class). IMO creating a local var does the job and it is less risky.

We could rename the `order` argument to `order_or_order_id`. Isn't that what is happening?

luisramos0: very interesting little comment...
because I dont think your main point is valid, we have to set and keep @order because it's used in the mailer views.
BUT
you made me realise that after this
https://github.com/openfoodfoundation/openfoodnetwork/pull/2589/commits/0de691b8216f5ec183c7098021b6d37cecd2b6ad
we can just remove this cancel_email method :-) and use the spree order_mailer one.
Your comment still applies… but it’s now spree code. I think we will keep (in this case, the not so good) spree code as it is… this in wrt to cancel_email method.

I really liked your order_or_order_id variable name so I decided to use it in the remaining methods (I love making code smells more explicit like this function parameter that takes all types of variations without indication).

As I was refactoring... I also extracted email subject related copy pasted logic to a new build_subject method.

And as I was improving this space, I also removed MailMethod dependency from both this order_mailer and the subscription mailer specs.

I kept the initial commit as it makes historical sense and added 2 new commits with what I mention in this comment.

luisramos0:
I'll wait for @mkllnk or someone else to review the last version fo this PR.

app/mailers/spree/order_mailer_decorator.rb
```diff
@@ -3,43 +3,42 @@
...
-    mail(to: @order.email, from: from_address, subject: subject)
-  end
-
-  def confirm_email_for_customer(order, resend = false)
-    find_order(order) # Finds an order instance from an id
-    subject = (resend ? "[#{t(:resend).upcase}] " : '')
-    subject += "#{Spree::Config[:site_name]} #{t('order_mailer.confirm_email.subject')} ##{@order.number}"
+  def confirm_email_for_customer(order_or_order_id, resend = false)
+    @order = find_order(order_or_order_id)
+    subject = build_subject("#{t('order_mailer.confirm_email.subject')}", resend)
```
mkllnk: You don't need the string interpolation any more:
```ruby
subject = build_subject(t('order_mailer.confirm_email.subject'), resend)
```
There are a couple more cases in this file.

luisramos0: ah, true, well spotted! Fixed.

PR #2613, opened 2018-08-29
[Cookies] Updated cookies policy page text

### LINKED ISSUE #2601: [Cookies] Update text of cookie policy page (WIP)
## Description
Cookie policy page needs to be updated to make it readable on mobile (multi columns tables needs to be removed) + some light corrections on text.

## Expected Behavior
Content is https://docs.google.com/document/d/1RWvlXFzgpPtkhWhrm2OoWThHKwY87bu8e2l8-sPHa4w/edit?usp=sharing

## Actual Behavior
Content is slightly different and tables are not readable
img_9439

## Steps to Reproduce
From footer click on cookie policy page.

## Context
In cookie epic.

## Severity
S4

## Possible Fix
See update cookie policy page content, I have update the tables to put the info on just one column, see example:
capture du 2018-08-28 11-03-50

luisramos0:
Also migrated translations to ruby translations and extracted repeated partial from cookies policy page view.

#### What? Why?

Closes #2601 

See issue.

#### What should we test?
The policy page content should be retested including the backoffice configurations to display/hide the GA/matomo parts.
Must be tested on mobile as well.

#### Release notes
Changelog Category: Fixed

Cookies policy page with better UX and updated text.


luisramos0:
Maybe multiple commits on policy_page view would have been better...

luisramos0:
This is how the table looks like now. I think it's much better.

screen shot 2018-08-29 at 23 55 45

screen shot 2018-08-29 at 23 54 10

sigmundpetersen:
@myriamboure what do you think of the proposed solution?

myriamboure:
yes it's good it's exactly what I put in the template ;-) let's move to test ready!

luisramos0:
I staged this PR here: https://staging.openfoodfrance.org/

myriamboure:
Testing notes: https://docs.google.com/document/d/1QzbpB9sU1bN0Y0x7_4mh49xOZJ7_Sic-C6a5ZAnmcew/edit#

@luisramos0 the last sentence on the cookie policy page is not translate, but I checked in transifex and the translation is there, https://www.transifex.com/open-food-foundation/open-food-network/translate/#fr/enyml/150924320?q=text%3A'be+aware' any idea why? All the other strings are translated so it doesn't see normal... same in backoffice the back office configuration menu are well translated in Transifex and not on the staging.
Those are not strings affected by this PR, strings affected are not translated but that's normal.

Is it something we should be worried about? I did the translation quite some time ago, so that sounds strange that they are not applied on staging today, and especially as most of them are.

luisramos0:
Hey @myriamboure thanks for testing!

I had the same issue locally. I run the transifex script locally and I get those missing translations. So, the fix for the problem is to run the tx script to download translations. So, it should be fine.
But In transifex's history I see you translated everything in the same day (11 days ago). And the last TX update was 10 days ago.
I dont understand why only some entries are left behind... except if you left some of the entries unreviewed in tx?
Anyway, the solution is always to update translations (run transifex script).
So, I think this is ready to go.

PR #2618, opened 2018-08-30
Favor #public_send over #send using Rubocop's cop

sauloperez:
#### What? Why?

Devs keep using `#send` although that method does not preserve
private/protected visibility. Watching after this turned out to be quite
time-consuming while doing code review.

Currently, the Style/Send cop doesn't enforce `#public_send` however
(that's what we want). It simply discourages the use of #send. See
https://github.com/rubocop-hq/rubocop/pull/2081#issuecomment-292251650
for details. So a new entry on the Code Conventions doc has been added
to overcome this limitation:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code-Conventions#prefer-public_send-over-send

#### What should we test?
A smoke test, including:

* Product import
* Subscriptions
* Uploading a new logo from Configuration > Contents
* Inventory items (from `admin/variant_overrides` URL)
* Capture (or any other modification to the payment) an order's payment (from `admin/orders/:order_id/payments` URL)
* Enterprise fees

Although very little, these features are affected by this change.

#### Release notes
Ensure devs use `#public_send` over `#send` by enabling a Rubocop cop.

Changelog Category: Added

### DIFF
--- a/app/controllers/shop_controller.rb
+++ b/app/controllers/shop_controller.rb
@@ -44,7 +44,7 @@ class ShopController < BaseController
   private
 
   def filtered_json(products_json)
-    if applicator.send(:rules).any?
+    if applicator.rules.any?
       filter(products_json)
     else
       products_json
--- a/app/controllers/spree/admin/payment_methods_controller_decorator.rb
+++ b/app/controllers/spree/admin/payment_methods_controller_decorator.rb
@@ -10,7 +10,7 @@ module Spree
       # Only show payment methods that user has access to and sort by distributor name
       # ! Redundant code copied from Spree::Admin::ResourceController with modifications marked
       def collection
-        return parent.send(controller_name) if parent_data.present?
+        return parent.public_send(controller_name) if parent_data.present?
         collection = if model_class.respond_to?(:accessible_by) &&
                          !current_ability.has_block?(params[:action], model_class)
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,7 +15,7 @@ module ApplicationHelper
   # spree.foo_path in any view rendered from non-spree-namespaced controllers.
   def method_missing(method, *args, &block)
     if (method.to_s.end_with?('_path') || method.to_s.end_with?('_url')) && spree.respond_to?(method)
-      spree.send(method, *args)
+      spree.public_send(method, *args)
     else
       super
     end
--- a/app/models/model_set.rb
+++ b/app/models/model_set.rb
@@ -12,7 +12,7 @@ class ModelSet
     @collection = attributes[:collection] if attributes[:collection]
 
     attributes.each do |name, value|
-      send("#{name}=", value)
+      public_send("#{name}=", value)
     end
   end
 
--- a/app/models/proxy_order.rb
+++ b/app/models/proxy_order.rb
@@ -19,7 +19,7 @@ class ProxyOrder < ActiveRecord::Base
   def state
     # NOTE: the order is important here
     %w(canceled paused pending cart).each do |state|
-      return state if send("#{state}?")
+      return state if __send__("#{state}?")
     end
     order.state
   end
@@ -32,7 +32,7 @@ class ProxyOrder < ActiveRecord::Base
     return false unless order_cycle.orders_close_at.andand > Time.zone.now
     transaction do
       update_column(:canceled_at, Time.zone.now)
-      order.send('cancel') if order
+      order.cancel if order
       true
     end
   end
@@ -41,7 +41,7 @@ class ProxyOrder < ActiveRecord::Base
     return false unless order_cycle.orders_close_at.andand > Time.zone.now
     transaction do
       update_column(:canceled_at, nil)
-      order.send('resume') if order
+      order.resume if order
       true
     end
   end
--- a/app/services/order_syncer.rb
+++ b/app/services/order_syncer.rb
@@ -83,7 +83,7 @@ class OrderSyncer
 
   def addresses_match?(order_address, subscription_address)
     relevant_address_attrs.all? do |attr|
-      order_address[attr] == subscription_address.send("#{attr}_was") ||
+      order_address[attr] == subscription_address.public_send("#{attr}_was") ||
         order_address[attr] == subscription_address[attr]
     end
   end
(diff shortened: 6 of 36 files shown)


luisramos0:
There are 50 warnings, this means that when a dev touches one of these files in a PR, they will have to fix it to the new convention. Correct?

sauloperez:
AFAIK that Cop doesn't have autocorrect but I can try to fix them

app/controllers/shop_controller.rb
```diff
@@ -44,7 +44,7 @@ def changeable_orders_alert
   private
 
   def filtered_json(products_json)
-    if applicator.send(:rules).any?
+    if applicator.rules.any?
```
sauloperez: I made this one public. If we use it from outside, it's part of applicator's public API

app/controllers/spree/admin/payment_methods_controller_decorator.rb
```diff
@@ -10,7 +10,7 @@ module Admin
       # Only show payment methods that user has access to and sort by distributor name
       # ! Redundant code copied from Spree::Admin::ResourceController with modifications marked
       def collection
-        return parent.send(controller_name) if parent_data.present?
+        return parent.controller_name if parent_data.present?
```
sauloperez: [controller_name](https://apidock.com/rails/v3.2.13/ActionController/Metal/controller_name/class) is a public method

app/helpers/application_helper.rb
```diff
@@ -15,7 +15,7 @@ def ng_form_for(name, *args, &block)
   # spree.foo_path in any view rendered from non-spree-namespaced controllers.
   def method_missing(method, *args, &block)
     if (method.to_s.end_with?('_path') || method.to_s.end_with?('_url')) && spree.respond_to?(method)
-      spree.send(method, *args)
+      spree.public_send(method, *args)
```
sauloperez: Rails' URL helpers are public

app/models/model_set.rb
```diff
@@ -12,7 +12,7 @@ def initialize(klass, collection, attributes={}, reject_if=nil, delete_if=nil)
     @collection = attributes[:collection] if attributes[:collection]
 
     attributes.each do |name, value|
-      send("#{name}=", value)
+      public_send("#{name}=", value)
```
sauloperez: the setters of a model are meant to be public

app/models/proxy_order.rb
```diff
@@ -41,7 +41,7 @@ def resume
     return false unless order_cycle.orders_close_at.andand > Time.zone.now
     transaction do
       update_column(:canceled_at, nil)
-      order.send('resume') if order
+      order.resume if order
```
sauloperez: these two methods are public

app/services/order_syncer.rb
```diff
@@ -83,7 +83,7 @@ def relevant_address_attrs
 
   def addresses_match?(order_address, subscription_address)
     relevant_address_attrs.all? do |attr|
-      order_address[attr] == subscription_address.send("#{attr}_was") ||
+      order_address[attr] == subscription_address.public_send("#{attr}_was") ||
```
sauloperez: AR's dirty attributes API is public

lib/discourse/single_sign_on.rb
```diff
@@ -42,7 +42,7 @@ def self.parse(payload, sso_secret = nil)
         if BOOLS.include? k
           val = ["true", "false"].include?(val) ? val == "true" : nil
         end
-        sso.send("#{k}=", val)
+        sso.public_send("#{k}=", val)
```
sauloperez: each of these K are defined as public attr_accessor in line 11

lib/open_food_network/paperclippable.rb
```diff
@@ -4,18 +4,18 @@
...
   module Paperclippable
     def self.included(base)
-      base.send :extend, ActiveModel::Naming
-      base.send :extend, ActiveModel::Callbacks
-      base.send :include, ActiveModel::Validations
-      base.send :include, Paperclip::Glue
+      base.extend(ActiveModel::Naming)
+      base.extend(ActiveModel::Callbacks)
+      base.include(ActiveModel::Validations)
+      base.include(Paperclip::Glue)
```
sauloperez: `extend` and `include` are public Module methods in Ruby 2.1.5 https://ruby-doc.org/core-2.1.5/Module.html#method-i-include

lib/open_food_network/paperclippable.rb
```diff
@@ -4,18 +4,18 @@
...
 
       # Paperclip required callbacks
-      base.send :define_model_callbacks, :save, only: [:after]
-      base.send :define_model_callbacks, :commit, only: [:after]
-      base.send :define_model_callbacks, :destroy, only: [:before, :after]
+      base.define_model_callbacks(:save, only: [:after])
```
sauloperez: it is public https://apidock.com/rails/ActiveModel/Callbacks/define_model_callbacks

lib/spree/api/testing_support/setup.rb
```diff
@@ -20,7 +20,9 @@ def sign_in_as_enterprise_user!(enterprises)
           let!(:current_api_user) do
             user = create(:user)
             user.spree_roles = []
-            enterprises.each { |e| user.enterprise_roles.create(enterprise: send(e)) }
+            enterprises.each do |enterprise|
+              user.enterprise_roles.create(enterprise: enterprise)
```
sauloperez: no point for a send when `enterprise` is a var

sauloperez:
Thanks for the heads-up. I think this now deserves a smoke test.

Matt-Yorkley:
Looks like the tests are smoking already... :trollface:

app/controllers/spree/admin/payment_methods_controller_decorator.rb
mkllnk: The original did not call `send(:controller_name)`. Doesn't it call a method with the same name as the controller?

sauloperez: :hankey: yes, I went too fast on this one. Thanks for the heads-up!

RachL:
FINALLY!
Sorry it took me so long ...
Everything looks good:
 https://docs.google.com/document/d/1FLZVVyGaH6ubjeM-bpXHzeN4GxUTsTVC6fADeCO1xGk/edit

PR #2633, opened 2018-09-01
WIP - Removed unnecessary css_splitter. See #1361

### LINKED ISSUE #1361: Duplicate CSS definitions from use of css_splitter
Original from Paul:

> I've been looking at the CSS structure of OFN a bit. `css_splitter` is used but it seems like many CSS blocks are being imported twice because:
> 
> * `all.css` uses `@import '*';`
> * `all_split2.css` uses `*= require 'darkswarm/all'` 
> 
> Wont both of these import all the darkswarm CSS files?

## Description

Our `all.css` file includes `all_split2.css` which is entirely contained in `all.css` already. So the include is unnecessary and lets visitors wait longer before they can see the page.

## Expected Behavior

No duplicate CSS code is loaded.

## Actual Behavior

180 kb of duplicate CSS code is loaded via `all_split2.css`.

## Steps to Reproduce

1. Load any page and observe the css include links: `curl -s https://openfoodnetwork.org.au/ | grep 'all.*css'`
2. Compare the two files, for example: `sed -i 's/;/\n/g' *.css; meld *.css`

## Severity

s3 as it affects all users. It doesn't prevent most people from using the platform, but can put customers off that have a slow internet connection (bad mobile signal).

## Your Environment

Observed on the Australia production server:

- all-c448b4e380da3e7c7b14b01981eab7bd.css: 377K
- all_split2.css: 180K

all.css seems to contain third party styles like Foundation, all_split2.css does not.

## Possible Fix

Remove the css splitter and references to all_split2.css.

luisramos0:
Work in Progress.

#### What? Why?

Closes #1361

Removing css_splitter, if it is not necessary, is one solution to #1361.

#### What should we test?
Frontoffice css should remaining the same and the website should look the same.

#### Release notes
Changelog Category: Removed
Removed css splitter.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -103,7 +103,6 @@ gem 'foundation_rails_helper', github: 'willrjmarshall/foundation_rails_helper',
 
 gem 'jquery-rails'
 gem 'jquery-migrate-rails'
-gem 'css_splitter'
 
 gem 'ofn-qz', github: 'openfoodfoundation/ofn-qz', ref: '60da2ae4c44cbb4c8d602f59fb5fff8d0f21db3c'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -243,9 +243,7 @@ GEM
       safe_yaml (~> 1.0.0)
     css_parser (1.3.5)
       addressable
-    css_splitter (0.4.5)
-      sprockets (>= 2.0.0)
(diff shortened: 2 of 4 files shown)


luisramos0:
superseeded by #2637

Issue #238, opened 2014-05-14
Build tool to extract data from Stroudco database


pmackay:
This would format the data into the generic format defined in #237

aidandunsdon:
Project and Source Code for Migration Tool is at  

https://github.com/Stroudco/stroudco-data-migration

PR #2636, opened 2018-09-02
Removed old files that were used with, now unsupported, heroku

luisramos0:
#### What? Why?

Raised from the review of #238.

These files are no longer required because heroku is no longer supporte.

#### What should we test?
I don't think this needs testing.

#### Release notes
Changelog Category: Removed
Removed old files used in, now unsupported, heroku.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,0 @@
-web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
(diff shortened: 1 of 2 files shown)


luisramos0:
ahah, I always feel weird about putting release notes for these tiny PRs... but well, I just added something :-)

PR #2637, opened 2018-09-02
Remove duplicate css_splitter references, delete split css file.

### LINKED ISSUE #1361: Duplicate CSS definitions from use of css_splitter

Vadlusk:
#### What? Why?
Deletes all references to css_splitter and an unnecessary split css file to speed up the UX. 
Closes #1361 
Unblocks #2565
Related to #2633 

#### What should we test?
Visit the site and make sure all style is still as desired. 

#### Release notes
Changelog Category: Removed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -103,7 +103,6 @@ gem 'foundation_rails_helper', github: 'willrjmarshall/foundation_rails_helper',
 
 gem 'jquery-rails'
 gem 'jquery-migrate-rails'
-gem 'css_splitter'
 
 gem 'ofn-qz', github: 'openfoodfoundation/ofn-qz', ref: '60da2ae4c44cbb4c8d602f59fb5fff8d0f21db3c'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -243,9 +243,7 @@ GEM
       safe_yaml (~> 1.0.0)
     css_parser (1.3.5)
       addressable
-    css_splitter (0.4.5)
-      sprockets (>= 2.0.0)
-    daemons (1.2.6)
+    daemons (1.2.2)
     dalli (2.7.2)
     database_cleaner (0.7.1)
     db2fog (0.8.0)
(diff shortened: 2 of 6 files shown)


luisramos0:
Thanks @Vadlusk I moved it to Test Ready, we need to wait for availability of a staging server. Someone will pick this up soon and stage it (maybe me). And then someone tests it in staging. And  then someone merges it into master.

Vadlusk:
Thank you @luisramos0 ! It was fun contributing. You and your team are all very kind.

luisramos0:
hey, what do you mean? WAS fun? it WILL be fun ;-)

luisramos0:
I just double checked that the split2 file is gone in staging katuma:
screen shot 2018-09-11 at 11 12 15

RachL:
Alright so I went through both front and back office. Front office twice actually (mobile and pc). Nothing changed so moving this to ready to go

PR #2638, opened 2018-09-03
Unit test variant stock

### LINKED ISSUE #2627: [Spree Upgrade] Add unit tests, doc and deprecation warnings to VariantStock
## Description

#2615 got merged but we're still missing unit tests for this new `VariantStock` module.

Also, we should document what happens with the various methods involved in that module as mentioned in https://github.com/openfoodfoundation/openfoodnetwork/pull/2615#discussion_r214063245, as well as add deprecation warnings to prevent us from relying on this old API (see https://github.com/openfoodfoundation/openfoodnetwork/pull/2615#discussion_r214064783).

## Context

We decided to merge #2615 because @mkllnk wasn't able to complete it. With a new PR we avoid any possible conflicts with his work.

sauloperez:
#### What? Why?

Closes #2627 

As explained in the issue, this adds unit tests, documentation and deprecation warnings to the new `VariantStock` module. The one that provides backward compatibility for the current stock methods such as `#on_hand` or `#on_demand`.

You can read the commits to understand minor details of the decisions taken, like https://github.com/openfoodfoundation/openfoodnetwork/pull/2638/commits/5c14a8c66ee20554549a7a8fdcbbcbd83733d758.

#### What should we test?

The upgrade is not ready to be tested yet.

#### Release notes

Changelog Category: Changed


daniellemoorhead:
@sauloperez 

Please, please, #assignyourself to PRs. Please 🙏

lib/open_food_network/variant_stock.rb
```diff
@@ -1,80 +1,136 @@
...
+    # Sets whether the variant can be ordered on demand or not. Note
+    # that although this modifies the stock item, it is not
+    # persisted.
+    #
+    # @raise [StandardError] when the variant has no stock item yet
     def on_demand=(new_value)
+      warn_deprecation(__method__)
+
       raise_error_if_no_stock_item_available
+      raise_error_if_multiple_stock_items
```
sauloperez: I'll try to convince you this deserves to become a DB-level constraint although opinions (See https://openfoodnetwork.slack.com/threads/convo/C4NDJT3FY-1535723508.000100/) were against it.

First, at this point, it is too late. The data is already inconsistent violating data integrity and we can cleaning it manually is the most we can do. And that is hard.

Second, nothing stops us data to become inconsistent by any other means. Adding this sort of barrier, doesn't prevent this from happening from a rails console, a controller or even a psql session. It's like *poner puertas al campo* aka. trying to stop the unstoppable.

As a result, unconfident programming comes up; the lines of code consuming `stock_items` become paranoid checking there's a single stock item throughtout the app.

That is why, I believe this would be better off replaced by `add_index :spree_stock_items, :variant_id, unique: true` in a migration. I know we don't like messing with Spree's schema but this doesn't modify the tables and the data integrity it provides well outweighs it.

lib/open_food_network/variant_stock.rb
```diff
@@ -1,19 +1,31 @@
 require 'active_support/concern'
 
-# These methods were available in Spree 1, but were removed in Spree 2.
-# We would still like to use them. Therefore we use only a single stock location
-# (default stock location) and use it to track the `count_on_hand` value that
-# was previously a database column on variants.
+# These methods were available in Spree 1, but were removed in Spree 2.  We
+# would still like to use them so that we still give support to the consumers
+# of this methods, making the upgrade backward compatible.
+#
```
luisramos0: I believe this module should be called Adapter, VariantStockAdapter.
And the documentation could become more simple:

-# This module is an adapter for OFN to work with Spree v2 code. 
-# 
-# This adapter simplifies the Spree 2 data model by using only a single stock item per variant.
-# Each stock item is associated to a single stock location per instance (default stock location).
-# The stock items are used to track the `count_on_hand` value (previously a database column on -variants).
-# See https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade%3A-Stock-locations
-# for details.

lib/open_food_network/variant_stock.rb
```diff
@@ -1,19 +1,31 @@
...
+# Therefore we use only a single stock item per variant, which is associated to
+# a single stock location per instance (default stock location) and use it to
+# track the `count_on_hand` value that was previously a database column on
+# variants. See
+# https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade%3A-Stock-locations
+# for details.
 #
 # We may decide to deprecate these methods after we designed the Network feature.
```
luisramos0: We may have the deprecation warning here, for example, "Some methods are or may become deprecated." but I dont think we should mention roadmap/future items in code comments ever. I'd remove the reference to Network feature.

lib/open_food_network/variant_stock.rb
```diff
@@ -1,19 +1,31 @@
...
 
     included do
-      after_save :save_stock
+      after_update :save_stock
     end
 
+    # Returns the number of items of the variant available in the
+    # stock. When allowing on demand, it returns infinite.
+    #
+    # Note this method relies on Spree's 2.0 #total_on_hand.
```
luisramos0: I would remove this comment. The code is self explanatory.

lib/open_food_network/variant_stock.rb
```diff
@@ -22,39 +34,65 @@ def on_hand
       end
     end
 
+    # Alias for Spree's 2.0 #total_on_hand
```
luisramos0: Instead of referring to the method used below. It would be more useful to explain what this method does, probably just "Number of items available in the stock for this variant."

lib/open_food_network/variant_stock.rb
```diff
@@ -22,39 +34,65 @@ def on_hand
...
+    # @raise [StandardError] when the track_inventory_levels config
+    # key is not set.
     def on_hand=(new_level)
       error = 'Cannot set on_hand value when Spree::Config[:track_inventory_levels] is false'
       raise error unless Spree::Config.track_inventory_levels
 
       self.count_on_hand = new_level
     end
 
+    # Sets the stock level. As opposed to #on_hand= it does not check `track_inventory_levels`'s value. It does ensure there's a stock item for the variant however raising otherwise. See #raise_error_if_no_stock_item_available for details.
```
luisramos0: More then explaining what's already in the code, I'd prefer to see here the why it is doing so. Why is it not checking "track_inventory_levels" here? same for on_hand=, why is it not checking for the existence of stock_item?

lib/open_food_network/variant_stock.rb
```diff
@@ -22,39 +34,65 @@ def on_hand
...
+    # @raise [StandardError] when the variant has no stock item yet
     def count_on_hand=(new_level)
       raise_error_if_no_stock_item_available
       overwrite_stock_levels new_level
     end
 
     def on_demand
-      stock_items.any?(&:backorderable?)
+      warn_deprecation(__method__)
+      stock_items.first.backorderable?
```
luisramos0: nice!!

lib/open_food_network/variant_stock.rb
```diff
@@ -22,39 +34,65 @@ def on_hand
...
     end
 
+    # Sets whether the variant can be ordered on demand or not. Note
+    # that although this modifies the stock item, it is not
+    # persisted.
```
luisramos0: what do you mean by "it is not persisted"? why would you use it then?

lib/open_food_network/variant_stock.rb
luisramos0: I see. I dont like the approach because we should know who's doing what and fix when it's not happening. 

"become paranoid checking there's a single stock item throughout the app"
The place from where code is "consuming stock_items" should be one and only one: it's the model and service layer of this part of the app. But I know we don't have that kind of isolation... So, lets do it!

sauloperez:
Thanks for your review @luisramos0 ! I tried to share more knowledge and be concise. Let me know if it's all clear as a dev reading the code.

luisramos0:
Awesome! Comments are better now, thanks!
I know it's probably the normal thing to do in Rails but I have to say again, I dont like the fact the name of the module is not stating clearly that it is an Adapter or a Concern. This VariantStock will never be used outside of Variant, it's an add on, it is not a standalone service/module/model, so I'd call it VariantStockConcern.

sauloperez:
Yep, I still need to take a look at that

lib/open_food_network/variant_stock.rb
```diff
@@ -77,7 +84,7 @@ def count_on_hand=(new_level)
     # track_inventory_levels only. It was initially introduced in
     # https://github.com/openfoodfoundation/spree/commit/20b5ad9835dca7f41a40ad16c7b45f987eea6dcc
     def on_demand
-      warn_deprecation(__method__)
+      warn_deprecation(__method__, 'Spree::Config[:track_inventory_levels]')
```
mkllnk: This is not right. Track inventory levels is a global setting that we don't use at all. Alternatives are `in_stock?` and `stock_items.first.backorderable?`.

lib/open_food_network/variant_stock.rb
```diff
@@ -125,5 +134,12 @@ def overwrite_stock_levels(new_level)
       # have only one stock location.
       stock_items.first.__send__(:count_on_hand=, new_level)
     end
+
+    def warn_deprecation(method_name, new_method_name)
```
mkllnk: I'm feeling a bit uneasy to introduce deprecation warnings without having a clear plan what else to use. We can refer to the stock items, but that may change again when we detailed the network feature. I would only deprecate the methods we are sure how to replace them.

lib/open_food_network/variant_stock.rb
mkllnk: Yes, I'm in favour of a constraint. That does make some of my earlier comments redundant.

lib/open_food_network/variant_stock.rb
```diff
@@ -22,39 +34,81 @@ def on_hand
...
 
+    # Persists the single stock item associated to this variant. As defined in
+    # the top-most comment, as there's a single stock location in the whole
+    # database, there can only be a stock item per variant. See StockItem's
+    # definition at
+    # https://github.com/openfoodfoundation/spree/blob/43950c3689a77a7f493cc6d805a0edccfe75ebc2/core/app/models/spree/stock_item.rb#L3-L4
+    # for details.
     def save_stock
-      stock_items.each(&:save)
+      stock_items.first.save
```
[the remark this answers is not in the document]
sauloperez: Any idea how? I'll take a look anyway.

sauloperez:
Ok, so I renamed and moved it @luisramos0 . Aligned with what you suggested I moved it to `concerns/` to follow Rails 4 convention (In Rails 4 there's a `models/concern/` by default).

It's an old misconception that all stuff in `models/` need to be AR's models and all other abstractions we come up with in `lib/`. Any of these abstractions are also part of our domain logic and as such, it makes sense they also live under `app/`. This [CodeClimate's blog post](https://codeclimate.com/blog/what-code-goes-in-the-lib-directory/) is a nice and short explanation of the topic.

luisramos0:
Awesome! I will do the same on OrderShippingMethod.
For me it makes perfect sense, /lib for generic not-ofn specific stuff and /app for ofn stuff.

What I am concerned about is the difficulty of creating a service culture: business logic like this should not be in models or concerns (models are for ORM and data), it should be in services. I believe the best place for this would be /app/services, but it would have to be a service, not a concern.

lib/open_food_network/variant_stock.rb
sauloperez: > Track inventory levels is a global setting that we don't use at all

is that explained/documented somewhere?

sauloperez:
> Track inventory levels is a global setting that we don't use at all

is that explained/documented somewhere @mkllnk ?

Also, `#in_stock?`'s definition uses exactly that config.

https://github.com/openfoodfoundation/openfoodnetwork/blob/c4d4b52668393e733b02582644c2acf1a7e0f0a1/lib/open_food_network/scope_variant_to_hub.rb#L25-L29

lib/open_food_network/variant_stock.rb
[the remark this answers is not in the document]
sauloperez: that config is gone @Matt-Yorkley

luisramos0:
2 approvals. moving to test ready.
I think @Matt-Yorkley  just offered some advise at one point, he was not planning to review the PR.
correct me if I am wrong.

luisramos0:
moving to ready to go actually, no tests here.

sauloperez:
I followed his advice but nothing relevant came out of it, at least for now.

PR #2614, opened 2018-08-29
Transifex

Transifex-Openfoodnetwork:
Automatically created by our CI server.

PR #2643, opened 2018-09-04
Fetched latest translations from transifex with tx pull --force

luisramos0:
#### What? Why?

Manual fetch from transifex, so that all menu translations are included in next release.

This is simply the results of 
> `tx pull --force`

If this is correct, it will supersede automatic process on #2614 

#### What should we test?
I am not sure we need to test this.

#### Release notes
Changelog Category: Added
Added latest updated translations for all languages.


sauloperez:
I translated few keys earlier today, a menu link among those, so could you please, pull again?

luisramos0:
yep, done.

luisramos0:
all right! Thanks guys.
Can be merged. I can prepare the release afterwards.

PR #2648, opened 2018-09-06
Fix ask anonymous users to login for order page

### LINKED ISSUE #2506: Prompt person to login if they use a link to an order and they are not logged in already.
## Description
If someone tries to access their order via a link of this format while logged out... https://staging1.openfood.com.au/orders/R244060841 ... they currently see an 'unauthorised' message. 
image
Instead of this, could we prompt them to login?

This is important for subscriptions, because a big part of susbcriptions is that customers are able to edit their subs. If they're logged out when they try to edit their sub order, they'll get this confusing and abrupt 'unauthorised' message.

When does this scenario occur?
The first subscription conf email that goes to subscription customers.
image
This only occurs when a shop allows customers to make changes to their order while the OC is open.

## Expected Behavior
The person is prompted to login. After logging in they're either a) taken to the order page if they're the right cusotmer or b) they see the unauthorised message again if it's not the right user for that order.

Alternatively I'd expect to see better messaging, like: 'unauthorised, please login to view this page'. 

## Actual Behavior
The person clicks the link. 
They see this:
image
If they do have the initiative to login, they'll then see the order page. But there's no clear messaging that they need to login.

## Steps to Reproduce
1. Setup a subscription in a shop which allows changes to be made
2. Open a subscription order and look at the email confirmation
3. Try to open that link in a browser window where you're not logged in.

## Context
It's a road block for customers who want to edit a sub order

## Bug Severity
s3

## Your Environment
PC Chrome, Staging1.

kristinalim:
#### What? Why?

Closes #2506

Subscription emails link to the order page if the order is associated with a user. But if the user is not yet logged in the OFN website and clicks on the link, they get an HTTP 401 (Unauthorized).

This redirects the user to the login page instead. After logging in, the user is redirected back to the order page for which they are now authorized to view.

This also adds automated tests for other scenarios for the order page.

#### What should we test?

As guest:

1. Place an order.
2. After that, you should be redirected to the order details page. It should load.
3. Load the URL in another window (same browser session). It should load.

As signed up user:

1. Place an order.
2. After that, you should be redirected to the order details page. It should load.
3. Load the URL in another window (same browser session). It should load.
4. Copy the URL.
5. Open the URL in a new browser session (through incognito window). This should redirect you to the login page.
6. Sign in as the user. This should redirect you to the order page, which should load.

#### Release notes

- Fix order page for anonymous users - The user is now redirected to the login page, then redirected back to the order page after login.

Changelog Category: Fixed

Authorization for the order page could have changed, so this might no longer be necessary or there could be a more straightforward solution.

### DIFF
--- a/spec/features/consumer/shopping/orders_spec.rb
+++ b/spec/features/consumer/shopping/orders_spec.rb
@@ -3,6 +3,79 @@ require 'spec_helper'
 feature "Order Management", js: true do
   include AuthenticationWorkflow
 
+  describe "viewing a completed order" do
+    let!(:distributor) { create(:distributor_enterprise) }
+    let!(:customer) { create(:customer, user: user, enterprise: distributor) }
+    let!(:order_cycle) { create(:simple_order_cycle, distributors: [distributor]) }
+
+    let!(:bill_address) { create(:address) }
+    let!(:ship_address) { create(:address) }
+    let!(:shipping_method) { create(:free_shipping_method, distributors: [distributor]) }
+
+    let!(:order) do
+      create(:order_with_credit_payment,
+        customer: customer,
+        user: user,
+        distributor: distributor,
+        order_cycle: order_cycle
+      )
+    end
+
+    before do
+      # For some reason, both bill_address and ship_address are not set
+      # automatically.
+      #
+      # Also, assigning the shipping_method to a ShippingMethod instance results
+      # in a SystemStackError.
+      order.update_attributes!(
+        bill_address: bill_address,
+        ship_address: ship_address,
+        shipping_method_id: shipping_method.id
+      )
+    end
+
+    context "when checking out as an anonymous guest" do
+      let(:user) { nil }
+
+      it "allows the user to see the details" do
+        # Cannot load the page without token
+        visit spree.order_path(order)
+        expect(page).to_not be_confirmed_order_page
+
+        # Can load the page with token
+        visit spree.order_path(order, token: order.token)
+        expect(page).to be_confirmed_order_page
+
+        # Can load the page even without the token, after loading the page with
+        # token.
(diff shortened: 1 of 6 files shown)


spec/features/consumer/shopping/orders_spec.rb
```diff
@@ -3,6 +3,79 @@
...
+      # For some reason, both bill_address and ship_address are not set
+      # automatically.
+      #
+      # Also, assigning the shipping_method to a ShippingMethod instance results
+      # in a SystemStackError.
+      order.update_attributes!(
+        bill_address: bill_address,
+        ship_address: ship_address,
+        shipping_method_id: shipping_method.id
+      )
```
luisramos0: why is this not done above like:

let!(:order) do
      create(:order_with_credit_payment,
        customer: customer,
        user: user,
        distributor: distributor,
        order_cycle: order_cycle
bill_address: bill_address,
        ship_address: ship_address,
        shipping_method_id: shipping_method.id
      )
    end

kristinalim: @luisramos0 The current state of `Spree::Order` and how the factories are set up it turns out do not allow that. Please see thread here: https://openfoodnetwork.slack.com/messages/C2GQ45KNU/convo/C2GQ45KNU-1536159429.000100/ That leads to `SystemStackError` `stack level too deep`.

luisramos0: ok, thanks. just asking, and learning :-)

mkllnk:

RachL:
All good :)

testing notes:
https://docs.google.com/document/d/1S8kjrVsu9Z4inXtu665O0pgt4yIXdTXiwiAVhp8-azc/edit#

PR #2649, opened 2018-09-06
Fix i18n translation for Store link in admin header for `fr`

### LINKED ISSUE #2632: Wrong translation on backoffice header
## Description
On English menu the top right link in back office is "store" and sends to the store (front office)
untitled presentation 1

On French instance this is automatically translated by Spree, there is no string in Transifex, and it's wrongly translated by "Enregistrer" (=save) which is pretty confusing. And users don't know they can go to the store when they click on it.

untitled presentation

## Steps to Reproduce
Go to any non English back office page.

## Severity
S4

## Possible Fix
Add the string so that it can be manually translated.

kristinalim:
Use more specific translation key for the "Store" link in the admin header

#### What? Why?

Closes #2632

We use the `spree_i18n` for base Spree translations.

Its language translations for the "store" key is sometimes like "shop" and sometimes like "save". It's possible there is a clash in usage between Spree add-ons, so websites have been using it in different ways.

To make this more specific and to allow custom translation, I changed the "Store" link in the admin section to use the new translation key "admin.header.store".

#### What should we test?

Go to admin section and check the "Store" (`en`) link in header.

This should be properly translated in non-English instances, especially for France for which the correct translation will not match the one used before this PR.

#### Release notes

- Fix translation for "Store" link in admin header for some languages

Changelog Category: Fixed


app/views/spree/layouts/admin/_login_nav.html.haml
```diff
@@ -11,4 +11,4 @@
       = link_to t(:logout), spree.logout_path
     %li{"data-hook" => "store-frontend-link"}
       %i.icon-external-link
-      = link_to t(:store), spree.root_path, :target => '_blank'
+      = link_to t("admin.header.store"), spree.root_path, target: "_blank"
```
sauloperez: could you check if Rails' [lazy lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup) would work with `t('.header.store')` here? Using lazy lookup is one of the conventions we have for i18n. See https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29#development

kristinalim: Thanks, @sauloperez. I updated this, and it's now `spree.layouts.admin.header.store` (but referencing through `.header.store`). Right now, I'm questioning whether we should still have `spree` there, but I find it acceptable.

Issue #2008, opened 2017-12-21
Make shipping_method.zones work correctly


sauloperez:
https://github.com/openfoodfoundation/openfoodnetwork/pull/1627/files

- [ ] Document EnterpriseHelper#available_shipping_methods
- [ ] Document EnterpriseHelper#available_payment_methods

luisramos0:
The first step was to remove within zone as in #1627 . I do that with new #2652 

I dont think I understand the checklist above:
- Document EnterpriseHelper#available_shipping_methods
- Document EnterpriseHelper#available_payment_methods

There are two parts to this:
- I don't think filtering shipping methods by zone will be done in EnterpriseHelper (this will filter methods by enterprise), spree must be doing it somewhere
- make sure the original customization works (if it is supposed to work)
       - do shipping methods need to appear before shipping address is selected?         
       - if so, make sure that is working after the upgrade

sauloperez:
I don't understand the checklist either :trollface: I guess these methods were not clear to me.

sigmundpetersen:
I don't know if this one should have been closed, but the connected PRs were merged. @luisramos0 could you reopen and put it were it belongs if it's not to be closed?

luisramos0:
The simple approach is to keep this open until we:

make sure the original customization works (if it is supposed to work)
- do shipping methods need to appear before shipping address is selected?
- if so, make sure that is working after the upgrade

luisramos0:
@sigmundpetersen 👍 removing myself from the assignee as this can be taken by anyone at some point in the future.

luisramos0:
one of the things we need to do regarding shipping_method zones is to always pass zone tests on pickup shipping_methods.

One of the zone validations done on shipping_methods is in shipphing_method.include?
We can decorate this method with something like this:

app/models/spree/shipping_method_decorator.rb

```
+  def include_with_pickup?(address)
+    return true unless require_ship_address
+    include_without_pickup?(address)
+  end
+  alias_method_chain :include?, :pickup
```

sauloperez:
We purposefully wanted to get rid of this kind of metaprogramming as it's not explicit and it goes out of hand real soon. That's why we started refactoring our customizations with extension hooks like https://github.com/openfoodfoundation/spree/pull/6

luisramos0:
fair enough, are you giving this one it a try with hooks?

[...]

PR #2652, opened 2018-09-06
[Spree 2 Upgrade] Removed unused shipping_method.within_zone, customization not needed

luisramos0:
#### What? Why?

This is the same as PR #1627 but now against the spree upgrade branch 2-0-stable.

Part of #2008 

We need to remove this customization in shipping_method.
This was verifying if a shipping_method is within the zone of the address selected. This is not done here in spree 2.
Next step is to clarify and fix:
- where this is done in spree 2 by default
- where will that be done in ofn2 (including ofn customization/filtering of methods per enterprise)

#### What should we test?
Cant be tested right now.

Part of adapting to the new multi shipments of spree 2. See issue and its epic.

### DIFF
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -42,14 +42,6 @@ Spree::ShippingMethod.class_eval do
     ]
   end
 
-  def within_zone?(order)
-    if order.ship_address
-      zone && zone.include?(order.ship_address)
-    else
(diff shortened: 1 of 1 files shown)


luisramos0:
Yeah, I understand that @mkllnk ! I think in this specific case of the upgrade, we are better off removing wrong code and having a list of things to fix in the end. In this case, the issue 2008 is not being close, but this removal is certainly correct. Thanks :+1:

PR #2625, opened 2018-08-31
Rename "entreprise" in i18n keys to "enterprise"

kristinalim:
#### What? Why?

Closes #2443

Inconsistent use between the two could cause confusion and possibly bugs.

```shell
$ git grep entreprise | grep ^config/locales/en.yml | wc -l
0
$ git grep entreprise | grep -v config/locales/ | wc -l
0
```

#### What should we test?

This does not involve changes to how the software works, but needs some actions per language on Transifex.

Smoke test of customer purchasing items, looking at places where there seem to be missing translations. Best if this can be tested in another language so missing translation keys would be more obvious.

#### Release notes

- Rename translation keys that use "entreprise" to use "enterprise" instead

Changelog Category: Changed

PR #2660, opened 2018-09-07
Translate and fix fallback errors when creating or updating OC

### LINKED ISSUE #2641: Translation missing in error message / OC creation
## Description
Found out while testing #2625
The "failed to create order cycle" message is not translated :

## Steps to Reproduce
Log in as an admin on a non English instance, go to Order Cycle menu, create a new OC and try to make mistakes (don't fill all the field for instance).

(the bug is in app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee:159 and 174)

## Severity
S4

## Your Environment
* Version used: v1.18

kristinalim:
#### What? Why?

Closes #2641

This translates the fallback errors when creating or updating an OC fails. _Fallback_, because some errors that we explicitly support in the code (e.g. when close date is set before open date) give a more specific error.

Then, for the error when updating, this was actually saying "Failure to _create_ order cycle" (should be "update"). This has also been addressed.

#### What should we test?

One way I let these errors appear is by removing authorization of a user for an enterprise while the user is filling in the OC form.

@RachL if you know a simpler way, kindly share. It's likely that when you encountered these errors after leaving some fields blank was only because staging [had inconsistent data](https://github.com/openfoodfoundation/openfoodnetwork/pull/2625#issuecomment-418390260) as @luisramos0 found. I couldn't get these errors to appear that way.

1. Browser Session 1: Add test user as manager to an enterprise.
2. Browser Session 2: As test user, open the new OC page and fill in the form, but do not submit.
3. Browser Session 1: Remove test user as manager for the enterprise.
4. Browser Session 2: Being unauthorized for the enterprise now, test user should get the translated error when submitting the form.

Repeat the same for editing an OC.

#### Release notes

- Fix and translate fallback errors when creating and updating an order cycle

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee
@@ -156,7 +156,7 @@ angular.module('admin.orderCycles').factory 'OrderCycle', ($resource, $window, S
         if response.data.errors?
           StatusMessage.display('failure', response.data.errors[0])
         else
-          StatusMessage.display('failure', 'Failed to create order cycle')
+          StatusMessage.display('failure', t('js.order_cycles.create_failure'))
 
     update: (destination, form) ->
       return unless @confirmNoDistributors()
@@ -171,7 +171,7 @@ angular.module('admin.orderCycles').factory 'OrderCycle', ($resource, $window, S
         if response.data.errors?
           StatusMessage.display('failure', response.data.errors[0])
         else
-          StatusMessage.display('failure', 'Failed to create order cycle')
+          StatusMessage.display('failure', t('js.order_cycles.update_failure'))
 
 
     confirmNoDistributors: ->
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2565,7 +2565,9 @@ See the %{link} to find out more about %{sitename}'s features and to start using
         This will set stock level to zero on all products for this
         enterprise that are not present in the uploaded file.
     order_cycles:
+      create_failure: "Failed to create order cycle"
       update_success: 'Your order cycle has been updated.'
+      update_failure: "Failed to update order cycle"
       no_distributors: There are no distributors in this order cycle. This order cycle will not be visible to customers until you add one. Would you like to continue saving this order cycle?'
     enterprises:
       producer: "Producer"


app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee
```diff
@@ -156,7 +156,7 @@ angular.module('admin.orderCycles').factory 'OrderCycle', ($resource, $window, S
         if response.data.errors?
           StatusMessage.display('failure', response.data.errors[0])
         else
-          StatusMessage.display('failure', 'Failed to create order cycle')
+          StatusMessage.display('failure', t('js.order_cycles.create_failure'))
```
luisramos0: I am not a big fan of this js namespace in en.yml.
I don't think we have a convention and we have a mix of js. entries and non-js entries being used in angular templates.
I think we need to define a convention and document it here:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29

I'd vote for not using the js namespace and do what's done in admin.subscriptions for example.
Basically, I think these entries should be under admin.order_cycles together with the translations of that page, no matter if it's rails views or JS.

what do you guys think @kristinalim @mkllnk ?

luisramos0: I was trying to check if this separation between rails and js translations is encouraged or recommended in i18n-js, but I dont think it is.

mkllnk: The idea of the js namespace was that it's clear that these keys are used in Javascript. If we had a complete js namespace, we wouldn't need to include the whole locale and all translations in the assets. We could reduce the size of the assets a lot.

If we decide that it's unrealistic to separate these namespaces, we can choose another namespace. If I understand your comment, you are suggesting `admin.orderCycles.create_failure` here, right?

I definitely agree that we should standardise it so that we all go in the same direction. Thank you for bringing it up.

luisramos0: yes, I am suggesting "admin.order_cycles.create_failure" which is together with the rest of the translations for this page. Or even "admin.order_cycles.form.create_failure" or "admin.order_cycles.edit.create_failure".

I was not aware of that advantage of the js namespace. Size of assets. It's a very good point!
Do you know how to split the files so the assets pipeline only picks up the js namespace? If that is possible, I'd change my vote actually. It'd be great to remove most translations from the assets.
So, I am changing my vote to agree with this commit and make it a convention. We can create issues to fix the main problems, for example, subscriptions. And we can create a "project" (maybe a wiki page?), to separate and remove most translations from assets.

Do you guys agree?

sauloperez: before we invest any time in it, do we know how much time reduction we get precompiling the assets?

mkllnk: @luisramos0 

> Do you know how to split the files so the assets pipeline only picks up the js namespace?

Yes. It's configurable: https://github.com/fnando/i18n-js#export-configuration-for-translations

@sauloperez 
> do we know how much time reduction we get precompiling the assets?

No. And I don't think it would be much. We were more discussing the speed-up for clients which don't have to download all translations. I reckon that is several hundred kb.

luisramos0: yes, the advantage is smaller assets for the client.
Locally, I have 4 languages enabled, I copy/pasted the I18n.translations line on all.js to a file:   768kb :-)

kristinalim: Nice point about reducing file size, @mkllnk. @luisramos0 I agree that it should be a convention.

Not sure @luisramos0 if you were talking about gradually separating the translations. If yes, IMO it would be most efficient to do the separation in one go. It sounds like a lot of testing needed otherwise, or a risk that it will take a really long time until we can benefit from the client-side just using its own namespace.

kristinalim: Hmm... I'm wondering again about the download size, if there would be that big of a difference if the production server has Gzip compression enabled... But even then, browser memory would be very happy. :grinning:

luisramos0: yes, I am ok with any implementation approach actually :-)

I checked the sizes a bit better. Just the en translations are 190k on the client, 2600 translations. It's ~7kb per 100 translations.
We have ~200 lines on js: right now, let's say this will move to 400 or 600 lines after migration. That will be 30kb or 40kb total.
We are talking about reducing 150kb (per active language in the instance).
We have to do this!

luisramos0: good point. I zipped the 190kb file, I get 55kb :-)
according to this, we would still get ~40kb per language.

luisramos0: Let's do it!

kristinalim:
I created issues #2721 and #2722, and also added the section ["Javascript translations: Put translation key under the "js." namespace"](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)#javascript-translations-put-translation-key-under-the-js-namespace) to the i18n wiki page.

There are already 2 code review approvals here, so moving to "Test Ready".

sstead:
**Testing notes**

The error messages appeared correctly in English. One note, the message read 'failed to create OC' regardless of whether the use is trying to create or edit an order. I don't think this really matters, but I suppose one could say 'failed to edit OC'

https://docs.google.com/document/d/1Ikt8tJ1gYX-a1Mmtpgp6B99XbapRz5y92wicmrQAnTA/edit#heading=h.xw2hpxaxawc9

mkllnk:
There should be two error messages:
- "Failed to create order cycle"
- "Failed to update order cycle"

Actually, I just checked the code and if see only the create error, then this pull request doesn't work properly. Let me check.

mkllnk:
I deployed it again and re-tested: it works.
screenshot from 2018-09-14 16-51-59

sstead:
Testing Notes

There are unique error messages now- "failed to update" and "failed to create"

https://docs.google.com/document/d/1Ikt8tJ1gYX-a1Mmtpgp6B99XbapRz5y92wicmrQAnTA/edit#

luisramos0:
hey @kristinalim I just saw the issues to move translation to js namespace and also the wiki update. great stuff, thanks!

Issue #2414, opened 2018-06-28
Translation missing in order cycle page


RachL:
## Description
Some translations are missing here :
![image](https://user-images.githubusercontent.com/6525576/42028742-704627c2-7acd-11e8-88f4-c4fa36488178.png)

## Steps to Reproduce
Go to order cycle page : /admin/order_cycles

## Context
A French user reported it. 

## Severity
S4

## Your Environment
* Version used: v1.15.2

daniellemoorhead:
Hey @RachL @myriamboure @HugsDaniel who looks after these kind of translation problems that are found in France? Is there a process to ensure that translation issues are quickly fixed? This seems pretty important for non-english speaking countries using the platform...

myriamboure:
@daniellemoorhead as discussed this morning we need some process, but for now we use them as good first issues to onboard new contributors, that helps them understand the Transifex logic and internationalization issues.

kristinalim:
Noting here that this has been addressed in #2659, #2661, #2671 It so happens that the nature of the translation issues are different, so although simple, I separated the PRs in case some might get merged quicker than the others.

PR #2661, opened 2018-09-07
Translate "COPY OF" when cloning an OC

kristinalim:
#### What? Why?

Addresses part of #2414

An OC is cloned as "COPY OF original_oc_name" even in non-English instances.

#### What should we test?

Test cloning an OC in a non-English instance. The name of the new OC should be translated.

#### Release notes

- Translate name of the new order cycle when duplicating an order cycle

Changelog Category: Fixed

### DIFF
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -121,7 +121,7 @@ class OrderCycle < ActiveRecord::Base
 
   def clone!
     oc = self.dup
-    oc.name = "COPY OF #{oc.name}"
+    oc.name = I18n.t("models.order_cycle.cloned_order_cycle_name", order_cycle: oc.name)
     oc.orders_open_at = oc.orders_close_at = nil
     oc.coordinator_fee_ids = self.coordinator_fee_ids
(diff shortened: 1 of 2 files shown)


mkllnk:

RachL:
As it is staged on an English instance, I can only check if nothing is broken which is correct as I still see "COPY OF" as usual:
file:///home/rachel/Images/Capture%20d%E2%80%99%C3%A9cran%20de%202018-09-17%2011-39-38.png

sauloperez:
Unless we don't translate it and stage it on a non-english isntance we won't see it working. We can either merge it as apparently doesn't break the app or translate it in a different language, pull translations, merge them here and then stage again. What do you vote for @RachL ?

sigmundpetersen:
But doesn't this PR need to be merged to master before it appears in Transifex? 🙂

RachL:
@sauloperez yes don't worry, I was just detailing how I've tested it since @kristinalim put in the testing process that it had to be tested on an non English instance.
Let's merge this and we will see in production how it translate, worse case scenario it does not and we are in same situation as today, but so far as long as nothing was broken translation have always worked. So let's move forward as would a usual PR go.

PR #2667, opened 2018-09-08
[Spree Upgrade] Adapt Address_Finder.last_used_ship_address to spree 2

### LINKED ISSUE #2663: [Spree Upgrade] Adapt Address_Finder.last_used_ship_address to spree 2
order.shipping_method relation is gone. The calculation of last_used_ship_address must use order.shipments.

See lib/open_food_network/address_finder.rb:79
and respective spec/lib/open_food_network/address_finder_spec.rb

This breaks quite a few tests in models/order_spec

luisramos0:
#### What? Why?

Closes #2663 

Finding last used address now uses the relation order#shipments and not order#shipping_method
Also, fixed respective spec by replacing reference to order.shipping_method_id with order.shipping_method (the new adapter to get to order.shipments).

#### What should we test?
Together with #2727, this makes address_finder_spec green and fixes quite a few tests in models/order_spec and others.

### DIFF
--- a/lib/open_food_network/address_finder.rb
+++ b/lib/open_food_network/address_finder.rb
@@ -79,7 +79,7 @@ module OpenFoodNetwork
 
     def last_used_ship_address
       return nil unless allow_search_by_email?
-      Spree::Order.complete.joins(:ship_address, :shipping_method).order('id DESC')
+      Spree::Order.complete.joins(:ship_address, shipments: :shipping_methods).order('id DESC')
         .where(email: email, spree_shipping_methods: { require_ship_address: true })
         .first.andand.ship_address
     end
--- a/spec/lib/open_food_network/address_finder_spec.rb
+++ b/spec/lib/open_food_network/address_finder_spec.rb
@@ -86,7 +86,7 @@ module OpenFoodNetwork
     describe "last_used_bill_address" do
       let(:distributor) { create(:distributor_enterprise) }
       let(:address) { create(:address) }
-      let(:order) { create(:completed_order_with_totals, user: nil, email: email, distributor: distributor) }
+      let(:order) { create(:completed_order_with_totals, user: nil, email: email, distributor: distributor, bill_address: nil) }
       let(:finder) { AddressFinder.new(email) }
 
       context "when searching by email is not allowed" do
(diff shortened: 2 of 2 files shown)


lib/open_food_network/address_finder.rb
```diff
@@ -79,7 +79,7 @@ def last_used_bill_address
 
     def last_used_ship_address
       return nil unless allow_search_by_email?
-      Spree::Order.complete.joins(:ship_address, :shipping_method).order('id DESC')
+      Spree::Order.complete.joins(:ship_address, shipments: :shipping_methods ).order('id DESC')
```
sauloperez: :+1:

luisramos0:
2 approvals, moving to ready to go.
@mkllnk let me know if there was a reason to keep it in code review.

lib/open_food_network/address_finder.rb
```diff
@@ -79,7 +79,7 @@ def last_used_bill_address
 
     def last_used_ship_address
       return nil unless allow_search_by_email?
-      Spree::Order.complete.joins(:ship_address, :shipping_method).order('id DESC')
+      Spree::Order.complete.joins(:ship_address, shipments: :shipping_methods).order('id DESC')
```
sauloperez: Did this join between shipments and shipping_methods show up somewhere else already? In that case it might be the right time to encapsulate it in an AR scope. This way we could refer back to it but in a decoupled way.

luisramos0: yes, it is coming up here as well: https://github.com/openfoodfoundation/openfoodnetwork/pull/2680/commits/75656668cdaade2bbc2a994f1b38642b73734b03

it's only 2 occurrences and they are different because one joins to ship_methods and the other joins to ship_methods and ship_address. this is a reason no to do it.

but the real reason I didnt do it it's because I dont know how to do it. can you be a bit more specific about what would you like to see here? sorry for my AR ignorance. thanks!

luisramos0: is this a new line on order_decorator?
has_many :shipping_methods, through: :shipments

sauloperez: In that case it may not be worth doing anything but FYI https://guides.rubyonrails.org/v3.2/active_record_querying.html#scopes

luisramos0: it would be
scope :with_shipping_methods, lambda {
    joins( ‘shipments: :shipping_methods’ )
end

ah, the merge one is nice.
thanks a lot for the links!!
ok, we keep it as it is. we will do this on a third occurrence for sure.

PR #2668, opened 2018-09-08
[Spree 2 Upgrade] Fixed models/order_spec issues related to order.shipping_method

### LINKED ISSUE #2664: [Spree Upgrade] Fix issues related to shipments in order_spec
Quite a few errors related to shipments in the order_spec.
We need to adapt to spree 2 shipments model
spec/models/spree/order_spec.rb

luisramos0:
#### What? Why?

Closes #2664 

#### What should we test?
Fixes tests in models/order_spec.

Adapts order_spec to spree 2 - order multi shipments.

### DIFF
--- a/spec/controllers/line_items_controller_spec.rb
+++ b/spec/controllers/line_items_controller_spec.rb
@@ -91,10 +91,9 @@ describe LineItemsController, type: :controller do
     end
 
     context "where shipping and payment fees apply" do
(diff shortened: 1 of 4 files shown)


luisramos0:
I think it's better we review and merge this PR although it is still not fixing all the tests in the spec. There is more work to be done, even in terms of shipments, we still need to make the tests related to shipping fees pass. I am moving this to code review because 1. this change makes sense as is, 2. it's a valid step in the right direction 3. avoids a PR that is too big. We can open a new PR later to finish fixing this complex spec.

PR #2654, opened 2018-09-06
[Spree 2 Upgrade] - Added adapted order.shipping_method

luisramos0:
#### What? Why?

Closes #2009 

WIP: This is the quick fix, it still needs cleaning up, docs and tests.

#### What should we test?
More specs fixed.

#### Release notes
Changelog Category: Fixed
Adapted to spree 2 multi shipments by adding adapted shipping method to order.

#### How is this related to the Spree upgrade?
This is part fo the spree upgrade.

PR #2669, opened 2018-09-08
[Spree 2 Upgrade] WIP - Added adapted order.ship_method=

### LINKED ISSUE #2662: [Spree Upgrade] Replace Order#shipping_method=
`Order#shipping_method=` is gone in Spree 2.0, so any line of code that uses it in OFN will fail. We can provide the same behaviour with an adapter to spree 2 models.

Related to #2009 that handles the absence of order.shipping_method

luisramos0:
#### What? Why?

Closes #2662

#### What should we test?
Spree upgrade issues cannot be tested yet.

Part of order multi-shipments epic.

#### Dependencies
Related to #2654

### DIFF
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -336,8 +336,13 @@ Spree::Order.class_eval do
   # Although Spree 2 supports multi shipments per order, in OFN we keep the rule one shipment per order
   #   Thus, this method sets the given shipping_method on the first and only shipment in the order
   def shipping_method=(shipping_method = nil)
-    return if shipments.empty?
-    shipments.first.shipping_method = shipping_method
+    if shipments.empty?
+      shipment = Spree::Shipment.new
+      shipment.order = self
+      shipment.save
+      shipments << shipment
+    end
+    shipments.first.add_shipping_method(shipping_method, true)
   end
 
   private
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -333,6 +333,13 @@ Spree::Order.class_eval do
     payments.select {|p| p.state == "checkout"} # Original definition
   end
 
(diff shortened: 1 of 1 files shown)


app/models/spree/order_decorator.rb
```diff
@@ -333,6 +333,18 @@ def pending_payments
     payments.select {|p| p.state == "checkout"} # Original definition
   end
 
+  # Although Spree 2 supports multi shipments per order, in OFN we keep the rule one shipment per order
+  #   Thus, this method sets the given shipping_method on the first and only shipment in the order
+  def shipping_method=(shipping_method = nil)
+    if shipments.empty?
+      shipment = Spree::Shipment.new
+      shipment.order = self
+      shipment.save
```
sauloperez: AFAIK we don't need this save as AR persists the order and all its associations when doing a `order.save`. Correct me if I'm wrong.

app/models/spree/order_decorator.rb
```diff
@@ -333,6 +333,18 @@ def pending_payments
...
+      shipments << shipment
+    end
+    shipments.first.add_shipping_method(shipping_method, true)
+  end
```
sauloperez: Have we investigated how this works with the way Spree creates shipments in https://github.com/openfoodfoundation/spree/blob/fe0a1311abb097bf3ac8001a24246c6cff5ecdbb/core/app/models/spree/order.rb#L506-L515 ? That's the only place I know in Spree's codebase where shipments are created.

luisramos0: yes, I have seen that code. I am closing this PR actually, we need to fix the specs and then we will see if this is still required.

luisramos0:
I am closing this PR for now, I dont think we will need this adapter. We will have to replace all places where shipping_method= is used. Mostly specs as listed in #2662

PR #2659, opened 2018-09-07
Simple translation fixes for admin area and profile page

kristinalim:
#### What? Why?

- Closes #2630 - Additionally, translates the "Package" and "Status" column headers.
- Closes #2629
- Some work for #2414 - Fixes the placeholder for the OC name search field and also the "Involving" filter label. "Any Enterprise" and "COPY OF" are more complicated in nature, and belong to another pull request.
- Related to #2646 - That issue seems to be fixed already. However, I noticed "Shop Title" and "Allow Charges" are not translated - these have been addressed in this PR.

Please also note that I changed "Shop Title" to "Shop Name". It seems more appropriate.

#### What should we test?

Do below in a translated, non-English OFN instance. For convenience, you can log in as enterprise manager when checking these.

- Enterprise index: Check the "Producer?", "Package", and "Status" column headers.
- Product form: Go to "Properties". Check the checkbox label for inheriting properties from supplier.
- OC index: Check the place holder for OC name search field and the "Involving" filter label.
- "Credit Cards" section in customer account page: Check the table heading for "Authorised Shops" and its tooltip popover. Check the column headers for "Authorised Shops".

#### Release notes

- Translate more text in the enterprise index, order cycle index, product form, and customer account page.

Changelog Category: Fixed

PR #2671, opened 2018-09-09
Translate "Any Enterprise" and "Any Schedule" in OC filters

kristinalim:
#### What? Why?

Fixes part of #2414, including translation of the "Any Schedule" option for the "Schedule" filter.

I should have included this #2659, but I was considering whether we should have a helper method for converting a hash in Ruby to a hash used in an HTML tag attribute. This seems risky though and has potential to be misused, so I didn't do that now because for now the risks seem to outweight the benefit.

#### What should we test?

Test this in a non-English instance.

As admin:

1. Go to the OC index.
2. Check the "Involving" filter. The first drop-down option "Any Enterprise" should be translated.
3. Check the "Schedule" filter. The first drop-down option "Any Schedule" should be translated.

#### Release notes

- Translate more text in the order cycle index

Changelog Category: Fixed

#### Dependencies

~This includes changes in #2659. That should be merged first and this could be rebased, or this could be merged into that branch before that branch is merged.~ (**Update**: This is no longer the case.)

### DIFF
--- a/app/views/admin/order_cycles/_filters.html.haml
+++ b/app/views/admin/order_cycles/_filters.html.haml
@@ -6,12 +6,12 @@
   .filter_select.four.columns
     %label{ :for => 'involving_filter' }=t('.involving')
     %br
-    %input.ofn-select2.fullwidth{ :id => 'involving_filter', type: 'number', blank: "{id: 0, name: 'Any Enterprise'}", data: 'enterprises', ng: { model: 'involvingFilter' } }
+    %input.ofn-select2.fullwidth{ id: 'involving_filter', type: 'number', blank: "{id: 0, name: '#{j t(".any_enterprise")}'}", data: 'enterprises', ng: { model: 'involvingFilter' } }
   - if subscriptions_enabled?
     .filter_select.four.columns
       %label{ :for => 'schedule_filter' }=t('admin.order_cycles.index.schedule')
       %br
-      %input.ofn-select2.fullwidth{ :id => 'schedule_filter', type: 'number', blank: "{id: 0, name: 'Any Schedule'}", data: 'schedules', ng: { model: 'scheduleFilter' } }
+      %input.ofn-select2.fullwidth{ id: 'schedule_filter', type: 'number', blank: "{id: 0, name: '#{j t(".any_schedule")}'}", data: 'schedules', ng: { model: 'scheduleFilter' } }
     .two.columns &nbsp;
   - else
     .six.columns &nbsp;
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -897,6 +897,8 @@ en:
       filters:
         search_by_order_cycle_name: "Search by Order Cycle name..."
         involving: "Involving"
+        any_enterprise: "Any Enterprise"
+        any_schedule: "Any Schedule"
       form:
         incoming: Incoming
         supplier: Supplier


app/views/admin/order_cycles/_filters.html.haml
```diff
@@ -6,7 +6,7 @@
   .filter_select.four.columns
     %label{ :for => 'involving_filter' }=t('.involving')
     %br
-    %input.ofn-select2.fullwidth{ :id => 'involving_filter', type: 'number', blank: "{id: 0, name: 'Any Enterprise'}", data: 'enterprises', ng: { model: 'involvingFilter' } }
+    %input.ofn-select2.fullwidth{ :id => 'involving_filter', type: 'number', blank: "{id: 0, name: '#{j t(".any_enterprise")}'}", data: 'enterprises', ng: { model: 'involvingFilter' } }
```
luisramos0: That j in #{j is a typo or am I about to learn something I didnt know?

app/views/admin/order_cycles/_filters.html.haml
```diff
@@ -11,7 +11,7 @@
     .filter_select.four.columns
       %label{ :for => 'schedule_filter' }=t('admin.order_cycles.index.schedule')
       %br
-      %input.ofn-select2.fullwidth{ :id => 'schedule_filter', type: 'number', blank: "{id: 0, name: 'Any Schedule'}", data: 'schedules', ng: { model: 'scheduleFilter' } }
+      %input.ofn-select2.fullwidth{ :id => 'schedule_filter', type: 'number', blank: "{id: 0, name: '#{j t(".any_schedule")}'}", data: 'schedules', ng: { model: 'scheduleFilter' } }
```
luisramos0: again, the same mystery j, hmmm...

app/views/admin/order_cycles/_filters.html.haml
sauloperez: No idea either

sauloperez: would you mind replacing that `:id` has rocket for consistency?

luisramos0: amazing, I'm learning!
Are you saying that we should use this j everywhere in all rails translations that are used by some form of js code in the process? 
We should document this in i18n wiki page.

kristinalim: @sauloperez I didn't notice that it was using the two hash syntaxes in the same line! Thanks, I appended a commit replacing this.

luisramos0: ping :-)
Can you please clarify "Are you saying that we should use this j everywhere in all rails translations that are used by some form of js code in the process?"

kristinalim: So, while starting to make changes to the wiki, I realized that I could use the Angular filter `t` here instead. I revised this PR. (There are now no more conflicts with #2659, so this branch no longer includes the changes there.)

The I18n wiki page already says to prefer this filter, so I did not edit that. I updated the Code Conventions doc though to remind developers to [use `j` when inserting an arbitrary Ruby string to a JS or Angular expression](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code-Conventions#use-j-when-inserting-an-arbitrary-ruby-string-in-a-js-or-angular-expression).

Requesting another review, sorry!

luisramos0: awesome, thank you!

luisramos0: hey, I think you misunderstanding the docs just as I did a few weeks ago!
I am not sure about this BUT the convention is that you should use the filter and not use the JS global t function in JS! BUT you should prefer rails translations (with lazy load) to js translation. Correct @mkllnk ?

luisramos0: imo, in this case I think you should keep the rails translation, not js.

mkllnk: Sorry for the confusion. We have to improve the documentation somehow.

- This is a Rails view. So we should use Rails' `I18n.t` method:
```haml
%input.ofn-select2.fullwidth{ :id => 'involving_filter', type: 'number', blank: "{id: 0, name: '#{t(".any_enterprise")}'}", data: 'enterprises', ng: { model: 'involvingFilter' } }
```
- In Angular views, we should use the `t` filter.
- In other Angular/JS code where filters are not easily accessible, we should use Javascript's `I18n.t` function.

kristinalim: I updated the wiki to reflect that we should still prefer the Ruby method in Rails views: ["Rails view translations: Prefer Ruby `t` method"](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)#rails-view-translations-prefer-ruby-t-method).

This PR has been updated as well. :slightly_smiling_face: This is still in "Code Review". Kindly just check again, then move to "Test Ready" when okay.

luisramos0:
Staged in https://staging.openfoodfrance.org

RachL:
Nothing broken here :) Moving to ready to go. Here is the screenshot:
capture d ecran de 2018-09-18 11-25-39

PR #2675, opened 2018-09-11
Remake orders index view

### LINKED ISSUE #2525: [Bulk invoice printing] Recreate the Spree view without deface overrides
## Description

Part 1 of the Bulk Invoice Printing epic.

As a small issue in preparation for the Bulk Invoice Printing epic, we need to recreate our own views for the admin Orders page. This means taking the Spree view, applying anything from the deface overrides we have previously created, adding them together into new view file and deleting the old overrides.

## Acceptance criteria

A new view should be created for the Orders page in the backend. Visually it should be the same as before.

Matt-Yorkley:
#### What? Why?

Closes #2525

Recreates the orders index view as a new HAML file with no deface overrides.

#### What should we test?

The admin orders index page should be working OK and look the same as before.

#### Release notes

No release notes.


Matt-Yorkley:
Also fixed a little filter alignment thing that was annoying me.

 ------- Before: -------
screenshot from 2018-09-11 12-22-11
 ------- After: -------
screenshot from 2018-09-11 12-22-16

sigmundpetersen:
Hey @Matt-Yorkley I don't think we add PRs to the epic, let's see what train driver @daniellemoorhead says

luisramos0:
Awesome, I believe this is the origin, correct me if I am wrong:
https://github.com/openfoodfoundation/spree/blob/step-6a/backend/app/views/spree/admin/orders/index.html.erb

sigmundpetersen:
@kristinalim just from earlier practice :) I don't know if there is an official process on it or if this specifically has been discussed. I'm OK with both approaches. @daniellemoorhead would have further insights 👍

Matt-Yorkley:
I thought connecting it just helped filtering PRs by epic in zenhub, but I'm not sure what else it's used for...

Matt-Yorkley:

app/assets/stylesheets/admin/orders.css.scss
```diff
@@ -87,3 +87,7 @@ div#group_buy_calculation {
 th.actions {
   white-space: nowrap;
 }
+
+table.index td.actions {
+  text-align: left;
+}
```
sauloperez: I'm not a CSS expert myself but Code Climate complaints about the use of the `table` and `td` element selectors.

RachL:
Did a sanity check on orders page. Everything looks as usual :)
Let's move forward!
Testing notes : https://docs.google.com/document/d/1g-MacYfRGjyhJ3ucsel1fTPatkQ_-QnUeTUwPUx_DMo/edit#

daniellemoorhead:
OK, so my thoughts @Matt-Yorkley @kristinalim @sigmundpetersen:

There isn't an official process, just what we've done in the past. As far as I'm aware we've never connect PRs to epics, we've connected them to the issues that are already in the epic. As far as I'm aware (correct me if wrong) by connecting the PR to the issue it should still display when you filter by epic on the board. 

What this doesn't allow for are multiple PRs for the same issue. Not sure what to do about these, and they do happen fairly frequently. But we don't have a solution for this

Let's think about what we're trying to achieve in terms of issue/epic hierarchy and tracking: 

- We have a "parent" epic that is the feature candidate, and that is kept on the product feature backlog project board so we can track the high level prioritised features through the delivery pipe.

- We have child epics underneath that parent which are used to group different parts of the work to be done (e.g. mobile navigation changes, or mobile shopping changes). We don't track these on the feature backlog, nor in the same way as an issue to be built (next dot point) _so perhaps this is a gap in our process as to where we should track them? But do they need to be tracked, can the product owner look after their status as part of the role?_

(Note (for @sigmundpetersen because I reckon you read that paragraph above and were thinking, "but..." 😆): Spree upgrade has been the exception to this but it's MAMMOTH and the idea is that a feature that's selected isn't EVER as big as this (bite off small chunks of features at a time whereas this is tech debt on a massive scale). 

- We have issues/stories that are moved from dev ready to in dev, etc, and built (as part of a feature, or tech debt, or an orphan bug). They are tracked on the zenhub delivery train board by the delivery team to ensure they are moving across, and if the product owner has set up a project for that feature then on there too (yep, duplication).

- PRs (for features) are connected to the issue that they resolve using zenhub. We use the zenhub delivery train board to track their progress across the different stages, with their issue else as orphans. The limited 1-1 aspect of linking PRs to issues is a problem when there are multiple issues and you're tracking issues with attached PR across the board. _Perhaps this problem means we should add PRs to epics...but I still don't like the idea of clogging a feature epic with PRs when it's the delivery of the issue that is most important.

And that is the end of my (very long) response.

mkllnk:
@daniellemoorhead @sigmundpetersen 

Connecting multiple PRs with one issue is possible, but in most cases it has a bad side effect. When one PR solves only part of an issue I don't connect it to the PR and I don't mention "closes #n". Otherwise the issue would get moved along with the PR and closed even though it's only partially solved.

Ideally we would create multiple issues if we need multiple PRs, but sometimes it doesn't seem worth the effort. Sometimes one PR is just refactoring to prepare the solution. So I think this is acceptable.

The last PR to completely solve an issue should be connected to it so that it gets closed.

Issue #2009, opened 2017-12-21
[Spree 2 Upgrade] Replace Order#shipping_method with Order#shipments 


sauloperez:
`Order#shipping_method` is gone in Spree 2.0, so any line of code that uses it in OFN will fail. We can provide the same behaviour but being spree-2.0 compatible without having to wait for a specific spree-2.0 branch :metal:

The work has already been started in: https://github.com/openfoodfoundation/openfoodnetwork/pull/1651

sauloperez:
It might to big of a PR to get all this done in one shot. Let's first investigate, post the outcomes and then we decide. We could create individual issues for each them and fix them bit by bit.

luisramos0:
Note from #2523 - when removing line items from a completed/finalised order, the shipments/stock_locations are updated in the order.contents.remove_variant method. This needs to be adapted. See #2523 for more details.

luisramos0:
Actions from my investigation:
- I have created an issue to handle the setting of order.shipping_method (shipping_method=) #2662 
- I have created PRs to implement adapters to both order.shipping_method and order.shipping_method= (respectively #2654 and #2669 )
- I have created issues and respective PRs to fix specs broken with errors related to this change (order spec issue #2664 and PR #2668 and adjustment_spec issue #2665 and PR #2670 )
- I have created an issue and PR to adapt last_used_ship_address in Address Finder #2663 and PR 2667
I have created separate items to keep all issues focused and all PRs small.

I am requesting the validation of the approach for #2654 (and #2669) [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2654#issuecomment-419717043).

daniellemoorhead:
Hey @luisramos0 the PR attached to this issue is merged, yet this is still in the code review column...can it be closed, or are you waiting on #2662 to be merged before it's done? 

If you're waiting, then suggest disconnecting #2654 and connecting #2662 instead. Or, if as it says in #2654 that it closes #2009 then go right ahead and close this 😄

luisramos0:
Thanks for the heads up @daniellemoorhead 

I disconnected this from #2654 , 2654 is starting but not closing this one.

There's more work to be done and it's not just 2662. Actually 2662 may not be necessary.

What's still not done in this issue is to go through all the changes in the original PR #1651 and see what's still valid. This issue acceptance criteria is the absence of errors related to order#shipping_method.

luisramos0:
[...]

PR #2677, opened 2018-09-11
[Spree Upgrade] Fixed db:load_sample_data

luisramos0:
#### What? Why?

Related to #2009 

Refactored the distributor.shipping_method builder. Basically, shipping_method has no zone_id in spree 2.

#### What should we test?
We can now run

`bundle exec rake openfoodnetwork:dev:load_sample_data RAILS_ENV=development`

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -184,13 +184,12 @@ namespace :openfoodnetwork do
 
       enterprise2 = Enterprise.find_by_name('Enterprise 2')
       enterprise2.sells = 'any'
-      enterprise2.shipping_methods.build(
-        name: 'Pickup',
-        zone_id: 3,
-        require_ship_address: true,
-        calculator_type: 'OpenFoodNetwork::Calculator::Weight',
-        distributor_ids: [enterprise2.id]
-      )
+      enterprise2.shipping_methods << FactoryBot.create(:shipping_method,
+                             name: 'Pickup',
+                             zone: zone,
+                             require_ship_address: true,
+                             calculator_type: 'OpenFoodNetwork::Calculator::Weight',
+                             distributors: [enterprise2])
       enterprise2.payment_methods << Spree::PaymentMethod.last
       enterprise2.save!
 
@@ -200,7 +199,9 @@ namespace :openfoodnetwork do
 
       CreateOrderCycle.new(enterprise2, variants).call
 
-      EnterpriseRole.create!(user: Spree::User.first, enterprise: enterprise2)
+      unless EnterpriseRole.where( user_id: Spree::User.first, enterprise_id: enterprise2 ).any?
+        EnterpriseRole.create!(user: Spree::User.first, enterprise: enterprise2)
+      end
     end
   end
 end


lib/tasks/dev.rake
```diff
@@ -200,7 +199,9 @@ namespace :openfoodnetwork do
 
       CreateOrderCycle.new(enterprise2, variants).call
 
-      EnterpriseRole.create!(user: Spree::User.first, enterprise: enterprise2)
+      unless EnterpriseRole.where( user_id: Spree::User.first, enterprise_id: enterprise2 ).any?
```
luisramos0: this makes the script rerunable.

lib/tasks/dev.rake
```diff
@@ -184,13 +184,12 @@ namespace :openfoodnetwork do
...
       enterprise2 = Enterprise.find_by_name('Enterprise 2')
       enterprise2.sells = 'any'
-      enterprise2.shipping_methods.build(
-        name: 'Pickup',
-        zone_id: 3,
-        require_ship_address: true,
-        calculator_type: 'OpenFoodNetwork::Calculator::Weight',
-        distributor_ids: [enterprise2.id]
-      )
+      enterprise2.shipping_methods << FactoryBot.create(:shipping_method,
```
mkllnk: I always saw FactoryBot's purpose to support specs. It would be better to avoid it in other parts of the code to reduce dependencies. If it's difficult to build an object then that's a symptom of a bigger problem, complex models with too many responsibilities.

sauloperez: that's an interesting point. Mine is more about the fact that sample data needs to be deterministic IMO. We all (not only testers) need to always know the data that gets set in the system so that we can test without having to learn it each time from scratch.

If I know `Enterprise X` has a members-only shop and I'm asked to test a feature of those, I can go straight ahead and do my tests with it.

luisramos0: What I think of FacoryBot here is: it works, it makes our life easer. It's not production code. No problem :-)

luisramos0: If it was production code I'd recommend not using factorybot and move the building code into a "Builder pattern" object that supports object creation.

sauloperez: I don't care so much about using FactoryBot but more about the reproducibility (whathever word :troll:) of our sample data.

luisramos0: I am not sure I understand exactly what change you are requesting.
Is it to replace factorybot with something else? If that is the case, if you agree, we'd approve this PR and create a separate issue to tackle that.

sauloperez: we need a reproducible state. Every time we run sample data we need to get the same result, so we better create the shipping method without FactoryBot as the latter will always generate random data, I believe. What do you think?

luisramos0: yes, I understand that and agree. But I dont think this is specific to this change or to shipping_method. If you look at this file. It's using FactoryBot for most of the data. So, we would need to rewrite this file and verify it creates the same data all the time.

mkllnk: Oh, you are right. I didn't realise that it was using FactoryBot already. Sorry. Different scope. :-)

sauloperez: ok, then I give up :+1:

sigmundpetersen:
There are some Codeclimate issues here @luisramos0 . Do we fix or isn't this a priority for this issue?

luisramos0:
Yeah @sigmundpetersen , there are quite a few warnings about complexity. But they are not related to this change.

PR #2680, opened 2018-09-11
[Spree Upgrade] Fixed inexistent order#shipping_method= in several specs

### LINKED ISSUE #2662: [Spree Upgrade] Replace Order#shipping_method=

### LINKED ISSUE #2679: [Spree Upgrade] Replace order.shipping_method_id in current_order_serializer
Mimic what's in this commit current_order_serializer:
https://github.com/openfoodfoundation/openfoodnetwork/pull/1651/commits/06345e91615b93b75d9e1b1411b7da33fe1dd607

app/serializers/api/current_order_serializer.rb:
    1  class Api::CurrentOrderSerializer < ActiveModel::Serializer
    2:   attributes  ... :shipping_method_id,
PR - #2729

luisramos0:
Closes #2662

Instead of writing an adapter (as in #2669 ), the best solution here is to fix all specs that need this method. Production code may actually not require to set shipping_method like this.

### Test
Acc Criteira: no "undefined method order.shipping_method=" errors in specs

### DIFF
--- a/lib/open_food_network/order_cycle_management_report.rb
+++ b/lib/open_food_network/order_cycle_management_report.rb
@@ -107,7 +107,7 @@ module OpenFoodNetwork
 
     def filter_to_shipping_method(orders)
       if params[:shipping_method_in].present?
-        orders.joins(:shipping_method).where(shipping_method_id: params[:shipping_method_in])
+        orders.joins(shipments: :shipping_methods).where(shipping_method_id: params[:shipping_method_in])
       else
         orders
       end
--- a/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
+++ b/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
@@ -8,12 +8,12 @@ describe Spree::Admin::Orders::CustomerDetailsController, type: :controller do
       let!(:user) { create(:user) }
       let(:address) { create(:address) }
       let!(:distributor) { create(:distributor_enterprise) }
-      let!(:shipping_method) { create(:shipping_method) }
+      let!(:shipment) { create(:shipment) }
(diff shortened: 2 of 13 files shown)


mkllnk:
@luisramos0 May I ask why you create WIP pull requests?

luisramos0:
:-)
The first valid reason is that I need early feedback. Not the case here.
The second not-so-valid reason is that I spent some time without success trying to run semaphore builds on my branches. The PR is the easier option to run the build against a branch.

mkllnk:
@luisramos0 Have you seen my documentation of configuring Semaphore? https://github.com/openfoodfoundation/openfoodnetwork/wiki/Continuous-Integration#semaphore-ci

luisramos0:
I have gone through all the specs mentioned in issue #2662. This is ready for review.

luisramos0:
yes, we need to continuously refactor. do let me know if you are thinking of some specific refactoring, in some cases we may just do it now.
Also, we should create refactoring issues as we go. Specially for the important ones.
For example, I have just created #2763.

luisramos0:
2 approvals, moving to ready to go.

sauloperez:
> Off topic: The Spree upgrade seems to make a lot of things more complicated, because we haven't found the right way yet to bridge the OFN domain with Spree's domain. We will have to do some refactoring once it's all green.

I totally understand your point and I believe we're getting better at that but, I the tight coupling to Spree's DB schema won't be easy to overcome.

Issue #2631, opened 2018-09-01
Change link to user guide in welcome email


RachL:
## Description
This is linked to issue #2593 If we enable instances to customize their user guide URL, the URL in the enterprise welcome email should change as well:
![image](https://user-images.githubusercontent.com/6525576/44944710-0e5a2800-addc-11e8-8821-38efef53b3fb.png)

So either we can change this automatically when #2593 is done, either each instance has to change the link manualy

## Acceptance criteria
If I've customize my URL, meaning I've done this from issue 2593:
In configuration > general settings, a field should be made available to add an external link. 
(Only super admins can change this link)
When a new enterprise register, the link to the user guide should be updated too.
Example of user guide link in French: https://ofnuserguidefr.gitbook.io/guide-utilisateur-open-food-france

PR #2686, opened 2018-09-12
Enable setting a customized user guide link in general settings

### LINKED ISSUE #2593: Make it possible to customize the link to user guide per instances
## Description
Link to user guide in admin mode is hard-wired at the moment. This issue is here to make changes to the code in order to allow each instance to have their own link.
image

## Acceptance criteria
In configuration > general settings, a field should be made available to add an external link. 
Only super admins can change this link.
For managers, when clicking on the user guide button, the user guide should open in a new page (same behaviour as now).
Example of user guide link in French: https://ofnuserguidefr.gitbook.io/guide-utilisateur-open-food-france

HugsDaniel:
#### What? Why?

Closes #2593 #2631 

In configuration > general settings, a field is available to add an external link for the user guide.

screenshot from 2018-09-12 09-50-02

#### What should we test?

A superadmin can add a custom user guide link.
Non superadmins can't.

#### Release notes

An admin can now add a custom user guide link in Configuration > General Settings.

Changelog Category: Added

#### Dependencies

This adds customizations to Spree::Config.

#### Documentation updates

There could be a line added where this new feature is mentioned but I'm not sure where to put it.

### DIFF
--- a/app/controllers/admin/contents_controller.rb
+++ b/app/controllers/admin/contents_controller.rb
@@ -24,10 +24,16 @@ module Admin
     private
 
     def preference_sections
-      Dir["app/models/preference_sections/*.rb"].map do |filename|
-        basename = 'PreferenceSections::' + File.basename(filename, '.rb').camelize
-        basename.constantize.new
-      end
+      [
+        PreferenceSections::HeaderSection.new,
+        PreferenceSections::HomePageSection.new,
+        PreferenceSections::ProducerSignupPageSection.new,
+        PreferenceSections::HubSignupPageSection.new,
+        PreferenceSections::GroupSignupPageSection.new,
+        PreferenceSections::MainLinksSection.new,
+        PreferenceSections::FooterAndExternalLinksSection.new,
+        PreferenceSections::UserGuideSection.new
+      ]
     end
   end
 end
--- a/app/controllers/admin/contents_controller.rb
+++ b/app/controllers/admin/contents_controller.rb
@@ -1,15 +1,9 @@
 module Admin
   class ContentsController < Spree::Admin::BaseController
     def edit
-      @preference_sections = [{name: I18n.t('admin.contents.edit.header'), preferences: [:logo, :logo_mobile, :logo_mobile_svg]},
-                              {name: I18n.t('admin.contents.edit.home_page'), preferences: [:home_hero, :home_show_stats]},
-                              {name: I18n.t('admin.contents.edit.producer_signup_page'), preferences: [:producer_signup_pricing_table_html, :producer_signup_case_studies_html, :producer_signup_detail_html]},
-                              {name: I18n.t('admin.contents.edit.hub_signup_page'), preferences: [:hub_signup_pricing_table_html, :hub_signup_case_studies_html, :hub_signup_detail_html]},
-                              {name: I18n.t('admin.contents.edit.group_signup_page'), preferences: [:group_signup_pricing_table_html, :group_signup_case_studies_html, :group_signup_detail_html]},
-                              {name: I18n.t('admin.contents.edit.main_links'), preferences: [:menu_1, :menu_1_icon_name, :menu_2, :menu_2_icon_name, :menu_3, :menu_3_icon_name, :menu_4, :menu_4_icon_name, :menu_5, :menu_5_icon_name, :menu_6, :menu_6_icon_name, :menu_7, :menu_7_icon_name]},
-                              {name: I18n.t('admin.contents.edit.footer_and_external_links'), preferences: [:footer_logo,
-                                                             :footer_facebook_url, :footer_twitter_url, :footer_instagram_url, :footer_linkedin_url, :footer_googleplus_url, :footer_pinterest_url,
-                                                             :footer_email, :community_forum_url, :footer_links_md, :footer_about_url]}]
+      @preference_sections = preference_sections.map do |preference_section|
+        { name: preference_section.name, preferences: preference_section.preferences }
+      end
     end
 
     def update
@@ -25,6 +19,15 @@ module Admin
       flash[:success] = t(:successfully_updated, :resource => I18n.t('admin.contents.edit.your_content'))
 
       redirect_to main_app.edit_admin_content_path
+    end
+
+    private
+
+    def preference_sections
+      Dir["app/models/preference_sections/*.rb"].map do |filename|
+        basename = 'PreferenceSections::' + File.basename(filename, '.rb').camelize
+        basename.constantize.new
+      end
     end
   end
 end
--- a/app/models/content_configuration.rb
+++ b/app/models/content_configuration.rb
@@ -71,5 +71,5 @@ EOS
   preference :footer_about_url, :string, default: "http://www.openfoodnetwork.org/ofn-local/open-food-network-australia/"
 
   #User Guide
-  preference :user_guide_link, :string, default: 'http://www.openfoodnetwork.org/platform/user-guide/'
+  preference :user_guide_link, :string, default: 'https://guide.openfoodnetwork.org/'
 end
--- a/app/models/content_configuration.rb
+++ b/app/models/content_configuration.rb
@@ -69,4 +69,7 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
 EOS
 
   preference :footer_about_url, :string, default: "http://www.openfoodnetwork.org/ofn-local/open-food-network-australia/"
+
+  #User Guide
+  preference :user_guide_link, :string, default: 'http://www.openfoodnetwork.org/platform/user-guide/'
 end
--- a/spec/features/admin/content_spec.rb
+++ b/spec/features/admin/content_spec.rb
@@ -18,23 +18,35 @@ feature %q{
     fill_in 'footer_twitter_url', with: 'http://twitter.com/me'
     fill_in 'footer_links_md', with: '[markdown link](/)'
     click_button 'Update'
-    page.should have_content 'Your content has been successfully updated!'
+    expect(page).to have_content 'Your content has been successfully updated!'
 
     visit root_path
 
     # Then social media icons are only shown if they have a value
-    page.should_not have_selector 'i.ofn-i_044-facebook'
-    page.should     have_selector 'i.ofn-i_041-twitter'
+    expect(page).not_to have_selector 'i.ofn-i_044-facebook'
+    expect(page).to     have_selector 'i.ofn-i_041-twitter'
 
     # And markdown is rendered
-    page.should have_link 'markdown link'
+    expect(page).to have_link 'markdown link'
   end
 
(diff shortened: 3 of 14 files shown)


app/controllers/spree/admin/general_settings_controller_decorator.rb
```diff
@@ -8,6 +8,7 @@ module GeneralSettingsEditPreferences
       def edit
         super
         @preferences_general << :bugherd_api_key
+        @preferences_general << :user_guide_link
```
HugsDaniel: I'm not sure about this, should I create a method below to fill `@preferences_general` and avoid these two lines ? Or something else ?

spec/features/admin/general_settings_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
```
HugsDaniel: Also not sure about where to put this file. I treated it as a feature test, is that correct ?

app/models/spree/app_configuration_decorator.rb
```diff
@@ -4,6 +4,9 @@
   # we can allow to be modified in the UI by adding appropriate form
   # elements to existing or new configuration pages.
 
+  # General Preferences
+  preference :user_guide_link, :string, default: nil
```
HugsDaniel: Should I add a default user guide link (http://www.openfoodnetwork.org/platform/user-guide/ for example). Right now if no link is provided, it just links to the current page.

HugsDaniel:
@RachL would it be ok what Luis said ? Put it in the content section

spec/features/admin/general_settings_spec.rb
mkllnk: Yes, that's fine. If you wanted to test the whole feature, you would visit the root path `/` and test if the URL of the link is correct. That would be more apt for a feature test.

HugsDaniel:
@mkllnk @RachL @luisramos0 I moved it to content config and tests in `content_spec.rb` where other content config features are tested. Thoughts ?

app/controllers/admin/contents_controller.rb
```diff
@@ -9,7 +9,8 @@ def edit
                               {name: I18n.t('admin.contents.edit.main_links'), preferences: [:menu_1, :menu_1_icon_name, :menu_2, :menu_2_icon_name, :menu_3, :menu_3_icon_name, :menu_4, :menu_4_icon_name, :menu_5, :menu_5_icon_name, :menu_6, :menu_6_icon_name, :menu_7, :menu_7_icon_name]},
                               {name: I18n.t('admin.contents.edit.footer_and_external_links'), preferences: [:footer_logo,
                                                              :footer_facebook_url, :footer_twitter_url, :footer_instagram_url, :footer_linkedin_url, :footer_googleplus_url, :footer_pinterest_url,
-                                                             :footer_email, :community_forum_url, :footer_links_md, :footer_about_url]}]
+                                                             :footer_email, :community_forum_url, :footer_links_md, :footer_about_url]},
+                              {name: I18n.t('admin.contents.edit.user_guide'), preferences: [:user_guide_link]}]
```
sauloperez: This `#edit` is getting hairy and hard to read. Honestly, I find it hard to understand what you just did. What about splitting these preference sections into objects that return each of this individual hashes and then we simply merge them into an array here? that is an approach that served me well in the past. Something like (I haven't tried it, but it should work :trollface:):

```ruby
# app/models/preference_sections/user_guide_section.rb

class UserGuideSection
  def name
    I18n.t('admin.contents.edit.user_guide')
  end

  def  preferences
    [:user_guide_link]
  end
end
```

```ruby
# app/controllers/admin/contents_controller.rb
module Admin
  class ContentsController < Spree::Admin::BaseController
    def edit
      @preference_sections = preference_sections.map do |preference_section|
        { name: preference_section.name, preferences: preference_section.preferences }
      end
    end

    private
  
    # Returns an array with the instances of preference sections, which all respond to #name and #preferences.
    def preference_sections
      Dir["models/preference_sections/**/*.rb"].map do |filename|
        basename = File.basename(filename, '.rb')
        basename.camelize.constantize.new
      end
  end
end

This makes it really easy to add, remove and change sections/preferences. And keep adding them all the time (whether they should be preferences in the UI or config settings in ofn_install it's another topic).

File: spec/features/admin/content_spec.rb [hunk 5]
```diff
@@ -18,23 +18,36 @@
...
-    ContentConfig.logo.to_s.should include "logo-white"
+    expect(ContentConfig.logo.to_s).to include "logo-white"
+  end
+
+  scenario "setting user guide link" do
+    fill_in 'user_guide_link', with: 'http://www.openfoodnetwork.org/platform/user-guide/'
+    click_button 'Update'
+
+    expect(page).to have_content 'Your content has been successfully updated!'
+    expect(ContentConfig.user_guide_link).to eq 'http://www.openfoodnetwork.org/platform/user-guide/'
```
sauloperez: being this a feature spec, this expectation shouldn't be here because it belongs to a deeper testing level (unit? something in between?). It does not exert the UI but a class. If line 50 passes, is because `ContentConfig` worked. I would simply remove it because `Spree::Preferences::FileConfiguration` is already tested by Spree.

HugsDaniel:
@sauloperez your solution works great ! It is cleaner indeed. Thoughts ?

app/controllers/admin/contents_controller.rb
```diff
@@ -26,5 +20,14 @@ def update
 
       redirect_to main_app.edit_admin_content_path
     end
+
+    private
+
+    def preference_sections
+      Dir["app/models/preference_sections/*.rb"].map do |filename|
+        basename = 'PreferenceSections::' + File.basename(filename, '.rb').camelize
+        basename.constantize.new
```
mkllnk: This is very clever. But I would prefer it more explicit. The problems I see:

- From the code I can't see what is loaded. I need to go to the directory to find out.
- I can't navigate to a section declaration using my IDE.
- The dependency on these classes cannot be detected automatically.
- One file not following the file name / class name convention breaks the code (class not found).
- We cannot define the order or the sections here. They may appear random depending on the file system.

app/models/content_configuration.rb
```diff
@@ -69,4 +69,7 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
 EOS
 
   preference :footer_about_url, :string, default: "http://www.openfoodnetwork.org/ofn-local/open-food-network-australia/"
+
+  #User Guide
+  preference :user_guide_link, :string, default: 'http://www.openfoodnetwork.org/platform/user-guide/'
```
mkllnk: I just followed that link and the current URL is https://guide.openfoodnetwork.org/. We should use that here.

app/controllers/admin/contents_controller.rb
```diff
@@ -26,5 +20,20 @@ def update
...
+        PreferenceSections::HeaderSection.new,
+        PreferenceSections::HomePageSection.new,
+        PreferenceSections::ProducerSignupPageSection.new,
+        PreferenceSections::HubSignupPageSection.new,
+        PreferenceSections::GroupSignupPageSection.new,
+        PreferenceSections::MainLinksSection.new,
+        PreferenceSections::FooterAndExternalLinksSection.new,
+        PreferenceSections::UserGuideSection.new
+      ]
+    end
```
HugsDaniel: @mkllnk something like this ?

app/controllers/admin/contents_controller.rb
```diff
@@ -26,5 +20,20 @@ def update
 
       redirect_to main_app.edit_admin_content_path
     end
+
+    private
+
+    def preference_sections
+      [
+        PreferenceSections::HeaderSection.new,
+        PreferenceSections::HomePageSection.new,
```
HugsDaniel: @mkllnk something like this ?

app/controllers/admin/contents_controller.rb
sauloperez: Just to give another opinion, let me answer these questions:

> From the code I can't see what is loaded. I need to go to the directory to find out.

Like anything in an Object-oriented system. And that's a great benefit; you can read what the code does at an abstract level to understand how an algorithm works. You can get into detail if you need to by reading the objects' implementations.

> I can't navigate to a section declaration using my IDE.

Indeed that is a con, but there's no direct reference to a particular section so checking the dir you'll see the section object you are interested in. No a big deal with so simple objects.

> The dependency on these classes cannot be detected automatically.

Probably the biggest downside. We just have the call to `Dir` which won't be automatically detected.

> One file not following the file name / class name convention breaks the code (class not found).

The file convention is being `*.rb` as shown in line 27. As for the class name, any name under that namespace will be picked. So unless you change this method or you don't declare the new class under the `PreferenceSections` namespace any will work.

> We cannot define the order or the sections here. They may appear random depending on the file system.

Easy to fix by using `sort_by` with a method each section object declares, and that could be even more explicit than a big array as we have now.

app/controllers/admin/contents_controller.rb
```diff
@@ -26,5 +20,20 @@ def update
...
+        PreferenceSections::ProducerSignupPageSection.new,
+        PreferenceSections::HubSignupPageSection.new,
+        PreferenceSections::GroupSignupPageSection.new,
+        PreferenceSections::MainLinksSection.new,
+        PreferenceSections::FooterAndExternalLinksSection.new,
+        PreferenceSections::UserGuideSection.new
```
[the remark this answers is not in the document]
HugsDaniel: Is there a way we can do both ? Respect OCP and have explicit code as Maikel suggests ?

app/controllers/admin/contents_controller.rb
mkllnk: >> We cannot define the order or the sections
> Easy to fix by using sort_by with a method each section object declares, and that could be even more explicit than a big array as we have now.

Adding something like priority levels to the sections so that they can be sorted creates a dependency between the sections. I can't change the position of one section without knowing the positions of all other sections. A list (array) solves this very well.

> The file convention is being *.rb

* The file needs to have the same name as the class it contains. (This fails for class names containing `API` or `URL` unless we name the files `a_p_i` and `u_r_l`.)
* They need to be in that directory. (You can't have them somewhere else and you can't have another rb file in that directory.)

These conventions are not a big deal, but I don't see any need for this flexibility here. It's a trade-off between having more code (but stupid code) and some auto-loading algorithm relying on conventions. In this case I voted for the dumb array, because it solves the sorting as well.

sstead:
**Testing Notes**

- Yes, the guide link is now configurable :)

https://docs.google.com/document/d/1uqQcthv0-sL4MsNrNCfKr9QdteQpJRqhsPZe_jYpjLo/edit#

RachL:
@sstead did you test #2631 as well? I just want to make sure as you did not mentioned it in testing notes and both issues ended up in the same PR.

PR #2692, opened 2018-09-12
Angularise orders page

### LINKED ISSUE #2526: [Bulk invoice printing] Angularise the current Orders page
## Description

Part 2 of the preparation for Bulk Invoice Printing, after #2525 is finished.

We need to change the admin Orders page so that the orders table uses angular instead of a plain Rails view, in preparation for adding more interactive functionality into the page.

## Acceptance Criteria

The Orders page is Angularised. The functionality should be the same as before, but working via Angular.

Matt-Yorkley:
#### What? Why?

Closes #2526

Angularises the admin orders index page in preparation for adding bulk actions.

- [x] Set up angular orderCtrl to fetch results via angular orders resource
- [x] Convert orders index table to using ng-repeat with fetched data
- [x] Change filters and pagination to use angular
- [x] Add table column sorting buttons with angular
- [x] Update specs

#### What should we test?

The orders page should look and function the same as before.

#### Release notes

Changed the admin orders page to use angular.

Changelog Category: Changed

#### Screenshots

Loading:

screenshot from 2018-09-18 15-43-58
--------------------------------------------------------------------------------

Loaded:

screenshot from 2018-09-18 15-50-11
--------------------------------------------------------------------------------

Pagination:

screenshot from 2018-09-18 15-50-27


app/controllers/spree/admin/orders_controller_decorator.rb
```diff
@@ -100,18 +100,17 @@ def update_distribution_charge
...
-      @search.result.includes([:user, :shipments, :payments]).
-        distributed_by_user(spree_current_user).
-        page(params[:page]).
-        per(params[:per_page] || Spree::Config[:orders_per_page])
-    end
+    @search = if json_request?
+                OpenFoodNetwork::Permissions.new(spree_current_user).editable_orders.ransack(params[:q])
+              else
+                Spree::Order.accessible_by(current_ability, :index).ransack(params[:q])
+              end
```
luisramos0: I think this would look better with an extractmethod. The name of the method is an opportunity to describe what is being done: user_orders?
@search = user_orders.ransack(params[:q])
...

```
def user_orders
  if json_request?
      OpenFoodNetwork::Permissions.new(spree_current_user).editable_orders
  else
    Spree::Order.accessible_by(current_ability, :index)
  end
end
```

app/serializers/api/admin/order_serializer.rb
```diff
@@ -16,7 +61,33 @@ def phone
     object.billing_address.nil? ? "a" : ( object.billing_address.phone || "" )
   end
 
+  def created_at
+    object.created_at.blank? ? "" : I18n.l(object.created_at, format: '%B %d, %Y')
+  end
+
   def completed_at
-    object.completed_at.blank? ? "" : object.completed_at.strftime("%F %T")
+    object.completed_at.blank? ? "" : I18n.l(object.completed_at, format: '%B %d, %Y')
```
luisramos0: Is this how we localize dates? nice! #I_am_learning

app/serializers/api/admin/order_serializer.rb
```diff
@@ -16,7 +61,33 @@ def phone
...
+  end
+
+  def pending_payments
```
luisramos0: from here to the bottom it's biz logic, it needs to go to a service or to the payment model. right?

app/views/spree/admin/orders/_angular_pagination.html.haml
```diff
@@ -0,0 +1,17 @@
+.pagination
```
luisramos0: I think this partial should be in /views/admin/shared. this is ready to be re-used, correct?

app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
```diff
@@ -1,26 +1,41 @@
-angular.module("admin.orders").controller "ordersCtrl", ($scope, $compile, $attrs, shops, orderCycles) ->
```
Matt-Yorkley: This file looks a bit confusing in the diff, but basically `ordersCtrl` was renamed to `orderCtrl` (above) and the contents of this file are new.

app/views/spree/admin/orders/index.html.haml
```diff
@@ -24,15 +24,47 @@
...
-      %th= sort_link @search, :payment_state,  t(:payment_state, :scope => 'activerecord.attributes.spree/order')
-      %th= sort_link @search, :shipment_state, t(:shipment_state, :scope => 'activerecord.attributes.spree/order')
-      %th= sort_link @search, :email,          t(:email, :scope => 'activerecord.attributes.spree/order')
-      %th= sort_link @search, :total,          t(:total, :scope => 'activerecord.attributes.spree/order')
+        %th
+          %a{'ng-click' => "sortOptions.toggle('created_at')"}
+            = t(:created_at, scope: 'activerecord.attributes.spree/order')
+            %span{'ng-show' => "sorting == 'created_at asc'"}= "&#x25B2;".html_safe
+            %span{'ng-show' => "sorting == 'created_at desc'"}= "&#x25BC;".html_safe
+      %th
```
mkllnk: There is a lot of code duplication here. Maybe a partial with local variables would be good?

app/assets/stylesheets/admin/orders.css.scss
```diff
@@ -91,3 +91,15 @@ th.actions {
 table.index td.actions {
   text-align: left;
 }
+
+.orders-loading {
```
mkllnk: Should the spinner look the same everywhere? There are 11 other uses of the spinner. It would be great if we could have one partial and one block of CSS following BEM to use everywhere.

Matt-Yorkley: Are we switching to BEM for all CSS now?

mkllnk: We haven't discussed this properly, but several people like to use BEM. We don't have to switch everything. It can be a tool to make naming easier and I just thought this is a good example for BEM.

Matt-Yorkley: Is it ok if I move that to a separate issue/PR? I can see 27 instances across 22 files that I would need to touch to make a single view and CSS for the spinner, and there's already quite a lot going on in this PR.

app/services/pending_payments.rb
```diff
@@ -0,0 +1,15 @@
+# Returns the capturable payment object for an order with balance due
+
+class PendingPayments
+  def initialize(order)
+    @order = order
+  end
+
+  def payment_object
+    @order.payments.select{ |p| p if p.state == 'checkout' }.first
```
sauloperez: can we use meaningful and communicative var names, please :pray: ? we don't pay by char :trollface:

app/services/pending_payments.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def can_be_captured?
+    payment_object && payment_object.actions.include?('capture')
+  end
+end
```
sauloperez: IMO this is a model an not a service. It represents a concept of our domain and not a command to be performed. I'd move it to `model/`. Btw, really nice abstraction :ok_hand:

app/serializers/api/admin/order_serializer.rb
```diff
@@ -8,6 +11,48 @@ def full_name
     object.billing_address.nil? ? "" : ( object.billing_address.full_name || "" )
   end
 
+  def distributor_name
+    object.distributor.andand.name
+  end
+
+  def show_path
+    return '' unless object.id
+    Spree::Core::Engine.routes_url_helpers.admin_order_path(object)
```
sauloperez: you could move `Spree::Core::Engine.routes_url_helpers` to a method to DRY things a bit.

app/serializers/api/admin/order_serializer.rb
```diff
@@ -8,6 +11,48 @@ def full_name
     object.billing_address.nil? ? "" : ( object.billing_address.full_name || "" )
   end
 
+  def distributor_name
+    object.distributor.andand.name
+  end
+
+  def show_path
+    return '' unless object.id
```
sauloperez: what about `object.id || ''`?

Matt-Yorkley: The object id isn't being returned by that method...

sauloperez: :poop: right. I shut up then

app/assets/stylesheets/admin/orders.css.scss
mkllnk: Sure. A separate PR sounds great.

app/services/pending_payments.rb
```diff
@@ -6,7 +6,7 @@ def initialize(order)
   end
 
   def payment_object
-    @order.payments.select{ |p| p if p.state == 'checkout' }.first
+    @order.payments.select{ |payment| payment if payment.state == 'checkout' }.first
```
mkllnk: I can find another thing to pick on. ;-) The `select` block is evaluated as boolean, so you can scrap the `payment if`:
```rb
@order.payments.select{ |payment| payment.state == 'checkout' }.first
```

But you actually don't need to iterate all payments just to select the first one in checkout. There is a better method:
```ruby
@order.payments.find{ |payment| payment.state == 'checkout' }
```

It's not useful in this case, but you could use lazy evaluation as well:
```rb
@order.payments.lazy.select{ |payment| payment.state == 'checkout' }.first
```

Lazy evaluation helps with big enumerations. You can even deal with infinity:
```ruby
(1..Float::INFINITY).lazy.select { |i| i == 5 }.first
# 5
```

app/services/pending_payments.rb
Matt-Yorkley: You really think this should be a model and not a service? It's task is just "return the payable payment object for an order".

sauloperez: To me, the key is that you named it after a noun and not a verb and you don't perform an action that changes the system's state.

app/services/pending_payments.rb
```diff
@@ -0,0 +1,15 @@
...
+  def initialize(order)
+    @order = order
+  end
+
+  def payment_object
+    @order.payments.find{ |payment| payment.state == 'checkout' }
+  end
+
+  def can_be_captured?
+    payment_object && payment_object.actions.include?('capture')
```
mkllnk: I just noticed that you call payment_object twice. It's not a big deal, but it will search through all payments twice as well. This should call it only once:
```ruby
payment_object.andand.actions.include?('capture')
```

app/services/pending_payments.rb
mkllnk: The purpose of this class is maybe not the modelling of pending payments. We could just call `order.pending_payments` for that. This service is working out if there is something to capture. There is not much logic to that though and I would probably get rid of the service all together and have it in the serialiser. It's shorter as well:

```ruby
  def capture_path
    pending_payment = object.pending_payments.first
    return '' unless object.payment_required? && pending_payment
    spree_routes_helper.fire_admin_order_payment_path(object, pending_payment.id, e: 'capture')
  end
```

app/serializers/api/admin/order_serializer.rb
```diff
@@ -8,6 +11,48 @@ def full_name
...
+    return '' unless object.shipment_state
+    spree_routes_helper.admin_order_shipments_path(object)
+  end
+
+  def ship_path
+    spree_routes_helper.fire_admin_order_path(object, e: 'ship')
+  end
+
+  def capture_path
+    payment_due = PendingPayments.new(object)
```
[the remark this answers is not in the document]
Matt-Yorkley: Thanks @mkllnk I totally missed Spree::Order.pending_payments! I'll remove the service now and update it.

app/services/pending_payments.rb
Matt-Yorkley: Done. That's much better.

Matt-Yorkley:

RachL:
I like the new pagination look @Matt-Yorkley ! The current page could be even more "bold" than the other pages, but that's just my opinion :)

Testing notes:

https://docs.google.com/document/d/1xmtWFQXCFFBX8EPPxBqvtYt-byQI_UeJ93EsOTmmDkM/edit#

PR #2701, opened 2018-09-13
2700 improve product refresh scheduling

### LINKED ISSUE #2700: ProductsCacheRefreshment fails on broken delayed jobs
## Description

Follow up from #2676.

Creating a new order cycle or modifying an existing order cycle triggers the `RefreshProductsCacheJob`. Before that job is created, it scans through all other jobs to see if there is already a job with the same task. It crashes if one of the other jobs fails to load.

## Expected Behavior

Creating a new order cycle always works.

## Actual Behavior

Creating a new order cycle fails if there are broken jobs in the database.

```
Started POST "/admin/order_cycles.json" for 182.239.162.19 at 2018-09-12 09:03:25 +0000
Processing by Admin::OrderCyclesController#create as JSON
  Parameters: {"order_cycle"=>{"incoming_exchanges"=>[{"enterprise_id"=>54, "incoming"=>true, "active"=>true, "variants"=>{"188"=>true}, "enterprise_fee_ids"=>nil}], "outgoing_exchanges"=>[{"enterprise_id"=>54, "incoming"=>false, "active"=>true, "variants"=>{"188"=>true, "undefined"=>false}, "pickup_time"=>"testing", "enterprise_fee_ids"=>nil}], "name"=>"Maikel's Test", "orders_open_at"=>"2018-09-12 00:00:00", "orders_close_at"=>"2018-09-13 00:00:00", "coordinator_id"=>54, "schedule_ids"=>nil, "coordinator_fee_ids"=>nil}}
Completed 500 Internal Server Error in 96.0ms (ActiveRecord: 25.5ms)

Delayed::DeserializationError (ActiveRecord::RecordNotFound, class: Spree::User, primary key: 66 (Couldn't find Spree::User with id=66)):
  lib/open_food_network/products_cache_refreshment.rb:35:in `pending_job?'
  lib/open_food_network/products_cache_refreshment.rb:20:in `refresh'
  lib/open_food_network/products_cache.rb:169:in `refresh_cache'
  lib/open_food_network/products_cache.rb:128:in `refresh_outgoing_exchange'
  lib/open_food_network/products_cache.rb:76:in `exchange_changed'
  app/models/exchange.rb:91:in `refresh_products_cache'
  lib/open_food_network/order_cycle_form_applicator.rb:73:in `add_exchange'
  lib/open_food_network/order_cycle_form_applicator.rb:47:in `block in go!'
  lib/open_food_network/order_cycle_form_applicator.rb:35:in `each'
  lib/open_food_network/order_cycle_form_applicator.rb:35:in `go!'
  app/services/order_cycle_form.rb:33:in `apply_exchange_changes'
  app/services/order_cycle_form.rb:19:in `block in save'
  app/services/order_cycle_form.rb:17:in `save'
  app/controllers/admin/order_cycles_controller.rb:41:in `create'
```

## Steps to Reproduce

1.  Sign up with an invalid email address that will fail to be delivered.
2.  Log in as admin and delete the unconfirmed user account.
3.  Try to create an order cycle.

## Context

This first happened on the US server. The email setup was broken for a while and a lot of those failed email jobs were sitting in the database.

## Severity

bug-s2: A critical feature can be broken in the wrong circumstances. There is no workaround for a user. A dev has to fix the database before order cycles can be created again.

## Your Environment

* Version used: 44faa11 after v1.19.0.

## Possible Fix

The class `OpenFoodNetwork::ProductsCacheRefreshment` checks if there is a job already before creating a new one. We could rescue exceptions, but it would be much better to make the check more efficient and more robust.

The handler of a delayed job is encoded like this:
```yml
--- !ruby/struct:RefreshProductsCacheJob\ndistributor_id: 2\norder_cycle_id: 1\n
```

We can encode the handler of a new job and query the database for a job with exactly the same handler:

```ruby
    def self.pending_job?(distributor, order_cycle)
      job = RefreshProductsCacheJob.new(distributor.id, order_cycle.id)
      Delayed::Job.
        where(locked_at: nil).
        where(handler: job.to_yaml).
        exists?
    end
```

This solution was already recommended by Rohan when he wrote the class.

mkllnk:
#### What? Why?

Closes #2700.

When broken delayed jobs are in the database, product cache refreshment fails as described in the issue.

#### What should we test?

1. Sign up with an invalid email address that will fail to be delivered: `a@b.void`
2. Log in as admin and delete the unconfirmed user account.
3. Try to create an order cycle with products.

#### Release notes

Fixed order cycle creation which failed when invalid user accounts were deleted.

Changelog Category: Fixed


lib/open_food_network/products_cache_refreshment.rb
```diff
@@ -21,21 +21,23 @@ def self.refresh(distributor, order_cycle)
...
-    def self.pending_job?(job)
-      Delayed::Job.
-        where(locked_at: nil).
-        where(handler: job.to_yaml).
-        exists?
-    end
-
-    def self.enqueue_job(job)
-      Delayed::Job.enqueue job, priority: 10
+    class << self
```
luisramos0: what happened here?
You sent me reading about eigenclasses... my head hurts :-)
In this particular case, is this a convention (use class << self instead of self. ) or is there any change in logic with this?

sauloperez: I'd like to check what the Ruby Style guide says about it and I'd stick to the community's default. Also, `private` doesn't work for class methods AFAIK, so I'm not sure here if this methods get hid.

luisramos0: hmm.. I read a bit more about this and actually, this is exactly how you create private class methods :-)

lib/open_food_network/products_cache_refreshment.rb
```diff
@@ -17,31 +17,27 @@
...
+      end
 
-      Delayed::Job.
-        where(locked_at: nil).
-        map(&:payload_object).
-        select { |j|
-          j.class == RefreshProductsCacheJob &&
-          j.distributor_id == distributor.id &&
-          j.order_cycle_id == order_cycle.id
-        }.any?
```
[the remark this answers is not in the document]
mkllnk: Yes, the logic is the same. We check for an identical job.

*BUT:* The previous implementation needed to call `payload_object` on each job. That call fails if the object can't be deserialised properly. That happens when the payload object is a user and that user doesn't exist any more. In that case `map(&:payload_object)` raised `Delayed::DeserializationError` caused by `ActiveRecord::RecordNotFound`. Instead of rescueing from that error in a separate each block, I simplified the implementation to just check for the serialised content in the db column `handler`. That avoids the error and is more efficient.

lib/open_food_network/products_cache_refreshment.rb
mkllnk: I know! I was a bit reluctant to use `class << self` here, but I followed rubocop and found that it is the widely accepted way to declare private class methods. I should have tackled one rubocop violation at a time to explain it better. Sorry.

The alternative is to declare it separately:
```ruby
private_class_method :refresh_job
private_class_method :pending_job?
private_class_method :enqueue_job
```

That is more repetitive, but you definitely know what it's doing. Should I change it?

luisramos0: :+1:
private_class_method is much easier to understand :-)
If it's also correct I'd use it.

luisramos0: yes, imo, much better.

PR #2728, opened 2018-09-14
[Spree Upgrade] Remove use of Order#shipping_method_id from ShippingMethodsControllerDecorator

### LINKED ISSUE #2010: [Spree Upgrade] Remove use of Order#shipping_method_id from ShippingMethodsControllerDecorator
`app/controllers/spree/admin/shipping_methods_controller_decorator.rb` references the column `shipping_method_id` from the `spree_orders` table and this won't be longer available in Spree 2.0.

The work has already been started in: https://github.com/openfoodfoundation/openfoodnetwork/pull/1659

Initially we thought of making shipping_methods soft deletable, but currently, the easier approach seems to just keep the existing method and disallow deleting shipping_methods that are being used in orders.

luisramos0:
#### What? Why?

Closes #2010
We do not proceed with the (a little long) work started in #1659, we just keep functionality as is: we just make do_not_destroy_referenced_shipping_methods work with spree 2 model.

Added spec to cover behaviour. Also tested this manually /admin/shipping_methods and it works.

### DIFF
--- a/app/controllers/spree/admin/shipping_methods_controller_decorator.rb
+++ b/app/controllers/spree/admin/shipping_methods_controller_decorator.rb
@@ -5,22 +5,10 @@ module Spree
       before_filter :load_hubs, only: [:new, :edit, :create, :update]
 
       # Sort shipping methods by distributor name
-      # ! Code copied from Spree::Admin::ResourceController with two added lines
       def collection
-        return parent.send(controller_name) if parent_data.present?
+        collection = super
+        collection = collection.managed_by(spree_current_user).by_name
 
-        collection = if model_class.respond_to?(:accessible_by) &&
-                         !current_ability.has_block?(params[:action], model_class)
-
-                       model_class.accessible_by(current_ability, action)
-
-                     else
-                       model_class.scoped
-                     end
-
-        collection = collection.managed_by(spree_current_user).by_name # This line added
-
-        # This block added
         if params.key? :enterprise_id
           distributor = Enterprise.find params[:enterprise_id]
           collection = collection.for_distributor(distributor)
--- a/app/controllers/spree/admin/shipping_methods_controller_decorator.rb
+++ b/app/controllers/spree/admin/shipping_methods_controller_decorator.rb
@@ -17,14 +17,10 @@ module Spree
         collection
       end
 
-      # This method was originally written because ProductDistributions referenced shipping
-      # methods, and deleting a referenced shipping method would break all the reports that
-      # queried it.
-      # This has changed, and now all we're protecting is Orders, which is a spree resource.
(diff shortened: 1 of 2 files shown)


sauloperez:
> I have not managed to make the auto test (included here) work, for some reason spree_get :index works fine but spree_delete :destroy is not reaching the controller... I don't understand.

In `log/test.log` I spotted the request was redirecting due to an unauthorized request. Done by /app/controllers/spree/admin/base_controller_decorator.rb:43:in `unauthorized' 

By stubbing `:current_spree_user` I managed to get the requests go through, so there's something not working with `quick_login_as_admin` but I don't know what.

luisramos0:
great, thanks a lot @sauloperez 
looking at the logs does help debugging sometimes  😢 
I'll try to find what's wrong with quick_login_as_admin...

luisramos0:
When the shipping_method is not deleted and the flash[error] statement is executed, the error is not displayed to the user in the page, I am leaving this issue to be solved as part of #2744

app/controllers/spree/admin/shipping_methods_controller_decorator.rb
```diff
@@ -29,14 +17,10 @@ def collection
...
-      # methods, and deleting a referenced shipping method would break all the reports that
-      # queried it.
-      # This has changed, and now all we're protecting is Orders, which is a spree resource.
-      # Do we really need to protect it ourselves? Does spree do this, or provide some means
-      # of preserving the shipping method information for past orders?
+      # Spree allows soft deletes of shipping_methods but our reports are not adapted to that.
+      # So, this method prevents the deletion (even soft) of shipping_methods that are referenced in orders.
       def do_not_destroy_referenced_shipping_methods
-        order = Order.where(:shipping_method_id => @object).first
+        order = Order.joins(shipments: :shipping_rates).where( spree_shipping_rates: { :shipping_method_id => @object } ).first
```
HugsDaniel: I think `Order.joins(shipments: :shipping_methods).first` does the job as well

HugsDaniel:
@luisramos0 I've tried to reproduce the issue with the flash[:error] you mentioned. In my browser I indeed don't see the flash[:error] but the shipping_method is not deleted. In feature tests for this case it's all green, RSpec finds the flash[:error]. I'm not sure why though :/

luisramos0:
Interesting @HugsDaniel 
Can you try save_and_open_screenshot and see if rspec is really seeing the error?
Anyway, I can have a look when your PRs are merged.

HugsDaniel:
@luisramos0 I tried save_and_open_page and rspec is really seeing the error

luisramos0:
:-O

PR #2729, opened 2018-09-14
[Spree Upgrade] Replace order.shipping_method_id in current_order_serializer

### LINKED ISSUE #2679: [Spree Upgrade] Replace order.shipping_method_id in current_order_serializer

luisramos0:
#### What? Why?

Mostly copied from https://github.com/openfoodfoundation/openfoodnetwork/pull/1651/commits/06345e91615b93b75d9e1b1411b7da33fe1dd607

Closes #2679 

#### What should we test?
current order serialiser spec should be green.
Cant be tested on the frontend yet.


spec/serializers/current_order_serializer_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'spec_helper'
+
+describe Api::CurrentOrderSerializer do
+  let(:distributor) { create(:distributor_enterprise) }
+  let(:oc) { create(:simple_order_cycle) }
+  let(:li) { create(:line_item, variant: create(:variant)) }
```
sauloperez: could we take the chance to improve the variables names to make them a bit more communicative? I'd rather name them `order_cycle` and `line_item` respectively.

app/serializers/api/current_order_serializer.rb
```diff
@@ -12,6 +12,10 @@ def payment_method_id
     object.payments.first.andand.payment_method_id
   end
 
+  def shipping_method_id
+    object.shipping_method.id
```
sauloperez: I'm might have seen it and I don't remember but, did we add a `#shipping_method` that uses `shipments` under the hood?

luisramos0: object.shipping_method takes you to the orderShippingMethod concern that uses order.shipments.

spec/serializers/current_order_serializer_spec.rb
luisramos0: :+1: done.
I have also added andand to the shipping_method_id method to make it work even if there are no shipments in the order.

PR #2734, opened 2018-09-17
Get useful feedback from Rubocop Metrics cops

sauloperez:
#### What? Why?

One of the biggest pros of linters like Rubocop is to get valuable feedback to
help write better code. The way we have Rubocop configured now we don't prevent
new code from adhere improved code quality and this is specially important when
touching code that already suffers from complexity.

It's the second time in few days that I miss Rubocop's warnings on code that I need to refactor.

Without all Rubocop's Metrics cops enabled there's no way to get this insights
and write better code. This enables them while regenerating the
`.rubocop_todo.yml` to hide the current violations.

So, next time we touch existing code that we think could be simpler, we should
go to `.rubocop_todo.yml` and remove any occurrences of the file in question.
This way we could Rubocop's feedback right in the editor. This is tremendously
helpful when refactoring. It shows you where to start.

#### What should we test?
No production code affected.

#### Release notes
Enable Rubocop's Metric cop to improve code quality

Changelog Category: Added


.rubocop.yml
```diff
@@ -172,30 +172,3 @@ Lint/AmbiguousRegexpLiteral:
...
-  Enabled: false
-
-Metrics/MethodLength:
-  Enabled: false
-
-Metrics/ParameterLists:
-  Enabled: false
-
-Metrics/PerceivedComplexity:
-  Enabled: false
```
sauloperez: these are all true by default

sauloperez:
I had to disable Code Climate's maintainability checks because they duplicate Rubocop's ones. See https://mastodon.social/@Gargron/99045485821820038, we're not the only ones to dislike this feature.

.rubocop.yml
```diff
@@ -174,28 +170,28 @@ Lint/AssignmentInCondition:
...
+  Max: 10
 
 Metrics/ParameterLists:
-  Enabled: false
+  Max: 5
 
 Metrics/PerceivedComplexity:
-  Enabled: false
+  Max: 7
```
sauloperez: These are the default values. They need to be specified so that we override the ones inherited from the rubocop_todo.yml, otherwise, the largest Max value found in the codebase will be used. More details: https://github.com/rubocop-hq/rubocop/issues/3264#issuecomment-259959753

luisramos0:
I dont think I understand what is the change.

"So, next time we touch existing code that we think could be simpler, we should
go to .rubocop_todo.yml and remove any occurrences of the file in question."

I didnt know this, we can put this in the code review checklist, all files touched by a commit have to be removed from this list.

I am all for being more demanding in terms of following the ruborules as long as we as reasonable and allow exceptions (like not having to do a 1 day refactoring detour to be able to change one line) ;-)

sauloperez:
Sorry @luisramos0 . There are lots of assumptions here. So step by step.

## How Rubocop works

So Rubocop allows genereting a `.rubocop_todo.yml` to set a threshold for the project. This is pretty well explained in its docs.

> If you have a code base with an overwhelming amount of offenses, it can be a good idea to use rubocop --auto-gen-config and add an inherit_from: .rubocop_todo.yml in your .rubocop.yml. The generated file .rubocop_todo.yml contains configuration to disable cops that currently detect an offense in the code by excluding the offending files, or disabling the cop altogether once a file count limit has been reached.

> By adding the option --exclude-limit COUNT, e.g., rubocop
--auto-gen-config --exclude-limit 5, you can change how many files are excluded before the cop is entirely disabled. The default COUNT is 15.

> Then you can start removing the entries in the generated .rubocop_todo.yml file one by one as you work through all the offenses in the code. 

## What I did

So to me, it's on the dev creating the PR to go check any offenses in the `.rubocop_todo.yml` for the files he touched. He might have improved them and the exclusion it's no longer needed.

As per my investigation, that is not the case for the `Metric`-type cops (lint rules in Rubocop's jargon). The `--auto-gen-config` argument does not list the exclusions but gets the largest value for each metric and sets it as `Max`. As a result, since `.rubocop_todo.yml`'s config is inherited in `.rubocop.yml` all complex bits of code in our codebase get hid. By adding a `Max` value for each of them in `.rubocop.yml` we then override these and so the offenses get displayed.

This helped me realize what it was that made `models/order_updater_decorator.rb` hard to follow for me, for instance.

## What I suggest

This PR makes lots of offenses appear in Code Climate but I think that's fine. This better reflects the status of our codebase and not what we have now a A score and no major code smells. This way it does not provide any benefit.

As for offenses displayed by Rubocop on our machines, as long as we agree on just refactoring what we think improves part of a code we're about to change, it shouldn't be a problem.

Matt-Yorkley:
So the changes should be fairly limited because the deactivated CodeClimate checks are already duplicated by Rubocop, right? It might even get a bit looser in some cases, for instance the new `PerceivedComplexity` will be 7, but CodeClimate currently is limiting it to 5. I'm not sure what the current `ClassSize` and `ModuleSize` are via CodeClimate, but I imagine 100 lines will make a lot of new violations!

If I understand correctly, the rubocop_todo hasn't been substantially cut down, so various files that have current violations will still be excluded from new checks until they are manually removed from the todo list, so the "1 day refactor to change a line of code" scenario won't actually be an issue?

sauloperez:
AFAIK Code Climate reads our config from the `.rubocop.yml` so it should take our settings. You're right, I deactivated Code Climate's own checks that are equivalent to Rubocop cops.

> If I understand correctly, the rubocop_todo hasn't been substantially cut down, so various files that have current violations will still be excluded from certain specific checks until they are manually removed from the todo list, so the "1 day refactor to change a line of code" scenario won't actually be an issue?

That's it. You'll only see Metrics-related offenses when you execute Rubocop.

sauloperez:
I see the same @kristinalim . I believe this won't happen after we merge this as Code Climate won't try to show all of them at once as it does know. It does so because the offenses are being introduced with this PR.

Issue #2744, opened 2018-09-18
[Spree Upgrade] Fix ShippingMethods Create page


luisramos0:
Fix page /admin/shipping_methods/new

Currently broken with an error related to zone id.
It's a heavily defaced page, probably better to use the opportunity to move Spree code to OFN and create a view without deface.
Also, if the approach selected is removing deface, it's probably worth doing it in master and then adapting it to spree 2 in 2-0-stable.
Also, maybe we remove deface not just for new but also for index and edit pages.

This issue is related to #2010, same controller

HugsDaniel:
May I take this one ?

luisramos0:
Hi Hugo, yes!
You can see what Matt did here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2675

You need to pick this
../spree/backend/app/views/spree/admin/shipping_methods
mix with this
./openfoodnetwork/app/overrides/spree/admin/shipping_methods
convert to haml
and put it here
./openfoodnetwork/app/views/admin

My only question is that I'd not put it in app/views/spree/admin as Matt did, I'd put it in app/views/admin

luisramos0:
You can see better as you go but it's probably better to have one PR per page: new, index and edit.
Also, as I say above, I think it's better to do this in master first.

HugsDaniel:
Also with spree step-6a used in master there's no method `zones` for Spree::ShippingMethods, it will end up crashing

luisramos0:
I'd check if step-6a is very different from 2-0-4 on this view.
If it is, I'd skip building it in master, but otherwise, I'd use step-6a to do this in master.
And then in 2-0-stable we can adapt to the changes like zone.
If you do it in 2-0-stable from 2-0-4-stable you will have to merge all changes that are done in master from now to the final migration...
What's easier?

luisramos0:
>>> I'm not sure rails will use the view if we put it in app/views/admin
yeah, that's true! we would have to change the route and controller. For this one, maybe easier to keep in spree/. Thanks, I was not seeing this problem.

HugsDaniel:
@luisramos0 there's a whole bunch of I18n syntax changes and a couple of sections added, it would seem easier to use directly 2-0-4-stable views

luisramos0:
all right, lets do that!

luisramos0:
Just leaving a note here to address question from #2728

HugsDaniel:
[...]

PR #2752, opened 2018-09-19
[Spree Upgrade] Replace Spree::ShippingMethods new view by our own and apply overrides

HugsDaniel:
#### What? Why?

Part of #2744 

Replacing Spree view for Spree:ShippingMethods new by OFN own view and applying Deface overrides.

#### What should we test?
The new shipping method view won't be working right now (because #2744), but all elemnts should be here and Rails should take our view instead of Spree's.

#### Release notes

No release notes

Changelog Category: Changed

### DIFF
--- a/app/views/spree/admin/shipping_methods/new.html.haml
+++ b/app/views/spree/admin/shipping_methods/new.html.haml
@@ -2,15 +2,15 @@
   = Spree.t(:new_shipping_method)
 - content_for :page_actions do
   %li
-    = button_link_to Spree.t(:back_to_shipping_methods_list), spree.admin_shipping_methods_path, :icon => 'icon-arrow-left'
+    = button_link_to Spree.t(:back_to_shipping_methods_list), spree.admin_shipping_methods_path, icon: 'icon-arrow-left'
 %div
-  = render :partial => 'spree/shared/error_messages', :locals => { :target => @shipping_method }
+  = render partial: 'spree/shared/error_messages', locals: { target: @shipping_method }
 %div
   = form_for [:admin, @shipping_method] do |f|
     %fieldset.no-border-top
-      = render :partial => 'form', :locals => { :f => f }
+      = render partial: 'form', locals: { f: f }
       .one.column &nbsp;
-      = render :partial => 'spree/admin/shared/hubs_sidebar', :locals => { f: f, klass: :shipping_method }
+      = render partial: 'spree/admin/shared/hubs_sidebar', locals: { f: f, klass: :shipping_method }
       .clear
       %div
-        = render :partial => 'spree/admin/shared/new_resource_links'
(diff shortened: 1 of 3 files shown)


app/views/spree/admin/shipping_methods/new.html.haml
```diff
@@ -0,0 +1,16 @@
+- content_for :page_title do
+  = Spree.t(:new_shipping_method)
+- content_for :page_actions do
+  %li
+    = button_link_to Spree.t(:back_to_shipping_methods_list), spree.admin_shipping_methods_path, :icon => 'icon-arrow-left'
+%div{"data-hook" => "admin_shipping_method_new_form_header"}
+  = render :partial => 'spree/shared/error_messages', :locals => { :target => @shipping_method }
+%div{"data-hook" => "admin_shipping_method_new_form"}
```
luisramos0: I think data-hook is only used to make things deface-able. Now that we are moving away from deface, I think we can remove these.
I am not sure about this, it needs to be confirmed by another dev...

Matt-Yorkley: Yep, looks like it's just for deface: https://github.com/spree/spree/issues/3187

Thanks for the pointer, I've just realised I need to remove those from the Spree I just recreated for the admin orders page. :)

HugsDaniel:
@luisramos0 the `zone_id` issue is coming from the form, I guess that'll be where we'll fix it

Matt-Yorkley:
**Spree view recreation tip of the day:** 
Most of our overrides live in the correct corresponding directories such as `overrides/spree/admin/shipping_methods/*`, but we have a bunch of overrides that are just left in `overrides/*`, and they may or may not apply to the view you're working on... :see_no_evil: 

screenshot from 2018-09-20 11-40-21

luisramos0:
hey, do we have a wiki page or discourse page with this topic?
I think it's worth having one with links to this pr and the order PR Matt just did for orders.
And add details as the above.

sauloperez:
Good idea @luisramos0 . Let's make it the easy for anyone to do one of this Defacifications.

sauloperez:
There was no discussion that I'm aware of but in there we're getting Spree's 2.0.4 version of the views, while in master we have 1.3.

We need to weight doing this with the added the cost of diverging much from master vs. doing this with 1.3 views in master and then update them with the Spree's 2.0.4 changes.

In any case, I don't think we'll encounter many merge conflicts due to the views. These files don't exist in `master` and I haven't seen any PR touching Deface overrides lately.

luisramos0:
I think we should consider case by case, we should prefer doing it in master.
in this case, there are new things from v2, zones and categories, I agreed with Hugo that it would be too much work to make it work in v1 and then having to change again for v2. I think in this case we will be better off managing any conflicts that may arise.

PR #2753, opened 2018-09-19
[Spree Upgrade] Replace Spree::ShippingMethods edit view by our own and apply overrides

HugsDaniel:
#### What? Why?

Part of #2744

Replacing Spree view for Spree:ShippingMethods edit by OFN own view and applying Deface overrides.

#### What should we test?

The edit shipping method view won't be working right now (because #2744), but all elements should be here and Rails should take our view instead of Spree's.

#### Release notes

No release notes


sigmundpetersen:
Remember to assign yourself to PRs @HugsDaniel and connect PR to issue. You can connect multiple PRs to one issue 👍

PR #2757, opened 2018-09-19
Deleted sample specs and one pending spec

luisramos0:
#### What? Why?

Deleting some dummy specs.
I am deleting the only file in spec/views which is a file with pending tests of a rabl view. I am sure we will be ok without this.

#### What should we test?
No tests, just deleting unused tests.


luisramos0:
I've only removed one pending view spec producers.json.rabl_spec.rb
I also felt like deleting it because it is a rabl view. we use AMS for this, correct? something to be addressed as we review our current approach to build API endpoints here: https://community.openfoodnetwork.org/t/road-map-for-ofns-api/931

luisramos0:
Hey guys, sorry for adding stuff to a PR ready to go (moved back to code review) but I randomly found another file that becomes dead with this PR, I wanted to add it to this dead code party if you don't mind ;-)

PR #2761, opened 2018-09-21
[Spree Upgrade] Fix feature specs for shipping_methods

HugsDaniel:
#### What? Why?

Part of #2744 

This fixes the feature specs for shipping_methods.
Following #2752 #2753 #2754 #2755 

Also it takes into account changes made in #2667 #2680 and #2728 which have not been merged yet but will affect these specs.

#### What should we test?

The specs are green with the changes from all PRs mentioned above.

#### Release notes

No release notes.

### DIFF
--- a/spec/features/admin/shipping_methods_spec.rb
+++ b/spec/features/admin/shipping_methods_spec.rb
@@ -20,23 +20,25 @@ feature 'shipping methods' do
 
       # Shows appropriate fields when logged in as admin
       visit spree.new_admin_shipping_method_path
-      page.should have_field 'shipping_method_name'
-      page.should have_field 'shipping_method_description'
-      page.should have_select 'shipping_method_display_on'
-      page.should have_field 'shipping_method_require_ship_address_true', checked: true
+      expect(page).to have_field 'shipping_method_name'
+      expect(page).to have_field 'shipping_method_description'
+      expect(page).to have_select 'shipping_method_display_on'
+      expect(page).to have_css 'div#shipping_method_zones_field'
+      expect(page).to have_field 'shipping_method_require_ship_address_true', checked: true
 
       # When I create a shipping method and set the distributors
       fill_in 'shipping_method_name', with: 'Carrier Pidgeon'
       check "shipping_method_distributor_ids_#{d1.id}"
       check "shipping_method_distributor_ids_#{d2.id}"
+      check "shipping_method_shipping_categories_"
       click_button 'Create'
 
       # Then the shipping method should have its distributor set
-      flash_message.should == 'Shipping method "Carrier Pidgeon" has been successfully created!'
+      expect(flash_message).to eq('Shipping method "Carrier Pidgeon" has been successfully created!')
 
       sm = Spree::ShippingMethod.last
-      sm.name.should == 'Carrier Pidgeon'
-      sm.distributors.should match_array [d1, d2]
+      expect(sm.name).to eq('Carrier Pidgeon')
+      expect(sm.distributors).to match_array [d1, d2]
     end
 
     it "at checkout, user can only see shipping methods for their current distributor (checkout spec)"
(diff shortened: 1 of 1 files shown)


spec/features/admin/shipping_methods_spec.rb
```diff
@@ -23,12 +23,14 @@
       page.should have_field 'shipping_method_name'
       page.should have_field 'shipping_method_description'
       page.should have_select 'shipping_method_display_on'
-      page.should have_field 'shipping_method_require_ship_address_true', checked: true
+      page.should have_css 'div#shipping_method_zones_field'
 
+      # binding.pry
```
luisramos0: to be removed?

spec/features/admin/shipping_methods_spec.rb
```diff
@@ -23,12 +23,14 @@
       page.should have_field 'shipping_method_name'
       page.should have_field 'shipping_method_description'
       page.should have_select 'shipping_method_display_on'
-      page.should have_field 'shipping_method_require_ship_address_true', checked: true
```
luisramos0: you should still verify the require_ship_address is there, right?

spec/features/admin/shipping_methods_spec.rb
```diff
@@ -23,12 +23,14 @@
...
+      page.should have_css 'div#shipping_method_zones_field'
 
+      # binding.pry
       # When I create a shipping method and set the distributors
       fill_in 'shipping_method_name', with: 'Carrier Pidgeon'
       check "shipping_method_distributor_ids_#{d1.id}"
       check "shipping_method_distributor_ids_#{d2.id}"
+      check "shipping_method_shipping_categories_"
```
luisramos0: why the underscore? I am just asking, I dont know what it should be.

spec/features/admin/shipping_methods_spec.rb
luisramos0: "the choice between Delivery and Pick Up, which are shipping categories now"
that's a very interesting assumption! because it would be good to get rid of this field. 
but I dont think it's correct. afaik in ofn v2 we will keep require_ship_address.

sauloperez: I believe we will keep it, yes

spec/features/admin/shipping_methods_spec.rb
HugsDaniel: That's how the selector appears in browser, it seems strange to me as well but rspec finds it and checks it, so all green here :/

luisramos0: I understand that. But why is it like that in the page? can you check the page source code and see what is creating that selector without an id?

HugsDaniel: It's coming form `check_box_tag`, I'm not sure why though but I keep investigating

HugsDaniel: It is sent to the server all right, I think he empty [] was just put there because Spree doesn't use the name of this particular field so they don't care. We can either put something like `shipping_method[shipping_categories][#{category.id)]` to have the correct id in the name (I tested it and it works fine), or we can just leave it like that. Since it's our view now I believe it's no harm to change it as we want. What do you guys think ?

luisramos0: great, thanks for investigating! I think you can decide. I'd put the category id in there.

luisramos0:
lets move this to code review @HugsDaniel ?

luisramos0:
I have rebased this locally and gave the shipping methods screens a try, it's looking good! Good job @HugsDaniel !

there's only one layout problem, the hubs list is not in the correct place:

screen shot 2018-10-18 at 15 43 01

should be:
screen shot 2018-10-18 at 15 42 50

@HugsDaniel do you want to fix this as part of this PR or do you want to create a new issue and PR for it? as you wish.

Thanks!

HugsDaniel:
You know what, I'll open a new PR instead, this is not really part of the scope of this PR

luisramos0:
ok, I have merged this one.
I'd open an issue for the problem I reported only if you are not fixing it now. If you are opening the PR in the next day or so, I'd skip the issue creation. At least, that's what I am doing for the little fixes in the spree upgrade.

PR #2775, opened 2018-09-26
[Spree Upgrade] Handle shipping_method_id in checkout

### LINKED ISSUE #2683: [Spree Upgrade] Handle shipping_method_id in checkout
We need to look at both
- app/assets/javascripts/darkswarm/controllers/checkout/checkout_controller.js.coffee
- app/assets/javascripts/darkswarm/services/checkout.js.coffee
and 
- app/views/checkout/_shipping.html.haml

and review/replace shipping_method_id usage.

For reference, other files that may need review/changes:
- app/controllers/checkout_controller.rb
- spec/requests/checkout/failed_checkout_spec.rb
- spec/requests/checkout/stripe_connect_spec.rb
- spec/javascripts/unit/darkswarm/services/checkout_spec.js.coffee
- spec/controllers/checkout_controller_spec.rb

See [this commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/1651/commits/05aeb03839ca02942d0e68894316f94f113a0e16) for line numbers.

luisramos0:
#### What? Why?

Closes #2683 and #2999 

Handles shipping_method_id in the checkout controller.

#### What should we test?
The checkout process should be tested and the shipping method selection verified.

Regarding #2999, these specs must be green in the build:
- spec/features/consumer/shopping/checkout_spec.rb:329
- spec/features/consumer/shopping/checkout_spec.rb:429

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -22,45 +22,41 @@ class CheckoutController < Spree::CheckoutController
   end
 
   def update
-    if @order.update_attributes(object_params)
-      check_order_for_phantom_fees
-      fire_event('spree.checkout.update')
-      while @order.state != "complete"
-        if @order.state == "payment"
-          return if redirect_to_paypal_express_form_if_needed
-        end
+    return update_failed unless @order.update_attributes(object_params)
 
-        next if advance_order_state(@order)
-
-        if @order.errors.present?
-          flash[:error] = @order.errors.full_messages.to_sentence
-        else
-          flash[:error] = t(:payment_processing_failed)
-        end
-        update_failed
-        return
+    check_order_for_phantom_fees
+    fire_event('spree.checkout.update')
+    while @order.state != "complete"
+      if @order.state == "payment"
+        return if redirect_to_paypal_express_form_if_needed
       end
-      if @order.state == "complete" ||  @order.completed?
-        set_default_bill_address
-        set_default_ship_address
 
-        ResetOrderService.new(self, current_order).call
-        session[:access_token] = current_order.token
+      next if advance_order_state(@order)
 
-        flash[:notice] = t(:order_processed_successfully)
-        respond_to do |format|
-          format.html do
-            respond_with(@order, :location => order_path(@order))
-          end
-          format.json do
-            render json: {path: order_path(@order)}, status: 200
-          end
-        end
+      if @order.errors.present?
+        flash[:error] = @order.errors.full_messages.to_sentence
       else
-        update_failed
+        flash[:error] = t(:payment_processing_failed)
       end
-    else
       update_failed
+      return
+    end
+    return update_failed unless @order.state == "complete" ||  @order.completed?
+
+    set_default_bill_address
+    set_default_ship_address
+
+    ResetOrderService.new(self, current_order).call
+    session[:access_token] = current_order.token
+
+    flash[:notice] = t(:order_processed_successfully)
+    respond_to do |format|
+      format.html do
+        respond_with(@order, :location => order_path(@order))
+      end
+      format.json do
+        render json: {path: order_path(@order)}, status: 200
+      end
     end
   end
 
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -22,8 +22,7 @@ class CheckoutController < Spree::CheckoutController
   end
 
   def update
-    shipping_method_id = object_params[:shipping_method_id]
-    object_params.delete(:shipping_method_id)
+    shipping_method_id = object_params.delete(:shipping_method_id)
 
     return update_failed unless @order.update_attributes(object_params)
 
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -22,13 +22,21 @@ class CheckoutController < Spree::CheckoutController
   end
 
   def update
+    shipping_method_id = object_params[:shipping_method_id]
+    object_params.delete(:shipping_method_id)
+
     return update_failed unless @order.update_attributes(object_params)
 
     check_order_for_phantom_fees
     fire_event('spree.checkout.update')
+
     while @order.state != "complete"
       if @order.state == "payment"
         return if redirect_to_paypal_express_form_if_needed
+      end
+
+      if @order.state == "delivery"
+        @order.select_shipping_method(shipping_method_id)
       end
 
       next if advance_order_state(@order)
--- a/app/models/concerns/order_shipping_method.rb
+++ b/app/models/concerns/order_shipping_method.rb
@@ -23,12 +23,13 @@ module OrderShippingMethod
   #
   # @return [Shipment]
   def select_shipping_method(shipping_method_id)
-    return if shipping_method_id.nil? || shipments.empty?
+    return if shipping_method_id.blank? || shipments.empty?
     shipment = shipments.first
 
     shipping_rate = shipment.shipping_rates.find_by_shipping_method_id(shipping_method_id)
     return unless shipping_rate
 
     shipment.selected_shipping_rate_id=(shipping_rate.id)
+    shipment.shipping_method
   end
 end
--- a/app/models/concerns/order_shipping_method.rb
+++ b/app/models/concerns/order_shipping_method.rb
@@ -19,9 +19,12 @@ module OrderShippingMethod
     shipments.first.shipping_method
   end
 
-  # Finds the shipment shipping_rate for the given shipping_method_id and selects that shipping_rate
+  # Finds the shipment's shipping_rate for the given shipping_method_id and selects that shipping_rate.
+  # If the selection is successful, it persists it in the database by saving the shipment.
+  # If it fails, it does not clear the current shipping_method selection.
   #
-  # @return [Shipment]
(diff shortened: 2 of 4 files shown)


app/controllers/checkout_controller.rb
```diff
@@ -22,45 +22,41 @@ def edit
   end
 
   def update
```
luisramos0: The diff is not helping.
I replaced
```
if X
    long logic
    ...
else
    update_failed
```
with
```
return update_failed unless X
long logic
...
```

luisramos0:
I believe this is all we need to do: [commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/2775/commits/2b3c694e17585f9bfca6c7268c1f37109e8f74c7)

The shipping_method_id logic in the browser is fine, nothing needs to be changed. The user selects a shipping_method and the id is sent to the server.
On the controller, instead of saving the shipping_method_id on the order (the field/column will be gone), we keep it locally on the controller until we get the order workflow to delivery (just after the shipments are created) and then we mark the shipping_rate for the shipping_method the user has selected as _selected_.
After this change we can drop the column shipping_method_id without a problem.

If you guys can validate this approach I'll proceed with making it ready: handling error scenarios, unit tests, dropping shipping_method_id column, etc

app/controllers/checkout_controller.rb
```diff
@@ -22,45 +22,41 @@ def edit
...
-        next if advance_order_state(@order)
-
-        if @order.errors.present?
-          flash[:error] = @order.errors.full_messages.to_sentence
-        else
-          flash[:error] = t(:payment_processing_failed)
-        end
-        update_failed
-        return
+    return update_failed unless @order.update_attributes(object_params)
```
mkllnk: :+1: for reducing the indentation level here. :-)

app/controllers/checkout_controller.rb
```diff
@@ -24,13 +24,21 @@ def edit
   def update
     return update_failed unless @order.update_attributes(object_params)
 
+    shipping_method_id = object_params[:shipping_method_id]
+    object_params.delete(:shipping_method_id)
```
mkllnk: Why do we delete that entry from the parameters? It has already been passed on two lines above.

app/models/concerns/order_shipping_method.rb
```diff
@@ -18,4 +18,17 @@ def shipping_method
...
   end
+
+  # Finds the shipment shipping_rate for the given shipping_method_id and selects that shipping_rate
+  #
+  # @return [Shipment]
+  def select_shipping_method(shipping_method_id)
+    return if shipments.empty?
+    shipment = shipments.first
+
+    shipping_rate = shipment.shipping_rates.find{ |shipping_rate| shipping_rate.shipping_method_id == shipping_method_id }
```
mkllnk: Is this an ActiveRecord relation? Then we could do `shipment.shipping_rates.find_by_shipping_method_id(shipping_method_id)` and let the database handle this.

app/models/concerns/order_shipping_method.rb
```diff
@@ -18,4 +18,17 @@ def shipping_method
...
+    return if !shipping_rate
+
+    shipment.selected_shipping_rate_id=(shipping_rate.id)
```
mkllnk: Is this the way Spree does it?

app/models/concerns/order_shipping_method.rb
mkllnk: And we can skip it if `shipping_method_id` is empty.

app/controllers/checkout_controller.rb
luisramos0: ah, it's a mistake. I need to add unit tests to this and make sure it works.
We will delete shipping_method_id before calling update_attributes because order.shipping_method_id will not be in the model (we will drop the column afterwards).

Even if ActiveRecord can handle inexistent attributes on update_attributes (I am not sure about this) it's better to keep this explicit: the difference between UI code (object_params) where order.shipping_method_id is used and DB/AR code (@order.update_attributes) where order.shipping_method_id is not present.

luisramos0:
Great @mkllnk 
thanks a lot for your awesome review here. and welcome back!
I'll get this documented, tested and get your (very useful) questions answered. Hopefully this week.

app/models/concerns/order_shipping_method.rb
luisramos0: no, in Spree's multi-step checkout process, the shipping_rates are rendered on the UI and the user selection is done on shipping_rates. We are not doing the same, we will let the user select a shipping_method and we will inject the selection on the shipping_rate ourselves. We could do the same even with single step checkout with some ajax call but we would have to change the rendering of the checkout page. I dont think this is necessary right now.

luisramos0:
@HugsDaniel rebased. lets look at the build :-D

spec/models/concerns/order_shipping_method_spec.rb
```diff
@@ -20,4 +20,43 @@
...
+      let(:shipping_method) { create(:shipping_method_with, :flat_rate) }
+      let(:shipment) { create(:shipment_with, :shipping_method, shipping_method: shipping_method) }
+
+      context "when shipping_method_id is valid for order" do
+        it "returns the shipments shipping_method" do
+          order.shipments = [shipment]
+          expect(order.select_shipping_method(shipping_method_id)).to eq shipment.shipping_method
+        end
+        #shipment.shipping_rates.find_by_shipping_method_id(shipping_method_id)
+        #       expect(Spree::PaymentMethod).to receive(:find).with(payment_method.id.to_s) { payment_method }
```
sauloperez: :warning: :fire:

app/models/concerns/order_shipping_method.rb
```diff
@@ -18,4 +18,18 @@ def shipping_method
     return if shipments.empty?
     shipments.first.shipping_method
   end
+
+  # Finds the shipment shipping_rate for the given shipping_method_id and selects that shipping_rate
+  #
+  # @return [Shipment]
+  def select_shipping_method(shipping_method_id)
+    return if shipping_method_id.nil? || shipments.empty?
```
sauloperez: [blank?](https://apidock.com/rails/v3.2.13/Object/blank%3F) might be what you need here.

spec/models/concerns/order_shipping_method_spec.rb
luisramos0: done

app/models/concerns/order_shipping_method.rb
luisramos0: :+1: done

luisramos0:
I've added test to the controller as well.
This PR is ready for final review!

spec/controllers/checkout_controller_spec.rb
```diff
@@ -85,6 +85,42 @@
...
 
+    context "#update with shipping_method_id" do
+      let(:test_shipping_method_id) { "111" }
+
+      before do
+        # stub order and resetorderservice
+        allow(ResetOrderService).to receive(:new).with(controller, order) { reset_order_service }
+        allow(reset_order_service).to receive(:call)
+        allow(order).to receive(:update_attributes).and_return true
+        controller.stub(:current_order).and_return(order)
```
sauloperez: please, use `allow` instead. There has to be a way to check this automatically :thinking:

spec/controllers/checkout_controller_spec.rb
```diff
@@ -85,6 +85,42 @@
...
+
+        # make order workflow pass through delivery
+        order.stub(:next).twice do
```
sauloperez: here too

spec/controllers/checkout_controller_spec.rb
luisramos0: :+1:
done

spec/controllers/checkout_controller_spec.rb
luisramos0: done

app/controllers/checkout_controller.rb
```diff
@@ -22,45 +22,49 @@ def edit
...
-
-        if @order.errors.present?
-          flash[:error] = @order.errors.full_messages.to_sentence
-        else
-          flash[:error] = t(:payment_processing_failed)
-        end
-        update_failed
-        return
+    shipping_method_id = object_params[:shipping_method_id]
+    object_params.delete(:shipping_method_id)
```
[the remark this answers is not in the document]
luisramos0: ah, thank you!

app/models/concerns/order_shipping_method.rb
```diff
@@ -18,4 +18,18 @@ def shipping_method
     return if shipments.empty?
     shipments.first.shipping_method
   end
+
+  # Finds the shipment shipping_rate for the given shipping_method_id and selects that shipping_rate
+  #
+  # @return [Shipment]
+  def select_shipping_method(shipping_method_id)
```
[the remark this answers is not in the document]
luisramos0: Interesting comment, thanks!
Initially I was tempted to call this method shipping_method= but then decided to move away from that and make it look more like a service method.
I think the best is to keep the code as is. I followed both suggestions as comments and a new assertion. See new commit.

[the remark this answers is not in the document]
kristinalim: I saw it. Thanks, @luisramos0!

PR #2778, opened 2018-09-27
Refactor entry processor

sauloperez:
#### What? Why?

This PR serves two purposes. Firstly, make product import possible to change, following the mantra

> Make the change easy (warning: this may be hard), then make the easy change
> Sandy Metz

so that `update_all(count_on_hand: 0)` can be changed in the upgrade with the appropriate implementation in Spree v2.0.

Secondly, as it was needed for the refactoring process, this PR's aim is also to increase the code coverage substantially. As most of these methods were rather complex adding tests for them was an impossible task. Now, we can be a bit more confident the next time we touch product import, which as I said, it's going to come soon.

⚠️ **I stronlgy recommend following commit by commit** as the final diff contains lots of small changes which make hard to understand the whole picture.

Surely the refactoring could go on and improve these classes even further but at this point we already achieved our goals.

#### What should we test?
Product Import.

#### Release notes
Refactor product import to increase its changeability and add tests

Changelog Category: Changed

Without this refactoring PR, making `#reset_absent_items` work in Sree 2.0 would be impossible to review.

### DIFF
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -4,11 +4,9 @@
 
 module ProductImport
   class EntryProcessor
-    delegate :products_reset_count, to: :reset_absent
-
     attr_reader :inventory_created, :inventory_updated, :products_created,
                 :variants_created, :variants_updated, :supplier_products,
-                :total_supplier_products
+                :total_supplier_products, :products_reset_count
 
     def initialize(importer, validator, import_settings, spreadsheet_data, editable_enterprises, import_time, updated_ids)
       @importer = importer
@@ -24,6 +22,7 @@ module ProductImport
       @products_created = 0
       @variants_created = 0
       @variants_updated = 0
+      @products_reset_count = 0
       @supplier_products = {}
       @total_supplier_products = 0
     end
@@ -65,7 +64,7 @@ module ProductImport
     def reset_absent_items
       return unless settings.data_for_stock_reset? && settings.reset_all_absent?
 
-      reset_absent.call
+      @products_reset_count = reset_absent.call
     end
 
     def reset_absent
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -198,12 +198,14 @@ module ProductImport
     end
 
     def assign_defaults(object, entry)
+      settings = Settings.new(@import_settings)
+
       # Assigns a default value for a specified field e.g. category='Vegetables', setting this value
       # either for all entries (overwrite_all), or only for those entries where the field was blank
       # in the spreadsheet (overwrite_empty), depending on selected import settings
-      return unless @import_settings.key?(:settings) && @import_settings[:settings][entry.supplier_id.to_s] && @import_settings[:settings][entry.supplier_id.to_s]['defaults']
+      return unless settings.defaults(entry)
 
-      @import_settings[:settings][entry.supplier_id.to_s]['defaults'].each do |attribute, setting|
+      settings.defaults(entry).each do |attribute, setting|
         next unless setting['active']
 
         case setting['mode']
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -85,6 +85,10 @@ module ProductImport
       end
     end
 
+    def strategy
+      @strategy ||= strategy_factory.new
+    end
+
     def total_saved_count
       @products_created + @variants_created + @variants_updated + @inventory_created + @inventory_updated
     end
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -4,7 +4,9 @@
 
 module ProductImport
   class EntryProcessor
-    attr_reader :inventory_created, :inventory_updated, :products_created, :variants_created, :variants_updated, :products_reset_count, :supplier_products, :total_supplier_products, :import_settings
+    delegate :products_reset_count, to: :reset_absent
+
+    attr_reader :inventory_created, :inventory_updated, :products_created, :variants_created, :variants_updated, :supplier_products, :total_supplier_products, :import_settings
 
     def initialize(importer, validator, import_settings, spreadsheet_data, editable_enterprises, import_time, updated_ids)
       @importer = importer
@@ -20,7 +22,6 @@ module ProductImport
       @products_created = 0
       @variants_created = 0
       @variants_updated = 0
-      @products_reset_count = 0
       @supplier_products = {}
       @total_supplier_products = 0
     end
@@ -60,7 +61,11 @@ module ProductImport
     end
 
     def reset_absent_items
-      ResetAbsent.new(self).call
+      reset_absent.call
+    end
+
+    def reset_absent
+      @reset_absent ||= ResetAbsent.new(self)
     end
 
     def total_saved_count
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -81,7 +81,7 @@ module ProductImport
     end
 
     def strategy
-      @strategy ||= strategy_factory.new
+      @strategy ||= strategy_factory.new(settings.updated_ids)
     end
 
     def total_saved_count
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -6,12 +6,14 @@ module ProductImport
   class EntryProcessor
     delegate :products_reset_count, to: :reset_absent
 
-    attr_reader :inventory_created, :inventory_updated, :products_created, :variants_created, :variants_updated, :supplier_products, :total_supplier_products, :import_settings
+    attr_reader :inventory_created, :inventory_updated, :products_created,
+                :variants_created, :variants_updated, :supplier_products,
+                :total_supplier_products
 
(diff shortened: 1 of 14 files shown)


spec/models/product_import/reset_absent_spec.rb
```diff
@@ -134,16 +133,9 @@
           .to receive(:permission_by_id?).with('1') { false }
       end
 
-      it 'does not reset anything' do
+      xit 'does not reset anything' do
```
luisramos0: are you leaving these pending?

app/models/product_import/inventory_reset.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def reset
+      relation.update_all(count_on_hand: 0)
+    end
+
+    private
+
+    attr_reader :excluded_items_ids
+
+    def relation
```
sauloperez: The upgrade's changes will then be limited here...

app/models/product_import/products_reset.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def reset
+      relation.update_all(count_on_hand: 0)
+    end
+
+    private
+
+    attr_reader :excluded_items_ids
+
+    def relation
```
sauloperez: ...and here.

luisramos0:
I like the way you really follow that mantra!
I thought it was Kent Beck's 2012 https://twitter.com/kentbeck/status/250733358307500032?lang=en

app/models/product_import/reset_absent.rb
```diff
@@ -0,0 +1,33 @@
+module ProductImport
+  class ResetAbsent
```
luisramos0: I prefer classes with nouns as names and more explicit, in this case I believe AbsentEnterprisesResetter
And I'd make the main method more contextual, in this case, reset() instead of call()

app/models/product_import/reset_absent.rb
```diff
@@ -0,0 +1,33 @@
...
+    def initialize(entry_processor, settings, strategy)
+      @entry_processor = entry_processor
+      @settings = settings
+      @strategy = strategy
+
+      @products_reset_count = 0
+    end
+
+    # For selected enterprises; set stock to zero for all products/inventory
+    # that were not listed in the newly uploaded spreadsheet
```
luisramos0: I recommend documenting the return value (@products_reset_count).

app/models/product_import/entry_processor.rb
```diff
@@ -60,45 +63,38 @@ def count_existing_items
...
     def reset_absent_items
-      # For selected enterprises; set stock to zero for all products/inventory
-      # that were not listed in the newly uploaded spreadsheet
-      return unless data_for_stock_reset?
-      suppliers_to_reset_products = []
-      suppliers_to_reset_inventories = []
+      return unless settings.data_for_stock_reset? && settings.reset_all_absent?
 
-      settings = @import_settings[:settings]
+      reset_absent.call
```
luisramos0: I am not a big fan of the delegate line above:
`delegate :products_reset_count, to: :reset_absent`

I'd prefer a similar but more explicit solution like this:
on top:
`     - attr_reader :products_reset_count`
on the initialize method
`     - @products_reset_count = 0`
and on this line:
`     - @products_reset_count = reset_absent.call`

I think this solution is less smart... which is better.

sauloperez:
You're probably right. I did a too lazy google look up :troll:

app/models/product_import/reset_absent.rb
```diff
@@ -0,0 +1,33 @@
...
+    def call
+      settings.enterprises_to_reset.each do |enterprise_id|
+        next unless entry_processor.permission_by_id?(enterprise_id)
+
+        strategy << enterprise_id.to_i
```
luisramos0: imo creating a new operator is too smart for this simple task. A new operator means anyone that reads this code needs to understand what this operator is, etc, etc.
Why not just create a support data structure called something like authorized_enterprises and extract the logic?
I think this is easier on the brain:
```
def call
   reset_stock(authorized_enterprises(settings.enterprises_to_reset))
end

def authorized_enteprsies(all_enterprises_to_reset)
     authorized_enterprises = []
     all_enterprises_to_reset.each do |enterprise_id|
         next unless entry_processor.permission_by_id?(enterprise_id)
         authorized_enterprises << enterprise_id.to_i
     end
     authorized_enterprises
end

in reset_stock we could have a new arg authorized_enterprises and we can easily handle the empty supplier_ids.

Comment (luisramos0) [reply app/models/product_import/reset_absent.rb]: this also means you can get rid of the supplier_ids attribute in the strategies

File: app/models/product_import/reset_absent.rb [hunk 8]
```diff
@@ -0,0 +1,33 @@
+module ProductImport
+  class ResetAbsent
+    attr_reader :products_reset_count
+
+    def initialize(entry_processor, settings, strategy)
+      @entry_processor = entry_processor
+      @settings = settings
+      @strategy = strategy
```
luisramos0: I think strategy could have a more explicit name like reset_stock_strategy

app/models/product_import/products_reset.rb
```diff
@@ -0,0 +1,35 @@
+module ProductImport
+  class ProductsReset
```
luisramos0: I prefer nouns for class names, in this case, ProductsResetter would be better, or ProductsResetStrategy

app/models/product_import/settings.rb
```diff
@@ -0,0 +1,37 @@
...
+    def updated_ids
+      @import_settings[:updated_ids]
+    end
+
+    def enterprises_to_reset
+      @import_settings[:enterprises_to_reset]
+    end
+
+    def importing_into_inventory?
+      settings && settings['import_into'] == 'inventories'
```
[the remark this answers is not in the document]
sauloperez: I don't understand it either but I didn't go into detail about the settings structure. By hiding it under the `Settings` carpet I hope we can improve it.

app/models/product_import/reset_absent.rb
luisramos0: yes, sounds good. I am learning Ruby conventions. Thanks!
I must say I dislike the conventions. It's OO, these are objects, not actions.
Services should have single responsibility, but that's very different from having a single method in its API.
For example:
- PasswordManager service updates the password but can also reset it.
- UserBuilder can create user in slightly different ways.

sauloperez: Sure! it's just that services following these conventions in Rails are simply implementing the Command pattern. We can totally differ from these and have more complex services and still store them in `app/services`. However, if we think the service is simple enough to have just `#call`, `#perform` or whatever method, is worth sticking to these conventions.

app/models/product_import/products_reset_strategy.rb
```diff
@@ -0,0 +1,34 @@
...
+      @excluded_items_ids = excluded_items_ids
+    end
+
+    def reset(supplier_ids)
+      @supplier_ids = supplier_ids
+
+      if supplier_ids.present?
+        relation.update_all(count_on_hand: 0)
+      else
+        0
```
sauloperez: we wouldn't need this conditional branch if we were in Rails 4.2 :sob: See https://stackoverflow.com/questions/4877931/how-to-return-an-empty-activerecord-relation#10001043

sauloperez:
Thanks a lot for the feedback @luisramos0 I think it greatly improved the PR :heart_eyes:

luisramos0:
all right! test ready?

sauloperez:

RachL:
@sauloperez @myriamboure I've tested product creation and update. Works fine! What other feature / scenario is really important to test while testing product import? I'm trying to create test files that I will be able to use quickly depending on the staging server.

Testing notes:
https://docs.google.com/document/d/1pcLpXp3g3cFfWvVDljOOdodDr5aEmxwPpjwYu8mWje8/edit#

myriamboure:
Have you tested as well inventory import, and maybe the "reset absent products" as well as it seems that's connected also to this work @RachL ?

sauloperez:
Yes, @RachL. My apologies for not mentioning but I miss the reset absent test case. I'm not used to PI myself so I'll digg into it and let you know what to test.

sauloperez:
Ok, so if I understood correctly it's basically the case described in 

https://github.com/openfoodfoundation/openfoodnetwork/blob/aa78b9e54c845cd5fcdcaa8ded6a6a516db9eccf/spec/features/admin/product_import_spec.rb#L177-L202

We need to check that when selecting `Reset absent products` indeed their stock level gets to 0.

sauloperez:
Btw, this can be shipped outside of the upgrade but it is required for it. It might not be clear in the description. cc @daniellemoorhead

RachL:
@sauloperez Thank you. If I've understood correctly all absent products should get their stock to 0, but in the lists there are 2 products whose stock did not change (other than my 2 mushrooms that are in my file) : Garlic and Fuji Apple

Do you know why ?
Testing notes:
https://docs.google.com/document/d/1pcLpXp3g3cFfWvVDljOOdodDr5aEmxwPpjwYu8mWje8/edit

Matt-Yorkley:
@RachL Those two products were present in the CSV file, right? All the other products were reset to zero? That's working as intended. :+1:

RachL:
@Matt-Yorkley oh you're right they were from enterprise 3 ! I did so many test I've forgot to clear Enterprise 3 products. We are good to go :)

PR #2615, opened 2018-08-30
2575 variant on hand api

mkllnk:
#### What? Why?

Closes #2575 and closes #2450.

Spree 2.0 dropped methods like `Variant.on_hand=` to manage stock. Since a lot of our code uses these methods, we are restoring the old interface here.

In a set of 927 related specs it fixes 237 tests (from 398 to 161).

A major caveat is that `on_demand` and `count_on_hand` can only be set when the variant is already saved in the database. Saving creates the stock items that hold this data (`after_create :create_stock_items` in Spree code). It would be much better if the stock items were build in memory when the variant is build. I'm not sure if later versions of Spree go that way.

I feel like I rushed with this. So I'm not 100% confident that I explored all the implications. We may discover more when we look at line items and the `increase` and `decrease` methods which I ignored here.

#### What should we test?

Manual testing is not available for the Spree upgrade yet.

#### Release notes

Not released yet.

#### How is this related to the Spree upgrade?

This is part of the Spree upgrade.

PR #2578, opened 2018-08-23
Spree2 fix test line items

luisramos0:
#### What? Why?

Closes #2577 

See issue for details.

#### What should we test?
More passing specs in ofn2 branch.

#### Release notes
No release notes.

#### How is this related to the Spree upgrade?
This is part of the spree upgrade. Adapting test specs to spree2.

PR #2779, opened 2018-09-27
[Spree Upgrade] Fixed several specs by defining line_items.target_shipmnt

luisramos0:
The fix is simply to provide target_shipment in test line_items.

#### What? Why?

Fixing more tests like in #2578 (setting target_shipment in line items).
Mostly errors uncovered by the fix #2615  (missing variant.on_hand=)

#### What should we test?
More working specs.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -280,7 +280,7 @@ FactoryBot.define do
     after(:create) { |c| c.set_preference(:per_kg, 0.5); c.save! }
   end
 
-  factory :order_with_totals_and_distribution, :parent => :order do #possibly called :order_with_line_items in newer Spree
+  factory :order_with_totals_and_distribution, parent: :order do
     distributor { create(:distributor_enterprise) }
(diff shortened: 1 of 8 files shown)


spec/factories.rb
```diff
@@ -280,7 +280,7 @@
     after(:create) { |c| c.set_preference(:per_kg, 0.5); c.save! }
   end
 
-  factory :order_with_totals_and_distribution, :parent => :order do #possibly called :order_with_line_items in newer Spree
+  factory :order_with_totals_and_distribution, :parent => :order do
```
sauloperez: can we dehashrocketize it now that we're at it?

luisramos0: ah, yes, but only this line please, otherwise I'll be creating merging issues all over...

sauloperez: Or you can run rubocop with the autofix argument to all factories afterwards :trollface:

PR #2780, opened 2018-09-28
[Spree Upgrade] Fix Stripe connect factory

### LINKED ISSUE #2781: [Spree Upgrade] Stripe Gateway Payment Method ActiveRecord config is broken
Issue represented by this build error:
```
      Failure/Error: let!(:payment_method) { create(:stripe_payment_method, distributors: [enterprise], preferred_enterprise_id: enterprise.id) }
      
      NoMethodError:
        undefined method `association_class' for nil:NilClass
      # ./spec/requests/checkout/stripe_connect_spec.rb:11:in `block (2 levels) in <top (required)>'
```

This is a problem in Spree::Gateway::StripeConnect, although it inherits from Spree::PaymentMethod,  it does NOT pick up the ActiveRecord mappings defined in OFN's PaymentMethod decorator.

This is seen in the factory stripe_payment_method when defining distributors_ids BUT this is not specific to the stripe payment method factory and this is not specific to payment_method distributors mapping because this error is also happening in the payment methods serializer with tag_list (another property defined in Payment Method decorator and used in StripeConnect).

This is the class structure:
a

luisramos0:
EDITED (ready for review now)

#### What? Why?

Closes #2781 
Fix the issue by copying distributors and distributors_ids from payment_method decorator to gateway decorator (a subclass of PaymentMethod, see issue for image) and then extract it to a concern.

This is not perfect but looks like it does solve the problem. I can't find a better solution.

#### What should we test?
This PR moves ~30 tests to different stages: ~10 are now passing, ~10 are now failing with errors related to "allow_backorders" and 10~ are now failing with other errors.

### DIFF
--- a/app/models/spree/concerns/payment_method_distributors.rb
+++ b/app/models/spree/concerns/payment_method_distributors.rb
@@ -1,0 +1,15 @@
+require 'active_support/concern'
+
+# This concern is used to duplicate the associations distributors and distributor_ids
+#   across payment method and gateway
+#   this fixes the inheritance problem https://github.com/openfoodfoundation/openfoodnetwork/issues/2781
+module Spree::PaymentMethodDistributors
+  extend ActiveSupport::Concern
+
+  def self.included(base)
+    base.class_eval do
+      attr_accessible :distributor_ids
+      has_and_belongs_to_many :distributors, join_table: 'distributors_payment_methods', class_name: 'Enterprise', foreign_key: 'payment_method_id', association_foreign_key: 'distributor_id'
+    end
+  end
+end
--- a/app/models/spree/gateway_decorator.rb
+++ b/app/models/spree/gateway_decorator.rb
@@ -2,4 +2,7 @@ Spree::Gateway.class_eval do
   # Default to live
   preference :server, :string, :default => 'live'
   preference :test_mode, :boolean, :default => false
+
+  attr_accessible :distributor_ids
+  has_and_belongs_to_many :distributors, join_table: 'distributors_payment_methods', :class_name => 'Enterprise', foreign_key: 'payment_method_id', association_foreign_key: 'distributor_id'
 end
--- a/app/models/spree/gateway_decorator.rb
+++ b/app/models/spree/gateway_decorator.rb
@@ -1,8 +1,9 @@
+require 'spree/concerns/payment_method_distributors'
+
 Spree::Gateway.class_eval do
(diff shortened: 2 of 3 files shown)


luisramos0:
I found a reference to these classes [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/initializers/spree.rb#L12).
And from the commit log on these 2 lines I found @oeoeaio trying to fix this problem in the past [here](https://github.com/openfoodfoundation/openfoodnetwork/commit/1059c170d1288c6ec5cee89ea894091a6fa2ab2a) and then with success [here](https://github.com/openfoodfoundation/openfoodnetwork/commit/b2e14711acc43ef80928416289da4296a341b319).

What was removed in this last commit b2e14711 is very similar to what I propose in this PR...

luisramos0:
Thanks @mkllnk those are great options. I'll check where this code is being used.

luisramos0:
thanks @mkllnk 
I have removed the rubocop upgrade, it will be handled in another PR.

app/models/spree/concerns/payment_method_distributors.rb
```diff
@@ -0,0 +1,15 @@
...
+# This concern is used to duplicate the associations distributors and distributor_ids
+#   across payment method and gateway
+#   this fixes the inheritance problem https://github.com/openfoodfoundation/openfoodnetwork/issues/2781
+module Spree::PaymentMethodDistributors
+  extend ActiveSupport::Concern
+
+  def self.included(base)
+    base.class_eval do
+      attr_accessible :distributor_ids
+      has_and_belongs_to_many :distributors, join_table: 'distributors_payment_methods', :class_name => 'Enterprise', foreign_key: 'payment_method_id', association_foreign_key: 'distributor_id'
```
sauloperez: can we dehashrocketize `:class_name`? I guess it's copy-paste leftover.

luisramos0: yes, I just used what was in payment_method_decorator
that was beautifully said btw: I dehashrocketize, you dehashrocketize, he dehashrocketizes
I have rebased and I have dehashrocketized it
#dehashrocketize

sigmundpetersen: hash de rocketized?

luisramos0: eheh

luisramos0:
@mkllnk can we merge this?

PR #2785, opened 2018-09-29
[Spree Upgrade] Fix order_updater_spec by using create (instead of build) in shipment

### LINKED ISSUE #2784: [Spree Upgrade] Fix order_updater broken specs
Order Updater specs broken with error related to shipments.

```
47) OrderUpdater#before_save_hook when any of the shipping methods doesn't require a delivery address populates the shipping address
      Failure/Error: next if shipping_method.require_ship_address
      
      NoMethodError:
        undefined method `require_ship_address' for nil:NilClass
      # ./app/models/order_updater.rb:52:in `block in shipping_address_from_distributor'
      # ./app/models/order_updater.rb:50:in `shipping_address_from_distributor'
      # ./app/models/order_updater.rb:35:in `before_save_hook'
      # ./spec/models/order_updater_spec.rb:135:in `block (4 levels) in <top (required)>'

```

luisramos0:
Also simplified spec by using prebuild shipment_with and shipping_method factories

#### What? Why?

Closes #2784

#### What should we test?
Order Updater spec should be green

### DIFF
--- a/spec/models/order_updater_spec.rb
+++ b/spec/models/order_updater_spec.rb
@@ -117,49 +117,30 @@ describe OrderUpdater do
 
   context '#before_save_hook' do
     let(:distributor) { build(:distributor_enterprise) }
-    let(:order) { build(:order, distributor: distributor) }
-    let(:shipment) { build(:shipment) }
-    let(:shipping_rate) do
-      Spree::ShippingRate.new(
-        shipping_method: shipping_method,
-        selected: true
-      )
-    end
+    let(:shipment) { create(:shipment_with, :shipping_method, shipping_method: shipping_method) }
 
     before do
-      shipment.shipping_rates << shipping_rate
-      order.shipments << shipment
+        order.distributor = distributor
+        order.shipments = [shipment]
     end
 
-    context 'when any of the shipping methods doesn\'t require a delivery address' do
-      let(:shipping_method) { build(:base_shipping_method, require_ship_address: false) }
+    context 'when shipping method is pickup' do
+      let(:shipping_method) { create(:shipping_method_with, :pickup) }
+      let(:address) { build(:address, firstname: 'joe') }
+      before { distributor.address = address }
 
-      let(:delivery_shipment) { build(:shipment) }
-      let(:delivery_shipping_rate) do
-        Spree::ShippingRate.new(
-          shipping_method: build(:base_shipping_method, require_ship_address: true),
-          selected: true
-        )
-      end
-
-      before do
-        delivery_shipment.shipping_rates << delivery_shipping_rate
-        order.shipments << delivery_shipment
-      end
-
-      it "populates the shipping address" do
+      it "populates the shipping address from distributor" do
         order_updater.before_save_hook
         expect(order.ship_address.firstname).to eq(distributor.address.firstname)
       end
     end
 
-    context 'when any of the shipping methods requires a delivery address' do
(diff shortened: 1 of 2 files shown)


luisramos0:
I have checked the build and spec/models/order_updater_spec.rb has no errors now.

spec/models/order_updater_spec.rb
```diff
@@ -101,49 +101,30 @@
...
     let(:distributor) { build(:distributor_enterprise) }
-    let(:order) { build(:order, distributor: distributor) }
-    let(:shipment) { build(:shipment) }
-    let(:shipping_rate) do
-      Spree::ShippingRate.new(
-        shipping_method: shipping_method,
-        selected: true
-      )
-    end
+    let(:shipment) { create(:shipment_with, :shipping_method, shipping_method: shipping_method) }
```
sauloperez: Isn't the use of the `:shipping_method` trait a bit redundant here? As we're passing a shipping method in this would be equivalent of the simpler `create(:shipment, shipping_method: shipping_method)`. Am I missing something?

spec/models/order_updater_spec.rb
```diff
@@ -101,49 +101,30 @@
...
-      let(:shipping_method) { build(:base_shipping_method, require_ship_address: false) }
-
-      let(:delivery_shipment) { build(:shipment) }
-      let(:delivery_shipping_rate) do
-        Spree::ShippingRate.new(
-          shipping_method: build(:base_shipping_method, require_ship_address: true),
-          selected: true
-        )
-      end
+    context 'when shipping method is pickup' do
```
sauloperez: thanks for sticking to the more meaningful pickup/delivery naming :clap:

spec/models/order_updater_spec.rb
```diff
@@ -101,49 +101,30 @@
...
+      let(:shipping_method) { create(:shipping_method_with, :pickup) }
```
sauloperez: is there any reason to have a separate shipping method factory other than better readability when using traits?

spec/models/order_updater_spec.rb
luisramos0: I dont think so.
The spree shipment factory does create(:shipping_method)
but the new ofn shipment_with factory has the same create(:shipping_method) but inside a transient block, which means that if shipping_method is provided, it will get used.
and in this case we are providing different shipping_methods, once as pickup and then as delivery.
Makes sense?

spec/models/order_updater_spec.rb
luisramos0: yeah, I could have extended the spree shipping_method factory, but I think it's better if we keep it with a different name. so, not just readability but also separation from spree code (I think it's similar to extending the class instead of using class_eval).

spec/models/order_updater_spec.rb
sauloperez: So, for me to understand, if no shipping method is provided none will be created, am I right?

luisramos0: no, if no shipping_method is provided the code inside transient is used, in this specific shipment_with factory, the default :shipping_method factory will be used.

mkllnk:
@sauloperez Can you give this another review?

luisramos0:
lets wait for the merge from master before we merge this one.

luisramos0:
Although there were no conflicts, this spec was a difficult merge on #2683 
So I decided to rebase this one and run the tests. It's all good. No code change since the approvals, so I am using my merger powers on this one.

Issue #2788, opened 2018-09-29
[Spree Upgrade] Fix pending specs in Order Syncer and Proxy Order (Subscriptions)


luisramos0:
order_syncer_spec is broken with 9 broken specs.
These specs were made pending but need to be fixed at some point.

Move all specs in proxy_order_spec to pending (order syncer and proxy order are specific to subscriptions).

See #2789 and #2902 and #3071

daniellemoorhead:
If this is blocked and can't be worked on (and thus not `In Dev`) should it be moved back into the Spree Upgrade backlog until it's unblocked? Ping @kristinalim 😄

kristinalim:
Fixing just the parts of the blocker issue #2696 that fix this one. I moved #2696 to In Dev as well.

And reversed the other dependency relationship, so this one blocks #3410.

Sorry!

PR #2789, opened 2018-09-29
[Spree Upgrade] Making broken OrderSyncerSpec tests pending, 9 tests to be fixed in issue #2788

luisramos0:
Related to #2788 
This is making 9 specs in order_syncer_spec pending. #2788  represents the work to fix this.

### DIFF
--- a/spec/services/order_syncer_spec.rb
+++ b/spec/services/order_syncer_spec.rb
@@ -1,3 +1,5 @@
+require 'spec_helper'
+
 describe OrderSyncer do
   describe "updating the shipping method" do
     let(:subscription) { create(:subscription, with_items: true, with_proxy_orders: true) }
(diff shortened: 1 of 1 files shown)


spec/services/order_syncer_spec.rb
```diff
@@ -1,3 +1,5 @@
+require 'spec_helper'
+
 describe OrderSyncer do
```
sauloperez: Isn't `OrderSyncer` a bit too core to make it pending? Doesn't it play an important role in placing orders or it isn't related to order placement?

luisramos0: I believed the docs, it's subscriptions specific:
# Responsible for ensuring that any updates to a Subscription are propagated to any
# orders belonging to that Subscription which have been instantiated

Issue #2665, opened 2018-09-08
[Spree Upgrade] Fix issues related to shipments in adjustment_spec


luisramos0:
Quite a few erros related to shipments in spec/models/spree/adjustment_spec.rb

luisramos0:
First step done in #2670 .

There is more work to be done, we still need to make the tests related to shipping fees pass.

luisramos0:
This issue will be closed when #2811, #2812 and #2816 are merged.

daniellemoorhead:
>This issue will be closed when #2811, #2812 and #2816 are merged.

Does the issue still need to sit in the code review column until these are merged? Or can it be unattached from the others and reattached to one of these?

luisramos0:
I think we can leave the issue connections as they are.
This is currently in code review and will be closed after all the 3 PRs are merged. I can do that manually hopefully next week after Maikels review.

PR #2811, opened 2018-10-02
[Spree Upgrade] Fixes shipping tests in adjustment spec

luisramos0:
#### What? Why?

Continues #2665 

Fixes shipping adjustment spec tests.
Enterprise Fees specs still need to be fixed.

#### What should we test?
Nothing for now. adjustment_spec should have 4 more green tests.


app/models/spree/shipment_decorator.rb
```diff
@@ -1,7 +1,13 @@
 module Spree
   Shipment.class_eval do
     def ensure_correct_adjustment_with_included_tax
-      ensure_correct_adjustment_without_included_tax
+      if adjustment
+        ensure_correct_adjustment_without_included_tax
+      elsif selected_shipping_rate_id
+        # ensures the attribute adjustment is set (which, in original spree method, even with the reload is not happening)
+        adjustment = shipping_method.create_adjustment shipping_method.adjustment_label, order, self, true, "open"
```
sauloperez: Is this `adjustment` local var really visible in line 11? 🤔

spec/factories.rb
```diff
@@ -374,6 +374,7 @@
         shipping_method { create :shipping_method }
       end
       after(:create) do |shipment, evaluator|
+        shipment.shipping_rates.destroy_all # remove existing shipping_rates from shipment
```
sauloperez: I would not add the comment IMO. It doesn't provide any extra info.

spec/models/spree/adjustment_spec.rb
```diff
@@ -61,38 +61,43 @@ module Spree
...
+        let(:line_item)       { create(:line_item, order: order) }
         let(:adjustment)      { order.adjustments(:reload).shipping.first }
 
-        it "has a shipping charge of $50" do
-          adjustment.amount.should == 50
+        describe "the shipping charge" do
+          it "is the adjustment amount" do
+            order.shipments = [shipment]
+
+            adjustment.amount.should == 50
```
sauloperez: Please, use RSpec3's `expect`.

spec/models/spree/adjustment_spec.rb
```diff
@@ -61,38 +61,43 @@ module Spree
...
             adjustment.included_tax.should == 0
           end
         end
 
         describe "when tax on shipping is enabled" do
-          before do
-            Config.shipment_inc_vat = true
-            Config.shipping_tax_rate = 0.25
-          end
+        before { Config.shipment_inc_vat = true }
```
sauloperez: Wrong alignment :warning:

sauloperez: Mmm Rubocop is not complaining?

app/models/spree/shipment_decorator.rb
luisramos0: adjustment is not a local var, adjustment is a has_one relation in shipment.
That is why the fix works.
This is the shipment method in spree:
https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/core/app/models/spree/shipment.rb#L273

spec/models/spree/adjustment_spec.rb
luisramos0: I am going to transpec'ize and rubocup'ize :-D this file when it's all green (when this, 2812 and 2816 are merged). Ok?

spec/models/spree/adjustment_spec.rb
luisramos0: rubocup is not running here, I dont know why. But I get it locally. Same as above, I'll get this sorted on a PR after this, 2812 and 2816 are merged.

spec/models/spree/adjustment_spec.rb
luisramos0: here: #2856

spec/models/spree/adjustment_spec.rb
luisramos0: here: #2856

app/models/spree/shipment_decorator.rb
luisramos0: ok, great, thanks a lot for the heads up.
You are completely right. I missed the simple fact I am setting the adjustment directly, not a property of it...
The solution would be to use self.adjustment as you suggest.

But, while debugging this and working on the order spec #2823 I realized we dont need this code. The fix will be on #2823 where the order.adjustments association is fixed.

On this PR, I have dropped this commit.

luisramos0:
@sauloperez can you please re-review?

spec/models/spree/adjustment_spec.rb
```diff
@@ -61,40 +61,43 @@ module Spree
...
+            order.shipments = [shipment]
 
-        it "has a shipping charge of $50" do
-          adjustment.amount.should == 50
+            adjustment.amount.should == 50
+          end
         end
 
         describe "when tax on shipping is disabled" do
+          before { Config.shipment_inc_vat = false }
```
mkllnk: This is not reverted after the test, is it? Depending on the test execution order it could have confusing side effects. But that's not the scope of your pull request. So no need to change.

PR #2812, opened 2018-10-02
[Spree Upgrade] Fix enterprise fees specs in adjustment_spec - tax_rate

luisramos0:
#### What? Why?

Continues #2665 

Fixes issue in Tax Rate decorator.

#### What should we test?
spec/models/spree/adjustment.rb:145 is now green.
and spec/models/spree/tax_rate_spec.rb is still green.

### DIFF
--- a/app/models/spree/tax_rate_decorator.rb
+++ b/app/models/spree/tax_rate_decorator.rb
@@ -25,9 +25,8 @@ module Spree
     # LineItems or Orders, so we mock out a line item here to fit the interface
     # that our calculator (usually DefaultTax) expects.
     def compute_tax(amount)
-      product = OpenStruct.new tax_category: tax_category
       line_item = LineItem.new quantity: 1
-      line_item.define_singleton_method(:product) { product }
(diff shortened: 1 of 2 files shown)


luisramos0:
I have used this [feedback in #2824](https://github.com/openfoodfoundation/openfoodnetwork/pull/2824#discussion_r223002950) from @sauloperez to improve this PR as well.

luisramos0:
ready for re-review ;-)

app/models/spree/tax_rate_decorator.rb
```diff
@@ -25,9 +25,8 @@ def adjust_with_included_tax(order)
     # LineItems or Orders, so we mock out a line item here to fit the interface
     # that our calculator (usually DefaultTax) expects.
     def compute_tax(amount)
-      product = OpenStruct.new tax_category: tax_category
       line_item = LineItem.new quantity: 1
-      line_item.define_singleton_method(:product) { product }
```
sauloperez: :scream:

PR #2816, opened 2018-10-02
[Spree Upgrade] Fix enterprise fees specs in adjustment_spec - enterprise_fee_aplicator

luisramos0:
#### What? Why?

Continues #2665 

This PR is fixing a problem in enterprise_fees_aplicator caused by calculated_adjustment.create_adjustment where the calculable is the line_item but the adjustment.source is changed to an order.

#### What should we test?
spec/models/spree/adjustment.rb:245 is now green.
spec/lib/open_food_network/enterprise_fee_applicator_spec.rb is still green.

### DIFF
--- a/lib/open_food_network/enterprise_fee_applicator.rb
+++ b/lib/open_food_network/enterprise_fee_applicator.rb
@@ -1,23 +1,26 @@
 module OpenFoodNetwork
   class EnterpriseFeeApplicator < Struct.new(:enterprise_fee, :variant, :role)
     def create_line_item_adjustment(line_item)
-      a = enterprise_fee.create_adjustment(line_item_adjustment_label, line_item.order, line_item, true)
-
-      AdjustmentMetadata.create! adjustment: a, enterprise: enterprise_fee.enterprise, fee_name: enterprise_fee.name, fee_type: enterprise_fee.fee_type, enterprise_role: role
-
-      a.set_absolute_included_tax! adjustment_tax(line_item, a)
+      create_adjustment(line_item_adjustment_label, line_item.order, line_item)
     end
 
     def create_order_adjustment(order)
-      a = enterprise_fee.create_adjustment(order_adjustment_label, order, order, true)
-
-      AdjustmentMetadata.create! adjustment: a, enterprise: enterprise_fee.enterprise, fee_name: enterprise_fee.name, fee_type: enterprise_fee.fee_type, enterprise_role: role
-
-      a.set_absolute_included_tax! adjustment_tax(order, a)
+      create_adjustment(order_adjustment_label, order, order)
     end
 
-
     private
+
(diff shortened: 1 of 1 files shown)


lib/open_food_network/enterprise_fee_applicator.rb
```diff
@@ -1,23 +1,32 @@
...
+    private
+
+    def create_adjustment(label, target, calculable)
+      adjustment = create_enterprise_fee_adjustment(label, target, calculable)
 
-      AdjustmentMetadata.create! adjustment: a, enterprise: enterprise_fee.enterprise, fee_name: enterprise_fee.name, fee_type: enterprise_fee.fee_type, enterprise_role: role
+      AdjustmentMetadata.create! adjustment: adjustment, enterprise: enterprise_fee.enterprise, fee_name: enterprise_fee.name, fee_type: enterprise_fee.fee_type, enterprise_role: role
 
-      a.set_absolute_included_tax! adjustment_tax(order, a)
+      adjustment.set_absolute_included_tax! adjustment_tax(adjustment)
```
sauloperez: Is there a particular reason why we don't move `adjustment_tax(adjustment)` inside `set_absolute_included_tax!`'s definition? It really smells like [Feature Envy](https://refactoring.guru/smells/feature-envy). Then this would become something like `adjustment.recalculate_absolute_included_tax!` or similar.

luisramos0: yes, this is feature envy, well spotted.
I agree BUT...
first, I am rushing to green, you know red-green-(and-then)-refactor, refactoring this type of code with a broken build is crazy.
but the main reason is that I suspect what this object is envious about is something that should belong to it, not to the neighbour :-) but I am not sure and dont want to refactor... back to the first reason...

Did I say I understand everything about adjustments and taxes? 😭 
I am not sure what's going on.
Maybe you can help? :-)
Why would the adjustment object that has an association tax_rate (belongs), have a decorator that defines a method tax_rates just for enterprise_fees? :-)
See this commit 4a9c17cb2838327d228ac3e75ee68bc46deddf60
and this one 2d2792225a607a06dbd06aab694030a5cfa04d95

This is why I am not moving much in this space. Maybe after we have a green build...

luisramos0:
@sauloperez can you please have a look at this one please? thanks!

lib/open_food_network/enterprise_fee_applicator.rb
sauloperez: Agree. I shot to quick; it might be too early to refactor :sweat: .

> Maybe you can help? :-)

You know :100: times more than me about these things. I haven't been through this part of the schema.

luisramos0: ok :+1:
can you please approve the PR?

Issue #2435, opened 2018-07-05
[Spree Upgrade] Spree 2.0 removes on_demand property from  Product and Variant


Matt-Yorkley:
## Description

`on_demand` is removed from Product and Variant in these 2 PRs for Spree 2.0:

https://github.com/spree/spree/commit/40665ab244277138801ee278f663fd2f6da7da22#diff-c1dfa6db7daf5dcbf0bf6caad4024294
https://github.com/spree/spree/commit/14a398cd3625a1ef7173c1661d7013d51ea92446#diff-c1dfa6db7daf5dcbf0bf6caad4024294

There are 81 mentions of `on_demand` throughout our app folder. 

I think our users regularly use this option for product stock. 

## Expected Behavior

Products can be set to `on_demand`.

## Actual Behavior

It's gone in 2.0!

## Animated Gif/Screenshot

![shrug-gif](https://media.giphy.com/media/13NUOwOLq0NJug/200w_d.gif) 

## Context

Spree Upgrade.

Matt-Yorkley:
This will need a bit of a spike, but after some discussions there was a mention of a solution where we allow backorders on items. This will probably need to include looking at our current use of the `allow_backorders` config setting and the previous issues relating to it. I can't remember if it is just a large number of specs that depend on it being turned off, or whether or not some of our customisations make it unusable.

https://github.com/openfoodfoundation/openfoodnetwork/search?q=allow+backorders&type=Issues

Matt-Yorkley:
Thanks @daniellemoorhead, this was just here to document it as a reminder. Yes it can be filed under "all the Spree upgrade things". I think it will be covered in the new plans to keep the existing API in terms of models/methods.

luisramos0:
Useful conversation: https://github.com/spree/spree/commit/1ca45b826934d345c0e39dfebe9a707f649ea2ab
From what I have seen, it has not come back since then.
From this conversation here looks like we need to use backorders instead:
https://groups.google.com/forum/#!topic/spree-user/kUHJljlN2hQ

myriamboure:
I don't really care about the use of backlorders in replacement of on_demand, but what I care for is the UX, as most users actually do check the box "on demand" on product as they don't manage stock levels within OFN.
My concern is that today backorder is setup at instance level, whereas every hub, for every product, should be able to decide if their customer can order more than stocked / order as much as they want without specified stock info. So the "allow backorder" info should be at product level. Is it the case?

[...]

PR #2817, opened 2018-10-03
[Spree Upgrade] Remove on_demand from Product serializers

### LINKED ISSUE #2802: [Spree Upgrade] Remove reference to on_demand in Product serializers
## Description
`Api::CachedProductSerializer` still calls `on_demand` on Product which cases a `NoMethodError`.

## Possible Fix

Similar to `VariantStock`, we need to provide an adapter that provides backward compatibility between our code and Spree 2.0's `Product`.

sauloperez:
#### What? Why?

Closes #2802

`on_demand` is removed from Product. There's no apparent way to provide an adapter as we did with `VariantStock` because the feature is simply removed from Spree (and so it is its DB column).

This change also moves the spec file to its appropriate place for the sake of consistency and RSpec3-izes the test example.

##### `on_demand's` future

This PR just makes it possible to fix lots of failures from the test suite. It is a temporal solution while we figure out how we implement `on_demand` from OFN side so that the following holds true:

> I don't really care about the use of backlorders in replacement of on_demand, but what I care for is the UX, as most users actually do check the box "on demand" on product as they don't manage stock levels within OFN.
My concern is that today backorder is setup at instance level, whereas every hub, for every product, should be able to decide if their customer can order more than stocked / order as much as they want without specified stock info. So the "allow backorder" info should be at product level. Is it the case?

Check out the discussion on https://github.com/openfoodfoundation/openfoodnetwork/issues/2435#issuecomment-426635995 for details.

#### What should we test?
The upgrade is not ready yet

### DIFF
--- a/app/models/spree/product_decorator.rb
+++ b/app/models/spree/product_decorator.rb
@@ -128,6 +128,14 @@ Spree::Product.class_eval do
 
   # -- Methods
 
+  def on_hand
+    if has_variants?
+      variants.map(&:on_hand).reduce(:+)
+    else
+      master.on_hand
+    end
+  end
+
   # Called by Spree::Product::duplicate before saving.
   def duplicate_extra(parent)
     # Spree sets the SKU to "COPY OF #{parent sku}".
--- a/app/serializers/api/product_serializer.rb
+++ b/app/serializers/api/product_serializer.rb
@@ -4,7 +4,7 @@ class Api::ProductSerializer < ActiveModel::Serializer
   # TODO
   # Prices can't be cached? How?
   def serializable_hash
-    cached_serializer_hash.merge uncached_serializer_hash
+    cached_serializer_hash.merge(uncached_serializer_hash)
   end
 
   private
@@ -36,7 +36,7 @@ class Api::CachedProductSerializer < ActiveModel::Serializer
   include ActionView::Helpers::SanitizeHelper
 
   attributes :id, :name, :permalink, :meta_keywords
-  attributes :on_demand, :group_buy, :notes, :description, :description_html
+  attributes :group_buy, :notes, :description, :description_html
   attributes :properties_with_values
 
   has_many :variants, serializer: Api::VariantSerializer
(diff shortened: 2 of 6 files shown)


app/models/spree/product_decorator.rb
```diff
@@ -128,6 +128,14 @@
 
   # -- Methods
 
+  def on_hand
+    if has_variants?
+      variants.map(&:on_hand).reduce(:+)
+    else
+      master.on_hand
+    end
+  end
```
sauloperez: This mimics current Spree implementation. It was required for the specs not to fail. This `on_hand` attribute will be addressed in an upcoming PR.

sauloperez:
I can't find any failure due to `on_demand` so it did the trick.

app/models/spree/product_decorator.rb
luisramos0: For my understanding, can you pls explain what do you mean by "will be addressed"?

sauloperez: I didn't put to much thought on what the actual behaviour for Product#on_hand should be now that Spree removes it. Needs to be done in a different PR anyway.

It's likely that it'll delegate to master's on_hand.

luisramos0: all right, thanks for explaining.

app/serializers/api/product_serializer.rb
```diff
@@ -36,7 +36,7 @@ class Api::CachedProductSerializer < ActiveModel::Serializer
   include ActionView::Helpers::SanitizeHelper
 
   attributes :id, :name, :permalink, :meta_keywords
-  attributes :on_demand, :group_buy, :notes, :description, :description_html
```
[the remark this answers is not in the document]
sauloperez: yep, that is exactly it and my point in the description about `on_demand's future`.

sauloperez:
Job 1: 468 examples, 53 failures, 2 pending
Job 2: 542 examples, 109 failures
Job 3: 932 examples, 111 failures, 5 pending
Job 4: 1116 examples, 99 failures, 4 pending

sauloperez:
See the `on_demand's future` section of the PR's description for details. You'll find the goal of this PR as well as the discussion about this feature.

mkllnk:
I know, I've seen that section. And I understand the value of fixing tests. Can we at least create an issue which outlines all the affected frontend parts that need fixing?

luisramos0:
ok, I have created the issue for this one: #2870 under #2435 
@mkllnk can you pls approve?

sauloperez:
thanks @luisramos0 . I had to temporally abandon this because the S2 kept me quite busy.

PR #2818, opened 2018-10-04
Move query from variant_overrides_controller to its model scope

nikobozi:
#### What? Why?

This PR moves query code from `VariantOverrideController` into `VariantOverride` model as a new scope query, in order to extract extra logic from controller and also to reuse an existing scope `for_hubs`

#### What should we test?
The existing rspec tests should pass along with the new one

#### Release notes
Move query code from `VariantOverrideController` into its Model scope and add a spec  

Changelog Category: Changed

### DIFF
--- a/app/models/variant_override.rb
+++ b/app/models/variant_override.rb
@@ -3,8 +3,6 @@ class VariantOverride < ActiveRecord::Base
 
   acts_as_taggable
 
-  attr_accessor :import_date
-
   belongs_to :hub, class_name: 'Enterprise'
   belongs_to :variant, class_name: 'Spree::Variant'
 
@@ -19,6 +17,12 @@ class VariantOverride < ActiveRecord::Base
 
   scope :for_hubs, lambda { |hubs|
     where(hub_id: hubs)
+  }
+
+  scope :distinct_import_dates, lambda {
+    select('DISTINCT variant_overrides.import_date').
+      where('variant_overrides.import_date IS NOT NULL').
+      order('import_date DESC')
   }
 
   localize_number :price
--- a/app/controllers/admin/variant_overrides_controller.rb
+++ b/app/controllers/admin/variant_overrides_controller.rb
@@ -60,7 +60,9 @@ module Admin
     end
 
     def inventory_import_dates
-      import_dates = VariantOverride.distinct_import_dates.for_hubs(editable_enterprises.collect(&:id))
+      import_dates = VariantOverride.
+        distinct_import_dates.
+        for_hubs(editable_enterprises.collect(&:id))
 
       options = [{ id: '0', name: 'All' }]
       import_dates.collect(&:import_date).map { |i| options.push(id: i.to_date, name: i.to_date.to_formatted_s(:long)) }
--- a/app/controllers/admin/variant_overrides_controller.rb
+++ b/app/controllers/admin/variant_overrides_controller.rb
@@ -60,11 +60,7 @@ module Admin
     end
 
     def inventory_import_dates
-      import_dates = VariantOverride.
-        select('DISTINCT variant_overrides.import_date').
-        where('variant_overrides.hub_id IN (?)
-        AND variant_overrides.import_date IS NOT NULL', editable_enterprises.collect(&:id)).
-        order('import_date DESC')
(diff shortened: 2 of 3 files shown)


app/models/variant_override.rb
```diff
@@ -3,8 +3,6 @@ class VariantOverride < ActiveRecord::Base
 
   acts_as_taggable
 
-  attr_accessor :import_date
```
nikobozi: This was overriding the ActiveRecord getter/setter, which is why querying didn't work for me. When defining new accessor methods, shouldn't they have a different name than the database field?

app/models/variant_override.rb
```diff
@@ -21,6 +19,12 @@ class VariantOverride < ActiveRecord::Base
     where(hub_id: hubs)
   }
 
+  scope :distinct_import_dates, lambda {
+    select('DISTINCT variant_overrides.import_date').
+      where('variant_overrides.import_date IS NOT NULL').
+      order('import_date DESC')
```
sauloperez: just as a minor detail, could we rely on ActiveRecord's hash style? So it'd be like `order(import_date: :desc)`. It decouples the scope from SQL a bit.

See the examples in https://apidock.com/rails/ActiveRecord/QueryMethods/order.

app/models/variant_override.rb
sauloperez: Did we mean [attr_accessible](https://apidock.com/rails/ActiveRecord/Base/attr_accessible/class) instead, perhaps :thinking: ?

app/models/variant_override.rb
```diff
@@ -21,6 +19,12 @@ class VariantOverride < ActiveRecord::Base
     where(hub_id: hubs)
   }
 
+  scope :distinct_import_dates, lambda {
+    select('DISTINCT variant_overrides.import_date').
```
sauloperez: A pity that we're still on Rails 3.2 otherwise we could make use of AR's [distinct](https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/distinct)

sauloperez:
neat @nikobozi :clap: ! Keep them coming!

app/models/variant_override.rb
nikobozi: Apparently this also works only after Rails 4 :/ as well as `.where.not()` which could have been used in the line above

sauloperez: Really? :sob:

app/models/variant_override.rb
Matt-Yorkley: Looks fine to delete :+1:

myriamboure:
"import dates" was something added by @Matt-Yorkley for product import but we are not using this info anymore. Not sure we will use it later either, the filter on import dates should have disappeared as it's not anymore in any place of the product import feature as we just redirect to product or inventory page after import. But as an admin I still see it on product page, so maybe it was not removed. Maybe it's safer to keep the product import info for now and we will see if we delete later on if we don't do anything with it?

sauloperez:
According to the code we show an import date filter in the inventory (variant overrides) page.

luisramos0:
I am not sure it's clear to everyone, I suggested deleting that import_date line of code because it's technically redundant (the line of code, not the db field). This would not impact on functionality at all.

luisramos0:
staged https://staging.openfoodfrance.org

The  inventory (variant overrides) page needs to be tested.

myriamboure:
@sauloperez @luisramos0 is there any issue attached? hard to test without understanding the problem... @RachL Luis said the inventory page (variant override) needs to be tested. Make sure you unhide columns to test all overrides possible. @Matt-Yorkley do you think we need to keep that "import date" at all? I am not sure there will be any use for it, it was useful when you first designed it to display what had been imported but UX was floady aspecially for inventory... Anyway here Rachel you know what to test.

sauloperez:
This didn't arise from an existing issue but from an improvement that @nikobozi contributed with which is worth adding @myriamboure . There's no issue.

RachL:
I've been a bit long to come back on this one, sorry. Ready to go!

My testing notes:

https://docs.google.com/document/d/1kQXbx_BJb373YBk3U60TzKyJe4owjgP1DRWwJC0eoHw/edit#

sauloperez:
Thanks again @nikobozi ! keep 'em coming! :metal:

Issue #2799, opened 2018-10-01
Producer does not appear in inventory


OliverUK:
## Description
If I go to the inventory for Stroudco and try and filter for Producer:Wharf Farm, it no longer comes up. They also don't come up if I leave Producer unfiltered and search for their produce. This happens for me logged in as me and it happens when I log in as super admin but it seems to work fine for Nick.

## Expected Behavior
Inventory screen loads produce for Wharf Farm

## Actual Behavior
Screen displays " No matching products found in your inventory "

## Animated Gif/Screenshot
http://recordit.co/k6dAumaPgP

## Context
We use the inventory to reset how much a producer has available each week. 

## Your Environment
Firefox 62.0.2 on Mac 10.12.6

OliverUK:
With the fix of #2769 this too seems to be working again now. @myriamboure should I close?

myriamboure:
Thanks @OliverUK we will close when we have tested and merged the patch to master, leave it open for now, I have added that issue number to the patch to know we have to close it when tested.

sigmundpetersen:
I guess we can close this now that #2769 is tested and merged @myriamboure ?

Issue #2769, opened 2018-09-24
Saving failed notification when saving product changes in spree - UK 


sineadfenton:
## Description

In UK production site, in spree. Click save on the product page and you see an error message when infact the request has been processed. This is not the intended flow, instead you should see a saved/complete notification.

## Expected Behavior

You should seed a "saved" notification when you click save and not 'Saving failed.Save failed due to invalid data:422'. This should only show when there is an error.

## Actual Behavior

You click save and Saving failed.Save failed due to invalid data:422 comes up

## Steps to Reproduce

1. Go to product page
2. Make a change - I changed the weight of an item.
3. Click save
4. Saving processes and then you get Saving failed with "Save failed due to invalid data:422" at the bottom of the screen
5. Refresh the page. You'll see your changes have actually gone through

## Animated Gif/Screenshot

![image](https://user-images.githubusercontent.com/28807449/45972367-7397e680-c033-11e8-9fd9-9f0c89b5c9b7.png)

## Context

Users will be confused and this there is an error when there is not. this will result in more support requests

## Your Environment

* Version used:
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): Desktop, windows
UK production was upgraded to 1.20.0 today (24/09) and this issue is new as of today.

## Possible Fix

Have the correct message when a task is successfully completed, and error show up only when there is an error.

sineadfenton:
![](https://github.trello.services/images/mini-trello-icon.png) [Saving failed.Save failed due to invalid data:422](https://trello.com/c/UfiOaWiF/247-saving-failedsave-failed-due-to-invalid-data422)

myriamboure:
@sineadfenton I logged in today on UK production with a hub manager user to reproduce the bug and see how to move it forward. I couldn't reproduct it, I tested on both price and weight and had the correct "product saved" message. Can you please describe more precisely which user was concerned, were you logged in as super admin? What product exactly did you try changing, for which type of user, etc. ? Is the user trying to change the product info the owner of the producer profile or a hub manager that has right to edit products on that producer? Without those info I don't see how we can move forward, if we can't reproduce the bug...
![capture du 2018-09-25 17-40-43](https://user-images.githubusercontent.com/8135376/46026049-ab159a00-c0ea-11e8-8491-21de4d7f966b.png)
[...]

PR #2831, opened 2018-10-05
Fix product import date error when some but not all variants have import date

kristinalim:
#### What? Why?

Addresses #2769 and #2799  (Verify before closing issues)

The product import date is used when loading product data for the "Bulk Update Products" page. This however raises an error when some but not all variants have import date.

#### What should we test?

1. Log in as manager of an enterprise that manages some products that were imported and some products that were manually created.
2. Go to the "Bulk Update Products" page.
3. Enable the "Import Date" column.
4. Check that both types of products are listed, and that the import date of imported products are visible.
 
#### Release notes

- Fix listing of a product when some variants have import date and some do not.

Changelog Category: Fixed

### DIFF
--- a/spec/models/spree/product_spec.rb
+++ b/spec/models/spree/product_spec.rb
@@ -714,4 +714,45 @@ module Spree
       end
     end
   end
+
+  describe "product import" do
+    describe "finding the most recent import date of the variants" do
+      let!(:product) { create(:product) }
+
+      let(:reference_time) { Time.zone.now.beginning_of_day }
+
+      before do
+        product.reload
+      end
+
+      context "when the variants do not have an import date" do
+        let!(:variant_a) { create(:variant, product: product, import_date: nil) }
+        let!(:variant_b) { create(:variant, product: product, import_date: nil) }
+
+        it "returns nil" do
+          expect(product.import_date).to be_nil
+        end
+      end
+
+      context "when some variants have import date and some do not" do
+        let!(:variant_a) { create(:variant, product: product, import_date: nil) }
+        let!(:variant_b) { create(:variant, product: product, import_date: reference_time - 1.hour) }
+        let!(:variant_c) { create(:variant, product: product, import_date: reference_time - 2.hour) }
+
+        it "returns the most recent import date" do
+          expect(product.import_date).to eq(variant_b.import_date)
+        end
+      end
+
+      context "when all variants have import date" do
+        let!(:variant_a) { create(:variant, product: product, import_date: reference_time - 2.hour) }
+        let!(:variant_b) { create(:variant, product: product, import_date: reference_time - 1.hour) }
+        let!(:variant_c) { create(:variant, product: product, import_date: reference_time - 3.hour) }
+
+        it "returns the most recent import date" do
+          expect(product.import_date).to eq(variant_b.import_date)
+        end
+      end
+    end
+  end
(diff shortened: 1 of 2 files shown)


Matt-Yorkley:
I've been testing this on staging and have now pushed this PR to UK production, and it has definitely resolved the issue. I did a few extra bits of checking (partly because of those specs that are failing on all builds) and I'd say this is good to go.

spec/models/spree/product_spec.rb
```diff
@@ -714,4 +714,45 @@ module Spree
...
   end
+
+  describe "product import" do
+    describe "finding the most recent import date of the variants" do
+      let!(:product) { create(:product) }
+
+      let(:reference_time) { Time.zone.now.beginning_of_day }
+
+      before do
+        product.reload
```
sauloperez: why a reload if we just create the product? Do we really need it?

myriamboure:
It seems to have fixed https://github.com/openfoodfoundation/openfoodnetwork/issues/2799 as well, so I added it in the thread so that we test and make sure it is fixed as well before closing issue.

spec/models/spree/product_spec.rb
kristinalim: Yes, this was to uncache the initially empty `product.variants` which was getting cached somewhere during setup of the product. :slightly_smiling_face: I prefer this over reloading just `product.variants`, which could leave you with outdated fields, e.g. (hypothetical) `order.total` after adding to `order.line_items`.

myriamboure:
Ok I'm testing this now as it needs to me merged to master asap

myriamboure:
Ok here are my testing notes: https://docs.google.com/document/d/1SXVmM6Gn0UnoErDTxFuoKp4jjM4xHTzqtsGJaEOw-bY/edit?usp=sharing

I found 2 very annoying issues:
- Now on order pages filters on email are not working anymore (I tested that as I had some clue from #2843 and it seemed when they reverted the fix it solved the issue
- Also I saw an old bug reappearing on internationalization of numbers, now when "use international treatment for separators" is enabled, if I type in price 6,50 it should transform in 6.50 so price remains the same even if separator is shown as . but now we have back the problem we had before, if I type 6,50 it gets transformed into 650.00 so as you see... the price changes.

There might be some other issues, I didn't test evberything but it seems this commit is breaking quite a lot of different things... I tested these things on an instance with V1.20 deployed and they are working fine there.

So @kristinalim this is back to you.
-

myriamboure:
Oups sorry on the internationalisation issue the error might come on my side, let me check and don't move until my next message :-)

myriamboure:
Ok I can confirm I didn't save the configuration setup so sorry, I will remove the comments and mention of it in this thread and the testing notes to avoid any confusion.

myriamboure:
Given discussion with @Matt-Yorkley on https://github.com/openfoodfoundation/openfoodnetwork/pull/2844 the bug that I noticed is not connected to that PR and was introduced by another commit after v1.20 release. SO this is ready to go then ! I open straight away the other issue for category duplication which is not connected either to that PR.

spec/models/spree/product_spec.rb
Matt-Yorkley: @sauloperez Can I merge this now? :heavy_check_mark: :thinking:

Issue #2739, opened 2018-09-17
Inventories broken: changes to inventory don't save


myriamboure:
## Description
It's not possible anymore to update a price, stock, etc. in inventory, we get the following error message:
![df28d41ac1d20369445d7df193b0735d112a71d0_1_654x500](https://user-images.githubusercontent.com/8135376/45647736-e69add80-bac6-11e8-887e-8f4764c09a8e.png)

## Expected Behavior
Inventory work and hub manager can save changes to prices, stock, etc.

## Actual Behavior
Not working.

## Steps to Reproduce
1. As a super admin or hub manager, go to products > inventories, choose a hub
2. Try to update a price, see the error message

## Context
Reported by a user in Canada, reproduced by Myriam on french instance

## Severity
S2 it doesn’t seem there is any workaround, other than modifying source info in product catalog (but not possible in some uses of inventory) or creating a new producer catalog to replace inventory but it’s not a quick workaround… not really usable.

## Your Environment
* Version used: v1.18 on French instance (not sure which version it is in Canada) so it's not linked to the last v1.19 release.
* Browser name and version: Firefox Quantum 62.0
* Operating System and version (desktop or mobile): Ubuntu 18.04.1 LTS

## Possible Fix
No idea...

myriamboure:
From @luisramos0 on Slack:

> looks like an authorization problem (401) on the variantoverides controller action bulk_update

**And more details:**

lets read together :joy:
   can [:admin, :index, :read, :update, :bulk_update, :bulk_reset], VariantOverride do |vo|
     next false unless vo.hub.present? && vo.variant.andand.product.andand.supplier.present?

     hub_auth = OpenFoodNetwork::Permissions.new(user).
       variant_override_hubs.
       include? vo.hub

     producer_auth = OpenFoodNetwork::Permissions.new(user).
       variant_override_producers.
       include? vo.variant.product.supplier

     hub_auth && producer_auth
   end
this is the logic used to validate if a user is authorized to bulk_update variantoverrides

These are things you can check on the instance, to see if some data config is broken:
- first line, the VO must have hub and the product of the VO must have a supplier
- last line - both hub_auth and producer_auth must be TRUE to grant authorization to this action
- hub_auth - user must be a manager of VO hub and VO.hub must be a distributor (sells != ‘none’)
- producer_auth - this one is more complicated but is related to permissions set in EnterpriseRelationships

myriamboure:
[...]

PR #2833, opened 2018-10-06
Add tests for permissions for variant overrides

kristinalim:
#### What? Why?

Relates to #2739 (Verify before closing issue)

This PR adds tests outlining the rules of permissions for variant overrides.

```
permissions for variant overrides
  when admin
    should have permission
  when user of the producer
    should not have permission
  when user of the distributor
    should not have permission
  when user of the distributor which is also the producer
    should have permission
  when owner of the distributor with add_to_order_cycle permission to the producer
    should not have permission
  when owner of the enterprise with create_variant_overrides permission to the producer
    should not have permission
    when the enterprise is not a distributor
      should not have permission
```

#### What should we test?
Build is green and all these new specs are green.

#### Release notes

Changelog Category: Added
Improved tests around user access to manage inventories.

Not related.

### DIFF
--- a/spec/models/spree/ability_spec.rb
+++ b/spec/models/spree/ability_spec.rb
@@ -514,5 +514,89 @@ module Spree
         end
       end
     end
+
+    describe "permissions for variant overrides" do
+      let!(:distributor) { create(:distributor_enterprise) }
+      let!(:producer) { create(:supplier_enterprise) }
+      let!(:product) { create(:product, supplier: producer) }
+      let!(:variant) { create(:variant, product: product) }
+      let!(:variant_override) { create(:variant_override, hub: distributor, variant: variant) }
+
+      subject { user }
+
+      let(:manage_actions) { [:admin, :index, :read, :update, :bulk_update, :bulk_reset] }
+
+      describe "when admin" do
+        let(:user) { create(:admin_user) }
+
+        it "should have permission" do
+          is_expected.to have_ability(manage_actions, for: variant_override)
+        end
+      end
+
+      describe "when user of the producer" do
+        let(:user) { producer.owner }
+
+        it "should not have permission" do
+          is_expected.not_to have_ability(manage_actions, for: variant_override)
+        end
+      end
+
+      describe "when user of the distributor" do
+        let(:user) { distributor.owner }
+
+        it "should not have permission" do
+          is_expected.not_to have_ability(manage_actions, for: variant_override)
+        end
+      end
+
+      describe "when user of the distributor which is also the producer" do
+        let(:user) { distributor.owner }
+        let!(:distributor) { create(:distributor_enterprise, is_primary_producer: true, sells: "any") }
+        let!(:producer) { distributor }
+
+        it "should have permission" do
+          is_expected.to have_ability(manage_actions, for: variant_override)
+        end
+      end
+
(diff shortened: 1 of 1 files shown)


mkllnk:
@kristinalim Moving this to "In Dev" as it doesn't seem ready for code review.

HugsDaniel:
All tests are green on both master and 0312bf93dde2bb9e976ccfaae954c991d6a0852f (which is the current version of Canada instance).

mkllnk:
@kristinalim What's the state of this PR? It got moved to test ready, because there is another fix for the associated issue. So I disconnected it. Is this ready for review?

kristinalim:
@mkllnk Yes, this can be tagged for Code Review. I wrote these tests to prove/understand the resulting permissions for variant overrides. Even after the bug fix of @luisramos0 in #2893, I think these tests would be handy to have in the code base.

I'll disconnect from the issue on Zenhub, update the PR title and description, and move to Code Review.

luisramos0:
hey @kristinalim I fetched this branch, rebased and run the tests. it's all green.
I also had a look at the code, all good.
I edited the description to make it ready to go.
Looks good for review and merge to me.

kristinalim:
Thanks for moving this forward @luisramos0! Somewhere along the way, I forgot this is already ready.

spec/models/spree/ability_spec.rb
```diff
@@ -514,5 +514,89 @@ module Spree
...
+
+        describe "when the enterprise is not a distributor" do
+          let!(:authorized_enterprise) do
+            create(:enterprise, sells: "none").tap do |record|
+              create(:enterprise_relationship, parent: producer, child: record, permissions_list: [:create_variant_overrides])
+            end
+          end
+
+          it "should not have permission" do
+            should_not have_ability(manage_actions, for: variant_override)
```
sauloperez: Could we use RSpec-3 syntax here? I know I sound boring and stubborn but sticking to this very old syntax makes it hard to stay-up-to-date. Let me know if you need help.

I'd like to automate this so I can avoid having to review syntax but I don't really know how to. Disabling `should` as RSpec's config allows is an all-or-nothing.

sauloperez: ok, as you guys want. It's just that if we don't all get used to not writing RSpec2 syntax it'll make it harder.

sigmundpetersen:
> What should we test?
>
> Build is green and all these new specs are green.

Merge?

kristinalim:
@sigmundpetersen Sorry, I'll push a minor change based on the review of @sauloperez first! :slightly_smiling_face: Thank you so much!

spec/models/spree/ability_spec.rb
kristinalim: I rebased the branch and also changed to RSpec 2 syntax.

@sauloperez Sorry about that! This was an old PR, before I found out about `is_expected` from you. :slightly_smiling_face: Odd but I insisted on using `should` when dealing with `subject` until that time.

@luisramos0 I changed only the syntax in the tests I added, so no extra problems when merging.

kristinalim:
This is good to go now.

spec/models/spree/ability_spec.rb
luisramos0: :+1: sorry, maybe I was too quick, my "no transpec until 2-0 merge" is not really valid for new tests...

PR #2845, opened 2018-10-08
Fix invalid variant creation

### LINKED ISSUE #2776: Able to end up with an invalid variant through web UI
Through the web UI alone, it is possible to end up with an invalid variant. This has big potential to mess up related code.

## Description

We normally presume `variant.save` and `variant.update_attributes` would succeed if we do not make changes the record that would make it invalid. When we have an already invalid variant, even if we don't change it and then attempt to save it, it would fail.

I haven't done more investigation on the consequences. I'm still just documenting at this point.

## Expected Behavior

It should not be possible to end up with an invalid variant in the database.

## Actual Behavior

I can create an invalid variant in the web UI by performing a series of actions.

## Steps to Reproduce

As enterprise manager:

1. Create a new product with "Items" as unit size.
2. Go to the "Bulk Edit Products" page.
3. Add a new variant for the product. Set unit to a non-numeric value, e.g. ~"1 med"~ "med".
4. Save the changes
5. Change the product "Unit" to weight, and save.

The variant at this point fails validation.

## Context

Some possibly related issues:

* #2769 
* #2764
* #2774 
* #2739 

## Severity

S2

sauloperez:
#### What? Why?

Closes #2776

It's been a bit hard to first understand and untangle `ProductSet`, which is the class responsible for the bulk update of products.

What I did is to check the presence of `variant.unit_value` when updating a product's `variant_unit`. Although there's a validation for that in `variant_decorator.rb`, it does not work when the product is not yet persisted :see_no_evil: (See my comment on the code).

Not only `ProductSet` should be improved, but also this unit related attributes spread between product and variant feels a bit messy. I'm not even sure validating a model checking other's attributes is even a good choice. (Shouldn't this go into a service? isn't too much for a model's responsibility). That's why I chose to:

* Duplicate this validation. Duplication is cheaper than a wrong abstraction.
* I fixed just the exact scenario we found with `variant_unit/unit_value`. I didn't implement a fix for any possible update to a product that may make its variants become invalid.

This will make the world a bit better than it was but it'll require a follow-up. We don't know if there are other possible scenarios.

#### What should we test?
Update products and variants through the "Bulk Edit Products" page, especially the case described in the issue above.

#### Release notes

Forbid changing a product's variant_unit when the variant has no unit_value.

Changelog Category: Fixed
 
#### Still missing :warning: 

- [x] Data migration to fix current cases.

### DIFF
--- a/app/models/spree/product_set.rb
+++ b/app/models/spree/product_set.rb
@@ -36,18 +36,17 @@ class Spree::ProductSet < ModelSet
 
   def update_product_only_attributes(product, attributes)
     variant_related_attrs = [:id, :variants_attributes, :master_attributes]
+    product_related_attrs = attributes.except(*variant_related_attrs)
 
-    if attributes.except(*variant_related_attrs).present?
-      product.assign_attributes(attributes.except(*variant_related_attrs))
+    return true if product_related_attrs.blank?
 
-      product.variants.each do |variant|
-        validate_presence_of_unit_value(product, variant)
-      end
+    product.assign_attributes(product_related_attrs)
 
-      product.save if errors.empty?
-    else
-      true
+    product.variants.each do |variant|
+      validate_presence_of_unit_value(product, variant)
     end
+
+    product.save if errors.empty?
   end
 
   def validate_presence_of_unit_value(product, variant)
@@ -58,19 +57,13 @@ class Spree::ProductSet < ModelSet
   end
 
   def update_product_variants(product, attributes)
-    if attributes[:variants_attributes]
-      update_variants_attributes(product, attributes[:variants_attributes])
-    else
-      true
-    end
+    return true unless attributes[:variants_attributes]
+    update_variants_attributes(product, attributes[:variants_attributes])
   end
 
   def update_product_master(product, attributes)
-    if attributes[:master_attributes]
-      update_variant(product, attributes[:master_attributes])
-    else
-      true
-    end
+    return true unless attributes[:master_attributes]
+    update_variant(product, attributes[:master_attributes])
   end
 
   def update_variants_attributes(product, variants_attributes)
--- a/app/models/spree/product_set.rb
+++ b/app/models/spree/product_set.rb
@@ -28,9 +28,35 @@ class Spree::ProductSet < ModelSet
     if found_model.nil?
       @klass.new(attributes).save unless @reject_if.andand.call(attributes)
     else
-      ( attributes.except(:id, :variants_attributes, :master_attributes).present? ? found_model.update_attributes(attributes.except(:id, :variants_attributes, :master_attributes)) : true) and
-      (attributes[:variants_attributes] ? update_variants_attributes(found_model, attributes[:variants_attributes]) : true ) and
-      (attributes[:master_attributes] ? update_variant(found_model, attributes[:master_attributes]) : true )
+      update_product_only_attributes(found_model, attributes) &&
+        update_product_variants(found_model, attributes) &&
+        update_product_master(found_model, attributes)
+    end
+  end
+
+  def update_product_only_attributes(product, attributes)
+    if attributes.except(:id, :variants_attributes, :master_attributes).present?
+      product.update_attributes(
+        attributes.except(:id, :variants_attributes, :master_attributes)
+      )
+    else
+      true
+    end
+  end
+
+  def update_product_variants(product, attributes)
+    if attributes[:variants_attributes]
+      update_variants_attributes(product, attributes[:variants_attributes])
+    else
+      true
+    end
+  end
+
+  def update_product_master(product, attributes)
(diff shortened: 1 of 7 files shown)


app/models/spree/variant_decorator.rb
```diff
@@ -18,11 +18,13 @@
   attr_accessible :unit_value, :unit_description, :images_attributes, :display_as, :display_name, :import_date
   accepts_nested_attributes_for :images
 
-  validates_presence_of :unit_value,
-    if: -> v { %w(weight volume).include? v.product.andand.variant_unit }
+  validates :unit_value, presence: true, if: -> (variant) {
+    %w(weight volume).include?(variant.product.andand.variant_unit)
+  }
```
sauloperez: This is the validation copied in `validate_presence_of_unit_value`. See below why this does not work for `ProductSet`:

```ruby
   41:       product.assign_attributes(attributes.except(*variant_related_attrs))
   42:
   43:       byebug
=> 44:       product.variants.each do |variant|
   45:         validate_presence_of_unit_value(product, variant)
   46:       end
   47:

...
(byebug) attributes
{:id=>618, :variant_unit=>"weight", :variant_unit_scale=>1}
(byebug) product.variant_unit
"weight"
(byebug) product.variants.first.product.variant_unit
"items"
```

We did update the value for the product object but this didn't affect the variant's value as they are different Ruby objects.

```
(byebug) product.object_id == product.variants.first.product.object_id
false
```

app/models/spree/product_set.rb
```diff
@@ -28,9 +28,35 @@ def update_attributes(attributes)
...
+
+  def update_product_only_attributes(product, attributes)
+    if attributes.except(:id, :variants_attributes, :master_attributes).present?
+      product.update_attributes(
+        attributes.except(:id, :variants_attributes, :master_attributes)
+      )
+    else
+      true
+    end
+  end
```
luisramos0: dont you prefer the early exit style for these if clauses?

  ```
return true unless attributes.except(:id, :variants_attributes, :master_attributes).present?
product.update_attributes(
   attributes.except(:id, :variants_attributes, :master_attributes)
)  
```

luisramos0:
@sauloperez I guess it's not wip anymore and you can remove the wip label, correct?

app/models/spree/product_set.rb
sauloperez: I do. I just want to touch them the least :trollface:. Will do

sauloperez:
Still missing a data migration as stated in the description and the issue itself.

sauloperez:
that's a good point. This is getting tricky but IMO you got to do that for the product and its variant when you having variants. 

I'll try to list the various concerns that are coming up and work on them separately and prioritizing properly. Agree?

sauloperez:
Moving it to test ready since the migration won't be manually tested by any tester.

sauloperez:

RachL:
@sauloperez quick question. I'm trying to add a variant but I get an error message saying that the unit cannot be blank:
capture d ecran de 2018-10-10 16-00-42

But I don't see the unit field.... Did I missed or misunderstood something somewhere? The product name is "test 2845"

sauloperez:
I've been looking at that variant and its product and although updating it did fail the first couple times, then it seems that by playing from bulk edit products I got it back to a valid state and I can't longer see that error message from the variant page. Can you reproduce that state? I can't.

There were no changes on `unit_description` so the only way I see we can reach that error is by having a variant without unit. AFAIK that attribute can only be set from the Bulk Edit Products page.

RachL:
@sauloperez yes I can reproduce if I follow the testing process of the issue. Meaning:

A. I create a new product (last test name : "new test")
B. I edit the product to go to the variant page and then I click on "new variant". This is what I get:
testvariant

If I go on an old product that already has a variant, i can add a new variant without error. Does this mean that I need first to edit the default variant that is created when I create a new product? I'm sorry if I'm waisting time by just not knowing how this is supposed to work :'( ping @myriamboure or @sstead if you have an idea of what I'm doing wrong.

sauloperez:
Thanks @RachL ! Yes, now I do manage to reproduce. I'll take a look.

sauloperez:
We'll give it a try by making `variant.unit_description` optional. This way we can test the actual issue which is not directly related to this attribute.

RachL:
Thanks @sstead it did work! Now in the context of the PR, tests do pass. But I guess we need to re-think a bit the variant creation process, it does seem very clear to me...

Testing notes :
https://docs.google.com/document/d/1WzCJahC-VO9vs3guBl7EQaNsJKxEdCsyenYnVYJCGuk/edit#

sauloperez:
Just to clarify, the issue @RachL found with `unit description` happens to be pre-existent. Nothing to do with this PR.

PR #2846, opened 2018-10-08
[Product Import] Rename enterprise fields and update template

### LINKED ISSUE #2481: [Product Import] Review csv template on landing page
## Description
We have review the UX of product import to avoid confusion, reduce the number of clics, and try to find some consistency.
Here we want to tackle the template part which was before included in issue #2455
The goal is to have csv template with all possible columns.

## Acceptance criteria

Templates for product listings and inventories are accurate. Especially, the template for inventory only has the necessary columns, some existing columns can be delete. Some column headers have been renamed for more clarity. See https://docs.google.com/document/d/1QDtKQe64enO5ZLnht9PxQ8bOlaMHCueJBMWSyW4zCng/edit# for headers in each template.

Matt-Yorkley:
#### What? Why?

Closes #2481

As part of #2481 it was decided to rename the names of the fields used in the CSV import. For importing to product list, the enterprise field has changed form "supplier" to "producer", and for inventory the fields "supplier" and "producer" are now "producer" and "distributor".

I've tried to make some of the naming more generic in the codebase, using "enterprise" where possible for lists such as "enterprise_ids" which will now be used either to check the "producer" for product list, or "distributor" for inventory.

#### What should we test?

Product Import generally, using the new names, and the downloadabe CSV templates.

#### Release notes

Renamed CSV fields in product import for clarity and updated CSV template.

Changelog Category: Changed

#### Documentation updates

Product Import guide will need to be updated to reflect the new field names.

### DIFF
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -166,6 +166,7 @@ module ProductImport
 
       product = Spree::Product.new
       product.assign_attributes(entry.attributes.except('id'))
+      product.supplier_id = entry.enterprise_id
       assign_defaults(product, entry)
 
       if product.save
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -29,7 +29,7 @@ module ProductImport
 
     def save_all(entries)
       entries.each do |entry|
-        if import_into_inventory?(entry)
+        if settings.importing_into_inventory?
           save_to_inventory(entry)
         else
           save_to_product_list(entry)
@@ -122,10 +122,6 @@ module ProductImport
       end
 
       @variants_updated += 1
-    end
-
-    def import_into_inventory?(entry)
-      entry.enterprise_id && settings.importing_into_inventory?
     end
 
     def save_new_inventory_item(entry)
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -5,8 +5,8 @@
 module ProductImport
   class EntryProcessor
     attr_reader :inventory_created, :inventory_updated, :products_created,
-                :variants_created, :variants_updated, :supplier_products,
-                :total_supplier_products, :products_reset_count
+                :variants_created, :variants_updated, :enterprise_products,
+                :total_enterprise_products, :products_reset_count
 
     def initialize(importer, validator, import_settings, spreadsheet_data, editable_enterprises, import_time, updated_ids)
       @importer = importer
@@ -23,8 +23,8 @@ module ProductImport
       @variants_created = 0
       @variants_updated = 0
       @products_reset_count = 0
-      @supplier_products = {}
-      @total_supplier_products = 0
+      @enterprise_products = {}
+      @total_enterprise_products = 0
     end
 
     def save_all(entries)
@@ -40,24 +40,24 @@ module ProductImport
     end
 
     def count_existing_items
-      @spreadsheet_data.suppliers_index.each do |_supplier_name, attrs|
-        supplier_id = attrs[:id]
-        next unless supplier_id && permission_by_id?(supplier_id)
+      @spreadsheet_data.enterprises_index.each do |_enterprise_name, attrs|
+        enterprise_id = attrs[:id]
+        next unless enterprise_id && permission_by_id?(enterprise_id)
 
         products_count =
           if settings.importing_into_inventory?
-            VariantOverride.where('variant_overrides.hub_id IN (?)', supplier_id).count
+            VariantOverride.where('variant_overrides.hub_id IN (?)', enterprise_id).count
           else
             Spree::Variant.
               not_deleted.
               not_master.
               joins(:product).
-              where('spree_products.supplier_id IN (?)', supplier_id).
+              where('spree_products.supplier_id IN (?)', enterprise_id).
               count
           end
 
-        @supplier_products[supplier_id] = products_count
-        @total_supplier_products += products_count
+        @enterprise_products[enterprise_id] = products_count
+        @total_enterprise_products += products_count
       end
     end
 
@@ -88,8 +88,8 @@ module ProductImport
       @products_created + @variants_created + @variants_updated + @inventory_created + @inventory_updated
     end
 
-    def permission_by_id?(supplier_id)
-      @editable_enterprises.value?(Integer(supplier_id))
+    def permission_by_id?(enterprise_id)
+      @editable_enterprises.value?(Integer(enterprise_id))
     end
 
     private
@@ -122,7 +122,7 @@ module ProductImport
     end
 
     def import_into_inventory?(entry)
-      entry.supplier_id && settings.importing_into_inventory?
+      entry.enterprise_id && settings.importing_into_inventory?
     end
 
     def save_new_inventory_item(entry)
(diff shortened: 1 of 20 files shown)


luisramos0:
Lots of "line is too long" errors on codeclimate. Are we really enforcing 80 chars per line? I think 80 is too short. I prefer 100.

Matt-Yorkley:
Some of these lines are starting to look really ugly with the 80 character limit...

app/models/product_import/entry_validator.rb
```diff
@@ -74,32 +74,48 @@ def enterprise_validation(entry)
       return if permissions_error entry
       return if primary_producer_error entry
 
-      entry.enterprise_id = @spreadsheet_data.enterprises_index[entry.enterprise][:id]
-      entry.public_send "#{enterprise_field}_id=", @spreadsheet_data.enterprises_index[entry.enterprise][:id]
+      entry.enterprise_id = \
```
luisramos0: is this slash needed?

Matt-Yorkley: Definitely not!

app/models/product_import/entry_processor.rb
```diff
@@ -135,7 +135,7 @@ def save_new_inventory_item(entry)
         @inventory_created += 1
         @updated_ids.push new_item.id
       else
-        @importer.errors.add("#{I18n.t('admin.product_import.model.line')} #{line_number}:", new_item.errors.full_messages)
+        @importer.errors.add("#{I18n.t('admin.product_import.model.line')} #{entry.line_number}:", new_item.errors.full_messages)
```
mkllnk: When duplicating the same code three or more times, I tend to extract it.

I would also use I18n for the string interpolation. Some languages may prefer something like "nth line:" instead of "Line n:".
```diff
-# with en.yml:
-#   line: Line
-"#{I18n.t('admin.product_import.model.line')} #{entry.line_number}:"

+# with en.yml:
+#   line: "Line %{number}:"
+I18n.t('admin.product_import.model.line', number: entry.line_number)
```

app/models/product_import/entry_processor.rb
```diff
@@ -166,6 +166,7 @@ def save_new_product(entry)
 
       product = Spree::Product.new
       product.assign_attributes(entry.attributes.except('id'))
+      product.supplier_id = entry.enterprise_id
```
mkllnk: The commit message doesn't explain why this change is needed. Is it necessary because of renaming supplier to enterprise? Or is it fixing a bug?

spec/features/admin/product_import_spec.rb
```diff
@@ -33,7 +33,7 @@
 
     it "validates entries and saves them if they are all valid and allows viewing new items in Bulk Products" do
       csv_data = CSV.generate do |csv|
-        csv << ["name", "supplier", "category", "on_hand", "price", "units", "unit_type"]
+        csv << ["name", "enterprise", "category", "on_hand", "price", "units", "unit_type"]
```
mkllnk: I think it's good practice to have code and spec changes in the same commit. It makes the use of git bisect possible and shows the impact of a code change.

app/models/product_import/spreadsheet_data.rb
```diff
@@ -33,10 +34,14 @@ def shipping_index
 
     private
 
+    def import_into_inventory?
```
mkllnk: We have some code duplication here. I'm wondering if the `entry` could know if it holds a producer or distributor.

app/models/product_import/entry_processor.rb
```diff
@@ -46,7 +46,10 @@ def count_existing_items
 
         products_count =
           if settings.importing_into_inventory?
-            VariantOverride.where('variant_overrides.hub_id IN (?)', enterprise_id).count
+            VariantOverride.where(
```
mkllnk: I thought `VariantOverride.where(hub_id: [enterprise_id])` would work as well. I just checked the class and this implementation is used for a scope already. You can write `VariantOverride.for_hubs([enterprise_id])`.

app/models/product_import/spreadsheet_data.rb
Matt-Yorkley: I've removed this method, as we already have it in the settings object. I'll call it from there instead, as it's now done that way in most places where that check is made.

mkllnk:
Great. What are your thoughts on my I18n comment?

sauloperez:
> Some of these lines are starting to look really ugly with the 80 character limit...

That was the default but we can increase it. What about 89? I believe that is the ancient standard-ish (my Vim highlights the 89th column).

Matt-Yorkley:
> I believe that is the ancient standard

@sauloperez What's the significance of these ancient standards? Did Moses come down the mountain bearing stone tablets with strictly 89 characters per line? :trollface:

app/models/product_import/entry_processor.rb
Matt-Yorkley: Yep, good thinking.

app/models/product_import/entry_processor.rb
Matt-Yorkley: It's because of the renaming, yes. The product object here needs the `supplier_id` to be set.

app/models/product_import/entry_processor.rb
Matt-Yorkley: I realised I've addressed this here and elsewhere in another PR. I'll remove this commit from here and address it (including your I18n suggestion) in https://github.com/openfoodfoundation/openfoodnetwork/pull/2842

sauloperez:
Well, you're right @Matt-Yorkley :trollface: It's the fear I share with @mkllnk of ending up with too much logic in a line (and that doesn't fit my small laptop screen).

sauloperez:

RachL:
@Matt-Yorkley I have troubles with shipping_category. In user guide it says that the column is not required, but I can't upload of file without it anymore. Plus, if I had it but leave blank as my 2 products have shipping category set to "none", the error stays the same....
Testing notes:
https://docs.google.com/document/d/1QYAXEXDJE8mlGLSNl30t9W7chhh_X6WQMt4tUgHMF4o/edit#

sauloperez:
Sorry guys I broke Katuma's staging while trying things out for `ofn-install`. I'll see if I can fix it soon. Meanwhile



Sorry for the inconvenience.

Matt-Yorkley:
@RachL I've re-tested this and these are my notes from 4 cases:

- Updating an existing product with no shipping_category column in CSV and shipping_category set to 'None' on the existing product: Success
- Updating an existing product with no shipping_category column in CSV and shipping_category set to 'Frozen' on the existing product: Error
- Updating existing product with empty shipping_category column in CSV and shipping_category set to 'None' on the existing product: Success
- Updating existing product with empty shipping_category column in CSV and shipping_category set to 'Frozen' on the existing product: Error

I think that's correct, no?

RachL:
@Matt-Yorkley Yes this is what I was expecting. however the two success cases you described didn't work on staging.

myriamboure:
Is it possible to restage so we can have some counter test? Maybe that was a problem with the staging as well, if we haven't the same results in the tests... If some dev can restage we will retest :-)

luisramos0:
staged: https://staging.openfoodfrance.org

RachL:
YES ! Now it is working good : no need for shipping_category in the file, the user guide is correct! Let's roll

sigmundpetersen:
So this should move on to Ready To Go @RachL ?

PR #2850, opened 2018-10-08
Password change logout

### LINKED ISSUE #2848: User logged out when changing password.
## Description

Follow-on from bug report #2840. When setting a new password under the account page, the user is logged out unexpectedly.

## Expected Behavior

User should stay logged in when setting a new password.

## Actual Behavior

User is unexpectedly logged out and bounced to the home page / login modal.

## Steps to Reproduce

1. Log in as a user
2. Go to /account, then Account Settings tab
3. Set a new password

## Context

Changing password.

## Severity

S4: annoying, but not critical.

## Possible Fix

Interestingly, the Spree::UsersController in the spree gem already has an apparent fix for this in the #update action, with the comment: `# this logic needed b/c devise wants to log us out after password changes`. Maybe it needs adjusting, or maybe we overrode it somewhere.

Matt-Yorkley:
#### What? Why?

Closes #2848

Adjusts a config setting so that users are not logged out when updating their password.

#### What should we test?

Users can change password via /account page without being logged out.

#### Release notes

Users will not be logged out whilst changing their password.

Changelog Category: Changed

### DIFF
--- a/spec/features/consumer/account/settings_spec.rb
+++ b/spec/features/consumer/account/settings_spec.rb
@@ -33,7 +33,6 @@ feature "Account Settings", js: true do
     end
 
     it "allows the user to change their password" do
-      allow(Spree::Auth::Config).to receive(:[]).with(:signout_after_password_change).and_return(false)
       initial_password = user.encrypted_password
 
       fill_in 'user_password', with: 'NewPassword'
--- a/spec/features/consumer/account/settings_spec.rb
+++ b/spec/features/consumer/account/settings_spec.rb
@@ -4,18 +4,22 @@ feature "Account Settings", js: true do
   include AuthenticationWorkflow
 
   describe "as a logged in user" do
-    let(:user) { create(:user, email: 'old@email.com') }
+    let(:user) do
+      create(:user,
+             email: 'old@email.com',
+             password: 'OriginalPassword',
+             password_confirmation: 'OriginalPassword')
+    end
 
     before do
       create(:mail_method)
       quick_login_as user
-    end
-
-    it "allows me to update my account details" do
       visit "/account"
-
       click_link I18n.t('spree.users.show.tabs.settings')
       expect(page).to have_content I18n.t('spree.users.form.account_settings')
+    end
+
+    it "allows the user to update their email address" do
       fill_in 'user_email', with: 'new@email.com'
 
       click_button I18n.t(:update)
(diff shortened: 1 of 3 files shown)


spec/features/consumer/account/settings_spec.rb
```diff
@@ -27,5 +31,16 @@
       click_link I18n.t('spree.users.show.tabs.settings')
       expect(page).to have_content I18n.t('spree.users.show.unconfirmed_email', unconfirmed_email: 'new@email.com')
     end
+
+    it "allows the user to change their password" do
+      fill_in 'user_password', with: 'NewPassword'
+      fill_in 'user_password_confirmation', with: 'NewPassword'
+
+      click_button I18n.t(:update)
+      expect(find(".alert-box.success").text.strip).to eq "#{I18n.t(:account_updated)} ×"
```
mkllnk: `find(".alert-box.success")` fails. It looks like the user is logged out after hitting Update.

spec/features/consumer/account/settings_spec.rb
```diff
@@ -27,5 +31,18 @@
       click_link I18n.t('spree.users.show.tabs.settings')
       expect(page).to have_content I18n.t('spree.users.show.unconfirmed_email', unconfirmed_email: 'new@email.com')
     end
+
+    it "allows the user to change their password" do
+      allow(Spree::Auth::Config).to receive(:[]).with(:signout_after_password_change).and_return(false)
```
Matt-Yorkley: @mkllnk I've added this to make the spec work, as the test environment doesn't pick up the value set in the spree initializer. Do you think this is ok, or should I try to set that config variable globally in the test environment instead? I'm wondering if it might interfere with other specs or changes to factories if it's set globally...

sauloperez: That's the way IMO. Changes to global state in tests => :boom:

mkllnk: I agree that changes to the global state are not good in tests in general. And I find this solution okay, but I see a problem in the fact that the logic is different for production and testing here. We have to add this line to the test case to make it behave the same way. That's basically code duplication in a very hidden place. Imagine we had to define the checkout state machine in the same spec that verifies that it's correct. It's not really testing the code used in production then.

Matt-Yorkley:
I agree @mkllnk that changing the settings for the production environment but not for tests doesn't seem like a great idea. I've switched to setting it in the global test config, and all the specs are green, so I guess it's ok. 

@sauloperez are you happy with that? https://github.com/openfoodfoundation/openfoodnetwork/pull/2850/commits/4760ebb80c908b70a9fd5a0d184a57a49d2b36b3

sauloperez:

myriamboure:
testing notes: https://docs.google.com/document/d/1gtzQWewyiLawJEyyH85xUSNQHNHcAitVvLPL4lu63qQ/edit?usp=sharing

@Matt-Yorkley i randomly did another test as change password was affected I tested the "forgot password" process and it seems that PR is breaking it, the email recovery email is not sent. See notes. I tested on current French production (on v1.21) and it is working so it feels like it can be connected to that PR... back to in dev then for now.

Matt-Yorkley:
@myriamboure I've just put this up on UK staging, and reset password emails are working fine with this PR. It'll leave it up if you want to re-test it.

Issue #2852, opened 2018-10-09
Remove parallel_tests gem


sauloperez:
## Description

`parallel_tests` gem was introduced by Rohan in 2015 before we used knapsack to parallelise spec runs. So we can remove it completely. See db47c01784df955cff00560f3c4b5454467025dd 2b3689fd933f99349f5284c61fc51ff1e875fb34 for details.

Stems from https://github.com/openfoodfoundation/openfoodnetwork/pull/2832

CoryWritesCode:
Has anyone picked this up yet for hacktoberfest?

sigmundpetersen:
Yes @CoryWritesCode this is now in Code Review with #2853 

The best way to find another issue is to install the Zenhub extension and pick one in the 'Dev Ready' column here https://github.com/openfoodfoundation/openfoodnetwork/labels/Hacktoberfest#boards?labels=hacktoberfest&activeFilters=labels&repos=6257856

sigmundpetersen:
BTW thanks a lot for your interest and welcome to Open Food Network. If you have any questions you can ask in our `dev` channel on [Slack](https://openfoodnetwork.org/slack-invite)

PR #2853, opened 2018-10-09
Remove parallel tests

mkllnk:
#### What? Why?

Replaces #2832

Dependabot suggested to update the gem parallel_tests which we don't use any more. I suggest to remove it.

This dependency was introduced by Rohan in 2015 before we used knapsack to parallelise spec runs. We were using our own CI server at the time, the days before Travis and Semaphore. This pull request is reverting db47c01784df955cff00560f3c4b5454467025dd and 2b3689fd933f99349f5284c61fc51ff1e875fb34

#### What should we test?

Not needed.

#### Release notes

Removed unused dependency `parallel_tests`.

Changelog Category: Removed


sauloperez:
Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/2852

Issue #1760, opened 2017-08-11
Instagram links not working 


sstead:
## Description
In Enterprise Setting > Social there is a field to insert an Instagram link. This link will appear on enterprise profiles. If the instagram page is https://www.instagram.com/happyfarm/ the link needs to be in the format "happyfarm". 
![image](https://user-images.githubusercontent.com/10843056/48240835-6d659a00-e428-11e8-9e10-c5c72dbca249.png)
Currently there is no pointer telling the user that this is the required format. As a results, there are multiple different ways that people include the link, most of which lead to broken links on their profile.

Link formats that will work: happyfarm
Link formats that do not work: /happyfarm, @happyfarm, https://www.instagram.com/happyfarm/ and www.instagram.com/happyfarm/

## Expected Behavior
Users can find out what format the link needs to be in.
If a user puts in a link in a format that's wrong, the link either corrects, or there is an error message.

## Actual Behavior
Users don't know what to enter. 
If they do enter an incorrect format there's no error message, so they don't know that the link will be broken. The link also doesn't autocorrect, so the instance can have many many profiles with broken links, which looks unprofessional.

## Steps to Reproduce
1. Go to any enterprise settings, and go to the "social" tab
2. See instagram field, no suggestion is shown.
3. Try to put in an incorrect format. You'll notice that the link is accepted even if wrong. Look the result from frontend on your profile, you'll see the link is broken.
4. Try to put Instagram link in the "happyfarm" only format, and look the result on the frontend. Link works.

## Context
This issue is leading to lots of broke links in the directory. 
It's also frustrating for users who are creating their profile because they don't know what format we want the link in.

## Severity
S4 (UX issue I would say as the software is working as expected but the user doesn't know what to do)

## Your Environment
PC, French production, Chrome

## Possible Fix
As at Noverber 9th 2018 the following fix has been decided. We tried to implement this fix via PRs https://github.com/openfoodfoundation/openfoodnetwork/pull/2859 and https://github.com/openfoodfoundation/openfoodnetwork/pull/2849, but as mentioned on 2849 there were some issues. 

Fix 1
- When a user first enters their social media links, if the instagram link is in any of the following four formats (@happyfarm, /happyfarm, https://www.instagram.com/happyfarm/ or www.instagram.com/happyfarm/) the form should convert the link to the correct shorthand format (happyfarm)
- If the system can't recognise the link, and can't convert it, there should be an error message when the user saves the form.
[...]

PR #2849, opened 2018-10-08
Fix for instagram links

albarnaz:
#### Problem

There is no suggestion in the Instagram empty field in enterprise profile setup menu. Some users copy-paste the full URL on their page. That doesn't work because what is expected is only their Instagram user pseudo.

image

#### FIX

Add a suggestion/model in the empty field that guides the user so that he enter the Instagram info at the expected format (as it is done for Twitter for instance).

Add migration to fix the wrong links already inputted, from "http://instagram.com/myUser" to "@myUser"

image

PR #2859, opened 2018-10-11
Instagram pattern

albarnaz:
#### Issue: #1760 

There is no suggestion in the Instagram empty field in enterprise profile setup menu. Some users copy-paste the full URL on their page. That doesn't work because what is expected is only their Instagram user pseudo.

image

#### FIX

Add an before save to check if the user enter the whole URL, and extract and store just the handle.

So, if the user enter "www.instagram.com/my_user", we store just "@my_user".

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -333,8 +333,7 @@ class Enterprise < ActiveRecord::Base
   end
 
   def instagram=(value)
-    regex = %r{\A(?:https?://)?(?:www\.)?instagram\.com/([a-zA-Z0-9._]{1,30})/?\z}
-    write_attribute(:instagram, value.try(:gsub, regex, '@\1'))
+    write_attribute(:instagram, value.try(:gsub, instagram_regex, '@\1'))
   end
 
   protected
@@ -344,6 +343,10 @@ class Enterprise < ActiveRecord::Base
   end
 
   private
+
+  def instagram_regex
+    %r{\A(?:https?://)?(?:www\.)?instagram\.com/([a-zA-Z0-9._]{1,30})/?\z}
+  end
 
   def name_is_unique
     dups = Enterprise.where(name: name)
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -80,6 +80,8 @@ class Enterprise < ActiveRecord::Base
   before_validation :set_unused_address_fields
   after_validation :geocode_address
 
+  before_save :check_instagram_pattern
+
   after_touch :touch_distributors
   after_create :set_default_contact
   after_create :relate_to_owners_enterprises
@@ -424,5 +426,10 @@ class Enterprise < ActiveRecord::Base
     Enterprise.distributing_products(self.supplied_products).
       where('enterprises.id != ?', self.id).
       each(&:touch)
+  end
+
+  def check_instagram_pattern
+    return if self.instagram.blank? || self.instagram.exclude?('www.instagram.com')
+    self.instagram = "@#{self.instagram.split('/').last}"
   end
 end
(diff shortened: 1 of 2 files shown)


sobrinho:
Maybe we can do something like:

```ruby
validates :instagram, format: /\A@[a-zA-Z0-9._]{1,30}\z/

def instagram=(value)
  write_attribute(:instagram, value.try!(:gsub, %r{\Ahttps?://(?:www.)?instagram.com/([a-zA-Z0-9._]{1,30})/?\z}, '@\1'))
end
```

That way we care about that manipulating on setter and we care about the format on the validation.

app/models/enterprise.rb
```diff
@@ -80,7 +80,7 @@ class Enterprise < ActiveRecord::Base
   before_validation :set_unused_address_fields
   after_validation :geocode_address
 
-  before_save :check_instagram_pattern
+  validates :instagram, format: /\A@[a-zA-Z0-9._]{1,30}\z/, :allow_blank => true
```
mkllnk: We try to keep the code style consistent. Can you change the last bit to `allow_blank: true`?

app/models/enterprise.rb
```diff
@@ -332,6 +332,10 @@ def can_invoice?
     abn.present?
   end
 
+  def instagram=(value)
+    write_attribute(:instagram, value.try(:gsub, %r{\Ahttps?://(?:www.)?instagram.com/([a-zA-Z0-9._]{1,30})/?\z}, '@\1'))
```
mkllnk: Dots `.` have a special meaning in regular expressions and match any character. We should consider that. I think this would be right:
```ruby
regex = %r{\Ahttps?://(?:www\.)?instagram\.com/([a-zA-Z0-9._]{1,30})/?\z}
write_attribute(:instagram, value.try(:gsub, regex, '@\1'))
```
I also extracted the expression into a variable to simplify it and comply with our line length restrictions.

spec/models/enterprise_spec.rb
```diff
@@ -149,10 +149,10 @@
       end
 
       context "prevent an wrong instagram link pattern" do
-        let!(:e) { create(:enterprise, instagram: 'www.instagram.com/my_user') }
+        let(:e) { build(:enterprise, instagram: 'instagram.com/my_user') }
 
-        it "expects the instagram attribute to be in the correct pattern" do
-          expect(e.instagram).to eq('@my_user')
+        it "expects the instagram attribute to not be valid" do
+          expect(e).to_not be_valid
```
mkllnk: Can we add more unit tests here to verify that the regular expressions work as intended?

In this case, it would be nice if it was valid and `instagram` could return `@my_user`. We probably need to modify the regex for this a bit more: `%r{\A(?:https?://)?(?:www\.)?instagram\.com/([a-zA-Z0-9._]{1,30})/?\z}`

We should test all common valid and invalid cases. Some ideas:

Valid input:

- `` empty string or `nil`
- `@user`
- `@my_www5.example`
- `http://instagram.com/user`
- `https://instagram.com/user/`
- `https://www.instagram.com/user`
- `instagram.com/user`

Invalid input:

- `@my-user`
- `https://facebook.com/user`
- `tagram.com/user`
- `https://instagram.com/user/preferences`

That should cover most common mistakes.

sigmundpetersen:
How is this related to #2849 ? Can the other be closed or are they complementary? @albarnaz

luisramos0:
they are complementary @sigmundpetersen

RachL:
Ready to go :) Testing notes: https://docs.google.com/document/d/1e8eWTX7YVYkkxYXpDZwXKiR_vUsRvpjrpqFvBElRfHI/edit#

mkllnk:
@RachL Sally found an issue with this pull request: https://github.com/openfoodfoundation/openfoodnetwork/pull/2849#issuecomment-437225515

We should have included in the test description that you need to check on a profile page if the Instagram link is actually working. :-D

RachL:
@mkllnk @sstead ouch my bad :/ But I've re-tested in our production and we don't get the error. Everything saves correctly, so I'm not sure if it was introduce by 2849. But anyway now it is reverted.

PR #2864, opened 2018-10-12
[Spree Upgrade] WIP Fix completed_order_with_fees tests in models/spree/order_spec and orders_controller spec

### LINKED ISSUE #2685: [Spree Upgrade][Models] Fix order_spec and orders_controller spec tests with shipment fees and taxes (3 broken tests)
Originated from #2009.
We have to find what's broken in the shipping fees calculation and fix them in spec/models/spree/order_spec.rb. It may be a code change, it may just be adapting tests.

Acc Criteria: order_spec tests passing.

luisramos0:
What? Why?
Closes #2685

What should we test?
Fixes 2 tests in order_spec and 1 in orders_controller_spec as described in #2685.

### DIFF
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -677,11 +677,13 @@ describe Spree::Order do
 
     context "removing line_items" do
       it "updates shipping and transaction fees" do
-        # Setting quantity of an item to zero
+        # Set quantity of an item to zero
         order.update_attributes(line_items_attributes: [{id: order.line_items.first.id, quantity: 0}])
+        order.shipment.save
+        order.update_payment_fees!
+        order.update!
 
         # Check if fees got updated
-        order.reload
         expect(order.adjustment_total).to eq expected_fees - shipping_fee - payment_fee
         expect(order.shipment.adjustment.included_tax).to eq 0.6
       end
@@ -697,7 +699,11 @@ describe Spree::Order do
         end
 
         it "does not attempt to update such adjustments" do
+          # Set quantity of an item to zero
           order.update_attributes(line_items_attributes: [{id: order.line_items.first.id, quantity: 0}])
+          order.shipment.save
+          order.update_payment_fees!
+          order.update!
 
           # Check if fees got updated
           order.reload
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -711,7 +711,16 @@ describe Spree::Order do
       let(:shipping_method) { create(:shipping_method, calculator: Spree::Calculator::FlatRate.new(preferred_amount: 0)) }
 
       it "updates shipping fees" do
+        # This is necessary to bypass memoised order's @shipment
+        #   Making the order object not cache the shipment as a @instance_variable but instead always use the shipments.first reference.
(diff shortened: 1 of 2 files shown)


app/models/spree/order_decorator.rb
```diff
@@ -226,6 +226,10 @@ def set_variant_attributes(variant, attributes)
     end
   end
 
+  def shipment
```
luisramos0: this is required for this test:
https://github.com/openfoodfoundation/openfoodnetwork/blob/2-0-stable/spec/models/spree/order_spec.rb#L711

app/models/spree/order_decorator.rb
```diff
@@ -226,6 +226,10 @@ def set_variant_attributes(variant, attributes)
...
+    shipments.first
+  end
+
```
luisramos0: this is required for this test:
https://github.com/openfoodfoundation/openfoodnetwork/blob/2-0-stable/spec/models/spree/order_spec.rb#L711

spec/models/spree/order_spec.rb
```diff
@@ -673,11 +673,13 @@
 
     context "removing line_items" do
       it "updates shipping and transaction fees" do
-        # Setting quantity of an item to zero
+        # Set quantity of an item to zero
         order.update_attributes(line_items_attributes: [{id: order.line_items.first.id, quantity: 0}])
+        order.shipment.save
+        order.update_payment_fees!
+        order.update!
```
luisramos0: I am not convinced order.update_attributes(line_items_attributes is a good way to change order attributes. But we must be using this somewhere in the code...

If we need this order.update_attributes(line_items_attributes, currently we have to do shipment.save, update_payment_fees and order.update! in the test (and probably later in the code) to fix the test (it's green), but it doesnt look like an acceptable solution.

What do you think?

app/models/spree/order_decorator.rb
mkllnk: I searched for uses of `order.shipment` in the OFN code base and could only find specs. Is that right? In that case I would suggest to fix the spec instead of adding to the decorator.

spec/models/spree/order_spec.rb
mkllnk: I'm wondering if `Spree::OrderUpdater` is what we want here. If not, then we probably want something similar to encapsulate the logic.

luisramos0:
I am not sure yet how to proceed here but I have added orders_controller_spec here (new commit) because the fix is exactly the same. Whatever we do for models orders_spec we have to do for order_controller_spec.

app/models/spree/order_decorator.rb
luisramos0: I have followed your advise. It's now done only in this spec and it works.

spec/models/spree/order_spec.rb
sauloperez: I'm a bit confused too. What the test tells me is that updating just removing a line item through an `update_attributes` is not enough to update the shipping and transaction fees :thinking:

luisramos0:
Rebased hoping for a magical fix, didn't happen 😭  we still need these calls to have these 3 specs green.
For now, I leave this unacceptable solution to the 3 specs as WIP.

luisramos0:
closing this PR. we have a better solution for #2685 in #3250

PR #2866, opened 2018-10-13
[Spree Upgrade] Rubocop fix

luisramos0:
Fixes 2-0-stable build issue with rubocop.
this is the result of "bundle update rubocop"

I dont understand the source of the problem but this is definitely the solution.


sauloperez:
Ready in https://github.com/openfoodfoundation/openfoodnetwork/pull/2867

PR #2869, opened 2018-10-15
[Bulk Invoice Printing] Bulk invoices

### LINKED ISSUE #2510: [Bulk Invoice Printing] Add checkboxes and "print invoices" bulk action button
### Description
**- As a:** enterprise user or a superadmin
**- On page:** admin/orders
**- I want to be able to:** Apply existing filters on order page, and choose single bulk action to generate in a new tab a PDF to print all corresponding invoices.
Note : as long as no order are selected, the print invoices button stays disable.

### Tasks

- [x] Add selectable checkboxes with Angular
- [x] Set up PDF invoice creation logic for multiple orders
- [x] Add bulk action button and endpoint
- [ ] Test large PDF resource use (potential timeouts?)

### Mockups:
image

image

image

### Acceptance Criteria
- All selected orders have an invoice in the PDF generated
- "Select all" limit is yet to define!

Matt-Yorkley:
#### What? Why?

Closes #2510

Adds bulk invoice printing to admin orders page. Orders can be selected with checkboxes, then a PDF containing an invoice for each selected order can be created.

#### What should we test?

Creating bulk invoice PDFs on the admin orders page.

#### Screenshots

screenshot from 2018-10-15 17-44-34

#### Release notes

Added bulk invoice printing to admin orders page.

Changelog Category: Added

### DIFF
--- a/app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
@@ -1,0 +1,30 @@
+angular.module("admin.orders").controller "bulkInvoiceCtrl", ($scope, $http, $timeout) ->
+  $scope.createBulkInvoice = ->
+    $scope.invoice_id = null
+    $scope.poll = 1
+    $scope.loading = true
+    $scope.message = null
+    $scope.error = null
+
+    $http.post('/admin/orders/invoices', {order_ids: $scope.selected_orders}).success (data) ->
+      $scope.invoice_id = data
+      $scope.pollBulkInvoice()
+
+  $scope.pollBulkInvoice = ->
+    $timeout($scope.nextPoll, 5000)
+
+  $scope.nextPoll = ->
+    $http.get('/admin/orders/invoices/'+$scope.invoice_id+'/poll').success (data) ->
+      $scope.loading = false
+      $scope.message = t('js.admin.orders.index.bulk_invoice_created')
+
+    .error (data) ->
+      $scope.poll++
+
+      if $scope.poll > 10
+        $scope.loading = false
+        $scope.error = t('js.admin.orders.index.bulk_invoice_failed')
+        return
+
+      $scope.pollBulkInvoice()
+
--- a/app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
@@ -21,7 +21,7 @@ angular.module("admin.orders").controller "bulkInvoiceCtrl", ($scope, $http, $ti
     .error (data) ->
       $scope.poll++
 
-      if $scope.poll > 10
+      if $scope.poll > 30
         $scope.loading = false
         $scope.error = t('js.admin.orders.index.bulk_invoice_failed')
         return
--- a/app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
@@ -5,13 +5,15 @@ angular.module("admin.orders").controller "bulkInvoiceCtrl", ($scope, $http, $ti
     $scope.loading = true
     $scope.message = null
     $scope.error = null
+    $scope.poll_wait = 5 # 5 Seconds between each check
+    $scope.poll_retries = 80 # Maximum checks before stopping
 
     $http.post('/admin/orders/invoices', {order_ids: $scope.selected_orders}).success (data) ->
       $scope.invoice_id = data
       $scope.pollBulkInvoice()
 
   $scope.pollBulkInvoice = ->
-    $timeout($scope.nextPoll, 5000)
+    $timeout($scope.nextPoll, $scope.poll_wait * 1000)
 
   $scope.nextPoll = ->
     $http.get('/admin/orders/invoices/'+$scope.invoice_id+'/poll').success (data) ->
@@ -21,7 +23,7 @@ angular.module("admin.orders").controller "bulkInvoiceCtrl", ($scope, $http, $ti
     .error (data) ->
       $scope.poll++
 
-      if $scope.poll > 30
+      if $scope.poll > $scope.poll_retries
         $scope.loading = false
         $scope.error = t('js.admin.orders.index.bulk_invoice_failed')
         return
--- a/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
@@ -73,9 +73,7 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
     $scope.fetchResults(newPage)
 
   $scope.bulkInvoice = ->
-    params = ''
-    angular.forEach $scope.selected_orders, (order_id) ->
-      params += 'order_ids[]='+order_id+'&'
-
-    $window.open('/admin/orders/invoices/show?'+params)
-    true
+    $http.post("/admin/orders/invoices", {order_ids: $scope.selected_orders}).success (data) ->
+      console.log(data)
+    .error (data) ->
+      console.log(data)
--- a/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
@@ -77,5 +77,5 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
     angular.forEach $scope.selected_orders, (order_id) ->
       params += 'order_ids[]='+order_id+'&'
 
-    $window.open('/admin/orders/bulk_invoice?'+params)
+    $window.open('/admin/orders/invoices/show?'+params)
     true
--- a/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
@@ -1,4 +1,4 @@
-angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor, Orders, SortOptions, $http, $window, $filter) ->
+angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor, Orders, SortOptions, $http, $timeout, $window, $filter) ->
   $scope.RequestMonitor = RequestMonitor
   $scope.pagination = Orders.pagination
   $scope.orders = Orders.all
@@ -12,6 +12,7 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
   $scope.checkboxes = {}
   $scope.selected = false
   $scope.select_all = false
+  $scope.poll = 0
 
   $scope.initialise = ->
     $scope.per_page = 15
@@ -72,8 +73,32 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
     $scope.page = newPage
     $scope.fetchResults(newPage)
 
-  $scope.bulkInvoice = ->
-    $http.post("/admin/orders/invoices", {order_ids: $scope.selected_orders}).success (data) ->
-      console.log(data)
-    .error (data) ->
-      console.log(data)
+  $scope.createBulkInvoice = ->
+    $scope.invoice_id = null
+    $scope.poll = 1
+    $scope.loading = true
+    $scope.message = null
+    $scope.error = null
+
+    $http.post('/admin/orders/invoices', {order_ids: $scope.selected_orders}).success (data) ->
+      $scope.invoice_id = data
+      $scope.pollBulkInvoice()
+
+  $scope.pollBulkInvoice = ->
+    console.log('Polling... ')
+
+    $timeout(->
+      $http.get('/admin/orders/invoices/'+$scope.invoice_id+'/poll').success (data) ->
+        $scope.loading = false
+        $scope.message = t('js.admin.orders.index.bulk_invoice_created')
+
+      .error (data) ->
+        $scope.poll++
+
+        if $scope.poll > 10
+          $scope.loading = false
+          $scope.error = t('js.admin.orders.index.bulk_invoice_failed')
+          return
+
+        $scope.pollBulkInvoice()
+    , 5000)
--- a/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
@@ -85,20 +85,20 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
       $scope.pollBulkInvoice()
 
   $scope.pollBulkInvoice = ->
-    console.log('Polling... ')
+    $timeout($scope.nextPoll, 5000)
 
-    $timeout(->
-      $http.get('/admin/orders/invoices/'+$scope.invoice_id+'/poll').success (data) ->
+  $scope.nextPoll = ->
+    $http.get('/admin/orders/invoices/'+$scope.invoice_id+'/poll').success (data) ->
+      $scope.loading = false
+      $scope.message = t('js.admin.orders.index.bulk_invoice_created')
+
+    .error (data) ->
+      $scope.poll++
+
+      if $scope.poll > 10
         $scope.loading = false
-        $scope.message = t('js.admin.orders.index.bulk_invoice_created')
+        $scope.error = t('js.admin.orders.index.bulk_invoice_failed')
+        return
 
-      .error (data) ->
-        $scope.poll++
+      $scope.pollBulkInvoice()
 
-        if $scope.poll > 10
-          $scope.loading = false
-          $scope.error = t('js.admin.orders.index.bulk_invoice_failed')
-          return
-
-        $scope.pollBulkInvoice()
-    , 5000)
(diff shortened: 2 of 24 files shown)


Matt-Yorkley:
I still need to ad some JS specs here, but any feedback is welcome in the meantime :)

app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
```diff
@@ -52,3 +52,11 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
   $scope.changePage = (newPage) ->
     $scope.page = newPage
     $scope.fetchResults(newPage)
+
+  $scope.bulkInvoice = ->
+    params = ''
+    angular.forEach $scope.selected, (selected, order_id) ->
+      params += 'order_ids[]='+order_id+'&' if selected
```
Matt-Yorkley: I was trying to find a nicer way to pass this data in a request at the same time as opening a new tab, but this was the only way that worked. I tried a few other options and some workarounds using `$http.post` and manipulating new windows/tabs, but they all either failed or triggered Firefox/Chrome to block them as suspicious popups :/

app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
```diff
@@ -52,3 +52,11 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
...
+
+    $window.open('/admin/orders/bulk_invoice?'+params)
+    true
```
Matt-Yorkley: This `true` would normally be unnecessary, but I encountered a weird Coffeescript bug that was throwing JS console errors, and this was the fix. 

Outlined here: https://github.com/angular/angular.js/issues/4853#issuecomment-28491586

app/controllers/spree/admin/orders_controller_decorator.rb
```diff
@@ -84,17 +85,35 @@ def resend
...
     respond_with(@order) { |format| format.html { redirect_to edit_admin_order_path(@order) } }
   end
 
+  def bulk_invoice
+    orders = params[:order_ids]
+
+    combined_pdf = CombinePDF.new
+
+    orders.each do |order_id|
+      @order = Spree::Order.find(order_id)
```
sauloperez: we should rather fetch all the orders in one shot to save a considerable amount of round-trips to the DB.

app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
Matt-Yorkley: I managed to make one with 15 orders, but yes I think I'll need to pass it to a background job, then possibly have the PDF saved to the disk, and maybe have an Angular function that periodically checks to see if the file exists yet, with a spinner / loading message...

app/views/spree/admin/orders/invoice.html.haml
```diff
@@ -30,13 +30,14 @@
         &nbsp;
       %td{ :align => "right" }
         %strong= "#{t('.to')}: #{@order.ship_address.full_name}"
-        - if @order.customer.code.present?
+        - if @order.andand.customer.andand.code.present?
```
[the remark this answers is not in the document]
Matt-Yorkley: Whilst testing I found a few orders that caused a fatal error here. After looking into it I found it wasn't caused by this PR, and those same orders threw a fatal error when making a regular invoice PDF.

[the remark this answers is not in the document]
Matt-Yorkley: Possibly due to legacy data?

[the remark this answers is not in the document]
sauloperez: 99% sure. It must be invalid data in DB :boom: . Could tell why do they fail and which ones? I'd like to have an issue to fix them.

sauloperez:
Still waiting for early review? Shall we move this back to In Dev?

app/views/spree/admin/orders/invoice.html.haml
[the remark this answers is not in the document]
Matt-Yorkley: Note: customer can be `nil` on an order, and code can be `nil` on a customer. We discussed implementing nil objects, but decided it might not be worth the effort for this one line.

Matt-Yorkley:
I've moved the processing to a background job, and added some asynchronous bits in the UI to check when it's finished and provide a link to the created file.

app/jobs/bulk_invoice_job.rb
```diff
@@ -0,0 +1,41 @@
...
+    orders.each do |order|
+      invoice = renderer.render_to_string pdf: "invoice-#{order.number}.pdf",
+                                          template: invoice_template,
+                                          formats: [:html], encoding: "UTF-8",
+                                          locals: { :@order => order }
+
+      pdf << CombinePDF.parse(invoice)
+    end
+
+    pdf.save "#{file_directory}/#{@filename}.pdf"
```
luisramos0: I think you need a invoice service with:
- all logic related to filenames, directories and filepaths used in both bulk_invoice_job and in invoices_controller
- invoice_template used in both bulk_invoice_job and orders_controller_decorator

app/assets/javascripts/admin/orders/orders.js.coffee
```diff
@@ -1 +1 @@
-angular.module("admin.orders", ['admin.indexUtils', 'ngResource'])
+angular.module("admin.orders", ['admin.indexUtils', 'ngResource', 'mm.foundation'])
```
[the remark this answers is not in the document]
Matt-Yorkley: `mm.foundation` includes our modal code. I'm pretty sure `ofn.admin` is a big module that would include a long list of things, including `mm.foundation`, so I think it's more efficient this way? I could be wrong though.

app/assets/javascripts/templates/admin/modals/bulk_invoice.html.haml
```diff
@@ -0,0 +1,16 @@
...
+  = t('js.admin.orders.index.bulk_invoices')
+
+%p.message{ng: {show: 'message'}}
+  {{message}}
+%p.error{ng: {show: 'error'}}
+  {{error}}
+
+%img.spinner{ src: "/assets/spinning-circles.svg", ng: { show: "loading" }}
+
+%a.button{target: '_blank', ng: {click: 'showBulkInvoice()', href: '/admin/orders/invoices/{{invoice_id}}', show: "!loading && !error"}}
```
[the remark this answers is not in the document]
Matt-Yorkley: I'll update the spacing. I have a feeling that rule is applied to Rails hashes but not HAML attributes...

app/controllers/spree/admin/invoices_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      def show
+        invoice_id = params[:id]
+
+        send_file(filepath(invoice_id), type: 'application/pdf', disposition: :inline)
+      end
+
+      def poll
+        invoice_id = params[:invoice_id]
```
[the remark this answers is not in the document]
Matt-Yorkley: That was how it came out in the generated routes. 

It looks like the solution involves waiting for Rails 4, or using an extra gem: https://stackoverflow.com/questions/6592038/change-the-name-of-the-id-parameter-in-routing-resources-for-rails

app/assets/javascripts/admin/orders/orders.js.coffee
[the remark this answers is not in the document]
luisramos0: if it works without this reference, Id remove it, it's better, less characters and complexity. but if it doesnt pick it up from ofn.admin, there's no other option.

app/controllers/spree/admin/invoices_controller.rb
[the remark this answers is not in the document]
luisramos0: I get it, it's because it considers "poll" to be another resource other than invoice and so the invoice id gets named "invoice_id" in the poll resource.

app/jobs/bulk_invoice_job.rb
Matt-Yorkley: Is the idea of services that they have one callable method? I think I'd need several callable methods to allow both `BulkInvoiceJob` and `InvoicesController` to access what they need. Is that ok?

app/assets/javascripts/admin/orders/orders.js.coffee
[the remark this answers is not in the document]
Matt-Yorkley: It has to be included here. It's not picked up from `ofn.admin`, because `ofn.admin` isn't included in `admin.orders`. Does that make sense?

app/jobs/bulk_invoice_job.rb
luisramos0: It's a very good question!

In general I'd say several methods is fine as long as you keep the service with a single responsibility, because these methods will be very simple I'd put them in an invoices_services.

I think making it mandatory for every service to have a "def call" and only one callable method is too strict and doesnt help, in this case creating 2 services is too much in my opinion.

what do you think @sauloperez? :-)

app/assets/javascripts/admin/orders/orders.js.coffee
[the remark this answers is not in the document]
luisramos0: ok, I was wondering how did I get the cookies modal to work without this but now I see that the cookies code is inside the darkswarm ng module that includes foundation :+1:

app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
```diff
@@ -45,3 +71,33 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, RequestMonitor,
...
+      .error (data) ->
+        $scope.poll++
+
+        if $scope.poll > 10
+          $scope.loading = false
+          $scope.error = t('js.admin.orders.index.bulk_invoice_failed')
+          return
+
+        $scope.pollBulkInvoice()
+    , 5000)
```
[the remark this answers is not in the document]
Matt-Yorkley: I've extracted the `$timeout` function.

I moved the invoices methods out of `OrdersCtrl` and into a new `BulkInvoiceCtrl` controller. The code in those methods feels very much like an angular controller to me: we're using `$scope`, attaching behaviour to the DOM, and returning user feedback into the DOM. What do you think?

app/jobs/bulk_invoice_job.rb
Matt-Yorkley: Ok, I've added a controversial service. :grimacing:

app/assets/javascripts/templates/admin/modals/bulk_invoice.html.haml
```diff
@@ -0,0 +1,16 @@
+%a.close-reveal-modal{ng: {click: '$close()'}}
```
luisramos0: sorry to be your rubocop but this needs spaces here :-D

app/jobs/bulk_invoice_job.rb
luisramos0: it's so much better!!!

app/controllers/spree/admin/invoices_controller.rb
```diff
@@ -0,0 +1,31 @@
+module Spree
+  module Admin
+    class InvoicesController < Spree::Admin::BaseController
+      respond_to :json
+
+      def create
+        invoice_service = BulkInvoiceService.new
+        invoice_service.create_bulk_invoice(params[:order_ids])
+
+        render text: invoice_service.id, status: :ok
```
sauloperez: Shouldn't it be `render json`?

app/services/bulk_invoice_service.rb
```diff
@@ -0,0 +1,56 @@
...
+                                          template: invoice_template,
+                                          formats: [:html], encoding: "UTF-8",
+                                          locals: { :@order => order }
+
+      pdf << CombinePDF.parse(invoice)
+    end
+
+    pdf.save "#{file_directory}/#{@id}.pdf"
+  end
+  handle_asynchronously :create_bulk_invoice
```
sauloperez: :see_no_evil: I find this too implicit and hidden. No one can tell from the consumer (the controller in this case) that this is executed in background and I think it's relevant, especially when debugging (and with background jobs, that happens often :trollface: )

I'd rather use a custom job class and call it as `Delayed::Job.enqueue(BulkInvoiceJob)`. Check the docs: https://github.com/collectiveidea/delayed_job/#custom-jobs

spec/controllers/spree/admin/invoices_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+describe Spree::Admin::InvoicesController, type: :controller do
+  let(:order) { create(:order_with_totals_and_distribution) }
+
+  before do
+    controller.stub spree_current_user: create(:admin_user)
```
sauloperez: `stub` is Spree2's age, I believe. Can we use `allow` here?

app/services/bulk_invoice_service.rb
Matt-Yorkley: I had a custom job at the beginning, but when I was refactoring it seemed really unnecessary, so I cut it. I can move it back to a custom job if you like, or maybe I could rename the method so it's clear it's a background job? `:start_pdf_job` or something similar?

luisramos0:
staged in https://staging.openfoodnetwork.org.au

Matt-Yorkley:
Yeah I had to add the loading popup when I did a second iteration. Once it's triggered it will build the PDF, it doesn't stop the process if the modal is closed. Maybe it needs some feedback and adjusting though? I added it at the last minute. 

I think if we stage on UK there should be a larger number of orders you can try, we have a lot more staging data.

RachL:
@Matt-Yorkley yes. The thing is: if the user closes the modal, there is no way he/she can access the result once the process is done, isn't it? So I see two options:
1. sending an email or displaying a message/ notification at the end of the process with a link to the pdf. Notification sounds better, but what if the user goes to another page? And if we introduce a message, does it mean we will leave it permanently on the page (or it vanishes at the end of the session)? If we choose this option, I guess we will have first to know which one is easier technically and then figure out the experience.
2. Not allowing the user to close the modal. But it's a shame if the job is already happening in the backround. Plus with a large number of invoices this might become very long for the user?
What do you think?

Matt-Yorkley:
@RachL I don't think sending an email is the way to go. I'm not sure how we would add a new notification to the page in terms of UI? 

Do we need some more inception on this last bit? I'm not clear on how to proceed with the next iteration.

Matt-Yorkley:
Yes, running the job in a background process means the page won't fail due to a timeout if it's longer than the server's timeout value (30 seconds by default, I think).

Blocking the user completely from closing the modal seems a bit problematic to me...

Matt-Yorkley:
@RachL I've staged this on UK staging now if you want to take another look. There should be plenty of data to work with.

myriamboure:
A- Can't we just open the PDF in a modal when the background job is done ? Even if the window was closed before, the end of the job triggers a new window to open ?
B- Else in order to release ASAP and be agile, I would just add a message in the window "if you close that window the PDF will not be generated" ;-) Easy no ? Even if it's not true because the background job continue, the result for the user is that she will have no PDF... so let'a jus warn her !
C- Then if we feel like it's not satisfying enough after users have started using it we can just, when user clicks on "print invoices" display a message saying that an email with a link will be sent as soon as the generation of the PDF is finished, and that it can take some minutes"

@RachL @Matt-Yorkley

daniellemoorhead:
@Matt-Yorkley @RachL have moved this to the Testing column as it's been staged and is awaiting testing. It was in the In Dev column. Move it back if this isn't correct :)

RachL:
I like your B proposal @myriamboure let's just add a sentence for now and keep the possibility to close the modal (if the user want to close because he is tired of waiting or did a mistake). 
Also I will open a discourse topic around notifications, to understand what's the best way to do it (if you read the conversation above Myriam you will see that emails for example where ruled out).

@Matt-Yorkley are you ok with this? I've tested on UK staging however, when I select more than 12 orders I have this message:
image
I think this need to be address before the first release :(

sauloperez:
B seems a good tradeoff as long as we work on a better solution while we release this 1st version. The current implementation allows us to have a better notification once the PDF is generated quite easily.

We basically keep asking the backend "is it already done? is it already done?..." aka. polling so we can show a message regardless of that modal.

Matt-Yorkley:
Is this ok?

screenshot from 2018-12-21 14-30-03
screenshot from 2018-12-21 14-30-14

RachL:
@Matt-Yorkley yes the message is perfect. However you remove the possibility to close the modal? Or we keep this message because the user can still close it by clicking outside the modal? Did you manage to create a PDF for more than 12 invoices?

Matt-Yorkley:
@RachL the modal will still close if the user clicks around the outside, but I removed the obvious button with the red X to lower the incentive to close it. What do you think?

RachL:
@Matt-Yorkley since the message is mentioning a closing action, I personally wouldn't mind if we keep it. Plus, once we will work on a notification, we will need a clear closing option.  But I guess it's not too hard to add it again in a 2nd iteration? If so, let's keep it this way and focus on checking if everything work (including when selecting a large number of invoices). Ok for you?

app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
```diff
@@ -21,7 +21,7 @@ angular.module("admin.orders").controller "bulkInvoiceCtrl", ($scope, $http, $ti
     .error (data) ->
       $scope.poll++
 
-      if $scope.poll > 10
+      if $scope.poll > 30
```
sauloperez: :warning: magic number. We should assign it to a constant with meaningul name and doc. Can be done in a separate pr if you prefer

RachL:
woooooouuuh this is looking good, I managed a file of 20 invoices easy peasy. I will stage this as well on UK staging to do a final test tomorrow morning with more orders. Bets are on :D
image

RachL:
@Matt-Yorkley on UK staging it fails when selecting 100 orders. Can you check if it is linked to UK staging or specific to the feature? But if it takes you more than an hour, I would advise that we release the feature without the possibility to show more than 50 orders per page, and we make another issue to deal with that. What do you think ?

image

Matt-Yorkley:
@RachL Okay, maybe :100: is too many! I set a maximum time for it to run at 2:30. So the options are: a) I can increase the maximum allowed time, or b) I can set the maximum invoices to 50.

Which do you prefer?

RachL:
@Matt-Yorkley okay so let's try to push it up to 4? If I remember correctly we won't have any issue with server timeout right?

Matt-Yorkley:
@RachL I've bumped it to 4 minutes. :+1:

Issue #2674, opened 2018-09-10
Make embedded_shopfronts_spec reliable


luisramos0:
## Description
The build is breaking very often on this spec.
We need to take a look.

This has been looked at on #2233 and #2468 and #2561 
Also related to moving to headless chrome: #2469 

## Expected Behavior
Reliable tests.

## Actual Behavior
Randomly broken build. Examples:

- https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/dependabot-bundler-uglifier-4-1-18/builds/3

## Context
We need a stable build without random broken specs.

PR #2876, opened 2018-10-16
Fix bug in subscriptions address controller where country states were not correctly returned

### LINKED ISSUE #2862: [Subscriptions] When setting up OC I can't select French states in address zone, but I can for Italy (on French instance)
## Description

French instance has three states/region bases setup for local addresses: France, Italy and Spain.
As a hub manager I try to setup a recurring order for a French customer, when setting up their address I can't proceed as I can't select any French state/region. 
But strangely if I choose "Italy" or "Spain" I can see the states/regions and choose.
It seems in subscription states change is not properly handled.

## Expected Behavior

I can setup a subscription for my customer.

## Actual Behavior

I can't as state for that customer cannot be setup.

## Steps to Reproduce

1. Go to "subscription", select hub, and "new subscription" on an instance that has multiple local states/regions configured.
2. When entering address, see when switching between states/regions and see.

## Animated Gif/Screenshot

capture du 2018-10-12 13-28-17
capture du 2018-10-12 13-28-42
capture du 2018-10-12 13-28-53

## Context

Reviewing subs scope and project with Rachel and testing some remaining issues.

## Severity

S2, this is a non-critical feature that is broken with no workaround and we have an active user of that feature in France.
We tried to first set up the customer address in customer list, but the states are still not visible in subscription and blocking setting up subs.

luisramos0:
the country id lookup up was not made exact (third parameter on the filter function) and states returned were incorrect. Added unit tests to cover different cases.

#### What? Why?

Closes #2862 

#### What should we test?
Subscriptions address state selection should now work correctly.
Country and state selectors on customer addresses new and edit screens were changed.
The code is now common across these two places.

#### Release notes
Changelog Category: Fixed
Fixed bug in subscriptions address country/state selection.

Not related.


spec/javascripts/unit/admin/subscriptions/controllers/address_controller_spec.js.coffee
```diff
@@ -0,0 +1,43 @@
...
+      expect(scope.statesFor(9)).toEqual []
+
+    it "returns states for country id in availableCountries with states", ->
+      expect(scope.statesFor(119)).toEqual states_in_spain
+
+    it "returns empty array for country id (11) in availableCountries but only as part of other country id (119)", ->
+      expect(scope.statesFor(11)).toEqual []
+
+    it "returns states for country id (19) in availableCountries with states even if other country ids contain the requested id (119)", ->
+      expect(scope.statesFor(19)).toEqual states_in_portugal
```
luisramos0: these last 2 tests were failing without the fix, both returning states_in_spain (adding true so the filter uses string exact match).

sigmundpetersen:
Build errors @luisramos0

sigmundpetersen:
Failing on a different one, the one reported in #2674

sigmundpetersen:
🚀

app/assets/javascripts/admin/subscriptions/controllers/address_controller.js.coffee
```diff
@@ -1,11 +1,21 @@
 angular.module("admin.subscriptions").controller "AddressController", ($scope, $filter, StatusMessage, availableCountries) ->
   $scope.countries = availableCountries
+
   $scope.statesFor = (country_id) ->
     return [] unless country_id
-    $filter('filter')(availableCountries, {id: country_id})[0].states
+    country = $filter('filter')(availableCountries, {id: country_id}, true)[0]
```
luisramos0: adding this ", true" is the fix to the issue, everything else in this PR is tests and refactoring.

mkllnk:

myriamboure:
@luisramos0 the issue is already there, there is a collaborative document where we listed the sample data we think we need, in a vision to enrich them step by step with new cases we need to test, to evolve it with agility. There is not yet sample data I think to test multistates instances, also I don't think multilingual, etc. So we definitely need to iterate on the file, but @mkllnk was disagreeing on the approach and @kirstenalarsen made some suggestion that probably it's time to reply to :-) https://community.openfoodnetwork.org/t/seed-data-development-provisioning-deployment/910/3

luisramos0:
ah, nice. will comment on discourse.

RachL:
@luisramos0 @myriamboure to move forward here, I need to create a bunch of state in 2 or 3 different countries right? Or something more?

luisramos0:
yes, the functional test is to have multiple countries with states and verify that both subscriptions and customers address selection work properly.

RachL:
@mkllnk @luisramos0 I've added states for Spain and France, but even countries do not appear in forms. Is there something else to change in order to have a several country instance?

mkllnk:
@RachL Yes, there is an important step that most people forget. You need to add the countries to the checkout zone. The currently active zone on staging is "Australia" under Settings / Zones. I just added Afghanistan to test that it's working. You can add Spain and France, then go to subscriptions and they should be available.

RachL:
@mkllnk can you be more specific? Like is it in general settings or enterprise settings? The only place I can think about that mention zones is here https://staging.openfoodnetwork.org.au/admin/zones but it is not meant to do what I search :D

myriamboure:
@RachL I just went to "zone" yes, checkout zone is where to people can checkout from, a shipping method is associated to a checkout zone. You will see the zone associated is Australia, so as @mkllnk suggested I added the two countries and now I can create a subs and see the relevant states, and Australian states are still as supposed, so you can finish testing and it seems ready to go :-)

RachL:
Ok thx! So for future me that would still have the question (I know I will forget this one): you need to EDIT the main country in the "zone" page. Here the url is then: https://staging.openfoodnetwork.org.au/admin/zones/3/edit

Testing notes: https://docs.google.com/document/d/1HpqeJ2I4l2z1NamFFi84ImKDqj0vhWFFMNJXK0Nee4k/edit#

PR #2888, opened 2018-10-18
[Spree Upgrade] Replace allow_backorders with variant.on_demand in the line_item_decorator

### LINKED ISSUE #2885: [Spree Upgrade] Replace usage of Config:allow_backorders with variant.on_demand in line_item_decorator
Same as #2803 but for app/models/spree/line_item_decorator.rb

luisramos0:
#### What? Why?

Closes #2885

Replaces calls to config::allow_backorders with variant.on_demand in line_item decorator.

#### What should we test?
- shopping/cart_spec with only 2 specs broken
In spec/models/spree/line_item_spec.rb, there's still an error but now related to the scoper and variant_overrides. To be addressed in another issue.

### DIFF
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -117,7 +117,7 @@ Spree::LineItem.class_eval do
   def sufficient_stock?
     return true if quantity == 0 # This line added
     scoper.scope(variant) # This line added
-    return true if Spree::Config[:allow_backorders]
+    return true if variant.on_demand
     if new_record? || !order.completed?
       variant.on_hand >= quantity
     else
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -103,16 +103,9 @@ feature "full-page cart", js: true do
         add_product_to_cart order, product_tax
       end
 
-      around do |example|
-        original = Spree::Config.allow_backorders
-        Spree::Config.allow_backorders = false
-        example.run
-        Spree::Config.allow_backorders = original
-      end
-
       it "prevents me from entering an invalid value" do
         # Given we have 2 on hand, and we've loaded the page after that fact
-        variant.update_attributes! on_hand: 2
+        variant.update_attributes!({ on_hand: 2, on_demand: false })
         visit spree.cart_path
 
         accept_alert 'Insufficient stock available, only 2 remaining' do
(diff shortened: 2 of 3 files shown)


app/models/spree/line_item_decorator.rb
```diff
@@ -117,7 +117,7 @@ def unit_value
   def sufficient_stock?
     return true if quantity == 0 # This line added
     scoper.scope(variant) # This line added
-    return true if Spree::Config[:allow_backorders]
+    return true if variant.on_demand
```
mkllnk: I'm not sure we need to handle `on_demand` here since that was not done in Spree. Spree just looked at the global config here, not the the specific on_demand value. I guess on_demand is checked somewhere else. But it doesn't hurt to double-check.

It's a bit annoying that we are so far from Spree now that we can't use the StockQuantifier. We need a solution for VariantOverrides to get rid of these decorators.

luisramos0: tbh, I am not sure what's the relation between stock_items and VariantOverrides (StockQuantifier). I need to look into that. This decorator needs to be looked at for sure. Probably a new issue. I'll investigate and create one if needed.

mkllnk: Spree moved all stock logic from the variants into a service called StockQuantifier. But unfortunately that service needs to query the database (stock items) to do its job. VariantOverrides contain their own stock data and don't use stock items. Therefore the StockQuantifier can't be used for it.

So before we could just feed different data (count_on_hand) into Spree's stock logic. Now we have to implement our own stock logic (a copy of Spree's old logic at the moment).

luisramos0: Is this a issue we have to create in the spree upgrade @mkllnk ?

mkllnk: No, I don't think we need to change anything unless we discover that we have to. The current code could just work.

sauloperez: > I'm not sure we need to handle `on_demand` here since that was not done in Spree. Spree just looked at the global config here, not the specific on_demand value. I guess on_demand is checked somewhere else. But it doesn't hurt to double-check.
> 
> It's a bit annoying that we are so far from Spree now that we can't use the StockQuantifier. We need a solution for VariantOverrides to get rid of these decorators.

I share @mkllnk here. Let's move it forward but keep an eye on redundant checks. I think soon we'll be familiar enough with this logic to confidently decide.

PR #2912, opened 2018-10-23
Admin orders refactor

### LINKED ISSUE #2911: Refactor Admin Orders controller
Follow on from working on #2908

## Description

Quite a few changes have gone through for Admin Orders, and the index action especially needs a refactor and tidy up (it's a mess of old and unnecessary code).

This is part of the Bulk Invoice Printing epic.

Matt-Yorkley:
#### What? Why?

Closes #2911

Following up on a refactor prompted by #2908

Tidying up the Admin::OrdersController and related endpoints after the recent changes have made lots of the legacy code there redundant. :fire:

#### What should we test?

Admin orders and BOM pages should work as before.

#### Release notes

Refactor admin orders controller and related code.

Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -23,17 +23,7 @@ Spree::Admin::OrdersController.class_eval do
   respond_to :html, :json
 
   def index
-    @results = SearchOrders.new(params, spree_current_user) if json_request?
-
-    respond_to do |format|
-      format.html
-      format.json do
-        render json: {
-          orders: serialized_orders,
-          pagination: @results.pagination_data
-        }
-      end
-    end
+    # Moved to api. Overriding the action so we only render the page template.
   end
 
   # Overwrite to use confirm_email_for_customer instead of confirm_email.
@@ -95,9 +85,5 @@ Spree::Admin::OrdersController.class_eval do
     unless @order.distribution_set?
       render 'set_distribution', locals: { order: @order }
     end
-  end
-
-  def serialized_orders
-    ActiveModel::ArraySerializer.new(@results.orders, each_serializer: Api::Admin::OrderSerializer)
   end
 end
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -23,7 +23,8 @@ Spree::Admin::OrdersController.class_eval do
   respond_to :html, :json
 
   def index
-    # Moved to api. Overriding the action so we only render the page template.
+    # Overriding the action so we only render the page template. An angular request
+    # within the page then fetches the data it needs from Api::OrdersController
   end
 
   # Overwrite to use confirm_email_for_customer instead of confirm_email.
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -20,50 +20,17 @@ Spree::Admin::OrdersController.class_eval do
 
   before_filter :require_distributor_abn, only: :invoice
 
-
   respond_to :html, :json
 
-  # Mostly the original Spree method, tweaked to allow us to ransack with completed_at in a sane way
   def index
-    params[:q] ||= {}
-    params[:q][:completed_at_not_null] ||= '1' if Spree::Config[:show_only_complete_orders_by_default]
-    @show_only_completed = params[:q][:completed_at_not_null].present?
-    params[:q][:s] ||= @show_only_completed ? 'completed_at desc' : 'created_at desc'
+    @results = SearchOrders.new(params, spree_current_user) if json_request?
 
-    # As date params are deleted if @show_only_completed, store
-    # the original date so we can restore them into the params
-    # after the search
-    created_at_gt = params[:q][:created_at_gt]
-    created_at_lt = params[:q][:created_at_lt]
-
-    params[:q].delete(:inventory_units_shipment_id_null) if params[:q][:inventory_units_shipment_id_null] == "0"
-
-    if !params[:q][:created_at_gt].blank?
-      params[:q][:created_at_gt] = Time.zone.parse(params[:q][:created_at_gt]).beginning_of_day rescue ""
-    end
-
-    if !params[:q][:created_at_lt].blank?
-      params[:q][:created_at_lt] = Time.zone.parse(params[:q][:created_at_lt]).end_of_day rescue ""
-    end
-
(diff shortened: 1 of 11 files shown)


app/views/spree/admin/orders/_filters.html.haml
```diff
@@ -1,38 +1,40 @@
 %div{"data-hook" => "admin_orders_index_search"}
-  = search_form_for [:admin, @search], html: { name: "orders_form", "ng-submit" => "fetchResults()"} do |f|
```
sauloperez: Is this `search_form_for` a Rails thing? I'm trying to understand what's the change that we did in this partial.

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+        end
+
+        it "retrieves a list of orders with appropriate attributes,
+            including line items with appropriate attributes" do
+
+          returns_orders(json_response)
+        end
+
+        it "formats completed_at to 'yyyy-mm-dd hh:mm'" do
+          completed_dates = json_response['orders'].map{ |order| order['completed_at'] }
```
sauloperez: We have [json_spec](https://github.com/collectiveidea/json_spec) in the Gemfile, which might be of help in this spec file.

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+        end
+      end
+    end
+
+    private
+
+    def returns_orders(response)
+      keys = response['orders'].first.keys.map(&:to_sym)
+      expect(order_attributes.all?{ |attr| keys.include? attr }).to be_truthy
+    end
```
sauloperez: I'm not sure this is worth the extra complexity. Isn't enough to check for an `id` field in `response['orders']`?

spec/controllers/spree/admin/orders_controller_spec.rb
```diff
@@ -30,112 +30,24 @@
...
-          controller.stub spree_current_user: coordinator.owner
-          spree_get :index, :format => :json
-        end
-
-        it "retrieves a list of orders" do
-          keys = json_response['orders'].first.keys.map{ |key| key.to_sym }
-          order_attributes.all?{ |attr| keys.include? attr }.should == true
-        end
+      before do
+        controller.stub spree_current_user: order.distributor.owner
```
sauloperez: let's use `allow` here, now that you are at it.

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+    def order_attributes
+      [
+        :id, :number, :full_name, :email, :phone, :completed_at, :display_total,
+        :show_path, :edit_path, :state, :payment_state, :shipment_state,
+        :payments_path, :shipments_path, :ship_path, :ready_to_ship, :created_at,
+        :distributor_name, :special_instructions, :payment_capture_path
+      ]
+    end
+
+    def pagination_data
```
sauloperez: Why not simply inline this hash into the single example that's using it? IMO indirection is more of a downside when reading tests.

spec/services/search_orders_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def pagination_data
+    {
+      results: 3,
+      pages: 1,
+      page: 1,
+      per_page: 15
+    }
+  end
```
sauloperez: Same as I suggested above about `pagination_data`

app/controllers/spree/admin/orders_controller_decorator.rb
```diff
@@ -20,58 +20,10 @@
...
-          pagination: {
-            results: @orders.total_count,
-            pages: @orders.num_pages.to_i,
-            page: params[:page].to_i,
-            per_page: params[:per_page].to_i
-          }
-        }
-      end
-    end
+    # Moved to api. Overriding the action so we only render the page template.
```
sauloperez: Ok, so then we render the template and then we request the orders with an Ajax call to the API?

app/views/spree/admin/orders/_filters.html.haml
Matt-Yorkley: `search_form_for` is an addition from the Ransack gem used for creating search forms. In order to use this alternate form helper here we have to construct a ransack search object in the controller and pass it to the view. So it's an additional dependency that adds more mess in the controller action, but actually it's no longer needed as we are passing the form data to Angular instead.

Matt-Yorkley: So changing this to the basic `form_for` helped to tidy up the controller.

spec/controllers/api/orders_controller_spec.rb
Matt-Yorkley: This was part of migrating tests from one area to another. Checking all the attributes are there means we know the serializer is returning what it should do, but it also makes this test more coupled with the code. Should I cut it?

app/controllers/spree/admin/orders_controller_decorator.rb
Matt-Yorkley: Yes. We just render the template and the Angular in the page makes it's request.

mkllnk: Future readers may find this confusing without more context. What moved to the api? What is in the page template? Maybe we could write a bit more to explain:
> This action renders the page template without orders. The template contains Javascript calls to fetch the order data from the API (`Api::OrderController`).

spec/controllers/api/orders_controller_spec.rb
mkllnk: I'm okay with leaving it as the work is done already.

sauloperez: It's just the I'm always very concerned about performance, complexity and changeability of the test suite as it tends to grow a lot. But your call, it's not a big deal.

app/controllers/spree/admin/orders_controller_decorator.rb
```diff
@@ -23,7 +23,8 @@
   respond_to :html, :json
 
   def index
-    # Moved to api. Overriding the action so we only render the page template.
+    # Overriding the action so we only render the page template. An angular request
+    # within the page then fetches the data it needs from Api::OrdersController
```
sauloperez: ❤️

sstead:
**Testing notes**
- fine
- 2 pre-existing bugs (orders listed multiple times in orders list, and delete not working in BOM)

https://docs.google.com/document/d/1jGxLzcpv8ap61rlJ_wfdNIqqUbFgmXzCu4G2-LCdfOg/edit#

Issue #2626, opened 2018-08-31
Avoid missing i18n translations


kristinalim:
## Description

There doesn't seem to be an easy way right now to make sure that translation keys we use indeed have corresponding entries in the main language file.

There's a chance to miss it if a developer introduces a new translation key but forgets to add it to the language file.

And if a translation is missing, the non-underscored version of the key is used, so a missing "new_feature" appears as "New Feature" (wrapped in `span.translation_missing`). This translation often still makes sense for the `development` language `en`, so you would think everything is okay when manually testing in the browser.

These are small enough changes that can help us avoid these missing translations:

* For translations used in the Ruby code, in automated tests (only `test` environment), raise error when encountering a missing translation key. We have good test coverage, so I think this would be very helpful.
* For translations used in the JS, warn in the browser console if a translation is missing.
* Make it more obvious in the UI if a certain text is only from the non-underscored version of the translation key. As mentioned, these are wrapped in `span.translation_missing` so can easily be styled.

## Expected Behavior

Warnings when using a translation key that is missing from the language file

## Actual Behavior

No changes in actual behaviour if translations are okay

## Steps to Reproduce

1. Add `t("missing_key")` to a page, and check page. The page would not complain, and render this as "Missing Key".
2. The following test would not complain:
```ruby
require 'spec_helper'

describe "Missing translation" do
  it "raises error when a translation is missing" do
    I18n.t("missing_key")
  end
end
```

## Context

Lack of confidence when adding/updating translation keys

kristinalim:
When I submitted this issue, I was only thinking about how this would affect application code, but it turns out this is also relevant in the decision of whether we should use `I18n.t` in specs or not. There is a discussion on this [here](https://community.openfoodnetwork.org/t/convention-decision-to-i18n-or-not-to-i18n-in-spec-assertions/1432/1).

About Task 1, @sauloperez recommended [this article](https://robots.thoughtbot.com/better-tests-through-internationalization) which has code for this. The code sets this up not just for the test environment but also for development. Should we do the same?

If we also do Task 1 for development, Task 3 will no longer be necessary because then `span.translation_missing` would never be generated there.

mkllnk:
I ran a CI run with exceptions on missing translations. https://github.com/openfoodfoundation/openfoodnetwork/commit/1c928474d1eebf98a8d9b59ea64766b26a5e136c

Here are the results:

```
Failures:

[...]

PR #2913, opened 2018-10-23
Add password confirmation, token and expired keys

### LINKED ISSUE #2905: Untranslated error message in reset password process
## Description

When asking new password, one string is untranslated (second field) and the error message if token expired is untranslated as well.

capture du 2018-10-22 16-41-47

## Severity

This is front end so more annoying... priority S3 I would say ;-)

HugsDaniel:
#### What? Why?

Closes #2905 

Add password confirmation, token and expired keys for devise i18n.

#### What should we test?

Translations are made with devise i18n keys.

Changelog Category: Added

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -248,6 +248,9 @@ en:
   pick_up: Pick up
   ok: Ok
   copy: Copy
+  password_confirmation: Password Confirmation
+  reset_password_token: Reset password token
+  expired: has expired, please request a new one
 
   actions:
     create_and_add_another: "Create and Add Another"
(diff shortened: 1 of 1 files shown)


sigmundpetersen:
Stage on FR @luisramos0 ?

luisramos0:
stage https://staging.openfoodfrance.org/

myriamboure:
It might not be super easy to reproduce @RachL to see the key, I have no idea what is the duration of the password reset expiration token? Maybe some dev can tell us? @Matt-Yorkley ? @mkllnk ? So we can plan the test to get it and see if the text is still ok.

RachL:
Don't worry that's why I've started it a Friday night @myriamboure so that the link would be expired on Monday ;)

"Password confirmation" is not translated. Is it just not translated in the french file but the string is there @HugsDaniel ?

image

mkllnk:
I also found a few untranslated text pieces with an automated test: https://github.com/openfoodfoundation/openfoodnetwork/issues/2626#issuecomment-433291504
I think it would be really good to get all these untranslated strings done quickly so that we can check for them in automated tests. That should avoid most of these issues.

HugsDaniel:
@RachL the key for `password_confirmation` is there (see `en.yml`) so the translation must be missing for french and other languages.

I can take care of the other missing translations in this PR, @mkllnk what do you think ? Or in a new PR ?

RachL:
@HugsDaniel ok then I propose we move this one forward. Please do another PR for the other missing translations :) it will be easier to follow.

mkllnk:
All good! My comment was about more work to be done in other PRs. :+1:

PR #2914, opened 2018-10-23
Fix multi lingual issue in cart and checkout: the user language selection was not being used

### LINKED ISSUE #2491: Selected LOCALE is not maintained with navigation
## Description
If you add a product to the cart on the Portuguese site, it switches to Spanish as you visit the cart.

- go to shop in Portuguese https://alpha.katuma.org/quinta-de-silvares-vizela/shop?locale=pt
- select an order cycle
- add a quantity
- click Edit Cart "Editar carrinho"
- the destination page is a page in spanish

## Expected Behavior
Should keep user language: portuguese in this example.

## Actual Behavior
Switches to Spanish.

## Steps to Reproduce
See description above.

## Context
Portugese version of Katuma not usable like this.

## Severity
S4

## Your Environment
* Version used: katuma live
* Operating System and version (desktop or mobile): OSX

luisramos0:
#### What? Why?

Closes #2491 

The function set_locale in the controllers gets the locale variable into the locale cookie, this was only done in:
 - Spree::Admin::BaseController (all OFN backoffice controllers extend this one directly or through ResourceController)
- BaseController (most our frontend controllers extend this one) 

But in the frontoffice, we were missing this set_locale in all controllers that do not extend BaseController, basically, controllers still relying on Spree code:
- controllers/CheckoutController (extends Spree::CheckoutController)
- controllers/spree/Spree::OrdersController.class_eval
- controllers/spree/Spree::UsersController

UsersController already had the set_locale call itself and the other two CheckoutController and OrdersController both extend StoreController, so, that was the destination of the new call to set_locale. See the code.

I have checked backoffice code and some pages and I couldnt find any similar problem there.

#### What should we test?
Cart and checkout process with a non-default language.

#### Release notes
Changelog Category: Fixed
Bug in non-default language: user language selection is no longer jumping to default selection on the cart/checkout process. It keeps the user language through the process.

not related.

### DIFF
--- a/app/controllers/spree/store_controller_decorator.rb
+++ b/app/controllers/spree/store_controller_decorator.rb
@@ -1,6 +1,9 @@
 class Spree::StoreController
   layout 'darkswarm'
 
+  include I18nHelper
+  before_filter :set_locale
+
   def unauthorized
     render 'shared/unauthorized', :status => 401
   end
--- a/spec/features/consumer/multilingual_spec.rb
+++ b/spec/features/consumer/multilingual_spec.rb
@@ -29,17 +29,15 @@ feature 'Multilingual', js: true do
       visit root_path(locale: 'es')
       expect(get_i18n_locale).to eq 'es'
       expect(get_i18n_translation('label_shops')).to eq 'Tiendas'
-      expect(page.driver.browser.cookies['locale'].value).to eq 'es'
+      expect_menu_and_cookie_in_es
       expect(page).to have_content '¿Estás interesada en entrar en Open Food Network?'
-      expect(page).to have_content 'TIENDAS'
 
       # it is not in the list of available of available_locales
       visit root_path(locale: 'it')
       expect(get_i18n_locale).to eq 'es'
       expect(get_i18n_translation('label_shops')).to eq 'Tiendas'
-      expect(page.driver.browser.cookies['locale'].value).to eq 'es'
+      expect_menu_and_cookie_in_es
       expect(page).to have_content '¿Estás interesada en entrar en Open Food Network?'
-      expect(page).to have_content 'TIENDAS'
     end
 
     context 'with a product in the cart' do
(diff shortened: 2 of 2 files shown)


app/controllers/spree/store_controller_decorator.rb
```diff
@@ -1,6 +1,9 @@
 class Spree::StoreController
   layout 'darkswarm'
 
+  include I18nHelper
+  before_filter :set_locale
```
sauloperez: Would you mind adding test for this in `spec/controllers/spree/store_controller_spec.rb`?

luisramos0: When I think that the it will more expensive to add the test than the value it provides I skip it.
My initial intuition about this test was right.
But well, if you ask. I just spent an hour trying to write this test without success.
Maybe you can help, this is my best try:

```
require 'spec_helper'

describe Spree::StoreController, type: :controller do

  include I18nHelper
  def params
    {locale: 'xp'}
  end

  def spree_current_user
  end

  controller(Spree::StoreController) do
    before_filter :unauthorized
    def index
      render text: ""
    end
  end

  it "redirects to home when unauthorized" do
    get :index
    response.should render_template("shared/unauthorized", layout: 'darkswarm')
  end

  it "calls set_locale before index" do
    expect(controller).to receive(set_locale)

    get :index
  end
end
```

luisramos0:
I have fixed the certificate on staging.katuma and the broken menu. It's all working now.

I've also tested the PR and it's looking good.

@RachL @myriamboure could you guys give this one a look so it gets included in todays release?
Thanks!

RachL:
@luisramos0 sorry I've missed this comment! I'm on it. Don't hesitate to ping me on slack next time ;)

RachL:
I think we are good to go! However I've did 2 orders one in Portuguese, one in Spanish and I've didn't received an email confirmation for the one in Portuguese. Maybe just email configuration missing on staging @sauloperez ?

Portugues:
image

Castellano:
image

sauloperez:
>  Maybe just Portuguese email configuration missing on staging @sauloperez ?

There shouldn't be any difference so might be something else. Can you create an issue to investigate it after reproducing it again?

Issue #2882, opened 2018-10-17
[Spree Upgrade] MailMethod is back from master into 2-0-stable and needs to be converted to spree 2 MailConfiguration


luisramos0:
## Description
There are around 20 test broken in the 2-0-stable build with error:
```
Failure/Error: create(:mail_method)       
       ArgumentError:
         Factory not registered: mail_method
```

Change in master that originated this:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2725/files

This is the type of solution already used as part of #1957:
https://github.com/openfoodfoundation/openfoodnetwork/commit/b5038fa5dc8483c17cabd6883770c6a0a9285848#diff-d0a5855e5f1cb6163b01e2c22d95d787R629

## Expected Behavior
Specs do not reference mail_method and emails are sent.

## Actual Behavior
specs broken.

## Steps to Reproduce
See for example user_spec

luisramos0:
Hey @mkllnk #2923  has been merged to 2-0-stable. So, it's ready to be fixed on the 2-0 branch.

setup_email is breaking ~50 tests here: https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/2-0-stable/builds/93

PR #2923, opened 2018-10-25
Extract email setup in specs for easy upgrade

mkllnk:
#### What? Why?

First part of #2882.

The way we set up email sending completely changes with Spree 2. This
change encapsulates that code in a single method so that it can be
changed easily and doesn't create further merge conflicts while we are
still working on the master branch and the Spree upgrade.

#### What should we test?

No test, passing automated tests should be enough.

#### Release notes

No change of production code.

This change should avoid merge conflicts with the Spree upgrade in the future. After this is merged to master, we need to merge master into 2-0-stable. We can then change the `setup_email` method and should not have another merge conflict because of this.

### DIFF
--- a/spec/support/email_helper.rb
+++ b/spec/support/email_helper.rb
@@ -1,0 +1,11 @@
+module OpenFoodNetwork
+  module EmailHelper
+    # Some specs trigger actions that send emails, for example creating an order.
+    # But sending emails doesn't work out-of-the-box. This code sets it up.
+    # It's here in a single place to allow an easy upgrade to Spree 2 which
+    # needs a different implementation of this method.
+    def setup_email
+      create(:mail_method)
+    end
+  end
+end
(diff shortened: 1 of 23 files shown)


spec/support/email_helper.rb
```diff
@@ -0,0 +1,11 @@
+module OpenFoodNetwork
+  module EmailHelper
+    # Some specs trigger actions that send emails, for example creating an order.
+    # But sending emails doesn't work out-of-the-box. This code sets it up.
+    # It's here in a single place to allow an easy upgrade to Spree 2 which
```
luisramos0: I wouldn't mention "spree upgrade" in comments any more. It will be past and outdated soon :-D

luisramos0: I'd remove the 3rd and 4th line.

mkllnk: We need to change this method in the Spree branch once this is all merged. I would remove those lines then. Having it in the master branch will make everybody aware that they shouldn't temper with this code during the Spree upgrade unless there is a really good reason -> avoiding conflicts.

PR #2930, opened 2018-10-25
[Spree Upgrade] WIP Fix line_items_controller destroy spec

### LINKED ISSUE #2929: [Spree Upgrade] Line Items Controller Spec (2 broken tests)
Fix this spec:

```
36) LineItemsController destroying a line item on a completed order where shipping and payment fees apply updates the fees
      Failure/Error: expect(order.adjustment_total).to eq initial_fees
      
        expected: 16
             got: 10.0 (#<BigDecimal:7f73235acab0,'0.1E2',9(18)>)
      
        (compared using ==)
      # ./spec/controllers/line_items_controller_spec.rb:107:in `block (4 levels) in <top (required)>'
```

luisramos0:
#### What? Why?

Closes #2929

### The problem
The problem is caused by the fact that destroying the line item as the controller is doing, does not run update_inventory on the shipment and so the shipment.inventory_unit of the deleted line_item stays on the shipment and the shipment adjustment (shipping fee) is not updated correctly.
### The solution
We need to make sure the line_item is saved so that the hook update_inventory runs and updates the shipment.inventory_units.
Another option would be to create a before_destroy or after_destroy on the line_item decorator to run update_inventory.
### TODO
TODO: The question is: how is this working in Spree and why do we need to do this? The Spree code should be doing this already in some way. Or is this a problem that was solved in some future commit in Spree? Some investigation is needed before going ahead with this solution.

#### What should we test?
Together with #2928 (that will come from master at some point), makes the test pass.


sauloperez:
> Or is this a problem that was solved in some future commit in Spree?

I bet that is the case

luisramos0:
@sauloperez shall we just go ahead with this fix then?

sauloperez:
I would spend some time digging through Spree's upcoming version to understand if they fixed it later on.

luisramos0:
Tests still not green after rebase, needs to be looked at.

PR #2934, opened 2018-10-26
Add offending files to rubocop config

mkllnk:
#### What? Why?

The file .rubocop_todo.yml is generated automatically and contains a
configuration to make all files pass. For a lot of cops it just lists
the offending files. But for some cops it sets a different metric.

Since we don't want these lax metrics, we override them in our config
file .rubocop.yml. This leads to a lot of offences again. This patch
lists all offending files for each cop to make rubocop pass. We can
improve the code over time and remove items from the list.

#### What should we test?

Check that CodeClimate passes.

#### Release notes

Fixed: Code style and quality tool Rubocop provides useful output for developers.

Changelog Category: Fixed

#### Documentation updates

Have we documented our use of Rubocop anywhere? Our practice of removing files from rubocop_todo has to be extended to the rubocop config file.


.rubocop.yml
```diff
@@ -200,3 +200,745 @@ Metrics/ParameterLists:
 
 Metrics/PerceivedComplexity:
   Max: 7
+
+# Additional todo list not generated by --autogen-config
+Metrics/LineLength:
```
sauloperez: ⚠️ this cop is already defined in line 192. The same for the ones below it

.rubocop.yml
```diff
@@ -200,3 +200,745 @@ Metrics/ParameterLists:
...
+  - app/helpers/angular_form_builder.rb
+  - app/models/product_import/entry_processor.rb
+  - app/models/product_import/entry_validator.rb
+  - lib/open_food_network/xero_invoices_report.rb
+  - spec/features/admin/reports_spec.rb
+
+Metrics/BlockNesting:
+  Max: 3
+  Exclude:
+  - app/controllers/checkout_controller.rb
```
sauloperez: How did you generate all these exclusions? It'd be great to doc it at the top of this file and also in a wiki page, if we are to maintain it ourselves.

sauloperez:
Ok, if I understand correctly this won't bring maintenance overhead as this additions come in `.rubocop.yml` and not `.rubocop_todo.yml`. However, what I am afraid of is

captura de pantalla 2018-10-29 a les 15 27 38

Does this mean Code Climate will now hide all this feedback and the score and code smells detected won't reflect reality? I feel like going back and forth with this 🤕 

I think we all agree but let me put this down for clarity. We want a solution that does the following:

1. Spots code smells and gives a real score to our codebase from which we can work on to improve it.
2. A way to know if the code that we add adheres to our style guide defined in `.rubocop.yml`
3. A way to track existing violations of that style guide on which we can work on little by little to eventually get rid of it. We want to this to not raise a million warnings while working on the codebase.

.rubocop.yml
mkllnk: I know that they are already defined above. I see the definition above as our target these lines down here is intermediate todo list. What do you suggest? Should we comment out the definitions above and add a comment like `# look below for current todo list`? I don't think we should include all the files in the list above, because it will be very difficult to read the general style guide.

.rubocop.yml
mkllnk: That was a manual process:
```
rubocop > rubo.log
# inspect log file to see which cops are failing
# copy cop configurations and add Exclude parameter
grep ExampleCop rubo.log | cut -d ":" -f 1 | sort -u >> .rubocop.yml
# use vim to add `- ` before each line
```

Do you think this process is valuable enough to include it in a comment?

mkllnk:
> Does this mean Code Climate will now hide all this feedback and the score and code smells detected won't reflect reality?

Yes. But since we are using `.rubocop_todo.yml`, we are already hiding a lot and the score does not reflect reality.

Code Climate supports custom rubocop configuration files: https://docs.codeclimate.com/docs/rubocop#section-using-a-non-standard-configuration-file
We can create `.codeclimate-rubocop.yml` containing our style guide, but not inheriting from the todo list and not containing all the exclusion from this pull request. The downside would be that we have two places to edit if we change the style guide. The upside is that we have useful feedback locally and Code Climate shows us a real score.

I personally would only go in that direction to make you happy. :-) I'm not using Code Climate for the score. For me it's just another quality check on each pull request.

mkllnk:
@sauloperez @luisramos0 This pull request now has the following goals:

- Ignore all current rubocop violations when run locally. It allows incremental improvements.
- Show all current violations in Code Climate. It will result in a realistic score.

The downside is that we have to configuration files now. If we change rules, we need to change both files. We could move those into an additional file and include it into the two configs.

But it also means that if we see an error in Code Climate and then run rubocop locally, it may not display that error. We may need to run `rubocop -c .rubocop-codeclimate.yml` which will then show all offences (old and new) and Code Climate's complaint may be about only one of them (the new one).

What do you think about this approach?

.rubocop.yml
sauloperez: aha, that's a good idea. Yes, I would explain it as I'm sure I'm won't be the last one that thinks that otherwise.

.rubocop.yml
sauloperez: I'd say so. The easier we make it for other devs, the greater the chances to have it up-to-date.

sauloperez:
Thanks @mkllnk I actually like it! I find it a good trade-off (and the only way to solve this whole thing :sweat: ). It may be a bit annoying to keep two files but is not a big deal and we leverage Code Climate. Let's see how it goes and let's iterate.

PR #2947, opened 2018-10-28
Pi/submenu link

### LINKED ISSUE #2603: [Product Import] Make product import submenu visible for all users + beta sentence
## Description
Today only superadmins can see the submenu. This task is about adding it for all the users.

We want also to add a sentence at the beginning of the page saying "This feature is still in beta: you may experience some errors while using it. Please don't hesitate to contact us at [Insert email]. "

Matt-Yorkley:
#### What? Why?

Closes #2603

Shows the product import menu item in the backend for superadmin users. Includes recreating the Spree view for product submenu.

#### What should we test?

Product import menu item is visible for admins.

#### Release notes

Product Import is now available for admin users.

Changelog Category: Added

### DIFF
--- a/app/views/admin/shared/_product_sub_menu.html.haml
+++ b/app/views/admin/shared/_product_sub_menu.html.haml
@@ -1,0 +1,6 @@
+= content_for :sub_menu do
+  %ul#sub_nav.inline-menu
+    = tab :products, match_path: '/products'
+    = tab :option_types, match_path: '/option_types'
+    = tab :properties
+    = tab :prototypes
--- a/app/views/admin/shared/_product_sub_menu.html.haml
+++ b/app/views/admin/shared/_product_sub_menu.html.haml
@@ -4,3 +4,5 @@
     = tab :option_types, match_path: '/option_types'
     = tab :properties
     = tab :prototypes
+    = tab :variant_overrides, url: main_app.admin_inventory_path, match_path: '/inventory'
+    = tab :import, url: main_app.admin_product_import_path, match_path: '/product_import'
--- a/spec/features/admin/product_import_spec.rb
+++ b/spec/features/admin/product_import_spec.rb
@@ -460,7 +460,7 @@ feature "Product Import", js: true do
 
   def proceed_to_validation
     expect(page).to have_selector 'a.button.proceed', visible: true
-    click_link I18n.t('admin.product_import.import.import')
+    within("#content") { click_link I18n.t('admin.product_import.import.import') }
     expect(page).to have_selector 'form.product-import', visible: true
     expect(page).to have_content I18n.t('admin.product_import.import.validation_overview')
   end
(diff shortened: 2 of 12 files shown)


app/views/admin/shared/_product_sub_menu.html.haml
```diff
@@ -0,0 +1,8 @@
+= content_for :sub_menu do
+  %ul#sub_nav.inline-menu
+    = tab :products, match_path: '/products'
+    = tab :option_types, match_path: '/option_types'
+    = tab :properties
+    = tab :prototypes
+    = tab :variant_overrides, url: main_app.admin_inventory_path, match_path: '/inventory'
+    = tab_without_cancan_check :import, url: main_app.admin_product_import_path, match_path: '/product_import'
```
luisramos0: An opportunity to improve the (currently random?) order of these entries? Can we move the Inventory entry next to second place after Products? maybe a question that @myriamboure can follow up?

spec/features/admin/product_import_spec.rb
```diff
@@ -366,7 +366,7 @@
 
   def import_data
     expect(page).to have_selector 'a.button.proceed', visible: true
-    click_link I18n.t('admin.product_import.import.import')
+    within "#content" do click_link I18n.t('admin.product_import.import.import') end
```
sauloperez: I'd vote for `{}` instead of `do end` for single-blocks. That's the Ruby's standard-ish. Strange that Rubocop didn't complain.

app/views/admin/shared/_product_sub_menu.html.haml
Matt-Yorkley: I think unless you log in as a superadmin, only Products and Inventory show in that submenu anyway. What's your thinking on changing the order?

luisramos0: I think it shouldn't be random, we should group things logically everywhere and seeing products and inventory separated by a random list of menu entries looks wrong.

myriamboure:
@Matt-Yorkley this was in the backlog but not in dev ready (unless I did a mistake ;-)), we were waiting for the upgrade, we want to check that major bugs still happening recently are solved before we make the feature publicly visible. This PR can move forward and be tested but would like to put on hold before merging until upgrade to upcoming release and check majors bugs (might be done tomorrow or in 2-3 days hopefully). To avoid user support problems and prepare correctly the release of that feature yeahhh !

Matt-Yorkley:
Sounds good @myriamboure. There might be some strings in Transifex as well to translate before this is properly relased :+1:

spec/features/admin/product_import_spec.rb
Matt-Yorkley: Done.

daniellemoorhead:
Hey @Matt-Yorkley given @sauloperez put the blocked label on this should it be moved to the Blocked column with all the others? Or is it being handled differently?

Matt-Yorkley:
Ok, I've fixed this one up and it's ready to go. It makes the product import page available to superadmin users.

RachL:
Nice!

@Matt-Yorkley I've just realized that Myriam's issue was mentioning access for all users. I don't know how I've read it before 🤦‍♀️  . But let's move with this PR, have more testing with super admins and real cases and then we will be able to add the link for all users. So I'm disconnecting the issue and moving this to ready to go.

PR #2948, opened 2018-10-28
Fix default time zone config

pfac:
#### What? Why?

* Starting the environment, even with just `bundle exec rake`, results
  in the following error:
  ```
  Value assigned to config.time_zone not recognized.Run "rake -D time"
  for a list of tasks for finding appropriate time zone names.
  ```

This change addresses the issue by:

* Adding the time zone setting to environment specific configurations,
  defaulting to UTC in development, and using Melbourne in tests.

#### What should we test?

For devs:

* Running `bundle exec rake` in a clean environment after setting up the project.

For testers:

* Check that order cycles close at the right time.

#### Release notes

* Changed `TIMEZONE` environment variable to be optional in development and test. The system will now default to `UTC` in development when the variable is not present, similar to what happens in Rails when no time zone is assigned.

Changelog Category: Changed

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -29,6 +29,12 @@ Openfoodnetwork::Application.configure do
   # Expands the lines which load the assets
   config.assets.debug = false
 
+  # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
+  # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
+  #
+  # To override this, set the appropriate locale in application.yml
+  config.time_zone = ENV.fetch("TIMEZONE", "UTC")
+
   config.i18n.fallbacks = [:en]
 
   # Show emails using Letter Opener
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -33,6 +33,8 @@ Openfoodnetwork::Application.configure do
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
 
+  config.time_zone = ENV.fetch("TIMEZONE", "Melbourne")
+
   # Tests assume English text on the site.
   config.i18n.default_locale = "en"
   config.i18n.available_locales = ['en', 'es']


pfac:
A single test failed on `spec/features/admin/orders_spec.rb:162`. Does not seem like something this change would cause. Is this a known randomly failing test, or did I actually break something?

pfac:
Olá Luís. Obrigado pelas boas-vindas, espero poder ajudar. :smile: 

I get that you're supposed to set the timezone in `application.yml` but this one variable is even preventing the environment from correctly load for a simple Rake task. It's unnecessary friction in my opinion, and this little change allows to move forward.

If you're worried about people actually forgetting to set this variable and the system not warning them, then I suggest making this an environment specific config. We set a default for it in development and tests, but we make it so production environments will fail to start if it is not set. Would that be better in your opinion?

(Yes, I am in Braga. Not sure I can make it to the tech meeting, but would love to at least try. The Slack you mentioned is not open for registration, seems to be invite only.)

pfac:
@sigmundpetersen worked perfectly, thanks :+1:

pfac:
> Would you mind adding a comment there to ask people to provide the appropriate locale in `application.yml`?

Not at all, I'll do that as soon as I have the chance.

pfac:
Comment added 👍

config/application.rb
```diff
@@ -107,7 +107,9 @@ class Application < Rails::Application
 
     # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
-    config.time_zone = ENV["TIMEZONE"]
+    #
+    # Be sure to set the appropriate locale in application.yml
+    config.time_zone = ENV.fetch("TIMEZONE", "UTC")
```
mkllnk: There is one little caveat. Our specs depend on Melbourne time. We could just replace `UTC` with `Melbourne` here, but the better solution is probably to set `config.time_zone = "Melbourne"` in `config/environments/test.rb`. Can you add that and check that `spec/controllers/admin/order_cycles_controller_spec.rb` is still passing?

pfac: Ok, you're right, that test was broken.

Changed the approach :+1:, since now all three environments have a different approach (development can use whatever, test needs `Melbourne`, and in production/staging we want to enforce setting it in `application.yml`).

sauloperez:
Thanks again @pfac !

PR #2949, opened 2018-10-28
Fix new preferences migration

pfac:
#### What? Why?

* In a clean environment, running the new preferences migration fails,
  due to a missing file from Spree core being manually required. This
  file has now been missing since ab707cf.

This change addresses the issue by:

* Copying the missing file from Spree Core 1.3.6.beta into the codebase.
  This fixes the issue for now, but also means that a migration merge
  and/or rewrite might be in order for the future.

#### What should we test?

* Running the migrations from scratch in a clean system. Before this change `rake db:migrate` fails with
  ```
  rake aborted!
  LoadError: cannot load such file -- spree/core/preference_rescue
  .../db/migrate/20120327000645_new_preferences.rb:1:in  `<top (required)>'
  .../ruby/2.1.5/bin/bundle:23:in `load'
  .../ruby/2.1.5/bin/bundle:23:in `<main>'
  
  Caused by:
  Polyglot::PolyglotLoadError: Failed to load 
  .../db/migrate/20120327000645_new_preferences.rb using extensions deface, rb
  .../ruby/2.1.5/bin/bundle:23:in `load'
  .../ruby/2.1.5/bin/bundle:23:in `<main>'
  Tasks: TOP => db:migrate
  ```

  After this change `db/migrate/20120327000645_new_preferences.rb:1` should work.

#### Release notes

* Fixed the new preferences migration, which was broken since ab707cf due to the Spree upgrade.

Changelog Category: Fixed

* This does not present a direct conflict with the Spree upgrade, but was caused by its first steps, and should be revised afterwards.

### DIFF
--- a/db/migrate/20120327000645_new_preferences.rb
+++ b/db/migrate/20120327000645_new_preferences.rb
@@ -1,4 +1,37 @@
-require 'spree/core/preference_rescue'
+# Spree 1.3.6.beta preference rescue implementation, required for the new
+# preferences migration, which is broken since commit ab707cf due to the
+# absence of this file.
+#
+# Migration: db/migrate/20120327000645_new_preferences.rb
+# Source: https://raw.githubusercontent.com/spree/spree/1-3-stable/core/lib/spree/core/preference_rescue.rb
+#
+# rubocop:disable all
+module Spree
+  class OldPrefs < ActiveRecord::Base
+    self.table_name = "spree_preferences"
+    belongs_to  :owner, :polymorphic => true
+    attr_accessor :owner_klass
+  end
+
+  class PreferenceRescue
+    def self.try
+      OldPrefs.where(:key => nil).each do |old_pref|
+        next unless owner = (old_pref.owner rescue nil)
+        unless old_pref.owner_type == "Spree::Activator" || old_pref.owner_type == "Spree::Configuration"
+          begin
+            old_pref.key = [owner.class.name, old_pref.name, owner.id].join('::').underscore
+            old_pref.value_type = owner.preference_type(old_pref.name)
+            puts "Migrating Preference: #{old_pref.key}"
+            old_pref.save
+          rescue NoMethodError => ex
+            puts ex.message
+          end
+        end
+      end
+    end
+  end
+end
+# rubocop:enable all
 
 class NewPreferences < ActiveRecord::Migration
 


pfac:
This is failing on Code Climate. I can either refactor the code to pass on Code Climate, or I can disable the checks for this file.

Refactoring would mean the code is no longer the file imported from Spree Core 1.3.6.beta, and should be maintained accordingly. On the other hand, disabling the checks would keep the file exactly as it was imported, meaning it is not permanent and we don't intend to maintain it.

What's the preferred approach here?

I'm gonna go ahead and disable the checks since is the fastest change, but let me know if you want to go the other way.

pfac:
Sure. Replaced the `require` with the content of the file.

Ideally I would patch the migration to be backwards compatible and not to use any of this, but since I don't even have a working environment yet I'm trying to move slowly to avoid breaking unnecessary stuff.

Let me know if you want me to change something else.

pfac:
@luisramos0 I'm still not able to run all the migrations, there are others with issues that I'm working on, if it interests the project.

I am aware that I can use `rake db:schema:load`, and it would skip the migrations. However I usually refrain from doing so in the projects I work on because re-running the migrations provides me this regression information. But if you tell me there is little interest in doing so I'll leave it alone. :smile: 

I went through the setup guides already, but just until the `db:schema:load` (where I tried a different route for the reason I mentioned). I also tried to setup the project up on Ubuntu, which is my main system and the one I'm using right now, but I am unable to install Ruby 2.1.5 as it was recently updated to 18.10. :stuck_out_tongue:

pfac:
The changes have now been rebased and squashed, and are now ready to merge.

sauloperez:
I wouldn't try to fix any other migrations @pfac. This won't provide much value now as we don't migrate new envs but load the schema instead (I think). I'm sure there are more gratifying things you can contribute with.

> I also tried to setup the project up on Ubuntu, which is my main system and the one I'm using right now, but I am unable to install Ruby 2.1.5 as it was recently updated to 18.10. 

You might already be aware of it but I also use Ubuntu together with [Rbenv](https://github.com/rbenv/rbenv) to have diverse ruby versions across my projects. I strongly recommend it.

pfac:
@sauloperez I use [asdf-vm](https://github.com/asdf-vm/asdf) which uses RBenv's ruby-build to install Ruby versions. Switched over from RBenv precisely. :wink: 

I believe the issue should be fixed in the upcoming weeks. Otherwise I'm already working on a plan B (with Docker :grin:). Until then I'll use my Mac for OFN. :+1:

PR #2959, opened 2018-10-30
Move default_url_options set up to an initializer

sauloperez:
### What? Why?

For some reason running `bundle exec rake db:migrate RAILS_ENV=staging` on a deployment fails with:

```
rake aborted!
NameError: uninitialized constant Spree::Config
```

Running `bundle exec rails server` for instance, does not. There must be a difference on the way a rake task and the rails commands load the app.

Moving this configuration to an initializer, at the end of the initialization process, fixes it. The constant `Spree::Config` is already loaded.

**This is preventing the release v1.22.0 from being staged and tested**

### What should we test?

This should fix the deployment to staging therefore, I can test it myself.

### DIFF
--- a/config/initializers/action_mailer.rb
+++ b/config/initializers/action_mailer.rb
@@ -1,0 +1,6 @@
+ActionMailer::Base.configure do |config|
+  if Rails.env.production? || Rails.env.staging?
+    # Use https when creating links in emails
+    config.default_url_options = { protocol: 'https', host: Spree::Config[:site_url] }
+  end
+end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -30,9 +30,6 @@ Openfoodnetwork::Application.configure do
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
 
-  # Use https when creating links in emails
-  config.action_mailer.default_url_options = { protocol: 'https', host: Spree::Config[:site_url] }
-
   # See everything in the log (default is :info)
   config.log_level = :info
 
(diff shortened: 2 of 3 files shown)


sauloperez:
thanks for the idea of trying out an initializer @Matt-Yorkley !

Matt-Yorkley:
teamwork

config/initializers/action_mailer.rb
```diff
@@ -0,0 +1,4 @@
+ActionMailer::Base.configure do |config|
+  # Use https when creating links in emails
+  config.default_url_options = { protocol: 'https', host: Spree::Config[:site_url] }
```
Matt-Yorkley: Ah, I think we actually want to only do this in staging and production. Test and development have very different settings for links, emails and the site url.

sauloperez: yes, that's right. We don't want to mess with that.

daniellemoorhead:
@sauloperez Pls, #assignyourself 😱

daniellemoorhead:
Also, how is this to be tested? Does it have to be tested? There's no detail in the description.

sauloperez:
sorry, sorry :trollface:

sauloperez:
> Also, how is this to be tested? Does it have to be tested? There's no detail in the description.

Added it. Thanks for the heads-up. I gave for granted the scope of this PR was understood.

sauloperez:
Fixed!

```
TASK [deploy : migrate database] ***********************************************
Wednesday 31 October 2018  09:09:47 +0100 (0:00:00.305)       0:01:20.374 *****
changed: [staging.katuma.org]
```

Issue #2960, opened 2018-10-31
Users inventory levels changed for unknown reason


sstead:
## Description
An Australian user reported that their Inventory levels (set on /inventory) changed without them doing anything on Oct 30th evening. The user checked the levels before opening the shop, and then found some time later that the numbers had changed to incorrect numbers. Thus it's unlikely to be caused by user error. The user is not using upload tool.

The user reported the on hand values for most of her approx 250 products had changed. Some were raised (from zero to >0) or lowered ( from e.g. 9 to 0). Example products impacted: carrots, cucumbers,  Rhon de nice zucchinis,  rockmelon, Fuji apples,  all of the avocado varieties,  kiwi fruit . These items were all originally set to 0 on hand, but after the error they had stock on hand.  The user could see no pattern/logic in the new on hand figures that appeared.

About the Enterprise - Terrewah Farm - This is a producer hub. They have permission to stock one other enterprise's products. The vast majority of products in her Inventory belong to Terrewah Farm. Should this user be usiing Inventory - probably not as it's mostly a producer shop- no need for overrides.
![image](https://user-images.githubusercontent.com/10843056/47760477-93908900-dd08-11e8-9c30-42d7aaf4d147.png)

One possibility is that this bug is related to https://github.com/openfoodfoundation/openfoodnetwork/pull/2893 . Why? Because this PR was merged to Aus production shortly before the issue was reported. This might not be related though. See discussion on the PR.

## Expected Behavior
Inventory levels should remain as levels last saved by user.

## Actual Behavior
Inventory levels have been reported to change without the users' involvement.

## Steps to Reproduce
? Unknown.

## Context
She received orders for items that should have been out of stock and vice versa.

## Severity
If this is a widely reaching but it's S2, but so far there's only been one report of issues. May be Aus only, or certain enterprise configuration only.

## Your Environment
Aus production.

kirstenalarsen:
@luisramos0 @mkllnk @myriamboure does this seem connected to the other discussions?

myriamboure:
> About the Enterprise - Terrewah Farm - This is a producer hub. They have permission to stock one other enterprise's products. The vast majority of products in her Inventory belong to Terrewah Farm. Should this user be usiing Inventory - probably not as it's mostly a producer shop- no need for overrides.

[...]

PR #2961, opened 2018-10-31
Revert "Merge pull request #2893 from luisramos0/deleted_products_bre…

sauloperez:
…ak_inventory"

#### What? Why?

This reverts commit 8a3f621b42dfa1248d6b647bcf53befe83790c28, reversing
changes made to 7cac4637fe566c32e17402ed3cf0b5ad7a19443d.


daniellemoorhead:
@sauloperez #assignyourself #please 🙏

mkllnk:
@luisramos0 I just read through the thread at https://github.com/openfoodfoundation/openfoodnetwork/pull/2893 and there are still two open questions:

1. Why are permissions in place if the user sells their own enterprise? Is this scenario still valid?
2. Does reverting fix anything? If the data changed because of the migration, this pull request doesn't change anything about it. The migration didn't have a `down` method and there is no way for a rollback via migration. The user already fixed the stock levels herself. So maybe no further code change is needed for Australia.

luisramos0:
Maikel, good questions:
1- both the code and the migration will revoke inventory items even if override hub_id = produdct supplier_id. we need to fix this in both code and migration.
2- In AUS now, everytime users create permissions, the data will be corrupted again. We should run quick fix (see below) to fix AUS data and revert the change. the more you keep the system as is the more inventories will be messed up.

See here:
https://github.com/openfoodfoundation/openfoodnetwork/issues/2960#issuecomment-434627038

luisramos0:
Actually, changing my mind.
@mkllnk is right on point 2. This code is only revoking permissions on delete (user needs to delete permissions) and will only affect the overrides of that permission (specific relation between 1 producer and 1 hub), no own overrides involved. Also, the code will review overrides when new permission is created but again it will not affect own overrides.

So, the bug is only in the migration. We need to fix it and run it again.

luisramos0:
And, correcting what I say above about point 1.
1- Only the migration will revoke inventory items even if override hub_id = product supplier_id.
The code will only revoke permissions that are part of the permission being created or deleted.

So, actually there is a problem in the code when a permission to manage own inventory is added and then removed (it's a edge case, but will break the inventory). the fix is to re-add the permission. I think we can ignore this case but I'll put this on 2893 for reference.

PR #2963, opened 2018-10-31
Fix variant_overrides permissions for overrides that belong to the supplier herself

### LINKED ISSUE #2960: Users inventory levels changed for unknown reason

luisramos0:
#### What? Why?

Closes #2960 

This PR sets all variant_overrides.permission_revoked_at to nil when variant_override.hub_id is the products supplier_id. Basically allowing all suppliers to manage their own inventory without having to set explicit permission to do so.

#### What should we test?
This should be empty:
```
select distinct vo.hub_id from spree_products p, spree_variants v, variant_overrides vo
openfoodnetwork-> where p.supplier_id = vo.hub_id
openfoodnetwork-> and p.id = v.product_id and vo.variant_id = v.id and vo.permission_revoked_at is not null;
```

#### Release notes
Changelog Category: Fixed
Fixed previous data migration by not revoking access to inventory management for suppliers who manage their own inventory but that do not have explicit enterprise permissions to do so.

not related.

#### Dependencies
Fixes problem in the migration in #2893

### DIFF
--- a/db/migrate/20181031105158_allow_all_suppliers_own_variant_overrides.rb
+++ b/db/migrate/20181031105158_allow_all_suppliers_own_variant_overrides.rb
@@ -1,0 +1,12 @@
+class AllowAllSuppliersOwnVariantOverrides < ActiveRecord::Migration
+  def up
+    # This migration is fixing a detail of previous migration RevokeVariantOverrideswithoutPermissions
+    #   Here we allow all variant_overrides where hub_id is the products supplier_id
+    #   This is needed when the supplier herself uses the inventory to manage stock and not the catalog
+    owned_variant_overrides = VariantOverride.unscoped
+      .joins(variant: :product).where("spree_products.supplier_id = variant_overrides.hub_id")
+
+    owned_variant_overrides.update_all(permission_revoked_at: nil)
+  end
+end
+
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20181020103501) do
+ActiveRecord::Schema.define(:version => 20181031105158) do
 
   create_table "account_invoices", :force => true do |t|
     t.integer  "user_id",    :null => false


myriamboure:
Would be great to include in the release I guess to avoid messing up people inventories in all instances I guess @sauloperez right?

Matt-Yorkley:
Yes, we shouldn't release without this PR. 

@luisramos0 What's the plan for testing and proceeding with this? Merge into master and test on Aus production before the release?

Matt-Yorkley:
It's only affecting variant overrides where the enterprise that owns the product is also the hub stocking the inventory, and making sure they have non-revoked permissions. Looks ok to me :+1:

db/migrate/20181031105158_allow_all_suppliers_own_variant_overrides.rb
```diff
@@ -0,0 +1,12 @@
+class AllowAllSuppliersOwnVariantOverrides < ActiveRecord::Migration
+  def up
+    # This migration is fixing a detail of previous migration RevokeVariantOverrideswithoutPermissions
+    #   Here we allow all variant_overrides where hub_id is the products supplier_id
+    #   This is needed when the supplier herself uses the inventory to manage stock and not the catalog
+    owned_variant_overrides = VariantOverride.unscoped
+      .joins(variant: :product).where("spree_products.supplier_id = variant_overrides.hub_id")
+
+    owned_variant_overrides.update_all(permission_revoked_at: nil)
+  end
```
sauloperez: This could have some impact on performance for users using OFN, depending on the number of `VariantOverride` in the DB. This is basically what's happening (I might be wrong, of course):

1. An `ActiveRecord::Relation` is built an assigned to `owned_variant_overrides`.
2. In line 9, this `ActiveRecord::Relation` is executed issuing an `UPDATE` statement against DB. The resulting statement will lock all variant overrides that fulfill `spree_products.supplier_id = variant_overrides.hub_id` for the duration of the update.

This means that if the number of variant overrides fulfilling that is large, users updating them at the same time we run the migration will have a bad experience. It is not until the DB transaction is finished that the lock is released.

I guess all these conditions will hardly happen together but can we check @mkllnk? We basically need to get a count and then decide.

sauloperez:
I agree with @luisramos0 . Let's ship v1.22.0 as it has been waiting too long already, get the ton of fixes in all instances, while we prepare a v1.22.1 with this fix if we need to.

We still think about releases as something that only happens once, which leads to people rushing and sneaking in PRs that are not ready. We should think about them as a metro; if you miss this one you can catch the next one, which comes in 5min.

Matt-Yorkley:
In that case we'll need to make sure the original commits are reverted, right?

sauloperez:
According to @luisramos0 's summary on the #bug channel (and in the PR) there's no need to.

Matt-Yorkley:
@sauloperez The summary that we wouldn't need to revert #2893 was added with the proviso that we would apply the second migration to fix it instead of reverting it. That migration is in this PR, so if we are not including this PR in the release but we are including #2893, then it will indeed break inventories for the other instances when it's deployed, right?

sigmundpetersen:
That's absolutely the way I understand @luisramos0 as well @Matt-Yorkley

sauloperez:
sure @Matt-Yorkley . That is equivalent to releasing v1.22.0 as it is and then separately deploying this branch on AUS to run the migration and then merging it into `master`. If you guys prefer I can include it in the release. No worries.

myriamboure:
Yes please include in the release, I would like to avoid all instance having the same issue as Australia had... it has an impact on users, twice, the fix also impact eh inventory levels will be changed a second time and user will have to check again. If @mkllnk  can merge this night you can include in the release tomorrow and ship it @sauloperez :-)

daniellemoorhead:
Damn, the build has failed on semaphore. @mkllnk this is your #1 priority today, ping @sstead if you need her to test it, let's get it onto aus production asap and then into the release!

mkllnk:
Good work @luisramos0.

Before I deploy, I want to outline the gaps in this process though. This is not failsafe code.

* This pull request assumes to be deployed with #2893. It doesn't actually fix the old migration. Running the migration in #2893 is still breaking things. Then running the new migration is fixing it again, but on a busy server there could be a gap in which orders are placed, screwing up the stock levels. If we don't fix this: Please only deploy when no orders are happening. Maybe shut down the application for the duration of the deploy.
* The Australian server is *broken* already. We know only of one affected shop so far. They fixed up their inventory levels and had orders since. If we now restore the old inventory levels, they will be wrong. I need a process to restore the current inventory levels for that shop. They have 240 products. That's not quick to do manually. Additionally, we don't know if other shops were affected as well. It would need a lot of research in the database to see if variants have been updated since the deploy and if these variants have been ordered. Only with all the detailed information we may be able to restore the correct stock levels.

I will do some sanity checks on the Australian database first, but I'm facing a trade off here: The more time I spend on this, the better I can restore the data. But time also increases the chances of more data corruption and false orders.

I also have an idea how we can make these things easier in the future: Migrations have to be reversible. If data is lost during a migration, we should write a log file containing that data. The migration should be able to restore the data from the log file. This is additional work, but fixing up migrations that have gone wrong are even more work.

mkllnk:
For documentation purposes:

```
owned_variant_overrides = VariantOverride.unscoped.joins(variant: :product).where("spree_products.supplier_id = variant_overrides.hub_id")
# count 731, affected businesses: 16
all_affected = owned_variant_overrides.where("(variant_overrides.count_on_hand is not null and spree_variants.count_on_hand != variant_overrides.count_on_hand) or (variant_overrides.on_demand is not null and variant_overrides.on_demand != spree_variants.on_demand)")
# count 541
stock_affected = owned_variant_overrides.where("(variant_overrides.count_on_hand is not null and spree_variants.count_on_hand != variant_overrides.count_on_hand)")
# count 540

Spree::Order.where("created_at > ?", 3.days.ago)
# count 1107
affected_orders = Spree::Order.where("created_at > ?", 3.days.ago).where(distributor_id: stock_affected.select(:hub_id))
# count 172

Spree::Order.where("created_at > ?", 3.days.ago).where(distributor_id: stock_affected.select(:hub_id)).select(:distributor_id).uniq
# count: 10

Enterprise.where(id: Spree::Order.where("created_at > ?", 3.days.ago).where(distributor_id: stock_affected.select(:hub_id)).select(:distributor_id).uniq).pluck(:name)
["Escape and Rebellion",
 "Picots Farm",
 "Herry's Harvest",
 "Fat Beets Food Hub",
 "Woggoon Farm Samford ",
 "Terrewah Farm",
 "New Life Farm",
 "Mush Love Gourmet Mushrooms",
 "Mirboo Pastured Poultry",
 "The Mandarin Bend"]
```

We will notify all these enterprises to check their stock levels after we deployed this pull request. Before we deploy this, I will also modify the stock levels of the farm we know updated their stock levels. I will choose the minimum as their main problem was getting orders for products they don't have.
```
terr = Enterprise.find_by_name "Terrewah Farm"
owned_variant_overrides.where(hub_id: terr).each { |v| next if v.on_demand; VariantOverride.unscoped.where(id: v).update_all(count_on_hand: v.variant.count_on_hand) if v.count_on_hand > v.variant.count_on_hand; }
```

Sorry, this is all a bit messy, but it's one-time code and I extracted some CSV sheets to verify the code. Worst case we have to adjust it manually.

luisramos0:
thanks for taking this one @mkllnk I am really sorry I could finish the clean up...

it looks good!
The real number of affected_orders should be lower than 172 as we are only looking at orders that refer directly to VOs, not all orders of all affected hubs, affected hubs may not have VOs for all their inventory.

Thanks again!

PR #2896, opened 2018-10-19
Use https on all links in emails in production

Matt-Yorkley:
#### What? Why?

Closes #2895

Links created by devise in emails were being built with http instead of https. Aside form this not being good practice in general, it seemed to be causing some issues on UK prod where certain user's browsers were complaining, and causing support issues to pile up.

#### What should we test?

Links in emails should now use https on staging and production.

- sign up
- sign up confirmation
- password reset
- order confirmation
- new enterprise
- enterprise manager invitation
- order cycle report for producer
- subscription emails: (testing one of the following should be enough)
  - confirmation
  - empty order
  - placement
  - failed payment
  - placement summary
  - confirmation summary

#### Release notes

Ensure links in emails use https. #2896 

Changelog Category: Changed

PR #2984, opened 2018-11-07
Specify only email link protocol setting

### LINKED ISSUE #2895: Email Confirm Links not working for some users
For a long time the UK has had an issue with some users not being able to confirm their email addresses.

Asking around other instances, I've not heard this happening in other places, so figured it was a UK config issue. After much (soul) searching it seems that the problem is actually with the way the spree configures URLs

## Expected Behavior

Links in OFN emails break for some users

## Actual Behavior

Links in all OFN emails will work

## Steps to Reproduce

We have not been able to reproduce this issue with consistency.

## Context

UK gets at least one user per week that has not been able to confirm their account or reset their password.

The workaround is them forwarding the email to UK support team as none of us experience the same problem.

We have not found a pattern in browser but haven't tracked mail clients or virus software.

## Severity

## Possible fix
Links in emails are always written with http rather than https. My hunch is that some specific virus protection or mail client in the UK is blocking these links as the http in the email does not match the https of the UK website.

Matt-Yorkley:
#### What? Why?

Closes #2895 

There has been some back and forth with this bugfix in #2896, where it seemed to work on staging but was patched at the last minute in #2959 due to an Ansible/deployment issue. This PR removes the patch in #2959.

I've tested this on UK staging before and after deploying the PR and it seems to be working correctly with these adjustments.

#### What should we test?

Email links should use https

#### Release notes

Patched a previous fix for https links in emails.

Changelog Category: Fixed

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -30,6 +30,9 @@ Openfoodnetwork::Application.configure do
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
 
+  # Use https in email links
+  config.action_mailer.default_url_options = { protocol: 'https' }
+
   # See everything in the log (default is :info)
   config.log_level = :info
 
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -30,6 +30,9 @@ Openfoodnetwork::Application.configure do
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
 
+  # Use https in email links
+  config.action_mailer.default_url_options = { protocol: 'https' }
+
   # See everything in the log (default is :info)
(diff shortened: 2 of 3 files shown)


luisramos0:
staged: http://staging.katuma.org

RachL:
Again all I see is https, hope I'm not missing anything :)
Testing notes:
https://docs.google.com/document/d/1_3LlSc_5Uf-g9ycVsNy_FWGDtofqfgPVvE0lC-WrGZk/edit
@sauloperez on UK and AUS staging I don't see "spree demo site" anymore and I have links that are working right away with the correct URL. Do you think Katuma staging could have the same set up?

mkllnk:
@sauloperez It's a known issue: https://github.com/openfoodfoundation/ofn-install/issues/243

In Australia we solve it by loading custom baseline data instead of resetting the whole database.

PR #2991, opened 2018-11-07
[Spree Upgrade] Import missing translations from Spree 2 to fix some tests

Matt-Yorkley:
#### What? Why?

Fixes a couple of errors in `Job #3` that look like this:

```
Failure/Error: expect(page).to have_content 'Your content has been successfully updated!'
    expected to find text "Your content has been successfully updated!" in "translation missing: en.successfully_updated Loading... 
```

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2046,6 +2046,9 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   not_visible: not visible
   you_have_no_orders_yet: "You have no orders yet"
   show_only_complete_orders: "Only show complete orders"
+  successfully_created: ! '%{resource} has been successfully created!'
+  successfully_removed: ! '%{resource} has been successfully removed!'
+  successfully_updated: ! '%{resource} has been successfully updated!'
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -2046,6 +2046,9 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   not_visible: not visible
   you_have_no_orders_yet: "You have no orders yet"
   show_only_complete_orders: "Only show complete orders"
+  successfully_created: ! '%{resource} has been successfully created!'
```
mkllnk: What are those exclamation marks doing there? I found [some hint](https://stackoverflow.com/a/9703472/3377535) that this makes it a "vanilla string", but I have no idea why it wouldn't be a vanilla string.

Matt-Yorkley: I copied these directly from Spree's translations. I'm not sure about the bang!

sauloperez: If that's what they do in Spree, I'd just keep it. I agree that it's weird. Shall we merge?

PR #2994, opened 2018-11-08
[Spree Upgrade] Fix delete enterprise fees spec

### LINKED ISSUE #2987: [Spree Upgrade][BackOffice] Fix deleting enterprise fee behaviour (1 broken test)
## Description

## Expected Behavior

## Actual Behavior

The spec fails.

## Steps to Reproduce

## Possible Fix

HugsDaniel:
#### What? Why?

Closes #2987 

For some reason, the destroy action was processed as HTML and not JS. Putting the format to JS in the delete URL fixed it.

#### What should we test?

`spec/features/admin/enterprise_fees_spec.rb:122` should pass.
Job#1 should be green :v:

This is part of fixing Spree 2.0 build.


HugsDaniel:
screenshot from 2018-11-08 11-49-38

freddy

Issue #2943, opened 2018-10-26
[Spree Upgrade][Models] Fix line Item spec (4 broken tests)


luisramos0:
Quite a few different errors in spec/models/spree/line_item_spec.rb
Most of the broken tests in line item spec are related to variant overrides and scoping. First we need to take care of #2559 

PR #3132 fixed one of the originally 11 broken tests (spec/models/spree/line_item_spec.rb:96).

The remaining 10 broken tests are:

```
  8) Spree::LineItem capping quantity at stock level caps at zero when stock is negative
     Failure/Error: update_attributes!(quantity: variant.on_hand) if quantity > variant.on_hand
     
     ActiveRecord::RecordInvalid:
       Validation failed: Quantity "Product #168 - 8121 (1g, S)" is out of stock.
     # ./app/models/spree/line_item_decorator.rb:65:in `cap_quantity_at_stock!'
     # ./spec/models/spree/line_item_spec.rb:102:in `block (3 levels) in <module:Spree>'

  9) Spree::LineItem capping quantity at stock level when a variant override is in place caps quantity to override stock level
     Failure/Error: li.remove_instance_variable(:@scoper)
     
     NameError:
       instance variable @scoper not defined
     # ./spec/models/spree/line_item_spec.rb:115:in `remove_instance_variable'
     # ./spec/models/spree/line_item_spec.rb:115:in `block (4 levels) in <module:Spree>'

  10) Spree::LineItem capping quantity at stock level when a variant override is in place when count on hand is negative caps at zero
      Failure/Error: li.remove_instance_variable(:@scoper)
      
      NameError:
        instance variable @scoper not defined
      # ./spec/models/spree/line_item_spec.rb:115:in `remove_instance_variable'
      # ./spec/models/spree/line_item_spec.rb:115:in `block (4 levels) in <module:Spree>'

  11) Spree::LineItem tracking stock when quantity is changed when the order is already complete when a variant override applies draws stock from the variant override
      Failure/Error: expect{line_item.increment!(:quantity)}.to_not change{Spree::Variant.find(variant.id).on_hand}
        expected `Spree::Variant.find(variant.id).on_hand` not to have changed, but did change from 4 to 3
      # ./spec/models/spree/line_item_spec.rb:147:in `block (5 levels) in <module:Spree>'

  12) Spree::LineItem tracking stock when a line item is destroyed when the order is already complete when a variant override applies restores stock to the variant override
[...]

PR #3004, opened 2018-11-08
[Spree Upgrade] Fix line Item specs with problems related to VariantStock.on_hand=

luisramos0:
#### What? Why?

Starts #2943 

#### What should we test?
Fixes 1 spec and moves another to a new error.

### DIFF
--- a/spec/models/spree/line_item_spec.rb
+++ b/spec/models/spree/line_item_spec.rb
@@ -98,7 +98,7 @@ module Spree
       end
 
       it "caps at zero when stock is negative" do
-        v.update_attributes(on_hand: -2)
+        v.update_attributes! on_hand: -2
         li.cap_quantity_at_stock!
         expect(li.reload.quantity).to eq 0
       end
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -16,6 +16,7 @@ module VariantStock
   extend ActiveSupport::Concern
 
   included do
+    attr_accessible :on_hand
     after_update :save_stock
   end
 


app/controllers/spree/credit_cards_controller.rb
```diff
@@ -26,6 +26,8 @@ def update
       else
         update_failed
       end
+      rescue ArgumentError => e
```
sauloperez: We need to watch out here. Since this is a very broad exception class we will rescue many cases we don't want to and that will be annoying, like probably errors in specs. What's the problem that we're trying to solve?

spec/models/spree/line_item_spec.rb
```diff
@@ -68,7 +68,8 @@ module Spree
       let!(:li) { create(:line_item, variant: v, quantity: 10, max_quantity: 10) }
 
       before do
-        v.update_attributes! on_hand: 5
+        v.on_hand = 5
+        v.save
```
luisramos0: @sauloperez this simple test only works if I save the variant after setting on_hand on it (uses VariantStock). I am not sure if we should keep this save here in the spec on in the VariantStock methods themselves.

sauloperez: `update_attributes!` requires [attr_accessible](https://apidock.com/rails/ActiveRecord/Base/attr_accessible/class) to work. I went ahead and commit a fix so that you see it in action. Do as you please with it.

We can fix the other attributes of VariantStock either here, in another PR or create a separate issue. Your call.

luisramos0: ok, if you think this update_attributes thing with attr_acessible is the best way to go here I can do it in this PR yes. thanks!
I wouldnt do for the other methods if I dont need them, yagni.

sauloperez: Good point. Let's see them failing first :ok_hand:

sauloperez: Good point. Let's see them failing first :ok_hand:

luisramos0: done, thanks for one more lesson!

PR #3008, opened 2018-11-09
[Spree Upgrade] Add an API endpoint for EnterpriseFeesController#destroy

### LINKED ISSUE #2987: [Spree Upgrade][BackOffice] Fix deleting enterprise fee behaviour (1 broken test)

### LINKED ISSUE #3000: [Spree Upgrade] Enterprise Fees - use json API endpoint
## Description

As discussed with #2994, we should switch to an API endpoint for this action (delete enterprise fee).

Also we should get rid of the `before_fitler`s in `enterprise_fees_controller.rb`.

HugsDaniel:
### WIP
Still one little thing to finish but can be reviewed so far.

#### What? Why?

Closes #3000 
Closes #2987 

This adds a new endpoint for destroying enterprise fees through the API.

The latter `do_not_destroy_referenced_fees` method in `admin/enterprise_fees_controller.rb` has been replaced by a `before_destroy` callback in `enterprise_fee.rb`.

Specs have been added to cover the new `api/enterprise_fees_controller.rb` destroy action.

**[TODO]** - In case of error, the error is rendered in JSON, so in the browser we will have some kind of json object inside the error banner, which is bad. We have to find a way (with Angular?) to display only the messages of these errors.

#### What should we test?

When deleting an enterprise fee that is not related to any product (in Products > Product Distributions), it should delete normally.
When deleting an enterprise fee that is related to a product, it should fail and display a message `That enterprise fee cannot be deleted as it is referenced by a product distribution: #{product.id} - #{product.name}.`

Also this was originally made to fix specs from `spec/features/admin/enterprise_fees_spec.rb:122`, which should now be green.

`spec/controllers/api/enterprise_fees_controller_spec.rb` should pass.

#### Release notes

EnterpriseFeesController#destroy is now going though an API endpoint.

Changelog Category: Changed

This changes the way enterprise fees are deleted and fixes some specs of the build.

### DIFF
--- a/app/assets/javascripts/admin/enterprise_fees/directives/delete_resource.js.coffee
+++ b/app/assets/javascripts/admin/enterprise_fees/directives/delete_resource.js.coffee
@@ -1,7 +1,7 @@
 angular.module('admin.enterpriseFees').directive 'spreeDeleteResource', ->
   (scope, element, attrs) ->
     if scope.enterprise_fee.id
-      url = '/admin/enterprise_fees/' + scope.enterprise_fee.id
+      url = '/api/enterprise_fees/' + scope.enterprise_fee.id
       html = '<a href="' + url + '" class="delete-resource icon_link icon-trash no-text" data-action="remove" data-confirm="' + t('are_you_sure') + '" url="' + url + '"></a>'
       #var html = '<a href="'+url+'" class="delete-resource" data-confirm="Are you sure?"><img alt="Delete" src="/assets/admin/icons/delete.png" /> Delete</a>';
       element.append html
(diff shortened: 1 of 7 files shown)


HugsDaniel:
New PR for this against master in #3011, closing this one.

luisramos0:
@HugsDaniel I found recently that you can just edit the title of the PR to change the branch you want to merge to. No need for a new PR in this situation.

If you also want to change the source branch, you can do it in two different ways:
- locally, renaming original branch to something else and create a new one with the same name as the PR one, then, push force to remote
- locally, create new branch with new name and push to original branch in remote (you can push from local branch A to remote branch B)
These are useful git gymnastics :-)

PR #3009, opened 2018-11-09
[Spree Upgrade] Authenticate API user through logged in user

HugsDaniel:
#### What? Why?

Related to #3000 

Use an alias_method to set the `current_api_user` to `spree_current_user` when a call to the API is made.

#### What should we test?

When calling the API while logged in in the app, the `current_api_user` should be the logged in user.

#### Release notes

The API user is now the app's logged in user, if any user is logged in. This prevents action with `authorize!` checks to fail, but parts of the API with no such check remain open for non authenticated users.

Changelog Category: Added

This comes from the fact that [this line](https://github.com/openfoodfoundation/spree/blob/69db1c090f3711088d84b524f1b94d25e6d21616/api/app/controllers/spree/api/base_controller.rb#L72) now doesn't have the `try_spree_current_user` setting, which means that if no API key is provided, the API user will be set to `Spree.user_class.new`, making any `authorize!` check for the failing.


app/controllers/api/base_controller.rb
```diff
@@ -13,5 +13,16 @@ class BaseController < Spree::Api::BaseController
...
+    private
+
+    # Use logged in user (spree_current_user) for API authentication (current_api_user)
+    def authenticate_user_with_spree_current_user
+      @current_api_user = try_spree_current_user
+      
+      # Authenticate API user
+      authenticate_user_without_spree_current_user
+    end
+    alias_method_chain :authenticate_user, :spree_current_user
```
sauloperez: Although this solves the problem, IMO there's no benefit on using `alias_method_chain` here. There are two things that lead me to think we should implement this differently:

* We've been actively removing uses of `alias_method_chain` in the upgrade and this goes in the opposite direction.
* It's been removed from Rails codebase itself and it's somewhat discouraged in the Rails community, for the reasons explained in https://yehudakatz.com/2009/03/06/alias_method_chain-in-models/ and https://www.justinweiss.com/articles/rails-5-module-number-prepend-and-the-end-of-alias-method-chain/.

If I'm not mistaken we can leverage the inheritance we're already using here. What about?

```suggestion
    def authenticate_user
      @current_api_user = try_spree_current_user
      super
    end
```

luisramos0: ah, yes, we are extending not class_evaling.... this was a SUPER suggestion @sauloperez :-)

sauloperez: lol @luisramos0

PR #3011, opened 2018-11-10
[WIP] Add an API endpoint for EnterpriseFeesController#destroy

### LINKED ISSUE #2987: [Spree Upgrade][BackOffice] Fix deleting enterprise fee behaviour (1 broken test)

### LINKED ISSUE #3000: [Spree Upgrade] Enterprise Fees - use json API endpoint

HugsDaniel:
### WIP
Still one little thing to finish but can be reviewed so far.

#### What? Why?

Closes #3000 
Closes #2987 

This adds a new endpoint for destroying enterprise fees through the API.

Specs have been added to cover the new `api/enterprise_fees_controller.rb` destroy action.

**[TODO]** - In case of error, the error is rendered in JSON, so in the browser we will have some kind of json object inside the error banner, which is bad. We have to find a way (with Angular?) to display only the messages of these errors.

#### What should we test?

When deleting an enterprise fee that is not related to any product (in Products > Product Distributions), it should delete normally.
When deleting an enterprise fee that is related to a product, it should fail and display a message `That enterprise fee cannot be deleted as it is referenced by a product distribution: #{product.id} - #{product.name}.`

`spec/controllers/api/enterprise_fees_controller_spec.rb` should pass.

#### Release notes

EnterpriseFeesController#destroy is now going though an API endpoint.

Changelog Category: Changed


app/models/enterprise_fee.rb
```diff
@@ -68,6 +69,16 @@ def ensure_valid_tax_category_settings
     return true
   end
 
+  def ensure_no_product_distributions
+    dependent_distribution = ProductDistribution.where(enterprise_fee_id: self).first
+    if dependent_distribution
```
Matt-Yorkley: CodeClimate is complaining about this line. It's as Luis pointed out the other day, we can use `return true unless dependent_distribution` here, remove the `end` below, and remove the indentation on the lines in between.

luisramos0: yep, although if you do it here you would be moving the error case to the base indentation which is not so obviously good, you want the base indentation to be the happy path of the method (anyway, it's a small detail, I am just commenting because I think it's interesting)

spec/controllers/api/enterprise_fees_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      context "when the fee is referenced by a product distribution" do
+        it "should render 403" do
+          spree_delete :destroy, id: referenced_fee.id, format: :json
+          expect(response.status).to eq 403
+        end
+
+        it "should display the correct error message" do
+          spree_delete :destroy, id: referenced_fee.id, format: :json
+          expect(response.body).to match(/That enterprise fee cannot be deleted/)
+        end
```
Matt-Yorkley: I think I would one more detail to these tests, here we are just checking the response code, but maybe we should check what has happened to the data as well. So under `it "removes the fee"` we could check the fee has actually been deleted from the database (as well as the response), and in the other we can check that it has not been deleted. 

I'd maybe merge the two tests for the referenced fee into one test, like `it "does not remove the fee"` and just have a couple of `expect` lines in the single test to check it's not deleted and check the response/message.

app/controllers/api/enterprise_fees_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  class EnterpriseFeesController < BaseController
+    respond_to :json
+
+    def destroy
+      authorize! :destroy, @enterprise_fee
+
+      if enterprise_fee.destroy
+        render text: I18n.t(:successfully_removed), status: 204
+      else
+        render json: { errors: enterprise_fee.errors }, status: 403
```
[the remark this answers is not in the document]
HugsDaniel: Yes, my problem is that we are calling the API so the response should be in JSON right ? Also, if I render text it'll be with something like `render text: @enterprise_fee.errors[:base], status: 403` which is not ideal for other errors that might occur. I'm looking for a way to render just the error's full message via Angular, so that I can keep rendering JSON.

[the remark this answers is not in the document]
HugsDaniel: @mkllnk @Matt-Yorkley any ideas on this one ? Should I render text for now and open an issue to remember it ?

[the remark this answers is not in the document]
mkllnk: You are right that the API should return JSON, but that can happen in a different pull request. For now, it's best to keep the current API (text) which is compatible with the used Spree code: https://github.com/openfoodfoundation/spree/blob/69db1c090f3711088d84b524f1b94d25e6d21616/backend/app/assets/javascripts/admin/admin.js.erb#L174-L196

We don't want do rely on Spree's backend code and we want to get away from jQuery as well. So it would be another task to implement an Angular directive to do the same thing, but with JSON instead of text.

spec/controllers/api/enterprise_fees_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    before do
+      allow(controller).to receive(:spree_current_user) { current_user }
+    end
+
+    describe "destroy" do
+      it "removes the fee" do
+        spree_delete :destroy, id: unreferenced_fee.id, format: :json
+        expect(response).to be_success
+        expect { unreferenced_fee.reload }.to raise_error ActiveRecord::RecordNotFound
```
sauloperez: RSpec's `change` matcher comes in handy for these cases :point_right: https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/change-matcher

Could be implement as

```suggestion
        it "removes the fee" do
          expect { spree_delete :destroy, id: unreferenced_fee.id, format: :json }
            .to change(EnterpriseFee).by(1)
        end
```

I find it a bit more readable as it clearly states what you're actually checking.

mkllnk: Does `change(EnterpriseFee).by(1)` work? I always thought it has to be  `change(EnterpriseFee.count).by(1)`.

sauloperez: yep, sorry I missed `count` there. You can use either `change { EnterpriseFee.count }` or `change(EntepriseFee, :count)`

HugsDaniel: I did it with change `{ EnterpriseFee.count } by -1` and it works fine !

luisramos0:
ok, @HugsDaniel, lets move to ready to be tested?

luisramos0:
staged in https://staging.openfoodfrance.org

RachL:
This is good to go!

Testing notes:

https://docs.google.com/document/d/1_q9FzY4vL_9oUoQTAeoX-kFOV0vdjj5uWAPNjk5HKA8/edit#

PR #2771, opened 2018-09-25
[Spree Upgrade] Remove obsolete Spree 2 backport on Order

mkllnk:
#### What? Why?

Closes #2520

We backported error handling logic for processing payments from Spree 2. We don't need this backport any more.

#### What should we test?

Not testable yet.

Issue #2520, opened 2018-08-07
[Spree Upgrade] Subs - Remove `process_payments!` override


mkllnk:
## Description

As summarised in https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade:-Checkout-customisations, we backported functionality from Spree 2 into our code. With the Spree upgrade, we can get rid of the code: https://github.com/openfoodfoundation/openfoodnetwork/blob/32846af23019e4f00d0f7b5fdb1d9fb797641912/app/models/spree/order_decorator.rb#L327-L344

## Expected Behavior

Less code, but everything still works.

## Actual Behavior

Duplicate code. We override Spree's code with the same functionality.

## Steps to Reproduce

1. Look at the code. ;-)

This is part of the Spree upgrade and the proposed change should be merged into the 2-0-stable branch.

mkllnk:
The Spree 2 method is almost exactly the same as our override. The only difference is an update in Spree 2 which would be good to have as well: 884c0ff7b1b4ea249ce46264e82b9b5a5c611896

mkllnk:
Removing the override breaks a few tests.
```
# Semaphore run 2
rspec ./spec/controllers/admin/subscriptions_controller_spec.rb:418 # Admin::SubscriptionsController cancel json as an enterprise user with authorisation when at least one associated order is still 'open' when no 'open_orders' directive has been provided renders an error, asking what to do
rspec ./spec/controllers/admin/subscriptions_controller_spec.rb:429 # Admin::SubscriptionsController cancel json as an enterprise user with authorisation when at least one associated order is still 'open' when 'keep' has been provided as the 'open_orders' directive renders the cancelled subscription as json, and does not cancel the open order
rspec ./spec/controllers/admin/subscriptions_controller_spec.rb:449 # Admin::SubscriptionsController cancel json as an enterprise user with authorisation when at least one associated order is still 'open' when 'cancel' has been provided as the 'open_orders' directive renders the cancelled subscription as json, and cancels the open order
rspec ./spec/controllers/admin/subscriptions_controller_spec.rb:517 # Admin::SubscriptionsController pause json as an enterprise user with authorisation when at least one associated order is still 'open' when no 'open_orders' directive has been provided renders an error, asking what to do
[...]

Issue #2953, opened 2018-10-30
[Spree Upgrade][Phase 2] Fix non-core features


luisramos0:
Non core features have been commented out in the build and will need to be fixed when the core functionality is ready to go.
These features are:
- reports
- subscriptions
- product import
- bulk invoices
- account invoices
- paypal integration

luisramos0:
for reference, I validated paypal integration is working in v2, I did a basic setup and managed to pay an order through a paypal sandbox account.

sauloperez:
Is this sandbox setup documented somewhere in the wiki? It'd be great to have it up to date as we did with Stripe recently.

luisramos0:
I put some details here: https://docs.google.com/document/d/15EdXKb4B3BS8gekD6_9PDdPTZsjUCAVS2011zRf55lM/edit#

PR #3012, opened 2018-11-12
[Spree Upgrade] Subs - Extend Order cleanly for subscriptions

mkllnk:
#### What? Why?

Part of #2520. It unblocks https://github.com/openfoodfoundation/openfoodnetwork/pull/2771.

Orders belonging to subscriptions get completed without payment. That
requires overriding Spree's functionality. Here we move this
customisation into its own file and override a different method to make
it less invasive and compatible with Spree 2.

In Spree 2, an order in payment state without pending payments is invalid.
Instead we skip the payment state by not requiring a payment for
automatically generated orders until the order cycle is closed.

This is the change that comes with Spree 2:
```diff
diff --git a/app/models/spree/order_decorator.rb b/app/models/spree/order_decorator.rb
index fb5cea69b..25c91568a 100644
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -336,6 +336,9 @@ Spree::Order.class_eval do
   # See commit: https://github.com/spree/spree/commit/5fca58f658273451193d5711081d018c317814ed
   # Allows GatewayError to show useful error messages in checkout
     def process_payments!
+      if pending_payments.empty?
+        raise Core::GatewayError.new Spree.t(:no_pending_payments)
+      else
         pending_payments.each do |payment|
           break if payment_total >= total
 
@@ -345,6 +348,7 @@ Spree::Order.class_eval do
             self.payment_total += payment.amount
           end
         end
+      end
     rescue Spree::Core::GatewayError => e # This section changed
       result = !!Spree::Config[:allow_checkout_on_gateway_error]
       errors.add(:base, e.message) and return result
```

#### What should we test?

- Checkout with several payment methods.
- Create a subscription.
- Let the order cycle close by passing the close-by date.
- Make sure the payment is triggered and a confirmation is sent.

#### Release notes

Changed: Payment processing logic has been prepared for future software updates.

Changelog Category: Changed

It makes our code more compatible with Spree 2.

### DIFF
--- a/app/models/extensions/order_subscriptions_extensions.rb
+++ b/app/models/extensions/order_subscriptions_extensions.rb
@@ -1,0 +1,14 @@
+# Used to prevent payments on subscriptions from being processed in the normal way.
+# Payments are skipped until after the order cycle has closed.
+module OrderSubscriptionsExtensions
+  # Override Spree method.
+  def payment_required?
+    super && !skip_payment_for_subscription?
+  end
+
+  private
+
+  def skip_payment_for_subscription?
+    subscription.present? && order_cycle.orders_close_at.andand > Time.zone.now
+  end
+end
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -1,3 +1,4 @@
+require 'extensions/order_subscriptions_extensions'
 require 'open_food_network/enterprise_fee_calculator'
 require 'open_food_network/distribution_change_validator'
 require 'open_food_network/feature_toggle'
@@ -8,6 +9,8 @@ ActiveSupport::Notifications.subscribe('spree.order.contents_changed') do |name,
 end
 
 Spree::Order.class_eval do
+  prepend OrderSubscriptionsExtensions
+
   belongs_to :order_cycle
   belongs_to :distributor, class_name: 'Enterprise'
   belongs_to :customer
@@ -345,13 +348,6 @@ Spree::Order.class_eval do
   rescue Spree::Core::GatewayError => e # This section changed
     result = !!Spree::Config[:allow_checkout_on_gateway_error]
     errors.add(:base, e.message) and return result
-  end
-
-  # Override or Spree method. Used to prevent payments on subscriptions from being processed in the normal way.
-  # ie. they are 'hidden' from processing logic until after the order cycle has closed.
-  def pending_payments
-    return [] if subscription.present? && order_cycle.orders_close_at.andand > Time.zone.now
-    payments.select {|p| p.state == "checkout"} # Original definition
   end
 
   private
--- a/spec/jobs/subscription_confirm_job_spec.rb
+++ b/spec/jobs/subscription_confirm_job_spec.rb
@@ -127,6 +127,7 @@ describe SubscriptionConfirmJob do
       before do
         allow(order).to receive(:payment_total) { 0 }
         allow(order).to receive(:total) { 10 }
+        allow(order).to receive(:payment_required?) { true }
         allow(order).to receive(:pending_payments) { [payment] }
       end
 
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -818,13 +818,22 @@ describe Spree::Order do
     end
   end
 
-  describe "finding pending_payments" do
-    let!(:order) { create(:order ) }
-    let!(:payment) { create(:payment, order: order, state: 'checkout') }
+  describe "payments" do
+    let(:payment_method) { create(:payment_method) }
+    let(:shipping_method) { create(:shipping_method) }
+    let(:order) { create(:order_with_totals) }
 
     context "when the order is not a subscription" do
-      it "returns the payments on the order" do
-        expect(order.reload.pending_payments).to eq [payment]
+      it "it requires a payment" do
+        expect(order.payment_required?).to be true
+      end
+
+      it "advances to payment state" do
+        advance_to_delivery_state(order)
+
+        order.next!
+
+        expect(order.state).to eq "payment"
       end
     end
 
@@ -835,8 +844,8 @@ describe Spree::Order do
       context "and order_cycle has no order_close_at set" do
         before { order.order_cycle.update_attributes(orders_close_at: nil) }
 
-        it "returns the payments on the order" do
-          expect(order.reload.pending_payments).to eq [payment]
+        it "requires a payment" do
+          expect(order.payment_required?).to be true
         end
       end
 
@@ -844,7 +853,7 @@ describe Spree::Order do
         before { order.order_cycle.update_attributes(orders_close_at: 5.minutes.ago) }
 
         it "returns the payments on the order" do
-          expect(order.reload.pending_payments).to eq [payment]
+          expect(order.payment_required?).to be true
         end
       end
 
@@ -852,9 +861,27 @@ describe Spree::Order do
         before { order.order_cycle.update_attributes(orders_close_at: 5.minutes.from_now) }
 
         it "returns an empty array" do
-          expect(order.reload.pending_payments).to eq []
+          expect(order.payment_required?).to be false
+        end
+
+        it "skips the payment state" do
+          advance_to_delivery_state(order)
+
+          order.next!
+
+          expect(order.state).to eq "complete"
         end
       end
+    end
+
+    def advance_to_delivery_state(order)
+      # advance to address state
+      order.shipping_method = shipping_method
+      order.next!
+
+      # advance to delivery state
+      create(:payment, order: order, payment_method: payment_method)
+      order.next!
     end
   end
 


app/models/extensions/order_subscriptions_extensions.rb
```diff
@@ -0,0 +1,14 @@
+# Used to prevent payments on subscriptions from being processed in the normal way.
+# Payments are skipped until after the order cycle has closed.
+module OrderSubscriptionsExtensions
```
mkllnk: Dear reviewers,

This file introduces a new structure. I simply didn't know where to put this file. It is not a Concern, but an Extension. It is very tightly coupled to Spree::Order and Subscription logic. So it doesn't fit into `lib`. It's not a service and it's not a model on its own. Is this the right place?

luisramos0: The name OrderSubscriptionsExtensions implies we can add more to this module. Modules and classes should have single responsibility and as such I'd name the module to what it is doing here and nothing else: OrderSubscriptionsPaymentRequired or OrderSubscriptionsProcessPayments.

re /extensions, I'd do an effort not to create another type of file in the codebase, it adds complexity that we should avoid.
I dont think it would be perfect, but I think it'd be better than this extension: a concern with a alias_method_chain on  payment_required? or better, process_payments.

I am sure @sauloperez will be thinking about spree hooks. The right solution for this would be to patch spree code to add a "before_process_payments" hook and then make this change in ofn code reimplementing the hook.

Personally, I'd go for the alias_method_chain on process_payments inside a concern.

luisramos0:
re "Orders belonging to subscriptions get completed without payment."
This is an exception and creates complexity. Can we avoid it?
Can we say subscription orders must have payments? We can inject a pending payment in the subscriptions orders based on info on the subscription.
I think this would be much better. We have to make an effort to avoid creating these exceptions in the system, they add up and make the system complex.

Matt-Yorkley:
Injecting a new "pending" payment object is basically what happens on normal orders, right? It's not a payment as such, more like a placeholder that says "this can be paid in the future". It seems like a good idea to follow the same process. Or is the issue that the payment method hasn't been chosen at that point, so a pending payment can't be created?

mkllnk:
@luisramos0 @Matt-Yorkley Thank you for your great feedback. It makes me look at the bigger picture again.

### The Spree model

If payment is not required:

* The checkout controller advances the order with `order.next`.
* The order goes from state `delivery` to `complete`.
* Done.

If payments are required:

* The user chooses a payment method.
* A payment is created and has the initial state `checkout`.
* All payments in state `checkout` are listed with `order.pending_payments`.
* The checkout controller advances the order with `order.next`.
* `order.process_payments!` is triggered before going to the `complete` state.
* The order calls `payment.process!` on each of the payments.
* If the payment succeeds, the totals are updated and the order is complete.
* Done.

### The special need of subscriptions

Subscriptions place an order and advance it to `complete` without triggering a payment. Payments are processed later in a delayed job. In the mean time, the user can call up and request to change the order before it has been paid. The card is charged only once.

This model aligns with conventional veggie box deliveries on a weekly or fortnightly basis. Customers don't want to get charged without confirmation. If they don't change the order after they got notified when the order cycle opened, we see that as consent. It avoids double charging and therefore transaction fees. Most customers also prefer being charged later than earlier.

But I'm also realising now that this looks very different from a CSA perspective. People pay the farmer early and get what the farmer can provide. Changes may be possible through additional orders with additional charges, cancelling orders, refunding or dealing with account credit. But discussing this logic needs to go through a broader inception phase.

### Possible implementations

* The current implementation overrides `order.pending_payments` to return an empty array for subscriptions before the end of an order cycle. That does the job, but blows up in Spree 2 if we remove our override of `process_payments!`. If we leave that override there, it will still work as before.

* My implementation allows to remove our override of `process_payments!`. Instead of overriding `pending_payments` it overrides `payment_required?` so that the whole payment step is skipped. This avoids a lot of complication and seems the least invasive code to me.

* We could also override `process_payments!` to explicitly deal with subscriptions. We can skip the processing for subscriptions there and call the original code otherwise. It's very similar to my implementation, just hooking in at a different method. Is that what you meant @sauloperez?

* We could somehow make the payment go in the `pending` state. Spree uses that for a two step payment. First a payment is authorised and goes to `pending`. Then the payment is actually processed and it goes to `complete` later. I haven't found the code for the second step yet. The problem is that `order.pending_payments` doesn't list those payments in state `pending` and it would blow up in Spree 2 again. This sounds like what @Matt-Yorkley suggested, but the `pending` state doesn't seem to solve the problem. It looks like we need to hook in somewhere else as well. Do you have a good idea for that @Matt-Yorkley?

* When `payment.process!` is called, it basically calls `gateway.purchase(amount)`. We could override something in the payment model to skip processing for subscriptions. But it seems to become more hidden the further we go up the call stack.

What do you think now? Do you have a preference or other good ideas?

luisramos0:
Great analysis @mkllnk we should move this to a wiki page :-)

After reading your analysis, I think we cannot change OFN core orders workflow just because of subscriptions. IMO there should be no reference to subscriptions on ofn orders code.
If we need to change order workflow or payment workflow because of subscriptions, we need to do it because it makes sense in the broader picture and as a global change to orders, not as an exception for subscriptions.
Basically, there should be no "if subscriptions" code in ofn core code but instead, if really needed, we should implement configurable endpoints that subscriptions code can use.
Another way of putting it, we have to make an effort to avoid using the word override :-)
So, I think we should explore the only implementation option that doesn't contain the word override, which is 4.
Regarding this option 4, in subscriptions, if the payment is not complete, the order is not complete. Can we make subscriptions work with incomplete orders? Do these orders need to be in the complete state? Why?
(I am sorry for asking so many questions but I believe this is very important)

Matt-Yorkley:
I was thinking of cash payments for example, that can be present on a completed order even though the payment hasn't been made yet. The enterprise user can click the "capture" button once it's been paid to trigger the payment state changing.

app/models/extensions/order_subscriptions_extensions.rb
sauloperez: > I am sure @sauloperez will be thinking about spree hooks

I'm too foreseeable :trollface: . That is exactly what I though right after reading the description. I tend to dislike `alias_method_chain` for what I shared in https://github.com/openfoodfoundation/openfoodnetwork/pull/3009#discussion_r233396535.

>  The right solution for this would be to patch spree code to add a "before_process_payments" hook and then make this change in ofn code reimplementing the hook.

Then, why not this? or any other extension hook we think best fits here?

luisramos0: With "The right solution" I meant "the solution that @sauloperez thinks is right" :-)
I think we have to explore option 4 below. Try to remove subscriptions code from order.

luisramos0:
@sauloperez yes, that is that possibility
- AND there is the possibility of making subscriptions use a new payment thing that is implemented in standard payment process, but it's not called "subscriptions something"
- AND even better, there is the possibility of making subscriptions work with the current order and payment process, with no exceptional logic.

mkllnk:
@luisramos0 asked:
> Can we make subscriptions work with incomplete orders? Do these orders need to be in the complete state? Why?

The actual requirements were the following use case:

* A subscription is created for a series of order cycles.
* At the beginning of an order cycle (when it opens) orders are placed for all subscriptions.
* Users receive order confirmations (which are triggered by completing orders).
* Users can still modify the order before the end of the order cycle.
* Users get charged when the order cycle closes.

Rob did think a lot about how to achieve this. We could have created incomplete orders at the beginning of the order cycle. But incomplete orders don't reserve stock levels. So if we want to model that an order is actually placed and the stock is reserved for the user, a complete order seems reasonable.

### Postponing this issue

I will move this PR and https://github.com/openfoodfoundation/openfoodnetwork/pull/2771 (which is blocked by this) to the non-core epic #2953. A proper solution seems a bit more complicated and I guess we can leave the current override of `pending_payments`. If that override becomes a problem, we need to revisit this.

### A better solution

I totally agree with avoiding overrides. Yes for good object oriented design. I'm not sure polymorphism is the right word for the solution here.

I'm assuming that we still want a completed order and delay the payment. Spree processes payments in payment methods. The cash payment method allows to check out without paying. Rob always saw Stripe as requirement for subscriptions, because it allows capturing the money automatically. But maybe we need a subscription payment method that decorates another payment method. It can contain the logic of delaying the payment while the order cycle is open. But once it's closed, it just calls the associated payment method. This type of composition could avoid overrides and separate the code better. But it may well be that there are some unknown obstacles that make this really difficult. Worth a try after the upgrade?

luisramos0:
i like your line of thought Maikel! i wonder how the "cash payment method allows to check out without paying" is implemented...

sauloperez:
I like it too

mkllnk:
@luisramos0 Have a look at:

https://github.com/openfoodfoundation/spree/blob/41906362d931695e0616194341a68d2c4c85aaaf/core/app/models/spree/payment_method/check.rb#L17

It basically has a capture method returning `true` without doing anything.

mkllnk:
@luisramos0 We deferred this to phase 2. Now we are in phase 2 and I just had a look at it again. What do you think we should do with these pull requests now?

The current implementation seems to work. In this pull request and #2771 I thought I could improve our way of overriding and make it more compliant with Spree 2. You and Pau pointed out that it's still overriding core logic for subscriptions and that it's not ideal. I agree. But I also don't see a reasonable way of implementing our ideal solution. I see three ways forward:

- We close these pull requests and leave everything as it is. It works. The comment about obsolete code in the decorator code may be confusing though. https://github.com/openfoodfoundation/openfoodnetwork/blob/32846af23019e4f00d0f7b5fdb1d9fb797641912/app/models/spree/order_decorator.rb#L327-L344
- I incorporate your feedback about not creating a new extension and we go with this pull request. It's not ideal but I think that it's better than what we have at the moment. We see the separation of core code and subscriptions as separate issue out of scope.
- We invest more time to try and separate subscriptions and core logic. I suspect that it may not be possible with current Spree code and that we either need to override something or start defining our own models because subscriptions require something that is restricted in the Spree world.

luisramos0:
Given where we are with the upgrade and looking at this code again, I'd go for option 2.

I dont mind the extension but I'd put those 2 short methods in the order_decorator itself. But if you prefer to keep the extension it's fine (I'd call it OrderSubscriptionsPaymentsExtension.

So, if you rename the extension (or better imo, if you move the methods to order_decorator), I am ok with going ahead with this PR. It's not perfect but definitely better than what it was.
Otherwise, if someone disagrees, I tend more to option 1 than option 3.

mkllnk:
Okay, I had another go at this. I realised several things:

1. Moving the methods into the override means that I can't use `super`. Since the implementation of `payment_required?` changes between Spree 1 and Spree 2, I thought its better to apply this change to the v2 branch instead of master.
2. Using v2 I came across merge conflicts and needed to update my spec. So it's better to do that now instead of when merging later. Otherwise you would have cursed me during the next merge. :wink: 
3. Using v2 also allows me to include the removal of `process_payments!` which was the original purpose of the endeavour.

@luisramos0 Have a look at https://github.com/openfoodfoundation/openfoodnetwork/compare/2-0-stable...mkllnk:2520-spree2-order-subscription?expand=1. Should I open a new pull request and close this one and #2771?

I think that an extension would have been cleaner but I went with your preference of using the decorator now. We will see what the next Spree upgrade requires. :wink:

luisramos0:
yeah, that's it. awesome! yes, it's not so clean to have them in the order_decorator, but it makes it more obvious we need to take this out of here soon :-)

Issue #2756, opened 2018-09-19
Impossible to duplicate a product with image


myriamboure:
This has now been reported by 2 hub users in France and seems pretty annoying.

## Description
The "duplicate product" button is not working. Click on it but nothing happens.

## Steps to Reproduce
1. Go to "product"
2. On any product, click on the icon "duplicate"

## Animated Gif/Screenshot
![capture du 2018-09-19 22-51-58](https://user-images.githubusercontent.com/8135376/45780808-a9695380-bc5e-11e8-9592-8b2a40e40ff5.png)

## Context
Reported by 2 French users, takes a long time to recreate products one by one and can't use duplicate feature. Tested on both Firefox and Chrome but users, verified by me I can reproduce the bug.

## Severity
S3 - You can create it manually, or now use product import feature.

## Your Environment

* Version used: v1.18 on French instance (not sure which version it is in Canada) so it's not linked to the last v1.19 release.
* Browser name and version: Firefox Quantum 62.0
* Operating System and version (desktop or mobile): Ubuntu 18.04.1 LTS

myriamboure:
![screenshot from 2018-09-20 12-01-22](https://user-images.githubusercontent.com/8135376/45826343-a15bf300-bcf4-11e8-9219-67a9430ea58e.png)
From @Matt-Yorkley investigation.

myriamboure:
![capture du 2018-09-20 19-50-12](https://user-images.githubusercontent.com/8135376/45836986-88603b80-bd0e-11e8-8b8c-1f8772202cf4.png)

![capture du 2018-09-20 19-50-59](https://user-images.githubusercontent.com/8135376/45836989-8c8c5900-bd0e-11e8-873a-68d08c97822a.png)

RachL:
hey @myriam "chez moi ça marche". For me it works on French production. Lousy UX though as you don't see it is loading the duplicated product... but it works :) You can see I did two copy of liqueur (yes I do choose my product wisely) for the producer "producteur test OFF"

myriamboure:
@RacheL strangely when the first user reported that I tested and it worked. Then second I retested and It didn't. With both super admin and hub manager... I will do some other test but there is definitely a bug here somewhere.

myriamboure:
I might have found the bug:
- if I am the owner of the producer enterprise behind the product, I can duplicate a product. 
- if I am not the owner, but just a manager of the producer enterprise behind the product, I can't.
[...]

PR #3022, opened 2018-11-13
Add specs for cloning a product with image

HugsDaniel:
#### What? Why?

Part of #2756 

This adds specs for cloning a product with image.
**Blocked by https://github.com/openfoodfoundation/spree/pull/19**

The issue itself is solved in Spree:https://github.com/openfoodfoundation/spree/pull/19
The implementation is from Spree 2-0 and solves the problem we had with product duplication.

#### What should we test?

After https://github.com/openfoodfoundation/spree/pull/19 is merged, specs should be green, and it should be possible to duplicate a product with an image.

#### Release notes

Fixed product with image duplication.

Changelog Category: Fixed

#### Dependencies

https://github.com/openfoodfoundation/spree/pull/19

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem 'nokogiri', '>= 1.6.7.1'
 gem 'web', path: './engines/web'
 
 gem 'pg'
-gem 'spree', github: 'openfoodfoundation/spree', branch: 'step-6a', ref: '69db1c090f3711088d84b524f1b94d25e6d21616'
+gem 'spree', github: 'openfoodfoundation/spree', branch: 'step-6a', ref: '41906362d931695e0616194341a68d2c4c85aaaf'
 gem 'spree_i18n', github: 'spree/spree_i18n', branch: '1-3-stable'
 gem 'spree_auth_devise', github: 'openfoodfoundation/spree_auth_devise', branch: 'spree-upgrade-intermediate'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -31,8 +31,8 @@ GIT
 
 GIT
   remote: https://github.com/openfoodfoundation/spree.git
-  revision: 69db1c090f3711088d84b524f1b94d25e6d21616
-  ref: 69db1c090f3711088d84b524f1b94d25e6d21616
+  revision: 41906362d931695e0616194341a68d2c4c85aaaf
+  ref: 41906362d931695e0616194341a68d2c4c85aaaf
   branch: step-6a
   specs:
     spree (1.3.99)
--- a/spec/controllers/spree/api/products_controller_spec.rb
+++ b/spec/controllers/spree/api/products_controller_spec.rb
@@ -8,6 +8,7 @@ module Spree
     let(:supplier2) { create(:supplier_enterprise) }
     let!(:product1) { create(:product, supplier: supplier) }
     let(:product_other_supplier) { create(:product, supplier: supplier2) }
+    let(:product_with_image) { create(:product_with_image, supplier: supplier) }
     let(:attributes) { [:id, :name, :supplier, :price, :on_hand, :available_on, :permalink_live] }
 
     let(:current_api_user) { build_stubbed(:user) }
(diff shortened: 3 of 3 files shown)


myriamboure:
It seems some tests are not passing @HugsDaniel :-(
One thing I don't understand, you put this as blocked, so can't we move it forward? It's a pretty annoying issue, does it have to wait for the Spree upgrade to be complete? If yes... well... we have to implement a manual support in every instance for such cases :-o We need to know to adapt ASAP.

HugsDaniel:
@myriamboure The resolution itself is in https://github.com/openfoodfoundation/spree/pull/19, these are just the tests covering this case.

The failing specs are these two, but it's expected since changes have not been merged in Spree yet. I put it in blocked because it is indeed blocked by https://github.com/openfoodfoundation/spree/pull/19.

myriamboure:
Ahhh ! Ok so the Spree PR needs to be reviewed, tested and merge but it is in the Spree repo, so doesn't appear in our zenhub board. Any idea how we could connect that repo in Zenhub as well? Devs need to know if needs to be reviewed ! @mkllnk @sauloperez how do you  handle review follow up on those type of PR?

myriamboure:
Maybe we leave this one in review ready so devs can review both together? ...

luisramos0:
All right. Just to double check the issue is really the basic case of cloning a product with an image, I went to https://staging.openfoodfrance.org and tried to clone a product with image, it fails. ✅

mkllnk:
@HugsDaniel I merged the blocking pull request. You can now update the Spree version used and rebase.

HugsDaniel:
@mkllnk I removed the last commit, rebased and push force, but it's still failing. Is it possible that it doesn't see the changes made in step-6a ?

sigmundpetersen:
Maybe you need to update ofn/spree revision used in Gemfile or something? As revision changed after openfoodfoundation/spree#19

HugsDaniel:
@sigmundpetersen how can I do that ?

sigmundpetersen:
Well I'm not really the expert here but I think updating this line https://github.com/openfoodfoundation/openfoodnetwork/blob/605f84d30ad6f62b5f58228a078c30d7479fb4f2/Gemfile#L16

@sauloperez or @luisramos0 would know best

HugsDaniel:
@luisramos0 it must have updated when I ran bundle install, I removed it to keep the same version

sauloperez:
> @HugsDaniel I am not 100% sure upgrading to bundle 1.17 will create trouble in deployment.

Correct me if I'm wrong @Matt-Yorkley but I think it won't. We should have our servers quite up-to-date as we always install the latest bundler when provisioning. See https://github.com/openfoodfoundation/ofn-install/blob/48476c4f012325350f55211820c83b8a362e0f28/playbooks/provision.yml#L45-L48 and https://github.com/openfoodfoundation/ofn-install/blob/master/roles/bundler/tasks/main.yml for reference.

Besides, as long as I've seen so far this bundler line in the Gemfile just causes a warning to make you upgrade soon.

luisramos0:
@sauloperez I was expecting Test Ready > Staging, etc... am I missing something?

sauloperez:
Aren't this just tests that actually pass? Sorry if I missed something?

luisramos0:
This is using the new change in step6-a so the fix is this PR. Right?
We need to validate the issue is fixed, right?

Maybe we can stage master now and verify?

myriamboure:
Yes, it would be great to test ;)

mkllnk:
Yes, this should have been tested, because it includes https://github.com/openfoodfoundation/spree/pull/19. Sally will test it now and then we'll deploy to production. :-)

sstead:
**Testing notes**
- I was able to successfully clone products with and without images, as a producer, a hub and a superadmin user. yay.

- One observation - cloning copies the image and product description, which seems possibly annoying, depending on the user.

https://docs.google.com/document/d/1laGMqG8WM6wtiJseTGFxIEu8wFenpsEkhAd4txMoVIo/edit#

luisramos0:
yes, cloning image was the original and current objective of the code :-)
I think it makes sense to clone the image as well.

Matt-Yorkley:

myriamboure:
It is definitely wanted that image and description is duplicated, if I clone a product it's the idea, I want a clone, if I want a different product I create a new one. Cool then !

PR #3040, opened 2018-11-13
[Spree Upgrade] Simplify completed_order_with_fees factory and make shipment_with an ofn factory

luisramos0:
#### What? Why?

There are quite a few things we can do to make these factories better. But I think it's better to wait until we have a green build. For now, this move will help get tests green because the default shipment with default shipping_method from spree is no longer in these order factories, only our shipment_with that contains a shipping_method with a configurable shipping_fee.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -281,12 +281,16 @@ FactoryBot.define do
   end
 
   factory :order_with_totals_and_distribution, parent: :order do
+    transient do
+      shipping_fee 3
+    end
+
     distributor { create(:distributor_enterprise) }
     order_cycle { create(:simple_order_cycle) }
 
-    after(:create) do |order|
-      p = create(:simple_product, :distributors => [order.distributor])
-      FactoryBot.create(:line_item_with_shipment, :order => order, :product => p)
+    after(:create) do |order, proxy|
+      p = create(:simple_product, distributors: [order.distributor])
+      FactoryBot.create(:line_item_with_shipment, shipping_fee: proxy.shipping_fee, order: order, product: p)
       order.reload
     end
   end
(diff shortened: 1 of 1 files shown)


spec/factories.rb
```diff
@@ -281,12 +281,16 @@
...
+    end
+
     distributor { create(:distributor_enterprise) }
     order_cycle { create(:simple_order_cycle) }
 
-    after(:create) do |order|
+    after(:create) do |order, proxy|
       p = create(:simple_product, :distributors => [order.distributor])
-      FactoryBot.create(:line_item_with_shipment, :order => order, :product => p)
+      FactoryBot.create(:line_item_with_shipment, shipping_fee: proxy.shipping_fee, :order => order, :product => p)
```
sauloperez: I would replace those hashrockets now that you are at it

spec/factories.rb
```diff
@@ -410,7 +422,18 @@
...
 
   factory :line_item_with_shipment, parent: :line_item do
-    target_shipment { create(:shipment, order: order) }
+    transient do
+      shipping_fee 3
+    end
+
+    target_shipment do
+      shipment = order.reload.shipments.first
+      if shipment == nil
```
sauloperez: the idiomatic way would be 
```suggestion
      if shipment.nil?
```

spec/factories.rb
```diff
@@ -410,7 +422,18 @@
   end
 
   factory :line_item_with_shipment, parent: :line_item do
-    target_shipment { create(:shipment, order: order) }
+    transient do
+      shipping_fee 3
+    end
+
+    target_shipment do
+      shipment = order.reload.shipments.first
```
sauloperez: is there any particular reason why we need a `reload` here? this will run an extra DB query so it'll add an overhead anywhere we use this factory.

spec/factories.rb
luisramos0: yep, done.

spec/factories.rb
luisramos0: Thanks a lot!!! Learning Ruby bit by bit :-)
I cant believe I couldn't write a line of Ruby 6 months ago, I can surely READ ruby fluently now :-D

spec/factories.rb
luisramos0: yes! these factories are all DB dependent :-( and this reload is really required here...

Issue #2485, opened 2018-07-19
Unfilled checkout panels default to Open in Mobile


kirstenalarsen:
## Expected Behavior
When a Customer goes to the Checkout, any panels that require information to be filled in will default to open so that it is obvious that they have to enter information.

If the Customer is already logged in or logs in, and has shopped previously, this will most likely be the Shipping and Payment panels

If they are a new Shopper it will also be the Details and Billing Info panels

## Actual Behavior
panels are closed and require Customer to click on 'expand', which is in small text and on the right hand side of the screen. So people get stuck

![image](https://user-images.githubusercontent.com/2911176/42942213-87ffef7a-8ba2-11e8-9f93-c048941b74dd.png)

## Steps to Reproduce
1. Go to Checkout on Mobile
2. See that some panels are closed and you need to click small link to open them - not obvious
3.
4.

kirstenalarsen:
See #2830 for additional information. Please apply this behaviour to both desktop and mobile

RachL:
Closing via #3047

PR #3047, opened 2018-11-14
Expand all steps in the checkout page by default

### LINKED ISSUE #2830: "next" on checkout sometimes jump checkout sections, or bring too low > UX confusing and leads to errors
**UPDATE 29.11.2018**
After PR 3047, on mobile when I click on "next" I am correctly taken to next section BUT too low, I see the  bottom of the section apparently and not the top, it's confusing.
Ex after address section when click on next:
img_9649
After shipping section when click on next:
img_9650

## Description
UX at checkout is confusing, both on desktop and mobile.
On desktop, navigating through the 4 sections is not consistent, when I click on next sometime it doesn't bring me to the next section but jump some sections. Sometimes it goes too "low" on the page and I have to scroll up to be able to fill in info. It complicates checkout and frustrates shoppers.
On mobile panels are closed and require Customer to click on 'expand', which is in small text and on the right hand side of the screen. So people get stuck.
mobile

## Expected Behavior
I can navigate easily through checkout sections both on desktop and mobile.

Proposed behavior: when a Customer goes to the Checkout, any panels that require information to be filled in will default to open so that it is obvious that they have to enter information.
If the Customer is already logged in or logs in, and has shopped previously, this will most likely be the Shipping and Payment panels
If they are a new Shopper it will also be the Details and Billing Info panels

## Actual Behavior
I can't navigate easily through checkout sections and it's hard to fill in checkout info both on desktop and mobile.

## Steps to Reproduce
On desktop and mobile
1. Put products in a cart on any shop (I tested with a registered user).
2. When you reach checkout page, fill in content of first section of checkout (if registered user and info saved, unhide the info and redo the process as if no info were saved) and click on "next", it brings you to the second section.
3. Fill in and click again on "next", it brings you directly to the 4th section, skipping the shipping method !
4. On mobile, see that some panels are closed and you need to click small link to open them - not obvious

## Animated Gif/Screenshot
Desktop behavior:
https://www.useloom.com/share/4dc286a795fe44ad847aa3be66d0e268

## Context
I regularly hit this issue, and especially today Irene from Sicily starting to use OFN hit it again in front of me and was distrurbed, she didn't understand how to checkout as she was sent directly to Payment and didn't see she had to choose a Shipping option, then got the error, etc. Pretty bad UX touching checkout.
For mobile that was also reported by Kirsten in #2485 

## Severity
S3, but pretty annoying S3 as most users are concerned at every checkout. They do with it but it's disturbing checkout.

## Your Environment

* Version used: V1.18 French instance

mllocs:
#### What? Why?

Closes #2830 and #2485 

Pretty straightforward, I still don't understand where those Angular directives (%accordion, %accordion-group...) come from though, a gem maybe? 😬 

#### What should we test?

The checkout page

#### Release notes

Expand all steps in the checkout page by default
Changelog Category: Changed


app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
```diff
@@ -1,6 +1,6 @@
 Darkswarm.controller "AccordionCtrl", ($scope, localStorageService, $timeout, $document, CurrentHub) ->
   key = "accordion_#{$scope.order.id}#{CurrentHub.hub.id}#{$scope.order.user_id}"
-  value = if localStorageService.get(key) then {} else { details: true, billing: false, shipping: false, payment: false }
+  value = { details: true, billing: true, shipping: true, payment: true }
```
mkllnk: Why did you remove the logic of keeping the current state? If parts of the form are already filled in, we don't need to expand them.

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
```diff
@@ -9,13 +9,7 @@ Darkswarm.controller "AccordionCtrl", ($scope, localStorageService, $timeout, $d
...
   $scope.show = (section)->
     $scope.accordion[section] = true
-    # If we call scrollTo() directly after show(), when one of the accordions above the
-    # scroll location is closed by show(), scrollTo() will scroll to the old location of
-    # the element. Putting this in a 50 ms timeout is enough delay for the DOM to
-    # have updated.
-    $timeout ->
-      $document.scrollTo($("##{section}"), offset_height, 500)
-    , 50
+    $document.scrollTo($("##{section}"), offset_height, 400)
```
mkllnk: Do we still need to scroll? If clicking next is collapsing the current section, the next one should be visible straight away. I think a bit of scrolling is easier to understand than random jumps in the document. Maybe it's just me, but I really don't like it when pages are scrolling for me. ;-)

luisramos0:
Welcome back @mllocs !

There's a ofn angular foundation fork here:
https://github.com/openfoodfoundation/angular-foundation/tree/master/src/accordion

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
mllocs: Sorry I don't really understand this line, looked like: if we find the key in localStorage, then the value is `{}`, otherwise we only expand details. This is the initial state of the accordion right? 

I would like to set the initial state of the accordion based on each step form being valid, so, we would only expand a step if its form is invalid (so needs the user's input). 

Could somebody guide me a bit on this? 😊 

Thanks!

mllocs: I was a bit confused with assigning `{}` when there's there's a saved state in localStorage, now I understand a bit more how it works. 😄 

I'm still working on this plus we are also discussing about some missing details, I'll push some updates in the next days!

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
mllocs: Changed that, now clicking Next won't scroll down. The code that scrollsTo a given section is only used when there are errors after submitting the form to scroll up to the step that has the error.

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
mllocs: OK, just pushed an update. So, what I propose is: instead of storing the accordion step state in localStorage, we set the open/close initial state based on each step being valid/invalid https://github.com/openfoodfoundation/openfoodnetwork/pull/3047/files#diff-22ad100d0355fdbb60c33f3143d2a215R9

So, if you already filled up some steps and come back, the steps that doesn't need user interaction will be closed 😄

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
mkllnk: Ah, thank you for the clarification. :+1:

app/assets/javascripts/darkswarm/controllers/checkout/accordion_controller.js.coffee
mkllnk: Sounds great to me! Less complexity and more deterministic behaviour. :-D

spec/features/consumer/shopping/variant_overrides_spec.rb
```diff
@@ -182,12 +181,10 @@ def complete_checkout
       fill_in "Postcode", with: "3066"
     end
 
-    toggle_shipping
     within "#shipping" do
       choose sm.name
     end
 
-    toggle_payment
```
sauloperez: Just removing this lines is a huge win for the test suite :muscle:

mllocs:
@mkllnk ready to review again ;)

spec/features/consumer/shopping/checkout_spec.rb
```diff
@@ -64,19 +64,18 @@
       let(:user) { create(:user) }
 
       def fill_out_form
-        toggle_shipping
```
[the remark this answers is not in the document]
mllocs: There are some of those toggles that are still necessary :)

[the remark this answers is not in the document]
mkllnk: Really?  Where else is it used?
```
> git fetch origin pull/3047/merge
> git checkout FETCH_HEAD
> git grep toggle_shipping
spec/support/request/checkout_workflow.rb:  def toggle_shipping
```

I can only find the definition, no usage.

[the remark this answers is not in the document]
mllocs: True, I just grepped the first one `toggle_details` which is the one that is still used 😬 
Removed the others!

sauloperez:

RachL:
Looking good! I found a bug but not related to this PR. @myriamboure or @sigmundpetersen  would you be available for an iOS testing? Especially can you try to buy Sally's expensive  (:D ) lemons and login after putting everything on your cart? On my mobile (an open source version of Android 7) the popin does a wild thing and I can't login anymore. I recently upgraded my OS so maybe it is not the case for everyone. If you have the issue as well I will open an issue for that.

Testing notes:
https://docs.google.com/document/d/1G3JWbHINjZjQnfMSYxHKC6yKBfPKyZMTlzY1CNK6xbY/edit#

sauloperez:
I restaged `master`, which includes this changes on https://staging.katuma.org so you guys can test with iOS.

luisramos0:
I put the pr-stage-es label here until katuma staging is free.

luisramos0:
ok, staging es is available.

PR #2955, opened 2018-10-30
Import Spree views for orders#edit in our codebase

HugsDaniel:
#### What? Why?
**WIP**

Closes #2956 
Related to #2416, this is the first step to move towards Spree 2.0 for orders edit page.

I imported Spree views from step-6a and integrated Deface modifications.

What remains to be done : 
- [x] Have dynamic classes work in _order_tabs.html.haml
- [x] Remove `data-hook` tags

#### What should we test?

The orders edit page is working properly. Also check the different screens in the workflow of editing orders, like choosing the distributor etc.

#### Release notes

Changelog Category: Added | Changed

#### How is this related to the Spree upgrade?

This is the first step before changing to the Spree 2.0 views

PR #3053, opened 2018-11-15
[Defacepocalypse] Import products#new view into our codebase and apply overrides

### LINKED ISSUE #3050: [Defacepocalypse] Convert products/new/replace_form.html.haml.deface to an OFN view
## Description

We need to refactor [app/overrides/spree/admin/products/new/replace_form.html.haml.deface](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/overrides/spree/admin/products/new/replace_form.html.haml.deface) in `master` to have the same behavior but using an OFN view. 

To do that we need to:

1. Bring Spree's [backend/app/views/spree/admin/products/new.html.erb](https://github.com/openfoodfoundation/spree/blob/step-6a/backend/app/views/spree/admin/products/new.html.erb) into this repo.
2. Apply OFN's [app/overrides/spree/admin/products/new/replace_form.html.haml.deface](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/overrides/spree/admin/products/new/replace_form.html.haml.deface) on top
3. Ensure all other overrides still work, which should be the ones in [app/overrides/spree/admin/products/_form](https://github.com/openfoodfoundation/openfoodnetwork/tree/master/app/overrides/spree/admin/products/_form)
3. Remove [app/overrides/spree/admin/products/new/replace_form.html.haml.deface](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/overrides/spree/admin/products/new/replace_form.html.haml.deface)

HugsDaniel:
#### What? Why?

Closes #3050

Import Spree's products#new view in our codebase and apply the overrides from `app/overrides/spree/admin/products/new/replace_form.html.haml.deface`.

#### What should we test?

The new product page should have all fields in the right place and work just fine

#### Release notes

Spree view for products#new have been imported in our codebase.

Changelog Category: Added

This is another Defacepocalypse step that will help us when going to Spree 2.0


app/views/spree/admin/products/new.html.haml
```diff
@@ -0,0 +1,117 @@
...
+      = button t('actions.create_and_add_another'), 'icon-repeat', :submit, value: 'add_another'
+      %span.or
+        = t(:or)
+      = link_to_with_icon 'icon-remove', t('actions.cancel'), admin_products_path, :class => 'button'
+
+
+:javascript
+  angular.element(document.getElementById("new_product")).ready(function() {
+    angular.bootstrap(document.getElementById("new_product"), ['admin.products']);
+  });
```
HugsDaniel: @sauloperez what should we do about this ?

sauloperez: I guess this the way to "boot" Angular, so I'd keep it. Have we solved it differently in other cases?

sauloperez:
thanks for being so quick @HugsDaniel !

app/views/spree/admin/products/new.html.haml
luisramos0: this is manual bootstrapping.
https://docs.angularjs.org/api/ng/function/angular.bootstrap
it's the same as a ng-app tag.
https://stackoverflow.com/questions/35018399/ng-app-or-angular-bootstrap

I think we can replace this with ng-app on the new_product element.
like
%fieldset.no-border-bottom{ id: "new_product", ng-app: "admin.products" }
we may need to define the correct controller in the same line like here:
app/overrides/spree/admin/products/_form/add_units_form.html.haml.deface:
    3: .variant_units_form{ 'ng-app' => 'admin.products', 'ng-controller' => 'editUnitsCtrl' }

luisramos0: can you give it a try Hugo and see if it works?
otherwise, we keep it as is...

app/views/spree/admin/products/new.html.haml
```diff
@@ -0,0 +1,117 @@
...
+        if (id.length) {
+          $('#product-from-prototype').load([ base_url, id ].join("/"));
+        } else {
+          $('#product-from-prototype').empty();
+        }
+      })
+      if (prototype_select.html() == "") {
+        prototype_select.change();
+      }
+    })(jQuery);
```
luisramos0: this belongs in an angular controller... maybe on the next PR?  🙈

app/views/spree/admin/products/new.js.erb
```diff
@@ -1,5 +1,5 @@
 <%# This chunk is just a copy of Spree's core/app/views/spree/admin/products/new.js.erb %>
-$("#new_product").html('<%= escape_javascript(render :template => "spree/admin/products/new", :formats => [:html], :handlers => [:erb]) %>');
+$("#new_product").html('<%= escape_javascript(render :template => "spree/admin/products/new", :formats => [:html], :handlers => [:haml]) %>');
 handle_date_picker_fields();
 <% unless Rails.env.test? %>
   $('.select2').select2();
```
luisramos0: can we convert this to haml?

app/views/spree/admin/products/new.html.haml
Matt-Yorkley: Applying ng-app to an element is indeed the solution. We can also use something like 
``` - content_for :app_wrapper_attrs do
		  = "ng-app='admin.products'"
	```
Which will instantiate angular on the whole page, in case we need to use it outside of that one element.

app/views/spree/admin/products/new.js.erb
HugsDaniel: It's the same thing with autocomplete [in your comment](https://github.com/openfoodfoundation/openfoodnetwork/pull/2955#issuecomment-434363720), maybe I can open an issue for these .js.erb files that have to be converted and moved ?

sauloperez: I had the same doubt but my concern is that it'll make it harder to compare it against Spree's one in the coming version. I guess however that if we convert OFN views to haml, we canvert this one too.

app/views/spree/admin/products/new.html.haml
HugsDaniel: I tried content_for at the top of the file and ng-app on the element, but it doesn't seem to work. This module is used to display the text area for description, with the two solutions I mention here it's present in the DOM but doesn't render. Could the `angular.element(document.getElementById("new_product")).ready(function()` have something to do with it ?

app/views/spree/admin/products/new.js.erb
luisramos0: yes, I think we need to get rid of erb completly.

luisramos0:
ok, looks like we agree, lets create issues for the points raised and test this PR as is.

mkllnk:

RachL:
Looking same as usual! Testing notes:

https://docs.google.com/document/d/1ioaPpY8Hzq_TZ7TGIpFdoYcSS8CwrVRAZ3wuB3DMGac/edit#

PR #3054, opened 2018-11-15
Fix/incorrect order dates

### LINKED ISSUE #3045: Order Dates are incorrect for some orders in Orders page
## Description

A number of UK users are reporting that the Order page is not showing the correct order dates for a number of orders. This is causing orders to be completely missed by some users as a number of users use the date functionality to search for orders.

Some orders are displaying with order dates that are before the order cycle has actually opened.

## Expected Behavior

The Order date in Orders page should reflect the date the order was completed.

## Actual Behavior

Dates are showing with dates previous to the date the user actually placed the order. Often this date appears to be a date before the Order Cycle actually opened. It appears that searching by Order Cycle is working correctly but searching by date is not showing the correct dates.

## Steps to Reproduce

On UK Prod this has affected 2 enterprises - Cultivate (last 2 order cycles) and Black Mountain
1. Use BOM to select orders by dates that reflect an order cycle
2. Then remove the dates and select by order cycle instead
3. Notice the difference in the order list, and in particular that the dates on some orders are not reflective of the dates that the order cycle was actually open.

## Animated Gif/Screenshot

orders_with_strange_order_dates png

This is an image showing some orders for an order cycle that actually opened on Nov 7th. The orders displayed are correct for the order cycle, but the dates are displaying incorrectly.

## Severity

* Version used: 1.20

## Possible Fix

spree_order holds 3 dates for an order. In the past I have seen issues arise when the wrong date is referenced. Might not be relevant but I thought I'd mention all the same.

sauloperez:
#### What? Why?

Closes #3045.

As agreed in the issue, I changed the orders page to use `completed_at` instead of `created_at`. As a result, that same page now shows `COMPLETED AT` in the table header instead of `ORDER DATE`. 

To make things consistent I also change BOM page to also show `COMPLETED AT`.

As we are now using always `completed_at` when "ONLY SHOW COMPLETE ORDERS" is not checked we show the following:

screenshot from 2018-11-15 16-04-31

IMO this clearly conveys the order is not yet finished. There's little value on showing the creation date even in this case.

#### What should we test?

The ordering in `/admin/orders` and `/admin/orders/bulk_management` pages involving the following cases (at least):

* An order that got created in the past, while an old OC was open, but completed now
* An order that got created now, in the current OC, and also completed now

Also using the "ONLY SHOW COMPLETE ORDERS" checbox:

* An order in cart state (the one that gets created after completing an order)

#### Release notes

Orders now display their completion date in the admin section so that orders started long ago are listed together with the orders of the same order cycle.

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/admin/line_items/controllers/line_items_controller.js.coffee
+++ b/app/assets/javascripts/admin/line_items/controllers/line_items_controller.js.coffee
@@ -26,8 +26,19 @@ angular.module("admin.lineItems").controller 'LineItemsCtrl', ($scope, $timeout,
       $scope.startDate = moment(OrderCycles.byID[$scope.orderCycleFilter].orders_open_at).format('YYYY-MM-DD')
       $scope.endDate = moment(OrderCycles.byID[$scope.orderCycleFilter].orders_close_at).startOf('day').format('YYYY-MM-DD')
 
-    RequestMonitor.load $scope.orders = Orders.index("q[state_not_eq]": "canceled", "q[completed_at_not_null]": "true", "q[completed_at_gteq]": "#{moment($scope.startDate).format()}", "q[completed_at_lt]": "#{moment($scope.endDate).add(1,'day').format()}")
-    RequestMonitor.load $scope.lineItems = LineItems.index("q[order][state_not_eq]": "canceled", "q[order][completed_at_not_null]": "true", "q[order][completed_at_gteq]": "#{moment($scope.startDate).format()}", "q[order][completed_at_lt]": "#{moment($scope.endDate).add(1,'day').format()}")
+    RequestMonitor.load $scope.orders = Orders.index(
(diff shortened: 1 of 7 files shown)


mkllnk:
@sauloperez This PR still has the wip label and is in the In Dev column. Should it go to test ready or are you still planning on modifying it?

sauloperez:
It's ready now. It wasn't obvious if the failing spec was due to this PR or a random failure. It turns out to be the latter :see_no_evil:

sauloperez:

RachL:
I think we are good here. I've managed to reproduce a strange bug that an Aus user saw previously. I will open a new issue for that.
Testing notes:
https://docs.google.com/document/d/1cNtkAYX_fQfqQz3D7uVvOZJPbGEhYKoyVg78Qv6Vlgg/edit#

PR #2932, opened 2018-10-25
2931 Refactor and prepare weight calculator for spree 2

luisramos0:
#### What? Why?

Closes #2931 

Make weight calculator respond to any object with the order attribute.
Move calculator::weight from models/open_food_network/calculator to models/calculator and remove models/open_food_network :tada:
Clean up rubocop warnings on weight and weight spec and transpec weight spec.

#### What should we test?
We can validate the weight calculator is still working correctly. No functional change.

#### Release notes
Changelog Category: Changed
Refactored Weight calculator code to prepare for Spree 2.

#### How is this related to the Spree upgrade?
This is preparation step for spree 2 and directly fixes #2931

PR #3072, opened 2018-11-17
Fix bug on calculator_decorator line_items_for where input is line_item with a nil order

### LINKED ISSUE #3025: [Spree Upgrade] Make calculator_decorator.line_items_for work with Stock::Package (1 broken test)
Similarly to what we did we the weight calculator in #2931, we need to make the calculator_decorator.line_items_for work with Stock::Package.

This will improve the spec spec/requests/checkout/failed_checkout_spec.rb:42.

luisramos0:
#### What? Why?

Take 2 on weight calculator (#2932), this time for a more fundamental change (remove line_items_for method all together).

The specific issue we address in the calculator is that the passed "object" could be a line_item that would "respond? :order" but with a nil order. In this case, the option to return [line_item] is the correct option.

#### What should we test?

Functional: we changed the way calculators work internally so we need to review all the points where calculators are used, not just weight. We need to review shipping fees, enterprise fees, and weight line items. I dont think it's possible these changes will break the calculations (the numbers), if something is wrong, things will just not work, no fees calculated, for example.

Tech: there could be a class loading issue with this PR, we need to watch in staging if there are any class loading error messages.

Specs: All specs related to calculators are green in the 2-0 branch. Fixes tests in spree upgrade issue #3025

#### Release Notes
Changelog Category: Fixed
Fixed bug in fees calculated based on weight. Bug introduced in previous release (#3072). This bug was applicable to orders with multiple line items with weight in order cycles with enterprise fees that use a weight calculator (this was validated in live environments and we didn't detect any case where this was applicable)

### DIFF
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -12,19 +12,5 @@ module Calculator
       total_weight = line_items.sum { |li| ((li.variant.andand.weight || 0) * li.quantity) }
       total_weight * preferred_per_kg
     end
-
-    private
-
-    def line_items_for(object)
-      if object.respond_to? :order
-        object.order.line_items
-      elsif object.respond_to? :line_items
-        object.line_items
-      elsif object.respond_to?(:variant) && object.respond_to?(:quantity)
-        [object]
-      else
-        raise "Unknown object type: #{object.inspect}"
-      end
-    end
   end
 end
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -10,10 +10,10 @@ feature "full-page cart", js: true do
     let!(:zone) { create(:zone_with_member) }
     let(:distributor) { create(:distributor_enterprise, with_payment_and_shipping: true, charges_sales_tax: true) }
     let(:supplier) { create(:supplier_enterprise) }
-    let!(:order_cycle) { create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], coordinator: create(:distributor_enterprise), variants: [product_tax.variants.first, product_fee.variants.first]) }
-    let(:enterprise_fee) { create(:enterprise_fee, amount: 11.00, tax_category: product_tax.tax_category) }
-    let(:product_tax) { create(:taxed_product, supplier: supplier, zone: zone, price: 110.00, tax_rate_amount: 0.1) }
-    let(:product_fee) { create(:simple_product, supplier: supplier, price: 0.86, on_hand: 100) }
+    let!(:order_cycle) { create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], coordinator: create(:distributor_enterprise), variants: [product_with_tax.variants.first, product_with_fee.variants.first]) }
+    let(:enterprise_fee) { create(:enterprise_fee, amount: 11.00, tax_category: product_with_tax.tax_category) }
+    let(:product_with_tax) { create(:taxed_product, supplier: supplier, zone: zone, price: 110.00, tax_rate_amount: 0.1) }
+    let(:product_with_fee) { create(:simple_product, supplier: supplier, price: 0.86, on_hand: 100) }
     let(:order) { create(:order, order_cycle: order_cycle, distributor: distributor) }
 
     before do
@@ -22,7 +22,7 @@ feature "full-page cart", js: true do
 
     describe "product description" do
       it "does not link to the product page" do
-        add_product_to_cart order, product_fee, quantity: 2
+        add_product_to_cart order, product_with_fee, quantity: 2
         visit spree.cart_path
         expect(page).to have_no_selector '.item-thumb-image a'
       end
@@ -33,7 +33,7 @@ feature "full-page cart", js: true do
 
       before do
         add_enterprise_fee percentage_fee
-        add_product_to_cart order, product_fee, quantity: 8
+        add_product_to_cart order, product_with_fee, quantity: 8
         visit spree.cart_path
       end
 
@@ -54,7 +54,7 @@ feature "full-page cart", js: true do
 
         before do
           add_enterprise_fee handling_fee
-          add_product_to_cart order, product_fee, quantity: 3
+          add_product_to_cart order, product_with_fee, quantity: 3
           visit spree.cart_path
         end
 
@@ -70,7 +70,7 @@ feature "full-page cart", js: true do
 
       context "when there are no admin and handling fees" do
         before do
-          add_product_to_cart order, product_fee, quantity: 2
+          add_product_to_cart order, product_with_fee, quantity: 2
           visit spree.cart_path
         end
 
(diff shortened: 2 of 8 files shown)


spec/models/calculator/weight_spec.rb
```diff
@@ -2,13 +2,13 @@
...
+    variant1 = build(:variant, weight: 10)
+    variant2 = build(:variant, weight: 20)
+    variant3 = build(:variant, weight: nil)
 
-    line_item1 = double(:line_item, variant: variant1, quantity: 1)
-    line_item2 = double(:line_item, variant: variant2, quantity: 3)
-    line_item3 = double(:line_item, variant: variant3, quantity: 5)
+    line_item1 = build(:line_item, variant: variant1, quantity: 1)
+    line_item2 = build(:line_item, variant: variant2, quantity: 3)
+    line_item3 = build(:line_item, variant: variant3, quantity: 5)
```
luisramos0: These changes from double to build are needed because "object.is_a?(Spree::LineItem)" will fail for a double.

app/models/calculator/weight.rb
```diff
@@ -12,19 +12,5 @@ def compute(object)
       total_weight = line_items.sum { |li| ((li.variant.andand.weight || 0) * li.quantity) }
       total_weight * preferred_per_kg
     end
-
-    private
-
-    def line_items_for(object)
```
mkllnk: Sorry @luisramos0, I didn't mean to just remove it. I have a similar memory to you that this was done for a reason. If I remember it correctly, it can happen that the weight calculator class is reloaded (maybe only in dev?) and it picks up the calculator class as parent without our override. It's missing the `line_items_for` implementation then. This is just a guess. I'm not sure what the actual problem was or if it even exists. One possible solution is to extract this code into a separate module that is included in the override and the weight calculator.

I'm just wondering if this change would actually be better in the master branch. What do you think?

sauloperez: But if reloading in development caused a problem with the decorators, we would have noticed by now I think. We have tons of these, although yes, `class_eval` can have nasty side effects.

luisramos0: I really dont feel like doing it now but the correct decision here is to do this in master first 😭

mkllnk: I know it's a pain, but it would really be better to do this in master first. We have proper testing there. We don't want random orders to fail after the Spree upgrade, because the calculator isn't working properly. A new day and new motivation?

luisramos0: ehehe, yes, will do this at some point. I move this PR to WIP for now.

luisramos0:
I have tested this locally, buying 3 items of a product of 5 kgs and paying enterprise fees per kg and shipping fees also per kg and with a payment method with "flexible rate" (first item one price, following items another price).

All the calculations are correct as far as I understand these things:
- 3 items x 5 kgs x 8.7eur enterprise fee = 130.5eur (I added the fee 3 times to the order cycle)
- 3 items x 5 kgs x 5.6 eur shipping fee = 84eur
-  10eur payment fee first item + 100eur x 2 payment fee 2nd and 3rd items = 210eur payment fees
screenshot 2018-11-30 at 22 36 10

luisramos0:
nice @kristinalim, thanks for your review!
You hit jackpot on this code review: finding a bug that is in production and was not detected previously 🎉 😭


I went back to reading the enterprise fees code and validated that #3072 does introduce a bug on **orders with multiple line items in order cycles with enterprise fees that use a weight calculator**.
(the method line_items_for is called with object = line_item, the current live code will return the line_item.order which is wrong, it should return the line_item itself). This problem is solved in this PR. So, I understand what you are saying and agree.

I immediately went to the main live instances to see how many cases we have, lucky we are, this is only used in FR live in the "Hub demo Open Food France" (I checked AUS, UK, FR and ES).

This was the SQL I used:
select count(*) from spree_orders where state = 'complete' and order_cycle_id in (select distinct(order_cycle_id) from exchanges ex, exchange_fees xfee, spree_calculators c where ex.id = xfee.exchange_id and xfee.enterprise_fee_id = c.calculable_id and c.calculable_type = 'EnterpriseFee' and c.type = 'Calculator::Weight');

In @myriamboure 's test on #3072, here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2932#issuecomment-436728582
the OC test order only has one line item with a kilo (the second item is only 10g). It explains why this was not seen.

I tested locally in master and this PR and it's obviously broken in master and it is working in this PR.

So, the only bad news is the bug in production.
The rest is good news:
- it's not happening to anyone
- we just need to test this PR and put it live

spec/models/spree/calculator/flat_percent_item_total_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe Spree::Calculator::FlatPercentItemTotal do
   let(:calculator) { Spree::Calculator::FlatPercentItemTotal.new }
-  let(:line_item) { instance_double(Spree::LineItem, amount: 10) }
```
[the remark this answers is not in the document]
luisramos0: I have been messing around so much with factories in spree upgrade that I have memorized many of these basics but I agree in this case it's very connected to the test :+1:
I pushed a new commit to fix this.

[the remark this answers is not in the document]
luisramos0: I had to rebase the PR, so I ended up squashing this change in the one commit.

luisramos0:
I had to rebase the PR to get the google maps pending specs out of the way...

luisramos0:
thanks @kristinalim I agree. I will do that.

luisramos0:
ok @kristinalim 
new checks done in live, release notes updated and a new commit with the spec.
ready for re-review.
thanks!

luisramos0:
@kristinalim can you please re-review this one? thanks!

spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -83,6 +83,33 @@
...
+          enterprise: order_cycle.coordinator, fee_type: 'admin') }
+
+        before do
+          product_fee.variants.first.update_attributes(unit_value: '2000.0')
+          product_tax.variants.first.update_attributes(unit_value: '5000.0')
+
+          add_enterprise_fee admin_fee
+
+          cart_service = CartService.new(order)
+          cart_service.populate(variants: { product_fee.variants.first.id => 3, product_tax.variants.first.id => 5 })
```
kristinalim: The quantity of 5 here really confused me because the cart total was only computing for 3 units. If the product factory's `on_hand` value of 3 for `product_tax` is important in this test, what do you think of specifying this attribute when defining `product_tax`?

kristinalim:
@sauloperez There have been important changes since your approval. I'll request another review from you to unmark it, and you or another dev can do the second review.

spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -83,6 +83,33 @@
...
+          cart_service = CartService.new(order)
+          cart_service.populate(variants: { product_fee.variants.first.id => 3, product_tax.variants.first.id => 3 })
+          order.update_distribution_charge!
+
+          visit spree.cart_path
+        end
+
+        it "shows the correct weight calculations" do
+          expect(page).to have_selector('#cart-detail')
+          expect(page).to have_selector '.cart-item-price',                 text: with_currency(2.86) # price + 1 * 2kg
```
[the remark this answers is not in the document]
luisramos0: Yes, pre-existing problem. I improved the variable names to product_with_tax. I tried to improve the comment as well. It's 1 eur per kilo.

sauloperez:
Nice! now it's way easier to understand.

sstead:
**Testing Notes**
- All of the fee calculators are functioning correctly - both when an order is first made, and dealing with any recalculations. I checked fees on shipping/payment methods and enterprise fees.

One possibly related glitch I noticed:
- There seemed to be a cacheing error (my guess) where a $1/kg fee was getting applied to an 'item' product which was previously a 500g product. The fee shouldn't get applied to products which are items. Seems the calculator was seeing the product as 500g still.

Pre-existing glitch I noticed:
- Got a snail page when creating a new enterprise fee when I left a field blank. We need to add a 'please fill in this field' error message instead of snail page. New bug here - https://github.com/openfoodfoundation/openfoodnetwork/issues/3307

https://docs.google.com/document/d/178DpKby-lpQMkMX_AF2UCzfXejiMlzHUVmzFU14XdFY/edit#

mkllnk:
I didn't see any errors while Sally was testing. Sally found that glitch is pre-existing. So this is ready to go.

luisramos0:
great, thanks @sstead and @mkllnk

PR #3076, opened 2018-11-18
Fix checkbox values on orders page

### LINKED ISSUE #3064: Wrong display in order page after using filters
## Description
This was reporter by an Australian user. I remember @sstead mentioned it in another issue, but I cannot find the issue anymore. At the time we did not manage to reproduced it. But now I think I've found how.

## Steps to Reproduce:
1. On order page /admin/orders?q[s]=completed_at+desc check "show unfulfilled orders only", then click on "filter orders"
2. On the same page, uncheck unfulfilled orders, and filter again
3. Now check “show complete orders only”. The page will show several times the last order :

## Severity

If you reload completely the page the display is again correct. However I don't know how many users will think about reloading their page...

## Your Environment

* Version used: v1.22
* Browser name and version: Firefow 63
* Operating System and version (desktop or mobile): Ubuntu 18 LTS

## Possible Fix
No idea :-)

Matt-Yorkley:
#### What? Why?

Closes #3064

The values of checkboxes were acting strangely when checked and then unchecked in the new angularised orders page.

#### What should we test?

Duplicate orders should not appear on the orders page when a checkbox is checked and then unchecked again.

#### Release notes

Fixed duplicate orders sometimes showing in orders page.

Changelog Category: Fixed


Matt-Yorkley:
This was a tricky one to figure out, but I think it's something to do with the value of the checkbox being 1 or 0, then being passed to angular, and then being passed to rails. Basically if the endpoint receives `'q[completed_at_not_null]'` with a value of `true` or `nil` it works fine, but it messes up the query if it receives the default `false` value from the checkbox.

luisramos0:
agree, could be a test or maybe a comment if the test is too difficult to write.

Matt-Yorkley:
@sauloperez Now that we're not going to pass `false` to the query any more, does it make sense to test that path? Or should I just add a comment?

sauloperez:
I would test the values that we do pass to `inventory_units_shipment_id_null` and `completed_at_not_null`; true and null. It shouldn't be too difficult with Jasmine.

mkllnk:
@sauloperez @luisramos0 Are you happy with this pull request?

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -141,6 +161,16 @@ module Api
...
     private
 
+    def serialized_orders(orders)
+      serialized_orders = ActiveModel::ArraySerializer.new(
+        orders,
+        each_serializer: Api::Admin::OrderSerializer,
+        root: false
+      )
+
+      JSON.parse(serialized_orders.to_json)
```
sauloperez: wouldn't to_json + parse be the same of doing nothing? just returning `serialized_orders`?

mkllnk: No, I think we want an array of hashes instead of a serializer.

sauloperez: Right, then [as_json](https://apidock.com/rails/v3.2.3/ActiveModel/Serializers/JSON/as_json) is what we need.

Matt-Yorkley: I played around with this for too long, and this was the only way I could find to get a green test. Notes:

- Doing nothing gives us objects instead of the expected json data.
- Using `#as_json` doesn't give the same output as what's returned from the server (I did try it originally) so the test fails. It gives us a hash that has symbols for keys instead of strings. 
- Subsequently using `#stringify_keys!` on the hash results in a failure as well, because the hash has nested data. The recursive version `#deep_stringify_keys!` is not available until Rails 4. 

Any other suggestions?

mkllnk: I'm okay with the current version.

sauloperez:
Staged on https://staging.openfoodnetwork.org.uk/

RachL:
Unable to reproduce the bug and all others filters are working fine :) This is ready https://docs.google.com/document/d/1m71XnDHKspLrDHhze-GHOcd5c258oFaFCLVD8Rbx5Qw/edit#

PR #3080, opened 2018-11-18
Pi/french issue

### LINKED ISSUE #2951: Product import not working on French instance with simple file
## Description

A new user has some excel files to populate to create their catalog in OFN. I did try to do the first import for them, and it is not working, and no error displays.

## Expected Behavior

I can upload a list of product with product import feature.

## Actual Behavior

I can't.

## Steps to Reproduce

1. On french instance, go to https://www.openfoodfrance.org/admin/product_import
2. Try to import a simple file like https://drive.google.com/open?id=1JKDfEcb-VhSL_wqEhXZl2nFoWkFKhIcF of course make sure the producer already exists
3. See the error, the file can't be uploaded, and no error message is displayed

## Animated Gif/Screenshot

capture du 2018-10-25 22-17-13

## Context

A new promising user on OFFrance want to upload their producer catalog.

## Severity
S3 - S2, you can still create manually products, but for hundreds of products it becomes complicated. Here I put S3 for now, we start with only one list that if it doesn't work we can fill in manually, it's only around 30 products...

Matt-Yorkley:
#### What? Why?

Closes #2951

An instance configuration on French production was causing unexpected product validation messages when creating a new product. This setting was forcing all products to require a tax_category to be set, and added an extra validation error into product objects at an unexpected point in the import process.

#### What should we test?

Importing products from CSV on an instance with this checkbox checked in the config:
screenshot from 2018-11-18 20-31-40
should show an error message if there is no tax_category set in the CSV file.

#### Release notes

Fixed tax category validations in product import for instances that enforce mandatory tax categories.

Changelog Category: Fixed

### DIFF
--- a/app/models/product_import/product_importer.rb
+++ b/app/models/product_import/product_importer.rb
@@ -100,7 +100,8 @@ module ProductImport
         entries[entry.line_number] = {
           attributes: entry.displayable_attributes,
           validates_as: entry.validates_as,
-          errors: entry.invalid_attributes
+          errors: entry.invalid_attributes,
+          product_validations: entry.product_validations
         }
       end
       entries.to_json
--- a/app/views/admin/product_import/_errors_list.html.haml
+++ b/app/views/admin/product_import/_errors_list.html.haml
@@ -7,3 +7,5 @@
       ( {{entry.attributes.display_name}} )
   %p.error{ng: {repeat: "(attribute, error) in entry.errors", show: "ignore_fields.indexOf(attribute) < 0" }}
     &nbsp;-&nbsp; {{error}}
+  %p.error{ng: {repeat: "(attribute, error) in entry.product_validations"}}
+    &nbsp;-&nbsp; {{error}}
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2395,6 +2395,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   validation_msg_relationship_already_established: "^That relationship is already established."
   validation_msg_at_least_one_hub: "^At least one hub must be selected"
   validation_msg_product_category_cant_be_blank: "^Product Category cant be blank"
+  validation_msg_tax: "^Tax Category is required"
   validation_msg_tax_category_cant_be_blank: "^Tax Category can't be blank"
   validation_msg_is_associated_with_an_exising_customer: "is associated with an existing customer"
   content_configuration_pricing_table: "(TODO: Pricing table)"


app/models/product_import/entry_validator.rb
```diff
@@ -243,7 +243,7 @@ def mark_as_new_product(entry)
       new_product.assign_attributes(entry.attributes.except('id'))
       new_product.supplier_id = entry.producer_id
 
-      if new_product.valid?
+      if new_product.valid? && new_product.errors.empty?
```
sauloperez: I don't understand this extra check. If the model is valid it won't have errors in the array (unless we add them ourselves outside of validations).

app/models/product_import/product_importer.rb
```diff
@@ -100,7 +100,8 @@ def entries_json
         entries[entry.line_number] = {
           attributes: entry.displayable_attributes,
           validates_as: entry.validates_as,
-          errors: entry.invalid_attributes
+          errors: entry.invalid_attributes,
+          product_validations: entry.product_validations
```
sauloperez: I know it's not strictly related to this PR, but IMO the JSON serialization should rather be extracted into a serializer. We could create an issue for it, if we all agree.

Matt-Yorkley: Issue added: https://github.com/openfoodfoundation/openfoodnetwork/issues/3105

app/models/product_import/entry_validator.rb
Matt-Yorkley: Ah, I think I added these whilst debugging because the outcome wasn't making sense. I've removed them now. :+1:

luisramos0:
staged: https://staging.openfoodfrance.org

luisramos0:
@Matt-Yorkley @sauloperez on staging this PR I noticed the commit messages of this PR: they are very specific and with no reference to the context. I wonder if this is ok or if we should always include context in every commit message. I ask this because I am not sure what's the best approach. You opinion?
Maybe an opportunity to improve this guide:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Making-a-great-commit

sauloperez:
I think they are fine and they stick to the those guidelines, which I think are quite good.

In my experience a great way to see this is from the perspective of someone new to the app. Is it easy to trace back to the decisions behind the code changes? Aka. Our experience with the Spree upgrade investigations.

luisramos0:
ok, thanks. I am glad I asked. Maybe I need to adapt some of my commit messages.

RachL:
@Matt-Yorkley @luisramos0 PR is working fine but the error message was not translated:
image

Is it linked to the PR or to staging?
Thx :)
Testing notes:
https://docs.google.com/document/d/1yu4UVizLSR6TcPkDD_Jyp28km1GGuJ3ktoog9ucNDqA/edit#

luisramos0:
@RachL it's how it works, first we get this tested and then merged and then validation_msg_tax will get to transifex from master, and after translated, back to master on a transifex merge.

RachL:
@luisramos0 yup

PR #3086, opened 2018-11-20
[Headless Chrome] - Travis config WIP

luisramos0:
Extracted travis wip commits from the headless chrome branch.
What do we do with these @mkllnk ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,6 @@ before_script:
   - cp config/application.yml.example config/application.yml
   - RAILS_ENV=test bundle exec rake db:create db:schema:load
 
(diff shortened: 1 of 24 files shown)


luisramos0:
closing, used wrong source branch...

luisramos0:
hey @mkllnk is your comment above a new issue? do we really need to do that or is it a nice to have?

mkllnk:
I was assuming that we need to change the Semaphore configuration to make this build work. But now I read all the other threads and it looks like we don't need to do it. If Semaphore just works, we don't need to do anything, great. :-)

PR #3087, opened 2018-11-20
[Headless Chrome] - Travis config WIP

luisramos0:
Extracted travis wip commits from the headless chrome branch.
What do we do with these @mkllnk ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,6 @@ before_script:
   - cp config/application.yml.example config/application.yml
   - RAILS_ENV=test bundle exec rake db:create db:schema:load
 
-  # Only install PhantomJS if it is not already present (ie. cached)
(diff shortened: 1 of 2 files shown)


mkllnk:
All this is not needed for Semaphore? I would still merge it in case we ever switch back to Travis. You never know, it's a moving market and it doesn't hurt to have these changes in the code.

sauloperez:
but if we move back to travis chances are that the config changed and we might keeping useless code. I'd say this is just YAGNI.

mkllnk:
I didn't want to give up Travis, because it's much easier to set up for other contributors. But if you think it's YAGNI, we should delete .travis.yml alltogether. It's broken now anyway.

mkllnk:
And this pull request will be here in case we want to know how to use headless chrome on Travis with the OFN. Closing it.

luisramos0:
:+1:

PR #3082, opened 2018-11-19
Remove all database.travis.yml

sauloperez:
#### What? Why?

We no longer use this file since we started with Semaphore CI.

#### What should we test?

No change on behavior.

#### Release notes
Removed old and unused database configuration for Travis CI.

Changelog Category: Removed

PR #3088, opened 2018-11-20
Run RSpec tests in custom engines (and some fixes, cleanup)

### LINKED ISSUE #3085: RSpec tests in custom engines are not being run
The `rspec` command which we use does not run tests of custom engines in `engines/web`.

## Description

#2521 (following [this discussion](https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377)) introduced custom engines representing domains. That added the new "web" engine in `engines/web`. However, the custom engine is not covered by the `rspec` command which we run and is used by the official CI accounts.

Tests in custom engines should be run.

## Expected Behavior

1. `GETTING_STARTED` should document testing the custom engines.
2. The Travis CI configuration file should run RSpec tests in the custom engines.

## Actual Behavior

These are not the case.

## Steps to Reproduce

Only for developers:

1. Introduce a failing test in `engines/web/spec/` and push to Semaphore and Travis.
2. There are no failing tests in Semaphore and Travis.

## Context

I added a new custom engine "Order Management" for #2616 and noticed that the engine tests were not covered by our CI configurations.

## Severity

kristinalim:
#### What? Why?

Closes #3085 

RSpec tests of the engine "web" are not being run by the commands we use.

This PR includes the following changes:

1. Allow tests in "web" engine to be run separately. Using the main app's `spec_helper.rb` for now, while we are still slowly moving to Rails engines.
2. Add Rake task `openfoodnetwork:specs:engines:rspec` to run RSpec tests for all engines.
3. Add Rake tasks `openfoodnetwork:specs:engines:ENGINE_NAME:rspec` to run RSpec tests of a specific engine.

For your reference, here are Semaphore builds which already run the engine tests on my fork:

- `cleanup-test_scripts` - https://semaphoreci.com/kristinalim/openfoodnetwork/branches/cleanup-test_scripts/builds/14
- `cleanup-test_scripts-failure` which adds a failing engine test - https://semaphoreci.com/kristinalim/openfoodnetwork/branches/cleanup-test_scripts-failure/builds/5

#### What should we test?

Only for developers:

1. Introduce a failing test in `engines/web/spec/`.
2. The failure should reflect in Semaphore.

The `.travis.yml` configuration file currently [does not build successfully](https://github.com/openfoodfoundation/openfoodnetwork/pull/3082#pullrequestreview-176609809).

#### Release notes

- Allow tests in "web" engine to be run with `bundle exec rspec` from the engine path.
- Add Rake task `openfoodnetwork:specs:engines:rspec` to run RSpec tests for all engines.
- Add Rake tasks `openfoodnetwork:specs:engines:ENGINE_NAME:rspec` to run RSpec tests of a specific engine.

Changelog Category: Fixed

#### Documentation updates

1. [Continuous Integration](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Continuous-Integration) - One of the jobs should be updated to run the Rake task `openfoodnetwork:specs:engines:rspec`.

#### After merge

1. Communicate this in `#dev` Slack channel.
2. Update project settings in Semaphore official account to run the Rake task `openfoodnetwork:specs:engines:rspec`.

### DIFF
--- a/lib/tasks/specs.rake
+++ b/lib/tasks/specs.rake
@@ -10,7 +10,7 @@ namespace :openfoodnetwork do
       end
 
       def execute_rspec_for_engine(engine_path)
-        system "cd #{engine_path.expand_path} && bundle exec rspec"
+        system "cd #{engine_path.expand_path} && DISABLE_KNAPSACK=true bundle exec rspec"
       end
 
       engine_paths = detect_engine_paths
--- a/lib/tasks/specs.rake
+++ b/lib/tasks/specs.rake
@@ -25,6 +25,20 @@ namespace :openfoodnetwork do
           end
         end
       end
+
+      namespace :all do
+        task :rspec do
+          success = true
+
+          engine_paths.each do |engine_path|
+            success = !!execute_rspec_for_engine(engine_path) && success
+          end
+
+          abort "Failure encountered when running tests for engines" unless success
+        end
+      end
+
+      task rspec: "all:rspec"
     end
   end
 end
--- a/lib/tasks/specs.rake
+++ b/lib/tasks/specs.rake
@@ -19,6 +19,7 @@ namespace :openfoodnetwork do
         engine_name = engine_name_for_engine(engine_path)
 
         namespace engine_name do
+          desc "Run RSpec tests for engine \"#{engine_name}\""
           task :rspec do
             success = execute_rspec_for_engine(engine_path)
             abort "Failure when running tests for engine \"#{engine_name}\"" unless success
@@ -27,6 +28,7 @@ namespace :openfoodnetwork do
       end
 
       namespace :all do
+        desc "Run RSpec tests for all engines"
         task :rspec do
           success = true
 
@@ -38,6 +40,7 @@ namespace :openfoodnetwork do
         end
       end
 
+      desc "Alias for openfoodnetwork:specs:engines:all:rspec"
       task rspec: "all:rspec"
     end
   end
--- a/lib/tasks/specs.rake
+++ b/lib/tasks/specs.rake
@@ -1,0 +1,30 @@
+namespace :openfoodnetwork do
+  namespace :specs do
+    namespace :engines do
+      def detect_engine_paths
+        Pathname("engines/").children.select(&:directory?)
+      end
+
+      def engine_name_for_engine(engine_path)
+        engine_path.basename.to_path
+      end
+
+      def execute_rspec_for_engine(engine_path)
+        system "cd #{engine_path.expand_path} && bundle exec rspec"
+      end
+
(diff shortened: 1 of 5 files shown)


kristinalim:
@sauloperez As far as I understand, the motivation behind packaging code into engines is to make them work independently. Running them in a single `rspec` process would not expose places where their libraries are still tightly coupled.

Although, I think at the start, most engines would not be able to stand on their own and could therefore still use the main application's `spec_helper.rb` which includes the main application (and all the other engines). For example, running `bundle exec rspec` in "web" raises `NoMethodError:
  undefined method `scenario' for #<Class:...>`, and it does not define nor require `BaseController` which it uses.

Do you think this is okay?

sauloperez:
I understand what you say. My point was just implementation. Instead of adding 3 new bash scripts I would give it a try at sticking to rake to solve what `script/specs/rspec_engines` covers.

kristinalim:
Thanks, I misunderstood! Yes, that sounds great. So a non `:environment` Rake task, for ruby scripting instead of bash.

sauloperez:
> So a non :environment Rake task, for ruby scripting instead of bash.

If they don't need to load the environment :+1:, but as you said they will not be able to stand on their own, right? so we might need it at least for now. In any case, rake supports both ways.

daniellemoorhead:
Hey @kristinalim 👋 

No progress on this for 22 days...I'm wondering if this needs to be closed for now and moved back to Dev Ready until your to do list is clear again?

#toomanythingsinthepipe
#toottoot

kristinalim:
Thanks @daniellemoorhead! It so happened that I finished the two other PRs in my In Dev set, so I worked on this again.

@sauloperez I changed this to use a Rake task instead of a script in `script/`. I also updated the issue description, so please read again if you're reviewing this.

The Code Climate check failure would be fixed with a rebase after #3223 is merged.

lib/tasks/specs.rake
```diff
@@ -0,0 +1,47 @@
+namespace :openfoodnetwork do
```
luisramos0: @sauloperez we wanted to get rid of this namespace in the rake tasks right?
it was discussed in @mkllnk's sample_data PR #3209. we should keep both tasks coherent in terms of namespaces. do you guys know what's the best approach here?

sauloperez: Thanks for pointing out. I thought about it but I prefered to not annoy again with that topic. My opinion is that we should get rid of it.

luisramos0: lets get rid of it then :-D

mkllnk: Let's have the discussion on https://community.openfoodnetwork.org/t/rake-task-naming/1531.

PR #3095, opened 2018-11-20
In admin/users, convert Spree.t calls to t calls and use lazy lookup

luisramos0:
Moved translations from spree to ofn's en.yml

This originated in a broken test in Spree Upgrade related to translations.

#### What should we test?
The translations in admin users screens: mostly list and edit. Only visible in other languages after merge and transifex. So, the best test we can do here is to verify the functionality of the page.

#### Release notes
Changelog Category: Fixed
Admin users screen are now using OFN's translations, not spree translations.

This will fix spec/features/admin/users_spec.rb:23 in the upgrade build.

### DIFF
--- a/app/views/spree/admin/users/_form.html.haml
+++ b/app/views/spree/admin/users/_form.html.haml
@@ -1,11 +1,11 @@
 .row
   .alpha.five.columns
     = f.field_container :email do
-      = f.label :email, Spree.t(:email)
+      = f.label :email, t(".email")
       = f.email_field :email, class: "fullwidth"
       = error_message_on :user, :email
     .field
-      = label_tag nil, Spree.t(:roles)
+      = label_tag nil, t(".roles")
       %ul
         - @roles.each do |role|
           %li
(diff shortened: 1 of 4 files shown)


app/views/spree/admin/users/_form.html.haml
```diff
@@ -1,26 +1,26 @@
 .row
   .alpha.five.columns
     = f.field_container :email do
-      = f.label :email, Spree.t(:email)
+      = f.label :email, t(".email")
```
sauloperez: sorry, what was the difference? I recall having a discussions about it but I don't remember.

luisramos0: the main diff here is the lazy loading instead of using en.yml root entries.
the Spree.t thing, from @mkllnk 's wisdom through slack:

"I just looked up what the difference is: `core/lib/spree/i18n.rb`
    
```
def translate(*args)
      @virtual_path = virtual_path

      options = args.extract_options!
      options[:scope] = [*options[:scope]].unshift(:spree)
      args << options
      super(*args)
    end
    alias_method :t, :translate
```

So we can use `Spree.t` if we want to translate something in the Spree namespace. For example `Spree.t('settings')` is the same as `t('spree.settings')`. But we want to get away from all Spree translations in the long term."

the conclusion is that we should not use Spree.t at all.

luisramos0:
there was a typo in en.yml file breaking the build. it's fixed now.

app/views/spree/admin/users/_form.html.haml
luisramos0: I have added this here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)#development
under "Add missing Spree translations"

luisramos0:
staged in http://staging.katuma.org

RachL:
Deleting a user is not working but it is a pre-existing bug / wrong UX so I would say we are good to go here as well.
Testing notes:
https://docs.google.com/document/d/1802xrFccxg2YUob5Dpvlfj0cPFSQAHRihmpbAJgVrVQ/edit#

Issue #3068, opened 2018-11-16
Inventory page says "on demand" in scenario where variant is not on demand for shop


kristinalim:
Inventory page says "on demand" in scenario where variant is not on demand for shop.

## Description

With the following settings:

* Producer - On Demand: `true`, Count on Hand: 10
* Variant Override - On demand: false, Count on Hand: blank

The effective stock level is the producer's count on hand of 10. But the Inventory page still says that stock is "On Demand".

## Expected Behavior

PENDING

## Actual Behavior

PENDING

## Steps to Reproduce

Signed in as producer:

1. Create a product.
2. Set the variant to have 0 count on hand, and save.
3. Set the variant to be "on demand", and save. (Step 3 should no longer be relevant after this, but it is. A different issue will be created for this bug, but meanwhile we want to take advantage of it for testing.)
4. Grant distributor permission "to add to order cycle" and "to add products to inventory".

Signed in as the distributor:

1. Add a payment method and a shipping method to the distributor.
2. In "Inventory Settings", enable this: "New products must be added to my inventory before they can be put into my shopfront".
2. Add the variant to the inventory:
    a. Go to Products > Inventory.
    b. Click "Review Now" to see the list of products that can be added.
    c. Click "Add" for the variant.
3. Create an ongoing order cycle, and include the variant.
4. With `true` "on demand" - Using: VO setting of "on demand" (true)
    a. Go to Products > Inventory.
    b. Tick the "on demand" checkbox, and save. The "on hand" status says "On demand".
    e. The variant appears in the shop.
5. With `false` "on demand" - Using: VO setting of "on demand" (false), supplier setting of "count on hand" (0)
    a. Go to Products > Inventory.
    b. Untick the "on demand" checkbox, and save. The "on hand" status still says "On demand", but this is not the case.
    e. The variant does not appear in the shop.

## Context

Related to #3058.

## Severity

bug-s2

## Possible Fix

PENDING

myriamboure:
I think the fix for this bug had been merged, if I'm mistaken @kristinalim please reopen.

Issue #3067, opened 2018-11-16
Hidden "count on hand" of on demand variant should never be relevant


kristinalim:
The producer "count on hand" of an "on demand" variant should never be relevant. But this is used when the variant override has false "on demand" and blank "count on hand".

## Description

When the variant override has false "on demand" and a blank "count on hand", the stock check looks at the producer's "count on hand" setting. But this value is no longer visible in the producer's UI when the producer's setting for "on demand" is checked. This became a problem in #3058.

See the table below. In scenario 4, the producer's "count on hand" setting is used when checking the stock even if it is supposed to be "on demand".

|  | VO | VO | Supplier | Supplier | Final | Final |
|-------|-------|-------|-------|-------|-------|-------|
|  | On Demand | On Hand | On Demand | On Hand | On Demand | On Hand |
|     1 | false |    10 | false |    20 | false |    10 |
|     2 | false |    10 |  true |    20 | false |    10 |
|     3 | false |   nil | false |    20 | false |    20 |
| 4 >>  | false |   nil |  true |    20 | false |    20 |
|     5 |   nil |    10 | false |    20 | false |    10 |
|     6 |   nil |    10 |  true |    20 | false |    10 |
|     7 |   nil |   nil | false |    20 | false |    20 |
|     8 |   nil |   nil |  true |    20 |  true |   nil |
|     9 |  true |    10 | false |    20 |  true |   nil |
|    10 |  true |    10 |  true |    20 |  true |   nil |
|    11 |  true |   nil | false |    20 |  true |   nil |
|    12 |  true |   nil |  true |    20 |  true |   nil |

The code used to generate this table is in [this comment](https://github.com/openfoodfoundation/openfoodnetwork/issues/3067#issuecomment-446519979).

## Expected Behavior

The producer's "count on hand" setting should never be used when the producer's setting for "on demand" is `true`.

## Actual Behavior

This is not the case when the variant override has `false` "on demand" and a blank "count on hand".

## Steps to Reproduce

See same section in #3066.

## Context

Related to #3058.

## Severity

Some users might not realize that their inventories are using the 0 "count of hand" of the producer, like what happened in #3058. This can result to variants unintentionally being hidden from the shopfront.

bug-s2

## Possible Fix

PENDING

Matt-Yorkley:
[...]

PR #3104, opened 2018-11-21
Less ambiguous field for "on demand" in the Inventory page

### LINKED ISSUE #3066: "Use supplier's 'on demand'" is not represented in the inventory page
"Use supplier's 'on demand'" is not represented in the inventory page.

## Description

A variant override may have any of these three values for "on demand":

1. `nil` - Use the supplier's setting
2. `true` - Override the supplier's setting with `true`
3. `false` - Override the supplier's setting with `false`

After an inventory item has been added:

1. Creating the variant override without clicking the "on demand" checkbox, e.g. by setting the price, creates the variant override with a `nil` setting. This is displayed as an unticked checkbox.
2. Ticking the checkbox sets this to true.
3. Unticking the checkbox sets this to `false`.

There are two problems here:

1. There is no way to tell between `nil` and `false` - both are represented by an unticked checkbox.
2. There is no way to go back to `nil`.

## Expected Behavior

1. User should be able to know if the setting is `nil` or `false`.
2. User should be able to go back to `nil`.

## Actual Behavior

User cannot.

## Steps to Reproduce

Signed in as producer:

1. Create a product.
2. Set the variant to have 0 count on hand, and save.
3. Set the variant to be "on demand", and save. (Step 3 should no longer be relevant after this, but it is. A different issue will be created for this bug, but meanwhile we want to take advantage of it for testing.)
4. Grant distributor permission "to add to order cycle" and "to add products to inventory".

Signed in as the distributor:

1. Add a payment method and a shipping method to the distributor.
2. In "Inventory Settings", enable this: "New products must be added to my inventory before they can be put into my shopfront".
2. Add the variant to the inventory:
    a. Go to Products > Inventory.
    b. Click "Review Now" to see the list of products that can be added.
    c. Click "Add" for the variant.
3. Create an ongoing order cycle, and include the variant.
4. With inventory item but no variant override:
    a. Go to Products > Inventory.
    b. Notice that the "on demand" checkbox is unticked.
    c. Go to the shop page. The variant would appear.
5. With variant override but `nil` "on demand" - Using: supplier setting of "on demand" (true)
    a. Go to Products > Inventory.
    b. Override the price for the variant, and save.
    c. The variant appears in the shop.
6. With `true` "on demand" - Using: VO setting of "on demand" (true)
    a. Go to Products > Inventory.
    b. Tick the "on demand" checkbox, and save.
    e. The variant appears in the shop.
7. With `false` "on demand" - Using: VO setting of "on demand" (false), supplier setting of "count on hand" (0)
    a. Go to Products > Inventory.
    b. Untick the "on demand" checkbox, and save.
    e. The variant does not appear in the shop.

This demonstrates that the unticked checkbox in 5 and 7 are different. And, there is no way to go back to scenario 5.

## Context

Related to https://github.com/openfoodfoundation/openfoodnetwork/issues/3058.

## Severity

bug-s2

## Possible Fix

We could have the checkbox to a SELECT field with three options:

1. `nil` - "Use producer setting"
2. `true` - "On Demand"
3. `false` - "Limited Stock"

kristinalim:
Change the UI control for "on demand" in the Inventory page from a checkbox to a SELECT field with three options: `nil`, `true`, and `false`.

#### What? Why?

Closes #3066

As discussed in the issue, this addresses the following problems:
* There is no way to tell between `nil` and `false` - both are represented by an unticked checkbox.
* There is no way to go back to `nil` from a `true` or `false` value.

#### What should we test?

Signed in as producer:

1. Create a product with first variant.
2. Grant distributor permission "to add products to inventory".

Signed in as the distributor, already with payment and shipping methods:

1. Add the variant to the inventory:
    a. Go to Products > Inventory.
    b. Click "Review Now" to see the list of products that can be added.
    c. Click "Add" for the variant.
2. Back in the Inventory page, under "On Demand", check that the selected value is ~blank~ "Use producer settings".
3. Select ~"On Demand"~ "Yes" and save. After that is successful, check that the selected value is still ~"On Demand"~ "Yes", both on the current page and when loading the Inventory page in another browser tab.
4. Repeat Step 3, but for ~"Limited Stock"~ "No".
5. Repeat Step 3, but for ~the blank~ "Use producer settings" option.

#### Release notes

- Update the UI for "On Demand" and "Count on Hand" in the Inventory page to allow display and selection of "Use producer stock settings", and to encourage compatible "On Demand" and "Count on Hand".

Changelog Category: Fixed

#### Dependencies

This is best merged together with #3198.

### DIFF
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -13,6 +13,10 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
   $scope.RequestMonitor = RequestMonitor
   $scope.selectView = Views.selectView
   $scope.currentView = -> Views.currentView
+  $scope.onDemandOptions = [
+    { description: t('js.yes'), value: true },
+    { description: t('js.no'), value: false }
+  ]
 
   $scope.views = Views.setViews
     inventory:    { name: t('js.variant_overrides.inventory_products'), visible: true }
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -14,8 +14,8 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
   $scope.selectView = Views.selectView
   $scope.currentView = -> Views.currentView
   $scope.onDemandOptions = [
-    { description: t('js.yes'), value: true },
-    { description: t('js.no'), value: false }
+    { description: t('js.variant_overrides.on_demand.yes'), value: true },
+    { description: t('js.variant_overrides.on_demand.no'), value: false }
   ]
 
   $scope.views = Views.setViews
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -110,6 +110,16 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
       .error (data, status) ->
         $timeout -> StatusMessage.display 'failure', $scope.updateError(data, status)
 
+  $scope.countOnHandPlaceholder = (variant, hubId) ->
+    variantOverride = $scope.variantOverrides[hubId][variant.id]
+
+    if variantOverride.on_demand
+      t('js.variants.on_demand.yes')
+    else if variantOverride.on_demand == false
+      ''
+    else
+      variant.on_hand
+
   $scope.clearCountOnHandUnlessLimitedStock = (hubId, variantId) ->
     variantOverride = $scope.variantOverrides[hubId][variantId]
     unless variantOverride.on_demand == false && variantOverride.count_on_hand?
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -131,9 +131,9 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
     variantOverride = $scope.variantOverrides[hubId][variant.id]
 
     suggested = $scope.countOnHandSuggestion(variant, hubId)
-    unless suggested == variantOverride.count_on_hand
-      variantOverride.count_on_hand = suggested
-      DirtyVariantOverrides.set hubId, variant.id, variantOverride.id, 'count_on_hand', suggested
+    return if suggested == variantOverride.count_on_hand
+    variantOverride.count_on_hand = suggested
+    DirtyVariantOverrides.set hubId, variant.id, variantOverride.id, 'count_on_hand', suggested
 
   # Suggest producer count_on_hand if variant has limited stock and variant override forces limited
   # stock. Otherwise, clear whatever value is set.
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -110,6 +110,10 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
       .error (data, status) ->
         $timeout -> StatusMessage.display 'failure', $scope.updateError(data, status)
 
+  # Variant override count_on_hand field placeholder logic:
+  #     on_demand true  -- Show "On Demand"
+  #     on_demand false -- Show empty value to be set by the user
+  #     on_demand nil   -- Show producer on_hand value
   $scope.countOnHandPlaceholder = (variant, hubId) ->
     variantOverride = $scope.variantOverrides[hubId][variant.id]
 
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -110,12 +110,6 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
       .error (data, status) ->
         $timeout -> StatusMessage.display 'failure', $scope.updateError(data, status)
 
-  $scope.selectLimitedStockIfCountOnHandSet = (hubId, variantId) ->
-    variantOverride = $scope.variantOverrides[hubId][variantId]
-    if variantOverride.count_on_hand? && variantOverride.count_on_hand != '' && variantOverride.on_demand != false
-      variantOverride.on_demand = false
-      DirtyVariantOverrides.set hubId, variantId, variantOverride.id, 'on_demand', false
-
   $scope.clearCountOnHandUnlessLimitedStock = (hubId, variantId) ->
     variantOverride = $scope.variantOverrides[hubId][variantId]
     unless variantOverride.on_demand == false && variantOverride.count_on_hand?
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -109,3 +109,15 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
         StatusMessage.display 'success', t('js.variant_overrides.stock_reset')
       .error (data, status) ->
         $timeout -> StatusMessage.display 'failure', $scope.updateError(data, status)
+
+  $scope.selectLimitedStockIfCountOnHandSet = (hubId, variantId) ->
+    variantOverride = $scope.variantOverrides[hubId][variantId]
+    if variantOverride.count_on_hand? && variantOverride.count_on_hand != '' && variantOverride.on_demand != false
+      variantOverride.on_demand = false
+      DirtyVariantOverrides.set hubId, variantId, variantOverride.id, 'on_demand', false
+
+  $scope.clearCountOnHandUnlessLimitedStock = (hubId, variantId) ->
+    variantOverride = $scope.variantOverrides[hubId][variantId]
+    unless variantOverride.on_demand == false && variantOverride.count_on_hand?
+      variantOverride.count_on_hand = null
+      DirtyVariantOverrides.set hubId, variantId, variantOverride.id, 'count_on_hand', null
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -120,8 +120,20 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
     else
       variant.on_hand
 
-  $scope.clearCountOnHandUnlessLimitedStock = (hubId, variantId) ->
-    variantOverride = $scope.variantOverrides[hubId][variantId]
-    unless variantOverride.on_demand == false && variantOverride.count_on_hand?
-      variantOverride.count_on_hand = null
-      DirtyVariantOverrides.set hubId, variantId, variantOverride.id, 'count_on_hand', null
+  # This method should only be used when the variant override on_demand is changed.
+  #
+  # Change the count_on_hand value to a suggested value.
+  $scope.updateCountOnHand = (variant, hubId) ->
+    variantOverride = $scope.variantOverrides[hubId][variant.id]
+
+    suggested = $scope.countOnHandSuggestion(variant, hubId)
+    unless suggested == variantOverride.count_on_hand
+      variantOverride.count_on_hand = suggested
+      DirtyVariantOverrides.set hubId, variant.id, variantOverride.id, 'count_on_hand', suggested
+
+  # Suggest producer count_on_hand if variant has limited stock and variant override forces limited
+  # stock. Otherwise, clear whatever value is set.
+  $scope.countOnHandSuggestion = (variant, hubId) ->
+    variantOverride = $scope.variantOverrides[hubId][variant.id]
+    return null unless !variant.on_demand && variantOverride.on_demand == false
+    variant.on_hand
--- a/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
+++ b/app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
@@ -14,6 +14,7 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
   $scope.selectView = Views.selectView
   $scope.currentView = -> Views.currentView
   $scope.onDemandOptions = [
+    { description: t('js.variant_overrides.on_demand.use_producer_settings'), value: null },
     { description: t('js.variant_overrides.on_demand.yes'), value: true },
     { description: t('js.variant_overrides.on_demand.no'), value: false }
   ]
--- a/app/views/admin/variant_overrides/_products_variants.html.haml
+++ b/app/views/admin/variant_overrides/_products_variants.html.haml
@@ -10,7 +10,8 @@
   %td.on_hand{ ng: { show: 'columns.on_hand.visible' } }
     %input{name: 'variant-overrides-{{ variant.id }}-count_on_hand', type: 'text', ng: {model: 'variantOverrides[hub_id][variant.id].count_on_hand'}, placeholder: '{{ variant.on_hand }}', 'ofn-track-variant-override' => 'count_on_hand'}
   %td.on_demand{ ng: { show: 'columns.on_demand.visible' } }
-    %input.field{ :type => 'checkbox', name: 'variant-overrides-{{ variant.id }}-on_demand', ng: { model: 'variantOverrides[hub_id][variant.id].on_demand' }, 'ofn-track-variant-override' => 'on_demand' }
+    %select{ name: 'variant-overrides-{{ variant.id }}-on_demand', ng: { model: 'variantOverrides[hub_id][variant.id].on_demand', options: 'option.value as option.description for option in onDemandOptions' }, 'ofn-track-variant-override' => 'on_demand' }
+      %option{ value: '' }= t(".on_demand.use_producer_settings")
   %td.reset{ ng: { show: 'columns.reset.visible' } }
     %input{name: 'variant-overrides-{{ variant.id }}-resettable', type: 'checkbox', ng: {model: 'variantOverrides[hub_id][variant.id].resettable'}, placeholder: '{{ variant.resettable }}', 'ofn-track-variant-override' => 'resettable'}
   %td.reset{ ng: { show: 'columns.reset.visible' } }
--- a/app/views/admin/variant_overrides/_products_variants.html.haml
+++ b/app/views/admin/variant_overrides/_products_variants.html.haml
@@ -8,7 +8,7 @@
   %td.price{ ng: { show: 'columns.price.visible' } }
     %input{name: 'variant-overrides-{{ variant.id }}-price', type: 'text', ng: {model: 'variantOverrides[hub_id][variant.id].price'}, placeholder: '{{ variant.price }}', 'ofn-track-variant-override' => 'price'}
   %td.on_hand{ ng: { show: 'columns.on_hand.visible' } }
-    %input{name: 'variant-overrides-{{ variant.id }}-count_on_hand', type: 'text', ng: { model: 'variantOverrides[hub_id][variant.id].count_on_hand', readonly: 'variantOverrides[hub_id][variant.id].on_demand != false' }, placeholder: '{{ variant.on_hand }}', 'ofn-track-variant-override' => 'count_on_hand'}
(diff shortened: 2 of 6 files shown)


spec/features/admin/variant_overrides_spec.rb
```diff
@@ -141,7 +141,7 @@
             fill_in "variant-overrides-#{variant.id}-sku", with: 'NEWSKU'
             fill_in "variant-overrides-#{variant.id}-price", with: '777.77'
             fill_in "variant-overrides-#{variant.id}-count_on_hand", with: '123'
-            check "variant-overrides-#{variant.id}-on_demand"
+            select I18n.t("js.variant_overrides.options.on_demand.unlimited"), from: "variant-overrides-#{variant.id}-on_demand"
```
luisramos0: I dont think you need I18n here, just t()

app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
```diff
@@ -13,6 +13,10 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
   $scope.RequestMonitor = RequestMonitor
   $scope.selectView = Views.selectView
   $scope.currentView = -> Views.currentView
+  $scope.onDemandOptions = [
+    { description: t('js.variant_overrides.options.on_demand.unlimited'), value: true },
+    { description: t('js.variant_overrides.options.on_demand.limited'), value: false }
```
luisramos0: I find that adding limited/unlimited to the on_demand concept is a bit confusing. Can we keep that complexity at the translations level and use something like this instead?
js.variant_overrides.options.on_demand.true and js.variant_overrides.options.on_demand.false
or js.variant_overrides.options.on_demand.active and js.variant_overrides.options.on_demand.inactive
These can still be translated to on_demand and Limited Stock.

config/locales/en.yml
```diff
@@ -2590,6 +2590,10 @@ See the %{link} to find out more about %{sitename}'s features and to start using
       now_out_of_stock: is now out of stock.
       only_n_remainging: "now only has %{num} remaining."
     variant_overrides:
+      options:
+        on_demand:
+          unlimited: "On Demand"
+          limited: "Limited Stock"
```
luisramos0: as per comment above, instead of unlimited/limited here I'd just use active/inactive or true/false.

app/views/admin/variant_overrides/_products_variants.html.haml
```diff
@@ -10,7 +10,8 @@
   %td.on_hand{ ng: { show: 'columns.on_hand.visible' } }
     %input{name: 'variant-overrides-{{ variant.id }}-count_on_hand', type: 'text', ng: {model: 'variantOverrides[hub_id][variant.id].count_on_hand'}, placeholder: '{{ variant.on_hand }}', 'ofn-track-variant-override' => 'count_on_hand'}
   %td.on_demand{ ng: { show: 'columns.on_demand.visible' } }
-    %input.field{ :type => 'checkbox', name: 'variant-overrides-{{ variant.id }}-on_demand', ng: { model: 'variantOverrides[hub_id][variant.id].on_demand' }, 'ofn-track-variant-override' => 'on_demand' }
+    %select{ name: 'variant-overrides-{{ variant.id }}-on_demand', ng: { model: 'variantOverrides[hub_id][variant.id].on_demand', options: 'option.value as option.description for option in onDemandOptions' }, 'ofn-track-variant-override' => 'on_demand' }
+      %option{ value: '' }
```
luisramos0: has this empty value been decided with product people? maybe we should give them a chance to chose between this "" or "Use Producer Settings" as you have suggested initially.

kristinalim: This was a last-minute decision I made while implementing when I realized we didn't need to label it, also in consideration that we might change the UI again for #3067.

But now, proposing the following much simpler set of options (should have thought of this earlier :grin:):

* Blank
* Yes
* No

Please see [my comment](https://github.com/openfoodfoundation/openfoodnetwork/issues/3066#issuecomment-440847944) in the issue.

kristinalim: @Matt-Yorkley :point_up: This is related also to your feedback.

kristinalim:
Noting that I've also updated the tests, to more comprehensively test setting of "on demand". I added these to make sure the option with "" value is not typecasted to `true`/`false`, and "false" is not typecasted to `true`.

config/locales/en.yml
```diff
@@ -2444,6 +2447,8 @@ See the %{link} to find out more about %{sitename}'s features and to start using
     pending: Pending
     shipped: Shipped
   js:
+    "yes": "Yes"
+    "no": "No"
```
sauloperez: we don't need double quotes for the keys, right?

sauloperez:
Congrats @kristinalim :clap: ! very good investigation and simple solution!

config/locales/en.yml
kristinalim: @sauloperez "yes" and "no" are [reserved](http://yaml.org/type/bool.html) in YAML, and can be [enclosed in quotes as a workaround](https://makandracards.com/makandra/24809-yaml-keys-like-yes-or-no-evaluate-to-true-and-false). :slightly_smiling_face: Interesting quirk. We actually have translations for "(en.)yes" and "(en.)no" earlier in the translations file too, just not under the "js" scope.

luisramos0: nice.

I'd put this in context js.products_variants.on_demand so that translators can do their magic if they want. Like in German this will sound better as "on_demand/limited".

luisramos0:
Moving this to in dev. Unless you disagree @kristinalim , we should move the translations to a on_demand specific scope.
Otherwise, ready for testing.

config/locales/en.yml
kristinalim: Thanks, @luisramos0! I've added a commit to translate those strings. They're in `js.variant_overrides.on_demand`, in case the translations for `on_demand` in the variants themselves would be different.

This is back in Code Review.

RachL:
Holy molyyyyyy I've deployed on staging.openfoodnetwork.org.au 🎉 



This is going to be a fun night

RachL:
@kristinalim while testing I've found one confusing thing IMO for the user:
If I switch from "yes" to "use producer settings" AND I forgot to clear my on hand column, like this for example:
image
Then the on hand column will be taken into account and not the "Use producer settings". As a user I would assume that the sentence reflects the choice made and not the figure put in on hand. To be more clear : in that case I would assume that the producer settings would be taken into account and not my "30" in on hand column.
What do other users think especially @luisramos0 @Matt-Yorkley @myriamboure as you reacted on the original issue?

Also on a minor note: can we choose in the future one style for our combo boxes? In this page only we have 3 different styles:
image
image
image
(we can have different colours though)

Testing notes:
https://docs.google.com/document/d/1beDA3wua0fGA7dtaH60klz6Qh9GmXUsKEG33qFCuBLE/edit#

myriamboure:
Definitely agree, and so there is a little confusion to clarify in the UX and some wrong behavior to fix.
 
If "use producer setting" is selected on the on demand column that will apply only to the stock information, not on the price, right? What I find confusing in UX is that it is in the "on demand" column but concerns also the stock on hand column as the on demand info work hand in hand with stock info, and not the price one:
- if a product is on demand (on demand = yes), even if the there is a stock on hand info the stock on hand will be ignored
- if a product is not on demand (on demand = no), in that case the stock on hand will be read and taken into account quantities will be removed from it when I shop.

Here if we select "use producer setting" it should ideally change back the stock on hand to the value from the producer on-hand stock and associated on demand info. So my suggestion for easy fix would be that:
- by default stock on hand and on demand info in a new product on inventory is inherited from producer, so default on hand = the one in producer catalog and on demand is set top "user producer setting".
- if user change stock on hand to something different than producer value, it means it's not anymore using the producer setting so automatically the on demand column is switched to "no" (if I put on hand value it means it's not on demand)
- If I switch back to "yes" the on hand quantity become greyed (not read anymore)
- If I switch back to "use producer setting" the stock on hand is moved back to producer original value in grey
I also suggest to rename "Use producer stock info" to clarify the potential confusion with price...

What do you think @kristinalim ?

@RachL I would test also when you shop from a shopfront sourcing from inventory that the stock is well treated given box selected,  but anyway we need first to agree on how to fix it.

RachL:
@myriamboure yep that's what I did. I gave Enterprise 2 the permission to sell Bob's farm product, and I opened an OC with Enterprise 2 and the product. That how I tested which stock number was taking into account (see notes).

kristinalim:
@RachL Thanks for testing this. @myriamboure Your recommended behaviour sounds good. I'm already starting to make the changes needed.

kristinalim:
I've made the following additional changes:

1. "Use producer settings" is now "Use producer stock settings".
2. The user can now change the count on hand value only when on demand is "No". Is this okay? Aside from reducing complexity of the UI code, maybe the user should consciously choose the on demand setting before changing count on hand.
3. The count on hand field is still visible when on demand is "Use producer stock settings" or "Yes", but it is disabled and grayed out.
4. We are now using correct placeholders for count on hand. This also fixes #3068.
5. When on demand is changed, we now clear the count on hand, **except** when the producer setting says limited stock and the variant override also forces limited stock. In that case, we pre-fill the field with the producer's count on hand.

| Producer | Producer | VO | VO | VO | VO |
|-|-|-|-|-|-|
| On Demand | On Hand | On Demand | OH Placeholder | OH Changed | OH Visible Value |
| true | nil | nil | "On demand" | blank | "On demand" |
| true | 20 | nil | "On demand" | blank | "On demand" |
| true | nil | true | "On demand" | blank | "On demand" |
| true | 20 | true | "On demand" | blank | "On demand" |
| true | nil | false | blank | blank | blank |
| true | 20 | false | blank | blank | blank |
| false | 20 | nil | 20 | blank | 20 |
| false | 20 | true | "On demand" | blank | On demand |
| false | 20 | false | blank | 20 | 20 or user input |

Please note that these changes are only for the UI. These reduce the chance from now on that we save awkward settings such as an "On Demand" variant override with count on hand, but this doesn't affect existing data nor prevent a custom API request from saving data this way.

I'm separated the commits for that and will submit another PR.

app/views/admin/variant_overrides/_products_variants.html.haml
```diff
@@ -8,9 +8,10 @@
...
   %td.on_hand{ ng: { show: 'columns.on_hand.visible' } }
-    %input{name: 'variant-overrides-{{ variant.id }}-count_on_hand', type: 'text', ng: {model: 'variantOverrides[hub_id][variant.id].count_on_hand'}, placeholder: '{{ variant.on_hand }}', 'ofn-track-variant-override' => 'count_on_hand'}
+    %input{name: 'variant-overrides-{{ variant.id }}-count_on_hand', type: 'text', ng: { model: 'variantOverrides[hub_id][variant.id].count_on_hand', readonly: 'variantOverrides[hub_id][variant.id].on_demand != false' }, placeholder: '{{ countOnHandPlaceholder(variant, hub_id) }}', 'ofn-track-variant-override' => 'count_on_hand'}
   %td.on_demand{ ng: { show: 'columns.on_demand.visible' } }
-    %input.field{ :type => 'checkbox', name: 'variant-overrides-{{ variant.id }}-on_demand', ng: { model: 'variantOverrides[hub_id][variant.id].on_demand' }, 'ofn-track-variant-override' => 'on_demand' }
+    %select{ name: 'variant-overrides-{{ variant.id }}-on_demand', ng: { model: 'variantOverrides[hub_id][variant.id].on_demand', change: 'updateCountOnHand(variant, hub_id)', options: 'option.value as option.description for option in onDemandOptions' }, 'ofn-track-variant-override' => 'on_demand' }
```
luisramos0: instead of this change attribute here it'd prefer to see a "watch" in the controller.

app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
```diff
@@ -105,3 +109,31 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
...
+
+  $scope.countOnHandPlaceholder = (variant, hubId) ->
+    variantOverride = $scope.variantOverrides[hubId][variant.id]
+
+    if variantOverride.on_demand
+      t('js.variants.on_demand.yes')
+    else if variantOverride.on_demand == false
+      ''
+    else
+      variant.on_hand
```
luisramos0: I see the tests are covering this but I'd still add a comment to this critical part of the code so that first time readers can follow.

I'd go for a method doc like this:
```js
// variant override count_on_hand field placeholder logic:
//     on_demand true  -- show "On Demand"
//     on_demand false -- show empty value to be set by the user
//     on_demand nil   -- show producer on_hand value
$scope.countOnHandPlaceholder = (variant, hubId) ->
    variantOverride = $scope.variantOverrides[hubId][variant.id]

      if variantOverride.on_demand == true
            t('js.variants.on_demand.yes')
      else if variantOverride.on_demand == false
            ''
      else
           variant.on_hand
```

app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee
```diff
@@ -105,3 +109,31 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
...
+  # This method should only be used when the variant override on_demand is changed.
+  #
+  # Change the count_on_hand value to a suggested value.
+  $scope.updateCountOnHand = (variant, hubId) ->
+    variantOverride = $scope.variantOverrides[hubId][variant.id]
+
+    suggested = $scope.countOnHandSuggestion(variant, hubId)
+    unless suggested == variantOverride.count_on_hand
+      variantOverride.count_on_hand = suggested
+      DirtyVariantOverrides.set hubId, variant.id, variantOverride.id, 'count_on_hand', suggested
```
luisramos0: I prefer early returns and if instead of unless:

 ```js
return if suggested == variantOverride.count_on_hand
variantOverride.count_on_hand = suggested
DirtyVariantOverrides.set hubId, variant.id, variantOverride.id, 'count_on_hand', suggested
```

File: config/locales/en.yml (hunk 5 above)
Comment (luisramos0) [reply config/locales/en.yml]: do you still need these js.yes and js.no here?

File: app/views/admin/variant_overrides/_products_variants.html.haml (hunk 6 above)
Comment (kristinalim) [reply app/views/admin/variant_overrides/_products_variants.html.haml]: @luisramos0 As far as I understand, doing a `watch` here would be too complicated because we're looking at a dynamic expression where the value is deeply nested.

Ideally, we have a separate service for `products_fields`, but it will take work. I'll see tomorrow if it's actually more simple than I think. How different would variant overrides be in 2.0?

File: app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee (hunk 7 above)
Comment (kristinalim) [reply app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee]: Done! :slightly_smiling_face: Thanks. https://github.com/openfoodfoundation/openfoodnetwork/pull/3104/commits/02c6f74bed966f80af9d6b9b06fe3e6d0cf4cf9d

File: app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee (hunk 8 above)
Comment (kristinalim) [reply app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee]: Changed in https://github.com/openfoodfoundation/openfoodnetwork/pull/3104/commits/9d423d586de5cd778095f187e57116fd1ccd0b53.

File: config/locales/en.yml (hunk 5 above)
Comment (kristinalim) [reply config/locales/en.yml]: Not anymore, @luisramos0.

Removed in https://github.com/openfoodfoundation/openfoodnetwork/pull/3104/commits/b3f440c17d44cf2721525d00595b434faf36a994. Thanks!

File: app/views/admin/variant_overrides/_products_variants.html.haml (hunk 6 above)
Comment (luisramos0) [reply app/views/admin/variant_overrides/_products_variants.html.haml]: let's ignore it, it was only in case it was straight forward.

**Comment by kristinalim** [comment]
FYI, this set of changes can go ahead of any changes for #3067, which is still in discussion.

File: app/assets/javascripts/admin/variant_overrides/controllers/variant_overrides_controller.js.coffee [hunk 9]
```diff
@@ -13,6 +13,10 @@ angular.module("admin.variantOverrides").controller "AdminVariantOverridesCtrl",
   $scope.RequestMonitor = RequestMonitor
   $scope.selectView = Views.selectView
   $scope.currentView = -> Views.currentView
+  $scope.onDemandOptions = [
```
mkllnk: Why didn't you include the third option in here?

kristinalim: I wasn't able to get this to work at first, but I think this was only in the initial version when I was using string counterparts of `null`, `true`, and `false`. Angular automatically added a blank option so I just *renamed* the label, but it turns out it wouldn't add this if there is a match.

Fixed already: https://github.com/openfoodfoundation/openfoodnetwork/pull/3104/commits/517a7c85f257d16aadc2c621561dbb7c721d3c6e

Thanks, @mkllnk! :slightly_smiling_face:

luisramos0:
staged here: https://staging.openfoodnetwork.org.uk

luisramos0:
asset compilation error on deployment:
```css
Sass::SyntaxError: Undefined variable: \"$disabled-light\".\n  (in /home/openfoodnetwork/apps/openfoodnetwork/current/app/assets/stylesheets/admin/components/input.css.scss)\n/home/openfoodnetwork/apps/openfoodnetwork/current/app/assets/stylesheets/admin/components/input.css.scss:4\n/home/openfoodnetwork/.gem/ruby/2.1.0/gems/sass-3.3.14/lib/sass/script/tree/variable.rb:49:in `_perform'\n/home/openfoodnetwork/.gem/ruby/2.1.0/gems/sass-3.3.14/lib/sass/script/tree/node.rb:50:in `perform'\n/home/openfoodnetwork/.gem/ruby/2.1.0/gems/sass-3.3.14/lib/sass/tree/visitors/perform.rb:366:in `visit_prop'\n/home/openfoodnetwork/.gem/ruby/2.1.0/gems/sass-3.3.14/lib/sass/tree/visitors/base.rb:36:in `visit'\n/home/openfoodnetwork/.gem/ruby/2.1.0/gems/sass-3.3.14/lib/sass/tree/visitors/perform.rb:152:in `block in visit'\n/home/openfoodnetwork/.gem/ruby
```

kristinalim:
@luisramos0 Thanks, I was able to reproduce the error when compiling the assets.

I've pushed the fix for this - it  should work now.

luisramos0:
ok, staged here: https://staging.openfoodnetwork.org.uk

RachL:
@kristinalim @luisramos0 @myriamb two things I hit when testing and I would like to understand if they are a blocker or not:
A. When I load the inventory the in stock column is greyed out and I cannot write inside the column so I cannot test the following proposition made by Myriam:
> if user change stock on hand to something different than producer value, it means it's not anymore using the producer setting so automatically the on demand column is switched to "no" (if I put on hand value it means it's not on demand)

Is this made on purpose ?
B. Even while working on an inventory with only 2 products, staging uk is slow (all production data are set up there, so it gives us a good view of live response time). So when I try to save I have this message:
image
Is this 100% a time out issue or can it be also a pb in the PR? Thanks :)

myriamboure:
About question A, @RachL you have selected "yes" in the "on demand" column so you "forced" on demand, it is normal that you cannot update the stoch on hand value, it is on demand ;-) Can you share your testing notes link ? I can finish testing tomorrow or saturday, or even tonight maybe ;-)

kristinalim:
> When I load the inventory the in stock column is greyed out and I cannot write inside the column so I cannot test the following proposition made by Myriam:
>
> if user change stock on hand to something different than producer value, it means it's not anymore using the producer setting so automatically the on demand column is switched to "no" (if I put on hand value it means it's not on demand)

@RachL This is how I intended it, and also wanted to check with you if this is okay. This simplifies the code and the logic, and I think it is also good that the user changes the "On Demand" setting first.

This makes it clearer that, when the user sets the value, they are choosing to use their own stock tracker (the stock count of the producer will not be affected) even if the VO and the producer COH are equal.

kristinalim:
Sorry, I need to make a follow-up change that will make it possible to change/clear non-blank COH of existing VOs that are "Use producer stock settings" or "Yes" for on demand. Right now, the UI expects the COH for these to already be blank, and wants these to remain blank, so the field is not editable.

@RachL Maybe that is part of your concern.

This follow-up change will no longer be needed with the PR in #3067.

myriamboure:
@kristinalim I don't have an objection that it stays as it is actually, I find the UX pretty clean. I'm testing, sorry I should have assigned myself, wait a minute I'll share my feedbacks, but looks pretty clean to me.

myriamboure:
So I tested different cases:
- override as a producer my own products in inventory
- override as distributor the producer inventory
- I tried to check with preexisting VOs how the switch behave in new system. I didn't find any :-( I couldn't test if a variant is overriden to "on demand = yes" in inventory it is well migrated in the new with that PR.

All seems pretty consistent, basically:
- by default in inventory it's "use producer stock info" so by default basically you cannot change the COH. You have to click on "no" and then you can update the COH field.
- if you select "yes" it set up COH to on demand and you can't modify it

An improvement on UX would be to enable, if "use producer stock info " or "yes" is selected, to change COH field but if I start typing then it switches on unlimited ? column to "no".

But it is acceptable as it is IMO. So up to you @kristinalim it's fixing the main problem so it's a 5 min improvement you can do it, else don't spend 1/2 day on it. The issue is fixed and that's the most important IMO, and UX is ok even if not optimal.

kristinalim:
Thanks, @myriamboure.

I updated the "Release Notes" and "Dependencies" sections of this PR.

Regarding what I said about existing VO records that have non-ideal stock configs, e.g. "Use producer stock settings" but with a COH override (possible to save this before this PR), I realized it is hard for the user to know if there is an override or just a placeholder/hint because the field is disabled (not editable). The other option of not disabling the field will also require more changes, so I just finished working on #3198 which will also migrate these existing VO records to ideal configs.

I prefer for this PR to be merged together with #3198.

myriamboure:
Ok so from what I understand it is ready to go once the other one is ready, so moving it to ready to go but blocked. Great !

PR #3126, opened 2018-11-23
2841  Associate customers with their user records

### LINKED ISSUE #2841: Validation email error with an old account with bug before email confirmation work
## Description
We have one user in France who created her account when there was a bug with private shopfronts. We discovered the bug with her actually. Today her account still have the following issue:
- she still appears as a "guest user" even though she has an account, she logs in, etc. 
capture du 2018-10-04 16-59-49
- the hub manager created a subscription for her but the subscription is failing, with the given error message
capture du 2018-10-08 14-57-40
(the message says that 5 recurring orders were planned for that OC, and 4 were treated successfully, 1 failed du to error:
R377155761: this email is already associated to a user and has already been validated. Please login to continue, or use another email.)

Also strangely if I look at the failing order R377155761 here is what I see
capture du 2018-10-08 15-06-02
so it is in "cart" status but with "balance due" which is somehow strange...

We are stuck, we can give no support, she doesn't have another email she can use, and anyway I tried that workaround but it's failing also (see #2840) 

## Expected Behavior
She has validated her email so she shouldn't be a guest anymore and not treated as such. She can get subscriptions.

## Steps to Reproduce
1. You need to connect as a super admin on French instance and see in customer the status for that user.
2. You can check the subscription failing status from the order R377155761

## Context
The user is trying to get subscription working... but it doesn't work for that particular customer.

## Severity
There is no workaround, but it concerns only one user, so I would say S3

## Your Environment

* Version used: v1.18

## Possible Fix
- Can we resend some confirmation link? Or check in data base the status for that user and fix it?

mkllnk:
#### What? Why?

Closes #2841.

When we introduced the Customer model, we didn't associate any existing
customers with users that have the same email address.
Later we decided to create that association when users sign up. But we didn't
update all the existing customers. We do that now for data consistency and to
solve several bugs.

#### What should we test?

Our staging data is too new to contain the legacy records that are fixed with this pull request. Only developers can test this against production data on their local machine.

#### Release notes

Fixed: Some customer records created before May 2018 caused subscriptions to fail and get repaired with this release.

Changelog Category: Fixed

#### Discourse thread

https://community.openfoodnetwork.org/t/subscriptions-and-customers/1512

### DIFF
--- a/db/migrate/20181123012635_associate_customers_to_users.rb
+++ b/db/migrate/20181123012635_associate_customers_to_users.rb
@@ -1,0 +1,42 @@
+# When we introduced the Customer model, we didn't associate any existing
+# customers with users that have the same email address.
+# Later we decided to create that association when users sign up. But we didn't
+# update all the existing customers. We do that now for data consistency and to
+# solve several bugs.
+#
+# - https://github.com/openfoodfoundation/openfoodnetwork/pull/2084
+# - https://github.com/openfoodfoundation/openfoodnetwork/issues/2841
+class AssociateCustomersToUsers < ActiveRecord::Migration
+  class Customer < ActiveRecord::Base
+  end
+
+  def up
+    save_customers
+    execute "UPDATE customers
+              SET user_id = spree_users.id
+              FROM spree_users
+              WHERE customers.email = spree_users.email
+               AND customers.user_id IS NULL;"
+  end
+
+  def down
+    customers = backed_up_customers
+    Customer.where(id: customers).update_all(user_id: nil)
+  end
+
+  def save_customers
+    customers = Customer.
+      joins("INNER JOIN spree_users ON customers.email = spree_users.email").
+      where(user_id: nil).all
+
+    File.write(backup_file, Marshal.dump(customers))
+  end
+
+  def backed_up_customers
+    Marshal.load(File.read(backup_file))
+  end
+
+  def backup_file
+    File.join("log", "customers_without_user_association.log")
+  end
+end
--- a/spec/support/request/shop_workflow.rb
+++ b/spec/support/request/shop_workflow.rb
@@ -13,7 +13,7 @@ module ShopWorkflow
   end
 
   def set_order(order)
-    ApplicationController.any_instance.stub(:session).and_return({order_id: order.id, access_token: order.token})
+    allow_any_instance_of(ApplicationController).to receive(:session).and_return({order_id: order.id, access_token: order.token})
   end
 
   def add_product_to_cart(order, product, quantity: 1)


mkllnk:
Thank you for your reviews. I started a new wiki page for migrations: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Database-migrations

luisramos0:
Awesome @mkllnk , I linked the new wiki page from here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code-Conventions

mkllnk:
Nice, thank you. It looks like we still didn't test enough. Paco reported the migration failing in France. I have to look into it.

So maybe a new proposal would be to run all migrations against all production data locally. I mean for each OFN instance:

- download production data and import into local dev database
- run migration
- roll back migration
- go to next instance

That takes a while and requires access to all these instances, but we can write a little script for it.

sauloperez:
For data migrations, a strategy that proved to be battle-tested in the past was to run them in production as soon as they got approved and before merging. Then if something came up, the migration could be fixed, like in this case.

luisramos0:
@mkllnk even if we dont do this process for migrations I'd like to see how you guys "download production data and import into local dev database". can you share a script?

mkllnk:
@luisramos0 There is a script in the repository that is broken now, because it hasn't been updated for a while: `script/mirror_db.sh`

I'm experimenting with a new version, but I haven't fixed the last part of modifying the data for development yet:
```bash
#!/bin/bash

# Used to pull data from production or staging servers into local dev database
# Useful for when you want to test a migration against production data, or see
# the effect of codebase changes on real-life data

# Usage: script/mirror_db.sh [ofn-staging1|ofn-staging2|ofn-prod]

set -e

if hash zeus 2>/dev/null && [ -e .zeus.sock ]; then
  RAILS_RUN='zeus r'
else
  RAILS_RUN='bundle exec rails runner'
fi

#if [[ "$1" = 'ofn-prod' ]]; then
#  DB_USER='openfoodnetwork'
#  DB_USER='openfoodweb'
#  DB_DATABASE='openfoodweb_production'
#elif [[ "$1" 
#
#else
#  DB_USER='ofn_user'
#  DB_DATABASE='openfoodnetwork'
#fi

#DB_USER='openfoodnetwork'
DB_USER='ofn_user'
DB_DATABASE='openfoodnetwork'

DB_OPTIONS=(--exclude-table-data=sessions --no-acl)

# -- Mirror database
echo "Mirroring database..."
dropdb -h localhost -U ofn open_food_network_dev
createdb -h localhost -U ofn open_food_network_dev
ssh -C "$1" "pg_dump -h localhost -U $DB_USER $DB_DATABASE ${DB_OPTIONS[@]}" | psql -h localhost -U ofn open_food_network_dev


echo "Please fix me"
exit
# -- Disable S3
echo "Preparing mirrored database..."
$RAILS_RUN script/prepare_imported_db.rb


# -- Mirror images
if hash aws 2>/dev/null; then
    echo "Mirroring images..."
    BUCKET=`echo $1 | sed s/-/_/ | sed "s/\\([0-9]\\)/_\1/" | sed s/prod/production/`
    aws s3 sync s3://$BUCKET/public public/

else
    echo "Please install the AWS CLI tools so that I can copy the images from $1 for you."
    echo "eg. sudo easy_install awscli"
fi
```

Issue #3016, opened 2018-11-13
[Spree Upgrade][FrontOffice] Fix restart checkout spec in checkout_controller (1 broken test)


luisramos0:
fix checkout controller restart spec.

First a shipment needs to be set in the order instead of the shipping_method_id, for example:
`order.shipments << shipment_pending`
instead of
`order.update_attribute(:shipping_method_id, shipment_pending.shipping_method_id)`

But afterwards all the assertions need to be verified or adapted.

```
16) CheckoutController#restart_checkout when the order is already in the 'cart' state does nothing
      Failure/Error: order.update_attribute(:shipping_method_id, shipment_pending.shipping_method_id)
      
      NoMethodError:
        undefined method `shipping_method_id' for #<Spree::Shipment:0x00555812914b68>
      # ./spec/controllers/checkout_controller_spec.rb:226:in `block (3 levels) in <top (required)>'

  17) CheckoutController#restart_checkout when the order is in a subsequent state resets the order state, and clears incomplete shipments and payments
      Failure/Error: order.update_attribute(:shipping_method_id, shipment_pending.shipping_method_id)
      
      NoMethodError:
        undefined method `shipping_method_id' for #<Spree::Shipment:0x007f80d03b1520>
      # ./spec/controllers/checkout_controller_spec.rb:226:in `block (3 levels) in <top (required)>'
```

luisramos0:
Related to #2683

sauloperez:
I got a bit stuck on 

https://github.com/openfoodfoundation/openfoodnetwork/blob/f593a700002cc899afeb5f8bc63a8f481ca1e7c5/spec/controllers/checkout_controller_spec.rb#L244-L258

IMO it depends too much on the actual state of the DB, payments, the order, etc. it's very hard to foresee them all. I suggest we extract the following

https://github.com/openfoodfoundation/openfoodnetwork/blob/f593a700002cc899afeb5f8bc63a8f481ca1e7c5/app/controllers/checkout_controller.rb#L161-L168

on a class and then, we test there in isolation. I'll be quicker.

luisramos0:
yes, sounds good.
services/checkout_restart.rb ;-)

luisramos0:
but maybe better in master first, right?

luisramos0:
this issue is directly related to #3027

sauloperez:
#3130 has been merged but AFAIK this issue is still open.

luisramos0:
[...]

PR #3130, opened 2018-11-23
Extract restart checkout

sauloperez:
#### What? Why?

Partially solves https://github.com/openfoodfoundation/openfoodnetwork/issues/3016

As agreed in https://github.com/openfoodfoundation/openfoodnetwork/issues/3016#issuecomment-439474695, it's worth extracting the logic to restart a checkout into a service and improve a dubious test that is failing in the upgrade branch.

#### What should we test?

Go through a checkout and cause an error, like a failed payment. This should display the error and restart the checkout as always. There should be no change in behavior.

#### Release notes

Refactor logic to restart a checkout to take it out of the controller.

Changelog Category: Changed

It will make a failing test pass in the `2-0-stable` branch.

### DIFF
--- a/spec/controllers/checkout_controller_spec.rb
+++ b/spec/controllers/checkout_controller_spec.rb
@@ -189,7 +189,7 @@ describe CheckoutController, type: :controller do
 
   describe "Paypal routing" do
     let(:payment_method) { create(:payment_method, type: "Spree::Gateway::PayPalExpress") }
-    let(:restart_checkout) { instance_double(RestartCheckout, restart_checkout: true) }
+    let(:restart_checkout) { instance_double(RestartCheckout, call: true) }
 
     before do
       allow(controller).to receive(:current_distributor) { distributor }
@@ -208,9 +208,7 @@ describe CheckoutController, type: :controller do
   end
 
   describe "#update_failed" do
-    let(:restart_checkout) do
-      instance_double(RestartCheckout, restart_checkout: true)
-    end
+    let(:restart_checkout) { instance_double(RestartCheckout, call: true) }
 
     before do
       controller.instance_variable_set(:@order, order)
@@ -219,7 +217,7 @@ describe CheckoutController, type: :controller do
 
     it "clears the shipping address and restarts the checkout" do
       expect(controller).to receive(:clear_ship_address)
-      expect(restart_checkout).to receive(:restart_checkout)
+      expect(restart_checkout).to receive(:call)
       expect(controller).to receive(:respond_to)
 
       controller.send(:update_failed)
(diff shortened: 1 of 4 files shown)


spec/controllers/checkout_controller_spec.rb
```diff
@@ -205,57 +208,19 @@
...
-
-      # NOTE: at the time of writing, it was not possible to create a shipment with a state other than
-      # 'pending' when the order has not been completed, so this is not a case that requires testing.
-      it "resets the order state, and clears incomplete shipments and payments" do
-        expect(order).to receive(:restart_checkout!).and_call_original
-        expect(order.shipping_method_id).to_not be nil
-        expect(order.shipments.count).to be 1
-        expect(order.adjustments.shipping.count).to be 1
-        expect(order.payments.count).to be 2
-        expect(order.adjustments.payment_fee.count).to be 2
```
luisramos0: I think this pre execution assertion are useful in such a test with multiple changes going on. I'd keep them.

sauloperez: I don't know... the fact that you need to set assertions because you are not sure what's the state the app is in feels like a symptom to me.

luisramos0: ok. it's up to you 👍 . for me it's more about readability of the test. with the assertions before you see the movement. for example, knowing that when you add shipment_pending there should be a order.adjustments.shipping is very specific.

sauloperez: That's a valid point although, to me, that doesn't belong here. Let's see what others think.

luisramos0: 👍

sauloperez:
To me, looking at `RestartCheckout` as a reader I can guess what it does while with `OrderStateService` it's not very clear. It 👃 that it'll do many things, not following the single responsibility principle.

If later on we have more single-purpose services throughout the controller and prefer to talk to a single object such as `OrderStateService` we can combine them in that facade 👍

sauloperez:
Ok, I'll document this later then :+1:

RachL:
Staged on https://staging.openfoodnetwork.org.au/ Will not test this right away, if someone else has time, help yourself.

RachL:
@sauloperez just to be sure, what is a failed payment ? Is it just the user failing to fil in like this:
image


or should I look for a wrong credit card number for example?

mkllnk:
@RachL No, the input needs to be valid so that a payment attempt is made. Stripe has a list of credit card numbers you can use to test error cases: https://stripe.com/docs/testing That can be very handy.

sauloperez:

RachL:
hmmm @sauloperez do you know how to set up stripe on katuma staging?

RachL:
I'm starting deployment on Aus staging.

RachL:
Stripe has really a nice playground :)

All good both desktop and mobile. Let's move it https://docs.google.com/document/d/151N7kCGMsFxu3Lk8jOvJr203Y9nPsBAX53gW4EInr34/edit#

Issue #2804, opened 2018-10-02
[Spree Upgrade][BackOffice] Fix reference to Product#on_hand= in bulk_product_update_spec (7 broken tests)


sauloperez:
## Description
Originally this issue included 3 similar errors in product_spec. 2 were fixed in #3125 and one was moved to a new separate issue #3137

Now this issue is only related to `spec/features/admin/bulk_product_update_spec.rb`
This spec creates products providing an `on_hand` value. As this method no longer exists. This results in the following type of error:

```ruby
  33) 
  As an Administrator
  I want to be able to manage products in bulk
 listing products displays an on hand count in a span for each product
      Failure/Error: p1 = FactoryBot.create(:product, on_hand: 15)
      
      NoMethodError:
        undefined method `on_hand=' for #<Spree::Product:0x0055d1b1522dc0>
      # ./spec/features/admin/bulk_product_update_spec.rb:58:in `block (3 levels) in <top (required)>'
```

There are currently 7 broken tests with this error (see in a comment below).

## Possible Fix

Similar to `VariantStock`, we need to provide an adapter that provides backward compatibility between our code and Spree 2.0's `Product`.
Related to https://github.com/openfoodfoundation/openfoodnetwork/issues/2802.

luisramos0:
We need to validate if this method is needed in live code (controllers?). If not needed, the specs should be adapted to set these values on the variant.

luisramos0:
List of errors:
```
 28) Spree::Product validations and defaults does not save when master is invalid
      Failure/Error: product.on_hand = "10,000"
      
      NoMethodError:
        undefined method `on_hand=' for #<Spree::Product:0x0055806be4e620>
      # ./spec/models/spree/product_spec.rb:29:in `block (3 levels) in <module:Spree>'

  29) Spree::Product validations and defaults when the product has variants saving a new product copies the properties on master variant to the first standard variant
      Failure/Error: product.on_hand = 3
      
      NoMethodError:
        undefined method `on_hand=' for #<Spree::Product:0x007effe2b67dd8>
      # ./spec/models/spree/product_spec.rb:109:in `block (5 levels) in <module:Spree>'

  30) Spree::Product validations and defaults when the product has variants saving a new product only duplicates master with after_save when no standard variants exist
      Failure/Error: product.on_hand = 3
      
      NoMethodError:
        undefined method `on_hand=' for #<Spree::Product:0x007effe342c2c0>
[...]

PR #3132, opened 2018-11-23
[Spree Upgrade] Add VariantStock.on_demand as attribute_accessible

### LINKED ISSUE #3017: [Spree Upgrade][FrontOffice] Fix cart_controller_specs errors related to on_hand numbers (3 broken tests)
Fix the 3 broken tests in cart_controller_spec.

Probably some detail related to the new variant.on_hand adapter introduced in #2615.

```
  24) CartController returning stock levels in JSON on success generating stock levels returns a hash with variant id, quantity, max_quantity and stock on hand
      Failure/Error:
        expect(controller.stock_levels(order, [v.id])).to eq(
          {v.id => {quantity: 2, max_quantity: 3, on_hand: 4}}
        )
      
        expected: {1523=>{:quantity=>2, :max_quantity=>3, :on_hand=>4}}
             got: {1523=>{:quantity=>2, :max_quantity=>3, :on_hand=>5}}
      
        (compared using ==)
      
        Diff:
        @@ -1,2 +1,2 @@
        -1523 => {:quantity=>2, :max_quantity=>3, :on_hand=>4},
        +1523 => {:quantity=>2, :max_quantity=>3, :on_hand=>5},
        
      # ./spec/controllers/cart_controller_spec.rb:36:in `block (4 levels) in <top (required)>'

  25) CartController returning stock levels in JSON on success generating stock levels includes all line items, even when the variant_id is not specified
      Failure/Error:
        expect(controller.stock_levels(order, [])).to eq(
          {v.id => {quantity: 2, max_quantity: 3, on_hand: 4}}
        )
      
        expected: {1529=>{:quantity=>2, :max_quantity=>3, :on_hand=>4}}
             got: {1529=>{:quantity=>2, :max_quantity=>3, :on_hand=>5}}
      
        (compared using ==)
      
        Diff:
        @@ -1,2 +1,2 @@
        -1529 => {:quantity=>2, :max_quantity=>3, :on_hand=>4},
        +1529 => {:quantity=>2, :max_quantity=>3, :on_hand=>5},
        
      # ./spec/controllers/cart_controller_spec.rb:42:in `block (4 levels) in <top (required)>'

  26) CartController returning stock levels in JSON on success generating stock levels includes an empty quantity entry for variants that aren't in the order
      Failure/Error:
        expect(controller.stock_levels(order, [v.id, v2.id])).to eq(
          {v.id  => {quantity: 2, max_quantity: 3, on_hand: 4},
           v2.id => {quantity: 0, max_quantity: 0, on_hand: 2}}
        )
      
        expected: {1535=>{:quantity=>2, :max_quantity=>3, :on_hand=>4}, 1538=>{:quantity=>0, :max_quantity=>0, :on_hand=>2}}
             got: {1535=>{:quantity=>2, :max_quantity=>3, :on_hand=>5}, 1538=>{:quantity=>0, :max_quantity=>0, :on_hand=>5}}
      
        (compared using ==)
      
        Diff:
        @@ -1,3 +1,3 @@
        -1535 => {:quantity=>2, :max_quantity=>3, :on_hand=>4},
        -1538 => {:quantity=>0, :max_quantity=>0, :on_hand=>2},
        +1535 => {:quantity=>2, :max_quantity=>3, :on_hand=>5},
        +1538 => {:quantity=>0, :max_quantity=>0, :on_hand=>5},
        
      # ./spec/controllers/cart_controller_spec.rb:48:in `block (4 levels) in <top (required)>'

```

luisramos0:
#### What? Why?

This is required so that it is mass-assigned from product_controller_decorator.bulk_update.
That makes the admin product list on_demand checkboxes work.
That helps the changes in PR #3139 close issue #2804 by fixing two extra tests. 

Also, this closes #3017 by fixing the 3 tests and this also fixes one of the 12 tests in #2943 

#### What should we test?
This PR fixes 6 tests.

Hey, let's stage this one for proper testing!!!!
Product List on_demand checkboxes are now working.

### DIFF
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -105,6 +105,7 @@ module VariantStock
     # unknown reasons, it does not pass the test.
     stock_items.each do |item|
       item.backorderable = new_value
+      item.save
     end
   end
 
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -126,11 +126,8 @@ module VariantStock
   end
 
   # Backwards compatible setting of stock levels in Spree 2.0.
-  # It would be better to use `Spree::StockItem.adjust_count_on_hand` which
-  # takes a value to add to the current stock level and uses proper locking.
-  # But this should work the same as in Spree 1.3.
   def overwrite_stock_levels(new_level)
-    stock_item.__send__(:count_on_hand=, new_level)
+    stock_item.adjust_count_on_hand(new_level - stock_item.count_on_hand)
   end
 
   # There shouldn't be any other stock items, because we should
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -16,7 +16,7 @@ module VariantStock
   extend ActiveSupport::Concern
 
   included do
-    attr_accessible :on_hand
+    attr_accessible :on_hand, :on_demand
     after_update :save_stock
   end
 
--- a/spec/features/admin/product_import_spec.rb
+++ b/spec/features/admin/product_import_spec.rb
@@ -19,11 +19,11 @@ feature "Product Import", js: true do
   let!(:shipping_category) { create(:shipping_category) }
 
   let!(:product) { create(:simple_product, supplier: enterprise2, name: 'Hypothetical Cake') }
-  let!(:variant) { create(:variant, product_id: product.id, price: '8.50', on_hand: '100', unit_value: '500', display_name: 'Preexisting Banana') }
-  let!(:product2) { create(:simple_product, supplier: enterprise, on_hand: '100', name: 'Beans', unit_value: '500', description: '', primary_taxon_id: category.id) }
-  let!(:product3) { create(:simple_product, supplier: enterprise, on_hand: '100', name: 'Sprouts', unit_value: '500') }
-  let!(:product4) { create(:simple_product, supplier: enterprise, on_hand: '100', name: 'Cabbage', unit_value: '500') }
-  let!(:product5) { create(:simple_product, supplier: enterprise2, on_hand: '100', name: 'Lettuce', unit_value: '500') }
+  let!(:variant) { create(:variant, product_id: product.id, price: '8.50', on_hand: 100, unit_value: '500', display_name: 'Preexisting Banana') }
+  let!(:product2) { create(:simple_product, supplier: enterprise, on_hand: 100, name: 'Beans', unit_value: '500', description: '', primary_taxon_id: category.id) }
+  let!(:product3) { create(:simple_product, supplier: enterprise, on_hand: 100, name: 'Sprouts', unit_value: '500') }
+  let!(:product4) { create(:simple_product, supplier: enterprise, on_hand: 100, name: 'Cabbage', unit_value: '500') }
+  let!(:product5) { create(:simple_product, supplier: enterprise2, on_hand: 100, name: 'Lettuce', unit_value: '500') }
   let!(:variant_override) { create(:variant_override, variant_id: product4.variants.first.id, hub: enterprise2, count_on_hand: 42) }
   let!(:variant_override2) { create(:variant_override, variant_id: product5.variants.first.id, hub: enterprise, count_on_hand: 96) }
 


sauloperez:
> Hey, let's stage this one for proper testing!!!!
Product List on_demand checkboxes are now working.

Is this possible?????

sauloperez:
You'll need a rebase :recycle: here

luisramos0:
These fields are now working!
screenshot 2018-11-23 at 12 08 26

luisramos0:
re staging: we need to try it at some point!

luisramos0:
This code is breaking 21 tests...

giphy 1

app/models/concerns/variant_stock.rb
```diff
@@ -105,6 +105,7 @@ def on_demand=(new_value)
     # unknown reasons, it does not pass the test.
     stock_items.each do |item|
       item.backorderable = new_value
+      item.save
```
sauloperez: I would really think twice to do this. These are the kind of things that cause the myriad of UPDATE and INSERT statements we see in development, resulting in multiple and pointless writes sometimes. 

This is just like another db column baked by AR and in that case, AR won't persist on assignments. I would not do it in our case either. To me, this is the responsibility of the consumer of this setter method.

luisramos0: ok, thinking twice... I write up my thoughts so you can comment:

1. first I wanted to see **if this works**. answer: yes! this fixes not just the 21 one broken tests but a couple more.

2. then, I wanted to **understand** why: I think I now understand what's going on, if we dont save on_demand, the call to on_hand is saving the stock_item and ignoring the on_demand value (stock_item.backorderable) set just before by the consumer. By saving when we call on_demand, the consumers request to set on_demand is not lost when on_hand is called.

3. **explaining** this change: I got into this because the initial version was not working correctly. With  stock_item.send count_on_hand= the stock item data was not saved correctly on mass assignments. I believe it's acceptable to say that these stock management methods on_hand and on_demand require a lock on update and thus a save (this is what spree does on stock_item.count_on_hand). one advantage of this approach is that it makes both on_hand and on_demand behave similarly. So, because it's stock management, I actually think it's a nice exception to the rule of letting the consumers control persistence.

4. **improvements** to this: I believe that if we keep these saves in on_hand and on_demand we make it better by removing attribute_accessible, i.e., these methods need to be called explicitly by the consumers.

3. I believe this the best of the **alternatives** we have so far. what other alternatives do we have? any ideas? maybe some kind of after_save hook on variant to persist the stock_item? I am not sure...

luisramos0:
This PR fixes 4 tests:
spec/controllers/cart_controller_spec.rb:35
spec/controllers/cart_controller_spec.rb:41
spec/controllers/cart_controller_spec.rb:47
spec/models/spree/line_item_spec.rb:92

app/models/concerns/variant_stock.rb
mkllnk: Good thoughts. Saving the stock items after saving the variant seems reasonable to me. Worth a try?

There is a difference between on_hand and on_demand in the way they have to be synchronised. on_demand is set by a manager and not modified during checkout. It has only two values. Locking is not essential.

The on_hand value is very critical though. And offering an admin interface to set stock levels is ambiguous. What if the manager counts the product on the shelf and finds that there are `5` while the database still shows `6` (someone was hungry last night). Now the manager enters `5` while a customer buys one at the same time. Should we save `5` or `4`? I think the new Spree 2 approach is wise in allowing only stock increase and decrease, no direct setting. It would end up with `4`.

As long as we keep this consistent (`new_level - stock_item.count_on_hand` uses the two values from the same point in time) it doesn't matter when we save it to the database.

luisramos0: >> Saving the stock items after saving the variant seems reasonable to me. 

Rethinking about it. I think the solution in this PR is better. It's more explicit than having callbacks.

Another approach could be to remove on_hand from mass assigments and keep on_demand as mass assignable. That would make the API a bit weird as users would have to save after on_demand but not after calling on_hand.
I think in this case, this would be the normal usage:
variant.on_demand = false
variant.on_hand = 4
variant.save
But I am not sure how to make this sequence work correctly:
variant.on_demand = false
variant.save
variant.on_hand = 4

Do you guys think that would be better than the current PR?

sauloperez: I just read both of your comments and let's go with this PR. After the upgrade there will be for sure a cleanup phase where we'll realize about the long term implications of all these small trade-offs #iterate

Thanks for making thoughtfull and objective discussions. We rock :muscle:!

luisramos0: ok, cool! This moves us to 90 broken tests in the 2-0-stable build.

luisramos0:
just to finish the gif story on this one:
giphy 2

sauloperez:
He was my hero @luisramos0 !

PR #3133, opened 2018-11-23
[Spree Upgrade] Fix ProductsRenderer specs

### LINKED ISSUE #3029: [Spree Upgrade] Products with count_on_hand = 0 should not be returned by product renderer
Fix spec/lib/open_food_network/products_renderer_spec.rb:51

Should be related to #2417

HugsDaniel:
#### What? Why?

Closes #3029 

Set `variant.on_demand` to false, so when setting `on_hand` to 0 the JSON response for available products does not include the product related to this variant.

#### What should we test?

`spec/lib/open_food_network/products_renderer_spec.rb` should pass.

### DIFF
--- a/spec/lib/open_food_network/products_renderer_spec.rb
+++ b/spec/lib/open_food_network/products_renderer_spec.rb
@@ -49,6 +49,7 @@ module OpenFoodNetwork
       end
 
       it "doesn't return products not in stock" do
+        variant.update_attribute(:on_demand, false)
         variant.update_attribute(:count_on_hand, 0)
         pr.products_json.should_not include product.name
       end


spec/lib/open_food_network/products_renderer_spec.rb
```diff
@@ -49,6 +49,7 @@ module OpenFoodNetwork
       end
 
       it "doesn't return products not in stock" do
+        variant.update_attribute(:on_demand, false)
         variant.update_attribute(:count_on_hand, 0)
```
sauloperez: you could use `update_attributes` here with both of them.

HugsDaniel: I tried with `variant.update_attribute(on_demand: false, count_on_hand: 0)` but it still fails with the same error

sauloperez: Note the extra s in `update_attributeS`. Did you try this? it's weird anyway, because it only changes from 2 UPDATE db statements to a single one.

HugsDaniel: I tried with `attributes`*, sorry I typed it wrong in GH

sauloperez: Oh, I figured. Anyway, it's not a big deal we just want to fix this failure.

luisramos0: you need this commit to make update_attributes work on on_demand, right?
https://github.com/openfoodfoundation/openfoodnetwork/pull/3132/commits/6cdaaa9e7d7e19dbcd997c29ba2b9648a229e82c

luisramos0: also, just a heads up, I think we have a few of these cases in the broken tests, the on_demand defaults to true in the factory but some tests expect it to be false to test stock levels.

PR #3134, opened 2018-11-23
Update Skylight from 1.6.1 to 1.7.1

sauloperez:
#### What? Why?

This doesn't bring too many improvements but it's the latest version can
work on our Rails 3.2 version.

You can read the changelog at
https://github.com/skylightio/skylight-ruby/blob/master/CHANGELOG.md#170-april-24-2018.

#### What should we test?
No change in behavior but we should check things work as expected after deploying to production.

#### Release notes

Update Skylight to the latest release supporting our Rails version 3.2.

Changelog Category: Changed

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -688,7 +688,7 @@ GEM
       json (>= 1.8, < 3)
       simplecov-html (~> 0.10.0)
     simplecov-html (0.10.2)
-    skylight (1.6.1)
+    skylight (1.7.1)
       activesupport (>= 3.0.0)
     spinjs-rails (1.3)
       rails (>= 3.1)


sauloperez:
Nope, due to the way it's declared

```rb
 gem 'skylight', '< 2.0'
```

mkllnk:
@Matt-Yorkley There are two schools when it comes to Gemfiles:

1. *No Gemfile.lock change without a Gemfile change.* A change in the gem versions should always have a reason. For example we upgrade a gem to gain a certain feature or fix a bug. That declaration goes into the Gemfile and affects Gemfile.lock.
2. *Keep the Gemfile as flexible as possible and use the Gemfile.lock to document the currently used versions.* This strategy makes it easier to keep all dependencies up-to-date even when it's not needed.

In my opinion we should combine both schools of thought. If we do a change because it's important to our software, we should make that clear in the Gemfile. If we only update to stay on top of things and avoid big changes, we can just update the Gemfile.lock.

sauloperez:
I also agree it's about a compromise of both and also, not all gems are so critical.

mkllnk:
Did you stage it somewhere @sauloperez? Even though this doesn't need user testing, it's always good to deploy gem changes first. Sometimes it blows up on system dependencies.

PR #3138, opened 2018-11-24
[Spree Upgrade] Merging master into 2-0-stable (third run in Nov2018)

luisramos0:
Same as previouss #3042
  Merging master into 2-0-stable. This is the result of:   

git checkout master  
git pull upstream master  
git checkout 2-0-stable
  git pull upstream 2-0-stable
  git checkout -b 2-0-stable-nov-24th
  git merge master

CONFLICT (content): Merge conflict in app/views/spree/admin/products/new.js.erb
     - Keep HEAD version (with wrapper) but update erb to haml
git add app/views/spree/admin/products/new.js.erb

CONFLICT (content): Merge conflict in Gemfile.lock
    - Keep HEAD in https://github.com/openfoodfoundation/spree.git
    - Keep only kgio (2.11.2), not kaminari/railties
    - Keep master in rainbow
git add Gemfile.lock

CONFLICT (content): Merge conflict in Gemfile
    - Keep HEAD

git add Gemfile

git commit
 git push origin 2-0-stable-nov-24th

I double checked the list of commits on GH and it looks good.


luisramos0:
The build with these changes from master fixed spec rspec ./spec/controllers/line_items_controller_spec.rb:21, only 11 broken tests now in #2929

config/locales/de_DE.yml
```diff
@@ -1140,13 +1139,13 @@ de_DE:
...
-  menu_1_url: "/shops"
+  menu_1_url: "/ shops"
   menu_2_title: "Karte"
   menu_2_url: "/map"
   menu_3_title: "Erzeuger"
-  menu_3_url: "/producers"
+  menu_3_url: "/ producers"
   menu_4_title: "Gruppen"
-  menu_4_url: "/groups"
+  menu_4_url: "/ groups"
```
sauloperez: Who is dealing with :de: translations? It's been already many times that I see this change, which is going to break their menus.

mkllnk: No sure. This one came from the machine translation. I fixed it now and left a comment. But it's difficult to explain this hack to translators.

sauloperez: Machine translation? Are they using google translate from Transifex without checking? It'd great to know whats happening because looks like is going to happen again.

mkllnk: Yes, Kirsten used machine translation to bootstrap the German locale. I think someone else is going through and reviewing it, but it's really difficult to tell in Transifex that this is a string that needs special attention.

Issue #3099, opened 2018-11-21
/shops taking 30+ seconds to load (UK)


luisramos0:
https://openfoodnetwork.org.uk/shops
consistently taking 30 secs to load

## Description
This is a bug.

## Expected Behavior
Page loads in "acceptable" time. I'd say we can close this if it's loads in under 10 seconds.

Edit: At #gathering19 we agreed 3-5 second load time is acceptable

## Actual Behavior
Page takes 30 secs to load.

## Steps to Reproduce
Go to 
https://openfoodnetwork.org.uk/shops
and wait 

## Context
@lin-d-hop says: "Fortunately most users direct straight to their shopfront so we dont get complaints about the slow page load"

## Severity
bug-s3

## Your Environment
* Version used: uk live Nov 20th
* Browser name and version: chrome latest
* Operating System and version (desktop or mobile): osx mojave

## Possible Fix
...

myriamboure:
Just tested on French production, it takes 20 sec to load from homepage to "shop" page.

sauloperez:
It's weird but I haven't found any trace of `ShopsController#index` in Skylight for UK. By checking on my machine this is what I see, and how I knew it `GET /shops` maps to that controller.

![screenshot from 2018-11-21 12-13-46](https://user-images.githubusercontent.com/762088/48837930-1e542900-ed87-11e8-8d01-66fbd84f0f84.png)

It's pretty obvious the performance hit is related to an N+1 on `producer_properties` and `spree_properties` database tables. See below

![screenshot from 2018-11-21 12-14-05](https://user-images.githubusercontent.com/762088/48838008-52c7e500-ed87-11e8-96d6-4edc915371ff.png)

```
Completed 200 OK in 996.5ms (Views: 790.2ms | ActiveRecord: 204.7ms)
```

sauloperez:
>  Could be that Skylight drops these requests due to them hitting a hard limit.

This message in `log/skylight.log` must be it

```
[SKYLIGHT] [1.6.1] failed to instrument span; msg=native Trace#start_span failed; endpoint=ShopsController#index
```

RachL:
> 
> 
> Just tested on French production, it takes 20 sec to load from homepage to "shop" page.

Could we agree on using Matomo for these stats :) ? Rather than our own individual network. Matomo shows an average generation time of 13 to 15 seconds. And like Lynne said, the main behaviour is not to reach this page.

luisramos0:
[...]

PR #3143, opened 2018-11-26
Enable rack-mini-profiler in development

sauloperez:
#### What? Why?

Partially solves #3099

This installs [rack-mini-profiler](https://github.com/MiniProfiler/rack-mini-profiler) profile the performance of the app in development. Although is better to profile in production mode, there's soooo much room for improvement in terms of performance that development mode is more than enough.

This is what I'm using to fully address #3099.

It generates reports like

captura de pantalla 2018-11-26 a les 19 56 41

#### Release notes

Add rack-mini-profiler to profile performance in development

Changelog Category: Added

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -150,4 +150,6 @@ group :development do
   # While we don't require this gem directly, no dependents forced the upgrade to a version
   # greater than 1.0.9, so we just required the latest available version here.
   gem 'eventmachine', '>= 1.2.3'
+
+  gem 'rack-mini-profiler', '< 1.0.0'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -584,6 +584,8 @@ GEM
     rack (1.4.7)
     rack-cache (1.8.0)
       rack (>= 0.4)
+    rack-mini-profiler (0.10.7)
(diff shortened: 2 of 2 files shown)


sauloperez:
I plan to add some doc in the wiki on how to profile the app once I finish.

As for staging, it won't since it's in the development's Gemfile groups.

luisramos0:
awesome @sauloperez this is very important! Let's get a wiki page started for this.
Is that method generating 294 SQL queries???? 😭

luisramos0:
I agree with @mkllnk let's stage this and see it's not running there.
and we can stsrt some documentation about it: @sauloperez how do I see it is running, how do I see it's not running? ;-)

sauloperez:
That screeshot I shared is the result of clicking on the little white box that shows up on the top-left corner of the screen. It displays the response times of the http requests.

luisramos0:
ok, I tested locally and it works, I can see the white box. Such a useful tool!!!

I bumped into a problem staging with semaphore to AUS. I will test this as soon as another staging server is available.

luisramos0:
ok, tested in staging FR, no sign of the profiler.
ready to go

PR #3144, opened 2018-11-27
Encode records with special chars as well

mkllnk:
Using Marshal.dump on the French production database raised an error:

    Encoding::UndefinedConversionError: "\xC3" from ASCII-8BIT to UTF-8

Replacing Marshal with YAML solves the problem. It is also more reliable
and human readable.

This code was run against the French, Australian and UK production
data successfully.

#### What should we test?

Nothing to test on staging.

#### Release notes

Fixed: Release v1.23.0 can be deployed on servers with non-English data.

Changelog Category: Fixed

#### Discourse thread

https://openfoodnetwork.slack.com/archives/CEBMTRCNS/p1543272844010600


pacodelaluna:
@pacodelaluna Thanks!

luisramos0:
already tested in production, so moving to ready to go.

PR #3152, opened 2018-11-28
 [Spree Upgrade] Fix product spec

### LINKED ISSUE #3137: [Spree Upgrade][BackOffice] Fix models/spree/product_spec (1 broken tests)
One remaining error in product_spec:

```
 28) Spree::Product validations and defaults does not save when master is invalid
      Failure/Error: product.on_hand = "10,000"
      
      NoMethodError:
        undefined method `on_hand=' for #<Spree::Product:0x0055806be4e620>
      # ./spec/models/spree/product_spec.rb:29:in `block (3 levels) in <module:Spree>'

```

Possible Fix:
This spec uses product.on_hand property to test a basic invalid product. Because on_hand is a special property, we should make this test use another property and test for it's invalidity on product creation, for example, product price.

This could be a solution (we would have to make an invalid price generate an error message, it's not at the moment). product_spec:29:

```
       it "does not save when master is invalid" do
        create(:stock_location)
        product = Product.new
        product.shipping_category = create(:shipping_category)
        product.supplier = create(:supplier_enterprise)
        product.primary_taxon = create(:taxon)
        product.name = "Apples"
        product.variant_unit = "weight"
        product.variant_unit_scale = 1000
        product.unit_value = 1

        # option 1 - broken price
        #product.price = "BROKEN PRICE"
        # option 2 - NO PRICE gets false but does not return any errors

        expect(product.save).to be false
        expect(product.errors[:price]).to include "is not a number"
      end
```

sauloperez:
#### What? Why?

Closes #3137

This test was checking the case of an invalid product's master variant. I switched it to check for one of the two validations we have on Variant.

### DIFF
--- a/spec/models/spree/product_spec.rb
+++ b/spec/models/spree/product_spec.rb
@@ -23,13 +23,11 @@ module Spree
       end
 
       it "does not save when master is invalid" do
-        s = create(:supplier_enterprise)
-        t = create(:taxon)
-        product = Product.new supplier_id: s.id, name: "Apples", price: 1, primary_taxon_id: t.id, variant_unit: "weight", variant_unit_scale: 1000, unit_value: 1
-        product.on_hand = "10,000"
-        expect(product.save).to be false
+        product = build(:product)
+        product.variant_unit = 'weight'
+        product.master.unit_value = nil
 
-        expect(product.errors[:count_on_hand]).to include "is not a number"
+        expect(product.save).to eq(false)
       end
 
       it "defaults available_on to now" do
--- a/app/models/spree/product_decorator.rb
+++ b/app/models/spree/product_decorator.rb
@@ -26,7 +26,6 @@ Spree::Product.class_eval do
   attr_accessible :variant_unit, :variant_unit_scale, :variant_unit_name, :unit_value
   attr_accessible :inherits_properties, :sku
 
-  # validates_presence_of :variants, unless: :new_record?, message: "Product must have at least one variant"
   validates_presence_of :supplier
   validates :primary_taxon, presence: { message: I18n.t("validation_msg_product_category_cant_be_blank") }
   validates :tax_category_id, presence: { message: I18n.t("validation_msg_tax") }, if: "Spree::Config.products_require_tax_category"


spec/models/spree/product_spec.rb
```diff
@@ -23,13 +23,11 @@ module Spree
...
-        t = create(:taxon)
-        product = Product.new supplier_id: s.id, name: "Apples", price: 1, primary_taxon_id: t.id, variant_unit: "weight", variant_unit_scale: 1000, unit_value: 1
-        product.on_hand = "10,000"
-        expect(product.save).to be false
+        product = build(:product)
+        product.variant_unit = 'weight'
+        product.master.unit_value = nil
 
-        expect(product.errors[:count_on_hand]).to include "is not a number"
+        expect(product.save).to eq(false)
```
sauloperez: it turns out `valid?` doesn't work for product and variant :sob: 

```ruby
[1] pry(#<RSpec::ExampleGroups::SpreeProduct::ValidationsAndDefaults>)> product.valid?
=> true
[2] pry(#<RSpec::ExampleGroups::SpreeProduct::ValidationsAndDefaults>)> product.master.valid?
=> true
[3] pry(#<RSpec::ExampleGroups::SpreeProduct::ValidationsAndDefaults>)> product.save!
ActiveRecord::RecordInvalid: Validation failed: Unit value can't be blank, Unit description can't be blank
from /home/pau/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activerecord-3.2.22.5/lib/active_record/validations.rb:56:in `save!'
[4] pry(#<RSpec::ExampleGroups::SpreeProduct::ValidationsAndDefaults>)>
```

God knows what we've broken...

mkllnk: That is weird. Same in Spree 1? I guess we will notice if this is a problem... :see_no_evil:

spec/models/spree/product_spec.rb
```diff
@@ -23,13 +23,11 @@ module Spree
...
-        s = create(:supplier_enterprise)
-        t = create(:taxon)
-        product = Product.new supplier_id: s.id, name: "Apples", price: 1, primary_taxon_id: t.id, variant_unit: "weight", variant_unit_scale: 1000, unit_value: 1
-        product.on_hand = "10,000"
-        expect(product.save).to be false
+        product = build(:product)
+        product.variant_unit = 'weight'
+        product.master.unit_value = nil
 
-        expect(product.errors[:count_on_hand]).to include "is not a number"
```
mkllnk: What was the reason to scrap this assertion?

sauloperez: We are no longer chechking count_on_hand's failure.

mkllnk: Okay, but why?

luisramos0: because it's not a core property of the product model any more. this initial test is testing a core property of the model.
I dont think we need to start the product model test by checking such an exceptional property like count_on_hand.

mkllnk: Ah, I see. :+1:

PR #3158, opened 2018-11-29
[Spree Upgrade] Scopes variants on order completion and on changes to completed orders

luisramos0:
#### What? Why?

See twin PR #3147 
Here we scope the variants before they are used for stock management, so that stock management is done on variant overrides (as prepared in #3147)

In terms of places we need to scope variants, there are quite a few:
- on order completion, scoping variants on order.finalize! is not needed, we need to scope variants in shipment.finalize as it reloads inventory_units.variants before moving stock (handled here)
- scoping variants on line_item.update_inventory (it's a before_save hook in line_item) before changes in line items of completed orders are synced with inventory  (handled here)
- there are quite a few more places we need to do this, we may leave that for a follow up PR

This is fixed with alias_method_chain in both cases, in both cases quite easy to understand ... what do you guys think? alternatives?

#### What should we test?
See #3147

### DIFF
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -127,6 +127,12 @@ Spree::LineItem.class_eval do
 
   private
 
+  def update_inventory_with_scoping
+    scoper.scope(variant)
+    update_inventory_without_scoping
+  end
+  alias_method_chain :update_inventory, :scoping
+
   # Override of Spree validation method
   # Added check for in-memory :skip_stock_check attribute
   def stock_availability
@@ -135,8 +141,7 @@ Spree::LineItem.class_eval do
   end
 
   def scoper
-	  return @scoper unless @scoper.nil?
-	  @scoper = OpenFoodNetwork::ScopeVariantToHub.new(order.distributor)
+    @scoper ||= OpenFoodNetwork::ScopeVariantToHub.new(order.distributor)
   end
 
   def calculate_final_weight_volume
--- a/app/models/spree/shipment_decorator.rb
+++ b/app/models/spree/shipment_decorator.rb
@@ -15,6 +15,18 @@ module Spree
       end
     end
 
+    # The shipment manifest is built by loading inventory units and variants from the DB
+    # These variants come unscoped
+    # So, we need to scope the variants just after the manifest is built
+    def manifest_with_scoping
+      manifest_without_scoping.each { |item| scoper.scope(item.variant) }
+    end
+    alias_method_chain :manifest, :scoping
+
+    def scoper
+      @scoper ||= OpenFoodNetwork::ScopeVariantToHub.new(order.distributor)
+    end
+
     private
 
     # NOTE: This is an override of spree's method, needed to allow orders


app/models/spree/line_item_decorator.rb
```diff
@@ -135,8 +141,7 @@ def stock_availability
   end
 
   def scoper
-	  return @scoper unless @scoper.nil?
-	  @scoper = OpenFoodNetwork::ScopeVariantToHub.new(order.distributor)
+    @scoper ||= OpenFoodNetwork::ScopeVariantToHub.new(order.distributor)
```
mkllnk: This could be quite inefficient. The scoper loads all variant overrides of an enterprise. Are we doing this for only one line item or for a long list of line items? Load time would grow with `#items * #overrides`.

mkllnk: A quick fix would be to load the override for this one line item and supply that to the scoper.

luisramos0: I don't think that's the scope of this PR or #3147 
This is the way of loading variant overrides in master right now: you load all of them once.
Do you think this needs to be changed? I'd open a separate issue for that.
Or am I missing something?

mkllnk: I think we always load variant overrides for a collection in master, all at once. Here you introduce a method that loads variant overrides individually, but you use the scoper that was designed for batch processing and loads all variant overrides. This pull request is introducing a performance issue by using the scoper differently to before. The quick fix would be only one additional line. It's not very efficient when doing that with a lot of variants, but much more efficient than the current code.

```suggestion
    @scoper ||= OpenFoodNetwork::ScopeVariantToHub.new(
      order.distributor,
      Hash[variant, VariantOverride.for(order.distributor, variant)]
    )
```

luisramos0: I don't get it :-)
I copied this code from here:
https://github.com/openfoodfoundation/openfoodnetwork/blob/66585e93a85f1d557a54345d52af1bac822f3a1f/app/models/spree/line_item_decorator.rb#L165-L168

and because variant_overrides is nil this will load all variants of the hub:
https://github.com/openfoodfoundation/openfoodnetwork/blob/66585e93a85f1d557a54345d52af1bac822f3a1f/lib/open_food_network/scope_variant_to_hub.rb#L3-L6

using the vo.indexed scope: https://github.com/openfoodfoundation/openfoodnetwork/blob/66585e93a85f1d557a54345d52af1bac822f3a1f/app/models/variant_override.rb#L30-L33

mkllnk: Ah, thank you! You are right that this a pre-existing problem, because it was used in the line item decorator already.

I still think it's very inefficient to load all variant overrides of a hub (potentially thousands) and then use only one of them. If that happens in a loop, we burn a lot of energy on nothing. But you are right, it's a pre-existing problem.

app/models/spree/shipment_decorator.rb
```diff
@@ -15,6 +15,18 @@ def update_adjustment_included_tax
       end
     end
 
+    # The shipment manifest is built by loading inventory units and variants from the DB
+    # These variants come unscoped
+    # So, we need to scope the variants just after the manifest is built
```
sauloperez: Weird line wrapping, no?

app/models/spree/line_item_decorator.rb
```diff
@@ -127,6 +127,12 @@ def sufficient_stock?
 
   private
 
+  def update_inventory_with_scoping
+    scoper.scope(variant)
+    update_inventory_without_scoping
+  end
+  alias_method_chain :update_inventory, :scoping
```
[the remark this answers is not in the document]
luisramos0: are you saying that on alias_method_chain we should call super instead of <method>_without_<modification> (update_inventory_without_scoping)? I dont think so, all uses and examples I see make use of the "without" call.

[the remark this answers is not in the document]
sauloperez: Sorry I was on my phone and I couldn't share a suggestion. What I meant was

```ruby
def update_inventory                                                                                                                                                                                         
 scoper.scope(variant)                                                                                                                                                                                      
 super                                                                                                                                                                                                      
end   
```

But this returns `NoMethodError: super: no superclass method `update_inventory' for #<Spree::LineItem:0x0055cc5b3a0610>`. Thanks `class_eval`... :sob:

PR #3164, opened 2018-12-03
[Spree Upgrade] Merging master into 2-0-stable (first run in Dec2018)

luisramos0:
Same as previouss #3138
   Merging master into 2-0-stable. 
This is the result of:   

git checkout master   
git pull upstream master   
git checkout 2-0-stable   
git pull upstream 2-0-stable   
git checkout -b 2-0-stable-dec-3rd   
git merge master
NO CONFLICTs :tada:
  git push origin 2-0-stable-dec-3rd

I double checked the list of commits on GH and it looks good.

The build looks good! There's just one extra PR we will need to remove Shipments from the recently de-defaced app/views/spree/admin/shared/_order_tabs.html.haml


mkllnk:
@luisramos0 Have you created an issue for removing shipments from that tab?

luisramos0:
nice, thanks for sharing the git commands.
it's a lot less characters, the result is the same but your local master and 2-0-stable branches are not updated, right? you would need a pull or merge to get them updated, correct?

mkllnk:
> your local master and 2-0-stable branches are not updated, right? you would need a pull or merge to get them updated, correct?

That's correct. I actually had a long phase of not having a local master branch and always referring to upstream/master. That prevented me from using a stale branch for from accidentally working on it and pushing it back to upstream.

At the moment I use a local master branch to push it to our production server for deploys. But once that's changed (ofn-install + Semaphore?), I can get rid of a local master again.

luisramos0:
ah, no local master... that is interesting!!! I am not sure I need to change my workflow but it's definitely good to know! thanks for sharing.

PR #3165, opened 2018-12-03
Fix send confirmation instructions specs

### LINKED ISSUE #3028: [Spree Upgrade][FrontOffice] Fix specs that send_confirmation_instructions (4 broken tests)
```
 69) Spree::User#create should send a confirmation email
      Failure/Error:
        expect do
          create(:user, email: 'new_user@example.com', confirmation_sent_at: nil, confirmed_at: nil)
        end.to send_confirmation_instructions
      
        expected #<Proc:0x007fe5b6b83820@/home/runner/openfoodnetwork/spec/models/spree/user_spec.rb:79> to send confirmation instructions
      # ./spec/models/spree/user_spec.rb:79:in `block (3 levels) in <top (required)>'

```
4 broken tests:
spec/features/consumer/account/settings_spec.rb:26
spec/controllers/admin/manager_invitations_controller_spec.rb:34
spec/features/admin/users_spec.rb:38
spec/features/consumer/authentication_spec.rb:9

I am not sure why but these tests don't always fail in the build.

sauloperez:
#### What? Why?

Closes #3028

These tests rely on the email being sent but ActionMailer's `perform_delivery` was set to false and as a result, the `deliveries` array was not populated. However, the emails were still working.


spec/features/consumer/authentication_spec.rb
```diff
@@ -1,14 +1,9 @@
 require 'spec_helper'
 
-feature "Authentication", js: true, retry: 3 do
+feature "Authentication", js: true do
```
luisramos0: 🤞

spec/features/consumer/account/settings_spec.rb
```diff
@@ -21,6 +21,8 @@
     end
 
     it "allows the user to update their email address" do
+      ActionMailer::Base.perform_deliveries = true
```
luisramos0: is this not something we can put in the EmailHelper? next to setup_email?

sauloperez:
Check it out @luisramos0 . I think this new helper method besides safer makes it a bit more evident to the reader.

PR #3166, opened 2018-12-03
[Spree Upgrade] Remove shipments option from orders tab (gone in spree 2)

luisramos0:
The admin orders page has been de-defaced and is now on ofn code as haml.
That de-deface work was merged in #3164. We now need to adapt it to spree 2 by removing the shipments tab that is gone in spree 2.

I've checked the build and it's looking good. This PR and #3164 move us from 87 to 84 broken tests.


luisramos0:
Rebased to get new build stats.

Issue #3096, opened 2018-11-21
[Spree Upgrade] Google Maps "NoApiKeys" warning in feature specs


HugsDaniel:
## Description

When running feature specs in 2-0-stable, I get this error : 
`Google Maps JavaScript API warning: NoApiKeys https://developers.google.com/maps/documentation/javascript/error-messages#no-api-keys`

I last got it when running `spec/features/consumer/shopping/checkout_spec.rb`, but it's the same for several feature specs. 

It doesn't make the specs fail, but it takes a lot of time by retrying 3 or 4 times to connect to Google API.

As @sauloperez and @mkllnk suggested, I open this issue so that we can investigate how not to depend on internet connection for specs.

sauloperez:
IMO this is a perfect fit for https://github.com/bblimke/webmock

mkllnk:
This is not a warning anymore: https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/master/builds/347

Our build is broken. We need to fix this as soon as possible.

luisramos0:
broken build, we stop the "production line" https://en.wikipedia.org/wiki/Andon_(manufacturing) :-)

sauloperez:
I gave it a try and Webmock can't help us in this case since it's not a regular HTTP request made through a Ruby HTTP client but a <script> tag.

I believe the culprit is https://github.com/openfoodfoundation/openfoodnetwork/blob/66585e93a85f1d557a54345d52af1bac822f3a1f/app/views/layouts/darkswarm.html.haml#L19.

I also tried adding a guard clause with `Rails.env.production?` but darkswarm depends on the `google` object. See below:

```
  1) Shops searches by URL
     Failure/Error: visit shops_path(anchor: "/?query=xyzzy")

     Capybara::Poltergeist::JavascriptError:
       One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details).

       ReferenceError: Can't find variable: google
       ReferenceError: Can't find variable: google
           at http://127.0.0.1:50035/assets/darkswarm/all.js:67070 in global code
       ReferenceError: Can't find variable: Darkswarm
       ReferenceError: Can't find variable: Darkswarm
           at http://127.0.0.1:50035/assets/web/all.js:2
       ReferenceError: Can't find variable: I18n
       ReferenceError: Can't find variable: I18n
[...]

PR #3168, opened 2018-12-04
Temporarily xit 2 broken specs (shops and producers search by url)

luisramos0:
#### What? Why?

Temp fix for #3096 so we can have a green build meanwhile.

#### What should we test?
Build should be green again.


luisramos0:
yes, pending is more appropriate. done ;-)

luisramos0:
it worked. ready to be merged.

Issue #3061, opened 2018-11-16
Wrong URL in footer leading to error 500


RachL:
## Description
In the footer there is a link to register. When using the link from a buyer account profile, the link leads to a page that does not exist.

## Expected Behavior
The link should leads to /register

## Actual Behavior
The link leads to /signup which does not exist

## Steps to Reproduce
1. From the home page, scroll down to the footer page and try the link "Register here" which should work perfectly (meaning the link leads you to the /register page)
2. Log in as a buyer
3. Go to your profile "my account"
4. Scroll down to the footer and try again the same link
5. A new tab opens and shows you the /signup page which does not exist so you get an 500 error and the beautiful snail

## Animated Gif/Screenshot
![image](https://user-images.githubusercontent.com/6525576/48622399-ed4eaf80-e9a6-11e8-9e10-8a4e2b513ee1.png)

![image](https://user-images.githubusercontent.com/6525576/48621924-6ea54280-e9a5-11e8-9490-7c77d41d1cda.png)

## Severity
bug-s4: it's annoying, but you can use it

I'm not sure many user will transform their account this way, but this happened in France during a test.

## Your Environment
* Version used: v1.22
* Browser name and version: Firefox 63
* Operating System and version (desktop or mobile): Ubuntu 18 LTS and Android 7 mobile

## Possible Fix
Change the URL in footer to /register instead of /signup

PR #3174, opened 2018-12-04
Fix broken registration link in footer

haseleyi:
#### What? Why?

Closes [#3061](https://github.com/openfoodfoundation/openfoodnetwork/issues/3061)

We have a broken link: In the account page for a buyer account, the footer features a registration call partial that routes the user to /signup instead of /register. 
This fixes things. But I'm not satisfied with it: why would registration_path resolve to /signup on only this page (or at all for that matter)? I'd rather address the root of the problem.

#### What should we test?

In a buyer account, navigate to My Account > Orders and click "Register here" in the footer.

#### Release notes

Fixed a broken registration link in the footer of our My Account page.

Changelog Category: Fixed

### DIFF
--- a/app/views/shared/_register_call.html.haml
+++ b/app/views/shared/_register_call.html.haml
@@ -1,5 +1,9 @@
 .alert-cta
   %h6
+    -# Please forgive the hard-coded link:
+       The more elegant 'registration_path' resolves to /signup due to spree_auth_device > config > routes.rb
+       This is one of several possible fixes. Long-term, we'd like to bring the accounts page into OFN.
+       View the discussion here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3174
     %a{href: "/register", target: "_blank"}
       = t '.selling_on_ofn'
       &nbsp;
--- a/app/views/shared/_register_call.html.haml
+++ b/app/views/shared/_register_call.html.haml
@@ -1,6 +1,6 @@
 .alert-cta
   %h6
-    %a{href: registration_path, target: "_blank"}
+    %a{href: "/register", target: "_blank"}
       = t '.selling_on_ofn'
       &nbsp;
       %strong
--- a/app/views/shared/_register_call.html.haml
+++ b/app/views/shared/_register_call.html.haml
@@ -1,9 +1,9 @@
 .alert-cta
   %h6
     -# Please forgive the hard-coded link:
-       The more elegant 'registration_path' resolves to /signup due to spree_auth_device > config > routes.rb
-       This is one of several possible fixes. Long-term, we'd like to bring the accounts page into OFN.
-       View the discussion here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3174
+    -# The more elegant 'registration_path' resolves to /signup due to spree_auth_device > config > routes.rb
+    -# This is one of several possible fixes. Long-term, we'd like to bring the accounts page into OFN.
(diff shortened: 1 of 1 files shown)


luisramos0:
good question: "why would registration_path resolve to /signup on only this page (or at all for that matter)?"
tricky one. Some clues:
this dependency: https://github.com/openfoodfoundation/openfoodnetwork/blob/1052d982550f0b7ac601b399f032291383ddad26/Gemfile#L18
takes you to this route:
https://github.com/openfoodfoundation/spree_auth_devise/blob/da9eecefc6fe13dedf4c6f3febec79caad839ec3/config/routes.rb#L18

Because the account page where you are is also on this routes file:
https://github.com/openfoodfoundation/spree_auth_devise/blob/da9eecefc6fe13dedf4c6f3febec79caad839ec3/config/routes.rb#L34
I think the registration_path defined on this routes file with /sign_up takes precedence the ofn's one in:
https://github.com/openfoodfoundation/openfoodnetwork/blob/1052d982550f0b7ac601b399f032291383ddad26/config/routes.rb#L19

There's also this devise_scope that I dont really know how it works.

I think we can consider this is a bug in OFN because we should not have re-written this registration named route. I dont think there is a good reason to do this but i am not 100% sure.

The proper fix could be to rename the registration path in the ofn context to separate it from the spree devise registration path.
Or maybe an easier fix would be to overwrite the devise registration path in ofn's routes.

Let's see what reviewers say.

mkllnk:
Very good detective work @luisramos0. Most front-end pages are our own implementation, but we are still using Spree's UsersController and override the main action. I guess that Spree's controller is referring to Spree routes instead of OFN routes. So I think we could solve this problem by implementing our own account page. It seems out of scope, but is actually a fix we want to do anyway.

So we have some options now:

- Explicitly link `/register`.
- Rename our register path and use that.
- Overwrite Spree's register path.
- Add a route to redirect from `/signup` to `/register`.
- Move accounts page into the OFN namespace and abandon Spree's controller.

I'm surprised that there are so many solutions to this simple problem. I'm voting for the last one, but would accept the current pull request until that is implemented. A working link wins over code style. It would be nice to add a comment though. Otherwise someone could have the brilliant idea to change it back to `registration_path` again, because that's the right way to do it.

luisramos0:
thanks @mkllnk 
I agree. That means we are approving this PR as long as a comment is added :-)

haseleyi:
Thank you both for the excellent breakdown of the issue and potential solutions. This is a thoughtful and inspiring community to learn from!

app/views/shared/_register_call.html.haml
```diff
@@ -1,6 +1,10 @@
 .alert-cta
   %h6
-    %a{href: registration_path, target: "_blank"}
+    -# Please forgive the hard-coded link:
+    -# The more elegant 'registration_path' resolves to /signup due to spree_auth_device > config > routes.rb
+    -# This is one of several possible fixes. Long-term, we'd like to bring the accounts page into OFN.
+    -# View the discussion here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3174
```
sauloperez: :ok_hand:

RachL:
Deploying in AU staging (yes I jumped one issue, but I'm in a meeting in 15 minutes, so I figured rather I test 1 small one rather than nothing :) )

RachL:
Link redirects correctly to https://staging.openfoodnetwork.org.au/register ! ready to go :)

Issue #12, opened 2013-07-22
Mailing list does not exist


wvengen:
I tried to signup for the mailing list mentioned in [Help-Out](https://github.com/eaterprises/openfoodweb/wiki/Help-Out), but that message bounced. It would be helpful if that would be corrected.

Issue #11, opened 2013-07-22
Error on first time checkout


dacook:
Immediately after seeding a fresh db: As a customer, I should be able to enter an address when checking out. 
It fails with error: 
        ActiveRecord::RecordNotFound in Spree::CheckoutController#update
        Couldn't find Spree::Address with ID=3 for Spree::Order with ID=1

Issue #10, opened 2013-07-12
Admin: 'New Product' button doesn't work when editing product.


dacook:
admin/products/{id}/edit

As an administrator, when I'm editing a product, and I click on 'New Product', I expect to be able to create a new product.
Instead it says loading then nothing happens.. Spree bug?

PR #3189, opened 2018-12-06
Make all mailers use appropriate locale when sending emails

### LINKED ISSUE #3043: [Multi language] Emails are being sent on instance default language, not on users selected locale
## Description
As a Portuguese user of Katuma (Iberian instance of OFN: https://alpha.katuma.org)
I need to receive emails in Portuguese
So that I can interact with the platform successfully

## Expected Behavior
Emails in Portuguese.

## Actual Behavior
Emails in Castellano.

## Steps to Reproduce
Given I go on Katuma and select Portuguese language
When I register as a new user
Then I get an email in Portuguese to confirm registration

## Context
Multi language instances #2808 

## Severity
bug-s3: it's a blocking issue for Portuguese users.

## Your Environment
Any multi language instance.

## Possible Fix
The user locale is already being saved on the database when the user selects her language. See "select locale from spree_users" and I18nHelper.set_locale
`spree_current_user.update_attributes!(locale...`

We need to use this when sending emails.
One particular test case is when the user selects language before logging in, we need to ensure emails are still sent in the users language upon registration. See Acceptance Criteria above "Steps to Reproduce" where the user is not logged in.

luisramos0:
#### What? Why?

Closes #3043 

In order to send emails in the correct locale we need to select the appropriate locale for each email type, for example, order confirmation should come from the user of the order.

Note for reviewers:
- I am happy to try a different approach. I could not extract/abstract I18n.with_locale because it's needed for both mail _and_ for the t calls in subject.
- I could test this locale change in all mailers (not just user mailer). It would be too much noise?

#### What should we test?
All emails need to be tested. In the default language of the instance used for testing and also in a  language different from the instance default.

#### Release notes
Changelog Category: Added

Emails all now sent in specific locales related to the users, not the instance default language.

### DIFF
--- a/app/controllers/user_registrations_controller.rb
+++ b/app/controllers/user_registrations_controller.rb
@@ -8,22 +8,22 @@ class UserRegistrationsController < Spree::UserRegistrationsController
   # POST /resource/sign_up
   def create
     @user = build_resource(params[:spree_user])
-    if resource.save
-      session[:spree_user_signup] = true
-      session[:confirmation_return_url] = params[:return_url]
-      associate_user
+    unless resource.save
+      return render_error(@user.errors)
+    end
 
-      respond_to do |format|
-        format.html do
-          set_flash_message(:success, :signed_up_but_unconfirmed)
-          redirect_to after_sign_in_path_for(@user)
-        end
-        format.js do
-          render json: { email: @user.email }
-        end
+    session[:spree_user_signup] = true
+    session[:confirmation_return_url] = params[:return_url]
+    associate_user
+
+    respond_to do |format|
+      format.html do
+        set_flash_message(:success, :signed_up_but_unconfirmed)
+        redirect_to after_sign_in_path_for(@user)
       end
-    else
-      render_error(@user.errors)
+      format.js do
+        render json: { email: @user.email }
+      end
     end
   rescue StandardError => error
     OpenFoodNetwork::ErrorLogger.notify(error)
--- a/app/controllers/user_registrations_controller.rb
+++ b/app/controllers/user_registrations_controller.rb
@@ -5,9 +5,13 @@ class UserRegistrationsController < Spree::UserRegistrationsController
 
   before_filter :set_checkout_redirect, only: :create
 
+  include I18nHelper
+  before_filter :set_locale
+
   # POST /resource/sign_up
   def create
     @user = build_resource(params[:spree_user])
+    @user.locale = I18n.locale.to_s
     unless resource.save
       return render_error(@user.errors)
     end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -1,7 +1,7 @@
 module I18nHelper
   def set_locale
     # Save a given locale
-    if params[:locale] && Rails.application.config.i18n.available_locales.include?(params[:locale])
+    if params[:locale] && available_locale?(params[:locale])
       spree_current_user.update_attributes!(locale: params[:locale]) if spree_current_user
       cookies[:locale] = params[:locale]
     end
@@ -12,5 +12,19 @@ module I18nHelper
     end
 
     I18n.locale = spree_current_user.andand.locale || cookies[:locale] || I18n.default_locale
+  end
+
+  def valid_locale(locale)
+    if available_locale?(locale)
+      locale
+    else
+      I18n.default_locale
+    end
+  end
+
+  private
+
+  def available_locale?(locale)
+    Rails.application.config.i18n.available_locales.include?(locale)
   end
 end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -15,7 +15,9 @@ module I18nHelper
   end
 
   def valid_locale(object_with_locale)
-    if object_with_locale.present? && object_with_locale.locale.present? && available_locale?(object_with_locale.locale)
+    if object_with_locale.present? &&
+       object_with_locale.locale.present? &&
+       available_locale?(object_with_locale.locale)
       object_with_locale.locale
     else
       I18n.default_locale
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -14,9 +14,9 @@ module I18nHelper
     I18n.locale = spree_current_user.andand.locale || cookies[:locale] || I18n.default_locale
   end
 
-  def valid_locale(locale)
-    if available_locale?(locale)
-      locale
+  def valid_locale(object_with_locale)
+    if object_with_locale.present? && object_with_locale.locale.present? && available_locale?(object_with_locale.locale)
+      object_with_locale.locale
     else
       I18n.default_locale
     end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -14,11 +14,11 @@ module I18nHelper
     I18n.locale = spree_current_user.andand.locale || cookies[:locale] || I18n.default_locale
   end
 
-  def valid_locale(object_with_locale)
-    if object_with_locale.present? &&
-       object_with_locale.locale.present? &&
-       available_locale?(object_with_locale.locale)
-      object_with_locale.locale
+  def valid_locale(user)
+    if user.present? &&
+       user.locale.present? &&
+       available_locale?(user.locale)
+      user.locale
     else
       I18n.default_locale
     end
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -17,9 +17,32 @@ describe Spree::UserMailer do
     setup_email
   end
 
-  it "sends an email when given a user" do
-    Spree::UserMailer.signup_confirmation(user).deliver
-    ActionMailer::Base.deliveries.count.should == 1
+  describe '#signup_confirmation' do
+    it "sends email when given a user" do
+      Spree::UserMailer.signup_confirmation(user).deliver
+      expect(ActionMailer::Base.deliveries.count).to eq(1)
+    end
+
+    describe "user locale" do
+      around do |example|
+        original_default_locale = I18n.default_locale
+        I18n.default_locale = 'pt'
+        example.run
+        I18n.default_locale = original_default_locale
+      end
+
+      it "sends email in user locale when user locale is defined" do
+        user.locale = 'es'
+        Spree::UserMailer.signup_confirmation(user).deliver
+        expect(ActionMailer::Base.deliveries.first.body).to include "Gracias por unirte"
+      end
+
+      it "sends email in default locale when user locale is not available" do
+        user.locale = 'cn'
+        Spree::UserMailer.signup_confirmation(user).deliver
+        expect(ActionMailer::Base.deliveries.first.body).to include "Obrigada por juntar-se"
+      end
+    end
   end
 
   # adapted from https://github.com/spree/spree_auth_devise/blob/70737af/spec/mailers/user_mailer_spec.rb
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -845,7 +845,6 @@ Layout/SpaceInsideHashLiteralBraces:
     - 'spec/features/admin/reports_spec.rb'
     - 'spec/features/consumer/shopping/checkout_spec.rb'
     - 'spec/helpers/checkout_helper_spec.rb'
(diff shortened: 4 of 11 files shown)


luisramos0:
Thank @mkllnk (I put numbers on your comments):
1. yes, it could be with_valid_locale,  I prefer to keep it as is and have helper.valid_locale available for use in other places
2. I tried to do that by wrapping the mail() method but each action has a t() call for the subject as well. So, it would have to be a wrapper that takes the base subject text and the subject translation key to be translated inside the wrapper... I couldn't find a simple solution for this!
3. that's a good idea but it would involve one extra query to the DB

I feel we can skip suggestions 1 and 3. I would love to implement suggestion 2, any idea how to integrate the subject translation in there?

mkllnk:
Interesting. I'm totally fine to ignore point 1. And now I do see the difficulty in point 2. It would only be possible if we did point 3 as well. And that would include another database query.

The closest to your attempt to wrap the `mail` method would be to provide a block for the subject that is evaluated later. Not a good solution either.

I'm sorry, I can't think of a better solution than the one you already implemented. :-)

spec/mailers/user_mailer_spec.rb
```diff
@@ -17,9 +17,32 @@
     setup_email
   end
 
-  it "sends an email when given a user" do
-    Spree::UserMailer.signup_confirmation(user).deliver
-    ActionMailer::Base.deliveries.count.should == 1
+  describe '#signup_confirmation' do
+    it "sends email when given a user" do
+      Spree::UserMailer.signup_confirmation(user).deliver
+      ActionMailer::Base.deliveries.count.should == 1
```
sauloperez: :warning: let's use `expect` here

sauloperez:
> 2- We are adding a nested block in every mailer action. We can remove that code duplication by using the around_action callback of mailers: edgeguides.rubyonrails.org/action_mailer_basics.html#action-mailer-callbacks. I'm not sure if that is available in our ruby version. Is there a way to implement it?

Unfortunately it's not available on Rails 3.2 :point_right: https://guides.rubyonrails.org/v3.2/action_mailer_basics.html

app/helpers/i18n_helper.rb
```diff
@@ -13,4 +13,20 @@ def set_locale
 
     I18n.locale = spree_current_user.andand.locale || cookies[:locale] || I18n.default_locale
   end
+
+  def valid_locale(object_with_locale)
+    if object_with_locale.present? &&
+       object_with_locale[:locale].present? &&
+       available_locale?(object_with_locale[:locale])
+      object_with_locale[:locale]
```
sauloperez: I think I'm missing something. Why do we use `[:locale]` instead of `.locale`? As far as I see the object passed is always a user object.

luisramos0: [] works for objects and hashes, right? this way it also works for hashes and, for example, I can write the tests without a user object, only a hash.

spec/mailers/user_mailer_spec.rb
luisramos0: :+1:

app/helpers/i18n_helper.rb
sauloperez: Isn't that YAGNI?

luisramos0: I was just being lazy, instead of using a user obj to test I wanted to test with a hash with only.
Anyway, I changed it, pls have a look.

sauloperez:

RachL:
@sauloperez @luisramos0 how well is Portuguese translated on Katuma. Emails are working fine, but even if I change to Portuguese, I keep getting emails in Spanish. Except switching language like this:
image

Is there another way to test it?

Testing notes so far: https://docs.google.com/document/d/1BajRK6yaTEwt7Z9oFS9dzHdUFdYoJGUxmBMC0TkaLCA/edit#

RachL:
Btw shiping email stays in English but this is a known/re-existent issue.

luisramos0:
I'll have a look.

luisramos0:
@RachL I see registration and password emails in Spanish but order confirmation email is in Portuguese. The guest checkout order confirmation email is in Spanish but the order from a logged in user is in Portuguese. Can you please check again so I am sure we are seeing the exact same thing.

RachL:
@luisramos0 yes I've check only with guest the second time you're right :/

luisramos0:
can you pls confirm that on checkout with logged in user you get a email in Portuguese? thanks

RachL:
@luisramos0 sorry if I was unclear :( : yes I confirm that I received an email in Portuguese.

sauloperez:
Then this is ready to go, am I right?

myriamboure:
@RachL or @luisramos0 can tell I didn't get the conversation ;-)

sigmundpetersen:
The tests are not all green in the testing notes either, so I wouldn't bet on it.

RachL:
@luisramos0 would it help if we release this PR as it is and then create a new issue for non logged in user? Right now this PR improves still a little bit the current situation.

luisramos0:
yes, I think we can do that but first I'd need you to test the other emails.
there are other emails that should be working that I dont think you have tested.

Let's review all emails sent by OFN (we may copy this to a wiki page afterwards).

Broken here means Email in Spanish, and Working means Email in Portuguese.

The ones you have tested:
- order confirmation for customer as GUEST - broken
- order confirmation for customer as GUEST - works!
- signup confirmation - broken
- user welcome email - **when is this one trigger? is it broken?**
- password change - broken
- order shipped - broken
- subscription - **broken?**

The ones not tested (most of these should work with this PR):
- enterprise.welcome (after creation on admin and also after registration)
- enterprise.manager_invitation (admin > enterprises)
- order cycle report email (I dont know where this one is trigger, I have only tested the email part, not the trigger from the UI)
- order cancelation email
- order confirmation for shop
- invoice_email

Shall we test these?

sauloperez:
> user welcome email - when is this one triggered? is it broken?

AFAIK that's what you get once you confirm the account.

luisramos0:
So, OFN sends 17 different emails after all 😂 
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Emails

There are **20 cases** listed in the wiki page.

This PR originally fixed 9 cases: #4.1, #4.2, #5.1, #5.2, #6, #8, #9, #10.1 and #11
@RachL originally tested 6 cases (all broken): #1, #2, #3, #5.1 (guest user), #7 and #12 and then confirmed #5.1 was working for logged in user.
These last commits fixed 2 cases: #1 and #2 (only tested locally so far).

**So currently, this PR fixes 11 cases 1, 2, 4.1, 4.2, 5.1, 5.2, 6, 8, 9, 10.1 and 11.** I suggest we make this list the scope of this issue.
**Here's a list of the remaining 9 cases: 3, 7, 10.2 and 12-17** I suggest we create issues for these, see details below:

**3. BROKEN forgotten password (not working in multi-language)**
This is done in devise (a dependency of spree) - I'll open an separate issue to take care of this one (the email itself is not styled, so we may also need to redo the email template).

**7. NOT TESTED order shipped**
I don't know how to test this one...

**10.2 BROKEN - enterprise manager - new account sign up email**
there's no way to know whats the new user locale, best option is to user current user locale. I'll open a separate issue to do that.

** 12. BROKEN subscription confirmation email**
I havent tested this but I am pretty sure the problem will be caused by the fact the customer is created with the subscription so there's no way of knowing it's locale. I'll create an issue to handle this and make it use the locale of the user creating the subscription.

** GUEST user emails **
Note about guest user: guest user is not registered in the DB so there is no place to store the user locale (I thought of using the session locale but the email is sent on delayedJob which is a different process without access to the session locale) - I'll open a separate issue for this one as well.
The same problem applies on resend confirmation in the backoffice and order cancelation.

luisramos0:
First these last 2 commits need to be reviewed, then we can restage and re-test.

app/controllers/user_registrations_controller.rb
```diff
@@ -8,22 +8,22 @@ class UserRegistrationsController < Spree::UserRegistrationsController
   # POST /resource/sign_up
   def create
     @user = build_resource(params[:spree_user])
-    if resource.save
-      session[:spree_user_signup] = true
-      session[:confirmation_return_url] = params[:return_url]
-      associate_user
+    unless resource.save
+      return render_error(@user.errors)
```
mkllnk: This could be a one-liner now:
```rb
return render_error(@user.errors) unless resource.save
```

`resource` and `@user` are the same, right?

mkllnk:
About the guest emails:

I think the clearest way to use Delayed Job is to create custom jobs for whatever we want to delay. These jobs can store the current locale of the guest and use it for sending the email.

In some cases I also wonder if we should send the email straight away. I know that sending emails from unicorn blocks that process and adds a delay for the user. But e.g. when the user resets the password, they can't do anything before receiving the email anyway. In this case delayed job lets the user wait. It would be great to document in the wiki which emails are delayed and which aren't.

sauloperez:
> In some cases I also wonder if we should send the email straight away. I know that sending emails from unicorn blocks that process and adds a delay for the user.

Not only the user but any other request. And this is specially relevant because we have a single app server.

IMHO all requests issued from the server must be handled from a background job, and this includes emails.

luisramos0:
thanks @mkllnk for the input, I'll create the guest user issue and add that info to the issue (that locale can be stored in the job).
in terms of delaying emails or not, I agree with Pau, all emails should be DelayedJobs.

luisramos0:
I am staging this in FR and give it a good try myself. I will post results here.

luisramos0:
I tested in staging FR and it's looking good on the log files. Unfortunately delayed job is not working properly in staging FR. We need to test this in staging ES or AUS.

luisramos0:
all right, staged in ES: https://staging.katuma.org

I tested the frontoffice part and it's looking good, my testing notes:
https://docs.google.com/document/d/18s6gQUGQUEKA4qXt9IAYyYrYs2lF2vlANG3UOD_01yc/edit?usp=sharing

I covered cases 1, 2, 4.2, 5.1 and 6.
We need to cover 4.1, 5.2, 8, 9, 10.1 and 11.
reference link: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Emails

RachL:
@luisramos0 thank you very much :) I've updated the wiki by just removing the "not be able to test the email" mention, so that it stays a list of all email send, with the broken part.

A summary of my tests:
4.1. enterprise welcome email - triggered after enterprise creation on admin/enterprises
I received the email in English. So I'm not sure, is it just not translated on Transifex?

5.2 order confirmation - triggered as resend in the backoffice
Yes! It works :)

8. order cancelation email
I see some words in Portuguese, so my guess is it works? See screenshot in the doc

**9. order invoice email **
It says “ Sally's Farm must have a valid ABN before invoices can be sent.” I don’t know what it means :( Not sure if this is really used...

**10.1 enterprise manager invite - new account - invitation email**
Never got the email

**11. order cycle producers notification (not working in multi-language)**
Never got the email

My guess is that we can ship it as it is and create issues for what's left. What do you think?

Screenshots at the top of my doc:

https://docs.google.com/document/d/1BajRK6yaTEwt7Z9oFS9dzHdUFdYoJGUxmBMC0TkaLCA/edit#

luisramos0:
Thanks!
Yes please, I'd merge this code as is and create issues for the small different things we find now.
Can this still be included in this release? If so I'd merge right away to get this in katuma before Jan 2019 orders start to come in in Portugal :-)

RachL:
@luisramos0 tbh I'm a bit lost with the release... It is already published right? I'm waiting for Aus staging to be free to ask it to be staged.So I guess we would need to wait for a new one? Or not do a release testing on the latter, but only on the future one combining the two?

Matt-Yorkley:
@RachL we can stage and test the current release on Aus staging when it's free.

app/controllers/user_registrations_controller.rb
luisramos0: sorry @mkllnk I forgot this comment...
yes, could be a one liner. but I also like this pattern because the one liner moves this major point in the happy path of the method (resource.save) to the end of the line...

yes, I think resource and @user are the same but I didnt want to touch it. anyway this assignment to @user and then usage of resource in the next line is very very ugly...

because this is already ready to go I'll skip changing these things.

luisramos0:
ok, I have created an epic to represent what still needs to be done: https://github.com/openfoodfoundation/openfoodnetwork/issues/3243

This one is definitely ready to go :-)

Issue #3122, opened 2018-11-22
[Spree Upgrade][BackOffice] Fix features/admin/products_spec (3 broken tests)


luisramos0:
Apart from #3018, fix features/admin/products_spec

```ruby
  36) 
    As an admin
    I want to set a supplier and distributor(s) for a product
 as an enterprise user editing a product
      Failure/Error: select 'Permitted Supplier', from: 'product_supplier_id'
      
      Capybara::ElementNotFound:
        Unable to find visible select box "product_supplier_id" that is not disabled
      # ./spec/features/admin/products_spec.rb:141:in `block (3 levels) in <top (required)>'

  37) 
    As an admin
    I want to set a supplier and distributor(s) for a product
 as an enterprise user deleting product properties
      Failure/Error: page.all('a.remove_fields').first.click
      
      NoMethodError:
        undefined method `click' for nil:NilClass
      # ./spec/features/admin/products_spec.rb:211:in `block (3 levels) in <top (required)>'

  38) 
    As an admin
    I want to set a supplier and distributor(s) for a product
 as an enterprise user deleting product images
      Failure/Error: defined_response.call
      
      NoMethodError:
        undefined method `admin_images_url' for #<ActionDispatch::Routing::RoutesProxy:0x007fb8577566f8>
      # ./lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `call'
      # ./lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `respond_with'
      # ./lib/open_food_network/rack_request_blocker.rb:37:in `call'
      # ------------------
      # --- Caused by: ---
      # Capybara::ExpectationNotMet:
      #   expected not to find visible css "table[data-hook='images_table'] td img", found 1 match: ""
      #   ./spec/features/admin/products_spec.rb:233:in `block (3 levels) in <top (required)>'
```

luisramos0:
#3197 fixed the first test, the two remaining ones are fixed in #3203

daniellemoorhead:
Shall we move this issue back to In Dev column then @luisramos0, because you're likely to have more PRs for it? I don't think there's anything to review on it right now.

#toottoot

luisramos0:
It's fixed!

Issue #3114, opened 2018-11-22
[Spree Upgrade][BackOffice] Fix admin/bulk_product_update spec (1 broken test)


luisramos0:
admin/bulk_product_update has a few errors in #2804, #3018 and #3091.
Apart from these, there is this error in bulk_product_update related to creating a variant directly on the bulk product edit page:

```ruby
  16) 
  As an Administrator
  I want to be able to manage products in bulk
 creating new variants
      Failure/Error: raise message if stock_items.empty?
      
      RuntimeError:
        You need to save the variant to create a stock item before you can set stock levels.
      # ./app/models/concerns/variant_stock.rb:125:in `raise_error_if_no_stock_item_available'
      # ./app/models/concerns/variant_stock.rb:77:in `count_on_hand='
      # ./app/models/concerns/variant_stock.rb:58:in `on_hand='
      # ./app/models/spree/product_set.rb:83:in `update_variant'
      # ./app/models/spree/product_set.rb:71:in `block in update_variants_attributes'
      # ./app/models/spree/product_set.rb:70:in `each'
      # ./app/models/spree/product_set.rb:70:in `update_variants_attributes'
      # ./app/models/spree/product_set.rb:61:in `update_product_variants'
      # ./app/models/spree/product_set.rb:32:in `update_attributes'
      # ./app/models/spree/product_set.rb:95:in `block in save'
      # ./app/models/spree/product_set.rb:94:in `each_value'
      # ./app/models/spree/product_set.rb:94:in `each'
      # ./app/models/spree/product_set.rb:94:in `all?'
      # ./app/models/spree/product_set.rb:94:in `save'
      # ./app/controllers/spree/admin/products_controller_decorator.rb:46:in `bulk_update'
      # ./lib/open_food_network/rack_request_blocker.rb:37:in `call'
      # ------------------
      # --- Caused by: ---
      # Capybara::ExpectationNotMet:
      #   expected to find text "Changes saved." in "Saving failed.Save failed due to invalid data:500"
      #   ./spec/features/admin/bulk_product_update_spec.rb:225:in `block (2 levels) in <top (required)>'
```

luisramos0:
The current error is this:

```
 1)
  As an Administrator
  I want to be able to manage products in bulk
 creating new variants
     Failure/Error: expect(updated_variant.on_hand).to eq 10

       expected: 10
            got: Infinity

       (compared using ==)
[...]

Issue #3091, opened 2018-11-20
[Spree Upgrade][BackOffice] Fix issue in Product edit variants - related to deface override in admin/variants_spec and admin/bulk_product_update (3 broken tests)


luisramos0:
There is something wrong in app/overrides/spree/admin/variants/_form/hide_unit_option_types.html.haml.deface
My best hypothesis is that this override is no longer compatible with the product/variants edit page flow. Needs further investigation.

```
 33) 
    As an admin
    I want to manage product variants
 creating a new variant
      Failure/Error: super
      
      ActionView::Template::Error:
        undefined local variable or method `option' for #<#<Class:0x007fe319974678>:0x007fe32a1df2d0>
      # ./app/helpers/application_helper.rb:20:in `method_missing'
      # ./lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `call'
      # ./lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `respond_with'
      # ./lib/open_food_network/rack_request_blocker.rb:37:in `call'
      # ./spec/features/admin/variants_spec.rb:17:in `block (2 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # NameError:
      #   undefined local variable or method `option' for #<#<Class:0x007fe319974678>:0x007fe32a1df2d0>
      #   ./app/helpers/application_helper.rb:20:in `method_missing'

  34) 
    As an admin
    I want to manage product variants
 editing unit value and description for a variant
      Failure/Error: super
      
      ActionView::Template::Error:
        undefined local variable or method `option' for #<#<Class:0x007fe319974678>:0x007fe31b0840e8>
      # ./app/helpers/application_helper.rb:20:in `method_missing'
      # ./lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `call'
      # ./lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `respond_with'
      # ./lib/open_food_network/rack_request_blocker.rb:37:in `call'
      # ------------------
      # --- Caused by: ---
      # Capybara::ExpectationNotMet:
      #   expected to find visible field "unit_value_human" that is not disabled with value "1" but there were no matches
      #   ./spec/features/admin/variants_spec.rb:46:in `block (2 levels) in <top (required)>'
```

luisramos0:
This also happens in admin/bulk_product_update_spec:

```
 20) 
  As an Administrator
[...]

Issue #3031, opened 2018-11-13
[Spree Upgrade][BackOffice] Fix redirection issue in admin products controller (2 broken tests)


luisramos0:
Fix spec/controllers/spree/admin/products_controller_spec.rb

```
 18) Spree::Admin::ProductsController creating a new product redirects to products when the user hits 'create'
      Failure/Error: response.should redirect_to spree.admin_products_path
        Expected response to be a <:redirect>, but was <200>
      # ./spec/controllers/spree/admin/products_controller_spec.rb:96:in `block (3 levels) in <top (required)>'

  19) Spree::Admin::ProductsController creating a new product redirects to new when the user hits 'add_another'
      Failure/Error: response.should redirect_to "/admin/products/new"
        Expected response to be a <:redirect>, but was <200>
      # ./spec/controllers/spree/admin/products_controller_spec.rb:116:in `block (3 levels) in <top (required)>'
```

This should be a simple fix after #2982 is fixed.

luisramos0:
Looking at this with the branch in #2982, I cant figure this one out...
The redirects are working correctly, but for some reason the 302 response is not showing up on the test. It does show on dev tools _ network when testing manually.
@sauloperez do you want to have a look at this one? Could be a quick one for you.

daniellemoorhead:
Hey @sauloperez now that #3206 is merged can this be closed, or will further work be done on it and therefore it should be back to the In Dev column?

#toottoot

luisramos0:
It's fixed.

PR #3197, opened 2018-12-08
[Spree Upgrade] Adapt Product level stock management to Spree 2

### LINKED ISSUE #2982: [Spree Upgrade] Fix accessing new product
## Description

When clicking on `+ NEW PRODUCT`, a 500 error is returned with the following error:

```
Completed 500 Internal Server Error in 438.5ms (ActiveRecord: 1.8ms)

ActionView::Template::Error (undefined method `backorderable?' for nil:NilClass):
    1: <%# This chunk is just a copy of Spree's core/app/views/spree/admin/products/new.js.erb %>
    2: $("#new_product").html('<%= escape_javascript(render :template => "spree/admin/products/new", :formats => [:html], :handlers => [:erb]) %>');
    3: handle_date_picker_fields();
    4: <% unless Rails.env.test? %>
    5:   $('.select2').select2();
  app/models/concerns/variant_stock.rb:87:in `on_demand'
  app/models/concerns/variant_stock.rb:31:in `on_hand'
  app/models/spree/product_decorator.rb:138:in `on_hand'
  app/views/spree/admin/products/new.js.erb:2:in `_ac8cba3fd61d6e0362ef0d2ce3c49503'


  Rendered /home/pau/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/actionpack-3.2.22.5/lib/action_dispatch/middleware/templates/rescues/_trace.erb (1.1ms)
  Rendered /home/pau/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/actionpack-3.2.22.5/lib/action_dispatch/middleware/templates/rescues/_request_and_response.erb (1.1ms)
  Rendered /home/pau/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/actionpack-3.2.22.5/lib/action_dispatch/middleware/templates/rescues/template_error.erb within rescues/layout (14.7ms)
  Rendered /home/pau/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/actionpack-3.2.22.5/lib/action_dispatch/middleware/templates/rescues/layout.erb (15.8ms)
```

## Expected Behavior

Clicking this button should show the order creation page.

## Actual Behavior

The 500 error decribed above gets raised and nothing happens on the page.

## Possible Fix

According to the comment present on `app/views/spree/admin/products/new.js.erb`

```
 This chunk is just a copy of Spree's core/app/views/spree/admin/products/new.js.erb
```

but it turns out this view got moved in https://github.com/openfoodfoundation/spree/commit/e1e175c1c

luisramos0:
#### What? Why?

Closes #2982, #3018 and #3030

For reviewers:
- see commits individually: commit messages should be self explanatory
- overall comment: it took me quite a lot of time to get to this solution, it's not a beautiful solution, but works...

#### What should we test?
New product page is now working.

Fixes the following specs:
- [X] spec/features/admin/products_spec.rb:23  (#3018)
- [X] spec/features/admin/products_spec.rb:62  (#3018)
- [x] spec/features/admin/products_spec.rb:108  (#3018)
- [X] spec/features/admin/bulk_product_update_spec.rb:174 (#3018)
- [x] spec/features/admin/bulk_product_update_spec.rb:686  (#3018)
- [X] spec/models/spree/product_set_spec.rb:26 (#3030)

Specs and respective issues that are closely related to this PR but will be fixed afterwards:
- spec/features/admin/bulk_product_update_spec.rb:531 (#3091)
- spec/features/admin/bulk_product_update_spec.rb:203 (#3114)
- spec/features/admin/products_spec.rb:201 (#3122)
- spec/features/admin/products_spec.rb:222 (#3122)
- spec/controllers/spree/admin/products_controller_spec.rb:79 (#3031)
- spec/controllers/spree/admin/products_controller_spec.rb:99 (#3031)

#### Dependencies
This PR needs #3190 to make the specs green.


app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -31,6 +31,18 @@ def index
     @show_latest_import = params[:latest_import] || false
   end
 
+  def create
```
mkllnk: I'm not a big fan of Rails' callbacks, but this seems the perfect use case for `around_action`:
```rb
around_action :delete_stock_params_and_set_after, only: :create, :update
```
https://guides.rubyonrails.org/action_controller_overview.html#after-filters-and-around-filters

luisramos0: ehehe, I didnt now that existed, thanks! but tbh I think these callbacks are terrible. It would reduce lines of code but the current solution is very explicit about what's doing. When I am doing things that are not standard (it's the case here) I like to be explicit. around_action would hide it on the top of the file. I prefer as is...

luisramos0:
Rebased so that we see the build results of the freshly merged #3190 together with this one.

luisramos0:
btw, first build I see below 50 broken tests (48).

app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -119,4 +131,22 @@ def strip_new_properties
...
     end
   end
+
+  def delete_stock_params_and_set_after
+    on_demand = params[:product].delete(:on_demand)
+    on_hand = params[:product].delete(:on_hand)
+
+    yield
+
+    set_stock_levels(@product, on_hand, on_demand) if @product.errors.empty?
```
sauloperez: You could use [valid?](https://apidock.com/rails/v3.2.13/ActiveResource/Validations/valid%3F) instead. It does this check for us.

app/models/concerns/product_stock.rb
```diff
@@ -0,0 +1,25 @@
...
+        variants.first.on_demand
+      end
+    else
+      master.on_demand
+    end
+  end
+
+  def on_hand
+    if has_variants?
+      variants.map(&:on_hand).reduce(:+)
```
sauloperez: I think AR's [sum](https://apidock.com/rails/v3.2.13/ActiveRecord/Calculations/sum) would be more performant as we wouldn't load all the variant objects, iterate to get their on_hand value and then sum them all. Something like:

```suggestion
      variants.sum(:on_hand)
```

app/models/concerns/product_stock.rb
```diff
@@ -0,0 +1,25 @@
+require 'active_support/concern'
+
+module ProductStock
+  extend ActiveSupport::Concern
+
+  def on_demand
+    if has_variants?
+      if variants.size > 1
+        raise 'Cannot determine product on_demand value of product with multiple variants'
```
sauloperez: what about implementing this as an early return? I think this better conveys the fact that it's an invalid case and it'd reduce a nesting level.

app/models/spree/product_set.rb
```diff
@@ -63,27 +63,37 @@ def update_product_variants(product, attributes)
...
   end
 
+  def create_variant(product, variant_attributes)
+      on_hand = variant_attributes.delete(:on_hand)
+      on_demand = variant_attributes.delete(:on_demand)
+
+      variant = product.variants.create(variant_attributes)
+
+      variant.on_demand = on_demand if on_demand.present?
+      variant.on_hand = on_hand.to_i if on_hand.present?
```
sauloperez: I don't really understand why we assign them this way and can't be done through `#create`. Can you remind me why we need it?

spec/models/concerns/product_stock_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+describe ProductStock do
+  let(:product) { create(:simple_product) }
+
+  context 'when product has no variants' do
+    before do
+      product.variants.first.destroy
+      product.variants.reload
+    end
+
+    scenario 'product.on_demand is master.on_demand' do
```
sauloperez: why don't we use `it` like all other unit tests? I'd rather keep the consistency to avoid confusing new devs.

app/models/concerns/product_stock.rb
luisramos0: I will do that.
this is your code btw ;-)
https://github.com/openfoodfoundation/openfoodnetwork/commit/75ad4254a6c50b28a2f3b36828d2e3428b20b774#diff-e6a3a7e1c82ac34e5e72899c9ceb109cR133

app/models/concerns/product_stock.rb
luisramos0: do you mean:
```ruby
 if has_variants?
   raise 'Cannot determine product on_demand value of product with multiple variants' if variants.size > 1
   variants.first.on_demand
else
   master.on_demand
end
```

app/models/spree/product_set.rb
luisramos0: it's because these assignments will save the variant themselves and that messes up the overall variant save process. I think this is ok, first you have to create the variant and then you can manage stock (there are stock items and stock movements involved in stock management).

app/models/concerns/product_stock.rb
sauloperez: Yes, I saw. I was probably too lazy

app/models/spree/product_set.rb
sauloperez: uf, this is getting messy... we'll need to fix these things once get the upgrade live.

luisramos0: I dont think it's too bad. you cant expect to create a variant and all it's parts in one mass_assignemnt. we should have a collection of services to do different things to models. I think it's totally fine to detach variant stock management from the variant model. the only thing we need to fix is to take these attributes out of the variant into a service. moving from variant_stock into some form of variant_stock_service.

so, in this case the actual change would be on the frontend where the "API" to create variants with stock settings disappears and the frontend needs to first create the variant and then set/manage stock levels (it would be a different API endpoint to manage stock levels).

sauloperez: definitely, we'll need to work on the frontend and that will simplify the backend.

app/models/concerns/product_stock.rb
mkllnk: I don't think ActiveRecord's `sum` works here, because `on_hand` is not a column in the database, right? Each variant actually needs to find all stock locations first. And the variants may be augmented by variant overrides.

luisramos0: hey @mkllnk , thanks, that makes sense!

luisramos0: I'll skip this. ok?

luisramos0:
ready for re-review @sauloperez @mkllnk 
thanks!

app/models/concerns/product_stock.rb
sauloperez: @mkllnk my point here was to get the sum straight from DB; way faster than Ruby.

mkllnk: Sure. :-)

Issue #3200, opened 2018-12-11
Missing "required field" stars in New Product form


haseleyi:
## Description

When we create a new product in the back office, leaving the "Unit Size" field blank results in an error message. We're missing a red star to indicate a required field there. Similarly, the "Value" field is missing a star.

![capture](https://user-images.githubusercontent.com/18274798/49770176-13126000-fc98-11e8-9666-236fac2ecf3d.PNG)

## Steps to Reproduce

1. Login and visit Administration.
2. Visit Products and create a new product.
3. Fill out all fields except for Unit Size and click Create to see the first error.
4. Fill out Unit Size but not Value to see the second error.

## Severity

s5? Few users will neglect the field, but when they do, an unexpected alert can be annoying.

## Your Environment

* Version used: 1.23.1 Mussel Mass
* Browser name and version: Firefox 63.0.3
* Operating System and version (desktop or mobile): Ubuntu 16.04

PR #3201, opened 2018-12-11
Add missing 'required field' stars in new product form

haseleyi:
#### What? Why?

Closes [#3200](https://github.com/openfoodfoundation/openfoodnetwork/issues/3200)

In the New Product form, we are missing some stars to denote required fields. This should fix that!

#### What should we test?

1. Login and visit Administration.
2. Visit Products and create a new product.
3. Check for "required field" stars on the Unit Size and Value fields.

#### Release notes

Added missing "required field" stars in New Product form.

Changelog Category: Fixed

### DIFF
--- a/app/views/spree/admin/products/new.html.haml
+++ b/app/views/spree/admin/products/new.html.haml
@@ -22,6 +22,7 @@
         .six.columns.alpha
           = f.field_container :units do
             = f.label :variant_unit_with_scale, t(:units)
+            %span.required *
             %select.select2.fullwidth{ id: 'product_variant_unit_with_scale', 'ng-model' => 'product.variant_unit_with_scale', 'ng-options' => 'unit[1] as unit[0] for unit in variant_unit_options' }
               %option{'value' => '', 'ng-hide' => "hasUnit(product)"}
             %input{ type: 'hidden', 'ng-value' => 'product.variant_unit', name: 'product[variant_unit]' }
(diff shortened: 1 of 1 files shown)


sigmundpetersen:
@haseleyi if you have Zenhub installed you can Connect the PR to the issue it closes. It's at the bottom. This makes it easier to track the PR/Issue set on the Zenhub board. 🙏

haseleyi:
@sigmundpetersen thanks for the tip! At the bottom of our board, I get "This is a read-only Zenhub Board. To edit this Board, you must have write permissions for this repo."

sigmundpetersen:
Yeah as I suspected @haseleyi , permissions. Thanks for checking.

RachL:
Meeting is 10min late so one more for the road 🚀 on Aus staging

myriamboure:
It becomes enables when you select a unit size @RachL I think it's acceptable even if not perfect ;-)

myriamboure:
It might be because when you select item the fields change, there is a new field new value and display as which is "unit name", so it's not unconsistent to say "choose first the unit type" and then we'll tell you what to fill in ? Anyway, I understand your point about and yes it wouldn't harm I guess to make the value field editable even before choosing the unit, but I don't know if that deserves an issue if we never had any feedback from users on it... but as you want, if there is an issue it would be s5 IMO.

RachL:
@myriamboure yes it is just that now with this PR we have a little red star next to the field. So it is a bit weird to see that you need to fill the field, but that actually you can't because the field is disable. Do you see what I mean?

myriamboure:
Yes I see @RachL :-) Got your point, I guess UX would be better yes if possible to modify even before selecting unit type.

PR #3203, opened 2018-12-11
[Spree Upgrade] Fix product properties and images deletion

### LINKED ISSUE #3122: [Spree Upgrade][BackOffice] Fix features/admin/products_spec (3 broken tests)

luisramos0:
#### What? Why?

Closes #3122 

This fixes two specs:
- spec/features/admin/products_spec.rb:202 scenario "deleting product properties"
- spec/features/admin/products_spec.rb:223 scenario "deleting product images"

#### What should we test?
These specs should be green.

### DIFF
--- a/app/controllers/spree/admin/images_controller_decorator.rb
+++ b/app/controllers/spree/admin/images_controller_decorator.rb
@@ -1,0 +1,6 @@
+Spree::Admin::ImagesController.class_eval do
+  # This will make resource controller redirect correctly after deleting product images.
+  # This can be removed after upgrading to Spree 2.1.
+  # See here https://github.com/spree/spree/commit/334a011d2b8e16355e4ae77ae07cd93f7cbc8fd1
+  belongs_to 'spree/product', :find_by => :permalink
+end
--- a/spec/features/admin/products_spec.rb
+++ b/spec/features/admin/products_spec.rb
@@ -210,7 +210,7 @@ feature %q{
       page.should have_field 'product_product_properties_attributes_0_value', with: 'fooval'
 
       # And I delete the property
-      page.all('a.remove_fields').first.click
+      page.all('a.delete-resource').first.click
       click_button 'Update'
 
       # Then the property should have been deleted


app/controllers/spree/admin/images_controller_decorator.rb
```diff
@@ -0,0 +1,6 @@
+Spree::Admin::ImagesController.class_eval do
+  # This will make resource controller redirect correctly after deleting product images.
+  # This can be removed after upgrading to Spree 2.1.
+  # See here https://github.com/spree/spree/commit/334a011d2b8e16355e4ae77ae07cd93f7cbc8fd1
+  belongs_to 'spree/product', :find_by => :permalink
```
sauloperez: sorry, I don't understand what this line has to do with that commit. I don't even know what a `belongs_to` does in a controller. What did you find out?

luisramos0: it's how resource controller works. this is how the collection_url is build:
https://github.com/openfoodfoundation/spree/blob/2-0-stable/backend/app/controllers/spree/admin/resource_controller.rb#L242

belongs_to is used to define the parent of the resource (in terms of routes) and then collection_url will use the parent to build the path, in this case, instead of resulting in admin_images_path (it doesnt exist) it builds the path correctly by using parent admin_product_images_path. this is how I fix it here.
The fix in spree is different, somewhere in the future they create a resource_controller. location_after_destroy that image_controller can override. I think myy fix is easier, and, anyway, can be deleted when upgrading to 2.1.

PR #3206, opened 2018-12-12
[Spree Upgrade] Fix redirection on admin products specs

### LINKED ISSUE #3031: [Spree Upgrade][BackOffice] Fix redirection issue in admin products controller (2 broken tests)

sauloperez:
#### What? Why?

Closes #3031 

In Spree v2.0 the product attributes being sent were no longer valid. Providing a shipping category and setting a stock location on the DB so that a stock item can be created fixes them.

Remember `Variant#create_stock_item`, which is defined as an `after_create` callback, relies on `StockLocation` to create the item.

#### What should we test?

This two tests should get green.

### DIFF
--- a/app/models/spree/product_decorator.rb
+++ b/app/models/spree/product_decorator.rb
@@ -275,19 +275,23 @@ Spree::Product.class_eval do
   # This fixes any problems arising from failing master saves, without the need for a validates_associated on
   # master, while giving us more specific errors as to why saving failed
   def save_master
-    begin
-      if master && (master.changed? || master.new_record? || (master.default_price && (master.default_price.changed? || master.default_price.new_record?)))
-        master.save!
-      end
+    if master && (
+        master.changed? || master.new_record? || (
+          master.default_price && (
+            master.default_price.changed? || master.default_price.new_record?
+          )
+        )
+      )
+      master.save!
+    end
 
     # If the master cannot be saved, the Product object will get its errors
     # and will be destroyed
-    rescue ActiveRecord::RecordInvalid
-      master.errors.each do |att, error|
-        self.errors.add(att, error)
-      end
-      raise
+  rescue ActiveRecord::RecordInvalid
+    master.errors.each do |att, error|
+      self.errors.add(att, error)
     end
+    raise
   end
 
   def sanitize_permalink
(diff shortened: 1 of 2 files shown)


spec/controllers/spree/admin/products_controller_spec.rb
```diff
@@ -74,45 +74,30 @@
   end
 
   context "creating a new product" do
-    before { login_as_admin }
+    let(:supplier) { create(:supplier_enterprise) }
+    let(:taxon) { create(:taxon) }
+    let(:shipping_category) { create(:shipping_category) }
+
+    let(:product_attrs) {
+      attributes_for(:product).merge(
```
sauloperez: `attributes_for` is one of FactoryBot's creation methods.

app/models/spree/product_decorator.rb
```diff
@@ -275,19 +275,23 @@ def ensure_standard_variant
...
-    rescue ActiveRecord::RecordInvalid
-      master.errors.each do |att, error|
-        self.errors.add(att, error)
-      end
-      raise
+  rescue ActiveRecord::RecordInvalid
+    master.errors.each do |att, error|
+      self.errors.add(att, error)
     end
+    raise
```
luisramos0: welcome to spree v3 where this is done: https://github.com/spree/spree/commit/fe6369f9d289d7d69b0970b576d77f41a1eb7061

spec/controllers/spree/admin/products_controller_spec.rb
luisramos0: nice!!! you can use a factory to generate a hash!

sauloperez: Yep, perfect fit for creation controller specs IMO.

PR #3209, opened 2018-12-13
2072 New sample data for testing

### LINKED ISSUE #2072: Provide new seed data in load_sample_data
## Description

The current [dev:load_sample_data](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/tasks/dev.rake#L6) adds quite a lot of records but not very useful when testing. AFAIK the `Enterprise 2` is the one that allows you placing an order right away, without having to add any payment or delivery method manually first.

We need enough sample data so that a tester can start testing the established test cases with the app after running the `dev:load_sample_data` rake task.

## Expected Behavior

After a branch is deployed in the desired environment, a developer can log into the machine and execute `db:reset` to erase the existing data in the DB, populate it with the seed data Spree provides, and then, execute `dev:load_sample_data`.

This provides all the data specified in: https://docs.google.com/document/d/1_va0Aw2yJ5BRe-HtuBT-OC8HNHIui0gvQ0uElWjDVs0/edit?usp=sharing

## Actual Behavior
The aforementioned rake task now populates the DB with the data that is not relevant for testing.

## Context

https://community.openfoodnetwork.org/t/seed-data-development-provisioning-deployment/910

## Severity

S-3. It's impacting the testing process a great deal and thus affecting our velocity.

## Your Environment

* Version used:
* Browser name and version:

## Possible Fix

Rewriting https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/tasks/dev.rake#L6

mkllnk:
#### What? Why?

Closes #2072

We need more and better structured sample data for dev testing and to bootstrap staging data. This pull request introduces a new rake task that aims to replace openfoodnetwork:dev:sample_data.

The generated data is mainly the data defined in https://docs.google.com/document/d/1_va0Aw2yJ5BRe-HtuBT-OC8HNHIui0gvQ0uElWjDVs0. There are the following differences:

* There is no difference between a producer profile and a shop profile. We just have enterprise profiles. Therefore I introduced the user Penny with her profile enterprise. The documented enterprise 1, Fred's farm, is a normal producer. And the documented enterprise 4, Mary's shop, is a normal shop.
* VAT has not been considered yet. I can do that in another iteration.
* The second sub-distributor and its order cycle are missing. I didn't see a need for it. That may come with VAT.
* Some of the naming is slightly different as it made more sense to me.
* I filled in some missing data like pickup instructions.

I tried to keep this as short as possible, but I spent 22 hours on this already. I made some effort though to make it easier to maintain and extend. It has much better code metrics than the old code.

#### What should we test?

This cannot be staged without a developers help. The developer needs to run the following to make this data active:

```sh
sudo /bin/systemctl stop delayed_job_openfoodnetwork.service
sudo /bin/systemctl stop unicorn_openfoodnetwork.service
mkdir -p ~/backup/
pg_dump -h localhost -U ofn_user openfoodnetwork | gzip > ~/backup/staging-`date +%Y%m%d%H%M%S`-before-reset.sql.gz
bundle exec rake db:reset ofn:sample_data
sudo /bin/systemctl start delayed_job_openfoodnetwork.service
sudo /bin/systemctl start unicorn_openfoodnetwork.service
```

Testers can then verify that the data is set up as specified. I think, @myriamboure worked on the data with certain test scenarios in mind. Would you like to share these so that we can verify that they work with this data?

#### Release notes

Changed: Added and improved sample data for testing and demo instances.

Changelog Category: Changed

Some of the data generation may have to be updated for the Spree upgrade.

#### Discourse thread

https://community.openfoodnetwork.org/t/seed-data-development-provisioning-deployment/910


mkllnk:
> I wonder if we should use the factories in spec/factories

I wondered the same. There is some potential code duplication here. But I didn't use those factories for two reasons:

1. They generate more data that is not explicitly specified by our task.
2. They generate some random data which is not deterministic and doesn't fit the purpose here.

While generating sample data and spec data are similar tasks, I don't like to tie the two together. I like to specify everything explicitly and design the factories to be used as easily as possible for our sample data. The old task used the spec factories, but I don't think it made it better.

One lesson from reading these factories is that our data structures are badly complex. Working on the models to make the factories easier would also reduce the potential of bugs a lot.

myriamboure:
Thanks @mkllnk ! I need to dive again into it which is not easy ;-) 

> There is no difference between a producer profile and a shop profile. We just have enterprise profiles. Therefore I introduced the user Penny with her profile enterprise. The documented enterprise 1, Fred's farm, is a normal producer. And the documented enterprise 4, Mary's shop, is a normal shop.

Then you created duplicates, as I created those two profiles to make sure we distinguish how they look like on the map if needed (4 is in white color for instance as distributor and not using the shopfront). Actually there are differences between a producer and a shop profile. 
If you wanted to only have one profile only entity (maybe that's enough...) then I would simply have deleted number 4 and transform number 1 into a "enterprise profile" that is described as a producer (so appear with the tractor and in green.

But anyway, it might not harm...

> VAT has not been considered yet. I can do that in another iteration.

I don't think you will be able to checkout without a VAT zone setup in configuration... don't you think ? If it's not a problem then no problem we can iterate. But we definitely do want to test that VAT is properly calculated, and VAT reports display correctly, etc. So we would need it to avoid to recreate VAT categories and rates in config each time we want to test some VAT potential impact of a PR.

> The second sub-distributor and its order cycle are missing. I didn't see a need for it. That may come with VAT.

The second hub distributor was setup to test inventories. As it's OC is sourcing info from inventory and not from product catalog, and this hub (Maryna hub) is the only one to use the inventory feature.
Also I used it to test a specific case : I created 2 OC on this hub to have a case setup with multiple OC at the same time.

Anyway, I'll write some test scenarios at the end of the document, will do it now ! Cheers :-)

myriamboure:
Oh sorry about the second sub-distributor maybe you meant "Maryse" ? I did create her because she runs a private shop so we can have a private shop case to do tests on it. Anyway, I'll clarify all that with the scenario.

lib/tasks/sample_data.rake
```diff
@@ -0,0 +1,656 @@
...
+        Spree::State.count > 0
+    end
+
+    module Logging
+      private
+
+      def log(message)
+        puts "[openfoodnetwork:sample_data:load] #{message}"
+      end
+    end
```
sauloperez: we could use the rails logger as any part of Rails. [TaggedLogging](https://apidock.com/rails/v3.2.13/ActiveSupport/TaggedLogging) is a perfect fit for this.

lib/tasks/sample_data.rake
```diff
@@ -0,0 +1,656 @@
...
+      def zone
+        zone = Spree::Zone.find_or_create_by_name!("Australia")
+        zone.members.create(zonable: country)
+        zone
+      end
+
+      def country
+        Spree::Country.find_by_iso(ENV.fetch('DEFAULT_COUNTRY_CODE'))
+      end
+    end
```
sauloperez: I don't see any reason why not to have this modules and classes in their own files and require them from the rake task. This is just Ruby code after all.

myriamboure:
Ok so I wrote at the end of the document the cases I had in mind when writing this document (it has been reviewed by Sally at that time...) https://docs.google.com/document/d/1_va0Aw2yJ5BRe-HtuBT-OC8HNHIui0gvQ0uElWjDVs0/edit so I guess in order to test we should somehow test that the different cases are setup as expected ? Please check @mkllnk that it's what you understood, as with the changes you made I think some of those test cases are not setup then. But anyway, we can test and see if that's ok as you did it to start with, and iterate.

lib/tasks/sample_data.rake
mkllnk: Ah, yes, that looks quite apt. But I don't see the benefit in this simple scenario. The change would be this:
```diff
-  puts "[openfoodnetwork:sample_data:load] #{message}"
+  @logger ||= ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
+  @logger.tagged("openfoodnetwork:sample_data:load") { @logger.info(message) }
```
That's much more complicated. What is the added value?

sauloperez: It's more about using a logger rather than puts. We have logging built-in in Rails that we could leverage. You can then add those tags using the formatter attribute of the Rails logger for instance.

https://logmatic.io/blog/rails-logger-best-practices/ seems to be a good post about the topic.

lib/tasks/sample_data.rake
sauloperez: yep, sounds good to me.

Matt-Yorkley:
This suggestion isn't strictly necessary, but: if we're creating enterprises with different configurations and relationships, can we describe it in each enterprise's profile description so that any dev or tester can click on them under /shops and see an explanation, like "This shop sells it's own produce, and supplies Enterprise X"?

sauloperez:
That is nice to have

luisramos0:
@sauloperez can we move forward with this PR? the namespace changes will be done somewhere else right?

myriamboure:
@mkllnk I guess we can start with what you have done, when testing I suggest we carefully document then / adapt the use cases I describe to the reality of what you have done. If we realize there are important cases missing we will comment when testing. Let's move forward.

lib/tasks/sample_data.rake
kristinalim: I agree with @sauloperez. The Rake task is too long and difficult to review, and can be split into multiple files in `lib/tasks/sample_data`.

lib/tasks/sample_data.rake
```diff
@@ -0,0 +1,656 @@
...
+          city: parts[1],
+          zipcode: parts[2],
+          state: state,
+          country: country
+        )
+      end
+
+      def zone
+        zone = Spree::Zone.find_or_create_by_name!("Australia")
+        zone.members.create(zonable: country)
```
kristinalim: This should be a bang method.

lib/tasks/sample_data.rake
```diff
@@ -0,0 +1,656 @@
...
+          variant_unit_scale: 1,
+          unit_value: 1,
+          on_demand: true
+        )
+        create_product_with_distribution(params)
+      end
+
+      def create_product_with_distribution(params)
+        product = Spree::Product.create_with(params).find_or_create_by_name(params[:name])
+        ProductDistribution.create(
```
kristinalim: Same here and the line above - these should be bang methods.

lib/tasks/sample_data.rake
```diff
@@ -0,0 +1,656 @@
...
+        create_override(shop, product)
+      end
+
+      def create_override(shop, product)
+        VariantOverride.create_with(
+          variant: product.variants.first,
+          hub: shop,
+          price: 12,
+          count_on_hand: 5
+        ).find_or_create_by_variant_id(product.variants.first.id)
```
kristinalim: This should be a bang method.

lib/tasks/sample_data.rake
```diff
@@ -0,0 +1,656 @@
...
+      end
+
+      private
+
+      def create_item(shop, product)
+        InventoryItem.create_with(
+          enterprise: shop,
+          variant: product.variants.first,
+          visible: true
+        ).find_or_create_by_variant_id(product.variants.first.id)
```
kristinalim: This should be a bang method.

lib/tasks/sample_data.rake
mkllnk: I changed it now as you suggested, but I don't see any added value in this case.

lib/tasks/sample_data.rake
mkllnk: I agree. I split it all up now.

lib/tasks/sample_data.rake
mkllnk: Thank you Kristina! Changing this revealed a validation error on product distributions. :-) It's fixed now.

sigmundpetersen:
It was a random Semaphore failure. It passed on a rebuild. But yes, fixing the inconsistent spec fails should be prioritized soon.

mkllnk:
Ah, good one. I never use the setup script. But yes, we should update that as well.

lib/tasks/dev.rake
```diff
@@ -316,6 +316,15 @@ namespace :openfoodnetwork do
...
       end
+      puts <<WARNING
+
+  Warning: This task is going to be replaced by:
+
+    $ bundle exec rake openfoodnetwork:sample_data:load
+
+  It contains more sample data.
+
+WARNING
```
sauloperez: nothing to do in this PR but Rails has tool for this. See https://apidock.com/rails/v3.2.13/ActiveSupport/Deprecation/warn/class

lib/tasks/sample_data/enterprise_factory.rb
```diff
@@ -0,0 +1,93 @@
...
+  # rubocop:disable Metrics/MethodLength
+  def enterprise_data(users)
+    [
+      {
+        name: "Penny's Profile",
+        owner: users["Penny Profile"],
+        is_primary_producer: false,
+        sells: "none",
+        address: address("25 Myrtle Street, Bayswater, 3153"),
+        long_description: <<DESC
```
sauloperez: [strip_heredoc](https://apidock.com/rails/v3.2.13/String/strip_heredoc) is perfect for this.

lib/tasks/sample_data/order_cycle_factory.rb
```diff
@@ -0,0 +1,111 @@
...
+      ["Mary's Online Shop", "Maryse's Private Shop"],
+      receival_instructions: "Please shut the gate.",
+      pickup_time: "dusk"
+    )
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  private
+
+  def create_order_cycle(name, coordinator_name, supplier_names, distributor_names, data)
```
sauloperez: keyword arguments would be perfect here to make calls to it a bit easier to understand. 5 arguments is already a lot.

lib/tasks/sample_data/payment_method_factory.rb
```diff
@@ -0,0 +1,52 @@
...
+  def create_cash_method(enterprise)
+    create_payment_method(
+      enterprise,
+      "Cash on collection",
+      "Pay on collection!",
+      Spree::Calculator::FlatRate.new
+    )
+  end
+
+  def create_card_method(enterprise)
```
sauloperez: At this point, it's clear to me that `enterprise` might be worth being an ivar. It's class state.

lib/tasks/sample_data/payment_method_factory.rb
```diff
@@ -0,0 +1,52 @@
...
+    create_payment_method(
+      enterprise,
+      "Credit card (fake)",
+      "We charge 1%, but won't ask for you details. ;-)",
+      Spree::Calculator::FlatPercentItemTotal.new(preferred_flat_percent: 1)
+    )
+  end
+
+  def create_payment_method(enterprise, name, description, calculator)
```
sauloperez: keyword arguments would make this also a bit more readable.

mkllnk:
Thank you for the suggestions @sauloperez. I learned a lot. I incorporated more of the feedback and since it's all approved now, I rebased it as well.

- The task is now simply `openfoodnetwork:sample_data` without `load` and another pr will change it to `ofn:sample_data`.
- I used the ActiveSupport deprecation feature in the old task.
- I used `strip_heredoc` for enterprise descriptions.
- I updated `bin/setup`.

I didn't go for named parameters, because that would have been a lot of redundancy. But maybe we will do it in another iteration.

lib/tasks/sample_data/enterprise_factory.rb
sauloperez: don't worry, it was just a suggestion to make you aware of that method if you weren't

sauloperez:
Shall we merge? did you @myriamboure take a look at it? let's iterate! :muscle:

sauloperez:
I just deployed to AU's staging and executed those commands but unfortunately, it failed. See the output in [output.log](https://github.com/openfoodfoundation/openfoodnetwork/files/2764068/output.log)

mkllnk:
I finally got around to testing this. It works in development environment, but fails in staging. Investigating...

mkllnk:
Okay, I found a workaround. Back to test-ready.

luisramos0:
this PR needs a rebase @mkllnk so that it gets the bundler script from #3302

RachL:
Then I'm moving it back to in dev

mkllnk:
Okay, I rebased it. Now it's back in test-ready.

luisramos0:
I tried to run the script but bundle is not there:
```
openfoodnetwork@staging:~/apps/openfoodnetwork/current$ bundle exec rake db:reset openfoodnetwork:sample_data
bash: bundle: orden no encontrada
```

I tried using ~/.rbenv/shims/bundle but then I see an error related to MAIL_HOST not being defined.

How do I fix this?

mkllnk:
I just logged in:
```
ssh staging.katuma.org
sudo su - openfoodnetwork
cd ~/apps/openfoodnetwork/current/
bundle # works
bundle exec rake db:reset openfoodnetwork:sample_data # failed because the task got renamed
bundle exec rake ofn:sample_data # works
sudo /bin/systemctl start delayed_job_openfoodnetwork.service
sudo /bin/systemctl start unicorn_openfoodnetwork.service
```

Either someone fixed it in the mean time or your environment wasn't right. I have an idea. Did you use `sudo su openfoodnetwork` instead of `sudo su - openfoodnetwork`? That dash is important to load the new user's environment. Otherwise it keeps the environment of ofn-admin.

mkllnk:
I just tested myself a little bit and found one issue. All payment methods are created with the development environment. So they are not active on staging which renders all order cycles invalid. I'll fix that and put it then back to test ready.

luisramos0:
thanks @mkllnk it was just a "bug" in my local ssh connection script, I was doing this:
`ssh -t openfoodnetwork@staging.katuma.org 'cd apps/openfoodnetwork/current; bash'`

I changed it to this and it's now working, thanks!
`ssh -t ofn-admin@staging.katuma.org 'sudo su - openfoodnetwork; cd apps/openfoodnetwork/current;'`

sauloperez:
Might be simpler to do `ssh openfoodnetwork@staging.katuma.org; cd apps/openfoodnetwork/current` @luisramos0

luisramos0:
ping @mkllnk did your last commit fix the problem? I wonder if this is just forgotten and should be in test ready

mkllnk:
I don't know yet. I was waiting for Aus staging to become available so that I can test it. :-)

mkllnk:

I had a quick look and shops are now appearing. :+1:

luisramos0:
awesome, thanks!

RachL:
Hey @mkllnk this is looking good! Some feedback and comments:

- I have an error 500 when I'm trying to checkout on all the open shops, but I didn't find why :/ I have added a screenshot at the end of @myriamboure 's doc https://docs.google.com/document/d/1_va0Aw2yJ5BRe-HtuBT-OC8HNHIui0gvQ0uElWjDVs0/edit
- Emails attached to users are not those described by Myriam, but I will propose on Discourse that we use something like [Mailtrapp ](https://mailtrap.io/how-it-works) on our staging servers. It will be more easy, especially if the team of tester is growing, isn't it?
But an important question: what are the password for all those user accounts? I'm deeply sorry if it's mentioned somewhere and I did not find it.
- I still can log in as spree@example.com. Is this normal?
- All products mentioned in Myriam's doc are not there, but I think there is enough to play with, so I'm good with starting with that :)

mkllnk:
> I have an error 500 when I'm trying to checkout on all the open shops

That's a bug. I have to find out why.

> Mailtrap

Great! I like the idea. Pau suggested another service but it would make all emails public. I find that too risky.

> what are the password for all those user accounts?

The passwords are random, unknown. You can set them through the admin panel. I'm not sure how to manage them because I don't want the password to all accounts public. It doesn't matter at the moment as the password for the admin user is public but I hope we will change that.

> All products mentioned in Myriam's doc are not there

Uh, you are right. I probably thought that could be in a second iteration.

RachL:
@mkllnk yes don't worry about the products it's fine for now :) Let's just focus on solving the bug.

About passwords, we can store them in the new password manager: https://community.openfoodnetwork.org/t/say-hello-to-bitwarden-our-password-manager/1583/2

luisramos0:
It's the payment method type/provider that is missing. I edited payment methods and saved with a provider and it worked.

sigmundpetersen:
I think this PR would benefit from a rebase to get some recent flaky spec fixes

luisramos0:
yes, @sigmundpetersen rebase or retry :-)


@mkllnk I see 4 pending points here:
1.  I have added mandatory payment method provider, it's looking good in staging AUS and it's working (see commit)
2. Then I bumped into the shipping methods being on a different zone, so I added a commit to make the zone the same as the environment checkout_zone, it looks good in staging AUS and it is working (see other commit).

3. And now, one more problem: for some reason Australia, the default country, is not part of Australia, the default zone... I will/check fix this one.

4. And on top of this (if I fix the data manually in AUS staging), there's one more problem, on checkout, I see this exception (which I dont think is related to any of the above):
```
Completed 500 Internal Server Error in 1624.9ms (ActiveRecord: 544.7ms)
RuntimeError (can't modify frozen Hash):
  app/models/spree/order_decorator.rb:418:in `update_adjustment!'
  app/models/spree/order_decorator.rb:174:in `block in update_shipping_fees!'
  app/models/spree/order_decorator.rb:172:in `update_shipping_fees!'
  app/controllers/checkout_controller.rb:134:in `advance_order_state'
  app/controllers/checkout_controller.rb:33:in `update'
```

mkllnk:
I noticed on Tuesday that the bugsnag initializer was missing. Maybe the provisioning needs be run to install it. I'm looking forward to putting these files into the repository and getting them out of ofn-install.

luisramos0:
all right, I found the solution for point 3, it was a typo in zone: zonable vs zonEable :-)
new commit to fix that.

I'll re-stage and see if point 4 is still there.

luisramos0:
It works in AUS! So, all issues above fixed.

In code review, I requested re-review from @sauloperez and @kristinalim (there 3 new commits).

I leave it in staging AUS https://staging.openfoodnetwork.org.au so you can all have a look at the new beautiful sample data!

mkllnk:
I just ran provisioning to fix the Bugsnag issue.

lib/tasks/sample_data/payment_method_factory.rb
```diff
@@ -33,21 +34,22 @@ def create_cash_method(enterprise)
...
       "Credit card (fake)",
       "We charge 1%, but won't ask for your details. ;-)",
       Spree::Calculator::FlatPercentItemTotal.new(preferred_flat_percent: 1)
     )
   end
 
-  def create_payment_method(enterprise, name, description, calculator)
-    card = enterprise.payment_methods.new(
+  def create_payment_method(provider_class, enterprise, name, description, calculator)
+    payment_method = provider_class.constantize.new(
```
sauloperez: What stops us from avoiding the call `constantize` call and just passing constants instead of strings?

luisramos0: just my ignorance, I didnt know I could just pass the class directly. done in a new commit.

luisramos0:
Made a very minor change to include @sauloperez comment.
Moving to test ready, @RachL back to you :-)

luisramos0:
I moved this to the top of test ready column... this PR is older than my son, 3 months :-D

RachL:
@luisramos0 yes but so far the rule was to take them in order. Should we really take into account the first date of the PR?

luisramos0:
I think we should use the same rule as in code review, if it's a re-review (in this case, a re-test) it should be moved to the top.
even if we dont do that, I do think we need to nail this PR down asap :-D

sauloperez:
I have the same feeling. This PR is getting too old.

luisramos0:
lets just merge it, please 🙏  🤣

luisramos0:
all right, staged in katuma and looking good:
http://staging.katuma.org

RachL:
Checkout now works fine! And I see that bogus was added as credit card possibilities. Awesome 👍 

I guess we can move forward and iterate now.

About iteration, what's the plan: discuss on discourse the type of data / config we want to add (example : stripe) and then but directly the issue here? Is this considered as tech debt?

Also just to be 100% sure, is super admin still spree@example.com @mkllnk ?

RachL:
oh and can we not put them on staging UK while we work on making this dataset bigger? @luisramos0 @mkllnk @sauloperez @Matt-Yorkley 

As we have several staging server and I'm currently the only using them, I guess we can dedicate UK for test that need large dataset? What do you think?

sauloperez:
what do you mean exactly? to avoid touching UK's staging for now?

sauloperez:
> About iteration, what's the plan: discuss on discourse the type of data / config we want to add (example : stripe) and then but directly the issue here? Is this considered as tech debt?

It should be treated like any other feature IMO.

RachL:
@sauloperez 

> what do you mean exactly? to avoid touching UK's staging for now?
yep

> It should be treated like any other feature IMO.
Meaning going through the voting process?

sauloperez:
> Meaning going through the voting process?

I meant in terms of prioritization and follow-up. To my understanding the feature it's already been started its implementation. It will take a bunch of small PRs at most, however.

mkllnk:
@RachL Yes, spree@example.com is still the default admin login. Merging... :tada:

PR #3210, opened 2018-12-13
[Spree Upgrade] Fix availability validator to include inventory_units in it's validation

### LINKED ISSUE #3037: [Spree Upgrade] Fix order_factory spec (2 broken tests)
Fix spec/services/order_factory_spec

```
72) OrderFactory create builds a new order based the provided attributes
      Failure/Error: expect(order.shipping_method).to eq shipping_method
      
        expected: #<Spree::ShippingMethod id: 375, name: "UPS Ground", created_at: "2018-11-12 17:14:29", updated_at: "...", display_on: "", deleted_at: nil, require_ship_address: true, description: nil, tracking_url: nil>
             got: nil
      
        (compared using ==)
      # ./spec/services/order_factory_spec.rb:38:in `block (3 levels) in <top (required)>'

 40) OrderFactory create when requested quantity is greater than available stock when no override is present when skip_stock_check is requested initialises the order with the requested quantity regardless
      Failure/Error: expect(order.line_items.find_by_variant_id(variant1.id).quantity).to eq 5
      
      NoMethodError:
        undefined method `quantity' for nil:NilClass
      # ./spec/services/order_factory_spec.rb:76:in `block (6 levels) in <top (required)>'

  41) OrderFactory create when requested quantity is greater than available stock when an override is present when skip_stock_check is requested initialises the order with the requested quantity regardless
      Failure/Error: expect(order.line_items.find_by_variant_id(variant1.id).quantity).to eq 6
      
      NoMethodError:
        undefined method `quantity' for nil:NilClass
      # ./spec/services/order_factory_spec.rb:99:in `block (6 levels) in <top (required)>'
```

PR #3038 fixes the first issue.

The second problem "nil:NilClass" is related to line_item.skip_stock_check. To be investigated.

### LINKED ISSUE #3090: [Spree Upgrade][BackOffice] Fix features/admin/bulk_order_management_spec - issue with variant on_hand levels (2 broken tests)
Fix spec/features/admin/bulk_order_management_spec (2 broken tests)

There are problems related to variant on_hand levels in  spec/features/admin/bulk_order_management_spec:
```
  1) 
  As an Administrator
 altering line item properties submitting data to the server when acceptable data is sent to the server displays an update button which submits pending changes
     Failure/Error: expect(page).to have_selector "#save-bar", text: "All changes saved"
       expected to find visible css "#save-bar" with text "All changes saved" but there were no matches. Also found "Fields with red borders contain errors.", which matched the selector but not all filters.
     # ./spec/features/admin/bulk_order_management_spec.rb:157:in `block (5 levels) in <top (required)>'

  2) 
  As an Administrator
 altering line item properties submitting data to the server when unacceptable data is sent to the server displays an update button which submits pending changes
     Failure/Error: expect(page).to have_content "exceeds available stock. Please ensure line items have a valid quantity."
       expected to find text "exceeds available stock. Please ensure line items have a valid quantity." in "Logged In As : jamil_brakus@vonruedenlegros.ca Account Logout Store DASHBOARD ORDERS PRODUCTS REPORTS CONFIGURATION USERS ORDER CYCLES GROUPS ENTERPRISES CUSTOMERS ORDERS BULK ORDER MANAGEMENT SUBSCRIPTIONS Bulk Order Management What's this? Fields with red borders contain errors. START DATE END DATE PRODUCER All SHOP All ORDER CYCLE All ACTIONS COLUMNS NAME ORDER DATE PRODUCER PRODUCT: UNIT QUANTITY MAX ASK? John Doe November 19, 2018 Enterprise 54 Product #28 - 4810: 1g, S \"Product #28 - 4810 (1g, S)\" is out of stock. 5"
     # ./spec/features/admin/bulk_order_management_spec.rb:171:in `block (5 levels) in <top (required)>'
```

### LINKED ISSUE #3204: [Spree Upgrade] Fix line Item spec - capping quantity at stock level (2 broken tests)
Originally part of #2943
Moving to separate issue to be handled separately.
This is one of the entries in #3172 

The 2 specs related to capping quantities:
- rspec ./spec/models/spree/line_item_spec.rb:100 # Spree::LineItem capping quantity at stock level caps at zero when stock is negative
- rspec ./spec/models/spree/line_item_spec.rb:126 # Spree::LineItem capping quantity at stock level when a variant override is in place when count on hand is negative caps at zero

### In v1
line item stock level check was overriden here (we added skip_stock_check)
https://github.com/openfoodfoundation/openfoodnetwork/blob/046ba14da4111176270d8e9ddaf8164b9dc59bf7/app/models/spree/line_item_decorator.rb#L160-L163

and using sufficient_stock? also overridden here (we added scoping):
https://github.com/openfoodfoundation/openfoodnetwork/blob/046ba14da4111176270d8e9ddaf8164b9dc59bf7/app/models/spree/line_item_decorator.rb#L145-L154

### In v2
The stock check in line item is now done using
`validates_with Stock::AvailabilityValidator` that depends on Stock::Quantifier. Stock::Quantifier depends on stock_items. Variant Overrides don't have stock items in our current approach to them in v2.
Also, sufficient_stock? is now implemented through a Quantifier. same challenge as for the stock check.

luisramos0:
#### What? Why?

Closes #3090 and #3037 (see skip_stock_check below) and #3204 (cap line item quantity at stock level).

As described in #3090 ([here](https://github.com/openfoodfoundation/openfoodnetwork/issues/3090#issuecomment-446777857)) this is a spree bug fixed in spree 2.4 that these OFN specs detect.

We could adapt the specs and ignore the spree bug, I dont think that's a valid alternative, specially because the bug is on a basic feature: changing line items quantity on a completed order.

Plain english explanation of the bug: in a completed order with a line_item with quantity 5, when you change the quantity from 5 to 4 on that completed order, because the 5 items were already removed from available stock on order completion, the stock availability check should check the inventory_units of the order (the 5 items) and succeed. From spree v2.0.4 to spree 2.4 the validator is checking if there are 4 (extra) available items in stock...

Bringing the commits from the fix in spree 2.4 to the current ofn spree fork would be painful, they include DB changes.
So, this PR, adapts the availability_validator to fix the bug.

One extra reason to do this is that we can introduce OFN specific stuff like:
- skip_stock_check condition (by including it in this validator we fix #3037), see commit "Fix order factory"
- capping line_item.quantity at zero (we fix #3204 with this fix), see commit "Fix capping quantity"

AND, we still need to handle scoping in this context.

#### What should we test?
2 specs from #3090, 2 from #3037 and 2 from #3204 are green:
- rspec ./spec/features/admin/bulk_order_management_spec.rb:151 - #3090
- rspec ./spec/features/admin/bulk_order_management_spec.rb:163 - #3090
- rspec ./spec/services/order_factory_spec.rb:73 - #3037
- rspec ./spec/services/order_factory_spec.rb:96 - #3037 
- rspec ./spec/models/spree/line_item_spec.rb:100 - #3204
- rspec ./spec/models/spree/line_item_spec.rb:126  - #3204

re bulk_order_management_spec, the test order was changed so that it includes inventory_units that can be checked by the validator.

### DIFF
--- a/app/models/spree/stock/availability_validator_decorator.rb
+++ b/app/models/spree/stock/availability_validator_decorator.rb
@@ -1,5 +1,8 @@
 Spree::Stock::AvailabilityValidator.class_eval do
   def validate(line_item)
+    # OFN specific check for in-memory :skip_stock_check attribute
+    return if line_item.skip_stock_check
+
     quantity = adapt_line_item_quantity_to_inventory_units(line_item)
 
     validate_quantity(line_item, quantity)
--- a/app/models/spree/stock/availability_validator_decorator.rb
+++ b/app/models/spree/stock/availability_validator_decorator.rb
@@ -3,10 +3,10 @@ Spree::Stock::AvailabilityValidator.class_eval do
     # OFN specific check for in-memory :skip_stock_check attribute
     return if line_item.skip_stock_check
 
-    quantity = adapt_line_item_quantity_to_inventory_units(line_item)
-    return if quantity == 0
+    quantity_to_validate = line_item.quantity - quantity_in_shipment(line_item)
+    return if quantity_to_validate < 1
 
-    validate_quantity(line_item, quantity)
+    validate_quantity(line_item, quantity_to_validate)
   end
 
   private
@@ -14,12 +14,12 @@ Spree::Stock::AvailabilityValidator.class_eval do
   # This is an adapted version of a fix to the inventory_units not being considered here.
   # See #3090 for details.
   # This can be removed after upgrading to Spree 2.4.
-  def adapt_line_item_quantity_to_inventory_units(line_item)
+  def quantity_in_shipment(line_item)
     shipment = line_item_shipment(line_item)
-    return line_item.quantity unless shipment
+    return 0 unless shipment
 
     units = shipment.inventory_units_for(line_item.variant)
-    line_item.quantity - units.count
+    units.count
   end
 
   def line_item_shipment(line_item)
--- a/app/models/spree/stock/availability_validator_decorator.rb
+++ b/app/models/spree/stock/availability_validator_decorator.rb
@@ -1,0 +1,37 @@
+Spree::Stock::AvailabilityValidator.class_eval do
+  def validate(line_item)
+    quantity = adapt_line_item_quantity_to_inventory_units(line_item)
+
+    validate_quantity(line_item, quantity)
+  end
+
+  private
+
+  # This is an adapted version of a fix to the inventory_units not being considered here.
+  # See #3090 for details.
+  # This can be removed after upgrading to Spree 2.4.
+  def adapt_line_item_quantity_to_inventory_units(line_item)
+    shipment = line_item_shipment(line_item)
+    return line_item.quantity unless shipment
+
+    units = shipment.inventory_units_for(line_item.variant)
+    line_item.quantity - units.count
+  end
+
+  def line_item_shipment(line_item)
+    return line_item.target_shipment if line_item.target_shipment
+    return line_item.order.shipments.first if line_item.order.present? && line_item.order.shipments.any?
+  end
+
+  # This is the spree v2.0.4 implementation of validate
+  # But using the calculated quantity instead of the line_item.quantity.
+  def validate_quantity(line_item, quantity)
+    quantifier = Spree::Stock::Quantifier.new(line_item.variant_id)
+    return if quantifier.can_supply? quantity
+
+    variant = line_item.variant
+    display_name = %Q{#{variant.name}}
+    display_name += %Q{ (#{variant.options_text})} unless variant.options_text.blank?
+    line_item.errors[:quantity] << Spree.t(:out_of_stock, :scope => :order_populator, :item => display_name.inspect)
+  end
+end
--- a/app/models/spree/stock/availability_validator_decorator.rb
+++ b/app/models/spree/stock/availability_validator_decorator.rb
@@ -4,6 +4,7 @@ Spree::Stock::AvailabilityValidator.class_eval do
     return if line_item.skip_stock_check
 
     quantity = adapt_line_item_quantity_to_inventory_units(line_item)
+    return if quantity == 0
 
     validate_quantity(line_item, quantity)
   end
--- a/spec/models/spree/stock/availability_validator_spec.rb
+++ b/spec/models/spree/stock/availability_validator_spec.rb
@@ -9,16 +9,11 @@ module Spree
         let(:order) { create(:order_with_line_items) }
         let(:line_item) { order.line_items.first }
 
-        before do
-          expect(order.shipment.inventory_units).to be_empty
-          expect(line_item.target_shipment).to be_nil
-        end
-
         context "available quantity when variant.on_hand > line_item.quantity" do
           it "suceeds" do
             line_item.quantity = line_item.variant.on_hand - 1
             validator.validate(line_item)
-            expect(line_item.errors[:quantity].size).to eq(0)
+            expect(line_item).to be_valid
           end
         end
 
@@ -26,14 +21,14 @@ module Spree
           it "fails" do
             line_item.quantity = line_item.variant.on_hand + 1
             validator.validate(line_item)
-            expect_product_out_of_stock_error
+            expect(line_item).not_to be_valid
           end
 
           it "succeeds with line_item skip_stock_check" do
             line_item.skip_stock_check = true
             line_item.quantity = line_item.variant.on_hand + 1
             validator.validate(line_item)
-            expect(line_item.errors[:quantity].size).to eq(0)
+            expect(line_item).to be_valid
           end
         end
       end
@@ -42,33 +37,19 @@ module Spree
         let(:order) { create(:completed_order_with_totals) }
         let(:line_item) { order.line_items.first }
 
-        before do
-          expect(line_item.quantity).to eq(1)
-          expect(line_item.variant.on_hand).to eq(4)
-
-          expect(order.shipment.inventory_units).to_not be_empty
-          expect(order.shipment.inventory_units_for(line_item.variant).size).to eq(1)
-        end
-
         context "when adding all variant.on_hand quantity to existing line_item quantity" do
           it "succeeds because it excludes existing inventory units from the validation" do
             line_item.quantity += line_item.variant.on_hand
             validator.validate(line_item)
-            expect(line_item.errors[:quantity].size).to eq(0)
+            expect(line_item).to be_valid
           end
 
           it "fails if one more item is added" do
             line_item.quantity += line_item.variant.on_hand + 1
             validator.validate(line_item)
-            expect_product_out_of_stock_error
+            expect(line_item).not_to be_valid
           end
         end
-      end
-
-      def expect_product_out_of_stock_error
-        quantity_error = line_item.errors[:quantity].first
-        expect(quantity_error).to include(line_item.variant.product.name)
-        expect(quantity_error).to include("out of stock")
       end
     end
   end
--- a/spec/models/spree/stock/availability_validator_spec.rb
+++ b/spec/models/spree/stock/availability_validator_spec.rb
@@ -1,0 +1,75 @@
+require 'spec_helper'
+
+module Spree
+  module Stock
+    describe AvailabilityValidator do
+      let(:validator) { AvailabilityValidator.new({}) }
+
+      context "line item without existing inventory units" do
+        let(:order) { create(:order_with_line_items) }
+        let(:line_item) { order.line_items.first }
+
+        before do
+          expect(order.shipment.inventory_units).to be_empty
+          expect(line_item.target_shipment).to be_nil
+        end
+
+        context "available quantity when variant.on_hand > line_item.quantity" do
+          it "suceeds" do
+            line_item.quantity = line_item.variant.on_hand - 1
+            validator.validate(line_item)
+            expect(line_item.errors[:quantity].size).to eq(0)
+          end
+        end
+
+        context "unavailable quantity when variant.on_hand < line_item.quantity" do
+          it "fails" do
+            line_item.quantity = line_item.variant.on_hand + 1
+            validator.validate(line_item)
+            expect_product_out_of_stock_error
+          end
+
+          it "succeeds with line_item skip_stock_check" do
+            line_item.skip_stock_check = true
+            line_item.quantity = line_item.variant.on_hand + 1
+            validator.validate(line_item)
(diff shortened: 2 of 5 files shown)


luisramos0:
@mkllnk @sauloperez this availability_validator_decorator is a disaster: class_eval with custom ofn stuff mixed with future spree bug fixes...
 I dont want to delay the upgrade but I am very open to other alternatives! that's also the reason I didnt even cover this with tests...

app/models/spree/stock/availability_validator_decorator.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    validate_quantity(line_item, quantity)
+  end
+
+  private
+
+  # This is an adapted version of a fix to the inventory_units not being considered here.
+  # See #3090 for details.
+  # This can be removed after upgrading to Spree 2.4.
+  def adapt_line_item_quantity_to_inventory_units(line_item)
```
mkllnk: What do you think about this name?
```suggestion
  def unshipped_quantity(line_item)
```

luisramos0:
ok, thanks @mkllnk for some encouragement!

I will cover it with tests now.

app/models/spree/stock/availability_validator_decorator.rb
luisramos0: thanks for this. I ended up changing the method to return a quantity and then called it quantity_in_shipment, because inventory_units can be already shipped as well.

luisramos0:
ok, I added tests. Ready for review.

luisramos0:
true :+1:

app/models/spree/stock/availability_validator_decorator.rb
```diff
@@ -0,0 +1,41 @@
...
+  # This is the spree v2.0.4 implementation of validate
+  # But using the calculated quantity instead of the line_item.quantity.
+  def validate_quantity(line_item, quantity)
+    quantifier = Spree::Stock::Quantifier.new(line_item.variant_id)
+    return if quantifier.can_supply? quantity
+
+    variant = line_item.variant
+    display_name = %Q{#{variant.name}}
+    display_name += %Q{ (#{variant.options_text})} unless variant.options_text.blank?
+    line_item.errors[:quantity] << Spree.t(:out_of_stock, :scope => :order_populator, :item => display_name.inspect)
```
sauloperez: we should stop using hashrockets for hashes and use symbols IMO. It's been 7 years since this syntax was introduced :see_no_evil:

spec/models/spree/stock/availability_validator_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    describe AvailabilityValidator do
+      let(:validator) { AvailabilityValidator.new({}) }
+
+      context "line item without existing inventory units" do
+        let(:order) { create(:order_with_line_items) }
+        let(:line_item) { order.line_items.first }
+
+        before do
+          expect(order.shipment.inventory_units).to be_empty
+          expect(line_item.target_shipment).to be_nil
```
sauloperez: why would we put an assertion in the setup phase? if you are trying to ensure the objects are in a particular state, why not create them like that? To me, this leads to unconfident tests. It's like we don't trust our test setup and factories and make the test suite even slower.

spec/models/spree/stock/availability_validator_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        end
+
+        context "available quantity when variant.on_hand > line_item.quantity" do
+          it "suceeds" do
+            line_item.quantity = line_item.variant.on_hand - 1
+            validator.validate(line_item)
+            expect(line_item.errors[:quantity].size).to eq(0)
```
sauloperez: what about using `expect(line_item).to be_valid`?. Conveys a bit more meaning IMO.

spec/models/spree/stock/availability_validator_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+          end
+        end
+
+        context "unavailable quantity when variant.on_hand < line_item.quantity" do
+          it "fails" do
+            line_item.quantity = line_item.variant.on_hand + 1
+            validator.validate(line_item)
+            expect_product_out_of_stock_error
```
sauloperez: what about replacing this with `expect(line_item).not_to be_valid`?

app/models/spree/stock/availability_validator_decorator.rb
luisramos0: yes, I know. but these couple of lines are spree code as is, I'd leave it as it is to make it easier for the next developer to see that.

spec/models/spree/stock/availability_validator_spec.rb
luisramos0: this is a familiar pattern to me, I am using standard app factories and before I do my testing I make sure the factories are giving me what I need for the test. so if someone changes the factories, they will need to take care of this spec. I think this is better than to force target_shipment null on the line_item because I want this spec to break if the line_item factory is changed in terms  of target_shipment. So, I want the future dev that changes the line_item.target_shipment in the generic factory to see this breaking.
same for shipment.inventory_units.

It's also avoiding having to do what the factories have already done.

luisramos0: You made me think with this one. And my conclusion is that I like this pattern.
This is the _given_ part in BDD (given/when/then), it's given, doesn't necessarily need to be composed of creation statements, can be assertion. I find it really natural to have expectations in this phase.

spec/models/spree/stock/availability_validator_spec.rb
luisramos0: yes, nice, thanks, done!

spec/models/spree/stock/availability_validator_spec.rb
luisramos0: done

spec/models/spree/stock/availability_validator_spec.rb
sauloperez: I don't know. It feels totally weird to me. It's the first time I see assertions in the setup phase.

> I make sure the factories are giving me what I need for the test. so if someone changes the factories, they will need to take care of this spec.

If the problem is to have valid factories, why not specific tests for them? you could even TDD while implementing the factories if you want. But why would we test their validity with every single use of them? it's enough to check it one time, not thousands. Think also how much slower our test suite would become if we applied this pattern in all tests.

> So, I want the future dev that changes the line_item.target_shipment in the generic factory to see this breaking.

If you change a generic factory tests that rely on it will break. Why the extra step?

> It's also avoiding having to do what the factories have already done.

But I wouldn't redo what factories do. If I understand you correctly, there's no point on setting up the test independently from the factory. If a particular factory sets up an object in a certain state, let's make use of it in the test. No point on redoing it.

That's what you when you first create a factory. You move the setup code you had in the test into a factory and then the test becomes simpler while making the setup reusable.

> This is the given part in BDD (given/when/then), it's given, doesn't necessarily need to be composed of creation statements, can be assertion.

But this is a plain simple unit test. The good old and arcane Setup, Exercise, Verify and Teardown phases. I feel like we're somehow reinventing things that are well established and understood by devs.

sauloperez: It might as well be that I'm missing something.

luisramos0: I put these lines because I think they are very important for the reader to understand the test conditions. basically, no target_shipment and no inv_units on the order.shipment
then later we have a test with no target_shipment and with inv_units.
these are things implicit in the very generic factories used order_with_line_items and then completed_order_with_totals.
I think it makes it obvious the conditions being tested.
but look, I am ok with anything actually, what's your suggestion? delete these expect statements? I can do that although I think the test wouldnt read so well.

luisramos0: re reinventing things, this is simple BDD at unit test level. nothing new.
a major dimension of BDD is test readability.

app/models/spree/stock/availability_validator_decorator.rb
mkllnk: I agree to keep Spree code in it's original form. One improvement would be to add a reference to the original code: https://github.com/spree/spree/blob/v2.0.4/core/app/models/spree/stock/availability_validator.rb

I wouldn't have introduced the guard clause either, but I'm happy with the current version.

Just to play around with ideas. If we wanted to keep and use the original code, we could have wrapped the line item in a new proxy object that acts like the original line item but returns our given quantity. That would make the separation clearer, but it's a lot of hassle for this little change. I'm still happy with the current version. :+1:

spec/models/spree/stock/availability_validator_spec.rb
mkllnk: If the *given* state is important for the test, I would move the expect statements into the test where they matter. In some cases it can be written like this:
```rb
expect(object).to be_empty
add_something
expect(object).to_not be_empty

# or
expect(object).to be_empty
do_something_not_adding
expect(object).to be_empty # still
```

And that's where people invented the `change` expectation. The alternative is what Pau said about making it explicit in the setup:
```rb
object = create(:object, empty: true)
# or
object = create(:object, items: [])
```

Then you trust the factory to do the right thing and don't need to repeat it in the test. I think either is valid and it depends on the context. But I agree with Pau that the `before` block is not the natural habitat of `expect` statements. ;-)

I guess the only thing that is really important to me here and I think we all agree on that already is to **not** rely on factory defaults:

```rb
# BAD
item = create(:line_item)
item.add_stock(3)
expect(item.stock).to eq 8
# the reader has to know that the factory uses a default stock of 5
```

luisramos0: ok, I will delete these lines.

>>> the underlying concern we all have is that we are not 100% confident about our factories

I trust the factories, with theses line I just wanted to communicate better the important details about them. the objective was readability.

luisramos0: > Personally, I find them too big and so hard to know what they actually create for me. 

I see. These expects are a simple solution to that problem, you use the factories and assert what you need from them before your test.

app/models/spree/stock/availability_validator_decorator.rb
luisramos0: I am actually fixing this code (at least the hash rocckets!) in the next PR here :-) https://github.com/openfoodfoundation/openfoodnetwork/pull/3232/files#diff-d5194b883a5ff6787c65bd6c4175a4b2R42

spec/models/spree/stock/availability_validator_spec.rb
sauloperez: > if the factories are changed, the tests will most likely break.

That's what I understand

luisramos0:
All right! Thank you guys! Merging...

Issue #3141, opened 2018-11-26
Rubocop `Metrics/BlockLength` doesn't like DSL in specs


kristinalim:
Rubocop complains that there are too many lines in the block.

## Description

See result from Code Climate:

![20181126121211-block-length](https://user-images.githubusercontent.com/2243/48992780-afa31100-f174-11e8-9d77-1842471888a4.png)

We would normally end up with long `describe` and `context` blocks because of how we nest them.

## Expected Behavior

Rubocop should not complain that the `describe` and `context` blocks are long.

## Actual Behavior

Rubocop complains that the `describe` and `context` blocks are long.

## Steps to Reproduce

Run the following:

```sh
bundle exec rubocop -c .rubocop_styleguide.yml spec/features/consumer/registration_spec.rb 
```

Notice the following:

```
Offenses:

spec/features/consumer/registration_spec.rb:3:1: C: Metrics/BlockLength: Block has too many lines. [129/25]
feature "Registration", js: true do ...
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/features/consumer/registration_spec.rb:7:3: C: Metrics/BlockLength: Block has too many lines. [83/25]
  describe "Registering a Profile" do
```

## Context

A new, long spec file for #3115 is making Rubocop complain of this.

## Severity

tech debt

PR #3213, opened 2018-12-14
Ignore `it` in Rubocop BlockLength

kristinalim:
#### What? Why?

Relates to #3141 which was only partly addressed by PR #3142 

We usually end up with long `it` blocks in Ruby specs because we are explicit, need to do a lot of scenario setup, and have to make many assertions. The current config makes Rubocop complain that the blocks are too long.

In #3142, we already made the cop ignore `context` and `describe` blocks. We should have done the same for `it` blocks.

Like in #3142, that wasn't a problem before because of this config in `.codeclimate.yml`:

```yaml
exclude_patterns:
- "spec/**/*"
```

However, recently, we added domain engines, which are in `engines/`. The above pattern does not exclude files in `engines/*/spec/`.

As I mentioned, we can also simply exclude `engines/*/spec/**/*`, but we should not do this.

#### What should we test?

Nothing to test, but the Code Climate check for this PR should pass.

#### Release notes

- Disable Rubocop check `Metrics/BlockLength` for `it` blocks.

Changelog Category: Changed

### DIFF
--- a/.rubocop_styleguide.yml
+++ b/.rubocop_styleguide.yml
@@ -187,7 +187,7 @@ Metrics/AbcSize:
   Max: 15
 
 Metrics/BlockLength:
-  ExcludedMethods: ["describe", "context"]
+  ExcludedMethods: ["context", "describe", "it"]
 
 Metrics/BlockNesting:
   Max: 3


sauloperez:
@Matt-Yorkley @mkllnk @HugsDaniel anyone against it? if not, please merge.

PR #3229, opened 2018-12-19
3127 More robust tax rates report

### LINKED ISSUE #3127: Loading Tax Rates report gave snail page
## Description
I was trying to view the Tax Rates report in Aus production, logged in as a regular enterprise user. When I tried to view the report I got an instant snail page (not a page timeout).

Could this be because Aus Production has 3 taxes setup?

## Expected Behavior
I should be able to view the report

## Actual Behavior
I got the snail page.

## Steps to Reproduce
1. Login to Aus Production (as enterprise or superadmin)
2. Go to report and select the Tax Rates report 
3. Select an enterprise 
4. Click Search and you should quickly see a snail page.

## Animated Gif/Screenshot
image

## Context
Trying to view tax report

## Severity
s4 - I'm not sure how many people use this report?

## Your Environment
PC, Chrome, Aus Production

mkllnk:
#### What? Why?

Closes #3127

It is possible to delete line items without deleting their associated fees which can include tax. Orders containing these orphaned fees broke the tax rates report.

While we should work on keeping our database clean and consistent, this change makes tax rate reporting much more robust and fixes the immediate issue #3127.

It also encapsulates the tax rate finding logic in a new class, not polluting Spree's namespace.

#### What should we test?

- Go to Enterprises / Enterprise Fees and add a packing fee with flat percent per item.
- Adjust the percentage to 10%.
- Include that fee in an order cycle.
- Create an order with multiple line items.
- Go to admin / orders / edit order.
- Observer that there is a fee for each item.
- Click the trash icon to delete an item, but don't click "Update And Recalculate Fees".
- Observe that the line item is gone, but the fee is still there.
- Go to the tax rates report.
- Get a report that includes the modified order.
- Observe that the page doesn't crash, but the tax rate of the enterprise fee is displayed.

#### Release notes

Fixed: The tax rates report is working again. It crashed with orders containing orphaned enterprise fees including tax.

Changelog Category: Fixed

### DIFF
--- a/app/services/tax_rate_finder.rb
+++ b/app/services/tax_rate_finder.rb
@@ -1,0 +1,33 @@
+# Finds tax rates on which an adjustment is based on.
+# For example a packaging fee may contain VAT. This service finds the VAT rate
+# for the tax included in the packaging fee.
+class TaxRateFinder
+  # @return [Array<Spree::TaxRate>]
+  def tax_rates(originator, source, amount, included_tax)
+    case originator
+    when Spree::TaxRate
+      [originator]
+    when EnterpriseFee
+      case source
+      when Spree::LineItem
+        tax_category = originator.inherits_tax_category? ? source.product.tax_category : originator.tax_category
+        tax_category ? tax_category.tax_rates.match(source.order) : []
+      when Spree::Order
+        originator.tax_category ? originator.tax_category.tax_rates.match(source) : []
+      end
+    else
+      find_closest_tax_rates_from_included_tax(amount, included_tax)
+    end
+  end
+
+  private
+
+  # shipping fees and adjustments created from the admin panel have
+  # taxes set at creation in the included_tax field without relation
+  # to the corresponding TaxRate, so we look for the closest one
+  def find_closest_tax_rates_from_included_tax(amount, included_tax)
+    approximation = (included_tax / (amount - included_tax))
+    return [] if approximation.infinite? || approximation.zero?
+    [Spree::TaxRate.order("ABS(amount - #{approximation})").first]
+  end
+end
--- a/app/services/tax_rate_finder.rb
+++ b/app/services/tax_rate_finder.rb
@@ -4,6 +4,13 @@
 class TaxRateFinder
   # @return [Array<Spree::TaxRate>]
   def tax_rates(originator, source, amount, included_tax)
+    find_associated_tax_rate(originator, source) ||
+      find_closest_tax_rates_from_included_tax(amount, included_tax)
+  end
+
+  private
+
+  def find_associated_tax_rate(originator, source)
     case originator
     when Spree::TaxRate
       [originator]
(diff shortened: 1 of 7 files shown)


app/services/tax_rate_finder.rb
```diff
@@ -0,0 +1,58 @@
...
+    if enterprise_fee.inherits_tax_category?
+      line_item.product.tax_category
+    else
+      enterprise_fee.tax_category
+    end
+  end
+
+  # shipping fees and adjustments created from the admin panel have
+  # taxes set at creation in the included_tax field without relation
+  # to the corresponding TaxRate, so we look for the closest one
```
luisramos0: thanks for the comment but I've read this code 10 times already and I still don't understand what is this concept of using the closest one... maybe an example would help here.

spec/services/tax_rate_finder_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        tax_rate,
+        shipment,
+        amount,
+        included_tax
+      )
+      expect(rates).to eq [tax_rate]
+    end
+
+    it "finds a close match" do
+      close_tax_rate
```
sauloperez: If this tax rate is only used here why not having the following?

```suggestion
      close_tax_rate = create_rate(0.25)
```

It's weird for the reader to see a method call named after a noun without an assignment. Also, having a let at the top it makes you think it's an object that will be used throughout all tests, while it's not.

app/services/tax_rate_finder.rb
sauloperez: It's weird yes, but it seems like from admin shipping fees and adjustments miss an association to tax rates. Shouldn't we fix this instead of having to infer which tax rate it is? it looks really broken to me. I suggest we create an issue.

myriamboure:
You talk about one tax report but there are two, one by rates, just want to make sure that the refactoring doesn't create bug in the second one, did you check that one as well @mkllnk ? Cause we only use the report "by tax rate" in France, I remember when Pierre worked on it it was quite complex to get it work... just checking ;-)

app/services/tax_rate_finder.rb
mkllnk: @luisramos0 I extended that comment and added an example. Let me know if you like it.

@sauloperez I agree that we should fix the interface as well. I opened #3278 for that.

spec/services/tax_rate_finder_spec.rb
mkllnk: Thank you for the feedback. I followed your suggestion and discovered that the spec was testing the wrong thing. Have a look at the last commit and let us know if you are happy with the pull request.

mkllnk:
@myriamboure My changes only affect the tax rates report, not the tax types report.

app/services/tax_rate_finder.rb
luisramos0: wow, thanks @mkllnk now I see why I didnt get it, it's broken and complicated :-) thanks for explaining in detail, explaining/understanding is a good first step to improve it :-)

mkllnk: Yes, I learned a lot just by writing that comment. I didn't know any of this when starting to work on this issue.

RachL:
Deploying on staging AU

RachL:
Everything seems food. thx for the detailed step to re-produce the issue @mkllnk !

My testing notes:

https://docs.google.com/document/d/1URkD6FmFR3Fg613iRI6c6ZA0C9MWdk3z4hJ0QE9Ph-Y/edit#

PR #3230, opened 2018-12-19
3222 Improve stock levels calculation code on cart populate and add specs to test handling of variant overrides

### LINKED ISSUE #3222: Broken stock levels on product list page when product added and removed quickly from basket
## Description
I am pretty sure there is a bug in this commit/line:
https://github.com/openfoodfoundation/openfoodnetwork/commit/a26266159cefcdc55c2e1bff7d5498a9d88c069d#diff-0e4d2deb3b86720d428f95f51c0c19dfR65
This is calculating stock_levels (on_hand) for variants that are listed in the request but are not present in the order (prior to this code is executed, the method li_stock_levels fetches stock levels for variants present in the order). The problem is that variants coming from the order are scoped in the cart service (are transformed into overrides) but this Spree::Variant.find call does not scope variants, so the on_hand value sent is the variants one when it should be the override on_hand value if a override exists.

I am not too sure about this logic but looking at the commit this code was added to fix a problem with multiple requests where frontend code requests stocklevels for a specific list of variants but when the server executes, the variants are not in the order any more (probably due to another request removing them). We should keep the logic but fix the bug.

## Expected Behavior
In some cases on the product list when adding and removing quantities, stock levels reported will be the variants values even if buying from a hub with inventory.

## Actual Behavior
Inventory (variant overrides) stock levels are reported instead.

## Steps to Reproduce
I didn't replicate this and probably better just to get this covered in specs. I believe manual verification will be difficult.

## Context
Detected while fixing Spree Upgrade build.

## Severity

## Possible Fix
So, a few things need to be done here:
1. create spec to validate the bug exists
2. fix the bug by scoping the variant before calling on_hand
3. refactor code! the method names at least need to be improved (the current best way to understand what's going on is to read the tests [here](https://github.com/openfoodfoundation/openfoodnetwork/commit/a26266159cefcdc55c2e1bff7d5498a9d88c069d#diff-cc6b7d88d7ef9e9a51819e616293808fR98))

The fix to the bug will be something like this:
```
      variant = Spree::Variant.find(variant_id)
      cartService.scoper.scope(variant)
      stock_levels[variant_id] = { quantity: 0, max_quantity: 0, on_hand: variant.on_hand }
```

I found this issue because this is where #3118 is breaking in the v2 branch. This is only seen in v2 variant overrides specs because order.add_variant is not working for the spec in #3118. To be investigated as part of #3118

luisramos0:
#### What? Why?

In this PR I wrote some specs to try to replicate #3222 but I didnt manage to reach the code, so this PR consists of a refactored version of the code, new specs to exercise related test cases and comments in spec and code to indicate the potential issue.

#### What should we test?
Interaction with the cart in terms of adding and removing items from it.

#### Release notes
Changelog Category: Changed
Improved code in cart population and stock levels management in the cart.

This was detected in the spree upgrade process, see #3222 for more details.

### DIFF
--- a/app/services/variants_stock_levels.rb
+++ b/app/services/variants_stock_levels.rb
@@ -5,19 +5,31 @@ class VariantsStockLevels
   def call(order, requested_variant_ids)
     variant_stock_levels = variant_stock_levels(order.line_items)
 
-    # Potentially, the following lines are dead code, they are never reached
-    # Additionally, variants are not scoped here and so the stock levels reported would be incorrect
+    # Variants are not scoped here and so the stock levels reported are incorrect
     # See cart_controller_spec for more details and #3222
-    li_variant_ids = variant_stock_levels.keys
-    (requested_variant_ids - li_variant_ids).each do |variant_id|
+    order_variant_ids = variant_stock_levels.keys
+    missing_variant_ids = requested_variant_ids - order_variant_ids
+    missing_variant_ids.each do |variant_id|
       variant_on_hand = Spree::Variant.find(variant_id).on_hand
       variant_stock_levels[variant_id] = { quantity: 0, max_quantity: 0, on_hand: variant_on_hand }
     end
+
+    # The code above is most probably dead code, this bugsnag notification will confirm it
+    notify_bugsnag(order, requested_variant_ids, order_variant_ids) if missing_variant_ids.present?
 
     variant_stock_levels
   end
 
   private
+
+  def notify_bugsnag(order, requested_variant_ids, order_variant_ids)
+    error_msg = "VariantsStockLevels.call with variants in the request that are not in the order"
+    Bugsnag.notify(RuntimeError.new(error_msg),
+                   requested_variant_ids: requested_variant_ids.as_json,
+                   order_variant_ids: order_variant_ids.as_json,
+                   order: order.as_json,
+                   line_items: order.line_items.as_json)
+  end
 
   def variant_stock_levels(line_items)
     Hash[
--- a/app/services/variants_stock_levels.rb
+++ b/app/services/variants_stock_levels.rb
@@ -5,6 +5,9 @@ class VariantsStockLevels
   def call(order, requested_variant_ids)
     variant_stock_levels = variant_stock_levels(order.line_items)
 
+    # Potentially, the following lines are dead code, they are never reached
+    # Additionally, the variants are not scoped here and so the stock levels reported would be incorrect
+    # See cart_controller_spec for more details and #3222
     li_variant_ids = variant_stock_levels.keys
     (requested_variant_ids - li_variant_ids).each do |variant_id|
       variant_on_hand = Spree::Variant.find(variant_id).on_hand
--- a/app/services/variants_stock_levels.rb
+++ b/app/services/variants_stock_levels.rb
@@ -1,0 +1,35 @@
+# Report the stock levels of:
+#   - all variants in the order
+#   - all requested variant ids
+class VariantsStockLevels
+  def call(order, requested_variant_ids)
+    variant_stock_levels = variant_stock_levels(order.line_items)
+
+    li_variant_ids = variant_stock_levels.keys
+    (requested_variant_ids - li_variant_ids).each do |variant_id|
+      variant_on_hand = Spree::Variant.find(variant_id).on_hand
+      variant_stock_levels[variant_id] = { quantity: 0, max_quantity: 0, on_hand: variant_on_hand }
+    end
+
+    variant_stock_levels
+  end
+
+  private
+
+  def variant_stock_levels(line_items)
+    Hash[
+      line_items.map do |line_item|
+        [line_item.variant.id,
+          { quantity: line_item.quantity,
+            max_quantity: line_item.max_quantity,
+            on_hand: wrap_json_infinity(line_item.variant.on_hand) }]
+      end
+    ]
+  end
+
+  # Rails to_json encodes Float::INFINITY as Infinity, which is not valid JSON
+  # Return it as a large integer (max 32 bit signed int)
+  def wrap_json_infinity(number)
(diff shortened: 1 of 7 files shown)


spec/services/variants_stock_levels_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    create(:line_item, order: order, variant: variant_in_the_order, quantity: 2, max_quantity: 3)
+  end
+  let!(:variant_in_the_order) { create(:variant, count_on_hand: 4) }
+  let!(:variant_not_in_the_order) { create(:variant, count_on_hand: 2) }
+
+  let(:variant_stock_levels) { VariantsStockLevels.new }
+
+  before do
+    order.reload
+  end
```
sauloperez: Have you tried removing it now that tests are finished? We modify an order object from subsequent lets in several tests and they don't reload it. Why this case would be different?

luisramos0: I copied this from the original test and now I tried removing it but it is needed @sauloperez 
I am not sure why but it must be because of some relation that is not set explicit in the factory and is created only on reload.
But I found an unnecessary line! let(:product) :-D I removed that line and rebased.

sauloperez: great! at least we removed code. It must definitely be one of those callbacks...

app/services/variants_stock_levels.rb
```diff
@@ -0,0 +1,38 @@
...
+#   - all requested variant ids
+class VariantsStockLevels
+  def call(order, requested_variant_ids)
+    variant_stock_levels = variant_stock_levels(order.line_items)
+
+    # Potentially, the following lines are dead code, they are never reached
+    # Additionally, variants are not scoped here and so the stock levels reported would be incorrect
+    # See cart_controller_spec for more details and #3222
+    li_variant_ids = variant_stock_levels.keys
+    (requested_variant_ids - li_variant_ids).each do |variant_id|
```
mkllnk: If I understand your comments correctly, you are saying that `requested_variant_ids - li_variant_ids` is always empty. That means that the inside of the `each` block is never executed. It does sound bizarre. There is one case I can think of:

- Apples have a stock level of 5.
- The user adds 5 apples to their cart.
- Someone else buys 5 apples at the same time
- The cart items are capped at stock levels and the apples are removed.
- The apples are not part of the order any more, but still in the request.

Sounds very unlikely, but may be possible. I also noticed that the cart service doesn't just look at `params[:variants]`, but also at `:products` and `:quantity`. Are there other interfaces sending requests with these ids?

Just playing around with the site I noticed an even more important bug: If an item is not available anymore, the server responds with 412, but the Javascript sees that as an error and enters an infinite loop of re-trying the request. The Update button stays on "Updating...". There should be an out-of-stock message instead.

luisramos0: I am not sure about you scenario because cart_populate is after cap_quantity_at_stock so the item would still be in the order. and is cap_quantity_at_stock is not removing the item from the order? maybe some race condition between requests...

> Sounds very unlikely, but may be possible. I also noticed that the cart service doesn't just look at params[:variants], but also at :products and :quantity. Are there other interfaces sending requests with these ids?

Yes, I know. I cant find other calls to populate. The main one only sends variants.cart_service.read_products_hash . It's dead code isn't it? even the specs are not reaching it.
shall I remove it?

For that last part, we need to create an issue.

luisramos0: I am tempted to add a bugsnag dead code detector on both the stock levels calculation and the :products hash. Shall I do that?

sauloperez: I wish we had app monitoring already in place 😭 these go there, to Datadog.

spec/controllers/cart_controller_spec.rb
```diff
@@ -8,7 +8,8 @@
 
     it "returns stock levels as JSON" do
       allow(controller).to receive(:variant_ids_in) { [123] }
-      allow(controller).to receive(:stock_levels) { 'my_stock_levels' }
+      allow(VariantsStockLevels).to receive(:new).and_return(variant_stock_levels = double())
```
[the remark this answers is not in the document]
luisramos0: yes, this was the pattern in this spec. I'll fix it now.
I ended up using this for variant_stock_levels but doing something else for the cart_service double.
And I also ended up moving specs around to make the spec better.
Please see new commit.

app/services/variants_stock_levels.rb
luisramos0: I am not sure how to use datadog for this... let me know if you think it's worth going through that now.

I have added bugsnag notifications, please see the new commit.

RachL:
@luisramos0 deployment failed on my side :( can you deployed it on Katuma? thx

luisramos0:
I have rebased and fixed rubocop_todo conflict. Before we test, let's just wait for the build to be green.

luisramos0:
just the usual unstable spec broken. ready for testing!

RachL:
@luisramos0  because of the broken spec I can't deploy it :disappointed: I don't have my magic button on semaphore

sigmundpetersen:
Hey @RachL I think you have the powers to rebuild the PR yourself? "Rebuild this revision" here https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3230/builds/16

sigmundpetersen:
It passed! 😄

RachL:
🎉 thx @sigmundpetersen !

Deploying on Aus staging

RachL:
Seems good both on desktop and mobile !

Testing notes: https://docs.google.com/document/d/1vP14xpXD8nInqpaSqxDyinyzmc_tSY8OUsF8v12PoaE/edit#

luisramos0:
oh my god @RachL amazing gif and amazing autonomy!!!

PR #3238, opened 2018-12-21
Fix/instagram pattern

albarnaz:
#### Issue: #1760 

There is no suggestion in the Instagram empty field in enterprise profile setup menu. Some users copy-paste the full URL on their page. That doesn't work because what is expected is only their Instagram user pseudo.

image

#### FIX

Add an before save to check if the user enter the whole URL, and extract and store just the handle.

So, if the user enter "www.instagram.com/my_user", we store just "my_user".

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -333,7 +333,7 @@ class Enterprise < ActiveRecord::Base
   end
 
   def instagram=(value)
-    write_attribute(:instagram, value.try(:gsub, instagram_regex, '\1'))
+    self[:instagram] = value.try(:gsub, instagram_regex, '\1')
   end
 
   protected
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -80,6 +80,8 @@ class Enterprise < ActiveRecord::Base
   before_validation :set_unused_address_fields
   after_validation :geocode_address
 
+  validates :instagram, format: /\A[a-zA-Z0-9._]{1,30}\z/, allow_blank: true
+
   after_touch :touch_distributors
   after_create :set_default_contact
   after_create :relate_to_owners_enterprises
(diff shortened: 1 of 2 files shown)


luisramos0:
Hello, thanks for your PR @albarnaz 
I triggered a build rerun to double check but an error in the registration spec may be related to this PR?

Also, there's a code climate issue:
Prefer self[:attr] = val over write_attribute(:attr, val). NEW
    write_attribute(:instagram, value.try(:gsub, instagram_regex, '\1'))
Severity: MinorFound in app/models/enterprise.rb by rubocop

albarnaz:
Hi @luisramos0 
I fixed the code climate issue, but this build error I think is not related to this PR.

sauloperez:
It must something related to this PR as the failing spec is `Registering a Profile Allows a logged in user to register a profile` right after the step

```ruby
fill_in 'enterprise_instagram', with: '@InStAgRaM'
perform_and_ensure(:click_button, "Continue", lambda { page.has_content? 'Finished!' })
```

There must be a validation error of some sort that doesn't move the page to the `Finished!` state.

albarnaz:
@mkllnk Ok, I will work on that.
Thanks.

daniellemoorhead:
Hi @albarnaz 👋 
Do you think you'll get time soon to finish this for a final review? It would be great to get it through to be tested and merged 😄

albarnaz:
Hi @daniellemoorhead 
Sorry for the daley.
Yes, I will finish it in the next days.

lin-d-hop:
Hi @albarnaz,
Just wondering if you've had any time for this? Keeps tripping people up here in the UK so would be super useful if we could manage to merge and release this!
Thanks!

daniellemoorhead:
@albarnaz given it's been a month I'm going to put this issue back into the `Bug Backlog` column so it can be picked up and finished either by yourself when you're available again or another dev who may want to complete the work.

Feel free to come back when you find you have some time, we hope to see you back online and working with us 😄 

Ping @lin-d-hop FYI

PR #2424, opened 2018-07-03
Replace therubyracer with mini_racer

frank-west-iii:
#### What? Why?

Closes #2100

This is being replaced to increase the reliability and speed of asset
compilation.

#### What should we test?

Assets are compiled in staging.

Issue #2100, opened 2018-02-28
Update or replace therubyracer


mkllnk:
## Description

We use the gem therubyracer 0.12.0 which is five years old. It embeds a very old version of the V8 Javascript engine. V8 had a lot of improvements done in these five years and a newer version could make our asset compilation faster and more secure. The Discourse people decided to replace it with the mini_racer and their benchmarks show that it has been a good decicion. https://github.com/discourse/mini_racer#benchmark-minification-of-discourse-applicationjs-both-minified-and-unminified

## Expected Behavior

Asset compilation is fast and succeeds.

## Actual Behavior

Asset compilation is slow and sometimes takes more memory than the server has.

## Steps to Reproduce

1. Run: `bundle exec rake RAILS_ENV=staging assets:precompile:primary RAILS_GROUPS=assets`

## Context

Someone contacted us via email about using the OFN as backend for an app. Their developer saw the outdated gem therubyracer as potential issue for security and for installing the OFN on a BSD operating system. Looking into that feedback I found that we would probably benefit from updating or replacing that gem anyway, even though the current community is not interested in using a BSD instead of Linux as operating system.

## Severity

Sometimes asset compilation can crash which requires a developer to run it manually again. Long asset compilation can cause short disruptions to users while deploying a new release, but usually we manage to do it without downtime. So I guess it's severity 5.

## Possible Fix

Update therubyracer or replace it with mini_racer.

kirstenalarsen:
@mkllnk @myriamboure @enricostano @sauloperez @daniellemoorhead  is this a bug or part of tech debt / upgrade dependencies? suspect if classified as bug-s5 if will never get done, but if we have way to track in upgrades it may be more likely to get picked up?

mkllnk:
I would classify it as tech debt, but I don't know how to do that with the new issue structure. Do we have a process or template for that?

enricostano:
Yep, tech debt. Not bug. Should be an icebox, right?

I'm still not sure how to procede about this specific kind of tasks, any idea @daniellemoorhead ?

myriamboure:
[...]

PR #3244, opened 2018-12-26
Revert to the ruby racer

luisramos0:
#### What? Why?

I can't make mini racer work locally (osx mojave) . I cant trace back to the point of failure but it did work at some point.
This is not only my problem as 2 new contributors (I believe all using osx as I am) have also reported this problem.

So, this PR reverts PR #2424.
I suggest we re-open #2100

This PR makes my life (and some new contributors with the same problem) a lot easier but makes the assets compilation move back from 211s to 316s (as noted on #2100).

#### What should we test?
Asset precompilation should still work correctly. We should make a sanity check of the app.

#### Release notes
Moved back to therubyracer for asset compilation to avoid trouble caused by the faster mini racer.
Changelog Category: Changed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,10 +88,7 @@ group :assets do
   gem 'compass-rails'
   gem 'coffee-rails', '~> 3.2.1'
 
-  gem 'mini_racer'
-  # We found that the following version of libv8 breaks the compilation of mini_racer.
-  # Nothing else depends on libv8.
-  gem 'libv8', '!= 6.7.288.46.1'
+  gem 'therubyracer', '=0.12.0'
 
   gem 'uglifier', '>= 1.0.3'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -511,7 +511,7 @@ GEM
       addressable (~> 2.3)
     letter_opener (1.6.0)
       launchy (~> 2.2)
-    libv8 (6.3.292.48.1)
+    libv8 (3.16.14.19)
     listen (3.0.8)
       rb-fsevent (~> 0.9, >= 0.9.4)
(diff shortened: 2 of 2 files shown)


luisramos0:
this build has broken twice already with random errors...
I wonder if miniracer made the build more stable, but I dont udnerstand how that could be...

luisramos0:
I have staged this in ES and confirmed precompile assets didn't error out.

sauloperez:
We should keep an eye on Mojave's support for libv8 and revert this change in the future if we can. I guess that's the underlying problem.

Matt-Yorkley:
Wow, 211s to 316s is a big jump. Is there an issue open still to investigate speeding up the precompile stage? Should we look for alternatives?

sauloperez:
IMO it's better if we focus on getting our app and dependencies up to date. This is just a symptom of how bad our situation is. Chances are that any other solutions might not work with deprecated Ruby and Rails versions...

sigmundpetersen:
Is this still on ES @luisramos0 @sauloperez ? Is it tested and OK?

luisramos0:
this is on ES yes.
it needs a sanity check from a tester.

RachL:
@luisramos0 @sauloperez staging ES URL redirects now to https://appstaging.katuma.org/ which I cannot access apparently... am I looking at the wrong place?

RachL:
thx @sauloperez I'm starting testing now :)

RachL:
Ready to go !

Testing notes : https://docs.google.com/document/d/1Ko0YqoytiPBnT_ls6lZeLxIm740FfNHLZbu2CylYx88/edit#

luisramos0:
oh my god, I cant believe, my master just works now :-D

PR #3245, opened 2018-12-26
Add spring, a rails preloader alternative to zeus, to our gemfile

luisramos0:
Also added /bin to gitignore, bin is used by spring.

#### What? Why?

zeus is not working on my osx mojave, this PR adds spring to the gemfile so that it can be used instead of zeus. I didnt manage to make it work without adding it to the gemfile. Is it acceptable for you guys to have this in the gemfile? any hints on how to make this work without changing the gemfile? Thanks!

#### What should we test?
Spring 1.1.3 should work with OFN.

Can be used locally:
```
bundle install
spring binstub rspec     (generates new rspec binary under /bin)
spring stop
time bin/rspec spec/models/order_cycle_spec.rb:154
time bin/rspec spec/models/order_cycle_spec.rb:154

```
The second run should be a lot faster.

#### Release notes
Changelog Category: Added
Added support for spring rails app preloader.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ vendor/bundle/
 coverage
 /reports/
 !/reports/README.md
+bin/
--- a/Gemfile
+++ b/Gemfile
@@ -145,6 +145,8 @@ group :development do
   gem 'guard-rails'
   gem 'guard-rspec', '~> 4.7.3'
   gem 'rubocop', '>= 0.49.1'
+  gem 'spring', '=1.1.3'
+  gem 'spring-commands-rspec'
 
   # 1.0.9 fixed openssl issues on macOS https://github.com/eventmachine/eventmachine/issues/602
   # While we don't require this gem directly, no dependents forced the upgrade to a version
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -43,3 +43,4 @@ vendor/bundle/
 coverage
 /reports/
 !/reports/README.md
+bin/
```
sauloperez: I don't see why we can't share the binstubs. I would not ignore it.

sauloperez:
https://github.com/rbenv/rbenv/wiki/Understanding-binstubs might be relevant for this.

.gitignore
luisramos0: The objective is to keep the number of files in the project to a minimum. Less complexity to take.
Also, most people will not use this as they are using zeus.
all you need to do is to run the commands above to generate them.

sauloperez: as you want. I might migrate to spring as well, to be stick to a single solution for the team.

sauloperez:
We should prepend those `spring` commands you listed in the description with `bundle exec` to ensure the project's spring is picked up.

sauloperez:
First run

```
Finished in 3.31 seconds (files took 2.7 seconds to load)
```

Second run

```
Finished in 0.55706 seconds (files took 0.86619 seconds to load)
```

Welcome to the speed of light @luisramos0 :tada: :trollface:

PR #3247, opened 2018-12-27
[Spree Upgrade] Provide anonymous user instead of nil

### LINKED ISSUE #3078: [Spree Upgrade][FrontOffice] Fix shopping/orders_spec - guest order (1 broken test)
Fix this spec:

```
1) Order Management viewing a completed order when checking out as an anonymous guest allows the user to see the details
     Failure/Error:
       create(:order_with_credit_payment,
         customer: customer,
         user: user,
         distributor: distributor,
         order_cycle: order_cycle
       )

     NoMethodError:
       undefined method `email' for nil:NilClass
     # /Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/bundler/gems/spree-f55722b38db7/core/lib/spree/testing_support/factories/order_factory.rb:6:in `block (3 levels) in <top (required)>'
….   
     # ./spec/features/consumer/shopping/orders_spec.rb:17:in `block (3 levels) in <top (required)>'

```

The new default spree order factory fetches user.email from order.email
https://github.com/openfoodfoundation/spree/blob/master/core/lib/spree/testing_support/factories/order_factory.rb#L6
This causes the error because the spec is using a nil user.

It will be related to this method
https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/spree/orders_controller_decorator.rb#L133
@kristinalim you may have a special insight into this one.
I believe the code is correct, we just need to adapt the test and/or the factories used.

sauloperez:
#### What? Why?

Closes #3078 

As explained in https://github.com/openfoodfoundation/spree_auth_devise/blob/0181835fb6ac77a05191d26f6f32a0f4a548d851/app/models/spree/user.rb#L26-L32, all orders have a user so checking out as a guest is just using an automatically generated user by means of `.anonymous!`.

#### What should we test?
spec/features/consumer/shopping/orders_spec.rb:17 should be fixed.

### DIFF
--- a/spec/features/consumer/shopping/orders_spec.rb
+++ b/spec/features/consumer/shopping/orders_spec.rb
@@ -36,7 +36,7 @@ feature "Order Management", js: true do
     end
 
     context "when checking out as an anonymous guest" do
-      let(:user) { nil }
+      let(:user) { Spree::User.anonymous! }
 
       it "allows the user to see the details" do
         # Cannot load the page without token


luisramos0:
oh, you just force-pushed. is it ready for review? the first build looks good, the spec is fixed.

sauloperez:
Yep, sorry I messed it with another branch. Now it's ready.

luisramos0:
rerunning build to see job 3 with 6 broken specs, not 8, should be random...

sauloperez:
Gald to see you happy @mkllnk 😜. Sorry, what is it that we do with customer's email?

mkllnk:
It's just that we have some data duplication around customer records in the database. Customers can be associated to an email (anonymous guest) and to a user id (registered user). Some of that is confusing and lead to bugs like https://github.com/openfoodfoundation/openfoodnetwork/issues/2841. Going with Spree's new way of having a user record for every guest could resolve some of these things.

luisramos0:
relates to this thread right Maikel?
https://community.openfoodnetwork.org/t/subscriptions-and-customers/1512

PR #3250, opened 2018-12-27
Open adjustments to enable update

### LINKED ISSUE #2685: [Spree Upgrade][Models] Fix order_spec and orders_controller spec tests with shipment fees and taxes (3 broken tests)

sauloperez:
#### What? Why?

Closes #2685

Spree supports updating line items as the model contains these two callbacks:

```ruby
after_save :update_order
after_destroy :update_order
```
(openfoodfoundation/spree:core/app/models/spree/line_item.rb@f55722b#L27-L28)

However, as the following paragraph clearly states

```
# Adjustments represent a change to the +item_total+ of an Order. Each adjustment
# has an +amount+ that can be either positive or negative.
#
# Adjustments can be open/closed/finalized
#
# Once an adjustment is finalized, it cannot be changed, but an adjustment can
# toggle between open/closed as needed
```
(openfoodfoundation/spree:core/app/models/spree/adjustment.rb@f55722b#L1-L7)

by means of the Adjustment's state machine once it is finalized it can't be updated. Then, if we want to have an up-to-date adjustment_total attribute for a complete order we must first open the adjustments. Otherwise, the following `immutable?` check does not pass 

```ruby
def update!
  return if immutable?
  # Fix for #3381
  # If we attempt to call 'source' before the reload, then source is currently
  # the order object. After calling a reload, the source is the Shipment.
  reload
  originator.update_adjustment(self, source) if originator.present?
  set_eligibility
end
```
(openfoodfoundation/spree:core/app/models/spree/adjustment.rb@f55722b#L84-L92)

I still can't tell what changed for this to fail in 2-0-stable.

#### What should we test?

The affected test should pass.


sauloperez:
I'd like to have feedback here. This is not yet finished now we need to see where this call to `adjustment.open` should go. I guess to each of the 4 LineItemsController that we have (including Spree decorators).

sauloperez:
After that it'd be worth to investigate the following lines 

https://github.com/openfoodfoundation/openfoodnetwork/blob/1007addddbc7b60dd530cb9ec67b1f4d851a8b68/app/models/spree/order_decorator.rb#L36-L37

https://github.com/openfoodfoundation/openfoodnetwork/blob/1007addddbc7b60dd530cb9ec67b1f4d851a8b68/app/models/spree/order_decorator.rb#L167-L184

Now that I understood a bit more how this works, I'm not sure we solved a problem adding more code while the fix was probably adding a call to `open` in the proper place.

sauloperez:
Yep but note how the state is restored right after. The call to update_totals cannot succeed.

I'll have to go through it again to notice any other implications per though.

luisramos0:
I am going through it right now.
What I concluded from analysing v1 is that these ofn methods using update_adjustment! are the ones updating the adjustments.amount (from 10 to 5 and from 6 to 3 in the test). In v2 this is not happening... I am trying to figure out why.

luisramos0:
But I agree with your comment above: it's probably possible to get rid of these methods by just placing a "adjustment.open" in the right place.

luisramos0:
Yeah, I dont get it...

The only thing I can add is that adding the update_adjustment! logic to order_updater.update_adjustments like this will also fix the test (without opening the adjustment in the test):
  def update_adjustments
      order.adjustments.reload.each { |adjustment| update_adjustment! adjustment }
      choose_best_promotion_adjustment
    end

I don't understand why order.update_payment_fees! is not picking up the line_item.quantity zero...

sauloperez:
Thanks for the investigation @luisramos0. It feels like the fix is to trigger the appropriate callbacks at the right moment for Spree logic to act. I'll see what else I can find.

sauloperez:
Rebased onto 2-0-stable to get the latest build and make sure I don't break any spec.

app/models/spree/order_decorator.rb
```diff
@@ -398,9 +398,11 @@ def ensure_customer
 
   def update_adjustment!(adjustment)
     return if adjustment.finalized?
+
     state = adjustment.state
     adjustment.state = 'open'
     adjustment.update!
+    update!
```
sauloperez: since manually opening the adjustment makes the test pass, this means that in v2 the `OrderUpdater` (the class that updates all the totals attributes) is not executed before the adjustment's state is restored in line 406. Therefore, triggering the update ourselves fixes it.

sauloperez:
This is the least invasive change I could think of to fix it. Once we get to green, and as we said, we should revisit these callbacks after phase 1. We don't want to fix things by triggering yet more callbacks but removing code and triggering just the ones that are needed. This will require digging really deep into Spree's models.

luisramos0:
all right, looks good enough!!!
do you want to do this for the other 2 tests?
rspec ./spec/models/spree/order_spec.rb:711 
rspec ./spec/controllers/spree/orders_controller_spec.rb:197
(#2685)

sauloperez:
Yep, let's see if I can them fixed as well.

sauloperez:
I'll keep the three fixes separate because they are not related. I opened https://github.com/openfoodfoundation/openfoodnetwork/pull/3349 for the 2nd one and I'll do the same for the last one.

sauloperez:
spec/models/spree/order_spec.rb:679 is passing, so this is ready.

sauloperez:
Good catch.

PR #3254, opened 2018-12-28
[Spree Upgrade] Create the default StockLocation in DB

### LINKED ISSUE #3253: [Spree Upgrade] Fix reference to backorderable_default from admin/products/new.html.haml
## Description

Our VariantStock module relies on the `Spree::StockLocation.first.backorderable_default` but in `2-0-stable` we don't create the default stock location yet.

This obviously causes a 500 error as follows:

```
...
Completed 500 Internal Server Error in 929.8ms (ActiveRecord: 34.8ms)

ActionView::Template::Error (undefined method `backorderable_default' for nil:NilClass):
    1: <%# This chunk is just a copy of Spree's backend/app/views/spree/admin/products/new.js.erb %>
    2: $("#new_product").html('<%= escape_javascript(render :template => "spree/admin/products/new", :formats => [:html], :handlers => [:haml]) %>');
    3: handle_date_picker_fields();
    4: <% unless Rails.env.test? %>
    5:   $('.select2').select2();
  app/models/concerns/variant_stock.rb:91:in `on_demand'
  app/models/concerns/variant_stock.rb:32:in `on_hand'
  app/models/concerns/product_stock.rb:19:in `on_hand'
  app/views/spree/admin/products/new.html.haml:57:in `block (2 levels) in _6397cc2bb50734c9d7283660ebd313d6'
  app/views/spree/admin/products/new.html.haml:53:in `block in _6397cc2bb50734c9d7283660ebd313d6'
  app/views/spree/admin/products/new.html.haml:3:in `_6397cc2bb50734c9d7283660ebd313d6'
  app/views/spree/admin/products/new.js.erb:2:in `_80cf50a20567230ea7daadc245fd7326'
  lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `call'
  lib/spree/core/controller_helpers/respond_with_decorator.rb:15:in `respond_with'
...
```

## Expected Behavior
`products/new` page should render successfully without errors and allow users to specify a value for the stock level.

## Actual Behavior
The page fails to render raising the error showed above.

## Steps to Reproduce
1. Go to `/admin/products`
2. Click on `+ NEW PRODUCT`

## Possible Fix

We need to add a data migration that creates that default StockLocation. Once executed this should be fixed.

sauloperez:
#### What? Why?

Closes #3253

The data migration creates a StockLocation in the configured default country of the instance with the name "OFN default". Check https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade%3A-Stock-locations
for details.

After running it a rails console should output something like

```
pry(main)> Spree::StockLocation.all
=> [#<Spree::StockLocation id: 5, name: "OFN default", created_at: "2018-12-28 12:16:43", updated_at: "2018-12-28 12:16:43", address1: nil, address2: nil, city: nil, state_id: 58, state_name: nil, country_id: 109, zipcode: nil, phone: nil, active: true, backorderable_default: false, propagate_all_variants: true>]
```

Then, if you create a product you should be able to get that same location as follows

```
pry(main)> variant.stock_items.first.stock_location
=> #<Spree::StockLocation id: 5, name: "OFN default", created_at: "2018-12-28 12:16:43", updated_at: "2018-12-28 12:16:43", address1: nil, address2: nil, city: nil, state_id: 58, state_name: nil, country_id: 109, zipcode: nil, phone: nil, active: true, backorderable_default: false, propagate_all_variants: true>
```

#### What should we test?
It should be possible to create a product from /admin

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -656,10 +656,11 @@ end
 
 FactoryBot.modify do
   factory :stock_location, class: Spree::StockLocation do
-    name { 'default' }
-
     # keeps the test stock_location unique
     initialize_with { DefaultStockLocation.find_or_create }
+
+    # Ensures the name attribute is not assigned after instantiating the default location
+    transient { name 'default' }
   end
 
   factory :shipment, class: Spree::Shipment do
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -656,8 +656,10 @@ end
 
 FactoryBot.modify do
   factory :stock_location, class: Spree::StockLocation do
+    name { 'default' }
+
     # keeps the test stock_location unique
-    initialize_with { Spree::StockLocation.find_or_create_by_name(name)}
+    initialize_with { DefaultStockLocation.find_or_create }
   end
 
   factory :shipment, class: Spree::Shipment do
--- a/app/services/default_stock_location.rb
+++ b/app/services/default_stock_location.rb
@@ -1,0 +1,15 @@
+# Encapsulates the concept of default stock location that OFN has, as explained
+# in https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade%3A-Stock-locations
+class DefaultStockLocation
+  NAME = 'OFN default'.freeze
+
+  def self.create!
+    country = Spree::Country.find_by_iso(ENV['DEFAULT_COUNTRY_CODE'])
+    state = country.states.first
+    Spree::StockLocation.create!(name: NAME, country_id: country.id, state_id: state.id)
+  end
+
+  def self.destroy_all
+    Spree::StockLocation.where(name: NAME).destroy_all
+  end
+end
--- a/app/services/default_stock_location.rb
+++ b/app/services/default_stock_location.rb
@@ -1,7 +1,7 @@
 # Encapsulates the concept of default stock location that OFN has, as explained
 # in https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade%3A-Stock-locations
 class DefaultStockLocation
-  NAME = 'OFN default'.freeze
+  NAME = 'default'.freeze
 
   def self.create!
     country = Spree::Country.find_by_iso(ENV['DEFAULT_COUNTRY_CODE'])
@@ -11,5 +11,9 @@ class DefaultStockLocation
 
   def self.destroy_all
     Spree::StockLocation.where(name: NAME).destroy_all
+  end
+
+  def self.find_or_create
+    Spree::StockLocation.find_or_create_by_name(NAME)
   end
 end
--- a/db/migrate/20180906094641_add_uniqueness_of_variant_id_to_spree_stock_items.rb
+++ b/db/migrate/20180906094641_add_uniqueness_of_variant_id_to_spree_stock_items.rb
@@ -1,3 +1,6 @@
+# Since OFN has only a single default StockLocation, variants in OFN can only
+# have a stock item. By adding this unique index we constraint that at DB level
+# ensuring data integrity.
 class AddUniquenessOfVariantIdToSpreeStockItems < ActiveRecord::Migration
   def change
     add_index :spree_stock_items, :variant_id, unique: true
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -50,3 +50,5 @@ end
 
 spree_user = Spree::User.first
 spree_user && spree_user.confirm!
+
+DefaultStockLocation.create!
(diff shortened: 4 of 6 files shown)


sauloperez:
Tested and working!

db/migrate/20181228115540_add_default_location.rb
```diff
@@ -0,0 +1,11 @@
+class AddDefaultLocation < ActiveRecord::Migration
+  def up
+    country = Spree::Country.find_by_iso(ENV['DEFAULT_COUNTRY_CODE'])
+    state = country.states.first
+    Spree::StockLocation.create!(name: 'OFN default', country_id: country.id, state_id: state.id)
```
mkllnk: Why did you go with `OFN default` instead of `default` as specified in the wiki? I don't really mind. I think we came up with `default`, because it's used by Spree sample data:
https://github.com/openfoodfoundation/spree/blob/f55722b38db7e706a8521c9091a0e00119bb4d20/sample/db/samples/stock.rb#L3

sauloperez: I didn't know we stated its name in the wiki. I just wanted to differentiate from Spree's stuff.

How does that Spree code relate to OFN? AFAIK that's Spree's sample data and there is no default location as per its migrations.

sauloperez:
Nice @mkllnk I had the same feeling while typing that code but with the default country but then I got lazy.

db/migrate/20181228115540_add_default_location.rb
sauloperez: for the record, openfoodfoundation/spree:sample/db/samples/stock.rb@f55722b#L3 is not executed with Spree's seed data (the one that we run when executing `bundle exec rake db:reset`) but from Spree's CLI, which I think is used to set up new Spree projects.

sauloperez:
I also added a constraint at app level to restrict StockLocation to a single record. I couldn't find a way to do it at DB level as well though.

app/models/default_stock_location.rb
```diff
@@ -0,0 +1,15 @@
+# Encapsulates the concept of default stock location that OFN has, as explained
+# in https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade%3A-Stock-locations
+class DefaultStockLocation
```
luisramos0: id prefer a service here, the more we avoid growing /models the better.

app/models/spree/stock_location_decorator.rb
```diff
@@ -1,9 +1,19 @@
 Spree::StockLocation.class_eval do
+  validate :just_default_location
```
luisramos0: i didnt get the "just" quickly. maybe ensure_single_stock_location is easier to understand?

spec/factories.rb
```diff
@@ -645,9 +645,6 @@
 
 FactoryBot.modify do
   factory :stock_location, class: Spree::StockLocation do
-    # keeps the test stock_location unique
-    initialize_with { Spree::StockLocation.find_or_create_by_name(name)}
-
```
luisramos0: are you sure this is not causing build errors?
I see "Validation failed: The default location already exists"

sauloperez: Then we need to fix those because the violate a constraint.

sauloperez:
The deeper I dig, the more complicated it gets... https://github.com/openfoodfoundation/spree/commit/2c48bfbaad

luisramos0:
1.
I am not aware of how are we managing spree migrations... are we including all migrations that spree brings? has that been done for 2.0.4 already?

2.
The link you share shows that spree was already doing what you are doing in this PR.
So, does this PR become unnecessary?

sauloperez:
> 1.

Yep, long ago. AFAIK there was an issue for it.

> 2.

We stepped onto each other. See my last comment.

mkllnk:
@sauloperez Would you like to clean-up the commit history before we merge this? There are a couple of detours.

luisramos0:
This is not ready to be merged.
we need to fix the 2 stock location specs and there is also a problem with spec/models/concerns/variant_stock_spec.rb:155
correct @sauloperez ?

sauloperez:
Yep, and yes. I need to also cleanup the history.

sauloperez:
This is ready now @luisramos0 @mkllnk

luisramos0:
I think this is pending @mkllnk re-review, right?

spec/factories.rb
```diff
@@ -656,8 +656,10 @@
 
 FactoryBot.modify do
   factory :stock_location, class: Spree::StockLocation do
+    name { 'default' }
```
mkllnk: This name doesn't make sense when using "OFN default" via `DefaultStockLocation.find_or_create`.

PR #3255, opened 2018-12-28
Fix menu translations

### LINKED ISSUE #3226: Take menu_*_url from 1 to 4 out of translations
## Description

Since we implemented the customizable main menu links (the ones shown on the home page), we've been hitting the problem of translating URLs over and over again.

The last of many examples is what just happened while deploying latest `fr_BE` translations to OFN Belgium. Theo did his job in Transifex, which part of it was to translate Boutique to Comptoir. I pulled these manually, I checked and by chance I found:

screenshot from 2018-12-18 12-15-15

If I wouldn't notice Theo probably would not have checked it and it'd have failed during the demo. 

I think it's the 7th time I have to warn someone the change will break the menu or I have a problem with this `menu_1_url` keys that are not meant to be translated. One of this cases led to Katuma being broken for almost a day.

So, it's time to fix it. Translators are not aware of this detail and they'll never be. Their are not devs.

## Expected Behavior

`menu_1_url`, `menu_2_url`, `menu_3_url` and `menu_4_url` are not meant to be translated, and as such they should not be in `config/locales` as keys.

## Actual Behavior

People keep translating them and this breaks the links as the app only has the routes `/shops`, `/map`, `/producers` and `/groups` defined. 

## Possible Fix

We should use Rails path helpers instead of translations to render the 4 first menu items in

https://github.com/openfoodfoundation/openfoodnetwork/blob/f36c04542577a75b88ff63fa7ac2219bbba1a767/app/views/shared/menu/_large_menu.html.haml#L14-L20

sauloperez:
#### What? Why?

Closes #3226

This aims to remove the common error that causes instances to point to the wrong menu URLs. It has happened way too many times and leads to people investing time investigating and fixing the problem. This, in the end, is money that could be better invested on other things.

I moved the `menu_*_url` translation keys to ENV vars so no one else will translate them by mistake. This will then require configuring the values of the vars per instance in ofn-install. 

#### What should we test?

AU and UK stagings should have the same custom menu as production. The links should be as the list below.

**AU**
* Shops => /shops
* Map => /map
* Producers => /producers
* Connect => https://openfoodnetwork.org/au/connect/
* Learn => https://openfoodnetwork.org/au/learn/

**UK**
* Shops => /shops
* Map => /map
* Services => https://about.openfoodnetwork.org.uk/services
* About => https://about.openfoodnetwork.org.uk/
* Blog => https://about.openfoodnetwork.org.uk/blog
* Support => https://about.openfoodnetwork.org.uk/support

#### Release notes
Replaced main menu's URL translations with ENV vars to avoid translating non-existent URLs.

Changelog Category: Changed


sauloperez:
Marking as WIP as, although it works, I still need to check if there's anything missing, besides the PR in ofn-install to migrate current instance's configuration.

sauloperez:
Let me know if you think we can simplify this further.

sauloperez:
> What we really want is the multilang feature for content and store it in the database.

But let's not build WordPress in Ruby :troll:

sauloperez:
@mkllnk the idea here is to take out those 4 URLs that are not translatable. For instance, shop's URL won't change between locales so I think no i18n feature or similar is needed here. Correct me if I'm wrong.

sauloperez:
I appreciate your feedback. That is why I opened as WIP with a proposal.

>  I am not really sure what's a good alternative apart from keeping things as they are.

I'm sure I took too long to raise the flag but I have spent way too many hours troubleshooting instances while I could have been bringing more value. And this still takes me time. So, IMO keeping things as they are is not an option.

I want everyone to understand I'm not being picky. I'm just trying to improve things.

> the cool functionality of having a configurable menu.

Did we add it because it was cool or was it a requirement? Since the beginning, I wondered if having entirely customizable was really the need. At least all instances show shops and the map. Can't we give room to extend the menu instead?

I'd appreciate we all could take some time to read the issue #3226 and think of ways to solve this.

sauloperez:
I've been thinking an the root cause is using things that are not valid URLs like `/tiendas` or `/Comptoirs`. Then, what if we could automatically know if a particular set of translations adds a URL that doesn't exist?

There might be better ways but that could be checked with a small script as if it was a linter that we run in CI. It would check if the entered URL returns a 200 ok. The examples I mention would not, resulting in a broken build from a Transifex branch.

mkllnk:
You know what the real solution would be? Storing the menu entries in the database and use the [https://github.com/globalize/globalize](globalize gem) to translate them to all languages. This is the perfect first example for the globalize gem and then we can extend the use to product names etc. That would actually be work towards a long-term solution.

luisramos0:
yes @mkllnk agree, it's a good first case but multilanguage content is not a priority right now:
https://community.openfoodnetwork.org/t/current-top-10-global-priorities/1433

daniellemoorhead:
He @sauloperez this has been sitting in `In Dev` for a long time with no progress. Are you working on it in the next week or should it go back to `Dev Ready` till you have time to pick it up again?

#toottoot

daniellemoorhead:
It's been over 3 weeks now @sauloperez and no change. Given your time constraints at the moment it seems better to move this back to the backlog, no, and pick it up when you have time again.

sauloperez:
I'll follow your advice. It's clear again that we should work on it next.

sauloperez:
I gave it a second iteration and encapsulated all the logic into this new `MenuItem` abstraction. Now it is way easier to understand than before. Please, re-review.

config/application.yml.example
```diff
@@ -62,3 +62,6 @@ SMTP_PASSWORD: 'f00d'
 # STRIPE_INSTANCE_PUBLISHABLE_KEY: "pk_test_xxxx" # This can be a test key or a live key
 # STRIPE_CLIENT_ID: "ca_xxxx" # This can be a development ID or a production ID
 # STRIPE_ENDPOINT_SECRET: "whsec_xxxx"
+
+# URLs for the site's main menu
+# MENU_3_URL: https://about.openfoodnetwork.org.uk/services
```
luisramos0: is this mandatory? what happens if this is not configured?

app/models/menu_item.rb
```diff
@@ -0,0 +1,43 @@
+class MenuItem
```
luisramos0: what a wonderful example of what I am talking about [here](https://community.openfoodnetwork.org/t/how-to-name-a-new-service-in-ofn/1530/7?u=luisramos0). option 6, non-ORM data objects like this should be in /services. It's funny how this is very obvious to my large scale java/c# service oriented conditioned brain but goes against MVC and "normal Rails". I think we should still follow it :-)
Basically, I advocate that /models is for ORM only.

luisramos0: Spree is a very good example of a mess of ORM and logic mixed together in /models, with for example shipments (db entity) mixed with packages (service entity).

mkllnk: It's not that obvious to me. It represents the data of a menu item. It doesn't fetch data from the database but from the routes, the locale and the environment. You could separate the logic into a stupid MenuItem, a pure data structure, and a MenuItemFactory which is a service creating those items. But having it in one place is nice as well. It's not a big class and follows the OO pattern of encapsulation.

I think it shouldn't matter where data is stored. Not every model has to be an ActiveRecord object.

luisramos0: I dont think we need to split between service and the business model. The point is that MenuItem is a business model with business logic (the mentioned option 6) and belongs in the service layer, never in the data access layer.

"I think it shouldn't matter where data is stored."
This is exactly where I disagree. I think this it the open door to the big mess we (and spree) have under /models. And if we dont start to move things to services (and organize services well) we will keep having a bigger mess under /models. It wont stop growing and it's already over 100 files.

luisramos0: A well organized app will look simple (very different from what we have). It will have only ActiveRecord  objects under /models and then in /services you will find a well separated list of service modules under different folders, each addressing specific parts of the business logic. This is where the complexity of the app will be gracefully addressed far from active record relations and callbacks. I have seen this in many apps and it makes a huge difference as the app grows in complexity.

We have some serious design errors in spree/ofn, for example, the logic to address adjustments (or the logic in order updater) has been put into /models and it should be 100% outside of it under /services and completely detached from callbacks and AR relations.

luisramos0: so my recommendation is to put this file as it is under engines/web/app/services and not together with the other +100 files under /models

config/application.yml.example
sauloperez: Screenshot from 2019-04-05 13-03-04

I let RSpec answer you :sunglasses: I'll open a PR on ofn-install to add the required ENV vars for each instance once I get your approvals.

app/models/menu_item.rb
sauloperez: Ok, my 2 cents here.

IMO this is a view-related abstraction and at first, I wanted to add it to `app/views` or perhaps in a new `app/presenters/` then I thought, "they won't like it and what I want to be addressed quickly will take longer. I don't care about the actual folder but the code changeability. Change as less a possible and hope for the best".

Then, if you ask me, as for my experience and the code conventions I came to use over time, aligned with the Rails community (I think), `models/` does not only contain AR objects but domain-related models not backed by a DB table. This grows over time as the team and the app mature. `services/` is used for command-like, purpose-specific classes; they perform an action. Things get larger and that is when folders like `queries/` and `presenters/` come up. They all benefit from the Rails autoloading: any Ruby code within `app/*/` will get evaluated and ready to be used.

I was quite strict about these things when I was younger but I came to realize that because of autoloading it doesn't matter much where you put things (and be happier :joy: ). It is the right OO abstraction and the overall code quality what makes an app changeable. It might even matter less for me as I use fuzzy finder and ctags. 

I also have the feeling that I waste my time in these discussions I want to rather focus on providing value. So, all the explanation is to conclude that I'm absolutely happy with anything you guys decide. I'm fine with any arbitrary decision we come up with as a team :muscle: .

As a side note, however, the mess with AR callbacks and logic you mention @luisramos0 is something all Rails app tend to suffer after sometime due to the way Rails couples it all in ActiveModel objects. It makes it really easy at first but requires OO discipline to overcome it when your app becomes something else than an MVP. Discipline teams need to learn, which takes time. IMO where you put the mess won't change things, either `models/`, `services/` or anywhere else.

sauloperez: So, to keep things agile, I'll go and move it to `services/` as that is what we've been doing lately.

sauloperez:
> This should avoid the most dangerous mistakes translators can make. I still think we need to find a better solution to the menu feature. Drawing data from three different sources and having different rules depending on the menu item number is very complex and will lead to confusion.

I have the same feeling. It was even a bit complex for me and I did my best as 1st iteration. The goal here is to spend our time on meaningful things and be happier. I've spent countless hours checking instances, finding out about the bug, fixing things manually, then preparing custom releases, and repeating the process again each time a translator makes the mistake. Not to mention the impact this has on our budget... the ultimate metric :trollface: 

So I think the goal is accomplished. Our life will be better after it :joy:

RachL:
On staging I could only tests those: 

Shops => /shops
Map => /map

The other links are production only. As the bug happened on the "shop" page mostly. I'm putting this to ready to go.

app/models/menu_item.rb
luisramos0: I agree, let's not waste our time in long discussions, I raised this point because I think this topic is very important. imo MVC thinking is poison at this stage and we need to focus on service oriented design separated from the data access layer.

sauloperez:
What about producers and groups? they are not retrieved from the translations file either but they should still point to /producers and /groups by default.

RachL:
@sauloperez those two menus are not used on UK data / staging

luisramos0:
you can activate them here: https://staging.openfoodnetwork.org.uk/admin/content/edit

RachL:
@luisramos0 @sauloperez good thing you made me retest :scream:

/producers is indeed working but I'm not sure it should. Indeed, the admin section in staging UK is not providing the "producer" menu. You can activate a "pricing" menu which URL is https://about.openfoodnetwork.org.uk/pricing-and-plans

But this URL is not seen anywhere in the staging shopfront. In production "pricing" menu leads correctly to https://about.openfoodnetwork.org.uk/pricing-and-plans/ (the url that is in the translations).

We are here in an edge case I think because UK is maybe the only instance not using the "basic" producer menu and I guess your development was made taking into account everybody does @sauloperez ?

Merging this will break UK's production menu. So I can't put it in ready to go. Moving this back to in dev. ping @lin-d-hop 

https://staging.openfoodnetwork.org.uk/groups is working fine.

luisramos0:
ok, first, we need to adapt en_GB to this change by making menu 1 to 4 translations match the static options: shops, maps, producers and groups. I just checked the file, it's really just changing key menu_4_title from Pricing back to Producers.
That is all fine and in UK we can then disable menu_3 and menu_4 (producers and groups) so that the menu starts as is with shops and maps only.

The only problem I see is that currently in UK, on top of these 4 slots, we have 4 more entries: PRICING, ABOUT, BLOG and SUPPORT and with this PR we only support 3 configurable options :-(

We can add one more configurable item and make them 8 (some dev required), or we need to remove one of the 4 entries in UK. Maybe we keep only blog or about, is that acceptable?

lin-d-hop:
Grrr. I didnt read all that closely but everything about this issue is frustrating, no? We had a good (albeit hacky) solution that just wasn't idiot proof enough and that meant that Pau wasted loads of time. Lessons learned.

For UK if you *need* to get rid of 'Blog' that is acceptable. Please do not replace or change any other menu item.

luisramos0:
ok, if we get rid of blog, this is what needs to happen in transifex en_GB:

current:
  menu_1_title: "Shops"
  menu_1_url: "/shops"
  menu_2_title: "Map"
  menu_2_url: "/map"
  menu_3_title: "Pricing"
  menu_3_url: "https://about.openfoodnetwork.org.uk/pricing-and-plans"
  menu_4_title: "Groups"
  menu_4_url: "/groups"
  menu_5_title: "About"
  menu_5_url: "https://about.openfoodnetwork.org.uk"
  menu_6_title: "Blog"
  menu_6_url: "https://about.openfoodnetwork.org.uk/blog"
  menu_7_title: "Support"
  menu_7_url: "https://about.openfoodnetwork.org.uk/support"

new:
  menu_1_title: "Shops"
  menu_2_title: "Map"
  menu_3_title: "Producers"
  menu_4_title: "Groups"
  menu_5_title: "Pricing"
  menu_5_url: "https://about.openfoodnetwork.org.uk/pricing-and-plans"
  menu_6_title: "About"
  menu_6_url: "https://about.openfoodnetwork.org.uk"
  menu_7_title: "Support"
  menu_7_url: "https://about.openfoodnetwork.org.uk/support"

and then we disable menu 3 and 4 after deploy.

I made exactly this change manually in en_GB.yml staging UK now, can you please have a look and validate it's correct?
Screenshot 2019-04-09 at 09 15 15
if it is correct, we can then do it in transifex.

@sauloperez I just found one problem in the backoffice content main links where the urls 1 to 4 become missing translations:
Screenshot 2019-04-09 at 09 10 03

This needs to be fixed.

sauloperez:
Thank you very much for your awesome investigation @luisramos0 ❤.

I will definitely go one by one with each instance's locale on my dev machine and let you know when it is ready again to be tested. I'll take the chance to document as well because it still feels too cumbersome 😕

luisramos0:
hey, instead of doing what I did above for uk, we can keep the menu uk as is with blog by changing menu 3 from Producers to Pricing with env vars...

luisramos0:
hey @sauloperez for the content config side, you just need to adapt this line of code in some way:
https://github.com/openfoodfoundation/openfoodnetwork/blob/6e727e12e2dfa3c83d919aeea5416644dc0ed1a6/app/views/admin/contents/_fieldset.html.haml#L6
for example:
```
      - menu_item = MenuItem.new(key.slice! "menu_")
      - text = menu_item.title + " - " + menu_item.url
```

sauloperez:
Ok, I'll take a look. The fewer changes the better.

sauloperez:
This has fallen again behind other priorities 😭 I really wish we could have it for v2 but... we better get Katuma server provisioned and monitored first and hope no one messes with these translations any time soon 🤞 .

sauloperez:
I hate having to close this because it's one of those bits that annoy me the most but it's sitting here for ages and we got other priorities. Bringing it into production involves some risk and requires time :man_shrugging: .

However, we those new languages coming in, we should be aware of this issue.

sauloperez:
I feel in peace now :trollface:

PR #3256, opened 2018-12-28
Replace feature spec with view spec

sauloperez:
#### What? Why?

There is not need to boot a browser and hit the whole stack to check how we display information in a template.

Furthermore, relying on feature specs by default not only leads to long-running and fragile test suites but also makes it impossible to have a proper testing pyramid.

While the former feature spec took 0.52272 seconds to finish, the new view spec took 0.39279 seconds. Nearly a 25% gain.

#### What should we test?
Nothing. This only affects the test suite.

#### Release notes
Replace feature spec with a more appropriate view spec to improve performance.

Changelog Category: Changed


sauloperez:
That's exactly the debate I wanted to raise. We can't afford to have so many feature specs, more than other kinds of test.

That's a very valid point what you suggest about the controller spec though.

sauloperez:
> Imagine someone renames or changes the Enterprise.is_hub scope. It will break the controller, but this test will still pass, because is_hub is mocked.

If we unit tested the scope, and we change its implementation, the test will fail. If you stop calling `is_hub` from the controller its test will fail. See below.

Changing `@enterprises` to the following

```ruby
@enterprises = Enterprise.is_distributor
```
the tests fail with

```ruby
  1) SitemapController#index lists all hub enterprises
     Failure/Error: expect(assigns(:enterprises)).to eq([hub])

       expected: [#<Enterprise id: nil, name: "Enterprise 1", description: "enterprise", long_description: "<p>Hello, ...ce_text: nil, display_invoice_logo: false, allow_order_changes: false, enable_subscriptions: false>]
            got: []
```

Here I made a choice. I sticked to unit testing the controller because I think an integration test wouldn't cover much more IMO.

sauloperez:
So what do you suggest? it's not clear to me if I should change something or if there's any agreement around the various test types.

mkllnk:
> >   Imagine someone renames or changes the Enterprise.is_hub scope. It will break the controller, but this test will still pass, because is_hub is mocked.
> If we unit tested the scope, and we change its implementation, the test will fail.

The spec of the scope will fail. And then you fix the spec of the scope. But you forget to change this controller and its spec. It still passes. We would get reports through Bugsnag, I guess.

I'm Sorry Pau. This was such a good idea and now we are pointing out all the downsides of this change. Initially we talked about a difference of 0.12993 seconds. Now we have a second spec already which may take longer than that. I like the underlying principle of having more unit tests and fewer feature specs. But in this case, I would suggest to leave it.

sauloperez:
>  I like the underlying principle of having more unit tests and fewer feature specs. 

That's what I wanted to point out. Our current feature-spec-by-default approach doesn't scale. I'm happy to see this trend is slowly changing though.

PR #3257, opened 2018-12-28
Refactor to reveal how we load enterprises in shop

sauloperez:
#### What? Why?

By sticking to Rails conventions we make it more obvious how (badly) we are loading enterprises in EnterprisesController#shop and shed some light on the obscure InjectionHelper.

This will also make it easier to improve its performance in the future as it's among the top offenders. See https://www.skylight.io/app/applications/ibo3NOqCYMnq/1545851520/1d/endpoints/EnterprisesController%23shop?responseType=html

This is just a tiny refactor, so at least we move things forward a bit. What I've found so far is that there is not a single N+1 that is not triggered by the InjectionHelper module.

#### What should we test?

The frontoffice's shop page should work as usual. You can access by clicking on a shop's link from /shops.

#### Release notes

Move the enterprises' load from the injection helper to the controller.

Changelog Category: Changed


app/controllers/enterprises_controller.rb
```diff
@@ -88,4 +108,14 @@ def reset_order_cycle(order, distributor)
...
+
+  def inject_json_ams(name, data)
+    options = {
+      each_serializer: Api::EnterpriseSerializer,
+      data: OpenFoodNetwork::EnterpriseInjectionData.new
+    }
+    serializer_instance = ActiveModel::ArraySerializer.new(data, options)
+
+    { name: name, json: serializer_instance.to_json }
+  end
```
sauloperez: I prefer to have this duplicate for now and then abstract things than having it buried in InjectionHelper where it's hard to understand.

app/controllers/enterprises_controller.rb
```diff
@@ -7,13 +7,33 @@ class EnterprisesController < BaseController
...
+                     [current_order_cycle]
+                   else
+                     OrderCycle.not_closed.with_distributor(current_distributor)
+                   end
+
+    enterprises = current_distributor
+      .plus_relatives_and_oc_producers(order_cycles)
+      .activated
+      .includes(address: :state)
+      .all
```
sauloperez: These two AR relations are just copied from InjectionHelper.

sauloperez:
@mkllnk it's not quite clear to me how I'm supposed to update `.rubocop_manual_todo.yml`. Do I have to update it manually cop by cop??

sauloperez:
Fixed. Thanks @mkllnk

sauloperez:
What I actually wanted to move out of the injection helper and be triggered from the controller is the data fetching (it can be extracted from the controller itself of course).

As a side effect, and since we need to pass arguments to the views, either as ivars or locals, the serialization went there as well but this is not final. It's just a first step towards improving performance and design. Now it's ugly and duplicate but I'll revisit it soon as I work more on individual shop's performance.

RachL:
_ Open and closed shop are correctly shown
- Filters are still working

This is good to go!

PR #3258, opened 2018-12-30
Serve Google maps JS lib from a proxy

### LINKED ISSUE #3096: [Spree Upgrade] Google Maps "NoApiKeys" warning in feature specs

sauloperez:
#### What? Why?

:warning: **disclaimer** this still totally WIP

Closes #3096

This requires us to install `sinatra` manually for now. You can do so by running `gem install sinatra`.

Before running any test or running the app in your dev env, boot this fake google maps server by running `ruby spec/fake_google_maps.rb`. This boots up a WEBrick process in foreground which has the /maps route configured.

Then, for the app to point to this server go add the following line in config/application.yml.

    GOOGLEAPIS_URL: '//localhost:4567/maps'

Now you can run the tests or the app normally. For instance, when executing spec/features/consumer/shopping/checkout_spec.rb:54 you can see the following in WEBrick's output.

    http://127.0.0.1:40921/cart -> /maps?libraries=places,geometry

This way we don't hit maps.googleapis.com while testing.

#### Missing

- [x] Populate GOOGLEAPIS_URL in test and set current value in prod
- [ ] Add sinatra in Gemfile so it works in CI
- [x] Boot fake_google_maps programmatically from spec so it works in CI
- [ ] Refactor to use just a simple rack app and not depend on Sinatra

#### Release notes


sauloperez:
> Was this your own idea or did you see that somewhere?

I grabbed some inspiration from https://github.com/thoughtbot/fake_stripe.

> Do we need Sinatra or can we place the JS file in public/ and use the test server started by rspec (aka localhost) as maps server?

Yep, I want to avoid Sinatra but I don't follow what you say about public and RSpec. AFAIK there is no way we can tell Capybara (who actually starts a server I think) to also serve this JS file.

> Why can't we use puffing billy? oesmith/puffing-billy

I started with Puffing Billy but it's not a fire and forget. By default, it blocks all HTTP requests and apparently, we have some. Also, it felt like adding a new big dependency that required me to change specs configuration wasn't really worth for just a JS file. I want to revisit it though and tune its config to just block this particular request. We'll see what's best.

mkllnk:
> Yep, I want to avoid Sinatra but I don't follow what you say about public and RSpec. AFAIK there is no way we can tell Capybara (who actually starts a server I think) to also serve this JS file.

When we start a feature spec, we automatically get a webserver answering all the web requests from the test browser. I was just wondering what would happen if we served the JS file through our normal application. For example, if you placed in in `public/maps.js`, you could just load it with a relative path: 
```
%script{src: "/maps.js"}
```

sauloperez:
I wouldn't implement something that is only meant for testing. It is not a feature that we want to add but a test double and as such it should stay within the test env limits IMO.

sauloperez:
> Running another server just to serve this response (and others in the future) is interesting but feels like adding unnecessary complexity.

I understand your concerns but I wouldn't worry too much. If what you care about is performance, it's way faster booting a process or a thread on your machine than issue an HTTP request to a remote server. And Puffing Billy isn't magic either.

> Billy spawns an EventMachine-based proxy server, which it uses to intercept requests sent by your browser. It has a simple API for configuring which requests need stubbing and what they should return.

mkllnk:
@sauloperez Here is another approach already used in our JS specs: https://github.com/mkllnk/openfoodnetwork/commit/727a4139715dcfe924f0f8dac747cf02ffb5e13d

Much simpler.

sauloperez:
but do we want to pollute our production's JS bundle with something that is only needed in testing? I'm not sure we can put an if either.

mkllnk:
No, we don't want to have that in production. I'm sure there is a way with an `if`. I just wanted to give you another idea how to solve this in a simpler way. We could just put the compiled mock js file in `public/` and use your approach of a variable to either load the mock or the real maps.

daniellemoorhead:
Also this one @sauloperez - if you don't have any plans to work on this in the next week then can you move it back to `Dev Ready` till you are free to pick it up again? It's been almost 6 weeks since the last activity on it.

#toottoot

sauloperez:
Closing until I manage to get some time for it. Thanks @daniellemoorhead !

PR #3273, opened 2019-01-02
Fix menu_1_url translation for fr_BE

sauloperez:
#### What? Why?

The first menu item in OFN BE is currently broken.


sauloperez:
This is what I deployed to OFN BE yesterday. As soon as we fix it in Transifex and we pull the changes and will be solved for next releases.

PR #3277, opened 2019-01-03
Revert "[Spree Upgrade] Set on_demand default value to false in test factories"

luisramos0:
PR 3246 got the 2-0-stable build very broken: https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/2-0-stable/builds/158
We should revert this small change and have a look afterwards.
Reverts openfoodfoundation/openfoodnetwork#3246

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -647,9 +647,6 @@ FactoryBot.modify do
   factory :stock_location, class: Spree::StockLocation do
     # keeps the test stock_location unique
     initialize_with { Spree::StockLocation.find_or_create_by_name(name)}
-
-    # sets the default value for variant.on_demand
-    backorderable_default false
   end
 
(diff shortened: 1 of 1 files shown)


sauloperez:
I wanted to do the same as soon as I found out but then I saw things get fixed by

luisramos0:
@sauloperez I am not sure you agree or not with this revert. Can you confirm. I think we should do this immediately so that the 2-0 build is back to "green" (24 broken specs).

luisramos0:
I am sorry but I am bypassing the second review here. Because it's a single line of code, because the build is broken without this and because this is blocking the merge from master #exceptional

luisramos0:
ok, looking good. we are back to 23 broken tests in 2-0-stable.

PR #3282, opened 2019-01-04
[Spree Upgrade] Fix credit card spec that was depending on current year

luisramos0:
With this commit https://github.com/openfoodfoundation/spree/commit/5d93d1dedb0c727c9a78770d35a4363a16fa1691, this static fix was meant to fail https://github.com/openfoodfoundation/openfoodnetwork/commit/3f33e28839cc60dd28175becad8ed590bc51bb6e

#### What? Why?

Fixes spec/serializers/credit_card_serializer_spec.rb:13

#### What should we test?
Spec should be green. I confirmed in the build that it is green.

### DIFF
--- a/spec/serializers/credit_card_serializer_spec.rb
+++ b/spec/serializers/credit_card_serializer_spec.rb
@@ -10,6 +10,6 @@ describe Api::CreditCardSerializer do
   end
 
   it "formats an identifying string with the card number masked" do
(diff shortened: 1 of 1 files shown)


luisramos0:
rebased to get latest build. all good here.
we have one unstable spec in the 2-0 build: spec/features/consumer/shopping/variant_overrides_spec.rb:154:

PR #3295, opened 2019-01-08
Update all locales with the latest Transifex translations

sauloperez:
#### What? Why?

Bringing in the latest translations with `tx pull --force` in preparation for the release v1.25.0.


sauloperez:
I checked an there are no menu_* translations broken.

Issue #3291, opened 2019-01-07
[i18n] wrong automatic translation for "state" on order page


myriamboure:
the column header "state" here in English platform:
![untitled presentation](https://user-images.githubusercontent.com/8135376/50770425-94551680-1287-11e9-9fc0-b30e2790b52b.jpg)

is badly automatically translated by Spree and no string exist in Transifex to fix it:
![untitled presentation 1](https://user-images.githubusercontent.com/8135376/50770423-8ef7cc00-1287-11e9-92b1-d77c20ae1ac4.jpg)

myriamboure:
Same @HugsDaniel this one is pretty annoying as really doesn't make sense... if you can fix it as well it would be great.

PR #3304, opened 2019-01-10
Add I18n keys for order state and order firstname and lastname filters

### LINKED ISSUE #3288: [i18n] Missing string and wrong automatic translation on orders page
On /admin/orders page, the filter "first name begins with" is not in transifex and cannot be translated.
In French the translation is so wrong we are a bit ashamed... "begins", translated "commence" in French is currently written "commmance"... 

We need the string in transifex to be able to fix that.

English:
untitled presentation

French:
untitled presentation 1

HugsDaniel:
#### What? Why?

Closes #3288 #3291 

Add I18n keys for order state and firstname/lastname order filters in our codebase.

#### What should we test?

After translations are added in Transifex, the labels should be correctly translated.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -30,6 +30,7 @@ en:
         shipment_state: Shipment State
         completed_at: Completed At
         number: Number
+        state: State
         email: Customer E-Mail
       spree/payment:
         amount: Amount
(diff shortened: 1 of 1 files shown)


RachL:
Everything looks fine :
image

PR #3305, opened 2019-01-10
Improve shops performance take 2

sauloperez:
#### What? Why?

Moves #3099 forward.

This is most of the stuff I had on my stash since last time I worked on #3099. It's just part of the refactoring needed to remove another one of the various N+1s that take place in `GET /shops`.

This is aligned with what I did in #3151, where an N+1 is removed by moving the data loading to the controller, or #3257 where moving the data loading brings some clarity.

Unlike #3151, here loading product and producer properties in a single query turns to be tougher. I had to stop at `#active` and give it some more thought. There's more on the backstage, but this is what it's shippable for now.

There won't be much performance improvement if any.

#### What should we test?

The distributed properties shown in `/shops` should keep working as expected. That is, those that belong to producers or products that belong or belonged to an order cycle.

#### Release notes

Refactor CachedEnterpriseSerializer's `#distributed_properties` to make it easier to remove an N+1 on the properties table in the future.

Changelog Category: Changed

### DIFF
--- a/app/serializers/api/cached_enterprise_serializer.rb
+++ b/app/serializers/api/cached_enterprise_serializer.rb
@@ -7,7 +7,7 @@ module Api
     cached
 
     def cache_key
-      object.andand.cache_key
+      enterprise.andand.cache_key
     end
 
     attributes :name, :id, :description, :latitude, :longitude,
@@ -24,55 +24,55 @@ module Api
     has_many :distributed_properties, serializer: PropertySerializer
 
     def pickup
-      services = data.shipping_method_services[object.id]
+      services = data.shipping_method_services[enterprise.id]
       services ? services[:pickup] : false
     end
 
     def delivery
-      services = data.shipping_method_services[object.id]
+      services = data.shipping_method_services[enterprise.id]
       services ? services[:delivery] : false
     end
 
     def email_address
-      object.email_address.to_s.reverse
+      enterprise.email_address.to_s.reverse
     end
 
     def hash
-      object.to_param
+      enterprise.to_param
     end
 
     def logo
-      object.logo(:medium) if object.logo?
+      enterprise.logo(:medium) if enterprise.logo?
     end
 
     def promo_image
-      object.promo_image(:large) if object.promo_image?
+      enterprise.promo_image(:large) if enterprise.promo_image?
     end
 
     def path
-      enterprise_shop_path(object)
+      enterprise_shop_path(enterprise)
     end
 
     def producers
-      relatives = data.relatives[object.id]
+      relatives = data.relatives[enterprise.id]
       ids_to_objs(relatives.andand[:producers])
     end
 
     def hubs
-      relatives = data.relatives[object.id]
+      relatives = data.relatives[enterprise.id]
       ids_to_objs(relatives.andand[:distributors])
     end
 
     def taxons
       if active
-        ids_to_objs data.current_distributed_taxons[object.id]
+        ids_to_objs data.current_distributed_taxons[enterprise.id]
       else
-        ids_to_objs data.all_distributed_taxons[object.id]
+        ids_to_objs data.all_distributed_taxons[enterprise.id]
       end
     end
 
     def supplied_taxons
-      ids_to_objs data.supplied_taxons[object.id]
+      ids_to_objs data.supplied_taxons[enterprise.id]
     end
 
     def supplied_properties
@@ -97,7 +97,7 @@ module Api
     end
 
     def active
-      data.active_distributors.andand.include? object
+      data.active_distributors.andand.include? enterprise
     end
 
     # Map svg icons.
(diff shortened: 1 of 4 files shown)


app/serializers/api/property_serializer.rb
```diff
@@ -1,4 +1,7 @@
 class Api::PropertySerializer < ActiveModel::Serializer
+  # TODO: stick to :presentation or :name not both because the are duplicate
```
luisramos0: imo, code should never ever have TODOs. TODOs belong in backlogs. So I think this should be either 1. a comment about how the code works ":presentation or :name are duplicates and could be one property only" or 2. an issue in GH

app/serializers/api/cached_enterprise_serializer.rb
```diff
@@ -81,25 +81,42 @@ def supplied_properties
...
-
+        Spree::Property.currently_sold_by(enterprise)
       else
-        product_properties = Spree::Property.ever_sold_by(object)
-        producer_property_ids = ProducerProperty.ever_sold_by(object).pluck(:property_id)
+        Spree::Property
+          .joins(products: { variants: { exchanges: :order_cycle } })
+          .merge(Exchange.outgoing)
+          .merge(Exchange.to_enterprise(enterprise))
+          .select('DISTINCT spree_properties.*')
```
luisramos0: I dont get this part. you just moved the code from Spree::Property.ever_sold_by to here, right?

sauloperez: Yep, that's it. Essentially, `#distributed_properties` issues few queries to the properties table but this makes things more complex than they should be. It's simpler to just ask the DB about properties that we care for than having to deal with `ProducerProperty` on one hand, `product_properties` on the other, then mapping their presentation attribute, etc. We tend to forget that SQL is declarative and end up implementing map-reduce.

A first step towards relying on SQL is to inline the queries to find how they can be merged together. Does it make sense?

app/serializers/api/property_serializer.rb
sauloperez: Ups, sorry my mistake. I left it there to turn it into something else. I share the opinion about TODOs

sauloperez: I created https://github.com/openfoodfoundation/openfoodnetwork/issues/3313

app/serializers/api/cached_enterprise_serializer.rb
luisramos0: yes it makes sense. but I wouldnt move/duplicate the SQL here if I was not changing it.

spec/controllers/shops_controller_spec.rb
```diff
@@ -15,4 +15,25 @@
...
     expect(response.body).to have_content(invisible_distributor.name)
   end
+
+  it 'renders distributed properties' do
+    duplicate_property = create(:property, presentation: 'dairy')
+    producer = create(:supplier_enterprise, properties: [duplicate_property])
+    property = create(:property, presentation: 'dairy')
+    product = create(:product, properties: [property])
+    producer.supplied_products << product
+
```
kristinalim: If you want to doing a smoke test here, maybe it is more appropriate to test that producer and product properties are covered, than checking that there are no duplicates.

luisramos0:
@sauloperez moving to "in dev", pls move to code review when you think is ready ;-)

spec/controllers/shops_controller_spec.rb
sauloperez: The absence of duplicates is already covered in the serializer specs (and it does work) but I added individual tests for product and producer properties in the controller as well.

sauloperez:
I believe I fixed all comments and it's ready to be tested. Agree @luisramos0 and @kristinalim ?

spec/serializers/api/cached_enterprise_serializer_spec.rb
```diff
@@ -27,28 +27,26 @@
...
-      order_cycle.exchanges << incoming_exchange
-      order_cycle.exchanges << outgoing_exchange
-
-      order_cycle.save
+      create(
+        :simple_order_cycle,
+        coordinator: shop,
+        suppliers: [producer],
+        distributors: [shop],
+        variants: [product.variants]
```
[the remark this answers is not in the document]
kristinalim: @sauloperez This seems to still work anyway, but shouldn't this be `variants: product.variants`? Minor change needed.

kristinalim:
@sauloperez From the PR description, this doesn't close the issue yet. I disconnected this from the issue on Zenhub, and then moved to Test Ready. :-)

RachL:
Properties are looking good. Moving to ready to go https://docs.google.com/document/d/1fBnxEQiuA3FOXMmUJbX04F35NhMnL4XQjFLo21gJEWo/edit#heading=h.hj26umg3kyxb

PR #3315, opened 2019-01-11
[Spree Upgrade] Phase 2 - Product Import ready - all specs running and green

### LINKED ISSUE #2791: [Spree Upgrade][PI] Fix Product Import specs (1 broken spec)
spec/models/product_importer_spec and spec/features/admin/product_import_spec are badly broken.
These specs were made pending but need to be fixed at some point.

Move all specs in products_reset_strategy_spec to pending.

See #2796  and #2902 

Status update: 36 broken specs:
- JOB 2 - 6 broken specs - products_reset_strategy_spec
   -- error: You need to save the variant to create a stock item before you can set stock levels.
- Job 3 - 5 x spec/features/admin/product_import_spec
   -- UX tests failing
- Job 3 - 25 x spec/models/product_importer_spec
   -- error: undefined method `-' for "100":String
   -- also, shipping_category is a mandatory product attribute

luisramos0:
#### What? Why?

Closes #2791

After #3425, product import is working in v2. Ready for testing in v2.

#### What should we test?
Product Import feature in v2.

### DIFF
--- a/spec/features/admin/product_import_spec.rb
+++ b/spec/features/admin/product_import_spec.rb
@@ -34,7 +34,7 @@ feature "Product Import", js: true do
     before { quick_login_as_admin }
     after { File.delete('/tmp/test.csv') }
 
-    xit "validates entries and saves them if they are all valid and allows viewing new items in Bulk Products" do
+    it "validates entries and saves them if they are all valid and allows viewing new items in Bulk Products" do
       csv_data = CSV.generate do |csv|
         csv << ["name", "producer", "category", "on_hand", "price", "units", "unit_type", "shipping_category_id"]
         csv << ["Carrots", "User Enterprise", "Vegetables", "5", "3.20", "500", "g", shipping_category_id_str]
@@ -101,7 +101,7 @@ feature "Product Import", js: true do
       expect(page).to have_no_selector 'input[type=submit][value="Save"]'
     end
 
-    xit "handles saving of named tax and shipping categories" do
+    it "handles saving of named tax and shipping categories" do
       csv_data = CSV.generate do |csv|
         csv << ["name", "producer", "category", "on_hand", "price", "units", "unit_type", "tax_category", "shipping_category"]
         csv << ["Carrots", "User Enterprise", "Vegetables", "5", "3.20", "500", "g", tax_category.name, shipping_category.name]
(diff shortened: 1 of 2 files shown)


luisramos0:
ping @RachL :-D

RachL:
@luisramos0 does this means also that from now on all product import bugs should be taking care of in v2 rather than v1?

luisramos0:
I'd keep PI bugfixes in master. PI is not like subscriptions, it's lot simpler.

the real challenge here are the "merge to 2-0" PRs like this https://github.com/openfoodfoundation/openfoodnetwork/pull/3610
**There's this point where we need to test all PRs in both versions!!!**

For example, sample data is fine in master but broken in v2 (I'll reate a PR today to fix it), but what about all other master PRs in #3610?

RachL:
> There's this point where we need to test all PRs in both versions...

RachL:
@luisramos0  I'm definitely doing something wrong here, any idea? Those are the 3 tests so far:

1. Product creation on existing enterprise : got a confusing error message saying something cannot be blank, but I have the feeling I've completed everything (sku and available_on columns are empty, but doing test number 3 I've tried adding those columns empty and it didn't change anything). If this is about unit type being empty, we are here in the case of items products, so we should be able to leave the column blank (so does the user guide says).
2. Create inventories : I tried to add existing products to an inventory. But the products were not found. I thought maybe I didn't write them well, so I added columns to update those products with a product import, which lead then to test number 3:
3. Updating products: in this case both product were found in the database (although with inventory before they were not) but the description is not matching the actual description although I copy pasted it... So I'm stuck => file product_listv2_test2

All files are here: https://drive.google.com/drive/u/0/folders/1fI0r3CuHEs0e2y3N46QHIUPJkMs4mfrf
Notes: https://docs.google.com/document/d/1g-dgFCLc9I4v667XcOFgsfYSccmCQLNMtsyxoH3xrkU/edit

luisramos0:
@RachL you are not doing anything wrong! You are testing, and it's broken...

point 1 - shipping category is now mandatory but the error is not showing correctly, it's "cant be blank", should be "shipping category cant be blank". I have added shipping category to your long file and the products are created. Can you please create a separate issue to fix the error message? the user guide will need to be updated as well, right?

point 2 - this is probably the same as point 3, we can see after point 3 is fixed.

point 3 - the problem is that the product description field is HTML (when you edit a product you can see you can use italic and bold and because it's a text area there will naturally be line breaks). Taking it in a csv and comparing the csv value with the value in the database is a recipe for disaster.
I dont know why we are doing this: why are we taking in the csv a field we can't update on the DB but we still validate it against the value in the DB?
Anyway, I can update products but I have to go to the DB and get the value from there and put it in the csv. This is how they look like:
Screenshot 2019-03-28 at 15 51 47

I think it's better if we create a separate issue to address this last problem. We need to agree on what will be the best solution...

RachL:
Ok so I've opened https://github.com/openfoodfoundation/openfoodnetwork/issues/3663 and https://github.com/openfoodfoundation/openfoodnetwork/issues/3664

Moving this one forward then, we will open each issue one by one as we fix things.

luisramos0:
ok, cool, thanks!!!
Did you manage to go past the error in point 1 by setting shipping category?

RachL:
@luisramos0 yes I did.

mkllnk:
Documentation changes? We need to create a todo list:  #3598

PR #3316, opened 2019-01-11
[Spree Upgrade] Phase 2 - Reports ready - all specs running and green (except enterprise fee report)

### LINKED ISSUE #3373: [Spree Upgrade] Fix reports_controller_spec (12 broken specs)
Job 2 - 12 x spec/controllers/spree/admin/reports_controller_spec.rb

```
 4) Spree::Admin::ReportsController Coordinator Enterprise User Orders & Fulfillment shows all orders in order cycles I coordinate
     Failure/Error: let(:resulting_orders) { assigns(:report).table_items.map(&:order) }
     
     NoMethodError:
       undefined method `table_items' for nil:NilClass
     # ./spec/controllers/spree/admin/reports_controller_spec.rb:62:in `block (2 levels) in <top (required)>'
     # ./spec/controllers/spree/admin/reports_controller_spec.rb:75:in `block (4 levels) in <top (required)>'
```

It’s a problem with the “assigns” method used in this spec to fetch the report object from the controller

luisramos0:
#### What? Why?

Closes #3373
All specs related to reports are green now.

There was [a change](https://github.com/openfoodfoundation/spree/commit/3685569db4dd0890faef1e08fb40f099c66cdc0b) in the model_class of the ReportsController that broke the ability_decorator. Here I adapt the ability decorator and the navigation helper.

#### What should we test?
All reports need to be validated as this is the first time we say reports are ready in v2.

Enterprise Fee Summary report is the only report that still needs to be fixed, that's in #3483 and is out of scope for this PR.

### DIFF
--- a/app/controllers/spree/admin/reports_controller_decorator.rb
+++ b/app/controllers/spree/admin/reports_controller_decorator.rb
@@ -250,7 +250,7 @@ Spree::Admin::ReportsController.class_eval do
       :xero_invoices,
       :packing
     ]
-    reports = all_reports.select { |action| can? action, :report }
+    reports = all_reports.select { |action| can? action, Spree::Admin::ReportsController }
     reports.map { |report| [report, describe_report(report)] }.to_h
   end
 
--- a/app/helpers/spree/admin/navigation_helper_decorator.rb
+++ b/app/helpers/spree/admin/navigation_helper_decorator.rb
@@ -20,6 +20,7 @@ module Spree
         klass = Spree::Order if klass == :bulk_order_management
         klass = EnterpriseGroup if klass == :group
         klass = VariantOverride if klass == :Inventory
+        klass = Spree::Admin::ReportsController if klass == :report
         klass
       end
       alias_method_chain :klass_for, :sym_fallback
--- a/app/models/spree/ability_decorator.rb
+++ b/app/models/spree/ability_decorator.rb
@@ -187,7 +187,7 @@ class AbilityDecorator
     # Reports page
     can [:admin, :index, :customers, :orders_and_distributors, :group_buys, :bulk_coop, :payments,
          :orders_and_fulfillment, :products_and_inventory, :order_cycle_management, :packing],
-        :report
+        Spree::Admin::ReportsController
     add_enterprise_fee_summary_abilities(user)
   end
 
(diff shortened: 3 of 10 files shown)


RachL:
@luisramos0 I think you were working on adding some live datas on stagings servers right? Do you think this is something we could do to better test reports? we could then compare production reports to v2 reports.

luisramos0:
yes, let's do it in FR staging, it has 4gb ram, AUS staging only has 2GBs.

RachL:
@luisramos0 wooooaaah you rock!!

RachL:
I've compared reports for OFN UK Demo Hub both on staging FR and UK prod. I found one difference in the payment method and delivery reports in the balance displayed. Do you know how the balance column is calculated @luisramos0 because I actually find the column on staging FR more accurate (just by reading the info) 😆 

https://docs.google.com/document/d/1HXHg3nIg_M4d_IOd2-IQJtBsGsb7bqOEdezUd3mXMe8/edit#

luisramos0:
I have never even opened this report before but I can have a look at the code.

It's the user balance.
The user balance from UserBalanceCalculator is calculated based on the user email and the distributor of the specific order.

The balance is 
**payment_total - completed_order_total**

where
payment_total = the sum of payment.amount of all **completed** payments in all orders (not cancelled) of that user in that distributor
completed_orders_total = the sum of order.total of all orders (not cancelled) of that user in that distributor

this is in [this code](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/open_food_network/user_balance_calculator.rb) and it's the same in both v1 and v2.

RachL:
@luisramos0 then I think there is a small bug in v2. The payment was in cash and was not payed. The balance should be 1,8 in both cases as the formula you gave me takes into account **_all_** orders. However in FR staging it seems to me that the balance is calculated **_per_** order. Do you agree? If so, shall I open an issue for this and we merge this or should this be fixed here?

luisramos0:
yeah, in the code it looks like it should include all orders. but in the report, I dont know. does it make sense to include the overall balance or the balance for each order. anyway, lets create a new issue and merge this one.
Thanks!

RachL:
I've created this issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/3677

PR #3322, opened 2019-01-14
Delete dead code and unused folders

luisramos0:
#### What? Why?

Removing a few unused things from the code base:
- spec/archive
- /doc (moved to images to wiki [here](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade:-Development-⛑), I have created a folder for all spree upgrade pages, only seen if you checkout the wiki)
- delete vendor empty folders (I have checked staging and live environments, these folders are empty)

#### What should we test?
Validate that the deployment and app work without the folders under /vendor.
We can do a very quick sanity check in staging.

#### Release notes
Changelog Category: Removed
Removed unused folders and files from the source code making it more simple.

#### Documentation updates
See wiki update above.


RachL:
Deploying on Aus staging

RachL:
Sanity check ok. Testing notes:
https://docs.google.com/document/d/18MzB0R8hjZcqPiKGiZJ33zdR3tgPyQHBvWcW8R8j7qc/edit#

Issue #3310, opened 2019-01-11
[Inventory] Tags can't be removed when too many columns are displayed


mkllnk:
## Description

The inventory table has several columns which are optional. When you enable several columns, the Tags column can become too small and the delete button of tags is hidden. It's not obvious how to remove tags in this scenario.

There are two workarounds:

1. Hide columns until the delete buttons appear.
2. Add another tag which recalculates the size of the tags field. The delete buttons are then visible and you can delete the new tag and whichever other tag you wanted to remove.

## Expected Behavior

Tags can always be removed from inventory items.

## Actual Behavior

The delete button may be hidden.

## Steps to Reproduce

1. Go to the Inventory `/admin/inventory`.
2. Select an enterprise.
3. Add products if its empty.
4. Add tags to a product and save.
5. Reload the page and select the enterprise again.
6. Click on "COLUMNS" and select "Enable Stock Reset?", "Inherit?" and "Tags".
7. Observe the tags and their "x" delete button.

## Animated Gif/Screenshot

Good view:
![screenshot from 2019-01-11 17-18-02](https://user-images.githubusercontent.com/3524483/51016811-c804a900-15c5-11e9-9069-939e80791425.png)

Bad view:
![screenshot from 2019-01-11 17-17-43](https://user-images.githubusercontent.com/3524483/51016818-cf2bb700-15c5-11e9-9508-c61cd2457a2a.png)

## Context

I'm working on #2469 (using Chrome as test browser) and one feature spec can't click the remove button, because it's not visible.

## Severity

s4: It affects only a few users, but they may get stuck and think that it doesn't work at all.

filipefurtad0:
While testing #6789 and going through `spec/features/admin/variant_overrides_spec.rb` I noticed the reference to this issue. This might be resolved now, since the introduction #6070 . 

I ran this spec after commenting out line 303, and it passes - the columns are now broad enough on a regular 1600 x 816 browser window. However, it fails again though if the window width is reduced to half a screen (800 px).

Maybe it is possible to define a minimal width, for the Tags column? Is this feasible? ping @jibees 

I think that could close this issue, and then, line 303 could be removed from the spec.

jibees:
Yeah, I agree we should find a decent solution (like `min-width` or maybe enlarge tag to its real size on mouse hover or truncate the text and always display the cross or mixture of all that). Maybe @Erioldoesdesign has a better idea? 
[...]

Issue #2469, opened 2018-07-13
Replace PhantomJS with headless Chrome


mkllnk:
We see a lot of test failures on our Continuous Integration servers. Some of them could be related to PhantomJS not being very consistent and reliable.

## Description

Rob has been working on using headless Chrome instead of PhantomJS for our feature tests. I thought that was a really good idea, but it is actually a lot of work to change our tests to work with Chrome. I'm also wondering if Chrome is what we want. In an ideal world, our tests would run with all browsers, but at the moment the tests don't pass in all browsers (different test behaviour) and we don't have enough CI power.

So assuming that we will test with only one browser, here is a little breakdown:

For PhantomJS:
- It's like an old Safari. We test compatibility with older browsers.
- It's made for testing and can raise Javascript errors as Ruby errors.
- We are used to it and we don't have to change tests.

For Chrome:
- PhantomJS is not maintained any more: https://github.com/ariya/phantomjs/issues/15344
- Chrome is more reliable.
- It behaves like a real browser, not something emulating a browser.
- Chrome has a global market share of ~50%. https://en.wikipedia.org/wiki/Usage_share_of_web_browsers
- A simple option changed on a developer's machine can activate real browsing to see what's going on (not headless).

I tend to switch to Chrome, but I wanted to double check with you if you have any concerns.

## Expected Behavior

Tests always pass.

## Actual Behavior

Tests sometimes pass.

## Steps to Reproduce

Open a pull request to trigger a test run.

## Context

When developing, our test servers often have different results to my own machine. And I can't really see what PhantomJS is doing apart from `save_and_create_screenshot`.

## Severity

s4 tech debt

## Possible Fix

Use headless Chrome or Firefox. Rob started that work and I added a little bit. Now we are managing a branch as a team, creating pull requests to fix the broken tests: https://github.com/openfoodfoundation/openfoodnetwork/compare/headless-chrome?expand=1

**7** Broken tests:
```
rspec ./spec/features/admin/variant_overrides_spec.rb:289 -> #3326

rspec ./spec/features/admin/bulk_order_management_spec.rb:485 -> #3325
rspec ./spec/features/admin/shipping_methods_spec.rb:79
rspec ./spec/features/admin/product_import_spec.rb:130

rspec ./spec/features/admin/customers_spec.rb:24
rspec ./spec/features/admin/enterprises/images_spec.rb:29
rspec ./spec/features/admin/enterprises/images_spec.rb:66
```

[...]

PR #3326, opened 2019-01-15
Cleanup specs

mkllnk:
#### What? Why?

Helps with #2469.

Working on the headless-chrome branch I found #3310 which made specs fail in Chrome. In this change I'm not fixing that bug, but simplify the spec code and remove dead code. I also added a workaround (one line) to make the spec pass in Chrome.

#### What should we test?

Only automated testing is affected.

#### Release notes

Changed: Improved test code quality.

Changelog Category: Changed

#### Dependencies

#2469 depends on this.

### DIFF
--- a/spec/support/request/admin_helper.rb
+++ b/spec/support/request/admin_helper.rb
@@ -1,8 +1,4 @@
 module AdminHelper
-  def have_admin_menu_item(menu_item_name)
-    have_selector "ul[data-hook='admin_tabs'] li", text: menu_item_name
-  end
-
   def toggle_columns(*labels)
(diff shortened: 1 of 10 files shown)


spec/support/request/admin_helper.rb
```diff
@@ -1,5 +1,13 @@
 module AdminHelper
-  def have_admin_menu_item(menu_item_name)
-    have_selector "ul[data-hook='admin_tabs'] li", text: menu_item_name
+  def toggle_columns(*labels)
```
sauloperez: niiice

PR #3328, opened 2019-01-15
[Spree Upgrade] Phase 2 - Product Import - Adapt product_reset_strategy

### LINKED ISSUE #2782: [Spree Upgrade][PI] Make Product Import update stock items
## Description

Product Import is now updating all `count_on_hand` columns of a bunch of variants in one shot, while in Spree v2.0 this column is moved to stock items. We need PI to reflect that, while keeping a single update statement for performance purposes.

luisramos0:
#### What? Why?
Change product import's product_reset_strategy from depending on the inexistent variant.count_on_hand DB field and instead make individual calls to variant.count_on_hand= defined in VariantStock.
Closes #2782.

#### What should we test?
Fixes 6 specs in products_reset_strategy_spec, moved to pending in #2791.

### DIFF
--- a/app/models/product_import/products_reset_strategy.rb
+++ b/app/models/product_import/products_reset_strategy.rb
@@ -7,18 +7,16 @@ module ProductImport
     def reset(enterprise_ids)
       @enterprise_ids = enterprise_ids
 
-      if enterprise_ids.present?
-        relation.update_all(count_on_hand: 0)
-      else
-        0
-      end
+      return 0 if enterprise_ids.blank?
+
+      reset_variants_count_on_hand(enterprise_variants_relation)
     end
 
     private
 
     attr_reader :excluded_items_ids, :enterprise_ids
 
-    def relation
+    def enterprise_variants_relation
       relation = Spree::Variant
         .joins(:product)
         .where(
@@ -29,6 +27,24 @@ module ProductImport
       return relation if excluded_items_ids.blank?
 
       relation.where('spree_variants.id NOT IN (?)', excluded_items_ids)
+    end
+
+    def reset_variants_count_on_hand(variants)
+      updated_records_count = 0
+      variants.each do |variant|
+        updated_records_count += 1 if reset_variant_count_on_hand(variant)
+      end
+      updated_records_count
+    end
+
+    def reset_variant_count_on_hand(variant)
+      begin
+        variant.count_on_hand = 0
+        return true if variant.count_on_hand.zero?
+      rescue StandardError => e
+        Rails.logger.info "Could not reset variant count on hand: #{e.message}"
+      end
+      false
     end
   end
 end
--- a/app/models/product_import/products_reset_strategy.rb
+++ b/app/models/product_import/products_reset_strategy.rb
@@ -38,13 +38,8 @@ module ProductImport
     end
 
     def reset_variant_count_on_hand(variant)
-      begin
-        variant.count_on_hand = 0
-        return true if variant.count_on_hand.zero?
-      rescue StandardError => e
-        Rails.logger.info "Could not reset variant count on hand: #{e.message}"
-      end
-      false
+      variant.count_on_hand = 0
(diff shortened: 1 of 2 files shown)


app/models/product_import/products_reset_strategy.rb
```diff
@@ -30,5 +28,23 @@ def relation
...
+      end
+      updated_records_count
+    end
+
+    def reset_variant_count_on_hand(variant)
+      begin
+        variant.count_on_hand = 0
+        return true if variant.count_on_hand.zero?
+      rescue StandardError => e
+        Rails.logger.info "Could not reset variant count on hand: #{e.message}"
```
sauloperez: I would prefer not to rescue the exception so that it reaches Bugsnag. That's the mechanism we have for error tracking while logs need to be manually grepped, so we won't notice unless we actively check it.

Then, we wouldn't need lines 43 and 47 as the falsy case will raise an exception, for all the others we can assume a truthy case.

luisramos0: I am not sure I follow you. can you give code example?
if there are 10 items to be reset and one is broken (either throws an exception or count_on_hand is not zero after calling count_on_hand=0), I want the main reset method to return 9.

mkllnk:
Did you change anything after I approved it? Or did you mean to pint @sauloperez?

luisramos0:
oh. sorry i wasnt clear. its because you moved it to in dev. i just wanted to let you know that i think it should be in code review now.

app/models/product_import/products_reset_strategy.rb
luisramos0: ping @sauloperez

sauloperez: My point is that in my experience logging without a proper log infrastructure that makes it easy to explore them they are not of much use for this case. Bugsnag is the error tracking service we use and it is better suited for that. So, if we really want to know when this might fail, we shouldn't rescue the exception. This way that error will show up in Bugsnag's dashboard.

Secondly, going in-depth, why would a `#coun_on_hand=` fail making `#count_on_hand` return a non-zero value? I guess the only feasible possibility is that there is no stock item available. In that case, I prefer it to fail quickly and laud so that we can fix things asap, rather than going it unnoticed and realizing when it's too late.

As a side note, it took me a while to realize that this setter does not behave like regular AR setters because it persists changes in DB and can raise exceptions. Did this happen to you guys before?

luisramos0: That is not my experience with logging at all. A normal log file on the server with messages like this is a valuable source of info. So I disagree with your approach of no logs, only Bugsnag. One point would be that we can build an application that logs these messages and later setup a proper logging framework/infrastructure. Having said that I am happy to follow your opinion here.

re second point, fair enough, count_on_hand should just work, if it doesnt' we get an RuntimeError.

re the last question, I know this code too well now :-(

so, I rebased and added one commit where I remove the rescue. is that what you were thinking about? if not please provide specific example.
Thanks!!! (why are you doing code review during a conference?)

sauloperez: I'm not saying no logs but as it is now it's quite useless. We need that infra. you mention and like any other error, I want it to be in Bugsnag so I can take action.

spec/models/product_import/products_reset_strategy_spec.rb
```diff
@@ -62,36 +55,37 @@
...
+          expect(updated_records_count).to eq(1)
+        end
+
+        context 'and there is an unresetable variant' do
+          before do
+            variant.stock_items = [] # this makes variant.count_on_hand raise an error
+          end
+
+          it 'returns correct number of resetted variants' do
+            expect { products_reset.reset(supplier_ids) }.to raise_error RuntimeError
```
sauloperez: Exactly. This will bubble up until Bugsnag error tracking code.

PR #3329, opened 2019-01-15
[Headless Chrome] Fix broken product import spec

Matt-Yorkley:
#### What? Why?

Fixes a spec in #2469.

I think this previous line was originally a workaround, but anyway... headless doesn't like interacting with hidden elements, so I just switched it to use the proper `select2_select` method.

#### What should we test?

Nothing to test. Reviewers should check that the spec is fixed. :heavy_check_mark:


sauloperez:
:tada: it's happening!

PR #3331, opened 2019-01-16
[Headless Chrome] Add chromedriver-helper to Gemfile for test and development

Matt-Yorkley:
#### What? Why?

Helps development of #2469

Adding `chromedriver-helper` to the Gemfile. Specs don't run on my machine using headless chrome without this. Unless anyone has any tips on making it run on Debian... @mkllnk ?

#### What should we test?

Nothing to test. Only applies to dev and test environments.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -118,6 +118,7 @@ group :test, :development do
   gem 'letter_opener', '>= 1.4.1'
   gem 'timecop'
   gem 'selenium-webdriver'
+  gem 'chromedriver-helper'
   gem 'rspec-retry'
   gem 'json_spec', '~> 1.1.4'
   gem 'unicorn-rails'


mkllnk:
I didn't need that gem on my Debian system, but I have chromedriver installed through apt:
```
apt install chromedriver
```

I personally trust the signed Debian packages more than something downloaded through a gem.

Matt-Yorkley:
I was going to play around with my environment and try getting it working without this gem, but I think this might be a good idea anyway. All devs will need to set up `chromedriver` when we switch, and this will now be handled with a `bundle install`. It should also help keep the same driver version across dev environments, and pre-empt potential setup issues for new contributors in different environments.

luisramos0:
@Matt-Yorkley 💯

PR #3333, opened 2019-01-16
[Headless Chrome] Fix shipping method tests

Matt-Yorkley:
#### What? Why?

Fixes a test on #2469 

Adjusts the `flash_message` method in spec helper to suit chrome.

#### What should we test?

Nothing to test. `shipping_methods_spec.rb` should be green.

### DIFF
--- a/spec/features/admin/shipping_methods_spec.rb
+++ b/spec/features/admin/shipping_methods_spec.rb
@@ -94,6 +94,9 @@ feature 'shipping methods' do
 
       check "shipping_method_distributor_ids_#{distributor1.id}"
       find(:css, "tags-input .tags input").set "local\n"
+      within(".tags .tag-list") do
+        expect(page).to have_css '.tag-item'
+      end
 
       click_button 'Create'
 
--- a/spec/support/request/web_helper.rb
+++ b/spec/support/request/web_helper.rb
@@ -42,7 +42,7 @@ module WebHelper
   end
 
   def flash_message
-    find('.flash', visible: false).text.strip
+    find('.flash', visible: false).text(:all).strip
   end
 
   def handle_js_confirm(accept=true)


Matt-Yorkley:
This is passing when I run the spec locally...?

mkllnk:
It doesn't pass on my computer. Do you use Spring or Zeus?

mkllnk:
Uh, I had one random pass. That made me think it may be a timing issue. This works reliably for me:
```
      sleep 0.5
      click_button 'Create'
```

I'm not sure why the tag entering needs a bit of time before we click Create.

Matt-Yorkley:
Thanks @mkllnk, we just needed to wait for Angular to finish doing it's thing before submitting the form. I did this by checking the tag UI element had been inserted.

screenshot from 2019-01-16 10-05-49

It's green! :heavy_check_mark:

spec/features/admin/shipping_methods_spec.rb
```diff
@@ -94,6 +94,9 @@
 
       check "shipping_method_distributor_ids_#{distributor1.id}"
       find(:css, "tags-input .tags input").set "local\n"
+      within(".tags .tag-list") do
+        expect(page).to have_css '.tag-item'
```
sauloperez: why not using a `find` instead of this line? afaik it'll block the execution until the css is found.

Having this tag-item it's not the assertion this particular spec was meant for, it is just part of the set up of the test. As such, using `expect` doesn't seem accurate to me in terms of meaning. And it's surely slower.

Matt-Yorkley: I thought `find` was immediate, whereas `expect` included some waiting if the element hadn't appeared yet in the DOM?

Matt-Yorkley:
@luisramos0 this is the last one! When we merge this we can rebase the `headless-chrome` branch on master and check the build. It should be all green and ready to roll. :grinning: :tada:

spec/features/admin/shipping_methods_spec.rb
sauloperez: I remember seeing it explained in Capybara's README but it is also explained in https://robots.thoughtbot.com/write-reliable-asynchronous-integration-tests-with-capybara#find-the-first-matching-element.

sauloperez:
I'm so looking forward! thanks @Matt-Yorkley for taking it!

spec/features/admin/shipping_methods_spec.rb
Matt-Yorkley: Done!

luisramos0:
shipping_methods_spec is not green @Matt-Yorkley 
```
  2) shipping methods as an enterprise user creating a shipping method
     Failure/Error: find('.flash', visible: false).text(:all).strip
     
     Capybara::ElementNotFound:
       Unable to find css ".flash"
     # ./spec/support/request/web_helper.rb:79:in `flash_message'
     # ./spec/features/admin/shipping_methods_spec.rb:101:in `block (3 levels) in <top (required)>'
```

luisramos0:
Hey @Matt-Yorkley I rebased headless-chrome, this PR needs rebasing as well.
BUT, in the new build (with the rebase and the gemlock fix) this spec is not broken! Do you understand why?
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3336/builds/2

Matt-Yorkley:
@luisramos0 that makes sense. We identified that this spec was a bit flakey, basically it was submitting a form whilst some angular was still loading, without explicitly waiting for it to finish. So it makes sense if it was randomly green on the other branch. I had it passing locally but failing in Semaphore a few times.

Matt-Yorkley:
Ah I see, it looks like it's failing fairly consistently since I changed from `expect` to `find`.  I'm 99% sure we need an `expect` in there and that `find` doesn't wait for missing DOM elements in the same way, so I'll change it back to how I had it originally if @sauloperez doesn't mind :trollface:

Matt-Yorkley:
Okay, `shipping_methods_spec.rb` is green again, and we have 2 new errors, one for `bulk_order_management_spec.rb` that didn't appear before, and one for `embedded_shopfronts_spec.rb` that just looks like it needs a nice chrome-flavoured `accept_js_alert` at line 101.

sauloperez:
ok. I need to investigate that myself then. There's no way around knowing Capybara better.

PR #3334, opened 2019-01-16
3332 [Enterprise Fee Summary] Disable report for non-superadmin users

### LINKED ISSUE #3332: [Enterprise Fee Summary] Hide report link with feature flag
Hide report link with feature flag.

## Description

This was recommended by @sauloperez [in Slack](https://openfoodnetwork.slack.com/messages/C2GQ45KNU/convo/C2GQ45KNU-1547186925.158100/).

The link should be visible only to superadmins. However, any enterprise user can also access the report by going directly to the URL.

## Expected Behavior

A non-superadmin user should not see "Enterprise Fee Summary" in the list of reports.

## Actual Behavior

All enterprise users see "Enterprise Fee Summary" in the list of reports.

## Steps to Reproduce

In the admin section:

1. Log in as a non-superadmin user.
2. Go to the admin section.
3. Click on "Reports".
4. Notice the "Enterprise Fee Summary" link.

kristinalim:
#### What? Why?

Closes #3332

This exposes the "Enterprise Fee Summary" link and resource to only superadmins. This will allow us to merge existing changes for the report to master even if there are still bugs.

This doesn't enable the feature for non-production environments. Note that this limits permission testing that can be done, so a thorough round of permission testing needs to be done before the feature toggle is enabled for all enterprise users.

#### What should we test?

As superadmin:

1. Go to "Reports". There should be a link to "Enterprise Fee Summary".
2. Click on the link. There should be no permission issues.
3. Generate a report that is expected to provide data. Data should be provided.

As manager of an enterprise with orders:

1. Go to "Reports". There should be no link to "Enterprise Fee Summary".
2. Go directly to URL path `/admin/reports/enterprise_fee_summary/new`. There should be a permissions issue.

#### Release notes

- Expose "Enterprise Fee Summary" only to superadmin users.

Changelog Category: Changed

#### Dependencies

This would be merged into the transitional branch "feature/enterprise_fee_summary", not "master".

### DIFF
--- a/app/models/spree/ability_decorator.rb
+++ b/app/models/spree/ability_decorator.rb
@@ -185,9 +185,9 @@ class AbilityDecorator
 
     # Reports page
     can [:admin, :index, :customers, :orders_and_distributors, :group_buys, :bulk_coop, :payments,
-         :orders_and_fulfillment, :products_and_inventory, :order_cycle_management, :packing,
-         :enterprise_fee_summary], :report
-    can [:admin, :new, :create], :enterprise_fee_summary
+         :orders_and_fulfillment, :products_and_inventory, :order_cycle_management, :packing],
+         :report
+    add_enterprise_fee_summary_abilities(user)
   end
 
   def add_order_cycle_management_abilities(user)
@@ -261,8 +261,8 @@ class AbilityDecorator
     # Reports page
     can [:admin, :index, :customers, :group_buys, :bulk_coop, :sales_tax, :payments,
          :orders_and_distributors, :orders_and_fulfillment, :products_and_inventory,
-         :order_cycle_management, :xero_invoices, :enterprise_fee_summary], :report
-    can [:admin, :new, :create], :enterprise_fee_summary
+         :order_cycle_management, :xero_invoices], :report
+    add_enterprise_fee_summary_abilities(user)
 
     can [:create], Customer
     can [:admin, :index, :update, :destroy, :show], Customer, enterprise_id: Enterprise.managed_by(user).pluck(:id)
@@ -284,6 +284,16 @@ class AbilityDecorator
     can [:destroy], EnterpriseRelationship do |enterprise_relationship|
       user.enterprises.include? enterprise_relationship.parent
     end
+  end
+
+  def add_enterprise_fee_summary_abilities(user)
+    feature_enabled = FeatureFlags.new(user).enterprise_fee_summary_enabled?
+    return unless feature_enabled
+
+    # Reveal the report link in spree/admin/reports#index
+    can [:enterprise_fee_summary], :report
+    # Allow direct access to the report resource
+    can [:admin, :new, :create], :enterprise_fee_summary
   end
 end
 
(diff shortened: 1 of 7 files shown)


app/models/spree/ability_decorator.rb
```diff
@@ -283,6 +283,16 @@ def add_relationship_management_abilities(user)
...
+
+  def add_enterprise_fee_summary_abilities(user)
+    feature_enabled = FeatureFlags.new(user).enterprise_fee_summary_enabled?
+    return unless feature_enabled
+
+    # Reveal the report link in spree/admin/reports#index
+    can [:enterprise_fee_summary], :report
+    # Allow direct access to the report resource
+    can [:admin, :new, :create], :enterprise_fee_summary
+  end
```
sauloperez: This file is already quite unmanageable and although I love the path you take @kristinalim I'd go one step further and split abilities into multiple files. It is as easy as registering the individual classes as we do in line 298. If you agree, could you create a tech debt issue for it?

spec/support/feature_flags_helper.rb
```diff
@@ -0,0 +1,10 @@
+module FeatureFlagsHelper
+  def set_feature_flag(user, feature, status)
+    feature_flags = instance_double(FeatureFlags, "#{feature}?" => status)
+    allow(FeatureFlags).to receive(:new).with(user) { feature_flags }
+  end
+
+  def unstub_feature_flags(user)
+    allow(FeatureFlags).to receive(:new).with(user).and_call_original
+  end
+end
```
sauloperez: I'm not sure it's really worth introducing new methods for what `allow` already does and with which all devs are familiar with. If feature flags were more tedious it'd make more sense.

spec/models/feature_flags_spec.rb
```diff
@@ -25,4 +25,22 @@
       end
     end
   end
+
+  describe ".enterprise_fee_summary_enabled?" do
```
sauloperez: the convention of the `.` prefix is used to note this is a class method while `#` for instance ones. As this case is the latter, I would stick to it and use `#`. It will be confusing otherwise.

sauloperez:
I'm really happy to see us adopting feature flags to ship more often :tada: !

spec/models/feature_flags_spec.rb
kristinalim: Agh, yes thanks! I copied that from the other example group - fixed both already.

spec/support/feature_flags_helper.rb
kristinalim: After thinking it over, I'm okay with removing these spec helpers. Done! Thanks, @sauloperez.

app/models/spree/ability_decorator.rb
kristinalim: Nice! I created the tech debt issue: #3345

kristinalim:
I've made changes based on the review.

@sauloperez kindly approve if you're already okay with this. :slightly_smiling_face:

kristinalim:
FYI, I rebased this PR's branch upon the recently rebased (upon master) `feature/enterprise_fee_summary`.

kirstenalarsen:
ok flag works. Fee report can be seen / used as superadmin but not as enterprise

kristinalim:
Okay, merging this to our transitional branch now. Thanks, @kirstenalarsen!

PR #3339, opened 2019-01-16
[Headless Chrome] Fix embedded_shopfront_spec unhandled alert

### LINKED ISSUE #2674: Make embedded_shopfronts_spec reliable

Matt-Yorkley:
#### What? Why?

Fixes a test in #2469 
Closes #2674 

This looks like it could be another flakey one that pops up occasionally based on timing. `Selenium::WebDriver::Error::UnhandledAlertError: unexpected alert open`.

#### What should we test?

No testing. `embedded_shopfront_spec` should be green.

### DIFF
--- a/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
+++ b/spec/features/consumer/shopping/embedded_shopfronts_spec.rb
@@ -115,13 +115,15 @@ feature "Using embedded shopfront functionality", js: true do
   end
 
   def login_with_modal
-    expect(page).to have_selector 'div.login-modal', visible: true
+    page.has_selector? 'div.login-modal', visible: true
 
     within 'div.login-modal' do
       fill_in "Email", with: user.email
       fill_in "Password", with: user.password
       find('input[type="submit"]').click
     end
+
+    page.has_no_selector? 'div.login-modal', visible: true
   end
 
   def logout_via_navigation


luisramos0:
is this also  improving #2674 ? :-)

Matt-Yorkley:
Yes, it should close that one :)

Matt-Yorkley:
I spent a while playing with this one, using `save_and_open_screenshot` to check the state of the page at various points in the test. The issue was with a popup appearing unexpectedly if the page hadn't finished loading before moving on to the next step in the test. There was a previous attempt to make the spec consistent but it wasn't working as intended. I've tweaked it and it is running consistently. I've checked by re-running the test about 15 times, and the issue seems to be fixed.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
```diff
@@ -110,8 +111,8 @@
   # When we click the "Update" button to save changes, we need to wait
   #   until it is actually saved and "loading" disappears before doing anything else.
   def wait_for_shop_loaded
-    page.has_no_content? "Loading"
-    page.has_no_css? "input[value='Updating cart...']"
```
sauloperez: We're clearly missing something. I thought these methods wait for the text/css to disappear.

luisramos0:
headless-chrome is merged to master and I'll delete the branch now.
I switched this to master but semaphre didnt react. I clicked the rebuild button on sempahore. I assume it will run the build based on master and not the branch, lets see.

sigmundpetersen:
Need a rebase or something?

mkllnk:
There are so many flaky specs. Build number 6...

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
mkllnk: I always struggle to find information about this. https://blog.codeship.com/faster-rails-tests/ suggests that `has_no_content?` is waiting. I just tested it:

- `page.has_content?` is waiting
- `page.has_no_content?` is waiting
- `page.has_no_css?` is waiting

[This post](https://stackoverflow.com/a/36842392/3377535) says that `has_content?` is just a wrapper for `have_content`.

@Matt-Yorkley How did you test that it doesn't block? There may be something else we need to wait for and your change made the failure only less likely.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
```diff
@@ -94,6 +94,7 @@
         wait_for_shop_loaded
         find('ul.right li#login-link a').click
         login_with_modal
+        expect(page).to_not have_selector 'div.login-modal', visible: true
```
mkllnk: This is the first of three waiting statements. The last one should be enough if it was working. I guess that these three statements just slow down execution and make the failure less likely. We need to find a way to determine when all the JS work is done.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
```diff
@@ -126,6 +127,7 @@ def login_with_modal
 
   def logout_via_navigation
     first('ul.right li.user-menu a').click
+    expect(page).to have_css 'ul.right ul.dropdown li a[title="Logout"]'
     find('ul.right ul.dropdown li a[title="Logout"]').click
```
mkllnk: `find` is waiting as well. We are waiting for the same thing twice.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
sauloperez: Yes, I also remember that article talking about it. Actually, I didn't know before.

Matt-Yorkley:
I've done some more investigating and I think I understand it a bit better now. The original lines in the `wait_for_shop_loaded` method were waiting correctly, but the second time it is called in the test is directly after `logout_via_navigation` is called, and the page hadn't reloaded yet before `wait_for_shop_loaded` was called the second time. So the checks to see if the page contained the loading elements was running while the login modal was still open and before the login had actually finished, so it "passed" incorrectly, and then left the page loading after that point which is what lead to the unexpected alert error.

I've updated the PR.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
```diff
@@ -115,13 +115,15 @@ def wait_for_shop_loaded
...
-    expect(page).to have_selector 'div.login-modal', visible: true
+    page.has_selector? 'div.login-modal', visible: true
 
     within 'div.login-modal' do
       fill_in "Email", with: user.email
       fill_in "Password", with: user.password
       find('input[type="submit"]').click
     end
+
+    page.has_no_selector? 'div.login-modal', visible: true
```
Matt-Yorkley: This check to see if the modal was closed before moving on is what we needed, I've moved it into the `login_with_modal` method instead of putting it in the test.

spec/features/consumer/shopping/embedded_shopfronts_spec.rb
Matt-Yorkley: See https://github.com/openfoodfoundation/openfoodnetwork/pull/3339#issuecomment-455655974

sauloperez:
Waiting to hear @mkllnk's opinion.

PR #3325, opened 2019-01-15
2469 Fix async js requests

mkllnk:
#### What? Why?

Fixes one spec in the headless chrome branch.
```
bundle exec rspec ./spec/features/admin/bulk_order_management_spec.rb:485
```

Helps with https://github.com/openfoodfoundation/openfoodnetwork/issues/2469.

#### What should we test?

Check that records are loaded correctly when filtering on these admin pages:

- customers
- bulk order management
- order cycles
- subscriptions
- products

It is unlikely that you can re-produce the original issue on master. But our automated specs cover that.

#### Release notes

Fixed: Searching as admin, e.g. for orders or products is more reliable when typing very fast (like a robot).

Changelog Category: Fixed

#### Dependencies

 #2469 depends on this.

PR #3343, opened 2019-01-16
[Headless Chrome] Fix datepicker interactions in bulk_order_management_spec

Matt-Yorkley:
#### What? Why?

Fixes another test for #2469 

This one was pretty nasty. I thought maybe headless chrome just really didn't like datepickers, but when I booted the app and tried inputting the dates the way we were doing it (as text), the orders didn't load well for me either. Anyway, I've updated the spec so it correctly opens the datepicker and clicks on the actual dates it needs to as if it were a user, and now it's green. :tada: 

#### What should we test?

No testing. The new failing test in `bulk_order_management_spec.rb` should be green.

### DIFF
--- a/spec/features/admin/bulk_order_management_spec.rb
+++ b/spec/features/admin/bulk_order_management_spec.rb
@@ -474,13 +474,17 @@ feature %q{
       end
 
       it "displays only line items whose orders meet the date restriction criteria, when changed" do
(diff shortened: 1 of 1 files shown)


mkllnk:
Oh no, I fixed that already in https://github.com/openfoodfoundation/openfoodnetwork/pull/3325. I thought you knew, because you reviewed that. Well, your solution mimics how most people probably use the datepicker.

PR #3347, opened 2019-01-17
[Spree Upgrade] Sets StockLocation.backorderable_default to false in test factories

### LINKED ISSUE #3114: [Spree Upgrade][BackOffice] Fix admin/bulk_product_update spec (1 broken test)

luisramos0:
This makes the default value of variant.on_demand be false

#### What? Why?
Closes #3114
Replaying #3246 but now making the necessary adaptations.

Notes:
I made most changes in one single commit on purpose, these changes are all closely related and are better looked at together.

I had to fix the :product factory modify part and set on_hand default value.
:simple_product is being used in many places and it's property on_demand as well. At some point we have to decide to use either :product or :simple_product. I think we should use :product and delete :simple_product. Something we should do at some point in the future.

#### What should we test?
see 3246 description https://github.com/openfoodfoundation/openfoodnetwork/pull/3246#issue-241033260

### DIFF
--- a/spec/models/spree/payment_method_spec.rb
+++ b/spec/models/spree/payment_method_spec.rb
@@ -28,16 +28,20 @@ module Spree
     end
 
     it "computes the amount of fees" do
-      pickup = create(:payment_method)
       order = create(:order)
-      expect(pickup.compute_amount(order)).to eq 0
-      transaction = create(:payment_method, calculator: Calculator::FlatRate.new(preferred_amount: 10))
-      expect(transaction.compute_amount(order)).to eq 10
-      transaction = create(:payment_method, calculator: Calculator::FlatPercentItemTotal.new(preferred_flat_percent: 10))
-      expect(transaction.compute_amount(order)).to eq 0
+
+      free_payment_method = create(:payment_method) # flat rate calculator with preferred_amount of 0
+      expect(free_payment_method.compute_amount(order)).to eq 0
+
+      flat_rate_payment_method = create(:payment_method, calculator: Calculator::FlatRate.new(preferred_amount: 10))
+      expect(flat_rate_payment_method.compute_amount(order)).to eq 10
+
+      flat_percent_payment_method = create(:payment_method, calculator: Calculator::FlatPercentItemTotal.new(preferred_flat_percent: 10))
+      expect(flat_percent_payment_method.compute_amount(order)).to eq 0
+
       product = create(:product)
       order.add_variant(product.master)
-      expect(transaction.compute_amount(order)).to eq 2.0
+      expect(flat_percent_payment_method.compute_amount(order)).to eq 2.0
     end
   end
 end
--- a/spec/models/spree/product_spec.rb
+++ b/spec/models/spree/product_spec.rb
@@ -711,26 +711,6 @@ module Spree
         e.variants(true).should be_empty
       end
     end
-
-    describe '#on_hand' do
-      let(:product) { create(:product) }
-
-      context 'when the product has variants' do
-        before { create(:variant, product: product) }
-
-        it 'returns the sum of the on_hand of its variants' do
-          expect(product.on_hand).to eq(Float::INFINITY)
-        end
-      end
-
-      context 'when the product has no variants' do
-        before { product.variants.destroy_all }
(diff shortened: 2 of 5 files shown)


spec/models/spree/payment_method_spec.rb
```diff
@@ -28,16 +28,20 @@ module Spree
...
     it "computes the amount of fees" do
-      pickup = create(:payment_method)
       order = create(:order)
-      expect(pickup.compute_amount(order)).to eq 0
-      transaction = create(:payment_method, calculator: Calculator::FlatRate.new(preferred_amount: 10))
-      expect(transaction.compute_amount(order)).to eq 10
-      transaction = create(:payment_method, calculator: Calculator::FlatPercentItemTotal.new(preferred_flat_percent: 10))
-      expect(transaction.compute_amount(order)).to eq 0
+
+      free_payment_method = create(:payment_method) # flat rate calculator with preferred_amount of 0
```
mkllnk: This is much better. Next time I would be tempted to convert the comment into code as well. Passing on the calculator may not be necessary, but if we rely on the calculator, we shouldn't trust the factory default. It would make the comment obsolete and still work if the default changed.

luisramos0: yes agree.
in this case this is not a default factory, it's [a default value in the object](https://github.com/openfoodfoundation/openfoodnetwork/blob/c178478a4f0210133e98e0fbcc25a4d96943a300/app/models/spree/payment_method_decorator.rb#L47).
maybe I should have called it default_free_payment_method...

luisramos0:
ah, yeah. the product form is not picking up the default product.on_demand value! I ignored that problem when I did the product_stock work because I knew I'd put this PR at some point. So, there is a problem there, the form doesnt pick up the default value. I think we can ignore.

spec/models/spree/product_spec.rb
```diff
@@ -711,26 +711,6 @@ module Spree
...
-        end
-      end
-
-      context 'when the product has no variants' do
-        before { product.variants.destroy_all }
-
-        it 'returns the on_hand of the master' do
-          expect(product.on_hand).to eq(product.master.on_hand)
-        end
-      end
```
sauloperez: Are these cases covered elsewhere?

luisramos0: yes, it's mentioned in the commit message: product_stock_spec is doing exactly the same

PR #3349, opened 2019-01-17
Fix shipment without fee

sauloperez:
#### What? Why?

Fixes one of the 3 failing specs in #2685

This ensures we can still use Order#shipment although Spree deprecates
it, while fixing a bug at the same time. The problem that was making the
test fail was on `Order#shipment` that Spree defines.

If the shipments association changes, `#shipment` returns stale data.
That is because the order object we might be using is still alive, and
so its @shipment ivar still holds an old shipment object (it's not nil)
and thus `@shipment ||= shipments.last` doesn't evaluate the right-hand
side of the expression.

Note that we need to `prepend` the evaluation of the concern (which it's
been rename) for our methods to take precedence over Spree ones. With
`include`, Spree's `#shipment` would still be picked up making the test
fail.

#### What should we test?
spec/models/spree/order_spec.rb:711 should pass.

### DIFF
--- a/app/models/concerns/order_shipment.rb
+++ b/app/models/concerns/order_shipment.rb
@@ -8,8 +8,12 @@ require 'active_support/concern'
 # for details.
 #
 # Methods in this module may become deprecated.
-module OrderShippingMethod
+module OrderShipment
   extend ActiveSupport::Concern
+
+  def shipment
+    shipments.first
+  end
 
   # Returns the shipping method of the first and only shipment in the order
   #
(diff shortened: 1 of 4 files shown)


mkllnk:
Yes, but I'm not sure what's better though. It's a patch that would benefit other projects as well if it wasn't for a deprecated method that nobody should be using any more.

luisramos0:
order.shipment is not really deprecated in OFN. because of our constraint of single shipment per order.

sauloperez:
Yes, I started my TDD cycle defining the method there but... after some thinking, this is a bug on Spree, so in normal conditions we would have opened a PR against Spree as it has nothing to do with our codebase. Besides, I feel uneasy making our decorators bigger.

Nonetheless, @luisramos0 I think you're right. It's an OFN decision to have `#shipment` which might live longer than Spree's deprecated method.

luisramos0:
yes, agree. Perfect!

sauloperez:
Take a look at the description. I updated it to reflect the changed approach.

sauloperez:
done

sauloperez:
@mkllnk what's your opinion? I merged too early :see_no_evil:

PR #3354, opened 2019-01-18
New Knapsack report

luisramos0:
#### What? Why?

Rebuilds the knapsack report to make the build jobs more balanced.

I use this PR to improve proxy_order spec that failed twice in the build.

#### What should we test?
the build jobs should take the same amount of time.


Gemfile.lock
```diff
@@ -748,9 +748,6 @@ GEM
       addressable (>= 2.3.6)
       crack (>= 0.3.2)
       hashdiff
-    websocket-driver (0.7.0)
-      websocket-extensions (>= 0.1.0)
-    websocket-extensions (0.1.3)
```
sauloperez: :warning: I saw this on my local machine as well, but I don't know when. Worth checking.

sauloperez:
Another timing issue:

```
Failures:

  1) 
  As an Administrator
 using page controls using drop down seletors distributor filter displays all line items when 'All' is selected from distributor filter
     Failure/Error: expect(page).to have_no_selector "tr#li_#{li2.id}"
       expected #has_no_selector?("tr#li_34") to return true, got false
     # ./spec/features/admin/bulk_order_management_spec.rb:322:in `block (5 levels) in <top (required)>'
```

Would you mind creating an issue for it_

Gemfile.lock
luisramos0: ok, it came with the headless chrome branch (probably an error) and was removed in the recent paper_trail upgrade PR (just because it run the bundle install). I'll rebase this PR and this will disappear.

luisramos0:
Rebased to remove gemfilelock commit and created #3358.
ready for re-review.

spec/models/proxy_order_spec.rb
```diff
@@ -196,6 +196,6 @@ def expect_cancelled_now(subject)
     # We still need to use be_within, because the Database timestamp is not as
     # accurate as the Rails timestamp. If we use `eq`, we have differing nano
     # seconds.
-    expect(subject.reload.canceled_at).to be_within(1.second).of Time.zone.now
+    expect(subject.reload.canceled_at).to be_within(2.second).of Time.zone.now
```
sauloperez: the time dependency should be solved with [Timecop](https://github.com/travisjeffery/timecop). Did this fail in the build @luisramos0 ?

luisramos0: with timecop.freeze, isn't it?
 I dont want to get into that now. what shall I do? revert this commit and create an issue?

sauloperez: Sorry, I wasn't clear. Yep, let's create a tech debt issue, please.

luisramos0: done here #3369

luisramos0:
the build times in master are a bit wild since headless chrome was merged... I am not sure this PR improved the situation.

Issue #2783, opened 2018-09-29
[Spree Upgrade][PI] Make Product Import create stock items (1 broken spec)


sauloperez:
## Description

In Product Import spreadsheet entries map directly to product, variant and variant override records. That is, an entry can become any of these entities in our DB.

With Spree 2.0 this gets more complex and `variant.count_on_hand` becomes `variant.stock_items.first.count_on_hand`. It can't be longer assumed that `entry.attributes == variant.attributes` (or product or variant override).

sauloperez:
`spec/models/product_importer_spec.rb:312` fails due to the same root cause. Product Import assumes variants have an `on_demand` column and that is no longer the case.

luisramos0:
the description is a bit outdated now because we have since then agreed that we will not play with stock items but instead use the variant_stock facade to handle these cases. done in #3355

luisramos0:
the only case where this is still a problem is the only broken spec after #3355:
spec/models/product_importer_spec.rb:684

In this spec, we test entry_processor.assign_defaults where we set the defaults.
The challenge here is that we set the defaults before we save the "object", but the object can be product or variant, SO, it's not straight forward to do the same as in #3355 where, for variants we save the variant and then set on_hand, for product we set on_hand in the variant.

Let use this issue to represent this broken spec, ok?

luisramos0:
Hey @mkllnk yes, I totally agree with the problem and I like the solution.
Interfaces would come handy here where you would have an Importable Interface and a Factory that takes an Entry and returns an Importable. You would have 5 implementations of Importable: new_product, new_variant, etc. What you describe is similar to this but without interfaces...

Having said that, I dont think this is strictly needed now, I think it's tech debt that we don't necessarily need to pay now (my opinion). We can add one more branch now and get the upgrade done.

The main reason I say this about this task is that ProductImport (although it has a few OO flaws, this is just one) is far from being core to OFN and we should invest quality refactoring time on core parts of the system, that's where it will pay off really quickly. For this reason I'd create a tech debt issue that is lower in priority than for example both #2952 and #2763
It's obviously your call, just wanted to share my opinion.

mkllnk:
@luisramos0 
> Interfaces would come handy here where you would have an Importable Interface and a Factory that takes an Entry and returns an Importable.

I totally agree. This is where I miss Java, too. In Ruby we don't have typed variables and therefore no support of interfaces. We only have unit tests to verify it something adheres to an interface.

[...]

PR #3355, opened 2019-01-18
[Spree Upgrade] Phase 2 - Fix product_importer spec

luisramos0:
Added now mandatory shipping_category_id to CSVs in PI tests
Adapted VariantStock to handle on_hand string values
Adapted EntryValidator and EntryProcessor to new VariantStock logic where on_hand is set at variant level, not at product level

#### What? Why?

Part of #2791

#### What should we test?
with this PR and #3328 and #3347:
models/product_importer_spec and features/admin/product_import_spec should both be green (29 specs fixed!)

Only spec/models/product_importer_spec.rb:684 is still broken, see #2783

### DIFF
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -293,9 +293,10 @@ module ProductImport
     end
 
     def convert_to_trusted_type(untrusted_attribute, trusted_attribute)
-      if trusted_attribute.is_a? Integer
+      case trusted_attribute
+      when Integer
         untrusted_attribute.to_i
-      elsif trusted_attribute.is_a? Float
+      when Float
         untrusted_attribute.to_f
       else
         untrusted_attribute.to_s
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -54,7 +54,11 @@ module ProductImport
     end
 
     def mark_as_new_variant(entry, product_id)
-      new_variant = Spree::Variant.new(entry.attributes.except('id', 'product_id'))
+      new_variant = Spree::Variant.new(entry.attributes.except('id', 'product_id', 'on_hand', 'on_demand'))
+      new_variant.save
+      new_variant.on_demand = entry.attributes['on_demand'] if entry.attributes['on_demand'].present?
+      new_variant.on_hand = entry.attributes['on_hand'] if entry.attributes['on_hand'].present?
+
       new_variant.product_id = product_id
       check_on_hand_nil(entry, new_variant)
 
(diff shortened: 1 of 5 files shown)


app/models/product_import/entry_validator.rb
```diff
@@ -283,7 +287,11 @@ def product_field_errors(entry, existing_product)
     end
 
     def attributes_match?(attribute, existing_product, entry)
-      existing_product.public_send(attribute) == entry.public_send(attribute)
+      if existing_product.public_send(attribute).is_a?(Numeric) || entry.public_send(attribute).is_a?(Numeric)
+        return existing_product.public_send(attribute).to_i == entry.public_send(attribute).to_i
+      else
+        return existing_product.public_send(attribute).to_s == entry.public_send(attribute).to_s
+      end
```
luisramos0: dear ruby experts, there must be a better way to do this, I just want to compare ignoring the type...

HugsDaniel: Why not compare only with `to_s` ? If one is numeric it'll be transformed into a string anyway.

luisramos0: Yes, it was my first try but some numbers will be converted to "1.0" and others to "1" and to_s compare will fail.

mkllnk: Oh dear, it looks like we are allowing too much flexibility. The existing product should always have the same type for an attribute, right? What about the CSV? Always string? If the specs supply other types, it's the problem of the spec.

I think you are generalising too much here. `Numeric` can be float like `2.1` but you just convert to integer, loosing the decimals: `2`.

If you have one trusted value (from the database) and one untrusted value (import) then I would always convert the untrusted value into the format of the trusted value and compare on that basis. So if we read an *on demand* column, we should convert that into a boolean and then compare. If we read a stock level, then as integer. Price as float and so on. If that fails we need to rescue and notify the user.

luisramos0:
I have just rebased this PR.

app/models/product_import/entry_validator.rb
luisramos0: ok, I have a new improved version that handles floats and only converts the entry to the attribute type, not the other way around.
what do you think @mkllnk ?

app/models/product_import/entry_validator.rb
```diff
@@ -283,7 +287,19 @@ def product_field_errors(entry, existing_product)
...
 
     def attributes_match?(attribute, existing_product, entry)
-      existing_product.public_send(attribute) == entry.public_send(attribute)
+      existing_product_value = existing_product.public_send(attribute)
+      entry_value = entry.public_send(attribute)
+      existing_product_value == convert_to_trusted_type(entry_value, existing_product_value)
+    end
+
+    def convert_to_trusted_type(untrusted_attribute, trusted_attribute)
+      if trusted_attribute.is_a? Integer
```
mkllnk: I just learned this a few weeks ago:
```ruby
case trusted_attributes
  when Integer #...
  when Float #...
  else #...
end
```

luisramos0: awesome, I tried something similar but couldnt make it work. thanks, done.

sauloperez:
`spec/models/product_importer_spec.rb:604` is still broken, sorry. Can you take a look @luisramos0 ? It seems that `count_on_hand=` is not being picked up.

luisramos0:
it's fine, this depends on #3328 and #3347.

luisramos0:
all good, confirmed in this build:
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3315/builds/4

Issue #2801, opened 2018-10-02
[Spree Upgrade][Reports] Do not use count_on_hand in products and inventory reports spec (3 broken specs)


sauloperez:
## Description

Products and inventory report specs update the variant's column `count_on_hand` in their setup. This column is now gone. See below:

https://github.com/openfoodfoundation/openfoodnetwork/blob/f0183cd17ec23fd52b637379f5859f0c2a574141/spec/features/admin/reports_spec.rb#L305-L318

## Possible Fix

Use the attribute accessor instead of directly updating the DB column. This will go through our `VariantStock` concern.

Also, it's worth taking the chance to move this report's specs to a separate file. `reports_spec.rb` is already too long.

PR #3361, opened 2019-01-20
2 0 report spec

### LINKED ISSUE #2805: [Spree Upgrade][Reports] Do not use count_on_hand in lettuce_share_report_spec.rb (1 broken spec)
## Description

LettueShareReport `only available items` spec updates the variant's `count_on_hand` column in its setup. This column is now gone. See below:

https://github.com/openfoodfoundation/openfoodnetwork/blob/f0183cd17ec23fd52b637379f5859f0c2a574141/spec/lib/open_food_network/lettuce_share_report_spec.rb#L50-L54

## Possible Fix

Use the attribute accessor instead of directly updating the DB column in line 51. This will go through our `VariantStock` concern, as already happens with the various variants created in the setup phase.

luisramos0:
#### What? Why?

Closes #2805 and #2801

#### What should we test?

Fixes the following 4 specs:
Job 1 - 1 x spec/features/admin/reports_spec.rb:204
Job 1 - 2 x spec/features/admin/reports_spec.rb:310
Job 2 - 1 x spec/lib/open_food_network/lettuce_share_report_spec


spec/features/admin/reports_spec.rb
```diff
@@ -307,11 +308,11 @@
       product2.set_property 'Organic', 'NASAA 12345'
       product1.taxons = [taxon]
       product2.taxons = [taxon]
-      variant1.update_column(:count_on_hand, 10)
+      variant1.count_on_hand = 10
       variant1.update_column(:sku, "sku1")
-      variant2.update_column(:count_on_hand, 20)
+      variant2.count_on_hand = 20
       variant2.update_column(:sku, "sku2")
-      variant3.update_column(:count_on_hand, 9)
```
sauloperez: This was persisting the changes to DB. Isn't it needed? I hope so!

spec/lib/open_food_network/lettuce_share_report_spec.rb
```diff
@@ -1,30 +1,32 @@
...
 
-        report.send(:grower_and_method, v).should == "Producer (Method)"
+        report.send(:grower_and_method, variant).should == "Producer (Method)"
       end
     end
 
     describe "gst" do
       it "handles tax category without rates" do
-        report.send(:gst, v).should == 0
+        report.send(:gst, variant).should == 0
```
sauloperez: Since these methods are private, we should test the public methods that call them instead of them. Can you take a look if this is doable or create an issue otherwise @luisramos0 ?

spec/features/admin/reports_spec.rb
luisramos0: variant_stock.count_on_hand goes around and ends up persisting the stock item count on hand as well.

I am not sure yet but with this comment you made me realize I must have left a hole in my variant_override solution for v2. There's no vo.count_on_hand= setter... so, if a Variant is scoped and count_on_hand= is used, the stock item of the variant is still updated.
Basically, ScopeVariantToHub is missing overrides for on_hand= and on_demand=
I'll investigate and create an issue if required.
This is not related to this PR.

spec/lib/open_food_network/lettuce_share_report_spec.rb
luisramos0: I see, I have removed send from the calls to table because it's not private.
For the other two gst and grower_and_method, the table method is returning their results directly in a data structure, so testing these methods individually is not too bad imo.
https://github.com/openfoodfoundation/openfoodnetwork/blob/8b1c1881f731d6b38290fea3c4ee7bcd69bc2c8b/lib/open_food_network/lettuce_share_report.rb#L33

we do have quite a bit of private methods testing in OFN. shall I create a generic issue with reference to this particular case?

luisramos0: ping @sauloperez

spec/features/admin/reports_spec.rb
luisramos0: today I came back to this very big question in v2. I have the answer.
**Why we don't need scope_variant_to_hub.count_on_hand= nor scope_variant_to_hub.on_demand= ?**
scope_variant_to_hub is what is added to the variant objects so that when methods are called, it will do what needs to be done on the VOs.

These are the multiple places where these would be needed and why they are not needed:
1 - on **checkout** cart_controller, cart_service will go through order_contents.add that will add variant to the line_item, the stock movements are done in order_inventory that uses stock_location and, through overriding in OFN, we make it go through variant.move that is overridden in scope_variant_to_hub and updates vo.count_on_hand
2 - in **order edit page** we use api/shipment that also uses order_contents.add, same process as above
3 - in the **product edit page** we go through admin/products/bulk_update that uses product_set that sets variant.on_hand and variant.on_demand directly, here we are not supposed to use overrides
4 - in **inventory edit page**, we use admin/variant_overrides/bulk_update that uses VariantOverrideSet that will set on_hand and on_demand directly on the VariantOverride columns, so, no scope_variant_to_hub is needed

This is why no on_hand=, count_on_hand= or on_demand= are NOT present in scope_variant_to_hub
Only scope_variant_to_hub.move is needed which is used in point 1 and 2 above.

**Summary, it's all good, nothing needs to be done.** ;-)

PR #3299, opened 2019-01-09
[Spree Upgrade] Fix order_with_totals_and_distribution factory by updating shipping fees

luisramos0:
#### What? Why?
Closes #3113 
By fixing the totals in the factory, the test sees an order with correct total of 13 (including the shipping fee of 3) and not just 10 (the line item total).
This makes the test payment amount move from 10 to 13 and, because it covers the final amount of 13, capturing the payment just works.
All credit to @mkllnk for finding out the source of the issue [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/3113#issuecomment-452599450).

#### What should we test?
spec in 3113 should be green (I have confirmed in the build).

PR #3366, opened 2019-01-21
Fix orders controller spec

sauloperez:
#### What? Why?

#### Release notes


spec/controllers/spree/orders_controller_spec.rb
```diff
@@ -168,21 +168,18 @@
...
-      let(:params) { { order: { line_items_attributes: {
-        "0" => {id: line_item1.id, quantity: 1},
-        "1" => {id: line_item2.id, quantity: 0}
-      } } } }
 
       before do
-        Spree::Config.shipment_inc_vat = true
-        Spree::Config.shipping_tax_rate = 0.25
+        allow(Spree::Config).to receive(:shipment_inc_vat) { true }
+        allow(Spree::Config).to receive(:shipping_tax_rate) { 0.25 }
```
luisramos0: this is awesome! no need to use around block and keep original setting and restore it afterwards...

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -130,6 +131,24 @@ def cancel
...
+    @order.updater.update_totals
+    @order.shipment.ensure_correct_adjustment_with_included_tax if @order.shipment
+  end
+
+  # Sets the adjustments to open to perform the block's action and closes them back again
+  def opening_adjustments
+    @order.adjustments.each(&:open)
+    yield
+    @order.adjustments.each(&:close)
+  end
```
luisramos0: maybe a better name with_open_adjustments?

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -130,6 +131,24 @@ def cancel
 
   private
 
+  # Updates the various denormalized total attributes of the order and
+  # recalculates the shipment taxes
+  def update_totals_and_taxes
+    @order.updater.update_totals
+    @order.shipment.ensure_correct_adjustment_with_included_tax if @order.shipment
```
luisramos0: this method will change in spree 2.2 https://github.com/spree/spree/commit/9548723ece02fd58c10e05d6a1bce69fe6b78ece#diff-a3fc88c87f17b31a30219f559940b2eaR258

Do you think order.update_shipping_fees! would work here? Or do you want to get rid of order.update_shipping_fees! ?

luisramos0:
@sauloperez before you leave the adjustments party, maybe you want to have a look at this one: https://github.com/openfoodfoundation/openfoodnetwork/pull/3299

spec/controllers/spree/orders_controller_spec.rb
sauloperez: Yep, I started doing that myself and then I though about this.

sauloperez:
> Maybe we should create a session for the gathering this year to discuss and share knowledge about this area of Spree?

sounds good to me. It's also a rather dark area of the code.

app/controllers/spree/orders_controller_decorator.rb
sauloperez: The actual fix is to execute https://github.com/openfoodfoundation/openfoodnetwork/blob/d8710d283ae4649d7207423f16bc1479a2861223/app/models/spree/shipment_decorator.rb#L10-L16 so that we update the `included_tax` attribute. That is what `shipment.save` was triggering, remember?

I'll create an issue for `#update_shipping_fees!` taking that spree 2.2 into consideration. Thanks for that commit!

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -130,6 +134,10 @@ def cancel
 
   private
 
+  def discard_empty_line_items
+    @order.line_items = @order.line_items.select { |li| li.quantity > 0 }
```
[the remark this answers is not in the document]
sauloperez: I personally avoid mutations but in this case, I think it makes sense.

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -134,6 +131,20 @@ def cancel
...
+  def update_totals_and_taxes
+    @order.updater.update_totals
+    @order.shipment.ensure_correct_adjustment_with_included_tax if @order.shipment
+  end
+
+  # Sets the adjustments to open to perform the block's action and closes them back again
+  def with_open_adjustments
+    @order.adjustments.each(&:open)
+    yield
+    @order.adjustments.each(&:close)
```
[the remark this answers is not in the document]
sauloperez: I'm also afraid of that. I can't tell if they always need to be closed. At least the tests so say but we never know...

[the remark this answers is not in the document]
sauloperez: done

app/controllers/spree/orders_controller_decorator.rb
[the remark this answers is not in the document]
mkllnk: Me too.

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -138,11 +138,19 @@ def update_totals_and_taxes
...
   def with_open_adjustments
+    previous_states = @order.adjustments.each_with_object({}) do |adjustment, hash|
+      hash[adjustment.id] = adjustment.state
+    end
     @order.adjustments.each(&:open)
+
     yield
-    @order.adjustments.each(&:close)
+
+    @order.adjustments.each_with_index do |adjustment, index|
```
mkllnk: - You are not using the `index` here.
- What if something adds an adjustment? We should only restore adjustments that are in our states hash.

sauloperez:
Sorry @mkllnk , I had to revert your mutation as it failed a couple specs. It turns out, that mutating makes `#count` still see 2 record when there is 1, which leads me to think that the mutation doesn't play well with AR's persistence and I think it makes sense.

sauloperez:
The entire orders_controller_spec.rb passing :tada:

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -130,6 +131,36 @@ def cancel
...
+
+    yield
+
+    previous_states.each do |adjustment_id, adjustment_pair|
+      adjustment = adjustment_pair[:adjustment]
+      previous_state = adjustment_pair[:previous_state]
+
+      adjustment.update_attribute(:state, previous_state)
+    end
+  end
```
luisramos0:
```suggestion
  def with_open_adjustments
    previous_states = build_adjustments_state_hash
    @order.adjustments.each(&:open)
    yield
    restore_adjustments_state(previous_states)
  end
  
  def build_adjustments_state_hash
     @order.adjustments.each_with_object({}) do |adjustment, hash|
      hash[adjustment.id] = { adjustment: adjustment, previous_state: adjustment.state }
    end
  end
  
  def restore_adjustments_state(adjustments_state_hash)
    adjustments_state_hash.each do |adjustment_id, adjustment_pair|
      adjustment = adjustment_pair[:adjustment]
      previous_state = adjustment_pair[:previous_state]

      adjustment.update_attribute(:state, previous_state)
    end
  end
```

luisramos0: I'd extract method ;-)

sauloperez: It's hairy I agree. I'd prefer to extract class but in another PR as IMO this is growing too much and chances are it'll take some time. Agree? If so, I'll create an issue.

luisramos0: sure!
if this was my PR I'd extract method and leave it there. extract class is even better but will take more time. it's really up to you.

app/controllers/spree/orders_controller_decorator.rb
sauloperez: Done in https://github.com/openfoodfoundation/openfoodnetwork/issues/3375

app/controllers/spree/orders_controller_decorator.rb
sauloperez: that's right, but when I see the rest of the controller I feel really bad adding a new method...

sauloperez: Done in https://github.com/openfoodfoundation/openfoodnetwork/issues/3376

luisramos0: :+1:

app/controllers/spree/orders_controller_decorator.rb
```diff
@@ -139,17 +139,21 @@ def update_totals_and_taxes
...
 
     yield
 
-    @order.adjustments.each_with_index do |adjustment, index|
-      adjustment.update_attribute(:state, previous_states[adjustment.id])
+    previous_states.each do |adjustment_id, adjustment_pair|
+      adjustment = adjustment_pair[:adjustment]
+      previous_state = adjustment_pair[:previous_state]
+
+      adjustment.update_attribute(:state, previous_state)
```
mkllnk: Now you have to hope that nothing deleted the adjustment or update_attribute will fail. :-) I thought of something like this:

```rb
  def with_open_adjustments
    previous_states = @order.adjustments.each_with_object({}) do |adjustment, hash|
      hash[adjustment.id] = adjustment.state
    end
    @order.adjustments.each(&:open)

    yield

    @order.adjustments.each do |adjustment|
      previous_state = previous_states[adjustment.id]
      adjustment.update_attribute(:state, previous_state) if previous_state
    end
end
```

sauloperez:
I addressed your comment @mkllnk and took the chance to rebase onto `2-0-stable` to get an up-to-date build.

Issue #3121, opened 2018-11-22
[Spree Upgrade][BackOffice] De-deface payment method new and adapt to spree 2 (1 broken test)


luisramos0:
Main config menu is showing up on the right side instead of the list of hubs that can be associated with the payment method.
payment_method_new is defaced. Needs to be de-defaced (in master first).

Errors in features/admin/payment_method_spec.rb:
```
12) 
    As a Super Admin
    I want to be able to set a distributor on each payment method
 updating a payment method
      Failure/Error: uncheck "payment_method_distributor_ids_#{@distributors[0].id}"
      
      Capybara::ElementNotFound:
        Unable to find visible checkbox "payment_method_distributor_ids_96" that is not disabled
      # ./spec/features/admin/payment_method_spec.rb:81:in `block (2 levels) in <top (required)>'

  13) 
    As a Super Admin
    I want to be able to set a distributor on each payment method
 creating a payment method assigning a distributor to the payment method
      Failure/Error: check "payment_method_distributor_ids_#{@distributors[0].id}"
      
      Capybara::ElementNotFound:
        Unable to find visible checkbox "payment_method_distributor_ids_99" that is not disabled
      # ./spec/features/admin/payment_method_spec.rb:24:in `block (3 levels) in <top (required)>'

  14) 
    As a Super Admin
    I want to be able to set a distributor on each payment method
 as an enterprise user creates payment methods
      Failure/Error: check "payment_method_distributor_ids_#{distributor1.id}"
      
      Capybara::ElementNotFound:
        Unable to find visible checkbox "payment_method_distributor_ids_116" that is not disabled
      # ./spec/features/admin/payment_method_spec.rb:147:in `block (3 levels) in <top (required)>'
```

luisramos0:
only one broken spec currently:

```
 10) 
    As a Super Admin
    I want to be able to set a distributor on each payment method
 updating a payment method
      Failure/Error: object.tag_list.join(",")
      
      ActionView::Template::Error:
        undefined method `association_class' for nil:NilClass
      # ./app/serializers/api/admin/payment_method/base_serializer.rb:6:in `tag_list'
      # (eval):6:in `_fast_attributes'
      # ./app/serializers/api/admin/payment_method_serializer.rb:4:in `serializable_hash'
      # ./app/helpers/admin/injection_helper.rb:123:in `admin_inject_json_ams'
[...]

PR #3370, opened 2019-01-23
[Spree upgrade] Workaround Rails inheritance bug in Spree::Gateway

mkllnk:
#### What? Why?

Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/3121

Due to a bug in ActiveRecord we need to load the tagging code in Gateway which should have inherited it from its parent PaymentMethod. We have to call it before loading the PaymentMethod decorator because the tagging code won't load twice within the inheritance chain.

I tested what happens when you assign tags to a PayPalExpress gateway and they are correctly stored with `:taggable_type => "Spree::PaymentMethod"`.

I also considered a tagging implementation that doesn't require decorating, but we would be working against the used gem which assumes to be called on the tagged ActiveRecord model. It may still be an option, but needs deeper study of the acts_as_taggable_on code.

#### What should we test?

If we test this: Create a PayPal payment method from the admin interface.

Please note: the affected `spec/features/admin/payment_method_spec.rb:72` executes further, but does not pass. The following failure means that this pull request is successful:
```
     Failure/Error: expect(page).to have_field 'Password', with: ''
       expected to find visible field "Password" that is not disabled with value "" but there were no matches. Also found "", which matched the selector but not all filters.
     # ./spec/features/admin/payment_method_spec.rb:108:in `block (2 levels) in <top (required)>'
```

This is part of the Spree upgrade.

### DIFF
--- a/config/initializers/spree.rb
+++ b/config/initializers/spree.rb
@@ -8,6 +8,16 @@
 
 require 'spree/product_filters'
 
+# Due to a bug in ActiveRecord we need to load the tagging code in Gateway which
+# should have inherited it from its parent PaymentMethod.
+# We have to call it before loading the PaymentMethod decorator because the
+# tagging code won't load twice within the inheritance chain.
+# https://github.com/openfoodfoundation/openfoodnetwork/issues/3121
+Spree::Gateway.class_eval do
+  acts_as_taggable
+  attr_accessible :tag_list
+end
+
 require "#{Rails.root}/app/models/spree/payment_method_decorator"
 require "#{Rails.root}/app/models/spree/gateway_decorator"
 
--- a/spec/models/spree/gateway_tagging_spec.rb
+++ b/spec/models/spree/gateway_tagging_spec.rb
@@ -1,0 +1,27 @@
+require "spec_helper"
+
+# An inheritance bug made these specs fail.
+# See config/initializers/spree.rb
+shared_examples "taggable" do |parameter|
+  it "uses the given parameter" do
+    expect(subject.tag_list).to eq []
+  end
+end
+
+module Spree
+  describe PaymentMethod do
+    it_behaves_like "taggable"
+  end
+
+  describe Gateway do
+    it_behaves_like "taggable"
+  end
+
+  describe Gateway::PayPalExpress do
+    it_behaves_like "taggable"
+  end
+
+  describe Gateway::StripeConnect do
+    it_behaves_like "taggable"
+  end
+end
--- a/spec/models/spree/gateway_tagging_spec.rb
+++ b/spec/models/spree/gateway_tagging_spec.rb
@@ -1,27 +1,56 @@
 require "spec_helper"
 
-# An inheritance bug made these specs fail.
-# See config/initializers/spree.rb
-shared_examples "taggable" do |parameter|
-  it "uses the given parameter" do
+# We extended Spree::PaymentMethod to be taggable. Unfortunately, an inheritance
+# bug prevented the taggable code to be passed on to the descendants of
+# PaymentMethod. We fixed that in config/initializers/spree.rb.
+#
+# This spec tests several descendants for their taggability. The tests are in
+# a separate file, because they cover one aspect of several classes.
+shared_examples "taggable" do |expected_taggable_type|
+  it "provides a tag list" do
     expect(subject.tag_list).to eq []
+  end
+
+  it "stores tags for the root taggable type" do
+    subject.tag_list.add("one")
+    subject.save!
+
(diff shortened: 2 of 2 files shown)


mkllnk:
I created a new issue #3379. The old one became a bit long and the agenda changed already.

mkllnk:
Semaphore failed at another point in the spec:
```
     Failure/Error: expect(first('tags-input .tag-list ti-tag-item')).to have_content "member"
       expected to find text "member" in ""
     # ./spec/features/admin/payment_method_spec.rb:94:in `block (2 levels) in <top (required)>'
```

That has to do with tags, so I should double-check my PR.

mkllnk:
Ah, I needed to include `attr_accessible :tag_list` as well. I'm still wondering if there is a better solution to implement tags without modifying Spree models...

luisramos0:
this PR fixes the same problem in 3 specs in spec/features/admin/subscriptions_spec.rb 
 :+1:

spec/models/spree/gateway_tagging_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+# See config/initializers/spree.rb
+shared_examples "taggable" do |parameter|
+  it "uses the given parameter" do
+    expect(subject.tag_list).to eq []
+  end
+end
+
+module Spree
+  describe PaymentMethod do
+    it_behaves_like "taggable"
```
sauloperez: why not moving this to the payment_method_spec.rb? That is where I'd go check that if I had to.

spec/models/spree/gateway_tagging_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  describe Gateway::PayPalExpress do
+    it_behaves_like "taggable"
+  end
+
+  describe Gateway::StripeConnect do
+    it_behaves_like "taggable"
+  end
+end
```
sauloperez: The same for the other models.

config/initializers/spree.rb
```diff
@@ -8,6 +8,16 @@
...
 
+# Due to a bug in ActiveRecord we need to load the tagging code in Gateway which
+# should have inherited it from its parent PaymentMethod.
+# We have to call it before loading the PaymentMethod decorator because the
+# tagging code won't load twice within the inheritance chain.
+# https://github.com/openfoodfoundation/openfoodnetwork/issues/3121
+Spree::Gateway.class_eval do
+  acts_as_taggable
+  attr_accessible :tag_list
+end
```
sauloperez: but if essentially this is a order-related bug doesn't a `require 'spree/payment_method'` in the appropriate place helps us here? apart from that, from https://github.com/rails/rails/issues/3847 looks like some gem could be messing things up as well.

mkllnk:
@kristinalim, thank you for nudging me to increase the test coverage. I hesitated because it makes everything more complicated. It actually took me several hours to write a spec I'm content with. The difficult trade-off was between a flexible shared example that works with any payment method and using our factories. Let me know what you think about the new commit.

spec/models/spree/gateway_tagging_spec.rb
mkllnk: This spec is actually not about the PaymentMethod but about its descendants. Since I'm testing several classes for the same thing I do it in one place instead of scattering these tests in the spec files of every descendant. We are actually testing Ruby's inheritance here which doesn't belong to one specific class.

mkllnk: I added more comments to explain this so that the next developer doesn't have the same question. Thank you for prompting. :+1: :-)

config/initializers/spree.rb
mkllnk: Yes, it is order-related. Yes, it could be a gem. It could even be Spree.

A `require 'spree/payment_method'` doesn't help, because the inheritance already happened when we access the Spree code. I tried that in many places, for hours, after Luis tried that in many places, for hours. You can have another go and maybe you can find a better solution. But we spent several dev days on this already and couldn't find any better solution.

Since this bug is hopefully fixed in the next version of Rails we shouldn't invest much more time.

One experiment I would support as a tech-debt task: Instead of adding tags to `Spree::PaymentMethod`, we could try to have a new class called `TaggablePaymentMethod` which basically decorates any other payment method. All code using tagging interface would use that class instead of using the payment method instance directly.

What do you think? Should I create an issue for that?

sauloperez: It's fine. It doesn't make sense to spend any more time on this. I wouldn't add the tech debt task either for now.

sauloperez:
Waiting for @kristinalim 's answer to merge this.

luisramos0:
🎉

Issue #2792, opened 2018-09-29
[Spree Upgrade][Reports] Fix Reports specs (13 broken specs)


luisramos0:
spec/features/admin/reports_spec and spec/controllers/spree/admin/reports_controller_spec are badly broken.
These specs were made pending but need to be fixed at some point.

See #2797 and #2902

luisramos0:
🎉

PR #3371, opened 2019-01-23
[Spree Upgrade] Fix xero invoices specs in reports_spec

luisramos0:
#### What? Why?

Part of  #2792 
This spec was adding the shipping adjustment manually but the adjustment is now already there as the callbacks do the job when the line items are added to the order. The adjustment the callbacks add is exactly the same as the one that was injected manually before, to get that we just need to set the included_tax config option.

#### What should we test?
3 specs fixed in Job 1 - spec/features/admin/reports_spec.rb:434
Only one spec missing in reports_spec.rb but it is related to account invoices.

### DIFF
--- a/spec/features/admin/reports_spec.rb
+++ b/spec/features/admin/reports_spec.rb
@@ -410,115 +410,119 @@ xfeature %q{
     let(:product1) { create(:taxed_product, zone: zone, price: 12.54, tax_rate_amount: 0, sku: 'sku1') }
     let(:product2) { create(:taxed_product, zone: zone, price: 500.15, tax_rate_amount: 0.2, sku: 'sku2') }
 
-    let!(:line_item1) { create(:line_item, variant: product1.master, price: 12.54, quantity: 1, order: order1) }
-    let!(:line_item2) { create(:line_item, variant: product2.master, price: 500.15, quantity: 3, order: order1) }
-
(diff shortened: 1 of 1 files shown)


spec/features/admin/reports_spec.rb
```diff
@@ -410,115 +410,119 @@
...
-      visit spree.admin_reports_path
-
-      click_link 'Xero Invoices'
-    end
-
-    around do |example|
-      Timecop.travel(Time.zone.local(2015, 4, 26, 14, 0, 0)) do
-        example.run
-      end
+      Spree::Config.shipment_inc_vat = true
```
mkllnk: Doesn't this alter the config for other specs as well? Can you mock it or reset it after the test with an `around` block?

luisramos0: yeah, thanks. I switched to mocks, it works!

spec/features/admin/reports_spec.rb
```diff
@@ -198,8 +198,8 @@
     let!(:adj_shipping) { create(:adjustment, adjustable: order1, label: "Shipping", amount: 100.55) }
 
     before do
-      Spree::Config.shipment_inc_vat = true
-      Spree::Config.shipping_tax_rate = 0.2
+      allow(Spree::Config).to receive(:shipment_inc_vat) { true }
+      allow(Spree::Config).to receive(:shipping_tax_rate) { 0.2 }
```
kristinalim: Great! The mocked version is much better.

PR #3380, opened 2019-01-24
Use patch to hide Paypal gateway password

### LINKED ISSUE #3379: [Spree Upgrade][BackOffice] Payment method (1 broken test)
Follow up from #3121.

```
bundle exec rspec spec/features/admin/payment_method_spec.rb:72
```

     Failure/Error: expect(page).to have_field 'Password', with: ''
       expected to find visible field "Password" that is not disabled with value "" but there were no matches. Also found "", which matched the selector but not all filters.
     # ./spec/features/admin/payment_method_spec.rb:108:in `block (2 levels) in <top (required)>'

mkllnk:
#### What? Why?

Closes #3379

When we updated our Gemfile for the Spree upgrade, we chose Spree's original 2-0-stable branch of better_spree_paypal_express. We would like to keep our two patches though. One of the two patches changes the password field to the type `password` and therefore hiding the password on the screen. 

#### What should we test?

`./spec/features/admin/payment_method_spec.rb:72` should pass. No user testing required.

#### Dependencies

This depends on #3370 to make the spec pass.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ gem 'spree_auth_devise', github: 'spree/spree_auth_devise', branch: '2-0-stable'
 # Our branch contains two changes
 # - Pass customer email and phone number to PayPal (merged to upstream master)
 # - Change type of password from string to password to hide it in the form
-gem 'spree_paypal_express', github: "spree-contrib/better_spree_paypal_express", branch: "2-0-stable"
+gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "2-0-stable"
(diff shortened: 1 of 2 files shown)


luisramos0:
for some reason this has not triggered a semaphore build. Maybe you can push again to trigger it?

Matt-Yorkley:
`./spec/features/admin/payment_method_spec.rb:72` fails with: 
```
ActionView::Template::Error:
   undefined method `association_class' for nil:NilClass
```

luisramos0:
This is good to go. The spec is only fixed with #3370 (see dependencies in the description). I am merging this one now.

Issue #3384, opened 2019-01-24
Product import crashing unexpectedly upon data upload


sineadfenton:
## Description

Product upload sporadically breaks

## Expected Behavior

Upload data and it processes and tells you how many 

## Actual Behavior

The data import appears to only be able to handle so much data - to manage this we've breaking up the upload into smaller chunks of about 50-100 rows at a time. 

We've noticed when doing this the page crashes sporadically and appears to only upload so many of the rows - so to check what's in I need to reupload the file, and see how many will be updated and how many still need to be created, on other occasions all the rows import but it's not clear, and other times it's totally fine.

This happens even after the import tells you to rectitfy errors etc.

## Steps to Reproduce

1. Product import - https://drive.google.com/open?id=19Z6M56Yq1w_Rz_TBa7cYbSX5hjbQuy64
2. It starts to load then crashes
3.
4.

## Animated Gif/Screenshot

http://recordit.co/VsiuEc6PbT

## Context

This is pretty time consuming and we have a potential client where we'll need to upload and update 7000 products.

## Severity

Part of the product import work - maybe s1?

## Your Environment

* Version used:
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): Windows

sineadfenton:
I will flag this up in the slack channel following our discussion @lin-d-hop

kristinalim:
The network panel says `save_data` was getting HTTP 504, confirming that this is a timeout issue.

I noticed two simultaneous `save_data` HTTP requests in the Network panel in the screen recording, and learned that product import processes the CSV in batches of 50. So if there are 200 rows, the browser starts four simultaneous HTTP requests to the server: one to process rows 1-50, another to process rows 51-100, and so on. I suppose this is to keep the server from processing too large a batch and then have the HTTP request time out.

[UK has two Rails instances](https://github.com/openfoodfoundation/ofn-install/blob/master/inventory/group_vars/all.yml#L95) (through Unicorn workers) and can process two HTTP requests at a time. Because these HTTP requests happen simultaneously, requests just start early but still wait for other requests (whether for the current import or other users) to finish and for a Rails instance to free up. Processing of a batch takes a while, so HTTP requests for other batches are likely to time out.

I'll make this start the HTTP request for processing next batches only after the previous batch has completed. A con of this approach is that the user needs to keep their browser open for later parts of the large CSV to be processed, but I think this is better than higher chance of requests timing out.

[...]

PR #3233, opened 2018-12-20
Pi/inventories

Matt-Yorkley:
#### What? Why?

Closes #3116 
Closes #3160

Updated the specs and CSV template for importing into inventories.

#### What should we test?

Product Import: importing into inventory. Check import to inventory works correctly, and existing inventory items set to hidden should be set to visible when updated via import.

#### Release notes

Updated inventory CSV template for Product Import.

Changelog Category: Changed

#### Documentation updates

The template has changed for inventory CSV.

PR #3387, opened 2019-01-24
[Product Import] Process only one batch of rows at a time

kristinalim:
#### What? Why?

Addresses #3384 

There doesn't seem to be JS or feature tests for the batch processing in particular. I'll work on adding these tests in a separate PR.

#### What should we test?

Do this for product import and inventory import:

1. Prepare a CSV file with more than 120 lines.
2. (For Chrome) Open the Network panel of Developer Tools. Filter results to `_data`: We want to see requests for `validate_data` and `save_data`.
3. Upload the file.
4. When the file is being validated, there should be 3 HTTP requests for `validate_data` that are not simultaneous but happen one at a time.
5. When the file is being saved, there should be 3 HTTP requests for `save_data` that happen also one at a time.
6. Test a random row from each batch of 50. The row should have been processed successfully.
7. Introduce an invalid row to each batch, and upload the file.
8. The validation errors for each of these invalid rows should show.

#### Release notes

- Change processing by batch of rows in product import and inventory import CSV files to happen one batch after another, not simultaneous.

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/admin/product_import/controllers/import_form_controller.js.coffee
+++ b/app/assets/javascripts/admin/product_import/controllers/import_form_controller.js.coffee
@@ -10,6 +10,7 @@ angular.module("admin.productImport").controller "ImportFormCtrl", ($scope, $htt
   $scope.updated_ids = []
   $scope.update_errors = []
 
+  $scope.batchSize = 50
   $scope.step = 'settings'
   $scope.chunks = 0
   $scope.completed = 0
@@ -51,14 +52,13 @@ angular.module("admin.productImport").controller "ImportFormCtrl", ($scope, $htt
   $scope.start = () ->
     $scope.started = true
     total = ams_data.item_count
-    size = 50
-    $scope.chunks = Math.ceil(total / size)
+    $scope.chunks = Math.ceil(total / $scope.batchSize)
 
     i = 0
 
     while i < $scope.chunks
-      start = (i*size)+1
-      end = (i+1)*size
+      start = (i * $scope.batchSize) + 1
+      end = (i + 1) * $scope.batchSize
       if $scope.step == 'import'
         $scope.processImport(start, end)
       if $scope.step == 'save'
--- a/app/assets/javascripts/admin/product_import/controllers/import_form_controller.js.coffee
+++ b/app/assets/javascripts/admin/product_import/controllers/import_form_controller.js.coffee
@@ -54,16 +54,24 @@ angular.module("admin.productImport").controller "ImportFormCtrl", ($scope, $htt
     total = ams_data.item_count
     $scope.chunks = Math.ceil(total / $scope.batchSize)
 
-    i = 0
+    # Process only the first batch.
+    $scope.processBatch($scope.step, 0, $scope.chunks)
 
-    while i < $scope.chunks
-      start = (i * $scope.batchSize) + 1
-      end = (i + 1) * $scope.batchSize
-      if $scope.step == 'import'
-        $scope.processImport(start, end)
-      if $scope.step == 'save'
-        $scope.processSave(start, end)
-      i++
+  $scope.processBatch = (step, batchIndex, batchCount) ->
+    start = (batchIndex * $scope.batchSize) + 1
+    end = (batchIndex + 1) * $scope.batchSize
+    withNextBatch = batchIndex + 1 < batchCount
+
+    promise = if step == 'import'
+      $scope.processImport(start, end)
+    else if step == 'save'
+      $scope.processSave(start, end)
+
+    processNextBatch = ->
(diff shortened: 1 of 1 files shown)


kristinalim:
**UPDATE:** I already added the feature test to this PR, and decided not to add a JS test anymore. I removed the note about these in the issue description.

app/assets/javascripts/admin/product_import/controllers/import_form_controller.js.coffee
```diff
@@ -51,19 +52,26 @@ angular.module("admin.productImport").controller "ImportFormCtrl", ($scope, $htt
...
+    promise = if step == 'import'
+      $scope.processImport(start, end)
+    else if step == 'save'
+      $scope.processSave(start, end)
+
+    processNextBatch = ->
+      $scope.processBatch(step, batchIndex + 1, batchCount)
+
+    # Process next batch whether or not processing of the current batch succeeds.
+    promise.then(processNextBatch, processNextBatch) if withNextBatch
```
luisramos0: I felt like changing to isLastBatch and return early, please ignore if you prefer your version.
```suggestion
     $scope.processBatch = (step, batchIndex, batchCount) ->
    startRecord = (batchIndex * $scope.batchSize) + 1
    endRecord = (batchIndex + 1) * $scope.batchSize
    isLastBatch = batchCount < batchIndex + 1

    promise = if step == 'import'
      $scope.processImport(startRecord, endRecord)
    else if step == 'save'
      $scope.processSave(startRecord, endRecord)

    return if isLastBatch
    
    # Process next batch whether or not processing of the current batch succeeds.
    processNextBatch = ->
      $scope.processBatch(step, batchIndex + 1, batchCount)
    promise.then(processNextBatch, processNextBatch)
```

kristinalim:
> this can have a bad impact on UX, user will have to wait...
> it all depends how quick the endpoint is returning.
>
> one alternative would be to write something like limited parallel calls to the server, for example limiting to 2 the number of processes running.

Without processing rows in batches, the default is there would only be one instance performing the import for the user, so I think the impact on UX is acceptable. Unless splitting into batches was done also to reduce waiting time of the users, @Matt-Yorkley, and not just to avoid timeout?

But actually @luisramos0, I prefer that users importing data will wait longer if that means it leaves alone 1 of the 2 Rails instances to process requests of other users.

@luisramos0 Let me know if you still want to explore having max 2 import requests at a time. I will create an issue for that if yes.

app/assets/javascripts/admin/product_import/controllers/import_form_controller.js.coffee
kristinalim: I made the changes using `isLastBatch` instead and then returning early: https://github.com/openfoodfoundation/openfoodnetwork/pull/3387/commits/b6d8611359ed2cd701ee7997f095487c4bdcba92

Thought to note that I used `==` in the expression:

```
isLastBatch = batchCount == batchIndex + 1
```

Thanks, @luisramos0! :slightly_smiling_face:

Matt-Yorkley:
@kristinalim it was just to avoid timeouts, I think waiting is fine. :+1:

RachL:
deploying on UK staging

RachL:
@kristinalim when testing what happened with empty on_demand cells for 2 lines on a second batch, I had an error at the saving stage, which prevented the 2 products to be imported. Is this the correct behaviour? The other type of errors I created where catch before importing.

Notes and file: https://docs.google.com/document/d/1muvvMKkA0dfUxNS75VxavsFudFXorVcojbLCrZW6Msc/edit

kristinalim:
Interesting, @RachL. Sounds like a separate bug. Let me know in case you want me to create the issue.

Just checking: There wasn't accidentally a space " " in the `on_hand` column, was there? That's the easiest possible explanation. I will attempt to reproduce later.

myriamboure:
About on_demand some explanation of expected behavior is in the other PR. As the other PR is not tested and merged, maybe it is normal that it is still bugging in this one ?

kristinalim:
@myriamboure I doubt that #3233 will fix this, because that is mostly for the inventory (variant overrides) while this is for the product and variant import. Variants and variant overrides also don't share the same code for validating `count_on_hand` and `on_demand` (variants do not have "Use producer stock settings").

And yes, noted about how `on_demand` and `on_hand` should be processed.

RachL:
@kristinalim I'm not sure I'm seeing a space in my file :( So we go forward with this PR and open an new issue for this case? ping @myriamboure

myriamboure:
If it improves the feature, as it is still invisible, yes let's merge, and open a new issue to fix it separately.

myriamboure:
When we identify clearly the issue we can decide if it blocks the release of the feature or not.

PR #3394, opened 2019-01-25
[Spree upgrade] 3021 update soft delete

### LINKED ISSUE #3021: [Spree Upgrade] Fix products_and_inventory_report pending spec (1 broken spec)
Fix spec spec/lib/open_food_network/products_and_inventory_report_spec.rb
moved to xdescribe in #3019

Job 4 - 1 x spec/lib/open_food_network/products_and_inventory_report_spec.rb:105
```
2) OpenFoodNetwork::ProductsAndInventoryReport As an enterprise user Filtering variants should filter deleted products
     Failure/Error: subject.filter(Spree::Variant.scoped).should match_array [product1.master, product1.variants.first]
     
       expected collection contained:  [#<Spree::Variant id: 2329, sku: "ABC", weight: #<BigDecimal:7f13b76be780,'0.1E-2',9(36)>, height: ni..."AUD", unit_value: 1.0, unit_description: "", display_name: nil, display_as: nil, import_date: nil>]
       actual collection contained:    [#<Spree::Variant id: 2329, sku: "ABC", weight: #<BigDecimal:7f13b76b3c40,'0.0',9(18)>, height: nil, ..."AUD", unit_value: 1.0, unit_description: "", display_name: nil, display_as: nil, import_date: nil>]
       the extra elements were:        [#<Spree::Variant id: 2331, sku: "ABC", weight: #<BigDecimal:7f13b76aaeb0,'0.0',9(18)>, height: nil, ..."AUD", unit_value: 1.0, unit_description: "", display_name: nil, display_as: nil, import_date: nil>]
     # ./spec/lib/open_food_network/products_and_inventory_report_spec.rb:109:in `block (4 levels) in <module:OpenFoodNetwork>'
```

this test
```
        it "should filter deleted products" do
          product1 = create(:simple_product, supplier: supplier)
          product2 = create(:simple_product, supplier: supplier)
          product2.delete
          subject.filter(Spree::Variant.scoped).should match_array [product1.master, product1.variants.first]
        end
```

product2.delete is not removing product2's variants from the scope as expected.

mkllnk:
#### What? Why?

Closes #3021

Update our code to use Spree's changed soft-delete API of models. I searched our code for every appearance of the word `delete` and dealt with all delete calls to products and variants. We didn't call delete for other models. https://github.com/openfoodfoundation/spree/commit/cd3add960ef04f5881b508a07a503b3f7cade883

There may be cases in which we were calling `destroy` before, assuming that the product or variant got really destroyed. That won't happen any more. I'm not aware of these cases and I guess we can live with that. Searching our code for every `destroy` would be too much.

There may also be a query for all variants or products that assumes to get deleted ones as well. Since deleted objects are now filtered in the default scope, we won't get them unless we append `with_deleted`. I changed one controller to deal with that. I'm not sure if there could be more.

#### What should we test?

- Create products and variants.
- Delete products and variants.

### DIFF
--- a/app/controllers/spree/admin/products_controller_decorator.rb
+++ b/app/controllers/spree/admin/products_controller_decorator.rb
@@ -78,8 +78,13 @@ Spree::Admin::ProductsController.class_eval do
     params[:q][:deleted_at_null] ||= "1"
 
     params[:q][:s] ||= "name asc"
-
-    @search = Spree::Product.ransack(params[:q]) # this line is modified - hit Spree::Product instead of super, avoiding cancan error for fetching records with block permissions via accessible_by
+    # The next line is modified.
+    # Hit Spree::Product instead of super, avoiding cancan error for fetching
+    # records with block permissions via accessible_by.
+    @collection = Spree::Product
+    @collection = @collection.with_deleted if params[:q].delete(:deleted_at_null).blank?
+    # @search needs to be defined as this is passed to search_form_for
+    @search = @collection.ransack(params[:q])
     @collection = @search.result.
       managed_by(spree_current_user). # this line is added to the original spree code!!!!!
       group_by_products_id.
--- a/app/controllers/spree/api/products_controller_decorator.rb
+++ b/app/controllers/spree/api/products_controller_decorator.rb
@@ -56,8 +56,8 @@ Spree::Api::ProductsController.class_eval do
   def product_scope
     if current_api_user.has_spree_role?("admin") || current_api_user.enterprises.present? # This line modified
       scope = Spree::Product
-      unless params[:show_deleted]
-        scope = scope.not_deleted
+      if params[:show_deleted]
+        scope = scope.with_deleted
       end
     else
       scope = Spree::Product.active
--- a/app/controllers/spree/api/products_controller_decorator.rb
+++ b/app/controllers/spree/api/products_controller_decorator.rb
@@ -33,7 +33,7 @@ Spree::Api::ProductsController.class_eval do
     authorize! :delete, Spree::Product
     @product = find_product(params[:product_id])
     authorize! :delete, @product
-    @product.delete
+    @product.destroy
     respond_with(@product, :status => 204)
   end
 
--- a/app/models/spree/product_decorator.rb
+++ b/app/models/spree/product_decorator.rb
@@ -202,20 +202,18 @@ Spree::Product.class_eval do
     end
   end
 
-  def delete_with_delete_from_order_cycles
+  def destroy_with_delete_from_order_cycles
     transaction do
       OpenFoodNetwork::ProductsCache.product_deleted(self) do
-        # Touch supplier and distributors as we would on #destroy
-        self.supplier.touch
         touch_distributors
 
         ExchangeVariant.where('exchange_variants.variant_id IN (?)', self.variants_including_master.with_deleted).destroy_all
 
-        delete_without_delete_from_order_cycles
+        destroy_without_delete_from_order_cycles
       end
     end
   end
-  alias_method_chain :delete, :delete_from_order_cycles
+  alias_method_chain :destroy, :delete_from_order_cycles
 
 
   def refresh_products_cache
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -105,14 +105,6 @@ Spree::Variant.class_eval do
     OpenFoodNetwork::EnterpriseFeeCalculator.new(distributor, order_cycle).fees_by_type_for self
   end
 
-  def delete
-    transaction do
-      self.update_column(:deleted_at, Time.zone.now)
-      ExchangeVariant.where(variant_id: self).destroy_all
-      self
-    end
-  end
-
   def refresh_products_cache
     if is_master?
       product.refresh_products_cache
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -35,7 +35,6 @@ Spree::Variant.class_eval do
 
   scope :with_order_cycles_inner, joins(exchanges: :order_cycle)
 
-  scope :not_deleted, where(deleted_at: nil)
   scope :not_master, where(is_master: false)
   scope :in_order_cycle, lambda { |order_cycle|
     with_order_cycles_inner.
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -106,15 +106,10 @@ Spree::Variant.class_eval do
   end
 
   def delete
-    if product.variants == [self] # Only variant left on product
-      errors.add :product, I18n.t(:spree_variant_product_error)
-      false
-    else
-      transaction do
-        self.update_column(:deleted_at, Time.zone.now)
-        ExchangeVariant.where(variant_id: self).destroy_all
-        self
-      end
+    transaction do
+      self.update_column(:deleted_at, Time.zone.now)
+      ExchangeVariant.where(variant_id: self).destroy_all
+      self
     end
   end
 
--- a/app/services/variant_deleter.rb
+++ b/app/services/variant_deleter.rb
@@ -6,7 +6,7 @@ class VariantDeleter
       return false
     end
 
-    variant.delete
+    variant.destroy
   end
 
   private
--- a/app/services/variant_deleter.rb
+++ b/app/services/variant_deleter.rb
@@ -1,0 +1,17 @@
+# Checks the validity of a soft-delete call.
+class VariantDeleter
+  def delete(variant)
+    if only_variant_on_product?(variant)
+      variant.errors.add :product, I18n.t(:spree_variant_product_error)
+      return false
+    end
+
+    variant.delete
+  end
+
+  private
+
+  def only_variant_on_product?(variant)
+    variant.product.variants == [variant]
+  end
+end
--- a/lib/open_food_network/products_and_inventory_report_base.rb
+++ b/lib/open_food_network/products_and_inventory_report_base.rb
@@ -31,11 +31,7 @@ module OpenFoodNetwork
     end
 
     def filter(variants)
-      filter_on_hand filter_to_distributor filter_to_order_cycle filter_to_supplier filter_not_deleted variants
-    end
-
-    def filter_not_deleted(variants)
-      variants.not_deleted
+      filter_on_hand filter_to_distributor filter_to_order_cycle filter_to_supplier variants
     end
 
     # Using the `in_stock?` method allows overrides by distributors.
--- a/app/controllers/spree/admin/variants_controller_decorator.rb
+++ b/app/controllers/spree/admin/variants_controller_decorator.rb
@@ -10,8 +10,11 @@ Spree::Admin::VariantsController.class_eval do
 
   def destroy
     @variant = Spree::Variant.find(params[:id])
-    @variant.delete # This line changed, as well as removal of following conditional
-    flash[:success] = I18n.t('notice_messages.variant_deleted')
+    if VariantDeleter.new.delete(@variant) # This line changed
+      flash[:success] = Spree.t('notice_messages.variant_deleted')
+    else
+      flash[:success] = Spree.t('notice_messages.variant_not_deleted')
+    end
 
     respond_with(@variant) do |format|
       format.html { redirect_to admin_product_variants_url(params[:product_id]) }
--- a/app/controllers/spree/api/variants_controller_decorator.rb
+++ b/app/controllers/spree/api/variants_controller_decorator.rb
@@ -3,7 +3,7 @@ Spree::Api::VariantsController.class_eval do
     @variant = scope.find(params[:variant_id])
     authorize! :delete, @variant
 
-    @variant.delete
+    VariantDeleter.new.delete(@variant)
     respond_with @variant, status: 204
   end
 end
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -146,7 +146,6 @@ class OrderCycle < ActiveRecord::Base
     Spree::Variant.
       joins(:exchanges).
       merge(Exchange.in_order_cycle(self)).
-      not_deleted.
       select('DISTINCT spree_variants.*').
       to_a # http://stackoverflow.com/q/15110166
   end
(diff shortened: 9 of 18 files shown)


app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -78,8 +78,13 @@ def collection
     params[:q][:deleted_at_null] ||= "1"
 
     params[:q][:s] ||= "name asc"
-
-    @search = Spree::Product.ransack(params[:q]) # this line is modified - hit Spree::Product instead of super, avoiding cancan error for fetching records with block permissions via accessible_by
+    # The next line is modified.
+    # Hit Spree::Product instead of super, avoiding cancan error for fetching
+    # records with block permissions via accessible_by.
+    @collection = Spree::Product
+    @collection = @collection.with_deleted if params[:q].delete(:deleted_at_null).blank?
```
sauloperez: why is this needed now? AFAIK ransack's [deleted_at_null predicate](https://github.com/activerecord-hackery/ransack/wiki/Basic-Searching#null) should already take the `deleted_at` column into account. From what I understand, if `:deleted_at_null` is 1, it'll fetch records where `deleted_at` is null, while if it's 0 it will fetch them all.

app/services/variant_deleter.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    variant.destroy
+  end
+
+  private
+
+  def only_variant_on_product?(variant)
+    variant.product.variants == [variant]
+  end
+end
```
sauloperez: Are we sure Spree doesn't deal with this case already? In any case, it feels to me this fits better in a model validation, which I suspect is already implemented by Spree.

app/controllers/spree/api/products_controller_decorator.rb
```diff
@@ -33,7 +33,7 @@ def soft_delete
     authorize! :delete, Spree::Product
     @product = find_product(params[:product_id])
     authorize! :delete, @product
-    @product.delete
+    @product.destroy
     respond_with(@product, :status => 204)
   end
```
sauloperez: can we create an issue to refactor this `#soft_delete`? Now that Spree has it we don't need a separate soft delete implementation. The frontend shouldn't care whether the delete is soft or not, IMO. It's an implementation detail.

app/models/spree/product_decorator.rb
```diff
@@ -202,20 +202,18 @@ def variant_unit_option_type
...
 
         ExchangeVariant.where('exchange_variants.variant_id IN (?)', self.variants_including_master.with_deleted).destroy_all
 
-        delete_without_delete_from_order_cycles
+        destroy_without_delete_from_order_cycles
       end
     end
   end
-  alias_method_chain :delete, :delete_from_order_cycles
+  alias_method_chain :destroy, :delete_from_order_cycles
```
sauloperez: 🙈 first time I see this.

app/models/spree/variant_decorator.rb
```diff
@@ -99,19 +98,6 @@ def fees_by_type_for(distributor, order_cycle)
...
   end
 
-  def delete
-    if product.variants == [self] # Only variant left on product
-      errors.add :product, I18n.t(:spree_variant_product_error)
-      false
-    else
-      transaction do
-        self.update_column(:deleted_at, Time.zone.now)
-        ExchangeVariant.where(variant_id: self).destroy_all
```
sauloperez: what happens with these exchange variants then? Are they destroyed on cascade somewhere else?

lib/open_food_network/products_and_inventory_report_base.rb
```diff
@@ -31,11 +31,7 @@ def child_variants
     end
 
     def filter(variants)
-      filter_on_hand filter_to_distributor filter_to_order_cycle filter_to_supplier filter_not_deleted variants
-    end
-
-    def filter_not_deleted(variants)
-      variants.not_deleted
+      filter_on_hand filter_to_distributor filter_to_order_cycle filter_to_supplier variants
```
sauloperez: how is this ever going to be fast... 😓

app/controllers/spree/admin/products_controller_decorator.rb
mkllnk: This is Spree code from radar's commit https://github.com/openfoodfoundation/spree/commit/cd3add960ef04f5881b508a07a503b3f7cade883#diff-08c8984d7cc18aa24712d5b26ec7defbR88.

The difference in Spree 2 is that `deleted_at is null` is now a default scope. If the ransack parameter `:deleted_at_null` is `true`, it will append the same condition again (not having any effect. And if `:deleted_at_null` is `false`, ransack will not modify the query and it has no effect either. The other solution would be to change the UI to send a `deleted_at_not_null` parameter. But I just went with original Spree code here.

app/services/variant_deleter.rb
mkllnk: It is not implemented by Spree. In Spree, variants are optional. If a product doesn't have variants, it always has the hidden master variant holding the data of the product.

In OFN, we always have variants and we ignore the master variant. But if the model enforces that a product needs at least one variant, we can't delete a product, because we can't delete the last variant. We would need to delete without callbacks, but that could miss other important callbacks.

If there was a good way of implementing it on the model, I would go for it. Do you have any ideas? One option I was thinking of is `validates variants, presence: true` on the product. But the product is Spree code and that would break a lot of other cases. That's why I thought it's better to implement this outside of Spree and leave Spree code work as it is.

app/controllers/spree/api/products_controller_decorator.rb
mkllnk: Done: #3405

app/models/spree/variant_decorator.rb
mkllnk: Yes, they are destroyed as dependents. I added a spec for that.

mkllnk:
@sauloperez Thank you a lot for you thorough review. :+1: Let me know what you think about my replies and if there is some documentation I could add so that future devs don't have these questions.

app/services/variant_deleter.rb
sauloperez: Do you know why we chose to ignore the master and always have a variant?

My ideas was to move that logic you implemented to a custom validation method in the variant decorator.

Also, I restriction like this one should be constrainted at the DB level as well, but I don't see how now.

sauloperez:
Yes, that's a good point. Let see of you can find how to doc these things a bit. That will be helpful.

app/services/variant_deleter.rb
mkllnk: > Do you know why we chose to ignore the master and always have a variant?

I'm not sure. I just remember that it was confusing to have normal variants and a master variant that is only active if there is no other variant. I think earlier in the project they decided to hide the master variant and deal only with real variants.

> My ideas was to move that logic you implemented to a custom validation method in the variant decorator.

Do you know if a validation on variants can distinguish if the destroy is only for the current variant or if the product is going to be destroyed as well? If we can't distinguish between these cases, we can't implement it on the model. The destroy is valid when the product is destroyed, but the destroy is invalid if the product persists and the last variant is to be destroyed.

mkllnk:
@sauloperez Are you happy with this pull request?

app/services/variant_deleter.rb
sauloperez: > The destroy is valid when the product is destroyed, but the destroy is invalid if the product persists and the last variant is to be destroyed.

can't we simply check these conditions in the validation method? Did you find any problem by calling `product.variants == [variant]` from a `before_destroy`?

sauloperez:
Sorry, I had pretty busy and exhausting days

app/services/variant_deleter.rb
sauloperez: There has to be one. We have to think harde.r

app/models/spree/product_decorator.rb
kristinalim: The loss of this data has implications in the Enterprise Fee Summary report. :worried: I'll create an issue.

app/services/variant_deleter.rb
kristinalim: One option... In a `before_destroy` callback of `Spree::Product`, we could set an instance variable in `Spree::Variant` to true (e.g. `@destroy_through_product` or `@skip_last_variant_validation`). Then, in `Spree::Variant`, we could skip the last variant validation if this is true.

There's [`marked_for_destruction?`](https://apidock.com/rails/v4.2.7/ActiveRecord/AutosaveAssociation/marked_for_destruction%3F) and [`destroyed_by_association`](https://apidock.com/rails/ActiveRecord/AutosaveAssociation/destroyed_by_association). but I don't think they're applicable here.

mkllnk: Good research @kristinalim. I just looked into those two methods and they don't seem to help. Your suggestion about the internal state variable is the only other option I can think of. So we have these two options:

* We add a `Product.before_destroy` callback to set `@requires_variants = false` and check that variable during `Variant.before_destroy`.
* We check for the validity of `Variant.destroy` in controller actions.

Since we are experiencing callback hell already I find the second option much more attractive. @luisramos0 @kristinalim What do you think? Or should we move this into a separate issue?

sauloperez: Let's see what @kristinalim thinks. I have the feeling that this tightly related to the Variant model and might be legit to have callbacks for it but it's also right that we're experiencing the callback hell. If you want to give this approach a try, fine for me :+1:

luisramos0: I think ofn and rails in general have one major problem which is models that are bloated and include business logic. the solution for this is to take as much as possible out of the models even if it is data manipulation like in this case. it helps keep the models simple and clean: a difficult thing to do!
this means I think model callbacks should be forbidden in this project at this scale :-D
Every action you do on a model should be explicit either in the model itself or in a service layer as in this case. variant_deleter is a very nice service and we should have more of them.

also, we have a working solution here so I think we should merge this PR and move on to different challenges ;-)

kristinalim: Presence of at least one non-master variant is an assumption that we use in many places. Enforcing this in the model level is safer, and will help ensure that product data will always be valid.

kristinalim: I think it's worth having a model callback for. :+1:

kristinalim: Oh sorry, @luisramos0, I didn't see your comment.

If that's the direction that OFN is taking, not just for `Spree::Variant.destroy`, then I'm okay with this.

mkllnk: Okay, so we have two proposals and half of us agree on one and the other half on the other. I find both solutions acceptable. One will make it harder to maintain the code, the other risks invalid data.

Dilemma

Well, my argument would be that a product without variants is still valid. We just want to prevent users from deleting the last variant. Nothing should blow up if we had such a product in the database.

I'm actually a big fan of creating robust code that doesn't rely on many constraints. That makes it more flexible and stable. Imagine we wanted to replace a variant. Wouldn't it be nice if we could remove the current variant and add another one without it complaining it between?

Or imagine we decided to allow deleting the last variant and when that happens we just check for an empty product and delete that one as well. And that could be done in the VariantDeleter without changing or removing any callbacks and validations. Life could be that simple! :-D

Simple life

luisramos0: ehehe, interesting.
@kristinalim "If that's the direction that OFN is taking" it's only my opinion :-D

luisramos0: Just to give some context to my opinion: it comes from knowing other systems that use datamapper pattern ORM, not AR ORM like Rails, and these large scale systems are more complex _and_ look much **better than all this callback stuff that is already unmanageable at this small scale**. The better solution is to take this logic out of models into services and **repositories** that handle logic and data in separate place from the models and their callbacks themselves.

sauloperez: Agree on trying that @kristinalim ?

kristinalim: Trying which, the `VariantDeleter` option? Yes, I'm okay with it. :slightly_smiling_face: Sounds good!

mkllnk: Great. :+1:

sauloperez:
`products_and_inventory_report_spec.rb` is passing but I'll merge `master` into this branch to see the latest build status. At this point we need to be extra sure there are no regressions in the build.

luisramos0:
I see this new broken spec:
spec/serializers/admin/for_order_cycle/enterprise_serializer_spec.rb:48:
could be related to this PR.

RachL:
@luisramos0 @mkllnk judging from the last comment, does this PR need to be put back in dev? If not, can one of you stage it on FR? If I'm not wrong this is not targeting master.

luisramos0:
i dont think we should test individual PRs before the overall test.
Rachel, stage FR is ready for a major round of tests.

RachL:
@luisramos0 so this PR should be merged directly?

sigmundpetersen:
I would suggest a green build now that we have reached that point in 2-0 :trollface:

luisramos0:
I am moving it to "in dev" until @mkllnk looks at the suspicious broken spec: spec/serializers/admin/for_order_cycle/enterprise_serializer_spec

@RachL right now it's a lot more important to make the overall test of v2 than to test this specific PR. I'd merge this to get the phase 2 build greener and at the same time take note so that we include this PR when we make the phase 2 overall test.

mkllnk:
@luisramos0 I fixed the serializer spec but there is still one failing: `spec/features/admin/customers_spec.rb:184`

Have you seen that before? It passes on my computer reliably. If you have no idea either I will start debugging in Semaphore.

sigmundpetersen:
The build is green now after a rebuild, so it was one of those flaky specs.

We should see if they appear after merge and fix them in the target branch.

luisramos0:
admin/customers_spec is a very flaky spec in v2 at least.

mkllnk:
Thank you @luisramos0, good feedback. I updated the comment and the commit message. I hope it's clearer now.

luisramos0:
thanks Maikel, I get it now :-)

RachL:
> @RachL right now it's a lot more important to make the overall test of v2 than to test this specific PR. I'd merge this to get the phase 2 build greener and at the same time take note so that we include this PR when we make the phase 2 overall test.

@luisramos0 @sigmundpetersen then I'm not sure I understand why this is sitting in test ready...

luisramos0:
I am moving this back to code review so that @sauloperez  reviews this last commit.
Afterwards, we merge it directly as discussed [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/3496#issuecomment-465367552).
Unless @sauloperez has a different opinion about the process to follow for v2 PRs.

RachL:
My testing notes:

https://docs.google.com/document/d/13twIeD2BqqQB51ktM_hXype5qBWJYB-TWgXpDLx7c0g/edit#heading=h.gqpc7g2rwaw

And with this I played with stock info and updated comments in https://github.com/openfoodfoundation/openfoodnetwork/issues/3573

luisramos0:
I wonder what happens to the OC when the products/variants are deleted? do they simply disappear gracefully?

RachL:
@luisramos0 yes they do. Sorry I forgot to mentioned this in my notes.

Issue #3392, opened 2019-01-25
[Spree Upgrade] Order Cycles don't update to reflect changes made after they're created (product cache in v2)


sstead:
## Description
I'm noticing that if I make changes to things after an OC has been created, the changes that I made won't carry through to that OC in the shopfront.

Changes I've made that have not carried forward into the shop include:
- Changing the category of a product - the shop filter still sees it as the past category
- Adding a fee to the OC - the fee won't show in the shop like it should
- Adding a new product to the OC - the new product won't show
- Price - if I change the product's price it won't update

## Expected Behavior
The shopfront OCs should always show the up to date information

## Actual Behavior
Once an OC is open in a shop it won't update 

## Steps to Reproduce
1. Open an OC with some products in it
2. See the price of the products in the shop
3. Change the price of the products
4. See - has the price updated?

## Animated Gif/Screenshot
https://www.useloom.com/share/403e9539031e402390468cb03284372b?focus_title=1&muted=1

## Severity
Not on production 
bug-s4: it's annoying, but you can use it

## Your Environment
French staging, Chrome, PC, Desktop, Spree upgrade

luisramos0:
this issue will be related to #3391 #productcache

luisramos0:
can we validate in staging FR that this is not happening with disabled cache and close it?
we have #3391 to represent the product cache issue.

RachL:
* Open an OC with some products in it => OK
* See the price of the products in the shop => OK
* Change the price of the products => OK
* See - has the price updated? => YES

@luisramos0 is it normal to not have a PR attached?

luisramos0:
oh my god, it just works! can we deploy live? 🤣 

-- is it normal to not have a PR attached?
welcome to ofn v2, it just works 🚀 

now, back to reality: staging FR has a hacked version of v2 that includes #3404 
I did this so that we dont have to wait for the PR to get approved, tested, merged to master, then merged to v2, to test this. this way we are already sure it was a problem with the cache that needs to be looked at as part of fixing #3391
I think we can close this one.
Thanks @RachL 

any other bugs in v2? go for it 💪

mkllnk:
@luisramos0 Can this issue be closed? It's a bit confusing to have an issue be treated like a PR. ;-)

luisramos0:
yes, let's close this.

PR #3404, opened 2019-01-29
[Spree Upgrade] Add checkbox to disable products cache

luisramos0:
#### What? Why?
I couldnt replicate #3392 locally and I was having serious trouble debugging it in FR staging so, after 6 hours of unsuccessful investigations around what is wrong with the products cache, I decided to implement this feature:
In configuration - cache settings there's a new checkbox that admins can use to disable the products cache all together, this will be very useful to detect cache problems in staging and live without having to go through the unreadable diffs currently available...

#### What should we test?
There are auto tests to validate this is working. If the system is working correctly, disabling the cache should not change anything apart from making the PLP slower.
I have tested this in staging FR and it is working as it is successfully detecting a problem in the cache:
- go to shopfront "Winery District Food Hub" - OC#2
- see Apples
- go to the backoffice and check the new checkbox to activate cache
- go back to the shofront and see Apples disappear from the shopfront 

This feature helps detect an existing problem in caching this product Apples.

#### Release notes
Changelog Category: Added
Easier cache debugging with a new checkbox for admins to disable products cache.

This will be very useful to debug the cache in v2.

### DIFF
--- a/app/views/admin/cache_settings/edit.html.haml
+++ b/app/views/admin/cache_settings/edit.html.haml
@@ -1,6 +1,19 @@
 - content_for :page_title do
   = t(:cache_settings)
 
+= form_tag main_app.admin_cache_settings_path, :method => :put do
+  .field
+    = hidden_field_tag 'preferences[enable_products_cache?]', '0'
+    = check_box_tag 'preferences[enable_products_cache?]', '1',  Spree::Config[:enable_products_cache?]
+    = label_tag nil, t('.enable_products_cache')
+  .form-buttons
+    = button t(:update), 'icon-refresh'
+
+%br
+%br
+
+%h4= t(:cache_state)
+%br
 %table.index
   %thead
     %tr
--- a/lib/open_food_network/cached_products_renderer.rb
+++ b/lib/open_food_network/cached_products_renderer.rb
@@ -27,19 +27,19 @@ module OpenFoodNetwork
     private
 
     def cached_products_json
-      return uncached_products_json unless Spree::Config[:enable_products_cache?]
+      return uncached_products_json unless use_cached_products?
 
-      if Rails.env.production? || Rails.env.staging?
-        Rails.cache.fetch("products-json-#{@distributor.id}-#{@order_cycle.id}") do
-          begin
-            uncached_products_json
-          rescue ProductsRenderer::NoProducts
-            nil
-          end
+      Rails.cache.fetch("products-json-#{@distributor.id}-#{@order_cycle.id}") do
+        begin
+          uncached_products_json
+        rescue ProductsRenderer::NoProducts
+          nil
         end
-      else
-        uncached_products_json
       end
+    end
+
+    def use_cached_products?
+      Spree::Config[:enable_products_cache?] && (Rails.env.production? || Rails.env.staging?)
     end
 
     def uncached_products_json
--- a/lib/open_food_network/cached_products_renderer.rb
+++ b/lib/open_food_network/cached_products_renderer.rb
@@ -27,6 +27,8 @@ module OpenFoodNetwork
     private
 
     def cached_products_json
+      return uncached_products_json unless Spree::Config[:enable_products_cache?]
+
       if Rails.env.production? || Rails.env.staging?
         Rails.cache.fetch("products-json-#{@distributor.id}-#{@order_cycle.id}") do
           begin
--- a/spec/lib/open_food_network/cached_products_renderer_spec.rb
+++ b/spec/lib/open_food_network/cached_products_renderer_spec.rb
@@ -5,7 +5,7 @@ module OpenFoodNetwork
   describe CachedProductsRenderer do
     let(:distributor) { double(:distributor, id: 123) }
     let(:order_cycle) { double(:order_cycle, id: 456) }
-    let(:cpr) { CachedProductsRenderer.new(distributor, order_cycle) }
+    let(:cached_products_renderer) { CachedProductsRenderer.new(distributor, order_cycle) }
 
     # keeps global state unchanged
     around do |example|
@@ -37,7 +37,7 @@ module OpenFoodNetwork
           end
 
           it "returns uncached products JSON" do
-            expect(cpr.products_json).to eq 'uncached products'
+            expect(cached_products_renderer.products_json).to eq 'uncached products'
           end
         end
 
@@ -47,7 +47,7 @@ module OpenFoodNetwork
           end
 
           it "returns the cached JSON" do
-              expect(cpr.products_json).to eq 'products'
+              expect(cached_products_renderer.products_json).to eq 'products'
           end
         end
       end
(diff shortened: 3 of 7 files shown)


mkllnk:
Is that something we should backport to master?

luisramos0:
ok, I switched this to master and tested locally. Looking good. Ready for review.

luisramos0:
rebasing to increase the chances of a green build.

app/views/admin/cache_settings/edit.html.haml
```diff
@@ -1,6 +1,19 @@
...
+    = check_box_tag 'preferences[enable_products_cache?]', '1',  Spree::Config[:enable_products_cache?]
+    = label_tag nil, t('.enable_products_cache')
+  .form-buttons
+    = button t(:update), 'icon-refresh'
+
+%br
+%br
+
+%h4= t(:cache_state)
+%br
```
sauloperez: are we sure that allowing any superadmin to play with this is the right move? IMO this belongs to code-level config only available to be modified by devs, which is something different than settings.

Besides, settings page already looks like a hard to manage Frankenstein.

I know we've added settings for other similar things and it seems unfair but IMO this distinction is pretty important. What do you guys think?

lib/open_food_network/cached_products_renderer.rb
```diff
@@ -27,6 +27,8 @@ def products_json
     private
 
     def cached_products_json
+      return uncached_products_json unless Spree::Config[:enable_products_cache?]
+
       if Rails.env.production? || Rails.env.staging?
```
sauloperez: What if we rather extract a method that takes these 3 conditions into account? I think this would decouple this method from these low-level details.

spec/lib/open_food_network/cached_products_renderer_spec.rb
```diff
@@ -7,19 +7,57 @@ module OpenFoodNetwork
...
+          end
+        end
+
+        context "enabled" do
+          before do
+            Spree::Config[:enable_products_cache?] = true
+          end
+
+          it "returns the cached JSON" do
+              expect(cpr.products_json).to eq 'products'
```
sauloperez: I know `cpr` was already defined but it'd be great if we could improve the naming so other devs find it easier to follow.

app/views/admin/cache_settings/edit.html.haml
luisramos0: I just dont want to have to login to the server to debug this in the process of debugging v2. So, we need this now in ofn v2, that's the reason I built it, and that's why it's already deployed in fr staging :-D
I think this is a lot more useful if available to superadmins.
Also, this is not a big risk at all, the cache will keep being updated, it will just not be picked up by the shopfront renderer.
So, I vote for inclusion of this option in the existing config page.

I think we can plan to split the config pages into two different groups: the app-admin pages and the system-admin pages like this one.

lib/open_food_network/cached_products_renderer.rb
luisramos0: I am merging all changes from master to 2-0 so when I work in master I avoid all refactoring 🙄 🤣 

this is very obviously a candidate for improvement, I'll do it now.

spec/lib/open_food_network/cached_products_renderer_spec.rb
luisramos0: I am just avoiding merge hell...
I'll do it now.

luisramos0:
Rebased to get a more reliable build.

lib/open_food_network/cached_products_renderer.rb
luisramos0: done

spec/lib/open_food_network/cached_products_renderer_spec.rb
luisramos0: done

lib/open_food_network/cached_products_renderer.rb
```diff
@@ -27,19 +27,21 @@ def products_json
...
+        rescue ProductsRenderer::NoProducts
+          nil
         end
-      else
-        uncached_products_json
       end
     end
 
+    def use_cached_products?
+      Spree::Config[:enable_products_cache?] && (Rails.env.production? || Rails.env.staging?)
```
sauloperez: What if we just set `enable_products_cache?` from each of the different `config/environments/` files: development, production, staging and test? Then we wouldn't need the `&&` at all and we could handle this config without touching this code :ok_hand:

spec/lib/open_food_network/cached_products_renderer_spec.rb
```diff
@@ -91,11 +129,11 @@ module OpenFoodNetwork
             end
 
             it "raises an error" do
-              expect { cpr.products_json }.to raise_error CachedProductsRenderer::NoProducts
+              expect { cached_products_renderer.products_json }.to raise_error CachedProductsRenderer::NoProducts
             end
 
             it "caches the products as nil" do
-              expect { cpr.products_json }.to raise_error CachedProductsRenderer::NoProducts
+              expect { cached_products_renderer.products_json }.to raise_error CachedProductsRenderer::NoProducts
```
sauloperez: thanks!

lib/open_food_network/cached_products_renderer.rb
luisramos0: I disagree because super admin would not be able to run the following test:

sauloperez: I'm was suggesting a different way to implement what you did in this line. No change in behavior at all, but now that I read the line again I don't understand why we exactly need the `(Rails.env.production? || Rails.env.staging?)` part of the conditional.

luisramos0: yes, cache is disabled in dev by default. that's it.

RachL:
@luisramos0 can you staged it to FR? What's the best option for me to ask for staging? Here or pinging everyone on slack?

luisramos0:
pinging here is fine @RachL 
but this is not v2, you can check the branch name at the top of the PR, it says master in this case (it will say 2-0-stable in v2 PRs), that means it needs to go to any staging server other than FR and be validated as any other v1 PR.

luisramos0:
My brain must have stopped (I closed this PR and deleted the branch...).
@RachL this is ready to staged and tested in any v1 staging server.

RachL:
@luisramos0 may I have a bit more info to test this in UK staging?

I see the checkbox is already checked after staging this. Am I looking at the right place? I went here: https://staging.openfoodnetwork.org.uk/admin/cache_settings/edit

Also, were there any particularities to the Apple product in staging FR that I should replicate?

luisramos0:
hey @RachL 
the checkbox should normally be checked: cache should be used.
This feature will be useful to detect problems BUT, in a healthy version like v1, checking or unchecking the checkbox should bring no changes to the system.
So, the tests should be to change catalog and OC data and check that what comes up in the shopfront is correct (with or without the checkbox activated).
I dont know what's causing the issue in v2, this feature will help find out.
Makes sense?

RachL:
With or without the checkbox activated, I see no difference in shopfront, before or after a change. testing notes : https://docs.google.com/document/d/1a45rFezcspQ84OKh7wUnS21M6Od3VgWfelhIGTzax0E/edit#

luisramos0:
awesome!

PR #3232, opened 2018-12-20
[Spree Upgrade] Scope variants in line items availability validator

luisramos0:
This PR includes all 5 commits from #3210, this PRs commits are the last 5.
#### What? Why?

Closes #3118 

When validating line_item.variant stock levels in stock/availability_validator we need to scope variants so that variant overrides stock levels are used instead. Here we re-use the the line_item.scoper so that the variant_overrides are not loaded several times in the handling fo line_items in the same request.

Additional problems solved:
- adapting line_item.sufficient_stock? to spree 2 by using the new variant.can_supply?
- fix variant.in_stock? by also using the new variant.can_supply?
- move stock_location.fill_status to VariantStock and fix it, it was totally dependent on stock_item methods and is not using already overridden variant methods (on_demand and total_on_hand)

#### What should we test?
This is one of the final steps in making cart and checkout handle stock levels correctly in v2!
the spec in 3118 (spec/features/consumer/shopping/variant_overrides_spec.rb:152) should be green.

#### Dependencies
This PR goes after #3210

PR #3429, opened 2019-02-01
[Spree Upgrade] 3428 Fix in_stock? and shopping/variant_overrides_spec

### LINKED ISSUE #3428: [Spree Upgrade] Fix consumer/shopping/variant_overrides_spec (2 broken specs)
I cant remember when was this introduced, I believe it was when headless chrome was merge to 2-0.

We have two broken specs:
- spec/features/consumer/shopping/variant_overrides_spec.rb:47 # shopping with variant overrides defined viewing products looks up stock from the override
- rspec ./spec/features/consumer/shopping/variant_overrides_spec.rb:152 # shopping with variant overrides defined creating orders does not show out of stock flags on order confirmation page

They could be intermittent fails but we need to have a look any way.

Re test in line 47: the overridden variants are not expected to be shown on the shopfront page and I can see they are listed disabled but the test code thinks they are enabled and so it fails.

Re test in line 152: there should be no out of stock message in the cart but it is showing up. I am not sure why is this behaviour correct when the override is out of stock. Needs investigation.
 
The errors:
```
 1) shopping with variant overrides defined viewing products looks up stock from the override
     Failure/Error: page.should_not have_content product1_variant2.options_text
       expected not to find text "2g, S" in "SHOPS MAP PRODUCERS GROUPS ABOUT English Castellano English Login Shopping @ ENTERPRISE 113 0 items Enterprise 113 Herndon Next order closing in 7 days Ready for time ABOUT ENTERPRISE 113 PRODUCERS CONTACT GROUPS Brand Ruby on Rails Product #78 - 9454 from Enterprise 114 Product #78 - 9454 1g, S $61.11 $0.00 Product #78 - 9454 2g, S $85.55 $0.00 Product #78 - 9454 4g, S $85.55 $0.00 Product #79 - 6119 from Enterprise 114 Product #79 - 6119 3g, S $85.55 $0.00 Product #80 - 9260 from Enterprise 114 Product #80 - 9260 5g, S $85.55 $0.00 Product #80 - 9260 6g, S $85.55 $0.00 Interested in getting on the Open Food Network? Register here OFN Global Home | News | About | Contact OFN Sites Developer | Community | User Guide Secure and trusted. Open Food Network uses SSL encryption (2048 bit RSA) everywhere to keep your shopping and payment information private. Our servers do not store your credit card details and payments are processed by PCI-compliant services. Keep in touch Email us Newsletter sign-up News Calendar Navigate Shops Map Producers Groups About Join us Create a listing, shop or group directory on the Open Food Network. Tell me more! Read our Terms and conditions | Find us on GitHub Open Food Network is a free and open source software platform. Our content is licensed with CC BY-SA 3.0 and our code with AGPL 3. We take good care of your data. See our cookies policy"
     # ./spec/features/consumer/shopping/variant_overrides_spec.rb:49:in `block (3 levels) in <top (required)>'
```

luisramos0:
#### What? Why?

Closes #3428

This is a bug in #3232 should have added unit tests... some of them would be broken. See new tests in scope_variant_to_hub_spec (2 of them were broken before this fix).

We cannot override Variant.in_stock? inside the concern VariantStock, we need to do it directly in the variant_decorator. The VariantStock.in_stock? was just not seen from the outside and was the source of #3428.
I think the VO shopping spec was a false positive before headless chrome and became a true error after it. This PR fixes the spec and the in_stock issue, this time, thanks to spring, with some unit testing covering the different cases.

#### What should we test?
the 2 specs in #3428 should be green.

### DIFF
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -125,11 +125,6 @@ module VariantStock
     on_demand || total_on_hand >= quantity
   end
 
-  # We override Spree::Variant.in_stock? to avoid depending on the non-overidable method Spree::Stock::Quantifier.can_supply?
-  def in_stock?(quantity = 1)
-    can_supply?(quantity)
-  end
-
   # Moving Spree::StockLocation.fill_status to the variant enables us to override this behaviour for variant overrides
   # We can have this responsibility here in the variant because there is only one stock item per variant
   #
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -87,6 +87,12 @@ Spree::Variant.class_eval do
     ]
   end
 
+  # We override in_stock? to avoid depending on the non-overridable method Spree::Stock::Quantifier.can_supply?
+  #   VariantStock implements can_supply? itself which depends on overridable methods
+  def in_stock?(quantity = 1)
+    can_supply?(quantity)
+  end
+
   def price_with_fees(distributor, order_cycle)
     price + fees_for(distributor, order_cycle)
   end
--- a/spec/features/consumer/shopping/variant_overrides_spec.rb
+++ b/spec/features/consumer/shopping/variant_overrides_spec.rb
@@ -39,12 +39,10 @@ feature "shopping with variant overrides defined", js: true, retry: 3 do
   end
 
   describe "viewing products" do
-    it "shows the overridden price" do
-      page.should_not have_price with_currency(12.22) # product1_variant1.price ($11.11) + 10% fee
+    it "shows price and stock from the override" do
       page.should have_price with_currency(61.11) # product1_variant1_override.price ($55.55) + 10% fee
-    end
+      page.should_not have_price with_currency(12.22) # product1_variant1.price ($11.11) + 10% fee
 
-    it "looks up stock from the override" do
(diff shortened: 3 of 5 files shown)


mkllnk:
Wow, what a learning. `Spree::Variant` has an instance method called `in_stock?`. We can override it in a decorator via `Spree::Variant.class_eval`. But if we try to override it via `include`, it doesn't work.

> In fact, mixed-in modules effectively behave as superclasses.
> https://stackoverflow.com/a/9828056/3377535

The `in_stock?` method in VariantStock was effectively the parent method. It was overridden by Spree's own implementation.

Fun fact: If Spree had called `super`, they would have called our `VariantStock.in_stock?`.

@luisramos0 Have you considered `prepend` instead of `include`? It fixes this scenario by executing the mixin later. There is a great post on SO: https://stackoverflow.com/a/4471202/3377535

luisramos0:
ah, interesting, I just tested it, I didnt know that it was working as a super class. thanks.

yes, that stackoverflow post is amazing, I think I have read it in the past, it is good to re-read, so many options!

I think we should leave VariantStock with include because it really is extra stuff on top of Variant, it is not supposed to override Variants methods.

luisramos0:
thanks!

PR #3435, opened 2019-02-03
Use Dir::Tmpname.make_tmpname for threadsafe file naming

### LINKED ISSUE #3423: [Product import] Parallel imports override each other
## Description

I think that we have a problem if several people try to use the product import within the same second. The file name contains a timestamp including seconds, but no other random part or incremental number. So if two people uploaded a CSV file in the same second, it would create a big mess.

```rb
    def save_uploaded_file(upload)
      filename = 'import' + Time.zone.now.strftime('%d-%m-%Y-%H-%M-%S')
      extension = '.' + upload.original_filename.split('.').last
      directory = 'tmp/product_import'
      Dir.mkdir(directory) unless File.exist?(directory)
      File.open(Rails.root.join(directory, filename + extension), 'wb') do |f|
        f.write(upload.read)
        f.path
      end
    end
```

## Expected Behavior

Many people can use the product upload and it never fails.

## Actual Behavior

It would try to write the same file. This could either fail or write data from two people into one file. Best case it's just invalid. Worst case a person is importing data from another person.

## Steps to Reproduce

1. Open two different windows and try to upload two different sets of products with the same second.

## Context

I was just looking through product import code for the Spree upgrade.

## Severity

I would say s3, because in some cases it's broken. The workaround is to upload when nobody else is uploading.

## Possible Fix

Use Ruby's `Dir::Tmpname` for a safe and unique place to upload a file.

Matt-Yorkley:
#### What? Why?

Closes #3423

Uploaded CSV files were using a time-based filename that was not adequately unique, so if two files were uploaded in the same second by different users it could cause an issue.

I've used `Dir::Tmpname.make_tmpname` as suggested by @mkllnk (thanks for the lesson!)

It's output looks like this in the Rails console, if anyone's interested:
```
irb(main):001:0> Dir::Tmpname.make_tmpname 'import', nil
=> "import20190203-28965-1j8l7ha"
```

#### What should we test?

No testing required. If the specs are green it should be fine.

#### Release notes

Changed naming method for temporary uploaded CSV files.

Changelog Category: Changed

### DIFF
--- a/app/controllers/admin/product_import_controller.rb
+++ b/app/controllers/admin/product_import_controller.rb
@@ -79,11 +79,9 @@ module Admin
     end
 
     def save_uploaded_file(upload)
-      filename = 'import' + Time.zone.now.strftime('%d-%m-%Y-%H-%M-%S')
-      extension = '.' + upload.original_filename.split('.').last
-      directory = 'tmp/product_import'
-      Dir.mkdir(directory) unless File.exist?(directory)
-      File.open(Rails.root.join(directory, filename + extension), 'wb') do |f|
+      extension = File.extname(upload.original_filename)
+      directory = Dir.mktmpdir 'product_import'
+      File.open(File.join(directory, "import#{extension}"), 'wb') do |f|
         data = UploadSanitizer.new(upload.read).call
         f.write(data)
         f.path


Matt-Yorkley:
@mkllnk I tried replacing `File` with `Tempfile`. They're not exactly interchangeable, and it seemed to cause failing tests elsewhere that I couldn't manage to solve. Also, it looks like `Tempfile` just imports `Dir::Tmpname` anyway and uses the same file naming technique.

mkllnk:
Oh no, how sad is that. Is it about the file getting deleted automatically? Sidetrack: When do we delete uploaded product sheets? Do we need a cleanup job?

If you look at the implementation of Tempfile#initialize, it's a lot more sophisticated than just calling Dir::Tmpname.make_tmpname. It has all that logic to ensure that it creates a unique file without any race conditions. I find it difficult to give that up.

I have one more idea that could work here: `Dir.mktmpdir` makes sure that the directory doesn't exist yet. It means that we have one additional directory for each file but it should be safe.

Other solutions like creating our own class are probably not feasible as the chance of collision is very small. What do you think?

sigmundpetersen:
Wasn't that fixed in #3430 ? A rebase should do

Matt-Yorkley:
> Do we need a cleanup job?

I think the /tmp folder gets cleared once per day, and clears things out based on their age. It should be automatic, but maybe we can investigate a bit more.

I'll try using `Dir.mktmpdir`. :+1:

sigmundpetersen:
So is this ready for re-review? @Matt-Yorkley @sauloperez @mkllnk

Matt-Yorkley:
Ready for re-review @mkllnk

app/controllers/admin/product_import_controller.rb
```diff
@@ -79,11 +79,9 @@ def check_spreadsheet_has_data(importer)
     end
 
     def save_uploaded_file(upload)
-      filename = 'import' + Time.zone.now.strftime('%d-%m-%Y-%H-%M-%S')
-      extension = '.' + upload.original_filename.split('.').last
-      directory = 'tmp/product_import'
-      Dir.mkdir(directory) unless File.exist?(directory)
-      File.open(Rails.root.join(directory, filename + extension), 'wb') do |f|
+      extension = upload.original_filename.split('.').last
```
sauloperez: Sorry to be late to the party but Ruby's [.extname](https://ruby-doc.org/core-2.1.5/File.html#method-c-extname) feels better than this.

Matt-Yorkley: The uploaded file at this point is an `ActionDispatch::Http::UploadedFile` object that doesn't respond to `.extname`. It only has the `.original_filename` attribute.

mkllnk: Interesting, it's actually a static method and slightly more clever than split. https://apidock.com/ruby/File/extname/class
```diff
-      extension = upload.original_filename.split('.').last
+      extension = File.extname(upload.original_filename)
```

app/controllers/admin/product_import_controller.rb
```diff
@@ -79,11 +79,9 @@ def check_spreadsheet_has_data(importer)
     end
 
     def save_uploaded_file(upload)
-      filename = 'import' + Time.zone.now.strftime('%d-%m-%Y-%H-%M-%S')
       extension = '.' + upload.original_filename.split('.').last
```
mkllnk: You still have the dot in here and below in `import.`. Pau's suggestion would be apt now:
```suggestion
       extension = File.extname(upload.original_filename)
```

sigmundpetersen:
I think a rebase would be good here? Both for conflicts and higher probability of a green build after some flaky spec fixes were merged to master

app/controllers/admin/product_import_controller.rb
Matt-Yorkley: Done :+1:

sauloperez:
ping @mkllnk . Waiting for your approval.

RachL:
@mkllnk @Matt-Yorkley I see in the test comments that no tests were required, is this true? Otherwise I guess I need to launch 2 import from 2 browsers, almost at the same time. Am I right?

sauloperez:
> Otherwise I guess I need to launch 2 import from 2 browsers, almost at the same time. Am I right?

I think so, but uploading two files in the same second is nearly impossible. We can only do a smoke test at most.

RachL:
Ok so I've used files from previous tests cases to add / update product. It seems to work as usual. Files used:
https://docs.google.com/document/d/1E95MmfFnANvbUBpKKDNl9M1BcCTcIKwFGrfFziq4oPg/edit#

PR #3194, opened 2018-12-07
[Spree Upgrade] Fix admin new order page

HugsDaniel:
#### What? Why?

Closes #3109 and #3107 and #2416
The spec in #3107 should be green.

@HugsDaniel said: 
New new/edit order page in back office! Upgraded Spree views to 2.0 and applied overrides.
There's no error anymore for blank line_items (it was added with deface), and line items can be added.
I used Spree default routes for these actions (see in `_routes.html.erb` to make this work. 

@luisramos0 said:
Regarding the fix to #2416 by adapting app/views/spree/admin/variants/_autocomplete.js.erb and the respective search.rabl file: this is a quick fix to the problem, there's plenty we could improve here, I suggest we leave for after the upgrade, I created this tech debt issue #3440

Remaining work to be done in follow up PRs:
- the display of order adjustments needs to be reviewed (adjustments included in price are not showing) #2940 (fixed in PR #3445)
- fix the flow between order details page and customer details page #3103 (fixed in PR #3436)
- scope variants when adding them to the order api/shipments (and probably get rid of admin/line_items_controller) #3422 (fixed in PR #3441)

**For the review, I recommend reviewers to follow the commits in order.**

PR #3436, opened 2019-02-03
[Spree Upgrade] Fix admin new order page flow (from order details to customer details)

### LINKED ISSUE #3103: [Spree Upgrade] Changing distributor and OC in the Edit Orders page (3 broken specs)
Expected behaviour: if an order is not complete, the order edit page should let manager change the distributor and the OC.

Related broken specs:
```
  3) 
    As an administrator
    I want to manage orders
 displays error when incorrect distribution for products is chosen
     Failure/Error: targetted_select2(value, options)
     
     Capybara::ElementNotFound:
       Unable to find visible css "#s2id_order_distributor_id"
     # ./spec/support/request/web_helper.rb:163:in `select2_select'
     # ./spec/features/admin/orders_spec.rb:116:in `block (2 levels) in <top (required)>'

  4) 
    As an administrator
    I want to manage orders
 can't change distributor or order cycle once order has been finalized
     Failure/Error: expect(page).to have_selector 'p', text: "Distributor: #{@order.distributor.name}"
       expected to find visible css "p" with text "Distributor: Enterprise 25" but there were no matches
     # ./spec/features/admin/orders_spec.rb:144:in `block (2 levels) in <top (required)>'

```

luisramos0:
#### What? Why?

Closes #3103

The new spree admin orders controller together with the variant search JS, after adding the variant, run a window.reload on order#update which redirects to customer details after the first variant is added to the order...

I dont think this behaviour of jumping to customer details after the first variant is added is acceptable. So, we need to find a way to keep the user in the order details page when variants are added and keep the behaviour of jumping to customer details if user clicks "update and recalculate" and the order is not complete.

This PR changes the spree admin orders controller to adapt to our needs by redirecting from update to edit when there are errors (and sends the errors in flash errors) so that when a variant is added and the page is reloaded, the current action is edit and the page doesn't jump to the customer details (the jump is configured on the update action).

#### What should we test?
The 3 specs in #3103 should be green but only together with #3194 (I have verified this in a [previous build](https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3436/builds/2))

### DIFF
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -27,6 +27,23 @@ Spree::Admin::OrdersController.class_eval do
     # within the page then fetches the data it needs from Api::OrdersController
   end
 
+  # Re-implement spree method so that it redirects to edit instead of rendering edit
+  #   This allows page reloads while adding variants to the order (/edit), without being redirected to customer details page (/update)
+  def update
+    unless @order.update_attributes(params[:order]) && @order.line_items.present?
+      @order.errors.add(:line_items, Spree.t('errors.messages.blank')) if @order.line_items.empty?
+      return redirect_to edit_admin_order_path(@order), :flash => { :error => @order.errors.full_messages.join(', ') }
+    end
+
+    @order.update!
+    if @order.complete?
+      redirect_to edit_admin_order_path(@order)
+    else
+      # Jump to next step if order is not complete
+      redirect_to admin_order_customer_path(@order)
+    end
+  end
+
   # Overwrite to use confirm_email_for_customer instead of confirm_email.
   # This uses a new template. See mailers/spree/order_mailer_decorator.rb.
   def resend
--- a/spec/controllers/spree/admin/orders_controller_spec.rb
+++ b/spec/controllers/spree/admin/orders_controller_spec.rb
@@ -4,29 +4,69 @@ describe Spree::Admin::OrdersController, type: :controller do
   include AuthenticationWorkflow
   include OpenFoodNetwork::EmailHelper
 
-  context "updating an order with line items" do
-    let!(:order) { create(:order) }
-    let(:line_item) { create(:line_item) }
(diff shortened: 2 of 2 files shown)


app/controllers/spree/admin/orders_controller_decorator.rb
```diff
@@ -27,6 +27,23 @@ def index
     # within the page then fetches the data it needs from Api::OrdersController
   end
 
+  # Re-implement spree method so that it redirects to edit instead of rendering edit
+  #   This allows page reloads while adding variants to the order (/edit), without being redirected to customer details page (/update)
+  def update
+    unless @order.update_attributes(params[:order]) && @order.line_items.present?
+      @order.errors.add(:line_items, Spree.t('errors.messages.blank')) if @order.line_items.empty?
```
sauloperez: Isn't this implemented in the model layer? If not, it should be there and be consistent with the other places from where we can place an order (the shopfront).

luisramos0: I think this is specific to the BO
see the spree code:
https://github.com/spree/spree/blob/58ef94cdb17310d62e110dd92a7d4b4d1f270335/backend/app/controllers/spree/admin/orders_controller.rb#L82

the orders in the shopfront are the carts so they don't have line items and the presence of line items is not relevant most of the time.
what you suggest would imply validating the present of line items in a specific stage of the order workflow.
I wouldnt go that way... what do you think?

sauloperez: In that case is probably is wiser to stick to this :+1:

luisramos0:
@sauloperez re tests, this is fixing 2 tests already.

luisramos0:
reconsidering testing, I need to add some controller level tests here: spec/controllers/spree/admin/orders_controller_spec.rb

luisramos0:
I have added specs for the update method (I added them to the same commit without changing the controller).
@sauloperez pls re-review, thanks!

PR #3441, opened 2019-02-05
[Spree Upgrade] Adapt api shipments controller to OFN by scoping variants before each action

### LINKED ISSUE #3422: [Spree Upgrade] Scope variants when they are added in the edit order page through api/shipments (2 broken specs)
## Description
#2416 is related to searching variants in the admin orders edit page before adding them to the order, this issue is related to the following step which adds the selected variant to the order. 
In spree 2 this is done through api/shipments now.
The process is working but we probably need scoping... that will probably fix this spec: spec/features/admin/variant_overrides_spec.rb:408

Adittionally, we can probably get rid of spree/admin/line_items_controller and delete the broken spec  spec/controllers/spree/admin/line_items_controller_spec.rb:91

## Acceptance criteria
No broken specs:
- spec/features/admin/variant_overrides_spec.rb:408
- spec/controllers/spree/admin/line_items_controller_spec.rb:91

luisramos0:
#### What? Why?

Closes #3422

Admin edit Order page is now correctly managing variant overrides in the order.

Additionally, because the edit order page is using the shipments api (that we are adapting in this PR), we can delete all the code related to the endpoint used in v1 spree::admin::line_items_controller.
This deletes the failing spec spec/controllers/spree/admin/line_items_controller_spec.rb:91

#### What should we test?
spec/features/admin/variant_overrides_spec.rb:408 is testing this behaviour and should be green now.

#### Dependencies
This PR depends on #3194 to work and make the specs green.


app/controllers/spree/api/shipments_controller_decorator.rb
```diff
@@ -0,0 +1,39 @@
...
+    variant = Spree::Variant.find(params[:variant_id])
+    OpenFoodNetwork::ScopeVariantToHub.new(@order.distributor).scope(variant)
+
+    quantity = params[:quantity].to_i
+
+    @order.contents.remove(variant, quantity, @shipment)
+    @shipment.reload if @shipment.persisted?
+    respond_with(@shipment, :default_template => :show)
+  end
+end
```
sauloperez: Is there something stopping us from inheriting from `Spree::Api::ShipmentsController instead`? We would need then to explicitly override the route though but we already know the dangers of class_evil.

luisramos0:
thanks for the review!
I'll give inheritance a try.

re tests, this type of coding (overriding frameworks code) is so poor that I don't feel motivated to write tests...  inheritance will help but anyway... not nice.
also the create action is covered by the spec we are fixing.
but I definitely agree, I must cover this controller with tests. I'll do that by copying more code :sob: from https://github.com/spree/spree/blob/2-1-stable/api/spec/controllers/spree/api/shipments_controller_spec.rb

luisramos0:
After some predicted stuckness with inheritance... and quite a lot of hours on this one I am pushing what I think is a decent coverage over the class_eval solution.
This is an interesting api end point if you care to read the specs and there are lots of things going on here with inventory_units and variant overrides, and I havent really thought about callbacks...
I did end up changing the behaviour of #create to reuse order.shipment because a second shipment is not allowed in the ofn order model.
@sauloperez pls have a look, thanks!

spec/controllers/spree/api/shipments_controller_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      it 'creates a shipment if order does not have a shipment' do
+        order.shipment.destroy
+        order.reload
+
+        spree_post :create, params
+
+        validate_response
+        expect(json_response["inventory_units"].size).to eq 2
+        expect(order.reload.line_items.first.variant.price).to eq(variant.price)
+      end
```
sauloperez: and what if  `@order.shipments.create` fails?

luisramos0: yeah, I wondered. spree code is not doing anything and not testing either...
I tested existing spree code and the new adaptation.

What do you think I should do?
1- change the code to rescue some exception and add error codes? we would have to handle errors in shipments.create, @order.contents.add and @shipment.save!
2- just cover an error case with a spec? create a spec that makes shipments.create raise and then assert it is raising?

luisramos0:
I have rebased and resolved conflicts.

spec/controllers/spree/api/shipments_controller_spec.rb
luisramos0: The answer to your question is
if shipments.create fails, the endpoint will return a http error code. for the user that means adding the first variant to a new order will fail.
option 2 above would be to cover this http error code with a spec.

sauloperez: ok, so that means it is somehow dealt with and it doesn't blow into pieces. IMO we should cover so with specs. We'll see then if we like what is returned and it needs to be worked on in an upcoming PR.

luisramos0: ok, I'll do that :+1:

luisramos0: ok, I'll do that.

luisramos0: done, pls have a look @sauloperez

sauloperez:
spec/features/admin/variant_overrides_spec.rb passing :+1:

luisramos0:
thanks a lot for pushing back @sauloperez , this was worth the time, it's an important endpoint!

PR #3444, opened 2019-02-06
3437 supplier name on invoice

### LINKED ISSUE #3437: Add Supplier name into default invoice template
Today there is no easy way to print individual packing sheets. Bulk Invoice Printing has made it much easier to generate and print all the Customer orders in one go, however they still can't be used as packing sheets because "the OFN current invoice does NOT have the supplier’s name on it. In order for Mary or Shannon to pack orders, they MUST know whose products to put into the order, because it is very likely that they have the same product/variant from multiple suppliers"

Current invoice looks like this ..

By adding the Supplier name to the invoice, the invoices will be able to be used as packing sheets

## What is to be done
The Supplier name should be added in small italic font underneath each product line, as it appears in the email confirmation (shown below)

Hub users can easily use Invoices, including via Bulk Invoice Printing to pack Customer orders

## Link to the "Product Development - Backlog" item in Discourse
https://community.openfoodnetwork.org/t/add-supplier-name-into-default-invoice-template/1453

mkllnk:
#### What? Why?

Closes #3437.

Simply adding the supplier name to each line item on the invoice allows a food hub to use the invoice as packing sheet.

#### What should we test?

- Create an order and look at the emailed template.
- Change the invoice template as super-admin in Settings/ Invoice Settings.
- View an order as admin and look at the invoice through the action "Print Invoice"

#### Release notes

Changed: Invoices now contain the supplier name for each item.

Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -37,9 +37,7 @@ Spree::Admin::OrdersController.class_eval do
   end
 
   def invoice
-    pdf = render_to_string pdf: "invoice-#{@order.number}.pdf",
-                           template: invoice_template,
-                           formats: [:html], encoding: "UTF-8"
+    pdf = InvoiceRenderer.new.render(@order)
 
     Spree::OrderMailer.invoice_email(@order.id, pdf).deliver
     flash[:success] = t('admin.orders.invoice_email_sent')
@@ -48,7 +46,7 @@ Spree::Admin::OrdersController.class_eval do
   end
 
   def print
-    render pdf: "invoice-#{@order.number}", template: invoice_template, encoding: "UTF-8"
+    render InvoiceRenderer.new.args(@order)
   end
 
   def print_ticket
@@ -60,10 +58,6 @@ Spree::Admin::OrdersController.class_eval do
   end
 
   private
-
-  def invoice_template
-    Spree::Config.invoice_style2? ? "spree/admin/orders/invoice2" : "spree/admin/orders/invoice"
-  end
 
   def require_distributor_abn
     unless @order.distributor.abn.present?
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -37,7 +37,7 @@ Spree::Admin::OrdersController.class_eval do
   end
 
   def invoice
-    pdf = InvoiceRenderer.new.render(@order)
+    pdf = InvoiceRenderer.new.render_to_string(@order)
 
     Spree::OrderMailer.invoice_email(@order.id, pdf).deliver
     flash[:success] = t('admin.orders.invoice_email_sent')
--- a/app/services/invoice_renderer.rb
+++ b/app/services/invoice_renderer.rb
@@ -1,0 +1,29 @@
+class InvoiceRenderer
+  def render_to_string(order)
+    renderer.render_to_string(args(order))
+  end
+
+  def args(order)
+    {
+      pdf: "invoice-#{order.number}.pdf",
+      template: invoice_template,
+      formats: [:html],
+      encoding: "UTF-8",
+      locals: { :@order => order }
+    }
+  end
+
+  private
+
+  def renderer
+    ApplicationController.new
+  end
+
+  def invoice_template
+    if Spree::Config.invoice_style2?
+      "spree/admin/orders/invoice2"
+    else
+      "spree/admin/orders/invoice"
+    end
+  end
+end
--- a/app/views/spree/admin/orders/_invoice_table2.html.haml
+++ b/app/views/spree/admin/orders/_invoice_table2.html.haml
@@ -17,6 +17,9 @@
       %tr
         %td
           = render 'spree/shared/line_item_name', line_item: item
+          %br
+          %small
+            %em= raw(item.variant.product.supplier.name)
         %td{:align => "right"}
           = item.quantity
         %td{:align => "right"}
--- a/spec/services/bulk_invoice_service_spec.rb
+++ b/spec/services/bulk_invoice_service_spec.rb
@@ -17,8 +17,7 @@ describe BulkInvoiceService do
       order.bill_address = order.ship_address
       order.save!
 
-      service.start_pdf_job([order.id])
-      Delayed::Job.last.invoke_job
+      service.start_pdf_job_without_delay([order.id])
 
       expect(service.invoice_created?(service.id)).to be_truthy
     end
--- a/spec/services/bulk_invoice_service_spec.rb
+++ b/spec/services/bulk_invoice_service_spec.rb
@@ -11,6 +11,17 @@ describe BulkInvoiceService do
 
       expect(Delayed::Job.last.payload_object.method_name).to eq :start_pdf_job_without_delay
     end
+
+    it "creates a PDF invoice" do
+      order = create(:completed_order_with_fees)
+      order.bill_address = order.ship_address
+      order.save!
+
+      service.start_pdf_job([order.id])
+      Delayed::Job.last.invoke_job
+
+      expect(service.invoice_created?(service.id)).to be_truthy
+    end
   end
 
   describe "#invoice_created?" do
--- a/spec/services/invoice_renderer_spec.rb
+++ b/spec/services/invoice_renderer_spec.rb
@@ -1,0 +1,15 @@
+require 'spec_helper'
+
+describe InvoiceRenderer do
+  let(:service) { described_class.new }
+
+  it "creates a PDF invoice" do
+    order = create(:completed_order_with_fees)
+    order.bill_address = order.ship_address
+    order.save!
+
+    result = service.render_to_string(order)
+
+    expect(result).to match /^%PDF/
+  end
+end
--- a/spec/services/invoice_renderer_spec.rb
+++ b/spec/services/invoice_renderer_spec.rb
@@ -3,13 +3,18 @@ require 'spec_helper'
 describe InvoiceRenderer do
   let(:service) { described_class.new }
 
-  it "creates a PDF invoice" do
+  it "creates a PDF invoice with two different templates" do
     order = create(:completed_order_with_fees)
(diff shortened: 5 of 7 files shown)


luisramos0:
build broken. looks related to the PR.

app/controllers/spree/admin/orders_controller_decorator.rb
```diff
@@ -37,9 +37,7 @@ def resend
   end
 
   def invoice
-    pdf = render_to_string pdf: "invoice-#{@order.number}.pdf",
-                           template: invoice_template,
-                           formats: [:html], encoding: "UTF-8"
+    pdf = InvoiceRenderer.new.render(@order)
```
luisramos0:
```suggestion
    pdf = InvoiceRenderer.new.render_to_string(@order)
```

app/services/invoice_renderer.rb
```diff
@@ -0,0 +1,29 @@
+class InvoiceRenderer
+  def render_to_string(order)
+    renderer.render_to_string(args(order))
```
sauloperez: This class would clearly benefit from moving `order` to an ivar populated in the constructor IMO

app/services/invoice_renderer.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def invoice_template
+    if Spree::Config.invoice_style2?
+      "spree/admin/orders/invoice2"
+    else
+      "spree/admin/orders/invoice"
+    end
+  end
+end
```
sauloperez: nice abstraction 👏

app/views/spree/admin/orders/_invoice_table2.html.haml
```diff
@@ -17,6 +17,9 @@
       %tr
         %td
           = render 'spree/shared/line_item_name', line_item: item
+          %br
+          %small
+            %em= raw(item.variant.product.supplier.name)
```
sauloperez: any chance we could pass something like `supplier_name` to this view? It'd reduce coupling a great deal but I'm not familiar with this view at all though.

app/services/invoice_renderer.rb
```diff
@@ -0,0 +1,29 @@
+class InvoiceRenderer
+  def render_to_string(order)
+    renderer.render_to_string(args(order))
+  end
+
+  def args(order)
```
sauloperez: this method is not used outside the class, right? in that case, it's better if we move it to private to keep the public API to the minimum.

spec/services/bulk_invoice_service_spec.rb
```diff
@@ -11,6 +11,17 @@
...
+    it "creates a PDF invoice" do
+      order = create(:completed_order_with_fees)
+      order.bill_address = order.ship_address
+      order.save!
+
+      service.start_pdf_job([order.id])
+      Delayed::Job.last.invoke_job
+
+      expect(service.invoice_created?(service.id)).to be_truthy
+    end
```
sauloperez: Doesn't it make more sense to have such integration-like spec in the controller and stick to unit-level in this spec file?

spec/services/invoice_renderer_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    result = service.render_to_string(order)
+    expect(result).to match /^%PDF/
+
+    allow(Spree::Config).to receive(:invoice_style2?).and_return true
+
+    alternative = service.render_to_string(order)
+    expect(alternative).to match /^%PDF/
+    expect(alternative).to_not eq result
+  end
```
sauloperez: IMO this deserves to be split in two tests, one for each `invoice_template` branch, which it'd make the class and the test more maintainable. Besides, as it is, looks more like a capybara test.

app/controllers/spree/admin/orders_controller_decorator.rb
mkllnk: Yes, thanks @luisramos0. I renamed that method at some point and forgot this call.

app/services/invoice_renderer.rb
mkllnk: I don't see it as clearly. I see great benefit in declaring all input variables for each method. It makes refactoring easier. It also allow to call the same renderer with different orders. The BulkInvoiceService should probably do that.

app/views/spree/admin/orders/_invoice_table2.html.haml
mkllnk: This is actually within a loop. The only variable for the template is `@order` and this template then goes through the items:
```rb
@order.line_items...each do |item|
  #...
end
```

We could put the content of the loop in a separate template, but I don't see the benefit here.

app/services/invoice_renderer.rb
mkllnk: The OrdersController calls this:
```
  def print
    render InvoiceRenderer.new.args(@order)
  end
```

I tried to move the `render` call into the InvoiceRenderer but that failed badly.

```
RuntimeError (ActionController::RackDelegation#content_type= delegated to @_response.content_type=, but @_response is nil: #<ApplicationController:0x005630be1fca30 @_routes=nil, @_action_has_layout=true, @_headers={"Content-Type"=>"text/html"}, @_status=200, @_request=nil, @_response=nil, @_config={}, @_prefixes=["application"], @_lookup_context=#<ActionView::LookupContext:0x005630be144340 @details_key=#<ActionView::LookupContext::DetailsKey:0x005630c488b0f8 @hash=-3205369017263900504>, @details=...
```

I think `render` tries to store variables like the content type in the request object of the controller. But in the service we just have a new ApplicationController without any request object stored in there. `render` is only to be called on a the active controller that actually received a request. This is another example why internal state is bad.

spec/services/bulk_invoice_service_spec.rb
mkllnk: I'm testing only the interface of this service here. What makes it integration-like? Do you mean the creation of an order in the database?

How would you unit-test the logic of this method?

spec/services/invoice_renderer_spec.rb
mkllnk: Haha, I had a comment there explaining why and then removed it. This is the second time that I removed a comment and then you ask me about it. I should never remove one of my comments again.

Here are some reasons:

- You can't test if the two results are different without rendering both results in the same test.
- Splitting the test means generating the pdf files several times which costs spec running time.

What do you think? Is it worth it splitting the spec into three and pay with 6 seconds more run time?

1. it renders an invoice pdf
2. it renders an alternative invoice pdf
3. it renders two different pdfs

mkllnk:
@sauloperez Thank you for your thorough review. I explained my decisions a bit more and it would be great to see what you think about it or what you can suggest.

spec/services/invoice_renderer_spec.rb
sauloperez: > This is the second time that I removed a comment and then you ask me about it. I should never remove one of my comments again.

:trollface:

sauloperez:
I'm quite aware that my review might be tiresome but I'd gladly help you if need it.

app/services/invoice_renderer.rb
sauloperez: I know there is a way to move render outside the controller although I never used it but I don't think we need it here. I was just talking about the `#args` method itself but I see what you did now :+1:

spec/services/bulk_invoice_service_spec.rb
mkllnk: Good point @sauloperez. This test isn't about Delayed Job at all. Your comment made me realise that I can invoke the method without delayed job to test it's actual execution. Have a look at my new commit. That's much clearer.

Also big thanks for the Sandy Metz talk about unit testing. Very simple and clear. My conclusion is though that I don't want to introduce doubles in this spec for several reasons:

- Doubles need to be kept in sync with the API of the thing they are imitating. Sandy mentions that you should have tools for that in order to make this stable and efficient. I don't think we have that in place. We don't have a double which verifies that the doubled method exists on the real object.
- This spec currently uses the public API of the service. Mocking requires knowledge of the internal workings. And that is bad, because changing the implementation of the service then requires changing the spec as well. The current spec should not need adjustments as long as the API stays the same.
- Sandy's main reason for doubles is to stay close to the object under test. We don't want to call `create_pdf` and then check for a file on the filesystem within the spec. The filesystem is far away. But my spec doesn't go far away at all. It uses only the API of the object under test.
- The remaining valid reason for doubles is speed. In this case though, I think that all the stability gained is worth the added 3 seconds.

What do you think?

app/views/spree/admin/orders/_invoice_table2.html.haml
mkllnk: I see, a very important point. I always hesitate to modify Spree code though. I'm also wondering if it's the responsibility of a line item to know the supplier name. Two ideas:

- Provide `LineItem#supplier` via the decorator. Less knowledge for the line item and the view can select what to display from the supplier: `raw(item.supplier.name)`.
- Or provide a helper that can select the supplier name of a line item, variant or product: `supplier_name(item)`.

If LineItem was our class I would definitely go for the first option, but since it's Spree, I'm not sure. What do you think?

spec/services/bulk_invoice_service_spec.rb
sauloperez: > We don't have a double which verifies that the doubled method exists on the real object.

That talk is old and RSpec's `instance_double` didn't exist at that time. Fortunately, things are better now and we already use it in our tests :tada: !

mkllnk: @sauloperez 
> I'm not a big fan of this handle_asynchronously magic

I totally agree! We should avoid that in future code.

> That talk is old and RSpec's instance_double and class_double didn't exist at that time. 

Thank you! I didn't know that. I definitely want to use that more. (Just not here. ;-) )

The outcome feels good now. :+1:

sstead:
**Testing notes**

- Hooray, a printed invoice now includes the producer name :)
- This is true regardless of the Invoice Settings (superadmin)

https://docs.google.com/document/d/1TwqaHrk6q6Ad49luLkoGC6LpCenTiXaijBsa4v8L8N8/edit?usp=sharing

PR #3445, opened 2019-02-06
[Spree Upgrade] Fix adjustments sections in admin order edit page

### LINKED ISSUE #2940: [Spree Upgrade] Included vs not included Adjustments in the edit order page (1 broken test)
There are no adjustments showing up on the edit order page in the following spec.
And we need to clarify what are order charges and what are price adjustments and how they map to the data model of order.adjustments.

```
  8) 
    As an administrator
    I want to manage orders
 as an enterprise manager viewing the edit page shows the order charges (non-tax adjustments)
     Failure/Error: expect(page).to have_selector "td", match: :first, text: adjustment.label
       expected to find visible css "td" with text "Shipping" within #<Capybara::Node::Element tag="tbody" path="//HTML[1]/BODY[1]/DIV[1]/DIV[1]/DIV[4]/DIV[1]/DIV[1]/DIV[1]/DIV[2]/DIV[2]/DIV[1]/DIV[1]/FIELDSET[1]/TABLE[1]/TBODY[1]"> but there were no matches
     # ./spec/features/admin/orders_spec.rb:244:in `block (6 levels) in <top (required)>'
     # ./spec/features/admin/orders_spec.rb:242:in `block (5 levels) in <top (required)>'
     # ./spec/features/admin/orders_spec.rb:241:in `block (4 levels) in <top (required)>'
```

and 

```
 9) 
    As an administrator
    I want to manage orders
 as an enterprise manager viewing the edit page shows the order taxes
     Failure/Error:
       within('table.index tbody#price-adjustments') do
         expect(page).to have_selector "td", match: :first, text: "Tax 1"
         expect(page).to have_selector "td.total", text: Spree::Money.new(10)
       end
     
     Capybara::ElementNotFound:
       Unable to find visible css "table.index tbody#price-adjustments"
     # ./spec/features/admin/orders_spec.rb:257:in `block (4 levels) in <top (required)>'
```

The problem will probably be in the order factory used. Some detail must be missing, probably how payment or shipment is setup in the factory order_with_taxes that needs to be adapted to spree 2.

luisramos0:
#### What? Why?

Closes #2940

Adding price_adjustments to a separate table in admin order edit page. This is mimicking what happens in v2.2. I dont think we should pull forward that commit from v2.2 because it depends on order associations not yet available in v2.0 and also because we may start customize this page to OFN needs.

#### What should we test?
Tax adjustments spec in #2940 (spec/features/admin/orders_spec.rb:266) is now green.

This will need to be manually tested in detail when we test v2.0 as a whole.

#### Dependencies
This is totally depend on #3194 so I left the commits in this PR. **This PR is composed of only one commit**, [this one](https://github.com/openfoodfoundation/openfoodnetwork/pull/3445/commits/3ed6aea9abed3de941c3b93823929e0a128f2b59).

### DIFF
--- a/app/views/spree/admin/orders/_form.html.haml
+++ b/app/views/spree/admin/orders/_form.html.haml
@@ -4,23 +4,8 @@
 
   = render :partial => "spree/admin/orders/shipment", :collection => @order.shipments, :locals => { :order => order }
 
-  - if order.adjustments.eligible.exists?
-    %fieldset.no-border-bottom
-      %legend= Spree.t('adjustments')
-      %table
-        %thead
-          %th= Spree.t('name')
-          %th= Spree.t('amount')
-        %tbody#order-charges.with-border{"data-hook" => "order_details_adjustments"}
-          - order.adjustments.eligible.each do |adjustment|
-            - next if ((adjustment.originator_type == 'Spree::TaxRate') and (adjustment.amount == 0)) || adjustment.originator_type == 'Spree::ShippingMethod'
-            %tr.total
-              %td
-                %strong
-                  = adjustment.label
-                  \:
-              %td.total.align-center
-                %span= adjustment.display_amount
+  = render :partial => "spree/admin/orders/_form/adjustments", :locals => { :adjustments => @order.price_adjustments, :order => order, :title => Spree.t(:line_item_adjustments)}
+  = render :partial => "spree/admin/orders/_form/adjustments", :locals => { :adjustments => @order.adjustments, :order => order, :title => Spree.t(:order_adjustments)}
 
   - if order.line_items.exists?
     %fieldset#order-total.no-border-bottom{"data-hook" => "order_details_total"}
--- a/app/views/spree/admin/orders/_form.html.haml
+++ b/app/views/spree/admin/orders/_form.html.haml
@@ -1,68 +1,35 @@
-%div
+%div{"data-hook" => "admin_order_form_fields"}
   - if @line_item.try(:errors).present?
-    = render partial: 'spree/shared/error_messages', locals: { target: @line_item }
-  = form_for @order, url: admin_order_url(@order), method: :put do |f|
-    %fieldset.no-border-top
-      = f.hidden_field :number
-      %table.index{"ng-show" => "distributionChosen()"}
-        %colgroup
-          %col{style: "width: 49%;"}/
-          %col{style: "width: 14%;"}/
-          %col{style: "width: 10%;"}/
-          %col{style: "width: 14%;"}/
-          %col{style: "width: 8%;"}/
-        %thead#line-items
-          %tr
-            %th
-              = t(:item_description)
-            %th.price
-              = t(:price)
-            %th.qty
-              = t(:qty)
-            %th.total
-              %span
-                = t(:total)
-            %th.orders-actions.actions
-        %tbody
-          = f.fields_for :line_items do |li_form|
-            = render partial: 'spree/admin/orders/line_item', locals: { f: li_form }
-        %tbody#subtotal.no-border-top
-          %tr#subtotal-row
-            %td{colspan: "3"}
-              %b
-                = t(:subtotal)
-                \:
-            %td.total.align-center
-              %span
-                = @order.display_item_total.to_html
-            %td.actions
-        %tbody#order-charges.no-border-top
-          - @order.adjustments.eligible.each do |adjustment|
-            %tr
-              %td{colspan: "3"}
+    = render :partial => 'spree/shared/error_messages', :locals => { :target => @line_item }
+
+  = render :partial => "spree/admin/orders/shipment", :collection => @order.shipments, :locals => { :order => order }
+
+  - if order.adjustments.eligible.exists?
+    %fieldset.no-border-bottom
+      %legend= Spree.t('adjustments')
+      %table
+        %thead
+          %th= Spree.t('name')
+          %th= Spree.t('amount')
+        %tbody#order-charges.with-border{"data-hook" => "order_details_adjustments"}
+          - order.adjustments.eligible.each do |adjustment|
+            - next if ((adjustment.originator_type == 'Spree::TaxRate') and (adjustment.amount == 0)) || adjustment.originator_type == 'Spree::ShippingMethod'
+            %tr.total
+              %td
                 %strong
                   = adjustment.label
                   \:
               %td.total.align-center
-                %span= adjustment.display_amount.to_html
-              %td.actions
-        %tbody#order-total.grand-total.no-border-top
-          %tr
-            %td{colspan: "3"}
-              %b
-                = t(:order_total)
-                \:
-            %td.total.align-center
-              %span#order_total
-                = @order.display_total.to_html
-            %td.actions
+                %span= adjustment.display_amount
 
-      = render partial: 'spree/admin/orders/_form/distribution_fields'
+  - if order.line_items.exists?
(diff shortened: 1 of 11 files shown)


app/views/spree/admin/orders/_form/_adjustments.html.haml
```diff
@@ -0,0 +1,14 @@
+- if adjustments.present?
```
sauloperez: what if instead, we move this conditional to wrap the lines 7 and 8 of app/views/spree/admin/orders/_form.html.haml? this would decouple this view from the details of adjustments and make it more maintainable. It feels like a responsibility that _adjustments.html.haml shouldn't be aware of and can save us some CPU cycles.

app/views/spree/admin/orders/_form/_adjustments.html.haml
```diff
@@ -0,0 +1,14 @@
...
+  %fieldset.no-border-bottom
+    %legend= title
+    %table>
+      %thead
+        %th= Spree.t('name')
+        %th= Spree.t('amount')
+      %tbody#order-charges.with-border
+        - adjustments.each do |adjustment|
+          - if (adjustment.originator_type != 'Spree::ShippingMethod') && !(adjustment.originator_type == 'Spree::TaxRate' && adjustment.amount == 0)
```
sauloperez: Along the same lines, this couples a lot this view to the inner details of the adjustment. If we don't want to render anything for these kinds of adjustments we can avoid rendering the table altogether right?

We could accomplish this by loading from DB or filtering only the adjustments we care about and passing those in the `adjustments` local. This would make things a lot more maintainable IMO.

app/views/spree/admin/orders/_form/_adjustments.html.haml
luisramos0: we would require 2 if statements in _form (see that "adjustments" here are different lists in _form)
if in spree v2.2. with shipment_adjustments, that would make 3 if statements.

app/views/spree/admin/orders/_form/_adjustments.html.haml
luisramos0: yes, I agree, I noticed that problem. this is a problem in the spree code.
my decision was not to deviate too much from the spree code...

do you think I should do some magic in _form to get only adjustments that pass the conditional?
to be honest, instead, I think we should aim to remove these conditionals, these views should just show what's in the DB...

app/views/spree/admin/orders/_form/_adjustments.html.haml
sauloperez: Yes, I know but that is something we would need to fix in `_form`. If it gets a bit messy is not `_adjustments.html.haml` responsibility.

app/views/spree/admin/orders/_form/_adjustments.html.haml
sauloperez: > do you think I should do some magic in _form to get only adjustments that pass the conditional?
to be honest, instead, I think we should aim to remove these conditionals, these views should just show what's in the DB... 

I would just load from DB those that we care about in this partial (shipping method nad tax rate) or do you mean we should display all kinds of adjustments?

app/views/spree/admin/orders/_form/_adjustments.html.haml
sauloperez: I created https://github.com/openfoodfoundation/openfoodnetwork/issues/3475 for this.

app/views/spree/admin/orders/_form/_adjustments.html.haml
sauloperez: I created https://github.com/openfoodfoundation/openfoodnetwork/issues/3476 for this

sauloperez:
:see_no_evil: I was totally convinced I was the second reviewer and this block due to my review

luisramos0:
we can still have @mkllnk's review and I can include Maikels feedback in a follow up PR.

app/views/spree/admin/orders/_form/_adjustments.html.haml
luisramos0: ok, thank for creating the issue. I think this can definitely be improved!

app/views/spree/admin/orders/_form/_adjustments.html.haml
luisramos0: :+1:

app/views/spree/admin/orders/_form/_adjustments.html.haml
luisramos0: yes, I meant that we should display all kinds of adjustments.

app/views/spree/admin/orders/_form/_adjustments.html.haml
```diff
@@ -0,0 +1,14 @@
+- if adjustments.present?
+  %fieldset.no-border-bottom
+    %legend= title
+    %table>
+      %thead
+        %th= Spree.t('name')
+        %th= Spree.t('amount')
+      %tbody#order-charges.with-border
```
kristinalim: It is now possible for the HTML ID `order-charges` to appear more than once in the page. I see that this is also an issue with [in the 2-2-stable version of the file](https://github.com/spree/spree/blob/2-2-stable/backend/app/views/spree/admin/orders/_adjustments.html.erb), but it isn't valid HTML.

app/views/spree/admin/orders/_form.html.haml
```diff
@@ -4,23 +4,8 @@
...
 
-  - if order.adjustments.eligible.exists?
-    %fieldset.no-border-bottom
-      %legend= Spree.t('adjustments')
-      %table
-        %thead
-          %th= Spree.t('name')
-          %th= Spree.t('amount')
-        %tbody#order-charges.with-border{"data-hook" => "order_details_adjustments"}
-          - order.adjustments.eligible.each do |adjustment|
```
kristinalim: The adjustments are no longer limited to those with `eligible: true`.

app/views/spree/admin/orders/_form/_adjustments.html.haml
luisramos0: yeah, good catch. I am not sure what to do? I feel tempted to add this to one of the tech debt issues that Pau created above. maybe #3475 ?

sauloperez: Sounds good to me @luisramos0

kristinalim: Hmm, it isn't tech debt though. It could get in quicker than #3475, so I would create a new issue for it. :slightly_smiling_face: I'll do that in a bit.

app/views/spree/admin/orders/_form.html.haml
luisramos0: ah, I just verified that adjustment.set_eligibility is removing adjustments with amount zero...
so, this needs to be fixed now. I will create a spree upgrade issue to fix this and the order-charges issue above.
Thanks for you review @kristinalim

luisramos0: #3477

app/views/spree/admin/orders/_form/_adjustments.html.haml
luisramos0: I just added this to #3477
both your points need to be fixed now @kristinalim they are not tech debt.

app/views/spree/admin/orders/_form.html.haml
kristinalim: @luisramos0 I also realized that the Enterprise Fee Summary still needs to be scoped to just `eligible: true`. :grin: Thanks.

PR #3449, opened 2019-02-06
[WIP] Use system version of chromedriver

mkllnk:
#### What? Why?

Our build is broken at the moment, because our chromedriver-helper is downloading the latest chromedriver which is not compatible with Semaphore's Chrome.

Here I remove the helper. We can add it with more time when we discussed the best solution to this problem.

#### What should we test?

Only automated tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -119,7 +119,6 @@ group :test, :development do
   gem 'letter_opener', '>= 1.4.1'
   gem 'timecop'
   gem 'selenium-webdriver'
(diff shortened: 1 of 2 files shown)


luisramos0:
you will need a bundle install to remove it from gemfile.lock Maikel

mkllnk:
Of course! Too quick in emergency mode!

mkllnk:
This quick fix didn't work.

PR #3451, opened 2019-02-07
Use localized currency symbols in Subs pages

### LINKED ISSUE #2875: [Subscriptions] Currency internationalization issue on subs
## Description

Currency symbol is $ on subscription pages, whatever the local currency setup.

## Expected Behavior

Currency adapt to local currency setup.

## Steps to Reproduce

1. On all screens of the subscription concerned (setup process when adding products and reviewing, "items" and "orders" tabs) you can see $ even on the French instance setup to €
capture du 2018-10-16 21-41-36

## Context

Reviewing the remaining things to be done to consider standing order v1 as done.

## Severity
S3, but not so relevant here as included in standing order project anyway

Matt-Yorkley:
#### What? Why?

Closes #2875

Shows localized currency symbols in the subscriptions UI.

#### What should we test?

Localized currency symbols should be shown in subscriptions.

#### Release notes

Fixed currency symbol localization in subscriptions.

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/admin/subscriptions/subscriptions.js.coffee
+++ b/app/assets/javascripts/admin/subscriptions/subscriptions.js.coffee
@@ -1,1 +1,1 @@
-angular.module("admin.subscriptions", ['ngResource','admin.indexUtils','admin.dropdown'])
+angular.module("admin.subscriptions", ['ngResource','admin.indexUtils','admin.dropdown', 'admin.utils'])
--- a/app/assets/javascripts/admin/utils/filters/localize_currency.js.coffee
+++ b/app/assets/javascripts/admin/utils/filters/localize_currency.js.coffee
@@ -1,0 +1,14 @@
+angular.module("admin.utils").filter "localizeCurrency", (currencyConfig)->
+  # Convert number to string currency using injected currency configuration.
+  (amount) ->
+    # Set country code (eg. "US").
+    currency_code = if currencyConfig.display_currency then " " + currencyConfig.currency else ""
+    # Set decimal points, 2 or 0 if hide_cents.
+    decimals = if currencyConfig.hide_cents == "true" then 0 else 2
+    # Set format if the currency symbol should come after the number, otherwise (default) use the locale setting.
+    format = if currencyConfig.symbol_position == "after" then "%n %u" else undefined
+    # We need to use parseFloat as the amount should come in as a string.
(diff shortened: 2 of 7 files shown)


sauloperez:
Is that an angular thing @mkllnk?

mkllnk:
It's our own Angular filter defined in `app/assets/javascripts/darkswarm/filters/localize_currency.js.coffee`. The I18n config is injected into every page. I think it takes into account the setting if the symbol should be before or after the number.

Matt-Yorkley:
@mkllnk It looks like the `localizeCurrency` filter is only available in `Darkswarm` modules. What's the best way to apply it to the backend? It can't be imported directly as it is, and we don't seem to have any current way of managing custom angular components that are shared between `Darkswarm` and `admin`? Hmm... actually it looks like each filter needs to be declared as part of a module, so the only option is to duplicate it under /admin?

mkllnk:
Cool. We don't have a good solution for code sharing between admin and darkswarm. I think the I18n code was just duplicated as well.

mkllnk:
Look what I found: https://github.com/openfoodfoundation/openfoodnetwork/blob/93aabf674152d157415a5a3ac867f81bceccb86c/app/assets/javascripts/admin/all.js#L55

Can we use that?

Matt-Yorkley:
That file is a bit of a hacky solution by the looks of it, and it's not a proper module:
```
# Declares the translation function t.
# You can use translate('login') or t('login') in Javascript.
window.translate = (key, options = {}) ->
  unless 'I18n' of window
    console.log 'The I18n object is undefined. Cannot translate text.'
    return key
  I18n.t(key, options)
window.t = window.translate
```

RachL:
@Matt-Yorkley @luisramos0 or @mkllnk can you tell me if I can stage this of should the code climate issue be fixed first?

mkllnk:
Usually I would request changes, but there are more style violations in the master branch. So I moved it into a tech-debt issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/3502

Let's proceed.

sauloperez:
I just approved it from Code Climate's UI. I tend to do that for these PRs whose violations that fall out of scope.

RachL:
Ready to go!

https://docs.google.com/document/d/1Qwf0QricsmntBNLXsF9sgNfxk18pGzISqxv8BSR4dlA/edit#

Issue #3383, opened 2019-01-24
[Flaky Spec] spec/features/admin/bulk_order_management_spec.rb:56


kristinalim:
## Description

This spec is unreliable: https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-2833/builds/5

## Context

Our Semaphore builds fail regularly because of these unreliable tests. If you encounter this in your build, please comment on this issue or add :+1: 

Error:

```
  As an Administrator
  I want to be able to manage orders in bulk
 listing orders displaying individual columns displays a column for order date
     Failure/Error: expect(page).to have_selector "td.date", text: o1.completed_at.strftime('%B %d, %Y'), visible: true
       expected to find visible css "td.date" with text "January 24, 2019" but there were no matches. Also found "", "", which matched the selector but not all filters.
     # ./spec/features/admin/bulk_order_management_spec.rb:58:in `block (4 levels) in <top (required)>'
```

sauloperez:
Is there something I can do to help here @kristinalim? I believe this is the flakiest spec file. Any idea about the root cause?

kristinalim:
@sauloperez I suspect there is a race condition somewhere, but I haven't been able to figure this out yet. Do you want to look into this?

sauloperez:
Yes, I would like to.

kristinalim:
Go ahead, @sauloperez. Thanks!

sauloperez:
@Matt-Yorkley I think you gave some thought to this as well. Would you mind sharing your ideas? Although it was much more unreliable in v2, I still see these specs failing from time to time.

Matt-Yorkley:
I'll take this on now, I've got a new idea to try...

sigmundpetersen:
hmmm still not good? https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/dependabot-bundler-combine_pdf-1-0-16/builds/1 😢

PR #3457, opened 2019-02-07
Soft delete customers

### LINKED ISSUE #1494: Delete a customer from your customer list impossible

sauloperez:
#### What? Why?

Closes #1494

This enables soft-deleting customers by means of `paranoia`. The same gem v2 uses. Therefore, I specified the same version Spree defines.

This allows disabling access to private shops while still showing any customer's orders and her details in reports.

#### What should we test?
* A customer can be removed from the customers' list regardless of the orders she might have placed (test with and without order)
* The removed customer can't access the private shop
* Reports still show the deleted customer's orders and her details

#### Release notes

Access to a private shop can be revoked by removing the customer from the list while still showing her orders from reports.

Changelog Category: Added

### DIFF
--- a/spec/controllers/admin/customers_controller_spec.rb
+++ b/spec/controllers/admin/customers_controller_spec.rb
@@ -181,7 +181,7 @@ describe Admin::CustomersController, type: :controller do
     end
 
     context 'when rendering json' do
-      context 'and the customer has no orders' do
+      context 'and the destroy succeeds' do
         it 'invokes before destroy callbacks' do
           allow(controller).to receive(:invoke_callbacks).with(:destroy, :after)
 
@@ -191,7 +191,7 @@ describe Admin::CustomersController, type: :controller do
 
         it 'returns a no_content status code' do
           spree_delete :destroy, id: customer, format: :json
-          expect(response).to have_http_status(:no_content)
+          expect(JSON.parse(response.body)['id']).to eq customer.id
         end
 
         it 'invokes after destroy callbacks' do
@@ -202,8 +202,11 @@ describe Admin::CustomersController, type: :controller do
         end
       end
 
-      context 'and the customer has orders' do
-        before { customer.orders << build(:order) }
+      context 'and the destroy fails' do
+        before do
+          allow(Customer).to receive(:find).with(customer.id.to_s).and_return(customer)
+          allow(customer).to receive(:destroy).and_return(false)
+        end
 
         it 'invokes before destroy callbacks' do
           allow(controller).to receive(:invoke_callbacks).with(:destroy, :fails)
@@ -214,7 +217,7 @@ describe Admin::CustomersController, type: :controller do
 
         it 'renders the errors' do
           spree_delete :destroy, id: customer, format: :json
-          expect(response.body).to match(I18n.t('admin.customers.destroy.has_associated_orders'))
+          expect(response.body).to match('errors')
         end
 
         it 'returns conflict HTTP status code' do
@@ -232,7 +235,7 @@ describe Admin::CustomersController, type: :controller do
     end
 
     context 'when rendering html' do
-      context 'and the customer has no orders' do
+      context 'and the destroy succeeds' do
         it 'invokes before destroy callbacks' do
           allow(controller).to receive(:location_after_destroy) { '/fix_location_after_destroy' }
           allow(controller).to receive(:invoke_callbacks).with(:destroy, :after)
@@ -256,8 +259,11 @@ describe Admin::CustomersController, type: :controller do
         end
       end
 
-      context 'and the customer has orders' do
-        before { customer.orders << build(:order) }
+      context 'and the destroy fails' do
+        before do
+          allow(Customer).to receive(:find).with(customer.id.to_s).and_return(customer)
+          allow(customer).to receive(:destroy).and_return(false)
+        end
 
         it 'invokes before destroy callbacks' do
           allow(controller).to receive(:location_after_destroy) { '/fix_location_after_destroy' }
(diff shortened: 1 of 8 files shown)


spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -171,4 +171,158 @@ def create_customer(enterprise)
...
+          allow(controller).to receive(:invoke_callbacks).with(:destroy, :after)
+
+          expect(controller).to receive(:invoke_callbacks).with(:destroy, :before)
+          spree_delete :destroy, id: customer, format: :html
+        end
+
+        it 'raises due to missing #location_after_destroy' do
+          expect {
+            spree_delete :destroy, id: customer, format: :html
+          }.to raise_error(NameError, /location_after_destroy/)
```
sauloperez: any how this method was implemented @mkllnk ? I couldn't find any trace including the metaprogramming I could think of.

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -171,4 +171,158 @@ def create_customer(enterprise)
...
+      context 'and the destroy fails' do
+        before do
+          allow(Customer).to receive(:find).with(customer.id.to_s).and_return(customer)
+          allow(customer).to receive(:destroy).and_return(false)
+        end
+
+        it 'raises due to missing template' do
+          expect {
+            spree_delete :destroy, id: customer, format: :js
+          }.to raise_error(ActionView::MissingTemplate)
```
sauloperez: It turns out that we never used checked the scenario the destroy didn't work when requesting it with JS. We need to cover it somehow.

sauloperez:
The only failing spec must be related to https://github.com/openfoodfoundation/openfoodnetwork/issues/3383.

mkllnk:
@sauloperez Great to see a quick solution. Thank you.

In my comment https://github.com/openfoodfoundation/openfoodnetwork/issues/1494#issuecomment-459636281 expressed concerns about deleting customers. People who ordered from an enterprise are their customers. That's why I suggested to introduce a new flag on the customer used for private shop fronts. That would separate the concerns "What is a customer" and "Who is allowed to see my shop front?"

I guess you could argue that a deleted_at field can be used as a "don't allowed to see private shop" flag. It's just not that obvious. What was your rationale?

sauloperez:
> I guess you could argue that a deleted_at field can be used as a "don't allowed to see private shop" flag. It's just not that obvious. What was your rationale?

As I saw on your explanation, it was what customers meant to solve. Associating a user with an enterprise is the way to grant access. If I don't allow you to purchase on my shop anymore, I don't want you on the list, therefore the association needs to be removed.

Besides, we enable other enterprises to manage their customers list as they want while OFN ensures the underlying data integrity.

Do you see any scenario this solution would break? Am I missing something?

luisramos0:
I think both approaches are fine but the explicit relationship flagging the permission is better than using the presence of the entry.
if a flag is used to grant access the presence of the customer record will be got connecting to orders through user.
in this space, I think what really needs to be dropped is the relationship between order and customer, it should go through user at all times.

sauloperez:
> in this space, I think what really needs to be dropped is the relationship between order and customer, it should go through user at all times.

so if I understood your point, you mean that customers table is redundant since from the order you can reach both the enterprise and the user, right?

luisramos0:
no, I dont think customers is redundant, it's the normal relationship table (that can hold data like permissions and other stuff) between user and enterprise. to remove customer from order and use only user, we need to sort out some edge cases like guest users and creating orders and subscriptions for users that have not registered yet...

mkllnk:
I agree with Luis. Both approaches are okay.

The customer model has been introduced for reporting. We may need to include the scope `with_deleted` in reports because they need to consider all customers regardless of their permission to access the shop.

Customer records are also used for tags. I actually don't know all the use cases tags and if this could break something. We should probably investigate this before going down the soft-delete route.

Just thinking, what if you deny a customer from your shop but then want to assign them a customer code for reporting? I guess that scenario would be broken.

And what do we do if customers are re-added? Do we need some logic to un-delete existing customer records?

I like to avoid these kind of questions by not mixing concerns and treating shop-access as a separate feature to customer records.

sauloperez:
> The customer model has been introduced for reporting. We may need to include the scope `with_deleted` in reports because they need to consider all customers regardless of their permission to access the shop.

That's what I thought but they do work. I still don't fully understand why.

> Customer records are also used for tags. I actually don't know all the use cases tags and if this could break something. We should probably investigate this before going down the soft-delete route.

As far as I know there is no problem with tags and tag rules but I'll check.

sauloperez:
Thanks a lot for your insightful comments @mkllnk . I close this until we reach an agreement in #1494 . I could distract us too much from the Spree upgrade.

PR #3462, opened 2019-02-08
Delete empty controller class eval and spec

luisramos0:
#### What? Why?
These are useless left overs from 0ad2978926f17e663b96bafffe4f8365e660af50
Correct? Is there ever a reason to have an empty class eval?

#### What should we test?
I dont think this controller is being used as the commit above says the BOM is using the admin ctrl not this api controller. The auto tests are validating that it's not a problem. We can make a simple sanity check on the BOM.

#### Release notes
Changelog Category: Removed
Removed old useless code.

### DIFF
--- a/app/controllers/spree/api/orders_controller_decorator.rb
+++ b/app/controllers/spree/api/orders_controller_decorator.rb
@@ -1,7 +1,0 @@
-Spree::Api::OrdersController.class_eval do
-
-  # We need to add expections for collection actions other than :index here
-  # because Spree's API controller causes authorize_read! to be called, which
-  # results in an ActiveRecord::NotFound Exception as the order object is not
-  # defined for collection actions
-end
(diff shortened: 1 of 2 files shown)


sauloperez:
Definitely useless :+1:

RachL:
I'm sorry but I want to be sure: BOM = Bulk Order Management = /admin/orders   and not admin/orders/bulk_management right @luisramos0 ?

luisramos0:
it's good that you ask!
BOM is bulk order management.
But actually, we should test both BOM and orders page.
Thanks!

RachL:
BOM and orders page seem to work fine.

testing notes : https://docs.google.com/document/d/1eGENj7_7VISCi-nOqJcR0xy--Bp3557tD9HgrMY1Hhc/edit#

PR #3463, opened 2019-02-08
Import Catalan translation from Transifex

sauloperez:
#### What? Why?

Although we've started translating to Catalan very long ago and focused on it recently, we never imported the file in the app. The language menu showed "English" for the "ca" locale and clicking on it switched to English as well instead of Catalan.

As we already had the Spanish translation, we moved along with it (too many things) delaying the end of the Catalan one forever. Truth is that all Katuma users' mother tongue is Catalan so they will be more than happy.

#### What should we test?
It should be possible to switch to Catalan when testing this from staging.katuma.org. Catalan 101: the menu should look like

screenshot from 2019-02-08 15-31-22

#### Release notes

Added Catalan translation which will work on Katuma.

Changelog Category: Added

### DIFF
--- a/config/locales/ca.yml
+++ b/config/locales/ca.yml
@@ -1,0 +1,2649 @@
+ca:
+  language_name: "Català"
+  activerecord:
+    attributes:
+      spree/order:
+        payment_state: Estat del Pagament
+        shipment_state: Estat de la Tramesa
+        completed_at: Completat a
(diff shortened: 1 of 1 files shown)


sauloperez:
:muscle:

sauloperez:
Working! :tada:

PR #3470, opened 2019-02-11
3426 [Flaky Spec] Address timing issues in enterprise users feature specs

### LINKED ISSUE #3426: [Flaky Spec] /spec/features/admin/enterprise_roles_spec.rb:131,147
## Description

These specs are unreliable: 
rspec ./spec/features/admin/enterprise_roles_spec.rb:131
rspec ./spec/features/admin/enterprise_roles_spec.rb:147

https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3209/builds/12

## Context

Error:

```
  4) 
  As an Administrator
  I want to manage relationships between users and enterprises
 as a site administrator using the enterprise managers interface can invite unregistered users to be managers
     Failure/Error: expect(page).to have_selector "table.managers"
       expected to find visible css "table.managers" but there were no matches. Also found "", which matched the selector but not all filters.
     # ./spec/features/admin/enterprise_roles_spec.rb:102:in `block (4 levels) in <top (required)>'
```

kristinalim:
Address timing issues in enterprise users feature specs.

#### What? Why?

Closes #3426 

#### What should we test?

Semaphore build should be green.

In the long term, this test should also not fail.

#### Release notes

* Address intermittently failing feature tests

Changelog Category: Fixed

### DIFF
--- a/spec/features/admin/enterprise_roles_spec.rb
+++ b/spec/features/admin/enterprise_roles_spec.rb
@@ -98,8 +98,7 @@ feature %q{
       before do
         click_link 'Enterprises'
         click_link 'Test Enterprise'
-        within('.side_menu') { click_link 'Users' }
-        expect(page).to have_selector "table.managers"
+        navigate_to_enterprise_users
       end
 
       it "lists managers and shows icons for owner, contact, and email confirmation" do
@@ -131,7 +130,7 @@ feature %q{
       it "shows changes to enterprise contact or owner" do
         select2_select user2.email, from: 'receives_notifications_dropdown'
         within('#save-bar') { click_button 'Update' }
-        within('.side_menu') { click_link 'Users' }
+        navigate_to_enterprise_users
 
         within 'table.managers' do
           within "tr#manager-#{user1.id}" do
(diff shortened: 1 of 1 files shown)


spec/features/admin/enterprise_roles_spec.rb
```diff
@@ -170,9 +172,16 @@
...
 
-
   private
 
+  def navigate_to_enterprise_users
+    within ".side_menu" do
+      click_link "Users"
+    end
+
+    expect(page).to have_selector "table.managers"
```
sauloperez: I wouldn't add this here as we're coupling the actual navigation with an assertion. As a result, in line 133 above we're adding an extra step that we don't need. That being said, I don't think line 101 should need it.

sauloperez: what do you think @kristinalim ?

kristinalim: @sauloperez I moved check for table presence out of the helper method. Build is green, and I moved this back to Code Review.

Issue #3427, opened 2019-02-01
[Spree Upgrade] Fix spree/admin/invoices_controller_spec (1 broken spec)


luisramos0:
Since the new bulk invoice feature was merged into 2-0 we have a new brokens spec in the v2 branch: spec/controllers/spree/admin/invoices_controller_spec.rb:25

This is the error:
```
 2) Spree::Admin::InvoicesController#poll when the file is available returns true
     Failure/Error: spree_get :poll, invoice_id: invoice_id
     
     Errno::ENOENT:
       No such file or directory @ rb_sysopen - /home/runner/openfoodnetwork/tmp/cache/paralleltests/3B7/CD1/.permissions_check.46948123415300.3175.131088
     # ./spec/controllers/spree/admin/invoices_controller_spec.rb:27:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Errno::ENOENT:
     #   No such file or directory @ rb_file_s_stat - /home/runner/openfoodnetwork/tmp/cache/paralleltests/3B7/CD1/spree%2Fapp_configuration%2Fredirect_https_to_http
     #   ./spec/controllers/spree/admin/invoices_controller_spec.rb:27:in `block (4 levels) in <top (required)>'
```

PR #3471, opened 2019-02-11
[Spree Upgrade] Make Bulk Invoices part of spree upgrade phase 2

luisramos0:
#### What? Why?
Move broken bulk invoices test to xdescribe in 2-0-stable branch so we can focus on core features first.

I have moved #3427 under phase 2 epic #2953 

#### Product Import quick fix
I am using this PR to move ProductImporter spec back to xdescribe. It was moved to describe by mistake [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/3355/files#diff-12f58c2e71bafa579c28e0f9ef45b8d4L4) as there's still one broken test in this spec that will be fixed as part of #2783, we can then remove the xdescribe.

#### What should we test?
spec should not be broken is this build (I have verified the build and the spec is not failing).

### DIFF
--- a/spec/controllers/spree/admin/invoices_controller_spec.rb
+++ b/spec/controllers/spree/admin/invoices_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe Spree::Admin::InvoicesController, type: :controller do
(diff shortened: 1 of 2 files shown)


luisramos0:
thank you @kristinalim  :-D

PR #3481, opened 2019-02-13
Simplify specs with stripe factory

mkllnk:
#### What? Why?

While working on fixing payment methods for Spree 2 I noticed that the `stripe_payment_method` factory doesn't work without additional parameters. A lot of specs share some duplicate code to set these parameters. This PR removes the duplication.

#### What should we test?

Automated tests only.

#### Release notes

Changed: Improved our automated tests.

Changelog Category: Changed

### DIFF
--- a/spec/features/consumer/shopping/checkout_spec.rb
+++ b/spec/features/consumer/shopping/checkout_spec.rb
@@ -160,8 +160,7 @@ feature "As a consumer I want to check out my cart", js: true, retry: 3 do
         let!(:stripe_pm) do
           create(:stripe_payment_method,
             distributors: [distributor],
-            name: "Stripe",
-            preferred_enterprise_id: distributor.id)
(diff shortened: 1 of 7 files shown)


spec/features/consumer/shopping/checkout_spec.rb
```diff
@@ -160,8 +160,7 @@ def fill_out_form
         let!(:stripe_pm) do
           create(:stripe_payment_method,
             distributors: [distributor],
-            name: "Stripe",
-            preferred_enterprise_id: distributor.id)
+            name: "Stripe")
```
sauloperez: No need for this name. It is the default.

mkllnk: Correct, the name actually doesn't matter for this spec at all. I removed it. Have a look at my new commit.

PR #3489, opened 2019-02-14
[Spree Upgrade] Fix database migrations by re-ordering migrations correctly

### LINKED ISSUE #3301: [Spree Upgrade] CreateDefaultStock migration fails with sample data
## Description

The migration CreateDefaultStock fails with test data.

## Expected Behavior

All migrations work with all data.

## Actual Behavior

```
==  CreateDefaultStock: migrating =============================================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::UndefinedColumn: ERROR:  column spree_stock_locations.active does not exist
LINE 1: ...d" WHERE "spree_stock_items"."variant_id" = 1 AND "spree_sto...
                                                             ^
: SELECT "spree_stock_items".* FROM "spree_stock_items" INNER JOIN "spree_stock_locations" ON "spree_stock_locations"."id" = "spree_stock_items"."stock_location_id" WHERE "spree_stock_items"."variant_id" = 1 AND "spree_stock_locations"."active" = 't'
/home/maikel/code/openfoodnetwork/app/models/concerns/variant_stock.rb:44:in `count_on_hand'
/home/maikel/code/openfoodnetwork/db/migrate/20180426145632_create_default_stock.spree.rb:9:in `block in up'
/home/maikel/code/openfoodnetwork/db/migrate/20180426145632_create_default_stock.spree.rb:7:in `each'
/home/maikel/code/openfoodnetwork/db/migrate/20180426145632_create_default_stock.spree.rb:7:in `up'
/home/maikel/.rbenv/versions/2.1.5/bin/bundle:23:in `load'
/home/maikel/.rbenv/versions/2.1.5/bin/bundle:23:in `<main>'

Caused by:
ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR:  column spree_stock_locations.active does not exist
...

Caused by:
PG::UndefinedColumn: ERROR:  column spree_stock_locations.active does not exist
```

## Steps to Reproduce

1. `git checkout master`
2. `bundle exec rake db:reset`
3. `bundle exec rake openfoodnetwork:dev:load_sample_data`
4. `git checkout 2-0-stable`
5. `bundle exec rake db:migrate`

## Context

Working on the Spree upgrade.

## Possible Fix

Define all needed models within the migration, not using the models defined in the code.

luisramos0:
#### What? Why?

Closes #3301

Fix database migrations by moving createdefaultstock after add_active_field_to_stock_locations

Migration createdefaultstock needs spree_stock_locations.active 
So I moved add_active_field_to_stock_locations before it. It was just 5 migrations after.

tbh I dont know if this is allowed/ok, but it fixes the problem.

#### What should we test?
db:migrate with sample data should work. see issue for details.


sauloperez:
But why wasn't this fixed in Spree itself? could you find any clue about whether got eventually fixed somehow in an upcoming version?

kristinalim:
Seeing in Spree `2-0-stable` that the migrate file has had many changes: [file](https://github.com/spree/spree/blob/2-0-stable/core/db/migrate/20130213191427_create_default_stock.rb), [commit history](https://github.com/spree/spree/commits/2-0-stable/core/db/migrate/20130213191427_create_default_stock.rb)

I agree with @sauloperez about ideally fixing Spree issues closely to how Spree addressed them. But considering above, for this particular type of issue, it doesn't seem sustainable. So, for this, I prefer the approach of defining the models in the DB migrations suggested by @mkllnk.

mkllnk:
This is the diff to Spree master where they haven't fixed this problem yet.
```diff
diff --git a/core/db/migrate/20130213191427_create_default_stock.rb b/core/db/migrate/20130213191427_create_default_stock.rb
index 1efda239af..ce3ac95bb1 100644
--- a/core/db/migrate/20130213191427_create_default_stock.rb
+++ b/core/db/migrate/20130213191427_create_default_stock.rb
@@ -1,12 +1,20 @@
-class CreateDefaultStock < ActiveRecord::Migration
+class CreateDefaultStock < ActiveRecord::Migration[4.2]
   def up
+    unless column_exists? :spree_stock_locations, :default
+      add_column :spree_stock_locations, :default, :boolean, null: false, default: false
+    end
+
     Spree::StockLocation.skip_callback(:create, :after, :create_stock_items)
-    Spree::StockItem.skip_callback(:save, :after, :process_backorders)
-    location = Spree::StockLocation.create(name: 'default')
-    Spree::Variant.all.each do |variant|
-      stock_item = location.stock_items.build(variant: variant)
+    Spree::StockLocation.skip_callback(:save, :after, :ensure_one_default)
+    location = Spree::StockLocation.new(name: 'default')
+    location.save(validate: false)
+
+    Spree::Variant.find_each do |variant|
+      stock_item = Spree::StockItem.unscoped.build(stock_location: location, variant: variant)
       stock_item.send(:count_on_hand=, variant.count_on_hand)
-      stock_item.save!
+      # Avoid running default_scope defined by acts_as_paranoid, related to #3805,
+      # validations would run a query with a delete_at column that might not be present yet
+      stock_item.save! validate: false
     end
 
     remove_column :spree_variants, :count_on_hand
@@ -15,7 +23,7 @@ class CreateDefaultStock < ActiveRecord::Migration
   def down
     add_column :spree_variants, :count_on_hand, :integer
 
-    Spree::StockItem.all.each do |stock_item|
+    Spree::StockItem.find_each do |stock_item|
       stock_item.variant.update_column :count_on_hand, stock_item.count_on_hand
     end
 
@@ -23,4 +31,3 @@ class CreateDefaultStock < ActiveRecord::Migration
     Spree::StockItem.delete_all
   end
 end
```

sauloperez:
you mean that is their final version, right? Let's patch it on our fork then. But I wouldn't go and fix all migrations embedding the models just the ones that break.

kristinalim:
@sauloperez Yes, just the ones that break.

luisramos0:
Looks like you guys have come to an agreement but I dont understand what it is. Do you approve the PR or is there another way you want to see the problem solved?

kristinalim:
@luisramos0 The opinion from the reviews is:

1. The timestamps should not be changed. Instead...
2. The failing DB migration scripts should be fixed by defining the models there.
3. The corresponding DB migration script _templates_ in the Spree fork should also be updated.

sauloperez:
I took me a while to understand 3. The point is that migrations added by a gem need to be installed in OFN, right? I'm not used to it and I nearly forgot.

luisramos0:
ok, can you confirm the solution is:
point 2 - add the StockLocation object without the active property to the createdefaultstock migration.
point 3 - copy the same change [here](https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/core/db/migrate/20130213191427_create_default_stock.rb)?

sauloperez:
If I understood @kristinalim correctly what you say is right but we need to backport that final version of the create_default_stock migration to our Spree fork, apply our fix and then install them as described under "4.2 Engine setup" in https://guides.rubyonrails.org/v3.2/engines.html.

According to that guide

> When run the next time, it will only copy over migrations that haven’t been copied over already. 

I think we'll need to remove the current one first.

luisramos0:
that doesn't sound simple. what is the benefit of point 3 again?
from what I understand it will make our spree fork migrations correct, but no one will ever use them again, right?

mkllnk:
With point 2 Kristina was referring to the recommendation from the [Migrations Guide](https://guides.rubyonrails.org/v3.2/migrations.html#using-models-in-your-migrations). The benefit is that we don't have to change timestamps. We don't have to worry about the order of migrations as much. We follow common code styles. Even if the migration is not used later on, sometimes people go back and read them to understand what happened. Declaring the model within the migration makes it easier to understand how the model looked like when the migration was run.

It's a bit tricky with this migration because it uses several models and their logic:

- Spree::StockLocation
- Spree::StockItem
- Spree::Variant

Usually we would need a minimal implementation of all of these. But maybe in this context we just implement the offending one to fix the issue.

mkllnk:
If the migrations in the Spree code are not run, I wouldn't bother fixing them.

luisramos0:
Re point 2, I created PR #3552 as an alternative to this PR. It's not a problem in the model, it's in the quantifier's code. Please let me know your thoughts.

re point 3, if it's only about making our fork of spree correct, I dont see the value of doing it.

sauloperez:
> re point 3, if it's only about making our fork of spree correct, I dont see the value of doing it.

IMO we should only do it if it fixes an error. I thought that was the case.

luisramos0:
replaced by #3552

PR #3496, opened 2019-02-14
[Spree Upgrade] Remove allow_backorders from product stock verification

### LINKED ISSUE #2886: [Spree Upgrade] Replace usage of Config:allow_backorders with product.on_demand
We need to replace all occurrences of Spree::Config::allow_backorders related to products with product.on_demand.
This happens in app/helpers/add_to_cart_helper.rb

luisramos0:
#### What? Why?

Closes #2886 

This can be done because product.has_stock? already uses product.on_demand and in v2 allow_backorders becomes on_demand.

#### What should we test?
The build looks good, but it's not green yet.

### DIFF
--- a/app/helpers/add_to_cart_helper.rb
+++ b/app/helpers/add_to_cart_helper.rb
@@ -1,8 +1,4 @@
 module AddToCartHelper
-  def product_out_of_stock
-    !@product.has_stock? && !Spree::Config[:allow_backorders]
-  end
-
   def distributor_available_for?(order, product)
     DistributionChangeValidator.new(order).distributor_available_for?(product)
   end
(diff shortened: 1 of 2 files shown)


sauloperez:
:clap:

luisramos0:
I was planning to add more stuff to this PR but I ended up creating other PRs targeting master. This is ready for review.

luisramos0:
rebased for a green build.

mkllnk:
Which Spree upgrade branches are going to be tested and which ones are merged straight away?

luisramos0:
I think we should keep merging straight away until we get to a point where it makes sense to start collecting feedback from individual PRs. Right now we only want overall feedback.

I also think we need to go through ALL the upgrade PRs merged so far and make a list of things that we need to verify. This list can then be passed to testers for validation. Maybe all devs involved should do similar exercise of thinking: "what is it that we need to validate that auto tests may not be validating?"

what do you think is the best approach @mkllnk ?

mkllnk:
Sounds good. We don't need to do testing while we know that there is a lot broken.

PR #3505, opened 2019-02-15
Adapt scope_variant_to_hub to new VO rules where on_demand=nil means use_producer_settings

luisramos0:
#### What? Why?

This is the continuation of #3198 where on_demand nil became explicitly use_producer_settings.
I specially like the new variant_override.stock_overridden? method where this rule becomes explicit.

#### What should we test?
We need to test inventory and possible scenarios where count_on_hand in the inventory is used when the use_producer_settings is set.

#### Release notes
Changelog Category: Changed
Inventory rules are now more strict about using producer settings in inventory where stock levels are always ignored.

This was detected as part of the spree upgrade when verifying inventory logic in v2. It is related to upcoming PR #3504

### DIFF
--- a/app/models/variant_override.rb
+++ b/app/models/variant_override.rb
@@ -62,6 +62,10 @@ class VariantOverride < ActiveRecord::Base
     count_on_hand.present?
   end
 
+  def use_producer_settings?
+    on_demand.nil?
+  end
+
   def decrement_stock!(quantity)
     if stock_overridden?
       decrement! :count_on_hand, quantity
--- a/app/models/variant_override.rb
+++ b/app/models/variant_override.rb
@@ -57,7 +57,7 @@ class VariantOverride < ActiveRecord::Base
   end
 
   def stock_overridden?
-    count_on_hand.present?
+    on_demand == false && count_on_hand.present?
   end
 
   def decrement_stock!(quantity)
--- a/app/models/variant_override.rb
+++ b/app/models/variant_override.rb
@@ -62,7 +62,7 @@ class VariantOverride < ActiveRecord::Base
     count_on_hand.present?
   end
 
-  def use_producer_settings?
+  def use_producer_stock_settings?
     on_demand.nil?
   end
 
--- a/app/models/variant_override.rb
+++ b/app/models/variant_override.rb
@@ -57,7 +57,9 @@ class VariantOverride < ActiveRecord::Base
   end
 
   def stock_overridden?
-    on_demand == false && count_on_hand.present?
+    # If count_on_hand is present, it means on_demand is false
+    #   See StockSettingsOverrideValidation for details
+    count_on_hand.present?
   end
 
   def decrement_stock!(quantity)
--- a/lib/open_food_network/scope_variant_to_hub.rb
+++ b/lib/open_food_network/scope_variant_to_hub.rb
@@ -29,15 +29,19 @@ module OpenFoodNetwork
       end
 
       def count_on_hand
-        return super unless @variant_override.andand.stock_overridden?
-
-        @variant_override.count_on_hand
+        if @variant_override.present? && @variant_override.stock_overridden?
+          @variant_override.count_on_hand
+        else
+          super
+        end
       end
 
       def on_demand
-        return super if @variant_override.andand.on_demand.nil?
-
-        @variant_override.andand.on_demand
+        if @variant_override.present? && !@variant_override.use_producer_settings?
+          @variant_override.on_demand
+        else
+          super
+        end
       end
 
       def decrement!(attribute, by = 1)
--- a/lib/open_food_network/scope_variant_to_hub.rb
+++ b/lib/open_food_network/scope_variant_to_hub.rb
@@ -29,21 +29,15 @@ module OpenFoodNetwork
       end
 
       def count_on_hand
-        @variant_override.andand.count_on_hand || super
+        return super unless @variant_override.andand.stock_overridden?
+
+        @variant_override.count_on_hand
       end
 
       def on_demand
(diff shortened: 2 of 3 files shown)


lib/open_food_network/scope_variant_to_hub.rb
```diff
@@ -29,21 +29,15 @@ def in_stock?
       end
 
       def count_on_hand
-        @variant_override.andand.count_on_hand || super
+        return super unless @variant_override.andand.stock_overridden?
```
sauloperez: It's been a while I don't read VOs code but `@variant_override` is set at this point, right? then we don't need to `andand` defensive programming. The same happens with all the other methods using it.

luisramos0: no, @variant_override is nil for variants that are not overriden (this code will be executed for all calls to variant.count_on_hand :-)
I think the alternative is something like:
```
return super unless @variant_override.present? && @variant_override.stock_overridden?
```
andand is shorter.

luisramos0:
Interesting... you are correct Maikel!

regarding the "big comment in variant override", it's already there inside concern StockSettingsOverrideValidation, it's pretty easy to follow actually.

Now that I re-read the code and my PR, I want to remove the change in variant_override.stock_overridden? and just leave a comment, please see my new commit.

It's all about these constraints being in the readers mind or not :-)

mkllnk:
> regarding the "big comment in variant override", it's already there inside concern StockSettingsOverrideValidation, it's pretty easy to follow actually.

Where is the comment? I can't find it in https://github.com/openfoodfoundation/openfoodnetwork/blob/ef786adcfce6bb37adf6777bf1bb3e4111413105/app/models/concerns/stock_settings_override_validation.rb.

lib/open_food_network/scope_variant_to_hub.rb
mkllnk: I was wondering why you didn't use `andand` in the next line and it took me half a minute to make sure that this logic is correct. I would have found this version easier to read:
```rb
if @variant_override.andand.stock_overridden?
  @variant_override.count_on_hand
else
  super
end
```
A bit more rubyish, more functional, less procedural. But maybe that's just my preference?

lib/open_food_network/scope_variant_to_hub.rb
```diff
@@ -29,21 +29,15 @@ def in_stock?
...
-            # If we're overriding the stock level of an on_demand variant, show it as not
-            # on_demand, so our stock control can take effect.
-            false
-          else
-            super
-          end
-        else
-          @variant_override.andand.on_demand
-        end
+        return super if @variant_override.andand.on_demand.nil?
```
mkllnk: I'm wondering if we should introduce an alias here:
```rb
def use_producer_settings?
  on_demand.nil?
end
```

It could also be:
```rb
def active?
  !on_demand.nil?
end
```

Not sure.

luisramos0: yeah, I like it but I am not sure. I ended up following the first option.

lib/open_food_network/scope_variant_to_hub.rb
luisramos0: I ended up making count_on_hand and on_demand be a simple if else statement and without using andand. see new commit.

luisramos0:
thanks for your detailed review @mkllnk 

regarding the comment, I am sorry, I was not clear, I meant that StockSettingsOverrideValidation is itself a good "comment". it explains the logic pretty well.

yes, this PR is not changing anything just adapting to the new rules by removing unreachable code (mostly in on_demand) and fixing a test that was testing that unreachable code (on_demand nil and count_on_hand present).
so, the main goal is to make scope_to_variant.on_demand simpler.
From your comment I realize the spec I changed is itself testing an invalid case (it works because we are not saving the VO).

So, I changed the spec and I followed both your suggestions above. New commit. Thanks!

app/models/variant_override.rb
```diff
@@ -57,9 +57,15 @@ def self.decrement_stock!(hub, variant, quantity)
   end
 
   def stock_overridden?
+    # If count_on_hand is present, it means on_demand is false
+    #   See StockSettingsOverrideValidation for details
     count_on_hand.present?
   end
 
+  def use_producer_settings?
```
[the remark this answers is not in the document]
kristinalim: Could you also update the line in `StockSettingsOverrideValidation` doing a `nil` check to use this instead?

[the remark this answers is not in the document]
luisramos0: I thought about changing it in StockSettingsOverrideValidation but I liked the fact that those methods had the three options explicit: on_demand = nil, on_demand = true and on_demand = false. But well, I followed your suggestion now :-)

I'll click the new button "re-request a review" above :-)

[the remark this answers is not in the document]
luisramos0: ah, I only saw #3531 afterwards...
I reverted the change in StockSettingsOverrideValidation, with the comment on 3531, StockSettingsOverrideValidation should definitely have on_demand.nil?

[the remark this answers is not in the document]
kristinalim: Okay, that makes sense to me @luisramos0. :+1: Approving this PR once build is green.

kristinalim:
**Correction:** Safe to move to Test Ready

RachL:
Inventory, is looking good! This is good to go.

Testing notes: https://docs.google.com/document/d/1B1R247A2iRdQmRzsMzTzvOyg32oI3NDK5MEcZpcFkEk/edit#

luisramos0:
thanks @RachL

PR #3506, opened 2019-02-15
Sanitize non-parsable BOM characters inserted by Excel

### LINKED ISSUE #3414: Import validation panel gives incorrect information -  column names in the wrong places
## Description

When doing a product import I used the product import template 

to create an import file that looks like this

The import validation then looks like this . . 

It tells me everything is fine, but the column names are all wrong. 
* The Producer does not appear
* Producer column contains SKU
* SKU column contains Product name
etc

The data are imported correctly then but the display is very confusing and users feels they can't do the import and it's bugging.

## Expected Behavior
The column names should match the data in the columns / the column names from the import file

## Actual Behavior
The column names in the import validation display are all muddled up, even if data are then imported correctly.

## Steps to Reproduce
1. Go to admin/product_import 
2. Download the product import template file
3. Set up some products to import, following instructions in user guide - https://guide.openfoodnetwork.org/advanced-features/products/product-and-inventory-import#1-import-new-products
4. Save your file as csv
5. Go back to product import
6. Select 'Product List', choose file, select your file
7. Select import 
8. View product validation

## Context
Was extremely confusing when I was trying to upload products

## Severity
s3 - non-critical feature and you can still do the import. However if you are a person doing this for the first time and don't know this it would likely put you off using it. Also I am imagining that if you have an actually incorrect file this is going to give you very unhelpful feedback to fix it

-->

## Your Environment
Chrome, Mac OS

### LINKED ISSUE #3448: [Product import] Support Byte Order Mark (BOM) character inserted by Excel
Support Byte Order Mark (BOM) character inserted by Excel.

## Description

Saving the CSV file with Excel sometimes insert the Byte Order Mark (BOM) UTF-8 character. Files with BOM do not seem to be parsed properly.

@RachL encountered the problem [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/3233#issuecomment-461056579). I imagine other users would encounter the same issue.

It seems that it is just the first column that the parser cannot read properly when there is a BOM.

## Expected Behavior

Importing the spreadsheet should work.

## Actual Behavior

Importing the spreadsheet does not work.

## Severity

Matt-Yorkley:
#### What? Why?

Closes #3448
Closes #3414

Microsoft Excel was sometimes adding a BOM character at the start of CSV files, which was causing the entire first column to not be parsed and disappear. 

#### What should we test?

The CSV file originally tested with here should now work: https://github.com/openfoodfoundation/openfoodnetwork/pull/3233#issuecomment-461056579

#### Release notes

Fixed another encoding issue with CSV files made in Excel.

Changelog Category: Fixed

### DIFF
--- a/app/services/upload_sanitizer.rb
+++ b/app/services/upload_sanitizer.rb
@@ -1,0 +1,18 @@
+# Formats uploaded files to UTF-8 encoding and strips unexpected BOM characters.
+# Takes an open File object as input
+class UploadSanitizer
+  def initialize(upload)
+    @data = upload
+  end
+
+  def call
+    @data.force_encoding('UTF-8')
+    strip_bom_character
+  end
+
+  private
+
+  def strip_bom_character
+    @data.gsub("\xEF\xBB\xBF".force_encoding("UTF-8"), '')
+  end
+end
--- a/app/controllers/admin/product_import_controller.rb
+++ b/app/controllers/admin/product_import_controller.rb
@@ -84,7 +84,8 @@ module Admin
       directory = 'tmp/product_import'
       Dir.mkdir(directory) unless File.exist?(directory)
       File.open(Rails.root.join(directory, filename + extension), 'wb') do |f|
-        f.write(upload.read)
+        data = UploadSanitizer.new(upload.read).call
+        f.write(data)
         f.path
       end
     end
--- a/spec/services/upload_sanitizer_spec.rb
+++ b/spec/services/upload_sanitizer_spec.rb
@@ -1,0 +1,20 @@
+require 'spec_helper'
+
+describe UploadSanitizer do
+  describe "#call" do
+    let(:upload) do
+      File.open("/tmp/unsanitized.csv", 'wb:ascii-8bit') do |f|
+        f << "\xEF\xBB\xBF"
+        f << "Test"
+      end
+    end
+    let(:service) { UploadSanitizer.new(File.open(upload).read) }
+
+    it "sanitizes the uploaded file" do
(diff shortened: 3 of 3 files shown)


app/services/upload_sanitizer.rb
```diff
@@ -0,0 +1,19 @@
...
+  def call
+    @data.force_encoding('UTF-8')
+    strip_bom_character
+    @data
+  end
+
+  private
+
+  def strip_bom_character
+    @data.gsub!("\xEF\xBB\xBF".force_encoding("UTF-8"), '')
```
sauloperez: IMO it'd be better to use its non-bang version to avoid the mutation which will make the line 11 unnecessary. This would return the stripped string already.

Matt-Yorkley: Done @sauloperez

RachL:
Both files from #3233 (comment) are now working properly! Producer column is correctly found :)

app/services/upload_sanitizer.rb
```diff
@@ -0,0 +1,18 @@
+# Formats uploaded files to UTF-8 encoding and strips unexpected BOM characters.
+# Takes an open File object as input
+class UploadSanitizer
```
luisramos0: here we took option 1 for the name of this service as described here:
https://community.openfoodnetwork.org/t/how-to-name-a-new-service-in-ofn/1530

we should make a decision and stick to some naming convention, otherwise, the service naming chaos will keep increasing.
@Matt-Yorkley @sauloperez @kristinalim could you guys please vote on your favorite options in the discourse thread above? thanks!

PR #3508, opened 2019-02-16
[Spree Upgrade] Fix bulk_order_mgmt_spec - 2-0 build GREEN !!!

### LINKED ISSUE #2927: [Spree Upgrade] Fix Upgrade build
This epic contains issues related to fixing the upgrade build.
We started this epic with 201 broken specs, we are now at 119 broken specs.

Meanwhile we extract FrontOffice issues to #3123 and BackOffice issues to #3124.

Most issues under these epics will be issues with the name of the spec and the test error message.

luisramos0:
#### What? Why?
Fix bulk_order_management, the factory order_with_distributor does not come with ship_address, ship_address is required to make the order workflow work because in the delivery step the shipping_methods are revalidated against the ship_address of the order, if ship_address is not set, no shipping_method will be available.

Closes #2927

#### What should we test?
Makes the 2-0 build green!

### DIFF
--- a/spec/features/admin/bulk_order_management_spec.rb
+++ b/spec/features/admin/bulk_order_management_spec.rb
@@ -591,8 +591,9 @@ feature %q{
 
     context "using action buttons" do
       context "using edit buttons" do
-        let!(:o1) { create(:order_with_distributor, state: 'complete', completed_at: Time.zone.now ) }
-        let!(:o2) { create(:order_with_distributor, state: 'complete', completed_at: Time.zone.now ) }
+        let(:address) { create(:address) }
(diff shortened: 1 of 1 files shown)


luisramos0:
Yeah, it's important detail of the order workflow to remember: 
`before_transition :to => :delivery, :do => :create_proposed_shipments`

and then
```
  def create_proposed_shipments
      **shipments.destroy_all**

      packages = Spree::Stock::Coordinator.new(self).packages
      packages.each do |package|
        shipments << package.to_shipment

```
and here the shipping_methods will be validated with ship_method.include? that needs order.ship_address

sauloperez:
I knew it! I was trying to grep something like `shipments.destroy_all` because I remembered reading similar code. Great team work @luisramos0 ! thanks!

PR #3511, opened 2019-02-18
Update Bugsnag from 5.5.0 to 6.11.1

sauloperez:
#### What? Why?

This brings lots of fixes and a great dose of enhancements such as
a rewritten `Delayed::Job` integration. Turns out our version is from 09 Nov 2017. Check them out in https://github.com/bugsnag/bugsnag-ruby/blob/master/CHANGELOG.md.

This came up after investigating why Katuma is not getting any errors on Bugsnag for a few months already, although we do experience them. It's better to update the library before reaching out to customer support.

#### What should we test?

Check that from a rails console we can notify Bugsnag.

#### Release notes

Update Bugsnag Ruby notifier library from 5.5.0 to 6.11.1

Changelog Category: Changed


sauloperez:
I did manage to deploy it to staging and use `Bugnsnag` from the Rails console successfully so this is safe to merge.

sauloperez:
I don't understand what I might be missing in Katuma staging then 🤔 I'll try to take a look.

mkllnk:
I just looked at Katuma staging and it doesn't have `config/initializers/bugsnag.rb`. That file should contain your API key: https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/bugsnag.rb.j2

sauloperez:
Yep, fixed in https://github.com/openfoodfoundation/openfoodnetwork/pull/3554#issuecomment-467536956

mkllnk:
#3554 is a good re-factor for fixing this but not the fix itself. Once #3554 is merged you would need to change this pull request to remove the `use_ssl` line. That's what blows up in this PR.

mkllnk:
For testing: we just need to stage it and make sure that the home page is still loading. Nothing else.

sauloperez:
I'll do it myself because I also want to check notifications actually get to Bugsnag.

sauloperez:
Tested and working. Although I had to reprovision for the changes in https://github.com/openfoodfoundation/ofn-install/pull/329 to take effect

mkllnk:
Great! Thank you for taking this on, Pau. It will save us lots of time.

PR #3512, opened 2019-02-18
3466,3478 [Enterprise Fee Summary] Scope report to eligible: true and non-zero adjustments

### LINKED ISSUE #3466: Don't show 0 fee rows in Enterprise Fee Summary Report
**## Description**

**- As an enterprise user
**- On page: Enterprise Fee Summary Report https://staging.openfoodnetwork.org.au/admin/reports/enterprise_fee_summary
**- I want to be able to:
See results without 'noise'

At the moment there are quite a few rows that effectively contain no information e.g. they are zero. I am noticing this particularly in relation to when there are no fees attached to payment and shipping methods - there is still a row there for them even though no fees have been collected.

I think that where no fees are attached or collected there should be no row in the report

**## Acceptance Criteria**
There are no rows with value ZERO in the Enterprise Fee Summary Report

### LINKED ISSUE #3478: [Enterprise Fee Summary] Adjustments need to be scoped to "eligible: true"
Adjustments need to be scoped to `eligible: true`.

## Description

This issue also causes #3466.

It seems that adjustments can exist even when the adjustment amount is 0, for promotions associated with an order even when they are not the final one applied, and for computed payment transaction fees even if the payment failed or was invalid.

The `eligible` attribute of the adjustment tells whether the adjustment really was applied to the order or not.

The Enterprise Fee Summary report should summarize only adjustments that are `eligible: true`.

## Expected Behavior

The report should not include adjustments with 0 amount or adjustments that are not actually applied.

## Actual Behaviour

The report includes these adjustments.

## Steps to Reproduce

For each of these tests, it would be convenient to create a new user with a unique name, to more easily see results in the report.

1. Configure the hub to allow pick-up shipping.
2. As superadmin: Create a payment method with "Bogus" as provider. Set a fee for this.
3. As superadmin: Configure the hub to allow cheque payment and payment through Bogus in Step 2.

Not sure if you've used Bogus payment method before. A sample card number that works with this test provider is `4111111111111111`. Many other card numbers such as `1111111111111111` should cause failure.

To check $0 amount:

1. Checkout with pick-up for shipping and cheque for payment.
2. There should be no shipping and payment adjustments for the customer in the report, but there are entries for this.

To check no longer applicable adjustments:

1. Attempt an in checkout using Bogus payment method.
2. There would be no payment adjustment for the customer in the report yet.
3. Complete the checkout with a valid card number for Bogus again.
4. There would be an entry for the payment transaction under the customer, but this is double the correct adjustment.

## Context

[Discovered randomly](https://github.com/openfoodfoundation/openfoodnetwork/pull/3445#discussion_r255901384) while reviewing PR #3445 for Spree upgrade.

## Severity

## Possible Fix

Apply `eligible: true` to the scope for the report.

kristinalim:
Report should not include `eligible: false` and $0 adjustments.

#### What? Why?

Closes #3466 
Closes #3478 

At the moment, `eligible: false` adjustments (such as payment transactions for failed payments), and $0 adjustments (such as for pick-up orders) are still included in the report.

These should not appear.

#### What should we test?

**Preparations for Testing**

For each of these tests, it would be convenient to create a new user with a unique name, to more easily see results in the report.

As superadmin:

1. Configure the hub to allow pick-up shipping method, with no shipping fee.
2. Create a payment method with "Bogus" as provider. Set a non-zero fee for this.
3. Configure the hub to allow both cheque payment and payment through Bogus in Step 

A sample card number that works with this test provider is `4111111111111111`. A sample invalid card number is `1111111111111111`.

**Test $0 adjustment is not included**

1. Check out with pick-up for shipping and cheque for payment.
2. There should be no shipping adjustment for the customer in the report.

**Check failed/invalid adjustment is not included**

1. Attempt an invalid checkout using Bogus payment method.
2. There would be no payment adjustment for the customer in the report yet.
3. Complete the checkout with a valid card number for Bogus again.
4. There would be an entry for the payment transaction under the customer. The amount should be correct, not double.

#### Release notes

- Exclude failed and invalid adjustments from Enterprise Fee Summary
- Exclude $0 adjustments from Enterprise Fee Summary

Changelog Category: Fixed

### DIFF
--- a/engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
+++ b/engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
@@ -46,7 +46,7 @@ module OrderManagement
 
         def find_adjustments
           chain_to_scope do
-            Spree::Adjustment
+            Spree::Adjustment.eligible
           end
         end
 
--- a/engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
+++ b/engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
@@ -17,7 +17,7 @@ module OrderManagement
         end
 
         def result
-          group_data.select_attributes
+          group_data.exclude_groups_with_zero_total.select_attributes
           @scope.all
         end
 
@@ -334,6 +334,10 @@ module OrderManagement
             if params.shipping_method_ids.present?
           filter_scope(spree_payment_methods: { id: params.payment_method_ids }) \
             if params.payment_method_ids.present?
+        end
+
+        def exclude_groups_with_zero_total
+          filter_scope("spree_adjustments.amount != 0")
         end
 
         def group_data
--- a/engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb
+++ b/engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb
@@ -151,6 +151,36 @@ describe OrderManagement::Reports::EnterpriseFeeSummary::ReportService do
     end
   end
 
+  describe "data exclusions" do
+    describe "invalid adjustments (through 'eligible') like failed payments" do
+      let!(:customer_order) { prepare_order(customer: customer) }
+
+      before do
+        # Make the payment fail. See Spree::Payment#revoke_adjustment_eligibility.
+        payment = customer_order.payments.first
+        payment.state = "failed"
+        payment.save!
+      end
+
+      it "is included" do
+        totals = service.list
+
+        expect(totals.length).to eq(2)
+
+        expected_result = [
+          ["Payment Transaction", "Sample Distributor", "Sample Payment Method", "Sample Customer",
+           nil, nil, nil, "2.00"],
+          ["Shipment", "Sample Distributor", "Sample Shipping Method", "Sample Customer",
+           nil, nil, "Platform Rate", "1.00"]
+        ]
+
+        expected_result.each_with_index do |expected_attributes, row_index|
+          expect_total_attributes(totals[row_index], expected_attributes)
+        end
+      end
+    end
+  end
+
   describe "handling of more complex cases" do
     context "with non-sender fee for incoming exchange and non-receiver fee for outgoing" do
       let!(:variant) do
--- a/engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb
+++ b/engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb
@@ -214,6 +214,31 @@ describe OrderManagement::Reports::EnterpriseFeeSummary::ReportService do
         end
       end
     end
+
+    describe "$0 mandatory adjustments" do
+      let!(:payment_method) do
+        create(:payment_method, :per_item, amount: 0, name: "Sample Payment Method")
+      end
+
+      let!(:customer_order) { prepare_order(customer: customer) }
+
+      it "is included" do
+        totals = service.list
+
+        expect(totals.length).to eq(2)
+
+        expected_result = [
+          ["Payment Transaction", "Sample Distributor", "Sample Payment Method", "Sample Customer",
+           nil, nil, nil, "0.00"],
+          ["Shipment", "Sample Distributor", "Sample Shipping Method", "Sample Customer",
+           nil, nil, "Platform Rate", "1.00"]
+        ]
+
+        expected_result.each_with_index do |expected_attributes, row_index|
+          expect_total_attributes(totals[row_index], expected_attributes)
+        end
+      end
+    end
   end
 
   describe "handling of more complex cases" do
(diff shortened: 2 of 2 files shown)


engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
```diff
@@ -336,6 +336,10 @@ def filter_by_fee(params)
             if params.payment_method_ids.present?
         end
 
+        def exclude_groups_with_zero_total
+          filter_scope("spree_adjustments.amount != 0")
```
luisramos0: I dont think you need to replicate model logic here, adjustments.set_eligibility is already giving you this:
https://github.com/openfoodfoundation/spree/blob/step-6a/core/app/models/spree/adjustment.rb#L65

```suggestion
          filter_scope("spree_adjustments.eligible = true")
```

sauloperez: Good catch. Then this can be turned into `filter_scope(eligible: true)` :ok_hand:

kristinalim: Thanks, @luisramos0.

My finding about this is that $0 adjustments are not automatically set to `eligible: false` if `mandatory: true`, which is the case for all/most types of order adjustments. The only adjustment I hadn't checked was `TagRule` (I haven't explored how to use this yet). Failed payment transaction fee adjustments are also set to `eligible: false`.

Do let me know if it seems I'm misunderstanding.

In [`CalculatedAdjustments#create_adjustment`](https://github.com/openfoodfoundation/spree/blob/step-6a/core/lib/spree/core/calculated_adjustments.rb#L27):

```ruby
def create_adjustment(label, target, calculable, mandatory=false)
  # ...
end
```

And notice the 4th argument below:

```
$ git grep "create_adjustment("
app/models/product_distribution.rb:20:    a = enterprise_fee.create_adjustment(adjustment_label_for(line_item), line_item.order, line_item, true)
app/models/spree/payment_decorator.rb:24:        payment_method.create_adjustment(adjustment_label, order, self, true)
app/models/tag_rule/discount_order.rb:8:    create_adjustment(I18n.t("discount"), subject, subject)
lib/open_food_network/enterprise_fee_applicator.rb:4:      a = enterprise_fee.create_adjustment(line_item_adjustment_label, line_item.order, line_item, true)
lib/open_food_network/enterprise_fee_applicator.rb:12:      a = enterprise_fee.create_adjustment(order_adjustment_label, order, order, true)
spec/models/enterprise_fee_spec.rb:112:      pd1.enterprise_fee.create_adjustment('foo1', line_item.order, line_item, true)
spec/models/enterprise_fee_spec.rb:113:      pd2.enterprise_fee.create_adjustment('foo2', line_item.order, line_item, true)
spec/models/enterprise_fee_spec.rb:125:      tax_rate.create_adjustment('foo', line_item.order, line_item)
spec/models/enterprise_fee_spec.rb:146:      pd1.enterprise_fee.create_adjustment('foo1', line_item1.order, line_item1, true)
spec/models/enterprise_fee_spec.rb:147:      pd2.enterprise_fee.create_adjustment('foo2', line_item1.order, line_item1, true)
spec/models/enterprise_fee_spec.rb:148:      pd3.enterprise_fee.create_adjustment('foo3', line_item2.order, line_item2, true)
spec/models/enterprise_fee_spec.rb:149:      pd4.enterprise_fee.create_adjustment('foo4', line_item2.order, line_item2, true)
spec/models/product_distribution_spec.rb:77:        adjustment = pd.enterprise_fee.create_adjustment('foo', line_item.order, line_item, true)
spec/models/product_distribution_spec.rb:85:        pd.enterprise_fee.create_adjustment('one', line_item.order, line_item, true)
spec/models/product_distribution_spec.rb:86:        pd.enterprise_fee.create_adjustment('two', line_item.order, line_item, true)
spec/models/spree/order_spec.rb:145:      a = ef.create_adjustment("adjustment", o, o, true)
spec/models/spree/order_spec.rb:153:      a = ef.create_adjustment("adjustment", o, o, true)
spec/models/spree/order_spec.rb:163:      sm.create_adjustment("adjustment", o, o, true)
spec/models/spree/order_spec.rb:171:      ef.create_adjustment("adjustment", li.order, li, true)
spec/models/tag_rule/discount_order_spec.rb:78:        shipping_method.create_adjustment("Shipping", order, order, true)
```

kristinalim: Interestingly, Spree's default `mandatory` is `false`.

luisramos0: yeah, makes sense.
I am not sure we are using mandatory correctly in these cases...

anyway, we need to keep your original approach here as we don't want to review the mandatory situation just because of this...

RachL:
This is looking good to me :)

Ready to go! Testing notes: https://docs.google.com/document/d/12jtljG3ABkNR4j4qhrKbnBCVxKdB2FWZFq9NxwT4K_8/edit#

PR #3531, opened 2019-02-22
Document stock validation of variant overrides

mkllnk:
#### What? Why?

Follow up from https://github.com/openfoodfoundation/openfoodnetwork/pull/3505#pullrequestreview-205530320

Our new stock validation logic can be a bit confusing. I added a table for a better overview to understand the different states.

#### What should we test?

No testing required.

#### Release notes

Added: Added code documentation of stock validation rules

Changelog Category: Added

### DIFF
--- a/app/models/concerns/stock_settings_override_validation.rb
+++ b/app/models/concerns/stock_settings_override_validation.rb
@@ -1,3 +1,21 @@
+# Validates the combination of on_demand and count_on_hand values.
+#
+# `on_demand` can have three values: true, false or nil
+# `count_on_hand` can either be: nil or a number
+#
+# This means that a variant override can be in six different stock states
+# but only three of them are valid.
+#
+# | on_demand | count_on_hand | stock_overridden? | use_producer_stock_settings? | valid? |
+# |-----------|---------------|-------------------|------------------------------|--------|
+# | 1         | nil           | false             | false                        | true   |
+# | 0         | x             | true              | false                        | true   |
+# | nil       | nil           | false             | true                         | true   |
+# | 1         | x             | ?                 | ?                            | false  |
+# | 0         | nil           | ?                 | ?                            | false  |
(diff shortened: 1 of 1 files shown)


app/models/concerns/stock_settings_override_validation.rb
```diff
@@ -1,3 +1,29 @@
...
+#
+# - true
+# - false
+# - nil
+#
+# `count_on_hand` can either be:
+#
+#  - nil
+#  - a number
+#
```
luisramos0: My preference goes for using less lines in comments so I'd do:
```suggestion
# `on_demand` can have three values: true, false or nil
# `count_on_hand` can either be: nil or a number 
```

luisramos0:
@kristinalim I think with this comment I prefer not to change this on_demand.nil? to user_producer_stock_settings? but I can change it if you prefer. let me know and I make the change.

kristinalim:
Sounds good to me, @luisramos0. :slightly_smiling_face: Happy to approve #3505 once green.

app/models/concerns/stock_settings_override_validation.rb
mkllnk: Good idea. I committed your changes.

mkllnk:
@kristinalim It looks like @luisramos0 is not introducing the `use_producer_stock_settings?` method anymore. I still think that it aids the understanding to have it in the table because it is used in the user interface. In this context `use_producer_settings?` makes sense. Do you think we need to denote that it's not a method? Should we replace `_` with spaces?

kristinalim:
@mkllnk Actually, #3505 is already using `use_producer_stock_settings?`. The PR is already in Test Ready, so the method name in the comment here can be renamed?

mkllnk:
I'll wait for the other PR being merged.

mkllnk:
Okay, @kristinalim I updated the documentation now.

Issue #3140, opened 2018-11-26
[Spree upgrade] Move Spree translations into the Spree namespace


mkllnk:
This is complementary to https://github.com/openfoodfoundation/openfoodnetwork/issues/3119 and https://github.com/openfoodfoundation/openfoodnetwork/issues/2187.

Spree 2 moved all i18n translation keys from the root namespace into the Spree namespace. For example `I18n.t("back_to_payments_list")` becomes `Spree.t("back_to_payments_list")` which is equivalent to `I18n.t("spree.back_to_payments_list")`.

We will find a lot of these broken translations during testing, but we can't possibly find all. This issue is to collect all translations we come across and don't want to forget. Please edit this description or post below if you are aware of any more translations we need to deal with.

This should be part of phase 2 in the Spree upgrade.

```
back_to_payments_list
```

PR #3534, opened 2019-02-22
Move registration process translations to lazy lookups to better organize the en.yml file

### LINKED ISSUE #2484: registration process - move translation keys into new structure
## Description
To avoid having all translation keys at the root of the en.yml file. A new structure for translation keys was created for step TYPE of the registration process and lazy look up was used as much as possible.
The TYPE step was changed as part of issue #2303 where we confirmed that if translation keys are moved from root to a new place, if we keep the key text, transifex will pick up the translations and translators will not have to translate everything again.

The goal is to make the en.yml file organized with a minimum number of entries at the root level.

## Testing
No change in behaviour, no change in transifex. This is just a refactor of the registration process translation keys.
Afterwards, we should test the registration process and make sure all existing translations are picked up automatically by transifex.

luisramos0:
#### What? Why?

Closes #2484 

Moving registration translations to a namespace in en.yml.

#### What should we test?
I have verified locally that all entries are picked up by the UI.
No translation keys are changed, just moved, so transifex will pick up existing translations.

The translations of the registration process need to be verified.

#### Release notes
Changelog Category: Changed
Improved the organisation of the translations of the enterprise registration process.

Doing this before we get to #3140


RachL:
I've registered in Spanish, Catalan and Portuguese, but I've just realized that this is touching the English translation. @luisramos0 if I stage this on UK (even if UK is to multilingual), will this work?

luisramos0:
yes, this can only be validated in English for now.
after we merge and transifex is merged as well, we can retest in a different language.

RachL:
thank you @luisramos0 ! Registration is well translated. Here are the screenshots (confirmation email and welcome email as well):

image

image

PR #3535, opened 2019-02-22
Move to master all en.yml entries added so far to v2 and add more missing v2 translations

### LINKED ISSUE #3140: [Spree upgrade] Move Spree translations into the Spree namespace

### LINKED ISSUE #3452: [i18n] missing translations orders, bulk orders and subscriptions
Some translations missing in current BE instance (and thus some others too I guess?)

orders:
dashboard-orders-missing-translations

bulk orders:
dashboard-bulkorders-missing-translations

create new subscription step 3:
dashboard-new-subscription-step3-missing-translations

luisramos0:
#### What? Why?

Putting all v2 missing translations in master en.yml is harmless to master and will enable us to start translating things in transifex using the normal process. All these translations can then be merged back to 2-0-stable. One commit here is just moving things from 2-0-stable to master, the other one is adding new ones.

In the end, en.yml file in the 2-0-stable branch should be the same as master only with entry removals (there's only one entry deleted so far).

Closes #3140

#### What should we test?
v2 is now translated, we need to find if there are any other missing entries.
I have verified each of these entries locally (very boring task!!!)
It's important we use a language that is not translated in spree base translations so we test #3452 is fixed in v2. I have used fr_BE for tests.

#### Release notes
Changelog Category: Added
Added translation entries that will be used for OFN v2, so that translators can start the translation work.

It's part of the spree upgrade.

#### Dependencies
This fixes #3452 but only in v2.


luisramos0:
sorry to add stuff after your review @sauloperez  I just added one more commit with more things needed for v2 translations, it's just more en.yml entries.

mkllnk:
Should we do a sanity check for the merge to master? Not that we accidentally override another translation... Having a browse through the admin interface and trying to spot weird text would be good.

RachL:
@mkllnk @luisramos0 can you confirm that staging UK is enough? I don't need multilingual to test this, but I need a staging with English as a language option?

luisramos0:
yes, for now this can only be validated in English

RachL:
thank you @luisramos0 I navigated through (I think) all backoffice pages in both producer and super admin. I think we are good :) Moving this to ready to go

sauloperez:
Would you mind fixing the conflict first @luisramos0 ?

luisramos0:
thanks! done.

PR #3536, opened 2019-02-22
Wait for angular to load in BOM spec

### LINKED ISSUE #3383: [Flaky Spec] spec/features/admin/bulk_order_management_spec.rb:56

Matt-Yorkley:
#### What? Why?

Closes #3383

Hopefully this will finally fix this flaky spec!

#### The issue:

There are a number of tests that fail occasionally in this file. Angular is making some asynchronous requests on the page, then updating the DOM when it's finished. The issue seems to be that Capybara is not waiting long enough, or is not waiting until Angular has finished loading. It's consistently green in people's dev environments, but fails sometimes on Semaphore, so I think the problem is partly due to the CI environment being slower?

#### My solution:

I've added a couple of helper methods for feature specs that enable us to interrogate angular directly from within a test, and find out if the `$http` module is still processing requests, or if the `RequestMonitor` class is still loading data. They should be re-usable in the future if we have similar issues.

#### What should we test?

Tests in `/spec/features/admin/bulk_order_management_spec.rb` should not be flaky.

### DIFF
--- a/spec/support/request/web_helper.rb
+++ b/spec/support/request/web_helper.rb
@@ -152,15 +152,17 @@ module WebHelper
     page.driver.browser.switch_to.alert.accept
   end
 
-  def wait_for_angular_requests
-    wait_until { angular_requests_finished }
+  def angular_http_requests_finished(angular_controller=nil)
+    scope_element = angular_controller ? "[ng-controller=#{angular_controller}]" : '.ng-scope'
+    page.evaluate_script("angular.element(document.querySelector('#{scope_element}')).injector().get('$http').pendingRequests.length == 0")
+  end
+
+  def request_monitor_finished(angular_controller=nil)
+    scope_element = angular_controller ? "[ng-controller=#{angular_controller}]" : '.ng-scope'
+    page.evaluate_script("angular.element(document.querySelector('#{scope_element}')).scope().RequestMonitor.loading == false")
   end
 
   private
-
-  def angular_requests_finished
-    page.evaluate_script('angular.element(".ng-scope").injector().get("$http").pendingRequests.length === 0')
-  end
 
   def wait_for_ajax
     wait_until { page.evaluate_script("$.active") == 0 }
--- a/spec/support/request/web_helper.rb
+++ b/spec/support/request/web_helper.rb
@@ -152,17 +152,22 @@ module WebHelper
     page.driver.browser.switch_to.alert.accept
   end
 
-  def angular_http_requests_finished(angular_controller=nil)
-    scope_element = angular_controller ? "[ng-controller=#{angular_controller}]" : '.ng-scope'
-    page.evaluate_script("angular.element(document.querySelector('#{scope_element}')).injector().get('$http').pendingRequests.length == 0")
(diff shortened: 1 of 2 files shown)


Matt-Yorkley:
I've tried various things with this flaky spec. I thought I had it that last time, but the test is still failing... 

:sob:

Matt-Yorkley:
Ok, I've had another go and I think I've nailed it this time.

spec/support/request/web_helper.rb
```diff
@@ -152,6 +152,16 @@ def accept_js_alert
...
+  def angular_http_requests_finished(angular_controller=nil)
+    scope_element = angular_controller ? "[ng-controller=#{angular_controller}]" : '.ng-scope'
+    page.evaluate_script("angular.element(document.querySelector('#{scope_element}')).injector().get('$http').pendingRequests.length == 0")
+  end
+
+  def request_monitor_finished(angular_controller=nil)
+    scope_element = angular_controller ? "[ng-controller=#{angular_controller}]" : '.ng-scope'
+    page.evaluate_script("angular.element(document.querySelector('#{scope_element}')).scope().RequestMonitor.loading == false")
+  end
+
```
luisramos0: I think this will make it more readable:
```suggestion
.
def angular_http_requests_finished(angular_controller=nil)
    scope_expr = scope_element_expression(angular_controller)
    pending_http_requests_script = "#{scope_expr}.injector().get('$http').pendingRequests.length == 0"
    page.evaluate_script(pending_http_requests_script)
end

 def request_monitor_finished(angular_controller=nil)
    scope_expr = scope_element_expression(angular_controller)
    request_monitor_script = "#{scope_expr}.scope().RequestMonitor.loading == false")"
    page.evaluate_script(request_monitor_script)
  end
  
  def scope_element_expression(angular_controller=nil)
      scope_element = angular_controller ? "[ng-controller=#{angular_controller}]" : '.ng-scope'
      scope_element_expression = "angular.element(document.querySelector('#{scope_element}')"
  end
```

Matt-Yorkley: I've tidied it up and added a comment to explain how the format for the angular controller name  and how it finds the scope if `nil` :+1:

sauloperez:
:clap:

sauloperez:
It ran three times in a row so it seems stable enough, for now :trollface:

sigmundpetersen:
... and it uncovered the next flaky BOM spec https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/master/builds/508 🙀

PR #3537, opened 2019-02-23
Fix nil values in on_hand column

### LINKED ISSUE #3419: [Product import] Saving error when on_demand is empty
## Description
Seen while testing #3387 : when testing what happened with empty on_demand cells for 2 lines on a second batch, I had an error at the saving stage, which prevented the 2 products to be imported.

## Expected Behavior
Any errors should be revealed before the actual import, not after.

## Steps to Reproduce
1. Use the file described here with producers matching your environment: https://docs.google.com/document/d/1muvvMKkA0dfUxNS75VxavsFudFXorVcojbLCrZW6Msc/edit# (Apparently github prevent me from uploading a .csv to the issue, too bad)
2. Click on import
3. Observe results

## Your Environment

* Version used: 1.25

### LINKED ISSUE #3523: [Product import] On_hand can currently not be left blank
## Description
Since changes made in inventory, these new rules have been introduce to PI:
>    1. If blank - Read as "Use producer stock settings", so "on_hand" should be blank.
    2. If "true", "TRUE", or "1" - Read as on demand of "Yes", so "on_hand" should be blank.
    3. If "false", "FALSE", or "0" - Read as on demand of "No", so "on_hand" is required.

However on the product side, a product cannot be created or updated with a blank on_hand column. If you set on_demand to 1, you need to set on_hand to 0.

## Steps to Reproduce
1. Log in as a super admin
2. Go to /admin/product_import
3. Try to upload a file while leaving on_demand to 1 and on_hand to 0 (you should see the below error message)

## Animated Gif/Screenshot
image

## Severity

## Your Environment
* Version used: v1.26.1

## Possible Fix
Allow on_hand to be blank

Matt-Yorkley:
#### What? Why?

Closes #3523

`on_hand` value should be set to zero for new products when the column is left blank.

#### What should we test?

Product Import: `on_hand` should be set to zero on new products when the column is left blank.

#### Release notes

Changelog Category: Fixed

### DIFF
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -254,6 +254,7 @@ module ProductImport
       new_product = Spree::Product.new
       new_product.assign_attributes(entry.attributes.except('id'))
       new_product.supplier_id = entry.producer_id
+      entry.on_hand = 0 if entry.on_hand.nil?
 
       if new_product.valid?
         entry.validates_as = 'new_product' unless entry.errors?
--- a/spec/models/product_importer_spec.rb
+++ b/spec/models/product_importer_spec.rb
@@ -50,7 +50,7 @@ describe ProductImport::ProductImporter do
         csv << ["Potatoes", "User Enterprise", "Vegetables", "6", "6.50", "2", "kg", "", ""]
         csv << ["Pea Soup", "User Enterprise", "Vegetables", "8", "5.50", "750", "ml", "", "0"]
         csv << ["Salad", "User Enterprise", "Vegetables", "7", "4.50", "1", "", "bags", ""]
-        csv << ["Hot Cross Buns", "User Enterprise", "Cake", "7", "3.50", "1", "", "buns", "1"]
+        csv << ["Hot Cross Buns", "User Enterprise", "Cake", nil, "3.50", "1", "", "buns", "1"]
       end
       File.write('/tmp/test-m.csv', csv_data)
       file = File.new('/tmp/test-m.csv')
@@ -122,7 +122,8 @@ describe ProductImport::ProductImporter do
 
       buns = Spree::Product.find_by_name('Hot Cross Buns')
       expect(buns.supplier).to eq enterprise
-      # buns.on_hand).to eq Infinity
+      expect(buns.on_hand).to eq Float::INFINITY
+      expect(buns.count_on_hand).to eq 0
       expect(buns.price).to eq 3.50
       expect(buns.unit_value).to eq 1
       expect(buns.variant_unit).to eq 'items'


RachL:
@Matt-Yorkley If I follow those rules for on_demand:

      1. If blank - Read as "Use producer stock settings", so "on_hand" should be blank.
       2. If "true", "TRUE", or "1" - Read as on demand of "Yes", so "on_hand" should be blank.
       3. If "false", "FALSE", or "0" - Read as on demand of "No", so "on_hand" is required.

With rule 3, I think that when on_demand is set to 0 I should see an error when importing a blank on_hand. The error should require me to set on_hand to something. Was this too complex ? If yes I think we can left it like this and mention it in user guide: a blank on_hand will result to a on_hand equal to 0 by default.

I noticed as well that this 0 by default rule now applies also for on_demand : I've uploaded a file with an empty on_demand and an empty on_hand. In this case, after import on_hand was left blank and on_demand was set to 0.
=> Shouldn't one of them be mandatory? I'm afraid that with large file this would go unnoticed by the user. Rule number 1 should be only true for inventory, not product import.

And finally with rule 2: If I set on_demand to 1 and on_hand to 50, the product is imported with on demand set to true. 🤔 I think that in this case we should see an error as well... what do you think ?

Testing notes and files: https://docs.google.com/document/d/1E95MmfFnANvbUBpKKDNl9M1BcCTcIKwFGrfFziq4oPg/edit#

Matt-Yorkley:
Ah yes! I think it didn't get fully adapted to the new logic of three on_demand states (true, false, and nil) in place of the old boolean state. I'll take another look. :+1:

daniellemoorhead:
And here's the PR for both those issues I commented on above. 

Still over 3 weeks - what's the plan with this PR? @Matt-Yorkley when will you be picking it up again to be finished? Doesn't seem to be in development as it's current position on the delivery pipe suggests...

RachL:
@daniellemoorhead Matt & I had a chat Tuesday March 12th where he told me he was getting it done during the week. That's the last news I have. Let's keep it here for a couple of days and I will put it back in the PI backlog if Matt has no more time left.

daniellemoorhead:
OK, it's been 5 days since you commented @RachL, and no word from @Matt-Yorkley. Sadly it doesn't look like it's going to be finished anytime soon so it's time to move it back 😞

Matt-Yorkley:
Ok, I'm back and will look at this now. We now need two sets of rules for how to handle the `on_demand` column after the changes to `on_demand` with inventory. @RachL do you have any preference on the UX side of things? 

For example, we could have:

**Importing Products:**
- blank or 0 -> not `on_demand`
- 1 -> `on_demand`

**Importing Inventory:**
- blank -> "Use producer stock settings"
- 0 -> not `on_demand`
- 1 -> `on_demand`

RachL:
@Matt-Yorkley awesome :)

Yes the rule is correct, but I have a few questions:

**Importing Products**:

 -   blank or 0 -> not on_demand

=> if the user does not fill on_hand here, does this mean that in this case we set on_hand to 0?
What happens if both on_hand and on_demand are empty? The proper case would be to display an error and not import. What do you think?

 -  1 -> on_demand

If both on_hand and on_demand are filed in, I think we should see an error and block import until the user cleans its file. Otherwise the other solution would be to say that on_demand always wins, but I'm afraid this is something we will forget. Any thoughts?

Matt-Yorkley:
If the user doesn't fill in the `on_demand` column with a `1`, it's not `on_demand`. If they also don't fill in the `on_hand` column, the product has no stock (and is not `on_demand`).

That sounds ok to me.

If both `on_hand` and `on_demand` are filled in,yes I think I would say `on_demand` wins. The `on_hand` value is irrelevant if a product is `on_demand`. I think you can do the same thing when adding a new product, no?

RachL:
@Matt-Yorkley you are right. I didn't know this. Let's do the same rule as when you add a product manually, it is indeed consistent!

Matt-Yorkley:
I think I'm going to handle the `on_demand` stuff in a separate PR and leave this one as it is.

Matt-Yorkley:
In which case, it's been approved twice and is ready for testing...

RachL:
@Matt-Yorkley did you submit a PR for the on_demand column?

Matt-Yorkley:
@RachL I tested it and mentioned the result here: https://github.com/openfoodfoundation/openfoodnetwork/issues/3419

PR #3538, opened 2019-02-23
Adjust expectations in flaky spec

### LINKED ISSUE #3473: [Flaky Spec] ./spec/features/admin/shipping_methods_spec.rb:82
## Description

This spec is unreliable:
./spec/features/admin/shipping_methods_spec.rb:82

https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/master/builds/473

## Context

Our Semaphore builds fail regularly because of these unreliable tests. If you encounter this in your build, please comment on this issue or add 👍

Error:
```

  1) shipping methods as an enterprise user creating a shipping method
     Failure/Error: expect(page).to have_flash_message message
       expected to find flash message #Shipping method "Teleport" has been successfully created!
     # ./spec/features/admin/shipping_methods_spec.rb:108:in `block (3 levels) in <top (required)>'

```

Matt-Yorkley:
#### What? Why?

Closes #3473

Tweaking a flaky spec to try to get it a bit more stable.

#### What should we test?

`/spec/features/admin/shipping_methods_spec.rb` should be green. Line 82 / 108.

### DIFF
--- a/spec/features/admin/shipping_methods_spec.rb
+++ b/spec/features/admin/shipping_methods_spec.rb
@@ -103,9 +103,8 @@ feature 'shipping methods' do
 
       click_button I18n.t("actions.create")
 
-      expect(page).to have_no_button I18n.t("actions.create")
-      message = "Shipping method \"Teleport\" has been successfully created!"
-      expect(page).to have_flash_message message
+      expect(page).to have_content I18n.t('editing_shipping_method')
+      expect(flash_message).to eq I18n.t('successfully_created', resource: 'Shipping method "Teleport"')
 
       expect(first('tags-input .tag-list ti-tag-item')).to have_content "local"
 


spec/features/admin/shipping_methods_spec.rb
```diff
@@ -103,9 +103,8 @@
 
       click_button I18n.t("actions.create")
 
-      expect(page).to have_no_button I18n.t("actions.create")
-      message = "Shipping method \"Teleport\" has been successfully created!"
-      expect(page).to have_flash_message message
+      expect(page).to have_content "Editing Shipping Method"
```
Matt-Yorkley: At this point in the test the page loads from the "new" shipping method page to "edit", showing the new shipping method that has been created. The sticking point seems to be that page load transition, which is a common problem with feature specs. I've changed the expectation here to check for the presence of the title of the next page instead of the absence of an element from the previous page.

spec/features/admin/shipping_methods_spec.rb
```diff
@@ -103,9 +103,8 @@
...
+      expect(flash_message).to eq 'Shipping method "Teleport" has been successfully created!'
```
Matt-Yorkley: I also changed this expectation to be in line with the syntax used throughout our other feature specs.

spec/features/admin/shipping_methods_spec.rb
sauloperez: Sounds good

spec/features/admin/shipping_methods_spec.rb
sauloperez: Let's not forget to use I18n in assertions as agreed on https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code-Conventions#always-use-i18nt-in-spec-assertions.

luisramos0:
the build is green, so ready to go.

PR #3539, opened 2019-02-24
Don't process proxy_order if the order has been cancelled

### LINKED ISSUE #3388: [Subscriptions] If I cancel a subs order, I still receive the email saying the order was made at the end of OC
## Description
As a customer I have a recurring order with hub A.
On week X, I receive the email when OC opens saying "an order has been made, click here to modify".
I click and cancel the order.
I receive an email saying the order has been cancelled.
But when the OC closes, I still receive an email saying the order has been made ! It is pretty confusing for the customer...

## Expected Behavior
If I cancel an order, in the context of a subscription, I shouldn't receive the mail triggered at the end of the OC saying the order has been made. After I receive the cancellation confirmation, no other email should be sent regarding that order.

## Actual Behavior
I do.

## Steps to Reproduce
1. Make sure the hub authorize modification of orders till the end of OC.
2. As hub : create a subscription for customer X, make sure you have an OC starting some minutes later with the relevant schedules and products in it.
3. In another browser, login the mailbox of the customer.
4. At the time the OC opens, check customer mailbox, you see the email saying an order has been made, and link to modify. Click on the link.
5. On the order view, click on "cancel the order". You see the confirmation.
6. Check customer mailbox: order cancellation received.
7. As a hub manager in the other browser  you can check : the order has been cancelled properly.
8. Wait for OC to close and check customer mailbox: you see a new email saying order has been made.

## Animated Gif/Screenshot
1- First email when OC opens
capture d ecran de 2019-01-24 18-21-21
2- Cancellation confirmed
capture d ecran de 2019-01-24 18-21-09
capture du 2019-01-24 18-21-49
3- Email confirmation when OC closes that order has been made :-( 
capture d ecran de 2019-01-24 18-35-34

## Context
Reported by French active user of the feature and beta tester, reported by one of their customer who faced the cas.

## Severity
As the feature is not officially released I don't put a severity, would be S2 else as there is no workaround to prevent that email to be sent and it's very confusing for a customer.

Matt-Yorkley:
#### What? Why?

Closes #3388 

Subscriptions create proxy orders which generate orders, which are processed when an order cycle closes. If the user or an admin cancels the order from the orders page before the OC closes, the proxy order still gets processed at the end of the OC (and confirmation emails are sent).

I've updated the `SubscriptionConfirmJob` to exclude proxy orders where the related order has been cancelled.

#### What should we test?

Follow the steps detailed in #3388. An order confirmation should not be sent at the end of the order cycle.

#### Release notes

Subscriptions: orders created by a subscription that have been manually cancelled before an OC closes will not be incorrectly processed.

Changelog Category: Fixed

### DIFF
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -25,7 +25,7 @@ class SubscriptionConfirmJob
   def proxy_orders
     ProxyOrder.not_canceled.where('confirmed_at IS NULL AND placed_at IS NOT NULL')
       .joins(:order_cycle).merge(recently_closed_order_cycles)
-      .joins(:order).merge(Spree::Order.complete)
+      .joins(:order).merge(Spree::Order.complete.not_state('canceled'))
   end
 
   def recently_closed_order_cycles
--- a/spec/jobs/subscription_confirm_job_spec.rb
+++ b/spec/jobs/subscription_confirm_job_spec.rb
@@ -58,6 +58,12 @@ describe SubscriptionConfirmJob do
       proxy_order.update_attributes!(confirmed_at: 1.second.ago)
       expect(proxy_orders).to_not include proxy_order
     end
+
+    it "ignores orders that have been cancelled" do
+      setup_email
+      proxy_order.order.cancel!
+      expect(proxy_orders).to_not include proxy_order
+    end
   end
 
   describe "performing the job" do


Matt-Yorkley:
@kristinalim This is my first PR delving into the guts of subscriptions. Does this look ok?

Matt-Yorkley:
@kristinalim would that affect subsequent orders in the subscription? 

I also noticed that if a `proxy_order` is cancelled from the subs UI it doesn't cancel `proxy_order.order`.

I've gone back to test it manually and now it's not creating any proxy orders in my dev environment so it's hard to retest... not sure if that's a bug? :frowning_face:

sauloperez:
I agree with @Matt-Yorkley setting the `canceled_at` to `proxy_orders` reads like the subscription is canceled rather than a particular order cycle's order, don't you think @kristinalim ?

I'm also afraid that canceling the order may cause confusion but I agree on trying it first. Then, I'd keep an eye on it and perhaps add just a dialog like "This order is part of a subscription. Do you still want to cancel this occurrence? This won't cancel any upcoming orders"

Matt-Yorkley:
I feel like subs needs a lot more work done on the spec side of things before tackling a lot of the changes in the backlog column...

sauloperez:
Let's focus our efforts there rather than wasting them on half-broken implementations

luisramos0:
there are some docs here: https://community.openfoodnetwork.org/t/subscriptions-technical-overview/1385
see ProxyOrder - canceled_at
maybe it will help

mkllnk:
Nice analysis, @kristinalim. I agree that it is inconsistent that `ProxyOrder.cancelled_at` depends on which UI element you use to cancel an order. It would probably be simpler if the ProxyOrder is not cancelled after it has been processed and the associated order is cancelled. It also means that Spree::Order doesn't have to trigger ProxyOrder. That can be tackled in a different issue though. @kristinalim Can we save some code and simplify it by going with your option 1? Otherwise I wouldn't bother unless it's causing other issues.

kristinalim:
Moving this to Test Ready, because this already has two approvals. The change needed for redefining `proxy_orders.canceled_at` to Option 1 (new issue #3566) has no conflicts with this PR.

@sauloperez and @luisramos0, in case you have a pending concern for this PR, feel free to move it back to Code Review. :slightly_smiling_face:

sauloperez:
Go ahead. I read your in-depth review but although I guess how it works I lack context (and focus to fully understand). I trust your judgement guys :+1:

RachL:
This is ready to go! https://docs.google.com/document/d/1w3ZrE6rB9fMbfWfdUffCJO6fGPDuw9WnzpfDXoz2pEo/edit#

Issue #3546, opened 2019-02-25
De-deface orders/adjustments page


luisramos0:
### What we should change and why (this is tech debt)
We should de-deface the adjustments page when editing an order in the backoffice.
There are mainly two changes we are doing with the overrides:
- we are adding tax to the adjustments
- we are not including adjustments state (open/close)

### Context
This was discovered in the spree upgrade, see #1830 
This is part of the overall de-deface effort #2807 

### Impact and timeline
Removing deface is important to make OFN views easier.

Issue #1830, opened 2017-09-14
[Spree Upgrade][Admin/Orders/Edit] Fix edit orders shipping adjustment


oeoeaio:
Related to Spree upgrade. `state` field has been added to adjustments, and does not seem to be playing perfectly with OFN yet.

Steps to reproduce.
- open up an order with shipping and/or payment fees in the backend
- click on 'Adjustments' and delete the adjustments for the shipping/payment fees
- click on 'Order Details', update the quantity of a line item and then 'Recalculate Fees'

Site will crash due to the following error:
````
NoMethodError spree/admin/orders#update
undefined method `state' for nil:NilClass
````

Stacktrace :
app/models/spree/order_decorator.rb:351
````ruby
def update_adjustment!(adjustment)
> state = adjustment.state
  adjustment.state = 'open'
  adjustment.update!(self)
  adjustment.state = state
````

sauloperez:
I never worked on this one, so unassigning myself. You might have a better picture after your work on adjustments @luisramos0 . Any feedback about that `state` field?

luisramos0:
I dont think this issue is up to date.

1. I am not sure this is being applied:
app/overrides/spree/admin/orders/_form/relabel_update_button.html.haml.deface

2. the code now looks like this:
 def update_adjustment!(adjustment)
    return if adjustment.finalized?
    state = adjustment.state
    adjustment.state = 'open'
    adjustment.update!
    adjustment.state = state
  end

We can get back to this when these are closed: #2935, #2938, #2940, #2941

luisramos0:
I did some investigation around this issue. I think we can title this issue as "fix order adjustments edit page", the example Rob gives above is one of them and is currently working for payment fees but not shipping fess. Here are my conclusions and questions:

##  Edit Adjustments
The first test would be to use the order adjustments pages to edit the values of the adjustments:

#### Edit Shipping Fee
Edit shipping adjustment is not possible. The adjustment amount is correctly updated on the DB but the an order.update is triggered and that updates all the adjustmentss and, because the shipping adjustment is OPEN, the value is reset to the original shipping method cost.
In v1 this recalculation only happens when in the order details page, when the button "update and recalculate fees" is pressed, if the shipping fee was edited by the user, the shipping feee is reset to the base shipping method cost. I think that's how it should remain in v2. I am not sure how to fix this... we probably need to set the shipping adjustment to closed after the controller updates it.

#### Edit Payment Fee
[...]

Issue #3440, opened 2019-02-05
Improve Variants autocomplete solution


luisramos0:
## Description
As part of the upgrade to v2 we put in a quick adaptation to the new spree v2 code and kept using Spree::Admin::VariantsController to search for variants to be added to an order, see #3194.

This issue represents the task to move the variants look up to the respective api endpoint Spree::Api::Variants.
This will enable us to delete the search.rabl template and also to clear up the challenge of exposing the variants stock_location to the api client: in the v2 spree API variants endpoint, the variants sent to the client include the stock items. Because in ofn v2 we only have one stock_item per variant (see VariantStock) maybe we will want to simplify the api::variants payload by not including the stock_items.
The challenge is that all the JS code that we use from spree around the autocomplete in admin orders page needs the stock_location info to work properly, so we have two options:
- adapt the spree JS code and the admin orders page (maybe too much work for nothing)
- as part of the rework of the api::variants endpoint, keep sending the stock_location info to the client as part of the variant as we are doing now in search.rabl.

## Expected Behavior
The behaviour of the admin orders page should be the same but the variants should be fetched from the api::variants endpoint.

PR #3548, opened 2019-02-25
[Spree Upgrade] Remove unused spree v2 features

luisramos0:
#### What? Why?

This PR is removing 3 things that come with spree 2 but we don't want to keep in OFN v2:
1 - stock locations management and stock transfers
2 - stock location split in orders
3 - adjustments states management

#### 1 - stock locations management and stock transfers
This PR de-defaces the main configuration menu:
 - exclude new spree 2 options stock locations and stock transfers: we only have on stock location in OFN yet so these don't make sense.
 - fix broken menu selection in v1 for OFN specific entries (if you go to https://staging.katuma.org/admin/stripe_connect_settings/edit you see Stripe Connect menu is not selected)

#### 2 - stock location split in orders
The split button is removed from the orders edit screen.

#### 3 - adjustments states management
It removes the ability to move adjustments state from open to close in the order adjustments page (added override to remove a line of the table that contains two buttons "open all adjustments" and "close all adjustments"). This is according to decision here https://github.com/openfoodfoundation/openfoodnetwork/issues/1830
I've created this tech debt issue to de-deface this view: https://github.com/openfoodfoundation/openfoodnetwork/issues/3546

#### What should we test?
1 - verify the main menu, all options work and the list is correct: it includes all the elements that are already there in v1 with the following differences: no stock locations and no stock transfers. Additionally, if an entry like Content is selected, the item should be highlighted.

2 - Go to the edit order page and verify that the split button is not present.

3 - Go to the edit order adjustments page and verify that the list of adjustments does not have buttons to change the state of the adjustments.

#### Dependencies
This PR conflicts with #3321 because of the config menu entries, the second one to be merged will have to rebased.


app/views/spree/admin/shared/_configuration_menu.html.haml
```diff
@@ -0,0 +1,28 @@
...
+      = configurations_sidebar_menu_item Spree.t(:taxonomies), admin_taxonomies_path
+      = configurations_sidebar_menu_item Spree.t(:shipping_methods), admin_shipping_methods_path
+      = configurations_sidebar_menu_item Spree.t(:shipping_categories), admin_shipping_categories_path
+      = configurations_sidebar_menu_item t(:enterprise_fees), main_app.admin_enterprise_fees_path
+      = configurations_sidebar_menu_item Spree.t(:analytics_trackers), admin_trackers_path
+      = configurations_sidebar_menu_item t('admin.cache_settings.show.title'), main_app.admin_cache_settings_path
+      = configurations_sidebar_menu_item t('admin.contents.edit.title'), main_app.edit_admin_content_path
+      = configurations_sidebar_menu_item t('admin.invoice_settings.edit.title'), main_app.edit_admin_invoice_settings_path
+      = configurations_sidebar_menu_item t('admin.matomo_settings.edit.title'), main_app.edit_admin_matomo_settings_path
+      = configurations_sidebar_menu_item t('admin.stripe_connect_settings.edit.title'), main_app.edit_admin_stripe_connect_settings_path
```
luisramos0: I didnt want to touch translations in this PR. We should probably remove all Spree.t entries (there are more) from ofn's code and move everything we can to our namespaces.

app/views/spree/admin/shared/_configuration_menu.html.haml
```diff
@@ -0,0 +1,30 @@
+- content_for :sidebar_title do
+  = Spree.t(:configurations)
```
luisramos0: I didnt want to touch translations in this PR. We should probably remove all Spree.t entries (there are more) from ofn's code and move everything we can to our namespaces.

app/assets/javascripts/admin/orders/controllers/order_controller.js.coffee
```diff
@@ -24,3 +24,7 @@ angular.module("admin.orders").controller "orderCtrl", ($scope, shops, orderCycl
 
   for shop in $scope.shops
     shop.disabled = !$scope.distributorHasOrderCycles(shop)
+
+  # Hides the split button introduced by spree in the order form
+  #   We only have one stock location in OFN so it's meaningless to split the order between stock locations
+  $('.split-item').hide()
```
luisramos0: I started by using angular.element here and it worked on the app and rpsec but karma kept failing. This direct use of jquery works for the app, rpsec and karma.

Matt-Yorkley: That might not be the best way to hide it. How about CSS display: none?

Matt-Yorkley: Or de-deface it and just remove the element from the view?

luisramos0: de-deface is a long way, this split code is mixed with js code that is working for us, the variant auto complete code and add variant to order with shipment selection. there's this issue that would/could include that: https://github.com/openfoodfoundation/openfoodnetwork/issues/3440

luisramos0: re the css, do you mean you think it's better to do .css( "display", "none" )?
it's "roughly" the same: http://api.jquery.com/hide/

luisramos0: oh, I think you mean that I should just add a css rule for this, no JS. I'll try that!

luisramos0: I have to use some long selector to override sprees css "display". I would keep this css version. what do you think?

luisramos0:
ok, agreed with @RachL on slack, I am reverting staging ES to a v1 staging server.

this PR needs a v2 staging server, staging AUS will be a v2 staging server in a few hours.

luisramos0:
ok, I resolved conflicts and tested in staging.
I found a very minor issue where the Content menu entry was not being highlighted. I added an extra commit: 77655bfffbdf99c11c564010b222c5ab6b83f1bd
I moving this back to code review  ☝️  @sauloperez @mkllnk

spec/features/admin/orders_spec.rb
```diff
@@ -285,6 +285,11 @@ def new_order_with_distribution(distributor, order_cycle)
         end
       end
 
+      scenario "cannot split the order in different stock locations" do
+        # There's only 1 stock location in OFN, so the split functionality that comes with spree should be hidden
+        expect(page).to_not have_selector '.split-item'
+      end
```
sauloperez: Just for next time, I think a view spec it's much better suited for this. We're just checking the views output and not emulating a user's behavior at all.

luisramos0: ok, that's interesting. It's just my ignorance, I've seen a view spec!
Would that be under a new folder /spec/views, something like [this](http://ruby-journal.com/how-to-write-rails-view-test-with-rspec/)?

mkllnk: Exactly. I read about view specs, maybe I wrote one once but neither OFN nor Fair Food use them. You could argue that views should be so trivial that they don't require specs. And maybe there is another unit that is better tested for this? That said, I don't object view specs, they may be useful in some cases and way cheaper than feature specs.

luisramos0: yes, I agree because views should be simple simple, but this is a very good example for a view spec.

sauloperez: Yep, I have only written few of them but when you need them they come in handy! The Rails testing book talks about it in https://github.com/thoughtbot/testing-rails/blob/master/book/types_of_tests/view_specs.md.

luisramos0: 👌

RachL:
@luisramos0 about main menu, it is really cool that it stays now for payment methods, too bad it's not the case for shipping methods :D

I wasn't sure whant you meant by "adjustments" states, but I did a comparison with UK staging and it looked the same, so I guessed it was good.

One question: I didn't see the menu "Jirafe" in main menu. Can you confirm that it was removed somewhere else and I missed it? If yes we can move this to ready to go and stage the next one :)

Notes: https://docs.google.com/document/d/1HI1VL-SvR6flu7HjKTbbkUGj2a_YbSwwzw3UpS_e1DI/edit#

luisramos0:
re "adjustments states", there were two big buttons there "open all adjustments" and "close all adjustments". Good that you didnt see them, that was the point :-)

Jirafe is gone in v2. I dont think anyone is using this, is it? anyway, it's not related to this PR.

RachL:
@luisramos0 nope I don't think anybody uses it :)

PR #3554, opened 2019-02-26
Bring in Bugsnag initializer from ofn-install

### LINKED ISSUE #1929: Bugsnag initializer should be in application code, not provisioning
Right now some instance use [Bugsnag](https://www.bugsnag.com/) to monitor application failures.

The configuration is set in provisioning [here](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/bugsnag.rb.j2) but was not activated, https://github.com/openfoodfoundation/ofn-install/pull/87 solves this.

Our opinion on this is that we should have the `bugsnag.rb` initializer in this repo, not in `ofn_install` and use `application.yml` to set all the credentials and any instance specific information through environment variables.

If an instance doesn't use Bugsnag they don't have to do anything, the initializer will just "early return".

This is one of those little things that make it easier to setup new instances and maintain existing one.

sauloperez:
#### What? Why?

Closes #1929

This copies ofn-install's `roles/app/templates/bugsnag.rb.j2` to `config/initializers/bugsnag.rb`. All tasks and templates regarding Bugsnag can then be :fire: from ofn-install.

As a result, it'll fix the issue that both Katuma and OFF are facing where the `config/initializer/bugsnag.rb` symlink to `shared/config/bugsnag.rb` does not exist thus, nothing gets notified to Bugsnag since December 3rd (according to customer support).

#### What should we test?

After the deploy any errors should appear in Bugsnag's dashboard. You can try running the following from a `bundle exec rails console` in staging

```ruby
Bugsnag.notify('foo')
```

#### Release notes

Changelog Category: Added

Bring into the app the Bugsnag initializer used in ofn-install. It won't be symlinked at deploy time anymore.

#### Dependencies

https://github.com/openfoodfoundation/ofn-install/pull/329

### DIFF
--- a/config/initializers/bugsnag.rb
+++ b/config/initializers/bugsnag.rb
@@ -1,0 +1,5 @@
+Bugsnag.configure do |config|
+  config.api_key = ENV['BUGSNAG_API_KEY']
+  config.notify_release_stages = %w(production staging)
+  config.use_ssl = true
+end


sauloperez:
Test in Katuma staging and working

sauloperez:
I need to check if a defensive `if` is really needed for instances that don't use Bugsnag (should be mandatory though).

luisramos0:
all right, if you have tested it in katuma it means it's ready to go!

sauloperez:
I have the hypothesys that it's due to the use of `none` and `""` in ofn-install but I didn't investigate @Matt-Yorkley.

sauloperez:
Double checked. If for any reason `ENV['BUGSNAG_API_KEY']` is nil `Bugsnag.notify` does not raise any error. No need for defensive code :+1:

Issue #2721, opened 2018-09-13
Transfer I18n translations used by the JS to the "js." namespace


kristinalim:
Preliminary work to reduce the speed and browser memory usage when loading the website

## Description

This was discussed [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2660#discussion_r216155016).

Before transferring a translation key, make sure that there are no other parts of the website using the translation. Otherwise, keep the old translation and just add a new one under "js.".

## Expected Behavior

All translation keys used in the JS are under the "js." namespace.

## Actual Behavior

Not all translation keys used in the JS are under the "js." namespace.

RachL:
@dacook as you dive into translation recently, do you know if this issue is still needed? We are doing a bit of clean up with Lynne :) 🧹

PR #3555, opened 2019-02-26
[Spree Upgrade] Improve order state translations in v2

### LINKED ISSUE #3542: [Spree upgrade] Translations missing for states in super admin /orders page
## Description
Translation key are missing for super admin on /orders.

For a producer it is fine however.

## Steps to Reproduce
1. Log in as a super admin in Staging FR
2. Go to https://staging.openfoodfrance.org/admin/orders?q%5Bs%5D=completed_at+desc

## Animated Gif/Screenshot
Super admin:

Producer:

## Severity

## Your Environment

* Version used: ??
* Browser name and version: Firefox 64 and Chromium 73
* Operating System and version (desktop or mobile): Ubuntu 18;10

luisramos0:
#### What? Why?

Closes #3542

Here we are improving the way we manage order state translations by using en.spree.order_state on haml code and js.admin.orders.order_state in angular (remember the objective with js translations set in #2721). See what translations were changed in "what to test" below.

#### What should we test?
We need to validate the places where we are changing translations.

In the orders list:
- the order states in the filters
- the name of the columns: Payment State, Shipment State, Completed At, Number, State, Customer E-Mail
- each value of the order state column
- each value of the shipment state column
- each value of the payments state column

In order edit page, the state of the order on the right side panel

In the subscriptions order panel, the orders state (this one can’t be tested yet in v2).

#### Dependencies
To be tested, this PR needs translations coming from master in #3535

### DIFF
--- a/app/assets/javascripts/admin/subscriptions/controllers/orders_panel_controller.js.coffee
+++ b/app/assets/javascripts/admin/subscriptions/controllers/orders_panel_controller.js.coffee
@@ -19,4 +19,4 @@ angular.module("admin.subscriptions").controller "OrdersPanelController", ($scop
     text = if closes_at > moment() then t('js.subscriptions.closes') else t('js.subscriptions.closed')
     "#{text} #{closes_at.fromNow()}"
 
-  $scope.stateText = (state) -> t("spree.order_state.#{state}")
+  $scope.stateText = (state) -> t("js.admin.orders.order_state.#{state}")
--- a/app/views/spree/admin/orders/_filters.html.haml
+++ b/app/views/spree/admin/orders/_filters.html.haml
@@ -11,7 +11,7 @@
       .field
         = label_tag nil, t(:status)
         = select_tag("q[state_eq]",
-            options_for_select(Spree::Order.state_machines[:state].states.collect {|s| [t("order_state.#{s.name}"), s.value]}),
+            options_for_select(Spree::Order.state_machines[:state].states.collect {|s| [t("spree.order_state.#{s.name}"), s.value]}),
             {include_blank: true, class: 'select2', 'ng-model' => 'q.state_eq'})
     .four.columns
       .field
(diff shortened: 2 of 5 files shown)


sauloperez:
I added the dependency. Do you want to proceed with the testing as usual after #3535 gets merged?

luisramos0:
yes, I think we can now switch to normal process for all v2 PRs with tests in staging

luisramos0:
I rebased this to get all the translations from #3535

luisramos0:
I manually put the latest **FR** translations from transifex on top of this PR and it's looking really good on my local but in AU staging there must be some cache of the JS translations and some are not showing up in French...
@kristinalim @sauloperez @mkllnk would you know what this is? JS translations not being picked up in the order states translations here: https://staging.openfoodnetwork.org.au/admin/orders

sigmundpetersen:
I've seen something like this before and I think there was some mention of asset precompile step on deploy. Do you know if this was performed? 

Maybe also a unicorn/nginx restart would be needed

mkllnk:
When we had issues with translations not appearing it was because not all assets got recompiled. It helped to delete `tmp/cache` but that won't help here. The server is now using ofn-install which creates a new directory during deploy, without `tmp`. The assets folder is shared though. Somehow it may think that the assets are already compiled. You can try to delete all compiled assets and then deploy again.

luisramos0:
Master Maikel! Thank you. I just deleted everything from apps/openfoodnetwork/shared/assets and re-deployed, fixed :-)

This is ready for testing in staging AUS.

luisramos0:
re canceled - I see there's a shipment_states.canceled entry missing, maybe that is causing the problem. I am not 100% sure. Anyway, it's better to create a separate issue as the fix will be a new translation key to master (needs a PR against master, this one is a v2 PR).

re other states - you can see the cart state by listing orders not completed. if you dont see any order in cart, just add an item to the cart on the frontoffice and you will see a new order in the backoffice.
I am not sure about the address and payment states, I think they will be seen when a shipping method validation or a payment fails respectively. Maybe this can be achieved in the backoffice by creating a payment that fails? I think this question deserves an issue, even if tech debt issue. we should clarify when would an order end up in one of these states...

RachL:
@luisramos0 I've just noticed that if you filter on "cart" the checkbox "show completed orders only" remains checked. So my guess is that I forgot to uncheck it during my tests. Let's move forward with this PR. I will create a new issue for the cancel button missing and if I see further translations missing in my tests, I will do the same.

PR #3558, opened 2019-02-26
[Spree Upgrade] WIP _Trying_ to fix unreliable customers_spec

### LINKED ISSUE #3520: [Spree upgrade][Flaky spec] `spec/features/admin/customers_spec.rb:184`

Pull requests for our Spree upgrade branch `2-0-stable` regularly fail `spec/features/admin/customers_spec.rb:184`. The unreliable spec feedback costs time.

### Context

https://github.com/openfoodfoundation/openfoodnetwork/pull/3394
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3394/builds/6

### Impact and timeline

This is not as important as flaky specs on the master branch but the sooner we fix it the faster we can deliver the Spree upgrade.

luisramos0:
#### What? Why?

Closes #3520
The spec opens the edit address dialog, puts a new value "New Address1" and save the data. The save action should close the dialog and spec goes on to expect the unerlying list to be updated with the new address. This is done through angular data bindings.

The specs fails ocasionally probably because the angular bindings do not have time to update the data and capybara is not waiting. I don't know how to find out when angular has finished its event processing...

#### What should we test?
customers_spec not broken after 10 runs :-)

### DIFF
--- a/spec/features/admin/customers_spec.rb
+++ b/spec/features/admin/customers_spec.rb
@@ -197,11 +197,11 @@ feature 'Customers' do
           fill_in 'address1', with: "New Address1"
           click_button 'Update Address'
 
-          expect(page).to have_selector "th.bill_address"
           expect(page).to have_content 'Address updated successfully.'
-          expect(page).to have_link 'New Address1'
           expect(customer4.reload.bill_address.address1).to eq 'New Address1'
+          expect(page).to have_selector 'th.bill_address'
 
+          expect(page).to have_link 'New Address1'
           first('#bill-address-link').click
           
           expect(page).to have_content 'Edit Billing Address'
--- a/spec/features/admin/customers_spec.rb
+++ b/spec/features/admin/customers_spec.rb
@@ -197,6 +197,7 @@ feature 'Customers' do
           fill_in 'address1', with: "New Address1"
           click_button 'Update Address'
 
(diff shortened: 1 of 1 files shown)


sigmundpetersen:
It failed again on the same line, but I think the error is a bit different

luisramos0:
really interesting errors...
I dont think there's any "New A" or "First345678" in the system, it looks like it's a problem filling in the values, "First345678" looks like a weird mix up between first name and phone number!?

```

 1) Customers as an enterprise user using the customers index updating a customer addresses updates the existing billing address
     Failure/Error: expect(customer4.reload.bill_address.address1).to eq 'New Address1'
     
       expected: "New Address1"
            got: "New A"
     
       (compared using ==)
     # ./spec/features/admin/customers_spec.rb:201:in `block (5 levels) in <top (required)>'

  2) Customers as an enterprise user using the customers index updating a customer addresses creates a new shipping address
     Failure/Error: expect(ship_address.firstname).to eq 'First'
     
       expected: "First"
            got: "First345678"
     
       (compared using ==)
     # ./spec/features/admin/customers_spec.rb:233:in `block (5 levels) in <top (required)>'


```

luisramos0:
#3620 will do it!

PR #3559, opened 2019-02-26
[Spree Upgrade] WIP Remove update_shipping_fees! and update_payment_fees! from order_decorator

luisramos0:
#### What? Why?

Related to #1830
Exploring the removal of order_decorator.update_shipping_fees and order_decorator.update_payment_fees.

This logic _obivously_ has to be moved to order_updater where all this is already happening in spree...

#### What should we test?
Build is green and order_decorator is not handling any kind of updates, all order updates are handled by order_updater.

### DIFF
--- a/app/controllers/line_items_controller.rb
+++ b/app/controllers/line_items_controller.rb
@@ -39,8 +39,6 @@ class LineItemsController < BaseController
     order = item.order
     order.with_lock do
       item.destroy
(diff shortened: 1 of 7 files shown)


luisramos0:
closing, 
#1830 addressed in #3666 
it looks like we are not going to delete update_shipping_fees and update_payment_fees yet...

PR #3560, opened 2019-02-27
2696,2788 [Spree Upgrade] Fix use of shipping method ID for subscriptions

### LINKED ISSUE #2696: [Spree Upgrade] Fix subscriptions.shipping_method_id (13 broken specs)
Subscriptions and spree 2 shipping_methods will need to work together.
There's a column subscriptions.shipping_method_id that, I believe, is used to create orders in a subscription. We need to verify how this process works with the new spree 2 way of creating shipments from shipping methods.

Files referencing shipping_method_id:
          - app/serializers/api/admin/subscription_serializer.rb (this used to be a separate issue #2732 that can be reopened if appropriate)
          - app/models/proxy_order.rb
          - app/services/order_syncer.rb
          - app/assets/javascripts/admin/subscriptions/services/subscription_functions.js.coffee
          - app/views/admin/subscriptions/_details.html.haml
          - app/views/admin/subscriptions/_table.html.haml
          - spec/controllers/admin/subscriptions_controller_spec.rb
          - spec/services/subscription_form_spec.rb
          - spec/services/order_syncer_spec.rb
          - spec/features/admin/subscriptions_spec.rb
          - spec/controllers/admin/proxy_orders_controller_spec.rb
          - spec/javascripts/unit/admin/subscriptions/services/subscription_functions_spec.js.coffee

kristinalim:
Fix use of shipping method ID for subscriptions.

#### What? Why?

- Closes #2696
- Relates to #2788 (does not close issue)

The shipping method in Spree 2.0 is not associated directly to orders, but through shipment shipping rates.

Before Spree 2, `OrderFactory` keeps subscription orders in cart state but already sets associations and creates associated records (like payments). But in Spree 2, transitioning an order to "delivery" state removes and regenerates shipments. Considering above, this means that the selected shipping method could be forgotten.

To work around this, the code that regenerates shipping rates for a shipment has been updated to remember the previously selected shipping method first and select it again after creating the new shipping rates.

#### What should we test?

Not worth testing as there are still some broken functionalities. But some unit and feature tests have been uncommented and now pass.

### DIFF
--- a/app/controllers/spree/admin/orders/customer_details_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders/customer_details_controller_decorator.rb
@@ -1,6 +1,21 @@
 Spree::Admin::Orders::CustomerDetailsController.class_eval do
   before_filter :set_guest_checkout_status, only: :update
 
+  def update
+    if @order.update_attributes(params[:order])
+      if params[:guest_checkout] == "false"
+        @order.associate_user!(Spree.user_class.find_by_email(@order.email))
+      end
+      while @order.next; end
+
+      @order.shipments.map &:refresh_rates
+      flash[:success] = Spree.t('customer_details_updated')
+      redirect_to admin_order_customer_path(@order)
+    else
+      render :action => :edit
+    end
+  end
+
   # Inherit CanCan permissions for the current order
   def model_class
     load_order unless @order
--- a/app/controllers/spree/admin/orders/customer_details_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders/customer_details_controller_decorator.rb
@@ -6,7 +6,8 @@ Spree::Admin::Orders::CustomerDetailsController.class_eval do
       if params[:guest_checkout] == "false"
         @order.associate_user!(Spree.user_class.find_by_email(@order.email))
       end
-      while @order.next; end
+
+      AdvanceOrderService.new(@order).call
 
       @order.shipments.map &:refresh_rates
       flash[:success] = Spree.t('customer_details_updated')
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -27,6 +27,18 @@ Spree::Admin::OrdersController.class_eval do
     # within the page then fetches the data it needs from Api::OrdersController
   end
 
+  def edit
+    @order.shipments.map &:refresh_rates
+
+    # Transition as far as we can go
+    while @order.next; end
+
+    # The payment step shows an error of 'No pending payments'
+    # Clearing the errors from the order object will stop this error
+    # appearing on the edit page where we don't want it to.
+    @order.errors.clear
+  end
+
   # Re-implement spree method so that it redirects to edit instead of rendering edit
   #   This allows page reloads while adding variants to the order (/edit), without being redirected to customer details page (/update)
   def update
--- a/app/controllers/spree/admin/orders_controller_decorator.rb
+++ b/app/controllers/spree/admin/orders_controller_decorator.rb
@@ -30,8 +30,7 @@ Spree::Admin::OrdersController.class_eval do
   def edit
     @order.shipments.map &:refresh_rates
 
-    # Transition as far as we can go
-    while @order.next; end
+    AdvanceOrderService.new(@order).call
 
     # The payment step shows an error of 'No pending payments'
     # Clearing the errors from the order object will stop this error
--- a/app/controllers/spree/admin/payments_controller_decorator.rb
+++ b/app/controllers/spree/admin/payments_controller_decorator.rb
@@ -1,6 +1,37 @@
 Spree::Admin::PaymentsController.class_eval do
   append_before_filter :filter_payment_methods
 
+  def create
+    @payment = @order.payments.build(object_params)
+    if @payment.payment_method.is_a?(Spree::Gateway) && @payment.payment_method.payment_profiles_supported? && params[:card].present? and params[:card] != 'new'
+      @payment.source = CreditCard.find_by_id(params[:card])
+    end
+
+    begin
+      unless @payment.save
+        redirect_to admin_order_payments_path(@order)
+        return
+      end
+
+      if @order.completed?
+        @payment.process!
+        flash[:success] = flash_message_for(@payment, :successfully_created)
+
+         redirect_to admin_order_payments_path(@order)
+      else
+        #This is the first payment (admin created order)
+        until @order.completed?
+          @order.next!
+        end
+        flash[:success] = Spree.t(:new_order_completed)
+        redirect_to edit_admin_order_url(@order)
+      end
+
+    rescue Spree::Core::GatewayError => e
+      flash[:error] = "#{e.message}"
+      redirect_to new_admin_order_payment_path(@order)
+    end
+  end
 
   # When a user fires an event, take them back to where they came from
   # (we can't use respond_override because Spree no longer uses respond_with)
--- a/app/controllers/spree/admin/payments_controller_decorator.rb
+++ b/app/controllers/spree/admin/payments_controller_decorator.rb
@@ -19,7 +19,7 @@ Spree::Admin::PaymentsController.class_eval do
 
          redirect_to admin_order_payments_path(@order)
       else
-        AdvanceOrderService.new(@order).call(true)
+        AdvanceOrderService.new(@order).call!
 
         flash[:success] = Spree.t(:new_order_completed)
         redirect_to edit_admin_order_url(@order)
--- a/app/controllers/spree/admin/payments_controller_decorator.rb
+++ b/app/controllers/spree/admin/payments_controller_decorator.rb
@@ -19,10 +19,8 @@ Spree::Admin::PaymentsController.class_eval do
 
          redirect_to admin_order_payments_path(@order)
       else
-        #This is the first payment (admin created order)
-        until @order.completed?
-          @order.next!
-        end
+        AdvanceOrderService.new(@order).call(true)
+
         flash[:success] = Spree.t(:new_order_completed)
         redirect_to edit_admin_order_url(@order)
       end
--- a/app/services/advance_order_service.rb
+++ b/app/services/advance_order_service.rb
@@ -5,15 +5,31 @@ class AdvanceOrderService
     @order = order
   end
 
-  def call
-    shipping_method_id = @order.shipping_method.id if @order.shipping_method.present?
+  def call(raise_on_error = false)
+    shipping_method_id = order.shipping_method.id if order.shipping_method.present?
+    options = { shipping_method_id: shipping_method_id }
+    raise_on_error ? advance_order!(options) : advance_order(options)
+  end
 
-    while @order.state != "complete"
-      break unless @order.next
+  private
 
-      if @order.state == "delivery"
-        @order.select_shipping_method(shipping_method_id) if shipping_method_id.present?
-      end
+  def advance_order(options)
+    until order.state == "complete"
+      break unless order.next
+      after_transition_hook(options)
+    end
+  end
+
+  def advance_order!(options)
+    until order.completed?
+      order.next!
+      after_transition_hook(options)
+    end
+  end
+
+  def after_transition_hook(options)
+    if order.state == "delivery"
+      order.select_shipping_method(options[:shipping_method_id]) if options[:shipping_method_id]
     end
   end
 end
--- a/app/services/advance_order_service.rb
+++ b/app/services/advance_order_service.rb
@@ -5,13 +5,20 @@ class AdvanceOrderService
     @order = order
   end
 
-  def call(raise_on_error = false)
-    shipping_method_id = order.shipping_method.id if order.shipping_method.present?
-    options = { shipping_method_id: shipping_method_id }
-    raise_on_error ? advance_order!(options) : advance_order(options)
+  def call
+    advance_order(advance_order_options)
+  end
+
+  def call!
+    advance_order!(advance_order_options)
   end
 
   private
+
+  def advance_order_options
+    shipping_method_id = order.shipping_method.id if order.shipping_method.present?
+    { shipping_method_id: shipping_method_id }
+  end
 
   def advance_order(options)
     until order.state == "complete"
--- a/app/services/advance_order_service.rb
+++ b/app/services/advance_order_service.rb
@@ -1,0 +1,11 @@
+class AdvanceOrderService
+  attr_reader :order
+
+  def initialize(order)
+    @order = order
+  end
+
+  def call
+    while order.next; end
+  end
+end
--- a/app/services/advance_order_service.rb
+++ b/app/services/advance_order_service.rb
@@ -6,6 +6,14 @@ class AdvanceOrderService
   end
 
   def call
-    while order.next; end
+    shipping_method_id = @order.shipping_method.id if @order.shipping_method.present?
+
+    while @order.state != "complete"
+      break unless @order.next
+
+      if @order.state == "delivery"
+        @order.select_shipping_method(shipping_method_id) if shipping_method_id.present?
+      end
+    end
   end
 end
--- a/app/services/order_factory.rb
+++ b/app/services/order_factory.rb
@@ -15,8 +15,10 @@ class OrderFactory
     set_user
     build_line_items
     set_addresses
+    create_shipment
     set_shipping_method
     create_payment
+
     @order
   end
 
@@ -64,6 +66,10 @@ class OrderFactory
 
   def set_addresses
     @order.update_attributes(attrs.slice(:bill_address_attributes, :ship_address_attributes))
+  end
+
+  def create_shipment
+    @order.create_proposed_shipments
   end
 
   def set_shipping_method
--- a/app/services/order_factory.rb
+++ b/app/services/order_factory.rb
@@ -15,6 +15,7 @@ class OrderFactory
     set_user
     build_line_items
     set_addresses
+    set_shipping_method
     create_payment
     @order
   end
@@ -63,6 +64,10 @@ class OrderFactory
 
   def set_addresses
     @order.update_attributes(attrs.slice(:bill_address_attributes, :ship_address_attributes))
+  end
+
+  def set_shipping_method
+    @order.select_shipping_method(attrs[:shipping_method_id])
   end
 
   def create_payment
--- a/app/services/order_syncer.rb
+++ b/app/services/order_syncer.rb
@@ -67,13 +67,12 @@ class OrderSyncer
 
   def update_shipment_for(order)
     shipment = order.shipment
+    return track_shipping_method_issue(order) if shipment.blank?
 
-    if shipment.andand.state == "pending" && shipment.shipping_method.id == shipping_method_id_was
+    if shipment.state == "pending" && shipment.shipping_method.id == shipping_method_id_was
       order.select_shipping_method(shipping_method_id)
-    else
-      unless shipment.andand.state == "pending" && shipment.shipping_method.id == shipping_method_id
-        order_update_issues.add(order, I18n.t('admin.shipping_method'))
-      end
+    elsif !(shipment.state == "pending" && shipment.shipping_method.id == shipping_method_id)
+      track_shipping_method_issue(order)
     end
   end
 
@@ -105,5 +104,9 @@ class OrderSyncer
     relevant_address_attrs.all? do |attr|
       order.ship_address[attr] == distributor_address[attr]
     end
+  end
+
+  def track_shipping_method_issue(order)
+    order_update_issues.add(order, I18n.t('admin.shipping_method'))
   end
 end
--- a/app/services/order_syncer.rb
+++ b/app/services/order_syncer.rb
@@ -69,8 +69,7 @@ class OrderSyncer
     shipment = order.shipment
 
     if shipment.andand.state == "pending" && shipment.shipping_method.id == shipping_method_id_was
-      shipment.update_attributes(shipping_method_id: shipping_method_id)
-      order.update_attribute(:shipping_method_id, shipping_method_id)
+      order.select_shipping_method(shipping_method_id)
     else
       unless shipment.andand.state == "pending" && shipment.shipping_method.id == shipping_method_id
         order_update_issues.add(order, I18n.t('admin.shipping_method'))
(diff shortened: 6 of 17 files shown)


app/services/order_syncer.rb
```diff
@@ -66,12 +66,16 @@ def update_payment_for(order)
...
+      shipment = Spree::Shipment.find(shipment.id)
+
+      select_shipping_method_for_shipment(shipment, shipping_method)
       order.update_attribute(:shipping_method_id, shipping_method_id)
     else
-      unless order.shipments.with_state('pending').where(shipping_method_id: shipping_method_id).any?
+      unless pending_shipment_using_shipping_method(order, shipping_method_id).present?
         order_update_issues.add(order, I18n.t('admin.shipping_method'))
       end
     end
```
luisramos0: see order_shipment.rb, in ofn v2 we have one very useful rule (it makes everything much easier): one order can only have one shipment, there will never be 2 shipments in an order. So you can use order.shipment.

also, spree::shipment.selected_shipping_rate is kept to single selection in spree::shipment.selected_shipping_rate_id=
so you can always use order.shipment.selected_shipping_rate

also, you can simply use **order_shipment.shipping_method** that redirects to spree::shipment.shipping_method that redirects to spree::shipment.selected_shipping_rate :-)

I think these facts will help simplify this code, right?

app/services/order_syncer.rb
```diff
@@ -106,4 +110,22 @@ def force_ship_address_required?(order)
       order.ship_address[attr] == distributor_address[attr]
     end
   end
+
+  def select_shipping_method_for_shipment(shipment, shipping_method)
+    shipping_rate = find_or_create_shipping_rate_for_shipping_method(shipment, shipping_method)
+    shipment.update_attributes(selected_shipping_rate_id: shipping_rate.id)
```
luisramos0: should be shipment.selected_shipping_rate_id=(shipping_rate.id)

app/services/order_syncer.rb
```diff
@@ -106,4 +110,22 @@ def force_ship_address_required?(order)
...
+  end
+
+  def find_or_create_shipping_rate_for_shipping_method(shipment, shipping_method)
+    shipping_rate = shipment.shipping_rates.where(shipping_method_id: shipping_method.id).last
```
luisramos0: as in order_shipment.select_shipping_method you can do:
shipping_rate = shipment.shipping_rates.find_by_shipping_method_id(shipping_method_id)

app/services/order_syncer.rb
```diff
@@ -66,12 +66,16 @@ def update_payment_for(order)
   end
 
   def update_shipment_for(order)
-    shipment = order.shipments.with_state('pending').where(shipping_method_id: shipping_method_id_was).last
+    shipment = pending_shipment_using_shipping_method(order, shipping_method_id_was)
+
     if shipment
```
luisramos0: maybe this can simply be
`if order.shipment.shipping_method == shipping_method_id_was?`

spec/services/order_syncer_spec.rb
```diff
@@ -34,27 +34,31 @@
...
 
         before do
           # Updating the shipping method on a shipment updates the shipping method on the order,
           # and vice-versa via logic in Spree's shipments controller. So updating both here mimics that
           # behaviour.
-          order.shipments.first.update_attributes(shipping_method_id: changed_shipping_method.id)
-          order.update_attributes(shipping_method_id: changed_shipping_method.id)
+          shipment = order.shipments.first
+          shipping_rate = shipment.shipping_rates.where(shipping_method_id: changed_shipping_method.id).first
+          shipment.update_attributes(selected_shipping_rate_id: shipping_rate.id)
```
luisramos0: this is probably order.shipment.select_shipping_method(changed_shipping_method.id)

spec/services/order_syncer_spec.rb
```diff
@@ -34,27 +34,31 @@
...
+          order.update_attributes!(shipping_method_id: changed_shipping_method.id)
```
luisramos0: see above regarding the column order.shipping_method_id

app/services/order_factory.rb
```diff
@@ -14,8 +14,22 @@ def create
...
     set_addresses
+
+    # Transition to :delivery state.
+    @order.next!
+
+    set_shipping_method
+
+    # Transition to :payment state.
+    @order.next!
+
```
luisramos0: you are assuming a specific order workflow here...

also, do you know what happens if one of these steps fails for some reason?
you could run @order.next! only if order.state is the expected state and you could do something if @order.next! fails

I am sorry I dont have a specific suggestion for this one...

app/services/order_factory.rb
```diff
@@ -65,6 +79,10 @@ def set_addresses
     @order.update_attributes(attrs.slice(:bill_address_attributes, :ship_address_attributes))
   end
 
+  def set_shipping_method
+    @order.update_attributes!(shipping_method_id: attrs[:shipping_method_id])
```
luisramos0: the column order.shipping_method_id is deprecated and will be deleted after v2 is live. we cannot rely on it.
this is how you do it in v2:
@order.shipment.select_shipping_method(attrs[:shipping_method_id])

app/services/order_syncer.rb
kristinalim: @luisramos0 Using `update_attributes` also works because `selected_shipping_rate_id` is `attr_accessible`: https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/core/app/models/spree/shipment.rb#L19 And specs would fail if in the future it won't allow mass assignment anymore.

kristinalim:
Thanks for the info, @luisramos0 - that's very helpful info and indeed makes this simpler.

app/services/order_factory.rb
kristinalim: @luisramos0 What do you think about this approach: https://github.com/openfoodfoundation/openfoodnetwork/pull/3560/commits/7206cd88995ba52b77763a22d377c8564505a164 https://github.com/openfoodfoundation/openfoodnetwork/pull/3560/commits/dd2f7ecac65e3222afddd20748243fcfd8691c66

With these changes, shipment records are now created without changing the order status (similar to what is done for payments), and we now remember the previously selected shipping method and select it again after the new shipping rates have been created.

kristinalim:
@luisramos0 Thanks for that feedback.

I have already been working on this, but then noticed this code in Spree `admin/orders#edit` which [transitions orders as far as it can](https://github.com/openfoodfoundation/spree/blob/v2.0.4/backend/app/controllers/spree/admin/orders_controller.rb#L57). :face_with_head_bandage:

```ruby
while @order.next; end
```

So that was why I kept seeing in the UI that subscription orders were completed... I suppose we're keeping this behaviour to keep the workflow similar to Spree? This behaviour wasn't in the previous version.

And then I checked Spree 2.4, and it looks like it remembers the shipping method when regenerating the shipping rates:

- [`admin/orders#edit`](https://github.com/spree/spree/blob/2-4-stable/core/app/models/spree/shipment.rb)
- [`Spree::Order#refresh_shipment_rates`](https://github.com/spree/spree/blob/2-4-stable/core/app/models/spree/order.rb#L556)
- [`Spree::Shipment#refresh_rates`](https://github.com/spree/spree/blob/2-4-stable/core/app/models/spree/shipment.rb#L221)

```ruby
      # StockEstimator.new assigment below will replace the current shipping_method
      original_shipping_method_id = shipping_method.try(:id)

      self.shipping_rates = Stock::Estimator.new(order).shipping_rates(to_package)

      if shipping_method
        selected_rate = shipping_rates.detect { |rate|
          rate.shipping_method_id == original_shipping_method_id
        }
        self.selected_shipping_rate_id = selected_rate.id if selected_rate
      end
```

Considering above, I think that overriding the `Spree::Order` method would be the best approach for now.

Let me know what you think. Moving this back to Code Review for further critique.

luisramos0:
yes, that while order.next is so painful!

I see you prefer this version. I still prefer the previous version or some form of service that handles this logic outside of spree models. The main reason is that this overriding is not necessary and by creating a service with a clear API we would be able to improve the messy spree checkout workflow code.

Let's see what other reviewers say.

luisramos0:
yes, the order workflow will fail (it will be stuck in delivery) if no shipping methods are available. another case is that the selection gets lost if the selected shipping method becomes unavailable. But that will be a problem in both implementations: in or out of the spree model.

kristinalim:
I was wrong in my previous statement, where I said future versions of Spree remember the shipping method. This remembering only happens in `Spree::Shopment#refresh_rates`, which happens separately from transitioning an order (I got confused). This is called in [`admin/orders#edit`](https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/backend/app/controllers/spree/admin/orders_controller.rb#L55) and [`admin/orders/customer_details#update`](https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/backend/app/controllers/spree/admin/orders/customer_details_controller.rb#L25), separate from `while @order.next; end`. I think the goal mostly is to make sure that `Spree::ShippingRate` records exist and are calculated where needed.

When transitioning, it still calls `Spree::Order#create_proposed_shipments` which continues to remove shipments, until [2-4-stable](https://github.com/spree/spree/blob/2-4-stable/core/app/models/spree/order/checkout.rb#L96).

> What if the method got changed before the end of the order cycle? Would that change get lost in the old implementation?

@mkllnk When the shipping method in the subscription is changed before the OC closes, `OrderSyncer` updates the shipping method in future orders IF the shipping method in the order is still the same as that in the subscription (meaning it wasn't customized in the order). This behaviour is described in specs [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/23cf0729984b4827c8b83146231e3c83f46732b7/spec/services/order_syncer_spec.rb#L11).

luisramos0:
@kristinalim fyi as it is related to this PR, as I was playing with live data in v2, I realized that refresh rates should only be called if order is not complete: https://github.com/spree/spree/blob/3c46d141c5cb2bedced63596178a0a79df3f0c9e/backend/app/controllers/spree/admin/orders_controller.rb#L54-L60

I think we'll need to bring this change to our spree 2-0-4 branch otherwise, editing an old order may just break it as shipping methods may not be setup appropriately for older orders.

kristinalim:
> I think we'll need to bring this change to our spree 2-0-4 branch otherwise, editing an old order may just break it as shipping methods may not be setup appropriately for older orders.

@luisramos0 I'll do it if you're not working on it yet. :slightly_smiling_face:

luisramos0:
no, I am not working on it. please go ahead :-)

I think we need to keep the workflow .next loop in these methods, because otherwise the page flow (moving automatically to customer details) will not work.

mkllnk:
@kristinalim 
> When the shipping method in the subscription is changed before the OC closes, OrderSyncer updates the shipping method in future orders

That's not what I meant. I was thinking of the scenario that a single order is changed without touching the subscription. From the description of your previous implementation it sounded like that change would get lost because you would use the proxy order as reference for the shipping method. If the change is done in the proxy order as well (I don't know) then it's no problem and your previous implementation was correct.

I'll wait for your next version and review again. :-)

kristinalim:
Thanks, @mkllnk!

> That's not what I meant. I was thinking of the scenario that a single order is changed without touching the subscription. From the description of your previous implementation it sounded like that change would get lost because you would use the proxy order as reference for the shipping method.

The implementation was using the previous shipping method of the _order_ before the transition, whether or not this was created for a subscription, so it wasn't losing the data. Does this answer your concern?

Anyway, I've already moved the order advancement code to a service (pretty okay with the idea now). :slightly_smiling_face: I'm just making a last change of having other places advancing the order state to use this service as well, if the initial order state is not specific.

kristinalim:
This is now back in Code Review. Thanks for your feedback, @luisramos0 and @mkllnk!

app/controllers/spree/admin/orders_controller_decorator.rb
```diff
@@ -27,6 +27,17 @@ def index
     # within the page then fetches the data it needs from Api::OrdersController
   end
 
+  def edit
+    @order.shipments.map &:refresh_rates
```
luisramos0: there are two points about this method:
- there's a unless statement that we will need before the refresh shipping rates (out of scope for this PR)
https://github.com/spree/spree/blob/3c46d141c5cb2bedced63596178a0a79df3f0c9e/backend/app/controllers/spree/admin/orders_controller.rb#L54

- and also, the order advancement logic was removed in later versions of spree 2-0:
https://github.com/spree/spree/commit/8208b4bd0871b73be97bb2a7c7c1e0e4c04b3073#diff-1e2ba31309f1f1abd2a4b626036d316f

these are both out of scope of this PR.

app/controllers/spree/admin/payments_controller_decorator.rb
```diff
@@ -1,6 +1,35 @@
...
+
+        flash[:success] = Spree.t(:new_order_completed)
+        redirect_to edit_admin_order_url(@order)
+      end
+
+    rescue Spree::Core::GatewayError => e
+      flash[:error] = "#{e.message}"
+      redirect_to new_admin_order_payment_path(@order)
+    end
+  end
```
luisramos0: It's not good that we need to copy spree code to our code base but I am actually very happy that we are reclaiming this payments code to OFN, because the spree code is not good in v2 and it's a big mess in v3.7 [here](https://github.com/spree/spree/blob/e2d6599715932215635baf5a709693d9e9daad20/backend/app/controllers/spree/admin/payments_controller.rb#L25).

this kind of logic must go to a service, it's not the responsibility of the controller.
So I see this as the first step to improve it on our side.

It's out of scope for this PR.

app/services/advance_order_service.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_reader :order
+
+  def initialize(order)
+    @order = order
+  end
+
+  def call(raise_on_error = false)
+    shipping_method_id = order.shipping_method.id if order.shipping_method.present?
+    options = { shipping_method_id: shipping_method_id }
+    raise_on_error ? advance_order!(options) : advance_order(options)
```
luisramos0: I see this raise_on_error and advance_order! as YAGNI 🔥

app/services/order_syncer.rb
```diff
@@ -66,14 +66,13 @@ def update_payment_for(order)
...
-      unless order.shipments.with_state('pending').where(shipping_method_id: shipping_method_id).any?
-        order_update_issues.add(order, I18n.t('admin.shipping_method'))
-      end
+    shipment = order.shipment
+    return track_shipping_method_issue(order) if shipment.blank?
+
+    if shipment.state == "pending" && shipment.shipping_method.id == shipping_method_id_was
+      order.select_shipping_method(shipping_method_id)
+    elsif !(shipment.state == "pending" && shipment.shipping_method.id == shipping_method_id)
+      track_shipping_method_issue(order)
```
luisramos0: this is too difficult to read,
we should simplify this, even if only extracted to methods with good names.

or maybe just remove the elsif,
if we just have two ifs one after the other, it's more readable and more correct, in case the first assignment fails, right?

```ruby
    if shipment.state == "pending" && shipment.shipping_method.id == shipping_method_id_was
      order.select_shipping_method(shipping_method_id)
    end
    unless shipment.state == "pending" && shipment.shipping_method.id == shipping_method_id
      track_shipping_method_issue(order)
     end
```
and then, is this correct?
```ruby
    track_shipping_method_issue(order) unless shipment.state == "pending"
    if shipment.shipping_method.id == shipping_method_id_was
      order.select_shipping_method(shipping_method_id)
    end
    track_shipping_method_issue(order) unless shipment.shipping_method.id == shipping_method_id
```

app/services/advance_order_service.rb
```diff
@@ -0,0 +1,35 @@
+class AdvanceOrderService
```
luisramos0: we went for option 4 here
https://community.openfoodnetwork.org/t/how-to-name-a-new-service-in-ofn/1530

app/services/advance_order_service.rb
kristinalim: @luisramos0 This has implications in the handling in the places where this is used. For example, when creating a payment and the order cannot be transitioned all the way to completion, this would rollback. But in the edit order page, this would just stop where the state cannot be transitioned anymore.

I actually considered merging the two. But aside from raising an error where needed, I was being cautious about how `order.state == "complete"` is not the same as `order.completed?`. (An order that is completed and then cancelled would return `false` for the first condition but `true` for the next.)

I prefer not touch this for now, and think we should look into that later. What do you think? Spree 2.4-stable does not use the bang version in `admin/payments#create` anymore, and it can be :fire: without much audit then.

**Spree 2.4-stable:**

```
$ grep -r order.next | egrep -v 'spec|testing_support'
src/frontend/app/controllers/spree/orders_controller.rb:              @order.next if @order.cart?
src/frontend/app/controllers/spree/checkout_controller.rb:        unless @order.next
src/backend/app/controllers/spree/admin/orders/customer_details_controller.rb:            @order.next
src/backend/app/controllers/spree/admin/payments_controller.rb:            while @order.next; end
src/backend/app/controllers/spree/admin/payments_controller.rb:            # If "@order.next" didn't trigger payment processing already (e.g. if the order was
src/api/app/controllers/spree/api/checkouts_controller.rb:        @order.next!
src/api/app/controllers/spree/api/checkouts_controller.rb:        while @order.next; end
src/api/app/controllers/spree/api/checkouts_controller.rb:          if @order.completed? || @order.next
src/core/lib/tasks/exchanges.rake:        while order.state != order.checkout_steps[-2] && order.next; end
```

**Current Spree version:**

```
$ grep -r order.next | egrep -v 'spec|testing_support'
frontend/app/controllers/spree/orders_controller.rb:              @order.next_transition.run_callbacks if @order.cart?
frontend/app/controllers/spree/checkout_controller.rb:        unless @order.next
backend/app/controllers/spree/admin/orders/customer_details_controller.rb:            while @order.next; end
backend/app/controllers/spree/admin/orders_controller.rb:        while @order.next; end
backend/app/controllers/spree/admin/payments_controller.rb:              @order.next!
backend/app/controllers/spree/admin/payments_controller.rb:      # where order.next is called which leaves the order in payment step
api/app/controllers/spree/api/orders_controller.rb:        if @order.valid? && @order.next
api/app/controllers/spree/api/checkouts_controller.rb:          state_callback(:after) if @order.next
api/app/controllers/spree/api/checkouts_controller.rb:        @order.next!
api/app/controllers/spree/api/checkouts_controller.rb:          if @order.valid? && @order.next
```

app/services/order_syncer.rb
kristinalim: The first snippet works is equivalent with a `return` in the first block. :+1: 

But the next version is not equivalent. The current version does not always call `track_shipping_method_issue(order)` if shipment state is not pending, but this would.

Thinking about this...

kristinalim: I added a commit. It's a lot more readable now: https://github.com/openfoodfoundation/openfoodnetwork/pull/3560/commits/7505dd0410043d51065122c7db65c2032ce32226

Thanks, @luisramos0! :slightly_smiling_face:

luisramos0: ah, yes, awesome, much better!

app/services/advance_order_service.rb
luisramos0: yeah, I see now, I think we need to keep it until 2.2

it was in v2.2 they removed this here: https://github.com/spree/spree/commit/4eade0fa78f426f5a3a8db15789dee48c2f87fa9?diff=unified
and then added the one without the bang here: https://github.com/spree/spree/commit/59198f74cf156e3b09dbf16f8eca523e28ebec79

luisramos0: thanks for the explanation!

mkllnk: The syntax is not clear though. You don't know what the boolean means from reading this:
```rb
AdvanceOrderService.new(@order).call(true)
```

You could go with a named parameter:
```rb
AdvanceOrderService.new(@order).call(raise_on_error: true)
```

But in this case I would go with the bang convention:
```rb
AdvanceOrderService.new(@order).call!
```

If you put `options` into its own method it becomes really easy.

app/controllers/spree/admin/orders/customer_details_controller_decorator.rb
```diff
@@ -1,6 +1,22 @@
 Spree::Admin::Orders::CustomerDetailsController.class_eval do
   before_filter :set_guest_checkout_status, only: :update
 
+  def update
+    if @order.update_attributes(params[:order])
+      if params[:guest_checkout] == "false"
```
sauloperez: I'm a bit surprised to compare a boolean value to a string. Rails does parse JSON to Ruby primitives. How is it possible that we end up with a String and not with a FalseClass instance 🤔 ? I would have expected to read this as `if !params[:guest_checkout]`

app/services/advance_order_service.rb
```diff
@@ -0,0 +1,35 @@
...
+  def call(raise_on_error = false)
+    shipping_method_id = order.shipping_method.id if order.shipping_method.present?
+    options = { shipping_method_id: shipping_method_id }
+    raise_on_error ? advance_order!(options) : advance_order(options)
+  end
+
+  private
+
+  def advance_order(options)
+    until order.state == "complete"
```
sauloperez: It'd be great to stick to the approach in line 24 for consistency IMO

app/controllers/spree/admin/orders/customer_details_controller_decorator.rb
kristinalim: @sauloperez This isn't the case when the string is sent to the server as HTTP form data. However, this string would still be converted to a boolean as we expect when assigned to an `ActiveRecord` boolean attribute. :slightly_smiling_face:

app/services/advance_order_service.rb
kristinalim: I gave this a try just now, @sauloperez, and confirmed that a completed but later cancelled order would attach an error here (`ActiveRecord::Base#errors`) whereas using `order.completed?` would not. `Spree::Order#completed?` looks at `completed_at`, not the order state.

We can actually remove `advance_order!` in Spree 2.2. (Discussion here: [heere](https://github.com/openfoodfoundation/openfoodnetwork/pull/3560#discussion_r265746751)) Are you okay with waiting until then so we avoid doing a more comprehensive check?

app/services/advance_order_service.rb
kristinalim: Thanks, @mkllnk! I changed this here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3560/commits/94606025c9c881c61fe52c07dc9adf4f86f35962 I also added a couple of specs to cover `call!`.

sauloperez:
Merging as suggested by @kristinalim . It seems to be passing those 13 specs.

PR #3563, opened 2019-02-28
3562 Fix error calculating payment fee using order-based calculator

### LINKED ISSUE #3562: Not possible anymore to manually capture a payment on production on some edge cases
## Description
From general order page, or if I'm in edit mode of a specific order, when I click on "capture payment", I got the snail page.

## Expected Behavior
I can capture a payment so the order is marked as paid.

## Actual Behaviour
It's not

## Steps to Reproduce
1. Go to https://www.openfoodfrance.org/admin/orders?q%5Bs%5D=completed_at+desc and try to capture a payment on an order.
2. Go into an order, sub menu "payments", same, try to capture a payment.

## Animated Gif/Screenshot
capture du 2019-02-27 20-30-45
capture d ecran de 2019-02-27 20-25-06

capture du 2019-02-28 16-43-56

## Context
On French production, was documenting how to manage refunds on OFN in the user guide...

## Severity
S2

## Your Environment
* Version used: 1.27
* Operating System and version (desktop or mobile): Ubuntu 1.18
* OFN Platform instance where you discovered the bug, and which version of the software they are using : French production

kristinalim:
#### What? Why?

Closes #3562 

The edge case (not sure) error happens with a particular sequence of order sequence of order creation, payment creation, and order completion when the payment method calculator is order-based.

I actually wasn't able to reproduce this from the UI using the usual checkout flow nor from the admin section, but there is a code path for this.

I was initially didn't want to remove the stock check portion of this code without a full audit, but it wasn't working and it seems more inappropriate to leave it there. Stock checking should already happen in the line items, and the line items should be the source of truth to the perspective of the payment fee calculator. @luisramos0 said this code was also [removed](https://github.com/openfoodfoundation/openfoodnetwork/commit/2ccbb87adcb3174bb5a64b57b6fc9dd4ab2b3f02#diff-e2ebcdd38cd6a972dc29886b9cad8ff1) in Spree v2.

#### What should we test?

(For below, create a cheque payment method with "Flat Percent" calculator for the distributor. Take note of the % rate.)

The following did not cause the reported problem, but should be tested.

**As customer:**

1. Make an order selecting the payment method earlier.
2. Check that the transaction fee is correct.

**As enterprise manager:**

1. Create an order selecting the payment method earlier.
2. Complete the order without capturing payment. Check that the transaction fee is correct.
3. Capture payment. Check that the transaction fee is still correct.
4. Create another order.
5. Complete the order without capturing payment.
6. Mark the uncaptured payment void. This should not fail.
7. Add another payment to the order. This should not fail.
8. Capture payment. Check that the transaction fee is still correct.

(**UPDATE:** Added Steps 4 to 8)

#### Release notes

- Fix calculation of payment transaction fee when the order is already completed, the calculator is order-based, and inventory tracking is enabled

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/payment_decorator.rb
+++ b/app/models/spree/payment_decorator.rb
@@ -33,7 +33,8 @@ module Spree
     # This is called by the calculator of a payment method
     def line_items
       if order.complete? && Spree::Config[:track_inventory_levels]
-        order.line_items.select { |li| inventory_units.pluck(:variant_id).include?(li.variant_id) }
+        inventory_item_ids = order.distributor.inventory_items.pluck(:variant_id)
+        order.line_items.select { |li| inventory_item_ids.include?(li.variant_id) }
       else
         order.line_items
       end
--- a/app/models/spree/payment_decorator.rb
+++ b/app/models/spree/payment_decorator.rb
@@ -32,12 +32,7 @@ module Spree
 
     # This is called by the calculator of a payment method
     def line_items
-      if order.complete? && Spree::Config[:track_inventory_levels]
-        inventory_item_ids = order.distributor.inventory_items.pluck(:variant_id)
-        order.line_items.select { |li| inventory_item_ids.include?(li.variant_id) }
-      else
-        order.line_items
-      end
+      order.line_items
     end
 
     # Pin payments lacks void and credit methods, but it does have refund
--- a/app/models/spree/payment_decorator.rb
+++ b/app/models/spree/payment_decorator.rb
@@ -4,6 +4,8 @@ module Spree
   Payment.class_eval do
     extend Spree::LocalizedNumber
 
+    delegate :line_items, to: :order
+
     has_one :adjustment, as: :source, dependent: :destroy
 
     after_save :ensure_correct_adjustment, :update_order
@@ -28,11 +30,6 @@ module Spree
 
     def adjustment_label
       I18n.t('payment_method_fee')
-    end
-
-    # This is called by the calculator of a payment method
-    def line_items
-      order.line_items
     end
 
     # Pin payments lacks void and credit methods, but it does have refund
--- a/spec/models/spree/payment_spec.rb
+++ b/spec/models/spree/payment_spec.rb
@@ -133,6 +133,28 @@ module Spree
         order.reload.update!
       end
 
+      context "when order-based calculator" do
+        let!(:shop) { create(:enterprise) }
+        let!(:payment_method) { create(:payment_method, calculator: calculator) }
+
+        let!(:calculator) do
+          Spree::Calculator::FlatPercentItemTotal.new(preferred_flat_percent: 10)
+        end
+
+        context "when order complete and inventory tracking enabled" do
+          let!(:order) { create(:completed_order_with_totals, distributor: shop) }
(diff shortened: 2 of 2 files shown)


spec/models/spree/payment_spec.rb
```diff
@@ -133,6 +133,28 @@ module Spree
...
+        let!(:calculator) do
+          Spree::Calculator::FlatPercentItemTotal.new(preferred_flat_percent: 10)
+        end
+
+        context "when order complete and inventory tracking enabled" do
+          let!(:order) { create(:completed_order_with_totals, distributor: shop) }
+          let!(:variant) { order.line_items.first.variant }
+          let!(:inventory_item) { create(:inventory_item, enterprise: shop, variant: variant) }
+
+          it "creates adjustment" do
```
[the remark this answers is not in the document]
kristinalim: Thanks @mkllnk, I should indeed use "pending" when introducing specs for bugs.

app/models/spree/payment_decorator.rb
```diff
@@ -30,15 +32,6 @@ def adjustment_label
...
     end
 
-    # This is called by the calculator of a payment method
-    def line_items
-      if order.complete? && Spree::Config[:track_inventory_levels]
-        order.line_items.select { |li| inventory_units.pluck(:variant_id).include?(li.variant_id) }
-      else
-        order.line_items
-      end
-    end
```
sauloperez: but where is this check performed then? I just want to understand.

kristinalim: @sauloperez Currently, stock is being checked upon `save` of the line item:

- Validate upon save: https://github.com/openfoodfoundation/spree/blob/step-6a/core/app/models/spree/line_item.rb#L15
- Original Spree implementation: [`Spree::LineItem#stock_availability`](https://github.com/openfoodfoundation/spree/blob/step-6a/core/app/models/spree/line_item.rb#L109) and [`Spree::LineItem#sufficient_stock?`](https://github.com/openfoodfoundation/spree/blob/step-6a/core/app/models/spree/line_item.rb#L61)
- OFN implementation: [`Spree::LineItem#stock_availability`](https://github.com/openfoodfoundation/openfoodnetwork/blob/4ed0831e4c540107d5b154e166f2753d2180fb75/app/models/spree/line_item_decorator.rb#L160) and [`Spree::LineItem#sufficient_stock?`](https://github.com/openfoodfoundation/openfoodnetwork/blob/4ed0831e4c540107d5b154e166f2753d2180fb75/app/models/spree/line_item_decorator.rb#L145)

RachL:
@kristinalim when I add a payment after putting the previous one as a void, I see that the transaction fee are added a second time. Is this a normal behaviour?

https://docs.google.com/document/d/1Quk4URoak4T-i1BFE0yOWFztnI9EFkF3xgM73q1O1Ts/edit#

kristinalim:
@RachL Sorry, I was supposed to take a further look. I noticed this with voided cheque payments, but I wasn't sure if this was just for cheque and we were just charging transaction fees for bounced cheques (which makes sense to me). I confirmed that this was the behaviour even before the PR.

I created a new issue here: #3584

RachL:
@kristinalim thank you! I'm moving this to ready to go then.

Issue #3525, opened 2019-02-20
Remove Product Distributions


luisramos0:
### What we should change and why (this is tech debt)
Product Distributions is a dead feature and should be removed from OFN.

### Context
From a slack conversation [here](https://openfoodnetwork.slack.com/archives/C59ADD8F2/p1546958052045600)

### Impact and timeline
Removing this will reduce code complexity and thus OFN's cost of maintenance.

luisramos0:
@mkllnk I just realized you used product distributions in the sample data PR here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3209/files#diff-6b925f809b3a412c614a3e69e37d6a78R89

Can you share what you know about product distributions? Is this a working feature? is it not replaced by OCs?

mkllnk:
I only included it to make the data valid. It's annoying code without any benefit. :-) I would love to see it :fire:.

luisramos0:
I found this comment from Rob, 2 years ago:
https://github.com/openfoodfoundation/openfoodnetwork/blob/6c83f67b86c1f14b003a057b299536909d8c8f1e/app/models/enterprise.rb#L139

I guess that's now :-)

luisramos0:
Another amazing line of code:
https://github.com/openfoodfoundation/openfoodnetwork/blob/6c83f67b86c1f14b003a057b299536909d8c8f1e/spec/models/spree/order_spec.rb#L463

sauloperez:
>  set_feature_toggle :order_cycles, false 

lol, it's clearly time to :fire:

PR #3569, opened 2019-03-01
Delete some more dead overrides and more dead code

luisramos0:
#### What? Why?
Can you believe it? Following the trace of some dead overrides I find 641 lines of smart and dead code scattered around 26 files!
But the question is open for reviewers and testers: is this really dead code? the build is green...

We are not using:
- view spree/checkout (except for /checkout/payment)
- view spree/products

and that makes most of distribution_change_validator dead code.
The change of distributor in the frontoffice is not permitted, if you change shop, the cart is made empty (done in js in ofnChangeHub).
The change of distributor in the backoffice is validated in the order model validator products_available_from_new_distribution.

This is the perfect preparation for #3525

#### What should we test?
green build!
This is basically touching the frontoffice: add to cart, checkout (different payment methods paypal and stripe) and then changing a submitted order in the frontoffice (I didnt know this existed!).
The validation is to make sure everything is working as normal.

#### Release notes
Changelog Category: Removed
Removed old unused code which makes ofn easier and faster to maintain and develop.


lib/open_food_network/distribution_change_validator.rb
```diff
@@ -9,8 +9,10 @@ def can_change_to_distribution?(distributor, order_cycle)
   end
 
   def variants_available_for_distribution(distributor, order_cycle)
-    product_distribution_variants = distributor.andand.product_distribution_variants || []
-    order_cycle_variants = order_cycle.andand.variants_distributed_by(distributor) || []
+    product_distribution_variants = []
```
mkllnk: I actually prefer the old version. Double-assignments are bad style.

luisramos0: I'll keep the new version. I dont like andand, I think it's bad in terms of readability, in the first version it's not clear that the result is [] if distributor is nil, in the new version, it's explicit.
I like double assignments in cases like this, they make code simple, forcing everything into single assignments makes the code unnecessarily complicated. this is avoiding using the readers brain capacity when it is not necessary.
but it's a personal preference...

RachL:
hello @luisramos0 . Is this concerning all OFN payment methods? Because I've tested bogus and stripe. But I have no idea for example how to test pin payments. Should I try and reach the AU team for this or are stripe and bogus enough?

Also, what did you mean by "changing a submitted order in the frontoffice" ?

my notes for future work: https://docs.google.com/document/d/12QDuekEmEAq5bvU7YwHzWI4qJYC7LTDSewpIMSmG8r4/edit#

luisramos0:
I think bogus and stripe is enough for this PR.

"Also, what did you mean by "changing a submitted order in the frontoffice" ?"
:-D exactly!!! I'll show you an old little corner of OFN that you may not know:
Screenshot 2019-03-12 at 16 25 46
select the second option
and then on your account, if you open the order, it will be changable.
the link for you test order is this one: https://staging.katuma.org/orders/R355081463

lib/open_food_network/distribution_change_validator.rb
sauloperez: I don't have a preference for double assignments or `||` but I do not like `andand` at all. To me, it feels like a symptom of unconfident programming which tends to lead to bloat.

RachL:
ah yes thank you @luisramos0 actually I do have tested this before, but I completely forgot that it was something you could turned off!

Issue #3285, opened 2019-01-05
[Spree Upgrade] Data Migration


luisramos0:
As discussed here: https://community.openfoodnetwork.org/t/staging-ofn-v2/1527/5
We need to make sure:
1- the db:migrate from v1 to v2 works correctly
2- the db seed works correctly
3- all v1 data in current live instances is migrated to the v2 data model

## What to test
I install v2 on a live instance running v1 and I get the db in the correct v2 state while the data in the DB has been adapted correctly to v2. We should cover different data tests we need to run for the migration. What are the changes in the data model that we need to adapt/migrate?

luisramos0:
@sauloperez I think you are more clear than I am about what needs to be done here. Can you help write up/clarify what exactly we need to do and test as pat of this task? Thanks!

luisramos0:
To validate the data migrations, we need to load live data into staging, run the migrations and check the result is good.
Discussion here: https://community.openfoodnetwork.org/t/testing-ofn-v2/1545/5

We will need to check all v1 historical data is stored in an appropriate manner in the v2 data model, order adjustments, order shipping_methods, enterprise shipping methods and payment methods. What else will we need to test in terms of data?

sauloperez:
I wouldn't harm to investigate the actual data migration Spree did on existing data, to check things are fine, like from on_hand attributes from their old to their new location, etc.

Then the actual check will be to generate data, migrate it and then use it again. If the data migrations were right we should see no errors.

luisramos0:
I have quite a few things todo but I think this is the highest priority right now.

Tomorrow I'll take a backup for UK live and load it into uk staging. After that I plan to run the v2 migrations and look at the data. At that point we will have a staging server with live like data running in v2. I'll report back here.

luisramos0:
I have spent quite a lot of time on this one...
- We now have a way to sanitize and truncate live data to be used for testing  #3579 
- 2 existing migrations were fixed #3591
- 2 new migrations added (picked up from future spree versions and adapted) #3602

I am currently testing locally with live data. I'll create issues as I find them.

sauloperez:
Are we done with this epic?

luisramos0:
there's 3591 above to be merged.

and I think @kristinalim question on #3591 is very important and we should investigate a little better:
"Do we have a system for validating a certain v1 instance will not lose any data when migrating?"

kristinalim:
[...]

PR #3579, opened 2019-03-06
Add sanitize and truncate data tasks

luisramos0:
#### What? Why?

Related to #3285

This is what I used to clean up the AUS data after copying it from live to staging.

Sanitize: I am happy to use another name for this task if anyone has a better name.

#### What should we test?
You can run the tasks on your database without errors.
Sanitize: no user private data should be available after running the task.
Truncate: no orders, products and order cycles older than 3 months should be in the database.

Note: even after this truncate, AUS staging is still mostly unusable by a super admin because of number of order cycles and products.

#### Release notes
Changelog Category: Added
Added tasks to clean up and truncate data in large databases, so it becomes possible to use it for testing in smaller servers.

This will be used to clean up data loaded from live into staging: this will enable us to test v2 with live data before the release.


lib/tasks/truncate_data.rake
```diff
@@ -0,0 +1,24 @@
+namespace :ofn do
+  namespace :data do
+    desc 'Truncate data'
+    task truncate: :environment do
+      guard_and_warn
```
sauloperez: This implicit dependency will bite us in the ass someday. What about using a more generic and reusable solution like https://github.com/adamcooke/rails-safe-tasks instead of our own?

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+      end
+    end
+
+    def guard_and_warn
+      if Rails.env.production?
+        puts "This task cannot be executed in production"
+        exit
+      end
+
+      message = "This will permanently change DB contents, are you sure you want to proceed? (y/N)"
```
sauloperez: but will `N` actually be the default? I thought you had to use `Question#default` for that

lib/tasks/truncate_data.rake
```diff
@@ -0,0 +1,24 @@
...
+      OrderCycle
+        .where("created_at < ?", 6.months.ago)
+        .destroy_all
+
+      # truncate Products not in an OC and older than 6 months
+      Spree::Product
+        .with_order_cycles_outer
+        .where('o_order_cycles.id IS NULL and spree_products.created_at < (?)', 6.months.ago)
+        .destroy_all
+    end
```
sauloperez: I don't understand why we need this truncation with a 6-month window. Can you provide more details in a code comment maybe?

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+       { first_name: "Helen", last_name: "Mitcham" },
+       { first_name: "Emma", last_name: "Low" },
+       { first_name: "Mandy", last_name: "Trust" }]
+    end
+
+    # We need to disable user coNFirmation emails that would otherwise be triggered
+    def disable_user_confirmation_emails
+      Spree::UserMailer.class_eval do
+        def confirmation_instructions(*) end
+      end
```
sauloperez: We really need to stop abusing class evil. What about disabling ActionMailer deliveries in this rake task? See perform_deliveries in https://guides.rubyonrails.org/v3.2/action_mailer_basics.html#action-mailer-configuration.

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+      paypal_express_checkout.save!
+    end
+
+    def sanitize_stripe_account(stripe_account)
+      stripe_account.stripe_user_id = stripe_account.stripe_publishable_key = nil
+      stripe_account.save!
+    end
+
+    def sanitize_batch(model_class, sanitize_method_name, canned_users = nil)
+      model_class.constantize.all.each_with_index do |object, index|
```
sauloperez: there is no need to constantize the string if we pass a constant, and AFAIK nothing stops us from doing so.

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+        sanitized_user = sanitized_user(canned_users, index) if canned_users
+
+        if sanitized_user
+          puts "Sanitizing #{model_class} #{object.id} with #{sanitized_user[:first_name]}"
```
sauloperez: Didn't Rubocop complain about the use of `puts` instead of `Rails.logger` :thinking: ?

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+          __send__ sanitize_method_name, object, sanitized_user
+        else
+          puts "Sanitizing #{model_class} #{object.id}"
+          __send__ sanitize_method_name, object
```
sauloperez: better to stick to `public_send` here. Didn't Rubocop complain either?

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+
+      # Sanitize payments related entities
+      sanitize_batch("Spree::PaymentMethod", :sanitize_payment_method)
+      sanitize_batch("Spree::CreditCard", :sanitize_credit_card)
+      sanitize_batch("Spree::Payment", :sanitize_payment)
+      sanitize_batch("Spree::PaypalExpressCheckout", :sanitize_paypal_express_checkout)
+      sanitize_batch("StripeAccount", :sanitize_stripe_account)
+
+      # Can't find the model for spree_paypal_accounts table (email and payer_id need to be deleted)
+      #   sanitize_batch("Spree::PaypalAccount", :sanitize_paypal_account, canned_users)
```
sauloperez: it might not be baked by an AR model. I couldn't find anything either

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+      ActiveRecord::Base.connection.execute("delete from spree_paypal_accounts")
+
+      # Delete all preferences that may contain sensitive information
+      ActiveRecord::Base.connection.execute("
+        delete
+        from spree_preferences
+        where key like '%gateway%'
+          OR key like '%billing_integration%'
+          OR key like '%s3%'
+        ")
```
sauloperez: A slightly higher-level method like [delete_all](https://apidock.com/rails/v3.2.13/ActiveRecord/Relation/delete_all) could come in handy here.

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+
+      Spree::LogEntry.destroy_all
```
sauloperez: First time I see this model. I have no idea but if we don't care about any callbacks we can use `delete_all` instead.

lib/tasks/sanitize_data.rake
```diff
@@ -0,0 +1,162 @@
...
+      sanitized_user[:email] = "#{index}_#{sanitized_user[:email]}"
+      sanitized_user
+    end
+
+    def build_canned_users
+      base_canned_users.each do |user|
+        user[:email] = "ofn@example.com"
+        user[:phone] = "0123456789"
+      end
+    end
```
sauloperez: Why do we modify these users at runtime with static data? Wouldn't it be simpler to just append it in `#base_canned_users` below?

lib/tasks/sanitize_data.rake
luisramos0: there was no default, I have added no as default now.

lib/tasks/sanitize_data.rake
luisramos0: done :+1:
thanks!

lib/tasks/sanitize_data.rake
luisramos0: yes, ActionMailer::Base.perform_deliveries = false would do the trick, I ended up switching to update_all that doesnt trigger the emails.

lib/tasks/sanitize_data.rake
luisramos0: yes, I switched.

lib/tasks/sanitize_data.rake
luisramos0: I ended up ditching this method all together and use update_all, so much faster...

lib/tasks/sanitize_data.rake
luisramos0: I made some good efforts in to using AR methods but ended up with sql, so much faster...

lib/tasks/sanitize_data.rake
luisramos0: :+1:

luisramos0: I removed the comment

lib/tasks/truncate_data.rake
luisramos0: I have added a comment at the top of the task

lib/tasks/sanitize_data.rake
luisramos0: I ended up ditching this approach all together.

lib/tasks/sanitize_data.rake
luisramos0: I ended up changing this one. Some others created issues with default scopes...

lib/tasks/truncate_data.rake
luisramos0: yes, that would be better... I think I have spent way too much time on this PR...

lib/tasks/data/truncate_data.rake
```diff
@@ -0,0 +1,75 @@
...
+      #   The sanitize task will work well with the ful range of 300k addresses
+      # sql_delete_from "spree_addresses where
+      #  id not in (select address_id from enterprise_groups) and
+      #  id not in (select address_id from enterprises) and
+      #  id not in (select bill_address_id from spree_orders where bill_address_id is not null) and
+      #  id not in (select ship_address_id from spree_orders where ship_address_id is not null) and
+      #  id not in (select address_id from spree_shipments where address_id is not null) and
+      #  id not in (select address_id from spree_shipments where address_id is not null) and
+      #  id not in (select bill_address_id from customers where bill_address_id is not null) and
+      #  id not in (select ship_address_id from customers where ship_address_id is not null)"
```
sauloperez: I guess we don't need this code anymore, right?

luisramos0: yes, I have just removed it in a new commit.

lib/tasks/data/sanitize_data.rake
```diff
@@ -0,0 +1,56 @@
...
+namespace :ofn do
+  namespace :data do
+    desc 'Sanitize data'
+    task sanitize: :environment do
+      guard_and_warn
+
+      Spree::User.update_all("email = concat(id, '_ofn_user@example.com'),
+                              login = concat(id, '_ofn_user@example.com'),
+                              unconfirmed_email = concat(id, '_ofn_user@example.com')")
+      Spree::Customer.update_all("email = concat(id, '_ofn_customer@example.com'),
```
mkllnk: This line means that customer emails won't be in sync with user emails any more. They should be the same. Maybe we could do something like this:
```rb
Spree::Customer.where("user_id IS NULL").update_all("email = concat(id, '_ofn_customer@example.com')...")
Spree::Customer.where("user_id IS NOT NULL").update_all("email = concat(user_id, '_ofn_user@example.com')...")
```

mkllnk: But in this case we lose all the examples where customer data is not in sync... It's a bit more complicated if we wanted to keep the entropy. So maybe we shouldn't bother unless there is a problem with this.

lib/tasks/data/sanitize_data.rake
```diff
@@ -0,0 +1,56 @@
...
+                                  name = concat('Customer Number ', id)")
+      Spree::Order.update_all("email = concat(id, '_ofn_order@example.com')")
+      Spree::Address.update_all("
+        firstname = concat('Ms. Number', id), lastname = 'Jones',  phone = '01234567890',
+        alternative_phone = '01234567890', address1 = 'Dummy address',
+        address2 = 'Dummy address continuation', city = 'Dummy City', zipcode = '0000',
```
mkllnk: Do you think it would be reasonable to keep the city and zipcode? That would avoid the map from displaying all in one spot.

lib/tasks/data/sanitize_data.rake
```diff
@@ -0,0 +1,56 @@
...
+      # Sanitize payments related entities
+      Spree::PaymentMethod.update_all("name = concat('Dummy Payment Method', id),
+                                       description = name")
+      Spree::CreditCard.update_all("
+        month = 12, year = 2020, start_month = 12, start_year = 2000,
+        cc_type = 'VISA', first_name = 'Dummy', last_name = 'Dummy', last_digits = '2543'")
+      Spree::Payment.update_all("response_code = null, avs_response = null,
+                                 cvv_response_code = null, identifier = null,
+                                 cvv_response_message = null")
+      Spree::PaypalExpressCheckout.update_all("token = null")
```
mkllnk: What about Stripe?

lib/tasks/data/sanitize_data.rake
```diff
@@ -0,0 +1,56 @@
+require 'highline'
+
+namespace :ofn do
+  namespace :data do
+    desc 'Sanitize data'
```
mkllnk: Sanitize can also mean to discard input data which doesn't comply to requirements. So if someone enters an invalid email address into a form. But it's also used for removing classified data. Maybe a native speaker could tell us if this name is good? I'm okay with it.

lib/tasks/data/sanitize_data.rake
luisramos0: it's in the following line: StripeAccount.delete_all
and also included in the Spree::Preference clean up below with the filter "gateway".

lib/tasks/data/sanitize_data.rake
luisramos0: Yeah, you are right. I have just renamed it to anonymize :-)
I think they are both right, actually:
Screenshot 2019-03-20 at 11 51 37

lib/tasks/data/sanitize_data.rake
luisramos0: yes, done.

lib/tasks/data/sanitize_data.rake
luisramos0: I like your suggestion to keep them in sync, I have added it in a new commit. I have tested it locally to validate it won't make the task take 3 days to run :-)

luisramos0: also, I dont understand how Spree::Customer got into here, it's just Customer. I could swear this was tested... well, now it's tested for sure!

lib/tasks/data/sanitize_data.rake
mkllnk: Oh dear, it was clearly too late for me last night.

luisramos0:
ok, code climate is happy now.

sauloperez:
Great work @luisramos0 !

sauloperez:
Moving this to test ready so that we get another pair of eyes on it and @RachL gets familiar with it. I guess you'll need to help her @luisramos0 . Let me know if you agree.

luisramos0:
I have tested this way too many times now 🐰 
Do you guys want to test this or shall we just merge?

luisramos0:
oh, I see your comment. ok I can load live sanitized truncated AUS data to aus staging.

luisramos0:
actually, I will merge this now to master and then merge it to 2-0 and then, from 2-0-stable, use it to load AUS staging.

sauloperez:
Your call!

Issue #3391, opened 2019-01-25
[Spree Upgrade] Shop showing incorrect products (product cache in v2)


sstead:
## Description
On the French staging server, which now holds the Spree Upgrade I created an order cycle for a hub shop. The OC contains Apple products. But when I get to the shop I see a list of random products that this shop doesn't even have permission to sell. I'm not sure where this product data is coming from - is it seed data for the new Staging server? I can't see these products in the backend as Super Admin....?

Either way I can't place an order, and it's not possible to meaningfully test checkout if I can't put the right products in the shop.

Update: I created a second OC and that OC correctly shows the right items, however I can't checkout in that OC, or others.

## Expected Behavior
The products in the shop should reflect the products in the Order Cycle.

## Steps to Reproduce
1. Open OC on French staging and see if the shop holds the correct items

## Animated Gif/Screenshot
This shows OC#1 - full of incorrect products
https://www.useloom.com/share/9e3ea68323764329a9ff0e9d207a6869

This shows OC#2 which correctly shows the Apple related products
https://staging.openfoodfrance.org/winery-district-food-hub/shop

## Severity
Because this bug is only on Staging, related to the Spree upgrade, I'll assign it s4
bug-s4: it's annoying, but you can use it

## Your Environment
https://staging.openfoodfrance.org/winery-district-food-hub/shop
Spree testing, desktop, Chrome, PC

luisramos0:
It's a cache error Admin - Configuration - Cache Settings (Mise en Cache):
![screenshot 2019-01-28 at 15 11 43](https://user-images.githubusercontent.com/1640378/51845226-0bbe1900-230f-11e9-9187-9ad0d5e6f21c.png)

luisramos0:
I spent an hour or so trying to find the problem but I couldnt find the cause yet. It's a cache issue for sure, and #3392 looks like it's also a cache issue.

luisramos0:
I now understand well the products cache code but it is not easy ([cache invalidation is one of the 2 hard problems in computer science](https://i.imgur.com/lJpH5nX.jpg)) because different events will trigger the cache invalidation (aka refresh_cache). After 6 hours or so trying a few things I'll leave this one for the next dev who wants to have a look at it.

Meanwhile I built a little feature to help debug the product cache: see #3404

sauloperez:
I'd like to take a look at it but I first want to get the green build

tschumilas:
[...]

PR #3581, opened 2019-03-06
Cache stylistic improvements

sauloperez:
#### What? Why?

This applies a bunch of small refactors that improve a little bit part of the product cache implementation by making it more readable and maintainable.

It also includes a bigger refactor I did to extract `DistributedValidProducts` out of the order cycle model. It brings cohesion to the model and more importantly it increases my confidence in the affected code, whose behavior took me long to figure out. 

It is the outcome of the in-depth exploration I did for #3391.

#### What should we test?

An order cycle with an obsolete master in distribution: If a product without variants is added to an order cycle, and then some variants are added to that product, but not to the order cycle, then the master variant should not available for customers to purchase.

Also, the `admin/cache_settings/edit` page should keep working. To be extra sure a dev should shut down delayed job in staging to see the diff.

#### Release notes

Refactor valid distributed products in an order cycle, which is used by `RefreshProductsCacheJob`.

Changelog Category: Changed

### DIFF
--- a/app/models/distributed_valid_products.rb
+++ b/app/models/distributed_valid_products.rb
@@ -1,0 +1,35 @@
+# Finds valid products distributed by a particular distributor in an order cycle
+class DistributedValidProducts
+  def initialize(order_cycle, distributor)
+    @order_cycle = order_cycle
+    @distributor = distributor
+  end
+
+  def all
+    variants = order_cycle.variants_distributed_by(distributor)
+    products = variants.map(&:product).uniq
+
+    valid_products = products.reject do |product|
+      product_has_only_obsolete_master_in_distribution?(product, variants)
+    end
+    product_ids = valid_products.map(&:id)
+
+    Spree::Product.where(id: product_ids)
+  end
+
+  private
+
+  attr_reader :order_cycle, :distributor
+
+  # If a product without variants is added to an order cycle, and then some variants are added
+  # to that product, but not the order cycle, then the master variant should not available for
+  # customers to purchase.
+  #
+  # This method is used by #valid_products_distributed_by to filter out such products so that
+  # the customer cannot purchase them.
+  def product_has_only_obsolete_master_in_distribution?(product, distributed_variants)
+    product.has_variants? &&
+      distributed_variants.include?(product.master) &&
+      (product.variants & distributed_variants).empty?
+  end
+end
--- a/app/models/distributed_valid_products.rb
+++ b/app/models/distributed_valid_products.rb
@@ -1,11 +1,20 @@
 # Finds valid products distributed by a particular distributor in an order cycle
+#
+# If a product without variants is added to an order cycle, and then some
+# variants are added to that product, but not the order cycle, then the master
+# variant should not available for customers to purchase. This class filters
+# out such products so that the customer cannot purchase them.
 class DistributedValidProducts
   def initialize(order_cycle, distributor)
     @order_cycle = order_cycle
     @distributor = distributor
   end
 
-  def all
+  # Returns an ActiveRecord relation without invalid products. Check
+  # #valid_products_distributed_by for details
+  #
+  # @return [ActiveRecord::Relation<Spree::Product>]
+  def relation
     variants = order_cycle.variants_distributed_by(distributor)
     products = variants.map(&:product).uniq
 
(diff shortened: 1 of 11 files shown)


app/models/distributed_valid_products.rb
```diff
@@ -0,0 +1,41 @@
+# Finds valid products distributed by a particular distributor in an order cycle
+#
+# If a product without variants is added to an order cycle, and then some
```
luisramos0: What do you mean by "product without variants"?

We have product.ensure_standard_variant that keeps products with variants at all the times.

This query returns no results in all AU, UK and FR live.
```
select * from spree_products
where not exists (select 1
                              from spree_variants
                              where product_id = spree_products.id
                                       and is_master = 'f');
```

app/models/distributed_valid_products.rb
```diff
@@ -0,0 +1,41 @@
...
+# variants are added to that product, but not the order cycle, then the master
+# variant should not available for customers to purchase. This class filters
+# out such products so that the customer cannot purchase them.
+class DistributedValidProducts
+  def initialize(order_cycle, distributor)
```
luisramos0: IMO, this should be in app/services, it's a service and it's doing one specific thing really well, finding the order cycle distributed products. I think it deserves Order Cycle in the name.   Order_Cycle_Distributed_Products?

luisramos0: OrderCycleProductsFinder.call
FindOrderCycleProducts.call
please cast you vote here ❤️  https://community.openfoodnetwork.org/t/how-to-name-a-new-service-in-ofn/1530/5

app/models/distributed_valid_products.rb
sauloperez: I just moved the comment and honestly, even some the test cases themselves were a bit dubious. I'll review it and make any necessary changes. Thanks for the heads-up!

app/models/distributed_valid_products.rb
sauloperez: lol. I first started adding it into models and didn't put any thought on it afterward. Do you think is worth keeping the notion of validation in its name? It felt important to me to know that we perform a sort of validation in here but what do you guys think?

luisramos0: ehehe, I'd not include Valid because the validation isn't valid 🤣 
Because we have the standard variant, master variants should never be added to OCs or exchanges, so this validation should go... doesnt need to be now, but it should go at some point.

app/models/distributed_valid_products.rb
luisramos0: I think this deserves a tech debt issue, to clarify the comment but also the fact that master variants are added to OCs, and why would that ever happen. see my comment below.

app/models/distributed_valid_products.rb
sauloperez: done

app/services/order_cycle_distributed_products.rb
```diff
@@ -0,0 +1,41 @@
...
+    @distributor = distributor
+  end
+
+  # Returns an ActiveRecord relation without invalid products. Check
+  # #valid_products_distributed_by for details
+  #
+  # @return [ActiveRecord::Relation<Spree::Product>]
+  def relation
+    variants = order_cycle.variants_distributed_by(distributor)
+    products = variants.map(&:product).uniq
```
mkllnk: This loads every product individually. I think this would be much more efficient:
```suggestion
    products = Product.where(id: variants.select(:product_id)).uniq
```

That combines it all into one query.
```sql
SELECT DISTINCT "spree_products".*
 FROM "spree_products"
  WHERE "spree_products"."id" IN (SELECT  product_id FROM "spree_variants" ...)
```

app/services/order_cycle_distributed_products.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    valid_products = products.reject do |product|
+      product_has_only_obsolete_master_in_distribution?(product, variants)
+    end
+    product_ids = valid_products.map(&:id)
+
+    Spree::Product.where(id: product_ids)
```
mkllnk: Have you tried using `valid_products` here? ActiveRecord maps to ids for us.

app/services/order_cycle_distributed_products.rb
sauloperez: This inevitably moves the N+1 to the block below, to the variants table, which would then require us to use [includes](https://apidock.com/rails/ActiveRecord/QueryMethods/includes). I think it's wiser to stop here because #3596 might turn this change pointless. We'll see there.

RachL:
Ok so this is working good, so we just need to test killing delayed_job. This is scheduled for tomorrow afternoon.

testing notes: https://docs.google.com/document/d/1jwTcr0q7pwuVSE0yYoHGfn3nUo-1lUZ8l8wPkWWBoAw/edit#

RachL:
Stoping delayed job and test this part was sucessfull! I've updated the notes, this is ready to go!

PR #3583, opened 2019-03-06
Handle deleted order cycle when refreshing cache

### LINKED ISSUE #3571: RefreshProductsCacheJob gets run multiple times because of failure when OC deleted

This should not fail. When a background job fails, it gets retried 25 times (`Delayed::Worker::MAX_ATTEMPTS`). This is unnecessary and a waste of time.

### Context

Noticed this in the `delayed_job.log` in OFN CA:

```
2019-02-20T18:17:49-0500: [Worker(delayed_job.0 host:openfoodnetwork.ca pid:29429)] Job RefreshProductsCacheJob (id=6898) FAILED (22 prior attempts) with ActiveRecord::RecordNotFound: Couldn't find OrderCycle with id=169
```

sauloperez:
#### What? Why?

Closes #3571 

When refreshing the cache of an order cycle that got deleted there is not need to retry the job. Essentially, it does not constitute a failure scenario so returning true does the job.

#### What should we test?

It's worth testing this in a staging environment. Check that after changing an order cycle products and then deleting it does not retry cache refreshes. I'll help to do so.

#### Release notes

Cache refreshes of a deleted order cycle no longer get retried.

Changelog Category: Fixed

### DIFF
--- a/app/jobs/refresh_products_cache_job.rb
+++ b/app/jobs/refresh_products_cache_job.rb
@@ -3,6 +3,8 @@ require 'open_food_network/products_renderer'
 RefreshProductsCacheJob = Struct.new(:distributor_id, :order_cycle_id) do
   def perform
     Rails.cache.write(key, products_json)
+  rescue ActiveRecord::RecordNotFound
+    true
   end
 
   private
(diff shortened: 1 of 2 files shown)


RachL:
Here are the steps we followed with @sauloperez  :

https://docs.google.com/document/d/1C9XteJyX3TKjLEc2ZRBxpcrU9ESG2kz_CNohWbszPxU/edit#

Issue #2786, opened 2018-09-29
[Spree Upgrade] Fix specs related to Subscriptions (pending specs)


luisramos0:
Subscription_mailer_spec is badly broken with 21 broken specs.
These specs were made pending but need to be fixed at some point.

spec/jobs/subscription_placement_job_spec.rb is badly broken.
These specs were made pending but need to be fixed at some point.

Moved all specs in subscription_form_spec to pending
Moved all specs in spec/features/admin/subscriptions_spec to pending
Moved all specs in subscription_confirm_job_spec to pending

See #2787 , #2795 and #2902

kristinalim:
Current status in #3589 branch:

* `spec/mailers/subscription_mailer_spec.rb` - Done
* `spec/jobs/subscription_placement_job_spec.rb` - Fixed in #3589 
* `spec/services/subscription_form_spec.rb` - Done
* `spec/features/admin/subscriptions_spec` - Done
* `spec/jobs/subscription_confirm_job_spec.rb` - Done

luisramos0:
Amazing 🎉

PR #3589, opened 2019-03-07
2788 [Subscriptions] Fix address issues when syncing subscriptions and orders

kristinalim:
Fix shipping and billing address issues when syncing subscriptions and orders.

#### What? Why?

This fixes remaining issues in #2788, related to how changes to subscriptions' bill address, ship address, and shipping method are carried over to the orders.

This PR includes changes in PR #3560, and cherry-picks commits for v1 in #3605, so should not be merged until those have been approved and merged to their respective base branches.

It turned out that many spec assertions were passing only by accident in v1. See more info here: #3605 The specs had to be fixed in v1 so we could make sure to preserve the behaviour in v2. The fix specific to v2 is only in these commits:

- https://github.com/openfoodfoundation/openfoodnetwork/pull/3589/commits/d1c3bbf3edc2ac222c22a8883c04f79db747b036
- https://github.com/openfoodfoundation/openfoodnetwork/pull/3589/commits/e98de7e6821e2452807075d39c7369b3ae109f91
- https://github.com/openfoodfoundation/openfoodnetwork/pull/3589/commits/8c7e176fa9dad9baac1cbae9133ca7ec028be67b
- https://github.com/openfoodfoundation/openfoodnetwork/pull/3589/commits/eb1961c5f3a2627d119dea7685e95f1d0002c7d6

#### What should we test?

1. Create a subscription with no end date for an open OC. Set pickup as the shipping method, but set billing details that is different from the shipping details.
2. "Edit" the subscription order generated. This will create the order. The order ship address should be: bill address name and phone number, and shop address.
4. Edit the subscription bill address. This should affect the bill and ship address of the order.
5. Change the order bill address.
6. Edit the subscription bill address again. This should not affect the order bill and ship address.
7. Change back order bill address to be the same as subscription bill address.
7. Switch subscription to delivery.
8. Edit the subscription bill address again. Check that the order bill address is updated.
9. Change the order bill address.
10. Edit the subscription bill address again. The order bill addresses should not be updated.

#### Dependencies

- #3560 
- #3605 
- #3640

### DIFF
--- a/app/models/order_updater.rb
+++ b/app/models/order_updater.rb
@@ -27,7 +27,7 @@ class OrderUpdater < SimpleDelegator
   def shipping_address_from_distributor
     return if order.shipping_method.blank? || order.shipping_method.require_ship_address
 
-    order.ship_address = order.__send__(:address_from_distributor)
+    order.ship_address = order.address_from_distributor
   end
 
   private
--- a/app/models/order_updater.rb
+++ b/app/models/order_updater.rb
@@ -21,6 +21,15 @@ class OrderUpdater < SimpleDelegator
     shipping_address_from_distributor
   end
 
+  # Sets the distributor's address as shipping address of the order for those
+  # shipments using a shipping method that doesn't require address, such us
+  # a pickup.
+  def shipping_address_from_distributor
+    return if order.shipping_method.blank? || order.shipping_method.require_ship_address
+
+    order.ship_address = order.__send__(:address_from_distributor)
+  end
+
   private
 
   def infer_payment_state
@@ -78,14 +87,5 @@ class OrderUpdater < SimpleDelegator
 
   def failed_payments?
     payments.present? && payments.valid.empty?
-  end
-
-  # Sets the distributor's address as shipping address of the order for those
-  # shipments using a shipping method that doesn't require address, such us
-  # a pickup.
-  def shipping_address_from_distributor
-    return if order.shipping_method.blank? || order.shipping_method.require_ship_address
-
-    order.ship_address = order.__send__(:address_from_distributor)
   end
 end
--- a/app/models/order_updater.rb
+++ b/app/models/order_updater.rb
@@ -84,11 +84,8 @@ class OrderUpdater < SimpleDelegator
   # shipments using a shipping method that doesn't require address, such us
   # a pickup.
   def shipping_address_from_distributor
-    shipments.each do |shipment|
-      shipping_method = shipment.shipping_method
-      next if shipping_method.require_ship_address
+    return if order.shipping_method.blank? || order.shipping_method.require_ship_address
 
-      order.ship_address = order.distributor.address
-    end
+    order.ship_address = order.__send__(:address_from_distributor)
   end
 end
--- a/app/services/order_syncer.rb
+++ b/app/services/order_syncer.rb
@@ -38,7 +38,7 @@ class OrderSyncer
     # are ignored or not.
     pickup_to_delivery = force_ship_address_required?(order)
     if !pickup_to_delivery || order.shipment.present?
-      update_ship_address_for(order) if (ship_address.changes.keys & relevant_address_attrs).any?
+      save_ship_address_in_order(order) if (ship_address.changes.keys & relevant_address_attrs).any?
     end
     if !pickup_to_delivery || order.shipment.blank?
       order.updater.__send__(:shipping_address_from_distributor)
@@ -57,11 +57,6 @@ class OrderSyncer
       return order_update_issues.add(order, I18n.t('bill_address'))
     end
     order.bill_address.update_attributes(bill_address.attributes.slice(*relevant_address_attrs))
-  end
-
-  def update_ship_address_for(order)
-    return unless ship_address_updatable?(order)
-    order.ship_address.update_attributes(ship_address.attributes.slice(*relevant_address_attrs))
   end
 
   def update_payment_for(order)
@@ -114,6 +109,11 @@ class OrderSyncer
     relevant_address_attrs.all? do |attr|
       order.ship_address[attr] == distributor_address[attr]
     end
+  end
+
+  def save_ship_address_in_order(order)
+    return unless ship_address_updatable?(order)
+    order.ship_address.update_attributes(ship_address.attributes.slice(*relevant_address_attrs))
   end
 
   def pending_shipment_with?(order, shipping_method_id)
--- a/app/services/order_syncer.rb
+++ b/app/services/order_syncer.rb
@@ -37,10 +37,10 @@ class OrderSyncer
     # switching from pick-up to delivery affects whether simultaneous changes to shipping address
     # are ignored or not.
     pickup_to_delivery = force_ship_address_required?(order)
-    if (ship_address.changes.keys & relevant_address_attrs).any?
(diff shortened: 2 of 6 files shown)


app/models/order_updater.rb
```diff
@@ -84,11 +84,9 @@ def failed_payments?
   # shipments using a shipping method that doesn't require address, such us
   # a pickup.
   def shipping_address_from_distributor
-    shipments.each do |shipment|
-      shipping_method = shipment.shipping_method
-      next if shipping_method.require_ship_address
+    return if order.shipment.andand.shipping_method.blank?
+    return if order.shipment.shipping_method.require_ship_address
```
luisramos0: I was just discussing the use of andand in another PR with @mkllnk 
maybe:
```ruby
    return if order.shipment.blank?
    return if order.shipment.shipping_method.blank?
    return if order.shipment.shipping_method.require_ship_address
```
or simply?
```ruby
    return if order.shipment.andand.shipping_method.andand.require_ship_address
```

mkllnk: Or we implement the law of Demeter and then write:
```rb
return if order.require_ship_address?
```

:smile:

kristinalim:
I've rebased this upon #3640 which already includes the changes in #3560 and #3605. So now, this is blocked only by #3640. When that is merged into `2-0-stable`, this should be rebased again.

luisramos0:
fyi #3640 has been merged

kristinalim:
Rebased and green, @luisramos0. Thanks.

app/services/order_syncer.rb
```diff
@@ -31,8 +31,19 @@ def sync!
...
+    # switching from pick-up to delivery affects whether simultaneous changes to shipping address
+    # are ignored or not.
+    pickup_to_delivery = force_ship_address_required?(order)
+    if (ship_address.changes.keys & relevant_address_attrs).any?
+      update_ship_address_for(order) unless pickup_to_delivery && order.shipment.blank?
+    end
+    unless pickup_to_delivery && order.shipment.present?
+      order.updater.__send__(:shipping_address_from_distributor)
+    end
+
```
luisramos0: this is a bit brain damaging, isnt it? :-)

I didnt manage to improve it much but I think this looks a bit easier to follow:
```ruby
    pickup_to_delivery = force_ship_address_required?(order)
    if !pickup_to_delivery || order.shipment.present?
        update_ship_address_for(order) if (ship_address.changes.keys & relevant_address_attrs).any?
    end
    if !pickup_to_delivery || order.shipment.blank?
      order.updater.__send__(:shipping_address_from_distributor)
    end
```

luisramos0: I'd extract this block above including the comment to a separate method, maybe into update_ship_address_for.

app/models/order_updater.rb
```diff
@@ -84,11 +84,8 @@ def failed_payments?
...
   # a pickup.
   def shipping_address_from_distributor
-    shipments.each do |shipment|
-      shipping_method = shipment.shipping_method
-      next if shipping_method.require_ship_address
+    return if order.shipping_method.blank? || order.shipping_method.require_ship_address
 
-      order.ship_address = order.distributor.address
-    end
+    order.ship_address = order.__send__(:address_from_distributor)
```
mkllnk: Shouldn't we make that method public if we use it?

app/services/order_syncer.rb
mkllnk: I see that making this method public would mean touching another class but I think it's worth it.

app/models/order_updater.rb
luisramos0: yeah, address_from_distributor is also a candidate for a nice little service that could also help below in shipping_address_from_distributor

app/services/order_syncer.rb
kristinalim: > this is a bit brain damaging, isnt it? :-)

Yes, I had a difficult time following code paths for this. :sweat: 

Thanks @luisramos0 and @mkllnk. I agree with your suggestions and have appended the commits. :slightly_smiling_face:

app/models/order_updater.rb
kristinalim: I've made this method public as well. :slightly_smiling_face:

luisramos0:
I think this is the last subs v2 PR and it closes #2786 as well, correct @kristinalim ?

RachL:
@kristinalim so far in my tests my order bill address never gets updated when I edit the subscription bill address.
I'm using enterprise 2 and I'm logged in as super admin all along on AU staging.

Maybe I don't look at the right place?

For instance:
> "Edit" the subscription order generated. This will create the order. The order ship address should be: bill address name and phone number, and shop address.

What do you mean by _This will create the order_ ? I used the order cycle closing time in order to create the order.

Also where do you exactly see the order ship/bill address?

I went to the "customer details" of the edit order submenu. But I guess it should show up some place else but I couldn't find where.... the invoice is showing the shipping address of my customer... Sorry if this is obvious :(

luisramos0:
I am unstaging this one from AUS to put the new test data there. this one can go next.

luisramos0:
this is now back to aus staging.

kristinalim:
Apologies, my comment on this is really late.

> This will create the order.

@RachL Immediately after creating a subscription, there are "orders" (proxy orders) listed under the subscription, but the real orders are not actually created until a background task that generates these is run.

I checked just now and see that this is actually only 5 minutes (I don't have this running automatically in my local machine), so you can just wait 5 minutes too. If you don't want to wait, clicking the link to "edit the order" would create the order. The order needs to exist for the rest of the testing steps.

> Also where do you exactly see the order ship/bill address?
>
> I went to the "customer details" of the edit order submenu. But I guess it should show up some place else but I couldn't find where....

This screen recording shows where the subscription order's addresses should show up, @RachL. Are we looking at different places?

20190502015400-subscription-order-addresses

I went through all the testing steps in my dev server and don't see a problem. I will also check the data you are using @RachL. Thank you!

RachL:
@kristinalim thank you for the tip! I didn't know the shortcut, but I was looking at the right place. It didn't worked last time, maybe something with staging it because now I could see each step more clearly!

However 2 things that did not worked:
- *Edit the subscription bill address. This should affect the bill and ship address of the order.* this only affected the bill address. Not the shipping address. Why should it update the ship address as well?
- The email title was not updated but I know there is a PR somewhere for this
- There is a translation key also missing here (but maybe not due to this PR?):
image


Updated testing notes:
https://docs.google.com/document/d/1DNbA0L2dLpmYLZrEkQx9GKI044bAgEcZDsFweYpUXPo/edit#

kristinalim:
> Edit the subscription bill address. This should affect the bill and ship address of the order. this only affected the bill address. Not the shipping address. Why should it update the ship address as well?

Sorry, my bad, Rachel! I was actually editing the **non-address** part of the subscription billing **details** (not **address**) here. So if you edit the _subscription's_ billing name or phone number, it should be copied over to the orders' billing and shipping details. The address part would not be copied to the orders' shipping details because the order is pickup and would still use the hub's address.

RachL:
Thank you @kristinalim as https://github.com/openfoodfoundation/openfoodnetwork/issues/3782 is created I'm moving this PR foward!

Issue #3358, opened 2019-01-18
[Flaky Spec] spec/features/admin/bulk_order_management_spec.rb:322,436


luisramos0:
## Description
This spec seems be unreliable: https://github.com/openfoodfoundation/openfoodnetwork/pull/3354#issuecomment-455585480

## Context
Our automated test fail regularly due to unreliable specs. If you encounter this in your build then comment on this issue or add :+1: 

Error:

```
Failures:

  1) 
  As an Administrator
  I want to be able to manage orders in bulk
 using page controls using drop down seletors distributor filter displays all line items when 'All' is selected from distributor filter
     Failure/Error: expect(page).to have_no_selector "tr#li_#{li2.id}"
       expected #has_no_selector?("tr#li_34") to return true, got false
     # ./spec/features/admin/bulk_order_management_spec.rb:322:in `block (5 levels) in <top (required)>'
```

kristinalim:
This unified diff [3358-intermittent_spec_failures.patch.txt](https://github.com/openfoodfoundation/openfoodnetwork/files/2791191/3358-intermittent_spec_failures.patch.txt) that adds a delay at the right place reproduces these errors consistently for me:

```sh
git apply 3358-intermittent_spec_failures.patch.txt
bundle exec rspec --tag focus
```

luisramos0:
@kristinalim after merging #3416 the master build failed on this spec, do you know why?
I'll re-open this issue.

sauloperez:
It seems like there is no particular spec failing but all of them in `bulk_order_management_spec.rb`. From the last 5 merges to master, 3 of them failed on lines 614, 51 and 588 :see_no_evil:

Matt-Yorkley:
:sob:

sigmundpetersen:
#3576 is the same type of failure, should probably be fixed in same PR

mkllnk:
Closed by https://github.com/openfoodfoundation/openfoodnetwork/pull/3590.

PR #3590, opened 2019-03-08
3576 Fix bulk order management race condition

### LINKED ISSUE #3576: [Flaky Spec] ./spec/features/admin/bulk_order_management_spec.rb:588
## Description

This spec is unreliable:
./spec/features/admin/bulk_order_management_spec.rb:588

https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/master/builds/523

## Context

Our Semaphore builds fail regularly because of these unreliable tests. If you encounter this in your build, please comment on this issue or add 👍

Error:
```

  As an Administrator
 using page controls bulk action controls when a filter has been applied only applies the delete action to filteredLineItems
     Failure/Error: expect(page).to have_no_selector "tr#li_#{li1.id}"
       expected #has_no_selector?("tr#li_79") to return true, got false
     # ./spec/features/admin/bulk_order_management_spec.rb:598:in `block (5 levels) in <top (required)>'

```

mkllnk:
#### What? Why?

Fixes #3576.

We have a very flaky spec that causes half of our build to fail. I think I fixed it now. The main problem was a method returning an array instead of a query object with a promise. Without the promise the page didn't wait for the request to complete.

#### What should we test?

Bulk order management page should show the "loading" message correctly and only show "no records found" if no records are listed. There is a change in the date filters as well.

#### Release notes

Fixed: Improved reliability of our automated tests.

Changelog Category: Fixed


app/assets/stylesheets/admin/angular.css.scss
```diff
@@ -0,0 +1,4 @@
+// https://docs.angularjs.org/api/ng/directive/ngCloak
+[ng\:cloak], [ng-cloak], [data-ng-cloak], [x-ng-cloak], .ng-cloak, .x-ng-cloak {
+  display: none !important;
+}
```
luisramos0: this is not pretty but really nice that we are hiding the "no order found" message!
do we need to keep this file name and location? it would be better to name the file angula-ng-cloak.css.scss and move it under /components. because it has the !important I think this will also work, right?

app/views/spree/admin/orders/bulk_management.html.haml
```diff
@@ -106,11 +106,11 @@
     %h1
       = t("admin.orders.bulk_management.loading")
 
-  %div{ :class => "sixteen columns alpha", 'ng-show' => '!RequestMonitor.loading && filteredLineItems.length == 0'}
+  %div{ class: "sixteen columns alpha", ng: {show: '!RequestMonitor.loading && filteredLineItems.length == 0', cloak: true}}
     %h1#no_results
       = t("admin.orders.bulk_management.no_results")
 
-  .margin-bottom-50{ 'ng-hide' => 'RequestMonitor.loading || filteredLineItems.length == 0' }
+  .margin-bottom-50{ ng: {hide: 'RequestMonitor.loading || filteredLineItems.length == 0', cloak: true } }
```
luisramos0: I think there's a space missing between { and hide
same for show in line 109 and repeat in line 160

app/assets/stylesheets/admin/angular.css.scss
mkllnk: I just replicated what we have in Darkswarm. But I agree that it's better in components. I will move it.

mkllnk:
@luisramos0 I think you are right about testing. I also attended to the style issues. Ready for review again.

luisramos0:
oh, we should update the "What should we test?" section.
Something like "Bulk order management page should show the "loading" message correctly and only show "no records found" if no records are listed"

mkllnk:
So great to have specs! I didn't know that orders were returned within an object instead of as an array. I amended my last commit to just provide the needed promise. I hope it's all good now. :crossed_fingers: 

Ready for review again.

sigmundpetersen:
No Bulk Order failures in 10 builds 🎉

sauloperez:
Tests are passing so this should be merged right away

luisramos0:
we can merge and test afterwards if you want. but i vote for a testing round. It's a tricky area this page loading business...

app/assets/javascripts/admin/resources/services/orders.js.coffee
```diff
@@ -9,7 +9,7 @@ angular.module("admin.resources").factory 'Orders', ($q, OrderResource, RequestM
       request = OrderResource.index params, (data) =>
         @load(data)
         (callback || angular.noop)(data)
-      RequestMonitor.load(request.$promise)
```
kristinalim: Is it really okay to remove this line? It doesn't seem necessary to remove, and many pages including the bulk order management use `RequestMonitor`.

kristinalim:
I agree with @luisramos0 about testing. :slightly_smiling_face:

app/assets/javascripts/admin/resources/services/orders.js.coffee
mkllnk: That call was redundant. The only call of this method is in the line items controller and it uses the RequestMonitor already:
https://github.com/openfoodfoundation/openfoodnetwork/blob/d129665d439671d82eeba89f7cb8c80e4eb0920a/app/assets/javascripts/admin/line_items/controllers/line_items_controller.js.coffee#L32

RachL:
This is ready to go! https://docs.google.com/document/d/1OPjo1VwyfOKawPymCzPIzt65-Jc6Q9kdDOjhum26u-g/edit#

luisramos0:
@mkllnk is this also closing #3358 ?

mkllnk:
Yes, pretty much all bulk order management specs where affected by this bug.

PR #3591, opened 2019-03-08
[Spree Upgrade] Fix Migrations to 2.0

luisramos0:
#### What? Why?

Related to #3285 

After testing db:migrate on live data from the AUS environment, I found a few problems, this PR fixes them:
- migrate tax categories to line items - bringing a fix from spree 2.2
- unique index on shipment.order_id - deleting shipments if order has more than one

#### What should we test?
This has been tested with live data now.

### DIFF
--- a/db/migrate/20180426145669_migrate_tax_categories_to_line_items.spree.rb
+++ b/db/migrate/20180426145669_migrate_tax_categories_to_line_items.spree.rb
@@ -1,8 +1,12 @@
 # This migration comes from spree (originally 20130802022321)
 class MigrateTaxCategoriesToLineItems < ActiveRecord::Migration
   def change
-  	Spree::LineItem.includes(:variant => { :product => :tax_category }).find_in_batches do |line_items|
-  	  line_items.each do |line_item|
+    Spree::LineItem.includes(:variant => { :product => :tax_category }).find_in_batches do |line_items|
+      line_items.each do |line_item|
+        next if line_item.variant.nil?
+        next if line_item.variant.product.nil?
+        next if line_item.product.nil?
+        next unless line_item.product.tax_category.present?
         line_item.update_column(:tax_category_id, line_item.product.tax_category.id)
       end
     end
--- a/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
+++ b/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
@@ -1,6 +1,23 @@
+# This migration is an OFN specific migration that enforces an order to have a single shipment at all times
 class AddUniquenessOfOrderIdToSpreeShipments < ActiveRecord::Migration
   def change
+    destroy_all_but_latest_shipments
+
     remove_index :spree_shipments, :order_id
     add_index :spree_shipments, :order_id, unique: true
+  end
+
+  private
+
+  # Destroy all but the latest shipment in each order
+  def destroy_all_but_latest_shipments
+    latest_shipments = Spree::Shipment.
+      select("order_id, MAX(updated_at) updated_at").
+      group(:order_id).
+      having("count(*) > 1")
+
+    Spree::Shipment.
+      joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id=latest_shipments.order_id AND spree_shipments.updated_at != latest_shipments.updated_at").
+      destroy_all
   end
 end
--- a/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
+++ b/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
@@ -16,8 +16,67 @@ class AddUniquenessOfOrderIdToSpreeShipments < ActiveRecord::Migration
       group(:order_id).
       having("count(*) > 1")
 
+    all_duplicated_shipments = Spree::Shipment.
+      joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id = latest_shipments.order_id")
+    backup_to_csv(all_duplicated_shipments)
+
     Spree::Shipment.
       joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id=latest_shipments.order_id AND spree_shipments.updated_at != latest_shipments.updated_at").
       destroy_all
+  end
+
+  def backup_to_csv(shipments)
+    CSV.open(csv_path, "w") do |csv|
+      csv << csv_header_row
+
+      shipments.each do |shipment|
+        csv << shipment_csv_row(shipment)
+      end
+    end
+  end
+
+  def csv_header_row
+    %w(
+      id
+      shipment.tracking
+      number
+      cost
+      shipped_at
+      order_id
+      shipping_method_id
+      address_id
+      created_at
+      updated_at
+      state
+    )
+  end
+
+  def shipment_csv_row(shipment)
+    [
+      shipment.id,
+      shipment.tracking,
+      shipment.number,
+      shipment.cost,
+      shipment.shipped_at,
+      shipment.order_id,
+      shipment.shipping_method_id,
+      shipment.address_id,
+      shipment.created_at,
+      shipment.updated_at,
+      shipment.state
+    ]
+  end
+
+  def csv_path
+    ensure_reports_path_exists
+    reports_path.join("duplicated_shipments_backup.csv")
+  end
+
+  def reports_path
+    Rails.root.join("reports")
+  end
+
+  def ensure_reports_path_exists
+    Dir.mkdir(reports_path) unless File.exist?(reports_path)
   end
 end
--- a/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
+++ b/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
@@ -20,9 +20,16 @@ class AddUniquenessOfOrderIdToSpreeShipments < ActiveRecord::Migration
       joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id = latest_shipments.order_id")
     backup_to_csv(all_duplicated_shipments)
 
-    Spree::Shipment.
-      joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id=latest_shipments.order_id AND spree_shipments.updated_at != latest_shipments.updated_at").
-      destroy_all
+    shipments_to_delete = Spree::Shipment.
+      joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id = latest_shipments.order_id AND spree_shipments.updated_at != latest_shipments.updated_at")
+    remove_association_to_adjustments(shipments_to_delete)
+    shipments_to_delete.destroy_all
+  end
+
+  def remove_association_to_adjustments(shipments)
+    Spree::Adjustment.
+      joins("INNER JOIN (#{shipments.to_sql}) shipments_to_delete ON shipments_to_delete.id = spree_adjustments.source_id and spree_adjustments.source_type = 'Spree::Shipment'").
+      update_all(source_id: nil, source_type: nil, originator_id: nil, originator_type: nil, mandatory: nil)
   end
 
   def backup_to_csv(shipments)
@@ -38,16 +45,16 @@ class AddUniquenessOfOrderIdToSpreeShipments < ActiveRecord::Migration
   def csv_header_row
     %w(
       id
-      shipment.tracking
+      tracking
       number
+      order_number
+      shipping_method_name
       cost
+      state
       shipped_at
-      order_id
-      shipping_method_id
-      address_id
       created_at
       updated_at
-      state
+      address_json
     )
   end
 
@@ -56,14 +63,14 @@ class AddUniquenessOfOrderIdToSpreeShipments < ActiveRecord::Migration
       shipment.id,
       shipment.tracking,
       shipment.number,
+      shipment.order.number,
+      shipment.shipping_method.andand.name,
       shipment.cost,
+      shipment.state,
       shipment.shipped_at,
-      shipment.order_id,
-      shipment.shipping_method_id,
-      shipment.address_id,
       shipment.created_at,
       shipment.updated_at,
-      shipment.state
+      shipment.address.to_json
     ]
   end
 


db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
```diff
@@ -1,5 +1,19 @@
...
+    # We keep the latest shipment in the order
+    Spree::Shipment
+      .select(:order_id)
+      .group(:order_id)
+      .having("count(*) > 1")
+      .each do |shipment|
+        Spree::Shipment
+          .where("order_id = #{shipment.order_id}")
+          .order("updated_at")
+          .first
```
mkllnk: So you destroy only one shipment here, right? Could there be more than one?

I have a different idea:
```rb
latest_shipments = Spree::Shipment.
  select("order_id, MAX(updated_at) updated_at").
  group(:order_id).
  having("count(*) > 1")

Spree::Shipment.
  joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id=latest_shipments.order_id AND spree_shipments.last_updated!=latest_shipments.updated_at").
  destroy_all
```

db/migrate/20180426145633_add_stock_location_id_to_spree_shipments.spree.rb
```diff
@@ -2,5 +2,9 @@
 class AddStockLocationIdToSpreeShipments < ActiveRecord::Migration
   def change
     add_column :spree_shipments, :stock_location_id, :integer
+
+    # This Spree migration has not come with data migration and we need all shipments to have a valid stock location
+    #   This update fixes that problem
+    Spree::Shipment.update_all("stock_location_id = #{Spree::StockLocation.first.id}")
```
sauloperez: Doesn't Spree migrate the data in an upcoming migration? Is a common strategy to relax constraints, migrate data and in a later migration, constraints are set back.

luisramos0: I moved this topic to a separate issue here, let's talk there: #3600

db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
luisramos0: accepted, thanks a lot!

db/migrate/20180426145669_migrate_tax_categories_to_line_items.spree.rb
```diff
@@ -3,6 +3,10 @@ class MigrateTaxCategoriesToLineItems < ActiveRecord::Migration
   def change
   	Spree::LineItem.includes(:variant => { :product => :tax_category }).find_in_batches do |line_items|
```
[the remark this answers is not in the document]
luisramos0: yes, done! thanks.

db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
```diff
@@ -1,6 +1,23 @@
...
+  # Destroy all but the latest shipment in each order
+  def destroy_all_but_latest_shipments
+    latest_shipments = Spree::Shipment.
+      select("order_id, MAX(updated_at) updated_at").
+      group(:order_id).
+      having("count(*) > 1")
+
+    Spree::Shipment.
+      joins("INNER JOIN (#{latest_shipments.to_sql}) latest_shipments ON spree_shipments.order_id=latest_shipments.order_id AND spree_shipments.updated_at != latest_shipments.updated_at").
+      destroy_all
```
kristinalim: It is unlikely for two shipments for the same order to have the same `updated_at`, but it is still possible.

1. Can we also check for the maximum ID to handle this case?
2. Should we consider the shipment state, e.g. prefer a shipment that is "shipped" over one that is "ready" or "pending"? (State machine [here](https://github.com/openfoodfoundation/spree/blob/step-6a/core/app/models/spree/shipment.rb#L66))

And sorry if this was already discussed before: Do we have a system for validating a certain v1 instance will not lose any data when migrating? For example, for this PR, if there was an order for which I had to ship twice, I might lose tracking data for the first shipment. Are we okay with this?

luisramos0: I have checked live data before doing this and we are talking about around 16 orders in each of main environments (AU, FR and UK). Each environment has 2/3 orders like this in 2019...
All the duplicate shipments of an order have the same workflow state. All the shipments in the orders with duplicate shipments have different updated_at, for example:
```
 29217 |          | H86046673735 |      |                            |   398707 |                 46 |     269274 | 2018-11-04 01:58:55.405496 | 2018-11-04 01:58:55.405496 | pending
 29218 |          | H46481084155 |      |                            |   398707 |                 46 |     269276 | 2018-11-04 01:58:55.690717 | 2018-11-04 01:58:55.690717 | pending
```
I have checked some addresses of these duplicates and they are also the same... the shipping_methods are also the same for the duplicates I checked.
So, I think your point is very valid but not applicable here. I dont think we need to spend more time with this. what do you think?

regarding your last question:
no, we dont have any automatic validation in place, we just make sure that in every PR we keep the data consistent.
I dont think we will have any trouble with this particular PR.
do you have some other part of the model in mind where we should think about this? It's an important topic...

sauloperez: It'd be great to get an idea of how long this takes with the snapshot we are currently working with. Unless it takes too long I wouldn't care much about `updated_at` or `id` or any other improvement.

We also need to be aware that we're instantiating the shipments to destroy them afterward (executing callbacks) and that takes time. Again, if it is quick enough I wouldn't care much.

luisramos0: we are talking about changing 20/40 records in a table with 20k/30k records. I dont think it will be a problem.

kristinalim: I think it's bad practice to assume that the latest shipment is the one that we should keep. And actually, I'm uncomfortable with the idea of losing the older shipment data too.

Relating this with @mkllnk's recommendation for a CSV export before. What about going ahead with the code to keep the latest shipment record, but exporting to a CSV the shipment data of the ones we destroy? The instances can then decide if they want to contact the affected enterprises - @luisramos0 you said there are few so it would be easy enough for them.

luisramos0: ok, I think your concern is very important, I moved the generic question to the data migration epic [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/3285#issuecomment-476709344).

it's not just bad practice, this would be totally unacceptable in another contexts (including other areas of the OFN model). but the further I look into the live data the more I think we dont need to worry here.

I'll consider our 4 main envinroments here.
I created a few queries.

#### duplicate shipments (multiple shipments in the same order)
```
select * from spree_shipments
INNER JOIN (select order_id from spree_shipments group by order_id having count(*) > 1) orders_with_duplicates ON spree_shipments.order_id = orders_with_duplicates.order_id;
```
This shows a total of 114 duplicates (~50 orders in total)
- fr live 39 rows
- uk prod 33 rows
- katuma prod 2 rows
- aus prod 40 rows

from these, I checked what were the ones where shipping_method or state was different between duplicates:

#### duplicate spree shipments with different states or shipping method
```
select * from spree_shipments
INNER JOIN (select order_id from spree_shipments group by order_id having count(*) > 1) orders_with_duplicates ON spree_shipments.order_id = orders_with_duplicates.order_id
where exists (select id from spree_shipments sp where sp.order_id = spree_shipments.order_id and sp.id != spree_shipments.id and (sp.state != spree_shipments.state OR sp.shipping_method_id != spree_shipments.shipping_method_id));

```
- FR prod - no entries
- in uk prod - 3 orders where the shipping method changed (one order in 2018, from Sept)
- katuma prod - no entries
- AUS prod - 5 orders where the shipping method changed (one order in 2018, from Nov)

and then,
#### duplicated shipments with address diferences
```
select * from spree_shipments
INNER JOIN (select order_id from spree_shipments group by order_id having count(*) > 1) orders_with_duplicates ON spree_shipments.order_id = orders_with_duplicates.order_id
where exists (select sp.id from spree_shipments sp
  inner join spree_addresses a1 on (a1.id = sp.address_id)
  inner join spree_addresses a2 on (a2.id = spree_shipments.address_id)
  where sp.order_id = spree_shipments.order_id and sp.id != spree_shipments.id and sp.address_id != spree_shipments.address_id and
    (
        a1.firstname != a2.firstname OR a1.lastname != a2.lastname OR a1.address1 != a2.address1 OR a1.address2 != a2.address2 OR a1.city != a2.city
        OR a1.zipcode != a2.zipcode OR a1.phone != a2.phone OR a1.state_id != a2.state_id OR a1.country_id != a2.country_id
        OR a1.company != a2.company OR a1.latitude != a2.latitude OR a1.longitude != a2.longitude OR a1.state_name != a2.state_name
        OR a1.alternative_phone != a2.alternative_phone
    )
  );
```
and this is empty in all environments.

So, we are really talking about 8 orders (2 from 2018) where the shipping method was updated.

luisramos0: I think you suggestion is acceptable, it just takes time...
I'll add a new commit with a backup process shortly.

luisramos0: ok, done [here](3742c70910897c2069ae5afc35e1778c5ae879ee).
Let me know if this was what you were thinking about.
Thanks!

kristinalim: @luisramos0 I didn't realize that the "H86046673735" and "H46481084155" in your examples above are Spree-generated shipment numbers - I thought they were user-input tracking numbers. :joy: 

Nevertheless, I still think we should extract these data because of the fact that the users could have deleted the shipments but they did not.

db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
```diff
@@ -16,8 +16,67 @@ def destroy_all_but_latest_shipments
...
+  def shipment_csv_row(shipment)
+    [
+      shipment.id,
+      shipment.tracking,
+      shipment.number,
+      shipment.cost,
+      shipment.shipped_at,
+      shipment.order_id,
+      shipment.shipping_method_id,
+      shipment.address_id,
```
kristinalim: The three above would be more useful to the instance owner if these were order number, shipping method name, and even just JSON of the address object.

db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
kristinalim: @luisramos0 I wondered why the users did not remove the previous shipments, and realized it could be because these shipments were associated with shipping fee adjustments: `source_type: "Spree::Shipment"` ...

Do the older shipments have associated adjustments? If yes, maybe it will work to just reassign the adjustments to the latest shipments?

luisramos0: I checked the db and there are adjustments for all shipments, most of the amounts are zero.
in FR and UK, the duplicate shipments have all the same adjustment. in AUS, only in 3 orders we have adjustments with different amounts.

I think I can add to the migration:
- backup all the adjustments of duplicate shipments
- delete all the adjustments of the shipments that are being deleted

is this acceptable for you? let me know.

luisramos0: H86046673735 is the shipment number, tracking is another field in the shipment.

db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
luisramos0: :+1: done in the new commit

db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
luisramos0: I have added one more commit to backup adjustments with the shipments and delete the adjustments before deleting the duplicate shipments.

kristinalim: @luisramos0 My preference is to keep the adjustment but make it like a custom unassociated adjustment (what you get when using "New Adjustment" button). From a quick test adding such an adjustment, seeing that `source`, `originator`, and `mandatory` are blank. What do you think about this approach?

If doing this way, I think we don't need to export affected adjustments.

The intention is to preserve as much of the users' current data  as we can.

luisramos0: I am not sure I understand , do you mean to update the adjustments of the deleted shipments and some fields to blank? If that's what you suggest, I dont like the idea of manipulating the data, things can break. I prefer to delete them.
But if you prefer the update, please be a bit more specific about your suggestion and I'll implement it, ok?

kristinalim: My idea was to do it like this, @luisramos0:

```ruby
adjustments_scope.update_all(source_id: nil, source_type: nil, originator_id: nil,
                             originator_type: nil, mandatory: nil)
```

That just removes their association with the shipment and shipping method, instead of deleting the adjustments.

db/migrate/20180426145669_migrate_tax_categories_to_line_items.spree.rb
```diff
@@ -1,8 +1,12 @@
...
 class MigrateTaxCategoriesToLineItems < ActiveRecord::Migration
   def change
-  	Spree::LineItem.includes(:variant => { :product => :tax_category }).find_in_batches do |line_items|
-  	  line_items.each do |line_item|
+    Spree::LineItem.includes(:variant => { :product => :tax_category }).find_in_batches do |line_items|
+      line_items.each do |line_item|
+        next if line_item.variant.nil?
+        next if line_item.variant.product.nil?
+        next if line_item.product.nil?
+        next unless line_item.product.tax_category.present?
```
kristinalim: Both variants and tax categories have default scope (both use `deleted_at IS NULL`, and tax category has `"spree_tax_categories"."is_default" = 't'`). These should be unscoped in the checks above - the DB still knows the tax category even if the variant and tax category have already been soft-deleted.

luisramos0: This code is copied from Spree.
I dont see much benefit in updating the tax category of a line item with a variant that has been deleted...
I think I am saying this because I dont know how to do it 🤣 

Is there a way to go through the associations like line_item.variant.unscoped?

db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
luisramos0: ok, I replaced the last commit with that.
I tested with uk live data and it works.
this is the SQL query:
```
UPDATE "spree_adjustments"
SET "source_id" = NULL, "source_type" = NULL, "originator_id" = NULL, "originator_type" = NULL, "mandatory" = NULL
WHERE "spree_adjustments"."id" IN
(SELECT "spree_adjustments"."id"
FROM "spree_adjustments"
INNER JOIN (SELECT "spree_shipments".*
          FROM "spree_shipments"
          INNER JOIN (
                 SELECT order_id, MAX(updated_at) updated_at
                  FROM "spree_shipments"
                  GROUP BY order_id HAVING count(*) > 1) latest_shipments
             ON spree_shipments.order_id = latest_shipments.order_id
                 AND spree_shipments.updated_at != latest_shipments.updated_at) shipments_to_delete
  ON shipments_to_delete.id = spree_adjustments.source_id
     and spree_adjustments.source_type = 'Spree::Shipment')
```

db/migrate/20180426145669_migrate_tax_categories_to_line_items.spree.rb
mkllnk: Yes, there is `unscoped` but it would return all variants, not just the ones of that line item. There should be a special scope `with_deleted` for this case though.

kristinalim: > I dont see much benefit in updating the tax category of a line item with a variant that has been deleted...

Thanks, @luisramos0. I'm okay with keeping this as is then. @mkllnk Are you okay with not touching this anymore too?

mkllnk: Yes, all good.

PR #3599, opened 2019-03-12
Bring in latest nl_BE translations

sauloperez:
#### What? Why?

As we didn't reach 100% translation Transifex does not sync the changes into the repo. However, the Belgian instance still needs it. We should aim for that 100% so that changes are always maintained by Transifex itself.

#### What should we test?

Get sign off from Theo that all looks ok.

#### Release notes

Flemish translation (nl_BE) which will be used by the Belgian instance
Changelog Category: Added


luisramos0:
hey, actually, I think we should test this language in staging. otherwise, we deploy into live, Theo will detect another problem, and we will have to do another release. because it's a new language, we can give it a try in staging and get sign off from Theo, what do you think?

mkllnk:
We need a better solution for translations. There is always a delay and we fail to keep code and translations in sync. In the few days it takes people to translate we already merge a handful of new pull requests that change the text and need translating.

One idea would be: Don't tie Transifex to master. Use the latest (draft) release for translations and instances can pull the latest translations during deploy (automated). Maybe a discussion for France.

luisramos0:
ah, yes, nice! I immediately thought of a ofn-install task for it refresh-translations. we could run it from semaphore in any specific server including staging :-D

kristinalim:
This is `pr-no-test` and can be moved to Ready to Go now, right?

sauloperez:
These ideas sound really good but I think they deviate from the actual problem we are experiencing and this is trying to solve. That is the fact that skipping the established release process, in this case bringing in translations into a custom branch branched off the last release and deploying it to BE because they couldn't wait, has this consequences: as the nl_BE wasn't enabled nor pulled in master as soon as we deployed v1.28.0 these translations got lost. This leads to someone having to the extra work that wasn't planned.

I'm not blaming anyone, just trying to bring the focus to what IMO is the actual problem. So note, that even with the improvement you suggest @mkllnk , that wouldn't have solved the problem in the first place. So to me, it feels more organizational than tech problem.

sauloperez:
@kristinalim I think @luisramos0 is right. It is not the first time it happens, one of those the usual /shops link was broken and we noticed afterwards.

kristinalim:
Ah cool! I didn't read the comments. :slightly_smiling_face: Yes, that also makes sense to me.

luisramos0:
I staged this to uk and saw the NL menu broken...
I asked for transifex admin power, thanks @sigmundpetersen 
I fixed the menu entries in transifex, pulled nl_BE and committed on top of this PR. See new commit with the correct menu entries.
I staged it in UK, it's looking good. Ready for testing.
@RachL maybe Theo can validate it now here? https://staging.openfoodnetwork.org.uk/?locale=nl_BE

RachL:
@luisramos0 Theo gave his blessing :)

luisramos0:
Blessed we are!

sauloperez:
Thanks a lot for the heads-up @luisramos0 ! It was trully needed!

Issue #3594, opened 2019-03-11
[Spree Upgrade] Fix shipping categories data migration issues


luisramos0:
## Description
As we migrate to v2, shipping methods are picked up from shipping categories and are filtered by product.shipping_category_id.
the column shipping_metho.shipping_category_id in v2 was transformed in a table spree_shipping_method_categories.
Currently in most v1 data, shipping_methods do not have a shipping_category_id defined, and most products do not have a shipping_category_id defined. That makes most processes break.

We need to set shipping category on all shipping methods and we need to set shipping category on all products.
The easiest solution should be to create a default shipping cateogry and assign all products and shipping methods without shipping category to it.

This is already done in Spree:
https://github.com/spree/spree/commit/af16cf1d7453bd30d204f92bff10352817bc922d#diff-846be66c71d8cf32f6749d49fd68726f
We need to replicate this.
I dont think we should fetch the first category as this migration above, we should create a "default" one always.

## Expected Behavior
This will make edit order page work. Currently, after migrating live data to v2, when editing orders, refresh_rates is executed and breaks old orders.

## Actual Behaviour
We can edit old completed orders without breaking the shipping methods set on the order.

### On a related topic
If refresh rates is not called on the orders edit page, the problem above will not be so bad.
We need to understand understand why our version of edit:
https://github.com/openfoodfoundation/spree/blob/f55722b38db7e706a8521c9091a0e00119bb4d20/backend/app/controllers/spree/admin/orders_controller.rb#L54-L60

is different from spree v2.0 solution:
https://github.com/spree/spree/blob/3c46d141c5cb2bedced63596178a0a79df3f0c9e/backend/app/controllers/spree/admin/orders_controller.rb#L54-L60

PR #3602, opened 2019-03-13
[Spree Upgrade] Add migrations necessary to successfully migrate data to v2

### LINKED ISSUE #3600: [Spree Upgrade] Fix shipments.stock_location_id migration issue
## Description
After migrating to v2, shipment.stock_location_id become null.

## Expected Behavior
After migrating v1 data to v2, we dont see any shipment.stock_location_id nulls.

## Actual Behaviour
shipment.stock_location_id is null for all shipments after migration.

## Steps to Reproduce
Fetch live data in v1 model, run db:migrate, inspect db.

## Possible Fix
Because we only have one stock location, we can make all existing shipments get the default stock location.
To do that we need to bring spree/spree@219e09b from spree v2.1.0

What's the best way to include it? Shall we simply copy the migration code to our branch as a new migration in OFN?

luisramos0:
#### What? Why?

Closes #3600 and #3594 

Add migrations necessary to successfully migrate live data to v2.

#### What should we test?
Migrating live data to v2 should keep the app working, namely:
- all shipments should have stock location defined
- all shipping methods and products should have a shipping category

#### Dependencies
This is the continuation of #3591

### DIFF
--- a/db/migrate/20190313142051_set_default_stock_location_on_shipments.rb
+++ b/db/migrate/20190313142051_set_default_stock_location_on_shipments.rb
@@ -1,0 +1,8 @@
+class SetDefaultStockLocationOnShipments < ActiveRecord::Migration
+  def up
+    if Spree::Shipment.where('stock_location_id IS NULL').count > 0
+      location = DefaultStockLocation.find_or_create
+      Spree::Shipment.where('stock_location_id IS NULL').update_all(stock_location_id: location.id)
+    end
+  end
+end
--- a/db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
+++ b/db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
@@ -1,0 +1,13 @@
+class AddShippingCategoryToShippingMethodsAndProducts < ActiveRecord::Migration
+  def up
+    # This is different from the equivalent Spree migration
+    #   Here we are creating the default shipping category even if there are already shipping categories
+    default_category = Spree::ShippingCategory.create!(name: "Default")
+
+    Spree::ShippingMethod.all.each do |method|
+      method.shipping_categories << default_category if method.shipping_categories.blank?
+    end
+
+    Spree::Product.where(shipping_category_id: nil).update_all(shipping_category_id: default_category.id)
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20190221214542) do
+ActiveRecord::Schema.define(:version => 20190313142540) do
 
   create_table "account_invoices", :force => true do |t|
     t.integer  "user_id",    :null => false


db/migrate/20190313142051_set_default_stock_location_on_shipments.rb
```diff
@@ -0,0 +1,8 @@
+class SetDefaultStockLocationOnShipments < ActiveRecord::Migration
+  def up
+    if Spree::Shipment.where('stock_location_id IS NULL').count > 0
+      location = DefaultStockLocation.find_or_create
```
mkllnk: I would have gone the other way, defining `StockLocation` and `Shipment` here to be independent of code changes. But I'm happy to go with this for now.

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
```diff
@@ -0,0 +1,13 @@
+class AddShippingCategoryToShippingMethodsAndProducts < ActiveRecord::Migration
+  def up
+    # This is different from the equivalent Spree migration
+    #   Here we are creating the default shipping category even if there are already shipping categories
+    default_category = Spree::ShippingCategory.create!(name: "Default")
```
mkllnk: Is it safe to assume that we don't have a shipping category yet? I guess so.

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    Spree::ShippingMethod.all.each do |method|
```
mkllnk: I know this comes from Spree and changing it is out of scope. But just for similar cases in the future `Spree::ShippingMethod.find_each` would be better as it iterates in batches when there are lots of records. https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

It only matters if there are more than 1000 records and we have ~500 in Australia. So it should be fine.

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
luisramos0: Even if we have other shipping categories (we will in some production environments), I think it's good that we get this default one. We can then use it to make shipping category optional at some point...

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
luisramos0: nice, thanks for pointing that out!

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
mkllnk: Okay, let me rephrase the question: Is it safe to assume that we don't have a `Default` shipping category yet? I guess this would fail then.

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
sauloperez: that applies to all updates we perform. I checked the one SetDefaultStockLocationOnShipments and I don't think it'll be a problem for the numbers we have so I prefer to lean towards code simplicity instead of performance. Agree on this?

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
sauloperez: There is neither an ActiveModel validation nor a DB constraint so we will have duplicates at most. Even in that case, I think we could just merge them later on right?

luisramos0: ah, I see you question now Maikel.
Default duplicates will not be a problem, all we need here is a shipping category to assign to all products and methods without a shipping category. after this, the users will have to set shipping categories on products and methods as they create them. and admins will be able to manage shipping categories as they want (including this Default one).

luisramos0: I agree with Pau, this will never fail as there is no constraint on unique shipping_category.name. And if we end up with duplicate names (there's no "Default" shipping category in any of our main instances right now), admins will be able to manage them as they want.

db/migrate/20190313142540_add_shipping_category_to_shipping_methods_and_products.rb
luisramos0: :+1:

luisramos0:
you guys have approved the PR so I assume you are ok with the two points raised above. I have tested this with AUS data, it works. So this is ready to be merged. Correct me if I am wrong.

PR #3604, opened 2019-03-14
[Spree Upgrade] Override default spree splitters config

luisramos0:
#### What? Why?
After #3594, I found that if you try to checkout with two products with different shipping categories, the order would be split into multiple shipments and checkout would break with an error on the DB constraint index_spree_shipments_on_order_id, unique: true.

The solution is to override default spree config and set one single splitter called Base that never splits the orders.

In spree, orders are always split by stock location and optionally by shipping categories and/or on_hand/backordered items.

#### Context
We have one single stock location and one shipment per order to significantly simplify the solution to manage stock at variant overrides level: we dont have stock items for variant overrides, just a single count_on_hand in the variant_overrides table.

#### What should we test?
Create an order with products with different shipping categories, you should be able to checkout.

Part of testing v2 with live data.

#### Spree bug and PR to fix
In the investigation that led to this PR I found a bug in the Spree adaptation of package_factory that was breaking all splitters except the Base splitter. But because this PR changes OFN to use only the Base splitter, we don't need to fix that bug right now.
Anyway, there's a new Spree PR that fixes this splitters problem (which is not necessary) but also fixes a problem in shipment.to_package that is required for OFN (#3607).
The Spree PR: https://github.com/openfoodfoundation/spree/pull/39

### DIFF
--- a/spec/models/spree/order/checkout_spec.rb
+++ b/spec/models/spree/order/checkout_spec.rb
@@ -36,9 +36,7 @@ describe Spree::Order do
       order.order_cycle = create(:simple_order_cycle,
                                  distributors: [order.distributor],
                                  variants: [order.line_items.first.variant, other_variant])
-      order.line_items = [order.line_items.first, create(:line_item,
-                                                         order: order,
-                                                         variant: other_variant)]
+      order.line_items << create(:line_item, order: order, variant: other_variant)
     end
 
     it "can progress to delivery" do
--- a/spec/models/spree/order/checkout_spec.rb
+++ b/spec/models/spree/order/checkout_spec.rb
@@ -2,9 +2,9 @@ require 'spec_helper'
 
 describe Spree::Order do
   describe 'event :restart_checkout' do
-    context 'when the order is not complete' do
-      let(:order) { create(:order) }
+    let(:order) { create(:order) }
 
+    context 'when the order is not complete' do
       before { allow(order).to receive(:completed?) { false } }
 
       it 'does transition to cart state' do
@@ -13,8 +13,6 @@ describe Spree::Order do
     end
 
     context 'when the order is complete' do
-      let(:order) { create(:order) }
-
       before { allow(order).to receive(:completed?) { true } }
 
       it 'raises' do
@@ -24,6 +22,34 @@ describe Spree::Order do
             /Cannot transition state via :restart_checkout/
           )
       end
+    end
+  end
+
+  describe "order with products with different shipping categories" do
+    let(:order) { create(:order_with_totals_and_distribution, ship_address: create(:address) ) }
+    let(:shipping_method) { create(:shipping_method, distributors: [order.distributor]) }
+    let(:other_shipping_category) { create(:shipping_category) }
+    let(:other_product) { create(:product, shipping_category: other_shipping_category ) }
+    let(:other_variant) { other_product.variants.first }
+
+    before do
+      order.order_cycle = create(:simple_order_cycle,
+                                 distributors: [order.distributor],
+                                 variants: [order.line_items.first.variant, other_variant])
+      order.line_items = [order.line_items.first, create(:line_item,
+                                                         order: order,
+                                                         variant: other_variant)]
+    end
+
+    it "can progress to delivery" do
+      shipping_method.shipping_categories << other_shipping_category
+
+      # If the shipping category package splitter is enabled,
+      #   an order with products with two shipping categories will be split into two shipments
+      #   and the spec will fail with a unique constraint error on index_spree_shipments_on_order_id
+      order.next
+      order.next
+      expect(order.state).to eq "delivery"
     end
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -83,6 +83,17 @@ module Openfoodnetwork
       ]
     end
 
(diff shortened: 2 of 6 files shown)


luisramos0:
yes, multi shipments will make upgrades easier for sure. but for that we need to make variant_overrides work with stock items...

spec/models/spree/order/checkout_spec.rb
```diff
@@ -26,4 +24,32 @@
...
+    let(:shipping_method) { create(:shipping_method, distributors: [order.distributor]) }
+    let(:other_shipping_category) { create(:shipping_category) }
+    let(:other_product) { create(:product, shipping_category: other_shipping_category ) }
+    let(:other_variant) { other_product.variants.first }
+
+    before do
+      order.order_cycle = create(:simple_order_cycle,
+                                 distributors: [order.distributor],
+                                 variants: [order.line_items.first.variant, other_variant])
+      order.line_items = [order.line_items.first, create(:line_item,
```
sauloperez: What about `order.line_items << create(:line_item, order: order, variant: other_variant)`?. AR is here to help you 😂

luisramos0: gosh, why did I do that? 🤣 
I fixed it in a new commit.

luisramos0:
Build very broken after rebase. Looks like this PR is incompatible with #3560. I need to have a look...

luisramos0:
ok, fixed, needs re-review on [this new commit](b780823970e2d4b53c8525d8f8ad78adea4c9bf4).

RachL:
@luisramos0 I didn't managed to checkout... 

This is the shop I've setup: https://staging.openfoodfrance.org/shipping-method-test/shop 

And my testing notes: https://docs.google.com/document/d/1_JBxzd2I1s0HMdiN4PWK_oYxFuN9e5lf8_F1MoTSZxM/edit#

luisramos0:
The problem is that the "Pick-up dude!" shipping method is available for products with the "Default" shipping category and you are checking out frozen and refrigerated products.

If you add frozen and refrigetated categories to the "Pick-up dude!" it will work. Please give it a try. That is what this issue resolves.

I think we need to create a separate issue for the error message on checkout. For your test, it should have some indication to the user that the shipping method is not valid.

RachL:
@luisramos0 ok sorry I didn't realized bot were linked, I will retest this. Is there a way to put a message for the hub manager and not the user? Like if I open an order cycle with only the wrong shipping category?

luisramos0:
it is possible, I think it will be more work...

RachL:
@luisramos0 I've added refrigeration and frozen categories to both of my shipping methods. I can't checkout my "Eggs" product (when I click on "Edit your cart" the page reload and that's it). When I look at the network tab I see something as a missing taxonomy with a 404 error.

I changed the "Eggs" category to alcohol & drinks but this didn't change anything. Same if I change "Eggs" shipping category I still can't checkout this product, so weird.

So I created a 3rd product form scratch : Rhubarb, put it on demand and added it to the order cycle. But it does not show up in the shopfront.

I believe I have cache issue, right? :scream: 

I also noticed the following errors:
- you cannot checkout with a product price setup at 0. But no error message appears => I will open an issue.
- if you change the product price in the variant from 0 to something it does not show up directly in the shopfront (I had to clear my browser cache a few times - but maybe this is normal)

luisramos0:
I had a look at this. It's not related to this PR. It's related to empty prices and also maybe the instance currency config, because it's uk data it's all in gbp and the staging server is EUR, so the cache update process breaks, probably because of this (this DB has 30000 prices...) or the empty prices.
I think we should open separate issue for empty prices (note that empty price is different from zero price).

I have updated the prices of eggs to be not empty and I manage to checkout.

RachL:
> I have updated the prices of eggs to be not empty and I manage to checkout.

@luisramos0 you saw an empty price for eggs? That's weird because I made the mistakes on Almonds and not Eggs. By the way, I still see in the shopfront the Almonds at 0. Even when browsing in private mode and clearing all cookies.

I will delete all those products and start again and I will open the issue for the empty prices.

RachL:
Everything is fine on this second test. I'm moving this to ready to go and will open issues for related bugs I saw.

https://docs.google.com/document/d/1_JBxzd2I1s0HMdiN4PWK_oYxFuN9e5lf8_F1MoTSZxM/edit#

RachL:
Update: OK with the new enterprise I created, I couldn't reproduce any of the errors. I managed to checkout with a product with an empty price, and updating the variant price displayed correctly in the shopfront

luisramos0:
you cant replicate, interesting... something went wrong with the empty price in the first test, maybe related to the cache... another idea I have is that some data from uk live came as GBP and we are using EUR on top of all that GBP data.

mkllnk:
So can this be merged?

luisramos0:
yes, can be merged, the discussion was now about empty prices unrelated to this PR.

sigmundpetersen:
@luisramos0 the 2-0 build broke after this merge https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/2-0-stable/builds/249 
It seems to still be broken on next build.

luisramos0:
problem addressed here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3718

PR #3605, opened 2019-03-14
[Subscriptions] (v1) Fix assertions in specs for OrderSyncer

kristinalim:
#### What? Why?

Relates to #2696 

There are some wrong assertions in specs for `OrderSyncer`, caused by addresses (subscription shipping, subscription billing, distributor) having the same attributes.

This shows how it can be dangerous to use static data in factories.

#### What should we test?

This only affects specs. No manual tests are needed.

#### Release notes

- Fix setup data and assertions for `OrderSyncer` specs.

Changelog Category: Fixed

This is in preparation for some fixes for subscriptions in v2. This is best to merge in v1 so we can be sure that the behaviour that we wish to achieve in v2 is indeed the behaviour in v1.

### DIFF
--- a/spec/services/order_syncer_spec.rb
+++ b/spec/services/order_syncer_spec.rb
@@ -134,14 +134,14 @@ describe OrderSyncer do
     let!(:original_ship_address) do
       create(:address, firstname: "Melanie", lastname: "Miller", address1: "Magenta")
     end
+    let!(:shipping_method) { create(:shipping_method, distributors: [distributor]) }
 
     let(:subscription) do
       create(:subscription, shop: distributor, bill_address: original_bill_address,
-                            ship_address: original_ship_address, with_items: true,
-                            with_proxy_orders: true)
+                            ship_address: original_ship_address, shipping_method: shipping_method,
+                            with_items: true, with_proxy_orders: true)
     end
 
-    let(:shipping_method) { subscription.shipping_method }
     let!(:order) { subscription.proxy_orders.first.initialise_order! }
     let!(:bill_address_attrs) { subscription.bill_address.attributes }
     let!(:ship_address_attrs) { subscription.ship_address.attributes }
@@ -149,7 +149,9 @@ describe OrderSyncer do
     let(:syncer) { OrderSyncer.new(subscription) }
 
     context "when a ship address is not required" do
-      before { shipping_method.update_attributes(require_ship_address: false) }
+      let!(:shipping_method) do
+        create(:shipping_method, distributors: [distributor], require_ship_address: false)
+      end
 
       context "when the bill_address on the order matches that on the subscription" do
         it "updates all bill_address attrs and ship_address names + phone" do
@@ -188,7 +190,9 @@ describe OrderSyncer do
     end
 
     context "when a ship address is required" do
-      before { shipping_method.update_attributes(require_ship_address: true) }
+      let!(:shipping_method) do
+        create(:shipping_method, distributors: [distributor], require_ship_address: true)
+      end
 
       context "when the bill_address on the order matches that on the subscription" do
         it "only updates bill_address attrs" do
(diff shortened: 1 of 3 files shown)


spec/services/order_syncer_spec.rb
```diff
@@ -213,7 +233,24 @@
...
+    let!(:shipping_method) do
+      create(:shipping_method, distributors: [distributor], require_ship_address: true)
+    end
+
+    let(:subscription) do
+      create(:subscription, shop: distributor, bill_address: original_bill_address,
+                            ship_address: original_ship_address, shipping_method: shipping_method,
+                            with_items: true, with_proxy_orders: true)
+    end
+
```
luisramos0: DRY
This is copy pasted from above, right?
I'd prefer to see this in a factory.

spec/services/order_syncer_spec.rb
```diff
@@ -222,38 +259,68 @@
...
+            it "updates ship_address attrs" do
+              expect(syncer.sync!).to be true
+              expect(syncer.order_update_issues.keys).not_to include order.id
+              order.reload
+              expect(order.ship_address.firstname).to eq "Ship"
+              expect(order.ship_address.lastname).to eq ship_address_attrs["lastname"]
+              expect(order.ship_address.address1).to eq "123 abc st"
+              expect(order.ship_address.phone).to eq "1123581321"
+            end
+          end
```
luisramos0: this is not simple but I get it now :-)
the spec helps a lot!

spec/services/order_syncer_spec.rb
```diff
@@ -146,8 +164,8 @@
           expect(order.bill_address.address1).to eq "123 abc st"
           expect(order.bill_address.phone).to eq "1123581321"
           expect(order.ship_address.firstname).to eq "Bill"
-          expect(order.ship_address.lastname).to eq ship_address_attrs["lastname"]
-          expect(order.ship_address.address1).to eq ship_address_attrs["address1"]
+          expect(order.ship_address.lastname).to eq bill_address_attrs["lastname"]
```
sauloperez: why not using the address' object rather than its hash of attributes? I don't see any benefit on using the hash. Just more code.

sauloperez: Also, by are we switching from ship to bill address?

kristinalim: @sauloperez This is using `_attrs` to keep the original ship and bill attributes after the order is initialized. Syncing modifies the `order` object including sometimes the `original_ship_address` and `original_bill_address`.

kristinalim: > Also, why are we switching from ship to bill address? Sorry, I find all this file very hard to follow.

When the shipping method is for a pickup (`require_ship_address == false`), syncing the order with the subscription disregards whatever ship address you have in the subscription. It forces the ship address of the order to be the distributor's address (because it is a pickup) and to use the orders bill contact details.

This is existing behaviour. I didn't actually change the app code here - it just so happened that the previous assertions passed because `order.ship_address`, `order.bill_address`, and `distributor.address` used the same attributes (`create(:address)`).

spec/services/order_syncer_spec.rb
kristinalim: Thanks for nudging me to do this, @luisramos0. Responding to this concern in a main comment, because this will need another review from the other reviewers.

kristinalim:
I made follow-up changes related to @luisramos0's recommendation to move the setup code for subscription addresses to a factory.

I checked the feasibility of randomizing bill and ship addresses for **all** sample subscriptions. As mentioned, so far, they use the address factory of Spree which generates addresses with the same attributes. POC PR for this shows that only `OrderSyncer` has failing specs: https://github.com/openfoodfoundation/openfoodnetwork/pull/3637

So I've appended to this PR:

* **Added address factory trait `randomized` for randomized attributes.** IMO, this should not be a trait but should be the default for all generated sample addresses (equivalence should be intentional in specs). For exploring randomizing attributes for all sample addresses, I created this: #3638 If feasible, then the `randomized` trait would just be transitional.
* **Changed subscription factory to use randomized ship and bill addresses.** Phone number is now also customized, and this revealed other wrong phone number references, fixed here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3605/commits/0c2a75a22745ae24ef0e92ad3ea6010a4aafd2de

Requesting reviews again, because you might have issue with the `randomized` trait. Thank you!

spec/services/order_syncer_spec.rb
sauloperez: Yes, I remember fixing that behavior for v2. Thanks for the explanation!

luisramos0:
PR no test, it means this can be merged @sauloperez right?

sauloperez:
Sorry, I missed this one

PR #3609, opened 2019-03-15
[Spree Upgrade] Finish up deleting account invoices

### LINKED ISSUE #3533: [Spree Upgrade] Delete all running/pending/failed delayedjobs before migrating to v2
## Description
From [this comment in #3321](https://github.com/openfoodfoundation/openfoodnetwork/pull/3321#issuecomment-466277043), we should include a migration to delete all delayed_jobs from the DB when we migrate to v2. This includes any traces all jobs that will no longer exist in v2 (account invoices jobs).

## Acceptance Criteria
No delayed jobs in the DB after we migrate to v2

luisramos0:
#### What? Why?
Closes #3533

Additionally this PR:
- cleans up monthly bill description that was left from #3321
- drops account invoices and billable periods tables, for some reason not dropped in #3321

#### What should we test?
Tests are required for the "monthly bill description" removal.
We need to test all the pages where the package selection appears:
- the welcome page that enterprises see after registration
- the drop down menu "change package" in the single enterprise dashboard
- the non-admin list of enterprises

### DIFF
--- a/app/views/admin/enterprises/_change_type_form.html.haml
+++ b/app/views/admin/enterprises/_change_type_form.html.haml
@@ -22,9 +22,6 @@
             .top
               %h3= t('.producer_shop')
               %p= t('.sell_your_produce')
-            .bottom
-              %monthly-pricing-description{ joiner: "newline" }
-
           %p.description
             = t('.producer_shop_description_text')
             %br
@@ -36,9 +33,6 @@
             .top
               %h3= t('.producer_hub')
               %p= t('.producer_hub_text')
-            .bottom
-              %monthly-pricing-description{ joiner: "newline" }
-
           %p.description
             = t('.producer_hub_description_text')
 
@@ -59,8 +53,6 @@
             .top
               %h3= t('.hub_shop')
               %p= t('.hub_shop_text')
-            .bottom
-              %monthly-pricing-description{ joiner: "newline" }
           %p.description
             = t('.hub_shop_description_text')
         .two.columns.omega
--- a/app/assets/javascripts/admin/enterprises/directives/monthly_pricing_description.js.coffee
+++ b/app/assets/javascripts/admin/enterprises/directives/monthly_pricing_description.js.coffee
@@ -1,8 +1,0 @@
-angular.module("admin.enterprises").directive "monthlyPricingDescription", (monthlyBillDescription) ->
-  restrict: 'E'
-  scope:
-    joiner: "@"
-  template: "<span ng-bind-html='billDescription'></span>"
-  link: (scope, element, attrs) ->
-    joiners = { comma: ", ", newline: "<br>" }
-    scope.billDescription = monthlyBillDescription.replace("{joiner}", joiners[scope.joiner])
--- a/app/assets/javascripts/templates/admin/panels/enterprise_package.html.haml
+++ b/app/assets/javascripts/templates/admin/panels/enterprise_package.html.haml
@@ -20,10 +20,6 @@
           {{ 'js.admin.panels.enterprise_package.hub_shop' | t }}
 
         %p
-          %strong
-            %monthly-pricing-description{ joiner: "comma" }
-
-        %p
           {{ 'js.admin.panels.enterprise_package.hub_shop_text1' | t }}
 
         %p
@@ -67,10 +63,6 @@
           {{ 'js.admin.panels.enterprise_package.producer_shop' | t }}
 
         %p
-          %strong
-            %monthly-pricing-description{ joiner: "comma" }
-
-        %p
           {{ 'js.admin.panels.enterprise_package.producer_shop_text1' | t }}
 
         %p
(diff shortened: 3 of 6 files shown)


db/migrate/20190315225354_delete_all_pending_delayed_jobs.rb
```diff
@@ -0,0 +1,5 @@
+class DeleteAllPendingDelayedJobs < ActiveRecord::Migration
+  def up
+    Delayed::Job.delete_all
```
mkllnk: I'm not sure this is a good idea. There may be pending jobs that are still important. I think we should delete specific job types that are now obsolete.

mkllnk:
We can also keep the pricing information for now and just remove the logic to bill based on that.

luisramos0:
@mkllnk that is all gone already:
- https://github.com/openfoodfoundation/openfoodnetwork/pull/3321/files#diff-59c655beec205769b9b08f9358ca527cL113
- https://github.com/openfoodfoundation/openfoodnetwork/pull/3321/files#diff-7b755a8572efec7fc38da020f5403136
- https://github.com/openfoodfoundation/openfoodnetwork/pull/3321/files#diff-346e21da75b3b2c24c5e7f2a53816ad3

For the placeholder for pricing, we already have 
js.admin.panels.enterprise_package.hub_profile_text1
and 
js.admin.panels.enterprise_package.hub_profile_text1
for each profile, there's plenty of space to write whatever the instance wants about pricing. I dont see why we would need a price specific textarea.

mkllnk:
Ah, great. :+1:

What do you think about the delayed jobs?

db/migrate/20190315225354_delete_all_pending_delayed_jobs.rb
luisramos0: ok, I have changed it to only delete obsolete jobs related to account invoices. Please have a look.
ping @kristinalim are you ok with this?

luisramos0:
ok, so I rebased to resolved conflicts and changed the delayedjobs migration to just delete the obsolete ones: 804375d1aa785e3c4c1c0244b7c4944a3afef255

luisramos0:
ehehe, yes, I thought that must be a shorter way to write this :-D

db/migrate/20190315225354_delete_all_pending_delayed_jobs.rb
kristinalim: @luisramos0 Yes, I'm okay with this.

app/views/admin/enterprises/_change_type_form.html.haml
```diff
@@ -36,9 +33,6 @@
             .top
```
kristinalim: Unless putting back `monthly-pricing-description`, the `top` CSS class might no longer be necessary here.

luisramos0:
@kristinalim this is a dead feature. this specific dynamic part is not configured even in Australia...
we are just deleting it here.
we may get to it later if prioritized but for now I'd not spend more time here.

app/views/admin/enterprises/_change_type_form.html.haml
luisramos0: you should be right, I'll have a look at this

db/migrate/20190315225354_delete_all_pending_delayed_jobs.rb
kristinalim: In relation to this comment, I did quick check in recent commits and spotted that `SubscriptionConfirmJob` and `SubscriptionPlacementJob` have been renamed. Pending jobs of this type would not have a corresponding job class in v2. I created #3643 to address this.

kristinalim:
I'm not sure I understand correctly. Was the point of @mkllnk's concern that some instances bill enterprises offline, and now they won't see the price? Or is that data unused too?

kristinalim:
> this specific dynamic part is not configured even in Australia...

So there is no pricing description visible in AU?

luisramos0:
no, pricing is not used anywhere, from live:

FR
Screenshot 2019-03-22 at 21 09 38
ES
Screenshot 2019-03-22 at 21 10 29
AU
Screenshot 2019-03-22 at 21 10 16
UK
Screenshot 2019-03-22 at 21 10 38

this last one is not dynamic is a translation in enGB:
always_free: Plans starting from £1/month

thats why I think we should just delete, it's all been dead for 3 years and can be relived easily if required.
the other option is really to make this monthlyBillingDescription work without the dynamic part, I don't think we should spend time doing that.

db/migrate/20190315225354_delete_all_pending_delayed_jobs.rb
luisramos0: thank you!

luisramos0:
ok, thanks @kristinalim 
I was planning to look at the css but I forgot about it and now that you approved the PR... I will leave it 😊 
moving to test ready

RachL:
This is good to go: https://docs.google.com/document/d/1zsKXh8kkQl-JPbUqz-cf8-WfHrjEV2B9a2ocd59NGlM/edit#

PR #3613, opened 2019-03-17
[Spree Upgrade] Add on_hand and on_demand to the variants create and edit pages

### LINKED ISSUE #3573: [Spree upgrade] Impossible to update stock info (on hand and on demand)
## Description
I can't see the field in order to update a stock info in a variant. When creating a product everything is ok.

## Steps to Reproduce
1. Go to https://staging.openfoodfrance.org/admin/products
2. Edit a variant. E.g. https://staging.openfoodfrance.org/admin/products/apple-cider-1-2/variants/12/edit
3. You won't see the on hand stock field neither the on demand checkbox. 

## Animated Gif/Screenshot
image

## Severity

## Your Environment

* Version used: ??

luisramos0:
#### What? Why?

Closes #3573 

##### shared/_product_tabs
Here I de-deface shared/_product_tabs and remove the new Stock Management option from it.

##### Variants create/edit Page
In spree v2, on_hand and on_demand were removed from the variants edit page (on_hand [here](https://github.com/openfoodfoundation/spree/commit/64d5843b83b5f4f94d78fe2327f44aa1c7679f13#diff-4cc0ba3139a87f563a1faedee946fe6c) and on_demand [here](https://github.com/openfoodfoundation/spree/commit/40665ab244277138801ee278f663fd2f6da7da22#diff-4cc0ba3139a87f563a1faedee946fe6c)).

These fields were moved to a new submenu/page "Stock Management" but because this new page includes the complexity of stock management, it looks at lot easier (for both the dev and the user!) to just re-add these two fields in the variants edit page. Maybe when/if we make OFN work with multiple stock locations we can move OFN to use the stock management page.

This variants edit page is obviously needing a de-deface. But with the de-deface it will also need more specs and a new angular controller to do what's currently being done in  on_demand_script.html.haml.deface for example. It will take some effort so I created a new tech debt issue for it #3615

Meanwhile, in this PR, I added a new deface override to add the two fields back to the page and adapted on_demand_script.html.haml.deface so that it works.
In terms of improvement, I didn't want to waste time improving the overrides so I decided to write feature specs on top of the page that will be valid after #3615 I added the specs in master #3617 these specs pass in master and pass in v2 with this PR (I tested locally and we can validate this when #3617 is merged into 2-0-stable).

#### What should we test?
The variants edit page should work as normal and the user should be able to manage on_hand and on_demand. These 2 fields should play well with changing the same values on the bulk edit page.
I have tested this locally and it's working correctly.

### DIFF
--- a/app/controllers/spree/admin/variants_controller_decorator.rb
+++ b/app/controllers/spree/admin/variants_controller_decorator.rb
@@ -6,9 +6,13 @@ Spree::Admin::VariantsController.class_eval do
   def create
     on_demand = params[:variant].delete(:on_demand)
     on_hand = params[:variant].delete(:on_hand)
+
     super
-    @object.on_demand = on_demand if @object.present?
-    @object.on_hand = on_hand.to_i if @object.present?
+
+    if @object.present? && @object.valid?
+      @object.on_demand = on_demand if on_demand.present?
+      @object.on_hand = on_hand.to_i if on_hand.present?
+    end
   end
 
   def search
--- a/app/controllers/spree/admin/variants_controller_decorator.rb
+++ b/app/controllers/spree/admin/variants_controller_decorator.rb
@@ -3,6 +3,14 @@ require 'open_food_network/scope_variants_for_search'
 Spree::Admin::VariantsController.class_eval do
   helper 'spree/products'
 
+  def create
+    on_demand = params[:variant].delete(:on_demand)
+    on_hand = params[:variant].delete(:on_hand)
+    super
+    @object.on_demand = on_demand if @object.present?
+    @object.on_hand = on_hand.to_i if @object.present?
+  end
+
   def search
     scoper = OpenFoodNetwork::ScopeVariantsForSearch.new(params)
     @variants = scoper.search
--- a/app/overrides/spree/admin/shared/_product_tabs/add_distributions.html.haml.deface
+++ b/app/overrides/spree/admin/shared/_product_tabs/add_distributions.html.haml.deface
@@ -1,5 +1,0 @@
-/ insert_bottom "[data-hook='admin_product_tabs']"
-
-- klass = current == 'Product Distributions' ? 'active' : ''
-%li{:class => klass}
-  = link_to_with_icon 'icon-tasks', t('admin.products.product_distributions'), product_distributions_admin_product_url(@product)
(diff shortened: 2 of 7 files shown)


lin-d-hop:
@luisramos0 Tested this one and found that:
As enterprise user nothing saves on the Edit Variant page
As superadmin user everything works as expected
https://docs.google.com/document/d/1K0inimZQxiY-Eh7oc45r_ttpjh23_9TPxA0ydfGhylE/edit#

Not sure whether this is caused by this change or whether this was an existing issue.
Up to you whether this is acceptable to move forward for now or if you want to look into this now. Possibly needs new issue for being able to save anything on the Edit Variant page?

luisramos0:
Thanks @lin-d-hop for testing!
I dont think it's related to user profile. I think it's just the variants, I found Rubhard and Potatoes1 to be broken. Artichoke for example, works well. Can you please confirm this is correct for you as well?

Tech explanation: this is a problem with variants in the system that do not have stock_items. I am not sure how did we end up with variants that do not have stock_items.
I could write code to handle this but I think we need to find out how this is happening.

So @lin-d-hop , I suggest you confirm the error is related to specific variants and not user profiles, and if that's correct we reset the database and try to re-create the problem (variants without stock_items).

luisramos0:
the problem is "related" to this PR (this PR makes the problem more visible) but the problem is not caused by this PR. So, I think we can merge this PR and create a separate issue for this problem.

lin-d-hop:
Thanks @luisramos0 
Rechecked and yes you are right. It's not the profile it is the product.
Updated testing doc to reflect passing tests.
https://docs.google.com/document/d/1K0inimZQxiY-Eh7oc45r_ttpjh23_9TPxA0ydfGhylE/edit#

luisramos0:
ok, thanks!

for reference this is the invalid data:

```
openfoodnetwork=> select * from spree_variants where deleted_at is null and not exists (select * from spree_stock_items where variant_id = spree_variants.id);
 id | sku | weight | height | width | depth | deleted_at | is_master | product_id | cost_price | position | cost_currency | unit_value | unit_description | display_name | display_as | import_date
 10 | ABC |   0.00 |        |       |       |            | f         |          5 |      17.00 |        2 | AUD           |          1 |                  |              |            |
 22 |     |   0.00 |        |       |       |            | f         |          8 |            |        3 | AUD           |          3 |                  |              |            |
(2 rows)

```

I have reset the DB in staging AU and tried to replicate the issue by creating and deleting variants in both the bulk edit page and the new variant page and it all looks good.
I think we should merge this PR and monitor this to see if it happens again.

@lin-d-hop can we move to ready to go?

luisramos0:
This is blocking #3640 so I am going to merge it. We can open new issue if required.

PR #3614, opened 2019-03-18
Support international decimals in weight calculator

### LINKED ISSUE #3335: , and . problem is back on enterprise fees
## Description
In countries using , as a separator for decimals, when I setup a "per kg" enterprise fee, and fill in with "5,50" for instance, when I save it is transformed to "5.00".
This was fixed by Pierre last year, I did tested that specifically I remember, so probably was reintroduced by refactor ? (@mkllnk maybe it will ring a bell)

## Expected Behavior
User can enter indifferently , or . for decimals separator, it will work

## Actual Behavior
It doesn't.

## Steps to Reproduce
1. Go to enterprise fee.
2. Create/update a fee with per kg basis with , as decimal separator.
3. Save, you see the decimals are not taken into account.

## Animated Gif/Screenshot
Before clicking save
capture du 2019-01-16 13-06-57
After clicking save
capture du 2019-01-16 13-07-25

## Context
Found the bug while doing support to big italian hub setting hub some pilot on OFN.

## Severity
S3, there is a workaround which is to enter a . as a separator but people don't think about it and think it's broken in most european countries.

## Possible Fix
Check the refactor done by Maikel some time ago if there is a clue there ? It was working before as far as I remember.

mkllnk:
#### What? Why?

Closes #3335.

A previous pull request added support for flexible decimal characters
when editing money amounts.
https://github.com/openfoodfoundation/openfoodnetwork/pull/1831

This pull request applies the same principle to the weight calculator
which was missed in the previous pull request.

#### What should we test?

1. Go to admin settings and make sure that "Use the international thousand/decimal separator logic" is enabled.
1. Go to the enterprise fees of an enterprise.
2. Create a fee with "Weight (per kg)" calculator.
3. Update the fee with `,` as decimal separator, for example `1,23`.
4. Save and make sure that the price is correct, for example `1.23`. It should not be `0` or `1.00`.

#### Release notes

Fixed: The international decimal separator logic can now be used for the weight calculator as well.

Changelog Category: Fixed

### DIFF
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -1,7 +1,11 @@
+require 'spree/localized_number'
+
 module Calculator
   class Weight < Spree::Calculator
+    extend Spree::LocalizedNumber
     preference :per_kg, :decimal, default: 0.0
     attr_accessible :preferred_per_kg
+    localize_number :preferred_per_kg
 
     def self.description
       I18n.t('spree.weight')
--- a/spec/lib/spree/localized_number_spec.rb
+++ b/spec/lib/spree/localized_number_spec.rb
@@ -1,7 +1,19 @@
 require 'spec_helper'
 
 describe Spree::LocalizedNumber do
-  context ".parse" do
+  describe ".parse" do
+    context "with point separator" do
+      it "captures the proper amount for a formatted string" do
+        expect(described_class.parse('5.67')).to eql 5.67
+      end
+    end
+
+    context "with comma separator" do
+      it "captures the proper amount for a formatted string" do
+        expect(described_class.parse('5,67')).to eql 5.67
+      end
+    end
+
     context "with decimal point" do
       it "captures the proper amount for a formatted string" do
         expect(described_class.parse('1,599.99')).to eql 1599.99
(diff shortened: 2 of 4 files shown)


mkllnk:
@luisramos0 
> the shared example is not testing the case of the bug, is it?

No, it's not. It's only verifying that the value is passed on to `Spree::LocalizedNumber` which deals with the conversion logic. So testing the comma case should be done in that module's spec. I assumed it was Spree code (because of the name) but it's actually within OFN. I now added some more examples to the spec and cleaned up the shared example. We should move it out of the Spree namespace as well but maybe we leave that to after the Spree upgrade.

RachL:
@mkllnk I wanted to check the changes on a shopfront, but since I staged this PR all shops on UK staging have an error 500 when you try to reach them. could this be linked to this PR? I will staged in on Katuma also when I'm finished with the release.

luisramos0:
I see a product distributions table error in the logs in staging UK, the table was dropped and this PR is still using it. A rebase/redeploy of this PR will fix it.

mkllnk:
Okay, I rebased this.

RachL:
Thank you both @mkllnk @luisramos0 ! Tested shopfront and we are good to go :)

PR #3617, opened 2019-03-18
Add spec to cover editing on_hand and on_demand values in the variants edit page

luisramos0:
#### What? Why?

This PR adds a spec to cover editing on_hand and on_demand in the variants edit page (from what I have seen so far, this was the most significant feature the build was not covering and so we only detected now that it was broken in v2 #3573).

Additionally and unrelated, I am fixing a typo in the en.yml file (new_shipping_category).

#### What should we test?
The build should be green which means the new test is passing. It's testing an existing feature.

This spec is also passing in the v2 PR that fixes this feature in v2: #3613

#### Release notes
Changelog Category: Added
Improved test coverage in the edit variants page.

This was detected as part of the spree upgrade and will be fixed in v2 as part of #3613

### DIFF
--- a/spec/features/admin/variants_spec.rb
+++ b/spec/features/admin/variants_spec.rb
@@ -58,6 +58,47 @@ feature %q{
     v.unit_description.should == 'bar'
   end
 
+  describe "editing on hand and on demand values", js:true do
+    let(:product) { create(:simple_product) }
+    let(:variant) { product.variants.first }
+
+    before do
+      login_to_admin_section
+    end
+
+    it "allows changing the on_hand value" do
+      visit spree.edit_admin_product_variant_path(product, variant)
+
+      page.should have_field "variant_on_hand", with: variant.on_hand
+      page.should have_unchecked_field "variant_on_demand"
+
+      fill_in "variant_on_hand", with: "123"
+      click_button 'Update'
+      page.should have_content %Q(Variant "#{product.name}" has been successfully updated!)
+    end
+
+    it "allows changing the on_demand value" do
+      visit spree.edit_admin_product_variant_path(product, variant)
+      check "variant_on_demand"
+
+      # on_hand reflects the change in on_demand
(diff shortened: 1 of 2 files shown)


spec/features/admin/variants_spec.rb
```diff
@@ -58,6 +58,47 @@
...
+    let(:variant) { product.variants.first }
+
+    before do
+      login_to_admin_section
+    end
+
+    it "allows changing the on_hand value" do
+      visit spree.edit_admin_product_variant_path(product, variant)
+
+      page.should have_field "variant_on_hand", with: variant.on_hand
```
mkllnk: We should use the new `expect` syntax in all new specs. But since the rest of the file uses `should`: :see_no_evil:

spec/features/admin/variants_spec.rb
```diff
@@ -58,6 +58,47 @@
...
+      page.should have_unchecked_field "variant_on_demand"
+
+      fill_in "variant_on_hand", with: "123"
+      click_button 'Update'
+      page.should have_content %Q(Variant "#{product.name}" has been successfully updated!)
```
mkllnk: I think these tests miss `page.should have_field "variant_on_hand", with: "123"` here.

sauloperez:
@mkllnk given your approval I'll assume you commented on things that are nice to have and can be addressed as tech-debt issues later on

spec/features/admin/variants_spec.rb
luisramos0: ah, yes, sorry, I switched that detector long ago in my brain otherwise I'd go crazy looking at all the shoulds and thinking about expects and not being able to change them to avoid merge conflicts...

spec/features/admin/variants_spec.rb
luisramos0: yes, I agree...

PR #3620, opened 2019-03-18
Fix flaky spec by waiting for modal animation

### LINKED ISSUE #3520: [Spree upgrade][Flaky spec] `spec/features/admin/customers_spec.rb:184`

mkllnk:
#### What? Why?

Closes #3520

Bootstrap's fading in of the customer edit modal and the webdriver's sendKeys method returning too early lead to a race condition: https://github.com/teamcapybara/capybara/issues/1890

This pull request introduces waiting for the modal. If we have the same problem in other specs, we should share the waiting method.

#### What should we test?

No tests.

#### Release notes

Part of Spree upgrade.

### DIFF
--- a/spec/features/admin/customers_spec.rb
+++ b/spec/features/admin/customers_spec.rb
@@ -184,6 +184,7 @@ feature 'Customers' do
         it 'updates the existing billing address' do
           expect(page).to have_content 'BILLING ADDRESS'
           first('#bill-address-link').click
+          wait_for_modal_fade_in
 
           expect(page).to have_content 'Edit Billing Address'
           expect(page).to have_select2 'country_id', selected: 'Australia'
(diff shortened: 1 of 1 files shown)


mkllnk:
Ah, yes, the 184 spec still failed because I applied the fix only to the 210 spec. Updated the commit. Let's see if that works now.

PR #3624, opened 2019-03-20
Add spec for registration link on Spree pages

### LINKED ISSUE #2873: Register link on checkout page leads to broken page
## Description

The registration call at the bottom of the page sometimes leads to the wrong URL. It should go to `/registration`, but from the checkout page it goes to `/signup` which is Spree's path for registrations. Since we are not using Spree's signup, the page is broken and shows the snail.

## Expected Behavior

On every page the "Register here" call leads to `/register`.

## Actual Behavior

On some pages (checkout) the register call leads to `/signup`.

## Steps to Reproduce

1. Visit a shop.
2. Add a product to the cart.
3. Go to the checkout and scroll down.
4. Click on "Register here".

## Animated Gif/Screenshot

screenshot from 2018-10-16 18-09-12

## Context

Bugsnag reported the broken page. See below.

## Severity

S3 as every instance is affected (I believe) and it can prevent customers from ordering. The workaround is to go to another page and then click on the link. Another workaround is to click on login and then select the register tab.

## Your Environment

* Version used: f8bccda6abe1c859e2506a3265178d818d1a1d5d

## Possible Fix

I'm not sure why the link doesn't render deterministically. But I noticed that there are Spree routes with the same names.
```
# rake routes
                registration GET  /register(.:format)              registration#index
           registration_auth GET  /register/auth(.:format)         registration#authenticate
                      signup GET  /signup(.:format)                spree/user_registrations#new
                registration POST /signup(.:format)                spree/user_registrations#create
       checkout_registration GET  /checkout/registration(.:format) spree/checkout#registration
update_checkout_registration PUT  /checkout/registration(.:format) spree/checkout#update_registration
```

Maybe there is a way to deactivate the unused Spree routes? Otherwise we can use our own helper to refer to the right path.

## Bugsnag error: ActionView::Template::Error in spree/user_registrations#new

**ActionView::Template::Error** in **spree/user_registrations#new**
store/all.css isn't precompiled

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5bc556570aae8f001943d029?event_id=5bc585ce0030994927d00000&i=gh&m=ci)

## Stacktrace

    gems/actionpack-3.2.22.5/lib/sprockets/helpers/rails_helper.rb:142 - digest_for

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5bc556570aae8f001943d029?event_id=5bc585ce0030994927d00000&i=gh&m=ci)

mkllnk:
#### What? Why?

Closes #2873

The `registration_path` helper resolves to `/signup` in Spree
controllers due to spree_auth_device > config > routes.rb.
We worked around that in:
https://github.com/openfoodfoundation/openfoodnetwork/pull/3174

Here we add a spec for this so that we can test more easily if we can
remove that workaround or detect it's accidental removal.

#### What should we test?

No test, spec changes only.

#### Release notes

Added: Added a new integration test to ensure that the registration link works on all pages.

Changelog Category: Added

### DIFF
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -20,17 +20,6 @@ feature "full-page cart", js: true do
       set_order order
     end
 
-    it "shows the right registration link" do
-      # We have an issue with the registration link within Spree controllers.
-      # The `registration_path` helper resolves to `/signup` due to
-      # spree_auth_device > config > routes.rb. This spec verifies that we have
-      # a solution to that problem.
-      add_product_to_cart order, product_with_fee
-      visit spree.cart_path
-      register_page = window_opened_by { click_link "Register here" }
-      within_window(register_page) { expect(page).to have_content "Welcome" }
-    end
-
     describe "product description" do
       it "does not link to the product page" do
         add_product_to_cart order, product_with_fee, quantity: 2
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -20,6 +20,17 @@ feature "full-page cart", js: true do
       set_order order
     end
 
+    it "shows the right registration link" do
+      # We have an issue with the registration link within Spree controllers.
+      # The `registration_path` helper resolves to `/signup` due to
+      # spree_auth_device > config > routes.rb. This spec verifies that we have
+      # a solution to that problem.
+      add_product_to_cart order, product_with_fee
+      visit spree.cart_path
+      register_page = window_opened_by { click_link "Register here" }
+      within_window(register_page) { expect(page).to have_content "Welcome" }
+    end
+
     describe "product description" do
       it "does not link to the product page" do
         add_product_to_cart order, product_with_fee, quantity: 2
--- a/spec/controllers/spree/orders_controller_spec.rb
+++ b/spec/controllers/spree/orders_controller_spec.rb
@@ -116,7 +116,7 @@ describe Spree::OrdersController, type: :controller do
       expect(flash[:info]).to eq("The hub you have selected is temporarily closed for orders. Please try again later.")
     end
 
-    describe "when an item has insufficient stock" do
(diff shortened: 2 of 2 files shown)


spec/features/consumer/shopping/cart_spec.rb
```diff
@@ -20,6 +20,17 @@
...
+    it "shows the right registration link" do
+      # We have an issue with the registration link within Spree controllers.
+      # The `registration_path` helper resolves to `/signup` due to
+      # spree_auth_device > config > routes.rb. This spec verifies that we have
+      # a solution to that problem.
+      add_product_to_cart order, product_with_fee
+      visit spree.cart_path
+      register_page = window_opened_by { click_link "Register here" }
+      within_window(register_page) { expect(page).to have_content "Welcome" }
+    end
```
sauloperez: It feels a bit too much to boot a browser to only check if that a link points to `/register`. Since in this case, we don't care about the user input or interaction involving different pages but the target of a link why don't we try using view specs? What are your thoughts?

mkllnk: I'm not 100% sure but I don't think the problem is with the view. I think the problem is with the controller and which helpers it has loaded. A controller spec may work. I can give that a go and see if we can reproduce the bug that way. Is it worth the dev time?

luisramos0:
wait, now I wonder how can a new spec fix a bug :-)
is #2873 already fixed?

mkllnk:
Yes, the bug is fixed. This spec just covers it in case it re-appears or we want to get rid of our workaround. I tested the spec. It fails using `registration_path`.

spec/features/consumer/shopping/cart_spec.rb
mkllnk: @sauloperez Please have a look at my new commit. I replaced the feature spec with two controller specs. It saves us three seconds and it's much more specific, closer to the problem.

Issue #3565, opened 2019-03-01
[Spree 2.1] Prepare and estimate upgrade to Spree 2.1


luisramos0:
### What we should change and why (this is tech debt)
We will upgrade to spree 2.1 at some point probably soon. v2.1 brings rails 4.

See discourse thread for context:
https://community.openfoodnetwork.org/t/spree-upgrade-go-live-on-v2-0-4-or-move-further-to-2-1-0/1534/4

We need to make the spree 2-1-0-stable  branch work (bundle install is broken) and then branch ofn 2-1-stable (from ofn 2-0-stable) and make it use this spree 2-1-0-stable.

The objective of this issue is to have bundle install working on both spree and ofn 2-1 and we see a build in semaphore for it.

The state of the build will help us estimate the effort involved in getting 2-1 live.

### Impact and timeline
Rails 4 ❤️

luisramos0:
Current status:
I have this OFN 2-1-stable branch: https://github.com/luisramos0/openfoodnetwork/commits/2-1-stable

It's only one commit updating to spree 2-1-0-stable, rails 4 and a few others.

Currently, trying to run
`bundle update rails sass-rails rails-i18n`

I am stuck with errors like this. I dont know how to fix these:
```
Bundler could not find compatible versions for gem "rails":
  In Gemfile:
    rails (~> 4.0.0)

    db2fog was resolved to 0.9.0, which depends on
      rails (< 5.0, >= 3.2.0)

    deface (= 1.0.0) was resolved to 1.0.0, which depends on
      rails (>= 3.1)

Bundler could not find compatible versions for gem "rails-i18n":
  In Gemfile:
    rails-i18n (~> 4.0.0)

    spree_i18n was resolved to 1.0.0, which depends on
      rails-i18n
```

sauloperez:
these need to be solved by appending the gems to the `bundle update`. It happened with v2 as well. I'm taking a look right now.

sauloperez:
After a few tries I decided to give a plain `bundle update` a go. https://gist.github.com/sauloperez/5aee4b6339f9d3e214560a805f76de04 is the output that I got. It will tell us what are the gems stopping us from upgrading.

luisramos0:
I have just learned that:
‘~> 2.2’  equals ['>= 2.2.0', '< 3.0']
while
‘~> 2.2.0’ equals ['>= 2.2.0', '< 2.3.0']

luisramos0:
[...]

PR #3627, opened 2019-03-20
Work on the Gemfile to make bundle update pass

sauloperez:
#### What? Why?

This comments out a number of gems that either are fetch from out-of-date forks, that are meant for Rails 3 only or that do not support Rails 4.

There a lot of work still to be done to have a working Gemfile.

It is the very first step towards v2.1. What is supposed to be the next upgrade we do after v2.


Gemfile
```diff
@@ -21,16 +21,16 @@ gem 'pg'
...
 # Our branch contains two changes
 # - Pass customer email and phone number to PayPal (merged to upstream master)
 # - Change type of password from string to password to hide it in the form
-gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "2-0-stable"
+# gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "2-0-stable"
 gem 'stripe', '~> 4.5.0'
 # We need at least this version to have Digicert's root certificate
 # which is needed for Pin Payments (and possibly others).
-gem 'activemerchant', '~> 1.78'
+# gem 'activemerchant', '~> 1.78'
```
sauloperez: If this particular version is not compatible with Spree we will need to first understand what is it that we need from activemerchant (I don't remember) and possibly bring in that Digicert's root certificate from somewhere else.

Gemfile
```diff
@@ -59,7 +58,7 @@ gem 'rabl'
 # Once Rails is updated to 5.x we should bump directly to 0.10.x
 gem "active_model_serializers", "0.8.4"
 gem 'oj'
-gem 'deface', '1.0.0'
+gem 'deface'
```
sauloperez: I even wonder if we need to specify deface ourselves. It is Spree itself that depends on it on a particular version.

sauloperez:
What should we do with this @luisramos0 ? Feel free to assign yourself if you want work on it. I unassign myself.

luisramos0:
this PR can be closed as I have put all this in 2-1-stable already.

luisramos0:
I am blocked here: https://github.com/openfoodfoundation/openfoodnetwork/issues/3565#issuecomment-475419217

can you have a quick look, maybe you will understand the errors in the build and why I dont see them locally?

PR #3634, opened 2019-03-21
[Spree Upgrade] Fix new sample data in v2

luisramos0:
#### What? Why?

This PR fixes sample data in v2.

#### What should we test?
You should be able to load sample data. We can re-test sample data in v2 and make sure it is working.

#### Dependencies
This PR includes the commit from #3633 so that we avoid conflicts.


lib/tasks/sample_data/product_factory.rb
```diff
@@ -74,8 +74,8 @@ def create_product(hash)
       variant_unit: "weight",
       variant_unit_scale: 1,
       unit_value: 1,
-      on_demand: true
+      on_demand: true,
+      shipping_category: Spree::ShippingCategory.find_or_create_by_name('Default')
     )
-    Spree::Product.create_with(params).find_or_create_by_name!(params[:name])
```
mkllnk: I think you are forgetting to create the products here.

lib/tasks/sample_data/product_factory.rb
```diff
@@ -74,8 +74,10 @@ def create_product(hash)
       variant_unit: "weight",
       variant_unit_scale: 1,
       unit_value: 1,
-      on_demand: true,
       shipping_category: Spree::ShippingCategory.find_or_create_by_name('Default')
     )
+    product = Spree::Product.create_with(params).find_or_create_by_name!(params[:name])
```
mkllnk: Ah, okay, you create the product in a later commit again. I find it easier when I can read pull requests commit by commit.

mkllnk:
I just had to use this one and it worked. :-)

lib/tasks/sample_data/product_factory.rb
luisramos0: ah, sorry, it was my mistake when reviewing the commit history...

lib/tasks/sample_data/shipping_method_factory.rb
```diff
@@ -47,7 +47,8 @@ def create_delivery(enterprise)
   def create_shipping_method(enterprise, params)
     params[:distributor_ids] = [enterprise.id]
     method = enterprise.shipping_methods.new(params)
-    method.zone = zone
+    method.zones << zone
+    method.shipping_categories << Spree::ShippingCategory.find_or_create_by_name('Default')
```
sauloperez: What about abstracting this into a class method that saves us from typing the name every time as we did with StockLocation? we reduce the chances of introducing a bug a great deal as well as having fewer things to think about.

luisramos0: I considered doing that but I decided not to do it because this is very different from default stock_location. Default stock location is seed data, it's very important for the system to work.
This DefaultShippingCategory is totally different, this is a value we use to migrate v1 instances with existing data to v2, it's not going to be reused. And here is really just a sample shipping category. Here it could be called "by Air".

luisramos0: so, having something like default stock location is tricky to mantain and can induce errors (our life would be easier without it) so by not doing the same here we are not creating another similar fragile situation or source of trouble.

luisramos0:
I merged 2-0-stable into this branch to resolve conflicts

RachL:
I've tested checkout + un created an enterprise with a product and an open OC and managed to checkout. This is good to go then!

PR #3637, opened 2019-03-21
[Proof of Concept] Different bill and ship addresses for subscriptions in specs

kristinalim:
Different bill and ship addresses for subscriptions in specs

#### What? Why?

Relates to #3605 

This PR is to check if there are other specs aside from those in `OrderSyncer` that would fail if we make bill and ship addresses in sample subscriptions different from each other.

This forces us to write correct assertions when adding specs affected by subscription addresses.

See this related review comment: https://github.com/openfoodfoundation/openfoodnetwork/pull/3605#discussion_r265896370


kristinalim:
I created this PR to see if other specs would fail with this change. It looks like only `OrderSyncer` is affected, which I was already aware about.

Closing this PR now.

PR #3639, opened 2019-03-22
Fix variant cache refreshing on delete

### LINKED ISSUE #993: RuntimeError in app/jobs/products_cache_integrity_checker_job.rb:6
## Error in OpenFoodNetwork

**RuntimeError** in **app/jobs/products_cache_integrity_checker_job.rb:6**
Products JSON differs from cached version for distributor: 178, order cycle: 199

[View on Bugsnag](https://bugsnag.com/yaycode/openfoodnetwork/errors/56cfa39507141bdbda05485f?event_id=573535363ce85ece7172ac7c)
## Stacktrace

```
app/jobs/products_cache_integrity_checker_job.rb:6 - perform
script/delayed_job:5 - <main>
```

[View full stacktrace](https://bugsnag.com/yaycode/openfoodnetwork/errors/56cfa39507141bdbda05485f?event_id=573535363ce85ece7172ac7c)

### LINKED ISSUE #3629: Deleting a variant has no direct effect on a shopfront
## Description
If you delete a variant, it keeps showing up on the shopfront until you save a second time changes on your product.

If someone attempts to add this deleted variant to cart it gets a infinite loader

I saw this while testing #3619 but could reproduced it on French production

## Steps to Reproduce
1. I deleted a variant that already existed on a staging server
2. I saw that it was correctly deleted from the OC but not the shopfront (I cleared my browser cache several times)
3. I've added a new variant to my product. At this stage - as I saved again - the variant previously deleted disappear from the shop. The new variant showed up correctly
4. I deleted this new variant. Again it did not showed up in the OC, but on the shopfront yes.
5. I change the price of another variant, hit save. Then my variant is correctly deleted from shop front.

I reproduce all this on French product but this time I also tried to add the deleted variant to card. I got an infinite loader on the add to cart button:

image

Reloading the page stop the loader, but my variant is still showing up. It's the v2 variant bottom page here: https://www.openfoodfrance.org/hub-demo-open-food-france/shop

## Animated Gif/Screenshot
image

image

## Severity

## Your Environment
* Version used: v1.28
* Browser name and version: Firefox 65

sauloperez:
#### What? Why?

Closes #3629 and closes #993.

As @mkllnk pointed out in our own implementation of variant destroy we skip all callbacks by calling ActiveRecord's [#delete](https://apidock.com/rails/v3.2.13/ActiveRecord/Relation/delete) instead of `#destroy`. 

The problem laid in the fact that the logic to refresh the cache was meant to be triggered by an `#around_destroy` and so it wasn't called.

In order to confidently fix the bug I first covered the controller with tests. I :heart: TDD. As a side note, having cache enabled in dev helped a lot :v: .

#### What should we test?

The exact scenario described in #3629. Deleting a variant from the product's variant section should remove it from any existing OC, which should, in turn, remove it from the shopfront.

#### Release notes

Update the cache when a variant listed in an order cycle is removed. This was the cause of removed variants still being shown in the shopfront.

Changelog Category: Fixed

### DIFF
--- a/app/controllers/spree/admin/variants_controller_decorator.rb
+++ b/app/controllers/spree/admin/variants_controller_decorator.rb
@@ -10,7 +10,8 @@ Spree::Admin::VariantsController.class_eval do
 
   def destroy
     @variant = Spree::Variant.find(params[:id])
-    @variant.delete # This line changed, as well as removal of following conditional
+    @variant.delete_and_refresh_cache
+
     flash[:success] = I18n.t('notice_messages.variant_deleted')
 
     respond_with(@variant) do |format|
--- a/app/controllers/spree/admin/variants_controller_decorator.rb
+++ b/app/controllers/spree/admin/variants_controller_decorator.rb
@@ -15,7 +15,7 @@ Spree::Admin::VariantsController.class_eval do
 
     respond_with(@variant) do |format|
       format.html { redirect_to admin_product_variants_url(params[:product_id]) }
-      format.js   { render_js_for_destroy }
+      format.js { render_js_for_destroy }
     end
   end
 
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -118,6 +118,11 @@ Spree::Variant.class_eval do
     end
   end
 
+  # Deletes the record, skipping callbacks, but it also refreshes the cache
+  def delete_and_refresh_cache
+    destruction { delete }
+  end
+
   private
 
   def update_weight_from_unit_value
--- a/spec/controllers/spree/admin/variants_controller_spec.rb
+++ b/spec/controllers/spree/admin/variants_controller_spec.rb
@@ -35,6 +35,58 @@ module Spree
           assigns(:variants).should match_array [v1, v2]
         end
       end
+
+      describe '#destroy' do
+        let(:variant) { create(:variant) }
+
+        context 'when requesting with js' do
+          before do
+            allow(Spree::Variant).to receive(:find).with(variant.id.to_s) { variant }
+            allow(variant).to receive(:delete)
+          end
+
+          it 'deletes the variant' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'js'
+            expect(variant).to have_received(:delete)
+          end
+
+          it 'shows a success flash message' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'js'
+            expect(flash[:success]).to eq(I18n.t('notice_messages.variant_deleted'))
+          end
+
+          it 'renders spree/admin/shared/destroy' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'js'
+            expect(response).to render_template('spree/admin/shared/_destroy')
+          end
+        end
+
+        context 'when requesting with html' do
+          before do
+            allow(Spree::Variant).to receive(:find).with(variant.id.to_s) { variant }
+            allow(variant).to receive(:delete)
+          end
+
+          it 'deletes the variant' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'html'
+            expect(variant).to have_received(:delete)
+          end
+
+          it 'shows a success flash message' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'html'
+            expect(flash[:success]).to eq(I18n.t('notice_messages.variant_deleted'))
+          end
+
+          it 'redirects to admin_product_variants_url' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'html'
+            expect(response).to redirect_to(
+              controller: 'spree/admin/variants',
+              action: :index,
+              product_id: variant.product.permalink
+            )
+          end
+        end
+      end
     end
   end
 end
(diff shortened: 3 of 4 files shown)


app/controllers/spree/admin/variants_controller_decorator.rb
```diff
@@ -10,12 +10,13 @@ def search
 
   def destroy
     @variant = Spree::Variant.find(params[:id])
-    @variant.delete # This line changed, as well as removal of following conditional
+    @variant.delete_and_refresh_cache
```
sauloperez: I'm not a big fan of this but I consider it a good-enough solution given the changes pointed out by @mkllnk in v2 regarding this. #conservative.

spec/controllers/spree/admin/variants_controller_spec.rb
```diff
@@ -59,6 +59,19 @@ module Admin
             spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'js'
             expect(response).to render_template('spree/admin/shared/_destroy')
           end
+
+          it 'refreshes the cache' do
```
mkllnk: If you expect these tests to fail, I would set them to `pending` in this commit and change them to `it` in the fixing commit. It makes it even clearer what exactly the commit fixes. And it makes the specs always pass which helps when running git bisect. Do you agree?

app/models/spree/variant_decorator.rb
```diff
@@ -118,6 +118,11 @@ def refresh_products_cache
     end
   end
 
+  # Deletes the record, skipping callbacks, but it also refreshes the cache
+  def delete_and_refresh_cache
+    destruction { delete }
```
luisramos0: this method name destruction... imo it's really not informative and could be better named but it makes me laugh, could be called implosion instead :-D

spec/controllers/spree/admin/variants_controller_spec.rb
```diff
@@ -35,6 +35,84 @@ module Admin
...
+
+          it 'shows a success flash message' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'js'
+            expect(flash[:success]).to eq(I18n.t('notice_messages.variant_deleted'))
+          end
+
+          it 'renders spree/admin/shared/destroy' do
+            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'js'
+            expect(response).to render_template('spree/admin/shared/_destroy')
+          end
```
luisramos0: why are these 3 specs above not one spec with 3 assertions? is it your testing style decision? I am curious...

spec/controllers/spree/admin/variants_controller_spec.rb
sauloperez: I've never thought of this approach. Interesting.

app/models/spree/variant_decorator.rb
sauloperez: lol. I was just trying to follow the previous naming but I can improve it of course.

spec/controllers/spree/admin/variants_controller_spec.rb
sauloperez: Yep, stems from http://www.betterspecs.org/#single and the conventions I used to work with in previous teams I guess.

luisramos0: ah, ok. but that's for unit testing, right? this is controller level testing, and in this case, it doesnt bring much added readability.
I am just trying it out to find out what I prefer. I think I prefer this faster shorter version:
```
        context 'when requesting with js' do
          before do
            allow(Spree::Variant).to receive(:find).with(variant.id.to_s) { variant }
            allow(variant).to receive(:delete)
          end

          it 'deletes the variant' do
            spree_delete :destroy, id: variant.id, product_id: variant.product.permalink, format: 'js'

            expect(variant).to have_received(:delete)
            expect(flash[:success]).to eq(I18n.t('notice_messages.variant_deleted'))
            expect(response).to render_template('spree/admin/shared/_destroy')
          end
```

luisramos0: thanks for the link!

mkllnk: I quite like to have one assertion per test case. When some of them fail, you get a better picture of what's broken and what's still working. They are also different features. You want to delete the variant and you want to inform the user about it. In this case I agree that your version is easier to read. -> trade off :man_shrugging:

sauloperez:
@luisramos0 I gave a try at `git merge fix-variant-cache-refreshing-on-delete` into `2-0-stable` and this is the conflict I encountered. I hope it helps.

```
diff --cc app/controllers/spree/admin/variants_controller_decorator.rb
index c422ae8b4,a87155b30..000000000
--- a/app/controllers/spree/admin/variants_controller_decorator.rb
+++ b/app/controllers/spree/admin/variants_controller_decorator.rb
@@@ -22,11 -10,9 +22,17 @@@ Spree::Admin::VariantsController.class_

    def destroy
      @variant = Spree::Variant.find(params[:id])
++<<<<<<< HEAD
 +    if VariantDeleter.new.delete(@variant) # This line changed
 +      flash[:success] = Spree.t('notice_messages.variant_deleted')
 +    else
 +      flash[:success] = Spree.t('notice_messages.variant_not_deleted')
 +    end
++=======
+     @variant.delete_and_refresh_cache
+
+     flash[:success] = I18n.t('notice_messages.variant_deleted')
++>>>>>>> fix-variant-cache-refreshing-on-delete

      respond_with(@variant) do |format|
        format.html { redirect_to admin_product_variants_url(params[:product_id]) }
```

RachL:
Drumroll! this works :) https://docs.google.com/document/d/1oIYdzyw_oJWLz75usb8X75l5hliWekiT303w_kKQ6CQ/edit

luisramos0:
@sauloperez just to double check: after merging master to 2-0, this PR should be reverted (in some git way) except for the tests in variants_controller_spec.rb, correct?

sauloperez:
Yep, exactly as I mention in my comment.

Issue #3625, opened 2019-03-20
[Spree upgrade] Cancelled state translation missing


RachL:
## Description
When an order has been cancelled, the state is not correctly translated.

## Steps to Reproduce
1. Create an order
2. Go to the admin order page and cancel it
3. The state of the order should change to cancelled, thus showing the missing translation shipment_states.canceled (whatever the language chosen)

## Animated Gif/Screenshot
![image](https://user-images.githubusercontent.com/6525576/54683408-e6c97380-4b11-11e9-87a9-5f7899d00d98.png)

## Severity
bug-s3: a feature is broken but there is a workaround

## Your Environment
* Version used: v2
* Browser name and version: Firefox 65
* Operating System and version (desktop or mobile): Ubuntu 18.10

PR #3641, opened 2019-03-22
[Spree Upgrade] Update spree revision to include last package_factory fix

### LINKED ISSUE #3607: [Spree Upgrade] Order edit page is not filtering shipping methods by the order distributor
## Description
When editing an order we can see the full list of shipping methods in the system.

## Expected Behavior
We should see only shipping methods of the order's distributor.

## Actual Behaviour
We see the full list of shipping methods in the system.

## Steps to Reproduce
Open any order, edit shipping method and check the list.

## Animated Gif/Screenshot
Example from my local instance with sanitized AUS live data (I see 500+ shipping methods for any order)
Screenshot 2019-03-15 at 12 40 50

## Context
Found while testing #3527

## Severity

luisramos0:
#### What? Why?

Closes #3607

ok, I figured this one out. With the new spree revision, the package shipping methods get filtered by distributor, so to get a green green build, we need to adapt the factories and specs to always set the order distributor on the shipping methods used in the test. This was already partially done in #3604 

#### What should we test?
For a green build, this PR requires #3604 and also #3667 (due to duplicate order-charges html id).

We need to verify issue #3607 is fixed.

#### Dependencies
This PR can only be tested and merged after 3604 and 3667 but can be reviewed now.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -31,7 +31,7 @@ GIT
 
 GIT
   remote: https://github.com/openfoodfoundation/spree.git
-  revision: f55722b38db7e706a8521c9091a0e00119bb4d20
+  revision: 060d0a5d8b31f68990fa792e0e88910c742d9a96
   branch: 2-0-4-stable
   specs:
     spree (2.0.4)
--- a/spec/controllers/spree/api/shipments_controller_spec.rb
+++ b/spec/controllers/spree/api/shipments_controller_spec.rb
@@ -25,7 +25,9 @@ describe Spree::Api::ShipmentsController, type: :controller do
     let(:error_message) { "broken shipments creation" }
 
     before do
-      order.update_attribute(:ship_address_id, order_ship_address.id)
+      order.update_attribute :ship_address_id, order_ship_address.id
+      order.update_attribute :distributor, variant.product.supplier
+      shipment.shipping_method.distributors << variant.product.supplier
     end
 
     sign_in_as_admin!
@@ -56,6 +58,7 @@ describe Spree::Api::ShipmentsController, type: :controller do
       it 'updates existing shipment with variant override if an VO is sent' do
         hub = create(:distributor_enterprise)
         order.update_attribute(:distributor, hub)
+        shipment.shipping_method.distributors << hub
         variant_override = create(:variant_override, hub: hub, variant: variant)
 
         spree_post :create, params
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -219,7 +219,9 @@ feature %q{
     end
 
     feature "viewing the edit page" do
-      let!(:different_shipping_method) { create(:shipping_method, name: "Different Shipping Method") }
+      let!(:shipping_method_for_distributor1) { create(:shipping_method, name: "Normal", distributors: [distributor1]) }
+      let!(:different_shipping_method_for_distributor1) { create(:shipping_method, name: "Different", distributors: [distributor1]) }
+      let!(:shipping_method_for_distributor2) { create(:shipping_method, name: "Other", distributors: [distributor2]) }
 
       background do
         Spree::Config[:enable_receipt_printing?] = true
@@ -227,10 +229,12 @@ feature %q{
         distributor1.update_attribute(:abn, '12345678')
         @order = create(:order_with_taxes,
                         distributor: distributor1,
+                        ship_address: create(:address),
                         product_price: 110,
                         tax_rate_amount: 0.1,
                         tax_rate_name: "Tax 1")
         Spree::TaxRate.adjust(@order)
+        @order.update_shipping_fees!
 
         visit spree.edit_admin_order_path(@order)
       end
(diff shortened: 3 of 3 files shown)


Gemfile.lock
```diff
@@ -59,7 +59,7 @@ GIT
       activemerchant (~> 1.34)
       acts_as_list (= 0.2.0)
       awesome_nested_set (= 2.1.5)
-      aws-sdk (~> 1.11.1)
+      aws-sdk (~> 1.3.4)
```
sauloperez: that is quite a big change. Have you checked why it happened? I find it weird to have these Gemfile.lock changes at this point. The revision moves us a bunch of commits forward only, right? :thinking:

luisramos0: this is just the upload/download of images from s3 right?
we are moving one commit forward only.
I dont understand why this happens.
shall I lock 1.11 on our side, in our Gemfile?

sauloperez: > this is just the upload/download of images from s3 right?

I guess so as it's defined as a `spree_core` dependency. I can't think of any other Spree feature that might need aws.

> I dont understand why this happens.

Me neither. I'll try to reproduce it in my machine.

sauloperez: Ok, same outcome on my machine. For some reason when resolving spree's dependencies again (through a `bundle update spree` bundler thinks these versions are better suited 🤷‍♀️ . So locking aws-sdk to 1.11 or above sounds good IMO.

sauloperez: I tried to do it myself but I found out why bundler was doing it in the first place:

```
Bundler could not find compatible versions for gem "aws-sdk":
  In Gemfile:
    aws-sdk (>= 1.11.1)

    spree was resolved to 2.0.4, which depends on
      spree_core (= 2.0.4) was resolved to 2.0.4, which depends on
        aws-sdk (~> 1.3.4)
```

So it turns out the Gemfile.lock was wrong. Maybe a wrongly solved merge conflict?

Nothing to do here then.

luisramos0: ah, nice, thanks for having a look!
spree 204 depends on 1.3.4 in spree repo [here](https://github.com/spree/spree/blob/v2.0.4/core/spree_core.gemspec#L23) and in ofn spree repo [here](https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/core/spree_core.gemspec#L23).
What I just found is that our current spree branch step-6a is on 1.11.1 since a [long time ago](https://github.com/openfoodfoundation/spree/commit/6c98889f7453983da874b89d842e1945438bb6d0).
So I think we need to open a new PR in ofn/spree/2-0-4 and change it to depend on 1.11.1
Then update this PR to use that new revision of ofn/spree/2-0-4 and keep 1.11.1. Makes sense?

luisramos0: ok, I went for it.
There's only the update of the spree revision in this PR now.
I'll update aws-sdk in a separate PR.

sauloperez: Glad that you carried on with the investigation. Sounds good :+1:

luisramos0: yeah, this is the only one breaking a fresh bundle install

sauloperez:
I gave CI a second chance by rebuilding it

luisramos0:
The build is good, I tested it thoroughly.
The last one was only broken in one orders_spec because I dont want to put #3667 in this PR.
All good.

luisramos0:
this is fixed in #3667 

 1) 
    As an administrator
    I want to manage orders
 as an enterprise manager viewing the edit page shows the order tax adjustments
     Failure/Error:
       within('tbody#order-charges') do
         expect(page).to have_selector "td", match: :first, text: "Tax 1"
         expect(page).to have_selector "td.total", text: Spree::Money.new(10)
       end
     
     Capybara::Ambiguous:
       Ambiguous match, found 2 elements matching visible css "tbody#order-charges"
     # ./spec/features/admin/orders_spec.rb:274:in `block (4 levels) in <top (required)>'

luisramos0:
great, thanks for the quick reviews guys!

both dependencies are merged now. I rebased so we have a green build.

moving to test ready.

luisramos0:
@RachL apart from testing #3607, because this is rebased it includes #3668 and so this is the right PR to test #3625 as well
and maybe we can check that the PRs that were tested in master are also working in v2: #3662, #3646 and #3644

luisramos0:
@RachL staging Katuma is ready on v2. I ended up using this PR because it is the last major v2 PR. and I wanted to test the data migration with live data and all major changes in, it worked perfectly 🎉

RachL:
@luisramos0 thx! I've checked them all and everything is fine. I didn't see the spanish translation on "producer profile" but I guess it is not done on transifex. Anyway it shouldn't postpone this PR to be merged. So this is ready to go https://docs.google.com/document/d/14fB7poxkCBKRX2o7wOivSG32IRsPEOWtbUzxmf7LILs/edit#

luisramos0:
nice!! thanks!
"I've checked them all" 💪

Issue #3162, opened 2018-12-01
[i18n] Untranslated "producer profile" in pack selection page (front end)


myriamboure:
![capture du 2018-12-02 00-13-02](https://user-images.githubusercontent.com/8135376/49333884-6107d380-f5c7-11e8-97a7-7e24eb69dc98.png)

I guess also "PRODUCER PROFILE" here should be replaced by the name of the enterprise that was just created...

amers185:
If this issue is still open, I would like to work on it.

sigmundpetersen:
Please have a look at our wiki page on i18n for an introduction on how we work on translations https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29

luisramos0:
fyi, this page will be tested in https://github.com/openfoodfoundation/openfoodnetwork/pull/3609
I dont think 3609 fixes this bug though.

amers185:
@sigmundpetersen Thank you. I am familiar with the project because I was part of a group of college students who recently worked on the OFN documentation issue. I want to continue contributing in an individual capacity even though our group project ended. @luisramos0 Shall I continue working on the issue even if it will be tested in 3609?

luisramos0:
hi @amers185 
nice! thanks for your interest!

I just checked. you can fix this, it's independent of #3609
I leave you a hint ;-) https://github.com/openfoodfoundation/openfoodnetwork/blob/fb29a7c7c210e22863a6bbd4a1619da0d1c75b19/app/views/admin/enterprises/welcome.html.haml#L7

amers185:
Ah thank you. So, on a high-level, this is not a missing key issue? Rather its more of not calling the right profile?

luisramos0:
no, you have to make the code call the t function so that it looks up the translation key.

amers185:
I feel rather dumb for asking this since I have only really inserted your code but this modification: "#{"producer " - if @enterprise.is_primary_producer? t('.producer_profile') - else t('.profile')}profile" in my IDE gives me an expected expression error at the dash before else and an expected herdoc end error? I have looked over different ways to try nested loops within tags for HAML but I am not quite share what I am doing wrong here. Further, how can I proceed to check for changes I make on the local host? As in, is there a go to strategy to navigate the local host given I do not know where this page is on the actual website? I cannot even access the french version on my local host. Again, my apologies for all these questions.

luisramos0:
[...]

PR #3644, opened 2019-03-22
Fixed producer profile translation bug

amers185:
#### What? Why?
 Untranslated "producer profile" in pack selection page (front end) 
Closes #[#3162]

Producer profile was hard coded as a string in English. I put the correct key-value pairs in the english file and then made t function call on them with respect to their status as primary producer. 

#### What should we test?

Once transifex does it magic. It will be important to check if the correct key is called by the t function for non-english websites.

#### Release notes

String interpolation cannot be relied upon.

 Fixed

### DIFF
--- a/app/views/admin/enterprises/welcome.html.haml
+++ b/app/views/admin/enterprises/welcome.html.haml
@@ -4,7 +4,8 @@
     %p
       = t('.welcome_text')
       %strong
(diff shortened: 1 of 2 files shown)


mkllnk:
You got two positive reviews, excellent. One of our testers will now look at it on a staging server and tell you if it works as expected.

RachL:
I don't see anything broken here, so this is good to go.

image

PR #3646, opened 2019-03-23
3528 Scope Customer Totals report also by variant (not just product) and use variant SKU

### LINKED ISSUE #3528: SKU showing up incorrectly on different reports for variant products
## Description

The reports are showing up different SKU values for the same product.

The product will have a code of XXXX, within the product report this is shown correctly, but in the order cycle customer totals this is incorrect showing up as XXXXY - two different reports of the order cycle customer total have been run, and both show different SKUs which are not the true SKU

## Expected Behavior

SKU should be the same for a product everywhere

## Actual Behaviour

The product will have a code of XXXX, within the product report this is shown correctly, but in the order cycle customer totals this is incorrect showing up as XXXXY - two different reports of the order cycle customer total have been run, and both show different SKUs which are not the true SKU

This is happening with hodmedods MASTER in UK production

## Steps to Reproduce

1. Look at a the SKU for Camelina   Seed, Non-Organic - 10 x 250g (250g x 10) - NGCSCX (left)
2.  Customer totals report run on the 12th states NGCSR2 (top right)
3. Same report run today only sates NGCS (middle right)
4. All products report states NGCSCX which is correct (bottom right)

## Animated Gif/Screenshot

image

## Context

SKU is a key field that allows OFN to operate with other systems. If it can't be reliable then it renders the reports useless, making OFN impossible to use for users that need to use reports to interoperate with other systems in their businesses.

## Severity

## Your Environment

* Version used:
* Operating System and version (desktop or mobile): windows 10

kristinalim:
Scope Customer Totals report also by variant (not just product) and use variant SKU.

(This includes the work of @mkllnk in his fork. Thank you, @mkllnk!)

#### What? Why?

Closes #3528

The report groups the line item data by product, among other attributes. We need to display the variant SKU, not the product SKU, so we need to group these by variant ~instead~ also.

#### What should we test?

1. Create two customers for a hub with an open OC. Make sure they have different names.
2. Prepare a purchasable Product A with two variants, and a purchasable Product B with one variant.
3. For Customer A, create an order purchasing all three variants.
4. For Customer B, create an order purchasing Variant A of Product A.
5. For Customer B, create another order purchasing Variant B of Product A.
6. Generate the Customer Totals report. Note the SKU and total for these orders. These should be correct.

#### Release notes

- In the Customer Totals report, separate rows for different variants and use the variant SKU.

Changelog Category: Fixed


kristinalim:
I updated the description of this commit: https://github.com/openfoodfoundation/openfoodnetwork/pull/3646/commits/46e7090d118df379511db3128a26cbf5e3038c1b, and also the title and description of this PR, to reflect that the grouping is by variant _in addition to_ (not "instead of") by product. No code changes.

In an earlier version, I just replaced the grouping by product to grouping by variant. The current version uses both because both are important.

RachL:
I hate myself. I literally spent 20 minutes on "I don't see SKU in the report, am I on the correct report?" Only to discover that SKU is called "product code" in the export 🤦‍♀️ 

Anyway this is good to go :)

https://docs.google.com/document/d/1cb3tylYDFc6YJcsTNsHrb14zc4tLbUE0wxjCNjBSkrs/edit#heading=h.hj26umg3kyxb

PR #3648, opened 2019-03-24
[Spree Upgrade] Fix "back to list" button overlap and remove configurations menu from payments methods list

### LINKED ISSUE #3543: [Spree upgrade] Buttons overlap when editing payment methods (French only)
## Description
When a producer adds a new payment method there is an overlap of one button at the end of the process. It is the same when editing an existing method.

EDIT: this happens only when displaying in French

## Steps to Reproduce and Animated Gif/Screenshot
1. Log in as a producer with French language activated
2. Create a new payment method or edit an existing one
3. When the page loads, you should see the button "go back to the list of payment methods" overlapping "new payment method" :

## Severity

## Your Environment
* Version used: ??

luisramos0:
#### What? Why?

Closes #3543

In this PR we fix the button layout  issue 3543 and we remove the configurations menu from the payments methods list.
@RachL I think you said it was good to have the configurations menu back but we can't because this page is also seen by non-admins coming from the enterprise management page (they would see the menu and all the links would take the user to the "no authorization" page). You can create an issue to detect if user is admin or not and show the menu if user is admin, OR show the menu if user is coming from the configurations and not from enterprise management.

#### What should we test?
- "back to list" buttons do not overlap.
- payment methods list doesnt show the configurations list.

### DIFF
--- a/app/assets/stylesheets/admin/components/page_actions.css.scss
+++ b/app/assets/stylesheets/admin/components/page_actions.css.scss
@@ -1,0 +1,10 @@
+.page-actions {
+  li {
+    margin-top: 1px;
+    margin-bottom: 1px;
+  }
+
+  a {
+    display: inline-block;
+  }
+}
--- a/app/overrides/spree/admin/payment_methods/index/remove_configuration_sidebar.deface
+++ b/app/overrides/spree/admin/payment_methods/index/remove_configuration_sidebar.deface
@@ -1,1 +1,1 @@
-remove "code[erb-loud]:contains(\"render :partial => 'spree/admin/shared/configuration_menu'\")"
+remove "erb[loud]:contains(\"render :partial => 'spree/admin/shared/configuration_menu'\")"


app/assets/stylesheets/admin/components/page_actions.css.scss
```diff
@@ -0,0 +1,7 @@
+.page-actions {
+  li {
```
sauloperez: isn't there a class attribute we could use instead? As far as I know, coupling to a particular DOM element is not very changeable.

luisramos0: no, the li entries don't have any class in the spree code. I think this is fine. page actions are a ul with li entries, and we want to change all those li entries.

kristinalim: This is how it would look in narrower screens, or using languages where the translation is even longer:

20190328031740-header-buttons-medium-screens

Changing the `display` property of the button from `inline` to `inline-block` could really help make this more responsive. I recommend this instead:

```css
.page-actions {
  li {
    margin-top: 1px;
    margin-bottom: 1px;
  }

  a {
    display: inline-block;
  }
}
```

The 1px margin (because we're not trying to be pixel-perfect) is to make sure there is space between rows of buttons.

Then this is how the buttons would look:

20190328033019-header-action-buttons-inline-block

app/overrides/spree/admin/payment_methods/index/remove_configuration_sidebar.deface
```diff
@@ -1 +1 @@
-remove "code[erb-loud]:contains(\"render :partial => 'spree/admin/shared/configuration_menu'\")"
\ No newline at end of file
+remove "erb[loud]:contains(\"render :partial => 'spree/admin/shared/configuration_menu'\")"
```
kristinalim: For other reviewers, it's related to this: https://github.com/spree/deface#upgrading-from-09-to-10 I had to look it up.

app/assets/stylesheets/admin/components/page_actions.css.scss
luisramos0: yeah, thanks @kristinalim 
now I ask myself, why did I use padding and not margin?
the inline-block tip is 💯 
I replaced my commit with your code 👍

RachL:
> I think you said it was good to have the configurations menu back but we can't because this page is also seen by non-admins coming from the enterprise management page (they would see the menu and all the links would take the user to the "no authorization" page).

Ok this makes sense. To be honest I would prefer us to work on the steps sometimes : to me it is illogical that the enterprise manager ends up on the global payment method page. When the enterprise users clicks on the go back button he should be redirected to the admin/enterprises/enterprisename/edit#/payment_methods page. Thus super admin and admin have two different pages, and we reduce the number of clicks for the manager. And no need to hide the left menu for the super admin. So this is something I have to put on discourse I guess.

Buttons are good:



And no left menu :



Ready to go!

luisramos0:
nice @RachL makes sense! discourse or gh... it's an improvement.

RachL:
@luisramos0 alright! Will open it this way

PR #3650, opened 2019-03-25
3487,2786 [Spree Upgrade] Address spec failures for SubscriptionConfirmJob

### LINKED ISSUE #3487: [Spree Upgrade] Fix subscription_confirm_job_spec (1 broken spec)
Broken spec, can be seen in this build:
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3317/builds/4

```
3) SubscriptionConfirmJob processing an order when payments need to be processed and no errors are added when updating payments when payments are processed without error sends only a subscription confirm email, no regular confirmation emails
     Failure/Error: expect(ActionMailer::Base.deliveries.count).to be 1
     
       expected #<Fixnum:3> => 1
            got #<Fixnum:1> => 0
     
       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     # ./spec/jobs/subscription_confirm_job_spec.rb:168:in `block (6 levels) in <top (required)>'
```

kristinalim:
#### What? Why?

Closes #3487
Relates to #2786 

This fixes order status advancement in the setup code for the specs.

#### What should we test?

This doesn't update app code. A green build is sufficient.

### DIFF
--- a/spec/jobs/subscription_confirm_job_spec.rb
+++ b/spec/jobs/subscription_confirm_job_spec.rb
@@ -11,7 +11,10 @@ describe SubscriptionConfirmJob do
     let(:order_cycle2) { create(:simple_order_cycle, coordinator: shop, orders_close_at: 61.minutes.ago, updated_at: 1.day.ago) }
     let(:schedule) { create(:schedule, order_cycles: [order_cycle1, order_cycle2]) }
     let(:subscription) { create(:subscription, with_items: true, shop: shop, schedule: schedule) }
-    let!(:proxy_order) { create(:proxy_order, subscription: subscription, order_cycle: order_cycle1, placed_at: 5.minutes.ago) }
+    let!(:proxy_order) do
+      create(:proxy_order, subscription: subscription, order_cycle: order_cycle1,
+                           placed_at: 5.minutes.ago)
+    end
     let!(:order) { proxy_order.initialise_order! }
     let(:proxy_orders) { job.send(:proxy_orders) }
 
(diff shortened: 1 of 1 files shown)


luisramos0:
@kristinalim the build is still broken:
```
  1) SubscriptionConfirmJob finding proxy_orders that are ready to be confirmed ignores orders that have been cancelled
     Failure/Error: proxy_order.order.cancel!
     
     StateMachine::InvalidTransition:
       Cannot transition state via :cancel from :cart (Reason(s): There are no items for this order. Please add an item to the order to continue.)
     # ./spec/jobs/subscription_confirm_job_spec.rb:64:in `block (3 levels) in <top (required)>'

```

kristinalim:
@luisramos0 All green now. :slightly_smiling_face:

spec/jobs/subscription_confirm_job_spec.rb
```diff
@@ -168,6 +176,7 @@
           end
 
           it "sends only a subscription confirm email, no regular confirmation emails" do
+            ActionMailer::Base.perform_deliveries = true
```
sauloperez: We must restore its value afterward to avoid messing the global state. You can use `performing_deliveries` for that as we did in spec/features/admin/users_spec.rb

kristinalim: Thanks, @sauloperez. Yes, I agree. I've fixed this here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3650/commits/f7b47fb622ceaafed01d0fab775f8cd001352651

kristinalim: I spoke too soon. This is the correct commit: https://github.com/openfoodfoundation/openfoodnetwork/pull/3650/commits/f0746eb64387ba17cc655f79dd57bf5efd554a10 Build is green again.

PR #3651, opened 2019-03-25
[Spree Upgrade] Fix Enterprise fee report specs

### LINKED ISSUE #3483: [Spree Upgrade] Fix spec/features/admin/reports/enterprise_fee_summaries_spec (1 broken spec)
## Description
spec/features/admin/reports/enterprise_fee_summaries_spec.rb has one broken spec probably related to user privileges. This spec has been moved to xdescribe as part of #3479, it needs to be fixed and moved back to describe.

The authorization problem is now resolved in #3316 but there are other problems in the spec.
The first one is that headless chrome does not support page.response_headers
But even if we remove these assertions (they assert the content type is CSV), there are other errors in 3 specs that need to be looked at.

## Expected Behavior
Spec is green and not xdescribe.

## Actual Behaviour
Spec is marked with xdescribe.

## Possible Fix
the workaround for the content type test doesnt seem simple here: https://about.gitlab.com/2017/12/19/moving-to-headless-chrome/
Maybe we just remove the assertions related to content type.

luisramos0:
#### What? Why?

Closes #3483

Chrome driver doesnt support file downloads with standard config so we need specific configuration to make it download files and have specs get the file contents.

#### What should we test?
Spec should be green.
We should test the enterprise fees summary report and make sure it's working correctly.

This build is green if I include the commits from #3316 (they fix the authorization problem, they are present in build number 2 in this PR). I have removed those commits now to make the review of this PR easier. So, this should be merged only after #3316.

### DIFF
--- a/spec/features/admin/reports/enterprise_fee_summaries_spec.rb
+++ b/spec/features/admin/reports/enterprise_fee_summaries_spec.rb
@@ -120,8 +120,9 @@ xfeature "enterprise fee summaries", js: true do
         it "generates file with data for all enterprises" do
           check I18n.t("filters.report_format_csv", scope: i18n_scope)
           click_on I18n.t("filters.generate_report", scope: i18n_scope)
-          expect(page.response_headers['Content-Type']).to eq "text/csv"
-          expect(page.body).to have_content(distributor.name)
+
+          expect(downloaded_filename).to include ".csv"
+          expect(downloaded_content).to have_content(distributor.name)
         end
       end
 
@@ -139,9 +140,11 @@ xfeature "enterprise fee summaries", js: true do
         it "generates file with data for the enterprise" do
           check I18n.t("filters.report_format_csv", scope: i18n_scope)
           click_on I18n.t("filters.generate_report", scope: i18n_scope)
-          expect(page.response_headers['Content-Type']).to eq "text/csv"
-          expect(page.body).to have_content(distributor.name)
-          expect(page.body).not_to have_content(other_distributor.name)
+
+          expect(downloaded_filename).to include ".csv"
+          csv_content = downloaded_content
+          expect(csv_content).to have_content(distributor.name)
+          expect(csv_content).not_to have_content(other_distributor.name)
         end
       end
     end
@@ -169,9 +172,11 @@ xfeature "enterprise fee summaries", js: true do
         select order_cycle.name, from: "report_order_cycle_ids"
         check I18n.t("filters.report_format_csv", scope: i18n_scope)
         click_on I18n.t("filters.generate_report", scope: i18n_scope)
-        expect(page.response_headers['Content-Type']).to eq "text/csv"
-        expect(page.body).to have_content(distributor.name)
-        expect(page.body).not_to have_content(second_distributor.name)
+
+        expect(downloaded_filename).to include ".csv"
+        csv_content = downloaded_content
+        expect(csv_content).to have_content(distributor.name)
+        expect(csv_content).not_to have_content(second_distributor.name)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -145,6 +155,7 @@ def restart_phantomjs
   config.include ActionView::Helpers::DateHelper
   config.include OpenFoodNetwork::DelayedJobHelper
   config.include OpenFoodNetwork::PerformanceHelper
+  config.include DownloadsHelper
```
kristinalim: Better if we only include this in feature specs:

```ruby
config.include DownloadsHelper, type: :feature
```

spec/features/admin/reports/enterprise_fee_summaries_spec.rb
```diff
@@ -96,60 +98,86 @@
...
 
-  describe "smoke test for generation of report based on permissions" do
-    before do
-      visit spree.new_admin_reports_enterprise_fee_summary_path
+  describe "csv downloads" do
+    around do |example|
+      clear_downloads
+      example.run
+      clear_downloads
     end
```
kristinalim: How about adding a method similar to `performing_deliveries` in `EmailHelper` for this?

luisramos0: ah, nice. I'll do that.

luisramos0:
@kristinalim all credit for this PR to the internet 🤣  DownloadsHelper for example was copied from here: https://collectiveidea.com/blog/archives/2012/01/27/testing-file-downloads-with-capybara-and-chromedriver

spec/features/admin/reports/enterprise_fee_summaries_spec.rb
luisramos0: done

luisramos0:
The build is fine, needs #3316.

luisramos0:
I just rebased this to see the green build.

RachL:
@luisramos0 I've loaded the same period of data in staging and in production. But the results don't match. Is there anything else than the date that can change the results of this report ?

https://docs.google.com/document/d/17-d1pT1y5eWYbCg-bY2NdmQRULZDoSdZSZPGPSeqKBY/edit#

luisramos0:
This PR only changes specs. So, this PR cannot be breaking the report. I think the scope of testing is more about validating the report itself in v2.

Note that the data is not only truncated, it's also anonymised and that can break data connections as we have seen in another report (for example, order.email will not match order.user.email).

@RachL can you share the exact filter you used so I can explain exactly why the numbers don't match between live users and anonimized users?

RachL:
@luisramos0 yes of course.

So on both platform I've chosen a date range From January the 1st to April the 1st. You can reach data without breaking anything with just this filter.

The first 3 lines in the results show the 5 per cent fee for a particular hub. These 3 lines are linked to three different users.

The first line has a different sum in staging than in production :

Staging: 

image

Production:

image

I've understood reading this: https://community.openfoodnetwork.org/t/report-that-isolates-various-enterprise-fees/1361/4 that there should be only one line per hub-feetype-customer-feeplacement. 

I thought that maybe they were aggregated by OC, so I've filtered through the "christmas holiday" OC. But I still see differences. So what am I missing?

luisramos0:
eheh, it took me a while to confirm that it's just because the data changed after the data load to staging...
we moved data from live uk to stg FR on the 29th of March and the data related to this 5 per cent fee has changed on the 30th Mars. those 0.5 are from the 30th of Mars...

Can you please use data fro Feb for example, less chances of mismatch :-)

RachL:
@luisramos0 it still didn't change anything I still had different figures... The feature is not yet documented enough in order for me to understand all the rules that applies. So I change my testing approach and did the follow tests :
- create an order and see if the fees displayed well in the report : they did
- past orders fees seems to display correctly as well.

This is the first time I'm testing this report in v2, so I guess staging AU is not updated with this compliant version right? When it is, I will ask Kirsten and Theresa to have a deeper look and help on creating the user guide for this.

luisramos0:
ok! staging AU does contain the same version of the reports because this PR is only fixing the specs..

PR #3654, opened 2019-03-26
Test inventory report to use variant overrides

### LINKED ISSUE #3237: [Spree Upgrade] Verify all points where scope_to_variant is used (non-core features)
After verifying scoping for core features in #3172, we need to do the same for non-core features.

This is the list of references:

```ruby
- 1
/Users/luisramos/dev/ofn/openfoodnetwork/app/services/order_factory.rb:
   77  
   78    def scoper
   79:     @scoper ||= OpenFoodNetwork::ScopeVariantToHub.new(shop)
   80    end
   81  end

- 2
/Users/luisramos/dev/ofn/openfoodnetwork/app/services/subscription_estimator.rb:
   43  
   44    def scoper
   45:     OpenFoodNetwork::ScopeVariantToHub.new(shop)
   46    end
   47  

- 3
/Users/luisramos/dev/ofn/openfoodnetwork/lib/open_food_network/products_and_inventory_report_base.rb:
   60        if params[:distributor_id].to_i > 0
   61          distributor = Enterprise.find params[:distributor_id]
   62:         scoper = OpenFoodNetwork::ScopeVariantToHub.new(distributor)
   63          variants.in_distributor(distributor).each { |v| scoper.scope(v) }
   64        else
```

mkllnk:
#### What? Why?

Closes #3237

For the Spree v2 upgrade we want to make sure that all code is using variant overrides correctly. This PR adds a couple of specs to cover the products and inventory reports.

#### What should we test?

Spec only changes, no tests required.

I tested the new specs with master and 2-0-stable, both passing.

### DIFF
--- a/spec/lib/open_food_network/products_and_inventory_report_spec.rb
+++ b/spec/lib/open_food_network/products_and_inventory_report_spec.rb
@@ -134,6 +134,32 @@ module OpenFoodNetwork
           subject.stub(:params).and_return(distributor_id: distributor.id)
           subject.filter(variants).should == [product2.variants.first]
         end
+
+        it "ignores variant overrides without filter" do
+          distributor = create(:distributor_enterprise)
+          product = create(:simple_product, supplier: supplier, price: 5)
+          variant = product.variants.first
+          order_cycle = create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], variants: [product.variants.first])
+          create(:variant_override, hub: distributor, variant: variant, price: 2)
+
+          result = subject.filter(variants)
+
+          expect(result.first.price).to eq 5
+        end
+
+        it "considers variant overrides with distributor" do
+          distributor = create(:distributor_enterprise)
+          product = create(:simple_product, supplier: supplier, price: 5)
+          variant = product.variants.first
+          order_cycle = create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], variants: [product.variants.first])
+          create(:variant_override, hub: distributor, variant: variant, price: 2)
+
+          allow(subject).to receive(:params).and_return(distributor_id: distributor.id)
+          result = subject.filter(variants)
+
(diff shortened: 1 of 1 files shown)


spec/lib/open_food_network/products_and_inventory_report_spec.rb
```diff
@@ -134,6 +134,32 @@ module OpenFoodNetwork
...
+        end
+
+        it "considers variant overrides with distributor" do
+          distributor = create(:distributor_enterprise)
+          product = create(:simple_product, supplier: supplier, price: 5)
+          variant = product.variants.first
+          order_cycle = create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], variants: [product.variants.first])
+          create(:variant_override, hub: distributor, variant: variant, price: 2)
+
+          allow(subject).to receive(:params).and_return(distributor_id: distributor.id)
```
sauloperez: It is a bad practice to stub the subject of the test because we are not testing the actual implementation. Do you see a way to overcome this? I guess we need to pass a different set of params instead but I haven't checked the report's implementation.

mkllnk: I agree that it's bad practice. But I think fixing it is beyond this pull request.

1. I used the same pattern as in the rest of the spec file. Solving this problem would be an improvement but it should be done for the whole file.
2. The `params` method is not provided by this controller but by one of the ancestors. Rails provides it. That's why I think it's okay to stub it. Ideally, all that filtering would happen in a service that is taking the params as input and we didn't need to stub it.

@sauloperez Are you okay to leave this as is? I don't like to get into refactoring work which then creates merge conflicts. Let's be stricter once this Spree upgrade is done.

sauloperez: Fine, let's just create a tech debt issue then.

PR #3659, opened 2019-03-27
Fix "What's this?" on admin dashboard

mkllnk:
#### What? Why?

When a user has no enterprises, the admin dashboard displays a "What's
this?" hint in the right corner of the enterprises tab. The link text
and the tooltip were both broken (missing translation).

#### What should we test?

I don't think the actual translation is worth testing. The only way to see this page is to delete all enterprises.

1. Delete all enterprises or reset the database. Don't load sample data.
2. Log in as super admin and verify there are no missing translations in the dashboard.

Instead I would suggest to test that the dashboard still displays correctly:

1. Log in as super admin and visit the dashboard.
2. Log in as enterprise user and visit the dashboard.

In both cases the "+ Create new" button is in the corner of the enterprises tab. Only when there are no enterprises, the fixed tooltip displays there.

#### Release notes

Fixed: Corrected a tooltip in the admin dash board.

Changelog Category: Fixed

### DIFF
--- a/app/views/spree/admin/overview/_enterprises_header.html.haml
+++ b/app/views/spree/admin/overview/_enterprises_header.html.haml
@@ -7,4 +7,4 @@
         = t "spree_admin_enterprises_create_new"
   - else
     %a{ "ofn-with-tip" => t('.ofn_with_tip') }
-      = t "admin_enterprise_groups_what_s_this"
+      = t "admin.whats_this"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1013,9 +1013,6 @@ en:
     shared:
       user_guide_link:
         user_guide: User Guide
-    overview:
-      enterprises_header:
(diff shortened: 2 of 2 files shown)


mkllnk:
Well, I'm not sure. We want the translation to be available in Transifex. That's why we did translation PRs in master. But now we are so close. If this is merged without testing, it doesn't really matter. @luisramos0 What do you think?

Feel free to merge it to either branch.

luisramos0:
I think it's better to do it in 2-0 and backport translations. that way we do it in 2-0 and test it properly there when we have the translations in v2.
so, a PR with the translations in master plus a PR depending on the first PR with the code changes?
Doing it in master and testing twice is worse than this. There's no easy way out of this one....

mkllnk:
Okay, I changed the base.

RachL:
Super admin:



Enterprise:



+ the create new enterprise is still working as well!

Issue #93, opened 2013-09-19
RSOD when I try to use the Supplier filter, when logged in as Enterprise User


kirstenalarsen:
http://www.bugherd.com/projects/21919/tasks/76

PR #3662, opened 2019-03-28
 Add shipping method name to orders detail report

### LINKED ISSUE #3446: orders_and_distributors report is missing shipment method
## Description
The orders_and_distributors report contains the details of the order, but not the related selected shipment method. When the distributor has multiple "pick-up"-type shipment methods, this missing information is crucial in managing the orders using this report.

## Expected Behavior
see shipment method in report table (and csv export)

## Actual Behavior
shipment method is missing

## Steps to Reproduce
1. create 2 shipment method of "pick-up" type, in different locations, for a single distributor
2. create 2 orders, with same products, each with a different "pick-up" point
3. create report, where you cannot distinguish the logistics details between the 2 orders

## Context
More efficient logistics to group orders grouped by pick-up location for shipment.

## Severity
bug-s4: it's annoying, but you can use it when combining with customers report and praying the user did not pick a different pickup location for this order... which is a dangerous bet.

## Your Environment
* Version used: .be instance
* Browser name and version: chrome 71.0
* Operating System and version (desktop or mobile): Win8 desktop

SDBowen:
#### What? Why?

Adds a Shipping Method column to the Orders And Distributors report. This allows the users to see the the shipping detail of each order.

Closes #3446 

Before:

Screenshot from 2019-03-27 19-39-00

After:

Screenshot from 2019-03-27 19-04-20

#### What should we test?

Generating a Orders and Distributors report will show the result of this PR.

### DIFF
--- a/lib/open_food_network/order_and_distributor_report.rb
+++ b/lib/open_food_network/order_and_distributor_report.rb
@@ -10,7 +10,8 @@ module OpenFoodNetwork
     end
 
     def header
-      [I18n.t(:report_header_order_date),
+      [
+        I18n.t(:report_header_order_date),
         I18n.t(:report_header_order_id),
         I18n.t(:report_header_customer_name),
         I18n.t(:report_header_customer_email),
@@ -28,7 +29,9 @@ module OpenFoodNetwork
         I18n.t(:report_header_distributor_address),
         I18n.t(:report_header_distributor_city),
         I18n.t(:report_header_distributor_postcode),
-        I18n.t(:report_header_shipping_instructions)]
+        I18n.t(:report_header_shipping_method),
+        I18n.t(:report_header_shipping_instructions)
+      ]
     end
 
     def search
@@ -95,6 +98,7 @@ module OpenFoodNetwork
         order.distributor.address.address1,
         order.distributor.address.city,
         order.distributor.address.zipcode,
+        order.shipping_method.name,
         order.special_instructions
       ]
     end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -20,7 +20,6 @@ Layout/AlignArray:
   Exclude:
     - 'lib/open_food_network/bulk_coop_report.rb'
     - 'lib/open_food_network/customers_report.rb'
-    - 'lib/open_food_network/order_and_distributor_report.rb'
     - 'lib/open_food_network/orders_and_fulfillments_report.rb'
     - 'lib/open_food_network/packing_report.rb'
     - 'spec/lib/open_food_network/order_grouper_spec.rb'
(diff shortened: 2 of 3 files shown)


lib/open_food_network/order_and_distributor_report.rb
```diff
@@ -28,6 +28,7 @@ def header
         I18n.t(:report_header_distributor_address),
         I18n.t(:report_header_distributor_city),
         I18n.t(:report_header_distributor_postcode),
+        I18n.t(:report_header_shipping_method),
```
SDBowen: Interesting, this line has failed with rubocop. I believe it wants everything aligned with the first line of the array.

Let me know if this is right, and I will amend!

luisramos0: the problem was already there, you can fix it by making this header method similar to the row_for method
```
def header
  [
    I18n.t(...),
    I18n.t(...)
  ]

```
if you do this you can also delete this line from the exceptions:
https://github.com/openfoodfoundation/openfoodnetwork/blob/fed2b6074320f814daac822fd6c2e0020a1749bd/.rubocop_todo.yml#L23

SDBowen: Easy enough! Fixing this now.

RachL:
Shipping method is now shown in the order and distribution report. 

image


The report section is working as usual. This is ready to go.

luisramos0:
I have merged this PR exceptionally because I am curious if the [contributors list](https://github.com/openfoodfoundation/openfoodnetwork/graphs/contributors) gets updated. @SDBowen would be contributor #93
I see that @amers185 has not been added after merging #3644 (although I see the [commit in #3634](https://github.com/openfoodfoundation/openfoodnetwork/pull/3644/commits/698c71f0392fe014f44030c724228550f4f9482a) is from unknown user Usama.

luisramos0:
yeah, so @SDBowen has gone into the list of contributors but @amers185 has not, probably because the way @amers185 has configured git locally the commit got a different user that doesn't match the GH user.

amers185:
@luisramos0 Hello Luis! What can I do correctly on my end to go on the list of contributors? I was working on the branch created by @kevinchristianson perhaps that has something to do with it? I have also cloned openfoodnetwork but I had trouble bringing it up to date because of insufficient permissions. So I decided to work on the branch created by @kevinchristianson

luisramos0:
it's not a problem with the branch, it's the commit itself.
something to do with your configuration of git.
maybe just this: https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-linked-to-another-user

amers185:
@luisramos0 I will look into this. Perhaps, I should have mentioned earlier but my name is Usama so that is in fact me.

luisramos0:
yeah, I got that it was you. It's probably your local account. Good luck with that, let us know on slack if you need help.

PR #3665, opened 2019-03-29
Relax stripe version requirement

mkllnk:
#### What? Why?

Follow up from #3656.

We are using the latest version of the `stripe` gem. We don't depend on
any particular version. I'm proposing to drop our dependency declaration
on a particular version and just track the current used version in
Gemfile.lock. That means fewer code changes when updating, which happens
quite frequently with this gem and dependabot.

#### What should we test?

No test required. The actual version of the gem is not changed here.

#### Release notes

Changed: Our dependency specification (`Gemfile`) became more accurate.

Changelog Category: Changed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ gem 'spree_auth_devise', github: 'openfoodfoundation/spree_auth_devise', branch:
 # - Change type of password from string to password to hide it in the form
 gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "spree-upgrade-intermediate"
 #gem 'spree_paypal_express', github: "spree-contrib/better_spree_paypal_express", branch: "1-3-stable"
-gem 'stripe', '~> 4.11.0'
+gem 'stripe'
 # We need at least this version to have Digicert's root certificate
 # which is needed for Pin Payments (and possibly others).
 gem 'activemerchant', '~> 1.78'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -23,7 +23,7 @@ gem 'spree_auth_devise', github: 'openfoodfoundation/spree_auth_devise', branch:
 # - Change type of password from string to password to hide it in the form
 gem 'spree_paypal_express', github: "openfoodfoundation/better_spree_paypal_express", branch: "spree-upgrade-intermediate"
 #gem 'spree_paypal_express', github: "spree-contrib/better_spree_paypal_express", branch: "1-3-stable"
-gem 'stripe', '~> 4.11.0'
+gem 'stripe'
```
sauloperez: I don't think this is a good idea. Assuming Stripe follows semantic versioning, this would enable updating to versions > 4, which could introduce breaking changes that would require working on our integration. I don't think it is something we want to do now. By sticking to the current version declaration we ensure we get fixes and backward compatible changes only.

> That means fewer code changes when updating, which happens quite frequently with this gem and dependabot.

If we still want to reduce the changed files, we could do so adding `gem 'stripe', '~> 4.11'`. This would still disable major version upgrades.

luisramos0: context: 4.11 was latest until yesterday.
This change would just make dependabot push for new versions of it, so we would always be able to see/test the updates. So, with this PR in, when stripe 5 is out, we will have a dependabot PR with it and we can decide what to do bases on build and tests. If we keep the version in the gem file dependabot will not let us know about stripe 5.

sauloperez: Sounds good too :+1:

PR #3666, opened 2019-03-29
Close adjustments so that updates are effective

### LINKED ISSUE #1830: [Spree Upgrade][Admin/Orders/Edit] Fix edit orders shipping adjustment

mkllnk:
#### What? Why?

Closes #1830

Spree 2.0 keeps shipping fee adjustments open unless they are manually    closed. But open adjustments cannot be edited.    To preserve updates, like changing the amount of the shipping fee    we close the adjustment first.

#### What should we test?

1. Place an order.
2. Log in as admin and go to the order.
3. Observe the shipping fee and total amount.
4. Click on Adjustments.
5. Click on the edit icon of the shipping fee.
6. Enter a different amount and click Continue.
7. Check the amount was updated.
8. Click on Order Details and check the shipping fee and totals have updated.

Part of the Spree Upgrade.

#### Documentation updates

**Deleting shipping fees**

If you delete a shipping fee and look at the order details again, the shipping fee will re-appear. You can use this behaviour if you want to re-calculate the shipping fee. If you don't want to charge a shipping fee you can set the amount of the shipping fee to `0`.

### DIFF
--- a/app/controllers/spree/admin/adjustments_controller_decorator.rb
+++ b/app/controllers/spree/admin/adjustments_controller_decorator.rb
@@ -3,7 +3,7 @@ module Spree
     AdjustmentsController.class_eval do
       prepend_before_filter :set_included_tax, only: [:create, :update]
       before_filter :set_default_tax_rate, only: :edit
-
+      before_filter :enable_updates, only: :update
 
       private
 
@@ -37,6 +37,17 @@ module Spree
         else
           params[:adjustment][:included_tax] = 0
         end
+      end
+
+      # Spree 2.0 keeps shipping fee adjustments open unless they are manually
+      # closed. But open adjustments cannot be edited.
+      # To preserve updates, like changing the amount of the shipping fee,
+      # we close the adjustment first.
+      #
+      # The Spree admin interface allows to open and close adjustments manually
+      # but we removed that functionality as it had no purpose for us.
+      def enable_updates
+        @adjustment.close
       end
     end
   end
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -350,6 +350,16 @@ feature %q{
           end
         end
       end
+
+      scenario "editing shipping fees" do
+        click_link "Adjustments"
+        page.find('td.actions a.icon-edit').click
+
+        fill_in "Amount", with: "5"
+        click_button "Continue"
+
+        expect(page.find("td.amount")).to have_content "$5.00"
+      end
     end
 
     scenario "creating an order with distributor and order cycle" do


app/controllers/spree/admin/adjustments_controller_decorator.rb
```diff
@@ -38,6 +38,17 @@ def set_included_tax
           params[:adjustment][:included_tax] = 0
         end
       end
+
+      # Spree 2.0 keeps shipping fee adjustments open unless they are manually
+      # closed. But open adjustments cannot be edited.
```
sauloperez: How did you get to this conclusion? our `#update_adjustment!` still opens the adjustment to update it so it feels contradictory.

mkllnk: Good question. I simply observed that `adjustment.update_attributes` didn't have any effect on an open shipping fee adjustment. But after I closed it, it did update the attributes. I *guess* that you need to open adjustments to reset them to their calculated value and you have to close them to set a custom value. Do you think we should dig deeper in the code to verify that theory?

mkllnk: I checked vanilla Spree and an adjustment can be updated either open or closed. But it triggers the after_save callback `update_adjustable` which calls `update!` on the order and that eventually triggers a reset of the adjustment. I guessed that before but here is the proper trace:

1. `adjustment.update_attributes` triggers `adjustable.update!`.
2. `Order#update!` triggers the `Spree::OrderUpdater`.
3. The updater updates all shipments and adjustments.
4. `Adjustment#update!` doesn't do anything if the adjustment is closed.
5. When the adjustment is open, `originator.update_adjustment` is called.
6. `ShippingMethod#Adjustment` recalculates the amount and luckily avoids callbacks: `adjustment.update_attribute_without_callbacks(:amount, compute_amount(calculable))`. Otherwise we would have an infinite loop.

All this seems Spree code. It's not something we accidentally caused with OFN overrides. So Spree defines this way:

- Open shipping adjustments are always recalculated -> can't be changed to a custom value.
- Closed shipping adjustments are not recalculated -> can be changed to a custom value.

This research presents another option though. We could try to call `adjustment.update_attribute_without_callbacks` to update the amount. I guess though that it wouldn't last long. The adjustment would probably be recalculated when anything in the callback jungle is triggered, e.g. order.next. I conclude that closing the adjustment is the "Spree thing" to do if we want to set a custom amount.

Do you think that's reasonable?

luisramos0: ah, yes, I followed this track in the past. Makes sense.
When you edit an adjustment, you need to close it, otherwise it gets recalculated. That's the spree logic.
finalized means you can't, you shouldnt change the value in any way.

sauloperez: Could you take the time to paste that amazing investigation in the Spree upgrade wiki section? It will save us time in the future.

mkllnk: :+1: Done. https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-Upgrade:-Orders-and-Adjustments-updates

RachL:
This is good to go https://docs.google.com/document/d/1Ks1lcvGxUaAiDHINnNcGnBaBfsQr-AfoUhbQkVrKKDA/edit#

I've noticed something weird on the order total, but I will test this further and see if a new issue needs to be open.

luisramos0:
nice nice nice!!!

PR #3667, opened 2019-03-29
3477 Render only eligible adjustments on the order page

### LINKED ISSUE #3477: [Spree Upgrade] Adjustments with amount zero are being displayed in the orders edit page
## Description
Order edit page is showing adjustments that have amount zero. These adjustments should be removed by using the scope adjustment.eligible. See [this review comment](https://github.com/openfoodfoundation/openfoodnetwork/pull/3445#discussion_r255901384).

Additionally, the id order-charges is being used several times making the html invalid. See [this comment](https://github.com/openfoodfoundation/openfoodnetwork/pull/3445#discussion_r255896692)

## Expected Behavior
Adjustments with amount zero should not be shown.
html div Id order-charges is NOT present multiple times in the html if order has more than one adjustment.

## Actual Behaviour
Adjustments with amount zero are listed.
html div Id order-charges is present multiple times in the html if order has more than one adjustment.

## Steps to Reproduce
Create an order with a adjustment with a zero amount (a specific example needs to be given here) and verify it is not showing on the orders edit page.

## Possible Fix
Use the eligible scope to fetch adjustments.

mkllnk:
#### What? Why?

Closes #3477 

The admin orders edit form was displaying adjustments even if they were    "not eligible". For example, an additional fee with amount `0` is not    eligible. They were already hidden in the adjustments view and we are    hiding them in reports.

#### What should we test?

1. Place an order.
2. Log in as admin and go to the order.
3. Go to adjustments and create an adjustment with zero amount.
4. After Saving you will see a success message but you won't see the adjustment. That's confusing but correct, I think.
5. Go back to the order details. You shouldn't see the adjustment there either.

Bad:
Screenshot from 2019-03-29 18-45-48

Good:
Screenshot from 2019-03-29 18-47-06

### DIFF
--- a/app/helpers/admin/orders_helper.rb
+++ b/app/helpers/admin/orders_helper.rb
@@ -4,16 +4,9 @@ module Admin
     #
     # We exclude shipping method adjustments because they are displayed in a
     # separate table together with the order line items.
-    #
-    # We also exclude tax rate adjustment with zero value.
     def order_adjustments_for_display(order)
       order.adjustments.eligible.select do |adjustment|
-        type = adjustment.originator_type
-
-        is_shipping_method_adjustment = (type == 'Spree::ShippingMethod')
-        is_zero_tax_rate_adjustment = (type == 'Spree::TaxRate' && adjustment.amount.zero?)
-
-        !is_shipping_method_adjustment && !is_zero_tax_rate_adjustment
+        adjustment.originator_type != "Spree::ShippingMethod"
       end
     end
   end
--- a/app/helpers/admin/orders_helper.rb
+++ b/app/helpers/admin/orders_helper.rb
@@ -1,7 +1,14 @@
 module Admin
   module OrdersHelper
     def order_adjustments(order)
-      order.adjustments.eligible
+      order.adjustments.eligible.select do |adjustment|
+        type = adjustment.originator_type
+
+        is_shipping_method_fee = (type == 'Spree::ShippingMethod')
+        is_zero_tax_rate = (type == 'Spree::TaxRate' && adjustment.amount.zero?)
+
+        !is_shipping_method_fee && !is_zero_tax_rate
+      end
     end
   end
 end
--- a/app/helpers/admin/orders_helper.rb
+++ b/app/helpers/admin/orders_helper.rb
@@ -1,0 +1,7 @@
+module Admin
+  module OrdersHelper
+    def order_adjustments(order)
+      order.adjustments.eligible
+    end
+  end
+end
--- a/app/helpers/admin/orders_helper.rb
+++ b/app/helpers/admin/orders_helper.rb
@@ -1,13 +1,19 @@
 module Admin
   module OrdersHelper
-    def order_adjustments(order)
+    # Adjustments to display under "Order adjustments".
+    #
+    # We exclude shipping method adjustments because they are displayed in a
+    # separate table together with the order line items.
+    #
+    # We also exclude tax rate adjustment with zero value.
+    def order_adjustments_for_display(order)
       order.adjustments.eligible.select do |adjustment|
         type = adjustment.originator_type
 
-        is_shipping_method_fee = (type == 'Spree::ShippingMethod')
-        is_zero_tax_rate = (type == 'Spree::TaxRate' && adjustment.amount.zero?)
+        is_shipping_method_adjustment = (type == 'Spree::ShippingMethod')
+        is_zero_tax_rate_adjustment = (type == 'Spree::TaxRate' && adjustment.amount.zero?)
 
-        !is_shipping_method_fee && !is_zero_tax_rate
+        !is_shipping_method_adjustment && !is_zero_tax_rate_adjustment
       end
     end
   end
--- a/app/views/spree/admin/orders/_form/_adjustments.html.haml
+++ b/app/views/spree/admin/orders/_form/_adjustments.html.haml
@@ -8,8 +8,7 @@
           %th= Spree.t('amount')
       %tbody.with-border
         - adjustments.each do |adjustment|
-          - if (adjustment.originator_type != 'Spree::ShippingMethod') && !(adjustment.originator_type == 'Spree::TaxRate' && adjustment.amount == 0)
-            %tr.total
-              %td.strong= adjustment.label + ":"
-              %td.total.align-center
-                %span= Spree::Money.new(adjustment.amount)
+          %tr.total
+            %td.strong= adjustment.label + ":"
+            %td.total.align-center
+              %span= Spree::Money.new(adjustment.amount)
--- a/app/views/spree/admin/orders/_form/_adjustments.html.haml
+++ b/app/views/spree/admin/orders/_form/_adjustments.html.haml
@@ -6,7 +6,7 @@
         %tr
           %th= Spree.t('name')
           %th= Spree.t('amount')
-      %tbody#order-charges.with-border
+      %tbody.with-border
         - adjustments.each do |adjustment|
           - if (adjustment.originator_type != 'Spree::ShippingMethod') && !(adjustment.originator_type == 'Spree::TaxRate' && adjustment.amount == 0)
             %tr.total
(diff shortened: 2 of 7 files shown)


mkllnk:
Ah, I see. We need to filter the adjustments before passing them on to the partial.

mkllnk:
@luisramos0 I added a few more commits that need a review.

- The duplicate HTML id issue is fixed (I didn't see that before).
- The spec is fixed.
- The empty adjustments table is hidden.

mkllnk:
@luisramos0 I added another little commit for a missing translation on that page. I also backported it to master for translations: https://github.com/openfoodfoundation/openfoodnetwork/pull/3679

But now I'm wondering if we need to do that in the future. I guess we will merge to master at some point. And then we can tell everybody to go on Transifex and translate. If we don't backport, that could be a lot in one go but that could be as efficient as translating bits and pieces from time to time. What do you think?

luisramos0:
re translations, I think we should put the translations in master now in some way. Previously I put the translations only in master and waited for the merge from master to 2-0, but I think backporting them is better now :+1:

Are we merging 2-0-stable to master? omg, when, how, is it that simple?

app/helpers/admin/orders_helper.rb
```diff
@@ -0,0 +1,14 @@
+module Admin
+  module OrdersHelper
+    def order_adjustments(order)
+      order.adjustments.eligible.select do |adjustment|
+        type = adjustment.originator_type
+
+        is_shipping_method_fee = (type == 'Spree::ShippingMethod')
```
luisramos0: I think it's important we are strict with naming so for me there's a big difference between shipping_method_fee, shipping_method_adjustment and shipping_fee. Fee's (just like hubs) dont exist in ofn code, we only have adjustments not fees. So I'd call this variable is_shipping_method_adjustment.

app/helpers/admin/orders_helper.rb
```diff
@@ -0,0 +1,14 @@
+module Admin
+  module OrdersHelper
+    def order_adjustments(order)
```
luisramos0: this is not really order_adjustments right?
this method is order_adjustments_displayed_in_the_adjustments_table_on_the_order_edit_page_which_excludes_shipping_method_adjustments_and_zero_tax_rate_adjustments 🤣 

because of this I'd call it order_adjustments_for_display and add a method comment with something like
"we exclude shipping method adjustments because they are displayed in a separate table together with the order line items"
and "we exclude tax rate adjustment with zero value because..." I actually dont know why tax rates need to be excluded here, they would be removed already as not eligible if amount is zero.

mkllnk:
@luisramos0 
> Are we merging 2-0-stable to master? omg, when, how, is it that simple?

That's probably a discussion for the next meeting. But yes, at some point we should probably do this:

- Last v1 release.
- Merge 2-0-stable to master.
- Extensive testing and bug fixing.
- Publish 2.0.0-beta pre-release.
- Test that release on small instances (Germany, Canada, Katuma).
- Observe and fix bugs.
- Release 2.0.0.
- Carefully deploy to one instance at a time, maybe leaving a couple of days in between to observe errors.
- Fix bugs and release 2.0.1, 2.0.2,...2.0.42. ;-)
- Deploy to Australia (3rd most traffic, oldest data)
- Deploy to UK (2nd most traffic, 2nd oldest data)
- Deploy to France (most traffic)

app/helpers/admin/orders_helper.rb
mkllnk: I agree, I added another commit (and then made it obsolete :wink:)

app/helpers/admin/orders_helper.rb
mkllnk: Thank you for bringing it up. I guess you could argue that helpers are to be used in views. So all helper are "for display" and therefore the suffix is redundant. :stuck_out_tongue_closed_eyes: 
In this case it's probably better to point it out though. 

I had a similar thought about the zero value adjustments. So I first covered that part with a spec and then removed the redundant logic. Much simpler, isn't it?

luisramos0: you refactored code to delete it afterwards :-)
re removing tax rate filter, we can do that. I am just not 100% sure it is redundant...

luisramos0:
nice summary! I created an issue for the "v2 release" here #3692

spec/features/admin/orders_spec.rb
```diff
@@ -267,7 +267,7 @@ def new_order_with_distribution(distributor, order_cycle)
       end
 
       scenario "shows the order tax adjustments" do
-        within('tbody#order-charges') do
+        within('fieldset', text: 'LINE ITEM ADJUSTMENTS') do
```
sauloperez: We should reference the translation key instead as per https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code-Conventions#always-use-i18nt-in-spec-assertions, right?

spec/features/admin/orders_spec.rb
```diff
@@ -283,6 +283,19 @@ def new_order_with_distribution(distributor, order_cycle)
...
+          :adjustment,
+          adjustable: @order,
+          label: "invalid adjustment",
+          amount: 0
+        )
+
+        visit spree.edit_admin_order_path(@order)
+
+        expect(page).to have_no_content adjustment.label
+      end
```
sauloperez: Have you considered writing a controller test instead? the test does not involve any user interaction or more than a page so it should be possible AFAIK. You know the benefits already, improved reliability, execution speed, etc.

luisramos0: @sauloperez the filtering is done in the view

app/views/spree/admin/orders/_form/_adjustments.html.haml
```diff
@@ -3,12 +3,12 @@
...
       %thead
-        %th= Spree.t('name')
-        %th= Spree.t('amount')
-      %tbody#order-charges.with-border
+        %tr
+          %th= Spree.t('name')
+          %th= Spree.t('amount')
+      %tbody.with-border
         - adjustments.each do |adjustment|
-          - if (adjustment.originator_type != 'Spree::ShippingMethod') && !(adjustment.originator_type == 'Spree::TaxRate' && adjustment.amount == 0)
```
luisramos0: I think we can delete admin/orders_spec line 258:
`next if (adjustment.originator_type == 'Spree::TaxRate') && (adjustment.amount == 0)`

spec/features/admin/orders_spec.rb
sauloperez: I see, and that is why I suggest a controller test. By using `render_views`, which we might use by default, we hit the view too.

luisramos0: eheh, good that I commented, I learned something. I didnt know that render_views trick 👍

spec/features/admin/orders_spec.rb
mkllnk: I thought about that. But then this was much easier for several reasons:

- The translation key was missing. While rendering of the page has a fallback and uses the key name, that doesn't happen when calling `I18n.t` in the spec. I added the missing translation now and could use it.
- Some Spree logic (maybe the view) is printing it all in capitals. So I would need to use this: `I18n.t('spree.admin.orders.form.line_item_adjustments').upcase` which is long and hard to read.

But we have a convention, so I better obey. :wink:

spec/features/admin/orders_spec.rb
mkllnk: Good idea. Have a look at my last commit. We are saving 7 seconds here.

app/views/spree/admin/orders/_form/_adjustments.html.haml
mkllnk: Done.

spec/features/admin/orders_spec.rb
luisramos0: 🤣

spec/features/admin/orders_spec.rb
sauloperez: Wow! 7 seconds it's a whole lot!

RachL:
This is ready to go ! https://docs.google.com/document/d/1ane0lt3y67wDM1sokJJ7dXNhIUorwTaNA9ZtED_8yvU/edit#

PR #3670, opened 2019-03-30
[Spree Upgrade] Remove dependency to spree frontend - routes and controllers

### LINKED ISSUE #3676: Make OFN independent of Spree Frontend
We depend on spree frontend for a few things but most of it is overriden in ofn.
To make our code easier we should just move everything we need from spree frontend (just a few things missing) to ofn and remove the dependency to spree frontend.

there's #3669 that removed all deps to assets and views and then #3670 will remove deps to routes and controllers. I think in the end we will be bringing ~100 lines of code from spree frontend to OFN.

### Context
This comes from analysis and discussion here:
https://community.openfoodnetwork.org/t/bye-bye-spree-or-spree-v2-1-and-beyond/1598/5

### Impact and timeline
This will open up for some more clean up of the code on the OFN side.

luisramos0:
#### What? Why?

Continues what was started in #3669
Closes #3676

In this PR we:
- move all spree frontend routes to ofn's spree.rb
- transform spree frontend controller decorators into controllers
- move the missing code in the controllers from spree frontend to ofn
- remove spree_frontend and spree_sample from ofn's gemfile
This way we become 100% independent of spree_frontend routes and controllers.

The best part is that we can now delete all the duplication, the unnecessary bits and refactor to our needs.

#### What should we test?
There a few things we need to test here:
- cart - including empty cart scenario
- checkout including failed payment method
- shopfront (product list) - add to cart
- switching language
- product import: sanity check

#### Release notes
Changelog Category: Removed
OFN does not depend on spree frontend gem anymore :tada:

#### Discourse thread
This is discussed [here](https://community.openfoodnetwork.org/t/bye-bye-spree-or-spree-v2-1-and-beyond/1598).

#### Dependencies
This PR needs #3669 (already merged).


config/routes/spree.rb
```diff
@@ -94,6 +94,16 @@
...
 
+  # For some unknown reason this route /orders/populate and it's redirect
+  #   make product import routes work in a form in
+  #   app/views/admin/product_import/import.html.haml
+  #.  This route is otherwise dead
+  populate_redirect = redirect do |params, request|
+    request.flash[:error] = Spree.t(:populate_get_error)
+    request.referer || '/cart'
+  end
+  get '/orders/populate', :via => :get, :to => populate_redirect
```
mkllnk: Which spec failed? I just checked out your pull request without this commit and `spec/features/admin/product_import_spec.rb` passed. Or do you actually remove the frontend gem for testing?

luisramos0: I run the build against a version of spree where I simply deleted the folder frontend. The branch is here:
https://github.com/luisramos0/spree/tree/2-0-4-without-frontend

You can try it by changing the gemfile:
gem 'spree', github: 'luisramos0/spree', branch: '2-0-4-without-frontend'

luisramos0: I didnt see the problem again! so I removed the commit with the populate route :-D
I did use different versions of the spree branch in the process: removing just the views first, then the controllers, maybe that caused the problem. With this last branch (no frontend folder) the build is green without the populate route.
thanks for asking :-D

luisramos0: because you have suggested removing the gem I went ahead and tried replacing the spree dependency with individual dependencies to the different spree gems. it worked! the build is green, all looks good! #exciting

luisramos0: in this PR I am separating spree the different spree gems and all works fine.

in the final PR #3724 I remove the spree_frontend gem and get the same product import spec errors that are fixed by adding the order/populate routes, you can have a look here @mkllnk 
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-3724/builds/2

Gemfile
```diff
@@ -18,7 +18,11 @@ gem 'pg'
 # OFN-maintained and patched version of Spree v2.0.4. See
 # https://github.com/openfoodfoundation/openfoodnetwork/wiki/Spree-2.0-upgrade
 # for details.
-gem 'spree', github: 'openfoodfoundation/spree', branch: '2-0-4-stable'
+gem 'spree_core', github: 'openfoodfoundation/spree', branch: '2-0-4-stable'
+gem 'spree_api', github: 'openfoodfoundation/spree', branch: '2-0-4-stable'
+gem 'spree_backend', github: 'openfoodfoundation/spree', branch: '2-0-4-stable'
+gem 'spree_frontend', github: 'openfoodfoundation/spree', branch: '2-0-4-stable'
+gem 'spree_sample', github: 'openfoodfoundation/spree', branch: '2-0-4-stable'
```
sauloperez: I wonder what this is. Smells like something we can skip?

luisramos0: yeah, I remove it in the final PR #3724

luisramos0:
I spent all my conservative thinking while working for a bank 😺 
Now I always run forward... 🤣 

It's a bit painful and slow to delay this but I understand why you would want to hold it. We still have some good testing coming up for v2 though.
@mkllnk what do you think?

sauloperez:
I totally understand. Let's see what others think.

Matt-Yorkley:
In terms of rolling out v2 as smoothly as possible I'd probably vote for the more conservative option, but it's up to you guys. I trust you either way :+1:

config/routes/spree.rb
mkllnk: Interesting. Yesterday, when I started a local server and requested the home page for the first time, the little profiler in the top left corner displayed a product import request that took really long. I couldn't reproduce it and I have no idea how that is possible but we will find out if it's a real problem.

mkllnk:
I'm voting for the conservative approach as well. Let's get the first release done and rolled out. We have to accelerate our releases...

sauloperez:
Shall we move it to the post v2 column?

luisramos0:
I'd like to follow the normal review process before that. is that ok?

config/routes/spree.rb
mkllnk: I can reproduce it now. Thank you. I still don't understand it though. If I pry into the view and run the failing spec I can issue these commands:
```
> url_for(:controller=>"admin/product_import", :action=>"import")
ActionController::RoutingError: No route matches {:controller=>"admin/product_import", :action=>"import"}
from /home/maikel/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/actionpack-3.2.22.5/lib/action_dispatch/routing/route_set.rb:540:in `raise_routing_error'
Caused by Journey::Router::RoutingError: Journey::Router::RoutingError
from /home/maikel/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/journey-1.0.4/lib/journey/formatter.rb:46:in `generate'

> main_app.url_for(:controller=>"admin/product_import", :action=>"import")
=> "http://127.0.0.1:44929/admin/product_import"

> Rails.application.routes.url_for(:controller=>"admin/product_import", :action=>"import")
=> "http://test.host/admin/product_import"
```

The route is clearly there and can be found. But the current scope seems to have a `url_for` method that is looking in the wrong place.

```
> method(:url_for)
=> #<Method: #<Class:0x00562666a25de0>(ActionView::Helpers::UrlHelper)#url_for>
```

That method is calling super. But where is that code? I pry in there and then:

```
> self.class.ancestors.select { |c| c.method_defined?(:url_for) }
=> [#<Class:0x00562666a25de0>, ActionView::Base, ActionView::Helpers::UrlHelper, ActionDispatch::Routing::UrlFor]
```

The first class is `self.class`. So the `super` call goes to `ActionView::Base but I can't find a definition in there. Just this interesting piece of code:

```rb
      # This method receives routes and helpers from the controller
      # and return a subclass ready to be used as view context.
      def prepare(routes, helpers) #:nodoc:
        Class.new(self) do
          if routes
            include routes.url_helpers
            include routes.mounted_helpers
          end

          if helpers
            include helpers
            self.helpers = helpers
          end
        end
      end
```

Calling `ActionView::Base.new.url_for` leads to `ActionDispatch::Routing::UrlFor`:
```rb
          _routes.url_for((options || {}).symbolize_keys.reverse_merge!(url_options))
```

That `_routes` is provided by the controller:

```
> Admin::ProductImportController.ancestors.select { |c| c.method_defined?(:_routes) }
=> [Admin::ProductImportController,
 Spree::Admin::BaseController,
 Spree::BaseController,
 #<Module:0x0056265fd5e5b8>,
 ApplicationController,
 #<Module:0x005626627f85f8>,
 ActionController::Base,
 AbstractController::UrlFor]
```

And there we go:
```
> Spree::Admin::BaseController._routes.url_for(:controller=>"admin/product_import", :action=>"import")
ActionController::RoutingError: No route matches {:controller=>"admin/product_import", :action=>"import"}
```

So that Spree controller seems to contain only Spree routes while the ApplicationController has our routes as well:
```
> ApplicationController._routes.routes.map do |route|   {alias: route.name, path: route.path.spec.to_s, controller: route.defaults[:controller], action: route.defaults[:action]} end.select { |a| a[:path].starts_with? "/admin/product_import" }
=> [{:alias=>"admin_product_import", :path=>"/admin/product_import(.:format)", :controller=>"admin/product_import", :action=>"index"},
 {:alias=>nil, :path=>"/admin/product_import(.:format)", :controller=>"admin/product_import", :action=>"import"},
 {:alias=>"admin_product_import_process_async", :path=>"/admin/product_import/validate_data(.:format)", :controller=>"admin/product_import", :action=>"validate_data"},
 {:alias=>"admin_product_import_save_async", :path=>"/admin/product_import/save_data(.:format)", :controller=>"admin/product_import", :action=>"save_data"},
 {:alias=>"admin_product_import_reset_async", :path=>"/admin/product_import/reset_absent(.:format)", :controller=>"admin/product_import", :action=>"reset_absent_products"}]

> Spree::Admin::BaseController._routes.routes.map do |route|   {alias: route.name, path: route.path.spec.to_s, controller: route.defaults[:controller], action: route.defaults[:action]} end.select { |a| a[:path].starts_with? "/admin/product_import" }
=> []
```

So it looks like removing the Spree frontend changes the loading of routes. I also noticed that putting in the orders/populate route adds it to the overview in `rake routes` three times, in different places. It seems that it also triggers OFN routes to appear in the Spree admin routes. There are more controllers inheriting from `Spree::Admin::BaseController` which could be affected by this:

```
app/controllers/admin/bulk_line_items_controller.rb
app/controllers/admin/cache_settings_controller.rb
app/controllers/admin/contents_controller.rb
app/controllers/admin/invoice_settings_controller.rb
app/controllers/admin/manager_invitations_controller.rb
app/controllers/admin/matomo_settings_controller.rb
app/controllers/admin/product_import_controller.rb
app/controllers/admin/stripe_accounts_controller.rb
app/controllers/admin/stripe_connect_settings_controller.rb
app/controllers/spree/admin/base_controller_decorator.rb
app/controllers/spree/admin/invoices_controller.rb
```

For example:
```
> ApplicationController._routes.url_for(:controller=>"admin/bulk_line_items", :action=>"index")
=> "http://test.host/admin/bulk_line_items"

> Spree::Admin::BaseController._routes.url_for(:controller=>"admin/bulk_line_items", :action=>"index")
ActionController::RoutingError: No route matches {:controller=>"admin/bulk_line_items"}
```

I feel a bit uneasy to solve this problem by defining a random route without understanding how it works. Do you think you could find out more?

luisramos0: Dr Maikel the detective! Amazing. It would take me ages to get to this detail... and I am afraid it will take me ages fo figure this out. But yes, I think I need to have a look...

sauloperez:
Sure, I just meant after it gets the approvals.

config/routes.rb
```diff
@@ -27,6 +27,7 @@
   get "/cart", :to => "spree/orders#edit", :as => :cart
   put "/cart", :to => "spree/orders#update", :as => :update_cart
   put "/cart/empty", :to => 'spree/orders#empty', :as => :empty_cart
+  match '/orders/:id/token/:token' => 'spree/orders#show', :via => :get, :as => :token_order
```
sauloperez: if it's only via `:get` we better replace `match` with `get` as in line 27. If I'm not mistaken it is something that gets deprecated in Rails 4 and will need fixing. I remember dealing with `match` routes in the upgrade I did.

config/routes/spree.rb
```diff
@@ -86,4 +86,18 @@
...
+
+  resources :states, :only => :index
+  resources :countries, :only => :index
+
+  # route globbing for pretty nested taxon and product paths
+  match '/t/*id', :to => 'taxons#show', :as => :nested_taxons
+
+  match '/unauthorized', :to => 'home#unauthorized', :as => :unauthorized
+  match '/content/cvv', :to => 'content#cvv', :as => :cvv
+  match '/content/*path', :to => 'content#show', :via => :get, :as => :content
```
sauloperez: see my comment about `match` above

spec/controllers/enterprises_controller_spec.rb
```diff
@@ -109,7 +109,7 @@
       it "redirects to the cart" do
         spree_get :shop, {id: current_distributor}
 
-        response.should redirect_to spree.cart_path
+        response.should redirect_to cart_path
```
sauloperez: does it work without `main_app` here?

config/routes.rb
luisramos0: yeah, this is done [here](https://github.com/spree/spree/commit/c73dbbc49e57a17d38d1332e3a1ea2eb9af15667#diff-77e25f6625b6d4d28d71cdbe70021df8) in spree.

I added a commit to do it.

config/routes/spree.rb
luisramos0: done as well

spec/controllers/enterprises_controller_spec.rb
luisramos0: yes, it's working.

```
[107, 116] in /Users/luisramos/dev/ofn/openfoodnetwork/spec/controllers/enterprises_controller_spec.rb
   107:       end
   108:
   109:       it "redirects to the cart" do
   110:         spree_get :shop, {id: current_distributor}
   111:         byebug
=> 112:         response.should redirect_to cart_path
   113:       end
   114:     end
   115:
   116:     it "sets order cycle if only one is available at the chosen distributor" do
(byebug) cart_path
"/cart"                                                                                                                                                                                                                                                      |  ETA: ??:??:??
(byebug)
```

luisramos0:
@sauloperez @mkllnk this is ready for review again
except for the "routing issue with random workaround", I dont really know what to do about it...

config/routes/spree.rb
luisramos0: ah, I found the problem! it's not random.

form_tag false... in product import view is generating this html:
Screenshot 2019-05-28 at 12 29 28

so, it's calling the orders_populate route.
I dont understand why the orders populate route is picked up here, but it doesnt matter, we can just fix it by setting the correct action in this form which is simply
form_tag main_app.admin_product_import_path
I have added 2 commits from #3724 in this PR and added this new one. I will close #3724 now.
ready for review!

luisramos0:
rebased to resolve conflicts with #3670 
this is ready for review

sauloperez:
It'd be great if we could the chance to fix the simplest Code Climate errors. I understand there are others that should be fixed in a another PR.

luisramos0:
@sauloperez :+1: I have fixed the easy ones.

config/routes/spree.rb
```diff
@@ -86,4 +86,41 @@
     get :order_cycle_expired, :on => :collection
     put :cancel, on: :member
   end
+
+  root :to => 'home#index'
+
+  resources :products
+
+  match '/locale/set', :to => 'locale#set'
```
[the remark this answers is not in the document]
luisramos0: good point, I ignored the controller but left the route.
I have removed the route now in [this new commit](32e33775c1580fe1107f030d068f58649dd23569), we use a query param.

app/controllers/spree/orders_controller.rb
```diff
@@ -40,10 +36,6 @@ def empty
     redirect_to spree.cart_path
   end
 
-  def accurate_title
-    @order && @order.completed? ? "#{Spree.t(:order)} #{@order.number}" : Spree.t(:shopping_cart)
-  end
```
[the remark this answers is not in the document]
luisramos0: accurate_title is a method in spree/core/lib/spree/core/controller_helpers/common.rb
```
# this is a hook for subclasses to provide title
   51:           def accurate_title
   52              Spree::Config[:default_seo_title]
   53            end

```
the override for the users view would be in the users_controller (/account). this override in the orders controller has no impact. I had a look at both the pages account and order in the frontoffice and they look ok.

app/controllers/spree/checkout_controller.rb
```diff
@@ -31,56 +30,17 @@ def edit
...
-         (!params[:state] && !@order.has_checkout_step?(@order.state))
-        @order.state = 'cart'
-        redirect_to checkout_state_path(@order.checkout_steps.first)
-      end
-    end
-  end
-
-  # Should be overriden if you have areas of your checkout that don't match
-  # up to a step within checkout_steps, such as a registration step
-  def skip_state_validation?
```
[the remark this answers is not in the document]
luisramos0: ah, good catch!
I delete the stub [in this new commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/3670/commits/6511177d5ca6c1c70404dd7139cbab98a1d9d5f9)

app/controllers/spree/orders_controller.rb
```diff
@@ -102,8 +102,6 @@ def update
       discard_empty_line_items
       with_open_adjustments { update_totals_and_taxes }
 
-      render(:edit) && return unless apply_coupon_code
```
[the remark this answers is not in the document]
luisramos0: it's weird code but no, the apply coupon method returns true if coupons are not used:
https://github.com/openfoodfoundation/spree/blob/46d6f8f5fd434105b0c69958276d1727a3066a97/frontend/app/controllers/spree/store_controller.rb#L12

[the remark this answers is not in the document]
luisramos0: it's in the end:
```
else
    true
```

config/routes/spree.rb
```diff
@@ -86,4 +86,18 @@
     get :order_cycle_expired, :on => :collection
     put :cancel, on: :member
   end
+
+  resources :products
+
+  get '/locale/set', :to => 'locale#set'
```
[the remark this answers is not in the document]
luisramos0: yes, I removed this route, see above.

config/routes/spree.rb
```diff
@@ -86,4 +86,18 @@
...
+
+  resources :states, :only => :index
+  resources :countries, :only => :index
```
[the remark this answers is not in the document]
luisramos0: another good catch :-)
this is fixed in spree 2.1 [here](https://github.com/spree/spree/commit/ba420fe70e30dde75513b9b9b722f4b7c422e8cc).
I fixed it in [this new commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/3670/commits/88655be5f02fd195953c10c27b8a8b30ffc1dd04)

config/routes/spree.rb
```diff
@@ -86,4 +86,18 @@
...
+
+  # route globbing for pretty nested taxon and product paths
+  get '/t/*id', :to => 'taxons#show', :as => :nested_taxons
```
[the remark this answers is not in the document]
luisramos0: yeah, I just copied it over.
now that you mention it, I deleted it in [this new commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/3670/commits/28f2697982327ee0a57a94ab7f24bea52034f525)

luisramos0:
great @kristinalim thanks a lot for your detailed review. I addressed every comment, please have a look :-)

config/routes/spree.rb
mkllnk: Wow! I'm so happy that you found the cause! :tada: :man_dancing:

luisramos0:
Test Ready! So good!  🎉

sigmundpetersen:
Does this need a rebase before we test @luisramos0 ?

luisramos0:
yes, thanks!
rebased, ready for testing again.

RachL:
Everything seems good! Here are my testing notes:

https://docs.google.com/document/d/1W1sy3XA7UYQSq329OE0FejZ0LcWr7lSEcgX-q77YI9U/edit#

mkllnk:
Wow, this is huge! It's a bit like a burn off. Slightly risky but you have more fertile soil afterwards. And it's just the first step. Wait for dinner and dessert.

burn off

luisramos0:
ahaha
yeah, this is awesome 🎉 
Screenshot 2019-06-24 at 20 38 48

PR #3673, opened 2019-04-01
Spree2/css

### LINKED ISSUE #3577: [Spree Upgrade] Clear all button translation does not fit
## Description

As show in the screenshot, the Clear all button has a 100px fixed width which causes translations not to fit.

screenshot from 2019-03-05 11-47-23

It turns out that this doesn't not happen in v1, although the button has also a fixed width of 100px.

screenshot from 2019-03-05 17-05-58

## Expected Behavior

The button should not have a fixed width and rather adapt to the text.

## Actual Behaviour

It currently cuts the translation's text.

## Steps to Reproduce

1. Navigate to `/admin/order_cycles`

## Context

I was using the v2 version of the app in Catalan and I found it.

## Severity

## Your Environment

* Version used: 72e9c1d00

Matt-Yorkley:
#### What? Why?

Closes #3577

Made a few CSS tweaks to fix some minor display issues introduced by Spree 2.

#### What should we test?

Take a quick look at the admin pages for Order Cycles and Inventory. And the admin menu in general.

### DIFF
--- a/app/assets/stylesheets/admin/openfoodnetwork.css.scss
+++ b/app/assets/stylesheets/admin/openfoodnetwork.css.scss
@@ -2,6 +2,7 @@
 
 input[type="submit"], input[type="button"], button, .button {
   cursor: pointer;
+  padding: 6.5px 15px;
 }
 
 .text-center {
--- a/app/assets/stylesheets/admin/components/admin-menu.css.scss
+++ b/app/assets/stylesheets/admin/components/admin-menu.css.scss
@@ -6,6 +6,7 @@
 
     a::before {
       font-weight: normal;
+      padding-top: 0;
     }
   }
 }
--- a/app/assets/stylesheets/admin/components/admin-menu.css.scss
+++ b/app/assets/stylesheets/admin/components/admin-menu.css.scss
@@ -1,0 +1,11 @@
+#admin-menu {
+  li {
+    a {
+      font-weight: 600;
+    }
+
+    a::before {
+      font-weight: normal;
+    }
+  }
+}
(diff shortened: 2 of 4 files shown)


Matt-Yorkley:
**Before:**
Screenshot from 2019-04-01 11-40-45

**After:**
Screenshot from 2019-04-01 11-40-31

**Before:**
Screenshot from 2019-04-01 11-43-56

**After:**
Screenshot from 2019-04-01 11-47-19

**Before:**
Screenshot from 2019-04-01 12-09-57

**After:**
Screenshot from 2019-04-01 12-10-13

**Before:** 
Screenshot from 2019-04-01 12-17-14

**After:**
Screenshot from 2019-04-01 12-16-58

**Before:**
Screenshot from 2019-04-01 12-31-01

**After:**
Screenshot from 2019-04-01 12-30-40

app/assets/stylesheets/admin/openfoodnetwork.css.scss
```diff
@@ -2,6 +2,7 @@
 
 input[type="submit"], input[type="button"], button, .button {
   cursor: pointer;
+  padding: 6.5px 15px;
```
luisramos0: this is adding padding to all buttons in the backoffice right? I havent seen how it looks... are you sure this is what we need? We may need to test more pages to verify this is not breaking the structure in any other page.

Matt-Yorkley: I had a quick look around, and yes this looks right. The font has changed for the backoffice, and the new one renders using slightly taller characters, which means the overall height of all buttons has increased slightly, making them look strange next to some of the other elements that have a fixed height that hasn't increased. This adjusts the padding a _tiny_ bit so they look like they did before and fit in visually with other elements.

luisramos0: awesome!

RachL:
@Matt-Yorkley I've tried to stage this one but it failed

RachL:
I navigated both in French and Italian, and it seems to be working good. Ready to go! A few examples:

image

image

image

PR #3674, opened 2019-04-01
Allow import to proceed when updating a product and `description` is set

### LINKED ISSUE #3664: [Spree upgrade] Better use of description field in Product Import
## Description
Description can be created but not updated through product import. This is causing problem when product have italic or bold text, that cannot be translated in the csv easily by a user (except if the user knows html).

Further more, it may have prevented to find products for inventory creation here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3315

## Steps to Reproduce
1. Use file product_listv2_test2 https://drive.google.com/drive/u/0/folders/1fI0r3CuHEs0e2y3N46QHIUPJkMs4mfrf on AU staging
2. You be able to update those products as description field is not matching the description in the DB

## Your Environment
* Version used: v2
* Browser name and version: Firefox 65

## Possible Fix
Not check the description field during upload?

Matt-Yorkley:
#### What? Why?

Closes #3664 

Allows import to proceed without errors when updating a product and the `description` field is set. The field can be set on creation, but not updated via import, so it will be ignored if updating a product.

#### What should we test?

Products can be updated when importing and the description field is set.

### DIFF
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -4,6 +4,8 @@
 
 module ProductImport
   class EntryValidator
+    SKIP_VALIDATE_ON_UPDATE = [:description].freeze
+
     # rubocop:disable Metrics/ParameterLists
     def initialize(current_user, import_time, spreadsheet_data, editable_enterprises,
                    inventory_permissions, reset_counts, import_settings, all_entries)
@@ -326,8 +328,7 @@ module ProductImport
     end
 
     def ignore_when_updating_product?(attribute)
-      attributes_to_ignore = [:description]
-      attributes_to_ignore.include? attribute
+      SKIP_VALIDATE_ON_UPDATE.include? attribute
     end
 
     def convert_to_trusted_type(untrusted_attribute, trusted_attribute)
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -314,7 +314,7 @@ module ProductImport
     def product_field_errors(entry, existing_product)
       EntryValidator.non_updatable_fields.each do |display_name, attribute|
         next if attributes_match?(attribute, existing_product, entry) || attributes_blank?(attribute, existing_product, entry)
-        next if attribute == :description
+        next if ignore_when_updating_product?(attribute)
         mark_as_invalid(entry, attribute: display_name, error: I18n.t('admin.product_import.model.not_updatable'))
       end
     end
@@ -323,6 +323,11 @@ module ProductImport
       existing_product_value = existing_product.public_send(attribute)
       entry_value = entry.public_send(attribute)
       existing_product_value == convert_to_trusted_type(entry_value, existing_product_value)
+    end
+
+    def ignore_when_updating_product?(attribute)
+      attributes_to_ignore = [:description]
+      attributes_to_ignore.include? attribute
     end
 
     def convert_to_trusted_type(untrusted_attribute, trusted_attribute)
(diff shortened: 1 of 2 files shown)


app/models/product_import/entry_validator.rb
```diff
@@ -282,6 +282,7 @@ def mark_as_existing_variant(entry, existing_variant)
     def product_field_errors(entry, existing_product)
       EntryValidator.non_updatable_fields.each do |display_name, attribute|
         next if attributes_match?(attribute, existing_product, entry) || attributes_blank?(attribute, existing_product, entry)
+        next if attribute == :description
```
luisramos0: I think this requires a spec.

kristinalim: This line would also be clearer if the condition is moved to a method with a more descriptive name, e.g. `ignore_attribute_when_updating_product?(attribute)`.

And then it is better to check this earlier in the block, so there is no need to compare values in `existing_product` and `entry` if the attribute should be skipped.

sauloperez:
Ping @Matt-Yorkley

Matt-Yorkley:
Test added and attribute check moved to a readable method. :+1:

spec/models/product_importer_spec.rb
```diff
@@ -247,6 +247,25 @@
     end
   end
 
+  describe "updating an exiting variant" do
+    before do
+      csv_data = CSV.generate do |csv|
+        csv << ["name", "producer", "description" ,"category", "on_hand", "price", "units", "unit_type", "display_name", "shipping_category"]
+        csv << ["Hypothetical Cake", "Another Enterprise", "New Description", "Cake", "5", "5.50", "500", "g", "Preexisting Banana", shipping_category.name]
+      end
+      @importer = import_data csv_data
```
sauloperez: I'd rather stick with `let` like other tests. See http://www.betterspecs.org/#let

app/models/product_import/entry_validator.rb
```diff
@@ -298,6 +299,11 @@ def attributes_match?(attribute, existing_product, entry)
       existing_product_value == convert_to_trusted_type(entry_value, existing_product_value)
     end
 
+    def ignore_when_updating_product?(attribute)
+      attributes_to_ignore = [:description]
```
sauloperez: I know this is a micro-optimization but these things are better suited as a constant so they are evaluated only once and not every single time the method is called. Something like 

```ruby
class EntryValidator
  ATTRIBUTES_TO_IGNORE = [:description].freeze
  (...)
  def ignore_when_updating_product?(attribute)
    ATTRIBUTES_TO_IGNORE.include? attribute
  end
end
```

RachL:
@Matt-Yorkley On Katuma's staging, I'm trying to upload this file : https://framadrop.org/r/JDBty-GmYE#j6PZpNPCb0EfkG0Q3CxVFEmcED/waHEIgwKjENjAXww=

But the import is stuck at:

image

Any idea what's the pb?

Matt-Yorkley:
@RachL I just tried to open your file...

CSV

Mã hóa này là gì?!

Matt-Yorkley:
There's also lots of these Microsoft error messages embedded in the file:

valeur

RachL:
(✖╭╮✖)

@Matt-Yorkley I rekon I used a file that someone else created, but I've already uploaded it on several staging server. For this test I only changed the producer's name and categories, and I did so by using Libre Office on Ubuntu making sure I was saving on the right format.

I just re-tried on Katuma staging (changing absolutely nothing) and it worked right away (apparently the first time as well):
image

I don't know what has happened the first time... I've tested changing info and everything is uploading correctly... leaving the description untouched. I guess we are ready to go with this one.

PR #3689, opened 2019-04-02
[Spree Upgrade] Fix split button reappearing when editing line item quantity

### LINKED ISSUE #3683: [Spree Upgrade] Split order button visible
## Description

When editing an individual order, when you click to edit a line item the 'split' button becomes visible. This needs to be hidden in v2.

## Expected Behavior

Editing a variant in an order should allow us to edit the qty but not to split the order.

## Actual Behaviour

The user is given the option to split the order.

## Steps to Reproduce

(See the gif for more clarity)
1. Orders tab -> Edit order
2. Order details -> Edit line item
3. Click and the split button appears

## Animated Gif/Screenshot

splitappears

## Severity

S3
## Your Environment

* Version used: v2

## Possible Fix

Hide the split button.

luisramos0:
#### What? Why?

Closes #3683 

#3683 is a bug in PR #3548 in [this commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/3548/commits/37685c8107ac1cc59ef9e3f866e70cbbe813c111).
We need to remove the split button from the dom so that when spree does a toggle on it [here](https://github.com/openfoodfoundation/spree/blob/060d0a5d8b31f68990fa792e0e88910c742d9a96/backend/app/assets/javascripts/admin/variant_autocomplete.js.erb#L112), the button doesnt re-appear.

#### What should we test?
Make sure 3683 is fixed.

### DIFF
--- a/app/assets/javascripts/admin/orders/controllers/order_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/order_controller.js.coffee
@@ -24,3 +24,8 @@ angular.module("admin.orders").controller "orderCtrl", ($scope, shops, orderCycl
 
   for shop in $scope.shops
     shop.disabled = !$scope.distributorHasOrderCycles(shop)
+
+  # Removes the split button introduced by spree in the order form
+  #   We only have one stock location in OFN so it's meaningless to split the order between stock locations
+  #   We delete it instead of hiding or changing CSS so that, when spree code toggles the element, nothing hapens
+  $('.split-item').remove()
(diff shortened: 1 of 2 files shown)


RachL:
Yes #3683 works! 

image

I've tried it on fresh orders and past orders. And I understood how I've missed it previously : I must have forgotten to test with super admin access... and those buttons are only available for super admin. Sorry :(

luisramos0:
awesome!

Issue #3691, opened 2019-04-03
Release v1.30


luisramos0:
TBD

sauloperez:
We'll prepare the release today and test it throughout the week.

sauloperez:
This is the build we'll use: https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/master/builds/586

sauloperez:
Release ready to be tested at https://github.com/openfoodfoundation/openfoodnetwork/releases/tag/untagged-f58a56110720c5e7a9e5 cc @RachL

RachL:
Here are my testing notes:
https://docs.google.com/document/d/1Z6JBWjQhVxZGCPzF2hecir8ePKVbmevsxHOA22U6zIU/edit#

I've introduced for the first time Paypal and BOM in the test release process.

I discovered 3 things but I don't think they are blocking the release:

- @sauloperez while retesting #3644 I've noticed that the variant does not disappear directly from the shop front. It takes some time (even incognito, with no cookies/cache etc) Is this something to considered?
- I've noticed that you could order through subs a product with insufficient stock. The stock of the product will then be negative. I have no idea if it made on purpose. I will open a separate issue for this.
- I couldn't checkout with my fake paypal visa card, only with "paypal balance". I think it is linked to my my fake buyer paypal account. I need to investigate. @luisramos0 when you tested did you used the visa card or the balance option?

luisramos0:
great, thanks a lot for all the testing @RachL 
re paypal, I cant remember!! I think I payed with my real paypal account and my real credit card!!! But on a sandbox.paypal.com URL... I am not sure...

sauloperez:
> I couldn't checkout with my fake paypal visa card, only with "paypal balance". I think it is linked to my fake buyer paypal account. I need to investigate. @luisramos0 when you tested did you used the visa card or the balance option?

You mean with one of those test credit card numbers they list?

luisramos0:
@sauloperez this needs to be looked at before releasing: https://github.com/openfoodfoundation/openfoodnetwork/pull/3694#pullrequestreview-224246651

RachL:
@sauloperez I don't know if paypal gived a list of credit card like stripe, but I only tested with the fake visa card that gets created with a dev account.

RachL:
Ah and yes #3639 definitely! Will change this in my post

sauloperez:
[...]

PR #3694, opened 2019-04-04
Fix outdated db2fog initializer

Matt-Yorkley:
#### What? Why?

Part one of: https://github.com/openfoodfoundation/ofn-install/issues/280

Updates the old deprecated initializer and uses the correct values provisioned by ofn-install here: https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/application.yml.j2

#### Release notes

Moved db2fog initializer into the main repo from ofn-install

Changelog Category: Changed


luisramos0:
oops, build broken:
```
NoMethodError: s3_backups_bucket preference not defined
/home/runner/openfoodnetwork/vendor/bundle/ruby/2.1.0/bundler/gems/spree-41906362d931/core/app/models/spree/preferences/preferable.rb:60:in `has_preference!'
/home/runner/openfoodnetwork/vendor/bundle/ruby/2.1.0/bundler/gems/spree-41906362d931/core/app/models/spree/preferences/preferable.rb:33:in `get_preference'
/home/runner/openfoodnetwork/config/initializers/db2fog.rb:5:in `<top (required)>'
/home/runner/openfoodnetwork/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22.5/lib/active_support/dependencies.rb:245:in `load'
/home/runner/openfoodnetwork/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22.5/lib/active_support/dependencies.rb:245:in `block in load'
/home/runner/openfoodnetwork/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22.5/lib/active_support/dependencies.rb:236:in `load_dependency'
/home/runner/openfoodnetwork/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22.5/lib/active_support/dependencies.rb:245:in `load'
/home/runner/openfoodnetwork/vendor/
```

Matt-Yorkley:
Well that was unexpected! Ok, it looks like Spree has a predefined key for storing that preference, and we've been using it wrong. `Spree::Config[:s3_backups_bucket]` should be `Spree::Config[:s3_bucket]`.

config/initializers/db2fog.rb
```diff
@@ -0,0 +1,7 @@
+# See: https://github.com/yob/db2fog
+DB2Fog.config = {
+    :aws_access_key_id     => Spree::Config[:s3_access_key],
```
sauloperez: so if these keys are empty does it fail loudly? or it simply skips?

Matt-Yorkley: I was considering wrapping it in an `if` statement and checking `Spree::Config[:s3_bucket]` is set... what's the best way to do this?

Matt-Yorkley: We have a cron job for db2fog in `scedule.rb` here: https://github.com/openfoodfoundation/openfoodnetwork/blob/07888b355a2888e118d6d59ce968ab88125ebadc/config/schedule.rb#L28

We don't configure this anywhere with ofn-install as far as I can tell, so this job always runs.

Matt-Yorkley: Looking at how the previous deprecated initializer was being loaded, and how things are set on servers that don't use S3 backups, I'd say that it quietly ignores or skips if the initializer settings are empty. 

I don't think this will be any different to the way it's being loaded currently, which is this: https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/initializers/0_depricated_db2fog.rb

Is that right?

Matt-Yorkley: The current initializer is prexifed with a zero in the filename so it gets loaded first, then any other initializer that gets added after (by ofn-install) that does the same configuration will overwrite it. That's the current setup. So we're keeping the exact same logic and initialization, but removing all the gymnastics performed in ofn-install to add a second initializer that overrides the first.

sauloperez: > We don't configure this anywhere with ofn-install as far as I can tell, so this job always runs.

And it always fails.

I think the simplest is to add the conditional on schedule.rb so that we don't schedule useless jobs, but IMO it is a separate issue that can be tackled in a separate PR.

Matt-Yorkley:
The deploying without secrets revolution has begun! :tada:

config/initializers/db2fog.rb
```diff
@@ -0,0 +1,7 @@
...
+    :aws_secret_access_key => Spree::Config[:s3_secret],
+    :directory             => Spree::Config[:s3_bucket],
```
mkllnk: :alarm_clock: Does this mean we are uploading our database to the public bucket where our assets are stored? The reason why there were two config files is that we have the Spree config which is for assets (public bucket) and a second config (possibly a different access key) to upload backups to a private bucket. Maybe I missed something but this looks like we completely messed it up.

luisramos0: 👀 that means the release needs to be blocked: #3691

Matt-Yorkley: I can confirm UK db backups are not public, but there are a number of issues here that need to be looked at. As far as I can tell they are pre-existing.

PR #3697, opened 2019-04-04
Add delayed job web

### LINKED ISSUE #3622: Add Delayed Job Web
## Description

We need to make it easier to check the behavior of Delayed Job in a given server and perform basic operations like retries or deletion of jobs.

A good way to do so is by setting up [Delayed Job Web](https://github.com/ejschmitt/delayed_job_web). This will be particularly useful for stagings where we try and mess up with things.

I raised the topic [ages ago](https://community.openfoodnetwork.org/t/delayed-job-monitoring/1099/2) but it never got into the priorities.

Check out https://community.openfoodnetwork.org/t/delayed-job-monitoring/1099/2 for a description of this web UI features.

Something to take into account is the authentication. Not everyone can have access to that UI meant for the core team.

sauloperez:
#### What? Why?

Closes #3622 

This adds https://github.com/ejschmitt/delayed_job_web to enable async job management such as retries, removal of failed jobs, etc.

Only superadmins can access this web UI though. Therefore any sysadmin that needs to use it should have access to the appropriate instance superadmin credentials in Bitwarden.

**For the UI to be accessible you need to log in as superadmin first**

Screenshot from 2019-04-04 16-03-17

#### What should we test?
As superadmin you should be able to access `/admin/delayed_job`. For the UI to be accessible you need to log in first.

#### Release notes

Add Delayed Job Web to enable delayed job management to superadmins through a web UI.

Changelog Category: Added

#### Discourse thread

https://community.openfoodnetwork.org/t/making-operations-a-first-class-citizen/1601

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem 'jwt', '~> 2.1'
 
 gem 'delayed_job_active_record'
 gem 'daemons'
+gem 'delayed_job_web'
 
 # Fix bug in simple_form preventing collection_check_boxes usage within form_for block
 # When merged, revert to upstream gem
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -277,6 +277,11 @@ GEM
     delayed_job_active_record (4.1.3)
       activerecord (>= 3.0, < 5.3)
       delayed_job (>= 3.0, < 5)
+    delayed_job_web (1.4.3)
+      activerecord (> 3.0.0)
+      delayed_job (> 2.0.3)
+      rack-protection (>= 1.5.5)
+      sinatra (>= 1.4.4)
     devise (2.2.8)
       bcrypt-ruby (~> 3.0)
       orm_adapter (~> 0.1)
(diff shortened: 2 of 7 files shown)


sauloperez:
> Did we switch the default dev branch to 2-0-stable already?

Whatever you guys decide. You're onto it. But I agree with you that it doesn't affect much. Even in this case it'd good to include it in v1.30.0

luisramos0:
yeah, I think we should only allow simple/small PRs like this one into master, otherwise it's too much trouble merging them and testing them in both branches...

RachL:
This is cool!

image

sauloperez:
@luisramos0 @mkllnk @Matt-Yorkley @kristinalim IMO this should make it to a v1.31.0. Even @RachL found it useful already. Thoughts?

luisramos0:
There's a cost to releasing a new version. I'd focus on moving issues forward in v2. Just my opinion.

sauloperez:
sure. It's just that we do have issues and I could use something like DJ web. I want to be able to have proper sleep after we deploy v2 for the first time.

luisramos0:
in #3742 I suggest we release v1.31.0 to al instances when we release v2.0.0 to Katuma ;-)

Issue #3685, opened 2019-04-02
[Spree Upgrade] Remove location field from Add Line Item to Order


lin-d-hop:
## Description

When adding a line item to an order I am presented with a table in v2 with information on location, qty on hand and the qty I choose to add to the order. As OFN is not making use of stock locations the locations field is superfluous and a confusing experience for the user. 

The proposal is to remove it.

## Expected Behavior

The table should only display 'On Hand' field and 'Qty' field

## Actual Behaviour

The 'Location' field is also displayed

## Steps to Reproduce

1. Orders -> Edit Order
2. Order Details -> Add line item to order
3. See the 'Location' field in the table

## Animated Gif/Screenshot

![addlineitems](https://user-images.githubusercontent.com/8669848/55407033-2dc55900-5555-11e9-8b57-94c3d3de5bb1.gif)

## Severity

S4
## Your Environment

* Version used: v2
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): Ubuntu
* OFN Platform instance where you discovered the bug, and which version of the software they are using. 

## Possible Fix

Hide the field from user view.

PR #3701, opened 2019-04-05
[Spree Upgrade] Fix two issues in admin orders edit page

### LINKED ISSUE #3684: [Spree Upgrade] Translation missing - add new line item to existing order
## Description

Missing translation when adding a line item to an existing order.

## Expected Behavior

Should use one of the existing keys for this: "admin.on_demand?"

## Actual Behaviour

Current key is on_demand - could just be a missing ?

## Steps to Reproduce

1. Orders -> Edit Order
2. Order Details -> Add line item to order
3. See the missing translation error.

## Animated Gif/Screenshot

addlineitems

## Severity

S4
## Your Environment

* Version used: v2
* Operating System and version (desktop or mobile): Ubunutu

luisramos0:
#### What? Why?

Closes #3684 and #3685 
Removes stock location from stock levels table and fix Spree.t entry bug.

This is the lazy man quick fix, this autocomplete.erb file needs to be converted to haml and the js code moved to angular. Next time... 🙏 

#### What should we test?
Order edit page works as usual, stock location column is not there and if variant is on demand, it shows the translation, not the Spree.t command.

### DIFF
--- a/app/views/spree/admin/variants/_autocomplete.js.erb
+++ b/app/views/spree/admin/variants/_autocomplete.js.erb
@@ -34,23 +34,20 @@
     <legend align="center"><%= Spree.t(:select_stock) %></legend>
       <table class="stock-levels" data-hook="stock-levels">
         <colgroup>
+          <col style="width: 60%;" />
           <col style="width: 30%;" />
-          <col style="width: 40%;" />
-          <col style="width: 20%;" />
           <col style="width: 10%;" />
         </colgroup>
         <thead>
-          <th><%= Spree.t(:location) %></th>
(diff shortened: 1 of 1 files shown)


luisramos0:
I want to skip it for now because it's not so easy. This template is loaded with spree js code. it would have to be a feature spec I think. I have added this to https://github.com/openfoodfoundation/openfoodnetwork/issues/3440

RachL:
This seems good to go!

My testing notes:
https://docs.google.com/document/d/1B6uA4BFsQcfKLObuaDkjkrb0SeQ4kxA8JPHKQLFer1I/edit#

PR #3709, opened 2019-04-07
Start upgrade to Rails 4

### LINKED ISSUE #3708: [Spree 2.1] Adapt to Rails 4 - subquery has too many columns
Looks like that in rails 4 we need to be a it more specific when using IN operator otherwise we get this type of error:
```
71) Spree::Product scopes access roles shows only products for given user
      Failure/Error: product.count.should == 1
      
        PG::SyntaxError: ERROR:  subquery has too many columns
        LINE 1: ...e_products"."deleted_at" IS NULL AND (supplier_id IN (SELECT...
                                                                     ^
        : SELECT COUNT(*) FROM "spree_products"  WHERE "spree_products"."deleted_at" IS NULL AND (supplier_id IN (SELECT "enterprises".* FROM "enterprises" INNER JOIN "enterprise_roles" ON "enterprises"."id" = "enterprise_roles"."enterprise_id" WHERE "enterprise_roles"."user_id" = $1))
      # ./spec/models/spree/product_spec.rb:326:in `block (4 levels) in <module:Spree>'
      # ------------------
      # --- Caused by: ---
      # PG::SyntaxError:
      #   ERROR:  subquery has too many columns
      #   LINE 1: ...e_products"."deleted_at" IS NULL AND (supplier_id IN (SELECT...
      #                                                                ^
      #   ./spec/models/spree/product_spec.rb:326:in `block (4 levels) in <module:Spree>'
```

In this case, we can have
supplier_id IN (SELECT "enterprises".*
we need to have 
supplier_id IN (SELECT "enterprises"."id"

I couldnt find reference to this here:
https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0

luisramos0:
#### What? Why?

Starts #3708
This makes some of our code in models compatible with rails 4 by making scopes select the appropriate field when using the IN operator.

This PR does not close the issue, there are quite a few more places where we have to do this, this PR serves as a sample of what we need to do so we can validate the approach and do the same for the rest of the codebase.

#### map, select or pluck
The trick here is to know when to use map, select or pluck. With this PR I have learned a few things including [how CollectionProxy works](https://technology.customink.com/blog/2014/10/07/rails-association-proxies/)!
Relations have the methods select and pluck. Arrays have the map method.
A scope returns a relation but an association returns a CollectionProxy that has both methods from a relation select and pluck plus all the array methods including map. So we can use map on an association but not on a scope.
The challenge is that select returns a relation while pluck returns an array. So, when we use pluck we can't really chain relations and we create an extra query just for that pluck statement.
SO, when we are handling relations that will be chained or used in other queries, we need to use select.

#### What should we test?
This 2-0 build should be green.
This code can also be seen in the 2-1-stable where it fixes a few specs.

I think we will be ok with a sanity check. Maybe we can test adding variants to an order cycle and removing them and make sure the shopfront reacts correctly. Making sure the products cache is active.

#### Release notes
Changelog Category: Changed
Started adapting our codebase to Rails 4.

Upgrading our code to rails 4 is part of the upgrade to spree 2-1.

### DIFF
--- a/app/controllers/spree/admin/base_controller_decorator.rb
+++ b/app/controllers/spree/admin/base_controller_decorator.rb
@@ -52,7 +52,7 @@ Spree::Admin::BaseController.class_eval do
   def active_distributors_not_ready_for_checkout
     ocs = OrderCycle.managed_by(spree_current_user).active
     distributors = ocs.map(&:distributors).flatten.uniq
-    Enterprise.where('id IN (?)', distributors).not_ready_for_checkout
+    Enterprise.where('enterprises.id IN (?)', distributors).not_ready_for_checkout
   end
 
   def active_distributors_not_ready_for_checkout_message(distributors)
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -111,7 +111,7 @@ class Enterprise < ActiveRecord::Base
     joins(:shipping_methods).
       joins(:payment_methods).
       merge(Spree::PaymentMethod.available).
-      select('DISTINCT enterprises.*')
+      select('DISTINCT enterprises.id')
   }
   scope :not_ready_for_checkout, lambda {
     # When ready_for_checkout is empty, return all rows when there are no enterprises ready for
@@ -165,14 +165,14 @@ class Enterprise < ActiveRecord::Base
       select('DISTINCT enterprises.*')
   }
 
-  scope :distributing_products, lambda { |products|
+  scope :distributing_products, lambda { |product_ids|
     exchanges = joins("
         INNER JOIN exchanges
           ON (exchanges.receiver_id = enterprises.id AND exchanges.incoming = 'f')
       ").
       joins('INNER JOIN exchange_variants ON (exchange_variants.exchange_id = exchanges.id)').
       joins('INNER JOIN spree_variants ON (spree_variants.id = exchange_variants.variant_id)').
-      where('spree_variants.product_id IN (?)', products).select('DISTINCT enterprises.id')
+      where('spree_variants.product_id IN (?)', product_ids).select('DISTINCT enterprises.id')
 
     where(id: exchanges)
   }
@@ -449,7 +449,7 @@ class Enterprise < ActiveRecord::Base
   end
 
   def touch_distributors
-    Enterprise.distributing_products(supplied_products).
+    Enterprise.distributing_products(supplied_products.select(:id)).
       where('enterprises.id != ?', id).
       find_each(&:touch)
   end
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -111,12 +111,12 @@ class Enterprise < ActiveRecord::Base
     joins(:shipping_methods).
       joins(:payment_methods).
       merge(Spree::PaymentMethod.available).
-      select('enterprises.id')
+      select('DISTINCT enterprises.*')
   }
   scope :not_ready_for_checkout, lambda {
     # When ready_for_checkout is empty, return all rows when there are no enterprises ready for
     # checkout.
-    ready_enterprises = Enterprise.ready_for_checkout
+    ready_enterprises = Enterprise.ready_for_checkout.select('enterprises.id')
     if ready_enterprises.present?
       where("enterprises.id NOT IN (?)", ready_enterprises)
     else
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -111,14 +111,14 @@ class Enterprise < ActiveRecord::Base
     joins(:shipping_methods).
       joins(:payment_methods).
       merge(Spree::PaymentMethod.available).
-      select('DISTINCT enterprises.id')
+      select('enterprises.id')
   }
   scope :not_ready_for_checkout, lambda {
     # When ready_for_checkout is empty, return all rows when there are no enterprises ready for
     # checkout.
     ready_enterprises = Enterprise.ready_for_checkout
     if ready_enterprises.present?
-      where("id NOT IN (?)", ready_enterprises)
+      where("enterprises.id NOT IN (?)", ready_enterprises)
     else
       where("TRUE")
     end
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -27,17 +27,30 @@ class OrderCycle < ActiveRecord::Base
 
   preference :product_selection_from_coordinator_inventory_only, :boolean, default: false
 
-  scope :active, lambda { where('order_cycles.orders_open_at <= ? AND order_cycles.orders_close_at >= ?', Time.zone.now, Time.zone.now) }
+  scope :active, lambda {
+    where('order_cycles.orders_open_at <= ? AND order_cycles.orders_close_at >= ?',
+          Time.zone.now,
+          Time.zone.now)
+  }
   scope :active_or_complete, lambda { where('order_cycles.orders_open_at <= ?', Time.zone.now) }
-  scope :inactive, lambda { where('order_cycles.orders_open_at > ? OR order_cycles.orders_close_at < ?', Time.zone.now, Time.zone.now) }
+  scope :inactive, lambda {
+    where('order_cycles.orders_open_at > ? OR order_cycles.orders_close_at < ?',
+          Time.zone.now,
+          Time.zone.now)
+  }
   scope :upcoming, lambda { where('order_cycles.orders_open_at > ?', Time.zone.now) }
-  scope :not_closed, lambda { where('order_cycles.orders_close_at > ? OR order_cycles.orders_close_at IS NULL', Time.zone.now) }
-  scope :closed, lambda { where('order_cycles.orders_close_at < ?', Time.zone.now).order("order_cycles.orders_close_at DESC") }
+  scope :not_closed, lambda {
+    where('order_cycles.orders_close_at > ? OR order_cycles.orders_close_at IS NULL', Time.zone.now)
+  }
+  scope :closed, lambda {
+    where('order_cycles.orders_close_at < ?',
+          Time.zone.now).order("order_cycles.orders_close_at DESC")
+  }
   scope :undated, -> { where('order_cycles.orders_open_at IS NULL OR orders_close_at IS NULL') }
   scope :dated, -> { where('orders_open_at IS NOT NULL AND orders_close_at IS NOT NULL') }
 
   scope :soonest_closing,      lambda { active.order('order_cycles.orders_close_at ASC') }
-  # TODO This method returns all the closed orders. So maybe we can replace it with :recently_closed.
+  # This scope returns all the closed orders
   scope :most_recently_closed, lambda { closed.order('order_cycles.orders_close_at DESC') }
 
   scope :soonest_opening,      lambda { upcoming.order('order_cycles.orders_open_at ASC') }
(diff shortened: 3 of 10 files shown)


mkllnk:
@luisramos0 I updated the description because Github thought that you were saying that the issue is closed by this PR. It doesn't understand negation as in "it doesn't close #1001110".

app/models/enterprise.rb
```diff
@@ -91,15 +91,15 @@ class Enterprise < ActiveRecord::Base
     joins(:shipping_methods).
       joins(:payment_methods).
       merge(Spree::PaymentMethod.available).
-      select('DISTINCT enterprises.*')
+      select('enterprises.id')
```
sauloperez: But this is not equivalent. Where we were getting all columns, now we just get the id.

app/controllers/spree/admin/base_controller_decorator.rb
```diff
@@ -53,7 +53,7 @@ def unauthorized
   def active_distributors_not_ready_for_checkout
     ocs = OrderCycle.managed_by(spree_current_user).active
     distributors = ocs.map(&:distributors).flatten.uniq
-    Enterprise.where('id IN (?)', distributors).not_ready_for_checkout
+    Enterprise.where('enterprises.id IN (?)', distributors).not_ready_for_checkout
```
sauloperez: Why do we need the table? AR infers it from the class you are calling to.

app/models/enterprise.rb
```diff
@@ -135,14 +135,14 @@ class Enterprise < ActiveRecord::Base
...
-  scope :distributing_products, lambda { |products|
+  scope :distributing_products, lambda { |product_ids|
     exchanges = joins("
         INNER JOIN exchanges
           ON (exchanges.receiver_id = enterprises.id AND exchanges.incoming = 'f')
       ").
       joins('INNER JOIN exchange_variants ON (exchange_variants.exchange_id = exchanges.id)').
       joins('INNER JOIN spree_variants ON (spree_variants.id = exchange_variants.variant_id)').
-      where('spree_variants.product_id IN (?)', products).select('DISTINCT enterprises.id')
+      where('spree_variants.product_id IN (?)', product_ids).select('DISTINCT enterprises.id')
```
sauloperez: Thanks! I like it when things are concise

app/models/enterprise.rb
```diff
@@ -408,7 +408,7 @@ def initialize_permalink
   end
 
   def touch_distributors
-    Enterprise.distributing_products(self.supplied_products).
+    Enterprise.distributing_products(self.supplied_products.select(:id)).
```
sauloperez: We don't need self here, so we could remove it while we are at it.

app/models/order_cycle.rb
```diff
@@ -47,7 +47,7 @@ class OrderCycle < ActiveRecord::Base
     if user.has_spree_role?('admin')
       scoped
     else
-      where('coordinator_id IN (?)', user.enterprises)
+      where('coordinator_id IN (?)', user.enterprises.map(&:id))
```
mkllnk: So your guess is that enterprises are loaded already and we don't need to start a special query here? Otherwise this would be more efficient:
```rb
where('coordinator_id IN (?)', user.enterprises.select("enterprises.id"))
```
It's difficult to judge in a scope though. We don't know who will be using the scope and what else they are doing.

Uh, I just realised that we are using the result in a sub-query here. So there is no additional query with `select`. It will just be inserted, right?
```sql
WHERE coordinator_id IN (SELECT enterprises.id FROM "enterprises" INNER JOIN ...)
```

In that case we potentially avoid querying the enterprises. In case of a big enterprise list, the sub query could also be much smaller than a long list of ids. At least, it would scale better.

Your thoughts on this?

sauloperez: Have we checked we are not affecting performance here? When passing the association I guess AR calls `enterprises_ids` which will do clever things, whereas with map we load the entire association and then iterate it.

I'd like to understand what's the actual change we are doing.

app/models/order_cycle.rb
```diff
@@ -222,7 +222,7 @@ def exchanges_carrying(variant, distributor)
   end
 
   def exchanges_supplying(order)
-    exchanges.supplying_to(order.distributor).with_any_variant(order.variants)
+    exchanges.supplying_to(order.distributor).with_any_variant(order.variants.map(&:id))
```
mkllnk: Same here. I'm not sure what performs better at the moment but I would guess that `select` scales better than `map`, especially because we use the result in sub queries and the database can handle it very efficiently.

app/controllers/spree/admin/base_controller_decorator.rb
luisramos0: because of the scope not_ready_for_checkout the query errors out with "ambiguous field id". this change fixes that problem in 2-1-stable.

app/models/enterprise.rb
luisramos0: ok, done

app/models/order_cycle.rb
luisramos0: the error we are fixing is that we can't have enterprises matching coordinator_id, we need enterprises ids. it just fails in rails 4.
can we use user.enterprises_ids? instead of user.enterprises.map(&:id)?
we could use select or map, but this case is better with map because the user.enterprises is already lazy loaded, with select we would trigger another db query.

luisramos0: I thought about it, user.enterprises are lazy loaded already and map is better. but seeing @sauloperez comment above made me realize user.enterprises_ids would be better. is user.enterprises_ids valid?

app/models/order_cycle.rb
luisramos0: this would not be a subquery, order.variants are already loaded at this point. so map is better.

luisramos0:
@mkllnk "Did you know about the use of sub queries when using relations?"

yes, but I dont think your example is always valid. it will not always be a subquery. I think that in the app user.enterprises will already be loaded and map will avoid an extra query.

app/models/enterprise.rb
luisramos0: the * is not allowed in rails 4 and ready_for_checkout is only used for entities in EnterpriseInjectionData.active_distributors.
I have tested and the method works perfectly fine and the enterprises are loaded correctly.
 https://github.com/openfoodfoundation/openfoodnetwork/blob/a45782f59b9f355c6b7649554496739f3c29abe9/lib/open_food_network/enterprise_injection_data.rb#L3

so I think we are good here.

app/models/order_cycle.rb
luisramos0: so, if order.variants is already loaded, AR will not make it a subquery. and using select on an already loaded order.variants will produce an extra query. probably because of the intense chaining of scopes...

luisramos0:
@mkllnk @sauloperez can you guys pls have a second look. I am pretty confident with the use of map instead of select...

luisramos0:
rebased to resolve conflicts with #3842

luisramos0:
fixed rubocop issues, ready for review.

app/models/enterprise.rb
kristinalim: Changing the `SELECT` to just use `enterprises.id` is misleading IMO because the scope name does not clarify that only the ID would be included.

This only works because `Enterprise.distributors_with_active_order_cycles` still uses `select("DISTINCT enterprises.*")` (see [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/0f461067ab763ae766521b9f132a30008a37e446/app/models/enterprise.rb#L132)), so the resulting `SELECT` is `DISTINCT enterprises.*, enterprises.id`.

In Rails 4. there is [`ActiveRecord::Associations::CollectionProxy#distinct`](https://api.rubyonrails.org/v4.2.11.1/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-distinct): (`v4.2.11`)

```
> puts User.joins(:recipes).distinct.to_sql
SELECT DISTINCT "users".* FROM "users" INNER JOIN "recipes" ON "recipes"."user_id" = "users"."id"
```

If we can't change `distributors_with_active_order_cycles` to not to use `DISTINCT enterprises.*`, I think we should leave this particular line as is for now and just wait for the Rails 4 upgrade before switching to `distinct`.

app/models/order_cycle.rb
kristinalim: I checked, calling `*_ids` multiple times triggers a SQL query each time it is called while the association is not lazy loaded. But if the association is already lazy loaded, it uses the cache.

```
> user.recipe_ids ; nil
   (0.7ms)  SELECT "recipes".id FROM "recipes" WHERE "recipes"."user_id" = $1  [["user_id", 1]]
=> nil
> user.recipe_ids ; nil
   (0.7ms)  SELECT "recipes".id FROM "recipes" WHERE "recipes"."user_id" = $1  [["user_id", 1]]
=> nil
> user.recipes.to_a ; nil
  Recipe Load (0.9ms)  SELECT "recipes".* FROM "recipes" WHERE "recipes"."user_id" = $1  [["user_id", 1]]
=> nil
> user.recipe_ids 
=> [1]
```

kristinalim: I'm quite uncomfortable that the scope is now making assumptions about whether or not the `enterprises` association is already lazy loaded. But yeah, happy about the way you do it here if the associations are already lazy-loaded anyway.

app/models/enterprise.rb
luisramos0: thanks @kristinalim

"and just wait for the Rails 4 upgrade"
this "select distinct enterprises.*" doesnt work in rails 4. all the changes in this PR are fixing broken specs in the spree 2-1 build (running on rails 4). I mean, this is the rails 4 upgrade 🎉 

I am not sure why this "distinct enterprises.*" is breaking stuff and not the one in distributors_with_active_order_cycles (maybe I just missed other broken specs in the 2-1 build related to distributors_with_active_order_cycles).

I think you are suggesting I revert this change until we fix this and the one in distributors_with_active_order_cycles.

what if I change the name of this scope to be more appropriate ready_for_checkout_ids?

let me know, I can either revert this or rename the scope if you think it's acceptable (I prefer to keep this because this fixes lots of tests in the 2-1 build).

thanks!

app/models/order_cycle.rb
luisramos0: maybe user.enterprise_ids would be better than map?

kristinalim: Hmm... Yeah actually @luisramos0, sounds good to me.

app/models/enterprise.rb
kristinalim: Ah yes, nice. :slightly_smiling_face:

> what if I change the name of this scope to be more appropriate ready_for_checkout_ids

I did a quick check and I do not understand either. :grinning: It's not ideal, but okay - that sounds good to me.

luisramos0: ok, I think it was my misconception that select * was not supported in rails 4. There's nothing wrong with the select * in rails 4

(there is something very wrong with "select *" in my mind, "select *" is an anti pattern, it should always be avoided, the specific fields should always be listed, but I'll ignore this case for now).

So, I went back to v2.1. to verify and what is wrong in the not_ready_for_checkout scope:
https://github.com/openfoodfoundation/openfoodnetwork/blob/98d1cbec8c7a0f1b71ed16e2e527f250bd8fe8b6/app/models/enterprise.rb#L101
the problem is that in rails 4 we can't do "id IN (scope)" where the scope does select *, we need to select id.

So, I added a new commit to this PR to make ready_enterprises select the id field from the ready_for_checkout scope: [6091b1c319811c95e9adc4221562b21389111d30](https://github.com/openfoodfoundation/openfoodnetwork/pull/3709/commits/6091b1c319811c95e9adc4221562b21389111d30)
two points about the commit:
- I checked and the select statement does not trigger additional queries
- the enterprises.* is needed otherwise we get a "ambiguous column" (like in other changes in this PR)

Let me know if this makes sense for you :-)
Thanks!

kristinalim: The commit looks great. :smile:

sigmundpetersen:
Conflicts @luisramos0

luisramos0:
yeah, thanks Sigmund!
rebased, ready for testing.

RachL:
With product cache active, I've tested:

- adding and removing variants
- checkout
- creating / editing products and impact in shop front
- inventory

Everything looks good.

luisramos0:
🎉

PR #3734, opened 2019-04-11
Removing a variant from OC will prevent user with that variant in the cart from checking out

### LINKED ISSUE #2091: If a user has items in cart and hub remove them from OC, he can still proceed to checkout and buy them
## Description
A user can buy items that were removed/deleted from an order cycle if they had those products in their cart before retrieval.

## Expected Behavior
If at checkout time some products I have in my cart have become unavailable because removed from the OC by the manager, I should not be able to checkout and be sent back to my cart with an error message on the given item saying they became unavailable and prompting me to modify my cart.

## Actual Behavior
If at checkout time some products I have in my cart have become unavailable because retrieved from the OC by the manager, I am still able to buy them.

## Steps to Reproduce
1. In browser A, user A put in his cart x quantity of item y. He is ready to checkout but has not yet clicked on the last step of the checkout.
2. At that precise moment, the shop manager modify this very same order cycle and remove the given item OR producer from the order cycle, and save the change.
3. After that, user 1 click on the last step of his checkout and manage to proceed and buy the product.

## Animated Gif/Screenshot
https://www.useloom.com/share/4dcbfb6a0025435ea4014a0780921050

## Context
We reported while working on #1491 as part of the job for new feature #1031 persistent cart.

## Severity
s3 or s2 (to be discussed, putting s3 for now)

## Your Environment
- Version used: 1.10.1

## Possible Fix
An attempt was made on #1767 but was not complete.

luisramos0:
#### What? Why?

Closes #2091

This case was totally absent of the code so I had to implement it. I refactored the validator and I am happy with it but there's lots of things we can do to simplify the orders controller and checkout controllers, but that's only after #3670

#### What should we test?
See issue.
When you delete a variant in the cart from the order cycle, the user will see this page if she goes to /cart or /checkout.

Screenshot 2019-04-13 at 20 49 40

#### Release notes
Changelog Category: Fixed
If a Hub removes a variant from an Order Cycle and that variant is in a user's cart, the user will now be informed and will not be able to checkout until the variant is removed from the cart.


spec/requests/checkout/failed_checkout_spec.rb
```diff
@@ -23,6 +23,10 @@
   end
 
   before do
+    order_cycle_distributed_variants = double(:order_cycle_distributed_variants)
+    allow(OrderCycleDistributedVariants).to receive(:new).and_return(order_cycle_distributed_variants)
+    allow(order_cycle_distributed_variants).to receive(:distributes_order_variants?).and_return(true)
+
```
luisramos0: I spent over an hour trying to make these tests pass by adding the test variants to the order cycle but got really stuck with it (and the 14, yes 14, before_filters on the checkout controller do not help).
I decided to go for mocking... I dont think it's ideal to mock a specific class when testing the controller but it works.

luisramos0:
I wanted to crack the "view spec challenge" in this PR by creating our first view spec...
I gave it half an hour and I can't make it work only because of a method in the partial called mini_image which apparently is a dynamic method defined in spree base_helper... meta-programming...

I dont think I'll be able to fix this so I will leave this here for reference for the next view spec.

We need to activate haml specs in application.rb:

```ruby
diff --git a/config/application.rb b/config/application.rb
index b86316c96..a4cabe38f 100644
--- a/config/application.rb
+++ b/config/application.rb
@@ -170,5 +170,9 @@ module Openfoodnetwork
+
+    config.generators do |generator|
+      generator.template_engine :haml
+    end
```

and this is how the specs will look like
spec/views/spree/orders/line_item.html.haml_spec.rb
```ruby
require "spec_helper"

describe "spree/orders/_line_item.html.haml", type: :view do
  describe "when unavailable_order_variants is empty" do
    before do
      assign(:unavailable_order_variants, nil)
    end

    it "does not display currently unavailable" do
      line_item = create(:line_item)

      render partial: 'line_item', locals: { variant: line_item.variant,
                                             line_item: line_item,
                                             item_form: nil }

      expect(rendered).to_not contain("Currently unavailable")
    end
  end
end
```

spec/controllers/checkout_controller_spec.rb
```diff
@@ -36,22 +36,40 @@
...
-  it "renders when both distributor and order cycle is selected" do
-    controller.stub(:current_distributor).and_return(distributor)
-    controller.stub(:current_order_cycle).and_return(order_cycle)
-    controller.stub(:current_order).and_return(order)
-    order.stub_chain(:insufficient_stock_lines, :present?).and_return false
-    get :edit
-    response.should be_success
+    before do
+      controller.stub(:current_order).and_return(order)
+      order.stub(:distributor).and_return(distributor)
```
sauloperez: If we are using RSpecs's `allow` syntax I would use it here as well for the sake of consistency.

spec/services/order_cycle_distributed_variants_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe "checking if an order can change to a specified new distribution" do
+    it "returns false when a variant is not available for the specified distribution" do
+      order.should_receive(:line_item_variants) { [1] }
+      subject.should_receive(:available_variants) { [] }
+      expect(subject.distributes_order_variants?(order)).to be false
+    end
+
+    it "returns true when all variants are available for the specified distribution" do
+      order.should_receive(:line_item_variants) { [1] }
+      subject.should_receive(:available_variants) { [1] }
```
sauloperez: Again, better to stick to RSpec 3 syntax rather than both 2 and 3.

luisramos0: :+1:

spec/controllers/checkout_controller_spec.rb
luisramos0: thanks, I am still a bit confused as I learned all this looking at the mixed ofn spec code :-)

RachL:
@luisramos0 it is looking good! However I experienced a lot of 500 errors while trying to checkout products on OFN UK demo shop. It didn't happened on other shops... and not on all products. For example I could checkout "X" or "Bouillon powder". My guess is that it is unrelated to this PR, correct?

https://docs.google.com/document/d/1yiUCjbJJvdHYtcAgMGCg0hXqTWPiwWe7r_u8Cum5KP8/edit#

sauloperez:
We can give it a try in Katuma staging @luisramos0 to be sure

RachL:
Everything looks good on Katuma staging, this is ready to go!
image

PR #3736, opened 2019-04-12
Install and set up newrelic_rpm gem for development environment

kristinalim:
#### What? Why?

The 3.x version of `newrelic_rpm` gem offers a local dashboard (does not require registration on their website) for tracking and analyzing SQL queries that were done for a Rails instance.

Over `rack-mini-profiler`, it provides the following:

* Tracking for multiple requests - Can analyze requests not just for the current page
* Centralized page for reviewing
* :point_up: Which allows us to easily analyze data for AJAX requests
* Ability to drill down to exact line of code (with stacktrace) where the SQL queries were executed
* Nice summaries

The next major version of the gem removes this local dashboard, so this doesn't seem like a good idea to introduce now. There is actually version 6.x already. But the 3.x version works for the OFN `2-0-stable` branch, and might still be worth keeping until we have improved performance issues.

This PR only adds this gem for the `development` environment.

To go to that quite helpful dashboard in `development` environment, go to `/newrelic`, e.g. http://localhost:3000/newrelic

#### What should we test?

Just test that any page loads. The admin dashboard might be a nice sample page for this.

#### Release notes

- Install and set up `newrelic_rpm` gem for the development environment.

Changelog Category: Added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ NERD_tree*
 coverage
 libpeerconnection.log
 /config/application.yml
+/config/newrelic.yml
 node_modules
 vendor/bundle/
 coverage
--- a/Gemfile
+++ b/Gemfile
@@ -142,6 +142,7 @@ end
 
 group :development do
   gem 'byebug', '~> 9.0.0' # 9.1 requires ruby 2.2
+  gem "newrelic_rpm", "~> 3.0"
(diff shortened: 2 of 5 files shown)


RachL:
I've tested:
- homepage
- shops
- maps
- producers
- groups
- admin dashboard, reports, enterprises, customers, OC, products and orders

This is ready to go!

PR #3758, opened 2019-04-28
3292 link popup text translation

### LINKED ISSUE #3292: [i18n] translation missing "insert url"
## Description

On entreprise settings / shop preferences

public php

## Steps to Reproduce
1. Log in as enterprise user or superadmin
2. Go to enterprise > Settings
3. Go to shop preferences
4. Try to add an URL you will see that "insert url" is not translated

## Severity

jazzdragon:
Closes #3292 

Getting close to translating pop-up alert from textAngular, can't figure out how to make the dynamic parsing of the translation work. Looks like this is an existing issue: https://github.com/textAngular/textAngular/issues/821

Filling out the rest of the info below in case this is seen through to completion.

UPDATE (Luis): removed double quotes and used t function, it's working now. I think this is ready for review and test.

#### Release notes
Ability to translate text on pop-up input field when adding a link to shopfront messages in admin configs.

Changelog Category: Changed

### DIFF
--- a/app/assets/javascripts/admin/enterprises/enterprises.js.coffee
+++ b/app/assets/javascripts/admin/enterprises/enterprises.js.coffee
@@ -18,8 +18,8 @@ angular.module("admin.enterprises", [
       '$delegate'
       (taTranslations) ->
         taTranslations.insertLink = {
-          tooltip: 'Insert / edit link',
-          dialogPrompt: "Please enter a URL to insert"
+          tooltip: "{{'admin.enterprises.form.shop_preferences.shopfront_message_link_tooltip' | t}}",
+          dialogPrompt: "{{'admin.enterprises.form.shop_preferences.shopfront_message_link_prompt' | t}}"
         }
         taTranslations
     ]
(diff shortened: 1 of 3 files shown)


luisramos0:
Nice, thanks for your PR @jazzdragon 
I gave it a try. It's working now. See new commit. The trick I used was to get the translation done immediately on our side by removing the double quotes.

(for some reason the {{ | t}} pattern doesnt work. [We prefer it over using the t function in js script](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)#javascript-translations-prefer-angular-filter-over-global-t-function). But I think it's acceptable to use the t function in cases we cant make the filter work)

I also added double quotes on the en.yml file. I cant find this rule written but I think we agreed all translations should be inside double quotes to avoid mistakes.

I think this is ready for review now (I have updated the description)

jazzdragon:
Thanks @luisramos0, I'm pretty sure I tried that and `t` was not in scope there. Odd...
For the tooltip text, the interpolation string does work, but I guess it would be better to keep them both the same.

luisramos0:
yes, I'd keep both the same.

--- I'm pretty sure I tried that and t was not in scope there. Odd...

maybe you didnt remove the double quotes?

RachL:
Everything is looking fine, moving to ready to go

image

luisramos0:
@jazzdragon your first PR I believe 🎉  thank you!

PR #3762, opened 2019-04-30
Store S3 keys in Config before calling db2fog

sauloperez:
#### What? Why?

This ensures that Db2fog always picks up the latest value of the S3_* env vars and not the one that was persisted last time. Now you can do things like `S3_BUCKET=xxx bundle exec rake b2fog:backup` if you had to.

#### What should we test?
Nothing

#### Release notes

Make Db2fog pick up the latest values of the S3 env vars and not the ones previously persisted.

Changelog Category: Fixed


sauloperez:
Superseded by #3763

PR #3764, opened 2019-04-30
Don't use s3 for images unless images bucket has been provisioned.

### LINKED ISSUE #3766: Can't use S3 only for backups without using it for image storage too
## Description

When I provisioned Katuma with the latest `ofn-install` and v1.30.0 few errors related to Paperclip starting popping up such as

Screenshot from 2019-04-30 21-39-21

The actual problem here is that katuma does not use S3 as image storage, only as backup storage. 

I tried disabling `use amazon s3 for images` from `/admin/image_settings/edit` with no success. Still the following line

https://github.com/openfoodfoundation/openfoodnetwork/blob/88cbf5bd36ab5b633ffc3f76efb41822072146b9/config/initializers/spree.rb#L28

will always get evaluated taking precedence over the UI.

## Expected Behavior

Images should be stored in the server and not in S3. Specifically, `Spree::Image.attachment_definitions[:attachment][:storage]` should return `nil` from a rails console in production. See https://github.com/openfoodfoundation/spree/blob/step-6a/backend/app/controllers/spree/admin/image_settings_controller.rb#L55 for details.

## Actual Behaviour

Paperclip is configured with S3 storage thus all image URLs are resolved assuming so and fail due to missing configuration options.

## Steps to Reproduce

1. tail -f log/delayed_job
2. enqueue a `ProductsCacheIntegrityCheckerJob` with valid settings like distributor_id: 93,
order_cycle_id: 505
3. The log will show a message like 

```
Job ProductsCacheIntegrityCheckerJob (id=323401) FAILED (6 prior attempts) with ArgumentError: missing required :bucket option
```

## Context

We never enabled backups in Katuma until now.

## Severity

## Your Environment

* Version used: v1.30.0, ofn-install `master`
* Browser name and version:

## Possible Fix

#3764

Matt-Yorkley:
#### What? Why?

Closes #3766 

Sets the `use_s3` (for images) option based on presence of `s3_images_bucket` from ofn-install instead of presence of s3 keys, for instances that use s3 for backups but not for images.

#### What should we test?

Provisioning Katuma shouldn't activate the use of s3 for images.

@sauloperez

### DIFF
--- a/config/initializers/spree.rb
+++ b/config/initializers/spree.rb
@@ -25,7 +25,7 @@ Spree.config do |config|
   config.s3_bucket = ENV['S3_BUCKET'] if ENV['S3_BUCKET']
   config.s3_access_key = ENV['S3_ACCESS_KEY'] if ENV['S3_ACCESS_KEY']
   config.s3_secret = ENV['S3_SECRET'] if ENV['S3_SECRET']
-  config.use_s3 = true if ENV['S3_ACCESS_KEY'] && ENV['S3_SECRET']
+  config.use_s3 = true if ENV['S3_BUCKET']
   config.s3_protocol = ENV.fetch('S3_PROTOCOL', 'https')
 end
 


sauloperez:
Cherry-picked the change into `v1.30.0-plus-use-s3-patch` to deploy to production and working :tada: Thanks!

luisramos0:
already tested in katuma, so ready to go.

PR #3768, opened 2019-05-01
[Spree Upgrade] Migrate spree_variants.on_demand to spree_stock_items.backorderable

### LINKED ISSUE #3767: [Spree Upgrade] Broken stock levels migration - all variants become on_demand after migration
## Description
After migrating a v1 db to v2, all variants become on_demand and no data is left available to fix it as the variants.on_demand field is deleted.

The data about on_demand is just being removed [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/2-0-stable/db/migrate/20180426145635_remove_on_demand_from_product_and_variant.spree.rb).

## Expected Behavior
After db migration, stock levels should be correct, not all variants are on_demand.

## Actual Behaviour
After db migration, all variants are on_demand.

## Steps to Reproduce
Load live data into a staging server and migrate the data to v2. Verify that all variants are on_demand.

## Severity
We cant go live without fixing this.

## Your Environment
I tested this in staging katuma with live katuma data.

## Possible Fix
The count_on_hand value is migrated [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/2-0-stable/db/migrate/20180426145632_create_default_stock.spree.rb).

When count_on_hand is migrated, stock_item.backorderable field is not present yet. It's only created [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/2-0-stable/db/migrate/20180426145644_add_backorderable_to_stock_item.spree.rb).

The easiest fix is probably to not remove spree_variants.on_demand so early on and just create a new migration that copies spree_variants.on_demand to stock_item.backordearble and then drops spree_variants.on_demand.

luisramos0:
#### What? Why?

Closes #3767
Copies data from spree_variants.on_demand to spree_stock_items.backorderable and then drops spree_variants.on_demand.

#### What should we test?
See issue #3767

### DIFF
--- a/db/migrate/20180426145635_remove_on_demand_from_product_and_variant.spree.rb
+++ b/db/migrate/20180426145635_remove_on_demand_from_product_and_variant.spree.rb
@@ -2,6 +2,6 @@
 class RemoveOnDemandFromProductAndVariant < ActiveRecord::Migration
   def change
     remove_column :spree_products, :on_demand
-    remove_column :spree_variants, :on_demand
+    # we are removing spree_variants.on_demand in a later migration
   end
 end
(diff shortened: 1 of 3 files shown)


RachL:
Afer speaking with Pau, this is ready to go!

PR #3788, opened 2019-05-04
Add display_name and display_as to admin edit variant page

### LINKED ISSUE #3249: Display_name not available on the variants create/edit page
## Description
As a farmer
When I create a variant for my product
I need to set the display name of the variant as it will be shown on the shopfront
Currently I can only do that on the bulk product edit page

This is very confusing for the user who is creating variants using the variants tab...

Additionally, and to add to the user confusion, the unit display_as value is also only available on the bulk product edit page.

## Expected Behavior
I can set the variant's display_name and unit's display_as fields of a variant when I edit a variant.

## Actual Behavior
These fields are missing from the page (see third screenshot):
screenshot 2018-12-27 at 14 59 34

screenshot 2018-12-27 at 14 59 49

screenshot 2018-12-27 at 14 59 58

## Steps to Reproduce
Go to bulk product edit page and click on the edit button of any variant

## Context
I found this in a user demo in Portugal and then I realised it can be easily fixed as part of the spree upgrade.

## Severity
I'd say bug-s3, it's totally broken UX on a very basic feature: I believe every single user will see this.

## Possible Fix
The missing fields are OFN specific fields. We need to de-deface the view and add these fields to it.
This is discussed in this thread: https://community.openfoodnetwork.org/t/product-option-types-vs-product-variant-unit-fields/1535/3
This is related to spree upgrade issue: #3091 and I believe we should fix this one together with it.

stveep:
#### What? Why?

Closes #3249

Adds "Display name" and "Display As" fields to Variant edit form.

#### What should we test?
Test that fields display and update the correct variant attributes

#### Release notes
Display name can now be set while creating/editing an individual variant

Changelog Category: Fixed 

#### Discourse thread

https://community.openfoodnetwork.org/t/product-option-types-vs-product-variant-unit-fields/1535/3

### DIFF
--- a/spec/features/admin/variants_spec.rb
+++ b/spec/features/admin/variants_spec.rb
@@ -118,4 +118,29 @@ feature %q{
     v.reload
     v.deleted_at.should_not be_nil
   end
+
+  scenario "editing display name for a variant", js:true do
+    p = create(:simple_product)
+    v = p.variants.first
+
+    # When I view the variant
+    login_to_admin_section
+    visit spree.admin_product_variants_path p
+    page.find('table.index .icon-edit').click
+
+    # It should allow the display name to be changed
+    page.should have_field "variant_display_name"
+    page.should have_field "variant_display_as"
+
+    # When I update the fields and save the variant
+    fill_in "variant_display_name", with: "Display Name"
+    fill_in "variant_display_as", with: "Display As This"
+    click_button 'Update'
+    page.should have_content %Q(Variant "#{p.name}" has been successfully updated!)
+
+    # Then the displayed values should have been saved
+    v.reload
+    v.display_name.should == "Display Name"
+    v.display_as.should == "Display As This"
+  end
 end
--- a/app/overrides/spree/admin/variants/_form/add_display_name.html.haml.deface
+++ b/app/overrides/spree/admin/variants/_form/add_display_name.html.haml.deface
@@ -1,8 +1,0 @@
-/ insert_top "[data-hook='variants']"
-
-.field
-  = f.label :display_name, t(:display_name)
-  = f.text_field :display_name, class: "fullwidth"
-.field
-  = f.label :display_as, t(:display_as)
-  = f.text_field :display_as, class: "fullwidth"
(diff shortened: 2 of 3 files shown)


stveep:
Here is the screenshot - just need to fix the test.

image

sauloperez:
I'll move to in test as soon as @stveep fixes that failing spec.

stveep:
Fixed!

CathyPearce:
The display_name and display_as fields are not appearing top left, as shown on the screen shot earlier
Discussed with Steve, who will look to change the display order
image

luisramos0:
yeah, it's related to the override add_unit_value_and_description.html.haml.deface that only appears
if @product.variant_unit != 'items' ;-)

luisramos0:
actually, imo we could move this PR forward as is. displayname and displayas are showing up after unit value and unit description: the trick is that these two fields will only show if product unit != items.

stveep:
I reordered the fields. One comment here is that I am not really clear what the difference between "Display Name" and "Display As" is - but hopefully the users of this page are, otherwise some more useful labels might be in order! (Labels can be changed in the translation files)

spec/features/admin/variants_spec.rb
```diff
@@ -118,4 +118,29 @@
...
+
+    # When I view the variant
+    login_to_admin_section
+    visit spree.admin_product_variants_path p
+    page.find('table.index .icon-edit').click
+
+    # It should allow the display name to be changed
+    page.should have_field "variant_display_name"
+    page.should have_field "variant_display_as"
+
```
kristinalim: Great that you added a test! :slightly_smiling_face:

It's okay to leave this as is, but for future pull requests, please use the newer `expect` syntax. https://github.com/rubocop-hq/rspec-style-guide#use-expect

luisramos0:
@stveep yes, maybe we can change the labels later, these are the fields that show up on the shopfront for each variant
Screenshot 2019-05-04 at 22 38 40
display_name on the left and display_as after the quantity field.

CathyPearce:
Checked the fields display as expected on the Variant screen - looks good
https://docs.google.com/document/d/16zu6kAgYtnjJn7Anh621kRiRJ0yGsrREnnT8kRVUgfU/

Issue #3793 has been raised to put placeholder details for both fields to further improve this screen.

luisramos0:
🎉

PR #3794, opened 2019-05-05
& replaced by OR when multiple property filters are selected

### LINKED ISSUE #985: When shopfront displays multiple property filters please replace '&' with 'OR'

m-budryte:
#### What? Why?

Closes #985 

#### What should we test?
Product property filters (when choosing more than 1) - should not have & and should have OR instead

Changelog Category: Changed

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
@@ -45,12 +45,12 @@ Darkswarm.controller "ProductsCtrl", ($scope, $filter, $rootScope, Products, Ord
   $scope.appliedTaxonsList = ->
     $scope.activeTaxons.map( (taxon_id) ->
       Taxons.taxons_by_id[taxon_id].name
-    ).join(t(' products_or ')) if $scope.activeTaxons?
+    ).join(" #{t('products_or')} ") if $scope.activeTaxons?
 
   $scope.appliedPropertiesList = ->
     $scope.activeProperties.map( (property_id) ->
       Properties.properties_by_id[property_id].name
-    ).join(t(' products_or ')) if $scope.activeProperties?
+    ).join(" #{t('products_or')} ") if $scope.activeProperties?
 
   $scope.clearAll = ->
     $scope.query = ""
--- a/app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
@@ -50,7 +50,7 @@ Darkswarm.controller "ProductsCtrl", ($scope, $filter, $rootScope, Products, Ord
   $scope.appliedPropertiesList = ->
     $scope.activeProperties.map( (property_id) ->
       Properties.properties_by_id[property_id].name
-    ).join(" & ") if $scope.activeProperties?
+    ).join(" OR ") if $scope.activeProperties?
 
   $scope.clearAll = ->
     $scope.query = ""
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
```diff
@@ -50,7 +50,7 @@ Darkswarm.controller "ProductsCtrl", ($scope, $filter, $rootScope, Products, Ord
   $scope.appliedPropertiesList = ->
     $scope.activeProperties.map( (property_id) ->
       Properties.properties_by_id[property_id].name
-    ).join(" & ") if $scope.activeProperties?
+    ).join(" OR ") if $scope.activeProperties?
```
sauloperez: We would also need to make it possible to translate this string to all supported languages. https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29 has all the information you may need.

Basically, we need to add a new key in `config/locales/en.yml`, the default language, and then refer to it in here. This way we'll use the appropriate copy for each instance.

m-budryte: done! (checked in the browser - working fine for me)

app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
```diff
@@ -45,12 +45,12 @@ Darkswarm.controller "ProductsCtrl", ($scope, $filter, $rootScope, Products, Ord
   $scope.appliedTaxonsList = ->
     $scope.activeTaxons.map( (taxon_id) ->
       Taxons.taxons_by_id[taxon_id].name
-    ).join(" & ") if $scope.activeTaxons?
+    ).join(" #{t('products_or')} ") if $scope.activeTaxons?
```
Matt-Yorkley: Hey @m-budryte have you tested this in the browser? I'm not sure the `#{...}` syntax will work in coffeescript, and I think we need to concatenate two spaces around the translation, like:
```js
).join(" " + t('products_or') + " ") if $scope.activeTaxons?
```

config/locales/en_BE.yml
```diff
@@ -1522,6 +1522,7 @@ en_BE:
   orders_changeable_orders_alert_html: This order has been confirmed, but you can make changes until <strong>%{oc_close}</strong>.
   products_clear_all: Clear all
   products_showing: "Showing:"
+  products_or: "OR"
```
Matt-Yorkley: Also because of the way we use Transifex, we only update translation keys in `en.yml`, and don't commit any changes to the other locale files as they are updated separately.

app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
Matt-Yorkley: I apologise! It looks like you can use the Ruby interpolation style here. It seems like my `coffeescript` is a bit rusty!

config/locales/en_BE.yml
Matt-Yorkley: Could you revert the changes in `en_BE.yml`, `en_CA.yml`, `en_US.yml`, and `en_GB.yml`? Thanks! :heart:

luisramos0:
moving to in dev, I think you just need to remove the translations from files other than en.yml

luisramos0:
I removed the changes from the files expect en.yml
this is ready for review now.

kristinalim:
Moved this PR to Test Ready! :tada:

RachL:
This ready to go :)

image

PR #3797, opened 2019-05-05
Remove non-functional Price field from Edit Product screen

### LINKED ISSUE #2979: Product master variant price
## Description

When a new product is created the price gets set on the master variant. If a product is edited later, there is a field for price on the edit product screen, but it does not update the prices in the frontend, as they are taken from the price on the variants, not the master price on the product.

## Expected Behavior

Fields in product edit page should work

## Actual Behavior

The price field does not work as expected

## Steps to Reproduce

1. Create a product with price: 5
2. Go to edit product and change the price field in the product details section of edit product.
3. The price in the frontend stays the same
4. Edit the price on the variant, and it will change correctly

## Screenshot

screenshot from 2018-11-06 13-04-02

## Context

Discovered by Luis while looking at Spree Upgrade issues.

## Severity
S4. The UI is confusing to the user and appears to be broken.

## Possible Fix

Remove that price field from the product edit page, as it is non-functional.

lin-d-hop:
#### What? Why?

Closes #2979 

The Price field is non functional and catches many users out when they 'update' a price but do not see the changes flow through to the live product in their shopfront.

Instead users must update variant prices.

#### What should we test?

Update:
We need to test:
Creating a new product and variant with and without price
Updating a product and variant price
Updating a product and variant but not updating the price.

Screenshot from 2019-05-05 14-41-13

#### Release notes

Enterprise managers can no longer access the non-functional product price field.

Changelog Category: Fixed 

none

#### Discourse thread

none

#### Dependencies

none

#### Documentation updates

none. Checked the user guide but found no reference.

### DIFF
--- a/app/controllers/spree/admin/products_controller_decorator.rb
+++ b/app/controllers/spree/admin/products_controller_decorator.rb
@@ -151,4 +151,8 @@ Spree::Admin::ProductsController.class_eval do
     variant.on_demand = on_demand if on_demand.present?
     variant.on_hand = on_hand.to_i if on_hand.present?
   end
+
+  def set_product_master_variant_price_to_zero
+    @product.price = 0 if @product.price.nil?
+  end
 end
--- a/app/views/spree/admin/products/_form.html.haml
+++ b/app/views/spree/admin/products/_form.html.haml
@@ -46,11 +46,6 @@
       = f.collection_select(:supplier_id, @producers, :id, :name, {:include_blank => true}, {:class => "select2"})
       = f.error_message_on :supplier
 
-    = f.field_container :price do
-      = f.label :price, raw(t(:price) + content_tag(:span, ' *', :class => 'required'))
-      = f.text_field :price, :value => number_to_currency(@product.price, :unit => '')
-      = f.error_message_on :price
-
     .clear
 
     - unless @product.has_variants?


lin-d-hop:
Ok pause on review of this. It needs more work....

lin-d-hop:
Ok this is ready to go. @luisramos0 Fancy a review?

config/initializers/spree.rb
```diff
@@ -39,6 +39,9 @@
   config.s3_secret = ENV['S3_SECRET'] if ENV['S3_SECRET']
   config.use_s3 = true if ENV['S3_BUCKET']
   config.s3_protocol = ENV.fetch('S3_PROTOCOL', 'https')
+
+  # Master product price field overridden by variant price thus not required.
+  config.require_master_price = false
```
luisramos0: there's no such thing as product.price. product.price is a delegate to product's  master variants price.

I'd change this explanation to: "we will create products with variants (master and standard) with no price and we will handle the consequences of that in the application"

if we let products be created without price, the products master variant will have no price and also the standard variant will have no price.

we need to test the implications of this. I think a few things will break on the shopfront if you create a product (without updating the variant price) and put it on sale...

**otherwise**, I think it is probably easier and safer to revert this line and instead just force the product price to zero on the product controller. what do you think?

sauloperez: Agree on forcing the price to 0 but I guess the variant's `price` db column has a default value defined already. 

This change really needs to be thoroughly checked.

sauloperez:
Agree with Luis on forcing the price to 0 but I guess the variant's `price` db column has a default value defined already. 

This change really needs to be thoroughly checked.

lin-d-hop:
@luisramos0 'owisit? (Northern English dialect for 'How is this?')

app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -151,4 +152,9 @@ def set_stock_levels(product, on_hand, on_demand)
     variant.on_demand = on_demand if on_demand.present?
     variant.on_hand = on_hand.to_i if on_hand.present?
   end
+
+  def set_product_master_variant_price_to_zero
+    return unless @product.price.nil?
+    @product.price = 0
```
luisramos0:
```suggestion
    @product.price = 0 if @product.price.nil?
```

lin-d-hop:
Thanks @luisramos0. Syntax updated along with notes of what to test in the main description.

lin-d-hop:
Can I move this to test ready? @luisramos0 @sauloperez

luisramos0:
:+1:

RachL:
Today in production, the price field is mandatory when using the "new product" button.

This has for some reason be removed here. If it is a technical reason, I suggest we remove the * in the field (will open a new issue for this).

If it is due to something else, shall we consider to fix this is this PR @lin-d-hop ?

The rest is working good! I tested in super admin and manager, and display in shopfront and checkout.

Notes:
https://docs.google.com/document/d/1DxjoHGkILciJHsSjvy5trZEDv2h4gs2wkFmwiaJDsWQ/edit#

luisramos0:
@RachL I think the problem with price field mandatory in new product page is not specific to this PR.

when you say "production" I assume you mean v1 like openfoodfrance.org, right?
I tried this in stg AUS with v2 but without this PR, and I can create a product leaving the price field empty. it makes the price be zero.
so, the red start (mandatory) in the price field in create product page is a bug and can be removed, but it's not introduced by this PR.

It's a simple but because this PR is ready to go if you agree, I think we should create a separate issue and PR for that one. It will be just to remove the red star.

RachL:
@luisramos0 oops yes sorry I forgot that it could have been introduce by spree v2. let's move this one then.

lin-d-hop:
:tada:

PR #3798, opened 2019-05-05
the translations for Remove in the exchange forms added

### LINKED ISSUE #3732: Missing translation when trying to remove a fee from an OC
## Description
When you add a fee to an OC, the "remove" link is not translated.

## Steps to Reproduce
1. Log in as a producer or a hub or a super admin
2. Go to /admin/order_cycles
3. Create on order cycle and add fees to it
4. You will the the "remove" link

## Animated Gif/Screenshot

image

## Severity

* Version used: v1.29
* Operating System and version (desktop or mobile): Desktop Ubuntu 18.10

m-budryte:
#### What? Why?

Closes #3732

Localised versions do not have 'Remove' translated in the exchanged forms

#### What should we test?
/admin/order_cycles/1/edit - add fee - make sure that 'Remove' is shown in the correct language

#### Release notes
I copied translations for 'Remove' in each localisation file and created corresponding new order_cycles.exchange_form.remove keys. Could not find it in pt-BR :( 

Changelog Category:  Fixed

### DIFF
--- a/app/views/admin/order_cycles/_exchange_form.html.haml
+++ b/app/views/admin/order_cycles/_exchange_form.html.haml
@@ -26,7 +26,7 @@
 
         = select_tag 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_enterprise_fee_id', nil, {'id' => 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_enterprise_fee_id', 'ng-model' => 'enterprise_fee.id', 'ng-options' => 'enterprise_fee.id as enterprise_fee.name for enterprise_fee in enterpriseFeesForEnterprise(enterprise_fee.enterprise_id)'}
 
-        = link_to t('remove'), '#', {'id' => 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_remove', 'ng-click' => 'removeExchangeFee($event, exchange, $index)'}
+        = link_to t('.remove'), '#', {'id' => 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_remove', 'ng-click' => 'removeExchangeFee($event, exchange, $index)'}
 
     = f.submit t('.add_fee'), 'ng-click' => 'addExchangeFee($event, exchange)', 'ng-hide' => '!enterprises[exchange.enterprise_id].managed && !order_cycle.viewing_as_coordinator'
   %td.actions
--- a/app/views/admin/order_cycles/_exchange_form.html.haml
+++ b/app/views/admin/order_cycles/_exchange_form.html.haml
@@ -26,7 +26,7 @@
 
         = select_tag 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_enterprise_fee_id', nil, {'id' => 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_enterprise_fee_id', 'ng-model' => 'enterprise_fee.id', 'ng-options' => 'enterprise_fee.id as enterprise_fee.name for enterprise_fee in enterpriseFeesForEnterprise(enterprise_fee.enterprise_id)'}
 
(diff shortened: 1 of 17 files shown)


app/views/admin/order_cycles/_exchange_form.html.haml
```diff
@@ -26,7 +26,7 @@
 
         = select_tag 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_enterprise_fee_id', nil, {'id' => 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_enterprise_fee_id', 'ng-model' => 'enterprise_fee.id', 'ng-options' => 'enterprise_fee.id as enterprise_fee.name for enterprise_fee in enterpriseFeesForEnterprise(enterprise_fee.enterprise_id)'}
 
-        = link_to 'Remove', '#', {'id' => 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_remove', 'ng-click' => 'removeExchangeFee($event, exchange, $index)'}
+        = link_to t('remove'), '#', {'id' => 'order_cycle_{{ exchangeDirection(exchange) }}_exchange_{{ $parent.$index }}_enterprise_fees_{{ $index }}_remove', 'ng-click' => 'removeExchangeFee($event, exchange, $index)'}
```
luisramos0: should be t('.remove') to pickup the lazy lookup, otherwise it will lookup in the root of the translations keys.

config/locales/ca.yml
```diff
@@ -876,6 +876,7 @@ ca:
         pickup_time_placeholder: "Preparat per (és a dir, data / hora)"
         receival_instructions_placeholder: "Instruccions de recepció"
         add_fee: 'Afegeix una comissió'
+        remove: Eliminar
```
luisramos0: we only update en.yml file
all the other files must be left as they are because they will be automatically updated through the transifex process

app/views/admin/order_cycles/_exchange_form.html.haml
sauloperez: I think @m-budryte 's idea was to use the global "remove" key present in the root of the file. Not sure which is better though.

config/locales/ca.yml
sauloperez: Yes. Remember all this is explained in https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29

app/views/admin/order_cycles/_exchange_form.html.haml
m-budryte: I've done it .remove first but order_cycle.exchanges.remove was not defined. Then I've added the keys inside exchange in every language file by copying the line from the global 'remove'. Then I was told not to change the localisation files as it's done through a different system. Current solution fixes the lack of translation straightaway and does not require extra work from translators but not sure if it's good for the future. :)

config/locales/ca.yml
m-budryte: argh that's my git revert not reverting properly hold on

m-budryte: have pushed again, pls check when you have time :)

app/views/admin/order_cycles/_exchange_form.html.haml
luisramos0: I am sorry but this PR, as is, is not correct.
you can either:
1. use the root remote t('remove') already [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/b443f5ece9490539c6e1b55cdefe2291317b02ea/config/locales/en.yml#L208), no change to en.yml
2. use relative and lazy look up t('.remove') and add the new relative key in en.yml as you are going in this PR

I approve both but I prefer option 2.

config/locales/en.yml
```diff
@@ -867,6 +867,7 @@ en:
         pickup_time_placeholder: "Ready for (ie. Date / Time)"
         receival_instructions_placeholder: "Receival instructions"
         add_fee: 'Add fee'
+        remove: 'Remove'
```
luisramos0: you only need this in option 2

luisramos0:
hello @m-budryte do you think you will pick up this PR again?
let us know, thanks!

m-budryte:
@luisramos0 Hi, unfortunately not for a week or so. Feel free to assign it to someone else or I will be back soon!

app/views/admin/order_cycles/_exchange_form.html.haml
luisramos0: ok, I added one commit, we go with option 2.

luisramos0:
ok, back to code review.

RachL:
Looking good!

Moving to ready to go

image

PR #3801, opened 2019-05-06
Fix /maps performance

### LINKED ISSUE #3756: Why /map is so slow?
## Description
I'm filing this using the bug template, not sure if you consider performance stuff as a bug or something else. Sorry if it's the wrong place and sorry if there is any other issue/pr addressing this.

Basically I'm wondering why `/map` is so slow to load:
https://openfoodnetwork.org.uk/map 52294 ms
https://app.katuma.org/map 19677 ms

Any insight?

## Expected Behavior
Be faster? :grimacing: :sweat_smile: 

## Actual Behaviour
Sloooow.

## Steps to Reproduce
Visit `/map` on any OFN instance.

## Context
:grimacing: 

## Severity
I'm not sure, feel free to change it.

## Your Environment

* Version used:
* Operating System and version (desktop or mobile): Arch Linux desktop
* OFN Platform instance where you discovered the bug, and which version of the software they are using: I tried on Katuma and UK

## Possible Fix
:sunglasses:

Matt-Yorkley:
#### What? Why?

Closes #3756

I knocked this up at the UK Devhack yesterday.

Speeds up loading of the `/map` page from ~44 seconds to ~1 second. If I have a couple of hours (and the green light) I can apply the same method to `/shops` and to the `/groups/*` pages, which are currently the bane of UK uptime.

#### What should we test?

- Map works, correct information is displayed in modals when clicking markers.
- The map tab in groups pages should also be checked.

#### Release notes

Improved performance on `/maps` page.

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/darkswarm/services/enterprise_modal.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/enterprise_modal.js.coffee
@@ -1,9 +1,12 @@
-Darkswarm.factory "EnterpriseModal", ($modal, $rootScope)->
+Darkswarm.factory "EnterpriseModal", ($modal, $rootScope, $http)->
   # Build a modal popup for an enterprise.
   new class EnterpriseModal
     open: (enterprise)->
       scope = $rootScope.$new(true) # Spawn an isolate to contain the enterprise
       scope.embedded_layout = window.location.search.indexOf("embedded_shopfront=true") != -1
 
-      scope.enterprise = enterprise
-      $modal.open(templateUrl: "enterprise_modal.html", scope: scope)
+      $http.get("/api/enterprises/" + enterprise.id + "/shopfront").success (data) ->
+        scope.enterprise = data
+        $modal.open(templateUrl: "enterprise_modal.html", scope: scope)
+      .error (data) ->
+        console.error(data)
--- a/app/helpers/injection_helper.rb
+++ b/app/helpers/injection_helper.rb
@@ -10,6 +10,14 @@ module InjectionHelper
     )
   end
 
+  def inject_enterprise_shopfront_list
+    inject_json_ams(
+      'enterprises',
+      Enterprise.activated.includes(address: :state).all,
+      Api::EnterpriseShopfrontListSerializer
+    )
+  end
+
   def inject_enterprise_and_relatives
     inject_json_ams "enterprises", current_distributor.relatives_including_self.activated.includes(address: :state).all, Api::EnterpriseSerializer, enterprise_injection_data
   end
--- a/app/controllers/api/enterprises_controller.rb
+++ b/app/controllers/api/enterprises_controller.rb
@@ -47,6 +47,12 @@ module Api
       end
     end
 
+    def shopfront
+      enterprise = Enterprise.find_by_id(params[:id])
+
+      render text: Api::EnterpriseShopfrontSerializer.new(enterprise).to_json, status: :ok
+    end
+
     private
 
     def override_owner
--- a/app/controllers/api/enterprises_controller.rb
+++ b/app/controllers/api/enterprises_controller.rb
@@ -1,5 +1,5 @@
 module Api
-  class EnterprisesController < Spree::Api::BaseController
+  class EnterprisesController < BaseController
 
     before_filter :override_owner, only: [:create, :update]
     before_filter :check_type, only: :update
--- a/app/serializers/api/enterprise_shopfront_list_serializer.rb
+++ b/app/serializers/api/enterprise_shopfront_list_serializer.rb
@@ -1,0 +1,52 @@
+# Represents the minimum details of an Enterprise when all shopfronts are being listed
+module Api
+  class EnterpriseShopfrontListSerializer < ActiveModel::Serializer
+    attributes :name, :id, :latitude, :longitude, :is_primary_producer, :is_distributor,
+               :visible, :path, :icon, :icon_font, :producer_icon_font
+
+    has_one :address, serializer: Api::AddressSerializer
+
+    def path
+      enterprise_shop_path(enterprise)
+    end
+
+    def icon
+      icons = {
+        hub: "/assets/map_005-hub.svg",
+        hub_profile: "/assets/map_006-hub-profile.svg",
+        producer_hub: "/assets/map_005-hub.svg",
+        producer_shop: "/assets/map_003-producer-shop.svg",
+        producer: "/assets/map_001-producer-only.svg",
+      }
+      icons[enterprise.category]
+    end
+
+    def icon_font
+      icon_fonts = {
+        hub: "ofn-i_063-hub",
+        hub_profile: "ofn-i_064-hub-reversed",
+        producer_hub: "ofn-i_063-hub",
+        producer_shop: "ofn-i_059-producer",
+        producer: "ofn-i_059-producer",
+      }
+      icon_fonts[enterprise.category]
+    end
+
+    def producer_icon_font
+      icon_fonts = {
+        hub: "",
+        hub_profile: "",
+        producer_hub: "ofn-i_059-producer",
+        producer_shop: "ofn-i_059-producer",
+        producer: "ofn-i_059-producer",
+      }
+      icon_fonts[enterprise.category]
+    end
+
+    private
+
+    def enterprise
+      object
+    end
+  end
+end
--- a/app/serializers/api/enterprise_shopfront_serializer.rb
+++ b/app/serializers/api/enterprise_shopfront_serializer.rb
@@ -1,0 +1,133 @@
+# Represents the properties of an Enterprise when viewing the details of listed shopfronts
+module Api
+  class EnterpriseShopfrontSerializer < ActiveModel::Serializer
+    include SerializerHelper
+
+    attributes :name, :id, :description, :latitude, :longitude, :long_description, :website,
+               :instagram, :linkedin, :twitter, :facebook, :is_primary_producer, :is_distributor,
+               :phone, :visible, :email_address, :hash, :logo, :promo_image, :path, :category,
+               :active, :producers, :orders_close_at, :hubs, :taxons, :supplied_taxons, :pickup,
+               :delivery
+
+    has_one :address, serializer: Api::AddressSerializer
+    has_many :supplied_properties, serializer: Api::PropertySerializer
+    has_many :distributed_properties, serializer: Api::PropertySerializer
+
+    def orders_close_at
+      OrderCycle.with_distributor(enterprise).soonest_closing.first.andand.orders_close_at
+    end
+
+    def active
+      enterprise.ready_for_checkout? && OrderCycle.active.with_distributor(enterprise).exists?
+    end
+
+    def pickup
+      shipping_types? :pickup
+    end
+
+    def delivery
+      shipping_types? :delivery
+    end
+
+    def email_address
+      enterprise.email_address.to_s.reverse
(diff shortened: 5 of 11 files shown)


mkllnk:
Woohoo! This is great. I'm very happy about the approach and it gives hope that we can solve other performance problems quickly.

app/assets/javascripts/darkswarm/services/enterprise_modal.js.coffee
```diff
@@ -1,9 +1,12 @@
...
       scope = $rootScope.$new(true) # Spawn an isolate to contain the enterprise
       scope.embedded_layout = window.location.search.indexOf("embedded_shopfront=true") != -1
 
-      scope.enterprise = enterprise
-      $modal.open(templateUrl: "enterprise_modal.html", scope: scope)
+      $http.get("/api/shops/" + enterprise.id).success (data) ->
+        scope.enterprise = data
+        $modal.open(templateUrl: "enterprise_modal.html", scope: scope)
+      .error (data) ->
+        console.log(data)
```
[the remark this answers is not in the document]
Matt-Yorkley: Agreed. This was mostly for debugging at the time.

app/controllers/api/shops_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Api
+  class ShopsController < Spree::Api::BaseController
+    respond_to :json
+
+    def show
+      enterprise_id = Enterprise.find(params[:id])
```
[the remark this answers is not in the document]
Matt-Yorkley: I wasn't sure at the time, we have a lot of things in the enterprise namespace, and these serializers and endpoints are for returning specific representations of an enterprise in it relation to customers.

Matt-Yorkley:
I'll try to doc this a bit here. Here's some annotated pics of some of the things we are doing to display information in the `map`, `shops`, and `groups` pages as a rough example. The notes are a good guide, but it's a lot worse than that. There are _nested_ N+1 queries all over the place,and these are a small subset of them:

`SHOPS:`
shops
`MAP:`
maps
`GROUPS:`
groups

These 3 pages have been doing this to UK production: :warning:

Screenshot from 2019-04-29 17-08-00
Screenshot from 2019-04-29 17-07-43

Currently we query and pre-load _vast_ amounts of related data for every single enterprise before loading these pages. My approach was to create a very slim serializer (**ShopsSerializer**) that pulls only the bare minimum of data for each enterprise to display the map pointers (or list shop names), then when the user clicks on the pointer we make a new request which pulls the data we need to display all the info for that single enterprise, and load it in the modal (this uses the **ShopSerializer**, which has much more data, but only for one shop). The shops also display related enterprises in the UI, like the producers of a hub, or the hubs that a producer sells to. This last bit of nested data on related enterprises is done with **ShopRelativesSerializer**, which again only queries what it needs and no more.

I hesitated to use the "Enterprise" namespace because it was already very full, and this seemed like something quite specific. Whatever name we go with, we can re-use these serializers and endpoint in `/shops` and `/groups`, as the data they fetch is exactly the same as `/maps`

app/controllers/api/shops_controller.rb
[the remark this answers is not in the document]
luisramos0: I think @sauloperez means the variable enterprise_id, not the controller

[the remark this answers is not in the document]
luisramos0: I think that if we add Shop to our domain vocabulary, we need to stick to it.
Shop will be a 1-to-1 relationship with Enterprise. The name is ok, it's pretty obvious it's a representation of an Enterprise for the FrontOffice.

But I think I'd prefer to see this in api/enterprises_controller and have a separate name for this representation. I am not sure what would be a good name for example enterprise_light_serializer or enterprise_serializer_for_shopfront...

app/helpers/injection_helper.rb
```diff
@@ -10,6 +10,14 @@ def inject_enterprises(enterprises = Enterprise.activated.includes(address: :sta
     )
   end
 
+  def inject_map_pointers()
+    inject_json_ams(
+        'enterprises',
+        Enterprise.activated.includes(address: :state).all,
+        Api::ShopsSerializer
+    )
+  end
```
luisramos0: we can keep the syntax as in other methods, right?

```suggestion
def inject_map_pointers
   inject_json_ams "enterprises", Enterprise.activated.includes(address: :state).all, Api::ShopsSerializer
end
```

app/helpers/injection_helper.rb
```diff
@@ -10,6 +10,14 @@ def inject_enterprises(enterprises = Enterprise.activated.includes(address: :sta
     )
   end
 
+  def inject_map_pointers()
```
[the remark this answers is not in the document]
luisramos0: the name of a method should not have reference to it's usage but to what it does. method names should look within only, so, no reference to maps and pointers should be made but only to what is happening inside the method, for example:

def inject_shops
or def inject_activated_shops_with_address_states

app/serializers/api/shop_relatives_serializer.rb
```diff
@@ -0,0 +1,17 @@
+class Api::ShopRelativesSerializer < ActiveModel::Serializer
```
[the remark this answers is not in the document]
luisramos0: relatives are producers right?
you are creating a new concept "relatives", is it really necessary?

maybe it does not need to be specific to the relatives, maybe this can be simply called EnterpriseThinSerializer, it's just a thin representation of an enterprise, it's not related to relationships between enterprises as the relatives implies.
I'd call this EnterpriseThinSerializer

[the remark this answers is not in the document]
luisramos0: regarding the names of the serializers:

option 1 "named serializers" would be to:
- rename Api::ShopSerializer to Api::EnterpriseShopfrontSerializer
- rename Api::ShopRelativesSerializer to Api::EnterpriseThinSerializer
- rename Api::ShopsSerializer to Api::EnterpriseWithIconsSerializer

option 2 "for serializers" would be to:
- rename Api::ShopSerializer to Api::EnterpriseSerializerForShopfront
- rename Api::ShopRelativesSerializer to Api::EnterpriseSerializerForRelatives
- rename Api::ShopsSerializer to Api::EnterpriseSerializerForMaps

option 3 "mix" would be to:
- rename Api::ShopSerializer to Api::EnterpriseSerializerForShopfront
- rename Api::ShopRelativesSerializer  to Api::EnterpriseThinSerializer
- rename Api::ShopsSerializer to Api::EnterpriseWithIconsSerializer

I vote to keep everything in the Enterprise namespace and not create yet another domain concept called Shop, including the api controller.

I think this is an important API design decision because we should then use the same approach everywhere.
ping @sauloperez  @mkllnk @kristinalim 
What do you guys think?

sauloperez:
Amazing explanation @Matt-Yorkley! Do you think we could convert it into a wiki page that we could refer to as move along with perf. improvements? It could serve as a design guideline and make it easier for other contributors to jump in.

app/helpers/injection_helper.rb
Matt-Yorkley: I was trying to keep the line under 100 characters, but I think that cop is disabled in injection_helper, so I could make it one line. I think I prefer it that way too, tbh. :+1:

Matt-Yorkley:
Thanks @luisramos0 ! I was really unsure about the naming and categorizing. Calling the representation `Shopfront` and having it under the `Enterprises` namespace is great.

luisramos0:
yeah, the explanation is amazing!!!

app/serializers/api/enterprise_shopfronts_serializer.rb
```diff
@@ -0,0 +1,53 @@
+# Represents the minimum details of an Enterprise when all shopfronts are being listed
```
luisramos0: ah, now I get it :-) this is the serializer for the shoplist
EnterpriseShoplistSerializer?

Matt-Yorkley: Yes, so we have:
1. A _thin_ representation of a shopfront for listing all shopfronts
2. A fatter representation of the details of a single shopfront
3. An even thinner representation of related producers or hubs, which is nested in `2`.

Matt-Yorkley:
I've been using the UK Maps API key and a large dataset (smaller than UK prod though).

`BEFORE:`
maps_before

`AFTER:`
maps_after

41s -> 0.7s :tada:

app/serializers/api/enterprise_shopfronts_serializer.rb
Matt-Yorkley: @luisramos0 I went with: 
1. EnterpriseShopfrontsSerializer 
2. EntrpriseShopfrontSerializer
3. EnterpriseThinSerializer

Does that sound ok? I've moved the API endpoint as well and scrapped the `api/shop_controller`.

luisramos0: yeah, awesome!
I dont get much context from "shopfronts" so I prefer EnterpriseShoplistSerializer

but with that loadtime improvement (40s - 1 sec) we will approve the PR even with a BananasSerializer ;-)

Matt-Yorkley: I've changed it to: `EnterpriseShopfrontListSerializer`

luisramos0: awesome!

Matt-Yorkley:
I've removed the `WIP` tag, this is ready for review

app/helpers/injection_helper.rb
```diff
@@ -10,12 +10,8 @@ def inject_enterprises(enterprises = Enterprise.activated.includes(address: :sta
...
   end
 
-  def inject_map_pointers
-    inject_json_ams(
-      'enterprises',
-      Enterprise.activated.includes(address: :state).all,
-      Api::ShopsSerializer
-    )
+  def inject_shopfronts
+    inject_json_ams 'enterprises', Enterprise.activated.includes(address: :state).all, Api::EnterpriseShopfrontListSerializer
```
luisramos0: this is now failing on codeclimate, line too long... some break lines will do
btw, this is should now be  def inject_enterprise_shopfront_list

Matt-Yorkley:
I added some specs coverage for the main parts, but I havn't managed to test the taxon-gathering in the serializers yet, I think it requires a more complex setup phase for the examples to work.

luisramos0:
moving to test ready as epr Pau's comment: "I'm up for shipping this as soon as we improve tiny bits and iterate from there."

Matt-Yorkley:

RachL:
Everything seems to be working good on UK map. I tried on old and created also a new enterprise.

Will stage this on staging Fr to test group now.

image

Matt-Yorkley:
It looks like the groups menu item is missing on UK staging, but you can still reach the page directly: https://staging.openfoodnetwork.org.uk/groups

RachL:
@Matt-Yorkley ah thx, I actually was too lazy to create one... if weren't I would have seen it 🙈 

Map is as fast as lightning!!! Group page on the other end :trollface: so not sure what is due to the page, or to the map. But everything is working as usual, so moving this to ready to go!

image

Matt-Yorkley:
The group pages are about 5-10% faster, but only as a coincidental side-effect. As long as the map modals still display there as before, it's fine.

luisramos0:
like the French say: super!!! 👏

enricostano:
@Matt-Yorkley

Issue #3792, opened 2019-05-05
Can't create or edit an order cycle on German instance


kirstenalarsen:
## Description
On the German instance

## Expected Behavior
When I go to create or edit an Order Cycle, the view should open, enable me to type in the 'name' box 

![image](https://user-images.githubusercontent.com/2911176/57190186-1403a100-6f5b-11e9-922a-638f56bdd564.png)

## Actual Behaviour
The view loads but I can't type in the Name field 

and my lists of Suppliers and Distributors are empty so I can't add any.

![image](https://user-images.githubusercontent.com/2911176/57190214-5d53f080-6f5b-11e9-973a-04fa930d464e.png)

I am set to 'any' and I have shipping and payment methods set up. There is a producer with lots of products (DorV-online) and I have permissions to add to order cycle and inventory.

When I use Browser tools I see this error
![image](https://user-images.githubusercontent.com/2911176/57190294-fb47bb00-6f5b-11e9-9601-ace8c9b4f73b.png)

then a bit later on reload I am seeing this
![image](https://user-images.githubusercontent.com/2911176/57190265-c63b6880-6f5b-11e9-917e-7d85cc2b97e4.png)

I have just created a completely new Enterprise and tried with this in case some kind of weird data problem, but same thing happens

## Steps to Reproduce

1. Go to german instance
2. Try to create or edit and order cycle

## Animated Gif/Screenshot
http://recordit.co/yI0PcvsTty

## Context
Germany meant to be launching tomorrow (Monday 6 May)

## Severity
S1 - Order cycle is a critical function. I can't find a workaround
-->

## Your Environment
Chrome MACOS
Tried Firefox as well, same problem

kirstenalarsen:
From Julia just now - 
![image](https://user-images.githubusercontent.com/2911176/57190624-88404380-6f5f-11e9-89c1-2c723fc04e8a.png)

mkllnk:
I just looked at this and it seems to be a performance problem. There are several things that need to be loaded and it just took one minute for me to load. Then I could edit everything.

kirstenalarsen:
hmmm - yes I just came to similar conclusion - as I noticed the 'loading ... ' thing in the bottom. But mine is still loading - way more than a minute. 

and now the loading thing has disappeared and I have these errors
![image](https://user-images.githubusercontent.com/2911176/57190665-4a8fea80-6f60-11e9-8223-b28b6a5f36a1.png)

[...]

Issue #3760, opened 2019-04-29
Performance Improvements


Matt-Yorkley:
Editing @Matt-Yorkley's original epic with the outcomes of the performance session at #gathering19

Performance is a feature.

This feature has taken a high priority agreed at #gathering19

Success Criteria for Performance:
- Pages will load in under 5 seconds. 3-5 is acceptable.
- OFN supports unlimited enterprises with 10,000 products. 

Priority pages (in order of decreasing priority)
- /maps - Merged
- /shops - Awaiting PR merge and release
- /producer - Awaiting PR merge and release
- /groups - Awaiting PR merge and release
- /admin/ edit order cycles - Spike
- /enterprise/shop - Dev ready
- /reports/orders+fulfillment - Dev ready
- /super-admin/enterprises
- /super-admin/order-cycles
- /bom - Spike
- /admin/products
- /admin/enterprises - only a problem for hubs with large numbers of producers
- /checkout

RachL:
@Matt-Yorkley thanks! There already was an epic but the title was not clear. So I've put all the issues of the previous one in yours and left only the epic in the bugbacklog. Ping @daniellemoorhead I know you wanted to have a look on those issues as well and grouping them. Does this make any sense?

Matt-Yorkley:
Great! Yeah I think that makes more sense. The old epic #3493 was more like a bug report. It's already s2 for France, and I've added UK to the title of the issue, because it's an s2 for us as well. 

With the new monitoring we've added in the last week we can quite safely say the performance issues with the above pages are the cause of the downtimes, so now we have a clearer idea on how to proceed.

RachL:
@Matt-Yorkley yes the new monitoring looks great! It is only provisioned in UK for now right?

Matt-Yorkley:
![coming-soon](https://media.giphy.com/media/3o72FkiKGMGauydfyg/giphy.gif)

sauloperez:
IMO we should also consider creating a specific performance backlog because lots of unrelated issues and epics will start soon popping up, as suggested by @luisramos0

luisramos0:
Awesome, in the gathering we should scope this so that we know what we want to address in the short term and what is not essential. Performance is as infinite topic and we will only be able to address a few things in the short term.

Matt-Yorkley:
[...]

PR #3802, opened 2019-05-06
3760 Optimizations related to OCs

kristinalim:
#### What? Why?

- Relates to #3792 
- Relates to #3760

This set of optimizations should affect not just the OC pages, but there should also be minor improvements in admin pages in general, and possible minor improvements for code that involve OCs.

#### What should we test?

**Edit OC page**

Smoke test of the edit OC page should pass. Test with an OC where the coordinator is not the only producer.

**Warning if there is an invalid OC setup**

Also, as admin, remove all shipping methods for a producer-hub with an active OC. You can easily undo this change if you simply edit the shipping methods and untick the producer-hub. The warning that there is an invalid OC setup should still appear at the top of the page.

#### Release notes

- Improve performance, especially in the edit order cycle page.

Changelog Category: Fixed

Needs discussion on whether this should be code reviewed and tested in the pre-v2.0 branch.

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -96,12 +96,10 @@ class Enterprise < ActiveRecord::Base
   scope :not_ready_for_checkout, lambda {
     # When ready_for_checkout is empty, ActiveRecord generates the SQL:
     # id NOT IN (NULL)
-    # I would have expected this to return all rows, but instead it returns none. To
-    # work around this, we use the "OR ?=0" clause to return all rows when there are
-    # no enterprises ready for checkout.
-    where('id NOT IN (?) OR ?=0',
-          Enterprise.ready_for_checkout,
-          Enterprise.ready_for_checkout.count)
+    # I would have expected this to return all rows, but instead it returns none. But we want to
+    # return all rows when there are no enterprises ready for checkout.
+    ready_enterprises = Enterprise.ready_for_checkout
+    ready_enterprises.present? ? where("id NOT IN (?)", ready_enterprises) : where("TRUE")
   }
   scope :is_primary_producer, where(:is_primary_producer => true)
   scope :is_distributor, where('sells != ?', 'none')
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -97,7 +97,11 @@ class Enterprise < ActiveRecord::Base
     # When ready_for_checkout is empty, return all rows when there are no enterprises ready for
     # checkout.
     ready_enterprises = Enterprise.ready_for_checkout
-    ready_enterprises.present? ? where("id NOT IN (?)", ready_enterprises) : where("TRUE")
+    if ready_enterprises.present?
+      where("id NOT IN (?)", ready_enterprises)
+    else
+      where("TRUE")
+    end
   }
   scope :is_primary_producer, where(:is_primary_producer => true)
   scope :is_distributor, where('sells != ?', 'none')
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -94,10 +94,8 @@ class Enterprise < ActiveRecord::Base
       select('DISTINCT enterprises.*')
   }
   scope :not_ready_for_checkout, lambda {
-    # When ready_for_checkout is empty, ActiveRecord generates the SQL:
-    # id NOT IN (NULL)
-    # I would have expected this to return all rows, but instead it returns none. But we want to
-    # return all rows when there are no enterprises ready for checkout.
+    # When ready_for_checkout is empty, return all rows when there are no enterprises ready for
+    # checkout.
     ready_enterprises = Enterprise.ready_for_checkout
     ready_enterprises.present? ? where("id NOT IN (?)", ready_enterprises) : where("TRUE")
   }
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -9,7 +9,7 @@ class OrderCycle < ActiveRecord::Base
   has_many :exchanges, :dependent => :destroy
 
   # These scope names are prepended with "cached_" because there are existing accessor methods
-  # :incoming_exchanges and :outgoing_exchanges. Future work can be done to name these properly.
+  # :incoming_exchanges and :outgoing_exchanges.
   has_many :cached_incoming_exchanges, conditions: { incoming: true }, class_name: "Exchange"
   has_many :cached_outgoing_exchanges, conditions: { incoming: false }, class_name: "Exchange"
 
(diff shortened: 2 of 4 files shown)


kristinalim:
### Setting up initial scenario

I did the following locally to set up the base scenario with which to compare pre-PR and post-PR:

1. Start with the DB set up with `script/setup`, which also loads minimal sample data:

    ```sh
    bundle exec rake db:drop
    script/setup
    ```

2. Update the 500-product spreadsheet sent by Julia for Germany, changing `producer` for all to "Fredo's Farm Hub" and `shipping_category` to "Pick-up". Save this as `db-performance_base.csv` in the application root.
3. Create the taxonomies and tax categories in the spreadsheet.
4. Add the shipping category "Pick-up".
5. Import the 500 products in the spreadsheet through the Product Import page.
6. Dump the DB for reuse: `pg_dump -U ofn -W -F t open_food_network_dev > db-performance_base.tar`

### Initial numbers

#### [/admin](http://localhost:3000/admin)

| Timestamp | Response Time | URL | |
|-|-|-|-|
| 14:37:55 | 553 ms | /admin | [Detail][SQL (41)] |

#### [/order_cycles](http://localhost:3000/admin/order_cycles)

| Timestamp | Response Time | URL | |
|-|-|-|-|
| 14:42:35 | 211 ms | /admin/order_cycles.json | [Detail][SQL (77)] |
| 14:42:35 | 160 ms | /admin/schedules.json | [Detail][SQL (27)] |
| 14:42:35 | 118 ms | /admin/enterprises/visible.json | [Detail][SQL (36)] |
| 14:42:34 | 432 ms | /admin/order_cycles | [Detail][SQL (33)] |

#### [Editing "Multi Shop OC" with 6 variants](http://localhost:3000/admin/order_cycles/4/edit)

| Timestamp | Response Time | URL | |
|-|-|-|-|
| 14:45:21 | 1236 ms | /admin/order_cycles/4.json | [Detail][SQL (170)] |
| 14:45:21 | 68 ms | /admin/schedules.json | [Detail][SQL (27)] |
| 14:45:21 | 154 ms | /admin/enterprise_fees/for_order_cycle.json | [Detail][SQL (44)] |
| 14:45:15 | 5683 ms | /admin/enterprises/for_order_cycle.json | [Detail][SQL (2079)] |
| 14:45:14 | 528 ms | /admin/order_cycles/4/edit | [Detail][SQL (30)] |

app/models/enterprise.rb
```diff
@@ -96,12 +96,10 @@ class Enterprise < ActiveRecord::Base
...
-    # I would have expected this to return all rows, but instead it returns none. To
-    # work around this, we use the "OR ?=0" clause to return all rows when there are
-    # no enterprises ready for checkout.
-    where('id NOT IN (?) OR ?=0',
-          Enterprise.ready_for_checkout,
-          Enterprise.ready_for_checkout.count)
+    # I would have expected this to return all rows, but instead it returns none. But we want to
+    # return all rows when there are no enterprises ready for checkout.
+    ready_enterprises = Enterprise.ready_for_checkout
+    ready_enterprises.present? ? where("id NOT IN (?)", ready_enterprises) : where("TRUE")
```
kristinalim: This reduces the number of queries only by 1 for each call.

app/models/order_cycle.rb
```diff
@@ -8,6 +8,14 @@ class OrderCycle < ActiveRecord::Base
 
   has_many :exchanges, :dependent => :destroy
 
+  # These scope names are prepended with "cached_" because there are existing accessor methods
+  # :incoming_exchanges and :outgoing_exchanges. Future work can be done to name these properly.
+  has_many :cached_incoming_exchanges, conditions: { incoming: true }, class_name: "Exchange"
+  has_many :cached_outgoing_exchanges, conditions: { incoming: false }, class_name: "Exchange"
+
+  has_many :suppliers, source: :sender, through: :cached_incoming_exchanges, uniq: true
+  has_many :distributors, source: :receiver, through: :cached_outgoing_exchanges, uniq: true
```
sauloperez: can you explain a little bit what's the difference between the SQL resulting of this and the previous implementation?

kristinalim: @sauloperez One effect is that `order_cycle.suppliers` and `order_cycle.distributors` would be cached, so running a method multiple times would query only once. This also makes it possible to use eager-loading, like: `order_cycle_scope.includes(:distributors, :suppliers)`

app/models/enterprise.rb
luisramos0: I think the comment above needs to be adapted to the new version.

also, quite a complicated thing happening, so I'd prefer to see the condition in multiple lines:
```suggestion
    if ready_enterprises.present?
       where("id NOT IN (?)", ready_enterprises)
    else
       where("TRUE")
    end
```

app/models/order_cycle.rb
luisramos0: looks good but I think we need to clean up a little bit.
the code and comments should be about the present. future work is represented on github.
can we move the references to the future to github tech debt issue?

there are two:
1 - "Future work can be done to name these properly."
2 - "TODO: DRY the incoming/outgoing clause used in several cases below" (this TODO is 5 years old, which is a common thing for TODOs)

even if we dont create the tech debt issue I think both points should be removed from the code.

app/models/enterprise.rb
kristinalim: Thanks, @luisramos0! Yes, I agree. Changed here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3802/commits/759316834154ed9c463c0ef06dcba469a56f2c72

app/models/order_cycle.rb
kristinalim: Regarding this old comment:

```
# TODO: DRY the incoming/outgoing clause used in several cases below
# See Spree::Product definition, scopes variants and variants_including_master
# This will require these accessors to be renamed
```

I checked these and didn't get useful information to understand the accessors, so I removed the entire comment.

Changed here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3802/commits/c0a5bcce92b4a22ce19b1e5cf3ea0391ffa2c79a

app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
```diff
@@ -28,15 +28,20 @@ def supplied_products
...
       object.supplied_products.visible_for(order_cycle.coordinator)
     else
       object.supplied_products
     end
   end
 
+  def products
+    return @products unless @products.nil?
+
+    @products = products_scope.includes(:supplier, master: [:images])
```
Matt-Yorkley: Can lines 40-42 just be replaced with this?
```
@products ||= products_scope.includes(:supplier, master: [:images])
```

kristinalim: Oh, right! I didn't notice, I was updating just the bottom part of the method. Code simplified here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3802/commits/56bc497fe272e1549c71228f56da3519044536c8 @luisramos0

luisramos0:
2 approvals = test ready

RachL:
This is ready to go !

https://docs.google.com/document/d/1XTPWaukvim3k1WG6PKaaIiUHl4Uzld8M7i8tktO_myM/edit#

PR #3814, opened 2019-05-06
Test empty variants_attrs in bulk_update

sauloperez:
#### What? Why?

This covers the scenario we saw in https://app.bugsnag.com/katuma/katuma/errors/5cd0595bece0b7001984f64c?event_id=5cd0595b0038be3b070d0000&i=sk&m=nw which raised our concerns.
There is some other context we might be missing that makes it not possible to reproduce.

I guess then, that this is an isolated case that does not prevent us from deploying v2.0.0. 

#### What should we test?

This is just a test so nothing to test.

#### Release notes
Check that sending empty `variant_attributes` through bulk update does not make the update fail.

Changelog Category: Added

### DIFF
--- a/spec/controllers/spree/admin/products_controller_spec.rb
+++ b/spec/controllers/spree/admin/products_controller_spec.rb
@@ -71,6 +71,57 @@ describe Spree::Admin::ProductsController, type: :controller do
         )
       end
     end
+
+    context 'when passing empty variants_attributes' do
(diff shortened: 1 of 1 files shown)


luisramos0:
I think this should be pointing at master, no need to make this a hotfix.

mkllnk:
I put this in Ready to Go. It's waiting for @sauloperez to decide to which branch to merge it to.

Issue #3816, opened 2019-05-07
Missing translations on Admin / Config / Content


mkllnk:
## Description

As super admin you can visit Admin -> Configuration -> Content to change logos, footer and other text. On a fresh instance, these fields should contain defaults depending on the default language. Unfortunately, translations are not available in content defaults and missing translations are shown.

## Expected Behavior

There is a default text in each field.

## Actual Behaviour

There is a translation missing message in each field.

## Steps to Reproduce

1. Use a new instance or reset the database or use a database that didn't set all content fields. Delete the cache via `rm tmp` in the app directory if you reset the database.
2. Log in as admin and go to Admin -> Configuration -> Content.

## Animated Gif/Screenshot

![Screenshot from 2019-05-07 16-11-56](https://user-images.githubusercontent.com/3524483/57275935-522bcc80-70e3-11e9-8869-d577c0059970.png)

## Context

I tried to add a new content field and my dev environment showed these errors when testing.

## Severity

s4 as it only affects super admins setting up a new instance.

## Your Environment

* Version used: 1.31.0

## Possible Fix

Don't define `ContentConfig` in an initializer while I18n is not available. Create a service `ContentConfig` instead. If that doesn't help, re-implement the storage of content, not using Spree's preferences.

RachL:
@sigmundpetersen @filipefurtad0 this issue is in dev, just checking it's the correct position?

filipefurtad0:
Hum, indeed thanks @RachL.
I don't think so - I guess this was automatically placed here after reopening the issue (there's nobody assiged). I'll move it to all the things :+1:

PR #3824, opened 2019-05-08
Aus content updates

mkllnk:
#### What? Why?

The Australian team updated their pricing model which needed some content changes on the website. Since we are using the source locale `en.yml` we need to change text via pull requests instead of going through Transifex.

We also wanted to change the alert banner on the home page. I guessed that this customisation could be desirable for other instances and made it customisable via the admin interface. It's still 100% downwards compatible and the usual register call is shown if you don't put a custom text in the admin panel. While the usual register call is translated into different languages, any custom text used for the alert banner would be static and not translated. So until we get to the multi-lang 2.0 feature, the custom banner may not be attractive to multi-lang instances.

#### What should we test?

- The alert banner and menu items look as usual.
- The alert banner can be customised via Admin / Configuration / Content.

This is on Australian production already. A quick test on staging that it doesn't mess up anything else would still be nice. :-)

#### Release notes

Added: The homepage alert banner can be customised by admin users.

Changelog Category: Added


app/models/content_configuration.rb
```diff
@@ -12,6 +12,20 @@ class ContentConfiguration < Spree::Preferences::FileConfiguration
...
+  # I intended to set a default for home_page_alert_html including I18n text.
+  #
+  #      default: "<a href='/register' target='_blank'>
+  #             #{I18n.t('shared.register_call.selling_on_ofn')}&nbsp;
+  #             <strong>#{I18n.t('shared.register_call.register')}
+  #             <i class='ofn-i_054-point-right'></i></strong>
+  #             </a>"
+  #
+  # Unfortunately, this configuration is initialised and cached before locales
+  # are loaded. So I chose to not set a default and use a translatable view instead.
```
luisramos0: I dont think we should tell a story in comments. imo the story can go in the commit message or the PR. comments should be straight to the point about the current state of the code. I'd keep only these 2 last lines:
# All the following defaults using I18n don't work.
# https://github.com/openfoodfoundation/openfoodnetwork/issues/3816

sigmundpetersen:
Yes, `en_AU.yml`. Then en.yml could be the source file that is actually not used on any instance. I think that would be good.

app/models/content_configuration.rb
mkllnk: You are right, I covered that in the issue now. There was a point in my development when that comment made more sense and I didn't want to lose the HTML code to paste. But in the end I kept the view and the default is not need. :+1:

lin-d-hop:
Tests look good!
Testing notes here:
https://docs.google.com/document/d/1UUm4MqJbktMiaOcWCSsrAuRpq5ZXDpZ6brxduSKMj3Y/edit#

PR #3833, opened 2019-05-09
Fix /groups pages performance

Matt-Yorkley:
#### What? Why?

Related to #3760 

Improves performance on `/groups` index and show pages. There's still a lot of mess around the way groups load data, but this should be a huge improvement in the mean time.

#### What should we test?

Groups pages should work as before, and load much more quickly. Also check the `/map` page and map tab in group pages, as this includes a serializer improvement that should decrease their load time a bit more.

#### Release notes

Improved performance for groups pages

Changelog Category: Fixed


Matt-Yorkley:
The groups `#index` and `#show` actions take around 40-60 seconds on UK production, and I'm 99% sure they are the cause of our intermittent downtime issues. This should get them to <1 second. :tada:

app/helpers/serializer_helper.rb
```diff
@@ -3,4 +3,13 @@ def ids_to_objs(ids)
...
     ids.map { |id| {id: id} }
   end
+
+  # Returns an array of the fields a serializer needs from it's object
+  # so we can #select only what the serializer will actually use
+  def required_attributes(model, serializer)
+    model_attributes = model.attribute_names
+    serializer_attributes = serializer._attributes.keys.map(&:to_s)
+
+    (serializer_attributes & model_attributes).map { |attr| "#{model.table_name}.#{attr}" }
```
luisramos0: do you know how much of a performance impact this has?
I'd not do this if it does give us a big improvement. it's complicated.

because it is not easy, this method really needs unit tests...

app/helpers/injection_helper.rb
```diff
@@ -12,6 +12,16 @@ def inject_enterprises(enterprises = Enterprise.activated.includes(address: :sta
...
+  def inject_groups
+    select_only = required_attributes EnterpriseGroup, Api::GroupListSerializer
+
+    inject_json_ams(
+      'groups',
+      EnterpriseGroup.on_front_page.by_position.select(select_only).includes(address: :state).all,
+      Api::GroupListSerializer
+    )
+  end
+
```
luisramos0: OCD mode triggered... can you please make the double quotes (or single quotes) consistent in this file? :-)

app/assets/javascripts/darkswarm/services/groups.js.coffee
```diff
@@ -1,14 +1,3 @@
...
-    groups_by_id: {} 
-    constructor: ->
-      for group in @groups
-        @groups_by_id[group.id] = group
-      @dereference()
-    dereference: ->
-      for group in @groups
-        Dereferencer.dereference group.enterprises, Enterprises.enterprises_by_id 
-      for enterprise in Enterprises.enterprises
-        Dereferencer.dereference enterprise.groups, @groups_by_id
```
luisramos0: :+1: good to see one more instance of this Dereferencering thing going away! We only have 4 or 5 more usages of it...

app/views/groups/index.html.haml
```diff
@@ -1,10 +1,7 @@
 - content_for(:title) do
   = t :groups_title
 
-= inject_enterprises
-
-:javascript
-  angular.module('Darkswarm').value('groups', #{render partial: "json/groups", object: @groups})
+= inject_groups
```
luisramos0: :+1: ideally this would be a new api endpoint, not in scope for this PR.

app/assets/javascripts/darkswarm/controllers/group_page_controller.js.coffee
```diff
@@ -1,10 +1,8 @@
-Darkswarm.controller "GroupPageCtrl", ($scope, group_enterprises, Enterprises, MapConfiguration, OfnMap, visibleFilter, Navigation) ->
+Darkswarm.controller "GroupPageCtrl", ($scope, enterprises, Enterprises, MapConfiguration, OfnMap, visibleFilter, Navigation) ->
   $scope.Enterprises = Enterprises
 
-  all_enterprises_by_id = Enterprises.enterprises_by_id
-
-  dereferenced_enterprises = group_enterprises.map (enterprise) =>
-    all_enterprises_by_id[enterprise.id]
+  dereferenced_enterprises = enterprises.map (enterprise) =>
+    Enterprises.enterprises_by_id[enterprise.id]
```
luisramos0: you have removed the dereferencing right? so these enterprises are not dereferenced.
I guess you can change the name of this dereferenced_enterprises variable.

app/assets/javascripts/darkswarm/services/groups.js.coffee
Matt-Yorkley: Yes, I've come to realise that anywhere this `Dereferencer` is being used in Angular, I can be sure that:
- something horrible is happening
- it's impossible to follow what it's doing

app/assets/javascripts/darkswarm/controllers/group_page_controller.js.coffee
Matt-Yorkley: Right, there's still some dereferencing going on in some connected places, but this part is now actually just `enterprises_by_id`. I'll update it.

app/helpers/serializer_helper.rb
Matt-Yorkley: I'll add some tests. It seems to give ~10% improvement on maps.

app/assets/javascripts/darkswarm/services/groups.js.coffee
luisramos0: LOLOL
it's documented here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Angular-and-OFN#dereferencing

app/views/groups/index.html.haml
Matt-Yorkley: Yes, I wanted to shift some of these ams injection parts to the API, but there were some problems when I tried it with /shops for example. We can do it at some point.

Matt-Yorkley: Spec added: https://github.com/openfoodfoundation/openfoodnetwork/pull/3833/commits/d9bafd56daaca7e98d030ae5205d15be7e8bcf55#diff-81148c208b0c1bdbfcd8e77f3fc82e57

app/helpers/serializer_helper.rb
Matt-Yorkley: Spec added: https://github.com/openfoodfoundation/openfoodnetwork/pull/3833/commits/d9bafd56daaca7e98d030ae5205d15be7e8bcf55#diff-81148c208b0c1bdbfcd8e77f3fc82e57

app/helpers/injection_helper.rb
luisramos0: :-D

app/assets/javascripts/darkswarm/controllers/group_page_controller.js.coffee
```diff
@@ -1,12 +1,10 @@
-Darkswarm.controller "GroupPageCtrl", ($scope, group_enterprises, Enterprises, MapConfiguration, OfnMap, visibleFilter, Navigation) ->
+Darkswarm.controller "GroupPageCtrl", ($scope, enterprises, Enterprises, MapConfiguration, OfnMap, visibleFilter, Navigation) ->
   $scope.Enterprises = Enterprises
 
-  all_enterprises_by_id = Enterprises.enterprises_by_id
+  enterprises_by_id = enterprises.map (enterprise) =>
+    Enterprises.enterprises_by_id[enterprise.id]
```
mkllnk: Is `enterprises_by_id` a list or a map? The name indicates that it's a map but the code seems to indicate it's a simple list of enterprises. What is the goal here?

Matt-Yorkley:
Ah yes, I'm loading the data for the partials via the API already in #3836, and it's required here.

Matt-Yorkley:
Moving these commits into #3836, as they are interdependent, and can't be tested in isolation.

PR #3836, opened 2019-05-09
Improve /shops, /producers, and /groups performance

### LINKED ISSUE #3099: /shops taking 30+ seconds to load (UK)

Matt-Yorkley:
#### What? Why?

Closes #3099

Improves performance on `/shops`, `/producers`, and `/groups` pages. I didn't quite manage to work the same miracles as #3801, but it's still a big improvement. With my non-production data:

- `/shops` is reduced from ~15 seconds to ~1.2
- `/producers` is reduced from ~28 to ~5
- `/groups` (both #index and #show) were around 20-40 seconds, reduced to 1-3 seconds

If we want <1 second page loads we'll need to completely overhaul the search and filtering on these pages. I figured out a hybrid solution where we still load a lot more data then we need to and do some dereferencing gymnasticis, but we do much less than before whilst keeping all the existing functionality.

#### What should we test?

Search and filtering work as before on both `/shops`, `/producers`, and `/groups` pages. All data should be displayed as before. Pages are faster. Note: caching makes a huge difference! The first time a page is loaded will take much longer than the second or third time.

Success criteria: 
- /shops, /producers and /groups all load within 3-5 seconds.
- Instances support shops with 10,000 products. (@lin-d-hop and @kirstenalarsen will explore importing to a staging server to enable this testing) Note: high-volume shops has a separate issue: #3853 

#### Release notes

Improved performance on shops, producers, and groups pages

Changelog Category: Fixed


Matt-Yorkley:
Okay, I've hit a snag: I was excluding enterprises that aren't shops, but the shops page has a checkbox for "show profiles", and these are enterprises that don't sell anything. If I include these in the results again, the load time balloons from ~1.5 seconds to ~9 seconds...

`SHOPS:`
profiles

`SHOPS AND PROFILES:`
profiles2

mkllnk:
It was once on my todo list to convert that profiles checklist into a button at the bottom and lazy-load them. But it was crossed off when other things became more important. Your tests show that it would still be good to do.

luisramos0:
some conflicts from the transpec PR...

Matt-Yorkley:
Looks like this also needs #3847...

Matt-Yorkley:
groups_before_after

Matt-Yorkley:
Specs are passing now :smile:

app/helpers/serializer_helper.rb
```diff
@@ -3,4 +3,13 @@ def ids_to_objs(ids)
...
   end
+
+  # Returns an array of the fields a serializer needs from it's object
+  # so we can #select only what the serializer will actually use
+  def required_attributes(model, serializer)
+    model_attributes = model.attribute_names
+    serializer_attributes = serializer._attributes.keys.map(&:to_s)
+
+    (serializer_attributes & model_attributes).map { |attr| "#{model.table_name}.#{attr}" }
+  end
```
sauloperez: does this really make a difference in terms of performance?

app/serializers/api/group_list_serializer.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  class GroupListSerializer < ActiveModel::Serializer
```
sauloperez: It'd be great to describe a bit what this is with a top level comment

app/views/enterprises/shop.html.haml
```diff
@@ -5,7 +5,7 @@
 - content_for(:image) do
   = current_distributor.logo.url
 
-= render partial: 'json/injection_ams', locals: enterprises
+= inject_enterprise_shopfront(@enterprise)
```
sauloperez: it'd be really good to stick to locals instead of implicit ivars. It's a lesson learned in the Rails community: the implicitness of ivars becomes soon a mess.

app/views/producers/index.html.haml
```diff
@@ -1,7 +1,7 @@
 - content_for(:title) do
   = t :producers_title
 
-= inject_enterprises
+= inject_enterprises(@enterprises)
```
sauloperez: here too

spec/helpers/serializer_helper_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper'
+
+describe SerializerHelper, type: :helper do
+  let(:serializer) do
+    class ExampleEnterpriseSerializer < ActiveModel::Serializer
+      attributes :id, :name
+    end
+    ExampleEnterpriseSerializer
+  end
```
sauloperez: you could simply define and use ExampleEnterpriseSerializer without a let. There is no restriction on which scope you define a class.

app/controllers/groups_controller.rb
```diff
@@ -1,10 +1,6 @@
 class GroupsController < BaseController
   layout 'darkswarm'
 
-  def index
-    @groups = EnterpriseGroup.on_front_page.by_position
-  end
-
```
sauloperez: what do we render now when accessing /groups then?

app/models/enterprise.rb
```diff
@@ -285,6 +285,14 @@ def distributed_taxons
       select('DISTINCT spree_taxons.*')
   end
 
+  def current_distributed_taxons
+    Spree::Taxon
+      .select("DISTINCT spree_taxons.*")
+      .joins(products: :variants_including_master)
+      .joins("INNER JOIN (#{current_exchange_variants.to_sql}) \
```
sauloperez: have we checked the EXPLAIN ANALYZE of this big query? this nesting here makes it quite large.

sauloperez:
I just left some minor comments. I'm happy to move this forward and address any of these in a separate PR if we want it.

app/assets/javascripts/darkswarm/controllers/hub_node_controller.js.coffee
```diff
@@ -1,9 +1,44 @@
...
+    $scope.shopfront_loading = true
+    $scope.toggle_tab(event)
+
+    $http.get("/api/enterprises/" + $scope.hub.id + "/shopfront")
+      .success (data) ->
+        $scope.shopfront_loading = false
+        $scope.hub = data
+        $scope.enterprise_details[$scope.hub.id] = $scope.hub
+      .error (data) ->
+        console.error(data)
```
luisramos0: (1) data loading should not be part of the controller.
it should be extracted to a separate file, like a resource, for example.

(2) also, I think we should continue the separation of darkswarm in domains like admin (I remember Rob mentioning this somewhere).
so, instead of 
app/assets/javascripts/darkswarm/controllers/hub_node_controller.js.co
we would have 
app/assets/javascripts/darkswarm/shops/hub_node_controller.js.coffee

and then this data loading code could be extracted to a new file:
app/assets/javascripts/darkswarm/shops/enterprises_resource.js.coffee

if you only implement suggestion (1) above, this code could be extracted to app/assets/javascripts/darkswarm/services/enterprise_resource.js.coffee (not an api endpoint)
or we could create a new file for api/enterprise
app/assets/javascripts/darkswarm/resources/api/enterprise_resource.js.coffee (for api endpoints)

this is important as we would make a lot more explicit the repeated code in darkswarm and admin:
app/assets/javascripts/admin/resources/resources/enterprise_resource.js.coffee

I think we need to agree on a location for these things. Otherwise, the existing mess will just keep growing.
What do you think?

app/assets/javascripts/darkswarm/controllers/producer_node_controller.js.coffee
```diff
@@ -1,6 +1,39 @@
...
+    $scope.shopfront_loading = true
+    $scope.toggle_tab(event)
+
+    $http.get("/api/enterprises/" + $scope.producer.id + "/shopfront")
+      .success (data) ->
+        $scope.shopfront_loading = false
+        $scope.producer = data
+        $scope.enterprise_details[$scope.producer.id] = $scope.producer
+      .error (data) ->
+        console.error(data)
```
luisramos0: same as above, I dont think this can be in a controller.

app/assets/javascripts/darkswarm/services/groups.js.coffee
```diff
@@ -1,14 +1,3 @@
...
-    constructor: ->
-      for group in @groups
-        @groups_by_id[group.id] = group
-      @dereference()
-    dereference: ->
-      for group in @groups
-        Dereferencer.dereference group.enterprises, Enterprises.enterprises_by_id 
-      for enterprise in Enterprises.enterprises
-        Dereferencer.dereference enterprise.groups, @groups_by_id
-
```
luisramos0: dereference 🔥 
🎉

app/controllers/groups_controller.rb
luisramos0: the same, this is done in the view now with inject_groups

app/assets/javascripts/darkswarm/controllers/hub_node_controller.js.coffee
Matt-Yorkley: Ah yes, I was planning on extracting these to a new resource when it was all working correctly. :+1:

luisramos0:
ping @Matt-Yorkley  #3842 created some conflicts here

app/assets/javascripts/darkswarm/controllers/hub_node_controller.js.coffee
sauloperez: better to address in a separate PR IMO. I'll be much cleaner.

Matt-Yorkley:
I've rebased this to pick up the change in #3847 and moved it to test ready.

lin-d-hop:
@matt-yorkley can you fill me in on what the effect of the caching is?
When I make changes to enterprise details, shipping methods etc the changes don't flow into the pages. When are caches updated? Can I force refresh? As a user how do I know that my changes will flow through?

lin-d-hop:
In my testing I discovered that fields in the modals don't update when changes are made. I haven't tested all fields. All of the fields I did test are affected. 

@matt-yorkley thinks this is rails caching.

I checked this on UK prod and changes made flow directly through to user.

Back to InDev.... sorry crew.

Matt-Yorkley:
Great testing :+1:

Matt-Yorkley:
@lin-d-hop I just had a look at this and it was working for me on UK staging. Updated shipping methods, enterprise; description, email address, phone number, and they all worked.

What page were you on specifically? What fields did you update? If you're on the `/shops` page and the browser adds a hash into the URL, when you hit the reload button it won't actually reload.

lin-d-hop:
I did not successfully update anything I tried on the maps and shops pages.
I tried updating contact information (phone, website), description, shipping methods and product categories.

@matt-yorkley - If you are confident then move back to test-ready and I'll retest. I've taken UK-staging back :-)

lin-d-hop:
Tested on UK staging. All looks good:
https://docs.google.com/document/d/1kiPkPx_MiNqPxSVZ1SDB0Ni83ybsZI-U8qvPeKHtVAg/edit#

Gave a quick test on ES staging just to check as I am certain that in my original testing updates didn't flow through (I really really want this PR merged, so I tried a lot of things to make sure it wasn't just me). Everything updated as expected. No idea what happened there. C'est la vie.

mkllnk:
Awesome. I think everyone wants a new release now and deploy it quickly. :smile:

kirstenalarsen:
I think I can't have this until we're on v2 - is that right? In which case suddenly I'm ready for v2.0 ;)

RachL:
@kirstenalarsen it would also need to be included in a release. Currently instances are upgraded to v2.1, which is currently the latest release that does not include this PR.

Issue #3830, opened 2019-05-09
RuntimeError in app/jobs/products_cache_integrity_checker_job.rb:6


sauloperez:
## Error in Katuma

**RuntimeError** in **app/jobs/products_cache_integrity_checker_job.rb:6**
Products JSON differs from cached version for distributor: 93, order cycle: 469

[View on Bugsnag](https://app.bugsnag.com/katuma/katuma/errors/5c864dfee822db0019f45a62?event_id=5cd3d0a00039c766cb880000&i=gh&m=ci)

## Stacktrace

    app/jobs/products_cache_integrity_checker_job.rb:6 - perform
    script/delayed_job:5 - <main>

[View full stacktrace](https://app.bugsnag.com/katuma/katuma/errors/5c864dfee822db0019f45a62?event_id=5cd3d0a00039c766cb880000&i=gh&m=ci)

*Created by Katuma via Bugsnag*

## Description

Since v2.0.0 this error that used to happen sporadically now happens more often. This results in surpassing the 250 events/day free quota we have in Katuma. I did see an exceptional increase of these errors while we used v2.0.0 in staging but it wasn't clear whether the issue was limited to staging or not.

This background job checks whether for a given (distributor, order_cycle) pair the cache is stale or not, meaning that the JSON response stored in the cache is not equal to the actual and latest JSON response. This means that we're missing at least a scenario where the cache should be refreshed.

Also, because we enqueue as many of these jobs as cacheable exchanges (exchanges whose order cycle has at least opening or closing date and is not closed yet) and every hour, the flood of recurrent bugsnag notifications is considerable.

## Expected Behavior

The products cache should always be kept up to date by refreshing it when any of the entities involved in a shopfront is updated.

## Actual Behaviour

The products cache is not kept in sync with the actual data updates.

## Steps to Reproduce

TBD

## Context

This has been introduced by v2.0.0.

## Severity

Still not clear. We need to find out if changing the order cycle or any other attribute refreshes the cache. In the meantime, the shopfront won't reflect its actual state.

## Your Environment

* Version used: v2.0.0
* OFN Platform instance where you discovered the bug, and which version of the software they are using: Katuma

## Possible Fix

There must be some new model or attribute added in v2 and involved in a shopfront that doesn't trigger a cache refresh while it should.

[...]

PR #3838, opened 2019-05-09
Limit cache settings to 5 results

sauloperez:
#### What? Why?

Related to  #3830

Helps identify the actual difference between the products cache and the latest shopfront state using `/admin/cache_settings/edit`. From the issue:

> Additionally, as explained in https://github.com/openfoodfoundation/openfoodnetwork/wiki/Products-cache#development, there is a settings page to debug cache's staleness but it currently crashes in Katuma for unknown reasons. 

Without this is very hard to know what model we might not be refreshing. See #3830 for details.

#### What should we test?

The `/admin/cache_settings/edit` should only list 3 cacheable exchanges.

#### Release notes

Limit the cache_settings page results to only 3 exchanges for performance reasons. It can crash a server.

### DIFF
--- a/app/controllers/admin/cache_settings_controller.rb
+++ b/app/controllers/admin/cache_settings_controller.rb
@@ -3,7 +3,7 @@ require 'open_food_network/products_cache_integrity_checker'
 module Admin
   class CacheSettingsController < Spree::Admin::BaseController
     def edit
-      @results = Exchange.cachable.map do |exchange|
+      @results = Exchange.cachable.limit(5).map do |exchange|
         checker = OpenFoodNetwork::ProductsCacheIntegrityChecker
           .new(exchange.receiver, exchange.order_cycle)
 


sauloperez:
It turns out that the actual query although bad is probably not one of our worst offenders

```sql
openfoodnetwork=> explain analyze SELECT "exchanges".* FROM "exchanges" INNER JOIN "order_cycles" ON "order_cycles"."id" = "exchanges"."order_cycle_id" WHERE "exchanges"."incoming" = 'f' AND (orders_open_at IS NOT NULL AND orders_close_at IS NOT NULL) AND (order_cycles.orders_close_at > '2019-05-09 21:54:26.511515' OR order_cycles.orders_close_at IS NULL);
                                                                                               QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Hash Join  (cost=12.93..46.48 rows=22 width=90) (actual time=0.815..1.104 rows=19 loops=1)
   Hash Cond: (exchanges.order_cycle_id = order_cycles.id)
   ->  Seq Scan on exchanges  (cost=0.00..31.57 rows=470 width=90) (actual time=0.013..0.548 rows=450 loops=1)
         Filter: (NOT incoming)
         Rows Removed by Filter: 684
   ->  Hash  (cost=12.65..12.65 rows=22 width=4) (actual time=0.368..0.368 rows=19 loops=1)
         Buckets: 1024  Batches: 1  Memory Usage: 9kB
         ->  Seq Scan on order_cycles  (cost=0.00..12.65 rows=22 width=4) (actual time=0.166..0.330 rows=19 loops=1)
               Filter: ((orders_open_at IS NOT NULL) AND (orders_close_at IS NOT NULL) AND ((orders_close_at > '2019-05-09 21:54:26.511515'::timestamp without time zone) OR (orders_close_at IS NULL)))
               Rows Removed by Filter: 442
 Planning time: 2.281 ms
 Execution time: 1.292 ms
(12 rows)
```

app/controllers/admin/cache_settings_controller.rb
```diff
@@ -3,7 +3,7 @@
 module Admin
   class CacheSettingsController < Spree::Admin::BaseController
     def edit
-      @results = Exchange.cachable.map do |exchange|
+      @results = Exchange.cachable.limit(5).map do |exchange|
```
luisramos0: what if the 6th exchange is the one with important diffs?

mkllnk: I'm wondering the same. I think the really slow thing on this page is running the `ProductsCacheIntegrityChecker` on every order cycle. To generate the diff it compiles the json from scratch. The best idea I have would be to just list the cached order cycles and link them to another view that shows only that one. But that's a new feature.

sauloperez:
Yes, sorry I missed the why.

To me, this page is not about seeing all of the stale cache entries. There's no point on that because we will never review them all. It is about seeing a little example of the things that are not going ok. This endpoint can't serve as an in-depth investigation of th e issue. 

You see things are wrong, then you log into the server and start digging.

luisramos0:
ok @sauloperez that makes sense but looks like you are limiting to 5 before filtering the broken ones, so if the first 5 are ok and the 6th is broken, you will not see that there is a problem...

sauloperez:
:hankey: that's right

luisramos0:
make it just show errors and show only 5 errors at max with for example an external error_counter

continue if checker.ok?
error_counter++
break if error_counter > 5

mkllnk:
But that still means that the page will be really slow if all cache entries are okay. I think that this feature is just broken for production environments at the moment. We don't have many order cycles in staging and development and can show the whole result there. Can we do pagination?

sauloperez:
This is still problematic but got over the original problem of not knowing the diff with https://github.com/openfoodfoundation/openfoodnetwork/pull/3839. This can still cause downtime but IMO it's not as urgent as other current priorities. I move it back to all the things but we should assess its priority.

luisramos0:
let's see how the v2 rollout goes, if cache becomes an issue again, we can get back to this one :+1:

PR #3839, opened 2019-05-09
Report cache diff in new tab

sauloperez:
#### What? Why?

Related to #3830

This will allow us to see the difference between the cache entry and the actual shopfront. Otherwise, there is no way to see what wasn't refreshed in the cache. 

Captura de pantalla 2019-05-10 a les 0 09 59

We'll see a new tab next to "job" ☝️ 

#### What should we test?

We should see a new tab in these Bugsnag notifications. Since I don't know how to reproduce it yet I'll just trigger the notification from the rails console.

#### Release notes
Add a new tab with the cache diff for products_cache_integrity_checker_job.rb errors in Bugsnag.

Changelog Category: Added

### DIFF
--- a/app/jobs/products_cache_integrity_checker_job.rb
+++ b/app/jobs/products_cache_integrity_checker_job.rb
@@ -3,7 +3,14 @@ require 'open_food_network/products_cache_integrity_checker'
 ProductsCacheIntegrityCheckerJob = Struct.new(:distributor_id, :order_cycle_id) do
   def perform
     unless checker.ok?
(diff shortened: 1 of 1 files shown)


mkllnk:
How do we trigger a cache diff on staging to see this tab?

Btw: Code Climate complains.

sauloperez:
Tested and working. This is what we see

Screenshot from 2019-05-10 14-37-19

luisramos0:
I dont think this is in katuma staging any longer :-)

Issue #3727, opened 2019-04-11
[Subs] Intermittent Failures on Payments


lin-d-hop:
## Description

We are getting weekly reports of payments failing in subscriptions.
This is not happening on all customers - generally one or two customers per week. A number of enterprises have reported this issue but on UK Prod it seems to be affecting Farmham Community Farm the most.

## Expected Behavior

Subscription payments should complete as expected without error

## Actual Behaviour

Occasional payment failures with the error:
Errors: You cannot use a Stripe token more than once

## Steps to Reproduce

I don't yet know how to reproduce this error, but chat to me on Slack and I'll share details of specific customers to help with searching the logs.

## Animated Gif/Screenshot

![Screenshot from 2019-04-11 10-45-04](https://user-images.githubusercontent.com/8669848/55947746-fc304a00-5c46-11e9-9eea-b0a012bceb0a.png)

## Context

This means that the hub needs to take payment for the subscription outside of OFN then manually update OFN to say that payment has in fact been received. This is a major headache for the hub.

## Severity

S2
## Your Environment

* Version used: v1.29
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): Desktop
* OFN Platform instance where you discovered the bug, and which version of the software they are using. UK Production

lin-d-hop:
![](https://github.trello.services/images/mini-trello-icon.png) [Failing Subs](https://trello.com/c/eBNWazqb/426-failing-subs)

lin-d-hop:
User asking about this again.

mkllnk:
I looked through the code and the token comes from Stripe. Each time we want to take a payment, we ask Stripe to generate a token for us and then we use that token for the purchase transaction. From the code, I can't see how the same token can be re-used twice. It could be a bug in the Stripe gem.

@lin-d-hop Do you know when this error first happened? I'm wondering if a new gem version could have a bug.

mkllnk:
I just tested the token method in development mode and I get different tokens:
```
Stripe::Token.create({card: card.gateway_payment_profile_id, customer: card.gateway_customer_profile_id}, stripe_account: "acct_1Exxx")
=> #<Stripe::Token:0x2ac68b43f798 id=tok_1EQUaVKbre...
=> #<Stripe::Token:0x2ac68b5788e4 id=tok_1EQUZx...
=> #<Stripe::Token:0x2ac68b6a6c48 id=tok_1EQUZqKb...
```

Oh, I just found this line:
[...]

PR #3840, opened 2019-05-10
Make only reusable credit cards the default card

mkllnk:
#### What? Why?

Hopefully fixes https://github.com/openfoodfoundation/openfoodnetwork/issues/3727. There may be other bugs causing the same issue.

When a user didn't have a default credit card and then checked out with
a credit card it became the default even when the user didn't intend to
store it. That lead to subscriptions trying to charge a one-time card
which fails.

Please note that subscriptions can still fail if customers remove their default credit card. But the error message should be much clearer now.

#### What should we test?

As admin:

- Make sure Stripe and subscriptions are enabled.

As enterprise:

- Make sure subscriptions are enabled.
- Create a customer.

As customer:

- Sign up.
- Go to the account settings and allow the enterprise to charge credit cards.
- If you have a default card there, delete it.
- Check out using a credit card (Stripe) without ticking the box to save it.
- Go to account settings and there should still be no default card.

As enterprise:

- Try to create a subscription for the customer. It should not be possible with Stripe.

As customer:

- Add a new card and it should automatically become the default.

As enterprise:

- Create a subscription with Stripe. It should now be possible.

#### Release notes

Fixed a bug that prevented subscriptions from charging customers that didn't set their default credit card.

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/credit_card_decorator.rb
+++ b/app/models/spree/credit_card_decorator.rb
@@ -31,13 +31,17 @@ Spree::CreditCard.class_eval do
 
   private
 
+  def reusable?
+    gateway_customer_profile_id.present?
+  end
+
   def default_missing?
     !user.credit_cards.exists?(is_default: true)
   end
 
   def ensure_single_default_card
     return unless user
-    return unless is_default? || default_missing?
+    return unless is_default? || (reusable? && default_missing?)
     user.credit_cards.update_all(['is_default=(id=?)', id])
     self.is_default = true
   end
--- a/spec/models/spree/credit_card_spec.rb
+++ b/spec/models/spree/credit_card_spec.rb
@@ -4,15 +4,27 @@ module Spree
   describe CreditCard do
     describe "setting default credit card for a user" do
       let(:user) { create(:user) }
+      let(:onetime_card_attrs) do
+        {user: user, gateway_payment_profile_id: "tok_1EY..."}
+      end
+      let(:stored_card_attrs) do
+        {
+          user: user,
+          gateway_customer_profile_id: "cus_F2T...",
+          gateway_payment_profile_id: "card_1EY..."
+        }
+      end
+      let(:stored_default_card_attrs) do
+        stored_card_attrs.merge(is_default: true)
+      end
 
       context "when a card is already set as the default" do
-        let!(:card1) { create(:credit_card, user: user, is_default: true) }
+        let!(:card1) { create(:credit_card, stored_default_card_attrs) }
 
         context "and I create a new card" do
-          let(:attrs) { { user: user } }
-          let!(:card2) { create(:credit_card, attrs) }
-
           context "without specifying it as the default" do
+            let!(:card2) { create(:credit_card, stored_card_attrs) }
+
             it "keeps the existing default" do
               expect(card1.reload.is_default).to be true
               expect(card2.reload.is_default).to be false
@@ -20,7 +32,7 @@ module Spree
           end
 
           context "and I specify it as the default" do
-            let(:attrs) { { user: user, is_default: true } }
+            let!(:card2) { create(:credit_card, stored_default_card_attrs) }
 
             it "switches the default to the new card" do
               expect(card1.reload.is_default).to be false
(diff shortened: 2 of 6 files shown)


app/models/spree/credit_card_decorator.rb
```diff
@@ -31,13 +31,17 @@ def save_requested_by_customer?
...
+  end
+
   def default_missing?
     !user.credit_cards.exists?(is_default: true)
   end
 
   def ensure_single_default_card
     return unless user
-    return unless is_default? || default_missing?
+    return unless is_default? || (reusable? && default_missing?)
```
luisramos0: this condition is so short and still so brain damaging :-)
I am not sure how to make it better. extract method probably.

spec/models/spree/credit_card_spec.rb
```diff
@@ -57,23 +66,30 @@ module Spree
...
       context "when no card is currently set as the default for a user" do
         context "and I create a new card" do
-          let(:attrs) { { user: user } }
-          let!(:card1) { create(:credit_card, attrs) }
-
           context "without specifying it as the default" do
+            let!(:card1) { create(:credit_card, stored_card_attrs) }
+
             it "sets it as the default anyway" do
               expect(card1.reload.is_default).to be true
```
luisramos0: from the PR description I thought we dont want this behaviour anymore. when user creates a new card without specifying it as the default, it will not make it default because it can be a one time card...

lin-d-hop:
Hi @mkllnk 
Any news on this one? I think it can safely be downgraded but based on the lack of response from the (previously very insistent) user I think this might be a good fix.
If it is not far from completion then shall we just leave as s2 and push through?

lin-d-hop:
There is a workaround, so for correctness this should be an s3.
I'll update... but since there is a PR I maintain we should just push it through...

spec/models/spree/credit_card_spec.rb
mkllnk: Sorry for the confusion. There are two ways to create a card:

1. Enter card details during checkout. This creates a one-time card. It should never be the default because it can't be re-used.
2. Add a card in the profile. This always creates a re-usable card. You add it to the profile to store it long term, not just for one checkout. When you add the first card, it should automatically become the default. This saves a couple of clicks.

In this test we create a "stored card" that is re-usable. It should become the default.

mkllnk:
@luisramos0 I fixed the specs. Ready for re-review.

spec/models/spree/credit_card_spec.rb
luisramos0: ah, I get it. thanks.

luisramos0:
I moved this to top of code review column because it's almost an s2 an because it's an old PR.

luisramos0:
Rebased this PR to get latest updates, for example, possible cause of trouble #3884

lin-d-hop:
@mklink

Testing notes:
I followed your step by step. Once Stripe was enabled on the enterprise I got a 500 error when trying to checkout.
Thinking this was related to #3879 @luisramos0 rebased the PR. However after redeploying the error remained.
I tested about 19:10 if you want to check the logs for clues. And left the PR on FR staging if you want to have a play.
I suspect this is a v2 related issue as the error message in logs is the same as #3879. Thus it might not be related to this PR but definitely need to check!

mkllnk:
Hm, something is wrong with the staging data. When I copy it to my local machine, I can't even see the shop products. The question is just: Did we mess up the data with experimental pull request code or can we mess up the data with current master code?

mkllnk:
@lin-d-hop I could finally reproduce the error locally and get in with the debugger. The problem was that the shipping method couldn't be applied because of a missing zone.

Screenshot from 2019-06-12 11-03-19

I selected EU_VAT and now I can check out. Do we know who created the shipping method? :soccer: :arrow_forward:

lin-d-hop:
Hi @maikel,

Can I check a test case with you:
1) Customer checks out and saves the card
2) This card is not automatically made a default card
3) Customer changes this card to be a default card in the accounts section.
4) Enterprise can successfully make a subscription using this card now it is saved as a default card.

When a card is saved from an order, will this card be able to be used for a subscription? Or would the user need to re-enter the card?
Should it be possible for a saved card from an order to be used in a subscription?

Testing notes here:
https://docs.google.com/document/d/1bD4UTIJrlQAsvQ63dV8qoSs4nVCj2KoonHZr6f_Y5D8/edit#

lin-d-hop:
This PR is ready to go however it might not successfully close the issue.
Let me know on the above Maikel. I'll open a new issue to ensure that a user can not add make a card a default card incorrectly.

mkllnk:
@lin-d-hop I have an answer for you. A card that is saved during checkout is re-usable and should work for subscriptions. So we can close #3727.

I also found something else interesting. In your testing you found that storing cards during checkout doesn't make them default cards. That used to be different before this PR. The intention was always that a new stored card becomes the default if there isn't a default already. That saves a subscription customer one step. They just have to authorize the enterprise and don't need to select their default card. It also makes the card pre-selected during the next checkout.

This little bug happens because the card is first stored as one-time card and then made re-usable. I have no idea why it's not stored as re-usable card straight away. But I can probably fix it with a one-line change. :crossed_fingers:

PR #3841, opened 2019-05-10
Move shipping methods translations to lazy lookups

luisramos0:
#### What? Why?

Make all translations in the shipping methods admin view use lazy lookup translations instead of the spree root namespace.

#### What should we test?
I tested each key locally.
The admin shipping methods page both list, edit and create should be correctly translated.

#### Release notes
Changelog Category: Changed
Better organize code and translations related to the shipping methods page.

This view was copied from spree during the upgrade but the translation calls to Spree.t were left.


RachL:
List:
image

Edit:
image

Create:

Superadmin:

PR #3847, opened 2019-05-10
Allow unauthenticated access to OFN API endpoints

Matt-Yorkley:
#### What? Why?

I just retested the maps changes, and found the authentication on the new endpoint is really inconsistent. Most of the time it allows access when the user is logged out, but sometimes it denies the request with the error: "You must specify an API key". 

api_error

Logging in and logging out again seems to sometimes change the allow/deny behaviour from one state to the other, but I can't discern a pattern to it.

I've made a small change to the authentication:
- Allows API access without authentication, but only for OFN controllers which inherit from `BaseController`.
- Spree will now initialize `@current_api_user` with an empty `Spree::User` object (basically a "guest") if the current user is not logged in.
- This does not affect the authentication for existing Spree::Api endpoints.
- We should always use `authorize!` when required on API actions.

#### What should we test?

The new map page always load the modal when the user is logged out. Logging in and out doesn't change the outcome.

#### Release notes

Changed API authentication for custom endpoints

Changelog Category: Changed

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -9,6 +9,7 @@ module Api
     include ActionController::UrlFor
     include Rails.application.routes.url_helpers
     use_renderers :json
+    check_authorization
 
     def respond_with_conflict(json_hash)
       render json: json_hash, status: :conflict
@@ -20,6 +21,14 @@ module Api
     def authenticate_user
       @current_api_user = try_spree_current_user
       super
+    end
+
+    # Allows API access without authentication, but only for OFN controllers which inherit
+    # from Api::BaseController. @current_api_user will now initialize an empty Spree::User
+    # unless one is present. We now also apply devise's `check_authorization`. See here for
+    # details: https://github.com/CanCanCommunity/cancancan/wiki/Ensure-Authorization
+    def requires_authentication?
+      false
     end
   end
 end
--- a/app/controllers/api/enterprises_controller.rb
+++ b/app/controllers/api/enterprises_controller.rb
@@ -1,11 +1,11 @@
 module Api
   class EnterprisesController < BaseController
-
     before_filter :override_owner, only: [:create, :update]
     before_filter :check_type, only: :update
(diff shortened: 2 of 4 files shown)


Matt-Yorkley:
@luisramos0 what do you think?

app/controllers/api/base_controller.rb
```diff
@@ -21,5 +21,13 @@ def authenticate_user
       @current_api_user = try_spree_current_user
       super
     end
+
+    # Allows API access without authentication, but only for OFN controllers which inherit
+    # from BaseController. Actions that require authentication should all use #authorize!
```
luisramos0: maybe change to "from Api::BaseController (that inherits from Spree::Api::BaseController)"

app/controllers/api/enterprises_controller.rb
```diff
@@ -77,5 +77,12 @@ def override_sells
     def override_visible
       params[:enterprise][:visible] = false
     end
+
+    # Allows API access without a logged in user for actions in this controller.
+    # Actions that require authentication should all use #authorize!
+    # @current_api_user will now initialize an empty Spree::User unless one is present.
```
sauloperez: IMO it's worth explaining why we needed this and the tradeoff we made. That context will be invaluable the next time we end up here, don't you think?

kristinalim: @sauloperez @Matt-Yorkley updated the comment - are you already happy with it?

Matt-Yorkley:
I'm prepping a release, but I think we need to get this PR in, otherwise we'll have intermittent issues with the new map page...

lin-d-hop:
Looks good to me as per Matt's testing notes.
Given the name of this PR I feel like I'm under testing this by just checking maps.

Reviewers: @kristinalim @sauloperez @luisramos0 just wanted to confirm that you agree this testing is sufficient?

https://docs.google.com/document/d/1pQCXLjf3-gRrJ8DAkEVeMrEXQc_rgmLT2pBhtAa7hr0/edit#

Issue #3855, opened 2019-05-19
SQL Injection Vulnerability in users_and_enterprises_report


Drumstickx:
## Description
The admin interface contains a SQL injection vulnerability on the user and enterprises report page.
The `enterprise_id_in[]` and `users_id_in[]` parameters in requests issued to the respective endpoint are not properly sanitized.
This allows the execution of arbitrary SQL queries (e.g., dropping a table as shown below).

## Steps to Reproduce
1. Setup a new postgresql database (to prevent data loss) as described in `GETTINGS_STARTED.md`
2. Run openfoodnetwork with the new database (`bundle exec rails server`)
3. Visit the users and enterprises report section in the admin web interface at `http://0.0.0.0:3000/admin/reports/users_and_enterprises` in a browser (log in with the default credentials)
4. Extract the value of the cookie header and the authenticity token parameter as follows:
	- 4.1. Open the network tab of the developer console of your browser
	- 4.2. Issue a POST request by searching for any enterprise (e.g., `Mary's Online Shop`) using the form on the `http://0.0.0.0:3000/admin/reports/users_and_enterprises` site
	- 4.3. Inspect the issued request in the developer console and extract the cookie header value (e.g. `_session_id=27f5b5fbd88863d5afc0e445eed189b2`) and the authenticity token value (e.g., `gYhFyg4ud5kf77svOXGQd6ZnMJqFiEeBBVC49tb21iY=`)
5. Craft a malicious request with the values extracted in 4., (e.g. using the [Advanced Rest Client](https://chrome.google.com/webstore/detail/advanced-rest-client/hgmloofddffdnphfgcellkdfbfbjeloo)) by injecting a malicious SQL query via the `enterprise_id_in[]` parameter (works the same way for `user_id_in[]`)
    - Example for a malicious parameter value: `3)%3B DROP TABLE sessions%3B SELECT enterprises.id, enterprises.created_at FROM enterprises WHERE enterprises.id IN(3`
	- I attached two screenshots of an example request configuration with the Advanced Rest Client as well as an export of this request that can be imported in the Advanced Rest Client and used for reproduction (cookie header and authenticity token have to be updated with values of a current session)
6. Send the request issued in 5. to delete the sessions database

## Severity
Vulnerability

## Your Environment
* Version used: v2.0.1 / master branch as of 59d4e3d9e2192524d807b8dbc4271b427adea1ac
* Browser name and version: Chromium Version 74.0.3729.157
[...]

PR #3856, opened 2019-05-19
Fixed sql injection vulnerability in users_and_enterprises_report.

Drumstickx:
See  #3855

### DIFF
--- a/.rubocop_styleguide.yml
+++ b/.rubocop_styleguide.yml
@@ -196,7 +196,6 @@ Metrics/ClassLength:
   Max: 100
   Exclude:
   - engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
-  - lib/open_food_network/users_and_enterprises_report.rb
 
 Metrics/ModuleLength:
   Max: 100
--- a/.rubocop_styleguide.yml
+++ b/.rubocop_styleguide.yml
@@ -196,6 +196,7 @@ Metrics/ClassLength:
   Max: 100
   Exclude:
   - engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
+  - lib/open_food_network/users_and_enterprises_report.rb
 
 Metrics/ModuleLength:
   Max: 100
--- a/lib/open_food_network/users_and_enterprises_report.rb
+++ b/lib/open_food_network/users_and_enterprises_report.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
 module OpenFoodNetwork
   class UsersAndEnterprisesReport
     attr_reader :params
@@ -119,3 +120,4 @@ module OpenFoodNetwork
     end
   end
 end
+# rubocop:enable Metrics/ClassLength
--- a/lib/open_food_network/users_and_enterprises_report.rb
+++ b/lib/open_food_network/users_and_enterprises_report.rb
@@ -50,24 +50,7 @@ module OpenFoodNetwork
         query = query.where("owner.id IN (?)", params[:user_id_in].split(',').map(&:to_i))
       end
 
-      query.order("enterprises.created_at DESC")
-        .select(["enterprises.name",
-                 "enterprises.sells",
-                 "enterprises.visible",
-                 "enterprises.is_primary_producer",
-                 "enterprises.created_at",
-                 "owner.email AS user_email"])
-        .to_a
-        .map { |x|
-        {
-          name: x.name,
-          sells: x.sells,
-          visible: (x.visible ? 't' : 'f'),
-          is_primary_producer: (x.is_primary_producer ? 't' : 'f'),
-          created_at: x.created_at.utc.iso8601,
-          relationship_type: 'owns',
-          user_email: x.user_email
-        }.stringify_keys }
+      query_helper(query, :owner, :owns)
     end
 
     def managers_and_enterprises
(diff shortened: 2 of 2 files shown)


lib/open_food_network/users_and_enterprises_report.rb
```diff
@@ -38,29 +39,70 @@ def table
...
+
+      query.order("enterprises.created_at DESC")
+        .select(["enterprises.name",
+                 "enterprises.sells",
+                 "enterprises.visible",
+                 "enterprises.is_primary_producer",
+                 "enterprises.created_at",
+                 "managers.email AS user_email"])
+        .to_a
+        .map { |x|
```
kristinalim: This block is almost exactly the same as the one in `owners_and_enterprises`. Perhaps you can extract this into a separate method?

lib/open_food_network/users_and_enterprises_report.rb
```diff
@@ -38,29 +39,70 @@ def table
...
-      ActiveRecord::Base.connection.execute(query).to_a
+      query = Enterprise.joins("LEFT JOIN enterprise_roles ON enterprises.id = enterprise_roles.enterprise_id")
+        .joins("LEFT JOIN spree_users AS managers ON enterprise_roles.user_id = managers.id")
+        .where("enterprise_id IS NOT NULL")
+
+      if params[:enterprise_id_in].present?
+        query = query.where("enterprise_id IN (?)", params[:enterprise_id_in].split(',').map(&:to_i))
+      end
+
+      query.where("user_id IS NOT NULL")
```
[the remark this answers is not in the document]
Drumstickx: You are right, I somehow missed that one.

.rubocop_styleguide.yml
```diff
@@ -196,6 +196,7 @@ Metrics/ClassLength:
   Max: 100
   Exclude:
   - engines/order_management/app/services/order_management/reports/enterprise_fee_summary/scope.rb
+  - lib/open_food_network/users_and_enterprises_report.rb
```
luisramos0: we need to fix rubocop problems and add local exceptions if really needed.
this line should not be here.

lib/open_food_network/users_and_enterprises_report.rb
```diff
@@ -38,29 +38,59 @@ def table
...
+
+      if params[:enterprise_id_in].present?
+        query = query.where("enterprises.id IN (?)",
+                            params[:enterprise_id_in].split(',').map(&:to_i))
+      end
+
+      if params[:user_id_in].present?
+        query = query.where("owner.id IN (?)", params[:user_id_in].split(',').map(&:to_i))
+      end
+
```
luisramos0: repeated code here, should be in a separate method:

```
filter_by_int_list_if_present(params[:enterprise_id_in], "enterprises.id")
filter_by_int_list_if_present(params[:user_id_in], "owner.id")

def filter_by_int_list_if_present(int_list, filtered_field_name)
   query.where("? IN (?)", filtered_field_name, split_int_list(int_list)) if int_list.present?
end

def split_int_list(int_list)
    int_list.split(',').map(&:to_i)
end
```

lib/open_food_network/users_and_enterprises_report.rb
```diff
@@ -38,29 +38,59 @@ def table
...
+        .where("user_id IS NOT NULL")
+
+      if params[:enterprise_id_in].present?
+        query = query.where("enterprise_id IN (?)",
+                            params[:enterprise_id_in].split(',').map(&:to_i))
+      end
+
+      if params[:user_id_in].present?
+        query = query.where("user_id IN (?)", params[:user_id_in].split(',').map(&:to_i))
+      end
```
luisramos0: same as above

.rubocop_styleguide.yml
luisramos0: oh, I just saw @kristinalim comment.
now I am confused, I am not sure what's the current usage of the different rubocop exception files and when we should use local exceptions like [this](https://github.com/openfoodfoundation/openfoodnetwork/blob/59d4e3d9e2192524d807b8dbc4271b427adea1ac/lib/tasks/sample_data/enterprise_factory.rb#L20).

luisramos0: me and @mkllnk talked about this and this is what we have agreed:
https://community.openfoodnetwork.org/t/management-of-rubocop-exception-files/1668
maybe we can agree on this and make this PR follow the rule agreed.
our current proposed approach is to:
- add local exception for this problem
- or, if it needs to be fixed at some point, add to the manual_todo list

lib/open_food_network/users_and_enterprises_report.rb
```diff
@@ -38,35 +38,64 @@ def table
...
+          user_email: x.user_email
+        }.stringify_keys }
     end
 
     def users_and_enterprises
       sort( owners_and_enterprises.concat managers_and_enterprises )
     end
 
+    def filter_by_int_list_if_present(query, filtered_field_name, int_list)
+      if int_list.present?
```
mkllnk: This can be written more rubyish:

```rb
    def filter_by_int_list_if_present(query, filtered_field_name, int_list)
      if int_list.present?
        query.where("#{filtered_field_name} IN (?)", split_int_list(int_list))
      else
        query
      end
    end
```

.rubocop_styleguide.yml
kristinalim: Thanks for bringing this to my attention, @luisramos0! This sounds good to me - thanks.

RachL:
As an enterprise manager, using this report is working fine.

As a super admin, if you tried to load all the data in UK staging you get a direct error 500.

image

I'm not sure it is linked to UK staging or not. Can we stage this on AU or ES? Staging failed for me on ES and semaphore's automated staging process for AU is currently broken AFAIK. @sauloperez @mkllnk can one of you stage this manually on one of these two?

sauloperez:
On it to stage it in ES.

sauloperez:
Done @RachL

RachL:
Thank you @sauloperez ! All good, this is ready to go :)

PR #3863, opened 2019-05-21
Make inserting delayed jobs from cron quiet

mkllnk:
#### What? Why?

`config/schedule.rb` defines entries for cron jobs. Every five minutes
it insert several delayed jobs via `script/enqueue`. A recent upgrade of
Postgresql (a year or two ago) made that script noisy. Every time we
insert a delayed job with that script it outputs:

    INSERT 0 1

Standard output like this is seen as error message by cron an forwarded
to the unix user's local mailbox in `/var/mail/openfoodnetwork`. This
file grows over time and accumulates tens of thousands of error
messages, hiding any important failures.

This patch makes inserting delayed jobs quiet. A failure is still
reported.

#### What should we test?

- Stage the pull request.
- Wait half an hour.
- DelayedJob should still be reported running. You can check that, e.g.: https://staging.openfoodnetwork.org.au/api/status/job_queue

#### Release notes

Fixed error reporting of scheduled background tasks.

Changelog Category: Fixed

### DIFF
--- a/script/enqueue
+++ b/script/enqueue
@@ -26,6 +26,7 @@ def psql
   raise "Missing config for #{env} environment" unless config
 
   "psql".tap do |s|
+    s << " --quiet"
     s << " --host #{config['host']}"     if config['host']
     s << " --user #{config['username']}" if config['username']
     s << " --port #{config['port']}"     if config['port']


RachL:
It's alive!

image

PR #3865, opened 2019-05-22
Add Australian English locale

mkllnk:
#### What? Why?

Closes https://github.com/openfoodfoundation/ofn-install/issues/438.

The Australian instance is using the source locale so far. This commit
just copies the source locale. Adding en_AU to Transifex is pending.

#### What should we test?

This should be tested on the Australian staging server after the locale has been changed. The homepage should look normal.

#### Release notes

Added support to for Australian language customisations.

Changelog Category: Added

#### Dependencies

 The ofn-install config for Australia needs updating.

#### Documentation updates

The wiki page https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29 is up to date.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,16 +1,14 @@
 # English language file
 # ---------------------
 #
-# This is the source language file maintained by the global OFN team and used
-# by the Australian OFN instance.
-#
+# This is the source language file maintained by the global OFN team.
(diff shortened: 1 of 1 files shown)


sauloperez:
Yep, we should create an issue for it

luisramos0:
ok, done #3868

PR #3876, opened 2019-05-27
3804 Fix editing cart when variant on_demand but its on_hand value is zero or negative 

### LINKED ISSUE #3804: If insufficient stock on editing cart, quantity shown becomes incorrect
## Description
I've tried editing a cart and reaching product's max stock level. I received a correct message warning me that I couldn't go further. However after I pressed "ok" on the message" the quantity ordered went down to 0 on the UI. But actually it has still well kept the 7 units I was trying to buy. Updating the cart leads to an error 500.

## Steps to Reproduce
1. Go to https://staging.katuma.org/carrot-release/shop
2. Order 7 quantity of "Wine with stock at 7"
3. On the cart, try to add one quantity. A message pops up saying there is only one left.
4. Press "ok"
5. On your cart copy of wine is now showing with 0 quantity
6. Try to raise the number and get again the warning saying there is no stock left (so the 7 quantity is still somewhere but not shown through the UI
7. Try the update button and get an error 500

The workaround is to go back to the shop where the 7 quantity is showing correctly and go back to the cart

## Animated Gif/Screenshot
Cart with 7 on quantity:
image
Rasing to 8 gives this correct message:
image
After pressing ok:
image
when raising again:
image

Trying the update button:
image

## Severity

## Your Environment
* Version used: v2

luisramos0:
#### What? Why?

Closes #3804

Added test to cover bug and added a quick fix to it.

#### What should we test?
Test the bug in 3804 and other scenarios related to editing the quantity in the cart.

#### Release notes
Changelog Category: Fixed
Fix editing cart bug when variant on_demand but its on_hand value is zero or negative.

### DIFF
--- a/app/views/spree/orders/_line_item.html.haml
+++ b/app/views/spree/orders/_line_item.html.haml
@@ -22,7 +22,7 @@
   %td.text-right.cart-item-price{"data-hook" => "cart_item_price"}
     = line_item.single_display_amount_with_adjustments.to_html
   %td.text-center.cart-item-quantity{"data-hook" => "cart_item_quantity"}
-    = item_form.number_field :quantity, :min => 0, "ofn-on-hand" => variant.on_hand, "ng-model" => "line_item_#{line_item.id}", :class => "line_item_quantity", :size => 5
+    = item_form.number_field :quantity, :min => 0, "ofn-on-hand" => "#{variant.on_demand && 9999 || variant.on_hand}", "ng-model" => "line_item_#{line_item.id}", :class => "line_item_quantity", :size => 5
   %td.cart-item-total.text-right{"data-hook" => "cart_item_total"}
     = line_item.display_amount_with_adjustments.to_html unless line_item.quantity.nil?
 
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -126,7 +126,7 @@ feature "full-page cart", js: true do
       end
     end
 
-    describe "updating quantities with insufficient stock available" do
+    describe "updating quantities" do
       let(:li) { order.line_items(true).last }
       let(:variant) { product_with_tax.variants.first }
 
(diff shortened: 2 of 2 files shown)


app/views/spree/orders/_line_item.html.haml
```diff
@@ -22,7 +22,7 @@
   %td.text-right.cart-item-price{"data-hook" => "cart_item_price"}
     = line_item.single_display_amount_with_adjustments.to_html
   %td.text-center.cart-item-quantity{"data-hook" => "cart_item_quantity"}
-    = item_form.number_field :quantity, :min => 0, "ofn-on-hand" => variant.on_hand, "ng-model" => "line_item_#{line_item.id}", :class => "line_item_quantity", :size => 5
+    = item_form.number_field :quantity, :min => 0, "ofn-on-hand" => "#{variant.on_demand && 9999 || variant.on_hand}", "ng-model" => "line_item_#{line_item.id}", :class => "line_item_quantity", :size => 5
```
Matt-Yorkley: We're displaying the on hand count as: `variant.on_demand && 9999 || variant.on_hand` ?

luisramos0: no, this is not the display value, this is the on-hand value used to verify stock in JS. this is how it is done in JS when this ofn-on-hand directive is used...

mkllnk: Can we use `Infinity` instead of 9999?

luisramos0: hey @mkllnk we can, actually I went for it. this is the code change:
Screenshot 2019-06-13 at 16 26 07

BUT, this 9999 can actually be a good limit because if a crazy user tries to checkout 100k  items, the DB may crash with 100k inserts on spree_inventory_units and there will also be a number that it will make the code break somewhere for being too large...

I dont like the hard coded number 9999 at all but I think it's better than Infinity.
What do you think?

mkllnk: Good thinking. The result is good but it's a mix of concerns. I'm okay with leaving it. Ideally we would then name the 9999 something like `item_order_limit`. It's probably logic that shouldn't be hidden in a view. But I don't want to stop the current PR. We can just leave it and act on it if it comes up again.

luisramos0: yes, I agree. the good part of this is that it's so obviously wrong. yes, item_order_limit  should be in a helper or service.

RachL:
The scenario in #3804 is now working. I tested as well the update and bin button. While testing I've discover that I can't empty my cart when shopping at Mary's Online shop. I get an error 500. I didn't find out why, other shops are fine. This is not related to this PR, so I moving this to ready to go !

luisramos0:
I rebased to fix the conflicts. There were only conflicts in the auto tests so no need to manually test again. Merging.

PR #3842, opened 2019-05-10
Fix a few rubocop warnings manually and run rubocop autocorrect

luisramos0:
#### What? Why?

I improved Rubocop situation a little manually and then realized rubocop auto-correct exists!

Updated rubocop auto generated todo list by running [the command in the top of the file](https://github.com/openfoodfoundation/openfoodnetwork/blob/59d4e3d9e2192524d807b8dbc4271b427adea1ac/.rubocop_todo.yml#L2).

Also updated rubocop manual todo list following the process described in the file.

#### What should we test?
We can run a sanity check.

#### Release notes
Changelog Category: Changed
Improved code readability.

PR #3880, opened 2019-05-28
3792 Optimize saving Order Cycle changes

kristinalim:
#### What? Why?

- Relates to #3792

This PR makes the edit OC page add and remove exchange variants in bulk, instead of doing one SQL call for each affected record.

This is an experiment in using the `activerecord-import` gem, which does bulk SQL inserts. (It's my first time to use this gem.) A similar approach can be used when addressing #3760 whenever applicable.

#### What should we test?

Test removing, adding, and saving as is variants in the edit OC page.

#### Release notes

- Optimize saving exchange variants through the edit order cycle page. 

Changelog Category: Fixed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem 'rails', '~> 3.2.22'
 gem 'rails-i18n', '~> 3.0.0'
 gem 'rails_safe_tasks', '~> 1.0'
 
+gem "activerecord-import"
 # Patched version. See http://rubysec.com/advisories/CVE-2015-5312/.
 gem 'nokogiri', '>= 1.6.7.1'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -160,6 +160,8 @@ GEM
       activesupport (= 3.2.22.5)
       arel (~> 3.0.2)
       tzinfo (~> 0.3.29)
+    activerecord-import (1.0.1)
+      activerecord (>= 3.2)
     activeresource (3.2.22.5)
       activemodel (= 3.2.22.5)
       activesupport (= 3.2.22.5)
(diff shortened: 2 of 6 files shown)


Matt-Yorkley:
This looks promising! Maybe we could use this on the other bulk pages as well (after we add pagination).

What effect does it have on the saving time? Any before and after results from `miniprofiler`?

luisramos0:
ping @kristinalim #3842 created some conflicts here

kristinalim:
Thanks, @Matt-Yorkley and @luisramos0!

Here are some numbers.

In below tests, I did each set starting from the same DB snapshot with ~2000 variants. I tested only twice for Before Changes and twice for After Changes, and got around the same numbers.

This does not affect loading of the edit OC page.

In the sample scenario, there is:

* Around 69% improvement when adding ~2000 variants to an incoming exchange
* Around 74% improvement when removing ~2000 variants from an incoming exchange

### Before Changes

#### Adding ~2000 variants to incoming exchange

Duration | Path | SQL
-|-|-
5929 ms | /admin/order_cycles/2.json | 2149

#### Removing ~2000 variants from incoming exchange

Duration | Path | SQL
-|-|-
4260 ms | /admin/order_cycles/2.json | 134

### After Changes

#### Adding ~2000 variants to incoming exchange

Duration | Path | SQL
-|-|-
1866 ms | /admin/order_cycles/2.json | 150

#### Removing ~2000 variants from incoming exchange

Duration | Path | SQL
-|-|-
1127 ms | /admin/order_cycles/2.json | 145

lin-d-hop:
This all looks good to me. Changes in variants on Edit Order Cycle screen flow through to shopfront.
https://docs.google.com/document/d/1C-ocnJK8H6Dl8A1lPcYsilC9BDf5qpVkqizbT0Qu24E/edit#
Nice one!

Issue #2656, opened 2018-09-07
Add link and description to /admin/reports


kirstenalarsen:
![image](https://user-images.githubusercontent.com/2911176/45201813-13c2e100-b2ba-11e8-98ae-76fa26b7d1ef.png)

Issue #2616, opened 2018-08-30
Build Enterprise Fee Summary Report 


kirstenalarsen:
https://docs.google.com/spreadsheets/u/1/d/1pnJwHgBo2sokdcwNoLquYlVh8ICUcx6X4J2ZAMnRyio/edit?usp=sharing
Sheet "THIS By Customer"

Fee Type | Enterprise Owner | Fee Name | Customer | Fee Placement | Fee calc on trans through | Tax Category | $$ SUM
-- | -- | -- | -- | -- | -- | -- | --
Admin | Coord 1 |   | Customer A | Coordinator | All |   | 2.3
Admin | Coord 1 |   | Customer B | Coordinator | All |   | 4.7
Admin | Hub 2 |   | Customer A | Outgoing | Hub 2 |   |  
Admin | Hub 2 |   | Customer B | Outgoing | Hub 2 |   |  
Mgmt | Coord 1 |   | Customer A | Incoming | Supplier 2 |   |  
Mgmt | Coord 1 |   | Customer B | Incoming | Supplier 2 |   |  
Packing | Supplier 1 |   | Customer A | Incoming | Supplier 1 |   |  
Packing | Supplier 1 |   | Customer B | Incoming | Supplier 1 |   |  
Transport | Coord 1 |   | Customer A | Incoming | Supplier 2 |   |  
Transport | Coord 1 |   | Customer B | Incoming | Supplier 2 |   |  
Transport | Hub 3 |   | Customer A | Outgoing | Hub 3 |   |  
Transport | Hub 3 |   | Customer B | Outgoing | Hub 3 |   |  
Transport | Supplier 1 |   | Customer A | Incoming | Supplier 1 |   | etc
Transport | Supplier 1 |   | Customer B | Incoming | Supplier 1 |   | etc

kirstenalarsen:
@mkllnk @myriamboure - do we have a 'standard' view on whether reports pages should all load empty rather than slowly bringing up a zillion results?

and @mkllnk if you have any thoughts on performance / good building of reports based on the challenges you've seen in others feel free to put them here?!

mkllnk:
For good performance it's important to filter as much as possible on database level and avoid queries in loops. But I think Kristina knows that very well. ;-)

It could be very helpful to create a diagram of where the data comes from. In this case it looks like we start with orders, filter them and fetch the remaining data for the report. But I haven't looked into it very much. Happy to discuss approaches with Kristina. :-)

[...]

PR #3882, opened 2019-05-28
3408 Remove feature flag for Enterprise Fee Summary

### LINKED ISSUE #3408: [Enterprise Fee Summary] Enable report for non-superadmin enterprise users
Enable report for non-superadmin enterprise users.

## Description

This reverses the effect of #3332.

However, checks relating to permissions need to be done in doing this.

This is best to be tested with existing OCs (if already available) that would already have data for this report. Before testing, it would be helpful to identify hubs, producers, and coordinators that would have data needed for Step 2 below.

Each of the following roles would be checked for Steps 2 and 3.

1. Admin
2. Hub
3. Producer
4. Coordinator

To test:

1. Check that HTML and CSV report generation work. These are using the same data objects, so except occasionally, for Steps 2 and 3 there is no need to check results for both formats - just checking one should be sufficient.
2. Check that the "Enterprise Fee Summary" link is available in the Reports page, as described in #2656.
3. For each user role listed above, check that these data are properly summarized:
a. Coordinator fee for order cycle
b. Coordinator fee for incoming exchange
c. Coordinator fee for outgoing exchange
d. Producer fee for incoming exchange
e. Distributor fee for outgoing exchange
f. Shipping fee
g. Payment method fee
4. Do a smoke test that the correct filter options are available to each user role.
5. Check that orders for orders cycles to which their enterprise is not participant are not available to the user.
6. Test that using each filter field works. (No need to do this for each user role.)

## Expected Behavior

This feature should be available to all enterprise users.

## Actual Behavior

This feature is only available to superadmins before this.

## Context

For full launch of Enterprise Fee Summary report.

kristinalim:
#### What? Why?

- Closes #3408

This mostly reverses/updates the changes done in #3334.

#### What should we test?

This is best to be tested with existing OCs (if already available) that would already have data for this report. Before testing, it would be helpful to identify hubs, producers, and coordinators that would have data needed for Step 2 below.

Each of the following roles would be checked for Steps 2 and 3.

1. Admin
2. Hub
3. Producer
4. Coordinator

To test:

1. Check that HTML and CSV report generation work. These are using the same data objects, so except occasionally, for Steps 2 and 3 there is no need to check results for both formats - just checking one should be sufficient.
2. Check that the "Enterprise Fee Summary" link is available in the Reports page, as described in #2656.
3. For each user role listed above, check that these data are properly summarized:
a. Coordinator fee for order cycle
b. Coordinator fee for incoming exchange
c. Coordinator fee for outgoing exchange
d. Producer fee for incoming exchange
e. Distributor fee for outgoing exchange
f. Shipping fee
g. Payment method fee
4. Do a smoke test that the correct filter options are available to each user role.
5. Check that orders for orders cycles to which their enterprise is not participant are not available to the user.
6. Test that using each filter field works. (No need to do this for each user role.)

#### Release notes

- Enable Enterprise Fee Summary report for non-superadmin enterprise users.

Changelog Category: Changed

### DIFF
--- a/app/models/feature_flags.rb
+++ b/app/models/feature_flags.rb
@@ -14,13 +14,6 @@ class FeatureFlags
     user.superadmin?
   end
 
-  # Checks whether the "Enterprise Fee Summary" is enabled for the specified user
-  #
-  # @return [Boolean]
-  def enterprise_fee_summary_enabled?
-    user.superadmin?
-  end
-
   private
 
   attr_reader :user
--- a/app/models/spree/ability_decorator.rb
+++ b/app/models/spree/ability_decorator.rb
@@ -188,7 +188,7 @@ class AbilityDecorator
     can [:admin, :index, :customers, :orders_and_distributors, :group_buys, :bulk_coop, :payments,
          :orders_and_fulfillment, :products_and_inventory, :order_cycle_management, :packing],
         Spree::Admin::ReportsController
-    add_enterprise_fee_summary_abilities(user)
+    add_enterprise_fee_summary_abilities
   end
 
   def add_order_cycle_management_abilities(user)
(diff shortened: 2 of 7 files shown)


kristinalim:
Thanks, @luisramos0 and @sauloperez!

I addressed the Rubocop issue here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3882/commits/2c52c1a6fc7a9de9dfbc87fdb351537b8f0e4b80

Also linking to [this discussion](https://github.com/openfoodfoundation/openfoodnetwork/issues/2616#issuecomment-428073918) about data permissions for the report. In summary, all parties in the OC can see all fees in the OC.

Moving this to Test Ready.

RachL:
Looking good :)

I think we are ready to go! https://docs.google.com/document/d/1UBCJWXomkrZCgV0mlJgevwdycbfFTv4A-E6ZBHw93D4/edit#

PR #3884, opened 2019-05-29
Remove matching of product shipping category with shipping method's category on checkout

### LINKED ISSUE #3879: [v2] Cannot checkout if a product shipping category is not in any of the shipping methods
## Description
Shipping categories are now mandatory on products and shipping methods.

But you can create a product with a shipping category that was not selected in the shipping methods you have created. When this happens, the hub manager does not see his mistake, but his buyers won't be able to checkout that product. 

## Steps to Reproduce
1. Log is as an instance manager.
2. Check that you shipping methods are all missing one shipping category
3. Create a product with that particular shipping category
4. Put this product in an open OC and try to checkout that product
5. At the last step of checkout, the payment will show a spinner, but nothing else happens. You cannot go further (error 500)

## Severity

## Your Environment
* Version used: v2

## Possible Fix
Several options:
A - Make checkout succeed whatever the shipping categories selected or not selected
B - When creating a shipping method make all the categories and zones selected by default
C - When creating a product, show in the shipping categories dropdown only the shipping categories selected in the shipping methods.

luisramos0:
#### What? Why?

Closes #3879
Users can now checkout products with shipping categories not supported by the shipping methods. 
To make this verification work properly it would require some work in the backoffice, this is a quick solution to the problem where checkout would not work and hub managers would have no way to know why.

#### What should we test?
Verify issue is fixed.

#### Release notes
Changelog Category: Changed
Users can now checkout products with shipping categories not supported by the shipping methods. The shipping categories are not verified at checkout time.

This problem was introduced by the upgrade.

### DIFF
--- a/app/models/stock/package.rb
+++ b/app/models/stock/package.rb
@@ -10,6 +10,16 @@
 #
 module Stock
   class Package < Spree::Stock::Package
+    # Returns all exsiting shipping categories.
+    #   It does not filter by the shipping categories of the products in the order.
+    #   It allows checkout of products with categories that are not the shipping methods categories
+    #   It disables the matching of product shipping category with shipping method's category
+    #
+    # @return [Array<Spree::ShippingCategory>]
+    def shipping_categories
+      Spree::ShippingCategory.all
+    end
+
     # Skips the methods that are not used by the order's distributor
     #
     # @return [Array<Spree::ShippingMethod>]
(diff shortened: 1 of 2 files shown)


RachL:
I could checkout a product with a frozen shipping category although my enterprise only added default and refrigerated as shipping categories on shipping methods :tada:

This is ready to go!
Notes: https://docs.google.com/document/d/17ykpXTXFwdHVzp4Qc0DfVopJ5hQkRXpG2zUgeOinW_M/edit#

Issue #3642, opened 2019-03-22
[i18n] "cancelled" for payment status is badly automatically translated by spree


myriamboure:
## Description
On order / edit an order / payments, the status "cancelled" (when a payment is cancelled) is badly automatically translated by Spree in French, by "empty" ("vide" in French). The string is not in transifex so I can't do anything.

## Animated Gif/Screenshot
![image](https://user-images.githubusercontent.com/8135376/54837480-cd1d5d00-4cc6-11e9-9302-7d6592376f72.png)
(in the red rectangle is the bad translation)

## Context
Reported by a user who got confused.

## Severity
S3

luisramos0:
I think this has been fixed in v2. moving to test ready to be tested.

RachL:
Indeed! 

![image](https://user-images.githubusercontent.com/6525576/58793438-6685bb00-85f6-11e9-98f4-d88ad1a8932c.png)


Closing

PR #3888, opened 2019-05-30
Add translation keys to actions in the order payments list page

luisramos0:
#### What? Why?

Related to #3642
Investigating 3642 I found the actions in the order payments list page were not translated. This PR fixes that.

#### What should we test?
Order payments list page, actions capture and void should be translatable now.

#### Release notes
Changelog Category: Fixed
Actions in the order payments list page are now translatable.


RachL:
Seems good to go!

image

image

PR #3890, opened 2019-05-30
Add missing translation keys to the create product page

### LINKED ISSUE #1467: Rephrase error message in create new product
A user said they found this error message confusing... The term 'variant' is not very meaningful to users, so perhaps we should not use it in customer facing messages.
image

luisramos0:
Fixing the fields names displayed in error messages.

#### What? Why?

Closes #1467

Adds missing translations keys, not just the one in the issue but all the mandatory fields in the page.

#### What should we test?
Go to the new product page and click "create" without fillin in any fields. You should see the error message for all fields.

Additionally, there's an extra field called "unit name" that shows up when "unit size" is "items": go to new product page, select only unit size items and see the new field name translated. After that click "create" without filling in any other field, the error message should show "variant unit name" which is now a translatable string.

#### Release notes
Changelog Category: Fixed
Errors in the create product page are now translatable.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -33,7 +33,11 @@ en:
       spree/payment:
         amount: Amount
       spree/product:
-        primary_taxon: Product Category
+        primary_taxon: "Product Category"
+        supplier: "Supplier"
+        shipping_category_id: "Shipping Category"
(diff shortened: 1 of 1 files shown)


RachL:
Seems good to go as well!

image

image

image

Issue #3862, opened 2019-05-21
Product stays in the product lists even when out of stock


luisramos0:
## Description

Whenever a new order is placed and stock is updated the cache is not updated resulting on wrong stock displayed in the shopfront. Since the products cache is only used in the shopfront, all other parts of the app are not affected; they'll display correct stock level.

## Expected Behavior

The shopfront does not display products that are out of stock.

## Actual Behaviour

When you add an out of stock product into the cart, a modal pop ups telling it ran out of stock and then the product greys out forbidding you to purchase it.

## Steps to Reproduce

1. Disable on demand and add some stock to a variant
2. Place an order consuming all the stock available
3. Then, check the updated stock in the variant from the backoffice
4. Try placing a new order adding that same product

## Context

https://github.com/openfoodfoundation/openfoodnetwork/issues/3830#issuecomment-492970792

## Severity

bug-s3: a feature is broken but there is a workaround

## Your Environment

* Version used: v2.0.1
* OFN Platform instance where you discovered the bug: Katuma

## Possible Fix

Add the necessary cache refreshing logic for the new stock item model.

sauloperez:
Note to myself. We need to investigate if the following method needs updating:

https://github.com/openfoodfoundation/openfoodnetwork/blob/4afa0cdca1b66d42f32506f1d62d9fc54b15d33c/app/models/spree/variant_decorator.rb#L121-L136

PR #3891, opened 2019-05-31
Bring changes introduced in v2.0.1 into 2-0-stable

sauloperez:
#### What? Why?

As we want to prepare a v2.0.2 release with the fix for https://github.com/openfoodfoundation/openfoodnetwork/issues/3862 and https://github.com/openfoodfoundation/openfoodnetwork/issues/3830 in it, we need to update the branch `2-0-stable`, the one we'll use to iterate on v2, with the changes that we introduced in the [v2.0.1](https://github.com/openfoodfoundation/openfoodnetwork/releases/tag/v2.0.1) plus some previous commits. We should only keep this one branch and tidy things up before we commit any mistakes :see_no_evil: .

#### What should we test?

No manual tests as there is no change in behavior.


luisramos0:
hmmm, I dont think this is necessary @sauloperez 
we dont need to keep this branch. I'd delete 2-0-stable.
we can build the 2-1-stable branch from master when we start to upgrade to spree 2-1.

Currently 2-1-stable is an old 2-0-stable plus some stuff, I was planning to rebuild 2-1-stable again from master after some major PRs are merged to master (rails 4 and remove spree frontend).

So,
option 1 - keep merging master to 2-0-stable
option 2 - delete 2-0-stable and start fresh from master in 2-1-stable

I'd go option 2.

sauloperez:
My question is: which branch do we pick to continue v2 roll-out? To my understanding, we want to deploy only v2 plus any fixes we might need so no unnecessary risks are introduced. Or are we ok with deploying `master`-based releases to this v2 servers, which includes much more than the deep changes in db/schema and core features v2 brings? That's simpler in terms of branch management, of course.

luisramos0:
ah, ok, I understand, I was talking about spree 2.1 upgrade, you are talking about the 2.0 roll out and 2.0.x versions. sorry, my mistake.

I still think we should delete 2-0-stable and use a branch of master. In this case, we can use 2-0-1 or even create a new one from it 2-0-2 if you want.

Option 1 - create PRs for #3862 and #3830 against 2-0-1 and release 2-0-2 out of the 2-0-1 branch
Option 2 - copy branch 2-0-1 into new branch 2-0-2 and create PRs for #3862 and #3830 against that new branch 2-0-2 from where we release 2-0-2

sigmundpetersen:
So it's Ready to Go then, right?

sauloperez:
Awesome, team! Let's go with @kristinalim's suggestion :rocket:

PR #3892, opened 2019-05-31
Hide product from shopfront when out of stock

### LINKED ISSUE #3830: RuntimeError in app/jobs/products_cache_integrity_checker_job.rb:6

### LINKED ISSUE #3862: Product stays in the product lists even when out of stock

sauloperez:
#### What? Why?

Closes #3862 and closes #3830

This ensures the cache is up-to-date when a user accesses the shopfront and we use the appropriate stock levels.

Besides, this makes the (very annoying) ProductsCacheIntegrityCheckerJob succeed and thus Bugsnag won't be flooded with failures anymore.

As for the implementation of this. I'm not happy at all. It's disgusting to use `class_eval` for this but it's a good trade-off to deliver this fix ASAP. Katuma's been experiencing this error far too long see_no_evil. I was working on a better implementation that required a bit of refactoring but it was taking too much time. I'll open another PR with it soon.

#### What should we test?

Follow the steps listed by @luisramos0 in https://github.com/openfoodfoundation/openfoodnetwork/issues/3862.

We also need to check that no errors like https://github.com/openfoodfoundation/openfoodnetwork/issues/3830 show up in Bugsnag while testing in staging

#### Release notes

Out of stock variants do not appear in the shopfront anymore and any stock change is properly reflected in the products cache.


spec/models/spree/variant_spec.rb
```diff
@@ -166,7 +166,8 @@ module Spree
       let(:variant) { create(:variant) }
 
       it "refreshes the products cache on save" do
-        expect(OpenFoodNetwork::ProductsCache).to receive(:variant_changed).with(variant)
+        # When creating the variant both the Variant and StockItem callbacks get executed
```
sauloperez: I prefer to optimize these things in a separate PR. From my experience, you can end up in a black hole very easily when dealing with them and we need this fix ASAP.

What I have in mind is using something like

```
def refresh_products_cache
  unless variant.id_was == nil # at this point we know we're creating a new variant
    # refresh logic
  end
end
```

Or perhaps simply an `after_update` and an `after_destroy` instead of this broad `after_save`.

lin-d-hop:
Ok I'm testing this now... though it still needs another code review, right?

lin-d-hop:
Testing notes here:
https://docs.google.com/document/d/1Zn9kE9FpaZPvPppPFo_LLRa77R8aPtmtCZnhcU2IP4w/edit#

@sauloperez Back to you as either @Matt-Yorkley or @kristinalim still need to review before merging.

Issue #1803, opened 2017-09-05
Missing translations : calendar


myriamboure:
## Description
In order cycle page (but I guess it can happen in other calendar pages), the calendar needs to be translatable
![illustration](https://user-images.githubusercontent.com/8135376/30079271-f22e5b7e-927f-11e7-960d-05a81fef31be.jpeg)

## Step to reproduce
1. Log in as enterprise user or super admin
2. Go to the order cycles menu
3. Create a new order cycle or edit a current one
4. See the calendar dropdown
5. This dropdown also appears in the "reports" menu

## Severity
bug-s4: it's annoying, but you can use it

luisramos0:
There are two types of calendars: datepickers where user can select a date (in the orders management page for example) and datetime pickers where user can select a datetime (in OC list and OC edit pages for example). This issue referes to datetimepickers.

I managed to translate date pickers in #3893 but I cant fix the datetimepickers...

I think the fix to translate datetime pickers will be similar to the datepickers, but I have not managed to find the trick...
For datetime pickers we may need to add the localization files of the plugin (something I didnt have to do to fix the datepickers).

Also, I have created #3894 to remove duplicated code around these pickers... we have 5 implementations of calendars in the code...

rmklaus12:
This is a bugger, isn't it? Had a go at this, with experienced help, but no luck so far. I did dig up the syntax `.datetimepicker.setLocale('de');`, of which I tried multiple iterations with no luck. Will keep it assigned to me, as I want to keep working on it.

luisramos0:
:muscle: it's hard, good luck!
The code is here /home/luisramos/dev/ofn/vendor/assets/javascripts/jquery-ui-timepicker-addon.js
No .setLocale there, just .regional.

see localization tab here:
https://trentrichardson.com/examples/timepicker/

it says we have to add our own translations!
So the solution could be to do:
```

$.datepicker.regional['ru'] = {
	**closeText: t('js.timepicker.closeText'),
	prevText: t('js.timepicker.prevText'),
   etc... for all of these:**
	nextText: 'След>',
	currentText: 'Сегодня',
	monthNames: ['Январь','Февраль','Март','Апрель','Май','Июнь',
	'Июль','Август','Сентябрь','Октябрь','Ноябрь','Декабрь'],
	monthNamesShort: ['Янв','Фев','Мар','Апр','Май','Июн',
	'Июл','Авг','Сен','Окт','Ноя','Дек'],
[...]

PR #3893, opened 2019-05-31
WIP Translate date pickers (not datetime pickers)

luisramos0:
Make datepickers use selected locale by settging the regional property to empty. This makes the datepicker use the selected locale.

#### What? Why?

Relates to #1803

I have created #3894 to remove code duplication.

#### What should we test?
All datepickers in the backoffice should be translated now. I tested orders list and a couple of reports with pt and fr.
This will not work for all languages, I tested with fr_BE and it doesnt work, I am not sure why. Maybe we can create an issue for these languages.

#### Release notes
Changelog Category: Fixed
Date pickers are now translated (not datetime pickers yet).


luisramos0:
this solution is not valid. will close as I am not planning to work on this soon.

PR #3895, opened 2019-05-31
Make weight calculator use line_item.final_weight_volume rather than variant.weight

### LINKED ISSUE #3661: Shipping fee "per kg" are not updated when real weight is captured
## Description
As a manager I have created a product "cheese (kg)" with a variant 1kg and a variant 0.5kg (500g).
When my customers order, they can then choose if they want 3kg, 1,5kg, etc.
The customer will pay a transport fee which is per kg based.

When the pieces of cheese are received, there is always a gap between ordered and delivered, as you can't be precise when you cut a piece of cheese ! 
So the manager when giving the cheese to the customer adjust the real weight in BOM. The price is automatically adjusted.

BUT the shipping fee which is per kg based is not...
Even if I try to go into the order and click on "update and recalculate fees", nothing changes.

## Expected Behavior
If there is a per kg shipping fee, if the weight of the order is changed, the fee should change accordingly.

## Actual Behaviour
It doesn't.

## Steps to Reproduce
1. Make sure you have a product "per kg" and a shipipng method fee "per kg"
2. Order that product.
3. In BOM page, change the weight, see the price changing, and save.
4. Go to the order, you see the shipping fee is unchanged. You can try to update and recalculate fee, nothing change.

## Animated Gif/Screenshot
Product for the test:
Capture du 2019-03-27 15-13-32
Order with 10ct shipping fee per kg
Capture du 2019-03-27 15-14-09
In BOM before changing volume:
Capture du 2019-03-27 15-14-54
After:
Capture du 2019-03-27 15-15-11
Unchanged shipping fee:
Capture du 2019-03-27 15-15-59

## Context
Reported by a user.

## Severity
The workaround is to capture a manual adjustment for every single order, which isn't an acceptable workaround. Only one user is concerned in France for now so I think it's reasonable to keep the severity to s3 and see with that user how they can handle...

## Your Environment

* Version used: 1.29
* OFN Platform instance where you discovered the bug, and which version of the software they are using. French production

## Possible fix
Calculate on g based and not kg based the fee ?

luisramos0:
This is for cases where the final weight is set manually in the BOM

#### What? Why?

Closes #3661 

Re 3661, the fees are being recalculated, the problem is that the manual weight was not being picked up.
The weight calculator for fees was always using the variant weight, the fixed value. When the final weight is set by the seller on the BOM, that value is stored in the line_item.final_weight_volume field, the weight calculator is now using that value if it is present.

#### What should we test?
We need to verify all cases where the weight calculator is used.

Make sure the issue is fixed and shipping fees are updated when final weight volume is updated in the BOM.
This should be tested for the simple case where variant unit is weight but also for variants where unit is items or volume: in these cases, the fees should be calculated based on the variant weight field that can be found when editing a variant even when unit is volume for example. The test is: update the weight of a volume based variant, add the variant to the order and make sure the weight based shipping fee is correctly calculated based on that weight, then, go to BOM and change the final weight volume and check the shipping fee is also correctly calculated. (I tested this and it worked in staging UK).

#### Release notes
Changelog Category: Fixed
Manual weight values introduced by the seller in the backoffice are now picked up when calculating fees by weight based calculators.

### DIFF
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -18,11 +18,16 @@ module Calculator
 
     def total_weight(line_items)
       line_items.sum do |line_item|
-        if line_item.final_weight_volume.present?
-          line_item.final_weight_volume / 1000
-        else
-          (line_item.variant.andand.weight || 0) * line_item.quantity
-        end
+        line_item_weight(line_item)
+      end
+    end
+
+    def line_item_weight(line_item)
+      if line_item.final_weight_volume.present?
+        # Divided by 1000 because grams is the base weight unit and the calculator price is per_kg
+        line_item.final_weight_volume / 1000
+      else
+        (line_item.variant.andand.weight || 0) * line_item.quantity
       end
     end
   end
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -13,7 +13,7 @@ module Calculator
 
     def compute(object)
       line_items = line_items_for object
-      total_weight(line_items) * preferred_per_kg
+      (total_weight(line_items) * preferred_per_kg).round(2)
     end
 
     private
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -13,7 +13,13 @@ module Calculator
 
     def compute(object)
       line_items = line_items_for object
-      total_weight = line_items.sum { |li| ((li.variant.andand.weight || 0) * li.quantity) }
+      total_weight = line_items.sum do |line_item|
+        if line_item.final_weight_volume.present?
+          line_item.final_weight_volume / 1000
+        else
+          ((line_item.variant.andand.weight || 0) * line_item.quantity)
+        end
+      end
       total_weight * preferred_per_kg
     end
   end
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -16,6 +16,8 @@ module Calculator
       total_weight(line_items) * preferred_per_kg
     end
 
+    private
+
     def total_weight(line_items)
       line_items.sum do |line_item|
         line_item_weight(line_item)
@@ -23,6 +25,7 @@ module Calculator
     end
 
     def line_item_weight(line_item)
+      return 0 if line_item.variant.product.andand.variant_unit != 'weight'
       if line_item.final_weight_volume.present?
         # Divided by 1000 because grams is the base weight unit and the calculator price is per_kg
         line_item.final_weight_volume / 1000
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -13,14 +13,17 @@ module Calculator
 
     def compute(object)
       line_items = line_items_for object
-      total_weight = line_items.sum do |line_item|
+      total_weight(line_items) * preferred_per_kg
+    end
+
+    def total_weight(line_items)
+      line_items.sum do |line_item|
         if line_item.final_weight_volume.present?
           line_item.final_weight_volume / 1000
         else
-          ((line_item.variant.andand.weight || 0) * line_item.quantity)
+          (line_item.variant.andand.weight || 0) * line_item.quantity
         end
       end
-      total_weight * preferred_per_kg
     end
   end
 end
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -45,6 +45,9 @@ module Calculator
       end
     end
 
+    #  Example: 2 (line_item.quantity) wine glasses of 125mL (line_item.variant.unit_value)
+    #    Customer ends up getting 350mL (line_item.final_weight_volume) of wine
+    #      that represent 2.8 (quantity_implied_in_final_weight_volume) glasses of wine
     def quantity_implied_in_final_weight_volume(line_item)
       (1.0 * line_item.final_weight_volume / line_item.variant.unit_value).round(3)
     end
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -26,13 +26,7 @@ module Calculator
 
     def line_item_weight(line_item)
       if line_item.final_weight_volume.present?
-        if line_item.variant.product.andand.variant_unit == 'weight'
-          # Divided by 1000 because grams is the base weight unit and the calculator price is per_kg
-          line_item.final_weight_volume / 1000.0
-        else
-          final_units = (1.0 * line_item.final_weight_volume / line_item.variant.unit_value).round(3)
-          weight_per_variant(line_item) * final_units
-        end
+        weight_per_final_weight_volume(line_item)
       else
         weight_per_variant(line_item) * line_item.quantity
       end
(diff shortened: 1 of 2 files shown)


luisramos0:
ahaha, yes! it feels weird I am not just fixing specs in the 2-0-stable branch 🤣

luisramos0:
interesting Kristina.
the original weight calculator uses variant.weight, this field is not updated for volumes and items.
so, the calculator as is will return zero for volumes and items.
I think we can keep that behaviour, so I am adding a clause to the method for that.

kristinalim:
@luisramos0 I think [this line](https://github.com/openfoodfoundation/openfoodnetwork/blob/a783f4609a1e1485bda25877ce8997c4a9b8e647/app/models/spree/variant_decorator.rb#L116) updating `Spree::Variant#weight` is only to sync the weight of the variant with its unit value when the unit type is weight? When the unit type is not weight, the user can specify the weight at the variant form (not Bulk Edit Product page).

luisramos0:
ah, nice one Kristina, I didnt see that part of this puzzle 🙄 
Following from that, I see that although variant.weight can be set (for products in volume or items), when the order is created, the line_item.final_weight_volume (maybe that's why it's called _volume) is not populated from that weight field but from the unit_value (which is the volume when product is on volume). [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/a783f4609a1e1485bda25877ce8997c4a9b8e647/app/models/spree/line_item_decorator.rb#L167)

I think it's better to leave this logic above as is with volume populating final_weight_volume.

So, I think the current code in this PR is ok as is. Do you see a better alternative?

app/models/calculator/weight.rb
```diff
@@ -13,8 +13,25 @@ def self.description
...
+    private
+
+    def total_weight(line_items)
+      line_items.sum do |line_item|
+        line_item_weight(line_item)
+      end
+    end
+
+    def line_item_weight(line_item)
+      return 0 if line_item.variant.product.andand.variant_unit != 'weight'
```
luisramos0: hi @kristinalim from slack you said:
"I think it's not valid to consider 0 weight for items that are not using weight as the unit type."

can you add more details? maybe I can investigate further.

kristinalim: @luisramos0 The use case I was thinking of is this: You have a 1L variant, which may or may not be 1kg in weight. The shop supports weight-based shipping, so they set the "Weight" in the variant form to 1.2kg.

With this guard clause, this line item would have 0 weight when calculating shipping, which should not be the case.

This is what I see the calculation would look like:

```
actual_quantity_of_variant = 1.0 * final_units_per_variant / total_units_per_variant
actual_kg_per_variant = actual_quantity_of_variant * kg_per_variant
```

luisramos0: sorry @kristinalim I am not sure I follow.
can you clarify what these values are? final_units_per_variant and total_units_per_variant
are you not repeating what's already done in line_item.calculate_final_weight_volume and line_item.unit_value?

I understand that we need to calculate weight even if line_item.variant.product.andand.variant_unit != 'weight'

So, I think that if variant_unit is volume or items but final_weight_volume is not present, we can still just use 
line_item.variant.andand.weight.
so, the only condition to change would be when variant_unit != weight and final_weight_volume.present?

something like this:
```suggestion
      if line_item.final_weight_volume.present?
            if line_item.variant.product.andand.variant_unit == 'weight'
                  # Divided by 1000 because grams is the base weight unit and the calculator price is per_kg
                  line_item.final_weight_volume / 1000
            else
                  line_item.final_weight_volume
            end
      else
        (line_item.variant.andand.weight || 0) * line_item.quantity
      end
```

what do you think?

kristinalim: Sorry @luisramos0 if there is something I do not understand correctly or I missed something.

Let us use these examples for a 3 L variant with 2.5 kg weight:

* Variant `unit_value`: 3 (L)
* Product `variant_unit`: volume
* Product `variant_unit_scale`: 1.0
* Variant weight: 2.5 (kg)

And weight-based shipping costing $6 per kg.

If I order 7 units of the variant, the shipping fee should be $105:

```
7 pc * (2.5 kg / pc) * ($6 / kg)
# => 105
```

This order sets "Final Weight/Volume" to 21. If the admin changes this value to 22 (as in 22 L), the code above would change recalculate the fee to $132:

```
22 L * ($6 / kg)
```

When it should be $110:

```
22 L * (2.5 kg / 3 L) * ($6 / kg)
```

I will post another comment with suggested code change.

kristinalim: I wanted to verify the logic, and ended up with a full commit. :joy:

@luisramos0 Please take a look here: https://github.com/kristinalim/openfoodnetwork/commit/e1fce8304d1c4cc5da1c5708e0a79c8e3bc449d4 I added specs testing how `final_weight_volume` is handled for "weight", "volume", and "items", and have updated the calculator as well. Let me know if that makes sense.

Because of the division performed for "volume" and "items" variants, I also added `.round(3)` to the fee calculation: https://github.com/kristinalim/openfoodnetwork/commit/0d6ba90ea1b5084b9efc087dc851833826354034

RachL:
@luisramos0 This PR is since early June in Code review. May I asked what's left? Is it blocked by anything?

luisramos0:
This is not blocked since last week. I now need to validate and incorporate Kristina's solution 🏅 
I will do that this week.

luisramos0:
ok, I have integrated Kristina's changes and separated the code in smaller methods.

@kristinalim can you please approve and move to test ready? thank you!

app/models/calculator/weight.rb
```diff
@@ -26,13 +26,7 @@ def total_weight(line_items)
...
     def line_item_weight(line_item)
       if line_item.final_weight_volume.present?
-        if line_item.variant.product.andand.variant_unit == 'weight'
-          # Divided by 1000 because grams is the base weight unit and the calculator price is per_kg
-          line_item.final_weight_volume / 1000.0
-        else
-          final_units = (1.0 * line_item.final_weight_volume / line_item.variant.unit_value).round(3)
-          weight_per_variant(line_item) * final_units
-        end
+        weight_per_final_weight_volume(line_item)
```
kristinalim: Nice cleanup!

luisramos0:
I am going to temporarily remove this from katuma staging so I can investigate #4111

lin-d-hop:
When I tested this updated the weight of the order in Orders -> Edit Orders everything worked fine. I was able to update the weight and the variable weight Shipping Cost updated correctly.

I then tried to update the weight of the order within the BOM. The resulting weight update would not save. This has not been happening in other testing so I suspect that it might be related to this PR.

Screenshot from 2019-08-23 10-59-55


Back to In Dev.

luisramos0:
@lin-d-hop in the edit order page we can only edit quantity, is that what you mean by "updated the weight of the order in Orders -> Edit Orders everything worked fine. I was able to update the weight and the variable weight Shipping Cost updated correctly."

or am I missing something?

luisramos0:
re the BOM, the first test I did was to go to stg FR and do the test with whatever is there and I get the same error. so, probably not specific to this PR. I'lll investigate further now.

luisramos0:
I dont understand what's going on in staging FR... but it's happening without this PR, so not related.

I staged this in staging UK and it's looking good. I improved "What should we test?" above.

Ready for testing.

RachL:
I believe this one is good to go!

Testing notes: https://docs.google.com/document/d/1fT_3ZjfRmankxpmKHtACoJ00TQsz4JdXO1tz29HKUck/edit#

luisramos0:
tough one, 3 months... we did it 🎉 
this completes the original TOP 5 S3 bugs epic :-)

PR #3896, opened 2019-05-31
Clear product.variants.unit_description if product.variant_unit is items

### LINKED ISSUE #1582: Unit description not cleared when switching between kg and items
Picked up in #1452. When switching variant_unit between `weight` products and `items`, the unit description text on variants lingers. It should to be removed.

issue1
issue2

## Severity
S3, not affecting so many users I guess and there is a workaround.

luisramos0:
When product.variant_unit is items, the product.variant_unit_name is used instead of the variant.unit_description (this field is not available in the variants edit page if product.variant_unit is items). So, we need to clear it.

#### What? Why?

Closes #1582

#### What should we test?
The product edit page needs to be tested as well as verifying the issue if fixed.

#### Release notes
Changelog Category: Fixed
Unit description (only used for weight or volume products) is cleared when switching a product unit from weight or volume to items.

### DIFF
--- a/spec/controllers/spree/admin/products_controller_spec.rb
+++ b/spec/controllers/spree/admin/products_controller_spec.rb
@@ -150,16 +150,30 @@ describe Spree::Admin::ProductsController, type: :controller do
   end
 
   describe "updating" do
+    let(:producer) { create(:enterprise) }
+    let!(:product) { create(:simple_product, supplier: producer) }
+
+    before do
+      @request.env['HTTP_REFERER'] = 'http://test.com/'
+      login_as_enterprise_user [producer]
+    end
+
+    describe "product variant unit is items" do
+      it "clears unit description of all variants of the product" do
+        product.variants.first.update_attribute :unit_description, "grams"
+        spree_put :update,
+                  id: product,
+                  product: {
+                    variant_unit: "items",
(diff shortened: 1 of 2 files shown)


spec/controllers/spree/admin/products_controller_spec.rb
```diff
@@ -150,17 +150,31 @@
   end
 
   describe "updating" do
+    let(:producer) { create(:enterprise) }
+    let!(:product) { create(:simple_product, supplier: producer) }
+
+    before do
+      @request.env['HTTP_REFERER'] = 'http://test.com/'
```
Matt-Yorkley: Why is this needed here?

mkllnk: Yeah, it doesn't seem to belong here. :man_shrugging:

luisramos0: must have been a copy paste error... removed in new commit.

RachL:
So I tested with the follwoing super-real scenario:
1 bag of 1g of carrots:
image

After moving to item:
image

It also displays correctly in inventory. Editing other fields still work as well as checkout and BOM.
An old issue fixed!!!

Issue #2505, opened 2018-07-27
[Subscriptions][Stock] If I edit a subscription while open, I can go into negative stock levels


sstead:
## Description
I'm a shop manager and I have an open order cycle with open subscription orders attached. A product is on zero stock. I edit one of the open susbcriptions, to add this item to the subscription. The stock level goes into negative values, even though my instance is set to not allow backorders. The sub order will be confirmed with this product in it at the close of the OC.

## Expected Behavior
The item should be added to the susbcription, but there's insufficient stock, so the customer should get 0 QTY.

## Actual Behavior
Rather than the customer getting 0 QTY of the product that was added to their subscription, they get the amount that was added, but the stock level goes into -ve values.

## Steps to Reproduce
1. Setup a shedule OC with an open subsccription in it.
2. Set the on hand value of a product that's not in the susbcription to zero. 
3. Now edit this susbcription and add the product with zero stock to it.
4. See if the stock level goes to -ve.
5. You'll see the order get confirmed as though there was enough stock on hand, when really the order should have 0 QTY for that item.

## Animated Gif/Screenshot
https://www.useloom.com/share/b164f07ce7984eaf8cdd09c2c725eb1c

## Context
This could arise when subs is used broadly. We need to be confident that the stock control functionality isn't compromised because it's a central OFN feature.

## Severity
s2 

## Your Environment
Staging1
PC Chrome

## Possible Fix
Ensure that stock level is checked when a sub is edited, if stock level is zero, the sub order should have 0 QTY and insufficient stock messaging.

daniellemoorhead:
@sstead this is listed as a severity 2 bug yet it hasn't come up in our conversations about what needs to be completed for subscriptions to be "done" for now.

Is this actually an s2? Is it needed? Any s2 bugs are supposed to be picked up as a priority...yet this one seems to have slipped through without being noticed....

What should we do with it?

sstead:
Eeep sorry. Yes this needs fixing, before subs is used more widely. Its not super likely to occur (maybe a case for s3) but it’s a bug in a core feature (stock control) so could also be s2?

daniellemoorhead:
@kirstenalarsen @myriamboure @RachL what do you think? Worthy of an s2? Should be fixed ASAP?

kirstenalarsen:
i think s3 - not catastrophic if it does happen, they can workaround / amend order, likely only a few - I'd say s3

kristinalim:
[...]

PR #3904, opened 2019-06-06
Fix ofn-on-hand directive and fix usage of variant.on_hand in subscriptions order factory

### LINKED ISSUE #2507: FrontOffice - Edit Completed Order - If product stock level is zero, customer can't adjust quantity
## Description
I'm a customer with a subscription. In my order I received QTY 4 of a product, these were the last 4 in stock, there's now an on hand value of zero for that product. I, the customer, go to edit my order (because my shop allows me to) as I only now want 2 of that product. I find that I can't change the qty and see an error message 'there is zero in stock'. I should be able to reduce my order to 3,2, or 1. But I can only have zero, or leave my order as it is at 4.

## Steps to Reproduce
1. Create a sub order where the order purchases all the remaining stock for an item.
2. Make sure the shop allows customers to edit orders
3. As the customer view your order and try to change the QTY for that item. You'll see the bug.

## Animated Gif/Screenshot
https://www.useloom.com/share/8a583f5fc1544beeb99942af551f3f35

## Context
I'm trying to edit a sub order. 
This may occur for regular orders in shops that allow changes too. 

## Severity
s3

## Your Environment
Chrome, PC, Staging1

## Possible Fix
Allow me to reduce the qty, and return any items to on hand stock.

luisramos0:
#### What? Why?

Closes #2507

This PR fixes two problems, I leave them in the same PR because they are closely related.

1. The first one is the fix to #2507 where the ofn-on-hand directive was not taking the finalized quantity to calculate how much _more_ stock is available to add to the completed order.

2. The other problem this PR is fixing is a problem introduced in https://github.com/openfoodfoundation/openfoodnetwork/commit/12eab1bfa9b8e6ece85c1e5dd6e3691beda1d687#diff-c3c4192f302cc77e9a8547012fe86ddb, since then variant.on_hand does not return infinity if variant is on_demand and subscriptions code was not adapted to that change at the time. This PR fixes that. I believe this problem was not seen yet in subscriptions because most of subs code works with skip_stock_check enabled.

#### What should we test?
We need to test 2507 is fixed.
the change in subscriptions code is covered with a auto test and should be verified with #2505 

#### Release notes
Changelog Category: Fixed
When editing completed orders, user will now see the correct out of stock warnings that match the actual stock available.

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
@@ -13,9 +13,7 @@ Darkswarm.directive "ofnOnHand", ->
       ngModel.$setDirty = setDirty
 
     ngModel.$parsers.push (viewValue) ->
-      on_hand = parseInt(attr.ofnOnHand)
-      finalized_quantity = parseInt(attr.finalizedquantity)
-      available_quantity = on_hand + finalized_quantity
+      available_quantity = scope.available_quantity()
       if parseInt(viewValue) > available_quantity
         alert t("js.insufficient_stock", {on_hand: available_quantity})
         viewValue = available_quantity
@@ -23,3 +21,8 @@ Darkswarm.directive "ofnOnHand", ->
         ngModel.$render()
 
       viewValue
+
+    scope.available_quantity = ->
+      on_hand = parseInt(attr.ofnOnHand)
+      finalized_quantity = parseInt(attr.finalizedquantity) || 0 # finalizedquantity is optional
+      on_hand + finalized_quantity
--- a/app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
@@ -14,9 +14,11 @@ Darkswarm.directive "ofnOnHand", ->
 
     ngModel.$parsers.push (viewValue) ->
       on_hand = parseInt(attr.ofnOnHand)
-      if parseInt(viewValue) > on_hand
-        alert t("js.insufficient_stock", {on_hand: on_hand})
-        viewValue = on_hand
+      finalized_quantity = parseInt(attr.finalizedquantity)
+      available_quantity = on_hand + finalized_quantity
+      if parseInt(viewValue) > available_quantity
+        alert t("js.insufficient_stock", {on_hand: available_quantity})
+        viewValue = available_quantity
         ngModel.$setViewValue viewValue
         ngModel.$render()
 
--- a/app/views/spree/orders/_line_item.html.haml
+++ b/app/views/spree/orders/_line_item.html.haml
@@ -22,7 +22,8 @@
   %td.text-right.cart-item-price{"data-hook" => "cart_item_price"}
     = line_item.single_display_amount_with_adjustments.to_html
   %td.text-center.cart-item-quantity{"data-hook" => "cart_item_quantity"}
-    = item_form.number_field :quantity, :min => 0, "ofn-on-hand" => "#{variant.on_demand && 9999 || variant.on_hand}", "ng-model" => "line_item_#{line_item.id}", :class => "line_item_quantity", :size => 5
+    - finalized_quantity = @order.completed? ? line_item.quantity : 0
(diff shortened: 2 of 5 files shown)


app/views/spree/orders/_line_item.html.haml
```diff
@@ -22,7 +22,8 @@
   %td.text-right.cart-item-price{"data-hook" => "cart_item_price"}
     = line_item.single_display_amount_with_adjustments.to_html
   %td.text-center.cart-item-quantity{"data-hook" => "cart_item_quantity"}
-    = item_form.number_field :quantity, :min => 0, "ofn-on-hand" => variant.on_hand, "ng-model" => "line_item_#{line_item.id}", :class => "line_item_quantity", :size => 5
+    - finalized_quantity = @order.completed? ? line_item.quantity : 0
```
sauloperez: Is the line_item.quantity column nullable? I'd like to trust its value to be an integer because we'll end up with lots of defensive programming otherwise.

app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
```diff
@@ -13,11 +13,16 @@ Darkswarm.directive "ofnOnHand", ->
...
+        viewValue = available_quantity
         ngModel.$setViewValue viewValue
         ngModel.$render()
 
       viewValue
+
+    scope.available_quantity = ->
+      on_hand = parseInt(attr.ofnOnHand)
+      finalized_quantity = parseInt(attr.finalizedquantity) || 0 # finalizedquantity is optional
+      on_hand + finalized_quantity
```
sauloperez: IMO this deserves some unit testing

app/views/spree/orders/_line_item.html.haml
luisramos0: I am not sure I understand your comment.
do you say that because you think this ternary op is defensive programming?
this is not defensive programming, this is calculating finalized_quantity (stock already taken from stock levels) and the finalized_quantity is not line_item.quantity if order is not complete because the stock was not finalized yet (taken from stock levels). do you see the point or do you mean something else?

luisramos0: maybe I need to add a comment to the code explaining this?

app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
luisramos0: yeah, agree but we only have one directive unit test. I gave it a try to create one for this now but it's not working... I remember I spent a huge amount of time in the past trying to test one directive without success...
and this one in particular using "ngModel.$setViewValue"...

app/views/spree/orders/_line_item.html.haml
mkllnk: I guess that he meant if `line_item.quantity` can be nil, we need to add more defensive programming here (and return 0 instead). But I just checked the schema and `quantity` can't be null in the database. So it shouldn't be nil here either.

app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
sauloperez: can we then create a tech debt issue with the problems you had :pray: ?

app/views/spree/orders/_line_item.html.haml
sauloperez: yes, that's my point. I wanted to be sure we are not defending against `line_item.quantity` being nil. That's a DB responsibility in this case.

app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
luisramos0: ah, yes, great point, I'll do that.
I dont know exactly what were the problems I had but a tech debt issue for angular directive unit tests will be good!

luisramos0: here: https://github.com/openfoodfoundation/openfoodnetwork/issues/3947

app/views/spree/orders/_line_item.html.haml
luisramos0: ok, thanks for clarifying. we are not defending here and we dont need to :-)

RachL:
@luisramos0 just to be sure, is this closing #2505 as well?

luisramos0:
@RachL no, 2505 is a different problem.
I rebased this one to resolve conflicts. ready for testing.

RachL:
The bug in #2507 is now solved: I can reduce my quantity with no errors. If I try to increase it above the the existing stock, I see an error. Overall subscriptions are working as usual. This is ready to go!

image

Issue #3438, opened 2019-02-05
Product Import code is tightly coupled to Spree models


mkllnk:
## Description

Product import assumes that columns in the spreadsheet correlate to attributes on products, variants or inventory items. This is not the case in Spree 2. For example, products and variants don't have `on_hand` or `on_demand`. One line in the spreadsheet affects more than one object in our data model. We end up with code like this:

```rb
          if object.public_send(attribute).blank? ||
             ((attribute == 'on_hand' || attribute == 'count_on_hand') &&
             entry.on_hand_nil)

            object.assign_attributes(attribute => setting['value'])
          end
```

```rb
    def check_on_hand_nil(entry, object)
      return if entry.on_hand.present?

      object.on_hand = 0 if object.respond_to?(:on_hand)
      object.count_on_hand = 0 if object.respond_to?(:count_on_hand)
      entry.on_hand_nil = true
    end
```

Product import uses a class `Entry` to represent one line in the spreadsheet. It can either be a product, a variant or an inventory item. The code has to handle these cases differently and there are a lot of branches for this. For example:
```rb
    def save_to_product_list(entry)
      save_new_product entry if entry.validates_as? 'new_product'

      if entry.validates_as? 'new_variant'
        save_variant entry
        @variants_created += 1
      end

      return unless entry.validates_as? 'existing_variant'

      # ...
    end
```

## Context

This came up during the Spree upgrade when I posted https://github.com/openfoodfoundation/openfoodnetwork/issues/2783#issuecomment-459601530.

## Severity

tech-debt

Product import is not a core feature, but it would be good to solve this before the next Spree upgrade. It will make the upgrade much easier.

## Possible Fix

We create new classes for the five different entry types.

- new_inventory_item
- existing_inventory_item
- new_product
- new_variant
- existing_variant

They could be named like `NewProduct` and `ExistingVariant`. They contain the processing logic for these types which should simplify the `EntryProcessor`. That processor has some code duplication in the methods `save_new_inventory_item`, `save_existing_inventory_item`, `save_new_product` etc. The new classes would need the following methods, for example:

- initialize(entry)
- assign_defaults(settings)
- valid?
- save
- errors

Most of those may be private and the only methods called by the processor are `new` and `save`.

luisramos0:
this is closely related to [this comment here](https://github.com/openfoodfoundation/openfoodnetwork/pull/3908#discussion_r292962110)
[...]

PR #3908, opened 2019-06-09
3877,3843 Improve handling of units and unit type in product import

### LINKED ISSUE #3843: Discrepancies in the number of products uploads in the v2 product import
## Description

Using the staging server I have tried to upload products. Doing 50 worked fine, and 50 went in. I upped the amount in the upload and started to get some issues.

https://docs.google.com/spreadsheets/d/1-n5PyoOqloRXWdw-DUGMoy0AWs5KeeQWKITnMIbDyyg/edit#gid=0

This dataset has 251 points, however when I upload only some have been uploaded (245), even with the importer stating it will upload all 251.

## Expected Behavior

It should upload the number it says it will. The ones that were missed did not come up with any errors so it's hard to see what the issue is

## Actual Behavior

I upload 251 products
Importer states no errors and all 251 will be uploaded
Upon completion, the importer tells you it has uploaded 245 - you cannot tell easily whats gone in and hasn't

## Steps to Reproduce

1. Upload this file on french staging
2. Observe numbers
3. import
4. observe second set of numbers and the descrepancy

## Animated Gif/Screenshot

image
image

When I upload the file again to see what is missing - this time it says there is an error however when you look at what it indicates, it does not look correct as the datafile is in the same format as all other entries 
image
image

## Context

Makes product import unreliable

## Your Environment

* Version used: v2

### LINKED ISSUE #3877: Unit Type is not correctly imported in Product Import v2
## Description

A file is uploaded and the variants contain different unit types, i.e. kg, g and and item (multipacks such as "g x 4 bags" which would then append to be 125_g x 4 bags_)

The product import review pages indicate the data will be imported as is in the csv files. However when you review the data, some datas unit types have changed. It looks like instead of maintaining the csv files input, it updates them all to use the same unit type. 

The might be the correct way, however in v1, we could do it as initially stated, and the main issue if this is how it is meant to be is you are initially told in the review page it will do something different to what the end result is, and it is then difficult to review the end result in BEP

## Expected Behavior

The importer should notify any changes.

## Actual Behaviour

Some data points change, whereas others dont, so it's really hard to see what the actual issue is that causes a change, especially when working with large data sets

## Steps to Reproduce

1. Upload - https://docs.google.com/spreadsheets/d/1DCHBuiBitkgjiAVR5jUG4Pz-E2UgGYDl0VgTyje2KTs/edit#gid=796587646
2. Note the import screen and all data points are fine
3. Upload
4.Check through the data and not that some have changed their formats when it comes to the unit type - S015Z | Alfalfa   Seeds - is an example

## Animated Gif/Screenshot

File going in:
File going in
what goes in:
What ends up going in

line 63 is an example of something that hasnt gone in correctly
https://docs.google.com/spreadsheets/d/1DCHBuiBitkgjiAVR5jUG4Pz-E2UgGYDl0VgTyje2KTs/edit#gid=796587646

## Context

We have large clients using a multipack format and need a way of accommodating this and being clear on what the tool can and cant do.

## Your Environment

* Version used:
* Browser name and version:

kristinalim:
#### What? Why?

- Closes #3877 
Edit - #3843 not solved by this. See comments below.

Let's say a shop already has a certain pre-existing product. We already show a validation error when importing a variant for this product with a _unit type_ or _variant unit name_ that does not match that of the product.

However, there are issues when the product does not yet exist before importing the file:

1. #3877 - When "Variant 1" in the import sheet uses "kg", and "Variant 2" uses "g". "Variant 2" is imported successfully but it uses the wrong unit type "kg".
2. #3843 - Remember that product import is processed in batches of 50, for performance reasons. When "Variant 1" in batch no. 1 uses "kg" and "Variant 2" in batch no. 2 uses "g", the import proceeds but "Variant 2" isn't saved without providing warning. The row for "Variant 2" passes validation because it is valid on its own, but not after "Variant 1" has been imported.

This PR adds validation for each row in the import spreadsheet. The _unit type_ or _variant unit name_ for a product-producer now needs to match that of the first row for that product-producer in the spreadsheet.

#### What should we test?

There is a chance of noticeable increase in processing time, although I doubt it because we're not doing anything that involves the DB. If this is noticeable while testing with the sample CSV files in the issues, kindly note.

* Test #3877 and #3843 using the spreadsheets provided in the respective issues.
* Test importing variant for an existing product:
    - With wrong and then correct unit type / variant unit name
* Test importing new product and a second variant for this product:
    - With second variant using wrong and then correct unit type / variant unit name
* Test importing new product and a second variant for this product which is in another batch (e.g. first variant in Row 4, second variant in Row 52):
    - With second variant using wrong and then correct unit type / variant unit name

#### Release notes

- Fix product import of a new product and subsequent variants, when the subsequent variants use the wrong unit type or variant unit name. More data validation is done now, which can result in slightly higher memory and CPU usage.

Changelog Category: Fixed

### DIFF
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -4,7 +4,8 @@
 
 module ProductImport
   class EntryValidator
-    def initialize(current_user, import_time, spreadsheet_data, editable_enterprises, inventory_permissions, reset_counts, import_settings)
+    def initialize(current_user, import_time, spreadsheet_data, editable_enterprises,
+                   inventory_permissions, reset_counts, import_settings, all_entries)
       @current_user = current_user
       @import_time = import_time
       @spreadsheet_data = spreadsheet_data
@@ -12,6 +13,7 @@ module ProductImport
       @inventory_permissions = inventory_permissions
       @reset_counts = reset_counts
       @import_settings = import_settings
+      @all_entries = all_entries
     end
 
     def self.non_updatable_fields
@@ -30,6 +32,7 @@ module ProductImport
         assign_enterprise_field(entry)
         enterprise_validation(entry)
         unit_fields_validation(entry)
+        variant_of_product_validation(entry)
 
         next if entry.enterprise_id.blank?
 
@@ -156,6 +159,17 @@ module ProductImport
 
       # variant_unit_name must be present if unit_type not present
       mark_as_invalid(entry, attribute: 'variant_unit_name', error: I18n.t('admin.product_import.model.conditional_blank')) unless entry.variant_unit_name && entry.variant_unit_name.present?
+    end
+
+    def variant_of_product_validation(entry)
+      return if entry.producer.blank? || entry.name.blank?
+
+      reference_entry = all_entries_for_product(entry).first
+      if entry.unit_type.present? && entry.unit_type.to_s != reference_entry.unit_type.to_s
+        mark_as_not_updatable(entry, "unit_type")
+      elsif entry.variant_unit_name.present? && entry.variant_unit_name.to_s != reference_entry.variant_unit_name.to_s
+        mark_as_not_updatable(entry, "variant_unit_name")
+      end
     end
 
     def producer_validation(entry)
@@ -332,6 +346,11 @@ module ProductImport
       entry.product_validations = options[:product_validations] if options[:product_validations]
     end
 
+    def mark_as_not_updatable(entry, attribute)
+      mark_as_invalid(entry, attribute: attribute,
+                             error: I18n.t("admin.product_import.model.not_updatable"))
+    end
+
     def import_into_inventory?
       @import_settings[:settings].andand['import_into'] == 'inventories'
     end
@@ -385,6 +404,20 @@ module ProductImport
       object.count_on_hand = entry.on_hand.presence
       object.on_demand = object.count_on_hand.blank? if entry.on_demand.blank?
       entry.on_hand_nil = object.count_on_hand.blank?
+    end
+
+    def all_entries_for_product(entry)
+      all_entries_by_product[entries_by_product_key(entry)]
+    end
+
+    def all_entries_by_product
+      @all_entries_by_product ||= @all_entries.group_by do |entry|
+        entries_by_product_key(entry)
+      end
+    end
+
+    def entries_by_product_key(entry)
+      [entry.producer.to_s, entry.name.to_s]
     end
   end
 end
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -164,12 +164,22 @@ module ProductImport
     def variant_of_product_validation(entry)
       return if entry.producer.blank? || entry.name.blank?
 
+      validate_unit_type_unchanged(entry)
+      validate_variant_unit_name_unchanged(entry)
+    end
+
+    def validate_unit_type_unchanged(entry)
+      return if entry.unit_type.blank?
       reference_entry = all_entries_for_product(entry).first
-      if entry.unit_type.present? && entry.unit_type.to_s != reference_entry.unit_type.to_s
-        mark_as_not_updatable(entry, "unit_type")
-      elsif entry.variant_unit_name.present? && entry.variant_unit_name.to_s != reference_entry.variant_unit_name.to_s
-        mark_as_not_updatable(entry, "variant_unit_name")
-      end
+      return if entry.unit_type.to_s == reference_entry.unit_type.to_s
+      mark_as_not_updatable(entry, "unit_type")
+    end
+
+    def validate_variant_unit_name_unchanged(entry)
+      return if entry.variant_unit_name.blank?
+      reference_entry = all_entries_for_product(entry).first
+      return if entry.variant_unit_name.to_s == reference_entry.variant_unit_name.to_s
+      mark_as_not_updatable(entry, "variant_unit_name")
     end
 
     def producer_validation(entry)
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -4,6 +4,7 @@
 
 module ProductImport
   class EntryValidator
+    # rubocop:disable Metrics/ParameterLists
     def initialize(current_user, import_time, spreadsheet_data, editable_enterprises,
                    inventory_permissions, reset_counts, import_settings, all_entries)
       @current_user = current_user
@@ -15,6 +16,7 @@ module ProductImport
       @import_settings = import_settings
       @all_entries = all_entries
     end
+    # rubocop:enable Metrics/ParameterLists
 
     def self.non_updatable_fields
       {
--- a/app/models/product_import/product_importer.rb
+++ b/app/models/product_import/product_importer.rb
@@ -238,18 +238,14 @@ module ProductImport
     end
 
     def build_entries_in_range
-      start_line = @import_settings[:start]
-      end_line = @import_settings[:end]
+      # In the JS, start and end are calculated like this:
+      # start = (batchIndex * $scope.batchSize) + 1
+      # end = (batchIndex + 1) * $scope.batchSize
+      start_data_index = @import_settings[:start] - 1
+      end_data_index = @import_settings[:end] - 1
 
-      (start_line..end_line).each do |i|
-        line_number = i + 1
-        row = @sheet.row(line_number)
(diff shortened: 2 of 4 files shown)


app/models/product_import/product_importer.rb
```diff
@@ -272,5 +266,14 @@ def delete_uploaded_file
       return unless @file.path == Rails.root.join('tmp', 'product_import').to_s
       File.delete(@file)
     end
+
+    def build_entries_from_rows(rows)
+      rows.each_with_index.inject([]) do |entries, (row, i)|
+        row_data = Hash[[headers, row].transpose]
+        entry = SpreadsheetEntry.new(row_data)
+        entry.line_number = i + 2
+        entries.push entry
```
luisramos0: let me ask a ruby question...
what is this entries where you call push? does it get to be @entries automagically?

app/models/product_import/product_importer.rb
```diff
@@ -267,11 +259,11 @@ def delete_uploaded_file
       File.delete(@file)
     end
 
-    def build_entries_from_rows(rows)
+    def build_entries_from_rows(rows, offset = 0)
       rows.each_with_index.inject([]) do |entries, (row, i)|
         row_data = Hash[[headers, row].transpose]
         entry = SpreadsheetEntry.new(row_data)
-        entry.line_number = i + 2
+        entry.line_number = offset + i + 1
```
luisramos0: I leave 2 questions:
1- here you are changing the previous behaviour called from build_entries right? it was i + 2 and now after this it's  offset (0) + i + 1
2- in this new version you skipped the logic in the break statement, right?    break if @sheet.last_row == line_number

and 2 recommendations
1- variable offset, I'd name it line_number_offset, much easier to understand
2- variable start_line and end_line  indicate a line but it's actually a number, better name is start_line_number and end_line_number

app/models/product_import/product_importer.rb
```diff
@@ -182,7 +182,7 @@ def init_product_importer
       @spreadsheet_data = SpreadsheetData.new(@entries, @import_settings)
       @validator = EntryValidator.new(@current_user, @import_time, @spreadsheet_data,
                                       @editable_enterprises, @inventory_permissions, @reset_counts,
-                                      @import_settings)
+                                      @import_settings, build_all_entries)
```
luisramos0: if I understand correctly, this way we are loading the full spreadsheet all the time so that we defend from these unit_types problems. I dont think there's any other way, but this will increase the import time, right? maybe users/testers should be warned?

app/models/product_import/product_importer.rb
kristinalim: @luisramos0 It's what the block names its first argument here:

```ruby
rows.each_with_index.inject([]) do |entries, (row, i)|
  # ...
end
```

Breaking it down:

```ruby
list = rows.each_with_index # => Enumerator
list.inject([]) do |entries, row_with_index|
  row, i = row_with_index
  # ...
  entries
end
```

app/models/product_import/product_importer.rb
kristinalim: Noting again that the batch size is 50. The `break if @sheet.last_row == line_number` was to make sure that, if there were 52 rows, the block would no longer need to be executed for rows 53 to 100.

I moved this logic out of the `build_entries_from_rows` method so we can reuse the method. Skipping the non-existent rows already happens here: `rows[(start_line - 1)..(end_line - 1)]`. This expression will end at the last element in `rows`.

```
> [1, 2, 3, 4, 5][0..3]
=> [1, 2, 3, 4]
> [1, 2, 3, 4, 5][0..4]
=> [1, 2, 3, 4, 5]
> [1, 2, 3, 4, 5][0..6]
=> [1, 2, 3, 4, 5]
```

app/models/product_import/product_importer.rb
kristinalim: Yes, you're right. It would increase import time, but fortunately nothing that would touch the database. We don't validate the items in this collection. I'll note this in the testing and release notes.

This mostly sets and reads the attributes. It would use more memory, but I guess only an MB or two even for larger import files (sample CSV file in #3843 with 252 rows has 76,955 characters). :sweat_smile:

kristinalim: :crossed_fingers: Hoping the difference is negligible.

@luisramos0 I updated the testing and release notes, FYI.

app/models/product_import/product_importer.rb
kristinalim: @luisramos0 About the Point 1, thanks. I updated the commit here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3908/commits/38519b2baeabab27fa9c7097d9130d461f0da5a7 I renamed the variable names and added a comment. I was getting confused among line number (with header, start at 1), row index (with header, start at 0), and data row index (no header, start at 0).

kristinalim:
Hi @luisramos0! Thanks for your review. I already replied to your comments and also updated the code.

app/models/product_import/product_importer.rb
luisramos0: [1, 2, 3, 4, 5][0..6] = [1, 2, 3, 4, 5]
:-O :+1:

app/models/product_import/product_importer.rb
luisramos0: ah, thanks for clarifying, I missed it.
This inject idiom is a but complex but now I understand it...

spec/features/admin/product_import_spec.rb
```diff
@@ -79,8 +79,10 @@
 
     it "displays info about invalid entries but no save button if all items are invalid" do
       csv_data = CSV.generate do |csv|
-        csv << ["name", "producer", "category", "on_hand", "price", "units", "unit_type"]
-        csv << ["Bad Carrots", "Unkown Enterprise", "Mouldy vegetables", "666", "3.20", "", "g"]
+        csv << ["name", "producer", "category", "on_hand", "price", "units", "unit_type", "shipping_category_id"]
```
Matt-Yorkley: The importer doesn't generally use `shipping_category_id` as a field (although it can). It's usually `shipping_category`, and uses the name instead of id. Could you use `shipping_category.name` in place of `shipping_category_id_str` and change the column name from `"shipping_category_id"` to `"shipping_category"`?

app/models/product_import/entry_validator.rb
```diff
@@ -386,5 +417,19 @@ def check_variant_override_stock_settings(entry, object)
...
+
+    def all_entries_by_product
+      @all_entries_by_product ||= @all_entries.group_by do |entry|
+        entries_by_product_key(entry)
+      end
+    end
+
+    def entries_by_product_key(entry)
+      [entry.producer.to_s, entry.name.to_s]
+    end
```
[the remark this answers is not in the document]
luisramos0: yeah, I dont think this creating bugs in its self and it is still manageable, but definitely deserves a tech debt issue that we can prioritize.

app/models/product_import/entry_validator.rb
```diff
@@ -4,6 +4,7 @@
 
 module ProductImport
   class EntryValidator
+    # rubocop:disable Metrics/ParameterLists
```
mkllnk: I think this should be dealt with. Either in a tech-debt task or in this pull request, I don't mind. But I agree with Pau that this is really bad style. It's like having global variables in PHP. ;-) Not blaming you. It's just a growth problem to deal with.

mkllnk:
@sauloperez I'm not clear what the blocker is for you. Do you think the style problems need to be addressed before merging this?

sauloperez:
@mkllnk I don't mind as long as we work on the tech debt issue right after. Complex code costs money and this one is a clear example. Moreover, I strongly believe it could be simpler.

luisramos0:
ok @sauloperez , I added a comment to https://github.com/openfoodfoundation/openfoodnetwork/issues/3438
and moved it to "dev ready"...

lin-d-hop:
Waiting on access to the test file from @sineadfenton

lin-d-hop:
Hi @kristina,
I did some arduous testing on this and found that it works well most of the time but does not quite solve #3843.

Details here:
https://docs.google.com/document/d/1QrbafB1BsDhveliknjOKWPZtN4G8UGfeTvWe-YGQHlk/edit#

To be honest I am unsure which is the correct behaviour so I will consult with other product folks and update the original issue.

This PR is a big improvement and I think we should go ahead and merge it. I will update #3843 with more information on the discrepency.

I edited the original issue so that #3843 will not automatically close when this is merged.

Issue #3844, opened 2019-05-10
Uploading Variants Causes Errors in V2 Product Import


sineadfenton:
## Description

When uploading files that contain different variants that need to be grouped together the importer seems to crash - there is no error message but the loading page never changes.

The difference here is that a file has 2 entries that need to be updated and then create 13 more - could the issue be that the importer cannot do this?

## Expected Behavior

Should upload as per usual

## Actual Behaviour

An 'invisible' error

When I refresh and attempt to reupload the data some of the data has gone in, and others haven't. There was no error indicated when I first imported the file during the initial checks

It looks like it stops after the 3rd entry where the data changes from unit type, ot variant unit type “bar”

## Steps to Reproduce

1. https://docs.google.com/spreadsheets/d/1-n5PyoOqloRXWdw-DUGMoy0AWs5KeeQWKITnMIbDyyg/edit#gid=788111594
2. Upload the above file
3. Import and see that the file never actually goes in, the page starts to load but never changes
4. Reupload the file to see that some have infact been uploaded

## Animated Gif/Screenshot

![image](https://user-images.githubusercontent.com/28807449/57527664-277e8580-7328-11e9-9864-afb48ceb63fe.png)

![image](https://user-images.githubusercontent.com/28807449/57527709-4e3cbc00-7328-11e9-98f3-5197c2da620c.png)

 GET https://staging.openfoodfrance.org/assets/admin/custom.css net::ERR_ABORTED 404
all-8e381660fc94b253a53474fe56c7de90.js:547 SyntaxError: Unexpected token P in JSON at position 0
    at JSON.parse (<anonymous>)
    at fromJson (all-8e381660fc94b253a53474fe56c7de90.js:543)
    at defaultHttpResponseTransform (all-8e381660fc94b253a53474fe56c7de90.js:546)
    at all-8e381660fc94b253a53474fe56c7de90.js:546
    at forEach (all-8e381660fc94b253a53474fe56c7de90.js:543)
    at transformData (all-8e381660fc94b253a53474fe56c7de90.js:546)
    at transformResponse (all-8e381660fc94b253a53474fe56c7de90.js:546)
    at processQueue (all-8e381660fc94b253a53474fe56c7de90.js:547)
    at all-8e381660fc94b253a53474fe56c7de90.js:547
    at Scope.$eval (all-8e381660fc94b253a53474fe56c7de90.js:547)
[...]

PR #3915, opened 2019-06-11
Consider deleted products when creating permalinks

mkllnk:
#### What? Why?

Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/3844

Spree's implementation and our implementation to create a unique permalink failed to notice conflicts with soft-deleted products. This patch looks at deleted products as well.

#### What should we test?

Test product import:
- Upload products with Rachel's test sheet: https://docs.google.com/spreadsheets/d/1-n5PyoOqloRXWdw-DUGMoy0AWs5KeeQWKITnMIbDyyg/edit#gid=788111594
- Delete some of the products.
- Upload the products again.
- It should succeed.
- The new products have a different permalink to the products that got deleted.

Test enterprise groups:
- Try to assign the same permalink to two groups.
- A new permalink should be created and the data should save.

#### Release notes

Fixed: Creating new products could fail if a product with the same name was deleted before.

Changelog Category: Fixed


lib/open_food_network/permalink_generator.rb
```diff
@@ -16,7 +23,23 @@ def find_available_value(existing, requested)
   end
 
   def create_unique_permalink(requested)
-    existing = self.class.where('id != ?', id).where("permalink LIKE ?", "#{requested}%").pluck(:permalink)
+    existing = others.where("permalink LIKE ?", "#{requested}%").pluck(:permalink)
```
luisramos0: this is a module included in the product model but we are a bit far from the product model here, I'd add a bit more context to this method name by calling it other_products instead of others.

mkllnk: Careful, this class is also used in `EnterpriseGroup`.

Instead of heaving an OFN specific service, I would rather see an independent module that can live in lib or in a separate gem. And that's where we could move to Friendly Id at some point.

Another solution would be to overwrite permalinks when deleting products so that the permalink is available to new products again. Spree did that but I'm not sure we want to re-use permalinks. We don't actually use them for products at the moment. Only for enterprises and groups, I think.

luisramos0: ah, I see.
Friendly Id looks good!

lin-d-hop:
Looks good to me. Tested a few iterations of deletes and replicated permalinks.

https://docs.google.com/document/d/12heXfkfJOdeOJN9p_CoA0S9lWE6GiQM0eUYIZVABqXs/edit#

PR #3916, opened 2019-06-11
OC schedules are now filtered to schedules of OCs with same coordinator

### LINKED ISSUE #3306: [Subscriptions] if I manage more than one shop, schedules for shops 2 is proposed in other shops and creates bugs.
## Description
User 1 is manager of Hub 1 and hub 2.
She creates schedule 1 for hub 1 and schedule 2 for hub 2,.
When creating an OC for hub 2, associating schedule 1 is proposed as an option and I can select it and save. 
This is confusing as hub manager believe they can reuse schedules over hubs, and if they give the same name to schedules for multiple hubs they manage, they can choose the wrong one.

## Expected Behavior
When setting up and OC for a hub, I am only proposed the schedules setup for that hub.

## Actual Behavior
I am proposed all the schedules for all the hubs I manage.

## Steps to Reproduce
1.  Make sure you have 2 schedules setup for 2 hubs.
2. Create an OC for one of the two hubs, and see.

## Animated Gif/Screenshot
capture du 2019-01-10 19-19-47

## Context
Discovered as a potential source of mistake for a beta tester of subs in France who had an OC where subs didn't work at all...

## Severity
s3

## Your Environment
* Version used: v1.25
* Operating System and version (desktop or mobile): Ubuntu 18.1

luisramos0:
#### What? Why?

Closes #3306

#### What should we test?
Schedules list in Order Cycle page is filtered to schedules of order cycles with same coordinator.
there are 4 pages where this needs to be tested:
- simple create order cycle
- simple edit order cycle
- incoming/outgoing create order cycle
- incoming/outgoing edit order cycle

#### Release notes
Changelog Category: Changed
Subscriptions schedules now cannot be re-used across order cycles coordinated by different enterprises.


app/controllers/admin/schedules_controller.rb
```diff
@@ -31,7 +31,16 @@ def index
...
+      if params[:enterprise_id]
+        filter_schedules_by_enterprise_id(permissions.visible_schedules, params[:enterprise_id])
+      else
+        permissions.visible_schedules
+      end
+    end
+
+    # Filter schedules by OCs with a given coordinator id
+    def filter_schedules_by_enterprise_id(schedules, enterprise_id)
+      schedules.select { |schedule| schedule.coordinators.map(&:id).include? enterprise_id.to_i }
```
sauloperez: Is there anything stopping us from fetching this from DB? It will be orders of magnitude faster

luisramos0: I opted to do this because I'd like to keep the permissions code in visible_schedules as is and not mix permissions with data fetching here:
https://github.com/openfoodfoundation/openfoodnetwork/blob/e15ccdbd74a7e2582d0cdd251ac118567db95224/lib/open_food_network/permissions.rb#L135

luisramos0: otherwise I think I'd have to mix the two parts in one query.

sauloperez: Have you played with AR's `#merge`?

luisramos0: ok, I'll give it try.

app/controllers/admin/schedules_controller.rb
```diff
@@ -31,7 +31,16 @@ def index
...
+      if params[:enterprise_id]
+        filter_schedules_by_enterprise_id(permissions.visible_schedules, params[:enterprise_id])
+      else
+        permissions.visible_schedules
+      end
+    end
+
+    # Filter schedules by OCs with a given coordinator id
+    def filter_schedules_by_enterprise_id(schedules, enterprise_id)
+      schedules.where("order_cycles.coordinator_id = ?", enterprise_id.to_i)
```
luisramos0: is this better @sauloperez ?

mkllnk: I find that much better. You are relying on the permissions class joining with order cycles but that's okay. You could have the same join here and Rails would filter out the duplicate join. And you can use the same hash syntax:
```rb
schedules.joins(:order_cycles).where(order_cycles: { coordinator_id: enterprise_id.to_i })
```

We don't need the distinct select because coordinators are uniq on each order cycle.

luisramos0: ah, nice, I do that! new commit.

RachL:
All 4 pages are working as usual an only showing the schedules of the coordinator :)

Subs are launching / ending as usual.

PR #3919, opened 2019-06-12
Db mirror script

mkllnk:
#### What? Why?

This updated database mirror script helps debugging and preparing migrations for production data. It can also copy all images from AWS.

#### What should we test?

No test on staging servers needed.

#### Release notes

Fixed: Updated a data mirror script for developers.

Changelog Category: Fixed

### DIFF
--- a/script/mirror_db.sh
+++ b/script/mirror_db.sh
@@ -4,9 +4,25 @@
 # Useful for when you want to test a migration against production data, or see
 # the effect of codebase changes on real-life data
 
-# Usage: script/mirror_db.sh [ofn-staging1|ofn-staging2|ofn-prod]
-
 set -e
+
+HOST="$1"
+BUCKET="$2"
+: ${DB_USER='ofn_user'}
+: ${DB_DATABASE='openfoodnetwork'}
+
+DB_OPTIONS=(
+--exclude-table-data=sessions
+--exclude-table-data=spree_log_entries
+--exclude-table-data=spree_state_changes
+--no-acl
+)
+
+if [ -z "$HOST" ]; then
+  echo "[Error] SSH host missing.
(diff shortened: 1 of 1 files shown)


mkllnk:
I just found that if you mirror the database from a server with a different country and currency, you need to change your application.yml. For example:

```yml
# Spree zone.
#CHECKOUT_ZONE: Australia
CHECKOUT_ZONE: France
#CURRENCY: AUD
CURRENCY: EUR
```

script/mirror_db.sh
```diff
@@ -4,45 +4,48 @@
...
-
 # -- Mirror database
 echo "Mirroring database..."
-echo "drop database open_food_network_dev" | psql -h localhost -U ofn open_food_network_test
-echo "create database open_food_network_dev" | psql -h localhost -U ofn open_food_network_test
-ssh $1 "pg_dump -h localhost -U $DB_USER $DB_DATABASE $DB_OPTIONS |gzip" |gunzip |psql -h localhost -U ofn open_food_network_dev
-
+dropdb -h localhost -U ofn open_food_network_dev
+createdb -h localhost -U ofn open_food_network_dev
+ssh -C "$HOST" "pg_dump -h localhost -U $DB_USER $DB_DATABASE ${DB_OPTIONS[@]}" | psql -h localhost -U ofn open_food_network_dev
```
Matt-Yorkley: So you're pulling a production database and adding it to your local machine. Wont the details be mismatched? eg. the owner will be `ofn_user` and the database will be `openfoodnetwork` in the production dump, but that won't match your dev settings, right?

mkllnk: Yes, the owner doesn't match but as long as you have the superuser role accessing the database it doesn't matter. The database name doesn't matter either, it's copied from `$DB_DATABASE` to `open_food_network_dev`.

PR #3923, opened 2019-06-12
User friendly error when creating a product with an image in an unsupported format

### LINKED ISSUE #764: Product creation crashes with wrong image
## Description

**Paperclip::Errors::NotIdentifiedByImageMagickError** in **spree/admin/images#create**
Paperclip::Errors::NotIdentifiedByImageMagickError

[View on Bugsnag](https://bugsnag.com/yaycode/openfoodnetwork/errors/54b04f4499b2cdc8b867a1f8?event_id=568a55383644892447eda417)
## Stacktrace

```
gems/paperclip-3.5.4/lib/paperclip/geometry_detector_factory.rb:10 - make
gems/paperclip-3.5.4/lib/paperclip/geometry.rb:26 - from_file
bundler/gems/spree-afcc23e489eb/core/app/models/spree/image.rb:36 - find_dimensions
```

[View full stacktrace](https://bugsnag.com/yaycode/openfoodnetwork/errors/54b04f4499b2cdc8b867a1f8?event_id=568a55383644892447eda417)

## Expected Behavior

The product image is either processed successfully (all images supported) or a helpful error message is given.

## Actual Behavior

The snail page appears and the user doesn't know why.

## Steps to Reproduce

1. Log in as enterprise user.
2. Go to products.
3. Create a product.
4. Fill in the details and upload an image that is not recognised. (Sorry, we don't have a test image yet.)
5. Submit the form.

## Animated Gif/Screenshot

Not recorded yet.

## Context

Australian users trying to create products.

## Severity

S3: The "create product" feature is broken in some cases. The workaround is to use a different image file.

## Your Environment

* Version used: 5c8a638320f06874065b0e87c237ba7853adc1f7

## Possible Fix

There are several ideas:

1. Investigate if we can support more image formats.
2. Show an error message telling that the image is no good.
3. Separate the image upload from the product creation form so that it's obvious that this step fails.

luisramos0:
#### What? Why?

Closes #764

Instead of a [stack trace](https://github.com/openfoodfoundation/openfoodnetwork/issues/764#issuecomment-501280800) the user now sees a nice error message:
Screenshot 2019-06-12 at 23 02 58

We do this by catching the exception in the controller.

#### What should we test?
Product creation still works correctly.
Uploading a good image still works.
Uploading an image in a weird format (I tested with exr) shows a friendly error.

#### Release notes
Changelog Category: Fixed
Creating a product with an unsupported image format now shows a user friendly error.

### DIFF
--- a/app/controllers/spree/admin/products_controller_decorator.rb
+++ b/app/controllers/spree/admin/products_controller_decorator.rb
@@ -33,6 +33,10 @@ Spree::Admin::ProductsController.class_eval do
     delete_stock_params_and_set_after do
       super
     end
+  rescue Paperclip::Errors::NotIdentifiedByImageMagickError
+    invoke_callbacks(:create, :fails)
+    @object.errors.add(:base, t('spree.admin.products.image_upload_error'))
+    respond_with(@object)
   end
 
   def update
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3011,6 +3011,8 @@ See the %{link} to find out more about %{sitename}'s features and to start using
             error_saving_payment: Error saving payment
             submitting_payment: Submitting payment...
       products:
+        image_upload_error: "The product image was not recognised. Please upload an image in PNG or JPG format."
+
         new:
           title: 'New Product'
         unit_name_placeholder: 'eg. bunches'
--- a/spec/controllers/spree/admin/products_controller_spec.rb
+++ b/spec/controllers/spree/admin/products_controller_spec.rb
@@ -147,6 +147,25 @@ describe Spree::Admin::ProductsController, type: :controller do
       spree_post :create, product: product_attrs, button: 'add_another'
       expect(response).to redirect_to spree.new_admin_product_path
     end
+
+    describe "when user uploads an image in an unsupported format" do
+      it "does not throw an exception" do
+        product_image = ActionDispatch::Http::UploadedFile.new({
+          :filename => 'unsupported_image_format.exr',
+          :content_type => 'application/octet-stream',
+          :tempfile => Tempfile.new('unsupported_image_format.exr')
+        })
+        product_attrs_with_image = product_attrs.merge(
+          images_attributes: {
+            '0' => { attachment: product_image }
+          })
+
+        expect do
+          spree_put :create, product: product_attrs_with_image
+        end.not_to raise_error Paperclip::Errors::NotIdentifiedByImageMagickError
+        expect(response.status).to eq 200
+      end
+    end
   end
 
   describe "updating" do


app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -33,6 +33,10 @@ def create
     delete_stock_params_and_set_after do
       super
     end
+  rescue Paperclip::Errors::NotIdentifiedByImageMagickError
+    invoke_callbacks(:create, :fails)
+    @object.errors.add(:base, t('spree.admin.products.image_upload_error'))
```
sauloperez: may I suggest we let the user which formats we accept? I think this is the kind of messages that confuse the user "which format am I supposed to use?"

luisramos0: tbh I have no clue what is that list...
I dont see this ImageMagick in our dependencies list... do you know if we use it?

mkllnk: Maybe we can recommend common formats?

```diff
- Error uploading image, you may need to use a different image format
+ This image was not recognised. Please upload an image in PNG or JPG format.

Matt-Yorkley: We install ImageMagick via ofn-install I think.

luisramos0: nice one, I found this list: https://imagemagick.org/script/formats.php

Matt: that would make sense.

I'll follow Maikel's advise, looks reasonable.

luisramos0: done

RachL:
@luisramos0 I didn't manage to make it work. I always have an error 500 even with good format like .png or .jpg (by the way I think we need to list these formats somewhere).

Here are my testing notes with link to the image:
https://docs.google.com/document/d/152ZQxXvcK2FAYURYcGmaIIwlH2Vh7NO3NM0RraIG-k8/edit# 

Back to you

luisramos0:
image uploads are broken in staging FR, can you pls test this in staging uk for example?

RachL:
@luisramos0 indeed it works on UK staging. however I've noticed that you added that friendly message only to product creation, not editing. Is that correct? So we need to open a new issue for that. Moreover, uploading a .svg image does not work as well, but the user is not warned, he only sees an empty spot:

image

luisramos0:
nice.

the goal here is really to guide the user who is creating a product and sees a snail because of the image but has no clue what field is causing the problem...

there's no image in product edit, only edit images, it's a different problem and not so important. see last paragraph here:
https://github.com/openfoodfoundation/openfoodnetwork/issues/764#issuecomment-501280800

I'd create a new issue for that and also for the svg problem you describe.

RachL:
@luisramos0 if you use the submenu Images in product edit and you upload an unknown format, you will see a snail / Error 500. Same pattern than in product creation. If you click on the image in the bulk edit page, then the spinner will stop at some point with no indication. This is indeed less bad. But yes let's merge this as it improves the current situation anyway. I've opened https://github.com/openfoodfoundation/openfoodnetwork/issues/3976 for the error 500 and https://github.com/openfoodfoundation/openfoodnetwork/issues/3975 for the other type of image file.

luisramos0:
yes :+1 in both those situations the user is explicitly updating the image, so the user will know the image is the problem. when creating a product, the user has no clue if it's the image or some other field.

Issue #3925, opened 2019-06-13
Duplicate / failed payments when payment processing gaps occur


emilyjeanrogers:
## Description

Inconsistencies have been identified with the way OFN responds when payment gaps occur (ie. payments are created and then rolled back / disappear). The following scenarios have been recorded:
a) Customer places an order and is charged twice by Stripe
b) Customer doesn't complete an order but is charged anyway
c) Customer's payment failed but customer was not notified (so order was presumed successful)

## Expected Behaviour

Processed payments should not disappear from the system. 
Errors after the payment processing have to be handled gracefully.

## Actual Behaviour

Gaps in processed payments have led to duplicate and/or incorrect charging in some cases.

Further information from @mkllnk investigation:
In Aus production there have been 1021 gaps the in the payment ids of which 9 correspond to payments that were processed and then rolled back (disappeared). 5 are failed payments and 4 are successful payments.  Of the successful payments, one is from November 2018 and the other three occurred in the last 4 weeks (May/June 2019) for the recently created Enterprise North Arm Farms. 

## Steps to Reproduce

The issue cannot be reproduced.

## Animated Gif/Screenshot

![North Arm Farms Order Confirmation $9](https://user-images.githubusercontent.com/47620442/59395473-7ad07180-8dc7-11e9-9bc0-157543a677c0.png)
![Stripe confirmation of duplicate charge $9](https://user-images.githubusercontent.com/47620442/59395475-7dcb6200-8dc7-11e9-99f0-bda6d657b6cd.png)

## Context

Customers have complained to the Enterprise (North Arm Farms) about duplicate charging, failed orders and being charged for orders that were not received. The Enterprise has had to  process refunds through Stripe. The Enterprise has also had to spend time investigating the problem. After 4 weeks of consistent issues, they now manually cross check all orders against Stripe reports to identify duplicates and issues before customers can complain.

## Severity

bug-s2: a critical feature is broken for some users: duplicated and failed payments. Could affect any user. We can't find a satisfactory workaround.

## Your Environment

* Version used: 1.31.0 
* OFN Platform instance: Aus Production

## Possible Fix

Maybe related to this past issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/3924

kirstenalarsen:
[...]

Issue #3924, opened 2019-06-12
ActiveRecord::RecordNotUnique in checkout#update


mkllnk:
## Description

The checkout fails with this error. It can happen after the payment which means that the payment will disappear (database rollback). That can cause duplicate charges and happened to an Australian enterprise. It happened in all four reported cases in https://github.com/openfoodfoundation/openfoodnetwork/issues/3925.

This error affected more enterprises though. In those cases we were lucky though and the order was paid by bank transfer instead of credit card.

It first happened 5th May 2019 and was raised 5 times since then (today 13 June 2019).

## Error in OpenFoodNetwork

**ActiveRecord::RecordNotUnique** in **checkout#update**
PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "spree_shipments_pkey"
DETAIL:  Key (id)=(36089) already exists.
: INSERT INTO "spree_shipments" ("address_id", "cost", "created_at", "id", "number", "order_id", "shipped_at", "shipping_method_id", "state", "tracking", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) RETURNING "id"

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5cce713175df8a001a341c4f?event_id=5cf4a18a003f0fe818910000&i=gh&m=ci)

## Stacktrace

    app/models/spree/order_decorator.rb:175 - block in update_shipping_fees!
    app/models/spree/order_decorator.rb:172 - update_shipping_fees!
    app/controllers/checkout_controller.rb:134 - advance_order_state
    app/controllers/checkout_controller.rb:33 - update

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5cce713175df8a001a341c4f?event_id=5cf4a18a003f0fe818910000&i=gh&m=ci)

*Created by Maikel via Bugsnag*

mkllnk:
This error is raised when trying to update shipping fees. Saving the shipment tries to create a new record. That indicates that it wasn't created properly in the first case.

https://github.com/openfoodfoundation/openfoodnetwork/blob/080689ccfa6197f32cef6b25dc86c5e68dfd7619/app/models/spree/order_decorator.rb#L35
[...]

PR #3926, opened 2019-06-13
Add rake task to report lost payments

mkllnk:
#### What? Why?

Related to #3924 and #3925.

In rare cases, payment records can get lost through database rollbacks. This new rake task tries to find log entries for lost transactions and reports them as CSV. It can help us until the underlying bug is fixed. It will also help us monitor if we really fixed the bug or if this happens again.

I'm not sure if we should make this more sophisticaed (run through schedule.rb) or keep it as temporary tool we want to get rid of as soon as possible.

#### What should we test?

No testing required.

#### Release notes

Added script for sys admins to find lost payments.

Changelog Category: Added

### DIFF
--- a/lib/tasks/missing_payments.rake
+++ b/lib/tasks/missing_payments.rake
@@ -1,0 +1,56 @@
+# Find gaps in the sequence of payment ids.
+# If there are gaps then see if there is a log entry with a payment result for
+# the now lost payment. If there are some then you probably want to follow up
+# with the affected enterprise and see if customers need to be refunded.
+#
+# ## Usage
+
+# Report of the last 35 days:
+#   rake ofn:missing_payments[35]
+namespace :ofn do
+  desc 'Find payments that got lost'
+  task :missing_payments, [:days] => :environment do |_task_, args|
+    days = args[:days].andand.to_i || 7
+    payments_sequence = Spree::Payment.where("created_at > ?", days.days.ago).order(:id).pluck(:id)
+    payments_range = (payments_sequence.first..payments_sequence.last).to_a
+    missing_payment_ids = payments_range - payments_sequence
+    puts "Gaps in the payments sequence: #{missing_payment_ids.count}"
+    log_entries = Spree::LogEntry.where(
+      source_type: "Spree::Payment",
+      source_id: missing_payment_ids
+    )
+    return if log_entries.empty?
+
+    CSV do |out|
+      out << headers
+      log_entries.each do |entry|
+        begin
+          details = Psych.load(entry.details)
+        rescue StandardError
+          Logger.new(STDERR).warn(entry)
+          next
+        end
+        out << row(details, details.params)
+      end
+    end
+  end
+
+  def headers
+    [
+      "Created", "Order", "Success", "Message", "Payment ID", "Action",
+      "Amount", "Currencty", "Receipt"
+    ]
+  end
+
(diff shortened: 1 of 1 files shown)


luisramos0:
you can extract method to make rubocop happier :-)

mkllnk:
@luisramos0 Ah, yes. Rubocop tricked me there. We are missing a value for BlockLength in our style guide and the todo file overrode it. I think that it's actually a mistake in our config. But I also fixed the style in this commit.

By the way, a maximum method length of 10 doesn't work when you define lists within a method. The headers could go in a constant but what about the values? Do you have a better idea?

luisramos0:
nice catch Maikel, I'd vote for keeping the limit of 10 and using rubocop exception in places where you need to a define long lists.

lib/tasks/missing_payments.rake
```diff
@@ -0,0 +1,56 @@
...
+# Report of the last 35 days:
+#   rake ofn:missing_payments[35]
+namespace :ofn do
+  desc 'Find payments that got lost'
+  task :missing_payments, [:days] => :environment do |_task_, args|
+    days = args[:days].andand.to_i || 7
+    payments_sequence = Spree::Payment.where("created_at > ?", days.days.ago).order(:id).pluck(:id)
+    payments_range = (payments_sequence.first..payments_sequence.last).to_a
+    missing_payment_ids = payments_range - payments_sequence
+    puts "Gaps in the payments sequence: #{missing_payment_ids.count}"
```
sauloperez: We should avoid using `puts` and use `Rails.logger` instead. The Rails logging is better suited for these.

sauloperez:
> keep it as temporary tool we want to get rid of as soon as possible.

:+1:

lib/tasks/missing_payments.rake
mkllnk: I don't see this as log message. It's the defined output of the program. It's like `ls` is listing files. I don't think it should use a logger for that.

mkllnk:
@sauloperez 
> Do we know whether the log entry gets inserted within the lifespan of the transaction that rolls back?

It gets re-inserted after rollback:

https://github.com/openfoodfoundation/spree/blob/46d6f8f5fd434105b0c69958276d1727a3066a97/core/app/models/spree/log_entry.rb#L6-L14

lib/tasks/missing_payments.rake
```diff
@@ -0,0 +1,56 @@
...
+# ## Usage
+
+# Report of the last 35 days:
+#   rake ofn:missing_payments[35]
+namespace :ofn do
+  desc 'Find payments that got lost'
+  task :missing_payments, [:days] => :environment do |_task_, args|
+    days = args[:days].andand.to_i || 7
+    payments_sequence = Spree::Payment.where("created_at > ?", days.days.ago).order(:id).pluck(:id)
+    payments_range = (payments_sequence.first..payments_sequence.last).to_a
```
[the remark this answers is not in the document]
mkllnk: You're right. Fixed in a new commit. Please re-review. :-)

sauloperez:
There are at least 2 of those 3 Code Climate complaints that can be easily fixed.

mkllnk:
@sauloperez I'm moving this to test ready.

mkllnk:
Ah, no, it's no-test. ;-) I'm merging it then.

PR #3932, opened 2019-06-13
Improve message in private shopfront when user is not logged in or not a customer

### LINKED ISSUE #1613: Improve 'customer-only' shopfront messaging
## Description
Fir private/"customer only" shops, the default messages for non logged-in users and UX is confusing.
We need to rephrase the messages for UX under the signup process. 

## Current and Expected Behavior
**1- When a visitor is not logged in and open the url of a private shopfront:**
image
BECOMES:
Only approved customers can access this shop.
If you're already an approved customer, **login** or **signup** to proceed. 
Want to start shopping here? **Contact** SHOP NAME and ask about joining.

**2- When a visitor is logged in and open the url of a private shopfront but is not part of the customer list of the shop:**
image
BECOMES:
Sorry, only approved customers can access this shop.
If you'd like to start shopping here, please **contact** SHOP NAME to ask about joining.

## Steps to Reproduce
1. Make sure user A is not part of shop A customer list
2. Not logged in, user A open url of shopfront A, you see message 1 above.
3. Then logs in and recharge the shopfront page. you see message 2 above.

## Severity
S3 as UX is pretty bad now

luisramos0:
#### What? Why?

Closes #1613

New messages:

Screenshot 2019-06-13 at 19 54 52
Screenshot 2019-06-13 at 19 55 06

I kept the first sentence shared between cases, they differ in the issue (the second one starts with "Sorry,"). I think it's ok as is. Let me know if you disagree, I can add it.

#### What should we test?
See steps in issue details.

#### Release notes
Changelog Category: Changed
Improved message to user in private shopfronts when user is not logged in or user is not a customer yet.

### DIFF
--- a/app/views/shop/_messages.html.haml
+++ b/app/views/shop/_messages.html.haml
@@ -7,7 +7,9 @@
         - if spree_current_user.nil?
           = t '.require_login_html',
             {login: ('<a auth="login">' + t('.login') + '</a>').html_safe,
-             register: ('<a auth="signup">' + t('.register') + '</a>').html_safe}
+             signup: ('<a auth="signup">' + t('.signup') + '</a>').html_safe,
+             contact: link_to(t('.contact'), '#contact'),
+             enterprise: current_distributor.name}
         - else
           = t '.require_customer_html',
             {contact: link_to(t('.contact'), '#contact'),
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1165,12 +1165,11 @@ en:
   shop:
     messages:
       login: "login"
-      register: "register"
+      signup: "signup"
       contact: "contact"
-      require_customer_login: "This shop is for customers only."
-      require_login_html: "Please %{login} if you have an account already. Otherwise, %{register} to become a customer."
-      require_customer_html: "Please %{contact} %{enterprise} to become a customer."
-
+      require_customer_login: "Only approved customers can access this shop."
+      require_login_html: "If you're already an approved customer, %{login} or %{signup} to proceed. Want to start shopping here? Please %{contact} %{enterprise} and ask about joining."
+      require_customer_html: "If you'd like to start shopping here, please %{contact} %{enterprise} to ask about joining."
 
   # Front-end controller translations
   card_could_not_be_updated: Card could not be updated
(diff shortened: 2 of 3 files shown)


lin-d-hop:
I tested this one on French staging in FR and EN. I found an issue when the user is not logged in.

1) When logged in as a user able to access the shopfront I am correctly taken into shopfront - PASS
2) When logged in as a user without access to the shopfront I am correctly shown the message - PASS
3) When not logged in I see the slug 500 error - FAIL

Passing back to dev ready. Ping @luisramos

luisramos0:
ah, for case 3, it's a problem with the existing translations. The existing translations use a key called %{register} that is not longer available in the key "require_login_html".

I fixed FR on the server and you can test with default en like this:
https://staging.openfoodfrance.org/mary-s-online-shop/shop?locale=en
and see that en_GB is broken
https://staging.openfoodfrance.org/mary-s-online-shop/shop?locale=en_GB

This will happen whenever translators use invalid placeholders, in this case, I am removing one placeholder (register) and that's why it fails. I think it's better to go ahead like this and make sure all languages are updated (at least remove the register key).

Basically, after this PR is merged you need to update this translation and not use the register placeholder:
https://www.transifex.com/open-food-foundation/open-food-network/translate/#en_GB/enyml/80855218?q=key%3Arequire_login_html

mkllnk:
That sounds like a really bad problem. An outdated locale can cause snails. I couldn't find a quick solution for that. But there must be some way to make this fail more quietly in production, similar to missing translations.

We could also provide legacy keys to a few releases so that the old locales won't crash. But that's a lot of hassle.

luisramos0:
yeah, I understand, I dont mind doing this after merge, I can go to transifex and replace register with signup in all languages in this key. I think it's the quicker approach.

RachL:
Alright so let's merge. I will check during release testing if we didn't forget it!

luisramos0:
actually, looks like transifex is helping us here.
these translations like "require_login_html" are removed and become untranslated in transifex, so the next transifex update PR will delete the entry unless translators do the new translation:
https://www.transifex.com/open-food-foundation/open-food-network/translate/#en_GB/enyml/80855218?q=key%3Arequire_login_html

Issue #3502, opened 2019-02-15
Comply to rubocop styles


mkllnk:
### What we should change and why (this is tech debt)

Code Climate seems to be more relaxed than our Rubocop rules and we have several style violations. We need to either fix them or re-generate the rubocop todo list. It's probably the best to do both: fix the recent and obvious violations and defer the ones that are less important.

### Context

Code Climate complained about https://github.com/openfoodfoundation/openfoodnetwork/pull/3451. Usually I would change the pull request before merging, but there are pre-existing issues and I didn't want to block the pipeline.

### Impact and timeline

Within the next days or weeks while we still remember the recent code changes that violated our code style.

luisramos0:
@mkllnk the rubocop todo list has been rebuilt recently. I wonder if this issue is still valid.

mkllnk:
Ah, yes, almost done. I actually just stumbled across some violations fixed in https://github.com/openfoodfoundation/openfoodnetwork/pull/3935. I'll associate that PR with this issue.

mkllnk:
Closed via #3935.

PR #3935, opened 2019-06-14
Update rubocop style

mkllnk:
#### What? Why?

Working on #3926 rubocop didn't complain about the block length in my rake task. I only found out about the style violation through Code Climate. This happened because of the automatically generated `.rubocop_todo.yml` file which set the allowed block length to 774. Here I adjust the config so that the default block length of 25 is enforced on all files except some files already violating the style which are now listed in `.rubocop_manual_todo.yml`.

I also quickly cleaned up pre-existing violations that must have slipped through our processes in the last weeks.

#### What should we test?

No testing.

#### Release notes

Fixed the enforcement of code style guidelines.

Changelog Category: Fixed

### DIFF
--- a/.rubocop_manual_todo.yml
+++ b/.rubocop_manual_todo.yml
@@ -493,6 +493,72 @@ Metrics/AbcSize:
   - spec/models/product_importer_spec.rb
   - spec/support/performance_helper.rb
 
+Metrics/BlockLength:
+  Max: 25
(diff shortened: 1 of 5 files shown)


sigmundpetersen:
So this closes #3502 I guess?

sauloperez:
I guess so but I'd let @mkllnk decide

PR #3940, opened 2019-06-15
Remove matching of shipping method zone to user address on checkout

### LINKED ISSUE #3928: [v2] Cannot checkout if a shipping zone is not specified for a shipping method
## Description

Shipping zones are now mandatory on shipping methods.

But you can create a shipping method without a shipping category. When this happens, the hub manager does not see his mistake, but his buyers won't be able to checkout that product.

Similar to #3879

## Expected Behavior

It should not be possible that a user successfully creates a shipping method that breaks checkout.

## Actual Behaviour

Above is possible

## Steps to Reproduce

1. Create a shipping method with no zones
2. Use in an order cycle
3. Try to check out on that order cycle
4. SLUGS

## Context

Found in testing.

## Severity

S3

## Your Environment

* Version used: v2 testing #3840 
* Operating System and version (desktop or mobile): Ubuntu Desktop
* OFN Platform instance where you discovered the bug, and which version of the software they are using.  UK staging

## Possible Fix

@luisramos0 suggests the easiest solution is to stop a check for shipping method zones on checkout. Agreed in Slack.

luisramos0:
#### What? Why?

Closes #3928
We go for option C in see comment in 3928 [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/3928#issuecomment-502394688).

#### What should we test?
Verify issue is fixed and that other checkout cases related to shipping methods are also working, for example, verifying that #3879 is still working.

#### Release notes
Changelog Category: Changed
To make everyone's life easier, no shipping method zone validation is now done on checkout. Users can now checkout using shipping methods without zone and also using an address outside the zones of the selected shipping method.

### DIFF
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -53,6 +53,14 @@ Spree::ShippingMethod.class_eval do
     spree_calculators.send model_name_without_spree_namespace
   end
 
+  # This is bypassing the validation of shipping method zones on checkout
+  # It allows checkout using shipping methods without zones (see issue #3928 for details)
+  #   and it allows checkout with addresses outside of the zones of the selected shipping method
+  def include?(address)
+    return false unless address
+    true
+  end
+
   def has_distributor?(distributor)
     distributors.include?(distributor)
   end
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -57,8 +57,7 @@ Spree::ShippingMethod.class_eval do
   # It allows checkout using shipping methods without zones (see issue #3928 for details)
   #   and it allows checkout with addresses outside of the zones of the selected shipping method
   def include?(address)
-    return false unless address
-    true
+    address.present?
   end
 
   def has_distributor?(distributor)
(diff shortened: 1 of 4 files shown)


app/models/spree/shipping_method_decorator.rb
```diff
@@ -53,6 +53,13 @@ def self.calculators
     spree_calculators.send model_name_without_spree_namespace
   end
 
+  # This is bypassing the validation of shipping method zones on checkout
+  # It allows checkout using shipping methods without zones (see issue #3928 for details)
+  #   and it allows checkout with addresses outside of the zones of the selected shipping method
+  def include?(address)
+    address.present?
+  end
```
sauloperez: this deserves a unit test IMO

luisramos0: yeah, thanks for the reminder 🥇 
done.

RachL:
I can checkout a product that has shipping categories with both different zones and different categories. The reverse is also true, I think this is good to go!

luisramos0:
@mkllnk "Which use do shipping zones now have?"
it's a valid question specially because the answer is none.... imo this is the best approach now to avoid wasting more time with something that is not top priority.

PR #3941, opened 2019-06-16
Add en_ZA.yml copied from en_GB.yml as requested by instance manager

### LINKED ISSUE #3939: Add en_ZA.yml language to OFN

ZA instance with own translation.

ZA managers can translate as they need in ZA.

luisramos0:
#### What? Why?

Closes #3939

#### What should we test?
No tests required. After merge we need to verify that the transifex process works correctly for this new language.

#### Release notes
Changelog Category: Added
Added en_ZA as a new language :tada:

### DIFF
--- a/config/locales/en_ZA.yml
+++ b/config/locales/en_ZA.yml
@@ -1,0 +1,3094 @@
+en_GB:
+  language_name: "English"
+  activerecord:
+    attributes:
+      spree/order:
+        payment_state: Payment State
+        shipment_state: Shipment State
+        completed_at: Completed At
+        number: Number
+        state: State
+        email: Customer E-Mail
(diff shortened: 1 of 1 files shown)


config/locales/en_ZA.yml
```diff
@@ -0,0 +1,3094 @@
+en_GB:
```
sigmundpetersen: Shouldn't this here be`en_ZA` @luisramos0 ?

luisramos0: yes, my bug, I only realized that after fixing it in transifex...
anyway, done.
it's not coming from transifex in this PR:
https://github.com/openfoodfoundation/openfoodnetwork/pull/3945

PR #3943, opened 2019-06-17
Bring back ‘Shop for X products at:’ title

### LINKED ISSUE #2112: On producer modal, "shop X products at" is gone
## Description
On the producer modal, there are links at the bottom but there is not anymore any indication of what those links are.

## Expected Behavior
An intro line, as it was before "shop for X products at"
bugnew

## Actual Behavior
No intro line, directly the shops url
screenshot from 2018-03-04 19-40-58

## Steps to Reproduce
1. Go on map and click on a producer icon, or from a shop below product name.

## Context
Was retesting old issues.

## Severity
S3, a bit annoying in term of UX and no UX workaround really... and affect everyone.

## Your Environment
- Version used: 1.10.1

## Possible Fix
Restore the line.

jonleighton:
#### What? Why?

Fixes #2112

It looks like this bug was accidentally introduced during a refactor in
e5ca494db83e067f6f6863619561d5e7e8206bbd. The `bo-html` attribute was
replaced with `ng-html`, but it appears that it should actually have
used `ng-bind-html`
(https://docs.angularjs.org/api/ng/directive/ngBindHtml) because the
former is not a valid AngularJS directive.

It was also necessary to bubble up the `.cta-container` class in order to
get the appropriate styling on the title.

#### Before
image

#### After

image

#### What should we test?

The producer modal

#### Release notes

Bring back the 'Shop for X products at:' title on the producer modal. This was
accidentally lost some time ago.

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/templates/partials/producer_details.html.haml
+++ b/app/assets/javascripts/templates/partials/producer_details.html.haml
@@ -2,14 +2,14 @@
 -# Do not show this for producer shops selling only their own produce,
 -# Since a shopping link will already have been displayed in hub_details.html.haml
 .row.active_table_row.pad-top{ "ng-if" => "enterprise.is_primary_producer && enterprise.hubs.length > 0 && !(enterprise.hubs.length == 1 && enterprise.hubs[0] == enterprise)"}
-  .columns.small-12
+  .columns.small-12.cta-container
     .row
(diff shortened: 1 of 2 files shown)


RachL:
This is looking on staging! Thanks and welcome @jonleighton :)

image

PR #3953, opened 2019-06-18
First stored credit card becomes default again

### LINKED ISSUE #3727: [Subs] Intermittent Failures on Payments

mkllnk:
#### What? Why?

Closes #3727 

A user can store their credit card when checking out. Due to a bug in https://github.com/openfoodfoundation/openfoodnetwork/pull/3840 it
didn't become the default card any more. This behaviour is now restored.
If there is no default card yet, a new card added during checkout
becomes the default card.

#### What should we test?

Not storing:

- Set up Stripe for a shop.
- Delete all credit cards of the current user.
- Check out using Stripe, not storing the credit card.
- The new card should not appear in the stored cards (account profile).

Storing:

- Check out using Stripe, storing the entered credit card.
- Check the stored cards. The new card should be there and should be the default.
- Repeat another checkout storing a new card.
- The new card should be stored but not become the default (the default stays the same).

Storing a card in the profile:

- Create a card in the profile.
- The previous default card should stay the same.
- Then delete the default card.
- Now there is no default card.
- Create a new card.
- The new card should be the default.

#### Release notes

Fixed: The first stored credit card is automatically the default card again.

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/credit_card_decorator.rb
+++ b/app/models/spree/credit_card_decorator.rb
@@ -15,7 +15,7 @@ Spree::CreditCard.class_eval do
   belongs_to :user
 
   after_create :ensure_single_default_card
-  after_save :ensure_single_default_card, if: :is_default_changed?
+  after_save :ensure_single_default_card, if: :default_card_needs_updating?
 
   # Allows us to use a gateway_payment_profile_id to store Stripe Tokens
   # Should be able to remove once we reach Spree v2.2.0
@@ -37,6 +37,10 @@ Spree::CreditCard.class_eval do
 
   def default_missing?
     !user.credit_cards.exists?(is_default: true)
+  end
+
+  def default_card_needs_updating?
+    is_default_changed? || gateway_customer_profile_id_changed?
   end
 
   def ensure_single_default_card
--- a/spec/models/spree/credit_card_spec.rb
+++ b/spec/models/spree/credit_card_spec.rb
@@ -83,11 +83,26 @@ module Spree
           end
         end
 
-        context "and the checkout creates a one-time-use card" do
+        context "and the checkout creates a card" do
           let!(:card1) { create(:credit_card, onetime_card_attrs) }
+          let(:store_card_profile_attrs) {
+            {
+              cc_type: "visa",
+              gateway_customer_profile_id: "cus_FH9HflKAJw6Kxy",
+              gateway_payment_profile_id: "card_1EmayNBZvgSKc1B2wctIzzoh"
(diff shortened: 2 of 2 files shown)


app/models/spree/credit_card_decorator.rb
```diff
@@ -15,7 +15,7 @@
   belongs_to :user
 
   after_create :ensure_single_default_card
-  after_save :ensure_single_default_card, if: :is_default_changed?
+  after_save :ensure_single_default_card, if: :default_card_needs_updating?
```
luisramos0:
```suggestion
  after_save :ensure_single_default_card
  
  def ensure_single_default_card
      return unless default_card_needs_updating?
```

I think these inline if statements are never a good idea and should be moved into the method itself.
actually, I think callbacks are never a good idea, in this case, that means that saving a credit card should be in a service.

mkllnk: > I think these inline if statements are never a good idea and should be moved into the method itself.

I agree. The difficulty here is that we have two different events calling the same method (create and save events). They don't share the same conditions.

I also agree with avoiding callbacks. I opened a new tech debt issue: #3966. Please let me know what you think about the two modelling options.

luisramos0: I see :+1:
I left my vote in the tech debt issue.

RachL:
@mkllnk thank you so much for you detailed testing notes :)

Everything is looking good! Notes: https://docs.google.com/document/d/13Z3Iy15o3kZgMYp3ZUCv-ZGCZaBOT4TnIFYWAutsqvY/edit#

PR #3887, opened 2019-05-30
Add docker configuration for quick local setup

pacodelaluna:
#### What? Why?
Let's add a Docker setup for local environment. The idea behind it is to allow developers or testers to setup quickly there environment.

Closes #2290. Already closed by added for reference.

It will add great value to the project I think, people will be able to quickly see how the app is working and then focus their energy on real issues. On the downside, we will have to maintain this configuration but we have also to consider it as documentation, and almost real-time-updated one, so I think it worth it.

#### What should we test?
I have added a `DOCKER.md` file which describes the process to follow in order to have the application working locally. Tell me if it needs adjustments.

#### Release notes
- Add Docker configuration for quick local setup

#### Documentation updates
I will add a link to the `DOCKER.md` file in the wiki once the PR is validated and merged.

PR #3954, opened 2019-06-18
DRY and make database.yml more flexible

### LINKED ISSUE #3981: new database.yml breaks local server
## Description
#3887 introduced docker and improved the database.yml file.

Now when I run (as usually) "rails s", I get this error:
```
Called from /Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:28:in `new'
Exiting
/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:1222:in `initialize': could not translate host name "db" to address: nodename nor servname provided, or not known (PG::ConnectionBad)
	from /Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:1222:in `new'
	from /Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:1222:in `connect'
	from /Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:324:in `initialize'
	from /Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:28:in `new'

```

## Expected Behavior
I can run rails s locally.

## Actual Behaviour
rails s broken.

## Possible Fix
Workaround:
It all works fine if I remove this line from the new database.yml:
https://github.com/openfoodfoundation/openfoodnetwork/blob/8b42bd205ae81ba38c0dcc35ac2f4b2ab62f1a5b/config/database.yml#L6

sauloperez:
Closes #3981

#### What? Why?

This enables us to work on https://github.com/openfoodfoundation/ofn-install/issues/387 and brings the
needed flexibility so things like https://github.com/openfoodfoundation/openfoodnetwork/pull/3887 or any CI don't require a custom version. That's what ENV vars are for!

For instance, I no longer need to mess with my LXC setup to connect to the DB. I just need to have 2 env vars.

This won't change how we deal with DB connections in production, which is still done through [roles/app/templates/postgresql.yml.j2](https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/postgresql.yml.j2)

#### What should we test?

the app should be able to connect to the DB without any problem. It's enough with the deployment process as it already connects to the DB.

#### Release notes

DRY and make database.yml more flexible with ENV vars 

Changelog Category: Changed

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,8 +3,8 @@ defaults: &defaults
   encoding: unicode
   pool: 5
   host: localhost
-  username: <%= ENV.fetch('DB_USERNAME', 'ofn') %>
-  password: <%= ENV.fetch('DB_PASSWORD', 'f00d') %>
+  username: <%= ENV.fetch('OFN_DB_USERNAME', 'ofn') %>
+  password: <%= ENV.fetch('OFN_DB_PASSWORD', 'f00d') %>
 
 development:
   <<: *defaults
--- a/config/database.yml
+++ b/config/database.yml
@@ -2,7 +2,7 @@ defaults: &defaults
   adapter: postgresql
   encoding: unicode
   pool: 5
-  host: localhost
+  host: <%= ENV.fetch('OFN_DB_HOST', 'localhost') %>
   username: <%= ENV.fetch('OFN_DB_USERNAME', 'ofn') %>
   password: <%= ENV.fetch('OFN_DB_PASSWORD', 'f00d') %>
 
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,19 +1,23 @@
-default: &default
+defaults: &defaults
   adapter: postgresql
-  encoding: utf8
+  encoding: unicode
   pool: 5
-  timeout: 5000
-  host: db
-  port: 5432
+  host: localhost
+  username: <%= ENV.fetch('DB_USERNAME', 'ofn') %>
+  password: <%= ENV.fetch('DB_PASSWORD', 'f00d') %>
 
 development:
-  <<: *default
+  <<: *defaults
   database: open_food_network_dev
-  username: ofn
-  password: f00d
 
 test:
-  <<: *default
+  <<: *defaults
   database: open_food_network_test
-  username: ofn
-  password: f00d
+
+production:
+  <<: *defaults
+  database: open_food_network_prod
+
+staging:
+  <<: *defaults
+  database: open_food_network_prod
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ services:
     environment:
       ADMIN_EMAIL: ofn@example.com
       ADMIN_PASSWORD: ofn123
+      OFN_DB_HOST: db
     command: >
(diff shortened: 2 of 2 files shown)


config/database.yml
```diff
@@ -1,33 +1,23 @@
-development:
+defaults: &defaults
   adapter: postgresql
   encoding: unicode
-  database: open_food_network_dev
   pool: 5
   host: localhost
-  username: ofn
-  password: f00d
+  username: <%= ENV.fetch('DB_USERNAME', 'ofn') %>
```
mkllnk: Can we call it `OFN_DB_USERNAME` and `OFN_DB_PASSWORD`? I'm just imagining other apps on my computer using the same variable names.

config/database.yml
```diff
@@ -1,33 +1,23 @@
...
+
+development:
+  <<: *defaults
+  database: open_food_network_dev
 
 test:
-  adapter: postgresql
-  encoding: unicode
+  <<: *defaults
   database: open_food_network_test
```
Matt-Yorkley: If the database names were ENV vars as well, we could use a single file in both repos and remove the duplication from ofn-install, right? Would that be preferable?

https://github.com/openfoodfoundation/ofn-install/blob/master/roles/app/templates/postgresql.yml.j2

mkllnk: Yes, that would be great!

By the way, the merge conflict here is with the merged docker pr which sets the database host to `db`. That fails in our current dev setup where it's `localhost`. Maybe we should take ownership of this PR while Pau is away and fix the config.

Matt-Yorkley:
I've handled the merge conflicts here, added the prefixes Maikel asked for, and fixed the new issue with the merged Docker changes that are breaking dev environments (#3981).

config/database.yml
Matt-Yorkley: Prefixes added :+1:

config/database.yml
Matt-Yorkley: We can make a separate issue for this in ofn-install.

config/database.yml
```diff
@@ -1,19 +1,23 @@
-default: &default
+defaults: &defaults
   adapter: postgresql
-  encoding: utf8
+  encoding: unicode
   pool: 5
-  timeout: 5000
-  host: db
-  port: 5432
+  host: <%= ENV.fetch('OFN_DB_HOST', 'localhost') %>
```
pacodelaluna: :+1:

PR #3959, opened 2019-06-19
Edit tooltip text to make it readable

### LINKED ISSUE #1848: One tooltip is off the page on OC edit page
Updated by Myriam 21.05.2018

## Description
One tooltip on order cycle edit page / "advance settings" can't be seen

## Expected Behavior
Tooltip can be read properly

## Actual Behavior
Tooltip can't be seen (Myriam just retested in 13inch screen for instance, so "regular" screen)
image

## Steps to Reproduce
1. You need to have one order cycle created already, then click on "edit" 
2. Click on "advance settings" button below the main menu
3. On this advance setting zone there is a tooltip below "choose products from", this tooltip is on one line only an can't be read properly.

## Context
Retesting to clean and specify old issues.

## Severity
S5

## Your Environment
PC, Aus production (+ French production), Chrome

## Possible Fix
Display tooltip on two lines?

jonleighton:
#### What? Why?

Fixes #1848 by rebasing the fix proposed in
https://github.com/Ecksi/openfoodnetwork/pull/1 (which was confusingly merged
to a forked repo) onto the current master.

#### Release notes

Fix display of tooltip in order cycle advanced settings so that the text
doesn't run off screen.

Changelog Category: Fixed

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -878,7 +878,7 @@ en:
         add_distributor: 'Add distributor'
       advanced_settings:
         title: Advanced Settings
-        choose_product_tip: You can opt to restrict all available products (both incoming and outgoing), to only those in %{inventory}'s inventory.
+        choose_product_tip: You can restrict products incoming and outgoing to only %{inventory}'s inventory.
         preferred_product_selection_from_coordinator_inventory_only_here: Coordinator's Inventory Only
         preferred_product_selection_from_coordinator_inventory_only_all: All Available Products
         save_reload: Save and Reload Page


jonleighton:
Thanks for the explanation -- I've updated the PR.

RachL:
hhmmmm what is the definition of "off the screen" applied here? Because to me it still feels that the tooltip is off the screen:

image

On a smaller screen it is then truncated.

The text of the tooltip should be appearing like this when the screen is smaller:

You can opt
to restrict all
available products
(both incoming and
outgoing to only those in OFN Demo hub's inventory)

What do you think@luisramos0 @Matt-Yorkley ?

luisramos0:
this PR only makes the message smaller, so it definitely improves the situation. And the message can be further reduced in transifex by translators. But maybe it will still be off the screen for other screen sizes... the PR should be merged. The question is if we close the issue or not.

RachL:
thank you @luisramos0 I understand better. This is not the only tooltip displaying badly, so I would be in favor of working maybe on a issue that would set up a responsive tooltip style that would work in many cases? In the meantime, moving to ready to go.

luisramos0:
yes, agree!

PR #3962, opened 2019-06-19
First 2 view specs in OFN

luisramos0:
#### What? Why?

This PR adds the first 2 view specs. This is testing code added in #3963 (I include code from 3963 here so that we can run the spec before 3963 is merged).

This was blocked by an error in CanCan code in the NavigationHelper decorator, thanks @mkllnk for the help!!!
It's now working and ready for review. Only the 2 last commits:
https://github.com/openfoodfoundation/openfoodnetwork/pull/3962/commits

#### What should we test?
Green build. No tests.

#### Release notes
Changelog Category: Added
Added a new type of automated test (view spec) that will ease writing tests for template views.

#### Dependencies
Can only be merged after  #3963

### DIFF
--- a/spec/views/spree/admin/orders/edit.html.haml_spec.rb
+++ b/spec/views/spree/admin/orders/edit.html.haml_spec.rb
@@ -1,0 +1,38 @@
+require "spec_helper"
+
+describe "spree/admin/orders/edit.html.haml" do
+  include AuthenticationWorkflow
+  helper Spree::BaseHelper # required to make pretty_time work
+
+  around do |example|
+    original_config = Spree::Config[:enable_invoices?]
+    example.run
+    Spree::Config[:enable_invoices?] = original_config
+  end
+
+  before do
+    controller.singleton_class.class_eval do
+      def current_ability
+        Spree::Ability.new(Spree.user_class.new)
+      end
+    end
+
+    allow(view).to receive_messages spree_current_user: create_enterprise_user
+
+    order = create(:completed_order_with_fees)
+    order.distributor = create(:distributor_enterprise)
+    assign(:order, order)
+    assign(:shops, [order.distributor])
+    assign(:order_cycles, [])
+  end
+
+  describe "order values" do
+    it "displays order shipping costs, transaction fee and order total" do
+      render
+
+      expect(rendered).to have_content("Shipping: UPS Ground $6.00")
+      expect(rendered).to have_content("Transaction fee: $10.00")
+      expect(rendered).to have_content("Order Total $36.00")
+    end
+  end
+end
--- a/spec/views/spree/admin/orders/index.html.haml_spec.rb
+++ b/spec/views/spree/admin/orders/index.html.haml_spec.rb
@@ -1,0 +1,33 @@
+require "spec_helper"
+
+describe "spree/admin/orders/index.html.haml" do
+  include AuthenticationWorkflow
+
+  around do |example|
+    original_config = Spree::Config[:enable_invoices?]
+    example.run
+    Spree::Config[:enable_invoices?] = original_config
+  end
+
+  before do
+    allow(view).to receive_messages spree_current_user: create_enterprise_user
+  end
+
+  describe "print invoices button" do
+    it "displays button when invoices are enabled" do
+      Spree::Config[:enable_invoices?] = true
+
+      render
+
+      expect(rendered).to have_content("Print Invoices")
+    end
+
+    it "does not display button when invoices are disabled" do
+      Spree::Config[:enable_invoices?] = false
+
+      render
+
+      expect(rendered).to_not have_content("Print Invoices")
+    end
+  end
+end
--- a/spec/views/spree/admin/orders/index.html.haml_spec.rb
+++ b/spec/views/spree/admin/orders/index.html.haml_spec.rb
@@ -10,6 +10,12 @@ describe "spree/admin/orders/index.html.haml" do
   end
 
   before do
+    controller.singleton_class.class_eval do
+      def current_ability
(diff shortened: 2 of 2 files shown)


amazzoccone:
I'm not a ruby dev, but are we sure this two lines works the same way?
`return '' if (klass = klass_for(options[:label])) && cannot?(:admin, klass)`
`return '' if klass = klass_for(options[:label]) and cannot?(:admin, klass)`

luisramos0:
Hi @amazzoccone thanks for your comment. Yeah, those lines are equivalent, the problem is that both fail. It's a namespace issue inside cancan code...

mkllnk:
Have you tried the suggestions in this post? https://stackoverflow.com/questions/27580136/rspec-test-cancan-uninitialized-constant-ability-nameerror

If I understand the issue correctly, we are trying to build the admin menu. Only the sections you are allowed to access should appear in the menu. As enterprise user you are not allowed to see "Configuration". So it loads a class for each label, e.g. "Spree::Configuration". Then it calls cancan which looks for an entry that the user has the `:admin` permission on `Spree::Configuration`. That cancan code fails to find `Ability` within the current namespace. I think it expects this code to run within a controller (error in `controller_additions.rb`) which we didn't load because it's a view spec.

So my best guess would be that we need to load `Spree::Ability` somewhere, maybe in the helper? Or does it need to be loaded within the view because the cancan code tries to access it relative to the current namespace?

CanCan is one of these gems where I think: "WTF have they done to make this work auto-magically?".

https://github.com/ryanb/cancan/blob/f2f40c7aac4a00a88651641129eaad71916c1c82/lib/cancan/controller_additions.rb#L3-L5

https://github.com/ryanb/cancan/blob/f2f40c7aac4a00a88651641129eaad71916c1c82/lib/cancan/controller_additions.rb#L356-L358

I also found that Spree is loading the Ability in a controller helper:
https://github.com/openfoodfoundation/spree/blob/46d6f8f5fd434105b0c69958276d1727a3066a97/core/lib/spree/core/controller_helpers/auth.rb#L18
Maybe loading that helper would work as well?

spec/views/spree/admin/orders/edit.html.haml_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    example.run
+    Spree::Config[:enable_invoices?] = original_config
+  end
+
+  before do
+    controller.singleton_class.class_eval do
+      def current_ability
+        Spree::Ability.new(Spree.user_class.new)
+      end
+    end
```
luisramos0: this is the solution to the Ability problem described earlier ☝️

spec/views/spree/admin/orders/edit.html.haml_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    assign(:order_cycles, [])
+  end
+
+  describe "order values" do
+    it "displays order shipping costs, transaction fee and order total" do
+      render
+
+      expect(rendered).to have_content("Shipping: UPS Ground $6.00")
+      expect(rendered).to have_content("Transaction fee: $10.00")
+      expect(rendered).to have_content("Order Total $36.00")
```
luisramos0: this type of test can potentially replace much slower tests like [these feature specs in spec/features/admin/orders_spec.rb](https://github.com/openfoodfoundation/openfoodnetwork/blob/8b42bd205ae81ba38c0dcc35ac2f4b2ab62f1a5b/spec/features/admin/orders_spec.rb#L241)

spec/views/spree/admin/orders/index.html.haml_spec.rb
```diff
@@ -10,6 +10,12 @@
   end
 
   before do
+    controller.singleton_class.class_eval do
+      def current_ability
+        Spree::Ability.new(Spree.user_class.new)
+      end
+    end
```
luisramos0: this is the solution to the Ability problem described earlier ☝️

mkllnk: Interesting. What is `controller` here? Is that just a basic test controller that doesn't know anything about our app?

That would make sense to me. Then the problem would be that a helper is relying on a controller method that isn't there until you add it here? That sounds like a design flaw causing a bad dependency.

luisramos0: yes, I think this controller is just a test controller that can be changed like this.
I got this from here:
https://relishapp.com/rspec/rspec-rails/v/2-0/docs/view-specs/view-spec#spec-with-view-that-accesses-helper-method-helpers

This is testing the view and we can manipulate/instrument the "mock" controller like this.

luisramos0:
yes, agree, the ability code can be put into a shared place. I leave that for the next view spec.

spec/views/spree/admin/orders/edit.html.haml_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+describe "spree/admin/orders/edit.html.haml" do
+  include AuthenticationWorkflow
+  helper Spree::BaseHelper # required to make pretty_time work
+
+  around do |example|
+    original_config = Spree::Config[:enable_invoices?]
+    example.run
+    Spree::Config[:enable_invoices?] = original_config
+  end
```
sauloperez: :clap:

spec/views/spree/admin/orders/edit.html.haml_spec.rb
sauloperez: Thank you :heart:  ! you know how much I've been pushing to have a balanced testing pyramid

sauloperez:
@mkllnk agree with @luisramos0 answer? in that case, you can merge it yourself.

PR #3964, opened 2019-06-19
Handle soft deleted products in orders and fulfillment report

### LINKED ISSUE #3956: When deleting a product, orders and fulfillment report shows a 500 error
## Description

When I delete a product purchased in an order, I cannot use the orders and fulfillment report anymore.

## Expected Behavior

Even long after a product has been deleted, I should be able to check the order and fulfillments report.

## Actual Behaviour

After an order is placed and then one of its products is deleted, when check the report crashes with the message `undefined method 'name' for nil:NilClass`.

## Steps to Reproduce

1. Place an order
2. Delete one of the products in said order
3. Go check any of the order and fulfillment reports
4. :boom: 

BE's Bugsnag error https://app.bugsnag.com/open-food-network-belgium/open-food-network-belgium/errors/5d068ad9bd21c60019ca7d36?event_id=5d068ad90041ec272ba70000&i=sk&m=nw

## Severity

## Your Environment

* Version used: v2.1.0

## Possible Fix

This is tightly related to https://github.com/openfoodfoundation/openfoodnetwork/issues/3903 and the solution will be similar to https://github.com/openfoodfoundation/spree/pull/41

sauloperez:
### What? Why?

Closes #3956

Because of the `default_scope` paranoia adds to models, soft-deleted rows are excluded from the result set. Therefore, when the report class tries to compute values from them finds `nil` everywhere.

#### Solution

The solution I came up with is disgusting. 

First of all, I did spend a few hours digging through the mess of this implementation trying to see if it was possible to solve this using `ActiveRecord::Relation`'s `#includes` to eager load the products and their variant but skipping the `default_scope`. Which should have look similar to:

```ruby
line_items = permissions.visible_line_items.merge(
  Spree::LineItem
    .where(order_id: orders)
    .joins(product: :variants)
    .includes(product: :variants)
)
```

if anyone wants to give a try, that's what https://github.com/openfoodfoundation/openfoodnetwork/blob/ab1ed53507bc4726ec779491849dbe0733d419a6/lib/open_food_network/reports/line_items.rb#L12 would become.

We were lucky because indeed we pass an `ActiveRecord::Relation` around but I didn't manage to stop AR from adding `deleted_at IS NULL` in that query above.

The only quick way to solve it that I found was querying the variant wherever it's needed, which does not cause any new N+1 other than the ones we have, I think. So there should be no change in performance.

I haven't investigated any further (it's very tough to read this code) but apparently, deleted products, although they `acts_as_paranoid` now, are included in the result set. Thus, they are not returned as `nil`.

#### The future of this report

We must stop this madness. What should have a rather straightforward change has become a deep dive into the way our fancy algorithm works. Also, I had a few "Stop this s***" moments not even being able to follow all those procs of the report class and trying to see how the AR Relation is passed around. (I even though it was a lack of respect to every one of us to have such code LOL).

So I'm fully committed to refactoring it as soon as we get to the performance epic. It won't be possible to increase its performance by performing the right SQL queries (and possibly adding one or more database views) unless we refactor it almost completely. Meanwhile, I say we don't invest any minute on it. And I'm eager to do so now that understand how the current implementation works.

#### What should we test?

We should test each of the subtypes of report combining both deleted and non-deleted products. All of them should show up in the report's aggregates.

The steps I took while developing are:

1. Place an order with one product
2. Place another one with another product
3. Delete the first one
4. Check all the subtypes of orders and fulfillments report and see if they show both products and their variants

#### Release notes

Deleted products and variants are successfully displayed in the orders and fulfillments report. 

Changelog Category: Fixed

### DIFF
--- a/lib/open_food_network/order_grouper.rb
+++ b/lib/open_food_network/order_grouper.rb
@@ -24,9 +24,12 @@ module OpenFoodNetwork
       sorted_groups.each do |property, items_by_property|
         branch[property] = build_tree(items_by_property, remaining_rules)
 
-        unless rule[:summary_columns].nil? || is_leaf_node(branch[property])
-          branch[property][:summary_row] = { items: items_by_property, columns: rule[:summary_columns] }
-        end
+        next if rule[:summary_columns].nil? || is_leaf_node(branch[property])
+
+        branch[property][:summary_row] = {
+          items: items_by_property,
+          columns: rule[:summary_columns]
+        }
       end
 
       branch
--- a/lib/open_food_network/order_grouper.rb
+++ b/lib/open_food_network/order_grouper.rb
@@ -18,11 +18,17 @@ module OpenFoodNetwork
     def group_and_sort(rule, remaining_rules, items)
       branch = {}
       groups = items.group_by { |item| rule[:group_by].call(item) }
+
       sorted_groups = groups.sort_by { |key, _value| rule[:sort_by].call(key) }
+
       sorted_groups.each do |property, items_by_property|
         branch[property] = build_tree(items_by_property, remaining_rules)
-        branch[property][:summary_row] = { items: items_by_property, columns: rule[:summary_columns] } unless rule[:summary_columns].nil? || is_leaf_node(branch[property])
+
+        unless rule[:summary_columns].nil? || is_leaf_node(branch[property])
+          branch[property][:summary_row] = { items: items_by_property, columns: rule[:summary_columns] }
+        end
       end
+
       branch
     end
 
--- a/lib/open_food_network/orders_and_fulfillments_report.rb
+++ b/lib/open_food_network/orders_and_fulfillments_report.rb
@@ -215,6 +215,8 @@ module OpenFoodNetwork
       end
     end
 
+    # Returns a proc for each column displayed in each report type containing
+    # the logic to compute the value for each cell.
     def columns
       case params[:report_type]
       when "order_cycle_supplier_totals"
--- a/lib/open_food_network/orders_and_fulfillments_report.rb
+++ b/lib/open_food_network/orders_and_fulfillments_report.rb
@@ -59,116 +59,168 @@ module OpenFoodNetwork
     def rules
       case params[:report_type]
       when "order_cycle_supplier_totals"
-        [{ group_by: proc { |line_item| line_item.product.supplier },
-           sort_by: proc { |supplier| supplier.name } },
-         { group_by: proc { |line_item| line_item.product },
-           sort_by: proc { |product| product.name } },
-         { group_by: proc { |line_item| line_item.full_name },
-           sort_by: proc { |full_name| full_name } }]
+        [
+          {
+            group_by: proc { |line_item| Spree::Variant.unscoped.find(line_item.variant_id).product.supplier },
+            sort_by: proc { |supplier| supplier.name }
+          },
+          {
+            group_by: proc { |line_item| Spree::Variant.unscoped.find(line_item.variant_id).product },
+            sort_by: proc { |product| product.name }
+          },
+          {
+            group_by: proc { |line_item| Spree::Variant.unscoped.find(line_item.variant_id).full_name },
+            sort_by: proc { |full_name| full_name }
+          }
+        ]
       when "order_cycle_supplier_totals_by_distributor"
-        [{ group_by: proc { |line_item| line_item.product.supplier },
-           sort_by: proc { |supplier| supplier.name } },
-         { group_by: proc { |line_item| line_item.product },
-           sort_by: proc { |product| product.name } },
-         { group_by: proc { |line_item| line_item.full_name },
-           sort_by: proc { |full_name| full_name },
-           summary_columns: [proc { |_line_items| "" },
-                             proc { |_line_items| "" },
-                             proc { |_line_items| "" },
-                             proc { |_line_items| I18n.t('admin.reports.total') },
-                             proc { |_line_items| "" },
-                             proc { |_line_items| "" },
-                             proc { |line_items| line_items.sum(&:amount) },
-                             proc { |_line_items| "" }] },
-         { group_by: proc { |line_item| line_item.order.distributor },
-           sort_by: proc { |distributor| distributor.name } }]
+        [
+          {
+            group_by: proc { |line_item| Spree::Variant.unscoped.find(line_item.variant_id).product.supplier },
+            sort_by: proc { |supplier| supplier.name }
+          },
+          {
+            group_by: proc { |line_item| Spree::Variant.unscoped.find(line_item.variant_id).product },
+            sort_by: proc { |product| product.name }
+          },
+          {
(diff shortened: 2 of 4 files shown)


sauloperez:
Apparently the issue with includes and unscoped https://github.com/rails/rails/issues/11036, it's not fixed until Rails 4.2 in https://github.com/rails/rails/pull/17360 :see_no_evil: 

EDIT Oh God, I found that a Coopdevs member has a Rails gem to fix it :scream: https://github.com/markets/unscoped_associations

luisramos0:
Nice effort Pau!
I'll have a look at the build.

luisramos0:
I ended up missing this one. I'll leave to Pau to handle this as he is coming back this week.

lib/open_food_network/order_grouper.rb
```diff
@@ -18,11 +18,20 @@ def build_tree(items, remaining_rules)
...
-        branch[property][:summary_row] = { items: items_by_property, columns: rule[:summary_columns] } unless rule[:summary_columns].nil? || is_leaf_node(branch[property])
+
+        next if rule[:summary_columns].nil? || is_leaf_node(branch[property])
+
+        branch[property][:summary_row] = {
+          items: items_by_property,
+          columns: rule[:summary_columns]
+        }
       end
+
```
sauloperez: All this is just refactoring to make Rubocop happy. No behaviour changes.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -41,9 +42,7 @@ def header
          I18n.t(:report_header_order_cycle), I18n.t(:report_header_payment_method), I18n.t(:report_header_customer_code), I18n.t(:report_header_tags),
          I18n.t(:report_header_billing_street), I18n.t(:report_header_billing_street_2), I18n.t(:report_header_billing_city), I18n.t(:report_header_billing_postcode), I18n.t(:report_header_billing_state),]
       else
-        [I18n.t(:report_header_producer), I18n.t(:report_header_product), I18n.t(:report_header_variant),
-         I18n.t(:report_header_amount), I18n.t(:report_header_curr_cost_per_unit), I18n.t(:report_header_total_cost),
-         I18n.t(:report_header_status), I18n.t(:report_header_incoming_transport)]
+        DefaultReport.new(self).header
```
luisramos0: I dont understand this 5th case, there are only 4 types in the UI:
Screenshot 2019-07-05 at 14 33 45

sauloperez: Me neither, but it's the one most of the tests use...

luisramos0: 🙈 <- I can never remember the code for this emoji, this time I had to look it up :-)

luisramos0: we should create a tech debt issue for this: make sure the different versions of the report are tested (at least the basics) and that this 5th option is removed if it is not being used. I dont think we need to fix this in this PR.

sauloperez: https://github.com/openfoodfoundation/openfoodnetwork/issues/4015

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -238,31 +282,55 @@ def columns
...
+    end
+
+    def line_item_name
+      proc { |line_item| find_variant(line_item.variant_id).full_name }
+    end
+
+    def line_items_name
+      proc { |line_items| find_variant(line_items.first.variant_id).full_name }
+    end
+
```
sauloperez: I aimed at DRYing the tones of duplication (which led me to too many stupid and missed cases that needed fixing) but it required too much time. Until here, we improved a bit.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -238,31 +282,55 @@ def columns
...
       else
-        [proc { |line_items| line_items.first.product.supplier.name },
-         proc { |line_items| line_items.first.product.name },
-         proc { |line_items| line_items.first.full_name },
-         proc { |line_items| line_items.sum(&:quantity) },
-         proc { |line_items| line_items.first.price },
-         proc { |line_items| line_items.sum { |li| li.quantity * li.price } },
-         proc { |_line_items| "" },
-         proc { |_line_items| I18n.t(:report_header_incoming_transport) }]
+        DefaultReport.new(self).columns
```
sauloperez: This extract class refactoring is what I intend doing with the other types of reports soon.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -238,31 +282,55 @@ def columns
...
+    end
+
     def permissions
       return @permissions unless @permissions.nil?
       @permissions = OpenFoodNetwork::Permissions.new(@user)
     end
 
     def total_units(line_items)
-      return " " if line_items.map{ |li| li.unit_value.nil? }.any?
+      return " " if not_all_have_unit?(line_items)
```
sauloperez: extracting methods allowed us to conform to Rubocop metrics cops.

lib/open_food_network/orders_and_fulfillments_report/default_report.rb
```diff
@@ -0,0 +1,63 @@
...
+        ]
+      end
+      # rubocop:enable Metrics/AbcSize
+
+      private
+
+      attr_reader :context
+    end
+  end
+end
```
sauloperez: It's way easier to grasp what a type of report does with a separate class. Otherwise, you simply get lost in the giant report class.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -238,31 +282,55 @@ def columns
...
+
       total_units = line_items.sum do |li|
-        scale_factor = ( li.product.variant_unit == 'weight' ? 1000 : 1 )
-        li.quantity * li.unit_value / scale_factor
+        product = find_variant(li.variant_id).product
+        li.quantity * li.unit_value / scale_factor(product)
```
Matt-Yorkley: Should this be `scale_factor(li.product)`?

Matt-Yorkley: And is this covered by tests or not?

sauloperez: product refers to the local var above. I can't tell now if this is covered but it's very likely.

luisramos0: Matt, here we cant refer directly to li.product because we want deleted products to be used, and that's why we need to go through find_variant.

RachL:
So:
- I bought the "Framboise rose" from the enterprise *Raspberry release*
- Then 2 salads
- I then deleted the product Framboise rose
- The reports are showing deleted and non deleted products accurately:

image

image

image

image

This is ready to go

luisramos0:
amazing @sauloperez with this tricky task 👏

PR #3968, opened 2019-06-21
Rescue checkout on shipping id conflict

mkllnk:
#### What? Why?

Helps to investigate https://github.com/openfoodfoundation/openfoodnetwork/issues/3924

A better solution would be to move the update of shipments out of the
after_save callback and deal with it in controllers. Unfortunately,
that's a big and tricky task.

Since this exception is causing a lot of pain for some Australian
farmers, I introduced more logging here to understand the problem
better. The issue was observed in OFN v1 and may disappear with v2. But
we don't know that and should monitor it.

#### What should we test?

- Checkout
- Creating an order in the admin interface
- Modifying an order in the admin interface and recalculate fees

*Note: this patch is used already in Australian production.*

#### Release notes

Added logging for a rare checkout issue.

Changelog Category: Added

### DIFF
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -166,7 +166,21 @@ Spree::Order.class_eval do
     shipments.each do |shipment|
       next if shipment.shipped?
       update_adjustment! shipment.adjustment if shipment.adjustment
-      shipment.save # updates included tax
+      begin
+        shipment.save # updates included tax
+      rescue ActiveRecord::RecordNotUnique => error
+        # This error was seen in production on `shipment.save` above.
+        # It caused lost payments and duplicate payments due to database rollbacks.
+        # While we don't understand the cause of this error yet, we rescue here
+        # because an outdated shipping fee is not as bad as a lost payment.
+        # And the shipping fee is already up-to-date when this error occurs.
+        # https://github.com/openfoodfoundation/openfoodnetwork/issues/3924
+        Bugsnag.notify(error) do |report|
+          report.add_tab(:order, attributes)
+          report.add_tab(:shipment, shipment.attributes)
+          report.add_tab(:shipment_in_db, Spree::Shipment.find_by_id(shipment.id).attributes)
+        end
+      end
     end
   end
 
--- a/app/models/spree/order_decorator.rb
+++ b/app/models/spree/order_decorator.rb
@@ -166,21 +166,23 @@ Spree::Order.class_eval do
     shipments.each do |shipment|
       next if shipment.shipped?
       update_adjustment! shipment.adjustment if shipment.adjustment
-      begin
-        shipment.save # updates included tax
-      rescue ActiveRecord::RecordNotUnique => error
-        # This error was seen in production on `shipment.save` above.
-        # It caused lost payments and duplicate payments due to database rollbacks.
-        # While we don't understand the cause of this error yet, we rescue here
-        # because an outdated shipping fee is not as bad as a lost payment.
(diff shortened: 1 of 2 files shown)


app/models/spree/order_decorator.rb
```diff
@@ -166,7 +166,21 @@ def update_shipping_fees!
...
+        # While we don't understand the cause of this error yet, we rescue here
+        # because an outdated shipping fee is not as bad as a lost payment.
+        # And the shipping fee is already up-to-date when this error occurs.
+        # https://github.com/openfoodfoundation/openfoodnetwork/issues/3924
+        Bugsnag.notify(error) do |report|
+          report.add_tab(:order, attributes)
+          report.add_tab(:shipment, shipment.attributes)
+          report.add_tab(:shipment_in_db, Spree::Shipment.find_by_id(shipment.id).attributes)
+        end
+      end
```
luisramos0: I'd prefer to see this extracted to method.

```ruby
  def update_shipping_fees!
     shipments.each do |shipment|
        next if shipment.shipped?
        update_adjustment! shipment.adjustment if shipment.adjustment
        save_shipment_with_rescue
   end
   
   def save_shipment_with_rescue
        shipment.save # updates included tax
   rescue ActiveRecord::RecordNotUnique => error
        # This error was seen in production on `shipment.save` above.
        # It caused lost payments and duplicate payments due to database rollbacks.
        # While we don't understand the cause of this error yet, we rescue here
        # because an outdated shipping fee is not as bad as a lost payment.
        # And the shipping fee is already up-to-date when this error occurs.
        # https://github.com/openfoodfoundation/openfoodnetwork/issues/3924
        Bugsnag.notify(error) do |report|
          report.add_tab(:order, attributes)
          report.add_tab(:shipment, shipment.attributes)
          report.add_tab(:shipment_in_db, Spree::Shipment.find_by_id(shipment.id).attributes)
        end
    end
```

mkllnk:
Ah, thank you, yes. I changed that around a couple of times and forgot about extracting a method in the end. I extracted it now and added a unit test.

mkllnk:
Weird, this error has not been logged in Bugsnag this order cycle. But I got two new errors instead.

[bugsnag_error_stacktrace_ActiveRecord__StatementInvalid_event_5d1849f50045ff9161bf0000.txt](https://github.com/openfoodfoundation/openfoodnetwork/files/3348258/bugsnag_error_stacktrace_ActiveRecord__StatementInvalid_event_5d1849f50045ff9161bf0000.txt)
[bugsnag_error_stacktrace_StateMachine__InvalidTransition_event_5d1948bc00469dfd87950000.txt](https://github.com/openfoodfoundation/openfoodnetwork/files/3348260/bugsnag_error_stacktrace_StateMachine__InvalidTransition_event_5d1948bc00469dfd87950000.txt)

I just don't know what exactly fails. An error must be rescued somewhere without it being reported. Maybe within the state machine?

luisramos0:
are these 2 errors on the same transaction or different cases?
I dont understand but looks like a failed payment and then restart checkout code is simply wrong trying to move state machine to cart. And callbacks are failing because the restart checkout process failed?
Could be that this is related to callbacks still being executed after the update process fails?

mkllnk:
They are actually completely different issues. Different enterprises even. Sorry for the confusion. The first one is related to this issue though. Strangely, I don't see the error report that this PR introduced.

RachL:
1.  Checkout: working great on stripe and cash
 2. Creating an order in the admin interface : no issues
 3. Modifying an order in the admin interface and recalculate fees : all working fine as well!

Ready to go!

PR #3970, opened 2019-06-25
Toggle features via env instead of initializer

mkllnk:
#### What? Why?

First part of https://github.com/openfoodfoundation/ofn-install/issues/444.

A FeatureToggle could be switched via a class_eval in an initializer.
The initializer was installed via ofn-install. We want to get rid of
custom, untracked initializers. Here I'm changing the FeatureToggle
class to use environment variables instead.

This change needs to be followed up with a change in ofn-install to use
the new environment variable. It affects only Australian production.

#### What should we test?

Staging servers should not show the Connect and Learn sections on the home page.

Screenshot from 2019-06-25 17-35-08

#### Release notes

Changed the technical way of switching experimental features on or off.

Changelog Category: Changed

### DIFF
--- a/lib/open_food_network/feature_toggle.rb
+++ b/lib/open_food_network/feature_toggle.rb
@@ -1,13 +1,21 @@
 module OpenFoodNetwork
   class FeatureToggle
-    def self.enabled?(feature)
-      !!features.with_indifferent_access[feature]
+    def self.enabled?(feature_name)
+      true?(env_variable_value(feature_name))
     end
 
     private
 
-    def self.features
-      { connect_learn_homepage: false }
+    def self.env_variable_value(feature_name)
+      ENV.fetch(env_variable_name(feature_name), nil)
+    end
+
+    def self.env_variable_name(feature_name)
+      "OFN_FEATURE_#{feature_name.to_s.upcase}"
+    end
+
+    def self.true?(value)
+      value.to_s.casecmp("true").zero?
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/open_food_network/feature_toggle.rb
```diff
@@ -1,13 +1,21 @@
...
 
     private
 
-    def self.features
-      { connect_learn_homepage: false }
+    def self.value(feature)
+      ENV.fetch(env_name(feature), nil)
+    end
+
+    def self.env_name(feature)
```
luisramos0: for extra clarity I'd name this env_variable_name or env_var_name instead.

lib/open_food_network/feature_toggle.rb
```diff
@@ -1,13 +1,21 @@
...
     def self.enabled?(feature)
-      !!features.with_indifferent_access[feature]
+      true?(value(feature))
     end
 
     private
 
-    def self.features
-      { connect_learn_homepage: false }
+    def self.value(feature)
```
luisramos0: for extra clarity I'd name this method env_variable_value
and the variable feature could be name feature_name instead

luisramos0: this would become easier to understand:
true?(env_variable_value(feature_name))

mkllnk:
Thank you @luisramos0! I was struggling with the naming. Even though I found my version really beautiful:
```rb
# What is the true value of feature?
true?(value(feature))
```

Your suggestions are much clearer. :+1:

RachL:
All good here, ready to go!

image

PR #3971, opened 2019-06-25
Avoid `undefined method 'order'` errors when removing lineitems from basket

### LINKED ISSUE #3969: Error 500 when trying to empty my cart when shopping at Mary's
## Description
I've noticed while testing #3804 that trying to empty your cart on Mary's Online Shop thanks to the little "bin" icon would result in an error 500.

With Maryse or Fred/Fredo/Freddy, couldn't replicate the issue. The page reloads and redirects back to the shop when hitting that icon.

What is making Mary's shop special is that matter?

## Steps to Reproduce

1. Go to a staging server and add item to your cart from Mary's Shop
2. On the /cart URL hit the little bin icon
3. See an error 500

## Animated Gif/Screenshot
image

## Severity

## Your Environment
* Version used: v2.1
* Browser name and version: Firefox 67
* Operating System and version (desktop or mobile): Ubuntu 19.10

Matt-Yorkley:
#### What? Why?

Closes #3969 

A recent change in `app/models/spree/calculator_decorator.rb:12` is now generating fatal (undefined method) errors when the object does not respond to `#order`. This should fix it.

#### What should we test?

Items can be removed from basket without generating a snail page.

#### Release notes

Fixed error when removing items from basket.

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/calculator_decorator.rb
+++ b/app/models/spree/calculator_decorator.rb
@@ -9,7 +9,7 @@ module Spree
         [object]
       elsif object.respond_to? :line_items
         object.line_items
-      elsif object.order.present?
+      elsif object.respond_to? :order
         object.order.line_items
       else
         [object]
--- a/app/models/spree/calculator_decorator.rb
+++ b/app/models/spree/calculator_decorator.rb
@@ -9,7 +9,7 @@ module Spree
         [object]
       elsif object.respond_to? :line_items
         object.line_items
-      elsif object.respond_to? :order
+      elsif object.respond_to?(:order) && object.order.present?
         object.order.line_items
       else
         [object]
--- a/spec/models/spree/calculator_spec.rb
+++ b/spec/models/spree/calculator_spec.rb
@@ -5,12 +5,14 @@ module Spree
     let(:calculator) { Spree::Calculator.new }
     let!(:enterprise) { create(:enterprise) }
     let!(:order) { create(:order) }
+    let!(:shipment) { create(:shipment) }
     let!(:line_item) { create(:line_item, order: order) }
     let!(:line_item2) { create(:line_item, order: order) }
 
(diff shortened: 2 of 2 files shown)


Matt-Yorkley:
The error is at line 12 in `app/models/spree/calculator_decorator.rb`, which was recently changed from this:
```ruby
elsif object.respond_to? :order
        object.order.line_items
```
to this: 
```ruby
elsif object.order.present?
        object.order.line_items
```
I've changed it back.

Matt-Yorkley:
I think this needs some tests, even though it's a private method inside a `class_eval`. It's been changed 5 times in the last 6 months and looks like it's quite a critical bit of code...

app/models/spree/calculator_decorator.rb
```diff
@@ -9,7 +9,7 @@ def line_items_for(object)
         [object]
       elsif object.respond_to? :line_items
         object.line_items
-      elsif object.order.present?
+      elsif object.respond_to? :order
```
luisramos0: it's not equivalent, there will be times where order will return nil and in that case we better return the object and not the nil order.
So maybe better to have the following?
```suggestion
      elsif object.respond_to? :order && object.order.present?
```

I am not sure you have seen but this is partially tested in weight_spec. I think it's very good we are adding tests for this specific class.

Matt-Yorkley: Yeah, it seems like this method is called from various different places, both in OFN code for calculators, and elsewhere in Spree code.

Matt-Yorkley: Done!

RachL:
Awesome! Emptying and editing the cart is working well on all shops now :)

2 days to notice, fix and merge. This was fast!!

PR #3751, opened 2019-04-22
Invalid pickup instrc

jgman:
#### What? Why?

Closes #3192 

Added validation of length on pickup_instructions in exchanges model, as well as added text limit on input field for pickup_instructions to only, allow a user to enter 255 characters.

#### What should we test?
Creating and updating a new order cycle with exchanges that have pickup instructions with vaid and invalid lengths.

#### Release notes

Changelog Category: Fixed

PR #3985, opened 2019-07-01
Convert several fields from string to text

### LINKED ISSUE #3192: ActiveRecord::StatementInvalid in admin/order_cycles#update
## Description

Entering pick up instructions that exceed 255 characters leads to a server error.

**ActiveRecord::StatementInvalid** in **admin/order_cycles#update**
PG::StringDataRightTruncation: ERROR:  value too long for type character varying(255)
: UPDATE "exchanges" SET "pickup_instructions" = 'Pick up options are:                                                                                                                                                                  - Theatre Royal courtyard (#30 Hargraves St) in Castlemaine between 5pm and 6pm on a Wednesday                  - Drop into the Harcourt Organic Farming Co-op shop (phew mouthful!) in Harcourt Weds or Friday 10am-4pm.', "updated_at" = '2018-12-07 00:23:13.552177' WHERE "exchanges"."id" = 66943

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5c09bd711a398a001e24458e?event_id=5c09bd710031f6a7ff2a0000&i=gh&m=ci)

    gems/activerecord-3.2.22.5/lib/active_record/connection_adapters/postgresql_adapter.rb:1163 - async_exec

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5c09bd711a398a001e24458e?event_id=5c09bd710031f6a7ff2a0000&i=gh&m=ci)

## Expected Behavior

It should not be possible to enter instructions that exceed the character limit of the database. Updating the order cycle should always succeed.

## Actual Behavior

A server error is possible.

## Steps to Reproduce

1. Edit an order cycle.
2. Change the pick up instructions to a long text.
3. Try to save the changes.

## Animated Gif/Screenshot

Not recorded yet.

## Context

One of our Australian users experienced this and it got reported through Bugsnag.

## Severity

S4 It's annoying when saving fails, but most people can figure out to use a shorter description.

## Your Environment

* Version used: 1052d982550f0b7ac601b399f032291383ddad26

## Possible Fix

Limit the text field to the limit of the database field. Maybe evaluate if the limit of the database field should be increased. How would it look like with longer instructions?

jonleighton:
#### What? Why?

See discussion here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/3751#issuecomment-503416955

Fixes #3192 to enable exchange pickup instructions to exceed 255 characters.

I have also done a pass over the schema to identify other fields which
would benefit from being a text rather than a string. However, I ignored
all `spree_*` tables because I didn’t want to mess up the ‘default’
Spree schema.

#### What should we test?

That the fields below can be longer than 255 characters.

#### Release notes

Enable several fields to contain more than 255 characters. These are:
enterprise description, enterprise pickup times, exchange pickup time, exchange
pickup instructions, exchange receival instructions.

Changelog Category: Fixed

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -90,7 +90,6 @@ class Enterprise < ActiveRecord::Base
   validates :permalink, uniqueness: true, presence: true
   validate :shopfront_taxons
   validate :enforce_ownership_limit, if: lambda { owner_id_changed? && !owner_id.nil? }
-  validates :description, length: { maximum: 255 }
 
   before_validation :initialize_permalink, if: lambda { permalink.nil? }
   before_validation :ensure_owner_is_manager, if: lambda { owner_id_changed? && !owner_id.nil? }
--- a/db/migrate/20190701002454_convert_string_fields_to_text.rb
+++ b/db/migrate/20190701002454_convert_string_fields_to_text.rb
@@ -1,0 +1,17 @@
+class ConvertStringFieldsToText < ActiveRecord::Migration
+  def up
+    change_column :enterprises, :description, :text
+    change_column :enterprises, :pickup_times, :text
+    change_column :exchanges, :pickup_time, :text
+    change_column :exchanges, :pickup_instructions, :text
+    change_column :exchanges, :receival_instructions, :text
+  end
+
+  def down
+    change_column :enterprises, :description, :string
+    change_column :enterprises, :pickup_times, :string
+    change_column :exchanges, :pickup_time, :string
+    change_column :exchanges, :pickup_instructions, :string
+    change_column :exchanges, :receival_instructions, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20190506194625) do
+ActiveRecord::Schema.define(:version => 20190701002454) do
 
   create_table "adjustment_metadata", :force => true do |t|
     t.integer "adjustment_id"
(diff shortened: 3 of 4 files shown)


RachL:
I could edit (shipping method called pickup) and create (shipping method called recogida) pick-up instructions with more than 255 characters (279 to be precise):

image

However this is breaking the shopfront checkout with Firefox (at least)  :

image

So we cannot merge this. @jonleighton we use browserstack as a cross browser testing tool. Let me know if you would need an access to fix the checkout layout.

Back to In dev.

jonleighton:
Thanks for the testing, I will come back to this when I have some time :+1:

sigmundpetersen:
@jonleighton do you think you will have some time to look at this in the near future?

jonleighton:
Sorry for being slow on this. I just had a quick look.

@RachL I believe what you have tested is not actually something that has been changed in this pull request. I think you are looking at the 'shipping methods' page (e.g. `/admin/shipping_methods/new`), which already allowed a description longer than 255 characters.

I think the reason you're seeing UI weirdness is because you've basically entered a single very long word ("blaaaaa..."). By default on the web this sort of thing will "overflow" the container it is in, the browser doesn't wrap or hyphenate the text.

Since words that are this long don't really come up, a better test would be to use a long sentence, but with separate words. (I tend to use [lipsum](https://lipsum.lipsum.com/) for this sort of thing.)

I'm not actually sure where to find "exchanges" in the UI. Can you offer some guidance @luisramos0?

luisramos0:
@jonleighton an exchange is the relationship between an enterprise and an Order Cycle (called either "incoming" for suppliers into the order cycle or "outgoing" for distributors of the order cycle). The fields I mentioned above are all found when editing an Order Cycle, each line of the tables in the Order cycle page is an exchange.
I hope now my hint above makes more sense:
"then the OC fields: exchanges.pickup_time is "ready for" and exchanges.pickup_instructions is "customer instructions" or "pick up instructions". And exchanges.receival_instructions are the "receival details"."

RachL:
@jonleighton indeed sorry, I don't know why I didn't thought of lorem ipsum. This is working and ready to go!

image

Issue #3860, opened 2019-05-21
[Performance]  On /enterprise/shop


RachL:
## Goal
**- As a:** regular user
**- On page:** /enterprise/shop
**- I want to be able to do:** access the page in less than 5 seconds.

## Success indicator
In the case where a shop has 10,000 products the page can display in less than 5 seconds with regular internet connection

## Part 1 - improve logging and data collection

- [x] Improve Memcache Logging - #515 (OFNInstall)

## Part 2 - backend performance improvements

- [x] Many inefficient queries - OCs Cannot have Master Variants #4251 

~~- [ ] Many inefficient queries - Spike #3853~~

~~- [ ] When a shopfront's products don't fit into cache, the shop page takes ages to load #4256~~

## Part 3 - product pagination

- [x] Spike Impact of pagination on cache.

- [x] Move product list search server-side #4040 

- [x] Move product category search server-side #4041

- [x] Pagination in enterprise/shop #4042 

### Barriers 

- The cache solution we use to load the product list doesn't allow for a paginated view.
- We do not have an existing UI design pattern available for pagination on the platform.

luisramos0:
I assume the solution here is to optimize the way the current "infinite scroll" is working now, correct?

RachL:
@luisramos0 yes, we talked about pagination at the gathering. But also understanding why the infinite scroll stops at some point

luisramos0:
ok.
i just had a quick look at a long list of products:
https://openfoodnetwork.org.uk/tamar-valley-food-hubs/shop

in this case, the products endpoint returns over 500kb in under a sec.
I'd guess these 500kb of products in json are too much for the angular code. I think we need to build pagination here.

1. Frontoffice is currently getting products from controllers/shop_controller#products
2. Maybe we should build a new api endpoint in controllers/api/products_controller for this with pagination
3. Or re-use the backend api endpoint that already has pagination: controllers/spree/api/products_controller#bulk_products

We can build pagination in 1, create 2 with pagination, OR create a frontoffice parallel in 3.

Matt-Yorkley:
We could start with improving the query? #3853

luisramos0:
could be Matt, I am not sure, if you build pagination, the query may be fine for each page?

Matt-Yorkley:
I just read your notes again. If we're returning 500kb of JSON, we're probably doing something completely crazy.

luisramos0:
[...]

PR #3988, opened 2019-07-02
Fix JS render-blocking in Darkswarm

Matt-Yorkley:
#### What? Why?

Related to  #3860

Moves javascript inclusion to the bottom of the `<body>` tag so that it doesn't block page rendering.

blocking

#### What should we test?

Sanity checks? No code has been altered, just moved slightly.

I tested this in Lighthouse and the page speed increase was really significant.

#### Release notes

Fixed render-blocking javascript loading to improve page speed.

Changelog Category: Fixed

### DIFF
--- a/app/views/layouts/darkswarm.html.haml
+++ b/app/views/layouts/darkswarm.html.haml
@@ -14,16 +14,7 @@
       = favicon_link_tag "/favicon-staging.ico"
     %link{href: "https://fonts.googleapis.com/css?family=Roboto:400,300italic,400italic,300,700,700italic|Oswald:300,400,700", rel: "stylesheet", type: "text/css"}
 
-    = yield :scripts
-    %script{:src => "https://js.stripe.com/v3/", :type => "text/javascript"}
-    %script{src: "//maps.googleapis.com/maps/api/js?libraries=places,geometry#{ ENV['GOOGLE_MAPS_API_KEY'] ? '&key=' + ENV['GOOGLE_MAPS_API_KEY'] : ''} "}
     = stylesheet_link_tag "darkswarm/all"
-    = javascript_include_tag "darkswarm/all"
-    = javascript_include_tag "web/all"
-
-    = render "layouts/i18n_script"
-    = render "layouts/bugherd_script"
-    = render "layouts/matomo_tag"
     = csrf_meta_tags
 
   %body{class: body_classes, ng: {app: "Darkswarm"}}
@@ -32,14 +23,6 @@
       = javascript_include_tag "iehack"
 
     = render "layouts/shopfront_script" if @shopfront_layout
-
-    = inject_current_hub
-    = inject_json "user", "current_user"
-    = inject_json "railsFlash", "flash"
-    = inject_taxons
-    = inject_properties
-    = inject_current_order
-    = inject_currency_config
 
     .off-canvas-wrap{offcanvas: true}
       .inner-wrap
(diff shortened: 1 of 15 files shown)


Matt-Yorkley:
Wow, this took UK Staging's lighthouse ranking from 42/100 to 65/100...

app/views/layouts/darkswarm.html.haml
```diff
@@ -50,4 +33,23 @@
...
+    = inject_json "railsFlash", "flash"
+    = inject_taxons
+    = inject_properties
+    = inject_current_order
+    = inject_currency_config
+    = yield :injection_data
+
+    = render "layouts/bugherd_script"
+    = render "layouts/matomo_tag"
     = render 'spree/shared/google_analytics'
```
luisramos0: - from what I read google analytics will not report partial page loads if the tag is not in the head. maybe the same for matomo. I wouldnt worry about this, at our scale better have a fast page and a little incorrect stats

app/views/layouts/darkswarm.html.haml
```diff
@@ -50,4 +33,23 @@
 
           #footer
     %loading
+
+    %script{:src => "https://js.stripe.com/v3/", :type => "text/javascript"}
+    %script{src: "//maps.googleapis.com/maps/api/js?libraries=places,geometry#{ ENV['GOOGLE_MAPS_API_KEY'] ? '&key=' + ENV['GOOGLE_MAPS_API_KEY'] : ''} "}
+    = javascript_include_tag "darkswarm/all"
+    = javascript_include_tag "web/all"
+    = render "layouts/i18n_script"
```
luisramos0: we need to make sure on slow page loads the page doesnt load without the translations - this needs testing  in pages with lots of content

app/views/layouts/darkswarm.html.haml
```diff
@@ -50,4 +33,23 @@
 
           #footer
     %loading
+
+    %script{:src => "https://js.stripe.com/v3/", :type => "text/javascript"}
+    %script{src: "//maps.googleapis.com/maps/api/js?libraries=places,geometry#{ ENV['GOOGLE_MAPS_API_KEY'] ? '&key=' + ENV['GOOGLE_MAPS_API_KEY'] : ''} "}
```
luisramos0: we need to test google maps and make sure it still works - why do we include google maps in ALL frontoffice pages? :-O

app/views/layouts/darkswarm.html.haml
```diff
@@ -50,4 +33,23 @@
...
+    = javascript_include_tag "darkswarm/all"
```
luisramos0: I am not 100% sure this is possible. Are the pages rendering normally? Maybe we need to test with a very slow internet connection.

Matt-Yorkley: I did some very brief testing and looking around, and it all seemed to be working.

app/views/layouts/darkswarm.html.haml
Matt-Yorkley: The Maps API is used on `/map`, `/shops`, `/groups`, and `/producers`. The Stripe API definitely doesn't need to be loaded on every page though!

mkllnk: Stripe recommends using it on every page because it helps them collecting data and detect fraud. But honestly, I would only load it if there is something in the cart of an order cycle that uses Stripe. Should we open an issue?
https://stripe.com/docs/stripe-js/reference#including-stripejs

The documentation says nothing about where to include it. So I guess it's fine at the bottom.

RachL:
I created a new enterprise, opened an OC and checkout a product.

The order appeared correctly. This is ready to go!

luisramos0:
did you see my comments @RachL 
I think we need to watch out for this one as we test the release. we need to see how the pages render particularly in the shopfront, if the intermediary stages of the page load (just after the page is loaded) are not too bad visually.

RachL:
@luisramos0 ah no I didn't see your comment. But I've done my checkout on stripe. Maps are part of the release testing, but not Matomo. I need to add that.

mkllnk:
Bad news for this one. It broke the checkout via Stripe. Since we are about to release, I will revert it and we can have a second try without time pressure.

There is one JS error on the checkout page:
```
ReferenceError: angular is not defined
checkout:895:27
<anonymous>
http://0.0.0.0:3000/checkout:895:27
```

The error comes from this script which is embedded in the checkout page:
```html
<script>
  angular.module('Darkswarm').value("stripeObject", Stripe("pk_test_uMu..."))
</script>
```

The solution could be simply adding the script to the collection of scripts loaded at the bottom of the page.

mkllnk:
@RachL 
> I've done my checkout on stripe

Did you enter a new card or did you use an existing one?

Matt-Yorkley:
> The solution could be simply adding the script to the collection of scripts loaded at the bottom of the page.

It was at the bottom of the page already, after Angular was loaded... :thinking:

Issue #3991, opened 2019-07-02
[mobile ux] Display bug on cart button


Matt-Yorkley:
## In a feature branch or master?
View cart feature branch 

## Description

There's a display bug in `master` at the moment. It's caused by a semi-hidden element that apparently hasn't worked for a while, because of a typo. It's only visible on the medium and small screens.

Here's the current cart button:

![broken_cart](https://user-images.githubusercontent.com/9029026/60526329-0ca50d80-9ce8-11e9-8bbe-949d24a9b9fa.png)

If I fix the missing element (the current shop button) it then looks like this:

![fixed_but_bad](https://user-images.githubusercontent.com/9029026/60526359-2181a100-9ce8-11e9-9bde-42074288b9ef.png)

It's not well styled. 

So the question is: do we want to fix it or remove it?

Matt-Yorkley:
I've fixed the visual bug in #3992, but I've left the non-visible current shop button alone for now.

daniellemoorhead:
@Matt-Yorkley I haven't seen this bug in my travels, can you tell me a) if it's still a problem, and b) whether you fixed it in the PR?

daniellemoorhead:
>@Matt-Yorkley I haven't seen this bug in my travels, can you tell me a) if it's still a problem, and b) whether you fixed it in the PR?

@Matt-Yorkley?

![](https://media.giphy.com/media/MkxZKy77sWoEw/giphy.gif)

Matt-Yorkley:
Fixed :+1:

PR #3992, opened 2019-07-02
Fix display bug in cart button on mobile menu

Matt-Yorkley:
#### What? Why?

Related to  #3991

Fixes a visual bug with the cart button in tablet and mobile views.

#### What should we test?

Cart button should not look like this when a shop is selected:

broken_cart

#### Release notes

Fixed a minor cart button display issue on mobiles

Changelog Category: Fixed

### DIFF
--- a/app/assets/stylesheets/darkswarm/menu.css.scss
+++ b/app/assets/stylesheets/darkswarm/menu.css.scss
@@ -147,6 +147,7 @@ nav.top-bar {
 
     a, span {
       color: white;
+      display: inline-block;
     }
 
     .count {


RachL:
@Matt-Yorkley next time don't forget to write in the issue on which device you've experienced the problem ;).

I couldn't replicate playing with browserstack on stagings that do not have this PR. Maybe because of recent improvement on "shopping @" button? However sanity check is passing on Katuma staging so I think we can go forward with this PR.

Testing on :
- Safari on Iphone XS
- Safari on Ipad Air 2019
- Chrome for Samsung Galaxy
- Firefox on my phone (Fairphone Open OS)

PR #4000, opened 2019-07-03
Fix failing migration by clearing schema cache

### LINKED ISSUE #3980: Database migrations to v2 fail in one go
## Description

I tried to run the database migrations on French production data and it failed. Just re-running it works. The issue is some cached knowledge within ActiveRecord.

## Expected Behavior

Migrations run without errors.

## Actual Behaviour

One migration fails:
```
==  AddUniquenessOfOrderIdToSpreeShipments: migrating =========================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

missing attribute: lock_version
/home/maikel/code/openfoodnetwork/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb:26:in `destroy_all_but_latest_shipments'
/home/maikel/code/openfoodnetwork/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb:4:in `change'
/home/maikel/.rbenv/versions/2.1.5/bin/bundle:23:in `load'
/home/maikel/.rbenv/versions/2.1.5/bin/bundle:23:in `<main>'

Caused by:
ActiveModel::MissingAttributeError: missing attribute: lock_version
/home/maikel/code/openfoodnetwork/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb:26:in `destroy_all_but_latest_shipments'
/home/maikel/code/openfoodnetwork/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb:4:in `change'
/home/maikel/.rbenv/versions/2.1.5/bin/bundle:23:in `load'
/home/maikel/.rbenv/versions/2.1.5/bin/bundle:23:in `<main>'
```

## Steps to Reproduce

1. Copy the French production database. There may be a better way but I haven't investigated that yet. I tried to load sample data and created an order but the migrations worked on that data.
2. `git checkout master`
3. `rake db:migrate`

## Context

I was working on #3978.

## Severity

The workaround is to run the migration twice. But we don't want to deal with a workaround like this when deploying v2. It extends the downtime and could be confusing in case there are other errors. And the admin will panic at first.

## Your Environment

* Version used: migrating from v1.31.0 to master (723499332a4c43011bbddab1d3fc6bc75be57ba0)

Matt-Yorkley:
#### What? Why?

Closes #3980 

Fixes a failing migration for the v2 rollout which appeared with French data. 

I confirmed the issue using a combination of ofn-install, Vagrant and French production data. Added this change and ran the migrations again (after resetting the database to the pre-migration state) and it's now proceeding without error.

More info here: https://github.com/openfoodfoundation/openfoodnetwork/issues/3980#issuecomment-508058021

#### What should we test?

Maybe @mkllnk can test the migrations again?

### DIFF
--- a/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
+++ b/db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
@@ -1,6 +1,9 @@
 # This migration is an OFN specific migration that enforces an order to have a single shipment at all times
 class AddUniquenessOfOrderIdToSpreeShipments < ActiveRecord::Migration
   def change
+    Spree::InventoryUnit.connection.schema_cache.clear!
+    Spree::InventoryUnit.reset_column_information
+
     destroy_all_but_latest_shipments
 
     remove_index :spree_shipments, :order_id


db/migrate/20180910155506_add_uniqueness_of_order_id_to_spree_shipments.rb
```diff
@@ -1,6 +1,9 @@
 # This migration is an OFN specific migration that enforces an order to have a single shipment at all times
 class AddUniquenessOfOrderIdToSpreeShipments < ActiveRecord::Migration
   def change
+    Spree::InventoryUnit.connection.schema_cache.clear!
+    Spree::InventoryUnit.reset_column_information
```
sauloperez: I don't think it makes any difference in terms of behavior but I find more intention reveling to use `ActiveRecord::Base` in this cases, so the next dev doesn't think it's something related to `InventoryUnit`.

Matt-Yorkley: It _is_ related to `InventoryUnit`. From my comment in the issue:

> It's failing with missing attribute: lock_version at the call to shipments_to_delete.destroy_all, and I'm pretty sure this is caused by the deletion of related objects with dependent: destroy, namely Spree::InventoryUnit, which has it's lock_version attribute removed in a migration a few steps before this one.

Matt-Yorkley: I could add a comment?

RachL:
@mkllnk it seems this one will be tested by you, so I thought I would ping you here.

PR #4002, opened 2019-07-03
Adapting queries in permissions to Rails 4

luisramos0:
#### What? Why?

Continues #3708

Adapts permissions and order_cycle_permissions to rails 4.
This PR **fixes 102 specs** in the 2-1-stable branch.
- 2-1-stable build without these commits: [here](https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/2-1-stable/builds/33) (1534 broken specs)
- 2-1-stable build with these commits: [here](https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/2-1-stable/builds/34) (1432 broken specs)

#### What should we test?
Green build. We should review permissions in OCs. It's a complex area that is well covered with automatic tests.

#### Release notes
Changelog Category: Changed
Adapt permissions code to Rails 4.

This is part of upgrading to spree 2.1

### DIFF
--- a/lib/open_food_network/order_cycle_permissions.rb
+++ b/lib/open_food_network/order_cycle_permissions.rb
@@ -216,7 +216,7 @@ module OpenFoodNetwork
 
     def order_cycle_exchange_ids_involving_my_enterprises
       # Any exchanges that my managed enterprises are involved in directly
-      @order_cycle.exchanges.involving(managed_enterprises).pluck :id
+      @order_cycle.exchanges.involving(managed_enterprises.select(:id)).pluck :id
     end
 
     def order_cycle_exchange_ids_with_distributable_variants
--- a/lib/open_food_network/order_cycle_permissions.rb
+++ b/lib/open_food_network/order_cycle_permissions.rb
@@ -14,61 +14,61 @@ module OpenFoodNetwork
     def visible_enterprises
       return Enterprise.where("1=0") if @coordinator.blank?
       if managed_enterprises.include? @coordinator
-        coordinator_permitted = [@coordinator]
-        all_active = []
+        coordinator_permitted_ids = [@coordinator]
+        all_active_ids = []
 
         if @coordinator.sells == "any"
           # If the coordinator sells any, relationships come into play
-          related_enterprises_granting(:add_to_order_cycle, to: [@coordinator]).pluck(:id).each do |enterprise_id|
-            coordinator_permitted << enterprise_id
+          related_enterprises_granting(:add_to_order_cycle, to: [@coordinator.id]).each do |enterprise_id|
+            coordinator_permitted_ids << enterprise_id
           end
 
           # As a safety net, we should load all of the enterprises invloved in existing exchanges in this order cycle
-          all_active = @order_cycle.suppliers.pluck(:id) | @order_cycle.distributors.pluck(:id)
+          all_active_ids = @order_cycle.suppliers.pluck(:id) | @order_cycle.distributors.pluck(:id)
         end
 
-        Enterprise.where(id: coordinator_permitted | all_active)
+        Enterprise.where(id: coordinator_permitted_ids | all_active_ids)
       else
         # Any enterprises that I manage directly, which have granted P-OC to the coordinator
-        managed_permitted = related_enterprises_granting(:add_to_order_cycle, to: [@coordinator], scope: managed_participating_enterprises ).pluck(:id)
+        managed_permitted_ids = related_enterprises_granting(:add_to_order_cycle, to: [@coordinator.id], scope: managed_participating_enterprises )
 
         # Any hubs in this OC that have been granted P-OC by producers I manage in this OC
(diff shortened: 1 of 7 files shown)


luisramos0:
I fixed the rubocop issues related to long lines but left the rest of the existing issues, abc, long class, long methods... messy messi code order_cycle_permissions:
messi

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -90,8 +90,8 @@ def visible_variants_for_incoming_exchanges_from(producer)
         Spree::Variant.joins(:product).where('spree_products.supplier_id = (?)', producer)
       else
         # All variants of the producer if it has granted P-OC to any of my managed hubs that are in this order cycle
-        permitted = EnterpriseRelationship.permitting(managed_participating_hubs).
-          permitted_by(producer).with_permission(:add_to_order_cycle).present?
+        permitted = EnterpriseRelationship.permitting(managed_participating_hubs.select(:id)).
```
mkllnk: Interesting, so you can pass a relation as enterprise ids in `where('child_id IN (?)', enterprise_ids)` and it gets executed as subquery? That's awesome!

lib/open_food_network/permissions.rb
```diff
@@ -183,7 +183,8 @@ def coordinated_order_cycles
 
     def related_enterprises_granting(permission, options = {})
       parent_ids = EnterpriseRelationship.
-        permitting(options[:to] || managed_enterprises).
+        # this options[:to].id needs to be verified/tested
```
mkllnk: Should we create an issue for this? It won't happen if it's just a comment.

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -53,20 +74,38 @@ def visible_enterprises
...
-          product_ids = Spree::Product.joins(:variants_including_master).where("spree_variants.id IN (?)", variant_ids).pluck(:id).uniq
-          producers_active_ids = Enterprise.joins(:supplied_products).where("spree_products.id IN (?)", product_ids).pluck(:id).uniq
+          variant_ids = Spree::Variant.joins(:exchanges).
+            where(
+              "exchanges.receiver_id IN (?)
+                AND exchanges.order_cycle_id = (?)
+                AND exchanges.incoming = 'f'",
+              managed_participating_hubs.select("enterprises.id"),
+              @order_cycle
+            ).pluck(:id).uniq
```
mkllnk: These changes make me see all these little things that could be optimised. But it's not the scope of this PR and we should probably only optimise if it's a problem.

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -196,12 +262,14 @@ def editable_variants_for_outgoing_exchanges_to(hub)
     private
 
     def user_manages_coordinator_or(enterprise)
-      managed_enterprises.pluck(:id).include?(@coordinator.id) || managed_enterprises.pluck(:id).include?(enterprise.id)
+      managed_enterprises.pluck(:id).include?(@coordinator.id) \
```
mkllnk: `\`?

Is that the recommended way of doing this? I thought it would be to have the operator on the first line and then ruby knows to look for the rest in the next line.

Of course there are guidelines. ;-)
https://github.com/rubocop-hq/ruby-style-guide#no-trailing-backslash

lib/open_food_network/permissions.rb
luisramos0: no, I remove that in a later commit. it is tested in many specs, it was just a reminder for me to check it later at that point.

lib/open_food_network/order_cycle_permissions.rb
luisramos0: yes! this class is complex and there's much work we could do to make this more simple and for sure more optimised. I tried to keep it as simple as possible.
getting your brain around each of these queries takes time.

lib/open_food_network/order_cycle_permissions.rb
luisramos0: ah, yes, I saw @kristinalim using this in another PR and I wanted to explore. I just realized that keeping the operator in the first line, as you say, does the trick.

sauloperez:
How are we going to deal with this. Are we merging to `master` and then rebasing `2-1-stable` onto it?

luisramos0:
Not much logic is changed actually in that commit, I am basically switching from entities to ids all over the place.

Rebase is possible but takes a little while, so for now I'd just merge master into 2-1-stable.

luisramos0:
also, re that long commit, in my latest PRs I am starting to always make small commits. I think it's better for reviewers 💪

RachL:
These are the scenarios I've tried:

1. A producer hub
2. Simple hub
3. More complex hub with multiple shops.

Permissions seemed to work as usual and checkout was working each time as well. I guess we are good to go.

luisramos0:
great! thanks!

PR #4030, opened 2019-07-11
LineItems can always access soft-deleted variants

### LINKED ISSUE #4016: Server error (slug of doom) running reports on SOME order cycles/dates
## Description

When trying to run a report on a past order cycle/date range, the server instantly returns the system error/slug of doom screen.  

I am getting the exact same result when running a packing report, orders and distributors report, and all of the Orders and Fulfillment reports as well.
The payment report returns the expected result on the same time period though.

## Expected Behavior

A report is returned based on the parameters set or OC selected.

## Actual Behaviour

The NO GROW server error screen is shown.  It comes up almost immediately, so there is no chance that it is some kind of server timeout at play.

## Steps to Reproduce

1. Sign in as a hub administrator (regular user, NOT Super User)
2. Go to the Reports tab.
3. Select a report type listed above as returning the error. (IE: Packing report)
4. EITHER: Select an affected date range OR an affected order cycle.  (In my case, for the CLFC test shop, February 17, 2019 to March 4, 2019 OR "Test Cycle 11 (17 Feb 08:00) - 03 Mar 16:00)"
5. Click Search

## Animated Gif/Screenshot

OFN-Error-PackingReport-ByDate
OFN-Error-PackingReport-ByOC

## Context

I was trying to access order cycle data while testing reports after V2 upgrade to see what/if anything had changed.

## Your Environment

* Version used:  OFN-CAN (I think it's v2.1 now?)
* Browser name and version:  Edge 77 and Chrome 75
* Operating System and version (desktop or mobile): Windows 10 Pro 1903

Matt-Yorkley:
#### What? Why?

Closes #4016

Looking at errors around soft-deleted variants and ways to work around them in the context of reports, I ended up with the question: "should a `LineItem` _always_ have access to it's soft-deleted variant so it can check it's attributes?". I think the answer is probably "yes"?

Options for working around soft-deleted associated objects with the paranoia gem are here: https://github.com/rubysherpas/paranoia#usage

It looks like there's a nicer way of doing this by applying a scope directly on the associations, but it's not available until Rails 4...

#### What should we test?

Reports should display correctly even if a purchased variant has been deleted. From the bug report, this affects: "packing report, orders and distributors report, and all of the orders and fulfilment reports". 

#### Release notes

Ensure LineItem can always access it's soft-deleted Variant.

Changelog Category: Changed

### DIFF
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -20,7 +20,7 @@ Spree::LineItem.class_eval do
 
   before_destroy :update_inventory_before_destroy
 
-  delegate :unit_description, to: :variant
+  delegate :product, :unit_description, to: :variant
 
   # -- Scopes
   scope :managed_by, lambda { |user|
@@ -73,6 +73,11 @@ Spree::LineItem.class_eval do
           AND spree_adjustments.originator_type='Spree::TaxRate')").
       where('spree_adjustments.id IS NULL')
   }
+
+  def variant
+    # Overridden so that LineItems always have access to soft-deleted Variant attributes
+    Spree::Variant.unscoped { super }
+  end
 
   def cap_quantity_at_stock!
     scoper.scope(variant)
--- a/spec/features/admin/reports/packing_report_spec.rb
+++ b/spec/features/admin/reports/packing_report_spec.rb
@@ -1,0 +1,42 @@
+require "spec_helper"
+
+feature "Packing Reports", js: true do
+  include AuthenticationWorkflow
+  include WebHelper
+
+  let(:distributor) { create(:distributor_enterprise) }
+  let(:oc) { create(:simple_order_cycle) }
+  let(:order) { create(:order, completed_at: 1.day.ago, order_cycle: oc, distributor: distributor) }
+  let(:li1) { build(:line_item_with_shipment) }
+  let(:li2) { build(:line_item_with_shipment) }
+
+  before do
+    order.line_items << li1
+    order.line_items << li2
+    quick_login_as_admin
+  end
+
+  describe "viewing a report" do
+    context "when an associated variant has been soft-deleted" do
+      it "shows line items" do
+        li1.variant.delete
+
+        visit spree.admin_reports_path
+
+        click_on I18n.t("admin.reports.packing.name")
+        select oc.name, from: "q_order_cycle_id_in"
+
+        find('#q_completed_at_gt').click
+        select_date(Time.zone.today - 1.days)
+
+        find('#q_completed_at_lt').click
+        select_date(Time.zone.today)
+
+        find("button[type='submit']").click
+
+        expect(page).to have_content li1.product.name
+        expect(page).to have_content li2.product.name
+      end
+    end
+  end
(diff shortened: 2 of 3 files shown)


sauloperez:
I'm not entirely sure about this. It is exactly the discussion I wanted to raise in https://github.com/openfoodfoundation/openfoodnetwork/issues/3956#issuecomment-503580831 but didn't have much feedback.

If we choose this approach then we'll need to purposefully hide deleted variants from orders. As a supplier you don't want to list in the shopfront a product you just deleted. That was the whole point.

I tend to lean towards explicitly "whitelisting" deleted rows wherever it's needed. I think it might make things more obvious but also we might have just a few of those places, mainly /admin.

what are your thoughts @mkllnk @kristinalim @luisramos0 ?

Matt-Yorkley:
Yeah, I wasn't totally sure about this. It definitely needs discussion.

I figure there could be dozens of scenarios where a line item not being able to access it's variant could break things, and we have no test coverage for any of them. So either we slowly find and fix each one, or we make sure line items can access soft-deleted variants.

This doesn't remove the default scope completely, it's only in the context of a line item looking for it's own variant.

> we'll need to purposefully hide deleted variants from orders

Are we sure about that? If an order is trying to get an attribute of a variant via an existing line item, we want it to succeed, right?

luisramos0:
Let me see if I can say something useful here:
I think the default scope should never include deleted variants.
We need to adapt code, mainly in admin, where deleted variants need to be used: in those cases, we include deleted variants. There may be a few...
Most accesses to line_item.variant will need to include deleted variants, so maybe we can change that association to always include deleted variants.

sauloperez:
ok, I buy it.

> Are we sure about that? I assumed the opposite. If an order is trying to get an attribute of a variant via an existing line item, we want it to succeed, right?

That's actually right. If a variant purchased in an old order gets deleted after completion, we still want to display it. This will happen for orders that are in the middle of the checkout too, we need to be aware of it.

We'll also need to create an issue to undo https://github.com/openfoodfoundation/openfoodnetwork/pull/3964/files#diff-8fbbbe23b3dc4fda696811e14d73a1f2R332. I can do it myself.

app/models/spree/line_item_decorator.rb
```diff
@@ -20,7 +20,7 @@
 
   before_destroy :update_inventory_before_destroy
 
-  delegate :unit_description, to: :variant
+  delegate :product, :unit_description, to: :variant
```
sauloperez: why is this needed exactly?

Matt-Yorkley: Because on the original LineItem model, `#product` is defined as another association using `through: :variant`, and we've basically removed that association.

Matt-Yorkley:
I tried the `-> { with_deleted }` notation, and it threw fatal errors. I think we can't use it on associations until Rails 4.

RachL:
I checkout the product "pommes", then deleted the product.

All reports are still working properly.

The same test on Katuma staging has shown error 500.

This is ready to go :tada:

luisramos0:
🎉

PR #4035, opened 2019-07-12
4033 Remove line item adjustments when line item removed

### LINKED ISSUE #4033: Invalid cart after removing line item with line item adjustment
## Description

When shopping as customer, the cart becomes invalid when you remove a line item with a line item based adjustment, such as an enterprise fee.

Removing the line item results to an HTTP 500 error, then the line item adjustment is not deleted while the associated line item is already gone. These are some implications (there may be more):

* The order total still includes the line item adjustment that is no longer visible in the shopfront. This adjustment is still listed in the order page for admins.
* Clicking "Empty Cart" in the cart page fails.
* Clicking "Update" in the cart page fails.

The customer can still check out with the wrong order total.

Adding back the variant which was removed from the cart fixes this.

## Expected Behavior

Deleting a line item from the cart should also remove the adjustment.
 
## Actual Behaviour

Deleting a line item from the cart leaves behind the adjustment.

## Steps to Reproduce

As customer:

1. Add two variants with associated enterprise fees in the order cycle.
2. Remove the first line item. (HTTP 500)
3. Go to the checkout page and notice the wrong order total.

As admin:

1. Go to the order page for the order that is still in "Cart" state.
2. Notice that the adjustment for the removed line item is still there.
3. Notice that the order total includes this adjustment.

As customer:

1. Go back to the cart page.
2. Click on "Empty Cart". (HTTP 500)
3. Change the quantity for the remaining line item, and click on "Update". (HTTP 500)

## Context

I was looking into #4022.

## Severity

kristinalim:
#### What? Why?

- Closes #4033 

This is a simple solution to removing the line item adjustments left behind when the line items for which they are source have been deleted.

#### What should we test?

To set up the OC:

1. Add at least two variants to an OC.
2. Set up a coordinator, an incoming, and an outgoing _line item based_ enterprise fee. You can use "Flat Percent (per item)" or other calculators except the following:
      - `Spree::Calculator::FlatRate`
      - `Spree::Calculator::FlexiRate`
      - `Spree::Calculator::PriceSack`
      - `Spree::Calculator::PerItem` (See [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/4033#issuecomment-511062897) why.)

As customer:

1. Add two variants with associated enterprise fees in the order cycle.
2. Remove the first line item.
3. Go to the checkout page. Check that the order total is correct.

As admin:

1. Go to the order page for the order that is still in "Cart" state.
2. Check that the adjustments for the removed line item are not there.
3. Check that the order total does not include these adjustments.

As customer:

1. Go back to the cart page.
2. Change the quantity for the remaining line item, and click on "Update". Reload the page, and check that it worked.
3. Click on "Empty Cart". Check that it worked.

#### Release notes

- Delete line item adjustments when the line items for which they are source have been deleted. #4033

Changelog Category: Fixed

### DIFF
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -398,7 +398,7 @@ describe Spree::Order do
         expect(order.line_items(:reload).map(&:variant)).to eq([v2])
       end
 
-      pending "removes the variant's adjustment" do
+      it "removes the variant's adjustment" do
         line_item = order.line_items.where(variant_id: v1.id).first
         adjustment_scope = Spree::Adjustment.where(source_type: "Spree::LineItem",
                                                    source_id: line_item.id)
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -366,6 +366,8 @@ describe Spree::Order do
     before do
       order.add_variant v1
       order.add_variant v2
+
+      order.update_distribution_charge!
     end
 
     it "removes the variant's line item" do
@@ -377,6 +379,34 @@ describe Spree::Order do
       expect do
         order.remove_variant v3
       end.to change(order.line_items(:reload), :count).by(0)
+    end
+
+    context "when the item has an associated adjustment" do
+      let(:distributor) { create(:distributor_enterprise) }
+
+      let(:order_cycle) do
+        create(:order_cycle).tap do |record|
+          create(:exchange, variants: [v1], incoming: true)
+          create(:exchange, variants: [v1], incoming: false, receiver: distributor)
+        end
+      end
+
+      let(:order) { create(:order, distributor: distributor, order_cycle: order_cycle) }
+
+      it "removes the variant's line item" do
+        order.remove_variant v1
+        expect(order.line_items(:reload).map(&:variant)).to eq([v2])
+      end
+
+      pending "removes the variant's adjustment" do
+        line_item = order.line_items.where(variant_id: v1.id).first
+        adjustment_scope = Spree::Adjustment.where(source_type: "Spree::LineItem",
+                                                   source_id: line_item.id)
+        expect(adjustment_scope.count).to eq(1)
+        adjustment = adjustment_scope.first
+        order.remove_variant v1
+        expect { adjustment.reload }.to raise_error
+      end
     end
   end
 
--- a/app/models/concerns/line_item_based_adjustment_handling.rb
+++ b/app/models/concerns/line_item_based_adjustment_handling.rb
@@ -1,0 +1,8 @@
+module LineItemBasedAdjustmentHandling
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :adjustments_for_which_source, class_name: "Spree::Adjustment", as: :source,
+                                            dependent: :destroy
+  end
+end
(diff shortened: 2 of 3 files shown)


Matt-Yorkley:
Wow! That was a quick turnaround. :clap:

kristinalim:
This doesn't apply to `Spree::Calculator::PerItem`. See [discussion](https://github.com/openfoodfoundation/openfoodnetwork/issues/4033#issuecomment-511034861).

spec/models/spree/order_spec.rb
```diff
@@ -378,6 +380,34 @@
...
+      end
+
+      it "removes the variant's adjustment" do
+        line_item = order.line_items.where(variant_id: v1.id).first
+        adjustment_scope = Spree::Adjustment.where(source_type: "Spree::LineItem",
+                                                   source_id: line_item.id)
+        expect(adjustment_scope.count).to eq(1)
+        adjustment = adjustment_scope.first
+        order.remove_variant v1
+        expect { adjustment.reload }.to raise_error
```
Matt-Yorkley: It's still expected to raise an error and not remove the adjustment?

kristinalim: @Matt-Yorkley Oh sorry - I had not specified what error. It's supposed to raise `ActiveRecord::RecordNotFound` because the adjustment had already been deleted.

RachL:
So :

1. I used "sanity farm" and transformed it as a hub that sell Fred's product as well.
2. I've added the fee on sanity farm as a distributor / coordinator
3. As a customer, removing the first line item correctly updated the total at checkout:
image

4. As an admin, the fees were not there anymore:
image

5. Back as a customer, I can empty the cart, but I can't update quantity anymore. Reloading the page does not change anything. Where is the update button that used to be there?

French staging:
image

Katuma staging (where the PR is):
image

ping @kristinalim @luisramos0 ?

luisramos0:
I think I know what's going on @RachL and it may not be related to this PR. I think it's a separate bug.
The order in katuma is in "address" state, not in "cart", that's why the update button is not showing.

I think it's a bug in the admin page, when you go there the order workflow is automatically pushed forward as much as possible, this is spree logic... because you went to the backoffice, the order/cart was moved from cart to address state...
I think we need to prevent this, at least for orders in the cart state...

RachL:
@luisramos0 you're right, I could reproduce it. So if this is not linked to the PR, let's move forward and merge it.

We will need another issue for this but I wonder if it is not an edge case: like when would a buyer want to update a cart after the manager already did something on it in the admin?

luisramos0:
yes, that's true, maybe an S4.

kristinalim:
I checked this, and want the issue to be at least an S3. @luisramos0 @RachL You might disagree.

For this to happen, you don't need the store owner to update the order - even just loading the order page transitions the new order from "cart" to "address". Once the order is in "address" state:

1. As admin, I cannot void the order in the admin section.
2. As customer, clicking the cart icon results in HTTP 500.
3. As customer, going directly to `/cart` works, but clicking the "Checkout" button there also results in HTTP 500.
4. If the customer logs out and signs back, they are using the same order/cart so they have the same issue.

@RachL Just let me know if you prefer I create the issue.

RachL:
@kristinalim yes go for it, it seems you have more insight. Thank you!

luisramos0:
thanks Kristina.
My assumptions for S4 was that the cart would render without the update button (as per Rachel's initial report and screenshot). you must be talking about a slightly different case where you get a 500 in the cart, right? can you share the stack trace of that 500?

PR #4043, opened 2019-07-15
Bring spree/admin/orders/shipment and shipment_manifest partials from spree_backend and fix an authorization bug

### LINKED ISSUE #4038: Managers are not able to delete line items from orders anymore
## Description
Hub managers should be able to delete line items from orders in the order edit page. This was possible in v1, and it is not in v2 anymore.

## Steps to Reproduce
1. Login with rachel.arnould[at]openfoodfrance.org on French production (v1.31)
2. Go to `/admin/orders/R854846648/edit` you will see this:
image

3. Login with the same account on French staging (v2.2)
4. Go to `/admin/orders/R854846648/edit` you will see this:
image

## Severity

## Your Environment
* Version used: see above
* Browser name and version: Firefox 67 desktop on Ubuntu 19.10

luisramos0:
#### What? Why?

Closes #4038 

#4038 is caused by a bug in Spree 2.0 that is fixed in Spree 2.4: https://github.com/spree/spree/pull/5624 

This fix switched the broken authorization on shipment.manifest.item (an OpenStruct) to the shipment.manifest.item.line_item. But shipment.manifest.item.line_item is only introduced in spree v2.2 [here](https://github.com/spree/spree/commit/bc47452562d88b21393890c95b322852720c3e00).
So, in this PR I am switching the authorization from shipment.manifest.item to simply shipment.

Additionally, this was the perfect opportunity to bring both shipment and shipment_manifest partials to OFN and convert them to haml.

#### What should we test?
Verify that the issue is fixed.

Because we are bringing these partials from spree we need to test them.
We need to verify that all features in this block work as expected:
Screenshot 2019-07-15 at 17 35 32

1. Verify that the title of the shipment is correct in different shipment states:
Screenshot 2019-07-15 at 17 33 58
2. Verify that the list of line items and its fields is correct
2.1 Verify that different variations of the variant type (unit, weight, etc) will display correctly in this page
3. Verify editing line item quantity
4. Verify adding new line items to the order
5. verify deleting line items from the order
6. Verify show and edit shipping method
7. Verify show and edit tracking number

#### Release notes
Changelog Category: Fixed
Fixed authorization bug in order edit page, managers can now edit their hub's orders.


RachL:
@luisramos0 I've noticed that the shipment state is always "by default" but I guess it should change depending on the shipping category right? Or is it another "default"?

Is this part of this PR as well or should I open another issue if this should indeed change depending on the shipping categories? So far it does not.

Notes: https://docs.google.com/document/d/1x0VFtKpyHkUziwbcZMBi0mXCYDT7QOwLlpK_5gLznk0/edit#

luisramos0:
that default is the stock location name and it will always be default...
I believe the translation of from is wrong, should be "de" not "par" 
the shipment state is "en attente"

because we only have one stock location called default and there will only be always only one shipment in the order, this line is not ideal...

currently the line is made up of:
- shipment code - shipment state - "packaged from" - stock location name
- H43534543 - en attente - "conditionnement par" - default

what if we keep the code of the shipment and state but remove the rest?
it's a very very easy change I can do on this PR. the header would become:

- shipment code - shipment state
- H43534543 - en attente

RachL:
@luisramos0 yes this makes more sense! Thank you :)

luisramos0:
I removed the two lines from the template.
I'd skip code review for these two lines and stage it right away ;-)

RachL:
Awesome! Moving to ready to go

Issue #4054, opened 2019-07-18
[API clean up] Make OFN independent of spree_api


luisramos0:
### What we should change and why (this is tech debt)
We need to drop spree_api dependency and bring what's needed to OFN.
Although there are quite a few controllers in spree_api that are not overridden in OFN, most of then will either not be applicable to OFN because of differences in models (like stock_transfers) or not funcional in OFN because of differences in processes (like checkouts_controller).

There are 24 controllers in spree_api:
- 5 are overridden in OFN and most probably working: line_items_controller_decorator, products_controller_decorator, shipments_controller_decorator, users_controller, variants_controller_decorator

The rest are all not overriden in OFN:

- 7 are most probably working: countries_controller, states_controller, zones_controller, taxonomies_controller, taxons_controller, images_controller, addresses_controller
- 4 are maybe functional but remotely useful for OFN right now: inventory_units_controller, option_types_controller, option_values_controller, return_authorizations_controller
- 4 are not functional and would need work to be functional: checkouts_controller (due to ofn checkout customizations), product_properties_controller, properties_controller (due to cache control, at least), payments_controller (due to adjustments)
- 4 are not applicable or not usable in OFN:  stock_locations_controller, stock_movements_controller, stock_items_controller (doesnt make sense to expose this level of granularity in the OFN api, stock levels are managed at variant level), orders_controller (OFN has it's own /api/orders_controller)

From this list we can already see that, **most probably**, we need to bring the 7 working ones to OFN and probably none of the other ones are needed to keep OFN working as is without spree_api.

The major advantage is that we will have the OFN api in a single place and the extra advantage is that this will enable us to convert the views of these 7 api resources from rabl to ams and thus have the OFN API build with one single uniform tech stack.

At some point in the future we will need to adapt (or rebuild!) checkout and payments to OFN so that we offer the ability for external apps to checkout using the OFN API. For now, we can ignore this and just keep the functionality we have.

### Context
This comes as part of the OFN API clean up process started with it's documentation in #3001

### Impact and timeline
This will need to be done before we clean up the api side of OFN, it will make things a lot easier because we are basically removing a lot of code in spree that we either dont use or is not working.

PR #4055, opened 2019-07-18
Remove dependency to spree_api - step 1 - controllers and routes

luisramos0:
#### What? Why?

Starts #4054

In this PR we bring from spree_api the controllers that are used in OFN as is. There may be more controllers we will want to expose in the OFN API in the future like the basic ones countries and states. We can do that in a second step.

One of the best things about this PR (it took me quite a while to make all the specs pass) was to bring tests from spree_api and mix them with existing tests on OFN side: this makes explicit that the API is currently not consistent, for example, the products and variants endpoints.

I fixed quite a lot of rubocop issues and I am leaving a few of the more complex.

#### What should we test?
These are the areas that need to be verified:
- create user account and edit user in backoffice
- BOM - list, add, edit and delete items
- Bulk Product management - test the list of products is correct, delete products, clone products, add, edit and remove variants
- Inventory page - test the list of products is correct and then basic features of adding and removing variants
- Order edit page - add, edit and delete line items from order. Make sure that enterprise fees are updated.
- Config/Product categories - create new category entry, use entry in product
- Product page - create product with specific category and then edit product and change that category
- Shopfront - filter products by category

#### Release notes
Changelog Category: Changed
OFN is no longer dependent on the controllers of the Spree API, this is a necessary step to make the OFN API consistent, currently the spree endpoints and the spree endpoints are not consistent.


app/controllers/spree/api/base_controller.rb
```diff
@@ -0,0 +1,130 @@
+require_dependency 'spree/api/controller_setup'
+
+module Spree
+  module Api
+    class BaseController < ActionController::Metal
+      include Spree::Api::ControllerSetup
```
sauloperez: I know this is an exact copy of the line in the Spree codebase and I don't suggest we change it here but for everyone to be aware, it is redundant as this could be written as `include ControllerSetup`. I'd like to remove these and other redundancies once we own the code completely...

app/helpers/spree/api/api_helpers.rb
```diff
@@ -0,0 +1,112 @@
+module Spree
+  module Api
+    module ApiHelpers
+      def required_fields_for(model)
```
sauloperez: This really looks like a duplicate of Rails 4's [strong parameters](https://github.com/rails/strong_parameters) :see_no_evil:

app/views/spree/admin/shared/_routes.html.erb
```diff
@@ -3,11 +3,6 @@
     :variants_search       => spree.admin_search_variants_path(:format => 'json'),
     :taxons_search         => spree.api_taxons_path(:format => 'json'),
     :user_search           => spree.admin_search_users_path(:format => 'json'),
-    :product_search        => spree.api_products_path(:format => 'json'),
-    :option_type_search    => spree.api_option_types_path(:format => 'json'),
-    :states_search         => spree.api_states_path(:format => 'json'),
-    :orders_api            => spree.api_orders_path(:format => 'json'),
-    :stock_locations_api   => spree.api_stock_locations_path(:format => 'json'),
-    :variants_api          => spree.api_variants_path(:format => 'json')
```
sauloperez: where do these go?

config/routes/api.rb
```diff
@@ -0,0 +1,33 @@
+Openfoodnetwork::Application.routes.draw do
+  namespace :api do
```
sauloperez: :heart_eyes:

config/routes/spree.rb
```diff
@@ -79,6 +92,13 @@
         end
       end
     end
+
+    resources :users do
+      member do
+        put :generate_api_key
+        put :clear_api_key
+      end
+    end
```
sauloperez: and why these new routes if we are only moving API controllers :thinking: ?

app/controllers/spree/api/base_controller.rb
luisramos0: and I hope soon this will be under only Api:: ;-)
Not long before we can kill the Spree namespace altogether in routes, controllers and views!!!

after spree_frontend is gone, the checkout (and orders, etc) controllers are ready to be moved from controller/spree to controllers/ and now the spree/api stuff is the same and can be moved from controller/spree/api to controllers/api 🎉

app/views/spree/admin/shared/_routes.html.erb
luisramos0: they are not needed, these are JS helpers for spree js code that we dont use.

config/routes/spree.rb
luisramos0: these non-api routes are related to api and are defined in spree_api [here](https://github.com/openfoodfoundation/spree/blob/fe0a1311abb097bf3ac8001a24246c6cff5ecdbb/api/config/routes.rb#L3)
and the code also comes from spree_api [here](https://github.com/openfoodfoundation/spree/blob/master/api/app/models/spree/user_decorator.rb)

luisramos0:
thanks a lot for you quick feedback @sauloperez

mkllnk:
@lin-d-hop Can you list all API endpoints that are currently used or you are planning to use with Zapier? It would be good to double-check that we are not removing something that is in use. We should add specs for these endpoints then.

app/controllers/spree/api/base_controller.rb
sauloperez: thanks for leading this @luisramos0 !

luisramos0:
ok, thanks guys.
I added a few more rubocop fixes and regenerated rubocop_todo

luisramos0:
Rebased to resolve conflicts.

sigmundpetersen:
So this should probably be restaged after rebase @RachL

luisramos0:
ah, yes @sigmundpetersen :+1:

RachL:
Alright. Everything is looking good EXCEPT adding product category as super admin, which is broken in both Firefox and Chrome:

image


image

It is working both in production and others stagings, so it clearly comes from this PR. Will ask on slack on the best approach here.

Testing notes: https://docs.google.com/document/d/1Tc9XwF3SbZUdZ4eA90-RAccdD27aD64b7WoaUFCpXSk/edit#

luisramos0:
The fix is very simple and does not affect any other parts of the app so I decided to add it to this PR.
I request a code review just for [this last commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/4055/commits/b9ddb39edccbe326a5ca23ce690154b2fc3fb38e).
I dont think we need to test everything again, just the taxonomies page.

daniellemoorhead:
@mkllnk does this need both you and Pau to review/it, given you both gave approval before? Can Kristina replace Pau as the approver?

sigmundpetersen:
Yes Pau approved it initially and the re-review is just for the additional commit after issue found in testing. So I think it's OK for @kristinalim to replace Pau for the re-review 👍

luisramos0:
awesome, test ready :tada:

for testers: I dont think we need to test everything again, just the taxonomies page.

RachL:
Taxonomies are back in business :muscle:

PR #4063, opened 2019-07-23
Remove dead code under views/spree/shared

luisramos0:
#### What? Why?

Remove 3 unused view templates related to the shopfront.

#### What should we test?
Verify that the shopfront works as expected: the list of products renders correctly, the filters/search work.
Verify the checkout works in particular the entry of addresses.
I don't think there will be anything missing.

#### Release notes
Changelog Category: Removed
Removed unused code from the frontoffice.


daniellemoorhead:
Hey @luisramos0 I staged this on UK staging and tested. The shopfront rendered fine, and search/filters are fine, but when I tried to go to the checkout I got the snail. I tried via the dropdown cart menu and also via the edit cart screen, both times straight to the snail.

Do you think it's this PR causing the problem?

RachL:
@daniellemoorhead can you provide the link to the shopfront you've used? Also don't forget to check the devtools, especialy the network tab (pressing F12 on your keyboard) sometimes it can give a little more insights :)

daniellemoorhead:
Apologies for lack of detail, I've recorded it all on a testing notes doc - https://docs.google.com/document/d/1Fq2GVt62GkLTIz6QEj-UHenPAx-YcZtqe3kInpHiFNs/edit?usp=sharing

daniellemoorhead:
Am going to try and test again...let's see what happens with another shop.

daniellemoorhead:
Test passed successfully this second time around. 

Testing notes: https://docs.google.com/document/d/1Fq2GVt62GkLTIz6QEj-UHenPAx-YcZtqe3kInpHiFNs/edit?usp=sharing

This is ready to go!

mkllnk:
Does it still fail in the first shop though? If you can make it fail, we should find out why and either file another bug or fix this pull request.

daniellemoorhead:
It didn't fail this time, no. And I ran through it on the same shopfront a
couple of times.

I can't retry with that same original shop because the order cycle closed :(

I can try and break it with a few more other shopfronts if that seems worth
doing @mkllnk <https://github.com/mkllnk>?

mkllnk:
> I can't retry with that same original shop because the order cycle closed :(

Can you change the close date of the order cycle to open it again? I'm tempted to just ignore it but a broken checkout sounds scary.

daniellemoorhead:
I won't have time to do this until next week @mkllnk
<https://github.com/mkllnk> so if the ticket can wait till then I will try
and figure out how to get into that shop and change it (never done anything
super admin-ish before but it's a good thing to learn sooner than later.

luisramos0:
Moving to test ready as it's better if we retest. It can wait for that.

RachL:
I manage to reproduce the snail on UK staging.

Shop : 
https://staging.openfoodnetwork.org.uk/new-dawn-traders/shop 

Here is the products I have chosen :
image

Here is the snail error 500:

image

@luisramos0 @mkllnk if I stage something else and the error 500 stays on this shop, is that proof enough it does not come from this PR?

mkllnk:
Yes. Some PRs could change the database layout or produce bad data that can persist. But this PR doesn't affect the database at all. So staging something else and retesting is good enough. You can also stage master to be sure that the other pull request isn't introducing the same bug as well.

RachL:
Same errors with master on that shop. This PR is then ready to go! Thank you @mkllnk

Issue #4060, opened 2019-07-22
Replace all rabl files with AMS serializers


luisramos0:
### What we should change and why (this is tech debt)
AMS is the way to serialize objects in OFN. We have a few rabl files that need to be converted, time to finish up the [clean up started 5 years ago](https://github.com/openfoodfoundation/openfoodnetwork/issues/280#issuecomment-61745824).

Currently, this is the list of rabl files we have:

- [x] ./app/views/admin/json/_enterprises.rabl #4059 
- [X] ./app/views/json/_order_cycle.rabl #4068
- [X] ./app/views/json/_flash.rabl #4059 
- [X] ./app/views/json/_groups.rabl #4059 
- [X] ./app/views/json/_taxon.rabl #4059 
- [X] ./app/views/json/_hubs.rabl #4059 
- [X] ./app/views/json/_current_user.rabl #4059 
- [X] ./app/views/json/_current_order.rabl #4059 
- [X] ./app/views/json/_enterprises.rabl #4059 
- [X] ./app/views/json/_producers.rabl #4059 
- [X] ./app/views/json/partials/_address.rabl #4059 
- [X] ./app/views/json/partials/_enterprise.rabl #4059 
- [X] ./app/views/json/partials/_hub.rabl #4059 
- [X] ./app/views/json/partials/_producer.rabl #4059 
- [X] ./app/views/json/_producer.rabl #4059 
- [X] ./app/views/spree/admin/variants/search.rabl #4064
- [ ] ./app/views/spree/admin/taxons/search.rabl
- [x] ./app/views/spree/api/variants/bulk_index.v1.rabl #4101
- [x] ./app/views/spree/api/variants/bulk_show.v1.rabl #4101
- [x] ./app/views/spree/api/products/bulk_index.v1.rabl #4101
- [x] ./app/views/spree/api/products/bulk_show.v1.rabl #4101
- [x] ./app/views/spree/api/users/authorise_api.v1.rabl #4111 (now in #4057)
- [x] ./app/views/api/product_images/update_product_image.v1.rabl #4073 
- [x] ./app/views/api/enterprises/bulk_index.v1.rabl #4059 
- [x] ./app/views/api/enterprises/bulk_show.v1.rabl #4101
- [x] ./app/views/api/order_cycles/bulk_index.v1.rabl #4059 
- [x] ./app/views/api/order_cycles/bulk_show.v1.rabl #4059 

### Context
[...]

PR #4064, opened 2019-07-23
Replace views/admin/variants/search.rabl with Api::Admin::VariantSerializer

luisramos0:
#### What? Why?

Part of  #4060

Converging what was in search.rabl into Api::Admin::VariantSerializer.

I fixed some rubocop issues, fixed a bug with the images code and simplified options_types and values display.

But mostly this is just converting and getting everything together, not reducing complexity.

But getting everything together makes it very clear the mess we have around variant names and unit names with 8 fields... :name, :options_text, :unit_value, :unit_description, :unit_to_display, :display_as, :display_name, :name_to_display ⚠️ 

#### What should we test?
Adding a variant to an order when: creating/editing an order and creating/editing a subscription.
Test the auto complete feature carefully (search for variants), make sure the fields for each result in the select box are correct and that the products can be added to the order and/or subscription.

#### Release notes
Changelog Category: Changed
Clean up old Spree code of search box for variants when adding variants to orders and subscriptions.


app/views/spree/admin/variants/_autocomplete.js.erb
```diff
@@ -16,16 +16,13 @@
       </ul>
       <ul class='variant-data'>
         <li class='variant-sku'><strong><%= t('admin.sku') %>:</strong> {{variant.sku}}</li>
-        <li class='variant-on_hand'><strong><%= t('on_hand') %>:</strong> {{variant.on_hand}}</li>
+        {{#if variant.on_demand }}
```
sauloperez: I have no idea what this `#if ` syntax is and the same time it's a JS file with ERB in it? :astonished:

luisramos0: this is an if clause in erb.
this is not a "JS file with erb" this is an "erb file with JS". the output of this erb file is a JS template (script type='text/template') that is injected by this complex JS code: https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/backend/app/assets/javascripts/admin/variant_autocomplete.js.erb

this should be a haml template under app/assets that angular code inject...
this PR solved this issue #3440 partially but the JS code still needs to be changed, but it's complicated, see the link above...

app/assets/stylesheets/admin/pages/subscription_line_items.css.scss
```diff
@@ -0,0 +1,11 @@
+#add-line-item {
+  fieldset {
+    td.vertical_align_top {
```
mkllnk: This is definitely an improvement and semantic naming would be even better. Or in this case, do you even need the class? Is the `#add-line-item fieldset td` selector enough?

spec/serializers/api/admin/variant_serializer_spec.rb
```diff
@@ -1,7 +1,30 @@
...
-  it "serializes a variant" do
+
+  it "serializes the variant name" do
     serializer = Api::Admin::VariantSerializer.new variant
+
+    expect(serializer.to_json).to match variant.name
+  end
+
+  it "serializes the variant options" do
+    serializer = Api::Admin::VariantSerializer.new variant
```
mkllnk: What about let?

```rb
let(:serializer) { Api::Admin::VariantSerializer.new(variant) }
let(:json) { serializer.to_json }
```

app/assets/stylesheets/admin/pages/subscription_line_items.css.scss
luisramos0: the third column with the button cannot have the class.
and I removed td in the following commit when reviewing rubocop issues.

sauloperez:
There's a valid Code Climate comment that should be considered

luisramos0:
ok @sauloperez, I changed to class  but kept the id as I am not sure it's not used in spree css/js

RachL:
@luisramos0 I used the Duhat release shop on French staging and created order as manager, removing and editing quantities. 

However for subs, after creating a new subs for "Duhat release" I cannot reached the `/admin/subscriptions` page anymore. There is an infinite spinner and an error 500 in the dev tools.

I deleted a variant before starting to test subs. maybe it is related? I didn't select this variant in the dropdown though (I wanted to test if the product appeared or not in the dropdown). However I selected a variant that is linked to the same master product (Rhubarbe en vrac).

The other shops still can access their /subs page so there is something going on with the data from this shop.

Can this be linked to this PR? Otherwise I think we are good to go. I could create and edit subs on other shops.
Super admin access are shared in bitwarden.

luisramos0:
The problem is related to a deleted variant Rhubarbe - Barquette and this variant is in a subscription. So, it's not related to this PR.
If you delete a variant that is used in a subscription, you break the subscriptions list for that shop...

PR #4067, opened 2019-07-23
Add icon to OC menu entry

luisramos0:
#### What? Why?

After moving the menu items we realize it's better to add an icon to the OC menu.
Before:
Screenshot 2019-07-23 at 18 42 23
After:
Screenshot 2019-07-23 at 18 49 08

I am not sure how to add a new icon to the list... so we re-use the same icon from products as a quick improvement.

#### What should we test?
Check the icon and the menu renders correcyly.

#### Release notes

Changelog Category: Changed
Re-organized main menu order, Order Cycles is now just after Products. The Order Cycles menu has an icon now.

### DIFF
--- a/app/views/spree/admin/shared/_tabs.html.haml
+++ b/app/views/spree/admin/shared/_tabs.html.haml
@@ -1,6 +1,6 @@
 = tab :dashboard, :route => :admin, :icon => 'icon-dashboard'
 = tab :products , :option_types, :properties, :prototypes, :variants, :product_properties, :taxons, :url => admin_products_path, :icon => 'icon-th-large'
-= tab :order_cycles, :url => main_app.admin_order_cycles_path
+= tab :order_cycles, :url => main_app.admin_order_cycles_path, :icon => 'icon-refresh'
 = tab :orders, :payments, :creditcard_payments, :shipments, :credit_cards, :return_authorizations, :url => admin_orders_path('q[s]' => 'completed_at desc'), :icon => 'icon-shopping-cart'
 = tab :reports, :icon => 'icon-file'
 = tab :configurations, :general_settings, :mail_methods, :tax_categories, :zones, :states, :payment_methods, :inventory_settings, :taxonomies, :shipping_methods, :trackers, :label => 'configuration', :icon => 'icon-wrench', :url => edit_admin_general_settings_path


sauloperez:
Wouldn't it be better to wait for Yuko's feedback here? I don't what we would get with a temporary icon.

RachL:
@sauloperez it was just to have a cleaner separation between the words "product" and "order cycles". Yes this require a deeper work, but the question was - given the new order - is it best to leave it without icon or with an icon.

What do you think?

app/views/spree/admin/shared/_tabs.html.haml
```diff
@@ -1,6 +1,6 @@
 = tab :dashboard, :route => :admin, :icon => 'icon-dashboard'
 = tab :products , :option_types, :properties, :prototypes, :variants, :product_properties, :taxons, :url => admin_products_path, :icon => 'icon-th-large'
-= tab :order_cycles, :url => main_app.admin_order_cycles_path
+= tab :order_cycles, :url => main_app.admin_order_cycles_path, :icon => 'icon-th-large'
```
mkllnk: I found this class in `spree-46d6f8f5fd43/backend/app/assets/stylesheets/admin/plugins/_font-awesome.scss`. Maybe `.icon-repeat` or `.icon-refresh` would work better?

luisramos0: awesome Maikel! I switched to icon-refresh:
Screenshot 2019-07-26 at 10 52 47

It's not like amazingly appropriate, but I think it's better :-D

luisramos0: @sauloperez @RachL ☝️ ok?

sauloperez: Better, yes!

RachL:
@luisramos0 so the thing is, we now have a longer menu for super admin which has different behavior across languages.

FR as super admin:
image

IT as super admin:
image

FR as manager:
image

IT as manager:
image

So there is a translation error in IT. I will stage it to Katuma and UK now to see other languages.

If there is only super admin which has a display issue, do we still go forward with this or shall we close and wait for a bigger work on our UI?

ping @lin-d-hop @sauloperez

luisramos0:
@RachL I tried super admin FR in master 
Screenshot 2019-07-30 at 12 47 44
and I also see the two lines menu. So, the problem is already in master. That means we should probably merge this PR anyway.

I think it was the previous PR that introduced the problem by changing the order of the menu, because MORE was including the longer Cycles de Vente and now it includes the shorter Groupes?

RachL:
:face_with_head_bandage: you are right... moving to ready to go then... Will try to figure something out in translations.

PR #4068, opened 2019-07-24
Convert order_cycle.rabl to Api::OrderCycleSerializer

luisramos0:
#### What? Why?

Part of #4060

We replace order_cycle.rabl with Api::OrderCycleSerializer and we use the serializer in both the haml view and the controller.

#### What should we test?
- change OC in the frontoffice and check product list updates correctly
- change from one shop to another in the frontoffice and make sure that, on the shop page you change to, the OC is correctly selected if there is only or the select box is empty if there is more than one open OC for that shop. For each OC selection make sure the product list is displaying correctly
- create and edit order cycles in the backoffice

#### Release notes
Changelog Category: Changed
Converted some of the last files in OFN in RABL to AMS.

### DIFF
--- a/app/controllers/admin/tag_rules_controller.rb
+++ b/app/controllers/admin/tag_rules_controller.rb
@@ -9,8 +9,8 @@ module Admin
     def map_by_tag
       respond_to do |format|
         format.json do
-          serialiser = ActiveModel::ArraySerializer.new(collection)
-          render json: serialiser.to_json
+          serializer = ActiveModel::ArraySerializer.new(collection)
+          render json: serializer.to_json
         end
       end
     end
--- a/app/controllers/shop_controller.rb
+++ b/app/controllers/shop_controller.rb
@@ -27,12 +27,12 @@ class ShopController < BaseController
       if oc = OrderCycle.with_distributor(@distributor).active.find_by_id(params[:order_cycle_id])
         current_order(true).set_order_cycle! oc
         @current_order_cycle = oc
-        render partial: "json/order_cycle"
+        render json: @current_order_cycle, serializer: Api::OrderCycleSerializer
       else
         render status: :not_found, json: ""
       end
     else
-      render partial: "json/order_cycle"
+      render json: current_order_cycle, serializer: Api::OrderCycleSerializer
     end
   end
 
--- a/app/helpers/injection_helper.rb
+++ b/app/helpers/injection_helper.rb
@@ -62,7 +62,9 @@ module InjectionHelper
   end
 
   def inject_current_order_cycle
-    inject_json_ams "orderCycleData", current_order_cycle, Api::OrderCycleSerializer
+    serializer = Api::OrderCycleSerializer.new(current_order_cycle)
+    json = serializer.object.present? ? serializer.to_json : "{}"
+    render partial: "json/injection_ams", locals: { name: "orderCycleData", json: json }
   end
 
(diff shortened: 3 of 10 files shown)


app/controllers/admin/tag_rules_controller.rb
```diff
@@ -9,8 +9,8 @@ class TagRulesController < ResourceController
     def map_by_tag
       respond_to do |format|
         format.json do
-          serialiser = ActiveModel::ArraySerializer.new(collection)
-          render json: serialiser.to_json
+          serializer = ActiveModel::ArraySerializer.new(collection)
```
mkllnk: Oh no, I like the UK spelling! :stuck_out_tongue: I wish we could convince the Rails team to switch to Serialisers...

Oh well, this is probably a good move then. Does that mean we will switch to American English in the whole code base for consistency? And documentation? No! :fearful:

luisramos0: tbh all I care about is consistency, I think we must be consistent in each word. We must not have a word spelled in two different ways in our codebase.

From my time coding in London with English and Australians devs I remember well that Australians were a lot more pro British spelling than the English themselves :-)

RachL:
Everything is looking good. Except one thing that I'm not sure has been introduced by this PR, but I'm recording this here:

- A while ago, I create "Nanny Ogg's farm" on Katuma's staging. Created OC and checkout products. All fine.
- While testing this PR, I've opened an OC for Nanny Ogg's farm as super admin.
- However the shop remained invisible from the shop list althought both product had on hand values. When visiting the shop I could see this:

image

As we discovered here : https://openfoodnetwork.slack.com/archives/CG7NJ966B/p1564466046014400 this happens when no shipping or payment method are there
- However as super admin I can see that Nanny Ogg has both
- So I loggedout and went as manager of the Nanny Ogg farm
- The /enterprise page told me I don't have payment method (red danger sign)
- But when looking at the payment method page I can see that there is a "cash" payment method. 
- So I've just hit "update" without changing anything and everything was fixed...

Any idea what can cause this and if it can be linked to this PR? @mkllnk @sauloperez I'm tagging you as you've reviewed the PR and because Luis' is on holiday this week.

app/controllers/admin/tag_rules_controller.rb
mkllnk: Yeah, and Germans living in Australia are even worse! :wink:

mkllnk:
I looked at the changes again and I couldn't find anything that could cause this problem. I'll put this in ready-to-go. @sauloperez Please merge if you agree.

sauloperez:
I don't see any changes that could affect that either so merging. I create a separate issue for that.

However, I don't fully understand what you mean by

> However the shop remained invisible from the shop list althought both product had on hand values. When visiting the shop I could see this:

what should be actually see in that screenshot you shared @RachL ?

RachL:
@sauloperez in the screenshot you see in the right corner it says "Orders are closed". + the shop didn't appeared in the list of opened shops.
This is happening if you don't set up a payment method. But in this case there was one. I needed to recent the payment method form for it to be taken into account.

luisramos0:
I dont think it's related to this PR but there is a mystery in the payment methods page:
@RachL said: "But when looking at the payment method page I can see that there is a "cash" payment method. So I've just hit "update" without changing anything and everything was fixed..."

PR #4073, opened 2019-07-24
Replace update_prod_image.rabl with ImageSerializer in api/product_image_controller

luisramos0:
#### What? Why?

Part of #4060 

Removing rabl file and using existing ImageSerializer.

A summary of my problem and Maikel's discovery: the Spree::Api::BaseController has [the Rabl responder activated](https://github.com/openfoodfoundation/spree/blob/8a8585a43cd04d1a50dc65227f337a91b18d66d5/api/app/controllers/spree/api/base_controller.rb#L10) because spree_api only uses rabl and the OFN Api::BaseController [does have AMS activated](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/api/base_controller.rb#L1-L11). So, as we convert rabl to AMS, we need to [switch the base_controller from Spree::Api::BaseController to Api::BaseController](https://github.com/openfoodfoundation/openfoodnetwork/pull/4073/commits/11a77043eb66f6b0e6548924a5472fffff4d0ca2) 💪

#### What should we test?
Upload product image in the bulk product edit page.

#### Release notes
Changelog Category: Changed
Converted one more rabl file to AMS (modernizing tech stack) when user uploads product image.

#### ARCHIVE
Initial report when I was blocked:
This is blocking all progress as I cant figure out how to convert these rabl template controllers to ams serializers.

Switching from respond_with for rabl under app/views to render :json with a serializer under app/serializers.
This works perfectly in an non-api controller: https://github.com/openfoodfoundation/openfoodnetwork/pull/4064/files#diff-b6e332682a5b8fba45654e386d82d12eR21
and the same does not work in an api controller, see change in this PR.

Any thoughts?

The build error is:
```
#<ActionView::MissingTemplate: Missing template api/product_images/update_product_image, spree/api/base/update_product_image with {:locale=>[:en], :formats=>[:js, :html], :handlers=>[:erb, :builder, :coffee, :haml, :rabl], :versions=>[:v1]}. Searched in:
  * "/Users/luisramos/dev/ofn/openfoodnetwork/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/bundler/gems/spree-8a8585a43cd0/api/app/views"
  * "/Users/luisramos/dev/ofn/openfoodnetwork/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/foundation-rails-5.5.2.1/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/gmaps4rails-1.5.6/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/bundler/gems/better_spree_paypal_express-27ad7165ea4c/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/bundler/gems/spree_auth_devise-0181835fb6ac/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/devise-2.2.8/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/bundler/gems/spree-8a8585a43cd0/backend/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/bundler/gems/spree-8a8585a43cd0/api/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/bundler/gems/spree-8a8585a43cd0/core/app/views"
  * "/Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/kaminari-0.14.1/app/views"
  * "/Users/luisramos/dev/ofn/openfoodnetwork/engines/web/app/views"
```

### DIFF
--- a/spec/features/admin/bulk_product_update_spec.rb
+++ b/spec/features/admin/bulk_product_update_spec.rb
@@ -742,7 +742,7 @@ feature '
     end
   end
 
-  describe "Updating product image with new upload interface" do
+  describe "Updating product image" do
     let!(:product) { create(:simple_product, name: "Carrots") }
 
     it "displays product images and image upload modal" do
@@ -755,6 +755,7 @@ feature '
 
         # Shows default image when no image set
         expect(page).to have_css "img[src='/assets/noimage/mini.png']"
+        @old_thumb_src = page.find("a.image-modal img")['src']
 
         # Click image
         page.find("a.image-modal").click
@@ -780,7 +781,7 @@ feature '
       within "table#listing_products tr#p_#{product.id}" do
         # New thumbnail is shown in image column
         @new_thumb_src = page.find("a.image-modal img")['src']
-        expect(@old_thumb_src) != @new_thumb_src
+        expect(@old_thumb_src).not_to eq @new_thumb_src
 
         page.find("a.image-modal").click
       end
--- a/app/controllers/api/product_images_controller.rb
+++ b/app/controllers/api/product_images_controller.rb
@@ -8,11 +8,11 @@ module Api
 
       if @product.images.first.nil?
         @image = Spree::Image.create(attachment: params[:file], viewable_id: @product.master.id, viewable_type: 'Spree::Variant')
-        respond_with(@image, status: 201)
+        render json: @image, serializer: ImageSerializer, status: :created
       else
         @image = @product.images.first
         @image.update_attributes(attachment: params[:file])
-        respond_with(@image, status: 200)
+        render json: @image, serializer: ImageSerializer, status: :ok
       end
     end
   end
--- a/app/controllers/api/product_images_controller.rb
+++ b/app/controllers/api/product_images_controller.rb
@@ -1,5 +1,5 @@
 module Api
-  class ProductImagesController < Spree::Api::BaseController
+  class ProductImagesController < BaseController
     respond_to :json
 
     def update_product_image
--- a/app/serializers/api/image_serializer.rb
+++ b/app/serializers/api/image_serializer.rb
@@ -1,8 +1,16 @@
 class Api::ImageSerializer < ActiveModel::Serializer
-  attributes :id, :alt, :small_url, :large_url
+  attributes :id, :alt, :thumb_url, :small_url, :image_url, :large_url
+
+  def thumb_url
+    object.attachment.url(:mini, false)
+  end
 
   def small_url
     object.attachment.url(:small, false)
+  end
+
+  def image_url
+    object.attachment.url(:product, false)
   end
 
   def large_url
--- a/app/views/api/product_images/update_product_image.v1.rabl
+++ b/app/views/api/product_images/update_product_image.v1.rabl
@@ -1,5 +1,0 @@
-object @image
-attributes(*image_attributes)
-attributes :viewable_type, :viewable_id
-node( :thumb_url ) { @product.images.first.attachment.url(:mini) }
-node( :image_url ) { @product.images.first.attachment.url(:product) }


luisramos0:
@sauloperez on slack:
it must be something about paths then
and content-type of course. It'll go pick the format infered from the content-type, so if the request does not specify it, it might not work. Check that out with a debugger (edited)

Luis Ramos  5 days ago
re paths: I want it not to look for views, just use the serializer
re content-type, this is executed correctly and render json:
what should I look for in the debugger?

sauloperez  5 days ago
and does the request have a content-type: application/json header?

Luis Ramos  5 days ago
ah, I see
content type is null
and format is js
I changed it to respond_with.format.js but Im still getting the same error: https://github.com/openfoodfoundation/openfoodnetwork/pull/4073/files#diff-f1514164d0f3c3554967cc8106d13f00R12
in the example that works outside the api the content type is also null and the format is text/html, so, not a big difference

mkllnk:
I'm writing this comment while I'm trying to solve the issue. I will document my steps in case they give any insight into troubleshooting such a problem.

* I ran the controller spec and identified the first bad commit: c8165fb5de1d406fb6f5240d5b50ba2ab8ad2322 wip
* I noticed that the commit has two changes. First, the change of the render call. And second, the deletion of the rabl file.
* I brought back the rabl file `git checkout HEAD~ app/views/api/product_images/update_product_image.v1.rabl` and found that the spec is passing. So maybe the changed render command didn't do anything?
* I emptied the rabl file. Now the response code is correct but the json content is empty. This confirms that it still renders the rabl file. (I love specs.)
* Looking at the controller again. Why does render not work or not work the way we want? I noticed that it's a descendant of `Spree::Api::BaseController`.
* `bundle show spree_api` and looking at the controller. There I see a concept that I wasn't aware of but it sounds related: `self.responder = Spree::Api::Responders::AppResponder`
* I found that responder under lib and its called `RablTemplate`. https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/api/lib/spree/api/responders/rabl_template.rb
* I tried to remove the responder setting. After modifying `api/app/controllers/spree/api/base_controller.rb` and `api/lib/spree/api/controller_setup.rb` the responder was the default `ActionController::Responder`.
* I want to know where the render call is really going. After removing the rabl file, I add a rescue block to the controller method and get the following backtrace:
  ```
  actionpack-3.2.22.5/lib/action_view/path_set.rb:58:in `find'",
  actionpack-3.2.22.5/lib/action_view/lookup_context.rb:122:in `find'",
  actionpack-3.2.22.5/lib/action_view/renderer/abstract_renderer.rb:3:in `find_template'",
  actionpack-3.2.22.5/lib/action_view/renderer/template_renderer.rb:35:in `determine_template'",
  actionpack-3.2.22.5/lib/action_view/renderer/template_renderer.rb:10:in `render'",
  actionpack-3.2.22.5/lib/action_view/renderer/renderer.rb:43:in `render_template'",
  actionpack-3.2.22.5/lib/action_view/renderer/renderer.rb:24:in `render'",
  actionpack-3.2.22.5/lib/abstract_controller/rendering.rb:111:in `_render_template'",
  actionpack-3.2.22.5/lib/abstract_controller/rendering.rb:104:in `render_to_body'",
  actionpack-3.2.22.5/lib/abstract_controller/rendering.rb:89:in `render'",
  actionpack-3.2.22.5/lib/action_controller/metal/rendering.rb:16:in `render'",
  ```
  This is good, we seem to be in pure Rails land. But I don't see anything about serializers in that code. I'm wondering if the controller needs some extension for that.
* I looked at the documentation at https://github.com/rails-api/active_model_serializers/tree/0-8-stable and it doesn't mention any special addition to controllers. Just the models have to be serialisable.
* Looking at a working example `Api::OrdersController`. It inherits from `BaseController`. Changing that in the `ProductImagesController` makes the test pass again. :heavy_check_mark: 
* I change the ImageSerializer to raise an exception to get the backtrace. That should help to identify the difference to above.
  ```
  active_model_serializers-0.8.4/lib/active_model/serializer.rb:468:in `rescue in attributes'",
  active_model_serializers-0.8.4/lib/active_model/serializer.rb:456:in `attributes'",
  active_model_serializers-0.8.4/lib/active_model/serializer.rb:480:in `_serializable_hash'",
  active_model_serializers-0.8.4/lib/active_model/serializer.rb:359:in `serializable_hash'",
  active_model_serializers-0.8.4/lib/active_model/serializer.rb:347:in `as_json'",
  activesupport-3.2.22.5/lib/active_support/json/encoding.rb:47:in `block in encode'",
  activesupport-3.2.22.5/lib/active_support/json/encoding.rb:77:in `check_for_circular_references'",
  activesupport-3.2.22.5/lib/active_support/json/encoding.rb:46:in `encode'",
  activesupport-3.2.22.5/lib/active_support/json/encoding.rb:31:in `encode'",
  activesupport-3.2.22.5/lib/active_support/core_ext/object/to_json.rb:16:in `to_json'",
  active_model_serializers-0.8.4/lib/active_model/serializer.rb:331:in `to_json'",
  actionpack-3.2.22.5/lib/action_controller/metal/renderers.rb:93:in `block in <module:Renderers>'",
  active_model_serializers-0.8.4/lib/action_controller/serialization.rb:48:in `block (2 levels) in <module:Serialization>'",
  actionpack-3.2.22.5/lib/action_controller/metal/renderers.rb:35:in `block in _handle_render_options'",
  ../../set.rb:263:in `each_key'",
  ../../set.rb:263:in `each'",
  actionpack-3.2.22.5/lib/action_controller/metal/renderers.rb:32:in `_handle_render_options'",
  actionpack-3.2.22.5/lib/action_controller/metal/renderers.rb:28:in `render_to_body'",
  actionpack-3.2.22.5/lib/abstract_controller/rendering.rb:89:in `render'",
  actionpack-3.2.22.5/lib/action_controller/metal/rendering.rb:16:in `render'",
  ```
  `AbstractController` calls `render_to_body` in both examples. But in the non-working example it simply calls itself. I guess that is default Rails behaviour. In the working example, the call goes to `ActionController::Renderers`.
* Looking at the Api::BaseController, there is some code that probably makes it work and Spree code is probably missing this: https://github.com/openfoodfoundation/openfoodnetwork/blob/35841066d1c7678e20faa21a970ca83c1be7b042/app/controllers/api/base_controller.rb#L1-L11

This took me 2.5 hours. I now know a lot more about rendering. :slightly_smiling_face: We now know that Spree controllers don't support AMS and that we need to use our own base controller for that.

@luisramos0 I guess you are now unblocked by using `class ProductImagesController < BaseController`?

luisramos0:
@mkllnk ❤️ I am very grateful you took the time to help, thanks a lot!

That was an amazing investigation and I learned a new trick: checking the stack trace in the controller to see where the call is coming from!

It's a simple conclusion actually: the Spree::Api::BaseController has the Rabl responder activated because spree_api only uses rabl and the OFN Api::BaseController does have AMS activated. So, as we convert rabl to AMS, we need to switch the base_controller from Spree::Api::BaseController to Api::BaseController 💪 

Thank you!

spec/features/admin/bulk_product_update_spec.rb
```diff
@@ -780,7 +781,7 @@
       within "table#listing_products tr#p_#{product.id}" do
         # New thumbnail is shown in image column
         @new_thumb_src = page.find("a.image-modal img")['src']
-        expect(@old_thumb_src) != @new_thumb_src
+        expect(@old_thumb_src).not_to eq @new_thumb_src
```
kristinalim: Nice catch, @luisramos0!

RachL:
I've tested uploading through the `admin/products`page and through the `admin/products/XX/edit`page and both are looking fine.

image

Issue #2807, opened 2018-10-02
[Defacepocalypse] Convert heavily defaced Spree views to OFN views


luisramos0:
This EPIC represents the technical initiative to move away from deface and implement [this convention](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code,-the-way-we-do-things#re-building-spree-views).
Some of this work will be done as part of the spree upgrade, but not necessarily.

PR #4047, opened 2019-07-16
De-deface some admin configuration pages and the general admin layout and menu

luisramos0:
#### What? Why?

There's no issue for this but this is part of #2807. In this PR I am de-defacing a few things in the backoffice:
- configurations - general settings
- configurations - image settings
- configurations - shipping categories
- main admin layout and admin/shared/head
- main admin menu

No functional changes done, just bringing code from spree, converting to haml and applying the defaces.

#### What should we test?
Make sure these pages show all fields and work as before:
- Configurations - General Settings page
- Configurations - Image Settings
- Configurations - Shipping Categories

The main admin menu has a new order of entries.

The main admin layout was de-defaced, in terms of testing I think a general test of menus, submenus and general page layouts should be enough. 

#### Release notes
Changelog Category: Changed
Removed deface from a few parts of the backoffice thus making OFN less entangled with Spree code.

PR #4074, opened 2019-07-24
[Defacepocalypse] De-deface product properties index

### LINKED ISSUE #1759: Confusing UX on Delete Properties
Update by Myriam 21.05.2018 to clarify what needs to be done.

## Description
While editing the product, there is an option to view 'PRODUCT PROPERTIES' This then shows any properties that have been assigned to the product. To delete a property you MUST click on the bin icon to the right of the product. Clicking the 'x' beside the property name does not result in a permanent removal of the property.

## Expected Behavior
This page would be much more clear and consistent with the rest of OFN if we simply stop the X appearing next to the property name. (See image, there is no X)
tothis
The extra click for 'Update' is still obviously a UX sub-optimality however the user is warned before leaving the page if they haven't clicked save which is a big improvement to the current situation.

## Actual Behavior
See image, there is a confusing X
thispart

## Steps to Reproduce
1. Go to "products", select one product and click "edit" then go to tab "product properties"
2. Try deleting a property using the X and/or the bin

## Context
Was reported by uk user https://community.openfoodnetwork.org.uk/t/deleting-product-properties/194

## Severity
S4

## Your Environment
PC, Retested on French production, Chrome

## Possible Fix
Remove the "X" on property label

HugsDaniel:
#### What? Why?

There's no issue but it's part of #2807 
Closes #1759 

I imported and de-defaced views from `spree/backend/app/views/spree/admin/product_properties`. So now we have two more views : 
 - `app/views/spree/admin/product_properties/_product_property_fields.html.haml`
 - `app/views/spree/admin/product_properties/index.html.haml`

Also added a `%th.no-border` to fix the weird shift between table headers and input fields.

Appart from that, no functional changes done, just bringing code from spree, converting to haml and applying the defaces.

#### What should we test?

Make sure this page works as before : 
 - Product -> Edit -> Product Properties

#### Release notes

Changelog Category: Changed
Removed deface from a few parts of the backoffice thus making OFN less entangled with Spree code.


luisramos0:
I found this last week, it works really well :-O
http://htmltohaml.com/

app/views/spree/admin/product_properties/index.html.haml
```diff
@@ -0,0 +1,72 @@
...
+    = hidden_field_tag 'clear_product_properties', 'true'
+
+:coffee
+  $(document).ready ->
+    $("#inherited_properties").toggle $("input#product_inherits_properties").is(':checked')
+  $("input#product_inherits_properties").change ->
+    $("#inherited_properties").toggle $(this).is(':checked')
+
+= javascript_tag do
+  var properties = #{raw(@properties.to_json)};
```
luisramos0: there's an indentation error here.
I found these js convertions from erb to haml to break easily. I dont think this raw and to_json will work like this...
Have you tested this? Give it a try. If it's broken and it's hard to fix, maybe it's acceptable to delete this autocomplete function?

HugsDaniel: I had errors when trying to use correct indentation with `= javascript_tag`, so I changed it to `:javascript`. No errors anymore, and the autocomplete works fine !

HugsDaniel:
> It would be good to also bring from spree and convert to haml the missing partials:
> 
> * 'spree/admin/shared/edit_resource_links'
> * 'spree/admin/shared/product_sub_menu'
> * 'spree/shared/error_messages'

Done!

app/views/spree/admin/product_properties/index.html.haml
```diff
@@ -60,13 +60,13 @@
   $("input#product_inherits_properties").change ->
     $("#inherited_properties").toggle $(this).is(':checked')
 
-= javascript_tag do
+:javascript
   var properties = #{raw(@properties.to_json)};
-
+  console.log(properties)
```
luisramos0: this needs to be removed

RachL:
@hugs I can add and edit properties. However, I cannot add them to products afterwards. I don't have a dropdown in `/admin/products/test/product_properties`
Here is what the dropdown currently looks like in french staging (without your PR):
image


And in staging ES with this PR I had this:

image

HugsDaniel:
Hey @RachL , how do you add properties to a product ?

About the dropdown, I have this in master currently : 

Capture d’écran de 2019-07-31 17-05-22

No dropdown, just a regular autocomplete. Can you check master ?

RachL:
😭 I hoped French staging had a recent version of master.  So this means the bug was introduce somewhere else...

In production the dropdown is still here. So it is really a bug that we will introduce with the new release.
Any idea which one of the latest merged PR could have caused this? ping @luisramos0 @sauloperez @mkllnk 

https://github.com/openfoodfoundation/openfoodnetwork/pulls?q=is%3Apr+merged%3A%3E2019-07-19T21%3A37%3A00%2B01%3A00+sort%3Aupdated-desc+base%3Amaster

About the properties, you can add them in `/admin/properties` @HugsDaniel

luisramos0:
:-(
I am sorry, I am doing some pretty risky changes lately... let me say it's totally worth it, such a good clean up!
I'd bet on #4047 again...

HugsDaniel:
@RachL there's no dropdown but there is still autocomplete on the property's name. I'm not sure it's blocked however, since the bug has nothing to do with de-defacing. The HTML is pretty much the same as master, there's just a problem elsewhere with select2 dropdowns.

RachL:
@hugs I've put the blocked label because once the dropdown is fixed, I don't know if the fix will have another impact on this PR. I just feel I cannot complete the test.

I will see if we squease a fix for the bug or not soon so I can re-test both.

RachL:
@HugsDaniel I've just staged master as part of the release test and I do have the dropdown... Can you check master again on your side?
I noticed also that at least with the version we have here the column titles are better align. So actually if we just add a placeholder to explain the autocompletion, this is good to go?

luisramos0:
I found #1759 that is related to this PR. Maybe something to look at as part of this PR.

HugsDaniel:
@RachL stupid question but the dropdown appears only for non admin users, I guess you tested it with both types ?

HugsDaniel:
I have this in this branch :
Capture d’écran de 2019-08-22 10-19-07

HugsDaniel:
@luisramos0 I tackled #1759 at the same time. @RachL can you test if the dropdown has two arrows on the right, like they have in my screenshot above ?

RachL:
@HugsDaniel yes I do have the dropdown, but only one arrow:

image


So I guess you only need to tackle the unsucessful checks?

sigmundpetersen:
Rebuild passed on Semaphore, so those build errors are random. So I guess good to go then?

PR #4075, opened 2019-07-24
[Defacepocalypse] De-deface payment methods index

HugsDaniel:
#### What? Why?

There's no issue but it's part of #2807

I imported and de-defaced views from `spree/backend/app/views/spree/admin/payment_methods`. So now we have more view :

`app/views/spree/admin/payment_methods/index.html.haml`

No functional changes done, just bringing code from spree, converting to haml and applying the defaces.

#### What should we test?

Make sure this view works as before : 
 - Configuration -> Payment Methods

#### Release notes

Changelog Category: Changed
Removed deface from a few parts of the backoffice thus making OFN less entangled with Spree code.


HugsDaniel:
@luisramos0 do we still remove the data-hooks ? I forgot that part.

luisramos0:
yes, you can remove then. I think we can also have a massive "delete data-hooks" PR in the end.
the challenge is that I think there's some spree JS code depending on data-hooks...

HugsDaniel:
Yeah sometimes there is, I try to test as much as I can each time I delete a data-hook.

RachL:
So as super admin I went to "configurations". Added Stripe and Paypal for Mary and Paypal for Maryse. Edited both and also checkout through the shopfront with these two payments methods.

All clear. Ready to go!

PR #4081, opened 2019-07-25
Bulk Edit Products pagination

### LINKED ISSUE #4076: Do not load all pages in Bulk Edit Products

Currently, the Bulk Edit Products page loads all pages regardless of how much you scroll. You might want to see only the first one. This adds a totally unnecessary load on the server.

[services/bulk_products.js.coffee](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/admin/services/bulk_products.js.coffee) uses [PagedFetcher](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee) to load them async.

### Context

This stems from https://github.com/openfoodfoundation/ofn-install/issues/477#issuecomment-514677459

### Impact and timeline

This should cut the negative impact on performance this page has which seems to be causing serious downtime.

Matt-Yorkley:
#### What? Why?

Closes #4076 

Pagination for Bulk Edit Products. Should resolve a number of BEP-related issues (there's a list somewhere).

#### What should we test?

Admin Products page functionality is all working as before, but with paginated results.

#### Release notes

Added pagination to Admin Products page

Changelog Category: Changed

### Screenshots

I've reused the elements from the orders page to try to get a bit of consistency, but it really feels like we need a UI person to design us a nice admin area...

`TOP:`
BEP1
`BOTTOM:`
BEP2

### DIFF
--- a/spec/features/admin/product_import_spec.rb
+++ b/spec/features/admin/product_import_spec.rb
@@ -170,6 +170,7 @@ feature "Product Import", js: true do
       expect(page).to have_selector 'div#s2id_import_date_filter'
       import_time = carrots.import_date.to_date.to_formatted_s(:long)
       select2_select import_time, from: "import_date_filter"
+      page.find('.button.icon-search').click
 
       expect(page).to have_field "product_name", with: carrots.name
       expect(page).to have_field "product_name", with: potatoes.name
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,10 @@ gem 'delayed_job_web'
 # When merged, revert to upstream gem
 gem 'simple_form', github: 'RohanM/simple_form'
 
+# Spree's default pagination gem (locked to the current version used by Spree)
+# We use it's methods in OFN code as well, so this is a direct dependency
+gem 'kaminari', '~> 0.14.1'
+
 gem 'andand'
 gem 'angularjs-rails', '1.5.5'
 gem 'aws-sdk'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -808,6 +808,7 @@ DEPENDENCIES
   jquery-rails (= 3.0.4)
   json_spec (~> 1.1.4)
   jwt (~> 2.2)
+  kaminari (~> 0.14.1)
   knapsack
   letter_opener (>= 1.4.1)
   listen (= 3.0.8)
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -54,6 +54,7 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     $scope.fetchProducts()
 
   $scope.fetchProducts = ->
+    removeClearedValues()
     params = {
       'q[name_cont]': $scope.query,
       'q[supplier_id_eq]': $scope.producerFilter,
@@ -64,6 +65,11 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     }
     RequestMonitor.load(BulkProducts.fetch(params).$promise).then ->
       $scope.resetProducts()
+
+  removeClearedValues = ->
+    delete $scope.producerFilter if $scope.producerFilter == "0"
+    delete $scope.categoryFilter if $scope.categoryFilter == "0"
+    delete $scope.importDateFilter if $scope.importDateFilter == "0"
 
   $timeout ->
     if $scope.showLatestImport
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -1,267 +1,267 @@
 angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout, $http, $window, BulkProducts, DisplayProperties, dataFetcher, DirtyProducts, VariantUnitManager, StatusMessage, producers, Taxons, SpreeApiAuth, Columns, tax_categories) ->
+  $scope.loading = true
+  $scope.loadingAllPages = true
+
+  $scope.StatusMessage = StatusMessage
+
+  $scope.columns = Columns.columns
+
+  $scope.variant_unit_options = VariantUnitManager.variantUnitOptions()
+
+  $scope.filterableColumns = [
+    { name: t("label_producers"),       db_column: "producer_name" },
+    { name: t("name"),           db_column: "name" }
+  ]
+
+  $scope.filterTypes = [
+    { name: t("equals"),         predicate: "eq" },
+    { name: t("contains"),       predicate: "cont" }
+  ]
+
+  $scope.optionTabs =
+    filters:        { title: t("filter_products"),   visible: false }
+
+
+  $scope.producers = producers
+  $scope.taxons = Taxons.all
+  $scope.tax_categories = tax_categories
+  $scope.filterProducers = [{id: "0", name: ""}].concat $scope.producers
+  $scope.filterTaxons = [{id: "0", name: ""}].concat $scope.taxons
+  $scope.producerFilter = "0"
+  $scope.categoryFilter = "0"
+  $scope.importDateFilter = "0"
+  $scope.products = BulkProducts.products
+  $scope.filteredProducts = []
+  $scope.currentFilters = []
+  $scope.limit = 15
+  $scope.query = ""
+  $scope.DisplayProperties = DisplayProperties
+
+  $scope.initialise = ->
+    SpreeApiAuth.authorise()
+    .then ->
+      $scope.spree_api_key_ok = true
+      $scope.fetchProducts()
+    .catch (message) ->
+      $scope.api_error_msg = message
+
+  $scope.$watchCollection '[query, producerFilter, categoryFilter, importDateFilter]', ->
+    $scope.limit = 15 # Reset limit whenever searching
+
+  $scope.fetchProducts = ->
     $scope.loading = true
     $scope.loadingAllPages = true
+    BulkProducts.fetch($scope.currentFilters, ->
+      $scope.loadingAllPages = false
+    ).then ->
+      $scope.resetProducts()
+      $scope.loading = false
 
-    $scope.StatusMessage = StatusMessage
+  $timeout ->
+    if $scope.showLatestImport
+      $scope.importDateFilter = $scope.importDates[1].id
 
-    $scope.columns = Columns.columns
+  $scope.resetProducts = ->
+    DirtyProducts.clear()
+    StatusMessage.clear()
 
-    $scope.variant_unit_options = VariantUnitManager.variantUnitOptions()
+  $scope.updateOnHand = (product) ->
+    on_demand_variants = []
+    if product.variants
+      on_demand_variants = (variant for id, variant of product.variants when variant.on_demand)
 
-    $scope.filterableColumns = [
-      { name: t("label_producers"),       db_column: "producer_name" },
-      { name: t("name"),           db_column: "name" }
-    ]
-
-    $scope.filterTypes = [
-      { name: t("equals"),         predicate: "eq" },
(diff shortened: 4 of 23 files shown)


Matt-Yorkley:
I had a bit of an issue with the "clear filters" button, which I resolved in https://github.com/openfoodfoundation/openfoodnetwork/pull/4081/commits/d65ba30082b5990588dff859c9a6d0d7effeeaaa

I wasn't satisfied with it at all, but I couldn't fix it any other way, and spent more time on it than it was worth.

sauloperez:
Could you share a screenshot or animated GIFs since this is changing the UI?

spec/features/admin/product_import_spec.rb
```diff
@@ -170,6 +170,9 @@
       expect(page).to have_selector 'div#s2id_import_date_filter'
       import_time = carrots.import_date.to_date.to_formatted_s(:long)
       select2_select import_time, from: "import_date_filter"
+      page.find('.button.icon-search').click
+
+      save_and_open_screenshot
```
sauloperez: :warning: :fire:

app/controllers/spree/api/products_controller_decorator.rb
```diff
@@ -66,6 +67,11 @@ def product_scope
     scope.includes(:master)
   end
 
+  def import_date_scope
```
sauloperez: I would at least test this in a controller spec.

Matt-Yorkley:
These tests are passing locally, and working with manual testing... :thinking:

app/controllers/spree/api/products_controller_decorator.rb
```diff
@@ -20,7 +20,7 @@ def bulk_products
 
     @products = product_query.order('created_at DESC').
       ransack(params[:q]).result.
-      page(params[:page]).per(params[:per_page])
+      page(params[:page] || 1).per(params[:per_page] || 50)
```
luisramos0: magic numbers should be a constants at the top:
DEFAULT_PAGE = 1
DEFAULT_PER_PAGE = 50

spec/controllers/spree/api/products_controller_spec.rb
```diff
@@ -176,5 +168,66 @@ module Spree
...
+          spree_get :bulk_products, { page: 1, per_page: 15, q: {supplier_id_eq: supplier.id} }, format: :json
+
+          expect(returned_product_ids).to eq [product2.id, product1.id]
+        end
+
+        it "filters results by product category" do
+          spree_get :bulk_products, { page: 1, per_page: 15, q: {primary_taxon_id_eq: taxon.id} }, format: :json
+
+          expect(returned_product_ids).to eq [product3.id, product2.id]
+        end
```
luisramos0: we should test pagination a bit better I think. maybe a test where page is not 1. and then test some limits like request page over the limit (page 2 of 50 records where per_page is 50).

app/controllers/spree/api/products_controller_decorator.rb
```diff
@@ -89,12 +89,16 @@ def render_paged_products(products)
 
     render text: {
       products: serializer,
```
luisramos0: can we be json-api inspired and call this data instead of products?

app/controllers/spree/api/products_controller_decorator.rb
```diff
@@ -89,12 +89,16 @@ def render_paged_products(products)
...
     }.to_json
   end
+
+  def pagination_data(results)
+    {
+      results: results.total_count,
+      pages: results.num_pages,
+      page: params[:page].to_i,
+      per_page: params[:per_page].to_i
+    }
```
luisramos0: I'd refer to see json-api structure here so we get our API pagination exemplar done:
https://jsonapi.org/examples/#pagination


```ruby
  meta: {
    totalPages: results.total_count
  },
  data: serializer,
  links: {
    self: "/api/products?page[number]=#{params[:page].to_i}&page[size]=1",
    first: "http://example.com/articles?page[number]=1&page[size]=1",
     prev: "http://example.com/articles?page[number]=2&page[size]=1",
     next: "http://example.com/articles?page[number]=4&page[size]=1",
     last: "http://example.com/articles?page[number]=13&page[size]=1"
  }
```

app/controllers/spree/api/products_controller_decorator.rb
```diff
@@ -66,6 +71,11 @@ def product_scope
     scope.includes(:master)
   end
 
+  def import_date_scope
+    import_date = params[:import_date].to_datetime
+    Spree::Variant.where(import_date: import_date..import_date + 24.hours)
+  end
```
sauloperez: I don't like making the models big but I think this would be better in the variant model as a AR scope. Definitely, nothing to couple the controller with.

app/controllers/spree/api/products_controller_decorator.rb
sauloperez: If we go with JSON API we'll have to take a look at http://jsonapi-rb.org/. What if we get this PR out with the current pagination data and discuss it later on so we don't delay this fix?

spec/controllers/spree/api/products_controller_spec.rb
Matt-Yorkley: I've used a smaller number of objects (2 per page), but added page tests. :+1:

app/controllers/spree/api/products_controller_decorator.rb
luisramos0: ok :+1: we can do that.
that library doesnt help much with pagination though, we will have to do that part ourselves.

app/controllers/spree/api/products_controller_decorator.rb
Matt-Yorkley: Moved to a new scope and tests added :+1:

app/controllers/spree/api/products_controller_decorator.rb
```diff
@@ -11,11 +14,16 @@ def managed
...
+      editable_products.merge(product_scope)
+
+    if params[:import_date].present?
+      product_query = product_query.imported_on(params[:import_date]).group_by_products_id
+    end
+
+    @products = product_query.order('created_at DESC').
       ransack(params[:q]).result.
-      page(params[:page]).per(params[:per_page])
+      page(params[:page] || DEFAULT_PAGE).per(params[:per_page] || DEFAULT_PER_PAGE)
```
sauloperez: I wanted to ask you to investigate the gem's config options so that we ensure `params[:per_page]` it's within a defined upper limit so no malicious users can blow up the server but... I don't even see which gem we're using so can you please specify it in our Gemfile? this is implicit dependency on Spree is particularly scary because I assume we're using whatever they might have defined.

Matt-Yorkley: Looks like it's `Kaminari`, and yes it's included by Spree.

Matt-Yorkley: Ok, I've added an entry in the Gemfile (with a note), and added an initializer with a max_per_page value for Kaminari. :+1:

Matt-Yorkley:
I'm a bit stuck with this. I've got 3 failing tests (one model, one api, one feature) that are all failing specifically because filtering on `import_date` fails to return any results.

- spec/models/spree/product_spec.rb:399
- spec/features/admin/product_import_spec.rb:135
- spec/controllers/spree/api/products_controller_spec.rb:215

All three tests pass every time when I run them locally. All three tests fail every time in CI. It seems to work fine in the browser, locally and on staging.

Matt-Yorkley:
Ok, the build is green now after another rebase...

daniellemoorhead:
Your testing notes aren't super comprehensive @Matt-Yorkley (speaking as a noob tester).... #perhapsharshbuttrue #goeasyonus #testingisnotfun

I _think_ I have identified all the pagination elements you added:

- Pagination elements are always displayed, regardless of whether there are multiple pages or not.

- Default maximum number of products on the page is 15.

- The pagination elements on the page include:
    - copy about number of products located at the top of the list (_xx_ Resultados encontrados. Viendo de _yy_ a _zz_)

   - is a drop down option to change the maximum number displayed to 50 or 100. This includes a search option attached to the dropdown that allows the user to search for one of (only) 3 results display numbers available.

    - controls at bottom of list: start, previous, page number, next, end. These are inactive if maximum products on page has not been reached.

I've been testing these elements and there are bits that don't work. I'm going to write up test notes, record what I've done, and then I'll put this back into Dev Ready to be picked up by you or the next available person.

daniellemoorhead:
Full testing notes here @Matt-Yorkley: https://docs.google.com/document/d/1YcI8st8cmYg-lv4S2RKvKrn15LgdFAehVotz7HGlPPk/edit?usp=sharing

Needs a fair bit of fixing. Sad face.

Matt-Yorkley:
Ah, you have to hit the "filer results" button after changing selections. This is consistent with the orders page UX, but we should review both at some point.

Matt-Yorkley:
Retesting notes:

I've removed the search bar from the dropdown:

no-search

The pagination elements are all working (as per my comment above):

> you have to hit the "filter results" button after changing selections. This is consistent with the orders page UX, but we could review both at some point.

I'm not sure about updating the pagination buttons at the bottom after deleting 15+ products. The buttons are currently static and based on the initial response from the server, and are a shared component that's used elsewhere. Shall we move that to another issue?

lin-d-hop:
@matt-yorkley can you fix the conflicts please?

RachL:
@Matt-Yorkley on the order page, the result are loading as soon as you select "50 per page" e.g. 
is this behavior possible on the product page as well for consistency?
Otherwise I would suggest to move the "filter results" below the dropdown. Because the order page is not working like that, it might be difficult for a manager to understand he/she has a click on a button _*above*_ the action button to see a result.

Other than that I agree that the behavior when we duplicate or delete can be done in a second iteration. This clearly shows we cannot add pagination without a bit of inception / mockups. I agree with @daniellemoorhead test criteria but at the same time I agree that those criteria where not in the original issue and that this PR is solving a huge problem of performance. Thus this PR should be able to move forward without those criteria fulfilled.

So I think we can move it to ready to go as soon as we have your feedback on the two different behavior on order page pagination vs product page pagination. What do you think?

Matt-Yorkley:
The results now update when the dropdown is changed. :+1:

RachL:
@Matt-Yorkley the build is failing, can I re-stage this safely though?

Matt-Yorkley:
These 3 specs are suddenly flaky again, and I can replicate the issue: if the timezone in the test environment is set to London, they pass, and if it's set to Melbourne, they fail...

- spec/models/spree/product_spec.rb:399
- spec/features/admin/product_import_spec.rb:135
- spec/controllers/spree/api/products_controller_spec.rb:215

RachL:
Great! Thanks @Matt-Yorkley we have a first version that is ready to go :tada:

@daniellemoorhead can you open a new issue with the second UX iteration?

luisramos0:
@Matt-Yorkley you have added a commit to product_decorator after PR is ready to go.
Merge or re-test? 🙈

Matt-Yorkley:
I was fixing a flaky spec affecting the CI build whilst it was being tested. @RachL do you want to do a quick re-check?

RachL:
I did a very quick check and it is looking good. Let's merge this!!

sauloperez:
:hankey: I was about to merge this but it needs a :recycle: (aka, rebase) :warning:

sauloperez:
I just got a Datadog monitoring alert because someone got a request timed out because of this.

screencapture-mail-google-mail-u-0-2019-09-04-19_12_47

To actually validate it, we should see no more alarms after deploying v2.4.0 https://github.com/openfoodfoundation/openfoodnetwork/issues/4214

sauloperez:
Good news! according to Datadog

> The monitor was last triggered at Thu Sep 05 2019 13:18:20 UTC

the exact same day v2.4.0 was deployed. So this clearly had a positive metricable impact.

PR #4089, opened 2019-07-29
Add doctype declaration in layouts/admin view

### LINKED ISSUE #4088: Product Bulk Edit page layout is a little broken
## Description

Screenshot 2019-07-29 at 12 24 20

The font size is bigger in the header (not necessarily a bad thing :-)) and the action icons are not centered, which is definetely a problem we need to solve now.

## Expected Behavior
Icons should be centered.

## Actual Behaviour
Icons are not centered.

## Steps to Reproduce
Open Bulk product edit page in a server with current master.

## Context
Styling problem introduced in #4047

## Severity
I am not sure, maybe bug-s2, it has not been released, it's only in master.

## Possible Fix
Some CSS rules priority has gone wrong and the browser is using default browser styles instead of OFN css.

luisramos0:
#### What? Why?

Closes #4088 
4808 was caused by the [missing doctype declaration](https://github.com/openfoodfoundation/openfoodnetwork/pull/4089/commits/9bc81f236c26f6025a04e8a80a45f8392d7bc302), this declaration tells the browser to read OFN pages as html5. Problem solved.

I used this PR to make a few more improvements:
- fix indentation problem and keep existing html structure - I dont think this has a functional impact
- remove dead code related to expand_variants, related to product auto complete which is the ancestor of variant_autocomplete - old code, no need to test manually
- move head script from admin layout to head partial - fonts

#### What should we test?
Verify 4088 is fixed and that the fonts are rendering as before.

#### Release notes
This is fixing a problem introduced in another PR of this release, no release notes.

### DIFF
--- a/app/views/spree/admin/shared/_head.html.haml
+++ b/app/views/spree/admin/shared/_head.html.haml
@@ -21,4 +21,7 @@
   = "jQuery.alerts.dialogClass = 'spree';"
   = raw "var AUTH_TOKEN = \"#{form_authenticity_token}\";"
 
+= render "layouts/bugherd_script"
+%link{'data-require' => "font-awesome@*", 'data-semver'=>"4.2.0", 'rel' => "stylesheet", 'href' => "//cdnjs.cloudflare.com/ajax/libs/font-awesome/4.2.0/css/font-awesome.css"}
+
 = yield :head
--- a/app/views/spree/layouts/admin.html.haml
+++ b/app/views/spree/layouts/admin.html.haml
@@ -3,9 +3,6 @@
   %head{"data-hook" => "admin_inside_head"}
     = render :partial => 'spree/admin/shared/head'
 
-    = render "layouts/bugherd_script"
-    %link{'data-require' => "font-awesome@*", 'data-semver'=>"4.2.0", 'rel' => "stylesheet", 'href' => "//cdnjs.cloudflare.com/ajax/libs/font-awesome/4.2.0/css/font-awesome.css"}
-
   %body.admin
     = admin_inject_currency_config
     = render "layouts/i18n_script"
--- a/app/views/spree/layouts/admin.html.haml
+++ b/app/views/spree/layouts/admin.html.haml
@@ -1,3 +1,4 @@
+!!!
 %html{:lang => "en"}
   %head{"data-hook" => "admin_inside_head"}
     = render :partial => 'spree/admin/shared/head'
--- a/app/views/spree/layouts/admin.html.haml
+++ b/app/views/spree/layouts/admin.html.haml
@@ -17,69 +17,69 @@
       - if flash[:success]
         .flash.success= flash[:success]
 
-    #progress
-      .wrapper
-        #spinner
-        .progress-message
-          = Spree.t(:loading)
-          \...
+      #progress
+        .wrapper
+          #spinner
+          .progress-message
+            = Spree.t(:loading)
+            \...
 
-    = render :partial => 'spree/admin/shared/alert', :collection => session[:alerts]
+      = render :partial => 'spree/admin/shared/alert', :collection => session[:alerts]
 
-    %header#header{"data-hook" => ""}
-      .container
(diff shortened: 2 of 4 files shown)


app/views/spree/admin/shared/_head.html.haml
```diff
@@ -21,4 +21,7 @@
   = "jQuery.alerts.dialogClass = 'spree';"
   = raw "var AUTH_TOKEN = \"#{form_authenticity_token}\";"
 
+= render "layouts/bugherd_script"
```
sauloperez: are we really using it? cc @RachL @kirstenalarsen

app/views/spree/layouts/admin.html.haml
```diff
@@ -1,85 +1,84 @@
+!!!
```
sauloperez: what does this do?

app/views/spree/admin/shared/_head.html.haml
luisramos0: yeah, same question for goggle analytics...

app/views/spree/layouts/admin.html.haml
luisramos0: it adds the doctype tag in the very beginning of each page.
which basically means "this doc is html5"
Screenshot 2019-07-29 at 17 00 58

luisramos0: otherwise it defaults to html4/legacy

app/views/spree/admin/shared/_head.html.haml
RachL: Google Analytics is used by UK and Australia is considering switching to Matomo: https://openfoodnetwork.slack.com/archives/CG7NJ966B/p1562290349053500

However I'm sorry but I'm not sure what you are referring to @sauloperez

luisramos0:
Second time (I also tried in #4088) I have to remove the de-deface of the app_wrapper code that adds the angular ng-app tag to every page. I cant make this work, it's a yield command in haml that produces html attributes...

Matt-Yorkley:
> Second time (I also tried in #4047) I have to remove the de-deface of the app_wrapper code that adds the angular ng-app tag to every page.

I think this line is wrong: `%div= yield(:app_wrapper_attrs).strip.html_safe`.

If you look at the original `erb`, it's this: `<div <%= yield(:app_wrapper_attrs).strip.html_safe %>>`. Here the attributes are being rendered _inside_ the opening div tag (before it closes).

In this line, the attributes are rendered inside the div itself, after the opening tag has closed: `%div= yield(:app_wrapper_attrs).strip.html_safe`.

app/views/spree/admin/shared/_head.html.haml
sauloperez: that Bugherd script :point_right: http://www.bugherd.com

luisramos0:
ah, yes @Matt-Yorkley thanks.
that was my first try, than I tried different things and none worked, for example:
%div{= yield(:app_wrapper_attrs).strip.html_safe}
or
%div{yield(:app_wrapper_attrs).strip.html_safe}

luisramos0:
@Matt-Yorkley anyway we dont need that for this PR, it's out of scope.

what we really need is the doctype magic !!! ;-)

app/views/spree/admin/shared/_head.html.haml
mkllnk: We used Bugherd in the past. I will check with the team here.

luisramos0: hey guys, this PR is just keeping Bugherd as it was, I'd say this is out of scope for this PR.
Because this PR is blocking the release, can we move it forward and handle Bugherd in a separate PR?

RachL:
So I've tested in:

- IE 18
- Firefox 68
- Chrome 75

I've tested product page, OC, reports and customers. All tables are now back to their previous font sizes and icons are centered :+1:

luisramos0:
great, thanks @RachL

Issue #4077, opened 2019-07-24
[Spike] Analyze impact of /api/products/bulk_products on DB


sauloperez:
### What we should change and why (this is tech debt)

We need to know what are the queries this request generates and see what impact they have on DB in terms of response time. Then, we can check how these can be optimized.

### Context

This stems from https://github.com/openfoodfoundation/ofn-install/issues/477.

### Impact and timeline

This should cut the negative impact on performance this page has which seems to be causing serious downtime.

luisramos0:
I am not sure we need to do this work. When a super admin loads the products update page, there are dozens?hundreds? of requests to the server for each page... I think we just need to do #4076

sauloperez:
Copying a relevant request log from https://github.com/openfoodfoundation/ofn-install/issues/477#issuecomment-514121571

```
Started GET "/api/products/bulk_products?page=29;per_page=20;" for 82.132.212.49 at 2019-07-22 15:56:04 +0000
(...)
Completed 200 OK in 64479.5ms (Views: 34634.5ms | ActiveRecord: 29843.7ms)
```
Notice the **ActiveRecord: 29843.7ms** we can hardly handle something like this unless we at least take a look. That is why I added spike to this issue.

Matt-Yorkley:
I'm doing the pagination now (it's nearly ready for review). I think the queries are still pretty bad even when they're smaller, so I think it's worth seeing if there's any quick wins (like simple N+1s we can strip out) to improve it.

luisramos0:
ok, agreed!

Matt-Yorkley:
```ruby
def spike
   if quick_wins_available?
      fix_it_now!
   else
      redirect_to_dev_ready_column
   end
end
```
:smile:

sauloperez:
I created https://github.com/openfoodfoundation/openfoodnetwork/pull/4092 while investigating this.

sauloperez:
Here's the outcome of the investigation. Basically, wherever I look I see N+1 which makes it harder to prioritize... but these are what look to the relevant things:

### All Api::Admin::ProductSerializer's attributes trigger a query

None of the many attributes and associations of the `Api::Admin::ProductSerializer`, including variants and the atttributes and associations of their `Api::Admin::VariantSerializer`, are eager loaded. Nested N+1? M*(N+1) :boom:?

### Mem-expensive enterprises query in /admin/products

Copied straight from the log

```
[...]

PR #4092, opened 2019-07-29
Remove N+1 while fetching distributors of OCs

sauloperez:
#### What? Why?

Closely related to  #4077. `/api/products/bulk_products` and `/admin/products` get always executed together. 

What used to be done as

```sql
SELECT "order_cycles".* FROM "order_cycles"
WHERE (order_cycles.orders_open_at <= '2019-07-29 17:45:20.137294'
  AND order_cycles.orders_close_at >= '2019-07-29 17:45:20.137333')

SELECT DISTINCT "enterprises".* FROM "enterprises"
INNER JOIN "exchanges" ON "enterprises"."id" = "exchanges"."receiver_id"
WHERE "exchanges"."order_cycle_id" = 1
  AND "exchanges"."incoming" = 'f'
(...)
SELECT DISTINCT "enterprises".* FROM "enterprises"
INNER JOIN "exchanges" ON "enterprises"."id" = "exchanges"."receiver_id"
WHERE "exchanges"."order_cycle_id" = 4
  AND "exchanges"."incoming" = 'f'
```
(notice we have N queries which just change the order_cycle_id)

It's now

```sql
SELECT "order_cycles".* FROM "order_cycles"
WHERE (order_cycles.orders_open_at <= '2019-07-29 17:45:20.137294'
  AND order_cycles.orders_close_at >= '2019-07-29 17:45:20.137333')

SELECT "exchanges".* FROM "exchanges"
WHERE "exchanges"."incoming" = 'f'
  AND "exchanges"."order_cycle_id" IN (1, 2, 3, 4)

SELECT "enterprises".* FROM "enterprises"
WHERE "enterprises"."id" IN (3, 4, 5, 6)
```

This has noticeable impact since the changed method belongs to the
`BaseController` seems to be executed in all HTML requests as it gets
called by

```ruby
before_filter :warn_invalid_order_cycles, if: :html_request?
```

#### What should we test?

The Bulk Edit Products (/admin/products) page should render exactly the same.

#### Release notes

The Bulk Edit Products page no longer loads distributors with N+1 queries.

Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/base_controller_decorator.rb
+++ b/app/controllers/spree/admin/base_controller_decorator.rb
@@ -61,7 +61,7 @@ Spree::Admin::BaseController.class_eval do
 
   def active_distributors_not_ready_for_checkout
     ocs = OrderCycle.managed_by(spree_current_user).active
-    distributors = ocs.map(&:distributors).flatten.uniq
+    distributors = ocs.includes(:distributors).map(&:distributors).flatten.uniq
     Enterprise.where('enterprises.id IN (?)', distributors).not_ready_for_checkout
   end
 


app/controllers/spree/admin/base_controller_decorator.rb
```diff
@@ -61,7 +61,7 @@ def model_class
 
   def active_distributors_not_ready_for_checkout
     ocs = OrderCycle.managed_by(spree_current_user).active
-    distributors = ocs.map(&:distributors).flatten.uniq
+    distributors = ocs.includes(:distributors).map(&:distributors).flatten.uniq
     Enterprise.where('enterprises.id IN (?)', distributors).not_ready_for_checkout
   end
```
sauloperez: A better improvement would be to totally refactor `active_distributors_not_ready_for_checkout` so that we get the enterprise ids with a complex but single query that avoids round trips to the DB but it is not as low-hanging-fruit as this first solution.

mkllnk: Another quick win for Ruby performance and Rails 4 compatibility:

```rb
    distributor_ids = ocs.includes(:distributors).map(&:distributor_ids).flatten.uniq
    Enterprise.where('enterprises.id IN (?)', distributor_ids).not_ready_for_checkout
```

mkllnk:
> Release notes

> The Bulk Edit Products page no longer loads distributors with N+1 queries.

Do you think that "N+1 queries" is a bit too technical for release notes?

sauloperez:
I'm not even sure about release notes as they are now given https://community.openfoodnetwork.org/t/release-process-and-communication-about-software-changes/1691 :trollface: . I'd keep the "N+1" for now. If we happen to explain the changes to our users, I will definitely do it differently.

Matt-Yorkley:
I this `no-test`? Shall we move it to `ready to go`?

app/controllers/spree/admin/base_controller_decorator.rb
sauloperez: I prefer to separate these two things because:

* This is already an improvement so I don't want to delay it to get both #iterative

sauloperez:
Yes, good point @Matt-Yorkley no need to bother Rachel here probably.

PR #4101, opened 2019-07-30
Convert spree/api/products and spree/api/variants views from rabl to AMS

luisramos0:
#### What? Why?

Part of #4060 

Here we convert spree/api/variants and spree/api/products from rabl to AMS and move routes and controllers out of the spree namespace into /api 🎉 

#### What should we test?

* Check that /admin/products is displaying variants with all attributes correctly.
* Check that a shop front is displaying variants correctly.
* Check that you can add an item to an order from the admin interface.
* Check that you can create a variant.
* Check that you can delete a variant.

#### Release notes
Changelog Category: Changed
Converted variants and products api endpoints from rabl to AMS.

#### Dependency
This PR depends on #4055 (the target branch of this PR is remove_spree_api, it's the branch of #4055).


luisramos0:
ah @mkllnk thanks so much for adding the testing part!

lin-d-hop:
These 'Blocked' issues in 'Test Ready'....
I'm not sure what to do with them.
@luisramos0... to test or not to test?

luisramos0:
They are blocked until the blockin issue is merged. The blocking issue
needs a code review...

luisramos0:
this is now ready for testing.

RachL:
@luisramos0 everything seems to work great as manager. As super admin however, creating and deleting variant gave me error 502 and 504 (time out).
Is it possible that the changes introduced here are making the performance worst on /products?

https://docs.google.com/document/d/1DKgv4F4TmLGYFYXrNPoM11W0cO6UfkMKiTr_h8kcJ6c/edit#

luisramos0:
yeah, I dont think the timeouts are specific to the actions, there are timeouts loading all the pages, this page loads all products in the database. Timeouts can make actions fail in some cases...

I deployed master to FR staging and I also see the timeouts exactly in the same page 45. So, I dont think this PR is making the problem worse.
Well, this PR could affect the performance of the page yes, could be slower or faster. I didnt compare the performance. I dont think the difference will be significant.

Anyway, this is fixed in #4081 where admin/products will only load one page of products :tada:

RachL:
Ok let's move it to ready to go then!

Issue #3853, opened 2019-05-16
/shop performance: product query improvements


Matt-Yorkley:
## Description

Loading enterprise/shop invokes n+1 queries. This results in a huge server load and impacts on performance.

This issue is capturing work to be done within the enterprise/shop performance feature.

## Expected Behavior

Shopfronts always load successfully.

## Actual Behaviour

Occasionally shopfronts fail to load and timeout

## Steps to Reproduce

1. Load enterprise/shop
2. Take a look on Datadog for number of queries run
3. See that we do around 11 queries for every product in the OC, so with 500 products, we make ~5500 queries. 

![Screenshot from 2019-08-02 15-57-40](https://user-images.githubusercontent.com/762088/62375317-5038a480-b53e-11e9-8227-b2654bc7d8f4.png)

## Context

This issue gives an overview of inflated querying and places to address them within enterprise/shop epic #3860 

Each of the possible fixes below should be addressed in a separate PR as much as possible such that they can be tested independently.

## Possible Fix

The performance of the products query for a shop is really bad. It starts here in the shop controller, but touches many different files:

https://github.com/openfoodfoundation/openfoodnetwork/blob/59d4e3d9e2192524d807b8dbc4271b427adea1ac/app/controllers/shop_controller.rb#L12-L24

https://github.com/openfoodfoundation/openfoodnetwork/blob/59d4e3d9e2192524d807b8dbc4271b427adea1ac/lib/open_food_network/cached_products_renderer.rb#L29-L39

https://github.com/openfoodfoundation/openfoodnetwork/blob/59d4e3d9e2192524d807b8dbc4271b427adea1ac/lib/open_food_network/products_renderer.rb#L34-L45

https://github.com/openfoodfoundation/openfoodnetwork/blob/59d4e3d9e2192524d807b8dbc4271b427adea1ac/app/services/order_cycle_distributed_products.rb#L17-L27

Matt-Yorkley:
With ~500 products in an OC:

![products](https://user-images.githubusercontent.com/9029026/57839579-e7535300-77be-11e9-93a8-a63b22d496c7.png)

lin-d-hop:
First round spike time boxed 1hr max.

Delete dead code (last one) duplicated in #4251

sauloperez:
[...]

PR #4107, opened 2019-07-31
Remove oc distributed products n+1

sauloperez:
#### What? Why?

Related to https://github.com/openfoodfoundation/openfoodnetwork/issues/3853

#### Release notes

### DIFF
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -15,8 +15,11 @@ class OrderCycleDistributedProducts
   #
   # @return [ActiveRecord::Relation<Spree::Product>]
   def relation
+    products = Spree::Product
+      .joins(:variants_including_master)
+      .merge(order_cycle.variants_distributed_by(distributor))
+
     variants = order_cycle.variants_distributed_by(distributor)
-    products = variants.map(&:product).uniq
 
     valid_products = products.reject do |product|
       product_has_only_obsolete_master_in_distribution?(product, variants)
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -15,25 +15,17 @@ class OrderCycleDistributedProducts
   #
   # @return [ActiveRecord::Relation<Spree::Product>]
   def relation
-    Rails.logger.debug "====== relation ======="
     products = Spree::Product
       .joins(:variants_including_master)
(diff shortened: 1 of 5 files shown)


Matt-Yorkley:
The de-N-plus-one-pocalypse has begun! :tada:

sauloperez:
LOL reminds me 



and I'm just following your command. You created the issue :joy:

app/services/order_cycle_distributed_products.rb
```diff
@@ -15,14 +15,25 @@ def initialize(order_cycle, distributor)
...
+      .merge(order_cycle.variants_distributed_by(distributor))
+
+    Rails.logger.debug "====== variants ======="
     variants = order_cycle.variants_distributed_by(distributor)
-    products = variants.map(&:product).uniq
 
+    Rails.logger.debug "====== reject ======="
     valid_products = products.reject do |product|
       product_has_only_obsolete_master_in_distribution?(product, variants)
     end
```
sauloperez: It is this block that causes the N+1s

Matt-Yorkley:
```ruby
hundreds_of.products do |n|
   n+1
end
```
Classic.

Matt-Yorkley: Yeah, I remember digging around in this code for a while trying some basic eager-loading solutions, but didn't manage to crack it.

This whole section looks pretty mad:
https://github.com/openfoodfoundation/openfoodnetwork/blob/59d4e3d9e2192524d807b8dbc4271b427adea1ac/app/services/order_cycle_distributed_products.rb#L17-L27

There's multiple queries of products and variants here in short succession. I was wondering if we could make it into a single complex query, with some fancy `#merge` magic?

Matt-Yorkley: Am I right in thinking there's another query to variants inside of `#product_has_only_obsolete_master_in_distribution`? So we get variants, get products of those variants, get fetch the variants of those products, and then finally query products again? Something like that?

sauloperez: Yep, exactly like that @Matt-Yorkley  

> I was wondering if we could make it into a single complex query, with some fancy #merge magic?

That's what I'm trying but it's painfully hard for me to turn imperative code to declarative :expressionless: :boom:

PR #4057, opened 2019-07-19
Remove dependency to spree_api - step 2 - routes and views

luisramos0:
#### What? Why?

Closes #4054 and #4060 and #4069

This PR finished up what is required to remove spree_api dependency.
The best and most important change here is: "Merge Spree::Api::BaseController with Api::BaseController. All api controllers inherit from Api::BaseController now."

#### What should we test?
This is a very big change, this is a basic verification plan I can think of (there's nothing specific to any of these views being changed except what was discussed in #4111):
- Admin - Config - Taxonomies - Edit taxonomies and use them when creating and editing products
- Admin - Enterprises - Shop Preferences - Category Ordering (make sure the list of taxons is correct)
- Admin - Products - Bulk update products and bulk edit variants
- Admin - Products - Category filter and Category column (view and edit)
- Admin - Products - Upload product image in the bulk product edit page
- Admin - Products - Inventory - update a variant override stock level
- Admin - OC - create and edit order cycles
- Shopfront - change OC in the frontoffice and check that the list of products renders correctly and the filters/search works
- Shopfront - verify the checkout works in particular the entry of addresses
- Admin - Orders - New order - create an order, add variants both normal variants as well as variants with overridden prices and stock and check they are picked up correctly (test the input to search for variants)
- Admin - Orders - Edit page - add, edit quantity and remove variants from the order. Ship the order.

#### Release notes
Changelog Category: Changed
Removed spree_api dependency, OFN API is now self sufficient and can now grow as required using a specific tech stack.

#### Dependencies

This PR goes on top of quite a few PRs that make up the base branch: remove_spree_api_2_base
- #4112 Convert spree/api/shipments views from rabl to ams 
- #4104 Convert spree/api/taxons views from rabl to AMS 
- #4101 Convert spree/api/products and spree/api/variants views from rabl to AMS 
- #4068 Convert order_cycle.rabl to Api::OrderCycleSerializer 
- #4073 Replace update_prod_image.rabl with ImageSerializer in api/product_image_controller 
- #4064 Replace views/admin/variants/search.rabl with Api::Admin::VariantSerializer 
- #4111 Delete now irrelevant authorize_api endpoint and logic (this one is now incorporated in this PR, see [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/4111#issuecomment-521614215) for details)
- #4063 Remove dead code under views/spree/shared 

This PR can only be tested when all PRs above are merged.

PR #4111, opened 2019-07-31
Delete now irrelevant authorize_api endpoint and logic

luisramos0:
OFN API is now authenticating all users, if no session and no key is provided an anonymous user will be created so that user can access public endpoints, authorization is then done at each individual endpoint. This makes this spree api auth call irrelevant

#### What? Why?

Part of #4060
Here we dont need to convert the rabl template, we delete it.

#### What should we test?
We need a sanity check on the bulk product edit page and the inventory page where this code was used.

#### Release notes
Changelog Category: Changed
Delete code not necessary in terms of API authorization.


luisramos0:
yeah, this endpoint was just verifying access to the api, what was happening in the base_controller for this authorize call will also happen for the actual call from angular.
it will authenticate the user, and now we have this magical step:
https://github.com/openfoodfoundation/openfoodnetwork/blob/6a41d9be877ede4052d1c672f13ab78e9bee9693/app/controllers/api/base_controller.rb#L22

RachL:
Here are the test I've carried:

1. As a super admin I've added a product for Fred's farm. After hitting the create button I was redirected here:
image

Usually this means the products are loading (we terribly need a spinner here). I've waited 5 minutes and nothing happened. Even if I've tried to select Fred's product.

I've tried to log out and log back in, with manager and super admin. But nothing, complete desert in terms of products. I've tried Chromium and Firefox. Nothing showing up in the browser.

Same for inventory.

Will tried in staging UK in order to see if I have the same behavior. Before staging the PR I could create a product as super admin and be redirected to the list of all products.

RachL:
@sigmundpetersen wait I was not finished testing it on Uk staging ;)

RachL:
ooops sorry actually @mkllnk is the one who moved it. Did you see something else Maikel?

RachL:
Ok so on UK staging everything is looking fine. Here are the test I did:

- Add products
- Edit products
- Delete product ok - recreating product that has been deleted is working as well
- Duplicate products
- Upload / remove images

Inventory
- Override price
- Override stock levels
- Check that override don't have an effect on the producer stock and price  level
- Reset stock levels

See how everything is looking in the shopfront and checkout the products.

One confusing thing (not linked to this PR) I've found - not sure if this need an issue :
- I created a product as a producer with a on hand stock at 7, and changed my mind and ticked the on-demand box.
- As a hub I set this product in my inventory as "Use producer settings". Currently instead of on demand I see the "7" in the on-hand value which can cause a confusion for the hub manager
- I've checkout the product with 11 quantities, now in inventory the on hand is showing at -11 while for the producer it remains unchanged and at on-demand.
So basically it will always show the quantity ordered... a bit weird I guess.

mkllnk:
Oh, sorry. I thought that your description of products not loading was bad enough to investigate. But you couldn't reproduce that issue? It's all working fine?

RachL:
@mkllnk don't worry. So I had the issue on Katuma staging with this PR. With the exact same PR in UK staging no problem. I've staged something else in ES now and products are back. So I guess I can retry again, maybe something went wrong when staging?

RachL:
Ok this is super weird... Staged on Katuma I again can't see any products (but before stagin gI could). Both super admin and managers. Just to be sure a dev should have a look.

Moving to dev ready as Luis is on holidays.

luisramos0:
ok, I have managed to replicate this problem.
The problem is that if we dont set the api key header in a previous api request, the products request will fail as an anonymous user will be used. This is a problem for spree api requests, not ofn api requests (the problem is already solved in the ofn api base_controller because [it now uses the current user to authenticate on api requests](https://github.com/openfoodfoundation/openfoodnetwork/blob/d5b37a5171a6e5104865c3f56e09be323e1a502a/app/controllers/api/base_controller.rb#L22)).
This problem in spree/api/base_controller is already solved in #4057 as we merge spree/api/base_controller into the ofn api/base_controller, thus, all api requests will try to use the app user to authenticate api requests [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/4057/files#diff-b4f92b4e361ffab4fbb91c68eb3f77bbR50).

SO, I could fix the problem in this PR and then test this and then we would have to test it again in #4057. To make it easier for everyone I am going to close this PR, add these commits in #4057 and we can test this when we test #4057 (4057 completely removes spree_api dependency).

ok?

oh, I think it worked in the other staging environment just because somehow before the products page was loaded there was a previous call to the api with the api key inserted (there are different places of the ofn app where the api key is still inserted as part of the api requests).

PR #4112, opened 2019-07-31
Convert spree/api/shipments views from rabl to ams

luisramos0:
#### What? Why?

Part of #4060 

Convert spree/api/shipments views from rabl to ams, new serializer. And then move the route and controller out of the spree namespace into /api :tada:

re rubocop, in this PR I moved classes and methods but I tried to avoid changing the code itself. Most rubocop issues now would require re-writting the code, imo, can be done when we change logic.

#### What should we test?
Admin orders - list - ship an order in the list of orders
Admin orders - new order - create an order, add variants both normal variants as well as variants with overridden prices and stock and check they are picked up correctly.
Admin orders - edit page - add, edit quantity and remove variants from the order. Ship the order.

#### Release notes
Changelog Category: Changed

Converted one more rabl file to AMS (modernizing tech stack).

#### Dependency
This PR depends on #4055 (the target branch of this PR is remove_spree_api, it's the branch of #4055).


RachL:
I didn't manage to ship an order, if I created the order as an admin. The shipping button just never appeared.

I've tested it in UK staging and I could ship an order created as an admin. There must be something here then. Moving to dev ready as Luis is on holiday.

Staging ES:

image

Staging UK:
image

Testing notes: https://docs.google.com/document/d/1qdu5RKcw0goXpgWWAMdKdN7iPfxXHUV5rNa22Mnd3Q4/edit#

luisramos0:
I think you have found a bug, it's a bug introduced in v2 not in this PR I think.

Even in another staging, you will see you cannot ship an order if the quantity is marked backordered (this happens for on_demand variants). you will only be able to ship if order is completed, payment exists and no backorders exist (basically variants with on_hand value).

this
Screenshot 2019-08-15 at 10 45 48

vs
this
Screenshot 2019-08-15 at 10 45 53

you can only ship an order [if the shipment is in the state ready](https://github.com/openfoodfoundation/openfoodnetwork/blob/d5b37a5171a6e5104865c3f56e09be323e1a502a/app/views/spree/admin/orders/_shipment.html.haml#L9)
and a shipment is only [ready if no backorders exists](https://github.com/openfoodfoundation/spree/blob/8a8585a43cd04d1a50dc65227f337a91b18d66d5/core/app/models/spree/shipment.rb#L207), this is not correct in OFN, if variant is on_demand (backorder) you should be able to ship the order anyway.

I think it will be easy to fix this problem by just removing that backorders logic from the shipment.determine_state method.

luisramos0:
so I think this is ready to go but blocked by #4055 
I created this bug #4158
I think it needs to be confirmed by tester.

Matt-Yorkley:
> you cannot ship an order if the quantity is marked backordered (this happens for on_demand variants)

@luisramos0 does this mean `on_demand` items should keep stock at zero and not reduce it to negative values?

luisramos0:
I think it's that but also all the fill_backorders logic in stock_item and inventory_units... I can only think of copying that spree code to our side and delete it all :-)

luisramos0:
@sigmundpetersen we need a tester to validate this before it's ready to go. ok?

sigmundpetersen:
Yep my bad, wrong column 👍

lin-d-hop:
These 'Blocked' issues in 'Test Ready'....
I'm not sure what to do with them.
@luisramos0... to test or not to test?

luisramos0:
rebased, ready for testing.

sigmundpetersen:
Build failing @luisramos0

luisramos0:
yeah, apparently unrelated error on enterprises page feature test. and it's not failing locally for me.
retrying....

luisramos0:
ok, this last rebase just fixed the random problem... 

I think we should retest this PR because a lot of related changes happened in master since the last test...

RachL:
Admin orders - list - ship an order in the list of orders => All good (except known issue with on demand)
Admin orders - new order - create an order, add variants both normal variants as well as variants with overridden prices and stock and check they are picked up correctly.
=> It is picked up correctly. 
Admin orders - edit page - add, edit quantity and remove variants from the order. Ship the order.

All good for shipping and deleting. However I've noticed that if you have a product with on_hand at 1, you can edit it to the order with quantity of 4. Which gives you afterwards this kind of display:
image


@luisramos0 I've tested this in staging FR with master and I couldn't reproduce... so could it be linked to this PR?

luisramos0:
good catch Rachel!
it can be related, anyway, I can fix that now that we are changing the view.
I wonder: should I remove the description (on_hand or backordered) from the quantity and just keep the number? (this came from spree but for us it''s a bit useless)

RachL:
@luisramos0 yes I was thinking about it. But how would I know that this issue is fixed? :trollface:

daniellemoorhead:
@luisramos0 the branch has conflicts so we haven't deployed to staging.

We'll start part 1 of this testing (better described as noobs learning how it works now) which is to do these actions on a staging server that doesn't have the PR on it and then when the conflict is resolved we'll test it to see if there are any differences.

Ping me when it's ready to go again.

daniellemoorhead:
Oops, my bad, I just realised I was looking at the wrong PR for me and Jen to test - it wasn't this one, it was the one below it. Apologies, pretend I never said anything!

luisramos0:
@RachL "But how would I know that this issue is fixed? :trollface:"
you would only see one number 4, not 3 and 1, would you agree to remove now?

RachL:
@luisramos0 but in this case it shouldn't allow me to add 4. But yes I agree that we can display only quantity. It will be easier for the user. It was useful to stop the bug though :)

luisramos0:
ah, wait, do you mean the product is not on_demand?

RachL:
@luisramos0 yes I've set up a on_hand level of 1. This is why it is showing me there are 3 quantities that are going to be backordered. But I couldn't reproduce this behavior on other staging/production server.

luisramos0:
ok, variant on_hand with stock 1 and you managed to change quantity from 1 to 4 
I'll check.

luisramos0:
rebased to resolved conflict on config/routes/spree.rb

luisramos0:
@RachL I cant see this behaviour if variant is always on_hand. if variant is on_hand when you update the quantity of the line item to a value above stock, it just doesnt change the quantity. No warning to the user: this is the same as in master right now. Maybe we could create an issue to add a error message to the user "quantity not available".

I can replicate the problem (part of the quantity is on_hand and part is backordered) if I set the stock to on_hand and stock to 1 and I add that item to the order, this will be seen as "1 x on_hand". Then I change the variant to on_demand and I use the orders edit page to add more quantity, this time the quantity is added and I can see what you see for example "1 x on_hand 1 x backordered".
I can replicate this behaviour in staging FR where we have master:
Screenshot 2019-09-08 at 12 57 36


SO, I dont think this is not a problem with this PR.
Actually, @Matt-Yorkley is about to fix this in PR #4240 where all quantities added to an order will be marked "on_hand" and so backordered will never show, even if variant is on_demand.

To me it looks like we can move this PR on and just make sure we test this scenario discussed here in #4240

RachL:
@luisramos0 I thought for a brief moment that I managed to reproduced it with a product that was into an inventory, but actually no... sorry I must have been looking at the wrong place. In anyway we will re-test this in the coming PR so moving on with this one, yes!

luisramos0:
🎉

PR #4124, opened 2019-08-04
Add SKU to order confirmation email

### LINKED ISSUE #4102: Print SKU in order confirmation email
For Hub owners that use the SKU for packing, and want to pack as orders arrive via email (i.e. when they see the email confirmations), they need to see the SKU

At the moment the order confirmation email doesn't show SKU

However the invoice printing from admin does. The proposal is to amend the order confirmation email to include the SKU in the same location as it displays on the invoice

SKUs appear on the order confirmation email when they are included in the product. The order confirmation appears as now if the product/variant doesn't have SKU

Matt-Yorkley:
#### What? Why?

Closes #4102 

Add SKU to order confirmation emails.

#### What should we test?

SKUs display in order emails. Shows a `-` in the column where they are missing.

#### Release notes

Added product SKUs to order confirmation emails

Changelog Category: Added

### DIFF
--- a/spec/mailers/order_mailer_spec.rb
+++ b/spec/mailers/order_mailer_spec.rb
@@ -4,6 +4,16 @@ describe Spree::OrderMailer do
   include OpenFoodNetwork::EmailHelper
 
   describe "order confimation" do
+    let(:bill_address) { create(:address) }
+    let(:distributor_address) { create(:address, address1: "distributor address", city: 'The Shire', zipcode: "1234") }
+    let(:distributor) { create(:distributor_enterprise, address: distributor_address) }
+    let(:shipping_instructions) { "pick up on thursday please!" }
+    let(:ship_address) { create(:address, address1: "distributor address", city: 'The Shire', zipcode: "1234") }
+    let(:order) {
+      create(:order_with_line_items, distributor: distributor, bill_address: bill_address, ship_address: ship_address,
+                     special_instructions: shipping_instructions)
+    }
+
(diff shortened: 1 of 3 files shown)


spec/mailers/order_mailer_spec.rb
```diff
@@ -13,41 +23,39 @@
...
         expect(ActionMailer::Base.deliveries.count).to eq(1)
-        expect(ActionMailer::Base.deliveries.first.to).to eq([@order1.email])
+        expect(ActionMailer::Base.deliveries.first.to).to eq([order.email])
+      end
+
+      it "should include SKUs" do
+        mail = Spree::OrderMailer.confirm_email_for_customer(order.id)
+
+        expect(mail.body.encoded).to include "SKU"
+        expect(mail.body.encoded).to include order.line_items.first.variant.sku
```
sauloperez: view specs are particularly suited for these things.

RachL:
I staged it both in UK and FR because I wanted to check how it goes with FR translation "Référence produit". Which is way longer than SKU. But the translation is not picked up yet.

@Matt-Yorkley  do you know if it will create a problem to have a longer title there? I have no idea how it was translated in German as well.

Otherwise this is good to go.

Manager:
image

Buyer:
image

Matt-Yorkley:
> do you know if it will create a problem to have a longer title there?

It's a regular HTML table, so it'll resize the columns to fit the content. Should be fine.

PR #4125, opened 2019-08-04
Variant overrides with on_demand

### LINKED ISSUE #4122: When an inventory override stock level to ondemand, the producer stock on hand is reduced as well
## Description

As part of the release testing I gave Marys'online shop the permissions to sell Pop release's products.

Pop release had choco pops on demand and Limonade on hand (0 left).

I left choco pops as "use producer settings" on Mary's inventory and I switch Limonade "on demand" in Mary's inventory. I noticed two things:

1. Mary does not see that Choco pops are "on demand", she only sees the stock level. Which is confusing.
2. If a shoper takes 34 limonade, the producer stock gets diminished, but it shouldn't

## Steps to Reproduce
1. As a producer, create a product with a on hand stock number
2. Give permission to a shop to add to inventory and to OC
3. As the shop; override the on hand product to "on demand"
4. As a buyer, checkout the product on the hub shop with a quantity higher than the on hand level you've set as a producer
5. Check the producer stock level: it now shows a negative quantity. 

## Animated Gif/Screenshot

Mary's inventory before checkout:

Pop's release stock after checkout on Mary's shop not on Pop release's shop:

## Severity

## Your Environment
* Version used: v2.2.2 (release testing)

Matt-Yorkley:
#### What? Why?

Closes #4122 

When an override had `on_demand: true`, the variant's stock was being decremented when it was sold.

#### What should we test?

When a distributor sells a variant override with `on_demand`set to `true`, it doesn't reduce the stock levels of the producer's variant.

#### Release notes

Fixed a stock issue with inventory using `on_demand`

Changelog Category: Fixed

### DIFF
--- a/lib/open_food_network/scope_variant_to_hub.rb
+++ b/lib/open_food_network/scope_variant_to_hub.rb
@@ -41,7 +41,11 @@ module OpenFoodNetwork
       #   - updates variant_override.count_on_hand
       #   - does not create stock_movement
       #   - does not update stock_item.count_on_hand
+      # If it is a variant override with on_demand:
+      #   - don't change stock or call super (super would change the variant's stock)
       def move(quantity, originator = nil)
+        return if @variant_override.andand.on_demand
+
         if @variant_override.andand.stock_overridden?
           @variant_override.move_stock! quantity
         else
--- a/spec/features/consumer/shopping/variant_overrides_spec.rb
+++ b/spec/features/consumer/shopping/variant_overrides_spec.rb
@@ -16,12 +16,14 @@ feature "shopping with variant overrides defined", js: true do
   let(:product1) { create(:simple_product, supplier: producer) }
   let(:product2) { create(:simple_product, supplier: producer) }
   let(:product3) { create(:simple_product, supplier: producer, on_demand: true) }
+  let(:product4) { create(:simple_product, supplier: producer) }
   let(:product1_variant1) { create(:variant, product: product1, price: 11.11, unit_value: 1) }
   let(:product1_variant2) { create(:variant, product: product1, price: 22.22, unit_value: 2) }
   let(:product2_variant1) { create(:variant, product: product2, price: 33.33, unit_value: 3) }
   let(:product1_variant3) { create(:variant, product: product1, price: 44.44, unit_value: 4) }
   let(:product3_variant1) { create(:variant, product: product3, price: 55.55, unit_value: 5, on_demand: true) }
   let(:product3_variant2) { create(:variant, product: product3, price: 66.66, unit_value: 6, on_demand: true) }
+  let(:product4_variant1) { create(:variant, product: product4, price: 77.77, unit_value: 7) }
   let!(:product1_variant1_override) { create(:variant_override, :use_producer_stock_settings, hub: hub, variant: product1_variant1, price: 55.55, count_on_hand: nil, default_stock: nil, resettable: false) }
   let!(:product1_variant2_override) { create(:variant_override, hub: hub, variant: product1_variant2, count_on_hand: 0, default_stock: nil, resettable: false) }
   let!(:product2_variant1_override) { create(:variant_override, hub: hub, variant: product2_variant1, count_on_hand: 0, default_stock: nil, resettable: false) }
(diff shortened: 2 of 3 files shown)


lib/open_food_network/scope_variant_to_hub.rb
```diff
@@ -41,7 +41,11 @@ def on_demand
       #   - updates variant_override.count_on_hand
       #   - does not create stock_movement
       #   - does not update stock_item.count_on_hand
+      # If it is a variant override with on_demand:
+      #   - don't change stock or call super (super would change the variant's stock)
       def move(quantity, originator = nil)
+        return if @variant_override.andand.on_demand
```
sauloperez: I think it's worth unit testing as well because someday (I hope) we'll want to refactor this

Matt-Yorkley:
I've got a single test failing locally and in CI. It still fails if I revert all the changes in this PR...? :thinking:

lib/open_food_network/scope_variant_to_hub.rb
Matt-Yorkley: I got a lot of errors trying to unit test this, as I think there needs to be line items and stock locations set up for it to work.

Matt-Yorkley: Done

spec/features/consumer/shopping/variant_overrides_spec.rb
```diff
@@ -146,6 +149,15 @@
       product1_variant1_override.reload.count_on_hand.should be_nil
     end
 
+    it "does not subtract stock from variants where the override has on_demand: true" do
```
mkllnk: I really like how you wrote the test first and have it in a separate commit. :+1:

I think that it's even better to declare a spec as pending when you know that it fails because of a bug. So you can run the specs and they "pass". The commit that fixed the bug then changes the spec from `pending` to `it` and you clearly see in the commit history which change fixed what spec (and you can rely on CI and `git bisect` when needed). But maybe that's just my preference?

daniellemoorhead:
>I just would like to flag that we have some inconsistent stock tracking now. Variants track the stock level even if they are on demand (backorderable). But VariantOverrides don't. That will confuse people in the long term and we need to plan our VariantOverride exit.

@rachel @lin-d-hop @sauloperez probs something to be reviewed and fixed?

daniellemoorhead:
@rachel I started trying to test this but I got stuck (admin interface isn't my strong point yet). I've staged it on UK and hoping we could go through it together after the delivery train catch up today?

RachL:
@daniellemoorhead sure, let's do that!

RachL:
> 
> 
> > I just would like to flag that we have some inconsistent stock tracking now. Variants track the stock level even if they are on demand (backorderable). But VariantOverrides don't. That will confuse people in the long term and we need to plan our VariantOverride exit.

> @Rachel @lin-d-hop @sauloperez probs something to be reviewed and fixed?

yes. I saw that two and got confused. Not sure how to tackle it yet though

sigmundpetersen:
@RachL @daniellemoorhead see Maikel's comment in [4122](https://github.com/openfoodfoundation/openfoodnetwork/issues/4122#issuecomment-518429210) as well. This is something that needs to be incepted and a great chance for a first story of the Network feature! :bowtie:

Matt-Yorkley:
From @mkllnk's comment in the issue:

> If we take that as the desired behaviour then we need to track stock on the variant override even though it's on-demand.

We could do that really easily here, if that's what we want. So to sum it up, when an override with `on_demand` is sold:

**In master:** variant stock is reduced, override stock is not reduced.
**In this PR:** neither are reduced.
**mklink's proposal:** variant stock is not reduced, override stock is reduced.

What do people think?  @kristinalim @sauloperez @RachL @lin-d-hop

RachL:
Maikel's proposal would make everything consistent. I find it always great to not have particular rules, which we forget with time.
But at the same time, I have the feeling we might be taking too much time on this and that the current solution is good enough?

A bit of mixed feelings so :) I would be happy either way I guess.

RachL:
We've tested with Danni this morning and I did a couple of more test this afternoon.

Testing notes: https://docs.google.com/document/d/1su3OWV1VfU-cgLpTaDLHOchbh9kF57VeqEjIRYadEgo/edit

This is ready to go.

mkllnk:
I moved this to test-ready because Pau's concerns had been addressed already and I wanted to speed this issue up because it's an S1.

The concern I raised needs some inception. I still approved this pull request because the current solution is solving the immediate problem. I will merge it. We can always improve things later.

luisramos0:
great @Matt-Yorkley 👏 

I think this bug only applies when the override is on_demand **and the override count_on_hand is null**, so I guess  to follow @mkllnk 's suggestion we would need to copy the variants stock and move it, something like this:

```ruby
def move(quantity, originator = nil)
       
         if @variant_override.andand.stock_overridden?
          @variant_override.move_stock! quantity	     
        else	     
            if @variant_override.andand.on_demand
                @variant_override.count_on_hand = 0
                @variant_override.move_stock! total_on_hand    
            else
                 super	
            end
        end	       
      end	     
```

anyway, I think it's better to keep it as you guys did!

PR #4132, opened 2019-08-06
Product Import beta

### LINKED ISSUE #2603: [Product Import] Make product import submenu visible for all users + beta sentence

Matt-Yorkley:
#### What? Why?

Closes #2603 

Makes the Product Import page visible for non-sysadmin users.

#### What should we test?

Product Import is now accessible to non-superadmin users.

#### Release notes

Product Import feature is now launched and in beta

Changelog Category: Added


app/helpers/spree/admin/navigation_helper_decorator.rb
```diff
@@ -18,6 +18,7 @@ def klass_for_with_sym_fallback(name)
         klass = Spree::Order if klass == :bulk_order_management
         klass = EnterpriseGroup if klass == :group
         klass = VariantOverride if klass == :Inventory
+        klass = ProductImport::ProductImporter if klass == :import
```
sauloperez: can you explain how is this used? I found out this helper is used to create tab links? but I only see a single use in our codebase

Matt-Yorkley: It's overriding a Spree helper, used in 5 files under `app/views/admin/shared`, wherever you see `= tab`. It probably needs a big tidy up at some point, but I didn't think it was a priority right now.

Matt-Yorkley: It applies `cancan` permissions to see whether or not it should display items in the submenus for the current user.

sauloperez: Would you mind adding the necessary Rubocop's `disable` to get it green in Code Climate while we are aware of the issue in the future?

daniellemoorhead:
I have no idea where I'll see this option...but I will do my best to figure it out. I will no doubt find it!

@Matt-Yorkley a screenshot would be helpful for next time :)

daniellemoorhead:
I found it, it was easy, I really didn't need a screenshot. Phew.

OK, so here is the 'import' link on a new shop for a new user I created (thus not as a super admin)
Screen Shot 2019-08-19 at 7 05 17 pm

And here is my shop on OFN AU, that doesn't have the link:

Screen Shot 2019-08-19 at 7 05 25 pm

So, this is good to go!

Issue #4127, opened 2019-08-06
[Spike] Check which release tests are already automated


sauloperez:
## Description
**- As a:** core team member
**- I want to be able to do:** gain efficiency and confidence on the release process so that I can focus my efforts on bringing more value to our users. 

## Acceptance Criteria

We need to get a list of the tests that are performed as a release testing (https://docs.google.com/document/d/15EdXKb4B3BS8gekD6_9PDdPTZsjUCAVS2011zRf55lM/edit#heading=h.alpdtdl2tc74) that are already included in our automated test suite and how well covered they are.

Then, we can decide which ones we stop testing manually.

daniellemoorhead:
I was talking to @mkllnk about this last week. We thought we could make a list and then each release we do we check off one of them. If it's not automated, we add that to be automated. If it is then we wait till the next release to check the next one.

@mkllnk did I remember correctly?

RachL:
Yep! But this issue is *only* about the list :) We would need separate issue for the automated tests.

sauloperez:
yep, not jumping into implementations yet. But yes, that iterative process you mention seems really good @daniellemoorhead

sauloperez:
I copy here the contents of the [Testing Handbook v2](https://docs.google.com/document/d/15EdXKb4B3BS8gekD6_9PDdPTZsjUCAVS2011zRf55lM/edit#heading=h.alpdtdl2tc74) for reference

## How to do a sanity check (release testing)?

- [ ] Check signing up for a customer or a producer
- [ ] Check that the producer account is correctly displayed on the map
- [ ] Complete your enterprise setting and check that both payment method and shipping method works
- [ ] Products : create, edit (add an image for example), duplicate (check that image is also duplicated), delete
- [ ] Can you create on OC? Does it show in front office? Can you duplicate / edit / delete it?
- [ ] Check out: test all method payments
- [ ] Order for producer: can you see it? Can you filter on the page? Can you ship it? Can you capture payments?
- [ ] Customers : add a customer and edit its credentials afterwards

### Subscriptions

- [ ] Can enable from superadmin?
- [ ] Can create schedule?
- [ ] Can trigger subs?
- [ ] Get emails?
- [ ] Out of stock messaging works?
- [ ] Editing sub, pausing sub?

### Inventory

- [ ] Override price
- [ ] Override stock levels
- [ ] Reset stock levels
- [ ] Embedded shops. 
[...]

PR #4135, opened 2019-08-07
Remove bugherd specs

sauloperez:
#### What? Why?

This is a quick cleanup that came up while working on https://github.com/openfoodfoundation/openfoodnetwork/issues/4127.

I don't think these are worth the time it takes to execute them every time but most importantly I believe we are not using Bugherd anymore.

#### What should we test?

Nothing. This only removes tests from the automated test suite.

#### Release notes

Remove old and unused Bugherd integration tests.

Changelog Category: Removed


sauloperez:
It might be worth adding an issue to completely remove Bugherd if we agree to.

PR #4136, opened 2019-08-08
Lock variants during checkout to avoid race condition

### LINKED ISSUE #4018: Two orders for the same (last) limited stock 'on hand' item are processed and stock set to -1
I am testing #3925 in V2. It appears that the previous problematic behaviour has gone (yay) but we have a new problem - it is allowing backorders and I can't turn that off?

I have one cheese in stock

I have two customers who have both placed cheese in the cart, set payment to Stripe / credit card and hit the 'Place Order Now' button at the same time.

Now, instead of both payments being taken but one order disappearing, both payments are taken and BOTH orders are processed.

So I have two successfully completed orders

And I have a stock level of -1

NB. If I now reload the shop, the cheese is still there even though there is no stock. Suggesting that 'show out-of-stock items' is set to yes :( If i put this into my cart I get immediately notified that I can't buy it - and some weird behaviour (that I will document if needed)

## Description

In V1 I had a Configuration option as super-admin to set 'Inventory Settings', which I/Aus instance and I think most (all?) other instances have set to NOT show out-of-stock items or Allow Backorders. When this was accidentally left checked (allowing backorders) people had problems.

I can no longer see this option in my Configuration panel in V2. I recall some discussions about this and am not surprised that the option is gone. I am concerned that perhaps the default has been set the wrong way in its absence? We NEED to disallow backorders as without it OFN stock management basically disappears for anyone not using 'on demand', which is any small farmer / hub working with small farmers . . 

Similarly with showing out-of-stock items - the default on this has been no, and using stock management to zero to ensure products aren't in the shop front is widespread if not standard behaviour in OFN users. Without this, they would have to manage all product availability via order cycles, which is unworkable
 
I am not sure if this is a bug, or if a decision was taken to do it this way? I am pretty concerned about it though - this would be a major loss of functionality for many of our users, likely making OFN un-useable for them. 

## Expected Behavior
If there is only one item in stock, it is only sold to one Customer. The other Customer is told "Sorry this item is out of stock" and can remove from cart and then place order

[If an item is out of stock, it is not shown in the Shopfront.]

## Actual Behavior
The item is sold twice. There is no notification to either Customer or Hub. Hub would need to monitor all products / variants to see which ones have gone negative and them amend all the relevant orders, process refunds etc as there will not be enough stock to fulfil the orders.

## Steps to Reproduce
As above
1. Create product with 1 in stock, put it in order cycle, open shop 
2. Two customers place item in cart  
3. Two customers hit Place Order Now at the same time
NB. I tested this with Stripe / credit card . . haven't replicated with other payment methods - suspect it would be exactly the same as the problem is in the inventory handling rather than the payment. Can check / confirm this if needed
4. Both orders are processed, both payments are taken, no notification

## Animated Gif/Screenshot
See screenshots above

## Context
I think that this change of setting nullifies OFN stock management in a major way. I don't think we can deploy to Aus until it is fixed. Curious about if / whether this affects other instances in the same way? Surely our users aren't the only ones that actually use stock levels?

## Severity
?? I'd be calling this S1 for Aus - ability to manage stock is a critical feature
Perhaps could argue S3 if they use order cycles to manage stock, but I really can't see that that is workable for most

## Your Environment
MACOS

## Possible Fix
Change the default inventory settings to how they have always / most commonly been set i.e. 
Allow Backorders - NO
Show out-of-stock items in shop - NO

mkllnk:
#### What? Why?

Closes #4018

Lock variants during checkout to avoid race condition

It was possible that several people bought the same variant even though there wasn't enough stock for everybody. That resulted in negative stock.

#### What should we test?

* Open an order cycle with a product that has stock `1`.
* Add that item to the your cart and go to the checkout, ready to click on the checkout button.
* Open a private window and add the same item to your cart and be ready to click on the checkout button.
* Arrange the windows so that you can click both checkout buttons very quickly, roughly within half a second.
* One order should succeed.
* The other order should fail and display an out-of-stock error message.

#### Release notes

Fixed: Multiple concurrent checkouts for the same variant are now waiting for each other and not reducing stock into the negative.

Changelog Category: Fixed


mkllnk:
I looked at this again and found why the specs were failing. The around filter tried to access `@order` before it was assigned in a later before filter.

This made me think about this a bit more. Initially I published this draft PR to show the concept and how easy it can be to solve the problem. But with little mistakes, this code is not effective at all. I will have another go at writing a spec for this. It will probably become complex, setting up multiple threads with multiple breakpoints to make sure we reliably reproduce the race condition plus effective tests. I'm not sure I will succeed. But I think that it's worth trying because race conditions during checkout are probably the worst race conditions of the application.

spec/controllers/checkout_controller_spec.rb
```diff
@@ -266,4 +266,80 @@
...
+          {
+            "payment_method_id" => payment_method.id,
+            "amount" => order.total
+          }
+        ],
+        "bill_address_attributes" => address_params,
+        "ship_address_attributes" => address_params,
+      }
+
+      # Starting two checkout threads. The first one will wait at the breakpoint.
```
luisramos0: I am curious about this: the breakpoint lock and synchronize instructions above will only be taken by the first thread? is that what this means?

spec/controllers/checkout_controller_spec.rb
```diff
@@ -266,4 +266,80 @@
...
+      # The second waits for the first one if it uses correct synchronisation.
```
luisramos0: I dont understand this part "if it uses correct synchronisation" maybe you can clarify in the comment?

spec/controllers/checkout_controller_spec.rb
mkllnk: Yes. If we didn't have the around filter then both threads would wait at the breakpoint. But since I implemented the locking variants in the around filter, the second thread will wait there and not get the breakpoint until the first thread is finished.

spec/controllers/checkout_controller_spec.rb
mkllnk: The spec looks at two cases:
1. The variant locking is implemented and everything works as it should. That's when "it uses correct synchronisation". The test passes.
2. The variant locking is not implemented (correctly) and the second thread doesn't wait for the first one. They end up in a race condition and the spec fails.

mkllnk:
> I wonder how often this new spec fails if you remove the new prepend_around_filter.

I ran it quite a few times and it never passed. The failure I had most of the time was a failure I have seen in Bugsnag before:

```
  1) CheckoutController concurrency waits for concurrent checkouts
     Failure/Error: order.next
     
     ActiveRecord::RecordNotUnique:
       PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_spree_shipments_on_order_id"
       DETAIL:  Key (order_id)=(113) already exists.
       : INSERT INTO "spree_shipments" ("address_id", "cost", "created_at", "number", "order_id", "shipped_at", "state", "stock_location_id", "tracking", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10) RETURNING "id"
     # ./app/controllers/checkout_controller.rb:177:in `advance_order_state'
     # ./app/controllers/checkout_controller.rb:42:in `update'
     # ./spec/controllers/checkout_controller_spec.rb:327:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # PG::UniqueViolation:
     #   ERROR:  duplicate key value violates unique constraint "index_spree_shipments_on_order_id"
     #   DETAIL:  Key (order_id)=(113) already exists.
     #   ./app/controllers/checkout_controller.rb:177:in `advance_order_state'
```
https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5d6b014950e7c00019884f18

spec/controllers/checkout_controller_spec.rb
luisramos0: ah, I get it, thanks for explaining!

spec/controllers/checkout_controller_spec.rb
luisramos0: oh I see, "if it uses correct sync" means "if we use locking for the process"

luisramos0:
nice, it's really cool we can test this and have a spec that validates it!!!

mkllnk:
@luisramos0 I addressed your issues in the last two commits. It's ready for review again.

app/services/current_order_locker.rb
```diff
@@ -0,0 +1,40 @@
+# Locks a controller's current order including its variants.
+#
+# It should be used when making major changes like checking out the order.
+# It can keep stock checking in sync and prevent overselling of an item.
+class CurrentOrderLocker
+  # This interface follows the ActionController filters convention:
```
Matt-Yorkley: Nice, I haven't seen this before :smile:

filipefurtad0:
Hi @mkllmk,

Followed the testing procedure, just as described:

- Opened an order cycle with a product that has stock 1.
- Added that item to the your cart and go to the checkout, ready to click on the checkout button.
- Open a private window and add the same item to your cart and be ready to click on the checkout button.
- Arrange the windows so that you can click both checkout buttons very quickly, roughly within half a second.

I tested this as above, and also using different accounts, on two different browsers, simultaneously. It all worked as expected:

- One order succeeded. -> Check.
- The other order should fail and display an out-of-stock error message. -> This here only happened with different accounts. When using the same account a redirection to the homepage occured, but no error message was shown. In both cases, only one article was bought, as confirmed by the emails.

Congratulations! :tada:
All good -> Moving to Ready to Go.


PS: At first, I got a snail 404 - page not found - at the moment of checking-out, but this happened during PR-deployment, which probably caused this. After deployment the 404 disappeared.

luisramos0:
great testing!!!

mkllnk:
Great, thank you Filipe!

PR #4141, opened 2019-08-08
OC performance

### LINKED ISSUE #4007: [Performance] admin edit order cycles queries 2

Part 2 of the N+1 issues with Admin Order Cycles. See spike here: #3858

**Pain point:** `/admin/order_cycles/<order_cycle_id>.json ` (show)

This uses the `app/serializers/api/admin/order_cycle_serializer.rb` serializer, and has more N+1 queries

```
lib/open_food_network/order_cycle_permissions.rb:27:in `visible_enterprises'
app/serializers/api/admin/order_cycle_serializer.rb:37:in `editable_variants_for_incoming_exchanges'
app/controllers/spree/admin/base_controller_decorator.rb:81:in `render_as_json'
app/controllers/admin/order_cycles_controller.rb:24:in `block (2 levels) in show'
app/controllers/admin/order_cycles_controller.rb:21:in `show'

SELECT DISTINCT "enterprises"."id" FROM "enterprises" INNER JOIN "exchanges" ON "enterprises"."id" = "exchanges"."sender_id" WHERE "exchanges"."order_cycle_id" = 200066 AND "exchanges"."incoming" = 't'; 
```

There's some N+1 queries here that need fixing, in `app/serializers/api/admin/order_cycle_serializer.rb`, in the methods: `#editable_variants_for_incoming_exchanges` and `#editable_variants_for_outgoing_exchanges`.

### Context

This is related to the admin edit order cycles spike: #3858

Matt-Yorkley:
#### What? Why?

Closes #4007

Poor performance with the endpoint: `/admin/order_cycles/<order_cycle_id>.json`.

I've tidied up some of the related code a bit, and removed some repeating queries. There's still a _lot_ of refactoring needed in these two files (tech debt):

[lib/open_food_network/order_cycle_permissions.rb](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/open_food_network/order_cycle_permissions.rb)
[app/serializers/api/order_cycle_serializer.rb](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/serializers/api/order_cycle_serializer.rb)

I used an OC with ~20 producers and ~500 products. Initially this endpoint was making `651` queries, it's now at `375`. I don't know if I can get it down much further. 

`Before:`
Screenshot from 2019-08-09 09-16-41
`After:`
Screenshot from 2019-08-09 11-43-59

~42% less queries :tada: 

#### What should we test?

Order Cycles admin functionality is working as before (create, view, update, etc).

#### Release notes

Improved performance on an order cycles API endpoint.

Changelog Category: Changed

### DIFF
--- a/app/serializers/api/admin/order_cycle_serializer.rb
+++ b/app/serializers/api/admin/order_cycle_serializer.rb
@@ -33,7 +33,7 @@ class Api::Admin::OrderCycleSerializer < ActiveModel::Serializer
   end
 
   def editable_variants_for_incoming_exchanges
-    sort_by_supplier_id(permissions.all_incoming_editable_variants.all)
+    variant_ids_by_supplier_id(permissions.all_incoming_editable_variants.all)
   end
 
   def editable_variants_for_outgoing_exchanges
@@ -79,17 +79,10 @@ class Api::Admin::OrderCycleSerializer < ActiveModel::Serializer
     @visible_enterprises ||= permissions.visible_enterprises
   end
 
-  def sort_by_supplier_id(variants)
-    collection = {}
-    variants.map do |variant|
-      supplier_id = variant.product.supplier_id
-
-      if collection.key? supplier_id
-        collection[supplier_id] << variant.id
-      else
-        collection[supplier_id] = [variant.id]
-      end
+  def variant_ids_by_supplier_id(variants)
+    grouped_by_supplier = variants.group_by(&:supplier_id)
+    grouped_by_supplier.each do |supplier_id, grouped_variants|
+      grouped_by_supplier[supplier_id] = grouped_variants.map(&:id)
     end
-    collection
   end
 end
--- a/app/serializers/api/admin/order_cycle_serializer.rb
+++ b/app/serializers/api/admin/order_cycle_serializer.rb
@@ -33,14 +33,7 @@ class Api::Admin::OrderCycleSerializer < ActiveModel::Serializer
   end
 
   def editable_variants_for_incoming_exchanges
-    # For each enterprise that the current user is able to see in this order cycle,
-    # work out which variants should be editable within incoming exchanges from that enterprise
-    editable = {}
-    visible_enterprises.each do |enterprise|
-      variants = permissions.editable_variants_for_incoming_exchanges_from(enterprise).pluck(:id)
-      editable[enterprise.id] = variants if variants.any?
-    end
-    editable
+    sort_by_supplier_id(permissions.all_incoming_editable_variants.all)
   end
 
   def editable_variants_for_outgoing_exchanges
@@ -84,5 +77,19 @@ class Api::Admin::OrderCycleSerializer < ActiveModel::Serializer
 
   def visible_enterprises
     @visible_enterprises ||= permissions.visible_enterprises
+  end
+
+  def sort_by_supplier_id(variants)
+    collection = {}
+    variants.map do |variant|
+      supplier_id = variant.product.supplier_id
+
+      if collection.key? supplier_id
+        collection[supplier_id] << variant.id
+      else
+        collection[supplier_id] = [variant.id]
+      end
+    end
+    collection
   end
 end
--- a/app/serializers/api/admin/order_cycle_serializer.rb
+++ b/app/serializers/api/admin/order_cycle_serializer.rb
@@ -25,10 +25,7 @@ class Api::Admin::OrderCycleSerializer < ActiveModel::Serializer
   end
 
   def exchanges
-    scoped_exchanges =
-      OpenFoodNetwork::OrderCyclePermissions.
-        new(options[:current_user], object).
-        visible_exchanges.by_enterprise_name
+    scoped_exchanges = permissions.visible_exchanges.by_enterprise_name
 
     ActiveModel::ArraySerializer.
       new(scoped_exchanges, each_serializer: Api::Admin::ExchangeSerializer,
@@ -39,9 +36,7 @@ class Api::Admin::OrderCycleSerializer < ActiveModel::Serializer
     # For each enterprise that the current user is able to see in this order cycle,
     # work out which variants should be editable within incoming exchanges from that enterprise
     editable = {}
-    permissions = OpenFoodNetwork::OrderCyclePermissions.new(options[:current_user], object)
-    enterprises = permissions.visible_enterprises
-    enterprises.each do |enterprise|
+    visible_enterprises.each do |enterprise|
       variants = permissions.editable_variants_for_incoming_exchanges_from(enterprise).pluck(:id)
       editable[enterprise.id] = variants if variants.any?
     end
@@ -52,9 +47,7 @@ class Api::Admin::OrderCycleSerializer < ActiveModel::Serializer
     # For each enterprise that the current user is able to see in this order cycle,
     # work out which variants should be editable within incoming exchanges from that enterprise
     editable = {}
-    permissions = OpenFoodNetwork::OrderCyclePermissions.new(options[:current_user], object)
-    enterprises = permissions.visible_enterprises
-    enterprises.each do |enterprise|
+    visible_enterprises.each do |enterprise|
       variants = permissions.editable_variants_for_outgoing_exchanges_to(enterprise).pluck(:id)
       editable[enterprise.id] = variants if variants.any?
     end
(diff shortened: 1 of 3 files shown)


Matt-Yorkley:
Notes: 

- Using an OC with ~20 producers and ~500 products.
- Working on the specific endpoint `/admin/order_cycles/<id>.json`
- Total queries were initially around ~670
- Currently down to ~230

Matt-Yorkley:
I also think the Angular for edit Order Cycles needs some serious work, but have no idea what tshirt size to give it. I found this code comment, which kind of sums it up:

https://github.com/openfoodfoundation/openfoodnetwork/blob/86aeb6a3c796ae09e15b5a576b4e8424e076aa37/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee#L79-L84

I think these will both need looking at:

[app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee)
[app/assets/javascripts/admin/order_cycles/services/enterprise.js.coffee](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/admin/order_cycles/services/enterprise.js.coffee)

app/serializers/api/admin/order_cycle_serializer.rb
```diff
@@ -74,4 +61,28 @@ def visible_variants_for_outgoing_exchanges
...
+      supplier_id = variant.product.supplier_id
+
+      if collection.key? supplier_id
+        collection[supplier_id] << variant.id
+      else
+        collection[supplier_id] = [variant.id]
+      end
+    end
+    collection
+  end
```
sauloperez: we got to start trusting our DB to work with data if we ever want to have decent performance. I know it's sometimes hard but there's no way around it. Have you tried with something like

```rb
variants
  .joins(:product)
  .order(:supplier_id)
  .select('supplier_id, spree_variants.*')

variants.group_by(&:supplier_id)
```

It'll return a hash like

```rb
"205"=>
  [#<Spree::Variant id: 3015, sku: "", weight: nil, height: nil, width: nil, depth: nil, deleted_at: nil, is_master: true, product_id: 1391, cost_price: nil, position: 1, cost_currency: "EUR", unit_value: 1.0, unit_description: "", display_name: nil, display_as: nil, import_date: nil>,                                                                               
   #<Spree::Variant id: 3016, sku: "", weight: nil, height: nil, width: nil, depth: nil, deleted_at: nil, is_master: false, product_id: 1391, cost_price: nil, position: 2, cost_currency: "EUR", unit_value: 1.0, unit_description: "", display_name: nil, display_as: nil, import_date: nil>]}
```

Because I assume that's the structure we need for the client, right? `Enumerable#group_by` does have a cost...

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -154,49 +151,30 @@ def editable_variants_for_incoming_exchanges_from(producer)
       end
     end
 
+    def all_incoming_editable_variants
+      valid_suppliers = visible_enterprises.map do |enterprise|
+        enterprise.id if user_manages_coordinator_or(enterprise)
+      end
+
+      Spree::Variant.includes(product: :supplier).
+        joins(:product).where('spree_products.supplier_id IN (?)', valid_suppliers)
```
sauloperez: you should be able to rely on AR syntax to get the supplier id but also isn't the following query enough?

```rb
Spree::Variant
  .joins(:product)
  .where(supplier_id: valid_suppliers)
```

given `supplier_id` is a column of `spree_products`.

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -154,49 +151,30 @@ def editable_variants_for_incoming_exchanges_from(producer)
...
+    end
```
sauloperez: We got to stop making this class longer. It's probably one of the few things we are sure won't do us any good. We should extract that into its own query objects or services or whatever you want. Besides, we already know this class is the source of several performance problems and needs to be seriously reconsidered.

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -252,18 +206,54 @@ def editable_variants_for_outgoing_exchanges_to(hub)
...
+
+    def active_outgoing_variants(hub)
+      @active_outgoing_variants ||= begin
+        active_variants = []
+        @order_cycle.exchanges.outgoing.where(receiver_id: hub).limit(1).each do |exchange|
+          active_variants = exchange.variants
+        end
+        active_variants
+      end
```
luisramos0: there'a a limit(1) in there. it means first_exchange, right?

```suggestion
     def active_outgoing_variants(hub)
      @active_outgoing_variants ||= begin
        
        first_outgoing_exchange = @order_cycle.exchanges.outgoing.where(receiver_id: hub).limit(1)
        if first_outgoing_exchange.present?
           first_outgoing_exchange.variants            
        else
              []
        end
      end
```

app/serializers/api/admin/order_cycle_serializer.rb
Matt-Yorkley: The hash has to be variant ids grouped by supplier ids, but yes I'll take a look at tidying it up.

sauloperez:
I think it'd be reasonable to pair with someone touching these deep parts of the data layer. There's a lot going on in these classes and a lot can go wrong. I would prefer two pairs of eyes. Don't you think?

sauloperez:
IMO it'd be wise to split this into smaller PRs so that we can have a good throughput of performance improvements including in releases. Also, we are touching quite deep stuff so the risk is high.

As I experienced in https://github.com/openfoodfoundation/openfoodnetwork/pull/4147 there are lots of easter eggs in the data layer that might cause PRs to take very long thus delaying little improvements to hit production.

Matt-Yorkley:
I submitted this a while ago and it got put aside, but I think it's worth getting some of this tidying up effort started before the other OC work begins. I'll take a look at finishing and resubmitting it.

lib/open_food_network/order_cycle_permissions.rb
Matt-Yorkley: I think we can simplify this further, into a one-liner:
```ruby
@order_cycle.exchanges.outgoing.where(receiver_id: hub).first.andand.variants || []
```

luisramos0:
all right :+1: is it ready for re-review?

sauloperez:
Yep, let us know. Feel free to ship this in two stages if you think it brings value. This way keep a high throughput of perf. improvements every week :ok_hand:

app/serializers/api/admin/order_cycle_serializer.rb
mkllnk: I think that we should change the name because it's not sorting anything. It's `group_variant_ids_by_supplier_id`.

I like Pau's recommendation to use group_by. Changing the variant objects to their ids makes it a bit more complicated. But if it needs to be:

```rb
variants.group_by(&:supplier_id).map { |supplier_id, variants| [supllier_id, variants.map(&:id)] }.to_h
```

It's still a bit complicated and apparently using a lot of memory. A way to use less would be this:
```rb
grouped = variants.group_by(&:supplier_id)
grouped.each { |supplier_id, variants|  grouped[supllier_id] = variants.map(&:id) }

# or shorter but lesser known method:
grouped.update(grouped) { |_, variants| variants.map(&:id) }
```

This is a lot simpler and then grouping is probably very fast in Ruby. But it's difficult to tell if it's actually better than Matt's version which iterates only once and doesn't create any intermediate objects.

I'm okay with keeping the current version, just rename the method. And in both cases we can do a little optimisation:

```rb
variants.select("spree_variants.id, spree_products.supplier_id")
```

But that's only good if we don't re-use the same relation somewhere else. Otherwise we will trigger an additional query here.

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -154,49 +151,30 @@ def editable_variants_for_incoming_exchanges_from(producer)
       end
     end
 
+    def all_incoming_editable_variants
+      valid_suppliers = visible_enterprises.map do |enterprise|
+        enterprise.id if user_manages_coordinator_or(enterprise)
```
mkllnk: The `if` will cause `nil` values in the new array. You can call `compact` on the result or go a different route:

```rb
visible_enterprises.select { |enterprise| user_manages_coordinator_or(enterprise) }.map(&:id)
```

Potentially, we could do it all in SQL but that's probably beyond this PR.

Matt-Yorkley:
Rebased for ruby upgrade

luisramos0:
rebased to resolve conflicts. ready for testing when the build is green.

daniellemoorhead:
@Matt-Yorkley semaphore build has failed, needs to be rebased?

Also, is it worth testing this not just in terms of "it still works as expected" but also "it's faster / more performant"? @lin-d-hop what do you think?

luisramos0:
I restarted the build. Our build is very flaky lately.
an alternative is to leave the perfomance validation to production when we have all these latest changes together.

RachL:
@daniellemoorhead during performance chat we agreed that performance was going to be tested in production. So the scope of testing remains "it still works as expected" and not "it's faster / more performant".

RachL:
Create, updating, duplicating deleting.

All works fine for super admin, admin and the shopfront info was correct as well.

Tested simple OC and complex OC.

This is ready to go !

PR #3700, opened 2019-04-05
Remove dependency to spree_auth_devise

luisramos0:
#### What? Why?

Closes #4014

We are removing the spree authentication module and moving what we need to OFN codebase.
There are no additions and very few modifications, this is a careful copy paste exercise.

I thought of making multiple PRs for this but then realized this would not be too big and it's much better to review and test this all in one go.

I fixed a lot of rubocop issues in this PR. I leave a few unresolved.

#### What should we test?
Quite a few things to be carefully tested in this PR.
- [x] - admin/users: list, search, create, edit (including only reseting password), delete, generate_api_key, clear_api_key and regenerate_api_key (in this PR we fix a bug: if you try to delete an user with orders, instead of getting the snail, you get an error message that can be translated)
- [x] - checkout: guest checkout and login before checkout, checkout with already_registered email should not be possible
- [x] - user account: create account, login, forgot password, show list of orders and change password in the frontoffice account
- [x] - user sessions: logout frontoffice and backoffice
- [x] - user_mailer: reset password instructions email, signup confirmation email, account confirmation instructions (verify email sender address)
- [x] - backoffice top nav bar: link to shopfront and account
- [x] - edit customer details in an order
- [x] - detect missing translations in all these tests

#### Release notes
Changelog Category: Changed
Removed dependency to spree_auth_devise, the spree adapted version of devise, an authentication framework. OFN now works directly with devise.

Issue #3202, opened 2018-12-11
Seeding process: Stop asking for input!


jhsu802701:
## Description
During the process of seeding the database, I'm asked for an email address and password for the admin user.

## Expected Behavior
When I seed the database in an app, I expect the script to proceed without my input.

## Actual Behavior
Although there is a default email address and password, the script can't go on until I confirm them.  I HATE babysitting processes.  I just want to finish these basic tasks and move on.

## Severity
bug-s4: it's annoying, but you can use it

## Possible Fix
Just change the script to use the default admin email address and password WITHOUT demanding input.

sauloperez:
I know it can be improved but have you tried using the [yes](https://en.m.wikipedia.org/wiki/Yes_(Unix)) command?

mkllnk:
This behaviour came from spree_auth_devise and the solution was to set an environment variable:

```sh
export AUTO_ACCEPT="true"
#export ADMIN_EMAIL="spree@example.com"
#export ADMIN_PASSWORD="spree123"
```

Anyway, we removed spree_auth_devise in https://github.com/openfoodfoundation/openfoodnetwork/pull/3700. We are not creating a user anymore and this is not happening. I will close this issue.

PR #4142, opened 2019-08-09
Seed user

mkllnk:
#### What? Why?

Since we merged #3700 the first default admin user hasn't been created when seeding the database. We were missing a login and couldn't create the sample data.

It was very easy now to also solve https://github.com/openfoodfoundation/openfoodnetwork/issues/3202 and get rid of the confirmation email (which is annoying for developers).

#### What should we test?

No testing required.

#### Release notes

Changed: Developers don't need to confirm admin user details when setting up the database.
Changed: The first automatically created admin user is not notified by email any more.

Changelog Category: Changed

### DIFF
--- a/db/default/users.rb
+++ b/db/default/users.rb
@@ -56,6 +56,7 @@ def create_admin_user
     say "\nWARNING: There is already a user with the email: #{email}, so no account changes were made.  If you wish to create an additional admin user, please run rake spree_auth:admin:create again with a different email.\n\n"
   else
     admin = Spree::User.new(attributes)
+    admin.skip_confirmation!
     if admin.save
       role = Spree::Role.find_or_create_by_name 'admin'
       admin.spree_roles << role
--- a/db/default/users.rb
+++ b/db/default/users.rb
@@ -34,7 +34,7 @@ def prompt_for_admin_email
 end
 
 def create_admin_user
-  if ENV['AUTO_ACCEPT']
+  if ENV.fetch("AUTO_ACCEPT", true)
     password = 'spree123'
     email = 'spree@example.com'
   else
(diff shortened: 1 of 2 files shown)


db/default/users.rb
```diff
@@ -34,7 +34,7 @@ def prompt_for_admin_email
 end
 
 def create_admin_user
-  if ENV['AUTO_ACCEPT']
+  if ENV.fetch("AUTO_ACCEPT", true)
```
Matt-Yorkley: Unless we explicitly set this to `false` in ofn-install, any username and password currently supplied via ENV vars will be ignored?

db/default/users.rb
```diff
@@ -56,6 +56,7 @@ def create_admin_user
     say "\nWARNING: There is already a user with the email: #{email}, so no account changes were made.  If you wish to create an additional admin user, please run rake spree_auth:admin:create again with a different email.\n\n"
   else
     admin = Spree::User.new(attributes)
+    admin.skip_confirmation!
```
Matt-Yorkley: :tada:

db/default/users.rb
mkllnk: Good catch @Matt-Yorkley! That could have been bad, production server with public admin account. What do you think about my new commit?

I would like to rewrite the whole thing but it's not a priority right now with S1 bugs around. So I tried to fix it as quickly as possible.

Matt-Yorkley: > What do you think about my new commit?

Perfect :ok_hand: 

> ...it's not a priority right now with S1 bugs around. So I tried to fix it as quickly as possible.

Agreed :+1:

Matt-Yorkley:
I just tried deploying and Ansible now hangs at the `seed user` task... :/

PR #4147, opened 2019-08-12
Remove order cycle distributed products N+1

sauloperez:
#### What? Why?

Related to https://github.com/openfoodfoundation/openfoodnetwork/issues/3853

Totally rewrites the logic to fetch products to be displayed in the shopfront; those that don't have an obsolete master. It does so by rewriting the slow and imperative Ruby logic with the following single pretty big SQL query.

```sql
SELECT DISTINCT(spree_products.*)
 FROM "spree_products"
INNER 
 JOIN "spree_variants" "variants_including_masters_spree_products"
   ON "variants_including_masters_spree_products"."product_id" = "spree_products"."id"
INNER 
 JOIN "exchange_variants" "exchange_variants_spree_variants_join"
   ON "exchange_variants_spree_variants_join"."variant_id"     = "variants_including_masters_spree_products"."id"
INNER 
 JOIN "exchanges"
   ON "exchanges"."id"                                         = "exchange_variants_spree_variants_join"."exchange_id"
 LEFT OUTER
 JOIN (
   SELECT *
     FROM inventory_items
WHERE enterprise_id                                            = 6701
      ) AS o_inventory_items
   ON o_inventory_items.variant_id                             = variants_including_masters_spree_products.id
 LEFT 
 JOIN (
   SELECT "spree_products".*
     FROM "spree_products"
    INNER 
     JOIN exchanges AS exchanges_suppliers
       ON exchanges_suppliers.sender_id                            = spree_products.supplier_id
     LEFT 
     JOIN "spree_variants"
       ON "spree_variants"."product_id"                            = "spree_products"."id"
      AND "spree_variants"."deleted_at" IS NULL
     LEFT 
     JOIN "exchange_variants"
       ON "exchange_variants"."variant_id"                         = "spree_variants"."id"
     LEFT 
     JOIN "exchanges" AS obsolete_exchanges
       ON "obsolete_exchanges"."id"                                = "exchange_variants"."exchange_id"
    WHERE ("spree_products".deleted_at IS NULL)
      AND (exchanges_suppliers.order_cycle_id                      = 2017)
      AND (
            obsolete_exchanges.incoming                              = false
       OR obsolete_exchanges.incoming IS NULL
          )
    GROUP BY "spree_products"."id"
   HAVING COUNT(*)                                                 > 1
      AND bool_or(
            is_master                                                = true
      AND exchange_variants.id IS NOT NULL
          )
  AND COUNT(exchange_variants.id)                              = 1
      ) AS products_with_obsolete_master
   ON variants_including_masters_spree_products.product_id     = products_with_obsolete_master.id
WHERE "exchanges"."order_cycle_id"                             = 2017
  AND "exchanges"."incoming"                                   = 'f'
  AND "exchanges"."receiver_id"                                = 6701
  AND (
        o_inventory_items.id IS NULL
   OR o_inventory_items.visible                                = true
      )
  AND (products_with_obsolete_master.id IS NULL)
  AND (spree_products.deleted_at IS NULL)
  AND (variants_including_masters_spree_products.deleted_at IS NULL)
```

:page_with_curl: http://tatiyants.com/pev/#/plans/plan_1565624404628 is its execution plan for Katuma production.

We still follow the previous code's idea of removing the invalid products (products with an obsolete master) from the result, but this time through a LEFT JOIN between variants and `products_with_obsolete_master`.

although tests are passing and the shopfront seems to render properly, the readability of the changed class needs to be improved **so this is still WIP**. Any hints and comments that could help on that are very welcome.

I literally suffered to get this query due to the complexity of the data model and the quality of our codebase, please be kind.

:information_source: **Please, leave any questions so I can help everyone understand**. I'll try to document the result of all these `JOIN`s somehow.

#### What should we test?

TBD

#### Release notes

Fetching the valid order cycle products for a shop does not perform N+1
Changelog Category: Changed

### DIFF
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -34,6 +34,7 @@ class OrderCycleDistributedProducts
       .group(:product_id)
   end
 
+  # TODO: filter by products supplied by the OC suppliers so we don't go through the whole products table.
   def products_with_obsolete_master
     query = <<-SQL
 SELECT "spree_products".*
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -43,14 +43,13 @@ class OrderCycleDistributedProducts
     Spree::Product
       .joins('LEFT JOIN "spree_variants" ON "spree_variants"."product_id" = "spree_products"."id" AND "spree_variants"."deleted_at" IS NULL')
       .joins('LEFT JOIN "exchange_variants" ON "exchange_variants"."variant_id" = "spree_variants"."id"')
-      .joins('LEFT JOIN "exchanges" ON "exchanges"."id" = "exchange_variants"."exchange_id"')
       .merge(distributor.inventory_variants)
       .group('"spree_products"."id"')
       .having(<<-SQL)
         COUNT(*) > 1
         AND bool_or(is_master = true
-        AND exchanges.id IS NOT NULL)
-        AND COUNT(exchanges.id) = 1
+        AND exchange_variants.id IS NOT NULL)
+        AND COUNT(exchange_variants.id) = 1
       SQL
   end
 end
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -15,8 +15,8 @@ class OrderCycleDistributedProducts
   #
   # @return [ActiveRecord::Relation<Spree::Product>]
   def relation
-    valid_product_ids = Spree::Variant
-      .joins(:exchanges)
+    valid_product_ids = order_cycle
+      .variants_distributed_by(distributor)
       .joins(<<-SQL.strip_heredoc)
         LEFT JOIN (
           #{products_with_obsolete_master.select('spree_products.id').to_sql}
@@ -24,10 +24,6 @@ class OrderCycleDistributedProducts
         ON spree_variants.product_id = products_with_obsolete_master.id
       SQL
       .where('products_with_obsolete_master.id IS NULL')
-      .merge(distributor.inventory_variants)
-      .merge(Exchange.in_order_cycle(order_cycle))
-      .merge(Exchange.outgoing)
-      .merge(Exchange.to_enterprise(distributor))
       .select(:product_id)
       .group(:product_id)
 
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -15,8 +15,11 @@ class OrderCycleDistributedProducts
   #
   # @return [ActiveRecord::Relation<Spree::Product>]
   def relation
+    products = Spree::Product
+      .joins(:variants_including_master)
+      .merge(order_cycle.variants_distributed_by(distributor))
+
     variants = order_cycle.variants_distributed_by(distributor)
-    products = variants.map(&:product).uniq
 
     valid_products = products.reject do |product|
       product_has_only_obsolete_master_in_distribution?(product, variants)
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -55,6 +55,6 @@ class OrderCycleDistributedProducts
   def distributed_current_variants(product)
     order_cycle
       .variants_distributed_by(distributor)
-      .merge(product.variants)
+      .where(id: product.variants)
   end
 end
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -15,22 +15,29 @@ class OrderCycleDistributedProducts
   #
   # @return [ActiveRecord::Relation<Spree::Product>]
   def relation
-    variants = order_cycle.variants_distributed_by(distributor)
-    products = variants.map(&:product).uniq
+    all_distributed_product_ids = all_distributed_products
+      .select(:product_id)
+      .map(&:product_id)
 
-    valid_products = products - products_with_obsolete_master
-    product_ids = valid_products.map(&:id)
+    product_ids_with_obsolete_master = products_with_obsolete_master
+        .map(&:id)
 
-    Spree::Product.where(id: product_ids)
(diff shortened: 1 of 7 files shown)


app/services/order_cycle_distributed_products.rb
```diff
@@ -15,27 +15,63 @@ def initialize(order_cycle, distributor)
...
+  def products_with_obsolete_master
+    Spree::Product
+      .joins('INNER JOIN exchanges AS exchanges_suppliers ON exchanges_suppliers.sender_id = spree_products.supplier_id')
+      .joins('LEFT JOIN "spree_variants" ON "spree_variants"."product_id" = "spree_products"."id" AND "spree_variants"."deleted_at" IS NULL')
+      .joins('LEFT JOIN "exchange_variants" ON "exchange_variants"."variant_id" = "spree_variants"."id"')
+      .joins('LEFT JOIN "exchanges" AS obsolete_exchanges ON "obsolete_exchanges"."id" = "exchange_variants"."exchange_id"')
+      .where('exchanges_suppliers.order_cycle_id = ?', order_cycle.id)
+      .where('obsolete_exchanges.incoming = false OR obsolete_exchanges.incoming IS NULL')
+      .group('"spree_products"."id"')
+      .having(<<-SQL.strip_heredoc)
```
sauloperez: this HAVING implements the constraints to actually get products with an obsolete master.

app/services/order_cycle_distributed_products.rb
```diff
@@ -15,27 +15,63 @@ def initialize(order_cycle, distributor)
...
+        COUNT(*) > 1
+        AND bool_or(is_master = true
+        AND exchange_variants.id IS NOT NULL)
```
sauloperez: :warning: this should be nested as it is part of the `bool_or` expression

sauloperez:
I'll proceed at splitting it into query objects which will hopefully make it easy to understand in chunks with proper documentation.

app/services/order_cycle_distributed_products.rb
```diff
@@ -15,27 +15,45 @@ def initialize(order_cycle, distributor)
...
   end
 
   private
 
   attr_reader :order_cycle, :distributor
 
-  # If a product without variants is added to an order cycle, and then some variants are added
-  # to that product, but not the order cycle, then the master variant should not available for
-  # customers to purchase.
-  def product_has_only_obsolete_master_in_distribution?(product, distributed_variants)
```
luisramos0: my main feedback is: there are no products with obsolete master in our DB, all products always have a "standard variant". we should enforce and trust that. Would that help simplify this query?

sauloperez: the meaning of variant here is `master: false`, I know it's a bit cumbersome because we use the terms interchangeably.

sauloperez:
My main concerns here are that I might be missing details of the intricacies of the `exchanges` or assuming some that might not be accurate potentially introducing regressions, like the scoping that I did in `SuppliedInOrderCycleQuery`.

Also, there is this bit about `inventory_items` that I might have broken as it's not taking into account the value of `prefers_product_selection_from_inventory_only?`

app/services/order_cycle_distributed_products.rb
luisramos0: I am not sure we are talking about the same thing :-)

can you clarify this statement?
"the meaning of variant here is master: false"

My comment was about "If a product without variants is added to an order cycle", there's no such thing in OFN now, all products have at least one "standard variant" (master: false)

luisramos0:
ok, before I review the query in details we need to agree on an approach about product_has_only_obsolete_master_in_distribution, I think it should be dead code: deleted.

app/services/order_cycle_distributed_products.rb
sauloperez: Ah! that's an interesting plot twist. Could you point to the place in the codebase where we enforce that? just to see if we really can get rid of this PR and just remove this class. That would like a dream :tada:.

luisramos0: first, tested in au, fr, uk and katuma, **products with no standard variant**, empty everywhere:
`select * from spree_products where not exists (select 1 from spree_variants where is_master = false and product_id = spree_products.id);
`

the code is in the product_decorator, ensure_standard_variant [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/models/spree/product_decorator.rb#L39)

there's even a migration doing it back in 2014 [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/db/migrate/20141003060622_add_standard_variant_to_products.rb)


SO, I looked into **exchanges with master variants**:
`select * from exchange_variants ev, spree_variants v where ev.variant_id = v.id and v.is_master = true;`

This query is empty in katuma and FR, the most recent instances.

In AU and UK, there are some old entries. In AU the latest is from 2015-05-22 and in UK the latest is from 2016-07-08.

The Order Cycles in au and uk with these master variants have closed around those dates as well, 2015 and 2016 respectively.

SO, looks like we can ignore that and move on to a world where OCs cannot have master variants :-)

what do you think?

sauloperez: > SO, looks like we can ignore that and move on to a world where OCs cannot have master variants :-)

That's awesome! but we'll to create a separate data migration to deal with those old OCs in UK and AU.

luisramos0: this is empty in both AU and UK:
`select v_master.id from spree_variants v_master where v_master.id in (select distinct v.id from exchange_variants ev, spree_variants v where ev.variant_id = v.id and v.is_master = true) and not exists (select id from spree_variants v where v.product_id = v_master.product_id and v.is_master = false);`

that means that all master variants in exchanges in AU and UK have **at least** one standard variant available. We can switch the master variant with one of the standard variants of the product. Because we are not changing order, only closed OCs I think this is fine.

mkllnk: Sounds good. We need a migration to remove those master variants from old order cycles and then we can remove this code.

sauloperez:
I created https://github.com/openfoodfoundation/openfoodnetwork/issues/4251 to properly scope the work we agreed on

PR #4153, opened 2019-08-13
Fix deployments

Matt-Yorkley:
#### What? Why?

Fixes an issue with deployments and the `ofn-install` CI build failing, since we merged #4142 this morning.

#### What should we test?

Deployments should work now (dev tested).

### DIFF
--- a/db/default/users.rb
+++ b/db/default/users.rb
@@ -56,7 +56,7 @@ def create_admin_user
     say "\nWARNING: There is already a user with the email: #{email}, so no account changes were made.  If you wish to create an additional admin user, please run rake spree_auth:admin:create again with a different email.\n\n"
   else
     admin = Spree::User.new(attributes)
-    admin.skip_confirmation!
+    admin.skip_confirmation!.skip_confirmation_notification!
     if admin.save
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Deployments are now fixed :heavy_check_mark:

db/default/users.rb
```diff
@@ -56,7 +56,7 @@ def create_admin_user
     say "\nWARNING: There is already a user with the email: #{email}, so no account changes were made.  If you wish to create an additional admin user, please run rake spree_auth:admin:create again with a different email.\n\n"
   else
     admin = Spree::User.new(attributes)
-    admin.skip_confirmation!
+    admin.skip_confirmation!.skip_confirmation_notification!
```
mkllnk: Interesting, I probably didn't see this one because I didn't have delayed job running locally.

Matt-Yorkley: This one only occurs on a fresh deploy :point_up:

Matt-Yorkley:
Ugh. There's a still an issue with this in the ofn-install CI build. :frowning_face:

PR #4154, opened 2019-08-14
Adjust user creation syntax

Matt-Yorkley:
#### What? Why?

Related to https://github.com/openfoodfoundation/openfoodnetwork/pull/4153 which was supposed to fix 2 bugs, but only fixed one of them.

`#skip_confirmation!` can't be chained... :see_no_evil: 

#### What should we test?

Dev test only. It'll need to be checked when deploying during a fresh install.

### DIFF
--- a/db/default/users.rb
+++ b/db/default/users.rb
@@ -35,8 +35,8 @@ end
 
 def create_admin_user
   if ENV.fetch("AUTO_ACCEPT", true)
-    password = ENV.fetch("ADMIN_PASSWORD", "spree123")
-    email = ENV.fetch("ADMIN_EMAIL", "spree@example.com")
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Ok, I just tested and readjusted this to fix another error caused by validation callbacks calling `downcase!` on the email ENV var, where it hadn't been used with `#dup`, so failed with a "can't modify frozen string" error. It looks like it's all working now... :roll_eyes:

sauloperez:
aha! that's exactly the error the Travis CI for ofn-install was reporting

sigmundpetersen:
This is ready right? 

Is this something we need to keep a note of somewhere:
> It'll need to be checked when deploying during a fresh install.

Matt-Yorkley:
I tested it in Vagrant this morning. Ready to go :+1:

PR #4157, opened 2019-08-15
4138 Avoid infinite loop when updating products

### LINKED ISSUE #4138: Users cannot save changes to product list or inventory
## Description

SELECTED Users on OFN-CAN cannot save changes (stock on hand, prices..) to existing products in either product list or inventory.  But it seems (from limited testing) that super admin can do so.  Yesterday I posted this on slack - but from my first tests,  I thought it was just affecting my own enterprises -  but this morning at least one other user reported this.  I have canvased users to see how widespread it is.   If it helps with narrowing this down,  the enterprises I know are affected:    Garden Party Flower Farm,   La Primavera Flower Farm  ---  these are both associated with 2 hubs:   The Local Flower Collective  and Local Flower Hub Preview shop.     

## Expected Behavior

Users should be able to manage their product list!!!
## Actual Behaviour

When users try to save a change to their product list (like updating stock) they get a message: Saving failed. Save failed due to invalid data:500.  When users try to save data in their inventory they get a different error:  I couldn't get authorisation to save those changes, so they remain unsaved.    As super-admin,  I am able to save changes. (I think - still testing this)

## Steps to Reproduce

To reproduce with sample data created by the Rake task `ofn:sample_data`:

1. Update `sells` to `any` on Freddy's Farm Shop to make it a producer distributor.
2. Update Freddy's Farm Shop OC to sell Fredo's products.
3. Update Fredo's Farm Hub OC to sell Freddy's products.
4. Attempting to edit Fredo's Farm products in the Bulk Edit Products results in "Saving failed.Save failed due to invalid data:500"

## Animated Gif/Screenshot

still testing -  will post screenshots and further info  -  but wanted this up there given its s1

## Context

As super admin -  users are sending me all their updates by email this morning so I can make their changes.  Not very sustainable

## Your Environment

* Version used: 2.2
* Operating System and version (desktop or mobile): windows 10
* OFN Platform instance where you discovered the bug, and which version of the software they are using.   OFN-CAN

mkllnk:
#### What? Why?

Closes #4138.

Rails supports caching based on the `updated_at` timestamp of records. The timestamp is updated on each save or manually via the `touch` method. When one record is updated, it can also touch other records to invalidate caches based on those records.

For example, when a variant is changed, we touch the producer enterprise. And when a producer is changed we touch all distributors of that producer's variants. Unfortunately, that can result in an infinite loop if an enterprise is a producer and a distributor and it sells products of another enterprise it also supplies to. Or simply: when two enterprises are selling each other's products.

When touching all distributors of a producer, we don't need to update all of the distributors' distributors as well. At least I can't think of a use case for that and it's not covered by specs.

So I changed the logic to stop the cascade of touching enterprises and only touch those enterprises which are direct distributors of a changed producer without further callbacks. It's also faster than before.

As far as I know, the timestamp is only used in the `Api::EnterpriseSerializer`. That serializer is used on these pages:

- `app/views/producers/index.html.haml`
- `app/views/shops/index.html.haml`
- `app/views/groups/show.html.haml`
- `app/views/checkout/edit.html.haml`
- `app/views/spree/orders/edit.html.haml`
- `app/views/spree/orders/show.html.haml`

I'm not sure though if we use all this information on all of these pages and it may be possible to clean up the serializer and reduce the number of events that cause cache invalidation which would simplify the code.

#### What should we test?

We need to make sure that the following pages are updated correctly:

- /producers
- /shops
- /groups -> select a group, check tabs
- /checkout/edit
- /orders/Order-Number and edit

These pages should update when changing the following information:

- A product changes its category (taxons, filters change).
- A product's properties change (I don't know what that is either).
- An enterprise's producer properties change.
- The owner of a product changes.

#### Release notes

Fixed a bug that could lead to certain enterprises not being able to update their products.

Changelog Category: Fixed

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -464,9 +464,11 @@ class Enterprise < ActiveRecord::Base
     self.permalink = Enterprise.find_available_permalink(name)
   end
 
+  # Touch distributors without them touching their distributors.
+  # We avoid an infinite loop and don't need to touch the whole distributor tree.
   def touch_distributors
     Enterprise.distributing_products(supplied_products.select(:id)).
       where('enterprises.id != ?', id).
-      find_each(&:touch)
+      update_all(updated_at: Time.zone.now)
   end
 end
--- a/spec/models/spree/variant_spec.rb
+++ b/spec/models/spree/variant_spec.rb
@@ -533,7 +533,7 @@ module Spree
         )
       end
 
-      pending "saves without infinite loop" do
+      it "saves without infinite loop" do
         expect(variant1.update_attributes(cost_price: 1)).to be_truthy
       end
     end
--- a/spec/models/spree/variant_spec.rb
+++ b/spec/models/spree/variant_spec.rb
@@ -515,6 +515,28 @@ module Spree
 
       it_behaves_like "a model using the LocalizedNumber module", [:price, :cost_price, :weight]
     end
+
+    context "in a circular order cycle setup" do
+      let(:enterprise1) { create(:distributor_enterprise, is_primary_producer: true) }
+      let(:enterprise2) { create(:distributor_enterprise, is_primary_producer: true) }
+      let(:variant1) { create(:variant) }
+      let(:variant2) { create(:variant) }
+      let!(:order_cycle) do
+        enterprise1.supplied_products << variant1.product
+        enterprise2.supplied_products << variant2.product
+        create(
+          :simple_order_cycle,
+          coordinator: enterprise1,
+          suppliers: [enterprise1, enterprise2],
+          distributors: [enterprise1, enterprise2],
+          variants: [variant1, variant2]
+        )
+      end
+
+      pending "saves without infinite loop" do
+        expect(variant1.update_attributes(cost_price: 1)).to be_truthy
+      end
+    end
   end
 
   describe "destruction" do


daniellemoorhead:
Hey @lin-d-hop and @mkllnk if it's ok with you both I'd like to save this one for Jen and I to test on Friday. It seems like a good one for a newbie. 

If it can't wait that long then let me know. Ta!

lin-d-hop:
This is a pretty big one to test @daniellemoorhead! Ping me any questions :-)

daniellemoorhead:
I've started testing this but it's huge so will take me a while.

@lin-d-hop I'm going through as super admin first (Maikel recommended), and then have to set myself as manager and check it all again. Testing notes so far are here: https://docs.google.com/document/d/1N-l_FdBewMrzsbd2A-l6IdIMyyALGcizPSB0pdf8VO4/edit?usp=sharing

Will sit down on sunday evening and do some more, hopefully get it done.

RachL:
I'm having an issue with properties. On staging ES I've used the enterprise "Nanny Ogg's Farm" which is part of the group "Unseen University".

This enterprise didn't have any property set up except the inherited "hello" property on products.

I've uncheked the property hello and added the property "certified organics" on the product:
image

and did the same on the enterprise :

image

but this new property is not shown in /shops neither in /groups. And "hello" keeps showing up (also I removed it from really all the products).
I will do a second test once we restage the PR for release testing this afternoon.

RachL:
Alright tested this on staging FR now that the PR is staged there as well and couldn't reproduce the problem. So I'm re-staging this in staging ES right now to re-test the original issue with OFN sample data.

RachL:
Updating Fredo's Farm products in the Bulk Edit Products works well after setting the loop up, this is ready to go!!

PR #4159, opened 2019-08-15
Add bugsnag notifications to both product_set and products_controller when stock update fails

luisramos0:
#### What? Why?

To try to debug  #4138 here we add bugsnag notifications to both product set (bulk product update) and products controller (product create and edit) when failing to update stock. This will give us more info about what's going on and in what cases this code is failing to create stock items. Adds more info to these bugsnag events: 21 events in [uk](https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5d4062fde75241001ce7c207) so far and  a few more in [ca](https://app.bugsnag.com/open-food-network-canada-1/open-food-network-canada/errors/5d4b6cfdd5b544001a5a9e26?filters%5Bevent.since%5D%5B0%5D=30d&filters%5Berror.status%5D%5B0%5D=open)

#### What should we test?
We should make sure we can create a product as usually, edit a product and use the bulk product update page normally: change product and variant attributes and create new variants.

#### Release notes
Changelog Category: Changed
Added notifications in case of error to the bulk product update page and the create/edit product page.

### DIFF
--- a/app/models/spree/product_set.rb
+++ b/app/models/spree/product_set.rb
@@ -101,6 +101,7 @@ class Spree::ProductSet < ModelSet
       variant.on_hand = on_hand.to_i if on_hand.present?
     rescue StandardError => error
       notify_bugsnag(error, product, variant, variant_attributes)
+      raise error
     end
   end
 
--- a/app/models/spree/product_set.rb
+++ b/app/models/spree/product_set.rb
@@ -17,9 +17,7 @@ class Spree::ProductSet < ModelSet
   #   variant.update_attributes( { price: xx.x } )
   #
   def update_attributes(attributes)
-    if attributes[:taxon_ids].present?
-      attributes[:taxon_ids] = attributes[:taxon_ids].split(',')
-    end
+    split_taxon_ids!(attributes)
 
     found_model = @collection.find do |model|
       model.id.to_s == attributes[:id].to_s && model.persisted?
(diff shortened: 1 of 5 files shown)


luisramos0:
tagging S1 just for awareness that this is connected to (but does not close) an s1 bug 4138

app/models/spree/product_set.rb
```diff
@@ -89,10 +89,19 @@ def create_variant(product, variant_attributes)
...
+      variant.on_hand = on_hand.to_i if on_hand.present?
+    rescue StandardError => error
+      Bugsnag.notify(error) do |report|
+        report.add_tab(:product, product.attributes)
+        report.add_tab(:product_error, product.errors.first) unless product.valid?
+        report.add_tab(:variant_attributes, variant_attributes)
+        report.add_tab(:variant, variant.attributes)
+        report.add_tab(:variant_error, variant.errors.first) unless variant.valid?
+      end
+    end
```
kristinalim: @luisramos0 Do we really want this? This would make saving appear successful, and the user might not realize that their data was not saved.

I think what we really need is feedback in the front-end pointing the user to the problem with their data, but of course that would take more time.

luisramos0: yes, agree. I dont think I can go for that right now: add user feedback in the frontend.

Do you think it's better to raise the original error and let the user know there's something wrong? I can do that.

luisramos0: ok, [new commit](https://github.com/openfoodfoundation/openfoodnetwork/pull/4159/commits/249a3c4e182eed65e7158251a34c03a1ced5fcfe) to make the exception fly out to the user, we keep the bugsnag info.

luisramos0: I have updated PR description as well.

daniellemoorhead:
@kristinalim @mkllnk can you both make it a priority to review this S1 and get it through to be tested? Thanks!

lin-d-hop:
Looks good to me 

https://docs.google.com/document/d/1Vgc6wVFRck-e7kOJocSUjxOWoBEXykQm-MW27XorEiw/edit#

PR #4162, opened 2019-08-16
Checkout is not refreshing products cache

### LINKED ISSUE #4020: Out-of-stock items shown in shop front
## Description

When an item goes out of stock, it still shows in the shop front. This is not the case when you update the stock as admin. It only happens if the item is bought through the shop front.

## Expected Behavior

If an item is NOT set to on demand, and is out of stock (which includes both 0 and negative numbers) it should NOT show in shop.

## Actual Behaviour

It appears in the shop and doesn't display out of stock message until customer tries to add to cart. Then shows as negative item in shop if the item has negative stock (0 otherwise).

## Steps to Reproduce

1. Open an order cycle with an item with stock `1`.
2. Add that item to your cart and check out.
3. Click on continue shopping.

## Animated Gif/Screenshot

 In this example the stock went down to `-1`:

This item is showing in the shop

When I try to add it to my cart I see this error message

and then the stock is shown on the shop page as if it is in my order

I can place the order and it is correct i.e. doesn't put -1 cheese in my order or reduce what I pay

## Context

This was discovered when investigating #3925 but it's not limited to that bug. It's also mentioned in #4018.

## Severity

s3: a feature is broken but there is a workaround

The workaround is to adjust the stock as admin or refresh the cache.

## Your Environment

* Version used: v2

## Possible Fix

I observed that the shop is still retrieving the old/cached products.json file. We have to either refresh the cache or retrieve the stock levels in a different way. This was probably introduced by v2.

luisramos0:
#### What? Why?

- Closes #4020 

#### What should we test?

As admin:

1. Create a new variant with limited stock (e.g. 5) supplied by Supplier A.
2. Create Shop A.
3. Set up an open OC for Shop A that includes this variant. Do not create a variant override for the variant.
4. Create Shop B.
5. Set up an open OC for Shop B that includes this variant. Do not create a variant override for the variant.
6. Create Shop C.
7. Set up an open OC for Shop C that includes this variant.
8. Create a variant override with different limited stock level (e.g. 8).

As customer:

1. Purchase all the stock of the variant (e.g. 5 from above) from Shop A, and check out.
2. Wait around 1 minute. Check that the variant is no longer visible in Shop A and Shop B. Check that the variant is still in Shop C.
3. Purchase all the stock of the variant (e.g. 8 from above) from Shop C, and check out.
4. Wait around 1 minute. Check that the variant is no longer in Shop C.

#### Release notes

- Refresh the variant cache when the supplier stock level for a non-override limited stock variant is updated without saving the variant. This fixes the issue of non-override limited stock variants still appearing in the shopfront after a customer purchases the last available stock.

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/stock_movement_decorator.rb
+++ b/app/models/spree/stock_movement_decorator.rb
@@ -1,0 +1,10 @@
+Spree::StockMovement.class_eval do
+  after_save :refresh_products_cache
+
+  private
+
+  def refresh_products_cache
+    return if stock_item.variant.blank?
+    OpenFoodNetwork::ProductsCache.variant_changed stock_item.variant
+  end
+end
--- a/spec/features/consumer/shopping/checkout_spec.rb
+++ b/spec/features/consumer/shopping/checkout_spec.rb
@@ -180,8 +180,19 @@ feature "As a consumer I want to check out my cart", js: true do
       end
     end
 
-    context "in the shopfront" do
+    context "in the shopfront with cache enabled" do
+      around do |example|
+        original_config = Spree::Config[:enable_products_cache?]
+        example.run
+        Spree::Config[:enable_products_cache?] = original_config
+      end
+
+      let(:control_product) { create(:taxed_product, supplier: supplier, price: 110, zone: zone, tax_rate_amount: 0.1) }
+      let(:control_variant) { control_product.variants.first }
+      let!(:order_cycle) { create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], coordinator: create(:distributor_enterprise), variants: [variant, control_variant]) }
+
       it "does not show item after all stock of an item is checked out (tesging cache update on checkout)" do
+        Spree::Config[:enable_products_cache?] = true
         variant.update_attributes on_hand: 5
         visit shop_path
 
@@ -200,7 +211,10 @@ feature "As a consumer I want to check out my cart", js: true do
         expect(page).to have_content "Your order has been processed successfully"
 
         visit shop_path
-        page.should_not have_content variant.product.name
+        # The presence of the control product ensures the list of products is fully loaded
+        #   before we verify the sold product is not present
+        page.should have_content control_product.name
+        page.should_not have_content product.name
       end
     end
 
--- a/spec/features/consumer/shopping/checkout_spec.rb
+++ b/spec/features/consumer/shopping/checkout_spec.rb
@@ -64,31 +64,6 @@ feature "As a consumer I want to check out my cart", js: true do
     context 'login in as user' do
       let(:user) { create(:user) }
 
-      def fill_out_form
-        choose sm1.name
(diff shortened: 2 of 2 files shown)


app/models/spree/stock_movement_decorator.rb
```diff
@@ -0,0 +1,10 @@
+Spree::StockMovement.class_eval do
+  after_save :refresh_products_cache
```
kristinalim: This is the sequence of method calls :face_with_head_bandage: up to `StockMovement` when a line item is destroyed for the problem scenario:

* `LineItem#update_inventory_before_destroy`
* `LineItem#update_inventory`, as alias calling
* `LineItem#update_inventory_with_scoping`
* `LineItem#update_inventory_without_scoping`, which calls
* Spree core `LineItem#update_inventory`
* `Spree::OrderInventory#verify`
* `Spree::OrderInventory#remove`
* `Spree::OrderInventory#remove_from_shipment`
* `Spree::InventoryUnit#destroy`
* `Spree::StockLocation#move` / `Spree::StockLocation#restock`
* Scoped `Spree::Variant#move`, which is `OpenFoodNetwork::ScopeVariantToHub::ScopeVariantToHub#move`
* `OpenFoodNetwork::ScopeVariantToHub::ScopeVariantToHub#move_stock!`
* `Spree::Variant#move` in `VariantStock`
* `Spree::StockMovement#create`

kristinalim: For overridden limited stock variants, the cache refresh happens in `VariantOverride`.

luisramos0: Thanks a lot for working on this Kristina!

That stack trace is for deleting line item of a completed order (Spree::OrderInventory#verify will only move stock on completed orders)

I think the basic problem scenario is the checkout flow which goes:
- Spree::Order#finalize!
- Spree::Shipment#finalize!
- Spree::Shipment#manifest_unstock
- Spree::StockLocation#unstock
- Spree::StockLocation#move where this line meets the line you described above

I think your solution is valid! I think we could alternatively do, if you prefer, refresh_products_cache inside VariantStock#move

luisramos0:
GH doesnt let me approve because I authored the first commit but I approve you commits Kristina, so this can be moved to test ready if you are happy with it.

daniellemoorhead:
>GH doesnt let me approve because I authored the first commit but I approve you commits Kristina, so this can be moved to test ready if you are happy with it.

@kristinalim are you happy to move this to Testing?

kristinalim:
Yes @daniellemoorhead, thanks! Moved this to second in the Test Ready below the other S2.

lin-d-hop:
Tested this by:

1) Following the instructions above - all worked fine. Cache refresh was instantaneous (did not have to wait 'about a minute')

2) Checking issue as described in #4020 - this is now fixed.

Ready to go :-D

PR #4164, opened 2019-08-18
Remove some unused deface overrides and move some others to their views

luisramos0:
#### What? Why?

The last bits of #2807 🎉 (only one override to go after current PRs are merged).

Deleted some unused overrides:
- Remove unused overrides of spree frontend code: spree/shared/nav_bar, footer and header
- Remove unused nav_bar sessions route and override
- Remove unused login_bar override of nav_bar
- Remove unused sidebar override

Move some others to their respective files:
- Delete unused frontend auth token script override and move the backend one to the head partial
- Move shared/user_form override to the view itself
- Remove users tab override and add it directly to the tabs partials where all other tabs are now
- Remove login_nav override (backoffice top nav bar) and move it to admin layout

#### What should we test?
In the backoffice:
- Make sure the users menu entry appears as before for super admins.
- Make sure you see the top right menu with links to account, logout and shopfront.  
- For the auth token script we just need to verify the api calls still work, successfully using bulk product update page will do the trick.

For the user form change, register a new account but dont confirm the email, go to /account#/settings and see the warning "email not confirmed".

To check some deleted files from old spree_frontend we can double check the menu in the frontoffice is working as usually.

#### Release notes
Changelog Category: Changed
Made codebase easier to grow removing some more html views overrides (deface).


daniellemoorhead:
OK, am looking at this one to test. Some questions @luisramos0:

>For the auth token script we just need to verify the api calls still work, successfully using bulk product update page will do the trick.

Does this mean I just have to make any change on this page? "Successfully using" is a little unclear to me, would help us newbie testers if you could be a little bit more specific here until we're super users and understand each of the functions and what success in using them equates to :) 

>To check some deleted files from old spree_frontend we can double check the menu in the frontoffice is working as usually.

Frontoffice = the main site, yes?

And is there anything in particular in menu we should look at, e.g. moves into small screen mode the same, has all the links its supposed to display? And is there a possibility with this defacing work that there may be discrepancies across browsers that we should be specifically looking for?

luisramos0:
hi @daniellemoorhead 
- bulk product page - yes, just loading it and maybe create a variant or something with different user profiles will do the trick

- frontoffice - yes, customers facing part of the platform

- menu - no need to test diff browsers or screen sizes. I think we should just make sure the entry appears for diff user profiles.

thanks!

daniellemoorhead:
Awesome, thanks Luis!

JenRSmith:
Testing completed. Assumption made that "successful" use of the Bulk Edit Products page is to add a variant as a super admin and a shop manager and not seeing this variant in a shop. 

@RachL and @luisramos0 can you confirm this is correct? 

https://docs.google.com/document/d/1zDT_JZZf-aAY2w_AByJ2zNe8t7ImD7Fi_HheZAfG8YY/edit?usp=sharing - testing notes here.

JenRSmith:
Testing cherry broken. Celebrating.

luisramos0:
awesome, welcome @JenRSmith 🎉

PR #4168, opened 2019-08-20
4160 Fix product report filters

### LINKED ISSUE #4160: Lettuceshare report no longer selecting by order cycle
## Description

Our biggest user in Aus has a specific report the 'Lettuceshare Report' that they use for providing info in a form to their network of buyers groups. It has stopped working with V2 . . 

"we create the Order Cycles for the next week and then use Lettuce Share to create a Wholesale Availability File for Customers to see the wholesale availability for the week ahead. Our co-op bulk buyers groups upload the csv file that is created to upload into their Lettuce share ordering systems for members.  This File that we create each week is essentially our price list and availability list for the Week so its really important that it is accurate and up to date. 
**This week the report is not reflecting the the current order cycle at all. It appears to have listed all the products and variants in the  Food Connect OFN system.**

I've attached the file created on Tuesday night this week after following this process: products_and_inventory_20190815 . An example that you will see is Kiwifruit is listed - it is not available and it is not showing as available in our order cycle when you search for Kiwi fruit. So the order cycles are correct but the file created through the lettuce share report completely is wrong.  It appears something has changed since the update that is affecting this report.

## Expected Behavior
The report should only include products / variants in the selected order cycle

## Actual Behavior
It is including all products / variants of the Distributor

## Steps to Reproduce

1. In OFN, select “Reports” from the top menu bar
2. Select “lettuceshare” - this will take about 30 seconds to load up
3. In order to ensure you are getting the correct information, the following filters must be set before downloading the report: [NB. Happy to give access to aus prod server if needed]
    * Distributor: Food Connect Wholesale 
    * Order Cycle: Master relevant to the week (eg. 3419 Master)
4. Check the box “DOWNLOAD AS CSV”, and click ‘Search’ which will start downloading the file to your computer.

## Context
Key user in Aus - this is extremely important to them as this report was built specifically to meet a critical need for them. But I understand that it likely matters to no one else :(

## Severity
s2 - non-critical, no workaround
OR
s3 - because of small number of users impacted (1)

## Possible Fix
Introduced with V2

mkllnk:
#### What? Why?

Closes #4160.

When filtering the LettuceShare report by distributor and order cycle, the order cycle filter was ignored. Too many variants ended up in the report.

#### What should we test?

All three Products & Inventory reports are affected by this change:

1. All products (same as Products & Inventory)
2. Inventory (on hand)
3. LettuceShare

These three reports share the same logic for filtering variants by:

- supplier
- order cycle
- distributor

Each combination of filters should be checked.

- no filters (test with "All products" report)
- filter by supplier and order cycle (test with "All products" report)
- filter by supplier and distributor (test with "Inventory (on hand)" report)
- filter by order cycle and distributor (test with "LettuceShare" report)

You will need at least two order cycles to test the order cycle filter. One of the order cycles should have multiple suppliers and multiple distributors to test those filters as well. And not all distributors should sell all products.

#### Release notes

Fixed: The Products & Inventory reports can be filtered by distributor and order cycle again. The latter got overridden by the former.

Changelog Category: Fixed

### DIFF
--- a/lib/open_food_network/products_and_inventory_report_base.rb
+++ b/lib/open_food_network/products_and_inventory_report_base.rb
@@ -65,7 +65,15 @@ module OpenFoodNetwork
     def filter_to_order_cycle(variants)
       if params[:order_cycle_id].to_i > 0
         order_cycle = OrderCycle.find params[:order_cycle_id]
-        variants.where(id: order_cycle.variants)
+        # There are two quirks here:
+        #
+        # 1. Rails 3 uses only the last `where` clause of a column. So we can't
+        #    use `variants.where(id: order_cycle.variants)` until we upgrade to
+        #    Rails 4.
+        #
+        # 2. `order_cycle.variants` returns an array. So we need to use map
+        #    instead of pluck.
+        variants.where("spree_variants.id in (?)", order_cycle.variants.map(&:id))
       else
         variants
       end
--- a/lib/open_food_network/products_and_inventory_report_base.rb
+++ b/lib/open_food_network/products_and_inventory_report_base.rb
@@ -65,15 +65,11 @@ module OpenFoodNetwork
     def filter_to_order_cycle(variants)
       if params[:order_cycle_id].to_i > 0
         order_cycle = OrderCycle.find params[:order_cycle_id]
-        # There are two quirks here:
-        #
-        # 1. Rails 3 uses only the last `where` clause of a column. So we can't
-        #    use `variants.where(id: order_cycle.variants)` until we upgrade to
-        #    Rails 4.
-        #
-        # 2. `order_cycle.variants` returns an array. So we need to use map
-        #    instead of pluck.
-        variants.where("spree_variants.id in (?)", order_cycle.variants.map(&:id))
+        variant_ids = Exchange.in_order_cycle(order_cycle).
+          joins("INNER JOIN exchange_variants ON exchanges.id = exchange_variants.exchange_id").
+          select("DISTINCT exchange_variants.variant_id")
+
+        variants.where("spree_variants.id IN (#{variant_ids.to_sql})")
       else
         variants
       end
--- a/spec/lib/open_food_network/products_and_inventory_report_spec.rb
+++ b/spec/lib/open_food_network/products_and_inventory_report_spec.rb
@@ -171,10 +171,49 @@ module OpenFoodNetwork
         end
 
         it "should do all the filters at once" do
+          pending "the order cycle filter is overriden by the distributor filter"
+          # The following data ensures that this spec fails if any of the
+          # filters fail. It's testing the filters are not impacting each other.
           distributor = create(:distributor_enterprise)
-          product1 = create(:simple_product, supplier: supplier)
-          product2 = create(:simple_product, supplier: supplier)
-          order_cycle = create(:simple_order_cycle, suppliers: [supplier], distributors: [distributor], variants: [product1.variants.first])
+          other_distributor = create(:distributor_enterprise)
+          other_supplier = create(:supplier_enterprise)
+          not_filtered_variant = create(:simple_product, supplier: supplier).variants.first
+          variant_filtered_by_order_cycle = create(:simple_product, supplier: supplier).variants.first
+          variant_filtered_by_distributor = create(:simple_product, supplier: supplier).variants.first
+          variant_filtered_by_supplier = create(:simple_product, supplier: other_supplier).variants.first
+          variant_filtered_by_stock = create(:simple_product, supplier: supplier, on_hand: 0).variants.first
+
+          # This OC contains all products except the one that should be filtered
+          # by order cycle. We create a separate OC further down to proof that
+          # the product is passing all other filters.
+          order_cycle = create(
+            :simple_order_cycle,
+            suppliers: [supplier, other_supplier],
+            distributors: [distributor, other_distributor],
+            variants: [
+              not_filtered_variant,
+              variant_filtered_by_distributor,
+              variant_filtered_by_supplier,
+              variant_filtered_by_stock
+            ]
+          )
+
+          # Remove the distribution of one product for one distributor but still
+          # sell it through the other distributor.
+          order_cycle.exchanges.outgoing.find_by_receiver_id(distributor.id).
+            exchange_variants.
+            find_by_variant_id(variant_filtered_by_distributor).
+            destroy
+
+          # Make product available to be filtered later. See OC comment above.
+          create(
+            :simple_order_cycle,
+            suppliers: [supplier],
+            distributors: [distributor, other_distributor],
+            variants: [
+              variant_filtered_by_order_cycle
+            ]
+          )
 
           allow(subject).to receive(:params).and_return(
             order_cycle_id: order_cycle.id,
@@ -182,7 +221,8 @@ module OpenFoodNetwork
             distributor_id: distributor.id,
             report_type: 'inventory'
           )
-          subject.filter(variants)
+
+          expect(subject.filter(variants)).to match_array [not_filtered_variant]
         end
       end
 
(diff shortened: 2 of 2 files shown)


lib/open_food_network/products_and_inventory_report_base.rb
```diff
@@ -65,7 +65,15 @@ def filter_to_distributor(variants)
...
-        variants.where(id: order_cycle.variants)
+        # There are two quirks here:
+        #
+        # 1. Rails 3 uses only the last `where` clause of a column. So we can't
+        #    use `variants.where(id: order_cycle.variants)` until we upgrade to
+        #    Rails 4.
+        #
+        # 2. `order_cycle.variants` returns an array. So we need to use map
+        #    instead of pluck.
+        variants.where("spree_variants.id in (?)", order_cycle.variants.map(&:id))
```
kristinalim: Considering an OC with no variants... The SQL here would end up being `spree_variants.id IN (NULL)`, which is not exactly what we want. But it has the same result (no matches when the selected order cycle has no variants) because `spree_variants.id` here would never be null anyway - the chain starts with `Spree::Variant`. :heavy_check_mark:

lin-d-hop:
Hrmmmm I think this is not quite working well yet....

Distributor1 with OrderCycleA contains products 1, 2, 3 from Producer1
Distributor2 with OrderCycleB contains products 1, 2 from Producer1
Producer1 has given enterprise permissions for Distributor1 and Distributor2 to add products to order cycles. 

Distributor1 is not an incoming or outgoing distributor for OrderCycleB.
Distributor2 is not an incoming or outgoing distributor for OrderCycleA.

I set the report filters to:
Distributor1 - Producer1 - OrderCycleB

Am I correct in thinking that this list should be empty because there are no products that match Distributor1 - OrderCycleB because Distributor1 has no direct relationship with OrderCycleB (only enterprise permissions)?

The Distributor and Producer filters work as expected together.
The Producer and Order Cycle filters work as expected together.
Distributor and Order Cycle filters together do not reflect the actual products in the order cycle, but the products for which the Distributor has enterprise permissions to add to Order Cycle.

Looking at the original bug description this is exactly the original issue so perhaps I've missed something. I double check this definitely deployed but it is late now and about the time my brain clocks off.

Back to you @mklink

lin-d-hop:
@mklink did you get a chance to look at this again? Let me know if I made a mistake in testing.

lib/open_food_network/products_and_inventory_report_base.rb
sauloperez: This adds a performance penalty and we should avoid it given our current situation, particularly with reports. I know `order_cycle.variants` returns an array but this doesn't mean we cannot extract that query and make it reusable.

I've been playing with it and the following patch does pass the tests but issue a single query

```diff
         #
         # 2. `order_cycle.variants` returns an array. So we need to use map
         #    instead of pluck.
-        variants.where("spree_variants.id in (?)", order_cycle.variants.map(&:id))
+        query = Spree::Variant.
+          joins(:exchanges).
+          merge(Exchange.in_order_cycle(order_cycle)).
+          select('DISTINCT spree_variants.*')
+
+        variants.merge(query)
       else
         variants
       end
```

We need to be extra careful with those old model methods that issue queries because it turns out most of them are the root cause of many performance issues :warning: .

sauloperez:
@lin-d-hop did you check if this happens in production as well. The change this PRs introduces seems rather equivalent to what we had.

mkllnk:
Hey, this has been on my todo list for a long time. I'm finally looking into it again.

@lin-d-hop I think I found the source of that issue. It's not a regression in this pull request. It is pre-existing behaviour. At the moment, the filters are independent.

> I set the report filters to:
> Distributor1 - Producer1 - OrderCycleB

This selects all variants that are
- sold in OrderCycleB and
- sold by Distributor1 and
- produced by Producer1.

That does not mean that the variant has been sold by Distributor1 in OrderCycleB. It can also be that Distributor1 sold it through another order cycle **and** it's available in OrderCycleB.

The original issue #4160 was that the order cycle filter didn't work at all. It would show products that were not available in the chosen order cycle. In your example, products 1 and 2 are in both order cycles and would still be displayed but product 3 should not be displayed when selecting OrderCycleB. This happened before.

I don't know what the actual requirement is for the exact filter combination. I guess that we don't need to change it as long as the using customers are happy which may be only one in Australia? @kirstenalarsen

lib/open_food_network/products_and_inventory_report_base.rb
mkllnk: Pau, my change didn't add a performance penalty. It was there before. I fixed a bug with the least amount of effort and risk while we are very low in resources.

Your suggestion duplicates code but avoids one query. I agree that it's appropriate to have custom queries in reports. It also avoids using the questionable `OrderCycle#variants` which should be refactored.

I added your change as a new commit, @sauloperez, have a look.

filipefurtad0:
Hi @mkllnk,
We retested the PR on two staging servers (Katuma and UK) with complete fresh data (new user, store, product, etc) and always ran into error 500 everytime we chose a Order Cycle on the Reports window.

https://docs.google.com/document/d/1YRERWVLxCQ-SWBCjJSUCKbVHBXuek0SNJjltoAeQZAU/edit#

RachL:
Oops @filipefurtad0 I forgot to tell you one more hint. When we got the first error 500 in Katuma, there was a Bugsnag notification on Slack. So I usually also add the link to the notification in the github comment in case it is helping somehow.

@mkllnk here is the link with the error: https://openfoodnetwork.slack.com/archives/CEF14NU3V/p1568367192046100

filipefurtad0:
Ahh true. Katuma does, I also forgot to paste it :-) Thank you Rachel :+1:

mkllnk:
Thank you for the detailed report. This broke in a change following Pau's review. The specs should have covered this but were only looking at parts of the logic instead of the whole thing together.

So basically, we changed this:
```diff
- variants.where("spree_variants.id in (?)", variants_in_order_cycle.map(&:id))
+ variants.merge(variants_in_order_cycle)
```

And our `variants_in_order_cycle` query adds a `DISTINCT` keyword to the `SELECT` clause which breaks the order clause.

```sql
-- works:
SELECT "spree_variants".* FROM "spree_variants"
 INNER JOIN "spree_products"
 ON "spree_products"."id" = "spree_variants"."product_id"
 ORDER BY spree_products.name

-- fails:
SELECT DISTINCT "spree_variants".* FROM "spree_variants"
 INNER JOIN "spree_products"
 ON "spree_products"."id" = "spree_variants"."product_id"
 ORDER BY spree_products.name

-- ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

mkllnk:
I can't take the `DISTINCT` out with the `merge` clause because it will end up in duplicating some of the variants. So I have a couple more ideas even though there are less elegant. @sauloperez Let me know what you think about these options:

```rb
# My first version. It has an additional query
variants.where("spree_variants.id in (?)", order_cycle.variants.map(&:id))
```

```rb
# Combination of `where` with a sub-query
variants_in_order_cycle = Spree::Variant.
  joins(:exchanges).
  merge(Exchange.in_order_cycle(order_cycle)).
  select("DISTINCT spree_variants.id")

variants.where("spree_variants.id IN (?)", variants_in_order_cycle)
```

```rb
# Avoiding the unnecessary join with variants but adding more custom sql
variant_ids = Exchange.in_order_cycle(order_cycle).
  joins("INNER JOIN exchange_variants ON exchanges.id = exchange_variants.exchange_id").
  select("DISTINCT exchange_variants.variant_id")

variants.where("spree_variants.id IN (#{variant_ids.to_sql})")
```

mkllnk:
I replaced the last commit with more optimised SQL.

sigmundpetersen:
Is this good to move to Test Ready then? Or does it need another review?

filipefurtad0:
Hi @mkllnk,

Tested different combinations on the filters, on four product variants, owned by 2 different producers: - found no issues! :tada: 
Good to go, thank you!

PR #4185, opened 2019-08-22
4176 Fix saving of shipping method when automatically advancing subscription order state

### LINKED ISSUE #4176: [Subs] Wrong Shipping Method Applied to Subscriptions
## Description

One specific user has reported that their Subscriptions are being generated with the same Shipping Method, regardless of the shipping method assigned in the Subscription.

This seems to be somehow related to the fact that the user uses tags to make some shipping methods available to some users and not to others. This might be a red herring, but would be my starting point as we are not having this reported by other subs users.

This is an S1 issue as it is costing the user money. The first occurrence resulted in a discounted shipping method being applied meaning that the customer lost 30% of the sale price. We are working to minimise the issue by removing the discounted shipping method, but still delivery charges are being missed. 

UK have agreed to credit the user.

## Expected Behavior

When subscription orders are created the correct shipping method is applied.

## Actual Behaviour

Subscriptions are being generated with incorrect shipping methods. The same Shipping method is being applied to all subscription orders in the OC.

## Steps to Reproduce

Chat to Lynne on Slack and I'll give you specific user details. I suspect a dev will want to try and recreate the bug using production data as it will be a little curly.

## Context

This bug is costing us money. Literal actual money. Automatic orders are generated with incorrect value due to the incorrect application of the Shipping Method.

## Severity

## Your Environment

* Version used:
* Browser name and version:

## Possible Fix

I suspect this is something curly in the application of Tagging to Subscriptions.
Part of this fix, I suspect, is improving our release and unit testing of tagging.

kristinalim:
#### What? Why?

- Closes #4176

When automatically advancing subscription orders from cart to completion, through `SubscriptionPlacementJob`, the original shipping method of the order is not retained.

We use `AdvanceOrderService` to wrap `order.next` with OFN-specific logic, such as retaining the shipping method when the order is transitioned to "delivery" state. `SubscriptionPlacementJob` still uses the `order.next` form, which loses the previously selected shipping method.

#### What should we test?

1. Create three shipping methods:
    * Volunteer discount with negative shipping fee.
    * Delivery, $1 shipping fee.
    * Express delivery, $2 shipping fee.
2. Create a subscription with the Express delivery as shipping method.
3. Open an order cycle for that subscription.
4. Wait six minutes for the order to be placed.
5. Check the shipping fee on the order. This should be $2 now.
6. Check other order details.
7. Repeat items 2 to 6 selecting Delivery as shipping method and expecting $1 shipping fee.

Now just double-checking:

- Close the order cycle.
- Wait six minutes for the confirmation emails to come through.
- Check the shipping fee.
- Check other order details.

#### Release notes

- Do not lose the previously selected shipping method for a subscription when automatically transitioning to "delivery" state through `SubscriptionPlacementJob`.

Changelog Category: Fixed

### DIFF
--- a/app/jobs/subscription_placement_job.rb
+++ b/app/jobs/subscription_placement_job.rb
@@ -59,7 +59,7 @@ class SubscriptionPlacementJob
   end
 
   def move_to_completion(order)
-    until order.completed? do order.next! end
+    AdvanceOrderService.new(order).call!
   end
 
   def unavailable_stock_lines_for(order)
--- a/spec/jobs/subscription_placement_job_spec.rb
+++ b/spec/jobs/subscription_placement_job_spec.rb
@@ -118,8 +118,12 @@ describe SubscriptionPlacementJob do
   end
 
   describe "processing a subscription order" do
-    let(:subscription) { create(:subscription, with_items: true) }
-    let(:shop) { subscription.shop }
+    let!(:shipping_method_created_earlier) { create(:shipping_method, distributors: [shop]) }
+    let!(:shipping_method) { create(:shipping_method, distributors: [shop]) }
+    let!(:shipping_method_created_later) { create(:shipping_method, distributors: [shop]) }
+
+    let(:shop) { create(:enterprise) }
+    let(:subscription) { create(:subscription, shop: shop, with_items: true) }
     let(:proxy_order) { create(:proxy_order, subscription: subscription) }
     let(:oc) { proxy_order.order_cycle }
     let(:ex) { oc.exchanges.outgoing.find_by_sender_id_and_receiver_id(shop.id, shop.id) }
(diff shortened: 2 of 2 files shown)


kristinalim:
@mkllnk Thank you very much for adding the "What should we test" part! :heart:

daniellemoorhead:
Leaving this one for you @lin-d-hop because my knowledge of subs is non-existent at the moment.

lin-d-hop:
All looks good!

Created a schedule with 3 subscriptions.
Each subscription had a different shipping method - a fee, free and an discount.
Upon OC opening each order was created with the correct shipping method.
Upon OC closing each order maintained the correct shipping method.

Ready to go!

luisramos0:
nice one!

PR #4201, opened 2019-08-29
Backordered items can be shipped

### LINKED ISSUE #4158: Cannot ship orders with variants on_demand
## Description
Manager cannot ship an order if the order contains variants that are on_demand. Variants on_demand will appear in the order as "quantity x backordered" (instead of "quantity x on_hand"), the shipment state will show pending instead of ready and the ship button (both in the order list and in the order page) will not appear. See images [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/4112#issuecomment-521584878).

## Expected Behavior
You can ship an order with on_demand variants.

## Actual Behaviour
You cannot ship orders with on_demand variants.

## Steps to Reproduce
1. place an order with a variant that is on_demand
2. try to ship the order

## Context
Detected in #4112 

## Severity
bug-s3
I dont think this is a critical feature.
The (not so good) workaround is to temporarily change the variant in the order to on_hand, ship the order and then move the variant back to on_demand.

## Your Environment
* Version used: staging katuma 2.2.2
* Operating System and version (desktop or mobile): mac desktop
* OFN Platform instance where you discovered the bug, and which version of the software they are using. 2.2.2

## Possible Fixes

- Remove backordered logic from the shipment state method [here](https://github.com/openfoodfoundation/spree/blob/8a8585a43cd04d1a50dc65227f337a91b18d66d5/core/app/models/spree/shipment.rb#L207).

- The items are marked as "backordered" when their stock is negative, and that sets the order to "backordered" state, which blocks is from being shipped. If we don't reduce stock of on_demand items below zero, it should be fixed.

Matt-Yorkley:
#### What? Why?

Closes #4158

Orders with backordered (negative stock) items could not be shipped. This included any orders with on_demand items. 

#### What should we test?

- Set a product stock level to 1, save it, then set it to `on_demand`
- Place an order with a large amount of that product
- Check the order can be successfully shipped (via admin)

#### Release notes

Fixed an issue with order shipment status getting stuck at 'pending' with on_demand items.

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/shipment_decorator.rb
+++ b/app/models/spree/shipment_decorator.rb
@@ -15,6 +15,14 @@ module Spree
       end
     end
 
+    # Override of Spree method to allow `on_demand` items to be shipped
+    def determine_state(order)
+      return 'canceled' if order.canceled?
+      return 'pending' unless order.can_ship?
+      return 'shipped' if state == 'shipped'
+      order.paid? ? 'ready' : 'pending'
+    end
+
     # The shipment manifest is built by loading inventory units and variants from the DB
     # These variants come unscoped
     # So, we need to scope the variants just after the manifest is built
--- a/spec/models/spree/shipment_spec.rb
+++ b/spec/models/spree/shipment_spec.rb
@@ -42,7 +42,7 @@ describe Spree::Shipment do
         pending_order.update!
       end
 
-      xit "can still be shipped" do
+      it "can still be shipped" do
         expect(pending_order.can_ship?).to be true
         expect(pending_order.ready_to_ship?).to be true
       end
--- a/spec/models/spree/shipment_spec.rb
+++ b/spec/models/spree/shipment_spec.rb
@@ -29,5 +29,23 @@ describe Spree::Shipment do
         expect(variants.sort_by(&:id)).to eq([deleted_variant, other_variant].sort_by(&:id))
       end
     end
+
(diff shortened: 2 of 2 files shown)


luisramos0:
they didnt because in spree logic, you need to first update stock to fulfil backorders, only then you can ship...

Matt-Yorkley:
> Have we checked whether Spree patched this in an upcoming version?

This is basically about the way we've mixed `backordered` logic with `on_demand` logic, so it's not really a Spree thing.

RachL:
@Matt-Yorkley in this shop:

https://staging.katuma.org/nanny-ogg-s-farm/shop

Buying carrots on hand can be shipped, but not on demand (error 500 - see Bugsnag notif on slack)

in this one:

https://staging.katuma.org/mary-s-online-shop/shop

I manage to buy beef on hand and on demand with the test you have described. Is it possible that the first shop has some particular data?

Side note for me: I didn't manage to display the variant "Fuji Apples 1g " in Mary's shop, I need to check if it happens with other products.

Matt-Yorkley:
@RachL I wonder if maybe the stock didn't go below zero in the second test?

RachL:
@Matt-Yorkley yes you are right I forgot the large amount part of the test... So I always get a snail / error 500 when the stock becomes negative but the item is on demand... Back to in dev guess :(

PR #4202, opened 2019-08-30
Remove Skylight

sauloperez:
#### What? Why?

Since we adopted Skylight to get response times across endpoints and
instances, we failed to get accurate numbers. Our Rails version is not
supported and thus Skylight fails to provide data for the slowest
endpoints, the ones we care about the most. Even with a supported one, we
could potentially hit any limits on tracing and have the same problem.

Recently, we started paying for Datadog's APM and the experience,
although it's still early, has been better. It makes it possible to
correlate between services and other metrics which helps to spot the
underlying issues.

Therefore, having two agents running on the server consumes system
resources so we better get rid of Skylight's one.

#### What should we test?

It does not affect the codebase.

#### Release notes

Remove response time profiling with Skylight

Changelog Category: Removed

#### Dependencies

It needs https://github.com/openfoodfoundation/ofn-install/pull/507

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,7 +85,6 @@ gem 'paperclip', '~> 3.4.1'
 gem 'rack-rewrite'
 gem 'rack-ssl', require: 'rack/ssl'
 gem 'roadie-rails', '~> 1.1.1'
-gem 'skylight', '< 2.0'
 gem 'spinjs-rails'
 
 gem 'combine_pdf'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -685,8 +685,6 @@ GEM
       rack (~> 1.4)
       rack-protection (~> 1.4)
       tilt (>= 1.3, < 3)
-    skylight (1.7.2)
-      activesupport (>= 3.0.0)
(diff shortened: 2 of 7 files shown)


sauloperez:
I need to check again but I checked in Katuma staging this week and every time I deployed this ansible got stuck in the task after `seed database` :thinking:

sauloperez:
Confirmed. This is somehow broken it also happened in UK staging

```
TASK [deploy : migrate database] **********************************************************
Friday 06 September 2019  13:48:04 +0200 (0:00:00.030)       0:00:53.518 ******
changed: [staging.openfoodnetwork.org.uk]

TASK [deploy : seed database] *************************************************************
Friday 06 September 2019  13:48:20 +0200 (0:00:16.228)       0:01:09.746 ******
(end of output)
```

sauloperez:
Ok, so it was somehow my fault. Running `ansible-playbook playbooks/deploy.yml --limit uk-staging -e "git_version=pull-request-4202"` after rebase did the trick. Merging after we get the :green_apple:

Issue #4203, opened 2019-08-30
Delete a product that has been ordered and orders/invoices cannot be updated


lin-d-hop:
## Description

A supplier loaded products onto OFN
These products were ordered
The supplier then decided they could not deliver the products and deleted them
The order products cannot be removed from the orders, cannot be removed from invoices.

## Expected Behavior

The soft-deleted products should still be able to be changed in while orders are not finalised.

## Actual Behaviour

The soft deleted products cause errors in BOM and Edit Order. The specific products cannot be altered to 0 qty. The specific products cannot be deleted. In fact the order doesn't seem to be update-able at all.

## Steps to Reproduce

1. Create products from a supplier
2. Order those products in an OC
3. Delete the products
4. Now edit the orders to try and change the quantities or remove the line items.
5. Also try changing the line items in this order in BOM

## Animated Gif/Screenshot

![Screenshot from 2019-08-30 14-13-04](https://user-images.githubusercontent.com/8669848/64023500-58c9ce00-cb30-11e9-8853-2296666f40b6.png)

## Context

A major UK hub is now having to send out incorrect invoices this week!

## Severity

bug-s2: a non-critical feature is broken, no workaround

## Your Environment

* Version used: v2.3.1
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): Desktop. 
* OFN Platform instance where you discovered the bug, and which version of the software they are using. UK

## Possible Fix

This is almost certainly related to:
#4053 
#3907 

Also overlaps with:
[Product chain data model
](https://community.openfoodnetwork.org/t/inception-session-product-chain-data-model/1285
)
[Invoicing and Delivery Note](https://community.openfoodnetwork.org/t/inception-session-delivery-note/1286)

lin-d-hop:
Downgrading to s3. The user has reported that they are happy enough with the workaround and the case that created it was a bit exceptional.

sigmundpetersen:
Okay I put this as test ready blocked by the merge of #4204 
We test it after that one is merged and if not fixed we put it back to bug backlog

RachL:
@lin-d-hop I'm not sure if I reproduced this right.... What happens at step 4 for you in production on the edit order page? Is there a button missing or an error?

lin-d-hop:
There was an error (see image). Box goes red and error message 'Fields with Red Borders Contain Errors.

Hard to miss. If you didn't notice this then it's fixed :tada:

[...]

Issue #3907, opened 2019-06-07
Delete product will break the shopfront (during the current OC) of all users who have bought that item in the current OC


RachL:
## Description
When I log in with my producer account in Katuma's production I get a weird error and cannot do anything else.

## Steps to Reproduce
Just log in with my account rachel.arnould+carrots@gmail.com

Yesterday I was testing stuff and when I was done I turned my enterprise as "not visible" anymore as I didn't want any real customer to see it. Maybe this is having an impact?

## Animated Gif/Screenshot
![image](https://user-images.githubusercontent.com/6525576/59101584-b7b0ea00-8929-11e9-9e71-886b1d6ab360.png)

## Severity
bug-s2: a non-critical feature is broken, no workaround

## Your Environment
* Version used: v.2.0.2
* Browser name and version: Firefox and Chromium
* Operating System and version (desktop or mobile): Desktop and mobile

luisramos0:
can you reproduce @RachL ? can you make the shop visible again and see if the error remains? would be good to know if it is related to that.

the error message points to a problem with the line items in the cart.
for reference, the code line where we are assigning line_item:
https://github.com/openfoodfoundation/openfoodnetwork/blob/568e3003ba43c713ac4fb5a57fddfbca086f7b98/app/assets/javascripts/darkswarm/services/cart.js.coffee#L13

RachL:
@luisramos0 yes I can reproduce @sauloperez as well. But the thing is I cannot do anything on the platform logged in with that user. A super admin needs to make the shop visible again. I can't.

sauloperez:
Good @luisramos0 I didn't find that skimming through the code. We need to investigate in DB what's wrong with that order and see which one is its root cause.

luisramos0:
the order will probably be the cart order, the only latest order of the user.

luisramos0:
Got it. Nice one!
@RachL you have bought Carrots and then deleted Carrots. That's causing this issue in the cart when showing bought items.

If you delete a product that was already sold as part of the current Order Cycle. You will break the cart and therefore the full frontoffice of all users who have bought that item in the current Order Cycle.

To replicate the issue:
- Add product to OC
- Checkout an order with that product
- Delete the product (while the OC is still open)
- Go back to shopfront

luisramos0:
it's very unlikely a user deletes a product that was just sold. Can we downgrade this to S3?

[...]

PR #4204, opened 2019-08-30
Cartastrophe

### LINKED ISSUE #4053: Delete a product that is sold in an OC and all buyers with this item in their cart will see an error 500
## Description
As a manager, if I delete a product that is sold in an OC, all buyers with this item in their cart will see an error 500 when trying to reach the website.

## Steps to Reproduce
1. As buyer, add a variant in your cart
2. In a private windows, connect as a manager and delete that product while leaving the OC open
3. Try to reach the website as a buyer again (ANY page) and you will see an error 500, 

## Severity

## Your Environment
* Version used: v2.2
* Operating System and version (desktop or mobile): Desktop Ubuntu 19.10

Matt-Yorkley:
#### What? Why?

Closes #4053

Could affect #3907, #4203, others?

Missing Spree::Price records were causing fatal errors in various scenarios where a product had been deleted.

Prices are now soft-deleted and can be accessed by associated variants.

#### What should we test?

- Add a product to the cart
- Go to backoffice and delete the product
- Go back to the cart
- The site is ok, and there's no error 500 

#### Release notes

Added soft-delete to price records to fix issues with soft-deleted products

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/price_decorator.rb
+++ b/app/models/spree/price_decorator.rb
@@ -1,6 +1,13 @@
 module Spree
   Price.class_eval do
+    acts_as_paranoid without_default_scope: true
+
     after_save :refresh_products_cache
+
+    # Allow prices to access associated soft-deleted variants.
+    def variant
+      Spree::Variant.unscoped { super }
+    end
 
     private
 
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -91,6 +91,11 @@ Spree::Variant.class_eval do
     can_supply?(quantity)
   end
 
+  # Allow variant to access associated soft-deleted prices.
+  def default_price
+    Spree::Price.unscoped { super }
+  end
+
   def price_with_fees(distributor, order_cycle)
     price + fees_for(distributor, order_cycle)
   end
--- a/db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
+++ b/db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
@@ -1,0 +1,6 @@
+class AddDeletedAtToSpreePrices < ActiveRecord::Migration
+  def change
+    add_column :spree_prices, :deleted_at, :datetime
+    add_index :spree_prices, :deleted_at
+  end
+end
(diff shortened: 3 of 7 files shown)


lin-d-hop:
:point_up: #checkastrophe @Matt-Yorkley (your pun was better)

Matt-Yorkley:
Okay... there's some weird custom delegation for price via a Spree method called `delegate_belongs_to`, which seems a bit flaky. Also, there's some callback somewhere that's removing the price object and replacing it with a blank one before it can be soft-deleted, so the object still isn't available... :disappointed:

app/models/spree/price_decorator.rb
```diff
@@ -1,7 +1,14 @@
 module Spree
   Price.class_eval do
+    acts_as_paranoid without_default_scope: true
+
     after_save :refresh_products_cache
 
+    # Allow prices to access associated soft-deleted variants.
+    def variant
+      Spree::Variant.unscoped { super }
+    end
```
luisramos0: I think this deserves a unit test.

app/models/spree/variant_decorator.rb
```diff
@@ -91,6 +91,11 @@ def in_stock?(quantity = 1)
     can_supply?(quantity)
   end
 
+  # Allow variant to access associated soft-deleted prices.
+  def default_price
+    Spree::Price.unscoped { super }
+  end
```
luisramos0: I think this deserves a unit test.

Matt-Yorkley:
> Have you checked whether this patch was introduced in Spree later on in their Git history?

They've done basically the same thing in 2.1:
https://github.com/spree/spree/blob/2bbf2a5a517e5ea62389cd90e58e3e91e14bf117/core/app/models/spree/price.rb#L26-L29
https://github.com/spree/spree/blob/2bbf2a5a517e5ea62389cd90e58e3e91e14bf117/core/app/models/spree/variant.rb#L88-L90

sauloperez:
That's a great validation :+1:

RachL:
Both the scenario in the PR and the linked issue are now fixed :muscle:

image


This is ready to go!

luisramos0:
We should retest #3907 and  #4203 right? this PR is maybe fixing them.

RachL:
@luisramos0 yes that's the plan, but not sure I will managed today, will keep people posted

Issue #4210, opened 2019-09-02
[Performance] UK Server is a temperamental beast


lin-d-hop:
## Description

This is possibly the most vague issue ever created....

UK server is temperamental. I find that in the mornings I can load pages about 3 times faster than in the afternoon. Some days we have multiple user complaints. Other days everything seems fine. 

I have no pattern to this. I have no explanation. All I have is frustration. And that is useful to no one.

## Expected Behavior

Pages load before the user's hair goes grey.

## Actual Behavior

Chaos.

## Steps to Reproduce

1. Ummmmmmm
2.
3.
4. :Shrugs:

## Context

We know we have a lot of performance issues.... but mostly they are consistent. This smells like there is something else going on on the server.

## Severity

I'm going to make an S3 for now.... but hoepfully a DevOps person will upgrade it :prayer:

## Your Environment

* Version used: v2.3.1 UK
* Browser name and version:
* Operating System and version (desktop or mobile):

sauloperez:
> I find that in the mornings I can load pages about 3 times faster than in the afternoon. Some days we have multiple user complaints.

We can start to correlate that with the available metrics. Could share dates when these happened?

lin-d-hop:
Will from now. 
Reported 3:34pm 02/09

lin-d-hop:
20:23 03/09
The following pages have taken 40+ seconds to load.
/Enterprises 
/Order_Cycles

lin-d-hop:
11:75 04/09
The site is also still regularly timing out when loading too (reports, order cycles).

lin-d-hop:
12:47 04/09
 It’s just taken me 1hr to fill in one order and it was an incredibly painful process.

lin-d-hop:
12:17 04/09
Also the system is very slow atm and my shop front won't even open at the moment

sauloperez:
I've been taking a look and these is what I found out from the screenshot below:



* The busiest hours are from 10h to 19h
* within that time window
  * `Spree::Api::ProductsController#bulk_products` is our worst ofender. Good news is that v2.4.0 includes a fix and we should see it go down the list
  * `ShopsController#index` follows very closely and it impacts even more users (not only admins).

luisramos0:
hey @sauloperez nice analysis and really nice perf data in datadog btw 👏 

point i is fixed and to be released.
point iv is very well known, we need pagination on the product list.
[...]

PR #4224, opened 2019-09-05
4210 Fetch and scope variants for shop in ProductsRenderer only once

kristinalim:
#### What? Why?

- References #4210 

This caches in instance variable `ProductsRenderer#all_variants_for_shop` so that the variants do not have to be fetched and scoped twice for the two times that this method is called.

This is a small code change that could considerably reduce the RAM and processing done for `RefreshProductsCacheJob`. It takes some time for Ruby garbage collection to happen, so for a shop that has thousands of variants in their order cycle, this could mean `number_of_variants + number_of_variant_overrides` fewer objects in memory, in addition to the I think only slightly reduced number of SQL queries.

(Maybe garbage collection is something we could look into too, if we're not doing this already.)

#### What should we test?

For below:

- Supplier stock settings - What you can set in Bulk Edit Products page
- Variant override stock settings - What you can set in Inventory page

Do the following with an open OC:

1. Create a new variant with unlimited stock (supplier).
2. Add the variant to the OC. The variant should be available in the shop.
3. Test the following:
    - 0 limited stock (supplier) - not available in shop
    - Positive limited stock (supplier) - available in the shop
4. In the Inventory page, add a variant override for the variant.
5. Test the following:
    - 0 limited stock (variant override) - not available in shop
    - Positive limited stock (variant override) - available in shop
    - 0 limited stock (variant override) again - wait for variant to disappear again
    - Unlimited stock (variant override) - available in shop

#### Release notes

- Fetch and scope variants for shop in `ProductsRenderer` used in `RefreshProductsCacheJob` only once.

Changelog Category: Changed

### DIFF
--- a/lib/open_food_network/products_renderer.rb
+++ b/lib/open_food_network/products_renderer.rb
@@ -54,14 +54,16 @@ module OpenFoodNetwork
     end
 
     def all_variants_for_shop
-      # We use the in_stock? method here instead of the in_stock scope because we need to
-      # look up the stock as overridden by VariantOverrides, and the scope method is not affected
-      # by them.
-      scoper = OpenFoodNetwork::ScopeVariantToHub.new(@distributor)
-      Spree::Variant.
-        for_distribution(@order_cycle, @distributor).
-        each { |v| scoper.scope(v) }.
-        select(&:in_stock?)
+      @all_variants_for_shop ||= begin
+                                   # We use the in_stock? method here instead of the in_stock scope
+                                   # because we need to look up the stock as overridden by
+                                   # VariantOverrides, and the scope method is not affected by them.
+                                   scoper = OpenFoodNetwork::ScopeVariantToHub.new(@distributor)
+                                   Spree::Variant.
+                                     for_distribution(@order_cycle, @distributor).
+                                     each { |v| scoper.scope(v) }.
+                                     select(&:in_stock?)
+                                 end
     end
 
     def variants_for_shop_by_id


sauloperez:
> It takes some time for Ruby garbage collection to happen,

Very good point. It's worth enabling Datadog's Ruby GC monitoring. According to https://docs.datadoghq.com/tracing/setup/ruby/ it will fetch Garbage collection statistics which will allow us to validate this claim.

lib/open_food_network/products_renderer.rb
```diff
@@ -54,14 +54,16 @@ def taxon_order
...
-      Spree::Variant.
-        for_distribution(@order_cycle, @distributor).
-        each { |v| scoper.scope(v) }.
-        select(&:in_stock?)
+      @all_variants_for_shop ||= begin
+                                   # We use the in_stock? method here instead of the in_stock scope
+                                   # because we need to look up the stock as overridden by
+                                   # VariantOverrides, and the scope method is not affected by them.
+                                   scoper = OpenFoodNetwork::ScopeVariantToHub.new(@distributor)
+                                   @all_variants_for_shop = Spree::Variant.
```
sauloperez: `@all_variants_for_shop =` is redundant here

Matt-Yorkley: Ah I missed that. Yeah, that shouldn't be there.

sauloperez:
Unfortunately, until we get rid of VariantOverrides we have a blocker to properly improve the performance of many endpoints and jobs.

lib/open_food_network/products_renderer.rb
kristinalim: Whoops! Sorry, that was from an earlier change. Changed already.

PR #4225, opened 2019-09-05
Add missing translations on product edit page

### LINKED ISSUE #4106: Lot's of missing translations on product edit page
## Description
Lots of translations seem to have disappeared in the product editing page :
1.
Capture du 2019-07-31 16-34-04

2. 
Capture du 2019-07-31 16-36-26

3. 
Capture du 2019-07-31 16-37-37

4. 
Capture du 2019-07-31 16-38-59

5.
Capture du 2019-07-31 16-40-15

## Steps to Reproduce

1. On non English speaking instance, go to `admin/products`
2. Choose a product
3. Go on `/edit`

## Severity
S3

## Your Environment

* OFN Platform instance where you discovered the bug, and which version of the software they are using : French v 2.2.1

HugsDaniel:
#### What? Why?

Closes #4106 

Add missing keys for translations and override Spree keys to use our own.

#### What should we test?
Missing translations should now appear. Verify all pages mentioned in the issue are rendering correctly in English.

#### Release notes
Adding missing translations in edit product pages.
Changelog Category: Added | Changed

### DIFF
--- a/app/views/spree/admin/shared/_edit_resource_links.html.haml
+++ b/app/views/spree/admin/shared/_edit_resource_links.html.haml
@@ -1,4 +1,4 @@
 .form-buttons.filter-actions.actions
-  = button Spree.t('actions.update'), 'icon-refresh'
-  %span.or= Spree.t(:or)
-  = button_link_to Spree.t('actions.cancel'), collection_url, icon: 'icon-remove'
+  = button t('update'), 'icon-refresh'
+  %span.or= t(:or)
+  = button_link_to t('cancel'), collection_url, icon: 'icon-remove'
--- a/app/views/spree/admin/shared/_edit_resource_links.html.haml
+++ b/app/views/spree/admin/shared/_edit_resource_links.html.haml
@@ -1,4 +1,4 @@
 .form-buttons.filter-actions.actions
-  = button t('update'), 'icon-refresh'
+  = button t(:update), 'icon-refresh'
   %span.or= t(:or)
-  = button_link_to t('cancel'), collection_url, icon: 'icon-remove'
+  = button_link_to t(:cancel), collection_url, icon: 'icon-remove'
--- a/app/views/spree/admin/product_properties/index.html.haml
+++ b/app/views/spree/admin/product_properties/index.html.haml
@@ -7,7 +7,7 @@
 - content_for :page_actions do
   %ul.tollbar.inline-menu
     %li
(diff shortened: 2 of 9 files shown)


HugsDaniel:
In variant edit, when ticking the "On demand" checkbox it shows a missing translation, even though the key is present :thinking: 
Strange thing is that it works in Spanish and French, just not in English... That's what makes the test fail. Any ideas ?

luisramos0:
it a collision @HugsDaniel 
you cannot add "variants: "Variants"" as there is already a variants that has nested keys. that's why variants.infinity doesnt work.

luisramos0:
there are more missing translations in edit variants, you can inspect and search for "missing translations"

also, :variants  is different from .variants, :variants will go to the route. I am not sure admin.products.variants is doing what you need.

luisramos0:
ok, ready for review.

luisramos0:
@HugsDaniel in the last delivery train catch up we decided this was hanging around for quite some time and someone should move it forward.

app/views/spree/admin/shared/_edit_resource_links.html.haml
```diff
@@ -1,4 +1,4 @@
 .form-buttons.filter-actions.actions
-  = button t('update'), 'icon-refresh'
+  = button t(:update), 'icon-refresh'
```
mkllnk: This is semantically the same, right? Just better way of writing it.

luisramos0: here we move from the key spree.actions.update to the key update.

RachL:
Alright I've staged this on staging UK and nothing looks broken.

Sorry to all new testers to test something as easy as this, but I need this one really badly in production and wanted to make sure this would be included in the next release... this PR is open since September the 5th.... :eyes: let's meeerge it!!

app/views/spree/admin/shared/_edit_resource_links.html.haml
mkllnk: Really? How does `t('update')` know to look in spree.actions.update? Or do you mean in combination with other commits?

luisramos0: 🙈 sorry Maikel, I saw a inexistent dot. yes, this is the same but better looking :-) and less  confusion with the lazylookup.

PR #4226, opened 2019-09-05
Add missing translations for orders

### LINKED ISSUE #4090: Missing translations / misleading automatic translations from Spree on orders pages
## Description

With v2 new badly automatic translations from Spree have appeared on the orders section, and there are also missing translation that (re)appeared... I'm pretty sure some of those were translated before the upgrade. We need to enable translation of those strings through transifex to be able to translate them properly.

## Animated Gif/Screenshot

1.
Untitled presentation
2. On the edit order page:
Untitled presentation(1)
3.
Untitled presentation(2)
4.
Untitled presentation(3)
5.
Untitled presentation(4)
6.
Untitled presentation(5)

## Context

User reports on OFFrance for part of those missing translations. We added some we noticed ourselves.

## Severity
What is the severity for translations, S3 ? 

## Your Environment

* OFN Platform instance where you discovered the bug, and which version of the software they are using : France on v2.2.1

HugsDaniel:
#### What? Why?

Closes #4090 

Add missing translations for order pages. Same, most of them where already here but not used. I overide a few of Spree's.

#### What should we test?

Missing translations now have keys in en.yml.

#### Release notes

Added missing translations for order pages.

Changelog Category: Added

### DIFF
--- a/app/views/spree/admin/adjustments/index.html.haml
+++ b/app/views/spree/admin/adjustments/index.html.haml
@@ -2,12 +2,12 @@
 
 - content_for :page_title do
   %i.icon-arrow-right
-  = Spree.t(:adjustments)
+  = t(:adjustments)
 
 - content_for :page_actions do
-  %li= button_link_to Spree.t(:new_adjustment), new_admin_order_adjustment_url(@order), :icon => 'icon-plus'
-  %li= button_link_to Spree.t(:back_to_orders_list), admin_orders_path, :icon => 'icon-arrow-left'
+  %li= button_link_to t(:new_adjustment), new_admin_order_adjustment_url(@order), :icon => 'icon-plus'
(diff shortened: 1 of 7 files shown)


luisramos0:
broken build, I rebuild to see if it was a flaky one.

RachL:
@HugsDaniel it looks good except one missing translation on order page:

image


Example here: https://staging.katuma.org/admin/orders/R521350017/edit

Is this just because there is no translation in English? Or is this something missing on your side?

luisramos0:
this is good to go Rachel. on_hand is a new key added in this PR so the Portuguese translation is not present: pt.on_hand

RachL:
ah thanks @luisramos0 ! But shouldn't it fallback to English?

luisramos0:
I am not sure why it's not showing the english version Rachel but I think it's related to the haml code used in this case which is just a plain call to the t function to translate:
= "#{count} x #{t(state.humanize.downcase)}"

I think we are ok here.

PR #4234, opened 2019-09-06
Remove integrity check for deleted variants in OCs

sauloperez:
#### What? Why?

We have no way to see the output of this weird RSpec integrity check
thus, it's pointless. It only adds unnecessary load the server.

However, the issue it checks seems to still be happening although
not very often. UK's server has 70 occurrences the last one being from
early 2019 while FR's server returns 10.

#### What should we test?

Nothing. This is just a harmless check that doesn't affect the behaviour of the app.

#### Release notes

Removed daily check regarding the integrity of order cycle variants for which we can't see the output.

Changelog Category: Removed


sauloperez:
Shall we create an issue to check this? it's clearly a bug if it happens.

PR #4236, opened 2019-09-06
4233 Add missing Spree route for Paypal

### LINKED ISSUE #4233: NoMethodErrorspree/paypal#express
This is a more specific and technical version of #4222 and may not be the cause of it. So opening a new issue. Severity should be the same as 4222 imo.

### Error in OpenFoodNetwork

link to bugsnag:
https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5d6ea59aa104a600184d9e67?filters[event.since][0]=30d&filters[error.status][0]=fixed

Screenshot 2019-09-06 at 09 18 26

### Possible fix
To fix the paypal express error (the checkout would fail, it's just the recovery so that user does not see the snail if paypal payment fails, user should be redirected to checkout page) we need to re-add this route:
get '/checkout/:state', :to => 'checkout#edit', :as => :checkout_state

mkllnk:
#### What? Why?

Closes #4233

Commit c56486d7aeaee822adc93f1f6d43be282a178a89 removed a couple of Spree routes that are not used in our code. But one of them is used by spree_paypal_express to redirect when something goes wrong. I added the needed route back in.

I also added a new feature spec for this case. Automated testing of a successful Paypal checkout may be difficult because it includes redirects to the Paypal website. Maybe we can fake it but I would do that in a separate task. I also think that an automated integration test for staging would be easier to set up and give us more value because it would test the real integration with the Paypal website.

The new spec adds 10 seconds to the test suite but I don't see a better way. A unit test would be on controller level within spree_paypal_express. But within that gem the test would pass. The problem is that the gem depends on Spree and we connect it to our lightweight version of Spree. That's a case for an integration test what I did here.

And I did some styling of the current checkout spec when I was analysing what we are testing already.

#### What should we test?

We should validate two scenarios:

Failed payment:

- Set up a Paypal payment method with invalid details (just don't enter a key).
- Checkout using that payment method.
- You should see an error message.

Successful payment:

- Set up a Paypal payment method with valid details, connected to a real testing account.
- Checkout using that payment method.
- The order should be successful and paid.

#### Release notes

Fixed: Buyers see a useful error message when Paypal isn't set up correctly.

Changelog Category: Fixed

### DIFF
--- a/config/routes/spree.rb
+++ b/config/routes/spree.rb
@@ -117,6 +117,9 @@ Spree::Core::Engine.routes.prepend do
 
   resources :products
 
+  # Used by spree_paypal_express
+  get '/checkout/:state', :to => 'checkout#edit', :as => :checkout_state
+
   get '/unauthorized', :to => 'home#unauthorized', :as => :unauthorized
   get '/content/cvv', :to => 'content#cvv', :as => :cvv
   get '/content/*path', :to => 'content#show', :as => :content
--- a/spec/features/consumer/shopping/checkout_paypal_spec.rb
+++ b/spec/features/consumer/shopping/checkout_paypal_spec.rb
@@ -43,8 +43,6 @@ feature "Checking out with Paypal", js: true do
 
   describe "as a guest" do
     it "fails with an error message" do
-      pending "The checkout_state_path is missing."
-
       visit checkout_path
       complete_the_form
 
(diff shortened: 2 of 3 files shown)


lin-d-hop:
Added the paypal account with incorrect details. It appears as though everything works successfully. This is obviously where people fail. I think this is the right solution for a quick s1 fix. It transfers the onus onto support people help people navigate. I think this is fine at the moment but worth mentioning.

Tested checkout with incorrect paypal set up:
Screenshot from 2019-09-10 11-39-24
We get an error message as we expect :check:

Testing with a real paypal account added:
Successfully redirected to paypal for payment.
Test payment made and order completed.

Ready to go!

luisramos0:
thanks @lin-d-hop

PR #4240, opened 2019-09-06
Disentangle on_demand and backordered

### LINKED ISSUE #4158: Cannot ship orders with variants on_demand

### LINKED ISSUE #4217: on_demand and backorderd
## Description

Purchasing `on_demand` products lowers the variant's stock, and can lower it below zero. Spree 2.0 then marks this variant as `backordered`, and also marks any shipments that contain this item as `backordered`. This then triggers some confusing feedback and some unwanted logic; the items and shipments are clearly marked as `backordered` in the UI which doesn't make sense in this context, and shipments in the `backordered` state can't be shipped and can't transition correctly. This can apply to any order with an `on_demand` item in it.

## Expected Behavior

`on_demand` items and related shipments are not labelled as `backordered`, orders containing `on_demand` items can be shipped.

## Steps to Reproduce

1. Create a product with stock: 1.
2. Set it to `on_demand`.
3. Order more than 1 of the product.
4. Look at the order and it's line_items as an enterprise user. (Confusing UI)
5. Try to ship the order. (Fails)

## Context

Investigating a fix for issue #4158, this sums up the root cause of the issue.
Also related to discussions in #4125, such as: https://github.com/openfoodfoundation/openfoodnetwork/pull/4125#issuecomment-519593092

## Severity

## Possible Fix

- Ensure `on_demand` items don't reduce the stock below zero. 
- Possibly add a migration to move all on_demand variants with negative stock levels back up to zero.

Matt-Yorkley:
#### What? Why?

Closes #4217 (original bug report: #4157)

Adjusts `on_demand` logic so that shipments no longer decrement an `on_demand` variant's stock levels. This means that `on_demand` items will no longer go into negative stock levels, and `on_demand` variants and their associated shipments will no longer be marked by Spree as `backordered`, which triggers a lot of unwanted logic.

#### What should we test?

Buying an `on_demand` variant does not reduce it's stock level.
Buying  large quantity of an `on_demand` variant does not cause the order to be unshippable.
Orders with `on_demand` variants do not get their shipment state marked as `backordered`.

#### Release notes

Variants marked `on_demand` no longer reduce stock levels when sold, and are not marked as `backordered`

Changelog Category: Changed

### DIFF
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -100,13 +100,13 @@ module VariantStock
   # Here we depend only on variant.total_on_hand and variant.on_demand.
   #   This way, variant_overrides only need to override variant.total_on_hand and variant.on_demand.
   def fill_status(quantity)
-    if on_hand >= quantity
-      on_hand = quantity
-      backordered = 0
-    else
-      on_hand = [0, total_on_hand].max
-      backordered = on_demand ? (quantity - on_hand) : 0
-    end
+    on_hand = if total_on_hand >= quantity || on_demand
+                quantity
+              else
+                [0, total_on_hand].max
+              end
+
+    backordered = 0
 
     [on_hand, backordered]
   end
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -117,6 +117,9 @@ module VariantStock
   def move(quantity, originator = nil)
     raise_error_if_no_stock_item_available
 
+    # Don't change variant stock if variant is on_demand
+    return if on_demand
+
     # Creates a stock movement: it updates stock_item.count_on_hand and fills backorders
     #
     # This is the original Spree::StockLocation#move,
--- a/db/migrate/20190908183511_set_backordered_inventory_to_on_hand.rb
+++ b/db/migrate/20190908183511_set_backordered_inventory_to_on_hand.rb
@@ -1,0 +1,5 @@
+class SetBackorderedInventoryToOnHand < ActiveRecord::Migration
+  def up
+    execute("UPDATE spree_inventory_units SET state = 'on_hand' WHERE state = 'backordered'")
+  end
+end
--- a/spec/models/spree/line_item_spec.rb
+++ b/spec/models/spree/line_item_spec.rb
@@ -132,6 +132,54 @@ module Spree
       end
     end
 
+    describe "reducing stock levels on order completion" do
+      context "when the item is on_demand" do
+        let!(:hub) { create(:distributor_enterprise) }
+        let(:bill_address) { create(:address) }
+        let!(:variant_on_demand) { create(:variant, on_demand: true, on_hand: 1) }
+        let!(:order) {
+          create(:order_with_totals_and_distribution,
+                 distributor: hub,
+                 bill_address: bill_address,
+                 ship_address: bill_address)
+        }
+        let!(:shipping_method) { create(:shipping_method, distributors: [hub]) }
+        let!(:line_item) { create(:line_item, variant: variant_on_demand, quantity: 10, order: order) }
+
+        before do
+          order.update_totals
+          order.payments << create(:payment, amount: order.total)
+          until order.completed? do break unless order.next! end
+          order.payment_state = 'paid'
+          order.select_shipping_method(shipping_method.id)
+          order.shipment.update!(order)
+        end
+
+        it "creates a shipment with backordered items" do
+          expect(order.shipment.manifest.first.quantity).to eq 10
+          expect(order.shipment.manifest.first.states).to eq 'on_hand' => 1, 'backordered' => 9
+          expect(order.shipment.manifest.first.variant).to eq line_item.variant
+        end
+
+        it "reduces the variant's stock level" do
+          expect(variant_on_demand.reload.on_hand).to eq(-9)
+        end
+
+        it "marks the inventory units as backorderd" do
+          backordered_units = order.shipments.first.inventory_units.any?(&:backordered?)
+          expect(backordered_units).to be true
+        end
+
+        it "marks the shipment as backorderd" do
+          expect(order.shipments.first.backordered?).to be true
+        end
+
+        it "does not allow the order to be shipped" do
+          expect(order.ready_to_ship?).to be false
+        end
+      end
+    end
+
     describe "tracking stock when quantity is changed" do
       context "when the order is already complete" do
         let(:shop) { create(:distributor_enterprise) }
@@ -186,12 +234,18 @@ module Spree
       let!(:hub) { create(:distributor_enterprise) }
       let!(:o) { create(:order, distributor: hub) }
       let!(:v) { create(:variant, on_demand: false, on_hand: 10) }
+      let!(:v_on_demand) { create(:variant, on_demand: true, on_hand: 1) }
       let!(:li) { create(:line_item, variant: v, order: o, quantity: 5, max_quantity: 5) }
+      let!(:li_on_demand) { create(:line_item, variant: v_on_demand, order: o, quantity: 99, max_quantity: 99) }
 
       before do
         # li#scoper is memoised, and this makes it difficult to update test conditions
         # so we reset it after the line_item is created for each spec
         li.remove_instance_variable(:@scoper)
+      end
+
+      context "when the variant is on_demand" do
+        it { expect(li_on_demand.sufficient_stock?).to be true }
       end
 
       context "when stock on the variant is sufficient" do
--- a/spec/models/spree/line_item_spec.rb
+++ b/spec/models/spree/line_item_spec.rb
@@ -161,8 +161,8 @@ module Spree
           expect(order.shipment.manifest.first.variant).to eq line_item.variant
         end
 
-        it "reduces the variant's stock level" do
-          expect(variant_on_demand.reload.on_hand).to eq(-9)
+        it "does not reduce the variant's stock level" do
+          expect(variant_on_demand.reload.on_hand).to eq 1
         end
 
         it "does not mark inventory units as backorderd" do
@@ -176,6 +176,11 @@ module Spree
 
         it "allows the order to be shipped" do
           expect(order.ready_to_ship?).to be true
+        end
+
+        it "does not change stock levels when cancelled" do
+          order.cancel!
+          expect(variant_on_demand.reload.on_hand).to eq 1
         end
       end
     end
--- a/spec/models/spree/line_item_spec.rb
+++ b/spec/models/spree/line_item_spec.rb
@@ -155,9 +155,9 @@ module Spree
           order.shipment.update!(order)
         end
 
-        it "creates a shipment with backordered items" do
+        it "creates a shipment without backordered items" do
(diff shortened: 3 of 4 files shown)


app/models/concerns/variant_stock.rb
```diff
@@ -100,7 +100,7 @@ def can_supply?(quantity)
   # Here we depend only on variant.total_on_hand and variant.on_demand.
   #   This way, variant_overrides only need to override variant.total_on_hand and variant.on_demand.
   def fill_status(quantity)
-    if on_hand >= quantity
+    if on_hand >= quantity || on_demand
```
luisramos0: do you know if we could we just get rid of backordered here and always return on_hand = quantity and backordered = 0?
now that I consider this, maybe we never get to backordered if variant is on_hand, that would mean we are fullfilling an order with outofstock variant.

so, this change may make, the else part of this if, dead code... what do you think?

Matt-Yorkley:
Ok, I've added a migration to change existing inventory_units from `backordered` to `on_hand`. 

Shall we move the second part of the cleanup to another issue/PR?

> bring StockItem#adjust_count_on_hand to ofn and remove process_backorders from it, this will make things faster and cleaner, maybe we can even bring the full stock item class (bye bye spree) as half of the class methods are related to this process_backorders

Matt-Yorkley:
> tbh I am a bit nervous with this PR but we have to do it!

Agreed on both counts!

luisramos0:
build is broken with a related error.

app/models/concerns/variant_stock.rb
```diff
@@ -117,6 +117,9 @@ def fill_status(quantity)
   def move(quantity, originator = nil)
     raise_error_if_no_stock_item_available
 
+    # Don't change variant stock if variant is on_demand
+    return if on_demand
```
mkllnk: This is an alternative fix to one of the products cache issues we had. :-)

db/migrate/20190908183511_set_backordered_inventory_to_on_hand.rb
```diff
@@ -0,0 +1,5 @@
+class SetBackorderedInventoryToOnHand < ActiveRecord::Migration
+  def up
+    Spree::InventoryUnit.backordered.update_all(state: 'on_hand')
```
mkllnk: It's better to avoid existing models. Either use plain SQL or define your own model here. Otherwise this will fail when we delete the `backordered` scope from InventoryUnit. I think you can just include this in the migration:

```rb
class SpreeInventoryUnit < ActiveRecord::Base
  scope :backordered, -> { where state: 'backordered' }
end
```
But plain SQL might be easier here.

sauloperez: Yes, please. It's very painful when later on the model changed what the migration was relying on 🙈

Might be time to introduce https://github.com/testdouble/good-migrations

spec/models/spree/line_item_spec.rb
```diff
@@ -132,6 +132,59 @@ module Spree
...
+        it "allows the order to be shipped" do
+          expect(order.ready_to_ship?).to be true
+        end
+
+        it "does not change stock levels when cancelled" do
+          order.cancel!
+          expect(variant_on_demand.reload.on_hand).to eq 1
+        end
+      end
+    end
```
sauloperez: IMO this belongs to a layer above, controller tests. This file contains line items unit tests but the examples that we added not even reference a line item.

sauloperez:
> I have the feeling there's an impact on other areas, Ill check a few things on Sunday/Monday.

Any findings @luisramos0 ? I think we all have that fear. It happened before.

As for the migration, as far as I understand we don't change behavior but fix broken data. So I don't see the need to make it reversible.

luisramos0:
I did check the code around... I didnt find anything, looks good. I am still a bit nervous.
Anyway, I think this is awesome, specially because it opens up the possibility of bringing the spree code to our side and delete quite a lot of it :tada

db/migrate/20190908183511_set_backordered_inventory_to_on_hand.rb
luisramos0: yes, I think that's correct Kristina.

luisramos0:
good point Kristina! I checked these scenarios and it's a valid alternative to process orders and update shipment etc through the orderUpdater for example, but I'd not go that way...
I think we will be ok if we leave shipment.states and order.shipment_states as they are. Also I don't think we will have a problem if we leave  order state as is.
What do you think is the problem of leaving these workflows as they are?

kristinalim:
> I think we will be ok if we leave shipment.states and order.shipment_states as they are. Also I don't think we will have a problem if we leave order state as is.
> What do you think is the problem of leaving these workflows as they are?

I need to test this a bit more in a few hours to verify (maybe my local dev data is not valid anymore), but as far as I saw, updating `StockItem` and `InventoryUnit` are not enough to recover an existing order with backordered items.

If we have no goal to recover existing orders and won't be eliminating the "backorder" shipment state of orders yet, then I think we should not touch existing backordered `InventoryUnit` records in the DB migration too, just leave them there for now.

luisramos0:
ok :+1:

Updating inventory_units state will fix the problem for all orders because the order edit page displays the inventory_units state [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/1d177a164bc71f7f6e8e6169f25cf8fad138c1aa/app/views/spree/admin/orders/_shipment_manifest.html.haml#L12).

I think you are worried the system stays in an incoherent state. By "recovering the orders" you mean moving the workflow to the correct place, right?

Matt-Yorkley:
I don't think the shipment state `backordered` actually exists in the database, it just gets shown as tht in the UI if the shipment contains any `backordered` items.

db/migrate/20190908183511_set_backordered_inventory_to_on_hand.rb
Matt-Yorkley: @mkllnk I've updated the migration to plain SQL.

luisramos0:
@kristinalim "I need to test this a bit more in a few hours"
Did you manage to test this? Any interesting findings?

kristinalim:
:v: So, I was able to confirm `Spree::Order#shipment_state` is indeed set to "backorder" for these orders. Please see the screen recording below:

20190914010200-backorder_shipment_state

After running the migrations in this branch, when the shipment state is "backorder" and I load the admin edit order page, the Ship option is not yet available but order advancement that happens behind the scenes changes the shipment state from "backorder" to "ready". After this first page load, it is possible to ship the order.

kristinalim:
I think this is okay to leave as is for now, but we should note this so we don't forget that there could be "backorder" shipment state in production data when eliminating backorder logic from OFN.

I'm seeing two relevant places for this backorder state:

* `Spree::OrderUpdater#update_shipment_state`
* `Spree::Shipment#determine_state`

kristinalim:
@Matt-Yorkley @luisramos0 Even with my concern above, what do you think of moving this PR forward and creating the tech debt issue for removing backorders and then noting possible "backorder" shipment state in existing data?

Matt-Yorkley:
Ah! I didn't realise the `spree_orders` table had it's own `shipment_state` column, I thought it was just getting it from the shipments. Confusingly, `Spree::Shipment` doesn't have a `backordered` state, but `Spree::Order` does. 

I'll add another migration. :+1:

Matt-Yorkley:
Ah, but we don't know whether to change it to `pending` or `ready` without checking all the other parts of the order...

sauloperez:
Awesome investigation @kristinalim . Let's approve the PR and do what you suggest :+1:

kristinalim:
I created #4267 for above concern. :zap:

luisramos0:
great stuff, thanks @kristinalim
test ready 🎉

RachL:
**Test case number 1** : Buying an on_demand variant does not reduce it's stock level.

I bought 3 carrots and 3 garlics from Freddo's farm hub. Carrots where at hidden stock 1 and garlic at 0.

The stock didn't move:

image

**Test case number 2** : Buying large quantity of an on_demand variant does not cause the order to be unshippable.

I bought 9999 Garlics at fred's farm hub (we will talk about bad breath another time...). And I could ship the order :tada:

**Test case number 3** : Orders with on_demand variants do not get their shipment state marked as backordered.

So this is true... however it is written "on_hand"... Anything to be concerned about @Matt-Yorkley ?

image

luisramos0:
in terms of test case 3, we have changed the tag from backordered to on_hand, that's why they show on_hand :+1:
I dont know why shipment state is "none" but I pretty sure it's not related to this PR. I have seen it before... the tech debt PR created could fix it: #4267 
I think this is good to go.

Matt-Yorkley:
> I think this is good to go.

With `shipment` and `payment` both displaying "none"?

luisramos0:
it's just a missing translation:
`= t(@order.shipment_state, scope: :shipment_states, default: [:missing, "none"])`

kristinalim:
By the way, I included this in the changelog as a "Fixed" item.

RachL:
@luisramos0 @ Matt-Yorkley About the missing translation. Do we need an issue? I see from Theresa's screenshot that the "none" is still appearing in production.

Moreover, the text in the quantity column is always showing the class "translation missing". Is this the new process to spot missing translation keys?

luisramos0:
They are two different issues actually:
- "the "none" is still appearing in production" - this was already the case before the release, yes, we need an issue to fix it
- "the class "translation missing"" - I am not sure I understand if it's showing missing translation for the user, or just the CSS class. I am not sure we can fix it with transifex or if we need an issue, can you share details here please so we can clarify? thanks!

luisramos0:
ok, I see the second point is now in https://github.com/openfoodfoundation/openfoodnetwork/issues/4302

RachL:
Yes, I will open a new issue for the "none"

PR #4242, opened 2019-09-06
Soft-delete price objects

### LINKED ISSUE #3907: Delete product will break the shopfront (during the current OC) of all users who have bought that item in the current OC

### LINKED ISSUE #4053: Delete a product that is sold in an OC and all buyers with this item in their cart will see an error 500

Matt-Yorkley:
#### What? Why?

Closes #4053
Closes #3907 
Possibly #4203 as well..?

Missing Spree::Price records were causing fatal errors in various scenarios where a product had been deleted.

Prices are now soft-deleted and can be accessed by associated variants.

Variants that have previously been soft-deleted and have had their associated price objects _hard_-deleted are removed from open carts via a migration.

#### What should we test?

- Add a product to the cart
- Go to backoffice and delete the product
- Go back to the cart
- The site is ok, and there's no error 500 

#### Release notes

Added soft-delete to price records to fix issues with soft-deleted products

Changelog Category: Fixed

### DIFF
--- a/db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
+++ b/db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
@@ -1,0 +1,6 @@
+class AddDeletedAtToSpreePrices < ActiveRecord::Migration
+  def change
+    add_column :spree_prices, :deleted_at, :datetime
+    add_index :spree_prices, :deleted_at
+  end
+end
--- a/db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
+++ b/db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
@@ -1,6 +1,5 @@
 class AddDeletedAtToSpreePrices < ActiveRecord::Migration
   def change
     add_column :spree_prices, :deleted_at, :datetime
-    add_index :spree_prices, :deleted_at
   end
 end
--- a/db/migrate/20190906165501_remove_broken_variants_from_carts.rb
+++ b/db/migrate/20190906165501_remove_broken_variants_from_carts.rb
@@ -1,0 +1,17 @@
+class RemoveBrokenVariantsFromCarts < ActiveRecord::Migration
+  def up
+    # Removes line_items from open carts where the variant has a hard-deleted price
+
+    variants_without_prices = execute(
+      'SELECT spree_variants.id FROM spree_variants
+      LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
+      WHERE spree_prices.id IS NULL'
+    ).to_a.map{ |v| v['id'] }
+
+    broken_line_items = Spree::LineItem.
+      joins(:variant).where('spree_variants.id IN (?)', variants_without_prices).
+      joins(:order).merge(Spree::Order.where(state: 'cart'))
+
+    broken_line_items.each(&:destroy)
+  end
+end
--- a/app/models/spree/price_decorator.rb
+++ b/app/models/spree/price_decorator.rb
@@ -1,6 +1,13 @@
 module Spree
   Price.class_eval do
+    acts_as_paranoid without_default_scope: true
+
     after_save :refresh_products_cache
+
+    # Allow prices to access associated soft-deleted variants.
+    def variant
+      Spree::Variant.unscoped { super }
+    end
 
     private
 
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -91,6 +91,11 @@ Spree::Variant.class_eval do
     can_supply?(quantity)
   end
 
+  # Allow variant to access associated soft-deleted prices.
+  def default_price
+    Spree::Price.unscoped { super }
+  end
+
   def price_with_fees(distributor, order_cycle)
     price + fees_for(distributor, order_cycle)
   end
--- a/db/migrate/20190913023137_replace_hard_deleted_prices.rb
+++ b/db/migrate/20190913023137_replace_hard_deleted_prices.rb
@@ -1,0 +1,11 @@
+class ReplaceHardDeletedPrices < ActiveRecord::Migration
+  def up
+    ActiveRecord::Base.connection.execute(
+      "INSERT into spree_prices (variant_id, amount, currency, deleted_at)
+      SELECT spree_variants.id, '0.00', (SELECT value FROM spree_preferences WHERE key = 'spree/app_configuration/currency'), now()
+      FROM spree_variants
+      LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
+      WHERE spree_prices.id IS NULL;"
+    )
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20190701002454) do
+ActiveRecord::Schema.define(:version => 20190830134723) do
 
   create_table "adjustment_metadata", :force => true do |t|
     t.integer "adjustment_id"
@@ -649,9 +649,10 @@ ActiveRecord::Schema.define(:version => 20190701002454) do
   add_index "spree_preferences", ["key"], :name => "index_spree_preferences_on_key", :unique => true
 
   create_table "spree_prices", :force => true do |t|
-    t.integer "variant_id",                               :null => false
-    t.decimal "amount",     :precision => 8, :scale => 2
-    t.string  "currency"
+    t.integer  "variant_id",                               :null => false
+    t.decimal  "amount",     :precision => 8, :scale => 2
+    t.string   "currency"
+    t.datetime "deleted_at"
   end
 
   add_index "spree_prices", ["variant_id"], :name => "index_spree_prices_on_variant_id"
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -28,6 +28,17 @@ feature "full-page cart", js: true do
       end
     end
 
+    describe "when a product is soft-deleted" do
+      it "shows the cart without errors" do
+        add_product_to_cart order, product_with_tax, quantity: 1
+        add_product_to_cart order, product_with_fee, quantity: 2
+        product_with_fee.destroy
+
+        visit main_app.cart_path
+        expect(page).to have_selector '.cart-item-price'
+      end
+    end
+
     describe "percentage fees" do
       let(:percentage_fee) { create(:enterprise_fee, calculator: Calculator::FlatPercentPerItem.new(preferred_flat_percent: 20)) }
 
--- a/spec/models/spree/price_spec.rb
+++ b/spec/models/spree/price_spec.rb
@@ -2,10 +2,10 @@ require 'spec_helper'
 
 module Spree
   describe Price do
-    describe "callbacks" do
-      let(:variant) { create(:variant) }
-      let(:price) { variant.default_price }
+    let(:variant) { create(:variant) }
+    let(:price) { variant.default_price }
 
+    describe "callbacks" do
       it "refreshes the products cache on change" do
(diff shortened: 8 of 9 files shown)


Matt-Yorkley:
I'm not sure about this migration...

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
```diff
@@ -0,0 +1,14 @@
...
+    # Removes line_items from open carts where the variant has a hard-deleted price
+
+    variants_with_prices = Spree::Variant.
+      joins('INNER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)').
+      select('spree_variants.id')
+
+    Spree::LineItem.
+      joins(:variant).where('spree_variants.id NOT IN (?)', variants_with_prices).
+      joins(:order).merge(Spree::Order.where(state: 'cart')).
+      delete_all
```
luisramos0: 1. is this the only migration we need here? I cant remember but maybe there is more data that needs to be cleaned up?

2. I think you should invert the first query because variants_with_prices can be too big
- I think you need to use arel for this (I never used it)
variants_without_price = select * from spree_variants v where not exists (select 1 from spree_prices where variant_id = v.id)
and then same second query with IN instead of NOT IN

Matt-Yorkley: I've updated the queries and tested them more thoroughly in the Rails console. :+1:

I'm not sure how much more we can clear up in terms of data. We can't restore hard-deleted prices, and I don't think we can reasonably remove line_items from completed orders.

db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
```diff
@@ -0,0 +1,6 @@
+class AddDeletedAtToSpreePrices < ActiveRecord::Migration
+  def change
+    add_column :spree_prices, :deleted_at, :datetime
+    add_index :spree_prices, :deleted_at
```
sauloperez: I think we've discussed this before but indexes on boolean columns are little to no value and make writes slower. I wouldn't add it.

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
```diff
@@ -0,0 +1,17 @@
...
+    # Removes line_items from open carts where the variant has a hard-deleted price
+
+    variants_without_prices = execute(
+      'SELECT spree_variants.id FROM spree_variants
+      LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
+      WHERE NOT EXISTS (SELECT 1 FROM spree_prices WHERE spree_prices.variant_id = spree_variants.id)'
+    ).to_a.map{ |v| v['id'].to_i }
+
+    broken_line_items = Spree::LineItem.
+      joins(:variant).where('spree_variants.id IN (?)', variants_without_prices).
```
sauloperez: I know it might not make much difference given the numbers of affected variants but this is a pattern I see over and over again that has performance and complexity implications. We do a lot of pointless back and forth to the DB in cases that the DB can handself itself without having to resort to Ruby.

There's no need to collect the variant ids in line 9 to pass them back again if we perform a subquery. That can be achieved passing an ActiveRecord::Relation to the where method like:

```ruby
variants_without_prices = 'SELECT spree_variants.id FROM spree_variants
LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
WHERE NOT EXISTS (SELECT 1 FROM spree_prices WHERE spree_prices.variant_id = spree_variants.id)'

broken_line_items = Spree::LineItem.
      joins(:variant).
      joins(:order).
      where(variant: { id: variants_without_prices }).
      merge(Spree::Order.where(state: 'cart')).

broken_line_items.each(&:destroy)
```

db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
Matt-Yorkley: It was recommended in the paranoia gem's documentation, but I can remove it :+1:

Matt-Yorkley: Removed :+1:

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
Matt-Yorkley: Updated :+1:

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
```diff
@@ -0,0 +1,15 @@
+class RemoveBrokenVariantsFromCarts < ActiveRecord::Migration
+  def up
+    # Removes line_items from open carts where the variant has a hard-deleted price
+
+    variants_without_prices = 'SELECT spree_variants.id FROM spree_variants
+      LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
+      WHERE NOT EXISTS (SELECT 1 FROM spree_prices WHERE spree_prices.variant_id = spree_variants.id)'
```
mkllnk: Wow, I've never seen this clause. I would have written this:
```sql
WHERE spree_prices.id IS NULL
```

In the joined table, all variants that don't have a price will have NULL in all spree_prices columns. @sauloperez What's your opinion on this?

Another way would be to use a subquery:
```sql
SELECT spree_variants.id FROM spree_variants
WHERE spree_variants.id NOT IN (SELECT spree_prices.id FROM spree_prices)
```
But I'm not sure what's more efficient. I believe that a subquery would be really slow in Mysql because it doesn't index that sub-query result. Not sure about Postgres.

sauloperez: Yep, good catch @mkllnk . We need to filter out variants that have price with that `WHERE spree_prices.id IS NULL`. @Matt-Yorkley https://blog.codinghorror.com/a-visual-explanation-of-sql-joins/ is your friend.

db/migrate/20190830134723_add_deleted_at_to_spree_prices.rb
kristinalim: > I think we've discussed this before but indexes on boolean columns add little to no value and make writes slower. I wouldn't add it.

I didn't know. This is useful information, thanks. @sauloperez

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
luisramos0: I think this is the better SQL:
```
variants_without_prices = 'SELECT spree_variants.id FROM spree_variants
      LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
      WHERE spree_prices.id IS NULL
```

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    broken_line_items = Spree::LineItem.
+      joins(:variant).where(variant: { id: variants_without_prices }).
```
kristinalim: This migration is failing - It currently compares `variant.id` against the SQL subquery string above. I don't actually know how to use subqueries with the `where` method yet.

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
kristinalim: I like the suggestion of @luisramos0. :+1:

luisramos0:
Hey, looking at this one again, why dont we "add deleted prices to deleted variants without prices"?

I think this is harmless:
```sql
insert into spree_prices (variant_id, amount, currency, deleted_at)
SELECT spree_variants.id, '0.00', (select value from spree_preferences where key = 'spree/app_configuration/currency'), now()
FROM spree_variants
LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
WHERE spree_prices.id IS NULL;
```

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
sauloperez: credits to you @mkllnk :trollface:

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
sauloperez: Might be a Rails specific version :thinking: . But I tested in a Rails console and it correctly performed a single query using the subquery above.

sauloperez:
That would keep the line items which is always better

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
Matt-Yorkley: These queries seem to give exactly the same results:
```ruby
SELECT spree_variants.id FROM spree_variants LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id) 
WHERE NOT EXISTS (SELECT 1 FROM spree_prices WHERE spree_prices.variant_id = spree_variants.id)
```

```ruby
SELECT spree_variants.id FROM spree_variants LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id) 
WHERE spree_prices.id IS NULL
```

Matt-Yorkley:
>  why dont we "add deleted prices to deleted variants without prices"?

What happens when someone orders that product for `0.00`?

luisramos0:
>>>  why dont we "add deleted prices to deleted variants without prices"?

> What happens when someone orders that product for `0.00`?

The variant is deleted so no one will buy it. Right?

db/migrate/20190906165501_remove_broken_variants_from_carts.rb
```diff
@@ -0,0 +1,17 @@
+class RemoveBrokenVariantsFromCarts < ActiveRecord::Migration
+  def up
+    # Removes line_items from open carts where the variant has a hard-deleted price
+
+    variants_without_prices = execute(
+      'SELECT spree_variants.id FROM spree_variants
+      LEFT OUTER JOIN spree_prices ON (spree_variants.id = spree_prices.variant_id)
+      WHERE spree_prices.id IS NULL'
+    ).to_a.map{ |v| v['id'] }
```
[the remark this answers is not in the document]
sauloperez: let me know next time because there must be a way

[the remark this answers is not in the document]
Matt-Yorkley: If you can provide an example that works, that would be awesome. :+1:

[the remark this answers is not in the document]
sauloperez: I did a quick check on a Rails console with some dummy AR but I'll keep it in mind next time

RachL:
Here is the test I've carried out on French staging:

1. I've added Fuji Apple on Mary's shop to my cart with my protonmail email address
2. I've deleted Fuji Apple as super admin
3. Then tried to checkout my cart with my personal account and got this:

image


I still could navigated within the site with no problem :+1: 🎉

luisramos0:
there was a missing db:migrate in this PR to update db/schema.rb, I run it now in master and pushed it directly to master:
https://github.com/openfoodfoundation/openfoodnetwork/commit/2683efdd3c8124cc5607ba6e5653a312c0bcb36d

Issue #4003, opened 2019-07-04
Orders stuck in 'Payment' state - what has happened to these customers?


kirstenalarsen:
## Description
While investigating #3925 I was exploring 'uncompleted orders' and particularly those that are stuck in 'Payment' state. This could be related to case 
c) Customer's payment failed but customer was not notified (so order was presumed successful)

@mkllnk commented in Slack that payment state means that the payment has failed e.g. invalid credit card, but orders should be put back into cart state .. so if this is not happening, what is happening to these users / orders - are they just disappearing?
[image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/80c5d6d4-2b49-45e1-b7d9-e4daab43fd32)

I am sad to report that there are quite a few - on Aus production 175 in total. While some of these go back a long way, some of them are more recent and in current active hubs. From a quick look through some, it seems that all are connected to credit cards. Recent ones Stripe, older ones paypal.

When I follow through into the order and then into the payment I see that it is in state 'checkout'
![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/b0c7064d-fff9-4007-8ebe-5ac4eee24359)

## Expected Behavior
- We shouldn't have orders stuck in Payment state - if the Payment fails the Order should be in Cart and the payment showing as failed
- If a Customer's credit card fails, they should be notified
- [Then what - I'm not sure! TBD] Notify customers to contact Hub with updated credit card details? Update their details? 

## Actual Behavior
I haven't yet managed to replicate this - will work on it. Currently hypothesis that it is a failed credit card

## Steps to Reproduce
To see the problem: 
1. go to orders, untick 'show only completed orders' and select State 'Payment' . . 
working on it
2. and/or go to a recent order cycle and just see all of them 

Replication not yet managed
1.
2.
3.
4.

## Context
It has scared me. Trying to work out problems with duplicating and missing payments and finding all these floating around 

There is also #4004 which I'm still trying to work out if it is a problem

## Severity
I think this will become either S1 or S2 . . putting S2 until we're clearer on what it is

Taking credit card payments and getting customer orders is a critical feature. Notifying customers to 
bug-s1: a critical feature is broken: checkout, payments, signup, login
bug-s2: a non-critical feature is broken, no workaround
bug-s3: a feature is broken but there is a workaround

## Your Environment
MacOS

## Possible Fix
Handling failed credit card payments?

[...]

PR #4252, opened 2019-09-09
On checkout of pickup orders, set ship_address to shipping_address_from_distributor instead of empty Spree::Address.default

### LINKED ISSUE #4005: StateMachine::InvalidTransition in checkout#update
## Error in OpenFoodNetwork

**StateMachine::InvalidTransition** in **checkout#update**
Cannot transition state via :restart_checkout from :payment (Reason(s): State cannot transition via "restart checkout")

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5d1948bc850871001b662ad5?event_id=5d1948bc00469dfd87950000&i=gh&m=ci)

## Stacktrace

    app/services/restart_checkout.rb:22 - reset_state_to_cart
    app/services/restart_checkout.rb:10 - call
    app/controllers/checkout_controller.rb:142 - update_failed
    app/controllers/checkout_controller.rb:40 - update

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5d1948bc850871001b662ad5?event_id=5d1948bc00469dfd87950000&i=gh&m=ci)

luisramos0:
#### What? Why?

Closes #4005 and probably #4003 

In CheckoutController#update, instead of clearing ship address by setting it to empty Spree::Address.default, we set shipping_address_from_distributor which is what's done already when order is finalized.

#4005 was happening because an **empty** ship address (set in clear_ship_address) makes restart_checkout fail. restart_checkout_spec was not detecting this because it was using a **nil** ship_address.
We now set the distributor ship address on ship_address and also test that scenario in both restart_checkout_spec and checkout_controller.

#### What should we test?
We need to test checkout manually. The scenario that changed was when checking out using a pick up ship method without ship address and using a broken bogus credit card, the user was getting a sail or just not seeing any errors, now the user should see an flash error: credit card payment failed.

Added by Maikel:
Verify every way the checkout can fail:
- Out of stock
- Payment method not set up correctly
- Invalid credit card
Test these with a shipping method that needs an address and without. Then try to checkout again and make sure that the right shipping address is either preserved or blank and needs filling in.

Extra scenario added by Luis:
On checkout:
- ship method is pickup, no ship address form
- checkout with failed payment and see payment error
- go back to ship methods and change to some delivery ship method
- make sure the ship address shown by default is not the distributor's address

#### Release notes
Changelog Category: Fixed
Fixed problem in user checkout when payment validation failed. User is now being informed of what is wrong in the checkout form.

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -152,6 +152,7 @@ class CheckoutController < Spree::CheckoutController
   end
 
   def update_failed
+    current_order.updater.shipping_address_from_distributor
     RestartCheckout.new(@order).call
 
     respond_to do |format|
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -152,7 +152,6 @@ class CheckoutController < Spree::CheckoutController
   end
 
   def update_failed
-    clear_ship_address
     RestartCheckout.new(@order).call
 
     respond_to do |format|
@@ -162,15 +161,6 @@ class CheckoutController < Spree::CheckoutController
       format.json do
         render json: { errors: @order.errors, flash: flash.to_hash }.to_json, status: :bad_request
       end
-    end
-  end
-
-  # When we have a pickup Shipping Method,
-  #   we clone the distributor address into ship_address before_save
-  # We don't want this data in the form, so we clear it out
-  def clear_ship_address
-    unless current_order.shipping_method.andand.require_ship_address
-      current_order.ship_address = Spree::Address.default
     end
   end
 
--- a/spec/controllers/checkout_controller_spec.rb
+++ b/spec/controllers/checkout_controller_spec.rb
@@ -78,28 +78,17 @@ describe CheckoutController, type: :controller do
       allow(controller).to receive(:current_order).and_return(order)
     end
 
-    it "does not clone the ship address from distributor when shipping method requires address" do
-      get :edit
-      expect(assigns[:order].ship_address.address1).to be_nil
-    end
-
-    it "clears the ship address when re-rendering edit" do
-      expect(controller).to receive(:clear_ship_address).and_return true
+    it "set shipping_address_from_distributor when re-rendering edit" do
+      expect(order.updater).to receive(:shipping_address_from_distributor)
       allow(order).to receive(:update_attributes).and_return false
       spree_post :update, format: :json, order: {}
     end
 
-    it "clears the ship address when the order state cannot be advanced" do
-      expect(controller).to receive(:clear_ship_address).and_return true
+    it "set shipping_address_from_distributor when the order state cannot be advanced" do
+      expect(order.updater).to receive(:shipping_address_from_distributor)
       allow(order).to receive(:update_attributes).and_return true
       allow(order).to receive(:next).and_return false
       spree_post :update, format: :json, order: {}
-    end
-
-    it "only clears the ship address with a pickup shipping method" do
-      allow(order).to receive_message_chain(:shipping_method, :andand, :require_ship_address).and_return false
-      expect(order).to receive(:ship_address=)
-      controller.send(:clear_ship_address)
     end
 
     context "#update with shipping_method_id" do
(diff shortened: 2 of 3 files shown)


app/controllers/checkout_controller.rb
```diff
@@ -165,15 +165,6 @@ def update_failed
     end
   end
 
-  # When we have a pickup Shipping Method,
-  #   we clone the distributor address into ship_address before_save
-  # We don't want this data in the form, so we clear it out
```
sauloperez: so we'll miss this behavior, right? users will see the address of the distributor which on the other hand seems good.

luisramos0:
thanks @mkllnk for the review and test scenarios!

--- I'm not clear on a case though: What happens when the shipping method requires an address? Do we set the distributor's address and the user won't see it during checkout because it's collapsed (done)?

When the shipping method requires an address, OrderUpdater#shipping_address_from_distributor does nothing so the user needs to input the address.
```
  def shipping_address_from_distributor
    return if order.shipping_method.blank? || order.shipping_method.require_ship_address
```

app/controllers/checkout_controller.rb
luisramos0: the behaviour is the same but instead of an empty address we inject the distributor address (instead of this clear_ship_address, we are calling OrderUpdater#shipping_address_from_distributor).

When the ship method is pick up, the ship address form is not shown on checkout, so the user will not see it.

One test scenario I did here (will move it to the description so testers can try):
- ship method is pickup, no ship address form 
- checkout with failed payment and see payment error
- go back to ship methods and change to some delivery ship method
- make sure the ship address shown by default is not the distributor's address

luisramos0:
I have updated the test scenarios in the description.

app/controllers/checkout_controller.rb
mkllnk: I was wondering about this case:

- ship method requires address
- checkout with failed payment
- what is the ship address now?

luisramos0: yes, I tested that. the ship address is not overritten, it will be whatever the user wrote  because the shipping_address_from_distributor will not update ship address if order.shipping_method.require_ship_address

RachL:
@luisramos0 my first tests results:

A. Checking out with an invalid Bogus credit card with Pick-up shipping method (no shipping address form) results in no errors for the user (error 400 in dev tools)
image


B. Trying to add stripe as a payment method on a shop and I get an error 500:

image

Could it be that the PR was incorrectly staged?
Notes to myself:
https://docs.google.com/document/d/1tcwWTCUaS_ZCzJp3bN4KB7AXOcFOCukcONGZ_q_VOpI/edit#

luisramos0:
the PR is correctly staged.

case A - the problem is credit card date. For some reason, if the card is expired (a specific problem) the flash message is not set. I remember seeing this while in dev but could not fix it quickly. As this is an S2, I think we should open a separate issue for this one.
I tried to checkout with same invalid card number and code but with a date after current date, and the error is displayed.

case B - is stripe setup in uk staging?

luisramos0:
case A - now I see you used Jan 2022, I will retest :-/

luisramos0:
case A - I cant replicate, unless I put an expired date.... can you try to replicate and if it happens click on the Response tab in the 400 error?

luisramos0:
case B - cant find the snail, can you add specific steps to get to the snail? Thanks!

RachL:
@luisramos0

Case B: I forgot yes stripe is not working in UK staging... :facepalm:
Will stage on to FR.

About A. I have the following message:
image

Yet I do have "pick-up as the only shipping method".... will start again on FR and keep you posted

sauloperez:
@RachL would you mind creating an issue to add/fix Stripe to UK staging?

sigmundpetersen:
https://github.com/openfoodfoundation/ofn-install/issues/283

RachL:
This is looking good on staging FR!

https://docs.google.com/document/d/1tcwWTCUaS_ZCzJp3bN4KB7AXOcFOCukcONGZ_q_VOpI/edit#

luisramos0:
🎉

PR #4260, opened 2019-09-14
Add default date range to Delivery Report

### LINKED ISSUE #4184: Add a default date range for Delivery Report
## Description

Since many shops and hubs have been trading for a significant time frame now the delivery report is large and is not downloading- the web page is timing out.  Unlike some of the other reports there is no default date range (say of a week or month) set for this report. This would prevent the need to download all the data for a hub.

## Expected Behavior

Delivery Report should download data for the last month or week only.

## Actual Behaviour

The default setting for downloading a delivery report is to access all the data generated by hub sales since it set up on OFN.

## Steps to Reproduce

1. Reports -> order cycle management -> delivery report
2. No default date range selected.
3.
4.

## Animated Gif/Screenshot

image

## Context

The delivery report is utilized weekly by many hubs to get orders packed and shipped to customers who purchased items.  Often it fails to download due to the excessive amount of data it is attempting to download at once.

## Severity

s-3 there is a workaround (payment methods report and then change to download delivery report) but it is annoying and large hubs are finding it difficult to operate without this function working as it should do.

## Your Environment

* Version used:
* Browser name and version:

jasonk357:
**What? Why?**

Closes #4184 

This adds a default date range, the most recent month, to the parameters sent to the `OrderCycleManagementReport`.

I think putting the work here in a background job would make more sense but given the context maybe that isn't necessary?

**What should we test?**

Delivery report, payment methods report, with and without dates.

**Release notes**
Assigns a default date range from one month ago to today for the delivery report and payment methods report.

Changelog Category: Fixed

Woo first OFN PR!

### DIFF
--- a/lib/open_food_network/order_cycle_management_report.rb
+++ b/lib/open_food_network/order_cycle_management_report.rb
@@ -2,9 +2,10 @@ require 'open_food_network/user_balance_calculator'
 
 module OpenFoodNetwork
   class OrderCycleManagementReport
+    DEFAULT_DATE_INTERVAL = 1.month
     attr_reader :params
     def initialize(user, params = {}, render_table = false)
-      @params = params
+      @params = sanitize_params(params)
       @user = user
       @render_table = render_table
     end
@@ -132,6 +133,13 @@ module OpenFoodNetwork
     def customer_code(email)
       customer = Customer.where(email: email).first
       customer.nil? ? "" : customer.code
+    end
+
+    def sanitize_params(params)
+      params[:q] ||= {}
+      params[:q][:completed_at_gt] ||= Time.zone.today - DEFAULT_DATE_INTERVAL
+      params[:q][:completed_at_lt] ||= Time.zone.today
+      params
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lin-d-hop:
I don't think sending via email is the solution.
I do this Rachel's idea - of having a task manager meaning you can click away and the little manager will provide you with a link from the menu when it's done - is a good idea.
It's explained better somewhere else....

filipefurtad0:
Hi @jasonk357, welcome :-)

The default date and time appears now, for both report types - thank you.

The "now" button on the calendar provides the date/time for the local system in use and not the actual hub, which might lead to some discrepancies when these are not on the same timezone. I am wondering whether the "now" button should provide the timestamp of the hub instead?

You can find details on
https://docs.google.com/document/d/1Fq7Dtc_ESzHTGnewJhfqlRx4M6r4ISMRb2mCHCE7fNM/edit#

filipefurtad0:
Hi @sigmundpetersen :slightly_smiling_face: I see you've staged this on FR previously. I've tested it and it seems ok, should I move it to Ready to Go for further review, or are you still checking something?

sigmundpetersen:
Hey @filipefurtad0 I just added the label after you staged it 😉 Please move to ready if you are happy with the test results 👍

sauloperez:
Congratulations @jasonk357 and thanks for this contribution! :tada:

PR #4261, opened 2019-09-14
Guard against malformed request referer

### LINKED ISSUE #3979: Homepage crashes with malformed `referer` from browser
## Description

Most browser tell our servers from which page they come through a field called `referer`. Yes, there is an `r` missing in the name. Anyway, browsers can be configured to disable the referer or set it to a fixed value. People scanning the site may also set an arbitrary referer.

One of our code parts is crashing when the referer is not valid. Here is the error report from Bugsnag:

```
NoMethodError
home#index
undefined method `downcase' for nil:NilClass
Jun 27th, 2019, 00:05:44 UTC

STACKTRACE

app/services/embedded_page_service.rb:76:in `current_referer': undefined method `downcase' for nil:NilClass (NoMethodError)
    from app/services/embedded_page_service.rb:30:in `embeddable?'
    from app/services/embedded_page_service.rb:15:in `embed!'
    from app/controllers/application_controller.rb:59:in `enable_embedded_shopfront'
    ...
```
https://app.bugsnag.com/open-food-france/open-food-france-prod/errors/5d14085234d4f3001965e011

## Expected Behavior

The application doesn't crash, no matter what weird data is sent to it. The test command below shows:
```
<title>Welcome to  Open Food Network</title>
```

## Actual Behaviour

The homepage is a snail page. The test command below shows:
```
  <title>We're sorry, but something went wrong (500)</title>
```

## Steps to Reproduce

1. Shell command:
    ```sh
    curl -s -e "hello" https://staging.openfoodnetwork.org.au | grep '<title>'
    ```

The command is testing one of our staging servers. If you test locally, the URL would be `http://0.0.0.0:3000/`.

## Context

Reported by Bugsnag.

## Severity

## Your Environment

* Version used: v1.31.0

## Possible Fix

It could be this simple change in `app/services/embedded_page_service.rb:76`:

```diff
- URI(@request.referer).host.downcase
+ URI(@request.referer).host.andand.downcase
```

But we should also test what happens if `referer` is an invalid URL. These cases should be tested in `spec/services/embedded_page_service_spec.rb`.

jasonk357:
**What? Why?**

Closes #3979 
When a request comes in the app attempts to parse the referer URI. This was causing an error when the referer field is malformed because there is no check that it was successful before calling `#downcase` on the string. 

**What should we test?**
Requests with malformed referer strings. 
`curl -s -e "hello" http://localhost:3000/ | grep '<title>'`
I also added a test case for this.

**Release notes**
Added a guard for malformed request referer strings.

Changelog Category: Fixed

### DIFF
--- a/app/services/embedded_page_service.rb
+++ b/app/services/embedded_page_service.rb
@@ -73,7 +73,9 @@ class EmbeddedPageService
 
   def current_referer
     return if @request.referer.blank?
-    URI(@request.referer).host.downcase
+    uri = URI(@request.referer)
+    return if uri.host.blank?
+    uri.host.downcase
   end
 
   def current_referer_without_www
--- a/spec/services/embedded_page_service_spec.rb
+++ b/spec/services/embedded_page_service_spec.rb
@@ -59,5 +59,16 @@ describe EmbeddedPageService do
         expect(response.headers['X-Frame-Options']).to eq 'DENY'
       end
     end
+
+    context "when the request's referer is malformed" do
+      let(:request) { ActionController::TestRequest.new('HTTP_HOST' => 'ofn-instance.com', 'HTTP_REFERER' => 'hello')}
+      before do
+        service.embed!
+      end
+
+      it "returns a 200 status" do
+        expect(response.status).to eq 200
+      end
+    end
   end
 end


daniellemoorhead:
> What should we test?
> Requests with malformed referer strings.
> curl -s -e "hello" http://localhost:3000/ | grep '<title>'
> I also added a test case for this.

Um, sorry but no idea how to test this...reading those instructions was like me trying to read something in Russian...help please! Ping @sauloperez

RachL:
image

Issue #4166, opened 2019-08-19
Create Shipping Method should auto-select Default Shipping Category


kirstenalarsen:
## Description
Similar to #4165, V2 has introduced requirement of Shipping Category in creation of a Shipping Method. 

This is meaningless to users, as are the Shipping Categories, so could we please just select Default by default so they don't have to think about it?

## Expected Behavior
Shipping Category automatically set to Default

## Actual Behaviour
Shipping method fails to save because Customer hasn't selected Shipping Method. There is an error message

## Steps to Reproduce

1. Create Shipping Method
2. Don't select Shipping Category
3. Try to Save
4.

![image](https://user-images.githubusercontent.com/2911176/63240897-0b7a7080-c295-11e9-86b5-d6c8915431d6.png)

## Context
Just annoying

## Severity

bug-s4: it's annoying, but you can use it

luisramos0:
this has been discussed somewhere else, I cant remember where. @RachL and @lin-d-hop were both involved, I cant remember the decision of that discussion.

RachL:
I don't think this is a bug, more a UX discussion. If indeed none of our users care about this we should have a selection by default. I'm labeling it "improvement".

So far, no users have send feedback in France on this. Let's use this issue to track feedback we might receive and make a decision later on. I'm not sure we have enough info to do it currently.

lin-d-hop:
I think simply having the compulsory field - red star will be enough. I think it would be good to begin to encourage users to really use the shipping category.

luisramos0:
what do you guys think about making both ship method creation and product creation pages look up the "Default" shipping category and select it by default?
From my code review here: https://github.com/openfoodfoundation/openfoodnetwork/pull/4262#pullrequestreview-288542802

RachL:
@luisramos0 yes I'm ok with that.

I'm sorry @lin-d-hop but when dan asked about it I've only read the issue description and not the comments below, so I've told him that he could go with having "default" selected by default.
And I agree with you Luis, the idea is to select "default" not the first shipping category which is available.

kirstenalarsen:
This seems to be done! yippee, closing :)

PR #4262, opened 2019-09-15
Prefill shipping category forms with default values

### LINKED ISSUE #4165: Selection of Shipping Category should be identified as required with a red asterisk in product creation
## Description

With V2 the User is now required to select a Shipping Category when creating a product. However the field is not highlighted with a red asterisk so there is no way for them to know that. This is a meaningless field to them so they are unlikely to do it without being told

If they have not selected a Shipping Category, then they see this message below and the Unit Size and Value fields are cleared. So they then select a meaningless Shipping Category field (say Default) and get error message that they don't have units

 

ARGHGGHG - could we make life more painful for a new user if we tried ;) ?

## Expected Behavior

The Shipping Category field should EITHER
- have a red asterisk to denote the need to select a value OR
- just default to Default so they don't have to think about it

## Actual Behavior
as above

## Steps to Reproduce

1. Create new product
2. Fill out all required fields but DON'T select a Shipping Category
3. Voila

## Animated Gif/Screenshot
http://recordit.co/z4DUokFeGH

## Severity

daningenthron:
#### What? Why?

Closes #4165

Shipping category is not marked as required on the Create Product form. Current configuration allows creation of only one shipping category. This prefills the form field.

#### What should we test?

`/spec/features/admin/products_spec.rb` has been updated to check for the prefilled value instead of adding a value when populating the form.

#### Release notes

This exposed a couple potential issues (possibly only in development environment?):
- Should a user be able to have more than one shipping category saved? There is only an option to edit the default category, not add a second one.
- Attempting to delete the existing shipping category causes an ActiveRecord error.

Changelog Category: Fixed


luisramos0:
if we go ahead with selecting default, it should be encapsulated in a service like DefaultStockLocation, see [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/services/default_stock_location.rb)

This will evolve the approach in existing migration AddShippingCategoryToShippingMethodsAndProducts.

Something like this?

```
# Encapsulates the concept of default shipping category
class DefaultShippingCategory
  NAME = 'Default'.freeze

  def self.create!
    Spree::ShippingCategory.create!(name: NAME)
  end

  def self.find_or_create
    Spree::ShippingCategory.find_or_create_by_name(NAME)
  end
end
```

and then use DefaultShippingCategory.find_or_create in the views

daningenthron:
Hi @luisramos0 thanks for clarifying the need for the default! This makes sense looking at #4166 and I agree that it would be better to address both problems with the same fix. I'll add the default service and tie it into both pages.

daningenthron:
I believe this sets up the service for use in product and shipping method creation. I adjusted the specs for these pages to use the Defaults as well, and also added it to the seeds file. Let me know what you think!

app/services/default_shipping_category.rb
```diff
@@ -0,0 +1,13 @@
+# Encapsulates the concept of default stock location in creation of a product or a shipping method.
+
```
luisramos0: this empty line can be removed (I wonder if rubocop is not flagging it)

app/services/default_shipping_category.rb
```diff
@@ -0,0 +1,13 @@
...
+class DefaultShippingCategory
+  NAME = 'Default'.freeze
+
+  def self.create!
+	  Spree::ShippingCategory.create!(name: NAME)
```
luisramos0: indentation.

spec/services/default_shipping_category_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+
+describe DefaultShippingCategory do
+	describe '.create!' do
```
luisramos0: indentation needs to be two spaces, not a tab

db/seeds.rb
```diff
@@ -52,3 +52,4 @@ def set_mail_configuration
 require File.join(File.dirname(__FILE__), 'default', 'users')
 
 DefaultStockLocation.find_or_create
+DefaultShippingCategory.find_or_create
```
luisramos0: there's a new line missing here

npm-debug.log
```diff
@@ -0,0 +1,27 @@
...
+4 verbose stack     at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:401:12)
+5 verbose cwd /Users/ingenthron/dev/Ovio/openfoodnetwork
+6 error Darwin 17.7.0
+7 error argv "/Users/ingenthron/.nodenv/versions/5.12.0/bin/node" "/Users/ingenthron/.nodenv/versions/5.12.0/bin/npm" "run" "build"
+8 error node v5.12.0
+9 error npm  v3.8.6
+10 error missing script: build
+11 error If you need help, you may report this error at:
+11 error     <https://github.com/npm/npm/issues>
+12 verbose exit [ 1, true ]
```
luisramos0: this file must be removed

daningenthron:
I think this failed on Semaphore bc of the Ruby version; rebased and resubmitted!

luisramos0:
60 commits... something weird with the rebase.
you can do:
git rebase upstream/master

luisramos0:
ouch :-O I was trying to fix the rebase problem for you and I just overwrote your branch with... nothing. 
To fix this and get back to where this was, you need to force PUSH your local copy to your remote  branch... I am sorry about this 🤕

spec/features/admin/shipping_methods_spec.rb
```diff
@@ -88,6 +89,7 @@
       within(".side_menu") do
         click_link "Shipping Methods"
       end
+      DefaultShippingCategory.find_or_create
```
Matt-Yorkley: Is this line needed?

spec/services/default_shipping_category_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'spec_helper'
+
+describe DefaultShippingCategory do
+  describe '.create!' do
+    it "names the location 'Default'" do
+      shipping_category = described_class.create!
+      expect(shipping_category.name).to eq('Default')
```
Matt-Yorkley: It's not a big deal, but there's lots of unnecessary parentheses in this file. You can use `eq 'Default'` here, for example. I'm surprised`rubocop` is not complaining. I'm not sure what IDE or editor you're using, but many have `rubocop` plugins that add tips for code style, and automated linting.

spec/features/admin/shipping_methods_spec.rb
daningenthron: Hi @Matt-Yorkley thanks for taking a look. This line was a hacky way to add a second default category to the one already provided by the Spree factory. I've added a new commit that overrides the Spree factory and cleans up the spec.

spec/services/default_shipping_category_spec.rb
daningenthron: I have the Rubocop linter added to Sublime 3 - it wasn't catching these, but I'll check the style guide to make sure that my settings are correct. Thanks!

spec/services/default_shipping_category_spec.rb
```diff
@@ -4,18 +4,18 @@
...
       shipping_category = described_class.create!
-      expect(shipping_category.name).to eq('Default')
+      expect(shipping_category.name).to eq 'Default'
     end
   end
 
   describe 'find_or_create' do
     context 'when a Default category already exists' do
-      let!(:category) do
+      let! :category do
```
luisramos0: you need to keep the parenthesis here to keep consistency with all other specs.
in OFN code, not using parenthesis is almost only used in specs after eq

spec/factories.rb
```diff
@@ -194,4 +194,9 @@
     # sets the default value for variant.on_demand
     backorderable_default false
   end
+
+  factory :shipping_category, class: Spree::ShippingCategory do
+    initialize_with { DefaultShippingCategory.find_or_create }
+    transient { name 'Default' }
+  end
```
luisramos0: :+1:

sigmundpetersen:
Semaphore build has 1 error

app/views/spree/admin/products/_shipping_category_form.html.haml
```diff
@@ -1,4 +1,4 @@
 = f.field_container :shipping_categories do 
   = f.label :shipping_category_id, t(:shipping_category)
-  = f.collection_select(:shipping_category_id, Spree::ShippingCategory.all, :id, :name, {:include_blank => true}, {:class => 'select2 fullwidth'})
+  = f.collection_select(:shipping_category_id, Spree::ShippingCategory.all, :id, :name, {:include_blank => false, :selected => DefaultShippingCategory.find_or_create.id}, {:class => 'select2 fullwidth'})
```
kristinalim: This selects "Default" no matter what, which would be a problem in this scenario:

1. In the new product form, fill in required product information except product name.
2. Select a non-default shipping category.
3. Submit the form.
4. In the product form showing the errors, the "Default" shipping category is selected even when it should be a different one.

How about setting the default category for the object in `spree/admin/products#new`? (`app/controllers/spree/admin/products_controller_decorator.rb`). Something like:

```ruby
def new
  @object.shipping_category = DefaultShippingCategory.find_or_create
  super
end
```

Then, this wouldn't be called after the user's unsuccessful submission of the form (`#create`).

app/views/spree/admin/shipping_methods/_form.html.haml
```diff
@@ -53,7 +53,7 @@
       = f.field_container :categories do
         - Spree::ShippingCategory.all.each do |category|
           = label_tag do
-            = check_box_tag('shipping_method[shipping_categories][]', category.id, @shipping_method.shipping_categories.include?(category))
+            = check_box_tag('shipping_method[shipping_categories][]', category.id, category == DefaultShippingCategory.find_or_create)
```
kristinalim: Same issue here - the "Default" checkbox is the only one selected when you load the edit shipping method page, even when it has been associated to all available shipping methods.

app/services/default_shipping_category.rb
```diff
@@ -0,0 +1,13 @@
+# Encapsulates the concept of default stock location in creation of a product or a shipping method.
+
+class DefaultShippingCategory
+  NAME = 'Default'.freeze
```
kristinalim: OFN is deployed in different places around the world, and translations are quite important. For example, there might be some languages where users might not understand what "Default" means.

I checked all [managed instances](https://github.com/openfoodfoundation/ofn-install/wiki/Current-deployment-status). and all still have the "Default" (English) shipping category. But if for example OFN France updates their "Default" to "Défaut", the "Default" category would automatically be created again so they would end up with 2.

This is not an immediate issue that should be fixed, I think. I'll create a separate issue for this.

app/views/spree/admin/products/_shipping_category_form.html.haml
daningenthron: Hi @kristinalim - thanks, this definitely makes sense. I had been thinking of a front-end approach with just the forms, instead of the object itself. I see the behavior you're describing in both forms upon submission. This seems to be working now, but I'll retest everything before submitting the PR again.

Regarding the question of the shipping method spec above (adding a second shipping category), I believe this has been resolved with the changes that I made to the shipping category factory.

app/views/spree/admin/shipping_methods/_form.html.haml
daningenthron: Thanks again for pointing this out! Since the shipping method accepts multiple categories, I used this variation to instantiate the class:

`def new
        @object.shipping_categories = [DefaultShippingCategory.find_or_create]
        super
end`

app/services/default_shipping_category.rb
daningenthron: This is a good point; I'll keep in mind that any user-facing text should be set up with a key in en.yml so Transifex can build out the translation. If you are setting up an issue for this, I modeled `default_shipping_category.rb` after `default_stock_location.rb`, which will need the same fix.

daningenthron:
@kristinalim Sorry for the delay! Was traveling and wanted to check this against the new controller before pushing. Let me know what you think - thanks!

app/services/default_shipping_category.rb
kristinalim: Thanks, @daningenthron!

I created issue #4360 for making the "Default" shipping category translatable.

I didn't create an issue for stock location anymore because we only support one stock location, so this is mostly used internally. :smiley:

JenRSmith:
Tested in UK and successfully showing default shipping method without requiring user input, and also retaining user selected shipping method rather than overriding with default. 

Testing notes here https://docs.google.com/document/d/1g2yCXRRgKhp3sst08j-qk4LuJ5W6GTgQw34J29rzFFw/edit#

Moved to Ready to Go

PR #4263, opened 2019-09-16
Index spree_orders on various columns

sauloperez:
#### What? Why?

**TL;DR We move from 56.643ms to 0.077ms in the 2nd most executed DB query** 

The following query

```sql
SELECT spree_orders . *
FROM spree_orders
WHERE spree_orders . user_id = ?
AND spree_orders . completed_at IS ?
AND spree_orders . created_by_id = ?
ORDER BY created_at DESC LIMIT ?
```

performs quite badly even though LIMIT is always 1 because:

* ORDER BY requires sorting by a column which is not indexed therefore
a sequential scan is performed.
* Although `completed_at` is indexed, `user_id` and `created_by_id` are
not causing a sequential scan.

To make it worse this query is executed very often in the following controllers among others also related to checkout:

* CartController#populate
* EnterprisesController#shop
* LineItemsController#bought
* ShopController#products
* ShopController#order_cycle

In some cases, this query alone accounts for [66.8% of the total time](https://app.datadoghq.com/apm/trace/2826003963900446028?colorBy=service&env=production&graphType=flamegraph&routing_hint=1568631644922&shouldShowLegend=true&sort=time&spanID=6392504358817399626) of the endpoint.

## Results

See by yourself. We move from 56.643ms to 0.077ms. Pretty neat.

```
openfoodnetwork=> explain analyze SELECT "spree_orders".* FROM "spree_orders" WHERE "spree_orders"."user_id" = 1 AND "spree_orders"."completed_at" IS NULL AND "spree_orders"."created_by_id" = 1 ORDER BY created_at DESC LIMIT 1;
                                                        QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=11753.03..11753.04 rows=1 width=195) (actual time=56.580..56.580 rows=0 loops=1)
   ->  Sort  (cost=11753.03..11753.04 rows=1 width=195) (actual time=56.578..56.578 rows=0 loops=1)
         Sort Key: created_at DESC
         Sort Method: quicksort  Memory: 25kB
         ->  Seq Scan on spree_orders  (cost=0.00..11753.02 rows=1 width=195) (actual time=56.571..56.571 rows=0 loops=1)
               Filter: ((completed_at IS NULL) AND (user_id = 1) AND (created_by_id = 1))
               Rows Removed by Filter: 256135
 Planning time: 0.252 ms
 Execution time: 56.643 ms
(9 rows)

openfoodnetwork=> CREATE INDEX ON spree_orders (completed_at, user_id, created_by_id, created_at);
CREATE INDEX
openfoodnetwork=> explain analyze SELECT "spree_orders".* FROM "spree_orders" WHERE "spree_orders"."user_id" = 1 AND "spree_orders"."completed_at" IS NULL AND "spree_orders"."created_by_id" = 1 ORDER BY created_at DESC LIMIT 1;
mit  (cost=8.45..8.46 rows=1 width=195) (actual time=0.030..0.030 rows=0 loops=1)
   ->  Sort  (cost=8.45..8.46 rows=1 width=195) (actual time=0.029..0.029 rows=0 loops=1)
         Sort Key: created_at DESC
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using spree_orders_completed_at_user_id_created_by_id_created_at_idx on spree_orders  (cost=0.42..8.44 rows=1 width=195) (actual time=0.021..0.021 rows=0 loops=1)
               Index Cond: ((completed_at IS NULL) AND (user_id = 1) AND (created_by_id = 1))
 Planning time: 0.199 ms
 Execution time: 0.077 ms
```

#### What should we test?

A dev should check that deploying correctly runs the migrations.

#### Release notes

Added missing index on spree_orders which speeds up the 2nd most executed DB query
Changelog Category: Added

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20190913023137) do
+ActiveRecord::Schema.define(:version => 20190916110029) do
 
   create_table "adjustment_metadata", :force => true do |t|
     t.integer "adjustment_id"
@@ -576,7 +576,7 @@ ActiveRecord::Schema.define(:version => 20190913023137) do
     t.integer  "created_by_id"
   end
 
-  add_index "spree_orders", ["completed_at"], :name => "index_spree_orders_on_completed_at"
+  add_index "spree_orders", ["completed_at", "user_id", "created_by_id", "created_at"], :name => "spree_orders_completed_at_user_id_created_by_id_created_at_idx"
   add_index "spree_orders", ["customer_id"], :name => "index_spree_orders_on_customer_id"
   add_index "spree_orders", ["number"], :name => "index_orders_on_number"
 
--- a/db/migrate/20190916105416_add_long_compound_index_on_spree_orders.rb
+++ b/db/migrate/20190916105416_add_long_compound_index_on_spree_orders.rb
@@ -1,0 +1,9 @@
+class AddLongCompoundIndexOnSpreeOrders < ActiveRecord::Migration
+  def change
+    add_index(
+      :spree_orders,
+      [:completed_at, :user_id, :created_by_id, :created_at],
+      name: 'spree_orders_completed_at_user_id_created_by_id_created_at_idx'
+    )
+  end
+end
--- a/db/migrate/20190916110029_drop_completed_at_index_on_spree_orders.rb
+++ b/db/migrate/20190916110029_drop_completed_at_index_on_spree_orders.rb
@@ -1,0 +1,5 @@
+class DropCompletedAtIndexOnSpreeOrders < ActiveRecord::Migration
+  def change
+    remove_index :spree_orders, :completed_at
+  end
+end


db/schema.rb
```diff
@@ -576,7 +576,7 @@
     t.integer  "created_by_id"
   end
 
-  add_index "spree_orders", ["completed_at"], :name => "index_spree_orders_on_completed_at"
```
sauloperez: This index is redundant because of PostgreSQL's left-most index columns usage.

As described in https://www.celerity.com/how-to-design-sql-indexes/:

> Additionally, multi-column indexes can still be used by the query planner even when all of the columns in the index aren’t constrained upon. For example, if you define a multi-column index against a user id, a start time, and an end time (in that order), that index can be used for queries against the user id alone, as well as the user id & start time together. The order of columns in the index definition matter: so long as your subset of columns follows the index definition left-to-right, that multi-column index is effectively and efficiently used.

sauloperez:
> I say this because this is a query specific index so I think it deserves a comment where the query is

It's actually used in many places. Take a look at https://app.datadoghq.com/apm/resource/rails-postgres/postgres.query/9b4df5714e78b76f?end=1568637056648&env=production&paused=false&start=1566045056648&index=apm-search

> If we change the query in the future, we should at least verify this index as well.

We do have the monitoring in place to watch for that so as long as we pay attention to Datadog we should be covered.

luisramos0:
I am ok with that but the calls listed are the controllers, there must be a common ground to these calls, some operation on a specific model or something.

sauloperez:
Oh yeah, definitely. I'll find out

sigmundpetersen:
Conflicts @sauloperez . Rebase?

sauloperez:
Alright, so this query is triggered by `#last_incomplete_spree_order` which is defined in spree in [core/config/initializers/user_class_extensions.rb](https://github.com/openfoodfoundation/spree/blob/master/core/config/initializers/user_class_extensions.rb) 

```ruby
def last_incomplete_spree_order
  spree_orders.incomplete.where(:created_by_id => self.id).order('created_at DESC').first
end
```

which in turn is used by OFN in https://github.com/openfoodfoundation/openfoodnetwork/blob/07ae51560e9063318fe388b428be6e3fd7d8869a/lib/spree/core/controller_helpers/order_decorator.rb#L20-L30

pretty much used everywhere.

luisramos0:
ah, yeah, thanks for clarifying that.
and that is a before_filter in the store_controller, basically, this is called in every frontoffice action 👍

sauloperez:
Please, stop touching the db/schema :trollface:

sauloperez:
I see now that for this case even a smoke test doesn't make sense. A DB can't affect any functionality.

This can only be deployed at most to check if the migrations work.

sauloperez:
Successfully deployed and migrated in fr-staging.

PR #4266, opened 2019-09-17
Enable Delayed Job analytics in Datadog

sauloperez:
#### What? Why?

Hopefully, this will provide extra information that will allow us to
better understand what are background workers spend their time on.

According to the docs

> Enable analytics for spans produced by this integration. true for on, nil to
> defer to global setting, false for off.

This is all the Delayed Job Datadog integration can do.

#### What should we test?

A dev should check the extra reported information in Datadog after deploying to production.

#### Release notes

Enable analysis of the spans produced by Datadog's Delayed Job integration

Changelog Category: Added

### DIFF
--- a/config/initializers/datadog.rb
+++ b/config/initializers/datadog.rb
@@ -1,6 +1,6 @@
 if ENV['DATADOG_RAILS_APM']
   Datadog.configure do |c|
     c.use :rails, service_name: 'rails'
-    c.use :delayed_job, service_name: 'delayed_job'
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Ah, I didn't see that in the docs. What does the analytics add?

sauloperez:
> What does the analytics add?

I only have that quote I shared in the description :trollface: . If adds any more information about the job itself it'll be welcome.

sauloperez:
Superseded by https://github.com/openfoodfoundation/openfoodnetwork/pull/4295

Issue #4072, opened 2019-07-24
[API clean up] Verify test coverage of API endpoints already in use


luisramos0:
### What we should change and why (this is tech debt)
We need to make sure the API endpoints currently being used by external systems like zapier are covered with specs so that we don't delete or break them accidentally.

Ping @lin-d-hop can please you comment on this issue with the endpoints you think need to verified?

### Context
Part of API clean up. First step of making our API great.

lin-d-hop:
Currently the only endpoint we are using is:
`/api/orders/${order_number}.json`

This returns:
https://gist.github.com/lin-d-hop/ad18e3568ad25069bd75ea15a32cd3ef

Obviously not all fields are used in the integration and this could be tidied quite a bit. However unless there is a good business case for tinkering this now perhaps it makes sense to leave as is until a more comprehensive API redesign?

lin-d-hop:
The other endpoint that I am in conversations with users about (though not yet using) is:
`/api/orders.json`
Particularly with the parameters:
- completed_at_gt
- completed_at_lt
- distributor_id

luisramos0:
Hi @lin-d-hop 
During my api clean up I mixed both entries above, I assumed /api/orders.json was needed and /api/orders/${order_number}.json could be needed soon.

In fact the /api/orders/${order_number}.json endpoint is coming from spree_api and disappears in https://github.com/openfoodfoundation/openfoodnetwork/pull/4057 :-(

I think we should move forward with this, with the current setup is quite easy to add a new endpoint to the API. and now we know where and how to do it :tada:

Can we open a new issue to re-add /api/orders/${order_number}.json after #4057 and go from there? For this issue we would need to know:
- What process is currently /api/orders/${order_number}.json?
- What fields are actually required from the huge payload referred above?
If the process is critical we should block #4057 from being merged until this new issue is also implemented.


btw, in terms of test coverage:
-  /api/orders/${order_number}.json this was being tested in spree but just basic validation that the endpoint is available and that it was implementing authorization (cannot see someone elses orders)
- /api/orders.json is tested in OFN

lin-d-hop:
Apologies for brevity....

 /api/orders/${order_number}.json is critical for our Zapier integration to accounting systems - used by a number of UK users (not sure about other places).

Thus #4057 needs to be blocked until this is fixed.

Will add to the iissue now.

RachL:
Outch. 

@luisramos0 

[...]

Issue #4146, opened 2019-08-12
Allow Enterprise Managers to add Shipping Notes in Edit Orders screen


lin-d-hop:
## Description

When Users are checking out they can leave a Shipping Note for the Enterprise  Manager:
![Screenshot from 2019-08-12 16-18-31](https://user-images.githubusercontent.com/8669848/62876448-e332c500-bd1c-11e9-8883-4a69ade35e10.png)

Back in v1 it was possible for the Enterprise Manager to update an Order with a Shipping Note. However in v2 this functionality has been lost and there is no workaround. 

## Expected Behavior

Hub Managers can add a Shipping Note to an order in the edit orders screen.

## Actual Behaviour

There is no longer an option to add a shipping note.

## Context

This has meant that a previously internal OFN feature is being done with scraps of paper for a significant UK food hub, adding mistakes to their process which can be costly.

## Your Environment

* Version used: v2.2.2
* OFN Platform instance: UK

## Possible Fix

With v2 a Tracking Number field has been added to the Orders screen:
![Screenshot from 2019-08-12 16-30-29](https://user-images.githubusercontent.com/8669848/62877274-9354fd80-bd1e-11e9-8021-ab4048458450.png)
![Screenshot from 2019-08-12 16-30-45](https://user-images.githubusercontent.com/8669848/62877280-96e88480-bd1e-11e9-97fc-e1ae7dc717d5.png)

An ideal solution would be to re-purpose this field to Shipping Notes and have the contents read and write to the same field that the customer can write to at Checkout.

luisramos0:
I dont think we need to repurpose an existing spree field to do something else. I think we have a better option.

The customer currently can write a shipping note on checkout, this is stored in a field called  order.special_instructions.
We can display this field in the order edit page (as "Shipping Notes") and make it editable by the manager. This will be stored in order.special_instructions. ok?

We can display this below tracking or replace tracking.

lin-d-hop:
Yep that sounds perfect. 
My intention on 're-purposing' the field was to keep the UI clean.
So replacing the 'tracking' field with this field is totally perfect :-)

lbwright22:
![](https://github.trello.services/images/mini-trello-icon.png) [Invoices & Pick Lists](https://trello.com/c/Vn1V1M9M/912-invoices-pick-lists)

RachL:
[...]

PR #4268, opened 2019-09-17
Add api/orders/{order_number} API endpoint

### LINKED ISSUE #4264: Re-add required endpoint /api/orders/{order_number}
## Description
In #4057 by removing spree_api we remove all api endpoints provided by spree.

We have one endpoint in this list that is a requirement for existing zapier integrations with accounting solutions.

## Expected Behavior
The endpoint is available with data about line_items and adjustments included. @lin-d-hop to add detailed field list at some point here.

## Actual Behaviour
The endpoint is gone if we merge #4057 so this is blocking 4057.

## Steps to Reproduce
Go to /api/orders/{order_number} and see a payload like [this example](https://gist.github.com/lin-d-hop/ad18e3568ad25069bd75ea15a32cd3ef).

## Context
Detected in #4072

## Severity
It's blocking 4057 which is part of the epic "API clean up".

## Possible Fix
We need to add a new endpoint in OFN using the OFN way of building API endpoints.

luisramos0:
#### What? Why?

Closes #4264 and #4072

Adds the new order detail endpoint by adding a new order detailed serializer that mostly uses existing serializers for the parts.

According to the issue, the spree endpoint was producing [this payload](https://gist.githubusercontent.com/lin-d-hop/ad18e3568ad25069bd75ea15a32cd3ef/raw/61f1cc63657b9a523467b3350d7ff345c240b7fc/order.json
). The new endpoint is producing something a little different (for the better I believe), see it [here](https://gist.githubusercontent.com/luisramos0/86805d7addc29f0dd90acb6b8ed2ef48/raw/c663300283ca39c52ce893e6a4aeb162e0a23ec9/order.json).
You can use this online tool for example to look at the data: http://jsonviewer.stack.hu/.

#### What should we test?
This endpoint is not being used by the app so app testing is not required.
All we need to verify is that the data in the json result includes all needed details for the current zapier integration.

(Added by @kristinalim) Testing for this PR will be done by @lin-d-hop, especially because we need the make sure this PR doesn't break existing integrations for OFN UK and their data. The following concerns need to be checked:

1. Will deployment of these changes not be destructive for those integrations and their data?
2. Do we need to specify and announce a deployment date for the release that includes this PR?

#### Release notes
Changelog Category: Added
Added API endpoint /api/orders/{order_number} that was removed with spree_api but is needed for current integrations. This endpoint is now implemented in OFN in the correct way.


app/serializers/api/address_serializer.rb
```diff
@@ -4,7 +4,11 @@ class Api::AddressSerializer < ActiveModel::Serializer
 
   attributes :id, :zipcode, :city, :state_name, :state_id,
              :phone, :firstname, :lastname, :address1, :address2, :city, :country_id,
-             :zipcode
+             :zipcode, :country_name
+
+  def country_name
+    object.country.andand.name
```
sauloperez: I don't think we are eager loading the country so this will issue an extra query to get it while a single one with a join would do. Please, check that out.

This with `has_many`  declarations in serializers is one of the most common causes of N+1s we have in the app.

app/controllers/api/orders_controller.rb
```diff
@@ -1,5 +1,10 @@
 module Api
   class OrdersController < BaseController
+    def show
+      authorize! :read, order
+      render json: @order, serializer: Api::OrderDetailedSerializer
```
sauloperez: I would just call it `Api::OrderSerializer` so it maps one to one with the model and we avoid some boilerplate IMO. Or do we have a class call like that already?

sauloperez: since you memoized the order below, you can use `order` instead of `@order` here.

app/serializers/api/order_detailed_serializer.rb
```diff
@@ -0,0 +1,12 @@
+module Api
+  class OrderDetailedSerializer < Api::Admin::OrderSerializer
+    has_one :shipping_method, serializer: Api::ShippingMethodSerializer
+    has_one :ship_address, serializer: Api::AddressSerializer
+    has_one :bill_address, serializer: Api::AddressSerializer
+
+    has_many :line_items, serializer: Api::LineItemSerializer
+    has_many :adjustments, serializer: Api::AdjustmentSerializer
```
sauloperez: ⚠️ see my comment above

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -156,6 +158,109 @@ module Api
...
+
+        it "when order number given is not in the systen" do
+          get :show, id: "X1321313232"
+          expect_resource_not_found
+        end
+
+        def expect_resource_not_found
+          expect(json_response).to eq( "error" => "The resource you were looking for could not be found." )
+          expect(response.status).to eq(404)
+        end
```
sauloperez: isn't [have_http_status](https://relishapp.com/rspec/rspec-rails/docs/matchers/have-http-status-matcher) enough? line 178 feels like testing Rails to me and that's already pretty well tested.

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -156,6 +158,109 @@ module Api
...
+      end
+
+      def expect_order
+        expect(response.status).to eq 200
+        expect_correct_order(json_response, order)
+      end
+
+      def expect_correct_order(json_response, order)
+        expect(json_response[:number]).to eq(order.number)
```
sauloperez: isn't the number enough? it's a unique way to identify it.

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -156,6 +158,109 @@ module Api
...
+        expect(json_response[:email]).to eq(order.email)
+      end
+
+      def expect_detailed_attributes_to_be_present(json_response)
+        expect(order_detailed_attributes.all?{ |attr| json_response.key? attr.to_s }).to eq(true)
```
sauloperez: I think either [include](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/include-matcher) or [have_attributes](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/have-attributes-matcher) RSpec matchers already implement this.

I usually check this with `expect(order.attributes).to include(expected_attributes_hash)` so less code of our own to maintain.

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -156,6 +158,109 @@ module Api
...
+
+        it "returns an order with all required fields" do
+          get :show, id: order.number
+          expect_order
+          expect_detailed_attributes_to_be_present(json_response)
+
+          expect(json_response[:bill_address][:address1]).to eq(order.bill_address.address1)
+          expect(json_response[:bill_address][:lastname]).to eq(order.bill_address.lastname)
+          expect(json_response[:ship_address][:address1]).to eq(order.ship_address.address1)
+          expect(json_response[:ship_address][:lastname]).to eq(order.ship_address.lastname)
```
sauloperez: you can do `expect(json_response[:bill_address]).to eq(order.bill_address.attributes)` instead.

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -156,6 +158,109 @@ module Api
...
+          expect(json_response[:shipping_method][:name]).to eq(order.shipping_method.name)
+          json_response[:adjustments].each do |adjustment|
+            if adjustment[:label] == "Transaction fee"
```
sauloperez: I would rather know the adjustment in advance and check the expected results rather than introducing this complexity in the test. We control the test's setup a phase so I don't see a reason why we need this dynamism. 

IMO we need to particularly watch over tests complexity to ease their maintainability. In my experience, highly complex and even metaprogrammed tests ended up not being understood and thus considered by other devs becoming a black box.

app/serializers/api/adjustment_serializer.rb
```diff
@@ -0,0 +1,8 @@
+module Api
+  class AdjustmentSerializer < ActiveModel::Serializer
+    attributes :id, :amount, :label, :eligible
+    attributes :source_type, :source_id
+    attributes :adjustable_type, :adjustable_id
+    attributes :originator_type, :originator_id
```
sauloperez: you can split it into multiple lines without having to call `attributes` many times. IMO that will highlight the fact that they all are attributes of the serializer and that will still be readable. Well, maybe you did it for consitency.

luisramos0: I did it for consistency with the orderserializer but now I see most of them dont have this.
so I fixed all serializers and removed duplicate attributes. new commit.

spec/controllers/api/orders_controller_spec.rb
luisramos0: yes, ok, I removed the email :+1:

spec/controllers/api/orders_controller_spec.rb
luisramos0: I did this just because I am not sure the adjustments will come always in the same order...
anyway, I reverted back to a more straight forward check:

```
          expect(json_response[:adjustments].first[:label]).to eq "Transaction fee"
          expect(json_response[:adjustments].first[:amount]).to eq order.adjustments.payment_fee.first.amount.to_s
          expect(json_response[:adjustments].second[:label]).to eq "Shipping"
          expect(json_response[:adjustments].second[:amount]).to eq order.adjustments.shipping.first.amount.to_s

```

spec/controllers/api/orders_controller_spec.rb
luisramos0: it's not testing rails, it's verifying that this controller plays well with the base_controller, but I agree it is not necessary here :+1:

luisramos0: I use have_http_status  now :+1:
thanks!

app/controllers/api/orders_controller.rb
luisramos0: Yeah, we have Api::OrderSerializer, Api::Admin::OrderSerializer and now Api::OrderDetailedSerializer.
I think this is ok for such a central class (not if it was a less important one). The first one is for the shopfront, the second one for the admin but without much detail and this one includes everything as required in this PR.

We could potentially merge Api::OrderSerializer and Api::Admin::OrderSerializer in one. But I think that's out of the scope of this PR.

I changed from @order to order :+1:

spec/controllers/api/orders_controller_spec.rb
luisramos0: not here because the serializer is not using all the attributes, just a few of them.

spec/controllers/api/orders_controller_spec.rb
luisramos0: yeah, I tried these in previous efforts and I cant make this work for some reason:
```

(byebug) json_response.keys
["id", "number", "user_id", "full_name", "email", "phone", "completed_at", "display_total", "edit_path", "state", "payment_state", "shipment_state", "payments_path", "ship_path", "ready_to_ship", "created_at", "distributor_name", "special_instructions", "payment_capture_path", "item_total", "adjustment_total", "payment_total", "total", "distributor", "order_cycle", "shipping_method", "ship_address", "bill_address", "line_items", "adjustments", "payments"]
(byebug) order_detailed_attributes
["number"]
(byebug) expect(json_response.keys).to include order_detailed_attributes

  1) Api::OrdersController#show as distributor owner returns an order with all required fields
     Failure/Error: expect(json_response.attributes).to include(order_detailed_attributes)
       expected ["id", "number", "user_id", "full_name", "email", "phone", "completed_at", "display_total", "edit_pat..._cycle", "shipping_method", "ship_address", "bill_address", "line_items", "adjustments", "payments"] to include ["number"]
     # (byebug):1:in `expect_detailed_attributes_to_be_present'
     # /Users/luisramos/.rvm/gems/ruby-2.1.5@openfoodnetwork/gems/byebug-9.0.6/lib/byebug/helpers/eval.rb:59:in `eval'

```

and no luck with have_attributes :-(

app/serializers/api/address_serializer.rb
luisramos0: I created this section in the code conventions:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Code-Conventions#performance

Can you help me write something on the "how to avoid this"?

do you know what's the code I should right in each case?

luisramos0:
In terms of changes from previous version, I included the items needed (adjustments, payments, line_items, etc) and then tried to reuse existing serializers as much as possible like Api::ShippingMethodSerializer, Api::AddressSerializer, Api::LineItemSerializer and Api::PaymentSerializer.
These serializers have slight differences from the original spree/rabl payoad, for example:
- addresses dont include a country and state object anymore, they have the fields: state_id, state_name, country_id and country_name
- shipments are gone (👍), we only have order.shipping_method with the details
- payments dont have a payment_method object any more, they just have the name of the payment_method
- line_item.variant changed quite a bit
- the order object gets new much needed attributes: order.distributor, order.order_cycle

From my review I think all these changes are acceptable and some actually good: simpler address and payment objects and no shipments, just ship method.

app/serializers/api/address_serializer.rb
luisramos0: I found [this option](https://stackoverflow.com/a/18134650/1812272). But this is only for nested attributes like object.line_items.variants. Not sure how to make the object itself eagerload for example adjustments...
but there must be a way and definitely something we should do in serializers.

luisramos0:
rebased to get to ruby 2.1.9

app/controllers/api/orders_controller.rb
sauloperez: Also for another PR but perhaps we should rename Api::OrderSerializer to Api::ShopfrontOrderSerializer or something else that reflects the shopfront use case.

.rubocop_manual_todo.yml
```diff
@@ -110,18 +110,6 @@ Metrics/LineLength:
...
-  - app/serializers/api/admin/enterprise_serializer.rb
-  - app/serializers/api/admin/exchange_serializer.rb
-  - app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
-  - app/serializers/api/admin/index_enterprise_serializer.rb
-  - app/serializers/api/admin/index_order_cycle_serializer.rb
-  - app/serializers/api/admin/line_item_serializer.rb
-  - app/serializers/api/admin/order_cycle_serializer.rb
-  - app/serializers/api/admin/subscription_serializer.rb
-  - app/serializers/api/admin/tag_rule_serializer.rb
-  - app/serializers/api/admin/variant_override_serializer.rb
```
sauloperez: OMG

spec/controllers/api/orders_controller_spec.rb
sauloperez: we need to compare the model attributes hash against the key/values we want and then I don't think we'll need these methods we abstracted. I'll try to give it a crack in your branch.

app/serializers/api/address_serializer.rb
sauloperez: we should instead call `#includes` from the controller (or whatever abstraction might be dealing with the data loading in each case) and pass on the data to the view layer. When leave that to perform the data loading is when N+1s pop-up.

In this cases the following patch

```rb
diff --git a/app/controllers/api/orders_controller.rb b/app/controllers/api/orders_controller.rb
index 03d4e8caa..bbbde5fe0 100644
--- a/app/controllers/api/orders_controller.rb
+++ b/app/controllers/api/orders_controller.rb
@@ -26,7 +26,10 @@ module Api
     end
 
     def order
-      @order ||= Spree::Order.find_by_number!(params[:id])
+      @order ||= Spree::Order
+        .where(number: params[:id])
+        .includes(line_items: :variant)
+        .first
     end
   end
 end
```

makes

```
openfoodnetwork@ofn:/opt/openfoodnetwork$ tail -n 200 log/development.log | grep spree_variants                                                                                              
  Spree::Variant Load (0.5ms)  SELECT "spree_variants".* FROM "spree_variants" WHERE "spree_variants"."id" = 6 LIMIT 1                                                                       
  Spree::Variant Load (0.5ms)  SELECT "spree_variants".* FROM "spree_variants" WHERE "spree_variants"."id" = 4 LIMIT 1                                                                       
  Spree::Variant Load (0.4ms)  SELECT "spree_variants".* FROM "spree_variants" WHERE "spree_variants"."id" = 8 LIMIT 1                                                                       
  Spree::Variant Load (0.4ms)  SELECT "spree_variants".* FROM "spree_variants" WHERE "spree_variants"."id" = 2 LIMIT 1     
```

become 

```
openfoodnetwork@ofn:/opt/openfoodnetwork$ tail -n 200 log/development.log | grep spree_variants
  Spree::Variant Load (0.5ms)  SELECT "spree_variants".* FROM "spree_variants" WHERE "spree_variants"."id" IN (6, 4, 8, 2) AND ("spree_variants".deleted_at IS NULL)  
```

Note however that as this is a `show` action we do perform queries like

```
SELECT "spree_line_items".* FROM "spree_line_items" WHERE "spree_line_items"."order_id" IN (4) ORDER BY spree_line_items.created_at ASC
```

for some of the involved tables but perform N+1 for tables related to variants and products, as well as those. I see things like 

```
SELECT "spree_stock_items".* FROM "spree_stock_items" WHERE "spree_stock_items"."variant_id" = 2 ORDER BY id ASC LIMIT 1
```

That should also be eager loaded by using `#includes`.

sauloperez:
have you considered when we would adopt this new endpoint in the Angular codebase?

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -156,6 +158,102 @@ module Api
...
+          expect(json_response[:shipping_method][:name]).to eq order.shipping_method.name
+
+          expect(json_response[:adjustments].first[:label]).to eq "Transaction fee"
+          expect(json_response[:adjustments].first[:amount]).to eq order.adjustments.payment_fee.first.amount.to_s
+          expect(json_response[:adjustments].second[:label]).to eq "Shipping"
+          expect(json_response[:adjustments].second[:amount]).to eq order.adjustments.shipping.first.amount.to_s
+
+          expect(json_response[:payments].first[:amount]).to eq order.payments.first.amount.to_s
+          expect(json_response[:line_items].size).to eq order.line_items.size
+          expect(json_response[:line_items].first[:variant][:product_name]). to eq order.line_items.first.variant.product.name
```
sauloperez: I meant something like

```diff
diff --git a/spec/controllers/api/orders_controller_spec.rb b/spec/controllers/api/orders_controller_spec.rb
index 778653663..bf22a3f5b 100644
--- a/spec/controllers/api/orders_controller_spec.rb
+++ b/spec/controllers/api/orders_controller_spec.rb
@@ -220,21 +220,37 @@ module Api
 
         it "returns an order with all required fields" do
           get :show, id: order.number
-          expect_order
-          expect_detailed_attributes_to_be_present(json_response)
 
-          expect(json_response[:bill_address][:address1]).to eq order.bill_address.address1
-          expect(json_response[:bill_address][:lastname]).to eq order.bill_address.lastname
-          expect(json_response[:ship_address][:address1]).to eq order.ship_address.address1
-          expect(json_response[:ship_address][:lastname]).to eq order.ship_address.lastname
+          expect(response.status).to eq 200
+          expect(json_response[:number]).to eq order.number
+
+          expect(json_response.symbolize_keys.keys).to include(*order_detailed_attributes)
+
+          expect(json_response[:bill_address]).to include(
+            'address1' => order.bill_address.address1,
+            'lastname' => order.bill_address.lastname
+          )
+
+          expect(json_response[:ship_address]).to include(
+            'address1' => order.ship_address.address1,
+            'lastname' => order.ship_address.lastname
+          )
+
           expect(json_response[:shipping_method][:name]).to eq order.shipping_method.name
 
-          expect(json_response[:adjustments].first[:label]).to eq "Transaction fee"
-          expect(json_response[:adjustments].first[:amount]).to eq order.adjustments.payment_fee.first.amount.to_s
-          expect(json_response[:adjustments].second[:label]).to eq "Shipping"
-          expect(json_response[:adjustments].second[:amount]).to eq order.adjustments.shipping.first.amount.to_s
+          expect(json_response[:adjustments].first).to include(
+            'label' => "Transaction fee",
+            'amount' => order.adjustments.payment_fee.first.amount.to_s
+          )
+          expect(json_response[:adjustments].second).to include(
+            'label' => "Shipping",
+            'amount' => order.adjustments.shipping.first.amount.to_s
+          )
+
+          expect(json_response[:payments].first).to include(
+            'amount' => order.payments.first.amount.to_s
+          )
 
-          expect(json_response[:payments].first[:amount]).to eq order.payments.first.amount.to_s
           expect(json_response[:line_items].size).to eq order.line_items.size
           expect(json_response[:line_items].first[:variant][:product_name]). to eq order.line_items.first.variant.product.name
         end
@@ -248,10 +264,6 @@ module Api
       def expect_correct_order(json_response, order)
         expect(json_response[:number]).to eq order.number
       end
-
-      def expect_detailed_attributes_to_be_present(json_response)
-        expect(order_detailed_attributes.all?{ |attr| json_response.key? attr.to_s }).to eq true
-      end
     end
 
     private
```

However, I'd question the need to check almost the entire JSON response. That might fit better in a view spec.

app/serializers/api/address_serializer.rb
luisramos0: ah, this is amazing!

This removes N+1 queries on line_items, variants, products, stock_items and prices :-O
```ruby
   @order ||= Spree::Order.
        where(number: params[:id]).
        includes(line_items: { variant: [:product, :stock_items, :default_price] }).
        first
```

I tried stock_locations but it didnt work, I cant see where the active scope is triggered for stock_locations.

app/serializers/api/order_detailed_serializer.rb
luisramos0: I think we are ok in terms of adjustments, there's one query for them.

luisramos0:
"have you considered when we would adopt this new endpoint in the Angular codebase? just for me to understand."

since v2 we dont have a show page for the order any longer, so I am not seeing any specific use case in the app for this endpoint. maybe it cab be used to build the edit order page for example. But we are building this for external use, it's needed for zapier.

spec/controllers/api/orders_controller_spec.rb
luisramos0: OMG!!!
expect(json_response.symbolize_keys.keys).to include(*order_detailed_attributes)
I learned 2 new ruby things in one line of code 🎉 
I didnt know symbolize_keys and I didnt know the * (splat) operator :-O
thank you!

yeah, I see your point about tooo many checks for a controller test (but dont want to change it now)... This is replacing the serializer unit tests really...

app/serializers/api/address_serializer.rb
luisramos0: I have not managed to act on the initial point about countries...
I tried includes with
{ ship_address: :country }
but it doesnt work...

spec/controllers/api/orders_controller_spec.rb
```diff
@@ -181,5 +287,13 @@ def order_attributes
         :distributor_name, :special_instructions, :payment_capture_path
       ]
     end
+
+    def order_detailed_attributes
+      [
+        :number, :item_total, :total, :state, :adjustment_total, :payment_total,
```
Matt-Yorkley: Do we need to list all of the attributes and test for them? It'll mean the spec has to change if the serializer is touched, but I don't see a real benefit.

Matt-Yorkley: Especially since you subsequently test most of them directly and more explicitly...

luisramos0: for external API endpoints like this one I think we should have these tests yes, ideally we would know what are the ones that are part of the public API (and cannot be removed). If the serializer changes we need to make sure these attributes remain in the payload as they are part of the public api.

I added a commit to remove repeated assertions :+1:

app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
```diff
@@ -6,7 +6,11 @@ class Api::Admin::ForOrderCycle::EnterpriseSerializer < ActiveModel::Serializer
              :is_primary_producer, :is_distributor, :sells
 
   def issues_summary_supplier
-    issues = OpenFoodNetwork::EnterpriseIssueValidator.new(object).issues_summary confirmation_only: true
+    issues =
+      OpenFoodNetwork::EnterpriseIssueValidator.
+        new(object).
+        issues_summary(confirmation_only: true)
```
kristinalim: I experience this sometimes and it feels like making the code less readable just to pass the rules we use for Rubocop. But I have no suggestion that would make styling existing code easier. :+1:

luisramos0:
yes @kristinalim 

1. yes, @lin-d-hop does
2. probably, @lin-d-hop to validate when she is available
3. maybe, let's see how the testing works


"Another thing is that I noticed shipments which includes the tracking code"
tracking code is new in v2 and is even considered for deletion in #4146, so definitely not required here.

lin-d-hop:
I think I'll need to test this one to verify @openfoodfoundation/testers 
Can I be assigned when it gets to Test Ready please? :-)

app/controllers/api/orders_controller.rb
```diff
@@ -19,5 +24,12 @@ def serialized_orders(orders)
         each_serializer: Api::Admin::OrderSerializer
       )
     end
+
+    def order
+      @order ||= Spree::Order.
+        where(number: params[:id]).
+        includes(line_items: { variant: [:product, :stock_items, :default_price] }).
```
Matt-Yorkley: Nice.I just did some work on a variant query, and eager-loading `:product, :stock_items, :default_price` made a massive difference. We should take a look and maybe add this to the other orders pages at some point.

lin-d-hop:
Sorry for the delay on this!
I've now used this new endpoint to integrate to accounting packages via Zapier testing the endpoint works with the existing integrations. Lots of API improvements and all the old things that we used are maintained. I'm happy this endpoint is ready to go.... and hoping I didn't miss anything.

Now just awaiting @kirstenL to feedback on how they are using the endpoint for integrations in Aus. Apparently they do have some more complex ones...

lin-d-hop:
Ok tested and I'm happy.
There is a small chance that this will break our Zapier integration as I don't know how sensitive Zapier is to small API changes. So it would be great to have a release date/time that will allow Aus and UK to test our integrations once deployed.
Thanks team! Sorry it took me so long to get to!

luisramos0:
ok, great thanks. Let's keep that in mind, next release needs to be on a agreed time to verify UK and AUS integration.

Another more reliable option is to connect the zapier integration to a staging environment... 
I'll merge but we can still deploy master to some staging environment before the release and verify zapier by connecting it to that staging server.

Issue #4050, opened 2019-07-16
Make OFN independent of spree_backend


luisramos0:
### What we should change and why (this is tech debt)
Most of spree_backend code is overriden in OFN thus creating a mess between OFN code on top of spree code. By removing spree_backend and moving what's needed to OFN, working with our code base becomes a LOT easier.

Before we do this we need to finish de-deface all spree_backend views as part of #2807, this is very close to completion now :tada:

After this is done, we can move the views and controllers that are not in OFN yet:
- 8 in the config space:
  - - [X] countries, states and zones #4290
  - - [X] general settings, image_settings and mail methods #4289
  - - [X] tax_categories, tax_rates and tax_settings #4292
  - - [X] taxonomies and taxons #4291
  - - [X] shipping_methods and shipping_categories #4269
  - - [X] payment_methods (merge controllers) #4371
- 5 in the product space:
  - - [X] images (copy controller and views to OFN) #4372
  - - [x] option_types and option_values (copy controllers and views to OFN) moved in #4389 but removed in #4397
  - - [x]  prototypes - the prototypes feature is not used since 2013 #4395
  - - [x] product_properties and properties - (copy properties views and the controllers) #4449
- 1 in the orders space:
  - - [x] return_authorizations (copy views and controller) #4450
- and 1 for user searches:
  - - [x] search (merge controller and get rid of rabl view, is this really used in the users admin page?) #4511

These are the 9 views we dont need to copy because we dont use them in OFN: banners, promotion_actions, promotion_rules, promotions, stock_items, stock_locations, stock_movements, stock_transfers, trackers (we should remove GA support).

We have 10 controller decorators that need to be merged with the original controllers (all code in original controller that is not needed should not be copied):
  - - [x] adjustments_controller_decorator.rb #4515
  - - [x] base_controller_decorator.rb #4512
  - - [x] orders_controller_decorator.rb #4513
  - - [x] customer_details_controller_decorator.rb #4513
  - - [x] overview_controller_decorator.rb #4514
  - - [x] payments_controller_decorator.rb #4509
  - - [x] products_controller_decorator.rb #4508
  - - [x] variants_controller_decorator.rb #4508
  - - [x] reports_controller_decorator.rb #4507
  - - [x] resource_controller_decorator.rb #4510
  - - [x] search_controller_decorator.rb #4511

And we have 3 controllers that are not decorated in OFN but may be needed, or maybe not:
[...]

PR #4269, opened 2019-09-17
Move shipping_methods and shipping categories from spree_backend to OFN

luisramos0:
### What? Why?
Part of #4050

This moves all things related to shipping methods and shipping categories from spree_backend to OFN.

### What should we test?
The code was just moved around so a basic test should do.

Existing bug around User feedback on delete:
The flash message after pressing the delete button is still not showing. It was broken before, I tried fixing it now but didnt manage to. Still not showing.
If ship method is deleted successfully, no flash message to user.
If the ship method is referenced by an order, the ship method is not deleted and no flash message to user is shown. But the ship method will disappear from the list and reappear if user refreshes page 🙈 

### Release notes
Changelog Category: Changed


daniellemoorhead:
>The flash message after pressing the delete button is still not showing. It was broken before, I tried fixing it now but didnt manage to. Still not showing.
If ship method is deleted successfully, no flash message to user.
If the ship method is referenced by an order, the ship method is not deleted and no flash message to user is shown. But the ship method will disappear from the list and reappear if user refreshes page 🙈

@lin-d-hop would this be worthy of an admin UX (or maybe it already is)?

filipefurtad0:
Hi :-)

What was tested:
Performed general test on Katuma as Administrator, creating user, adding/deleting, managing order cycles.
Created user, ordered a product.

Observed:
All worked fine. Verified the bugs you previously mentioned:
- The flash message after pressing the delete button is still not showing.
- If ship method is deleted successfully, no flash message to user.
- If the ship method is referenced by an order, the ship method is not deleted and no flash message to user is shown. But the ship method will disappear from the list and reappear if user refreshes page

I didn't detect any further errors, from moving the code from spree_backend to OFN.
Moving to Ready to Go! :tada:

luisramos0:
well done Filipe!

PR #4286, opened 2019-09-20
Finish transpec

### LINKED ISSUE #2097: Disable `should` syntax in specs
## What? Why?

We are still using the very old RSpec should syntax. Chances are that this syntax won’t work in upcoming versions of RSpec and that may prevent us from upgrading RSpec in the future.

The problem is that we waste valuable time and energy reviewing the syntax used in a PR when this is better done by machines. We would be better off focusing on architecture and implementation.

This stems from https://github.com/openfoodfoundation/openfoodnetwork/pull/2084, which is a good example of this.

## Proposed solution

This needs to be disabled with RSpec itself. See https://relishapp.com/rspec/rspec-expectations/docs/syntax-configuration#disable-should-syntax.

## Dependencies

We tried doing it in https://github.com/openfoodfoundation/openfoodnetwork/pull/2180 but as explained in https://github.com/openfoodfoundation/openfoodnetwork/pull/2180#issuecomment-376897246, it can't be done until we reach Spree 3.0.

luisramos0:
#### What? Why?
With the upgrade to ruby 2.1.9 I gave it a try to my local feature specs. Still a few errors but good enough to finish transpec'ing all our specs 🎉 

Result:Expect is mandatory now!
Screenshot 2019-09-20 at 10 34 18

Closes #2097

#### What should we test?
No test, spec changes only.

#### Release notes
Changelog Category: Changed
All OFN tests are now using modern rspec syntax.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -91,6 +91,11 @@ RSpec.configure do |config|
   # Retry
   config.verbose_retry = true
 
+  # Force use of expect (over should)
+  config.expect_with :rspec do |expectations|
+    expectations.syntax = :expect
+  end
+
   # DatabaseCleaner
   config.before(:suite)          { DatabaseCleaner.clean_with :deletion, except: ['spree_countries', 'spree_states'] }
   config.before(:each)           { DatabaseCleaner.strategy = :transaction }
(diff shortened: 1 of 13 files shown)


spec/spec_helper.rb
```diff
@@ -91,6 +91,11 @@
   # Retry
   config.verbose_retry = true
 
+  # Force use of expect (over should)
+  config.expect_with :rspec do |expectations|
+    expectations.syntax = :expect
+  end
```
sauloperez: THANK YOU

PR #4287, opened 2019-09-20
Replace therubyracer with mini_racer.

### LINKED ISSUE #2100: Update or replace therubyracer

luisramos0:
#### What? Why?
With ruby 2.1.9 I gave this a new try without success but a little bit more google time and I found the solution 🎉  in this [blog post](https://blog.driftingruby.com/updated-to-mojave/).

Closes #2100
Redos #2424 (that was reverted in #3244)

This is being replaced to increase the reliability and speed of asset compilation.

#### What should we test?
Assets are compiled in staging.
A successful deploy to a staging server is enough.

#### Release notes
Changelog Category: Changed
Move from therubyracer to mini_racer to reduce deploy time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -104,7 +104,10 @@ group :assets do
   gem 'coffee-rails', '~> 3.2.1'
   gem 'compass-rails'
 
-  gem 'therubyracer', '=0.12.0'
+  gem 'mini_racer', '0.1.15'
+  # Previously we found that libv8 6.7.288.46.1 breakis the compilation of mini_racer.
+  # Now we see that we need to set the version explicitly. Nothing else depends on libv8.
+  gem 'libv8', '6.3.292.48.1'
 
   gem 'uglifier', '>= 1.0.3'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -465,7 +465,7 @@ GEM
       addressable (~> 2.3)
     letter_opener (1.7.0)
       launchy (~> 2.2)
-    libv8 (3.16.14.19)
+    libv8 (6.3.292.48.1)
     listen (3.0.8)
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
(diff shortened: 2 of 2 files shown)


luisramos0:
staging to katuma worked!
but the precompile time is the same as before 🤷‍♂

sigmundpetersen:
Yeah there were some comments on not much improved performance in the previous PR as well, no? #2424
Try another deploy

Gemfile
```diff
@@ -104,7 +104,10 @@ group :assets do
   gem 'coffee-rails', '~> 3.2.1'
   gem 'compass-rails'
 
-  gem 'therubyracer', '=0.12.0'
+  gem 'mini_racer', '0.1.15'
+  # Previously we found that libv8 6.7.288.46.1 breakis the compilation of mini_racer.
+  # Now we see that we need to set the version explicitly. Nothing else depends on libv8.
+  gem 'libv8', '6.3.292.48.1'
```
sauloperez: I thought we tried this already

luisramos0:
staging to FR to see if it works.

luisramos0:
staging worked :+1:

PR #4288, opened 2019-09-20
Add spec for api/admin/order_cycle_serializer

luisramos0:
#### What? Why?

Relates to #4007
Adds some simple specs to complex and untested order_cycle_serializer.

I thought I'd grow this one as I worked on OC page but actually, these visible_* and editable_* methods on the OC will go (#4480 etc) so I'll skip adding more unit tests for them. We can keep this small spec in the meantime.

#### What should we test?
This is adding specs.

#### Release notes
Changelog Category: Added
Added new automated tests to a serializer.

### DIFF
--- a/spec/serializers/api/admin/order_cycle_serializer_spec.rb
+++ b/spec/serializers/api/admin/order_cycle_serializer_spec.rb
@@ -1,0 +1,27 @@
+require "spec_helper"
+
+module Api
+  module Admin
+    describe OrderCycleSerializer do
+      let(:order_cycle) { create(:order_cycle) }
+      let(:serializer) { Api::Admin::OrderCycleSerializer.new order_cycle, current_user: order_cycle.coordinator.owner }
+
+      it "serializes an order cycle" do
+        expect(serializer.to_json).to include order_cycle.name
+      end
+
+      it "serializes the order cycle with exchanges" do
+        expect(serializer.exchanges.to_json).to include "\"#{order_cycle.variants.first.id}\":true"
+      end
+
+      it "serializes the order cycle with editable_variants_for_incoming_exchanges" do
+        expect(serializer.editable_variants_for_incoming_exchanges.to_json).to include order_cycle.variants.first.id.to_s
+        expect(serializer.editable_variants_for_incoming_exchanges.to_json).to_not include order_cycle.distributors.first.id.to_s
+      end
+
+      it "serializes the order cycle with editable_variants_for_outgoing_exchanges" do
+        expect(serializer.editable_variants_for_outgoing_exchanges.to_json).to include order_cycle.variants.first.id.to_s
+      end
+    end
+  end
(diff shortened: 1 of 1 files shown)


spec/serializers/api/admin/order_cycle_serializer_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    describe OrderCycleSerializer do
+      let(:order_cycle) { create(:order_cycle) }
+      let(:serializer) { Api::Admin::OrderCycleSerializer.new order_cycle, current_user: order_cycle.coordinator.owner }
+
+      it "serializes an order cycle" do
+        expect(serializer.to_json).to match order_cycle.name
+      end
+
+      it "serializes the order cycle with exchanges" do
+        expect(serializer.exchanges.to_json).to include order_cycle.variants.first.id.to_s
```
mkllnk: Searching for a random number (id) in a string can easily lead to false positives. I would probably write it like this:

```rb
exchanges = JSON.parse(serializer.exchanges.to_json)
expect(exchanges).to include include({"variants" => {order_cycle.variants.first.id.to_s => true}})
```

I know it's a mind boggling matcher. The first include matches any element of the array and the second include is a partial match of the hash.

A compromise would be to match part of the JSON to avoid the id `8` being matched in `18`:
```rb
expect(serializer.exchanges.to_json).to include "\"order_cycle.variants.first.id\":true"
```

But that requires knowledge of the JSON format used and formatting would break it. And it's still not testing that the id belongs to a variant.

sigmundpetersen:
@luisramos0 did you want to address Maikel's comments above or is everything good here?

spec/serializers/api/admin/order_cycle_serializer_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "spec_helper"
+
+module Api
+  module Admin
+    describe OrderCycleSerializer do
+      let(:order_cycle) { create(:order_cycle) }
+      let(:serializer) { Api::Admin::OrderCycleSerializer.new order_cycle, current_user: order_cycle.coordinator.owner }
+
+      it "serializes an order cycle" do
+        expect(serializer.to_json).to match order_cycle.name
```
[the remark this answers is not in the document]
luisramos0: I changed to include. match works in the same way for strings but can take a regexp. not needed in this case.

spec/serializers/api/admin/order_cycle_serializer_spec.rb
luisramos0: yeah, good point! I used your last suggestion 👍

luisramos0:
@sigmundpetersen yes, low priority work. done ;-)

sauloperez:
@luisramos0 feel free to merge it as soon as it passes CI

PR #4289, opened 2019-09-20
Move general_settings, mail_methods and image_settings from spree_backend to OFN

luisramos0:
#### What? Why?

Part of #4050

This moves all things related to general_settings, mail_methods and image_settings from spree_backend to OFN. No change in functionality, although it's a good opportunity to make small changes to these pages, let me know.

#### What should we test?
The 3 pages must work correctly as before. All properties need to be updatable and the effects should be as expected. There are quite a few small details maybe we should check.

I have verified locally almost every checkbox updates the database field correctly.

General Settings: we should test the checkboxes work in the page but maybe it's too much unnecessary work to verify the effects on the app of every checkbox...
 
Mail Methods: same as General Settings, maybe we should disable email deliveries and check no emails are sent?

Image Settings: I have tested locally the database values are updated correctly. I am not sure how to test this page... setup s3 images for a staging server? That would be a dev-test.

#### Release notes
Changelog Category: Changed


RachL:
Checkboxes of general settings: all good except I can find where the confidentiality policy is shown in the front end. I remembered a second link just near the cookie policy link in the cookies banner. 
This is not linked to this PR, so it should not blocked it but do you remember where this link is shown @luisramos0 ?

Mail methods : I thought disabling the checkbox would prevent mail to be sent, but I can still receive order confirmation email. How to disable them completely?

luisramos0:
rebased to resolve conflict.

luisramos0:
it's next to the cookies policy in the banner,
I can see it locally with this branch if I define the URL in general settings:
Screenshot 2019-10-02 at 21 48 36

luisramos0:
I tested the mail methods in staging FR and the page is working correctly, the db setting is correct.
I also tested locally and this code is indeed executed on save:
ActionMailer::Base.perform_deliveries = false

so, if the emails are still being sent after you uncheck the box, we should create a new issue and investigate.

luisramos0:
I tested on staging UK and unchecking the checkbox will not stop emails from being sent. That indicates it's an existing bug.

RachL:
@luisramos0 about the cookie policy I've found my mystake: the field does allow a full copy / paste of the URL. It worked well when I've removed the "http://" :+1:

There is only one thing left to test and it is images. As we need a dev for it, I'm putting the dev-test label.

kristinalim:
I did these after deploying to staging (AU):

1. Enabled Amazon S3, and set bucket details and access credentials.
2. Uploaded photo for a product. :heavy_check_mark: 
3. Saw uploaded photo for product. :heavy_check_mark: 
4. Checked that the image URL is in the newly set S3 bucket. :heavy_check_mark: 
5. Removed the image.
6. Disabled Amazon S3.
7. Uploaded photo for a product.
8. Confirmed that the new image URL is no longer in the S3 bucket but in instance server storage. :heavy_check_mark: 

Locally, I also checked that changing a few other image settings work. :heavy_check_mark:

kristinalim:
Moving this to Ready to Go! :tada:

luisramos0:
awesome Kristina, thanks.

PR #4290, opened 2019-09-20
Move zones, countries and states from spree_backend to OFN

luisramos0:
### What? Why?
Part of #4050

This moves all things related to zones, countries and zones from spree_backend to OFN. No change in functionality, although it's a good opportunity to make small changes to these pages, let me know.

### What should we test?
The 3 pages must work correctly as before.
Zones: create and edit zone, add and remove countries and states from a zone.
Countries: edit countries.
States: create state, edit state, add new state to country. Delete new state.
Existing bug: when deleting an existing state, if state is used somewhere in the system: very ugly error flash message.

### Release notes
Changelog Category: Changed


RachL:
This seems ready to go!

- Zones: create and edit zone, add and remove countries and states from a zone is working good. 
- Countries: editing works
- States: CRUD works
- Existing bug: when deleting an existing state, if state is used somewhere in the system: very ugly error flash message => yes indeed:
image

Issue #2944, opened 2018-10-27
Taxonomies - Sorting entries in a taxonomy is broken


luisramos0:
## Description
In taxonomies, if you edit a taxonomy and try to sort the entries ((they are called taxons internally)) with drag and drop, apparently the actions work but if you refresh the page, the order of the entries goes back to what it was before the change.

## Expected Behavior
The order set with drag and drop actions should remain.

## Actual Behavior
It's not possible to change the order of the entries in a taxonomy.

## Steps to Reproduce
Admin  > Configuration > Taxonomies and see that the sorting of taxonomies is broken.

## Context
https://community.openfoodnetwork.org/t/how-to-manage-product-categories-in-different-languages-in-a-multilingual-instances/1471/2

## Severity
S3? - a feature is broken and there is no workaround but the "sorting of product categories" is a very unimportant feature

## Your Environment
Local

* Version used: latest master
* Browser name and version: chrome 69
* Operating System and version (desktop or mobile): macos highsierra

## Possible Fix
I believe there's something in the management/updating of the spree_taxons.position DB field. I couldnt find the solution quickly... 
Meanwhile, If #2945 gets done, this issues would include reverting the alphabetical order and using the order defined by the user in the taxonomies screen.

luisramos0:
I think this is a good first issue if you have experience with the tech stack but are new to OFN.

YanhanLyu:
Hi, I'm happy to take this issue as my first one!

sigmundpetersen:
Go ahead @YanhanLyu !

daniellemoorhead:
Welcome @YanhanLyu 😄 
Looking forward to seeing your first PR come through the team for review!

haseleyi:
Regarding the ordering of taxons, are these taxons in the sample data supposed to be alphabetically ordered when we visit this view?

![image](https://user-images.githubusercontent.com/18274798/49676120-4d2bf980-fa2d-11e8-95cd-ba24819fda72.png)

haseleyi:
Sounds good thanks! Just to understand how the current code works for context, where does the alphabetical sorting kick in? I created new taxonomies and children therein and wasn't able to find it. I ask because we're currently ordering by the taxon's name property, but if that works, then ordering by the position property sounds promising.

sigmundpetersen:
I think the alphabetic sorting relates to the dropdown in new/edit product, c.f. #2945 #2946 
And not this main Taxonomy page

[...]

PR #4291, opened 2019-09-20
Move taxons and taxonomies from spree_backend to OFN 

luisramos0:
### What? Why?
Part of #4050

This moves all things related to taxonomies and taxons from spree_backend to OFN. No change in functionality.

### What should we test?
The pages must work correctly as before.
Taxonomies: create taxonomy, sort taxonomies. Edit taxonomy: (use right click) add, rename, remove and edit Taxons.
Existing issues:
- cannot delete taxonomy even if newly created if more than one level of taxons.
- cannot sort taxons inside taxonomy #2944

### Release notes
Changelog Category: Changed


Matt-Yorkley:
There's quite a few Codeclimate issues. Do you want to handle them separately?

luisramos0:
this is spree code, these are the not easy ones (complexity, etc). I dont want to fix all of them, a lot of work...

luisramos0:
moving to test ready.

RachL:
@luisramos0 code climate is complaining. Can I still test it?

luisramos0:
yes @RachL I dont think we need to fix all issues in the spree code.

RachL:
@luisramos0 I wonder : how can I actually create a taxonomy? I would need to drag it above "products" which is our main taxonomy, right?

luisramos0:
you use the button "new taxonomy"

RachL:
@luisramos0 :woman_facepalming: I was looking at the edit taxonomy page, so sorry.

Taxonomies:CRUD is working. Sorting is working in the UI, but not in the product category list
Taxons : add and edit is working.

I believe we will need another issue for sorting taxonomies. But this is pre-existing so it is ready to go!

luisramos0:
thank you!
there's an issue already for the sorting problem: #2944

RachL:
@luisramos0 this issues covers the sorting within a taxonomy so the sorting of taxons. But I tested as well the sorting of taxonomies. And nothing happened on the categories dropdown when creating a product for example. But I guess for now sorting taxonomies is less prioritary.

luisramos0:
yeah, and is the order of taxonomies used anywhere now? I mean, the product categories are not taxonomies, they are taxons inside the Product Categories taxonomy.

luisramos0:
I dont think I understand what is weird. We have a taxonomies system and we use one of the taxonomies as product categories. What's the confusion?

sigmundpetersen:
In 'New Product' you can choose the taxonomy as on of the categories, e.g. 'Products', 'Others'. I guess that's the weird thing you mention, no @RachL ? They're sorted alphabetically among the other taxons/categories

luisramos0:
ah, my mistake, all taxonomies and all taxons are shown on the product categories list. I think only taxons inside the Products taxonomy should be shown. But that's not the case now.

PR #4292, opened 2019-09-20
Move tax_rates, tax categories and tax settings from spree_backend to OFN

luisramos0:
### What? Why?
Part of #4050

This moves all things related to tax_rates, tax categories and tax settings from spree_backend to OFN. No change in functionality.

### What should we test?
The pages must work correctly as before.
Tax Categories: create, edit and delete tax category.
Tax rates: create, edit and delele tax rate.
Tax Settings: make sure the selections are saved.
I dont think we need to verify that these settings are working in the app, nothing as changed in the app, if these pages show the correct info, the info will be used in the app as before.

### Release notes
Changelog Category: Changed


RachL:
- Tax Categories: CRUD is working. Noticed the default is quite useless now as there is the "none" tax category which was hardcoded somewhere I guess and is selected by default at product creation
- Tax rates: CRUD is working. 
- Tax Settings: selections are correctly saved.

Moving to ready to go!

PR #4293, opened 2019-09-20
Improve user email change flash alert message

### LINKED ISSUE #4031: Add a message for superadmin when changing users email.
## Description
In the backoffice, when editing a user as super admin requires a confirmation on the new email. However the superadmin may not know that and see only that his email changes have not be taken into account. A notification / message should be added.

## Expected Behavior
Admin should understand easily that changing an email requires a new confirmation in order to be taken into account.

## Actual Behaviour
After changing the email and saving the form, the user email remains the same and the message says that it has worked.

## Steps to Reproduce
1. Go to `admin/users` 
2. Click on an email
3. Change the email (with a mailbox you have access to)
4. Hit update
5. A green message will tell you that it has worked. But nothing has changed: the old email is still there. However the new email got a confirmation request email.
6. Click on the confirmation email and validate the new email
7. Go to `admin/users` 
8. See that the new email is correctly there

## Animated Gif/Screenshot
image

## Context
Found when moving spree_auth_devise code to ofn.

## Severity

## Your Environment
* Version used: v2 
* Operating System and version (desktop or mobile): macos
* OFN Platform instance where you discovered the bug, and which version of the software they are using. Katuma live v2.1.0

## Possible Fix

Add a message saying "The account will be updated once the new email is confirmed."

jasonk357:
Closes #4031

When editing a user as super admin requires a confirmation on the new email. However the superadmin may not know that and see only that his email changes have not be taken into account. This changes the message from "Account Updated!" to "The account will be updated once the new email is confirmed."

Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/users_controller.rb
+++ b/app/controllers/spree/admin/users_controller.rb
@@ -46,11 +46,12 @@ module Spree
             @user.spree_roles = roles.reject(&:blank?).collect{ |r| Spree::Role.find(r) }
           end
 
-          if params[:user][:email] != @user.email
-            flash.now[:success] = Spree.t(:email_updated)
-          else
-            flash.now[:success] = Spree.t(:account_updated)
-          end
+          message = if params[:user][:email] != @user.email
+                      Spree.t(:email_updated)
+                    else
+                      Spree.t(:account_updated)
+                    end
+          flash.now[:success] = message
         end
         render :edit
       end
--- a/app/controllers/spree/admin/users_controller.rb
+++ b/app/controllers/spree/admin/users_controller.rb
@@ -46,7 +46,11 @@ module Spree
             @user.spree_roles = roles.reject(&:blank?).collect{ |r| Spree::Role.find(r) }
           end
 
-          flash.now[:success] = Spree.t(:account_updated)
+          if params[:user][:email] != @user.email
+            flash.now[:success] = Spree.t(:email_updated)
+          else
+            flash.now[:success] = Spree.t(:account_updated)
+          end
         end
         render :edit
       end
--- a/app/controllers/spree/admin/users_controller.rb
+++ b/app/controllers/spree/admin/users_controller.rb
@@ -46,7 +46,7 @@ module Spree
             @user.spree_roles = roles.reject(&:blank?).collect{ |r| Spree::Role.find(r) }
           end
 
-          message = if params[:user][:email] != @user.email
+          message = if new_email_unconfirmed?
                       Spree.t(:email_updated)
                     else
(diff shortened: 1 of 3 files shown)


jasonk357:
> nice!!! it should be easy to add a test for this here:
> https://github.com/openfoodfoundation/openfoodnetwork/blob/9807b432f957c53ba7b448530635923391671b49/spec/features/admin/users_spec.rb#L55

Ah sorry about that! Added a spec.

sauloperez:
Code Climate is complaining @jasonk357 but it should be pretty straightforward.

app/controllers/spree/admin/users_controller.rb
```diff
@@ -46,7 +46,12 @@ def update
             @user.spree_roles = roles.reject(&:blank?).collect{ |r| Spree::Role.find(r) }
           end
 
-          flash.now[:success] = Spree.t(:account_updated)
+          message = if params[:user][:email] != @user.email
```
Matt-Yorkley: Hi @jasonk357 thanks for the PR :smiley: 

These lines are good logic-wise, but could be a bit clearer for future devs trying to understand the code. It might be nice to extract this conditional logic into a human-readable (private) method. For example:

```ruby
     ...

     message = if new_email_unconfirmed?
          Spree.t(:email_updated)
        else
          Spree.t(:account_updated)
        end
     flash.now[:success] = message
  end
  render :edit
end

...

private

def new_email_unconfirmed?
   params[:user][:email] != @user.email
end
```

jasonk357: c46fe7b

filipefurtad0:
Hi @jasonk357,

I've staged this PR in Katuma and made a change on the contact email of a given user, under backoffice mode, and :tada: one gets the message, indicating the requirement of confirmation :-)
#4293_kat

However, if I press the Update button a second time, the notifiifcation changes to "Account Updated!" although this isn't yet the case.
#4293_kat_2

So, the PR does address the issue, but I feel it could be improved: further pressing the Update button should display the requirement of confirmation again. Is this a small change and the PR is Ready to Go?

Matt-Yorkley:
Thanks @filipefurtad0!

> if I press the Update button a second time, the notification changes to "Account Updated!" although this isn't yet the case.

In terms of moving this forward, I think we can probably merge this as it is and address this issue in another PR.

luisramos0:
@filipefurtad0 on your second click the email field has not changed, it's still the old confirmed email, so there's no reference to the email field change.
I think your requirement is to show the warning whenever the user is updated and there's an unconfirmed email. I'd say that's a different issue and I'd go and merge this as is.

PR #4296, opened 2019-09-22
Make subscription line items handle soft deleted variants

### LINKED ISSUE #4062: Subs page not displaying if a line item in an order was deleted
## Description

In French production, we cannot reach the demo hub subscription page: there is an infinite spinner. A line item with no variant is causing this.

## Steps to Reproduce
1. Log in as rachel.arnould[at]openfoodfrance.org or super admin
2. Try to reach `/admin/subscriptions` for the Hub demo Open Food France
3. The spinner shows up but never ends

## Severity

## Your Environment
* Version used:v.2.2.1

## Possible Fix

 
The underlying issue is a line item with a missing variant. It probably got deleted at some point.

```
app/serializers/api/admin/subscription_line_item_serializer.rb:8:in `description': undefined method `product' for nil:NilClass (NoMethodError)
```
https://app.bugsnag.com/open-food-france/open-food-france-prod/errors/5d36638487f0ee0018332a37

A simple check for `nil` can fix the serializer. But we also need to find out how this can happen. We should either prevent the deletion of a variant that is still in use or update all subscriptions or make line items aware of deleted variants.

This is a duplicate of #3948.

luisramos0:
This ensures subscription list page works. The variant can be removed from the subscription by the user in the edit subscription page

#### What? Why?

Closes #4062

This makes the subs list page work. It keeps the deleted variant in the subscription.
According to my manual test it looks good because when the user goes to edit the order, the deleted variant shows up but it is flagged it is not available in any order cycle, so the user will be able to remove it from the subscription.

#### What should we test?
Verify that the issue is fixed.
Additionally verify that the variant can be removed from the subscription by the user.

#### Release notes
Changelog Category: Fixed
Subscriptions list page can now handle subscriptions with deleted products.

### DIFF
--- a/app/models/subscription_line_item.rb
+++ b/app/models/subscription_line_item.rb
@@ -10,6 +10,11 @@ class SubscriptionLineItem < ActiveRecord::Base
     (price_estimate || 0) * (quantity || 0)
   end
 
+  # Ensure SubscriptionLineItem always has access to soft-deleted Variant attribute
+  def variant
+    Spree::Variant.unscoped { super }
+  end
+
   # Used to calculators to estimate fees
   alias_method :amount, :total_estimate
 
--- a/spec/serializers/api/admin/subscription_line_item_serializer_spec.rb
+++ b/spec/serializers/api/admin/subscription_line_item_serializer_spec.rb
@@ -1,0 +1,25 @@
+require 'spec_helper'
+
+module Api
+  module Admin
+    describe SubscriptionLineItemSerializer do
+      let(:subscription_line_item) { create(:subscription_line_item) }
+      
+      it "serializes a subscription line item with the product name" do
+        serializer = described_class.new(subscription_line_item)
+
+        expect(serializer.to_json).to match subscription_line_item.variant.product.name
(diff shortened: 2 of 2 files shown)


RachL:
The tests I did:

- I created a new user with the account rachel.arnould+subs[at]gmail.com
- Created 2 product (Product 1 and product 2)
- And opened a new sub with these two products
- After deleting product 1, I can see there is a problem within  the subs (no more infinite spinner :tada: )

image

However I cannot remove the deleted product from the subscription:

image

(I've tried through the edit page as well).

This PR is fixing the original bug so I think we should go on and merge it.

Does this need a new issue or could it be fixed while working on : https://github.com/openfoodfoundation/openfoodnetwork/issues/4203 @luisramos0 ? I can update the description there.

luisramos0:
hmmm, I think I edited the subscription itself not the items and there I managed to remove the product. Can you try that?

RachL:
@luisramos0 yes,when you do that it seems to have worked, but when you open up again the subs, the 2 products are still there.

PR #4298, opened 2019-09-22
Make unit description visible in the variant edit page even for products which variant_unit is items

### LINKED ISSUE #3649: Create New Variant Fails if Unit Type is set to "Item"
## Description

When creating a new variant via the +New Variant link on the product page, it will fail EVERY time if the unit type is set to "item."

## Expected Behavior

The form should include the "Unit description" field, which is required to complete submission and create the new variant.

## Actual Behaviour

The "Unit descirption" field does not appear on the add variant screen if the unit type is set to "item." Because it is missing, it can not be submitted, and form validation fails because this mandatory (but hidden) field is blank when it is submitted.

## Steps to Reproduce

1. Go to products page.
2. Click "+ New Variant" button in upper right.
3. Complete the form as presented.
4. Submit.
5. Receive error message.

## Animated Gif/Screenshot
VariantError-ItemUnit

## Context

I am unable to create new variants of a product if the unit type is set to "Item." If it is changed to any other type (weight, volume) on the product page, everything works as expected. For "item" types, it does not.

It IS still possible to create a new variant through the bulk edit screen, expanding the product, and clicking the + on the left. But this also fails in a significant way. While it allows creating a new variant, there is NO way to add a description of it so long as the unit type is set to "item." You can also NOT edit the description of **any existing product** variant with an "item" unit type.  Everything works as expected for all other unit types. It all seems to hinge on the unit description field being missing for "item" types.

## Severity

bug-s3

I would honestly consider this an s2, because there really is no workaround for being able to give "item" types a variant description. Setting the unit to weight or volume, creating a description, then changing it back almost works - except it adds a unit quantity to the description, including a "0" if you leave it blank.  But there is one functioning workaround ... create a completely new product instead of a variant. So I ultimately went with s3.

## Your Environment

* Version used: Production
* Browser name and version: Chrome 73
* Operating System and version (desktop or mobile): Windows 10 Pro desktop
* OFN Platform instance where you discovered the bug, and which version of the software they are using. OFN-Canada ... I have no idea what version it's running on. 

## Possible Fix

Include the "Unit description" field in the add variant form, exactly the same way it is for literally every other unit variant. Seems that would solve the issue in a hurry, because it's already working for every other unit type.

In discussion on Slack, Luis Ramos added this:
> Luis Ramos   
> the code is actually excluding the fields explicitly, I have no idea why.. that will have to be investigated as part of your issue
> the code:    - if @product.variant_unit != ‘items’

luisramos0:
#### What? Why?

Closes #3649

This is also reverting #3896 as it becomes unnecessary to clear the field. The user can now edit the value as needed.

#### What should we test?
We need to test the variant edit page for the 3 types of variants: weight, volume and items.
We need to make a sanity check by making the product with "items" unit available for sale and checkout with it, to verify the change is not breaking the checkout.

#### Release notes
Changelog Category: Changed
Variant.unit_description field is now editable for variants with items units.

### DIFF
--- a/app/controllers/spree/admin/products_controller_decorator.rb
+++ b/app/controllers/spree/admin/products_controller_decorator.rb
@@ -44,8 +44,6 @@ Spree::Admin::ProductsController.class_eval do
     delete_stock_params_and_set_after do
       super
     end
-
-    clear_variants_unit_description if @object.variant_unit == 'items'
   end
 
   def bulk_update
@@ -189,11 +187,5 @@ Spree::Admin::ProductsController.class_eval do
 
   def set_product_master_variant_price_to_zero
     @product.price = 0 if @product.price.nil?
-  end
-
-  def clear_variants_unit_description
-    @object.variants.each do |variant|
-      variant.update_attribute :unit_description, ''
-    end
   end
 end
--- a/app/views/spree/admin/variants/_form.html.haml
+++ b/app/views/spree/admin/variants/_form.html.haml
@@ -14,9 +14,9 @@
         = text_field_tag :unit_value_human, nil, {class: "fullwidth", 'ng-model' => 'unit_value_human', 'ng-change' => 'updateValue()'}
         = f.text_field :unit_value, {hidden: true, 'ng-value' => 'unit_value'}
 
-      .field
-        = f.label :unit_description, t(:spree_admin_unit_description)
-        = f.text_field :unit_description, class: "fullwidth", placeholder: t('admin.products.unit_name_placeholder')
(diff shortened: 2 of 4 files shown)


filipefurtad0:
Hi @luisramos0,
This looks great now :-) All good!

Notes here:
https://docs.google.com/document/d/1dZtPu-uTz5Cg21eys7FhrfJEk7wbo2bqQ3Fip2YOykU/edit#

Moving to Ready to Go

luisramos0:
thanks @filipefurtad0

PR #4299, opened 2019-09-22
Remove unused field orders.shipping_method_id

luisramos0:
#### What? Why?

Dead field since v2...

#### What should we test?
- Sanity check of checkout related to shipping methods selection.
- Updating old orders and make sure shipping method selection is preserved.
- Verify that Subscription Orders get the correct subscription shipping method

#### Release notes
Changelog Category: Removed
Remove unused database field: order.shipping_method.id.

### DIFF
--- a/db/migrate/20190922201034_drop_orders_shipping_method_id.rb
+++ b/db/migrate/20190922201034_drop_orders_shipping_method_id.rb
@@ -1,0 +1,9 @@
+class DropOrdersShippingMethodId < ActiveRecord::Migration
+  def up
+    remove_column :spree_orders, :shipping_method_id
+  end
+
+  def down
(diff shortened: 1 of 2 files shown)


kristinalim:
@luisramos0 This needs a rebase.

RachL:
Here are the test I've carried:

1. Sanity check of checkout related to shipping methods selection. => I've used Bear farm on Katuma staging, both pick-up and delivery shipping methods are working fine at checkout.

2. Updating old orders and make sure shipping method selection is preserved. => I've updated quantity, products and addresses. All seemed fine.

3. Verify that Subscription Orders get the correct subscription shipping method => They do !

This is ready to go :)

luisramos0:
thank you Rachel!

PR #4303, opened 2019-09-23
Add missing translation for order.shipment.state "on hand" (with a space between on and hand)

### LINKED ISSUE #4302: Error on orders page -  string translation?
## Description

The order page is showing an error in the quantity field since the latest release.
huron orders error

## Expected Behavior

the field should report the quantity ordered

## Actual Behaviour

field has error as shown

## Steps to Reproduce

1.place an order 
2.open the order to edit,  see the error
3.
4.

## Context

NOTE -  I'm seeing this order in a hub I'm just setting up.  Its a food bank,  so the prices of the products are $0.    Is it possible that a 0 priced product is causing this error?    Or as the error suggests is this a missing translation string?

## Your Environment

* Version used:
* Browser name and version:

Happening with the new 2.5.0  deployment today.

luisramos0:
#### What? Why?

Closes #4302
Adds the missing translation key.

#### What should we test?
Verify the issue is fixed now.

#### Release notes
Changelog Category: Added
Added missing translation.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -278,6 +278,7 @@ en:
   maestro_or_solo_cards: "Maestro/Solo cards"
   backordered: "Backordered"
   on_hand: "On Hand"
+  "on hand": "On Hand"
   ship: "Ship"
 
   actions:


luisramos0:
I missed @Matt-Yorkley (because I dont know the answer) this is our key. I dont understand how this is happening but it does fix the problem.

filipefurtad0:
Test performed:
On staging-ES, performed a new order. Error described in issue #4302 is fixed, translation "On Hand" now appears:

PR #4305, opened 2019-09-23
Make translations in new product page relative and specific to the page

### LINKED ISSUE #4276: A translation has disappeared in new product creation and create confusion in user's mind
## Description
The field to put the nb of units have lost the previous translation with Spree upgrade.
On French production :
- before (from image in user guide) :

- now (current production)

I was doing support to a user who had put the price in this field, because "montant" (new automatic translation from spree I guess) means more like price in French... where we expect nb of units. It is pretty confusing and induces users mistakes.

This string needs to be translatable again so we can put the right translation.

## Severity
bug-s3

## Your Environment

* Version used: 2.4.1

luisramos0:
#### What? Why?

Closes #4276

Make translations in new product page lazylookups and specific to this page.

#### What should we test?
The new product page should render correctly in English. Translations can only be tested afterwards.

#### Release notes
Changelog Category: Added
Added missing translation.

### DIFF
--- a/app/views/spree/admin/products/new.html.haml
+++ b/app/views/spree/admin/products/new.html.haml
@@ -3,17 +3,17 @@
 = form_for [:admin, @product], :html => { :multipart => true } do |f|
   .twelve.columns.alpha
     %fieldset.no-border-bottom{ id: "new_product" }
-      %legend{align: "center"}= t(:new_product)
+      %legend{align: "center"}= t(".new_product")
       .twelve.columns.alpha
         .six.columns.alpha
           = f.field_container :supplier do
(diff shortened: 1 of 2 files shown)


filipefurtad0:
Hey @luisramos0,
The PR changes the target field and other fields also, but somehow not all of them. "Unité Affichéé" remains unafected by the fix. Is this the desired/expected behaviour? Moveing to Ready to Go.

Before the fix:
#4305_fr_before

After the fix:
#4305_fr_after

luisramos0:
looks good, display as (unité affiche) and product category are in other view files and were already relative keys, so they remain as they were: correct :+1:
thanks @filipefurtad0 🎉

PR #4311, opened 2019-09-24
Replace deprecated validates_presence_of

mkllnk:
#### What? Why?

Related to #3708.

Before we upgrade to Rails 4 we should remove all usages of deprecated code in Rails 3. This validation syntax is the only deprecation I found so far:
https://guides.rubyonrails.org/3_0_release_notes.html#patches-and-deprecations

#### What should we test?

- Create a subscription.
- Try to omit required fields like the address or the shop.
- It should not proceed when required fields are missing.
- Creating a subscription with all required fields should succeed.

#### Release notes

Changed: Updated the syntax of our code to prepare for the next upgrades.

Changelog Category: Changed

This may help during the next Spree upgrade.

### DIFF
--- a/app/services/subscription_validator.rb
+++ b/app/services/subscription_validator.rb
@@ -9,8 +9,8 @@ class SubscriptionValidator
 
   attr_reader :subscription
 
-  validates_presence_of :shop, :customer, :schedule, :shipping_method, :payment_method
-  validates_presence_of :bill_address, :ship_address, :begins_at
+  validates :shop, :customer, :schedule, :shipping_method, :payment_method, presence: true
+  validates :bill_address, :ship_address, :begins_at, presence: true
   validate :shipping_method_allowed?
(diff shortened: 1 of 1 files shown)


mkllnk:
That's weird. I can't imagine it's a race condition in a model spec. And the second built failed at the same point. It could be about the test execution order which is determined by Knapsack. I ran it locally with Knapsack and it passed... :woman_shrugging:

I'll re-run Semaphore again and we will see if it's flaky.

sigmundpetersen:
Green now ✅

RachL:
This is ready to go!

The test I did:

- Created a new enterprise (with product and open OC)
- Created a subscription.
- I've tried to omit fields on each steps
- Each time I was blocked by a friendly message within the UI
- Once I've field all the required fields, my subscription was set up correctly and order went through 🎉 

Some examples:

image

image

PR #4321, opened 2019-09-25
Delete old pt-BR translations file (it's now correctly called pt_BR)

luisramos0:
#### What? Why?

This removes an old file.

#### What should we test?
Nothing.

#### Release notes

Changelog Category: Fixed
Removed old translations file pt-BR, it's now called pt_BR.


sigmundpetersen:
`pt_BR.yml` was added in #4294 actually, so this should be changed to removal of `pt-BR.yml`, or in a new PR

Issue #4042, opened 2019-07-15
Shop product list pagination


daniellemoorhead:
### Task
- Add pagination to the shop products list, so that only a small number of products display initially in the shop and the user interacts with the page to display more.

- **STILL TO BE DETERMINED:** What is the best design for implementing this? Do we make it an auto-magical display of the next part of the list when they scroll to the bottom? Or traditional 1-2-3-4-etc page buttons and/or next-previous buttons? Or just a load more button? And what happens if they go away and then come back to the page, do we take that user back to the first part of the list, or can we remember where they were?

### Acceptance Criteria
As a user
When I first go to a shop
Then I will see a truncated list of products
And ...?

### Screenshots
coming once solution design completed.

Matt-Yorkley:
I think #4040 and #4041 will have to be done at the same time as this, they can't be separated.

PR #4271, opened 2019-09-18
Remove logic related to master variants in Exchanges and migrate data

luisramos0:
#### What? Why?
Closes #4251 

Master variants cannot be added to Exchanges so this code was catering for a situation that never happens now and impacting performance. We remove the code in this PR and run a migration to ensure old data is also correct by replacing master variants with a standard variant.

I tested the migration on local copies of both AUS and UK live databases. It works.

#### What should we test?
We need to test adding variants to Order Cycles and then seeing them in the shopfront.
Because we are just removing code, I don't think there's nothing else that needs verification.

EDITED: this code change was done in another PR, here we just need to see that the data migration works in staging. that's it.

#### Release notes
Changelog Category: Changed
Improved performance of shopfront/products list by removing unnecessary code.

PR #4323, opened 2019-09-25
WIP Build an API endpoint to list products in the shopfront (move code from shop/products to api/products/shop)

luisramos0:
#### What? Why?

Relates to #4042 

#### What should we test?
TODO

#### Release notes
Changelog Category: Added
Added new API endpoint to list products in a given shop

### DIFF
--- a/app/controllers/api/products_controller.rb
+++ b/app/controllers/api/products_controller.rb
@@ -8,6 +8,30 @@ module Api
 
     skip_authorization_check only: [:show, :bulk_products, :overridable]
 
+    def bulk_products
+      product_query = OpenFoodNetwork::Permissions.new(current_api_user).
+        editable_products.merge(product_scope)
+
+      if params[:import_date].present?
+        product_query = product_query.imported_on(params[:import_date]).group_by_products_id
+      end
+
+      @products = product_query.order('created_at DESC').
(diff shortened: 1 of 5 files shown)


Matt-Yorkley:
I added a quick comment in #4271 that's relevant to this API endpoint. It should help a lot here.

What do you think to moving in this direction (filter and paginate before scoping variants):
https://github.com/openfoodfoundation/openfoodnetwork/compare/master...Matt-Yorkley:shop_endpoint

app/controllers/api/products_controller.rb
```diff
@@ -8,6 +8,42 @@ class ProductsController < Api::BaseController
 
     skip_authorization_check only: [:show, :bulk_products, :overridable]
 
+    def shop
+      renderer = ProductsRenderer.new(current_distributor, current_order_cycle)
+      products = ProductsFilterer.new(renderer.products).call
```
Matt-Yorkley: The method on the renderer is `#products_json`

luisramos0: it was json... now that we dont have cache I dont think we should waste time parsgin and unparsing json... some work to do here.

luisramos0:
closing, this was done in #4334

PR #4324, opened 2019-09-25
Reports memoization

### LINKED ISSUE #4316: Use Permissions from controller for Orders and Fulfillment report and memoize methods

The `OpenFoodNetwork::Permissions` object has some expensive methods like `editable_line_items` which is called more than once when generating different types of the Orders and Fulfillment report.

We can memoize some of these method results to make minor improvements to performance, not just this report.

Additionally, the `Permissions` object created and used in the controller for `#orders_and_fulfillment` can be passed to `OrdersAndFulfillmentsReport` to better take advantage of the memoization.

### Context

This was created in relation to the spike #4130 for the Orders and Fulfillment report.

### Impact and timeline

This is expected to have minimal impact to the report. But there could be a reduction of a couple of queries here and for some other actions.

Matt-Yorkley:
#### What? Why?

Closes #4316

Passes the (already loaded) Permissions object into the `OrdersAndFulfillmentsReport` and adds some memoization.

#### What should we test?

Orders and fulfilment reports should work as before, but faster.

#### Release notes

Improved orders and fulfilment report performance with memoization

Changelog Category: Changed

### DIFF
--- a/lib/open_food_network/orders_and_fulfillments_report.rb
+++ b/lib/open_food_network/orders_and_fulfillments_report.rb
@@ -6,9 +6,10 @@ include Spree::ReportsHelper
 module OpenFoodNetwork
   class OrdersAndFulfillmentsReport
     attr_reader :params
-    def initialize(user, params = {}, render_table = false)
+    def initialize(user, permissions, params = {}, render_table = false)
       @params = params
       @user = user
+      @permissions = permissions
       @render_table = render_table
     end
 
@@ -288,6 +289,8 @@ module OpenFoodNetwork
 
     private
 
+    attr_reader :permissions
+
     def supplier_name
       proc { |line_items| find_variant(line_items.first.variant_id).product.supplier.name }
     end
@@ -302,11 +305,6 @@ module OpenFoodNetwork
 
     def line_items_name
       proc { |line_items| find_variant(line_items.first.variant_id).full_name }
-    end
-
-    def permissions
-      return @permissions unless @permissions.nil?
-      @permissions = OpenFoodNetwork::Permissions.new(@user)
     end
 
     def total_units(line_items)
--- a/lib/open_food_network/orders_and_fulfillments_report.rb
+++ b/lib/open_food_network/orders_and_fulfillments_report.rb
@@ -6,9 +6,8 @@ include Spree::ReportsHelper
 module OpenFoodNetwork
   class OrdersAndFulfillmentsReport
     attr_reader :params
-    def initialize(user, permissions, params = {}, render_table = false)
+    def initialize(permissions, params = {}, render_table = false)
       @params = params
-      @user = user
       @permissions = permissions
       @render_table = render_table
     end
--- a/lib/open_food_network/reports/line_items.rb
+++ b/lib/open_food_network/reports/line_items.rb
@@ -12,11 +12,11 @@ module OpenFoodNetwork
         line_items = permissions.visible_line_items.merge(Spree::LineItem.where(order_id: orders))
         line_items = line_items.supplied_by_any(params[:supplier_id_in]) if params[:supplier_id_in].present?
 
-        # If empty array is passed in, the where clause will return all line_items, which is bad
-        line_items_with_hidden_details =
-          permissions.editable_line_items.empty? ? line_items : line_items.where('"spree_line_items"."id" NOT IN (?)', permissions.editable_line_items)
+        hidden_line_items = line_items_with_hidden_details(permissions, line_items)
 
-        line_items.select{ |li| line_items_with_hidden_details.include? li }.each do |line_item|
+        line_items.select{ |li|
+          hidden_line_items.include? li
+        }.each do |line_item|
           # TODO We should really be hiding customer code here too, but until we
           # have an actual association between order and customer, it's a bit tricky
           line_item.order.bill_address.andand.assign_attributes(firstname: I18n.t('admin.reports.hidden'), lastname: "", phone: "", address1: "", address2: "", city: "", zipcode: "", state: nil)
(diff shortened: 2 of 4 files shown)


Matt-Yorkley:
I picked an arbitrary date range and active hub in my dev environment for testing. The queries dropped from 168 -> 136 -> 131 -> 79.

**Before:**
Screenshot from 2019-09-25 19-49-37
**After first commit:**
Screenshot from 2019-09-25 19-50-45
**After second commit:**
Screenshot from 2019-09-25 20-07-58
**After third commit:**
Screenshot from 2019-09-25 20-14-23

lib/open_food_network/reports/line_items.rb
```diff
@@ -12,9 +12,14 @@ def self.list(permissions, params)
...
+        editable_line_items = permissions.editable_line_items.pluck(:id)
+
         # If empty array is passed in, the where clause will return all line_items, which is bad
-        line_items_with_hidden_details =
-          permissions.editable_line_items.empty? ? line_items : line_items.where('"spree_line_items"."id" NOT IN (?)', permissions.editable_line_items)
+        line_items_with_hidden_details = if editable_line_items.empty?
+                                           line_items
+                                         else
+                                           line_items.where('"spree_line_items"."id" NOT IN (?)', editable_line_items)
+                                         end
```
luisramos0: this can be extracted to method right? line_items_with_hidden_details

lib/open_food_network/reports/line_items.rb
```diff
@@ -23,7 +29,7 @@ def self.list(permissions, params)
           line_item.order.ship_address.andand.assign_attributes(firstname: I18n.t('admin.reports.hidden'), lastname: "", phone: "", address1: "", address2: "", city: "", zipcode: "", state: nil)
           line_item.order.assign_attributes(email: I18n.t('admin.reports.hidden'))
         end
-        line_items
+        @list = line_items
```
kristinalim: :warning:

This caches the line items, which is based on the `permissions` argument, in the class. So if we pass a `Permissions` object for a different user in the same thread, the result for the first one would be used.
   
   ```ruby
class MyClass
  def self.my_method
    return @result unless @result.nil?
    @result = Time.now
  end
end

MyClass.my_method
# => 2019-09-26 13:09:07 +0800
sleep 3
MyClass.my_method
# => 2019-09-26 13:09:07 +0800
```

Unicorn would also retain this result in memory throughout the life of the worker, which would be a memory leak.

File: lib/open_food_network/orders_and_fulfillments_report.rb [hunk 3]
```diff
@@ -6,9 +6,10 @@
 module OpenFoodNetwork
   class OrdersAndFulfillmentsReport
     attr_reader :params
-    def initialize(user, params = {}, render_table = false)
+    def initialize(user, permissions, params = {}, render_table = false)
```
kristinalim: Do we still use this `user` argument?

Matt-Yorkley: Yeah, it's used twice.

lib/open_food_network/reports/line_items.rb
Matt-Yorkley: Oops! Yes, I just saw the build.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -47,12 +48,12 @@ def header
...
 
     def search
-      Reports::LineItems.search_orders(permissions, params)
+      Reports::LineItems.search_orders(@permissions, params)
     end
 
     def table_items
       return [] unless @render_table
-      Reports::LineItems.list(permissions, params)
+      Reports::LineItems.list(@permissions, params)
```
sauloperez: you wouldn't need this change if we were using ruby's `attr_reader` which I strongly recommend due to the encapsulation and other improvements it provides over ivars.

lib/open_food_network/reports/line_items.rb
```diff
@@ -25,6 +23,16 @@ def self.list(permissions, params)
...
         line_items
       end
+
+      def self.line_items_with_hidden_details(permissions, line_items)
+        editable_line_items = permissions.editable_line_items.pluck(:id)
+
+        if editable_line_items.empty?
+          line_items
+        else
+          line_items.where('"spree_line_items"."id" NOT IN (?)', editable_line_items)
```
sauloperez: :warning: this will cause an N+1. We'll need to weight the benefits of leveraging the permissions memoization versus introducing these N queries to see if it's really worth shipping it as it is...

If I'm not mistaken the call to `Enumerable#select` on line 15 will execute the AR built up until then, which in turn will cause the AR above to be executed for every step of the iteration. We should be able to INNER JOIN `line_items` with `editable_line_items` before that and make PostgreSQL return only the data we care about. Do you see any problems on that?

lib/open_food_network/orders_and_fulfillments_report.rb
Matt-Yorkley: Ah, it's not used! I'll remove it

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -288,6 +288,8 @@ def columns
 
     private
 
+    attr_reader :permissions
```
sauloperez: :heart:

sauloperez:
Moving to Test Ready as @kristinalim 's comment was also addressed.

filipefurtad0:
Hi @Matt-Yorkley :-)

I've tested this on Katuma using a Superadmin account, in order to generate heavier reports. Everything worked fine, but I measured no significant change in load times. I've looked at the Load value (in red) on the Chrome Console, and obtained around 10 sec, both before and after staging the PR. Perhaps the data is still not big enough for a measurable effect...

Would this be a good approach in testing this PR? (Rookie-question :-)

#4324_kat_a

Matt-Yorkley:
As long as its working, I think its OK. There are lots of factors contributing to load times here, and this is only a small piece :±1:

filipefurtad0:
Ok, thanks! Moving it to Ready to Go :+1:

Issue #4308, opened 2019-09-24
Eager-load obvious associations used in OrdersAndFulfillmentsReport


kristinalim:
### What we should change and why (this is tech debt)

* Supplier Totals - variant, product, supplier
* Supplier Totals by Distributor - order, variant, product
* Distributor Totals by Supplier - order, variant, product
* Customer Totals - order, bill address, variant, product - **Added:** shipment

We are adding _n_ queries for **each line item** involved in the report. Eager-loading, we can significantly reduce the number of queries and the response time.

Query example:

![20190924152204-eager_load_order_and_distributor](https://user-images.githubusercontent.com/2243/65491402-8e6c9780-dee1-11e9-90b1-4b35f94585b4.png)

Stack trace:

![20190924153908-eager_load_order_stack_trace](https://user-images.githubusercontent.com/2243/65491461-a9d7a280-dee1-11e9-9b9b-3e0e8ed0b72c.png)

### Context

This was created in relation to the spike #4130 for the Orders and Fulfillments report.

### Impact and timeline

This is expected to significantly reduce the response time.

sauloperez:
We should stop seeing traces like https://app.datadoghq.com/apm/trace/4523883766251097946?colorBy=service&env=production&graphType=flamegraph&routing_hint=1569838147022&shouldShowLegend=true&sort=time&spanID=3279947065713045773 in Datadog.

PR #4325, opened 2019-09-26
4312 Use association instead of OrderAndFulfillmentsReport#find_variant

### LINKED ISSUE #4312: find_variant in OrdersAndFulfillmentsReport does not cache result and does not use LineItem#variant

There are multiple calls for `find_variant` per line item. The existing `LineItem#variant` is already unscoped, so we can use that.

```ruby
# Returns the variant with the given id, including soft-deleted ones
def find_variant(id)
  Spree::Variant.unscoped.find(id)
end
```

`variant` can then be eager-loaded, to further reduce the number of SQL queries.

All report subtypes use this method.

### Context

### Impact and timeline

This is expected to have significant effect on the response time, and will significantly reduce the number of SQL queries.

kristinalim:
#### What? Why?

- Closes #4312 

There is some reduction in response time and number of SQL queries for some report subtypes. The effect will be greater when we already have #4308 which would eager-load variants.

#### What should we test?

Smoke test for generating the Orders and Fulfillment report subtypes. Just a quick test that there are no obvious issues when generating and filtering each subtype.

#### Release notes

In the Orders and Fulfillment reports, use the Rails association for fetching a line item's variant, to reduce the number of SQL queries needed and improve performance.

Changelog Category: Changed

### DIFF
--- a/lib/open_food_network/orders_and_fulfillments_report.rb
+++ b/lib/open_food_network/orders_and_fulfillments_report.rb
@@ -60,30 +60,30 @@ module OpenFoodNetwork
       when "order_cycle_supplier_totals"
         [
           {
-            group_by: proc { |line_item| find_variant(line_item.variant_id).product.supplier },
+            group_by: proc { |line_item| line_item.variant.product.supplier },
             sort_by: proc { |supplier| supplier.name }
           },
           {
-            group_by: proc { |line_item| find_variant(line_item.variant_id).product },
+            group_by: proc { |line_item| line_item.variant.product },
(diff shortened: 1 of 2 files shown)


filipefurtad0:
Hi @kristinalim :-)

Performed smoke test on the reports section, namely generating the Orders and Fulfillment report subtypes. Also had a look at order cycles, bulk orders etc. Everything seemed fine and pretty much as before, nothing to report - which is a good sign I guess :+1:
Moving PR to Ready to Go.

Issue #3731, opened 2019-04-11
Missing translation on order cycle loader


RachL:
## Description
When you wait for the OC to load, the loader is not translated

## Steps to Reproduce
1. Log in as a producer or a hub or a super admin
2. Go to /admin/order_cycles
3. You will see a loader before the order cycles appear
4. It is not translated

## Animated Gif/Screenshot

![image.png](https://images.zenhubusercontent.com/595e669a61b4ca3b0a737135/870168df-842d-48f1-8535-49d449bf768a)

## Severity
bug-s3: a feature is broken but there is a workaround

## Your Environment

* Version used: v1.29
* Browser name and version: Firefox 66
* Operating System and version (desktop or mobile): Ubuntu 18.10

PR #4327, opened 2019-09-26
4315 Set default date range for Orders and Fulfillment report

### LINKED ISSUE #4315: Pre-fill Orders and Fulfillments report date filters with start and end dates

Mistakenly running the report for all-time when the user needs data only for a shorter period would unnecessarily hog the Unicorn worker for _n_ more seconds/minutes. This small UI change is a small effort that could avoid that.

Pre-fill the start and end date fields with reasonable values, to encourage the user to set the period they need so they don't generate a report for all-time unless they actually need that.

### Context

### Impact and timeline

This would reduce unnecessary spikes in response time, and improve user experience too.

kristinalim:
#### What? Why?

- Closes #4315

Pre-fill the start and end date fields with default values, to encourage the user to set the date range they need. This is to avoid having users generate a report for all-time if they don't actually need that data.

The date range this PR uses is:

* Beginning of day, 1 year ago
* Beginning of day, today

When selecting a date and time in the datetime JS picker, the format of the date is e.g. `2019-09-26 00:00`. After submitting the form, the format becomes `2019-09-26 00:00:00 +1000`. This PR pre-fills the field with the longer datetime format. Best if these three states use the same format, but that is outside the scope of this PR.

I considered moving this inside `OrdersAndFulfillmentsReport`, but I think that's premature until we have another report using default filters.

#### What should we test?

Test that the date range fields are pre-filled with:

* Beginning of day, 1 year ago
* Beginning of day, today

Smoke test that the date range filter still works, and test that selecting another date range is not overwritten when submitting the form.

#### Release notes

- Set a default date range for the Orders and Fulfillment report.

Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/reports_controller_decorator.rb
+++ b/app/controllers/spree/admin/reports_controller_decorator.rb
@@ -109,7 +109,7 @@ Spree::Admin::ReportsController.class_eval do
   end
 
   def orders_and_fulfillment
-    params[:q] ||= {}
+    params[:q] ||= orders_and_fulfillment_default_filters
 
     # -- Prepare Form Options
     permissions = OpenFoodNetwork::Permissions.new(spree_current_user)
@@ -276,5 +276,11 @@ Spree::Admin::ReportsController.class_eval do
 
   def timestamp
     Time.zone.now.strftime("%Y%m%d")
+  end
+
+  def orders_and_fulfillment_default_filters
+    now = Time.zone.now
+    { completed_at_gt: (now - 1.month).beginning_of_day,
+      completed_at_lt: (now + 1.day).beginning_of_day }
   end
 end


app/controllers/spree/admin/reports_controller_decorator.rb
```diff
@@ -277,4 +277,10 @@ def url_for_report(report)
   def timestamp
     Time.zone.now.strftime("%Y%m%d")
   end
+
+  def orders_and_fulfillment_default_filters
+    now = Time.zone.now
+    { completed_at_gt: (now - 1.month).beginning_of_day,
+      completed_at_lt: (now + 1.day).beginning_of_day }
+  end
```
sauloperez: I'd be conservative and limit it to 1 week only. I haven't checked the numbers but a month can surely cause :boom: in many cases which would lead to a poor experience for the user.

luisramos0: it's a requirement here:
https://github.com/openfoodfoundation/openfoodnetwork/issues/4315#issuecomment-535428200

@lin-d-hop would "1 week ago" be ok as default start date?

luisramos0: ping @lin-d-hop @RachL

sigmundpetersen: there's also recent related discussion in the issue (#4315)

kristinalim:
@sauloperez Are you happy for this to be moved to Test Ready already?

For reference: https://github.com/openfoodfoundation/openfoodnetwork/issues/4315#issuecomment-540953001

sauloperez:
I'm happy with the investigation you did data-driven for the win!

sorry about the timing. I need to tune my GH notification settings :see_no_evil:

filipefurtad0:
Hi @kristinalim,

Smoke tested, all seems ok. Default date is correctly set to 1 month, in accordance to the code posted above. The shown +2000 default time is due to time change on my time zone, last Sunday.

Screenshot at 2019-10-29 10-39-58

I noticed the "END" didn't get translated. It's a minor detail, should I open an issue, on such cases?
Moving to Ready to Go.

luisramos0:
yes Filipe, new issue. 
end is a missing translation not in en.yml

RachL:
@filipefurtad0 nice catch! Yes please open the issue :) you can mirror this one in terms of label etc https://github.com/openfoodfoundation/openfoodnetwork/issues/3731

filipefurtad0:
Thanks @RachL and @luisramos0 :-) Issue opened :tada: 
https://github.com/openfoodfoundation/openfoodnetwork/issues/4412

PR #4330, opened 2019-09-26
Update bug template

RachL:


sigmundpetersen:
I have been wondering, should we include a section for a Workaround if there is one? We often discuss this in the issues, and it's also nice to have for reference until the underlying issue is fixed.

RachL:
@sigmundpetersen very good idea! I've added a commit to the file

Issue #4040, opened 2019-07-15
Shop product list search changes to server-side call


daniellemoorhead:
### Task
- Change the product search function on shops to be a server-side call, so that search will continue to work when pagination is added (#4042).

- Note: this will remove the 'see-results-as-you-type' aspect that the front-end call allows for now. **Server-side real-time results are out of scope** (unless the dev doing this work can say it's an easy addition and won't add complexity and/or excessive time to complete the work.

### Acceptance Criteria
As a user
When I type a search term into the product search field
And hit enter
Then the page will reload with a list of the search results
And will display the relevant categories
And will give me the option to clear the current search
And will allow me to do a different search

### Screenshots
<img width="953" alt="Screen Shot 2019-07-15 at 4 24 42 pm" src="https://user-images.githubusercontent.com/10949052/61224088-387bb680-a71e-11e9-968e-92521dff9c7a.png">

PR #4334, opened 2019-09-27
Products renderer

Matt-Yorkley:
#### What? Why?

Related to #4040 

Removes some big N+1s and makes it possible to use pagination and filtering on the results shown in the shop/products endpoint. `Openfoodnetwork::ProductsRenderer` can now have `ransack` and `kaminari` methods chained into the main query to add filtering and pagination (without N+1 troubles or incorrect stock levels due to scoping the variants after applying the pagination.)

Also incorporates #4271 plus proposed changes to remove `OrderCycleDistributedProducts` completely. :fire:

#### What should we test?

Shop products load correctly, including variant overrides.

#### Release notes

Improved `ProductsRenderer` query for fetching shop products

Changelog Category: Changed


Matt-Yorkley:
I still need to add some specs for the new service, but basically we can now paginate and filter a shop's products without creating a total mess. 

E.g. we can change this:
```ruby
    def load_products
      return unless @order_cycle

      ShopProductsService.new(@distributor, @order_cycle).relation.
        order(taxon_order).
        each { |product| scoper.scope(product) }
    end
```

to this:

```ruby
    def load_products
      return unless @order_cycle

      ShopProductsService.new(@distributor, @order_cycle).relation.
        order(taxon_order).
        ransack(params[:q]).result.page(params[:page]).per(params[:per_page]) # yay!
        each { |product| scoper.scope(product) }
    end
```

Matt-Yorkley:
Using an OC with ~500 products. Baseline number of queries: **5492**.
Screenshot from 2019-09-29 16-22-59

After working on product query: **3965**.
Screenshot from 2019-09-29 16-26-37

After working on variant query: **3252**.
Screenshot from 2019-09-29 16-37-20

We can probably get these queries even lower, but it will improve greatly when we are paginating these results to 10 products at a time instead of all 500 of them at once.

app/services/shop_products_service.rb
```diff
@@ -0,0 +1,76 @@
...
+      where(query_stock_with_overrides)
+  end
+
+  def query_stock_with_overrides
+    "( #{variant_not_overriden} AND ( #{variant_on_demand} OR #{variant_in_stock} ) )
+      OR ( #{variant_overriden} AND ( #{override_on_demand} OR #{override_in_stock} ) )
+      OR ( #{variant_overriden} AND ( #{override_on_demand_null} AND #{variant_on_demand} ) )
+      OR ( #{variant_overriden} AND ( #{override_on_demand_null}
+                                      AND #{variant_not_on_demand} AND #{variant_in_stock} ) )"
+  end
```
sauloperez: we'll need to take a close it to this particular bit of the query. Definitely faster than N+1 though.

app/services/shop_products_service.rb
```diff
@@ -0,0 +1,76 @@
...
+# The stock-checking includes on_demand and stock level overrides from variant_overrides.
+
+class ShopProductsService
+  def initialize(distributor, order_cycle)
+    @distributor = distributor
+    @order_cycle = order_cycle
+  end
+
+  def products_relation
+    Spree::Product.where(id: stocked_products)
```
sauloperez: having to fetch the product rows from the ids collected in `stocked_products` seems a waste of resources to me. We can start building the AR from `Spree::Product` and group by `spree_variants.product_id` and we will avoid unnecessary code we'll need to maintain, wasted Ruby GC time, slow Ruby processing on the product ids and instantiating bloated ActiveModel objects plus the network round trip, among others.

I suggest we create an issue to fix it in the next sprint.

lib/open_food_network/products_renderer.rb
```diff
@@ -29,17 +27,20 @@ def products_json
...
-        relation.
-        order(taxon_order).
-        each { |product| scoper.scope(product) }.
-        select do |product|
-          !product.deleted? && product.has_stock_for_distribution?(@order_cycle, @distributor)
-        end
+      @shop_products ||= begin
+        scoper = ScopeProductToHub.new(@distributor)
+
+        shop_products_service.products_relation.
```
sauloperez: I find a bit confusing the fact that this first method is not in its own line like the methods below. Well, I admit it's also TOC :trollface:

lib/open_food_network/products_renderer.rb
```diff
@@ -53,25 +54,23 @@ def taxon_order
...
-                                   Spree::Variant.
-                                     for_distribution(@order_cycle, @distributor).
-                                     each { |v| scoper.scope(v) }.
-                                     select(&:in_stock?)
-                                 end
+    def variants_for_shop
+      @variants_for_shop ||= begin
+        scoper = OpenFoodNetwork::ScopeVariantToHub.new(@distributor)
+
+        shop_products_service.variants_relation.
```
sauloperez: See my comment above about a  method per line.

app/services/shop_products_service.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def stocked_variants_and_overrides
+    Spree::Variant.
+      joins("LEFT OUTER JOIN variant_overrides ON variant_overrides.variant_id = spree_variants.id
+            AND variant_overrides.hub_id = #{@distributor.id}").
+      joins(:stock_items).
+      where(query_stock_with_overrides)
+  end
+
+  def query_stock_with_overrides
```
sauloperez: I guess this eager loads data that would have been loaded with N+1 by `ScopeVariantToHub` otherwise?

Matt-Yorkley: Yes, it includes the post-scoping stock check in the main query so that it works when paginating results. Otherwise if we fetch 10 products first, then scope, then check stock with overrides and remove the out-of-stock products, the paginated results might return 4 or 7, or whatever, instead of 10. Does that make sense?

app/services/shop_products_service.rb
```diff
@@ -0,0 +1,76 @@
+# Returns a (paginatable) AR object for the products or variants in stock for a given shop and OC.
+# The stock-checking includes on_demand and stock level overrides from variant_overrides.
+
+class ShopProductsService
```
luisramos0: I dont think this name is appropriate. Shop is not a clear concept in OFN, if we use this concept we need to define it first.
I'd prefer for example to keep the previous name OrderCycleDistributedProducts, it makes it very clear what product list this is.
I dont think we can be casual about naming otherwise we end up with things like what we have already in the DB with supplier_id, enterprise_id, distributor_id, hub_id, producer_id and... shop_id (it's a column in the subscription table).

luisramos0: or we can call it EnterpriseProductsService

lib/open_food_network/products_renderer.rb
Matt-Yorkley: TOC? Yeah, I put the service and service method call on the same line, then additional methods that alter the service's output on separate lines. I can change it if you like.

app/services/shop_products_service.rb
Matt-Yorkley: OK, I think OrderCycleDistributedProducts is better than EnterpriseProducts.

app/services/shop_products_service.rb
Matt-Yorkley: It's still a single AR object right? We're not making an extra query here.

app/services/shop_products_service.rb
Matt-Yorkley: Changed to `OrderCycleDistributedProducts` :+1:

lib/open_food_network/products_renderer.rb
sauloperez: sorry lost in langauges, TOC == OCD. I find it simpler all methods in their line but that's just my brain.

app/services/shop_products_service.rb
sauloperez: You are saying this triggers a singler query where `stocked_products` becomes a subquery? Have you checked if that is the case?

Matt-Yorkley: Yes, that's the beauty of it! All the methods here only build AR objects with subqueries. This test passes, for example:

```ruby
    describe "stocked_products" do
      it "returns an AR object" do
        stocked_products = described_class.new(distributor, order_cycle).send(:stocked_products)
        expect(stocked_products).to be_an ActiveRecord::Relation
      end
    end
```

sauloperez: awesome!

RachL:
These are the tests I carried:

- Filters and search on an existing and new enterprise
- Changing price and stock through inventory (keeping producer settings or not): everything looked fine. I've did it on products with several variants and product with only one variant
- Checkout is working fine
- the number max of variant in a shop I've tested this with is around 50 products

I think we are good to go!

luisramos0:
@Matt-Yorkley 👏

sauloperez:
Yay! So looking forward to see this in production 😍

PR #4340, opened 2019-09-30
Delete ISSUE_TEMPLATE.md

sigmundpetersen:
#### What? Why?
This ISSUE_TEMPLATE.md file is the old way of handling templates on Github. OFN is now using the new method for handling templates, so this file can be removed. 
This template still comes up as 'Default template' when opening an issue from the Zenhub board. This gives 2 almost equal template choices which is confusing.

My first PR 🎉 🙂


sauloperez:
Congrats @sigmundpetersen!

PR #4342, opened 2019-10-01
makes edit button action open a new tab

### LINKED ISSUE #4338: Product list: enable action button to be opened in a new tab
## Description
Currently you cannot open the edit product button in a new tab.

This means that when you have to update several products (for example updating the tax category), you need to open the edit page and go back to the full list of products before you can select another product to edit. 
If you had used filters, all filters are removed when reloading the page. While we are not tackling this behavior in this issue, it means it is difficult to see which product you have already edited vs. which product you didn't.
When opening all of them in several new tabs, you can more easily manage them.

## Steps to Reproduce
1. Go to `/admin/products`
2. Filter the results for one particular producer
3. Try to edit a product using the following icon: 
image
4. See that you cannot open the target page in a new tab
5. Open the edit page, then go back to `/admin/products`. Notice all your filter are lost (which is let's say for now a correct behavior)

## Your Environment
* Version used: v2.5
* Browser name and version: All browser and version

### LINKED ISSUE #4388: Sample data should include Orders
We should add orders to sample data so that a freshly loaded dev/test environment already includes some orders that can be used for testing.

### Context
The questions was raised here:
https://github.com/openfoodfoundation/openfoodnetwork/issues/4378
and is related to this thread:
https://community.openfoodnetwork.org/t/seed-data-development-provisioning-deployment/910/24

btw, the names of the sample enterprises can be changed according to what was agreed in the thread.

### Impact and timeline
This is a super easy task and will make our sample data a lot better.

### Possible implementation
We need to create OrderFactory under lib/tasks/sample_data and link it in lib/tasks/sample_data.rake

JacksonBates:
#### What? Why?

Closes #4338

This PR changes the `window.location...` action attached to the admin/product edit buttons to use `window.open('...', '_blank')` instead, which has the effect of opening a new tab. 

This was requested as a short term solution to losing page filters when editing a product.

#### What should we test?
Click the edit button for a product on the admin/products page to confirm new tab.

#### Release notes
Updates the edit button action on the admin/products page to open a new tab, in order to preserve filters on the products page.

Changelog Category: Changed

### DIFF
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -110,7 +110,7 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
 
   $scope.editWarn = (product, variant) ->
     if (DirtyProducts.count() > 0 and confirm(t("unsaved_changes_confirmation"))) or (DirtyProducts.count() == 0)
-      window.location = "/admin/products/" + product.permalink_live + ((if variant then "/variants/" + variant.id else "")) + "/edit"
+      window.open("/admin/products/" + product.permalink_live + ((if variant then "/variants/" + variant.id else "")) + "/edit", "_blank")
 
 
   $scope.toggleShowAllVariants = ->
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -108,9 +108,13 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     $scope.categoryFilter = "0"
     $scope.importDateFilter = "0"
 
+  confirm_unsaved_changes = () -> (DirtyProducts.count() > 0 and confirm(t("unsaved_changes_confirmation"))) or (DirtyProducts.count() == 0)
+  
+  editProductUrl = (product, variant) -> "/admin/products/" + product.permalink_live + ((if variant then "/variants/" + variant.id else "")) + "/edit"
+
   $scope.editWarn = (product, variant) ->
-    if (DirtyProducts.count() > 0 and confirm(t("unsaved_changes_confirmation"))) or (DirtyProducts.count() == 0)
-      window.open("/admin/products/" + product.permalink_live + ((if variant then "/variants/" + variant.id else "")) + "/edit", "_blank")
+    if confirm_unsaved_changes()
+      window.open(editProductUrl(product, variant), "_blank")
 
 
   $scope.toggleShowAllVariants = ->
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -108,9 +108,11 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     $scope.categoryFilter = "0"
     $scope.importDateFilter = "0"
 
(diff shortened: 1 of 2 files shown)


JacksonBates:
Thanks for the feedback @luisramos0.

I was a bit hasty admittedly - my 11:30pm brain thought it was a good idea to jump right in!

I'll have a look at your suggestions this evening. I noticed running `rspec spec` just fell over, rather than actually running the tests properly, so I'll need to double check how I set up my environment to get the tests running locally.

luisramos0:
yeah, our specs take almost one hour to run. better run the specs you want to check only:
in this case
rspec spec/features/admin/bulk_product_update_spec.rb

JacksonBates:
@luisramos0 I've got the tests passing locally (getting rspec and selenium working on my machine was a tricky issue in itself).

I'll look at your suggestions for refactoring over the next day or two.

spec/features/admin/bulk_product_update_spec.rb
```diff
@@ -516,11 +516,16 @@
       it "shows an edit button for products, which takes the user to the standard edit page for that product" do
         expect(page).to have_selector "a.edit-product", count: 2
 
-        within "tr#p_#{p1.id}" do
-          find("a.edit-product").click
+        new_window = window_opened_by do 
+          within "tr#p_#{p1.id}" do
+            find("a.edit-product").click
+          end
         end
```
luisramos0: nice trick!
just a thought: is this really necessary?
if the expect URI is being evaluated too early, the usual solution in OFN's code base is to expect some DOM element in the destination page before the required verification, in this case the URL, is done.

JacksonBates: I believe that is necessary, since the feature request was to open a new window.

I suppose we could examine the DOM on the existing page for the presence of a _blank target on the appropriate link, but I think checking for the existence of the new window is a truer test for covering the required behaviour. Although I'm happy to implement either case (or another you can think of) -- you're call.

I do need to change the description of the test, though. I realise I kept the old one, whilst changing what the test looked for. :man_facepalming:

luisramos0: ok, I missed the fact the request is to open a new window, thus the test. It's perfect!

JacksonBates:
I'm happy to refactor editWarn, and I should fix up the description of the two tests I touched before it gets merged.

JacksonBates:
Hi @luisramos0 - apologies for the delay in finishing this up. Just moved house and there were some legal dramas involved :open_mouth:

I've made the changes you suggested if you'd like to review?

Do you have a preference for where these new functions live in the file - I figured just before they were used made it readable, but I'm happy to follow other patterns already in practise in the codebase.

app/assets/javascripts/admin/bulk_product_update.js.coffee
```diff
@@ -108,9 +108,13 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     $scope.categoryFilter = "0"
     $scope.importDateFilter = "0"
 
+  confirm_unsaved_changes = () -> (DirtyProducts.count() > 0 and confirm(t("unsaved_changes_confirmation"))) or (DirtyProducts.count() == 0)
```
luisramos0: I'd break line after -> so that the line is shorter and declaration and implementation are in diff lines.
same for the editProductUrl method

JacksonBates: Done :)

RachL:
Thanks a lot for your contribution @JacksonBates !

I've tested this on a staging server and it looks good :) Among opening in a new tab I've also tested:
- product creation
- bulk and single edit
- cloning
- deleting

Both on super admin and admin

This is ready to go :)

PR #4345, opened 2019-10-02
Shopfront pagination

### LINKED ISSUE #4040: Shop product list search changes to server-side call

### LINKED ISSUE #4041: Shop product list category filters change to server-side call
### Task
- Change the product category filter function on shops to be a server-side call, so that filters continue to work when pagination is added (#4042).

### Acceptance Criteria
As a user
When I on a product category filter
Then the page will reload with a list of the products in that category
And will give me the option to clear the current filter
And will allow me to filter with a different category

### Screenshots
Screen Shot 2019-07-15 at 4 36 20 pm

### LINKED ISSUE #4042: Shop product list pagination

Matt-Yorkley:
#### What? Why?

Closes #4040 
Closes #4041 
Closes #4042 

Adds pagination to shop product lists and moves search and filtering server-side. Also removes the product cache...

#### What should we test?

Shop products should load more or less as before. Things that need testing:

- Enterprise fees are applied correctly in shop
- Variant overrides are working
- Taxons and Properties buttons are working
- Search and filter products is working
- Products hidden by tag rules are hidden

#### Release notes

Added pagination to shop products list and removed product cache.

Changelog Category: Changed


app/assets/javascripts/darkswarm/services/order_cycle_resource.js.coffee
```diff
@@ -0,0 +1,15 @@
+Darkswarm.factory 'OrderCycleResource', ($resource) ->
+  $resource('/api/order_cycles/:id.json', {}, {
+    'products':
+      method: 'GET'
+      isArray: true
+      url: '/api/order_cycles/:id/products.json'
```
luisramos0: we are killing the .json in the api endpoints.

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,38 @@
+module Api
+  class OrderCyclesController < BaseController
+    respond_to :json
+
+    def products
+      products = OpenFoodNetwork::ProductsRenderer.new(current_distributor, current_order_cycle, params).products_json
```
luisramos0: this is not going to work after #4057 where we merge Api::BaseController and Spree::ApiBaseController (Spree::Core::ControllerHelpers::Order that defines current_order will not be available in the API, which I think is very good).
I dont think we should rely on these global methods. both current_distributor, current_order_cycle they must be taken from params in the controller before being sent to the service.

also, the params variable is something specific to the world of controllers, should never get into the service world. instead of sending params we should get whatever params we need and send them as individual variables to the service. in this case I think that means three separate variables: q, page and per_page

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      render json: products
+    rescue OpenFoodNetwork::ProductsRenderer::NoProducts
+      render status: :not_found, json: ''
+    end
+
+    def filters
```
luisramos0: is there a better way we can do this in terms of API endpoints?
if we keep it we need at least a better name for this endpoint. "filters" is too specific to the shopfront, not directly related to the OrderCycle domain...
is there any reason to keep taxons and properties in the same endpoint? we can probably have two separate endpoints.

/api/order_cycles/:id/taxons - this is not really generic taxons, it's only product categories, so maybe:
/api/order_cycles/:id/categories

and then: /api/order_cycles/:id/properties

I think these are better names ☝️

Matt-Yorkley: Yeah this bit is WIP. Its: "taxons and properties of variants distributed in this order cycle".

Matt-Yorkley:
> we should rush... to avoid a huge PR

This PR will suddenly get a lot smaller when the ProductsRenderer PR is merged ;)

app/services/order_cycle_distributed_products.rb
```diff
@@ -42,6 +114,22 @@ def query_stock_with_overrides
...
+
+  def override_not_hidden_by_rule
+    return "FALSE" unless overrides_to_hide.any?
+    "variant_overrides.id NOT IN (#{overrides_to_hide.join(',')})"
+  end
+
+  def override_shown_by_rule
+    return "FALSE" unless overrides_to_show.any?
+    "variant_overrides.id IN (#{overrides_to_show.join(',')})"
+  end
```
luisramos0: if if the code s coupled with this class, we need to extract this code to a separate class. Maybe TagRuleApplicatorQuery? and inside both override_shown_by_rule and override_not_hidden_by_rule and all sub methods

Matt-Yorkley: Already done ;)

Matt-Yorkley:
Okay, this is getting pretty close.

In regards to the new filters endpoint, I think maybe we should move it to a separate issue.

Basically this PR introduces a small regression, which is that the filter buttons shown in the shop will show the filters for the enterprise, but not scoped to the order cycle. So if a hub has two order cycles; one with `fish`, `bread`, and `milk`, and the other with `fish`, `bread`, and `fruit`, the filter buttons will show `fish`, `bread`, `milk`, and `fruit`.

It's not difficult to fix, but including it here would make the PR substantially larger.

app/services/product_tag_rules_filterer.rb
```diff
@@ -0,0 +1,107 @@
+class ProductTagRulesFilterer
+  def initialize(distributor, customer, variants)
```
luisramos0: I think it's better to call this variants_relation to make it explicit this is a relation

app/services/order_cycle_distributed_products.rb
```diff
@@ -36,74 +36,7 @@ def stocked_variants_and_overrides
...
-    @hide_rules ||= customer_applicable_rules.select{ |rule| rule.preferred_matched_variants_visibility == 'hidden'}
-  end
-
-  def show_rules
-    customer_applicable_rules - hide_rules
-  end
-
-  def customer_tagged?(rule)
-    customer_tag_list.include? rule.preferred_customer_tags
+    ProductTagRulesFilterer.new(distributor, customer, stocked_variants).call
```
luisramos0: 👍

lib/open_food_network/products_renderer.rb
```diff
@@ -27,25 +29,33 @@ def products_json
...
+      return unless order_cycle
 
       @products ||= begin
-        scoper = ScopeProductToHub.new(@distributor)
+        results = distributed_products.products_relation.order(taxon_order)
 
-        distributed_products.products_relation.
-          order(taxon_order).
-          each { |product| scoper.scope(product) }
+        filter_and_paginate(results).each { |product| product_scoper.scope(product) }
```
luisramos0: I think we need a little comment here explaining to the future dumb devs (ourselves next year :-)) looking at this code. Something like "distributed_products is already taking Overrides into account in the query, here we need to scope the products so that the overrides are taken into account on the serialization process that follows"
and btw, is this explanation correct?

app/serializers/api/product_serializer.rb
```diff
@@ -70,4 +38,12 @@ def variants
...
     options[:master_variants][object.id].andand.first
   end
+
+  def price
+    if options[:enterprise_fee_calculator]
+      object.master.price + options[:enterprise_fee_calculator].indexed_fees_for(object.master)
+    else
+      object.master.price_with_fees(options[:current_distributor], options[:current_order_cycle])
+    end
+  end
```
luisramos0: I think this needs a spec in product_serializer_spec

lib/open_food_network/products_renderer.rb
```diff
@@ -4,19 +4,21 @@ module OpenFoodNetwork
   class ProductsRenderer
```
luisramos0: it would be great to see this file flying to app/services

spec/lib/open_food_network/products_renderer_spec.rb
```diff
@@ -6,7 +6,8 @@ module OpenFoodNetwork
     let(:distributor) { create(:distributor_enterprise) }
     let(:order_cycle) { create(:simple_order_cycle, distributors: [distributor]) }
     let(:exchange) { order_cycle.exchanges.to_enterprises(distributor).outgoing.first }
-    let(:pr) { ProductsRenderer.new(distributor, order_cycle) }
+    let(:customer) { create(:customer) }
```
luisramos0: I wonder if this is enough testing for such a core component.

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,59 @@
+require 'open_food_network/products_renderer'
+
+module Api
+  class OrderCyclesController < BaseController
```
luisramos0: We need a new spec and some tests api/order_cycles_controller_spec

app/services/product_tag_rules_filterer.rb
```diff
@@ -0,0 +1,107 @@
+class ProductTagRulesFilterer
```
luisramos0: We need a new spec and some tests services/product_tag_rules_filterer_spec

app/assets/javascripts/darkswarm/controllers/order_cycle_controller.js.coffee
```diff
@@ -32,4 +33,4 @@ Darkswarm.controller "OrderCycleChangeCtrl", ($scope, $timeout, OrderCycle, Prod
     Products.update()
     Cart.reloadFinalisedLineItems()
     ChangeableOrdersAlert.reload()
-    # Reload Filters from new endpoint after changing OC here
+    $rootScope.$broadcast 'orderCycleSet'
```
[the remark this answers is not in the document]
Matt-Yorkley: Sometimes it's when it changes, or sometimes when it's set for the first time on page load. So it's "set or changed". Maybe "selected"?

lib/open_food_network/products_renderer.rb
Matt-Yorkley: I was just thinking that!

spec/lib/open_food_network/products_renderer_spec.rb
Matt-Yorkley: I'm currently writing test for the tag rule filtering part, which is scoped by customer. The ProductRenderer itself isn't really affected by `customer`.

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: I'm just finishing it off :smile:

app/services/product_tag_rules_filterer.rb
Matt-Yorkley: Already added :+1:

app/assets/javascripts/darkswarm/controllers/order_cycle_controller.js.coffee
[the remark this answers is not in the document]
luisramos0: up to you. I prefer changed, in my mind it includes changed from nothing to something: the first time.

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: Moved to separate endpoints :+1:

spec/lib/open_food_network/products_renderer_spec.rb
Matt-Yorkley: Okay, unit tests added in the service spec, and integration tests added in controller spec :+1:

app/serializers/api/product_serializer.rb
Matt-Yorkley: Ah yes. I just combined the old `ProductSerializer` and `CachedProductSerializer` here, but it looks like there were no specs for either of them.

luisramos0: this is being set in the ProductsRenderer.products_json
you extracted it to a method with def enterprise_fee_calculator

Matt-Yorkley: Serializer spec added :+1:

luisramos0: did you remove it?
in case there's an enterprise fee in the OC, the product prices must include that fee.

it looks like variant.price_with_fees used EnterpriseFeeCalculator#fees_for while this uses EnterpriseFeeCalculator#indexed_fees_for 

I am not sure what's the difference, could be an optimized way of calculating the fees?

this commit https://github.com/openfoodfoundation/openfoodnetwork/commit/cf0031d9a560775596950bb1d655cc17b7191ae9
maybe it just added an optimized way of doing it, on the same class...
and we probably want to keep using it.

app/assets/javascripts/darkswarm/services/products.js.coffee
```diff
@@ -1,26 +1,35 @@
...
+      params['distributor'] = currentHub.id
+
+      OrderCycleResource.products params, (data)=>
+        @products = [] unless load_more
+        @fetched_products = data
         @extend()
         @dereference()
         @registerVariants()
+        for product in @fetched_products
+          @products.push product
```
sauloperez: can't we do `@products = @products.concat(@fetched_products)`?

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+        params.slice(:q, :page, :per_page)
+      ).products_json
+
+      render json: products
+    rescue ProductsRenderer::NoProducts
+      render status: :not_found, json: ''
+    end
+
+    def taxons
+      taxons = Spree::Taxon.joins(:products).
```
sauloperez: I'd move `joins` to a separate line, as usual :trollface:

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+        where(spree_products: { id: distributed_products(distributor, order_cycle, customer) }).
+        select('DISTINCT spree_taxons.*')
+
+      render json: ActiveModel::ArraySerializer.new(taxons, each_serializer: Api::TaxonSerializer)
+    end
+
+    def properties
+      properties = Spree::Property.joins(:products).
```
sauloperez: here as well

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def customer
+      @current_api_user.andand.customer_of(distributor) || nil
+    end
+
+    def distributed_products(distributor, order_cycle, customer)
+      OrderCycleDistributedProducts.new(distributor, order_cycle, customer).products_relation
+    end
+  end
+end
```
sauloperez: It's pretty need that we start a new API controller but I'd really go the REST way following Rails conventions which will benefit us in many ways. Our codebase already suffers too much of fighting said approach and we pay the price IMO.

That means we should have 3 controllers under an `OrderCycles` routes namespace each implementing the `index` action. This would stick to the convention of `index`, `show`, `edit` and `delete` controller actions.

However, I'm happy to try this out and iterate but I'd ask you to please create an issue for it.

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    def products
+      products = ProductsRenderer.new(
+        distributor,
+        order_cycle,
+        customer,
+        params.slice(:q, :page, :per_page)
+      ).products_json
+
+      render json: products
+    rescue ProductsRenderer::NoProducts
```
sauloperez: This is something I never liked about the way we render the products. It feels to me that both the `ProductsRenderer` and the controller contain unnecessary code to handle the case of an empty list of products and even handled with exceptions. We should return whatever is there: either empty or not empty but always a JSON array.

Would you agree on creating an issue?

app/services/product_tag_rules_filterer.rb
```diff
@@ -0,0 +1,111 @@
...
+  def initialize(distributor, customer, variants_relation)
+    @distributor = distributor
+    @customer = customer
+    @variants_relation = variants_relation
+  end
+
+  def call
+    return variants_relation unless distributor_rules.any?
+
+    filter!(variants_relation)
```
sauloperez: I would remove the bang as it doesn't follow any of the meanings it could have in Ruby land: either a method that performs a mutation (Ruby) or a method that may raise an exception (Rails). IMO it can be very confusing for anyone used to Ruby.

app/services/products_renderer.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def initialize(distributor, order_cycle, customer, args = {})
+    @distributor = distributor
+    @order_cycle = order_cycle
+    @customer = customer
+    @args = args
+  end
+
+  def products_json
+    raise NoProducts unless order_cycle && distributor && products
```
sauloperez: and this what feels broken to me. We say `NoProducts` when what we really mean is some custom flavor of [ArgumentError[(https://apidock.com/ruby/v2_1_10/ArgumentError).

Besides, I would raise it from the `initialize` method right away.

app/services/products_renderer.rb
```diff
@@ -0,0 +1,92 @@
...
+  def product_scoper
+    OpenFoodNetwork::ScopeProductToHub.new(distributor)
+  end
+
+  def enterprise_fee_calculator
+    OpenFoodNetwork::EnterpriseFeeCalculator.new distributor, order_cycle
+  end
+
+  def filter_and_paginate(query)
+    query.ransack(args[:q]).result.page(args[:page] || 1).per(args[:per_page] || 10)
```
sauloperez: We should move these magic numbers to constants but I'd first check whether the pagination gem we use (I don't remember which one of the available) and if it allows defining a global default. Didn't we implement something like that not long ago?

app/services/products_renderer.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def distributed_products
+    OrderCycleDistributedProducts.new(distributor, order_cycle, customer)
+  end
+
+  def taxon_order
+    if distributor.preferred_shopfront_taxon_order.present?
+      distributor
+        .preferred_shopfront_taxon_order
+        .split(",").map { |id| "primary_taxon_id=#{id} DESC" }
```
sauloperez: Again, I'd be consistent and have all lines separate.

app/services/products_renderer.rb
```diff
@@ -0,0 +1,92 @@
...
+        .join(",") + ", name ASC"
```
sauloperez: I guess we only perform this `preferred_shopfront_taxon_order` entanglement in this class. If not, we should really encapsulate it. What a bad idea to store this preference in a string :man_facepalming:

spec/services/products_renderer_spec.rb
```diff
@@ -0,0 +1,117 @@
+require 'spec_helper'
+
+describe ProductsRenderer do
+  let(:distributor) { create(:distributor_enterprise) }
+  let(:order_cycle) { create(:simple_order_cycle, distributors: [distributor]) }
+  let(:exchange) { order_cycle.exchanges.to_enterprises(distributor).outgoing.first }
+  let(:customer) { create(:customer) }
+  let(:pr) { ProductsRenderer.new(distributor, order_cycle, customer) }
```
sauloperez: Can we rename it to `products_renderer`? IMO acronyms like this hurt readability

spec/services/products_renderer_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    before do
+      exchange.variants << p1.variants.first
+      exchange.variants << p2.variants.first
+      exchange.variants << p3.variants.first
+      exchange.variants << p4.variants.first
+    end
+
+    it "sorts products by the distributor's preferred taxon list" do
+      allow(distributor).to receive(:preferred_shopfront_taxon_order) { "#{t1.id},#{t2.id}" }
+      products = pr.send(:products)
```
sauloperez: I know this a lot to test in this class but unit tests only cover the public API of a class. If we are not confident about this method is because the class is too big and should be split into smaller ones instead. 

Would you agree on creating an issue for it?

spec/services/products_renderer_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    it "strips html from description" do
+      product.update_attribute(:description, "<a href='44'>turtles</a> frogs")
+      json = pr.products_json
+      expect(json).to include "frogs"
+      expect(json).not_to include "<a href"
+    end
+
+    it "returns price including fees" do
+      # Price is 19.99
+      allow_any_instance_of(OpenFoodNetwork::EnterpriseFeeCalculator).
```
sauloperez: :warning: the `any_instance` are not recommended as lead to brittle tests. It might make things easy in here but pretty please, let's not lower the guard :pray: 

Even RSpec docs do not recommend them https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance

> This feature is sometimes useful when working with legacy code, though in general we discourage its use for a number of reasons:

app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
```diff
@@ -1,41 +1,65 @@
...
+    return unless order_cycle_id
+
+    params = {
+      id: order_cycle_id,
+      distributor: currentHub.id
+    }
+    OrderCycleResource.taxons params, (data)=>
+      $scope.supplied_taxons = {}
+      data.map( (taxon) ->
+        $scope.supplied_taxons[taxon.id] = Taxons.taxons_by_id[taxon.id]
```
sauloperez: I don't fully understand what we're doing here. We already loaded taxons in `Taxons` but request them through `OrderCycleResource`. Also, is this assignment some sort of whitelisting? I'd appreciate if you explain it a bit more as method doc in the code itself.

app/serializers/api/product_serializer.rb
Matt-Yorkley: I'll put it back in and add some test coverage in the Serializer spec.

app/services/product_tag_rules_filterer.rb
Matt-Yorkley: Interestingly all the services that deal with tag rules seem have a `#filter!` method with a bang. I've removed it here. :+1:

app/services/products_renderer.rb
Matt-Yorkley: You should see the `tag_rules` data model... :scream:

spec/services/products_renderer_spec.rb
Matt-Yorkley: These tests were mostly moved across from an existing spec, so I didn't spend a long time rewriting every line to make it prettier. The tests here were pre-existing, and I didn't want to delete them.

Given that this service is incredibly important, I think there is some value in unit testing these private methods, and not just the public interface. What do you think?

app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
Matt-Yorkley: I think basically there's some dereferencing magic going on, where `Taxons` contains additional data for each taxon. I'm trying to work with the existing structures here without having to rewrite mountains of Angular. :face_with_head_bandage:

spec/services/products_renderer_spec.rb
Matt-Yorkley: This is also an old test that was just copied across when the `ProductsRenderer` namespace was changed. 

What would you like it to be?

app/assets/javascripts/darkswarm/services/products.js.coffee
Matt-Yorkley: Apparently everything breaks when I apply that suggestion, so I'm going to say "no we can't". :trollface:

app/services/products_renderer.rb
Matt-Yorkley: Constants added :+1:

I'm actually not sure about messing with global pagination defaults. I think we can start getting unexpected results, like limits being applied "automagically" on queries that will return incorrect results when limited.

Matt-Yorkley:
On making smaller PRs: I tried to do as much as possible in the previous ProductsRenderer PR to minimise the inevitable large size of this PR, but there's so much that needs to change at once here, it was never going to be small. I can't see how to split this PR into several chunks without them being broken and untestable.

On the cache: the previous setup was caching the entire query of all products and variants at once, and as we've seen it was over 1MB of json. With these new changes (moving the searching server-side), any caching done would have to be scoped with the query string and filters. So the number of cache entries would increase exponentially, as there would be a unique entry for every single combination of search term and selected filter. This would make the caching _much_ less effective, and I think there would be consequences to running the cache refresh callbacks all over the codebase. 

On performance: the previous PR removed most of the N+1 issues from the ProductRenderer and reduced the overall number of queries by 61%, making it much more efficient and lightweight. Combined with the new pagination regime of returning only 10 results at a time, we get an incredibly fast (and small) response from the endpoint.

spec/services/products_renderer_spec.rb
sauloperez: I'm fine if these were moved but tests that check private methods are not pretty or ugly they simply cost money as private methods are less stable than a public API but also are a symptom of underlying complexity. They are a code smell.

That's why I would ship this but create an issue to work on it right away.

spec/services/products_renderer_spec.rb
sauloperez: These can be solved by injection test doubles in the right place so that in this case it's not `any` `EnterpriseFeeCalculator` but the one you expect. I'm not asking to change it but just to be aware of the bad practice.

app/assets/javascripts/darkswarm/controllers/products_controller.js.coffee
sauloperez: then this should be reflected as code comments IMO so the next dev is aware of it.

sauloperez:
Yes, I later realized that the cache is totally tied to the current implementation. Still, [featture toggles](https://www.martinfowler.com/articles/feature-toggles.html) is something I'd like to investigate in the feature as a way to ship smaller PRs. I see it as a key ingredient towards having more frequent releases with lower risk.

In any case, you did an amazing job here @Matt-Yorkley congrats! :clap:

spec/controllers/api/order_cycles_controller_spec.rb
```diff
@@ -0,0 +1,158 @@
+require "spec_helper"
+
+module Api
+  describe OrderCyclesController, type: :controller do
```
mkllnk: Ideally, this would have been included in the commit adding the controller.

app/services/products_renderer.rb
mkllnk: I agree in general but Matt just moved this code and didn't write it. The pull request is big enough as is and we should separate concerns.

sauloperez:
I haven't mentioned but thanks for moving stuff out of `lib/open_food_network/` @Matt-Yorkley !! It's one of those things that made me think "Thank you!!".

app/services/order_cycle_distributed_products.rb
```diff
@@ -2,36 +2,108 @@
...
-            AND variant_overrides.hub_id = #{@distributor.id}").
+            AND variant_overrides.hub_id = #{distributor.id}").
       joins(:stock_items).
       where(query_stock_with_overrides)
+
+    if distributor_rules.any?
+      stocked_variants = apply_tag_rules(stocked_variants)
+    end
+
+    stocked_variants
```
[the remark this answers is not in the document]
Matt-Yorkley: This bit is refactored in a later commit :+1:

spec/controllers/api/order_cycles_controller_spec.rb
Matt-Yorkley: Ideally, yes. The controller changed several times through the PR, I added the spec once it was stable...

app/services/product_tag_rules_filterer.rb
```diff
@@ -0,0 +1,111 @@
...
+
+    filter(variants_relation)
+  end
+
+  private
+
+  attr_accessor :distributor, :customer, :variants_relation
+
+  def distributor_rules
+    @distributor_rules ||= TagRule::FilterProducts.for(distributor).all
```
kristinalim: This needs the `prioritised` scope, right?

app/controllers/api/order_cycles_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    respond_to :json
+
+    skip_authorization_check
+
+    def products
+      products = ProductsRenderer.new(
+        distributor,
+        order_cycle,
+        customer,
+        params.slice(:q, :page, :per_page)
```
kristinalim: Can we also sanitize `params[:q]` here? Not sanitizing might make this endpoint more powerful, but I think this search feature should only allow filters that we explicitly want to support.

sauloperez: This deserves then a separate method. I was thinking about following the [strong params](https://github.com/rails/strong_parameters) convention and so we are a little closer but feel free to dismiss it.

sauloperez:
> so much less code to maintain.

so much less pain and suffering on the ops side...

app/services/product_tag_rules_filterer.rb
Matt-Yorkley: Added :+1:

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: I've added the `strong_parameters` gem and used it here. :+1:

Matt-Yorkley: Hmm... nope, I'll have to remove it. The gem was discontinued in 2014 and seems to have some issues. I don't think we can use it. :disappointed:

Matt-Yorkley: @kristinalim I've added a new method for restricting the params here: https://github.com/openfoodfoundation/openfoodnetwork/pull/4345/commits/aa0217da08d65fbb4bfc7c9ab7909f749027dc1b

It's not very pretty. `#slice` doesn't take nested arguments. Let me know if there's a nicer way to do this!

kristinalim: Yes, not very pretty. :wink: But I think that's okay for now, it's in a separate method and we can easily update when we have upgraded Rails.

sauloperez: Yes, that was my point. Adding the gem it's too much. It was integrated into Rails 4 if I'm not mistaken. We can wait.

sauloperez:
I'm more than happy with this!! Can't wait to see it in production and see all the spikes disappear from Datadog's DJ jobs as well as shop endpoints :rocket: !!

lin-d-hop:
Starting testing here if anyone else comes to contribute :-)

https://docs.google.com/document/d/190zBttky5Yl1cJk1Pwkv3v6FTGYbcJzKAieYwa22lgc/edit#

lin-d-hop:
Found an issue with properties. Steps to recreate:
1) Add properties to the whole enterprise
2) Add properties to products
3) Filter by the product property
4) Filtering does not work.
5) I then removed the Enterprise level properties and filtering still did not work.

Matt-Yorkley:
Thanks for the in-depth testing!  I'll take a look. I think I added some basic specs for product properties but not these enterprises level properties. 

Do they apply to all the products in an enterprise? If that's the cade they will be useless as filters...?

lin-d-hop:
Yes they are useless as filters, but with those filters applied the product level filters don't seem to work.
So the problem is not to do with the enterprise level filters, however after applying (and removing) these filters the product level filters do not work.

You can test the 'Grass Fed' filter now with Dorv on AU staging. It does nothing. I only applied this filter to a single product.

Matt-Yorkley:
Ok, I think I've fixed it. I also expanded the test coverage for using both taxon and property filters.

mkllnk:
@Matt-Yorkley There is one tiny Code Climate issue.

Matt-Yorkley:
Note: the new build needs to be re-staged before testing again.

lin-d-hop:
restaging now

Matt-Yorkley:
Ah... product properties are definitely working, but I might need another look at producer properties.

RachL:
@Matt-Yorkley I see you've added some commits after your last comment. Is this ready for testing again?

Matt-Yorkley:
Yes, ready for testing :+1:

lin-d-hop:
I would assume with the change /admin/cache_settings/ can go from super admin...?
I guess a low priority and even if I open an issue it won't be done. In case you're feeling unusually OCD @matt_yorkley? ;-)

Matt-Yorkley:
Pretty sure I removed it already..?

lin-d-hop:
Not from the menu ;-)

Matt-Yorkley:
Updated :+1:

lin-d-hop:
Alright.... testing is complete. I feel like I gave it a good and proper workthrough.

I'm moving this to Ready to Go..... BOOM!

lin-d-hop:
I'm guessing this makes a major dot release. I'm going to write a blog post in it :-D

luisramos0:
ah. good point. yes. v2.6

On Tuesday, 22 October 2019, Lynne <notifications@github.com> wrote:
> I'm guessing this makes a major dot release. I'm going to write a blog
post in it :-D
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub, or unsubscribe.<
https://ci5.googleusercontent.com/proxy/qRIARGPMGSKC2Ks2zRn_Yl6NdtI48gU2R_-ppVEfZJQNiqKPiNRI9uNxhWhMibpB27T26SgmOwp_1undUGffg82INYKfEPglVX5nVckjsiYcZj5-IhlOIt-rWAulJY53Vdb3S1LxMmJEOb0_xaiTtiMr5uh7N7nvW4uRHG-0V4ciaMR3kuj3-Ir8jfT0_vpzZP4SvQ2thoZvhjkrJB9VbNm6W6QMSPF9i-ZqAvQFyQ=s0-d-e1-ft#https://github.com/notifications/beacon/AAMQPOVIPQ3UBN3YCX5NRZ3QP3X65A5CNFSM4I4V6G6KYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEB5SQBA.gif>

PR #4351, opened 2019-10-03
Add constraint to validate unit_value and unit_description are not null at the same time for a given variant

### LINKED ISSUE #4216: Cannot clone order cycle - grumpy cat
## Description

A user in Australia is trying to clone an order cycle. They are getting a grumpy cat (see below). This is the 'old' error message screen that should have been replaced with the snail. So it suggests that we are touching some weird part of the code?

They have now sent a follow up email saying that "After the message with the cat picture. It does show a copy of the order cycle but it only has 73 variants and not the 521 variants  that the Master order cycle contains."

I have asked whether they can think of any other clues, and also for the name of the order cycle.

I have just tried cloning 3719 Master, and replicated. This is what dev tools look like - fails really fast!

## Expected Behavior
OC should be cloned

## Actual Behaviour
Grumpy cat and incomplete OC created

## Steps to Reproduce

1. Go to Aus prod
2. Clone 3719 Master OC, or it seems any OC as Food Connect manager

I have tried in another hub and it seems ok 

## Context
User clones their order cycles to get shops open for the week. This needs to be done in next hour. Working on a workaround but pretty stressful 

## Severity
S1 - cloning order cycles is a critical feature for many users, especially with this many products

luisramos0:
This is already done at AR Variant model level [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/e6408161dbefc7da320b247ed5ac97472602c325/app/models/spree/variant_decorator.rb#L23)

#### What? Why?

Closes #4216

This will prevent 4216 from happening again. If the situation happens again, it will not break the OC page and we will see a bugsnag message with the SQL error.

#### What should we test?
There were no examples in live DBs with the problem so we can just verify the deployment works in staging.

#### Release notes
Changelog Category: Fixed
Added verification to avoid issue #4216 from happening again.

### DIFF
--- a/db/migrate/20191003091125_add_variant_unit_value_unit_description_not_null_constraint.rb
+++ b/db/migrate/20191003091125_add_variant_unit_value_unit_description_not_null_constraint.rb
@@ -1,0 +1,10 @@
+# Ensures variant.unit_value and variant.unit_description are not null at the same time for a given variant
+class AddVariantUnitValueUnitDescriptionNotNullConstraint < ActiveRecord::Migration
+  def up
+    execute "ALTER TABLE spree_variants ADD CONSTRAINT unit_value_or_unit_desc CHECK (unit_value is not null or unit_description is not null);"
+  end
+
+  def down
+    execute "ALTER TABLE spree_variants DROP CONSTRAINT unit_value_or_unit_desc;"
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20190916110029) do
+ActiveRecord::Schema.define(:version => 20191003091125) do
 
   create_table "adjustment_metadata", :force => true do |t|
     t.integer "adjustment_id"


Matt-Yorkley:
I think this might need some rigorous testing. Apparently Rails has some [issues with constraints](https://edgeguides.rubyonrails.org/active_record_migrations.html#types-of-schema-dumps):

> `db/schema.rb` cannot express everything your database may support such as triggers, sequences, stored procedures, check constraints, etc. While migrations may use `execute` to create database constructs that are not supported by the Ruby migration DSL, these constructs may not be able to be reconstituted by the schema dumper. If you are using features like these, you should set the schema format to `:sql` in order to get an accurate schema file that is useful to create new database instances.

Additional discussion here: https://stackoverflow.com/questions/4664520/how-do-i-add-a-check-constraint-in-a-rails-migration

Matt-Yorkley:
Also, will this statement definitely achieve the desired effect? 

`unit_value is not null or unit_description is not null`

I read that as "one of the fields must be null", not "both fields can't be null".

luisramos0:
yeah, I see.
I tried doing that now but when I run migrate I get into this problem:
https://stackoverflow.com/questions/35999906/pg-dump-invalid-option-i-when-migrating

having the schema as is is better than having it in sql, right?

what's the right approach here? Should I try to fix the conversion of schema to sql or should we find another way around the constraint?

note, this check is already in place in the code:
https://github.com/openfoodfoundation/openfoodnetwork/blob/e6408161dbefc7da320b247ed5ac97472602c325/app/models/spree/variant_decorator.rb#L23

sauloperez:
> having the schema as is is better than having it in sql, right?

not necessarily, the only reason why it's an rb file is to be agnostic about the underlying DB adapter so that you can change DBMS at will. i.e. from to Postgres to Mysql without touching the app code. In my experience this doesn't happen often and it won't any time soon for us.

So, if we want to leverage PostgreSQL CHECK constraints there's no way around having a SQL schema. Example: https://github.com/coopdevs/timeoverflow/pull/465/files#diff-f9ae8314082d8229b38da27dd6c2642d.

we can try splitting this into two PRs 1st structure.sql and 2nd adding the constraint so we lower risks, perhaps.

Otherwise, we'll need to forget about CHECK constraints but there's a lot we can benefit from using PostgreSQL and that's only through structure.sql, I'm afraid. There are sometimes gems to provide this missing capabilities like https://github.com/on-site/active_record-postgres-constraints but these feel like patches and add yet more dependencies to our stack.

luisramos0:
yeah, thanks everyone for their feedback. I think we agree that we will close this and leave things as they are.

sauloperez:
> I think that SQL is overly specific and can cause other issues later on. Sometimes it's good to just follow the Rails convention even if it's not ideal.

It's a Rails convention as well and properly documented. I see long-lasting Rails apps with a structure.sql and no complaints no reason not to move to it in the future. I see we'll be in this situation more and more as we want to leverage PostgreSQL capabilities but agree. Not now.

Issue #4100, opened 2019-07-30
Missing translation in non English enterprise welcome email


myriamboure:
## Description

Two strings are not translated in the welcome email an enterprise receive when a new enterprise is created on the OFN
The strings are not in Transifex.

![Capture du 2019-07-30 22-10-14](https://user-images.githubusercontent.com/8135376/62161580-d70b3880-b316-11e9-8489-7b9dea8f79d5.png)

## Context

Myriam retested stuff while updating French user guide.

## Severity
s3, a bit annoying especially as one of the first emails non English speaker users receive

## Your Environment

* OFN Platform instance where you discovered the bug, and which version of the software they are using : French 2.2.1

PR #4352, opened 2019-10-05
Add missing translation in non English enterprise welcome email

tkusuki:
#### What? Why?

Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/4100

Two strings are not translated in the welcome email an enterprise receive when a new enterprise is created on the OFN

#### What should we test?

In Rails console, call `EnterpriseMailer.welcome` to any enterprise account.
```ruby
enterprise = Enterprise.first
EnterpriseMailer.welcome(enterprise).deliver
```
Check the result in `tmp/letter_opener/.../rich.html`

Captura de tela de 2019-10-05 17-04-48

Added by @kristinalim:

Limited testing can be done for this in staging, because translations for other locales will not be available yet until this PR is merged into master and translations from Transifex are pulled in.

For now, we can only test the `en` locale, which staging Australia has as "English". You can deploy this in Australia staging, then:

1. Switch to English there (`en`).
2. Create a new enterprise.
3. Check the welcome email. There should be no obvious issues.

#### Release notes

Changelog Category: Added
Added translation to _"Open Food Network User Guidefred"_ and _"Admin Panel"_ links

### DIFF
--- a/app/views/enterprise_mailer/welcome.html.haml
+++ b/app/views/enterprise_mailer/welcome.html.haml
@@ -1,18 +1,17 @@
 %h3
-  = "#{t(:email_welcome)}!"
+  = "#{t(".email_welcome")}!"
 %p.lead
   %strong
     = @enterprise.name
-  = "#{t(:email_registered)} #{ Spree::Config.site_name }!"
+  = "#{t(".email_registered")} #{ Spree::Config.site_name }!"
 
 %p
-  = t :email_userguide_html, link: link_to(t(:userguide), ContentConfig.user_guide_link)
+  = t(".email_userguide_html", link: link_to(t(".userguide"), ContentConfig.user_guide_link))
+%p
+  = t(".email_admin_html", link: link_to(t(".admin_panel"), spree.admin_url))
 
 %p
-  = t :email_admin_html, link: link_to(t(:admin_panel), spree.admin_url)
-
-%p
-  = t :email_community_html, link: link_to(t(:join_community), ContentConfig.community_forum_url)
+  = t(".email_community_html", link: link_to(t(".join_community"), ContentConfig.community_forum_url))
 
 = render 'shared/mailers/signoff'
 
--- a/app/views/enterprise_mailer/welcome.html.haml
+++ b/app/views/enterprise_mailer/welcome.html.haml
@@ -6,10 +6,10 @@
   = "#{t(:email_registered)} #{ Spree::Config.site_name }!"
 
 %p
-  = t :email_userguide_html, link: link_to('Open Food Network User Guide', ContentConfig.user_guide_link)
+  = t :email_userguide_html, link: link_to(t(:userguide), ContentConfig.user_guide_link)
 
 %p
-  = t :email_admin_html, link: link_to('Admin Panel', spree.admin_url)
+  = t :email_admin_html, link: link_to(t(:admin_panel), spree.admin_url)
 
 %p
   = t :email_community_html, link: link_to(t(:join_community), ContentConfig.community_forum_url)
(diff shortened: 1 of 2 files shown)


app/views/enterprise_mailer/welcome.html.haml
```diff
@@ -6,10 +6,10 @@
   = "#{t(:email_registered)} #{ Spree::Config.site_name }!"
 
 %p
-  = t :email_userguide_html, link: link_to('Open Food Network User Guide', ContentConfig.user_guide_link)
+  = t :email_userguide_html, link: link_to(t(:userguide), ContentConfig.user_guide_link)
 
 %p
-  = t :email_admin_html, link: link_to('Admin Panel', spree.admin_url)
+  = t :email_admin_html, link: link_to(t(:admin_panel), spree.admin_url)
```
luisramos0: we are migrating al translations to lazy lookup as explained here:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29#development

so, in this case, instead of a root level key, we need a relative key
here it needs to be
`= t :email_admin_html, link: link_to(t('.admin_panel'), spree.admin_url)`

and in the en.yml file it needs to be in the specific place of this file
in this case
app/views/enterprise_mailer/welcome.html.haml 
makes a i18n relative  path of: enterprise_mailer.welcome
so this translation needs to be enterprise_mailer.welcome.admin_panel
[here](https://github.com/openfoodfoundation/openfoodnetwork/blob/8a732568e7459bdee8ef07e545c243c7d7689e82/config/locales/en.yml#L138)

luisramos0: you could change all the entries in this file (email_welcome, email_registered and join_community) to lazy lookup by replacing :key with '.key' and adding entries in en.yml under enterprise_mailer.welcome

tkusuki: > we are migrating al translations to lazy lookup as explained here:
> https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29#development
> 
> so, in this case, instead of a root level key, we need a relative key
> here it needs to be
> `= t :email_admin_html, link: link_to(t('.admin_panel'), spree.admin_url)`
> 
> and in the en.yml file it needs to be in the specific place of this file
> in this case
> app/views/enterprise_mailer/welcome.html.haml
> makes a i18n relative path of: enterprise_mailer.welcome
> so this translation needs to be enterprise_mailer.welcome.admin_panel
> [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/8a732568e7459bdee8ef07e545c243c7d7689e82/config/locales/en.yml#L138)

Thanks @luisramos0  for posting the i18n link. :smile: 
I've changed as you suggest, can you take a look?

luisramos0: awesome. I just double checked that all these keys are only used in this page, so it's fine to delete them from where they were in the en.yml file :+1:

tkusuki: Yeah.. I checked here, especially if `t: email_welcome` could be used for other e-mails. 😊

filipefurtad0:
Hey @kristinalim!

Would it be ok to test this in UK? Since it also has en as local language. Or must it be AU only:

"For now, we can only test the en locale, which staging Australia has as "English". You can deploy this in Australia staging"

I am asking because AU is busy at the moment :-)
Thank you!

luisramos0:
I think uk staging test is enough, if the keys were not correct, you would see automated translations of the keys and not the content in en.yml :+1:

Issue #4256, opened 2019-09-12
When a shopfront's products don't fit into cache, the shop page takes ages to load


lin-d-hop:
## Description

When a shopfront's products don't fit into cache, the shop page takes ages to load. This, at times, can lead to timeouts in which the shopfront just fails to load altogether.

When the list of products for a distributor and OC is not cached the DB is consulted to do that same lookup. This is incredibly slow.

This is done through the `/shop/products` endpoint.

In particular this is effecting Mercia Food Hub in UK. The shopfront is losing sales. Customers are emailing to say they aren't ordering because the site is failing to load.

## Expected Behavior

Shopfronts load successfully without timing out.

## Actual Behaviour

Sometimes shopfronts are failing to load.

## Steps to Reproduce

1. Identify a shopfront with products missing from the product cache
2. Try to load the shopfront
3. It will take an unacceptable length of time and possibly timeout.

## Animated Gif/Screenshot

Insert GIF of an endless OFN spinner here

## Context

Shopfronts have reported not being able to load and losing customers as a result.

## Severity

Technically this is an S1. However I'm not going to label it as such just yet as we have a plan and it would be nicer to just do the issues than treat it as an S1.

## Your Environment

* Version used:
* Browser name and version:
* Operating System and version (desktop or mobile):
* OFN Platform instance where you discovered the bug, and which version of the software they are using. 

## Possible Fix

To explore this we first need:
- Memcache logging https://github.com/openfoodfoundation/ofn-install/issues/515

lin-d-hop:
@sauloperez I have no idea what you mean in your last comment

lin-d-hop:
Hi all,
I've upgraded this to an S2. 
The user reported 3 people got in touch to say they didn't order from the shop because the shop didn't load in the last order cycle. The actual number effected is unknown.

Although we need a real solution to this, the main priority right now is to have this shops products load into the cache. Any workaround that enables that is fine to remove the S2 label.

Thanks all.

lin-d-hop:
This is the shop:
https://openfoodnetwork.org.uk/mercia-food-hub/shop

I've tried warming the cache and that hasn't helped.
The user is also willing to remove products from the OC that are offending just to get the shop to load.....

lin-d-hop:
This is the console during the RQ to this shop. 1.3mins waiting for products.

[...]

PR #4356, opened 2019-10-08
4256 Configure dalli store value max bytes from env variable

kristinalim:
#### What? Why?

- References #4256

We need larger data to be stored in the Rails cache. PR https://github.com/openfoodfoundation/ofn-install/pull/531 takes care of configuring the memcached server to use 4 MB. But additionally, the memcached client which in this case is `dalli` also needs to be configured to allow storing values up to this size.

The `ofn-install` PR is not a requirement for this PR - we use the default value of 1 MB when no `MEMCACHED_VALUE_MAX_MEGABYTES` environment variable is specified, so we should be okay.

In the `dalli` gem, the default value of `value_max_bytes` is 1 MB (`1 * 1024 * 1024`). See [here](https://github.com/petergoldstein/dalli/blob/v2.7.10/lib/dalli/server.rb#L25-L26).

#### What should we test?

`dev-test`: Test that storing objects up to 4 MB works. The following code for 1 MB will be helpful:

```
cache_key = "sample-test"
megabytes = 1

Rails.cache.fetch(cache_key)
# => nil
Rails.cache.write(cache_key, "." * (megabytes * 1024 * 1024 - 100))
# => 6064378373231083520
Rails.cache.fetch(cache_key).length
# => 1048476
Rails.cache.write(cache_key, "." * (megabytes * 1024 * 1024 + 100))
# => false
Rails.cache.fetch(cache_key).length
# => 1048476
Rails.cache.delete(cache_key)
Rails.cache.fetch(cache_key)
# => nil
```

A quick manual test also that shopfront loads fine should be done as well.

#### Release notes

- Allow maximum `memcached` value size to be specified as an environment variable, to allow data of larger shops to be cached in order to improve performance.

Changelog Category: Changed

#### Dependencies

Ideal if https://github.com/openfoodfoundation/ofn-install/pull/531 could be merged and provisioned before this PR is deployed, but not required.

### DIFF
--- a/config/application.yml.example
+++ b/config/application.yml.example
@@ -56,3 +56,5 @@ SMTP_PASSWORD: 'f00d'
 # STRIPE_INSTANCE_PUBLISHABLE_KEY: "pk_test_xxxx" # This can be a test key or a live key
 # STRIPE_CLIENT_ID: "ca_xxxx" # This can be a development ID or a production ID
 # STRIPE_ENDPOINT_SECRET: "whsec_xxxx"
+
+MEMCACHED_VALUE_MAX_MEGABYTES: 4
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -40,7 +40,9 @@ Openfoodnetwork::Application.configure do
   # config.logger = SyslogLogger.new
 
   # Use a different cache store in production
-  config.cache_store = :dalli_store
+  memcached_value_max_megabytes = ENV.fetch("MEMCACHED_VALUE_MAX_MEGABYTES", 1).to_i
+  memcached_value_max_bytes = memcached_value_max_megabytes * 1024 * 1024
+  config.cache_store = :dalli_store, { value_max_bytes: memcached_value_max_bytes }
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server
   # config.action_controller.asset_host = "http://assets.example.com"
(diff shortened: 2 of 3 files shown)


kristinalim:
I tested this PR (deploy) with https://github.com/openfoodfoundation/ofn-install/pull/531 (provision), and it works until approximately `4 * 1024 * 1024` characters:

```ruby
cache_key = "sample-test"
megabytes = 4

Rails.cache.fetch(cache_key)
# => nil
Rails.cache.write(cache_key, "." * (megabytes * 1024 * 1024 - 100))
# => 12970366926827028480
Rails.cache.fetch(cache_key).length
# => 4194204
Rails.cache.write(cache_key, "." * (megabytes * 1024 * 1024 + 100))
# => false
Rails.cache.fetch(cache_key).length
# => 4194204

Rails.cache.delete(cache_key)
```

kristinalim:
Moving this to Test Ready, for testing by a dev aside from myself. :slightly_smiling_face:

luisramos0:
--- Ideal if openfoodfoundation/ofn-install#531 could be merged and provisioned before this PR is deployed, but not required.

I have deployed this to staging FR from local with of-install 531 and the test fails.
Maybe we do need to provision before.

kristinalim:
With "the test fails", did you mean the test here @luisramos0? https://github.com/openfoodfoundation/openfoodnetwork/pull/4356#issuecomment-539461736 If yes, it would indeed fail without the two PRs merged first.

But this PR on its own should be a non-breaking change - it merely allows the max size to be updated through an env variable.

kristinalim:
*the two PRs provisioned + deployed

luisramos0:
ok, tested after provisioning, it works :tada:

PR #4359, opened 2019-10-09
Add asterisks to required fields in New Order Cycle form

### LINKED ISSUE #3910: Add red * on all the mandatory fields when creating a new OC
## Description
**- As a:** enterprise user or super admin
**- On page:** /admin/order_cycles/new

**- I want to be able to do:** Be able to see more clearly which fields are mandatory to fill in.

The current mandatory fields are:
- Name
- Ready for

The proposal is to add a red * besides each field name. This is what is being done on product creation:

image

This feedback came out of https://community.openfoodnetwork.org/t/some-potential-ofn-improvements-from-a-small-normandy-producer/1663/2

It will help us to show more easily that the "ready for" field is mandatory.

## Acceptance Criteria
This issue can be closed if the following tests can be done:

*Scenario 1*

1. Log in as an enterprise user
2. Go to /admin/order_cycles/new
3. You can see red * on all the mandatory fields. When they are all filled up, the update button is enhanced (current behaviour)

*Scenario 2*

1. Log in as a super admin
2. Go to /admin/order_cycles/new
3. You can see red * on all the mandatory fields. When they are all filled up, the update button is enhanced (current behaviour)

daningenthron:
#### Closes #3910

Although the "Create" button is not enabled unless these required fields are complete, this clarifies workflow for users/admins while creating a new order cycle.

#### What should we test?

One item to look out for is an adjustment to an `en.yml` value - the English translation of "Orders Open At" was trimmed to "Orders Open" to match "Orders Close" and to keep this on one line with the asterisk.

#### Release notes

I believe this covers both user and admin use cases - let me know if any changes are needed!

Changelog Category: Added

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -928,7 +928,7 @@ en:
         new_schedule: New Schedule
       name_and_timing_form:
         name: Name
-        orders_open: Orders open at
+        orders_open: Orders open
         coordinator: Coordinator
         orders_close: Orders close
       row:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -928,7 +928,7 @@ en:
         new_schedule: New Schedule
       name_and_timing_form:
         name: Name
-        orders_open: Orders open
+        orders_open: Orders open at
         coordinator: Coordinator
         orders_close: Orders close
       row:
--- a/spec/features/admin/order_cycles_spec.rb
+++ b/spec/features/admin/order_cycles_spec.rb
@@ -181,10 +181,16 @@ feature '
       select2_select 'My coordinator', from: 'coordinator_id'
       click_button "Continue >"
 
-      # And I fill in the basic fields
+      # I cannot save before filling in the required fields
+      expect(page).to have_button("Create", disabled: true)
+
+      # After I fill in the basic fields
       fill_in 'order_cycle_name', with: 'Plums & Avos'
       fill_in 'order_cycle_orders_open_at', with: order_cycle_opening_time
       fill_in 'order_cycle_orders_close_at', with: order_cycle_closing_time
+
+      # I can save the form
+      expect(page).to have_button("Create", disabled: false)
 
       # And I add a coordinator fee
       click_button 'Add coordinator fee'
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -930,7 +930,7 @@ en:
         new_schedule: New Schedule
       name_and_timing_form:
         name: Name
-        orders_open: Orders open at
+        orders_open: Orders open
```
luisramos0: "One item to look out for is an adjustment to an en.yml value - the English translation of "Orders Open At" was trimmed to "Orders Open" to match "Orders Close" and to keep this on one line with the asterisk."

This will be different in other languages. Translators will break this without knowing. But I assume having it broken in two lines is not too bad, or is it? Maybe a screenshot will help us decide.

spec/features/admin/order_cycles_spec.rb
```diff
@@ -181,11 +181,17 @@
...
+      # I cannot save before filling in the required fields
+      expect(page).to have_button("Create", disabled: true)
+
+      # After I fill in the basic fields
       fill_in 'order_cycle_name', with: 'Plums & Avos'
       fill_in 'order_cycle_orders_open_at', with: order_cycle_opening_time
       fill_in 'order_cycle_orders_close_at', with: order_cycle_closing_time
 
+      # I can save the form
+      expect(page).to have_button("Create", disabled: false)
```
luisramos0: maybe we can add a comment here about the existing incoherence between the button showing and the actual mandatory fields? as far as I understand the button is showing even if the mandatory fields pickup_time and pickup_instructions are not filled in.

luisramos0:
hey @daningenthron thanks for one more PR!!!

config/locales/en.yml
daningenthron: I agree - the English is actually the shortest translation; any other language will definitely break across the line, so it's not worth making the adjustment just for the English. I'll revert to the original value.

daningenthron:
Confirmed with Rachel that `orders_open_at` and `orders_close` are not required fields. That's been updated here, and `en.yml` reverted. Since the spec covers two create interfaces, one that requires `pickup_time` and one that does not, I tried to be clearer with the scenario/describe statements.

RachL:
The red stars are in place :)

image


Thanks @daningenthron !

PR #4371, opened 2019-10-14
Move Payment methods from spree_backend to OFN

luisramos0:
#### What? Why?

Part of #4050

This moves payment methods from spree_backend to OFN. No change in functionality.

#### What should we test?
Make sure payment methods pages work correctly: list payment methods, create payment methods, edit payment methods and delete payment methods.

#### Release notes
Changelog Category: Changed


luisramos0:
The code climate issues are existing spree code issues. Complexity, etc...

sauloperez:
We might need to run the rubocop_todo again @luisramos0 ? We could approve the PR in Code Climate and do that in a separate PR. Otherwise, we should disable the particular cops with inline comments.

luisramos0:
ok, I'll create a PR with a new rubocop todo maybe with a single dev review so it goes quick into master

filipefurtad0:
Hi @luisramos0,

Created, Read, Edited and Deleted payment methods. Tested with cash and paypal in AU-staging. All smooth, moving to Ready to Go :-)

PR #4372, opened 2019-10-14
Move Images from spree_backend to OFN

luisramos0:
#### What? Why?

Part of #4050

This moves images methods from spree_backend to OFN. No change in functionality.

#### What should we test?
Make sure images pages work correctly: list images, add image, edit image and delete image.

We dont really support images per variant, only images on variant "All" will work. This was not changed as part of this PR. I have detected that if you upload an image to a specific variant, afterwards you cant edit or delete that particular image. This is broken before and after this PR...
Users should only use variant: "All" in this page. An issue can be created to remove the variant select box from this page.

#### Release notes
Changelog Category: Changed
Imported views related to product image management from spree_backend to make OFN more independent of Spree.


filipefurtad0:
Hi again @luisramos0,

Tested according to notes. Listing, adding, editing and delete images:
-> all good

Reproduced the issues you mentioned:
- only images on variant "All" will work.
- after uploading animage to a specific variant, you cant edit or delete that particular image afterwards

Moving to Ready to Go.
Thank you! :-)

filipefurtad0:
Opss

luisramos0:
great, thank you Filipe!

PR #4376, opened 2019-10-16
Remove product from Order Cycles if product supplier changes

### LINKED ISSUE #4366: Reports show ghost variants
## Description

Several reports, filtered by order cycle, show variants that are not in the order cycle. This happens after a product has been assigned to a different supplier. Changing the supplier has the affect that all variants of that product disappear from all order cycles. They are not shown on the order cycle edit page and are not shown in the shop front. But they are still connected to the order cycle via the old supplier and show up in reports.

The affected reports include:

- inventory reports including Lettuce Share
- all products report

## Expected Behavior

1. When filtering a report by order cycle, only products actually visible in the order cycle should show up.
2. When the supplier of a product changes, it should not affect order cycles. This is undefined though and needs some specification.
  - Option 1: We remove the product from all order cycles. This is like the current behaviour but without ghost entries.
  - Option 2: We update all order cycles to change the supplier of the product in the order cycles as well. This would leave the product in the order cycle. This is not possible though if the coordinator of the order cycle doesn't have permission to add products form the new supplier.

## Actual Behaviour

The changed product disappears from all order cycles but remains on reports. These ghost products are carried forward when cloning order cycles.

## Steps to Reproduce

1. You need two producers and one order cycle.
2. Add a product from one producer to the order cycle and observe that it's showing in the LettuceShare report.
3. Go to the products page and change the supplier of that product.
4. Observe that the product is not visible in the order cycle any more but still shows up in the LettuceShare report.

## Workaround

If the coordinator of the order cycle has full permissions over the supplier, they can delete the product and create it again.

## Severity

The mentioned workaround is not possible for all users. But if you think it's enough of a workaround, we can demote this bug to s3.

## Your Environment

* Version used: v2.5.1

## Possible Fix

Option 1 mentioned above would be my favourite solution. Remove products from all order cycles when the owner changes. I would usually suggest to do this in the controller to avoid complex callbacks but a callback may be right to ensure data consistency:

After saving a product, if the owner changed, delete all associated exchange variants. Since we don't have the products cache any more, we can delete all of them in a single SQL query to avoid a performance bottleneck. We don't have a model for ExchangeVariant and Exchange doesn't have any other callbacks to consider.

luisramos0:
#### What? Why?
Closes #4366 
With a new supplier the rules/permissions to add a product to an Order Cycle may be different and also the reports may break because of inconsistent data, SO we need to remove all the product variants from all Order Cycles if the supplier of a product is changed.

#### What should we test?
We only need to test editing a product in two places: the edit individual product page and the bulk product edit page.
- Add a product to an Order Cycle
- Change supplier of product
- Verify all the variants of the product are removed from the Order Cycle

Also make sure changing other product attributes still works correctly.

#### Release notes
Changelog Category: Fixed
When the supplier of a product is changed the product is automatically removed from all Order Cycles so that data remains consistent. The product can be manually re-added to the Order Cycle.

### DIFF
--- a/app/controllers/spree/admin/products_controller_decorator.rb
+++ b/app/controllers/spree/admin/products_controller_decorator.rb
@@ -50,7 +50,7 @@ Spree::Admin::ProductsController.class_eval do
 
     delete_stock_params_and_set_after do
       super
-      remove_product_from_order_cycles if original_supplier != @product.supplier_id
+      ExchangeVariantDeleter.new.delete(@product) if original_supplier != @product.supplier_id
     end
   end
 
@@ -191,13 +191,6 @@ Spree::Admin::ProductsController.class_eval do
     else
       product.master
     end
-  end
-
-  def remove_product_from_order_cycles
-    variant_ids = @product.variants.map(&:id)
-    ExchangeVariant.
-      where(variant_id: variant_ids).
-      delete_all
   end
 
   def set_product_master_variant_price_to_zero
--- a/app/controllers/spree/admin/products_controller_decorator.rb
+++ b/app/controllers/spree/admin/products_controller_decorator.rb
@@ -46,11 +46,11 @@ Spree::Admin::ProductsController.class_eval do
   end
 
   def update
-    original_supplier = @product.supplier_id
+    original_supplier_id = @product.supplier_id
 
     delete_stock_params_and_set_after do
       super
-      ExchangeVariantDeleter.new.delete(@product) if original_supplier != @product.supplier_id
+      ExchangeVariantDeleter.new.delete(@product) if original_supplier_id != @product.supplier_id
     end
   end
 
--- a/app/controllers/spree/admin/products_controller_decorator.rb
+++ b/app/controllers/spree/admin/products_controller_decorator.rb
@@ -46,8 +46,11 @@ Spree::Admin::ProductsController.class_eval do
   end
 
   def update
+    original_supplier = @product.supplier_id
+
     delete_stock_params_and_set_after do
       super
+      remove_product_from_order_cycles if original_supplier != @product.supplier_id
     end
   end
 
@@ -188,6 +191,13 @@ Spree::Admin::ProductsController.class_eval do
     else
       product.master
     end
+  end
+
+  def remove_product_from_order_cycles
+    variant_ids = @product.variants.map(&:id)
+    ExchangeVariant.
+      where(variant_id: variant_ids).
+      delete_all
   end
 
   def set_product_master_variant_price_to_zero
--- a/app/models/spree/product_set.rb
+++ b/app/models/spree/product_set.rb
@@ -57,7 +57,6 @@ class Spree::ProductSet < ModelSet
   def update_product_only_attributes(product, attributes)
     variant_related_attrs = [:id, :variants_attributes, :master_attributes]
     product_related_attrs = attributes.except(*variant_related_attrs)
-
     return true if product_related_attrs.blank?
 
     original_supplier = product.supplier_id
@@ -66,9 +65,10 @@ class Spree::ProductSet < ModelSet
     product.variants.each do |variant|
       validate_presence_of_unit_value(product, variant)
     end
-    product.save if errors.empty?
+    return false unless product.errors.empty? && product.save
 
     ExchangeVariantDeleter.new.delete(product) if original_supplier != product.supplier_id
+    true
   end
 
   def validate_presence_of_unit_value(product, variant)
--- a/app/models/spree/product_set.rb
+++ b/app/models/spree/product_set.rb
@@ -46,13 +46,15 @@ class Spree::ProductSet < ModelSet
 
     return true if product_related_attrs.blank?
 
+    original_supplier = product.supplier_id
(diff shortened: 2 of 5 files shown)


app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -190,6 +193,11 @@ def product_variant(product)
     end
   end
 
+  def remove_product_from_order_cycles
+    variant_ids_csv = @product.variants.map(&:id).join(', ')
+    ActiveRecord::Base.connection.execute("DELETE FROM exchange_variants WHERE variant_id IN (#{variant_ids_csv})")
+  end
```
luisramos0: if we agree this is the way to go I can move this method a a service ExchangeVariantDeleter

app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -190,6 +193,11 @@ def product_variant(product)
     end
   end
 
+  def remove_product_from_order_cycles
+    variant_ids_csv = @product.variants.map(&:id).join(', ')
+    ActiveRecord::Base.connection.execute("DELETE FROM exchange_variants WHERE variant_id IN (#{variant_ids_csv})")
```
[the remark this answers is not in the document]
luisramos0: no, there are no master variants in OCs as per previous/recent PR.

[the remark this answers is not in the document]
luisramos0: ah! I completely missed the ExchangeVariant model!
Yes, I can use it 👍

app/controllers/spree/admin/products_controller_decorator.rb
```diff
@@ -46,8 +46,11 @@ def create
   end
 
   def update
+    original_supplier = @product.supplier_id
```
mkllnk: I would call it `original_supplier_id`.

app/services/exchange_variant_deleter.rb
```diff
@@ -0,0 +1,8 @@
+class ExchangeVariantDeleter
+  def delete(product)
+    variant_ids = product.variants.map(&:id)
+    ExchangeVariant.
+      where(variant_id: variant_ids).
```
mkllnk: You can avoid a query here:
```suggestion
      where(variant_id: product.variants.select(:id)).
```

app/models/spree/product_set.rb
```diff
@@ -43,16 +57,18 @@ def update_product(found_model, attributes)
...
     product.assign_attributes(product_related_attrs)
 
     product.variants.each do |variant|
       validate_presence_of_unit_value(product, variant)
     end
+    return false unless product.errors.empty? && product.save
 
-    product.save if errors.empty?
+    ExchangeVariantDeleter.new.delete(product) if original_supplier != product.supplier_id
+    true
```
luisramos0: this is rubocop (use guard clause) forcing me to write confusing code.. I think this would be easier to understand:
```ruby
    if product.errors.empty? && product.save
      ExchangeVariantDeleter.new.delete(product) if original_supplier != product.supplier_id
      return true
    else
      return false
    end
```

luisramos0:
this is ready for review now.

luisramos0:
I think it's much better to have this in a separate class because it's not straight forward stuff, deleting a variant from a exchange is a very specific thing and making it stand alone in a class makes it very obvious. It's a pattern I recommend, when you have to write non-straight forward code, you should isolate it as much as possible. Mixing non-straight forward code with straight forward code will create a mess.

app/models/spree/product_set.rb
mkllnk: Maybe the deletion is in the wrong method? What about his one?

```diff
  def update_product(found_model, attributes)
+   oc_updater = ExchangeVariantDeleter.new(found_model) # caching the current supplier id
    update_product_only_attributes(found_model, attributes) &&
+     oc_updater.delete_if_supplier_changed &&
      update_product_variants(found_model, attributes) &&
      update_product_master(found_model, attributes)
  end
```

There are actually two ideas in here. One is separating updating-the-product and deleting-the-variants to simplify boolean logic and avoid branches with returns. This can also happen with a new method.

The other idea was to move the supplier storage into the service but it doesn't actually save any lines of code. So I don't think you should do that one but I wanted to share the idea.

luisramos0: ah, this is interesting, thanks  for sharing!
in new commit, I moved it to update_product as suggested :+1: but keeping the logic as is because the delete method may not return truthy and we still want to run the following update methods.

RachL:
As a manager, and from the bulk edit product page, I did the following:

- I've change the product bamboo, from supplier "bamboo farm" to supplier "fred's farm".
- The product was removed from the OC accordingly. I've added to the OC of the new supplier without issue, and could change price, name, picture, stock without any problem from the bulk edit product page.
- I also could checkout the product.

I did all these test again on single product edit page. Everything worked!

This is ready to go :tada:

RachL:
Oh and I also checked the lettuce share report which seemed accurate :+1:

luisramos0:
ah, nice, thanks @RachL

PR #4389, opened 2019-10-21
Move Option Types from spree_backend to OFN

luisramos0:
#### What? Why?

Part of #4050

This moves option types from spree_backend to OFN. No change in functionality.

#### What should we test?
Make sure option types can be created, edited and deleted.

#### Release notes
Changelog Category: Changed


luisramos0:
this is not required if we delete the page in #4397

PR #4393, opened 2019-10-23
Avoid asking the cloud if an image exists

### LINKED ISSUE #4392: Backoffice not available for some users ? (Error 500)
## Description

Some users cannot reach the /admin URL (or any URL in the backoffice without seeing the snail).

I'm trying to find a pattern and it is possible it only affect **owners** and not **managers**. But I need more data to confirm.

Super admin cannot reach these enterprise profile as well. Error 500

## Steps to Reproduce

1. Log in as super admin on French production
2. Try to reach `/admin/enterprises/le-jardin-de-deux-main/edit`or `/admin/enterprises/hub-demo-open-food-france/edit`
3. See an error 500
4. Try to log in as manager of these hubs : same error
5. Try another hub like `/admin/enterprises/nantes-chantenay/edit` No problem. What's the difference between the two of them?
6. Try to reach the enterprise "Hub demo Open Food France" with rachel.arnould[at]openfoodfrance.org (manager) everything is working well...

## Severity

## Your Environment
* Version used: v2.5.3
* Browser name and version: Firefox 69

mkllnk:
#### What? Why?

Closes #4392

Amazon's DNS is failing at the moment and some users can't access the admin panel because of this error.

Here I'm avoiding the network query.
While `exists?` asks the storage server if the file is actually there,
`file?` just checks if we have the file name stored in the database
and the file should be there. It's much faster and less error prone.

There is also the `present?` method which does the same as `file?` but Rubocop complains about `unless ...present?`. It's probably a good idea to use a different method name as it's not exactly the same as Rails' `present?`.

#### What should we test?

- You need a staging server with Amazon S3 setup, e.g. Australian staging, UK staging. *Not* French staging.
- You need an enterprise with a logo or promo image.
- Just visit the edit enterprise page and it should display.
- If the Amazon servers are up(if you can see https://s3.amazonaws.com/ofn-prod/public/images/enterprises/logos/19/original/pots_empil_C3_A9s_3.jpg?1535969877), you can test uploading a new image.

#### Release notes

Fixed: Don't crash on enterprise admin page when Amazon S3 is down.

Changelog Category: Fixed

### DIFF
--- a/app/serializers/api/admin/enterprise_serializer.rb
+++ b/app/serializers/api/admin/enterprise_serializer.rb
@@ -62,7 +62,7 @@ class Api::Admin::EnterpriseSerializer < ActiveModel::Serializer
   #   #   medium: LOGO_MEDIUM_URL
   #   # }
   def attachment_urls(attachment, versions)
-    return unless attachment.exists?
+    return unless attachment.file?
 
     versions.each_with_object({}) do |version, urls|
       urls[version] = attachment.url(version)


Matt-Yorkley:
Looks like there could be a bit more to this bug: 

bugsnag

RachL:
OFN UK demo Hub has now cute little balloons:

https://staging.openfoodnetwork.org.uk/ofn-uk-test-hub/shop#/about

I can access the edit page both as admin and super admin :+1:

https://staging.openfoodnetwork.org.uk/admin/enterprises/ofn-uk-test-hub/edit#/images

This is ready to go!

PR #4394, opened 2019-10-23
Merge duplicated product_sub_menu partials

luisramos0:
#### What? Why?

Remove duplicated and outdated partial. The spree version did not include all entries, so, for example, when editing a product the sub menu entry for inventory was not there.

Before:
Screenshot 2019-10-23 at 12 34 47

After:
Screenshot 2019-10-23 at 12 33 30

#### What should we test?
MAke sure the product sub menu is showing correctly on all pages under product like edit variants, edit images, create product, etc.

#### Release notes
Changelog Category: Fixed
Product sub menu entries are now consistent in all pages under the Product menu.

### DIFF
--- a/app/views/admin/enterprise_groups/index.html.haml
+++ b/app/views/admin/enterprise_groups/index.html.haml
@@ -1,4 +1,4 @@
-= content_for :page_title do
+- content_for :page_title do
   = t 'admin_enterprise_groups'
 
 - if admin_user?
--- a/app/views/admin/order_cycles/index.html.haml
+++ b/app/views/admin/order_cycles/index.html.haml
@@ -1,4 +1,4 @@
-= content_for :page_title do
+- content_for :page_title do
   = t :admin_order_cycles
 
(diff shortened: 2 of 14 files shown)


filipefurtad0:
Hi @luisramos0,

I've staged this in Katuma, but found no obvious difference between the menus (before and after). Does this PR require staging on FR?

luisramos0:
hey Filipe, I think you need to be super admin to see the properties sub menu.
And if that is the case you can verify the issue is solved in this PR.

I just compared to staging FR on the edit variants page:

FR:
Screenshot 2019-10-31 at 12 42 00

ES:
Screenshot 2019-10-31 at 12 41 54

filipefurtad0:
Ok, thats it then. Had to check it as super-admin. Then becomes obvious that the PR works. Thank you for the help!

filipefurtad0:
Thanks @luisramos0. I guess I really was looking at the wrong place, sorry for that and thank you.

PR #4395, opened 2019-10-23
Delete prototypes tables and all references in pages

luisramos0:
#### What? Why?

Prototypes is a dead feature. Only one prototype create in 2012 in the AUS live database. All other instances have no prototypes. It's actually not working with the mix with option types.

#### What should we test?
We can test that product properties feature is working correctly.
Add producer properties and product properties and use them in products and the shopfront.

#### Release notes
Changelog Category: Removed
MAkes app and code simpler by removing unused feature "prototypes," a feature implemented in Spree but never used in OFN.

### DIFF
--- a/app/views/spree/admin/product_properties/index.html.haml
+++ b/app/views/spree/admin/product_properties/index.html.haml
@@ -8,14 +8,10 @@
   %ul.tollbar.inline-menu
     %li
       = link_to_add_fields Spree.t(:add_product_properties), 'tbody#product_properties', class: 'icon-plus button'
-    %li
-      %span#new_ptype_link
-        = link_to Spree.t(:select_from_prototype), available_admin_prototypes_url, remote: true, 'data-update' => 'prototypes', class: 'button icon-copy'
 
 = form_for @product, url: admin_product_url(@product), method: :put do |f|
   %fieldset.no-border-top
     .add_product_properties
-    #prototypes
     = image_tag 'select2-spinner.gif', plugin: 'spree', style: 'display:none;', id: 'busy_indicator'
 
     %table.index.sortable{"data-sortable-link" => update_positions_admin_product_product_properties_url}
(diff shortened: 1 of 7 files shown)


RachL:
@sauloperez same comment here than in option types for me. Moreover, here I think the Australian team needs to be aware of that prototype in AU live before this goes into a release ping @kirstenalarsen

luisramos0:
pls see here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4397#issuecomment-547014072

the prototype is linked to old products and variants. I checked and I couldnt find anything close to being recent related to that prototype from 2012.

luisramos0:
I guess this is our product review :-)
Screenshot 2019-10-29 at 08 29 18

sigmundpetersen:
There are conflicts. Does this need a rebase @luisramos0?

filipefurtad0:
Thank you @sigmundpetersen and @luisramos0,

I have followed the testing instructions:
- added product details and labels, and verified that they appear in the shopfront -> PASS

However, I am a bit unsure I understood what the prototype tables feature is, and may have missed something. I also looked up in the user guide and didn't find anything on this. Please let me know, if this needs any further testing. Moving to Ready to Go.

RachL:
@filipefurtad0 "labels" aka product properties can be added to the product level OR to the enterprise level. When applied to the enterprise level, all products from that enterprise have the property.

I'm not sure this is what @luisramos0 ment by producer properties, but I think it might be worth testing this part as well :) So moving back to testing

luisramos0:
correct Rachel, producer properties = enterprise level properties
I think we should give it a go on this PR.

filipefurtad0:
Thank you @RachL and @luisramos0 for pointing this out.

-> Build failed at Katuma after rebuild, allthough on slack/devops-notifications the message appears as it was deployed.

In anycase, I've added enterprise level properties, and both - enterprise level and product level - now show on the shopfront. As before, the "Select From Prototype" button is gone, from the Product Properties page, so I think all is good.

Perhaps the failing rebuild remains an issue?
Thanks again :-)

PR #4397, opened 2019-10-23
Remove option types menu entry

luisramos0:
#### What? Why?

It's not possible to create custom option types in OFN and use them in products/variants. In OFN we have static list of option types that can be used in products: weight, volume and items.
This page is a legacy spree page for when the option types could be associated with products.

#### What should we test?
We are only removing the menu entry, make sure the menu entry is not there under the Products menu.

#### Release notes
Changelog Category: Removed
Remove Option Types menu entry, this is a legacy feature from Spree that is not usable in OFN.

### DIFF
--- a/app/views/spree/admin/shared/_product_sub_menu.html.haml
+++ b/app/views/spree/admin/shared/_product_sub_menu.html.haml
@@ -1,6 +1,5 @@
 - content_for :sub_menu do
   %ul#sub_nav.inline-menu
     = tab :products, match_path: '/products'
-    = tab :option_types, match_path: '/option_types'
     = tab :properties
     = tab :prototypes


RachL:
@sauloperez I'm not sure what the product review is about. If I understand clearly, we are removing something that is anyway not working. If so, it's a good news and this PR should go forward. 
If it was not the case, this conversation should have occurred in discourse first, before any PR would have been submitted.

luisramos0:
hello @RachL I can clarify.


original slack message:
https://openfoodnetwork.slack.com/archives/CG7NJ966B/p1571695576075000

The product review required here is: "this is dead feature zone, correct? shout if you think these pages cannot be removed for some reason" :-)

RachL:
Ah ok. Yes this is dead feature as Sigmund has answered before :)

filipefurtad0:
Hi @luisramos0,

Logged in as superadmin, the "custom option types" submenu dissappeaers, when staging this PR. Before:

crop_Screenshot at 2019-11-01 11-26-06

and after:
crop_Screenshot at 2019-11-01 11-48-00

So, all good -> moving to Ready to Go.

PR #4400, opened 2019-10-24
Task to remove limit of ent. that can be created

sauloperez:
## What? Why?

It does so by updating a user's enterprise_limit attribute to the maximum integer the database supports.

This is used at least in Katuma to remove the limitation of the number of enterprises a user can create. This is the agreement the community reached for the pricing plans.

Eventually, this logic could be triggered with a button from the UI but for now this is for internal usage only.

This is the result

Screenshot from 2019-10-24 13-31-31

I followed https://blog.10pines.com/2019/01/14/testing-rake-tasks/ and https://thoughtbot.com/blog/test-rake-tasks-like-a-boss to tests the rake task.

#### What should we test?

#### Release notes

Rake task to remove the limit of enterprises a user can create.

Changelog Category: Added

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -1,0 +1,27 @@
+require 'csv'
+
+namespace :ofn do
+  desc 'remove the limit of enterprises a user can create'
+  task :remove_enterprise_limit, [:user_id] => :environment do |_task, args|
+    RemoveEnterpriseLimit.new(args.user_id).call
+  end
+
+  class RemoveEnterpriseLimit
+    # rubocop:disable Style/NumericLiterals
+    MAX_INTEGER = 2147483647
+    # rubocop:enable Style/NumericLiterals
+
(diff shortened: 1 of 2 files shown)


lib/tasks/users.rake
```diff
@@ -0,0 +1,27 @@
+require 'csv'
+
+namespace :ofn do
+  desc 'remove the limit of enterprises a user can create'
+  task :remove_enterprise_limit, [:user_id] => :environment do |_task, args|
+    RemoveEnterpriseLimit.new(args.user_id).call
```
mkllnk: Just out of curiosity, why do you store the user id in the service object instead of just doing `RemoveEnterpriseLimit.call(args.user_id)`?

sauloperez: Just inertia. I'm used to do so as most of the times classes tend to be more complex than this one.

PR #4407, opened 2019-10-27
Add tax category to sample products

luisramos0:
#### What? Why?

Sample data is broken on my local setup, for some unknown reason, Spree::Config.products_require_tax_category is true when loading sample data and so the script fails.

This Pr fixes that by adding a tax category to all products created.

#### What should we test?
Sample data now loads without errors related to tax category being mandatory.

#### Release notes
Changelog Category: Fixed
Fixed OFN sample data script.


sauloperez:
Being a bit more flexible here so something this simple doesn't get blocked

Issue #4379, opened 2019-10-17
[Performance] POC Edit Order Cycle


lin-d-hop:
## Description

As an enterprise I want to be able to create/edit an order cycle with 10000 products in an acceptable length of time.

After iterations on discussions in the #performance channel we have decided that our main UX strategy will be to minimise the disruption to the users in terms of experience.

From an implementation perspective we know that the angular in the page is so complex that a rewrite is favourable.

There are two areas of implementation for which we have questions:
1) Will it be better to split the page into steps or to keep as a single page as it is now? 
([See here for overview of steps](https://docs.google.com/document/d/1F8_5xiAzGcg7S_QkJyX6FQ5-m0qhon_UziYz9m6XJFk/edit))

2) Will it be better to implement infinite scroll or paged results when loading products per enterprise?

## Acceptance Criteria & Tests

This is a Proof of Concept. We are looking for something that roughly works, but won't stand up to thorough testing. The point of this issue is to give devs the opportunity to explore and understand the merits of different implementations. Then Product team can play and potentially run by users to understand pain points of the new design.

- Products list will be sorted alphabetically.
- 10000 product OC can be loaded in under 3(?) minutes

Matt-Yorkley:
I think this POC needs some more thought on the technical implementation. I'm not sure yet if splitting the page into 3 parts is necessary or not. The main issue is saving the user selections of included/excluded variants per supplier.

If we imagine an OC with 5 suppliers and each has 2000 products, this interface would be swamped in thousands of these boxes in each tab:

![oc-edit](https://user-images.githubusercontent.com/9029026/67334336-e4048600-f519-11e9-91ec-ee994fb96407.png)

The main issue is that every time we hit the save button we would need to post an array with 10,000 items to the server, and save them all at once. I think we'll have to move to a more asynchronous strategy where each time the user selects or deselects a variant, a request is made that saves the value for that single variant, and we don't have a big save button at the end. I suppose the "select all" button would need to make a slightly different request, and we can handle it with different logic on the server side.

Does that sound good?

Matt-Yorkley:
We also now have some very streamlined services for rendering paginated products in the shopfront, and I think we could potentially re-use some of them for rendering available products in an OC on the admin side.

Matt-Yorkley:
[...]

PR #4408, opened 2019-10-28
Remove dead code from OC edit page code

luisramos0:
#### What? Why?

Related to #4379

Delete dead code, see commit messages for details.

#### What should we test?
Create and edit OCs. Validate only variants are listed in the exchanges. Products can never be selected and master variants are also not displayed (master variants are never displayed in the system now).

#### Release notes
Changelog Category: Removed
Delete dead code, make Order cycle pages easier to maintain.


luisramos0:
yes, thanks @mkllnk 💪

luisramos0:
I messed up this branch by force pushing new stuff into it. I reverted and it's now as it was when approved :+1:

filipefurtad0:
Hey @luisramos0,

Found no issues on creating and editing order cycles. Master variants are not shown for selection, so I think all is good here -> Moving to Ready to Go.

Perhaps a minor UX issue though:
I feel the "Select All" thick should disappear, after you unselect a variant. It should only be thicked, when all variants are selected, too - which does not happen on the pic below. Would this be something to open an issue on?

image

luisramos0:
great, thanks Filipe!

That is a valid problem, you need to click twice on select all if you want to select all afterwards.
I am working on changing this list of products as a continuation of #4422, the select all button will be changed. So, you can create an issue or maybe just keep it in mind as we will need to test this in the next few weeks for sure.

filipefurtad0:
Thanks for the feedback :-)
I will keep this in mind, untill the next test is due.

Issue #2971, opened 2018-11-05
[Spree Upgrade] Soft-delete enterprisess


sauloperez:
## Description

Spree 2.0 made products soft-deletable in https://github.com/spree/spree/commit/3a9bbab32855e94484ae522b7f002867e1a5915b. That means that destroys now become updates. This is why the test below fails in `2-0-stable`

https://github.com/openfoodfoundation/openfoodnetwork/blob/50a06d1bd9a88b35381b41e95a90a266a3640264/spec/models/enterprise_spec.rb#L36-L43

Now the product doesn't get deleted from DB, then the enterprise does get deleted and that leaves the product orphan, which causes

```shell
PG::ForeignKeyViolation: ERROR:  update or delete on table "enterprises" violates foreign key constraint "spree_products_supplier_id_fk" on table "spree_products"
       DETAIL:  Key (id)=(5) is still referenced from table "spree_products".
       : DELETE FROM "enterprises" WHERE "enterprises"."id" = $1
```

The solution here is to make enterprises soft-deletable as well by means of [paranoia](https://github.com/rubysherpas/paranoia).

## Expected Behavior

When deleting an enterprise, its supplied products should be marked as deleted as well as the enterprise itself. None of them will then be visible to users. This should make `spec/models/enterprise_spec.rb:36` pass.

## Actual Behavior

`spec/models/enterprise_spec.rb:36` fails as shown above.

## Possible Fix

We need to add [paranoia](https://github.com/rubysherpas/paranoia) in our Gemfile and then run the migration to add the `deleted_at` column to the `enterprises` table. Then, enable it by calling `act_as_paranoid`.

We should watch for calls to `delete` as these will skip Paranoia's logic and thus get completely deleted from DB :warning:

kirstenalarsen:
could we check this with @myriamboure - is it going to mean that we get closer or further from requirements to handle deleting things? And if Paranoia makes soft deletes easier, will that mean we can actually soft delete enterprises - omg please say yes?

Matt-Yorkley:
It shouldn't really affect requirements for deleting data, and yes it will mean we soft-delete enterprises.

mkllnk:
I made some progress in using the new delete API in https://github.com/openfoodfoundation/openfoodnetwork/pull/3394. It makes the related specs pass. We haven't implemented soft-deleting enterprises though. Can we deal with that as a separate feature not part of the Spree upgrade?

sauloperez:
[...]

PR #4409, opened 2019-10-28
Add rake task to empty a distributor enterprise

sauloperez:
#### What? Why?

This wipes out all proxy orders, orders, inventory units and line items of the specified distributor enterprise.

This is useful for the case of a hub that's been trying the platform and then wants to clean up the mess to proceed to regular operation. This also ensures we don't charge them for these initial tests.

#### What should we test?

#### Release notes

Rake task to remove proxy orders, orders, inventory units and line items of a distributor enterprise.

Changelog Category: Added

### DIFF
--- a/lib/tasks/enterprises.rake
+++ b/lib/tasks/enterprises.rake
@@ -1,6 +1,17 @@
 require 'csv'
 
 namespace :ofn do
+  desc 'deletes all orders and associated entities of an enterprise'
+  task :truncate_distributor, [:distributor_id] => :environment do |_task, args|
+    orders = Spree::Order.where(distributor_id: args.distributor_id)
+
+    inventory_units = Spree::InventoryUnit.where(order_id: orders.pluck(:id))
+    inventory_units.delete_all
+
+    ProxyOrder.where(order_id: orders.pluck(:id)).destroy_all
+    orders.destroy_all
+  end
+
   namespace :dev do
     desc 'export enterprises to CSV'
     task export_enterprises: :environment do
--- a/spec/lib/tasks/enterprises_rake_spec.rb
+++ b/spec/lib/tasks/enterprises_rake_spec.rb
@@ -1,0 +1,45 @@
+require 'spec_helper'
+require 'rake'
+
+describe 'enterprises.rake' do
+  describe ':truncate_distributor' do
+    let!(:distributor) { create(:enterprise) }
+    let!(:order) { create(:order, distributor: distributor) }
+
+    before do
+      Rake.application.rake_require 'tasks/enterprises'
+      Rake::Task.define_task(:environment)
+      Rake::Task["ofn:truncate_distributor"].reenable
+    end
+
+    it 'destroys all orders of the specified distributor' do
+      Rake.application.invoke_task "ofn:truncate_distributor[#{distributor.id}]"
+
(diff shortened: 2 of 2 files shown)


spec/lib/tasks/enterprises_rake_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      expect(Spree::InventoryUnit.where(order_id: order.id)).to be_empty
+    end
+
+    it 'destroys their line items' do
+      create(:line_item, order: order)
+
+      Rake.application.invoke_task "ofn:truncate_distributor[#{distributor.id}]"
+
+      expect(Spree::LineItem.where(order_id: order.id)).to be_empty
+    end
```
luisramos0: I think you should check adjustments and shipments as well.

lib/tasks/enterprises.rake
```diff
@@ -1,6 +1,17 @@
 require 'csv'
 
 namespace :ofn do
+  desc 'deletes all orders and associated entities of an enterprise'
+  task :truncate_distributor, [:distributor_id] => :environment do |_task, args|
+    orders = Spree::Order.where(distributor_id: args.distributor_id)
+
+    inventory_units = Spree::InventoryUnit.where(order_id: orders.pluck(:id))
```
mkllnk: If you used `select` instead of `pluck`, it would combine it in one query.

lib/tasks/enterprises.rake
```diff
@@ -1,6 +1,17 @@
...
+    inventory_units.delete_all
```
mkllnk: I know that `delete_all` is much faster but if someone decided to add a destroy callback to inventory units, they will probably not change this class.

lib/tasks/enterprises.rake
```diff
@@ -1,6 +1,17 @@
...
+
+    ProxyOrder.where(order_id: orders.pluck(:id)).destroy_all
+    orders.destroy_all
```
mkllnk: What happens if the orders have payments? Should all this be in a transaction to roll back everything in case one thing fails to delete?

sauloperez:
> A bit like Github asking you to type the repository name before you can delete it.

yes, this feels the best we can do.

lib/tasks/enterprises.rake
sauloperez: To be clear, my only purpose is to just share what I needed now but not to cover all possible cases. IMO this script should evolve as the need arises.

sauloperez:
I was about to close this as it's not a priority and I have other things to do. I could give it a second chance so we have this know-how persisted in our codebase which will be useful if we ever implement enterprise removal as a feature.

Do you both agree on testing adjustments and shipments and merging? I'll add a note in the task itself to warn the dev this is destructive. Implementing any guard mechanism will take the time I don't have and chances are we only use this very sporadically.

mkllnk:
Let's mention #2971 here so that this code is linked to the issue of deleting enterprises. We can then close this and come back to the code when we want to implement a proper solution. Closed pull requests are not lost.

luisramos0:
agree with Maikel, close and keep the link to the issue :+1:
we will definetely come back here when enterprises delete is a priority.

Maybe we should do the same for the delete OC PR as well, link it to the delete OC issue if there is one.

PR #4411, opened 2019-10-28
Add rake tasks to remove OCs

sauloperez:
#### What? Why?

Adds one to remove a single order cycle and another one that removes all the order cycles of the specified coordinator.

This is useful for the case of a hub that's been trying the platform and then wants to clean up the mess to proceed to regular operation. This also ensures we don't charge them for these initial tests.

#### What should we test?

#### Release notes

Rake tasks to remove a single order cycle or all the ones belonging to a coordinator enterprise.

Changelog Category: Added

### DIFF
--- a/lib/tasks/order_cycles.rake
+++ b/lib/tasks/order_cycles.rake
@@ -1,0 +1,19 @@
+namespace :ofn do
+  desc 'destroys the specified order cycle'
+  task :remove_order_cycle, [:order_cycle_id] => :environment do |_task, args|
+    remove_order_cycle(args.order_cycle_id)
+  end
+
+  desc 'destroys all order cycles of the specified coordinator'
+  task :remove_coordinated_order_cycles, [:coordinator_id] => :environment do |_task, args|
+    order_cycles = OrderCycle.where(coordinator_id: args.coordinator_id)
+    order_cycles.each { |order_cycle| remove_order_cycle(order_cycle.id) }
+  end
+
+  private
+
+  def remove_order_cycle(order_cycle_id)
+    CoordinatorFee.where(order_cycle_id: order_cycle_id).delete_all
+    OrderCycle.find(order_cycle_id).destroy
+  end
+end
--- a/spec/lib/tasks/order_cycles_rake_spec.rb
+++ b/spec/lib/tasks/order_cycles_rake_spec.rb
@@ -1,0 +1,53 @@
+require 'spec_helper'
+require 'rake'
+
+describe 'order_cycles.rake' do
+  describe ':remove_order_cycle' do
+    before do
+      Rake.application.rake_require 'tasks/order_cycles'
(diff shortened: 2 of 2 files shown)


spec/lib/tasks/order_cycles_rake_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        expect {
+          Rake.application.invoke_task "ofn:remove_order_cycle[#{order_cycle.id}]"
+        }.to change(OrderCycle, :count).by(-1)
+      end
+
+      it 'removes the order cycle coordinator fees' do
+        expect {
+          Rake.application.invoke_task "ofn:remove_order_cycle[#{order_cycle.id}]"
+        }.to change(CoordinatorFee, :count).by(-1)
+      end
```
luisramos0: I think you need to check exchanges are gone as well as exchange_variants

sauloperez: Good point. I'll check that however none of this failed due to DB constraints. I managed to run this in production.

luisramos0: yeah, but is the data gone from these tables? I would just triple check we dont end up with stale inconsistent data that can/will break the app in some random part.

lib/tasks/order_cycles.rake
```diff
@@ -0,0 +1,19 @@
+namespace :ofn do
+  desc 'destroys the specified order cycle'
+  task :remove_order_cycle, [:order_cycle_id] => :environment do |_task, args|
```
mkllnk: Isn't this easier in the admin interface?

lib/tasks/order_cycles.rake
```diff
@@ -0,0 +1,19 @@
...
+    remove_order_cycle(args.order_cycle_id)
+  end
+
+  desc 'destroys all order cycles of the specified coordinator'
+  task :remove_coordinated_order_cycles, [:coordinator_id] => :environment do |_task, args|
```
mkllnk: I think that we should supply id and name to add an additional layer of safety.

lib/tasks/order_cycles.rake
sauloperez: AFAIK we don't remove the coordinator fees applied to the OC when destroying it. That's why.

lib/tasks/order_cycles.rake
sauloperez: I can ask for the coordinator's permalink, as you suggested in another PR.

sauloperez:
I appreciate @mkllnk . I pushed all the scripts after running them though. I'm happy to address all your comments before merging because I'm not in a rush to have them merged.

sauloperez:
Sorry guys, I don't have the time to finish this. This doesn't add much and it's pretty easy to redo it in the future if we need to so closing.

PR #4417, opened 2019-10-30
Ensure variants returned in #variant_relation are DISTINCT

### LINKED ISSUE #4413: Shopfronts Not Loading all products
## Description

With the new release Tamar Valley Food Hub shopfront is not loading all products.

The shopfront scrolls to a certain point and then stops. The spinner spins for a while, then it times out without all the products there.

Run this with DevTools and you'll see and interesting and indicative error:

Screenshot from 2019-10-29 15-42-27

I suggest you just load this shop yourself to see it:
https://openfoodnetwork.org.uk/tamar-valley-food-hubs/shop

## Expected Behavior

We see an endless scroll of products loading without error

## Actual Behaviour

After 3 pages of products the shopfront fails to load further

## Steps to Reproduce

1. Go to https://openfoodnetwork.org.uk/tamar-valley-food-hubs/shop
2. Scroll Down
3. They should have 500+ products in their shop
4.

## Severity

S1

## Your Environment

* Version used: Mercao Rapido
* Browser name and version: Chrome and Firefox

Matt-Yorkley:
#### What? Why?

Closes #4413

Fixes an issue with shopfront pagination where duplicates were sometimes returned in the results. This was present in OCs where; 1) multiple products from the same supplier in the OC had the exact same name and 2) the hub was using the feature for reordering products by a custom taxon list.

The issue was discovered on UK production. It probably had minimal impact beyond a single OC where both of these conditions were met.

#### What should we test?

Deploying `v2.6.0` to UK Staging will show the bug, when viewing the shopfront for Tamar Valley Food Hub and scrolling down far enough. The javascript console will show an error and the loading spinner will be stuck, when reaching the `goats milk` product.

Deploying this PR to UK Staging will fix the issue. Scrolling to the bottom of the full product list is possible. Both versions of the duplicate `goats milk` product will be displayed correctly.

#### Release notes

Fixed an issue with duplicate product names affecting shopfront pagination

Changelog Category: Fixed

### DIFF
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -15,7 +15,8 @@ class OrderCycleDistributedProducts
   def variants_relation
     order_cycle.
       variants_distributed_by(distributor).
-      merge(stocked_variants_and_overrides)
+      merge(stocked_variants_and_overrides).
+      select("DISTINCT spree_variants.*")
   end
 
   private
--- a/app/services/order_cycle_distributed_products.rb
+++ b/app/services/order_cycle_distributed_products.rb
@@ -9,7 +9,7 @@ class OrderCycleDistributedProducts
   end
 
   def products_relation
-    Spree::Product.where(id: stocked_products)
+    Spree::Product.where(id: stocked_products).group("spree_products.id")
   end
 
   def variants_relation
--- a/app/services/products_renderer.rb
+++ b/app/services/products_renderer.rb
@@ -63,10 +63,10 @@ class ProductsRenderer
     if distributor.preferred_shopfront_taxon_order.present?
       distributor
         .preferred_shopfront_taxon_order
-        .split(",").map { |id| "primary_taxon_id=#{id} DESC" }
-        .join(",") + ", name ASC"
+        .split(",").map { |id| "spree_products.primary_taxon_id=#{id} DESC" }
+        .join(", ") + ", spree_products.name ASC"
     else
-      "name ASC"
+      "spree_products.name ASC"
     end
   end
 
--- a/app/services/products_renderer.rb
+++ b/app/services/products_renderer.rb
@@ -64,7 +64,7 @@ class ProductsRenderer
       distributor
         .preferred_shopfront_taxon_order
         .split(",").map { |id| "spree_products.primary_taxon_id=#{id} DESC" }
-        .join(", ") + ", spree_products.name ASC"
+        .join(", ") + ", spree_products.name ASC, spree_products.id ASC"
     else
       "spree_products.name ASC"
     end
--- a/spec/controllers/api/order_cycles_controller_spec.rb
+++ b/spec/controllers/api/order_cycles_controller_spec.rb
@@ -187,6 +187,46 @@ module Api
       end
     end
 
+    context "with custom taxon ordering applied and duplicate product names in the order cycle" do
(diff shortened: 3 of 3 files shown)


luisramos0:
nice!!!!

we need some for of testing to validate the issue and the fix.

looking at the code I wonder/suspect Api::ProductSerializer#master is not needed and quite a few lines of code can be deleted?

app/services/order_cycle_distributed_products.rb
```diff
@@ -9,7 +9,7 @@ def initialize(distributor, order_cycle, customer)
   end
 
   def products_relation
-    Spree::Product.where(id: stocked_products)
+    Spree::Product.where(id: stocked_products).group("spree_products.id")
```
mkllnk: Is this is an unusual use of `group`? Why didn't you go for the same approach as before? `select("spree_products.*")`

Matt-Yorkley: I wanted to make sure there were no duplicates in the results, and initially tried using DISTINCT in the query here, but I ran into an obscure Rails/Postgres issue when the taxon ordering feature is applied. It caused a lot of fatal errors and took a while to figure out. I settled on using group by id as a workaround.

Further reading: 

https://stackoverflow.com/questions/12693089/pgerror-select-distinct-order-by-expressions-must-appear-in-select-list/18367248

https://github.com/rails/rails/issues/520

Matt-Yorkley:
I couldn't make a working spec for this initially, but I'll try again.

Matt-Yorkley:
Nope, I can't replicate this bug in a test...

Matt-Yorkley:
Okay, I've got it. Ready for review!

luisramos0:
@Matt-Yorkley the new specs are green in master... I'll try to understand what's missing.

kristinalim:
The specs are also green for me in master...

The second commit touching the spec (67a5a1cdc2e809fce34ff097bbdc686f8606787d) causes a failure for me, but unrelated:

```
  1) Api::OrderCyclesController with custom taxon ordering applied and duplicate product names in the order cycle displays products in correct order across multiple pages
     Failure/Error: expect(product_ids).to eq [product7.id, product8.id, product2.id]

       expected: [249, 250, 244]
            got: [250, 249, 244]

       (compared using ==)
     # ./spec/controllers/api/order_cycles_controller_spec.rb:212:in `block (3 levels) in <module:Api>'
```

luisramos0:
all right, took me a while to get this right!

First, this will fix our problem!

I have this commit that changes the spec:
https://github.com/Matt-Yorkley/openfoodnetwork/pull/8/commits/d1a01b12b199ad0f2411fa43a37e3f906bd2599c

see the commit message

I also wrote a feature spec to replicate this but I think this new controller spec is enough 💪

app/services/order_cycle_distributed_products.rb
mkllnk: Ah, okay, I ran into that issue before. It's frustrating.

filipefurtad0:
Hi @Matt-Yorkley,

Followed your testing instructions: worked just as described :tada :-)
-> moving to Ready to Go

For pics and testing notes please see:
https://docs.google.com/document/d/1DVuxciWJL4DN0_XpmZ5RigS1tzsTyZBqyOOj_s2fxDA/

Cheers!

luisramos0:
great Filipe!

PR #4426, opened 2019-11-03
Orders and Fulfilment reports

### LINKED ISSUE #4309: Using SQL query for each type of adjustment for each line item in Customer Totals report

See examples of queries:

20190924165114-multiple_queries_for_adjustments

Is it a good solution to use just a single SQL query for all of these, for all types of adjustments and all line items used in the report, and calculate these values outside the DB?

Stack Trace 1:

```
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/app/models/spree/line_item_decorator.rb:105:in `price_with_adjustments'
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/app/models/spree/line_item_decorator.rb:115:in `amount_with_adjustments'
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/lib/open_food_network/orders_and_fulfillments_report.rb:254:in `block in columns'
```

Stack Trace 2:

```
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/app/models/spree/order_decorator.rb:264:in `admin_and_handling_total'
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/lib/open_food_network/orders_and_fulfillments_report.rb:155:in `block in rules'
```

Stack Trace 3:

```
/mnt/code/system/rbenv/versions/2.1.9/lib/ruby/gems/2.1.0/bundler/gems/spree-8a8585a43cd0/core/app/models/spree/order.rb:325:in `ship_total'
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/lib/open_food_network/orders_and_fulfillments_report.rb:156:in `block in rules'
```

Stack Trace 4:

```
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/app/models/spree/order_decorator.rb:268:in `payment_fee'
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/lib/open_food_network/orders_and_fulfillments_report.rb:157:in `block in rules'
```

Stack Trace 5:

```
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/app/models/spree/line_item_decorator.rb:105:in `price_with_adjustments'
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/app/models/spree/line_item_decorator.rb:115:in `amount_with_adjustments'
/mnt/code/projects/openfoodnetwork/openfoodnetwork/src/lib/open_food_network/orders_and_fulfillments_report.rb:254:in `block in columns'
```

### Context

### Impact and timeline

This is expected to significantly reduce the response time.

### LINKED ISSUE #4314: N+1 for shipment, selected shipping rate, and shipping method in Customer Totals report

This is related to #4310. Any work for this should be coordinated with #4213 

The shipment, shipping rate, and shipping method records should be fetched in batch. At the moment, there are 2 queries (aside from query to be removed in #4310) for each non-group row in the Customer Totals report.

### Context

### Impact and timeline

This is expected to have some effect on the response time, and it will significantly reduce the number of SQL queries.

### LINKED ISSUE #4319: Eager-load associations for Variant#options_text used in Orders and Fulfillment report

I separated this from #4308, because I started looking into this before and it wasn't simple.

We are using `options_text` is used for `Spree::Variant#full_name` used in different parts of the code.

### Context

### Impact and timeline

This is expected to have observable impact on the Orders and Fulfillment report. The same strategy could be used in other parts of the code that call `Spree::Variant#full_name` for multiple variants, such as AFAIK the edit OC page.

Matt-Yorkley:
#### What? Why?

Closes #4309 
Closes #4319
Closes #4314

- Adapts `line_item_decorator#price_with_adjustments`, `order_decorator#admin_and_handling_total`, `order_decorator#payment_fee`, and `order_decorator#ship_total` so that they automatically work with eager-loaded `Spree::Adjustment` objects without triggering additional queries.

- Adds eager-loading of various associated objects to the `OrdersAndFulfilment` reports.

#### What should we test?

Orders and Fulfilment reports should work correctly.

#### Release notes

Improved loading of adjustments in Order and Fulfilment reports

Changelog Category: Changed


Matt-Yorkley:
Okay, this removes almost all of the N+1 queries from Orders and Fulfilment reports, mostly Customer Totals. 

For a selected range with a medium amount of data, total pageload times were either significantly reduced or cut in half, and in the report subtypes:

- Customer Totals report: queries reduced by **~82%**
- Supplier Totals by Distributor: queries reduced by **~21%**
- Distributor Totals by Supplier: queries reduced by **~55%**

Part of this involved importing some methods for calculating adjustments, and adapting them to use preloaded objects instead of queries. It's not very pretty, but the effects are dramatic.

Matt-Yorkley:
The last remaining issue with N+1s that applies to all reports, but also to various other parts of the app (loading products in shopfronts for example) and causes huge performance hits is the `variant.full_name` method, which calls the `#full_name` and `#options_text` methods in the `VariantAndLineItemNaming` module. As far as I can tell, this module creates serious performance problems in almost all areas of the app.

https://github.com/openfoodfoundation/openfoodnetwork/blob/master/lib/open_food_network/variant_and_line_item_naming.rb

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
```diff
@@ -190,7 +190,25 @@ def columns
...
+      end
+
+      def ship_total(order)
+        order.adjustments.select{ |a| a.originator_type == 'Spree::ShippingMethod' }.
+          map(&:amount).sum.to_f
+      end
+
+      def payment_fee(order)
+        order.adjustments.select{ |a| a.originator_type == 'Spree::PaymentMethod' }.
+          map(&:amount).sum.to_f
```
luisramos0: this is "copy paste change" to something very similar... :-( 
at least, we must move these methods close to the original methods (because they will change at the same time in the future), these 3 methods have to go to order_decorador. I am not sure about the best names but maybe admin_and_handling_total_alternative...

but better than that would be to replace the original methods. I dont see why we cant replace the original methods in order_decorator. do you?

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
```diff
@@ -210,6 +212,27 @@ def payment_fee(order)
...
+      end
+
+      def line_item_adjustments(line_item)
+        line_item.order.adjustments.select{ |a| a.source_id == line_item.id }
+      end
+
+      def shipping_method(line_items)
+        line_items.first.order.shipments.first.
+          andand.shipping_rates.andand.first.andand.shipping_method
+      end
```
luisramos0: same as for orders but worse, I dont think this is acceptable, it's adding variations to a system (adjustments) that is already a mess...

we need at least to move all this to line_item_decorator, we need to consider merging/replacing existing methods in line_item_decorador.

Also, we should probably consider creating new services for these methods (and the existing ones in the order and line_item models):
- in order : order_adjustments_service
- here: line_item_adjustments_service

lib/open_food_network/reports/helper.rb
```diff
@@ -0,0 +1,42 @@
+module OpenFoodNetwork
+  module Reports
+    module Helper
```
luisramos0: I see you tried to move everything into an helper. I dont think this is good practice, these is OFN's core business logic and needs to be kept consistent in the core, not in a report helper.
I suggest moving this to the existing models and if not possible, move both the existing ones in the models and these methods into services.

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
Matt-Yorkley: Yeah I wasn't sure how to get around this duplication and increased coupling. We can't just replace the original methods, they're closely linked but not compatible.

Matt-Yorkley: We could add these as alternate methods in `line_item_decorator` and `order_decorator`. If we try to include this directly in the original methods I think we will have a lot more branching and complexity.

Basically the originals will trigger an additional query to fetch the associated objects (which is generally useful). Here we need the same logic, but applied to eager-loaded associated objects that have already been initialized.

Matt-Yorkley: Is there some way to check whether or not the associated object have already been fetched and initialized at the point they are needed? Like:

```ruby
if adjustments_already_eager_loaded_and_initialized?
   payment_fee_from_loaded_objects
else
   payment_fee
end
```

luisramos0: maybe this? https://stackoverflow.com/questions/1376212/how-to-determine-if-rails-association-is-eager-loaded

Matt-Yorkley: Okay, I've started adapting the code using `if adjustments.loaded?` and it seems to work well. This would also mean any similar queries that eager-load adjustments should work without triggering additional queries...

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
Matt-Yorkley: This method was needed because now we are eager-loading more associated data, and need to access: `line_items.first.order.shipments.first.shipping_rates.first.shipping_method` in multiple places, but in the result set any of these links in the chain of associations can be `nil`.

app/models/spree/adjustment_decorator.rb
```diff
@@ -2,6 +2,9 @@
 
 module Spree
   Adjustment.class_eval do
+    PAYMENT_FEE_SCOPE = { originator_type: 'Spree::PaymentMethod' }.freeze
+    SHIPPING_SCOPE = { originator_type: 'Spree::ShippingMethod' }.freeze
+    ELIGIBLE_SCOPE = { eligible: 't' }.freeze
```
luisramos0: this is "eligible: true" in spree

app/models/spree/adjustment_decorator.rb
```diff
@@ -19,7 +22,9 @@ module Spree
 
     scope :with_tax,       -> { where('spree_adjustments.included_tax > 0') }
     scope :without_tax,    -> { where('spree_adjustments.included_tax = 0') }
-    scope :payment_fee,    -> { where(originator_type: 'Spree::PaymentMethod') }
+    scope :payment_fee,    -> { where(PAYMENT_FEE_SCOPE) }
+    scope :shipping,       -> { where(SHIPPING_SCOPE) }
+    scope :eligible,       -> { where(ELIGIBLE_SCOPE) }
```
luisramos0: why constants? why not just the normal definition of the scope?

app/services/order_adjustments_fetcher.rb
```diff
@@ -0,0 +1,64 @@
+class OrderAdjustmentsFetcher
```
luisramos0: this needs a class level comment or a comment somewhere explaining the reason behind using .loaded?

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
```diff
@@ -190,7 +190,16 @@ def columns
...
+        [{ variant: { product: :supplier },
+           order: [:bill_address, :ship_address, :order_cycle, :adjustments, :payments,
+                   :user, :distributor, shipments: { shipping_rates: :shipping_method }] }]
+      end
+
+      private
+
+      def shipping_method(line_items)
+        line_items.first.order.shipments.first.
+          andand.shipping_rates.andand.first.andand.shipping_method
```
luisramos0: I think this change is incorrect, through OrderShipment you have order.shipping_method so this can stay:
line_items.first.order.shipping_method

I think that selecting shipping_rates.andand.first is wrong, we should use shipment.shipping_method that gets the  selected shipping_method

app/models/concerns/adjustment_scopes.rb
```diff
@@ -0,0 +1,7 @@
+module AdjustmentScopes
+  extend ActiveSupport::Concern
+
+  PAYMENT_FEE_SCOPE = { originator_type: 'Spree::PaymentMethod' }.freeze
+  SHIPPING_SCOPE = { originator_type: 'Spree::ShippingMethod' }.freeze
+  ELIGIBLE_SCOPE = { eligible: 't' }.freeze
+end
```
luisramos0: missing empty line at the end

app/services/order_adjustments_fetcher.rb
```diff
@@ -0,0 +1,66 @@
...
+      adjustments.public_send scope
+    end
+  end
+
+  def admin_and_handling_fees
+    if adjustments.loaded?
+      adjustments.select do |adjustment|
+        scope_hash = ELIGIBLE_SCOPE
+
+        adjustment.public_send(scope_hash.keys.first) == scope_hash.values.first &&
```
luisramos0: this scope_hash trick is not obvious to understand, maybe this can extracted to a separate method with a good name in some way?
name of the method: what is this == condition doing? 
maybe also a variable with a good name for scope_hash.keys.first

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
Matt-Yorkley: If I change this back to `line_items.first.order.shipping_method` and check the generated queries again, there is a huge increase in queries to `shipping_rate` (one for every shipment in every order in the results). It looks like this:

shipping_rates

We are essentially doing: `Spree::LineItem.includes(order: { shipments: { shipping_rates: :shipping_method } })` to get the shipping method eager-loaded correctly, and it seems that a consequence of that is that calling `line_item.shipping_method` directly triggers a new query ever time, whereas the slightly messy method above with the full chain (and lots of `#andand`) does not trigger new queries.

app/services/order_adjustments_fetcher.rb
Matt-Yorkley: I've extracted a couple of methods here to make it more readable.

```ruby
def adjustments_eager_loaded?
    adjustments.loaded?
end

def match_by_scope(adjustment, scope)
    adjustment.public_send(scope.keys.first) == scope.values.first
end
```

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
luisramos0: ok, at least we need to make it correct

in spree Shipment.rb:
```
    def shipping_method
      selected_shipping_rate.try(:shipping_method) || shipping_rates.first.try(:shipping_method)
    end
```

shipping_rates.where(selected: true).first || shipping_rates.first.try(:shipping_method)

so, the alternative would be:
```
 def shipping_method(line_items)
      shipping_rates = line_items.first.order.shipments.first.andand.shipping_rates
      shipping_rates.andnad.where(selected: true).andand.first || shipping_rates.andand.first.try(:shipping_method)
end
```

but again the suggestion could be the same as for the orders method.
we can re-implement shipment_decorator#shipping_method and use the is_loaded?

I am sorry I am asking for so many changes!

luisramos0:
ah! option values, does it work? that's diabolical!  👏 
I think it should be a separate PR Matt, it will require some specs of their own and also specific manual testing.

Matt-Yorkley:
> I think it should be a separate PR

Yeah, that's a good idea. It'll touch on lots of other things as well when we start correctly eager-loading those objects on all queries that fetch multiple variants. :+1:

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
Matt-Yorkley: Can an order order have multiple shipping methods in OFN?

Matt-Yorkley: Okay, I've adapted `Spree::Shipment#selected_shipping_rate` and rechecked the queries. LGTM. :+1:

Matt-Yorkley: Hmmm... the build is broken now, and it indicates the shipping methods selected previously were correct, and the ones selected now are incorrect...

Matt-Yorkley: I adapted it with:

```ruby
def selected_shipping_rate
      if shipping_rates.loaded?
        shipping_rates.find(&:selected)
      else
        shipping_rates.where(selected: true).first
      end
 end
```

...but a few specs that check for the correct selection of a `shipping_rate` fail. I think the original method works because it's part of a single query with a number of joins, so the `shipping_rate` that gets selected is automatically the one that's related to both the `shipment` and `shipping_method`. Does that make sense?

lib/open_food_network/permissions.rb
```diff
@@ -100,9 +90,9 @@ def visible_line_items
       editable = editable_line_items.pluck(:id)
 
       # Any from visible orders, where the product is produced by one of my managed producers
-      produced = Spree::LineItem.where(order_id: visible_orders.pluck(:id)).joins(:product).
-        where("spree_products.supplier_id IN (?)",
-              managed_enterprises.is_primary_producer.pluck(:id))
+      produced = Spree::LineItem.where(order_id: visible_orders.pluck(:id)).
```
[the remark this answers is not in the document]
Matt-Yorkley: I started cleaning up up this file a little bit (there's one or two less queries triggered now, and it's a bit cleaner) but I realised it would be a much bigger job to get real results here, so I didn't want to delve any deeper. I've mentioned it elsewhere in related open issues.

Matt-Yorkley:
Rebased for Ruby

luisramos0:
rebased to resolved conflicts. no brainer. ready for testing with a green build.

RachL:
Ok so I've tested:
- web and csv display
- filtering : producer, hub, date OC
- admin, super admin
- All sub reports.

Didn't see anything wrong. This is ready to go!

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
luisramos0: I just realized that I forgot about this point...
I know this has been merged but I think this is a bug.

I dont think the shipping method is selected in any other way while fetching the DB, it's only selected in the code: shipment.shipping_method...
I think we need to verify this.

lin-d-hop: @luisramos0 I'll look in release testing.
Can you give a little more indication of what I should specifically test? That shipping methods are correctly pulled into reports? Is this general or is there a specific edge case I am look at?

luisramos0: this is specific to the customer_totals report.
to try to see a bug here we would need an order with multiple shipping methods and verify that the shipping method shown in the report is not the one selected by the user.
but I think the current code selects the first shipping method, so we need to select the second shipping method available for that order.

lin-d-hop: Ok I'm confused. The UI never allows more than one shipping method. I know that the data model generally allows this but this is always limited in the UI.
I've certainly never seen a case of an order with >1 shipping method and have no idea how it would come about.

luisramos0: yes, 1 order = 1 shipping method but when you checkout there is a list of shipping methods you can choose and all those shipping methods are copied to the order in the database.

luisramos0: the one you select is marked in a field shipping_rate.selected

lin-d-hop: Ok. Lots of reports use the shipping_method.first approach I believe. This is not a new thing is it?

luisramos0: it is a new thing, I am pretty sure this is the only place where we are doing this.

lin-d-hop: And you are right that this introduces a bug. And I think this needs to be fixed before the release.

PR #4431, opened 2019-11-05
Fix migration mismatch

Matt-Yorkley:
It looks like this was probably changed whilst resolving a merge conflict somewhere. The number in the schema doesn't match the number in the last migration file, which is breaking the `ofn-install` CI build (as well as any new deployments on fresh servers).

https://travis-ci.org/openfoodfoundation/ofn-install/jobs/606429808#L2109

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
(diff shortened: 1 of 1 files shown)


luisramos0:
I am going to merge this now.

PR #4434, opened 2019-11-06
Don't test existence of a field used in next spec

sauloperez:
#### What? Why?

If the next example doesn't find the quick_search field it will fail so
no need to duplicate it with the consequent costly test setup.

#### What should we test?

Nothing this affects the automated tests only.

#### Release notes

Remove duplicate feature spec for Bulk Order Management quick search

Changelog Category: Removed


sauloperez:
I found it while checking the spec just below it. We might change that one in the future.

PR #4447, opened 2019-11-08
Clickable steps on the OC edit page

luisramos0:
#### What? Why?

This PR makes the steps in the OC edit page clicable. The user can jump to a specific step now by clicking the name of the step: General Settings, Incoming or Outgoing.

#### What should we test?
Verify the step names are clickable and the destination pages are correct.

#### Release notes
Changelog Category: Changed
The steps in the Order Cycle edit page are now clickable so it easier for the user to go to the page desired: General Settings, Incoming or Outgoing.

### DIFF
--- a/app/views/admin/order_cycles/_wizard_progress.html.haml
+++ b/app/views/admin/order_cycles/_wizard_progress.html.haml
@@ -8,11 +8,11 @@
       = t("admin.order_cycles.wizard_progress.outgoing")
   - else
     %li{ class: "#{'current' if action_name == 'edit'}" }
-      %a{href: "#{main_app.edit_admin_order_cycle_path(@order_cycle)}"}
+      %a{ href: main_app.edit_admin_order_cycle_path(@order_cycle) }
         = t("admin.order_cycles.wizard_progress.edit")
     %li{ class: "#{'current' if action_name == 'incoming'}" }
-      %a{href: "#{main_app.admin_order_cycle_incoming_path(@order_cycle)}"}
+      %a{ href: main_app.admin_order_cycle_incoming_path(@order_cycle) }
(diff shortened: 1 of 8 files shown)


app/views/admin/order_cycles/_wizard_progress.html.haml
```diff
@@ -1,3 +1,18 @@
...
+  - if action_name == 'new'
+    %li{ class: "current" }
+      = t("admin.order_cycles.wizard_progress.edit")
+    %li
+      = t("admin.order_cycles.wizard_progress.incoming")
+    %li
+      = t("admin.order_cycles.wizard_progress.outgoing")
+  - else
+    %li{ class: "#{'current' if action_name == 'edit'}" }
+      %a{href: "#{main_app.edit_admin_order_cycle_path(@order_cycle)}"}
```
mkllnk: Why not `%a{href: main_app.edit_admin_order_cycle_path(@order_cycle)}`? Is that due to some Haml bug?

luisramos0: string interpolation was needed in a previous version of the code, now it's not. I removed it in a new commit. thanks!

luisramos0:
blocked: this can only be tested after #4422

luisramos0:
rebased to resolve conflicts

luisramos0:
looks like #4422 is looking good so this can now be tested.

This PR should not be merged before #4422, we need to merge #4422 and then switch the target branch of this PR. But we can test this meanwhile.

RachL:
Apparently I can't take a screenshot that shows my mouse pointer, but the pointer does change when moving the pointer to the steps title, so you understand it is clickable.

This is good to go :) 

Same as before I will the blocker label to be sure we merge in order.

PR #4450, opened 2019-11-08
Move return authorizations from spree_backend to OFN

luisramos0:
#### What? Why?

Part of #4050

This moves return authorizations from spree_backend to OFN. No change in functionality.

#### What should we test?
Make sure the return authorizations page is work correctly for a given order.

#### Release notes
Changelog Category: Changed
Imported views related to return authorizations management from spree_backend to make OFN more independent of Spree.

### DIFF
--- a/app/controllers/spree/admin/return_authorizations_controller.rb
+++ b/app/controllers/spree/admin/return_authorizations_controller.rb
@@ -1,0 +1,24 @@
+module Spree
+  module Admin
+    class ReturnAuthorizationsController < ResourceController
+      belongs_to 'spree/order', find_by: :number
+
+      update.after :associate_inventory_units
+      create.after :associate_inventory_units
+
+      def fire
+        @return_authorization.public_send("#{params[:e]}!")
+        flash[:success] = Spree.t(:return_authorization_updated)
+        redirect_to :back
+      end
+
+      protected
+
+      def associate_inventory_units
+        (params[:return_quantity] || []).each do |variant_id, qty|
+          @return_authorization.add_variant(variant_id.to_i, qty.to_i)
+        end
+      end
+    end
+  end
+end
--- a/app/views/spree/admin/return_authorizations/_form.html.haml
+++ b/app/views/spree/admin/return_authorizations/_form.html.haml
@@ -1,0 +1,48 @@
+%div
+  %table.index
+    %thead
+      %tr
+        %th= t('.product')
+        %th= t('.quantity_shipped')
+        %th= t('.quantity_returned')
+        %th= t('.return_quantity')
+    %tbody
+      - @return_authorization.order.shipments.shipped.collect{|s| s.inventory_units.all}.flatten.group_by(&:variant).each do | variant, units|
+        - tr_class = cycle('odd', 'even')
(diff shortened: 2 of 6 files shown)


app/controllers/spree/admin/return_authorizations_controller.rb
```diff
@@ -0,0 +1,24 @@
+module Spree
+  module Admin
+    class ReturnAuthorizationsController < ResourceController
+      belongs_to 'spree/order', find_by: :number
+
+      update.after :associate_inventory_units
+      create.after :associate_inventory_units
+
+      def fire
+        @return_authorization.public_send("#{params[:e]}!")
```
mkllnk: This is a terrible pattern for security. I can't even see any authorisation here. The client can call any method ending in `!` on this object. Good to have that visible in our code now.

RachL:
@luisramos0 what or where is the return authorizations page?  :grin:

luisramos0:
ahah, it's a dead feature but I think we decided to keep it at some point during the upgrade to spree v2...
there is 1 authorization in Uk live, 2 in AU and 3 in France...

anyway, it's in orders, first you have to ship the order to be able to register return authorizations.
If there are issues in this PR I think it's better if we verify the feature in the current version to verify if it's a new problem or an existing one. Thanks!

filipefurtad0:
Hey @luisramos0,

I am not so familiar with this feature, but I had a look before and after staging the PR, and all behaved similarly. I was able to create, edit, delete Return Authorizations, and these had the effect on the values in the orders table.

I found some small issues/possible improvements, but I think these are not related with this PR:

- in the field Amount / RMA any value in quantity is accepted. This may lead to negative values on the orders table (see below). Is it supposed to be so?
- similarly, I guess Return Quantity should be =< Quantity Shipped

image


Should I open Issues on these? Or is this rather a fearture which will be removed in the future?

I moved it to Ready to Go, but please let me know if there is something more specific I should check. Thank you! :-)

app/controllers/spree/admin/return_authorizations_controller.rb
Matt-Yorkley: Um... I'm not sure we should merge this as is?

luisramos0:
rebased to resolve conflict.

Thanks Filipe!
Yeah, this is such a corner of the app that I think we can leave those issues as they are...

Merging.

app/controllers/spree/admin/return_authorizations_controller.rb
Matt-Yorkley: Or at least make an issue to change this line, and put it at the top of Tech Debt.

luisramos0: this is live code .
this is not only spree 2 code but also spree master code 🙈  🤕 
https://github.com/spree/spree/blob/98fa319554e9bad01fee9a9f56500eea99847d29/backend/app/controllers/spree/admin/return_authorizations_controller.rb#L11

the security risk is really just adding return authorizations.

I'll create a tech debt issue.
https://github.com/openfoodfoundation/ofn-security/issues/21

PR #4452, opened 2019-11-08
Make "Ready for" field show a red border when added to the list of outgoing exchanges

luisramos0:
#### What? Why?
This PR makes the red border on the pick up time (ready for) show immediately after a new distributor is added to the list of outgoing exchanges.

Tech comment: I tried quite a few things but the new form entry was not picking up the state (dirty). setTimeout was the only way to make this work...

#### What should we test?
Add a new distributor to the OC, the save button doesn't become active and now it's obvious why: the ready for field has a red border.

#### Release notes
Changelog Category: Changed
The "Ready for" field in the outgoing settings of the Order Cycle shows a red border indicating clearly it is a mandatory field.

### DIFF
--- a/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
@@ -1,5 +1,5 @@
 angular.module('admin.orderCycles')
-  .controller 'AdminOrderCycleExchangesCtrl', ($scope, $controller, $filter, $window, $location, OrderCycle, Enterprise, EnterpriseFee, Schedules, RequestMonitor, ocInstance, StatusMessage) ->
+  .controller 'AdminOrderCycleExchangesCtrl', ($scope, $controller, $filter, $window, $location, $timeout, OrderCycle, Enterprise, EnterpriseFee, Schedules, RequestMonitor, ocInstance, StatusMessage) ->
     $controller('AdminEditOrderCycleCtrl', {$scope: $scope, ocInstance: ocInstance, $location: $location})
 
     $scope.supplier_enterprises = Enterprise.producer_enterprises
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
```diff
@@ -37,5 +37,12 @@ angular.module('admin.orderCycles')
       $event.preventDefault()
       OrderCycle.addDistributor($scope.new_distributor_id)
 
+    $scope.setPickupTimeFieldDirty = (index) ->
+      setTimeout ->
+        last_outgoing_exchange_pickup_time_field_name = "order_cycle_outgoing_exchange_" + index + "_pickup_time"
+        $scope.order_cycle_form[last_outgoing_exchange_pickup_time_field_name].$setDirty()
+        $scope.$apply()
```
Matt-Yorkley: Is `$scope.$apply()` definitely needed here?

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
```diff
@@ -37,5 +37,12 @@ angular.module('admin.orderCycles')
       $event.preventDefault()
       OrderCycle.addDistributor($scope.new_distributor_id)
 
+    $scope.setPickupTimeFieldDirty = (index) ->
+      setTimeout ->
```
Matt-Yorkley: I think in Angular it's recommended to use the built-in `$timeout` service instead of `window.setTimeout`. I think `setTimeout` can have issues in different browsers, maybe the `$timeout` service applies some workarounds?

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
Matt-Yorkley: Update; if you use the `$timeout` service, you can remove `$scope.$apply()`.

luisramos0: really nice one @Matt-Yorkley, apply was needed yes but with $timeout it's not needed. nice! new commit.

luisramos0:
this can be tested after #4422

luisramos0:
looks like #4422 is looking good so this can now be tested.

This PR should not be merged before #4422, we need to merge #4422 and then switch the target branch of this PR. But we can test this meanwhile.

luisramos0:
looks like #4422 is looking good so this can now be tested.

This PR should not be merged before #4422, we need to merge #4422 and then switch the target branch of this PR. But we can test this meanwhile.

RachL:
@jenRSmith I'm sorry I will test this one as well because it's kind of a package with the new OC page as the release will be prepared tomorrow.

RachL:
Looking good!

image


I'm moving to ready to go and adding the block label so that we are sure the PR are merged in order.

PR #4455, opened 2019-11-11
In the OC edit page, load products only when each products tab is opened by user

### LINKED ISSUE #4144: OC edit performance: Angular

I think the Angular for edit Order Cycles will need to be reworked if we want 10000 products in an OC. I found this code comment, which kind of sums it up:

https://github.com/openfoodfoundation/openfoodnetwork/blob/86aeb6a3c796ae09e15b5a576b4e8424e076aa37/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee#L79-L84

These services will probably both need looking at:

[app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee)
[app/assets/javascripts/admin/order_cycles/services/enterprise.js.coffee](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/admin/order_cycles/services/enterprise.js.coffee)

### Context

Working on OC performance.

luisramos0:
#### What? Why?
Non tech explanation: there's no functional change in this PR. This is a preparation for loading exchange products in pages which will allow us to handle thousands of products in each exchange.

Continuation of #4379 and starts #4476
Closes #4144

Tech explanation: here I basically moved the repeat statement
"supplied_products | filter:productSuppliedToOrderCycle" to a new server endpoint
and that will allow us to improve it and convert it into a optimized (and paginated?) DB query as well as remove the algorithm from the js side, namely, OrderCycle.productSuppliedToOrderCycle.

The supplied products of both incoming and outgoing exchanges are now loaded into separate data structures (one for each enterprise, no global supplied_products anymore) and can now be loaded when the panel of products is opened for the first time.
This is happening on the outgoing exchanges. The incoming exchanges are still loading all the products on page load so that the total number of variants is displayed to the user. We can LATER move this products count to a specific server side call.

Here we maintain all the exchange "editable_*" and "visible_*" data structures, that data comes from the server as part of each exchange and is applied as a filter in the ng-repeat. We can improve this later by moving it as well to this new ExchangeProductsController!

The next big step would be to build pagination on the exchangeproducts controller.

#### What should we test?
This is a big change. We need to validate every aspect of the Order Cycle create and edit pages, both single enterprise and multiple enterprise.

#### Release notes
Changelog Category: Changed
The Order Cycle edit page is now faster in some situation as less data is loaded to the browser.


app/assets/javascripts/templates/admin/panels/exchange_distributed_products.html.haml
```diff
@@ -11,7 +11,7 @@
 
     .exchange-products
       -# Scope product list based on permissions the current user has to view variants in this exchange
-      .exchange-product{'ng-repeat' => 'product in supplied_products | filter:productSuppliedToOrderCycle | visibleProducts:exchange:order_cycle.visible_variants_for_outgoing_exchanges | orderBy:"name"' }
+      .exchange-product{'ng-repeat' => 'product in enterprises[exchange.enterprise_id].supplied_products | filter:visibleProducts:exchange:order_cycle.visible_variants_for_outgoing_exchanges | orderBy:"name"' }
```
luisramos0: this line change is the single point of this PR.
- supplied_products used to contain all products from all enterprises, it's not available any longer.
- now we have enterprises[exchange.enterprise_id].supplied_products that contains the supplied products for each enterprise and is loaded from the server only when the user opens the products panel and is already filtered on the server (no need for expensive productSuppliedToOrderCycle)

app/controllers/api/exchanges_products_controller.rb
```diff
@@ -0,0 +1,134 @@
...
+    # Lists products for an Enterprise in an Order Cycle
+    #   This is the same as index below but when the Exchange doesn't exist yet
+    #
+    #   Parameters are: enterprise_id, order_cycle_id and incoming
+    #     order_cycle_id is optional if for incoming
+    def show
+      load_data_from_params
+
+      render_products
+    end
```
luisramos0: this action should be index as well, but... 
the index endpoint corresponds to api/exchanges/{exchange_id}/products
but when there's no exchange_id yet, we call api/exchanges/products and that is sent to the action show of the exchanges resource (not the exchange_products resource).
To redirect to this controller I did this in the routes:
`resources :exchanges, only: [:show], controller: 'exchanges_products' do`
This make api/exchanges/products redirect to this show action in the ExchangeProducts controller.

This is not good, is it better to have this a new ExchangesController? Is there a way to redirect this URL (api/exchanges/products) also to the index action in the ExchangesProductsController?

app/controllers/api/exchanges_products_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+      enterprises_for_outgoing_exchange.each do |enterprise|
+        products.push( *supplied_products(@exchange.order_cycle, enterprise).to_a )
+
+        products.each do |product|
+          unless product_supplied_to_order_cycle?(product)
+            products.delete(product)
+          end
+        end
+      end
+      products
```
luisramos0: this array work (*) was just a literal conversion of the angular algorithm and is transformed, in a later commit, in a normal AR query with relations.

app/assets/javascripts/templates/admin/panels/exchange_distributed_products.html.haml
Matt-Yorkley: Awesome! Is there any chance we can strip out some of that Angular filtering on `visibleProducts` at some point as well?

Matt-Yorkley: I have a feeling it's something we want to do server-side instead of in the browser...

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
```diff
@@ -1,7 +1,27 @@
-angular.module('admin.orderCycles').controller 'AdminOrderCycleIncomingCtrl', ($scope, $controller, $location, Enterprise, ocInstance) ->
+angular.module('admin.orderCycles').controller 'AdminOrderCycleIncomingCtrl', ($scope, $controller, $location, Enterprise, OrderCycle, ocInstance) ->
   $controller('AdminOrderCycleExchangesCtrl', {$scope: $scope, ocInstance: ocInstance, $location: $location})
 
   $scope.view = 'incoming'
 
-  $scope.enterpriseTotalVariants = (enterprise) ->
-    Enterprise.totalVariants(enterprise)
+  $scope.exchangeTotalVariants = (exchange) ->
+    return unless this.enterprises? && this.enterprises[exchange.enterprise_id]?
```
Matt-Yorkley: I could be totally wrong here, but I don't think referencing `this` is normally used in Angular. In the context of a controller I think `$scope` is essentially `this`?

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
```diff
@@ -29,13 +28,29 @@ angular.module('admin.orderCycles')
...
-    $scope.addDistributor = ($event) ->
-      $event.preventDefault()
-      OrderCycle.addDistributor($scope.new_distributor_id)
-
     $scope.removeDistributionOfVariant = (variant_id) ->
       OrderCycle.removeDistributionOfVariant(variant_id)
+
+    $scope.loadExchangeProducts = (scope, exchange) ->
+      return if scope.enterprises[exchange.enterprise_id].supplied_products_fetched?
+      scope.enterprises[exchange.enterprise_id].supplied_products_fetched = true
```
Matt-Yorkley: This passing of scope objects between two controllers looks pretty funky. I think if the controllers are on the same page and nested in the same elements they can probably already share a scope? 

Otherwise generally if you need to pass things between multiple controllers, it's better to create a service that holds the shared data and have each controller interact with it.

app/assets/javascripts/admin/order_cycles/services/order_cycle.js.coffee
```diff
@@ -71,18 +75,6 @@ angular.module('admin.orderCycles').factory 'OrderCycle', ($resource, $window, S
     removeExchangeFee: (exchange, index) ->
       exchange.enterprise_fees.splice(index, 1)
 
-    productSuppliedToOrderCycle: (product) ->
```
Matt-Yorkley: :tada: :fire: :fire:

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
```diff
@@ -29,13 +28,29 @@ angular.module('admin.orderCycles')
...
+      incoming = true if $scope.view == 'incoming'
+      params = { exchange_id: exchange.id, enterprise_id: exchange.enterprise_id, order_cycle_id: $scope.order_cycle.id, incoming: incoming}
+      Product.index params, (products) ->
+        scope.enterprises[exchange.enterprise_id].supplied_products = products
+
+    # Register listeners to capture first toggle open of the products panel of the exchange
+    $scope.exchangeListChanged = ->
+      panelRows = angular.element(".panel-row")
+      for panelRow in panelRows
+        panelCtrl = angular.element(panelRow).controller('panelCtrl')
```
Matt-Yorkley: This looks a bit funky as well, picking the elements out of the DOM by their CSS selectors, then triggering methods on their controllers...

I'd have to take a bit of a deeper look before I could give a decent suggestion though.

app/assets/javascripts/templates/admin/panels/exchange_distributed_products.html.haml
luisramos0: yes!
I left these for later to avoid huge PRs. I created issues now:
visible (filter only) - https://github.com/openfoodfoundation/openfoodnetwork/issues/4480
editable (these need a new flag on each variant) - https://github.com/openfoodfoundation/openfoodnetwork/issues/4481

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
luisramos0: they are not exactly equivalent.
at some point $scope was not working and this was. there's a hirerchy of scopes and controllers here... maybe that was causing the trouble.
anyway, fixed in new commit. 26d3219deec3ecac061190f0db1a75774056af9f
thanks :+1:

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
```diff
@@ -3,5 +3,21 @@ angular.module('admin.orderCycles').controller 'AdminOrderCycleIncomingCtrl', ($
...
+    return unless enterprise.supplied_products?
+
+    enterprise.numVariants = $scope.countVariants(enterprise.supplied_products)
+
+  $scope.countVariants = (products) ->
+    return 0 unless products
+
+    numVariants = 0
+    for product in products
+      numVariants += product.variants.length
```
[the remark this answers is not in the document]
luisramos0: this method is deleted in follow up PR #4471 as this count is done on the server. do you mind if we ignore this?

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
luisramos0: yes, thanks for flagging this.
I was having some trouble with some scopes not having the info I needed. but now it all works well just using $scope. New commit: 217ae774864f3a0d51df5be48744af3010fa6082

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
luisramos0: yes 🙈 
all I need is to "run some code when user opens the panel"
I could not get to something better than this 🤕

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
```diff
@@ -39,3 +39,26 @@ angular.module('admin.orderCycles')
...
+    initPanel = (scope)  ->
+      # TEMP Insert some dummy data into enterprise.supplied_products so that the UI keeps on working
+      scope.enterprises[scope.exchange.enterprise_id].supplied_products = [{id: 2, name: "asdasd" + scope.exchange.id }]
+
+    # Register listener to capture first toggle open of the products panel
+    exchangeProdutsInitialized = []
+    registerToggleListener = ->
+      panelRows = angular.element(".panel-row")
+      if panelRows.length == 0
+        $timeout(registerToggleListener, 500)
```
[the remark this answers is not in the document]
luisramos0: yes, I managed to remove the timeout. but I couldnt get to a clean version for the process of registering the listener for panel open... same as Matt's comment above 🤕

app/assets/javascripts/admin/order_cycles/services/product.js.coffee
```diff
@@ -0,0 +1,22 @@
...
+  {
+    Product: Product
+    products: {}
+    loaded: false
+
+    index: (params={}, callback=null) ->
+      Product.index params, (data) =>
+        @products[params.enterprise_id] = []
+        for product in data
+          @products[params.enterprise_id].push(product)
```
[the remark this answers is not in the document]
luisramos0: ah, thanks for commenting, I just realized this is completely unnecessary!
The original supplied_products needed to be kept at the service but now we have the products in the controller... no need to keep any of this, I just pass the data fetched to the callback! Thanks 💪 
New commit: 18f2f339db148f574d5847b49dd45fe43f1fb8ed

app/assets/javascripts/admin/order_cycles/services/product.js.coffee
```diff
@@ -0,0 +1,22 @@
+angular.module('admin.orderCycles').factory('Product', ($resource) ->
+  Product = $resource('/admin/exchanges/:exchange_id/products.json', {}, {
```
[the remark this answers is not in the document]
luisramos0: yes, agree, this should be called ProductResource in this case.
In terms of the service name, I am not using byEnterprise as you suggest because I removed the internal data structure (see comment and commit below).

So, I renamed the service to something a lot more appropriate ExchangeProduct and the resource to ExchangeProductResource. New commit: d68bcb7a1cd2ce5157b99b2c8f20b06dccc6e279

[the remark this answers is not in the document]
luisramos0: I think this point is resolved, right?

config/routes/admin.rb
```diff
@@ -18,6 +18,10 @@
       end
     end
 
+    resources :exchanges do
+      resources :products, controller: 'exchanges_products'
```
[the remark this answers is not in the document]
luisramos0: ah, yes ExchangeProductsController is better than ExchangesProductsController :+1:
I will rename it.
the single resource here is a product in an exchange, the param[:id] would be for that.
I think it's correct that the exchange_id comes in params[:exchange_id]

app/controllers/api/exchanges_products_controller.rb
```diff
@@ -1,6 +1,16 @@
 module Api
   class ExchangesProductsController < Api::BaseController
-    skip_authorization_check only: [:index]
+    skip_authorization_check only: [:index, :show]
+
+    def show
```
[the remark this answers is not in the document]
luisramos0: ExchangeProductsController is about products that CAN be added to an exchange. I'll add this to the controllers documentation. I dont think this can be mixed up with exchange controller that handles data in the exchange. Even for handling adding variants to an exchange I'd probably go for a new controller called ExchangeVariants.

I think that after I extracted the ExchangeProductsRenderer this ExchangeProductsController became nice and small just handling the different combinations of the params given.

The only major standards/consistency flaw here is that I cant make 
api/exchanges/products (without exchange id) redirect to the ExchangeProductsController#index, it goes to ExchangeController#show and manage to redirect it to ExchangeProductsController#show which is better but not nice...

This controller resource is an product in an exchange. So the show action should

config/routes/admin.rb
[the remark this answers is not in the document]
luisramos0: ok, renamed :+1:
2f7700c9c380e1f6f6635afd538dc84288aec2c5

app/controllers/api/exchanges_products_controller.rb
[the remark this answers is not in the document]
luisramos0: ah, I managed to make api/exchanges/products go to "exchange_products#index" 🎉 
e177dc5cd5b0c91481c00a95e04640150987b056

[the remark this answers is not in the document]
luisramos0: ah, I managed to make api/exchanges/products go to "exchange_products#index" 🎉
e177dc5

app/controllers/api/exchanges_products_controller.rb
luisramos0: this is not fixed :tada:

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
luisramos0: ah, I managed to remove this stuff with a ng-init 🎉  
89f83f0f52802b6e7e7b1d29394f805ef589f688

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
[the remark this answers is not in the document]
luisramos0: ok, I managed to remove this  🎉  
89f83f0f52802b6e7e7b1d29394f805ef589f688

luisramos0:
hey @mkllnk and @Matt-Yorkley 
Thanks a lot for your reviews!!!
I think I managed to address all the points raised 🎉 
Back to you ;-)

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
Matt-Yorkley: Will the ng-init in that last commit load all the panel data when the page loads instead of on click?

luisramos0: no, because the panel template is only loaded when the panel is opened. I tested this :+1:

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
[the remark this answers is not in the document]
mkllnk: All good then.

app/controllers/api/exchanges_products_controller.rb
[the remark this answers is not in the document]
mkllnk: Hm, I think I preferred to have two separate methods instead of if-else. It means that Rails is doing the if-else for you. But it's not that important. I'm happy for you to leave it as is.

sigmundpetersen:
ping @luisramos0, conflicts

luisramos0:
thanks @sigmundpetersen 
rebased to resolve conflict, a very simple one.

luisramos0:
Heads up testers, #4471 builds on top of this PR and both are in test ready.
Maybe easier to stage and test only #4471 and verify both PRs there.

Issue #4433, opened 2019-11-06
Fetch the Docker's ruby version from .ruby-version


sauloperez:
### What we should change and why (this is tech debt)

Every time we upgrade the Ruby version we need to update the Dockerfile as well and we tend to forget about it. And these upgrades are happening quite often recently.

We should read the current ruby version from [.ruby-version](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/.ruby-version) to populate the following env var

https://github.com/openfoodfoundation/openfoodnetwork/blob/52ab6c52bf5a5bfee4fcd4fed1ac3d2b3bd4bafb/Dockerfile#L9

### Context

This came up in https://github.com/openfoodfoundation/openfoodnetwork/pull/4301#issuecomment-550353957

### Impact and timeline

This removes at least one of the tasks involved in a Ruby upgrade.

PR #4463, opened 2019-11-14
Use .ruby-version when building Docker container

jonleighton:
The Ruby version was updated, but the Dockerfile wasn’t. This meant that
the Docker environment was broken. This change should prevent similar
breakage in the future by making .ruby-version the source of truth about
the correct version.


jonleighton:
Fixes https://github.com/openfoodfoundation/openfoodnetwork/issues/4433

sauloperez:
Tested and working!

PR #4469, opened 2019-11-18
Fix shipping_rate fetching in customer_totals_report

### LINKED ISSUE #4468: [Release Blocker] Incorrect Shipping Method in OC Customer Totals Report
## Description

Introduced in 2.6.3.

The Order Cycle Customer Totals report is not showing the correct shipping method. Instead is it showing the first available shipping method no matter which one was selected for the order.

@luisramos0 identified the bug here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4426#discussion_r346749067
Thanks Luis!

Personally as this is our most heavily used report I think this bug blocks the release. Open to others disagreeing.

## Expected Behavior

Order cycle Customer Totals report displays the shipping method that was selected on the order.

## Actual Behaviour

Order cycle Customer Totals report always displays the same shipping method, the first available one.

## Steps to Reproduce

1. Complete an order where you had a choice of multiple shipping methods and you did not choose the first one
2. View the order in the OC Customer Totals report.
3. The incorrect shipping method is displayed. Fees are correct

## Severity

## Your Environment

* Version used: 2.6.3
* Browser name and version: 

## Possible Fix

See the discussion following this comment:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4426#discussion_r346749067

Matt-Yorkley:
#### What? Why?

Closes #4468 

Fixes querying of shipping methods in customer and totals reports.

#### What should we test?

Orders where there are several shipping methods available to the producer and the one selected for the order was not the first one; should display correctly in customer and totals reports.

#### Release notes

Patched displaying of shipping method information in customer and totals reports

Changelog Category: Fixed

### DIFF
--- a/lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
+++ b/lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
@@ -198,8 +198,13 @@ module OpenFoodNetwork
       private
 
       def shipping_method(line_items)
-        line_items.first.order.shipments.first.
-          andand.shipping_rates.andand.first.andand.shipping_method
+        shipping_rates = line_items.first.order.shipments.first.
+          andand.shipping_rates
+
+        return unless shipping_rates
+
+        shipping_rate = shipping_rates.find(&:selected) || shipping_rates.first
+        shipping_rate.try(:shipping_method)
       end
     end
   end
--- a/spec/lib/open_food_network/orders_and_fulfillments_report/customer_totals_report_spec.rb
+++ b/spec/lib/open_food_network/orders_and_fulfillments_report/customer_totals_report_spec.rb
@@ -2,14 +2,7 @@ require "spec_helper"
 
 RSpec.describe OpenFoodNetwork::OrdersAndFulfillmentsReport::CustomerTotalsReport do
   let!(:distributor) { create(:distributor_enterprise) }
-
   let!(:customer) { create(:customer, enterprise: distributor) }
-
-  let!(:order) do
-    create(:completed_order_with_totals, line_items_count: 1, user: customer.user,
-                                         customer: customer, distributor: distributor)
-  end
-
   let(:current_user) { distributor.owner }
   let(:permissions) { OpenFoodNetwork::Permissions.new(current_user) }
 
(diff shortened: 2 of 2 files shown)


lin-d-hop:
Thanks for all the new specs @Matt-Yorkley :heart:

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
```diff
@@ -198,8 +198,13 @@ def line_item_includes
...
       def shipping_method(line_items)
-        line_items.first.order.shipments.first.
-          andand.shipping_rates.andand.first.andand.shipping_method
+        shipping_rates = line_items.first.order.shipments.first.
+          andand.shipping_rates
+
+        return unless shipping_rates
+
+        shipping_rate = shipping_rates.find(&:selected) || shipping_rates.first
+        shipping_rate.try(:shipping_method)
```
luisramos0: Is this repetition really needed? if it is it should be merged with already existing methods in OrderShipment (shipment and shipping_method) and Shipment (selected_shipping_rate)

luisramos0: I am only making this point here because this is copy/pasted core business logic. If it was some report specific logic I wouldn't worry.

Matt-Yorkley: I initially tried to adapt the related method directly, but there were various spec failures in Subs and order-related services. I'm not sure how long it would take to get a green build going in that direction; it's preferable but might not be  quick.

Matt-Yorkley: I didn't have much time and I felt the urgency around unblocking the release, but maybe the sensible option is to revert for now and hold back the reports improvements until the next release?

luisramos0: that's one option as well. I think we can go either way: revert to original state or keep this PR as is, as long as we improve it in the next release.

Do you want to share a PR with the change of the methods that break the specs, maybe I can help fix it?

luisramos0: the "loaded?" approach didn't work?

luisramos0:
So, I am approving this PR and bypassing 2nd approval so the release can happen today.
The release has been tested with this as a patch so the easiest way to go is to merge this into master and work on this as a new PR.

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
Matt-Yorkley: No, I this this again (in the `Spree::Shipment` decorator), which I think is the way to go:

```ruby
def selected_shipping_rate
      if shipping_rates.loaded?
        shipping_rates.find(&:selected)
      else
        shipping_rates.where(selected: true).first
     end
end
```

but there were multiple failures here (I think Job #3 is just a flaky one): https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-4469/builds/2

luisramos0: ok, I'll have a look at that!

Matt-Yorkley:
New issue here: #4474

Issue #4424, opened 2019-11-02
Add supplier_id index to spree_products table


Matt-Yorkley:
### What we should change and why (this is tech debt)

We're missing a `supplier_id` index on the `spree_products` table. This index is commonly used and will result in sequential scans. It's a foreign key, and has high cardinality.

### Context

Looking into potential useful indexes for performance.

Example traces:

https://app.datadoghq.com/apm/resource/rails-postgres/postgres.query/3e58832d71271611?end=1572733957676&env=production&index=apm-search&paused=false&start=1570141957676
https://app.datadoghq.com/apm/resource/rails-postgres/postgres.query/79a87739d97ae77f?end=1572729769909&env=production&index=apm-search&paused=false&start=1570137769909
https://app.datadoghq.com/apm/resource/rails-postgres/postgres.query/8478cbff71034945?end=1572729762684&env=production&index=apm-search&paused=false&start=1570137762684

luisramos0:
Moving this to tech debt backlog to remove the performance column in the pipe. This is probably the top priority in terms of performance issues right now. This would have a very good impact I am sure.

PR #4470, opened 2019-11-18
Improve readability of OC permissions...

luisramos0:
#### What? Why?

Relates to #4455
These methods should work on AR relations and not depend on variables like producer.
This is just a first clean up step.

#### What should we test?
Just code restructure, I don't think this needs testing.

#### Release notes
Changelog Category: Changed
Improved code readability.

### DIFF
--- a/lib/open_food_network/order_cycle_permissions.rb
+++ b/lib/open_food_network/order_cycle_permissions.rb
@@ -120,36 +120,38 @@ module OpenFoodNetwork
 
     # Find the variants that a user can POTENTIALLY see within incoming exchanges
     def visible_variants_for_incoming_exchanges_from(producer)
-      return Spree::Variant.where("1=0") unless @order_cycle
-
-      if user_manages_coordinator_or(producer)
-        # All variants belonging to the producer
-        Spree::Variant.joins(:product).where('spree_products.supplier_id = (?)', producer)
+      if @order_cycle &&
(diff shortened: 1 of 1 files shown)


lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -120,38 +120,40 @@ def visible_exchanges
...
+      if @order_cycle && user_manages_coordinator_or(producer)
+        all_variants_supplied_by(producer)
       else
-        Spree::Variant.where("1=0")
+        no_variants
       end
     end
 
+    def all_variants_supplied_by(producer)
+      Spree::Variant.joins(:product).where('spree_products.supplier_id = (?)', producer)
```
Matt-Yorkley: We _definitely_ need to add an index to the `spree_products` table on `supplier_id`...

Matt-Yorkley: https://github.com/openfoodfoundation/openfoodnetwork/issues/4424

lib/open_food_network/order_cycle_permissions.rb
```diff
@@ -120,38 +120,40 @@ def visible_exchanges
...
+    end
+
+    def no_variants
+      Spree::Variant.where("1=0")
```
mkllnk: I'm looking forward to Rails 4: `Spree::Variant.none`

PR #4471, opened 2019-11-18
Paginate Exchange Products API endpoint

### LINKED ISSUE #4476: Pagination in the OC create/edit pages when listing the products of an enterprise
## Description
As a hub manager
When managing the products in an Order Cycle
I want to be able to manage a large list of products from an enterprise without crashing the server
So that I can onboard shops with large catalogs

This is related to the OC POC #4379 and the new 3 steps UI #4422.

When I load the Order Cycle, not all products available are loaded immediately and when I open the list of products of an enterprise in that OC I see only a fraction of the list and I can load more or all products.
The total numbers of products for each enterprise must be correct "number of selected products" / "number of total products available".
The "select all" option must work correctly.

Optional but probably achievable within this issue: I can edit products in an OC with 5k products without crashing the server.

luisramos0:
#### What? Why?

Follow up from #4455 
Closes #4476 
This PR makes the OC pages able to handle exchanges with thousands of products by paginating the server queries to list exchange products.

More up to date demo:
http://g.recordit.co/HixXjlxPvk.gif

This PR opens up a few more major possibilities for improving performance, see isues #4480, #4481, #4482, #4487 and #4488

#### What should we test?
Verify the OC create and edit pages work correctly and specifically validate the list of products is loaded correctly when an exchange has more than 100 products (100 is the size of each page of products).

We should test the interaction between the load all products and the select all buttons in the exchange panel: if we click "select all" all checkboxes of the loaded products should be ticked.
One detail that may not be obvious to the user:
- in the incoming settings, when user clicks on "select all" we need to load all products from the server and then select all the checkboxes (this process is done automatically by the page, the user sees a spinner and then all products appear loaded in the page and selected)
- in the outgoing settings, when user clicks on "select all" we _dont_ need to load all products before we select all - this is much faster but not all products/checkboxes will be visible in the page (only the first 100 or more if user has manually clicked "load more" or "load all")

#### Release notes
Changelog Category: Changed
The pages to create and edit Order Cycles can now handle enterprises providing or selling thousands of products. This is accomplished by handling the list of products in small pages instead of handling the full list of products all the time.

#### Dependencies
This cannot be merged before #4455


app/controllers/api/exchange_products_controller.rb
```diff
@@ -52,23 +67,23 @@ def load_data_from_other_params
...
+        pagination: pagination_data(paginated_products)
       }.to_json
     end
 
-    def pagination_data(results)
+    def pagination_data(paginated_products)
       {
-        results: results.total_count,
-        pages: results.num_pages,
+        results: paginated_products.total_count,
```
mkllnk: Is this the same as the count of the new endpoint?

luisramos0: no, this is products count, the new endpoint is variants count.

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
```diff
@@ -9,19 +9,23 @@ angular.module('admin.orderCycles').controller 'AdminOrderCycleIncomingCtrl', ($
...
     OrderCycle.addSupplier $scope.new_supplier_id
+
+  # To select all variants we first need to load them all from the server
+  #
+  # This is only needed in Incoming exchanges as here we use supplied_products,
+  #   in Outgoing Exchanges the variants are loaded as part of the Exchange payload
+  $scope.selectAllVariants = (exchange, selected) ->
+    $scope.loadAllExchangeProducts(exchange).then ->
+      $scope.setExchangeVariants(exchange, $scope.suppliedVariants(exchange.enterprise_id), selected)
+      $scope.$apply()
```
Matt-Yorkley: Having to manually call `$scope.apply()` is a code smell in Angular...

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
```diff
@@ -9,19 +9,23 @@ angular.module('admin.orderCycles').controller 'AdminOrderCycleIncomingCtrl', ($
...
-    for product in products
-      numVariants += product.variants.length
-    numVariants
+    return enterprise.numVariants
 
   $scope.addSupplier = ($event) ->
     $event.preventDefault()
     OrderCycle.addSupplier $scope.new_supplier_id
+
+  # To select all variants we first need to load them all from the server
```
Matt-Yorkley: Is this the best way to handle selecting all variants? Can't we decouple the saving of the "selected" state from loading all the products?

So instead of fetching and loading all the available products in the exchange first, then sending another request to include all those products in the exchange (and save it), can't we just have an endpoint on the server-side that handles the task of "select all the available products in this exchange" and call that directly without doing the first step?

luisramos0: yes! that's exactly #4487

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
luisramos0: yes, agree. this is the only way I managed to refresh the page after the load process finishes. angular is not triggering a re-render.
anyway, this is supposed to disappear as we implement #4487

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
Matt-Yorkley: Ah sorry, I wasn't looking at all the issues at once.

luisramos0: no problem, it's good you described the issue in your comment without looking at it :-)

luisramos0:
This PR builds on top of #4455 so maybe easier to stage and test both PRs as one.

I am saying this because #4455 test case is: "This is a big change. We need to validate every aspect of the Order Cycle create and edit pages, both single enterprise and multiple enterprise."

RachL:
@luisramos0 here are some feedback from my round of testing:

1. On AU staging I have a weird count of products in outcoming exchange:

image

Why is the number selected higher?

Incoming seems correct:

image

I've ended up like this by unselecting products in incoming and editing several times and going back and forth between the two steps. Maybe an edge case as it only happens in this huge OC, the second OC I created on AU staging with only 10 products has a correct count...

Doing it in one go, meaning if I unselect 1 product in incoming, when I go to outgoing for the first time the tab is correct, but not the sentence XX of XX product loaded. Is it because in one case you count products and in the other variants?

image


2. On a single producer OC, I have some translation missing: is this only because we need to go through transifex?

image


Finally, here are the other tests I've carried and which seemed fine: 

- OC creation
- Editing OC: dates, name and removing / adding products and relation to the shopfront
- Duplicating OC

all this is working good :+1: I did not test subscription: anything related to schedule should be looked at here?

It is super cool to have the big spinner in the incoming :+1:

image

I would even add the big spinner when we load the steps (there is the sentence "loading" but a bit small :-) Anyway that's out of the scope here.

Also, in delivery train I forgot to speak about https://github.com/openfoodfoundation/openfoodnetwork/issues/4503 => IF you need to work again on this, is it too much to fix the CSS at the same time?

luisramos0:
thanks @RachL 
I looked at point 1 now.
Re numbers on the summary "x / y select " - I believe this is pre-existing problem and now I know why it happens. This is caused by products that changed suppliers. There are products selected in the OC that have a different supplier. Supplier caled "bb". If we change supplier again back to Dorv the numbers will become correct again. I think this can go live with this or maybe we should check if it is pre-existing. I'd open a separate issue for this.
re numbers of products - argh I have not thought about this problem! this will confuse the users... but the numbers are correct, above we see number of variants, below we see numbers of products :-(
what shall we do about this?
to change the loaded totals to number of variants instead of products, theres some extra calculation that needs to be done. it's not difficult but will take a few more hours of dev.

luisramos0:
Point 2 is a bug, those line should not show up in the simple form where we just load all the products for now.
I am fixing it now.

luisramos0:
I'd say we can skip code review as I just moved files around. Meanwhile reviews can comment/review if they think is required.
I'd say this is ready for test again: we need to recheck that the panels work on the multi exchange form and that now the simple form works and does not show the totals.

luisramos0:
I have redepoyed this last version to au staging (now running on ruby 2.3.7).

RachL:
Alright so the error is fix! Moving to ready to go and creating the other issues.

luisramos0:
great, thanks @RachL 
I have merged the first PR and changed the base of this one. I will wait for the master build and the build of this one to be green before I merge this one.

PR #4475, opened 2019-11-19
Add Bullet logging to test suite

Matt-Yorkley:
#### What? Why?

WIP: quick test of the Bullet gem with Semaphore.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -152,6 +152,7 @@ group :test, :development do
 end
 
 group :test do
+  gem 'bullet', '< 5.5.0' # For active-record 3.x
   gem 'simplecov', require: false
   gem 'webmock'
   # See spec/spec_helper.rb for instructions
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -169,6 +169,9 @@ GEM
     bugsnag (6.12.2)
       concurrent-ruby (~> 1.0)
     builder (3.0.4)
+    bullet (5.4.3)
+      activesupport (>= 3.0.0)
+      uniform_notifier (~> 1.10.0)
     byebug (9.0.6)
     cancan (1.6.10)
     capybara (2.18.0)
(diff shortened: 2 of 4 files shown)


Matt-Yorkley:
Ok... it looks like there's maybe a conflict with one of the Bullet dependencies and some of our code, so lots of specs are failing. On the other hand the N+1 reports are working, and there are LOTS of entries in the logs.

luisramos0:
😲 this is awesome, it even tells you what to do!
"Add to your finder: :includes => [:bill_address]"

luisramos0:
and then the performance nirvana:
Bullet.raise = true  # raise error if n+1 query found

luisramos0:
you can do "gem 'bullet', group: 'development'" to start with.

luisramos0:
there are 782 alerts! a lot are repeated, for example:
/home/runner/openfoodnetwork/app/serializers/api/admin/order_serializer.rb:12:in `full_name'

appears in 32 alerts.

I think we should follow this up! maybe we should create an epic specifically for this #exciting

Matt-Yorkley:
There's tons of errors related to variant option naming. There's a PR going through soon that changes it, I'll rebase this once it's merged and have another look...

Matt-Yorkley:
Also from looking at the github issues for this gem, it looks like there could be some false-positive and false-negative problems. I'll play with it again next week and see how we can extract some value from it.

Matt-Yorkley:
> /home/runner/openfoodnetwork/app/serializers/api/admin/order_serializer.rb:12:in `full_name' appears in 32 alerts

That should be fixed in the open PR for variant naming. :+1:

Matt-Yorkley:
The build is green now (apart from a single flaky spec), so the question is: how do we want to use the output from this gem?

It seems like a valuable resource. We could do a quick blast through the list and try to add eager-loading in all the suggested places? Whether we want to merge this PR and permanently include Bullet reports in our CI is a different question.

luisramos0:
ah, I dont think we should forget about this.
I will re-open this PR temporarily now, capture the output of bullet in the CI build and create an tech debt issue for this based on the current output.

luisramos0:
ok, I have created https://github.com/openfoodfoundation/openfoodnetwork/issues/4673

PR #4507, opened 2019-11-27
Bring spree_backend reports controller to OFN

luisramos0:
#### What? Why?

Part of #4050

Here we bring spree/admin/reports_controller to OFN and merge with its decorator.

The remaining rubocop issues are length and complexity related and were added to the manual_todo.

#### What should we test?
Not much is changed here so I'd say we should do a sanity check on every report. Just to make sure navigation and execution works.

#### Release notes
Changelog Category: Changed
Moved reports controller code to OFN to remove dependency from Spree.


RachL:
Looking good :ok_hand: 

Tested redenring and filtering.

luisramos0:
ouch, I just realized I completely forgot about the sales report. This PR has removed the sales report!
From the slack conversation I had the feeling most people would want to keep it... @RachL can you advise here pls? :-( shall I create a PR to re-add the report?
I am sorry about this!

RachL:
@luisramos0 I will have a meeting next week with Lynne to set up zapier maybe I won't be needing the report anymore. But AU is relying only on reports for now I believe, and maybe they use this one to cross check everything? Something to check with @kirstenalarsen @mkllnk

mkllnk:
We have our own hack to get that data. I think that @kirstenalarsen was maybe using it to double-check? I don't think we actually need it in Australia.

kirstenalarsen:
HI @luisramos0 @mkllnk - I missed this sorry, PLEASE can we have it back, will explain why in slack

PR #4508, opened 2019-11-27
Bring spree_backend products and variants controllers to OFN 

luisramos0:
#### What? Why?

Related to #4050

Bring products and variants controllers from spree and merge with ofn's decorators.

Most code is already in OFN for both products and variants.

Some rubocop isssues fixed, rubocop manual todo updated to cover remaining issues.

#### What should we test?
We only some sanity checks around this functionality:

In the products submenu Variants:
- Create variant
- Delete a variant

In the edit order page:
- Search for a variant to add to an order

In terms of products:
- Create and edit a product
- Bulk edit products: edit product details, create variant, edit variant details and clone product

#### Release notes
Changelog Category: Changed
Broght some code that manages products and variants pages from spree_backend to make OFN independent of spree.


luisramos0:
I realized the products routes were outside the admin namespace. I moved them into the admin namespace now. I also rebased to resolved the conflict.

app/controllers/spree/admin/products_controller.rb
```diff
@@ -1,30 +1,86 @@
...
+      } }      
+
+      def new
+        @object.shipping_category = DefaultShippingCategory.find_or_create
+        super
+      end
+
+      def create
+        delete_stock_params_and_set_after do
+          super
```
mkllnk: Just to clarify. This `super` was in the override before. And it called `RessourceController.create`, not the overridden `ProductsController.create`, right? In that case, this is all good. :-) This confusion is a good reason to get rid of these overrides.

luisramos0: yes, super is not the class overridden by class_eval but rather the class overridden by inheritance.

filipefurtad0:
Hi @luisramos0,

In the products submenu Variants:
- Created variant and 
- Deleted a variant
These appeared and disappeader in the shopfront, as they should.

In the edit order page:
Search for a variant to add to an order -> Checked, they appear and can be inserted in the shopping cart.

In terms of products:
Bulk edit products: edit product details, create variant, edit variant details and clone product -> All good. All these changes work fine from the Product Menu (bulk) and changes appear in the shopfront.

I think all is fine here! :-)
Moving to ready to go.

luisramos0:
thanks @filipefurtad0 
I rebased to fix a conflict on a rubocop file (cannot impact functionality in any way).

PR #4509, opened 2019-11-27
Bring spree_backend payments controller to OFN

luisramos0:
#### What? Why?

Related to #4050

We bring payments_controller from spree_Backend and merge the decorator in OFN.

We fix some rubocop issues but leave abc size and method length on create and fire for later. These are reflected in the new todo_manual file.

#### What should we test?
We need to verify the page to capture payments in the order.
Create payment, capture payment, cancel payment.
We should test these things above with several order states as well as several payment methods.

There's this edge case where we delete the payment method and the list of payments of a past order with that payment method still shows the name of that deleted payment method correctly. I wonder if this is working in the current version. I havent tested this.

#### Release notes
Changelog Category: Changed
Brought the backoffice page that captures payments from spree to make OFN independent of Spree.


Matt-Yorkley:
Needs a rebase @luisramos0

sigmundpetersen:
Also needs a rebase

luisramos0:
:+1: done

RachL:
@luisramos0 when I click on a payment method of an order, I get an error 500 and I don't see this in production. Could it be linked to this PR?

Example link: https://staging.katuma.org/admin/orders/R064445101/payments/174

image


image

luisramos0:
It was a bug in my migration of the code, I didn't see there was a show action. This link on the "payment method" column goes to a page that shows the payment...
It's now working. I just staged it to ES.

luisramos0:
ready for testing in katuma.

RachL:
Great I think this is ready to go!

I've tested creating, capturing and cancelling on all 3 payments methods.

Also

>There's this edge case where we delete the payment method and the list of payments of a past order with that payment method still shows the name of that deleted payment method correctly. I wonder if this is working in the current version. I havent tested this.

This is working as well :tada:

PR #4510, opened 2019-11-27
Bring spree_backend resource controller to OFN

luisramos0:
#### What? Why?

Related to #4050

Here we move the original spree backend resource controller to OFN and add two small changes from the OFN decorator.

I have fixed the easy rubocop issues, the ones left behind are complexity or class/method length related.

#### What should we test?
This controller is used all over the backoffice but nothing was changed so a green build already indicates most of the behavior is OK.

I think we can test the following actions:
- bulk update different variants in the inventory page
- create enterprise
- create subscription and add line items to it

#### Release notes
Changelog Category: Changed
Imported the basic resource controller that supports all controllers of the backoffice from spree_backend to make OFN more independent of Spree.


luisramos0:
Finishing the spree_backend epic, I realized something was missing in this PR: actioncallbacks. I added 2 new commits to this PR. Moved back to code review for validation of the last 2 commits.

RachL:
-     bulk update different variants in the inventory page: All good
-     create enterprise: All good
-     create subscription and add line items to it: All good

Moving to ready to go!

luisramos0:
:tada: quite an epic one this PR!!! resource_controller is on our side now.

PR #4513, opened 2019-11-27
Bring spree_backend orders controller to OFN 

luisramos0:
#### What? Why?

Relates to #4050 

Here we bring orders and customer details controllers from spree and merge them with the existing decorators.

Fixed some rubop issues and updated rubocop manual todo with remaining rubocop issues like abc size and method length.

#### What should we test?
We need to test the full admin orders workflow. Mainly the workflow, not the details.
Listing orders - here particularly the actions of shipping and capturing payments of an order from the orders list page.
Create/edit order - make sure we can add line items and add customer details as usual.
We should also validate that the actions work correctly like printing invoice and cancelling/resuming the order. 
We should verify, on the edit order page, the order state and total value as well as the shipment and payment state appear correctly on the right side menu: the shipment/payment state only appear if the order is complete.
On the customer details the address form fields were moved from spree to ofn here so we should check it renders correctly all the fields.

#### Release notes
Changelog Category: Changed
Brought from spree_backend code related to editing an order in the backoffice so we can make OFN independent of Spree.


app/views/spree/admin/shared/_address_form.html.erb
```diff
@@ -0,0 +1,72 @@
+<% if use_billing %>
```
luisramos0: I created #4584 to convert this to haml at a later stage.

luisramos0:
rebase to resolve simple conflict.

RachL:
>We need to test the full admin orders workflow. Mainly the workflow, not the details.

Workflow doing ok as usual.

>Listing orders - here particularly the actions of shipping and capturing payments of an order from the orders list page.

All good

> Create/edit order - make sure we can add line items and add customer details as usual.

All good

> We should also validate that the actions work correctly like printing invoice and cancelling/resuming the order.

All good

> We should verify, on the edit order page, the order state and total value as well as the shipment and payment state appear correctly on the right side menu: the shipment/payment state only appear if the order is complete.

Yes!

>On the customer details the address form fields were moved from spree to ofn here so we should check it renders correctly all the fields.

This is ready to go!

luisramos0:
rebased to resolve minor conflicts

Issue #3912, opened 2019-06-10
Unable to create new product (snail) with Value field empty


luisramos0:
## Description
User cant create product.

![Screenshot 2019-06-10 at 13 11 51](https://user-images.githubusercontent.com/1640378/59194625-8af70f80-8b81-11e9-8f18-aab2b98b39a6.png)

I am still talking to the user to confirm this is her problem but I did find a bug.

## Steps to Reproduce
When creating a product with unit type items, user fills in all required fields except the unit value field which is mandatory.

## Actual Behaviour
After submitting the form, the user should see the same page with a standard form error saying "unit value field cant be blank".

![Screenshot 2019-06-10 at 13 11 58](https://user-images.githubusercontent.com/1640378/59194626-8af70f80-8b81-11e9-8ad5-7fb0b13f5f7c.png)

## Expected Behavior
After submitting the form, user sees snail page.

## Context
First real bug coming from Portuguese users.

## Severity
bug-s3
the workaround is to call support that will call developers that will look at the logs and will tell support "unit value field is mandatory" that will tell the user "unit value field is mandatory" :-D

## Your Environment
* Version used: Katuma OFN v2.0.2
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): osx 10.14

## Possible Fix
the bug is in this line:
https://github.com/openfoodfoundation/openfoodnetwork/blob/568e3003ba43c713ac4fb5a57fddfbca086f7b98/app/controllers/spree/admin/products_controller_decorator.rb#L142

for some reason product.valid? is returning true but product.errors is not empty:
![Screenshot 2019-06-10 at 13 05 24](https://user-images.githubusercontent.com/1640378/59194929-8717bd00-8b82-11e9-8321-2d1499ac7195.png)

Probably related to the fact that this field unit_description is a delegate in the product model:
`delegate_belongs_to :master, :unit_value, :unit_description`

The fix is to either use product.errors instead of product.valid? here or change something in the model to get product.errors and product.valid? consistent again.

luisramos0:
Just replicated this one in katuma live, v2.5.0

mkllnk:
This issue has been linked to an error in Bugsnag
[**RuntimeError** in **spree/admin/products#create**](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5db4208c2ab8f4001989c256?event_id=5e816edf00595296c5f60000&i=gh&m=li)

PR #4514, opened 2019-11-27
Bring spree_backend overview controller to OFN

luisramos0:
#### What? Why?

Related to #4050

Here we basically move the overview decorator to the original controller from spree_backend because the original controller was basically empty.

#### What should we test?
We need to validate that the main admin dashboard works as usual. Probably better to validate it for the different scenarios:
- redirect to enterprises if first access and enterprise profile (profile, producer, hub) was not chosen yet
- single enterprise dashboard if user manages only 1 enterprise
- multiple enterprises dashboard if user manages more than 1 enterprise

#### Release notes
Changelog Category: Changed
Imported the controller related to the dashboard from spree_backend to make OFN more independent of Spree.


filipefurtad0:
Hi @luisramos0 :-)

Checked the test cases, all good.

Not related to this PR (tested before and after staging):
Noticed one gets an Snail/Error 500, when adding a new product (link below) if not specifying Unit Value:
https://openfoodnetwork.org.au/admin/products/new

There are several S3s at All The Things, but didn't find this one. Should I open a new S3?

In any case, I'll move this to Ready To Go :-)

luisramos0:
@filipefurtad0 you posted a link to production, did you test in staging?

filipefurtad0:
I see now I didn't. In AU-staging the confirmation email sends me back to production, and I didn't notice it - sorry for that.
Moving back to test ready.

filipefurtad0:
Double checked the test cases, same result. I think everything is fine. So, moving back to Ready to Go.

I still get Error 500, now on the link below, and if one don't insert a value for unit size.
https://staging.openfoodnetwork.org.au/admin/products/new

I haven't tested much in AU-staging, and I realize now that this is a particular aspect of the email confirmation here. Will keep my eye open for this.

luisramos0:
ah, tricky emails!

ok, thanks Filipe.

I think you are seeing this issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/3912

filipefurtad0:
Yes! That's it Luis - issue: #3912
I'll be watching out for those tricky emails.
Thanks again.

RachL:
@filipefurtad0 @luisramos0 that confirmation link can be fixed by super admin. But I don't have the login. Do you guys have it for staging AU?

luisramos0:
ah, is it the site url in general config? I have just fixed it in staging AU :-)
thanks @RachL

RachL:
@luisramos0 yes it is! But I don't know if this could be changed "permanently" in the way we build our test data. often, we end up with spree demo URL after we put fresh seed data for example. I guess in this case we've put some production data on AU and ended up with the production URL in the config?

filipefurtad0:
Thank you to both! @luisramos0 @RachL This is a great discussion, especially since I am not so clear on how to test with seed- / test- / production-data. Should we move to Slack? Would definitivelly be grateful for some insigts. There recently was a discussion I had some difficulty following:
https://openfoodnetwork.slack.com/archives/C02TZ6X00/p1575446899162200

PR #4515, opened 2019-11-27
Bring spree_backend adjustments controller to OFN

luisramos0:
#### What? Why?

Related to #4050

Here we delete the adjustments controller decorator and merge it with the original adjustments controller from spree_backend.

#### What should we test?
The page to manage the adjustments in an order should work as is. Add adjustments, edit adjustments and delete adjustments.
There's some code to guess the tax rate of the adjustment. The tax rate is not stored in the adjustment, only the value of the tax after computation. So there's some code to "guess" what tax rate the adjustment belongs to...

#### Release notes
Changelog Category: Changed
Imported routes and the controller related to order adjustments from spree_backend to make OFN more independent of Spree


app/controllers/spree/admin/adjustments_controller.rb
```diff
@@ -1,12 +1,23 @@
 module Spree
   module Admin
-    AdjustmentsController.class_eval do
+    class AdjustmentsController < ResourceController
+      belongs_to 'spree/order', find_by: :number
+      destroy.after :reload_order
```
Matt-Yorkley: What's the reason for the `destroy.after` syntax here instead of `after_destroy`?

luisramos0: interesting, I think it's the resource_controller callbacks extension... it's spree code and it's still in master:
https://github.com/spree/spree/blob/597cd1578705b49de44525efefdc7805a40cfa6b/backend/app/controllers/spree/admin/adjustments_controller.rb#L6

https://github.com/spree/spree/blob/master/backend/lib/spree/backend/action_callbacks.rb

As I understand it, this is used in resource_controller to setup callbacks, for example, after destroy, that are executed after the object is updated but before rendering:
https://github.com/spree/spree/blob/597cd1578705b49de44525efefdc7805a40cfa6b/backend/app/controllers/spree/admin/resource_controller.rb#L79

luisramos0: basically, controller level callbacks. IMO Rails at it's worse where callback logic "learned" in models is extrapolated to controllers.

luisramos0:
Moving to test ready as I think we need to keep the controllers callbacks for now.

app/controllers/spree/admin/adjustments_controller.rb
Matt-Yorkley: Ugh. Yeah, it's not pretty...

RachL:
CRUD adjustments is working good, and I tested 2 tax rates and the result was correct. It is interesting to see how this is working vs the fees system.

Issue #2891, opened 2018-10-19
Upgrade the Ruby version and Ruby gems


jhsu802701:
As of 2018-10-18, this app has Ruby 2.1.5 (released 2014-11-13), Rails 3.2.22 (released June 16, 2015), nokogiri 1.6.8.1 (released October 03, 2016), and pg 0.21.0 (released June 12, 2017).

I will work on updating this app.  I have experience upgrading the Adopt-A-Tree app (https://github.com/OpenTwinCities/adopt-a-tree).

luisramos0:
There are things that could be already adapted in our current versions that will work with future versions. Is this not correct?
"Write OFN in current ruby 2.1 in a way that the migration to ruby 2.3 will be a breeze."
"Write OFN in current rails 3.2 in a way that the migration to rails 4 will be a breeze."
Is this not a valid project we could open to the community?
I am making this question because I think this is a generic and interesting topic for the community to help out (unlike the complex spree upgrade!).
This is a question for other core devs.

jhsu802701:
How can I join the Slack channel?

sauloperez:
> I am making this question because I think this is a generic and interesting topic for the community to help out (unlike the complex spree upgrade!).

That's an interesting idea. Theoretically, this would remove a great barrier of entry to new Ruby contributors. There's no need to be fully aware of OFN's and Spree's data model but of Rails/Ruby changes and I assume that's a potentially bigger set of devs.

To my understadning

> Write OFN in current ruby 2.1 in a way that the migration to ruby 2.3 will be a breeze.

It shouldn't require any action as Ruby upgrades are backward compatible although I don't know off the top of my head what 2.3 brings in.

> Write OFN in current rails 3.2 in a way that the migration to rails 4 will be a breeze.

I've been here already 🙈 . This will indeed require refactoring some calls to ActiveRecord methods and possibly other changes to the Gemfile. Someone can start with the former if we prepare detailed and scoped issues. Upgrade guide: https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0

mkllnk:
Quick scan into the guide.

* Update from put to patch. We have some uses:
```
> grep put config/routes/*
config/routes/admin.rb:        put :register
config/routes/admin.rb:      put :bulk_update, on: :collection
config/routes/admin.rb:      put :cancel, on: :member
[...]

PR #4301, opened 2019-09-23
Use Ruby 2.2.10

Matt-Yorkley:
#### What? Why?

Part 1 of #2891 

Bumps Ruby version to `2.2.10`, giving many (over 1 thousand?) improvements:

- 2.2.0 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_0/ChangeLog
- 2.2.1 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_1/ChangeLog
- 2.2.2 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_2/ChangeLog
- 2.2.3 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_3/ChangeLog
- 2.2.4 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_4/ChangeLog
- 2.2.5 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_5/ChangeLog
- 2.2.6 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_6/ChangeLog
- 2.2.7 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_7/ChangeLog
- 2.2.8 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_8/ChangeLog
- 2.2.9 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_9/ChangeLog
- 2.2.10 https://svn.ruby-lang.org/repos/ruby/tags/v2_2_10/ChangeLog
#### What should we test?

- [x] Provision a staging server with: https://github.com/openfoodfoundation/ofn-install/pull/530
- [x] Stage this PR
- [x] Check everything looks OK
- [x] Try running the CI build with `2.2.10`

#### Release notes

Bumped Ruby version from `2.1.9` to `2.2.10`

Changelog Category: Changed

PR #4250, opened 2019-09-09
Bump Ruby version from 2.1.5 to 2.1.9

Matt-Yorkley:
#### What? Why?

Minor Ruby version bump from 2.1.5 to 2.1.9. We gain a _vast_ number of updates and fixes:

- 2.1.6 Changelog: https://svn.ruby-lang.org/repos/ruby/tags/v2_1_6/ChangeLog
- 2.1.7 Changelog: https://svn.ruby-lang.org/repos/ruby/tags/v2_1_7/ChangeLog
- 2.1.8 Changelog: https://svn.ruby-lang.org/repos/ruby/tags/v2_1_8/ChangeLog
- 2.1.9 Changelog: https://svn.ruby-lang.org/repos/ruby/tags/v2_1_9/ChangeLog

#### What should we test?

Needs some dev input:

- Provision a staging server with https://github.com/openfoodfoundation/ofn-install/pull/514
- Deploy this PR
- Test everything looks ok

#### Release notes

Bumped Ruby from 2.1.5 to 2.1.9

Changelog Category: Changed

PR #4518, opened 2019-11-28
Bump Ruby to 2.3.7

Matt-Yorkley:
#### What? Why?

Related to #2891 

Bumps Ruby to `2.3.7` :tada: I had to drop down from `2.3.8` because Semaphore apparently doesn't support it. 

Full Ruby changelogs are here:

- 2.3.0 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_0/ChangeLog
- 2.3.1 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_1/ChangeLog
- 2.3.2 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_2/ChangeLog
- 2.3.3 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_3/ChangeLog
- 2.3.4 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_4/ChangeLog
- 2.3.5 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_5/ChangeLog
- 2.3.6 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_6/ChangeLog
- 2.3.7 https://svn.ruby-lang.org/repos/ruby/tags/v2_3_7/ChangeLog

#### What should we test?

General release type test of core functionality.

#### Release notes

Bumped Ruby version to `2.3.7`

Changelog Category: Changed


Matt-Yorkley:
It's green!

luisramos0:
"Rails 3.2 reached its end of life-cycle with the release of Rails 5.0 on June 30th, 2016"
Ruby 2.3.0  was released on 2015-12-25 
Ruby 2.3.7 was released on 2018-03-28

"Rails versions 3.2 to 4.2 were built to work with any version of Ruby from 1.9. 3 to 2.2, allowing flexibility"

I assume this will be the last ruby upgrade before the rails upgrade. Is that correct?
I think ruby 2.4 will only be ok with Rails 4.2 (not rails 4). [link](https://stackoverflow.com/questions/9087116/which-ruby-on-rails-is-compatible-with-which-ruby-version)
I hope this ruby 2.3 will not create trouble when upgrading to rails 4.0.

Matt-Yorkley:
Yeah I think this will be the last one.

mkllnk:
I'm concerned about Rail 4.0 being untested with Ruby 2.3. I haven't been able to run the Rails specs yet to confirm it it works. I would like test a couple more things. Otherwise I would say let's just try it. Worst case would be that we have to downgrade Ruby during the next upgrade. :-(

Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby "2.2.10"
+ruby "2.3.7"
```
mkllnk: I'm wondering if we should remove this line altogether. I believe that the purpose is to make Bundler fail fast if the used Ruby version isn't right. [Docs](https://bundler.io/v1.7/gemfile_ruby.html). But our application does work with several Ruby versions at the moment (2.1, 2.2, 2.3). So I would skip this requirement and see the `.ruby-version` file as source of truth to select the best version.

sigmundpetersen:
I guess this is dependent on https://github.com/openfoodfoundation/ofn-install/pull/549 before merge? :blush:

Gemfile
sauloperez: yes, I think we'll get to it as soon as it breaks our development with the two versions.

luisramos0:
I provisioned staging-fr with latest ofn-install (with 549) and now I started the deploy of this PR:
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/servers/fr-staging/deploys/137

luisramos0:
looking good, ready for testing.

RachL:
Ok so I've taken the same testing approach than #4250 and #4301, which means I did a sanity check:

1- Login
2- Create a product and it appears correctly in the shop
3- Checkout (paypal & cash)
4- See the order on the order view (+ capture and ship and edit)

The rest will be tested during release testing like previsouly. So I'm moving this to ready to go :tada:

sigmundpetersen:
@luisramos0 should the open PR's in the pipe be rebased after this merge?

sauloperez:
No need. Now the servers have the current and upcoming Ruby version installed :tada: we learned the lesson.

PR #4524, opened 2019-11-29
Packing reports performance

Matt-Yorkley:
#### What? Why?

Related to #4521 

Adds missing eager-loading in packing reports. Reduces total queries by ~76% and drastically reduces page load times (should be less than half the previous time).

#### What should we test?

Packing reports are working correctly.

#### Release notes

Improved performance in pacing reports.

Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/reports_controller_decorator.rb
+++ b/app/controllers/spree/admin/reports_controller_decorator.rb
@@ -222,7 +222,8 @@ Spree::Admin::ReportsController.class_eval do
   end
 
   def suppliers_of_products_distributed_by(distributors)
-    distributors.map { |d| Spree::Product.in_distributor(d) }.flatten.map(&:supplier).uniq
+    distributors.map { |d| Spree::Product.in_distributor(d).includes(:supplier).all }.
+      flatten.map(&:supplier).uniq
   end
 
   # Load order cycles the current user has access to
--- a/lib/open_food_network/packing_report.rb
+++ b/lib/open_food_network/packing_report.rb
@@ -95,7 +95,7 @@ module OpenFoodNetwork
     def columns
       if is_by_customer?
         [proc { |line_items| line_items.first.order.distributor.name },
-         proc { |line_items| customer_code(line_items.first.order.email) },
+         proc { |line_items| customer_code(line_items.first.order) },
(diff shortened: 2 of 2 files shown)


lib/open_food_network/packing_report.rb
```diff
@@ -146,8 +147,8 @@ def is_by_customer?
       params[:report_type] == "pack_by_customer"
     end
 
-    def customer_code(email)
-      customer = Customer.where(email: email).first
+    def customer_code(order)
+      customer = order.customer
```
mkllnk: This is fixing a bug as well. The report could have fetched the customer code given by another enterprise.

app/controllers/spree/admin/reports_controller_decorator.rb
```diff
@@ -222,7 +222,8 @@ def my_suppliers
   end
 
   def suppliers_of_products_distributed_by(distributors)
-    distributors.map { |d| Spree::Product.in_distributor(d) }.flatten.map(&:supplier).uniq
+    distributors.map { |d| Spree::Product.in_distributor(d).includes(:supplier).all }.
```
mkllnk: Definitely options for improvement here, but I agree it's out of scope. We probably want something like `Enterprise.is_primary_producer.supplying_variant_in(...)`.

Matt-Yorkley:
I'm taking another look using live data from UK prod. It's pretty crazy..!

Matt-Yorkley:
OK its basically crashing my laptop. I retested with production data and this reduced reports loading for one day from ~32 seconds to ~30 seconds. It also shows the huge difference we have in production, and the extent of other problems related to permissions when the amount of data is exponentially larger.

Matt-Yorkley:
Retesting with production data; ~38% reduction in queries (from 1465 to 903), total page load reduced from 32.8 seconds to 30.2 seconds:

BEFORE:
Screenshot from 2019-11-30 11-51-40
AFTER:
Screenshot from 2019-11-30 12-15-59

luisramos0:
great!
25th takes around that time 30secs.
the 25th and the 26th together take around 8mins for me.

lin-d-hop:
This looks good. Tested at producer and food hub - Pack by Customer and Pack by Supplier.
Good to go.

PR #4526, opened 2019-11-29
3149 Fix tooltips in orders list

### LINKED ISSUE #3149: Customers notes are truncating across different orders
## Description

Customers placing notes with their orders seem to sporadically been appearing in other customers orders. 

## Expected Behavior

Notes made by a customer should only show on their own orders

## Actual Behavior

There have been two peculairties here which we are tryig to pin down

1. Oliver who uses firefox and safari on a mac, see customers notes replicated on the incorrect orders.
I however on windows, chrome and firefox see the comments matching up to the correct orders. 

2. When you change the page, order notes from the first page are showing up on the second page - as seen in the video. This affects all users regardless of their browser/OS

If it's better to split these I can do so, but as we get different results as we replicate the issues I've gone for putting our observations down.

## Steps to Reproduce

Difference in views
Orders R136242427 and R074566242

Page truncation
1. Go to orders and filter by stroudco
2. Look at the order notes for R383184313  on page 1, it states they are away
3. Look at the order notes for the second customer on page 2 R864052464 - it copies the one outlined above

http://recordit.co/nzMXru6bEr

## Animated Gif/Screenshot

First issue where we see different things:

What oliver sees -  firefox, safari, mac
r136242427 order - olivers
r074566242 - orders - olivers

What I see - chrome, firefox, windows
r074566242 - orders
r136242427 order

Second issue where there is a truncation
http://recordit.co/nzMXru6bEr

## Context

Difficult to know which information is correct on first glance and this could result in orders being dealt with incorrectly.

## Your Environment

* Version used:
* Browser name and version:

kristinalim:
#### What? Why?

- Closes #3149 

This does two things:

* Set max width and enable text wrapping for the tooltip content
* Fix how the tooltips are associated with orders in the orders table

The 2nd is accomplished by fixing Angular tracking of rows in the orders list. `ng-repeat` was tracking items by table row index. As far as `ng-repeat` was concerned, rows with the same table row index in any page (pagination) were for the same item, so it wouldn't bother relinking the `ofn-with-tip` directive.

Example scenario: Order A in row 1 of the first page and Order B in row 1 of the second page have special instructions. Navigate to the second page, and notice that the tooltip for Order B uses the special instructions for Order A.

#### What should we test?

In the Orders page, find a table row number for which there are multiple pages (pagination) where the order has special instructions (e.g. row 2 in both pages 1 and 3 have special instructions). The tooltip for the order in the later page should not use the special instructions of the order in the earlier page.

Also test wrapping of text for long special instructions.

Smoke test of these in the "Orders" page is also needed:

* The link to edit order still works
* Selection of orders for "Print Invoices" works

This change affects all similar looking tooltips, e.g. "Edit" and "Capture" in the Orders page. But I can't think of any special tooltips that need to be tested.

#### Release notes

- Fix text wrapping and using of the correct order for the special instructions tooltip in the Orders page.

Changelog Category: Fixed

### DIFF
--- a/app/assets/stylesheets/admin/components/tooltip.css.scss
+++ b/app/assets/stylesheets/admin/components/tooltip.css.scss
@@ -1,0 +1,4 @@
+#powerTip {
+  max-width: 240px;
+  white-space: normal;
+}
--- a/app/views/spree/admin/orders/index.html.haml
+++ b/app/views/spree/admin/orders/index.html.haml
@@ -47,7 +47,7 @@
 
       %th.actions
   %tbody
-    %tr{ng: {repeat: 'order in orders track by $index', class: {even: "'even'", odd: "'odd'"}}, 'ng-class' => "'state-{{order.state}}'"}
+    %tr{ng: {repeat: 'order in orders track by order.id', class: {even: "'even'", odd: "'odd'"}}, 'ng-class' => "'state-{{order.state}}'"}
       %td.align-center
         %input{type: 'checkbox', 'ng-model' => 'checkboxes[order.id]', 'ng-change' => 'toggleSelection(order.id)'}
       %td.align-center


RachL:
> In the Orders page, find a table row number for which there are multiple pages (pagination) where the order has special instructions (e.g. row 2 in both pages 1 and 3 have special instructions). The tooltip for the order in the later page should not use the special instructions of the order in the earlier page.

That was long to set up :D but all good!

> Also test wrapping of text for long special instructions.

image


> This change affects all similar looking tooltips, e.g. "Edit" and "Capture" in the Orders page. But I can't think of any special tooltips that need to be tested.

All good!

Tested in latest Firefox, Edge and Safari.

Ready to go!

luisramos0:
nice!

PR #4527, opened 2019-12-02
Improve Subscriptions logging

### LINKED ISSUE #4462: Add Logging to Subscriptions
## Description

As a developer I want to be able to see logs for Subscriptions.

[From @luisramos0]
There's a design flaw in the subs module: absence of logging.
This subs module is full of automagic stuff like generating proxy orders and then real orders, and then completing those orders and sending summaries to people...

Again and again as the code is used things will happen and users will want to know what has happened? why this or why that. Sometimes the user will fail, miss the point, sometimes we will fail, bugs.

To support such a process we need logging that lets us monitor what automatic processes were executed, and what/why did they do what they did.

More context for this issue is given in the comments in #4396

Suggested implementation:
Enable existing logs on delayedjob, and add a custom logging capability to the app and issue some specific log messages from the code.

1. Dev Test?
2. 
3. 
4.

This is an improvement or tech debt... not really a bug.
I am adding an S2 label as this is replacing #4396.

If you disagree that this is an S2 please say so.... but I really hope you don't. UK keep getting new subs users and every one that comes on makes me want to cry due to the black hole of problems we've experienced.

luisramos0:
#### What? Why?

Progresses #4462 but we will need to check if linking schedules to OCs is fully tracked now.

Add paper_trail to OCs and Schedules and add extra logging to Subscriptions background processes.

Here we do a few different things (I keep it in one PR because this is all related and it makes the review and test process easier, I think):

A - Add paper_trail to OCs and Schedules: this enables us to see what changes were done to these models and when they were created/updated/deleted. We also add a custom_data field to paper_trail versions table so we can have extra info for whatever we need, in this case, we are adding: for Schedules, the OCs they are associated with, and for OCs, the schedules they are associated with. (I thought this is a lighter solution that introducing gem paper_trail-association_tracking).
This addresses the problem in the S2 that led to #4462: when did the user linked the schedule with the OC? I don think it will cover 100% of the cases: I think the user will still be able to associated a schedule to an OC without any trace on the DB. That's why I dont think this issue will close #4462

B - Add different log statements in the proxy orders sync process and in the subs placement and confirmation background jobs so that we know what subs and orders were processed in each run. These statements are not going to solve all our debugging needs but it's a start, these basic ones are specially useful when the process fails to run, we will be able to check the logs and see if it didn't run or if it did run without producing the expected output.

#### What should we test?
This PR can only be tested by a dev and doesnt require a testers to verify anything.

A - We should edit the OC and edit a schedule and verify the table versions contain the previous data on those models.

B - We should setup a subscription or two, edit them and verify that the sync process worked and there are logs mentioning what happened. After that, when the OC opens, we need to verify the placement job worked and there are logs mentioning what happened, including what emails were sent. After that, when the OC closes, we need to verify the confirmation job worked and there are logs mentioning what happened, including what emails were sent.

#### Release notes
Changelog Category: Added
Subscriptions module now has much better logging and it will be easier from now on to debug and explain what happened in specific cases.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1202,6 +1202,7 @@ ActiveRecord::Schema.define(:version => 20191023172424) do
     t.string   "whodunnit"
     t.text     "object"
     t.datetime "created_at"
+    t.string   "custom_data"
   end
 
   add_index "versions", ["item_type", "item_id"], :name => "index_versions_on_item_type_and_item_id"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20191023172424) do
+ActiveRecord::Schema.define(:version => 20191202165700) do
 
   create_table "adjustment_metadata", :force => true do |t|
     t.integer "adjustment_id"
@@ -1196,9 +1196,9 @@ ActiveRecord::Schema.define(:version => 20191023172424) do
   add_index "variant_overrides", ["variant_id", "hub_id"], :name => "index_variant_overrides_on_variant_id_and_hub_id"
 
   create_table "versions", :force => true do |t|
-    t.string   "item_type",  :null => false
-    t.integer  "item_id",    :null => false
-    t.string   "event",      :null => false
+    t.string   "item_type",   :null => false
+    t.integer  "item_id",     :null => false
+    t.string   "event",       :null => false
     t.string   "whodunnit"
     t.text     "object"
     t.datetime "created_at"
--- a/lib/open_food_network/proxy_order_syncer.rb
+++ b/lib/open_food_network/proxy_order_syncer.rb
@@ -17,27 +17,34 @@ module OpenFoodNetwork
     end
 
     def sync!
-      return sync_all! if @subscriptions
+      return sync_subscriptions! if @subscriptions
+
       return initialise_proxy_orders! unless @subscription.id
 
-      create_proxy_orders!
-      remove_orphaned_proxy_orders!
+      sync_subscription!
     end
 
     private
 
-    def sync_all!
+    def sync_subscriptions!
       @subscriptions.each do |subscription|
         @subscription = subscription
-        create_proxy_orders!
-        remove_orphaned_proxy_orders!
+        sync_subscription!
       end
     end
 
     def initialise_proxy_orders!
       uninitialised_order_cycle_ids.each do |order_cycle_id|
+        Rails.logger.info "Initializing Proxy Order " \
+          "of subscription #{@subscription.id} in order cycle #{order_cycle_id}"
         proxy_orders << ProxyOrder.new(subscription: subscription, order_cycle_id: order_cycle_id)
       end
+    end
+
+    def sync_subscription!
+      Rails.logger.info "Syncing Proxy Orders of subscription #{@subscription.id}"
+      create_proxy_orders!
+      remove_orphaned_proxy_orders!
     end
 
     def create_proxy_orders!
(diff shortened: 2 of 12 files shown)


sigmundpetersen:
Build is failing and takes very long

sigmundpetersen:
Build is failing

lib/open_food_network/proxy_order_syncer.rb
```diff
@@ -17,29 +17,36 @@ def initialize(subscriptions)
...
-      remove_orphaned_proxy_orders!
+      sync_subscription!
     end
 
     private
 
-    def sync_all!
+    def sync_subscriptions!
       @subscriptions.each do |subscription|
         @subscription = subscription
```
mkllnk: It took me a minute to find this line to understand that your change is correct here. It should really be `sync_subscription!(subscription)` instead of a leaking ivar. Out of scope here though.

db/schema.rb
```diff
@@ -1202,6 +1204,7 @@
     t.string   "whodunnit"
     t.text     "object"
     t.datetime "created_at"
+    t.string   "custom_data"
```
mkllnk: This line belongs to your first commit!

luisramos0: :+1: fixed

lib/open_food_network/proxy_order_syncer.rb
luisramos0: yeah, the use of these ivars is a bit weird in this class...

luisramos0:
there's something wrong with the timestamps in the order_cycle_schedules... it's breaking the build. I can figure it out...
```

       PG::NotNullViolation: ERROR:  null value in column "created_at" violates not-null constraint
       DETAIL:  Failing row contains (6, 6, 6, null, null).
       : INSERT INTO "order_cycle_schedules" ("schedule_id", "order_cycle_id") VALUES (6, 6) RETURNING "id"
```

any idea?

Matt-Yorkley:
I think I saw that whilst playing with the 3-0-stable build as well. The created_at field should set automatically, right? Maybe there's something missing in the schema?

mkllnk:
I've never seen this before but here is a theory: We don't have an ActiveRecord class called `OrderCycleSchedule`. The table enables a many-to-many relationship but is not a model on its own. Maybe ActiveRecord doesn't support timestamps on relationship tables. And would it make sense? Each row is either there or not, `updated_at` would never get updated, right?

luisramos0:
For now I have removed the timestamp in OrderCycleSchedules so we can progress with the other changes.
<details>
This is the commit I removed:
Screenshot 2020-02-07 at 10 06 10
</details>

This way a user can still link/unlink a schedule from an OC without any trace or log. So the original issue is still open after this PR.

luisramos0:
I need a second approval here before I test, @Matt-Yorkley can you review please?

luisramos0:
ok, this is looking as expected in staging FR.
This is on table versions where new field custom_data holds useful info (although the field is not explicitly identified in the table).

Create OC 35 with a schedule 5:
image

Add schedule 5 to OC 33:
image

Create Schedule 5 with OC 34:
image

Update the name of the schedule will list the OCs linked to it:
image

BUT editing a schedule (click on the schedule name) and change the list of OCs, either add or remove OCs, will NOT apppear in teh version DB... this is the only missing use case to close #4462

luisramos0:
Not exactly ideal when the dev tests their own PR.
@Matt-Yorkley or @mkllnk could you please confirm my testing above makes sense and it's clear what remains to be done. If you think all is good, you can merge this PR.

mkllnk:
@Matt-Yorkley Can you take this one on?

luisramos0:
well spotted Pau, I had my local db messed up with code from different PRs.
It's fixed in a new commit :+1:

sigmundpetersen:
@sauloperez do you approve the last commit? 
Did you want to test this as well afterwards?

sauloperez:
All looks good. Thanks @sigmundpetersen . I didn't get to see the logging in all scenarios but I saw some and I've full confidence on what @luisramos0 tested.

PR #4538, opened 2019-12-05
Make product order deterministic

### LINKED ISSUE #4537: French prod: not all product are loading on a shopfront
## Description
A shopfront is not loading entirely. This user has embedded shopfronts on his website. The other shopfronts are working fine.

## Steps to Reproduce

1. Go visit https://www.openfoodfrance.org/micromarche-villeenbois/shop
2. Only the first products will appear

## Severity

## Your Environment
* Version used: v.2.6.5
* Operating System and version (desktop or mobile): Both

mkllnk:
#### What? Why?

Closes #4537

When products are sorted by name and two products have the same name,
their order is undefined. With pagination, two pages can have a
different order of products with the same name which then means that two
pages can return the same product.

Sorting by product id makes sure that the result is always in the same
order, for every page.

I didn't add any automated tests for this because we can't create an "undefined" order returned from the database that would reliably fail.

#### What should we test?

Create a shopfront with more than 10 products. The 10th and 11th product in the shopfront should have the same name. Do all products load?

#### Release notes

Fixed a bug where two products with same name could limit the number of products displayed in the shopfront.

Changelog Category: Fixed

### DIFF
--- a/app/services/products_renderer.rb
+++ b/app/services/products_renderer.rb
@@ -66,7 +66,7 @@ class ProductsRenderer
         .split(",").map { |id| "spree_products.primary_taxon_id=#{id} DESC" }
         .join(", ") + ", spree_products.name ASC, spree_products.id ASC"
     else
-      "spree_products.name ASC"
+      "spree_products.name ASC, spree_products.id"
     end
   end
 


RachL:
I created a shop with 12 products and 2 of witch have the same name: all products correctly appear! This is ready to go 💪 

image

Issue #4322, opened 2019-09-25
Subscription that uses weight calculator cannot be saved or unpaused


kristinalim:
## Description

Unpausing a subscription that uses a weight calculator does not succeed https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5d8b408fb208330019de81bc

## Steps to Reproduce

Picking one shop:
1. Enable subscriptions for the shop. Go to the Settings page for the enterprise, and enabling Shop Preferences > Subscriptions.
2. Create a weight-based shipping fee for the shop. Go to Enterprises > Shipping Methods for the enterprise, and add a shipping method that uses weight calculator. Set the shipping fee after creating the shipping method, e.g. 5.
3. Update one product in the shop to use volume, e.g. 3L.
4. Edit one variant in Step 3 and set a weight, e.g. 2.2kg.
5. Create a new subscription that uses the shipping method, and add the variant. This works. :heavy_check_mark: 
6. Wait for the orders under the subscription in Subscriptions to no longer show "Pending". :heavy_check_mark: 
6. Try to edit the subscription date. It is not saved but there is no warning. :warning: 
7. Try to pause the subscription. It works. :heavy_check_mark: 
8. Try to unpause the subscription. It does not work. :warning: 

## Context

See error on Bugsnag:
https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5d8b408fb208330019de81bc?event_id=5d8b408f0050355630460000&i=sk&m=nw

## Severity

bug-s3: a feature is broken but there is a workaround

luisramos0:
question from @RachL on slack: what is the workaround?

lin-d-hop:
Is this specifically linked to the Weight Calculator?
I have a subscription that does not use the Weight Calculator that does not unpause.

Is the workaround to remove and recreate the subscription?

luisramos0:
yes, this is specific to Weight Calculator

luisramos0:
better create another issue maybe there's another bugsnag entry for your problem.

sauloperez:
OMG... I was right now writing an issue about this :see_no_evil:

sauloperez:
[...]

PR #4539, opened 2019-12-06
Make weight calculator work for SubscriptionLineItems

### LINKED ISSUE #4410: Subs cannot be edited
## Description

When a manager or super admin tries to edit a subscription, the changes are not taken into account.

## Steps to Reproduce

1. Got to `/admin/subscriptions`
2. Create a new subs, and edit it or click straight on the edit button of a subs already existing
3. Change something e.g. add a product or change a quantity. Save
4. The display shows the new info
5. Now come back to `/admin/subscriptions` and check your subscription again. You will see nothing has been actually saved.

## Workaround
The workaround is to cancel the subs and create a new one.

## Severity

## Your Environment
* Version used: v2.5.3

luisramos0:
#### What? Why?

Closes #4410 and #4322 

Make weight calculator work for SubscriptionLineItems.
We do this by making the calculator test if line_item responds to final_weight_volume field. See commit messages for more details.

#### What should we test?
Create subscription with a shipping method with a weight based calculator. Make sure you can now edit the subscription in both the subs list and the edit subs page. Make sure you can paused/unpaused the subscription.

#### Release notes
Changelog Category: Fixed
Fixed bug in subscriptions with shipping methods with price based on weight. These subscriptions can now be edited and paused/unpaused.

### DIFF
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -25,7 +25,7 @@ module Calculator
     end
 
     def line_item_weight(line_item)
-      if line_item.final_weight_volume.present?
+      if final_weight_volume_present?(line_item)
         weight_per_final_weight_volume(line_item)
       else
         weight_per_variant(line_item) * line_item.quantity
@@ -33,13 +33,18 @@ module Calculator
     end
 
     def weight_per_variant(line_item)
-      line_item.variant.andand.weight || 0
+      if variant_unit(line_item) == 'weight'
+        # The calculator price is per_kg so we need to convert unit_value to kg
+        convert_g_to_kg(line_item.variant.andand.unit_value)
+      else
+        line_item.variant.andand.weight || 0
+      end
     end
 
     def weight_per_final_weight_volume(line_item)
-      if line_item.variant.product.andand.variant_unit == 'weight'
-        # Divided by 1000 because grams is the base weight unit and the calculator price is per_kg
-        line_item.final_weight_volume / 1000.0
+      if variant_unit(line_item) == 'weight'
+        # The calculator price is per_kg so we need to convert final_weight_volume to kg
+        convert_g_to_kg(line_item.final_weight_volume)
       else
         weight_per_variant(line_item) * quantity_implied_in_final_weight_volume(line_item)
       end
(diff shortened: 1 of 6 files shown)


app/models/calculator/weight.rb
```diff
@@ -25,21 +25,26 @@ def total_weight(line_items)
...
+      end
     end
 
     def weight_per_final_weight_volume(line_item)
-      if line_item.variant.product.andand.variant_unit == 'weight'
-        # Divided by 1000 because grams is the base weight unit and the calculator price is per_kg
-        line_item.final_weight_volume / 1000.0
+      if variant_unit(line_item) == 'weight'
+        # The calculator price is per_kg so we need to convert final_weight_volume to kg
+        convert_g_to_kg(line_item.final_weight_volume)
```
sauloperez: :ok_hand:

sauloperez:
Didn't we say in https://github.com/openfoodfoundation/openfoodnetwork/issues/4322#issuecomment-546399235 we needed to implement `SubscriptionLineItem#final_weight_volume`?

luisramos0:
Hello Pau, yes we did but I changed my mind after analysis. This PR is what I suggest we do: make weight calculator handle line_items without final_weight_volume. This way we dont need to replicate logic for SubscriptionLineItems and all the weight logic stays in this calculator.

This logic, needed for SubscriptionLineItems, also applies to line_items without final_weight_volume with variant_unit weight. This way we can relax about final_weigth_volume being mandatory for this calculation. It makes the model more simple with one less restriction and the calculator more robust.

RachL:
@luisramos0 I've tested subs with the wieght calculator in shipping and without and I can both edit and unpause :tada:

In terms of automated tests, do we need to introduce some tests during release testing on this or are we now covered?

luisramos0:
I think we are good in terms of testing. the new unit test makes sure the error is shown in the UI and the new functional test validates there is no error.

sauloperez:
> In terms of automated tests, do we need to introduce some tests during release testing on this or are we now covered?

@RachL I was as excited as you about it :joy: . I think this means we can stop testing subs in releases. That's what we last spoke, right? I guess we need to confirm with test coverage.

PR #4542, opened 2019-12-08
Add timestamp to log entries

luisramos0:
#### What? Why?

In addition to #4527

Add timestamp to Rails logger in staging and prod so that info in logs can be accurately compared with data in the DB while debbuging. Logs will look like this now:
`I, [2019-12-05 10:49:57#13624]  INFO -- : Syncing Proxy Orders of subscription 1`

#### What should we test?
Make sure that the logs include a timestamp in staging.
Make sure the logs are on log/staging.log

#### Release notes
Changelog Category: Added
Added timestamp to log entries to improve accuracy on analysis.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -34,13 +34,14 @@ Openfoodnetwork::Application.configure do
   config.action_mailer.default_url_options = { protocol: 'https' }
 
   # See everything in the log (default is :info)
-  config.log_level = :info
+  # config.log_level = :debug
 
-  # Configure logging
-  Rails.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
-  Rails.logger.formatter = Logger::Formatter.new
-  Rails.logger.datetime_format = "%Y-%m-%d %H:%M:%S"
-  config.logger = Rails.logger
+  # Configure logging for Rails 3.2:
+  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(Rails.root.join("log", "#{Rails.env}.log")))
+  config.logger.formatter = Logger::Formatter.new
+  config.logger.datetime_format = "%Y-%m-%d %H:%M:%S"
+  # Once we get to Rails 4.0, we can replace the above with:
+  #config.log_formatter = Logger::Formatter.new.tap { |f| f.datetime_format = "%Y-%m-%d %H:%M:%S" }
 
   # Use a different cache store in production
   memcached_value_max_megabytes = ENV.fetch("MEMCACHED_VALUE_MAX_MEGABYTES", 1).to_i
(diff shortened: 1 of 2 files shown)


luisramos0:
i did that initially, added to applicationn.rb but the build on semaphore was broken with too many log entries. I am not sure how to control the log level on semaphore while still keep the log formatting in dev and test.

mkllnk:
Ouch, okay. So my best idea would be to still apply it only to staging and production by default.

The dev and test logs are actually filling up my disk from time to time. By default we could assume that we don't need logs (at least locally). We can easily activate them when we need them.

What would be the best way to control it on CI? An environment variable that we can set before clicking "rebuild"? Or changing the a config variable in a new commit?

mkllnk:
I deployed this to au-staging and watched the `log` directory. The delayed job log appeared but `staging.log` did not appear after some browsing around.

I started `bundle exec rails console` and was presented with debug logs with timestamps on STDOUT. I think we need to set the log level and the file:
```
> pp Rails.logger
#<ActiveSupport::TaggedLogging:0x0000563813ea4c40
 @logger=
  #<Logger:0x0000563813ea4d80
   @default_formatter=
    #<Logger::Formatter:0x0000563813ea4d08 @datetime_format=nil>,
   @formatter=
    #<Logger::Formatter:0x0000563813ea4c18
     @datetime_format="%Y-%m-%d %H:%M:%S">,
   @level=0,
   @logdev=
    #<Logger::LogDevice:0x0000563813ea4cb8
     @dev=#<IO:<STDOUT>>,
     @filename=nil,
     @mon_count=0,
     @mon_mutex=#<Thread::Mutex:0x0000563813ea4c90>,
     @mon_owner=nil,
     @shift_age=nil,
     @shift_size=nil>,
   @progname=nil>>
```

mkllnk:
@luisramos0 I thought I could speed up our dev pipe by working on this myself. It was good to get my head around Rails' config as well.

So I found that everything will be easier with Rails 4 but for now, we have to replicate a bit of Rails code to achieve the same log files.

I also found that `Rails.logger` is not initialized during the configuration. After the config, Rails takes all the config info to create a logger and set it to `Rails.logger`. So we don't have to assign that variable. And if we set `config.logger` then Rails will just use that one. Stuff like `config.log_level` is ignored in that case.

I tested this case locally with `bundle exec rails console production` and staging. It would be great if you could give it a review and test it on a real staging server.

luisramos0:
awesome, thanks Maikel!
I'll give it a try.

luisramos0:
thanks Maikel and apologies for taking so long to get this tested :-(

the logs are looking good in staging :+1:
filename is staging.log and they include the timestamp.

PR #4545, opened 2019-12-09
Make whenever properly read the S3 bucket

sauloperez:
#### What? Why?

For unknown reasons the magic [Figaro](https://github.com/laserlemon/figaro) does to turn keys in `config/application.yml` into ENV vars that can be read through Ruby's `ENV[]` is not working in `config/schedule.rb`.

As a result, the `db2fog` tasks are not translated into cron entries which led to **not having automatic backups**.

:warning: **UK, FR, AU and Katuma are all missing the backup entries in crontab.** So it must be affecting everyone. :warning: 

I checked the change directly in Katuma staging and backups now show up in S3.

I also took the time to see if we could get some output into a `log/cron.log` but the [gem we use](https://github.com/itbeaver/db2fog/blob/master/lib/db2fog.rb) doesn't seem to ouptut anything :see_no_evil: We could monitor the backups and set up an alert if we had it...

#### What should we test?

After the deploy, ssh into the staging server as `openfoodnetwork` and run  `crontab -e`. Then update the line that runs the backup (should be the 2nd one) to have the following content

```
* * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=staging bundle exec rake db2fog:backup --silent'
```
save and exit. Now go check S3. There should be a backup in a minute.

Deploy again when done to restore the backup frequency.

#### Release notes

Fixed automatic database backups to read the S3 bucket name. 

Changelog Category: Fixed

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -13,11 +13,11 @@ job_type :enqueue_job,  "cd :path; :environment_variable=:environment bundle exe
 
 
 every 1.day, at: '2:45am' do
-  rake 'db2fog:clean' if ENV['S3_BACKUPS_BUCKET']
+  rake 'db2fog:clean' if app_config['S3_BACKUPS_BUCKET']
 end
 
 every 4.hours do
-  rake 'db2fog:backup' if ENV['S3_BACKUPS_BUCKET']
+  rake 'db2fog:backup' if app_config['S3_BACKUPS_BUCKET']
 end
 
 every 5.minutes do


sauloperez:
@mkllnk @RachL @Matt-Yorkley would you mind checking your S3 buckets to double-check when was the last backup? I couldn't with the credentials listed in Bitwarden.

RachL:
@sauloperez Pau if you can't I can't either ;-)

sauloperez:
> What is the problem? No backups for katuma live?

Yep, last backup dates May 7, 2019 10:11:11 AM GMT+0200. And I assume it's the cause for others since there's no cron entry for their backups.

> @sauloperez Pau if you can't I can't either ;-) 

We should look into that. Maybe the credentials are not correct in Bitwarden?

RachL:
@sauloperez but then who has the correct credentials? @Matt-Yorkley maybe?

mkllnk:
I can confirm that the last Australian back is from 17 April 2019.

mkllnk:
I'm wondering how to test this. I could configure backups for au-staging and see if it works?

sauloperez:
> I'm wondering how to test this. I could configure backups for au-staging and see if it works?

Exactly @mkllnk . The deploy playbook will run the `whenever` command that turns `config/schedule.rb` entries into cron ones. What you can do then to not wait 4 hours is to modify `openfoodnetwork`'s crontab to run every minute. Then, you should see a new backup in S3.

sauloperez:
> For unknown reasons

I found out why. The `whenever` [command](https://github.com/javan/whenever/blob/master/bin/whenever) does not boot the app as we're used to with Rails commands, therefore no gems will be loaded. In this case, Figaro.

Matt-Yorkley:
> I found out why. The `whenever` [command](https://github.com/javan/whenever/blob/master/bin/whenever) does not boot the app as we're used to with Rails commands, therefore no gems will be loaded, in this case, Figaro.

Nice investigation! :heart:

sauloperez:
Whoever testes this, I updated the testing steps in the description.

luisramos0:
the aws FR credentials in bitwarden dont work, but the ones for AUS do, so I think I'll be able to test this in AU-staging.

sauloperez:
> the aws FR credentials in bitwarden dont work

I created https://github.com/openfoodfoundation/openfoodnetwork/issues/4552 for it @RachL

RachL:
@sauloperez thanks, but again I'm totally powerless... if what's in bitwarden don't work I don't know who has the correct info...

@pacodelaluna do you?

luisramos0:
@sauloperez test fail I think, I deployed this PR to staging FR and the crontab does not include the backup:
```
# Begin Whenever generated tasks for: /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb at: 2019-12-11 13:46:54 +0100
0,5,10,15,20,25,30,35,40,45,50,55 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current; RAILS_ENV=staging bundle exec script/enqueue HeartbeatJob 0'

0,5,10,15,20,25,30,35,40,45,50,55 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current; RAILS_ENV=staging bundle exec script/enqueue SubscriptionPlacementJob 0'

0,5,10,15,20,25,30,35,40,45,50,55 * * * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current; RAILS_ENV=staging bundle exec script/enqueue SubscriptionConfirmJob 0'

# End Whenever generated tasks for: /home/openfoodnetwork/apps/openfoodnetwork/current/config/schedule.rb at: 2019-12-11 13:46:54 +0100
```

pacodelaluna:
@RachL My Bitwarden account is blocked... Seems the only way to unblock it is to delete it...

sauloperez:
@pacodelaluna I don't see anything wrong with your account. It uses the email "francois@kokali.net".

sauloperez:
@luisramos0 we missed something because I removed the crontab backup entry, deployed, Ansible reported the change and now the crontab has the backup entry again.

sauloperez:
I'm taking the lead and merging this. I tested it thoroughly and it's quite important.

sauloperez:
I can confirm backups are working again. If that wasn't the case for a particular instance, it's a different issue.

luisramos0:
ok :+1:

but lets check all the instances, ok?
- AU
- UK
- FR
- KA - ✅ 
- BE
- DE
- US
- CA

PR #4548, opened 2019-12-10
Improve truncate data process

luisramos0:
#### What? Why?
Delete extra inventory_units that are for some unknown reason connected to shipments of orders to delete, but not connected to orders to delete

#### What should we test?
This problem appeared with uk live data and I tested it locally with UK live data. I dont think we need to validate this again.

#### Release notes
Changelog Category: Changed
Improved the data truncation process that we use to move data from live environments to other testing environments.

### DIFF
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -9,6 +9,9 @@ namespace :ofn do
 
       sql_delete_from "
         spree_inventory_units #{where_order_id_in_orders_to_delete}"
+      sql_delete_from "
+        spree_inventory_units
+        where shipment_id in (select id from spree_shipments #{where_order_id_in_orders_to_delete})"
 
       truncate_adjustments
 


luisramos0:
yeah, good point @kshlyk we will have to do that to make delete of orders and enterprises possible in the app. for example for: https://github.com/openfoodfoundation/openfoodnetwork/issues/2971

sigmundpetersen:
Hey @luisramos0 did you want to follow up Maikel's comment here?

luisramos0:
yes @sigmundpetersen thanks, done.

luisramos0:
just to clarify, this needs a second approval from a core dev.

luisramos0:
@Matt-Yorkley can you approve this one please? Thanks!

lib/tasks/data/truncate_data.rake
```diff
@@ -9,6 +9,9 @@ namespace :ofn do
 
       sql_delete_from "
         spree_inventory_units #{where_order_id_in_orders_to_delete}"
+      sql_delete_from "
+        spree_inventory_units
+        where shipment_id in (select id from spree_shipments #{where_order_id_in_orders_to_delete}"
```
kristinalim: This seems to be missing a closing parenthesis.

mkllnk: I added it now @kristinalim.

luisramos0: ah, I dont know how did this happen. I used this code before but must have made the mistake when committing it...
thanks Maikel.
I tested it now locally and it is working. With the extra parenthesis :-)

Issue #4547, opened 2019-12-10
Timeout on Orders and Fulfilment Report


lin-d-hop:
## Description

This report is timing out in UK and Australia.... detail

## Expected Behavior

Report loads in 3-4 minutes

## Actual Behaviour

Timeout at 12 minutes in UK

## Steps to Reproduce

1. UK Timeout:
Customer Totals Report
A user with only access to one enterprise - lynnedavis[at]gmail[dot]com
Tamar Valley Food Hub
Dates: 2019-11-29 to 2019-12-05
OC: TVFH20191206

## Workaround

To run the report 1 or 2 days at a time.

## Severity

bug-s2: a non-critical feature is broken, no workaround

## Your Environment

* Version used: 2.6.6
* Browser name and version: Chrome Firefox
* Operating System and version (desktop or mobile):

luisramos0:
First thing I notice is that you have permissions on Tamar Valley Apple Co-op which is a producer, not a distributor. That means that the algorithm to hide customer details is having to work a lot.

luisramos0:
I wonder if you have changed permissions of your user meanwhile, with the co-op enterprise, as a producer, I can run the report:
![Screenshot 2019-12-10 at 10 26 08](https://user-images.githubusercontent.com/1640378/70521683-8a8eff80-1b37-11ea-8fd5-647d445fe024.png)

luisramos0:
I assumed this is wrong and I am now testing with a user with access to the Hub account.

From 2019-11-29+00:00 to 2019-12-03+00:00 the report runs in 2.4 mins.

But for the full date range it times out after 2.8 mins...
I read above "Timeout at 12 minutes in UK" - do you see 12mins in the network tab of your browser?

luisramos0:
the full dates are taking 6.28mins locally. I will now try to understand where it is taking all this time.

PR #4549, opened 2019-12-10
Use arel in order permissions visible orders and editable orders so that we dont have queries with gigantic IN clauses

luisramos0:
#### What? Why?

Relates and improves the situation of #4547
Problem: the | operators here were converting the relations to long lists of IDs, in our current particular issue, an IN clause with 100k order_ids.
Now we use arel and build one single query for visible_orders and editable_orders.

#### What should we test?
We need to make sure the report is working correctly and hiding line items for producers who are not distributors. One particular case that makes the code slightly more complicated is that these producers should only see their line items, not all the line_items in the orders they have access to: so a producer should only see their products in the orders.

This PR improves the performance of the report from 50secs to 42secs in my tests.

There's a small change in order_and_distributor report as well. I think we can make a verification of this report as well.

#### Release notes
Changelog Category: Changed
Improved query of Orders and Fulfillment customer totals report and the performance of the report.

### DIFF
--- a/app/services/permissions/order.rb
+++ b/app/services/permissions/order.rb
@@ -7,17 +7,25 @@ module Permissions
 
     # Find orders that the user can see
     def visible_orders
-      Spree::Order.where(id:
-        managed_orders.select(:id) |
-        coordinated_orders.select(:id) |
-        produced_orders.select("spree_orders.id"))
+      Spree::Order.
+        with_line_items_variants_and_products_outer.
+        where(
+          Spree::Order.arel_table.
+            # Grouping keeps the 2 where clauses from produced_orders_where_values inside parentheses
+            #   This way it makes the OR work between the 3 types of orders:
+            #     produced, managed and coordinated
+            grouping(produced_orders_where_values).
+            or(managed_orders_where_values).
+            or(coordinated_orders_where_values)
+        )
     end
 
     # Any orders that the user can edit
     def editable_orders
-      Spree::Order.where(id:
-        managed_orders.select(:id) |
-        coordinated_orders.select(:id) )
+      Spree::Order.where(
+        managed_orders_where_values.
+          or(coordinated_orders_where_values)
+      )
     end
 
     def visible_line_items
@@ -28,27 +36,35 @@ module Permissions
 
     # Any line items that I can edit
     def editable_line_items
-      Spree::LineItem.where(order_id: editable_orders)
+      Spree::LineItem.where(order_id: editable_orders.select(:id))
     end
 
     private
 
     # Any orders placed through any hub that I manage
-    def managed_orders
-      Spree::Order.where(distributor_id: @permissions.managed_enterprises.select("enterprises.id"))
+    def managed_orders_where_values
+      Spree::Order.
+        where(distributor_id: @permissions.managed_enterprises.select("enterprises.id")).
+        where_values.
+        reduce(:and)
     end
 
     # Any order that is placed through an order cycle one of my managed enterprises coordinates
-    def coordinated_orders
-      Spree::Order.where(order_cycle_id: @permissions.coordinated_order_cycles.select(:id))
+    def coordinated_orders_where_values
+      Spree::Order.
+        where(order_cycle_id: @permissions.coordinated_order_cycles.select(:id)).
+        where_values.
+        reduce(:and)
     end
 
-    def produced_orders
+    def produced_orders_where_values
       Spree::Order.with_line_items_variants_and_products_outer.
         where(
           distributor_id: granted_distributor_ids,
           spree_products: { supplier_id: enterprises_with_associated_orders }
-        )
+        ).
+        where_values.
+        reduce(:and)
     end
 
     def enterprises_with_associated_orders
@@ -69,7 +85,7 @@ module Permissions
 
     # Any from visible orders, where the product is produced by one of my managed producers
     def produced_line_items
-      Spree::LineItem.where(order_id: visible_orders.select(:id)).
+      Spree::LineItem.where(order_id: visible_orders.select("DISTINCT spree_orders.id")).
         joins(:product).
         where(spree_products:
         {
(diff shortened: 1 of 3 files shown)


app/services/permissions/order.rb
```diff
@@ -28,27 +36,35 @@ def visible_line_items
...
     end
 
     private
 
     # Any orders placed through any hub that I manage
-    def managed_orders
-      Spree::Order.where(distributor_id: @permissions.managed_enterprises.select("enterprises.id"))
+    def managed_orders_where_values
+      Spree::Order.
+        where(distributor_id: @permissions.managed_enterprises.select("enterprises.id")).
```
sauloperez: why don't we `inner join` orders with `@permissions.managed_enterprises`? By checking the scope as is defined in `Enterprise` I don't see why we need the DB to go fetch the ids, even if within the same query. I'd check if doing so brings any improvement.

app/services/permissions/order.rb
```diff
@@ -7,17 +7,25 @@ def initialize(user)
...
+        with_line_items_variants_and_products_outer.
+        where(
+          Spree::Order.arel_table.
+            # Grouping keeps the 2 where clauses from produced_orders_where_values inside parentheses
+            #   This way it makes the OR work between the 3 types of orders:
+            #     produced, managed and coordinated
+            grouping(produced_orders_where_values).
+            or(managed_orders_where_values).
+            or(coordinated_orders_where_values)
+        )
```
sauloperez: IMO it'd be worth splitting this class into smaller purpose-specific classes like `visible_orders`. Even if only used within this `*_orders` methods, it'd give us room for comments on extracted private methods like the one above. Up to you. Just suggesting.

luisramos0:
To my great disappointment and surprise, I have just confirmed that this PR, although it changes the query (from very long ids list to a subquery) it does not improve performance of the report in the S2, the report keeps taking 42sec (including #4550).

SO, I wonder if we should discard this PR as it makes the code more complicated. **I think we should keep this PR** as I think the long list of IDs will make the query blow up at some point. One answer here says there's a technical limit of 32k values, I dont think it's true as we are already past that 32k values.
https://stackoverflow.com/questions/1009706/postgresql-max-number-of-parameters-in-in-clause

app/services/permissions/order.rb
```diff
@@ -28,27 +36,35 @@ def visible_line_items
...
+        where_values.
+        reduce(:and)
     end
 
     # Any order that is placed through an order cycle one of my managed enterprises coordinates
-    def coordinated_orders
-      Spree::Order.where(order_cycle_id: @permissions.coordinated_order_cycles.select(:id))
+    def coordinated_orders_where_values
+      Spree::Order.
+        where(order_cycle_id: @permissions.coordinated_order_cycles.select(:id)).
```
Matt-Yorkley: Semaphore is complaining about an ambiguous `id` reference. Maybe this needs to be `"DISTINCT spree_orders.id"` as well?

luisramos0:
This PR "arelizes" the visible_orders part of the query and #4551 arelizes the visible line_items part of the query. But #4551 has a disastrous impact on performance. The DB is not able to handle the complex query.

app/services/permissions/order.rb
luisramos0: I am always up for some extraction :-)
in this case though I prefer to keep it together because, we are at around 100lines for the class, still ok, and the methods are very closely related: in #4551 you see the line_items query uses the orders query where clauses. Maybe on the next iteration, ok?

app/services/permissions/order.rb
luisramos0: ah, this is a really good clue!!! but this condition is placed within a series of clauses with ANDs and ORs...
So, instead of
```sql
OR spree_orders.distributor_id IN
           (SELECT enterprises.id
            FROM enterprises
            INNER JOIN enterprise_roles ON enterprise_roles.enterprise_id = enterprises.id
            WHERE (enterprise_roles.user_id = 204238) ))

```
we could do
```sql
INNER JOIN enterprises ON enterprises.id = spree_orders.distributor_id
INNER JOIN enterprise_roles ON enterprise_roles.enterprise_id = enterprises.id
  WHERE (enterprise_roles.user_id = 204238)
```
but the other conditions would not work because this is just one of the conditions, the user does not have to be the manager of the enterprise in all cases...

we could do just for enterprises because the distributor is always there:
```sql
INNER JOIN enterprises ON enterprises.id = spree_orders.distributor_id
WHERE (enterprises.is_primary_producer = 't'
     AND enterprises.id IN ( SELECT enterprise_id FROM enterprise_roles
                                           WHERE (enterprise_roles.user_id = 204238) ))
```

but on top of this there's the super admin condition...

luisramos0: maybe you have a clear idea about how the activerecord code should look like?

app/services/permissions/order.rb
luisramos0: yeah, it's in order_and_distributors report, fixed in new commit.

app/services/permissions/order.rb
sauloperez: I think we will naturally end up with multiple classes as we evolve this crazy queries

sauloperez:
> I think the long list of IDs will make the query blow up at some point

That's exactly what I fear but perhaps it is better to wait and give it another look. The next release already greatly benefits from #4550, isn't it @luisramos0 ?

sigmundpetersen:
Needs a rebase 👷

luisramos0:
rebased

luisramos0:
I think it would be good to verify the report performance in PROD before and after this PR. Just to make sure it does not make it slower. In my local tests with prod data the improvement was minimal but noticeable.

RachL:
Alright so I've checked the order and fulfillment section with all the sub reports.

I've done this using old test data : relationship between Mary and Fred (how Fred sees the report) and with a fresh enterprise.

In both cases results where consistent with orders. Moving to ready to go.

luisramos0:
all right!
can someone take a benchmark result for some results with some data set in prod now, so we can then verify the impact of this PR after the next release? @lin-d-hop @RachL

PR #4560, opened 2019-12-11
Billing address in pdf invoice fixed

### LINKED ISSUE #4473: Invoices are showing shipping address instead of billing address
## Description
Invoices are showing shipping address instead of billing address.

The label is invoice_billing_address and then the field picked up is the ship_address 
https://github.com/openfoodfoundation/openfoodnetwork/blob/9c8c71bd0877a2ffbf3dc68fc75a0049ffa68c78/app/views/spree/admin/orders/invoice2.html.haml#L45

## Steps to Reproduce
1. Go to `/admin/orders?q[s]=completed_at+desc`
2. Select an order and generate a PDF invoice
3. Compare the address on the invoice and the address mentioned in customer detail menu at `/admin/orders/ORDERNUMBER/customer`
4. See that the invoice is not mentioning billing address as it should.

## Workaround
None I can recall except looking at the order customer details?

## Severity

## Your Environment
* Version used: 2.6.2

kshlyk:
#### What? Why?

Closes #4473 

Shipping address was displayed in PDF invoice instead of billing address. I've changed the code to display billing address.

#### What should we test?
1. **PRINT INVOICES** button on `admin/orders` page. Should be displayed billing address in the generated invoice.
2. **Actions - Print invoice** button on particular order's page in Admin area. Should be displayed billing address in the generated invoice.

#### Release notes

PDF invoices now show the billing address instead of the shipping address. Please let us know if you need the shipping address on your invoices.

Changelog Category: Fixed | Changed

### DIFF
--- a/app/views/spree/admin/orders/invoice2.html.haml
+++ b/app/views/spree/admin/orders/invoice2.html.haml
@@ -44,14 +44,14 @@
       %td{ :align => "right" }
         = t :invoice_billing_address
         %br
-        %strong= @order.ship_address.full_name
+        %strong= @order.bill_address.full_name
         - if @order.andand.customer.andand.code.present?
           %br
           = "Code: #{@order.customer.code}"
         %br
-        = @order.ship_address.address_part1
+        = @order.bill_address.address_part1
         %br
-        = @order.ship_address.address_part2
+        = @order.bill_address.address_part2
 
 = render 'spree/admin/orders/invoice_table2'
 


Matt-Yorkley:
Do we need some product input on this? I think a lot of hubs use the invoices as packing sheets, which means they will need the shipping address (for matching packed boxes with deliveries). @kirstenalarsen ?

RachL:
@Matt-Yorkley currently the label is billing address but the content is shipping address, which is confusing for both users and eaters. The billing address is mandatory on an invoice. Maybe we can add a shipping address but that would require another issue.

lin-d-hop:
Agree with @RachL. If this is problematic for some users we can create another issue and update the 'Customer Instructions' to 'Delivery Instructions' that can include shipping method and customer notes.

Matt-Yorkley:
OK, just checking :+1:

RachL:
@kshlyk Hello! I've stage this, but I can still see the shipping address, example:

image


image

Do you know where the problem can come from?

kshlyk:
Hi @RachL. I'm checking

kshlyk:
The thing is we have two styles of invoice. There is a checkbox in `/admin/invoice_settings/edit` called `USE THE ALTERNATIVE INVOICE MODEL THAT INCLUDES TOTAL TAX BREAKDOWN PER RATE AND TAX RATE INFO PER ITEM (NOT YET SUITABLE FOR COUNTRIES DISPLAYING PRICES EXCLUDING TAX)`. 
I've changed one style. I think I should change the second one to keep logic transparent for customers.

RachL:
@kshlyk great. How do you want to proceed, will you continue to work on this PR?

kshlyk:
Yes, I will. But I need product input. Is billing address suitable in this variant of invoice layout? It looks slightly strange within "From To" context. Don't you think so?

RachL:
@kshlyk Ah interesting, I thought it was a translation problem. Because currently this "from to" does not appear in production. Legally the invoice needs the billing address. So if you need to remove this variant for now I think we can go wit it.

@luisramos0 do you happen to know if this variant is used somewhere? I don't remember seeing a "from - to" display somewhere.

luisramos0:
in my head this issue is about invoice2 which is used in FR. that is what the PR changes and that is what I'd test and ship.
if we want to change the order template we need to see who's using it, how to change it, etc.

I think, I am not sure, the other template is used yes, UK or AU, @lin-d-hop ?

RachL:
Aaaaaaaah so the issue does not exist in UK/AU? If I stage this on FR staging it should work?

luisramos0:
it's only dependent on the checkbox in the invoice settings. you just need to activate the alternative model to test this PR.

RachL:
Ok so I retested this and indeed the two templates will now show different addresses:

"French" template:
image

The other template (left is "from" and right is "to")
image

I would find it weird that this template doesn't require a billing address as well. I opened the issue not just based on one template (I actually didn't know we had two :D ). But I agree we shouldn't postponed merging the work that has already be done, especially in holidays season.

@lin-d-hop or @kirstenlarsen if you need a change on your template please open a new issue.

kshlyk:
Thank you @RachL. Merry Christmas and Happy New Year!

luisramos0:
great, thanks @RachL

PR #4575, opened 2019-12-13
Fix route on "continue" button on admin adjustments page

### LINKED ISSUE #4128: Workflow of creating an order as an admin is broken
## Description
When creating an order as an admin, pressing "continue" on the adjustments step will result in the loss of the order thus created.

The workaround is to not click on "continue", but go directly to the submenu "Payments". But for some users this workaround is difficult to find. We've had complaints in France.

## Steps to Reproduce
1. Log in as a manager (admin)
2.  Go to `/admin/orders/new`
3. (Select a shop if you are a super admin or a manager with several shop) and select an OC
4. Add products.
5. Save. This should get you to the next step: customer credential info
6. Add a customer.
7. Go to next step `/adjustments`
8. If you click on "continue" you are redirected to `/admin/orders` and your order is lost

## Workaround
9. Re-create your order again, but this time rather than clicking on "continue" go strait to "payments" in the right menu. As soon as a payment method is chosen your order will be correctly recorded.

## Animated Gif/Screenshot
image

## Severity

## Your Environment
* Version used: v2.2.2
* Browser name and version: Firefox 68

Matt-Yorkley:
#### What? Why?

Closes #4128 

The route on the "continue" button on the admin order adjustments was incorrect, and redirected to the orders index page in some cases, which was breaking the admin order creation workflow.

The previous logic was branching: if the order is in cart state, link to the new payments page, otherwise link to the orders index page (which is really unhelpful).

It turns out the main payments route will show the new payment form if no payments currently exist, and show the existing payments if there are existing payments. So the branching was not needed, the single route works well for both cases.

#### What should we test?

Admin order creation workflow from the adjustments page. For both new orders without payments, and completed orders with payments, the "continue" button (from adjustments) should redirect correctly to the payments page. If the order has no payments, it should show the new payments form. If the order already has payments it should show the existing payments.

#### Release notes

Fixed "continue" button route on admin order adjustments page.

Changelog Category: Fixed

### DIFF
--- a/app/views/spree/admin/adjustments/index.html.haml
+++ b/app/views/spree/admin/adjustments/index.html.haml
@@ -10,4 +10,4 @@
 
 = render :partial => 'adjustments_table'
 
-= button_link_to t(:continue), @order.cart? ? new_admin_order_payment_url(@order) : admin_orders_url, :icon => 'icon-arrow-right'
+= button_link_to t(:continue), admin_order_payments_url(@order), :icon => 'icon-arrow-right'


RachL:
>Admin order creation workflow from the adjustments page. For both new orders without payments, and completed orders with payments, the "continue" button (from adjustments) should redirect correctly to the payments page. If the order has no payments, it should show the new payments form. If the order already has payments it should show the existing payments.

It works :tada:

PR #4578, opened 2019-12-13
Removing duplicate API method soft_delete for both products and variants

### LINKED ISSUE #4278: [API clean up] Move soft_delete to destroy in both api/products and api/variants
The actions soft_delete and destroy are duplicated. We dont have hard_deletes any longer so we need to move soft_delete to destroy and remove destroy from both api/products and api/variants.

DELETE api/product/21/soft_delete should be converted to DELETE api/product/21
This endpoint is used in app/assets/javascripts/admin/bulk_product_update.js.coffee
API doc in swagger needs to be updated afterwards.

### Context
This came up from the API documentation work #3001 

### Impact and timeline
This is kind of urgent as the duplicated actions dont have similar implementations...

kshlyk:
#### What? Why?

Closes #4278 

There were two different API methods: `destroy` and `soft_delete` for products and variants. `Soft_delete` method performed deleting through manual operating with `deleted_at` field. `Destroy` method performed deleting through `destroy` method invocation, but DB deleting indeed didn't performed as Spree Commerce used `acts_as_paranoid` for both product and variant models. As a result we had two similar methods.

I've removed the `soft_delete` method and replace it's invocations on `destroy` method invocation. I've accordingly changed specs and swagger.yaml.

#### What should we test?
We should test bulk products edit page: /admin/products. We need to check products and variants deleting.

#### Release notes
product's `soft_delete` API method (`/api/products/apples/soft_delete` for example) has been removed;
variant's `soft_delete` API method (`/api/products/apples/variants/3/soft_delete` for example) has been removed;

Changelog Category: Changed

#### Documentation updates
API documentation should be updated

### DIFF
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -142,7 +142,7 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     if confirm("Are you sure?")
       $http(
         method: "DELETE"
-        url: "/api/products/" + product.id + "/soft_delete"
+        url: "/api/products/" + product.id
       ).success (data) ->
         $scope.products.splice $scope.products.indexOf(product), 1
         DirtyProducts.deleteProduct product.id
@@ -157,7 +157,7 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
         if confirm(t("are_you_sure"))
           $http(
             method: "DELETE"
-            url: "/api/products/" + product.permalink_live + "/variants/" + variant.id + "/soft_delete"
+            url: "/api/products/" + product.permalink_live + "/variants/" + variant.id
           ).success (data) ->
             $scope.removeVariant(product, variant)
     else
--- a/app/controllers/api/products_controller.rb
+++ b/app/controllers/api/products_controller.rb
@@ -42,8 +42,8 @@ module Api
     def destroy
       authorize! :delete, Spree::Product
       @product = find_product(params[:id])
-      @product.update_attribute(:deleted_at, Time.zone.now)
-      @product.variants_including_master.update_all(deleted_at: Time.zone.now)
+      authorize! :delete, @product
+      @product.destroy
       render json: @product, serializer: Api::Admin::ProductSerializer, status: :no_content
     end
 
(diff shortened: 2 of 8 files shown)


luisramos0:
there's a broken spec:
```
1) Api::VariantsController as an administrator doesn't delete the only variant of the product
     Failure/Error: expect(assigns(:variant).errors[:product]).to include "must have at least one variant"
     
     NoMethodError:
       undefined method `errors' for nil:NilClass
     # ./spec/controllers/api/variants_controller_spec.rb:160:in `block (3 levels) in <top (required)>'
```

kshlyk:
Thank you, @luisramos0. I'm working on broken spec.

kshlyk:
Thank you @mkllnk. It's extremely important notice. Unfortunately I've missed it. It's my fault.
I think it is not good idea to allow users to destroy records through API as we've decided to remain both product and variant destroy method in 'soft' implementation. But currently as we have no external API consumers we can control this behaviour by `with_deleted` param.

luisramos0:
nice catch @mkllnk :+1:

I think we should keep it as simple as possible. I dont think we need product input.
We should have a single endpoint to delete products and that endpoint should soft delete products.

so I'd add a clause like 
@product.destroy if @product.delete_at != null

I think the race condition of two users clicking delete of the same product at the same time is very remote possibility. I think we have many race conditions oissues like this one the app. I'd forget about it.
If you think we should fix it we can add some safety to this process :+1:

kshlyk:
> so I'd add a clause like
> @product.destroy if @product.delete_at != null

I guess you mean just the opposite there? To prevent destroying if record has been already deleted. Don't you?

luisramos0:
yes! :-)

kshlyk:
Btw shall we handle this case on cancan level to deny users to delete already deleted records? What do you make about it @luisramos0, @mkllnk ?

luisramos0:
I dont think so, delete deleted records is not a feature we need or was asked for and so we should not implement it.

Matt-Yorkley:
@mkllnk I think we're actually using the [paranoia](https://github.com/rubysherpas/paranoia) gem, which is a reimplementation of the original [acts_as_paranoid](https://github.com/ActsAsParanoid/acts_as_paranoid) gem.

I don't think it has this issue with calling #destroy twice..?

kshlyk:
@Matt-Yorkley is right. We use paranoia gem. It has the same declaration `acts_as_paranoid` in model, but don't have problem with double invocation of destroy method.

RachL:
Everything is looking good in /admin/products. Deleting especially is working fine. Moving to ready to go!

luisramos0:
done.
thank you @kshlyk 🎉

PR #4580, opened 2019-12-14
Use asynchronous requests for order capture and ship actions

### LINKED ISSUE #4377: Don't reload Orders page after payment is captured
## Description
- As an enterprise user
- On page Orders (https://openfoodnetwork.org.au/admin/orders)
- I want to be able to quickly capture payment on multiple orders, including where I have selected them according to filters  
see discussion at https://community.openfoodnetwork.org/t/enable-quick-capture-of-payments-for-a-batch-of-orders/1478/9

Enterprise user should be able to set filters on the orders page to select a group of orders, and then 'capture payments' on these without losing their filter selection and being returned to home orders page each time

Create multiple orders with cash payments i.e. order is complete but not paid. Orders that are complete but not paid have a small tick at the right hand side which allows the payment to be captured

Set filters to select the orders you are looking for e.g. all orders from a particular user, order cycle or distributor etc

Hit capture payment button on the first order

Page SHOULD NOT reload to Orders page with no filters selected. 

The capture payment action should take place while keeping the same selection of filters on the page. So the page should then look like this

[This should NOT happen]

Matt-Yorkley:
#### What? Why?

Closes #4377

Actions on capture and ship buttons in the admin orders page are now handled asynchronously, with some new user feedback added.

#### What should we test?

Capture and ship buttons in admin orders page now work without reloading the page and show some basic user feedback.

#### Release notes

Use asynchronous requests for capture and ship actions on admin orders page 

Changelog Category: Changed

### DIFF
--- a/app/serializers/api/admin/order_serializer.rb
+++ b/app/serializers/api/admin/order_serializer.rb
@@ -1,8 +1,8 @@
 class Api::Admin::OrderSerializer < ActiveModel::Serializer
   attributes :id, :number, :user_id, :full_name, :email, :phone, :completed_at, :display_total,
              :edit_path, :state, :payment_state, :shipment_state,
-             :payments_path, :ship_path, :ready_to_ship, :created_at,
-             :distributor_name, :special_instructions, :payment_capture_path,
+             :payments_path, :ready_to_ship, :ready_to_capture, :created_at,
+             :distributor_name, :special_instructions,
              :item_total, :adjustment_total, :payment_total, :total
 
   has_one :distributor, serializer: Api::Admin::IdSerializer
@@ -28,15 +28,9 @@ class Api::Admin::OrderSerializer < ActiveModel::Serializer
     spree_routes_helper.admin_order_payments_path(object)
   end
 
-  def ship_path
-    spree_routes_helper.fire_admin_order_path(object, e: 'ship')
-  end
-
-  def payment_capture_path
+  def ready_to_capture
     pending_payment = object.pending_payments.first
-    return '' unless object.payment_required? && pending_payment
-
-    spree_routes_helper.fire_admin_order_payment_path(object, pending_payment.id, e: 'capture')
+    object.payment_required? && pending_payment
   end
 
   def ready_to_ship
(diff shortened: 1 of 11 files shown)


Matt-Yorkley:
RowActions

app/serializers/api/admin/order_serializer.rb
```diff
@@ -28,15 +28,9 @@ def payments_path
...
-    spree_routes_helper.fire_admin_order_path(object, e: 'ship')
-  end
-
-  def payment_capture_path
+  def ready_to_capture
     pending_payment = object.pending_payments.first
-    return '' unless object.payment_required? && pending_payment
-
-    spree_routes_helper.fire_admin_order_payment_path(object, pending_payment.id, e: 'capture')
+    object.payment_required? && pending_payment
```
luisramos0: this code has no unit tests so far.

Matt-Yorkley: There's a quick check that the value is included in the serialised object. Do you think this line deserves it's own test?

luisramos0: I couldnt find api/admin/order_serializer_spec.rb

Matt-Yorkley: Ah, the output of the serializer is tested in: `spec/controllers/api/orders_controller_spec.rb`.

luisramos0: ah, ok.
I think we should have unit tests for the serializers. I dont think it's a requirement for this PR, I just thought of mentioning it.

RachL:
This is awesome as well!! The next release will bring little things, but big improvements in the day to day actions :)

I tested different filtering option and also reloading and everything looks good!

@Matt-Yorkley the loader and the check button are nicely designed :muscle:

Do you know how I can test an error on capturing or shipping?

Matt-Yorkley:
> Do you know how I can test an error on capturing or shipping?

I'm not really sure how to manually trigger those kinds of errors in testing... 

Maybe create an order as admin using a payment method, then remove that payment method from the enterprise's list before hitting the capture button?

RachL:
@Matt-Yorkley ah indeed this is working:

image

 There is no message, but the user will have to contact support. So we can monitor it. Also the current version in production has the same behavior, so this is ready to go!

PR #4585, opened 2019-12-15
Bring spree_backend calculator fields partial to OFN

luisramos0:
#### What? Why?

Related to #4050

Here we bring the code that renders the calculators selector in both the payment and shipping methods in the admin and convert it to haml.

#### What should we test?
We need to test the shipping method and payment method create and edit pages and make sure in both cases the selection of calculators (fees) work correctly. We need to test the forms and then maybe test one simple case where the calculator is applied and make sure what was defined in the form was applied to the order.

#### Release notes
Changelog Category: Changed
Brought the code to render calculators on payment and shipping methods from spree to OFN to make OFN independent of Spree.


lin-d-hop:
Testing Notes:
Created a shipping method for each calculator type
Created a payment method for each calculator type

I then went to create a basket and got a whole lot of ugly:
CalculatorsTest

The page hangs in this way trying to load the whole product list. At this point the usual flow would just show the Cart.

Ping @luisramos0 Not sure if this is specific to this issue or is known.... awaiting your insights :-)

lin-d-hop:
Calculators are working well:
Screenshot from 2020-01-20 12-20-30

PR #4587, opened 2019-12-15
Bring navigation helper from spree_backend

luisramos0:
#### What? Why?

Related to #4050 

Brings the navigation helper from spree.

#### What should we test?
We need to validate the destination pages of all the menus and submenus are correct (some of this is verified by the build):
- main menu:  dashboard, products, ocs, orders, reports, config, enterprises, customers, groups and users
- product submenu: products, properties, inventory and import
- users submenu: usders and roles
- enterprises submenu: enterprises and permissions
- orders submenu: orders, bulk order mgmt and subscriptions

We should verify the icons are as usual on the main items.
We should verify security is working correctly: for example, only super admin sees Users, or user with a "profile only" enterprise will not see the product import tool.

We need to test the submenu on the configurations page, we just need to make sure the links work and that the selected entry is highlighted.

Additionally, there are a couple of HTML helper methods that were also copied, they are used in many places, we should verify a few usages just to be safe:
- edit an order
- delete a payment method
- go back to enterprises list when editing an enterprise

#### Release notes
Changelog Category: Changed
Brought from spree_backend code related to site navigation so that we can make ofn independe of Spree.


luisramos0:
yeah, except for klass_for 🙈

filipefurtad0:
Hey @luisramos0,
The PR has conflicts - could you please have a look?
Cheers :-)

luisramos0:
:+1: rebased to resolve conflicts. ready for testing again.

filipefurtad0:
Hi @luisramos0,

First off - thank you for the detailed testing notes!

Validated the destination pages of all the menus and submenus, admin and superadmin. All seems good.
-> Not even sure this is worth mentioning, but small on the icons as superadmin, which appear a bit out of line:
image

instead of:

image

Permissions and security seem good:
- only super admin sees Users
- a non-producer with a "profile only" enterprise will not see the product import tool.

Configurations Page works fine as Admin.
-> On Superadmin, I noticed the Configuration separator doesnt get highlighted, when selected.

image

The submenu works fine though (both in admin and superadmin), the links work and that the selected entry is highlighted.

-> I noticed the link (as superadmin):
https://staging.katuma.org/admin/payment_methods
Takes you away from the configuration menu, unlike all submenu entries.

All good with HTML helper methods tests:
- editing an order
- delete a payment method
- go back to enterprises list when editing an enterprise

I think despite these small issues - and the PR-unrelated issues, please see below - the PR is good to go. I'll add the feedback label just to double check.


--->>> Issues found not related to this PR:

-> In the main Menu, Orders leads to this link:
https://staging.katuma.org/admin/orders?q%5Bs%5D=completed_at+desc

Do we need the question mark and all that follows? Or just 
https://staging.katuma.org/admin/orders
would do it just as well?

-> Possible bug: editing a previous order worked, provided that there is enough in stock. If not, the edition appears accepted, but does not make any change, and no error is shown.

-> verified the S1 found by @danielle earlier today:
https://github.com/openfoodfoundation/openfoodnetwork/issues/4706

-> Katuma staging does not display correctly the OCM page:
 https://staging.katuma.org/admin/reports/order_cycle_management

luisramos0:
cool, thanks Filipe. Quite a few details to check.

Filipe, about all minor problems, the most important thing about each of them is: did they come with this PR or are they existing? It's useful to check another staging environment and see if it happens. I checked stg FR for the highlighting of Configuration and it is working there. That means this is a bug introduced by this PR. Therefore this is not ready to go.
We can leave this in test ready, I'll fix it tomorrow.

Thanks!!!

filipefurtad0:
Hi @luisramos0,

I missed that one, sorry. Also, I need to get superadmin acess for staging servers other than Katuma.
And I think - for clarity - I should have chosen to add my test notes on a google-doc.

I'll do that later on today.
Thanks again!

luisramos0:
ok, I have fixed the problem with the configurations menu not being selected.
Actually I made it better than it was before, now, whenever you select a submenu in any menu the main menu will still be highlighted. For example, when you go to Products / Search, the Products menu will still be highlighted.
Ready for testing I'd say.

filipefurtad0:
Thanks @luisramos0,

The highlighting works fine for all of them, on the Configurations menu - except when you clich the Mail Method Setting:
https://staging.katuma.org/admin/mail_method/edit

image

Perhaps you forgot about this one? All others work fine.

luisramos0:
ah, I thought I have tested all of them.
No, I have not forgotten it, I included it but it's not working.
It's a pre-existing bug so I'll skip fixing it in this PR.
We can open a separate issue to fix it (the name of the route needs to be changed but we need to check if it would break other things...).

If there's nothing else this would be ready to go. Can you confirm Filipe?
Thanks!

filipefurtad0:
Ok, then I'll open that separate issue too - I was hoping it would be somethig simple :-)

So, from my side that's everything - Moving it to ready to go!
Thank you for your feedback :rocket:

Issue #1215, opened 2016-11-18
Extra content on new payment method error page


RohanM:
When creating a payment method for enterprise, I didn’t select the hub initially to which the payment method was being created.
When I click on create without selecting the Hub, the page refreshes with an error message and prompts to “Atleast one hub must be selected”, but after a few seconds, there is an extra sub page loaded below this page, with the header , openfoodnetwork logo and all the header content and showing the existing payment methods.
I have seen such behavior sometimes in the regular setup too, but yeah could replicate it here.
Not sure if its as per requirement, but the header appearing twice doesn’t sound right !! 
I tried the same with updating the payment method, and found no issue !! 

Source:
https://docs.google.com/document/d/1AL1ZchxAkBHmOO1w_gZA56lSegwXt9OYwPJlOLutMvQ/edit

![ofn-payment-methods-error-loading-below](https://cloud.githubusercontent.com/assets/154210/20413457/42e50e80-ad81-11e6-93c7-2c551a4eaa9e.png)

daniellemoorhead:
@lin-d-hop is this is something worth of the UK Current backlog?

myriamboure:
Just retested and this is still happening.

RachL:
If needed, the test I did is described over there: https://github.com/openfoodfoundation/openfoodnetwork/issues/3225

RachL:
this is now fixed 🎉

PR #4606, opened 2019-12-18
Select by default single Hub/Shop option on creation of payment/shipping method

### LINKED ISSUE #4428: Default selection of Hub/Shop on creation of payment / shipping method when there is only one option
When creating a payment or shipping method, you have to select which hub/hubs that method should be attached to

It is extremely easy to miss and then the user gets lots of red

If there is only one option for selection of Hub / Shop, that should be selected by default, then they don't have to think about it

cyrillefr:
- added a helper
- added mode(new/edit) in payment/shipping views
- updated checkbox creation
- added tests

#### What? Why?

Closes #4428 

Checking the hubs checkbox in creation mode when there is only one would save time as it is mandatory to have at least one.
Then, the related error message should not show up because the checkbox has been forgotten.

#### What should we test?
When creating a new shipping or payment: 
- when one hub, it should be checked by default
- when two or more hubs, there should not be some default check

#### Release notes

Changelog Category:  Changed

### DIFF
--- a/app/helpers/spree/admin/base_helper_decorator.rb
+++ b/app/helpers/spree/admin/base_helper_decorator.rb
@@ -21,6 +21,14 @@ module Spree
 
         link_to_with_icon('icon-trash', name, '#', html_options) + f.hidden_field(:_destroy)
       end
+
+      def add_check_if_single(mode, count)
+        if mode == :new && count == 1
+          { checked: true }
+        else
+          {}
+        end
+      end
     end
   end
 end
--- a/app/helpers/spree/admin/base_helper_decorator.rb
+++ b/app/helpers/spree/admin/base_helper_decorator.rb
@@ -21,14 +21,6 @@ module Spree
 
         link_to_with_icon('icon-trash', name, '#', html_options) + f.hidden_field(:_destroy)
       end
-
-      def add_check_if_single(mode, count)
-        if mode == :new && count == 1
-          { checked: true }
-        else
-          {}
-        end
-      end
     end
   end
 end
--- a/app/views/spree/admin/payment_methods/edit.html.haml
+++ b/app/views/spree/admin/payment_methods/edit.html.haml
@@ -12,7 +12,7 @@
   %fieldset.no-border-top
     = render partial: 'form', locals: { f: f }
     .one.column &nbsp;
-    = render partial: 'spree/admin/shared/hubs_sidebar', locals: { f: f, klass: :payment_method }
+    = render partial: 'spree/admin/shared/hubs_sidebar', locals: { f: f, klass: :payment_method, mode: :edit }
     .clear
     .filter-actions.actions
       = button t(:update), 'icon-refresh'
--- a/app/views/spree/admin/payment_methods/edit.html.haml
+++ b/app/views/spree/admin/payment_methods/edit.html.haml
@@ -12,7 +12,7 @@
   %fieldset.no-border-top
     = render partial: 'form', locals: { f: f }
     .one.column &nbsp;
-    = render partial: 'spree/admin/shared/hubs_sidebar', locals: { f: f, klass: :payment_method, mode: :edit }
+    = render partial: 'spree/admin/shared/hubs_sidebar', locals: { f: f, klass: :payment_method }
     .clear
     .filter-actions.actions
       = button t(:update), 'icon-refresh'
--- a/spec/features/admin/shipping_methods_spec.rb
+++ b/spec/features/admin/shipping_methods_spec.rb
@@ -5,7 +5,7 @@ feature 'shipping methods' do
   include WebHelper
 
   before :each do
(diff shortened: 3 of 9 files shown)


spec/features/admin/shipping_methods_spec.rb
```diff
@@ -65,6 +65,20 @@
       expect(page).to have_content "That shipping method cannot be deleted as it is referenced by an order: #{o.number}."
       expect(Spree::ShippingMethod.find(@sm.id)).not_to be_nil
     end
+
+    scenario "checking a single distributor is checked by default" do
+      d1 = Enterprise.first
```
luisramos0: single letter variables are discouraged although we still have some in the code base. like above o should be order and sm should be shipping_method (you can fix these ones above as well if you want).
in this case it needs to be distributor1 or even first_distributor

spec/features/admin/shipping_methods_spec.rb
```diff
@@ -65,6 +65,20 @@
...
+      visit spree.new_admin_shipping_method_path
+      expect(page).to have_field "shipping_method_distributor_ids_#{d1.id}", checked: true
+    end
+
+    scenario "checking more than a distributor displays no default choice" do
+      d1 = create(:distributor_enterprise, name: 'Aeronautical Adventures')
+      d2 = create(:distributor_enterprise, name: 'Nautical Travels')
```
luisramos0: I wonder what kind of food these enterprises sell :-)

app/views/spree/admin/payment_methods/edit.html.haml
```diff
@@ -12,7 +12,7 @@
   %fieldset.no-border-top
     = render partial: 'form', locals: { f: f }
     .one.column &nbsp;
-    = render partial: 'spree/admin/shared/hubs_sidebar', locals: { f: f, klass: :payment_method }
+    = render partial: 'spree/admin/shared/hubs_sidebar', locals: { f: f, klass: :payment_method, mode: :edit }
```
luisramos0: do you really need mode? if editing and there's only one hub you can force the check as well, even if it is already selected. the code would be simpler.

app/helpers/spree/admin/base_helper_decorator.rb
```diff
@@ -21,6 +21,14 @@ def link_to_remove_fields(name, f, options = {})
...
         link_to_with_icon('icon-trash', name, '#', html_options) + f.hidden_field(:_destroy)
       end
+
+      def add_check_if_single(mode, count)
+        if mode == :new && count == 1
+          { checked: true }
+        else
+          {}
+        end
+      end
```
luisramos0: this is not a good location for this method.
this should probably go into a new helper, I see it's not easy to find a good name for it as it is shared between shipping and payment methods, I'd maybe go for helpers/admin/enterprises_helper.rb

cyrillefr: I have put the helper method in a new helper

app/views/spree/admin/payment_methods/edit.html.haml
cyrillefr: I have though about that too. My mind was split. I removed the mode variable and reverted the updated templates.

spec/features/admin/shipping_methods_spec.rb
cyrillefr: Vegatables that float :)
I have updated the names.

spec/features/admin/shipping_methods_spec.rb
cyrillefr: I have fixed each case I have found in the two testes I have updated.

luisramos0: thank you!!!

spec/features/admin/shipping_methods_spec.rb
luisramos0: :-)

app/views/spree/admin/payment_methods/edit.html.haml
luisramos0: oh, so it just works without the mode thing, definitely better!!! #simplicity

app/helpers/spree/admin/base_helper_decorator.rb
luisramos0: 👍

RachL:
Welcome @cyrillefr and thank you so much for this! It's really gonna help onboarding new enterprises :)

When only one enterprise, it is correctly selected by default:
image

When several enterprises, the user needs to make a choice. We still have the ugly page error on payment methods, but I think there is already an issue about this.

image

moving to ready to go!

sigmundpetersen:
Yes the existing issue is #1215

RachL:
@sigmundpetersen ❤️ thank you. I try to find it but didn't

Issue #4595, opened 2019-12-17
Make OFN independent of spree_backend js, css, fonts and images


luisramos0:
## Description
This is part of #4050 but it may require more work (or more copy paste from spree to OFN than I initially thought). This issue will be used to make that analysis.

We need to move to OFN anything we need from this folder:
https://github.com/openfoodfoundation/spree/tree/2-0-4-stable/backend/app/assets

The biggest challenge is to find out, in the middle of code that is so old, what is actually required for OFN and what is not...

## Acceptance Criteria & Tests
OFN does not depend on any js, css, fonts or images from spree_backend.

# Analysis
Here I list all the dependencies and flag if they are needed in OFN or not. The checkboxes indicate work done so only the items not needed are ticked in the beginning of this work.
My plan is to copy these dependencies to specific folders in OFN:
- js dependencies to app/vendor/javascript
- css dependencies to app/vendor/stylesheets
- js code to app/assets/javascripts/admin/spree (new folder)
- css code to app/assets/stylesheets/admin/spree (new folder)

### js dependencies
Here we analyze generic JS dependencies.

From admin/spree_backend.js:
- [x] #4607 //= require jquery-migrate-1.0.0 - we can remove this as we already have it through gemfile dep in ofn
- [X] //= require jquery.ui.all - already in OFN, can be ignored
- [x] #4607 //= require modernizr - let's keep it to keep compatibility with very old browsers
- [x] #4611 //= require jquery.cookie - this is used by jstree and jstree is required for taxons admin
- [X] //= require jquery.delayedobserver - this is dead code in spree_backend, can be ignored
- [x] #4611 //= require jquery.jstree/jquery.jstree - this is needed for taxons admin
- [x] #4611 //= require jquery.alerts/jquery.alerts - one usage of jConfirm to confirm a delete of a taxonomy - we should make it a normal confirm and do not copy the dependency to OFN
- [x] #4608 //= require jquery.powertip - this is used to render tooltips in ofn, needs to be copied
- [x] #4614 //= require jquery.vAlign - used to align checkboxes - keep.
- [x] #4607 //= require css_browser_selector_dev - this dep eases the implementation of browser specific tweaks in stylesheets/admin/hacks
- [x] #4614 //= require spin - there are very few situations where this is used - probably easier to keep, we could remove...
- [X] //= require trunk8 - this is dead code in spree_backend - can be ignored
- [X] //= require jquery.adaptivemenu - already in OFN
[...]

PR #4611, opened 2019-12-18
Bring js code to make taxonomies admin work without spree_backend

luisramos0:
#### What? Why?

Continues to #4595

We need to bring jstree from spree_backend so that the taxonomies admin page works in the backoffice. We remove one usage of jConfirm to avoid having to bring jquery.alert as well.

We create a folder admin/spree where we will put the not angularized JS legacy code coming from spree_backend. It's not much code but I think it's better if we keep it isolated under admin/spree/.

#### What should we test?
The taxons admin page in the backoffice is working correctly, particularly setting up hierarchies and renaming and deleting elements.

#### Release notes
Changelog Category: Changed
Brought from spree_backend code to make taxons admin page work without spree code so that we make ofn independent of spree.


luisramos0:
I rebased and tested deploying. All good. Ready for testing.

lin-d-hop:
ping @luisramos0 #conflicts

luisramos0:
rebased :+1:

lin-d-hop:
Tested creating a taxon and associating with a product. 
Taxons can be deleted when not associated with a product and cannot be deleted afterwards.
Taxons in different hierarchies are squashed into a flat list.
Renaming works.
All changes flow correctly into shopfront and descriptions.
All behaviour is as the current taxonomy system on production.

Ready to go :-)

PR #4613, opened 2019-12-19
Bring remaining variants autocomplete (and related) JS code from spree_backend

luisramos0:
#### What? Why?

Continues #4595

Bring variant_autocomplete as well as shipments js code to make orders edit page work. We also bring address_states used in the customer details page in the order edit page.

To support this we need to require admin/handlebar_extensions to make translations work inside handlbars templates.
Here spree uses underscore js to call functions like _.find, instead of bringing underscore we require existing (used in the frontoffice) lodash.underscore.js

#### What should we test?
Order edit page and customer details page particularly the auto complete features in them: search variants and search customers.

#### Release notes
Changelog Category: Changed
Bring variant autocomplete code from spree_backend to make ofn independent of Spree.


app/assets/javascripts/admin/spree/orders/shipments.js.erb
```diff
@@ -0,0 +1,64 @@
...
+      url: url
+    }).done(function( msg ) {
+      window.location.reload();
+    }).error(function( msg ) {
+      console.log(msg);
+    });
+  }
+  // $('[data-hook=admin_order_edit_form] a.ship').click(handle_ship_click);
+
+  //handle shipping method edit click
```
Matt-Yorkley: Are all these commented out lines in the originals files as well?

luisramos0: yes, these events are registered in spree_backend. I will remove spree_backend and these comments in https://github.com/openfoodfoundation/openfoodnetwork/pull/4621/files

luisramos0:
Rebased and tested deploying. All good. Ready for testing.

lin-d-hop:
In Edit Products and Customers
Tested a range of combinations - start, mid, end - of name/email/code all relevant fields for the search.
Everything worked a treat.
Ready to go.

luisramos0:
great, thanks for testing!!!

PR #4616, opened 2019-12-19
Bring some JS and CSS code from spree_backend related to orders admin

luisramos0:
#### What? Why?

Continues #4595 

#### What should we test?
- Verify the edit product page has the usual style and positioning of elements.
- Verify you can add, edit and remove images from a product
- Verify the orders search page and the order edit page have the usual style and positioning of elements.
- In the list of payments of an order, make sure the outstanding balance tag is in the correct place with correct style

#### Release notes
Changelog Category: Changed
Bring javascript code related to orders from spree_backend so we can make OFN independent of Spree.


luisramos0:
I rebased and added a dependency needed (admin/globals/variables.scss which is already added on another PR)  to make the precompilation process work here.
Ready for testing.

lin-d-hop:
This is looking good.
Tested all steps and compared to UK production. Everything is exactly as it should be.
Ready to go!

lin-d-hop:
@luisramos0 except now there is a conflict

luisramos0:
it's a tiny conflict in the all css file, I dont think this can create problems, I will merge now.

PR #4617, opened 2019-12-19
Bring some js and css code from spree_backend needs in the general configurations

luisramos0:
#### What? Why?
Continues #4595

#### What should we test?
**Image settings:**
- no need to test the actual working of the setting, just the form
- verify you can add a new style and remove it and also that you can add a new s3 header (removing a s3 header is broken, we can open a new issue if we want to fix this)

**States:**
- Editing the list of states of a country, when you change from one country to another, the link to create a new state should be updated. Add a state to a country and then change to another country and try to add a state to that new country. The second state should be added to the second country.

**Calculator Fields:**
- Edit a shipping method and make sure that when you change the calculator, you can see the warning: "If you are changing the calculator type, you must save first before you can edit the calculator settings"

**Zones:**
- Verify the edit zone page and make sure you can switch between country based to state based and the panel on the right switches accordingly and works correctly.

#### Release notes
Changelog Category: Changed

Bring code from spree_backend related to general configurations so that we can make OFN independent of Spree.

### DIFF
--- a/app/assets/javascripts/admin/spree/calculator.js
+++ b/app/assets/javascripts/admin/spree/calculator.js
@@ -1,0 +1,16 @@
+$(function() {
+  var calculator_select = $('select#calc_type')
+  var original_calc_type = calculator_select.attr('value');
+  $('.calculator-settings-warning').hide();
+  calculator_select.change(function() {
+    if (calculator_select.attr('value') == original_calc_type) {
+      $('div.calculator-settings').show();
+      $('.calculator-settings-warning').hide();
+      $('.calculator-settings').find('input,textarea').prop("disabled", false);
+    } else {
+      $('div.calculator-settings').hide();
+      $('.calculator-settings-warning').show();
+      $('.calculator-settings').find('input,texttarea').prop("disabled", true);
+    }
+  });
+})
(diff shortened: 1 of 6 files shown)


luisramos0:
rebased and tested deployment process. all good, ready for testing.

lin-d-hop:
Image settings. The form appears to work. When I click new the form offers me 2 new options and I have to remove one to save.

Screenshot from 2020-01-31 11-45-51

Screenshot from 2020-01-31 11-47-09


States - adding states to different countries persists as expected

Calculators - warning shows correctly in Shipping and Payment methods

Zones - Works correctly but when I go to add a new zone I am shown 2 additional zones fields instead of one. Same problem as above.

Screenshot from 2020-01-31 11-45-30

As this double up of fields is happening in multiple places I fear that we might find this popping up all over which will create an additional support demand for new instances as we have to explain that to save you must delete on of the additional fields.

Not sure if this will also flow through to user facing areas.

So despite it being small and easily worked around I suggest we just fix this here and now.

Back to In Dev.

luisramos0:
ah, the code is duplicated between spree and OFN now, that's why we see two entries...
This will go away when we remove spree_backend 

"we might find this popping up all over " - we are only changing these pages, nothing else. There's no risk of seeing this in other places.

Agree, we shouldnt put this in produiction.
But I think I'll leave this for when I remove spree_backend, that way this will just work as is.

luisramos0:
Closing, this is going in as part of #4621

PR #4618, opened 2019-12-20
Load DB2Fog configuration from file if it exists

mkllnk:
#### What? Why?

This pull request is a request for comments. I would like to configure backups independently to
- avoid moving the AU backups bucket,
- enable France to start a new backups bucket in Europe more easily,
- have the freedom to choose non-Amazon backups and
- avoid replicating DB2Fog's config interface.

What do you think about the following approach? Do you have a better idea?

This enables us to define a custom configuration file in ofn-install
(not yet implemented yet).

At the moment, the AWS S3 configuration for assets and backups are tightly coupled. This is an issue when buckets have different access keys or are in different regions. Since the Australian server was setup with backups in a different region to the assets, we need a way to configure this or move the bucket.

#### What should we test?

Dev test:

- Use a staging server with backups configured.
- Copy the example file to create a db2fog configuration.
- Run: `RAILS_ENV=staging bundle exec rake db2fog:backup`
- Verify that the backup has been uploaded.

#### Release notes

Added: The backup location can now be configured independently. Backups can be stored in another country and with different access keys than images.

Changelog Category: Added

Backups don't depend on the Spree configuration anymore.

#### Dependencies

This will require a change in ofn-install to be fully automated.

#### Documentation updates

Documentation updates should be done with the ofn-install pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ public/stylesheets
 public/images
 public/spree
 config/abr.yml
+config/db2fog.yml
 config/initializers/feature_toggle.rb
 NERD_tree*
 coverage
--- a/config/db2fog.yml.example
+++ b/config/db2fog.yml.example
@@ -1,0 +1,6 @@
+---
+:aws_access_key_id: "AYAYAYAYAAAAAAAAA"
+:aws_secret_access_key: "ThisIsAnExampleKeyWhichIsAtLeastFourtyCharsLong"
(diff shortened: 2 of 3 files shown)


mkllnk:
I'm closing this in favour of https://github.com/openfoodfoundation/openfoodnetwork/pull/4653. Once we can configure the region, the need for a custom config is not urgent.

PR #4620, opened 2019-12-20
Bring basic CSS rules from spree backend

luisramos0:
#### What? Why?

Continues #4595

Bring basic css from spree_backend.

#### What should we test?
I think we should compare visually some of the main pages of the backoffice to an instance without this PR and try to verify that the pages look the same in terms of style: size of components, position of components and color of components.

#### Release notes
Changelog Category: Changed
Brought all the basic css rules from spree_backend so that we can make ofn independent of Spree.


luisramos0:
yeah, I considered that option. I kept the spree JS code in a separate folder under admin/spree because it's not angularized, very different from ours. But in CSS from now on I thought we should probably just mix everything and start changing the rules in these files if needed and merge these files with our files if we have the opportunity. Like merging admin/variables and new admin/globals/variables.

luisramos0:
I just rebased and tested the deployment process (I had to add some imports in the css files to make the deployment work. All good now.
Ready for testing.

lin-d-hop:
Compared FR staging to UK production
Checked top and second level menu pages.
Checked all Edit sections (order, product)
Checked Enterprise settings
Everything looks as pretty as ever.
Good to go!

lin-d-hop:
@luisramos0 conflicts

luisramos0:
Some conflicts related to the order the css files are imported.
I think we should have a quick look at this one again.
Just making sure no major problems appear, for example, in the list of orders in admin/orders.

luisramos0:
I rebased again. I think we can just deploy and verify that are no display errors in the orders list page.

filipefurtad0:
Hey @luisramos0,

Staged this PR in FR and compared it with AU in production.
All seems the same, so moving to ready to go!
Thanks!

PR #4576, opened 2019-12-13
Updating mini_racer and libv8 at the same time

luisramos0:
#### What? Why?

Closes #[the issue number this PR is related to]

#### Release notes

Changelog Category: Added | Changed | Deprecated | Removed | Fixed | Security

PR #4627, opened 2019-12-26
Update libv8 and keep it only in Gemfile.lock as a dependency of mini_racer

luisramos0:
#### What? Why?

I tried doing this in #4576 but failed. I dont understand why.

Anyway, instead of upgrading libv8 in the gemfile #4624, we can now move it back to gemfile lock only, it's not a direct dependency, it's only used by mini_racer.

#### What should we test?
We need to stage this and make sure assets compilation works correctly and the website renders correctly.

#### Release notes
Changelog Category: Changed
Upgraded libv8 dependency.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,9 +107,6 @@ group :assets do
   gem 'compass-rails'
 
   gem 'mini_racer', '0.2.4'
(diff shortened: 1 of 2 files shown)


RachL:
I think this is good to go!

PR #4630, opened 2019-12-27
Show counts including all variants in order cycle exchanges

### LINKED ISSUE #4594: Change OC outgoing pagination count to variants
## Description
**- As a:** Manager and super admin
**- On page:** `/admin/order_cycles/XXX/outgoing`
**- I want to be able to :** see a count in number of _variants_ not in number of _products_ when I unfold the outgoing products

*E.g. the total number 6387 in the screenshot above is the number we should see everywhere. 6204 is the number of products*

*Scenario 1*
1. Logged in as a single producer shop
2.  I can create, update, read and delete an OC

*Scenario 2*
1. Logged in as a single hub shop
2.  I can create, update, read and delete an OC
3. When editing or creating the outgoing products step, all the numbers I see should count the number of variants in the system. 

*Scenario 3*
1. Same as scenario 2 but logged in as a multiple distributor shop

*Scenario 4*
1. Same as scenario 2 but logged in as super admin

Matt-Yorkley:
#### What? Why?

Closes #4594 

Adapts order cycle exchange panels to show full count of all variants.

#### What should we test?

Editing an order cycle: exchanges show full count of all variants.

#### Release notes

Changed pagination counts in order cycle exchange UI

Changelog Category: Changed

all_variants

### DIFF
--- a/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
@@ -58,9 +58,8 @@ angular.module('admin.orderCycles')
 
       incoming = true if $scope.view == 'incoming'
       params = { exchange_id: exchange.id, enterprise_id: exchange.enterprise_id, order_cycle_id: $scope.order_cycle.id, incoming: incoming, page: page}
-      ExchangeProduct.index params, (products, num_of_pages, num_of_products) ->
+      ExchangeProduct.index params, (products, num_of_pages) ->
         enterprise.num_of_pages = num_of_pages
-        enterprise.num_of_products = num_of_products
         enterprise.supplied_products.push products...
 
     $scope.loadMoreExchangeProducts = (exchange) ->
--- a/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
@@ -8,13 +8,6 @@ angular.module('admin.orderCycles')
     $scope.productsLoading = ->
       RequestMonitor.loading
 
-    $scope.exchangeLoadedVariants = (exchange) ->
-      loaded_variants = 0
-      angular.forEach $scope.enterprises[exchange.enterprise_id].supplied_products, (product) ->
-        loaded_variants += product.variants.length
-
-      loaded_variants
-
     $scope.setSelectAllVariantsCheckboxValue = (exchange, totalNumberOfVariants) ->
       exchange.select_all_variants = $scope.exchangeSelectedVariants(exchange) >= totalNumberOfVariants
 
@@ -55,12 +48,15 @@ angular.module('admin.orderCycles')
 
       return if enterprise.last_page_loaded? && enterprise.last_page_loaded >= page
       enterprise.last_page_loaded = page
+      enterprise.loaded_variants ?= 0
 
       incoming = true if $scope.view == 'incoming'
       params = { exchange_id: exchange.id, enterprise_id: exchange.enterprise_id, order_cycle_id: $scope.order_cycle.id, incoming: incoming, page: page}
       ExchangeProduct.index params, (products, num_of_pages) ->
         enterprise.num_of_pages = num_of_pages
         enterprise.supplied_products.push products...
+        angular.forEach products, (product) ->
+          enterprise.loaded_variants += product.variants.length
 
     $scope.loadMoreExchangeProducts = (exchange) ->
       $scope.loadExchangeProducts(exchange, $scope.enterprises[exchange.enterprise_id].last_page_loaded + 1)
--- a/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
@@ -8,6 +8,13 @@ angular.module('admin.orderCycles')
     $scope.productsLoading = ->
       RequestMonitor.loading
 
+    $scope.exchangeLoadedVariants = (exchange) ->
+      loaded_variants = 0
+      angular.forEach $scope.enterprises[exchange.enterprise_id].supplied_products, (product) ->
+        loaded_variants += product.variants.length
+
+      loaded_variants
+
     $scope.setSelectAllVariantsCheckboxValue = (exchange, totalNumberOfVariants) ->
       exchange.select_all_variants = $scope.exchangeSelectedVariants(exchange) >= totalNumberOfVariants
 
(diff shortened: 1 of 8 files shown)


luisramos0:
ah, one more point: I think we should change to "39 of 49 **Variants** Loaded" to be more clear.

app/assets/javascripts/admin/order_cycles/controllers/order_cycle_exchanges_controller.js.coffee
```diff
@@ -51,9 +58,8 @@ angular.module('admin.orderCycles')
 
       incoming = true if $scope.view == 'incoming'
       params = { exchange_id: exchange.id, enterprise_id: exchange.enterprise_id, order_cycle_id: $scope.order_cycle.id, incoming: incoming, page: page}
-      ExchangeProduct.index params, (products, num_of_pages, num_of_products) ->
+      ExchangeProduct.index params, (products, num_of_pages) ->
         enterprise.num_of_pages = num_of_pages
-        enterprise.num_of_products = num_of_products
```
luisramos0: instead of an extra method above "exchangeLoadedVariants" I'd implement the logic here in loadExchangeProducts.
instead of enterprise.num_of_products = num_of_products
we can do:

initialize enterprise.loaded_variants = 0 before the call to  ExchangeProduct.index
`enterprise.loaded_variants = 0`

and then inside:
```
angular.forEach products, (product) ->
        enterprise.loaded_variants += product.variants.length
```

Matt-Yorkley: I can't tell if there's any real benefit in one over the other? Do you think it would be more efficient to increment the count as the variants are loaded?

luisramos0: I think it would be slightly faster and less complex with one less method. but I dont think it's a massive difference, that's why I approve the PR as is. it's your call :-)
my first comment in 2020 happy new decade!

mkllnk: Hm, maybe there is not much in it but I imagine Luis' version to be more efficient. The variants are counted once after they have been loaded. In the current version, they are calculated every time `exchangeLoadedVariants` is called. I don't know how often Angular executes that.

And since we are already talking about preferred styles, I have another version here: :wink: 
```coffee
enterprise.loaded_variants = products.reduce((sum, product) => sum + product.variants.length, 0))
```

luisramos0: :-)

luisramos0:
yesh, thanks @mkllnk 

@RachL shall I change from "Products Loaded" to "Variants Loaded"?

RachL:
@luisramos0 yes maybe that's best. Worse case we can change in translations :)

Matt-Yorkley:
@mkllnk and @luisramos0 

I've renamed "Products" to "Variants", updated the `loaded_variants` counting methods (second suggestion didn't work), and found and fixed a pre-existing javascript console error related to clashing `$digest` cycles.

app/assets/javascripts/admin/order_cycles/controllers/incoming_controller.js.coffee
```diff
@@ -28,4 +28,3 @@ angular.module('admin.orderCycles').controller 'AdminOrderCycleIncomingCtrl', ($
   $scope.selectAllVariants = (exchange, selected) ->
     $scope.loadAllExchangeProducts(exchange).then ->
       $scope.setExchangeVariants(exchange, $scope.suppliedVariants(exchange.enterprise_id), selected)
-      $scope.$apply()
```
luisramos0: was this causing the console error? this is needed for some cases... the case is:
- on the incoming exchanges when you have only a few pages loaded and you "select all", it will load all variants first and then select them all. in this case, this apply() is making the "select all" select box become selected, otherwise it wont re-render after all variants are loaded and selected and it will remain unchecked (which is incorrect).

maybe we need to run this apply only in this case, which is the case where loadAllExchangeProducts actually loads something. I'd guess the console error is for the case where loadAllExchangeProducts is not loading anything (all products are already loaded)

Matt-Yorkley: Maybe I'll just revert it and open a tech-debt issue for now :+1:

Matt-Yorkley: Issue: #4676

lin-d-hop:
On testing this I noticed a couple things:

1) This doesn't show for single producer shops. There is no count at all. I assume this is out of scope for this issue.

2) A few weird cases testing on UK staging with Cambridge Food Hub
VariantCount

The values are different in the Tab and in the Panel (when you click to open the tab. 160 vs 163.
Consistently is says x/0 selected. Surely the total should be 163, not 0.

Ping @matt-yorkley.

luisramos0:
Looks like the problem you are seeing is unrelated to this PR...
I dont understand why it only fails here, but this change will fix the problem:

```diff
--- a/app/controllers/api/exchange_products_controller.rb
+++ b/app/controllers/api/exchange_products_controller.rb
@@ -31,7 +31,7 @@ module Api
       render text: {
         count: Spree::Variant.
           not_master.
-          where(product_id: products).
+          where(product_id: products.select(&:id)).
           count
       }.to_json
     end
```

luisramos0:
It's not just this problem, there's something else going on. I copied uk staging data to my computer and I am having a look. This enterprise Cambridge FoodHub looks like a really good test case, Ill investigate further :+1:

luisramos0:
This test case is very good!
I found another bug in the counting of variants. The simple count of variants (to display the summary before any tab is opened) is not considering order_cycle config that sets "coordinator inventory only". For example in Hodmedods MASTER there are 46 inventory items that are hidden and those variants are considered in the counting but are not displayed...

This problem is closely related to this PR but it is pre-existing. But this PR makes it worse because the numbers before this PR were variants against products (not obvious to see the problem). Now it's variants against variants and the issue is obvious!
I think I'll create another PR to fix the problem but we should mark this blocked and only merge this when the new one I'll create is also ready to be merged.

Let me know if you dont understand something I said above.

luisramos0:
Another problem I see is that the pagiantion is done on the products but we display variants.
Each page is 100 products but that is displayed as variants, so we get random page sizes like here:

Screenshot 2020-01-24 at 10 18 44

luisramos0:
ok, #4718 is ready for code review.

luisramos0:
Now that 4718 is merged, we should rebase this and test it again 💪

Matt-Yorkley:
Rebased :+1:

lin-d-hop:
@Matt-Yorkley  more conflicts :-(

Matt-Yorkley:
Conflict resolved :+1:

filipefurtad0:
Hey @Matt-Yorkley,

This looks great. Below - left, before - and - right, after:

image

One now sees all the variants.

I think this is ready to go.

@lin-d-hop : You have tested this before, would you like to have a look still?

RachL:
I think this is good to go, no need for @lin-d-hop to review it :-)

PR #4631, opened 2019-12-27
Increase space for buttons in savebar layout

### LINKED ISSUE #4503: New OC bottom page buttons do not fit well with long text
## Description

When translating the button in French for example, the text is longer, leaving the buttons to display like this:

French:

Spanish:

Instead of:

## Steps to Reproduce

Make sure to check a non English instance.

1. Go to `/admin/order_cycles/`
2. Create or edit an OC
3. See bottom buttons

## Severity

## Your Environment
* Version used: v2.6.4

Matt-Yorkley:
#### What? Why?

Closes #4503 

Increases space for buttons on the right hand side of the admin savebar to accommodate for different sizes due to translations.

#### What should we test?

Save bar buttons on the OC edit page are rendering ok in different languages.

#### Release notes

Fixed layout issue with save bar buttons in different languages

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/templates/admin/save_bar.html.haml
+++ b/app/assets/javascripts/templates/admin/save_bar.html.haml
@@ -1,9 +1,9 @@
 #save-bar.animate-show{ ng: { show: 'dirty || persist || StatusMessage.active()' } }
   .container
-    .eight.columns.alpha
+    .seven.columns.alpha
       %h5#status-message{ ng: { show: "StatusMessage.invalidMessage == ''", style: 'StatusMessage.statusMessage.style' } }
         {{ StatusMessage.statusMessage.text || "&nbsp;" }}
       %h5#status-message{ ng: { show: "StatusMessage.invalidMessage !== ''" }, style: 'color: #da5354' }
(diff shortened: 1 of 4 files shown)


Matt-Yorkley:
Ah, it works in Spanish, but not French. Do we need all these buttons crammed into this space? The savebar was designed to have two buttons, and now it has five.

I tried adding some padding and keeping them on multiple lines but it doesn't look good...

luisramos0:
yeah, do you have a suggestion @Matt-Yorkley ? I think the "previous" button is the least useful. we could drop it now that the steps are clickable above...
could you paste a screenshot with the 5 buttons in French spanning 2 lines with padding?

@RachL could we have shorter French translations? Maybe just "retour" on the last one?

Matt-Yorkley:
Lets drop the "previous" button for now. I think that's the best option here. :+1:

RachL:
FR:

image

IT:
image

I didn't tested in Spanish or Portuguese because it is actually not translated yet. Ping @sauloperez @luisramos0 maybe you will want to have a look some time on Transifex.

Looking good in FR and IT. Moving to ready to go! Let's see how user react to the previous button being gone.

PR #4633, opened 2019-12-28
Fix bug in frontoffice order details where the product image was a broken link

### LINKED ISSUE #4632: Product image link is broken in orders detail in the frontoffice
## Description
When a customer clicks the product image in the order details page (this page is the checkout confirmation page and can also be accessed through account/orders/specific order) the customer is sent to the app homepage...

## Expected Behavior
Because we dont have a PDP, the easier solution is to remove the link and make the image non-clickable.

## Actual Behaviour
Customer is sent to homepage.

## Steps to Reproduce
Checkout an order and click the image of the product in the order confirmation page.

## Severity

## Your Environment
AUS live for example.

luisramos0:
#### What? Why?

Closes #4632

This PR make the product image non-clickable.
This is an very low priority bug but it was blocking progress of #4050 so I fixed it now.

#### What should we test?
Make sure the image is not clickable.

#### Release notes
Changelog Category: Fixed
Fixed a bug in order details page by making product images non-clickable (the link was broken).

### DIFF
--- a/app/views/spree/orders/_bought.html.haml
+++ b/app/views/spree/orders/_bought.html.haml
@@ -17,9 +17,9 @@
 
         %div.item-thumb-image
           - if variant.images.length == 0
-            = link_to mini_image(variant.product), variant.product
+            = mini_image(variant.product)
           - else
-            = link_to image_tag(variant.images.first.attachment.url(:mini)), variant.product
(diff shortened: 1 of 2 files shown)


lin-d-hop:
Looks good to me :-)

PR #4638, opened 2020-01-02
Replace adjustment.open with adjustment.fire_events(open) to avoid method name conflict

### LINKED ISSUE #4581: Cannot edit cart - snail
## Description
Editing quantity in cart is producing a snail on OFN_CAN.  But if I use the browser back arrow at the snail to go back to the cart -  the cart edit is made.   

## Expected Behavior
No snail

## Actual Behaviour
Snail on cart edit -  but if you use the browzer arrow to go back - the change is there
## Steps to Reproduce
1.go to a shop on OFN-CAN
2.put an item in your cart and try to edit
3. get the snail and use your browzer arrows to go back
4. see the change is made

## Workaround
work around is using the browzer back arrow - but will users know to do this?

## Severity

bug-s1

-->

## Your Environment

* Version used:
* Browser name and version:

luisramos0:
#### What? Why?

Closes #4581

We make this change although we cant replicate the problem. We believe this will fix the conflict and therefore the bug.

#### What should we test?
Try to replicate the problem. Probably only verifiable in OFN_CAN live where the problem exists.

#### Release notes
Changelog Category: Fix
Fixed a bug when editing quantities in the cart.

### DIFF
--- a/app/controllers/spree/orders_controller.rb
+++ b/app/controllers/spree/orders_controller.rb
@@ -176,7 +176,7 @@ module Spree
       previous_states = @order.adjustments.each_with_object({}) do |adjustment, hash|
         hash[adjustment.id] = adjustment.state
       end
-      @order.adjustments.each(&:open)
+      @order.adjustments.each { |adjustment| adjustment.fire_events(:open) }
 
       yield
 


RachL:
I tested:
- Editing quantity
- Deleting product
- Empyting cart

All are working good. I'm moving this to ready to go. If the problem still happens in production, we can reopen.

PR #4641, opened 2020-01-05
3.0 Initial fixes

Matt-Yorkley:
Closes #3706 (routes are now working correctly)

This PR is targetting the `3-0-stable` branch.

Introduces a number of fixes to get the app booting and the build working better with Spree 2.1 / Rails 4.

PR #4643, opened 2020-01-05
3.0 Strong Parameters

Matt-Yorkley:
Related to #4641

Starts addressing #4644 

Starts the move towards strong parameters in OFN, and fixes various tests in the `3.0` build. 

Adds some temporary methods to print a list of the related parameters in Semaphore when a test fails due to forbidden params, to help with debugging.


Matt-Yorkley:
I thought it might be useful to keep the attributes visible in each model until the strong params work is finished. 5ffc5fc and 265afdf should both be tidied up at the end.

luisramos0:
ah, I completely agree with Maikel. I had that exact same comment going through my mind but I ignored it :-)

Matt-Yorkley:
I thought that at first, but after digging in to it I think strong params is the way to go. Partly because Spree is already using it, and partly because the protected_attributes gem actually introduces other issues (and spec failures) that at would need to be fixed in order to _not_ use strong params.

luisramos0:
I am ok with that if you have an idea of the size of the tasks:
A - use protected_attributes gem and fix the problems
B - implement strong parameters in all of OFN 

is A not just a couple of hours and B quite a few days of dev?
I am afraid that implementing (and testing) B could take a while.

mkllnk:
Is there a way in between? Can we keep protected attributes and move to strong parameters where the protected attributes cause trouble?

I agree that strong parameters are the way to go. I just hoped that we can deal with it later.

mkllnk:
Or here is another idea, we could add strong parameters to master to prepare for the upgrade: https://stackoverflow.com/questions/13852534/strong-parameters-in-rails-3-2-8

Matt-Yorkley:
> we could add strong parameters to master to prepare for the upgrade

It's a nice idea, but we would hit the issue that Spree 2.1 has it's strong parameters implementation included and fully integrated into all it's controllers, so to do it in master we would probably have to include some kind of hacky backport into our current Spree version...

mkllnk:
But didn't we import most of Spree's controllers so that the upgrade doesn't help us here and we still have to do it ourselves?

I'm not sure if it's a good idea to have this discussion in asynchronous text. Maybe let's organise a dev call.

Matt-Yorkley:
Yeah, I think we need to merge all the spree back end work into master, merge the other bits in 3-0-stable, then merge master into 3-0-stable, then take a look and come up with a solid plan.

luisramos0:
We need to rebase 30 and create issues for each temp fix that is there.

PR #4651, opened 2020-01-07
Remove unused presenter

mkllnk:
#### What? Why?

I just came across this dead code. The usage of this presenter got removed seven years ago:
c5d031a6d7e18ca0d9c8e15fc26be392afdd5ed7

#### What should we test?

Create an order cycle

#### Release notes

Removed unused code.

Changelog Category: Removed


daniellemoorhead:
Testing: I successfully cloned an order cycle on Katuma staging, and was able to access that order cycle on the shop.

Screen Shot 2020-01-26 at 4 55 21 pm

Screen Shot 2020-01-26 at 4 56 21 pm

I wasn't able to check out as UK staging is not working for checkouts due to a payment method not being set somewhere. 

@mklink I'm not sure if it's enough to have cloned the OC and also not checked out? I attempted to create a new one from scratch but couldn't make it display and I think that's more my fault than the fault of this PR. If I need to create one from scratch (I probably do, right, to be 100% sure all is ok?) then I'll properly learn how to create the order cycle using the userguide (instead of trying to do it on the fly between looking after kids on a Sunday :-) ). Let me know and I shall proceed as you direct!

mkllnk:
@daniellemoorhead The checkout is not important for this PR. The important bit is the creation of the order cycle, especially where you select products to be displayed in the OC. It would be great if you tested that part.

An OC not displaying is usually due to a misconfigured shipping or payment method.

daniellemoorhead:
Ok @mkllnk I'll give it a go on.

daniellemoorhead:
Testing notes:
Mac/Chrome

I created a new order cycle for a shop on UK staging, and was able to see that order cycle in the shop.
Screen Shot 2020-01-28 at 8 09 42 pm

Screen Shot 2020-01-28 at 8 09 14 pm

I also cloned an OC and it worked as well.

PR #4656, opened 2020-01-08
Use strings for all environment variables

mkllnk:
#### What? Why?

When using a fresh setup with the example application config, Figaro complained:

    WARNING: Use strings for Figaro configuration. 25 was converted to "25".
    WARNING: Use strings for Figaro configuration. 4 was converted to "4".

Those numbers have been converted to strings in our example config.

#### What should we test?

No test, this doesn't affect any production code.

#### Release notes

Fixed syntax of an example config file.

Changelog Category: Fixed


luisramos0:
this will break emails if it fails. In case the example is followed and the change is made in the secrets file. right?

PR #4657, opened 2020-01-08
Remove Product import code for xls suppport

luisramos0:
#### What? Why?
PI only works with csv right now: see [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/4625#issuecomment-569228696)

This can be reverted if we start supporting xls and ods in the future

btw, roo-xls is only needed to support legacy xls types, roo itself supports xlsx.

Relates to #4625

#### What should we test?
Verify csv upload works as normal. A simple test should be enough.

#### Release notes
Changelog Category: Removed
Removed code to support xls file upload in product import. We dont support xls uploads.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,7 +94,6 @@ gem 'wkhtmltopdf-binary'
 gem 'foreigner'
 gem 'immigrant'
 gem 'roo', '~> 2.8.2'
-gem 'roo-xls', '~> 1.1.0'
 
 gem 'whenever', require: false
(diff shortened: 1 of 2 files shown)


luisramos0:
rebased to resolve conflicts

luisramos0:
rebased to resolved conflicts.

lin-d-hop:
Gave a simple release test and products upload correctly.
Ready to go.

PR #4667, opened 2020-01-12
Rails 4 prep

Matt-Yorkley:
#### What? Why?

Pulls across some commits for upgrading to Rails 4. Mostly minor changes in syntax.

#### What should we test?

I think a green build will be enough to validate these changes.

#### Release notes

Adapted some syntax in preparation for Rails 4.

Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/base_controller.rb
+++ b/app/controllers/spree/admin/base_controller.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
 module Spree
   module Admin
     class BaseController < Spree::BaseController
@@ -121,11 +122,15 @@ module Spree
 
       def render_as_json(data, options = {})
         ams_prefix = options.delete :ams_prefix
-        if [Array, ActiveRecord::Relation].include? data.class
+        if each_serializer_required?(data)
           render options.merge(json: data, each_serializer: serializer(ams_prefix))
         else
           render options.merge(json: data, serializer: serializer(ams_prefix))
         end
+      end
+
+      def each_serializer_required?(data)
+        ['Array', 'ActiveRecord::Relation'].include?(data.class.name)
       end
 
       def serializer(ams_prefix)
@@ -140,3 +145,4 @@ module Spree
     end
   end
 end
+# rubocop:enable Metrics/ClassLength
--- a/app/controllers/spree/admin/base_controller.rb
+++ b/app/controllers/spree/admin/base_controller.rb
@@ -97,7 +97,7 @@ module Spree
       def active_distributors_not_ready_for_checkout
         ocs = OrderCycle.managed_by(spree_current_user).active
         distributors = ocs.includes(:distributors).map(&:distributors).flatten.uniq
-        Enterprise.where('enterprises.id IN (?)', distributors).not_ready_for_checkout
+        Enterprise.where(id: distributors.map(&:id)).not_ready_for_checkout
       end
 
       def active_distributors_not_ready_for_checkout_message(distributors)
--- a/app/controllers/spree/admin/base_controller.rb
+++ b/app/controllers/spree/admin/base_controller.rb
@@ -1,4 +1,3 @@
-# rubocop:disable Metrics/ClassLength
 module Spree
   module Admin
     class BaseController < Spree::BaseController
@@ -16,11 +15,10 @@ module Spree
       # Warn the user when they have an active order cycle with hubs that are not ready
       # for checkout (ie. does not have valid shipping and payment methods).
       def warn_invalid_order_cycles
-        distributors = active_distributors_not_ready_for_checkout
+        return if flash[:notice].present?
 
-        return if distributors.empty? || flash[:notice].present?
-
-        flash[:notice] = active_distributors_not_ready_for_checkout_message(distributors)
+        warning = OrderCycleWarning.new(spree_current_user).call
+        flash[:notice] = warning if warning.present?
       end
 
       # This is in Spree::Core::ControllerHelpers::Auth
@@ -94,24 +92,6 @@ module Spree
 
       private
 
-      def active_distributors_not_ready_for_checkout
-        ocs = OrderCycle.managed_by(spree_current_user).active
-        distributors = ocs.includes(:distributors).map(&:distributors).flatten.uniq
-        Enterprise.where(id: distributors.map(&:id)).not_ready_for_checkout
-      end
-
-      def active_distributors_not_ready_for_checkout_message(distributors)
-        distributor_names = distributors.map(&:name).join ', '
-
-        if distributors.count > 1
-          I18n.t(:active_distributors_not_ready_for_checkout_message_plural,
-                 distributor_names: distributor_names)
-        else
-          I18n.t(:active_distributors_not_ready_for_checkout_message_singular,
-                 distributor_names: distributor_names)
-        end
-      end
-
       def html_request?
         request.format.html?
       end
@@ -145,4 +125,3 @@ module Spree
     end
   end
 end
-# rubocop:enable Metrics/ClassLength
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -115,7 +115,7 @@ class Enterprise < ActiveRecord::Base
   scope :not_ready_for_checkout, lambda {
     # When ready_for_checkout is empty, return all rows when there are no enterprises ready for
     # checkout.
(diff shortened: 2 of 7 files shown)


app/controllers/spree/admin/base_controller.rb
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
```
luisramos0: I thought these exceptions/comments were for things we accept we will keep forever.
I think we should try at some point to make this class shorter, so I'd say this should be in rubocop manual.

Matt-Yorkley: I've refactored the BaseController a bit to make it shorter.

luisramos0: nice refactor👌

app/models/enterprise.rb
```diff
@@ -117,7 +117,7 @@ class Enterprise < ActiveRecord::Base
     # checkout.
     ready_enterprises = Enterprise.ready_for_checkout.select('enterprises.id')
     if ready_enterprises.present?
-      where("enterprises.id NOT IN (?)", ready_enterprises)
+      where("enterprises.id NOT IN (?)", ready_enterprises.map(&:id))
```
mkllnk: I think that we can use `pluck(:id)` here to be more efficient. The `select` isn't needed then. But this could have a performance impact, the array can be big.

luisramos0: pluck could create an extra query, right?
I think this should be select :-)
https://github.com/openfoodfoundation/openfoodnetwork/pull/3709
I think we should agree on some rules for using map vs select vs pluck :-)

Matt-Yorkley: > I think we should agree on some rules for using map vs select vs pluck :-)

That sounds good! And we could add it to the wiki

Matt-Yorkley:
> Can you link me to some information about the issue with subqueries in Rails 4?

It's difficult to find anything concrete for this specific error. Basically when a subquery is not specifying which columns to fetch and defaulting to `SELECT *`, it throws a fatal error. It seems to be specific to the Postgres and Rails 4. I guess it enforces more efficient queries by forcing you to select only the columns you will actually use in the subquery.

luisramos0:
there's this issue for it:
https://github.com/openfoodfoundation/openfoodnetwork/issues/3708

app/models/enterprise.rb
mkllnk: Thank you for the clarification. I now think that I know how it works.

In this particular case, there was a select that would lead to a sub-query. The `present?` call creates a separate query just looking if anything is there without loading the enterprises. Then applying `map` would first load all the enterprise objects and then copy the ids into an array. A `pluck` call would do that without creating the enterprise objects (less Ruby work).

If we used the enterprise objects somewhere else, we should use map to re-use the same array and not load the enterprises again. But I don't think we do it here, so pluck would be better than map.

Matt, you explained that `select *` is a problem with sub-queries in Rails 4. That's logical. We should definitely avoid that. In this case, we do `select('enterprises.id')` in line 118 which should allow us to do an efficient sub-query. That should be no problem with Rails 4, right?

With my current knowledge, we don't need `map` here and should use a sub-query. We could move the `select` from line 118 down to the 120 to make it clearer that we are selecting the ids for the `IN` query part.

luisramos0: awesome Maikel 👌

Matt-Yorkley: I've done a bit more reading on Rails subqueries. The docs seem to suggest we need to remove both the `select` _and_ the `pluck`, and leave Rails to build the subquery..? 

I'll have a play with it in the Rails 4 branch and see if I can improve things without Postgres complaining.

Matt-Yorkley: There's some other interesting errors when playing with this in the Rails 4 branch, including: `ActiveRecord::StatementInvalid: PG::ProtocolViolation: ERROR:  bind message supplies 0 parameters, but prepared statement "" requires 1`.

Some relevant discussion is here: https://github.com/rails/rails/issues/20077

Apparently there are some other issues with subqueries using associations / collections / prepared statements. Adding `#to_sql` on some of the pieces seems to work in some cases, eg: `distributor_id: user.enterprises.to_sql`

In this specific case, the nice Rails 4 way is to use the `#not` method to avoid the prepared statement, with something like: `where.not(id: ready_enterprises)`, but we can't do that in `master`.

Matt-Yorkley: Aha! Yeah it looks like it's mostly the prepared statements in our scopes that are causing issues. In the `#managed_by` scope in  `enterprise_fee.rb` this line throws a "subquery has too many columns" error:

```ruby
where('enterprise_id IN (?)', user.enterprises)
```

but this does not: 

```ruby
where(enterprise_id: user.enterprises)
```

luisramos0: nice investigation Matt!!
is there a reason you are not satisfied with how we have been fixing things so far with:
where('enterprise_id IN (?)', user.enterprises.select(&:id))

Matt-Yorkley: I need to redo the other subquery PR and check some things in Rails 4. It looks like it will be better if we avoid prepared statements and favour the hash syntax, passing an `ActiveRecord::Relation` object (as mentioned above). And in that case I think we generally don't need the `select`, as `ActiveRecord` should in theory figure it out and build the subquery correctly?

luisramos0: I am not sure I agree.
Prepared statements are a good thing in general. I dont think we should plan to just avoid them.
I dont think we should let active record control things, we should control things and we should name the exact fields we want in the query.
These are general rules I have seen used in other projects: as much prepared statements as possible and lists of fields in a query as specific as possible.

mkllnk: In general, the new Rails 4 way `where.not(id: ready_enterprises)` sounds better than what we have at the moment. But for compatibility with Rails 3 and 4, I would prefer `      where("enterprises.id NOT IN (?)", ready_enterprises.select("enterprises.id"))` for now. The reason I prefer this version is that it's very explicit what's happening, it's compatible with many versions and avoids the Rails 3 bug where multiple `where(id: collection)` statements override each other.

Instead of optimising the way we use the ActiveRecord interface further, I would also suggest looking at the underlying problem and changing our logic here. This negation is really expensive. It would be better to find criteria for enterprises that make them not being ready for checkout and using those in a direct query.

One problem is that there can be several reasons why an enterprise is not ready. And we are querying them all together. Instead, we could actually query those individually and change the UI to say:

- These enterprises are missing a shipping method: ...
- These enterprises are missing a payment method: ...
- ...

@Matt-Yorkley I noticed that your most recent version still uses `pluck`. Does `select('enterprises.id')` still raise the too-many-column error? I don't understand that because `enterprises.id` is only one column, isn't it? Or are there many more in JOIN clauses that count as well?

app/controllers/spree/admin/base_controller.rb
```diff
@@ -15,11 +15,10 @@ class BaseController < Spree::BaseController
       # Warn the user when they have an active order cycle with hubs that are not ready
       # for checkout (ie. does not have valid shipping and payment methods).
       def warn_invalid_order_cycles
-        distributors = active_distributors_not_ready_for_checkout
+        return if flash[:notice].present?
 
-        return if distributors.empty? || flash[:notice].present?
-
-        flash[:notice] = active_distributors_not_ready_for_checkout_message(distributors)
+        warning = OrderCycleWarning.new(spree_current_user).call
```
sauloperez: With this abstraction the intent is much clearer now :ok_hand:

kristinalim:
@mkllnk Are you happy with this PR already? :slightly_smiling_face:

mkllnk:
I'm still concerned about the performance of `pluck` in that one example. If it's really needed to be compatible with Rails 4, okay. But I got mixed answers here, nothing explaining why we can't use `select` in this example, selecting only once column in the subquery.

Matt-Yorkley:
> I'm still concerned about the performance of pluck in that one example.

I'll try it again in Rails 4

Matt-Yorkley:
Aha! The resulting query here looked like this in Rails 4: 

...NOT IN (SELECT DISTINCT **enterprises.*, enterprises.id** FROM "enterprises" JOIN...

The scope that's being negated here uses `.select('DISCINCT enterpises.*)` at the end. I've used #except here to override it, and it now works in Rails 4 without #pluck.

mkllnk:
Great work @Matt-Yorkley! Only Code Climate is the usual grinch. :wink:

sauloperez:
All green so merging!

Issue #4639, opened 2020-01-03
[Spike] Automate Stripe testing


lin-d-hop:
## Description

As a tester and developer I want to know if Stripe testing can be fully automated. 

Currently testers test that new enterprises can be connected to a Stripe account and this account can then be used to take payments in every release. This is the main automation win for testers.

## Acceptance Criteria & Tests

To give an overview - the following is a list of all the places users/enterprises interact with Stripe in the platform:
1. Enterprise Owners/Managers can take payments from backoffice Orders -> Payments
2. Shoppers can save cards from Accounts interface
3. Shoppers can give permission for enterprises to charge their cards in Accounts interface
4. Shoppers can buy with cards at Checkout
5. Shoppers can save cards at Checkout
6. Enterprise Owners/Managers can create subscriptions with cards that a shopper has previously saved and given permission for the enterprise to charge from
7. Enterprise Owners/Manager can refund - partially and fully - credit owning to shoppers.

This is a spike so no testing required.

Matt-Yorkley:
Am I right in thinking the plan is something like this?

- Set up a Stripe testing account / API key that we will use in the CI build
- Add the API key to the test suite
- Write new specs that actually make real calls to Stripe, using the various "test cards" from their documentation

luisramos0:
Hey Matt, we dont have a plan. Kristina was assigned to this just before she stopped contributing regularly and I have never really had a look at it.

Your plan is one option.
I found this now: https://github.com/stripe/stripe-mock

luisramos0:
@mkllnk mentioned a few tools they use in CERES? VCR? I think now is the right time for us to decide what's the tech approach to this challenge. First we need a list of options :-) @mkllnk can you comment? Thanks!

mkllnk:
Well, Ceres Fair Food doesn't use Stripe. The used payment gateway has simple API calls without webhooks. But my idea to got forward would be:

* Use [VCR](https://github.com/vcr/vcr/) to record and playback API interactions from Ruby side. So if we call the Stripe API, we have to do that once to record the response. The response is stored in a file in the repository and automated tests will use it to replay that response without ever needed to contact Stripe. If you want to test the real interaction with Stripe again, just delete the file and it gets re-recorded.
* Use [Puffing Billy](https://github.com/oesmith/puffing-billy) to mock API calls within the browser. It allows us to simulate Stripe's behaviour when a user enters a new card.
[...]

Issue #4180, opened 2019-08-22
Create and integrate PaymentIntent for Stripe payments


kristinalim:
## Description

- Related to #4170 

Ongoing and new Stripe payment records should have a Payment Intent set up. This will allow us to handle additional actions, such as customer authentication, if needed for a payment.

Also explore if this should be done at the Order level, but I think this approach would be a lot more complicated considering that we support other payment methods too.

[Gist of Setup Intents](https://stripe.com/docs/api/setup_intents):

> A SetupIntent guides you through the process of setting up a customer's payment credentials for future payments. For example, you could use a SetupIntent to set up your customer's card without immediately collecting a payment.

[Gist of Payment Intents](https://stripe.com/docs/api/payment_intents):

> A PaymentIntent guides you through the process of collecting a payment from your customer. **We recommend that you create exactly one PaymentIntent for each order or customer session in your system.** You can reference the PaymentIntent later to see the history of payment attempts for a particular session.

(Added by @kristinalim on 2019-11-12:)

Not to imply that we create the PaymentIntents even for non-Stripe checkouts/payments. It is more of, if we create a PaymentIntent, do we associate this with the order or the payment record?

For example, if the PaymentIntent is associated with the payment record, if I try to make a failed charge and then have a separate successful charge, these would be under two Stripe PaymentIntent objects.

However, if we have one one PaymentIntent object for each order that would use a Stripe payment, we can then have it recorded on Stripe as:

```
PaymentIntent object (for Order N)
- Payment attempt A
- Partial payment B
- Adjustment payment C
```

## Acceptance Criteria

Each new and ongoing Stripe payment record should have a Payment Intent on Stripe. This stage extends on the work done in #4173  to ensure that OFN is working effectively and efficiently with the Stripe API.

## Success factors = expected outcome 

**1) Shopper Checkout - The shopper does not wish to save the card:**
For devs: 
- The checkout flow should support SCA if requested from the bank. 
- Use 'on-session' flow.

For testers:
- [ ] Test checkout with a test card that requires SCA - 4000 0025 0000 3155
This card should request an authentication. Stripe will trigger a modal asking for the customer to authenticate.

- [ ] Test checkout with a test card that does not require SCA - 4242 4242 4242 4242
This card should checkout without authentication.

**2) Shopper Checkout - If the shopper DOES wish to save the card** 
For devs:
- SCA is requested and SetupIntent stored every time.
- Use 'off-session' flow.

For testers:
[...]

Issue #4173, opened 2019-08-20
Update backend payment collection page to use Payment Intents API


kristinalim:
## What is the problem we are solving

- References #4170 

Setting new credit card info for Stripe payment method should use the Payment Method and Setup Intents API and support SCA. The `off-session` mode would be applicable. (Maybe Payment Intents and `on-session` mode is more appropriate for checkout that will not save the card? Needs to be checked.)

Do this for the following pages:

1. Customer checkout page
2. Customer saved cards page
3. Admin new payment page for order

Support for non-authenticated cards for the following should remain intact and should be tested:

1. Checkout
2. Capturing of payment from admin section
3. Subscriptions

[Gist of Setup Intents](https://stripe.com/docs/api/setup_intents):

> A SetupIntent guides you through the process of setting up a customer's payment credentials for future payments. For example, you could use a SetupIntent to set up your customer's card without immediately collecting a payment.

[Gist of Payment Intents](https://stripe.com/docs/api/payment_intents):

> A PaymentIntent guides you through the process of collecting a payment from your customer. **We recommend that you create exactly one PaymentIntent for each order or customer session in your system.** You can reference the PaymentIntent later to see the history of payment attempts for a particular session.

## Success factors = expected outcome 

**1) Shopper Checkout - The shopper does not wish to save the card:**
For devs: 
- The checkout flow should support SCA if requested from the bank. 
- Use 'on-session' flow.

For testers:
- [ ] Test checkout with a test card that requires SCA - 4000 0025 0000 3155
This card should request an authentication. Stripe will trigger a modal asking for the customer to authenticate.

- [ ] Test checkout with a test card that does not require SCA - 4242 4242 4242 4242
This card should checkout without authentication.

**2) Shopper Checkout - If the shopper DOES wish to save the card** 
For devs:
- SCA is requested and SetupIntent stored every time.
- Use 'off-session' flow.

For testers:
- [ ] Test checkout with a test card that requires SCA - 4000 0025 0000 3155
This card should request an authentication. Stripe will trigger a modal asking for the customer to authenticate.

- [ ] Test checkout with a test card that does not require SCA - 4242 4242 4242 4242
This card should request an authentication. Stripe will trigger a modal asking for the customer to authenticate.

**3) The shopper adds a new saved card in Account -> Cards**
For devs:
- SCA is requested and SetupIntent stored every time.
- Use 'off-session' flow.

For testers:
[...]

Issue #4170, opened 2019-08-20
Support for Strong Customer Authentication (SCA) in Stripe integration - Part 2


kristinalim:
## Description

Banks will start declining payments that have not been SCA-authenticated in countries that will require this. We need to start supporting SCA, and thankfully Stripe makes this easy/manageable.

[See migration guidelines here.](https://stripe.com/docs/strong-customer-authentication/migration)

See discussion here #3927 and the articles [here](https://stripe.com/docs/strong-customer-authentication/migration).

Issues:
1. Completion of Part 1 #4211 
2. Customer should see updated terms of off-session charging through Stripe #4175
3. Explanation if SCA authentication required when adding Stripe payment for an order in the admin section #4178
4. Explanation in email if payment for subscription failed because of SCA authentication #4179
5. Allow a customer to perform further action for a Stripe payment if needed #4181
6. Allow a customer to perform further action for a Stripe card if needed #4182
7. Inform customers n days before their subscription orders are due payment if likely to need SCA authentication #4183

## Acceptance Criteria

The problem will be solved when there is:

1. User flow for customers to save and use new authenticated card info
2. User flow for admin to set unauthenticated card info for orders, with proper notification when the card cannot be saved because authentication is required
3. User flow for customers to perform SCA authentication on an existing stored card if needed
4. User flow for customers to perform SCA authentication for an existing Stripe payment attempt if needed
5. Custom text in failed payment email for customers whose subscription payment failed because of lack of SCA authentication (Check: Is this needed for some non-subscription orders?)
6. Email notification for customers likely to need SCA authentication _n_ days before their subscription orders are due payment. - I think this should be less than 7 days. Stripe releases charge authorization if not captured within 7 days.
7. Enterprises are informed when subscription payments fail for shoppers

kristinalim:
Regarding current item 2: "User flow for admin to set unauthenticated card info for orders, with proper notification when the card cannot be saved because authentication is required"

It is still possible that SCA authentication will not be needed for the customer or the payment. But If it is required at this point, you are not able to proceed to save the card on Stripe because the customer cannot authenticate. (Unless we want shop owners to be able to enter authentication info for customers, e.g. when they are communicating directly. I suspect this would be violating the regulation though.)

lin-d-hop:
Notes from running tests on our payment flows using test cards specified by Stripe to trigger error flows:

1) Once-off payment for an order with a new card requiring 3D secure authentication:
[...]

PR #4672, opened 2020-01-13
Add new Stripe payment method compatible with the new Stripe Payment Intents API

### LINKED ISSUE #4686: Add new Stripe SCA payment method compatible with the new Stripe Payment Intents API
## Description
A new payment method called Stripe SCA should be added to OFN. This payment method should work as the exiting Stripe payment method but work with the Payment Intents API instead of the Charges API.
This payment method will not support the new SCA auth processes yet. It should just work like the existing stripe payment method. After this is done we can work on this payment method to implement the SCA processes.

This new payment method will not work with cards already in the system (only new ones) and will not allow users to add a card in the account section, the only way to save a card for now will be to checkout and use the "save card" option. These extra steps will be implemented in issue #4685 and will make the new payment method fully functional.

A new payment method Stripe SCA is available and works in the following processes:

1. make a payment without saving card
2. make a payment saving card
3. make a payment using previously saved card
4. delete saved card
5. capture the payment in the admin payments section of an order
6. as a user authorize a hub to use your card and verify  the subscription payments work with that card (the card must be added by checking out with it and not through the "add card" form)

luisramos0:
#### What? Why?

Relates to #4180
Closes #4686 and #4173

The decision to duplicate the existing Stripe payment method has the objective of making the migration as smooth as possible and also of making it possible for us, after we migrate all users, to delete the original stripe payment method (and its code), and keep only the new one (that uses the new SCA compatible APIs).
Although this new API was made by stripe for European SCA it does work worldwide, we don't need to keep both integrations :+1:

This PR is not introducing any new auth processes yet. It should just work like the existing stripe payment method.

I have added entries to rubocop_manual_todo for the remaining rubocop issues. Code climate is not picking the exceptions list from rubocop_manual_todo...

In terms of tests we are not adding any feature tests here as we plan to get something new with #4639

#### What should we test?
We need to verify that the existing Stripe payment method works.
We need to verify the new payment method, it should just work like the existing stripe payment method.
Processes we need to test:
1. make a payment without saving card
2. make a payment saving card
3. make a payment using previously saved card
4. add a card on account section and use it to checkout
5. delete saved card
6. capture the payment in the admin payments section of an order
7. as a user authorize a hub to use your card and verify the subscription payments work with that card

The following details are valid before #4719, if we are testing 4719, all cases above should work.
<details>
As discussed in comments below: case 4 above is not working yet for this payment method. But case 7, subscriptions, is working.
Because case 4 doesnt work (if you add a card in the account section and use it in a subscription with this new payment methods, the payments will fail). To add a card that works for a subscription with this new payment method you need to checkout an order with this new payment method with the save card option enabled. After that go to account to authorize the card you have used to checkout.
</details>

#### Release notes
Changelog Category: Added
Added new payment method that integrates with the new Stripe Payment Intents API that is compatible with SCA. This is a initial preparation for us to implement the more complex payment authentication flows that banks will start to enforce in Europe soon.

### DIFF
--- a/app/assets/javascripts/admin/payments/services/payment.js.coffee
+++ b/app/assets/javascripts/admin/payments/services/payment.js.coffee
@@ -21,16 +21,7 @@ angular.module('admin.payments').factory 'Payment', (AdminStripeElements, curren
               year: @form_data.card_year
               verification_value: @form_data.card_verification_value
           }
-        when 'stripe'
-          angular.extend munged_payment.payment, {
-            source_attributes:
-              gateway_payment_profile_id: @form_data.token
-              cc_type: @form_data.cc_type
-              last_digits: @form_data.card.last4
-              month: @form_data.card.exp_month
-              year: @form_data.card.exp_year
-          }
-        when 'stripe_sca'
+        when 'stripe', 'stripe_sca'
           angular.extend munged_payment.payment, {
             source_attributes:
               gateway_payment_profile_id: @form_data.token
--- a/app/assets/javascripts/admin/payments/services/payment.js.coffee
+++ b/app/assets/javascripts/admin/payments/services/payment.js.coffee
@@ -30,11 +30,22 @@ angular.module('admin.payments').factory 'Payment', (AdminStripeElements, curren
               month: @form_data.card.exp_month
               year: @form_data.card.exp_year
           }
+        when 'stripe_sca'
+          angular.extend munged_payment.payment, {
+            source_attributes:
+              gateway_payment_profile_id: @form_data.token
+              cc_type: @form_data.cc_type
+              last_digits: @form_data.card.last4
+              month: @form_data.card.exp_month
+              year: @form_data.card.exp_year
+          }
       munged_payment
 
     purchase: ->
       if @paymentMethodType() == 'stripe'
         AdminStripeElements.requestToken(@form_data, @submit)
+      else if @paymentMethodType() == 'stripe_sca'
+        AdminStripeElements.createPaymentMethod(@form_data, @submit)
       else
         @submit()
 
--- a/app/controllers/spree/admin/payment_methods_controller.rb
+++ b/app/controllers/spree/admin/payment_methods_controller.rb
@@ -110,7 +110,7 @@ module Spree
                      else
                        Gateway.providers.reject{ |p| p.name.include? "Bogus" }.sort_by(&:name)
                      end
-        @providers.reject!{ |p| p.name.ends_with? "StripeConnect" } unless show_stripe?
+        @providers.reject!{ |provider| stripe_provider?(provider) } unless show_stripe?
         @calculators = PaymentMethod.calculators.sort_by(&:name)
       end
 
@@ -134,18 +134,28 @@ module Spree
       # current payment_method is already a Stripe method
       def show_stripe?
         Spree::Config.stripe_connect_enabled ||
-          @payment_method.try(:type) == "Spree::Gateway::StripeConnect"
+          stripe_payment_method?
       end
 
       def restrict_stripe_account_change
         return unless @payment_method
-        return unless @payment_method.type == "Spree::Gateway::StripeConnect"
+        return unless stripe_payment_method?
         return unless @payment_method.preferred_enterprise_id.andand > 0
 
         @stripe_account_holder = Enterprise.find(@payment_method.preferred_enterprise_id)
         return if spree_current_user.enterprises.include? @stripe_account_holder
 
         params[:payment_method][:preferred_enterprise_id] = @stripe_account_holder.id
+      end
+
+      def stripe_payment_method?
+        @payment_method.try(:type) == "Spree::Gateway::StripeConnect" ||
+          @payment_method.try(:type) == "Spree::Gateway::StripeSCA"
+      end
+
+      def stripe_provider?(provider)
+        provider.name.ends_with?("StripeConnect") ||
+          provider.name.ends_with?("StripeSCA")
       end
     end
   end
--- a/app/controllers/spree/admin/payment_methods_controller.rb
+++ b/app/controllers/spree/admin/payment_methods_controller.rb
@@ -149,13 +149,12 @@ module Spree
       end
 
       def stripe_payment_method?
-        @payment_method.try(:type) == "Spree::Gateway::StripeConnect" ||
-          @payment_method.try(:type) == "Spree::Gateway::StripeSCA"
+        ["Spree::Gateway::StripeConnect",
+         "Spree::Gateway::StripeSCA"].include? @payment_method.try(:type)
       end
 
       def stripe_provider?(provider)
-        provider.name.ends_with?("StripeConnect") ||
-          provider.name.ends_with?("StripeSCA")
+        provider.name.ends_with?("StripeConnect", "StripeSCA")
       end
     end
   end
--- a/app/models/spree/gateway/stripe_connect.rb
+++ b/app/models/spree/gateway/stripe_connect.rb
@@ -9,12 +9,6 @@ module Spree
 
       attr_accessible :preferred_enterprise_id
 
-      CARD_TYPE_MAPPING = {
-        'American Express' => 'american_express',
-        'Diners Club' => 'diners_club',
-        'Visa' => 'visa'
-      }.freeze
-
       def method_type
         'stripe'
       end
@@ -75,11 +69,6 @@ module Spree
         creditcard = token_from_card_profile_ids(creditcard)
 
         [money, creditcard, options]
-      end
-
-      def update_source!(source)
-        source.cc_type = CARD_TYPE_MAPPING[source.cc_type] if CARD_TYPE_MAPPING.include?(source.cc_type)
-        source
       end
 
       def token_from_card_profile_ids(creditcard)
--- a/lib/open_food_network/subscription_payment_updater.rb
+++ b/lib/open_food_network/subscription_payment_updater.rb
@@ -30,8 +30,8 @@ module OpenFoodNetwork
     end
 
     def card_required?
-      payment.payment_method.is_a?(Spree::Gateway::StripeConnect) ||
-        payment.payment_method.is_a?(Spree::Gateway::StripeSCA)
+      [Spree::Gateway::StripeConnect,
+       Spree::Gateway::StripeSCA].include? payment.payment_method.class
     end
 
     def card_set?
--- a/lib/open_food_network/subscription_payment_updater.rb
+++ b/lib/open_food_network/subscription_payment_updater.rb
@@ -30,7 +30,8 @@ module OpenFoodNetwork
     end
 
     def card_required?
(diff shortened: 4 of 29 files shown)


Matt-Yorkley:
Nice work! Looks like some of the stubbed requests in the new specs need adjusting a bit...

luisramos0:
yes, thanks @Matt-Yorkley I am finishing that spec today. I wanted to get some early feedback :+1:

lib/stripe/profile_storer.rb
```diff
@@ -4,17 +4,22 @@
...
+      @stripe_account_id = stripe_account_id
     end
 
     def create_customer_from_token
       token = @payment.source.gateway_payment_profile_id
       response = @provider.store(token, options)
 
       if response.success?
-        attrs = source_attrs_from(response)
+        if response.params['customer'] # Payment Intents API
```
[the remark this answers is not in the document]
luisramos0: yes, I did that in a later commit :+1:

lib/stripe/profile_storer.rb
```diff
@@ -24,11 +29,13 @@ def create_customer_from_token
...
-        email: @payment.order.email,
-        login: Stripe.api_key,
-        address: address_for(@payment)
-      }
+      options = {
+                  email: @payment.order.email,
+                  login: Stripe.api_key,
+                  address: address_for(@payment)
+                }
+      options = options.merge({ stripe_account: @stripe_account_id }) if @stripe_account_id.present?
```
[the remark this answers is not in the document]
luisramos0: yeah, nice. I added this.

luisramos0:
ok, I rebased and added the two last commits: make one method better looking and add some missing translations.

There are two test scenarios that are not working yet. 4 and 7.
4. Adding a card in the account section
the problem here is that this customer form does not have any association with a payment method, so we dont know if this will be used for a stripe payment method with SCA or the legacy stripe api. Currently the form is adding the card to the legacy stripe api and these cards will only work with payment methods that use the legacy stripe api (non sca). I am not sure how we should proceed here because while the new sca payment method is not completed and all user data migrated, we will still want the old stripe payment methods to work.

7. subscriptions
Without point 4 above cards can only be added through checking out with save card option.
There are a few more things to be done on the subscriptions front to make this work.

I think we should ship this PR as is if no one disagress. the new stripe sca payment method will be available but only works in limited list of use cases. We can then improve it until we can replace the old one.

lin-d-hop:
Thanks @luisramos0 I agree with your approach above. 

Do you foresee any issues when we come to merge the old API user data with the new API user data. As the above approach means we'll have split user data including potentially the same user across both APIs. I guess this is totally handled by Stripe?

RachL:
Something I'm not sure to understand: if we upgrade the API, doesn't it means that we should disable the current form that saves card in the legacy API?
I mean, migration is always a mess, so the less card to migrate the better?

luisramos0:
@RachL this is not related to upgrading to the latest api version.
this is related to using a different api: charges api (non sca compliant) vs payment intents api (sca compliant).
All stripe api versions include the charges api but we want to upgrade so that we get the recent versions of the payment intents api with all the sca features.

We can have both payment methods work at the same time, we just need to see what to do with this customer form.

if you agree we can just go forward with this PR as is. I can start investigate the easiest approach afterwards:
- maybe I can make the add card form add to both apis
- maybe I can make the cards added through the old api (charges api) work with the new api (payment intents api) - this would be great because we wouldnt need to migrate the user cards
- other approach

luisramos0:
@lin-d-hop answering your question: I still have to see if I can use the existing data in the new api. I havent managed so far but I think it will be possible.

Conclusion: lets go forward with this PR that adds the new payment method (Stripe SCA) and works with the new payment intents api. This way we can already validate it in staging and live. And meanwhile we can work on making this payment method ready for the full process.

luisramos0:
ok, it looks like subscriptions are working :-)

This is ready to be reviewed and tested :+1:
@mkllnk I have added the 4 last commits since your review.
@Matt-Yorkley can you please review?
Thanks!!!!

luisramos0:
1. I dont think this will be a problem because the new payment method is basically working. The only case not working is for users to add a card in the account section. Frontoffice. And they will see it doesnt work because the payments will fail. Also I think we will be ready very soon and we will probably rename the existing stripe one to something like "Stripe (legacy)". Having said that: what do other people think? Shall we name this new payment method "Stripe SCA (beta)"?

2. "a card added with the old API is not usable in the intents API " this is not a fact and probably it is incorrect, I need to investigate further.
It is legal to add a card as an admin, yes, for example, buying by phone. what we will have is a process to send the customer an email to authorize the transaction if that is what the bank requires. These cases are covered in the stripe epic issues if you want to see more details: https://github.com/openfoodfoundation/openfoodnetwork/issues/4170

app/assets/javascripts/admin/payments/services/payment.js.coffee
```diff
@@ -30,11 +30,22 @@ angular.module('admin.payments').factory 'Payment', (AdminStripeElements, curren
               month: @form_data.card.exp_month
               year: @form_data.card.exp_year
           }
+        when 'stripe_sca'
```
Matt-Yorkley: Is duplicating this block necessary? It's identical to the one above.

app/controllers/spree/admin/payment_methods_controller.rb
```diff
@@ -134,19 +134,29 @@ def load_hubs
...
         return unless @payment_method.preferred_enterprise_id.andand > 0
 
         @stripe_account_holder = Enterprise.find(@payment_method.preferred_enterprise_id)
         return if spree_current_user.enterprises.include? @stripe_account_holder
 
         params[:payment_method][:preferred_enterprise_id] = @stripe_account_holder.id
       end
+
+      def stripe_payment_method?
+        @payment_method.try(:type) == "Spree::Gateway::StripeConnect" ||
```
Matt-Yorkley: `["Spree:Gateway::StripeConnect", "Spree::Gateway::StripeSCA"].include? @payment_method.try(:type)` ?

app/controllers/spree/admin/payment_methods_controller.rb
```diff
@@ -134,19 +134,29 @@ def load_hubs
...
+          @payment_method.try(:type) == "Spree::Gateway::StripeSCA"
+      end
+
+      def stripe_provider?(provider)
+        provider.name.ends_with?("StripeConnect") ||
```
Matt-Yorkley: `provider.name.ends_with?("StripeConnect", "StripeSCA")` ?

lib/open_food_network/subscription_payment_updater.rb
```diff
@@ -30,7 +30,8 @@ def create_payment
     end
 
     def card_required?
-      payment.payment_method.is_a? Spree::Gateway::StripeConnect
+      payment.payment_method.is_a?(Spree::Gateway::StripeConnect) ||
```
Matt-Yorkley: `[Stripe::Gateway::StripeConnect, Spree::Gateway::StripeSCA].include? payment.payment_method.class` ?

spec/requests/checkout/stripe_sca_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+          put update_checkout_path, params
+
+          expect(response.status).to be 400
+
+          expect(json_response["flash"]["error"]).to eq "payment-intent-failure"
+          expect(order.payments.completed.count).to be 0
+        end
+      end
+
+      context "when the payment_metho request returns an error message" do
```
Matt-Yorkley: `payment_metho` :trollface:

spec/requests/checkout/stripe_sca_spec.rb
```diff
@@ -0,0 +1,281 @@
+require 'spec_helper'
+
+describe "checking out an order with a Stripe SCA payment method", type: :request do
```
Matt-Yorkley: :heart:

app/models/spree/gateway/stripe_connect.rb
```diff
@@ -9,12 +9,6 @@ class StripeConnect < Gateway
 
       attr_accessible :preferred_enterprise_id
 
-      CARD_TYPE_MAPPING = {
```
Matt-Yorkley: Isn't this needed anywhere for Stripe Connect? Are these changes to the old Stripe Connect code going to break it?

lib/open_food_network/subscription_payment_updater.rb
Matt-Yorkley: There are several of these checks that all look very similar. Maybe extracting it might be nice at some point?

app/models/spree/gateway/stripe_connect.rb
luisramos0: no, when adapting the new method I found this is dead code in the existing method.
in the commit message for this deletion you can see the commit when the code became dead:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4672/commits/785e89624cf02b084afd26cb3c9e116d4252664c

spec/requests/checkout/stripe_sca_spec.rb
luisramos0: fixed :+1:

app/assets/javascripts/admin/payments/services/payment.js.coffee
luisramos0: I dont know the exact syntax for the when with two cases...

but actually, this answers to the other comments below: these changes are optimized for deletion of the stripe connect method very soon.

do you mind if I leave my version, basically separate lines for each method? it will be straight forward to delete them soon.

app/controllers/spree/admin/payment_methods_controller.rb
luisramos0: :+1:

app/controllers/spree/admin/payment_methods_controller.rb
luisramos0: :+1:

lib/open_food_network/subscription_payment_updater.rb
luisramos0: we will remove the existing stripe method soon so it will become what it was before...
but it would be good to have a central place where this check is done yes, I thought it would take a little bit to do it and I decided to skip it for now...

luisramos0:
thanks for the review Matt.

Yeah, we can change its name or make it super admin only.
I am pretty confident we will make this migration really quickly so I'd just go for it as is.

luisramos0:
I managed to have 2 typos in this last commit.
I fixed that last commit and also rebased.

luisramos0:
I am blocking this PR with #4719.

There's no point in testing this PR and then test 4719 again. 4719 implements a different logic to store cards (puts them in the stripe platform account of the instance and then clones them to connected stripe accounts of the seller).

Let's wait for 4719 to be ready for testing and test it instead (it includes this PR code).

luisramos0:
rebased to resolve conflicts and pushed to upstream so that this is also seen in #4719

luisramos0:
tested as part of 4719, merging.

PR #4681, opened 2020-01-16
Rails 4 prep: subqueries

### LINKED ISSUE #3708: [Spree 2.1] Adapt to Rails 4 - subquery has too many columns

Matt-Yorkley:
#### What? Why?

Closes #3708 

Fixes all remaining issues related to the Postgres error: `PG::SyntaxError: subquery has too many columns` in preparation for Rails 4.

#### What should we test?

A green build should be enough.

#### Release notes

Updated lack of `select` in subqueries in preparation for Rails 4.

Changelog Category: Changed

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -317,7 +317,7 @@ class Enterprise < ActiveRecord::Base
   def distributed_taxons
     Spree::Taxon.
       joins(:products).
-      where('spree_products.id IN (?)', Spree::Product.in_distributor(self)).
+      where(spree_products: { id: Spree::Product.in_distributor(self).select(:id) }).
       select('DISTINCT spree_taxons.*')
   end
 
@@ -333,7 +333,7 @@ class Enterprise < ActiveRecord::Base
   def supplied_taxons
     Spree::Taxon.
       joins(:products).
-      where('spree_products.id IN (?)', Spree::Product.in_supplier(self)).
+      where(spree_products: { id: Spree::Product.in_supplier(self).select(:id) }).
       select('DISTINCT spree_taxons.*')
   end
 
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -317,7 +317,7 @@ class Enterprise < ActiveRecord::Base
   def distributed_taxons
     Spree::Taxon.
       joins(:products).
-      where(spree_products: { id: Spree::Product.in_distributor(self).select(:id) }).
+      where(spree_products: { id: Spree::Product.select(:id).in_distributor(self) }).
       select('DISTINCT spree_taxons.*')
   end
 
@@ -333,7 +333,7 @@ class Enterprise < ActiveRecord::Base
   def supplied_taxons
     Spree::Taxon.
       joins(:products).
-      where(spree_products: { id: Spree::Product.in_supplier(self).select(:id) }).
+      where(spree_products: { id: Spree::Product.select(:id).in_supplier(self) }).
       select('DISTINCT spree_taxons.*')
   end
 
--- a/app/models/enterprise_fee.rb
+++ b/app/models/enterprise_fee.rb
@@ -27,7 +27,7 @@ class EnterpriseFee < ActiveRecord::Base
     if user.has_spree_role?('admin')
       scoped
     else
-      where('enterprise_id IN (?)', user.enterprises)
+      where('enterprise_id IN (?)', user.enterprises.select('enterprises.id').map(&:id))
     end
   }
 
--- a/app/models/enterprise_relationship.rb
+++ b/app/models/enterprise_relationship.rb
@@ -22,7 +22,8 @@ class EnterpriseRelationship < ActiveRecord::Base
   }
 
   scope :involving_enterprises, ->(enterprises) {
-    where('parent_id IN (?) OR child_id IN (?)', enterprises, enterprises)
+    enterprise_ids = enterprises.map(&:id)
+    where('parent_id IN (?) OR child_id IN (?)', enterprise_ids, enterprise_ids)
   }
 
   scope :permitting, ->(enterprise_ids) { where('child_id IN (?)', enterprise_ids) }
--- a/app/models/exchange.rb
+++ b/app/models/exchange.rb
@@ -48,8 +48,9 @@ class Exchange < ActiveRecord::Base
       select('DISTINCT exchanges.*')
   }
   scope :with_product, lambda { |product|
+    exchange_variant_ids = product.variants_including_master.select('spree_variants.id').map(&:id)
     joins(:exchange_variants).
-      where('exchange_variants.variant_id IN (?)', product.variants_including_master)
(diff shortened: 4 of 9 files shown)


app/models/enterprise.rb
```diff
@@ -317,7 +317,7 @@ def category
   def distributed_taxons
     Spree::Taxon.
       joins(:products).
-      where('spree_products.id IN (?)', Spree::Product.in_distributor(self)).
+      where(spree_products: { id: Spree::Product.select(:id).in_distributor(self) }).
```
luisramos0: I guess this would look better? is it not the same?
Spree::Product.in_distributor(self).select(:id)

app/models/enterprise.rb
```diff
@@ -333,7 +333,7 @@ def current_distributed_taxons
   def supplied_taxons
     Spree::Taxon.
       joins(:products).
-      where('spree_products.id IN (?)', Spree::Product.in_supplier(self)).
+      where(spree_products: { id: Spree::Product.select(:id).in_supplier(self) }).
```
luisramos0: same as above

app/models/enterprise_fee.rb
```diff
@@ -27,7 +27,7 @@ class EnterpriseFee < ActiveRecord::Base
     if user.has_spree_role?('admin')
       scoped
     else
-      where('enterprise_id IN (?)', user.enterprises)
+      where('enterprise_id IN (?)', user.enterprises.select('enterprises.id').map(&:id))
```
luisramos0: from #3709 where I found that:
"A scope returns a relation but an association returns a CollectionProxy that has both methods from a relation select and pluck plus all the array methods including map. So we can use map on an association but not on a scope."

enterprises is an association, so you can use pluck here

BUT, why dont you just do select and return a relation, I dont think you need to return an array here.

app/models/enterprise_relationship.rb
```diff
@@ -22,7 +22,8 @@ class EnterpriseRelationship < ActiveRecord::Base
   }
 
   scope :involving_enterprises, ->(enterprises) {
-    where('parent_id IN (?) OR child_id IN (?)', enterprises, enterprises)
+    enterprise_ids = enterprises.map(&:id)
+    where('parent_id IN (?) OR child_id IN (?)', enterprise_ids, enterprise_ids)
```
luisramos0: With map you are forcing an extra query. I think you should use select here and make it a subquery:

```suggestion
        enterprise_ids = enterprises.select(&:id)
```

app/models/exchange.rb
```diff
@@ -48,8 +48,9 @@ class Exchange < ActiveRecord::Base
       select('DISTINCT exchanges.*')
   }
   scope :with_product, lambda { |product|
+    exchange_variant_ids = product.variants_including_master.select('spree_variants.id').map(&:id)
```
luisramos0: same as above, I think .select.map is equivalent to .pluck

luisramos0: but in this case it would also be better to just use select because it can go as a subquery in the following in clause.

app/models/spree/order_decorator.rb
```diff
@@ -55,8 +55,10 @@
     else
       # Find orders that are distributed by the user or have products supplied by the user
       # WARNING: This only filters orders, you'll need to filter line items separately using LineItem.managed_by
+      user_enterprise_ids = user.enterprises.select("enterprises.id").to_a
```
luisramos0: why convert to array if you can keep a relation that can become a subquery? I'd just remove the .to_a here

app/models/spree/order_decorator.rb
```diff
@@ -65,7 +67,8 @@
     if user.has_spree_role?('admin')
       scoped
     else
-      where('spree_orders.distributor_id IN (?)', user.enterprises)
+      distributor_ids = user.enterprises.select('enterprises.id').map(&:id)
+      where(spree_orders: { distributor_id: distributor_ids })
```
luisramos0: same here, I think this should be just select

app/models/spree/payment_method_decorator.rb
```diff
@@ -19,8 +19,9 @@
     if user.has_spree_role?('admin')
       scoped
     else
+      user_enterprise_ids = user.enterprises.select('enterprises.id').map(&:id)
```
luisramos0: same here, just select

app/models/spree/shipping_method_decorator.rb
```diff
@@ -14,8 +14,9 @@
     if user.has_spree_role?('admin')
       scoped
     else
+      user_enterprise_ids = user.enterprises.select('enterprises.id').map(&:id)
       joins(:distributors).
-        where('distributors_shipping_methods.distributor_id IN (?)', user.enterprises).
+        where('distributors_shipping_methods.distributor_id IN (?)', user_enterprise_ids).
```
luisramos0: same here, just select

app/models/spree/variant_decorator.rb
```diff
@@ -49,7 +49,9 @@
   }
 
   scope :for_distribution, lambda { |order_cycle, distributor|
-    where('spree_variants.id IN (?)', order_cycle.variants_distributed_by(distributor))
+    distributed_variant_ids = order_cycle.variants_distributed_by(distributor).
+      select('spree_variants.id').map(&:id)
+    where('spree_variants.id IN (?)', distributed_variant_ids)
```
luisramos0: same here...

PR #4691, opened 2020-01-20
Disable two codeclimate SCSS rules

Matt-Yorkley:
#### What? Why?

Disables the following two `codeclimate` SCSS rules:

- **All properties have to be ordered alphabetically.** This is really time-consuming and doesn't really add anything.
- **Decimals have to be written without the preceding zero**. Time consuming and pointless. Also makes the rules slightly less readable.

lint-order
leading-zero

These two rules get triggered almost any time our CSS is touched.

#### Release notes

Relaxed some SCSS rules

Changelog Category: Changed

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -12,6 +12,10 @@ plugins:
         enabled: false
       VendorPrefix:
         enabled: false
+      LeadingZero:
+        enabled: false
+      PropertySortOrder:
+        enabled: false
   duplication:
     enabled: true
     exclude_patterns:


Matt-Yorkley:
:pray:

mkllnk:
I know it's annoying when it complains about existing issues but don't you think we should apply these rules to new code?

I don't really care about the leading zero. An order on the properties could be nice though. I'm open to be convinced though. Do you have an example when it doesn't make sense to order them alphabetically?

Matt-Yorkley:
It just takes a lot of time. If you indent something or move it you have to go back and manually reorder every line alphabetically.

luisramos0:
@Matt-Yorkley I think we should ignore the rule if we dont feel like re-ordering the items, like in your case now. No problem. But I think we should keep the rule for the future. I think we should have an order.

Issue #4695, opened 2020-01-20
Checkout broken if weight calculator used in a payment or shipping method AND a product in the cart has unit_value ZERO


luisramos0:
## Description
If a shipping or payment method uses the weight calculator and a variant being checked out has variant unit different from weight (items for example) and has the unit value set to zero by the user. The checkout button wont work and will not show an error to the user.

## Expected Behavior
Checkout works and the product is included in the order.
The shipping method cost is calculated in some default way: what does zero unit_value mean? "zero items", "zero kilos", "zero liters" is a bit non-sense. The cost should probably default to zero in these cases.

## Actual Behaviour
Final checkout button does nothing. No error is shown.

## Steps to Reproduce
Setup a shipping method with a weight based calculator.
Create a product with variant_unit items and unit_value 0.
Add to OC
Checkout the product.
Bang.

## Workaround
Change the shipping method to something else.
Change the product unit value to something else.

## Severity
This was S2 until we discovered what it was. It's now an S3 I think. I got two cases in Katuma where users are setting unit value to zero. And weight based calculators are very common. This will happen again for sure.

## Your Environment
Katuma

* Version used:
* Browser name and version: firefox latest
* Operating System and version (desktop or mobile): osx catalina

## Possible Fix
Bugsnag - https://app.bugsnag.com/katuma/katuma/errors/5de6470bd3c5fe001a1ae122?filters[event.since][0]=30d&filters[error.status][0]=open

luisramos0:
this happened again in Portugal. This time the user sent the farmer a screenshot with the new error from #4699

Matt-Yorkley:
If the unit is set to `items` there's no way we can calculate weight on them whether it's `0` or not, right? So we can just skip those variants when doing weight calculations?

lin-d-hop:
That sounds right to me...

luisramos0:
I dont think that's correct Matt, if you read the code of the weight calculator you see that if variant_unit is items, the variant.weight field is used instead.

luisramos0:
I think this issue is as simple as making weight calculator a bit more resilient by handling this particular data situation without blowing up.

Matt-Yorkley:
> the variant.weight field is used instead

That's probably `nil` by default though, right?

luisramos0:
yeah, but method weight_per_variant is taking care of that with the || 0

luisramos0:
in the case:
product.variant_unit = items
variant.weight = not zero
[...]

PR #4699, opened 2020-01-21
Rescue unknown checkout errors

### LINKED ISSUE #4696: No error is displayed to the user when the checkout process fails on the server side
## Description
The user is not notified when the checkout button fails on the server.

## Expected Behavior
The user should be notified that an error occurred. If no details are known a generic error message can be shown telling the user to contact support.

## Actual Behaviour
The checkout page just remains available to be edited.

## Steps to Reproduce
You can follow for example #4695

## Workaround
no workaround

## Severity
bug-s2 - this is a disaster UX problem, whenever there's a problem like #4695 the users have no clue what happened after clicking the crucial checkout button. A farmer reported "users thought the order was placed".

## Possible Fix
We need to check the checkout request to the server and print a generic error if it blows up.

mkllnk:
#### What? Why?

Closes #4696

The user was not notified when the checkout button fails on the server. I now catch every possible standard error, send it to Bugsnag and return a generic message to the user.

#### What should we test?

Go through the description of #4695 to trigger a failing checkout. You should see a red error message popping up.

Screenshot from 2020-01-22 10-42-39

#### Release notes

Display an error message if the checkout fails due to an unknown bug.

Changelog Category: Added

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -68,6 +68,10 @@ class CheckoutController < Spree::CheckoutController
         render json: { path: order_path(@order) }, status: :ok
       end
     end
+  rescue StandardError => error
+    Bugsnag.notify(error)
+    flash[:error] = I18n.t("checkout.failed")
+    update_failed
   end
 
   # Clears the cached order. Required for #current_order to return a new order
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -68,6 +68,9 @@ class CheckoutController < Spree::CheckoutController
         render json: { path: order_path(@order) }, status: :ok
       end
     end
+  rescue Spree::Core::GatewayError => error
+    # This is done for all actions in the Spree::CheckoutController.
+    rescue_from_spree_gateway_error(error)
   rescue StandardError => error
     Bugsnag.notify(error)
     flash[:error] = I18n.t("checkout.failed")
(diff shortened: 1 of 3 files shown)


app/controllers/checkout_controller.rb
```diff
@@ -68,6 +68,13 @@ def update
         render json: { path: order_path(@order) }, status: :ok
       end
     end
+  rescue Spree::Core::GatewayError => error
+    # This is done for all actions in the Spree::CheckoutController.
+    rescue_from_spree_gateway_error(error)
```
luisramos0: nice catch :+1:

config/locales/en.yml
```diff
@@ -1181,6 +1181,7 @@ en:
     already_ordered:
       cart: "cart"
       message_html: "You have an order for this order cycle already. Check the %{cart} to see the items you ordered before. You can also cancel items as long as the order cycle is open."
+    failed: "The checkout failed. Please let us know so that we can process your order."
```
luisramos0: I wonder if this is the best message. From a user perspective: who is "us" and do I contact this "us"?

mkllnk: I know, it's ambiguous. Notifying the shop owner is probably the most important thing to sort out the order. We can see the error in Bugsnag but another prompt with details would be nice as well. It's difficult to give specific instructions if we don't know what the error is.

In the context of the shop, I hope that "us" is seen as the distributor and that people click on the Contact tab. I will ask our instance managers what they think.

sauloperez: in the meantime, I'm pretty sure they know the distributor and how to reach him out (facebook, twitter, website, phone number, etc.). I'd go with this and improve it later if we need to.

RachL:
This is working great!

image

PR #4709, opened 2020-01-23
Pluralize common variant unit names

### LINKED ISSUE #4172: When an article is setup by "item", if the quantity is more than one in the variant, the plural is not applied anymore
## Description

I remember that before v2, when a product was sold per "item", if the variant was for example "2 items", and the name of the item was "piece", it was displayed as "2 pieces" by default (with a s for plural).
It is not anymore in v2.

## Expected Behavior

If sold per item, if quantity more than 1, plural should be applied automatically

## Actual Behaviour

It is not

## Steps to Reproduce

1. Go to products page, setup a product per item, with two variants, put 2 as a quantity for the second and see, the item name is not plural

## Animated Gif/Screenshot

Capture du 2019-08-20 13-36-11

## Context

I saw that when doing customer support via skype screen sharing and noticed it.

## Severity
bug-s3: a feature is broken but there is a workaround : we can manually write the display as field but we need to do it for every plural quantity, need to think about it...

## Your Environment

* Version used: v2.2.3 I think (or 2.2.2)
* Browser name and version: Chrome and Firefox
* OFN Platform instance where you discovered the bug: French production

mkllnk:
#### What? Why?

Closes #4172

This adds the most popular unit names as singular and plural to our
locale for translation. The added Javascript performs a reverse lookup
to find the right singular/plural form of a unit name in that language.

I explained why I chose this solution on the issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/4172#issuecomment-577447644

#### What should we test?

Test in English because we haven't translated this yet.

* Go to admin -> products.
* Change a product to items instead of weight or volume.
* Add a unit name like `bunch` or `loaf`.
* Change the quantity including `0`, `1` and `2` of the variant and observe the change in the description.
* Save the changes.
* Go to the shopfront and verify the product description there.

#### Release notes

Unit descriptions like "1 bunch" or "4 boxes" can now be translated more accurately. The previous logic would sometimes fail to find the right plural, especially in languages other than English.

Changelog Category: Changed

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2721,7 +2721,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
       have_an_account: "Already have an account?"
       action_login: "Log in now."
 
-  # Most popular names used in variant_unit_name.
+  # Singular and plural forms of commonly used words.
   # We use these entries to pluralize unit names in every language.
   #
   # Extracted with the following query:
@@ -2731,7 +2731,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   #   puts "      one: \"#{name}\""
   #   puts "      other: \"#{name}s\"";
   # }
-  unit_names:
+  inflections:
     each:
       one: "each"
       other: "each"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2786,6 +2786,24 @@ See the %{link} to find out more about %{sitename}'s features and to start using
     tray:
       one: "tray"
       other: "trays"
+    piece:
+      one: "piece"
+      other: "pieces"
+    pot:
+      one: "pot"
+      other: "pots"
+    bundle:
+      one: "bundle"
+      other: "bundles"
+    flask:
+      one: "flask"
+      other: "flasks"
+    basket:
+      one: "basket"
+      other: "baskets"
+    sack:
+      one: "sack"
+      other: "sacks"
 
   producers:
     signup:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2732,86 +2732,57 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   #   puts "      other: \"#{name}s\"";
   # }
   unit_names:
-    # Used 379 times.
-    # Used 138 times.
     each:
       one: "each"
       other: "each"
-    # Used 332 times.
-    # Used 145 times.
-    # Used 93 times.
     bunch:
       one: "bunch"
       other: "bunches"
-    # Used 118 times.
-    # Used 63 times.
     pack:
       one: "pack"
       other: "packs"
-    # Used 90 times.
-    # Used 72 times.
     box:
       one: "box"
       other: "boxes"
-    # Used 81 times.
-    # Used 49 times.
     bottle:
       one: "bottle"
       other: "bottles"
-    # Used 71 times.
-    # Used 56 times.
     jar:
       one: "jar"
       other: "jars"
-    # Used 65 times.
     head:
       one: "head"
       other: "heads"
-    # Used 56 times.
-    # Used 41 times.
     bag:
       one: "bag"
       other: "bags"
-    # Used 53 times.
-    # Used 30 times.
     loaf:
       one: "loaf"
       other: "loaves"
-    # Used 43 times.
     single:
       one: "single"
       other: "singles"
-    # Used 41 times.
     tub:
       one: "tub"
       other: "tubs"
-    # Used 36 times.
     punnet:
       one: "punnet"
       other: "punnets"
-    # Used 32 times.
     packet:
       one: "packet"
       other: "packets"
-    # Used 30 times.
-    # Used 22 times.
     item:
       one: "item"
       other: "items"
-    # Used 29 times.
-    # Used 24 times.
     dozen:
       one: "dozen"
       other: "dozens"
-    # Used 26 times.
     unit:
       one: "unit"
       other: "units"
-    # Used 24 times.
     serve:
       one: "serve"
       other: "serves"
-    # Used 20 times.
     tray:
       one: "tray"
       other: "trays"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2721,6 +2721,101 @@ See the %{link} to find out more about %{sitename}'s features and to start using
       have_an_account: "Already have an account?"
       action_login: "Log in now."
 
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -2721,6 +2721,101 @@ See the %{link} to find out more about %{sitename}'s features and to start using
...
+  #
+  # Extracted with the following query:
+  # Spree::Product.group(:variant_unit_name).order("count_all DESC").count.each { |name, count|
+  #   puts "    # Used #{count} times."
+  #   puts "    #{name&.parameterize('_')}:"
+  #   puts "      one: \"#{name}\""
+  #   puts "      other: \"#{name}s\"";
+  # }
+  unit_names:
+    # Used 379 times.
```
luisramos0: I dont think we need to keep these numbers in this file. The numbers are useful to see what keys to have here but we I think we should remove these "statistics" for here.

lib/open_food_network/option_value_namer.rb
```diff
@@ -72,5 +71,39 @@ def scale_for_unit_value
...
+        key = @unit_keys[unit_name.downcase]
+
+        return unit_name unless key
+
+        I18n.t(key, scope: "unit_names", count: count, default: unit_name)
+      end
+
+      private
+
+      def unit_key_lookup
```
luisramos0: I think this is overcomplicated. Why not just load the key/value we need?

lib/open_food_network/option_value_namer.rb
```diff
@@ -72,5 +71,39 @@ def scale_for_unit_value
...
     end
+
+    def pluralize(unit_name, count)
+      I18nUnitNames.instance.pluralize(unit_name, count)
+    end
+
+    # Provides efficient access to unit name inflections.
+    # The singleton property ensures that the init code is run once only.
+    # The OptionValueNamer is instantiated in loops.
+    class I18nUnitNames
```
luisramos0: I think this break the single responsibility rule. We should have this in a separate class.
UnitNamePlural or Pluralizer or I18nUnitNames.
This (including option_value_names) should all go to something like app/services/naming

config/locales/en.yml
mkllnk: Okay, I thought it could be interesting to compare if we update it one day. But I'll remove them.

lib/open_food_network/option_value_namer.rb
mkllnk: Yeah, I need to add some documentation and explain it step by step. So the user enters "fardo". We can't look up "fardo" with I18n because there is no key named that. So we need to look up which key relates to "fardo" and we find `:bale`. Then we can translation `t(:bale, count: 2)` and get "fardos".

Or did you understand this part and you are hinting at something else?

lib/open_food_network/option_value_namer.rb
mkllnk: I agree, I will move it.

mkllnk:
@luisramos0 Did you see the js spec covering the pluralizer?

If we move the pluralizer code (js and ruby) to their own classes, should we make the solution independent of variant_unit_name? In that case, we should not use the scope `unit_names` in the locale but something more general like `inflections`.

lib/open_food_network/option_value_namer.rb
luisramos0: ok, I didnt understand that. It's not straight forward. Now I understand why this is necessary :+1:

luisramos0:
yeah, I'd generalize it. but only if it doesnt add even more complexity.

mkllnk:
@luisramos0 This is ready for review again.

- I added the missing words used in French.
- I moved the I18n scope from `unit_names` to `inflections` to be more general here.
- I moved the Ruby to its own lib file. I don't think it depends on our application and therefore doesn't need to be a service.
- I realised that I didn't consider the case of people switching locales. In our case that's mainly important for the enterprise user creating variants. They need to create variants in the same language as their locale. Otherwise it won't pluralise. There is no way around this because we can't know which language the user input is and one word can mean different things in different languages. So I added support for multiple locales, almost regretting going down the caching path. It adds so much more complexity.
- And I added more comments to explain everything. What do you think?

lib/open_food_network/i18n_inflections.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module OpenFoodNetwork
+  # Provides efficient access to unit name inflections.
```
luisramos0: this "unit name" should be replaced by "word" as this is no longer related specifically to unit_names. right?

lib/open_food_network/i18n_inflections.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module OpenFoodNetwork
```
luisramos0: I'd love to kill this namespace. First by moving the OFN related services to app/services (most of them). We could create a generic lib folder like lib/utils for these generic classes and existing generic single classes like Locking, Paperclippable, and RackRequestBlocker. What do you think?

mkllnk: Hm, yes, it would be nice to get rid of the namespace. But we need to come up with better namespaces. I don't find `Util` much better. We could use `I18n::Inflections` here but it could look like it's part of the official I18n library. It's just our add-on.

Moving files into their own namespaces would really help to see their independence. And we could create gems if we wanted to share those parts with other projects.

luisramos0: I see what you mean. Util is not great but I think lib/util is better than lib/OpenFoodNetwork because it makes it obvious the code is not dependent on OFN code.
lib/utils would just be for those small util class (like Inflections).
I think a separate folder under lib would be for a group of related classes that are independent of OFN like we have in Stripe.
In Java the folder lib/common is used for these classes.

Probably a topic for discourse, not this PR :-)

RachL:
@mkllnk I staged this on 2 different staging servers, but I still end up with an error when trying to test

> Change a product to items instead of weight or volume.


image


I've tried this on Melon Farm and Suma Whole Foods (Cocoa powder) on Uk staging.

And my latest release test Kimchi Farm 2 on staging FR

luisramos0:
@RachL  when you change to items you must specific the variant name
is that the problem?

luisramos0:
btw PR #4780 will bring meaningful error messages back to this page!

RachL:
@luisramos0 ah indeed this is working if I add a unit name before saving. 🤦‍♀ Moving to ready to go!

PR #4714, opened 2020-01-23
Add require to fix loading issue. Some other Permissions module was being loaded

luisramos0:
#### What? Why?

Fixes problem when loading order cycles locally for some developers due to loading order of classes.

#### What should we test?
No test required, this will not break anything for sure. Could be tested locally by another dev but problably ok to merge and include in the release right away as it will be validated in the release testing process.

#### Release notes
Fix class loading issue that was breaking the Order Cycle list page in some local dev environments.
Changelog Category: Fixed

### DIFF
--- a/app/helpers/order_cycles_helper.rb
+++ b/app/helpers/order_cycles_helper.rb
@@ -1,3 +1,5 @@
+require 'open_food_network/permissions'
+
 module OrderCyclesHelper
   def current_order_cycle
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
We're ignoring these "frozen string" CodeClimate suggestions for now, right?

luisramos0:
no, we are going for it and it must be added to all files possible from now on.
I added it to this file. Maybe we should go for a PR that adds it to most files...

PR #4718, opened 2020-01-24
Make Order Cycle exchange tab summary not count inventory variants that are hidden

luisramos0:
#### What? Why?

Fixes bug detected in #4630
In the order cycle incoming settings page, if the OC is configured to show only coordinator inventory variants, the count of variants (Y) displayed in the tab summary "X / Y selected" was not correct. It was including variants that were hidden in the OC coordinator inventory. The hidden variants we not listed below so the numbers would not match.
In staging UK, the Cambridge Food Hub had an incoming exchange from "Hodmedods MASTER" with 46 hidden variants; the total should be 163 and not 209 as displayed.

Additionally, a bug in the variant counts is fixed "subquery with too many columns", the controller method to count variants needed a select(&:id) on products as mentioned in #4630.

#### What should we test?
OC incoming exchange page should show correct number of variants and exclude variants hidden in the coordinator inventory if that's the OC config.

#### Release notes
Changelog Category: Fixed
Fixed a problem in the OC edit pages where hidden inventory variants were being counted in the summary totals of each exchange.

### DIFF
--- a/app/services/exchange_products_renderer.rb
+++ b/app/services/exchange_products_renderer.rb
@@ -12,6 +12,14 @@ class ExchangeProductsRenderer
     end
   end
 
+  def exchange_variants(incoming, enterprise)
+    variants_relation = Spree::Variant.
+      not_master.
+      where(product_id: exchange_products(incoming, enterprise).select(&:id))
+
+    filter_visible(variants_relation)
+  end
+
   private
 
   def products_for_incoming_exchange(enterprise)
@@ -21,12 +29,16 @@ class ExchangeProductsRenderer
   def supplied_products(enterprises_query_matcher)
     products_relation = Spree::Product.where(supplier_id: enterprises_query_matcher)
 
+    filter_visible(products_relation)
+  end
+
+  def filter_visible(relation)
     if @order_cycle.present? &&
        @order_cycle.prefers_product_selection_from_coordinator_inventory_only?
-      products_relation = products_relation.visible_for(@order_cycle.coordinator)
+      relation = relation.visible_for(@order_cycle.coordinator)
     end
 
-    products_relation
+    relation
   end
 
   def products_for_outgoing_exchange
--- a/app/services/exchange_products_renderer.rb
+++ b/app/services/exchange_products_renderer.rb
@@ -1,3 +1,5 @@
+require 'open_food_network/order_cycle_permissions'
+
 class ExchangeProductsRenderer
   def initialize(order_cycle, user)
     @order_cycle = order_cycle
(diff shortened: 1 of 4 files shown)


app/services/exchange_products_renderer.rb
```diff
@@ -12,6 +12,14 @@ def exchange_products(incoming, enterprise)
     end
   end
 
+  def exchange_variants(incoming, enterprise)
+    variants_relation = Spree::Variant.
+      not_master.
+      where(product_id: exchange_products(@incoming, @enterprise).select(&:id))
+
+    filter_visible(variants_relation)
```
luisramos0: this is the key addition in this PR.

luisramos0:
There was a typo in the controller code. Fixed. Build should be green now.

luisramos0:
rebased to resolve simple conflict.

lin-d-hop:
@RachL @kirstenalarsen 
It would be good to have a product perspective on this.
I find the numbers confusing. On the top line we show Variants selected.
Then when you expand it shows products selected.
Since in this page we can only select variants does it make sense to display the expanded view as number of products?
It does give more information which is good, but it also leaves me scratching my head for a moment while I put the numbers together.
Interested in thoughts.

Screenshot from 2020-01-31 11-04-35

lin-d-hop:
I feel like this has been discussed. Sorry if this was the spec and I missed it. Let me know :-)

RachL:
@lin-d-hop as far as I understood we will end up with counting variants everywhere: https://github.com/openfoodfoundation/openfoodnetwork/issues/4594

This PR just fixes a bug to enable #4630 to work fully

luisramos0:
hey @lin-d-hop sorry, maybe I was not clear.
This PR only resolves a bug in the count of variants (the number in the summary above).
After we merge this one we can test #4630 where we change products loaded to variants loaded and the numbers above and below will match!

luisramos0:
the only thing to test here is really whats written above:
"OC incoming exchange page should show correct number of variants and exclude variants hidden in the coordinator inventory if that's the OC config."

Steps would be:
- Config OC to show inventory only
- Hide a variant in inventory
- See the number change in the OC edit page

luisramos0:
to make it clear, this PR is only blocking 4630 because if we put 4630 in and the numbers of variants above and below dont match, that would be even more confusing for the user.

lin-d-hop:
Tested on UK staging
Hiding and adding products to inventory updates the values in the OC page.
Ready to go!

luisramos0:
nice, thanks!

PR #4719, opened 2020-01-26
Make StripeSCA payment method work with existing credit cards (the ones saved through the Stripe Charges API)

### LINKED ISSUE #4685: Make new Stripe SCA payment method (Stripe Payment Intents API) work with cards already stored through the old api (Stripe Charges API)
## Description
This is a part of #4173 - this will make the new Stripe SCA payment method able to replace the existing Stripe payment method. 

When a user stores a card in the account section, the add card form uses the old stripe charges api to store the card. Currently the new Stripe SCA payment method is not working well with these cards. We need to make this work.

This is what is currently happening in the spree_credit_cards DB right now.

```
gateway_customer_profile_id |  gateway_payment_profile_id
----------------------------+------------------------------
                            | tok_1G1vH8GzkMsqbAZX26ai42HZ -  checkout with **Stripe**  not saving card
 cus_GZ3Pl7DWFo5fpg         | card_1G1vIVGzkMsqbAZXegrNIusw - checkout with **Stripe**  saving card
 cus_GZ3VpwwHg64tVO         | card_1G1vOJGzkMsqbAZX7JaD2M9Q - add card in account section (uses old API)
                            | pm_1G1vKgG0lscsYhYacGoC85jv - checkout **Stripe SCA** not saving card
 cus_GZ3TNw3EjFPk1f         | pm_1G1vMEG0lscsYhYa04bGSd6l - checkout **Stripe SCA** saving card
```
The old api stores tok_ (tokens) for one time payments and when saving a card uses card_ (cards) for the card and cus_ (customer) for the customer.
The new api uses pm_ (payment methods) for one time payments as well as for the card when saving cards. It also uses cus_ (customer) to store the user id on the stripe side.

The one time payments on the old api are not important as they are not reusable. We can ignore the entries with tokens.

I believe all we need to do is to make the new api calls work with the card objects. In some way we need to convert a card_ id into a pm_ id. This could be done as a migration or on the fly when the card is used.

Maybe we can make the payment process convert card_ to pm_ and store thee new pm_ in the database.

We have two use cases:
1 - A customer with a saved card that has been used with the existing Stripe payment method should be able use this same card with the new Stripe SCA payment method. (this includes 2 cases: cards saved on checkout with "old" Stripe payment method and cards added in account through "add a card")
2 - A customer should be able to add a card in the account section and use it with the new Stripe SCA payment method.

luisramos0:
#### What? Why?

Closes #4685
Here we make all cards stored with StripeSCA go to the platform account (which is what happens with the existing Stripe payment method). To use these cards in the Payment Intents API we need to clone these cards to the hubs stripe accounts (the connected accounts) before we use them. This process of cloning will also work for existing cards saved through the Charges API :+1:
So, this PR makes this payment method complete and compatible with the existing cards in the system.

#### What should we test?
First verify that #4672 is working by following it's test guide.
Afterwards, see acceptance criteria for this PR in #4685.

#### Release notes
Changelog Category: Changed
Made new Stripe SCA payment method compatible with credit cards already stored in Stripe. This payment method is now complete and we can now start to migrate all users to this payment method.

### DIFF
--- a/lib/stripe/card_cloner.rb
+++ b/lib/stripe/card_cloner.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+# Here we clone
+# - a card (card_*) stored in a customer in a platform account
+# into
+# - a payment method (pm_*) in a new customer in a connected account
+#
+# This process is used in the migration between using the Stripe Charges API (stripe_connect)
+#   and the Stripe Payment Intents API (stripe_sca)
+#
+# This process can be deleted once all hubs are running on the new stripe_sca method and all cards in the system have been migrated to the new payment_methods
+#   Basically, when all DBs have no card_* values in credit_card.gateway_payment_profile_id
+module Stripe
+  class CardCloner
+    def clone!(credit_card, stripe_account_id)
+      return unless credit_card.gateway_payment_profile_id.starts_with?('card_')
+
+      new_payment_method = clone_payment_method(credit_card, stripe_account_id)
+      new_customer = Stripe::Customer.create({ email: credit_card.user.email },
+                                             stripe_account: stripe_account_id)
+      attach_payment_method_to_customer(new_payment_method.id, new_customer.id, stripe_account_id)
+
+      credit_card.update_attributes gateway_customer_profile_id: new_customer.id,
+                                    gateway_payment_profile_id: new_payment_method.id
+      credit_card
+    end
+
+    private
+
+    def clone_payment_method(credit_card, stripe_account_id)
+      card_id = credit_card.gateway_payment_profile_id
+      customer_id = credit_card.gateway_customer_profile_id
+
+      Stripe::PaymentMethod.create({ customer: customer_id, payment_method: card_id },
+                                   stripe_account: stripe_account_id)
+    end
+
+    def attach_payment_method_to_customer(payment_method_id, customer_id, stripe_account_id)
+      Stripe::PaymentMethod.attach(payment_method_id,
+                                   { customer: customer_id },
+                                   stripe_account: stripe_account_id)
+    end
+  end
+end
--- a/lib/stripe/credit_card_cloner.rb
+++ b/lib/stripe/credit_card_cloner.rb
@@ -1,44 +1,49 @@
 # frozen_string_literal: true
 
 # Here we clone
-# - a card (card_*) stored in a customer in a platform account
+#   - a card (card_*) or payment_method (pm_*) stored in a customer in a platform account
 # into
-# - a payment method (pm_*) in a new customer in a connected account
+#   - a payment method (pm_*) in a new customer in a connected account
 #
-# This process is used in the migration between using the Stripe Charges API (stripe_connect)
-#   and the Stripe Payment Intents API (stripe_sca)
+# This is required when using the Stripe Payment Intents API:
+#   - the customer and payment methods are stored in the platform account
+#       so that they can be re-used across multiple sellers
+#   - when a card needs to be charged, we need to create it in the seller's stripe account
 #
-# This process can be deleted once all hubs are running on the new stripe_sca method and all cards in the system have been migrated to the new payment_methods
-#   Basically, when all DBs have no card_* values in credit_card.gateway_payment_profile_id
+# We are doing this process every time the card is charged:
+#   - this means that, if the customer uses the same card on the same seller multiple times,
+#       the card will be created multiple times on the seller's account
+#   - to avoid this, we would have to store the IDs of every card on each seller's stripe account
+#       in our database (this way we only have to store the platform account ID)
 module Stripe
   class CreditCardCloner
-    def clone!(credit_card, stripe_account_id)
-      return unless credit_card.gateway_payment_profile_id.starts_with?('card_')
+    def clone(credit_card, connected_account_id)
+      new_payment_method = clone_payment_method(credit_card, connected_account_id)
 
-      new_payment_method = clone_payment_method(credit_card, stripe_account_id)
       new_customer = Stripe::Customer.create({ email: credit_card.user.email },
-                                             stripe_account: stripe_account_id)
-      attach_payment_method_to_customer(new_payment_method.id, new_customer.id, stripe_account_id)
+                                             stripe_account: connected_account_id)
+      attach_payment_method_to_customer(new_payment_method.id,
+                                        new_customer.id,
+                                        connected_account_id)
 
-      credit_card.update_attributes gateway_customer_profile_id: new_customer.id,
-                                    gateway_payment_profile_id: new_payment_method.id
-      credit_card
+      [new_customer.id, new_payment_method.id]
     end
 
     private
 
-    def clone_payment_method(credit_card, stripe_account_id)
-      card_id = credit_card.gateway_payment_profile_id
+    def clone_payment_method(credit_card, connected_account_id)
+      platform_acct_payment_method_id = credit_card.gateway_payment_profile_id
       customer_id = credit_card.gateway_customer_profile_id
 
-      Stripe::PaymentMethod.create({ customer: customer_id, payment_method: card_id },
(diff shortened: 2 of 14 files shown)


lib/stripe/card_cloner.rb
```diff
@@ -0,0 +1,44 @@
...
+
+# Here we clone
+# - a card (card_*) stored in a customer in a platform account
+# into
+# - a payment method (pm_*) in a new customer in a connected account
+#
+# This process is used in the migration between using the Stripe Charges API (stripe_connect)
+#   and the Stripe Payment Intents API (stripe_sca)
+#
+# This process can be deleted once all hubs are running on the new stripe_sca method and all cards in the system have been migrated to the new payment_methods
```
mkllnk: I'm reviewing this on Github and need to scroll horizontally which is a bit awkward. Can I have shorter lines, please?

lib/stripe/card_cloner.rb
```diff
@@ -0,0 +1,44 @@
...
+      return unless credit_card.gateway_payment_profile_id.starts_with?('card_')
+
+      new_payment_method = clone_payment_method(credit_card, stripe_account_id)
+      new_customer = Stripe::Customer.create({ email: credit_card.user.email },
+                                             stripe_account: stripe_account_id)
+      attach_payment_method_to_customer(new_payment_method.id, new_customer.id, stripe_account_id)
+
+      credit_card.update_attributes gateway_customer_profile_id: new_customer.id,
+                                    gateway_payment_profile_id: new_payment_method.id
+      credit_card
```
mkllnk: You don't need a return value here, do you?

Once we converted a card to the new API, we can't switch back to Stripe Connect, can we?

spec/lib/stripe/credit_card_cloner_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      let(:new_payment_method_id) { "pm_4567" }
+      let(:stripe_account_id) { "acct_456" }
+      let(:customer_response_mock) { { status: 200, body: customer_response_body } }
+
+      let(:credit_card) { create(:credit_card, user: create(:user)) }
+
+      before do
+        allow(Stripe).to receive(:api_key) { "sk_test_12345" }
+
+        stub_request(:post, "https://api.stripe.com/v1/customers")
```
mkllnk: I'm wondering if it's time for the vcr gem. We used it for Fair Food and it's working really well. It records real API responses to repeat them offline. So you get real data and you are setup to repeat it against the live API whenever you want.

lib/stripe/card_cloner.rb
luisramos0: yes, this was fixed in later commits :+1:

lib/stripe/card_cloner.rb
luisramos0: this was changed in later commits.

This process clones a card in a platform account to the connected account.
This process is backward compatible! Because the original stripe method will take the card in a platform account and create a "token" in the connected account, this method takes the card in the platform account and create another card in the connected account.

spec/lib/stripe/credit_card_cloner_spec.rb
luisramos0: I like this webmock way of testing things but @kristinalim will be looking into testing these things, maybe a good pointer for #4639

lib/stripe/credit_card_cloner.rb
```diff
@@ -0,0 +1,52 @@
...
+#   - a payment method (pm_*) (in a new customer) in a connected account
+#
+# This is required when using the Stripe Payment Intents API:
+#   - the customer and payment methods are stored in the platform account
+#       so that they can be re-used across multiple sellers
+#   - when a card needs to be charged, we need to create it in the seller's stripe account
+#
+# We are doing this process every time the card is charged:
+#   - this means that, if the customer uses the same card on the same seller multiple times,
+#       the card will be created multiple times on the seller's account
```
Matt-Yorkley: I've got a couple of questions, just out of interest:

- What are the consequences of the same card being created multiple times on the seller's account?
- If a card already exists on the seller's account, can't we retrieve it using the Stripe API instead of recreating it?

luisramos0: 1. until now what was created every time on the seller's account was the token, there were as many tokens as payments on these accounts. the stripe dashboard doesnt show the tokens I think. Now we are doing the same for cards, this appears in the dashboard on the customers tab, each time there is a payment there will be a new customer, even if the card has been previously used. I dont think this is a big problem, it's just a detail in the stripe UI for the sellers.
2. yes we can, it's just more code and more work to get that done. I dont think there's much value in implementing that.

filipefurtad0:
Hey @luisramos0,

Performing the first step of the testing notes:
1. make a payment without saving card

Generated the error:

_No such PaymentMethod: pm_1GEbg1LE3KX9OdcFl4vKeSul; OAuth key or Stripe-Account header was used but API request was provided with a platform-owned payment method ID. Please ensure that the provided payment method matches the specified account._ 

Please see testing notes:
https://docs.google.com/document/d/1wQzeV_NB-QaFF5dCrG4bBx-7lve66vc0o1ZRP-SWjQY/edit#

I'll stop testing now, so removing the staging-label.

luisramos0:
🙈 I messed up my local branches between my local linux and my local max.... I was lucky and managed to recover the lost code. It was only one commit to fix exactly the use case you tested:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4719/commits/404e7c1f37127070f5cbf151cc6694c54e316ae1

let's just watch the new build and then stage and test again 👍

luisramos0:
ok, looking good, I restaged in katuma and managed to checkout with stripe on fabulous funcho :-)

filipefurtad0:
Hi @luisramos0,

Awsome! This looks very good indeed. :tada: 

Testing notes:
https://docs.google.com/document/d/1wQzeV_NB-QaFF5dCrG4bBx-7lve66vc0o1ZRP-SWjQY/edit#

I do think there is a small issue with the checking of payment details, of a given order, for the new Stripe SCA. But since this is not the focus of the PR, and all the test cases worked fine, I am moving this to ready to go.

Thanks!

lin-d-hop:
Hiya @filipefurtad0 

Thanks so much for testing this!

I'm a little nervous about introducing regressions. Did you check in with others that these aren't related to this PR? Seems if inspecting individual payments work on prod and not here we might have a problem?

A user facing slug is pretty bad. Would be great to avoid if possible. Though I'm not sure if the pages with slugs are used much (I doubt it).

My personal preference would be not to introduce these slugs. Though if @luisramos0 and/or @RachL are happy then I'm happy.

RachL:
I've hit the same issue in a previous PR concerning some spree backend migration, maybe it is linked?

Worth to have it checked I believe.

filipefurtad0:
@lin-d-hop and @RachL,
Thank you for your feedback on this!
Definitively worth a second look - to avoid having these slugs later on and at least to make sure these are not related to this PR, in particular. Will get back at this.
Thanks again :-)

luisramos0:
great testing @filipefurtad0 
snail 1 is a bug in this PR - no template to show stripeSCA payment details - this template was now added
snail 2 is a pre-existing bug - payment details template could not handle nil credit card (deleted by user for example), it is now able to handle that case, it will show empty credit card details and the payment amount.
Added 2 commits to fix these two problems and one auto spec for each case :muscle: 

Changes are minor, @Matt-Yorkley or @mkllnk could you have a quick look please at these two last commits?
Thanks!

luisramos0:
Thanks @mkllnk 
Moving to Test Ready again.

filipefurtad0:
Hi @luisramos0,

I re-tested all test cases:
- the test-guide from #4672
- and acceptance criteria for this PR, as in #4685

Also checked for both Snail-cases on the test notes.
All good, this is ready-to-go!

luisramos0:
great, thanks Filipe!!!

PR #4722, opened 2020-01-28
Bump Code Climate's Rubocop version to their latest

mkllnk:
#### What? Why?

I noticed that Code Climate is using an older version of Rubocop. That explains why we see different results locally. In this PR I chose the most recent version supported by Code Climate which is closer to our used version.

Code Climate seems to be a bit behind which clashes with our Dependabot updates. Maybe we should not follow Dependabot here and always update rubocop manually. It needs manual intervention anyway. Let's see.

#### What should we test?

No test needed.

#### Release notes

Fine-tuned the configuration of our automated code quality control.

Changelog Category: Changed

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,7 +2,7 @@ version: "2"
 plugins:
   rubocop:
     enabled: true
-    channel: "rubocop-0-57"
+    channel: "rubocop-0-76"
     config:
       file: ".rubocop_styleguide.yml"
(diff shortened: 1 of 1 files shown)


sauloperez:
I think those CodeClimate issues deserve a look we could fix some and some others to the todo.

sauloperez:
I went ahead and approved it in Code Climate. This is already good as it is and we can fix those issues later.

PR #4733, opened 2020-01-29
Fixing incorrect payment and balance calculation

### LINKED ISSUE #4732: Reports Balance Calculation does not match User Balance Calculation
## Description

## Expected Behavior

## Actual Behaviour

## Steps to Reproduce

5. Cringe at the difference.

## Workaround

## Severity

## Your Environment

* Version used: 2.7.1 UK

lin-d-hop:
#### What? Why?

Closes #4732

#### What should we test?

To test:

1. With a shopper account create some orders and payments.
2. As Hub manager make adjustments to the orders and payments. Including voiding payments and orders. Add multiple payments to an order and include addition payments so that the orders also have more payments than the order amount.
3. As you make the changes in step 2 check the Shopper balance in Accounts -> Transactions. Compare this to the Balance and Payment Totals shown in Reports -> Order Cycle Management Reports -> Payment Methods Report.

#### Release notes

Changelog Category: Fixed

#### Discourse thread

First step in 

#### Documentation updates

### DIFF
--- a/lib/open_food_network/order_cycle_management_report.rb
+++ b/lib/open_food_network/order_cycle_management_report.rb
@@ -77,7 +77,7 @@ module OpenFoodNetwork
        ba.phone,
        order.shipping_method.andand.name,
        order.payments.first.andand.payment_method.andand.name,
-       order.payments.first.amount,
+       order.payment_total,
        OpenFoodNetwork::UserBalanceCalculator.new(order.email, order.distributor).balance]
     end
 
@@ -92,7 +92,7 @@ module OpenFoodNetwork
        sa.phone,
        order.shipping_method.andand.name,
        order.payments.first.andand.payment_method.andand.name,
-       order.payments.first.amount,
+       order.payment_total,
        OpenFoodNetwork::UserBalanceCalculator.new(order.email, order.distributor).balance,
        has_temperature_controlled_items?(order),
        order.special_instructions]
--- a/lib/open_food_network/user_balance_calculator.rb
+++ b/lib/open_food_network/user_balance_calculator.rb
@@ -6,25 +6,13 @@ module OpenFoodNetwork
     end
 
     def balance
-      payment_total - completed_order_total
+      -completed_orders.sum(&:outstanding_balance)
     end
 
     private
 
-    def completed_order_total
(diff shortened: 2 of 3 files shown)


lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -130,6 +130,10 @@ def is_payment_methods?
       params[:report_type] == "payment_methods"
     end
 
+    def payments_total(payments)
+      payments_total = payments.sum(&:amount)
+    end
```
lin-d-hop: Assignment not needed. Thanks CodeClimate :-)

lin-d-hop: Ping devs - I have a question here.
This is broken as it is also showing Void payments and Incomplete payments.
I would be very grateful if someone could help me strip out Void and Incomplete payments such that the Payments total is accurate.
Thanks in advance!

luisramos0: you shouldnt do biz logic here, the models do that for you.
Maybe you just need order.payment_total ?

lin-d-hop: That is exactly the kind of wisdom I was looking for... Wisdom to help noobs :-)

lib/open_food_network/user_balance_calculator.rb
```diff
@@ -6,25 +6,14 @@ def initialize(email, distributor)
     end
 
     def balance
-      payment_total - completed_order_total
+      completed_orders.sum(&:outstanding_balance)
```
luisramos0: I think this is correct but youare inverting the sign of the balance.
order.outstanding_balance is the positive valuee that buyer must pay.
this balance method here was "user balance".
order is 30 you have paid 10. order.outstand_balance = 20 this method -20

luisramos0:
```suggestion
      - completed_orders.sum(&:outstanding_balance)
```

RachL:
Nice! This is working well :-)

I noticed that the buyer's profile does not list the payment as "paid" if you paid and capture more money than your order amount.

But this is already in production, so this PR can go. Moving to ready to go and creating an issue.

PR #4742, opened 2020-02-02
Edit new_resource partial

### LINKED ISSUE #4109: Missing translations on new delivery method setup page
## Description

URL: `/admin/shipping_methods/new`

Capture du 2019-07-31 22-15-16

Also, when I change the calculator type, a little sentence appear below, that is untranslated :
Capture du 2019-07-31 22-39-23

myersca1:
#### What? Why?

Closes #4109 

New_resource_links file was referenced in multiple places, but did not exist. The existing .erb file did not work, so a .haml file was created in its place. 

#### What should we test?
Verify that all pages referencing this file work properly with the translations.

Changelog Category: Fixed

### DIFF
--- a/app/views/spree/admin/shared/_new_resource_links.html.haml
+++ b/app/views/spree/admin/shared/_new_resource_links.html.haml
@@ -1,4 +1,4 @@
 .form-buttons.filter-actions.actions
-  = button t(:create), 'icon-ok'
+  = button t('actions.create'), 'icon-ok'
   %span.or= t(:or)
-  = button_link_to t(:cancel), collection_url, icon: 'icon-remove'
+  = button_link_to t('actions.cancel'), collection_url, icon: 'icon-remove'
--- a/app/views/spree/admin/shared/_new_resource_links.html.haml
+++ b/app/views/spree/admin/shared/_new_resource_links.html.haml
@@ -1,0 +1,4 @@
+.form-buttons.filter-actions.actions
+  = button t(:create), 'icon-ok'
+  %span.or= t(:or)
+  = button_link_to t(:cancel), collection_url, icon: 'icon-remove'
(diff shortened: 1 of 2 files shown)


app/views/spree/admin/shared/_new_resource_links.html.haml
```diff
@@ -0,0 +1,4 @@
+.form-buttons.filter-actions.actions
+  = button t(:create), 'icon-ok'
+  %span.or= t(:or)
+  = button_link_to t(:cancel), collection_url, icon: 'icon-remove'
```
luisramos0: I wonder why the change from link_to_with_icon to button_link_to. What is the differrence between these two methods? Will the user notice any difference?

app/views/spree/admin/shared/_new_resource_links.html.haml
```diff
@@ -0,0 +1,4 @@
+.form-buttons.filter-actions.actions
+  = button t(:create), 'icon-ok'
```
luisramos0: Looking at the code now, to fix the issue, all you needed would be to change the translation calls from Spree.t to t. because spree.actions.create and spree.actions.cancel are not there in the en.yml file but both are under actions.create (Spree.t is similar to t. but adds spree in the namespace).

Here you are moving to root level :create and :cancel. I think it's ok, I would prefer to keep these under the actions namespace, so I'd use t('actions.create') and t('actions.cancel') instead of t(:create) and t(:cancel).
The OR can stay as t(:or).

app/views/spree/admin/shared/_new_resource_links.html.haml
blainebillings: The only difference is purely aesthetic - whether or not it looks like a button, as far as I could tell, but maybe I am mistaken.

app/views/spree/admin/shared/_new_resource_links.html.haml
blainebillings: Thank you very much - that makes complete sense. We went ahead and updated with the change.

RachL:
hehe even already on transifex 💪 

image


This is ready to go!

luisramos0:
Congrats @myersca1 🎉

RachL:
ah yes welcome @myersca1 !

Where are my manners 🙈

PR #4743, opened 2020-02-03
Restore ability to load more than 100 products

mkllnk:
The simple order cycle form was not updated to use the new pagination.

#### What? Why?

#### Release notes


luisramos0:
I think there are a couple of problems to be done here, so I want for the easier approach: #4744

If we want to do this we need to:
- add summary count "25 of 130 Variants loaded"
- we would need to make the "select all" button work with paginated results
- we would need to make the simple create OC page work correctly as well
- remove copy pasted code - the code should not be repeated in two angular controllers

I think it will take maybe one day of dev time to handle these things plus testing to get these things right. I dont think this is a priority if 4744 solves the problem.

mkllnk:
I agree with all your points above. But now we are in performance tech-debt again. We should solve it properly. I'll create an issue and then we can prioritise.

mkllnk:
Closing this because the original issue has been solved. We now have some tech-debt: #4755.

luisramos0:
nice, thanks Maikel.

PR #4744, opened 2020-02-03
Make pagination optional in the ExchangeProductsController

luisramos0:
#### What? Why?

This is an alternative to https://github.com/openfoodfoundation/openfoodnetwork/pull/4743
Instead of adding pagination to the simple OC edit page it just make pagination optional on the server and so the simple OC page starts to load all products at once.

#### What should we test?
Simple create and edit OC page should work correctly and load more than 100 products if applicable.

#### Release notes
Changelog Category: Fixed
Fix problem in simple OC edit page. It's now loading over 100 products if available.

### DIFF
--- a/app/controllers/api/exchange_products_controller.rb
+++ b/app/controllers/api/exchange_products_controller.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 # This controller lists products that can be added to an exchange
+#
+# Pagination is optional and can be required by using param[:page]
 module Api
   class ExchangeProductsController < Api::BaseController
-    DEFAULT_PAGE = 1
     DEFAULT_PER_PAGE = 100
 
     skip_authorization_check only: [:index]
@@ -49,6 +50,8 @@ module Api
     end
 
     def paginated_products
+      return products unless pagination_required?
+
       products.
         page(params[:page] || DEFAULT_PAGE).
         per(params[:per_page] || DEFAULT_PER_PAGE)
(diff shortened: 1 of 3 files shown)


app/controllers/api/exchange_products_controller.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 # This controller lists products that can be added to an exchange
+#
+# Pagination is optional and can be required by using param[:page]
 module Api
   class ExchangeProductsController < Api::BaseController
-    DEFAULT_PAGE = 1
```
Matt-Yorkley: This is still used elsewhere in the class, right?

luisramos0: not any more. If no page number is provided in the request, it will not paginate results. That's what the PR changes.

luisramos0: My mistake, there's one occurrence. I need to remote it, it's not needed now.

luisramos0: ok, fixed 👍

app/controllers/api/exchange_products_controller.rb
```diff
@@ -49,8 +50,10 @@ def renderer
     end
 
     def paginated_products
+      return products unless pagination_required?
+
       products.
-        page(params[:page] || DEFAULT_PAGE).
```
Matt-Yorkley: It would be nice to add this back when we can.

spec/controllers/api/exchange_products_controller_spec.rb
```diff
@@ -51,12 +51,27 @@ module Api
         let(:exchange) { order_cycle.exchanges.outgoing.first }
         let(:products_relation) { Spree::Product.includes(:variants).where("spree_variants.id": exchange.variants.map(&:id)) }
 
-        it "paginates results" do
-          spree_get :index, exchange_id: exchange.id, page: 1, per_page: 1
+        before do
+          stub_const("Api::ExchangeProductsController::DEFAULT_PER_PAGE", 1)
```
sauloperez: `stub_const` is legacy and we should use:

```suggestion
          stub_const("Api::ExchangeProductsController::DEFAULT_PER_PAGE", 1)
          allow(
```

luisramos0:
imho I think this is good enough to stay as is. I think it's a valid feature to have optional pagination.

lin-d-hop:
Imported 120 products to a simple producer shop.
Created an order cycle and added all products. 
Went to the shopfront and saw all 120 products in the shopfront.

This is good to go :-)

NOTE
There was a short delay before all the products appeared for selection in the OC. This is not a problem, but might explain why when @RachL tested on Australian staging the OC Edit page did not load any products (total 6000+ products). Apparently this page used to work and no longer does with this number of products. Something to note.

luisramos0:
are you on fosdem? @RachL 's the internet speed will matter for that test with 6k...

RachL:
@luisramos0 yes will try again at home on wednesday, but this is not a blocker.

PR #4759, opened 2020-02-04
Order admin products

### LINKED ISSUE #3909: Sorts product per alphabetical order
## Description
**- As a:** Enterprise user or super admin user
**- On page:** /admin/products
**- I want to be able to do:** See products per alphabetical order.

This user feedback came from Guillaume, see: https://community.openfoodnetwork.org/t/some-potential-ofn-improvements-from-a-small-normandy-producer/1663

## Acceptance Criteria
This issue can be closed if the following tests can be done:

*Scenario 1*
1. Log in as an enterprise user
2. Go to /admin/products
3. The list of product shall appear in alphabetical order with filter or without filters.

*Scenario 2*
1. Log in as a super admin
2. Go to /admin/products
3. The list of product shall appear in alphabetical order with filter or without filters.

pacodelaluna:
Closes #3909 

This change is needed to helper some users to have a clearer view of their catalogue. I decided to allow to sort on Name column instead of forcing the order on Name ASC.

To test, you can go to /admin/products page and click on the Name column to check that your products are getting properly sorted.

#### Release notes
Allow Admin products to be sorted by name

Changelog Category: Added


app/assets/javascripts/admin/bulk_product_update.js.coffee
```diff
@@ -103,9 +106,16 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     $scope.categoryFilter = "0"
     $scope.importDateFilter = "0"
 
+  $scope.$watch 'sortOptions', (sort) ->
+    if sort && sort.predicate != ""
+      $scope.sorting = sort.predicate + ' desc' if sort.reverse
+      $scope.sorting = sort.predicate + ' asc' if !sort.reverse
+      $scope.fetchProducts()
```
luisramos0: this is copy pasted code so I'd extract it, maybe SortOptions itself would be a good place for it?

```ruby
  $scope.$watch 'sortOptions', (sort) ->
       return unless sort && sort.predicate != ""

       $scope.sorting = sort.getSortingExpr()
       $scope.fetchProducts()   
            
and then in sortOptions:
    getSortingExpr: () ->
         sortingExpr = this.predicate + ' desc' if this.reverse
         sortingExpr = this.predicate + ' asc' if !this.reverse
         sortingExpr
```

you could call this from the orders controller as well.

luisramos0:
You decided to add the sorting button right?
It would be good to see a screenshot in the PR and also validate it with someone from Product.

pacodelaluna:
I need to fix tests, especially to setup the default order at the controller side too. Will come with additional patch asap.

pacodelaluna:
To avoid changing the default behaviour, now you can now just filter the products list by name.

Screenshot from 2020-02-05 08-55-15

Tell me if any issue with this.

luisramos0:
ping @lin-d-hop @RachL 
imho I think it should sort alphabetically by name by default.

pacodelaluna:
@luisramos0 You are right, it is pretty clear on the ticket, I will then force the order to `name asc`.

RachL:
@luisramos0 @pacodelaluna @lin-d-hop I'm fine also if we leave it this way first. In order to not confuse people. I really don't know if some of them out there like to have them ordered by date. I only hear from people who want to have them ordered alphabetically. Moving to alphabetical order by default means you can't have them by date at all anymore (expect if you did an import). So with this current solution we have both, that's interesting in some way. So I would be fine to first ship it as is.

pacodelaluna:
I was also thinking that we are using API Products on Admin Products page, and the API call may be used in other places, so to change the default order may have side effects. It is just assumption but safer for now I think.

app/controllers/api/products_controller.rb
```diff
@@ -136,6 +142,10 @@ def render_paged_products(products)
       }.to_json
     end
 
+    def query_params_with_defaults
+      params[:q].to_h.reverse_merge!('s' => 'created_at desc')
```
sauloperez: I'd rather not mutate to avoid nasty side effects with the params so I'd rather use `#reverse_merge` instead of `#reverse_merge!`

pacodelaluna: Done

sigmundpetersen:
There are errors in the last build

pacodelaluna:
Getting random failures in tests, trying to fix this asap...

luisramos0:
if you have changed the default order of the products you will need to adapt the specs.

RachL:
@pacodelaluna @luisramos0 can this be staged and tested meanwhile or should we wait?

luisramos0:
no, it's back to in dev.

pacodelaluna:
Sorry about the mess, got a regression in the middle of my commits and was blind after it (random failures appeared as order was skipped on scope). 
It should be all good now :crossed_fingers:  Tell me if any issue.

luisramos0:
ready for testing again :+1:

RachL:
Alright, this is quite interesting: even if you change the filters or search : the ordering is kept.

I believe it is a really good first option that I can send to users in order for them to decide which default ordering they prefer.

Thank you @pacodelaluna :tada:

luisramos0:
great, thanks @pacodelaluna :tada:

PR #4760, opened 2020-02-04
Merge Spree::CheckoutController with CheckoutController and clean it up

### LINKED ISSUE #2763: [Refactoring] Merge Spree::CheckoutController with our CheckoutController
### Description
As mentioned [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/2366#issuecomment-407663487), we are decorating Spree's CheckoutController and then extending it. This needs to be cleared up.

We should merge https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/spree/checkout_controller_decorator.rb
	and https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/controllers/checkout_controller.rb

### Expected output
We have one single place where all overrides to the Spree Checkout Controller are done. One possibility to be considered is to have our own CheckController without extending or decorating the Spree one. If what's on the Spree side is small enough to be copied to our side.

## Tests
No change in functionality. the checkout on the website should work correctly.

luisramos0:
#### What? Why?

Closes #2763

While working on StripeSCA and adding more logic to the checkout controller I realized it was nonsense to have such a critical part of the OFN app in such a bad state: long controller, extending a useless Spree::CheckoutController, full of logic and with a lot of "complexity" rubocop issues...
Obviously time for a clean up!

This PR merges the controllers as described in #2763 and also fixes all rubocop issues by extracting logic to different methods.
The only remaining rubocop issue is the size of the class: some logic was extracted to 3 separate classes but the class is still 238 lines long.

In the end, I had to keep Spree::CheckoutController only for a route that paypal_express uses. It only redirects to CheckoutController. At some point we can modigy the paypal_express gem to go directly to our CheckoutController (the checkout_state_path [here](https://github.com/openfoodfoundation/better_spree_paypal_express/blob/master/app/controllers/spree/paypal_controller.rb)).

#### What should we test?
We need to test all parts of the checkout process including exceptions:
- Verify simple checkout functionality is working correctly.
- Verify paypal payments and errors during payment are working correctly
- Verify stripe payments and errors during payment are working correctly (including paying with saved cards)
- Verify the order addresses can be successfully saved as default addresses and reused later

#### Release notes
Changelog Category: Changed
Major refactor of the checkout code to make it cleaner, easier, simpler and ready for the future!


app/services/checkout/form_data_adapter.rb
```diff
@@ -0,0 +1,68 @@
...
+    private
+
+    # For payment step, filter order parameters to produce the expected
+    #   nested attributes for a single payment and its source,
+    #   discarding attributes for payment methods other than the one selected
+    def move_payment_source_to_payment_attributes!
+      return unless @params[:payment_source].present? &&
+                    payment_source_params = delete_payment_source_params!
+
+      @params[:order][:payments_attributes].first[:source_attributes] = payment_source_params
```
sauloperez: It is strongly encouraged not to mutate the params object as can lead to potential and hard to spot bugs (I can try to find an article I read about the recently). This is just the same idea of not mutating data shared by multiple objects in a system. I'd rather `#dup` the params and return that one instead.

sauloperez:
This looks awesome but I'll need to give it another read. I'd appreciate if you could point out what you just merged and what you modified other than the abstractions you came up with. It would ease the review process.

luisramos0:
It's 100% refactoring Pau. Absolutely no change (including keeping the mutation of params).

app/controllers/checkout_controller.rb
```diff
@@ -46,26 +46,7 @@ def update
...
-      flash[:error] = if @order.errors.present?
-                        @order.errors.full_messages.to_sentence
-                      else
-                        t(:payment_processing_failed)
-                      end
-      update_failed
-      return
-    end
-    return update_failed unless @order.state == "complete" || @order.completed?
+    return unless checkout_workflow(shipping_method_id)
```
mkllnk: `update_failed` is a rendering method. I personally prefer to have calls to rendering with in action method to see all possible outcomes. If you did `return update_failed unless ...` here, it would simplify the `checkout_worflow` method, reduce duplication and enable us to move that logic to a service.

app/services/checkout/form_data_adapter.rb
luisramos0: ok, I am dupping the params on the initializer so that all work is done of a copy of it 👌

luisramos0: this is now the only change in this PR :-) all other logic should be exactly the same.

app/controllers/checkout_controller.rb
```diff
@@ -153,22 +152,6 @@ def set_default_ship_address
     end
   end
 
-  def check_order_for_phantom_fees
```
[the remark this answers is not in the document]
luisramos0: nice one, thanks for fixing it Maikel 💪

app/controllers/checkout_controller.rb
```diff
@@ -82,11 +82,11 @@ def update
         render json: { path: order_path(@order) }, status: :ok
       end
     end
-  rescue Spree::Core::GatewayError => error
+  rescue Spree::Core::GatewayError => e
```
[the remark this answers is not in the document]
luisramos0: I cant find any references but I swear rubocop made me do this :-)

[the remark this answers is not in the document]
luisramos0: you can see that everywhere in our code we have "=> e" not "=> error"

.rubocop_manual_todo.yml
```diff
@@ -372,7 +372,6 @@ Metrics/AbcSize:
   - app/controllers/spree/admin/taxons_controller.rb
   - app/controllers/spree/admin/users_controller.rb
   - app/controllers/spree/admin/variants_controller.rb
-  - app/controllers/spree/checkout_controller.rb
```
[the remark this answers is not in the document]
luisramos0: yes, just some mess up between commits, it was fixed in previous commits.

app/controllers/checkout_controller.rb
sauloperez: I agree with @mkllnk. Sticking to a Boolean return type in `checkout_workflow` would bring clean separation of concerns BUT this PR is amazing and I prefer to iterate instead.

luisramos0: yes, I see what you mean BUT this change is related to complexity of the method. This method is already in the limit of complexity so I had to move everything to checkout_workflow.

app/controllers/checkout_controller.rb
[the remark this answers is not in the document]
sauloperez: I don't like cryptic var names either but better consistent than not IMO, at least in this case where we know it's an exception.

app/controllers/spree/checkout_controller.rb
```diff
@@ -0,0 +1,8 @@
+module Spree
+  class CheckoutController < Spree::StoreController
```
sauloperez: I think it's worth adding the commit message here as doc for the next dev. Perhaps someone else finds out in the future. Also, the purpose of this controller will become more obvious.

luisramos0: good point! done.

RachL:
Here are my tests results:

- Verify simple checkout functionality is working correctly.

I tested cash, paypal and stripe. All good :)

- Verify paypal payments and errors during payment are working correctly

Errors are all dealt with on the paypal side as far as I know

image


Let me know @luisramos0 if we can test it otherwise? Might be good for release testing as well.

- Verify stripe payments and errors during payment are working correctly (including paying with saved cards)

All good.

image

- Verify the order addresses can be successfully saved as default addresses and reused later

All good!

This looks ready to go :-)

luisramos0:
ah, no, that's it. I think I mentioned it just to see if there was any case I didnt know about :+1:

PR #4766, opened 2020-02-05
Packing reports include customer first name in alphabetical ordering

### LINKED ISSUE #4765: Packing Reports - Order by name does not include first name
## Description

The Packing reports (Customer and Supplier) order the orders by customer name. Unfortunately this include last name but not first name.

This recently broke a script I created to merge the data from two reports into a single report for a user. They suddenly had two customers with the same surname. The other report orders by surname and firstname so to make the script run more quickly I used the ordering to match the data. I tried updating the script but google scripts are slow and timeout quickly so I figured it would be better to just fix the bug in the report.

## Expected Behavior

Packing reports should list orders by LastName then FirstName, along with all the other sorting that happens in the reports.

## Actual Behaviour

Packing reports should list orders by LastName, along with all the other sorting that happens in the reports. If two orders have the LastName then first name ordering is random.

## Steps to Reproduce

1. Create a few orders with the same last name and different first names
2. Go to the packing reports
3. Check the ordering of the orders in both reports. Only last name is used in the sorting.

## Animated Gif/Screenshot

Let me know if you need this. For privacy I'm not sharing the live data and I haven't replicated with fake data. I'll share the live data in a DM on slack.

## Severity

Only really a problem in my scripts.... 

## Your Environment

* Version used: Hot Kimchi

lin-d-hop:
#### What? Why?

Closes #4765

The Packing reports (Customer and Supplier) order the orders by customer name. Unfortunately this include last name but not first name.

This recently broke a script I created to merge the data from two reports into a single report for a user. They suddenly had two customers with the same surname. The other report orders by surname and firstname so to make the script run more quickly I used the ordering to match the data. I tried updating the script but google scripts are slow and timeout quickly so I figured it would be better to just fix the bug in the report.

#### What should we test?

Create a few orders with the same last name and different first names
Go to the packing reports
Check the ordering of the orders in both reports. The reports should now sort alphabetically by first name too.

#### Release notes

Packing reports now include first name in alphabetical ordering.

Changelog Category: Fixed

### DIFF
--- a/lib/open_food_network/packing_report.rb
+++ b/lib/open_food_network/packing_report.rb
@@ -52,7 +52,9 @@ module OpenFoodNetwork
           { group_by: proc { |line_item| line_item.order.distributor },
             sort_by: proc { |distributor| distributor.name } },
           { group_by: proc { |line_item| line_item.order },
-            sort_by: proc { |order| order.bill_address.lastname },
+            sort_by: proc { |order| order.bill_address.lastname } },
+          { group_by: proc { |line_item| line_item.order },
+            sort_by: proc { |order| order.bill_address.firstname },
             summary_columns: [proc { |_line_items| "" },
                               proc { |_line_items| "" },
                               proc { |_line_items| "" },
@@ -88,7 +90,9 @@ module OpenFoodNetwork
          { group_by: proc { |line_item| line_item.full_name },
            sort_by: proc { |full_name| full_name } },
          { group_by: proc { |line_item| line_item.order },
-           sort_by: proc { |order| order.bill_address.lastname } }]
+           sort_by: proc { |order| order.bill_address.lastname } },
+         { group_by: proc { |line_item| line_item.order },
+           sort_by: proc { |order| order.bill_address.firstname } }]
       end
     end
 


filipefurtad0:
Hi @lin-d-hop,

Bug verification:
- As a user: Carried out several purchases - same user, but I changed the first name
- As an Admin: Verified the fact that new orders are not displayed alphabetically, under the Packing reports.

Testing:
- Staged this in Katuma, and proceeded as before:
- As a user: Carried out several purchases - same user, but I changed the first name
- As an Admin: Verified the fact that new orders are not displayed alphabetically, under the Packing reports (below you may find customers AFilipe, BFilipe, ZFilipe, out of order):

Screenshot at 2020-02-19 16-55-41

This does not seem to be the expected result. Please drop me a line if I am missing something - like some obvious way to order them - moving to In_Dev.

lin-d-hop:
Hi @filipefurtad0 
Can you confirm that all of these orders are the same distributor, order cycle etc
If so can you move this back to dev ready.
Thanks

filipefurtad0:
Hi @lin-d-hop,
Thanks for your feedback. Yes, these orders were all made subsequently, the only difference being the first letter of the first name.
Moving back to dev ready.
Cheers!

daniellemoorhead:
@lin-d-hop checking in - think you'll get a revision through on this sometime soon in amongst the masses of customer support and onboarding you're doing? Want someone else to attempt to finish it off?

PR #4776, opened 2020-02-07
Remove the assets group from the gemfile (taken from spree upgrade branch)

luisramos0:
#### What? Why?

The assets group disappears in rails 4, we can do it now.

#### What should we test?
I am not sure but probably just deploy and load some pages in the frontoffice and the backoffice. 

#### Release notes
Changelog Category: Changed
Prepare gemfile  for rails 4 upgrade.


RachL:
This is looking fine :) Moving to ready to go!

PR #4786, opened 2020-02-11
Make StripeSCA (with new authorize/purchase workflow) work with Subscriptions again

luisramos0:
#### What? Why?

Here we complete #4757
After #4757 simple subscription payments with StripeSCA are not working, this PR makes the simple authorization process work.
Next step will be to work with setupIntents to handle cases where the authorization process will require further steps from the customer.

This PR includes a really good refactoring that breaks the PaymentSetup class into 2 separate classes PaymentSetup (for all payments) and StripePaymentSetup (only for stripe payment methods). Some hard work involved in removing tests to private methods and mocking of private methods of the class under test...

This PR adds the StripeScaPaymentAuthorize class that is used from the subs confirm job as well as from the checkout process in the StripeRedirect step.

#### What should we test?
We need to validate the checkout process with StripeSCA payment methods as in #4757 because of the simple change to the StripeRedirect process. This should all be fine.
And then the major test will be to verify that payments with StripeSCA in subscriptions are now working as long as the cards used do not require extra auth steps.

#### Release notes
Changelog Category: Fixed
StripeSCA payment method now works for subscriptions as well.

### DIFF
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -57,10 +57,13 @@ class SubscriptionConfirmJob
     return true unless order.payment_required?
 
     setup_payment!(order)
-    return false if order.errors.present?
+    return false if order.errors.any?
+
+    authorize_payment!(order)
+    return false if order.errors.any?
 
     order.process_payments!
-    return false if order.errors.present?
+    return false if order.errors.any?
 
     true
   end
@@ -70,6 +73,12 @@ class SubscriptionConfirmJob
     return if order.errors.any?
 
     OrderManagement::Subscriptions::StripePaymentSetup.new(order).call!
+  end
+
+  def authorize_payment!(order)
+    return if order.subscription.payment_method.class != Spree::Gateway::StripeSCA
+
+    OrderManagement::Subscriptions::StripeScaPaymentAuthorize.new(order).call!
   end
 
   def send_confirmation_email(order)
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -67,6 +67,9 @@ class SubscriptionConfirmJob
 
   def setup_payment!(order)
     OrderManagement::Subscriptions::PaymentSetup.new(order).call!
+    return if order.errors.any?
+
+    OrderManagement::Subscriptions::StripePaymentSetup.new(order).call!
   end
 
   def send_confirmation_email(order)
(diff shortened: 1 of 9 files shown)


luisramos0:
I think the blocked label was misguiding reviewers to skip this PR. This is only blocked for testing, not for review.

app/jobs/subscription_confirm_job.rb
```diff
@@ -57,10 +57,13 @@ def process_payment!(order)
     return true unless order.payment_required?
 
     setup_payment!(order)
-    return false if order.errors.present?
+    return false if order.errors.any?
```
mkllnk: Why do you prefer `any?` over `present?` here?

luisramos0: I change it here just for consistency with the calls in the same method.

luisramos0:
thanks for your review Maikel!

filipefurtad0:
Hey @luisramos0,

In the shopfront, validated that regular payments using
- Paypal
- StripeConnect
- Stripe SCA with authentication, without saving card, saving card previously and adding a card in account and using it.

In regards to the changes in this PR I tested newly created subscriptions, with 3 different settings, namely payment methods respectively using the payment methods below:
- Cash
- Stripe-Connect
- Stripe SCA

A condition for this test is adding a card with no further auth. requirements, like the 4242424... in the user settings, and allowing charges from the respective shop in which the subscriptions were created.

Tested each run of the subscription by following the procedure described in:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Testing-Subscriptions 
(which is really great btw :rocket: )
 
-> All emails arrived as expected. A particular notice was taken that the order placement/confirmations showed the correct payment method, and values from each purchase.

-> The payments were verified in the Stripe-Dashboard: they appear as "Succeded", with the respective values.

-> The checkouts appear on the customer's transactions, with the respective values.

I'd say this is working perfectly - amazing! :tada: 
Ready to go.

PS: just for completeness sake, I ran the subscription a 4th time, with Stripe-SCA and a 3D card requiring authentication. All fine, except the payment, which failed as expected.

luisramos0:
awesome, thanks a lot Filipe. Really happy to see the rails console commands are helping.

From todays issue, I think we need to include "taking stripe payments in the BO" as part of these test plans :+1: for next time and certainly for the next release testing.

filipefurtad0:
Yes indeed @luisramos0, the commands are most useful.
And I fully agree, I'll include the BO in the upcoming release testing for Stripe testing. Also, I'm adding these standard Stripe-tests to our Testing Handbook.
Thanks again for all the help!

PR #4787, opened 2020-02-12
Add new domain/engine Catalog

luisramos0:
#### What? Why?

After Web and OrderManagement the next engine/domain to be created is Catalog where we can move the ProductImport code into to start with.

This PR adds the Calalog engine and moves on ProductImport class. This class is just an example that can be followed to move all the Product Import code here. Briefly discussed in #3438

I am ignoring the Frozen string literals warnings on code climate, it's too much work to keep up with these...

#### What should we test?
A green build should be fine but let's make a simple sanity check of product import that will validate the backoffice is working and product import isnt broken with a new moved file.

#### Release notes
Changelog Category: Changed
Created the 3rd engine/domain in OFN: Catallog where the part of the app that handles Products, Variants, OrderCycles and Product Import will be :muscle:


RachL:
This is weird, I can't see why this is not working:

image


This is my upload file: https://framadrop.org/r/br8l6rRuam#UWmwIxMAC7zGQGILGmBmuIyov+2au1vwZv0Q+oMRmTY=

Any idea @luisramos0 ?

sigmundpetersen:
Hey @Rachel there was a mention in #4871 about `fr-staging` and prod import

> Ok, I think this is a pre-existing issue and it only applies to French staging. There's a different config option that's set there, which means all products must have a Tax Category in order to be valid. If you fill in the tax_category column the import is successful. It's odd that no validation message is shown though.

I guess that's it? 
ping @luisramos0

luisramos0:
yes, cool be @sigmundpetersen 
this needs to be retested with that in mind?

RachL:
This was indeed the root cause! Moving to ready to go

PR #4799, opened 2020-02-18
4771 Back to shop button

### LINKED ISSUE #4771: [mobile ux] On edit cart page the continue shopping button and the edit order button on the orders page takes you to the shop tab
This can be merged straight to master 🎉 

## Description

**- As a:** shopper
**- On page:** edit cart and also on the user account orders page
**- I want to be able to:** go back to the shop tab when I click 'Continue shopping' or 'Edit order'

**Current behaviour**
- continue shopping button defaults to the home tab
- If home tab does not exist it goes to the shop tab

**Expected behaviour**
- continue shopping button always defaults to the shop tab

mkllnk:
#### What? Why?

Closes #4771

Since the default shop page is now showing the shopfront message instead of the products, the "Continue shopping" buttons were leading there. It would make more sense though to go straight to the products in two cases:

- On the cart page.
- On the edit order page.

#### What should we test?

- Add a shopfront message to a shop.
- Change the shops settings to allow for editing orders.
- Go to the shop and verify that you are landing on the home tab.
- Add products to the cart.
- Go to the cart and click "Continue shopping".
- You should end up on the products page (not home tab).
- Checkout out and edit the order.
- Click on "Continue shopping" and you should get to the products again.
- Clicking on the "Shopping@..." button in the main navigation menu should take you to the home tab though.

#### Release notes

Fixed: The "Continue shopping" button takes you back to the products again even if the shop has a home tab.

Changelog Category: Fixed

### DIFF
--- a/app/helpers/shared_helper.rb
+++ b/app/helpers/shared_helper.rb
@@ -16,4 +16,8 @@ module SharedHelper
   def admin_user?
     spree_current_user.andand.has_spree_role? 'admin'
   end
+
+  def current_shop_products_path
+    "#{main_app.enterprise_shop_path(current_distributor)}#/shop"
+  end
 end
--- a/app/views/spree/orders/edit.html.haml
+++ b/app/views/spree/orders/edit.html.haml
@@ -22,7 +22,7 @@
     - if @order.line_items.empty?
       %div.row{"data-hook" => "empty_cart"}
         %p= t(:your_cart_is_empty)
-        %p= link_to t(:continue_shopping), main_app.shop_path, :class => 'button continue'
+        %p= link_to t(:continue_shopping), current_shop_products_path, :class => 'button continue'
 
     - else
       %div{"data-hook" => "outside_cart_form"}
--- a/app/views/spree/orders/form/_cart_links.html.haml
+++ b/app/views/spree/orders/form/_cart_links.html.haml
@@ -1,6 +1,6 @@
 .row.links{'data-hook' => "cart_buttons"}
   .columns.large-8{"data-hook" => ""}
-    %a.button.large.secondary{href: main_app.shop_path}
+    %a.button.large.secondary{href: current_shop_products_path}
       %i.ofn-i_008-caret-left
       = t :orders_edit_continue
   .columns.large-4.text-right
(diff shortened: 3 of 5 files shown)


sigmundpetersen:
Hey @JenRSmith I see you deployed this to `es-staging`. For future reference, if you want to avoid anything other to be deployed over it you can add the `pr-staged-es` label to the PR. (I added it for you on this one).

daniellemoorhead:
@mkllnk sitting with @JenRSmith and there's a couple of things that need fixing:

- [ ] When I am editing my order the button on that edit cart screen is actually not 'Continue shopping', it's 'Back to store' and is blue. It also takes me to the home tab, not the shop tab. I don't have a problem with the blue button and the copy on that button (I assume it was set up that way for a reason) however the link needs to take the user to the shop tab, not the home tab.

- [ ] When I click 'edit' on my orders page as a shopper, if I'm not in that shop the 'Back to store' button **does not** display - it should. It **does** display if I'm currently shopping in that shop. Recording of this here: https://www.loom.com/share/b54661024fb04a93bfc7bf27200684fd

All the other tests above passed correctly, it works fine when you're in the middle of doing the order itself, it's only when you're editing the order that these problems show up.

There's also a new bug found where the edit option disappears on mobile, you can only cancel. Will log this separately to be fixed as part of the edit cart work.

mkllnk:
@daniellemoorhead 
* Point 1: No problem, I will change the destination.
* Point 2: This is a bit more tricky. The reason it doesn't display the button is because you are shopping somewhere else and if you would go to the order's shop, you would lose your cart of the other shop. I think that this is the best solution as long as we can't have multiple carts (one for each shop). Do you have a better idea?

daniellemoorhead:
As discussed on the phone @mkllnk the only change required for the second point is to display the blue button if the user is not in a different shop (ie. "shop-less").

Ping @JenRSmith so that you know what the decision was when you come to retest this :)

mkllnk:
@JenRSmith This is ready for testing again.

JenRSmith:
@daniellemoorhead regarding the second point, it doesn't appear possible to become shop-less. I discussed this with @mkllnk & tried to become shop-less by closing the order cycle, but this still displays the current shop in the shopping@ field and the 'Return to Store' goes to the correct place - the shop. 

Figured this was the correct behaviour & all other tests passed. Testing notes here - https://docs.google.com/document/d/1IrOMnqMcS7gq4QmjpWNnJhi77PMgp4liTd6_9JAvgWI/edit#

Issue #2247, opened 2018-04-24
Delete old sessions from the database


mkllnk:
We are storing session data in the database, but never delete them. Our database fills up with useless session data.

## Description

We use https://github.com/rails/activerecord-session_store which is part of the Rails 3 core.

## Expected Behavior

Sessions older than a month should get deleted.

## Actual Behavior

Session data persists forever even if the user's browser doesn't have the session id any more.

## Steps to Reproduce

1. Log into a long running instance database.
2. Run: `select updated_at from sessions limit 5;`
3. Observe dates that are years old.

```
openfoodnetwork=> select updated_at from sessions limit 5;
         updated_at         
----------------------------
 2015-06-11 02:48:17.72724
 2015-06-11 02:48:18.089722
 2015-06-11 02:48:22.233834
 2015-06-11 02:48:22.443063
 2015-06-11 02:48:26.397765
(5 rows)
```

## Context

I was just about to implement the session store in another project and came across `rake db:sessions:trim`. That rake task is not included in our version of Rails and we never clean up sessions.

## Possible Fix

Check how https://github.com/rails/activerecord-session_store implements the session trimming. Maybe it's just including the gem and adding a task to our cron schedule.

enricostano:
One day we should probably investigate if switching to an in-memory DB for sessions (Redis, etc) would be a good approach for us.

In any case we can check right now how this is affecting Postgres performance due to both accesses and size of the table.

mkllnk:
Yeah, one day. But for now I'm very reluctant to add more complexity to our setup.

How do you check the affect on performance? Benchmark a query, delete all old entries and benchmark again?

enricostano:
You can ask Postgresql and @sauloperez :joy:

sauloperez:
Setup Skylight and monitoring infrastructure and then we can speak again :trollface: .

Jokes aside, my bet is that this will hardly be a problem for our DBs any time soon. As long as we have an index for the appropriate column/s and we don't end up with 100M rows in a month the reads will still be fast enough.

mkllnk:
> @mkllnk what's the problem you're experiencing? Disk space? DB performance?

Nothing I could measure. I just thought it's a possible optimisation, nothing urgent.

luisramos0:
s5 should be in all the things right @sigmundpetersen ?

luisramos0:
keeping this as tech debt only.

mkllnk:
[...]

PR #4809, opened 2020-02-20
redirect to shops list when an enterprise is not found

### LINKED ISSUE #4727: User should not see a snail, when accessing OFN via direct link to a shop that does not exist or is inactive
## Description
We have over 8k in AUS and 2.5 in UK events of this type:
https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5dd433e559edc60018caf959?filters[event.since][0]=30d&filters[error.status][0]=open

Screenshot 2020-01-28 at 15 22 59

Users type the correct shop URL but the shop name refers to a non existing shop or an inactive shop. The user sees a snail (and a bugsnag event is created).
This damages reputation and seo.

## Expected Behavior
The user can be redirected to the list of shops. Ideally an flash message is displayed "The shop you are looking for doesn't exist or is inactive on OFN. Please check other shops."

## Actual Behaviour
Snail.

## Steps to Reproduce
Inexistent enterprise:
https://openfoodnetwork.org.uk/theres-no-place-like-home/shop

Inactive/not visible shop:
https://openfoodnetwork.org.uk/pomeroys-rare-breeds-robyns-famhouse/shop
Create a shop and set it so that it's not visible in the frontoffice then visit it's shopfront using it's permalink.

## Workaround
None.

## Severity
We are talking about hundreds of (potential) customers seeing the OFN snail page.

## Your Environment
* Version used: UK and AU live
* Browser name and version: firefox
* Operating System and version (desktop or mobile): osx catalina

## Possible Fix
We just need a redirect to the shop list page and flash message. A few lines of code on the controller.

jeduardo824:
#### What? Why?

This PR closes #4727. I'm just rescuing the exception when the enterprise is not found and redirecting to the shops_path with a flash message.

#### What should we test?

If the user is redirected to the shops_path with a flash message when the enterprise not exists or is unavailable.

#### Release notes

Redirect to list of shops when an enterprise is not found or is unavailable.

Changelog Category: Changed

#### Discourse thread

#2247

### DIFF
--- a/spec/controllers/enterprises_controller_spec.rb
+++ b/spec/controllers/enterprises_controller_spec.rb
@@ -139,4 +139,18 @@ describe EnterprisesController, type: :controller do
       expect(response.status).to be 409
     end
   end
+
+  context "checking access on nonexistent enterprise" do
+    before do
+      spree_get :shop, id: "some_nonexistent_enterprise"
+    end
+
+    it "redirects to shops_path" do
+      expect(response).to redirect_to shops_path
+    end
+
+    it "shows a flash message with the error" do
+      expect(request.flash[:error]).to eq(I18n.t(:enterprise_shop_show_error))
+    end
+  end
 end
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -63,8 +63,6 @@ class EnterprisesController < BaseController
   end
 
   def reset_order
-    distributor = Enterprise.is_distributor.find_by_permalink(params[:id]) ||
-                  Enterprise.is_distributor.find(params[:id])
     order = current_order(true)
 
     reset_distributor(order, distributor)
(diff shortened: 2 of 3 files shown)


jeduardo824:
I tried to extract this logic to a method but, unfortunately, I got a type mismatch error
Screen Shot 2020-02-19 at 21 23 42

I also tried to call `and return` after the redirect_to, but I got an error of too many redirects.

luisramos0:
ah, I see, it's a problem when the error occurs.
First idea I had was to move back to use the generic rescue at controller level, that way you wont get too many redirects problem I guess.
But maybe the best solution that will also work is to move the rescue to the end of the shop action "def shop"

jeduardo824:
@luisramos0 What do you think about it?

spec/controllers/enterprises_controller_spec.rb
```diff
@@ -139,4 +139,18 @@
...
+    before do
+      spree_get :shop, id: "some_nonexistent_enterprise"
+    end
+
+    it "redirects to shops_path" do
+      expect(response).to redirect_to shops_path
+    end
+
+    it "redirects to shops_path" do
+      expect(request.flash[:error]).to eq(I18n.t(:enterprise_shop_show_error))
```
luisramos0: you can move this expect to the previous test or give this test a better name related to showing an error

jeduardo824: Sorry. Cmd C + Cmd V problems hahaha.

jeduardo824: All good now.

RachL:
Cool!

image


This is ready to go :)

sauloperez:
So happy to merge this one! thanks a lot @jeduardo824 ! I believe this should stop all those annoying 404 Bugsnag notifications.

jeduardo824:
Awesome! I'm here to help!

PR #4812, opened 2020-02-20
Change Result of PriceSack Calculation from Integers to Floats

### LINKED ISSUE #4654: 'Price Sack' Shipping does not allow non-integer figures
## Description

The Price Sack fee calculator almost works. It allows orders under a certain amount to be charged one value and orders over that amount to be charged another. This is very useful for offering free shipping for orders over x amount etc.

When using the calculator it accepts values that are non-integer in the Normal Amount adn Discount Amount fields. however at checkout the figure is rounded down. £3.99 becomes £3.

## Expected Behavior

These fields should allow non-integer numbers eg £3.50 and this should follow through to checkout.

## Actual Behaviour

These fields allow non-integer numbers eg £3.50 but this does not follow through to checkout.

## Steps to Reproduce

1. Create a shipping fee with Price Sack calculator
2. Enter a non-integer amount to the Normal Amount or Discount Amount fields
3. Create a basket with a value appropriate to the amount entered in the minimal amount field to trigger the desired fee
4. Go to checkout and find the fee rounded down.

## Animated Gif/Screenshot

Screenshot from 2020-01-07 14-24-12

Screenshot from 2020-01-07 14-24-20

## Workaround

Use only integer values on this calculator

## Severity

bug-s3: non-critical but makes the feature significantly less usable. For fees the difference of a whole pound can be significant so the workaround is not really acceptable. 

## Your Environment

* Version used: 2.6.8 UK

blainebillings:
#### What? Why?

Closes #4654

PriceSack calculator would truncate the order cost at checkout. The change required editing the method call of "to_i" on the cost to "to_f" so that they would be treated as floats rather than integers.

#### What should we test?
While we are now seeing floats on the checkout page, I am not exactly sure what the expected PriceSack calculation is and how it should be working (would you be willing to provide a quick explanation?). Please test to make sure that the prices are as expected, and if there are any further errors, I will gladly jump on them.

#### Release notes
The only code that was edited was changing two calls of "to_i" to "to_f" in the price_sack_decorator.rb file.

Changelog Category: Fixed


luisramos0:
it should fairly easy to add a spec to spec/models/spree/calculator/price_sack_spec.rb covering this  change

blainebillings:
Just added a unit test - never used spec before, so please let me know if there is something I should change to make it how it should look. The tests passed when run, so hopefully it works correctly.

spec/models/spree/calculator/price_sack_spec.rb
```diff
@@ -8,6 +8,13 @@
     calculator.preferred_discount_amount = 1
     calculator
   end
+  let(:floatCalculator) do
```
luisramos0: this needs to be float_calculator

spec/models/spree/calculator/price_sack_spec.rb
```diff
@@ -16,13 +23,23 @@
     it "uses the preferred normal amount" do
       expect(calculator.compute(line_item)).to eq(10)
     end
+    context "and preferred normal amount is float" do
```
luisramos0: instead of floatCalculator I think you can just do here:
before do
calculator.update_attribute :preferred_normal_amount, 10.4
end

luisramos0:
the spec is ok :+1: 
the alternative would be, instead of mixing the existing tests with the float tests, to create a separate context for the float case

spec/models/spree/calculator/price_sack_spec.rb
blainebillings: Hmm... doing this seems to be giving the error "PG::NotNullViolation: ERROR:  null value in column "calculable_id" violates not-null constraint".

blainebillings:
That makes sense - I can go ahead make the change if that would be preferable instead.

spec/models/spree/calculator/price_sack_spec.rb
luisramos0: oh, that's probably because the object is not ready to be persisted (update_attribute persists the object).
so you could try in memory:
before do
calculator.preferred_normal_amount = 10.4
end

spec/models/spree/calculator/price_sack_spec.rb
sauloperez: +1

daniellemoorhead:
Hi @blainebillings, just checking in on how you're going with this? It's been almost a months since last activity, do you think you'll get a revised PR through or shall we move this back to Dev Ready to potentially be picked up by someone else?

daniellemoorhead:
Closing for now.

luisramos0:
this one is basically done @daniellemoorhead shall I finish it? it's a few minutes of work.

luisramos0:
I cant push to this branch, I created a new PR with these commits and 2 more from me: #5251

PR #4813, opened 2020-02-20
If there's only one OC available to the customer because there are tag rules hidding other OCs, select that OC by default

### LINKED ISSUE #3170: Limit the available order cycles to select from for tagged shoppers
## Description

An order cycle has been tagged for specific customers for a shopfront that uses registered only access.
Most shoppers will only have one order cycle that they can choose from, however these shoppers still have to click the order cycle selection box and choose the only available order cycle.

## Expected Behavior

Users should only see the available cycles

## Actual Behavior

Users see an initial empty box and have to click down to the select the one available option

## Steps to Reproduce

1. Create an order cycle
2. Tag customer and then tag the order cycle so only one should be visible
3. Go to the shop front

## Animated Gif/Screenshot

image

## Context

Users may think that there is no order cycle available for them - a particular issue for new users

## Severity

## Your Environment

* Version used:
* Browser name and version:

luisramos0:
#### What? Why?

Closes #3170 
This is a quick fix to the issue.
Several tech debt problems in the code were left for later:
- TagRuleApplicator changing a relation with reject! and making #size return a different value from #count :see_no_evil:  I added comments to make it more explicit.
- a big and unnecessary filter mess in EnterrpisesController and BaseController with filters resetting the order and then the set_order_cycles filter... this could be a simple list of actions performed by code in extracted services.

#### What should we test?
We need to verify the issue is fixed and that OC selection is correct when we jump between shops in the frontoffice.

#### Release notes
Changelog Category: Fixed
Fixed problem in shopfront where users had to select the only available OC (no default selections) if in the system there was one or more OCs the user couldn't see.

### DIFF
--- a/app/services/shop/order_cycles_list.rb
+++ b/app/services/shop/order_cycles_list.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# Lists available order cycles for a given customer in a given distributor
+
+class OrderCyclesList
+  def initialize(distributor, customer)
+    @distributor = distributor
+    @customer = customer
+  end
+
+  def call
+    order_cycles = OrderCycle.with_distributor(@distributor).active
+      .order(@distributor.preferred_shopfront_order_cycle_order)
+
+    applicator = OpenFoodNetwork::TagRuleApplicator.new(@distributor,
+                                                        "FilterOrderCycles",
+                                                        @customer.andand.tag_list)
+    applicator.filter!(order_cycles)
+
+    order_cycles
+  end
+end
--- a/app/services/shop/order_cycles_list.rb
+++ b/app/services/shop/order_cycles_list.rb
@@ -1,22 +1,23 @@
 # frozen_string_literal: true
 
 # Lists available order cycles for a given customer in a given distributor
+module Shop
+  class OrderCyclesList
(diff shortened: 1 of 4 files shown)


app/services/shop/order_cycles_list.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+# Lists available order cycles for a given customer in a given distributor
+module Shop
+  class OrderCyclesList
```
sauloperez: pretty cool what you are lately doing of extracting even more abstractions :clap: :clap:

filipefurtad0:
Hey @luisramos0,
I see this branch is currently having conflicts. Once they are resolved I'll test this PR.
Cheers!

luisramos0:
done :+1:

filipefurtad0:
Hi @luisramos0,

Verified the existing bug, which is solved by staging this PR. The available OC is now selected by default.

Checked if all is ok with the display of other OC's, on other shopfronts.

Looks great -> Ready to go!

RachL:
I know a few hubs who will be super exited by this one ❤️

luisramos0:
yes, a great example of the virtues of the papercut process :tada:

PR #4814, opened 2020-02-20
Add new cops (disabled) to rubocop config

luisramos0:
#### What? Why?

The new version of rubocop comes with new cops we need to add to config and disable for now. We can add them later if we want, right now, I dont think we should spend time here.

#### What should we test?
No tests needed.

#### Release notes
Changelog Category: Changed
Updated rubocop config.


sauloperez:
I'm taking the lead and merging this so we reduce the list of PRs in code review. If someone disagrees, please open a PR.

PR #4816, opened 2020-02-21
1860 Remove style from pasted text for shopfront messages

### LINKED ISSUE #1860: [mobile ux] Copy-paste in shopfront message display strange white strips
This can be merged straight to master 🎉 

## Description
When copy-pasting from another shopfront or any text file, into the "message when shopfront open/close" (in menu enterprise setting > shop preferences), here is what happens. Some shops have done that apparently and then their shopfront is not looking good.

bug

mkllnk:
#### What? Why?

Closes #1860

When you edit the shopfront message and paste text or HTML into the editor, it preserves the formatting. It also preserved colours including backgrounds. That could lead to ugly messages. We now keep the basic formatting but remove all additional styles like colours and fonts.

Before:
Screenshot from 2020-02-21 14-16-46
After:
Screenshot from 2020-02-21 14-17-03

I also upgraded the editor library.

I did *not* change existing messages. Shopfront messages that already have weird formatting will keep their formatting. The solution is to go to edit the enterprise, select the whole text and click the clear formatting button, then save.

#### What should we test?

- Go to edit an enterprise and edit the shopfront message and the shop closed message.
- Test basic functionality of the editor to make sure that the update didn't break anything.
- Test with different browsers (I'm not sure what our requirements are).
- Paste plain text and verify that nothing is lost.
- Paste colourful text, including background colours and make sure that the colours disappear.
- Check that pasted bold text and links are preserved.
- Save your result and and verify the display in the shopfront.

Check that other text fields using the same editor code work as before:

- Enterprise about text
- Enterprise group about text
- New product form
- Edit product form

#### Release notes

Changed: When you edit the shopfront message and paste text, it preserves the formatting but strips additional styles like colours and fonts.

Changelog Category: Changed


daniellemoorhead:
Good job :)

app/assets/javascripts/admin/all.js
```diff
@@ -19,7 +19,10 @@
 //= require angular
 //= require angular-resource
 //= require angular-animate
-//= require angular-sanitize
+// angular-sanitize is replaced by textAngular-sanitize further down.
+// We should include only one.
+// https://github.com/textAngular/textAngular#where-to-get-it
+// require angular-sanitize
```
luisramos0: This is removing the dependency so I think you should remove all these lines here including the comment and just add a simple line comment above the textangular-sanitize require:

```
// this replaces angular-sanitize
//= require textAngular-sanitize.min.js

```

luisramos0: the reference to "We should include only one." and link would go to the commit message.

mkllnk: Good idea. I added a commit with that. :heavy_check_mark:

mkllnk:
@luisramos0 
> ideally we should be doing these upgrades in a dependencies list.

Yes. I didn't find a gem for it. We could start using another tool like bower but I'm not familiar with it and it's out of scope of this PR.

> It looks like you have upgraded textAngular, textAngular-rangy and textAngular-sanitize. From what versions did you upgrade from/to exactly?

I have no idea from which version I upgraded. That was one of my reasons to upgrade. The commit message now contains the new version number: v1.5.16

> and how did you find the code for the new versions?

The comment contains a link to the Github repository. I went to releases, downloaded the zip file and copied the minified files over.

> This upgrade means we shoul verify all places where textAngular is used, right?

Yes, I'll add that to the test list.

luisramos0:
gems for javascript are not always up to date and the list of naturally not so complete as if we use bower or other js package manager.
I'd agree this change is out of scope for this PR but I think you are saying we just need to remove the texangular js code and add this to the gemfile:
gem 'text-angular-rails', '~> 1.5', '>= 1.5.16'

luisramos0:
@sauloperez are you ok with this? can we move to test ready?

sauloperez:
IMO it'd be much better if we used a package manager (preferably one that takes the asset pipeline into account) but we're not making our situation worse than it was and we are fixing a bug so :+1:

daniellemoorhead:
**TESTING NOTES**

I pasted some green highlighted copy to a shop on AU staging (Carrot Farm) and the highlight did not display in the text field.

What I did notice, however, was that it pasted in bold even though the font in the Google Doc was not bold, it was normal.

So I tested this on another staging server, and it did the same thing. Thus, it's an existing bug and not a problem for this particular change.

@mkllnk this change is therefore good to go 🎉

PR #4821, opened 2020-02-22
[Spree 2.1] Fix creation of enterprises/users in specs

### LINKED ISSUE #4646: [Spree 2.1] Enterprise Role saving errors
It looks like `EnterpriseRole`, or possibly just the factory that creates them when creating enterprises with users is broken.

This is causing a large number of failing specs, including most of the remaining API test failures.

When an enterprise is created with user objects attached, we get this error (whilst the factory is saving the enterprise object):

```
622) Spree::Admin::SearchController Distributor Enterprise User searching for known users when search query is not an exact match returns a list of users that I share management of enteprises with
       Failure/Error: let!(:enterprise) { create(:enterprise, owner: owner, users: [owner, manager]) }
       
       ActiveRecord::RecordInvalid:
         Validation failed: Enterprise roles is invalid, Enterprise roles is invalid
       # ./spec/controllers/spree/admin/search_controller_spec.rb:9:in `block (3 levels) in <top (required)>'
```

A test with this line will fail: 

```
let!(:enterprise) { create(:enterprise, owner: user1, users: [user1, user2]) }
```

But for some reason this will pass, and allow the test to continue:

```
let!(:enterprise) { create(:enterprise, owner: user1) }
before { enterprise.users << [user1, user2] }
```

luisramos0:
#### What? Why?

Closes #4646

We have to move the definition of the association to after creation in the factory.

#### What should we test?
The error "Enterprise roles is invalid" is gone from the build.

#### Release notes
No release notes for upgrade isssues.

### DIFF
--- a/app/controllers/api/enterprise_attachment_controller.rb
+++ b/app/controllers/api/enterprise_attachment_controller.rb
@@ -1,3 +1,5 @@
+require 'api/admin/enterprise_serializer'
+
 module Api
   class EnterpriseAttachmentController < BaseController
     class MissingImplementationError < StandardError; end
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -93,9 +93,9 @@ class Enterprise < ActiveRecord::Base
   validate :enforce_ownership_limit, if: lambda { owner_id_changed? && !owner_id.nil? }
 
   before_validation :initialize_permalink, if: lambda { permalink.nil? }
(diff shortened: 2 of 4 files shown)


luisramos0:
I had to change the last commit re images to use update_attributes
enterprise.logo = proxy.logo was not persisting the connection between the image and the enterprise.
This fixes 10 specs in spec/controllers/api/logos_controller_spec.rb and spec/controllers/api/promo_images_controller_spec.rb

PR #4822, opened 2020-02-22
WIP Add strong params permit to subs controller and adapt to new mass_assignemnt logic in rails 4

### LINKED ISSUE #4820: [Spree 2.1] Fix problem with shipment.shipping_method in subscriptions (90 specs)
## Description
There are 90 broken specs all related to subs with the following error on shipment validation:
```
 60) OrderSyncer changing the billing address when a ship address is required when the bill_address on the order doesn't match that on the subscription does not update bill_address or ship_address on the order
      Failure/Error: payment_method.create_adjustment(adjustment_label, order, self, true)
      
      NoMethodError:
        undefined method `create_adjustment' for nil:NilClass
      # ./app/models/spree/payment_decorator.rb:26:in `ensure_correct_adjustment'
      # ./app/services/order_factory.rb:82:in `create_payment'
      # ./app/services/order_factory.rb:20:in `create'
      # ./app/models/proxy_order.rb:55:in `initialise_order!'
      # ./spec/services/order_syncer_spec.rb:146:in `block (3 levels) in <top (required)>'
```

## Possible Fix
The logic implemented to define the ship_method on the subs orders must be broken with spree 2.1.

luisramos0:
#### What? Why?

Closes #4820

#### Release notes

### DIFF
--- a/app/services/order_factory.rb
+++ b/app/services/order_factory.rb
@@ -35,13 +35,9 @@ class OrderFactory
   end
 
   def create_order
-    @order = Spree::Order.create!(create_attrs)
-  end
-
-  def create_attrs
     create_attrs = attrs.slice(:customer_id, :order_cycle_id, :distributor_id)
-    create_attrs[:email] = customer.email
-    create_attrs
+    @order = Spree::Order.create!(create_attrs)
+    @order.email = customer.email
   end
 
   def build_line_items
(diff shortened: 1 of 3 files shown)


app/services/order_factory.rb
```diff
@@ -56,17 +57,24 @@ def build_line_items
...
   end
 
   def set_user
     @order.update_attribute(:user_id, customer.user_id)
   end
 
   def set_addresses
+    byebug
+    # this mass assignment isnt working... we need an alternative...
     @order.update_attributes(attrs.slice(:bill_address_attributes, :ship_address_attributes))
```
Matt-Yorkley: We're assigning nested attributes for an associated object here, right? 

I think it's a bit different in Rails 4: https://stackoverflow.com/a/23495802

luisramos0: do you want to tackle this one @Matt-Yorkley? I havent really started, this is all I have so far and I still dont know how to adapt to rails 4.

luisramos0: this is #4820

luisramos0:
Replaced by #4827

PR #4823, opened 2020-02-22
Add some require statements to specs

luisramos0:
#### What? Why?

These requires are needed in the 3-0-stable build.

#### What should we test?
no tests needed apart from a green build.

### DIFF
--- a/spec/lib/open_food_network/bulk_coop_report_spec.rb
+++ b/spec/lib/open_food_network/bulk_coop_report_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'open_food_network/bulk_coop_report'
 
 include AuthenticationWorkflow
 
--- a/spec/lib/open_food_network/customers_report_spec.rb
+++ b/spec/lib/open_food_network/customers_report_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'open_food_network/customers_report'
 
 module OpenFoodNetwork
   describe CustomersReport do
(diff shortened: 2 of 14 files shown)


Matt-Yorkley:
This is totally blocking the current 3-0 build, right?

luisramos0:
yes, it breaks it badly.
to avoid these errors, you can use the 3-0-latest branch to work on the upgrade:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4825

Matt-Yorkley:
When I was looking at this before I ended up doing the exact same thing; having multiple branches and a big "all the changes together" branch, but it was a bit of a mess...

Matt-Yorkley:
Is there some underlying reason why these files are no longer being loaded as expected..?

luisramos0:
"having multiple branches and a big "all the changes together" branch, but it was a bit of a mess..."
that's how I worked most of the time in the last upgrade. I kept a little map of branches and what was already in the "latest" or not. It'ss quite important because we need code review and at the same time you want to be looking at the edge.

re this PR, I dont know why this is a problem in 3-0-stable...

sauloperez:
The way we decorate Spree classes in OFN and the fact that we brought Spree code recently, this change doesn't surprise me. I wouldn't even bother finding the root cause, tbh :man_shrugging:

luisramos0:
awesome, I'll merge master into 3-0-stable now :tada: New PR coming in a minute.

sauloperez:
weird. I couln't merge it myself even with three attempts :thinking:

luisramos0:
github is a bit unstable these days...

the merge PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/4861

PR #4825, opened 2020-02-22
WIP 3-0-stable LATEST BUILD - includes all pending PRs

luisramos0:
This includes cherry-picks from all pending PRs.


Matt-Yorkley:
We might need to look at this spree resource controller code a bit more.

We've imported this method:
```
def permitted_resource_params
params.require(object_name).permit!
end
```
And it's using the Template Method pattern, right? 

A couple of thoughts:

Firstly, this default method using `#permit!` will permit _everything_ and is supposed to be overridden by any (spree) controllers that inherit from it.

Secondly, we've added a few attributes to default spree objects (supplier_id in Product for example) so we might need to tweak some of those permitted attributes in some places.

luisramos0:
yeah, maybe we should make that method raise an exception!

luisramos0:
This temp LATEST is now the same as the last PR luisramos0:vo_fix :tada:

Issue #4644, opened 2020-01-05
[Spree 2.1] Implement Strong Parameters


Matt-Yorkley:
Rails 4 introduces Strong Paramaters, and we need to adapt the codebase to use it.

Spree's implementation is here:

https://github.com/spree/spree/blob/4cc2c2967352d3fe7652bd873e492c074ffcc3da/core/lib/spree/core/controller_helpers/strong_parameters.rb
https://github.com/spree/spree/blob/4cc2c2967352d3fe7652bd873e492c074ffcc3da/core/lib/spree/permitted_attributes.rb
https://github.com/spree/spree/blob/4cc2c2967352d3fe7652bd873e492c074ffcc3da/core/spec/lib/spree/core/controller_helpers/strong_parameters_spec.rb

luisramos0:
We need to see what we are going to do with this.
It's not mandatory to adopt strong parameters in rails 4 because there's a gem, protected_attributes, that just keeps the behaviour as in rails 3.
@Matt-Yorkley you mentioned that protected_attributes was creating other problems. I cant remember if there was any link to that.

Anyway, I am commenting here just because if we decide to implement Strong parameters we need to remember to remove the gem protected_attributes from the Gemfile in the upgrade branch :+1:

luisramos0:
I found that when spree added their solution for strong parameters a lot of the attr_accessible entries were removed from the models here:
https://github.com/openfoodfoundation/spree/commit/98bbbd7a9f6f2075878064b64039fa69596164e8#diff-d8e6742595fa0b47b62606ef77ddbe04

if we want to make the solution with protected_attributes work we need to re-add these attr_acessible entries to the model decorators on our side. I tried with one and it worked well. It should be straight forward to get these re-added.

Matt-Yorkley:
So the change here is that the whitelisting of which attributes can be updated has moved from models to controllers in Rails 4+.

If we use `strong_paramaters` we only need to whitelist the params in `our` controllers.

If we use `protected_attributes`, not only will we have to re-add all of the `attr_accessible` declarations in all the Spree models we use, but as we run bundler and update our (huge) list of gems, some of them that are designed for Rails 4 will not work, as the models they include will not have `attr_accessible` declarations, and this will cause fatal errors. So we will also need to patch and maintain `attr_accessible` declarations in other gems. As things progress and we update our gems more and more, we will have to add more patches and update existing patches where things have changed...

luisramos0:
[...]

PR #4827, opened 2020-02-22
[Spree 2.1] Implement Strong Parameters in various controllers

luisramos0:
#### What? Why?

#4828 disabled protected_attributes and started to implement strong parameters on some  controllers.
This PR continues that task #4644 

#### What should we test?
More green specs with no ForbiddenAttribute error.

### DIFF
--- a/app/controllers/admin/inventory_items_controller.rb
+++ b/app/controllers/admin/inventory_items_controller.rb
@@ -2,6 +2,8 @@ module Admin
   class InventoryItemsController < ResourceController
     respond_to :json
 
+    prepend_before_filter :permit_params
+
     respond_override update: { json: {
       success: lambda { render_as_json @inventory_item },
       failure: lambda { render json: { errors: @inventory_item.errors.full_messages }, status: :unprocessable_entity }
@@ -13,6 +15,10 @@ module Admin
     } }
 
     private
+
+    def permit_params
+      params.require(:inventory_item).permit!
+    end
 
     # Overriding Spree method to load data from params here so that
     # we can authorise #create using an object with required attributes
--- a/app/controllers/admin/inventory_items_controller.rb
+++ b/app/controllers/admin/inventory_items_controller.rb
@@ -2,8 +2,6 @@ module Admin
   class InventoryItemsController < ResourceController
     respond_to :json
 
-    prepend_before_filter :permit_params
-
     respond_override update: { json: {
       success: lambda { render_as_json @inventory_item },
       failure: lambda { render json: { errors: @inventory_item.errors.full_messages }, status: :unprocessable_entity }
@@ -16,18 +14,14 @@ module Admin
 
     private
 
-    def permit_params
-      params.require(:inventory_item).permit!
-    end
-
-    # Overriding Spree method to load data from params here so that
+    # Overriding resource_controller method to load data from params here so that
     # we can authorise #create using an object with required attributes
     def build_resource
-      if parent_data.present?
-        parent.public_send(controller_name).build
-      else
-        model_class.new(params[object_name]) # This line changed
-      end
+      model_class.new(permitted_resource_params)
+    end
+
+    def permitted_resource_params
+      params.require(:inventory_item).permit(:enterprise_id, :variant_id, :visible)
     end
   end
 end
--- a/app/controllers/spree/admin/payments_controller.rb
+++ b/app/controllers/spree/admin/payments_controller.rb
@@ -82,7 +82,7 @@ module Spree
            source_params = params.delete(:payment_source)[params[:payment][:payment_method_id]]
           params[:payment][:source_attributes] = source_params
         end
-        params[:payment]
+        params.require(:payment).permit(:amount, :payment_method_id, :source_attributes)
       end
 
       def load_data
--- a/app/controllers/spree/admin/taxons_controller.rb
+++ b/app/controllers/spree/admin/taxons_controller.rb
@@ -117,12 +117,10 @@ module Spree
       private
 
       def taxon_params
-        params.require(:taxon).permit(permitted_params)
-      end
-
-      def permitted_params
-        [:name, :parent_id, :position, :icon, :description, :permalink,
-         :taxonomy_id, :meta_description, :meta_keywords, :meta_title]
+        params.require(:taxon).permit(
+          :name, :parent_id, :position, :icon, :description, :permalink,
+          :taxonomy_id, :meta_description, :meta_keywords, :meta_title
+        )
       end
     end
   end
--- a/app/controllers/spree/admin/taxons_controller.rb
+++ b/app/controllers/spree/admin/taxons_controller.rb
@@ -89,7 +89,7 @@ module Spree
           @update_children = true
         end
 
-        if @taxon.update_attributes(params[:taxon])
+        if @taxon.update_attributes(taxon_params)
           flash[:success] = flash_message_for(@taxon, :successfully_updated)
         end
 
@@ -112,6 +112,17 @@ module Spree
         @taxon = Taxon.find(params[:id])
         @taxon.destroy
         respond_with(@taxon) { |format| format.json { render json: '' } }
+      end
+
+      private
+
+      def taxon_params
+        params.require(:taxon).permit(permitted_params)
+      end
+
+      def permitted_params
+        [:name, :parent_id, :position, :icon, :description, :permalink,
+         :taxonomy_id, :meta_description, :meta_keywords, :meta_title]
       end
     end
   end
--- a/app/controllers/spree/admin/zones_controller.rb
+++ b/app/controllers/spree/admin/zones_controller.rb
@@ -21,6 +21,12 @@ module Spree
         @states = State.order(:name)
         @zones = Zone.order(:name)
       end
+
+      def permitted_resource_params
+        params.require(:zone).permit(
+          :name, :description, :default_tax
+        )
+      end
     end
   end
 end
--- a/app/controllers/admin/bulk_line_items_controller.rb
+++ b/app/controllers/admin/bulk_line_items_controller.rb
@@ -29,7 +29,7 @@ module Admin
       # See https://github.com/rails/rails/blob/3-2-stable/activerecord/lib/active_record/locking/pessimistic.rb#L69
       # and https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE
       order.with_lock do
-        if @line_item.update_attributes(params[:line_item])
+        if @line_item.update_attributes(line_item_params)
           order.update_distribution_charge!
           render nothing: true, status: :no_content # No Content, does not trigger ng resource auto-update
         else
@@ -72,6 +72,10 @@ module Admin
 
     def order
       @line_item.order
+    end
+
+    def line_item_params
+      params.require(:line_item).permit(:price, :quantity, :final_weight_volume)
     end
   end
 end
--- a/app/controllers/admin/enterprise_relationships_controller.rb
+++ b/app/controllers/admin/enterprise_relationships_controller.rb
@@ -25,7 +25,7 @@ module Admin
     private
 
     def enterprise_relationship_params
-      params.require(:enterprise_relationship).permit(:parent_id, :child_id, :permissions_list)
+      params.require(:enterprise_relationship).permit(:parent_id, :child_id, permissions_list: [])
     end
   end
 end
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -96,6 +96,7 @@ module Admin
 
     def permitted_resource_params
       params.require(:schedule).permit(
+        :id,
         :name,
         order_cycle_ids: []
       )
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -95,7 +95,10 @@ module Admin
     end
 
     def permitted_resource_params
-      params.permit(:name, order_cycle_ids: [])
+      params.require(:schedule).permit(
+        :name,
+        order_cycle_ids: []
+      )
     end
   end
 end
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -93,5 +93,9 @@ module Admin
       syncer = OpenFoodNetwork::ProxyOrderSyncer.new(subscriptions)
       syncer.sync!
     end
+
+    def permitted_resource_params
+      params.require(:schedule).permit(:name, order_cycle_ids: [])
+    end
   end
 end
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -95,7 +95,7 @@ module Admin
     end
 
     def permitted_resource_params
-      params.require(:schedule).permit(:name, order_cycle_ids: [])
+      params.permit(:name, order_cycle_ids: [])
     end
   end
 end
--- a/app/controllers/spree/admin/adjustments_controller.rb
+++ b/app/controllers/spree/admin/adjustments_controller.rb
@@ -66,6 +66,12 @@ module Spree
       def enable_updates
         @adjustment.close
       end
+
+      def permitted_resource_params
+        params.require(:adjustment).permit(
+          :label, :amount, :included_tax
+        )
+      end
     end
   end
 end
--- a/app/controllers/spree/admin/images_controller.rb
+++ b/app/controllers/spree/admin/images_controller.rb
@@ -34,6 +34,12 @@ module Spree
       def destroy_before
         @viewable = @image.viewable
       end
+
+      def permitted_resource_params
+        params.require(:image).permit(
+          :attachment, :viewable_id, :alt
+        )
+      end
     end
   end
 end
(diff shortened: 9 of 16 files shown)


app/controllers/spree/admin/products_controller.rb
```diff
@@ -160,10 +160,16 @@ def collection_actions
...
-        collection_hash = Hash[params[:products].each_with_index.map { |p, i| [i, p] }]
+        collection_hash = Hash[products_params.each_with_index.map { |p, i| [i, p] }]
         Spree::ProductSet.new(collection_attributes: collection_hash)
       end
 
+      def products_params
+        params.require(:products).map do |product|
+          ActionController::Parameters.new(product.to_hash).permit(:id, :name)
+        end
+      end
```
Matt-Yorkley: Can't we do this?
```ruby
      def products_params
        params.require(:products).map do |product|
          product.permit(:id, :name)
        end
      end
```

luisramos0: yes we can!!! I'll switch.
Thanks!

Matt-Yorkley: Awesome, that looks a lot nicer ad is easier to read. :+1:

app/controllers/admin/order_cycles_controller.rb
```diff
@@ -235,5 +235,14 @@ def require_order_cycle_set_params
     def ams_prefix_whitelist
       [:basic, :index]
     end
+
+    def order_cycle_params
+      return params[:order_cycle] if params[:order_cycle].empty?
```
Matt-Yorkley: You've added this in a couple of places. Why it is it needed?

app/controllers/admin/inventory_items_controller.rb
```diff
@@ -14,6 +16,10 @@ class InventoryItemsController < ResourceController
 
     private
 
+    def permit_params
+      params.require(:inventory_item).permit!
```
Matt-Yorkley: I don't think we should be using `#permit!`...

app/controllers/admin/subscriptions_controller.rb
```diff
@@ -149,11 +149,25 @@ def strip_banned_attrs
...
+      return params[:subscription] if params[:subscription].empty?
+
+      params.require(:subscription).permit(
+        :shop_id, :schedule_id, :customer_id,
+        :payment_method_id, :shipping_method_id,
+        :begins_at, :ends_at,
+        :canceled_at, :paused_at,
+        :subscription_line_items_attributes => [:id, :quantity, :variant_id],
+        :bill_address_attributes => permitted_address_attributes,
+        :ship_address_attributes => permitted_address_attributes
```
Matt-Yorkley: Nice :+1: 

This is calling Spree's strong params, right?

app/controllers/admin/variant_overrides_controller.rb
```diff
@@ -92,5 +92,11 @@ def collection_errors
       full_messages.each { |fm| errors.add(:base, fm) }
       errors
     end
+
+    def variant_overrides_params
+      params.require(:variant_overrides).map do |variant_override|
+        variant_override.permit(:id, :price, :count_on_hand, :sku, :on_demand)
```
Matt-Yorkley: :+1:

app/controllers/spree/admin/payment_methods_controller.rb
```diff
@@ -40,8 +40,8 @@ def update
           @payment_method = PaymentMethod.find(params[:id])
         end
 
-        payment_method_params = params[ActiveModel::Naming.param_key(@payment_method)] || {}
-        attributes = params[:payment_method].merge(payment_method_params)
+        update_params = params[ActiveModel::Naming.param_key(@payment_method)] || {}
+        attributes = payment_method_params.merge(update_params)
```
Matt-Yorkley: It's not very clear what's happening here, or why...? Can it be extracted a bit so it's easier to read?

app/controllers/spree/admin/payment_methods_controller.rb
```diff
@@ -100,6 +100,10 @@ def show_provider_preferences
 
       private
 
+      def payment_method_params
+        params.require(:payment_method).permit!
```
Matt-Yorkley: Again, I don't think we should be using `#permit!` unless there's absolutely no other option...

app/controllers/spree/admin/payments_controller.rb
```diff
@@ -82,7 +82,7 @@ def object_params
            source_params = params.delete(:payment_source)[params[:payment][:payment_method_id]]
           params[:payment][:source_attributes] = source_params
         end
-        params[:payment]
+        params.require(:payment).permit(:amount, :payment_method_id, :source_attributes)
```
Matt-Yorkley: Does this method still return the same output as before?

app/controllers/spree/admin/resource_controller.rb
```diff
@@ -251,6 +251,10 @@ def object_url(object = nil, options = {})
         end
       end
 
+      def permitted_resource_params
+        params.require(object_name).permit!
```
Matt-Yorkley: As you mentioned before, we could have this method return an error instead of permitting everything. It's a bold move, but I think it might be the right choice...

app/controllers/spree/admin/taxons_controller.rb
```diff
@@ -113,6 +113,17 @@ def destroy
...
         respond_with(@taxon) { |format| format.json { render json: '' } }
       end
+
+      private
+
+      def taxon_params
+        params.require(:taxon).permit(permitted_params)
+      end
+
+      def permitted_params
```
Matt-Yorkley: Just for consistency... this is different to how you've done it in all the other controllers (two methods instead of one).

app/controllers/spree/admin/payments_controller.rb
Matt-Yorkley: I guess it must do.. :+1:

app/controllers/admin/order_cycles_controller.rb
luisramos0: some of the tests send empty params[:order_cycle] and the following require/permit statement fails.

app/controllers/admin/subscriptions_controller.rb
luisramos0: yes!

app/controllers/spree/admin/payment_methods_controller.rb
luisramos0: yes I can. it's spree code :-)

app/controllers/spree/admin/payment_methods_controller.rb
luisramos0: this is coming from spree. I'll improve this :+1:

app/controllers/spree/admin/payments_controller.rb
luisramos0: yes, it's equivalent.

app/controllers/spree/admin/resource_controller.rb
luisramos0: ok, I agree, I'll do that in this PR

app/controllers/spree/admin/taxons_controller.rb
luisramos0: this is spree code, I'll make it consistent.

app/controllers/admin/inventory_items_controller.rb
luisramos0: done in new commit :+1:

app/controllers/spree/admin/payment_methods_controller.rb
luisramos0: ok, done in new commit.

app/controllers/spree/admin/payment_methods_controller.rb
luisramos0: done by listing the params.
I also fixed the respective spec.

app/controllers/spree/admin/resource_controller.rb
luisramos0: done, it's now raising am error.

app/controllers/spree/admin/taxons_controller.rb
luisramos0: ok, done, merged into one method only.

luisramos0:
ready for review again @Matt-Yorkley

app/controllers/admin/enterprises_controller.rb
```diff
@@ -311,5 +311,20 @@ def location_after_save
...
+      params.require(:enterprise).permit(
+        :name, :is_primary_producer, :visible, :permalink,
+        :contact_name, :email_address, :phone, :sells, :owner_id,
+        :website, :facebook, :instagram, :linkedin, :twitter,
+        :abn, :acn, :charges_sales_tax, :display_invoice_logo,
+        :invoice_text, :description, :long_description, :promo_image,
+        :preferred_product_selection_from_inventory_only, :preferred_shopfront_message,
+        :preferred_shopfront_closed_message, :preferred_shopfront_taxon_order,
+        :preferred_shopfront_order_cycle_order, :require_login,
+        :allow_guest_orders, :allow_order_changes, :enable_subscriptions)
```
Matt-Yorkley: If we change this to include:
```ruby
...
   :allow_guest_orders, :allow_order_changes, :enable_subscriptions,
   producer_properties_attributes: [:id, :property_name, :value, :_destroy])
end
```

...we fix another 6+ specs related to saving producer properties. I can submit it separately after this is merged, if it's easier.

Also in the above example, `producer_properties_attributes` is one of those annoying arrays of unknown length where we need to permit attributes on each item in the array. With this array syntax it apparently **Just Works™**...

app/controllers/spree/admin/resource_controller.rb
```diff
@@ -251,8 +251,11 @@ def object_url(object = nil, options = {})
         end
       end
 
+      # Permit specific list of params
+      #
+      # Example: params.require(object_name).permit(:name)
       def permitted_resource_params
-        params.require(object_name).permit!
+        raise "All extending controllers need to override the method permitted_resource_params"
```
Matt-Yorkley: :muscle:

app/controllers/spree/admin/taxons_controller.rb
```diff
@@ -117,12 +117,10 @@ def destroy
       private
 
       def taxon_params
-        params.require(:taxon).permit(permitted_params)
-      end
-
-      def permitted_params
-        [:name, :parent_id, :position, :icon, :description, :permalink,
-         :taxonomy_id, :meta_description, :meta_keywords, :meta_title]
+        params.require(:taxon).permit(
```
Matt-Yorkley: :+1:

app/controllers/spree/admin/resource_controller.rb
luisramos0: I have 37 exceptions in the latest build... I'll fix them now.

app/controllers/admin/enterprises_controller.rb
luisramos0: in what spec? I only see one fixed in admin/enterprises_controller

app/controllers/admin/enterprises_controller.rb
```diff
@@ -316,15 +316,22 @@ def enterprise_params
...
         :preferred_shopfront_order_cycle_order, :require_login,
-        :allow_guest_orders, :allow_order_changes, :enable_subscriptions)
+        :allow_guest_orders, :allow_order_changes, :enable_subscriptions,
+        address_attributes: permitted_address_attributes
+      )
+    end
+
+    # Used in ResourceController#create
+    def permitted_resource_params
+      enterprise_params
```
Matt-Yorkley: Looking at this I'm wondering if it would be DRYer if we just define `permitted_resource_params` in each of our controllers, and call that in the methods where we use params, instead of defining two methods everywhere...

luisramos0: yeah, I see, I prefer to keep it like this so that resource controller stuff is not the default. I think we should try to remove resource controller all together with its "callbacks and stuff.."
I like the default as enterprise_params.

luisramos0: yeah, I see, I prefer to keep it like this so that resource controller stuff is not the default. I think we should try to remove resource controller all together with its "callbacks and stuff.."
I like the default as enterprise_params.

app/controllers/admin/enterprises_controller.rb
Matt-Yorkley: Producer properties tests in `spec/features/admin/enterprises_spec.rb`.

Matt-Yorkley: I've just noticed we also need `user_ids: []` in the list as well.

Sorry for the constant stream of annoying additions. :sweat_smile:

app/controllers/spree/admin/resource_controller.rb
luisramos0: ok, all exceptions fixed now with a few more simple commits to review :-)

app/controllers/admin/enterprises_controller.rb
luisramos0: done :+1:

Matt-Yorkley:
You left a `byebug` in the zones controller commit: https://github.com/openfoodfoundation/openfoodnetwork/blob/a9dae09fab7e16dfc6544f4636ad329be5c0baaf/app/controllers/spree/admin/zones_controller.rb

app/controllers/spree/admin/zones_controller.rb
```diff
@@ -21,6 +21,13 @@ def load_data
         @states = State.order(:name)
         @zones = Zone.order(:name)
       end
+
+      def permitted_resource_params
+        byebug
```
Matt-Yorkley: :point_up:

luisramos0: :see_no_evil:  fixed

app/services/checkout/form_data_adapter.rb
```diff
@@ -20,7 +22,20 @@ def initialize(params, order, current_user)
     end
 
     def order_params
-      @params[:order]
+      @params.require(:order).permit(
```
Matt-Yorkley: This helps fix a Paypal spec :heavy_check_mark: :heavy_check_mark: :muscle:

app/controllers/admin/inventory_items_controller.rb
```diff
@@ -14,14 +14,14 @@ class InventoryItemsController < ResourceController
 
     private
 
-    # Overriding Spree method to load data from params here so that
+    # Overriding resource_controller method to load data from params here so that
     # we can authorise #create using an object with required attributes
     def build_resource
-      if parent_data.present?
```
sauloperez: why are we removing this? are we really sure?

luisramos0: yes :+1: parent data is used when there is a parent like product on the images controller or the order in the adjustments controller...

luisramos0:
I have added a few more strong params commits to this PR.
I am now thinking we should extract these lists of attributes. Like Spree did but into a service.
I think I'll create a service app/services/permitted_attributes where I'll put the lists of permitted attributes.

Matt-Yorkley:
> I am now thinking we should extract these lists of attributes

I assumed we would do that at some point :+1::+1:

I think it would be good to merge this PR and iterate afterwards instead of adding any more to it.

luisramos0:
I think this PR is too big and not good enough.... so I'd like to close it and create separate PRs for separate controllers and include extracted versions in those PRs.

sauloperez:
> I am now thinking we should extract these lists of attributes. Like Spree did but into a service.

Do you mean all the permitted attributes of all the models? If that is the case I'm not sure if that would be a win.

luisramos0:
Something like this:
https://github.com/openfoodfoundation/spree/blob/2-1-0-stable/core/lib/spree/permitted_attributes.rb

It's not all the models, just the models that are mass_assigned from controllers, it's a shorter list, this PR basically has them all already.

I will give it a try next week and see what's the best approach.
We can have:
1 - one single class with attributes of all these models
2 - one single class with attributes of only the most complicated models like order, product and order_cycle
3 - a few classes list for example app/services/permitted_attributes/order.rb
Lets see.

sauloperez:
I choose 3 from the list above.

luisramos0:
thanks for the feedback :+1:

luisramos0:
ok, I am working on this now, this is the first PR with an extracted service, ready for review:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4876/commits/fd2cf7295e74d6229bd3d349060539067baa77f3

luisramos0:
ok, job done.
I left the most simple cases in this PR, ready for final review (no changes).

And I extracted 7 PRs out of this one, all ready for review:
- payment methods	https://github.com/openfoodfoundation/openfoodnetwork/pull/5034
- order cycle 	https://github.com/openfoodfoundation/openfoodnetwork/pull/4876
- checkout 	https://github.com/openfoodfoundation/openfoodnetwork/pull/5035
- users 	https://github.com/openfoodfoundation/openfoodnetwork/pull/5036
- subs 	https://github.com/openfoodfoundation/openfoodnetwork/pull/5037
- products et al 	https://github.com/openfoodfoundation/openfoodnetwork/pull/5038
- enterprises and customers 	https://github.com/openfoodfoundation/openfoodnetwork/pull/5039

PR #4828, opened 2020-02-22
[Spree 2.1] Strong params

Matt-Yorkley:
#### What? Why?

Starts the implementation of strong paramaters in `3-0-stable` #4644 

- Removes `protected_attributes` gem
- Removes all declarations of `attr_accessible`
- Removes use of `without_protection: true` in `#create` calls.
- Adds some temporary debugging help for strong params in the CI build
- Adds some `#permit` calls where needed in a couple of controller actions


Matt-Yorkley:
@luisramos0 

This commit makes all the strong-params-related issues **visible in the CI build**: https://github.com/openfoodfoundation/openfoodnetwork/pull/4828/commits/eb43e39722bf046c40e5755f2c120e3d9de96c6c

I think we basically need to get the three initial commits plus this debugging commit merged into the `3-0-stable` and iterate on filling out the reset of the `#permit` declarations after that...

What do you think?

app/controllers/spree/admin/base_controller.rb
```diff
@@ -12,6 +12,12 @@ class BaseController < Spree::BaseController
       before_filter :set_locale
       before_filter :warn_invalid_order_cycles, if: :html_request?
 
+      # Temporary measure to help debugging strong_parameters
+      rescue_from ActiveModel::ForbiddenAttributesError, with: :print_params
+      def print_params
+        raise ActiveModel::ForbiddenAttributesError, params.to_s
+      end
```
luisramos0: Do we really need this one? spree_admin_base inherits from spree_base (in spree core) that inherits from appcontroller above

luisramos0:
ok, let's do this!

Did you test the 2 fixes in the controllers? Are they fixing the specs?
If so, let's merge this PR and carry on this work.

Matt-Yorkley:
So now we've decoupled from spree_backend with the previous controllers... I think we might need to backport some Spree 2.1 bits to help out here. The resource_controller in particular...

luisramos0:
yeah, this bit:
https://github.com/openfoodfoundation/spree/blob/95d3ccb32f2e4016d0fc73f39446d1739da56c50/backend/app/controllers/spree/admin/resource_controller.rb#L254

Matt-Yorkley:
> this bit

Yes, and some of the related code changes in the basic `create` and `update` actions...

app/controllers/spree/admin/base_controller.rb
Matt-Yorkley: Removed.

Matt-Yorkley:
> Did you test the 2 fixes in the controllers? Are they fixing the specs?

It was tested and working previously. I can't see much in the current build until we merge the other bits though...

luisramos0:
feel free to push to my branch 3-0-latest so you can see it running with the latest stuff

luisramos0:
I just pushed this PR into 3-0-latest
Let's see how is the build with everything together:
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-4825/builds/3

luisramos0:
ah, interesting! There are only 480 broken specs now and ~150 have ForbiddenAttributesError.
I guess we have work to do now :+1:

Matt-Yorkley:
Awesome :muscle:

It should be clearer what needs doing after this

luisramos0:
I am on it now :-)

luisramos0:
yeah, it fixed 60/80 specs. Around 80/90 specs to go now with OFN controllers.

sauloperez:
Are you aware of https://github.com/rails/strong_parameters#migration-path-to-rails-4 @Matt-Yorkley ?

luisramos0:
I was not aware of that guide :-O 
We should try to use that script anyway and see if we are missing any attributes.

Matt-Yorkley:
Ah, nice. I didn't see that script. I tried a different "Rails 4 upgrade" script previously and it was _really_ unhelpful...

PR #4843, opened 2020-02-25
[Spree 2.1] Product import specs

Matt-Yorkley:
For some reason some of the superfluous attributes being assigned here have started throwing fatal errors in Rails 4 instead of being silently ignored...

Example error: 
```
ActiveRecord::AssociationTypeMismatch:
       Spree::ShippingCategory(#47179900444260) expected, got String(#47179814002740)
```

This isn't pretty, but it seems to fix all 56 of the broken product import tests (when run on top of `luisramos0/3-0-latest`).

### DIFF
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -62,7 +62,10 @@ module ProductImport
     end
 
     def mark_as_new_variant(entry, product_id)
-      new_variant = Spree::Variant.new(entry.attributes.except('id', 'product_id', 'on_hand', 'on_demand'))
+      new_variant = Spree::Variant.new(
+        entry.assignable_attributes.except('id', 'product_id', 'on_hand', 'on_demand',
+                                           'variant_unit', 'variant_unit_scale', 'primary_taxon_id')
+      )
       new_variant.save
       new_variant.on_demand = entry.attributes['on_demand'] if entry.attributes['on_demand'].present?
       new_variant.on_hand = entry.attributes['on_hand'] if entry.attributes['on_hand'].present?
@@ -295,7 +298,7 @@ module ProductImport
 
     def mark_as_new_product(entry)
       new_product = Spree::Product.new
-      new_product.assign_attributes(entry.attributes.except('id'))
+      new_product.assign_attributes(entry.assignable_attributes.except('id', 'on_hand', 'on_demand', 'display_name'))
       new_product.supplier_id = entry.producer_id
       entry.on_hand = 0 if entry.on_hand.nil?
 
@@ -307,7 +310,9 @@ module ProductImport
     end
 
     def mark_as_existing_variant(entry, existing_variant)
-      existing_variant.assign_attributes(entry.attributes.except('id', 'product_id'))
+      existing_variant.assign_attributes(
+        entry.assignable_attributes.except('id', 'product_id', 'variant_unit', 'variant_unit_scale', 'primary_taxon_id')
+      )
       check_on_hand_nil(entry, existing_variant)
 
       if existing_variant.valid?
--- a/app/models/product_import/spreadsheet_entry.rb
+++ b/app/models/product_import/spreadsheet_entry.rb
@@ -46,6 +46,10 @@ module ProductImport
       attrs.except(*non_product_attributes)
     end
 
+    def assignable_attributes
+      attributes.except(*non_assignable_attributes)
+    end
+
     def displayable_attributes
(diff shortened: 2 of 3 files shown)


app/models/product_import/entry_validator.rb
```diff
@@ -307,7 +309,9 @@ def mark_as_new_product(entry)
     end
 
     def mark_as_existing_variant(entry, existing_variant)
-      existing_variant.assign_attributes(entry.attributes.except('id', 'product_id'))
+      existing_variant.assign_attributes(
+        entry.attributes.except('id', 'product_id', 'shipping_category', 'tax_category')
+      )
```
luisramos0: I get  "unknown attribute: producer" here: https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-4825/builds/17

luisramos0: locally if I do:
```diff
diff --git a/app/models/product_import/entry_validator.rb b/app/models/product_import/entry_validator.rb
index f9b7ba8ab..d81997f39 100644
--- a/app/models/product_import/entry_validator.rb
+++ b/app/models/product_import/entry_validator.rb
@@ -62,8 +62,12 @@ module ProductImport
     end
 
     def mark_as_new_variant(entry, product_id)
+      puts entry.attributes
       new_variant = Spree::Variant.new(
-        entry.attributes.except('id', 'product_id', 'on_hand', 'on_demand', 'shipping_category', 'tax_category')
+        entry.attributes.except('id', 'product_id', 'on_hand', 'on_demand', 'shipping_category', 'tax_category',
+          'category', 'units', 'unit_type', 'variant_unit', 'variant_unit_scale', 'unscaled_units',
+          'enterprise', 'enterprise_id', 'producer', 'producer_id', 'primary_taxon_id'
+          )
       )
       new_variant.save
       new_variant.on_demand = entry.attributes['on_demand'] if entry.attributes['on_demand'].present?
@@ -310,7 +314,10 @@ module ProductImport
 
     def mark_as_existing_variant(entry, existing_variant)
       existing_variant.assign_attributes(
-        entry.attributes.except('id', 'product_id', 'shipping_category', 'tax_category')
+        entry.attributes.except('id', 'product_id',
+          'category', 'units', 'unit_type', 'variant_unit', 'variant_unit_scale', 'unscaled_units',
+          'enterprise', 'enterprise_id', 'producer', 'producer_id', 'primary_taxon_id',
+          'shipping_category', 'tax_category')
       )
       check_on_hand_nil(entry, existing_variant)

```
I get this one green: 
/dev/openfoodnetwork (3-0-latest) $ bin/rspec spec/models/product_importer_spec.rb:228

we probably need a better way of doing this... at least extract this list into a method...

Matt-Yorkley: I think it just needs `producer` and not the others.

> we probably need a better way of doing this...

Agreed.

luisramos0: I added only producer and the spec was still failing...

Matt-Yorkley: I've updated it and it's now passing. I also found some dead code for a sub-feature that's not used anymore... :fire:

luisramos0: great!
but we can define default in ProdImport, right?
It'd feel safer if that dead code was deleted and verified in master.

Matt-Yorkley: Moved to #4856

luisramos0: awesome!

app/models/product_import/spreadsheet_entry.rb
```diff
@@ -92,5 +96,10 @@ def non_product_attributes
        'product_validations', 'inventory_validations', 'validates_as',
        'save_type', 'on_hand_nil', 'has_overrides']
     end
+
+    def non_assignable_attributes
+      ['producer', 'producer_id', 'category', 'shipping_category', 'tax_category',
+       'units', 'unscaled_units', 'unit_type', 'enterprise', 'enterprise_id']
+    end
```
sauloperez: I would rather move this array to a constant (and call #freeze on it so we forbid mutation). This way Ruby evaluates at once at boot time and not in every request. The frozen string comment already avoids the string allocations on every request so there's less gain with the constant though.

Matt-Yorkley: Done.

Matt-Yorkley: Also rebased to fix the build...

Issue #4645, opened 2020-01-05
[Spree 2.1] Adapt PaypalExpress gem


Matt-Yorkley:
The paypal gem is currently disabled, and can't be installed with bundler. The `2-1-stable` branch explicitly requires `spree_core ~> 2.1.1`, but we are using `2.1.0`.

See: https://github.com/openfoodfoundation/better_spree_paypal_express/blob/76e113d61cb1a24645f9b0975a472f03cd0fc2a7/spree_paypal_express.gemspec#L26

A large number of the remaining failing tests are related to this.

luisramos0:
Some quickfixes are in the branch, they need to be reverted/fixed:
      - file spec/controllers/spree/paypal_controller_spec.rb deleted
      - file app/controllers/spree/paypal_controller_decorator.rb  deleted
      - the import spree_paypal_express is commented in app/assets/javascripts/admin/all.js
      - the gem spree_paypal_express is commented in the Gemfile

luisramos0:
spree 2.1.1 was released 2 days after spree 2.1.0 and just gemfile differences, so we may just move to 2.1.1 immediately.
https://github.com/spree/spree/compare/v2.1.0...v2.1.1

I'll give it a try.

luisramos0:
At the same time looks like paypal gem 2-0-stable and 2-1-stable are exactly the same, so first, I'll try to run everything with the existing paypal 2-0-stable gem.

luisramos0:
note for the future, this is exactly what we did:
"patched better_spree_paypal_express 2-1-stable to point to spree 2.1.0 instead of 2.1.1"

Matt-Yorkley:
In more detail:

- Copied `2-1-stable` (contains upstream changes for Spree 2.1) to new `2-1-0-stable` branch
- Cherry-picked two OFN commits from previous `2-0-stable` branch
- Changed restriction in gemspec from `spree_core ~> 2.1.1` to `spree_core ~> 2.1.0`

PR #4852, opened 2020-02-26
[Spree 2.1] Paypal gem test

Matt-Yorkley:
Related to #4645 

This uses the (new) `2-1-0-stable` branch of the paypal gem. It's the `2-1-stable` branch plus a couple of old OFN modifications (from 2-0-stable) cherry-picked on top, plus a minor modification to the gemspec to allow it to be used with spree_core `2.1.0`.

Rebased on top of `luisramos0/3-0-latest` for better spec visibility.

The new paypal gem branch is here: https://github.com/openfoodfoundation/better_spree_paypal_express/commits/2-1-0-stable


luisramos0:
we are doing the same thing :see_no_evil: 
https://github.com/openfoodfoundation/openfoodnetwork/pull/4825/commits/ef47527dfac69aad40a08e622c66d3a3e024c6fd

Matt-Yorkley:
Okay... all but one of the Paypal-related specs are now passing with the new adapted gem. :tada:

The final one is a feature spec that tests for a flash (error) message during checkout. The flash message is expected to contain: `PayPal failed`, but the error message being shown seems to be a generic one that was added in a recent commit which altered checkout failure messages: `The checkout failed. Please let us know so that we can process your order.`

I'm not sure if this is caused by the Paypal gem not working, or by the recent change in checkout error responses...

Matt-Yorkley:
Aha! It looks like there was a sneaky `ActiveModel::ForbiddenAttributesError` hiding somewhere deep in the checkout process. The unrecognised error got caught by the new rescue block in #4699, which interrupted the regular error response in the test...

I'll do some more investigating.

Matt-Yorkley:
All passing after a rebase on `luisramos/3-0-latest` and an extra commit for updating a broken spree_routes helper. I think it's ready :tada:

I'll submit a new PR when we've merged all the open ones, there's some messy conflicts at the moment.

Matt-Yorkley:
This PR was for testing only. Already merged in: #4862

PR #4856, opened 2020-02-27
Delete some dead code in Product Import (fixes 1 broken spec)

Matt-Yorkley:
Related to #4843 

#### What? Why?

Deletes some dead code for a Product Import sub-feature that was removed months ago, where default values for each column could be applied via the UI.

#### What should we test?

A basic Product Import check should be enough.

#### Release notes

Removed some dead code from Product Import

Changelog Category: Removed

Also fixes a spec in the current spree upgrade...

### DIFF
--- a/app/models/product_import/entry_processor.rb
+++ b/app/models/product_import/entry_processor.rb
@@ -122,7 +122,6 @@ module ProductImport
 
     def save_new_inventory_item(entry)
       new_item = entry.product_object
-      assign_defaults(new_item, entry)
       new_item.import_date = @import_time
 
       if new_item.valid? && new_item.save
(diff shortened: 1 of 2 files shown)


luisramos0:
The spec that is deleted in this PR is broken on the upgrade branch, so I am adding this PR to the rails upgrade epic so we know there's one spec being "fixed" here.

RachL:
Basic product import working fine. Here is my test file: https://framadrop.org/r/N_dGlB8f6a#xC+uzo/X0HZ3swoZ5n97qFiKyeUcSeVS5eOoiFBdp3U=

Moving to ready to go!

PR #4857, opened 2020-02-27
Remove dead spec in SchedulesController

luisramos0:
#### What? Why?
Removed dead spec...
This spec is broken in rails 4 and looking at the commit where it was introduced and making an overview of the schedules code, my conclusion is that we never need an HTML list of schedules, why would we test that it renders an empty collection!?

#### What should we test?
green build and test that schedules setup works: we need to test that we can create and assign schedules to order cycles in the order cycle list page as well as in the order cycles edit pages.

#### Release notes
Changelog Category: Removed
Remove unnecessary test.

### DIFF
--- a/spec/controllers/admin/schedules_controller_spec.rb
+++ b/spec/controllers/admin/schedules_controller_spec.rb
@@ -10,19 +10,6 @@ describe Admin::SchedulesController, type: :controller do
     let!(:coordinated_schedule) { create(:schedule, order_cycles: [coordinated_order_cycle] ) }
     let!(:uncoordinated_schedule) { create(:schedule, order_cycles: [other_order_cycle] ) }
 
-    context "html" do
-      context "where I manage an order cycle coordinator" do
-        before do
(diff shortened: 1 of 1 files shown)


RachL:
Assigning and creating is working fine, so I'm moving this to ready to go.

However I noticed something and I'm not sure it is the correct behavior (but it is the same in production): when removing a schedule that is only used once, the schedule disappears completely. I cannot re-add it again, I need to recreate a schedule. How is it on database side: does removing the schedule from the last OC it was attached to delete the circle @luisramos0?

sauloperez:
> does removing the schedule from the last OC it was attached to delete the circle @luisramos0?

What do you mean by circle @RachL ?

RachL:
lol @sauloperez I meant schedule :D

luisramos0:
interesting case @RachL 
the schedule is in the DB BUT it's not associated to any OC so it wont show in the list of schedules available.
This only happens if the schedule is removed from _all_ OCs of the enterprise.

The workaround is to create a new schedule with the same name. It will work fine.
We could create an issue S4 probably?

PR #4861, opened 2020-02-28
[Spree 2.1] Merge master into 3-0-stable

luisramos0:
#### What? Why?

I just had to resolve some simple conflicts in the Gemfile :muscle:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -622,8 +622,8 @@ GEM
       sprockets (>= 2.8, < 4.0)
     state_machine (1.2.0)
     stringex (1.5.1)
-    stripe (5.11.0)
-    temple (0.8.2)
+    stripe (5.15.0)
+    test-unit (3.3.5)
       power_assert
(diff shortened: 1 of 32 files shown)


Gemfile.lock
```diff
@@ -622,8 +622,8 @@ GEM
       sprockets (>= 2.8, < 4.0)
     state_machine (1.2.0)
     stringex (1.5.1)
-    stripe (5.11.0)
-    temple (0.8.2)
```
Matt-Yorkley: What's this `temple` gem? Looks like it's used by HAML. I can still see it in my version of 3-0-stable... was it removed in master?

Matt-Yorkley: Ok, it's not in `master`... :heavy_check_mark:

luisramos0: it's a dependency of haml 5 (rails 4) but of haml 4 (rails 3)

PR #4862, opened 2020-02-28
[Spree 2.1] Re-introduce Paypal gem

### LINKED ISSUE #4645: [Spree 2.1] Adapt PaypalExpress gem

Matt-Yorkley:
Closes #4645 

This uses the (newly added) `2-1-0-stable` branch of the paypal gem. It's the upstream `2-1-stable` branch plus a couple of old OFN modifications (from our version of `2-0-stable`), plus a minor modification to the gemspec to allow it to be used with spree_core `2.1.0`.

The branch is here: https://github.com/openfoodfoundation/better_spree_paypal_express/tree/2-1-0-stable

Also re-adds temporarily removed paypal bits in the codebase and fixes a small spree routes_helper issue.

### DIFF
--- a/app/controllers/spree/paypal_controller_decorator.rb
+++ b/app/controllers/spree/paypal_controller_decorator.rb
@@ -1,0 +1,45 @@
+Spree::PaypalController.class_eval do
+  before_filter :enable_embedded_shopfront
+  before_filter :destroy_orphaned_paypal_payments, only: :confirm
+  after_filter :reset_order_when_complete, only: :confirm
+
+  def cancel
+    flash[:notice] = Spree.t('flash.cancel', scope: 'paypal')
+    redirect_to main_app.checkout_path
+  end
+
+  # Clears the cached order. Required for #current_order to return a new order
+  # to serve as cart. See https://github.com/spree/spree/blob/1-3-stable/core/lib/spree/core/controller_helpers/order.rb#L14
+  # for details.
+  def expire_current_order
+    session[:order_id] = nil
+    @current_order = nil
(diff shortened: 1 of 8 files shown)


app/controllers/spree/paypal_controller_decorator.rb
```diff
@@ -0,0 +1,50 @@
...
+  # for details.
+  def expire_current_order
+    session[:order_id] = nil
+    @current_order = nil
+  end
+
+  private
+
+  def permit_parameters!
+    params.permit(:token, :payment_method_id, :PayerID)
```
luisramos0: this is weird, do you need PayerId to be capitalized?

Matt-Yorkley: https://github.com/openfoodfoundation/better_spree_paypal_express/blob/e28e4a8c5cedba504eea9cdad4be440d277d7e68/app/controllers/spree/paypal_controller.rb#L54

:man_shrugging:

luisramos0: ok

That emoji is very useful man shrugging! :man_shrugging:

Issue #4777, opened 2020-02-07
[Spree 2.1] Replace various quickfixes in the upgrade branch


luisramos0:
This issue/epic represents all the quickfixes we have currently in the upgrade branch that must be analysed, fixed, removed before the branch can go live.

TODO:
- [x] in config/routes/spree.rb "resources :credit_cards" is commented out #4862 
- [x] Simple Form is commented in the Gemfile and the file config/initializers/simple_form.rb is deleted #4865 #5010
- [x] the [override in app/models/spree/order_decorator.rb](https://github.com/openfoodfoundation/openfoodnetwork/commit/41421d6e213f0c76d2b9b5b2c169e4d794d2df77) of price_adjustments is not needed, we can simply use the new association line_item_adjustments - this was moved to its own issue outside this upgrade epic #5045
- [x] in the Enterprise model the setting acts_as_gmappable is commented #4867
- [x] in app/helpers/application_helper.rb FoundationRailsHelper::FlashHelper is commented - moved to its own issue (part of the upgrade) #5046
- [x] In the Gemfile we have the following commented gems:
    - [x] spring is commented #5044
    - [x] newrelic_rpm is commented #5044
    - [x] ofn-qz is commented #4866 
    - [x] turbo-sprockets-rails3 is commented (is not needed, see [here](https://github.com/ndbroadbent/turbo-sprockets-rails3/issues/61#issuecomment-15737594))

luisramos0:
Adding an extra point for consideration: why is gem 'i18n-js', '~> 3.5.1' downgraded to 3.2.1 in 3-0-stable. Needs to be investigated.

Matt-Yorkley:
I think some of these bits might be dead code essentially, or things that we don't need to replace.

- Is `ofn-qz` something we actually use?
- `turbo-sprockets-rails3` is not coming back, right?
- The modified version (forked by an OFN dev) of the `simple_form` gem we use was added in _2012_, and has this note: `Fix bug in simple_form preventing collection_check_boxes usage within form_for block. When merged, revert to upstream gem`.
- We don't use `newrelic` any more right?

luisramos0:
- newrelic local dashboard was added here: https://github.com/openfoodfoundation/openfoodnetwork/pull/3736

ofn-qz is used in print receipts, an admin order action. I'd assume it's a feature we need to keep!?

RachL:
[...]

PR #4865, opened 2020-02-28
[Spree 2.1] Simple Form

Matt-Yorkley:
Related to #4777

The current modified version of the `simple_form` gem is incompatible with Rails 4.

The comment in the gemfile says: 

```
# Fix bug in simple_form preventing collection_check_boxes usage within form_for block. 
# When merged, revert to upstream gem"
gem 'simple_form', github: 'RohanM/simple_form'
```

I checked, and the single commit that this old fork was designed for was actually merged in the upstream repo later the same day: **19th June 2012**.

I'm not sure how we test this exactly, I think it applies a few default CSS classes on certain elements in views..?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,9 +43,7 @@ gem 'daemons'
 gem 'delayed_job_active_record'
 gem 'delayed_job_web'
 
-# Fix bug in simple_form preventing collection_check_boxes usage within form_for block
-# When merged, revert to upstream gem
-# gem 'simple_form', github: 'RohanM/simple_form'
+gem 'simple_form'
 
 # Spree's default pagination gem (locked to the current version used by Spree)
 # We use it's methods in OFN code as well, so this is a direct dependency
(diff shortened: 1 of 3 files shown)


Matt-Yorkley:
I thought this was adding defaults to form elements in regular `form_for` blocks? Maybe we don't need it then...?

luisramos0:
@mkllnk I wonder if you have more info about this dependency. What do you think? Shall we try to remove this?

mkllnk:
This was added before I joined OFN. I don't know if it's actually used or not. I searched the history for `simple_form_for` and found this commit last mentioning it: https://github.com/openfoodfoundation/openfoodnetwork/commit/44e33b2b69e8f3a5a9b9c8ce633a1f6ed04b087f

That is an indicator that Rohan intended to use it but then never used that scaffold. I would try to remove it in master and see what happens. That's when visual testing would be ace. It also looks like we can remove that scaffold which is the only file in `lib/templates`.

Matt-Yorkley:
Great, I've made a PR to remove it in `master`: #5010. I'll close this PR.

luisramos0:
nice :muscle:

PR #4867, opened 2020-02-28
[Spree 2.1] acts_as_gmappable

Matt-Yorkley:
Part of #4777

Fixes maps in Rails 4 :tada:

In the new `2.x.x` version of the `gmaps4rails` gem (for Rails 4+), `acts_as_gmappable` doesn't exist any more, the gem has been overhauled and a lot of the other functionality has shifted to the `geocoding` gem.

It looks like we already use the `geocoding` gem correctly in `Spree::Address` decorator. It needed a slight config tweak, but now it works. The `gmaps4rails` gem seems to display the maps fine (I tested both locally using the UK maps API key).

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -17,8 +17,6 @@ class Enterprise < ActiveRecord::Base
 
   self.inheritance_column = nil
 
-  # acts_as_gmappable process_geocoding: false
-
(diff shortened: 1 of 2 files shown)


Matt-Yorkley:
Update: it looks like maps will probably just work with a couple of tweaks.

PR #4876, opened 2020-03-02
[Spree 2.1] Implement strong params in admin Order cycles controller

luisramos0:
#### What? Why?

Implements #4644 for order_cycles_controller.
This is what was on #4827 plus the last commit where I extract the new service and add a spec.

#### What should we test?
More green specs :tada:

### DIFF
--- a/app/services/permitted_attributes/order_cycle.rb
+++ b/app/services/permitted_attributes/order_cycle.rb
@@ -1,0 +1,46 @@
+module PermittedAttributes
+  class OrderCycle
+    def initialize(params)
+      @params = params
+    end
+
+    def call
+      return @params[:order_cycle] if @params[:order_cycle].empty?
+
+      @params.require(:order_cycle).permit(
+        :name, :orders_open_at, :orders_close_at, :coordinator_id,
+        :incoming_exchanges => permitted_exchange_attributes,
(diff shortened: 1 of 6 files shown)


app/services/permitted_attributes/order_cycle.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def call
+      return @params[:order_cycle] if @params[:order_cycle].empty?
+
+      @params.require(:order_cycle).permit(
+        :name, :orders_open_at, :orders_close_at, :coordinator_id,
+        :incoming_exchanges => permitted_exchange_attributes,
+        :outgoing_exchanges => permitted_exchange_attributes,
+        :schedule_ids => [], :coordinator_fee_ids => []
```
Matt-Yorkley: Any reason for using the hash-rocket style here? We could do this:
```ruby
:name, :orders_open_at, :orders_close_at, :coordinator_id,
incoming_exchanges: permitted_exchange_attributes,
outgoing_exchanges: permitted_exchange_attributes,
schedule_ids: [], coordinator_fee_ids: []
```

luisramos0: yeah, I thought this was causing problems somewhere, but apparently not :+1:

Issue #4782, opened 2020-02-10
Order Cycle Distributor Totals by Supplier results in Snail


emilyjeanrogers:
## Description

One Enterprise (Prom Coast Food Collective) on the Aus Instance has reported being unable to run the Order Cycle Distributor Totals by Supplier report. They have been trying repeatedly over 2 days and it has only successful run the report once. 

Their Order Cycle finishes tonight at 9pm (11/02/2020) and they will need to run the report between 9pm - 10pm (Aus time) 

## Expected Behaviour

Report should run successfully

## Actual Behaviour

Snail

## Steps to Reproduce

I have successfully run the report as Superadmin, and as a Manager of the specific enterprise. I have been unable to reproduce the error.

## Animated Gif/Screenshot

![Screen Shot 2020-02-11 at 10 21 05 AM](https://user-images.githubusercontent.com/47620442/74199871-8018ef80-4cb9-11ea-9198-898de6e9287d.png)

## Workaround

The only workaround is that someone at OFN will need to run the report for them at 9pm tonight, as they can't run it themselves. 

## Severity

S2 - unless the above workaround is acceptable - then S3

bug-s1: a critical feature is broken: checkout, payments, signup, login
bug-s2: a non-critical feature is broken, no workaround
bug-s3: a feature is broken but there is a workaround
bug-s4: it's annoying, but you can use it
bug-s5: we can live with it, only a few users impacted

https://github.com/openfoodfoundation/openfoodnetwork/wiki/Bug-severity

mkllnk:
I just ran the report as super-admin and it worked. But it took over 2 minutes and the process used a lot of memory, it's now at 60%. Our swap is exhausted. I restarted unicorn to free some memory but after running the report again, it used 58% of memory again. I think that means that it fails if another worker tries to run the same report because two processes can't use that much memory at the same time.

I will increase the swap size for now but I agree that this is S2 and we need to work on the performance of this report.

sauloperez:
I'm checking Wormly and I have a few doubts @mkllnk you might know the answer for

* Do we only have 2-day data retention? The RAM graph has only data until Feb 11th
* Do you know when did the user check the report?
* When did we restart the unicorn?

mkllnk:
The data reporting had crashed a long time ago and I didn't monitor that. I restarted the reporting on that day when I investigated and observed the memory consumption of the unicorn worker manually.
[...]

PR #4879, opened 2020-03-02
Inner join visible orders

sauloperez:
#### What? Why?

Improves (but doesn't close yet) #4782.

This replaces the underlying SQL to get the visible orders, which has quite far-reaching consequences as this bit of logic controls which orders a user can see in reports. Instead of using `LEFT JOIN` we use `INNER JOIN`. This means we won't load all line_items, variants and products of the DB but only the ones that match the line_items that belong to the specified orders.

This has an immediate effect on the time it takes to get the data from the DB. On staging data, goes from this [complex query plan](https://explain.dalibo.com/plan/hUj) which reports 2.46s of execution time to this other [query plan](https://explain.dalibo.com/plan/Ycj). This time the execution drops to 1.04s with a reduction (very likely but I haven't checked) on RAM usage.

#### What should we test?

We should test the distributor totals by supplier report but also at least two other reports. I'd say the ones enterprises use the most but I can only guess. My bet is:

* Packing
* Order cycle management

#### Release notes

Build reports reducing the pressure on the DB by refactoring a permissions SQL query. 
Changelog Category: Changed

### DIFF
--- a/app/controllers/spree/admin/reports_controller.rb
+++ b/app/controllers/spree/admin/reports_controller.rb
@@ -217,7 +217,11 @@ module Spree
       end
 
       def render_report(header, table, create_csv, csv_file_name)
-        send_data csv_report(header, table), filename: csv_file_name if create_csv
+        if create_csv
+          @csv_report = csv_report(header, table)
+          send_data @csv_report, filename: csv_file_name
+        end
+
         @header = header
         @table = table
         # Rendering HTML is the default.
--- a/app/services/permissions/order.rb
+++ b/app/services/permissions/order.rb
@@ -10,7 +10,7 @@ module Permissions
     # Find orders that the user can see
     def visible_orders
       Spree::Order.
-        with_line_items_variants_and_products_outer.
+        with_line_items_variants_and_products.
         where(visible_orders_where_values)
     end
 
--- a/spec/controllers/spree/admin/reports/distributor_totals_by_supplier_spec.rb
+++ b/spec/controllers/spree/admin/reports/distributor_totals_by_supplier_spec.rb
@@ -1,0 +1,249 @@
+require 'spec_helper'
+
+describe Spree::Admin::ReportsController, type: :controller do
+  let(:csv) do
+    <<-CSV.strip_heredoc
+      Hub,Producer,Product,Variant,Amount,Curr. Cost per Unit,Total Cost,Total Shipping Cost,Shipping Method
+      Mary's Online Shop,Freddy's Farm Shop,Beef - 5kg Trays,1g,1,12.0,12.0,"",Shipping Method
+      Mary's Online Shop,Freddy's Farm Shop,Fuji Apple,2g,1,4.0,4.0,"",Shipping Method
+      Mary's Online Shop,Freddy's Farm Shop,Fuji Apple,5g,5,12.0,60.0,"",Shipping Method
+      Mary's Online Shop,Freddy's Farm Shop,Fuji Apple,8g,3,15.0,45.0,"",Shipping Method
+      "",TOTAL,"","","","",121.0,2.0,""
+    CSV
+  end
+
+  before do
+    DefaultStockLocation.create!
+
+    delivery  = marys_online_shop.shipping_methods.new(
+      name: "Home delivery",
+      require_ship_address: true,
+      calculator_type: "Spree::Calculator::FlatRate",
+      distributor_ids: [marys_online_shop.id]
+    )
+    delivery.shipping_categories << DefaultShippingCategory.find_or_create
+    delivery.calculator.preferred_amount = 2
+    delivery.save!
+  end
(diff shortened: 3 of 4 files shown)


spec/controllers/spree/admin/reports/distributor_totals_by_supplier_spec.rb
```diff
@@ -0,0 +1,249 @@
+require 'spec_helper'
+
+describe Spree::Admin::ReportsController, type: :controller do
```
sauloperez: I tried to mimic the report I generated with sample data in staging. All of it without shortcuts so I managed to control all the data involved, which enabled me to understands all the bits and pieces of the report generation. I don't plan to keep after I finally close the issue. Meanwhile, it provides the lab conditions necessary to iterate quickly.

app/controllers/spree/admin/reports_controller.rb
```diff
@@ -217,7 +217,11 @@ def render_content?
       end
 
       def render_report(header, table, create_csv, csv_file_name)
-        send_data csv_report(header, table), filename: csv_file_name if create_csv
+        if create_csv
+          @csv_report = csv_report(header, table)
+          send_data @csv_report, filename: csv_file_name
+        end
```
luisramos0: from "rapper style" code to "daily language" code :heart:

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -82,8 +82,7 @@ def scale_factor(product)
     end
 
     def order_permissions
-      return @order_permissions unless @order_permissions.nil?
-      @order_permissions = ::Permissions::Order.new(@user)
+      @order_permissions ||= ::Permissions::Order.new(@user)
```
luisramos0: and from "dumb style" to "eloquent style" :heart:

app/controllers/spree/admin/reports_controller.rb
sauloperez: Note however that my whole point was to introduce an ivar (which I absolutely dislike) so I could test the csv output. I wanted to avoid any test mocks if possible.

sauloperez:
Yeah, that's what meant with the inner joins, and it is also my hope 😅

app/controllers/spree/admin/reports_controller.rb
luisramos0: ah, I see, much better anyway!

sauloperez:
This apparently harmless change

```diff
     def order_permissions
-      return @order_permissions unless @order_permissions.nil?
-      @order_permissions = ::Permissions::Order.new(@user)
+      @order_permissions ||= ::Permissions::Order.new(@user)
     end
```

broke spec/features/consumer/shopping/embedded_groups_spec.rb:14. It seems totally unrelated but... you know, embedded things...

luisramos0:
have you tried rerunning the build, embedded has a flaky spec.

sauloperez:
> have you tried rerunning the build, embedded has a flaky spec.

yes, like 3 times and it still failed. But I've seen that same spec in a build from one of your branches so it might as well be just an exceptionally flaky spec :man_shrugging: I just don't want to delay this PR any further.

sauloperez:
This would benefit from a second approval but while we wait the problem is still there. So given the reduced scope of this change, I assume the risk and move it to test ready.

RachL:
@sigmundpetersen why did you changed it back?

sigmundpetersen:
Yeah sorry, there are 2 PRs connected to the issue and I don't think the other one is Test ready. Maybe we should disconnect this one and test it seperately? And let the other remain in Code Review.

RachL:
ok, thanks :) done!

RachL:
This is looking good so far on staging.

I have checked with several cases the OC reports and the packing report.

I did the other with simple producer profile.

The results were fine.

@sauloperez I didn't test the performance part, because I guess this will be looked at in production, correct?

sauloperez:
> @sauloperez I didn't test the performance part, because I guess this will be looked at in production, correct?

Exactly. The affected SQL query is taking less than half the time in "lab" conditions so I'm hoping this is going to have a noticeable and positive impact. We'll see.

app/services/permissions/order.rb
```diff
@@ -10,7 +10,7 @@ def initialize(user)
     # Find orders that the user can see
     def visible_orders
       Spree::Order.
-        with_line_items_variants_and_products_outer.
+        with_line_items_variants_and_products.
```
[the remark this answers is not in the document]
Matt-Yorkley: Are we missing test coverage for some cases there?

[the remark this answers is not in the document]
sauloperez: > I think it was possible to see empty carts this way

LEFT JOINS would allow that but I think we somehow filter them in a later step

PR #4889, opened 2020-03-02
[Spree 2.1] Fix ProductSet spec (1 broken spec)

luisramos0:
#### What? Why?

Remove dead code that creates products in product_set.
This will fix a broken spec in the upgrade build because in rails 4 the product_id attribute causes the update to break.

This code product_set is never used to create products, only update and so this code is never used. 

#### What should we test?
Create product and bulk update products and variants.
Make sure the attributes like on_hand, display_name, variant_unit_name are saved correctly.

#### Release notes
Changelog Category: Removed
Delete some dead code that will ease the rails upgrade.

### DIFF
--- a/spec/models/spree/product_set_spec.rb
+++ b/spec/models/spree/product_set_spec.rb
@@ -12,7 +12,6 @@ describe Spree::ProductSet do
         let(:collection_hash) do
           {
             0 => {
-              product_id: 11,
               name: 'a product',
               price: 2.0,
               supplier_id: create(:enterprise).id,
@@ -25,11 +24,10 @@ describe Spree::ProductSet do
           }
         end
 
-        it 'creates it with the specified attributes' do
+        it 'does not create a new product' do
           product_set.save
 
-          expect(Spree::Product.last.attributes)
-            .to include('name' => 'a product')
+          expect(Spree::Product.last).to be nil
         end
       end
 
--- a/app/models/spree/product_set.rb
+++ b/app/models/spree/product_set.rb
@@ -1,6 +1,6 @@
 class Spree::ProductSet < ModelSet
   def initialize(attributes = {})
-    super(Spree::Product, [], attributes, proc { |attrs| attrs[:product_id].blank? })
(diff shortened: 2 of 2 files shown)


spec/models/spree/product_set_spec.rb
```diff
@@ -12,7 +12,6 @@
         let(:collection_hash) do
           {
             0 => {
-              product_id: 11,
```
Matt-Yorkley: Are we sure this won't ever be passed to `ProductSet` as an attribute?

luisramos0: no but we can test and anyway, what does it mean? it should be :id
having a separate attribute product_id for creation is just complicating and then it's  dead code :-)

luisramos0: Matt, I had a second look at both PRs. we can go with [yours](https://github.com/openfoodfoundation/openfoodnetwork/pull/4884/files) if you prefer but I think the little clean up in this one is good.

Matt-Yorkley: I also much prefer the cleanup in your PR, I just have a suspicion that the test example probably represents a real case, so the :product_id attribute in the test example might still be passed to ProductSet somewhere.

luisramos0: Adding details to what I have verified:
ProductSet is only used in spree/admin/products_controller on a collection scenario. I tested the bulk product update page and verified the product_id is not sent from there.
The bulk product update page does not create products in any circumstances.
The product creation happens in this controller but through ResourceController#create.

luisramos0:
I think I added the no-test label by mistake. There are some test steps above. Test Ready.

filipefurtad0:
Hi @luisramos0,

As admin, I created several products, and tried out bulk editing the different paramenters:
on_hand, display_name, variant_unit_name.
These changes are saved as expected, and appear on the shop front of a (non-admin) user.
All works fine, ready-to-go :+1:

PR #4916, opened 2020-03-04
[Spree 2.1] Make Spree::OrderUpdater update all adjustments, not just promotions and shipment adjustments

### LINKED ISSUE #4915: [Spree 2.1] Fix spree/admin/orders_controller_spec and spree/orders_controller_spec and admin/orders_helper_spec (3 broken specs)
3 specs broken:
spec/controllers/spree/orders_controller_spec.rb:248
spec/controllers/spree/admin/orders_controller_spec.rb:21
spec/helpers/admin/orders_helper_spec.rb:19

######

Invalid adjustment is being rendered because it is still seen as eligible.
The invalid adjustment (zero amount) never receives a call to update! and so eligibility is not correctly set. In master the adjustment receives a update! call and the eligibility is set correctly before rendering.

```
  2) Spree::Admin::OrdersController#edit view shows only eligible adjustments
     Failure/Error: expect(response.body).to_not match adjustment.label
     
       expected "<!DOCTYPE html>\n<html lang='en'>\n<head data-hook='admin_inside_head'>\n<meta content='text/html; c...ee.api_key = \"4a1d50e99b0858a36bd25281e3e8339417295fa63d8eb013\";\n</script>\n\n</body>\n</html>\n" not to match "invalid adjustment"
       Diff:
       @@ -1,2 +1,469 @@
       -invalid adjustment
       +<!DOCTYPE html>
       +<html lang='en'>
```

luisramos0:
#### What? Why?

This fixes #4915 
This spree commit is the obvious cause of the problem:
https://github.com/openfoodfoundation/spree/commit/39f9b3e9fed24f3ce8cacf3ec38b808d143550dd
it is changing from updating all adjustments to updating only shipment and promotions adjustments.
Payment adjustments are left behind and that's exactly what's failing in #4915

In terms of the solution: I dont understand why adding the method to OFN's OrderUpdater doesnt work, only adding it to the decorator works...

This updater code is a mess! Spree::OrderUpdater itself is a disaster. We have a delegator on top of it to do our own stuff. But the worse thing is that in Spree 2.2 there are massive changes to this class.... I wonder what can we do to improve this mess now or later...

#### What should we test?
Green spec.

### DIFF
--- a/app/models/spree/order_updater_decorator.rb
+++ b/app/models/spree/order_updater_decorator.rb
@@ -1,0 +1,6 @@
+Spree::OrderUpdater.class_eval do
+  # Override spree method to make it update all adjustments as in Spree v2.0.4
+  def update_shipping_adjustments
+    order.adjustments.reload.each { |adjustment| adjustment.update! }
+  end
+end


Matt-Yorkley:
So we just need to find where `adjustment.update!` is being called from in `master`..?

luisramos0:
looking in master could help. I dont really understand how could update! be called in master...

Matt-Yorkley:
> I dont really understand how could update! be called in master...

Like this: https://github.com/openfoodfoundation/openfoodnetwork/issues/4915#issuecomment-599032263

luisramos0:
I created the issue https://github.com/openfoodfoundation/openfoodnetwork/issues/5155
To be discussed when we plan the 2.2.0 upgrade or skip it!

PR #4920, opened 2020-03-04
Bring Spree::Variant#active from Spree so that we can improve it

luisramos0:
#### What? Why?

Fixes spec/controllers/api/variants_controller_spec.rb:94, the variants returned by #scope are not readonly anymore in the rails 4 branch. 
Original code in spree: https://github.com/openfoodfoundation/spree/blob/8a8585a43cd04d1a50dc65227f337a91b18d66d5/core/app/models/spree/variant.rb#L49
See commit message.

### What should we test?
Edit and delete variants in both the bulk_product_update page and the variants list and edit pages.

#### Release notes
Changelog Category: Changed
Adapted variants code to work in rails 4.

### DIFF
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -88,11 +88,12 @@ Spree::Variant.class_eval do
   def self.active(currency = nil)
     # "where(id:" is necessary so that the returned relation has no includes
     # The relation without includes will not be readonly and allow updates on it
-    where(id: joins(:prices).
-                where(deleted_at: nil).
-                where('spree_prices.currency' => currency || Spree::Config[:currency]).
-                where('spree_prices.amount IS NOT NULL').
-                select("spree_variants.id"))
+    where("spree_variants.id in (?)", joins(:prices).
+                                        where(deleted_at: nil).
+                                        where('spree_prices.currency' =>
+                                          currency || Spree::Config[:currency]).
+                                        where('spree_prices.amount IS NOT NULL').
+                                        select("spree_variants.id"))
   end
 
   # We override in_stock? to avoid depending on the non-overridable method Spree::Stock::Quantifier.can_supply?
--- a/app/models/spree/variant_decorator.rb
+++ b/app/models/spree/variant_decorator.rb
@@ -85,6 +85,10 @@ Spree::Variant.class_eval do
     ]
   end
 
+  def self.active(currency = nil)
+    where(id: joins(:prices).where(deleted_at: nil).where('spree_prices.currency' => currency || Spree::Config[:currency]).where('spree_prices.amount IS NOT NULL').select("spree_variants.id"))
+  end
+
   # We override in_stock? to avoid depending on the non-overridable method Spree::Stock::Quantifier.can_supply?
   #   VariantStock implements can_supply? itself which depends on overridable methods
   def in_stock?(quantity = 1)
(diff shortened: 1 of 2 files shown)


app/models/spree/variant_decorator.rb
```diff
@@ -85,6 +85,10 @@ def self.indexed
     ]
   end
 
+  def self.active(currency = nil)
+    where(id: joins(:prices).where(deleted_at: nil).where('spree_prices.currency' => currency || Spree::Config[:currency]).where('spree_prices.amount IS NOT NULL').select("spree_variants.id"))
```
Matt-Yorkley: Can this be tidied up a bit? :pray:

luisramos0: yes :+1:

Matt-Yorkley: That's 100 times better, thanks :heart:

app/models/spree/variant_decorator.rb
```diff
@@ -85,6 +85,16 @@ def self.indexed
     ]
   end
 
+  def self.active(currency = nil)
+    # "where(id:" is necessary so that the returned relation has no includes
+    # The relation without includes will not be readonly and allow updates on it
+    where(id: joins(:prices).
```
mkllnk: I just remembered another workaround for Rails 3:

```suggestion
    where("spree_variants.id in (?)", joins(:prices).
```

luisramos0: nice!!!!
I confirmed this works:
```ruby
puts Spree::Variant.where("spree_variants.id in (?)", [1]).where("spree_variants.id in (?)": [1,2,3]).to_sql
SELECT "spree_variants".* FROM "spree_variants"  WHERE "spree_variants"."id in (?)" IN (1, 2, 3) AND ("spree_variants".deleted_at IS NULL) AND (spree_variants.id in (1))
 => nil 

```

app/models/spree/variant_decorator.rb
```diff
@@ -86,7 +86,13 @@ def self.indexed
   end
 
   def self.active(currency = nil)
-    where(id: joins(:prices).where(deleted_at: nil).where('spree_prices.currency' => currency || Spree::Config[:currency]).where('spree_prices.amount IS NOT NULL').select("spree_variants.id"))
+    # "where(id:" is necessary so that the returned relation has no includes
```
sauloperez: could you elaborate a bit more? which includes and what does the `id IN ()` do? if we could get rid of it it could be a perf. gain. which is pretty needed. I don't mind investigating this further in another issue though.

luisramos0: the includes are prices, the IN clause makes the method return variants only without prices which makes the relation NOT be readonly.
I dont think there's a significant performance impact.

mkllnk:
@luisramos0 The Semaphore fail looks weird. Any idea?

luisramos0:
rebased to fix build problems.

luisramos0:
@mkllnk this branch must have been created when master was broken with broken engine specs.
I just rebased, all good now :ok_hand:

filipefurtad0:
Hi @luisramos0,

Edited and delete variants in:
- bulk edit page admin/products
- and in each individual product/variant pages.

Verified that these changes take effect not only in these pages but also in:
- Inventory (admin)
- frontshop/cart (user)
- reports, after checking-out some of these recently changed variants. Also, it was verified that the on variant changes have a retroactive effect those appearing in previously created reports. This is the expected behaviour.

All good here - moving to ready to go.

PR #4925, opened 2020-03-05
Adapt remaining AR finders to rails 4

luisramos0:
#### What? Why?

Final migration of finders to rails 4 style.

#### What should we test?
We should verify order customer details page is working correctly: it displays the details introduced on checkout and it allows manager to change some customer details.
We should verify for example product variants and images can be added, edited and deleted as usual.  

#### Release notes
Changelog Category: Changed
Migrated some finders to rails 4 style, this will be necessary for rails 4.1 upgrade.

### DIFF
--- a/app/controllers/spree/admin/orders/customer_details_controller.rb
+++ b/app/controllers/spree/admin/orders/customer_details_controller.rb
@@ -51,7 +51,7 @@ module Spree
         end
 
         def load_order
-          @order = Order.find_by_number!(params[:order_id], include: :adjustments)
+          @order = Order.find_by!({ number: params[:order_id] }, include: :adjustments)
         end
 
         def check_authorization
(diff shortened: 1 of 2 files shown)


filipefurtad0:
Hey @luisramos0,

Thanks for the testing notes Luis.

Checked that the order customer details page is working as before:
- the details introduced on checkout are displayed correctly
- the manager can change some customer details
- after these changes, these appear correctly on the invoices and orders (on the customers account)

Verified as well that:

- product variants can be added and edited as usual. In what deleting a variant concerns, I found an issue not introduced by this PR - but which relates to bug #5947.

Newly created items can always be deleted. But it seems to be a pattern, that limited stock products/variants cannot be deleted after orders have been placed. Orders placed on new items with unlimited stock, can be deleted. This worked, before and after this PR, so we are all good here.

- product images
can be added, edited and deleted as usual - all good here

- variant images
can be added but not deleted nor edited - a known bug #5731, which I now updated as it seems editing is not possible as well

This is ready to go - thank you!

luisramos0:
Nice, thanks Filipe!

PR #4930, opened 2020-03-06
Merge master into 3-0-stable

luisramos0:
Bring latest master commits to the upgrade branch.

Conflicts:
CONFLICT (content): Merge conflict in db/schema.rb
CONFLICT (content): Merge conflict in app/models/spree/gateway/stripe_connect.rb
CONFLICT (content): Merge conflict in app/models/schedule.rb
CONFLICT (content): Merge conflict in Gemfile.lock
CONFLICT (content): Merge conflict in spec/models/product_importer_spec.rb
CONFLICT (content): Merge conflict in app/controllers/enterprises_controller.rb


app/models/schedule.rb
```diff
@@ -1,10 +1,8 @@
 class Schedule < ActiveRecord::Base
-  has_and_belongs_to_many :order_cycles, join_table: 'order_cycle_schedules'
   has_paper_trail meta: { custom_data: :order_cycle_ids }
 
-  has_many :coordinators, uniq: true, through: :order_cycles
-
-  attr_accessible :name, :order_cycle_ids
+  has_and_belongs_to_many :order_cycles, join_table: 'order_cycle_schedules'
+  has_many :coordinators, -> { uniq }, through: :order_cycles
```
sauloperez: what is that `uniq` here? I don't see any var defined with that name, at least in its surroundings :thinking: is it maybe a scope?

Gemfile
```diff
@@ -154,12 +154,12 @@ end
 group :development do
   gem 'byebug', '~> 9.0.0' # 9.1 requires ruby 2.2
   gem 'debugger-linecache'
-  gem "newrelic_rpm", "~> 3.0"
+  #gem "newrelic_rpm", "~> 3.0"
```
sauloperez: tbh, I would just remove it. I never really used it because I get what I need with mini-profiler.

Gemfile
```diff
@@ -4,13 +4,12 @@ git_source(:github) { |repo_name| "https://github.com/#{repo_name}.git" }
 
 gem 'i18n', '~> 0.6.11'
 gem 'i18n-js', '~> 3.6.0'
-gem 'rails', '~> 3.2.22'
-gem 'rails-i18n', '~> 3.0.0'
+gem 'rails', '~> 4.0.0'
```
sauloperez: Is there a specific reason why we don't point to the latest patch? I found out the latest from Rails 4 series is [4.0.11.1](https://weblog.rubyonrails.org/2014/11/19/Rails-4-0-11-1-and-4-1-7-1-have-been-released/). There are security fixes we should benefit from.

Gemfile
```diff
@@ -21,22 +20,22 @@ gem 'pg', '~> 0.21.0'
...
 
 # We need at least this version to have Digicert's root certificate
 # which is needed for Pin Payments (and possibly others).
-gem 'activemerchant', '~> 1.78'
+gem 'activemerchant', '~> 1.78.0'
 
-gem 'devise', '~> 2.2.5'
-gem 'devise-encryptable', '0.2.0'
+gem 'devise', '~> 3.0.1'
+gem 'devise-encryptable'
```
sauloperez: isn't there a dependency on the actual devise version? we protect ourselves from backward-incompatible changes with the Gemfile.lock but might be good to point that out in the Gemfile too.

Gemfile
```diff
@@ -60,20 +59,21 @@ gem 'db2fog'
...
-gem 'sass-rails', '~> 3.2.3'
+gem 'sass'
+gem 'sass-rails'
 gem 'truncate_html'
 gem 'unicorn'
 
 # AMS is pinned to 0.8.4 because 0.9.x is a complete re-write, as is 0.10.x
 # Once Rails is updated to 5.x we should bump directly to 0.10.x
 gem "active_model_serializers", "0.8.4"
+gem 'activerecord-session_store'
```
sauloperez: can we use all the versions?

Gemfile
```diff
@@ -60,20 +59,21 @@ gem 'db2fog'
...
 gem 'acts-as-taggable-on', '~> 3.4'
 gem 'angularjs-file-upload-rails', '~> 2.4.1'
 gem 'blockenspiel'
 gem 'custom_error_message', github: 'jeremydurham/custom-err-msg'
 gem 'dalli'
-gem 'deface', '1.0.2'
+gem 'deface'
```
sauloperez: I guess we can trust all versions since it's maintained by the Spree team.

Gemfile
```diff
@@ -129,9 +129,9 @@ group :test, :development do
   gem 'atomic'
   gem 'awesome_print'
   gem 'capybara', '>= 2.18.0' # 3.0 requires nokogiri 1.8
-  gem 'database_cleaner', '0.7.1', require: false
-  gem "factory_bot_rails", require: false
-  gem 'fuubar', '~> 2.5.0'
+  gem 'database_cleaner', require: false
+  gem "factory_bot_rails", '4.8.2', require: false
```
sauloperez: Why this particular version? https://rubygems.org/gems/factory_bot_rails/versions/4.10.0 seems to be the latest version supporting Rails 4. The next one requires `railties >= 4.2.0`.

luisramos0:
I thought your review was really weird, I thought why were you reviewing things already in 3-0-stable branch?
then I realized the target branch for this PR was wrong. it was master, it should have been 3-0-stable. I changed it now...
Sorry about this. My mistake.
Anyway, your comments are valid comments for the 3-0-stable branch!

sauloperez:
ouch. Yep, in any case these are things worth checking before merging to `master`.

sauloperez:
wrong button :trollface:

luisramos0:
yes, before merging to master, but not necessarily before merging master into 3-0-stable. can you approve the PR?

app/models/schedule.rb
luisramos0: must be this https://apidock.com/rails/ActiveRecord/QueryMethods/uniq

luisramos0:
I created this issue to represent the points you raised here Pau:
https://github.com/openfoodfoundation/openfoodnetwork/issues/5011

sauloperez:
Thank you!

Gemfile
luisramos0: I ended up re-adding it. it's harmless.

Gemfile
luisramos0: I think latest is https://rubygems.org/gems/rails/versions/4.0.13
I'll give it a try now

Gemfile
luisramos0: now I see we are on 0.2.0 in master and on 0.1.2 in this branch...

Gemfile
luisramos0: yes, we are on the latest. we need to test embedded shopfronts to see that sessions are working correctly.

Gemfile
luisramos0: we can remove deface.

Gemfile
luisramos0: the build is green, will upgrade :+1:

Gemfile
luisramos0: I will switch to 020 which is the latest and keeep Gemfile as is because 0.2.0 is the latest.

Gemfile
luisramos0: 4.10.0 works fine :+1:

PR #4932, opened 2020-03-06
Adapt line items scope to include deleted variants and deleted products - rails 4 fix

luisramos0:
#### What? Why?

Fixes spec/controllers/spree/admin/reports_controller_spec.rb:209 in the upgrade branch.

I can see this was introduced in spree 2.1 https://github.com/openfoodfoundation/spree/commit/7c89051efbed7066a830ab193408c1ae99559c51
I am not sure it's related but deleted products are now not seen in this spec.

Looking at the joins(:product) statement I am not sure how this is working in rails 3 as deleted products should not be in the default scope of line_items...

Anyway, in this PR I delete the dead scope LineItem#supplied_by and I change the scope LineItem#supplied_by_any (only used in repports) to include deleted variants and products. This is done by joining the tables with SQL... is there a better way to do this? Is this good enough?

#### What should we test?
We should test order fulfillment reports with orders that include variants of deleted products and make sure those still appear in the report.

#### Release notes
Changelog Category: Changed
Adapted reports code to rails 4 and spree 2.1 code.


app/models/spree/line_item_decorator.rb
```diff
@@ -50,13 +50,12 @@
...
-      where('spree_products.supplier_id = ?', enterprise)
-  }
   scope :supplied_by_any, lambda { |enterprises|
-    joins(:product).
-      where('spree_products.supplier_id IN (?)', enterprises)
+    joins("LEFT OUTER JOIN spree_variants
+              ON spree_line_items.variant_id = spree_variants.id
+           LEFT OUTER JOIN spree_products
+              ON spree_variants.product_id = spree_products.id").
+      where("spree_products.supplier_id IN (?)", enterprises)
```
luisramos0: This is the same as the previous version BUT now includes deleted variants and products.
The objective here is simply to do something like:
    joins(:unscoped_product)
I wonder if there is a better way to do this without the left outer join SQL? Is this good enough?

mkllnk: It's not the same. `joins(:product)` creates inner joins, not outer joins. Another approach would be:

```rb
product_ids = Spree::Product.where(supplier_id: enterprises).select(:id)
variant_ids = Spree::Variant.where(product_id: product_ids).select(:id)
where("spree_line_items.variant_id IN (?)", variant_ids)
```

Inner joins are probably easier and shorter. Apparently this got fixed for Rails 4.2:
https://github.com/rubysherpas/paranoia/issues/330#issuecomment-279979421

luisramos0: -- It's not the same. joins(:product) creates inner joins, not outer joins. 
I missed that :see_no_evil: :face_with_head_bandage: 
it's not a big problem because all line items must have variants and all variants must have products. but it's a mistake :+1:

I tested and fixed your suggestion, we need to use unscoped to remove the default scopes:
```ruby
product_ids = Spree::Product.unscoped.where(supplier_id: enterprises).select(:id)
variant_ids = Spree::Variant.unscoped.where(product_id: product_ids).select(:id)
where("spree_line_items.variant_id IN (?)", variant_ids)
```
I accepted your suggestion :+1: 
Thanks Maikel!!

luisramos0: the SQL in rails 3 (2 queries..):
  Spree::Variant Load (0.6ms)  SELECT id FROM "spree_variants" WHERE "spree_variants"."product_id" IN (SELECT id FROM "spree_products" WHERE "spree_products"."supplier_id" IN (4))
  Spree::LineItem Load (0.8ms)  SELECT "spree_line_items".* FROM "spree_line_items" WHERE (spree_line_items.variant_id IN (10,9,12,11,8,7))
 
perfect in rails 4:
Spree::LineItem Load (2.6ms)  SELECT "spree_line_items".* FROM "spree_line_items" WHERE (spree_line_items.variant_id IN (SELECT id FROM "spree_variants" WHERE "spree_variants"."product_id" IN (SELECT id FROM "spree_products" WHERE "spree_products"."supplier_id" IN (4))))

mkllnk: Ah, unscoped, of course! That looks good to me but the SQL is worrying me. Worst case, logged in as super-admin, you have permission to see all enterprises which will load all variant ids in the database and copy them into a query, right? If that's right, it could reach some limits and blow up.

I'm not sure, but I would probably go back to your previous solution with proper inner joins.

luisramos0: yeah, the SQL problem is only in rails 3 right?
I am tempted to switch this PR to the rails 4 branch, does that makes sense to you?

mkllnk: Yes, I'm happy either way.

luisramos0:
ok, switched to the rails 4 branch, ready for review!

Issue #3608, opened 2019-03-15
Clean up useless data from live DBs to improve db backups speed and potentially app performance


luisramos0:
### What we should change and why (this is tech debt)
We will soon have tasks to sanitize and truncate data from production  #3579 This is useful to fetch data from live environments into testing environments.

Now we should create a new task for a different purpose: to reduce the live DB size by deleting data that is not needed.
Maybe it can be called ofn:data:clean_up

There are at least 3 good candidates to be deleted or truncated (keep only latest entries):
- spree_state_changes
- spree_log_entries
- sessions

This requires further investigation about what exactly are these tables and if they can really be deleted.

**Deleting all the data in these 3 tables in the AUS live database reduces the ZIP database size from 485MB to 67MB!**

### Context
this was found as part of Spree 2 data migrations.
This issue can include #2247 but it's not specific to sessions.

### Impact and timeline
This will make backups run faster and take a lot less space.
And it can improve app performance: #3099

sauloperez:
I think this requires more than a rake task and fits more in ofn-install than here. In the past I investigated https://www.percona.com/doc/percona-toolkit/LATEST/pt-archiver.html for instance but never go to use it in production.

luisramos0:
-  more than a rake task
there are many things you can do in this space, the scope here would be to simply delete unnecessary entries. 

- fits more in ofn-install than here
could be, I am not sure, I could easily see a job running that runs the tasks regularly (for example, delete sessions older than 1 month), that would be better in the app code, right?

- pt-archiver.html
the scope of this issue is not about archiving data, it's really just the simple case of deleting data that is not necessary for the operation

mkllnk:
To break it down by table:

Aus production

- full backup: 480 MB
- without sessions: 107 MB
- without log entries: 106 MB
- without state changes: 70 MB

So I think sessions is a big one. Then let's look at state changes. Log entries is not really worth it but we can delete it while we are at it.

None of this really affects the performance of the app directly, I think. It's only the backup generation (which can affect the performance of the app).

sauloperez:
Given these numbers it would be more about disk-space savings, right? No idea if that is currently a problem. Worth checking to be sure.

mkllnk:
For me it's more about processing times. Creating or downloading backups in half the time is great.

sauloperez:
Current numbers in FR production

[...]

PR #4935, opened 2020-03-06
Allow data archiving using the :truncate_data task

### LINKED ISSUE #4782: Order Cycle Distributor Totals by Supplier results in Snail

sauloperez:
#### What? Why?

Closes #4782 (I hope)

This refactors the `:truncate_data` rake task to allow for the removal of old data from an OFN instance as a mean for data archiving. Older data can still be recovered from an older backup. We just need to know when the data archiving was performed and go find the backup we made right
before.

It extracts the `TruncateData` service to make it possible to test without the annoyance of rake task testing. It's not a straightforward thing as you can see in https://thoughtbot.com/blog/test-rake-tasks-like-a-boss. It's a headache we can avoid if we conform to the testing pyramid.

This also gives us room to iterate on our data archiving strategy as we move along.

All credit goes to @lin-d-hop for insisting on the idea of data archiving.

##### Heads-up
:warning: to truncate data in staging as we used to you'll need to run 

```
$ bundle exec rake "ofn:data:truncate[3]"
```
That's because we switched the default 3 months to 24.

#### What should we test?

Test that loading production data in staging and then truncating it keeps working as expected. There should be no data integrity issues.

#### Release notes

Refactored truncate data rake task and default to 24 months instead of 3.
Changelog Category: Changed

### DIFF
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -28,8 +28,7 @@ namespace :ofn do
     task :truncate, [:months_to_keep] => :environment do |_task, args|
       warn_with_confirmation
 
-      months_to_keep = args.months_to_keep.to_i
-      TruncateData.new(months_to_keep).call
+      TruncateData.new(args.months_to_keep).call
     end
 
     def warn_with_confirmation
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -9,10 +9,11 @@ require 'tasks/data/truncate_data'
 namespace :ofn do
   namespace :data do
     desc 'Truncate data'
-    task truncate: :environment do
+    task :truncate, [:months_to_keep] => :environment do |_task, args|
       guard_and_warn
 
-      TruncateData.new.call
+      months_to_keep = args.months_to_keep.to_i
+      TruncateData.new(months_to_keep).call
     end
 
     def guard_and_warn
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -40,7 +40,7 @@ namespace :ofn do
       Are you sure you want to proceed? (y/N)
       MSG
 
-      exit unless HighLine.new.agree(message) { |q| q.default = "N" }
+      exit unless HighLine.new.agree(message) { |question| question.default = "N" }
     end
   end
 end
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -6,6 +6,22 @@ require 'tasks/data/truncate_data'
 # This task can be used to significantly reduce the size of a database
 #   This is used for example when loading live data into a staging server
 #   This way the staging server is not overloaded with too much data
+#
+# This is also aimed at implementing data archiving. We assume data older than
+# 2 years can be safely removed and restored from a backup. This gives room for
+# hubs to do their tax declaration.
+#
+# It's a must to perform a backup right before executing this. Then, to allow
+# for a later data recovery we need to keep track of the exact moment this rake
+# task was executed.
+#
+# Execute this in production only when the instance users are sleeping to avoid any trouble.
+#
+# Example:
+#
+# $ bundle exec rake "ofn:data:truncate[24]"
+#
+# This will remove data older than 2 years (24 months).
 namespace :ofn do
   namespace :data do
     desc 'Truncate data'
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -10,20 +10,18 @@ namespace :ofn do
   namespace :data do
     desc 'Truncate data'
     task :truncate, [:months_to_keep] => :environment do |_task, args|
-      guard_and_warn
+      warn_with_confirmation
 
       months_to_keep = args.months_to_keep.to_i
       TruncateData.new(months_to_keep).call
     end
 
-    def guard_and_warn
-      if Rails.env.production?
-        Rails.logger.info("This task cannot be executed in production")
-        exit
-      end
-
-      message = "\n <%= color('This will permanently change DB contents', :yellow) %>,
-                are you sure you want to proceed? (y/N)"
+    def warn_with_confirmation
+      message = <<-MSG.strip_heredoc
+      \n
+      <%= color('This will permanently change DB contents. Please, make a backup first.', :yellow) %>
+      Are you sure you want to proceed? (y/N)
+      MSG
       exit unless HighLine.new.agree(message) { |q| q.default = "n" }
     end
   end
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -34,7 +34,7 @@ namespace :ofn do
     def warn_with_confirmation
       message = <<-MSG.strip_heredoc
       \n
-      <% highlighted_message = "This will permanently change DB contents. Please, make a backup first." %>
+      <% highlighted_message = "This will permanently change DB contents. This is not meant to be run in production as it needs more thorough testing." %>
       <%= color(highlighted_message, :blink, :on_red) %>
       Are you sure you want to proceed? (y/N)
       MSG
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -35,10 +35,12 @@ namespace :ofn do
     def warn_with_confirmation
       message = <<-MSG.strip_heredoc
       \n
-      <%= color('This will permanently change DB contents. Please, make a backup first.', :yellow) %>
+      <% highlighted_message = "This will permanently change DB contents. Please, make a backup first." %>
+      <%= color(highlighted_message, :blink, :on_red) %>
       Are you sure you want to proceed? (y/N)
       MSG
-      exit unless HighLine.new.agree(message) { |q| q.default = "n" }
+
+      exit unless HighLine.new.agree(message) { |q| q.default = "N" }
     end
   end
 end
--- a/lib/tasks/data/truncate_data.rake
+++ b/lib/tasks/data/truncate_data.rake
@@ -1,3 +1,8 @@
+# frozen_string_literal: true
+
+require 'highline'
+require 'tasks/data/truncate_data'
+
 # This task can be used to significantly reduce the size of a database
 #   This is used for example when loading live data into a staging server
 #   This way the staging server is not overloaded with too much data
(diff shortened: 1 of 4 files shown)


luisramos0:
nice, the first commit is doing too many things. Would you be able to hack the commit history and start with a simple commit that just moves the code from the task to the service? Otherwise it is very painful to see what has actually changed.

sauloperez:
I'll do my best to split it into two commits but if take a closer look at the deletion code I don't change a single line. I just work on the `#date` method to allow specifying a date from the command line.

sauloperez:
Done. I hope this is now easier to follow @luisramos0

lib/tasks/data/truncate_data.rake
```diff
@@ -1,81 +1,28 @@
...
-      sql_delete_from "
-        exchange_fees where exchange_id
-        in (select id from exchanges #{where_oc_id_in_ocs_to_delete})"
-      sql_delete_from "exchanges #{where_oc_id_in_ocs_to_delete}"
+    def warn_with_confirmation
+      message = <<-MSG.strip_heredoc
+      \n
+      <%= color('This will permanently change DB contents. Please, make a backup first.', :yellow) %>
+      Are you sure you want to proceed? (y/N)
+      MSG
```
sauloperez: This now looks like

Screenshot from 2020-03-06 17-56-57

lib/tasks/data/truncate_data.rake
```diff
@@ -9,10 +9,11 @@ require 'tasks/data/truncate_data'
 namespace :ofn do
   namespace :data do
     desc 'Truncate data'
-    task truncate: :environment do
+    task :truncate, [:months_to_keep] => :environment do |_task, args|
       guard_and_warn
 
-      TruncateData.new.call
+      months_to_keep = args.months_to_keep.to_i
+      TruncateData.new(months_to_keep: months_to_keep).call
```
luisramos0: can be just       TruncateData.new(months_to_keep).call

lib/tasks/data/truncate_data.rake
luisramos0: Is there a color like :blinking_and_large_red instead of :yellow :-)

lib/tasks/data/truncate_data.rake
```diff
@@ -1,81 +1,44 @@
...
+#
+# It's a must to perform a backup right before executing this. Then, to allow
+# for a later data recovery we need to keep track of the exact moment this rake
+# task was executed.
+#
+# Execute this in production only when the instance users are sleeping to avoid any trouble.
+#
+# Example:
+#
+# $ bundle exec rake "ofn:data:truncate[12]"
```
sauloperez: I'm thinking it's safer to make the 2-years window the default instead of the current 3 months. Thoughts?

lib/tasks/data/truncate_data.rb
```diff
@@ -0,0 +1,95 @@
...
+    sql_delete_from "spree_line_items #{where_order_id_in_orders_to_delete}"
+    sql_delete_from "spree_payments #{where_order_id_in_orders_to_delete}"
+    sql_delete_from "spree_shipments #{where_order_id_in_orders_to_delete}"
+  end
+
+  def remove_transient_data
+    Spree::ReturnAuthorization.delete_all
+    Spree::StateChange.delete_all
+    Spree::LogEntry.delete_all
+    sql_delete_from "sessions"
```
luisramos0: I am not sure we can just wipe this out for production.

Spree::ReturnAuthorization
this is not something we can do it prod, although the last time I checked with had under 10 records across all DBs.
There must be a reason for this to be here... I wonder if this is necessary for the staging process, otherwise we could remove this one from the task.

The othres, sessions, StateChange and LogEntry we should probably add a date filter to the delete statement.
,

lib/tasks/data/truncate_data.rake
luisramos0: :+1:

luisramos0: 2 years would be 24 here, right?

sauloperez: Oh shit. I need to rest, yes :joy:

lib/tasks/data/truncate_data.rake
sauloperez: Yes, things evolved since I introduced the keyboard argument and now I see this is just redundant :+1:

lib/tasks/data/truncate_data.rake
sauloperez: bold red will do the trick :trollface:

lib/tasks/data/truncate_data.rb
sauloperez: Let me go through what think of each and see if we find a sweet spot:

* No idea what ReturnAuthorizations is. Something related to returned shipments? we need to investigate.
* StateChange: since this is useful for troubleshooting a month data seems enough.
* LogEntry: the same
* Sessions: this affects users that actively purchasing so I think 2 weeks data is more than enough. Others can afford to log in again.

lib/tasks/data/truncate_data.rake
sauloperez: LOL https://github.com/JEG2/highline/blob/bf489b632afd92c68df0b2c1f348afb66b5495aa/test/test_highline.rb#L475

luisramos0: ahah, now we have to use it :-)

lib/tasks/data/truncate_data.rb
sauloperez: I'm going to go with this.

luisramos0: there are only 5 spree_return_authorizations in aus live but we cant just delete them, there is one in 2020.
My code is a lazy shortcut, the ret auth are linked to orders.
So all you need to do is:
      sql_delete_from "spree_turn_authorizations" #{where_order_id_in_orders_to_delete}"

I havent tested but I think this will do it :+1:

sauloperez: I just check what these authorizations are. Yep, your suggestion makes sense.

lib/tasks/data/truncate_data.rake
```diff
@@ -1,81 +1,46 @@
...
-    def truncate_order_cycle_data
-      sql_delete_from "coordinator_fees #{where_oc_id_in_ocs_to_delete}"
-      sql_delete_from "
-        exchange_variants where exchange_id
-        in (select id from exchanges #{where_oc_id_in_ocs_to_delete})"
-      sql_delete_from "
-        exchange_fees where exchange_id
-        in (select id from exchanges #{where_oc_id_in_ocs_to_delete})"
-      sql_delete_from "exchanges #{where_oc_id_in_ocs_to_delete}"
+      exit unless HighLine.new.agree(message) { |q| q.default = "N" }
```
luisramos0: q is a one letter variable

lib/tasks/data/truncate_data.rake
```diff
@@ -6,6 +6,22 @@ require 'tasks/data/truncate_data'
 # This task can be used to significantly reduce the size of a database
 #   This is used for example when loading live data into a staging server
 #   This way the staging server is not overloaded with too much data
+#
+# This is also aimed at implementing data archiving. We assume data older than
+# 2 years can be safely removed and restored from a backup. This gives room for
+# hubs to do their tax declaration.
+#
+# It's a must to perform a backup right before executing this. Then, to allow
+# for a later data recovery we need to keep track of the exact moment this rake
```
luisramos0: I wonder if a timestamp is actually written to the logs when executing the task. Maybe we can log it when initializing the TruncateData service?

lib/tasks/data/truncate_data.rake
```diff
@@ -1,81 +1,46 @@
...
-      sql_delete_from "spree_adjustments where source_type = 'Spree::Payment'
-        and source_id in (select id from spree_payments #{where_order_id_in_orders_to_delete})"
-      sql_delete_from "spree_adjustments where source_type = 'Spree::LineItem'
-        and source_id in (select id from spree_line_items #{where_order_id_in_orders_to_delete})"
-    end
+    def warn_with_confirmation
+      message = <<-MSG.strip_heredoc
+      \n
+      <% highlighted_message = "This will permanently change DB contents. Please, make a backup first." %>
+      <%= color(highlighted_message, :blink, :on_red) %>
```
sauloperez: This now looks like

Screenshot from 2020-03-16 18-25-14

lib/tasks/data/truncate_data.rake
sauloperez: that would improve the situation a bit but the way to keep track of the event IMO is through a Datadog event like we do for deploys and provisions, so it doesn't get lost. Unfortunately, AU is not in Datadog, the 2nd best option we have is sending an event to Slack, I think.

luisramos0: datadog, awesome :+1:

lib/tasks/data/truncate_data.rake
luisramos0: :-)

lib/tasks/data/truncate_data.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  private
+
+  attr_reader :date
+
+  def logging
+    Rails.logger.info("TruncateData started with truncation date #{date}")
+    yield
+    Rails.logger.info("TruncateData finished")
+  end
```
sauloperez: we now see 

```sh
I, [2020-03-16 19:02:34#4746]  INFO -- : TruncateData started with truncation date 2018-03-16 20:02:34 +0100                                                                            
(...)
I, [2020-03-16 19:02:35#4746]  INFO -- : TruncateData finished

File: lib/tasks/data/truncate_data.rb [hunk 9]
```diff
@@ -36,7 +42,7 @@ def remove_transient_data
     Spree::ReturnAuthorization.delete_all
     Spree::StateChange.delete_all("created_at < '#{1.month.ago.to_date}'")
     Spree::LogEntry.delete_all("created_at < '#{1.month.ago.to_date}'")
-    sql_delete_from "sessions"
+    Session.delete_all("created_at < '#{2.weeks.ago.to_date}'")
```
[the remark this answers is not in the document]
sauloperez: I tried but this doesn't work with `delete_all` and making it work would require a combination of `where` + `delete_all` which in turn will make it harder to stub things in the test. I prefer to wait for our meeting.

luisramos0:
Hello guys, I wont be able to attend the meeting tomorrow. I leave my thoughts here:

I understand you concerns Maikel :+1:

I am 100% sure you can delete the data from the prod data as long as you keep a backup. You dont need to keep the records in your live database. I think the first point is not valid here.

Re second point, I understand the concern for sure!
- This will only touch 2 year old data, any problems arising will be related to that data. It's very unlikely this data is actually used.
- This process is already very slow, if you do it with AR I think it will never end... we would have to start from scratch with AR which would be too much work I believe.
- This process has already been tested in staging servers with live data lately, all the consistency problems I have seen there were related to data anonymization, none were related to truncation.
- I think we can run this process with live data and put it in a staging server. From there we will be able to see if data consistency is a problem.
- Lastly, in the unlikely event of seeing consistency problems in prod, we will have the backups to mitigate the situation.

sauloperez:
Thanks very much @luisramos0 ! We'll go through your ideas :ok_hand:

sauloperez:
@mkllnk and I had a chat and agreed on:

* Test this further with production data in my local env
* We'll thoroughly check the impact of data archiving with different thresholds, not only 2 years.
* We'll go through the list of destroy callbacks that may apply to these models and see if they got covered.
* Meanwhile, Aussies will talk to hubs and producers to see who checks old data regularly
* They let them know data will still be there but not online
* Then, we'll go with regular testing.

sauloperez:
Putting this on-hold until https://github.com/openfoodfoundation/openfoodnetwork/pull/5022, https://github.com/openfoodfoundation/openfoodnetwork/pull/5050, https://github.com/openfoodfoundation/openfoodnetwork/pull/5030 are out to the masses with release v2.8.1

sauloperez:
After #5030 got deployed on April 1st we see a noticeable improvement on the DB usage but there are still nasty outliers

Screenshot from 2020-04-15 19-25-56
a month data on orders_and_fulfillments report. Note we enabled APM for UK and AU on March 28th, hence the latency increase since then.

I'll wait for our meeting regarding infrastructure & performance to decide what we do with this.

sauloperez:
it's been a while since we introduced those performance improvements that blocked this PR. This wasn't discussed in last Friday's performance but I see it could be something to try it out quite easily if need to iterate once on this round of performance improvements. Triumvirate's DBs are rather bloated.

Aside from that, IMO it's a must be take `remove_transient_data` out of this PR and ship separately. These tables are very large and surely affect the overall DB performance for no reason.

Screenshot from 2020-04-28 09-28-20
Screenshot from 2020-04-28 09-28-07
Screenshot from 2020-04-28 09-27-54

Then, we could a bit of dev testing to the rest and merge it, considering it just an improvement over this staging-only rake task, and nothing meant to be run in production.

Matt-Yorkley:
I was looking at those tables the other day, they really need a clean-up. 

spree_state_changes seems to make a new record every time an order, payment, or anything with state machine changes, and records the old value and update time. Do we even need it at all? It seems like 8 million junk records.

The sessions table has really ancient records, can we add a simple rake task for deleting sesdions older than 3 months and just run it once a month (via schedule.rb)?

luisramos0:
fyi there's #3608 for that.

sauloperez:
Alright. I added one last commit and rebased onto master. Nothing has changed other than the goal. Instead of running this in production, I consider it a refactor it'll come in handy in the future, so we can merge after a bit of dev testing.

This gets us one step closer to archiving old data in production but misses the thorough testing. We can do that if the need comes up again.

luisramos0:
ok, I am ok with that approach :+1:
We should still test it with some prod db locally to detect major issues, I can try this with one of the live DBs.
I will let you know.

luisramos0:
I tested this locally with uk live data.
I loaded the data, switched payment methods to development, truncated data (no errors), verified some tables were (a lot) smaller, run the server, visited shop, checked out an order and went to the backoffice to open that order, all good :+1:
Merging.

PR #4940, opened 2020-03-07
[Spree 2.1] Fix issues in EnterprisesController

luisramos0:
#### What? Why?

Fixes two problems in Api enterprise controller. See commit message.

Fixes spec/features/consumer/registration_spec.rb:24

#### What should we test?

Green spec.

### DIFF
--- a/app/controllers/api/enterprises_controller.rb
+++ b/app/controllers/api/enterprises_controller.rb
@@ -15,7 +15,7 @@ module Api
       user_ids = params[:enterprise].delete(:user_ids)
       @enterprise = Enterprise.new(params[:enterprise])
       if @enterprise.save
-        save_enterprise_users(user_ids)
+        @enterprise.user_ids = user_ids
         render text: @enterprise.id, status: :created
       else
         invalid_resource!(@enterprise)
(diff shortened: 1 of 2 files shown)


app/controllers/api/enterprises_controller.rb
```diff
@@ -72,5 +76,15 @@ def override_sells
...
     end
+
+    def save_enterprise_users(user_ids)
+      return unless user_ids
+
+      user_ids.each do |user_id|
+        next if @enterprise.user_ids.include? user_id.to_i
+
+        @enterprise.users << Spree::User.find(user_id)
+      end
```
sauloperez: I believe we can let ActiveRecord do all this for us (and possibly a bit more efficiently). Looking at the associations Rails guide [collection_singular_ids=(ids)](https://guides.rubyonrails.org/association_basics.html#methods-added-by-has-many-collection-singular-ids-ids) seems to be what we want.

luisramos0: I started by trying
@enterprise.user_ids << user_ids

that's what collecction_singular_ids means, right?
but it doesnt work!

sauloperez: I don't know if you already did it but this method is a setter so should be used like

```ruby
@enterprise.user_ids = user_ids
```

luisramos0: ah, no, maybe that's it :face_with_head_bandage: :see_no_evil:

luisramos0: nice, it works. see new commit.
thanks for the lesson!!!

PR #4953, opened 2020-03-09
Fix problem with misssing params in schedules controller

### LINKED ISSUE #4943: [Spree 2.1] Fix params issue in schedules controller (2 specs)
rspec ./spec/features/admin/schedules_spec.rb:22
rspec ./spec/features/admin/schedules_spec.rb:68

These specs are broken since the last update of the params in the schedule controller.

Something different with the params in different scenarios:
in feature/schedule_spec:45  CREATE passes with params.permit(:name,order_cycle_ids: [])
and the schedules_controlller spec needs: params.permit(schedule: [:name, order_cycle_ids: []  ]

luisramos0:
#### What? Why?

Closes #4943

This is hack to fix the problem described in #4943
I dont know the existing mechanism (working in master) that copies params like params[:name] into params[:schedule] becoming params[:schedule][:name]
My best hypothesis is that this mechanism is not working for order_cycle_ids because it is an HABTM association in Schedule and not a basic attribute or a has_many association.

#### What should we test?
green specs in 4943

### DIFF
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -3,6 +3,7 @@ require 'open_food_network/proxy_order_syncer'
 
 module Admin
   class SchedulesController < ResourceController
+    before_filter :adapt_params, only: [:create, :update]
     before_filter :check_editable_order_cycle_ids, only: [:create, :update]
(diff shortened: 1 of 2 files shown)


luisramos0:
I did check the form, it works exactly the same and send exactly the same data to the server. the difference is only at the route/controller level.

Issue #4947, opened 2020-03-07
[Spree 2.1] Fix broken authentication spec (1 spec)


luisramos0:
rspec ./spec/features/consumer/authentication_spec.rb:66

Some problem with devise not validating that the password confirmation doesn't match...

luisramos0:
I have upgraded devise to 3.2.2 in the 3-0-latest branch and now we have these 20 extra broken specs:

1. rspec ./spec/features/admin/users_spec.rb:68 
2. rspec ./spec/features/admin/users_spec.rb:100 
3. rspec ./spec/features/admin/users_spec.rb:120
4. rspec ./spec/models/spree/user_spec.rb:76 
5. rspec ./spec/controllers/user_confirmations_controller_spec.rb:23 
6. rspec ./spec/controllers/user_confirmations_controller_spec.rb:29 
7. rspec ./spec/controllers/user_confirmations_controller_spec.rb:34 
8. rspec ./spec/controllers/user_confirmations_controller_spec.rb:39 
9. rspec ./spec/controllers/user_confirmations_controller_spec.rb:45 
10. rspec ./spec/controllers/user_confirmations_controller_spec.rb:51
11. rspec ./spec/controllers/user_confirmations_controller_spec.rb:63 
12. rspec ./spec/controllers/user_confirmations_controller_spec.rb:69 
13. rspec ./spec/features/consumer/authentication_spec.rb:66 
14. rspec ./spec/features/consumer/authentication_spec.rb:73
15. rspec ./spec/controllers/spree/users_controller_spec.rb:66
16. rspec ./spec/controllers/admin/manager_invitations_controller_spec.rb:36 
17. rspec ./spec/controllers/admin/manager_invitations_controller_spec.rb:45 
18. rspec ./spec/controllers/admin/manager_invitations_controller_spec.rb:61 
19. rspec ./spec/features/consumer/confirm_invitation_spec.rb:18 
20. rspec ./spec/controllers/user_passwords_controller_spec.rb:35

luisramos0:
Here's the PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/4952

I am not planning to work on this one right now so anyone can take this one forward from here by fixing the specs above :muscle:

luisramos0:
Looking at the broken specs I can see that upgrading devise is not fixing the original problem. It looks like we dont need to upgrade devise right now.
Maybe we can bypass my PR that upgrades devise and just fix the authentication spec.

[...]

PR #4954, opened 2020-03-09
Delete templates that are not used

luisramos0:
#### What? Why?

Found while working in #4947 

The login/signup templates are in javascripts/templates

#### What should we test?
Are there many ways a user can get to the login/signup modal?
We can make some basic verification of the login/signup process and make sure it's working as usual.
We can try to access these old login templates by trying a few URLs on the browser:
/login - should redirect to /#/login
/admin - should redirect to /#/login and then after login back to /admin

Any other ways to login?

#### Release notes
Changelog Category: Removed
Deleted dead code related to user login and signup.


sauloperez:
IMO this is small enough to move it to test ready with just an approval. So doing it.

filipefurtad0:
Hi @luisramos0,

Tested the redirection of both:
/login -> /#/login
/admin -> #/login?after_login=%2Fadmin

Both work fine. Also, resetting the password logs a user automatically in. So, using the token receiver per-email, 

Link per Email ->redirects to-> https://staging.openfoodfrance.org/user/spree_user/password/edit?reset_password_token=xxxxxxxxxxxxxxxxxxxxxxxxx

which then ->redirects to-> Dashboard

This works both for regular customers and admins. I can't think of any other ways to login.
Thanks for deleting the dead code - Moving to ready to go!

luisramos0:
bom dia!
awesome, thanks!

Issue #4962, opened 2020-03-10
Can't recreate deleted variants: [Show Deleted] button


filipefurtad0:
## Description
The [Show Deleted] button never shows results, despite having previously deleted variants:
![image](https://user-images.githubusercontent.com/49817236/76267426-9f784d80-6262-11ea-9eaf-79067b29a525.png)

![image](https://user-images.githubusercontent.com/49817236/76267509-f716b900-6262-11ea-83c3-eb4244166d4d.png)

## Expected Behavior
The deleted variants should appear.

## Actual Behaviour
Always, an empty list (no results message) appears.

## Steps to Reproduce
1. Go to products menu
2. Create a product, and several variants
3. Delete some of the previously created variants
4. Go to https://staging.katuma.org/admin/products/PRODUCTNAME/variants
5. Press the [Show Deleted]

## Workaround
A possible workaround would be creating a product and/or variants again.

## Severity

bug-s3: a feature is broken but there is a workaround

## Your Environment

Version used: v2.7.8 Striped Espargo
Browser name and version: Chrome
Operating System and version (desktop or mobile): desktop

filipefurtad0:
Just like #4961, this one is fixed by #4963, so I think I can close it.
Closing.

luisramos0:
@filipefurtad0 fyi when a PR is connected to an issue, when we merge the PR the issue is closed automatically :+1:

PR #4963, opened 2020-03-10
Fix 2 broken buttons in the product variants list page

### LINKED ISSUE #4961: [+ New Variant] button -> Snail, instead of new Variant
## Description

Using the button [+ New Variant] at the Variants menu causes no action (the spinner briefly appears, see below), generating a Bugnsnag error. A snail does not appear to the user.

## Expected Behavior
One should be redirected to the link:
https://staging.katuma.org/admin/products/PRODUCTNAME/variants/new

## Actual Behavior
Nothing much. Just the spinner flashes quickly, but nothing else is visible to the user. An error is displayed in Bugsnag 

## Steps to Reproduce

1. Go to Products
2. Edit a variant -> https://staging.katuma.org/admin/products/PRODUCTNAME/variants
3. Press the [+ New Product] button

## Workaround
Alternatively one can create a new Variant under the Products menu:
https://staging.katuma.org/admin/products/

## Severity

## Your Environment

* Version used: v2.7.8 Striped Espargo

luisramos0:
#### What? Why?

Closes #4961 and #4962

Very simple fixes to two issues. I think #4961 was a regression from removing spree_backend but I think 4962 has been there for longer...

I added a test to cover #4962 
There is already a test covering #4961 I dont understand how does it passes without this fix... I managed to replicate the problem manually and solve it with the fix...

#### What should we test?
Very both issues are fixed.

#### Release notes

Changelog Category: Changed
Fix variants list page by making both "new variant" and "show deleted" buttons work again

### DIFF
--- a/app/controllers/spree/admin/variants_controller.rb
+++ b/app/controllers/spree/admin/variants_controller.rb
@@ -59,7 +59,7 @@ module Spree
         @collection ||= if @deleted.blank?
                           super
                         else
-                          Variant.where(product_id: parent.id).deleted
+                          Variant.unscoped.where(product_id: parent.id).deleted
                         end
         @collection
       end
--- a/app/views/spree/admin/variants/index.html.haml
+++ b/app/views/spree/admin/variants/index.html.haml
@@ -49,6 +49,6 @@
   - content_for :page_actions do
     %ul.inline-menu
       %li#new_var_link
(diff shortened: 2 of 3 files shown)


filipefurtad0:
Thanks for the swift fix on this one @luisramos0!

So, the:
- [+ New Variant]  button now works great, just as before.
- [Show Deleted] now shows all the previously deleted variants. I found no way/option to recreate the previously deleted variants, but I must say I never saw this working before. So, the button does what it should :-)

Moving to ready to go!

luisramos0:
nice!

PR #4968, opened 2020-03-10
Revert changes that broke existing usage of this report

### LINKED ISSUE #4967: Order totals missing from Order Cycle Management reports
## Description

Recent change updated the OCM reports to include the paid payment total.
However this breaks the current usage for UK users.
Hence, reverting.

## Expected Behavior

The field shows the amount for the payment method used.

## Actual Behaviour

Amount field shows the amount that has been paid.

## Steps to Reproduce

1. Reports - Order cycle management report - Payment Methods
2. View the Amount field

## Animated Gif/Screenshot

image (1)

## Severity

bug s4

## Your Environment

* Version used: v2.7.7
* Browser name and version:

lin-d-hop:
#### What? Why?

Closes #4967 

Reverting a change that broke the use of this report for UK users.
The Amount field must show the value of the payment method for uncaptured payments. Meaning that we can't use payment_total in this use case.

#### What should we test?

1. Place an order with a manual method eg Cash. Do not capture this payment
2. Go to the Order Cycle Management reports.
3. Check that the Amount field in both reports populated with the value of the payment placed at checkout.

#### Release notes

Fixed a bug hiding uncaptured payments from Order Cycle Management reports

Changelog Category: Fixed

### DIFF
--- a/lib/open_food_network/order_cycle_management_report.rb
+++ b/lib/open_food_network/order_cycle_management_report.rb
@@ -77,7 +77,7 @@ module OpenFoodNetwork
        ba.phone,
        order.shipping_method.andand.name,
        order.payments.first.andand.payment_method.andand.name,
-       order.payment_total,
+       order.payments.first.amount,
        OpenFoodNetwork::UserBalanceCalculator.new(order.email, order.distributor).balance]
     end
 
@@ -92,7 +92,7 @@ module OpenFoodNetwork
        sa.phone,
        order.shipping_method.andand.name,
        order.payments.first.andand.payment_method.andand.name,
-       order.payment_total,
+       order.payments.first.amount,
        OpenFoodNetwork::UserBalanceCalculator.new(order.email, order.distributor).balance,
        has_temperature_controlled_items?(order),
        order.special_instructions]


lib/open_food_network/order_cycle_management_report.rb
```diff
@@ -77,7 +77,7 @@ def payment_method_row(order)
        ba.phone,
        order.shipping_method.andand.name,
        order.payments.first.andand.payment_method.andand.name,
-       order.payment_total,
+       order.payments.first.amount,
```
sauloperez: how would that handle the case of multiple payments? I guess that's a scenario possible to reach by creating payments from /admin

luisramos0: you could do
`order.payments.sum(&:amount)`
but that would also be incorrect because it would sum failed payments with finalized payments.
we would need a rule to use payment_total and, if no finalized payments are available, use the uncaptured payment.

could be
`(order.payment_total.nil? || order.payment_total.zero?) ? order.payments.first.amount : order.payment_total `

filipefurtad0:
Hey @lin-d-hop,

First, reproduced the issue. So, before staging:
- Placed an order with a manual method eg Cash, not capturing.
- Checked Order Cycle Management reports.

Result was:
Amount: 0
Ballance: - 66 

After Staging this PR:
Amount: 66
Ballance: - 66

Then, after manually capturing the payment:
Amount: 66
Ballance: -0

Ballance has a minus before zero, but this was already the case before this PR.

All good here, thank you!
Moving to ready to go.

emilyjeanrogers:
Assume that this will resolve this issue that was just raised in Aus: https://github.com/openfoodfoundation/openfoodnetwork/issues/5004

Will close #5004

PR #4969, opened 2020-03-10
Fix spec in rails 4 branch by making code in OrderCycleFormApplicator a bit more resilient

### LINKED ISSUE #4945: [Spree 2.1] Fix order cycles admin (1 spec)
rspec ./spec/features/admin/order_cycles_spec.rb:801 # 

Something wrong with editing order cycle, variant ids are not sent to the server?

luisramos0:
#### What? Why?

Closes #4945

In rails 4 probably due to strong params validation the variants hash comes as nil instead of an empty hash. We make this code handle that situation.

#### What should we test?
Green build is enough.

#### Release notes
Changelog Category: Changed
Adapt code to rails 4.

### DIFF
--- a/lib/open_food_network/order_cycle_form_applicator.rb
+++ b/lib/open_food_network/order_cycle_form_applicator.rb
@@ -184,6 +184,8 @@ module OpenFoodNetwork
     end
 
     def variants_to_a(variants)
+      return [] unless variants
+
       variants.select { |_k, v| v }.keys.map(&:to_i)
     end
   end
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
I had a quick look at this the other day, but I couldn't figure out why `attrs[:variants]` was empty in these tests in `3-0-stable` and not in `master`...

luisramos0:
I went this way because it is empty in master! But nil in 3-0-stable :-)

Issue #4898, opened 2020-03-02
[Spree 2.1] Fix services/restart_checkout_spec and requests/checkout/stripe_connect_spec (8 broken specs)


luisramos0:
```
 46) RestartCheckout#call when the order is in a subsequent state when order ship address is nil resets the order state, and clears incomplete shipments and payments
      Failure/Error: config.after(:each)            { DatabaseCleaner.clean }
      
      SystemStackError:
        stack level too deep
      # ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # PG::NotNullViolation:
      #   ERROR:  null value in column "calculable_type" violates not-null constraint
      #   DETAIL:  Failing row contains (596, null, 184, null, null, null).
      #   ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'

  47) RestartCheckout#call when the order is in a subsequent state when order ship address is not empty resets the order state, and clears incomplete shipments and payments
      Failure/Error: config.after(:each)            { DatabaseCleaner.clean }
      
      SystemStackError:
        stack level too deep
      # ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # PG::NotNullViolation:
      #   ERROR:  null value in column "calculable_type" violates not-null constraint
      #   DETAIL:  Failing row contains (600, null, 187, null, null, null).
      #   ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'

  48) RestartCheckout#call when the order is in a subsequent state when order ship address is empty does not reset the order state nor clears incomplete shipments and payments
      Failure/Error: config.after(:each)            { DatabaseCleaner.clean }
      
      SystemStackError:
        stack level too deep
      # ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # PG::NotNullViolation:
      #   ERROR:  null value in column "calculable_type" violates not-null constraint
      #   DETAIL:  Failing row contains (604, null, 190, null, null, null).
      #   ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'

```

These 3 specs are also failing on the clean up stage after the spec succeeds so I am adding them to the scope of this issue:
rspec ./spec/requests/checkout/stripe_connect_spec.rb:134
[...]

PR #4974, opened 2020-03-11
[Spree 2.1] Bypass broken payment rollback callback

### LINKED ISSUE #4890: [Spree 2.1] models/payments_spec and lib/open_food_network/user_balance_calculator_spec (3 broken specs)
```
25) Spree::Payment applying transaction fees to Stripe payments when the payment fails makes the transaction fee ineligible and finalizes it
      Failure/Error: config.after(:each)            { DatabaseCleaner.clean }
      
      SystemStackError:
        stack level too deep
      # ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # NoMethodError:
      #   undefined method `any?' for nil:NilClass
      #   ./app/models/spree/stock/availability_validator_decorator.rb:27:in `line_item_shipment'

  26) Spree::Payment applying transaction fees to Stripe payments when the payment information is invalid makes the transaction fee ineligible and finalizes it
      Failure/Error: config.after(:each)            { DatabaseCleaner.clean }
      
      SystemStackError:
        stack level too deep
      # ./spec/spec_helper.rb:113:in `block (2 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # NoMethodError:
      #   undefined method `any?' for nil:NilClass
      #   ./app/models/spree/stock/availability_validator_decorator.rb:27:in `line_item_shipment'

```

luisramos0:
#### What? Why?

Closes #4890 and #4898

The cause of the issues: a save statement inside an after_rollback that fails, starts a rollback inside a rollback that tries to save...
I cant figure out why the save inside the rollback is failing BUT anyway, this is just wrong so I suggest we remove this code:

We fix the problem by bypassing [the callback after_rollback](https://github.com/openfoodfoundation/spree/blob/b7ad5b473f9e38c5a1882550b2b4e348f4824f1f/core/app/models/spree/payment.rb#L35). This callback actually saves the payment: yes, a callback that saves the payment after the order transaction is rollbacked. Pretty crazy. I dont understand why fixing this at model level with callbacks was seen as a good idea in Spree. There's an alternative PR in Spree (this is also back in 2013) that fixes this at controller level that is deemed a hack...

This callback was added here:
https://github.com/spree/spree/commit/7772043a1f4e3b675d6daabea798f2a0ec2fddeb
Our build is looking great with this fix and I tested manually locally and the case discribed in the spree PR is not happening in OFN.

### DIFF
--- a/app/models/spree/payment_decorator.rb
+++ b/app/models/spree/payment_decorator.rb
@@ -12,6 +12,13 @@ module Spree
 
     localize_number :amount
 
+    # We bypass this after_rollback callback that is setup in Spree::Payment
+    # The issues the callback fixes are not experienced in OFN:
+    #   if a payment fails on checkout the state "failed" is persisted correctly
+    def persist_invalid
+      return
+    end
+
     def ensure_correct_adjustment
       revoke_adjustment_eligibility if ['failed', 'invalid'].include?(state)
       return if adjustment.try(:finalized?)


Matt-Yorkley:
Niiiiiiiiiiiiiice :smile:

Matt-Yorkley:
So we definitely don't need it? The related spec is `it "should invalidate if payment method doesnt support source"` and the comments mention issues where the credit card type is unsupported by the gateway...

PR #4978, opened 2020-03-12
Fix reloading issue in dev environment

Matt-Yorkley:
I constantly get `NameError: uninitialized constant Spree::AuthenticationHelpers` when touching local files and then reloading a page, and have to restart my rails server every time (in development). I read the other day that this is the best way to fix the issue, and it seems to work well... basically it will automatically reload the required class if and when it needs to.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,5 @@
 require 'open_food_network/referer_parser'
-require 'spree/authentication_helpers'
+require_dependency 'spree/authentication_helpers'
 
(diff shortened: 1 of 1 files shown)


sauloperez:
IMO a green build is enough. This only affects dev and it should behave as a regular `require` in production.

Some context: https://stackoverflow.com/a/5214667, https://apidock.com/rails/ActiveSupport/Dependencies/Loadable/require_dependency

luisramos0:
ok :+1:

PR #4980, opened 2020-03-12
[Spree 2.1] Line items destroy route

### LINKED ISSUE #4922: [Spree 2-1] Fix spec/controllers/line_items_controller_spec.rb:45 (1 broken spec)
Some routing issue where route is not seen if params are not provided, I tried adding json to params but it doesnt fix it, only adding an item id will make it work, which is not the test.
```

 14) LineItemsController destroying a line item on a completed order without a line item id fails and raises an error
      Failure/Error: delete :destroy
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"destroy", :controller=>"line_items"}
      # ./spec/controllers/line_items_controller_spec.rb:46:in `block (5 levels) in <top (required)>'
```

Matt-Yorkley:
Closes #4922

Removes a test from line_items_controller_spec.rb

Rails 4 does not recognise requests to destroy routes without ids as valid routes, and raises a routing error instead of a 404.

Fixes:
```
 14) LineItemsController destroying a line item on a completed order without a line item id fails and raises an error
      Failure/Error: delete :destroy

      ActionController::UrlGenerationError:
        No route matches {:action=>"destroy", :controller=>"line_items"}
      # ./spec/controllers/line_items_controller_spec.rb:46:in `block (5 levels) in <top (required)>'
```

### DIFF
--- a/spec/controllers/line_items_controller_spec.rb
+++ b/spec/controllers/line_items_controller_spec.rb
@@ -41,13 +41,6 @@ describe LineItemsController, type: :controller do
 
       before { allow(controller).to receive_messages spree_current_user: item.order.user }
 
-      context "without a line item id" do
-        it "fails and raises an error" do
-          delete :destroy
-          expect(response.status).to eq 404
-        end
-      end
-
       context "with a line item id" do
         let(:params) { { format: :json, id: item } }
 


sauloperez:
IMO that test is useless. As you say @luisramos0 that's already covered by Rails.

Matt-Yorkley:
Ok. I don't think testing the case where we are passing id as an empty string has any value. I'll update the PR to just delete the test. :+1:

luisramos0:
:+1:

Issue #4958, opened 2020-03-09
Specs in engines are not executed in CI!?


luisramos0:
### What we should change and why (this is tech debt)

There are some broken specs in ent fee summary report.
The CI build needs to run these tests so we see the broken specs, so too tasks here:
- add engine specs to semaphore jobs
- fix the broken specs

### Context
@kristinalim detected this in #4783

### Impact and timeline
This is as serious as a broken build, needs to be fixed immediately.

Matt-Yorkley:
As far as I can tell from the documentation on Rails Engines, they're not supposed to be tested in this way. They're designed to act like stand-alone components and not be tested in the full context of the app that's including them.

Maybe we need to move those specs back into the main spec folder if we want to test the engines in this way..?

Matt-Yorkley:
I had a go at adjusting the settings for knapsack / rspec to try to include our engines as well, but the results were a bit flaky.

Matt-Yorkley:
Copying the spec files from the engines back into the main /spec directory works well...

luisramos0:
:-( copying the specs back to the main app would be something to do only if we dont find any other solution. 
I thought @kristinalim had sorted this problem before when she created the orders engine.

One option I see would be to add extra commands on the build jobs for each engine. Maybe that's what Kristina did in the past?

Matt-Yorkley:
Conclusion summary:

- Engines are designed for things like `devise` that can be included in any app, but are not dependent on a specific app context
- The specs inside an engine are supposed to be run in a separate CI build. There's a way to include a small/generic "dummy" app inside the engine's spec folder to see how it integrates with the including app, but that's _not_ what we're trying to do here
- If we want to test how an engine behaves with _our_ app, we need to put those specs in the main `/spec` folder

kristinalim:
Hi @luisramos0 and @Matt-Yorkley! I added the Rake task namespace `ofn:specs:engines` and added a job for it before. I don't remember why it was temporarily removed, IIRC it was because it wasn't playing well with Knapsack.

kristinalim:
With "job", I mean Semaphore or Travis job at that time.

kristinalim:
This is the related issue: #3088

luisramos0:
IMHO the specs in the engine are not supposed to run only against the code in the engines, that would be to hard to make the engines code completely independent of the app already.

For now we can run the engine specs from root folder just like
bundle exec rspec engines/web/spec

[...]

PR #4989, opened 2020-03-12
Adapt task that runs engine specs

luisramos0:
#### What? Why?
On top of changing the semaphore config in #4958 here we are adapting the rake task that runs the engine specs to be successful, basically running the specs from the root folder with the main_app as context. Because we do this, the engine specific spec_helpers are not needed.

#### What should we test?
The rake tasks successfully runs the engine specs.

#### Release notes

Changelog Category: Changed
Fixed task that executes the tests in the rails engines.


luisramos0:
ok, rebased and tested it myself locally again. it works :+1:

luisramos0:
we have a very flaky spec spec/features/consumer/shopping/embedded_groups_spec.rb:17 
I'll merge this anyway.

PR #4992, opened 2020-03-13
[Spree 2.1] Fix routes in enterprise_fee_summaries controller spec

Matt-Yorkley:
Closes #4877

The app was loading these routes correctly, but the controller test was not. We have an odd situation where we are running a controller test from the main /spec folder, for a controller in the main app, but the routes for it are defined in an engine. 

Apparently rspec has some known issues with cases like this: https://tanzu.vmware.com/content/blog/writing-rails-engine-rspec-controller-tests

Of the three suggested solutions, one worked, one didn't, and the other involved some really messy monkey-patching...

Fixes:
```
  36) Spree::Admin::Reports::EnterpriseFeeSummariesController#new renders the report form
      Failure/Error: get :new
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"new", :controller=>"spree/admin/reports/enterprise_fee_summaries"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:16:in `block (3 levels) in <top (required)>'

  37) Spree::Admin::Reports::EnterpriseFeeSummariesController#create when the parameters are valid sends the generated report in the correct format
      Failure/Error: post :create, report: { start_at: "2018-10-09 07:30:00" }, report_format: "csv"
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"create", :controller=>"spree/admin/reports/enterprise_fee_summaries", :report=>{:start_at=>"2018-10-09 07:30:00"}, :report_format=>"csv"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:26:in `block (4 levels) in <top (required)>'

  38) Spree::Admin::Reports::EnterpriseFeeSummariesController#create when the parameters are invalid renders the report form with an error
      Failure/Error: post :create, report: { start_at: "invalid date" }, report_format: "csv"
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"create", :controller=>"spree/admin/reports/enterprise_fee_summaries", :report=>{:start_at=>"invalid date"}, :report_format=>"csv"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:36:in `block (4 levels) in <top (required)>'

  39) Spree::Admin::Reports::EnterpriseFeeSummariesController#create when some parameters are now allowed renders the report form with an error
      Failure/Error: post :create, report: { distributor_ids: [other_distributor.id] }, report_format: "csv"
      
[...]

PR #4993, opened 2020-03-13
Change engine routes to better work with specs in rails 4

### LINKED ISSUE #4877: [Spree 2.1] Fix enterprise_fee_report route issue (5 specs)
For some reason the routes in the order_management engine are not being loaded.
This causes 5 specs to fail:

```
  36) Spree::Admin::Reports::EnterpriseFeeSummariesController#new renders the report form
      Failure/Error: get :new
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"new", :controller=>"spree/admin/reports/enterprise_fee_summaries"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:16:in `block (3 levels) in <top (required)>'

  37) Spree::Admin::Reports::EnterpriseFeeSummariesController#create when the parameters are valid sends the generated report in the correct format
      Failure/Error: post :create, report: { start_at: "2018-10-09 07:30:00" }, report_format: "csv"
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"create", :controller=>"spree/admin/reports/enterprise_fee_summaries", :report=>{:start_at=>"2018-10-09 07:30:00"}, :report_format=>"csv"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:26:in `block (4 levels) in <top (required)>'

  38) Spree::Admin::Reports::EnterpriseFeeSummariesController#create when the parameters are invalid renders the report form with an error
      Failure/Error: post :create, report: { start_at: "invalid date" }, report_format: "csv"
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"create", :controller=>"spree/admin/reports/enterprise_fee_summaries", :report=>{:start_at=>"invalid date"}, :report_format=>"csv"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:36:in `block (4 levels) in <top (required)>'

  39) Spree::Admin::Reports::EnterpriseFeeSummariesController#create when some parameters are now allowed renders the report form with an error
      Failure/Error: post :create, report: { distributor_ids: [other_distributor.id] }, report_format: "csv"
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"create", :controller=>"spree/admin/reports/enterprise_fee_summaries", :report=>{:distributor_ids=>["1772"]}, :report_format=>"csv"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:50:in `block (4 levels) in <top (required)>'

  40) Spree::Admin::Reports::EnterpriseFeeSummariesController#create filtering results based on permissions applies permissions to report
      Failure/Error: post :create, report: {}, report_format: "csv"
      
      ActionController::UrlGenerationError:
        No route matches {:action=>"create", :controller=>"spree/admin/reports/enterprise_fee_summaries", :report=>{}, :report_format=>"csv"}
      # ./spec/controllers/spree/admin/reports/enterprise_fee_summaries_controller_spec.rb:67:in `block (4 levels) in <top (required)>'

```

luisramos0:
#### What? Why?

We move enterprise_fee_summary_controller and views to the order_management engine.
We move cookies_spec to the web engine.
We adapt engine routes so that they are now only extending the main_app routes (instead of creating engine specific routes in for example OrderManagement::Engine.routes).

We move the enterprise_fee_summary report to a new engine specific URL outside admin: This is where we start to see the separation between domains/modules
I think this is awesome :tada: http://0.0.0.0:3000/order_management/reports/enterprise_fee_summary/new

These two changes start a new rule: specs that need engine routes need to be in that engine.
And these changes also fix #4877 , this PR is an alternative to #4992 

#### What should we test?
We need to validate the enterprise fees summary report is working. A simple test will do.
We need to validate the cookies banner is showing. A simple test will do.
We need to have a green build :+1:

#### Release notes
Changelog Category: Changed
Adapt routes in Engines/Domains to work with specs in rails 4.


engines/order_management/spec/controllers/order_management/reports/enterprise_fee_summaries_controller_spec.rb
```diff
@@ -0,0 +1,81 @@
+require "spec_helper"
+
+describe OrderManagement::Reports::EnterpriseFeeSummariesController, type: :controller do
```
luisramos0: I changed this at the same timee as I moved the file (should have been a separate commit): the namespace here was Spree::Admin::Reports

engines/order_management/app/controllers/order_management/reports/enterprise_fee_summaries_controller.rb
```diff
@@ -0,0 +1,64 @@
+module OrderManagement
+  module Reports
+    class EnterpriseFeeSummariesController < Spree::Admin::BaseController
```
luisramos0: I changed this at the same time as I moved the file (should have been a separate commit): the namespace here was Spree::Admin::Reports

engines/order_management/spec/controllers/order_management/reports/enterprise_fee_summaries_controller_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+    end
+  end
+
+  def i18n_scope
+    "order_management.reports.enterprise_fee_summary"
+  end
+
+  def new_template_path
+    "order_management/reports/enterprise_fee_summaries/new"
```
luisramos0: I changed this at the same time as I moved the file (should have been a separate commit): the path was "spree/admin/reports/..."

engines/catalog/config/routes.rb
```diff
@@ -1,2 +1,2 @@
-Catalog::Engine.routes.draw do
```
sauloperez: I don't entirely agree with this change and the one in `config/routes.rb`. I don't like messing with the app's routes but prefer each engine to deal with its own in an isolated way from the rest. In the end, it's "just a gem" that could even be published to RubyGems. Then, the lines in `config/routes.rb`, the way I see is the explicit "installation" of those gems in our app. If we keep the analogy, just imagine using a gem that modifies your app. You would definitely complain.

luisramos0: these are not gems, I dont think the analogy and idea of "publishing to rubygems" will ever work. these engines in the end will always depend on some core modules and they will always depend on each other to some extent (how would you do order management with products from catalog? I can see how the main app will disappear or become a simple "core module".
anyway, I agree with you on the routes and I think it would be better to have engine specific routes, but I couldnt make it work...
I dont see an alternative.

luisramos0: so, would you agree if we say the routes like this are not ideal and we aim to switch back to engine specific routes (I'd create a tech debt issue for it) and we go forward like this for now?

luisramos0: :+1: issue created: #5091 5122

luisramos0:
thanks Maikel,
this is the now long discussion about engines in [discourse](https://community.openfoodnetwork.org/t/breaking-ofn-down-into-domains/1377/26).
the result was this wiki page that summarizes the vision for it:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Tech-Doc:-How-OFN-is-organized-in-Domains-using-Rails-Engines

I am not sure what exact naming you have doubts?
Is it the engines names?
I named **web** and **catalog** according to what was initially defined, Kristina has named the order_management engine. It was originally meant to be just **orders**.

yes, I'll  flag this URL change to the product team :+1:

spec/models/spree/user_spec.rb
```diff
@@ -76,9 +76,11 @@
     it "should send a confirmation email" do
       setup_email
 
-      expect do
-        create(:user, email: 'new_user@example.com', confirmation_sent_at: nil, confirmed_at: nil)
-      end.to send_confirmation_instructions
+      performing_deliveries do
```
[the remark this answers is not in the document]
luisramos0: no :see_no_evil: sorry I forgot to explain.
Here I am backporting #5048 to master.

mkllnk:
It's the order management name. I didn't expect reports to be in there and I think that we could have a separate engine just for reports. I hadn't seen that Kristina introduced that engine before, sorry.

I would love a rewrite of the reports but there are more urgent things to do right now...

luisramos0:
thanks Maikel.

Yes Kristina created order_management and added the enterprise fee summary report in there. I agree with this approach.

We can discuss this. Basically:
- option 1: have 6 domains and put specific feature code into these engines (PI into Catalog, Reports and Subscriptions into Order Management)
- option 2: have 9 or more domains: the 6 base ones plus at least 3 more domains for PI, Subs and Reports.

I prefer option 1, I think 6 domains is enough to make each small enough. I think we can extract reports and subs from OrderManagement in the future if we think the domain is too complex.

mkllnk:
> I prefer option 1, I think 6 domains is enough to make each small enough. I think we can extract reports and subs from OrderManagement in the future if we think the domain is too complex.

All good. :+1:

filipefurtad0:
Hey @luisramos0,
I'm writing briefly to point out that this PR has a few failed checks issues on codeclimate, but the build is green - is this ok to test?
Thank you.

luisramos0:
yes, pls go ahead.

filipefurtad0:
Hi @luisramos0 


- Enterprise fees summary report is working nicely. Added a few orders and the costs appear on the report.
- Cookies banner is showing is there, as usual.
- Build is green +1

Ready to go!

luisramos0:
great, thanks Filipe!

PR #4995, opened 2020-03-13
[Spree 2.1] Staging and Production configs

### LINKED ISSUE #4900: [Spree 2.1] App broken in staging
I tried deploying a Rails 4 branch to staging, and it was totally broken. I think we need to adjust some of the configs for staging and production environments.

For example we have `config.eager_load = false` for staging and production, but I think this should be `true`.

The Rails log showed some odd-looking cache entries as well, so we might need to review the settings for `memcached`..?

Matt-Yorkley:
Closes #4900 

Uses recommended eager_load settings for staging and production. From Rails itself:
```
Please update your config/environments/*.rb files accordingly:

  * development - set it to false
  * test - set it to false (unless you use a tool that preloads your test environment)
  * production - set it to true
```

Also fixes logging configs for Rails 4 in staging and production.

Already staged and tested; it works :tada:

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -38,13 +38,8 @@ Openfoodnetwork::Application.configure do
   # Note: This config no longer works with our new logging strategy
   # config.log_level = :debug
 
-  # Configure logging for Rails 3.2:
-  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(Rails.root.join("log", "#{Rails.env}.log")))
-  config.logger.level = Logger::INFO
-  config.logger.formatter = Logger::Formatter.new
-  config.logger.datetime_format = "%Y-%m-%d %H:%M:%S"
-  # Once we get to Rails 4.0, we can replace the above with:
-  #config.log_formatter = Logger::Formatter.new.tap { |f| f.datetime_format = "%Y-%m-%d %H:%M:%S" }
+  # Configure logging:
+  config.log_formatter = Logger::Formatter.new.tap { |f| f.datetime_format = "%Y-%m-%d %H:%M:%S" }
 
   # Use a different cache store in production
   memcached_value_max_megabytes = ENV.fetch("MEMCACHED_VALUE_MAX_MEGABYTES", 1).to_i
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,7 +1,7 @@
 Openfoodnetwork::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
 
-  config.eager_load = false
+  config.eager_load = true
 
   # Code is not reloaded between requests
   config.cache_classes = true
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -35,8 +35,8 @@ Openfoodnetwork::Application.configure do
   # Use https in email links
   config.action_mailer.default_url_options = { protocol: 'https' }
 
-  # Note: This config no longer works with our new logging strategy
-  # config.log_level = :debug
+  # Set log level (default is :debug in Rails 4)
+  config.log_level = :info
 
   # Configure logging:
   config.log_formatter = Logger::Formatter.new.tap { |f| f.datetime_format = "%Y-%m-%d %H:%M:%S" }
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -38,13 +38,8 @@ Openfoodnetwork::Application.configure do
   # Note: This config no longer works with our new logging strategy
   # config.log_level = :debug
 
-  # Configure logging for Rails 3.2:
-  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(Rails.root.join("log", "#{Rails.env}.log")))
-  config.logger.level = Logger::INFO
-  config.logger.formatter = Logger::Formatter.new
-  config.logger.datetime_format = "%Y-%m-%d %H:%M:%S"
-  # Once we get to Rails 4.0, we can replace the above with:
-  #config.log_formatter = Logger::Formatter.new.tap { |f| f.datetime_format = "%Y-%m-%d %H:%M:%S" }
+  # Configure logging:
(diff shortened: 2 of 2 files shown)


Matt-Yorkley:
Still not working. Delayed job took a reaaaally long time to restart, and there are weird-looking log entries relating to some of the caching:

```
D, [2020-03-13T11:37:14.977482 #4868] DEBUG -- :   SQL (0.8ms)  UPDATE "spree_preferences" SET "value_type" = $1, "updated_at" = $2, "value" = $3 WHERE "spree_preferences"."id" = 12117  [["value_type", :string], ["updated_at", Fri, 13 Mar 2020 11:37:14 GMT +00:00], ["value", "--- GBP\n...\n"]]
D, [2020-03-13T11:37:14.978339 #4868] DEBUG -- :    (0.6ms)  COMMIT
D, [2020-03-13T11:37:14.985320 #4868] DEBUG -- :   Spree::Country Load (0.7ms)  SELECT "spree_countries".* FROM "spree_countries" WHERE "spree_countries"."iso" = 'GB' LIMIT 1
D, [2020-03-13T11:37:14.997199 #4868] DEBUG -- : Cache write: /spree/app_configuration/default_country_id
D, [2020-03-13T11:37:14.999889 #4868] DEBUG -- :   Spree::Preference Load (0.9ms)  SELECT "spree_preferences".* FROM "spree_preferences" WHERE "spree_preferences"."key" = '/spree/app_configuration/default_country_id' ORDER BY "spree_preferences"."id" ASC LIMIT 1
D, [2020-03-13T11:37:15.001297 #4868] DEBUG -- :    (0.3ms)  BEGIN
D, [2020-03-13T11:37:15.004012 #4868] DEBUG -- :   SQL (0.8ms)  UPDATE "spree_preferences" SET "value_type" = $1, "updated_at" = $2, "value" = $3 WHERE "spree_preferences"."id" = 12118  [["value_type", :integer], ["updated_at", Fri, 13 Mar 2020 11:37:15 GMT +00:00], ["value", "--- 44\n...\n"]]
```
Bits like `["value", "--- GBP\n...\n"]` and `["value", "--- 44\n...\n"]` don't look right...?

Matt-Yorkley:
Ok, we're getting flat `500` responses directly from nginx (not Rails). I tried running the precompile rake task again manually and it looks like:

- it's possibly not compiling CSS at all (just JS)
- the fingerprints on the JS files that nginx is looking for don't seem to match the ones that have been created when precompiling

Output from precompile task:
- /home/openfoodnetwork/apps/openfoodnetwork/current/public/assets/application-9ef100151d29f9e23c96620e52a052cf.js
- /home/openfoodnetwork/apps/openfoodnetwork/current/public/assets/darkswarm/all-**08007d797d725e7f9242a3fc39a9cd88**.js

Output from nginx error log:
- "/home/openfoodnetwork/apps/openfoodnetwork/current/public/assets/darkswarm/all-c3b1c6b5ddc5eff742d2c7e22cd3abab.css" failed (2: No such file or directory)
- "/home/openfoodnetwork/apps/openfoodnetwork/current/public/assets/darkswarm/all-**0e4b6dd4790d17793b34c5f5bff7d3d0**.js" failed (2: No such file or directory)

luisramos0:
2 things come to my mind:
- we have turbosprockets commented out in this branch: https://github.com/ndbroadbent/turbo-sprockets-rails3
- we havent removed application.js yet in 3-0-stable, only master

Matt-Yorkley:
From the turbo-sprockets-rails3 repo:

> turbo-sprockets-rails3 was created to solve a problem with the Rails 3 asset pipeline. These problems have all been solved in Rails 4, so the gem will not be necessary

So I think we should be ok there.

Matt-Yorkley:
Ok, it looks like it was actually just the (Rails 3) logging configuration settings that were breaking everything... :see_no_evil:

luisramos0:
it's working!

luisramos0:
are we keeping the live data or shall we put simple sample data in it?

Matt-Yorkley:
I guess we can keep the data as it is for now, it'll be a more useful test to have more realistic data.

Matt-Yorkley:
I temporarily added the JS and CSS manifest changes from `master` as a quick test, but I've removed them again :+1:

RachL:
If you do change the data at some point, we should then define another server with some live data. Having one staging with "real" test cases is very useful

Matt-Yorkley:
I've added the `dev-test` label as the logging configs are changed here and we need to check for the log_level issue in Rails 4.

luisramos0:
shall we block this until we get the latest master into 3-0-stable (I am going to create a new "merge master into 3-0-stable" PR now) and then put this PR on top of it and test?

Matt-Yorkley:
The current logging settings in master will necessarily be removed by these commits anyway...

I just checked the logs in UK staging and it looks like they are currently using `:debug`. I re-checked the Rails 4 defaults and it's changed from `:info` to `:debug`...

Matt-Yorkley:
In other news it looks like staging PRs via Semaphore doesn't work currently. I think maybe Semaphore merges the PR with `master` before staging it.

This works though:
```
ansible-playbook playbooks/deploy.yml --limit uk-staging -e "pull-request-4995"
```
:+1:

Matt-Yorkley:
Tested on UK staging; the logs are pretty now, and we only get entries for `:info` level (and above).

luisramos0:
great, lets wait for a second review here just to see if any other ideas/concerns arise :+1:

Matt-Yorkley:
I've seen some intermittent issues when deploying branches, I guess we can come back to this after merging `master` into `3-0-stable` and finishing the `strong_parameters` bits...

luisramos0:
yes, thanks Matt!!

luisramos0:
the Logageddon Hotfix #5006 is in so we need to adapt this PR now.

luisramos0:
ok, I rebased, looking good, needs to be tested.

luisramos0:
ok, I staged this in uk and it's looking good, merging.

Issue #4991, opened 2020-03-12
Fix broken master build


luisramos0:
## Description
The broken specs in the order management engine were not being executed by CI so we didnt know. Now the specs are being executed and they are breaking thee master build.
We need to fix these specs:
```
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:98 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService grouping and sorting of entries groups and sorts entries correctly
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:165 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService data exclusions invalid adjustments (through 'eligible') like failed payments is included
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:200 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService data exclusions non-mandatory $0 adjustments (through 'eligible') is included
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:225 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService data exclusions $0 mandatory adjustments is included
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:272 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService handling of more complex cases with non-sender fee for incoming exchange and non-receiver fee for outgoing fetches data correctly
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:349 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService handling of more complex cases with order-based enterprise fee calculator fetches data correctly
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:412 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService filtering results based on permissions when admin includes all order cycles
rspec ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:424 # OrderManagement::Reports::EnterpriseFeeSummary::ReportService filtering results based on permissions when enterprise owner for distributor does not include unrelated order cycles
[...]

PR #4996, opened 2020-03-13
Fix broken master build

luisramos0:
#### What? Why?

Improves #4991 

Here I adapt factory code to v2.
Now there are 4 specs failing with missing data from the report, maybe some other v2 change with impact. It needs further investigation.

#### What should we test?
green build.

#### Release notes
Changelog Category: Changed
Fix broken specs that are valid but were not being executed by CI previously.

### DIFF
--- a/app/models/spree/calculator/per_item_decorator.rb
+++ b/app/models/spree/calculator/per_item_decorator.rb
@@ -13,14 +13,15 @@ module Spree
     def compute(object = nil)
       return 0 if object.nil?
 
-      preferred_amount * line_items_for(object).reduce(0) do |sum, value|
-        value_to_add = if matching_products.blank? || matching_products.include?(value.product)
-                         value.quantity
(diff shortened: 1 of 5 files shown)


sauloperez:
@luisramos0 I think I found the solution for the last failing specs in https://github.com/luisramos0/openfoodnetwork/pull/5. It needs some more work but we're pretty close. I stop now because it's Saturday :see_no_evil:

luisramos0:
ok, build is green. let's merge this.

@sauloperez this is indeed in the top 10 slowest specs 44secs:
 2.79 seconds average (44.63 seconds / 16 examples) ./engines/order_management/spec/services/order_management/reports/enterprise_fee_summary/report_service_spec.rb:3

I have created https://github.com/openfoodfoundation/openfoodnetwork/issues/5002 for that.

sauloperez:
> I have created #5002 for that.

Awesome!

PR #5005, opened 2020-03-17
Retry on payment and balance calculations

### LINKED ISSUE #4732: Reports Balance Calculation does not match User Balance Calculation

lin-d-hop:
#### What? Why?

Closes #4732

This pull request is a re-attempt at #4732.
This time the PR reverts back to the previous incarnation of not using Spree's user_balance. Instead it follows the implementation more closely.

The change was necessary as the user_balance functionality is effective on a shopper by shopper basis (as in Account -> Transactions). This method has been modified in newer Spree to be more useful in the use case of this report. In our version of Spree it is more pragmatic to follow the same logic and re-implement.

#### What should we test?

To test:

- With a shopper account create some orders and payments.
- As Hub manager make adjustments to the orders and payments. Including voiding payments and orders. Add multiple payments to an order and include addition payments so that the orders also have more payments than the order amount.
- Cancel an order with a payment.
- Void a payment on an active order
- As you make the changes in step 2 check the Shopper balance in Accounts -> Transactions. Compare this to the Balance and Payment Totals shown in Reports -> Order Cycle Management Reports -> Payment Methods Report.

In the last incarnation cancelling orders and voiding payments caused issues. I didn't note to test those cases last time. Whoops!

#### Release notes

Changelog Category: Fixed

#### Discourse thread
First step in

#### Documentation updates

### DIFF
--- a/lib/open_food_network/user_balance_calculator.rb
+++ b/lib/open_food_network/user_balance_calculator.rb
@@ -12,11 +12,11 @@ module OpenFoodNetwork
     private
 
     def completed_order_total
-      completed_orders.sum &:total
+      completed_orders.sum(&:total)
     end
 
     def payment_total
-      payments.sum &:amount
+      payments.sum(&:amount)
     end
 
     def completed_orders
@@ -28,7 +28,7 @@ module OpenFoodNetwork
     end
 
     def payments
-      Spree::Payment.where(order_id: all_orders, state: "completed")	      
+      Spree::Payment.where(order_id: all_orders, state: "completed")
     end
   end
 end
--- a/lib/open_food_network/user_balance_calculator.rb
+++ b/lib/open_food_network/user_balance_calculator.rb
@@ -6,13 +6,29 @@ module OpenFoodNetwork
     end
 
     def balance
-      -completed_orders.sum(&:outstanding_balance)
+      payment_total - completed_order_total
     end
 
     private
 
+    def completed_order_total
+      completed_orders.sum &:total
+    end
+
+    def payment_total
+      payments.sum &:amount
+    end
+
     def completed_orders
-      Spree::Order.where(distributor_id: @distributor, email: @email).complete.not_state(:canceled)
+      all_orders.complete.not_state(:canceled)
+    end
+
+    def all_orders
+      Spree::Order.where(distributor_id: @distributor, email: @email)
+    end
+
+    def payments
+      Spree::Payment.where(order_id: all_orders, state: "completed")	      
     end
   end
 end
--- a/lib/open_food_network/order_cycle_management_report.rb
+++ b/lib/open_food_network/order_cycle_management_report.rb
@@ -77,7 +77,7 @@ module OpenFoodNetwork
        ba.phone,
        order.shipping_method.andand.name,
        order.payments.first.andand.payment_method.andand.name,
-       order.payment_total,
+       order.payments.first.amount,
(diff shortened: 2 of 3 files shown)


lib/open_food_network/user_balance_calculator.rb
```diff
@@ -6,13 +6,29 @@ def initialize(email, distributor)
...
-      Spree::Order.where(distributor_id: @distributor, email: @email).complete.not_state(:canceled)
+      all_orders.complete.not_state(:canceled)
+    end
+
+    def all_orders
+      Spree::Order.where(distributor_id: @distributor, email: @email)
+    end
+
+    def payments
+      Spree::Payment.where(order_id: all_orders, state: "completed")
```
luisramos0: so you are putting back what was there before, right?
https://github.com/openfoodfoundation/openfoodnetwork/pull/4733/files

to do this correctly using git you should get a fresh branch and run:
git revert baf1ecb4366159f063d69fdc77b54c493799406d
commit this revert and then do your new changes.
that would make the change explicit and easy to follow.
this way we dont know what is revert or what are you changing now.

But I think this is simple enough to get us there, I think I figured it out: you reverted and changed this payments method to use all_orders instead of completed_orders, is this correct explanation? is that all?

lib/open_food_network/user_balance_calculator.rb
```diff
@@ -6,13 +6,29 @@ def initialize(email, distributor)
...
+    def payment_total
+      payments.sum &:amount
+    end
+
     def completed_orders
-      Spree::Order.where(distributor_id: @distributor, email: @email).complete.not_state(:canceled)
+      all_orders.complete.not_state(:canceled)
+    end
+
+    def all_orders
```
luisramos0: this would better called user_orders

lib/open_food_network/user_balance_calculator.rb
```diff
@@ -6,13 +6,29 @@ def initialize(email, distributor)
...
+    end
+
+    def payment_total
+      payments.sum &:amount
+    end
+
     def completed_orders
-      Spree::Order.where(distributor_id: @distributor, email: @email).complete.not_state(:canceled)
+      all_orders.complete.not_state(:canceled)
+    end
```
luisramos0: I'd call this 
def completed_and_not_canceled_orders

lib/open_food_network/user_balance_calculator.rb
```diff
@@ -6,13 +6,29 @@ def initialize(email, distributor)
...
+
+    def all_orders
+      Spree::Order.where(distributor_id: @distributor, email: @email)
+    end
+
+    def payments
```
luisramos0: I'd add a comment to def payments
 `# Lists all complete user payments including payments in incomplete or canceled orders`

lib/open_food_network/user_balance_calculator.rb
lin-d-hop: yeah that's it. Payments are read from all_orders and order totals only from completed orders. This was causing issues on payments received on cancelled orders in the first incarnation.

lin-d-hop:
@luisramos0 
Thanks for all the feedback. I'll do all those things and try again :-)

lin-d-hop:
Replacing...

PR #5010, opened 2020-03-17
Remove simple_form

Matt-Yorkley:
It looks like we don't use `simple_form` at all, so we can remove it.

This was discovered during the upgrade in #4865

#### What should we test?

We're pretty sure this will do nothing at all, but there's a small chance that this will affect the way forms are displayed, so we need a quick sanity check that forms look and function as before. 

- Find some forms in public and admin pages
- Check they're not visually displaying in a weird way
- Fill them out and check they submit correctly

#### Release Notes

Removed unused simple_form gem and related files


luisramos0:
I think you can also remove:
- config/initializers/simple_form.rb
- config/locales/simple_form.en.yml

filipefurtad0:
Hey @Matt-Yorkley ,

The notes below on behalf of Daniel Ekow Melfah, giving here his first contribution as a tester :-) Welcome Daniel! 🎉 

"I have filled out forms in both the public pages and admin pages. Submitting the forms work successfully on the staging branch. I also performed some of the sanity checks listed in the doc. I think the PR is good to go."

Moving to ready to go!

PR #5016, opened 2020-03-19
Revert "Inner join visible orders"

mkllnk:
Reverts openfoodfoundation/openfoodnetwork#4879

The attempt to improve performance failed as described in https://github.com/openfoodfoundation/openfoodnetwork/issues/4782#issuecomment-600686410.

@luisramos0 accepted this PR on Slack and @sauloperez initiated this. So I will see this as reviewed and it doesn't need testing because it's going back to old, well-tested behaviour.


sauloperez:
Thanks for taking care of it @mkllnk :1st_place_medal: Sometimes I'm glad we have a team oversees :joy:

PR #5023, opened 2020-03-20
Display shipping and billing address on invoices

### LINKED ISSUE #5019: Invoice displaying the enterprise address instead of the customer's address
## Expected Behaviour
The customer's address displays under their name on the invoice.

## Actual Behaviour
The enterprise's address displays instead, even though it's showing the correct name and email/phone number.

## Animated Gif/Screenshot
Image from iOS (2)

## Workaround
Go into the application and find the customer address and write it on the invoice. But this is a shit workaround when you have A LOT of orders.

## Severity
I'm calling this a sev-2 as the workaround is pretty unacceptable and this is a core part of the fulfillment process.

**bug-s2: a non-critical feature is broken, no workaround**

-->

## Your Environment

* Version used:
* Browser name and version:

mkllnk:
#### What? Why?

Closes #5019

We have two invoice templates. Each instance can decide which template to use. Australia uses the default template and France uses the alternative template for legal compliance.

Both templates used to display the customer's shipping address and the invoice was also used as packing sheet and for delivery. The French template displays the customer's billing address since #4560 for legal compliance.

If orders are not delivered but picked up at a hub, the shipping address is set to the hub's address. In that case the Australian template was confusing because it displayed the hub's address as the customer's address. In case of home delivery, we still want the shipping address though.

We decided to display both, the customer's billing address and the shipping address in case of delivery to a different address. The hub's address is always displayed on the left side under "From". To make it more clear, we changed the layout a bit and moved special delivery instructions to the top. That should be easier to see during delivery.

New layout with customer code:
Screenshot from 2020-03-20 15-27-10

Different addresses:
Screenshot from 2020-03-20 15-34-46

With instructions:
Screenshot from 2020-03-20 15-27-45

Pickup with proper layout as PDF:
Screenshot from 2020-03-20 15-54-16

#### What should we test?

* Log in as super-admin and check under Configuration that the invoice template is the default template, not the alternative one.
* Set up an order cycle with two shipping methods: one delivery and one pickup
* Verify three cases:
  - Pickup: The user's billing address shows on the left.
  - Delivery to separate address: Billing and delivery address show.
  - Special instructions: They show at the top and not at the bottom of the invoice.

#### Release notes

The default invoice template (used by some instances) now displays the billing address and shipping details in the header.

Changelog Category: Changed

### DIFF
--- a/app/views/spree/admin/orders/invoice.html.haml
+++ b/app/views/spree/admin/orders/invoice.html.haml
@@ -17,7 +17,7 @@
       %td{ :align => "right" }
         %h4= @order.order_cycle.andand.name
     %tr{ valign: "top" }
-      %td{ :align => "left" }
+      %td{ align: "left", colspan: 3 }
         %strong= "#{t('.from')}: #{@order.distributor.name}"
         - if @order.distributor.abn.present?
           %br
@@ -26,10 +26,14 @@
         = @order.distributor.address.full_address
         %br
         = @order.distributor.contact.email
-      %td{width: "10%" }
+    %tr{ valign: "top" }
+      %td{ colspan: 3 }
         &nbsp;
-      %td{ :align => "right" }
-        %strong= "#{t('.to')}: #{@order.bill_address.full_name}"
+    %tr{ valign: "top" }
+      %td{ align: "left" }
+        %strong= "#{t('.to')}:"
+        %br
+        = @order.bill_address.full_name
         - if @order.andand.customer.andand.code.present?
           %br
           = "#{t('.code')}: #{@order.customer.code}"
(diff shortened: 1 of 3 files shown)


app/views/spree/admin/orders/invoice.html.haml
```diff
@@ -26,30 +26,42 @@
...
+          %br
+          = @order.ship_address.full_name
+          %br
+          = @order.ship_address.full_address
+          %br
+          = @order.ship_address.phone
+        - if @order.special_instructions.present?
+          %br
+          %br
+            %strong= t :customer_instructions
```
luisramos0: we have parenthesis everywhere in the code in cases like this, I'd keep with local idiom
t(:customer_instructions)

EDIT: I now see you just copied it from below :-)

app/views/spree/admin/orders/invoice.html.haml
```diff
@@ -26,30 +26,42 @@
...
+        = @order.bill_address.full_address
         %br
         - if @order.andand.customer.andand.email.present?
           = "#{@order.customer.email},"
         = "#{@order.bill_address.phone}"
+      %td
+        &nbsp;
+      %td{ align: "left", style: "border-left: .1em solid black; padding-left: 1em" }
+        %strong= "#{t('.shipping')}: #{@order.shipping_method&.name}"
+        - if @order.shipping_method&.require_ship_address
```
luisramos0: we have a better helper:
.delivery?

lin-d-hop:
Can we get this in the next release please?
@filipefurtad0 a testing priority?
Thanks!!

filipefurtad0:
Hi @mkllnk and @lin-d-hop,

Tested as below:
- Superadmin: Checked that the default Invoice option is activated
- Admin: Set up order cycles and the delivery and pick-up options
- User: Checked-out according to the three test cases

pickup: The user's billing address appears on the left
delivery: Billing (left) and delivery (right) address show.
w/special instructions: They show at the top, under the delivery address, as described in this PR.

This is good to go.

Issue #5029, opened 2020-03-20
Timeout on Packing Reports


lin-d-hop:
## Description

Due to unprecedented demand this week one of our hubs had a timeout on the Packing Reports.

User: Tamar Valley Food Hubs
OC: TVFH 20200320
Report: Packing Reports - Pack By Customer
Date run: 19/03/20

## Expected Behavior

No timeout

## Actual Behaviour

Timout

## Steps to Reproduce

1. admin/reports
2. Pack By Customer report
3. Slug

## Severity

Not sure.
This week they worked around. 
Next week we are expecting sales to grow further, based on the level of orders we've already seen for them for their next order cycle (it closes next Weds!)

I'm setting as an S3 but it will be an S1 as soon as they can't get it to run.

## Your Environment

* Version used: UK 2.7.9
* Browser name and version:
* Operating System and version (desktop or mobile): Desktop

luisramos0:
We should run this locally with full uk data and report what are the results before and after
 #5016 
That will show us what will be the impact of the next release in this report.

daniellemoorhead:
Hey @lin-d-hop is this the next priority S3 for UK? We moved it to the backlog column but doesn't have an instance label on it, I can't remember if it was supposed to be labelled yours or should be in All the Things?

lin-d-hop:
THis is outdated and is being looked at in other work currently. Closing

PR #5030, opened 2020-03-20
Improve reports performance

Matt-Yorkley:
#### What? Why?

Related to #5029 and #4782

I started playing with Aus production data locally, and had a look at reports again.

From my initial idea in Slack:

> I'm sure we're missing something. It's a problem of combining different sets, right? We build these sets like "editable_orders", "editable_line_items" and then combine them in a huge query but I think we can reduce the scope of those sets before building them...

> when we search reports we have filters that limit the results to a date range, set of distributors, etc.

> but it seems like we're applying that only after doing lots of big query building that includes things like "all the line items a user can edit, ever", which could be 50,000+

**TLDR;** if we add the scoping from the search filters into the logic of building sets like `editable_line_items` _before_ combining them into the query, it reduces page loads in various reports by **50-90%**, and the horrifying amount of `line_item` ids added to the `IN` clause is reduced to a reasonable amount.

#### What should we test?

- Reports are all working in general. 
- Should be loading significantly faster than before (on a server with lots of data).

This PR affects:
- **bulk_coop** reports
- **order_and_distributor** reports
- **orders_and_fulfillments** reports
- **packing** reports

#### Release notes

Improve reports performance

Changelog Category: Changed

### DIFF
--- a/app/services/permissions/order.rb
+++ b/app/services/permissions/order.rb
@@ -2,24 +2,28 @@ require 'open_food_network/permissions'
 
 module Permissions
   class Order
-    def initialize(user)
+    def initialize(user, search_params = nil)
       @user = user
       @permissions = OpenFoodNetwork::Permissions.new(@user)
+      @search_params = search_params
     end
 
     # Find orders that the user can see
     def visible_orders
-      Spree::Order.
+      orders = Spree::Order.
         with_line_items_variants_and_products_outer.
         where(visible_orders_where_values)
+      orders = orders.complete.not_state(:canceled).search(search_params).result if search_orders?
+      orders
     end
 
     # Any orders that the user can edit
     def editable_orders
-      Spree::Order.where(
-        managed_orders_where_values.
-          or(coordinated_orders_where_values)
-      )
+      orders = Spree::Order.
+        where(managed_orders_where_values.
+          or(coordinated_orders_where_values))
+      orders = orders.complete.not_state(:canceled).search(search_params).result if search_orders?
+      orders
     end
 
     def visible_line_items
@@ -34,6 +38,12 @@ module Permissions
     end
 
     private
+
+    attr_reader :search_params
+
+    def search_orders?
+      search_params.present?
+    end
 
     def visible_orders_where_values
       # Grouping keeps the 2 where clauses from produced_orders_where_values inside parentheses
--- a/app/services/permissions/order.rb
+++ b/app/services/permissions/order.rb
@@ -13,8 +13,8 @@ module Permissions
       orders = Spree::Order.
         with_line_items_variants_and_products_outer.
         where(visible_orders_where_values)
-      orders = orders.complete.not_state(:canceled).search(search_params).result if search_orders?
-      orders
+
+      filtered_orders(orders)
     end
 
     # Any orders that the user can edit
@@ -22,8 +22,8 @@ module Permissions
       orders = Spree::Order.
         where(managed_orders_where_values.
           or(coordinated_orders_where_values))
-      orders = orders.complete.not_state(:canceled).search(search_params).result if search_orders?
-      orders
+
+      filtered_orders(orders)
     end
 
     def visible_line_items
@@ -41,7 +41,13 @@ module Permissions
 
     attr_reader :search_params
 
-    def search_orders?
+    def filtered_orders(orders)
+      return orders unless filter_orders?
+
+      orders.complete.not_state(:canceled).search(search_params).result
+    end
+
+    def filter_orders?
       search_params.present?
     end
 
--- a/lib/open_food_network/reports/line_items.rb
+++ b/lib/open_food_network/reports/line_items.rb
@@ -12,9 +12,7 @@ module OpenFoodNetwork
       end
 
       def list(line_item_includes = nil)
-        line_items = @order_permissions.
-          visible_line_items.
-          merge(Spree::LineItem.where(order_id: orders.result))
+        line_items = @order_permissions.visible_line_items.in_orders(orders.result)
 
         if @params[:supplier_id_in].present?
           line_items = line_items.supplied_by_any(@params[:supplier_id_in])
--- a/app/models/spree/line_item_decorator.rb
+++ b/app/models/spree/line_item_decorator.rb
@@ -36,6 +36,10 @@ Spree::LineItem.class_eval do
     end
   }
 
+  scope :in_orders, lambda { |orders|
+    where(order_id: orders)
+  }
+
   # Find line items that are from order sorted by variant name and unit value
   scope :sorted_by_name_and_unit_value, -> {
     joins(variant: :product).
--- a/lib/open_food_network/bulk_coop_report.rb
+++ b/lib/open_food_network/bulk_coop_report.rb
@@ -128,7 +128,7 @@ module OpenFoodNetwork
 
     def order_permissions
       return @order_permissions unless @order_permissions.nil?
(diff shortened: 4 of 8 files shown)


app/services/permissions/order.rb
```diff
@@ -2,24 +2,26 @@
 
 module Permissions
   class Order
-    def initialize(user)
+    def initialize(user, params = nil)
```
luisramos0: we should only send params down to service classes if really needed. This is not the case. We dont need the full params object here, only params[:q]. I'd say you need to pass params[:q] and give it a good name at this service class level.
This params[:q] is specific to ransack, right? Maybe order_search_params or ransack_order_search_params?
This will make
def search_orders?
  order_search_params.present?
end

app/services/permissions/order.rb
```diff
@@ -2,24 +2,26 @@
...
+      orders
     end
 
     # Any orders that the user can edit
     def editable_orders
-      Spree::Order.where(
-        managed_orders_where_values.
-          or(coordinated_orders_where_values)
-      )
+      orders = Spree::Order.where(managed_orders_where_values.or(coordinated_orders_where_values))
```
luisramos0: I think this is so complex that it deserves the breaklines :-)

app/services/permissions/order.rb
```diff
@@ -2,24 +2,26 @@
...
+      orders = orders.complete.not_state(:canceled).search(@params[:q]).result if search_orders?
```
luisramos0: Is this change correct? Should the reports only list complete not cancelled orders?

luisramos0: did you verify the SQL produced? Does it look good as part of the massive query?

lib/open_food_network/reports/line_items.rb
```diff
@@ -12,9 +12,7 @@ def orders
       end
 
       def list(line_item_includes = nil)
-        line_items = @order_permissions.
-          visible_line_items.
-          merge(Spree::LineItem.where(order_id: orders.result))
+        line_items = @order_permissions.visible_line_items.in_orders(orders.result)
```
luisramos0: I think this is not correct, I cant remember why. I think it's because of the SQL that visible_line_items generates, you cannot simply add a where clause, you need to use merge, which does something else, I cant remember what...

luisramos0: I am sure I have tried this before and it didnt work. I cant remember why....

app/services/permissions/order.rb
luisramos0: if this is correct, I think it needs a spec.

lib/open_food_network/reports/line_items.rb
Matt-Yorkley: I know there can be issues with multiple #where calls directly in the same query chain, but I think its different if they're in named scopes?

app/services/permissions/order.rb
Matt-Yorkley: > did you verify the SQL produced? Does it look good as part of the massive query? 

I checked the actual output of the queries with real data, using rack-mini-profiler both before and after.

app/services/permissions/order.rb
Matt-Yorkley: Yes, I need to tidy up a bit.

lib/open_food_network/reports/line_items.rb
luisramos0: I didnt mean there would be a problem on the rails side building the query. I meant that I think the query that results from this is not correct for what we need.

app/services/permissions/order.rb
luisramos0: ok, great. that means you validated the replacement of the merge with the names scope as well.

Matt-Yorkley: > Is this change correct? Should the reports only list complete not cancelled orders?

All the reports this PR affects use `Reports::LineItems` to get their line items, which uses this method (unchanged):
https://github.com/openfoodfoundation/openfoodnetwork/blob/981778bc98ce43dde6ff57da8272b997038745be/lib/open_food_network/reports/line_items.rb#L48-L50

lib/open_food_network/reports/line_items.rb
Matt-Yorkley: The #merge here gives us the intersection of both queries, right? So we could write it like this:
```ruby
line_items_in_orders_result = Spree::LineItem.where(order_id: orders.result)
visible_line_items = @order_permissions.visible_line_items
visible_line_items_in_orders_result = visible_line_items.merge(line_items_in_orders_result)
```

Is that not equivalent to `@order_permissions.visible_line_items.in_orders(orders.result)`?

Matt-Yorkley:
Comparison of SQL output with this PR and without it (loading one day of data, non-superadmin user for a large enterprise, with AUS data):

https://gist.github.com/Matt-Yorkley/7a1a711178bc0e7b38e5a65f25953a61

daniellemoorhead:
@mkllnk @sauloperez would be great to have another reviewer on this, feels super important to have this PR merged for the next release candidate.

app/services/permissions/order.rb
Matt-Yorkley: > if this is correct, I think it needs a spec

Specs added :+1:

lib/open_food_network/reports/line_items.rb
Matt-Yorkley: I called #to_sql on both versions and the output matches :+1:

app/services/permissions/order.rb
```diff
@@ -35,6 +39,12 @@ def editable_line_items
 
     private
 
+    attr_reader :search_params
+
+    def search_orders?
+      search_params.present?
+    end
```
mkllnk: I'm not sure with ransack but maybe this could be simpler. I'm imagining using this method in the cases above:

```suggestion
    def filtered_orders
      if search_params.present?
        Spree::Order.complete.not_state(:canceled).search(search_params).result
      else
        Spree::Order.scoped
      end
    end
```

```rb
    def editable_orders
      filtered_orders.where(
        managed_orders_where_values.
          or(coordinated_orders_where_values)
      )
    end
```

Matt-Yorkley:
> When you measured the performance improvement, did you also check that the reports had exactly the same output?

Yes, I ran some reports with a lot of results, and diffed the final report output.

app/services/permissions/order.rb
Matt-Yorkley: I tried making these _exact_ changes and there were some issues with the specs; the returned results were not correct. 

I've added a new commit where it's refactored in a fairly similar way, adding a #filtered_orders method. Also renamed #search_orders? to #filter_orders?

sauloperez:
@Matt-Yorkley would you mind running an `EXPLAIN ANALYZE` on the production server you took those queries you shared in the gist? It'd cool to then have the result on another gist plus a link to https://explain.dalibo.com/. This way we can use this PR as a baseline next time. I found that very useful while doing my investigations.

filipefurtad0:
Hey @Matt-Yorkley,

Amazing!

In AU-staging, as superadmin:
- bulk_coop reports and order_and_distributor reports ->> Load times dropped by a factor of 3. 
- orders_and_fulfillments reports and packing reports ->> still require more than 2 sec load time (snail). However, this was already the case before, and also I verified that in a regular admin account these reports work fine and are notriously faster.

Other reports seem to work fine as well.

I'd say this is a huge improvement :tada: 
Moving to ready-to-go.

sauloperez:
I'm so looking forward to this!! :scream:

PR #5031, opened 2020-03-21
Retry on payment and balances calculation (again)

### LINKED ISSUE #4732: Reports Balance Calculation does not match User Balance Calculation

lin-d-hop:
#### What? Why?
Closes #4732
Replaces #5005

This pull request is a re-attempt at #4732 and a reattempt at #5005 including @luisramos0's feedback.

The PR reverts back to the previous incarnation of not using Spree's user_balance. Now it follows the spree implementation more closely.

The change was necessary as the spree user_balance functionality is only effective on a shopper by shopper basis (as in Account -> Transactions). This method has been modified in newer Spree to be more useful in the use case of this report. In our version of Spree it is more pragmatic to follow the same logic and re-implement.

#### What should we test?

To test:

- With a shopper account create some orders and payments.
- As Hub manager make adjustments to the orders and payments. Including voiding payments and orders. Add multiple payments to an order and include addition payments so that the orders also have more payments than the order amount.
- Cancel an order with a payment.
- Void a payment on an active order
- As you make the changes in step 2 check the Shopper balance in Accounts -> Transactions. Compare this to the Balance and Payment Totals shown in Reports -> Order Cycle Management Reports -> Payment Methods Report.

In the last incarnation cancelling orders and voiding payments caused issues. I didn't note to test those cases last time. Whoops!

#### Release notes

Changelog Category: Fixed

#### Discourse thread
First step in

#### Documentation updates

### DIFF
--- a/spec/lib/open_food_network/user_balance_calculator_spec.rb
+++ b/spec/lib/open_food_network/user_balance_calculator_spec.rb
@@ -69,14 +69,70 @@ module OpenFoodNetwork
           create(:order_with_totals_and_distribution,
                  user: user1, distributor: hub1,
                  completed_at: 1.day.ago, state: "canceled")
-        } # total=10
+        } # total=13 (10 + 3 shipping fee)
         let!(:p4) {
           create(:payment, order: o4, amount: 20.00,
                            state: "completed")
         }
 
         it "does not include canceled orders in the balance" do
-          expect(UserBalanceCalculator.new(o4.email, hub1).balance).to eq(-9) # = 15 + 2 - 13 - 13
+          expect(UserBalanceCalculator.new(o4.email, hub1).balance).to eq(11) # = 15 + 2 + 20 - 13 - 13
+        end
+      end
+
+      context "with void payments" do
+        let!(:o4) {
+          create(:order_with_totals_and_distribution,
+                 user: user1, distributor: hub1,
+                 completed_at: 1.day.ago)
+        } # total=13 (10 + 3 shipping fee)
+        let!(:p4) {
+          create(:payment, order: o4, amount: 20.00,
+                           state: "void")
+        }
+
+        it "does not include void in the balance" do
+          expect(UserBalanceCalculator.new(o4.email, hub1).balance).to eq(-22) # = 15 + 2 - 13 - 13 - 10
+        end
+      end
+
+      context "with invalid payments" do
+        let!(:o4) {
+          create(:order_with_totals_and_distribution,
+                 user: user1, distributor: hub1,
+                 completed_at: 1.day.ago)
+        } # total=13 (10 + 3 shipping fee)
+        let!(:p4) {
+          create(:payment, order: o4, amount: 20.00,
+                           state: "invalid")
+        }
+
+        it "does not include invalid payments in the balance" do
+          expect(UserBalanceCalculator.new(o4.email, hub1).balance).to eq(-22) # = 15 + 2 - 13 - 13 - 10
+        end
+      end
+
+      context "with multiple payments on single order" do
+        let!(:o4) {
+          create(:order_with_totals_and_distribution,
+                 user: user1, distributor: hub1,
+                 completed_at: 1.day.ago)
+        } # total=13 (10 + 3 shipping fee)
+        let!(:p4) {
+          create(:payment, order: o4, amount: 4.00,
+                           state: "completed")
+        }
+        let!(:p5) {
+          create(:payment, order: o4, amount: 5.00,
+                           state: "completed")
+        }
+        let!(:p6) {
+          create(:payment, order: o4, amount: 6.00,
+                           state: "completed")
+        }
+
+        it "includes orders with multiple payments in the balance" do
+          expect(UserBalanceCalculator.new(o4.email, hub1).balance).to eq(-7) # = 15 + 2 + 4 + 5 + 6 - 13 - 13 - 10
         end
       end
     end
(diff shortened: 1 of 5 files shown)


spec/lib/open_food_network/user_balance_calculator_spec.rb
```diff
@@ -76,7 +76,7 @@ module OpenFoodNetwork
         }
 
         it "does not include canceled orders in the balance" do
-          expect(UserBalanceCalculator.new(o4.email, hub1).balance).to eq(-9) # = 15 + 2 - 13 - 13
+          expect(UserBalanceCalculator.new(o4.email, hub1).balance).to eq(11) # = 15 + 2 + 20 - 13 - 13
```
Matt-Yorkley: There's lots of single-letter variables in this spec like `p4`, `o3`. Can we use more descriptive naming like `order_cancelled` instead of `o4`? Or at least `order4`?

Matt-Yorkley: There's also a lot of **magic numbers**. The `10 + 3` numbers in the comments are default values from the `order_with_totals_and_distribution` factory, right?

Instead of the comment `# = 15 + 2 + 20 - 13 - 13`, could we make it more clear/explicit where all the numbers are coming from? Something like:
```ruby
expected_balance = payment1.amount + payment2.amount + payment4.amount - order1.total - order2.total
expect(UserBalanceCalculator.new(user1.email, hub1).balance).to eq(expected_balance)
```

Maybe we could name `p4` as `cancelled_order_payment` for extra clarity?

lin-d-hop:
@Matt-Yorkley 
While I appreciate your feedback I really don't have time right now and this is an actual bug for an actual user.
I didn't make the specs worse - in fact I improved them - we just aren't taking the opportuntity to make them perfect.
Please can I be let off right now?
:pray:

spec/lib/open_food_network/user_balance_calculator_spec.rb
lin-d-hop: Agree on the single letter variables. The magic numbers are only in comments. They could just as easily be deleted and no one would notice.

Matt-Yorkley:
@lin-d-hop I added a quick tidy-up for the spec and pushed it to your branch.

lin-d-hop:
Thanks @Matt-Yorkley 
#priorities ;-)

Matt-Yorkley:
I'll kick this into `test ready` for now. Maybe @luisramos0 wants to review the spec refactor commit?

spec/lib/open_food_network/user_balance_calculator_spec.rb
luisramos0: now that I re-review I see the name of this test should be better, it says it does not include cancelled orders but it includes payments in cancelled orders.
it should be more specific " it "includes complete payments in canceled orders but does not include the canceled orders amount" do"

luisramos0:
ok, I can change it, we make it a group PR!

lin-d-hop:
This is why the OFN code base is such high quality...

luisramos0:
ok, done, ready for testing if you dont disagree with the description of the spec :-)

filipefurtad0:
Hi @lin-d-hop and @lbwright22,

Here are my test notes:
https://docs.google.com/document/d/17p063PzhSKk1pf-4Z8ucHeKD2IRiBOgQd9sJRTuQoTg/edit#

I must say, I did not find this testing easy at all - but maybe I lack experience with the platform - so I'm happy to get more feedback on this one.

lbwright22:
Thanks @filipefurtad0 
I have the same findings and hence this issue is not resolved @luisramos0 @lin-d-hop 
https://docs.google.com/document/d/1vDNDCaDmfMww02FqE1oJ4wNpqHn56kpuI8gqVbQTTYg/edit?usp=sharing

Customer's balance is correct and consistent between the customer transaction history, orders and payment method reports for cases when payments are void-ed BUT not when orders are cancelled.

Cancelled orders result in the payments associated with that order are lost from the customer's interface and no evidence of them.  The payment does contribute to their user balance on the payment method report though.

There is also an anomaly when the customer's account is in credit overall (see document)

filipefurtad0:
Great testing @lbwright22, welcome!

So in this case, and whenever we find the issue at hand is not solved, we move the PR back to In-Dev (which I just did). If all is well, we move it further down the pipe to "Ready to go".

I am also removing the AU-staging tag.

Thanks again your help!

sigmundpetersen:
Shall we close this @lin-d-hop ? Or do you foresee work continuing on this any time soon?

luisramos0:
I am looking at this PR now.
This PR looks good :+1:  My first hypothesis is that it is actually making things better. Just not yet perfectly handling cancelled orders.

Note about resumed orders: exploring cancelled orders I realize a cancelled order goes from state complete to cancelled BUT if resumed, it goes to a state resumed... not completed. I think most of our code is not taking into account resumed orders, only complete ones. I think this deserves a separate issue: handling resumed orders across the system, in particular in reports and in specific in terms of impact on user balances.

For now, I'll ignore this case of resumed orders and just make sure this all works for **cancelled orders: their totals should not be considered for user balances but their payments should**.

luisramos0:
ok, I think there's nothing to be fixed in this PR. It does fix the report balance correctly but the user account page still shows the wrong result. This PR makes the report more correct. We need to fix the user account page now.

The easier way to add completed payments of canceled orders to the user account page is to add the canceled orders to it, like this:
image

This is what the PR is currently doing. You can stage and have a look :+1: 
I think the biggest issue is that this page doesn't include the order state, just the shipment state, in this screenshot we could infer the last order is cancelled only by the shipment state, but that's not enough... what do we do? can we just add the order state here?
What do you think about this approach? Any alternatives?

luisramos0:
Tech note: the code behind this user account page is old and has been refactored more recently (2017 :-)) but it is not well structured. It could use a good clean up, for example, it should depend on the API instead of relying on a number of internal injectors of data and serializers...

PR #5044, opened 2020-03-22
[Spree 2.1] Re-add spring and newrelic_rpm

luisramos0:
#### What? Why?

From #4777 
We un-comment spring and newrelix_rpm from the gemfile, they are not causing trouble any more.
Spring is working really well actually!

#### What should we test?
green build


Matt-Yorkley:
FYI I tried out the newrelic rpm in `master`, and it didn't seem to be working. I think it needs a bit more configuration. Lets go ahead with this though :+1:

Issue #5049, opened 2020-03-23
Brief downtimes on French prod


Matt-Yorkley:
All servers are experiencing record usage, and France specifically has had 2 small downtimes in the last 24h.

We need to investigate the data a bit to find the pain points and then maybe plan some follow-up firefighting...

As a side note, our lack of full monitoring across all servers is a big drawback at this point.

Matt-Yorkley:
I did some quick investigation on this.

I can't see any glaring specific things that are obvious causes of the small downtimes, I think it's just high usage. There were one or two actions that maybe used slightly more memory at the relevant times (I won't mention them specifically) but they all appear to be totally normal usage.

One thing I did spot (near the top of the list) was some wildly inefficient queries going on in the `CheckoutController` that I'm pretty sure are totally unnecessary. We're querying mountains of product properties and producer properties data in the cart edit page, and I don't think we actually need it there at all (it's just a consequence of re-using fat serializers in places they shouldn't be re-used).

Also I think we it would be really valuable at this point to have a little crack at getting some quick wins via #4763

lin-d-hop:
Yes please!

Matt-Yorkley:
I rechecked the `CheckoutController` idea, and actually it seems like impact would be fairly minimal.

I'd still recommend we try #4673 though.

luisramos0:
the number of visits today was not as big as last week... I see the access log in matomo and I cant find anything special there...

Issue #4673, opened 2020-01-14
Remove N+1 queries detected by Bullet


luisramos0:
### What we should change and why (this is tech debt)
We should add eager loading to avoid N+queries and remove unused eager loading.
Bullet was added to CI in #4475 and the output of it's run on 2020-01-14 is stored here:
https://gist.github.com/luisramos0/f044fd8e0f381c0793a995f7d373732e

There are a lot of entries but also a lot of repetition. Repeated entries are probably the best candidates to be addressed first.

For example this entry appears 51 times in the build:
```
Unused Eager Loading detected
  Spree::Order => [:adjustments]
  Remove from your finder: :includes => [:adjustments]
```

This other one appears 90 times in the file!
```
GET http://127.0.0.1:43198/enterprise-52/shop
Need Counter Cache
  Enterprise => [:shipping_methods]
N+1 Query detected
  Enterprise => [:shipping_methods]
  Add to your finder: :includes => [:shipping_methods]
N+1 Query method call stack
  /home/runner/openfoodnetwork/app/models/enterprise.rb:341:in `ready_for_checkout?'

```

If these entries come up many times in this log, it's because it is a code path crossed many times in the specs, and that probably means it is something that is executed a lot of times in production.

### Context
This came up in #4475 as part of the performance improvement epics recently addressed.

### Impact and timeline
This is not the top priority but it represents a really good and rather easy way to improve performance of the app in general. Defenetely something we should address at some point, for example, the next time we want to work on the app's overall performance.

Matt-Yorkley:
If anyone picks this up again (we still have really basic N+1s all over the app) I found another nice option for usage. Add the bullet gem, then add this to `environments/development.rb`:
```ruby
config.after_initialize do
  # To use, start a rails server with: `ENABLE_BULLET=true rails s`
  if Rails.env.development? && ENV.fetch('ENABLE_BULLET', nil)
    Bullet.enable = true
    Bullet.rails_logger = true
    Bullet.alert = true # Display bullet results immediately in a Javascript alert
  end
end
```

RachL:
I'm closing this epic as there are no issues attached.

PR #5050, opened 2020-03-23
Performance tweaks 2

Matt-Yorkley:
#### What? Why?

Related to #5049 and #4673 

Removes various N+1 queries from a number of pages. Partly directed by `bullet` reports, and partly by SQL output in `rack-mini-profiler`. Tested locally with Aus production data.

Significantly improves load times on groups and producers pages, and in **admin order edit**, which was especially bad with full production data.

#### What should we test?

Sanity check:
- groups
- producers
- admin order edit

#### Release notes

Performance improvements for producers and groups pages, and admin order edit

Changelog Category: Changed

### DIFF
--- a/lib/open_food_network/enterprise_injection_data.rb
+++ b/lib/open_food_network/enterprise_injection_data.rb
@@ -1,7 +1,8 @@
 module OpenFoodNetwork
   class EnterpriseInjectionData
-    def active_distributors
-      @active_distributors ||= Enterprise.distributors_with_active_order_cycles.ready_for_checkout
+    def active_distributor_ids
+      @active_distributor_ids ||=
+        Enterprise.distributors_with_active_order_cycles.ready_for_checkout.pluck(:id)
     end
 
     def earliest_closing_times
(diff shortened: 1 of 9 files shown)


lib/open_food_network/enterprise_injection_data.rb
```diff
@@ -1,7 +1,8 @@
 module OpenFoodNetwork
   class EnterpriseInjectionData
     def active_distributors
-      @active_distributors ||= Enterprise.distributors_with_active_order_cycles.ready_for_checkout
+      @active_distributors ||=
```
luisramos0: this needs to be renamed to active_distributor_ids

luisramos0: or the pluck can be moved to the caller methods.

filipefurtad0:
Hi @Matt-Yorkley,

Tested basic funcionality of the pages:

- groups
- producers
- admin order edit

All good in terms of functionality/sanity.
Good to go.

PR #5052, opened 2020-03-23
1760-Instagram-Link-Validation

### LINKED ISSUE #1760: Instagram links not working

reuschelcm:
#### What? Why?

Closes #1760

This PR returns Instagram validation to the enterprise social tab. It also accounts for the new potential input conditions:

- @instagram_handle
- https://instagram.com/@instagram_handle

These are both common (and, in the case of the second, incorrect) ways Instagram users enter in their handles on social tabs. 

This new validation also ignores case for both the Instagram handle (as these are always lowercase), and for the link, as a single erroneous capitalized character will cause it to validate as incorrect. (ie. a mistype of the link) 

#### What should we test?
Within the enterprise_spec.rb test suite, I've reincorporated the old validation tests as well as a few more to account for the new conditions specified above. This should test multiple variations of correct formats (ie with or without the entire link, the use of '@', etc), as well as incorrect formats and invalid links. 

#### Release notes

Added Instagram handle validation for the enterprise social tab. This allows for less confusion when the user enters their Instagram handle, even in different ways. 

Changelog Category: Added 

#### Discourse thread

https://github.com/openfoodfoundation/openfoodnetwork/issues/1760
https://github.com/openfoodfoundation/openfoodnetwork/pull/3238

### DIFF
--- a/spec/models/enterprise_spec.rb
+++ b/spec/models/enterprise_spec.rb
@@ -147,7 +147,6 @@ describe Enterprise do
       end
 
       context "prevent an wrong instagram link pattern" do
-
         it "expects to be invalid the instagram attribute https://facebook.com/user" do
           e = build(:enterprise, instagram: 'https://facebook.com/user')
           expect(e).to_not be_valid
--- a/spec/models/enterprise_spec.rb
+++ b/spec/models/enterprise_spec.rb
@@ -165,7 +165,7 @@ describe Enterprise do
 
       context "accepted pattern" do
         it "validates empty instagram attribute" do
-          e = build(:enterprise)
+          e = build(:enterprise, instagram: '')
           expect(e).to be_valid
           expect(e.instagram).to eq ""
         end
--- a/spec/models/enterprise_spec.rb
+++ b/spec/models/enterprise_spec.rb
@@ -220,7 +220,6 @@ describe Enterprise do
           expect(e).to be_valid
         end
 
-
         it "renders the expected pattern" do
           e = build(:enterprise, instagram: 'instagram.com/user')
           expect(e.instagram).to eq('user')
--- a/spec/models/enterprise_spec.rb
+++ b/spec/models/enterprise_spec.rb
@@ -146,82 +146,93 @@ describe Enterprise do
         expect(enterprise.contact).to eq enterprise.owner
       end
 
-      context "prevent an wrong instagram link pattern" do
-        it "expects to be invalid the instagram attribute https://facebook.com/user" do
+      context "prevent a wrong instagram link pattern" do
+        it "invalidates the instagram attribute https://facebook.com/user" do
           e = build(:enterprise, instagram: 'https://facebook.com/user')
           expect(e).to_not be_valid
         end
 
-        it "expects to be invalid the instagram attribute tagram.com/user" do
+        it "invalidates the instagram attribute tagram.com/user" do
           e = build(:enterprise, instagram: 'tagram.com/user')
           expect(e).to_not be_valid
         end
 
-        it "expects to be invalid the instagram attribute https://instagram.com/user/preferences" do
+        it "invalidates the instagram attribute https://instagram.com/user/preferences" do
           e = build(:enterprise, instagram: 'https://instagram.com/user/preferences')
           expect(e).to_not be_valid
         end
       end
 
       context "accepted pattern" do
-        it "expects to be valid empty instagram attribute" do
+        it "validates empty instagram attribute" do
           e = build(:enterprise)
           expect(e).to be_valid
+          expect(e.instagram).to eq ""
         end
 
-        it "expects to be valid the instagram attribute @my-user" do
+        it "validates the instagram attribute @my-user" do
           e = build(:enterprise, instagram: '@my-user')
           expect(e).to be_valid
+          expect(e.instagram).to eq "my-user"
         end
 
-        it "expects be valid the instagram attribute user" do
+        it "validates the instagram attribute user" do
           e = build(:enterprise, instagram: 'user')
           expect(e).to be_valid
+          expect(e.instagram).to eq "user"
         end
 
-        it "expects be valid the instagram attribute my_www5.example" do
+        it "validates the instagram attribute my_www5.example" do
           e = build(:enterprise, instagram: 'my_www5.example')
           expect(e).to be_valid
+          expect(e.instagram).to eq "my_www5.example"
         end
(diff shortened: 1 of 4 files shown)


reuschelcm:
Yeah those code issues are on me. I'll go ahead and refactor them. 

Those other conflicts were an issue with versioning. I think I made those changes in between a PR. I'll checkout my branch again and fix them.

luisramos0:
Great, looking good :+1:

You need to pull latest master and rebase this PR.

You also need to keep the bundler version at v1, not v2.

luisramos0:
:+1:

spec/models/enterprise_spec.rb
```diff
@@ -145,6 +145,86 @@
       it "sets the enterprise contact to the owner by default" do
         expect(enterprise.contact).to eq enterprise.owner
       end
+
+      context "prevent an wrong instagram link pattern" do
```
mkllnk: Little typo:
```suggestion
      context "prevent a wrong instagram link pattern" do
```

spec/models/enterprise_spec.rb
```diff
@@ -145,6 +145,86 @@
...
+        it "expects to be invalid the instagram attribute https://facebook.com/user" do
```
mkllnk: It's better English to write it like this:
```suggestion
        it "invalidates the instagram attribute https://facebook.com/user" do
```

spec/models/enterprise_spec.rb
```diff
@@ -145,6 +145,86 @@
...
+
+      context "accepted pattern" do
+        it "expects to be valid empty instagram attribute" do
+          e = build(:enterprise)
+          expect(e).to be_valid
+        end
+
+        it "expects to be valid the instagram attribute @my-user" do
+          e = build(:enterprise, instagram: '@my-user')
+          expect(e).to be_valid
```
mkllnk: It would be nice to test the result here as well.
```suggestion
          expect(e).to be_valid
          expect(e.instagram).to eq "my-user"
```

All valid cases should be tested this way. Otherwise this method would pass the test as well:
```rb
  def instagram=(value)
    self[:instagram] = ""
  end
```

reuschelcm: Sorry for the slight absence. Currently working on a fix to the spec file and adding the suggested changes!

This is a very good point, and that should also help me figure out the issue with that enterprise test. Thanks!

sigmundpetersen:
Build is green ✔️ 
Is this ready for review again @reuschelcm ?

reuschelcm:
@sigmundpetersen Yup, it should be ready to be reviewed!

mkllnk:
@reuschelcm I just remembered an important issue here. We have lots of existing enterprises that are currently valid but won't be valid with the new validation. For example, one of our enterprises has the instagram field `https://www.instagram.com/dandaraganorganicbeef/`.

If we accept your change as is, all these enterprises will be invalid and can't be saved unless we change the link. We need a database migration to fix the old enterprises. Is that something you would be happy to add?

reuschelcm:
@mkllnk That's right, that's right, that had slipped my mind. I'll go ahead and see if I can add  migration! There is an old db migration script from 2018 that I might be able to template from. I'll take a look at it and see how to integrate it.

mkllnk:
Great, just make sure to not use the Enterprise model. You can define your own model in the migration if you need to but referencing code outside of migrations is very brittle and should be avoided.

RachL:
hello @reuschelcm how are doing? Are you still going to work on this migration or should we close this PR for someone else to pick it up?

daniellemoorhead:
Will close this PR and make the issue available for someone else. @reuschelcm feel free to open and finish if you decide to come back 🙂

PR #5060, opened 2020-03-24
Fix weight calculator

### LINKED ISSUE #4695: Checkout broken if weight calculator used in a payment or shipping method AND a product in the cart has unit_value ZERO

Matt-Yorkley:
#### What? Why?

Closes #4695

Fixes issue where weight calculator was causing fatal errors by returning `BigDecimal::NaN` in some cases.

In the case where `variant.product.unit_type` was `items`, and `variant.unit_value` was set to `0`, and `variant.weight` was not set; we were **dividing by zero** in one of the calculations.

#### What should we test?

Checkout should work when: 
- one of the products in the order uses "items" as the type and has unit_value "0".
- the shipping method uses the weight calculator

#### Release notes

Fixed problematic edge case for shipping methods using weight calculator

Changelog Category: Fixed

### DIFF
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -54,6 +54,8 @@ module Calculator
     #    Customer ends up getting 350mL (line_item.final_weight_volume) of wine
     #      that represent 2.8 (quantity_implied_in_final_weight_volume) glasses of wine
     def quantity_implied_in_final_weight_volume(line_item)
+      return line_item.quantity if line_item.variant.unit_value.zero?
+
       (1.0 * line_item.final_weight_volume / line_item.variant.unit_value).round(3)
     end
 
--- a/app/models/calculator/weight.rb
+++ b/app/models/calculator/weight.rb
@@ -54,7 +54,7 @@ module Calculator
     #    Customer ends up getting 350mL (line_item.final_weight_volume) of wine
     #      that represent 2.8 (quantity_implied_in_final_weight_volume) glasses of wine
     def quantity_implied_in_final_weight_volume(line_item)
-      return line_item.quantity if line_item.variant.unit_value.zero?
+      return line_item.quantity if line_item.variant.unit_value.to_f.zero?
 
       (1.0 * line_item.final_weight_volume / line_item.variant.unit_value).round(3)
     end
--- a/spec/models/calculator/weight_spec.rb
+++ b/spec/models/calculator/weight_spec.rb
@@ -156,19 +156,24 @@ describe Calculator::Weight do
     let(:product) {
       create(:product, variant_unit: 'items', variant_unit_scale: nil, variant_unit_name: "bunch")
     }
-    let(:variant1) { create(:variant, product: product, unit_value: 0, weight: nil) }
-    let(:variant2) { create(:variant, product: product, unit_value: 0, weight: 10.0) }
-    let(:line_item1) { create(:line_item, variant: variant1, quantity: 1) }
-    let(:line_item2) { create(:line_item, variant: variant2, quantity: 1) }
+    let(:line_item) { create(:line_item, variant: variant, quantity: 1) }
 
     before { subject.set_preference(:per_kg, 5) }
 
-    it "uses zero weight if variant.weight is not present" do
-      expect(subject.compute(line_item1)).to eq 0
+    context "when variant.weight is present" do
+      let(:variant) { create(:variant, product: product, unit_value: 0, weight: 10.0) }
+
+      it "uses the variant weight" do
+        expect(subject.compute(line_item)).to eq 50.0
+      end
     end
 
-    it "uses the variant weight if variant.weight is present" do
-      expect(subject.compute(line_item2)).to eq 50.0
+    context "when variant.weight is nil" do
+      let(:variant) { create(:variant, product: product, unit_value: 0, weight: nil) }
+
+      it "uses zero weight" do
+        expect(subject.compute(line_item)).to eq 0
+      end
     end
   end
 end
--- a/spec/models/calculator/weight_spec.rb
+++ b/spec/models/calculator/weight_spec.rb
@@ -141,7 +141,7 @@ describe Calculator::Weight do
     end
 
     context "when the product uses item unit" do
-      let!(:product_attributes) { { variant_unit: "items", variant_unit_scale: nil, variant_unit: "pc", display_as: "pc" } }
+      let!(:product_attributes) { { variant_unit: "items", variant_unit_scale: nil, variant_unit_name: "pc", display_as: "pc" } }
       let!(:variant_attributes) { { unit_value: 3.0, weight: 2.5, display_as: "pc" } }
 
       it "is correct" do
--- a/spec/models/calculator/weight_spec.rb
+++ b/spec/models/calculator/weight_spec.rb
@@ -163,19 +163,11 @@ describe Calculator::Weight do
 
     before { subject.set_preference(:per_kg, 5) }
 
-    it "returns NaN if variant.weight is not present" do
-      expect(subject.compute(line_item1).nan?).to be_truthy
-    end
-
-    xit "uses zero weight if variant.weight is not present" do
+    it "uses zero weight if variant.weight is not present" do
       expect(subject.compute(line_item1)).to eq 0
     end
 
-    it "returns NaN if variant.weight is present" do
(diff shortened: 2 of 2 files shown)


app/models/calculator/weight.rb
```diff
@@ -54,6 +54,8 @@ def weight_per_final_weight_volume(line_item)
     #    Customer ends up getting 350mL (line_item.final_weight_volume) of wine
     #      that represent 2.8 (quantity_implied_in_final_weight_volume) glasses of wine
     def quantity_implied_in_final_weight_volume(line_item)
+      return 0 if line_item.variant.unit_value.zero? # Don't divide by zero ;)
```
luisramos0: you could default to 1 here, that would mean the variant.weight field would be used. this way the weight will be zero even if the weight field on the variant is defined.

luisramos0: no, better, I think we should default to line_item.quantity.
it would read "if unit_value is zero, the quantity_implied_in_final_weight_volume cannot be calculated so we use the line_item.quantity

Matt-Yorkley: We're checking for the presence of `final_weight_volume`, but it always appears to be present and defaults to zero. I don't think we should use the volume calculations at all when `variant_unit` is "items"...?

Matt-Yorkley: It still returns zero anyway, because the calculator sets `:preferred_per_kg` to `0.0`, and we multiply the result by that... :see_no_evil:

app/models/calculator/weight.rb
```diff
@@ -54,11 +54,14 @@ def weight_per_final_weight_volume(line_item)
     #    Customer ends up getting 350mL (line_item.final_weight_volume) of wine
     #      that represent 2.8 (quantity_implied_in_final_weight_volume) glasses of wine
     def quantity_implied_in_final_weight_volume(line_item)
+      return weight_per_variant(line_item) if line_item.variant.unit_value.zero?
+
       (1.0 * line_item.final_weight_volume / line_item.variant.unit_value).round(3)
     end
 
     def final_weight_volume_present?(line_item)
```
Matt-Yorkley: This check was always returning `true`; the default value is a `BigDecimal` object with value `0.0`...

Matt-Yorkley:
@luisramos0 I added an extra test, changed some key bits of the logic and extracted the "never divide by zero" bit into a separate commit. Needs a careful re-review...

app/models/calculator/weight.rb
luisramos0: final_weight_volume is not always present, for example for subscriptions. the check fixed a bug if you look into the commit history.

it's a good question: I think you can change the final weight volume of line items even if the variant_unit is items... in these cases final_weight_volume  becomes variant.unit_value time line_item quantity.
So, 
product.variant_unit items
variant.display_name "5 roses"
variant.unit_value 5
variant.weight 20g
line_item.quantity 2
line_item.final_weight_volume 10 (variant.unit_value * line_item.quantity)

for this line item weight calculator should calculate 40grams

but if user changes line_item.final_weight_volume from 10 to 8 (8 roses were actually delivered) the weight calculator will compute correctly 32grams :+1:  20 * (8/5)

So, I think it's good as is.

luisramos0: the field name should not be final_weight_volume
as I demonstrate above it's actually final_weight_volume_units, in this case final_units.

app/models/calculator/weight.rb
luisramos0: I added it to fix a bug
e9e6aa77d88ebc6d93277d85bbc2e4eca1f558d2

app/models/calculator/weight.rb
```diff
@@ -57,8 +57,13 @@ def quantity_implied_in_final_weight_volume(line_item)
...
     end
 
+    def use_final_weight_volume?(line_item)
+      final_weight_volume_present?(line_item) && !line_item.variant.unit_value.zero?
+    end
+
     def final_weight_volume_present?(line_item)
-      line_item.respond_to?(:final_weight_volume) && line_item.final_weight_volume.present?
+      line_item.respond_to?(:final_weight_volume) && line_item.final_weight_volume.present? &&
+        !line_item.final_weight_volume.zero?
```
luisramos0: looking at the DB I suspect a zero final_weight_volume is actually meaningful (set by user) and we should use it... I'd not change that behavior.

app/models/calculator/weight.rb
```diff
@@ -57,8 +57,13 @@ def quantity_implied_in_final_weight_volume(line_item)
       (1.0 * line_item.final_weight_volume / line_item.variant.unit_value).round(3)
     end
 
+    def use_final_weight_volume?(line_item)
+      final_weight_volume_present?(line_item) && !line_item.variant.unit_value.zero?
```
luisramos0: I dont think this is correct. If variant_unit is weight, you dont need unit_value to use final_weight_volume.
I think the previous version of the code is better where we just change quantity_implied_in_final_weight_volume

Matt-Yorkley:
Version 3 is ready for review. I played a lot with additional test cases and it was a minefield of `NaN` errors...

Matt-Yorkley:
Also I noticed the other tests in this spec file all use `:build` in the line item factories instead of `:create`, which skips the callback that calculates and saves the `final_weight_volume`...

Matt-Yorkley:
Oops... there was an extra `end` in one of your commit lines that broke the build. It's fixed now :+1:

luisramos0:
oh, I am sorry! I must have messed it up while splitting the code in two commits...

app/models/calculator/weight.rb
```diff
@@ -54,6 +54,8 @@ def weight_per_final_weight_volume(line_item)
     #    Customer ends up getting 350mL (line_item.final_weight_volume) of wine
     #      that represent 2.8 (quantity_implied_in_final_weight_volume) glasses of wine
     def quantity_implied_in_final_weight_volume(line_item)
+      return line_item.quantity if line_item.variant.unit_value.zero?
```
mkllnk: I looked at au-prod:

```rb
Spree::Variant.where(unit_value: 0).count
# => 38

Spree::Variant.where(unit_value: nil).count
# => 630
```

I think we can catch these cases as well:

```suggestion
      return line_item.quantity if line_item.variant.unit_value.to_i.zero?
```

What do you think? Is that case covered somewhere else?

mkllnk: I'm not sure if this is a valid case but we do have data like this:
```rb
item = Spree::LineItem.where("final_weight_volume > 0").joins(:variant).where("spree_variants.unit_value is null").last
Calculator::Weight.new.compute(item)
# TypeError: nil can't be coerced into BigDecimal
```

filipefurtad0:
Hi @Matt-Yorkley,

I have been trying to reproduce the bug - but it does not seem to be silent anymore. So, now one gets a warning "The checkout failed. Please let us know so that we can process your order." 

In any case, after following the test cases - all seems to work fine. Checkout is now possible, and the delivery costs are set by default to zero, on the delivery "items" as units. Also, I checked that everything works normally, and that the calculator works normally for weight items.

Good to go.

Matt-Yorkley:
Thanks @filipefurtad0! Awesome testing :muscle:

PR #5064, opened 2020-03-24
Fix sample data and custom paper_trail config on order_cycles and schedules on rails 4

luisramos0:
#### What? Why?

Fixes sample data and also adapts paper_trail custom data config.

#### What should we test?
Build green.
Sample data still works in master.
Table versions still has custom_data field with list of ids when creating and editing schedules and order cycles.

#### Release notes
Changelog Category: Changed
Adapt sample data code to rails 4.
Adapt logging in subscriptions related code (schedules) to work with rails 4.


luisramos0:
I have tested this in es staging, looks good.

PR #5065, opened 2020-03-24
[Spree 2.1] Upgrade some dependencies... rails 4.0.13 :-)

### LINKED ISSUE #5011: [Spree 2.1] Improve Gemfile
A couple of details need to be checked in the Gemfile in 3-0-stable as wisely suggested by @sauloperez 
This issue represents the work to go through the comments and clarify each point.

This task is probably better done when the build is green!

My main thoughts here are:
- we should probably go for rails 4.0.11 directly
- I think deface can be removed

### Context
From
https://github.com/openfoodfoundation/openfoodnetwork/pull/4930

### Impact and timeline
TODO when the build is green.

luisramos0:
#### What? Why?

Closes #5011 thanks @sauloperez for raising all the points!

Here we upgrade rails to 4.0.13, factory_bot_rails to 4.10.0,  devise-encryptable to 0.2.0 and we remove the direct dependency to deface (we dont use it anymore!).

#### What should we test?
The latest build is working fine with this PR, only flaky specs.


Gemfile
```diff
@@ -74,7 +74,6 @@ gem 'angularjs-file-upload-rails', '~> 2.4.1'
 gem 'blockenspiel'
 gem 'custom_error_message', github: 'jeremydurham/custom-err-msg'
 gem 'dalli'
-gem 'deface'
```
sauloperez: I kind of forgot about it (that's what our brain does to get over traumas, right? :trollface: ) but it was such a pain...

Gemfile
```diff
@@ -131,7 +130,7 @@ group :test, :development do
   gem 'awesome_print'
   gem 'capybara', '>= 2.18.0' # 3.0 requires nokogiri 1.8
   gem 'database_cleaner', require: false
-  gem "factory_bot_rails", '4.8.2', require: false
+  gem "factory_bot_rails", '4.10.0', require: false
```
sauloperez: is there a particular reason why we can use only 4.10.0 and not any patches to come? Just wondering if we can set the version a bit looser so that dependabot fins no trouble upgrading it for us.

luisramos0: I just followed your advise:
https://github.com/openfoodfoundation/openfoodnetwork/pull/4930#discussion_r392170518
we will be fine until rails 4.2 anyway.

sauloperez: My point is that in this case, we can do `~> 4.10` so running a `bundle update factory_rails_bot` will pick up any patches on the 4.10 series.

Issue #5003, opened 2020-03-17
Failed Orders / Payments for Prom Coast Food Collective


emilyjeanrogers:
## Description

Customers of Prom Coast Food Collective in Aus have reported being unable to place orders. Possibly some / all are seeing an error "Check Payment Pathway". 

Investigation of Orders for this food hub shows approx 50 successful (completed / paid) orders and approx 180 unsuccessful or incomplete (cart / balance due) orders. Most of the incomplete orders are empty carts with $0 owing. 

## Expected Behaviour

Payment Successful for all orders

## Actual Behaviour

What we know so far is that some customers are unable to check out and are getting an error message that says 'check payment pathway'. They are using Stripe as the only available payment method. They have used Stripe consistently for ages. The Order Cycle opened today (17th March 2020) and as at 2.45pm there have been 60 successful Stripe transactions.

Waiting for more information from customer about the unsuccessful transactions. 

## Steps to Reproduce

For some customers, they are trying to check out and getting this error. Not all customers. We don't know why yet

Could it be new Stripe-SCA? We have not yet switched them, should we try doing it?

## Severity

bug-s1: a critical feature is broken: checkout, payments, signup, login

mkllnk:
I can't find anything related in Bugsnag and I can't find that error message anywhere. Maybe it's something we should ask Stripe about. It's just difficult to replicate because people won't give us their credit card details for testing. :trollface:

kirstenalarsen:
![image](https://user-images.githubusercontent.com/2911176/76839552-ce597980-6889-11ea-8fbb-8e69185e7e8c.png)

kirstenalarsen:
![image](https://user-images.githubusercontent.com/2911176/76839728-20020400-688a-11ea-9ff7-2f8e58814526.png)

![image](https://user-images.githubusercontent.com/2911176/76839777-390ab500-688a-11ea-985a-c71acc5a9414.png)

luisramos0:
A failed payment will not empty the cart.
If a user loads the shop page, an empty cart will be created.
Also, immediately after checkout, a new empty cart will be created for the user. So, those 180 empty carts are not relevant for the analysis.


They have monthly OCs and are now running an exception second OC in March, these are the number of orders for each OC:
JANUARY OC:
  state   | count 
----------+-------
 canceled |     1
 cart     |  1194             - of which 111 are not empty (~9.3%)
 complete |   105

FEBRUARY OC:
  state   | count 
[...]

PR #5070, opened 2020-03-25
Translate credit card brand so that active merchand code handles the payment correctly

### LINKED ISSUE #5068: Failures at Checkout with StripeSCA and mastercards
## Description

After switching to Stripe SCA users a reporting a small but significant number of failed payments.

The errors are caught and reported in Bugsnag:
https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5e79c71a41fa51001712e4f6?filters[event.since][0]=30d&filters[error.status][0]=new

For active Stripe users I have reverted their payment method from Stripe SCA to the old Stripe payment method.

## Expected Behavior

Successful Payments

## Actual Behaviour

Error on payment

## Steps to Reproduce

1. Currently unsure of the cause but I think Luis has some ideas
2.
3.
4.

## Animated Gif/Screenshot

image1

## Workaround

Revert to Stripe payment method

## Severity

## Your Environment

* Version used: 2.8 UK
* Browser name and version: Multiple
* Operating System and version (desktop or mobile): Multiple

luisramos0:
#### What? Why?

I _believe_ it closes #5068

We convert card types so that active merchant handles the payment correctly. For some reason I didnt copy this logic from the old stripe payment method. We need this.
This is done in checkout and backend payments collection.

Additionally, I added a simple console.log statement to LOG errors on JS calls to Stripe. We will have the console log in the browser available to see what went wrong. This was discussed here: https://github.com/openfoodfoundation/openfoodnetwork/issues/5003#issuecomment-600246636
This is how it will look like in the user browser console:
image

#### What should we test?
Verify stripeSCA payments are working as before.
We can test the logging by checking out with a test stripe card but inserting a invalid expiry date. It shows a red message but you can click checkout anyway and check the console log.
#### Release notes
Changelog Category: Fixed
Fix a problem in stripeSCA for some card types like mastercard.

### DIFF
--- a/app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
+++ b/app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
@@ -14,6 +14,7 @@ angular.module("admin.payments").factory 'AdminStripeElements', ($rootScope, Sta
       @stripe.createToken(@card, cardData).then (response) =>
         if(response.error)
           StatusMessage.display 'error', response.error.message
+          console.log(JSON.stringify(response.error))
         else
           secrets.token = response.token.id
           secrets.cc_type = @mapCC(response.token.card.brand)
@@ -29,9 +30,10 @@ angular.module("admin.payments").factory 'AdminStripeElements', ($rootScope, Sta
       @stripe.createPaymentMethod({ type: 'card', card: @card }, @card, cardData).then (response) =>
         if(response.error)
           StatusMessage.display 'error', response.error.message
+          console.log(JSON.stringify(response.error))
         else
           secrets.token = response.paymentMethod.id
-          secrets.cc_type = response.paymentMethod.card.brand
+          secrets.cc_type = @mapCC(response.paymentMethod.card.brand)
           secrets.card = response.paymentMethod.card
           submit()
 
--- a/app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
+++ b/app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
@@ -14,10 +14,10 @@ angular.module("admin.payments").factory 'AdminStripeElements', ($rootScope, Sta
       @stripe.createToken(@card, cardData).then (response) =>
         if(response.error)
           StatusMessage.display 'error', response.error.message
-          console.log(JSON.stringify(response.error))
+          console.error(JSON.stringify(response.error))
         else
           secrets.token = response.token.id
-          secrets.cc_type = @mapCC(response.token.card.brand)
+          secrets.cc_type = @mapTokenApiCardBrand(response.token.card.brand)
           secrets.card = response.token.card
           submit()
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
```diff
@@ -29,9 +30,10 @@ angular.module("admin.payments").factory 'AdminStripeElements', ($rootScope, Sta
       @stripe.createPaymentMethod({ type: 'card', card: @card }, @card, cardData).then (response) =>
         if(response.error)
           StatusMessage.display 'error', response.error.message
+          console.log(JSON.stringify(response.error))
```
sauloperez: You might as well use [console.error](https://developer.mozilla.org/en-US/docs/Web/API/Console/error) so it'll show the red signal on the dev tools and show the message in red. We won't miss it this way.

filipefurtad0:
Hi @luisramos0,

Tested the console logging as you suggested, it works fine! This is great.

In what concerns the #5068 issue: I first reproduced it, and before staging this PR observed that payment with the mastercard 5555555555554444 shows the error the error reported in #5068:

- "The checkout failed. Please let us know so that we can process your order."

and the console shows the error:

- "uncaught exception: Object"

After staging this PR, both the regular Visa and Mastercard now yielded the reported error:

4242424242424242 | Visa | Any 3 digits | Any future date
5555555555554444 | Mastercard | Any 3 digits | Any future date

- "The checkout failed. Please let us know so that we can process your order."

and the console shows the same error:

- "uncaught exception: Object"

image

This now happens both for regular Stripe and Stripe SCA.

Also, this PR appears to break SCA payments. Using SCA payment method and SCA Card: 4000002500003155 results in Snail, console pic below:

image

Moving back to dev.

luisramos0:
I think you just experienced some issue in the stg FR. What stripe config is there?
In your comment above you dont seem to have tested regular visa before the PR. I think it would break as well.

This issue 5068 is not replicable with 5555555555554444, payments with this card should work in both stripe and stripeSCA. Before and after this PR.

We need to test this again.

filipefurtad0:
Thank you @luisramos0 ,
An issue with the staging server settings would indeed explain the findings.
I'm moving to test ready for further testing.

luisramos0:
I tested this further, this PR is not correct. I'll add a new commit in a few minutes.

app/assets/javascripts/admin/payments/services/stripe_elements.js.coffee
luisramos0: I change it now. we will get a bugsnag alert :-)

filipefurtad0:
Hi @luisramos0 ,

I am having a look in AU-staging, before staging this PR. Some observations:

a) In this shop (Tapaskoop), both SCA and Stripe connect are set, and SCA-authentication is working fine, payments are successful with 3D cards, also with the simple 4242424242424242 Visa-card. So, I think SCA-settings are fine.

b) With Stripe-Connect, both 4242424242424242 and 5555555555554444 work fine too. However, while attempting to use the mastercard 5555555555554444 with the Stripe-SCA payment method I observe the error message in #5068, as before on the FR-staging.

c) after attempting to checkout with Stripe-SCA and the 5555555555554444 and getting the error "The checkout failed. Please let us know so that we can process your order." other payment methods seem to get "blocked" (i.e. fail), on that checkout session.

d) after seeing this error, all other payment methods fail - even simple cash payment - or the 4242424242424242 or 5555555555554444 cards, which normally work on Stripe-Connect (as in b)).

e) The workaround is to use Stripe-SCA with 4242424242424242 or a 3D-secure card, which works.

f) Another workaround is to go back to shop and back to cart, avoiding to use the combination 5555555555554444+StripeSCA. Doing this, allows one to use some other method/card, like, a) or b) or cash.

Summarizing, 5555555555554444+StripeSCA does not work and kind of "blocks" the checkout for other payment methods, so I can't verify your previous statement: "with 5555555555554444, payments with this card should work in both stripe and stripeSCA. Before and after this PR."

I'm staging the PR, and see what happens :-)

luisramos0:
nice testing!!!
it looks like you are having problems with mastercard after all, the perfect scenario would be that this PR fixes that problem :-)
I think that the problem with mastercard on stripeSCA may leave the order in a broken state and that explains why subsequent attempts also fail.

filipefurtad0:
Awsome @luisramos0 !

- the cards 4242424242424242 and 5555555555554444 work both on Stripe-SCA and Stripe-Connect
- The console shows logs, like errors with the validity of the card.
- Secure3D cards work as expected on Stripe-SCA.

I think it closes #5068 :tada: and I guess what I observed in the FR-Staging was the "blocking" of the order. 

Moving to ready to go! :v:

luisramos0:
nice, I am not sure about the failure of the mastercard before this PR (maybe a replication of 5068, maybe not) but anyway, it's great it's working and we should move forward because this change is needed anyway.

But I made a process mistake here because I have changed the code and didnt wait for a re-review. I pinged devs on slack but I forgot to ask you (Filipe) to wait for the review :-(
Anyway, before we merge we need a review from another dev...

app/assets/javascripts/darkswarm/services/stripe_elements.js.coffee
```diff
@@ -34,26 +35,35 @@ Darkswarm.factory 'StripeElements', ($rootScope, Loading, RailsFlashLoader) ->
...
+      switch cardBrand
         when 'MasterCard' then return 'master'
         when 'Visa' then return 'visa'
         when 'American Express' then return 'american_express'
         when 'Discover' then return 'discover'
         when 'JCB' then return 'jcb'
         when 'Diners Club' then return 'diners_club'
 
+    # Maps the brand returned by Stripe's paymentMethodsAPI to that required by activemerchant
+    mapPaymentMethodsApiCardBrand: (cardBrand) ->
```
sauloperez, line 60: I understand this PR pretty needed that is why I approve, but we'll need to remove this duplication right after.

luisramos0: ah, I think we have discussed this before. we dont have a way to share code between darkswarm and admin!

Issue #5075, opened 2020-03-25
Silent Checkout Failures


lin-d-hop:
## Description

We are experiencing silent checkout failures that we cannot explain.
This issue is a place to discuss and explore... as we don't have a huge amount of information yet.

We know that many card errors from Stripe are not creating error messages visible to the users.

We also suspect that there might be a rare race condition when someone checks out with a sold out item in the cart. This is unconfirmed so probably outside the scope of this specific issue, but something to bear in mind when reporting such things.

## Expected Behaviour

If checkout fails the user should be notified and this should be recorded in Bugsnag

## Actual Behaviour

There are times when the checkout fails silently.

## Steps to Reproduce

This section was added by @luisramos0 
One example we can reproduce (although not exactly a bug because the user is shown an error message next to the card number).
1. Go to checkout
2. Pay with Stripe
3. Insert an invalid card number or a past expiry date
4. Verify there is an error message displayed below the input field
![image](https://user-images.githubusercontent.com/1640378/77553700-0c1f5900-6ead-11ea-8901-286c20c29d2a.png)

5. Press "Place order now"
6. Watch the spinning "Processing payment..." for 10secs
7. No flash error shown on top of the page

There is a way to see the error message in this case:
Same as above but on step 6, instead of watching, click anywhere on the page while the "processing payment..." is there and see the flash error being displayed.

## Severity

bug-s2: a non-critical feature is broken, no workaround

## Your Environment

* Version used: Reported in UK, FR, 2.8
* Browser name and version:
* Operating System and version (desktop or mobile):

Matt-Yorkley:
Related PRs in progress: #5076, #5060
Related issues / discussions: #5003

lin-d-hop:
Another report of a silent failed checkout with a card user in the UK.

lin-d-hop:
![](https://github.trello.services/images/mini-trello-icon.png) [Order from Cultivate?](https://trello.com/c/cqVrqFQB/767-order-from-cultivate)

lin-d-hop:
![](https://github.trello.services/images/mini-trello-icon.png) [Order from Cultivate?](https://trello.com/c/cqVrqFQB/767-order-from-cultivate)

lin-d-hop:
@luisramos0 You are spot on with the steps to reproduce. From a user:

[...]

PR #5076, opened 2020-03-25
Remove Rails Foundation Helper and improve error display and logging in checkout controller

### LINKED ISSUE #5046: [Spree 2.1] Re-add or remove FoundationRailsHelper
The gemfile is pointing to a rails 3 fork on willrjmarshall's repo...
`gem 'foundation_rails_helper', github: 'willrjmarshall/foundation_rails_helper', branch: 'rails3'`

Also, in app/helpers/application_helper.rb FoundationRailsHelper::FlashHelper is commented.

Is this dependency needed where? Are the flash messages working in the admin section? They must be because the specs test them, do they look good?

### Context
Extracted from #4777

### Impact and timeline
This must be resolved now.

luisramos0:
#### What? Why?
Closes #5046 and #5075

There are 5 changes in this PR:
**5046** - we remove rails_foundation_helper (removed code can be seen [here](https://github.com/willrjmarshall/foundation_rails_helper/tree/master/lib/foundation_rails_helper)):
- flash_helper - I have tested a few flash messages and they are all showing up as before
- form_helper (f_form_for) - I cant find any differences in the checkout form with and without this

**5075** - errors on calls to stripe now show as flash messages to the user (like an invalid card number not detected by the credit card form)

**Additionally but related to 5075**, in this PR I am improving the checkout controller so that if there is an error in some part of the server side checkout process, the flash message to the user will always show the errors present on the order object.
**And additionally but also related to 5075**, in this PR I am generating a bugsnag error every time the checkout process fails. I think this will be very useful to debug of the mysteries of failed checkouts. Because it's such a critical part of the app I think we should have these alerts. For an example of a case, see #5075

Improve **color of field specific error messages** in checkout page.
Before:
https://files.slack.com/files-pri/T02G54U79-F010MTLHMFD/image.png
image
After:
image

#### What should we test?
rails_foundation_helper:
- verify that error flash messages on top of the page show up in checkout and in the admin forms (for example the warning when enterprises dont have payment methods defined)
- make sure the checkout form looks visually equal to what it was before

Errors in stripe calls: see 5075 for a simple way to simulate this.

Improving checkout controller: this change is a safe guard, I am not sure when this will happen or how to test it but imo it makes sense to add this code.

Verify the field specific error messages in the checkout page have white text now and are easier to read.

#### Release notes
Changelog Category: Fixed
Fixed silent checkout when calls to stripe were failing. User will now see an error message.
Removed rails_foundation_helper, an old dependency no longer needed.
Improved checkout controller to provide user with an error message in more error scenarios.

### DIFF
--- a/app/assets/javascripts/darkswarm/services/stripe_elements.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/stripe_elements.js.coffee
@@ -15,6 +15,7 @@ Darkswarm.factory 'StripeElements', ($rootScope, Loading, RailsFlashLoader) ->
         if(response.error)
           Loading.clear()
           RailsFlashLoader.loadFlash({error: t("error") + ": #{response.error.message}"})
+          @triggerAngularDigest()
         else
           secrets.token = response.token.id
           secrets.cc_type = @mapCC(response.token.card.brand)
@@ -32,11 +33,16 @@ Darkswarm.factory 'StripeElements', ($rootScope, Loading, RailsFlashLoader) ->
         if(response.error)
           Loading.clear()
           RailsFlashLoader.loadFlash({error: t("error") + ": #{response.error.message}"})
+          @triggerAngularDigest()
         else
           secrets.token = response.paymentMethod.id
           secrets.cc_type = response.paymentMethod.card.brand
           secrets.card = response.paymentMethod.card
           submit()
+
+    triggerAngularDigest: ->
+      # $evalAsync is improved way of triggering a digest without calling $apply
+      $rootScope.$evalAsync()
 
     # Maps the brand returned by Stripe to that required by activemerchant
     mapCC: (ccType) ->
--- a/Gemfile
+++ b/Gemfile
@@ -112,7 +112,6 @@ gem 'momentjs-rails'
 gem 'turbo-sprockets-rails3'
 
 gem "foundation-rails"
-gem 'foundation_rails_helper', github: 'willrjmarshall/foundation_rails_helper', branch: "rails3"
 
 gem 'jquery-migrate-rails'
 gem 'jquery-rails', '3.1.5'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -65,16 +65,6 @@ GIT
       rails-i18n
       spree_core (>= 1.1)
 
-GIT
-  remote: https://github.com/willrjmarshall/foundation_rails_helper.git
-  revision: 4d5d53fdc4b1fb71e66524d298c5c635de82cfbb
-  branch: rails3
-  specs:
-    foundation_rails_helper (0.4)
-      actionpack (>= 3.0)
-      activemodel (>= 3.0)
-      railties (>= 3.0)
-
 PATH
   remote: engines/catalog
   specs:
@@ -740,7 +730,6 @@ DEPENDENCIES
   foreigner
   foundation-icons-sass-rails
   foundation-rails
-  foundation_rails_helper!
   fuubar (~> 2.5.0)
   geocoder
   gmaps4rails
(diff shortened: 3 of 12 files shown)


Matt-Yorkley:
15 build errors, only one is flaky. Moving back to `in dev` for now.

app/assets/javascripts/darkswarm/services/stripe_elements.js.coffee
```diff
@@ -15,6 +15,7 @@ Darkswarm.factory 'StripeElements', ($rootScope, Loading, RailsFlashLoader) ->
         if(response.error)
           Loading.clear()
           RailsFlashLoader.loadFlash({error: t("error") + ": #{response.error.message}"})
+          $rootScope.$apply()
```
Matt-Yorkley, line 18: Was this added because it wasn't working in the browser, or because it wasn't working in the javascript spec?

luisramos0: in the browser! this is the main thing in this PR, it fixes #5075

Matt-Yorkley: Calling `$apply()` manually generally makes Angular a lot more brittle. Angular wraps almost everything in an `$apply()` call by default, and if `$apply()` ever gets called inside another `$apply()` call for any reason, it throws a fatal error. So I'm a bit nervous about putting it in critical checkout methods...

If it's absolutely needed, there are some safer options that will achieve the same thing, but won't ever throw fatal errors: wrap it in a `$timeout()` call (with zero "time"), or I think `$scope.$evalAsync()` does something similar but possibly in a more efficient way...

More info here: https://www.bennadel.com/blog/2605-scope-evalasync-vs-timeout-in-angularjs.htm

Matt-Yorkley: I wish we could get rid of Angular... :see_no_evil:

luisramos0: nice one Matt.
both apply with timeout and evalAsync fix the problem as well.
I thought evalAsync is nice (if we believe what the article explains) but it is not that known to other devs as $apply(), so I decided to use it but extract a method to explain why it is there.
Thanks!

luisramos0: here
https://github.com/openfoodfoundation/openfoodnetwork/pull/5076/commits/957b398a549a569d9946ba6fff720dfc057d7664

Matt-Yorkley: Ah, and it works? I thought the code to be run had to be passed in as a callback? Like: 
```
$rootScope.$evalAsync( ->
   Loading.clear()
   RailsFlashLoader.loadFlash({error: t("error") + ": #{response.error.message}"})
)
```
:man_shrugging:

Matt-Yorkley: As long as we're not directly calling `$apply` here, I'm happy :+1: :+1:

filipefurtad0:
Olá @luisramos0,

This looks good, great improvements. Reproduced the silent error on checkout, with an outdated card, as described in #5075. 

Staged the PR and tested/verified the following:
- error flash messages on top of the page show up in checkout and in the admin forms (for example the warning when enterprises dont have payment methods defined)
- the checkout form looks visually equal to what it was before
- Errors in stripe calls (as in #5075) - a message is now shown to the user, the spinner does not appear.
- verified the field specific error messages in the checkout page have white text - much easier to read now.

Thank you for the great PR description and test notes! Did not verify that a bugsnag error appears, I wonder how I would do that. I'm moving to ready to go, as this was not a test case.

luisramos0:
awesome, thanks Filipe.
we dont need to test bugsnag :+1:

PR #5087, opened 2020-03-27
Fix bug in subscriptions logging by making versions.custom_data a longer field

### LINKED ISSUE #5086: Snail when adding schedule to order cycle (or creating a new schedule) when schedule is in more than around 40 OCs
## Description
Snail when adding schedule to order cycle (or creating a new schedule) when schedule is in more than around 40 OCs.

The logging system blows up when too many Ordercycles are associated with a single schedule.
Bugsnag:
https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5e7dd4c615afa70017451d6c?event_id=5e7dd4c60059500d25370000&i=sk&m=nw

## Expected Behavior
User can add as many OCs to a schedule as they want.

## Actual Behaviour
Snail after a certain number of OCs.

## Steps to Reproduce
Add a schedule to 40 Order Cycles.

## Workaround
I think the user found a workaround because the latest OCs were associated with the schedule in the live DB. Maybve removing past OCs from the schedule).

## Severity
probably S3 because it looks they have found a workaround.
I am going to fix it now because the logging is a new thing and the fix is super easy.

## Your Environment
live UK

luisramos0:
#### What? Why?

Closes #5086

Make custom_data longer so a long list of schedules and order_cycles can be stored in the versions table.

Even if we cant replicate #5086 this is a valid change we should put in. That's why I prepared this PR immediately because it will avoid any other occurrences of the bugsnag error:
https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5e7dd4c615afa70017451d6c?event_id=5e7dd4c60059500d25370000&i=sk&m=nw

#### What should we test?
We need to add a schedule to many order cycles (40 was the number that broken it in uk live) and make sure the UI still works without any snail.

#### Release notes
Changelog Category: Fixed
Improved subscription logging system by making it more resilient to larger sets of data.

### DIFF
--- a/db/migrate/20200327105910_change_versions_custom_data_to_text.rb
+++ b/db/migrate/20200327105910_change_versions_custom_data_to_text.rb
@@ -1,0 +1,9 @@
+class ChangeVersionsCustomDataToText < ActiveRecord::Migration
+  def up
+    change_column :versions, :custom_data, :text
+  end
+
+  def down
+    change_column :versions, :custom_data, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20200209163549) do
+ActiveRecord::Schema.define(:version => 20200327105910) do
 
   create_table "adjustment_metadata", :force => true do |t|
     t.integer "adjustment_id"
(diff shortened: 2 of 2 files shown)


Matt-Yorkley:
The data saved in that field will perpetually grow bigger and bigger over time, right?

luisramos0:
only if the same schedule is used all the time. the order cycles list in the schedule will be bigger and bigger.

luisramos0:
agree with Pau, this is not going to break anything for sure. pr-no-test

Matt-Yorkley:
> only if the same schedule is used all the time

And that's exactly what _is_ happening live, right?

RachL:
I'm interested by this: is it 04 live OC or just 40 OC in the database? Because the same schedule is *always* used all the time yes...

luisramos0:
yeah, people tend to use the same schedule all the time so those schedules will have a long list of OCs as time goes by, so this text field will grow large, yes. The case live is with 40 OCs and the varchar 255 chars long exploded. that's why we are changing to text here.

PR #5088, opened 2020-03-27
Do not reset the dev env when booting docker

sauloperez:
#### What? Why?

The current web container's command destroys anything you might have in your local DB from a previous session, assuming you always want to start from a clean environment. This is hardly the case and makes `docker-compose up` take quite long. What if you just stopped containers temporally while developing? All the sample orders you created are gone :see_no_evil: 

This changes the approach to not assume anything. If you need to install a new gem or reset your DB just run the commands you would without docker. You can run anything you want with `docker-compose run web bundle exec <rails/rake command>` anyway.

For someone setting things for the first time, the `Dockerfile` process still installs all dependencies.

#### What should we test?

Nothing. I know it works because I always comment out these lines.

#### Release notes

Do not reset the development environment when starting docker
Changelog Category: Changed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,10 +28,7 @@ services:
       ADMIN_PASSWORD: ofn123
       OFN_DB_HOST: db
     command: >
-      bash -c "(bundle check || bundle install) &&
-               wait-for-it -t 30 db:5432 &&
-               bundle exec rake db:reset &&
-               bundle exec rake db:test:prepare ofn:sample_data || true &&
+      bash -c "wait-for-it -t 30 db:5432 &&
                rm -f tmp/pids/server.pid &&
                bundle exec rails s -p 3000 -b '0.0.0.0'"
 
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -28,10 +28,7 @@ services:
       ADMIN_PASSWORD: ofn123
       OFN_DB_HOST: db
     command: >
-      bash -c "(bundle check || bundle install) &&
-               wait-for-it -t 30 db:5432 &&
-               bundle exec rake db:reset &&
-               bundle exec rake db:test:prepare ofn:sample_data || true &&
```
Matt-Yorkley, line 34: So the test db and sample data won't be loaded when it's first used?

luisramos0:
this is not working for a fresh setup, we need to re-add the initial setup instructions somewhere.
I added them to Dockerfile and Docker.md here: https://github.com/coopdevs/openfoodnetwork/pull/2
I think it's working now for a fresh install with some extra manual commands to be done.

Issue #5069, opened 2020-03-25
Failed checkout on FR prod related out of stock product in cart


RachL:
## Description

I have a user who cannot checkout on Guillaume's hub `Le jardin de deux mains`  hub ID 370

She added 11 products to her cart, filed all her info, but she cannot checkout : both stripe and cash.

I cannot find her in the list of orders, she a new customer (potential guest even) with email I can give in PM.

What happens is that when clicking on the final button... nothing happens. Silent checkout fail.

## Steps to Reproduce
1. Don't know yet :/
2.
3.
4.

## Workaround
We activated temporaly cash, but that's not it either.

## Severity
bug-s1: a critical feature is broken: checkout, payments, signup, login

## Your Environment
* Version used: v2.8
* Browser name and version: Firefox 73
* Operating System and version (desktop or mobile): Desktop

RachL:
Ok so current lead: some product like turnip went out of stock while she was trying to checkout. She re-validated the cart with me on the phone, so it is possible that we have a problem of checking stock at the cart stage.

RachL:
So yes, he turnip became out of stock, this is why checkout was failing. They could checkout without turnip;

Matt-Yorkley:
Firstly, there's nothing at all special about the **troublesome turnip** that broke the checkout. 

I played with some specs, and it looks like this is a very simple case:
- If a product is added to the cart and goes out of stock _before_ the user visits the checkout page, they get redirected to the cart page with an "out of stock" error
- If the product goes out of stock _after_ they've loaded the checkout page... the checkout breaks, and it's not pretty...

lin-d-hop:
The troublesome turnip made me laugh out loud.

I honestly think we are seeing this a lot this week. Since sales quadrupled in the space of 2 weeks its not surprising.

Can we fix this? ASAP? 
Like... can we get it in the next release?

So many mysterious cart failures and it really really doesn't look good :-(

Matt-Yorkley:
I need to investigate a bit more...

Matt-Yorkley:
Something was wrong with my initial tests. Looks like the checkout handles that case fine, and we have a separate feature spec for it...

So... back to looking for theories. :disappointed:

lin-d-hop:
Ergh, twas too good to be true

Matt-Yorkley:
It does look related to two people buying the last units of a product at the same time...

lin-d-hop:
[...]

PR #5089, opened 2020-03-27
Cart variant scoping

### LINKED ISSUE #3222: Broken stock levels on product list page when product added and removed quickly from basket

Matt-Yorkley:
#### What? Why?

Closes #3222 
Possibly cause of S2: #5069

The code in the original issue has been moved around and refactored several times since it was opened, but the issue was still there.

Adds two tests for cases where `VariantsStockLevels` was used when populating the cart but no scoping was being done to the variants, and fixes both cases so the correct stock values are returned.

#### What should we test?

Adding/removing variants with variant overrides in the cart, should give correct feedback based on the override's stock, not the variant's stock.

#### Release notes

Updated cart stock logic to include variant overrides

Changelog Category: Fixed

### DIFF
--- a/app/services/variants_stock_levels.rb
+++ b/app/services/variants_stock_levels.rb
@@ -14,22 +14,10 @@ class VariantsStockLevels
       variant_stock_levels[variant_id] = { quantity: 0, max_quantity: 0, on_hand: variant.on_hand, on_demand: variant.on_demand }
     end
 
-    # The code above is most probably dead code, this bugsnag notification will confirm it
-    notify_bugsnag(order, requested_variant_ids, order_variant_ids) if missing_variant_ids.present?
-
     variant_stock_levels
   end
 
   private
-
-  def notify_bugsnag(order, requested_variant_ids, order_variant_ids)
-    error_msg = "VariantsStockLevels.call with variants in the request that are not in the order"
-    Bugsnag.notify(RuntimeError.new(error_msg),
-                   requested_variant_ids: requested_variant_ids.as_json,
-                   order_variant_ids: order_variant_ids.as_json,
-                   order: order.as_json,
-                   line_items: order.line_items.as_json)
-  end
 
   def variant_stock_levels(line_items)
     Hash[
--- a/app/services/variants_stock_levels.rb
+++ b/app/services/variants_stock_levels.rb
@@ -1,6 +1,8 @@
 # Report the stock levels of:
 #   - all variants in the order
 #   - all requested variant ids
+require 'open_food_network/scope_variant_to_hub'
+
 class VariantsStockLevels
   def call(order, requested_variant_ids)
     variant_stock_levels = variant_stock_levels(order.line_items)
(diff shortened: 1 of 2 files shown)


app/services/variants_stock_levels.rb
```diff
@@ -34,12 +34,21 @@ def notify_bugsnag(order, requested_variant_ids, order_variant_ids)
...
+           on_hand: variant.on_hand,
+           on_demand: variant.on_demand }]
       end
     ]
   end
+
+  def scoped_variant(distributor, variant)
+    return variant if distributor.blank?
+
+    OpenFoodNetwork::ScopeVariantToHub.new(distributor).scope(variant)
```
luisramos0: I think you need to memoize the scoper like this
```ruby
  def scoper(distributor)
    @scoper ||= OpenFoodNetwork::ScopeVariantToHub.new(distributor)
  end
```
this will avoid loading all the hubs overrides several times:
```ruby
  class ScopeVariantToHub
    def initialize(hub, variant_overrides = nil)
      @hub = hub
      @variant_overrides = variant_overrides || VariantOverride.indexed(@hub)
    end
```

luisramos0:
I have moved this up in the test ready column as it is linked to an S2 #5069

filipefurtad0:
Hi @Matt-Yorkley, 
I tried reproducing the bug, by trying out the basic functionality of product stock override, but no success - I didn't reproduce @RachL 's S2 bug #5069. Cash checkout works fine, as well as warnings for stock limitations imposed by inventory settings.

So I staged the PR, and:
- Backoffice ->limited the stock in inventory (override)
- Shopfront -> added several products to cart, testing stock override.
- Verified that the limitation set in the inventory works, for the respective variants 
- Checked out with cash.

I believe this is good to go.

PR #5103, opened 2020-03-29
remove blue sentence in payment method setting

### LINKED ISSUE #5092: Remove blue sentence in payment method setting
## Description
Since the start of the work on SCA we have now a sentence that is displayed below the provider dropdown:
image

We should remove it : it's a bit ugly and does not help at all.

## Steps to Reproduce
1. Go to `/admin/payment_methods/new`
2. Switch from cash to paypal or stripe
3. Boom

## Severity

## Your Environment
* Version used: v2.8

jeduardo824:
#### What? Why?

Closes #5092 

When you select an item from the list of providers on the payment_method settings, there's an ugly blue sentence showing up below the provider selector. The problem was happening because the `@payment_method` variable was being printed on the view. I removed it, and everything seems fine now.

Before:
Screen Shot 2020-03-29 at 20 34 32

After:
Screen Shot 2020-03-29 at 20 26 45

#### What should we test?

The Payment Method settings

#### Release notes

Remove blue sentence under the provider selector on the payment method settings page

Changelog Category: Fixed

### DIFF
--- a/app/views/spree/admin/payment_methods/_provider_settings.html.haml
+++ b/app/views/spree/admin/payment_methods/_provider_settings.html.haml
@@ -1,4 +1,3 @@
-= @payment_method
 - case @payment_method
 - when Spree::Gateway::StripeConnect
   = render 'stripe_connect'


filipefurtad0:
Olá @jeduardo824 ,
Verified that before staging, this appeared as described in the issue.
Thanks for this improvement - the blue line is gone after staging this PR.
Moving to ready to go.

PR #5121, opened 2020-04-01
Update vendor/assets/angular-google-maps.min.js

### LINKED ISSUE #5109: Make frontoffice app work even if stripe and/or maps URLs are down or break with JS errors
## Description
Originated in #5080

We need to make our app independent of stripe and maps. If there are problems with these URLs and/or scripts, the webapp needs to keep working.

Stripe and maps are down, the webapp keeps working.
Stripe js code blows up with some JS error, the webapp keeps working.

Matt-Yorkley:
What? Why?

Closes #5109 

Fixes an issue where if the js library from maps.googleapis.com failed to load in the `<head>`, all of our subsequent Angular would completely break.

Updates the `angular-google-maps.js` to a newer version, which apparently handles this case gracefully.

See: https://github.com/angular-ui/angular-google-maps

Note: `bluebird.js` is a new dependency of `angular-google-maps.js`.

#### What should we test?

- **Test phase 1:** Check `/maps` is working and searching for enterprises by location in the `/shops` page is working.
- **Test phase 2:** Get a dev to comment out the maps js from the darkswarm layout (line 42) and reload unicorn. This will simulate the maps js failing to load. Test again, checking that the login modal works, and the `/shops` page is not totally broken. The `/maps` page _should_ be broken.

#### Release notes

Updated angular-google-maps

Changelog Category: Fixed


luisramos0:
is this in dev or test ready?

Matt-Yorkley:
I'm just double-checking something I made need to adjust some directive names that have changed namespace...

luisramos0:
I will be offline tomorrow but whatever you do here to make it work I approve :rofl:

Matt-Yorkley:
Ok, this was a bit of a nightmare. I upgraded `angular-google-maps` and it fixed the problem (versions after 2.0.0 gracefully handle the maps js not loading). But then it turned out a lot of the later versions were buggy and created additional problems, so I had to play around with downgrading the version again to find one that actually worked :see_no_evil: 

There's also another part of this issue: the rest of our Angular code will generally work now if the maps js fails to load, except for the pages that rely on maps functionality. Obviously that includes `/map`, but it also includes `/shops`, which uses it extensively for location-based searching. 

I've added some tweaks to our Geocoding service so that the `/shops` page still works in that case.

app/assets/javascripts/darkswarm/services/geo.js.erb.coffee
```diff
@@ -1,6 +1,6 @@
 Darkswarm.service "Geo", ->
   new class Geo
-    OK: google.maps.GeocoderStatus.OK
+    OK: google?.maps?.GeocoderStatus?.OK
```
luisramos0, line 3: :-)

app/assets/javascripts/darkswarm/all.js.coffee
```diff
@@ -9,6 +9,7 @@
 #= require angular-animate
 #= require angular-resource
 #= require lodash.underscore.js
+#= require bluebird.js
```
sauloperez, line 12: A comment here stating this is required by angular-google-maps.min.js, would come in handy.

Matt-Yorkley: Comment added :+1:

filipefurtad0:
Hey @Matt-Yorkley ,

I've deployed and went through Test phase 1, but I noticed now there are some check pending. Should I re-stage after all checks are completed and ping you, after I repeat phase 1?

Cheers.

filipefurtad0:
Hi @Matt-Yorkley ,

I went through: 
- Test phase 1: Checked /maps is working and searching for enterprises by location in the /shops page as well

- Test phase 2: After your commenting-out of the maps js from the darkswarm layout (line 42) the maps page was broken (pic below) as well as searching shops by location in the shopfront. However, shops were working, and logging-in to different accounts worked well.

image

Moving to ready to go!

sauloperez:
this is pretty cool! we're making OFN resilient! :tada:

PR #5133, opened 2020-04-02
Cache counts used in homepage for 24 hours

### LINKED ISSUE #5123: Cache home page counters for a day

We need to cache the home page so that we reduce a bit the pressure on the instance, namely the DB. The counts of producers, shops, shoppers, and orders that we show at the bottom of the home page are the root cause of 2 of the 10 top queries where our servers spend the most time.

In the last week, we got reports of the app being slowed and these queries taking more than 2 seconds (hard to know if as a consequence or cause). We even got one at 3.15min :see_no_evil: 

It was agreed in the last delivery train that we'll cache those counters for a day. And we want to explore HTTP caching. It might be the best bang for the buck.

We must beware of https://github.com/openfoodfoundation/openfoodnetwork/blob/78170bc7098df3415a3e86766e41aaba9da10500/app/controllers/application_controller.rb#L131-L135.

### Context

Now, that information is far below the fold, it's very unlikely users check it, and if they do, the information doesn't need to be real-time accurate. With this steady increase on requests we are experiencing we need to make better use of the system resources before we get :fire: #scalability.

### Impact and timeline

Saving some requests on one of the most visited pages (it's among the top 5 of visited URLs), will already have a noticeable impact.

Matt-Yorkley:
#### What? Why?

Closes #5123

Caches instance-level stats displayed in homepage for 24 hours to avoid constant re-fetching of the same values via expensive queries.

#### What should we test?

Stats show on the homepage OK. The queries are triggered once per day.

#### Release notes

Cache stats displayed on homepage for 24 hours.

Changelog Category: Changed

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -5,12 +5,23 @@ class HomeController < BaseController
 
   def index
     if ContentConfig.home_show_stats
-      @num_distributors = Enterprise.is_distributor.activated.visible.count
-      @num_producers = Enterprise.is_primary_producer.activated.visible.count
-      @num_users = Spree::Order.complete.count('DISTINCT user_id')
-      @num_orders = Spree::Order.complete.count
+      @num_distributors = cached_count('distributors', Enterprise.is_distributor.activated.visible)
+      @num_producers = cached_count('producers', Enterprise.is_primary_producer.activated.visible)
+      @num_orders = cached_count('orders', Spree::Order.complete)
+      @num_users = cached_count(
+        'users', Spree::Order.complete.select('DISTINCT spree_orders.user_id')
+      )
     end
   end
 
   def sell; end
+
+  private
+
+  # Cache the value of the query count for 24 hours
+  def cached_count(key, query)
+    Rails.cache.fetch("home_stats_count_#{key}", expires_in: 1.day, race_condition_ttl: 10) do
+      query.count
+    end
+  end
 end


mkllnk:
Testing:

- [x] I deployed this to au-staging.
- [x] I verified that the stats still display the same as before.
- [x] I verified that they stay the same after I placed an order (cached).
- [ ] I haven't verified that the stats are updated after 24 hours. That's not possible if we want to include this in the next release. I think it's ready to go.

luisramos0:
awesome, thanks Maikel for taking care of this one!
I like the fact you guys did it twice this week: Matt finished your work with bugsnagJS and now you finish Matts work here :rocket:

sauloperez:
Even beyond adding any missing indexes [slow counting](https://wiki.postgresql.org/wiki/Slow_Counting) is even described in PostgreSQL wiki. They even implemented Count estimates. So, we got to be realistic here and weigh the pros and cons of the many counts we have.

> I'm expecting though that new instances will tell us that it's broken because they create an enterprise and it's not reflected there...

In this case, @mkllnk still taking that into account the performance gains outweigh this cost IMO.

luisramos0:
@RachL removed the label but nothing was deployed to AU so the test is still valid.

I looked at au staging now.
Original numbers from Maikel's post above:
770 food producers
178 food shops
1,254 food shoppers
4,609 food orders

Now:
image

The new food order is being counted :+1: 
This is working correctly :tada:

sauloperez:
Good news! it is still too early but checking a month's data, the query `SELECT COUNT ( DISTINCT user_id ) FROM spree_orders WHERE ( completed_at IS NOT ? )` shows up in the 9th position. Whereas checking a day's data it doesn't show up in the first page! :clap:

PR #5142, opened 2020-04-03
Don't load distributed properties on inactive distributors

Matt-Yorkley:
Part of #5143 

#### What? Why?

We don't need to query **distributed properties** and **distributed product properties** on enterprises that aren't active distributors, because they don't have products in active order cycles.

Tested locally with production data loaded; this change cuts page load on `/shops` by ~75% and removes ~300 very expensive and completely superfluous queries.

#### What should we test?

This needs a server with a good amount of data, probably Aus Staging...

- `/shops` page is working ok in general
- all shops can be filtered by "**Type**" and "**Delivery**" as before
- open shops can be filtered by their product properties as before
- when filtering by product "**Properties**" with open and closed shops, only open shops are shown in the results

#### Release notes

Improved performance on `/shops` page.

Changelog Category: Changed

### DIFF
--- a/spec/features/consumer/groups_spec.rb
+++ b/spec/features/consumer/groups_spec.rb
@@ -60,8 +60,8 @@ feature 'Groups', js: true do
       let!(:group) { create(:enterprise_group, enterprises: [d1, d2], on_front_page: true) }
       let!(:order_cycle) { create(:simple_order_cycle, distributors: [d1, d2], coordinator: create(:distributor_enterprise)) }
       let(:producer) { create(:supplier_enterprise) }
-      let(:d1) { create(:distributor_enterprise) }
-      let(:d2) { create(:distributor_enterprise) }
+      let(:d1) { create(:distributor_enterprise, with_payment_and_shipping: true) }
+      let(:d2) { create(:distributor_enterprise, with_payment_and_shipping: true) }
       let(:p1) { create(:simple_product, supplier: producer) }
       let(:p2) { create(:simple_product, supplier: create(:supplier_enterprise)) }
       let(:ex_d1) { order_cycle.exchanges.outgoing.where(receiver_id: d1).first }
--- a/app/serializers/api/cached_enterprise_serializer.rb
+++ b/app/serializers/api/cached_enterprise_serializer.rb
@@ -73,12 +73,16 @@ module Api
 
     # This results in 3 queries per enterprise
     def distributed_properties
+      return [] unless active
+
       (distributed_product_properties + distributed_producer_properties).uniq do |property_object|
         property_object.property.presentation
       end
     end
 
     def distributed_product_properties
+      return [] unless active
+
       properties = Spree::Property
         .joins(products: { variants: { exchanges: :order_cycle } })
         .merge(Exchange.outgoing)
@@ -91,6 +95,8 @@ module Api
     end
 
     def distributed_producer_properties
+      return [] unless active
+
       properties = Spree::Property
         .joins(
           producer_properties: {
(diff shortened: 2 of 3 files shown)


Matt-Yorkley:
I updated the specs, as some of them were not using active distributors in the test setup. 

I also had to change the cached_enterprise_serializer spec as it was checking that distributed properties _should_ be fetched and serialized even if the enterprise is not an active distributor...

Please correct me if this doesn't seem right! :pray:

spec/controllers/shops_controller_spec.rb
```diff
@@ -7,7 +7,8 @@
   let!(:distributor) { create(:distributor_enterprise) }
 
   before do
-    allow(OpenFoodNetwork::EnterpriseInjectionData).to receive(:active_distributor_ids) { [distributor.id] }
+    allow_any_instance_of(OpenFoodNetwork::EnterpriseInjectionData).
```
Matt-Yorkley, line 10: This line was actually not working at all in the tests...

spec/features/consumer/groups_spec.rb
```diff
@@ -60,8 +60,8 @@
       let!(:group) { create(:enterprise_group, enterprises: [d1, d2], on_front_page: true) }
       let!(:order_cycle) { create(:simple_order_cycle, distributors: [d1, d2], coordinator: create(:distributor_enterprise)) }
       let(:producer) { create(:supplier_enterprise) }
-      let(:d1) { create(:distributor_enterprise) }
-      let(:d2) { create(:distributor_enterprise) }
+      let(:d1) { create(:distributor_enterprise, with_payment_and_shipping: true) }
+      let(:d2) { create(:distributor_enterprise, with_payment_and_shipping: true) }
```
Matt-Yorkley, line 64: These are now active / open shops in the `/shops` page :+1:

luisramos0:
Are you not changing behaviour of the page?

I can filter to show closed shops and some inactive enterprises show like this:
image

But some show like this:
image

are you not removing this list below the enterprise with this PR?
If the list remains, all good.
If this list will disappear, maybe it is still an acceptable PR but it needs product approval.

Matt-Yorkley:
I'm pretty sure those are loaded asynchronously now with a different serializer, but I'll double-check...

Matt-Yorkley:
Yep, it's asynchronous, uses a different serializer, and is still loading that info as before. :+1:

Matt-Yorkley:
It will change he behavior of the page in that the user won't be able to filter closed shops by the properties of products they would theoretically sell if they were open...

luisramos0:
ok, filtering by properties will not return any closed shops, for example, this (Demeter shops in OFF) will only return the open shop after this PR:
image
I'll post this on product-curation for validation.

luisramos0:
I wonder if this has any impact on the producers page where there are similar filters as well.

Matt-Yorkley:
It doesn't affect the producers page as `supplied_properties` is untouched in the serializer.

Matt-Yorkley:
It's changing one small part of some of the existing behavior (only for closed shops) in a way that makes sense: not filtering on **properties of distributed products** for shops that are **not distributing any products**.

luisramos0:
"properties of **distributed** products for shops that are not distributing any products."
This distributed word makes me see this is introducing a **very minor issue**, in my example above of searching for shops with demeter products in closed shops": because all we are would not show with this PR are hubs that have had OCs opened selling demeter products but today they dont have any.

sauloperez:
We gave it the ok in yesterday's product curation meeting :point_right: https://community.openfoodnetwork.org/t/product-curation-meeting-10-notes-7-april-2020/1903

filipefurtad0:
Hey again @Matt-Yorkley,

Staged this in UK-Staging:
-  /shops page is working ok in general - nothing unusual
-  "Type" and "Delivery" filters works on all shops, displaying both closed and open shops.
-  after showing open and closed shops, the "Properties" filter only shows open shops in the results.
- "Properties" works fine on open shops

Moving to ready to go.

Matt-Yorkley:
Rebased to resolve merge conflict...

Issue #5131, opened 2020-04-02
Unsaved changes in the shop inventory


SophieDuponcheel:
## Description

On the Oxfam Magasins du Monde Rochefort shop catalogue, I try to modify the stocks of products coming from different producers. These same producers supply other Oxfam stores. Oxfam-Rochefort has full permission to add products and change the stocks of products in its catalogue. Changes to stocks have been recorded each time. Since 2 days it doesn't work anymore for any catalogue (for other stores it doesn't work either). I should be able to change the stocks as I want, but an error message appears at the bottom of the page telling me that the changes have not been saved.

This is a real problem because the sotcks are very different from one store to another, so we need to have the right stock up to date on the catalogue of the different stores.

## Expected Behavior

I should be able to change the stocks as I want

## Actual Behaviour

an error message appears at the bottom of the page telling me that the changes have not been saved.

## Steps to Reproduce

https://www.openfoodnetwork.be/admin/inventory

1. I'm going to the Oxfam Rochefort World Shops store catalogue, 
2. I change the stock of "noodles 250g" to 3
3. I'm saving the changes, it takes a long time to load.
4. An error message appears, the changes have not been saved. 

## Animated Gif/Screenshot

![Screenshot from 2020-04-02 13-51-50](https://user-images.githubusercontent.com/8669848/78251390-2e3a5c00-74e9-11ea-8f64-b94c4b5612b8.png)

504 Error

## Severity

bug-s2: a non-critical feature is broken, no workaround

## Your Environment

* Version used: 2.8.2
* Browser name and version: Chrome and firefox
* Operating System and version (desktop or mobile): PC windows 10 and ubuntu

tschumilas:
Is this French error the same as this English error @RachL ?   If so -  it seems that inventories are broken for this login only.  Add a different manager login and try again @SophieDuponcheel   
It could be a different problem -  I'll test today, I'm doing other inventory testing anyway.  But I've broken inventories and gotten this message 3 times now -   and I have no idea why.  I think something to do with adjusting permissions,  and then inventory?   Anyway here is my issue - is it the same: https://github.com/openfoodfoundation/openfoodnetwork/issues/5090

luisramos0:
yeah, reducing permissions for the user doing this change may help as a workaround.
So a workaround **could** be: make sure the user trying to execute the task has only access to the products that need to be changed.

This inventory page completely breaks if there is a single broken variant that the user has access to.

[...]

PR #5145, opened 2020-04-03
Performance improvements to Inventory page

luisramos0:
#### What? Why?

Improves situation in #5131

This PR removes on N+1 query on variant_overrides.variant.product

#### What should we test?
We need to validate inventory is working as usual.

#### Release notes
Changelog Category: Changed
Improved a little the performance of the inventory page.

### DIFF
--- a/app/controllers/admin/variant_overrides_controller.rb
+++ b/app/controllers/admin/variant_overrides_controller.rb
@@ -73,8 +73,10 @@ module Admin
     end
 
     def collection
-      @variant_overrides = VariantOverride.includes(:variant).for_hubs(params[:hub_id] || @hubs)
-      @variant_overrides.select { |vo| vo.variant.present? }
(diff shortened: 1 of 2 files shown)


filipefurtad0:
Hi @luisramos0,
There is an issue with codeclimate check:
Line is too long. [103/100]
If understand this correctly, it refers to a comment, so it should be fine to stage/test - right?

Matt-Yorkley:
Yeah, it's fine to stage and test :+1:

filipefurtad0:
Hi @luisramos0,

Tested the functionality of the inventory feature, checking that product overrides and changes made in the back-office had the expected effect in the shopfront - all good.

Specifically I made sure that the issue described in #5131 is now solved: saving changes was fast, and no error message appeared.

Moving to ready to go.

luisramos0:
great!!

PR #5146, opened 2020-04-03
Add PROFILE var to set production-like settings

sauloperez:
#### What? Why?

Related to #5131 

I took this from a recent newsletter I got (from the guy behind http://www.speedshop.co/). Sometimes replicating performance issues locally is actually slower than production due to dev mode settings (code reloading, etc.), heavy dev-only gems and the asset pipeline.

The PROFILE env var switches these settings all at the same time, giving us an environment closer to production, essential for reliable profiling. Then, rack-mini-profiler is going to be more accurate.

Apparently it's something [RubyGems](https://github.com/rubygems/rubygems.org/blob/b026df86ae9df0e4d4c7dae2f220fe2219d52350/config/environments/development.rb#L72-L92) and [CodeTriage](https://github.com/codetriage/CodeTriage/commit/a3c957647dc039d73036c565620643cf216e30c7) both use.

I provided useful to investigate cache issues with some assets.

#### What should we test?

Development only

#### Release notes

Added PROFILE env var for production-like performance profiling.

Changelog Category: Added

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -11,7 +11,7 @@ Openfoodnetwork::Application.configure do
   config.cache_classes = !!ENV["PROFILE"]
 
   # :file_store is used by default when no cache store is specifically configured.
-  # config.cache_store = :file_store
+  config.cache_store = :memory_store if !!ENV["PROFILE"]
 
   # Log error messages when you accidentally call methods on nil.
   config.whiny_nils = true
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -1,10 +1,14 @@
 Openfoodnetwork::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
+  #
+  # PROFILE switches several settings to a more "production-like" value
+  # for profiling and benchmarking the application locally. All changes you
+  # make to the app will require restart.
 
   # In the development environment your application's code is reloaded on
   # every request.  This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.cache_classes = !!ENV["PROFILE"]
 
   # :file_store is used by default when no cache store is specifically configured.
   # config.cache_store = :file_store
@@ -14,7 +18,7 @@ Openfoodnetwork::Application.configure do
 
   # Show full error reports and disable caching
   config.consider_all_requests_local       = true
-  config.action_controller.perform_caching = false
+  config.action_controller.perform_caching = !!ENV["PROFILE"]
 
   # Don't care if the mailer can't send
   config.action_mailer.raise_delivery_errors = false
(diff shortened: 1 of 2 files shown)


sauloperez:
Yes, this deserves a section somewhere. I'll get to it.

config/environments/development.rb
```diff
@@ -28,8 +32,14 @@
...
 
+  # Generate digests for assets URLs.
+  #
+  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
+  # yet still be able to expire them through the digest params.
+  config.assets.digest = !!ENV["PROFILE"]
+
   # Expands the lines which load the assets
-  config.assets.debug = false
+  config.assets.debug = !!ENV["PROFILE"]
```
luisramos0, line 42: I gave this PR a try but maybe this settings is messing it up.
I dont get the assets bundled :man_shrugging: 
thsi is not production like:
image

sauloperez:
you need to precompile them running `bundle exec rake assets:precompile:primary assets:precompile:nondigest` as we do in deployments.

Sorry, I have an extra fix for profiling coming plus the documentation but I had other duties on top of this :pray:

luisramos0:
ok, do you know how do I test "config.assets.digest" and cache_classes?

sauloperez:
oh yes, you are right, `config.assets.debug = true` prevents Rails from serving bundled assets in a `application.js` or `application.css`. I'll fix it.

sauloperez:
@luisramos0 I fixed it and extended the documentation. I hope it's now clear for everyone. Do you find it enough or do you still think we need to doc in the wiki?

kristinalim:
Without the `PROFILE` set, I am now getting 364 requests when loading the home page:

20200428135646-assets_pr

While it is just 42 in master:

20200428140948-assets_master

This wasn't part of the plan, right @sauloperez?

sauloperez:
That's very strange @kristinalim . Just booting things up with `docker-compose up` with PROFILE set up and visiting the `/` results in a total of 41 requests, 3 of which are `/mini-profiler-resources`, like those the show up in your screenshot.

When I do the same without PROFILE the number goes up to 362 (one for each JS asset file) and still 3 for rack mini profiler. Are you sure you have the env var set up?

Other than that, anything related to rack mini profiler falls outside of the scope of this PR and it's one of those things that make this closer to production but not exactly the same.

kristinalim:
@sauloperez Sorry if I misunderstand.

It's also around the same for me: "When I do the same without PROFILE the number goes up to 362 (one for each JS asset file) and still 3 for rack mini profiler."

My point was, did you intend e.g. for `all.js` to be split in multiple files? They were concatenated in the same `all.js` before this PR.

sauloperez:
Now I understand, yes, that's was also intended. It's done with

```
config.assets.debug = !ENV["PROFILE"]
```

This way we can more easily debug the assets. I don't care that much about the JS files but other assets such images and fonts.

config/environments/development.rb
```diff
@@ -28,8 +32,25 @@
...
+  #
+  # Disabling asset debugging still requires that assets be compiled for each
+  # request. You can avoid that by precompiling the assets as in production:
+  #
+  #   $ bundle exec rake assets:precompile:primary assets:precompile:nondigest
+  #
+  # You can remove them by simply running:
+  #
+  #   $ bundle exec rake assets:clean
+  config.assets.debug = !ENV["PROFILE"]
```
luisramos0, line 53: I missed this single !
that means this PR changes everyone's local setup to separate all js files in their dev tools.
I dont like this, search all files in devtools (CTRL SHIFT F) is not so fast as searching in one single file with all the JS.

But if I set PROFILE I get all the other production like settings....
Do you mind if I create a PR to change this line to maybe use another env var like ENV["DEBUG_ASSETS"]?

sauloperez: fine for me :+1:

luisramos0: Here's the PR:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5344

I found this also makes my local server slower. I think it's because of the hundreds of js files that need to be downloaded and used.

PR #5153, opened 2020-04-04
Bye bye Google Analytics

### LINKED ISSUE #5149: Remove support for Google Analytics Trackers
## Description
The trackers page has not been copied to OFN before removing spree_backend. That means this page is giving a snail now in all production servers:
https://openfoodfrance.org/admin/trackers

I checked all 8 globally managed servers and only UK and AU have a tracker in the database, both from 2016:
AU UA-62912229-3 updated on 2016-05-25
UK UA-77349822-1 updated on 2016-05-05

After slack validation with Rachel, Lynne and Kirsten all confirmed we can remove support of Google Analytics. :heart: 
There a quick clean up to be done. We need to remove all code related to trackers and google analytics including the DB table with the trackers.

btw these trackers have been removed from spree in v3.5 into a separate gem: https://guides.spreecommerce.org/release_notes/3_5_0.html#extracted-analytics-tracker-into-an-extension

luisramos0:
#### What? Why?

Closes #5149

Removes support for GA. IT removes the trackers config page as well as the GA related content in the cookies banner and cookies explanation page. 

This will remove the trackers from AU and UK live and thus make the apps a bit faster for the user.

#### What should we test?
Very trackers option is gone from configurations.
Verify cookies banner GA preference on the general settings config page is also gone.
Verify cookies banner and cookies policy page still work with and without matomo activated:
- de-activate matomo toggle on general settings
image
- verify the cookies policy page have the correct content without matomo info
homepage:
image
- activate matomo and verify the cookies policy page now includes matomo info
image
- In the above cases, verify GA info is not there :-) 

#### Release notes
Changelog Category: Removed
Removed support for Google Anayltics


engines/web/app/views/web/angular_templates/cookies_policy.html.haml
```diff
@@ -44,32 +44,20 @@
...
+    = t 'legal.cookies_policy.statistics_cookies_matomo_desc_html'
 
   %table{ng: { controller:"CookiesPolicyModalCtrl"}}
-    - if Spree::Config.cookies_policy_ga_section
-      = render_cookie_entry( "__utma", t( "legal.cookies_policy.cookie_analytics_utma_desc" ) )
-      = render_cookie_entry( "__utmt", t( "legal.cookies_policy.cookie_analytics_utmt_desc" ) )
-      = render_cookie_entry( "__utmb", t( "legal.cookies_policy.cookie_analytics_utmb_desc" ) )
-      = render_cookie_entry( "__utmc", t( "legal.cookies_policy.cookie_analytics_utmc_desc" ) )
-      = render_cookie_entry( "__utmz", t( "legal.cookies_policy.cookie_analytics_utmz_desc" ) )
-    - if Spree::Config.cookies_policy_matomo_section
```
mkllnk, line 67: What if an instance isn't using Matomo. Why would we still display the cookie entry?

db/migrate/20200404090436_remove_ga_cookies_preference.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveGaCookiesPreference < ActiveRecord::Migration
+  def up
+    Spree::Preference
```
mkllnk, line 3: We should define that model within the migration to not depend on app code.

engines/web/app/views/web/angular_templates/cookies_policy.html.haml
luisramos0: there a if statement above already.
it used to be  if Spree::Config.cookies_policy_matomo_section || Spree::Config.cookies_policy_ga_section
and now it's just  if Spree::Config.cookies_policy_matomo_section
so this one is not needed anymore.

before
```
if matomo or ga
      HEADER
       if matomo    -- this is the IF you are asking about
            MATOMO
       if ga
            GA
```

after:
```
if matomo
      HEADER
      MATOMO

```

db/migrate/20200404090436_remove_ga_cookies_preference.rb
luisramos0: :+1: done

luisramos0:
yes, all credit to @RachL 
some of us are really happy about this one :tada: 
next gmaps :-)

luisramos0:
rebased to remove simple conflict.

filipefurtad0:
Hi @luisramos0 ,

This looks great. Staged this PR on Katuma and tested according to your notes. Compared the effects with another staging server (FR):

As Superadmin, verified that: 
- Trackers option is gone from configurations
- Cookies banner GA preference on the general settings config page is gone.
- Cookies banner and cookies policy page still work with and without matomo activated, i.e.:
- Deactivated matomo toggle on general settings
- Checked the policy page content is without matomo info
- Reactivated this and that the cookies policy page now includes matomo info

GA is as good as gone :tada:

luisramos0:
nice one Filipe!

Issue #5143, opened 2020-04-03
Reduce performance impact of /shops


sauloperez:
### What we should change and why (this is tech debt)

We need to reduce the impact the `/shops` page is having on our servers load by reducing its response time with quick and easy optimizations. It's holding its position as the worst app endpoint in the last month.

### Context

With the increased demand and the scarcity of resources in FR and UK servers this page's response times are going nuts, and this causing the DB to be on :fire: as well.

Today, the percentile 95 reaches 15secs sometimes, like this [request](https://app.datadoghq.com/apm/trace/5148661426660309856?env=production&sort=time&colorBy=service&graphType=flamegraph&shouldShowLegend=true ). And it's getting worse.

![Screenshot from 2020-04-03 13-35-27](https://user-images.githubusercontent.com/762088/78356784-80928000-75b0-11ea-9f49-5bcec053ce93.png)
Count of requests taking more than 10s in the last 4h for UK and FR together.

Even the latency distribution is taking a very weird and far from good shape.

![Screenshot from 2020-04-03 13-42-28](https://user-images.githubusercontent.com/762088/78357045-0282a900-75b1-11ea-8e32-498d8ae601b1.png)

### Impact and timeline

Any improvement in this page will have a big impact on the server load. Doing more with the same resources it's a good way to scale-up the system, isn't it?

sauloperez:
I tried an aggressive caching strategy using Rails [page caching](https://guides.rubyonrails.org/caching_with_rails.html#page-caching) but that won't work due to the dynamic nature of this page. Even if we go with other lower-level strategies dealing ourselves with cache invalidation, I'm afraid the complexity will be pretty big and it'll take time.

My personal conclusion is that we'll need to work pretty soon on refactoring /shops if we want to survive this :chart_with_upwards_trend: As per @matt-yorkley’s insights, there are no quick wins there unless we :fire: the tech-debt.

Matt-Yorkley:
>  the percentile 95 reaches 15secs sometimes

It's worse than that... :grimacing: 

![Screenshot from 2020-04-03 18-50-20](https://user-images.githubusercontent.com/9029026/78385390-0c6dd180-75dc-11ea-96db-e6a8cb6a4189.png)

sauloperez:
OH MY F***** GOD :see_no_evil:

Matt-Yorkley:
We can have another look at the data on this when the new release is deployed.

sauloperez:
[...]

PR #5156, opened 2020-04-04
Load closed shops in a separate request on /shops page

Matt-Yorkley:
#### What? Why?

Part of #5143

Loads only open shops by default on the `/shops` page, and loads closed shops separately when the "Show Closed Shops" button is clicked.

#### What should we test?

- `/shops` page loads a lot faster than before. 
- Closed shops are displayed as before, after the "Show Closed Shops" button is clicked. 
- Filtering works as before.

#### Release notes

Load closed shops separately on /shops page

Changelog Category: Changed

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/hub_node_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/hub_node_controller.js.coffee
@@ -24,7 +24,7 @@ Darkswarm.controller "HubNodeCtrl", ($scope, HashNavigation, CurrentHub, $http, 
     $scope.shopfront_loading = true
     $scope.toggle_tab(event)
 
-    $http.get("/api/enterprises/" + $scope.hub.id + "/shopfront")
+    $http.get("/api/shops/" + $scope.hub.id)
       .success (data) ->
         $scope.shopfront_loading = false
         $scope.hub = data
--- a/app/controllers/api/shops_controller.rb
+++ b/app/controllers/api/shops_controller.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Api
+  class ShopsController < BaseController
+    respond_to :json
+    skip_authorization_check only: [:show, :closed_shops]
+
+    def show
+      enterprise = Enterprise.find_by_id(params[:id])
+
+      render text: Api::EnterpriseShopfrontSerializer.new(enterprise).to_json, status: :ok
+    end
+
+    def closed_shops
+      @active_distributor_ids = []
+      @earliest_closing_times = []
+
+      serialized_closed_shops = ActiveModel::ArraySerializer.new(
+        ShopsListService.new.closed_shops,
+        each_serializer: Api::EnterpriseSerializer,
+        data: OpenFoodNetwork::EnterpriseInjectionData.new
+      )
+
+      render json: serialized_closed_shops
+    end
+  end
+end
--- a/app/assets/javascripts/darkswarm/controllers/enterprises_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/enterprises_controller.js.coffee
@@ -9,8 +9,12 @@ Darkswarm.controller "EnterprisesCtrl", ($scope, $rootScope, $timeout, $location
   $scope.show_closed = false
   $scope.filtersActive = false
   $scope.distanceMatchesShown = false
+  $scope.closed_shops_loaded = false
 
   $scope.$watch "query", (query)->
+    $scope.resetSearch(query)
+
+  $scope.resetSearch = (query) ->
     Enterprises.flagMatching query
     Search.search query
     $rootScope.$broadcast 'enterprisesChanged'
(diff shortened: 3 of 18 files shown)


luisramos0:
This makes me think about API first: what is the enterprises and shops endpoints we want to have?

Maybe we should separate enterprises from shops? An enterprise includes profiles and producers, etc Shops must have respective OrderCycle? Or are all producers closed_shops?

I think the concept of "recently closed shops" is important. An enterprise that never had an OC is very different from an enterprise that just closed it's weekly OC on a Friday and will open in on Tuesday as usual. The API should be prepared to take this.

Maybe we can have enterprises list all enterprises:
/api/enterprises list all enterprises
and then have shops endpoints:
/api/enterprises/shops  lists all open shops
/api/enterprises/shops/closed  lists all enterprises not ready for checkout

what do you think?

Matt-Yorkley:
> Maybe we should separate enterprises from shops?

Yes, I was thinking about looking at that (after making sure it all works). We now have two "shop" actions in `api/enterprises_controller`; `shopfront` (which is basically a show action) and `closed_shops`... maybe we could move these to a new `api/shops_controller` ? 

Both of these actions require this in the enterprise controller: `skip_authorization_check only: [:shopfront, :closed_shops]` but all the other actions don't. 

I feel like `api/enterprise_controller` is for CRUD actions on the actual enterprise resource (with authorisation), whereas these shop actions are for reading publicly-accessible representations of an enterprise that is selling to customers.

> I think the concept of "recently closed shops" is important. An enterprise that never had an OC is very different from an enterprise that just closed it's weekly OC on a Friday and will open in on Tuesday as usual. The API should be prepared to take this.

In theory yes, but we don't currently have anything that needs that specific info from the API, so I'd say YAGNI...

> Maybe we can have enterprises list all enterprises: ... /api/enterprises list all enterprises

I don't think we have a current use for that endpoint either...

luisramos0:
awesome! yes to api/shops_controller!

we dont have /api/enterprises, the list, just because we do inject_enterprises, otherwise we would need it. It's one of those that we should have, it's a very basic feature missing from the OFN api right now: list all enterprises (regardless of open/closed shops).

kirstenalarsen:
agree with @luisramos0 - having enterprises on the API would be ace - not least opening up our abilities to simply pull that information into other, better directory and mapping interfaces - so that we could potentially quickly and cheaply have a much more effective directory function :)

app/controllers/api/shops_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def show
+      enterprise = Enterprise.find_by_id(params[:id])
+
+      render text: Api::EnterpriseShopfrontSerializer.new(enterprise).to_json, status: :ok
+    end
+
+    def closed_shops
+      @active_distributor_ids = []
+      @earliest_closing_times = []
```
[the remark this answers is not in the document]
Matt-Yorkley: These are used in `OpenFoodNetwork::EnterpriseInjectionData.new` a few lines later, and will trigger big queries (that we don't need in the case of closed shops) unless these two ivars are already set.

Matt-Yorkley:
@mkllnk 

> ...three buttons show for a very short time

Fixed :+1:

> A loading indicator would be awesome

Added :+1:

> the filter options are not updating correctly

Give it another try. The taxons "Filter by Type" are loading correctly and working. Are they set up correctly in yoyr environment? Clicking the "Filter by Property" buttons also works correctly...

app/assets/javascripts/darkswarm/controllers/hub_node_controller.js.coffee
```diff
@@ -24,7 +24,7 @@ Darkswarm.controller "HubNodeCtrl", ($scope, HashNavigation, CurrentHub, $http,
     $scope.shopfront_loading = true
     $scope.toggle_tab(event)
 
-    $http.get("/api/enterprises/" + $scope.hub.id + "/shopfront")
+    $http.get("/api/shops/" + $scope.hub.id)
```
sauloperez, line 27: :heart_eyes:

app/controllers/api/shops_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    respond_to :json
+    skip_authorization_check only: [:show, :closed_shops]
+
+    def show
+      enterprise = Enterprise.find_by_id(params[:id])
+
+      render text: Api::EnterpriseShopfrontSerializer.new(enterprise).to_json, status: :ok
+    end
+
+    def closed_shops
```
sauloperez, line 14: IMO this should be a filter specified by URL query param for an `API::ShopsController#index` so we follow Rails REST conventions. But I want to discuss this in another PR and without any rush. This is awesome as it is.

filipefurtad0:
Hey @Matt-Yorkley,

I had a look at the upload times of the /shops page, including closed shops, before and after staging this in UK-staging. It was in both cases around 30s, but I noticed my CPUs were over the top, so perhaps the limiting step is my system.

After UK-staging:
- Type, Property and Delivery filters worked fine, showing both open and closed shops. I also had a closer look at the "Type" filter, since @mkllnk observed some issues before. I chose different combinations, found no issues.
- Closed shops can be toggled, as usual.

Good to go.

PR #3742, opened 2019-04-15
Merge 2-0-stable to master

luisramos0:
Are we merging 2-0-stable with a PR like this?
Or do we force push 2-0-stable into master?

@sauloperez @mkllnk @Matt-Yorkley @kristinalim

PR #4964, opened 2020-03-10
Fix proxy orders controller in rails 4 by removing the use of responders

luisramos0:
#### What? Why?

Closes #4951 by fixing the first spec.

We bypass using the responder that was looking for a template and just render the json. Looks like it works like this :+1:

#### What should we test?

We should test pausing and resuming a subscription as well as cancelling the subscription. We should verify these actions are still working.
I'd be very happpy to teach a tester how to use this:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Testing-Subscriptions

#### Release notes

Changelog Category: Changed
Adapt code to the rails 4 upgrade.

PR #5158, opened 2020-04-04
Merge 3-0-stable to master

luisramos0:
2-0-stable was merged in #3742 with:
image

v2 was 790 commits:
image

Here we are currently at:
image

v3 is 362 commits.

v3 has quite a lot of lines and commits, not good indicators of how easy this one was compared to the previous one :-)


luisramos0:
this build should be green after:
- these are merged in master and brought to 3-0-stable: #4841, #4920, #4964 and #4993
- these are merged to 3-0-stable: #5157, #5159

PR #5161, opened 2020-04-04
Add optional OOBGC support to Unicorn

Matt-Yorkley:
This is a proposal for an experiment that might help clear up memory in a more ordered fashion by doing small bits of garbage collection between requests...

More info here: http://tmm1.net/ruby21-oobgc/

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -122,6 +122,7 @@ gem 'ofn-qz', github: 'openfoodfoundation/ofn-qz', ref: '60da2ae4c44cbb4c8d602f5
 
 group :production, :staging do
   gem 'ddtrace'
+  gem 'gctools'
 end
 
 group :test, :development do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -410,6 +410,7 @@ GEM
     fuubar (2.5.0)
       rspec-core (~> 3.0)
       ruby-progressbar (~> 1.4)
+    gctools (0.2.4)
     geocoder (1.1.8)
     gmaps4rails (1.5.6)
     haml (4.0.7)
@@ -731,6 +732,7 @@ DEPENDENCIES
   foundation-icons-sass-rails
   foundation-rails
   fuubar (~> 2.5.0)
+  gctools
   geocoder
   gmaps4rails
   haml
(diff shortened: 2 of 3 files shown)


Matt-Yorkley:
Apparently it was built to improve the memory cleanup on `github.com` itself, albeit quite a few years ago...

This was just an idea, I'm not 100% sure if we should use it. And actually testing it would be tricky, as we probably can't get meaningful results unless it's in production. Maybe we could use the new Ruby GC monitoring with Datadog on Staging an automate sending a lot of requests to the server..?

luisramos0:
what's the risk of putting this in prod? what type of problem we can have? GC running too many times and taking too much CPU?

Maybe we should go with #5162 and leave this for later?

sauloperez:
> automate sending a lot of requests to the server..? 

I'm thinking the same. We should prepare some sort of load benchmarking

sauloperez:
IMO we should try this out in production. It's a big improvement on Ruby's GC which I can't see how it could make things worse other than our particular nuts workload. It's also impossible to predict without testing it in production.

What if we prepare a release only with this early next week and deploy it one or two days after we see the perf. improvements of the regular release? just to clearly see the impact of this alone.

Matt-Yorkley:
> It's a big improvement on Ruby's GC which I can't see how it could make things worse other than our particular nuts workload. It's also impossible to predict without testing it in production.

Agreed, I think this is _probably_ really awesome, but we have no idea until we try it in production. On the other hand, the primary reason for wanting to try this was downtime on French Prod due to lack of memory, and that is no longer a problem.

I guess the biggest potential benefit now is doing smaller (smarter) bits of GC at regular intervals between requests instead of waiting for ages and having all the GC done at once (which in theory could cause a CPU spike at totally random times).

I don't think there's any urgent need for this now, so I'm happy to put this in the sysadmin backlog and try it out when things are bit less crazy, to be honest.

sauloperez:
> I don't think there's any urgent need for this now, so I'm happy to put this in the sysadmin backlog and try it out when things are bit less crazy, to be honest.

Fine for me, but I would do it in a week or two. It's performance we'll get almost for free so how we could say no? given you dedicated some time to it, it feels to me like an investment with no returns unless it hits production.

Matt-Yorkley:
This was a cool idea, but Unicorn is long gone. Hasta la vista.

PR #5162, opened 2020-04-04
Add optional unicorn-worker-killer configs

Matt-Yorkley:
This is another proposal for an experiment that might help free up memory.

Today I noticed 3 of our unicorn workers were using a normal amount of memory (about 6-9% of total, each) and one worker was using around 60% of total memory (about 5GB) and had been that way for a long time.

Screenshot from 2020-04-04 20-26-41

I killed that worker (they restart automatically) and we got 60% of total memory back instantly:

Screenshot from 2020-04-04 20-28-06

There's a very nice gem that will do this gracefully and automatically when a worker has served a set amount of requests or when it get too big in it's memory usage, and I think we should use it!

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -1,4 +1,13 @@
 # This file is used by Rack-based servers to start the application.
 
+if ENV.fetch('KILL_UNICORNS', false) && ['production', 'staging'].include?(ENV['RAILS_ENV'])
+  # Gracefully restart individual unicorn workers if they have:
+  # - performed between 25000 and 30000 requests
+  # - grown in memory usage to between 700 and 850 MB
+  require 'unicorn/worker_killer'
+  use Unicorn::WorkerKiller::MaxRequests, 25_000, 30_000
+  use Unicorn::WorkerKiller::Oom, (700 * (1024**2)), (850 * (1024**2))
+end
+
 require ::File.expand_path('../config/environment', __FILE__)
 run Openfoodnetwork::Application
--- a/config.ru
+++ b/config.ru
@@ -5,8 +5,12 @@ if ENV.fetch('KILL_UNICORNS', false) && ['production', 'staging'].include?(ENV['
   # - performed between 25000 and 30000 requests
   # - grown in memory usage to between 700 and 850 MB
   require 'unicorn/worker_killer'
-  use Unicorn::WorkerKiller::MaxRequests, 25_000, 30_000
-  use Unicorn::WorkerKiller::Oom, (700 * (1024**2)), (850 * (1024**2))
+  use Unicorn::WorkerKiller::MaxRequests,
+      ENV.fetch('UWK_REQS_MIN', 25_000).to_i,
+      ENV.fetch('UWK_REQS_MAX', 30_000).to_i
+  use Unicorn::WorkerKiller::Oom,
+      ( ENV.fetch('UWK_MEM_MIN', 700).to_i * (1024**2) ),
+      ( ENV.fetch('UWK_MEM_MAX', 850).to_i * (1024**2) )
 end
 
 require ::File.expand_path('../config/environment', __FILE__)
--- a/Gemfile
+++ b/Gemfile
@@ -122,6 +122,7 @@ gem 'ofn-qz', github: 'openfoodfoundation/ofn-qz', ref: '60da2ae4c44cbb4c8d602f5
 
 group :production, :staging do
   gem 'ddtrace'
+  gem 'unicorn-worker-killer'
 end
 
 group :test, :development do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -411,6 +411,8 @@ GEM
       rspec-core (~> 3.0)
       ruby-progressbar (~> 1.4)
     geocoder (1.1.8)
+    get_process_mem (0.2.5)
+      ffi (~> 1.0)
     gmaps4rails (1.5.6)
     haml (4.0.7)
       tilt
@@ -668,6 +670,9 @@ GEM
     unicorn-rails (2.2.1)
       rack
       unicorn
+    unicorn-worker-killer (0.4.4)
+      get_process_mem (~> 0)
+      unicorn (>= 4, < 6)
     uuidtools (2.1.5)
     warden (1.2.7)
       rack (>= 1.0)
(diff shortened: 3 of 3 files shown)


sauloperez:
yep, I think so too. It's time we give unicorn killer a try. Now it's evident the app doesn't play with memory and with reports that's no suprise.

config.ru
```diff
@@ -1,4 +1,13 @@
 # This file is used by Rack-based servers to start the application.
 
+if ENV.fetch("KILL_UNICORNS", false) && (Rails.env.production? || Rails.env.staging?)
+  # Gracefully restart individual unicorn workers if they have:
+  # - performed between 10000 and 15000 requests
+  # - grown in memory usage to between 500 and 600 MB
+  require 'unicorn/worker_killer'
+  use Unicorn::WorkerKiller::MaxRequests, 10_000, 15_000
+  use Unicorn::WorkerKiller::Oom, (500 * (1024**2)), (600 * (1024**2))
```
luisramos0, line 9: how many requests do we have per day? We should base this config on that number too avoid too many restarts.

I wonder if the mem values are not too low and will trigger too many restarts. On your screenshot the other two processes are taking 700MB and 555MB right? Maybe 1GB would be better?

Also, what's the upper limit for? if the worker is taking more than the max 600MB will it not be restarted?

luisramos0: after this is live, is there a way to monitor how many restarts were triggered by these rules?

luisramos0: The actual limit is decided by rand() between memory_limit_min and memory_limit_max per worker, to prevent all workers to be dead at the same time.

Matt-Yorkley: > to prevent all workers to be dead at the same time

Exactly. There should only be one restarted at a time, so the others are still available. And it only does the restart whilst a worker is _between_ requests, not busy.

> how many requests do we have per day?

I took a quick look at UK Prod and got ~64k for today. So I guess with high usage, it would be roughly every 2 hours. Maybe we can go higher for both of these settings...

Matt-Yorkley: I've updated the values. Also note this will not be used unless the instance explicitly enables it by setting `ENV['KILL_UNICORNS']`.

Matt-Yorkley: Some interesting stats form French Prod for comparison:

Screenshot from 2020-04-04 22-47-25

Here we can see:
- the workers have all been running for ~256 hours
- two of them are sitting nicely at 590 and 620 MB, which looks reasonable
- two of them are eating 1.5 and 1.7 GB, which really doesn't look great

I think if we had `unicorn-worker-killer` running, French Prod would have roughly an extra 2GB of free memory right now. Likwise, UK Prod wouldn't be deep in swap today.

Matt-Yorkley: These numbers from `ps` are actually in kilobytes, so they're fractionally lower than advertised: 
1711476 / 1024 = **1671 MB**
1494168 / 1024 = **1459 MB**
620824 / 1024 = **606 MB**
594400 / 1024 = **580 MB**

I think if a worker goes over 700MB, it's time for a shift change.

Maybe we can add some ENV vars to make these upper and lower bounds configurable though...

Matt-Yorkley: More stats; last night on UK Prod we went up to 16GB of RAM. This morning our four workers look like this:

Screenshot from 2020-04-05 09-50-39

That's 3 sitting nicely at 470-520 MB, and one at **~6GB**...

Matt-Yorkley: And finally, some terrifying stats from Aus. This is the memory graph for the last 2 days (it was fairly flat earlier in the week):

Screenshot from 2020-04-05 10-11-58

And the 3 unicorn workers currently:

Screenshot from 2020-04-05 10-12-56

That's 3 workers using between **3.1 and 3.4 GB** each...

Matt-Yorkley: @mkllnk

luisramos0:
lets wait for @mkllnk 's review :+1:

Matt-Yorkley:
I adjusted the conditional in `config.ru`; we can't use `Rails.env.production?` because `Rails` isn't initialized yet (the app hasn't been launched at this point).

luisramos0:
-- When a worker is restarted there's a log entry
great!

shall we merge or another dev should test?

Matt-Yorkley:
I'm testing now and working on an `ofn-install` PR. Making sure the ENV vars are actually available at the point they're needed is proving tricky.

Matt-Yorkley:
> With the current patterns, I would allow up to 2 GB per worker to avoid frequent restarts. That would ensure that our three workers are not allowed to occupy more than half of the memory.

Not all instances have a spare 6 GB of memory sitting around, some have less than that in total. If multiple workers can run for days under heavy usage on our biggest servers without going over ~650 MB, I think refreshing the ones that reach 700-850 MB is pretty reasonable.

sauloperez:
> I'm testing now and working on an ofn-install PR. Making sure the ENV vars are actually available at the point they're needed is proving tricky.

But `/etc/default/openfoodnetwork` already contains the RAILS_ENV and some others. That should be enough, right?

sauloperez:
> It's still one particular report that is run once a month that occupies all the memory without freeing it. If we clean after that, we should be pretty good for the rest of the month.

I think this is a dangerous assumption I understand we might survive a little longer this way but that means that the infrastructure depends on a particular usage pattern to provide a reliable service. What if new users follow a different approach? having all services in a single VPS exacerbates this problem and its consequences.

Matt-Yorkley:
> But `/etc/default/openfoodnetwork` already contains the RAILS_ENV and some others. That should be enough, right?

See comment here: https://github.com/openfoodfoundation/ofn-install/pull/589#discussion_r403999562

Matt-Yorkley:
Now in production in UK and France. 

Here's a nice example of a single unicorn worker going berserk around 19:45, before being gracefully and automatically restarted by `unicorn-worker-killer` at 19:49:

Screenshot from 2020-04-06 19-52-38

This was France's memory today (unfortunately before deploying the `unicorn-worker-killer` patch):

Screenshot from 2020-04-06 20-00-47

Two unicorn workers go nuts and their memory use grows without stopping until around 16:00, when the server runs out of memory, the two rogue workers crash completely, and we have some downtime.

luisramos0:
great :clap:

sauloperez:
great report and effort @Matt-Yorkley :clap:

> the two rogue workers crash completely, and we have some downtime.

That's a perfect argument. Better to kill a worker at a time than having downtime. And the more workers, the less impact that has.

PR #5169, opened 2020-04-06
make shop name a link on /account

### LINKED ISSUE #5102: My Account: make shop names links to open Shop
## Description
- As a Customer
- on page https://openfoodnetwork.org.au/account#/orders
- I can easily navigate to My Shops (the ones I have ordered through

This way the My Account / Orders page can become a useful way to navigate to my 'favourite' shops, which otherwise I need to know the links for, or go through the Search. This is particularly useful for Hubs / Shops that are not visible as only accessible to members etc - it is too easy for them to navigate away and have no way of getting back

It would be good to do the same thing on the Transactions page, which is formatted a bit more nicely already
https://openfoodnetwork.org.au/account#/transactions

From the Customer accounts pages Orders and Transactions, Customers can click on name of Shop/Hub and be taken to that Shop

jeduardo824:
#### What? Why?

Closes #5102 

Under /account all the past and open orders are listed with their respective Shops. This PR makes the shop name a link to the shop page. I just changed the shop name element to have a href with /shop_permalink/shop.

#### What should we test?
On /account, we should test if a click on the shop name, inside the past/open orders table, redirects us to the shop page.

#### Release notes

Make the shop name a link to the shop page on the account page.

Changelog Category: Changed

### DIFF
--- a/app/views/spree/users/_open_orders.html.haml
+++ b/app/views/spree/users/_open_orders.html.haml
@@ -13,7 +13,8 @@
         %tr.order-row
           %td.order1
             %a{"ng-href" => "{{::order.path}}", "ng-bind" => "::order.number"}
-          %td.order2{"ng-bind" => "::Orders.shopsByID[order.shop_id].name"}
+          %td.order2
+            %a{"ng-href" => "{{::Orders.shopsByID[order.shop_id].hash}}#{main_app.shop_path}", "ng-bind" => "::Orders.shopsByID[order.shop_id].name"}
           %td.order3.show-for-large-up{"ng-bind" => "::order.changes_allowed_until"}
           %td.order4.show-for-large-up{"ng-bind" => "::order.item_count"}
           %td.order5.text-right{"ng-class" => "{'credit' : order.total < 0, 'debit' : order.total > 0, 'paid' : order.total == 0}","ng-bind" => "::order.total | localizeCurrency"}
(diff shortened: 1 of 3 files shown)


app/views/spree/users/_open_orders.html.haml
```diff
@@ -13,7 +13,8 @@
         %tr.order-row
           %td.order1
             %a{"ng-href" => "{{::order.path}}", "ng-bind" => "::order.number"}
-          %td.order2{"ng-bind" => "::Orders.shopsByID[order.shop_id].name"}
+          %td.order2
+            %a{"ng-href" => "{{::Orders.shopsByID[order.shop_id].hash}}/shop", "ng-bind" => "::Orders.shopsByID[order.shop_id].name"}
```
luisramos0, line 17: The URL should be built using the rails routes helpers.
This is angular code so you cannot do:
`main_app.enterprise_shop_path(order.distributor)`
but I think it would be better if you use shop_path instead so:

`"ng-href" => "{{::Orders.shopsByID[order.shop_id].hash}}#{main_app.shop_path}"`

jeduardo824: I considered that too, but I did not know the right way to do that I angular. Thanks for clarifying.

luisramos0: yeah, you just have to know what is ruby and what is angular and remember ruby runs on the server (the #{} part) and then angular (the {{}} part) runs on the browser. Sometimes this creates confusing code like here. I think this is the best we can do here now :+1:

filipefurtad0:
Hey @jeduardo824,

Writing on behalf of the actual tester for this PR Daniel @Ekow244,

Daniel verified that the Shops on /account#/orders now show hyperlinks directing to the respective shop:

Before this PR:
image


After this PR:
image

Thank you Eduardo and Daniel.
Good to go!

jeduardo824:
Awesome!

PR #5174, opened 2020-04-06
Allow changing the connection pool size

sauloperez:
#### What? Why?

This allows us to tune the pool size for UK. The hypothesis from @kristinalim is:

> From what I understand, it can result to Rails processes waiting for each other to complete, while the DB server can take more simultaneous connections.

#### What should we test?
We'll do it in production.

#### Release notes

Allow specifying the size of the database connection pool from an env var
Changelog Category: Added

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   adapter: postgresql
   encoding: unicode
-  pool: 5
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
I think this file gets ignored in production. There's not much value in adding it for development? See Kristina's PR: https://github.com/openfoodfoundation/ofn-install/issues/591

PR #5175, opened 2020-04-06
Memoize ScopeProductToHub in product list

Matt-Yorkley:
#### What? Why?

Memoizing this scoper means we avoid fetching all of the hub's variant overrides every time we scope a product (which happens in it's initializer). This happens for every product loaded when displaying a shops's product list, which is currently one of our **top 3 pain points** in Datadog :fire:.

#### What should we test?

A quick check to make sure product lists are loading in the shop page should be enough :+1:

#### Release notes

Improved performance on product list endpoint

Changelog Category: Changed

### DIFF
--- a/app/services/products_renderer.rb
+++ b/app/services/products_renderer.rb
@@ -40,7 +40,7 @@ class ProductsRenderer
   end
 
   def product_scoper
-    OpenFoodNetwork::ScopeProductToHub.new(distributor)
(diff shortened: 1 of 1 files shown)


filipefurtad0:
Hi @Matt-Yorkley ,

Checked the basic functionality of the product list, in the shop front, tested filters as well - found nothing unusual. Seems good to go.

Moving it down the pipe.

PR #5178, opened 2020-04-07
Dont populate empty cart

sauloperez:
#### What? Why?

I noticed in development that the server received a POST /cart/populate even though my cart was empty and the page was just loading. I couldn't add any items yet.

When loading the page $watchGroup calls the listener function for every listed line item but with a set variant and null quantity and max_quantity. There's no point on computing an order change when there was none.

This saves an empty request on the second most used endpoint of the app, especially busy when users are placing orders.

#### What should we test?

We should check we still can add and remove items from the cart and come back and forth from the /shops page without any browser alert. We should then check we can place the order as usual.

#### Release notes

Do not send populate the cart on page load with no user request.
Changelog Category: Changed

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/shop_variant.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/shop_variant.js.coffee
@@ -5,5 +5,9 @@ Darkswarm.directive "shopVariant", ->
   scope:
     variant: '='
   controller: ($scope, Cart) ->
-    $scope.$watchGroup ['variant.line_item.quantity', 'variant.line_item.max_quantity'], ->
+    $scope.$watchGroup [
+      'variant.line_item.quantity',
+      'variant.line_item.max_quantity'
+    ], (new_value, old_value) ->
+      return if old_value[0] == null && new_value[0] == null
       Cart.adjust($scope.variant.line_item)
--- a/spec/javascripts/unit/darkswarm/services/cart_spec.js.coffee
+++ b/spec/javascripts/unit/darkswarm/services/cart_spec.js.coffee
@@ -45,6 +45,13 @@ describe 'Cart service', ->
     Cart.adjust(order.line_items[0])
     expect(Cart.line_items.length).toEqual 0
 
+  it "does not add an item in the cart without quantity", ->
+    Cart.line_items = []
+
+    spyOn(Cart, 'orderChanged')
+    order.line_items[0].max_quantity = 0
+    expect(Cart.orderChanged).not.toHaveBeenCalled()
+
   it "sums the quantity of each line item for cart total", ->
     order.line_items[0].quantity = 2
     expect(Cart.total_item_count()).toEqual 2


Matt-Yorkley:
I'm wondering if this relates to the case where the user has items in a cart with one shop, then navigates to a different shop. In that case we probably do need to reset/empty their current cart...?

sauloperez:
that's an interesting point but I assume this handled already or someone we have complained by now.

app/assets/javascripts/darkswarm/directives/shop_variant.js.coffee
```diff
@@ -6,4 +6,4 @@ Darkswarm.directive "shopVariant", ->
     variant: '='
   controller: ($scope, Cart) ->
     $scope.$watchGroup ['variant.line_item.quantity', 'variant.line_item.max_quantity'], ->
-      Cart.adjust($scope.variant.line_item)
+      Cart.adjust($scope.variant.line_item) if $scope.variant.line_item.quantity > 0
```
luisramos0, line 9: I think we need to send quantity = 0 to remove the line_item :warning: 
have you tested that adding a product and then reducing quantity to zero again will work?

sauloperez: yeah, right. No, I haven't gone into much detail while testing. I'll do that next.

app/assets/javascripts/darkswarm/directives/shop_variant.js.coffee
```diff
@@ -5,5 +5,9 @@ Darkswarm.directive "shopVariant", ->
   scope:
     variant: '='
   controller: ($scope, Cart) ->
-    $scope.$watchGroup ['variant.line_item.quantity', 'variant.line_item.max_quantity'], ->
+    $scope.$watchGroup [
+      'variant.line_item.quantity',
+      'variant.line_item.max_quantity'
+    ], (new_value, old_value) ->
+      return if old_value[0] == null && new_value[0] == null
```
sauloperez, line 12: this is how we can know it's the first page load and nothing was selected.

Matt-Yorkley:
I did some manual testing of this and it looks good, no problems. I think the issue with the Karma specs is that the fix is in the directive, (which I think is correct) but your spec changes are in the service, and they don't quite match up. Apparently we don't test directives...?

sauloperez:
> I'm not sure the best way to test it. Maybe we could add a feature test that checks that the cart populate endpoint doesn't get called when the shop page is loaded and the form hasn't been touched yet?

yeah, that sounds enough.

spec/javascripts/unit/darkswarm/services/cart_spec.js.coffee
```diff
@@ -45,6 +45,13 @@ describe 'Cart service', ->
     Cart.adjust(order.line_items[0])
     expect(Cart.line_items.length).toEqual 0
 
+  it "does not add an item in the cart without quantity", ->
+    Cart.line_items = []
+
+    spyOn(Cart, 'orderChanged')
+    order.line_items[0].max_quantity = 0
+    expect(Cart.orderChanged).not.toHaveBeenCalled()
```
Matt-Yorkley, line 53: I think this test is enough here actually :+1:

spec/javascripts/unit/darkswarm/services/cart_spec.js.coffee
```diff
@@ -77,7 +84,7 @@ describe 'Cart service', ->
       expect(Cart.update_enqueued).toBe(true)
 
   describe "updating the cart", ->
-    data = {variants: {}}
+    data = {variants: 1: {quantity: 1, max_quantity: null}}
```
Matt-Yorkley, line 87: If we revert this change the specs all pass. This is messing with tested calls to `Cart.update()`, but the code you've changed is only affecting `Cart.adjust()`, so I think it's correct to revert this bit.

Matt-Yorkley: @sauloperez

sauloperez: yep, it was just leftovers from my dirty iterations

sauloperez: Ok, so I'll remove this plus the meaningless change on cart.js.coffee and publish the PR

sauloperez:
Thanks for your help @Matt-Yorkley !

filipefurtad0:
Hey @sauloperez ,

Checked the usual /cart operations:
- add and remove items from the cart
- went back and forth from the /shops page
- place the order as usual.

All good.

Made sure the POST /cart/populate request is now gone.

After staging this PR:
5178_after_PR

On a different staging-server (i.e., without staging this PR):
5178_before_PR

Ready to go!

PR #5183, opened 2020-04-07
Add Fragment Caching on shops and producers pages

pacodelaluna:
#### What? Why?

Let's add some Fragment Caching on these 2 pages as they are very long to be generated.

Related to #5143 

From DataDog, we can see that Postgres requests are taking sometimes:
Screenshot from 2020-04-07 14-51-34

I am not seeing any dynamic code on the fragment I am caching, most of the logic is done on front end side with Angular.
Let's consider this as a patch in order to give some air to our servers.

#### What should we test?
We should test that /producers and /shops pages are working as expected, for logged user and for unlogged user.

Changelog Category: Changed

### DIFF
--- a/app/views/producers/index.html.haml
+++ b/app/views/producers/index.html.haml
@@ -1,32 +1,32 @@
-- cache 'cached_producers_page_content', expires_in: 30.minutes do
-  - content_for(:title) do
-    = t :producers_title
+- content_for(:title) do
+  = t :producers_title
 
-  - content_for :injection_data do
+- content_for :injection_data do
+  - cache 'cached_producers_page_content', expires_in: 5.minutes do
     = inject_enterprises(@enterprises)
 
-  .producers{"ng-controller" => "EnterprisesCtrl", "ng-cloak" => true}
-    .row
-      .small-12.columns.pad-top
-        %h1
-          = t :producers_headline
+.producers{"ng-controller" => "EnterprisesCtrl", "ng-cloak" => true}
+  .row
+    .small-12.columns.pad-top
+      %h1
+        = t :producers_headline
 
-    = render partial: "shared/components/enterprise_search"
-    = render partial: "producers/filters"
+  = render partial: "shared/components/enterprise_search"
+  = render partial: "producers/filters"
 
-    .row
-      .small-12.columns
-        .active_table
-          %producer.active_table_node.row.animate-repeat{id: "{{producer.path}}",
-          "ng-repeat" => "producer in filteredEnterprises = (Enterprises.producers | searchEnterprises:query | taxons:activeTaxons | properties:activeProperties:'supplied_properties')",
-          "ng-controller" => "ProducerNodeCtrl",
-          "ng-class" => "{'closed' : !open(), 'open' : open(), 'inactive' : !producer.active}",
-          id: "{{producer.hash}}"}
+  .row
+    .small-12.columns
+      .active_table
+        %producer.active_table_node.row.animate-repeat{id: "{{producer.path}}",
+        "ng-repeat" => "producer in filteredEnterprises = (Enterprises.producers | searchEnterprises:query | taxons:activeTaxons | properties:activeProperties:'supplied_properties')",
+        "ng-controller" => "ProducerNodeCtrl",
+        "ng-class" => "{'closed' : !open(), 'open' : open(), 'inactive' : !producer.active}",
+        id: "{{producer.hash}}"}
 
-            .small-12.columns
-              = render partial: 'producers/skinny'
-              = render partial: 'producers/fat'
+          .small-12.columns
+            = render partial: 'producers/skinny'
+            = render partial: 'producers/fat'
 
-          = render partial: 'shared/components/enterprise_no_results'
+        = render partial: 'shared/components/enterprise_no_results'
 
-  = render partial: "shared/footer"
+= render partial: "shared/footer"
--- a/app/views/producers/index.html.haml
+++ b/app/views/producers/index.html.haml
@@ -1,31 +1,32 @@
-- content_for(:title) do
-  = t :producers_title
+- cache 'cached_producers_page_content', expires_in: 30.minutes do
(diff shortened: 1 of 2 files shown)


pacodelaluna:
It is just an idea, maybe I am missing something.

app/views/producers/index.html.haml
```diff
@@ -1,31 +1,32 @@
-- content_for(:title) do
-  = t :producers_title
+- cache 'cached_producers_page_content', expires_in: 30.minutes do
```
pacodelaluna, line 1: I put 30 minutes but we can adjust it for sure.

pacodelaluna: I was not that inspired for the cache key ever... Well, let's validate the approach first.

sauloperez: Take a look at https://app.datadoghq.com/logs/analytics?agg_q=&cols=core_host%2Ccore_service%2Clog_response_time&from_ts=1586152800000&index=main&live=false&messageDisplay=inline&query=%40http.url_details.path%3A%28%22%2Fshops%22%20OR%20%22%2Fshops%2F%22%29%20host%3Awww.openfoodnetwork.org.uk&step=300000&stream_sort=desc&to_ts=1586188020000&top_n=5&viz=timeseries. If we go with 5min we would save around 10reqs every 5min. At 13:30h we would have performed just 1 out of 38 requests.

sauloperez:
The concerns I have are any possible dynamics bits like multilanguage? What if I request the page now in Catalan?

pacodelaluna:
@sauloperez I have moved the cache to only seize the JSON serialized objects. No need to update caceh keys with locales then.

I have also set the expiration to 5 minutes.

pacodelaluna:
I totally agree that cache should be used carefully, and that the best approach here would be to refactor the enterprise logic, with a sharp API replacing the serialized JSON, and Angular calling it but it is something quite big and we were looking for quick wins to solve our last performance issues.

luisramos0:
awesome! we need validation from product and probably an update of the user guide?
basically, opening and closing OCs can have a delay of up to 5 minutes to be seen on the shops page

pacodelaluna:
I am actually wondering if we will faced such delays, I mean apart if we create an order cycle that needs to be released within 5 min.
Angular should be able to deal with these delays, because actually it is an issue we can have right now. I mean what happens if I am letting the /shops page open the whole day, with some displayed order cycles being closed during this day, I am pretty sure that the front logic will deal with such an issue. Especially that the shop detail will be reach with a separated request.

kirstenalarsen:
Just quick comment that I think 5 min delay in shops page is ok. Not ideal but ok. Most people send users directly to their own shops rather than via this page so if they're letting people know about an oc opening should be fine

More annoying case is that the order cycle appears open for 5 minutes after it is actually closed - with potential clicks through to closed shop. But the first time someone loads the page is fine yes? So it's just if they're reloading the page (which they would do if wanting most up-to-date information) - hmmmm

RachL:
I'm not sure I'm following... Are we caching the list of enterprises or the OC open/close information?

Matt-Yorkley:
I've just realised this could be super annoying in staging and development...

Change some order cycles to check something -> wait 5 minutes before it's actually usable.

pacodelaluna:
Maybe better to use this on Producers page first, where there is no order cycles timings controls. I guess the 5 min gap is much less sensitive on this page. What do you think?

Analyzing the Shops page logic, I am seeing that the closing-in time display is managed through Angular, and it seems there is a special method, `current()`, calculating the message but I am not getting yet where it is done. Will go on tomorrow.

luisramos0:
current() is just to not display the time on the shop the user is currently shopping at.
the time is fetched from hub.orders_close_at which is fetched from the server and thus cached. I dont think it's a problem to have a 5 minutes delay on this time.

I'll start a thread in slack to decide if we go ahead with this or not.

luisramos0:
Because of the stale data annoyance, I think my vote goes for leaving this PR on hold until the next release so that we can see the impact of other /shops PRs.

RachL:
@Matt-Yorkley 

> Change some order cycles to check something -> wait 5 minutes before it's actually usable.

We can go on the enterprise profile and copy-paste the URL. Not ideal, but it works.

luisramos0:
go ahead from Rachel and Lynne, I am moving this to test ready.

luisramos0:
ok, slack conversation https://app.slack.com/client/T02G54U79/CAVTM01QB/1586425625.108300
we are not going forward yet with this one.
I am not sure in what column to leave it. test ready?

pacodelaluna:
> Maybe better to use this on Producers page first, where there is no order cycles controls. I guess the 5 min gap is much less sensitive on this page. What do you think?

What about this comment? 

If performance are good for now, we can drop this, no problem, as I said, it was just to give some air to some cores on production instances (UK and FR ones), you know the temperature increased by 2 degrees in Paris since these spikes ;)

pacodelaluna:
I guess we can close this ticket for now, right?

sauloperez:
I'd like to share a couple of ideas while we wait for this to be back:

* What if we wrap this cache call with an ENV var check so a dev can enable/disable this at will. Would this alleviate some people's concerns @RachL @kirstenalarsen @lin-d-hop ? It could rather be a resilience mechanism we use when we experience chaos.
* Would you mind exploring which associations we would need to `touch: true` if we used `cache(@enterprise)` instead? this way Rails will handle cache expiration for us and we'll have much greater gains (we would serve from cache as much as we can)

daniellemoorhead:
Hey all, with all the caching work @Matt-Yorkley did are we now able to gratefully acknowledge the path this PR took us down and then close it?

Matt-Yorkley:
Sounds good :+1: 

We can see what the new caching changes look like in production over the next week, and we also have a nice foundation now for multiple caching techniques which could be used in other parts of the app.

Thanks @pacodelaluna !

PR #5184, opened 2020-04-07
Inventory loading

### LINKED ISSUE #5181: Admin Inventory performance
From discussions in Slack channel `uk_down_apr6`, we need to address some issues with the inventory page ASAP.

Matt-Yorkley:
#### What? Why?

Closes #5181 

Improves performance on admin inventory page.

#### What should we test?

Generally; we need to test that the functionality on the Admin Inventory page is working correctly and there are no regressions in the basic behavior.

More specifically, we're looking at improving the performance of the additional requests that are triggered when the page loads, which fetch the products that the user is able to add to inventory. If you want to test this bit, and you'll need a staging server with lots of data, and a login that has access to a hub with permissions to add lots of producer's products to their inventory. If you load the page with the network tab open, you'll be able to see the requests as they're loading. They'll look like this:

Screenshot from 2020-04-09 10-36-11

Another hot tip if you really want to get into it: if you click on one of those requests and then click the **Timing** tab, you should see an indication of how long the request took. Here's two examples showing the response time of the request in the bottom right:

With `master`, 6.20 seconds:
Screenshot from 2020-04-09 11-25-30

With this PR, 469 milliseconds:
Screenshot from 2020-04-09 11-25-47

Note: I don't recommend testing this as **superadmin** on `master` branch :warning:

#### Release notes

Improved performance on admin inventory page

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
+++ b/app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
@@ -2,19 +2,24 @@ angular.module("admin.indexUtils").factory "PagedFetcher", (dataFetcher) ->
   new class PagedFetcher
     # Given a URL like http://example.com/foo?page=::page::&per_page=20
     # And the response includes an attribute pages with the number of pages to fetch
-    # Fetch each page async, and call the processData callback with the resulting data
-    fetch: (url, processData, onLastPageComplete) ->
-      dataFetcher(@urlForPage(url, 1)).then (data) =>
-        processData data
+    # Fetch each page async, and call the pageCallback callback with the resulting data
+    # Developer note: this class should not be re-used!
+    page: 1
+    last_page: 1
 
-        if data.pages > 1
-          for page in [2..data.pages]
-            lastPromise = dataFetcher(@urlForPage(url, page)).then (data) ->
-              processData data
-          onLastPageComplete && lastPromise.then onLastPageComplete
-          return
-        else
-          onLastPageComplete && onLastPageComplete()
+    fetch: (url, pageCallback) ->
+      @fetchPages(url, @page, pageCallback)
 
     urlForPage: (url, page) ->
       url.replace("::page::", page)
+
+    fetchPages: (url, page, pageCallback) ->
+      dataFetcher(@urlForPage(url, page)).then (data) =>
+        @page++
+        @last_page = data.pages
+
+        pageCallback(data) if pageCallback
+
+        if @page <= @last_page
+          @fetchPages(url, @page, pageCallback)
+
--- a/app/controllers/api/products_controller.rb
+++ b/app/controllers/api/products_controller.rb
@@ -74,7 +74,7 @@ module Api
 
       @products = paged_products_for_producers producers
 
-      render_paged_products @products
+      render_paged_products @products, ::Api::Admin::ProductSimpleSerializer
     end
 
     # POST /api/products/:product_id/clone
@@ -128,10 +128,10 @@ module Api
         page(params[:page]).per(params[:per_page])
     end
 
-    def render_paged_products(products)
+    def render_paged_products(products, product_serializer = ::Api::Admin::ProductSerializer)
       serializer = ActiveModel::ArraySerializer.new(
         products,
-        each_serializer: ::Api::Admin::ProductSerializer
+        each_serializer: product_serializer
       )
 
       render text: {
--- a/app/controllers/api/products_controller.rb
+++ b/app/controllers/api/products_controller.rb
@@ -69,10 +69,10 @@ module Api
     end
 
     def overridable
-      producers = OpenFoodNetwork::Permissions.new(current_api_user).
-        variant_override_producers.by_name
+      producer_ids = OpenFoodNetwork::Permissions.new(current_api_user).
+        variant_override_producers.by_name.select('enterprises.id')
 
-      @products = paged_products_for_producers producers
+      @products = paged_products_for_producers producer_ids
 
       render_paged_products @products, ::Api::Admin::ProductSimpleSerializer
     end
@@ -118,11 +118,11 @@ module Api
       ]
     end
 
-    def paged_products_for_producers(producers)
+    def paged_products_for_producers(producer_ids)
       Spree::Product.scoped.
         merge(product_scope).
         includes(variants: [:product, :default_price, :stock_items]).
-        where(supplier_id: producers).
+        where(supplier_id: producer_ids).
         by_producer.by_name.
         ransack(params[:q]).result.
         page(params[:page]).per(params[:per_page])
--- a/app/controllers/api/products_controller.rb
+++ b/app/controllers/api/products_controller.rb
@@ -121,6 +121,7 @@ module Api
     def paged_products_for_producers(producers)
       Spree::Product.scoped.
         merge(product_scope).
+        includes(variants: [:product, :default_price, :stock_items]).
         where(supplier_id: producers).
         by_producer.by_name.
         ransack(params[:q]).result.
--- a/app/models/spree/stock/quantifier.rb
+++ b/app/models/spree/stock/quantifier.rb
@@ -1,0 +1,28 @@
+module Spree
+  module Stock
+    class Quantifier
+      attr_reader :stock_items
+
+      def initialize(variant)
+        @variant = variant
+        @stock_items = Spree::StockItem.joins(:stock_location).where(:variant_id => @variant, Spree::StockLocation.table_name =>{ :active => true})
+      end
+
+      def total_on_hand
+        if Spree::Config.track_inventory_levels
+          stock_items.sum(&:count_on_hand)
+        else
+          Float::INFINITY
+        end
+      end
+
+      def backorderable?
+        stock_items.any?(&:backorderable)
+      end
+
+      def can_supply?(required)
+        total_on_hand >= required || backorderable?
+      end
+    end
+  end
+end
--- a/app/models/spree/stock/quantifier.rb
+++ b/app/models/spree/stock/quantifier.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Spree
   module Stock
     class Quantifier
--- a/app/models/spree/stock/quantifier.rb
+++ b/app/models/spree/stock/quantifier.rb
@@ -5,7 +5,7 @@ module Spree
 
       def initialize(variant)
         @variant = variant
-        @stock_items = Spree::StockItem.joins(:stock_location).where(:variant_id => @variant, Spree::StockLocation.table_name =>{ :active => true})
+        @stock_items = fetch_stock_items
       end
 
       def total_on_hand
@@ -22,6 +22,16 @@ module Spree
 
       def can_supply?(required)
         total_on_hand >= required || backorderable?
+      end
+
+      private
+
+      def fetch_stock_items
+        # Don't re-fetch associated stock items from the DB if we've already eager-loaded them
+        return @variant.stock_items.to_a if @variant.stock_items.loaded?
+
+        Spree::StockItem.joins(:stock_location).
+          where(:variant_id => @variant, Spree::StockLocation.table_name => { active: true })
       end
     end
   end
--- a/app/models/spree/stock/quantifier.rb
+++ b/app/models/spree/stock/quantifier.rb
@@ -11,11 +11,7 @@ module Spree
       end
 
       def total_on_hand
-        if Spree::Config.track_inventory_levels
-          stock_items.sum(&:count_on_hand)
-        else
-          Float::INFINITY
-        end
+        stock_items.sum(&:count_on_hand)
       end
 
       def backorderable?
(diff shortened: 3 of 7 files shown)


Matt-Yorkley:
I've done a quick round of performance improvements here, the results are good so far. Looking at the overridable products endpoint that is called when this page loads:

**BEFORE:**
- total page load is around **12 seconds**
- the endpoint makes **1100+ queries** (this can be hit many times in one page load)
- the endpoint takes **~6 seconds** to load

Screenshot from 2020-04-07 15-24-50

**AFTER:**
- total page load is around **1.2 seconds** (~90% less)
- the endpoint makes **23 queries** (~98% less)
- the endpoint takes **~0.6 seconds** to load (~90% less)

Screenshot from 2020-04-07 15-22-16

Matt-Yorkley:
The next issue is that the page spams the endpoint with many requests at once, for example with 8 pages it makes 8 requests all at once. Now that the performance is 10 times better, I'm wondering if we can either:
- just have one large request that returns all overridable products
- make sure the requests wait until the last one has loaded before making the next request
- just render all the overridable products using AMS on page load

Matt-Yorkley:
I just managed to use the page as superadmin with production data! :open_mouth: :tada:

sauloperez:
> Or maybe we just go with this for now and see how it looks?

we need to deliver as many improvements as we can. I'd say, if this works, let's get it in and see if it can get better, but at least we've got this.

app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
```diff
@@ -2,19 +2,26 @@ angular.module("admin.indexUtils").factory "PagedFetcher", (dataFetcher) ->
   new class PagedFetcher
     # Given a URL like http://example.com/foo?page=::page::&per_page=20
     # And the response includes an attribute pages with the number of pages to fetch
-    # Fetch each page async, and call the processData callback with the resulting data
-    fetch: (url, processData, onLastPageComplete) ->
-      dataFetcher(@urlForPage(url, 1)).then (data) =>
-        processData data
+    # Fetch each page async, and call the pageCallback callback with the resulting data
+    # Developer note: this class should not be re-used!
```
sauloperez, line 6: :joy:

app/controllers/api/products_controller.rb
```diff
@@ -121,16 +121,17 @@ def product_query_includes
     def paged_products_for_producers(producers)
       Spree::Product.scoped.
         merge(product_scope).
+        includes(variants: [:product, :default_price, :stock_items]).
```
sauloperez, line 124: so it's this what saves those many queries?

RachL:
> I just managed to use the page as superadmin with production data! 😮 🎉

GET OUT 😮 😄

app/models/spree/stock/quantifier.rb
```diff
@@ -0,0 +1,40 @@
...
+      def fetch_stock_items
+        # Don't re-fetch associated stock items from the DB if we've already eager-loaded them
+        return @variant.stock_items.to_a if @variant.stock_items.loaded?
+
+        Spree::StockItem.joins(:stock_location).
+          where(:variant_id => @variant, Spree::StockLocation.table_name => { active: true })
+      end
+    end
+  end
+end
```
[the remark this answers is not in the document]
Matt-Yorkley: I kept the commit history clear, the original Spree version of the class is in one commit, and the changes are in the next

app/controllers/api/products_controller.rb
```diff
@@ -70,11 +70,11 @@ def bulk_products
 
     def overridable
       producers = OpenFoodNetwork::Permissions.new(current_api_user).
-        variant_override_producers.by_name
+        variant_override_producers.by_name.select('enterprises.id')
 
       @products = paged_products_for_producers producers
 
-      render_paged_products @products
+      render_paged_products @products, ::Api::Admin::OverridableProductSerializer
```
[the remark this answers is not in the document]
Matt-Yorkley: Drastically reducing both the volume of returned data and number of queries triggered to fetch it, yes.

app/controllers/api/products_controller.rb
Matt-Yorkley: There were some huge N+1s here, yes.

luisramos0:
this needs some careful review, I can do it in 8hrs from now.

app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
Matt-Yorkley: It felt  like it would have been irresponsible to touch this file without leaving this comment :see_no_evil:

app/serializers/api/admin/overridable_product_serializer.rb
```diff
@@ -0,0 +1,16 @@
+class Api::Admin::OverridableProductSerializer < ActiveModel::Serializer
+  attributes :id, :name
```
luisramos0: are you sure none of the other attributes are needed in the inventory page? like :sku or :variant_unit?
just asking.

app/serializers/api/admin/overridable_variant_serializer.rb
```diff
@@ -0,0 +1,27 @@
+class Api::Admin::OverridableVariantSerializer < ActiveModel::Serializer
```
luisramos0: We should describe our code by what it does, not by its uses.
I think this rule applies here really well.

Although this is used in the overridable action, this has nothing specific to it.
I'd call this Api::Admin::VariantSimpleSerializer (or Api::Admin::VariantBasicSerializer)
Because if you do that then, on top of making ready for reuse in the future, you could make the original serializer Api::Admin::VariantSerializer extend this Api::Admin::SimpleVariantSerializer and that would avoid the copy/paste of all the methods and attributes.


This is also applicable to Api::Admin::OverridableProductSerializer: Api::Admin::ProductSimpleSerializer although I am not sure the serializer that extends the basic one could redefine has_many :variants to point to the extended serializer. We would need to test.

I think this is important to manage the natural growth and complexity of our serializers.
If you don't disagree, can I ask you to change the names of the serializers at least? Thanks!

app/controllers/api/products_controller.rb
```diff
@@ -70,7 +70,7 @@ def bulk_products
 
     def overridable
       producers = OpenFoodNetwork::Permissions.new(current_api_user).
-        variant_override_producers.by_name
+        variant_override_producers.by_name.select('enterprises.id')
```
luisramos0, line 73: the variable and parameter name should be changed to producer_ids for correctness.

app/models/spree/stock/quantifier.rb
```diff
@@ -0,0 +1,28 @@
...
+        @stock_items = Spree::StockItem.joins(:stock_location).where(:variant_id => @variant, Spree::StockLocation.table_name =>{ :active => true})
+      end
+
+      def total_on_hand
+        if Spree::Config.track_inventory_levels
+          stock_items.sum(&:count_on_hand)
+        else
+          Float::INFINITY
+        end
+      end
```
luisramos0, line 19: track_inventory_levels is always true in OFN
so this should be:
```
  def total_on_hand
      stock_items.sum(&:count_on_hand)
   end
```

app/models/spree/stock/quantifier.rb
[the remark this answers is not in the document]
luisramos0: I just had a big trip into the spree 2 upgrade times where this class was at the core of so many headaches (managing stock in stock items for normal variants but without stock items for overridded variants).

here are some memories:
FYI we have a few places where this quantifier is avoided since spree 2:
    - Line_item#sufficient_stock? - overridden in OFN to redirect to Variant#can_supply that uses Variant#total_on_hand
    - Stock::AvailabilityValidator#validate - overridden in OFN (for a couple of other reasons including) to redirect to Variant#can_supply? that uses Variant#total_on_hand
    - Variant#in_stock?  - overridden in OFN to redirect to Variant#can_supply? that uses Variant#total_on_hand
    - **Variant#total_on_hand - this is the only place where Quantifier is used** and only for not-overrided variants (variant override does a different thing without StockItems).

app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
Matt-Yorkley: This `PagedFetcher` is only used in one place, in inventory. When we either switch to the network model or spend time implementing pagination here we can get rid of this class, but until then I don't think it should be used elsewhere.

app/serializers/api/admin/overridable_variant_serializer.rb
Matt-Yorkley: I was just thinking about the original cause of this particular part of the issue; that a serializer built for one job that needs a very large set of data is re-used for another job where it only needs a small set of data, because the class name just suggests that it should be used for serializing products.

But yes, I do agree. I'll change the naming as suggested :+1:

Matt-Yorkley:
Thanks @luisramos0! :heart: I'll make those changes in the morning.

> the user will have to wait 50secs until they can use the page? 

I actually found the total time was not so different. I see it locally but it's clear in production as well: the server can't handle being hit with all the requests at once, processes them very slowly and basically in a linear way, but under a huge amount of stress. The endpoint is _incredibly_ slow and CPU intensive (in master) and can take a _very_ long time for the server to actually return all those requests, even when they are all sent at once. 

> Is the user aware that the data is not completely loaded? what happens if the user submits the form when the later pages are still loading, will it work?

This is exactly the problem we have already with this page (in master); the user is not aware it is still loading data, and if they submit things it doesn't necessarily work. It's the cause of: https://github.com/openfoodfoundation/openfoodnetwork/issues/5131

Improving the response time from 6 seconds to 0.6 seconds will help :+1:

Doing a big pagination overhaul here would be great, but I don't think it will be a small job.

app/assets/javascripts/admin/index_utils/services/paged_fetcher.js.coffee
```diff
@@ -2,19 +2,26 @@ angular.module("admin.indexUtils").factory "PagedFetcher", (dataFetcher) ->
...
+      dataFetcher(@urlForPage(url, page)).then (data) =>
+        @page++
+        @last_page = data.pages
+
+        pageCallback && pageCallback data
+
+        if @page <= @last_page
+          @fetchPages(url, @page, pageCallback, lastPageCallback)
+        else
+          lastPageCallback && lastPageCallback()
```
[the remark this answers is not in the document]
luisramos0: after I read your last comment I thought that, if the user cant save the data before all pages are loaded, maybe we should actually use this lastPageCallback to give the user some form of feedback of when the form is ready to be used!?

app/serializers/api/admin/overridable_variant_serializer.rb
Matt-Yorkley: Done.

app/serializers/api/admin/overridable_product_serializer.rb
Matt-Yorkley: I've just double-checked and we don't need these attributes on products here, as far as I can tell. This page is all about variants.

Matt-Yorkley: I've added an extra commit to remove another two unused methods from the `ProductSimpleSerializer`, it's now slimmer.

Matt-Yorkley:
This PR certainly doesn't make inventory perfect, it just aims at making it 10x less awful as quickly as possible.

Matt-Yorkley:
> what do you think about my idea of using the lastPageComplete callback?
I think it could improve UX a lot for slower loads. Could we create an issue to implement it or do you think this PR will be enough?

We can see how things look after this PR. The issue with data not finishing loading is pre-existing. It's possible it will be good enough after this, or maybe we will need to add some additional UI feedback, but it'll definitely need careful inception, dev, and testing if we go for that.

Matt-Yorkley:
And here's the volume of json data returned per request (same data, different branches).

**BEFORE:**
Screenshot from 2020-04-09 10-47-00 (copy)

**AFTER:**
Screenshot from 2020-04-09 10-47-00

luisramos0:
ok, awesome!

filipefurtad0:
Hi @Matt-Yorkley ,

Whoa... Impressive numbers below!
Page load times, from staging-FR, Fredo's Farm Hub 1311 products.
(left, master /// right, this PR)

image

Also tested the basic functionality of inventory, product review, add or hide, stock override functions. These are also visible in the shopfront, for a regular user.

Moving to ready to go :tada:

luisramos0:
40secs to 0.4 secs :rofl: :rocket:

Matt-Yorkley:
I'm not sure if the timing numbers on the right are adjusted for the waiting time of all the (badly queued) requests or something...?

Anyway, at the bottom we have: 

Finish: **1:06 min** -> Finish: **12.44 s**

Fuck yeah.

Thanks again for the awesome testing @filipefurtad0!  :heart:

app/models/spree/stock/quantifier.rb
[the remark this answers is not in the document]
sauloperez: do we have that know-how written down somewhere @luisramos0 ? If I had to find out all that it'd took quite a lot of time :see_no_evil:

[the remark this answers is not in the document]
luisramos0: this is an analysis of all calls to Quantifer. It's basically saying that during spree upgrade most methods that used this Quantifier class were overriden in OFN.
Bye bye spree will just remove all this complexity because there will only be one call to Quantifier :+1:

PR #5186, opened 2020-04-07
Take sku overrides into account in customer totals report

### LINKED ISSUE #5185: Inventory SKUs are not reflected in customer totals report
## Description
- When variant SKUs are overridden in the inventory view, these SKUs do not flow through 
- A hub that is stocking products for a wide array of suppliers (eg BBFH) is likely to have their own SKU system that they will want to use, as opposed to those used by suppliers. Functionality allowing SKUs to be overridden is already in place but there is just a bit a gap here in terms of where that information is shown.

## Expected Behaviour
- SKUs overridden in the Inventory are visible in the 'Customer Totals'  Orders and Fulfillments report.

## Actual Behaviour
- Only product 

## Steps to Reproduce
1. Create a variant of a product
2. Add variant to inventory for a hub
3. Override the variant's sku in the inventory
4. Open an order cycle which includes the variant
5. Place an order for the variant
6. Open the Customer Totals report. The SKU for the variant will not display the overridden value.

## Animated Gif/Screenshot
Difficult to screenshot

## Workaround
N/A

## Severity
bug-s5?

## Your Environment
Aus production, Chrome and Firefox

## Possible Fix
Scope variants to the relevant distributor before rendering the sku.

oeoeaio:
#### What? Why?

Closes #5185 

#### What should we test?
- SKUs of variants overridden in the inventory are visible in the Customer Totals orders and fulfillments report.

#### Release notes
- SKUs set in the product inventory view will now be displayed correctly in the Order Cycle Customer Totals report. 

Changelog Category: Fixed


sauloperez:
I might be seeing ghosts. @oeoeaio is that you??

Matt-Yorkley:
I had to reload the page to make sure Github wasn't broken! :sweat_smile:

lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
```diff
@@ -206,6 +211,11 @@ def shipping_method(line_items)
         shipping_rate = shipping_rates.find(&:selected) || shipping_rates.first
         shipping_rate.try(:shipping_method)
       end
+
+      def scoper_for(distributor_id)
+        @scopers_by_distributor_id[distributor_id] ||=
+          OpenFoodNetwork::ScopeVariantToHub.new(distributor_id)
```
Matt-Yorkley, line 217: I'm slightly worried about the potential impact of this, given that `ScopeVariantToHub` fetches all of the hub's overrides in it's initializer, in the context of large hubs with a lot of overrides (France has a few) that are already struggling to load reports...

oeoeaio: Fair enough. I'm pretty rusty and lots has changed (nice work!), any suggestions or examples of a better approach to this kind of problem?

Matt-Yorkley: It might be OK. I guess we can just test it in a staging server with lots of data? 

I don't think there's an easy alternative for displaying the alternate SKUs. 

I'm thinking this change might also need some product team input...?

oeoeaio: Righto, who do I ping? Some acceptance criteria that I can work towards would be really helpful.

Matt-Yorkley: Maybe @RachL and @lin-d-hop could weigh in on it...

We're changing the SKUs displayed in the customer totals report to show the SKUs from variant overrides instead of the original variant SKU. It could be a big surprise for some users. Maybe it would just need an announcement note in the release?

lin-d-hop: I do agree that it is the technically correct behaviour. But also concerned of the consequences.
SKU is rarely used in the UK so my preference would be this fix.

Would like to hear @RachL's thoughts.

The biggest concern would be performance implications. @Matt-Yorkley can you help us to be as certain as possible this won't increase any server explosion rates?

RachL: hmmmm I have no idea if people use SKU on my instance... but if they do I think the change is logical. So I guess we are good to go 👍

luisramos0:
Hello @oeoeaio :wave: 
Moving to In Dev as it is a draft PR, we can move to Code Review once it's not a draft PR.

oeoeaio:
Hey @luisramos0 👋, and thanks, I've marked it as ready for review.

Matt-Yorkley:
> I think @Matt-Yorkley tried to do it at some point for the shopfront endpoints

I added some "only fetch the objects within the daterange" gymnastics to the line items returned in permissions checks (for reports) a while back in #5030. Is that the one? Maybe we could try to do something similar for the various scoping processes? I'm not sure how easy it would be.

luisramos0:
ah, no, I meant something completely different (older), anyway, you suggestion is awesome!
Implementing a variation of VariantOverride#indexed that uses the report date range to filter overrides used in line_items in that range and then inject it as the 2nd param when creating the scoper.

oeoeaio:
> Implementing a variation of VariantOverride#indexed that uses the report date range to filter overrides used in line_items in that range and then inject it as the 2nd param when creating the scoper.

So the theory is that the shear number of overrides being loaded into memory will be the problem, rather than the time taken to query for relevant overrides? I am happy to have a crack and building a list of overrides (based on the report parameters) to inject into the scoper if pulling up a smaller number of overrides is desirable.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -89,5 +99,13 @@ def order_permissions
     def report_line_items
       @report_line_items ||= Reports::LineItems.new(order_permissions, options)
     end
+
+    def report_variant_overrides
+      @report_variant_overrides ||=
+        Reports::VariantOverrides.new(
+          line_items: order_permissions.visible_line_items,
+          distributor_ids: report_line_items.orders.result.pluck('DISTINCT distributor_id'),
```
oeoeaio, line 107: Not super happy with this approach, but I couldn't think of a better one. Any ideas?

I had an alternative implementation where the `Reports::VariantOverrides` class was only responsible for loading variant overrides for a single distributor at a time, which made it easier to just inject the relevant distributor each time a new one was encountered, but it obviously resulted in more queries. See 2a843bf33. Let me know which you think is preferable.

oeoeaio:
Hey @luisramos0, @Matt-Yorkley, I've had another go at this. The new implementation should only load variant overrides that are relevant to the line items in the report.

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -102,7 +102,10 @@ def report_line_items
 
     def report_variant_overrides
       @report_variant_overrides ||=
-        Reports::VariantOverrides.new(order_permissions.visible_line_items).indexed
+        Reports::VariantOverrides.new(
+          line_items: order_permissions.visible_line_items,
+          distributor_ids: report_line_items.orders.result.pluck('DISTINCT distributor_id'),
```
Matt-Yorkley, line 107: `report_line_items.orders.result` is an `ActiveRecord::Relation` here, so I think in this case `#select` will be more efficient than `#pluck` for building the `distributor_ids` part of the subquery, e.g:
```ruby
distributor_ids: report_line_items.orders.result.select(:distributor_id),
```

lib/open_food_network/orders_and_fulfillments_report.rb
Matt-Yorkley: Lets go with this version :+1:

lib/open_food_network/orders_and_fulfillments_report.rb
oeoeaio: Done :)

lib/open_food_network/reports/variant_overrides.rb
```diff
@@ -0,0 +1,36 @@
...
+          .joins(:variant)
+          .preload(:variant)
+          .where(
+            hub_id: distributor_ids,
+            variant_id: line_items.select(:variant_id)
+          )
+      end
+
+      def hash_of_hashes
+        Hash.new { |h, k| h[k] = {} }
```
luisramos0, line 32: To avoid single letter variables this should be hash and key, right?

lib/open_food_network/reports/variant_overrides.rb
```diff
@@ -0,0 +1,36 @@
...
+  module Reports
+    class VariantOverrides
+      def initialize(line_items:, distributor_ids:)
+        @line_items = line_items
+        @distributor_ids = distributor_ids
+      end
+
+      def indexed
+        variant_overrides.each_with_object(hash_of_hashes) do |variant_override, indexed|
+          indexed[variant_override.hub_id][variant_override.variant] = variant_override
```
luisramos0, line 13: Can we break this line in two and use a good variable name to make it easier for the reader?
maybe (a bit too long but easier to understand what is going on):
```
    variant_overrides.each_with_object(hash_of_hashes) do |variant_override, indexedVariantOverridesForAllDistributors|
         indexedVariantOverridesForDistributor = indexedVariantOverridesForAllDistributors[variant_override.hub_id]
         indexedVariantOverridesForDistributor[variant_override.variant] = variant_override

```   
I think it would be good to add a comment to the original VariantOverride#indexed and/or to this method explaining what is the data structure we are creating.

lib/open_food_network/reports/variant_overrides.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module OpenFoodNetwork
+  module Reports
+    class VariantOverrides
```
luisramos0, line 5: I think there's a better and more reusable structure for this code.
I think this should be a service class in app/services (it's not related to reports in its implementation only in its usage).
According to our convention the resource name of the service should be first, so something like VariantOverridesIndex or VariantOverridesIndexed.
The cool thing about this move is that you can move VariantOverride#indexed (the original one) into this service as well (this can be done in a later move/PR).

lib/open_food_network/reports/variant_overrides.rb
```diff
@@ -0,0 +1,36 @@
...
+        variant_overrides.each_with_object(hash_of_hashes) do |variant_override, indexed|
+          indexed[variant_override.hub_id][variant_override.variant] = variant_override
+        end
+      end
+
+      private
+
+      attr_reader :line_items, :distributor_ids
+
+      def variant_overrides
```
luisramos0, line 21: This method name could be a bit more specific. These are variant overrides of the given distributors_ids that correspond to the given variant_ids.
Maybe filtered_variant_overrides ?

lib/open_food_network/orders_and_fulfillments_report.rb
```diff
@@ -89,5 +99,13 @@ def order_permissions
     def report_line_items
       @report_line_items ||= Reports::LineItems.new(order_permissions, options)
     end
+
+    def report_variant_overrides
+      @report_variant_overrides ||=
+        Reports::VariantOverrides.new(
+          line_items: order_permissions.visible_line_items,
+          distributor_ids: report_line_items.orders.result.select('DISTINCT distributor_id'),
+        ).indexed
```
luisramos0, line 108: Personally I dont like the keyword parameters, specially because they are both mandatory in this case. I'd just use normal parameters.

lib/open_food_network/reports/variant_overrides.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module OpenFoodNetwork
+  module Reports
+    class VariantOverrides
+      def initialize(line_items:, distributor_ids:)
```
luisramos0, line 6: This class should not take line_items, they are not needed, you can just take variant_ids and call with line_items.select(:variant_id)

lib/open_food_network/orders_and_fulfillments_report.rb
luisramos0: I think the keywork parameters here are adding some readability, I think the problem here is the complexity, it should be extracted to a separate method, like this:
```ruby
    Reports::VariantOverrides.new(report_distributor_ids, variants_of_visible_line_items).indexed
   
    def variants_of_visible_line_items
        order_permissions.visible_line_items.select(:variant_id)
    end
    def report_distributor_ids
       report_line_items.orders.result.select('DISTINCT distributor_id')
    end
```

this is not so important but I thought of sharing.

oeoeaio: I've removed the keyword params, I personally like them for public methods just to make calls a little more readable, but very happy to stick with convention / preferences of the core team.

lib/open_food_network/reports/variant_overrides.rb
oeoeaio: Done

lib/open_food_network/reports/variant_overrides.rb
oeoeaio: I've added a comment, but I've kept this block the way that it is. I personally find it more readable the way that it is, but it you have a strong preference for breaking it down in the way you have , let me know and I'll make the change.

lib/open_food_network/reports/variant_overrides.rb
oeoeaio: Done.

lib/open_food_network/reports/variant_overrides.rb
oeoeaio: Done.

lib/open_food_network/reports/variant_overrides.rb
oeoeaio: Done.

lib/open_food_network/reports/variant_overrides.rb
luisramos0: no, it's ok :+1:

lib/open_food_network/orders_and_fulfillments_report.rb
luisramos0: yeah, it's a detail, the main problem here is really the complexity in building the values. A good rule of thumb is: one dot per line of code.

kirstenalarsen:
Hi people - I did a quick test of this to try and get it through in this release. As I'm not really a tester I don't have a set-up with lots of data to check performance so still best if someone else does that . . 

But I am not sure if this implementation is correct and need another couple of brains. This is what I did:
1. put an sku on two variants of a product ('var a sku' and 'var b sku')

Placed an order, looked at Customer Totals report
image

Correctly showing product skus

2. Put both variants into inventory and put a different sku on them 
image

3. Changed my inventory setting to Coord Inv only (unnecessary as these will override in my shop anyway but just to be sure)

4. Placed another order - so it's using the inventory variants this time

5. Ran the Customer Totals report. 

It is showing the inventory skus - yay. But it is showing them for all items in the report, including those from before I created the inventory and set those new skus

image

So the question is - is this correct? Once something is placed into inventory for the Shop and there is a new SKU - should this SKU be retrospectively applied to all orders in the past, including those before there was an inventory override and before the SKU existed? Overriding information that would have been attached to orders created with the product SKU? 

I am a bit agnostic on this - think it's likely a rare enough case that we don't need to worry too much and it does make sense that once someone puts it in inventory and attaches a SKU they DO mean that to refer to all previous sales of it . . so maybe it's ok?

oeoeaio:
Hi @kirstenalarsen! Thanks for testing this out.

> Once something is placed into inventory for the Shop and there is a new SKU - should this SKU be retrospectively applied to all orders in the past, including those before there was an inventory override and before the SKU existed?

That is the intended behaviour from my perspective. It is far easier to implement this way (I'm actually not even sure I can think of a way to make it not apply retrospectively, as far as I can recall SKU is a property of variants / overrides, not of line items). I have a use-case implementing it this way too: at BBFH, new variants are often added to the shop on an ad-hoc basis while an order cycle is open (as we get new information about products that are available). Often this is done in a rush to get them onto the shop before we send our weekly newsletter. We may not have time to fill in all of the details required to link products to external systems (i.e. SKU for our POS). We would usually reconcile our orders with other systems after an order cycle has closed, and it is at this point that we would probably go back and look at any new products and add them to our POS or match them to an existing POS product. It is only at this point that an SKU would be set most of the time, so it would be useful for us to have any SKUs added to the inventory apply retrospectively.

filipefurtad0:
Hi @oeoeaio,
And thanks @kirstenalarsen for picking this up! - I'd say that's great testing :-)

Verified all the findings. Also checked whether this is reversible: Checking the "Inherit?" works, and sets back the SKU settings from the shop. The general functionality of reports does not seem affected. All good.

I terms of testing I'd say this is good to go,
Thanks Rob!

Issue #4826, opened 2020-02-22
Make OFN independent of spree_core


luisramos0:
### What we should change and why (this is tech debt)
Quick summary of the reasons to do this:
- A lot of OFN code is decorating Spree models (from my quick analysis we decorate around 50% of all spree models!), there would be a huge advantage in having the spree core merged with the decorations so we can increase quality, test coverage, etc.
- On top of that spree 2.2, the next upgrade, includes a major refactoring of the spree adjustments logic, OFN has a lot of adaptations on top of the existing adjustments logic for example to include enterprise fees as adjustments. Adapting the OFN code to this new refactoring would be a nightmare. And not upgrading spree blocks us from the main issue which is upgrading rails.
- One huge advantage will be to see all the unnecessary/unused complexity from spree go away, like promotions or stock movements/transfers.

In this epic we remove spree_core dependency and bring to OFN the spree models that we really need on our side. We will need to move a few 1000s of lines of code but the task is relatively simple to execute :tada: 

We need to start with a detailed analysis of what decorators exist in OFN and what files from spree_core we will need to bring.

luisramos0:
This is the list of 235 files in spree_core we will have to handle (copy, merge or ignore).

133 are live code classes:
- 74 code classes will have to be copied to OFN
- 40 code classes have to be merged with the existing decorators
- 8 code classes will not be copied but need some adaptations in OFN
- 11 code classes  may not need to be copied

102 are test code classes (a lot of these will be ignored):
- 42 specs (and spec helpers) will have to be copied to OFN (or ignored)
- 22 specs have to merged with existing versions on OFN (or ignored)
- 38 are the spree test factories that will have to be copied if they are used in OFN

This is the full list with short comments (result of my first analys):

<details>
core/config/initializers/user_class_extensions.rb MAYBE COPY remove reference to dead Spree::Core::UserBanners
core/db/default/spree/roles.rb COPY
core/db/default/spree/zones.rb MAYBE COPY

core/vendor/assets/javascripts/jsuri.js COPY
core/vendor/assets/stylesheets/normalize.css COPY
core/vendor/assets/stylesheets/skeleton.css COPY

core/app/assets/javascripts/spree.js.coffee COPY (but remove unused ajax method)
core/app/assets/javascripts/admin/handlebar_extensions.js COPY

core/app/controllers/spree/base_controller.rb COPY

[...]

PR #5195, opened 2020-04-09
Delete dead code from several places...

luisramos0:
#### What? Why?

I found all this deadliness while doing the analysis to [remove spree_core dependency](https://github.com/openfoodfoundation/openfoodnetwork/issues/4826#issuecomment-611698328).

This is a good example of what we will get from removing spree_core, it will be a major clean up of our code base in just a few weeks :heart_eyes:

#### What should we test?
**Checkout**
Checkout quantity 2 of a variant with on_hand = 2. Afterwards, verify on_hand moved to 0. Edit order and change quantity to 1. Verify on_hand moved to 1. Edit order and remove line item, verify on_hand moved back to 2.

**Unit value**
In /admin/products set a product unit to Weight (g).
Edit one of its variants and verify you have a well formatted text field "Weight **(g)**". Change the value to 10000 and save.
Go back to /admin/products and set the product unit to Weight **(Kg)**.
Edit the same variant and verify you have a well formatted text field Weight **(Kg)** and the value is 10 now.

**Currency symbol**
Verify that the currency symbol appears where it was appearing before, for example on the checkout page and in some reports.

**Shop**
Verify you can filter products in a shop by product properties.

#### Release notes
Changelog Category: Removed
Removed 276 lines of dead code.


app/models/spree/property.rb
```diff
@@ -2,11 +2,16 @@ module Spree
   class Property < ActiveRecord::Base
     has_many :product_properties, dependent: :destroy
     has_many :products, through: :product_properties
+    has_many :producer_properties
```
Matt-Yorkley, line 5: Can we really delete producer properties here? :thinking:

Matt-Yorkley:
LOL! 

I hadn't read @sauloperez's review when I left my review.

filipefurtad0:
Hi @luisramos0 .

Thank you for the great testing notes. Went through them:

**Checkout**
- Checked-out quantity 2 of a variant with on_hand = 2.
- Afterwards, verified on_hand moved to 0.
- Edited the order and changed the quantity to 1.
- Verified on_hand moved to 1.
- Edited the order and removed the line item (by editing the order, as a user) and verified on_hand moved back to 2.

**Unit value**
- In /admin/products set a product unit to Weight (g).
- Edited one of its variants and verified that the text field "Weight (g)" is well formatted. Change the value to 10000 and save.
- Went back to /admin/products and set the product unit to Weight (Kg).
- Edited the same variant and verified that the text field Weight (Kg) is well formatted and that the value is 10 now.

Briefly checked whether this works for T as well, kL/L/mL.

**Currency symbol**
Verified that the currency symbol appeared as before:
- Customer: shopfront, orders, transactions
- Admin: orders, products and variants, reports such as payments, orders_and_fulfillment, sales_tax

**Shop**
Verified filters work as usual.



Did not spot anything unusual - ready to go!
276 code lines less :tada:

PR #5196, opened 2020-04-09
Admin Customers performance

Matt-Yorkley:
#### What? Why?

Improves performance in admin customers page.

#### What should we test?

Admin customers page is working correctly, should be much faster for large hubs.

#### Release notes

Improved performance on admin customers page

Changelog Category: Fixed

### DIFF
--- a/app/controllers/admin/customers_controller.rb
+++ b/app/controllers/admin/customers_controller.rb
@@ -67,7 +67,7 @@ module Admin
       enterprise_id = Enterprise.managed_by(spree_current_user).
         select('enterprises.id').find_by_id(params[:enterprise_id])
 
-      Customer.of(enterprise_id)
+      Customer.of(enterprise_id).includes(user: :credit_cards)
     end
 
     def load_managed_shops
--- a/app/controllers/admin/customers_controller.rb
+++ b/app/controllers/admin/customers_controller.rb
@@ -96,7 +96,10 @@ module Admin
       customer_tags = ::ActsAsTaggableOn::Tag.
         joins(:taggings).
         includes(:taggings).
-        where(taggings: { taggable_id: Customer.of(managed_enterprise_id), context: 'tags' })
+        where(taggings:
+                { taggable_type: 'Customer',
+                  taggable_id: Customer.of(managed_enterprise_id),
+                  context: 'tags' })
 
       customer_tags.each_with_object({}) do |tag, indexed_hash|
         customer_id = tag.taggings.first.taggable_id
--- a/app/controllers/admin/customers_controller.rb
+++ b/app/controllers/admin/customers_controller.rb
@@ -67,7 +67,7 @@ module Admin
       enterprise_id = Enterprise.managed_by(spree_current_user).
         select('enterprises.id').find_by_id(params[:enterprise_id])
 
-      Customer.of(enterprise_id).includes(user: :credit_cards)
+      Customer.of(enterprise_id).includes(:bill_address, :ship_address, user: :credit_cards)
     end
 
     def load_managed_shops
--- a/app/controllers/admin/customers_controller.rb
+++ b/app/controllers/admin/customers_controller.rb
@@ -17,8 +17,9 @@ module Admin
       respond_to do |format|
         format.html
         format.json do
-          tag_rule_mapping = TagRule.mapping_for(Enterprise.where(id: params[:enterprise_id]))
-          render_as_json @collection, tag_rule_mapping: tag_rule_mapping
+          render_as_json @collection,
+                         tag_rule_mapping: tag_rule_mapping,
+                         customer_tags: customer_tags_by_id
         end
       end
     end
@@ -64,10 +65,13 @@ module Admin
     def collection
       return Customer.where("1=0") unless json_request? && params[:enterprise_id].present?
 
-      enterprise_id = Enterprise.managed_by(spree_current_user).
-        select('enterprises.id').find_by_id(params[:enterprise_id])
+      Customer.of(managed_enterprise_id).
+        includes(:bill_address, :ship_address, user: :credit_cards)
+    end
 
-      Customer.of(enterprise_id).includes(:bill_address, :ship_address, user: :credit_cards)
+    def managed_enterprise_id
+      @managed_enterprise_id ||= Enterprise.managed_by(spree_current_user).
+        select('enterprises.id').find_by_id(params[:enterprise_id])
     end
 
     def load_managed_shops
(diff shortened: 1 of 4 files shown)


Matt-Yorkley:
Tested locally with a medium-sized hub (using production data). In production this endpoint regularly takes **40+ seconds** to load.

BEFORE: **14.6 seconds** and **~3200 queries**

Screenshot from 2020-04-09 16-29-43

AFTER: **3.3 seconds** and **26 queries**

Screenshot from 2020-04-09 17-48-47

app/controllers/admin/customers_controller.rb
```diff
@@ -17,8 +17,9 @@ def index
       respond_to do |format|
         format.html
         format.json do
```
Matt-Yorkley, line 19: It would be really nice to extract this to the API, but I think it would require a lot more changes to be piled on top of this PR...

app/controllers/admin/customers_controller.rb
```diff
@@ -64,8 +65,13 @@ def destroy
...
 
-      enterprise = Enterprise.managed_by(spree_current_user).find_by_id(params[:enterprise_id])
-      Customer.of(enterprise)
+      Customer.of(managed_enterprise_id).
+        includes(:bill_address, :ship_address, user: :credit_cards)
+    end
+
+    def managed_enterprise_id
+      @managed_enterprise_id ||= Enterprise.managed_by(spree_current_user).
+        select('enterprises.id').find_by_id(params[:enterprise_id])
```
luisramos0, line 74: this select can go to the end of the line, right? I think it will make it easier to understand.

app/controllers/admin/customers_controller.rb
```diff
@@ -80,5 +86,23 @@ def user_can_create_customer?
...
+    def tag_rule_mapping
+      TagRule.mapping_for(Enterprise.where(id: managed_enterprise_id))
+    end
+
+    # Fetches tags for all customers of the enterprise and returns a hash indexed by customer_id
+    def customer_tags_by_id
+      customer_tags = ::ActsAsTaggableOn::Tag.
+        joins(:taggings).
+        includes(:taggings).
+        where(taggings: { taggable_id: Customer.of(managed_enterprise_id), context: 'tags' })
```
luisramos0, line 99: I was trying to do this for the inventory page: https://github.com/openfoodfoundation/openfoodnetwork/issues/5131#issuecomment-608345245
nice!

app/controllers/admin/customers_controller.rb
```diff
@@ -80,5 +86,23 @@ def user_can_create_customer?
...
+
+      customer_tags.each_with_object({}) do |tag, indexed_hash|
+        customer_id = tag.taggings.first.taggable_id
+        indexed_hash[customer_id] ||= []
+        indexed_hash[customer_id] << tag.name
+      end
+    end
```
luisramos0, line 106: I'd call indexed_hash   customer_tags_by_id instead.

app/controllers/admin/customers_controller.rb
Matt-Yorkley: Yeah, this was a bit of a pain. Initially I had a look through the repo for the `ActsAsTaggable` gem and there was lots of open issues with people basically screaming "HOW CAN I EAGER-LOAD TAGS??!?!!", but with no resolutions. :see_no_evil:

I think we have to be a bit careful if we try to re-use this bit of code elsewhere, as I think we do some weird modifications to tagging in some places. I think we're abusing the gem a bit already. Basically I think queries that fetch certain tag-related are different depending on the model. There's polymorphic associations and plenty of "automagic" things going on in the background that is really opaque and difficult to follow :see_no_evil:

app/controllers/admin/customers_controller.rb
Matt-Yorkley: I tried that previously because I thought it looked nicer, but the method names inside and outside the `#each_with_object` block conflicted, and everything broke, so I put it back... :man_shrugging:

app/controllers/admin/customers_controller.rb
Matt-Yorkley: I think I tried that previuously and it didn't work, the `#find_by_id` didn't want to take a chained `#select` after it, and complained loudly... :man_shrugging: 

I'll give it another try :+1:

Matt-Yorkley: Nope, all the tests in customers_controller_spec immediately break with the suggested change, throwing:
```ruby
NoMethodError: private method `select' called for #<Enterprise:0x000055ff6dd0c100>
```

app/controllers/admin/customers_controller.rb
luisramos0: yeah, it's great it's working here!

Matt-Yorkley:
I just double-checked some of the SQL before and after and I realised I forgot something crucial here! It's a polymorphic association, so we have to specify the class the tags are for: 

Screenshot from 2020-04-10 20-18-46

New commit added :point_up:

Matt-Yorkley:
> Hopefully we can extract this into a reusable service as some point and utilise it elsewhere.

I had a quick look and it looks like we can use this to improve reports...

app/controllers/admin/customers_controller.rb
luisramos0: ok, thanks for trying. the find_by method doesnt return a relation, it returns the enterprise.

luisramos0:
awesome, we should create an issue for that like: extract customer_tags_by_id and re-use it :tada:

app/models/spree/user.rb
```diff
@@ -125,7 +125,12 @@ def can_own_more_enterprises?
     end
 
     def default_card
-      credit_cards.where(is_default: true).first
+      # Don't re-fetch associated cards from the DB if they're already eager-loaded
+      if credit_cards.loaded?
+        credit_cards.to_a.find(&:is_default)
+      else
+        credit_cards.where(is_default: true).first
```
sauloperez, line 132: have you tried defining this as an association and using AR association's [:condition](https://apidock.com/rails/v3.2.13/ActiveRecord/Associations/ClassMethods/has_one)? This feels to me like something AR should do and not us. 

We can improve it in a different PR if you prefer. I don't want to delay this PR.

filipefurtad0:
Hey @Matt-Yorkley ,

Uff. This one was a brainer to set up - and, well, made me very grateful this and all performance efforts :-)

In AU and as superadmin:
- I had a look in reports for an enterprise with plenty of customers, found Ballantyne (Thornbury).
- Turned over to  which has over a hundred. navigated to Users > Permissions > and added a user to manage this enterprise
- logged into this users account
- navigated to /admin/customers/ 
- Reloaded the page and had a few screenshots, before and after the PR

Somehow, I could not "measure" a tangible difference... Load times for the /admin/customers/ page:


Before:
image

After this PR:
image

Perhaps the customer number isn't large enough?
Tried performing the usual operations here, as well, as in the users page. I found nothing unusual.

I'm moving this to ready to go, but adding the feedback label, just in case you feel there is anything more I should try here.

Matt-Yorkley:
Thanks Filipe! Interesting results. I just had another look, and the optimisations here will apply more to larger hubs that have any/all of the following:
- lots of customers
- lots of card payments
- lots of completed orders
- using tagging on customers

I think **Prom Coast Food Hub** is a good one to go for if you're testing large hubs on Aus Staging.

filipefurtad0:
Ah - that one has over 600 customers. I guess I should just have asked :-) 
Great, I'll have a second look at this.
Thanks @Matt-Yorkley

filipefurtad0:
Ok! Now things look quite different.

Time to refresh page:
Before PR: ~10 sec
After PR: ~5 sec

Time to display data, after clicking the [ Show all (589 more) ] button:
Before PR ~30 sec
After PR ~20 sec

Pic before PR:
BEFORE_PR

Pic after PR:
AFTER_PR

Thanks for your help @Matt-Yorkley 
Good to go.

PR #5197, opened 2020-04-09
Cart populate performance

Matt-Yorkley:
#### What? Why?

Adds some quick improvements to `CartController#populate`, which currently seems to perform very well in some conditions and very badly in others. It _think_ this difference is probably just when the number of items in the cart is very large. This endpoint is in our **top 3 pain points** in Datadog.

Removes a couple of N+1 queries.

#### What should we test?

Adding and removing items in the cart works. Loading (when adding a new item to the cart) should be faster if the number of items already in the cart is large.

#### Release notes

Improved performance on CartController#populate for large carts

Changelog Category: Fixed

### DIFF
--- a/app/services/variants_stock_levels.rb
+++ b/app/services/variants_stock_levels.rb
@@ -8,7 +8,8 @@ class VariantsStockLevels
     variant_stock_levels = variant_stock_levels(order.line_items)
 
     order_variant_ids = variant_stock_levels.keys
-    missing_variants = Spree::Variant.where(id: (requested_variant_ids - order_variant_ids))
+    missing_variants = Spree::Variant.includes(:stock_items).
+      where(id: (requested_variant_ids - order_variant_ids))
 
     missing_variants.each do |missing_variant|
       variant = scoped_variant(order.distributor, missing_variant)
--- a/app/services/variants_stock_levels.rb
+++ b/app/services/variants_stock_levels.rb
@@ -8,10 +8,12 @@ class VariantsStockLevels
     variant_stock_levels = variant_stock_levels(order.line_items)
 
     order_variant_ids = variant_stock_levels.keys
-    missing_variant_ids = requested_variant_ids - order_variant_ids
-    missing_variant_ids.each do |variant_id|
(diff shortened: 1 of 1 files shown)


filipefurtad0:
Hey @Matt-Yorkley,

A considerable reduction in load time of the /cart page, with 99 items in cart is apearent, but times vary a bit, on each reload.

- before: around 25sec.
- after: around 15sec.

Tested adding and removing items in the cart works - all good.

Moving to ready to go.

RachL:
@filipefurtad0 a tip when testing the cart:e it can be nice to add stuff like : the product runs out of stock or fees are added by managers during customers journey from cart to checkout. Should be well covered by automated tests, but we never know :)

filipefurtad0:
Thank you @RachL - I oversaw that, great that you pointing this out.
I build was still on the server, so I had a second look at it. 

Inventory actions work fine - like different cases of products running out of stock.

The second test case you mention is an interesting one. I checked this with and without the PR, so I think this is anyway good to go. However: 
- If a user has items in a cart, and the price is changed by the shop owner during the user's journey to checkout, no changes appear to the user i.e., the user pays the price which the product had when he put in the cart.
- What's more - and this is maybe a bit of an incoherence: if the user goes back to the shopfront, he will _see_ the new price, he may add items with the new price, to the previously added, but on checkout the value of the purchase is based on the previous price per unit.

So lets say for example, the customer is buying cheese (how original :trollface: ) at 8 eur per item:

image

- The user proceeds to checkout (but does not finish the process)
- Meanwhile, admin changes this to 10 eur per Item.
- User goes back to shopfront, sees the new price, but adds two more items:

image

- He moves further to the cart:

image

- and to checkout, paying the previous price per unit.

image

This is observed on this PR and on staging-fr, which has a totally different PR staged. So, again, I don't think this is related to this PR.

But I guess this really isn't expected/desired behaviour, right?
(more on this on Slack)

Thanks again, for pointing out for further testing.

luisramos0:
nice testing!
We can agree to change this behavior but it doesnt look broken to me. The user is buying the product at the price they saw it initially advertised.

Issue #5192, opened 2020-04-09
JS errors on checkout (bugsnagJS)


luisramos0:
## Description
Since we added bugsnagJS we are seeing two common JS errors on checkout that are happening to quite a lot of users:
![image](https://user-images.githubusercontent.com/1640378/78877032-25490d80-7a48-11ea-82b0-9f7d6982784e.png)
Just in UK 93 and 53 users respectively.

These errors do not necessarily mean the checkout is broken for all these users, but I think we need to investigate these errors as they can be causing checkout problems. 

## Expected Behavior
No JS errors related to checkout. 

## Actual Behaviour
JS errors.

## Steps to Reproduce
Unkonwn so far.

type error on bugsnag:
https://app.bugsnag.com/yaycode/open-food-network-js-uk/errors/5e847cf9751a150018bc4723?filters[event.since][0]=30d&filters[error.status][0]=open

UnhandledRejection on bugsnag:
https://app.bugsnag.com/yaycode/open-food-network-js-uk/errors/5e847f9022ed50001754372a?filters[event.since][0]=30d&filters[error.status][0]=open

## Severity
I'd put this as S2, it's a critical feature that may be broken.

## Your Environment
links to UK prod but it's happening on other servers as well.

## Possible Fix
In one case we are getting http 400 responses from the server so it should be possible to at least   add some more specific debug info to the error message to help future debugging.

luisramos0:
So, unhandled rejections are false alarms, they are all user errors, real examples are:
![image](https://user-images.githubusercontent.com/1640378/78984035-4596cc00-7b1d-11ea-9ea1-f0700d5d67d3.png)
or
![image](https://user-images.githubusercontent.com/1640378/78984062-4f203400-7b1d-11ea-9270-24447f32e97f.png)

I will try to understand why this is generating a bugsnag error on the browser (it should only be flagged on teh server side).

This is the JS explanation of unhandledrejection:
"The unhandledrejection event is sent to the global scope of a script when a JavaScript Promise that has no rejection handler is rejected"

luisramos0:
I will improve the checkout Javascript error handling code  in #5202

Matt-Yorkley:
Are these cases covered in feature specs?

luisramos0:
yes, checkout_spec tests a failed payment with bogus payment method. This test relies on this js code to pass.
Here spec/features/consumer/shopping/checkout_spec.rb:412

luisramos0:
[...]

PR #5202, opened 2020-04-10
Improve Checkout error handling in JS

luisramos0:
- Switch from old success/error to modern then/catch structure: .catch() will get a few more errors then .errors().
- Add try/catch inside catch to detect any errors parsing the response error payload including when the response doesnt include any error or flash information. In these cases log to bugsnag (throw the error again) and inform the user so we dont have silent checkout errors.

#### What? Why?

Related to #5192

#### What should we test?
We need to test checkout error scenarios:
- Redirect to cart page if insufficient stock: go to checkout with variant X in the cart, go to the backoffice and make on_hand zero for variant X, return to checkout page and click "place order", you should be redirected to the cart page with a message insufficient stock
- Try to checkout with invalid email address, for example, "ofn@org"
- Try to pay with an invalid credit card (bogus or stripe payment methods)
- Try to pay with a StripeSCA card where authentication is required for example 4000 0027 6000 3184 and fail the authentication on the stripe test page, you should see an flash error on the checkout page when you are redirected to the checkout page.

#### Release notes
Changelog Category:  Changed
Improved resiliency of the checkout code that handles error scenarios.

### DIFF
--- a/app/assets/javascripts/darkswarm/services/checkout.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/checkout.js.coffee
@@ -22,11 +22,14 @@ Darkswarm.factory 'Checkout', ($injector, CurrentOrder, ShippingMethods, StripeE
           if response.data.path
             Navigation.go response.data.path
           else
+            throw response unless response.data.errors || response.data.flash
+
             Loading.clear()
             @errors = response.data.errors
             RailsFlashLoader.loadFlash(response.data.flash)
         catch error
-          RailsFlashLoader.loadFlash("Unkown error occurred")
+          RailsFlashLoader.loadFlash(error: t("checkout.failed")) # inform the user about the unexpected error
+          throw error # generate a BugsnagJS alert
 
     # Rails wants our Spree::Address data to be provided with _attributes
     preprocess: ->
--- a/app/assets/javascripts/darkswarm/services/checkout.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/checkout.js.coffee
@@ -19,18 +19,23 @@ Darkswarm.factory 'Checkout', ($injector, CurrentOrder, ShippingMethods, StripeE
         Navigation.go response.data.path
       .catch (response) =>
         try
-          if response.data.path
-            Navigation.go response.data.path
-          else
-            throw response unless response.data.errors || response.data.flash
-
-            Loading.clear()
-            @errors = response.data.errors
-            RailsFlashLoader.loadFlash(response.data.flash)
+          @handle_checkout_error_response(response)
         catch error
-          Loading.clear()
-          RailsFlashLoader.loadFlash(error: t("checkout.failed")) # inform the user about the unexpected error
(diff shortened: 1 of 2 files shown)


luisramos0:
tagging bug-s2 for clarity.

filipefurtad0:
Hi @luisramos0,

Had a look at this scenario of insufficient stock in the cart, and the test cases:
- The user gets redirected to cart page if the products' stock is reduced by the admin/managers during customers journey from cart to checkout. This red banner appears, with "An item in your cart has become unavailable." as well as the red text "Insufficient stock available, only 2 remaining".
    Try to checkout with invalid email address, for example, "ofn@org"
    Try to pay with an invalid credit card (bogus or stripe payment methods)
    Try to pay with a StripeSCA card where authentication is required for example 4000 0027 6000 3184 and fail the authentication on the stripe test page, you should see an flash error on the checkout page when you are redirected to the checkout page.


An item in your cart has become unavailable.

filipefurtad0:
Hi @luisramos0,

Had a look at the described test scenarios:

1) Checked that the user gets redirected to cart page if the products' stock is reduced by the admin/managers during customers journey from cart to checkout. This red banner appears, with "An item in your cart has become unavailable." as well as the red text "Insufficient stock available, only 2 remaining".
- however, at this stage, the user gets "stuck" in the /cart page. Even if the number of items is reduced till the "insufficient stock" warning disappears, one does not seem to be able to proceed to checkout or to the shopfront. Here, the only way out, i.e. workaround is either a) pressing the trash-bin and removing the items from the cart, or b) using the address bar to type an address. I am pretty sure, this is not introduced by this PR - as I have observed this recently - so I will open an issue on this one.

2) Checking out by clicking the "place order" with invalid email address like ofn@ofn displays the error "Customer E-mail is invalid"

image

The field itself accepts an address like ofn@ofn; it does reject the email address if the @ sign is removed, thus preventing the user from clicking "place order".

3) Paying with an invalid credit card (bogus or stripe payment methods) -> Generates the red banner "Error: Your card number is invalid."

4) Trying to pay with a StripeSCA card 4000 0027 6000 3184 and fail the authentication on the stripe test page, shows the flash error on the checkout page, after being redirected to the checkout page.

image
- The error message displays the correct content. I am not too sure about using the underscores on this message. I think for the user this is perhaps not so readable/friendly.

5) I added the test-case in which one completes authentication but uses a card with insufficient funds. The correct error message is displayed "Your card has insufficient funds." in the red banner, on the checkout page.

I'd say this is ready to go. I will open issues on the respective issues found on 1) and 4).

luisramos0:
thanks Filipe, nice testing!!!

1) I am not sure but I think there is already an issue in gh for that, please try to search for it before creating a new one.

4) yes, in stripeSCA I decided to show the user the message that comes from Stripe directly. It's not nice but it's probably more informative than a generic error message. I think we can create a lower priority issue for the stripeSCA backlog.

filipefurtad0:
Thank you @luisramos0,

Created issues
1) #5239 S3 and
2) #5240 S5

luisramos0:
thanks Filipe!

PR #5223, opened 2020-04-13
Change default user from spree@example.com to ofn@example.com :heart: 

luisramos0:
#### What? Why?

Bye bye spree@example.com :heart: 
Welcome ofn@example.com with password ofn123

#### What should we test?
Reset your local DB with bundle exec db:reset
Run local server rails s
Login as ofn@example.com :heavy_check_mark: 

Everything else is tested as part of the build.

#### Release notes
Changelog Category: Changed
We say goodbye to the spree test user spree@example.com and welcome the new test user ofn@example.com

### DIFF
--- a/db/default/users.rb
+++ b/db/default/users.rb
@@ -6,13 +6,13 @@ def prompt_for_admin_password
     password = ENV['ADMIN_PASSWORD'].dup
     say "Admin Password #{password}"
   else
-    password = ask('Password [spree123]: ') do |q|
+    password = ask('Password [ofn123]: ') do |q|
       q.echo = false
       q.validate = /^(|.{5,40})$/
       q.responses[:not_valid] = 'Invalid password. Must be at least 5 characters long.'
       q.whitespace = :strip
     end
-    password = 'spree123' if password.blank?
+    password = 'ofn123' if password.blank?
   end
 
   password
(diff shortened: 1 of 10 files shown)


db/default/users.rb
```diff
@@ -23,20 +23,20 @@ def prompt_for_admin_email
...
 
   email
 end
 
 def create_admin_user
   if ENV.fetch("AUTO_ACCEPT", true)
-    password = ENV.fetch("ADMIN_PASSWORD", "spree123").dup
-    email = ENV.fetch("ADMIN_EMAIL", "spree@example.com").dup
+    password = ENV.fetch("ADMIN_PASSWORD", "ofn123").dup
+    email = ENV.fetch("ADMIN_EMAIL", "ofn@example.com").dup
```
sauloperez, line 39: these are the values specified as ENV vars in the docker-compose.yml. Could you please update that too?

luisramos0: sorry I dont understand what do you mean.
the values in docker-compose are already ofn@example.com ofn123 credits to @pacodelaluna btw

sauloperez: I mean that since the env vars defined in the docker-compose.yml specify the values that are now the defaults, there's no need to provide them. Without them, password and email will be the same.

luisramos0: ah, ok!
done :+1: 
and tested locally, it works.

PR #5227, opened 2020-04-14
Update Docker documentation

### LINKED ISSUE #5226: Update Docker documentation
## Description
The gem updates is missing from the Docker md file.

## Expected Behaviour
To follow the documentation from scratch will lead you to a working environment.

## Actual Behaviour
You need to do some steps manually like gem update and migrations run, and it is fine. We should just put them inside the documentation.

pacodelaluna:
#### What? Why?
Some step were missing inside the documentation if you run the composer from scratch.

Closes #5226 

#### What should we test?
We should run the docker process written inside DOCKER.md file and check it is working fine.

Changelog Category: Changed

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -35,8 +35,15 @@ Download the Docker images and build the containers:
 $ docker-compose build
 ```
 
+Install the required gems:
+
+```sh
+$ docker-compose run web bundle exec bundle install
+```
+
 Setup the database and seed it with sample data:
 ```sh
+$ docker-compose run web bundle exec rake db:create
 $ docker-compose run web bundle exec rake db:reset
 $ docker-compose run web bundle exec rake db:test:prepare
 $ docker-compose run web bundle exec rake ofn:sample_data
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,10 @@ services:
       ADMIN_PASSWORD: ofn123
       OFN_DB_HOST: db
     command: >
-      bash -c "wait-for-it -t 30 db:5432 &&
+      bash -c "(bundle check || bundle install) &&
+               wait-for-it -t 30 db:5432 &&
+               bundle exec rake db:create &&
+               bundle exec rake db:migrate &&
                rm -f tmp/pids/server.pid &&
                bundle exec rails s -p 3000 -b '0.0.0.0'"
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,8 +28,7 @@ services:
       ADMIN_PASSWORD: ofn123
       OFN_DB_HOST: db
     command: >
-      bash -c "(bundle check || bundle install) &&
-               wait-for-it -t 30 db:5432 &&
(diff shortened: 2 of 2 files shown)


pacodelaluna:
Actually, it was made this way initially, the DOCKER.md file was made for this, don't know why we added the `db:reset` part at some point but better to have it aside for smoother container restart.

luisramos0:
we have recently agreed to revert this change @pacodelaluna here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5088

luisramos0:
I think this is a good sign, it means docker is not just for first timers but also for us.
The db:reset is now in Docker.md.

re create/migrate, I think it will much slower for first timers than db:reset (that uses db:schema:load instead of running all migrations)

pacodelaluna:
Well, if you follow the `Docker.md` file, you will never run `create` and `migrate` initially, the `reset` may come before and then, when you restart the container, you use `'migrate` in order to keep your migrations updated on container start, same with `bundle install` and gem updates. I am not getting why this should be done manually, we will need to run it manually each time we rebase our current branch, maybe I am just to lazy :)

I agree then that we could remove the `db:create` part.

I think that the `bundle install` part is also missing inside the DOCKER.md file. Maybe we should add it right?

luisramos0:
>> if you follow the Docker.md file, you will never run create and migrate initially
I just found that  db:reset includes db:setup that includes db:create but only in rails 4.
In rails 3, db:reset includes db:setup but db:setup does not include db:create
Maybe you need the db create in rails3. I am not sure because I tested the latest version of this Docker.md and it worked for me in rails 3.

>> same with bundle install
initial bundle install is already done in the Dockerfile: https://github.com/openfoodfoundation/openfoodnetwork/pull/5088/files#diff-3254677a7917c6c01f55212f86c57fbfR43

>> we will need to run it manually each time we rebase our current branch
yes, correct, we agreed that it is better to have to do it manually when needed that to do it everytime you want to start the container (which is much more often). The main objective of #5088 is to make container start/stop faster.

pacodelaluna:
@luisramos0 I have just updated the documentation, or else we can close this ticket, no worries.

DOCKER.md
```diff
@@ -35,8 +35,15 @@ Download the Docker images and build the containers:
 $ docker-compose build
 ```
 
+Install the required gems:
+
+```sh
+$ docker-compose run web bundle exec bundle install
```
luisramos0, line 41: this is not needed, this already in the Dockerfile

docker-compose.yml
```diff
@@ -28,8 +28,7 @@ services:
       ADMIN_PASSWORD: ofn123
       OFN_DB_HOST: db
     command: >
-      bash -c "(bundle check || bundle install) &&
-               wait-for-it -t 30 db:5432 &&
+      bash -c "wait-for-it -t 30 db:5432 &&
```
luisramos0: I think bundle install was left there because it is fairly quick to run on every restart.

PR #5231, opened 2020-04-14
Add DFC Provider engine

### LINKED ISSUE #5217: Create an new API endpoint for DFC Catalog
We need to serve the Entreprise catalog on a specific API supporting the JSON-LD format required for DFC import feature.

A new endpoint will displayed the Enterprise Catalog in the JSON-LD format.
We will also need to put this catalog under an authentication layer, dealing witth OIDC tokan that will be sent from DFC application. This will be done in a separated ticket.

The data format has to follow the JSON-LD format (https://json-ld.org/) define into the ontology used by the DFC protoype (https://github.com/datafoodconsortium/ontology).

To avoid complex inceptions at this step, please find a example for the JSON-LD format to use for the Catalog:
```
{"@context":{"DFC":"http://datafoodconsortium.org/ontologies/DFC_FullModel.owl#","@base":"https://my-app.com/api/dfc/v1/"},"@id":"/entreprise/13616/catalog","DFC:supplies":[{"DFC:description":"DFC-Oeufs 12 pièces","DFC:quantity":12,"@id":"/suppliedProduct/item21345","DFC:hasUnit":{"@id":"/unit/Piece"}},{"DFC:description":"DFC-Oeufs 6 pièces","DFC:quantity":6,"@id":"/suppliedProduct/item21344","DFC:hasUnit":{"@id":"/unit/Piece"}},{"DFC:description":"DFC-Paniers de légumes 12 Kg.","DFC:quantity":12,"@id":"/suppliedProduct/item21347","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Paniers de légumes 5 Kg.","DFC:quantity":5,"@id":"/suppliedProduct/item21346","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Pesto Fanes de radis 200g 1 pièce","DFC:quantity":1,"@id":"/suppliedProduct/item21340","DFC:hasUnit":{"@id":"/unit/Piece"}},{"DFC:description":"DFC-Pesto Basilic pot 200g 1 pièce","DFC:quantity":1,"@id":"/suppliedProduct/item21339","DFC:hasUnit":{"@id":"/unit/Piece"}},{"DFC:description":"DFC-Pesto fanes de carottes 200g 1 pièce","DFC:quantity":1,"@id":"/suppliedProduct/item21341","DFC:hasUnit":{"@id":"/unit/Piece"}},{"DFC:description":"DFC-potimarron part 250 g.","DFC:quantity":250,"@id":"/suppliedProduct/item21342","DFC:hasUnit":{"@id":"/unit/Gram"}},{"DFC:description":"DFC-Potimarron 2 Kg.","DFC:quantity":2,"@id":"/suppliedProduct/item21343","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Radis ronds 1 Kg.","DFC:quantity":1,"@id":"/suppliedProduct/item21338","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Tomates Ananas 1 Kg.","DFC:quantity":1,"@id":"/suppliedProduct/item21335","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Tomates Mix tomates anciennes pour sauce 5 Kg.","DFC:quantity":5,"@id":"/suppliedProduct/item21337","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Tomates Coeur de boeuf 1 Kg.","DFC:quantity":1,"@id":"/suppliedProduct/item16095","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Tomates Rose de Berne 1 Kg.","DFC:quantity":1,"@id":"/suppliedProduct/item21334","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Tomates Mis tomates anciennes 3 Kg.","DFC:quantity":3,"@id":"/suppliedProduct/item21336","DFC:hasUnit":{"@id":"/unit/Kilogram"}},{"DFC:description":"DFC-Tomates Tomates andines 1 Kg.","DFC:quantity":1,"@id":"/suppliedProduct/item21333","DFC:hasUnit":{"@id":"/unit/Kilogram"}}],"@type":"DFC:Entreprise"}
```

## Useful information for inception

https://github.com/openfoodfoundation/openfoodnetwork/wiki/API-documentation

https://community.openfoodnetwork.org/t/is-the-ofn-api-ready-to-be-used-by-external-clients/1740/6

### LINKED ISSUE #5218: Find a way to link the OIDC token to an user
DFC application will do API calls with an OIDC token. These tokens will be related to an account created here: https://login.lescommuns.org/auth/realms/master/protocol/openid-connect/auth?client_id=4461746120466f6f6420436f6e736f727469756d&scope=openid&response_type=code&redirect_uri=http%3A%2F%2F185.25.192.252%3A8080%2Flogin%2Fauth%2Fcb&state=Sit1yrILNQx5oOm49n7Lo0GA52DNYwmS-gbeKYLtTsw&session=false

The token is from JWT format and we can extract information from it, like the 'email'. This can be used to link the user from OFN database.

The DFC Provider Catalog API is able to link a OFN user from an Access Token present inside the request header (OAuth standard, Authorization Bearer).

## Useful information for inception
Need to ask Simon about this.

pacodelaluna:
#### What? Why?
We need to provide an API to help DFC application to synchronize our catalog.

Closes #5217
Closes #5218 

This engine will provide an isolated stack for this API. It will help to serve the catalog with the proper format. And also to authenticate the user.

#### What should we test?
You can test that the API is returning the proper catalog in JSON-LD format. An example of an endpoint will be `/api/dfc_provider/enterprises/1/products`. It will also support `/api/dfc_provider/enterprises/default/products`, fetching the first enterprise of the user for this.

Example:
```
curl -v -X GET -H "Authorization: Bearer "OIDC_token"" -H "Content-Type: application/json" http://localhost:3000/api/dfc_provider/enterprises/default/products.json
```

#### Release notes
- Add DFC Provider engine to implement the Products Catalog of an user with DFC compliance

Changelog Category: Added

### DIFF
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -28,7 +28,12 @@ module DfcProvider
       private
 
       def check_enterprise
-        @enterprise = @user.enterprises.first
+        @enterprise =
+          if params[:enterprise_id] == 'default'
+            @user.enterprises.first
+          else
+            @user.enterprises.where(id: params[:enterprise_id]).first
+          end
 
         return if @enterprise.present?
 
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 # Controller used to provide the API products for the DFC application
-module Api
-  module DfcProvider
-    class ProductsController < Api::BaseController
+module DfcProvider
+  module Api
+    class ProductsController < ::Api::BaseController
+      skip_before_filter :authenticate_user
       before_filter :set_enterprise
+      before_filter :authenticate_user
+      skip_authorization_check
 
       def index
         products = @enterprise.
@@ -19,6 +22,10 @@ module Api
       end
 
       private
+
+      def authenticate_user
+        @current_api_user = @enterprise.owner
+      end
 
       def set_enterprise
         @enterprise = ::Enterprise.find(params[:enterprise_id])
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -4,13 +4,8 @@
 module DfcProvider
   module Api
     class ProductsController < ::ActionController::Metal
-      include ActionController::Head
-      include AbstractController::Rendering
-      include ActionController::Rendering
-      include ActionController::Renderers::All
-      include ActionController::MimeResponds
-      include ActionController::ImplicitRender
-      include AbstractController::Callbacks
+      include Spree::Api::ControllerSetup
+
       # To access 'base_url' helper
       include ActionController::UrlFor
       include Rails.application.routes.url_helpers
@@ -19,8 +14,6 @@ module DfcProvider
                     :check_enterprise,
                     :check_user,
                     :check_accessibility
-
-      respond_to :json
 
       def index
         products = @enterprise.
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -16,9 +16,8 @@ module DfcProvider
       include Rails.application.routes.url_helpers
 
       before_filter :check_authorization,
-                    :check_enterprise,
                     :check_user,
-                    :check_accessibility
+                    :check_enterprise
 
       respond_to :json
 
@@ -28,7 +27,7 @@ module DfcProvider
           includes(:product, :inventory_items)
 
         products_json = ::DfcProvider::ProductSerializer.
-          new(@enterprise, products, base_url).
+          new(products, base_url).
           serialized_json
 
         render json: products_json
@@ -37,7 +36,7 @@ module DfcProvider
       private
 
       def check_enterprise
-        @enterprise = ::Enterprise.where(id: params[:enterprise_id]).first
+        @enterprise = @user.enterprises.first
 
         return if @enterprise.present?
 
@@ -47,7 +46,7 @@ module DfcProvider
       def check_authorization
         return if access_token.present?
 
-        head :unauthorized
+        head :unprocessable_entity
       end
 
       def check_user
@@ -55,13 +54,7 @@ module DfcProvider
 
         return if @user.present?
 
-        head :unprocessable_entity
-      end
-
-      def check_accessibility
-        return if @enterprise.owner == @user
-
-        head :forbidden
+        head :unauthorized
       end
 
       def base_url
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -3,11 +3,24 @@
 # Controller used to provide the API products for the DFC application
 module DfcProvider
   module Api
-    class ProductsController < ::Api::BaseController
-      skip_before_filter :authenticate_user
-      before_filter :set_enterprise
-      before_filter :authenticate_user
-      skip_authorization_check
+    class ProductsController < ::ActionController::Metal
+      include ActionController::Head
+      include AbstractController::Rendering
+      include ActionController::Rendering
+      include ActionController::Renderers::All
+      include ActionController::MimeResponds
+      include ActionController::ImplicitRender
+      include AbstractController::Callbacks
+      # To access 'base_url' helper
+      include ActionController::UrlFor
+      include Rails.application.routes.url_helpers
+
+      before_filter :check_authorization,
+                    :check_enterprise,
+                    :check_user,
+                    :check_accessibility
+
+      respond_to :json
 
       def index
         products = @enterprise.
@@ -23,16 +36,44 @@ module DfcProvider
 
       private
 
-      def authenticate_user
-        @current_api_user = @enterprise.owner
+      def check_enterprise
+        @enterprise = ::Enterprise.where(id: params[:enterprise_id]).first
+
+        return if @enterprise.present?
+
+        head :not_found
       end
 
-      def set_enterprise
-        @enterprise = ::Enterprise.find(params[:enterprise_id])
+      def check_authorization
+        return if access_token.present?
+
+        head :unauthorized
+      end
+
+      def check_user
+        @user = authorization_control.process
+
+        return if @user.present?
+
+        head :unprocessable_entity
+      end
+
+      def check_accessibility
+        return if @enterprise.owner == @user
+
+        head :forbidden
       end
 
       def base_url
         "#{root_url}api/dfc_provider"
+      end
+
+      def access_token
+        request.headers['Authorization'].to_s.split(' ').last
+      end
+
+      def authorization_control
+        DfcProvider::AuthorizationControl.new(access_token)
       end
     end
   end
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -3,16 +3,8 @@
 # Controller used to provide the API products for the DFC application
 module DfcProvider
   module Api
-    class ProductsController < ::ActionController::Metal
-      include ActionController::Head
-      include AbstractController::Rendering
-      include ActionController::Rendering
-      include ActionController::Renderers::All
-      include ActionController::MimeResponds
-      include ActionController::ImplicitRender
-      include AbstractController::Callbacks
+    class ProductsController < ::ActionController::Base
       # To access 'base_url' helper
-      include ActionController::UrlFor
       include Rails.application.routes.url_helpers
 
       before_filter :check_authorization,
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -18,11 +18,11 @@ module DfcProvider
           inventory_variants.
           includes(:product, :inventory_items)
 
-        products_json = ::DfcProvider::ProductSerializer.
+        serialized_data = ::DfcProvider::ProductSerializer.
           new(products, base_url).
-          serialized_json
+          serialized_data
 
-        render json: products_json
+        render json: serialized_data
       end
 
       private
--- a/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
+++ b/engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
@@ -4,8 +4,13 @@
 module DfcProvider
   module Api
     class ProductsController < ::ActionController::Metal
-      include Spree::Api::ControllerSetup
-
+      include ActionController::Head
+      include AbstractController::Rendering
+      include ActionController::Rendering
+      include ActionController::Renderers::All
+      include ActionController::MimeResponds
+      include ActionController::ImplicitRender
+      include AbstractController::Callbacks
       # To access 'base_url' helper
       include ActionController::UrlFor
       include Rails.application.routes.url_helpers
@@ -14,6 +19,8 @@ module DfcProvider
                     :check_enterprise,
                     :check_user,
                     :check_accessibility
+
+      respond_to :json
 
       def index
         products = @enterprise.
--- a/engines/dfc_provider/app/serializers/dfc_provider/product_serializer.rb
+++ b/engines/dfc_provider/app/serializers/dfc_provider/product_serializer.rb
@@ -1,3 +1,7 @@
+# frozen_string_literal: true
+
+# Serializer used to render the products passed
+# into JSON-LD format based on DFC ontology
 module DfcProvider
   class ProductSerializer
     def initialize(enterprise, products, base_url)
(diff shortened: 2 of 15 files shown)


pacodelaluna:
First steps, still need to add specs, refactor the serializer with a better solution, and implement the authentication layer, but we have a base now.

engines/dfc_provider/dfc_provider.gemspec
```diff
@@ -0,0 +1,22 @@
...
+
+# Maintain your gem's version:
+require "dfc_provider/version"
+
+# Describe your gem and declare its dependencies:
+Gem::Specification.new do |spec|
+  spec.name        = 'dfc_provider'
+  spec.version     = DfcProvider::VERSION
+  spec.authors     = ['Admin OFF']
+  spec.email       = ['admin@openfoodfrance.org']
```
luisramos0, line 13: I wouldnt add OFF specific contact here, I'd make this generic like in the other engines:
  s.authors     = ["developers@ofn"]

engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+module DfcProvider
+  module Api
+    class ProductsController < ::ActionController::Metal
+      include ActionController::Head
+      include AbstractController::Rendering
+      include ActionController::Rendering
+      include ActionController::Renderers::All
+      include ActionController::MimeResponds
+      include ActionController::ImplicitRender
+      include AbstractController::Callbacks
```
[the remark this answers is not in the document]
pacodelaluna: Ah right, actually I started with less and add more and more code slowly :) I will update this.

engines/dfc_provider/app/serializers/dfc_provider/product_serializer.rb
```diff
@@ -0,0 +1,38 @@
...
+        {
+          "DFC:description" => variant.name,
+          "DFC:quantity" => variant.total_on_hand,
+          "@id" => variant.id,
+          "DFC:hasUnit" => { "@id" => "/unit/#{variant.unit_description.presence || 'piece'}" }
+        }
+      end
+    end
+  end
+end
```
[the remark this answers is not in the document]
pacodelaluna: We will need to work on the serializer for sure, but I would suggest to wait being sure which one to use, as we need something supporting the JSON-LD format too. Maybe I can take a look at this in a next PR.

engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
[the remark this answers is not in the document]
pacodelaluna: Ok not working because of this:
´´´
include CanCan::ControllerAdditions
include Spree::Core::ControllerHelpers::Auth
´´´
It is imposing an authentication layer which is not used on this API.

engines/dfc_provider/app/serializers/dfc_provider/product_serializer.rb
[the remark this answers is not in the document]
pacodelaluna: This gem maybe, https://github.com/ruby-rdf/json-ld, but it needs further inceptions.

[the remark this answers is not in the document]
luisramos0: yeah, for the LD side of things you can use whatever you think is best, there should be a few nice gems for that.
I was talking about a vanilla json serializer that we could use everywhere in the app, like the gems mentioned in the discourse thread above.

pacodelaluna:
I have deployed this branch to FR staging for further tests.

pacodelaluna:
@luisramos0 I have made a last change to this PR to avoid the enterprise reference inside the endpoint as the DFC application doesn't know this id. It will select the first enterprise related to the user. It will be enough for the POC version and we will see later if we need to extedn this.

The new endpoint is `/api/dfc_provider/enterprise/products.json`.

luisramos0:
I move this to code review so it goes through the normal process.

pacodelaluna:
It was tested on FR stage server, it goes well, the data are processed on DFC application.

engines/dfc_provider/config/routes.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+DfcProvider::Engine.routes.draw do
+  namespace :api do
+    scope :dfc_provider, as: :dfc_provider, path: '/dfc_provider' do
+      resource :enterprise, only: :none do
+        resources :products, only: %i[index]
```
Matt-Yorkley, line 7: For consistency I'd go with `only: [index]` instead of `only: %i[index]` here.

engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+# Controller used to provide the API products for the DFC application
+module DfcProvider
+  module Api
+    class ProductsController < ::ActionController::Metal
```
Matt-Yorkley, line 6: I'm wondering if this whole controller could be a lot cleaner if it inherited from `ActionController::Base` instead of `Metal`. Basically `ActionController::Metal` is a stripped-down version of `ActionController::Base`, with all of these bits like `ActionController::MimeResponds` removed.

What do you think?

Matt-Yorkley, line 6: I'm wondering if this whole controller could be a lot cleaner if it inherited from `ActionController::Base` instead of `Metal`? Basically `ActionController::Metal` is a stripped-down version of `ActionController::Base`, with all of these bits like `ActionController::MimeResponds` removed.

What do you think?

engines/dfc_provider/app/serializers/dfc_provider/product_serializer.rb
```diff
@@ -0,0 +1,37 @@
...
+    def serialized_json
+      {
+        "@context" =>
+        {
+          "DFC" => "http://datafoodconsortium.org/ontologies/DFC_FullModel.owl#",
+          "@base" => @base_url
+        },
+        "@id" => "/enterprise/products",
+        "DFC:supplies" => serialized_products
+      }.to_json
```
Matt-Yorkley, line 21: In this `#serialized_json` method, `#to_json` is called explicitly, and then we're calling `render json: products_json` as well in the controller. I don't think we need to do both, `render json:` will already do that, right? So I think we can remove `#to_json` here.

engines/dfc_provider/spec/controllers/dfc_provider/api/products_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      context 'with an authenticated user' do
+        before do
+          allow_any_instance_of(DfcProvider::AuthorizationControl)
+            .to receive(:process)
+            .and_return(user)
+        end
+
+        context 'with an enterprise' do
+          before { get :index }
```
Matt-Yorkley, line 34: There's a lot of overlapping `before` blocks in this spec. I think we normally have these `#get` calls inside the test blocks, like:

```ruby
context 'with an enterprise' do
   it 'is successful' do
      get :index
      expect(response.status).to eq 200
   end
```

engines/dfc_provider/spec/controllers/dfc_provider/api/products_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+          it 'is successful' do
+            expect(response.status).to eq 200
+          end
+
+          it 'renders the related product' do
+            expect(response.body)
+              .to include("\"DFC:description\":\"#{product.variants.first.name}\"")
```
Matt-Yorkley, line 42: If this test is just checking that the specific product has been included by making sure it's name is present in the json (and not testing the whole json output formatting generally), we could just do:
`.to include product.variants.first.name`
or 
`.to match product.variants.first.name`

engines/dfc_provider/spec/controllers/dfc_provider/api/products_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+require 'spec_helper'
+
+describe DfcProvider::Api::ProductsController, type: :controller do
+  render_views
+
+  let(:user) { create(:user) }
+  let(:enterprise) { create(:distributor_enterprise, owner: user) }
+  let(:product) do
+    create(:simple_product, supplier: enterprise )
```
Matt-Yorkley, line 11: This fits on one line, right?

engines/dfc_provider/spec/controllers/dfc_provider/api/products_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+          end
+
+          it 'renders the related product' do
+            expect(response.body)
+              .to include("\"DFC:description\":\"#{product.variants.first.name}\"")
+          end
+        end
+
+        context 'without a recorded enterprise' do
+          let(:enterprise) { create(:enterprise) }
```
Matt-Yorkley, line 47: This `#let` call without the bang (eg: `#let!`) will not create or persist this enterprise object until it's directly called, and I'm not sure it's not being called anywhere in this test..?

pacodelaluna: Actually it is called inside the `let!(:visible_inventory_item)` call, so we are good.

luisramos0:
Moving to In Dev until review feedback is addressed.

engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
pacodelaluna: Yes actually, working fine with it.

engines/dfc_provider/app/serializers/dfc_provider/product_serializer.rb
pacodelaluna: Well, the `serialized_json` method is supposed to return JSON so I think it is good to keep it this way, no? Or should I rename the method? `serialized_hash` ?

engines/dfc_provider/spec/controllers/dfc_provider/api/products_controller_spec.rb
pacodelaluna: Ok, why not.

engines/dfc_provider/spec/controllers/dfc_provider/api/products_controller_spec.rb
pacodelaluna: I am used to put all the test setup outside the test, but if you prefer I can put the page call inside.

engines/dfc_provider/app/serializers/dfc_provider/product_serializer.rb
pacodelaluna: Ok, I have put `serialized_hash` as it is not JSON now.

engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
pacodelaluna: Same same :)

pacodelaluna:
I have implemented all the suggestions, I have also changed a bit the endpoint to support more cases with `/api/dfc_provider/enterprises/:entreprise_id/products.json`, where you can put `entreprise_id` to `default` in order to catch the first entreprise related to the user.

pacodelaluna:
Ah yes, I understand the issue, so we agree on rebasing only once the code is validated, right? By the way, I am force-pushing but always using new commits.

luisramos0:
thanks. yes, rebase and clean commit history after PR approval. or just leave as is if it looks ok.

-- By the way, I am force-pushing but always using new commits.

interesting, if you dont change the initial commits the force-push will not re-write the initial commits (you can see this in github because the old commits will not come down to after your comment), it looks like you have changed you initial commits. It's probably related to how you are doing it with git.

pacodelaluna:
What I wanted to say is that you will see the whole list as I am rebasing, but I am using new commits for new changes, so you can see the difference with new commits from the last list. I won't do this anymore, no rebase before approval, sounds much more readable.

luisramos0:
ah, I understand! But I dont know what was the last commit the last time I reviewed.

pacodelaluna:
@luisramos0 What is the process once the PR is accepted? Who is in charge of mergng it to master? The reviewer or the owner or someone else?

luisramos0:
After 2 code review approvals, this goes to Test ready, I just added the dev-test ready so that a dev (other than thee author) can test this in staging (I think trying out the new endpoint will be enough.
After that the issue is moved to Ready to Go and any core dev can merge to master at that point.

pacodelaluna:
Ok I see, is this process written somewhere? I was checking the wiki but I couldn't find it, so I was thinking to update it on this wiki, https://github.com/openfoodfoundation/openfoodnetwork/wiki/Making-a-great-pull-request, please tell me if I should.

luisramos0:
good question! It's written here:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/The-process-of-review,-test,-merge-and-deploy
It was a bit outdated, I updated it now, thanks for asking!

engines/dfc_provider/app/controllers/dfc_provider/api/products_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        return if @enterprise.present?
+
+        head :not_found
+      end
+
+      def check_authorization
+        return if access_token.present?
+
+        head :unprocessable_entity
```
luisramos0, line 46: I wonder if 401 wouldnt be more appropriate here.

engines/dfc_provider/app/services/dfc_provider/authorization_control.rb
```diff
@@ -0,0 +1,31 @@
...
+
+# Service used to authorize the user on DCF Provider API
+# It controls an OICD Access token and an enterprise.
+module DfcProvider
+  class AuthorizationControl
+    def initialize(access_token)
+      @access_token = access_token
+    end
+
+    def process
```
luisramos0, line 11: looking at it in more detail, this method name can be easily improved, this is getting the user from the token.

I'd call it find_ofn_user

```ruby
def find_ofn_user
      decode_token
      Spree::User.where(email: @payload['email']).first
end
```

luisramos0:
ok, here are my test instructions and notes:

- First go to Admin > Enterprise and open the enterprise you want to get the catalog for, find the enterprise id there, in this case the id was 4
- then open Users tab of the enterprise and make sure the user you want to use for testing is a manager of the enterprise (super admin wont see the data!), in this case the user was: admin@openfoodfrance.org
- then go to Admin > Users and find the user you want to use for testing: get the user api key or generate a new one, in my case it was 592a53fc3bbf70af65f2d2074b56bc94d20d4ac62c44df6b
- Open a rails console and type:
```ruby
require 'jwt'
payload = { email: 'admin@openfoodfrance.org' }
token = JWT.encode payload, nil, 'none'
puts token
``` 
in this case it printed eyJhbGciOiJub25lIn0.eyJlbWFpbCI6ImFkbWluQG9wZW5mb29kZnJhbmNlLm9yZyJ9.

- open a terminal window and type:
`curl -v -H "X-Spree-Token: 592a53fc3bbf70af65f2d2074b56bc94d20d4ac62c44df6b" -H "Content-Type: application/json" -H "Authorization: Bearer "eyJhbGciOiJub25lIn0.eyJlbWFpbCI6ImFkbWluQG9wZW5mb29kZnJhbmNlLm9yZyJ9."" https://staging.openfoodfrance.org/api/dfc_provider/enterprises/default/products`

- Verify that you get a jsonLD document with the products of that enterprise :tada:

PR #5235, opened 2020-04-15
Remove smtp config from admin config page

### LINKED ISSUE #1534: Email sending fails after changing mail method
**Scenario:** An OFN instance is configured to use a normal email account to send emails. The password of that email account is changed. The OFN admin has to update the password within OFN as well.

**Problem:** After this change, notification emails fail.

**Background:**
Emails are sent using a mail method configured via Spree (Admin, Configuration, Mail Methods). An admin can change the mail method. While the Rails application knows about this change, the independent Delayed::Job processes don't notice. They continue to use the old mail method which may not be working any more.

**Solution:**
The Delayed::Job processes should reload the Spree configuration. I'm not sure what the trigger should be though. Some possibilities:

- Always reload the configuration before using it. This is the safest method, but it could also be slow.
- Reload once hour. This could be another scheduled Delayed::Job. You still have to wait in hour to see the effect of a change.
- Reload on failure. This would be good for recovering in case Spree configuration doesn't work. But it won't update if the old configuration doesn't cause an error. For example, if you want to use a different email address, but the old one is still valid.

Other ideas:

- We could add a "Restart delayed job" button to the admin interface.
- We could add a refresh-config job after changing the config.

luisramos0:
#### What? Why?

Closes #1534

This is a quick fix to 1534. We remove the fields because they need dev intervention: start/stop delayedjob.
Before:
image
After:
image

These values are updated on every deploy through the seeds task. The values are set in the secrets file used by ofn-install. This needs to be tested in staging as part of this PR.

#### What should we test?
We are still able to use the existing fields in this form and they still work.
As part of this PR we need to test that ofn-install will update the smtp password in the preferences table during a normal deploy.

#### Release notes
Changelog Category: Changed
Remvoed mail settings config from the UI as they can only be changed by developers.


kristinalim:
This PR needs both `dev-test` and user test. Tagging this `dev-test` first.

sauloperez:
UI-wise things work but I'm wondering about 

> One thing I can add in #5235 is to remove mail methods from spree_preference in a migration.

luisramos0:
thanks for mentioning that. I forgot about that.
I rebased to resolve conflicts in db/schema and added the migration.

db/migrate/20200429122446_drop_mail_methods.rb
```diff
@@ -0,0 +1,16 @@
+class DropMailMethods < ActiveRecord::Migration
+  def up
+    drop_table :spree_mail_methods
+
+    # delete mail_method preferences associated with the old MailMethod model
+    execute "DELETE FROM spree_preferences WHERE key LIKE 'spree/mail_method%'"
```
sauloperez, line 6: this will skip smtp_username and smtp_password though. Is that what we want?

sauloperez:
Confirmed it works. 

```
openfoodnetwork=> select * from spree_preferences where key ilike '%mail_host%';
-[ RECORD 1 ]---------------------------------
id         | 2
value      | smtp.mandrillapp.com
created_at | 2019-07-23 10:01:26.371498
updated_at | 2020-04-29 11:43:10.463866
key        | spree/app_configuration/mail_host
value_type | string

openfoodnetwork=> delete from spree_preferences where key ilike '%mail_host%';
DELETE 1
openfoodnetwork=> select * from spree_preferences where key ilike '%mail_host%';
 id | value | created_at | updated_at | key | value_type 
----+-------+------------+------------+-----+------------
(0 rows)

openfoodnetwork=> select * from spree_preferences where key ilike '%mail_host%';
-[ RECORD 1 ]---------------------------------
id         | 502
value      | smtp.mandrillapp.com
created_at | 2020-04-29 13:58:04.582884
updated_at | 2020-04-29 13:58:04.582884
key        | spree/app_configuration/mail_host
value_type | string
```

db/migrate/20200429122446_drop_mail_methods.rb
luisramos0: yes, I think smtp_username and smtp_password are the values set in MailConfiguration and used in MailSettings.

luisramos0:
great, thanks Pau!

db/migrate/20200429122446_drop_mail_methods.rb
sauloperez: ah ok, got it. These are old Spree preference and not ours. I wasn't sure what you were aiming at.

filipefurtad0:
Hi @sauloperez ,

Tested the remaining fields:

- send email
- send copy
- intercept email

All work fine - ready to go.

luisramos0:
oh, this was quick!!!

Issue #5242, opened 2020-04-16
Notify producer email behaving in an overzealous manner


RachL:
## Description

Clicking the notify producer button can send 8 to 13 times the same email :boom:

## Steps to Reproduce

 In French production it is happening with hub 1395

## Animated Gif/Screenshot

![image](https://user-images.githubusercontent.com/6525576/79462356-aa5b9600-7ff7-11ea-8e61-22edce980f09.png)

## Severity

If this can damage our email reputation I would say s2 :(

## Your Environment
* Version used: v2.9.2
* Browser name and version: Firefox 76
* Operating System and version (desktop or mobile): Desktop

sauloperez:
We've seen errors like https://app.bugsnag.com/open-food-france/open-food-france-prod/errors/5e94a9c53d908b0017b28cfc?filters%5Bevent.since%5D%5B%5D=1d&filters%5Berror.status%5D%5B%5D%5Btype%5D=eq&filters%5Berror.status%5D%5B%5D%5Bvalue%5D=open&filters%5Bsearch%5D%5B%5D%5Btype%5D=eq&filters%5Bsearch%5D%5B%5D%5Bvalue%5D=Job&event_id=5e9862d50059ecf518a00000 since April 13th. My bet is that this failure causes DJ to retry the job and thus, emailing again recipients that did get the email successfully.

sauloperez:
Confirmed, a job for the hub 1395 is the last one that failed.

![Screenshot from 2020-04-16 16-32-28](https://user-images.githubusercontent.com/762088/79468993-ec88d580-7fff-11ea-844d-a87cf67fc68b.png)

sauloperez:
Just to be clear, the OrderCycleNotificationJob is enqueued once but delivers as many emails as suppliers. If any of these fail, DJ retries the entire job. That's why successful recipients get spammed while the failed ones keep failing.

luisramos0:
Nice one Pau.

>>  we should enqueue a job per supplier

Yes, very easy. Basically moving the to first lines (each) in OrderCycleNotificationJob#perform to OrderCyclesController#notify_producers

PR #5243, opened 2020-04-16
Temporally disable OC notification retries

sauloperez:
#### What? Why?

This stops DJ from retrying this job when failed. I suspect this is the root cause of https://github.com/openfoodfoundation/openfoodnetwork/issues/5242.

This changes the logs **for this job** from

```
2020-04-16T14:40:59+0000: [Worker(delayed_job host:staging.katuma.org pid:4446)] Job OrderCycleNotificationJob (id=236046) RUNNING                                                             
2020-04-16T14:40:59+0000: [Worker(delayed_job host:staging.katuma.org pid:4446)] Job OrderCycleNotificationJob (id=236046) FAILED (2 prior attempts) with RuntimeError: Pau debugging 
```
to
```
2020-04-16T14:42:33+0000: [Worker(delayed_job host:staging.katuma.org pid:4853)] Job OrderCycleNotificationJob (id=236047) RUNNING
2020-04-16T14:42:33+0000: [Worker(delayed_job host:staging.katuma.org pid:4853)] Job OrderCycleNotificationJob (id=236047) FAILED (0 prior attempts) with RuntimeError: Pau debugging
2020-04-16T14:42:33+0000: [Worker(delayed_job host:staging.katuma.org pid:4853)] Job OrderCycleNotificationJob (id=236047) FAILED permanently because of 1 consecutive failures
```

#### What should we test?

This should be deployed ASAP.

### DIFF
--- a/app/jobs/order_cycle_notification_job.rb
+++ b/app/jobs/order_cycle_notification_job.rb
@@ -6,4 +6,9 @@ OrderCycleNotificationJob = Struct.new(:order_cycle_id) do
       ProducerMailer.order_cycle_report(supplier, order_cycle).deliver
     end
   end
+
+  # Temporally disable DJ's retries until we fix https://github.com/openfoodfoundation/openfoodnetwork/issues/5242.
+  def max_attempts
+    1
+  end
 end


sauloperez:
Already deployed

Matt-Yorkley:
Thoughts:
- Have you fully considered the potential consequences would this have for other jobs, for example:  subscriptions placements? 
- Can't we just fix the actual reason for the failure in #5242 (which we now know, and should be very easy)? 
- We want failed jobs to be retried until they pass, right?

sauloperez:
> Have you fully considered the potential consequences would this have for other jobs, for example: subscriptions placements?

It shouldn't have any. This only applies to this job. I clarified the description.

Matt-Yorkley:
> This only applies to this job.

Okay :+1:

Disabling it on FR and AU might be a good idea for now, but I don't think this should be merged. We should fix the specific problem, which is:
- a deleted variant of a line_item is not being found correctly in the mail view
- this triggers an error during the HTML rendering
- the email _is actually sent_ but DelayedJob thinks the task failed, so keeps retrying it

If an email fails to be sent for any other reason, like the DNS lookup for Sendgrid temporarily failing (which happened the other week on FR Prod) we still definitely want to retry sending the email.

sauloperez:
sure, it was just a temporal measure but I wouldn't mind merging as long as we revert it back when the issue is fixed. In the end, this represents the current state of those two servers and it could also be needed in all other instances.

kirstenalarsen:
I have just been alerted that this is still happening! @mkllnk could you please make sure this patch is deployed to aus asap? if it is not already?

mkllnk:
That's weird. A user reported that this is happening again but the patch is deployed. There were two new jobs which failed and both had only one attempt. I removed the jobs just to be sure but now we don't have any data to look at.

sauloperez:
Just to clarify this avoids having a second attempt but not the first, otherwise, no one would receive any OC notification, even those that don't have any problem.

kirstenalarsen:
User reports that it is this eight producers - she got each email twice today - Fleurieu Milk, Alnda Farm, Feather & Peck, Falkai Farm, Fleurieu garlic, Two Hills & Creek, Salad Greens & Kitchen Herbs, The Garden Farmers. Looking into the setup of these - is very strange

I'm currently checking out the owners / managers etc

sauloperez:
Do we know what's the relationship of the reporting user with these producers? It'd be great if we could know the ids of the enterprises and order cycle.

Either someone clicked on notifying producers again or we're still seeing attempts of a job that got enqueued before we deployed this fix (quite unlikely). The last potential cause could be that we are no longer running this patch in AU but that'll be the case when we deploy v2.9.3, which didn't happen yet.

sauloperez:
The fix at https://github.com/openfoodfoundation/openfoodnetwork/pull/5244 landed in production with v2.9.4 so this is no longer needed.

PR #5244, opened 2020-04-16
Include deleted products in OC mailer

### LINKED ISSUE #5242: Notify producer email behaving in an overzealous manner

sauloperez:
#### What? Why?

Closes #5242

This includes products and variants that have been soft-deleted so we show them in the OC supplier email notification even if they got deleted after the user purchase.

The root cause is a known Rails bug. See 63d04eace1a697447fb2e71bc2a340acd33e2fa2 for details.

Both in FR and AU the root cause of the last error was a deleted variant  (I haven't checked each of the occurrences of this error).

#### What should we test?

We need to check that we can notify suppliers of an order cycle where at least a variant got deleted after the order was placed. For that, place the orders, delete one of the purchased variants and then, click on "Notify producers" from the OC edit page. Then, try out the same but deleting a product instead.

In both cases, we shouldn't see any error related to it in either Bugsnag or in `/admin/delayed_job/failed`.

#### Release notes

Fix order cycle producer notification failing when a product or variant got deleted after the order was placed.

Changelog Category: Fixed

### DIFF
--- a/app/mailers/producer_mailer.rb
+++ b/app/mailers/producer_mailer.rb
@@ -2,28 +2,31 @@ class ProducerMailer < Spree::BaseMailer
   include I18nHelper
 
   def order_cycle_report(producer, order_cycle)
-    @producer = producer
-    @coordinator = order_cycle.coordinator
-    @order_cycle = order_cycle
-    line_items = line_items_from(@order_cycle, @producer)
-    @grouped_line_items = line_items.group_by(&:product_and_full_name)
-    @receival_instructions = @order_cycle.receival_instructions_for @producer
-    @total = total_from_line_items(line_items)
-    @tax_total = tax_total_from_line_items(line_items)
+    unscoping do
+      @producer = producer
+      @coordinator = order_cycle.coordinator
+      @order_cycle = order_cycle
 
-    I18n.with_locale valid_locale(@producer.owner) do
-      order_cycle_subject = I18n.t('producer_mailer.order_cycle.subject', producer: producer.name)
-      subject = "[#{Spree::Config.site_name}] #{order_cycle_subject}"
+      line_items = line_items_from(@order_cycle, @producer)
+      @grouped_line_items = line_items.group_by(&:product_and_full_name)
+      @receival_instructions = @order_cycle.receival_instructions_for @producer
+      @total = total_from_line_items(line_items)
+      @tax_total = tax_total_from_line_items(line_items)
 
-      return unless has_orders?(order_cycle, producer)
+      I18n.with_locale valid_locale(@producer.owner) do
+        order_cycle_subject = I18n.t('producer_mailer.order_cycle.subject', producer: producer.name)
+        subject = "[#{Spree::Config.site_name}] #{order_cycle_subject}"
 
-      mail(
-        to: @producer.contact.email,
-        from: from_address,
-        subject: subject,
-        reply_to: @coordinator.contact.email,
-        cc: @coordinator.contact.email
-      )
+        return unless has_orders?(order_cycle, producer)
+
+        mail(
+          to: @producer.contact.email,
+          from: from_address,
+          subject: subject,
+          reply_to: @coordinator.contact.email,
+          cc: @coordinator.contact.email
+        )
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


luisramos0:
:+1: you logic makes sense but we need to explain the error on absent line_item.variant while the solution is joining deleted products.

We are only unscoping variants in line_items_decorator with the accessor method: def variant
Do the scopes see this? I dont think so. If that is correct, if you do Spree::LineItem.includes(variant:) as in this mailer, you dont get the deleted variants. I think this would also mean your solution is not enough to fix the bugsnag error. We would need some form of  **includes(variant   "with deleted".**
I hope this is helpful.

sauloperez:
I thought so too but adding a byebug I could see it going through the line_item_decorator#variant. I'll dig deeper.

sauloperez:
@luisramos0 I found the root cause but we need to agree on a solution. This is a tricky one. Check out 63d04eace1a697447fb2e71bc2a340acd33e2fa2 for details

app/models/spree/line_item_decorator.rb
```diff
@@ -131,7 +131,9 @@ def display_included_tax
     Spree::Money.new(included_tax, currency: currency)
   end
 
-  delegate :display_name, to: :variant
+  def display_name
+    variant.display_name
+  end
```
sauloperez, line 136: :hankey: we don't need this change. It simply changes the RuntimerError to a NilClass error

luisramos0:
ah, nice, so my hint was on the right direction.

I think this can have implications in a few other places in the app when handling deleted products and deleted variants. Maybe we need to do this trick in some other places!?

sauloperez:
> Maybe we need to do this trick in some other places!?

Yes, that's what I meant. There must be plenty of places where we need this so I say we investigate a couple of potentials solutions after this is merged so as not to delay this patch.

I'll undo the draft and push the final version. I hope we can sneak it into the release.

app/mailers/producer_mailer.rb
```diff
@@ -2,28 +2,31 @@ class ProducerMailer < Spree::BaseMailer
...
-      return unless has_orders?(order_cycle, producer)
-
-      mail(
-        to: @producer.contact.email,
-        from: from_address,
-        subject: subject,
-        reply_to: @coordinator.contact.email,
-        cc: @coordinator.contact.email
-      )
+    unscoping do
```
luisramos0, line 5: The code inside this unscoping should be extracted to a method. this method is already too long.

Maybe a better name would be   with_unscoped_products_and_variants?

sauloperez: Agree. This is totally awful.

sauloperez:
I took the chance to remove an extra N+1. There still some re User, Enterprise and Adjustment but I want to finish today :trollface:

app/mailers/producer_mailer.rb
Matt-Yorkley: I've just run into a similar problem in another PR, and applying this `with_unscoped_products_and_variants` technique fixed the issue...

Matt-Yorkley:
I've just been looking in Datadog. All the previous bad endpoints look great, and this one is near the top of the list now, so it's great you removed the N+1 :tada: There's also a lot of runtime errors for the missing line items. We should get this in production!

sauloperez:
Yes, we definitely need this in production ASAP. I could prepare v2.9.4 later today since v2.9.3 already got published.

filipefurtad0:
Hi @sauloperez ,

First, I verified the bug - in AU-Staging. Triggered bugsnag by notifying producers, after deleting a variant from an already made/paid order, from a subscription. At first, I did not get any email. I repeated the process, for deleting a product. Ever since, I keep getting these ghost-emails... So, let's just say I've met the bug :-)

Staged the PR in FR-staging, and repeated the process. I added products from two different producers. All good with email notifications before, and after deleting i) a variant ii) a product, from one of the producers. They both receive the notifications triggered from the OC.

Looks good to me - moving to ready to go.

PR #5247, opened 2020-04-16
Added placeholders in variant edit page

### LINKED ISSUE #3793: Add placeholder in variant edit page
As suggested by @CathyPearce while testing #3788

## Description
**- As a:** manager or super admin
**- On page:** Variant edit page /admin/products/PRODUCTNAME/variants
**- I want to be able to do:** 
Add placeholders on Display name and Display as fields. The idea is to have the same behavior than in product bulk edit page where we add as placeholder the value of the main product.

This will help the manager to know which variant they are editing.

## Acceptance Criteria
1. Log in as an admin
2. The behavior on the bulk product page should not change, meaning, when adding variants you can see placeholders with the value from the main product. Example :

3. When using the variant edit page, the current behaviour is with no placeholder :

and we would like to have a placeholder in (in this example) the display as field :

4. If nothing is written in the display name, we should see a placeholder as well

chrishil1:
#### What? Why?

Closes #3793 

Added requested placeholders for Display Name and Display As to variant edit page as "eg Tomatoes" and "eg 2 kg" respectively. Can be easily changed to any other message of interest in en.yml.

#### What should we test?
Variant edit page has these placeholders that disappear once user enters input to text field.

#### Release notes
Added placeholders to variant edit page for clarification when user is inputting data.

Changelog Category: Added

Changed Form file is located under the Spree subfolder in views directory.

### DIFF
--- a/app/views/spree/admin/variants/_form.html.haml
+++ b/app/views/spree/admin/variants/_form.html.haml
@@ -1,10 +1,10 @@
 .label-block.left.six.columns.alpha{'ng-app' => 'admin.products'}
   .field
     = f.label :display_name, t('.display_name')
-    = f.text_field :display_name, class: "fullwidth", placeholder: t('admin.products.display_name_placeholder')
+    = f.text_field :display_name, class: "fullwidth", placeholder: t('.display_name_placeholder')
   .field
     = f.label :display_as, t('.display_as')
-    = f.text_field :display_as, class: "fullwidth", placeholder: t('admin.products.display_as_placeholder')
+    = f.text_field :display_as, class: "fullwidth", placeholder: t('.display_as_placeholder')
 
   - if product_has_variant_unit_option_type?(@product)
     - if @product.variant_unit != 'items'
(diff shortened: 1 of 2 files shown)


app/views/spree/admin/variants/_form.html.haml
```diff
@@ -1,10 +1,10 @@
 .label-block.left.six.columns.alpha{'ng-app' => 'admin.products'}
   .field
     = f.label :display_name, t('.display_name')
-    = f.text_field :display_name, class: "fullwidth"
+    = f.text_field :display_name, class: "fullwidth", placeholder: t('admin.products.display_name_placeholder')
```
luisramos0, line 4: you should use relative paths, this should be
 placeholder: t('.display_name_placeholder')

and then you need to add the keys in en.yml in line 3381 and 3382, which is path spree.admin.variants.form (the path of this haml file)

chrishil1: Just made these changes.

filipefurtad0:
Hi @chrishil1 ,

I staged this and checked whether this appears, as described. Looking good:

image


Thank you for this improvement!
Ready to go.

PR #5251, opened 2020-04-17
Change Result of PriceSack Calculation from Integers to Floats, clone #4812

luisramos0:
See #4812

### DIFF
--- a/app/models/spree/calculator/price_sack_decorator.rb
+++ b/app/models/spree/calculator/price_sack_decorator.rb
@@ -17,9 +17,9 @@ module Spree
       order_amount = line_items_for(object).map { |x| x.price * x.quantity }.sum
 
       if order_amount < min
-        cost = preferred_normal_amount.to_i
+        cost = preferred_normal_amount.to_f
       elsif order_amount >= min
-        cost = preferred_discount_amount.to_i
+        cost = preferred_discount_amount.to_f
       end
 
       cost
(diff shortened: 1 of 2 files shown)


sauloperez:
I'm taking the shortcut and moving to Test Ready, considering how much time the original PR has been sitting there and how trivial the fix is.

filipefurtad0:
Hi @blainebillings ,

After this PR, PriceSack settings accept and display non-integers. Depending on the validation of the boundary condition these appear as expected in checkout, and are not rounded any more.

Considering:

- Minimal amount = 10
- Normal amount = 5.9
- Discount amount = 1.9

An order amount of 15 now displays the Discount amount of 1.9, while an order amount of 8 displays the Normal amount of 5.9.

Awesome, thank you for this improvement! Think this is ready to go :tada: 

However, I think this PR and the RSpec unit testing do not consider the case in which the Minimal amount (boundary condition) is a float.

I don't think this should be a big issue as probably users will not be so inclined to set floats as Minimal amounts. Perhaps a future improvement. Anyway, some further test cases below.

Considering a constant value of Minimal amount = 16, and the order amount values below:

Order Amount = 15.9 
The displayed shipping cost is 5.9, which is the expected value (as 15.9 < 16.9).

Order Amount = 16 
The displayed shipping cost is 1.9, while it rather should be 5.9 (as 16 < 16.9)

Order Amount = 16.1 
The displayed shipping cost is 1.9, while it rather should be 5.9 (as 16.1 < 16.9)

RachL:
@filipefurtad0 I think that if this PR improves the current situation, we can ship it, and create a second issue to improve the rest. It's really in the case where we make the situation worse with a PR that we need to continue working on it.

filipefurtad0:
Thank you for your advice here @RachL. It surely improves situation for what it was created, so I'll open new issue concerning the remaining improvements.

luisramos0:
great testing Filipe!
Yeah, minimal amount has not been considered...

Issue #4186, opened 2019-08-23
Order data inconsistency between Line Items and Inventory Items (was: snail in admin edit orders)


mkllnk:
## Description

Some orders can't be edited by an admin. The Bugsnag error suggests that a variant got removed from an order (I guess it was out of stock) but the order's manifest is still referencing the variant. The view displaying the edit order screen fails to find the line item for the missing variant and crashes trying to access it.

## Expected Behavior

You can always edit an order as admin.

## Actual Behaviour

We can't edit at least one order.

## Steps to Reproduce

1. Visit https://openfoodnetwork.org.au/admin/orders/R451317181/edit

## Context

> Customer says they tried to order 3 of a product, but there were only 2 available, so they rectified this in the final screen before paying, then 'got kicked out' and had to start again. When they started again they found that the first part of the order was still there, finished the order, made payment and thought it went through fine. Didn't receive confirmation email though, and payment didn't go through. (Went to pick up box that week and it wasn't there).

That order is in `address` state and can't be edited.

I also think that variants removed from the cart because they are out of stock, causes this error: https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5c625e38b659cb00191b1307

## Severity

bug-s2: a non-critical feature is broken, no workaround
## Your Environment

* Version used: v2.2.2
* OFN Platform instance: OFN Australia

## Bugsnag

**ActionView::Template::Error** in **spree/admin/orders#edit**
undefined method `single_money' for nil:NilClass

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5e7c2499183e6e0018839df6)

## Stacktrace

    app/views/spree/admin/orders/_shipment_manifest.html.haml:10 - block in _0864f0f33beb8e759403eadb8313605c
    app/views/spree/admin/orders/_shipment_manifest.html.haml:1 - each
    app/views/spree/admin/orders/_shipment_manifest.html.haml:1 - _0864f0f33beb8e759403eadb8313605c
    app/views/spree/admin/orders/_shipment.html.haml:34 - _bc7e946539bd48bcf6974a1142286d17
    app/views/spree/admin/orders/_form.html.haml:5 - _9fb4e2d06260a01ce10cc4614bd31893
    app/views/spree/admin/orders/edit.html.haml:30 - _aa3ef3a74b75f7d5a0de15cf8b658c89

Matt-Yorkley:
[...]

PR #5253, opened 2020-04-17
Make edit order page work even if inventory_items dont have a corresponding line_item in the order

luisramos0:
#### What? Why?

Improves #4186 as the user will be able to use the edit order page without errors.
The data inconsistency and the underlying problem remains unknown and unresolved.

I have done a big refactoring of order_spec, it was needing one! The actual change is very small.

#### What should we test?
I can create the situation in 4186 by deleting the line item of an order in the database and keeping the inventory item in there but I am not sure how to do it with the app only. So, I am not sure how we should test this manually...

#### Release notes
Changelog Category: Fixed
Make edit order page work correctly even if the underlying order data is not consistent in terms of inventory items and line items.


sauloperez:
> I can create the situation in 4186 by deleting the line item of an order in the database

the easiest might be that the tester asks you to do so :shrug:

filipefurtad0:
Hi @luisramos0 !

Thank you so much for access to the FR server and all the help with the SQL commands for the table queries. Amazing stuff.

Here is a description of the process. I verified the bug in katuma- and fr-staging. Then, after staging this PR, proceeded as follows:

1) ofn-app: created a two product order (two line items), as admin/or customer order, #R656741034 - OK

image


2) psql console: checked if the the two line items appear
=> select * from spree_line_items where order_id in (select id from spree_orders where number like '#R656741034');

image

So, two line items visible, 336 and 337, variants 3106 and 3104, respectively. 

3) psql console: deleting one of the line items
=> delete from spree_line_items where id in (select max(id) from spree_line_items where order_id in (select id from spree_orders where number like '#R656741034'));

4) psql console: checking only one of the line items appear (one got deleted) Basically, repeating the command in 2) , item 337 is gone.

image

5) ofn-app: checking whether it is possible to edit the order.

It is indeed possible to edit the order. However I expected the remaining item to appear. Instead, none appears:

image

Shipping the order, I get the following confirmation email, where both items are shown:

image

However, on the customers side, the order appears as it should - with only the expected item remaining 

image


So, in summary:
- the purpose of the PR was to be able to edit the order and this is fulfilled.

However, there are discrepancies:
- no items appear for the Admin under that order
- what is shown in the SQL table matches the order from the customers perspective
- shipping the order sends a Shipment Notification where the deleted item appears again.

I'm labelling this w/ feedback needed.

luisramos0:
Nice testing!

We are testing a data inconsistency scenario in itself. So, not all things will match... we are not fixing the root cause here.

The customer is seeing the correct info because that page is looking at the line items table :+1: 
The shipment email is showing both items because the inventory is still there in the shipment (that's the inconsistency that will be seen in these cases.
But I was expecting you to be able to see the remaining line item it in the admin page :-(
I tried it myself and it's not showing the line item for some reason, in my local tests the behavior was different.

Anyway, this is a little better than without the PR, isn't it? One of the now possible solutions here is to cancel the order.
I am going to merge this.
Tomorrow morning I can investigate this better and make another PR if I find some improvement.

filipefurtad0:
Hey @luisramos0 ,

Thank you for your feedback! Sure, definitively much better this way. I was wondering why the remaining line item did not show up it in the admin page, so I decided to check somewhere else, how it behaved - hopefully it also helps in improving this.

I am happy that it's merged and on time for the release :tada:

luisramos0:
yes, fantastic testing :clap:

luisramos0:
ok Filipe, I have found why you were not seeing the other line item, that is fixed here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5350

filipefurtad0:
Nice! Thank you Luis :+1:

PR #5256, opened 2020-04-18
Fix sorting of orders list

### LINKED ISSUE #5190: Order sort by column header doesn't work
## Description
When you are in the Orders page of backend (e.g. https://openfoodnetwork.org.au/admin/orders ) each column header is hyperlinked, but clicking on them doesn't trigger any action. 

## Expected Behaviour
When you click a column header it should sort by that column. When you click again it should sort in the reverse order. e.g. clicking Shipment State should sort all 'Pending' to the top, clicking again should sort all 'Completed' to the top.

## Actual Behaviour
You click on the header and nothing happens. For some column headers an arrow appears pointing up or down, but this doesn't make anything happen.

## Steps to Reproduce
Go to Order page. Click a column header. Observe nothing happening 😀 

## Workaround
Workaround is to download CSV file and sort, but for some of these that's not a simple sort as the field you want to sort by (e.g. Shipment State) is applied just to order total but not all of the items in the order.

## Severity

## Your Environment
* Version used: v2.9.0
* Browser name and version: Chrome Version 80.0.3987.149
* Operating System and version (desktop or mobile): Mac OSX El Capitan Desktop

dacook:
Fix column sorting for admin/orders. Now when you click a column header, it refreshes the filtered results with the correct sorting (like for Products).

#### What? Why?

Closes #5190 

When you are in the Orders page of backend (e.g. https://openfoodnetwork.org.au/admin/orders ) each column header is hyperlinked, but clicking on them doesn't trigger any action.

There was a wrong copied function call in the coffeescript. Oops.

#### What should we test?

When you click a column header it should sort by that column. When you click again it should sort in the reverse order. e.g. clicking Shipment State should sort all 'Pending' to the top, clicking again should sort all 'Completed' to the top.

Repeat this test with at least 16 orders so that there is more than one page of orders. The sorting should happen across pages.

#### Release notes

Fix column sorting for admin/orders

Changelog Category:  Fixed

### DIFF
--- a/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
@@ -67,7 +67,7 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, $timeout, Reque
     return unless sort && sort.predicate != ""
 
     $scope.sorting = sort.getSortingExpr()
-    $scope.fetchProducts()
+    $scope.fetchResults()
   , true
 
   $scope.capturePayment = (order) ->
--- a/spec/javascripts/unit/admin/customers/controllers/customers_controller_spec.js.coffee
+++ b/spec/javascripts/unit/admin/customers/controllers/customers_controller_spec.js.coffee
@@ -20,7 +20,6 @@ describe "CustomersCtrl", ->
       {id: 109, name: "Australia", states: [{id: 55, name: "ACT", abbr: "ACT"}]}
     ]
 
-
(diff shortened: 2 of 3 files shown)


dacook:
I haven't added any tests for this, let me know if required.

dacook:
Great, thanks @luisramos0 , now we've both learnt something!
I guess this is up to the stage of user testing?

luisramos0:
yes, we use zenhub, if you have the plugin you will see Maikel moved to Test Ready which is a state that Testers are looking at, someone will stage this and test it sooon. The PR will be updated at that time :+1:

PR #4790, opened 2020-02-13
4779 payment method display

mkllnk:
#### What? Why?

Closes #4779

If we had multiple failed payments and then a successful payment, the
order confirmation was displaying the payment method of the first failed
payment. That was confusing and is now changed to the last payment
method.

#### What should we test?

Sanity check:

- Check out normally.
- Confirm the "Paid via" display in the order confirmation page.
- Confirm the "Paid via" display in the order confirmation emails (customer and enterprise).

New behaviour:

- Check out with a failing payment. I think there are certain credit card numbers for Stripe that generate a failure.
- Then check out with a cash payment method.
- Confirm the "Paid via" display in the order confirmation page.
- Confirm the "Paid via" display in the order confirmation emails (customer and enterprise).

#### Release notes

Fixed: Order confirmations now display the last used payment method instead of the first.

Changelog Category: Fixed

PR #5257, opened 2020-04-18
StripeSCA - reuse better method to fetch last payment of an order to avoid nasty bugs in the future

luisramos0:
#### What? Why?

In StripeSCA context, reuse better method to fetch last payment introduced in #4790 by @mkllnk 

#### What should we test?
We should repeat the test in #4790 and then:
 verify StripeSCA checkout still works for normal cards and cards that require extra SCA auth, for example, 4242... and 4000 0027 6000 3184.

#### Release notes
Changelog Category: Changed
Improved how stripeSCA handles orders with multiple payments. This will avoid hard to track bugs in the future.

### DIFF
--- a/app/views/spree/shared/_order_details.html.haml
+++ b/app/views/spree/shared/_order_details.html.haml
@@ -13,9 +13,9 @@
     .pad
       .text-big
         = t :order_payment
-        %strong=  OrderPaymentFinder.last_payment_method(order)&.name
+        %strong=  last_payment_method(order)&.name
       %p.text-small.text-skinny.pre-line
-        %em= OrderPaymentFinder.last_payment_method(order)&.description
+        %em= last_payment_method(order)&.description
 
     .order-summary.text-small
       %strong
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -153,10 +153,12 @@ class CheckoutController < Spree::StoreController
   end
 
   def valid_payment_intent_provided?
-    params["payment_intent"]&.starts_with?("pi_") &&
-      @order.state == "payment" &&
-      @order.payments.last.state == "pending" &&
-      @order.payments.last.response_code == params["payment_intent"]
+    return false unless params["payment_intent"]&.starts_with?("pi_")
+
+    last_payment = OrderPaymentFinder.new(@order).last_payment
+    @order.state == "payment" &&
+      last_payment&.state == "pending" &&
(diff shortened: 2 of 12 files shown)


app/views/spree/shared/_order_details.html.haml
```diff
@@ -13,9 +13,9 @@
     .pad
       .text-big
         = t :order_payment
-        %strong=  OrderPaymentFinder.last_payment_method(order)&.name
+        %strong=  last_payment_method(@order)&.name
```
mkllnk, line 16: You substituted `order` with `@order`. Was that intentional. I think it's better to use `order` because it makes clear where the variable comes from and makes it easier to re-use partials (compared to ivars).

mkllnk: Okay, I looked it up. This partial is only used in `app/views/spree/orders/show.html.haml` which also has `@order` defined. So this change doesn't break anything but the partial contains a dozen references to `order` and none to `@order`. So I think it would be a lot clearer to use `order` here. I will change it.

luisramos0:
yes, copy paste error Maikel, thanks for fixing it!

filipefurtad0:
Hi @luisramos0 

I checked for any wrong "memory" effects, concerning failed payments on receipts/emails/order confirmations as in #4803 

After failing a payment (ex. with a card triggering the incorrect_cvc code) and reattempting with cash, everything works and looks as expected, i.e., no discrepancies between the failed and the successful payment method.

image


I verified whether StripeSCA checkout still works for:
- normal cards (Visa, Visa Debit, Mastercard, ES international card)
- cards that require extra SCA auth, ex.: 4000 0027 6000 3184 and 4000003800000446.

All works, all good -> moving to ready to go!

luisramos0:
I saw your notes.
If there is a first payment with stripe that fails and then a checkout with uncaptured cash payment.
The system is assuming the cash payment isnt valid because it's not captured, so it shows the finalized payment that failed. I think you would want to see "Pending payment" instead?

filipefurtad0:
That's it @luisramos0,

For cases 1) and 2) I was expecting something like "Pending payment" or "Balance Due", to be displayed in /orders. I guess the good thing is, that the failed payment appears in the payments table /orders/ordernumber/payments.

For the test case 3) the behaviour is the opposite: the message is correctly displayed in the /orders page (Saldo Pendente, in the pic) but I would expect to see the a failed payment entry on the payments table /orders/ordernumber/payments.

PR #5261, opened 2020-04-18
[Mobile] Make search box font size be 16px so that no zoom happens on iphone

### LINKED ISSUE #4535: [mobile ux] Field focus on iPhones change the zoom level of a page and user needs to resize it
This can be merged directly to master 🎉 

## Expected Behavior
When I do a search of shop products on an iPhone, the browser does not zoom in and resize so that the page no longer fits the screen.

## Actual Behaviour
When I do a search of shop products on an iPhone, the browser zooms in and resizes the page, which means when I click out of the search bar I then need to resize the page back to original in order to continue browsing. This is really annoying!

## Steps to Reproduce
1. Go to a shop on an iPhone (any browser)
2. Click on the search bar
3. See the page zoom in
4. Click out of the search bar, the page remains at the incorrect zoom level.

## Animated Gif/Screenshot
https://airtable.com/tblPoizwAxVfccvOz/viwIhkyM0Fc83qbts/recafWYVSJtNntiZE

## Workaround
Resize the page to fit the screen

## Severity

**bug-s4: it's annoying, but you can use it**

## Your Environment
* Version used: n/a
* Browser name and version: Firefox and Safari, latest versions
* Operating System and version (desktop or mobile): Checked on iPhone 6s and X

## Possible Fix

There are several solutions at https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone. The simplest would be to increase the font size to at least 16px to prevent zooming.

## Here's a video of the wacky!!!
Courtesy of Yuko :P
(Had to zip the video as github says NO to videos!)
[OFN-20200108-mobile-search-field-resize.mov.zip](https://github.com/openfoodfoundation/openfoodnetwork/files/4038531/OFN-20200108-mobile-search-field-resize.mov.zip)

--> CSS

luisramos0:
#### What? Why?

Closes #4535

#### What should we test?
See issue for details.
I tested this on staging ES with my iphone 4, it works (the fix works, I am not sure about my iphone 4 :rofl:)

#### Release notes
Changelog Category: Changed
Make font size bigger on product search to improve UX on iphone

### DIFF
--- a/app/assets/stylesheets/darkswarm/_shop-inputs.css.scss
+++ b/app/assets/stylesheets/darkswarm/_shop-inputs.css.scss
@@ -9,6 +9,9 @@
 
   input#search {
     @include medium-input(rgba(0, 0, 0, 0.3), #777, $clr-brick);
+
+    // avoid zoom on iphone, see issue #4535
+    font-size: 1rem;
   }
 
   // ordering


daniellemoorhead:
It works for me on an iPhone 6s too!

sauloperez:
Let's take a shortcut straight to Test Ready given how simple it is.

daniellemoorhead:
@luisramos0 this works fine on the search bar (as described on the issue), so we can merge this.

However:
- It's also an issue on the log in / sign up / forgot password modal (though not as bad). <- should we create another story for this, or do it as part of this issue?

- It's also an issue on the select quantity field for product listings. <- @mkllnk the select quantity field is changing with the epic you are working on, but it might be a problem depending on how @yukoosawa designs the min/max fields for bulk products. So can you please keep on eye on the focus if it's a field you have to type in please 🙏

luisramos0:
ah, I was not aware of that.
I cant find those extra problems on #4535? Am I missing something?
If they are not mentioned in the issue, I think we should go ahead with PR and issue and open new issues for those other cases.

Issue #5150, opened 2020-04-03
Re-assess disabled browser caching


Matt-Yorkley:
We disabled browser caching to fix a bug a long time ago. We should take a look at reverting it.

Original Issue: #1213
Original PR: #2268

PR #5273, opened 2020-04-20
Strategic Caching

Matt-Yorkley:
#### What? Why?

Experimental caching proposal based on pain points from today's data and investigating which relevant bits can feasibly be cached.

- Adds some traditional-ish (longer-lived, auto-expiring) caching on some very heavy queries that are used in various places (including shops, groups) via `OpenFoodNetwork::EnterpriseInjectionData`.
- Caches final output of two big AMS injection arrays (all taxons and all properties!) in darkswarm layout based on the `:updated_at` attribute of the last updated object for those classes (auto-expiring).
- Avoids running product list queries if an OC is closed (fail early)
- Adds some time-based fragment-caching on the AMS data injected in `/shops` (index) and `/shop` (show) with conservative expirations (**15 seconds**). This data is things like `description`, `twitter_handle`, `supplied_taxons`.
- Adds some time-based (post-render) Action Caching in order cycle taxons and properties API endpoints (**30 seconds**). This data populates the filters in a shop page for filtering a specific shop's product list.
- Adds the ability to use Action Caching in the API
- Extracts caching-related code into a central service


luisramos0:
Interesting PR!

The word strategic took me to what I think should be solution to this problem: API.
Injected json in html is not the way to go. But making the page call API endpoints that have good caching strategies. We could do very similar things to this PR using HTTP caching on the user browser as well as having our own API cache, with a simple reverse proxy.

Matt-Yorkley:
Currently we can't do any HTTP caching until this is resolved: https://github.com/openfoodfoundation/openfoodnetwork/issues/5150 :see_no_evil:

app/controllers/api/order_cycles_controller.rb
```diff
@@ -15,26 +17,36 @@ def products
...
+          joins(:products).
+          where(spree_products: { id: distributed_products }).
+          select('DISTINCT spree_taxons.*')
+      end
 
       render json: ActiveModel::ArraySerializer.new(taxons, each_serializer: Api::TaxonSerializer)
     end
 
     def properties
+      properties = Rails.cache.fetch("oc-properties-#{order_cycle.id}", expires_in: 30.seconds) do
```
sauloperez, line 35: can we use a constant instead of a magic number for the `expires_in`? It will make future changes slightly easier.

app/helpers/cache_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+  end
+
+  # Gets the updated_at timestamp of the most recently updated record of a given class
+  def latest_timestamp_by_class(cached_class)
+    @_memoized_timestamps ||= Hash.new do |hash, key|
+      hash[key] = key.maximum(:updated_at).to_i
+    end
+    @_memoized_timestamps[cached_class]
+  end
```
sauloperez, line 17: I don't see what's the actual benefit of this approach compared to Rail's [compound cache key](https://guides.rubyonrails.org/caching_with_rails.html#fragment-caching). There the expiration happens by simply fetching a new cache key whereas we have to check all the models' `update_at`.

app/controllers/api/order_cycles_controller.rb
```diff
@@ -15,26 +17,36 @@ def products
...
-      taxons = Spree::Taxon.
-        joins(:products).
-        where(spree_products: { id: distributed_products }).
-        select('DISTINCT spree_taxons.*')
+      taxons = Rails.cache.fetch("oc-taxons-#{order_cycle.id}", expires_in: 30.seconds) do
+        Spree::Taxon.
+          joins(:products).
+          where(spree_products: { id: distributed_products }).
+          select('DISTINCT spree_taxons.*')
+      end
```
luisramos0, line 29: If we go ahead with this, all the caching logic needs to be extracted to services. All calls to Rails.cache need to be in a cache specific class, in this case for example: have a cached_data service that takes the cache key and yields back to this method and is re-used for the different places where we do this caching?

app/helpers/cache_helper.rb
luisramos0, line 17: could this be a service instead?

Matt-Yorkley: Rails's cache key applies to a single record, right? Like `cache: @post`. Here we're creating a usable cache key for all the records for a class, like `Spree::Taxon`, and it expires when any taxon is updated. Is there a built-in Rails way to do that? I wasn't trying to avoid `touch` here.

I added memoization afterwards because there can be multiple calls to the method for the same class in a request cycle, so we can avoid re-fetching the same value for the same class. The initial method (without the memoization) was:

```ruby
def latest_timestamp_by_class(cached_class)
  cached_class.maximum(:updated_at).to_i
end
```

luisramos0:
yeah, we would start by making #5150 specific to the cart endpoint!

Matt-Yorkley:
Yeah, I'm not sure we should merge this either, I was just trying to explore some ideas. :+1:

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: I realised we can use ActionCaching here instead (avoid the magic number issue in a similar way). It's much cleaner, and I think it's more efficient as it caches the data _after_ the output from the render call has been generated.

app/helpers/cache_helper.rb
Matt-Yorkley: I've refactored this to a service and added more comments.

Matt-Yorkley: Moved to service :+1:

sauloperez: > Rails's cache key applies to a single record, right? Like cache: @post

that's right. However, if this is per-class and checking again the Rails guides, this seems to then totally duplicate AR's built-in query cache? Are we sure we're not caching what's already cached?

Sources:

* https://guides.rubyonrails.org/caching_with_rails.html#sql-caching
* https://coderwall.com/p/tvybhq/bypassing-activerecord-query-cache

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: Hmmm, turns out fancy Action Caching doesn't play nicely with API controllers in Rails. It's got issues with any controller that inherits from `ActionController::Metal` instead of `ActionController::Base` and the workarounds don't look very nice.

app/helpers/cache_helper.rb
Matt-Yorkley: Aha, so we can just use the simpler/cleaner method and not bother memoizing it?

app/controllers/api/order_cycles_controller.rb
```diff
@@ -4,8 +4,11 @@ class OrderCyclesController < Api::BaseController
     respond_to :json
 
     skip_authorization_check
+    caches_action :taxons, :properties, expires_in: 30.seconds unless Rails.env.test?
```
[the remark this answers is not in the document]
Matt-Yorkley: It's slightly tricky there, as we have caching enabled in test.rb because we use caching on enterprise data, and that's tested in feature specs.

app/services/cache_service.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+  end
+
+  # Yields a cached query, expired by the most recently updated record for a given class.
+  # E.g: if *any* Spree::Taxon record is updated, all keys based on Spree::Taxon will auto-expire.
+  def self.cached_data_by_class(cache_key, cached_class)
+    Rails.cache.fetch "#{cache_key}-#{cached_class}-#{latest_timestamp_by_class(cached_class)}" do
+      yield
+    end
+  end
```
Matt-Yorkley, line 16: Rails's normal built-in cache magic works with single records. Under the hood it creates a cache key derived from the record's `:updated_at` value to auto-expire related caches when the record is updated.

What we're gaining here with `#cached_data_by_class` and `#latest_timestamp_by_class` is usable cache keys that auto-expire when _any_ records for a given class are updated. So for example we can cache some AMS that serializes _all taxons_, and it'll be invalidated when _any_ taxon is updated. So if nobody has updated a `Spree::Taxon` in the last 4 hours, and we've fragment-cached a big chunk of taxons AMS in a view (after it's been fetched, serialized and _rendered_) using `latest_timestamp_by_class(Spree::Taxon)` in the key, we can keep serving the same cached data the whole time instead of rebuilding it on every page load. :tada:

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: Update: we can use Action Caching in the API.

app/models/concerns/api_action_caching.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# API controllers inherit from ActionController::Metal to keep them slim and fast.
+# This concern adds the minimum requirements needed to use Action Caching in the API.
+
```
Matt-Yorkley, line 5: This is exciting! Normally in an API endpoint we go through:

1. Fetching a bunch of data
2. Serializing it
3. Rendering the output

All 3 of these steps are CPU-bound. By using Action Caching in controllers (similar to Fragment Caching in views) we can cache the entire output of an action, _after_ the render. So we're caching after point 3, instead of after point 1 or 2.

sauloperez: Yes, it'll be a huge win if we can use action caching. My biggest concern are query params since action caching doesn't seem to use them in the cache key but there seems to be an option for that: https://stackoverflow.com/a/2166419/2838228

Matt-Yorkley: I did some testing and it seemed to use the entire URL as the cache key, which included order id and distributor id in this case. I'll take a closer look...

Matt-Yorkley: Updated :+1:

app/controllers/api/order_cycles_controller.rb
```diff
@@ -1,11 +1,18 @@
...
     include EnterprisesHelper
-    respond_to :json
+    include ApiActionCaching
 
     skip_authorization_check
+    skip_before_filter :authenticate_user, :ensure_api_key, only: [:taxons, :properties]
+
+    caches_action :taxons, :properties,
+                  expires_in: CacheService::FILTERS_EXPIRY,
+                  cache_path: proc { |controller| controller.request.url }
```
luisramos0, line 11: Do I understand correctly that this action caching is not related to the cache_service below? If this is correct, maybe we need to move this to a new PR.

I think this looks a promising form of API caching!

I am not sure I am missing something but can I approve this part of the PR and reject what's happening on the injectiondata side of this PR? If we move the injection_data methods to API endpoints we can than use this technique to cache them?

anyway, great work @Matt-Yorkley :muscle:

Matt-Yorkley: I think in some cases it might be detrimental (performance-wise) to move all of the injected data to endpoints. We inject 7 different sets of AMS data in the Darkswarm layout for example (and more on specific pages). If we add an extra 7 requests per page load, we would be hitting the stack an extra 7 times on every page. The API controllers are thinner than the standard controllers, but not that thin. So maybe performance could get worse in that scenario..?

Matt-Yorkley: There are some bits of injection data that would be really great to extract to the API, but would currently take a large amount of work (like the /shops page).

lib/open_food_network/enterprise_injection_data.rb
```diff
@@ -1,28 +1,51 @@
...
+          OrderCycle.earliest_closing_times
+        end
+      end
     end
 
     def shipping_method_services
-      @shipping_method_services ||= Spree::ShippingMethod.services
+      @shipping_method_services ||= begin
+        CacheService.cached_data_by_class("shipping_method_services", Spree::ShippingMethod) do
+          Spree::ShippingMethod.services
```
Matt-Yorkley, line 22: @luisramos0 You don't like this bit?

Here for example when we call `Spree::ShippingMethod#services` we're running a huge query which involves a join across all shipping methods and all distributors in the entire database, which is really nasty. When the query is done we parse _each item_ via `#map` to arrange it in a giant hash in a specific way. We call this method really often! :see_no_evil: 

With `CacheService#cached_data_by_class` we can cache the resulting Hash and keep serving it (in about a millisecond) until the next time somebody changes a shipping method, at which point the cache entry will be automatically expired.

What's not to love? :sweat_smile:

luisramos0: And I am recognizing the marvelous feelings of the start of a caching project and also thinking about the very painful ends of such projects (remember the end of product cache?).

Are you now more convinced this is the way to go? I am really worried about the complexity introduced and all the trouble this can cause #twohardthings

how is the cache expired "automatically"?

app/controllers/api/order_cycles_controller.rb
luisramos0: I think we should move all of them to the API and then optimize the API to make it faster. I think it's very important we rely on API based solutions internally so that we can grow our public API and it's performance!

I wonder how difficult it would be. The payload would be exactly the same, we would only have to replace the angular code (it's now a variable with the name of the injected data in the controllers dependencies list) with a resource that fetches data from the API as well as creating API endpoints that will be equal to the injectos. I think we have to make this way: all these would be become public API endpoints.

sauloperez: I agree with your view @luisramos0 but I understand your point @Matt-Yorkley . That's a potential downside BUT if we consider the whole picture (browser rendering, ajax calls, server latency) we might even get quicker overall browser response time if we intelligently fetch things asynchronously and don't block the page rendering. You know, there's a lot of science in there.

app/controllers/api/order_cycles_controller.rb
```diff
@@ -70,11 +77,11 @@ def permitted_ransack_params
     end
 
     def distributor
-      Enterprise.find_by_id(params[:distributor])
+      @distributor ||= Enterprise.find_by_id(params[:distributor])
     end
 
     def order_cycle
-      OrderCycle.find_by_id(params[:id])
+      @order_cycle ||= OrderCycle.find_by_id(params[:id])
```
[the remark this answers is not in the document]
sauloperez: thoughts?

app/controllers/home_controller.rb
```diff
@@ -18,9 +18,11 @@ def sell; end
 
   private
 
-  # Cache the value of the query count for 24 hours
+  # Cache the value of the query count
   def cached_count(key, query)
-    Rails.cache.fetch("home_stats_count_#{key}", expires_in: 1.day, race_condition_ttl: 10) do
+    CacheService.cache("home_stats_count_#{key}",
+                       expires_in: CacheService::HOME_STATS_EXPIRY,
```
sauloperez, line 24: if `CacheService` is the owner of that constant, we might as well not leak the `expires_in` outside, don't you think? I talking about having something like

```
CacheService.cache("home_stats_count_#{key}", race_condition_ttl: 10) do
end
```

while still having the option to override the `expires_in` if we need to.

lib/open_food_network/enterprise_injection_data.rb
sauloperez: I share your feelings @luisramos0 . If we go down this path we need be extra careful but pretty please, @Matt-Yorkley let's split this up and go bit by bit so we can measure.

As for the cache expiration @luisramos0 I believe https://signalvnoise.com/posts/3113-how-key-based-cache-expiration-works is the established reference, in the Rails community at least.

luisramos0: thanks for the link Pau! I now understand the cached_data_by_class code.

It still depend on the "touch" calls. Are we sure the touch calls introduced in this PR are enough for all the cached data? I'd say we can only do this if we cover each data update case with unit tests. For example, in the case of active_distributor_ids, we need to be sure that when an OC is added to an enterprise this cache key is invalidated (the Enterprise object wont be updated), right? Is that happening?

Matt-Yorkley: > how is the cache expired "automatically"?

I'll try to clarify it a bit here for reference...

Lets look at an example of automatically expiring cache keys, where we fetch all products in the database but we cache the content per day:

```ruby
day = Time.now.strftime("%Y-%m-%d") # => "2020-04-14"

Rails.cache.fetch("auto-expire-after-#{day}") do
   Products.all
end
```

When we pass a key to `Rails.cache.fetch`, it will return the cached result if a cache entry with that key exists, otherwise run whatever is in the block and save it under the provided key.

So the first time the above code runs on a given day, the result will be cached under a cache key called for example `auto-expire-after-2020-04-14`. For the rest of the day, any time this code runs, the key we're _passing in_ to `#fetch` will still be `auto-expire-after-2020-04-14`, so Rails will return the cached value instead of running the query, because an entry with the specified key exists. 

As soon as this code runs after midnight, the key being passed in will _change_ to: `auto-expire-after-2020-04-15`. At this point a key with that value is _not found_ in the cache, so Rails runs the query in the block (`Products.all`), and saves the result under the specified key. Then for the rest of the day this code will serve the new cached value and not run the query, and so on. The cache entry automatically expires when `Time.now.strftime("%Y-%m-%d")` changes.

Side note; luckily Memcached automatically handles removal of old / unused cache entries so we don't have to worry about cleaning up. that's one of the main things it is designed to do, and do well.

So now another example...

```ruby
most_recently_updated_product_in_the_database = Product.maximum(:updated_at)
timestamp = most_recently_updated_product_in_the_database.to_i

Rails.cache.fetch("auto-expire-me-after-#{timestamp}") do
   Products.all
end
```

In this example we get the `:updated_at` value of the most recently updated `Product` record in the database. We make a timestamp from that value, eg: `1586822400`. We then cache the big product query, using the timestamp in the key. Something like: `auto-expire-me-after-1586822400`.

The first time this code runs, the corresponding key will not be found, so Rails will run the query and save the result under the given cache key. If we run it again and again it will serve the cached content and _not_ trigger the query.

If we then _update and save_ a `Product` record (any product), the timestamp we're generating in the key will not match the old one, the cache key will _not_ be found, Rails will run the query and will save the new result under the new key (eg: `auto-expire-me-after-1586999999`). So: the cache entry in this code is auto-expired whenever a `Product` is updated in the database.

Matt-Yorkley: > we need to be sure that when an OC is added to an enterprise this cache key is invalidated

When an `Exchange` is updated, we already `touch` the receiver (distributor). We definitely need some extra specs!

app/controllers/home_controller.rb
Matt-Yorkley: I was thinking about extracting the whole thing into a specific `#home_stats` method in `CacheService`...?

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: On the `/shops` page specifically we have a lot of Geocoding functionality that happens in javascript libraries as part of the search. Moving all the logic to the backend would be tricky. We would also then have a lot of different combinations of search params for the endpoint which would make it unfeasible to cache the output nicely. I'd still love to do that at some point though. It's desperately needed but it's not a small job.

Matt-Yorkley: I generally agree on moving things to the API and then making it faster though! I just don't think that will happen very soon.

lib/open_food_network/enterprise_injection_data.rb
luisramos0: great, thanks for the details, that's what I understood from the link above :+1:
To complete your detailed description, in that product cache we are including variants of the product. If a variant is updated, the product must be touched, sometimes this happens implicitly (some callback), sometimes it happens explicitly (touch).
The problem with this approach is really object composition, you need to make sure that any change to any part of the product serialization will touch the product object.

The checklist from any entry like this would be:
A - what objects compose the serialization of the main object? this needs to be recursive!! for example, updates to stock items must touch products in some way :see_no_evil: 
B - write a unit test that makes sure an update to every part of the payload will touch the main object (and therefore refresh the cache).

Matt-Yorkley: Ok, it turns out three of the six queries here ultimately depend on the `OrderCycle.active` scope, so they won't work with normal `touch` calls:

https://github.com/openfoodfoundation/openfoodnetwork/blob/f366aa260516788e9c283ec9785206420a52654a/app/models/order_cycle.rb#L29-L33

I've cut those three.

Matt-Yorkley: > updates to stock items must touch products in some way

Yes, the example was a simple one. I've intentionally avoided using this on anything to do with actual products because it would be a total nightmare, as you say.

Matt-Yorkley: Taxons and shipping methods are probably not updated very often, don't use composition, and are connected to other objects in a fairly simple way where we can use `touch`, but most importantly: we fetch all taxons and all shipping methods in the database on every Darkswarm page.

app/views/layouts/darkswarm.html.haml
```diff
@@ -48,8 +48,10 @@
     = inject_current_hub
     = inject_current_user
     = inject_rails_flash
-    = inject_taxons
-    = inject_properties
+    - cache "inject-all-taxons-#{CacheService.latest_timestamp_by_class(Spree::Taxon)}" do
+      = inject_taxons
+    - cache "inject-all-properties-#{CacheService.latest_timestamp_by_class(Spree::Property)}" do
+      = inject_properties
```
Matt-Yorkley, line 54: This is really great! We're caching two big AMS arrays (all the taxons and properties in the database) _after_ they're fetched, serialized, and rendered, and expiring the cache entries automatically. We'll be saving a lot of CPU time here on every page load.

luisramos0: :+1:
we need an auto test that covers the case where we just assign an existing property to a product and that change is reflected in the cached value. what are all possible ways of assigning properties to products? I think it's in the product properties page.
The same applies to taxons. I dont think we have a way to add taxons to a product!? Only changing the category of the product (primary taxon).

Matt-Yorkley: In this specific cases in this view, the objects being rendered here are literally just `Spree::Taxon.all` and `Spree::Property.all`, with no absolutely relation to whether or not they're applied to any products, or producers, or anything. And their serializers are not calling associated data.

But yes, we'll need to do that in the queries being cached in `EnterpriseInjectionData`, which do involve some joins and associations and need much more careful handling :+1:

luisramos0: ok :+1: You still need to test for the create taxon case and create properties.
You can create properties from both the properties page and the product properties page.

Matt-Yorkley: Saving a new record always sets it's `:updated_at` value to the current time, right?

Matt-Yorkley: I've removed some of the caching on one of the more complicated queries, updated the touches and touch specs and added some code comments.

Matt-Yorkley:
If we want to squeeze the most out of this, and considering the product list endpoint is one of the most hit, we could potentially do this as well in `OrderCyclesController`:

```ruby
caches_action :products,
              expires_in: CacheService::PRODUCTS_EXPIRY, # 15 seconds
              cache_path: proc { |controller| controller.request.url },
              unless: proc { params[:q].present? || (params[:page].to_i > 1) }
```

This caches the first page of results in the product list for 15 seconds. It ignores caching if the `params[:page]` is greater than one, or any query params have been sent, so we don't create unique cache entries for every possible combination of pages and search params. 

This will probably help a lot in stampede conditions. It's the only bit of caching I think we can feasibly apply to this endpoint.

I'm assuming the fact that the first 10 products will be slightly stale (for a maximum 15 seconds) is okay, because:

- when a product is added to the cart we already do an asynchronous call with a stock check, and if the stock has changed the data on the page is updated and the user is informed
- as soon as the user adds a search word or clicks a filter, the whole dataset will change anyway and the stale data (maximum 15 seconds) will be gone
- last week we had **22.6k** requests in one hour to `/shop`, and each one will hit this exact endpoint, with these exact params (basically no params) as soon as the page loads

Any thoughts?

Matt-Yorkley:
Closing; split into 4 separate PRs.

PR #5283, opened 2020-04-21
Fix disappearing tags issue

### LINKED ISSUE #5282: Tag display does not work for managers
## Description

As a manager of a hub, if you tag someone else than yourself, the tag will not be displayed in the list of customers. However it is still working.

A bit disturbing for our users: they have no possibility to check if they have tagged the right person.

## Steps to Reproduce

1. On demo hub in French production (ID 212) I've tagged an OC as "prive". This means that all customers with the tag "vip" can see it and not the others.
2. My openfood account is vip - my gmail account is not
3. Connected as openfood I can see the OC, connected as gmail I cannot - so far so good
4. But as a manager of the hub I cannot see who is tagged with that tag (I've added it to Myriam and myself). The only customer I see with the tag "vip" is a customer that was tagged a few years ago.

## Animated Gif/Screenshot

image

## Severity

## Your Environment
* Version used: v2.9.4

Matt-Yorkley:
#### What? Why?

Closes #5282 

#### What should we test?

Tags in admin customers page are working properly.

Go to admin, customers and select an enterprise. You will need multiple customers with multiple tags. Some of those tags should be on multiple customers. Then save and reload the page.

On the old version (production) you can observe that a tag that has been used for multiple customers will appear only on one customer. So each tag is displayed only once. But with this fix, you should see the tag on all customers it was assigned to.

#### Release notes

Fixed problem with disappearing tags on admin customers page.

### DIFF
--- a/app/controllers/admin/customers_controller.rb
+++ b/app/controllers/admin/customers_controller.rb
@@ -102,9 +102,11 @@ module Admin
                   context: 'tags' })
 
       customer_tags.each_with_object({}) do |tag, indexed_hash|
-        customer_id = tag.taggings.first.taggable_id
-        indexed_hash[customer_id] ||= []
-        indexed_hash[customer_id] << tag.name
+        tag.taggings.each do |tagging|
+          customer_id = tagging.taggable_id
+          indexed_hash[customer_id] ||= []
+          indexed_hash[customer_id] << tag.name
+        end
       end
     end
   end


app/controllers/admin/customers_controller.rb
```diff
@@ -102,9 +102,11 @@ def customer_tags_by_id
                   context: 'tags' })
 
       customer_tags.each_with_object({}) do |tag, indexed_hash|
-        customer_id = tag.taggings.first.taggable_id
-        indexed_hash[customer_id] ||= []
-        indexed_hash[customer_id] << tag.name
+        tag.taggings.each do |tagging|
```
mkllnk, line 105: This makes sense to me now. One tag is associated to several customers and we need to look at all of them.

For another PR in the future: It might be easier to load Taggings instead of Tags and iterate over them.

mkllnk:
If any tester wants to pick this up, I already staged it on au-staging.

kirstenalarsen:
right i see it's in aus staging - i'll have a little crack at it

kirstenalarsen:
wow the difference that clear testing instructions makes!! This is a world apart from when I used to test things - thanks @Matt-Yorkley - makes really easy

image

yep all good
image

deploy? 

btw. if it's not affecting the database then I am less concerned to rush to deploy - as in the shops and tags should still be working and it's just this page that's not. But I get how off-putting that is, so agree let's just get it patched

@mkllnk

lauriewayne:
Your North American fanbase thanks you!!! <3

PR #5289, opened 2020-04-22
Cart populate

Matt-Yorkley:
#### What? Why?

Improves performance in `CartController#populate`. This is currently at the top of the list of worst-performing / most-used endpoints in Datadog :warning::fire:. 

I've made some reasonable improvements here, but there's still a huge mess of really expensive adjustments re-calculations that are triggered over the course of this action, and these are run on every item in the cart, every time a product is added or removed from the cart.

#### What should we test?

Adding and removing items from the cart generally, from the `/shop` page.

- Fees, adjustments, and alternate prices from applicable variant overrides are correct when adding / removing products in the cart
- Clicking the up/down buttons when adding quantity to cart: there should be a short (half second) wait before the changes are actually sent to the server in a request. The wait should be extended with multiple clicks, eg: clicking it really fast 5 times will result in 1 request, not 5.
- Mouse scroll when the mouse if over the add to cart up/down buttons does not make the number change wildly (and trigger extra cart changes)

Screenshot from 2020-04-28 22-07-20

#### Release notes

Improved performance in CartController#populate endpoint (add item to cart)

### DIFF
--- a/app/controllers/cart_controller.rb
+++ b/app/controllers/cart_controller.rb
@@ -14,9 +14,7 @@ class CartController < BaseController
       cart_service = CartService.new(order)
 
       if cart_service.populate(params.slice(:products, :variants, :quantity), true)
-        fire_event('spree.cart.add')
-        fire_event('spree.order.contents_changed')
-
+        order.update_distribution_charge!
         order.cap_quantity_at_stock!
         order.update!
 
--- a/app/controllers/cart_controller.rb
+++ b/app/controllers/cart_controller.rb
@@ -4,24 +4,26 @@ class CartController < BaseController
   before_filter :check_authorization
 
   def populate
+    order = current_order(true)
+
     # Without intervention, the Spree::Adjustment#update_adjustable callback is called many times
     # during cart population, for both taxation and enterprise fees. This operation triggers a
     # costly Spree::Order#update!, which only needs to be run once. We avoid this by disabling
     # callbacks on Spree::Adjustment and then manually invoke Spree::Order#update! on success.
     Spree::Adjustment.without_callbacks do
-      cart_service = CartService.new(current_order(true))
+      cart_service = CartService.new(order)
 
       if cart_service.populate(params.slice(:products, :variants, :quantity), true)
         fire_event('spree.cart.add')
         fire_event('spree.order.contents_changed')
 
-        current_order.cap_quantity_at_stock!
-        current_order.update!
+        order.cap_quantity_at_stock!
+        order.update!
 
         variant_ids = variant_ids_in(cart_service.variants_h)
 
         render json: { error: false,
-                       stock_levels: VariantsStockLevels.new.call(current_order, variant_ids) },
+                       stock_levels: VariantsStockLevels.new.call(order, variant_ids) },
                status: :ok
       else
         render json: { error: true }, status: :precondition_failed
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -44,6 +44,15 @@ module VariantStock
     #   This provides a default value for variant.on_demand using Spree::StockLocation.backorderable_default
     return Spree::StockLocation.first.backorderable_default if new_record? || deleted?
 
+    # This can be removed unless we have seen this error in Bugsnag recently
+    if stock_item.nil?
+      Bugsnag.notify(
+        RuntimeError.new("Variant #stock_item called, but the stock_item does not exist!"),
+        object: as_json
+      )
+      return Spree::StockLocation.first.backorderable_default
+    end
+
     stock_item.backorderable?
   end
 
--- a/app/models/concerns/variant_stock.rb
+++ b/app/models/concerns/variant_stock.rb
@@ -40,9 +40,9 @@ module VariantStock
 
   # Checks whether this variant is produced on demand.
   def on_demand
-    # A variant that has not been saved yet, doesn't have a stock item
+    # A variant that has not been saved yet or has been soft-deleted doesn't have a stock item
     #   This provides a default value for variant.on_demand using Spree::StockLocation.backorderable_default
-    return Spree::StockLocation.first.backorderable_default if stock_items.empty?
+    return Spree::StockLocation.first.backorderable_default if new_record? || deleted?
 
     stock_item.backorderable?
   end
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -240,8 +240,8 @@ class OrderCycle < ActiveRecord::Base
   end
 
   def exchanges_supplying(order)
-    variants_relation = Spree::Variant.joins(:line_items).merge(Spree::LineItem.in_orders(order))
-    exchanges.supplying_to(order.distributor).with_any_variant(variants_relation)
+    variant_ids_relation = Spree::LineItem.in_orders(order).select(:variant_id)
+    exchanges.supplying_to(order.distributor).with_any_variant(variant_ids_relation)
   end
 
   def coordinated_by?(user)
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -240,7 +240,8 @@ class OrderCycle < ActiveRecord::Base
   end
 
   def exchanges_supplying(order)
-    exchanges.supplying_to(order.distributor).with_any_variant(order.variants.map(&:id))
+    variants_relation = Spree::Variant.joins(:line_items).merge(Spree::LineItem.in_orders(order))
+    exchanges.supplying_to(order.distributor).with_any_variant(variants_relation)
   end
 
   def coordinated_by?(user)
--- a/app/services/cart_service.rb
+++ b/app/services/cart_service.rb
@@ -34,7 +34,7 @@ class CartService
     quantity = quantity.to_i
     max_quantity = max_quantity.to_i if max_quantity
     variant = Spree::Variant.find(variant_id)
-    OpenFoodNetwork::ScopeVariantToHub.new(@distributor).scope(variant)
+    scoper.scope(variant)
 
     return unless quantity > 0 && valid_variant?(variant)
 
@@ -67,6 +67,10 @@ class CartService
   end
 
   private
+
+  def scoper
+    @scoper ||= OpenFoodNetwork::ScopeVariantToHub.new(@distributor)
+  end
 
   def read_variants(data)
     @variants_h = read_products_hash(data) +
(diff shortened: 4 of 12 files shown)


Matt-Yorkley:
Looking at a cart with 20 items, using production data, before and after:

Screenshot from 2020-04-28 21-02-06

Screenshot from 2020-04-28 00-37-07

Total load time reduced by ~32% and queries reduced by ~57%.

Not bad, but as I said there's still crazy stuff going on in this endpoint related to re-calcuating all adjustments every time an item is added, mostly in: `enterprise_fee_calculator.rb`.

app/controllers/cart_controller.rb
```diff
@@ -4,24 +4,24 @@ class CartController < BaseController
...
     # during cart population, for both taxation and enterprise fees. This operation triggers a
     # costly Spree::Order#update!, which only needs to be run once. We avoid this by disabling
     # callbacks on Spree::Adjustment and then manually invoke Spree::Order#update! on success.
     Spree::Adjustment.without_callbacks do
-      cart_service = CartService.new(current_order(true))
+      cart_service = CartService.new(order)
 
       if cart_service.populate(params.slice(:products, :variants, :quantity), true)
-        fire_event('spree.cart.add')
-        fire_event('spree.order.contents_changed')
```
Matt-Yorkley, line 16: I removed these two event calls here. I can't find anything anywhere in `spree_core` or in our codebase that actually listens on the `spree.cart.add` event, and the only listener we have on `spree.order.contents_changed` just calls `#update_distribution_charge!` on the current order, so I've added that call explicitly (below) in place of the event.

Also we were using `current_order` multiple times here, and calling things on it, and it seemed to be re-fetching the order every time.

app/services/cart_service.rb
```diff
@@ -15,28 +15,35 @@ def populate(from_hash, overwrite = false)
...
 
-  def attempt_cart_add_variants(variants)
-    variants.each do |v|
-      if varies_from_cart(v)
-        attempt_cart_add(v[:variant_id], v[:quantity], v[:max_quantity])
-      end
+  private
+
+  def attempt_cart_add_variants(variants_data)
+    loaded_variants = indexed_variants(variants_data)
```
Matt-Yorkley, line 28: Here we're loading all the variants we will need in the following code in advance (with their associated objects eager-loaded), to avoid loading them all one-by-one (and loading all their associated objects one-by-one as well). 

I did some renaming of variables in this class to distinguish between the hash of `variant_data` that's coming from posted `params`, and actual variant objects.

app/controllers/cart_controller.rb
luisramos0: hey, lets burn all the fire_events! It's worth doing it now that we have verified they are not used anywhere in Spree!
There are only two more:

app/controllers/checkout_controller.rb:
   50:     fire_event('spree.checkout.update')

This :point_up: can just be deleted

app/controllers/spree/orders_controller.rb:
  ```
        if @order == current_order
          fire_event('spree.order.contents_changed')
        else
          @order.update_distribution_charge!
        end
```
Reading the code, I'd just replace these lines with
@order.update_distribution_charge!

File: app/models/concerns/variant_stock.rb [hunk 3]
```diff
@@ -40,9 +40,9 @@ def on_hand=(new_level)
 
   # Checks whether this variant is produced on demand.
   def on_demand
-    # A variant that has not been saved yet, doesn't have a stock item
+    # A variant that has not been saved yet or has been soft-deleted doesn't have a stock item
     #   This provides a default value for variant.on_demand using Spree::StockLocation.backorderable_default
-    return Spree::StockLocation.first.backorderable_default if stock_items.empty?
+    return Spree::StockLocation.first.backorderable_default if new_record? || deleted?
```
luisramos0, line 45: nice catch!
Maybe there will be situations where the record isnt new, the record isnt deleted and still there is no stock_item... :boom: 
I'd keep both your conditions and maybe add the previous conditional on stock_items.empty? #defensive programming

app/models/order_cycle.rb
```diff
@@ -240,7 +240,8 @@ def exchanges_carrying(variant, distributor)
   end
 
   def exchanges_supplying(order)
-    exchanges.supplying_to(order.distributor).with_any_variant(order.variants.map(&:id))
+    variants_relation = Spree::Variant.joins(:line_items).merge(Spree::LineItem.in_orders(order))
```
luisramos0, line 243: with_any_variant takes variant ids
why not simply:
variants_relation = Spree::LineItem.in_orders(order).select(:variant_id)

app/services/cart_service.rb
```diff
@@ -143,11 +138,22 @@ def check_variant_available_under_distribution(variant)
...
-    order.find_line_item_by_variant Spree::Variant.find(variant_id)
+  def line_item_for_variant(variant)
+    order.find_line_item_by_variant variant
   end
 
   def variant_ids_in_cart
     @order.line_items.pluck :variant_id
   end
+
+  def indexed_variants(variants_data)
```
luisramos0, line 149: this is only used in the private method attempt_cart_add_variants, I'd move this below that method to reduce scrolling up and down.

app/models/concerns/variant_stock.rb
luisramos0: in any case, this needs to be unit tested in variant_stock_spec.

luisramos0:
please dont force push on this PR so I dont have to read it all over again, thanks!

app/controllers/cart_controller.rb
luisramos0: I think this is out of scope for this PR and also it will require very different testing (checkout, etc) so I created a tech debt PR for this:
#5324

app/models/concerns/variant_stock.rb
Matt-Yorkley: The problem is, if it's not new or deleted and we call `stock_items.empty?`, we trigger a pointless query in this line that counts the variant's stock items to see if any exist. If it's not new or deleted it will always have one.

In cases like the cart where we're checking variant stock on lots of variants, if we have 30 items in the cart we will trigger this query 30 times. I'm pretty sure this happens elsewhere in the app as well, where we do a big N+1 query because of this `stock_items.empty?` check...

luisramos0: I agree,
I am just not 100% sure about "If it's not new or deleted it will always have one."
If this is true then this code:
`return Spree::StockLocation.first.backorderable_default if new_record? || deleted? || stock_items.empty?`
will never run stock_items.empty?
That's why I'd add it to the end of the line to 100% sure.

Matt-Yorkley: I see what you mean. But won't it always run the last bit in this example, when a variant is not new and not deleted?

Matt-Yorkley: I've added a Bugsnag call for the case where there's a saved, non-deleted variant that does not have a `stock_item`, and ensured that the method will return what it previously would have returned in the case where `stock_items.empty? == true`.

luisramos0: ok, I see. yes, it will call stock_items.empty? in those cases.
I wonder if stock_items.any? would be better
or even stock_items.first.nil? (which is what stock_item does later: stock_items.first)

I am ok with your proposed solution, I think we will see some bugsnags.
In this case I'd prefer to have a defensive code solution that does not rely on the state of the data.

luisramos0: can you please add a unit test to variant_stock_spec?

luisramos0: awesome, thanks!

spec/models/concerns/variant_stock_spec.rb
```diff
@@ -90,13 +90,33 @@
...
+        expect(variant.stock_items.count).to eq 0
+      end
+
       it 'returns stock location default' do
         expect(variant.on_demand).to be_falsy
       end
     end
+
+    context 'when the variant has been soft-deleted' do
+      let(:deleted_variant) { create(:variant) }
```
mkllnk, line 106: I think you could save the `before` block.

```suggestion
      let(:deleted_variant) { create(:variant).tap(&:destroy) }
```

app/services/cart_service.rb
```diff
@@ -135,11 +139,22 @@ def check_variant_available_under_distribution(variant)
...
+    @indexed_variants ||= begin
+      variant_ids_in_data = variants_data.map{ |v| v[:variant_id] }
+
+      Spree::Variant.where(id: variant_ids_in_data).
+        includes(:stock_items, :product).
+        all.
+        index_by(&:id)
```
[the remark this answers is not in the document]
Matt-Yorkley: Big lists? This is only used on the cart add/remove endpoint, and only fetches the items that will definitely be used in the cart calculations that happen shortly after. Fetching all the variants we need at once (with their stock items etc) is much better than fetching them one by one here.

spec/models/concerns/variant_stock_spec.rb
Matt-Yorkley: Done :+1:

app/services/cart_service.rb
```diff
@@ -135,11 +139,22 @@ def check_variant_available_under_distribution(variant)
...
     @order.line_items.pluck :variant_id
   end
+
+  def indexed_variants(variants_data)
+    @indexed_variants ||= begin
+      variant_ids_in_data = variants_data.map{ |v| v[:variant_id] }
+
+      Spree::Variant.where(id: variant_ids_in_data).
+        includes(:stock_items, :product).
+        all.
```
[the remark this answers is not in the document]
Matt-Yorkley: Done :+1:

filipefurtad0:
Hi @Matt-Yorkley 

Did all the manual checks:
- Variant overrides have behave correctly when adding / removing products in the cart
- Verified the delay between clicking and the moment the changes are sent to the server.
- It's possible to add as many items as possible, with just one request
- The mouse scroll over the add to cart up/down buttons is now gone.

Awesome! Ready to go.

spec/features/consumer/shopping/shopping_spec.rb
```diff
@@ -536,4 +537,10 @@ def shows_products_without_customer_warning
     expect(page).to have_no_content "This shop is for customers only."
     expect(page).to have_content product.name
   end
+
+  def wait_for_debounce
+    # The auto-submit on these specific form elements (add to cart) now has a small built-in
+    # waiting period before submitting the data...
+    sleep 0.6
```
sauloperez, line 544: Capybara has better built-in mechanisms :see_no_evil: I suffered a bit with hidden sleeps in a test suite in the past.

Matt-Yorkley: In this case the expectation will totally break using Capybara's built in expectations...

PR #5295, opened 2020-04-23
Enterprise serializer optimisations

Matt-Yorkley:
#### What? Why?

Adds some simple optimisations to serialization of enterprise data:

- A bit of needed memoizing and eager-loading
- Avoids querying **supplied_taxons** and **supplied_properties** for enterprises that are **not suppliers**

#### What should we test?

I'm not quite sure about the best way to test this. A sanity check on displaying product properties and taxons? 

Basically if an enterprise is marked as not being a producer, we now won't try to find the taxons and properties of their own products. I think it's not even possible to add products as a non-producer?

#### Release notes

Improved performance of enterprise serialization

### DIFF
--- a/app/serializers/api/cached_enterprise_serializer.rb
+++ b/app/serializers/api/cached_enterprise_serializer.rb
@@ -113,7 +113,7 @@ module Api
     end
 
     def active
-      data.active_distributor_ids.andand.include? enterprise.id
+      @active ||= data.active_distributor_ids.andand.include? enterprise.id
     end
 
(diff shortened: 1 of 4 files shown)


filipefurtad0:
Hey @Matt-Yorkley ,

All good with the the sanity check on displaying product properties and taxons. Found nothing unusual.

And yes, I think you are right: it's not possible to add products as a non-producer hub, unless a specific producer explicitly allows them to.

Thank you for this further improvement!

PR #5304, opened 2020-04-27
5299 Impossible to filter orders with several tags

### LINKED ISSUE #5299: Impossible to filter order page with several OC
## Description

When selecting several OC as a search field on the order page, the result only shows one OC.

## Steps to Reproduce
1. Go to `/orders?q[s]=completed_at+desc`
2. Select several order cycles (OC) in the search field
3. Check the results

## Severity

## Your Environment
* Version used: v2.9.3

dacook:
#### What? Why?

Closes #5299

On the Admin Orders page (`/admin/orders`), there are two tag options in the filter:

 * Distributors
 * Order Cycles

When multiple tags for either of these are selected, the filter only uses the last tag, all others are ignored.

This fix ensures that all selected tags are used in the filter. It also removes a redundant duplicate line.

#### What should we test?
Test filtering with zero, one and multiple tags for both Distributors and Order Cycles

#### Release notes
Fix filtering by multiple tags on the Admin Orders page.

Changelog Category: Fixed

### DIFF
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -39,7 +39,6 @@ feature '
     end
 
     scenario "filter by multiple order cycles" do
-      # Set up 3 orders and 3 order cycles
       order_cycle_2 = create(:simple_order_cycle, name: 'Two')
       order_cycle_3 = create(:simple_order_cycle, name: 'Three')
       order_cycle_4 = create(:simple_order_cycle, name: 'Four')
@@ -54,11 +53,9 @@ feature '
       quick_login_as_admin
       visit 'admin/orders'
 
-      # Select order cycle 2 and 3
       multi_select2_select 'Two', from: 'q_order_cycle_id_in'
       multi_select2_select 'Three', from: 'q_order_cycle_id_in'
 
-      # click link "Filter Results"
       page.find('.filter-actions .button.icon-search').click
 
       # Order 2 and 3 should show, but not 4
--- a/spec/features/admin/orders_spec.rb
+++ b/spec/features/admin/orders_spec.rb
@@ -38,6 +38,36 @@ feature '
                   visible: :all)[:innerHTML]).to have_content(/.*Four.*Three.*Two/m)
     end
 
+    scenario "filter by multiple order cycles" do
+      # Set up 3 orders and 3 order cycles
+      order_cycle_2 = create(:simple_order_cycle, name: 'Two')
+      order_cycle_3 = create(:simple_order_cycle, name: 'Three')
+      order_cycle_4 = create(:simple_order_cycle, name: 'Four')
+
+      order_2 = create(:order_with_credit_payment, user: user, distributor: distributor,
+                                                  order_cycle: order_cycle_2)
+      order_3 = create(:order_with_credit_payment, user: user, distributor: distributor,
+                                                  order_cycle: order_cycle_3)
+      order_4 = create(:order_with_credit_payment, user: user, distributor: distributor,
+                                                  order_cycle: order_cycle_4)
+
+      quick_login_as_admin
+      visit 'admin/orders'
+
+      # Select order cycle 2 and 3
+      multi_select2_select 'Two', from: 'q_order_cycle_id_in'
+      multi_select2_select 'Three', from: 'q_order_cycle_id_in'
+
+      # click link "Filter Results"
+      page.find('.filter-actions .button.icon-search').click
+
+      # Order 2 and 3 should show, but not 4
+      expect(page).to have_content order_2.number
+      expect(page).to have_content order_3.number
+      expect(page).to_not have_content order_4.number
+
+    end
+
     context "with a capturable order" do
       before do
         order.finalize! # ensure order has a payment to capture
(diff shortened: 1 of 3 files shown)


dacook:
Hi @mkllnk , I just pushed a Jasmine spec and realised you mentioned a feature spec instead. 

The last line of my test is failing as I wasn't able to test the returned result. So I'm guessing this is why you mentioned a feature spec.

Is it worth keeping the Jasmine spec? 
And would this be the right place to add a feature spec: https://github.com/openfoodfoundation/openfoodnetwork/blob/master/spec/features/admin/order_cycles_spec.rb#L72 ?

mkllnk:
@dacook 
> Is it worth keeping the other part of the Jasmine spec?

Yes, good work. Those specs are great. Just remove the last line.

You could simulate the response but that doesn't help here. That would only help to test the code parsing the server response which is not part of this issue.

> And would this be the right place to add a feature spec

That doesn't look right. This is not about order cycles but about orders.

dacook:
Thanks @mkllnk, I got side-tracked the other day and added a Jasmine test, then realised I can't test the return results.
So I tried a feature spec today but am having trouble getting the multi select to work so far (spec/features/admin/orders_spec.rb:41).

dacook:
Hi @mkllnk , I've figured things out in the light of day, tests are now ready, thanks for your help. I've cleaned up my commits on this same branch.

spec/features/admin/orders_spec.rb
```diff
@@ -38,6 +38,36 @@
...
+
+      # Select order cycle 2 and 3
+      multi_select2_select 'Two', from: 'q_order_cycle_id_in'
+      multi_select2_select 'Three', from: 'q_order_cycle_id_in'
+
+      # click link "Filter Results"
+      page.find('.filter-actions .button.icon-search').click
+
+      # Order 2 and 3 should show, but not 4
+      expect(page).to have_content order_2.number
```
mkllnk, line 65: So this spec fails in this commit and then passes with your fixing commit, is that right?

dacook: Yep that's right, order 2 was not showing, as it was only showing orders from the last selected order cycle. I kept separate commits in case anyone needed to confirm the failing test, but will make sure to commit together in the future.

mkllnk: No, separate is fine but make the spec pass in every commit (change it twice).

spec/features/admin/orders_spec.rb
```diff
@@ -38,6 +38,36 @@
                   visible: :all)[:innerHTML]).to have_content(/.*Four.*Three.*Two/m)
     end
 
+    scenario "filter by multiple order cycles" do
+      # Set up 3 orders and 3 order cycles
```
luisramos0, line 42: this comment is too obvious, I think it should be removed

spec/features/admin/orders_spec.rb
```diff
@@ -38,6 +38,36 @@
...
+                                                  order_cycle: order_cycle_2)
+      order_3 = create(:order_with_credit_payment, user: user, distributor: distributor,
+                                                  order_cycle: order_cycle_3)
+      order_4 = create(:order_with_credit_payment, user: user, distributor: distributor,
+                                                  order_cycle: order_cycle_4)
+
+      quick_login_as_admin
+      visit 'admin/orders'
+
+      # Select order cycle 2 and 3
```
luisramos0, line 57: this comment is also obvious.

spec/features/admin/orders_spec.rb
```diff
@@ -38,6 +38,36 @@
...
+      multi_select2_select 'Two', from: 'q_order_cycle_id_in'
+      multi_select2_select 'Three', from: 'q_order_cycle_id_in'
+
+      # click link "Filter Results"
+      page.find('.filter-actions .button.icon-search').click
+
+      # Order 2 and 3 should show, but not 4
```
luisramos0, line 64: this comment is also obvious

spec/javascripts/unit/admin/orders/controllers/orders_controller_spec.js.coffee
```diff
@@ -51,3 +52,15 @@ describe "ordersCtrl", ->
 
       expect($scope.sorting).toEqual 'number asc'
       expect($scope.fetchResults).toHaveBeenCalled()
+
+  describe "filtering orders", ->
+    it "filters orders by all selected order cycles", ->
+
```
luisramos0, line 58: we dont usually have blank lines after the it "" line, this can be removed

luisramos0:
I'll keep this in code review until @dacook comes back with feedback about my minor comments.

dacook:
Thanks @luisramos0 , it wasn't as obvious for me ;)
Have made the suggested changes.

luisramos0:
ok @dacook  :+1: moving to test ready

filipefurtad0:
Hi @dacook - welcome! :tada: 

This looks great. I tested the Distributors and Order Cycles filters on Admin Orders page: the PR allows selecting more than one tag. Combining these filters also work as before.

Thank you for this improvement - ready to go.

PR #5317, opened 2020-04-28
[Caching] Timed caching on shops data

Matt-Yorkley:
#### What? Why?

Related to #5273 

Adds some short time-based caching on AMS data in `/shops` (index) and `/shop` (show), after it is fetched, serialized, and rendered.

**Note:** this data includes things like shop `name,` `description`, `twitter_handle`, `supplied_taxons`, and is cached for **up to 15 seconds**. It applies to some of the enterprise data displayed in both `/shops` and `/shop` pages.

#### What should we test?

Enterprise data in both `/shops` and `/shop` pages takes _up to_ 15 seconds (from the last time the page was loaded) to update, after it's changed in the backoffice. This should include things like enterprise name, description, and if the shop is open or closed.

Dev test: caching this data is working correctly. You can see clear messages in the logs with `log_level = :debug`, and see the queries are not being re-run. 

#### Release notes

Added 15-second caching on enterprise data in /shops and /shop pages

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -19,10 +19,8 @@ class HomeController < BaseController
   private
 
   # Cache the value of the query count
-  def cached_count(key, query)
-    CacheService.cache("home_stats_count_#{key}",
-                       expires_in: CacheService::HOME_STATS_EXPIRY,
-                       race_condition_ttl: 10) do
+  def cached_count(statistic, query)
+    CacheService.home_stats(statistic) do
       query.count
     end
   end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -18,9 +18,11 @@ class HomeController < BaseController
 
   private
 
-  # Cache the value of the query count for 24 hours
+  # Cache the value of the query count
   def cached_count(key, query)
-    Rails.cache.fetch("home_stats_count_#{key}", expires_in: 1.day, race_condition_ttl: 10) do
+    CacheService.cache("home_stats_count_#{key}",
+                       expires_in: CacheService::HOME_STATS_EXPIRY,
+                       race_condition_ttl: 10) do
       query.count
     end
   end
--- a/app/services/cache_service.rb
+++ b/app/services/cache_service.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class CacheService
+  HOME_STATS_EXPIRY = 1.day.freeze
   FILTERS_EXPIRY = 30.seconds.freeze
 
   def self.cache(cache_key, options = {})
--- a/app/services/cache_service.rb
+++ b/app/services/cache_service.rb
@@ -36,5 +36,19 @@ class CacheService
     def self.ams_all_properties_key
       "inject-all-properties-#{CacheService.latest_timestamp_by_class(Spree::Property)}"
     end
+
+    def self.ams_shops
+      [
+        "shops/index/inject_enterprises",
+        { expires_in: SHOPS_EXPIRY }
+      ]
+    end
+
+    def self.ams_shop(enterprise)
+      [
+        "enterprises/shop/inject_enterprise_shopfront-#{enterprise.id}",
+        { expires_in: SHOPS_EXPIRY }
+      ]
+    end
   end
 end
--- a/app/services/cache_service.rb
+++ b/app/services/cache_service.rb
@@ -3,6 +3,7 @@
 class CacheService
   HOME_STATS_EXPIRY = 1.day.freeze
   FILTERS_EXPIRY = 30.seconds.freeze
+  SHOPS_EXPIRY = 15.seconds.freeze
 
   def self.cache(cache_key, options = {})
     Rails.cache.fetch cache_key.to_s, options do
--- a/app/services/cache_service.rb
+++ b/app/services/cache_service.rb
@@ -25,6 +25,14 @@ class CacheService
     cached_class.maximum(:updated_at).to_i
   end
 
+  def self.home_stats(statistic)
+    Rails.cache.fetch("home_stats_count_#{statistic}",
+                      expires_in: HOME_STATS_EXPIRY,
+                      race_condition_ttl: 10) do
+      yield
+    end
+  end
+
   module FragmentCaching
     # Rails' caching in views is called "Fragment Caching" and uses some slightly different logic.
     # Note: supplied keys are actually prepended with "view/" under the hood.
--- a/app/services/cache_service.rb
+++ b/app/services/cache_service.rb
@@ -35,7 +35,7 @@ class CacheService
 
   module FragmentCaching
     # Rails' caching in views is called "Fragment Caching" and uses some slightly different logic.
-    # Note: supplied keys are actually prepended with "view/" under the hood.
+    # Note: keys supplied here are actually prepended with "views/" under the hood.
 
     def self.ams_all_taxons_key
       "inject-all-taxons-#{CacheService.latest_timestamp_by_class(Spree::Taxon)}"
--- a/app/views/enterprises/shop.html.haml
+++ b/app/views/enterprises/shop.html.haml
@@ -6,7 +6,8 @@
   = current_distributor.logo.url
 
 - content_for :injection_data do
-  = inject_enterprise_shopfront(@enterprise)
+  - cache "enterprises/shop/inject_enterprise_shopfront-#{@enterprise.id}", expires_in: CacheService::SHOPS_EXPIRY do
+    = inject_enterprise_shopfront(@enterprise)
 
 %shop.darkswarm
   - if @shopfront_layout == 'embedded'
--- a/app/views/enterprises/shop.html.haml
+++ b/app/views/enterprises/shop.html.haml
@@ -6,7 +6,7 @@
   = current_distributor.logo.url
 
 - content_for :injection_data do
-  - cache "enterprises/shop/inject_enterprise_shopfront-#{@enterprise.id}", expires_in: CacheService::SHOPS_EXPIRY do
+  - cache(*CacheService::FragmentCaching.ams_shop(@enterprise)) do
     = inject_enterprise_shopfront(@enterprise)
 
 %shop.darkswarm
--- a/app/views/shops/index.html.haml
+++ b/app/views/shops/index.html.haml
@@ -2,7 +2,7 @@
   = t :shops_title
 
 - content_for :injection_data do
-  - cache "shops/index/inject_enterprises", expires_in: CacheService::SHOPS_EXPIRY do
+  - cache(*CacheService::FragmentCaching.ams_shops) do
     = inject_enterprises(@enterprises)
 
 #panes
--- a/app/views/shops/index.html.haml
+++ b/app/views/shops/index.html.haml
@@ -2,7 +2,8 @@
   = t :shops_title
 
 - content_for :injection_data do
-  = inject_enterprises(@enterprises)
+  - cache "shops/index/inject_enterprises", expires_in: CacheService::SHOPS_EXPIRY do
+    = inject_enterprises(@enterprises)
 
 #panes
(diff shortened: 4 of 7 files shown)


Matt-Yorkley:
Notes on potential impact of this change:

This includes things like enterprise `name`, `description`, `email_address`, and whether or not they are open or closed. In some cases, like the drop-down section with more info when you click a shop on the `/shops` page, it won't be used at all (the data will be fresh). This caching applies from the moment the enterprise manager hits save, and will last _at most_ 15 seconds. If the page hasn't been accessed yet in the last 15 seconds, the data actually _will_ be updated without delay.

mkllnk:
@Matt-Yorkley Can you add some notes how to test this as a dev?

I installed memcached and added the production config to the dev environment:
```rb
# Use a different cache store in production
  memcached_value_max_megabytes = ENV.fetch("MEMCACHED_VALUE_MAX_MEGABYTES", 1).to_i
  memcached_value_max_bytes = memcached_value_max_megabytes * 1024 * 1024
  config.cache_store = :dalli_store, { value_max_bytes: memcached_value_max_bytes }
```

I don't understand though why the default file store wouldn't work.

I also changed the shop expiry to 1500000 seconds. Then I load the shop page, changed the enterprise name and reloaded the shop page. I can see all updates, no caching. I put `binding.pry` into `inject_enterprise_shopfront` to see if that's run and it runs every time. How do I activate caching?

I would like to test if the shopfront can break with stale data.

Matt-Yorkley:
After installing `memcached` and setting these configs everything worked for me:

```yaml
config.cache_store = :dalli_store
config.action_controller.perform_caching = true
```

app/controllers/home_controller.rb
```diff
@@ -18,9 +18,11 @@ def sell; end
 
   private
 
-  # Cache the value of the query count for 24 hours
+  # Cache the value of the query count
   def cached_count(key, query)
-    Rails.cache.fetch("home_stats_count_#{key}", expires_in: 1.day, race_condition_ttl: 10) do
+    CacheService.cache("home_stats_count_#{key}",
+                       expires_in: CacheService::HOME_STATS_EXPIRY,
+                       race_condition_ttl: 10) do
```
luisramos0, line 25: I think the service boundary here isnt good.
If CacheService knows about HOME_STATS_EXPIRY then it can know about it all. I'd do:
CacheService.home_stats_count do
    ...

otherwise I'd move HOME_STATS_EXPIRY to the home controller.

app/views/enterprises/shop.html.haml
```diff
@@ -6,7 +6,8 @@
   = current_distributor.logo.url
 
 - content_for :injection_data do
-  = inject_enterprise_shopfront(@enterprise)
+  - cache "enterprises/shop/inject_enterprise_shopfront-#{@enterprise.id}", expires_in: CacheService::SHOPS_EXPIRY do
```
luisramos0, line 9: same here, if SHOPS_EXPIRY is in the CacheService, at least the key should be in the service as well.

spec/services/cache_service_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      CacheService.stub(:latest_timestamp_by_class) { timestamp }
+    end
+
+    it "caches data by timestamp for last record of that class" do
+      CacheService.cached_data_by_class("test-cache-key", Enterprise) do
+        "TEST"
+      end
+
+      expect(CacheService).to have_received(:latest_timestamp_by_class).with(Enterprise)
+      expect(rails_cache).to have_received(:fetch).with("test-cache-key-Enterprise-#{timestamp}")
```
luisramos0, line 34: This spec validates the internals of the implementation, not the behavior.
We shouldnt stub internal methods of the class under test. Only it's external limits.
I'd expect Enterprise.maximum to be mocked here and the expectation is that the rails cache is called with that value.

```ruby
  before do
      expect(rails_cache.stub(:fetch)
      allow(Enterprise).to receive(:maximum).with(:updated_at).and_return(timestamp)
    end

    it "caches data by timestamp for last record of that class" do
      CacheService.cached_data_by_class("test-cache-key", Enterprise) do
        "TEST"
      end

      expect(rails_cache).to have_received(:fetch).with("test-cache-key-Enterprise-#{timestamp}")
```

app/views/shops/index.html.haml
```diff
@@ -2,7 +2,8 @@
   = t :shops_title
 
 - content_for :injection_data do
-  = inject_enterprises(@enterprises)
+  - cache "shops/index/inject_enterprises", expires_in: CacheService::SHOPS_EXPIRY do
+    = inject_enterprises(@enterprises)
```
luisramos0, line 6: I dont think we should go forward with this without feature tests.
I think we should test both that the cache is working as well as cache invalidation: load the page, wait a second, load again and assert data wasnt reloaded, change the data underneath, reload page and see the fresh data was reloaded.

luisramos0:
ok, I reviewed the 3 caching PRs. My conclusion is that I think we should definitely go ahead with all the 3 PRs, great work Matt. I dont think they can be dev-tests, we need feature tests doing the verification of all 3 PRs.

app/controllers/home_controller.rb
sauloperez: agree. I pointed something similar in the original PR.

app/views/shops/index.html.haml
Matt-Yorkley: I've just had a look at feature tests for this, but they won't work for any timed caches using the basic `:file_store` or `:memory_store` caching options. We would need to have `memcached` (or something similar) available in the machines the test suite runs on, and on any dev machines where the tests would be run locally.

Matt-Yorkley: I'm not sure how feasible that would be. It applies to 2 of the 3 PRs, but the "caching by class" one can be covered with feature tests.

Matt-Yorkley: Actually I'll check the timed caching with basic cache stores, the documentation seems to say conflicting things in different places, maybe it will work...

sauloperez: what about a controller test checking a hit and a miss stubbing the underlying store? I don't think it'll be feasible or makes sense getting any deeper than this.

Matt-Yorkley: When I try to stub the Rails cache I either get errors or it says it hasn't received any calls, even though I can see the cache reads and writes in `logs/test.log`... :disappointed:

Matt-Yorkley:
I managed to create a feature spec here :tada:. Looks like it's only the API caching that doesn't work in tests.

app/services/cache_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def self.home_stats(statistic)
+    Rails.cache.fetch("home_stats_count_#{statistic}",
+                      expires_in: HOME_STATS_EXPIRY,
+                      race_condition_ttl: 10) do
+      yield
+    end
+  end
+
+  module FragmentCaching
```
luisramos0, line 35: I wonder why is this mixed up with the previous code.
Looks like it would make a nice FragmentCacheService

spec/features/consumer/caching/shops_caching_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it "caches data for all enterprises, with the provided options" do
+      visit shops_path
+
+      key, options = CacheService::FragmentCaching.ams_shops
+      expect_cached "views/#{key}", options
+    end
+
+    it "keeps data cached for a short time on subsequent requests" do
+      # One minute ago...
+      Timecop.travel(Time.zone.now - 1.minute) do
```
mkllnk, line 19:
```suggestion
      Timecop.travel(1.minute.ago) do
```

I reckon that `ago` is a terrible language design on Rails' part but very convenient.

spec/features/consumer/caching/shops_caching_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        visit shops_path
+
+        expect(page).to have_content distributor.name
+
+        distributor.name = "New Name"
+        distributor.save
```
mkllnk, line 25: Always good to use `save!` in specs. If `save` fails this spec would always pass.

spec/features/consumer/caching/shops_caching_spec.rb
mkllnk: But would it be better to use `SHOPS_EXPIRY` here? One minute is a magic number. Maybe something like `SHOPS_EXPIRY + 2 * Capybara.default_max_wait_time`? The `2` is the number of page visits here.

Actually, if the server is really slow, `expect` can wait up to 30 seconds and then the spec fails. Flaky. If we are sure that this page is fast, we can reduce Capybara's wait time here to make sure that we are within the cache time. Or, and I like this idea more, we stub the caching time to be a multiple of Capybara's wait time.

spec/features/consumer/caching/shops_caching_spec.rb
Matt-Yorkley: Done.

spec/features/consumer/caching/shops_caching_spec.rb
Matt-Yorkley: `10.minutes.ago` should be safe for any theoretical flakyness, right? It's still a magic number, but we just need it to be longer than 30-150 seconds.

app/services/cache_service.rb
sauloperez: mmm doesn't this smell as merge conflict with your other PR? or am I looking at the wrong file?

spec/services/cache_service_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+describe CacheService do
+  let(:rails_cache) { Rails.cache }
+
+  describe "#cache" do
+    before do
+      rails_cache.stub(:fetch)
```
sauloperez, line 8: this syntax is from the old past. We should use `allow()`: https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/basics/allowing-messages

spec/features/consumer/caching/shops_caching_spec.rb
Matt-Yorkley: It hasn't actually failed on a slow machine yet, right?

app/services/cache_service.rb
Matt-Yorkley: Yes, I split the original PR into 4, but actually they all need to touch things or add sections in the same files :see_no_evil:

Matt-Yorkley: The merge conflicts should be very easy to resolve though.

spec/services/cache_service_spec.rb
Matt-Yorkley: Updated :+1:

spec/features/consumer/caching/shops_caching_spec.rb
luisramos0: It's a really good point Maikel!
I think using 10 or 30 minutes is enough here. I'd add a comment explaining the magic: basically a number bigger than cache time plus any capybara wait time.

Matt-Yorkley:
I think the feature tests cover the `dev-test` part here, which was just "check the caching is working". I'll move it to the other test column.

Matt-Yorkley:
I had to add an extra commit here to update the feature spec. :point_up: The test scenario was slightly overlapping the timed caching on shops data.

Matt-Yorkley:
Yeah it's not ideal for sure. One alternative is to ignore the visible elements in the page, and check the updated values directly in the HTML (for the rendered AMS we're testing). It works that way without interference from the other caching.

filipefurtad0:
Hi @Matt-Yorkley ,

Changing the name, description, twitter, supplier properties on the admin side take the expected action in both /shops and /shop pages, always up to 15 sec (way below, but will probably have a noticeable effect in production).

All good!

PR #5318, opened 2020-04-28
[Caching] API Action Caching on shop filters

Matt-Yorkley:
#### What? Why?

Related to #5273 

- Adds an includable module that optionally enables the use of ActionCaching in any API controller
- Adds ActionCaching on the `Api::OrderCyclesController` taxons and properties endpoints. 

**Note:** this data populates the filter buttons in an individual shop page for filtering a specific shop's product list. The data is cached for **up to 30 seconds**. It does not include the filter lists in the `/shops` index page.

#### What should we test?

Taxon and Property filter buttons are shown correctly when viewing the product list in the `/shop` page. If the taxons or properties are changed on a product, the buttons shown in the shop will take up to 30 seconds from the last time the page was loaded to update. Take care of this note when testing: https://github.com/openfoodfoundation/openfoodnetwork/pull/5318#issuecomment-620614110

**Dev test:** caching this data is working correctly. You can see clear messages in the logs with `log_level = :debug`, and see the queries are not being re-run.

#### Release notes

Added action caching to Api::OrderCycleCantroller taxons and properties endpoints

### DIFF
--- a/app/controllers/api/order_cycles_controller.rb
+++ b/app/controllers/api/order_cycles_controller.rb
@@ -5,7 +5,9 @@ module Api
 
     skip_authorization_check
 
-    caches_action :taxons, :properties, expires_in: CacheService::FILTERS_EXPIRY
+    caches_action :taxons, :properties,
+                  expires_in: CacheService::FILTERS_EXPIRY,
+                  cache_path: proc { |controller| controller.request.url }
 
     def products
       products = ProductsRenderer.new(
--- a/app/controllers/api/order_cycles_controller.rb
+++ b/app/controllers/api/order_cycles_controller.rb
@@ -1,9 +1,11 @@
 module Api
   class OrderCyclesController < Api::BaseController
     include EnterprisesHelper
-    respond_to :json
+    include ApiActionCaching
 
     skip_authorization_check
+
+    caches_action :taxons, :properties, expires_in: CacheService::FILTERS_EXPIRY
 
     def products
       products = ProductsRenderer.new(
--- a/app/models/concerns/api_action_caching.rb
+++ b/app/models/concerns/api_action_caching.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# API controllers inherit from ActionController::Metal to keep them slim and fast.
+# This concern adds the minimum requirements needed to use Action Caching in the API.
+
+module ApiActionCaching
+  extend ActiveSupport::Concern
+
+  included do
+    include ActionController::Caching
+    include ActionController::Caching::Actions
+    include AbstractController::Layouts
+
+    # This config is not passed to the controller automatically with ActionController::Metal
+    self.cache_store = Rails.configuration.cache_store
+
+    # ActionController::Caching asks for a controller's layout, but they're not used in the API
+    layout false
+  end
+end
--- a/app/models/concerns/api_action_caching.rb
+++ b/app/models/concerns/api_action_caching.rb
@@ -13,7 +13,7 @@ module ApiActionCaching
 
     # These configs are not assigned to the controller automatically with ActionController::Metal
     self.cache_store = Rails.configuration.cache_store
-    self.perform_caching = ActionController::Base.perform_caching
+    self.perform_caching = true
 
     # ActionController::Caching asks for a controller's layout, but they're not used in the API
     layout false
--- a/app/models/concerns/api_action_caching.rb
+++ b/app/models/concerns/api_action_caching.rb
@@ -11,8 +11,9 @@ module ApiActionCaching
     include ActionController::Caching::Actions
     include AbstractController::Layouts
 
-    # This config is not passed to the controller automatically with ActionController::Metal
+    # These configs are not assigned to the controller automatically with ActionController::Metal
     self.cache_store = Rails.configuration.cache_store
+    self.perform_caching = ActionController::Base.perform_caching
 
     # ActionController::Caching asks for a controller's layout, but they're not used in the API
     layout false
--- a/spec/features/consumer/caching/shops_caching_spec.rb
+++ b/spec/features/consumer/caching/shops_caching_spec.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+feature "Shops caching", js: true, caching: true do
+  include WebHelper
+  include UIComponentHelper
+
+  let!(:distributor) { create(:distributor_enterprise, with_payment_and_shipping: true, is_primary_producer: true) }
+  let!(:order_cycle) { create(:open_order_cycle, distributors: [distributor], coordinator: distributor) }
+
+  describe "API action caching on taxons and properties" do
+    let!(:taxon) { create(:taxon, name: "Cached Taxon") }
+    let!(:taxon2) { create(:taxon, name: "New Taxon") }
+    let!(:property) { create(:property, presentation: "Cached Property") }
+    let!(:property2) { create(:property, presentation: "New Property") }
+    let!(:product) { create(:product, taxons: [taxon], primary_taxon: taxon, properties: [property]) }
+    let(:exchange) { order_cycle.exchanges.to_enterprises(distributor).outgoing.first }
+
+    let(:test_domain) { "#{Capybara.current_session.server.host}:#{Capybara.current_session.server.port}" }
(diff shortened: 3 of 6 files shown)


Matt-Yorkley:
Notes on potential impact of this change:

In most cases this won't even be noticable, for example with 10 products in the `Vegetables` category and 10 in the `Fruit` category: if I remove the `Fruit` category from 3 products, the total list of all filters will still be `Fruit` and `Meat`, because there still exist products in both categories. But if I have 99 products in the `Meat` category and one single product in `Vegetables` category and I remove it, then it will take _at most_ 30 seconds before that change is shown in the shop.

lin-d-hop:
I feel like it would be great to also have a tester test these changes. Maybe just because all caching makes me nervous.

Would you agree @Matt-Yorkley? Even just a sanity check?
If so it would be great to put detailed testing notes for testers too.

Matt-Yorkley:
@lin-d-hop yes, I'm just in the process of writing it up and thinking about how best to test it.

Matt-Yorkley:
So: I can't get this to work in the test environment. It looks like Action Caching is not triggered in tests at all. The entries in this specific case are definitely not being written or read from the cache,in either feature or controller specs, even if caching is explicitly enabled, and even if I temporarily switch to different cache store providers, like `memcached`. I've tried stubbing various things instead, but haven't found any way to make a usable test. I think there is a lot of Rails "automagic" stuff going on under the hood, and I've been trying to pick it all part, but haven't achieved anything useful in 3-4 hours.

Here are some log entries from _outside_ of the test environment:

Screenshot from 2020-04-30 16-55-30
Screenshot from 2020-04-30 16-56-32

Matt-Yorkley:
Aside from the fact this doesn't work in the test environment, in this PR we're using a core Rails feature in quite a basic way, and very directly. Do we need to test basic use of core Rails features? It's incredibly easy to manually very that it's doing what it's supposed to.

luisramos0:
I think the test here would not test the rails feature but it would need to test that the OFN behavior is correct even if we use the rails feature.

I think it is very important we test this, specially if we want to keep re-using it. I think we could/should!
Have you tried using
config.action_controller.perform_caching = true
instead of 
ActionController::Base.perform_caching
I wonder if they are the same.

Matt-Yorkley:
> ActionController::Base.perform_caching

Aha! API Controllers don't inherit from the base controller... I'll have to add the same hack to `ActionController::Metal` as well.

Okay, I'll have another go.

luisramos0:
We have to love this code review process, I just throw a bit of a random idea and it makes you find the solution :-)

Matt-Yorkley:
I finally got a feature spec that works and tests the behavior :tada:

app/models/concerns/api_action_caching.rb
```diff
@@ -11,8 +11,9 @@ module ApiActionCaching
     include ActionController::Caching::Actions
     include AbstractController::Layouts
 
-    # This config is not passed to the controller automatically with ActionController::Metal
+    # These configs are not assigned to the controller automatically with ActionController::Metal
     self.cache_store = Rails.configuration.cache_store
+    self.perform_caching = Rails.configuration.action_controller.perform_caching
```
luisramos0, line 16: next step will be to make this an env var or something that we can easily enable/disable :+1:

spec/spec_helper.rb
```diff
@@ -117,9 +117,14 @@ def restart_phantomjs
   # Enable caching in any specs tagged with `caching: true`. Usage is exactly the same as the
   # well-known `js: true` tag used to enable javascript in feature specs.
   config.around(:each, :caching) do |example|
-    caching = ActionController::Base.perform_caching
+    caching = Rails.configuration.action_controller.perform_caching
+
+    Rails.configuration.action_controller.perform_caching = example.metadata[:caching]
     ActionController::Base.perform_caching = example.metadata[:caching]
```
luisramos0, line 123: have you verified if you still need  ActionController::Base.perform_caching  here?

Matt-Yorkley: It was needed for the other PRs, yes.

luisramos0: I meant if the other PRs work only with Rails.configuration.action_controller.perform_caching

Matt-Yorkley:
Hmmm okay... this is not fun at all.

- The `Rails.cache` object is passed around the app in a CONSTANT. The cache store is assigned to various places (like controllers) at various different times in a messy way :sob:
- If caching is not enabled at class load time, Rails ignores `caches_action` in a controller, so if we enable caching at test time it will be ignored :sob:
- Rails caching specs like this _pass_ in `spring` but _fail_ with `bundle exec rspec...` due to class loading weirdness: https://stackoverflow.com/a/45970527

app/controllers/api/order_cycles_controller.rb
```diff
@@ -1,10 +1,14 @@
 module Api
   class OrderCyclesController < Api::BaseController
     include EnterprisesHelper
-    respond_to :json
```
sauloperez, line 4: not needed anymore?

spec/features/consumer/caching/shops_caching_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      expect(page).to have_content taxon2.name
+      expect(page).to have_content property2.presentation
+    end
+  end
+
+  def expect_cached(key, options = {})
+    expect(Rails.cache.exist?(key, options)).to be true
+  end
+end
```
sauloperez, line 68: I see we test caching using feature specs most of the time. Given we only visit the shop page, is there something feature specs provide that controllers specs wouldn't? all these specs are adding up to our already slow test suite.

spec/services/cache_service_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper'
+
+describe CacheService do
+  let(:rails_cache) { Rails.cache }
+
+  describe "#cache" do
+    before do
+      rails_cache.stub(:fetch)
```
sauloperez, line 8: We shouls tick to RSpec 3's syntax and use [allow](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/basics/allowing-messages). It does the same you want to do here with `have_received`.

app/controllers/api/order_cycles_controller.rb
Matt-Yorkley: Nope, this is already declared in `Api::BaseController`.

spec/features/consumer/caching/shops_caching_spec.rb
Matt-Yorkley: Luis asked for feature specs specifically. I think it probably makes sense here to do a full integration test to verify this behavior in it's full context..?

spec/services/cache_service_spec.rb
Matt-Yorkley: Updated in: https://github.com/openfoodfoundation/openfoodnetwork/pull/5317

spec/features/consumer/caching/shops_caching_spec.rb
sauloperez: ok, :shrug: let's not forget about the testing pyramid. I'm a bit scared.

Matt-Yorkley: I think if we got into unit tests here we'd basically just be testing Rails features.

Matt-Yorkley: There's some unit tests for the `CacheService` methods.

filipefurtad0:
Hi @Matt-Yorkley ,

I did a general sanity test on how the changes on properties and taxons appear in the shopfront - all good.

I couldn't really observe the 30 sec delay in a reproducible way. Deleting a variant (admin) and refreshing the page in the shopfront (user) sometimes resulted in a delay, i.e., while the variant itself disappeared immediately, at times, the taxon remained. Clicking on it displayed no results, as expected.

I think this is ready for the dev testing - adding label and moving to the Dev-Test column.

Matt-Yorkley:
Awesome, thanks @filipefurtad0. Yes, it should be difficult to reproduce or notice! 

When lots of shoppers are online hopefully it will make a big difference in CPU usage. :tada:

filipefurtad0:
Yes! I'm sure it will have impact - thanks for the feedback @Matt-Yorkley ! :heart:

PR #5319, opened 2020-04-28
[Caching] Caching data by class

Matt-Yorkley:
#### What? Why?

Related to #5273 

- Adds some caching to queries and AMS arrays, where the results are a big array of objects from a single class. The expiration of the cache entries is based on the last time a record for that class was changed. 
- Adds some related `touch` calls for these cases.

Detailed explanation here: https://github.com/openfoodfoundation/openfoodnetwork/pull/5273#discussion_r413779989

#### What should we test?

Here we are caching some data related to taxons, properties (enterprise and product), and shipping_methods. It's not applied to a specific page, but should effect any non-admin pages. 

When:
- a taxon, property, or shipping_method are changed
- a shipping_method is added to an enterprise's list of shipping_methods
- a taxon is applied to a product (assigned as primary_taxon)

...then these changes should be reflected correctly in the frontend _without_ any delay, when the page is refreshed.

Dev test: caching this data is working correctly. You can see clear messages in the logs with `log_level = :debug`, and see the queries are not being re-run.

#### Release notes

Added class-based caching for large arrays of simple objects

### DIFF
--- a/app/views/layouts/darkswarm.html.haml
+++ b/app/views/layouts/darkswarm.html.haml
@@ -48,9 +48,9 @@
     = inject_current_hub
     = inject_current_user
     = inject_rails_flash
-    - cache "inject-all-taxons-#{CacheService.latest_timestamp_by_class(Spree::Taxon)}" do
+    - cache CacheService::FragmentCaching.ams_all_taxons_key do
       = inject_taxons
-    - cache "inject-all-properties-#{CacheService.latest_timestamp_by_class(Spree::Property)}" do
+    - cache CacheService::FragmentCaching.ams_all_properties_key do
       = inject_properties
     = inject_current_order
     = inject_currency_config
--- a/app/views/layouts/darkswarm.html.haml
+++ b/app/views/layouts/darkswarm.html.haml
@@ -48,8 +48,10 @@
     = inject_current_hub
     = inject_current_user
     = inject_rails_flash
-    = inject_taxons
-    = inject_properties
+    - cache "inject-all-taxons-#{CacheService.latest_timestamp_by_class(Spree::Taxon)}" do
+      = inject_taxons
+    - cache "inject-all-properties-#{CacheService.latest_timestamp_by_class(Spree::Property)}" do
+      = inject_properties
     = inject_current_order
     = inject_currency_config
     = yield :injection_data
--- a/spec/features/consumer/caching/darkwarm_caching_spec.rb
+++ b/spec/features/consumer/caching/darkwarm_caching_spec.rb
@@ -18,6 +18,18 @@ feature "Darkswarm data caching", js: true, caching: true do
 
   describe "caching injected taxons and properties" do
     it "caches taxons and properties" do
+      expect(Spree::Taxon).to receive(:all) { [taxon] }
+      expect(Spree::Property).to receive(:all) { [property] }
+
+      visit shops_path
+
+      expect(Spree::Taxon).to_not receive(:all)
+      expect(Spree::Property).to_not receive(:all)
+
+      visit shops_path
+    end
+
+    it "invalidates caches for taxons and properties" do
       visit shops_path
 
       taxon_timestamp1 = CacheService.latest_timestamp_by_class(Spree::Taxon)
--- a/spec/features/consumer/caching/darkwarm_caching_spec.rb
+++ b/spec/features/consumer/caching/darkwarm_caching_spec.rb
@@ -2,7 +2,7 @@
 
 require "spec_helper"
 
-feature "Caching", js: true, caching: true do
+feature "Darkswarm data caching", js: true, caching: true do
   let!(:taxon) { create(:taxon, name: "Cached Taxon") }
   let!(:property) { create(:property, presentation: "Cached Property") }
 
--- a/spec/features/consumer/caching_spec.rb
+++ b/spec/features/consumer/caching_spec.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+feature "Caching", js: true, caching: true do
+  let!(:taxon) { create(:taxon, name: "Cached Taxon") }
+  let!(:property) { create(:property, presentation: "Cached Property") }
+
+  let!(:producer) { create(:supplier_enterprise) }
+  let!(:distributor) { create(:distributor_enterprise, with_payment_and_shipping: true, is_primary_producer: true) }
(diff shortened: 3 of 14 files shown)


Matt-Yorkley:
Notes on potential impact:

These cache entries are auto-expired when an associated record is changed in the database. There is some additional DB overhead from `touch` calls when: 

- changing the taxons applied to a product
- assigning a shipping method to an enterprise

The CPU savings should greatly outweigh the small additional costs during these updates.

app/views/layouts/darkswarm.html.haml
```diff
@@ -48,8 +48,10 @@
     = inject_current_hub
     = inject_current_user
     = inject_rails_flash
-    = inject_taxons
-    = inject_properties
+    - cache "inject-all-taxons-#{CacheService.latest_timestamp_by_class(Spree::Taxon)}" do
+      = inject_taxons
+    - cache "inject-all-properties-#{CacheService.latest_timestamp_by_class(Spree::Property)}" do
+      = inject_properties
```
luisramos0, line 54: I think this must have feature tests.

Matt-Yorkley: Added.

spec/spec_helper.rb
```diff
@@ -114,6 +114,15 @@ def restart_phantomjs
...
 
+  # Enable caching in any specs tagged with `caching: true`. Usage is exactly the same as the
+  # well-known `js: true` tag used to enable javascript in feature specs.
+  config.around(:each, :caching) do |example|
+    caching = ActionController::Base.perform_caching
+    ActionController::Base.perform_caching = example.metadata[:caching]
+    example.run
+    ActionController::Base.perform_caching = caching
+  end
+
```
luisramos0, line 125: nice :ok_hand:

spec/features/consumer/caching_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      taxon.name = "Changed Taxon"
+      taxon.save
+      property.presentation = "Changed Property"
+      property.save
+
+      visit shops_path
+
+      taxon_timestamp2 = CacheService.latest_timestamp_by_class(Spree::Taxon)
+      expect_cached"views/#{CacheService::FragmentCaching.ams_all_taxons_key}"
```
luisramos0, line 44: there's a space missing between cached and "views... I wonder how this is working, is ruby smart enough to auto-fix this?

spec/features/consumer/caching_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      property_timestamp1 = CacheService.latest_timestamp_by_class(Spree::Property)
+      expect_cached "views/#{CacheService::FragmentCaching.ams_all_properties_key}"
+
+      toggle_filters
+
+      within "#hubs .filter-row" do
+        expect(page).to have_content taxon.name
+        expect(page).to have_content property.presentation
+      end
+
```
luisramos0, line 35: I think this test would be better if at this stage we verified the cache was actually used, so here I'd visit shops again and assert the cache was re-used and not re-newed.

spec/features/consumer/caching_spec.rb
Matt-Yorkley: Fixed :+1:

spec/features/consumer/caching_spec.rb
Matt-Yorkley: And in order to do that we have to start mocking the classes we're looking at in this integration test..? What's the best way to approach this?

spec/features/consumer/caching/darkwarm_caching_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+feature "Darkswarm data caching", js: true, caching: true do
+  let!(:taxon) { create(:taxon, name: "Cached Taxon") }
+  let!(:property) { create(:property, presentation: "Cached Property") }
+
+  let!(:producer) { create(:supplier_enterprise) }
+  let!(:distributor) { create(:distributor_enterprise, with_payment_and_shipping: true, is_primary_producer: true) }
```
luisramos0, line 10: I wonder why rubocop is not complaining about line too long here

spec/models/spree/shipping_method_spec.rb
```diff
@@ -109,5 +109,15 @@ module Spree
...
+
+    describe "touches" do
+      let!(:distributor) { create(:distributor_enterprise) }
+      let!(:shipping_method) { create(:shipping_method) }
+      let(:add_distributor) { shipping_method.distributors << distributor }
+
+      it "is touched when applied to a distributor" do
+        expect{ add_distributor }.to change { shipping_method.reload.updated_at}
+      end
+    end
```
luisramos0, line 121: I think these cache specs should be on their own file like shipping_method_cache_spec.rb

spec/models/spree/taxon_spec.rb
```diff
@@ -28,5 +28,23 @@ module Spree
         expect(Taxon.distributed_taxons(:current)).to eq(e.id => Set.new([t1.id]))
       end
     end
+
+    describe "touches" do
```
luisramos0, line 32: I think these cache specs should be on their own file like taxons_cache_spec.rb

spec/features/consumer/caching_spec.rb
luisramos0: :+1:

app/views/layouts/darkswarm.html.haml
```diff
@@ -48,8 +48,10 @@
     = inject_current_hub
     = inject_current_user
     = inject_rails_flash
-    = inject_taxons
-    = inject_properties
+    - cache CacheService::FragmentCaching.ams_all_taxons_key do
```
sauloperez, line 51: isn't it too redundant? what about `CacheService::Fragment`

spec/features/consumer/caching/darkwarm_caching_spec.rb
sauloperez: very likely that this file is whitelisted in rubocop config.

sigmundpetersen: Is it a typo in the file name as well here? darkwarm -> darkswarm?
If this file is added as part of this PR maybe it could be updated?

filipefurtad0:
Hi @Matt-Yorkley ,

Added and removed:
- a new taxon, property, or shipping_methods
- taxons to existing produtcs

All these changes were visible without delay, after the first page refresh.
Ready to go!

PR #5316, opened 2020-04-28
[Spree 2.1] Merge master into 3-0-stable

luisramos0:
It took me 1.5 hours of merging stuff. I think it went well.

Here are the conflicts and how I soolved each of them:
<details>

> 
> CONFLICT (modify/delete): spec/services/subscription_form_spec.rb deleted in master and modified in HEAD. Version HEAD of spec/services/subscription_form_spec.rb left in tree.
> 	re-apply 53645517af834f678b17e3094921267c353b2f8c that changed spec/services/subscription_form_spec.rb in engines/order_management/spec/services/order_management/subscriptions/form_spec.rb
> 	git rm spec/services/subscription_form_spec.rb
> 	git add engines/order_management/spec/services/order_management/subscriptions/form_spec.rb
> 
> CONFLICT (modify/delete): lib/open_food_network/proxy_order_syncer.rb deleted in master and modified in HEAD. Version HEAD of lib/open_food_network/proxy_order_syncer.rb left in tree.
> 	reapply 98fdbb86218b2c7a2e0f7722832385042fcd2ec2 that changed lib/open_food_network/proxy_order_syncer.rb in engines/order_management/app/services/order_management/subscriptions/proxy_order_syncer.rb
> 	git rm lib/open_food_network/proxy_order_syncer.rb
> 	git add engines/order_management/app/services/order_management/subscriptions/proxy_order_syncer.rb
>   
> CONFLICT (modify/delete): app/services/subscription_form.rb deleted in master and modified in HEAD. Version HEAD of app/services/subscription_form.rb left in tree.
> 	re-apply c3897b2f1c0776ca4e16491900659fa55768cea0 that changed app/services/subscription_form.rb in engines/order_management/app/services/order_management/subscriptions/form.rb
> 	git rm app/services/subscription_form.rb
> 	git add engines/order_management/app/services/order_management/subscriptions/form.rb
> 
> CONFLICT (content): Merge conflict in spec/services/order_cycle_form_spec.rb
> 	keep both
> 	git add spec/services/order_cycle_form_spec.rb
> 
> CONFLICT (content): Merge conflict in spec/features/admin/orders_spec.rb
> 	keep master
> 	apply 80e56084367972baf4fc7580f023792173693cbe to spec/features/admin/order_spec.rb
> 	git add spec/features/admin/orders_spec.rb
> 	git add spec/features/admin/order_spec.rb
> 
> CONFLICT (content): Merge conflict in lib/open_food_network/orders_and_fulfillments_report.rb
[...]

PR #5320, opened 2020-04-28
[Spree 2.1] Fix variant overrides specs and shops spec

luisramos0:
#### What? Why?

Two last commits to make the 3-0-stable build green.
The first commit fixes a problem in the variant_overrides_controller by making load_data run before load_resource (in resource_controller).
The second commit fixes a problem in the Enterprise scope that is breaking the closed shop controller spec:  .present? calls .length that breaks with ActiveModel::MissingAttributeError Exception: missing attribute: address_id (probably trying to load the collection from the db?). Calling any? works well.

#### What should we test?
Together with #5316 I think this makes the 3-0-stable build green.

### DIFF
--- a/app/models/enterprise.rb
+++ b/app/models/enterprise.rb
@@ -118,7 +118,7 @@ class Enterprise < ActiveRecord::Base
       except(:select).
       select('DISTINCT enterprises.id')
 
-    if ready_enterprises.present?
+    if ready_enterprises.any?
       where("enterprises.id NOT IN (?)", ready_enterprises)
     else
       where(nil)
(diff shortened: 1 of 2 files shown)


app/models/enterprise.rb
```diff
@@ -118,7 +118,7 @@ class Enterprise < ActiveRecord::Base
       except(:select).
       select('DISTINCT enterprises.id')
 
-    if ready_enterprises.present?
+    if ready_enterprises.any?
```
sauloperez, line 121: `any?` will perform a COUNT thus, the `else` branch at line 124 seems pointless to me.

PR #5322, opened 2020-04-28
Delete some dead views

luisramos0:
#### What? Why?

It's incredible how these things hide in our codebase...

#### What should we test?
The homepage should render correctly.

#### Release notes
Changelog Category: Removed
Delete some dead code from old homepage.


filipefurtad0:
Hi @luisramos0 

The home page looks the same as before to my human eyes - the source code of the home-page was identical as well: that was indeed dead code :-)

Thanks for spotting this!

Issue #5306, opened 2020-04-27
Tax report is showing order.created_at instead of completed_at


RachL:
## Description

When looking at the report `/admin/reports/sales_tax` the date shown is `order.updated_at`, we need it to be set at `completed_at` which would then be the same date as shown in `/admin/orders?q[s]=completed_at+desc`. 

## Steps to Reproduce
1. Connected as a manager, check the dates of the report `/admin/reports/sales_tax`

2. See some of them are different from `/admin/orders?q[s]=completed_at+desc` for the same order number

## Animated Gif/Screenshot
Example in French production :

Order page:

![image](https://user-images.githubusercontent.com/6525576/80372479-e1e9fe00-8893-11ea-821f-c0951b5c4e26.png)

Tax report:

![image](https://user-images.githubusercontent.com/6525576/80372447-d5fe3c00-8893-11ea-833a-949f8833d939.png)

## Severity
bug-s2: a non-critical feature is broken, no workaround

## Your Environment
* Version used: v2.9.5
* Browser name and version: Firefox 76
* Operating System and version (desktop or mobile):Desktop

kirstenalarsen:
Disagree that this is s2. It has always been like this and no one noticed until now

RachL:
@kirstenalarsen I don't kbow if you read the full discussion on slack but the reason for it to be tagged S2 was that there is no workaround (no other reports are showing the date) and that it was a really simple job to do (2min).

PR #5325, opened 2020-04-29
Add order number and date/time to order cycles customer report

### LINKED ISSUE #5214: Add Order number and date/time of order completion to Order Cycles - Customer Totals report
## Description

- As an enterprise user
- On page admin/reports Customer Totals report - https://openfoodnetwork.org.au/admin/reports/orders_and_fulfillment
- I want to be able to see the Order number and the date/time of order completion for each order

These new fields should be added to each row and they should be created as two new columns at the far right - so the last two fields of the table (so that they don't mess up any existing macros people have created using this report) 
 
1. Run the Order Cycle - Customer Totals report
2. See the Order Number and date/time of order completion on each row of the table, at the far right (last two columns)

oeoeaio:
#### What? Why?

Closes #5214 

Adds the order `number` and `completed_at` date to the far right hand side of the order cycles customer totals report. This is so as to maintain the existing position of other columns in case users have developed tools that depend on the order of the columns.

#### What should we test?
(From issue):
1. Run the Order Cycle - Customer Totals report
2. See the Order Number and date/time of order completion on each row of the table, at the far right (last two columns)

#### Release notes
- Order number and date/time of completion are now displayed at the far right hand side of the order cycle customer totals report. 

Changelog Category: Added

### DIFF
--- a/lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
+++ b/lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
@@ -36,7 +36,9 @@ module OpenFoodNetwork
          I18n.t(:report_header_customer_code), I18n.t(:report_header_tags),
          I18n.t(:report_header_billing_street), I18n.t(:report_header_billing_street_2),
          I18n.t(:report_header_billing_city), I18n.t(:report_header_billing_postcode),
-         I18n.t(:report_header_billing_state)]
+         I18n.t(:report_header_billing_state),
+         I18n.t(:report_header_order_number),
+         I18n.t(:report_header_date)]
       end
       # rubocop:enable Metrics/AbcSize
       # rubocop:enable Metrics/MethodLength
@@ -92,7 +94,9 @@ module OpenFoodNetwork
               proc { |_line_items| "" },
               proc { |_line_items| "" },
               proc { |_line_items| "" },
-              proc { |_line_items| "" }
+              proc { |_line_items| "" },
+              proc { |line_items| line_items.first.order.number },
+              proc { |line_items| line_items.first.order.completed_at.strftime("%F %T") },
             ]
           },
           {
@@ -187,7 +191,9 @@ module OpenFoodNetwork
           proc { |line_items| line_items.first.order.bill_address.andand.address2 },
           proc { |line_items| line_items.first.order.bill_address.andand.city },
           proc { |line_items| line_items.first.order.bill_address.andand.zipcode },
-          proc { |line_items| line_items.first.order.bill_address.andand.state }
+          proc { |line_items| line_items.first.order.bill_address.andand.state },
+          proc { |line_items| line_items.first.order.number },
+          proc { |line_items| line_items.first.order.completed_at.strftime("%F %T") },
         ]
       end
       # rubocop:enable Metrics/AbcSize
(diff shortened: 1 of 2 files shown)


lib/open_food_network/orders_and_fulfillments_report/customer_totals_report.rb
```diff
@@ -92,7 +94,9 @@ def rules
               proc { |_line_items| "" },
               proc { |_line_items| "" },
               proc { |_line_items| "" },
-              proc { |_line_items| "" }
+              proc { |_line_items| "" },
+              proc { |line_items| line_items.first.order.number },
+              proc { |line_items| line_items.first.order.completed_at.strftime("%F %T") },
```
luisramos0, line 99: This approach to format the date looks ok, we have two more ways to render dates in reports:
- including timezone as in this to_local_time here:
https://github.com/openfoodfoundation/openfoodnetwork/blob/ed15bc422e92ea5cd8289f9d13b2a9c135709056/lib/open_food_network/users_and_enterprises_report.rb#L128
- and in Order and distributor report and sales tax report we are just using the date field directly: https://github.com/openfoodfoundation/openfoodnetwork/blob/ed15bc422e92ea5cd8289f9d13b2a9c135709056/lib/open_food_network/order_and_distributor_report.rb#L82 and https://github.com/openfoodfoundation/openfoodnetwork/blob/407d890d23c2001ad3060eec6633ab8bd1783ab3/lib/open_food_network/sales_tax_report.rb#L60

Changing the created_at to completed_at on this last sales_tax_report would give us a bonus issue #5306 :-)

Maybe we should fix all these in one go: make all use completed_at (all these reports look only at completed orders) and use this formatting you have introduced in this PR.

luisramos0: I wonder if we should get rid of the timezone part in users_and_enterprises_report.rb or use it everywhere.

oeoeaio: > I wonder if we should get rid of the timezone part in users_and_enterprises_report.rb or use it everywhere.

My assumption would be that Rails parses datetimes into the local time zone anyway? The time zone is not actually being rendered in the output of `to_local_datetime`, so I don't think the `in_time_zone` has a big impact...?

I think we can probably use `order.completed_at.strftime("%F %T")` pretty much universally.

luisramos0: cool, let's do that :+1:

Matt-Yorkley:
LGTM :+1:

>  in case users have developed tools that depend on the order of the columns

Maybe we should ping @RachL and @lin-d-hop in case they have any relevant hacks here on exported reports...

luisramos0:
Test Ready?

RachL:
I do need that OC number and date to make my spreadsheets more secure and correct as I'm trying to match data from other reports. So I'm all in :)
If we could have this on packing sheets as well that would be heaven :) I will add it as papercut as well

filipefurtad0:
Hi @oeoeaio ,

The columns Order Number and Date now appear, on the far right side of the report. Newly placed orders show up correctly, as well - all good.

Thanks for this improvement!

PR #5329, opened 2020-04-29
Correct translations

romale:
Without it, some labels not translated to other language.
And, please check code:
en.admin.subscription.subscriptions - does not exists
but
en.admin.subscriptions.subscriptions - is exist

#### What? Why?

Because some translation does not work

#### What should we test?

Translation

#### Release notes

Changelog Category: Added | **Changed** | Deprecated | Removed | **Fixed** | Security

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -914,6 +914,11 @@ en:
         save_and_back_to_list: "Save and Back to List"
         choose_products_from: "Choose Products From:"
       incoming:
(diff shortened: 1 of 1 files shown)


filipefurtad0:
Hi @romale,

I had a look at the page OC and checked the rendering of the pages which this translation takes effect - I found issues in rendering, all should be fine here.

Thank you for this contribution! :tada: 

Moving to ready to go.

PR #5332, opened 2020-04-30
Add timestamps to OC schedule join table

### LINKED ISSUE #4462: Add Logging to Subscriptions

sauloperez:
#### What? Why?

Closes #4462

This is critical to debugging bugs related to subscriptions.

Essentially, `has_and_belongs_to_many` doesn't give us the option for any other column that the foreign keys themselves:

> A has_and_belongs_to_many association creates a direct many-to-many connection with another model, with no intervening model.

Source: https://guides.rubyonrails.org/v3.2/association_basics.html#the-has_and_belongs_to_many-association

Note, however, that there's no way to update an order_cycle_schedule, that I can think of but `updated_at` doesn't do any harm.

#### What should we test?

We need to ensure that a subscription can be created and modified as usual.

#### Release notes

The `order_cycle_schedules` join table has `created_at` and `updated_at` columns.

Changelog Category: Added

### DIFF
--- a/app/models/order_cycle.rb
+++ b/app/models/order_cycle.rb
@@ -16,7 +16,8 @@ class OrderCycle < ActiveRecord::Base
   has_many :suppliers, source: :sender, through: :cached_incoming_exchanges, uniq: true
   has_many :distributors, source: :receiver, through: :cached_outgoing_exchanges, uniq: true
 
-  has_and_belongs_to_many :schedules, join_table: 'order_cycle_schedules'
+  has_many :schedules, through: :order_cycle_schedules
+  has_many :order_cycle_schedules
   has_paper_trail meta: { custom_data: proc { |order_cycle| order_cycle.schedule_ids.to_s } }
 
   attr_accessor :incoming_exchanges, :outgoing_exchanges
--- a/app/models/order_cycle_schedule.rb
+++ b/app/models/order_cycle_schedule.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class OrderCycleSchedule < ActiveRecord::Base
+  belongs_to :schedule
+  belongs_to :order_cycle
+end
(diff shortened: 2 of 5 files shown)


Matt-Yorkley:
Do we need a quick `dev-test` as well as a regular test here?

luisramos0:
yeah, I think so. Check the new timestamps in the db after assigning schedules to OCs in the Edit schedule modal.

sauloperez:
well, that feels like testing Rails to me and I already did it but fine for me :+1:

Matt-Yorkley:
I did a quick check and the timestamp was there. Moving to `test-ready` for the sanity check on creating and updating subs :+1:

sauloperez:
This is sufficiently covered by specs IMO

luisramos0:
I'd still do the test ready part. I am ok with merging without it though if you want to include it in this release :+1:

sauloperez:
I tested we can create a schedule, link to and unlink it from an OC and a subscription can be created as well. All good here.

Issue #5173, opened 2020-04-06
Shop can see whether an order has refund owing on Orders page


kirstenalarsen:
## Description

- As an enterprise user (shop or hub)
- On page admin/orders
- I want to be able to see which orders have refunds owing 

Currently when an order is paid it looks like this
![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/a7bef105-5e61-4047-af4f-43fbf9295aa7)

and when a change has been made to the order so the customer is owed money (or has an account credit) it displays like this
![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/c1824032-8d3c-42c7-b82f-8a7046f1aaf3)

The Total column shows the revised total of the order, but I cannot see how much I actually owe them, in this case $5
![image.png](https://images.zenhubusercontent.com/54b83fb37c6ccc6277a6ea46/a57b6bd6-c9b1-4e83-91cb-d7d9dbed0944)

Could we print this amount on the Orders page somewhere, so that the Shop can at least see at a glance what they owe. 

Suggested solution to print the amount in brackets after the link in the Payment State column, so 
- Credit Owed (40.00)
- Balance Due ($40.00)
not sure if there should be a -ve sign?

- ?? leads to ability to process refund from this page as per [wishlist item](https://community.openfoodnetwork.org/t/batch-processing-of-refunds/1893). I started creating an issue to do a single one but think it might be a  bit tricky

 
## Acceptance Criteria & Tests

1. Create a customer order and accept the payment
2. Remove an item from the order
3. Go to admin/orders and filter to see that order (if necessary)
4. The amount owing to the customer should show on the page, so I can easily see how much needs to be refunded, or how much has been added to their account credit

luisramos0:
papercut size :+1:
related to https://github.com/openfoodfoundation/openfoodnetwork/pull/5031/files

mathewdbutton:
Hey @kirstenalarsen I was going to try and pick this issue up if that's cool.

A quick mock up of what I was thinking
![Screen Shot 2020-04-28 at 10 28 22 pm](https://user-images.githubusercontent.com/1100707/80487107-9f154c80-899f-11ea-8f94-83eea9127f8b.png)

I wondered if the text should go underneath the payment state to take into consideration any refunds/balance due that run several digits long and squish up the table.

>Credit Owed (40.00)
>Balance Due ($40.00)
[...]

PR #5341, opened 2020-05-01
Show outstanding balance on orders page

mathewdbutton:
#### What? Why?

Closes [#5173]

Prior to this change, there was no way of knowing how much was owed (or
in need of refunding) at a glance for each order on the orders index page,
this information was only available by clicking through to an order
payment page

This change adds the outstanding balance for each order.

There was some discussion about whether a refund should be a -ve number or not, I chose to keep it as a -ve number to keep it consistent with the information displayed on the payments page.

#### What should we test?

- Make an order with shipping at a shop as a cash payment
- Navigate to the order tab in the admin portal, you should see the balance outstanding for that order underneath the payment state. 
- Now capture the payment for that order 
- Change the order to pick up instead of delivery, the order should now display the amount required for a refund.

Screen Shot 2020-05-01 at 9 50 18 pm

#### Release notes
Show outstanding balance for each order on orders index page.

Changelog Category: Added

### DIFF
--- a/app/serializers/api/admin/order_serializer.rb
+++ b/app/serializers/api/admin/order_serializer.rb
@@ -19,7 +19,7 @@ class Api::Admin::OrderSerializer < ActiveModel::Serializer
   def display_outstanding_balance
     return "" unless ["balance_due", "credit_owed"].include?(object.payment_state)
 
-    "(#{object.display_outstanding_balance.to_html})"
+    object.display_outstanding_balance.to_s
   end
 
   def edit_path
--- a/app/serializers/api/admin/order_serializer.rb
+++ b/app/serializers/api/admin/order_serializer.rb
@@ -3,7 +3,7 @@ class Api::Admin::OrderSerializer < ActiveModel::Serializer
              :edit_path, :state, :payment_state, :shipment_state,
              :payments_path, :ready_to_ship, :ready_to_capture, :created_at,
              :distributor_name, :special_instructions,
-             :item_total, :adjustment_total, :payment_total, :total
+             :item_total, :adjustment_total, :payment_total, :total, :display_outstanding_balance
 
   has_one :distributor, serializer: Api::Admin::IdSerializer
   has_one :order_cycle, serializer: Api::Admin::IdSerializer
@@ -14,6 +14,12 @@ class Api::Admin::OrderSerializer < ActiveModel::Serializer
 
   def distributor_name
     object.distributor.andand.name
+  end
+
+  def display_outstanding_balance
+    return "" unless ["balance_due", "credit_owed"].include?(object.payment_state)
+
+    "(#{object.display_outstanding_balance.to_html})"
   end
 
   def edit_path
--- a/app/serializers/api/admin/order_serializer.rb
+++ b/app/serializers/api/admin/order_serializer.rb
@@ -17,7 +17,7 @@ class Api::Admin::OrderSerializer < ActiveModel::Serializer
   end
 
   def display_outstanding_balance
-    return "" unless ["balance_due", "credit_owed"].include?(object.payment_state)
+    return "" if object.outstanding_balance.zero?
 
     object.display_outstanding_balance.to_s
   end
--- a/app/views/spree/admin/orders/index.html.haml
+++ b/app/views/spree/admin/orders/index.html.haml
@@ -68,7 +68,7 @@
         %span.state{'ng-class' => 'order.payment_state', 'ng-if' => 'order.payment_state'}
           %a{'ng-href' => '{{order.payments_path}}' }
             {{'js.admin.orders.payment_states.' + order.payment_state | t}}
-            %span{'ng-if' =>'order.display_outstanding_balance'}
+            %span{'ng-if' => 'order.display_outstanding_balance'}
               ({{order.display_outstanding_balance}})
       %td.align-center
         %span.state{'ng-class' => 'order.shipment_state', 'ng-if' => 'order.shipment_state'}
--- a/app/views/spree/admin/orders/index.html.haml
+++ b/app/views/spree/admin/orders/index.html.haml
@@ -68,6 +68,7 @@
         %span.state{'ng-class' => 'order.payment_state', 'ng-if' => 'order.payment_state'}
           %a{'ng-href' => '{{order.payments_path}}' }
             {{'js.admin.orders.payment_states.' + order.payment_state | t}}
+            {{order.display_outstanding_balance}}
       %td.align-center
         %span.state{'ng-class' => 'order.shipment_state', 'ng-if' => 'order.shipment_state'}
           {{'js.admin.orders.shipment_states.' + order.shipment_state | t}}
--- a/app/views/spree/admin/orders/index.html.haml
+++ b/app/views/spree/admin/orders/index.html.haml
@@ -68,7 +68,8 @@
         %span.state{'ng-class' => 'order.payment_state', 'ng-if' => 'order.payment_state'}
           %a{'ng-href' => '{{order.payments_path}}' }
             {{'js.admin.orders.payment_states.' + order.payment_state | t}}
-            {{order.display_outstanding_balance}}
+            %span{'ng-if' =>'order.display_outstanding_balance'}
+              ({{order.display_outstanding_balance}})
       %td.align-center
         %span.state{'ng-class' => 'order.shipment_state', 'ng-if' => 'order.shipment_state'}
           {{'js.admin.orders.shipment_states.' + order.shipment_state | t}}
--- a/spec/factories/order_factory.rb
+++ b/spec/factories/order_factory.rb
@@ -46,8 +46,12 @@ FactoryBot.define do
     distributor { create(:distributor_enterprise) }
     order_cycle { create(:simple_order_cycle) }
 
-    after(:create) do |order|
-      create(:payment, amount: order.total + 10_000, order: order, state: "completed")
+    transient do
+      credit_amount { 10_000 }
+    end
+
+    after(:create) do |order, evaluator|
+      create(:payment, amount: order.total + evaluator.credit_amount, order: order, state: "completed")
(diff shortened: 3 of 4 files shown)


lin-d-hop:
A few ideas for edge cases on Order Balance - might be useful in testing:

- completed orders, cancelled orders
- multiple payments with different payment methods
- voiding payments
- adding things and cancelling things on the order including line items and different charges
- payment methods with fees and without fees

app/serializers/api/admin/order_serializer.rb
```diff
@@ -16,6 +16,12 @@ def distributor_name
     object.distributor.andand.name
   end
 
+  def display_outstanding_balance
+    return "" unless ["balance_due", "credit_owed"].include?(object.payment_state)
+
+    "(#{object.display_outstanding_balance.to_html})"
```
luisramos0, line 22: the parenthesis here are presentation, so they should be in the view not in the serializer. We may want to show this value somewhere else without the parenthesis.
So this last line becomes:
`object.display_outstanding_balance.to_html`

app/serializers/api/admin/order_serializer.rb
```diff
@@ -16,6 +16,12 @@ def distributor_name
     object.distributor.andand.name
   end
 
+  def display_outstanding_balance
+    return "" unless ["balance_due", "credit_owed"].include?(object.payment_state)
```
luisramos0, line 20: There's some biz logic here that would be better in a service. if we do that, we could, at a later point, add to that service the logic in OrderUpdater#infer_payment_state).

It's not mandatory for this PR but ideally we would have this line like
    return "" unless PaymentStateService(object).outstanding_balance?

And then, in that new service, we would handle the complexity of outstanding balance versus payment_state.

I wrote this as a comment, not in scope for this PR, BUT then I went to the UK live database and checked these fields...
- I see 25 orders where payment_state is **credit_owed** and outstanding balance is 0
- I see 29 orders where payment_state is **balance_due** and outstanding balance is negative (payment total > total)
- I see 76 completed orders where payment_state is **failed** and there is a balance due (payment total < total)

I think we need to ignore payment_state in this code and just use object.outstanding_balance != 0

app/serializers/api/admin/order_serializer.rb
Matt-Yorkley: @luisramos0 I'm not sure sure about this part: 

> the to_html should probably go to the view as well

`Spree::Order#display_outstanding_balance` returns a `Spree::Money` object which includes custom implementations for both `#to_s` and `#to_html`. I think in this case we probably do want to call one of these in the serializer side rather than in the view. `#to_s` is maybe the better option though. What do you think?

Agreed on the parenthesis moving to the view :+1:

luisramos0: yeah, agree, one of them must be used.
I'd prefer to_s, but if it's too much trouble we can keep to_html.

mathewdbutton: > the parenthesis here are presentation, so they should be in the view not in the serializer. We may want to show this value somewhere else without the parenthesis.
> So this last line becomes:
> `object.display_outstanding_balance`
> 

@luisramos0 Yes good point, I seem to have forgotten that this isn't a decorator for a view, it's the api response. I'll add something to the angular side of things generate the parenthesis if there is a value present.

> EDIT: the to_html should probably go to the view as well, we dont want html encoded characters in the json payload, do we?

This was me trying to see how other places use some of the spree models, I found that it's used here https://github.com/openfoodfoundation/openfoodnetwork/blob/8087e6b31b1e3a0a881cbaa0c2f67d3db45b8370/app/serializers/api/admin/order_serializer.rb#L47
But you make a good point this is silly, I should have just gone to `to_s` I was just not sure if there was some convention.

app/serializers/api/admin/order_serializer.rb
mathewdbutton: Right so it seems like there are a whoooole bunch of edge cases I've clearly failed to think of.

Yeah this is much better, I've just had a wee hunt around the code base and it seems we do something similar here https://github.com/openfoodfoundation/openfoodnetwork/blob/b568d254467c4df1c891b917104599af3a4fdc97/app/views/spree/admin/payments/index.html.haml#L15 so there is definitely some scope to add this kind of logic to a PaymentStateService. But as you say it might be out of scope for this issue.

mathewdbutton:
> A few ideas for edge cases on Order Balance - might be useful in testing:
> 
>     * completed orders, cancelled orders
> 
>     * multiple payments with different payment methods
> 
>     * voiding payments
> 
>     * adding things and cancelling things on the order including line items and different charges
> 
>     * payment methods with fees and without fees


@lin-d-hop Thanks for the suggestions. I've tested a bunch of these and they seem to produce expected results.

app/serializers/api/admin/order_serializer.rb
mathewdbutton: @luisramos0  So I've just been testing this out locally with a few more order/payment combinations
Screen Shot 2020-05-03 at 2 21 45 pm

Seems to work as expected, I'm just wondering about the 'Failed' Payment state and whether it makes sense to have the outstanding balance shown there or not? What do you think?

app/serializers/api/admin/order_serializer.rb
mathewdbutton: I've made the change to use `to_s` and to add the parentheses in the angular view instead. I wasn't sure if I needed to amend my commit with these changes or make a new commit (I made a new one to not lose the comment stream)

luisramos0: Sometimes the existing code is a good example, sometimes it's not :-)
I dont think it's necessarily wrong to put html encoded data in the api payload, but it seems we agree to_s is better so we can use it here and start doing that everywhere, including changing that order_serializer to_html at some point.

We dont always follow this rule but I think if a review is going on, the author should not clear the commit history so that the reviewers can see what happened. The author can clear up the commit history after the discussion is over, if it's not too much trouble.

app/serializers/api/admin/order_serializer.rb
luisramos0: I like this table.
I think we should show the balance whenever it is not zero, regardless of the payment_state.

mathewdbutton: I've updated the condition to not use payment state. Thanks for the input @luisramos0

app/serializers/api/admin/order_serializer.rb
```diff
@@ -16,6 +16,12 @@ def distributor_name
     object.distributor.andand.name
   end
 
+  def display_outstanding_balance
+    return "" if object.outstanding_balance.zero?
```
luisramos0, line 20: Looking at this now, why not just return 0 and adapt the angular code with != 0?

mathewdbutton: Hah.... you're right I could do this, leave the view logic to the front end again. I'll try and get onto this tomorrow and the cleanup the commits. 

Thanks for the help @luisramos0

mathewdbutton: I was just thinking about this some more. We're using `outstanding_balance` to figure out if the balance is zero but we're sending a different value, `display_outstanding_balance.to_s` in the API response, which will be "$0.00" or whatever the currency symbol is and the dollar amount, which might make this small change tricker than I thought if we need to check if outstanding_balance == <currency_symbol>0.00 .  

I think this is where it would be more useful if the front end formatted the amount in the correct locale instead of it coming from the back end, but that might be a task for another time.

luisramos0: ah, yes, I agree. Let's keep it as is it :+1:

app/views/spree/admin/orders/index.html.haml
```diff
@@ -68,6 +68,8 @@
         %span.state{'ng-class' => 'order.payment_state', 'ng-if' => 'order.payment_state'}
           %a{'ng-href' => '{{order.payments_path}}' }
             {{'js.admin.orders.payment_states.' + order.payment_state | t}}
+            %span{'ng-if' =>'order.display_outstanding_balance'}
```
luisramos0, line 71: there's a space missing after =>

daniellemoorhead:
@mathewdbutton once you've sorted out whether you do the missing space after => we can move this to test ready and have someone test it. 🎉 Not sure if you have the power to move something into the test ready column, if not yell out and one of us can do it for you.

app/views/spree/admin/orders/index.html.haml
mathewdbutton: Done!

mathewdbutton:
Hi @daniellemoorhead, formatting all sorted. Uhh I don't know if I do, I'm not sure where I would go to move this ticket so I might need some help, thanks.

daniellemoorhead:
It's in test ready @mathewdbutton, thanks to @Matt-Yorkley sorting it out - thank you for finishing the last little bit! 🎉

filipefurtad0:
Hi @mathewdbutton - welcome!

I staged your PR and followed your precise testing notes - all works as expected:

image

Thank you for this improvement :tada:

mathewdbutton:
@filipefurtad0 COOOOOL!

Thanks for confirmation, this is so darn neat.

daniellemoorhead:
Thanks so much for your contribution @mathewdbutton, and please don't be a stranger. We'd love to see you sticking around if and when you have the time to contribute again. The new job will no doubt take up loads of brain capacity in the coming months (as will the zillion Zoom calls you'll have), but if you ever feel like coming back we'd love to have you 😄

mathewdbutton:
@daniellemoorhead I certainly will, I'm already eyeing up my next ticket :). Thanks for having me, it was a great environment to contribute to.

daniellemoorhead:
We'll get you onboard more permanently one day @mathewdbutton, I just know it!

PR #5344, opened 2020-05-02
Move PROFILE debug assets flag to another ENV var

luisramos0:
#### What? Why?
This improves the situation from #5146 where we would need to either have cached assets and classes OR debug assets.
This PR enables us to have both profile and debug assets disabled by default.

#### What should we test?
After this PR the default behaviour is the same as before #5146 

#### Release notes
Changelog Category: Added
Added DEBUG_ASSETS env var to enable asset debugging through a env var.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -50,7 +50,7 @@ Openfoodnetwork::Application.configure do
   # You can remove them by simply running:
   #
   #   $ bundle exec rake assets:clean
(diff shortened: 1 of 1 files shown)


luisramos0:
I am going to merge this now to get it immediately.
@sauloperez approved it here:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5146#discussion_r417409713

PR #5347, opened 2020-05-02
[Mobile] Hide OC selector if shop requires login/customer

### LINKED ISSUE #4529: [mobile ux] Reduce prominence of 'select OC' when Customer can't use it
## Description
When a shop is set to 'private' i.e. only registered customers can place orders, it still shows the order cycle selector. If there is more than one order cycle available, it shows the big red 'choose when you want your order' message. The selector doesn't work as the shop isn't active for this customer. It is unnecessarily confusing!

## Expected Behavior
* If a shop is private and the user on the page is not a Customer
* Definitely - the red 'choose when you want your order' box should NOT appear
* Maybe - the OC selector should just be hidden / not appear

## Actual Behaviour
The OC selector looks like you can use it, and there is a big red message box telling you you should, even though you can't

## Steps to Reproduce
* Create a private shop (using shop preferences) 
* Go to shopfront as a user that is not registered as a customer

## Workaround
Send instructions to customers via email or other channel to try and reduce confusion

## Severity
S4

## Your Environment
Mac, Chrome

luisramos0:
#### What? Why?

Closes #4529

#### What should we test?
Make the shop private and verify the require login or require customer messages appear correctly and, when those messages appear, the OC selector does not appear.

#### Release notes
Changelog Category: Changed
Hide OC selector is users needs to login or if user is not a customer of the shop.

### DIFF
--- a/app/views/shopping_shared/_order_cycles.html.haml
+++ b/app/views/shopping_shared/_order_cycles.html.haml
@@ -1,7 +1,7 @@
 - content_for :injection_data do
   = inject_current_order_cycle
 
-- unless no_open_order_cycles?
+- unless no_open_order_cycles? || require_customer?
   %ordercycle{"ng-controller" => "OrderCycleCtrl", "ng-cloak" => true,
     "ng-class" => "{'requires-selection': !OrderCycle.selected()}"}
     %form.custom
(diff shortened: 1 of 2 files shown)


daniellemoorhead:
**Testing notes**

Tested on iPhone SE/Safari, Mac/Chrome, Windows 10/Chrome, Samsung S8/Chrome.



Allllllelujah, the OC selector has GONE!!

BEFORE
Screen Shot 2020-05-06 at 8 47 17 pm


AFTER
Screen Shot 2020-05-06 at 8 53 58 pm

PR #5350, opened 2020-05-03
Improve order edit page in data inconsistency scenario (follow up from S2 #4186)

luisramos0:
#### What? Why?
This is a follow up from #5253 that improves the situation for #4186
This is handling a data inconsistency problem that is not critical therefore I am not writing specs for this. The next step is to investigate the root cause, but that's an S3 now so we leave it for later.

In this PR I am simply fixing my code in #5253, break is aborting the cycle through units, if one unit is broken (no line item found) we can still iterate through the other units so that the user sees the remaining items in the order instead of a empty order.
See Filipe's testing comments in #5253 for more details.

#### What should we test?
Have a look at the order from #5253
https://staging.openfoodfrance.org/admin/orders/R656741034/edit
No line items before this PR, one line item after this PR.

And/or repeat tests of #5253 and make sure the remaining line item (not the broken one) is now displayed.

#### Release notes
Changelog Category: Changed
Improved the handling of a data inconsistent scenario in the orders edit page.

### DIFF
--- a/app/views/spree/admin/orders/_shipment_manifest.html.haml
+++ b/app/views/spree/admin/orders/_shipment_manifest.html.haml
@@ -1,26 +1,26 @@
 - shipment.manifest.each do |item|
   - line_item = order.find_line_item_by_variant(item.variant)
-  - break if line_item.blank?
 
-  %tr.stock-item{ "data-item-quantity" => "#{item.quantity}" }
-    %td.item-image
-      = mini_image(item.variant)
-    %td.item-name
-      = item.variant.product_and_full_name
-    %td.item-price.align-center
-      = line_item.single_money.to_html
-    %td.item-qty-show.align-center
-      - item.states.each do |state,count|
-        = "#{count} x #{t(state.humanize.downcase, scope: [:spree, :shipment_states], default: [:missing, "none"])}"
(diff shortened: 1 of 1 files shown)


filipefurtad0:
Hi @luisramos0 ,
Line item from the order R656741034 is now showing up under /admin/orders/R365651865/edit, there it is:
image

I repeated the process, creating a similar 2 variant order but deleted a different variant this time by replacing max(id) with the id of the corresponding line item (in this case, spinach). This works as expected in the terminal:
- before deletion, both variants appear, in the terminal and in the app
- deleted line item with id=377, via terminal
- checked that there was one line item left, via terminal
- in the frontend, editing the respective order R365651865 now shows the remaining item (lentils):

image

Interestingly, after the line item deletion, the first time you visit https://staging.openfoodfrance.org/admin/orders/R365651865/edit you see a price discrepancy (in the pic above you see 31.9 Eur, but should be 16 Eur.), the order is displaying the previous total. But I don't think this is an issue for now, as refreshing the page or Pressing Update and Recalculate Fees solves this, and displays "credit owed" to the customer - so it's quite solid! :tada: 

Good to go!

luisramos0:
great testing Filipe, thanks!

PR #5353, opened 2020-05-03
Sort invoices in bulk invoice with completed_at desc order

### LINKED ISSUE #5321: Sort invoices in bulk invoice printing with completed_at desc order
## Description
**- On page:** `/admin/orders?q[s]=completed_at+desc` 
**- I want to be able to do:** 
Today when I select orders, the "print invoices" button does give me only the orders I have selected, however it does not keep the order I see in the order page.

If keeping the order is not possible (because too complex regarding filters etc), having them ordered by `completed_at` date would already be a big improvement!

1. Log in as enterprise manager
2. Go to `/admin/orders?q[s]=completed_at+desc` 
3. Use the filters and sort the page in a specific way
4. Then select the orders and click the print button
5. Invoice within the PDF should have the same order as what the user sees on `/admin/orders?q[s]=completed_at+desc`

pacodelaluna:
#### What? Why?

Closes #5321 

We need to find a way to pass the order of the list to the Print Invoice action. But first on this ticket, we will just focus on ordering with completed_at desc order.

#### What should we test?
On `/admin/orders`, when by clicking on `Print Invoices`, we need to control that the PDF generated is ordering the order following completed_at desc order.

#### Release notes
- Sort orders with completed_at desc order while using Print Invoices action

Changelog Category:  Changed

### DIFF
--- a/app/services/bulk_invoice_service.rb
+++ b/app/services/bulk_invoice_service.rb
@@ -7,9 +7,8 @@ class BulkInvoiceService
 
   def start_pdf_job(order_ids)
     pdf = CombinePDF.new
-    orders = Spree::Order.where(id: order_ids)
 
-    orders.each do |order|
+    orders_from(order_ids).each do |order|
       invoice = renderer.render_to_string(order)
 
       pdf << CombinePDF.parse(invoice)
@@ -28,6 +27,10 @@ class BulkInvoiceService
   end
 
   private
+
+  def orders_from(order_ids)
+    Spree::Order.where(id: order_ids).order("completed_at DESC")
+  end
 
   def new_invoice_id
     Time.zone.now.to_i.to_s
--- a/app/services/bulk_invoice_service.rb
+++ b/app/services/bulk_invoice_service.rb
@@ -8,7 +8,7 @@ class BulkInvoiceService
   def start_pdf_job(order_ids)
     pdf = CombinePDF.new
 
-    orders_from(order_ids).each do |order|
+    orders_from(order_ids).find_each do |order|
       invoice = renderer.render_to_string(order)
 
       pdf << CombinePDF.parse(invoice)
--- a/app/services/bulk_invoice_service.rb
+++ b/app/services/bulk_invoice_service.rb
@@ -8,7 +8,7 @@ class BulkInvoiceService
   def start_pdf_job(order_ids)
     pdf = CombinePDF.new
 
-    orders_from(order_ids).find_each do |order|
+    orders_from(order_ids).each do |order|
       invoice = renderer.render_to_string(order)
 
       pdf << CombinePDF.parse(invoice)
--- a/spec/services/bulk_invoice_service_spec.rb
+++ b/spec/services/bulk_invoice_service_spec.rb
@@ -47,4 +47,11 @@ describe BulkInvoiceService do
       expect(filepath).to eq 'tmp/invoices/1234567.pdf'
     end
   end
+
+  describe "#orders_from" do
+    it "orders with completed desc" do
+      expect(service.send(:orders_from, [1, 2]).to_sql)
+        .to include('ORDER BY completed_at DESC')
+    end
+  end
 end
--- a/spec/services/bulk_invoice_service_spec.rb
+++ b/spec/services/bulk_invoice_service_spec.rb
@@ -50,8 +50,11 @@ describe BulkInvoiceService do
 
   describe "#orders_from" do
     it "orders with completed desc" do
-      expect(service.send(:orders_from, [1, 2]).to_sql)
-        .to include('ORDER BY completed_at DESC')
+      order_old = create(:order_with_distributor, :completed, completed_at: 2.minutes.ago)
+      order_older = create(:order_with_distributor, :completed, completed_at: 3.minutes.ago)
+
+      expect(service.send(:orders_from, [order_older.id, order_old.id]).pluck(:id))
+        .to eq([order_old.id, order_older.id])
     end
   end
 end
--- a/spec/services/bulk_invoice_service_spec.rb
+++ b/spec/services/bulk_invoice_service_spec.rb
@@ -60,11 +60,9 @@ describe BulkInvoiceService do
       order_oldest = create(:order_with_distributor, :completed, completed_at: 4.minutes.ago)
       order_older = create(:order_with_distributor, :completed, completed_at: 3.minutes.ago)
 
-      # This is the creation order provided `find_each` which invalidates
-      # our intended sorting by `completed_at`:
       expect(renderer).to receive(:render_to_string).with(order_old).ordered.and_return("")
-      expect(renderer).to receive(:render_to_string).with(order_oldest).ordered.and_return("")
       expect(renderer).to receive(:render_to_string).with(order_older).ordered.and_return("")
+      expect(renderer).to receive(:render_to_string).with(order_oldest).ordered.and_return("")
 
       order_ids = [order_oldest, order_old, order_older].map(&:id)
       service.start_pdf_job_without_delay(order_ids)
--- a/spec/services/bulk_invoice_service_spec.rb
+++ b/spec/services/bulk_invoice_service_spec.rb
@@ -49,12 +49,25 @@ describe BulkInvoiceService do
   end
 
   describe "#orders_from" do
+    let(:renderer) { InvoiceRenderer.new }
+
+    before do
+      allow(InvoiceRenderer).to receive(:new).and_return(renderer)
+    end
+
     it "orders with completed desc" do
       order_old = create(:order_with_distributor, :completed, completed_at: 2.minutes.ago)
+      order_oldest = create(:order_with_distributor, :completed, completed_at: 4.minutes.ago)
       order_older = create(:order_with_distributor, :completed, completed_at: 3.minutes.ago)
 
-      expect(service.send(:orders_from, [order_older.id, order_old.id]).pluck(:id))
-        .to eq([order_old.id, order_older.id])
+      # This is the creation order provided `find_each` which invalidates
+      # our intended sorting by `completed_at`:
+      expect(renderer).to receive(:render_to_string).with(order_old).ordered.and_return("")
+      expect(renderer).to receive(:render_to_string).with(order_oldest).ordered.and_return("")
+      expect(renderer).to receive(:render_to_string).with(order_older).ordered.and_return("")
+
+      order_ids = [order_oldest, order_old, order_older].map(&:id)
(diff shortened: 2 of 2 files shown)


pacodelaluna:
Right now I am using a fixed order, `completed_at DESC`, as I am not succeeding yet to get this order from the JS controller. Will check it out later.

app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
```diff
@@ -8,7 +8,7 @@ angular.module("admin.orders").controller "bulkInvoiceCtrl", ($scope, $http, $ti
     $scope.poll_wait = 5 # 5 Seconds between each check
     $scope.poll_retries = 80 # Maximum checks before stopping
 
-    $http.post('/admin/orders/invoices', {order_ids: $scope.selected_orders}).success (data) ->
+    $http.post('/admin/orders/invoices', {order_ids: $scope.selected_orders, q: {s: 'completed_at DESC'}}).success (data) ->
```
pacodelaluna, line 11: My big issue is that I don't know how to get easily the list order at this step. Lacking some Angular basics.

pacodelaluna:
The scope of this ticket changed, we could actually directly use the forced order inside the service, but I am thinking that we could let it this way if we want to get the same order than the list later, a part of the logic will be already there.

app/assets/javascripts/admin/orders/controllers/bulk_invoice_controller.js.coffee
Matt-Yorkley: I think I would add this in the pdf service instead (see below).

app/services/bulk_invoice_service.rb
```diff
@@ -5,9 +5,9 @@ def initialize
     @id = new_invoice_id
   end
 
-  def start_pdf_job(order_ids)
+  def start_pdf_job(order_ids, search_params)
     pdf = CombinePDF.new
-    orders = Spree::Order.where(id: order_ids)
+    orders = Spree::Order.where(id: order_ids).ransack(search_params).result
```
Matt-Yorkley, line 10: We can do the ordering here, right? 
```ruby
orders = Spree::Order.where(id: order_ids).order("completed_at DESC")
```

pacodelaluna:
@Matt-Yorkley @mkllnk I have simplify the logic, and add a test, tell me if any issue.

spec/services/bulk_invoice_service_spec.rb
```diff
@@ -47,4 +47,11 @@
       expect(filepath).to eq 'tmp/invoices/1234567.pdf'
     end
   end
+
+  describe "#orders_from" do
+    it "orders with completed desc" do
+      expect(service.send(:orders_from, [1, 2]).to_sql)
```
mkllnk, line 53: This test doesn't execute the query and doesn't verify that it works. It also depends on the internal implementation. It's better to test objects at the interface where they interact with other objects, e.g. the renderer. I would to the following:

* Create three orders with varying `completed_at` dates, unsorted, e.g. 5 hours ago, 3 hours ago, 4 hours ago.
* Inject an InvoiceRenderer mock object.
* Verify that the renderer's `render_to_string` method is called with the newest order first and the oldest order last.

app/services/bulk_invoice_service.rb
```diff
@@ -7,9 +7,8 @@ def initialize
 
   def start_pdf_job(order_ids)
     pdf = CombinePDF.new
-    orders = Spree::Order.where(id: order_ids)
 
-    orders.each do |order|
+    orders_from(order_ids).each do |order|
```
mkllnk, line 11: This will load all orders into the memory at once and then loop through them. It would be better to limit memory consumption by loading orders in batches:
```suggestion
    orders_from(order_ids).find_each do |order|
```
https://apidock.com/rails/ActiveRecord/Batches/find_each

This is out of scope of this PR though. You don't have to do that change but if you do, it should be in a separate commit.

RachL:
@pacodelaluna I'm pinging you just in case you haven't seen Maikel's last comment :)

app/services/bulk_invoice_service.rb
pacodelaluna: @mkllnk Ok, I was not expecting users to check more than 2000 chekboxes but I may miss a usage case :) Batches is a security for sure.

spec/services/bulk_invoice_service_spec.rb
pacodelaluna: Yes, it was a light test, no extra loads required but coverage was less, sure. I will check out your suggestions.

pacodelaluna:
@RachL Yes sorry, i was a bit busy lately, but I will try to update this asap.

app/services/bulk_invoice_service.rb
mkllnk: I'm not sure but I thought there was a checkbox for "check all". And if we don't have it yet, it may come in the future.

pacodelaluna:
@mkllnk I have added 2 patches, one for the batches, and one for the specs. I couldn't make the render_to_string control as the service is rendering directly a PDF file, not sure we want to generate PDF for this test.

spec/services/bulk_invoice_service_spec.rb
```diff
@@ -50,8 +50,11 @@
 
   describe "#orders_from" do
     it "orders with completed desc" do
-      expect(service.send(:orders_from, [1, 2]).to_sql)
-        .to include('ORDER BY completed_at DESC')
+      order_old = create(:order_with_distributor, :completed, completed_at: 2.minutes.ago)
+      order_older = create(:order_with_distributor, :completed, completed_at: 3.minutes.ago)
+
+      expect(service.send(:orders_from, [order_older.id, order_old.id]).pluck(:id))
+        .to eq([order_old.id, order_older.id])
```
mkllnk, line 57: This test passes even if I remove `.order("completed_at DESC")`. The reason is that you create the orders in the same order as you expect them to be retrieved later. The database is very likely to provide rows in that order by default. That's why I suggested to create three orders and create them in an irregular sequence like:

1. old
2. oldest
3. older

To show you what I mean, I changed the test so that it covers the real sequence of rendering calls. And the test failed. I then found that my suggestion to use `find_each` overrides the `ORDER BY` clause. So I led you to code that actually doesn't solve the problem here. That's why proper testing is so important. I will add my commits to your branch for your review. If you are happy with my solution, you can squash all the commits into one.

We still have the potential performance problem if we try to render lots of invoices at once but let's leave it and see if it actually becomes a problem. I suggested to solve it because the solution seemed so simple. But we don't need to add complexity right now.

app/services/bulk_invoice_service.rb
sauloperez: I'd also like to find the orders in batches but `#find_each` clashes with the `order("completed_at DESC")`, as stated in https://apidock.com/rails/v3.2.13/ActiveRecord/Batches/find_in_batches

> It’s not possible to set the order. That is automatically set to ascending on the primary key (“id ASC”) to make the batch ordering work.

spec/services/bulk_invoice_service_spec.rb
sauloperez: Ah, I see you already found it. I give it as approved then @mkllnk

RachL:
@filipefurtad0 @sauloperez I need some advise on how to proceed here.

I've tested several manager account and it is working great. 

However it is not working as super admin:

image

I'm tempted to create a second issue for that because super admin does not use it. But maybe it is introducing a bigger error that I'm not seeing? Thoughts?

filipefurtad0:
Hey @RachL ,

Here's what I've found:

As super-admin:
- I've chosen 2 orders and attempted to bulk-print: it generates the pdf but triggers one 422 error in the console.
- attempting to bulk-print 3 orders resulted in a non-stopping 422 error, even after one cancels the print action (pressing Esc). It just keeps on sending the request...

Also as a hub manager:
- I tried printing large amounts and the pdf is generated, but the error is also triggered here:
image

I'm really not so sure on the impact... Bulk printing orders is only on FR, right?

RachL:
@filipefurtad0 which hub manager did you choose? I've printed around 50 orders as hub manager and it worked. I think something in the data must trigger it.

Bulk Invoice Printing is used everywhere not only in FR. Why I'm saying this can help French hubs is that meanwhile we haven't implemented our invoice compliance stuff, hubs are manually editing the printing invoices. Reports are giving data to do that but ordered by date. Since covid all hubs have now a larger amount of orders per week and they are spending a lot of time to find the order they want to edit in the 200 orders they printed.
They've calculated that get the PDF ordered by date could save them around 2h per week.

But I guess if we are seeing too much errors, we cannot rush this and it needs further work. let's add all the relevant data we can get and move this back in dev?

sigmundpetersen:
Maybe try staging this somewhere else to test if the error is consistent across stagings? Could get us further in understanding if the problem is in data or code.

filipefurtad0:
Before, in FR I used the Hub Tapaskoop. Below on Hub Espargos Place:
image

I see, it's a really useful feature, so maybe it's good to go with it. I have no estimate on how often this will happen and on impact. More thoughts on this? Perhaps fix it and prioritize it as a patch? Or due to the (low?) risk rather merge and roll back if necessary?

RachL:
@filipefurtad0 I propose one last test: let's stage something else in FR and see if we end up with the same problem. If yes, it means the PR is not changing anything and we can merge. If not, better go on the safe side of things?

RachL:
@filipefurtad0  I have the same problem with BIP on staging FR with PR #5482 . So the error is not coming from this PR.
How did super admin performed on staging ES with this PR? Could you edit like a PDF with 100 orders?
Now if the problem is in Master I wonder if this is a release blocker or not. I've tested production, and it is working in super admin :/ So the pb lies either in staging FR data, either in some code on master...

filipefurtad0:
I agree: the PR is not breaking this.

It worked well for 15 orders in ES as superadmin (I've updated my previous comment, plz. see the 2nd pic), which was more that I observed in FR as superadmin, it was breaking already with 3.

In all cases the error 422 can be seen in the console.

Also, I've checked again, and with another PR staged in ES and - as superadmin - printing a 100 orders is successful but a lot of 422s are generated anyway:

image

So, I would say it's good for merging, and an issue should be open for the finding you describe.

RachL:
Ok I'm moving to ready to go. And once the release is ready I will play with several staging server to try to narrow down the problem 👍

sauloperez:
Moving on with this. The problem lays somewhere else as you guys said. This PR's change is absolutely minimal.

PR #5354, opened 2020-05-03
Enhancement/5277 be able to filter orders according to shipping method

### LINKED ISSUE #5277: Enable Bulk Printing of Invoices destined for specific Shipping Method only
## Description

As a manager of a busy hub which has multiple shipping methods (ie different collection points, delivery routes and times) I would like to filter my orders from that weeks order cycle and print the invoices for each shipping method in bulk separately. 
By being able to do so, I can then keep the paper work separate. This will help with putting the correct paper work with a customer's order when packing for each different drop off or delivery point.  It would save a lot of time.

To do this I would like on the Admin-> Orders page, where all the orders are listed, to be able to sort orders by Shipping Method as well as Distributor and Order Cycle.
Once only those orders associated with a given shipping method are present, I can 'Select All' and print invoices in bulk.

1. At the end of an order cycle, visit Admin -> Orders page
2. Filter Orders from Distributor A, Order Cycle AA and with Shipping Method AAA
3. Select All which fit this criteria
4. Select Print Invoices
5. Repeat for Shipping Methods BBB, CCC etc

jeduardo824:
#### What? Why?

Closes #5277 

#### What should we test?
We need to test the Shipping Method filter at `/admin/orders` and check if it is working properly.

#### Release notes

Add Shipping Method filter to Orders page

Changelog Category: Added


jeduardo824:
Hey @luisramos0 could you take a look at that and leave your thoughts? Thank you

jeduardo824:
I'm getting an error on ransack when I try to use the Shipping Method filter, probably because I added `shipping_method_id` to the `ransackable_attributes` array.

I tried a lot to ransack identify the `shipping_method_id` as a "searchable" attribute but no success. 

```
NoMethodError: undefined method `type' for nil:NilClass
from /Users/eduardo/.asdf/installs/ruby/2.3.7/lib/ruby/gems/2.3.0/gems/ransack-0.7.2/lib/ransack/adapters/active_record/context.rb:25:in `type_for'
```

app/services/search_orders.rb
```diff
@@ -24,7 +24,12 @@ def pagination_data
   attr_reader :params, :current_user
 
   def fetch_orders
-    @search = ::Permissions::Order.new(current_user).editable_orders.ransack(params[:q])
+    @search = ::Permissions::Order.new(current_user)
+                                  .editable_orders
+                                  .joins(shipments: { shipping_rates: :shipping_method })
+                                  .where(spree_shipping_rates: { selected: true })
+                                  .select('spree_orders.*, spree_shipping_rates.shipping_method_id as shipping_method_id')
+                                  .ransack(params[:q])
```
luisramos0, line 32: Nice one @jeduardo824 almost there!

I am not sure how to do this, @Matt-Yorkley can you advise here?

one option here is to bypass ransack for this attribute. remove it from params[:q] and send it as a normal param and them use that param to add an extra filter to the query like:
```ruby
.joins(shipments: :shipping_rates })
.where("spree_shipping_rates.selected == true and spree_shipping_rates.shipping_method_id = #{params[shipping_method_id]}")
```
btw, we dont need to join shipping_method, shipping_method_id in shipping_rates is enough to filter

jeduardo824: I thought about doing that. My concern is when the user doesn't use the Shipping Method filter. I could create a scope to return based on the shipping_method_id and use that only when params[:shipping_method_id] is not null.

luisramos0: yes, you can build the query incrementally, I wouldnt create a scope, I'd just do it here:
```
 def fetch_orders
    @search = ::Permissions::Order.new(current_user)
                                  .editable_orders
    if params[shipping_method_id]
         @search = @search.joins(shipments: :shipping_rates })
                            .where("spree_shipping_rates.selected == true and spree_shipping_rates.shipping_method_id = #{params[shipping_method_id]}")
     end

      @search = @search.ransack(params[:q])

```

jeduardo824: Good idea. I will try and update my PR.

jeduardo824: It worked!

luisramos0: awesome! some errors in the build, can you have a look?

jeduardo824: Sure

jeduardo824:
Hey guys, I did not forget about this. Just a crazy week/weekend. I will do my best to fix the CI this week.

luisramos0:
:+1: thanks for letting us know!

jeduardo824:
What do you think? @luisramos0

app/services/search_orders.rb
```diff
@@ -24,13 +24,30 @@ def pagination_data
...
+  end
+
+  def shipping_method_query
+    base_query
+      .joins(shipments: :shipping_rates)
+      .where(spree_shipping_rates: {
+               selected: true,
+               shipping_method_id: params[:shipping_method_id]
+             })
+  end
```
luisramos0, line 49: how about just this method?
```ruby
  def search_query
     base_query = ::Permissions::Order.new(current_user).editable_orders
     return base_query unless params[:shipping_method_id]

     base_query
        .joins(shipments: :shipping_rates)
        .where(spree_shipping_rates: {
                selected: true,
               shipping_method_id: params[:shipping_method_id]
             })
  end
```

PR #5359, opened 2020-05-04
Add NavigationCheck to order cycles edit form

### LINKED ISSUE #5108: Autosave missing when moving from Incoming to Outgoing Products setting up or editing Order Cycles
## Description

When creating a hub order cycle you can move from screen two (incoming products) to screen three (outgoing products) by clicking on the tab at the top rather than using the buttons on the bottom right of the screen. 
When editing a hub order cycle you can move between screens one, two and three by clicking on the tab at the top rather than using the buttons on the bottom right of the screen. 
Navigating by the top menu bar means that any changes made to an order cycle before moving page are lost.

## Expected Behavior

Navigating by top menu bar of the hub order cycle should automatically save changes made or the hyperlinks should be removed from the top menu bar to force users to select 'save' or 'save and next' to navigate.

## Steps to Reproduce

1. Make change to incoming products of an order cycle.
2. Move to outgoing products section by clicking on the top menu '3. outgoing products'
3. Move back to 2. incoming products and changes made in step 1 are lost.
4.

## Workaround

image

image

## Severity

bug-s4

## Your Environment

* Version used:
* Browser name and version:

mbudm:
#### What? Why?

Closes #5108

This adds a similar navigation check that is used on the enterprises form to the order_cycles edit form. This check pops up a javascript alert for the user when any attempt to change the window.location (or page refresh) occurs.

#### What should we test?
This can be manually tested by:
- Editing an order cycle for a hub (an enterprise that 'sells any').
- Make any change that triggers the activation of the save buttons, and the disabling of the next button
- Click any button that navigates to a new page (eg a nav item or the wizard links)
- An alert will display asking you to confirm that you want to leave the 'site' (as chrome calls it)
Screen Shot 2020-05-04 at 10 58 14 pm

I've also attempted to modify the spec for this form, `spec/features/admin/order_cycles_spec.rb` but am getting odd behavior that I haven't figured out yet (hence why this PR is WIP). I'm getting a selenium timeout on the ["editing an order cycle"](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/spec/features/admin/order_cycles_spec.rb#L428) scenario which makes modifying the test to handle these alerts difficult. I'm curious to see how CI fares.

#### Release notes
Alert administrators when an order cycle form is unsaved

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
+++ b/app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
@@ -1,5 +1,5 @@
 angular.module('admin.orderCycles')
-  .controller 'AdminEditOrderCycleCtrl', ($scope, $controller, $filter, $location, $window, OrderCycle, Enterprise, EnterpriseFee, StatusMessage, Schedules, RequestMonitor, ocInstance) ->
+  .controller 'AdminEditOrderCycleCtrl', ($scope, $controller, $filter, $location, $window, OrderCycle, Enterprise, EnterpriseFee, StatusMessage, Schedules, RequestMonitor, NavigationCheck, ocInstance) ->
     $controller('AdminOrderCycleBasicCtrl', {$scope: $scope, ocInstance: ocInstance})
 
     order_cycle_id = $location.absUrl().match(/\/admin\/order_cycles\/(\d+)/)[1]
@@ -18,5 +18,12 @@ angular.module('admin.orderCycles')
 
     $scope.submit = ($event, destination) ->
       $event.preventDefault()
+      NavigationCheck.clear()
       StatusMessage.display 'progress', t('js.saving')
       OrderCycle.update(destination, $scope.order_cycle_form)
+
+    warnAboutUnsavedChanges = ->
+      if $scope.order_cycle_form?.$dirty
+        t('admin.unsaved_confirm_leave')
+
+    NavigationCheck.register(warnAboutUnsavedChanges)
--- a/spec/features/admin/order_cycles_complex_nav_check_spec.rb
+++ b/spec/features/admin/order_cycles_complex_nav_check_spec.rb
@@ -1,0 +1,37 @@
+require 'spec_helper'
+
+feature '
+    As an administrator
+    I want to be alerted when I navigate away from a dirty order cycle form
+', js: true do
+  include AuthenticationWorkflow
+
+  scenario "alert when navigating away from dirty form" do
+    # Given a 'complex' order cycle form
+    oc = create(:order_cycle)
+
+    # When I edit the form
+    quick_login_as_admin
+    visit edit_admin_order_cycle_path(oc)
+
+    wait_for_edit_form_to_load_order_cycle(oc)
+   
+    expect(page).to have_selector '.wizard-progress .current a', text: '1. GENERAL SETTINGS'
+    expect(page.find('#order_cycle_name').value).to eq(oc.name)
+    expect(page).to have_button("Save", disabled: true)
+    fill_in 'order_cycle_name', with: 'Plums & Avos'
+
+    # Then the form is dirty and an alert warns about navigating away from the form
+    expect(page).to have_button("Save", disabled: false)
+    expect(page).to have_selector '.wizard-progress a', text: '2. INCOMING PRODUCTS'
+    accept_alert do
+      click_link '2. Incoming Products'
+    end
+  end
+
+  private
+
+  def wait_for_edit_form_to_load_order_cycle(order_cycle)
+    expect(page).to have_field "order_cycle_name", with: order_cycle.name
+  end
+end
--- a/app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
+++ b/app/assets/javascripts/admin/enterprises/controllers/enterprise_controller.js.coffee
@@ -29,7 +29,7 @@ angular.module("admin.enterprises")
     # from a directive "nav-check" in the page - if we pass it here it will be called in the test suite,
     # and on all new uses of this contoller, and we might not want that.
     enterpriseNavCallback = ->
-      if $scope.enterprise_form != undefined && $scope.enterprise_form.$dirty
+      if $scope.enterprise_form?.$dirty
         t('admin.unsaved_confirm_leave')
 
     # Register the NavigationCheck callback


app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
```diff
@@ -18,5 +19,14 @@ angular.module('admin.orderCycles')
 
     $scope.submit = ($event, destination) ->
       $event.preventDefault()
+      $scope.navClear()
       StatusMessage.display 'progress', t('js.saving')
       OrderCycle.update(destination, $scope.order_cycle_form)
+
+    # Provide a callback for generating warning messages displayed before leaving the page. 
+    orderCycleNavCallback = ->
+      if $scope.order_cycle_form != undefined && $scope.order_cycle_form.$dirty
```
luisramos0, line 28: coffee script has the safe navigation operator like ruby but it's with a question mark.
this becomes
f $scope.order_cycle_form?.$dirty

app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
```diff
@@ -18,5 +19,14 @@ angular.module('admin.orderCycles')
...
+        t('admin.unsaved_confirm_leave')
+
+    # Register the NavigationCheck callback
+    NavigationCheck.register(orderCycleNavCallback)
```
luisramos0, line 32: missing empty line at the end?

app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
luisramos0: can be changed on enterpriseCtrl as well

mbudm: Ok but like `.ts` - is that effectively saying, I'm certain that this exists? If so are we certain of that?

mbudm:
> Timeouts on specs could be a local setup problem.

> fyi, you can remove the headless keyword from spec_helper.rb and see chrome running the spec.
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Testing-and-Rspec-Tips#debug-tests

Yeah I tried removing headless, it actually was more troublesome. For example this fun failure where selenium couldn't see the button as it was obscured by the status message on the save bar.:
Screen Shot 2020-05-04 at 9 36 55 pm

I had a little chat with @mkllnk and a suggestion arose that I'm now working on - split up the order_cycle_spec into smaller specs for simple, complex and list scenarios. In theory this should make them less error prone locally.

mkllnk:
> I had a little chat with @mkllnk and a suggestion arose that I'm now working on - split up the order_cycle_spec into smaller specs for simple, complex and list scenarios. In theory this should make them less error prone locally.

I don't think that splitting the specs into multiple files will make the specs fail less. But it will help you (and all future devs) to find the best place to insert your spec for this pull request.

You could experiment with setting a large screen size when not useing the headless option.

app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
luisramos0: Its saying "I am not certain that this property exists, so return false if it doesnt exist or navigate further (to $dirty) if it does"

luisramos0:
Hello @mbudm usually we tend to troubleshoot local setup problems on slack, more people will see and help there.
Can you share you problem there?
In the meantime, have you tried running a specific spec and not the full file? Does that help?

mbudm:
Thanks @luisramos0 and @mkllnk I will move the local rspec discussion to slack.

In the meantime I thought the best way forward is to add a simple test that covers this change. I've pushed as a new squashed commit so the test is in the same commit as the change.

app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
mbudm: Ace

mbudm: Fixed with 7a41494

spec/features/admin/order_cycles_complex_nav_check_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+', js: true do
+  include AuthenticationWorkflow
+
+  scenario "alert when navigating away from dirty form" do
+    # Given an order cycle with all the settings
+    oc = create(:order_cycle)
+    oc.suppliers.first.update_attribute :name, 'AAA'
+    oc.suppliers.last.update_attribute :name, 'ZZZ'
+    oc.distributors.first.update_attribute :name, 'AAAA'
+    oc.distributors.last.update_attribute :name, 'ZZZZ'
```
luisramos0, line 15: I wonder why do we need these updates?
FYI you could use simple_order_cycle factory like create(:simple_order_cycle) if that would reduce number of lines here.

mbudm: Yeah good point, my copy pasta is showing. Removed in 223521d

From what I've learnt I think it needs `create(:order_cycle)` over `create(:simple_order_cycle)` to get the complex form variant though.

app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
```diff
@@ -18,5 +19,14 @@ angular.module('admin.orderCycles')
 
     $scope.submit = ($event, destination) ->
       $event.preventDefault()
+      $scope.navClear()
```
mkllnk, line 22: Did you try just calling `NavigationCheck.clear()` here? I'm not sure if you need to expose that function to the scope.

app/assets/javascripts/admin/order_cycles/controllers/edit.js.coffee
```diff
@@ -18,5 +19,14 @@ angular.module('admin.orderCycles')
...
       StatusMessage.display 'progress', t('js.saving')
       OrderCycle.update(destination, $scope.order_cycle_form)
+
+    # Provide a callback for generating warning messages displayed before leaving the page. 
+    orderCycleNavCallback = ->
+      if $scope.order_cycle_form?.$dirty
+        t('admin.unsaved_confirm_leave')
+
+    # Register the NavigationCheck callback
+    NavigationCheck.register(orderCycleNavCallback)
```
mkllnk, line 32: The name `orderCycleNavCallback` doesn't tell us anything about what it does and you need to explain it in a comment. Maybe a name like `warn_about_unsaved_changes` would be better?

spec/features/admin/order_cycles_complex_nav_check_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  scenario "alert when navigating away from dirty form" do
+    # Given a 'complex' order cycle form
+    oc = create(:order_cycle)
+
+    # When I edit the form
+    quick_login_as_admin
+    visit edit_admin_order_cycle_path(oc)
+
+    wait_for_edit_form_to_load_order_cycle(oc)
```
mkllnk, line 17: Nice spec! I like the added meaning here. :+1:

mbudm:
> It's good to only add commits during review but now we can clean it up.
> We usually don't cherry-pick from master unless necessary, of course. Did the ddtrace update solve something for you?

Okidoke. I rebased once already and included the ddtrace commit, Will do it again and ignore that as it's already in master

luisramos0:
awesome, Test Ready, a tester will pick it up now soon.

filipefurtad0:
Hi @mbudm,

I checked the fix this PR is introducing: works perfectly. Thank you for this improvement, and kudos for your effort with the specl :tada: 

Ready to go!

luisramos0:
nice, well done @mbudm :tada:

Issue #5239, opened 2020-04-16
Stuck on cart, if item becomes out of stock during shopping journey


filipefurtad0:
## Description

After a customer visits the shopfront and adds items to cart, it can happen that the stock is reduced, making those items unavailable, before the customer checks out. If this happens, and the customer attempts to check out, he is redirected to the cart page, where the red banner displays with "An item in your cart has become unavailable." Also, in front of the respective item, a red text indicates "Insufficient stock available, only xxx remaining" - pic below.

Now, at this point, the user is "stuck" in the /cart page. Even if the number of items is reduced (either introducing a number or using the small down-arrow) till the "insufficient stock" warning disappears, it is not possible, as the buttons below in [brackets], only leave you on the current cart page:
- to proceed to checkout [ Checkout > ] 
- go back to the shopfront. [ < Continue Shopping ]
- Going "back" in the browser does not work either.

## Expected Behavior

It should be possible to leave the cart page, either by:
- correcting the number of items and pressing  [ Checkout > ]
- go back to the shopfront. [ < Continue Shopping ], for example, to choose from another product

## Actual Behaviour
Customer gets stuck on cart page.

## Steps to Reproduce

C - Customer
A - Admin

C - add X items to the cart.
C - Proceed to Checkout page
A - In Products, reduce the amount of available items, to a value below X
C - Press [Place Order Now]
-> boom!

## Animated Gif/Screenshot

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/359027ee-a3d0-476f-9e4e-1ef26b590da8)

## Workaround

Here, the only way out, is pressing the trash-bin and removing the items from the cart.

## Severity

bug-s3: a feature is broken but there is a workaround
(however, perhaps this affects few users? S3 and a half? :-)

## Your Environment

* Version used: v2.9.2
* Browser name and version: Firefox 75
* Operating System and version (desktop or mobile): Ubuntu 18.04

RachL:
This is something I've hit in previous test and never manage to properly reproduce, thanks @filipefurtad0 !
However the probability for customers to hit that one is higher than ever. All our hubs have little stock and all customers are ordering during the first hours of the order cycle.

I would make this one an s2. Thougths? @lin-d-hop @luisramos0 ?

luisramos0:
I can replicate but I am not stuck in the cart page. It's just incredible bad UX :-(
Have you tried to reduce quantity and press Update?

[...]

PR #5361, opened 2020-05-04
Cart with soft-deletion

### LINKED ISSUE #5358: Cart Problems in Production
We've now had a couple of downtime incidents caused by a cart bug, on multiple production servers.

If the `CartController#populate` endpoint (add-to-cart) returns an error code, the page will keep trying to post data to the cart endpoint again and again and again. Twice now we've had the add-to-cart endpoint return a `412` (we don't know why yet), and then it subsequently kept re-posting the data over and over, and this can cause serious problems.

More details in a thread from Friday 01 May in the (private) channel **#devops-notifications**

#### Proposed fix(es):

- [ ]  Investigate and fix whatever caused the initial `412`. #5361 
- [ ] Fix the problem of the add-to-cart code continually spamming the endpoint with lots of requests if an error is returned. Maybe just show an error to the user, and _not_ continually re-POST the data. #5371

Matt-Yorkley:
#### What? Why?

Closes #5358 

Adds various missing specs for variant soft-deletion issues and adjusts `CartService` logic so that if a variant is soft-deleted whilst it's in an active cart or is added to a cart, the user gets useful feedback (and no fatal error problems), the item is removed and acts as if it has gone out of stock.

#### What should we test?

Load a shop page, then delete a variant that's already shown on the page, then add that variant to the cart. It should show an out-of-stock modal and remove it from the cart.

#### Release notes

Fixed issues around adding deleted variants to cart.

### DIFF
--- a/app/models/spree/stock/quantifier.rb
+++ b/app/models/spree/stock/quantifier.rb
@@ -11,6 +11,8 @@ module Spree
       end
 
       def total_on_hand
+        return 0 if @variant.deleted?
+
         stock_items.sum(&:count_on_hand)
       end
 
--- a/app/models/spree/stock/quantifier.rb
+++ b/app/models/spree/stock/quantifier.rb
@@ -11,6 +11,8 @@ module Spree
       end
 
       def total_on_hand
+        # Associated stock_items no longer exist if the variant has been soft-deleted. A variant
+        # may still be in an active cart after it's deleted, so this will mark it as out of stock.
         return 0 if @variant.deleted?
 
         stock_items.sum(&:count_on_hand)
--- a/app/services/cart_service.rb
+++ b/app/services/cart_service.rb
@@ -29,11 +29,15 @@ class CartService
 
     variants_data.each do |variant_data|
       loaded_variant = loaded_variants[variant_data[:variant_id].to_i]
+
+      if loaded_variant.deleted?
+        remove_deleted_variant(loaded_variant)
+        next
+      end
+
       next unless varies_from_cart(variant_data, loaded_variant)
 
-      attempt_cart_add(
-        loaded_variant, variant_data[:quantity], variant_data[:max_quantity]
-      )
+      attempt_cart_add(loaded_variant, variant_data[:quantity], variant_data[:max_quantity])
     end
   end
 
@@ -41,10 +45,14 @@ class CartService
     @indexed_variants ||= begin
       variant_ids_in_data = variants_data.map{ |v| v[:variant_id] }
 
-      Spree::Variant.where(id: variant_ids_in_data).
+      Spree::Variant.with_deleted.where(id: variant_ids_in_data).
         includes(:default_price, :stock_items, :product).
         index_by(&:id)
     end
+  end
+
+  def remove_deleted_variant(variant)
+    line_item_for_variant(variant).andand.destroy
   end
 
   def attempt_cart_add(variant, quantity, max_quantity = nil)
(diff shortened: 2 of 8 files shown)


app/models/spree/stock/quantifier.rb
```diff
@@ -11,6 +11,8 @@ def initialize(variant)
       end
 
       def total_on_hand
+        return 0 if @variant.deleted?
```
luisramos0, line 14: maybe useful to have a comment explaining why we need this line?
I imagine something like "if variant is deleted and is still in the cart, the zero total_on_hand will make it out of stock and force it's removal from the cart" ?

Also, and probably more important, I think you need a similar "if super.deleted?" on ScopeVariantToHub#def total_on_hand ?

Matt-Yorkley: Yeah I was looking at a couple of cases like this. With some minor adjustments we can have overridden stock still function on one enterprise if a variant us deleted on another enterprise. Like if a distributor has 200 stock items for an override, but the variant is soft-deleted by their supplier (maybe as a complete surprise to the distributor) they still have that stock (inventory), and they still need to sell it, right? 

It raises some interesting questions though...

Matt-Yorkley: Basically if the override explicitly has stock, ScopeVariantToHub#total_on_hand will have valid stock items that could theoretically be used.

luisramos0: that makes sense, but I dont think it will work on the shop/cart/checkout right? I think we should keep things very very simple. Maybe we should create an issue to verify that scenario and fix it?

Matt-Yorkley: I had a quick play with it and I think it will totally work in shop/cart/checkout with some very small adjustments. In another issue though :+1:

I'll add the explanatory comment here for now.

Matt-Yorkley:
Rebased :+1:

app/models/spree/stock/quantifier.rb
luisramos0: nice comment!

app/services/cart_service.rb
```diff
@@ -29,24 +29,32 @@ def attempt_cart_add_variants(variants_data)
...
 
-      Spree::Variant.where(id: variant_ids_in_data).
+      Spree::Variant.with_deleted.where(id: variant_ids_in_data).
         includes(:default_price, :stock_items, :product).
         index_by(&:id)
     end
   end
 
+  def remove_deleted_variant(variant)
+    line_item_for_variant(variant).andand.destroy
```
sauloperez, line 55: I think in the long term we should soft-delete the line item upon soft-deletion of the variant, ala ON DELETE CASCADE only that it'll be an `UPDATE` statement in this case.

filipefurtad0:
Awesome @Matt-Yorkley ,

Checked two scenarios here:

User: Loads /shopfront (variant x is visible)
Admin: Deletes variant x
User: Adds variant x to the cart (pressing the up arrow a couple of times)
-> the "Reduced stock available" modal kicks in :tada: and the variant x becomes greyed-out.

I think this is the main aim of this PR, this is ready to go.

The other scenario I checked is:

User: Loads /shopfront (variant x is visible)
User: Adds variant x to the cart (pressing the up arrow a couple of times)
Admin: Deletes variant x
-> In this case, the modal is not shown.
User: pressing Cart or Edit your cart will lead to the situation described on issue #5239 

While this could be improved it surely introduces a huge improvement. Should I open a separate issue for the second test-case?

PR #5365, opened 2020-05-05
Split order_cycle_spec into smaller specs grouped by theme

mbudm:
#### What? Why?

Supports #5108 

The order cycle spec contains 25 scenarios and is > 1000 LOC. This change splits it into more manageable chunks so that it is quicker to run locally and easier to debug when one scenario fails.

The groupings are, complex, simple and list. 

#### What should we test?
Nothing to test except running the new smaller tests

#### Release notes
Split up the order cycle spec to make it more comprehensible

Changelog Category: Changed


mbudm:
Further splitting of the complex spec has now gotten me to the point where I can see a consistent error, which is somewhat progress. I'e I get this error:

```sh
Failure/Error: check "order_cycle_incoming_exchange_0_select_all_variants"
     
     Capybara::ElementNotFound:
       Unable to find visible checkbox "order_cycle_incoming_exchange_0_select_all_variants" that is not disabled
     # ./spec/features/admin/order_cycles/complex_editing_multiple_product_pages_spec.rb:37:in `block (3 levels) in <top (required)>'
```

When running either of:
```sh
bundle exec rspec spec/features/admin/order_cycles/complex_editing_multiple_product_pages_spec.rb 
```
or 
```sh
bundle exec rspec spec/features/admin/order_cycles/*
```

Whereas previously with many scenarios in one spec file, the errors were different, pointing to a cascading error effect, where one error or side effect in a selenium instance(?) triggers another error which is actually not the problem.

This is now more manageable to test in the Chrome test runner, which I'll do next.

mbudm:
I'm going to pause this now and have a think. Isolating problems by creating smaller specs was helpful. This is to do with the practicalities of running long specs and being able to read output (eg we have a lot of verbose warn ing messages that we will be able to remove when we upgrade to rails 4.2).

The issue here seems to be primarily the sticky footer. Sometimes Capybara/selenium cannot find elements that are hidden by the footer element.
I tried this capability setting: however it does not apply to Chrome: https://stackoverflow.com/questions/36439701/selenium-webdriver-how-do-i-set-capabilities-elementscrollbehavior-to-1-for-a#37171021

Other options are:
- hide the footer when attempting to interact with an element that _may_ be obscured. Problem is what is a reliable way of knowing an element is obscured?
- manually scroll the page so that the target element is the centre of the window
- add options to find the element even if it is not visible

mbudm:
There is a `scroll_to` method in [Capybara 3.13](https://github.com/teamcapybara/capybara/blob/d7e9efbbcd71c6e2abfebd86bc8b51ec8408597e/History.md#version-3130), so I suggest we use a placeholder script to do the same and then replace when we upgrade.

spec/features/admin/order_cycles/complex_editing_multiple_product_pages_spec.rb
```diff
@@ -34,7 +34,10 @@
     end
 
     scenario "select all products" do
-      check "order_cycle_incoming_exchange_0_select_all_variants"
+      # replace with scroll_to method when upgrading to Capybara >= 3.13.0
+      cb_id = "order_cycle_incoming_exchange_0_select_all_variants"
+      page.execute_script("document.getElementById('#{cb_id}').scrollIntoView()")
```
mkllnk, line 39: Is this possible?

I would have thought that the "view" is the browser window and that it can still be hidden by an overlay. How did you test this?

spec/features/admin/order_cycles/complex_editing_multiple_product_pages_spec.rb
```diff
@@ -34,7 +34,10 @@
     end
 
     scenario "select all products" do
-      check "order_cycle_incoming_exchange_0_select_all_variants"
+      # replace with scroll_to method when upgrading to Capybara >= 3.13.0
+      cb_id = "order_cycle_incoming_exchange_0_select_all_variants"
```
mkllnk, line 38: `cb_id` is a bit cryptic. I would call it `checkbox_id`.

spec/features/admin/order_cycles/complex_editing_multiple_product_pages_spec.rb
mbudm: Hi @mkllnk - I checked it visually by removing 'headless' in `spec_helper.rb` and the behavior of scrollIntoView is [as specified](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView):

> "the top of the element will be aligned to the top of the visible area of the scrollable ancestor"

spec/features/admin/order_cycles/complex_editing_multiple_product_pages_spec.rb
mbudm: Good point - changed to be more informative

luisramos0:
Hi @mbudm 
There's some mismatch between your GH user and your local git user:
image

I cant remember the fix to this but it's an easy local git config I think.

On another note, I recommend you install zenhub browser extension so you can see the board and also you can move PRs around: usually from In Dev to Code Review when the PR is ready for review.

mkllnk:
We can also add to the [mailmap](https://git-scm.com/docs/git-check-mailmap) file to help Git make the connection.
https://github.com/openfoodfoundation/openfoodnetwork/blob/bb8401a0fc56db36f5d426d7a666f834ef3a27a6/.mailmap#L1-L2

spec/features/admin/order_cycles/complex_editing_multiple_product_pages_spec.rb
mkllnk: Ah, it's going to the top like clicking on an anchor. That makes it most likely visible, yes. :smile:

mkllnk:
Since this change affects specs only, we can merge it without manual testing.

PR #5370, opened 2020-05-05
Don't render twice

### LINKED ISSUE #5369: Broken shopfront in an edge case where OC is closed but products are fetched from database
## Error in OpenFoodNetwork UK

**AbstractController::DoubleRenderError** in **api/order_cycles#products**
Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5eb156ca1a8b83001807eff7?event_id=5eb156ca005b3525783d0000&i=gh&m=ci)

## Stacktrace

    app/controllers/api/base_controller.rb:74 - error_during_processing

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5eb156ca1a8b83001807eff7?event_id=5eb156ca005b3525783d0000&i=gh&m=ci)

Matt-Yorkley:
#### What? Why?

Closes #5369 

Not much more to say here that isn't already in the title.

#### What should we test?

I'm not sure we can test this one easily. The products list endpoint that displays products should not return an error when the shopfront is closed, but actually viewing a closed shopfront is quite difficult, and the product list shouldn't actually load successfully in this case anyway...

Edit: I've had another look, and I think we can do this to verify it's fixed:
- go to an open shop with an open order cycle
- look in the browser dev console in the network tab and find the request for the product list
- copy the full URL of the request, it'll have `/api/order_cycles` and `/products` in it, and `?distributor=` and some other bits
- close the order cycle
- paste that URL into the browser
- you shouldn't see the snail
- you shouldn't see much of anything else either, it'll be a blank json response

#### Release notes

Fixed double-rendering of products in unusual closed shopfront edge case

### DIFF
--- a/app/controllers/api/order_cycles_controller.rb
+++ b/app/controllers/api/order_cycles_controller.rb
@@ -7,7 +7,7 @@ module Api
     skip_before_filter :authenticate_user, :ensure_api_key, only: [:taxons, :properties]
 
     def products
-      render_no_products unless order_cycle.open?
+      return render_no_products unless order_cycle.open?
 
       products = ProductsRenderer.new(
         distributor,


filipefurtad0:
Hi @Matt-Yorkley ,

I found two requests with the description you indicate when the shopfront page is loaded:
1) https://staging.openfoodfrance.org/api/order_cycles/54/taxons?distributor=38
2) https://staging.openfoodfrance.org/api/order_cycles/54/properties?distributor=38

Both these output the same, before and after the closing the OC:

1) the JSON has content:
[{"id":2,"name":"Vegetables","permalink":"products/vegetables","icon":"/assets/default_taxon.png","pretty_name":"Products -\u003E Vegetables","position":6,"parent_id":1,"taxonomy_id":1}]

in 2) it's an empty JSON []

Refreshing the shopfront leads the user to the homepage, as usual.

No snail, no bugsnag.
All good :+1:

PR #5371, opened 2020-05-05
Cart add error handling

Matt-Yorkley:
#### What? Why?

Second part of #5358 

Now that the other part of the issue is fixed in #5361 we shouldn't see this case arise at all. In the rare event that an edge case pops up we will now:

- at least show the user some feedback letting them know something has gone wrong, instead of the cart breaking and nothing happening
- _not_ keep continuously re-posting the data to the endpoint if the endpoint is fatally failing to process it anyway (which kills the server)

#### What should we test?

If #5361 hasn't been merged yet, we can probably reproduce a cart error with the test steps from that PR:

> Load a shop page, then delete a variant that's already shown on the page, then add that variant to the cart.

I can't figure out how to actually get the cart to throw an error any other way.

#### Release notes

Improved fatal error handling on `CartController#populate` endpoint

### DIFF
--- a/app/assets/javascripts/darkswarm/services/cart.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/cart.js.coffee
@@ -50,7 +50,6 @@ Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $roo
         @popQueue() if @update_enqueued
 
       .error (response, status)=>
-        @scheduleRetry(status)
         @update_running = false
 
     compareAndNotifyStockLevels: (stockLevels) =>
@@ -86,13 +85,6 @@ Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $roo
           quantity: li.quantity
           max_quantity: li.max_quantity
       {variants: variants}
-
-    scheduleRetry: (status) =>
-      console.log "Error updating cart: #{status}. Retrying in 3 seconds..."
-      $timeout =>
-        console.log "Retrying cart update"
-        @orderChanged()
-      , 3000
 
     saved: =>
       @dirty = false
--- a/app/assets/javascripts/darkswarm/services/cart.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/cart.js.coffee
@@ -1,4 +1,4 @@
-Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $rootScope, $resource, localStorageService) ->
+Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $rootScope, $resource, localStorageService, RailsFlashLoader) ->
   # Handles syncing of current cart/order state to server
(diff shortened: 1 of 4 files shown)


Matt-Yorkley:
A fatal error from the cart endpoint now looks like this:

Screenshot from 2020-05-05 16-22-58

app/assets/javascripts/darkswarm/services/cart.js.coffee
```diff
@@ -50,7 +50,7 @@ Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $roo
         @popQueue() if @update_enqueued
 
       .error (response, status)=>
-        @scheduleRetry(status)
+        RailsFlashLoader.loadFlash({error: t('js.cart.add_to_cart_failed')})
```
luisramos0, line 53: I think we shoul keep the retry, the request can fail and we can retry. I'd keep the retry with a number, like 3 or 5.

Matt-Yorkley: If we post a set of data to the cart and it definitely can't be processed, does it make sense to post exactly the same data again?

Here the user can still try to add or remove items after seeing the message, and it will send the new (different) data to the cart.

sauloperez: agree with @Matt-Yorkley . It's not a unreliable integration where retrying would make sense, such as connecting to a third-party API, etc. IMO

app/assets/javascripts/darkswarm/services/cart.js.coffee
```diff
@@ -87,13 +87,6 @@ Darkswarm.factory 'Cart', (CurrentOrder, Variants, $timeout, $http, $modal, $roo
           max_quantity: li.max_quantity
       {variants: variants}
 
-    scheduleRetry: (status) =>
```
sauloperez, line 90: :see_no_evil: so dangerously naive...

filipefurtad0:
Hey @Matt-Yorkley ,

Repeating the first test from #5361 the error message "There was a problem adding this product to the cart. Perhaps it has become unavailable or the shop is closing." Attempting to reach the cart leads to a page reload, with that given variant not to be found any more. This was done previous to the merge of this PR #5361, so it correctly catching this edge case :+1: 

It does not catch the second tested scenario in #5361, which leads to the stuck-in-the-cart situation. As before, surely the best would be to prevent reaching this state. Thoughts?

Matt-Yorkley:
> It does not catch the second tested scenario in #5361 , which leads to the stuck-in-the-cart situation.

Yeah, it looks like we still have some issues around the cart page. It's next on my list :+1:

app/assets/javascripts/darkswarm/services/cart.js.coffee
luisramos0: ok, if you guys agree, lets go.
I was thinking about the case: server is not available and cart becomes inconsistent with the order on the backend.

PR #5373, opened 2020-05-05
Dont refresh shipping rates when editing an order in the backoffice

### LINKED ISSUE #5367: Desactivating a shipping method with fees leads to wrong delivery method and payment status in previous orders
## Description

This is something that I have found while following hub 922 in French prod.

This hub has 2 delivery method: pickup and delivery by bike. Delivery by bike has a fixed fee of 6 euros.

A typical week for them goes like this:

- On Tuesday they open their OC with meat products. Everyone has a choice between home delivery and pick up because they have only a few customers buying meat

- On Friday they open their OC with vegetables. A lot more customers are involved, so they open in 2 stages :
1. From 7 am to 1 PM their coop members who can choose between home delivery and pick-up thanks to a special OC tag
2. At 1PM they remove the tag on the OC: everyone gets to order. However only the first 40 people will get to be able to select home delivery. So they check delivery report. Once they hit 40 home deliveries they de-activate the home delivery shipping method.

When the week start again, they start this cycle again.

However, when they deactivate the home delivery shipping method, all the previous orders display then the only shipping method available : pick up, with no fees. So all these orders are now showing a credit due to the customer of 6 euros, which is wrong: these customer where delivered by bike, so should NOT be refunded the bike fees.

All reports are wrong too.

## Steps to Reproduce
 
Orders on French prod that are incorrect: # R088866556, # R008087351,  # R266862838  ... but maybe many that I have troubles to find  :sob:

To reproduce on staging:
1. Create 2 shipping method, one with fees, the other without
2. Order something with the shipping method with fees
3. De-activate that shipping method
4. Check the order, see that it is now showing the shipping method with no fees and a credit owned to the customer

## Workaround

I can't find a workaround. For this week, I thought on changing manually all previous order to refund the correct amount. However this is currently a nightmare because :
1. The only report giving me shipping fees and shipping method is OC customer totals, and the line with the fees are only the total line, so I cannot filter on both the fees and the shipping method to get an accurate list of all orders in that case
2. When trying to change the shipping method as a manager, the amount of the fee is not correct: it is showing a 6,33 euros amount, which is the fee + a 5,5 % VAT. But shipping method fees are entered VAT incl. I will open another issue for that
3. Even if I fix this, I don't have a workaround for deactivating the delivery on Friday. Tags are not manageable for this. They have 200 customers a week. Or would it work to just had a tag that said this shipping method is invisible for everyone ? And then remove it when needed?
Even if it would fix it, I cannot chase all my hubs de-activating shipping methods. So I think we need a proper solution here?

## Severity

## Your Environment
* Version used: v2.9.7

## Possible Fix

De-activated shipping method should stay available for the previous orders or for the manager when editing an order?

luisramos0:
#### What? Why?

Closes #5367

#### Release notes

### DIFF
--- a/app/controllers/spree/admin/orders/customer_details_controller.rb
+++ b/app/controllers/spree/admin/orders/customer_details_controller.rb
@@ -25,7 +25,6 @@ module Spree
 
             AdvanceOrderService.new(@order).call
 
-            @order.shipments.map(&:refresh_rates)
             flash[:success] = Spree.t('customer_details_updated')
             redirect_to admin_order_customer_path(@order)
           else
--- a/app/controllers/spree/admin/orders_controller.rb
+++ b/app/controllers/spree/admin/orders_controller.rb
@@ -33,8 +33,6 @@ module Spree
       end
 
       def edit
-        @order.shipments.map(&:refresh_rates)
-
         AdvanceOrderService.new(@order).call
 
         # The payment step shows an error of 'No pending payments'


RachL:
@luisramos0 I guess you will close this and we can move #5367 to all the things?

luisramos0:
yes, this was a draft, and not really valid, we will need to adapt the refresh_rates process. We cant remove it otherwise new shipping methods will not be available to be selected, which would be another problem.

PR #5375, opened 2020-05-05
[Spree 2.1] OFN v3 comes with sample orders :-)

### LINKED ISSUE #4388: Sample data should include Orders

luisramos0:
#### What? Why?

Closes #4388
I got tired of creating orders manually on a fresh local db :-)

I had some issue making this work in master because of the order workflow, it works in rails 4 so we can keep it there.

#### What should we test?
Load sample data and see the new orders :-)

#### Release notes
Changelog Category: Added
Added sample orders to sample data


luisramos0:
ok, I needed an order with shipped items to test return authorizations in rails 4 branch :-)
I came to this PR.
I found a little problem with the shipment state and fixed it with a simple reload. New commit.
I can now see return authorizations and the shipment looks good.

I am going to consider this tested and merge this directly, I hope you guys don't mind :pray:

PR #5380, opened 2020-05-07
Update jquery-ui-rails version to 4.1.1

### LINKED ISSUE #5201: Missing jquery-ui image
Possibly since some of our Spree asset changes a while back, we've been getting _tons_ of errors for this:

```
open() "/home/openfoodnetwork/apps/openfoodnetwork/current/public/assets/jquery-ui/ui-bg_flat_75_ffffff_40x100-bd8d4ac9eb16341c108b8ef81637c3f5.png" 
failed (2: No such file or directory)
```

It's some tiny jquery-ui image file that apparently isn't being loaded as expected now in the asset pipeline. The file is located in the `jquery-ui-rails` gem.

It's a very small thing, but the downsides to this are:

- we get vast amounts of error logs for this that are just noise, making it harder to find actual errors
- when it's not found (which is all the time) the server returns a whole HTML page worth of data in the response

#### Possible solutions:

- investigate and fix it so we return the image correctly
- remove it from the relevant CSS file (it looks like it might actually not be needed anyway)

ckhung5:
#### What? Why?
There are some noise about assets not found related to jquery-ui-rails gem.

Relates to #5201 but doesn't solve it.

~~This change will solve the noise for unable to find `ui-bg_flat_75_ffffff_40x100.png`. This is a bug in the gem itself and they introduce a `depend_on` as mentioned in https://github.com/jquery-ui-rails/jquery-ui-rails/pull/59 and https://github.com/jquery-ui-rails/jquery-ui-rails/pull/60.~~

So we only just need to update jquery-ui-rails to version 4.1.1

#### What should we test?

Steps to reproduce the bug here: https://github.com/openfoodfoundation/openfoodnetwork/pull/5380#issuecomment-625021684

We should also do a sanity check on admin datepickers and dropdowns with search fields.

#### Release notes

Update jquery-ui-rails version to 4.1.1 to introduce some depend_on assets in the gem as there are currently some assets errors in the application.

Changelog Category: Fixed

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -418,7 +418,7 @@ GEM
     jquery-rails (3.1.5)
       railties (>= 3.0, < 5.0)
       thor (>= 0.14, < 2.0)
-    jquery-ui-rails (4.0.5)
+    jquery-ui-rails (4.1.2)
       railties (>= 3.1.0)
     json (1.8.6)
     json_spec (1.1.5)
@@ -726,7 +726,7 @@ DEPENDENCIES
   immigrant
   jquery-migrate-rails
   jquery-rails (= 3.1.5)
-  jquery-ui-rails (~> 4.0.0)
+  jquery-ui-rails (~> 4.1.1)
   json_spec (~> 1.1.4)
   jwt (~> 2.2)
   kaminari (~> 0.14.1)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -726,7 +726,7 @@ DEPENDENCIES
   immigrant
   jquery-migrate-rails
   jquery-rails (= 3.1.5)
-  jquery-ui-rails (~> 4.1.1)
+  jquery-ui-rails (~> 4.1.2)
   json_spec (~> 1.1.4)
   jwt (~> 2.2)
   kaminari (~> 0.14.1)
--- a/Gemfile
+++ b/Gemfile
@@ -107,7 +107,7 @@ gem "foundation-rails"
 
 gem 'jquery-migrate-rails'
 gem 'jquery-rails', '3.1.5'
-gem 'jquery-ui-rails', '~> 4.0.0'
+gem 'jquery-ui-rails', '~> 4.1.1'
 gem 'select2-rails', '~> 3.4.7'
 
 gem 'ofn-qz', github: 'openfoodfoundation/ofn-qz', ref: '60da2ae4c44cbb4c8d602f59fb5fff8d0f21db3c'
--- a/Gemfile
+++ b/Gemfile
@@ -107,7 +107,7 @@ gem "foundation-rails"
 
 gem 'jquery-migrate-rails'
 gem 'jquery-rails', '3.1.5'
-gem 'jquery-ui-rails', '~> 4.1.1'
+gem 'jquery-ui-rails', '~> 4.1.2'
 gem 'select2-rails', '~> 3.4.7'
 
 gem 'ofn-qz', github: 'openfoodfoundation/ofn-qz', ref: '60da2ae4c44cbb4c8d602f59fb5fff8d0f21db3c'


mkllnk:
How did you verify that the gem update fixes the bug if you couldn't re-produce it?

I think you would need to pre-compile assets on your local machine and change the code to not compile them on-the-fly in development to re-produce this error. You'll find the options in `config/environments`.

At least I can re-produce this in staging:

 - Log in as enterprise user.
 - Go to Admin -> Orders.
 - Click on a date field and observer the network tab in the developer console.

Screenshot from 2020-05-07 14-24-14

That 404 error needs to be gone.

ckhung5:
> How did you verify that the gem update fixes the bug if you couldn't re-produce it?
> 
> I think you would need to pre-compile assets on your local machine and change the code to not compile them on-the-fly in development to re-produce this error. You'll find the options in `config/environments`.
> 
> At least I can re-produce this in staging:
> 
> * Log in as enterprise user.
> * Go to Admin -> Orders.
> * Click on a date field and observer the network tab in the developer console.
> 
> Screenshot from 2020-05-07 14-24-14
> 
> That 404 error needs to be gone.

Cool. Thank you for showing how to reproduce the error. I will check it out at my side :)

Matt-Yorkley:
> At least I can re-produce this in staging

Looks like we have some easy steps for a tester to verify :+1:

Gemfile.lock
```diff
@@ -418,7 +418,7 @@ GEM
     jquery-rails (3.1.5)
       railties (>= 3.0, < 5.0)
       thor (>= 0.14, < 2.0)
-    jquery-ui-rails (4.0.5)
+    jquery-ui-rails (4.1.2)
```
luisramos0, line 421: the upgrade is actually to 4.1.2 right?
I guess the gemfile can have 4.1.2 instead of 4.1.1?

ckhung5: @luisramos0 Yes. It looks like it is using 4.1.2 even it is told to upgrade to 4.1.1 . I can change this .

ckhung5:
I tried testing it locally. I manage to tweek the `development.rb` for not auto compile assets but it broke the angular part where the loading icon keeps loading, there is angular error in chrome console and the haml file does not load correctly.

Screen Shot 2020-05-08 at 11 54 40 am

How are the haml files being compile?

ckhung5:
Seems like I screw up the precompile assets. I cleared them out and re-compile again. This time, there is no error and I used the same way that @mkllnk said. 

Screen Shot 2020-05-08 at 12 24 11 pm even in Master branch. 

I not too sure what to do to re-produce in my local :(

mkllnk:
Did you delete your compiled assets and re-compile them before testing with master? Otherwise it will probably find the asset that was compiled (copied) in your fixed branch.

Anyway, it's a good upgrade anyway so we'll test this in staging.

ckhung5:
@mkllnk Yes.. I did clean the compiled assets and re-compile it.

filipefurtad0:
Hi @ckhung5,

I followed @mkllnk's test procedure, and checked the 404 error on requesting the file:
ui-bg_flat_75_ffffff_40x100-8a71ce304ee5d4e3ffc09efecc9b28e8.png

Before the PR:
image

the 404 error, is visible, but:

after the PR:
image

the error is still visible.

I found no issues on a basic sanity check, but from the pics above, I am not entirely sure the PR solves the issue... Thoughts?

luisramos0:
@filipefurtad0 this PR is just upgrading a dependency so, if, from your perspective, nothing is worse than before, we should merge. If the issue is not solved we can keep the issue open.

filipefurtad0:
Hi @luisramos0,
thanks for your feedback.

The testing notes led to the comment from @mkllnk #5380 (comment) above, which states "That 404 error needs to be gone.", hence my doubts here.

But I did not find loss in functionality, so I think this is good for merging.
Thanks again!

ckhung5:
Hi @filipefurtad0 .

Pardon me for asking. Do you do a `docker-compose build` if you are using docker or `bundle install` if you are using `bundle exec rails server` ? 

It is very odd that I can't reproduce the error.

Matt-Yorkley:
@ckhung5 Testing happens on our staging servers. They are similar to the production environment and don't use Docker...

PR #5286, opened 2020-04-22
4966 missing translation admin prod list

rmklaus12:
#### What? Why?

Closes #4966 

Translation keys were missing on the name field of the admin Products list. This fix (adding the appropriate keys to the config/locales/en.yml file resolved the missing translation keys issue. 

#### What should we test?
Is translation present on the name field within the admin Products List at /admin/products?

#### Release notes
Translation now working on the admin Products List

Changelog Category:  Fixed

#### Dependencies
Not dependent, but is related (in the resolution) to #5246 

#### Documentation updates
No updates required

Issue #4966, opened 2020-03-10
Missing translation key in admin product list


RachL:
## Description
Column "name" has a missing translation key

## Steps to Reproduce
1. Go to `/admin/products`

## Animated Gif/Screenshot
![image](https://user-images.githubusercontent.com/6525576/76326761-6b943b00-62e9-11ea-81d4-8ac153e10971.png)

## Severity
bug-s3: a feature is broken but there is a workaround

## Your Environment
* Version used: v2.7.8
* Browser name and version: Firefox 73
* Operating System and version (desktop or mobile): Desktop

rmklaus12:
I've had a poke at this and believe I fixed it? My first issue and it hasn't been assigned to me, but wanted to get a handle on the translations using this as my first issue test. Does it need to be assigned before I do a PR?

RachL:
Hi @rmklaus12 just putting a comment on the issue is enough to start a PR :) Thank you! Once the PR is there it will be linked to this issue.
We use transifex for all translation related issues. I'm unclear how it translate to your local environment, but this page on the wiki should contain all info: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-%28i18n%29

Don't hesitate to ping us on slack if it is not the case!

rmklaus12:
@RachL - I believe the code for the translation itself was fine, it was just missing the keys in the en.yml file, which is what I added. I have been reading up on the other bit, all the pieces are coming together :-) I look forward to any feedback to see if I did miss something! I will do the PRs for both now.

PR #5389, opened 2020-05-08
Fix translation keys in en.yml file from previous conflicting PRs

rmklaus12:
#### What? Why?

Resolves conflict between #5286 and #4966. 

Updated en.yml to reflect changes made previously in two separate PRs. A conflict between the two created errors within the en.yml file

#### What should we test?
1. Is translation present on the orders list (number, state, payment_state, shipment_state, email, and total fields) at /admin/orders?q[s]=completed_at+desc
2. Is translation present on the name field within the admin Products List at /admin/products?

#### Release notes
Resolves conflict in en.yml file which was breaking the general settings translation keys

Changelog Category: Fixed

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3354,6 +3354,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
           unit: "Unit"
       shared:
         sortable_header:
+          name: "Name"
           number: "Number"
           state: "State"
           payment_state: "Payment State"
@@ -3361,9 +3362,6 @@ See the %{link} to find out more about %{sitename}'s features and to start using
           email: "Email"
           total: "Total"
       general_settings:
-      shared:
-        sortable_header:
-          name: "Name"
         edit:
           legal_settings: "Legal Settings"
           cookies_consent_banner_toggle: "Display cookies consent banner"


luisramos0:
I wonder why you needed to create two PRs, let me know if you have any doubts about git, we can help :+1:

rmklaus12:
> I wonder why you needed to create two PRs, let me know if you have any doubts about git, we can help 👍

I think I need to start work later so I can work with you ;-) There were two separate issues I grabbed that both changed en.yml. Now I know to look at the big picture and make sure I'm not doing two PRs on the same file! In this case, two issues that can be one PR...is it okay to address both the changes and then do one PR?

luisramos0:
hello, my question was not about the problem (I understand the cause of the problem).
My question was about this PR and PR #5390. Why two?

rmklaus12:
@luisramos0 - this closed PR was an error on my part. I set the other one as ready for review. Thank you for your help!

luisramos0:
ok, cool, I am only asking because there could be something I could help with :+1:

rmklaus12:
Thank you! I appreciate it and will reach out when I have questions :-)

PR #5392, opened 2020-05-08
Make "backoffice only" ship methods work and remove option "frontoffice only"

luisramos0:
#### What? Why?
This doesnt fix the root cause of #5367 but I think it's a good workaround. And it makes something that was broken work.

The "display on" field of shipping methods was basically ignored everywhere before this PR.
After this PR:
- the "back office only" ship methods will not show up on checkout :tada:
- we will not see the "frontoffice only" option: making shipping methods only available in checkout (this would not be straight forward to fix because we would need to make the backoffice ignore these shipping methods...).

So, re #5367, managers instead of unlinking the ship methods from their enterprise, they can just set them as "backoffice only" so that it's not available on checkout.

#### What should we test?
We need to verify ship methods tagged with this "back office only" tag do not show up in the checkout. We need to verify these methods will still be available in admin when editing an order.

#### Release notes
Changelog Category: Fixed
Shipping methods can now be marked "Backoffice only" and that will work: the shipping method will not be available for users in the checkout process but will be available in the backoffice when editing an order.

### DIFF
--- a/app/helpers/enterprises_helper.rb
+++ b/app/helpers/enterprises_helper.rb
@@ -14,7 +14,7 @@ module EnterprisesHelper
   def available_shipping_methods
     return [] if current_distributor.blank?
 
-    shipping_methods = current_distributor.shipping_methods
+    shipping_methods = current_distributor.shipping_methods.display_on_checkout.to_a
 
     applicator = OpenFoodNetwork::TagRuleApplicator.new(current_distributor, "FilterShippingMethods", current_customer.andand.tag_list)
     applicator.filter!(shipping_methods)
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -30,6 +30,7 @@ Spree::ShippingMethod.class_eval do
   }
 
   scope :by_name, -> { order('spree_shipping_methods.name ASC') }
+  scope :display_on_checkout, -> { where("display_on is null OR display_on = ''") }
 
   # Return the services (pickup, delivery) that different distributors provide, in the format:
   # {distributor_id => {pickup: true, delivery: false}, ...}
--- a/app/models/spree/shipping_method_decorator.rb
+++ b/app/models/spree/shipping_method_decorator.rb
@@ -30,7 +30,7 @@ Spree::ShippingMethod.class_eval do
   }
 
   scope :by_name, -> { order('spree_shipping_methods.name ASC') }
-  scope :display_on_checkout, -> { where("display_on is null OR display_on = ''") }
+  scope :display_on_checkout, -> { where("spree_shipping_methods.display_on is null OR spree_shipping_methods.display_on = ''") }
 
   # Return the services (pickup, delivery) that different distributors provide, in the format:
   # {distributor_id => {pickup: true, delivery: false}, ...}
--- a/app/views/spree/admin/shipping_methods/_form.html.haml
+++ b/app/views/spree/admin/shipping_methods/_form.html.haml
@@ -18,7 +18,7 @@
       .alpha.three.columns
         = f.label :display_on, t(:display)
       .omega.eight.columns
-        = select(:shipping_method, :display_on, Spree::ShippingMethod::DISPLAY.collect { |display| [t(".#{display}"), display == :both ? nil : display.to_s] }, {}, {class: 'select2 fullwidth'})
+        = select(:shipping_method, :display_on, [[t(".both"), nil], [t(".back_end"), "back_end"]], {}, {class: 'select2 fullwidth'})
         = error_message_on :shipping_method, :display_on
 
   .row
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3221,9 +3221,8 @@ See the %{link} to find out more about %{sitename}'s features and to start using
           zone: "Zone"
           calculator: "Calculator"
           display: "Display"
-          both: "Both"
-          front_end: "Front End"
-          back_end: "Back End"
+          both: "Both Checkout and Back office"
+          back_end: "Back office only"
           no_shipping_methods_found: "No shipping methods found"
         new:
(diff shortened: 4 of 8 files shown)


tschumilas:
I want to make sure I understand this -  because right now I don't.   Here's the use case:   Producer sets shipping method A.   Producer opens Order Cycle 1.  Customer 1 buys and selects Shipping method A.  Shipping method A is now attached to their order that the producer sees,   and on the customers' order confirmation,  and on the customer's invoice.     Producer removes (or makes 'back office only' ) shipping Method A because that time slot is full.  Producer adds (shipping method B). Order Cycle is STILL OPEN .  Customer 2 places an order -  and does NOT see the option for Shipping method A.  Customer 2 chooses shipping method B.    So now we have:  Customer 1's invoice with shipping method A  -  and that method 'sticks' even if the producer opens and adjusts an order.   And we have Customer 2's order with shipping method B - and that method sticks even if the producer adjusts the order.    And in the reports -  each customer will have the correct shipping method showing -- no matter at what time the order cycle is run?  


My follow up will be what happens to reports when Order Cycle 2 opens with new shipping methods.

Sorry - its just this is a really frequent use case right now as users are changing shipping methods to  make for pickup windows  to ensure physical distancing.   I want to make sure I can explain it.

luisramos0:
yes, that's all correct.


--- what happens to reports when Order Cycle 2 opens with new shipping methods.

I think the answer is nothing.

app/helpers/enterprises_helper.rb
```diff
@@ -14,7 +14,7 @@ def current_customer
   def available_shipping_methods
     return [] if current_distributor.blank?
 
-    shipping_methods = current_distributor.shipping_methods
+    shipping_methods = current_distributor.shipping_methods.display_on_checkout.to_a
```
mkllnk, line 17: Why `to_a`?

app/models/spree/shipping_method_decorator.rb
```diff
@@ -30,6 +30,7 @@
   }
 
   scope :by_name, -> { order('spree_shipping_methods.name ASC') }
+  scope :display_on_checkout, -> { where("display_on is null OR display_on = ''") }
```
mkllnk, line 33: It's better to include the table name `spree_shipping_methods` in case we join other tables which have the same column (in the future). But this is much easier to read. I'm happy to leave this as is until future versions of Rails make it `or` easier.

app/helpers/enterprises_helper.rb
luisramos0: otherwise it's a relation and it was not working with it. 
I cant remember exactly the error but I think it's related to some magic inside TagRuleApplicator (called just after this).

app/models/spree/shipping_method_decorator.rb
luisramos0: I thought about this actually and decided to leave it like this for simplicity... yeah, I am not sure...

Matt-Yorkley: I think adding the table name explicitly is a good idea

luisramos0: ok, I have added it :+1:

filipefurtad0:
Hi @luisramos0,

Verified/reproduced the issue: 
- as admin, unchecking the shipping method automatically assigns another shipping method to an order, which may cause "fake" price adjustments.
- as superadmin, changing the "Display" to "Back End" does not remove the shipping method from checkout.

After this PR:
- as superadmin: there are only two options, to set "Display" - "Back end" or "Both" (Front End is gone)
- "Back end" option removes the shipping method from the customers checkout page, and leaves the order untouched, the shipping method is there, the costs remain the same.
- verified that root cause is still there, but this seems like a great workaround for the issue at hand :tada: 

Ready to go!

filipefurtad0:
Ah, btw., I saw (a bit too late) that there is a "Line is too long" issue from codeclimate... But I guess this should not influence the test/PR, right? Hope so.

luisramos0:
thanks @filipefurtad0 :-)

RachL:
@luisramos0 @filipefurtad0 I'm checking this as part of the release testing and the switch is only available as super admin. Yet when I read the PR description I understand it should be available for manager as well.

We introduce a text with #5399 and it's a bit confusing if you cannot see the dropdown as a manager but the text tells you that you can.

filipefurtad0:
Hey @RachL ,

Yes, that's a good point. The text is only accurate for superadmins, where the dropdown appears - I missed that detail.

If we change this text now, maybe we could also mention the tag system to hide shipping methods as well (as you describe in your notes - point 3, in #5367)? I guess this is always an option, which works too.

From the release notes, I understood this PR rather as a fix for an existing feature (the dropdown menu for superadmin) and not it's introduction for hub managers. But I guess it would make sense, as hub manager to be able use this dropdown as well.

RachL:
@filipefurtad0 Luis has written this in the PR:

> managers instead of unlinking the ship methods from their enterprise, they can just set them as "backoffice only" so that it's not available on checkout.

This sentence makes me understand that this is not super admin only. If it was designed to be super admin only, it is really not a suitable workaround for #5367 ...

So I guess that before releasing for now we need to change the text and include something like "contact your instance manager". But this will increase the support load and won't be something we can do on the long term.

luisramos0:
Hello, it's my mistake, I didnt see that the flag was super admin only. It's a simple line I can remove and it will show the option to anyone with access to the edit page, shall I do that? do you think this will create problems elsewhere?

Rob added this if statement to hide this from managed in 2014:
https://github.com/openfoodfoundation/openfoodnetwork/commit/004548a4fe514246679842c65c1778aec473447a

luisramos0:
I need to add an automated test to verify automatically the manager can edit the field, but basically, this is the PR:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5493

PR #5395, opened 2020-05-08
Skip unreliable caching test

sauloperez:
#### What? Why?

This spec failed two times in a row and it's blocking the release.


Matt-Yorkley:
This is a weird one. The flaky spec is not to do with the timings at all. 

In the test, we check the last saved timestamp for `Spree::Taxon`, then explicitly update and save a taxon, then check the last saved timestamp again. In some cases in Semaphore it doesn't recognise the taxon as having been updated. I ran it 20 times locally and it was green every time.

PR #5396, opened 2020-05-08
Make test not depend on response's order

sauloperez:
#### What? Why?

I'm assuming this is failing in CI due to the order in which the closed shops are returned.

This is blocking the release's build.

#### Release notes

Make test not rely on the order of the returned records.

Changelog Category: Fixed


sauloperez:
That fixed it but the build is still red due to #5395

Issue #5078, opened 2020-03-25
Deploy OFN free from external data collection such as google.


albjeremias:
## What is the problem we are solving 

The problem is that after I deployed OFN with docker-compose, i went to look into the network requests and I was really disappointed to find out that there are requests to google and to stripe, both of them companies which survive from data mining/collection from users.

I also posted on the OFN forum hope you don't feel as spam, but in there i'm requesting instructions on how-to disable it, and here on please to don't set to default to track OFN users.
https://community.openfoodnetwork.org/t/i-need-help-on-configuring-ofn-without-external-data-collection/1889

## Success factors = expected outcome 

I expected that OFN by default does do any network request to external servers at least as default.
I known this is not a simple issue, but i believe its very important.

## Useful information for inception

docker-compose using v2.8.0

Matt-Yorkley:
Is this something you'd like to work on, @albjeremias ?

It would be great to move away from Google maps services, we currently have various bits of functionality that depend on them for Geocoding, searching for local food enterprises by location, and the map display page. 

In terms of using Stripe, we take online payments in lots of countries internationally and the alternatives there are pretty limited. We won't be moving away from Stripe any time soon.

albjeremias:
I could work on it.. im not savvy with ruby.. but with support i think i can tackle little bits of it! :) as of stripe, i guess its a matter of not being activated by default?

cillian:
:wave:... I got Open Street Map working [here](https://github.com/cillian/openfoodnetwork/commit/87427e49c18d60fb7b2dfbd69d1655c6ecc98aa8). The search works a bit differently than usual but if that's not a problem I could open a pull request?

![ofn-open-street-map](https://user-images.githubusercontent.com/101088/81427419-b3e9a100-9152-11ea-98e6-c4274ccbbc42.gif)

![open-street-map-config](https://user-images.githubusercontent.com/101088/81427705-12af1a80-9153-11ea-95bc-80c1535572e6.png)

Although I haven't got tests working locally yet so I haven't checked if they are all passing.

Matt-Yorkley:
Nice! Sure, open a pull request @cillian and we'll take a look.

RachL:
Exciting!!! 😍

luisramos0:
After amazing work from @cillian we have an alternative to gmaps that uses OSM for maps.
[...]

PR #5398, opened 2020-05-10
Add support for using Open Street Map on the Map and Group pages.

cillian:
#### What? Why?

For https://github.com/openfoodfoundation/openfoodnetwork/issues/5078

This lets people use Open Street map.

ofn-open-street-map

The map is displayed using https://leafletjs.com/

The search field for the Open Street Map works differently than searching on Google Maps. It matches producers by their name or address instead of matching place names all over the world because this way easier to implement I found.

To enable Open Street Map go to the _Admin -> Configuration -> Content_ section and click 'Open Street Map Enabled'.

open-street-map-config

The _Open Street Map Provider Name_ setting can be used to configure different tile providers thanks to the Leaflet-providers extension (https://github.com/leaflet-extras/leaflet-providers)

Some tile providers require an API key, this can provided in JSON format e.g. `{ apiKey: 123 }` in the _Open Street Map Provider Options_ setting.

Each tile provider has their own usage policy so this should be checked before enabling Open Street Map.

#### What should we test?

- Enabling Open Street Map in _Admin > Configuration > Content_.
- The map page and search field.
- The group map page and search field.

Note I haven't got automated tests working locally yet but I will try check they are passing later.

#### Release notes

Adds support for displaying maps usings Open Street Map.

Changelog Category: Added

#### Discourse thread

https://community.openfoodnetwork.org/t/i-need-help-on-configuring-ofn-without-external-data-collection/1889

#### Documentation updates

- https://github.com/openfoodfoundation/openfoodnetwork/wiki/Set-and-Deploy-on-Heroku (Mention Open Street Map is an option in _Post-install setup_)
- https://github.com/openfoodfoundation/ofn-install/wiki/Administration (Add Open Street Map to _External Services_ section)


Matt-Yorkley:
Thanks so much! Leaflet looks like exactly what we needed. :smile:

I think we could probably bring back some of the lost search functionality with a bit of tweaking, possibly via the `geocoding` gem (which also can be configured to not use google), but it would need some investigating.

We'll probably need to make some configuration changes in the `ofn-install` repo as well, and make sure we can transition without disruption (I'm not asking you to do any of this, just mentioning it).

Exciting stuff! It might be a while before it can be merged though :+1:

luisramos0:
You are making many people happy with this PR @cillian :-)

I think that instead of making this PR ready to replace google maps, I'd just adapt the config in this PR to make OSM the default maps solution in OFN.
If I read the code correctly this leaves google maps as the default. Ideally we would solve the original request and make this limited (for now!) solution be the default. Managers can go to the back office and active google maps if they want (I think initially everyone will keep using google maps because of geocoding).

I'd say we should ship this (default OSM + google maps as the extra solution). From there we can gradually improve this OSM solution and make it replace google maps for all instances.
Thoughts?

I'll do a detailed code review when we agree on what's the general approach to this PR.

Anyway, this PR made my day!

RachL:
> (I think initially everyone will keep using google maps because of geocoding).

It is true we will need to add some routes and update some details for some farms. But we already do hit for Google and it is super easy on OSM. For the CSA network in France, the farmer can use this tool : https://nominatim.openstreetmap.org/
when adding his address. If he can't find his address, there is a link to add his address or ask support so that someone adds his address onto OSM. It is working pretty good.
So the big question is how our enterprise address field will play with OSM. That's something we should add in the tests here: add a new enterprise and see if it shows up on the map.

Also I understand that in countries like Australia where OSM is really poor in data, the need for Google maps is still there. I think the most important thing is to keep the choice. Whether or not one should be default and not the other I have no opinion. However if OSM is default, we should make sure that the deployment does not change this for all current instance. Could endup with a lot of problems.

luisramos0:
Just to be clear, making google maps not be the default is what would close #5078. I think #5078 is important.

cillian:
:+1: ..sounds good, I'll leave it with you all for now then. If it needs any changes please let me know.

Matt-Yorkley:
I'm just having a little play with this now. It's working really well. I think I'll try to make some improvements to it at some point, but for now I think I agree we can just merge it and iterate.

We should add a notice in the Wiki though: the current default provider has usage restrictions and should only only be used for a bit of testing. It shouldn't be used long-term or on a larger production server. We should also have a look at the alternate free and paid providers that can be used, and maybe recommend one.

app/views/groups/show.html.haml
```diff
@@ -32,13 +34,16 @@
...
-                    %map-osm-tiles
-                    %map-search
-                    %ui-gmap-markers{models: "mapMarkers", fit: "true",
-                    coords: "'self'", icon: "'icon'", click: "'reveal'"}
+              .map-container{"ng-if" => "(isActive(\'/map\') && (mapShowed = true)) || mapShowed"}
+                - if ContentConfig.open_street_map_enabled
+                  = render partial: 'map/open_street_map'
+                - else
+                  %map
+                    %ui-gmap-google-map{options: "map.additional_options", center: "map.center", zoom: "map.zoom", styles: "map.styles", draggable: "true"}
```
Matt-Yorkley, line 42: Note: we should refactor this repeated code block in `app/views/groups/show.html.haml` and `app/views/map/index.html.haml` and move it into a partial. It needs a slightly different `models` attribute in each one on the `ui-gmap-markers` element.

app/models/preference_sections/map_section.rb
```diff
@@ -0,0 +1,15 @@
+module PreferenceSections
+  class MapSection
```
Matt-Yorkley, line 2: Note: the preferences are hard to find in the config section. We could move them to a new map settings page.

app/assets/javascripts/darkswarm/directives/open_street_map.js.coffee
```diff
@@ -0,0 +1,100 @@
...
+
+    displayEnterprises = ->
+      for enterprise in Enterprises.enterprises
+        if enterprise.latitude? && enterprise.longitude?
+          marker = buildMarker(enterprise, { lat: enterprise.latitude, lng: enterprise.longitude }, enterprise.name).addTo(map)
+          enterpriseNames.push(enterpriseName(enterprise))
+          markers.push(marker)
+
+    displaySearchField = () ->
+      new Autocomplete('#open-street-map--search',
```
Matt-Yorkley, line 79: Note: we can use a different search box using Geocoding (like the /shops search) and get the old search behavior back. If we do that we can probably ditch the Autocomplete javascript dependency. See `geocoding` gem.

app/views/groups/show.html.haml
Matt-Yorkley: Done here: #5494

filipefurtad0:
Hi @cillian - thank you for this PR!

One issue though: I see codeclimate has several types of issues to resolve - perhaps these should be addressed before testing?

app/serializers/api/open_street_map_config_serializer.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+#
```
luisramos0, line 2: this should be an empty line.

and if you have one more minute you can fix the other warning 
image

by doing:
module Api
     class OpenStreetMapConfigSerializer < ActiveModel::Serializer
         ...
     end
end

luisramos0: I think we can ignore the other 2 warnings :+1: 

Good point @filipefurtad0 :+1:

cillian:
Hi @filipefurtad0, @luisramos0 

No problem, that's done now. Let me know if it needs more changes.

luisramos0:
Nice, thanks @cillian Ready for testing :+1:

Matt-Yorkley:
I've rebased this to resolve some merge conflicts and added a slight adjustment that was needed after the rebase.

filipefurtad0:
Hi @cillian ,

Thanks for tackling this!

The PR definitively removes dependency from google maps, as shown on the console (second pic, below) but somehow the map is not being displayed.

Before this PR
image

After this PR, and enabling Open Street Map in the Content section (superadmin):
image

This might be due to staging issues; maps never really worked well on fr-staging - right @RachL? Perhaps a good idea to test this on a different server.

More on this soon @cillian - thanks again :+1:

RachL:
@filipefurtad0 yes gmaps is not working in FR. But I thought it was specific to gmaps. Worth checking on another staging.

filipefurtad0:
Alright,
Different staging and server, show the expected result:

Before:
image

After:
image

Awesome!
Thank you @cillian :tada:

kirstenalarsen:
Can anyone tell me what would happen to Aus enterprise addresses that aren't on osm if we switched this? Would they just disappear from the map? Would we need to ask all existing enterprises to check and create themselves, do it ourselves, or is there an automated way to do it?

Matt-Yorkley:
We're adding an optional alternative, not really changing anything yet. It needs some more work, but the idea is that instances will be able to use it without any of the problems you've listed.

kirstenalarsen:
thanks @Matt-Yorkley I understand that :) I was just wondering if instances like Aus did want to switch to it, if we did switch to it, what would happen?

kirstenalarsen:
or in saying 'it needs more work' you're imagining that there would be solutions to those problems that would mean that we could switch?

luisramos0:
this issue will help to understand that Kirsten: https://github.com/openfoodfoundation/openfoodnetwork/issues/5542
The main difference is the search bar is searching for enterprise names, not locations.
Then, because we dont have any "manual geocoding", I think new enterprises will just not appear on the map. I think existing geocoded enterprises would appear on the map.
Is this correct, Matt?

cillian:
That's correct any enterprise with the `latitude` and `longitude` already set on the address record should appear. The search bar doesn't search for locations unfortunately although if you type in an address it will filter enterprises matching that address in the drop down.

Matt-Yorkley:
Actually the geocoding we do when an enterprise is saved is handled by a separate gem, which is not connected to the maps javascript. So new enterprises would still appear on the map. :+1:

I've already written up a bit of a roadmap detailing the bits of work would need to be done to solve those issues.

luisramos0:
wait, but if geocoding still works if you switch to osm maps, then the only problem with using osm is the search. I wonder if that doesn't make some instances want to switch immediately.

Matt-Yorkley:
It's close, but not close enough...

tschumilas:
So -  I activated this on OFN_CAN  -  and there are a few issues  you can see:   first -  on the main page, the map opens to Japan -  is there a way to have it default open to the instance?      Second -  the map formats like a banner image,  versus a large square for SOME but not all of our groups pages.   See these links:  https://openfoodnetwork.ca/map,      See this group map:  https://openfoodnetwork.ca/groups/https-www-stcatharines-ca-en-experiencein-stcatharinesfarmersmarket-asp#/map  versus this group map:  https://openfoodnetwork.ca/groups/guelph-farmers-market#/map

Any idea why the difference in the group maps?

And -  now that I see this,  I probably want to go back to google maps for OFN-CAN until we fix this more -   can I just back out?   If I just uncheck open street map in configuration -  will it go back?  (maybe I shouldn't have experimented )

RachL:
@tschumilas yes please switch back to google maps. There quite a few work to do before we can use this on live instance: https://github.com/openfoodfoundation/openfoodnetwork/issues/5542

tschumilas:
OK -  will do,  sorry I jumped the gun here.  I'm guessing I just de-select the open street map and presto - I'll be back.?

Issue #5300, opened 2020-04-24
Products Sold Exceed Max Stock Level


lbwright22:
## Description

Large UK Hub reported that for their OC 20/04/20 :
1) at least two producers found orders for products in their order cycle summaries which exceeded the max. stock level they placed.
These producers were Hester's Harvest and Hoe Made Acre
2) customer's received invoices containing products listed as purchased with zero quantity:
R781345776 Asparagus Green (250g) Hester's Harvest
R877041463 Sausages- usually frozen- Pork Honey Roast and Mustard 500g Trehill Farm
3) producer found products were left with negative stock levels after the order cycle (Asparagus for Hester's Harvest had stock of -3: initial stock level was 50, OC supplier report says 53 were sold)

There may be other examples not yet found.

## Expected Behavior

Product with finite stock level should not be available for sale once maximum capacity has been reached.

## Actual Behaviour

Finite stock level ('in stock') exceeded during an order cycle.

## Steps to Reproduce

1.
2.
3.
4.

## Animated Gif/Screenshot

![HHarvest](https://user-images.githubusercontent.com/45104035/80196301-19974280-8615-11ea-8e2f-88848683212e.jpg)
![R877](https://user-images.githubusercontent.com/45104035/80196492-6713af80-8615-11ea-9722-94d3c47482e9.jpg)
![R781](https://user-images.githubusercontent.com/45104035/80196510-6e3abd80-8615-11ea-954e-67627f8de8db.jpg)

## Severity

bug S2
## Your Environment

* Version used:
* Browser name and version:
* Operating System and version (desktop or mobile):

lbwright22:
![](https://github.trello.services/images/mini-trello-icon.png) [Tamar negative stock](https://trello.com/c/PLpYAc75/944-tamar-negative-stock)

luisramos0:
I will not have time to look into this any further until Monday. Maybe another dev will?

I just looked at the stock movement in the DB, it looks like there was a rush to the asparguss, yes. But movements were a few seconds apart... maybe enough to make our concurrency solution to break...

![image](https://user-images.githubusercontent.com/1640378/80205555-b7ddd500-8622-11ea-9d90-ecd28f6faa57.png)

tschumilas:
Just had a report from a hub that also had this happen.,  Getting details from her - but she says it was a really popular product -  it happended last week and this week
I'll get details here.

emilyjeanrogers:
[...]

PR #5406, opened 2020-05-12
5300 Avoid race conditions in Spree::StockItem

kristinalim:
#### What? Why?

- References #5300

I've looked into #5300, exploring if the negative inventory levels could be related to a missing callback for guest checkout, subscriptions, and the now removed backorders, but this doesn't seem to be the case.

The issue seems to be that stock levels in `Spree::StockItem`, which we check for stock levels before completing an order, could be updated by another order before the stock level is updated by the current order.

This PR introduces two changes:

* Optimistic locking for `Spree::StockItem` via `lock_version`.
* Require `count_on_hand` in `Spree::StockItem` to be positive if not backorderable.

#### What should we test?

Scenario 1: Stock restrictions for customer checking out should still work. Open two browser sessions, one logged in as admin (Session A), another logged in as customer (Session B).

1. **Session A:** Set the stock level of a product to limited stock with e.g. 4.
2. **Session B:** Add less than total stock left to cart, e.g. 3.
3. **Session A:** Update the stock level of the product to be less than the quantity in cart, e.g. 2.
4. **Session B:** Attempt to check out. This should warn of insufficient stock.
5. **Session B:** Reduce quantity in cart, e.g. 2.
6. **Session B:** Checkout should succeed.

Scenario 2: Stock restrictions for admin setting up an order should still work. Open two browser sessions, both logged in as admin.

1. **Session A:** Set the stock level of a product to limited stock with e.g. 4.
2. **Session B:** Create an order, Add less than total stock left to cart, e.g. 3.
3. **Session A:** Update the stock level of the product to be less than the quantity in cart, e.g. 2.
4. **Session B:** Attempt to complete the order. This should succeed but reduce the quantity in the order to the max available.

#### Release notes

Address race condition resulting in orders exceeding available stock.

Changelog Category: Fixed

### DIFF
--- a/app/models/spree/stock_item.rb
+++ b/app/models/spree/stock_item.rb
@@ -1,0 +1,52 @@
+module Spree
+  class StockItem < ActiveRecord::Base
+    belongs_to :stock_location, class_name: 'Spree::StockLocation'
+    belongs_to :variant, class_name: 'Spree::Variant'
+    has_many :stock_movements, dependent: :destroy
+
+    validates_presence_of :stock_location, :variant
+    validates_uniqueness_of :variant_id, scope: :stock_location_id
+
+    attr_accessible :count_on_hand, :variant, :stock_location, :backorderable, :variant_id
+
+    delegate :weight, to: :variant
+
+    def backordered_inventory_units
+      Spree::InventoryUnit.backordered_for_stock_item(self)
+    end
+
+    def variant_name
+      variant.name
+    end
+
+    def adjust_count_on_hand(value)
+      self.with_lock do
+        self.count_on_hand = self.count_on_hand + value
+        process_backorders if in_stock?
+
+        self.save!
+      end
+    end
+
+    def in_stock?
+      self.count_on_hand > 0
+    end
+
+    # Tells whether it's available to be included in a shipment
+    def available?
+      self.in_stock? || self.backorderable?
+    end
+
+    private
+      def count_on_hand=(value)
+        write_attribute(:count_on_hand, value)
+      end
+
+      def process_backorders
+        backordered_inventory_units.each do |unit|
+          return unless in_stock?
+          unit.fill_backorder
+        end
+      end
+  end
+end
--- a/app/models/spree/stock_item.rb
+++ b/app/models/spree/stock_item.rb
@@ -2,23 +2,22 @@
 
 module Spree
   class StockItem < ActiveRecord::Base
+    acts_as_paranoid
+
     belongs_to :stock_location, class_name: 'Spree::StockLocation'
     belongs_to :variant, class_name: 'Spree::Variant'
     has_many :stock_movements, dependent: :destroy
 
     validates :stock_location, :variant, presence: true
-    validates :variant_id, uniqueness: { scope: :stock_location_id }
+    validates :variant_id, uniqueness: { scope: [:stock_location_id, :deleted_at] }
     validates :count_on_hand, numericality: { greater_than_or_equal_to: 0, unless: :backorderable? }
 
-    attr_accessible :count_on_hand, :variant, :stock_location, :backorderable, :variant_id
-
     delegate :weight, to: :variant
+    delegate :name, to: :variant, prefix: true
 
     def backordered_inventory_units
       Spree::InventoryUnit.backordered_for_stock_item(self)
     end
-
-    delegate :name, to: :variant, prefix: true
 
     def adjust_count_on_hand(value)
       with_lock do
@@ -36,6 +35,10 @@ module Spree
     # Tells whether it's available to be included in a shipment
     def available?
       in_stock? || backorderable?
+    end
+
+    def variant
+      Spree::Variant.unscoped { super }
     end
 
     private
--- a/app/models/spree/stock_item.rb
+++ b/app/models/spree/stock_item.rb
@@ -6,8 +6,8 @@ module Spree
     belongs_to :variant, class_name: 'Spree::Variant'
     has_many :stock_movements, dependent: :destroy
 
-    validates_presence_of :stock_location, :variant
-    validates_uniqueness_of :variant_id, scope: :stock_location_id
+    validates :stock_location, :variant, presence: true
+    validates :variant_id, uniqueness: { scope: :stock_location_id }
 
     attr_accessible :count_on_hand, :variant, :stock_location, :backorderable, :variant_id
 
--- a/app/models/spree/stock_item.rb
+++ b/app/models/spree/stock_item.rb
@@ -17,9 +17,7 @@ module Spree
       Spree::InventoryUnit.backordered_for_stock_item(self)
     end
 
-    def variant_name
-      variant.name
-    end
+    delegate :name, to: :variant, prefix: true
 
     def adjust_count_on_hand(value)
       with_lock do
--- a/app/models/spree/stock_item.rb
+++ b/app/models/spree/stock_item.rb
@@ -47,7 +47,7 @@ module Spree
 
     def process_backorders
       backordered_inventory_units.each do |unit|
-        return unless in_stock?
+        break unless in_stock?
 
         unit.fill_backorder
       end
--- a/app/models/spree/stock_item.rb
+++ b/app/models/spree/stock_item.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 module Spree
   class StockItem < ActiveRecord::Base
     belongs_to :stock_location, class_name: 'Spree::StockLocation'
@@ -39,15 +40,17 @@ module Spree
     end
 
     private
-      def count_on_hand=(value)
-        write_attribute(:count_on_hand, value)
-      end
 
-      def process_backorders
-        backordered_inventory_units.each do |unit|
-          return unless in_stock?
-          unit.fill_backorder
-        end
+    def count_on_hand=(value)
+      write_attribute(:count_on_hand, value)
+    end
+
+    def process_backorders
+      backordered_inventory_units.each do |unit|
+        return unless in_stock?
+
+        unit.fill_backorder
       end
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


app/models/spree/stock_item_decorator.rb
```diff
@@ -0,0 +1,3 @@
+Spree::StockItem.class_eval do
+  validates :count_on_hand, numericality: { greater_than_or_equal_to: 0, unless: :backorderable? }
+end
```
luisramos0, line 3: now that we are going "bye bye spree" you can bring the full class from spree in these cases :heart: 
If you are up for it you can even bring a few more related things (that way we can save one iteration on the manual testing) like stock_movement and stock_location.
reference: https://github.com/openfoodfoundation/openfoodnetwork/issues/4826#issuecomment-611698328

kristinalim: Thank you for this, @luisramos0! For now, I just brought in `Spree::StockItem`.

app/models/spree/stock_item.rb
```diff
@@ -0,0 +1,57 @@
...
+      in_stock? || backorderable?
+    end
+
+    private
+
+    def count_on_hand=(value)
+      self[:count_on_hand] = value
+    end
+
+    def process_backorders
```
luisramos0, line 49: nice, all this backorders code can now be deleted :tada: in a separate PR I mean :-)

db/migrate/20200514174526_reset_negative_nonbackorderable_count_on_hand_in_stock_items.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def up
+    Spree::StockItem.where(backorderable: false)
+      .where("count_on_hand < 0")
+      .update_all(count_on_hand: 0)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
luisramos0, line 17: :ok_hand:

spec/models/spree/line_item_spec.rb
```diff
@@ -101,7 +101,7 @@ module Spree
       end
 
       it "caps at zero when stock is negative" do
-        v.update_attributes! on_hand: -2
+        v.__send__(:stock_item).update_column(:count_on_hand, -2)
```
Matt-Yorkley, line 104: Why is `__send__` needed here? Doesn't `variant.stock_item` work?

kristinalim: @Matt-Yorkley Yes, it turns out `Spree::Variant#stock_item` is a private method.

app/models/spree/stock_item.rb
kristinalim: This would reduce a lot of complexity! :fire:

app/models/spree/stock_item.rb
```diff
@@ -10,6 +10,7 @@ class StockItem < ActiveRecord::Base
     # rubocop:disable Rails/UniqueValidationWithoutIndex
     validates :variant_id, uniqueness: { scope: :stock_location_id }
     # rubocop:enable Rails/UniqueValidationWithoutIndex
+    validates :count_on_hand, numericality: { greater_than_or_equal_to: 0, unless: :backorderable? }
```
mkllnk, line 13: This will make the checkout fail. Did you test what the user would see in this scenario?

spec/models/spree/stock_item_spec.rb
```diff
@@ -8,7 +8,7 @@
   subject { stock_location.stock_items.order(:id).first }
 
   it 'maintains the count on hand for a variant' do
-    expect(subject.count_on_hand).to eq 10
+    expect(subject.count_on_hand).to eq 15
```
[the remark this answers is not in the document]
kristinalim: Hi @mkllnk! This spec file was originally copied directly from Spree core, and the `count_on_hand` before this PR is already 15. It was brought it in and updated for OFN code.

luisramos0:
@kristinalim please move back to Code Review when you think this is ready to be re-reviewed. Thanks!

luisramos0:
I am having a look at this.
I thought of creating PRs on top of Kristina's branch but Kristina may not be available in the next few days so I decided to work directly on Kristinas branch. I started by rebasing the PR and update stock_item to spree 2.1 (rails 4 version).

app/models/spree/stock_item.rb
```diff
@@ -7,7 +7,9 @@ class StockItem < ActiveRecord::Base
     has_many :stock_movements, dependent: :destroy
 
     validates :stock_location, :variant, presence: true
+    # rubocop:disable Rails/UniqueValidationWithoutIndex
```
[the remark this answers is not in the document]
luisramos0: I think we can do that. I moved this to the rubocop todo list.

luisramos0:
I have found that checkout controller already contains this:
rescue ActiveRecord::StaleObjectError
So we will be ok on that front.

With this last version I tried reproducing the problem by placing two orders at the same time. I only get the correct behavior, one of the sessions gets this:
image

I am now trying to write a spec that simulates two users checking out at the same time with insufficient stock for both.
I am working at controller level extending spec/controllers/checkout_controller_concurrency_spec.rb

mkllnk:
> With this last version I tried reproducing the problem by placing two orders at the same time. I only get the correct behavior

And with master? Can you actually re-produce the error? I still don't know why the synchronisation doesn't work in this case.

luisramos0:
I was trying to extend  https://github.com/openfoodfoundation/openfoodnetwork/blob/9e730e6445231c16396de2da7c33bff26c0aed74/spec/controllers/checkout_controller_concurrency_spec.rb

With a spec that would test for different orders with the same stock limited variant.
I can create a lot of deadlocks but they are not consistent...

I realize that the existing spec still passes without this:
image

And the second thread doesnt even get to the load_order call back.

Which is exactly the same problem in my attempt [here](https://github.com/luisramos0/openfoodnetwork/commit/cf5379238e042399d93af4dec4d81b72d0664005).
I dont understand why the second thread doesnt even get to load_order... I cant underastand what's going on.

Anywa, I have spent too much time on this already without success. So I am leaving it.
I suggest we just go ahead with this PR as is. Thoughts?

luisramos0:
rebased to resolve conflicts.

app/models/spree/stock_item.rb
sauloperez: why is that?

sauloperez:
This looks ok but I noticed this while reading https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html

> This locking mechanism will function inside a single Ruby process. To make it work across all web requests, the recommended approach is to add lock_version as a hidden field to your form.

We have two workers on our infra so this will be needed.

app/models/spree/stock_item.rb
luisramos0: as far as I can see, this should never happen, if it happens and if it blows up we will see it in bugsnag.

luisramos0:
yeah, this PR is not fixing 5300, just improving the situation.

@sauloperez can you clarify what exactly is preventing you from approving this PR? Is it the question to @mkllnk ?

sauloperez:
I didn't notice this isn't closing #5300 . If this is about introducing optimistic locking this is Test ready IMO.

filipefurtad0:
Hi @kristinalim ,

I tried to reproduce the issue as reported in #5300, but without success. I thank you for the detailed test cases, I went through them:

- **Scenario 1**: Opened two browser sessions, one logged in as admin (Session A), another logged in as customer (Session B).

Session A: I've set the stock level of a product to the limited stock of 4.
Session B: Added 3 items
Session A: Updated the stock level of the product to 2.
Session B: Attempted to check out. This warned of insufficient stock, and redirected to the /cart page, as shown below:

image

Session B: Reduced quantity in cart to 2; While in the /cart page, pressing Update triggered the warning in the red tab appear _again, although the quantity in car was already compatible with the available stock_ -> a "memory-effect" was observed while testing another PR. I will investigate this issue - system retaining previous state and showing previously shown warnings - and report later. <- After pressing Update a second time and making the warning disappear redirection to the checkout page followed.
Session B: Checkout succeed.

- **Scenario 2**: Opened two browser sessions, both logged in as admin.

Session A: I've set the stock level of a product to limited stock, in this case, 4 items.
Session B: Created an order, adding less than total stock left to cart: 3 items.
Session A: Updated the stock level of the product to be less than the quantity in cart: I've set it to 2 items.
Session B: Attempting to complete the order displays the message "Error saving payment":

image

and triggers Bugnsnag:
https://app.bugsnag.com/open-food-france/open-food-france-prod/errors/5f11cd5d9f0dd90017a69e3d?event_id=5f11cd5d005b6e2ca49f0000&i=sk&m=nw

- I have reproduced Scenario 2 with the same values in master-branch: the order is successfully created (Session B) and I verified that negative stock values are allowed:

image

Scenario 2 doesn't seem to be working as expected. Perhaps this PR could be improved? I'm moving it into In Dev for now. The other option would be to merge and open an issue on this concern. Please let me know if you find this beneficial in this case.

RachL:
@filipefurtad0 I don't think Kristina will be able to pick this up as she has no time to contribute at the moment. As it looks to me that it is already improving the situation, without making what we have in prod worse, I would merge and open a dedicated issue.  What do you think?

filipefurtad0:
Thank you your feedback @RachL ,

I agree with you, the situation is improved: as Scenario 2 shows, after this PR it's not possible to get negative stock levels, through backoffice orders - which is the second change introduced by the PR.

I'll open an issue about the error.

Thanks again!

luisramos0:
It's great you can replicate the problem Filipe! I thought we would not be able to replicate it.

I agree with the approach but I think 5784 is a sub-issue of 5300 and thus still an S2 to be worked on. Does that make sense?
If we can replicate 5784 locally I think it will be straight forward to add some defensive code to let the user know the product is out of stock.

filipefurtad0:
Cheers to @kristinalim and the wonderful testing notes! :tada: 
Upgraded to S2, and moved to dev-ready.

Thanks @luisramos0 !

Issue #4572, opened 2019-12-13
[mobile ux] 4. Order cycle design update


daniellemoorhead:
## What is the problem we are solving 
Currently users do not notice the order cycle section as they are not used to ordering in cycles. 

## Success factors = expected outcome 
- [ ] Users start to notice that the order cycle section and therefore there is a decrease in displaying the 'select an order cycle' message

## Design overview
**HOW IT LOOKS NOW**
_Mobile_
<img width="477" alt="Screen Shot 2019-12-13 at 4 24 03 pm" src="https://user-images.githubusercontent.com/10949052/70771385-03ef5200-1dc5-11ea-880f-98c69018b48b.png">

_Tablet_
<img width="1183" alt="Screen Shot 2019-12-13 at 4 23 38 pm" src="https://user-images.githubusercontent.com/10949052/70771402-136e9b00-1dc5-11ea-8d60-ededfcae9e91.png">

_Desktop_
<img width="706" alt="Screen Shot 2019-12-13 at 4 23 50 pm" src="https://user-images.githubusercontent.com/10949052/70771392-09e53300-1dc5-11ea-849d-40ca38fb9833.png">

**HOW IT WILL LOOK IN FUTURE**
_Mobile_
<img width="244" alt="Screen Shot 2019-12-13 at 4 25 48 pm" src="https://user-images.githubusercontent.com/10949052/70771482-5af52700-1dc5-11ea-9226-715c4e45cf73.png">

_Tablet_
<img width="472" alt="Screen Shot 2019-12-13 at 4 26 05 pm" src="https://user-images.githubusercontent.com/10949052/70771487-5e88ae00-1dc5-11ea-952e-582766aca738.png">

_Desktop_
<img width="779" alt="Screen Shot 2019-12-13 at 4 26 19 pm" src="https://user-images.githubusercontent.com/10949052/70771488-621c3500-1dc5-11ea-854f-aabc47f93740.png">

_Design for different states_
![OC Selector](https://user-images.githubusercontent.com/19785399/76047421-9a4c9300-5f30-11ea-893e-09a1e01064dc.png)

Design details are kept in our Zeplin account: https://app.zeplin.io/project/59532764a6c78ff3c534b372/dashboard

daniellemoorhead:
@RachL is there any way we can see that this design fixes the issue shoppers have in understanding that they have to select an order cycle?

RachL:
@daniellemoorhead 

I think so yes.

[...]

PR #5411, opened 2020-05-12
Fix closed shop message not displaying full-width

Matt-Yorkley:
#### What? Why?

This came up in a design sweep, tracked here: https://github.com/openfoodfoundation/openfoodnetwork/issues/4572#issuecomment-626374074

Fixes layout for orders closed message. It now looks like this:

Screenshot from 2020-05-12 18-26-35

Previously the dark grey background was cropped to the inner content width instead of expanding to the full page width.

#### What should we test?

Orders closed messages are fullwidth as intended.

#### Release notes

Adjusted layout for display of orders closed message in shopfront.

### DIFF
--- a/app/views/shop/messages/_closed_shop.html.haml
+++ b/app/views/shop/messages/_closed_shop.html.haml
@@ -1,15 +1,16 @@
-.row.closed-shop-header
-  .small-12.columns
-    .content{ "darker-background" => true }
-      %h4
-        .warning-sign
-          .rectangle
-          %strong !
-        .message
-          = t :shopping_oc_closed
-      %p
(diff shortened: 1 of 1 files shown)


Matt-Yorkley:
Yep, things get that margin at the edge when they're inside `.row` and `.columns`. Sometimes it's not so easy to rearrange though, and you have to move lots of different elements around to "escape" the row...

filipefurtad0:
Hey @Matt-Yorkley 
The orders closed info is now fullwidth as intended by the PR. The About, Producers and Contact are not changed and appear correctly.
The PR is ready to go - thanks!

Issue #5416, opened 2020-05-13
[Very Flacky spec] new Job Logger spec is breaking in lots of branches


luisramos0:
It's works on my computer but most builds are broken with:
```
  1) JobLogger.logger passes the message to the Logger instance
     Failure/Error: expect(job_logger).to have_received(:info).with('log message')
     
       (InstanceDouble(Logger) (anonymous)).info("log message")
           expected: 1 time with arguments: ("log message")
           received: 0 times
     # ./spec/jobs/job_logger_spec.rb:15:in `block (3 levels) in <top (required)>'

```

This was created in #5334

sauloperez:
I'll take a look now that I have it fresh in my mind.

PR #5417, opened 2020-05-13
Fix Job logger spec

luisramos0:
#### What? Why?

Related to #5416

#### What should we test?
green build


sauloperez:
Already solved in https://github.com/openfoodfoundation/openfoodnetwork/pull/5430/

PR #5388, opened 2020-05-07
[Spree 2.1] Fix enterprise package selection submit button

luisramos0:
#### What? Why?

Closes #5360

The target route needs to be patch in rails 4, put wont work. I extended the registration process spec to cover this step.

#### What should we test?
Verify the issue is solved.

#### Release notes
Changelog Category: Fixed
Fixed enterprise package selection submit button

PR #5383, opened 2020-05-07
[Spree 2.1] Fix Update and Calculate Fees button

luisramos0:
#### What? Why?

Closes #5362

Problem with strong params where order object is not sent to the server and is not needed. Strong params validation was expecting it.

#### What should we test?
Verify issue is fixed.
I recommend verifying that the bug only happens with completed orders, not orders in cart.

#### Release notes
Changelog Category: Fixed
Fix bug in strong params implementation in admin orders controller.

PR #5421, opened 2020-05-13
[Spree 2.1] Add missing strong params to less important controllers missed by specs

luisramos0:
#### What? Why?

Fixes a few minor problems in the rails 4 branch:
- cannot edit country name
- cannot add shipping category
- cannot add/edit return authorization to shipped orders

These cases were not covered by specs. I found them by searching for controllers that inherit from resource controller but do not define required params.
I added specs for them now.

#### What should we test?
Verify that you can
- edit country name
- add new shipping category
- add/edit return authorization to shipped orders

### DIFF
--- a/app/controllers/spree/admin/countries_controller.rb
+++ b/app/controllers/spree/admin/countries_controller.rb
@@ -1,6 +1,9 @@
 module Spree
   module Admin
     class CountriesController < ResourceController
+
+      protected
+
       def permitted_resource_params
         params.require(:country).
           permit(:name, :iso_name, :states_required)
--- a/app/controllers/spree/admin/countries_controller.rb
+++ b/app/controllers/spree/admin/countries_controller.rb
@@ -1,6 +1,11 @@
 module Spree
   module Admin
     class CountriesController < ResourceController
+      def permitted_resource_params
+        params.require(:country).
+          permit(:name, :iso_name, :states_required)
+      end
+
       def collection
         super.order(:name)
       end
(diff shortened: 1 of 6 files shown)


Matt-Yorkley:
Some basic test coverage would be good.

luisramos0:
ok, I have added specs :+1:
Back to code review.

app/controllers/spree/admin/shipping_categories_controller.rb
```diff
@@ -1,6 +1,10 @@
 module Spree
   module Admin
     class ShippingCategoriesController < ResourceController
+      def permitted_resource_params
```
Matt-Yorkley, line 4: Are these declarations normally `private` (just for consitency)?

app/controllers/spree/admin/countries_controller.rb
```diff
@@ -1,6 +1,11 @@
 module Spree
   module Admin
     class CountriesController < ResourceController
+      def permitted_resource_params
```
Matt-Yorkley, line 4: Same here.

luisramos0: ah, good point. protected.

luisramos0:
rebased to get latest 3-0-stable before testing.

filipefurtad0:
Ok! I'm ready for this one :+1:

luisramos0:
done, staged :+1:

filipefurtad0:
Hey @luisramos0 
Verified the bugs before, and that now one can:

- edit country name
- add new shipping category
- add/edit return authorization to shipped orders

All good!


PS: I noticed that between staging the two previous PRs #5383 and #5388 the language selector is gone. I'm sorry, I can't narrow it down, only noticed this now:
image

I'm opening an issue on this.

luisramos0:
you dont need to open an issue re language selector. it's a manual fix on uk staging that has probably just gone, I need to put it back.

PR #5427, opened 2020-05-13
Upgrade jquery-ui-rails to v4.2.1

luisramos0:
#### What? Why?

Follow up from #5380
I tried upgrading to 5.0.5 but it didnt work. But we can upgrade from 4.1 to 4.2 :+1:

#### What should we test?
We can try and see if this time #5201 is fixed. 

We should also do a sanity check on admin datepickers and dropdowns with search fields.

#### Release notes
Changelog Category: Changed

Bump jquery-ui-rails to 4.2.1

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -108,7 +108,7 @@ gem "foundation-rails"
 
 gem 'jquery-migrate-rails'
 gem 'jquery-rails', '3.1.5'
-gem 'jquery-ui-rails', '~> 4.1.2'
+gem 'jquery-ui-rails', '~> 4.2'
 gem 'select2-rails', '~> 3.4.7'
 
 gem 'ofn-qz', github: 'openfoodfoundation/ofn-qz', ref: '60da2ae4c44cbb4c8d602f59fb5fff8d0f21db3c'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -425,8 +425,8 @@ GEM
     jquery-rails (3.1.5)
       railties (>= 3.0, < 5.0)
       thor (>= 0.14, < 2.0)
-    jquery-ui-rails (4.1.2)
-      railties (>= 3.1.0)
+    jquery-ui-rails (4.2.1)
+      railties (>= 3.2.16)
     json (1.8.6)
     json_spec (1.1.5)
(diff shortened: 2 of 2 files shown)


Matt-Yorkley:
Can we remove the version pin and let bundler and dependabot figure it out?

luisramos0:
no, that's what I was trying to figure out.
The build is totally broken with 6 and also with 5.0.5 (latest v5) but works with this 4.2.1 (latest v4).
Looks like we should keep this one :+1:

luisramos0:
there's just the job_logger_Spec broken here :+1:

Gemfile
```diff
@@ -108,7 +108,7 @@ gem "foundation-rails"
 
 gem 'jquery-migrate-rails'
 gem 'jquery-rails', '3.1.5'
-gem 'jquery-ui-rails', '~> 4.0.0'
+gem 'jquery-ui-rails', '~> 4.2'
```
Matt-Yorkley, line 111: As with the other PR, shall we make it `4.2.1` here? The `Gemfile.lock` is using `4.2.1` anyway.

luisramos0: We are not stuck at 4.2.1 we are only stuck below v5. I think this is the correct way to say we can take any v4 above 4.2 version but not v5.
I think an upgrade to v4.3 or above would be welcome and will most probably work.
Does this make sense to you or do you prefer 4.2.1. here?

Matt-Yorkley: Ah, I see :okay_hand:

Matt-Yorkley:
I'll kick this into `test ready` with only 1 review, just to keep things moving. It's a very simple dependency version bump, and can be verified easily in testing.

filipefurtad0:
Hey @luisramos0, 

The request

https://staging.katuma.org/assets/jquery-ui/ui-bg_flat_75_ffffff_40x100-8a71ce304ee5d4e3ffc09efecc9b28e8.png

is still a 404:

image

The update to v4.2.1 causes no issues on datepickers and dropdowns with search fields - did a sanity check and they work as before - which is good news!

Moving to ready to go.

luisramos0:
thanks Filipe!

PR #5429, opened 2020-05-14
5400 List each order only once in Orders And Distributors report

### LINKED ISSUE #5400: Repeated entries in Orders and distributors Report
## Description

Orders and distributors report at:

admin/reports/orders_and_distributors

shows more than the expected entries. For example, an order of 4 items shows up in 16 times in the table. If the number of items n in that order is 1, then all is good. So, for n > 1 then a n^n behaviour is observed (pic below).

It's worth noticing this was observed before the latest release v2.9.8, across different staging servers.

## Expected Behavior

The report should show one line per item in that order - right?

## Actual Behaviour

Items appear repeatedly in the report. 

## Steps to Reproduce

1. User - Place an order with n items
2. Admin - Check the report and observe that the number of lines in the table relate to the number of items in the order, in a exponential fashion, n^n, ex.: 4 items = 16 entries.

## Animated Gif/Screenshot

Order #R387418680:

Order R387418680 corresponds to order_id=1017:

## Severity

I am rating this an S3, but thoughts are welcome, as I am not sure on the real impact of this issue.

## Your Environment

* Version used: v2.9.8

## Possible Fix

Not really a fix, but related with update of specs for these reports:
Check and update test coverage of Orders and Distributors report #4746

mkllnk:
#### What? Why?

Closes #5400

In January, an SQL optimisation lead to listing orders multiple times in the Orders And Distributors report. I now de-duplicated the entries.

13633e8 replaced a `where id in (?)` query with `outer join ... where`. The join lead to multiple rows per order. This was okay in most reports because they use that query only to further query line items. But the Orders And Distributors report was iterating over those duplicated orders. A `select distinct` clause corrects this here.

I quickly checked all other uses of the affected `visible_orders` method and it appears that all other usages are fine.

Many reports would benefit from refactoring to reduce the complexity. The code has grown over time and became difficult to maintain.

#### What should we test?

- User - Place an order with more than one item
- Admin - Check the report and observe that the number of lines in the table relate to the number of items in the order.

#### Release notes

Removed duplicate rows in the Orders And Distributors report.

Changelog Category: Fixed

### DIFF
--- a/lib/open_food_network/order_and_distributor_report.rb
+++ b/lib/open_food_network/order_and_distributor_report.rb
@@ -34,7 +34,9 @@ module OpenFoodNetwork
     end
 
     def search
-      @permissions.visible_orders.complete.not_state(:canceled).search(@params[:q])
+      @permissions.visible_orders.select("DISTINCT spree_orders.*").
+        complete.not_state(:canceled).
+        search(@params[:q])
     end
 
     def table
--- a/.rubocop_manual_todo.yml
+++ b/.rubocop_manual_todo.yml
@@ -226,7 +226,6 @@ Layout/LineLength:
   - spec/lib/open_food_network/group_buy_report_spec.rb
   - spec/lib/open_food_network/lettuce_share_report_spec.rb
   - spec/lib/open_food_network/option_value_namer_spec.rb
-  - spec/lib/open_food_network/order_and_distributor_report_spec.rb
   - spec/lib/open_food_network/order_cycle_form_applicator_spec.rb
   - spec/lib/open_food_network/order_cycle_permissions_spec.rb
   - spec/lib/open_food_network/order_grouper_spec.rb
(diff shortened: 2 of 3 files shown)


lib/open_food_network/order_and_distributor_report.rb
```diff
@@ -34,7 +34,9 @@ def header
     end
 
     def search
-      @permissions.visible_orders.complete.not_state(:canceled).search(@params[:q])
+      @permissions.visible_orders.select("DISTINCT spree_orders.*").
+        complete.not_state(:canceled).
+        search(@params[:q])
```
sauloperez, line 39: If we split the line, I'd keep all methods in their own line for consistency and readability, but that's my opinion.

luisramos0: I have been doing that as well, I think it's easier to read.

Matt-Yorkley:
Needs a rebase; there's a failing spec that should be fixed in `master`.

filipefurtad0:
Hi @mkllnk ,

I placed an order with two items (order #R114388831, order_id = 1886), verifying that only two entries show up in the Orders And Distributors report. This fix appears be retroactive as well: previous orders do not appear in duplicate any more :tada: 

Thank you Maikel!

PR #5438, opened 2020-05-15
Cart stock UX

### LINKED ISSUE #5239: Stuck on cart, if item becomes out of stock during shopping journey

Matt-Yorkley:
#### What? Why?

Closes #5239 

Updates various bits of UX to make it clearer to the user where they need to adjust their selected quantities when an item has reduced in stock.

#### What should we test?

UX on cart page when an item's stock has been reduced below the currently selected level.

#### Release notes

Improved cart page UX for carts with insufficient stock

### DIFF
--- a/app/assets/javascripts/darkswarm/controllers/cart_form_controller.js.coffee
+++ b/app/assets/javascripts/darkswarm/controllers/cart_form_controller.js.coffee
@@ -1,0 +1,2 @@
+Darkswarm.controller "CartFormCtrl", ($scope) ->
+
--- a/app/assets/javascripts/darkswarm/directives/disable_dynamically.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/disable_dynamically.js.coffee
@@ -1,0 +1,12 @@
+# Allows disabling of link buttons via disabled attribute.
+# This is normally ignored, ie the link appears disabled but is still clickable.
+
+Darkswarm.directive "disableDynamically", ->
+  restrict: 'A'
+
+  link: (scope, element, attrs) ->
+    element.on 'click', (e) ->
+      if attrs.disabled
+        e.preventDefault()
+      return
+
--- a/app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
@@ -1,0 +1,17 @@
+Darkswarm.directive "validateStockQuantity", ->
+  restrict: 'A'
+  require: "ngModel"
+
+  link: (scope, element, attr, ngModel) ->
+    ngModel.$parsers.push (selectedQuantity) ->
+      if parseInt(selectedQuantity) > scope.available_quantity()
+        ngModel.$setValidity('stock', false);
+      else
+        ngModel.$setValidity('stock', true);
+
+      selectedQuantity
+
+    scope.available_quantity = ->
+      on_hand = parseInt(attr.ofnOnHand)
+      finalized_quantity = parseInt(attr.finalizedquantity) || 0 # finalizedquantity is optional
+      on_hand + finalized_quantity
--- a/app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
@@ -1,6 +1,7 @@
 Darkswarm.directive "validateStockQuantity", (StockQuantity) ->
   restrict: 'A'
   require: "ngModel"
+  scope: true
 
   link: (scope, element, attr, ngModel) ->
     ngModel.$parsers.push (selectedQuantity) ->
--- a/app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
@@ -4,10 +4,10 @@ Darkswarm.directive "validateStockQuantity", ->
 
   link: (scope, element, attr, ngModel) ->
     ngModel.$parsers.push (selectedQuantity) ->
-      if parseInt(selectedQuantity) > scope.available_quantity()
-        ngModel.$setValidity('stock', false);
-      else
-        ngModel.$setValidity('stock', true);
+      valid_number = parseInt(selectedQuantity) != NaN
+      valid_quantity = parseInt(selectedQuantity) <= scope.available_quantity()
+
+      ngModel.$setValidity('stock', (valid_number && valid_quantity) );
 
       selectedQuantity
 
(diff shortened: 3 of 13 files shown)


Matt-Yorkley:
- Quantity box highlighted (red border) in the standard warning red colour
- Red border removed when stock is okay
- Update button is highlighted when: any changes have been made to the form **and** the form has no validation errors
- "Continue Shopping" and "Checkout" buttons disabled unless the cart has been updated and is usable

Cart-Stock-UX

luisramos0:
Hey @Matt-Yorkley can you update the messages as well with a cta of some kind: "Please adjust the quantity and press Update" This can be on the flash and maybe also on the line item message.

app/assets/javascripts/darkswarm/controllers/cart_form_controller.js.coffee
```diff
@@ -0,0 +1,2 @@
+Darkswarm.controller "CartFormCtrl", ($scope) ->
+
```
luisramos0, line 2: ah, this bootstraps angular basics on the form? I didnt know this :ok_hand:

app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
```diff
@@ -0,0 +1,17 @@
+Darkswarm.directive "validateStockQuantity", ->
+  restrict: 'A'
+  require: "ngModel"
+
+  link: (scope, element, attr, ngModel) ->
+    ngModel.$parsers.push (selectedQuantity) ->
+      if parseInt(selectedQuantity) > scope.available_quantity()
+        ngModel.$setValidity('stock', false);
+      else
+        ngModel.$setValidity('stock', true);
```
luisramos0, line 10:
```javascript
      valid? = parseInt(selectedQuantity) > scope.available_quantity()
      ngModel.$setValidity('stock', valid?);
```

luisramos0: re parseInt - what happens if user types a letter in the quantity field?

app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
```diff
@@ -0,0 +1,17 @@
...
+
+      selectedQuantity
+
+    scope.available_quantity = ->
+      on_hand = parseInt(attr.ofnOnHand)
+      finalized_quantity = parseInt(attr.finalizedquantity) || 0 # finalizedquantity is optional
+      on_hand + finalized_quantity
```
luisramos0, line 17: I think this is complicated code already in ofnOnHand, I dont think we can copy paste this type of logic. It needs to be extracted to a service to be re-used here.

app/assets/javascripts/darkswarm/directives/disable_dynamically.js.coffee
```diff
@@ -0,0 +1,13 @@
+# Allows disabling of link buttons via disabled attribute.
+# This is normally ignored, ie the link appears disabled but is still clickable.
+
+Darkswarm.directive "disableDynamically", ->
+  restrict: 'A'
+
+  link: (scope, element, attrs) ->
+    element.on 'click', (e) ->
+      console.log(attrs)
```
luisramos0, line 9: this needs to be deleted right?

app/assets/javascripts/darkswarm/directives/disable_dynamically.js.coffee
```diff
@@ -0,0 +1,13 @@
+# Allows disabling of link buttons via disabled attribute.
+# This is normally ignored, ie the link appears disabled but is still clickable.
```
luisramos0, line 2: are you saying "ng-disabled"" doesnt work in these cases? do you know why? It looks like this directive shouldnt be needed.

app/assets/javascripts/darkswarm/controllers/cart_form_controller.js.coffee
Matt-Yorkley: I think if there's no controller on an element, there's no `$scope` to play with. Initially I thought I would need to add some more code to this controller over the course of the PR, but in the end there was nothing else needed.

app/assets/javascripts/darkswarm/directives/disable_dynamically.js.coffee
Matt-Yorkley: It does not work! It appears to be disabled, but is still clickable. I tried hacking the `href` property instead with a conditional so the link would not function when the conditions are not met, but that didn't work either. :see_no_evil: 

SO discussion here: https://stackoverflow.com/questions/23425254/enable-disable-anchor-tags-using-angularjs

app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
Matt-Yorkley: I've made this more explicit :+1:

Matt-Yorkley: Done.

app/assets/javascripts/darkswarm/directives/validate_stock_quantity.js.coffee
Matt-Yorkley: I've made this more explicit :+1:

app/assets/javascripts/darkswarm/services/stock_quantity.js.coffee
```diff
@@ -0,0 +1,7 @@
+Darkswarm.factory "StockQuantity", ->
```
luisramos0, line 1: nice, thank you!!!

filipefurtad0:
This looks awesome @Matt-Yorkley,

After this PR, upon stock reduction during checkout:
- the warning in the red banner "An item in your cart has become unavailable. Please update the selected quantities." is shown
- the "Qty" field gets highlighted in red
- after reducing the value to an allowed number (in accordance with the modal) the update button get's fully red - can't miss it.
- while in this state, the Continue Shopping and Checkout buttons are greyed-out

This is good to go :tada:

PR #5440, opened 2020-05-15
Fix Grumpy Cat on shop page (cart with items from a closed OC)

### LINKED ISSUE #5340: Grumpy cat (http 422) trying to access the shopfront
## Error in OpenFoodNetwork UK

**ActiveRecord::RecordInvalid** in **enterprises#shop**
Validation failed: Distributor or order cycle cannot supply the products in your cart

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5e760f81ae44a400180bda7d?event_id=5eaaf702005b345d717d0000&i=gh&m=ci)

## Stacktrace

    app/controllers/enterprises_controller.rb:74 - reset_order

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5e760f81ae44a400180bda7d?event_id=5eaaf702005b345d717d0000&i=gh&m=ci)

luisramos0:
#### What? Why?

Closes #5340

This issue was caused by landing on a shop page with some variants in the cart from an Order Cycle that was no longer visible to the user (either closed or some new tag rules). The code is now smart enough to verify the order_cycle is not visible and will empty the order.

Thsi was a simple if statement on the enterprise_controller but sometimes we have to make the world a better place, I mean OFN, so I went on an afternoon of refactoring by extracting new OrderCartRest service (and add some unit tests) and renaming 2 other services according to service naming conventions so now we have related services together:
- OrderCheckoutRestart (restart order workflow)
- OrderCompletionReset (reset order after completion)
- OrderCartReset (reset order on shop page).

It's when we extract these services (and see how complex they are) that we see how obviously overloaded our controllers are!!! We need more of this in OFN.

#### What should we test?
Verify the issue is fixed. When OC is closed and there are items in the cart, the cart will be cleared up and the normal redirection rules will apply:
1- no other OCs: flash message and redirect to shops page
2- if there is only one more OC, that one will be selected and displayed
3- if there are two or more OCs, the red Oc selector will show up.

We should also try to break it by changing things in the Backoffice and load different shops pages #variation #exploratorytesting

Note: now I realize I havent tested with 3 OCs.
#### Release notes
Changelog Category: Fixed
Fixed grumpy cat page (422 http error) when Order Cycle was closed while user was shopping (the broken case was only when there other Order Cycles open for the same shop)

### DIFF
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -97,7 +97,14 @@ class EnterprisesController < BaseController
 
   def reset_order_cycle(order, distributor)
     order_cycles = Shop::OrderCyclesList.new(distributor, current_customer).call
-    order.order_cycle = order_cycles.first if order_cycles.size == 1
+
+    select_first_order_cycle(order, order_cycles)
+  end
+
+  def select_first_order_cycle(order, order_cycles)
+    return unless order_cycles.size == 1
+
+    order.order_cycle = order_cycles.first
   end
 
   def shop_order_cycles
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -98,6 +98,11 @@ class EnterprisesController < BaseController
   def reset_order_cycle(order, distributor)
     order_cycles = Shop::OrderCyclesList.new(distributor, current_customer).call
 
+    if order.order_cycle.present? && !order_cycles.include?(order.order_cycle)
+      order.order_cycle = nil
+      order.empty!
+    end
+
     select_default_order_cycle(order, order_cycles)
   end
 
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -98,11 +98,11 @@ class EnterprisesController < BaseController
   def reset_order_cycle(order, distributor)
     order_cycles = Shop::OrderCyclesList.new(distributor, current_customer).call
 
-    select_first_order_cycle(order, order_cycles)
+    select_default_order_cycle(order, order_cycles)
   end
 
-  def select_first_order_cycle(order, order_cycles)
-    return unless order_cycles.size == 1
+  def select_default_order_cycle(order, order_cycles)
+    return unless order.order_cycle.blank? && order_cycles.size == 1
 
     order.order_cycle = order_cycles.first
   end
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -112,14 +112,6 @@ class EnterprisesController < BaseController
     order.order_cycle = order_cycles.first
   end
 
-  def shop_order_cycles
-    if current_order_cycle
-      [current_order_cycle]
-    else
-      OrderCycle.not_closed.with_distributor(current_distributor)
-    end
-  end
-
   def set_noindex_meta_tag
     @noindex_meta_tag = true unless current_distributor.visible?
   end
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -65,51 +65,10 @@ class EnterprisesController < BaseController
   def reset_order
     order = current_order(true)
 
-    reset_distributor(order, distributor)
-
-    reset_user_and_customer(order) if try_spree_current_user
-
-    reset_order_cycle(order, distributor)
-
-    order.save!
+    OrderCartReset.new(order, params[:id], try_spree_current_user, current_customer).call
   rescue ActiveRecord::RecordNotFound
     flash[:error] = I18n.t(:enterprise_shop_show_error)
     redirect_to shops_path
-  end
-
-  def distributor
-    @distributor ||= Enterprise.is_distributor.find_by_permalink(params[:id]) ||
-                     Enterprise.is_distributor.find(params[:id])
-  end
-
-  def reset_distributor(order, distributor)
-    if order.distributor && order.distributor != distributor
-      order.empty!
-      order.set_order_cycle! nil
-    end
-    order.distributor = distributor
-  end
-
-  def reset_user_and_customer(order)
-    order.associate_user!(spree_current_user) if order.user.blank? || order.email.blank?
-    order.__send__(:associate_customer) if order.customer.nil? # Only associates existing customers
-  end
-
-  def reset_order_cycle(order, distributor)
-    order_cycles = Shop::OrderCyclesList.new(distributor, current_customer).call
-
-    if order.order_cycle.present? && !order_cycles.include?(order.order_cycle)
-      order.order_cycle = nil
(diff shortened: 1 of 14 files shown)


spec/controllers/enterprises_controller_spec.rb
```diff
@@ -112,6 +108,19 @@
...
+      order.order_cycle = order_cycle1
+      order.save
+      order_cycle1.update_attribute :orders_close_at, Time.zone.now
+
+      spree_get :shop, id: distributor
+
+      expect(controller.current_order.distributor).to eq(distributor)
+      expect(controller.current_order.order_cycle).to eq(order_cycle2)
+      expect(controller.current_order.line_items).to be_empty
+    end
```
Matt-Yorkley, line 122: Do you think it's worth adding specs for the cases here? 

https://github.com/openfoodfoundation/openfoodnetwork/compare/master...Matt-Yorkley:enterprise-controller-stock

- Line item variants are all in stock
- Line item variants are not all in stock
- A line item's variant has been soft-deleted
- A line item's variant has been removed from the OC

app/controllers/enterprises_controller.rb
```diff
@@ -65,49 +65,12 @@ def enough_stock?
...
-  end
-
-  def shop_order_cycles
-    if current_order_cycle
-      [current_order_cycle]
-    else
-      OrderCycle.not_closed.with_distributor(current_distributor)
-    end
-  end
-
```
sauloperez, line 110: You'll go to heaven for this :heart_eyes:

app/services/order_completion_reset.rb
```diff
@@ -1,6 +1,8 @@
-# Builds a new order based on the one specified. This implements the "continue
-# shopping" feature once an order is completed.
-class ResetOrderService
+# frozen_string_literal: false
+
+# Resets a completed order by building a new order based on the one specified.
+# This implements the "continue shopping" feature once an order is completed.
```
sauloperez, line 4: pretty useful :ok_hand:

app/controllers/enterprises_controller.rb
luisramos0: :angel:

luisramos0: or :imp:

spec/controllers/enterprises_controller_spec.rb
luisramos0: I was planning to do this as part of this PR but now that Pau moved it to Test Ready with the two approvals, we can probably get this into the release.
I will work on these specs anyway and create a new PR. I am a bit all over 10's of PRs, I may get to this only on Monday.

sauloperez: IMO it's fine as is. The services seem mostly covered with unit tests and we might just need to increase the integration coverage at controller level a little bit.

filipefurtad0:
Hi @luisramos0 

I've checked this bug and the test cases you describe.

In summary, the bug is gone :tada:
However, I can't seem to trigger the warning - which was displayed before the PR, when being redirected to the homepage:
image

The test cases:
Closing OC while on checkout, with a product on cart from that OC:

1- no other OCs: redirect to shops page, but the blue flash message does note appear
image
2- if there is only one more OC, that one will be selected and displayed - OK

3- if there are two or more OCs, the red Oc selector will show up - OK (verified this with 3 open OCs as well)

If two open OCs have the same products, and one of them is closes before checkout, the products on the cart are not automatically allocated to the OC remaining open. Instead, one is redirected to the /shop page, either with remaining OC open (test case 2) or with the selector showing up in red (case 3).

Other tests:
- changing prices to items in cart - all good
- keeping the OC open but removing item - leads to the now fixed stuck on cart issue :tada: 
- adding other items to OC - all good

This is ready to go. It think the warning would be nice to have, but I don't think we should block this, for this reason. I can open a separate issue to add this.

luisramos0:
ok, thanks Filipe. Yes, let's open an issue for that missing message. We need to tackle it now.
I dont understand how it can redirect but not show the warning: the two things are just one after the other in the code... maybe some problem displaying the flash message?

filipefurtad0:
Hi @luisramos0 ,

I checked this PR again, and found a small but relevant imprecision in my notes above (confusion between shops /shops /shop). In test case 1, the user gets redirected to the /shop page, as the pic shows (not confuse with /shops - sorry for that).

To see which flash message to expect in these test cases, I had a look in the code. So, should it be this error:

a) "The shop you are looking for doesn't exist or is inactive on OFN. Please check other shops."

or this info? 

b) "The order cycle you've selected has just closed. Please try again!"

I could trigger both these flash messages in my system, but only when visiting or getting redirected to the /shops page. This can happen if the user is in the checkout menu with items in cart, and:

a) the shop closes (by changing the package from shop/hub to producer profile) and the user attempts to edit is cart (navigating back with the browser): 

image

b) the OC closes on checkout page and the user clicks Boutiques:

image


So, both of these flash messages appear in the /shops# page. In the covered test cases the user rather gets redirected to the /shop page (not /shops or shops), so could this be the reason why I am not observing any the flash messages?

Should this be dealt in a separate issue (should I open one)?


PS - I found other issues which I don't think are related to this PR (opening issues on these):

- Bug switching between different shops, which can be observed by the disagreement between the address bar and the actual page the user finds himself in (gif):

switching_between_shops


- Missing translation for the flash error b) - it seems not to exist in the en.yml

luisramos0:
Hey Filipe, thanks for the detailed update!

I understand the problem, the user has a cart and is shopping, the manager closes the OC in the backoffice (there are no other OCs), user loads the shopfront and the cart is empty and the closed shop message displays. The closed shop message is already evidently saying the cause of the empty cart. I think we are ok here.

So we would change the test case and leave it as is:
Instead of:
"no other OCs: flash message and redirect to shops page"
We have:
"no other OCs: stay on home page where the closed shop message is displayed"


I think you did find a bug though: the missing translation of that flash message in b) :+1:

luisramos0:
re that gif... @Matt-Yorkley will be happy to use his new catchphrase about caching issues!
Can you have a look at this last gif Matt?

If anyone sees that one again I think we should open an issue for it.

luisramos0:
actually, I can replicate this problem in staging FR.
You just have to access the shop with a direct link I think.
click here: https://staging.openfoodfrance.org/fricando-hub/shop
and then click here: https://staging.openfoodfrance.org/chop-suey-farm/shop
Look at the pages...

luisramos0:
I cant replicate this in prod. Maybe a problem with the release?

luisramos0:
I can replicate this in staging FR by clicking around actually :see_no_evil: 
- click /shops in main menu
- click fricando shop
- click /shops  in main menu
- click chop seuy shop (I see Fricando shop :see_no_evil: )

Issue #5433, opened 2020-05-15
Release v2.9.9 Filter Coffee


mkllnk:
Steps: 

- [x] Include translations
- [x] [Draft new release]
- [x] Notify #instance-managers of user-facing changes.
- [ ] Test: https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/2-9-9/builds/2
- [x] Publish and notify #global-community
- [x] Deploy and notify #instance-managers
- [x] Nudge next release manager

The full process is described at https://github.com/openfoodfoundation/openfoodnetwork/wiki/Releasing.

[Draft new release]: https://github.com/openfoodfoundation/openfoodnetwork/releases

lin-d-hop:
https://docs.google.com/document/d/1rXLmg7q-wBwnYj7FO6J1KnG172g86Y5k3GPZ5WoBoXA/edit#

Ready to go team. Lovely lovely.

filipefurtad0:
Hey @mkllnk and @lin-d-hop ,
I've stumbled across a loss in functionality in the OC selector. Basically, the clickable area of the OC selected is now significantly reduced, which makes the selector appear as not working. In staging, only the green area (pic below) is possible to activate, to see the drop-down list of open OCs:

![image](https://user-images.githubusercontent.com/49817236/82200671-3878b400-98f7-11ea-900c-7a679173dcd2.png)

In production, it's easy to observe that the whole white area is clickable.

I've double checked on different server/browser combination, but could you please confirm these observations @lin-d-hop ?

luisramos0:
I can replicate, I am trying to find what caused this problem.
Looks like the mobile search filters PR #5330 is the only mobile PR in this release. It's quite a big PR, I am trying to understand what commit caused the problem.

luisramos0:
ok, this is caused by #5330
I have found the commit that introduced this problem. it's the second commit in #5330
https://github.com/openfoodfoundation/openfoodnetwork/pull/5330/commits/f46ca0c5956084bd5a028d8f580c8fc7587941ff

luisramos0:
[this line](https://github.com/openfoodfoundation/openfoodnetwork/pull/5330/commits/f46ca0c5956084bd5a028d8f580c8fc7587941ff#diff-5adb00cccb8e23d76783b6c406af4714R12) moves the z-index of the tab buttons on the left over the OC selector making the selecor clickable above the tabs line:
[...]

PR #5451, opened 2020-05-18
Fix tab-buttons element overlapping order cycle selector

Matt-Yorkley:
#### What? Why?

Quick fix for UX bug recorded here: https://github.com/openfoodfoundation/openfoodnetwork/issues/5433#issuecomment-630089588

This is a quick fix for the release, and means the drop-shadow isn't shown nicely over the new search bar. I think we can address that for the next release.


luisramos0:
I moved this commit to the branch 2.9.9 in another PR.

luisramos0:
re-opening so we can merge this change to master as well.

luisramos0:
I created an S5 to represent the lack of drop shadow after this PR:
https://github.com/openfoodfoundation/openfoodnetwork/issues/5477

PR #5452, opened 2020-05-18
Fix new order translation

### LINKED ISSUE #5094: Missing/wrong translation key in admin / new order
## Description

When creating a new order as an admin, the word is not correct in French, so I suspect that we reuse another key here instead of having a dedicate one.

## Steps to Reproduce
1. Go to `/admin/orders/new`
2. The translation key for the title is not correct

## Animated Gif/Screenshot

image

## Severity

## Your Environment
* Version used: v2.8

sharonfruit:
#### What? Why?

Closes #5094
"New Order" was mis-translated because "new" was being appended to "order" instead of using the "new order" key. Added logic to use "new order" key on a new order.

#### What should we test?
Load the new order page.

#### Release notes
Fix for new order translation.

Changelog Category: Fixed

### DIFF
--- a/app/views/spree/admin/orders/new.html.haml
+++ b/app/views/spree/admin/orders/new.html.haml
@@ -1,13 +1,10 @@
-- content_for :page_title do
-  = t(:new)
-
 - content_for :page_actions do
   %li= button_link_to t(:back_to_orders_list), spree.admin_orders_path, :icon => 'icon-arrow-left'
 
 = admin_inject_shops(module: 'admin.orders')
 = admin_inject_order_cycles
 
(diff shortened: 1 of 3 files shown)


sharonfruit:
@luisramos0 Your proposal makes sense. This work item was re-assigned to @rmklaus12 because I'm not able to execute this in a timely fashion (I'm a bit overloaded at the moment at work). I made a pull request with what I had so that they could modify what I had. My Docker image doesn't run at the moment and I haven't had time to delve into it.

@rmklaus12 can you take over?

sharonfruit:
Closing . . . refer to PR 5482

PR #5453, opened 2020-05-18
Fix tab-buttons element overlapping order cycle selector

luisramos0:
#### What? Why?

Same as #5451 but on 2.9.9 patch.

This is a quick fix to a release blocking issue #5433

#### What should we test?
Verify the issue from 5433 is fixed.

#### Release notes
Changelog Category: Fixed
Fix issue with clickable area on the Order Cycle selector.


Matt-Yorkley:
@luisramos0 this commit isn't in `master`, is it?

luisramos0:
good point! we need to put this in master!
Maybe we can point your original PR to master and merge it there.

luisramos0:
like this: https://github.com/openfoodfoundation/openfoodnetwork/pull/5451

PR #5456, opened 2020-05-18
[Spree 2.1] Move all spree_core dependencies to OFN so we can upgrade them later

luisramos0:
#### What? Why?

I did this trick to remove deface from spree_core and get nokogiri upgraded on our side.
I thought that now that we are basically settled on the ideia that spree 2-1-0 will be the last spree version ofn uses (anyone has second thoughts about this? maybe it is a good time to discuss) we could do it for all dependencies of spree core. I relaxed them a bit in spree here (to allow rails 4.1 upgrade soon) https://github.com/openfoodfoundation/spree/pull/45 and added them to OFN in this PR.

This will enable us to upgrade some of these immediately after this PR.

#### What should we test?
There are no actual version changes in this PR, just the requirements were moved around so a green build and a simple sanity check will be enough.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,20 @@ gem 'pg', '~> 0.21.0'
 # for details.
 gem 'spree_core', github: 'openfoodfoundation/spree', branch: '2-1-0-stable'
 
+### Dependencies brought from spree core
+gem 'acts_as_list', '= 0.2.0'
+gem 'awesome_nested_set', '~> 3.0.0.rc.1'
+gem 'cancan', '~> 1.6.10'
+gem 'ffaker', '~> 1.16'
+gem 'highline', '= 1.6.18' # Necessary for the install generator
+gem 'httparty', '~> 0.11' # For checking alerts.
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -20,7 +20,21 @@ gem 'pg', '~> 0.21.0'
 # OFN-maintained and patched version of Spree v2.0.4. See
 # https://github.com/openfoodfoundation/openfoodnetwork/wiki/Tech-Doc:-OFN's-Spree-fork%F0%9F%8D%B4
 # for details.
-gem 'spree_core', github: 'openfoodfoundation/spree', branch: '2-1-0-stable'
+gem 'spree_core', github: 'luisramos0/spree', branch: '2-1-0-without-deps'
```
luisramos0, line 23: this will have to be updated once the spree PR is merged.

luisramos0:
This is not blocked and is ready for review.

luisramos0:
done :+1:

PR #5457, opened 2020-05-19
Fix cart fields sharing same scope

### LINKED ISSUE #5455: Stock levels not taking the expected effect on /cart page
## Description

The stock levels for different products set incorrect limits in the cart page. When increasing the amount of a product with existing stock, the  modal for insufficient stock is triggered once the stock of one of the items' amounts is exceeded. This happens with inventory set take the settings from the producer. See pics/example below.

## Expected Behavior

It should be possible to increase the amount of a given stock item, if the number of items does not exceed the settings in the backoffice.

## Actual Behaviour

The adjustment of the number of items in the cart page is incorrectly limited - apparently the stock levels from other products are considered.

## Steps to Reproduce

1. Set stock values for two different products A and B, for example 20 an 100 stock levels, respectively
2. As a customer and while in the shopfront, insert the 20 units of each on the cart. Proceed to cart.
3. In the /cart page, try to increase the amount of items for product B.
4. The modal will appear, stating the level of 20 cannot be exceeded, while in fact still 80 items are left

Please see pictures of this example below.

## Animated Gif/Screenshot

Right side - admin page, Left side - customer

Inventory is set to take the settings from the producer:

## Workaround

The workaround is not to use the /cart page to introduce changes to the cart.

## Severity

## Your Environment

* Version used: PR for release  2.9.9
* Browser name and version: Firefox 2.9.9

Matt-Yorkley:
#### What? Why?

Closes #5455

All cart page quantity fields were displaying a single max quantity instead of a different value for each one, because the directive for capping stock on the fields didn't have individual scopes on each one.

#### What should we test?

Maximum quantities on cart page should be accurate and not weirdly duplicated with the same repeated value. So if there are three products in the cart with stock levels of 2, 3 and 4, the maximum stock values should be 2, 3, and 4 and not 2, 2, 2 or 4, 4, 4 for example.

#### Release notes

Fixed maximum quantity issue on cart page.

### DIFF
--- a/app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
+++ b/app/assets/javascripts/darkswarm/directives/on_hand.js.coffee
@@ -1,6 +1,7 @@
 Darkswarm.directive "ofnOnHand", ->
   restrict: 'A'
   require: "ngModel"
+  scope: true
 
   link: (scope, elem, attr, ngModel) ->
     # In cases where this field gets its value from the HTML element rather than the model,
--- a/spec/features/consumer/shopping/cart_spec.rb
+++ b/spec/features/consumer/shopping/cart_spec.rb
@@ -176,7 +176,7 @@ feature "full-page cart", js: true do
       end
 
       describe "with insufficient stock available" do
-        xit "prevents user from entering invalid values" do
+        it "prevents user from entering invalid values" do
           add_product_to_cart order, product_with_fee
 
           variant.update_attributes!(on_hand: 2, on_demand: false)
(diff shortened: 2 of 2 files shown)


Matt-Yorkley:
Okay, this needs a feature spec to test the regression, but basically this is the fix. As far as I can tell this bug is not new.

luisramos0:
ok, I am moving this to Test Ready so we can include in the release. shall we put this against 2.9.9 branch @Matt-Yorkley ?

filipefurtad0:
Hi @Matt-Yorkley,

The fix works as exoected. The stock levels set in the backoffice now work independently, for each item, in the /cart. The changes in the /cart page are successfully updated and it is possible to perform checkout with those set amounts.

Ready to go - Thanks for the fix!

PR #5458, opened 2020-05-19
Fix display issue on order cycle edit form

Matt-Yorkley:
#### What? Why?

Some of the form elements on the first OC edit page were covered up by the save bar in some cases. This ensures they are always visible and clickable.

#### What should we test?

Order cycle (admin) edit pages: the buttons at the bottom of forms are not covered up by the save bar. See screenshots below; the "Add Coordinator Fee" button is not readable or clickable.

**BEFORE:**

Screenshot from 2020-05-19 11-54-53

**AFTER:**

Screenshot from 2020-05-19 11-54-42

#### Release notes

Fixed layout issue on admin order cycle edit pages.

### DIFF
--- a/app/assets/stylesheets/admin/order_cycles.scss
+++ b/app/assets/stylesheets/admin/order_cycles.scss
@@ -1,5 +1,9 @@
 @import "variables";
 
+form.order_cycle {
+  margin-bottom: 5em;
+}
+
 #schedule-dialog {
   table {
     border: none;


filipefurtad0:
Hi @Matt-Yorkley,

I can reproduce your observations on the staging environment, on my desktop:
Before this PR:
image

After this PR:
image

But this is specially observable in mobile, tested this on mobile as well - this PR fixes this nicely for both platforms :tada: 

Ready to go!

Issue #5459, opened 2020-05-19
[Spree 2.1] Issue creating schedules in tests


luisramos0:
## Description
Many [broken tests in the build](https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/3-0-stable/builds/81), I think they all relate to this schedule_id null in order_cycles_schedules table:
![image](https://user-images.githubusercontent.com/1640378/82315180-d8e7da80-99c2-11ea-80ff-70a408201cd4.png)

I believe this will be related to https://github.com/openfoodfoundation/openfoodnetwork/pull/5332

luisramos0:
I googled this a bit but I cant fix it :-(
ping @sauloperez, any idea?

Matt-Yorkley:
So the join table `order_cycle_schedules` requires both `order_cycle_id` and `schedule_id` to not be `null`, but the suite is trying to create/save them without those attributes? Maybe it needs a tweak in some of the related factories?

Matt-Yorkley:
A schedule is being created with a collection of order cycles, but on save; the schedule itself is what's missing an id. Didn't we encounter some problems like this in the upgrade before, around persistence of associated collections? It sounds familiar.

luisramos0:
good point Matt, maybe we need to change the schedule factories to take OCs as a transient and then only assign them on a after block, that way the schedule will already be created and have an id.

Matt-Yorkley:
We had something similar with variants, right? And enterprise users?

luisramos0:
yes, this was done here
https://github.com/openfoodfoundation/openfoodnetwork/commit/bd0245229b731a250ff148b5174505ab4a3b1e52#diff-68c83872d66e936451119a72143b1acc

here
https://github.com/openfoodfoundation/openfoodnetwork/commit/1dc7b5793c4a323b1e79c862698974469db53c06#diff-68c83872d66e936451119a72143b1acc
and here
https://github.com/openfoodfoundation/openfoodnetwork/commit/e64d21d81dfd967fa9ad0ca6c2c93d17e6beec9f#diff-2222b289dbf415a1e772247a402dd6fd

PR #5468, opened 2020-05-19
Revert 5332 "Add timestamps to OC schedule join table" in the rails 4 branch

luisramos0:
#### What? Why?
This reverts commit 1903134e139e495442b1f630d1a3d8ff083a7435 the single commit in PR #5332
This resolves the problem in #5459 but 5459 goal still needs to be addressed in the rails 4 branch.

#### What should we test?
A better build than before!


Matt-Yorkley:
What was the error here? Do we have an issue for this in the upgrade?

luisramos0:
yes #5459

luisramos0:
we are fixing the problems in #5487 instead of reverting. closing this.

PR #5476, opened 2020-05-20
Add endpoint for api user to retrieve info about itself, mainly to fe…

luisramos0:
…tch their own api key

#### What? Why?

This would give the user something like this, no need to call the admin:
image

#### Release notes


luisramos0:
the api key can be fetched in the html of some pages, see here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/API-documentation#authorization

this Pr is not really needed.

PR #5484, opened 2020-05-22
Added translation keys to error messages partial and en.yml file to a…

### LINKED ISSUE #5336: Missing translation key in error message
## Description
When adding a product, the error message appearing when you forget some mandatory fields is partly untranslated.

## Steps to Reproduce
1. Go to `admin/products`
2. Try to add a product without a producer
3. See error message with 2 sentences untranslated

## Severity

## Your Environment
* Version used: 2.9.6

rmklaus12:
…dd translation to error messages

#### What? Why?

Closes #5336 
Error message on the new products page was not fully translating to French. Additional translation keys needed to be added. 

#### What should we test?
Go to admin/products
Try to add a product with empty required fields
Verify first two lines of error message is fully translated. 

#### Release notes
Changed syntax in `_error_messages.html.haml` to normal t() syntax (was Spree.t())
Added translation keys to en.yml file as:
en.spree.errors.messages.errors_prohibited_this_record_from_being_saved
en.spree.errors.messages.there_were_problems_with_the_following_fields

Changelog Category: Fixed


config/locales/en.yml
```diff
@@ -3091,6 +3091,8 @@ See the %{link} to find out more about %{sitename}'s features and to start using
       update: "Update"
     errors:
       messages:
+        errors_prohibited_this_record_from_being_saved: "errors prohibited this record from being saved"
```
luisramos0, line 3094: This is incorrect, there's a count so the structure has to be different here.
When you bring translations from spree you should look at what was the spree translation here:
https://github.com/openfoodfoundation/spree/blob/2-0-4-stable/core/config/locales/en.yml

Please make sure you test for the 2 cases (singular and plural).

This key is not in the correct place, if you remove Spree., the keys have to be moved out of the spree: namespace in the en.yml file. This can go to errors.messages in line 260 in en.yml file

rmklaus12: Thank you @luisramos0 for your comments. I see now that these need to move out of the spree: namespace. To test for singular and plural, do I need to just add both string variations? Or is there another method?
Also, the area of line 260 does not seem to be relevant to errors, or is that okay as long as the keys are out of the Spree: namespace?
Pardon all the questions and thank you for your help!

luisramos0: you have to add the translations as they are in spree and then you can test it by changing the count of the data, in this case, have a situation where target.errors.count is > 1
Make sure you are able to test this locally. For example, editing a product and having errors on two fields at the same time?

I dont know what I meant by that line 260 :see_no_evil: 
I think we can use lazy lookup here.
t(".errors_prohibited_this_record_from_being_saved")
and then:
```diff
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3089,6 +3089,11 @@ See the %{link} to find out more about %{sitename}'s features and to start using
 
     actions:
       update: "Update"
+    shared:
+      error_messages:
+        errors_prohibited_this_record_from_being_saved:
+          one: "1 error prohibited this record from being saved"
+          other: ! "%{count} errors prohibited this record from being saved"
     errors:
       messages:
         blank: "can't be blank"

```

rmklaus12: Screen Shot 2020-05-29 at 1 10 26 pm
Screen Shot 2020-05-29 at 1 10 44 pm

rmklaus12: Testing looks good for me locally - see screenshots for two versions (plural and singular). 
Added count method value for error count and one / other keys in en.yml to satisfy both conditions (plural and singular)

app/views/spree/shared/_error_messages.html.haml
```diff
@@ -1,11 +1,10 @@
...
-      \:
+      =  t("errors_prohibited_this_record_from_being_saved", count: target.errors.count)
     %p
-      = Spree.t(:there_were_problems_with_the_following_fields)
+      = t("there_were_problems_with_the_following_fields")
       \:
     %ul
       - target.errors.full_messages.each do |msg|
-        %li= msg
+        %li= msg
```
sauloperez, line 10: we're missing a newline at the end of the file :warning:

app/views/spree/shared/_error_messages.html.haml
```diff
@@ -1,11 +1,10 @@
 - if target && target.errors.any?
   #errorExplanation.errorExplanation
     %h2
-      = Spree.t(:errors_prohibited_this_record_from_being_saved, count: target.errors.count)
-      \:
+      =  t("errors_prohibited_this_record_from_being_saved", count: target.errors.count)
```
sauloperez, line 4: I agree with @luisramos0 's comment:

> I think we can use lazy lookup here.
> t(".errors_prohibited_this_record_from_being_saved")

Docs: https://guides.rubyonrails.org/i18n.html#lazy-lookup

luisramos0: @rmklaus12 watch out for the validity of your translations tests (I never use the expected translation because it can be mixed up with translations coming from spree which is what is happening in your test, I always add a custom number to the end of the string), in this case, your call to t is not using lazy loading but your keys are correctly placed for the lazyloading. You need a dot here for your new translations to be picked up.

app/views/spree/shared/_error_messages.html.haml
rmklaus12: Added extra line

app/views/spree/shared/_error_messages.html.haml
rmklaus12: Ah yes, thank you. Tested with extra characters and working with the dot in place now.

rmklaus12:
@luisramos0 - happening again! I, however, do not have the option to rerun the build. But would very much like to. Thanks.

PR #5487, opened 2020-05-22
[Spree 2.1] Allow to create order_cycle_schedules

### LINKED ISSUE #5459: [Spree 2.1] Issue creating schedules in tests

sauloperez:
#### What? Why?

Closes #5459

There's no way we can create an order_cycle_schedules if the schedule doesn't have an id, which we can't get if to persist it we need an OC first, which in turn, will create an order_cycle_schedules.

#### What should we test?

Nothing. Tests related to this issue should pass

### DIFF
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -30,20 +30,20 @@ module Admin
 
     def create
       if params[:order_cycle_ids].blank?
-        return respond_with(@object)
+        return respond_with(@schedule)
       end
 
-      @object.attributes = permitted_resource_params
-      @object.save!
+      @schedule.attributes = permitted_resource_params
+      @schedule.save!
 
-      @object.order_cycle_ids = params[:order_cycle_ids]
-      if @object.save
+      @schedule.order_cycle_ids = params[:order_cycle_ids]
+      if @schedule.save
         sync_subscriptions_create
 
-        flash[:success] = flash_message_for(@object, :successfully_created)
-        respond_with(@object)
+        flash[:success] = flash_message_for(@schedule, :successfully_created)
+        respond_with(@schedule)
       else
-        respond_with(@object)
+        respond_with(@schedule)
       end
     end
 
@@ -96,7 +96,7 @@ module Admin
       result = editable_order_cycles(requested)
 
       params[:schedule][:order_cycle_ids] = result
-      @object.order_cycle_ids = result
+      @schedule.order_cycle_ids = result
     end
 
     def editable_order_cycles(requested)
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -7,7 +7,7 @@ module Admin
     before_filter :check_editable_order_cycle_ids_create, only: [:create]
     before_filter :check_editable_order_cycle_ids_update, only: [:update]
     before_filter :check_dependent_subscriptions, only: [:destroy]
-    update.after :sync_subscriptions
+    update.after :sync_subscriptions_update
 
     respond_to :json
 
@@ -125,21 +125,19 @@ module Admin
       @permissions = OpenFoodNetwork::Permissions.new(spree_current_user)
     end
 
-    def sync_subscriptions
+    def sync_subscriptions_update
       return unless params[:schedule][:order_cycle_ids]
 
-      removed_ids = @existing_order_cycle_ids - @schedule.order_cycle_ids
-      new_ids = @schedule.order_cycle_ids - @existing_order_cycle_ids
-      return unless removed_ids.any? || new_ids.any?
-
-      subscriptions = Subscription.where(schedule_id: @schedule)
-      syncer = OrderManagement::Subscriptions::ProxyOrderSyncer.new(subscriptions)
-      syncer.sync!
+      sync_subscriptions
     end
 
     def sync_subscriptions_create
       return unless params[:order_cycle_ids]
 
+      sync_subscriptions
+    end
+
+    def sync_subscriptions
       removed_ids = @existing_order_cycle_ids - @schedule.order_cycle_ids
       new_ids = @schedule.order_cycle_ids - @existing_order_cycle_ids
       return unless removed_ids.any? || new_ids.any?
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -5,7 +5,7 @@ module Admin
   class SchedulesController < ResourceController
     before_filter :adapt_params, only: [:update]
     before_filter :check_editable_order_cycle_ids_create, only: [:create]
-    before_filter :check_editable_order_cycle_ids, only: [:update]
+    before_filter :check_editable_order_cycle_ids_update, only: [:update]
     before_filter :check_dependent_subscriptions, only: [:destroy]
     update.after :sync_subscriptions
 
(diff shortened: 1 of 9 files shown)


luisramos0:
hey Pau, that looks good! Your PR from master was reverted, so I guess you want to had it here plus the fix.

Matt-Yorkley:
All green :heavy_check_mark:

Matt-Yorkley:
The PR for reverting it hasn't been merged to `3-0-stable` yet, right? #5468

sauloperez:
@luisramos0 can we close #5468? This makes the entire proxy_order_syncer_spec pass and it's only missing two specs from spec/features/admin/schedules_spec.rb.

luisramos0:
ah, yes, sorry, my mistake. closed the revert PR now.
only some schedules specs missing now :muscle: 

and then some other stuff in ShopsController, and some DFC provider problems. these will be for different PRs.

Matt-Yorkley:
Ok, so lets move ahead with this and make issues for the remaining DFC and ShopsController bits.

engines/order_management/spec/services/order_management/subscriptions/proxy_order_syncer_spec.rb
```diff
@@ -16,107 +18,131 @@ module Subscriptions
...
+
+            let(:schedule) { create(:schedule, order_cycles: [oc]) }
+            let(:subscription) do
+              build(
+                :subscription,
+                begins_at: now + 1.minute,
+                ends_at: now + 2.minutes,
+                schedule: schedule
+              )
+            end
```
luisramos0, line 112: I wonder why this is needed if it's already in line 53 above.

sauloperez: It can probably be DRYed a bit but it comes at a cost in tests. There were so many tangled dependencies that I had to duplicate all `lets` to remove all dependencies and then fix test by test. In summary, it was DRY but at the expense of changeability.

sauloperez:
Last commit fixes the issue. It doesn't happen when updating because there is already a schedule_id at that point.

app/models/schedule.rb
```diff
@@ -3,9 +3,8 @@ class Schedule < ActiveRecord::Base
 
   has_many :order_cycles, through: :order_cycle_schedules
   has_many :order_cycle_schedules, dependent: :destroy
-  has_many :coordinators, -> { uniq }, through: :order_cycles
 
-  validates :order_cycles, presence: true
+  has_many :coordinators, -> { uniq }, through: :order_cycles
```
luisramos0, line 7: I wonder why is GH DIFF failing here?  Is there some special hidden character here? This is not a new line, the validates line was just removed and an empty line was added above this one, right?

app/controllers/admin/schedules_controller.rb
```diff
@@ -28,6 +28,25 @@ def index
       end
     end
 
+    def create
+      invoke_callbacks(:create, :before)
```
luisramos0, line 32: can you please delete this callback code? I am pretty sure this is not used.
3 less lines of code in this method.

app/controllers/admin/schedules_controller.rb
```diff
@@ -28,6 +28,25 @@ def index
...
+      @object.attributes = permitted_resource_params
+      @object.save!
+
+      @object.order_cycle_ids = params[:order_cycle_ids]
+      if @object.save
+        invoke_callbacks(:create, :after)
+        flash[:success] = flash_message_for(@object, :successfully_created)
+        respond_with(@object) do |format|
+          format.html { redirect_to location_after_save }
+          format.js   { render layout: false }
```
luisramos0, line 42: there's only one way to create a schedule in the OC page. I think you should delete the format.html here.
This is the new schedule request accept:
image

engines/order_management/spec/services/order_management/subscriptions/proxy_order_syncer_spec.rb
```diff
@@ -210,7 +228,19 @@ module Subscriptions
...
+                    schedule = create(:schedule, order_cycles: [oc])
+                    subscription = build(
+                      :subscription,
+                      begins_at: now + 1.minute,
+                      ends_at: now + 2.minutes,
+                      schedule: schedule
+                    )
+
+                    syncer = ProxyOrderSyncer.new(subscription)
+                    syncer.sync!
```
luisramos0, line 242: I am sorry I don't understand this copy pasted code.
For example here, under the context "and the proxy order has not already been placed" there are 7 specs, 3 of them need this exact same setup, why not group these 3 in a new sub-context, give it some nice name, and have this code once in a before block?

app/controllers/admin/schedules_controller.rb
```diff
@@ -3,7 +3,7 @@
 
 module Admin
   class SchedulesController < ResourceController
-    before_filter :adapt_params, only: [:create, :update]
+    before_filter :adapt_params, only: [:update]
```
luisramos0, line 6: I wonder if we should just copy the update method from resource controller as well, this line would be removed and it would become:

```ruby
      def update
        adapt_params
        if @object.update_attributes(permitted_resource_params)
          flash[:success] = flash_message_for(@object, :successfully_updated)
          respond_with(@object) do |format|
            format.js   { render layout: false }
          end
        else
          respond_with(@object)
        end
      end
```

and if we bring the destroy method like this:
```ruby
      def destroy
        if @object.destroy
          flash[:success] = flash_message_for(@object, :successfully_removed)
          respond_with(@object) do |format|
            format.js   { render partial: "spree/admin/shared/destroy" }
          end
        else
          respond_with(@object) do |format|
            format.html { redirect_to collection_url }
          end
        end
      end
```

and then :tada: 
class SchedulesController < Spree::Admin::BaseController

It's for another PR I think.

sauloperez:
> Did you forget to make this a real PR (not draft)?

not really, I'm just doing the cleanup you suggest right now :joy: . I'm aiming to un-abstract this unnecessary stuff but there might be more stuff I didn't foresee with `sync_subscription`.

app/controllers/admin/schedules_controller.rb
sauloperez: Yes, the more `before_filter` and callbacks I remove is becoming evident that the inheritance doesn't provide much. Indeed, for an upcoming PR.

app/models/schedule.rb
sauloperez: yes, that's it. I think git is correct here since this was at line 6 and now is at line 7. I'm not saying it's easy, it was hard for my brain too :joy:

sauloperez:
Well, this is as far as I can go cleaning up the controller without you having headache while reviewing.

sauloperez:
Oh Gosh, one missing OC spec :see_no_evil:

sigmundpetersen:
If it's the ./spec/features/admin/order_cycles_spec.rb:738 one I think that is flaky, and also in master 👍

app/controllers/admin/schedules_controller.rb
```diff
@@ -57,17 +77,38 @@ def adapt_params
...
-    def check_editable_order_cycle_ids
+    def check_editable_order_cycle_ids_create
+      return unless params[:order_cycle_ids]
+
+      requested = params[:order_cycle_ids]
+      @existing_order_cycle_ids = []
+
+      result = editable_order_cycles(requested)
+
+      params[:order_cycle_ids] = result
```
luisramos0, line 88: I'd change the method name, it's not just checking, it's changing them. I am not sure what a better name would be.

What about this version?
```ruby
  def define_editable_order_cycle_ids_for_create
      return unless params[:order_cycle_ids]

      @existing_order_cycle_ids = []
      params[:order_cycle_ids] = editable_order_cycles(params[:order_cycle_ids])
  end
```

app/controllers/admin/schedules_controller.rb
```diff
@@ -57,17 +77,38 @@ def adapt_params
...
       requested = params[:schedule][:order_cycle_ids]
-      @existing_order_cycle_ids = @schedule.persisted? ? @schedule.order_cycle_ids : []
-      permitted = OrderCycle.where(id: params[:schedule][:order_cycle_ids] | @existing_order_cycle_ids).merge(OrderCycle.managed_by(spree_current_user)).pluck(:id)
+      @existing_order_cycle_ids = @schedule.order_cycle_ids
+
+      result = editable_order_cycles(requested)
+
+      params[:schedule][:order_cycle_ids] = result
+      @schedule.order_cycle_ids = result
+    end
```
luisramos0, line 101: I'd maybe change the method name and remove local variable assignments, they dont add much do they?

```ruby
    def define_editable_order_cycle_ids_for_update
      return unless params[:schedule][:order_cycle_ids]

      @existing_order_cycle_ids = @schedule.order_cycle_ids
      params[:schedule][:order_cycle_ids] = editable_order_cycles(params[:schedule][:order_cycle_ids])

      @schedule.order_cycle_ids = result
    end
```

app/controllers/admin/schedules_controller.rb
```diff
@@ -82,9 +123,19 @@ def permissions
...
     end
 
-    def sync_subscriptions
+    def sync_subscriptions_update
       return unless params[:schedule][:order_cycle_ids]
 
+      sync_subscriptions
+    end
+
+    def sync_subscriptions_create
```
luisramos0, line 132: I think this looks better with the for sync_subscriptions_for_create
same for update

app/controllers/admin/schedules_controller.rb
```diff
@@ -30,12 +30,19 @@ def index
 
     def create
       invoke_callbacks(:create, :before)
+
+      if params[:order_cycle_ids].blank?
```
luisramos0, line 32: is this new behavior? we dont create a schedule without OCs.

luisramos0:
Good effort Pau, thanks!

app/controllers/admin/schedules_controller.rb
sauloperez: Nope. It was perfectly covered by

https://github.com/openfoodfoundation/openfoodnetwork/blob/c73defa8a1eaca230f58196058c8ae4e5bfec827/spec/controllers/admin/schedules_controller_spec.rb#L133-L137

I wouldn't have noticed otherwise, tbh. If you think about the UI it makes sense.

Screenshot from 2020-05-29 15-21-25


@oeoeaio good a good job in terms of code coverage :ok_hand:

app/controllers/admin/schedules_controller.rb
sauloperez: yes, good catch

app/controllers/admin/schedules_controller.rb
luisramos0: :+1:

sauloperez:
No failures related to order_cycle_schedules :heavy_check_mark:

luisramos0:
:tada:

PR #5490, opened 2020-05-23
Change minimal amount in price sack calculator to float

### LINKED ISSUE #5303: Change PriceSack Minimal amount from Integer to Float
## Description

'Price Sack' Shipping method does not consider non-integer figures as Minimal amount, which is used as boundary condition for applying the Normal or the Discount amount. This is related to issue #4654 and PR #4812

While non-integer figures are accepted in the settings of the shipping method, one observes incorrect calculation in on checkout, caused by the truncation of a non-integer Minimal amount. When this happens, the shipping costs are calculated considering the Discount amount in cases which the Normal amount should be considered instead.

## Expected Behavior

'Price Sack' Shipping method should display the correct costs. Minimal amount non-integer values should be considered when these are compared with the order amounts.

## Actual Behaviour

Minimal amount values get truncated, leading to incorrect shipping cost calculations.

## Steps to Reproduce

1. As admin, and set up a shipping method using Price Sack and these exemplary values:

- Minimal amount = 16.5
- Normal amount = 5.0
- Discount amount = 1.0

2. Go to the shopfront, and place an order of 16.
3. Choose the PriceSack shipping method.
4. Observe that the displayed shipping cost is 1.0, while it rather should be 5.0 (as 16 < 16.5)

## Workaround

Using integers for the Minimal amount works fine.

## Severity

S4 - has a workaround, and perhaps not many users are impacted.

## Your Environment

* Version used: v2.9.5 
* Operating System and version (desktop or mobile): Ubuntu 18.04

luchiago:
#### What? Why?

Closes #5303 

Change the minimal amount of price sack from integer to float

#### What should we test?

Create a price sack with float minimal amount

#### Release notes

Changelog Category: Fixed

### DIFF
--- a/spec/models/spree/calculator/price_sack_spec.rb
+++ b/spec/models/spree/calculator/price_sack_spec.rb
@@ -49,6 +49,32 @@ describe Spree::Calculator::PriceSack do
     end
   end
 
+  context "minimal amount is float" do
+    before do
+      calculator.preferred_minimal_amount = 16.5
+      calculator.preferred_normal_amount = 5
+      calculator.preferred_discount_amount = 1
+      line_item.quantity = 2
+    end
+
+    context "with price bellow minimal amount" do
+      let(:price) { 8 }
+
+      it "returns the correct value of cost" do
+        expect(calculator.compute(line_item)).to eq(5)
+      end
+    end
+
+    context "with price above minimal amount" do
+      let(:price) { 8.5 }
+
+      it "returns the correct value of cost" do
+        expect(calculator.compute(line_item)).to eq(1)
+      end
+    end
+
+  end
+
   context "extends LocalizedNumber" do
     it_behaves_like "a model using the LocalizedNumber module", [:preferred_minimal_amount, :preferred_normal_amount, :preferred_discount_amount]
   end
--- a/app/models/spree/calculator/price_sack_decorator.rb
+++ b/app/models/spree/calculator/price_sack_decorator.rb
@@ -13,7 +13,7 @@ module Spree
     end
 
     def compute(object)
-      min = preferred_minimal_amount.to_i
+      min = preferred_minimal_amount.to_f
       order_amount = line_items_for(object).map { |x| x.price * x.quantity }.sum
 
       if order_amount < min


spec/models/spree/calculator/price_sack_spec.rb
```diff
@@ -49,6 +49,31 @@
...
+    context "with price bellow minimal amount" do
+      let(:price) { 8 }
+
+      it "returns the correct value of cost" do
+        expect(calculator.compute(line_item)).to eq(5)
+      end
+    end
+
+    context "with price above minimal amount" do
+      let(:price) { 8.5 }
```
sauloperez, line 68: sorry, is this correct? In line 54 I see `preferred_minimal_amount = 16.5` so it seems that the code contradicts the `context` description. Am I missing something?

luisramos0: ah, yes, you are missing quantity = 2
So 8.5 * 2 = 17 which is above 16.5
it's an existing readability problem in the spec.

luchiago: @sauloperez @luisramos0 It's true, thanks for the alert, I didn't notice this. I will change the price to 8.25, which gives 8.25 * 2 = 16.5. But the context is for above the minimal amount (16.5), which 17 is, doesn't sound correct?

luisramos0: I dont think you need to change the value, you could maybe make the quantity = 2 more obvious in some way.

luchiago: @luisramos0 Something like this?
```ruby
  context "minimal amount is float" do
    before do
      calculator.preferred_minimal_amount = 16.5
      calculator.preferred_normal_amount = 5
      calculator.preferred_discount_amount = 1
      line_item.quantity = 2
    end
```

sauloperez: Yes, I think it's fine. Although the assignment has no effect, next time I read this I'll know the dev was trying to tell me I should know the quantity to understand the test :ok_hand:

luchiago: Yeah, I agree. Great advice @sauloperez, I'll keep in mind.

luchiago: I made an amend commit with this change, not worth another commit for this, alright?

filipefurtad0:
Hey @luchiago !

I've set this values for testing:
image

Before the PR the mininal amount = 10.5 is considered an integer (10), and a cart equaling this value gets the discount amount (set as 4 eur). After this PR, a cart of 10 eur does not get the PriceSack discount amount, but rather the Normal amount 5. All good :-) 

Thank you for adding the specs for this as well :tada:

PR #5492, opened 2020-05-24
Include adjustments with negative taxes on Sales Tax reports

### LINKED ISSUE #5428: Sales tax report is showing amounts per types without adjustments
## Description

No idea if this is a bug or a new feature, but this is preventing a user to do his official VAT declaration... :facepalm:

Indeed in France we have several VAT rates, let's keep it simple for understanding the issue: wine is at a 20% VAT rate and cheese at a 5,5% VAT rate. 

So an order can have products with different tax rates.... but also adjustments with different tax rates as well...

## Expected Behavior

I think we need to have for each tax type the amount with adjustments, so that the sum of all tax type equals the total tax amount.

## Actual Behaviour

The 1,51 euros in the 5,5% tax type is the previous amount of tax on products _before_ a negative adjustment with a 5,5% tax has been added.

## Steps to Reproduce

1. Take an order without adjustments and have a look at the `admin/reports/sale_tax` per type
2. Add an adjustment on this order with a tax type
2. Check `admin/reports/sale_tax` again
3. See the amount in the tax rate you have selected has not changes but the total tax amount reflects correctly the adjustment

## Workaround

No idea, I've been doing now the report manually for them, but I've been lucky they only had 5,5% VAT adjustments. In the future they will also have 20% VAT adjustments, and because of https://github.com/openfoodfoundation/openfoodnetwork/issues/5422 I'm not sure what tax rate they've used in adjustment... But I can't also find the adjustment tax value without looking through each order one by one... :boom:

## Severity

? 

## Your Environment

* Version used: v2.9.8

pacodelaluna:
#### What? Why?

Closes #5428

On Sales Tax reports, with the Tax Rates filter, you can see that the tax is not managed when an adjustment is negative. This was because we were not including negative tax rate in the calculation, so I just removed the scope applying this filter.

#### What should we test?
Add a negative adjustment on an order. Then go on `/admin/reports/sales_tax`, select the `Tax Rates` filter, then control that you can see the negative amount for the tax.

#### Release notes
- Deal with negative tax amounts for adjustments on Sales Tax reports

Changelog Category: Changed


app/models/spree/order_decorator.rb
```diff
@@ -320,13 +322,6 @@ def tax_adjustment_totals
     end
   end
 
-  def price_adjustment_totals
-    Hash[tax_adjustment_totals.map do |tax_rate, tax_amount|
-      [tax_rate.name,
-       Spree::Money.new(tax_amount, currency: currency)]
-    end]
-  end
-
```
pacodelaluna, line 329: This method is not used anywhere.

app/models/spree/order_decorator.rb
```diff
@@ -305,12 +305,14 @@ def total_tax
   end
 
   def tax_adjustments
-    adjustments.with_tax +
-      line_items.includes(:adjustments).map { |li| li.adjustments.with_tax }.flatten
+    adjustments + price_adjustments
```
luisramos0, line 308: if you want to include negatives you can change the scope to include negatives
scope :with_tax,       -> { where('spree_adjustments.included_tax > 0') }

but here you are just including all adjustments, including the ones with zero tax. that means the method become non-sense where tax_adjustments returns all adjustments.

I think you want to keep the method as tax_adjustment but exclude included_tax ZERO.

pacodelaluna: Yes, you are right, I was not thinking of this case, and it will also solve the comment you put above for `adjustment.included_tax.zero?` control. I will fix this, need to cross-check side effect related to this scope use.

pacodelaluna:
@luisramos0 Thanks for your feedback, I will take a look at it this weekend, extract the logic into a service makes sense yes.

pacodelaluna:
@luisramos0 Sorry I was busy lately but I have refactored the code within a service. I have changed the scope to now include also the negative tax values, there will be a side effect but I think it is a desired effect. Tell me if any issue.

sauloperez:
it doesn't look good. Using `PROFILE` from master:

Screenshot from 2020-06-11 12-12-42

Using `PROFILE` from this branch:

Screenshot from 2020-06-11 12-13-14

pacodelaluna:
@sauloperez Oh ok, I am a bit surprised about this performance issue as I thought I was mostly moving code inside the PR. I will take a deeper look thsi weekend, I saw that there was an other service, ` OrderAdjustmentsFetcher`, I initially didn't want to add too much refactoring but it seems we really need it...

sauloperez:
I'm pretty sure it's scoped within `OrderTaxAdjustmentsFetcher#all`. Surely can be refactored to a single query starting from `Spree:Adjustment`.

sauloperez:
@pacodelaluna I gave it a try and added a couple of commits. Take a look at them and let me know what you think. They remove the N+1 and effectively fetch the adjustments in one shot.

There is at least one more N+1 related to adjustments in this report but this PR brings already a better world.

pacodelaluna:
@sauloperez Great job! Thanks for your help on this!

sigmundpetersen:
The conflict on this branch needs to be resolved before testing. A rebase would be good?

pacodelaluna:
@sigmundpetersen Just did the rebase, should be fine for testing.

filipefurtad0:
Hey @pacodelaluna!

I checked/reproduced what @RachL described in the issue. So, placed order R006812641, which can be seen in Tax Rates report:

image

Below, the same order, after adjustment with the amount £-20.00 (£-3.33 included tax):
image


After this PR the report is rendered with the correct values:
image

Awesome, thanks Paco :tada: 
Ready to go.

PS: I found that clicking on that order-hyperlink leads to /admin/orders/R006812641 -> which is neither related to this PR nor to the upgrade - checked on staging-es as well. Opening a separate issue.

PR #5493, opened 2020-05-25
Make display_on attribute available for managers to change in shipping and payment methods

luisramos0:
#### What? Why?

Fixes a problem introduced in #5392 where a message is shown to managers but the option was only available to admins. This PR makes the option available to all users of this page.

#### What should we test?
We dont need to test #5392 again, I think we just need to validate the option shows up for managers as well as admins now.

#### Release notes
Changelog Category: Added
Make shipping and payment methods option display_on available to managers.


filipefurtad0:
Hey @luisramos0,

This is good to go - the selector now appears for enterprise managers as well, and works as expected. Also, the warning has been included in the en.yml - perfect!

Ready to go.

Issue #5372, opened 2020-05-05
Error 404 if an OC closes during checkout: /checkout -> /undefined


filipefurtad0:
## Description

A customer adds an item and proceeds to the checkout page. Error 404 will be displayed when the customer attempts to finish the purchase (by pressing the Place order now), if the respective order-cycle closes while the customer is in the /checkout page.

## Expected Behavior

The customer should be redirected to the shopfront, a modal or a warning should be displayed that the chosen item(s) is/are nor longer available.

## Actual Behaviour

Redirection to the page /undefined - Snail 404

## Steps to Reproduce

1. User: populate the cart
2. User: proceed to check-out
3. Admin: close the order-cycle from which some products from the above cart are included
4. User: press Place order now (/checkout)
5. User gets redirected to /undefined - error 404

## Workaround

Manually navigating to the previous page will redirect the customer to the shopfront.

## Severity

No workaround, but an edge case - s3

## Your Environment

* Version used: v2.9.7
* Browser name and version: Firefox 75
* Operating System and version (desktop or mobile): Desktop

## Possible Fix

This is perhaps related to issue #5369 , since it also deals with the closing of an order-cycle on during the shopping-journey.

lin-d-hop:
The agreed solution is that in the case that the order cycle closes mid shopping the user will see an improved error message explaining what happened.

filipefurtad0:
Thanks for commenting this @lin-d-hop :heart:  I'll update the "Expected behaviour" on this.

Erioldoesdesign:
agreed with the above solution for papercut-ness

buttttttt just noting here how good it would be to have a countdown timer to when the order closes throughout the checkout process. Perhaps like when there's 1 hour till order cycle close theres a message that overlays on that shop.
Lots of online stores do this for sale periods and how long a system remembers items in a basket:

![image](https://user-images.githubusercontent.com/11681324/97929547-56c95380-1d61-11eb-80e8-b962f2a1c547.png)

![image](https://user-images.githubusercontent.com/11681324/97929597-706a9b00-1d61-11eb-9d6a-72490373ca6b.png)

The sale usage of this UI is pretty icky and pushy sales but in this case I think it could be useful to have a timeout. But worth testing with shoppers to be sure.

filipefurtad0:
Related Bugsnag log:
[...]

PR #5501, opened 2020-05-26
Enterprises Controller: reset_distributor must be called before any call to memoized current_distributor

### LINKED ISSUE #5496: Switching between shops does not work correctly
## Description

Switching between different shops does not work as expected: while logged in and shopping in shop A, clicking Shops in the main menu and choosing shop B from the shop list will first display again the shopfront of shop A (were customer was initially)... The address bar shows the address of shop B. See gif below.

## Expected Behavior

Switching shops by clicking on the main menu should work, and lead you to the shopfront you clicked on.

## Actual Behaviour

Clicking on another shop first showed the shopfront of the previous shop.

## Steps to Reproduce

Not always observable if in guest mode - Only occurs if logged in.

1. Log in (admin, customer...)
2. Click /shops in main menu
3. Choose an open shop A
4. Click again in /shops in main menu 
5. Choose another shop (other than the one you are), shop B
6. You should observe shop A again, while the address bar clearly shows the address of shop B...

## Animated Gif/Screenshot

switching_between_shops

## Workaround

Refreshing the page solves this.

## Severity

While it is not always observable (only while logged in) and it does have a workaround (refreshing page) this is quite a critical feature - switching shops... it's a release blocker - S2.

## Your Environment

* Version used: Release draft 2.9.10

luisramos0:
#### What? Why?

Closes #5496

The bug was caused by a hidden logical problem created by memoized current_distributor, current_customer that uses current_distributor cant be called before reset_distributor, otherwise the original (and outdated) distributor was memoized and returned by current_distributor...

The new spec in enterprises_controller validates the problem is fixed.

Moving current_distributor and current_customer to a service would be awesome here but would take too much time because they are called from quite a few places.

#### What should we test?
Verify bug is resolved.
We should probably give it a simple validation of #5440 as well

#### Release notes
Fix stale data problem when navigating across different shops.

### DIFF
--- a/app/controllers/enterprises_controller.rb
+++ b/app/controllers/enterprises_controller.rb
@@ -65,7 +65,10 @@ class EnterprisesController < BaseController
   def reset_order
     order = current_order(true)
 
-    OrderCartReset.new(order, params[:id], try_spree_current_user, current_customer).call
+    # reset_distributor must be called before any call to current_customer or current_distributor
+    order_cart_reset = OrderCartReset.new(order, params[:id])
+    order_cart_reset.reset_distributor
+    order_cart_reset.reset_other!(try_spree_current_user, current_customer)
   rescue ActiveRecord::RecordNotFound
     flash[:error] = I18n.t(:enterprise_shop_show_error)
     redirect_to shops_path
--- a/app/services/order_cart_reset.rb
+++ b/app/services/order_cart_reset.rb
@@ -2,24 +2,11 @@
 
 # Resets an order by verifying it's state and fixing any issues
 class OrderCartReset
-  def initialize(order, distributor_id, current_user, current_customer)
+  def initialize(order, distributor_id)
     @order = order
     @distributor ||= Enterprise.is_distributor.find_by_permalink(distributor_id) ||
                      Enterprise.is_distributor.find(distributor_id)
-    @current_user = current_user
-    @current_customer = current_customer
   end
-
-  def call
-    reset_distributor
-    reset_user_and_customer if current_user
-    reset_order_cycle
-    order.save!
-  end
-
-  private
-
-  attr_reader :order, :distributor, :current_user, :current_customer
 
   def reset_distributor
     if order.distributor && order.distributor != distributor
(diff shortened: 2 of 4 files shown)


RachL:
I've tested the scenario in the issue:


    Log in (admin, customer...)
    Click /shops in main menu
    Choose an open shop A
    Click again in /shops in main menu
    Choose another shop (other than the one you are), shop B
    You should observe shop A again, while the address bar clearly shows the address of shop B...

About grumpy cat:

1. Single OC, all good:
image

2. However with 2 OC open or 3 and if I close one of them(e.g. with Phillippe shop), I'm redirected to the home page like in test number 1.

Is this intended? It's not clear to me while reading #5440 . In any case I don't think it is a blocker and we can merge this work as is and improve in a second round. At least we don't have a grumpy cat anymore :)

ping @luisramos0 @filipefurtad0 @kristinalim

filipefurtad0:
I agree @RachL, this is good to go.

If the OC is closed while on /checkout and I click the back button on the browser (like in Nick's video on the issue https://github.com/openfoodfoundation/openfoodnetwork/issues/5340#issuecomment-622345005) I always land on the /shop page, and observe OC selector (if any left), but no flash message.

I've searched a lot (as this was indicated as an outcome of a test case) and the only way I found to observe the flash was clicking directly in Shops, in the menu, instead of clicking back on the browser, from /checkout. So... How do you do it? (magic!) :-)

Well, main point is probably that the release is not blocked any more :tada:

luisramos0:
yeah, that redirect is not correct Rachel, can you replicate it? when you "I am redirected", in what exact page are you and what do you do? page refresh or click somewhere?

RachL:
@luisramos0 @filipefurtad0 yes I can replicate. You need to do the following:

1. Go to shop with several OC open and add products to the cart, but don't go any further. Stay on the shop page, leave the browser tab opened

2. Close the OC you have just used with another browser window

3. Come back to your tab with the shop
4. Click on the cart on the main menu and "proceed to checkout"

There the redirect happens.I also reproduced it while being on the cart page. I'm sorry, I didn't think we could have different behavior, so I didn't test the checkout page. Now I get a snail with this scenario but with closing the shop while trying to checkout...might be my internet connexion, I will retry in the morning...

RachL:
Outch, wasn't my internet access, I could reproduce the snail this morning it's a 404 error:

My test was: I had a user who was on /checkout, I closed the OC, then I made this user click on the checkout button at the bottom of the page. I ended up with a 404:

image

However when I press the browser back button I indeed ended up with the shop open on the other OC, no flash message.

Is that 404 something you saw earlier @filipefurtad0 ?

filipefurtad0:
Hey @RachL ,

Thanks for checking this again.
Yes I saw this, it's #5372

RachL:
@filipefurtad0 great!

luisramos0:
Hey, do we need a new issue for the redirect problem? I mean, if there are 2 OCs and the hub closes one of them the redirect should go to the order OC, not to /shops.

filipefurtad0:
Yes @luisramos0,
Redirection should be to /shop, not shops. I'll open an issue and link it #5440.
Thanks for mentioning this.

PR #5502, opened 2020-05-26
[mobile ux] Fix missing drop-shadow

### LINKED ISSUE #5477: [mobile ux] Re-add drop shadow to tabs menu in the shopfront
## Description
The quickfix #5451 removed the dropshadow from the tabs menu in the shopfront, see here for more details:
https://github.com/openfoodfoundation/openfoodnetwork/issues/5433#issuecomment-630126531

Now we need to find a way to add this drop shadow again.

## Severity

## Possible Fix
The initial fix was to move the z-index of the tabs menu up so that the menu stands above the content below but that created the bug that 5451 fixes: it was hiding the OC selector. We need an alternative approach.

Matt-Yorkley:
#### What? Why?

Closes #5477

Adjusts z-index on overlapping elements to ensure drop-shadow is shown. This was a small regression in the last release.

#### What should we test?

Drop-shadow is show in desktop view between shop tabs and search bar:

Screenshot from 2020-05-26 13-03-32

Also check the previous issue in the release is still fixed (the full area of the order cycle selector is clickable). See here for notes: https://github.com/openfoodfoundation/openfoodnetwork/issues/5433#issuecomment-630089588

#### Release notes

Re-added missing shadow under tabs in shop page

### DIFF
--- a/app/assets/stylesheets/darkswarm/distributor_header.css.scss
+++ b/app/assets/stylesheets/darkswarm/distributor_header.css.scss
@@ -11,7 +11,7 @@ section {
   display: block;
   background: $white;
   position: relative;
-  z-index: 2;
+  z-index: 20;
 
   .details {
     box-sizing: border-box;
--- a/app/assets/stylesheets/darkswarm/shop_tabs.css.scss
+++ b/app/assets/stylesheets/darkswarm/shop_tabs.css.scss
@@ -9,6 +9,7 @@
     color: $dark-grey;
     box-shadow: $distributor-header-shadow;
     position: relative;
+    z-index: 10;
 
     .columns {
       display: flex;


app/assets/stylesheets/darkswarm/distributor_header.css.scss
```diff
@@ -11,7 +11,7 @@ section {
   display: block;
   background: $white;
   position: relative;
-  z-index: 2;
+  z-index: 20;
```
mbudm, line 14: Perhaps I'm not looking in the right place but it looks like this z-index property can just be removed? It doesn't seem to have any effect?
Screen Shot 2020-05-27 at 11 22 11 am

Matt-Yorkley: The previous issue noted here: https://github.com/openfoodfoundation/openfoodnetwork/issues/5433#issuecomment-630089588 is broken again if we remove that line.

daniellemoorhead:
### Testing notes



Good to go!

Issue #4806, opened 2020-02-19
Update counts in OC edit page to exclude hidden overrides


Matt-Yorkley:
## Description

Part of #4355 

Counts shown in OC edit page include hidden overrides that cannot be added to the OC, resulting in confusing feedback.

See additional screenshots and notes here: https://github.com/openfoodfoundation/openfoodnetwork/issues/4355#issuecomment-587414047 and here: https://github.com/openfoodfoundation/openfoodnetwork/issues/4355#issuecomment-587419377 

## Expected Behavior

Variant counts shown in the OC edit page should make sense.

## Actual Behaviour

The numbers are misleading, eg: "4 of 2 Variants loaded".

## Steps to Reproduce

1. Add some variants to inventory, and hide them
2. View the variant counts for a related OC

## Workaround

None.

## Severity

bug-s4: it's annoying, but you can use it

luisramos0:
Hmm. trying to figure out an easy way out of this one...

I thought I could make hidden variants visible but not editable. They would show up but be disabled and with a new tooltip:
![Screenshot from 2020-05-26 15-35-36](https://user-images.githubusercontent.com/1640378/82913722-92682200-9f66-11ea-87f7-ba92f8d4ecb0.png)
The counts would be correct because they would be included now.
Thoughts?

@Matt-Yorkley the code would be something like this: #5508

Matt-Yorkley:
LGTM :+1:

Does it work OK with the select all button?

luisramos0:
this needs product feedback, probably better to double check that @RachL @lin-d-hop and @kirstenalarsen dont see any issues with this decision: merge the two cases 1. visible only variant (user has no permissions to add to exchange) and 2. hidden variant (variant is marked as hidden in inventory)

RachL:
@luisramos0 just too be sure: the idea is that instead of not counting visible only variants and hidden variants, we would continue to count them but display them to the user. Is that correct?

I think I like the logic because it can save time. User don't waste time in searching why they are missing variants. But I have 2 questions:

- what is the scope of the visible only variant? What kind of permissions are in place for this case to happen?
- how does this rule apply with the select all? With a very large catalog, it would be difficult to find those that are unselected. Can we put them on top of the list?

luisramos0:
>>  what is the scope of the visible only variant? What kind of permissions are in place for this case to happen?

It's complicated... this is the description in the code:
[...]

PR #5508, opened 2020-05-26
OC edit page: make hidden variants visible but not editable

luisramos0:
…ing reasons for a variant to not be editable by the user

#### What? Why?

#### Release notes


luisramos0:
as discussed here, this solution is not good enough:
https://github.com/openfoodfoundation/openfoodnetwork/issues/4806#issuecomment-648001412

PR #5530, opened 2020-05-29
Remove spinjs-rails dependency and use spinning-circles.svg instead.

### LINKED ISSUE #3491: Investigate removal of spinjs-rails gem

Our Gemfile contains spinjs-rails and we include the code in our front-end theme called Darkswarm. But we think that it's unused. This task includes:

- Find pages with the current spinner and verify they are working.
- Remove the gem from the Gemfile and `app/assets/javascripts/darkswarm/all.js.coffee`.
- Verify that the spinner is still working.

### Context

Dependabot prompted to update the gem version in https://github.com/openfoodfoundation/openfoodnetwork/pull/3460 which made us think about its use.

### Impact and timeline

 Keeping this gem dependency means regular reviews of updates of this gem. It should be removed as soon as possible. The task should only take a couple of hours.

cillian:
#### What? Why?

Closes #3491

The spinjs-rails gem provides a CSS3 a spinning activity indicator. This commit removes the gem so we don't have to keep the gem up-to-date and uses the spinning-circles.svg which is used already in other places.

This is the loading message using spinjs.

loading-with-spinjs

This is the loading message using spinning-circles.svg .  One problem here is that spinner circles are a bit difficult to see on the blue background. Not sure the best way to fix that yet.

loading-without-spinjs

#### What should we test?

The spinjs spinner can be seen when viewing or editing taxonomies in /admin/taxonomies. It is loaded from the .ajaxStart call in app/assets/javascripts/admin/spree/progress.coffee

To test or restyle this spinner you can change the display property of the #progress element in app/assets/stylesheets/admin/components/progress.scss to 'block'.

#### Release notes

Remove spinjs-rails dependency and use existing spinning-circles.svg instead.

Changelog Category: Changed


sauloperez:
being it SVG, you should be able to use `fill: currentColor` to change the color of the spinner. More details: https://css-tricks.com/cascading-svg-fill-color/

sauloperez:
I went ahead and added a commit to fix it. I thought it was easier to do it than explaining it. It now looks like 

Screenshot from 2020-06-03 11-15-34

app/views/spree/layouts/_admin_body.html.haml
```diff
@@ -10,11 +10,29 @@
     .flash.success= flash[:success]
 
   #progress
-    .wrapper
-      #spinner
-      .progress-message
-        = Spree.t(:loading)
-        \...
+    / By Sam Herbert (@sherb), for everyone. More @ http://goo.gl/7AJzbL
```
luisramos0, line 13: you are inlining an SVG directly in a layout view? It doesn't look good. Any alternative?

luisramos0:
Moving to code review for discussion, the inline svg doesn't look good to me.

app/views/spree/layouts/_admin_body.html.haml
sauloperez: we'll need to move it to a partial and reuse it in app/views/spree/layouts/bare_admin.html.haml too. Would that feel ok @luisramos0 ?

luisramos0: yes, a specific partial is better, but why not a svg file?

sauloperez: because it is then impossible to style it with CSS and style the spinner different than the image we have now. I understand you might be skeptical but it's an incredibly common technique to replace icon fonts, for instance. Just check your GH's dashboard and look up `svg`. It's basically a lot more changeable.

luisramos0: hmmm, interesting, I didnt know that, thanks!
So let's move it to a partial?
Shall we have a specific folder for svgs in app/views?

sauloperez: > Shall we have a specific folder for svgs in app/views?

I think it's too early for that. We don't even know if we'll do this again any time soon. It's something we can consider if it happens.

luisramos0: ok :+1:

cillian:
@sauloperez good to know about `fill: currentColor` thanks, looks good.

luisramos0:
Moving to in dev until the svg is moved to a partial as agreed above. Just to clear the code review column a bit.

sauloperez:
@cillian feel free to move that to a partial yourself. There are other things that are keeping me busy ATM.

filipefurtad0:
Hey @cillian ,

I checked that the rails-spinner is now gone.

Before this PR:
image

After this PR, the .sng takes it's place:
image

All is working fine - ready to go. Thank you!

PS: However, I did try several changes to the file under:
/apps/openfoodnetwork/current/app/assets/stylesheets/admin/components/progress.scss

on the staging-server, but couldn't observe any changes on the spinner. Any ideas why?

cillian:
Hey @filipefurtad0 thanks

It looks like I'm getting the same issue in development in Firefox too. If I change `progress.css` I don't see the changes after pressing reload but if I do a hard reload to clear cache with `ctrl + f5` it shows the changes. In Chrome it doesn't seem to be a problem. I wonder could that be it?

luisramos0:
@filipefurtad0 can I merge or should I wait?

sauloperez:
Ups, sorry, I didn't see that last comment. I think we answered the issue in Slack.

filipefurtad0:
Thank you @luisramos0 - it's fine to merge.

I tried changing browser and hard reload as well @cillian, but after chatting with our devs I've found out that such css content is precompiled, so such changes won't appear before they compiled again in the staging server. So, it's something on the staging our side.
Thank for your feedback anyway.

PR #5532, opened 2020-05-30
Fix admin config separator withou highlight

### LINKED ISSUE #4730: Configuration in Main Menu does not highlight correctly
### Description

Main menu entries are highlighted when selected, and during operations in submenu entries.

### Expected Behaviour

Logged in as Superadmin, one can click on Configurations under the main Menu, which highlights the Configurations button, as expected:

### Actual Behaviour

Upon selection of the Mail Method Settings submenu, the highlighting dissappears:

### Steps to Reproduce

- Log in as SuperAdmin.
- Click on the Configurations button (main menu).
- Click on the submenus, namely on the Mail Method Settings
- Highlighting on the Configurations button dissappears

### Context

Editing general configurations, as SuperAdmin.

### Severity
bug-s5

Impact is minimal. The customer will not know why the change was not possible and will not know how to resolve it. An appropriate error message will improve customer experience. The work around is to re-enter the products.

### Your Environment
Version used: v2.2.2
OFN Platform instance: Staging OFN Katuma

### Possible Fix

The name of the route needs to be changed but we need to check if it would break other things...

luchiago:
#### What? Why?

Closes #4730 

A little error caused by the name of the controller. My solution was to fix it, putting it on plural
Before:
before

After:
after

#### What should we test?

Just go to configurations and access `mail method`

#### Release notes

Change the mail method route name to plural, to be recognized when selected in configurations on admin

Changelog Category: Fixed

### DIFF
--- a/app/controllers/spree/admin/mail_methods_controller.rb
+++ b/app/controllers/spree/admin/mail_methods_controller.rb
@@ -23,7 +23,7 @@ module Spree
       rescue StandardError => e
         flash[:error] = Spree.t('admin.mail_methods.testmail.error') % { e: e }
       ensure
-        redirect_to edit_admin_mail_method_url
+        redirect_to edit_admin_mail_methods_url
       end
 
       private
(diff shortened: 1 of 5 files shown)


luchiago:
@luisramos0 Alright, I added to the description of the PR the testing case that you said.

PR #5536, opened 2020-05-30
Set :path in `whenever` configs.

### LINKED ISSUE #5535: Duplicate crontab entries via whenever gem
We now have duplicates of all our crontab entries, due to the change in app paths:

This needs a fix and then a cleanup of the old entries...

Matt-Yorkley:
This was defaulting to the new symlinked app path target, creating a new value each time, which causes whenever to create duplicate crontab entries for each deploy...

#### What? Why?

Part of #5535 

Fixes a config issue causing the whenever gem to create duplicate crontab entries.

#### What should we test?

Duplicate crontab entries are not created on each deploy.

#### Release notes

Fixed an issue with scheduled cron jobs.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -6,11 +6,11 @@ require 'yaml'
 app_config = YAML.load_file(File.join(__dir__, 'application.yml'))
 
 env "MAILTO", app_config["SCHEDULE_NOTIFICATIONS"] if app_config["SCHEDULE_NOTIFICATIONS"]
+set :path, ENV['WHENEVER_PATH'] if ENV['WHENEVER_PATH']
 
 # If we use -e with a file containing specs, rspec interprets it and filters out our examples
 job_type :run_file, "cd :path; :environment_variable=:environment bundle exec script/rails runner :task :output"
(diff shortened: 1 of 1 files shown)


config/schedule.rb
```diff
@@ -6,12 +6,12 @@
 app_config = YAML.load_file(File.join(__dir__, 'application.yml'))
 
 env "MAILTO", app_config["SCHEDULE_NOTIFICATIONS"] if app_config["SCHEDULE_NOTIFICATIONS"]
+set :path, '/home/openfoodnetwork/apps/openfoodnetwork/current'
```
luisramos0, line 9: I dont think we can have this path in ofn. we need to move this to ofn-install in some way, right?

Matt-Yorkley: Updated

Matt-Yorkley: Associated PR here: https://github.com/openfoodfoundation/ofn-install/pull/631

luisramos0: :ok_hand:

Matt-Yorkley:
Closing; it looks liek we can do it all on the ofn-install side without modifying the main repo here.

Issue #5533, opened 2020-05-30
[Spree 2.1] Broken spec - EnterprisesController missing routes


Matt-Yorkley:
Errors:

```
  1) EnterprisesController as an enterprise owner creating an enterprise creates as sells=any when it is not a producer
     Failure/Error: spree_post :create, { enterprise: new_enterprise_params }
     
     AbstractController::ActionNotFound:
       The action 'create' could not be found for EnterprisesController
     # ./spec/controllers/api/enterprises_controller_spec.rb:33:in `block (4 levels) in <module:Api>'

  2) EnterprisesController as an enterprise owner creating an enterprise saves all user ids submitted
     Failure/Error:
       spree_post :create, {
         enterprise: new_enterprise_params.
           merge({ user_ids: [enterprise_owner.id, manager1.id, manager2.id] })
       }
     
     AbstractController::ActionNotFound:
       The action 'create' could not be found for EnterprisesController
     # ./spec/controllers/api/enterprises_controller_spec.rb:43:in `block (4 levels) in <module:Api>'

  3) EnterprisesController as an enterprise manager submitting a valid image I can update enterprise image
     Failure/Error: spree_post :update_image, logo: 'a logo', id: enterprise.id
     
     AbstractController::ActionNotFound:
       The action 'update_image' could not be found for EnterprisesController
     # ./spec/controllers/api/enterprises_controller_spec.rb:71:in `block (4 levels) in <module:Api>'

  4) EnterprisesController as an non-managing user submitting a valid image I can't update enterprise image
     Failure/Error: spree_post :update_image, logo: 'a logo', id: enterprise.id
     
     AbstractController::ActionNotFound:
       The action 'update_image' could not be found for EnterprisesController
     # ./spec/controllers/api/enterprises_controller_spec.rb:90:in `block (4 levels) in <module:Api>'
```

Matt-Yorkley:
I think this is caused by this Devise user mappings issue: https://stackoverflow.com/a/4321055

luisramos0:
This could be some mess up between spree routes and Ofn routes. shop and enterprise controllers are both ofn only routes, maybe devise (or the test) is only using/loading/seeing spree routes.

luisramos0:
I tried playing with api_post instead of spree_post
I see that removing use_route: :spree from ControllerHacks#api_process
I get this one to fail with similar problems as this issue:
spec/controllers/api/variants_controller_spec.rb:72:
[...]

PR #5537, opened 2020-05-30
[Spree 2.1] Make devise play nicely with controller specs in Rails 4

### LINKED ISSUE #5504: [Spree 2.1] Broken spec - ShopsController missing routes
These were introduced recently from merging master. 

Errors:

```

  1) ShopsController as a non-authenticated user #show returns shopfront data for an enterprise
     Failure/Error: spree_get :show, id: producer.id
     
     AbstractController::ActionNotFound:
       The action 'show' could not be found for ShopsController
     # ./spec/controllers/api/shops_controller_spec.rb:27:in `block (4 levels) in <module:Api>'

  2) ShopsController as a non-authenticated user #closed_shops returns data for all closed shops
     Failure/Error: spree_get :closed_shops, nil
     
     AbstractController::ActionNotFound:
       The action 'closed_shops' could not be found for ShopsController
     # ./spec/controllers/api/shops_controller_spec.rb:37:in `block (4 levels) in <module:Api>'

```

Matt-Yorkley:
#### What? Why?

Hopefully closes #5504 and #5533...

Looks like an obscure issue related to setting up devise with custom user mappings (`spree_user`) in controller tests. We've had a few devise issues so far, hopefully this is the last... 

More info here: https://stackoverflow.com/questions/4291755/rspec-test-of-custom-devise-session-controller-fails-with-abstractcontrollerac

#### What should we test?

Green specs on controllers, no `AbstractController::ActionNotFound` errors like this:

```
1) EnterprisesController as an enterprise owner creating an enterprise creates as sells=any when it is not a producer
     Failure/Error: spree_post :create, { enterprise: new_enterprise_params }
     
     AbstractController::ActionNotFound:
       The action 'create' could not be found for EnterprisesController
     # ./spec/controllers/api/enterprises_controller_spec.rb:33:in `block (4 levels) in <module:Api>'
```

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -133,6 +133,13 @@ RSpec.configure do |config|
 
   config.before(:all) { restart_driver }
 
+  # Configure Devise to work nicely with controller specs
+  # See: https://github.com/openfoodfoundation/openfoodnetwork/issues/5533
+  config.before(:each, :controller) do
+    setup_controller_for_warden
+    @request.env["devise.mapping"] = Devise.mappings[:spree_user]
+  end
+
   # Geocoding
   config.before(:each) { allow_any_instance_of(Spree::Address).to receive(:geocode).and_return([1, 1]) }
 


luisramos0:
hey Matt, it seems like #5504 is gone from 3-0-stable and this PR has no effect on #5533.
What do you think?

Matt-Yorkley:
This PR doesn't fix the issue, no. I hoped it would.

I have a strong feeling the two issues are related, even though #5504 has magically disappeared from the build and been replaced with #5533.

- Both cases are for API controllers (`api/shops_controller_spec` and `api/enterprises_controller_spec`)
- Both cases couldn't be replicated locally but consistently failed in Semaphore
- Both cases appeared randomly and all the tests in the relevant spec file failed with a weird missing route error, as if the routes didn't exist for any of the actions for that controller (but did exist for al the other API controllers)
- It looks like after some things were merged (possibly re-arranging the order that specs are run in each Semaphore node), the errors jumped from one API controller spec file to the other

It's very weird.

PR #5538, opened 2020-05-30
Fix Payment Methods table to show proper provider name

### LINKED ISSUE #4592: Payment Methods Table in Dashboard does not display in Plain Language
## Description
Table showing Payment Methods in Dashboard includes non-intuitive language that would not make sense to a customer. eg. Spree::PaymentMethod::Check, Environment: Production 

## Expected Behaviour
Table fields should display information in plain language

## Actual Behaviour
Log in as an Enterprise
Access Dashboard
Select Payment Methods 
Click on Manage Payment Methods button

## Animated Gif/Screenshot

Screen Shot 2019-12-12 at 4 50 01 PM

## Workaround
Table is still visible, but much of the information is meaningless to a customer

## Severity
S4

## Your Environment

Australian Instance

jeduardo824:
#### What? Why?

Closes #4592 

Before my fix, the provider name column in the Payment Methods table was showing the class name like: `Spree::Gateway::PaypalExpress`. The codebase already has a class method to clean those names, and I used it on the view.

#### What should we test?
Check if the Payment Methods table shows the provider name properly.

#### Release notes
Fix the Payment Method table to show the provider name properly

Changelog Category: Fixed

Before
Screen Shot 2020-05-30 at 18 03 29

After
Screen Shot 2020-05-30 at 18 03 46

### DIFF
--- a/spec/features/admin/payment_method_spec.rb
+++ b/spec/features/admin/payment_method_spec.rb
@@ -185,16 +185,6 @@ feature '
       expect(page).not_to have_content payment_method3.name
     end
 
-    it "shows only the providers of the existing payment methods" do
-      payment_method1
-      payment_method2
-      payment_method3
-
-      visit spree.admin_payment_methods_path
-
-      expect(page).to have_content "Cash/EFT/etc. (payments for which automatic validation is not required)", count: 2
-    end
-
     it "does not show duplicates of payment methods" do
       payment_method1
       payment_method2
--- a/spec/features/admin/payment_method_spec.rb
+++ b/spec/features/admin/payment_method_spec.rb
@@ -185,6 +185,16 @@ feature '
       expect(page).not_to have_content payment_method3.name
     end
 
+    it "shows only the providers of the existing payment methods" do
+      payment_method1
+      payment_method2
+      payment_method3
+
+      visit spree.admin_payment_methods_path
+
+      expect(page).to have_content "Cash/EFT/etc. (payments for which automatic validation is not required)", count: 2
+    end
+
     it "does not show duplicates of payment methods" do
       payment_method1
       payment_method2
--- a/spec/views/spree/admin/payment_methods/index.html.haml_spec.rb
+++ b/spec/views/spree/admin/payment_methods/index.html.haml_spec.rb
@@ -1,0 +1,28 @@
+require "spec_helper"
+
+describe "spree/admin/payment_methods/index.html.haml" do
+  before do
+    controller.singleton_class.class_eval do
+      helper_method :new_object_url, :edit_object_url, :object_url
+
+      def new_object_url() "" end
+
+      def edit_object_url(object, options = {}) "" end
+
+      def object_url(object = nil, options = {}) "" end
+    end
+
+    assign(:payment_methods, [
+      create(:payment_method),
+      create(:payment_method)
+    ])
+  end
+
+  describe "payment methods index page" do
+    it "shows only the providers of the existing payment methods" do
+      render
+
+      expect(rendered).to have_content "Cash/EFT/etc. (payments for which automatic validation is not required)", count: 2
+    end
+  end
+end
--- a/app/views/spree/admin/payment_methods/index.html.haml
+++ b/app/views/spree/admin/payment_methods/index.html.haml
@@ -32,7 +32,7 @@
             - method.distributors.each do |distributor|
               = distributor.name
               %br/
-          %td= method.type
+          %td= method.class.clean_name
           %td.align-center= method.environment.to_s.titleize
           %td.align-center= method.display_on.blank? ? t('.both') : t('.' + method.display_on.to_s)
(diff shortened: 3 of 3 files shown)


spec/features/admin/payment_method_spec.rb
```diff
@@ -185,6 +185,16 @@
...
 
+    it "shows only the providers of the existing payment methods" do
+      payment_method1
+      payment_method2
+      payment_method3
+
+      visit spree.admin_payment_methods_path
+
+      expect(page).to have_content "Cash/EFT/etc. (payments for which automatic validation is not required)", count: 2
+    end
```
sauloperez, line 196: I would rather move this test to a lower level. Controller at least although I think it's the perfect use-case for a view test. The thing we're testing isn't worth booting a browser and going through the entire stack IMO

sauloperez:
Pretty useful fix @jeduardo824 ! thanks!

spec/features/admin/payment_method_spec.rb
jeduardo824: I'm trying to create the view spec but I'm getting an error. I will push what I did until now. Could you help me?

Screen Shot 2020-06-08 at 00 16 08

luisramos0: that's an helper method defined in the parent ResourceController, maybe the controller is not fully loaded for view specs?
maybe you can load the helper in the view spec explicitly?

if it was an helper it would be:
helper Spree::BaseHelper
but this is an helper method, maybe we can at least try to validate this theory by defining the method in the class_eval of the spec:

```
    controller.singleton_class.class_eval do
      def current_ability
        Spree::Ability.new(Spree.user_class.new)
      end

      def new_object_url
          ...
      end
    end

```

I am sorry, this is not very good code, but I think it's worth taking these ugly shortcuts to get more view specs as we go, this is the 5th spec now :tada:

jeduardo824: I will try!

jeduardo824: @luisramos0 I tried with your approach! No success

jeduardo824: Screen Shot 2020-06-12 at 18 54 02

luisramos0: re my approach, can you please push the code you have tried and the error you get?

jeduardo824: @luisramos0 I tried to define the method inside the `class_eval` block, but it keeps complaining about the undefined method "new_object_url". The same error that I posted when you recommended defining inside the class_eval.

jeduardo824: Screen Shot 2020-06-08 at 00 16 08

jeduardo824: I was able to fix stubbing the method "new_object_url". I will try to finish and I will push the spec.

jeduardo824: Done!

jeduardo824: I was forgetting to call "helper_method", sorry @luisramos0 ! I updated the PR with the view test.

spec/views/spree/admin/payment_methods/index.html.haml_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        ""
+      end
+
+      def edit_object_url(object, options = {})
+        ""
+      end
+
+      def object_url(object = nil, options = {})
+        ""
+      end
```
luisramos0, line 20: To make the hack look smaller I'd:
- remove the comment line # Fill...
- use the single line notation:
```
      def new_object_url() "" end
      def edit_object_url(object, options = {}) "" end
      def object_url(object = nil, options = {}) "" end
```

jeduardo824: I agree with the single line notation but I would stay with the comment. I think it will help new developers understand why those methods are there.

jeduardo824: But you understand this codebase more than me so I will be happy to remove it if you think it is not necessary (I particular don't like code with comments but I thought it would be useful on this case)

luisramos0: it's fine, as you wish. it triggered me because it's about the future. no comment should be about the future, just the current state.

jeduardo824: Removed!

Matt-Yorkley:
Moving to test ready. It looks like Pau's request is resolved.

filipefurtad0:
Hi @jeduardo824 ,

The class names are now removed:

Before:
image

And after:
image

Good to go, Thank you!

PR #5540, opened 2020-06-01
Missing translation key "return_authorizations"

romale:
#### What? Why?

Add missing translation on orders tabs.

#### What should we test?
Return authorizations tab entry is translated. See screenshot below.

#### Release notes
Changelog Category: Added
Added missing translation for return authorizations menu entry.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2905,7 +2905,6 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   first: "First"
   previous: "Previous"
   last: "Last"
-  
   spree:
     your_order_is_empty_add_product: "Your order is empty, please search for and add a product above"
     add_product: "Add Product"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2879,7 +2879,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   customer_details: "Customer Details"
   adjustments: "Adjustments"
   payments: "Payments"
-
+  return_authorizations: "Return Authorizations"
   payment: "Payment"
   payment_method: "Payment Method"
   shipment: "Shipment"
@@ -2905,7 +2905,6 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   previous: "Previous"
   last: "Last"
   
-  return_authorizations: "Return Authorizations"
   spree:
     your_order_is_empty_add_product: "Your order is empty, please search for and add a product above"
     add_product: "Add Product"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2880,7 +2880,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   adjustments: "Adjustments"
   payments: "Payments"
   return_authorizations: "Return Authorizations"
-  
+
   payment: "Payment"
   payment_method: "Payment Method"
   shipment: "Shipment"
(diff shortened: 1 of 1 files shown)


luisramos0:
Please post a screenshot of the page where you see the missing translation.

romale:
image

config/locales/en.yml
```diff
@@ -2904,7 +2904,8 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   first: "First"
   previous: "Previous"
   last: "Last"
-
+  
+  return_authorizations: "Return Authorizations"
```
luisramos0, line 2908: the right place for this is line 2882 above where the other menu entries are.

luisramos0:
thanks for the valid PR :+1:
Can you please move the translation to line 2882?

romale:
> thanks for the valid PR +1
> Can you please move the translation to line 2882?

ok, it's ready

config/locales/en.yml
```diff
@@ -2879,7 +2879,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   customer_details: "Customer Details"
   adjustments: "Adjustments"
   payments: "Payments"
-
+  return_authorizations: "Return Authorizations"
   payment: "Payment"
```
luisramos0, line 2883: can you leave the empty line as before?

config/locales/en.yml
```diff
@@ -2904,7 +2904,7 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   first: "First"
   previous: "Previous"
   last: "Last"
-
+
```
luisramos0, line 2907: there are extra characters here.

romale:
> there are some extra empty characters in line 2907

it was before my change. delete?

luisramos0:
please have a look at the diff in gh:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5540/files

return authorizations belongs to the group of keys above it (it's the orders menu with Customer Details etc), so the empty line would be after it.

config/locales/en.yml
romale: corrected

config/locales/en.yml
romale: corrected

config/locales/en.yml
```diff
@@ -2904,7 +2905,6 @@ See the %{link} to find out more about %{sitename}'s features and to start using
   first: "First"
   previous: "Previous"
   last: "Last"
-
```
luisramos0, line 2907: I dont think we should remove this line, this is the start of one of the major parts of this file, the spree key.

config/locales/en.yml
luisramos0: I am sorry about the little problems but you have added extra spaces or a tab where it should be empty line
always have a look at https://github.com/openfoodfoundation/openfoodnetwork/pull/5540/files

romale:
> all right, it's now correct +1
> 
> would be better to squash it into one commit, instead of 7.
> You would need to do
> git rebase -i master
> and replace pick with squash on all lines except the first one where you can leave pick

is there any instruction how to do this? i'm not git/github ninja/guru

sauloperez:
Thanks for the contribution @romale ! https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history is a good walkthrough :ok_hand: or https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d. 

Whenever you feel lost with Git, https://ohshitgit.com/ comes in handy.

luisramos0:
moving to test ready with one review as it is very simple PR.

filipefurtad0:
Hi @romale,

Thank you for this PR!

The page it should appear in is rendering well in english, so that means that the en.yml file is well formulated and should show the translation in russian, once the translation key is complete:

image

Thanks again!

Issue #5385, opened 2020-05-07
[Spree 2.1] Broken spec spec/models/concerns/product_stock_spec.rb:19


luisramos0:
### What we should change and why (this is tech debt)
This spec is flaky
spec/models/concerns/product_stock_spec.rb:19

I cant replicate this locally....

### Context
See branch 3-0-stable

luisramos0:
This fails consistently in the build but I cant replicate locally.

PR #5541, opened 2020-06-01
Delete dead code related to products without variants

luisramos0:
#### What? Why?

Related to #5385 
A product always has what's called a standard variant. This variant is created on saving a product without variants. This means has_variants? will always return true except before saving a product. We can delete all calls to has_variants in places where the product has been saved.

#### What should we test?
We should verify we can create and edit a product as well as create and edit variants.
We should try to delete the last variant from a product and see that it is not possible.
Adding variant specific images should also work correctly (altough they cant be used, the shopfront will only use the main product image).

#### Release notes
Changelog Category: Removed
Deleted dead code related to handling products without variants.

### DIFF
--- a/app/models/spree/product_decorator.rb
+++ b/app/models/spree/product_decorator.rb
@@ -158,18 +158,6 @@ Spree::Product.class_eval do
     self.class.in_order_cycle(order_cycle).include? self
   end
 
-  # overriding to check self.on_demand as well
-  def has_stock?
-    has_variants? ? variants.any?(&:in_stock?) : (on_demand || master.in_stock?)
-  end
-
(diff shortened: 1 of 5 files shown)


sauloperez:
I'm moving this for the sake of the pipe's throughput

filipefurtad0:
Hi @luisramos0 ,

I've checked the usual process of creating/editing new products and variants. Created three variants and set different pics for each - All three variants appear on the shopfront, but with the picture of the "main variant".

image

It's possible to delete these, with exception of the last one which, as expected, triggers the modal "Attempting to delete the last variant triggers the modal " The last variant cannot be deleted!"

All good - ready to go.

PR #5545, opened 2020-06-02
Add configuration to enable Matomo Tag Manager

### LINKED ISSUE #5510: Add configuration for Matomo Tag Manager
## Description
As an administrator of an instance I want to be able to add Matomo Tag Manager to the instance, so that I can add analytics events without development work

This was triggered by this problem that needs more data to help make product decisions. Adding events for this scenario manually will likely be as much work as adding support for MTM, and MTM will be enable n on developers to curate more events in future.

1. In Configuration > Matomo Settings a field will be available to ad the Matomo Tag Manager URL
2. Once this field is added, Matomo Tag Manager code will be embedded in the page, and a Matomo tracking code will not (if specified in Matomo Settings). Matomo Tag Manager overrides Matomo.
3. In the Matomo Tag Manager container, added Triggers and Tags will fire from the OFN instance as pages are navigated, forms submitted and elements clicked

mbudm:
#### What? Why?

Closes #5510

This adds a field to the Matomo configuration screen in the Admin UI. It lets a Matomo Tag Manager URL be set.

When the Matomo Tag Manager URL is set, the Matomo Tag Manager embed code is added to every page in the app.

Originally I thought that it would be best to have the Matomo Tag Manager setting override the existing Matomo Analytics setting. However I have found in testing that a combination of the two works more reliably. Therefore I have changed this so that both embed codes can be added if needed.

I have however updated the Matomo Analytics embed code to match what is now suggested by Matomo. Mostly this is the name change from piwik to matomo. Although the old snippet still works I imagine it is a good idea to keep up with the latest.

#### What should we test?
If testing locally, tracking is blocked so firstly you need to change your local host to a custom 'domain':

```rb
# config/environments/development.rb
Rails.application.routes.default_url_options[:host] = 'test.ofn'
```
```sh
# /etc/hosts
127.0.0.1  test.ofn
```
Then:
- Add a Matomo Tag Manager URL to the admin setting. (I suggest using the OFN AUSTRALIA DEV container as this has triggers and tags set up, or you can import this into your Matomo site:
[container_dmE1FP2S_Remove pageviews.json.zip](https://github.com/openfoodfoundation/openfoodnetwork/files/4714582/container_dmE1FP2S_Remove.pageviews.json.zip))
- Add a Matomo Analytics URL and Site ID (if one isn't set already)
- Navigate through the checkout process a couple of times, triggering an error (expired CC is easiest), and also doing a successful checkout resulting in a success banner

You should be able to see the events and pageviews in the Matomo analytics screen - Visitor Log.
Screen Shot 2020-06-02 at 10 07 52 am

#### Release notes
Added support for Matomo Tag Manager

Changelog Category: Added 

#### Discourse thread
Related directly to this discussion, where the need for detailed analytics was raised: https://community.openfoodnetwork.org/t/secure-products-availability-during-checkout-process/1933/13
Related indirectly, as with MTM events can be created anywhere: https://community.openfoodnetwork.org/t/enable-matomo-to-measure-backoffice-and-actions/1554

#### Documentation updates
A guide to using Matomo Tag Manager will be needed (TBD)


app/views/layouts/_matomo_tag.html.haml
```diff
@@ -1,15 +1,23 @@
+
+- if Spree::Config.matomo_tag_manager_url.present?
+  :javascript
+    var _mtm = _mtm || [];
+    _mtm.push({'mtm.startTime': (new Date().getTime()), 'event': 'mtm.Start'});
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    var u="#{Spree::Config.matomo_tag_manager_url}";
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u; s.parentNode.insertBefore(g,s);
 - if Spree::Config.matomo_url.present?
```
sauloperez, line 9: I'm being a bit picky here, but, with haml's lack of curly braces etc., I think a blank line between the snippet above and the `if` would make things more readable. We'll then see at a glance that these are two separate snippets.

app/views/layouts/_matomo_tag.html.haml
```diff
@@ -1,15 +1,23 @@
...
     (function() {
       var u="#{Spree::Config.matomo_url}";
-      _paq.push(['setTrackerUrl', u+'piwik.php']);
+      _paq.push(['setTrackerUrl', u+'matomo.php']);
       _paq.push(['setSiteId', '#{Spree::Config.matomo_site_id}']);
       var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
-      g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
-    })();
+      g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
+    })();
```
sauloperez, line 23: Github is complaining about a missing newline :warning:

app/views/layouts/_matomo_tag.html.haml
mbudm: np

app/views/layouts/_matomo_tag.html.haml
mbudm: Sorted

app/views/layouts/_matomo_tag.html.haml
```diff
@@ -1,15 +1,24 @@
+
+- if Spree::Config.matomo_tag_manager_url.present?
+  :javascript
+    var _mtm = _mtm || [];
+    _mtm.push({'mtm.startTime': (new Date().getTime()), 'event': 'mtm.Start'});
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    var u="#{Spree::Config.matomo_tag_manager_url}";
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u; s.parentNode.insertBefore(g,s);
+
```
luisramos0, line 9: This new setting, matomo_tag_manager_url, will never work without matomo_url being set, right?
I'd put this code and conditional **below** the basic matomo tag/conditional.

If matomo_url is not set and the tag_manager is set, I wonder if it doesnt blow up the js code on the browser. In that case I'd put this matomo_tag_manager_url code/condition **inside** the matomo_url conditional.

app/views/layouts/_matomo_tag.html.haml
```diff
@@ -1,15 +1,24 @@
...
+    var _paq = window._paq || [];
     _paq.push(["setDocumentTitle", document.domain + "/" + document.title]);
     _paq.push(["setCookieDomain", "*.#{Spree::Config.site_url}"]);
     _paq.push(["setDomains", ["*.#{Spree::Config.site_url}"]]);
     _paq.push(['trackPageView']);
     _paq.push(['enableLinkTracking']);
     (function() {
       var u="#{Spree::Config.matomo_url}";
-      _paq.push(['setTrackerUrl', u+'piwik.php']);
+      _paq.push(['setTrackerUrl', u+'matomo.php']);
```
luisramos0, line 20: Thanks! When I did this they had just renamed from piwik :+1:

config/locales/en.yml
```diff
@@ -415,8 +415,10 @@ en:
...
+        matomo_tag_manager_url: "Matomo Tag Manager URL"
+        info_html: "Matomo is a Web and Mobile Analytics application. You can either host Matomo on-premises or use a cloud-hosted service. See <a href='http://matomo.org' target='_blank'>matomo.org</a> for more information."
         config_instructions_html: "Here you can configure the OFN Matomo integration. The Matomo URL below should point to the Matomo instance where the user tracking information will be sent to; if it is left empty, Matomo user tracking will be disabled. The Site ID field is not mandatory but useful if you are tracking more than one website on a single Matomo instance; it can be found on the Matomo instance console."
+        config_instructions_tag_manager_html: "Setting the Matomo Tag Manager url enables Matomo Tag Manager. This tool lets non developers set up analytics events. The Matomo Tag Manager URL is copied from the Install Code section of the Matomo Tag Manager. Ensure you select the right container and environment as these options change the URL."
```
luisramos0, line 421: URL should be uppercase
"the Matomo Tag Manager URL enables"

I dont see the need to indicate who can do it, it could simply be:
"This tool lets you set up analytics events."

luisramos0:
Moving to In Dev until dev feedback is given (just to keep Code Review column with the minimum number of PRs, there are lot of them in there these days).

app/views/layouts/_matomo_tag.html.haml
mbudm: Hi @luisramos0 after further investigation I changed the approach to allowing both scripts at the same time - covered in the PR description:

> Originally I thought that it would be best to have the Matomo Tag Manager setting override the existing Matomo Analytics setting. However I have found in testing that a combination of the two works more reliably. Therefore I have changed this so that both embed codes can be added if needed.

config/locales/en.yml
mbudm: Ok

mbudm:
Further updates:
- fixed copy to use correct case and non user specific
- checked that cookies still match policy (documented in PR detail)
- added screenshots showing changes to admin screen (documented in PR detail)
- clarified that scripts can be run concurrently

luisramos0:
thanks for your feedback Steve!

We havent discussed as a group yet but I think it's better that after the initial review, the author of the PR doesnt force push. In this case, it's not clear what was changed since my last review. I believe it was only the text of the translation, right?

It's a minor thing but UX wise I think it would be better to have the text below the tag manager input box, this way it looks more related to the fields above. What do you think?
image

I should have pinged @RachL here actually.
@RachL It can be done in a separate PR but do you think we should or need to change the cookies policy text to indicate instances are using tags? Or do you think we will be fine as we are?

sauloperez:
Agree with @luisramos0 's comment about UX :point_up:

app/views/layouts/_matomo_tag.html.haml
```diff
@@ -1,15 +1,24 @@
+
+- if Spree::Config.matomo_tag_manager_url.present?
+  :javascript
+    var _mtm = _mtm || [];
+    _mtm.push({'mtm.startTime': (new Date().getTime()), 'event': 'mtm.Start'});
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    var u="#{Spree::Config.matomo_tag_manager_url}";
```
Matt-Yorkley, line 7: Would it be possible to use human-readable variable naming here instead of single-letter variables?

RachL:
@luisramos0 one thing that is not clear to me is: are we introducing new cookies the user will be able to switch off here or not? If yes then I think we need to update the cookie banner to let them choose whether or not they want their actions to be tracked. They could choose the general tracking that does not include tags and therefor actions...?

luisramos0:
>> are we introducing new cookies the user will be able to switch off here or not?

No, we are not introducing new ccokies, but we will be gathering more behavioral data from the users.

>> They could choose the general tracking that does not include tags and therefor actions...?

Yes, users could opt out of this tags. I think the question is: do we need to implement the opt out or is the basic opt-out enough? I read a bit about this now and it looks like we don't need anything else, but I am not sure.

app/views/layouts/_matomo_tag.html.haml
mbudm: We could do but I kind of like keeping it close to the provided embed code, as that way we can more easily see if it needs updating - eg that made it easier for me to spot that filenames of piwik had changed to matomo.

mbudm:
@RachL and @luisramos0 I've looked into the opt out iframe that we use for Matomo and this is a big switch that turns of all Matomo tracking, it has no option for choosing more granular settings like allowing pageviews and forms (currently tracked by the main analytics tool) or events that we are now going to add via the Tag Manager.

However, everything that we intend to do via the tag manager we can also do using the  basic Matomo analytics tracking, it just takes longer as we have to make a code change each time.

Therefore I think the current opt out option should remain unchanged

mbudm:
On the UX, I played around with adding sections, like some of the other settings pages, this sort of made sense as I was thinking about Matomo and Matomo Tag Manager as two separate tools. However the more I read, the more I realise that isnt the case. Tag Manager is really just an add on to the main Matomo Analytics tool.
Screen Shot 2020-06-10 at 2 41 30 pm

So I prefer what @luisramos0 and @sauloperez suggested, simply moving the MTM instruction below that field. I have pushed this change has to this PR:
Screen Shot 2020-06-10 at 3 25 01 pm

mbudm:
@filipefurtad0 - great. It looks like you have pageviews, they are represented by the folder icon in the real time view. The visits log detail which pages you visit.
Screen Shot 2020-06-11 at 9 43 32 am

To test tracking events, you need to ad some to a Matomo Tag Manager container. I've attached a container zip to this PR you can use that tracks the appearance of alerts (success/error) on pages.
https://github.com/openfoodfoundation/openfoodnetwork/files/4714582/container_dmE1FP2S_Remove.pageviews.json.zip

You import this in the Matomo Tag Manager after creating a container. Versions -> import.
Screen Shot 2020-06-11 at 9 46 31 am

filipefurtad0:
Hey @mbudm,

Thanks for the container file - I was not finding the place to upload it yesterday, so I set events and triggers manually in the the Tag Container, but was seeing nothing under "View detailed visit log". I see something today already, so it needed some time I think.

I uploaded the file you shared today, and sent it live, so I'm hoping to see some triggered errors later on. Thank you for your help!

PR #5549, opened 2020-06-02
[Spree 2.1] Make Variant Overrides spec less flaky and... comment it anyway

### LINKED ISSUE #4957: [Flaky spec] Fix features/admin/variant_overrides_specs currently commented out (2 specs)
spec/features/admin/variant_overrides_spec.rb:193
spec/features/admin/variant_overrides_spec.rb:210

This spec is not failing all the time #flaky

luisramos0:
#### What? Why?

I improved the spec. It is still falling. I commented it in the end.
This enables us to continue with the upgrade.
We have issue #4957 to fix the spec later after the upgrade.

#### What should we test?
green spec.

### DIFF
--- a/spec/features/admin/variant_overrides_spec.rb
+++ b/spec/features/admin/variant_overrides_spec.rb
@@ -199,6 +199,9 @@ feature "
 
             expect do
               click_button 'Save Changes'
+
+              # We need to wait_until because the save action is not fast enough for the have_content matcher
+              wait_until { page.find("#status-message").text != "Saving..." }
               expect(page).to have_content "I couldn't get authorisation to save those changes, so they remain unsaved."
             end.to change(VariantOverride, :count).by(0)
           end
--- a/spec/features/admin/variant_overrides_spec.rb
+++ b/spec/features/admin/variant_overrides_spec.rb
@@ -195,7 +195,8 @@ feature "
             fill_in "variant-overrides-#{variant.id}-count_on_hand", with: '123'
             expect(page).to have_content "Changes to one override remain unsaved."
 
-            user.enterprises.clear
+            # Set a user without suficient permissions
+            allow_any_instance_of(Spree::Admin::BaseController).to receive(:current_spree_user).and_return(build(:user))
 
             expect do
               click_button 'Save Changes'
--- a/spec/features/admin/variant_overrides_spec.rb
+++ b/spec/features/admin/variant_overrides_spec.rb
@@ -190,7 +190,7 @@ feature "
             end
           end
 
-          it "displays an error when unauthorised to access the page" do
+          xit "displays an error when unauthorised to access the page" do
             fill_in "variant-overrides-#{variant.id}-price", with: '777.77'
             fill_in "variant-overrides-#{variant.id}-count_on_hand", with: '123'
             expect(page).to have_content "Changes to one override remain unsaved."


sauloperez:
The build failed on that spec

```
Failure/Error: expect(page).to_not have_content "Saving..."
       expected not to find text "Saving..." in "Logged in as : domenica@welchtreutel.com Account Logout Store DASHBOARD PRODUCTS ORDER CYCLES ORDERS REPORTS ENTERPRISES CUSTOMERS PRODUCTS INVENTORY IMPORT Inventory What's this? QUICK SEARCH SHOP Enterprise 31 Enterprise 31 PRODUCER All IMPORTED All There are 2 new products available to add to your inventory. × ACTIONS VIEWING: INVENTORY PRODUCTS COLUMNS Saving... PRODUCER PRODUCT PRICE ON HAND ON DEMAND? Enterprise 33 Product #31 - 3751 1g Use producer stock settings Yes No Enterprise 35 Product #33 - 903 2g, S Use producer stock settings Yes No"
```

Could it be that CI is too slow for the configured Capybara timeout? `have_content` is one of the matchers listed https://rubydoc.info/github/jnicklas/capybara/master/Capybara/RSpecMatchers,  which wait for the result.

luisramos0:
ah, yeah, is there a way we can set rspec to wait a little longer on this specific validation?

sauloperez:
I've never done that, but there must be a way to provide an argument to the matcher, somehow :thinking:

Matt-Yorkley:
I think there's a custom method called `wait_until` that we use in one or two cases like this one...

luisramos0:
ah, nice, giving it a try :crossed_fingers:

luisramos0:
looks good :+1: trying again.

Matt-Yorkley:
Still good. I'll run it once more and we can check again.

spec/features/admin/variant_overrides_spec.rb
```diff
@@ -199,6 +199,9 @@
 
             expect do
               click_button 'Save Changes'
+
+              # We need to wait_until because the save action is not fast enough for the have_content matcher
+              wait_until { page.find("#status-message").text != "Saving..." }
```
sauloperez, line 204: the only thing I know is that this method was long discouraged and it seems it got removed in Capybara 2.0 according to  https://groups.google.com/forum/#!topic/ruby-capybara/qQYWpQb9FzY, but to my surprise, we added it back in

https://github.com/openfoodfoundation/openfoodnetwork/blob/12000d2d4482936794f040d50dfaf5c6da5239f7/spec/support/request/web_helper.rb#L90-L96

This solution is rather similar to https://thoughtbot.com/blog/automatically-wait-for-ajax-with-capybara. It seems to me that the root cause might be the same, so :+1:

luisramos0:
:-( broken:
image
something else is wrong? I think this means it waited 30 secs for it...

sauloperez:
Could be something blocking that ajax request?

luisramos0:
if the next build is green maybe we should go ahead with this change anyway as it makes it better?

luisramos0:
damn it, broken again. What do we do?

luisramos0:
ok, I think I have a more reliable solution, I think the problem is related to the user object in the controller not being reloaded sometimes.

sauloperez:
It was agreed that we'll go with this to unlock the merge into master and I'll check a better fix later on, so merging.

PR #5552, opened 2020-06-03
Remove unused stock photo

sauloperez:
#### What? Why?

I found a couple of photos with a watermark, which made me think they are not used. It turns out there a few more for which a `git grep` returned nothing.

#### What should we test?

We should check in the home section that all pages show correctly and don't miss a background or regular image.

#### Release notes

Unused photos meant to be used in the home.

Changelog Category: Removed


sauloperez:
These are the ones I removed:

home1
home2
home3
home-oranges
home-strawberries
macbook

filipefurtad0:
Hey @sauloperez,

All as before. The groups and show the usual backgrounds and so does the homepage, which has the second pic from your list as a background, in higher resolution (so, it's another file).

All good here - thanks!

PR #2916, opened 2018-10-23
Add Portuguese and German to list of momentJS locales

luisramos0:
The main usage is in the order cycle closure time in the shopfront.

#### What? Why?

Portuguese katuma shopfronts have the order cycle closure time indication in english. This fixes the problem.
German was also missing. Now fixed:
screen shot 2018-10-23 at 17 57 58

#### What should we test?
In Portuguese or German, the shopfront message should not display "in x days" but instead a localed message.

#### Release notes
Changelog Category: Fixed
Fixed Portuguese and German translation problem in the shopfront order cycle closure time.

#### How is this related to the Spree upgrade?
not related.

PR #5554, opened 2020-06-03
Add Catalan to the list of moment.js files

sauloperez:
#### What? Why?

This will precompile the Catalan language file as well and thus, things like 

Screenshot from 2020-06-03 18-14-28

should be shown in Catalan.

#### What should we test?

Check that OC timings are displayed in Catalan instead of English.

#### Release notes

Display time-related copies in Catalan when that's the selected language.

Changelog Category: Fixed

### DIFF
--- a/app/assets/javascripts/admin/all.js
+++ b/app/assets/javascripts/admin/all.js
@@ -73,6 +73,7 @@
 //= require i18n/translations
 //= require darkswarm/i18n.translate.js
 //= require moment
+//= require moment/ca.js
 //= require moment/de.js
 //= require moment/en-gb.js
 //= require moment/es.js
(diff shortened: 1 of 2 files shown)


sauloperez:
This showed up in the console while checking /admin/order_cycles

```
Error parsing the date/time string: Wrong time format
date/time string = 2020-03-03 00:00:00 +0100
timeFormat = HH:mm
dateFormat = yy-mm-dd
```

sauloperez:
Yes, I thought the same! I just checked their GH repo and it looks like they support tons of languages but first we need to understand why this is not working as is yet. Do you know where is it really used? I just guessed the OC selector I show in the screenshot was one of them.

luisramos0:
this is the solution for the translation problem in the OC selector, I did this for pt and de before and it worked:
https://github.com/openfoodfoundation/openfoodnetwork/pull/2916

sauloperez:
Superseded by #5807

Issue #5542, opened 2020-06-01
Google-free maps


Matt-Yorkley:
I'm adding this epic in the backlog to track any bits people do to move towards OFN being independent of Google Maps. Contributions welcome!

We're making a big leap forwards with this PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/5398

There are still various bits that would need to be done before we can start migrating our current production servers to alternate providers.

- [ ] Clarify which other parts of the app depend on Google Maps JS (there are some other parts) for geocoding in javascript
- [ ] Make some alternative javascript services for geocoding, via some new API endpoints that use the Geocoder gem (which is already used elsewhere in the app, and can be easily configured with alternate providers)
- [ ] Improve map page searchbar functionality. The current OSM implementation only searches by names of OFN enterprises, not locations. This would need require the new non-google geocoding service, and some simple method calls to recenter the map on the found location.
- [ ] OSM map is totally blank (grey) when no enterprises are present. Maybe we could show the map and a message?

sigmundpetersen:
Could/should #3179 be attached to this work? (Upgrading the geocoder dependency)

diegoamc:
Hi!

What is the status of this issue? I would like to help if possible.

Is there a task list on what's left to be done?

RachL:
Hi @Zoyolin thanks for your comment 🤗 

Gmaps was used first as 11 years ago in Australia (when OFN was launched)  OSM was not accurate / used.

Our community would love to have OSM working now but as you can see there is still a lot to do, so unless a volunteer picks this works up, we need to find funding, which is a long process.

Zoyolin:
Thanks @RachL. Sorry I didn't realise how global OFN is and how much effort has already been put together to support OSM. I realised Ireland and Belgium's default is already OSM from leaflet. Indeed Australia, UK, France and USA are on ABC's map from Google.

RachL:
@Zoyolin yes it's a setting you can access as super admin in `/admin/contents/edit`

But main issue with open street map currently is that the implementation let you search for enterprises, not location (unless you use the scroll). So if you don't know the enterprise name, the search is a bit difficult (although the address of the enterprise shows up in results as well if you type a city for example).

filipefurtad0:
> yes it's a setting you can access as super admin at the bottom of the page /admin/contents/edit

An additional hint on this: it seems that enabling this option breaks the map around the first step of the enterprise registration process, at `/register`:

[...]

PR #5556, opened 2020-06-04
Rename Angular Geo service

Matt-Yorkley:
Related to #5542 

#### What? Why?

Renames our Angular `Geo` service to `GmapsGeo` for clarity, as it depends on direct calls to the google maps javascript included in the darkswarm layout. 

This is another small step towards moving away from dependency on Google for maps functionality. We'll need an alternate Angular service that provides the same functionality by different means. See notes in the epic: #5542 

#### What should we test?

Sanity check that maps is working and searching in the /shops page returns results by location, eg "London".

#### Release notes

Renamed Angular "Geo" service to "GmapsGeo"

### DIFF
--- a/app/assets/javascripts/darkswarm/services/enterprises.js.coffee
+++ b/app/assets/javascripts/darkswarm/services/enterprises.js.coffee
@@ -1,4 +1,4 @@
-Darkswarm.factory 'Enterprises', (enterprises, ShopsResource, CurrentHub, Taxons, Dereferencer, Matcher, Geo, $rootScope) ->
+Darkswarm.factory 'Enterprises', (enterprises, ShopsResource, CurrentHub, Taxons, Dereferencer, Matcher, GmapsGeo, $rootScope) ->
   new class Enterprises
     enterprises: []
     enterprises_by_id: {}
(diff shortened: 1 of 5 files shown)


filipefurtad0:
Hey @Matt-Yorkley ,

Sanity check for maps shows nothing unusual, all good. Searching for a location works - as long as there is an enterprise located there. This was introduced by OpenStreeMaps, so all good.
Ready to go.

filipefurtad0:
Hey @Matt-Yorkley ,

I've staged this again and checked now the functionality you describe and couldn't verify that in the shops page the search by location is possible. It does seem to be the case in production, so I think this PR is introducing this:

search_OSM

I see this was merged already: I had updated the comment above, but this was not visible to @sauloperez. Perhaps we should revert this merge and move this back to dev?

sauloperez:
ups, thank God you mentioned me now :sweat_smile: . Reverting

Matt-Yorkley:
@filipefurtado0 was the openstreetmap setting turned off again in admin config after the previous tests?

Matt-Yorkley:
Yep, OpenStreetMap was still enabled on Katuma Staging:

Screenshot from 2020-06-11 16-04-50

I just disabled it and the location searching works :+1:

Matt-Yorkley:
Ironically the point of this PR is to highlight which bits of functionality on the site will be broken when `OpenStreetMaps` is enabled and `GoogleMaps` is disabled. :trollface:

filipefurtad0:
Ok - awesome @Matt-Yorkley,
Sorry for the false alarm! I tthought it would be best to mention this asap ;-)

So all is good and no need to revert :+1:

Matt-Yorkley:
> Sorry for the false alarm! I thought it would be best to mention this asap

Definitely! I hadn't thought about the settings, sorry I didn't mention it in the testing notes.

Thanks again! :heart:

PR #5557, opened 2020-06-04
Improve performance in various reports

Matt-Yorkley:
#### What? Why?

Discovered by accident when rummaging around in reports code.

Brings a massive performance improvement to various reports, when run by a user with access to a large amount of enterprises and products.

#### What should we test?

Affects the following report types:
- customers
- products_and_inventory
- order_cycle_management
- packing

#### Release notes

Improved performance on various reports


Matt-Yorkley:
Results using production data, user with permissions for a large number of enterprises and products:

**When just loading the page, not even submitting the form to generate a report (before and after):**

Screenshot from 2020-06-04 12-03-57

Screenshot from 2020-06-04 12-00-28

**When actually submitting the form, with around 300 rows returned (before and after):**

Screenshot from 2020-06-04 12-04-39

Screenshot from 2020-06-04 12-01-26

So that's around 80-85% reduction in page load...

app/controllers/spree/admin/reports_controller.rb
```diff
@@ -247,8 +247,10 @@ def my_suppliers
       end
 
       def suppliers_of_products_distributed_by(distributors)
-        distributors.map { |d| Spree::Product.in_distributor(d).includes(:supplier).all }.
-          flatten.map(&:supplier).uniq
+        supplier_ids = Spree::Product.in_distributors(distributors).
+          select('spree_products.supplier_id')
+
+        Enterprise.where(id: supplier_ids)
```
sauloperez, line 253: I'm assuming you use this as a trick to get unique enterprises. Why don't you ask that the DB exactly that in the query above? grouping by supplier id replaces the lack of `#distinct` in Rails 3.2.

app/models/spree/product_decorator.rb
```diff
@@ -84,6 +84,12 @@
       select('distinct spree_products.*')
   }
 
+  scope :in_distributors, lambda { |distributors|
```
sauloperez, line 87: but `#in_distributors` was used already. Was it a method instead of a scope? Won't we have two conflicting definitions?

Matt-Yorkley: There's an `in_distributor` scope, but it only takes a single object. This one takes a relation, with multiple distributors.

app/controllers/spree/admin/reports_controller.rb
Matt-Yorkley: It's about passing an active record relation to the subquery, instead of querying distributors, then using #map on the result, then running queries for each one in the list.

app/models/spree/product_decorator.rb
luisramos0: Isnt this copy-paste-free version better?
```ruby
  # Products distributed via the given distributor through an OC
  scope :in_distributor, lambda { |distributor|
    distributor = distributor.respond_to?(:id) ? distributor.id : distributor.to_i

    in_distributor_base(distributor).
      select('distinct spree_products.*')
  }

  scope :in_distributors, lambda { |distributors|
    in_distributor_base(distributors).uniq
  }

  scope :in_distributor_base, lambda { |distributors|
    with_order_cycles_outer.
      where('(o_exchanges.incoming = ? AND o_exchanges.receiver_id IN (?))', false, distributors)
  }

```

PR #5559, opened 2020-06-04
Make display only shops work again

### LINKED ISSUE #5423: Display only shopfronts not displaying products
## Description
A display only shopfront arises from an OC where shipping and payment methods have been disabled.  No customer can checkout,  but the shopfront still displays available products as per the OC.    In OFN-CAN we are using these with farmers' markets to give customers a dynamic list of what is available in the market that week.  Customers shop from individual vendors.

It was working fine -  I had 6 markets set up with display only shops for the season.  But this morning they have all called -  their display shop is no longer working.   I did a test,  and indeed,  I can no longer create a display only shop.  This has happened,  I believe, since the last release.

## Expected Behavior
Above -  disabling shipping methods and payment methods in an order cycle should result in display only for an active OC with a visible shop. 

## Actual Behaviour
Display only shop shows as a closed shop -  eventhough OC is active and enterprise is visible.

## Steps to Reproduce
Set up an OC for a hub 'normally' (with shipping & payment methods)
Check that its visible and works.
Disable the hub's shipping and payment methods.  
See the pop up 'warning'  that the hub has no active shipping & payment methods
See that the shop is now showing as closed

Open shop:

demo market with payment and shipping methods open

After shipping and payment methods have been disabled:

display only shop does not open

Animated Gif/Screenshot

## Workaround
I know of no workaround

## Severity
Help with this?   There is no workaround.  We have users set up expecting this as a feature.  S2?

-->

## Your Environment
desktop - chrome - latest

* Version used:
* Browser name and version:

luisramos0:
#### What? Why?

Closes #5423

When a distributor has an open OC but it doesnt have valid ship/pay methods, the shop should still render.
It's a simple change that was reverted recently, I added a feature spec to validate the current behaviour.

We could improve the UX, for example, not letting user add to cart or don't redirect user to /shops but keep user in shop page. We agreed we would just add the feature back as it was before because it was only an half a hour job :heavy_check_mark: 

#### What should we test?
We need to test the shop front renders correctly in different scenarios:
- when there are no open OCs
- when there is an open OC but it's hidden by some tag
- when the shop requires login
- (verify issue) when there is an open OC but the shop doesnt have ship/payment methods defined
In this last case, products can be added to the cart, but when user goes to the cart or checkout, it should be redirected to shops saying the shop is closed.

#### Release notes
Changelog Category: Changed
Make display only shops work again

### DIFF
--- a/spec/features/consumer/shopping/shopping_spec.rb
+++ b/spec/features/consumer/shopping/shopping_spec.rb
@@ -232,6 +232,28 @@ feature "As a consumer I want to shop with a distributor", js: true do
           expect(page).not_to have_content variant3.display_name
         end
       end
+
+      context "when the distributor has no available payment/shipping methods" do
+        before do
+          distributor.update_attributes shipping_methods: [], payment_methods: []
+        end
+
+        # Display only shops are a very useful hack that is described in the user guide
+        it "still renders a display only shop" do
+          visit shop_path
+          expect(page).to have_content product.name
+
+          # Add product to cart
+          fill_in "variants[#{variant.id}]", with: '1'
+          wait_for_debounce
+          expect(page).to have_in_cart product.name
+          wait_until { !cart_dirty }
+
+          # Try to go to cart
+          visit main_app.cart_path
+          expect(page).to have_content "The hub you have selected is temporarily closed for orders. Please try again later."
+        end
+      end
     end
 
     describe "group buy products" do
@@ -463,11 +485,13 @@ feature "As a consumer I want to shop with a distributor", js: true do
         visit shop_path
         expect(page).to have_content "Orders are closed"
       end
+
       it "shows the last order cycle" do
         oc1 = create(:simple_order_cycle, distributors: [distributor], orders_open_at: 17.days.ago, orders_close_at: 10.days.ago)
         visit shop_path
         expect(page).to have_content "The last cycle closed 10 days ago"
       end
+
       it "shows the next order cycle" do
         oc1 = create(:simple_order_cycle, distributors: [distributor], orders_open_at: 10.days.from_now, orders_close_at: 17.days.from_now)
         visit shop_path
--- a/app/views/shopping_shared/tabs/_shop.html.haml
+++ b/app/views/shopping_shared/tabs/_shop.html.haml
@@ -4,7 +4,8 @@
 
   - if no_open_order_cycles?
     = render partial: "shop/messages/closed_shop"
-
   - else
     = render partial: "shop/messages/select_oc"
-    = render partial: "shop/products/form"
+
+  -# Rendering the form, even if there are no open OCs, makes display only shops possible
+  = render partial: "shop/products/form"


Matt-Yorkley:
So what happens when there are shipping or payment methods but the OC is closed? Do we still see the product list partial for closed shops?

Matt-Yorkley:
Can we set a helper method like `display_only_shop?` that returns a boolean, and maybe disable the add to cart buttons using that?

luisramos0:
~~ So what happens when there are shipping or payment methods but the OC is closed? Do we still see the product list partial for closed shops?

no, the OC is closed, no products will render.


~~ Can we set a helper method like display_only_shop? that returns a boolean, and maybe disable the add to cart buttons using that?

Yes but that would be more work and code. I'd would just leave it as it was before. And add a simple spec to validate the case.

spec/features/consumer/shopping/shopping_spec.rb
```diff
@@ -232,6 +232,28 @@
...
+          # Add product to cart
+          fill_in "variants[#{variant.id}]", with: '1'
+          wait_for_debounce
+          expect(page).to have_in_cart product.name
+          wait_until { !cart_dirty }
+
+          # Try to go to cart
+          visit main_app.cart_path
+          expect(page).to have_content "The hub you have selected is temporarily closed for orders. Please try again later."
+        end
```
sauloperez, line 255: can't we check the feature from a controller spec? if this is about showing the products form when there are no open OCs it should. You know, the "inverted" testing pyramid we have always concerned me.

luisramos0: not in this case, this is really a feature. It would be somewhat testable in a view spec but I think we should go with feature spec to make sure it works as described in the user guide.

Matt-Yorkley:
I think I'll move this along the pipe as I know @tschumilas is desperate to get this functionality back. Assuming @sauloperez is happy with that..?

tschumilas:
Thank-you thank-you thank-you @Matt-Yorkley !

filipefurtad0:
Hi @luisramos0 ,

I've verified the issue and went through the test cases:

- when there are no open OCs and
- when there is an open OC but it's hidden by some tag

The shopfront is closed:
image

- when the shop requires login the shop is closed with message below:
image

- setting payment methods as backend only - it is not possible to delete **all** payment/delivery methods: more on this below - makes the warning pop up, in orders, dashboard, (which is expected):
image

and allows verifying that the PR solves the issue: when there is an open OC but with no ship/payment methods - set as backoffice only - the shopfront is visible:
image

It's possible to add items to the cart, but proceeding to checkout redirects to the homepage and shows the modal:
image

According to your test notes this is not the expected behaviour -> redirection to homepage instead of shops. Should I open an issue for this?

In any case, the PR solves the underlying issue and is good to go.

Thank you Luís - also for the detailed testing notes!

PS: I've noticed that once delivery/payment methods are set, the last one cannot be deleted - no warning, or error. Removing the last tick and updating just does not work. The alternative is using backend only - used for testing here. I'm wondering if this intended to be this way. Maybe an improvement would be a warning to let the user know the consequences of removing the last method?

luisramos0:
Awesome!


>>>   According to your test notes this is not the expected behaviour -> redirection to homepage instead of shops. Should I open an issue for this?

The best redirect would be to the shopfront actually. I would ignore this for now.

>>> PS: I've noticed that once delivery/payment methods are set, the last one cannot be deleted - no warning, or error. Removing the last tick and updating just does not work. The alternative is using backend only - used for testing here. I'm wondering if this intended to be this way. This is not introduced by this PR or the v3 upgrade. Maybe an improvement would be a warning to let the user know the consequences of removing the last method?

oh, there's no way to remove the last payment method from an enterprise? @tschumilas needs to know about this. To have a display only sho now we need to make the last method backend only.
I'd create an issue for this (I'd guess S3), I think the user should be able to remove the last method or at least be warned.

PR #5564, opened 2020-06-07
Improve Dockerfile

dmarcoux:
#### What? Why?
I want to start contributing to the project and while having a look at the Docker setup, I've found some minor improvements.

#### What should we test?
The Docker setup, so building the Docker-Compose `web` service with `docker-compose build web`. This will rebuild the image from the Dockerfile.

#### Release notes
Shallow clone specific release for rbenv and ruby-build and run bundler install in parallel with the amount of available CPUs.

Changelog Category: Changed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,8 @@ WORKDIR /usr/src/app
 COPY .ruby-version .
 
 # Install Rbenv & Ruby
-RUN git clone https://github.com/rbenv/rbenv.git ${RBENV_ROOT} && \
-    git clone https://github.com/rbenv/ruby-build.git ${RBENV_ROOT}/plugins/ruby-build && \
+RUN git clone --depth 1 --branch v1.1.2 https://github.com/rbenv/rbenv.git ${RBENV_ROOT} && \
+    git clone --depth 1 --branch v20200520 https://github.com/rbenv/ruby-build.git ${RBENV_ROOT}/plugins/ruby-build && \
     ${RBENV_ROOT}/plugins/ruby-build/install.sh && \
     echo 'eval "$(rbenv init -)"' >> /etc/profile.d/rbenv.sh && \
     rbenv install $(cat .ruby-version) && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,4 +43,5 @@ RUN wget https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.z
 
 # Copy code and install app dependencies
 COPY . /usr/src/app/
-RUN bundle install
+# Run bundler install in parallel with the amount of available CPUs
+RUN bundle install --jobs="$(nproc)"


Dockerfile
```diff
@@ -13,8 +13,8 @@ WORKDIR /usr/src/app
 COPY .ruby-version .
 
 # Install Rbenv & Ruby
-RUN git clone https://github.com/rbenv/rbenv.git ${RBENV_ROOT} && \
-    git clone https://github.com/rbenv/ruby-build.git ${RBENV_ROOT}/plugins/ruby-build && \
+RUN git clone --depth 1 --branch v1.1.2 https://github.com/rbenv/rbenv.git ${RBENV_ROOT} && \
+    git clone --depth 1 --branch v20200520 https://github.com/rbenv/ruby-build.git ${RBENV_ROOT}/plugins/ruby-build && \
```
luisramos0, line 17: there's the downside of having to update these versions later, I wonder if this makes the process significantly faster. If it does, let's keep this :+1:

Dockerfile
```diff
@@ -43,4 +43,5 @@ RUN wget https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.z
 
 # Copy code and install app dependencies
 COPY . /usr/src/app/
-RUN bundle install
+# Run bundler install in parallel with the amount of available CPUs
+RUN bundle install --jobs="$(nproc)"
```
luisramos0, line 47: nice one!
I think we can use this one in ofn-install (our ansible magic toolbox to manage our servers) here:
https://github.com/openfoodfoundation/ofn-install/blob/eb6ed5b449e5ad676ff9601648119bb9eacc6b13/roles/deploy/tasks/deploy.yml#L55

Dockerfile
dmarcoux: Cloning a specific release with `--branch` does have the downside of having to change it when we need to update to the latest release. On the other side, this makes the installation of those tools reproducible since we're always building the same version.

As for the speed, it's not relevant when we're talking about `--branch`. This is when `--depth 1` comes into play. I've put below some numbers after timing the git clone commands with `time`.

For ruby-build:
```
$ time git clone --quiet https://github.com/rbenv/ruby-build.git 
0.50s user 0.14s system 15% cpu 4.053 total (~4 seconds)

$ time git clone --quiet --depth 1 https://github.com/rbenv/ruby-build.git
0.06s user 0.05s system 6% cpu 1.785 total (~1.7 seconds)
```

For rbenv:
```
$ time git clone --quiet https://github.com/rbenv/rbenv.git 
0.15s user 0.07s system 15% cpu 1.421 total (~1.4 seconds)

$ time git clone --quiet --depth 1 https://github.com/rbenv/rbenv.git 
0.05s user 0.05s system 6% cpu 1.438 total (~1.4 seconds)
```

So as we can see here, using `--depth 1` does make the cloning of `ruby-build` a bit faster. For `rbenv`, it's pretty much the same since the repository doesn't have many commits.

luisramos0: ah, ok, I agree, it's better to always use the same version :+1:

sauloperez:
everything working as usual :+1:

PR #5565, opened 2020-06-08
Bug/5467 Shopfront search box returns all records when none is expected

### LINKED ISSUE #5467: Shopfront search box returns all records when none is expected
## Description

The search box in the shopfront (/shops) is showing all records when it should return zero results. This appears to be introduced by the latest release 2.9.9 #5433 .

## Expected Behavior

No results should be returned if there are no shops matching the text inserted in the search box.

## Actual Behaviour

All the shops are being returned.

## Steps to Reproduce

1. Go to /shops
2. Insert a string in the search box which does not match any criteria in the shop list
3. Observe all results being returned

## Animated Gif/Screenshot

The video from @luisramos0 shows the behaviour of the search box on a staging server with release 2.9.9 deployed, and compares it with a production server:

Peek 2020-05-19 11-27

## Workaround

The search box can still be used, but only when the input partially matches the existing shops.

## Severity

## Your Environment

* Version used: 2.9.9.

jeduardo824:
#### What? Why?

Closes #5467 

Hey guys, this is a suggested fix. The showed shops are the "distance matches" shops. We can erase them too, but since I don't know the expected behavior from the product view, I only fixed it to show the message "No results found."

After:
ezgif com-video-to-gif

#### Release notes

Fix the shopfront search box to show only matching records

Changelog Category:  Fixed

### DIFF
--- a/app/views/shops/_hubs.html.haml
+++ b/app/views/shops/_hubs.html.haml
@@ -9,9 +9,7 @@
 
   .row
     .small-12.columns
-      .name-matches{"ng-show" => "nameMatchesFiltered.length > 0"}
-        %h2
-          = t :hubs_matches
+      .name-matches{"ng-show" => "nameMatchesFiltered.length > 0 || query.length > 0"}
         = render "hubs_table", enterprises: "nameMatches"
 
       .distance-matches{"ng-if" => "nameMatchesFiltered.length == 0 || distanceMatchesShown"}


jeduardo824:
I will be happy to make the appropriate changes if you guys already know how the feature should be designed. 
Correct me if I'm mistaken, but I think this is a "product question." The current behavior is rendering a lot of inconsistent results because it searches with the name, location, etc... As far as I investigated, this is the reason that sometimes there are no results, and other times results are shown.

luisramos0:
I think this needs more investigation from someone who understands (or finds out) how the feature is supposed to work. 
Shall we close this PR (we still keep it for next time someone tries to fix it)?

sauloperez:
yes, this needs more product thinking and probably a broader discussion. Thanks a lot for taking the time @jeduardo824 though and apologies.

PR #5566, opened 2020-06-08
Add handling of CSV::MalformedCSVError for product_importer

### LINKED ISSUE #5132: No error message when product import's CSV is malformed
## Description

A Bugsnag error notification related to a malformed CSV just came in: https://app.bugsnag.com/open-food-network-belgium/open-food-network-belgium/errors/5e85ed88fdafed00193f6108?event_id=5e85ed8800599ff16e1e0000&i=sk&m=nw

As this error is not handled by the app, the user won't see any notice and won't have visual clue about the broken format.

## Expected Behavior

The user should know the file is not a valid CSV.

## Actual Behaviour

The import fails without any visual indication to the user

## Steps to Reproduce

Import products from a file that has at least one unquoted field containing the characters \r or \n.

## Workaround

none

## Severity

## Possible Fix

This method

https://github.com/openfoodfoundation/openfoodnetwork/blob/fe8beb5448b8b7d91af2c03c805a130e020f9cd3/app/models/product_import/product_importer.rb#L226-L240

should handle CSV::MalformedCSVError errors as well.

rioug:
#### What? Why?

Closes #5132 

The change adds handling of CSV::MalformedCSVError so it will display an error message when someone tries to import products with a malformed csv file. Namely, a file with a non quoted field containing the characters \r or \n.

#### What should we test?

Test the product import function, accessible via admin > products > import
You can use the file linked in this comment :https://github.com/openfoodfoundation/openfoodnetwork/issues/5132#issuecomment-640323032

A flash message with the following error "Importer Product Import encountered a malformed CSV: Unquoted fields do not allow \r or \n (line 3)." should appear upon uploading a problematic file.

Note, an unquoted \n will result in a file with some non valid line, which will give at least one error on the next import screen. It will not display a flash message.

#### Release notes

Add handling of malformed CVS file for the product import function. 

Changelog Category: Fixed

I have no idea what's involved with the upgrade, so I can't comment on any possible conflict. I'll just note that the product importer is used in this controller : 
 `ProductImportController < Spree::Admin::BaseController`

### DIFF
--- a/app/models/product_import/product_importer.rb
+++ b/app/models/product_import/product_importer.rb
@@ -238,14 +238,18 @@ module ProductImport
       end
       []
     rescue CSV::MalformedCSVError => e
-      # NOTE the init_product_importer method calls build_entries and
-      # buils_all_entries resulting in the error being raised twice
-      unless errors.added?(:importer, I18n.t('admin.product_import.model.malformed_csv',
-                                             error_message: e.message))
-        errors.add(:importer, I18n.t('admin.product_import.model.malformed_csv',
-                                     error_message: e.message))
-      end
+      add_malformed_csv_error e.message
       []
+    end
+
+    # This error is raised twice because init_product_importer calls both
+    # build_entries and buils_all_entries
+    def add_malformed_csv_error(error_message)
+      unless errors.added?(:importer, I18n.t('admin.product_import.model.malformed_csv',
+                                             error_message: error_message))
+        errors.add(:importer, I18n.t('admin.product_import.model.malformed_csv',
+                                     error_message: error_message))
+      end
     end
 
     def build_entries_in_range
--- a/app/models/product_import/product_importer.rb
+++ b/app/models/product_import/product_importer.rb
@@ -237,6 +237,15 @@ module ProductImport
                                      error_message: e.message))
       end
       []
+    rescue CSV::MalformedCSVError => e
+      # NOTE the init_product_importer method calls build_entries and
+      # buils_all_entries resulting in the error being raised twice
+      unless errors.added?(:importer, I18n.t('admin.product_import.model.malformed_csv',
+                                             error_message: e.message))
+        errors.add(:importer, I18n.t('admin.product_import.model.malformed_csv',
+                                     error_message: e.message))
+      end
+      []
     end
 
     def build_entries_in_range
--- a/spec/features/admin/product_import_spec.rb
+++ b/spec/features/admin/product_import_spec.rb
@@ -377,6 +377,24 @@ feature "Product Import", js: true do
       expect(page).to have_no_selector 'input[type=submit][value="Save"]'
       File.delete('/tmp/test.csv')
     end
+
+    it "handles cases where files contains malformed data" do
+      csv_data = "name,producer,category,on_hand,price,units,unit_type,shipping_category\n"
+      csv_data += "Malformed \rBrocolli,#{enterprise.name},Vegetables,8,2.50,200,g,#{shipping_category.name}\n"
+
+      File.write('/tmp/test.csv', csv_data)
+
+      visit main_app.admin_product_import_path
+      attach_file 'file', '/tmp/test.csv'
+      click_button 'Upload'
+
+      expect(page).to have_no_selector '.create-count'
+      expect(page).to have_no_selector '.update-count'
+      expect(page).to have_no_selector 'input[type=submit][value="Save"]'
+      expect(flash_message).to match(I18n.t('admin.product_import.model.malformed_csv', error_message: ""))
+
+      File.delete('/tmp/test.csv')
+    end
   end
 
   describe "handling enterprise permissions" do
--- a/spec/features/admin/product_import_spec.rb
+++ b/spec/features/admin/product_import_spec.rb
@@ -378,7 +378,7 @@ feature "Product Import", js: true do
       File.delete('/tmp/test.csv')
     end
 
-    it "handles cases where files contains malformed data" do
+    it "handles cases where files contain malformed data" do
       csv_data = "name,producer,category,on_hand,price,units,unit_type,shipping_category\n"
       csv_data += "Malformed \rBrocolli,#{enterprise.name},Vegetables,8,2.50,200,g,#{shipping_category.name}\n"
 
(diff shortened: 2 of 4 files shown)


spec/models/product_importer_spec.rb
```diff
@@ -169,11 +169,28 @@
     end
   end
 
+  describe "when uploading a spreadsheet with some malformed data" do
+    # Use a simple string as CVS.generate will do some escaping
```
luisramos0, line 173: there's a typo here CSV.generate

spec/models/product_importer_spec.rb
```diff
@@ -169,11 +169,28 @@
...
+    let(:csv_data) {
+      csv = "name,producer,category,on_hand,price,units,unit_type,shipping_category\n"
+      csv += "Good Carrots,#{enterprise.name},Vegetables,5,3.20,500,g,#{shipping_category.name}\n"
+      csv += "Malformed \rBrocolli,#{enterprise.name},Vegetables,8,2.50,200,g,#{shipping_category.name}\n"
+    }
+    let(:importer) { import_data csv_data }
+
+    # NOTE an unquoted \n will create a non valid line which will fail entry validation hence why we are only testing with \r
+    it "should raise an unquoted field error if data include unquoted filed with \r character" do
```
luisramos0, line 182: typo in filed

app/models/product_import/product_importer.rb
```diff
@@ -237,6 +237,15 @@ def rows
                                      error_message: e.message))
       end
       []
+    rescue CSV::MalformedCSVError => e
+      # NOTE the init_product_importer method calls build_entries and
```
luisramos0, line 241: can you please remove NOTE here as well?
just
`# the init_product_importer method calls build_entries and`

luisramos0: A rule for code comments I love is: if you need to comment a line of code, extract method and make it a method comment. In this case:
```ruby
  rescue CSV::MalformedCSVError => e
     add_malformed_csv_error(e.message)
     []
  end

  # This error is raised twice because init_product_importer calls both build_entries and buils_all_entries
  def add_malformed_csv_error(error_message)
      return if errors.added?(:importer, I18n.t('admin.product_import.model.malformed_csv', error_message: error_message))

      errors.add(:importer, I18n.t('admin.product_import.model.malformed_csv',error_message: error_message))
  end
```       

I think this looks nicer, what do you think?

spec/features/admin/product_import_spec.rb
```diff
@@ -377,6 +377,24 @@
       expect(page).to have_no_selector 'input[type=submit][value="Save"]'
       File.delete('/tmp/test.csv')
     end
+
+    it "handles cases where files contains malformed data" do
```
luisramos0, line 381: typo: files contain

spec/models/product_importer_spec.rb
luisramos0: NOTE is not needed, you can just right your comment after #

luisramos0: I dont understand "we are only testing with \r"
The test csv contains \n
Why do you say you are only testing with \r?

app/models/product_import/product_importer.rb
luisramos0: this change will also help with the codeclimate warning:
Assignment Branch Condition size for rows is too high. [<2, 21, 6> 21.93/15]

spec/models/product_importer_spec.rb
rioug: The test data has \n at the end of the lines to create a "valid" csv file. I mentioned we only test with \r as adding a \n in a random place will not trigger the error we are testing, only a \r will. Does that make sense ? Maybe I need to reword my comment.

app/models/product_import/product_importer.rb
rioug: > A rule for code comments I love is: if you need to comment a line of code, extract method and make it a method comment. 

Make sense, although I feel like It's just adding a method to comply with the codeclimate rules. Anyway it does make the code a bit more readable.

Rubocop still throw a warning with the suggested changes : 
`app/models/product_import/product_importer.rb:226:5: C: Metrics/AbcSize: Assignment Branch Condition size for rows is too high. [<2, 15, 5> 15.94/15]`

I am not sure if there is a way to make it better, what do you think ?

luisramos0: >>  although I feel like It's just adding a method to comply with the codeclimate rules.

imo code climate rules are very important and readability is fundamental. It's adding a method to make the code a lot more readable and less complex. The complexity rule is very important. Messy code will create bugs and head aches.

The warning is there but now it's 15.94 , not 21. It means it's less complex.
I think we can leave it like that :+1:

rioug:
I fixed the various typos and made the suggested changes. It should be good now.
Thanks for the feedback !

spec/models/product_importer_spec.rb
```diff
@@ -169,11 +169,28 @@
...
+  describe "when uploading a spreadsheet with some malformed data" do
+    # Use a simple string as CSV.generate will do some escaping
+    let(:csv_data) {
+      csv = "name,producer,category,on_hand,price,units,unit_type,shipping_category\n"
+      csv += "Good Carrots,#{enterprise.name},Vegetables,5,3.20,500,g,#{shipping_category.name}\n"
+      csv += "Malformed \rBrocolli,#{enterprise.name},Vegetables,8,2.50,200,g,#{shipping_category.name}\n"
+    }
+    let(:importer) { import_data csv_data }
+
+    # an unquoted \n will create a non valid line which will fail entry validation hence why we are only testing with \r
```
luisramos0, line 181: oh, I got it now!
If you add a \n in the middle of a line it will create a non valid line :+1: 
I was confused because the \n is there! maybe "an unquoted \n in the middle of a line" could help the next reader :-)

filipefurtad0:
Hey @rioug ,

Reproduced the bug - Error 500 and bugsnag got triggered, when uploading the file you attached (Thanks!).

After this PR it's the warning is visible (instead of error 500/bugsnag).
Awesome, thank you for this PR!

image

PS: I noticed there are issues on Code climate, which shows the "Assignment Branch Condition size for rows is too high" warning. Is this a minor detail? Ping @luisramos0 @Matt-Yorkley

luisramos0:
yeah, that codeclimate issue was discussed in the code review and improved, although not fixed.

PR #5569, opened 2020-06-09
Add translation keys to en.yml file for variant_override error messag…

### LINKED ISSUE #5204: Missing translation key in inventory error message
## Description

The error message states "count on hand" in English

## Steps to Reproduce
1. Go to `/admin/inventory`
2. Leave count on hand and on demand check box blank
3. Try to save
4. See error message

## Animated Gif/Screenshot

image

## Severity

## Your Environment
* Version used: v2.9.1

rmklaus12:
#### What? Why?
Closes #5204 
On admin/inventory, the "Count on Hand" error is not being translated like the rest of the message is. 
Screen Shot 2020-06-09 at 4 09 42 pm
Needed to add translation keys to locales/en.yml file. This error was tracked to an Activerecord attribute and adding the key to en.activerecord.attributes.variant_override.count_on_hand, allowed the error to be translated. 

#### What should we test?

1. Go to /admin/inventory
2. Leave 'On Hand' field blank
3. Set 'On Demand?' to 'No'
4. Click Save
5. With translations added, the message should be fully translated. 
6. I tested locally by changing the en.activerecord.attributes.variant_override.count_on_hand string to "COH"
Screen Shot 2020-06-09 at 4 18 15 pm

#### Release notes
Added translations keys to allow full error message to be properly translated.

Changelog Category: Added

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -44,6 +44,8 @@ en:
         base: "Credit Card"
       order_cycle:
         orders_close_at: Close date
+      variant_override:
+        count_on_hand: "On Hand"
     errors:
       models:
         spree/user:


config/locales/en.yml
```diff
@@ -44,6 +44,8 @@ en:
         base: "Credit Card"
       order_cycle:
         orders_close_at: Close date
+      variant_override:
+        count_on_hand: "Count on Hand"
```
luisramos0, line 48: I think it's better if the translation is "On hand" so that it is the same as the variant field.
So this would become:
`count_on_hand: "On Hand"`
Translators can still change it in transifex.

rmklaus12: change made

filipefurtad0:
Hey @rmklaus12 :-)

Looks fine:
image
Merging.

PR #5576, opened 2020-06-09
Reports rewrite POC

Matt-Yorkley:
I had a little play around over the weekend with what a reports rewrite might look like and I came up with this... 

I implemented a new base Report class using the template method pattern and extracted the main bits of functionality into 3 services that basically deal with: data wrangling, aggregating/summarising, and rendering. 

I re-implemented the "Packing by Customer" report with the new Report class. I had to hack some of the views a bit, so this will definitely break other reports in it's current state (it's a quick POC)... 

With the new packing report:

- The report's output is the same as before
- The definition of the packing report is 10 times easier to read, and a lot more concise
- I did some quick performance comparison using production data, and the packing report renders **twice as fast** as before
- The report can be rendered in various file formats or fetched via the API

In general, with reports based on this new class:

- Defining a new report or changing an existing one is so easy we could put the `good-first-issue` tag on it
- Any report can be created with 3 simple inputs (current user, search params, customisation options) and rendered on the screen, as plain CSV format, as Excel spreadsheets, as OpenOffice spreadsheets, or as JSON in an API endpoint, with the same data
- There's various places we could add additional aggregating/grouping/reordering functions very easily, and they would be available in all reports
- The report's data is easy to adjust, sort and generally manipulate before it's rendered, so it'll be easy to add various UX options that would have been difficult before, for example the user could: show/hide columns, reorder columns, show/hide summary rows, etc. before exporting to a spreadsheet
- It should be incredibly easy to switch existing reports to this new style

### Examples of a report definition:

Defining the main collection of objects looks like this:
```ruby
def collection
   Spree::LineItem.includes(:product, :order).where(order_id: visible_orders).
      ransack(ransack_params).result
end
```

Defining the report's headings and what the rows should contain looks like this:
```ruby
def report_rows(objects)
   {
      order_number: object.order.number,
      product_name: object.variant.product.name,
      quantity: object.quantity,
      amount: object.amount
   }
end
```

Defining ordering of the results by several different columns looks like this:
```ruby
def order_by
   [:supplier, :order, :last_name]
end
```

Defining a grouping and including a summary row after each group that includes the sum of values for that group for :quantity and :price columns would look like this:
```ruby
def summary_group
   [:supplier]
end

def summary_row
   { title: 'TOTALS:', sum: [:quantity, :price] }
end
```

Hiding a column that was used in grouping or ordering but shouldn't be output in the report would look like this:
```ruby
def hide_columns
   [:customer_id]
end
```

The report definition doesn't need much more than that, and the new Report class handles the rest...

#### Rendering options:

Screenshot from 2020-06-09 19-22-10

#### Downloading in Excel format:

Screenshot from 2020-06-09 19-26-42

#### Output from the corresponding API endpoint:

Screenshot from 2020-06-09 09-30-32

### DIFF
--- a/app/services/report_summariser.rb
+++ b/app/services/report_summariser.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+class ReportSummariser
+  def initialize(group_column, report_rows, report)
+    @group_column = group_column
+    @report_rows = report_rows
+    @report = report
+  end
+
+  def call
+    insert_summary_rows unless exclude_summaries?
+
+    @report_rows
+  end
+
+  private
+
+  def insert_summary_rows
+    grouped_rows = []
+    previous_grouping = nil
+
+    @report_rows.each_with_index do |row, row_index|
+      current_grouping = row[@group_column]
+
+      if previous_grouping.present? && current_grouping != previous_grouping
+        grouped_rows << build_summary_row(@group_column, previous_grouping, @report.summary_row)
+      end
+
+      grouped_rows << row
+
+      if last_row?(row_index)
+        grouped_rows << build_summary_row(@group_column, previous_grouping, @report.summary_row)
+      end
+
+      previous_grouping = current_grouping
+    end
+
+    @report_rows = grouped_rows
+  end
+
+  def build_summary_row(group_column, group_key, options)
+    summary_row = initialize_empty_row
+    group_rows = @report_rows.select{ |row| row[group_column] == group_key }
+
+    summary_row[:summary_row_title] = options[:title]
+
+    (options[:sum] || []).each do |sum_column|
+      summary_row[sum_column] = group_rows.sum{ |group_row| group_row[sum_column] }
+    end
+
+    (options[:show_first] || []).each do |first_column|
+      summary_row[first_column] = group_rows.first[first_column]
+    end
+
+    summary_row
+  end
+
+  def initialize_empty_row
+    row = {}
+    @report.headers.each do |key|
+      row[key.to_sym] = ""
+    end
+
+    row
+  end
+
+  def last_row?(row_index)
+    @report_rows.length == row_index + 1
+  end
+
+  def exclude_summaries?
+    @report.options[:exclude_summaries]
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -87,6 +87,7 @@ gem 'wkhtmltopdf-binary'
 gem 'foreigner'
 gem 'immigrant'
 gem 'roo', '~> 2.8.3'
+gem 'spreadsheet_architect'
 
 gem 'whenever', require: false
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -143,6 +143,9 @@ GEM
       json (~> 1.4)
       nokogiri (>= 1.4.4)
       uuidtools (~> 2.1)
+    axlsx_styler (1.0.0)
+      activesupport (>= 3.1)
+      caxlsx (>= 2.0.2)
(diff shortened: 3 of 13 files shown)


Matt-Yorkley:
I think we could also easily dump the report data into visualisation tools like https://chartkick.com/ or https://www.reportskit.co/ and dispaly the as beautiful charts in the admin dashboard as well...

Matt-Yorkley:
Yes, I'm not suggesting we merge this as is. Firstly it was an experiment to see what nice reports code might look like and check the performance. Secondly, its a proposal for discussion. Do we want to go in this direction?go in this direction.

luisramos0:
I see :+1:

Matt-Yorkley:
So the proposal (if we decide to go in this direction) would be firstly to tidy up the rough edges of this PR, namely the bits I skipped over in order to make a working POC as rapidly as possible:
- fix up the views and the way the form helpers interact with the report object
- add test coverage for the new classes and services
- move the new code to the order_management engine

and then switch the other reports to using this new class and services:
- define the rest of the existing reports in this new style
- :fire::fire: all of the previous reports code

RachL:
@Matt-Yorkley if you want to move forward I think you need to propose it during product curation meeting? That would help to get it the visibility and priority? I'm not sure why this has not been brought up in this morning meeting, too bad.
Maybe something to discuss at the report meeting next Tuesday.

lin-d-hop:
I'm also confused about this proof of concept.

I understand this is an exciting thing to explore but we're definitely not at the point of building this yet. This is purely tech driven without product input. And now we've had 4 core team members looking at it so even if it is a volunteer project the impacts ripple.

Can we park the tech explorations of reports for a while?

kirstenalarsen:
it's a pretty exciting POC in my view - thanks @Matt-Yorkley :) I think this brings a lot of potential useful info to the next reports meeting in terms of feasibility and potential pathways which seemed to be the big question out of the last one. You make it sound all quite doable 🤞

sauloperez:
Well, I think the goal of this PoC of bringing down to earth what's possible and what's not was already accomplished. It helps a great deal at imagining what can the new reports look like in terms of implementation but that's it.

I'm with @lin-d-hop that we still need to define in terms of product what it is that needs to be built and not get too caught up with a particular implementation.

app/services/report_summariser.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      if last_row?(row_index)
+        grouped_rows << build_summary_row(@group_column, previous_grouping, @report.summary_row)
+      end
+
+      previous_grouping = current_grouping
+    end
+
+    @report_rows = grouped_rows
+  end
```
sauloperez, line 39: IMO the moment we make some processing/aggregation in Ruby is when things get slow to a crawl and get worse as more requirements pile up. The way I see it, we should move all the burden to the DB.

There's much more we can benefit from PostgreSQL. It has many advanced analytical capabilities and specifically [Window Functions](https://www.postgresql.org/docs/9.6/tutorial-window.html) I think should provide us with all we need.

Matt-Yorkley: Would all the reports queries have to be done in SQL if we did this though? Or how would that work nicely with things like Ransack and ActiveRecord scopes?

Matt-Yorkley: Did I mention it's already twice as fast as our current implementation? :trollface:

sauloperez: sure, but I would also like the leverage the feaures PostgreSQL so we maintain less code.

Matt-Yorkley:
Superseded by #5753

PR #5593, opened 2020-06-12
Fix link from report to order

sauloperez:
#### What? Why?

Fixes the link which now points to a page that doesn't exist and thus causes a slug.

I haven't found any other report rendering the order number.

#### What should we test?

Render any of the tax reports and click on any order number. It should lead to the order edit page.

#### Release notes

Fix broken order link from tax reports
Changelog Category: Fixed

### DIFF
--- a/app/views/spree/admin/reports/_link_order.html.haml
+++ b/app/views/spree/admin/reports/_link_order.html.haml
@@ -1,1 +1,2 @@
-%a.edit-order{href: "/admin/orders/#{value}"}= value
+- order_number = value
+= link_to order_number, edit_admin_order_url(order_number), class: 'edit-order'


filipefurtad0:
Hey @sauloperez ,

Awesome - this works now. Unaware of your PR I opened an issue for this yesterday #5626 which this PR closes now.

Ready to go!

RachL:
@sauloperez would have been great to open the link in a new tab I think :) was this discussed somewhere?

sauloperez:
>  was this discussed somewhere?

Not it was just me fixing something that annoyed me while developing something else. I can quickly make it open on a separate tab if that's what we want.

PR #5594, opened 2020-06-12
[Spree 2.1] Fix "subquery has too many columns" in reports query

Matt-Yorkley:
Fixes "subquery has too many columns" in reports query.


Matt-Yorkley:
There's a coupe of flaky specs still, but the reports errors are all fixed :heavy_check_mark:

Issue #5572, opened 2020-06-09
500 Error on Payment Methods Report


lin-d-hop:
## Description

The Order Cycle Management -> Payments Methods report loads all orders before even loading the filter/search page.

To find Payments Methods it uses a line:
`pm = o.payments.first.payment_method` -> payment.first is nil

If ONE order that is available to the user from ANY of their enterprises fails to find a payment method then the whole page crashes - 500Error:
`undefined method `payment_method' for nil:NilClass`

Bugsnag: https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5ea17c7622a1830018e1f834?filters[event.since][0]=30d&filters[error.status][0]=open

Since the system allows an order without a payment (?) we need this report to run regardless.

## Expected Behavior

The report, ya know, works.

## Actual Behaviour

SLUG

## Steps to Reproduce

I have not been able to track down the enterprise that this order is associated with yet - but ther user noreply@openfoodnetwork.org.uk is associated with it. This user is the one I use for all reports integrations.

1. Logged in as a user associated with an order without a payment 
2. Go to Payment Methods report
3. Slug! (500 error)

## Severity

bug-s2: a non-critical feature is broken, no workaround

## Your Environment

* Version used: 2.10.0
* Browser name and version:
* Operating System and version (desktop or mobile):

## Possible Fix

From @luisramos0 on Slack:
Should be very easy to address (like this o.payments.first&.payment_method)

kirstenalarsen:
is this the same thing? https://github.com/openfoodfoundation/openfoodnetwork/issues/4977

RachL:
@kirstenalarsen Well what I understand is that in @lin-d-hop 's case it's a super user that has the problem and not one enterprise in particular, which makes it difficult to find the order causing this.

That's the difference I see between the two.

lin-d-hop:
Definitely not related to #4977.

Payment Totals report is not the Order Cycle Management -> Payment Methods report.

RachL:
@lin-d-hop it's not the same report, but I think it is the same cause: some corrupt data.

lin-d-hop:
Could well be related and due to an order not having a payment associated:
From Payments Report code ->
'payments.first.payment_method.name'

The key difference is that the data is loaded before the page renders so the corrupted order makes the page totally unusable for any user associated with the order.

Issue #4977, opened 2020-03-12
Snail when running Payment Totals Report 


emilyjeanrogers:
## Description

Aus customer Prom Coast Wholefoods Collective are getting a snail when running the Payment Totals Report. Snail is generated when including a date more than 12 months before today (11th March 2020). The snail comes up when trying to run this report as both a Manager of the Customer and as Superadmin. Have also tested for other Enterprises with no problems.

## Expected Behavior
Report should run successfully for any date range.

## Actual Behaviour

It shows a snail. The error has been logged:
https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5e68221b18c0ef00177d7370

## Steps to Reproduce

1. Log in as Superadmin (or add yourself as a manager for Prom Coast Wholefoods Collective)
2. Go to Reports
3. Select Payment Totals Report
4. Set Date range for a date including more than 12 months prior to today's date  
OR  
set date range to include dates prior to 11th March 2019

## Workaround
No workaround.

## Severity
S2

bug-s2: a non-critical feature is broken, no workaround

## Possible Fix
It looks like a payment method is missing. The safe navigation operator `&.` may help here:

```
NoMethodError · undefined method `name' for nil:NilClass 
lib/open_food_network/payments_report.rb:106
```
```rb
proc { |orders| orders.sum { |o| o.payments.select { |payment| payment.completed? && (payment.payment_method.name.to_s.include? "EFT") }.sum(&:amount) } },
```

RachL:
@emilyjeanrogers I already had this because there was a particular problem with an order. Have you tried narrowing the search and see for which set of dates the error occurs?

kirstenalarsen:
@emilyjeanrogers is this an s2? or is an s3 - not sure why it got downgraded? Seems likely connected to https://github.com/openfoodfoundation/openfoodnetwork/issues/5572

lin-d-hop:
Very unlikely to be connected to #5572

#5572 is specific to payment methods report and is related to an order not having a payment associated. It is a data corruption issue and renders this report totally unusable ALWAYS for any user that is associated with that order.

I fail to see how they can be related.

emilyjeanrogers:
I have reproduced this error today and been able to narrow it to a more specific date range. The snail occurs for any Payment Totals Report run for Prom Coast Food Collective prior to March 12th 2019.

sigmundpetersen:
[...]

PR #5595, opened 2020-06-12
Remove N+1 fetching payments in report

sauloperez:
#### What? Why?

Removes an N+1 related to payments

#### What should we test?

The payments report should work as expected while being a little faster.

#### Release notes

Remove N+1 from payments report
Changelog Category: Changed

### DIFF
--- a/lib/open_food_network/payments_report.rb
+++ b/lib/open_food_network/payments_report.rb
@@ -41,7 +41,7 @@ module OpenFoodNetwork
       return [] unless @render_table
 
       orders = search.result
-      payments = orders.map { |o| o.payments.select(&:completed?) }.flatten # Only select completed payments
+      payments = orders.includes(:payments).map { |o| o.payments.select(&:completed?) }.flatten # Only select completed payments
       case params[:report_type]
       when "payments_by_payment_type"
         payments
--- a/lib/open_food_network/payments_report.rb
+++ b/lib/open_food_network/payments_report.rb
@@ -41,7 +41,10 @@ module OpenFoodNetwork
       return [] unless @render_table
 
       orders = search.result
-      payments = orders.includes(:payments).map { |o| o.payments.select(&:completed?) }.flatten # Only select completed payments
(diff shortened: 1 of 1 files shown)


filipefurtad0:
Hey @sauloperez,
This relates to #5492, right? I'd test one after the other then.

sauloperez:
I can merge them together if that's going to make it easier for you by they touch two different reports.

filipefurtad0:
Thanks @sauloperez, not necessary. Thought this related to eventual performance issues introduced by the PR I mentioned. No need to merge them :+1:

filipefurtad0:
Hey @sauloperez ,

Tested this in Staging-UK, as superadmin.
In Payments by Type there is a decrease of around 3 sec. load time :tada: 

However - and not introduced by this PR - the Payment Totals appears to be broken:

image

https://openfoodnetwork.slack.com/archives/CEF14NU3V/p1592334661341000

Perhaps related with #5572?

sigmundpetersen:
@filipefurtad0 there's an issue with snail on Payment Totals #4977.
It has been switching between s3 and s2. What do you think? S2?

lib/open_food_network/payments_report.rb
```diff
@@ -41,7 +41,10 @@ def table_items
       return [] unless @render_table
 
       orders = search.result
-      payments = orders.map { |o| o.payments.select(&:completed?) }.flatten # Only select completed payments
+      payments = orders.includes(:payments).map do |order|
+        order.payments.select(&:completed?)
+      end.flatten
+
```
Matt-Yorkley, line 47: I think something like this might be even more efficient:
```ruby
payments = Spree::Payment.completed.where(order_id: orders.select('spree_orders.id'))
```

sauloperez: Yes, I also wanted to get rid of this `map` but it was so quick and non-distracting to do this that I thought to improve in a later PR.

PR #5599, opened 2020-06-13
Make Enter submit the form on both products and orders admin

### LINKED ISSUE #4437: 'Enter' key activates search on Products (and Orders?) page

When repeatedly using search box to find and deal with products in Products page, it is really irritating to have to actually hit 'filter results' before anything happens. It would be great if hitting enter, which I (and I suspect many others) do instinctively, activated the search.

Pretty standard to type something in a search box and have 'enter' then do the search?

I suspect there are issues with consistency of UX across the admin back-end, but I wonder if this is one small thing that might be good?

when I type something in the search box on the Products page, I can hit 'enter' and the search runs
when I type something in a search box on the Orders page, I can hit 'enter' and the search runs

luisramos0:
#### What? Why?

Closes #4437

It feels like too much copy paste for such a simple feature. I left the first commit so reviewers can see the alternative directive approach. The alternative approach was not working for the orders page, so I decided to go with ng-keypress.

On the orders page we only add the behavior to free text fields.

#### What should we test?
Products search /admin/products - type something on quick search and press enter, see the results being updated.
Orders search /admin/orders - type something on any of the text fields (Invoice number, email, first name and last name) and press enter, see results being updated

#### Release notes
Changelog Category: Added
Pressing enter will submit the search form on admin orders and products.

### DIFF
--- a/app/assets/javascripts/admin/bulk_product_update.js.coffee
+++ b/app/assets/javascripts/admin/bulk_product_update.js.coffee
@@ -64,10 +64,6 @@ angular.module("ofn.admin").controller "AdminProductEditCtrl", ($scope, $timeout
     RequestMonitor.load(BulkProducts.fetch(params).$promise).then ->
       $scope.resetProducts()
 
-  # Used by directive enterToSubmit
-  $scope.submit = ->
-    $scope.fetchProducts()
-
   removeClearedValues = ->
     delete $scope.producerFilter if $scope.producerFilter == "0"
     delete $scope.categoryFilter if $scope.categoryFilter == "0"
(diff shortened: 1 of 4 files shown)


luisramos0:
yes, agree but i couldnt make it work for orders page. Do you want to have a look? Could be a very simple thing.

sauloperez:
I'm of the opinion that better done than nothing, for this. Then, we can spend some time improving it but knowing this is already done.

sauloperez:
so @Matt-Yorkley . Which way do you want to go?

filipefurtad0:
Hi @luisramos0 ,

Thank you for your notes on the issue and in this PR.

I verified that the PR makes hitting enter work, as an alternative to manually pressing the Filter Results button on:

- Products page
-- Quick search field
- Orders Page
-- Fields Invoice number, Email, First name begins with, Last name begins with

Good to go,
Thank you!

PR #5601, opened 2020-06-15
Enhancement/5454 make links to shops under Groups > Producers open in new tab

### LINKED ISSUE #5454: Can the links to shops on the group page open into new tabs?
## Description
At present on a group page - see for example: https://openfoodnetwork.ca/groups/www-openfoodnetwork-ca-justfoodfarmstand#/producers,   each of the links to different farm shops opens in the same tab.   Is there a way to change the links so each opens in a new tab?

In OFN_CAN we have a number of farmers' markets set up as groups.  But,  because we don't have split payments,  each vendor is taking their own payment.   These users are wondering if we could help things along by having the links open in new tabs - that way when a shopper checks out from one market vendor,  they will still have the market group tab open  to go back to find the next vendor to shop from.

I know this isn't a bug - its an enhancement,  but I'm wondering if its a paper cut?  Anyone? 

## Expected Behavior
Operating as expected - just not great user experience.

## Steps to Reproduce

1.
2.
3.
4.

## Your Environment

* Version used:
* Browser name and version:

jeduardo824:
#### What? Why?

Closes #5454 

#### What should we test?
If the shops' links are opening in a new tab properly (both producers name and shops when the tab expands).

#### Release notes
Make shops' links on groups page open in a new tab (both producer name and shops when the tab expands).
Changelog Category: Changed

Result:
ofn
ofn

### DIFF
--- a/app/views/producers/_fat.html.haml
+++ b/app/views/producers/_fat.html.haml
@@ -79,8 +79,8 @@
     .row.cta-container
       .columns.small-12
         %a.cta-hub{"ng-repeat" => "hub in producer.hubs | orderBy:'-active'",
-        "ng-href" => "{{::hub.path}}", "ng-attr-target" => "{{ embedded_layout ? '_blank' : undefined }}",
-        "ng-class" => "::{primary: hub.active, secondary: !hub.active}"}
+        "ng-href" => "{{::hub.path}}", "ng-attr-target" => "_blank",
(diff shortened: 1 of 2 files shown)


jeduardo824:
@tschumilas Requested changes! Please, do not merge!

jeduardo824:
Ready to go!!

filipefurtad0:
Hey @jeduardo824 ,

Nice! These links now open in separate tabs - Thank you for this improvement.
Good to go.

PR #5605, opened 2020-06-15
Update spree revision and upgrade ransack to 1.2.3

luisramos0:
#### What? Why?

This will pick up the last change in spree to relax rails dependency and allow the upgrade to rails 4.1 with spree 2.1.
We upgrade ransack to 1.2.3 which is a good thing in itself but also needed for rails 4.1.

#### What should we test?
Green build.
The build is already testing this but we should manually verify search pages and make sure the filters sent in the query params are working properly. This can be tested by search on the orders and also products in the admin and make sure the results match the used filters.

#### Release notes
Changelog Category: Changed
Updated spree revision so we can upgrade the rails version in OFN.
Upgraded ransack, the library we use to search data in the database.


filipefurtad0:
Hi @luisramos0 ,

Tested:

Products
- producer
- category
- import date
- quick search

and Orders:
- date range
- invoice nr.
- first and surname
- (un)ticking "Show only complete orders"
- shipping methods
- order cycles
- distributors
- status
- email

Checked all of them independently and some of the combinations. The drop-down searches work fine and the results correspond to what's expected. I'd say this is good to go.

luisramos0:
Thanks Filipe!!!

Issue #5386, opened 2020-05-07
[Rails 4.1] Upgrade OFN to Rails 4.1


luisramos0:
## What is the problem we are solving 
This issue represents the effort to upgrade rails to 4.1.

This is related to #4826 but not necessarily blocked by it as we can now upgrade rails without upgrading spree.

This discourse thread contains more info about this:
https://community.openfoodnetwork.org/t/bye-bye-spree-or-spree-v2-1-and-beyond/1598/30

I read the rails 4.1 upgrade guide and I estimated this as 3 weeks of dev time. I wonder if the dual boot is worth the effort if these upgrades without spree look so easy for an app like OFN.
Would be good to get a second dev to do this  estimate :-)

NickMoignard:
Should we stop upgrading at 4.1? Rails 6 + updated gems have fewer security vulnerabilities.

luisramos0:
Hello Nick, no! We want to go all the way, this is only the next step. We have been making this top priority for some time now (upgraded to spree 2.0 and we are about to finish upgrade to spree 2.1 which uses rails 4.1, we will then remove spree_core and upgrade rails on our side, this issue).

What do you think? Would be great to get opinions from new contributors!

NickMoignard:
Awesome, I agree with your quote it is quite a massive task, perhaps 4 weeks.

'BTW, the dual boot #4800 work is a good issue to be picked up by experienced new contributors!' awesome i'll have a look now.

luisramos0:
I think I found a way where we can do this upgrade without having to finish #4826 first here: #5456 

That means we can start thinking about this upgrade. The official upgrade guidelines are here:
https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-0-to-rails-4-1

From this list, some of the points we may need to address are:
- https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#mutator-methods-called-on-relation
- https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#changes-on-default-scopes

Upgrades (taken from spree 2-3-stable that runs on rails 4.1) that may be needed:
- [ ]   s.add_dependency 'acts_as_list', '= 0.3.0'
- [ ]   s.add_dependency 'awesome_nested_set', '~> 3.0.0.rc.3'
- [ ]   s.add_dependency 'aws-sdk-v1', '1.64.0'
[...]

PR #5610, opened 2020-06-15
[Rails 4.1] Start Rails 4.1 upgrade

luisramos0:
#### What? Why?
Starts #5386

The upstream/branch rails-4-1 is a copy of master.

This PR upgrades rails to 4.1 and adds a few initial fixes to make the build better.

#### What should we test?
No major errors in the build.


luisramos0:
ah, I forgot to mention, the build will look much better after #5605
I think that error will be gone.
After #5605 is merged to master, I can merge master into rails-4-1 and then rebase #5611 - that's where the better 4.1 build will be.

luisramos0:
ok, I got #5605 into rails 4-1- branch and merged this PR as well.
I rebased #5611 so the latest build is this one:
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/rails-4-1/builds/23

144 broken specs. Not bad. I think there are still a few quick fixes to be done.

PR #5611, opened 2020-06-15
[Rails 4.1] Mutator methods called on Relation

luisramos0:
#### What? Why?

This has been separated in several PRs:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5919
and
https://github.com/openfoodfoundation/openfoodnetwork/pull/5921

This are the commits in this PR that may still be needed after the PRs above reach rails-4-1 branch from master.
rails 4-1 respond_with error
8fe1175518fe6429152296660cdff901e7ea218f

assertion problem - this should probably be squashed together with the commit already in rails-4-1 branch:
9b323e34146b0bcfd8d1f29af5f44539b61856ba 

this is a factory fix - I dont think this is needed, maybe just a local test db problem:
3a078983df9791a95c051a4998955af5d5bffdc1

### DIFF
--- a/app/controllers/cart_controller.rb
+++ b/app/controllers/cart_controller.rb
@@ -1,5 +1,3 @@
-require 'spree/core/controller_helpers/order'
-
 class CartController < BaseController
   before_action :check_authorization
 
--- a/app/controllers/spree/orders_controller.rb
+++ b/app/controllers/spree/orders_controller.rb
@@ -1,6 +1,3 @@
-require 'spree/core/controller_helpers/order'
-require 'spree/core/controller_helpers/auth'
-
 module Spree
   class OrdersController < Spree::StoreController
     include OrderCyclesHelper
--- a/app/helpers/checkout_helper.rb
+++ b/app/helpers/checkout_helper.rb
@@ -26,7 +26,7 @@ module CheckoutHelper
 
   def display_checkout_admin_and_handling_adjustments_total_for(order)
     adjustments = order.adjustments.eligible.where('originator_type = ? AND source_type != ? ', 'EnterpriseFee', 'Spree::LineItem')
-    Spree::Money.new adjustments.sum(&:amount), currency: order.currency
+    Spree::Money.new adjustments.sum(:amount), currency: order.currency
   end
 
   def checkout_line_item_adjustments(order)
@@ -34,7 +34,7 @@ module CheckoutHelper
   end
 
   def checkout_subtotal(order)
-    order.item_total + checkout_line_item_adjustments(order).sum(&:amount)
+    order.item_total + checkout_line_item_adjustments(order).sum(:amount)
   end
 
   def display_checkout_subtotal(order)
--- a/app/helpers/checkout_helper.rb
+++ b/app/helpers/checkout_helper.rb
@@ -18,7 +18,7 @@ module CheckoutHelper
     enterprise_fee_adjustments = adjustments.select { |a| a.originator_type == 'EnterpriseFee' && a.source_type != 'Spree::LineItem' }
     adjustments.reject! { |a| a.originator_type == 'EnterpriseFee' && a.source_type != 'Spree::LineItem' }
     unless exclude.include? :admin_and_handling
-      adjustments << Spree::Adjustment.new(label: I18n.t(:orders_form_admin), amount: enterprise_fee_adjustments.sum(&:amount))
+      adjustments << Spree::Adjustment.new(label: I18n.t(:orders_form_admin), amount: enterprise_fee_adjustments.sum(:amount))
     end
 
     adjustments
--- a/app/models/spree/line_item.rb
+++ b/app/models/spree/line_item.rb
@@ -178,7 +178,7 @@ module Spree
     end
 
     def included_tax
-      adjustments.included_tax.sum(&:included_tax)
+      adjustments.included_tax.sum(:included_tax)
     end
 
     def tax_rates
--- a/app/models/spree/order.rb
+++ b/app/models/spree/order.rb
@@ -739,15 +739,15 @@ module Spree
     end
 
     def shipping_tax
-      adjustments(:reload).shipping.sum(&:included_tax)
+      adjustments(:reload).shipping.sum(:included_tax)
     end
 
     def enterprise_fee_tax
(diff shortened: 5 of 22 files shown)


luisramos0:
Probably something for this PR as well:
```

  2) shipping methods as a site admin creating a shipping method owned by some distributors
     Failure/Error:
       @hubs = Enterprise.managed_by(spree_current_user).is_distributor.sort_by! do |d|
         [(@shipping_method.has_distributor? d) ? 0 : 1, d.name]
       end
     
     NoMethodError:
       undefined method `sort_by!' for #<Enterprise::ActiveRecord_Relation:0x00007fc01467cbf0>
     # ./app/controllers/spree/admin/shipping_methods_controller.rb:53:in `load_hubs'
     # ./lib/open_food_network/rack_request_blocker.rb:36:in `call'
     # ./spec/features/admin/shipping_methods_spec.rb:22:in `block (3 levels) in <top (required)>'

```

lib/open_food_network/tag_rule_applicator.rb
```diff
@@ -14,6 +14,7 @@ def initialize(enterprise, rule_type, customer_tags = [])
     def filter!(subject)
       return unless subject.respond_to?(:any?) && subject.any?
 
+      subject = subject.to_a if subject.class.to_s.end_with?("ActiveRecord_Relation") || subject.class.to_s.end_with?("ActiveRecord_AssociationRelation")
```
Matt-Yorkley, line 17: Looking through the code at where this is actually called from, in most cases we have a relation and `#to_a` is called on it before passing it as the argument in `#filter!`. So `#filter!` will either be given an array or a relation? But by the time the method is finished the `subject` will be an array anyway, right...?

Can we just call `#to_a` on `subject` without the conditional check?

Matt-Yorkley: We can just cut line 17 and change line 18 to this
```
subject.to_a.reject! do |element|
```
and the result will be the same, right?

Matt-Yorkley:
> undefined method `sort_by!' for #<Enterprise::ActiveRecord_Relation:0x00007fc01467cbf0>

That just needs a `#to_a` before the `#sort_by!` as well, right? It's an `Array` method.

lib/open_food_network/tag_rule_applicator.rb
Matt-Yorkley: I'm pretty sure I saw this in the Rails 4.0 upgrade as well, methods that checked the class of the passed object started failing, as they were not always the expected `ActiveRecordRelation`, but sometimes `ActiveRecordAssociationRelation` instead, and some of the behavior of relations shifted a bit and broke a few things...

luisramos0: I am pretty sure I tried to do just to_a and found some errors. I cant find them now and it just works with to_a :+1: 
awesome, thanks!
I still have a problem, will comment on that other place.

app/helpers/injection_helper.rb
```diff
@@ -141,7 +141,7 @@ def inject_rails_flash
   end
 
   def inject_json_ams(name, data, serializer, opts = {})
-    if data.is_a?(Array) || data.class.to_s.end_with?("ActiveRecord_Relation")
+    if data.is_a?(Array) || data.class.to_s.end_with?("ActiveRecord_Relation") || data.class.to_s.end_with?("ActiveRecord_AssociationRelation")
```
Matt-Yorkley: I wonder if there's a nicer way of handling these cases that doesn't involve explicitly listing multiple class names. Here we're essentially trying to check if the object being passed is singular or plural, right? I'm not sure, but we could possibly use: 

```ruby
if data.respond_to? :each
```

Otherwise, the `Array`, `ActiveRecord::Relation` and `ActiveRecord::AssociationRelation` classes all include the `ActiveModel::ArraySerializerSupport` module, so we could do this:

```ruby
if data.class.include? ActiveModel::ArraySerializerSupport
```

The result will be true for any of the array-like object we are trying to catch in this conditional check.

Matt-Yorkley: As a side note:
```ruby
object.class.ancestors.include? ActiveRecord::Relation
```
returns `true` for both `ActiveRecord::Relation` and `ActiveRecord::AssociationRelation` objects.

luisramos0: this is outdated Matt, I fixed this now by creating two methods, one for lists and one for single objects. Have a look at the latest version.

app/helpers/injection_helper.rb
```diff
@@ -79,73 +79,75 @@ def inject_current_order_cycle
...
-  def inject_json_ams(name, data, serializer, opts = {})
-    if data.is_a?(Array)
-      opts = { each_serializer: serializer }.merge(opts)
-      serializer = ActiveModel::ArraySerializer
-    end
+  def inject_json_list(name, data, serializer, opts = {})
+    opts = { each_serializer: serializer }.merge(opts)
+    serializer = ActiveModel::ArraySerializer
 
+    inject_json(name, data, serializer, opts)
```
luisramos0, line 147: here.
I havent checked the build yet, but I think we can just do data.to_a here.

app/helpers/injection_helper.rb
```diff
@@ -4,7 +4,7 @@ module InjectionHelper
   include SerializerHelper
 
   def inject_enterprises(enterprises = nil)
-    inject_json_ams(
+    inject_json_list(
```
Matt-Yorkley, line 7: Maybe `inject_json_array` might be a bit more idiomatic than `inject_json_list`?

Matt-Yorkley:
Looks good in general, it just needs to be split into 3 or 4 PRs :+1:

lib/open_food_network/tag_rule_applicator.rb
luisramos0: this is in https://github.com/openfoodfoundation/openfoodnetwork/pull/5919 now

app/helpers/injection_helper.rb
luisramos0: :+1: this is done in https://github.com/openfoodfoundation/openfoodnetwork/pull/5919

luisramos0:
Closing this. I am moving this updated branch to upstream so that we can create PRs on top of it.

Issue #2003, opened 2017-12-20
Flat percent calculator in shipping/payment method only applies to products total (excl. ent. fee)


myriamboure:
## Expected Behavior
When "flat percent" calculator is chosen, the % is supposed to be calculated on the total amount charged in the order ([see user guide to check](https://openfoodnetwork.org/user-guide/producer-set-up-guide/shipping-methods/)), which is logical.

## Actual Behavior
The % is not calculated on the total amount charged in the order, but on the products total cost, excluding any entreprise fee that was added by the hub.

## Steps to Reproduce
1. In "entreprise"  menu, select an entreprise and go in "shipping method"
2. Chose calculator "flat percent" and save
3. In the corresponding shopfront, make an order.
4. Go in "orders", open your order, you will see the detail amount for shipping fee = % x products price

## Gif
![screenshot from 2017-12-20 15-25-08](https://user-images.githubusercontent.com/8135376/34191822-13cd4fbc-e59e-11e7-8ac2-39bb990af173.png)
![untitled presentation](https://user-images.githubusercontent.com/8135376/34191792-eb1aa29a-e59d-11e7-95b7-04f80c7cce43.jpg)

## Context
I was tackling the calculator translations issue and did some test, and realized it was not working as supposed to do.

## Severity
3

Potentially affecting all users using payment or shipping method fee, but there is a workaround, which is to set up a higher % on the shipping and payment fee to include the entreprise fee in them so that the calculated amount is good. I don't know if it's many users or not, could be severity 2 if very few users use that.

## Your Environment
* Version used: 1.10.1
* Browser name and version: Chrome 53.0.2785.116 (64-bit)
* Operating System and version (desktop or mobile): Ubuntu 16.04 LTS

luisramos0:
This looks like a feature to me: "include enterprise fees in shipping fees calculations". 

This calculator comes from Spree and computes on the "items total" (it's how it's called FlatPercentItemTotal) and does not include any adjustments.

Anyway, in terms of implementation, it is probably better to leave FlatPercentItemTotal and create a new calculator FlatPercentIncludingEnterpriseFees...
it will be something like this but will need to look up existing enterprise fees:
https://github.com/openfoodfoundation/openfoodnetwork/blob/187fbd788be9c7f283047ef9ae72fc6715d61ce3/app/models/spree/calculator/flat_percent_item_total_decorator.rb#L11

drummer83:
Hope I got his right...
[...]

PR #5613, opened 2020-06-15
Move all Calculators from spree to OFN and out of the Spree namespace

luisramos0:
#### What? Why?

Closes #5678 
Something I wanted to do for a long time.
We move all needed code from core/app/models/spree/calculator to OFN. And then we move what's in app/models/spree/calculator to app/models/calculator where there are already 2 calculators :ok_hand: 

#### What should we test?
First of all we need to verify that the calculators defined before this PR is deployed will still work correctly after the deployment of this PR. For example, create an order with some fees (with specific  calculators, for example, flat_rate). After deploying this PR, go to admin/orders and press "Update and recalculate fees", the fees should still be correct.

We need to make a simple verification of each calculator in each part of the app where they are used. Maybe not a multiplication of these two things but for example:
- test FlatPercentItemTotal and FatRate and PerItem calculators on a shipping method
- test PriceSack on a payment method
- test FlexiRate on an enterprise fee
- test DefaultTax calculator in Tax Rates admin/tax_rates/ verify the tax is still applied. The code was just moved, not changed, so we don't really need to test all the details here.

#### Release notes
Changelog Category: Changed
Simplified a lot the calculators code used to calculate taxes and fees in OFN.


Matt-Yorkley:
Relevant failing specs and a merge conflict, moving to in dev. Looks good though :smile:

luisramos0:
yes, there was a problem in the last refactoring and also conflicts with #5492, all resolved now. Ready for review again.

db/migrate/20200616162646_move_all_calculators_outside_the_spree_namespace.rb
```diff
@@ -0,0 +1,39 @@
...
+  def revert_calculator(calculator_base_name)
+    update_calculator("Calculator::" + calculator_base_name,
+                      "Spree::Calculator::" + calculator_base_name)
+  end
+
+  def update_calculator(from, to)
+    Spree::Calculator.connection.execute(
+      "UPDATE spree_calculators SET type = '" + to + "' WHERE type = '" + from + "'"
+    )
+  end
```
sauloperez, line 38: I'm just wondering. Why do we do it with plain SQL? do you want to skip updating the timestamps and going through the validations? In that case, AR allows for that and IMO the intent is more clear but that's a minor thing that felt worth sharing.

Pretty good migration :ok_hand:

luisramos0: I definitely don't trust spree code for callbacks and magic, hence SQL :-)
But now I am curious how it would look like with AR. Can you share an example?

luisramos0:
rebased to resolve conflicts.

db/migrate/20200616162646_move_all_calculators_outside_the_spree_namespace.rb
sauloperez: Fair enough but :hankey: why did I ask :trollface:

sauloperez: https://apidock.com/rails/v4.0.2/ActiveRecord/Persistence/update_columns is what you need.

luisramos0: I am sorry for giving you more work, It's just an opportunity for me to learn.
I see what you mean! thanks.
I could change it, but if you dont mind I will go as is, just to save the time of converting things and testing.

luisramos0:
rebased to resolve conflict in schema.rb

db/migrate/20200616162646_move_all_calculators_outside_the_spree_namespace.rb
sauloperez: Yes, don't worry I was just kidding. You know I like sharing this know-how

Matt-Yorkley:
More conflicts...

luisramos0:
ok, rebased to resolve conflicts. Ready for review again.

luisramos0:
build is good, just flaky.

sauloperez:
I just rebuilt it to get a green build. I think we can move it to test ready afterwards.

luisramos0:
It needs a second review, right?

sauloperez:
yes, it does but... last time we hear from someone else was 22 days ago...

luisramos0:
:tada: :tada: :tada: :tada:

filipefurtad0:
Hey @luisramos0 ,

Thank your for the detailed testing notes!

Before staging this PR I've set enterprise fees, of all available kinds:

p1 - Flat percent (per item)
p2 - Weight (per kg)
p3 - Flat rate (per order)
p4 - Flat rate (per item)
p5 - Flexible rate (per item)
p6 - Price Sack


Test cases/steps:

- Added these fees to the respective Order Cycle and placed an order (5 items of each product).
- Checked the order confirmation


Then, I staged this PR:

- Clicked "Updated and recalculate fees" on the previously placed order: the order total remains the same as before - OK
- Placed a similar order as before (5 items of each product): the order total remains the same as before - OK
- Observed that the fees applied per order - p3, p4, p6 - appear as "Admin & Handling" on /cart /checkout and invoice - OK


After setting "charges GST" in the for that hub, and by changing the Tax category of the different enterprise fees (inheriting from product or others) apply taxes over:

- products and enterprise fees
- only products
- only enterprise fees

This worked as expected.


Also checked if calculators still work correctly on shipping/payment methods. I followed your testing notes:

- checked FlatPercentItemTotal and FatRate and PerItem calculators on a shipping method
- tested PriceSack and FlexibleRate on a payment methods

The calculators rendered the expected value for each order; fees operate on the price per item _without_ including the respective enterprise fee, for example, Price Sack, or flat-percent - this is a known issue:
- "flat percent" calculator in shipping/payment method only applies to products total (excl. ent. fee) #2003


- Also, to check cross functionality (between this PR and previous versions of the app) I recalculated various orders, from different hubs, as a superadmin. Everything looked ok.


This PR really touches a lot of corners of the app!
But looks ready to go, so one step closer to "bye bye Spree" :tada:
Thanks Luis!


> Please note that this PR was tested with the following failing spec:

rspec ./spec/models/variant_override_spec.rb:36 # VariantOverride scopes fetching variant overrides indexed by variant does not include overrides for soft-deleted variants

luisramos0:
bam :boom: , merged 
thanks Filipe!

Issue #5619, opened 2020-06-16
[Flaky spec] ./spec/features/consumer/caching/darkwarm_caching_spec.rb:32


sigmundpetersen:
```
  1) Darkswarm data caching caching injected taxons and properties invalidates caches for taxons and properties
     Failure/Error: expect(taxon_timestamp1).to_not eq taxon_timestamp2
     
       expected: value != 1592233121
            got: 1592233121
     
       (compared using ==)
     # ./spec/features/consumer/caching/darkwarm_caching_spec.rb:62:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:129:in `block (2 levels) in <top (required)>'
```

https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/master/builds/1496

Matt-Yorkley:
Ah, I re-enabled this in the Spree upgrade branch when updating the caching code for Rails 4. It's still flaky though.

sigmundpetersen:
This is still failing I'm afraid.

PR #5621, opened 2020-06-16
Set flaky caching spec to pending

Matt-Yorkley:
#### What? Why?

Related to #5619

This flaky spec got re-enabled during the upgrade. It's still flaky. Passing locally but failing intermittently in Semaphore.

There's a tech-debt issue open to fix it.


luisramos0:
I think this spec is important, I think we should move the tech debt issue to dev ready now.
I cant find the issue.

sigmundpetersen:
#5619 (in Tech debt prioritized)

PR #5624, opened 2020-06-16
v3 image precompile

### LINKED ISSUE #5622: Images broken in v3
## Description
Looks like images are broken in v3 in AUS and UK staging servers. We need to investigate.

## Expected Behavior
Product images show up.

## Actual Behaviour
No images.

## Steps to Reproduce
Go anywhere on staging AUS and check product images.

## Workaround
none yet.

## Severity

We need to look at this immediately if we want to release this week.

Matt-Yorkley:
#### What? Why?

Closes #5622

Ensures images in `/assets/images**` are precompiled and moved correctly to `/public/assets/images` in Rails 4.

#### What should we test?

All images load correctly in Rails 4. This search icon seems to be a good test (it's been missing):

Screenshot from 2020-06-16 16-50-17

#### Release notes

Fixed missing image issues in Rails 4 asset compilation

### DIFF
--- a/app/assets/stylesheets/admin/welcome.css.scss
+++ b/app/assets/stylesheets/admin/welcome.css.scss
@@ -7,7 +7,7 @@
 
     @include fullbg;
     background-color: black;
-    background-image: url("/assets/home/tagline-bg.jpg");
+    background-image: image-url("home/tagline-bg.jpg");
     background-repeat: no-repeat;
     background-position: center center;
     margin-bottom: 2em;
--- a/app/assets/stylesheets/darkswarm/_shop-navigation.css.scss
+++ b/app/assets/stylesheets/darkswarm/_shop-navigation.css.scss
@@ -62,7 +62,7 @@ ordercycle {
     }
 
     select {
-      background-image: url('/assets/white-caret.svg');
+      background-image: image-url('white-caret.svg');
       background-size: 30px auto;
       background-position-x: 102%;
     }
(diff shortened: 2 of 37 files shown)


luisramos0:
ah, nice! This will not be related to the access denied problem I have seen in staging AU, right?

Matt-Yorkley:
> This will not be related to the access denied problem I have seen in staging AU, right?

No, I think that's an S3 config issue, but this issue is present there as well. I'm just doing some more tweaking here with asset path helpers, which have apparently also changed a bit in Rails 4...

Matt-Yorkley:
Okay, after updating the image helpers the searchbar icon is now showing and the generated link has changed to this:

Screenshot from 2020-06-16 18-08-00

Note: it didn't have the fingerprint added before.

Matt-Yorkley:
It looks like there's a couple more issues with our use of image helpers in views, I think the fix will be similar to the last commit.

Screenshot from 2020-06-16 18-12-26

luisramos0:
awesome, looking very good!

Matt-Yorkley:
Screenshot from 2020-06-16 18-53-05

filipefurtad0:
Hey @Matt-Yorkley 

Thank you for pointing out the missing search icon. Before the PR:
image

It's back after staging this PR :+1: 
image

Added new images to existing products, these show up in the shop front.

Would be great to test this in staging-AU or to check if the pics appear there after merging, but it looks good for now.
Moving to ready to go.

PS: I was keeping my fingers crossed, but unfortunately this does does not appear to solve #5624. But it does look related with the +1 sign.

Matt-Yorkley:
Thanks @filipefurtad0 I'll merge this and start looking at the maps icons and other potential image issues.

PR #5638, opened 2020-06-18
Tech Debt/5631 [Rails 4.1] Fix spec spec/controllers/api/order_cycles_controller_spec.rb

### LINKED ISSUE #5631: [Rails 4.1] Fix spec spec/controllers/api/order_cycles_controller_spec.rb (13 broken specs)
This error in this build:
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-5611/builds/2

```
 46) Api::OrderCyclesController#products when tag rules apply does not return variants hidden for this specific customer
      Failure/Error: json_response.map{ |product| product['id'] }
      
      TypeError:
        no implicit conversion of String into Integer
      # ./spec/controllers/api/order_cycles_controller_spec.rb:225:in `[]'
      # ./spec/controllers/api/order_cycles_controller_spec.rb:225:in `block in product_ids'
      # ./spec/controllers/api/order_cycles_controller_spec.rb:225:in `each'
      # ./spec/controllers/api/order_cycles_controller_spec.rb:225:in `map'
      # ./spec/controllers/api/order_cycles_controller_spec.rb:225:in `product_ids'
      # ./spec/controllers/api/order_cycles_controller_spec.rb:135:in 
```

If I debug product_ids I see the controller looking for a template:
image

We need some trick in the controller to make it render the json using the serializer provided and not look for a template.

jeduardo824:
#### What? Why?

Closes #5631 

#### Release notes

Changelog Category: Fixed

### DIFF
--- a/app/controllers/api/order_cycles_controller.rb
+++ b/app/controllers/api/order_cycles_controller.rb
@@ -20,7 +20,7 @@ module Api
         search_params
       ).products_json
 
-      render json: products
+      render plain: products, content_type: "application/json"
     rescue ProductsRenderer::NoProducts
       render_no_products
     end
@@ -31,13 +31,15 @@ module Api
         where(spree_products: { id: distributed_products }).
         select('DISTINCT spree_taxons.*')
 
-      render json: ActiveModel::ArraySerializer.new(taxons, each_serializer: Api::TaxonSerializer)
+      render plain: ActiveModel::ArraySerializer.new(
+        taxons, each_serializer: Api::TaxonSerializer
+      ).to_json, content_type: "application/json"
     end
 
     def properties
-      render json: ActiveModel::ArraySerializer.new(
+      render plain: ActiveModel::ArraySerializer.new(
         product_properties | producer_properties, each_serializer: Api::PropertySerializer
-      )
+      ).to_json, content_type: "application/json"
     end
 
     private


jeduardo824:
I don't think this is the perfect solution, I will research more over the weekend. 

I found it here:
https://github.com/rails/rails/issues/23923#issuecomment-189728223

luisramos0:
ah, interesting!
But it does fix the problem, doesnt it?

The error on a few of the specs is another thing, right?
 147) Api::OrderCyclesController#properties loads properties for distributed products in the order cycle
       Failure/Error: properties = json_response.map{ |property| property['name'] }
       
       TypeError:
         no implicit conversion of String into Integer
       # ./spec/controllers/api/order_cycles_controller_spec.rb:166:in `[]'
       # ./spec/controllers/api/order_cycles_controller_spec.rb:166:in `block (4 levels) in <module:Api>'
       # ./spec/controllers/api/order_cycles_controller_spec.rb:166:in `each'
       # ./spec/controllers/api/order_cycles_controller_spec.rb:166:in `map'
       # ./spec/controllers/api/order_cycles_controller_spec.rb:166:in `block (3 levels) in <module:Api>'

jeduardo824:
Yes, it is another thing. This is another endpoint. Probably the fixes are similar, but the problems are not related.

jeduardo824:
@luisramos0 Fixed, now the entire spec should run green!

jeduardo824:
Actually, I investigated the problem a lot on Friday, and I think this is an incompatibility between the gem https://github.com/rails-api/active_model_serializers and Rails. We should start thinking about moving our serializers into something like https://github.com/Netflix/fast_jsonapi

luisramos0:
ah, that explanation makes sense!!!

Yes to the move from AMS. There's a thread here about that: https://community.openfoodnetwork.org/t/alternatives-to-ams/1717
Do you know any of these alternatives? Please share you opinion!

I have created an issue for a fast_json_api POC on OFN: #5654

PR #5659, opened 2020-06-23
Filter orders on inclusive dates in admin/orders

### LINKED ISSUE #5555: Make date filter inclusive of the date selected
## Description
**- As a:** enterprise user or super admin
**- On page:** `/admin/orders?q[s]=completed_at+desc`
**- I want to be able to do:** when selecting dates, I would like the search results to take the end date I've selected into account.

Currently the dates are selected at midnight, which means the start date is taken into account but not the end date. This is confusing for users.

We need to change the date selection to `completed_at_lteq` and `completed_at_gteq`

Reference: https://github.com/activerecord-hackery/ransack/wiki/Basic-Searching#lteq-less-than-or-equal-to

1. Got to `/admin/orders?q[s]=completed_at+desc`
2. Select a start and an end date
3. You should see the orders from these dates included in the search result.
4. Change the date selected and see the result are changed accordingly.

Do this test as enterprise manager, but also as super admin.

dmarcoux:
#### What? Why?

Closes #5555

Dates from the date range weren't inclusive when filtering orders.

#### What should we test?
Test the date range on `admin/orders`.

#### Release notes
Filter orders on inclusive dates in admin/orders

Changelog Category: Changed

### DIFF
--- a/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
+++ b/app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
@@ -23,10 +23,13 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, $timeout, Reque
     $scope.fetchResults()
 
   $scope.fetchResults = (page=1) ->
+    startDateWithTime = $scope.appendStringIfNotEmpty($scope['q']['completed_at_gteq'], ' 00:00:00')
+    endDateWithTime = $scope.appendStringIfNotEmpty($scope['q']['completed_at_lteq'], ' 23:59:59')
+
     $scope.resetSelected()
     params = {
-      'q[completed_at_lt]': $scope['q']['completed_at_lt'],
-      'q[completed_at_gt]': $scope['q']['completed_at_gt'],
+      'q[completed_at_gteq]': startDateWithTime,
+      'q[completed_at_lteq]': endDateWithTime,
       'q[state_eq]': $scope['q']['state_eq'],
       'q[number_cont]': $scope['q']['number_cont'],
       'q[email_cont]': $scope['q']['email_cont'],
@@ -41,6 +44,11 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, $timeout, Reque
       page: page
     }
     RequestMonitor.load(Orders.index(params).$promise)
+
+  $scope.appendStringIfNotEmpty = (baseString, stringToAppend) ->
+    return baseString unless baseString
+
+    baseString + stringToAppend
 
   $scope.resetSelected = ->
     $scope.selected_orders.length = 0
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/admin/orders/controllers/orders_controller.js.coffee
```diff
@@ -23,10 +23,15 @@ angular.module("admin.orders").controller "ordersCtrl", ($scope, $timeout, Reque
     $scope.fetchResults()
 
   $scope.fetchResults = (page=1) ->
+    startDate = $scope['q']['completed_at_gteq']
+    $scope.formattedStartDate = if startDate then moment(startDate).startOf('day').format() else null
+    endDate = $scope['q']['completed_at_lteq']
+    $scope.formattedEndDate = if endDate then moment(endDate).endOf('day').format() else null
```
luisramos0, line 29: 1. have you tried to just use completed_at_lteq with $scope['q']['completed_at_gteq'] ? I thought it could work just with the date.
2. do you really need to usse $scope for $scope.formattedEndDate, wouldnt simply formattedEndDate work?
3. assigments with conditionals and assigning to null are 2 things we should try to avoid, is there a way we can do this without those two things?
Removing $scope and assigment with conditional
```
endDate = $scope['q']['completed_at_lteq']
formattedEndDate = null
formattedEndDate = moment(endDate).endOf('day').format() if endDate
```
and maybe you can just remove the line "formattedEndDate = null" because undefined on formattedEndDate will just work when set on the params?

luisramos0:
Additionally, it would be awesome to have a unit test for this code.
You can basically copy the one there and adapt to these fields. It will be very similar to it:
https://github.com/openfoodfoundation/openfoodnetwork/blob/9e730e6445231c16396de2da7c33bff26c0aed74/spec/javascripts/unit/admin/orders/controllers/orders_controller_spec.js.coffee#L56

dmarcoux:
Could someone help me out with the test? I'm not sure how I can fix it.

luisramos0:
I can give it a try :+1:
First I converted it to unit test: just to test that Order.index was being called with the correct argument.
In this version I had to use moment in the spec as well because of the timezone.
Then I had a look at the SQL generated on the server and verified that the date generated with moment with timezone would not work correctly with ransack/postgres.
I concluded the best way would be to not use moment (no timezone convertion at all) and just send the date with string times appended, postgres does the rest of the work.
So here's how I think we should go forward:
https://github.com/dmarcoux/openfoodnetwork/pull/1/files

filipefurtad0:
Hi @dmarcoux , welcome :tada: 

I staged your PR and checked how the date filters work now, in the orders page.

As seen below, a query with the same start and end date returns the orders from that day:

image

This was not the case before, such a query would return zero results. This was tested both as superadmin and admin.

Thank you so much for this improvement Dany!

dmarcoux:
Thank you to @luisramos0 for helping me out.

luisramos0:
you are welcome!!!

PR #5660, opened 2020-06-23
Migrate all spree pref. keys to the new format

### LINKED ISSUE #5657: User cant checkout - the app is not seeing the app settings stored in the DB
## Description
Katuma is the first v3 deployment.
Originally a checkout problem:
User clicked 7 times on checkout but comunication with Stripe fails.
We found it through bugsnag:
https://app.bugsnag.com/katuma/katuma/errors/5ef1aba43cf9f70017838bec?filters[event.since][0]=30d&filters[error.status][0]=open

But now we see it's a config problem:
image

image

## Steps to Reproduce

1.
2.
3.
4.

## Your Environment

* Version used:
* Browser name and version:

sauloperez:
#### What? Why?

Added by Luis:
Closes #5657

We realized in Spree v2.1 they follow this format instead and this is what's causing issues to Katuma production. 

This will remove the duplicate ones and convert the current preferences to the new thus, keeping the values.

https://github.com/spree/spree/issues/3905 is exactly what happened to us.

#### What should we test?

All instance configurations should contain the correct values after running the migration.

#### Release notes

Fix the spree preferences keys to conform to Spree's v2.1 format

Changelog Category: Fixed

### DIFF
--- a/db/migrate/20200623140437_fix_preferences_keys.rb
+++ b/db/migrate/20200623140437_fix_preferences_keys.rb
@@ -1,6 +1,9 @@
 class FixPreferencesKeys < ActiveRecord::Migration
   def up
-    new_preferences = Spree::Preference.where("key like '/%'")
+    unmigrated_preferences = Spree::Preference.exists?(['key NOT LIKE ?', '/%'])
+    return unless unmigrated_preferences
+
+    new_preferences = Spree::Preference.where("key LIKE '/%'")
     new_preferences.delete_all
 
     Spree::Preference.update_all("key = '/' || key")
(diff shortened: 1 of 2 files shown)


db/migrate/20200623140437_fix_preferences_keys.rb
```diff
@@ -0,0 +1,11 @@
+class FixPreferencesKeys < ActiveRecord::Migration
+  def up
+    new_preferences = Spree::Preference.where("key like '/%'")
```
sauloperez, line 3: All keys that start with a slash got created today.

luisramos0:
I have to share I am a bit concerned about the deletes.
Do you see what happens if you run this migration twice?

sauloperez:
I added bit a of defensive coding as agreed on Slack plus the cache flush.

Issue #4598, opened 2019-12-18
[mobile ux] Add products to the cart redesign


daniellemoorhead:
## In a feature branch or master?
Product listing feature branch 

## Description
**- As a:** shopper
**- On page:** the product list in a shop
**- I want to be able to do:** quickly add or remove products in my cart

## Acceptance Criteria & Tests
- [ ] Standard product listing

**HOW IT LOOKS NOW**
<img width="122" alt="Screen Shot 2019-12-18 at 2 30 47 pm" src="https://user-images.githubusercontent.com/10949052/71053671-4e951380-21a3-11ea-8df9-d33c6090ec1a.png">
<img width="148" alt="Screen Shot 2019-12-18 at 2 31 06 pm" src="https://user-images.githubusercontent.com/10949052/71053652-3ae9ad00-21a3-11ea-8469-77b133e77c64.png">

**HOW IT NEEDS TO LOOK**
<img width="77" alt="Screen Shot 2019-12-18 at 2 31 40 pm" src="https://user-images.githubusercontent.com/10949052/71053679-58b71200-21a3-11ea-812c-ef8bf44cb5d7.png">
<img width="77" alt="Screen Shot 2019-12-18 at 2 31 46 pm" src="https://user-images.githubusercontent.com/10949052/71053683-5a80d580-21a3-11ea-9ded-e405fe4d9fe3.png">

- [ ] Group buy min/max product listing

**HOW IT LOOKS NOW**
On mobile:
![Screen Shot 2020-05-26 at 5 26 52 pm](https://user-images.githubusercontent.com/10949052/82872245-1d9ce400-9f76-11ea-8330-2d59bdcf162d.png)

On desktop:
![Screen Shot 2020-05-26 at 5 26 25 pm](https://user-images.githubusercontent.com/10949052/82872251-1f66a780-9f76-11ea-8bd2-2183d173e767.png)

**HOW IT NEEDS TO LOOK**
On mobile:
![OFN Shop Product List Bulk Order 20200518](https://user-images.githubusercontent.com/10949052/82873350-b849f280-9f77-11ea-912d-b3b75e3170b3.png)

![OFN Shop Product List Bulk Order Modal 20200518](https://user-images.githubusercontent.com/10949052/82871850-8768be00-9f75-11ea-8e05-603418392fa0.png)

On desktop:
[...]

PR #5667, opened 2020-06-25
Stabilise shopping specs and open them for change

mkllnk:
#### What? Why?

Preparing for #4598 and maybe solving #4957.

#### What should we test?

No test, only specs are affected.

#### Release notes

Preparations for more mobile friendly shop fronts.

Changelog Category: Changed

### DIFF
--- a/spec/features/consumer/shopping/shopping_spec.rb
+++ b/spec/features/consumer/shopping/shopping_spec.rb
@@ -104,9 +104,8 @@ feature "As a consumer I want to shop with a distributor", js: true do
             expect(page).to have_content with_currency(1020.99)
 
             # -- Cart shows correct price
-            fill_in "variants[#{variant.id}]", with: 1
-            toggle_cart
-            within(".cart-sidebar") { expect(page).to have_content with_currency(1020.99) }
+            click_add_to_cart variant
+            expect(page).to have_in_cart with_currency(1020.99)
             toggle_cart
 
             # -- Changing order cycle
@@ -119,10 +118,8 @@ feature "As a consumer I want to shop with a distributor", js: true do
             # ng-animate means that the old product row is likely to be present, so we ensure
             # that we are not filling in the quantity on the outgoing row
             expect(page).not_to have_selector "tr.product-cart"
-            within('product:not(.ng-leave)') { fill_in "variants[#{variant.id}]", with: 1 }
-
-            wait_for_cart
-            within(".cart-sidebar") { expect(page).to have_content with_currency(19.99) }
+            within('product:not(.ng-leave)') { click_add_to_cart variant }
+            expect(page).to have_in_cart with_currency(19.99)
           end
 
           describe "declining to clear the cart" do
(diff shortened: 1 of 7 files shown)


spec/support/request/shop_workflow.rb
```diff
@@ -1,13 +1,28 @@
...
+
   def wait_for_cart
-    first("#cart").click
-    within '.cart-sidebar' do
-      expect(page).to_not have_link "Updating cart..."
+    # Wait for debounce
+    #
+    # The auto-submit on these specific form elements (add to cart) now has a small built-in
+    # waiting period before submitting the data...
+    sleep 0.6
```
luisramos0, line 13: I think this should be a method in itself as it was in the shopping_spec, that way the multiline comment becomes a method comment.

spec/support/request/shop_workflow.rb
```diff
@@ -34,6 +49,58 @@ def add_product_to_cart(order, product, quantity: 1)
...
 
+  # Add an item to the cart
+  #
+  # At the moment, the user enters the quantity into an input field.
+  # But with the coming mobile-friendly UX, the user will click a button to
+  # add an item, hence the naming.
+  #
+  # This is temporary code. The duplication will be removed by the mobile
+  # product listings feature. This has been backported to avoid merge
+  # conflicts and to make the current build more stable.
```
luisramos0, line 60: This type of comment should never land on the codebase. The fact that is temporary doesn't need to be there. I'd remove this comment from the code and add it as a comment on the PR.

spec/features/consumer/shopping/shopping_spec.rb
```diff
@@ -322,9 +301,7 @@
         variant.update on_hand: 5
         visit shop_path
 
-        accept_alert 'Insufficient stock available, only 5 remaining' do
-          fill_in "variants[#{variant.id}]", with: '10'
-        end
+        click_add_to_cart variant, 5
```
luisramos0, line 304: Why did the behaviour change? Is there not an alert anymore? why?

spec/support/request/shop_workflow.rb
mkllnk: Yes, I agree. I'll leave it in this case though because the mobile code will remove it again. It won't be needed.

spec/support/request/shop_workflow.rb
mkllnk: My reason for the comment was so that if someone sees the code later that they don't start refactoring it. Somebody may come across the code without seeing the commit. I'll leave it for now. It will be removed soon anyway.

spec/features/consumer/shopping/shopping_spec.rb
mkllnk: Oh, well spotted! That's a mistake. In the new mobile UX it won't be possible to click on the button once the stock limit is reached. But this spec change should happen then, not now. I will change it.

sigmundpetersen:
Merge ready 😊

PR #5677, opened 2020-06-26
Bring ControllerHelpers from Spree and remove spree root path

### LINKED ISSUE #5688: [ByeBye Spree] Move ControllerHelpers to OFN
Move all ControllerHelpers to OFN and merge with decorators.

These files:
- core/lib/spree/core/controller_helpers.rb COPY
- core/lib/spree/core/controller_helpers/auth.rb  MERGE WITH DECORATOR
- core/lib/spree/core/controller_helpers/common.rb COPY
- core/lib/spree/core/controller_helpers/order.rb  MERGE WITH DECORATOR
- core/lib/spree/core/controller_helpers/respond_with.rb  MERGE WITH DECORATOR
- core/lib/spree/core/controller_helpers/ssl.rb COPY

### Context
Part of #4826

### Impact and timeline

luisramos0:
#### What? Why?

Closes #5688 and tries to address bugsnag alert:
https://app.bugsnag.com/open-food-network-belgium/open-food-network-belgium/errors/5ef562d88dd7d4001716d32d?event_id=5ef562d8005b8aa2242a0000&i=sk&m=nw&pivot_tab=event

Brings all ControllerHelpers from Spree.

In ControllerHelpers#Auth, change spree.root to main_app.root.
Delete spree.root route as it is not working. Move /unauthorized from spree to the main app and render itr in OFN's HomeController only. The unauthorized method in ControllerHelpers#Auth will do the redirect either to login or to this unauthrozed route in the home controller now.

And best of all I killed try_spree_current_user :tada: 

I created this issue to remove SSL code but decided to leave it for later: https://github.com/openfoodfoundation/openfoodnetwork/issues/5691

#### What should we test?
We need to run a good sanity check.
Things that could go wrong:
- current_order - user keeps the cart with items even if they log out or change browsers or visit different parts of the website (except changing shop)
- unauthorized redirect
        - we need to try to access specific pages in the app that the user doesnt have access and also access a page that requires login without being logged in. This will be mostly in the backoffice naturally.
        - another useful test will be to try to edit a specific order (direct URL admin/orders/R34343634/edit) that the current logged in user doesn't have access to.
        - another case, similar to the previous will be to access a specific order in the frontoffice (/orders/R316728373) but not as the customer of the order, will the manager of the enterprise of the order see it? Will a user that is not loggeg in be correclty redirected to the login page?
- auto generated api key - is a user without api key generated (see /admin/users) able to navigate the basckoffice normally like search orders in /admin/orders?
- languages - can user change language in the app?

#### Release notes
Changelog Category: Fixed
Fixed routing issue and brought controller code from Spree to simplify and improve it on the OFN side.


lib/spree/core/controller_helpers.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Core
+    module ControllerHelpers
+      def self.included(klass)
+        klass.class_eval do
+          include Spree::Core::ControllerHelpers::Common
+          include Spree::Core::ControllerHelpers::Auth
+          include Spree::Core::ControllerHelpers::RespondWith
+          include Spree::Core::ControllerHelpers::Order
+        end
+      end
```
sauloperez, line 13: I wonder if this not the same as 

```ruby
module ControllerHelpers
include Spree::Core::ControllerHelpers::Common
  include Auth
  include RespondWith
  include Order
end
```

it seems to me.

spec/features/admin/authentication_spec.rb
```diff
@@ -9,17 +9,14 @@
   let!(:enterprise) { create(:enterprise, owner: user) } # Required for access to admin
 
   scenario "logging into admin redirects home, then back to admin" do
-    # This is the first admin spec, so give a little extra load time for slow systems
-    Capybara.using_wait_time(120) do
```
sauloperez, line 13: :see_no_evil:

luisramos0: :rofl:

lib/spree/core/controller_helpers.rb
luisramos0: yes! I thought about it, there is class loading and order of loading. I was not 100% sure it wouldnt cause trouble so I left it as is. There's plenty to improve, I tried to keep to the very simple improvements to reduce risk.

luisramos0:
rebased to resolve conflicts.

spec/features/admin/authentication_spec.rb
mkllnk: I remember that one! That was tricky. Somehow this spec would always fail on CI but work locally. Finally we found that it was the first spec running and that the first spec always need a while to boot up...

filipefurtad0:
Hi @luisramos0,

Performed a sanity check, focusing on the test cases you point out - thank you for providing the valuable test-cases!

1) current_order - user keeps the cart with items even if they log out or change browsers or visit different parts of the website (except changing shop) - all good here

2) unauthorized redirect:

2.a) as a **customer**, trying to access **admin** page /admin/customers; /reports; /order_cycles; /products; orders; /enterprises -> Authorisation Failure
2.b) as an **admin**, trying to access **superadmin** page admin/general_settings/edit; /enterprise_groups -> Authorisation Failure
2.c) another useful test will be to try to edit a specific order (direct URL admin/orders/R34343634/edit) that the current logged in user doesn't have access to.

-> Here, I verified that:
- an admin cannot edit orders from other hubs;
- a customer can only edit their own;
- superadmin edits all

2.d) another case, similar to the previous will be to access a specific order in the frontoffice (/orders/R316728373) but not as the customer of the order, will the manager of the enterprise of the order see it?

-> I checked that the manager of the order as well as the superadmin are able to see the order in the frontoffice (I didn't know about this one, so I checked in another server, prior to staging)

2.e) Will a user that is not logged in be correctly redirected to the login page, if attempting to view his order?

-> When the user is logged out, attempting to view/edit his order (frontoffice) generates error 500, instead of being redirected to the login page:

After this PR (staging-UK):

image

Bugsnag error:
https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5f0f9ebe20d57900175766ac?event_id=5f0f9ebe005bfeec704f0000&i=sk&m=nw



Before this PR - redirection to login (staging-FR):

image


3) auto generated api key - is a user without api key generated (see /admin/users) able to navigate the basckoffice normally like search orders in /admin/orders?

-> All good here. Proceeded as follows:
- superadmin: cleared the key for an existing hub admin 
- hub admin: logged in and browsed through the backoffice
- superadmin: refreshed the page for that user and verified that another API key got generated automatically

4) Languages - can user change language in the app?
All good, the user can change the language, log-out, log back in after clearing cookies/private window and his preferences language are still there.


**Summary**

All looks great, except point 2.e) a visible snail to the user, which appears to be introduced by this PR. I would move this back to In Dev for a second look.

filipefurtad0:
Thank you @luisramos0 ,

Yes, I see it - redirection to login with warning in the background now works, no snail:
image

Looks good to go now.
Thanks!

luisramos0:
great, thanks Filipe.
Yes, let's ask @sauloperez or @mkllnk to approve the last commit before we merge.

sauloperez:
Yep, looks good to me :+1:

PR #5680, opened 2020-06-26
If no enterprises have been geocoded yet make sure Open Street Map displays correctly

cillian:
#### What? Why?

Before if Open Street Map was enabled and no geocoded enterprises were present the map wouldn't display correctly, it would display a gray/blank element. These changes make sure the map displays correctly even if no geocoded enterprises present. It also includes a setting to change the default map coordinates.

This is a start for the last item in #5542 

#### What should we test?

1. Enabling Open Street Map in Admin > Configuration > Content.
2. View the /map page or group map where there are no geocoded enterprises yet.
3. Check the map displays correctly and a gray/blank element is not there.

Extra testing note from Matt: turn Open Street Map off again in the staging server after you've finished testing this please! :sweat_smile: 

#### Release notes

Prevent blank Open Street Map when no geocoded enterprises present.

Changelog Category: Fixed


app/assets/javascripts/darkswarm/directives/open_street_map.js.coffee
```diff
@@ -61,28 +61,47 @@ Darkswarm.directive 'ofnOpenStreetMap', ($window, Enterprises, EnterpriseModal,
...
+      if geocodedEnterprises().length > 0
+        averageAngle("latitude")
+      else
+        openStreetMapConfig.open_street_map_default_latitude || -37.4713077
+
+    initialLongitude = () ->
+      if geocodedEnterprises().length > 0
+        averageAngle("longitude")
+      else
+        openStreetMapConfig.open_street_map_default_longitude || 144.7851531
```
luisramos0, line 101: imo this directive is already too long with too many responsibilities. Could you please extract some of these methods to a service?
can be under app/assets/javascripts/darkswarm/services/
You can extract averageAngle method or maybe the full displayMap method.

If you extract a service you will be able to add some unit tests that are missing.

luisramos0: Another smaller thing, re the number here

144.7851531 is a magic number should be assigned to a variable with a name that explains what it is. Something like this?
defaultLongitude = 144.7851531

do we really need the value here? it is already set as the default on the server side.

cillian: No problem @luisramos0 I extracted out a `MapCentreCalculator` service and added tests which is good because I think the averageAngle method was not working in certain scenarios. Although I haven't done a displayMap service yet.

I also removed the hardcoded latitude and longitude e.g 144.78. You're right it's probably unlikely that we need it.

Thanks

luisramos0: oh, brilliant, that's it, thanks a lot!!
It's the best thing when we decide to write a few tests and end up finding a bug :muscle:

app/assets/javascripts/darkswarm/directives/open_street_map.js.coffee
```diff
@@ -11,34 +11,6 @@ Darkswarm.directive 'ofnOpenStreetMap', ($window, Enterprises, EnterpriseModal,
...
-        averagePositiveAngle
-      else if positiveAngles.length == 0
-        averageNegativeAngle
-      else if averagePositiveAngle > averageNegativeAngle
-        averagePositiveAngle - averageNegativeAngle
-      else
-        averageNegativeAngle - averagePositiveAngle
-
     buildMarker = (enterprise, latlng, title) ->
       icon = L.icon
```
Matt-Yorkley, line 15: Where's this `L` variable coming from? I can't see where it's set...?

app/assets/javascripts/darkswarm/directives/open_street_map.js.coffee
```diff
@@ -61,28 +33,47 @@ Darkswarm.directive 'ofnOpenStreetMap', ($window, Enterprises, EnterpriseModal,
 
     displayMap = ->
       setMapDimensions()
-      averageLatitude = averageAngle("latitude")
-      averageLongitude = averageAngle("longitude")
       zoomLevel = 6
       map = L.map('open-street-map')
       L.tileLayer.provider(openStreetMapProviderName, openStreetMapProviderOptions).addTo(map)
-      map.setView([averageLatitude, averageLongitude], zoomLevel)
+      map.setView([initialLatitude(), initialLongitude()], zoomLevel)
```
Matt-Yorkley, line 39: These two methods are only called in this one place, right? I think I would extract `#initialLatitude` and `#initialLongitude` (and possibly `#geocodedEnterprises`) to the service as well, and just have something like this:
```js
map.setView([MapCentreCalculator.initialLatitude(), MapCentreCalculator.initialLongitude()], zoomLevel)
```

luisramos0:
Moving to In Dev until @cillian 's feedback.

app/assets/javascripts/darkswarm/directives/open_street_map.js.coffee
cillian: @Matt-Yorkley yes that works well. I have moved those method into the `MapCentreCalculator` service now.

app/assets/javascripts/darkswarm/directives/open_street_map.js.coffee
cillian: That `L` is the Leaflet JS library, it gets included here https://github.com/openfoodfoundation/openfoodnetwork/blob/master/app/assets/javascripts/darkswarm/all.js.coffee#L11

app/assets/javascripts/darkswarm/services/map_centre_calculator.js.coffee
```diff
@@ -0,0 +1,36 @@
...
+
+    initialLongitude: =>
+      if Enterprises.geocodedEnterprises().length > 0
+        @_calculate("longitude", Enterprises.geocodedEnterprises())
+      else
+        openStreetMapConfig.open_street_map_default_longitude
+
+    _calculate: (angleName, coordinates) =>
+      positiveAngles = []
+      negativeAngles = []
```
sauloperez, line 18: these two arrays can be removed right? We append values to them but I don't see them used.

sauloperez:
Pretty neat @cillian !

app/assets/javascripts/darkswarm/services/map_centre_calculator.js.coffee
cillian: @sauloperez ah yes, I should have spotted that, it's fixed now in 
e6ab2ae

luisramos0:
Verifying the issue before staging this.
I deleted all longitude and latitude data from staging katuma DB with this SQL.
image


Google maps centers around melbourne by default with no enterprise.
Checking the OSM option in the content settings, I see a grey OSM map:
image

After deployment, google maps still works and OSM now shows the default location in AUS (it's not so zoomed in as in google maps).
I set the address of an enterprise to Barcelona and the map moves there and shows the enterprise.
I deactivated OSM after this, google maps showing the enterprise in Barcelona :+1: 

This is ready to go :heavy_check_mark:

PR #5697, opened 2020-06-29
Quick css fix in main menu

### LINKED ISSUE #5671: Styling in main menu (Superadmin)
## Description

While testing translation PR #5589 I noticed the item "More" on the main menu is slightly misplaced. This appears only to affect  Superadmin in Language dashboard.

## Expected Behavior

Menu should look like this ("More" should appear translated, in production):
image

## Actual Behaviour

image

## Steps to Reproduce

1. Log in as Superadmin and check how the main menu looks like in French language.

## Severity

## Your Environment

* Version used:
* Browser name and version:

## Possible Fix

Worth checking how does it look like in production, with the due translation key.

renanbona:
#### What? Why?

Closes #5671 

BEFORE the changes:
before_the_changes_french

AFTER the changes:
new_menu_french
new_menu_english

#### What should we test?

Access the `/admin` page and the main menu should be ok with both langues (English and French)

#### Release notes

Fix css in main menu (Superadmin) 

Changelog Category: Fixed

### DIFF
--- a/app/assets/stylesheets/admin/components/navigation.scss
+++ b/app/assets/stylesheets/admin/components/navigation.scss
@@ -67,15 +67,17 @@ nav.menu {
 
   li {
     min-width: 90px;
+    font-size: 12px;
 
     a {
-      display: block;
+      display: flex;
+      justify-content: center;
       padding: 25px 20px;
       color: $color-1 !important;
       text-transform: uppercase;
       position: relative;
       text-align: center;
-      font-weight: 600;      
+      font-weight: 600;
 
       i {
         display: inline;


luisramos0:
Thanks for your PR!

I am not sure about reducing the fontsize. I thought we were using a plugin that adds entries to MORE if they dont fit. So the solution here would be to make one extra entry like customers go into the MORE submenu.

renanbona:
OK, I will check this plugin and see if I cant make the customers option go to the MORE submenu.

luisramos0:
I remember now, it's vendor/assets/javascripts/jquery.adaptivemenu.js

renanbona:
I've been doing some tests in the master branch, and apparently this bug has already been fixed. So I'm thinking about closing this PR. Can you have a quick look and see if you can reproduce this bug? @luisramos0

luisramos0:
I think the problem is still there but only in some circumstances? I know you have to be super admin and maybe it's only in French. And if it works in French superadmin I think there will be another condition that is causing this sometimes...

RachL:
heya @renanbona are still working on this issue or should we close your PR for someone else to pick it up?

renanbona:
Hey @RachL I'll work on this again this Friday.

renanbona:
Actually this bug is already fixed in this PR https://github.com/openfoodfoundation/openfoodnetwork/pull/5658/files#diff-e35182d90157bf50b09e78cb6a31defaR2808. The translation of word "More" was added in `fr.yml`. So I'm closing this PR because I've tested in both browsers Google Chrome and Mozilla in different monitors and the bug is already fixed. 

main_menu_fixed

PR #5699, opened 2020-06-29
Make button border radius consistent

sauloperez:
#### What? Why?

This makes this button consistent with the others next to it. From

Screenshot from 2020-06-29 12-00-39

to

Screenshot from 2020-06-29 12-00-59

#### What should we test?

The "saved order" button from the order confirmation page should have the same border radius as the other two buttons

#### Release notes

Fix radius of order saved button in the order confirmation page

Changelog Category: Fixed


filipefurtad0:
Hi @sauloperez ,

This seems to work fine:

Before this PR
image

After this PR
image

Ready to go.

Issue #5643, opened 2020-06-19
Rails 4 fingerprinted assets


Matt-Yorkley:
We still have some issues with non-fingerpritned assets being called from the codebase. Rails 4 basically does not do non-fingerprinted assets, at all.

More background here: https://github.com/openfoodfoundation/openfoodnetwork/issues/5622#issuecomment-645891351

So basically it looks like we still have a bunch of places in the codebase where we are calling assets in a way that will not include the fingerprint, and we need to fix that.

Matt-Yorkley:
It looks like this applies to fonts as well as images, and we'll need to check basically anything that's in the `/app/assets/*` directories.

sauloperez:
I started with https://github.com/openfoodfoundation/openfoodnetwork/pull/5644

filipefurtad0:
Hi - Just a brief feedback from my side: During testing of the release I haven't found anything major missing from the app.

sauloperez:
@luisramos0 I think that all assets work so far due to the combination of ofn-install's (dangerous) symlink plus assets being placed in `public/` that @Matt-Yorkley described in the other issue. But it's a bit messy and makes all feel uneasy and lower our confidence so we better get as close to Rails default as we can.

Matt-Yorkley:
> This is bug right? There will be images and icons missing from the app?

It's a worrying issue, but it shouldn't actually affect current OFN production servers at the moment, unless we start adding new assets or changing old ones. It's a bit complicated but we should probably fix it.

Matt-Yorkley:
Basically the work that needs to be done will look a bit like [this PR](https://github.com/openfoodfoundation/openfoodnetwork/pull/5635), but for some of the other assets (like fonts and maybe some other images) that are still being generated with without fingerprints, as "non-digest assets" are now completely ignored by Rails.

Matt-Yorkley:
As a side note, on any _new_ servers that are deployed, lots of assets will be broken for sure until we fix this.

sauloperez:
It think it'll be fairly straightforward to find out which assets are processed with the `assets:precompile:nondigest` rake task in Rails 3.2. https://github.com/rails/rails/blob/e17e25cd23e8abd45b1706463dd57c90fa6dcb7c/actionpack/lib/sprockets/assets.rake#L33-L57 is the code responsible for doing so.

luisramos0:
[...]

PR #5701, opened 2020-06-29
Noimage

Matt-Yorkley:
#### What? Why?

Part of #5643 

Moves "noimage" images (which are used when a product has no image uploaded yet) under /public so they can be accessed easily in non-digest form.

#### What should we test?

Images for products and variants that are shown when they have no uploaded image (the default OFN logo) should be working in the shopfront, cart dropdown, cart page, and admin products page.

#### Release notes

Adjusted default product image paths for Rails 4.


Matt-Yorkley:
Working as intended on UK staging.

app/views/spree/admin/variants/_autocomplete.js.erb
```diff
@@ -4,7 +4,7 @@
       {{#if variant.image }}
         <img src='{{variant.image}}' />
       {{ else }}
-        <img src='<%= image_path("noimage/mini.png") %>' />
+        <img src='/noimage/mini.png' />
```
sauloperez, line 7: but don't we need Rails url helpers so the fingerprinted URL is picked up?

Matt-Yorkley: I made a slightly different choice for this specific case. These default images are used in various API endpoints, serializers and also in javascript files where generating the fingerprint is a bit messy. Moving them under /public means we can use the non-fingerprinted versions of these images (which is what we were doing before, anyway).

What do you think?

sauloperez: I see. That's good enough

filipefurtad0:
Hey @Matt-Yorkley ,

I decided to stage this in another server, I took staging-FR.

Checked that the logo is there when no pic is associated with that item on:

- Shopfront
image


- Cart-dropdow

image

- Admin Products Page

image

- Cart Page

The logo seems to be missing here:

Staging-FR
image

Indeed in staging UK it looks fine:

Staging-UK
image

I'm re-staging this PR in UK and post again in a few moments.

Matt-Yorkley:
Yes, there's a bit missing. Thanks for spotting it!

I'm adding another commit now that should fix it :+1:

filipefurtad0:
Just checked in UK, after re-staging:
image

Matt-Yorkley:
@filipefurtad0 the latest build should be good :+1:

@sauloperez and @luisramos0 I've added 2 extra commits here :point_up:

filipefurtad0:
All good now @Matt-Yorkley ,

image

Good as well on the previous spots on the app:
image

image

Good to go!

filipefurtad0:
Short note:
I realize now I skipped the code review step before testing moving to ready to go. I hope this is ok! Ping @luisramos0 @sauloperez

app/views/spree/orders/_bought.html.haml
```diff
@@ -16,10 +16,10 @@
       %td.cart-item-description
 
         %div.item-thumb-image
-          - if variant.images.length == 0
-            = mini_image(variant.product)
+          - if variant.product.images.length == 0
+            = image_tag("/noimage/mini.png")
```
luisramos0: AFAIK the current status re variant images is that variant specific images can be added to the system in the product images page BUT they are not used anywhere. SO, we can make all our code go to variant.product.images ONLY. Which is what you are doing here. So I think this is ok, just sharing my thoughts :-)

filipefurtad0:
Hi again @luisramos0 @Matt-Yorkley ,

I've found something which may be related to this PR, on the admin orders page /admin/orders/Rxxxxx/edit:

image

While testing I'm pretty sure I did not test this, didn't think of it :-/
Opening an issue and linking to this one.

PR #5703, opened 2020-06-30
Checkout addresses

### LINKED ISSUE #5702: Require timestamps on all models that have them

- [ ] Replace all code that sets timestamps to `nil`.
- [ ] Change all timestamp columns to disallow null values.

### Context

Rails used to have mandatory timestamps by default (e.g. `created_at not null`). Rails 4 dropped the default value and we had some code crashing when it tried to set a timestamp to null. So we relaxed the timestamp requirement and made it optional. But since we don't know which other bugs that could introduce and Rails 5 defaults to enforcing timestamps again, we should fix our code and enforce timestamps again. The full discussion is on #5663.

### Impact and timeline

Since we don't know if the relaxed timestamps can cause other bugs, we should do this as soon as possible. Possible bugs could be:

- Orders are not displayed if they don't have a timestamp.
- Whole reports could be empty if an order without timestamps is involved.
- Certain pages could show a snail because they try to display a timestamp that doesn't exist.
- ...

Matt-Yorkley:
#### What? Why?

Closes #5702 
Related to #5663 and possibly to #5705

- Set timestamps to not be nullable again.
- Fixes problems with updating default shipping / billing addresses during checkout processing.

So I think a nasty error was happening (due to the way we were updating default address records) part-way through the transaction for placing and order, which was then aborted/rolled back in a messy way (causing timestamps on _associated_ objects to be NULL), and possibly triggering other issues within postgres itself (see below). Basically there was some really, really nasty stuff happening here that can only be descried as a **fail cascade**.

There's some related background in this Rails issue in a case where an exception is raised during a transaction at the point that postgres is dealing with **caching important values relating to it's internal state** (which we actually see some evidence of in the stacktraces on Katuma) and essentially all hell breaks loose: https://github.com/rails/rails/pull/17607

Other related reading: Rails 4 has introduced **breaking changes to the behavior of #clone** (apparently without mentioning it in the upgrade guide :unamused: ), which I think is what caused the problematic exception in this case. See: https://github.com/rails/rails/blob/4-0-stable/activerecord/lib/active_record/core.rb#L217-L220

We didn't see this in the build before, _mostly_ because of this: https://github.com/openfoodfoundation/openfoodnetwork/pull/5663#issuecomment-651655981 and partly because of a tiny bit of missing spec coverage for a particular case.

#### What should we test?

We need to very carefully test the "save as default billing address" and "save as default shipping address" checkout options with:
- guest checkout
- registered user checkout where the user does not already have default addresses saved
- registered user checkout where the user does have default addresses saved

Update: it looks like it's actually just the scenario where the user already has default addresses saved, and they select the "save as default shipping/billing address" option again, which updates their existing default addresses.

#### Release notes

Fixed bug with updating default addresses during checkout in v3


Matt-Yorkley:
I just added some explanatory notes in each commit :+1:

Matt-Yorkley:
I now have feature specs that replicate this bug, and fail horribly with the error we've seen. The failing specs are fixed by the commit in this PR.

spec/features/consumer/shopping/checkout_spec.rb
```diff
@@ -94,6 +94,32 @@
...
+
+            expect(order.reload.bill_address.address1).to eq '123 Your Head'
+            expect(order.reload.ship_address.address1).to eq '123 Your Head'
+
+            expect(order.customer.bill_address.address1).to eq '123 Your Head'
+            expect(order.customer.ship_address.address1).to eq '123 Your Head'
+
+            expect(user.reload.bill_address.address1).to eq '123 Your Head'
+            expect(user.reload.ship_address.address1).to eq '123 Your Head'
+          endI now have tests
```
luisramos0, line 120: typo here

luisramos0:
wow!!!! I had to give it a try, I can confirm, you are right!!!! :clap: 

I added the spec to master and see it passing!
I added the constraints and see the spec failing
image

and then added your fix and it's green again.

:clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap:

luisramos0:
This will be the cause and fix to the multiple stripe payments right Matt?
https://github.com/openfoodfoundation/openfoodnetwork/issues/5705

spec/features/consumer/shopping/checkout_spec.rb
Matt-Yorkley: Removed :+1:

Matt-Yorkley:
> This will be the cause and fix to the multiple stripe payments right Matt?

I really, really hope so.

spec/features/consumer/shopping/checkout_spec.rb
Matt-Yorkley: I force-pushed out of habit. Sorry Luis! :see_no_evil:

luisramos0: it's ok! just dont change the magic spec!

sauloperez:
But the root cause was the issue with `#clone` and the unordered rollback made it just hard to spot, right?

Matt-Yorkley:
I think problems due to the #clone issue caused an exception, which triggered a rollback, which fucked up, and the combination of some or all of the above _possibly_ then caused an internal issue in postgres itself, on top of all that.

sigmundpetersen:
So this is Test ready then? I guess we move it to the top as well?

Matt-Yorkley:
I've adjusted the migration here to guard against NULL values whilst the table's NULL constraint is being altered. It needs some re-review and then it would be good to test deploying it to a staging server that _does_ have some NULL timestamps.

luisramos0:
In uk live there are for example orders without these values, I think we will be ok :+1:

Let's get another dev to review  this before we test. What can go wrong?

Matt-Yorkley:
Interesting, there is a single order in UK Prod with NULL values in timestamps. I assume that order could create problems...

db/migrate/20200630070422_require_timestamps.rb
```diff
@@ -0,0 +1,253 @@
+class RequireTimestamps < ActiveRecord::Migration
+  def up
+    current_time = Time.now
```
sauloperez, line 3: Have you checked what's the actual value that the ALTER TABLE statement gets called with? It might not matter much in this context but we should always use Rails' `Time.zone.now`. It's a bit fuzzy in my memory but I remember `Time.now` was a no. What I do know is that without `zone` is the only way to include time-zone information. According to https://thoughtbot.com/blog/its-about-time-zones we should let AR do the UTC conversion.

Matt-Yorkley:
> Before that method, you had to change the schema in graduall and backward-compatible steps.

Yeah, it's new in Rails 4 :tada:

It would have been a huge pain to do this nicely without this option...

sauloperez:
> It would have been a huge pain to do this nicely without this option...

well, it required patience more than anything

sauloperez:
I tested this locally and it works as intended. An order without timestamps (I set them to null from psql) ends up with timestamps set to the current time in UTC and `not null` column modifiers :ok_hand:

filipefurtad0:
Hey @Matt-Yorkley ,

Checked the 3 scenarios you mentioned, focusing on Stripe-SCA payments:

i) guest checkout - OK
ii) registered user checkout where the user does not already have default addresses saved - OK
iii) registered user checkout where the user does have default addresses saved - OK

In iii) I tested 8 different options/combinations of the checkboxes on checkout, inserting new addresses when required
- Save as default billing address 
- Shipping address same as billing address?
- Save as default shipping address

These appear as expected in the dashboard on Stripe.

Also inserted a bit random and unusual characters (on addresses, names). In all cases checkout proceeded as expected.

Ready to go.

PR #5707, opened 2020-06-30
Add space after translations so that translators dont have to guess a space is needed

luisramos0:
#### What? Why?
This was seen in Brasil while the translations are correct without space at the end.

In the welcome email after confirmation:
image

In the shops page:
image

There should spaces introduced by the code, not the translations.

We will need to remove the extra space in transifex for the languages where the space was added in transifex, the translation keys are: welcome_to and closing and, after this PR, they should NOT have extra space in the end.

#### What should we test?
Make sure the welcome email in pt-BR will come with a space and also /shops displays correctly.

#### Release notes
Changelog Category: Fixed
Corrected spacing issue in email subject and /shops

### DIFF
--- a/app/assets/javascripts/darkswarm/filters/dates.js.coffee
+++ b/app/assets/javascripts/darkswarm/filters/dates.js.coffee
@@ -12,4 +12,4 @@ Darkswarm.filter "sensible_timeframe", (date_in_wordsFilter)->
     if moment().add(2, 'days') < moment(date, dateFormat)
       t 'orders_open'
     else
-      t('closing') + date_in_wordsFilter(date)
+      t('closing') + ' ' + date_in_wordsFilter(date)
--- a/app/mailers/spree/user_mailer.rb
+++ b/app/mailers/spree/user_mailer.rb
@@ -20,7 +20,7 @@ module Spree
       @user = user
       I18n.with_locale valid_locale(@user) do
         mail(to: user.email, from: from_address,
-             subject: t(:welcome_to) + Spree::Config[:site_name])
+             subject: t(:welcome_to) + ' ' + Spree::Config[:site_name])
       end
     end
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -207,8 +207,8 @@ en:
         explainer: Automatic processing of these orders failed for an unknown reason. This should not occur, please contact us if you are seeing this.
 
   home: "OFN"
-  title: Open Food Network
-  welcome_to: 'Welcome to '
+  title: "Open Food Network"
+  welcome_to: "Welcome to"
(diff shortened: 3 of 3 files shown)


RachL:
This is awesome 😍

sauloperez:
I think this needs a rebase to get the build green @luisramos0

luisramos0:
yes, rebased :+1:

sauloperez:
I don't think this needs to wait for a second approval and it'll make FR and Katuma happy.

filipefurtad0:
Hey @luisramos0 ,

I'm thinking how to test the Brazilian translation: these settings are defined on deployment or the PR, right? I'm not sure on how to do this, happy for further advice here. Thanks!

RachL:
@filipefurtad0 there is a space problem on private shop while you visit them when you are not loggedin.

So maybe if you try to test this on staging it will be fixed now?

luisramos0:
I forgot to remove the spaces from the base en.yml file. Done now.
I think it's fine to test this as a normal translations issue, just make sure it looks good in staging.

filipefurtad0:
Thanks @RachL and @luisramos0 ,

I was unsure in this case, since the testing notes mention specifically to test the email in pt-BR. On the other hand, and as mentioned in the description of this PR, these changes are introduced by the code, so they should be observable in the email confirmation subject in English language, which they are:

- before this PR (second line, star-underlined); after this PR (first line, not underlined):
image

I was expecting the change in en.yml to take immediate effect:
from  `welcome_to: 'Welcome to '` to  `welcome_to: "Welcome to"`

since this is the case on the closing shop info in /shops which look the same before and after the PR:

Before:
image

After:
image

The code line adds a space:
t('closing') + ' ' + date_in_wordsFilter(date)

And the change in en.yml:
  `closing: "Closing "` ->  `closing: "Closing"`
removes it.

This is similar to:
`welcome_to: 'Welcome to '` ->  `welcome_to: "Welcome to"`

The only difference is `'` vs. `"`. So I guess the double inverted-comma is a requirement for a Transifex key. In the "welcome" case a new key is introduced (and will only take effect later on in production) while in the "closing" case a change in an existing key takes immediate effect, and is visible in staging. Is this correct?

I think this is good to go.


Also mentioned on the **What? Why?** section:

> We will need to remove the extra space in transifex for the languages where the space was added in transifex, the translation > keys are: welcome_to and closing and, after this PR, they should NOT have extra space in the end.

I am opening a translation issue in this regard.

Thank you!

luisramos0:
>>> The only difference is ' vs. ". S

I am not sure what's the difference but one other difference is that one is ruby the other is javascript, I think that will be the reason for the difference.

PR #5709, opened 2020-06-30
Replace empty? with blank? and replace unless empty? with if present?

### LINKED ISSUE #5648: Logout during Subscription creation
During release testing Filipe encountered a possible issue whilst creating a subscription: at the point of adding a product, got logged out unexpectedly and whilst logging back in, this error appeared in Bugsnag: 
```
NoMethodError: undefined method `empty?' for nil:NilClass
Location: app/services/permitted_attributes/subscription.rb:10 - call
```

After logging in again subscriptions seemed to work correctly, and the issue wasn't reproduced(?). 

We might need to adjust the logic a bit here:

https://github.com/openfoodfoundation/openfoodnetwork/blob/dd9d9e331677539783be44015867b08bb410314c/app/services/permitted_attributes/subscription.rb#L10

luisramos0:
#### What? Why?
From [here](https://apidock.com/rails/Object/blank%3F):
image

and present? is just !blank?

Closes #5648
This does not address the unexpected logout in 5648 just the error part, after this PR the error, if there is one, will be different.
I wonder if we should close #5648 and wait for another occurrence or bugsnag error OR we leave it open. Thoughts?

#### What should we test?
I am not sure we need to test this manually. I think this is a pr-no-test as the changes are all logically equivalent.

#### Release notes
Changelog Category: Changed
Made the code a bit more resilient and able to handle edge cases with nil values in variables.

### DIFF
--- a/app/controllers/admin/enterprises_controller.rb
+++ b/app/controllers/admin/enterprises_controller.rb
@@ -146,7 +146,7 @@ module Admin
         enterprises = OpenFoodNetwork::OrderCyclePermissions.new(spree_current_user, @order_cycle)
           .visible_enterprises
 
-        unless enterprises.empty?
+        if enterprises.present?
           enterprises.includes(
             supplied_products:
               [:supplier, master: [:images], variants: { option_values: :option_type }]
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -241,7 +241,7 @@ class CheckoutController < Spree::StoreController
   def update_failed(error = RuntimeError.new(order_error))
     Bugsnag.notify(error)
 
-    flash[:error] = order_error if flash.empty?
+    flash[:error] = order_error if flash.blank?
     checkout_failed
     update_failed_response
   end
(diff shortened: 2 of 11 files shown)


Matt-Yorkley:
Lots of failures in the build here... 

`#empty?` has a slightly different implementation in the `Enumerable` class, right? So the results might be slightly different if the object it's being called on is a hash or an array ass opposed to something that just inherits from `Object`?

luisramos0:
I think the changes are logically equivalent because blank will always call empty and then empty will have specific implementations that were already being used.

I think the build is broken because of problems in master. I rebased now. Let's have a look at this new build.

luisramos0:
I found a little problem here:;
image

@order_cycles is not populated in the cart (orders_controller) but this helper method is used in the cart views. Currently, when @order_cycles is nil, no_open_order_cycles? returns nil :see_no_evil:

I am ignoring this problem for now.

sauloperez:
I had a boolean logic overflow in my brain @luisramos0 :trollface: I trust you on that.

I have no clear idea about keeping #5648 open or not but we could perhaps leave it open while we work on v3 regressions? It is anyway that happened for which we don't have an explanation.

sauloperez:
I take it you don't have any objection @Matt-Yorkley . Feel free to move it back if that's not the case.

sigmundpetersen:
So everyone agrees this can be merged without test?

sauloperez:
Ah I missed @luisramos0 tagged it with no-test. Let's wait a bit to hear from @Matt-Yorkley then.

Matt-Yorkley:
:ok_hand:

PR #5712, opened 2020-07-01
Add payment details to invoice

### LINKED ISSUE #5208: Invoices display incorrect information about payments if order has been edited
## Description

Changes to an order that has been paid for result in incorrect information on the invoice - it either shows 'paid' or 'unpaid' and does not reflect payments that have been made

I have an order for $26 and the payment has been collected

The invoice correctly shows PAID

I remove an item, leaving the order $8 in credit as shown on the Payments tab

Orders now shows me Credit Owed (albeit not the amount, see #5173)

But if I regenerate the Invoice it still just shows as PAID

I add items to the order so that there is a BALANCE DUE 

and now the new Invoice just shows NOT PAID and doesn't recognise that any payment has been made

I think there is a very simple (papercut?) way to resolve this . . see below

## Expected Behavior

Changes to the order are reflected in correct payment information on a newly generated Invoice

1. Payment Summary box should include the Payment State of the Order, not just Paid or Not Paid
- If the order has a Balance Due, the right hand side of the blue box should say 'Balance Due ($xx)'
- If the order is in Credit, the right hand side of the blue box should say 'Credit ($xxx)'

2. Payment history on this order should be included on the invoice. This can just be included under the Payment Summary, with a title 'Payment History' and then just show the information from the payments tab

image

I think this information could replace what is currently there as it will be more up-to-date and informative. The only thing that would be lost is the descriptive text in the payment method. We could include that in the table of payment information if we thought necessary? @tschumilas @RachL @lin-d-hop ?

## Actual Behaviour
Invoices generated after an order has been changed are totally unhelpful and uninformative re. additional payments required or credits

## Steps to Reproduce

1. Create an order and capture the payment
2. Generate invoice 
3. Amend the order in a way that puts it in balance owing or credit owed
4. Generate invoice - see incorrect information

## Workaround
No workaround, the invoice is just wrong. The only thing we can tell people is to create an invoice outside OFN

## Severity
S3 - we've lived a long time with this and pushed people out of ofn for invoicing, but with our current influx this is something we have been hit with many times this week. 

For the simplicity of the fix perhaps S2 - if invoicing is considered a critical feature

## Possible Fix
As above

NB. Couldn't decide whether this was a bug or a papercut so just went with bug as in invoices simply displaying incorrect information

mbudm:
#### What? Why?

Closes #5208

Add more detail on payment history to invoices. This helps to clarify the invoice when the status is balance due or credit owing.

Concerns/possible changes are:
- the payment partial is in the order_mailer view. The change will apply to other templates using this partial (emailed invoices?) I need to clarify if this expected/desired.
- I added the css for the payments table to the same `scss` file where the invoice table is defined, `app/assets/stylesheets/mail/email.scss` which feels a bit like a dumping ground - should we create a separate scss file for the invoice styles?
- I'm not sure about the column alignment, I've suggested in the issue that the amount column is moved to the right of the payment method column, for better balance and easier scanning of the figures
- If there are no payments, should the info currently added to the invoice be shown? Also I'm unsure if there are ever orders without payments, as a checkout must occur for an order to appear?

#### What should we test?
Detailed steps to recreate are in the issue: https://github.com/openfoodfoundation/openfoodnetwork/issues/5208
- create an order that is balance due or credit owing
- print the invoice
- the balance due or credit owing state is shown, including the amount. the payment history is also shown

Screen Shot 2020-07-01 at 10 46 35 am

A paid (completed) order - with a failed credit card attempt:
Screen Shot 2020-07-01 at 12 16 28 pm

#### Release notes
- Changed order invoices where balance owing or credit due, with details on payment history

Changelog Category: Changed


sigmundpetersen:
FYI those failing enterprise_controller specs are also in master and will be fixed by #5706

mbudm:
Thanks @sigmundpetersen - I think the failing `./spec/mailers/subscription_mailer_spec.rb` are on this change though so will review that tomorrow

app/views/spree/admin/orders/_list.html.haml
```diff
@@ -0,0 +1,16 @@
+%table.payment-list
```
luisramos0, line 1: this is the payment list so the partial would be better named as _payments_list?

app/views/spree/admin/orders/_list.html.haml
```diff
@@ -0,0 +1,16 @@
...
+  %thead
+    %tr
+      %th= "#{Spree.t('date')}/#{Spree.t('time')}"
```
luisramos0, line 4: always use lazylookups for translations and never Spree.t (it adds the spree namespace), always use I18n or in this case, just t
this would be t('.date') and t('.time') or maybe just t('.datetime') as it is specific to this template.
and then you can add the key to en.yml in spree.admin.order.payments_list.date around line 3236 in en.yml

app/views/spree/admin/orders/_list.html.haml
```diff
@@ -0,0 +1,16 @@
...
+      %th.amount= t(:amount)
+      %th= t(:payment_method)
+      %th.payment-state= t(:payment_state)
```
luisramos0, line 7: same here, please use lazy lookups like t('.amount')

app/views/spree/admin/orders/_list.html.haml
```diff
@@ -0,0 +1,16 @@
...
+  %tbody
+    - payments.each do |payment|
+      %tr
+        %td= pretty_time(payment.created_at)
+        %td.amount= payment.display_amount.to_html
+        %td
+          %span.payment-method-name= payment_method_name(payment)
+          %span.payment-method-description= payment.payment_method.description
+        %td.payment-state= t(payment.state, scope: :payment_states, default: payment.state.capitalize)
```
luisramos0, line 16: there's a missing empty line.
in this case, because it's a list of standard payment states I think lazylookup can be skipped.

app/views/spree/order_mailer/_payment.html.haml
```diff
@@ -3,12 +3,15 @@
     - if @order.paid?
       = t :email_payment_paid
     - else
-      = t :email_payment_not_paid
+      - if @order.outstanding_balance?
+        = @order.outstanding_balance < 0 ? t(:credit_owed) : t(:balance_due)
```
luisramos0, line 7: is this code not already somewhere else? this is logic that should not be in a view but in a service, can we extract to service and call it from here?

app/views/spree/order_mailer/_payment.html.haml
```diff
@@ -3,12 +3,15 @@
...
-%h4
-  = t :email_payment_method
-  %strong=  last_payment_method(@order)&.name
-%p
-  %em= last_payment_method(@order)&.description
-%p &nbsp;
+- if @order.payments.any?
+  = render partial: 'list', locals: { payments: @order.payments }
+- else
+  .alpha.twelve.columns.no-objects-found= t(:order_has_no_payments)
```
luisramos0, line 17: this should use lazy lookup

app/views/spree/order_mailer/_payment.html.haml
```diff
@@ -3,12 +3,15 @@
...
   %strong
     = t :email_payment_summary
-%h4
-  = t :email_payment_method
-  %strong=  last_payment_method(@order)&.name
-%p
-  %em= last_payment_method(@order)&.description
-%p &nbsp;
+- if @order.payments.any?
+  = render partial: 'list', locals: { payments: @order.payments }
```
luisramos0, line 15: the invoices views are calling this partial in order_mailer and then this partial is calling the _list partial in /orders again?
if list is used here in order_mailer it should be here.
I think you can simply move /orders/_list.html.haml  to /order_mailer/_payment_list.html.haml
Is the changed supposed to go into the order emails as well?

app/assets/stylesheets/mail/email.scss
```diff
@@ -107,6 +109,49 @@ table.order-summary {
   }
 }
 
+// put in separate order invoice scss file?
+table.payment-list {
+  width: 100%;
```
luisramos0, line 114: "should we create a separate scss file for the invoice styles?"

yes, I think this should go to a separate file :+1:

I'd keep the reference to emails in the name as invoices are only using email partials. Unless you want to create partials that are specific to invoices.

app/views/spree/admin/orders/_list.html.haml
luisramos0: more info here: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)#development

mbudm: Cool - thanks for the link to the specific wiki docs, that helps a lot. I updated the docs a bit to make them more clear

app/views/spree/order_mailer/_payment.html.haml
mbudm: The `/order_mailer/_payment_list.html.haml` partial is shared currently between the order email and the order invoice. And what I'm doing here is replicating the table from `app/views/spree/admin/payments/_list.html.haml` in the invoice and email.

I assume that the email should also be changed as the same problem exists there, in credit owing and balance due situations the info is incorrect.

What is fiddly is that both views are in diff locations, `spree/admin/order` path, the other is at `spree/order_mailer`, which means the latter has access to helper methods like pretty_time (that I figured out come from the `Spree::Admin::BaseHelper`) and the former does not, so I put the partial in `spree/admin/order` but now I see it isn't found from the mailer 😅 .

Options that spring to mind:
- Should I move this partial to `app/views/spree/admin/shared`?
- I'm pretty sure we don't want the Spree based pretty_time method in more places, is there an alternate time formatter/technique we prefer that doesn't depend on Spree?
- why is the order_mailer outside the admin dir anyway? Isn't it part of the orders admin form?
- leave the order_mailer alone and make this much clearer with a new invoice view

app/views/spree/admin/orders/_list.html.haml
luisramos0: :+1: nice, thanks for updating the docs :ok_hand: 
There was a dot missing in the example, lazy lookups need the initial dot:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)/_compare/61e41777a1596cd032bcc05fb1f5e7c1786f6f2f

app/views/spree/order_mailer/_payment.html.haml
mbudm: Update - I'm trying out the shared partial idea - seems promising

mbudm:
Updated screenshots - changed the column order as discussed in the issue #5208 - invoice PDF:

Screen Shot 2020-07-07 at 10 29 54 pm

Confirmation email:
Screen Shot 2020-07-07 at 9 59 35 pm

And as it is affected by the addition of a helper, checking that the payments page is unchanged:
Screen Shot 2020-07-07 at 10 01 40 pm

app/helpers/order_helper.rb
```diff
@@ -4,4 +4,8 @@ module OrderHelper
   def last_payment_method(order)
     OrderPaymentFinder.new(order).last_payment&.payment_method
   end
+
+  def outstanding_balance_label(order)
```
mbudm, line 8: @luisramos0 you mentioned extracting this logic to a service but it looked to me like it's more a helper type of thing...?

app/mailers/spree/order_mailer_decorator.rb
```diff
@@ -3,6 +3,7 @@
   helper CheckoutHelper
   helper SpreeCurrencyHelper
   helper OrderHelper
+  helper Spree::Admin::PaymentsHelper
```
mbudm, line 6: Not sure if there's an issue with bringing in yet more helpers here

mbudm:
Almost forgot invoice2:

app/assets/stylesheets/mail/payment_list.scss
```diff
@@ -0,0 +1,47 @@
...
+    th.payment-state {
+      text-align: right;
+    }
+  
+    td.payment-state {
+      text-align: right;
+      text-transform: uppercase;
+    }
+  }
+
```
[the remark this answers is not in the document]
luisramos0: ignore, this was fixed in later commits.

mbudm:
> oh, the build still has some relevant broken specs.
OK cool - I let them all run last night after doing some adhoc test runs locally, wasn't sure if the `/spec/mailers/subscription_mailer_spec.rb` was correlation or causation, i.e it was the right area but I wasn't clear how these chasnges had affected it - will investigate more deeply today.

mbudm:
One remaining failure on semaphore passes locally: spec/models/variant_override_spec.rb

filipefurtad0:
Hi @mbudm ,

I verified the issue described and used the steps described in #5208 to test the PR.

I checked 4 cases, in which invoices are generated:
 
- Bulk printing an invoice (pdf), on the Orders page (/admin/orders?q[s]=completed_at+desc)
- the functions under order edit page, (/admin/orders/R422624534/edit)
      - Print invoice (pdf)
      - Send the invoice (pdf per email)
      - Resend (email, no pdf)

Verified that the generated pdf documents above were identical in all cases.

1) By printing/sending an invoice of a complete/paid order: we can that the PR introduces information about the payment history:


- Invoice per email:

image



- PDF-invoice:

image

 
An issue not introduced by this PR: It's noticeable from the pics above, that shipping instructions appear on the email version version of the invoice, but are missing from the PDF file.


2) Removing an item from an order which is paid for (left side, before this PR, right side after this PR)

image

From the changes introduced by this PR, I would expected that "Credit Owed (£-12.00)" information would appear, somewhere in the area marked above. This is also not appearing in the email version.




3) Adding further items, and a second (insufficient) payment now appears as well:

image



This looks good for merging, thank you for these improvements!

Regarding the issues found:
- Missing shipping instructions on PDF-invoices (an issue will be opened)
- Missing Credit Owed (£-xx.xx) information on Email- and PDF-invoices: does this need a separate issue? Would it be papercut-size?

luisramos0:
Re the 2 issues:
>>> Missing shipping instructions on PDF-invoices (an issue will be opened)

Why do you think this is an issue? is it not a feature request? are the delivery instructions necessary on the invoice? why?

>> Missing Credit Owed (£-xx.xx) information on Email (order confirmation) and PDF-invoices: does this need a separate issue? Would it be papercut-size?

I think this is a bug in this PR. I think it could be the translation label :credit_owner missing from the en.yml file.
I think we could let @mbudm had a quick look at this second point and if it's easy fix it in this PR. What do you think @filipefurtad0 ?

filipefurtad0:
Hey @luisramos0,


Re the 2 issues:

            Missing shipping instructions on PDF-invoices (an issue will be opened)

Why do you think this is an issue? is it not a feature request? are the delivery instructions necessary on the invoice? why?

That's correct, that would be a feature request - not a bug. Invoice and order confirmation don't necessarily need the same info. Thanks!

        Missing Credit Owed (£-xx.xx) information on Email (order confirmation) and PDF-invoices: does this need a separate issue? Would it be papercut-size?

I think this is a bug in this PR. I think it could be the translation label :credit_owner missing from the en.yml file.
I think we could let @mbudm had a quick look at this second point and if it's easy fix it in this PR. What do you think @filipefurtad0 ?

Sounds good to me, especially knowing it's an easy fix. Moving back to In Dev :+1: 

Thanks again Luis!

filipefurtad0:
Hey @luisramos0,

 
> > Re the 2 issues:
> 
> ```
>         Missing shipping instructions on PDF-invoices (an issue will be opened)
> ```
> 
> > Why do you think this is an issue? is it not a feature request? are the delivery instructions necessary on the invoice? why?
> 

That's correct, that would be a feature request - not a bug. Invoice and order confirmation don't necessarily need the same info. Thanks!


> 
> ```
>     Missing Credit Owed (£-xx.xx) information on Email (order confirmation) and PDF-invoices: does this need a separate issue? Would it be papercut-size?
> ```
> 
> > I think this is a bug in this PR. I think it could be the translation label :credit_owner missing from the en.yml file.
> > I think we could let @mbudm had a quick look at this second point and if it's easy fix it in this PR. What do you think @filipefurtad0 ?
> 

Sounds good to me, especially if looks like an easy fix. Moving back to In Dev +1 
Thanks again Luis!

Matt-Yorkley:
Rebased, resolved conflicts, and the missing "credit owed" display should now be fixed.

spec/lib/spree/core/mail_interceptor_spec.rb
```diff
@@ -5,8 +5,8 @@
 # We'll use the OrderMailer as a quick and easy way to test. IF it works here
 # it works for all email (in theory.)
 describe Spree::OrderMailer do
-  let(:order) { Spree::Order.new(email: "customer@example.com") }
-  let(:message) { Spree::OrderMailer.confirm_email(order) }
```
luisramos0, line 9: this is not PR specific Matt, this is breaking in master. I am fixing it now in a separate PR :+1:

filipefurtad0:
Thanks @Matt-Yorkley 
the PR adds the "Credit Owed" info to the invoice:

image

Ready to go.

PR #5314, opened 2020-04-28
4599 restyle product listings

mkllnk:
#### What? Why?

Closes #4599

This pull request includes the following changes:

- No more red, grey, black colour palette.
- Photos now display on mobile
- Product title wraps (same as current version)
- Producer name truncates
- Weight/size wraps
- Price breakdown "pie" displays on mobile. *(Some of the pop-up is not visible on mobile, it's partly off-screen but that should be solved in #4600.)*
- Listing content layout change
- Removal of category icon
- Display product properties under description
- Display truncated product description, 'more' link to open product detail modal. *(The designs show a two-line description on small screens and a "more" button. I went for a simpler implementation that show the ellipsis after the first line. [Yuko gave her okay](https://github.com/openfoodfoundation/openfoodnetwork/issues/4599#issuecomment-617518941).)*
- Clicking on product name or image also opens product detail modal
- Removal of per-product sub-total display on mobile
- Categories design has less padding without affecting the categories design in the filter section.

Screenshot from 2020-04-28 16-55-47
Screenshot from 2020-04-28 17-11-34

#### What should we test?

Create a shopfront with:

* A short and a long producer name.
* Short and long product names.
* Short and long product descriptions.
* Some product properties.
* Products with and without bulk buy.

Then test the shopfront display in multiple browsers at multiple screen sizes. Resizing the window can be good to spot glitches between typical sizes.

Add products to you cart to make sure that it works as before.

#### Release notes

Updated product list design

Changelog Category: Changed

#### Dependencies

This will be merged into a feature branch. It will need two more issues before going live:

- #4598
- #4600

#### Documentation updates

Some screenshots may need replacing.

Issue #4671, opened 2020-01-13
[mobile ux] Remove icons from categories


daniellemoorhead:
## In a feature branch or master?
This can be merged directly to master 🎉 

## Description
**- As a:** instance manager
**- On page:** product categories
**- I want to be able to do:** have consistency with my categories and the way they display to users

## Acceptance Criteria & Tests
**Current behaviour**
Small icons display beside categories for _some_ instances
<img width="146" alt="Screen Shot 2020-01-13 at 12 47 52 pm" src="https://user-images.githubusercontent.com/10949052/72229466-f1a24700-3602-11ea-86a1-94431342e101.png">

**Changed behaviour**
Remove these icons so that there is consistency, no categories have them. 

Reason: as we add / remove / change categories we don't have icons to match them.

daniellemoorhead:
I'm pretty sure all these are gone now. Is that correct @mkllnk and @Matt-Yorkley, they disappeared from the product list view and the filter view?

mkllnk:
Yes, my pending pull request removes those items from the product listing.

PR #5713, opened 2020-07-01
Remove taxon icons

Matt-Yorkley:
#### What? Why?

These icons are inconsistently used across instances and have already started to be removed in a few places ( for example in the mobile UX redesign). They regularly trigger a long string of 404 responses in some servers, but it's not even consistent :see_no_evil: 

:fire: :fire:

#### What should we test?

Taxon icons are not present in the /shops index page, or on the /shop page. Admin (superadmin?) pages related to editing taxons are working correctly (view, update, etc).

#### Release notes

Removed taxon icons


sauloperez:
Agree but does the product team and the affected instance managers know about it?

luisramos0:
Can we have a screenshot of before and after to see what icons are being removed?
Yeah, this should have product input. We should ask for that but include an alternative with cost: how difficult it is to keep these icons and make them work consistently?

RachL:
@luisramos0 I believe they are only working for AU and UK currently. And maybe one day we will want to provide shops with entries by product category. So we won't use icons but images. 
So I wouldn't waste time on them.

sigmundpetersen:
Isn't this what is already done by @mkllnk through #4671 / #5314 ?

mkllnk:
My work removed the icon from the product listing (https://github.com/openfoodfoundation/openfoodnetwork/pull/5314/commits/e1422e132b379a01a6fdef9e8bd5d123989efaa9). But this pull request is removing the whole feature including the upload of icons etc.

I would like to know what the product team thinks about this, otherwise great initiative!

Matt-Yorkley:
Rebased. The build had some weird EnterpriseController errors... :+1:

sauloperez:
That's what broke the build and @luisramos0 fixed, I believe.

app/views/shops/_fat.html.haml
```diff
@@ -10,7 +10,6 @@
       .trans-sentence
         %div
           %span.fat-taxons{"ng-repeat" => "taxon in hub.taxons"}
-            %render-svg{path: "{{taxon.icon}}"}
```
luisramos0, line 13: this is removing the icons from the shops page, I think we need a screenshot with before/after so we can see the impact.

sauloperez: yes, seeing we don't mess up with margins and padding would be good

Matt-Yorkley: This PR vs Katuma prod:

**Filters in /shops**

Screenshot from 2020-07-14 12-05-12
Screenshot from 2020-07-14 12-05-50

**Filters in shop page**

Screenshot from 2020-07-14 12-07-23
Screenshot from 2020-07-14 12-07-01

No change, these icons are already missing on most servers.

luisramos0:
ok, these icons on the top right will also go, correct?
uk
image
AUS
image

sauloperez:
I think so @luisramos0 this is how it looks in Katuma

Screenshot from 2020-07-15 17-02-25

We could as well remove that `taxon-flag` altogether. It won't have any use.

cc @yukoosawa @daniellemoorhead

mkllnk:
The ribbon is already removed in the new product listing branch.

filipefurtad0:
Hi @Matt-Yorkley ,

This looks good,

- /shops page:


Before this PR 

image

After this PR

image


- /shop page:


before this PR

image

after this PR

image


Checked Admin/Superadmin pages related to editing taxons: all works as before.

Ready to go.

PR #1570, opened 2017-05-12
Add plurals

sseerrggii:
We need to add these plurals in order to have them in transifex and translate them in other languages

PR #5718, opened 2020-07-02
Available locales bug

### LINKED ISSUE #5669: Available locales login bug
## Description

Error 500 observed by @RachL when logging in to staging-FR - possibly with account: rachel.arnould+bamboo@protonmail.com

Bugsnag error:

https://app.bugsnag.com/open-food-france/open-food-france-prod/errors/5ef354f0dc9a3f001732340b?event_id=5ef354f0005b757586b80000&i=sk&m=nw

**Unhandled error**
I18n::InvalidLocale: "en_GB" is not a valid locale
Location
app/helpers/i18n_helper.rb:14 - set_locale

## Possible Fix

Not much more information, but perhaps related to #5648?

Matt-Yorkley:
#### What? Why?

Closes #5669 

Adds some defensive code to guard against cases where a user has a locale saved that is not in the available_locales list, possibly due to the available locales changing between user logins. This was causing fatal errors in some edge cases.

#### What should we test?

Login with an account that has selected a locale that is not in the available_locales list on the server resets the locale to the default, and does not cause a fatal error.

#### Release notes

Fixed error caused by available locales changing between logins

### DIFF
--- a/app/controllers/spree/user_sessions_controller.rb
+++ b/app/controllers/spree/user_sessions_controller.rb
@@ -6,7 +6,6 @@ module Spree
     include Spree::Core::ControllerHelpers::Common
     include Spree::Core::ControllerHelpers::Order
     include Spree::Core::ControllerHelpers::SSL
-    include I18nHelper
 
     ssl_required :new, :create, :destroy, :update
     ssl_allowed :login_bar
@@ -54,9 +53,8 @@ module Spree
     def ensure_valid_locale_persisted
       # When creating a new user session we have to wait until after a successful
       # login to be able to persist a locale on the current user
-      return unless spree_current_user && !available_locale?(spree_current_user.locale)
 
-      spree_current_user.update!(locale: I18n.default_locale)
+      UserLocaleSetter.ensure_valid_locale_persisted(spree_current_user)
     end
   end
 end
--- a/app/controllers/spree/user_sessions_controller.rb
+++ b/app/controllers/spree/user_sessions_controller.rb
@@ -12,7 +12,7 @@ module Spree
     ssl_allowed :login_bar
 
     before_action :set_checkout_redirect, only: :create
-    after_action :ensure_valid_locale, only: :create
+    after_action :ensure_valid_locale_persisted, only: :create
 
     def create
       authenticate_spree_user!
@@ -51,7 +51,9 @@ module Spree
       session["spree_user_return_to"] = nil
     end
 
-    def ensure_valid_locale
+    def ensure_valid_locale_persisted
+      # When creating a new user session we have to wait until after a successful
+      # login to be able to persist a locale on the current user
       return unless spree_current_user && !available_locale?(spree_current_user.locale)
 
       spree_current_user.update!(locale: I18n.default_locale)
--- a/app/controllers/spree/user_sessions_controller.rb
+++ b/app/controllers/spree/user_sessions_controller.rb
@@ -6,11 +6,13 @@ module Spree
     include Spree::Core::ControllerHelpers::Common
     include Spree::Core::ControllerHelpers::Order
     include Spree::Core::ControllerHelpers::SSL
+    include I18nHelper
 
     ssl_required :new, :create, :destroy, :update
     ssl_allowed :login_bar
 
     before_action :set_checkout_redirect, only: :create
+    after_action :ensure_valid_locale, only: :create
 
     def create
       authenticate_spree_user!
@@ -47,6 +49,12 @@ module Spree
     def redirect_back_or_default(default)
       redirect_to(session["spree_user_return_to"] || default)
       session["spree_user_return_to"] = nil
+    end
+
+    def ensure_valid_locale
+      return unless spree_current_user && !available_locale?(spree_current_user.locale)
+
+      spree_current_user.update!(locale: I18n.default_locale)
     end
   end
 end
--- a/app/controllers/spree/user_sessions_controller.rb
+++ b/app/controllers/spree/user_sessions_controller.rb
@@ -52,9 +52,10 @@ module Spree
 
     def ensure_valid_locale_persisted
       # When creating a new user session we have to wait until after a successful
-      # login to be able to persist a locale on the current user
+      # login to be able to persist a selected locale on the current user
 
-      UserLocaleSetter.ensure_valid_locale_persisted(spree_current_user)
+      UserLocaleSetter.new(spree_current_user, params[:locale], cookies).
+        ensure_valid_locale_persisted
     end
   end
 end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -1,6 +1,6 @@
 module I18nHelper
   def set_locale
-    UserLocaleSetter.new(spree_current_user, params[:locale], cookies).call
+    UserLocaleSetter.new(spree_current_user, params[:locale], cookies).set_locale
   end
 
   def valid_locale(user)
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -4,6 +4,6 @@ module I18nHelper
   end
 
   def valid_locale(user)
-    UserLocaleSetter.new(user).valid_locale_for_user
+    UserLocaleSetter.new(user).valid_current_locale
   end
 end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -24,11 +24,11 @@ module I18nHelper
     end
   end
 
-  private
-
   def available_locale?(locale)
     Rails.application.config.i18n.available_locales.include?(locale)
   end
+
+  private
 
   def current_user_locale
     spree_current_user.andand.locale
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -1,17 +1,17 @@
 module I18nHelper
   def set_locale
-    # Save a given locale
+    # Save a given locale from params
     if params[:locale] && available_locale?(params[:locale])
       spree_current_user&.update!(locale: params[:locale])
       cookies[:locale] = params[:locale]
     end
 
     # After logging in, check if the user chose a locale before
-    if spree_current_user && spree_current_user.locale.nil? && cookies[:locale]
-      spree_current_user.update!(locale: params[:locale])
+    if current_user_locale.nil? && cookies[:locale]
+      spree_current_user&.update!(locale: params[:locale])
     end
 
-    I18n.locale = spree_current_user.andand.locale || cookies[:locale] || I18n.default_locale
+    I18n.locale = current_user_locale || cookies[:locale] || I18n.default_locale
   end
 
   def valid_locale(user)
@@ -28,5 +28,9 @@ module I18nHelper
 
   def available_locale?(locale)
     Rails.application.config.i18n.available_locales.include?(locale)
+  end
+
+  def current_user_locale
+    spree_current_user.andand.locale
   end
 end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -1,10 +1,6 @@
 module I18nHelper
   def set_locale
-    # Save a given locale from params
-    if params[:locale] && available_locale?(params[:locale])
-      spree_current_user&.update!(locale: params[:locale])
-      cookies[:locale] = params[:locale]
-    end
+    save_locale_from_params
 
     # After logging in, check if the user chose a locale before
     if current_user_locale.nil? && cookies[:locale] && available_locale?(params[:locale])
@@ -29,6 +25,13 @@ module I18nHelper
   end
 
   private
+
+  def save_locale_from_params
+    return unless params[:locale] && available_locale?(params[:locale])
+
+    spree_current_user&.update!(locale: params[:locale])
+    cookies[:locale] = params[:locale]
+  end
 
   def current_user_locale
     spree_current_user.andand.locale
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -4,6 +4,6 @@ module I18nHelper
   end
 
   def valid_locale(user)
-    UserLocaleSetter.valid_locale_for_user(user)
+    UserLocaleSetter.new(user).valid_locale_for_user
   end
 end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -11,7 +11,7 @@ module I18nHelper
       spree_current_user&.update!(locale: params[:locale])
     end
 
-    I18n.locale = current_user_locale || cookies[:locale] || I18n.default_locale
+    I18n.locale = valid_current_locale
   end
 
   def valid_locale(user)
@@ -32,5 +32,15 @@ module I18nHelper
 
   def current_user_locale
     spree_current_user.andand.locale
+  end
+
+  def valid_current_locale
+    if available_locale?(current_user_locale)
+      current_user_locale
+    elsif available_locale?(cookies[:locale])
+      cookies[:locale]
+    else
+      I18n.default_locale
+    end
   end
 end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -1,49 +1,9 @@
 module I18nHelper
   def set_locale
-    save_locale_from_params
-
-    # After logging in, check if the user chose a locale before
-    if current_user_locale.nil? && cookies[:locale] && available_locale?(params[:locale])
-      spree_current_user&.update!(locale: params[:locale])
-    end
-
-    I18n.locale = valid_current_locale
+    UserLocaleSetter.new(spree_current_user, params[:locale], cookies).call
   end
 
   def valid_locale(user)
-    if user.present? &&
-       user.locale.present? &&
-       available_locale?(user.locale)
-      user.locale
-    else
-      I18n.default_locale
-    end
-  end
-
-  def available_locale?(locale)
-    Rails.application.config.i18n.available_locales.include?(locale)
-  end
-
-  private
-
-  def save_locale_from_params
-    return unless params[:locale] && available_locale?(params[:locale])
-
-    spree_current_user&.update!(locale: params[:locale])
-    cookies[:locale] = params[:locale]
-  end
-
-  def current_user_locale
-    spree_current_user.andand.locale
-  end
-
-  def valid_current_locale
-    if available_locale?(current_user_locale)
-      current_user_locale
-    elsif available_locale?(cookies[:locale])
-      cookies[:locale]
-    else
-      I18n.default_locale
-    end
+    UserLocaleSetter.valid_locale_for_user(user)
   end
 end
--- a/app/helpers/i18n_helper.rb
+++ b/app/helpers/i18n_helper.rb
@@ -7,7 +7,7 @@ module I18nHelper
     end
 
     # After logging in, check if the user chose a locale before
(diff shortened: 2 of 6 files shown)


Matt-Yorkley:
@sauloperez do you want a migration as well? I haven't written one yet...

sauloperez:
Yes, it's fine. I just answered in https://github.com/openfoodfoundation/openfoodnetwork/issues/5669#issuecomment-653013372

spec/features/consumer/authentication_spec.rb
```diff
@@ -135,5 +136,39 @@
...
+
+        it "logs in successfully and resets the user's locale to the default" do
+          fill_in_and_submit_login_form(user)
+          expect_logged_in
+
+          expect(page).to have_content I18n.t(:home_shop, locale: :en).upcase
+          expect(user.reload.locale).to eq "en"
+        end
+      end
+    end
```
sauloperez, line 172: I don't really see why we need this to be feature specs instead of controller. This logic only affects that controller and I'm pretty sure you can stub the cookie from controller specs. I'm worried about the pace at which increase the build's execution time

Matt-Yorkley: I tried it in a couple of different lower-down specs first, but it didn't replicate the error, everything was passing. I managed to replicate it in a feature spec though. I think this case involves some "automagic" things going on between Spree, Rails and Devise, but I'm not sure exactly.

Matt-Yorkley: In the stackstace, the only lines of code that were actually in our codebase was one method in `I18nHelper`.

app/controllers/spree/user_sessions_controller.rb
```diff
@@ -48,5 +50,11 @@ def redirect_back_or_default(default)
       redirect_to(session["spree_user_return_to"] || default)
       session["spree_user_return_to"] = nil
     end
+
+    def ensure_valid_locale
+      return unless spree_current_user && !available_locale?(spree_current_user.locale)
+
+      spree_current_user.update!(locale: I18n.default_locale)
```
luisramos0, line 57: I dont think we need something specific for this controller?
I think we should just do this in this controller like we do in all other controllers:
before_action :set_locale

app/helpers/i18n_helper.rb
```diff
@@ -1,17 +1,17 @@
...
+    # Save a given locale from params
     if params[:locale] && available_locale?(params[:locale])
       spree_current_user&.update!(locale: params[:locale])
       cookies[:locale] = params[:locale]
     end
 
     # After logging in, check if the user chose a locale before
-    if spree_current_user && spree_current_user.locale.nil? && cookies[:locale]
-      spree_current_user.update!(locale: params[:locale])
+    if current_user_locale.nil? && cookies[:locale] && available_locale?(params[:locale])
```
luisramos0, line 10: I dont understand why are you mixing cookies with params in this if clause?

luisramos0: did you mean available_locale?(cookies[:locale]) ?

luisramos0: and the line below would be
      spree_current_user&.update!(locale: cookies[:locale])
?

app/helpers/i18n_helper.rb
```diff
@@ -1,17 +1,17 @@
 module I18nHelper
   def set_locale
-    # Save a given locale
+    # Save a given locale from params
```
luisramos0, line 3: I think this should be an extracted method, can be called saveLocaleFromParams

app/helpers/i18n_helper.rb
luisramos0: I think this needs to be extracted to a method with a good name, if my corrections are correct this would be called saveLocaleFromCookies?

app/helpers/i18n_helper.rb
```diff
@@ -24,9 +24,23 @@ def valid_locale(user)
...
+
+  def current_user_locale
+    spree_current_user.andand.locale
+  end
+
+  def valid_current_locale
+    if available_locale?(current_user_locale)
+      current_user_locale
+    elsif available_locale?(cookies[:locale])
+      cookies[:locale]
```
luisramos0, line 41: if the cookies logic above is changed you dont need this because you have already saved cookies locale to the user.

app/controllers/spree/user_sessions_controller.rb
Matt-Yorkley: The thing with this controller is that `spree_current_user` only exists _after_ the controller action has finished, as it's handling login. Before and during the action there is no logged in user, and in some cases there is _still_ no logged in user after the action has finished (for example if the given the username or password were incorrect). So the parts that persist the locale on the user's account in `I18nHelper#set_locale` do not work in this specific action, in this specific controller...

app/helpers/i18n_helper.rb
Matt-Yorkley: Yes, this looked really strange to me as well, but I didn't want to alter the existing logic too much. The original lines are:
```ruby
# After logging in, check if the user chose a locale before
if spree_current_user && spree_current_user.locale.nil? && cookies[:locale]
  spree_current_user.update!(locale: params[:locale])
end
```

I've essentially refactored a tiny bit, and ensured we don't persist `params[:locale]` if it's not available, but the original logic is not changed. You're right though, it looks like it's either not doing anything useful, or not doing what it's intended to do...?

app/controllers/spree/user_sessions_controller.rb
luisramos0: ok, what about
after_action :set:_locale, only: :create
?

and then make set_locale handle it? I think we need to keep this logic in one place only.
If we really need to do something different here I think we need to extract set_locale to a service and call that service from here too, even if a different method than set_locale.

app/helpers/i18n_helper.rb
luisramos0: I think we need a bigger hammer on this PR. I think we must go for a bigger clean up here. This code is terrible and it's not so difficult to make it simple. Let me know, I can give it a try on Monday if you prefer.

app/controllers/spree/user_sessions_controller.rb
Matt-Yorkley: So we do need to _set_ the locale _before_ the page loads, but in this case we also have to wait until _after_ the action to be able to persist a valid locale...

app/helpers/i18n_helper.rb
Matt-Yorkley: > I think we need a bigger hammer

:smile: 

Yeah, I was going for the soft approach here and trying to leave things intact, but lets smash things instead :+1:

I'll take another look at this `cookies[:locale]` mess.

app/helpers/i18n_helper.rb
Matt-Yorkley: This method needs to return the most reasonable valid locale for the current page request, but the user may or may not be logged in. If they're not logged in, I think we we do need to look at the `cookies[:locale]` value, as a lower priority than the user's locale (if one is present), and higher priority than returning `I18n.default_locale`.

app/helpers/i18n_helper.rb
Matt-Yorkley: Yeah okay I think this cookies bit is a mess and I'm not sure the related specs are even doing anything useful either. :see_no_evil: 

In the existing code we start with this conditional: 
```ruby
spree_current_user && spree_current_user.locale.nil? && cookies[:locale]
```
This is checking a case where the user **is** logged in, there is **not** a locale set on the user record, but there **is** a locale set in cookies. I guess this scenario is: a user that as never explicitly selected a locale before has selected a locale (as a guest), then logged in. Or maybe a guest user selects a locale and then registers?

Then we do this if it's true:
```ruby
spree_current_user.update!(locale: params[:locale])
```
...persisting a locale to the user record, but from the params instead of the cookies (the value in the cookies is ignored). I'm not sure there's a case where the user is both in the process of logging in **and** the locale is being passed in the params at the same time? Anyway, we should probably be checking if `cookies[:locale]` is valid in this case, and persisting that value (not the params value) if the user doesn't have a locale already...

luisramos0:
we should probably remove this as part of this PR:
https://github.com/openfoodfoundation/openfoodnetwork/blob/d2db9b3a66ae09580363e4f500e31206f1e33953/config/application.rb#L136
what do you think?


also, when running specs locally, sometimes (random?) I get this error:
image

I hope/wonder if this is going away with this PR or if we need to do something else.

app/helpers/i18n_helper.rb
sauloperez: > we should probably be checking if cookies[:locale] is valid in this case, and persisting that value (not the params value) if the user doesn't have a locale already...

agree. I assume the cookie is needed for guest users and we cannot get rid of it, right?

sauloperez:
If we're going with the bigger hammer and I'm in favor of including `enforce_available_locales` as discussed in https://github.com/openfoodfoundation/openfoodnetwork/issues/5669#issuecomment-652330683, yes :+1:

Matt-Yorkley:
Can we handle this `enforce_available_locales` issue in another PR? I'd like to investigate it a bit before we start changing it, but I think it's separate to these changes. In some cases it's actually very useful to be able to "hack" a different locale, like checking default translations in Staging, for example.

It's not entirely clear why `I18n.config.enforce_available_locales = false` is not working, or working in some cases and not others. 

I think we have some other mess there that needs closer looking at... for example we have a funky alternate version of `#available_locales` in `lib/open_food_network/i18n_config.rb`...

app/services/user_locale_setter.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def self.valid_locale_for_user(user)
+    if user.andand.locale.present? && available_locale?(user.locale)
+      user.locale
+    else
+      I18n.default_locale
+    end
+  end
+
+  def self.available_locale?(locale)
```
luisramos0, line 30: why class method? why not just the instance and private method?

app/services/user_locale_setter.rb
```diff
@@ -0,0 +1,62 @@
...
+    I18n.locale = valid_current_locale
+  end
+
+  def ensure_valid_locale_persisted
+    return unless current_user && !available_locale?(current_user.locale)
+
+    current_user.update!(locale: valid_current_locale)
+  end
+
+  def self.valid_locale_for_user(user)
```
luisramos0, line 22: maybe valid_locale_for(user) ?

app/services/user_locale_setter.rb
```diff
@@ -0,0 +1,62 @@
...
+    if user.andand.locale.present? && available_locale?(user.locale)
```
luisramos0, line 23: user&.locale.present?  ?

app/services/user_locale_setter.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    current_user&.update!(locale: params_locale)
+    cookies[:locale] = params_locale
+  end
+
+  def available_locale?(locale)
+    self.class.available_locale?(locale)
+  end
+
+  def current_user_locale
```
luisramos0, line 49: this is only used in valid_current_locale, in my book it should be after valid_current_locale. I wonder if you follow the opposite positioning rule, it's possible and equally valid :-)

app/services/user_locale_setter.rb
luisramos0: there's repetition between this method and valid_current_locale.
The problem is that this is called from the mailed that can be used in delayedjob where there are no cookies.
I wonder if we could just add a if cookies on line 56 below and just reuse the methods.
That would mean remove this method and just make valid_current_locale public.

app/controllers/spree/user_sessions_controller.rb
luisramos0: :+1:

app/helpers/i18n_helper.rb
luisramos0: yeah, that code looked like non-sense... but I am not sure I understand everything that is going on here.
Anyway, I think this PR is a great step forward and we can iterate from here if we find any issues.

app/helpers/i18n_helper.rb
luisramos0: ok, I think this makes sense.

app/services/user_locale_setter.rb
Matt-Yorkley: Yes, I pulled it into the service but it's used in contexts where calling cookies will throw an error. I'll tidy this up a bit more :+1:

app/services/user_locale_setter.rb
Matt-Yorkley: I had to make this available as an instance method because #valid_locale_for_user was an instance method. Updated :+1:

Matt-Yorkley:
I made the class methods instance methods, tidied up the public interface and removed some duplication. It looks pretty nice now :smile:

spec/features/consumer/authentication_spec.rb
Matt-Yorkley: I've added more unit tests here as well, but a couple of small integration tests that tell us whether or not users can actually log in to the app without hitting a fatal error seem like a good idea. 

Valen la pena, no? :sweat_smile:

luisramos0: I wonder what kind of tests in this area we already have. We have an authentication_spec. There's also a login in registration_spec. I wonder if we could add some locale variant in one of these login specs.

Matt-Yorkley: Already added in authentication_spec :+1:

app/services/user_locale_setter.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  private
+
+  attr_reader :current_user, :params_locale, :cookies
+
+  def save_locale_from_params
+    return unless params_locale && available_locale?(params_locale)
+
+    current_user&.update!(locale: params_locale)
```
sauloperez, line 39: why do we need to defend from an empty current_user? it's mandatory in here but if the controller has it as nil we have a much bigger problem that needs solving and not hiding IMO

app/services/user_locale_setter.rb
```diff
@@ -0,0 +1,50 @@
...
+    cookies[:locale] = params_locale
+  end
+
+  def available_locale?(locale)
+    Rails.application.config.i18n.available_locales.include?(locale)
+  end
+
+  def current_user_locale
+    current_user&.locale
```
sauloperez, line 48: same here

spec/features/consumer/authentication_spec.rb
sauloperez: Fair enough. I was convinced we login through the UI in several feature specs but it doesn't look like it now.

filipefurtad0:
Hey @Matt-Yorkley ,

First off, I aimed at reproducing the bug:
- I logged in
- Set Italian as the language preference
- Logged out.
- Made sure that the language preferences are kept, by logging in into another browser.

Via ssh, I changed application.yml on staging-FR to remove `it` from AVAILABLE_LOCALES:

from
```
cat apps/openfoodnetwork/current/config/application.yml | grep LOCALES
AVAILABLE_LOCALES: fr,it,en_FR

```

to
```
openfoodnetwork@vps189799:~$ cat apps/openfoodnetwork/current/config/application.yml | grep LOCALES
AVAILABLE_LOCALES: fr,en_FR
```
Italian language option is gone in the app:
image

Having this changed, and **before** staging this PR:

- Logging in to staging-FR with that particular account and previously set Italian language preference breaks the page - Snail - Error 500. Bugsnag reports (I believe this is the triggered error):

ActionView::MissingTemplate · Missing template spree/home/index, spree/store/index, spree/base/index, application/index with {:locale=>[:fr, :en], :formats=>[:html], :handlers=>[:erb, :builder, :raw, :ruby, :coffee, :haml]}. Searched in: * "/home/openfoodnetwork/apps/openfoodnetwork/releases-old/2020-07-03-185325/app/views" * "/home/openfoodnetwork/.gem/ruby/2.3.0/gems/foundation-rails-5.5.2.1/app/views" * "/home/openfoodnetwork/.gem/ruby/2.3.0/gems/devise-3.0.4/app/views" * "/home/openfoodnetwork/.gem/ruby/2.3.0/bundler/gems/better_spree_paypal_express-e28e4a8c5ced/app/views" * "/home/openfoodnetwork/.gem/ruby/2.3.0/bundler/gems/spree-0b0c422369c8/core/app/views" * "/home/openfoodnetwork/.gem/ruby/2.3.0/gems/kaminari-0.14.1/app/views" * "/home/openfoodnetwork/apps/openfoodnetwork/releases-old/2020-07-03-185325/engines/web/app/views" * "/home/openfoodnetwork/apps/openfoodnetwork/releases-old/2020-07-03-185325/engines/order_management/app/views" 

Which seems to differ from the error previously reported:
https://github.com/openfoodfoundation/openfoodnetwork/issues/5669#issue-645420131

So I'm wondering whether this is was the right way to reproduce the bug.

Anyway, **after** staging this PR:

- Logging to with that particular account and previously set Italian language preference does not breaks the page, and sets the page to the default locale (FR).

Awesome! :tada: 
Good to go! 


Only thing I'm wondering is why removing the previously set locale did not trigger the same error, as reported in the original issue. Any thoughts on this?

Matt-Yorkley:
Amazing testing! Thanks :heart_eyes:

luisramos0:
yeah, it's awesome you are learning to update things on the server Filipe!
One of the things you can do next time is to restart the server after making those changes. I think that could give a different output.
To restart the server you need to login as ofn-admin:
ssh ofn-admin@staging.openfoodfrance.org
and then run:
sudo systemctl restart unicorn_openfoodnetwork.service


I am not sure what are the conditions that cause the error you saw or if it's even related to locales. BUT I think I have fixed that error in this PR #5677

filipefurtad0:
Thank you Luis!

I repeated the test, with build #1570 staged:
- set a local variant as Italian, for a user; logged-out.
- edited application.yml file, removing the `it`
- restarted the server with the commands provided
- logged in, on a private window.
- observed no error...

I repeated this with the current build (which is previous to merging of this PR). But couldn't reproduce the error.

luisramos0:
thanks Filipe for repeating the test.
I am not sure why you didnt see the error. Maybe we can give it a try when we test the release?

RachL:
@filipefurtad0 @luisramos0 I don't have ssh access. Can one of you have a look? The release is ready to go, but still staged in FR.

filipefurtad0:
I'm having a look right now, I'm trying to reproduce it in staging UK.

filipefurtad0:
Good news,
I repeated this test in staging-UK and was able to reproduce the error. Setting `fr` as the preferred language for a given user, deleting it from AVAILABLE_LOCALES, restarting the server and attempting to log back in breaks the app:

image

This corresponds to the error triggered in Bugnsnag:

**I18n::InvalidLocale home#index
"fr" is not a valid locale**

https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5ed679ba3484f00017b6676c?filters[event.since][0]=30d&filters[error.status][0]=open

Which coincides with the previously reported one! :tada: 
#5669 (comment)

However, I don't know why:
a) it did not show up in devops-notifications (staging-UK)
b) it did not get triggered in bugsnag before, when testing this in staging-FR - or maybe it did, and was mixed up with so many repetitions from other errors?

Anyway, this IS the issue this PR aims to solve and it does: staging the release draft in staging-UK brings back the app from it's broken state, and sets the default language for the user.


Just to be sure, did the same on staging-FR (with the draft for the release staged):
- Set `it` as the preferred language for a given user
- deleted it from AVAILABLE_LOCALES
- restared the server
- logged back in
- ...it didn't break, looks solid! :muscle: 

Worth double checking, all good.

luisramos0:
Awesome!!!

PR #5729, opened 2020-07-06
Remove Image settings page and standardise Paperclip styles

### LINKED ISSUE #5695: Remove unchangable config settings from Configuration
## Description
Image settings are not something that can be changed without the help of a dev: running commands on the server, moving images around, etc.
I think we can probably delete completely this page:
image

The only part I have doubts are the paperclips styles, can this be changed? Will they have an impact anywhere on the app? Are these configs useful?

## Expected Behavior
No image settings in Configuration.

## Actual Behaviour
Image settings can be changed and images in the app will break.

## Steps to Reproduce
There are few cases:
- change attachments URL and see that all existing images are now gone.
- change use s3 for images and see that all images are now gone.

## Severity
I think this is the cause of S1 #5668

## Your Environment
Seen in BE prod.

luisramos0:
#### What? Why?

Closes #5695

This PR is paired with changes in ofn-install to set and change these setting from now on: https://github.com/openfoodfoundation/ofn-install/pull/645

Apart from removing the pages we are adding some code to spree initializers to pick up ofn-instrall config if they exist.

#### What should we test?
After this PR and after image styles were removed from ofn-install as part of https://github.com/openfoodfoundation/ofn-install/pull/645 there is no way to update styles though the UI or through ofn-install.
We end up with only two ways to update styles:
- changing the default in the code like in this PR: https://github.com/openfoodfoundation/openfoodnetwork/pull/6223 
- use the new tools: https://github.com/openfoodfoundation/openfoodnetwork/pull/6014

This PR becomes a manual test only:
- image settings are gone from configuration
- images are rendering correctly as before this PR

#### Release notes
Changelog Category: Removed
Removed image settings page from the app, image settings are now done in ofn-isntall.


knapsack_rspec_report.json
```diff
@@ -331,7 +329,6 @@
   "spec/features/admin/external_services_spec.rb": 0.3893253803253174,
   "spec/models/stock/package_spec.rb": 0.43047094345092773,
   "spec/controllers/admin/stripe_connect_settings_controller_spec.rb": 0.27138352394104004,
-  "spec/controllers/spree/admin/image_settings_controller_spec.rb": 0.2584362030029297,
```
sauloperez, line 334: I think this will unbalance our jobs. We might need a full re-balance and update this file.

config/initializers/spree.rb
```diff
@@ -38,7 +57,39 @@
...
+    Spree::Image.attachment_definitions[:attachment].delete :storage
+  end
+
+  Spree::Image.attachment_definitions[:attachment][:styles] =
+    ActiveSupport::JSON.decode(config.attachment_styles).symbolize_keys!
+  Spree::Image.attachment_definitions[:attachment][:path] = config.attachment_path
+  Spree::Image.attachment_definitions[:attachment][:default_url] =
+    config.attachment_default_url
+  Spree::Image.attachment_definitions[:attachment][:default_style] =
+    config.attachment_default_style
```
sauloperez, line 90: What about bringing in the image model from spree and define the attachments as usual for Paperclip? The first place I'd go find the settings is in the model because that's the convention. Also, I find tedious having to deal with hashes here instead of the `has_attached_file` DSL, and is not easy to follow either IMO.

knapsack_rspec_report.json
luisramos0: yes, but on another PR. I havent managed to get a new knapsack report the last time I tried, the single job build expired after 3 hours :man_shrugging:

config/initializers/spree.rb
luisramos0: I'd have to investigate how do it, change the code and test it.
This way I am keeping the code as is (this is exactly what was on the controller) and just moving it around. I see changing code a bit out of scope for this PR. It would be more effort and risk. I'd prefer to go forward as is so that we save dev time. What do you think?

knapsack_rspec_report.json
sauloperez: sure, I didn't mean to do it here.

sigmundpetersen:
Some conflicts that should be resolved @luisramos0

mkllnk (approved):
I reviewed this pull request a while ago but somehow it wasn't submitted. My draft was still sitting here. I commented on one thing that I think we should change but now I think that we should get this through and follow up with another pull request.

Since we set the same styles for all instances, I think we should hardcode them in the Spree::Image model. There is no use case for having different styles in different countries. We all use the same design and if someone finds an optimisation, we should share it with everybody.

There's another issue I came across: When we deploy this, the new styles will come into effect. But until the whole thumbnail generation is complete, there will be broken images. That's not due to different sizes but because of different file names (jpg instead of jpeg and png). The only way around that is to run the thumbnail generation with new styles before deploying this.

The rollout of this is driving me a bit crazy. It's not that hard when you look at it calmly but I want to have it all done at the same time and there are quite a few steps. ;-)

mkllnk:
### Testing notes


-    verify we can change image settings through ofn-install (testing PR 64) :heavy_check_mark: 
-    verify those changes are correctly picked up in the application :red_circle: 

The effective style was not changed. All styles should have `jpg` as format:

```diff
- "{\"mini\":[\"48x48#\",\"jpg\"],\"small\":[\"227x227#\",\"jpg\"],\"product\":[\"240x240\\u003E\",\"\"],\"large\":[\"600x600\\u003E\",\"jpg\"]}"
+ "{\"mini\":[\"48x48#\",\"jpg\"],\"small\":[\"227x227#\",\"jpg\"],\"product\":[\"240x240>\",\"jpg\"],\"large\":[\"600x600>\",\"jpg\"]}"
```

But there is no change.

Afterwards we should also do a manual verification that:

-    image settings are gone from configuration :heavy_check_mark: 
-    images are rendering correctly as before this PR :red_circle: 

The attachment_path was changed which breaks all images:
```diff
- public/images/spree/products/:id/:style/:basename.:extension
+ public/spree/products/:id/:style/:basename.:extension
```

sigmundpetersen:
Hey @luisramos0 are the build failures unrelated to the PR?
They seem consistent.

mkllnk (approved):
Yay, we are getting there! :tada: 

Really good work, finally ditching Spree::Config which has annoyingly cached values.

My last dev test had two fails. I think that the first one was caused by the cache which should be resolved by the new commits here. The second fail was caused by a missing attachment path config for au-staging. I opened https://github.com/openfoodfoundation/ofn-install/pull/658 to fix that. The dev test of this PR is blocked by that config.

app/models/spree/image.rb
```diff
@@ -5,9 +5,14 @@ class Image < Asset
...
+    # This is where the styles are used in the app:
+    # - mini: used in the BackOffice: Bulk Product edit page and Order Cycle edit page
+    # - small: used in the FrontOffice: product list page
+    # - product: used in the BackOffice: Product Image upload modal in the Bulk Product edit page and Product image edit page
+    # - large: used in the FrontOffice: product modal
     has_attached_file :attachment,
-                      styles: { mini: '48x48>', small: '100x100>',
-                                product: '240x240>', large: '600x600>' },
+                      styles: { mini: ["48x48#", :jpg], small: ["227x227#", :jpg],
+                                product: ["240x240>", :jpg], large: ["600x600>", :jpg] },
```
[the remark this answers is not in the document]
luisramos0: Don't we need to deploy this immediately after we convert the styles with the task?

app/models/spree/image.rb
```diff
@@ -51,26 +47,23 @@ def no_attachment_errors
       false
     end
 
-    # Spree stores attachent definitions in JSON. This converts the style name and format to
-    # strings. However, when paperclip encounters these, it doesn't recognise the format.
-    # Here we solve that problem by converting format and style name to symbols.
-    # See also: ImageSettingsController decorator.
-    #
-    # eg. {'mini' => ['48x48>', 'png']} is converted to {mini: ['48x48>', :png]}
-    def self.format_styles(styles)
```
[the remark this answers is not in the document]
luisramos0: :+1:

app/models/spree/image.rb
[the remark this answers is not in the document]
mkllnk: Yes, so #6014 needs to be deployed already. It could avoid delays if we run the scripts in #6014 early then run it again with the new images just before we want to deploy this here.

app/models/spree/image.rb
```diff
@@ -51,26 +47,23 @@ def no_attachment_errors
...
+                     secret_access_key: Spree::Config[:s3_secret],
+                     bucket: Spree::Config[:s3_bucket] }
+        Spree::Image.attachment_definitions[:attachment][:storage] = :s3
+        Spree::Image.attachment_definitions[:attachment][:s3_credentials] = s3_creds
+        Spree::Image.attachment_definitions[:attachment][:s3_headers] =
+          ActiveSupport::JSON.decode(Spree::Config[:s3_headers])
+        Spree::Image.attachment_definitions[:attachment][:bucket] = Spree::Config[:s3_bucket]
+      else
+        Spree::Image.attachment_definitions[:attachment].delete :storage
+      end
```
sauloperez, line 66: We should replace all these `Spree::Image` with `self`

sauloperez (changes requested):
yay, so much :fire: !

app/models/spree/image.rb
```diff
@@ -51,26 +47,23 @@ def no_attachment_errors
...
+    def self.set_attachment_attributes(attribute_name, attribute_value)
+      Spree::Image.attachment_definitions[:attachment][attribute_name] = attribute_value
     end
 
-    def self.reformat_styles
-      Spree::Image.attachment_definitions[:attachment][:styles] =
-        format_styles(Spree::Image.attachment_definitions[:attachment][:styles])
+    def self.set_s3_attachment_definitions
+      if Spree::Config[:use_s3]
+        s3_creds = { access_key_id: Spree::Config[:s3_access_key],
```
sauloperez, line 56: we can extract that into a method to make this a bit more readable. Code Climate complaints about it too

app/models/spree/image.rb
luisramos0: nice :+1: actually it can be removed because then self becomes redundant (rubocop helps remembering that).

app/models/spree/image.rb
luisramos0: :+1:

sauloperez (approved):

luisramos0:
The restart checkout spec is failing in semaphore but passing locally. I dont see how this PR could be related to it...

sigmundpetersen:
Interestingly the same specs fail in another PR as well https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-6110/builds/1 ping @filipefurtad0 

So I guess that also indicates it's not specific to this PR. I haven't seen it fail in master though. Let's see if it will start failing after merging this 🙈

luisramos0:
good point @sigmundpetersen thanks
I think we can go ahead then and have a look at it if it starts failing in master.

filipefurtad0:
Thanks for pinging @sigmundpetersen :+1: 
I rebuilt the other PR6110 and it failed again. Also, it looks like `/spec/services/order_checkout_restart_spec.rb` it's failing in master [now](https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/master/builds/1751)...

Matt-Yorkley:
@luisramos0 there's two big changes here and I'm wondering if we can decouple them?

- removing the admin page for image settings
- changing the image sizes as part of the product images changes (which requires the image resizing to be done before this could be deployed?)

luisramos0:
Hey @Matt-Yorkley the change in image.rb has_attached_file :attachment will not have any effect unless we run the commands to resize the images. Are you suggesting I create a separate PR just for that line in image.rb has_attached_file :attachment?

luisramos0:
ok @Matt-Yorkley I have rebased this PR and added a commit to remove the change in the default styles.

Matt-Yorkley (approved):

sigmundpetersen:
Hey @luisramos0 is this still blocked by #6014 or is it rather #6223 that is blocked?
(Or nothing blocked anymore? 🙂 )

sauloperez:
As far as I understand nothing is blocked anymore.

luisramos0:
yes, this is not blocked by anything now :+1: that was the point of the change.
But #6223 is very closely related to whatever we do in #6014

luisramos0:
So this one is ready for manual testing now :+1:

mkllnk:
I think that we are going in circles here. The breaking change in this PR is that the image settings are not read from the database any more and instead defined in the code. The current settings in the production databases are not standardised and any standardised setting would break some servers. The Australian server converts all product images to JPG while other servers keep them in their original format which is often PNG and sometimes JPEG. After deploying the current version of this pull request (using original image formats), Australia would suddenly lose all PNG and JPEG images. Because of the style declarations, the code would look for a PNG image while we only stored the JPG image.

So this still depends on the tools to convert the images first and the last commit doesn't bring any improvement.

Matt-Yorkley:
> Australia would suddenly lose all PNG and JPEG images.

I don't think it will. The image format forces uploaded images to be created in that format. If you don't specify a format it will still find whatever images have been saved previously.

Matt-Yorkley:
If we use the jpg format eg: `["227x227#","jpg"]` it will break images on all servers except Aus (and Germany).

If we leave out the jpg format eg: `"227x227#"` it will not break any images on any servers.

mkllnk:
I still disagree with that one:
> If we leave out the jpg format eg: small: "227x227#" it will not break any images on any servers.

It's difficult to focus the conversation on all the different pull requests. Maybe let's continue on Slack?

Matt-Yorkley:
I've done a `dev-test` here. I changed the image settings in the UI to something other than the ones in this PR before staging.

- Is the image setting page removed? :heavy_check_mark: 
- Are existing product images still shown? :heavy_check_mark: 
- Are the hard-coded image settings in `Spree::Image` now the single source of truth for image configs, and the old database-persisted configs are no longer used? :heavy_check_mark: 
- Does image upload work for products, and apply the expected formatting? :heavy_check_mark:

luisramos0:
thanks Matt!

PR #5733, opened 2020-07-06
Move lib/spree to OFN 

### LINKED ISSUE #5734: [ByeBye Spree] Move lib/spree to OFN
Move all needed files from core/lib to OFN and merge with decorators.

These files (mostly handled in #5733):
- [x] core/lib/spree/core.rb COPY
- [x] core/lib/spree/i18n.rb COPY
- [x] core/lib/spree/i18n/base.rb COPY
- [x] core/lib/spree/i18n/initializer.rb COPY
- [x] core/spec/lib/i18n_spec.rb COPY
- [x] core/lib/spree/money.rb COPY
- [x] core/spec/lib/spree/money_spec.rb COPY
- [x] core/lib/spree/responder.rb MAYBE COPY

- [x] - core/lib/spree/core/environment.rb COPY
- [x] - core/lib/spree/core/environment_extension.rb COPY
- [x] - core/lib/spree/core/environment/calculators.rb COPY

- [x] - core/lib/spree/core/ext/active_record.rb COPY (this one is handled in #5758)

- [x] - core/lib/spree/core/calculated_adjustments.rb COPY
- [x] - core/spec/lib/calculated_adjustments_spec.rb COPY and move to core/spec/lib/spree
- [x] - core/lib/spree/core/delegate_belongs_to.rb COPY
- [x] - core/lib/spree/core/gateway_error.rb COPY
- [x] - core/lib/spree/core/mail_interceptor.rb MAYBE COPY
- [x] - core/spec/lib/mail_interceptor_spec.rb COPY and move to core/spec/lib/spree
- [x] - core/lib/spree/core/mail_settings.rb COPY
- [x] - core/spec/lib/mail_settings_spec.rb COPY and move to core/spec/lib/spree
- [x] - core/lib/spree/core/permalinks.rb COPY
- [x] - core/lib/spree/core/product_duplicator.rb COPY
- [x] - core/spec/models/spree/product_duplicator_spec.rb COPY
- [X] - core/lib/spree/core/product_filters.rb NO COPY - I dont think this is used in OFN
- [x] - core/spec/models/spree/product_filter_spec.rb NO COPY, see previous
- [x] - core/lib/spree/core/s3_suppport.rb COPY
- [x] - core/lib/spree/core/token_resource.rb COPY
- [x] - core/spec/lib/token_resource_spec.rb COPY and move to core/spec/lib/spree
- [X] - core/lib/spree/core/version.rb NO COPY - I dont think this will be needed when we drop spree_core 

### Context
Part of #4826

### Impact and timeline

luisramos0:
#### What? Why?

Closes #5734
We bring most of what is needed from spree_core/lib/spree
After this PR, we are basically missing app/models (big one) and some of the test_support stuff :tada:

#### What should we test?
Quite a few core features impacted with these files.
- Verify basic order fees and shipping fees are working
- Create product and edit details
- Calculators - are all types of calculators available in the app? I dont think we need to test them individually. Just verify you can select all of them in a shipping method and in an enterprise fee for example.
- Mail Interceptor - in Mail method config make sure you can intercept emails (last option in the page)
- Mail Settings - make sure the mail settings are preserved as they were before the PR was staged. For example, if deliveries are activated and the domain emails are sent from.
- Permalinks - make sure enterprise/shop permalinks are correctly generated and existing ones are correctly parsed (for example, access an existing shop and check the permalink is correct)
- S3 support - verify S3 connection is working and images are rendered correctly according to settings
- I18n - make sure you can change locale as usual
- Product duplicator - verify product cloning is working well

#### Release notes
Changelog Category: Added
Brought the last batch of core lib classes (we are only missing spree models now) from spree_core so we can become independent of Spree.

### DIFF
--- a/lib/spree/core.rb
+++ b/lib/spree/core.rb
@@ -1,0 +1,56 @@
+require 'rails/all'
+require 'active_merchant'
+require 'acts_as_list'
+require 'awesome_nested_set'
+require 'cancan'
+require 'kaminari'
+require 'mail'
+require 'paperclip'
+require 'paranoia'
+require 'ransack'
+require 'state_machine'
+
+module Spree
+
+  mattr_accessor :user_class
+
+  def self.user_class
+    if @@user_class.is_a?(Class)
+      raise "Spree.user_class MUST be a String object, not a Class object."
+    elsif @@user_class.is_a?(String)
+      @@user_class.constantize
+    end
+  end
+
+  # Used to configure Spree.
+  #
+  # Example:
+  #
+  #   Spree.config do |config|
+  #     config.site_name = "An awesome Spree site"
+  #   end
+  #
+  # This method is defined within the core gem on purpose.
+  # Some people may only wish to use the Core part of Spree.
+  def self.config(&block)
+    yield(Spree::Config)
+  end
+end
+
+require 'spree/core/version'
+require 'spree/core/engine'
+
+require 'spree/i18n'
+require 'spree/money'
+require 'spree/promo/coupon_applicator'
+
+require 'spree/core/delegate_belongs_to'
+require 'spree/core/ext/active_record'
+require 'spree/core/permalinks'
+require 'spree/core/token_resource'
+require 'spree/core/calculated_adjustments'
+require 'spree/core/product_duplicator'
+
+ActiveRecord::Base.class_eval do
+  include CollectiveIdea::Acts::NestedSet
+end
--- a/lib/spree/core.rb
+++ b/lib/spree/core.rb
@@ -35,7 +35,7 @@ module Spree
   #
   # This method is defined within the core gem on purpose.
   # Some people may only wish to use the Core part of Spree.
-  def self.config(&block)
+  def self.config
     yield(Spree::Config)
   end
 end
--- a/lib/spree/core.rb
+++ b/lib/spree/core.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'rails/all'
 require 'active_merchant'
 require 'acts_as_list'
@@ -11,15 +13,16 @@ require 'ransack'
 require 'state_machine'
 
 module Spree
-
   mattr_accessor :user_class
 
   def self.user_class
     if @@user_class.is_a?(Class)
       raise "Spree.user_class MUST be a String object, not a Class object."
-    elsif @@user_class.is_a?(String)
-      @@user_class.constantize
     end
+
+    return unless @@user_class.is_a?(String)
+
+    @@user_class.constantize
   end
 
   # Used to configure Spree.
--- a/lib/spree/core.rb
+++ b/lib/spree/core.rb
@@ -45,7 +45,6 @@ require 'spree/core/engine'
 
 require 'spree/i18n'
 require 'spree/money'
-require 'spree/promo/coupon_applicator'
 
 require 'spree/core/delegate_belongs_to'
 require 'spree/core/ext/active_record'
--- a/app/controllers/spree/base_controller.rb
+++ b/app/controllers/spree/base_controller.rb
@@ -1,0 +1,16 @@
+require 'cancan'
+require_dependency 'spree/core/controller_helpers/strong_parameters'
+
+class Spree::BaseController < ApplicationController
+  include Spree::Core::ControllerHelpers::Auth
+  include Spree::Core::ControllerHelpers::RespondWith
+  include Spree::Core::ControllerHelpers::SSL
+  include Spree::Core::ControllerHelpers::Common
+  include Spree::Core::ControllerHelpers::Search
+  include Spree::Core::ControllerHelpers::StrongParameters
+  include Spree::Core::ControllerHelpers::Search
+
+  respond_to :html
+end
+
+require 'spree/i18n/initializer'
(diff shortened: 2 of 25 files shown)


sauloperez:
awesome :tada: ! there are just a few code climate issues that I think are worth fixing. Some related to `include` and a couple more that show at the top of the list.

luisramos0:
I verified each warning and I left the ones I feel are risky to change and/or I am not sure how to test.

Maybe you can help me:
image

Can I really change to from &block to _block or should it be &_block or it doesn't matter?

This one:
image

Is this really true? These private methods will not be private in the class this module is being added to? I am not too sure.


This one:
image
I am not changing this! Respect for the magic!
You know, class loading order or whatever other magical stuff could be going on...

sauloperez:
I can take a look

sauloperez:
This should be it. Now that I've checked the code :see_no_evil:. I hate including modules into Rails core classes. It's fancy to use but it's not worth it IMO.

luisramos0:
Yes, it's not nice, it's used in the product model as well as in the variant model and the gateway model. We can remove it later

lib/spree/core.rb
```diff
@@ -35,7 +35,7 @@ def self.user_class
   #
   # This method is defined within the core gem on purpose.
   # Some people may only wish to use the Core part of Spree.
-  def self.config(&block)
+  def self.config
```
luisramos0, line 38: this will blow up if called with parameters, right?

sauloperez: no, it won't because of the way blocks in methods work. It will if no block is passed. See below :point_down: 

```ruby
irb(main):001:0> def foo
irb(main):002:1> yield 'hola'
irb(main):003:1> end
=> :foo
irb(main):004:0> foo
LocalJumpError: no block given (yield)
        from (irb):2:in `foo'
        from (irb):4
        from /home/pau/.rbenv/versions/2.3.7/bin/irb:11:in `<main>'
irb(main):005:0> foo { |msg| puts msg }
hola
=> nil
```

luisramos0: Awesome, I'll make this change :+1: 
Thanks!!!

luisramos0: awesome, thanks!!!

luisramos0:
can you still approve the PR after adding a commit?

sauloperez:
Unless someone disagrees I'm making the call to this to Test Ready. Given the time it's been sitting there and the nature of the fixes we're doing recently, it's fairly easy that it'll end up with merge conflicts what will bring up the costs of the tech debt.

sauloperez:
YES! the future is indeed bright @mkllnk :rocket: !

filipefurtad0:
Hey @luisramos0 !

Indeed a lot going on here, let's go step-wise:

- Verified basic order fees, shipping fees, enterprise are working
- Created products and edited details
- Checked all types of calculators are available in the app, under enterprise/shipment/transaction fees. They are available and work correctly - did a smoke test, on several of them.
- Mail Interceptor works correctly: when creating orders, the intercept recipient gets the email, and not the customer.
- Mail Settings such deliveries being set to activated and the domain emails are sent from remain as 
- Permalinks work correctly for previous ones and are created in the correct way for new enterprises. If a permalink already exists, then numbers are added in the end.
- tested I18n by changing locale to different languages, all good.
- Product duplicator works as before. Products get cloned but not their varieties

So far so good.

However, I'm not too sure on how to test S3 support. The configurations appear the same as before staging the PR. But some images were broken on staging-FR (since yesterday, after dev-test - remember Luís?) and I broke it a bit further, by accidentally removing the access credentials to S3. So, I would need help to fix this and the configuration issue which came up before, on staging-fr.
I staged this or in staging-UK, in which S3 was not activated, although there were some credentials saved. Activating S3 breaks the images, which may related to the need to point out an image bucket, I'm not sure. How can we best check this?

Possible issues found (not related to this PR):

- While "Enable Mail Delivery" and "Send Mails As" configurations are retained (as mentioned above), additional email settings like "Send Copy of All Mails To" and "Intercept Email Address" are erased every time a new PR is deployed. I'm guessing this is intended/expected behavior, but maybe worth to mention

- I could not get "Send Copy of All Mails To" option to work properly. I did not find any issue currently dealing with this, so I think it's best to open one.

I would say the PR looks good for merging, but maybe better to clarify the S3 access? Awaiting advice here. Thank you!

luisramos0:
Nice Filipe, thank you!

I was trying to fix S3 in FR or UK staging but then I realized S3 is perfectly working and configured in staging AU so I just run the test myself now:
- before PR deployment - check images in S3, upload new image to S3 and verify the S3 link in the shopfront - PASS
- after deploy - same test - PASS

Awesome, I am merging this PR.

filipefurtad0:
Ah so they were indeed broken on both servers - thanks for checking this @luisramos0 !
And for testing this in AU as well :+1:

luisramos0:
The master build is broken because this PR is incompatible with PR https://github.com/openfoodfoundation/openfoodnetwork/pull/5763
There's some code that needs to be adapted for this PR to work with the changes in 5763.
This will fix the build. The errors are related to mail interceptor, so **maybe** it will be related to #5932

luisramos0:
The master build issue is not related to #5962 at all.
It's really just an imcompatibility between 5763 and 5733. This PR here 5733 didnt take into account changes from 5763 where order_mailer.confirm_email was removed.

I just created PR https://github.com/openfoodfoundation/openfoodnetwork/pull/5935 to fix the mater build:
https://github.com/openfoodfoundation/openfoodnetwork/pull/5935

Issue #5721, opened 2020-07-02
Image ofn-logo-mobile.svg broken - Windows 10, IE11


filipefurtad0:
## Description

The image ofn-logo-mobile.svg seems to be broken, on Windows 10, IE11, at lower resolutions and/or smaller screens. This can be observed seen by zooming in and out, when on this browser/system combination.

First observed in staging-UK (v3) on Browserstack screenshots, while testing PR #5682, however not a v3-regression, as this is also it's also observable in staging-es. Confirmed in production as well.

## Expected Behavior

Image should appear in Windows 10, IE 11 (black logo, pic below)
![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/eeeb4c23-c07b-4c9e-b073-4e8dbf43279b)

## Actual Behaviour

Broken image logo appears instead (see gif further below).

## Steps to Reproduce

1. Go to a Windows 10 IE11 or use Browserstack
2. Go to the OFN homepagre
3. Change the window size or zoom in/out to see the broken logo appear.

## Animated Gif/Screenshot

![zoom_windows_ie11](https://user-images.githubusercontent.com/49817236/86261887-0ecdd080-bbb7-11ea-922e-8f71b03e5b81.gif)

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/a75cbb78-7c1a-429f-bf6c-2adc22625824)

## Severity

Seems only to affect this configuration, but it's a major one, and it's observable on our home page:
bug-s3: a feature is broken but there is a workaround

## Your Environment

* Version used: Observable across staging versions and in production as well
* Browser name and version: IE 11
* Operating System and version (desktop or mobile): Windows 10 (Browserstack)

drummer83:
I believe this one progressed to more browsers now. I've seen it on Chrome, Firefox and Edge and in DE, UK, AU, USA.
![20220614_111127.jpg](https://user-images.githubusercontent.com/1790176/173548460-b6e2aa1a-5f42-40b4-9f3a-8c8915b20414.jpg)
![Screenshot_20220614-111227_Chrome.jpg](https://user-images.githubusercontent.com/1790176/173548500-8538f1db-2403-43bc-92ae-e49287c60303.jpg)
![Screenshot_20220614-111315_Chrome.jpg](https://user-images.githubusercontent.com/1790176/173548376-cc3e5e88-1a1c-488d-aff8-4f243a62d100.jpg)

[...]

PR #5749, opened 2020-07-08
Ensure the hero image doesn't pixelate on hamburger menu open

### LINKED ISSUE #1253: [mobile ux] When I expand the hamburger menu on mobile, background image zoom in and overly pixelated
This can be merged directly to master 🎉 

## Description
On mobile/tablet, when expand hamburger menu, the background image zoom in excessively giving a pixelated look.

## Current behavior
Pixelated view, background image zoomed
46837804-66844100-ce02-11e8-84ed-b2b68e34221d

## Expected behavior
No zoom in the image
46837891-a0554780-ce02-11e8-9437-387910acfaee

mbudm:
#### What? Why?

Closes #1253 

This ensures the hero image doesn't attempt to use the full height of all page content:
Screen Shot 2020-07-08 at 9 12 40 pm

Not sure exactly why this happens, but when the mobile nav is opened the hero image at #tagline:before uses the height of the full window - often around 4000px. Adding max-height of 100% to the nearest safe parent prevents this behaviour, without pulling apart the nav css to figure out exactly what's happening.

#### What should we test?
Toggle open the hamburger menu
The hero image should stay the same size

Highly recommend we test this on multiple browsers... (I've tried chrome and safari on mac)

#### Release notes
Fixed hero image pixelating

Changelog Category: Fixed


mbudm:
Hi @luisramos0 - here is the example without the change - you can see that the `max-height:100%;` property is disabled, and the image is enlarged greatly because the height of `.inner-wrap` is almost 4000px (the `#tagline:before` element that contains the background image inherits it's height from `.inner-wrap`).
Screen Shot 2020-07-14 at 12 49 35 pm

When that css property is re-enabled, the image decreases back to the expected height ~ 800px in this case:
Screen Shot 2020-07-14 at 12 48 58 pm

If you aren't seeing this the this is either:
- a cache issue, try clearing cache
- a browser issue, I've tested this on chrome and safari on osx. I could make a static version and go the browsertesting route if that seems to be the issue

mbudm:
Grrr Firefox... Investigating

mbudm:
I stripped back the generated html and got a rudimentary static version, which I've [hosted here](https://mbudm.github.io/scratchpad/index.html). This is pretty broken - but importantly the bit that does work is the tagline layout.

You can check this by toggling the properties I've added - which make up this fix, the image will pixelate again:
Screen Shot 2020-07-14 at 7 28 10 pm

I then used this static mock to test on a random variety of browsers/os:

Opera on windows
Screen Shot 2020-07-14 at 7 21 06 pm

IE11 on windows
Screen Shot 2020-07-14 at 7 19 58 pm

A Galaxy device
Screen Shot 2020-07-14 at 7 19 10 pm

Edge on windows
Screen Shot 2020-07-14 at 7 17 56 pm

luisramos0:
ah, yeah, I forgot to mention, I am on firefiox.
Firefox is 5/6% on most instances but 15% in France. Definitely a major browser. Also because it's the main open source browser.

I am ok with having two height entries, we just need a short code comment to explain that they work for different browsers.

filipefurtad0:
Hi @mbudm ,

I verified issue #1253 and, after staging this PR, went through several environment combinations on Browserstack.

Manually:
- iPhone 11, iPhone 8 - Chrome, Safari
- Samsung Galaxy S20, Samsung Galaxy S9 - Chrome, Firefox

With a small ruby script (click - wait - take screenshot) - woohoo, it worked! :tada: (except for IE)
- Windows 10, Windows 8 - Chrome, Firefox, Edge, IE
- OS X Catalina, OS X High Sierra - Chrome, Firefox, Safari, Edge

You may find all the pics on the google doc:
https://docs.google.com/document/d/1W384cjrdC8AGQYG4GRtYFppgTQLIXekennxWK0BlRUg/edit#

Do you have a Gmail account/access to this folder Steve? If not, do drop me a line so I can add you.

While testing, I've observed the known issue #5721.

This looks great! Ready to go :-)

Issue #4983, opened 2020-03-12
Continue the Subs Placement job when a subscription order creation fails


lin-d-hop:
## Description
Currently the Subscriptions Placement job fails to place an order it kills the Subs Placement job altogether. This means that:
a) We only see one failed order and the reason for it in the Bugsnags
b) If the failure would not have effected every order we end up with a bigger problem than necessary, particularly from a user perspective.

When the Subs placement job fails to place an order we want the job to continue trying for every subscription.

This will mean that problems will be more visible to support people as well.

## Acceptance Criteria & Tests
When the Subs placement job fails an additional message should be sent to Bugsnag and Slack to:

If one sub placement fails, don't kill the job. Post an error to bugsnag and continue placing subscriptions in the job. Potentially with delayed retries. I think a developer will be able to share best practice on this.

I am not sure how a tester will test this.

I think it might be a devtest issue. If any devs can see how testers can test this I would like to hear.

PR #5751, opened 2020-07-08
Add rescue statements with bugsnag alerts to subs jobs

### LINKED ISSUE #4981: Improve Notifications when Subs Placement job fails
## Description

Currently the Subscriptions Placement job fails with very specific error messages related to the order placement that failed. When the Subs Placement job errors the whole job is killed and the remaining subscriptions are not placed. This issue is to update the notifications that are sent to Bugsnag and Slack on failure of the Subs Placement job.

When the Subs placement job fails an additional message should be sent to Bugsnag and Slack to:
- Indicate clearly that a Subs Placement job failed, meaning subscriptions were not placed
- Indicate the name of the Enterprise for which the job failed.
- Indicate the name of the Schedule and Order Cycle for which the job failed if possible.

With the enterprise name an support person should be able to figure out the Schedule and Order Cycle, but if these can be included in the message it will save time and increase confidence of the support person to take the appropriate mitigating action.

I am not sure how a tester will test this.

I think it might be a devtest issue. If any devs can see how testers can test this I would like to hear.

luisramos0:
## What? Why?

With this, when an order placement or confirmation fails, there's a bugsnag alert for it and the job continues processing the rest of the orders.
This basically fixes both #4981 and #4983
Closes #4981 and #4983

#### What should we test?
Subscriptions are really tricky to test with all the processing scenarios. I think this is a good opportunity to think about and improve our test scenarios for error cases. Let's work this together when this comes to Test Ready?

We need to test some error scenarios in subscriptions:
- for confirm job: expired credit card in one of the orders, confirm the rest of the orders still go through (I dont think there's a way to ensure the order with the error is not the last one. maybe we should have two orders with problems in one given test and make sure both errors show up in bugssnag)
- for placement job: tbh I cant remember a real case where the placement job failed... thoughts?

#### Release notes
Changelog Category: Added
The jobs that process subscriptions orders placement and completion are more resilient now and will always process all orders even if one specific order fails.

### DIFF
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -50,6 +50,8 @@ class SubscriptionConfirmJob
     else
       send_failed_payment_email(order)
     end
+  rescue StandardError => e
+    Bugsnag.notify(e, order: order)
   end
 
   def process_payment!(order)
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -61,13 +61,16 @@ class SubscriptionConfirmJob
     raise if order.errors.present?
     return unless order.payment_required?
 
+    prepare_for_payment!(order)
+    order.process_payments!
+    raise if order.errors.any?
+  end
+
+  def prepare_for_payment!(order)
     setup_payment!(order)
     raise if order.errors.any?
 
     authorize_payment!(order)
-    raise if order.errors.any?
-
-    order.process_payments!
     raise if order.errors.any?
   end
 
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -94,5 +94,7 @@ class SubscriptionConfirmJob
     order.update!
     record_and_log_error(:failed_payment, order, error_message)
     SubscriptionMailer.failed_payment_email(order).deliver
+  rescue StandardError => e
+    Bugsnag.notify(e, order: order, error_message: error_message)
   end
 end
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -45,30 +45,30 @@ class SubscriptionConfirmJob
   def confirm_order!(order)
     record_order(order)
 
-    if process_payment!(order)
-      send_confirmation_email(order)
-    else
-      send_failed_payment_email(order)
-    end
+    process_payment!(order)
+    send_confirmation_email(order)
   rescue StandardError => e
-    Bugsnag.notify(e, order: order)
-    send_failed_payment_email(order, e.message)
+    if order.errors.any?
+      send_failed_payment_email(order)
+    else
+      Bugsnag.notify(e, order: order)
+      send_failed_payment_email(order, e.message)
+    end
   end
 
+  # Process the order payment and raise if it's not successful
   def process_payment!(order)
-    return false if order.errors.present?
-    return true unless order.payment_required?
+    raise if order.errors.present?
+    return unless order.payment_required?
 
     setup_payment!(order)
-    return false if order.errors.any?
+    raise if order.errors.any?
 
     authorize_payment!(order)
-    return false if order.errors.any?
+    raise if order.errors.any?
 
     order.process_payments!
-    return false if order.errors.any?
-
-    true
+    raise if order.errors.any?
   end
 
   def setup_payment!(order)
--- a/app/jobs/subscription_confirm_job.rb
+++ b/app/jobs/subscription_confirm_job.rb
@@ -52,6 +52,7 @@ class SubscriptionConfirmJob
     end
   rescue StandardError => e
     Bugsnag.notify(e, order: order)
+    send_failed_payment_email(order, e.message)
   end
 
   def process_payment!(order)
@@ -89,9 +90,9 @@ class SubscriptionConfirmJob
     SubscriptionMailer.confirmation_email(order).deliver
   end
 
-  def send_failed_payment_email(order)
+  def send_failed_payment_email(order, error_message = nil)
     order.update!
-    record_and_log_error(:failed_payment, order)
+    record_and_log_error(:failed_payment, order, error_message)
     SubscriptionMailer.failed_payment_email(order).deliver
   end
 end
--- a/app/jobs/subscription_placement_job.rb
+++ b/app/jobs/subscription_placement_job.rb
@@ -29,8 +29,14 @@ class SubscriptionPlacementJob
 
   def place_order_for(proxy_order)
     JobLogger.logger.info("Placing Order for Proxy Order #{proxy_order.id}")
-    proxy_order.initialise_order!
+    initialise_order(proxy_order)
     place_order(proxy_order.order)
+  end
+
+  def initialise_order(proxy_order)
+    proxy_order.initialise_order!
+  rescue StandardError => e
+    Bugsnag.notify(e, subscription: proxy_order.subscription, proxy_order: proxy_order)
   end
 
   def place_order(order)
@@ -42,8 +48,9 @@ class SubscriptionPlacementJob
 
     move_to_completion(order)
     send_placement_email(order, changes)
-  rescue StateMachine::InvalidTransition
-    record_and_log_error(:processing, order)
+  rescue StandardError => e
+    record_and_log_error(:processing, order, e.message)
+    Bugsnag.notify(e, order: order)
   end
 
   def cap_quantity_and_store_changes(order)
--- a/engines/order_management/app/services/order_management/subscriptions/summarizer.rb
+++ b/engines/order_management/app/services/order_management/subscriptions/summarizer.rb
@@ -22,12 +22,15 @@ module OrderManagement
         summary_for(order).record_issue(type, order, message)
       end
 
-      def record_and_log_error(type, order)
+      def record_and_log_error(type, order, error_message = nil)
         return record_issue(type, order) unless order.errors.any?
 
         error = "Subscription#{type.to_s.camelize}Error"
         line1 = "#{error}: Cannot process order #{order.number} due to errors"
-        line2 = "Errors: #{order.errors.full_messages.join(', ')}"
+
+        error_message ||= order.errors.full_messages.join(', ')
+        line2 = "Errors: #{error_message}"
+
         JobLogger.logger.info("#{line1}\n#{line2}")
         record_issue(type, order, line2)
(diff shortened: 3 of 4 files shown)


app/jobs/subscription_confirm_job.rb
```diff
@@ -50,6 +50,8 @@ def confirm_order!(order)
     else
       send_failed_payment_email(order)
     end
+  rescue StandardError => e
+    Bugsnag.notify(e, order: order)
```
sauloperez, line 54: Re #4981 we already have a bugsnag errors when the processing fails :point_right: https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5eb1a9a5e67a47001777d396?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=delayed. What we need to do is rescue the `StateMachine::InvalidTransition` error and then execute `send_failed_payment_email(order)`. 

However, this won't be enough because `SubscriptionConfirmJob#process_payment!` and `Spree::Order#process_payment!` can both return boolean or raise. Therefore, we need to keep that `else` branch above. Yes, those methods need to reduce their complexity. It's a pain in the ass to work with such broad interface/return types. And we're paying its costs already.

app/jobs/subscription_placement_job.rb
```diff
@@ -29,10 +29,16 @@ def proxy_orders
...
     JobLogger.logger.info("Placing Order for Proxy Order #{proxy_order.id}")
-    proxy_order.initialise_order!
+    initialise_order(proxy_order)
     place_order(proxy_order.order)
   end
 
+  def initialise_order(proxy_order)
+    proxy_order.initialise_order!
+  rescue StandardError => e
+    Bugsnag.notify(e, subscription: proxy_order.subscription, proxy_order: proxy_order)
```
sauloperez, line 39: There's no actual need for the call to notify. Workers already notify unhandled exceptions in [config/initializers/delayed_job.rb](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/config/initializers/delayed_job.rb). If what we want is the extra diagnosis details from the subs. I'd rather go with https://docs.bugsnag.com/platforms/ruby/rails/#attaching-custom-diagnostics in that initializer. We already do it in few places.

app/jobs/subscription_confirm_job.rb
luisramos0: the objective of this statement is not to generate the bugsnag, the bugsnag is already there but it was blowing up the process, now the bugsnag will be generated and the process will continue to the next order. right?

luisramos0: I think we need to rescue StandardError/any exception, otherwise it will blow up. I think I am answering the following comment already: I dont think we need to have a worker for each order, that sounds like over-engineering for this problem. This simple rescue StandardError for each order already does the job of continuing processing the following orders. Let's discuss this tomorrow.

I agree we need to send the failed email anyway :+1: I added a new commit to do that, please have a look.

app/jobs/subscription_placement_job.rb
luisramos0: As mentioned above, this is already the solution for #4983 (continue processing the rest of the orders), it's a simple catch all and log that enables the job to continue doing what's doing. I think this is way easier and equally effective than creating a job to process each proxy_order.

app/jobs/subscription_confirm_job.rb
luisramos0: For context: we did a pair programming session and we agreed persisting summaries and making each order being processed by a separate job is a good idea. We will get there. For now we can go with this quicker approach of adding the rescue statements.

app/jobs/subscription_confirm_job.rb
```diff
@@ -45,27 +45,33 @@ def recently_closed_order_cycles
   def confirm_order!(order)
     record_order(order)
 
-    if process_payment!(order)
-      send_confirmation_email(order)
-    else
+    process_payment!(order)
+    send_confirmation_email(order)
+  rescue StandardError => e
```
sauloperez, line 50: This will now handle errors like https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5f104f65e9f6490017ad5250?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=delayed, which just happened.

filipefurtad0:
Hi @luisramos0 and @lin-d-hop ,

I probed different types of (invalid) cards, in order to assess how other (valid) orders would get influenced. Addressing the point you raise Luis, i.e., to ensure the order with the error is not the last one, groups of three subscriptions were tested, per schedule:

In staging FR - OC (id 74)
- subscription (id. 12) - customer A (invalid card)
- subscription (id. 13) - customer B (valid card)
- subscription (id. 13) - customer A (invalid card)

"Sandwiching" the valid order was reflected on confirmation emails, which at least indicates, that this processing order was followed, so it could be preceded from an invalid order.

In all tests, customer B had the valid 424242... card.

Moved on to the following test cases, following the procedure described [here](https://github.com/openfoodfoundation/openfoodnetwork/wiki/Testing-Subscriptions):

**- Test case 1**

In test case 1, customer A used a valid 3D card, which is accepted the page `/account#/cards` but not supported in the backoffice, so payment is bound to fail.

Emails arrive as expected, after:
`SubscriptionPlacementJob.new.perform`
each customer: order creation
hub: placed subscription orders

`SubscriptionConfirmJob.new.perform`
for customer A: email explaining payment failed (on two orders)
for customer B: order placement
hub: confirmed subscription orders


Before staging the PR
image


After staging the PR
image

Having an failed order does not seem to influence the placement of an order with a valid card. However, this did not seem to trigger any bugsnag errors, neither on Slack nor on the Bugsnag dashboard.


**- Test case 2**

In test case 2, customer A was set with a fraudulent card (4100000000000019), but payment should blocked. 

In this case, emails do not arrive as expected, after:

`SubscriptionPlacementJob.new.perform`
each customer: order creation
hub: placed subscription orders

But after confirmation `SubscriptionConfirmJob.new.perform` no email arrives:
customer: no placement of orders nor email or failing payment
hub: no subscription confirmation

Looking at the /orders page, we see:

Before this PR
image

After this PR
image

Both before and after staging the PR, the "middle"/valid order is not placed. Also here: no bugsnag errors.


So the process seems to break, using card a valid but fraudulent card, but works for a 3D-secure cards, in the sense that payments are not processed, while not preventing legitimate orders to be placed.


I am wondering why are we not seeing bugnsnag, in any of these cases. Which cases should be tested in order to observe this?
Grateful for feedback.

luisramos0:
thanks Filipe, I will have to try myself. I'll get back to you, maybe only next week, I have quite a few things on my plate right now.

sauloperez:
I can take a look if time allows @luisramos0 It's pretty important to get this ASAP

sauloperez:
another example https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5f291c95eb9ead0018c3fc42?event_id=5f291c95005b6b54d3ef0000&i=sk&m=nw. This might be relevant while testing @filipefurtad0

filipefurtad0:
Hi team,  

I had a second go at the second test case, once again. All seemed to work now:

Setting up a fraudulent card (4100000000000019) blocks the payment for the first and third orders (customer A), but the middle one proceeds fine (customer B).


Emails arrive as expected, after:
SubscriptionPlacementJob.new.perform
each customer: order creation
hub: placed subscription orders

SubscriptionConfirmJob.new.perform
for customer A: email explaining payment failed (on two orders)
for customer B: order placement
hub: confirmed subscription orders, pointing out that two of the placed orders failed confirmation

_Paiement échoué (2 commandes)_

_Le traitement automatique des paiements pour ces commandes a échoué. L'erreur a été affichée à l'endroit pertinent._
_R433887116,R738717157_


This bugsnag error is triggered:

https://app.bugsnag.com/open-food-france/open-food-france/errors/5f311d925ef88a00186859e8?event_id=5f311d92005b6ec70e9f0000&i=sk&m=nw

Looking at the /orders page, we see:

image


So, I guess this is the expected result from this PR. I can't really explain why it did not work before (repeatedly). There was no additional commits to this PR, so maybe human error from the tester (me)? Feels strange but reassuring to see the bugsnag error now.

Good to go from my side.

sauloperez:
For the record, the relevan log lines from `logs/delayed_job.log`:

```
2020-08-10T12:12:31+0200: Confirming Order for Proxy Order 22
2020-08-10T12:12:34+0200: Issue in Subscription Order 1545: failed_payment
2020-08-10T12:12:34+0200: Confirming Order for Proxy Order 23
2020-08-10T12:12:40+0200: Confirming Order for Proxy Order 24
2020-08-10T12:12:43+0200: Issue in Subscription Order 1547: failed_payment
```

and from `logs/staging.log`:

```
E, [2020-08-10 12:12:34#15269] ERROR -- : Gateway Error
E, [2020-08-10 12:12:34#15269] ERROR -- :   --- !ruby/object:ActiveMerchant::Billing::Response
params:                                               
  error:                                              
    charge: ch_1HEXzZKuuB1fWySnZJUi4MN8  
    code: card_declined    
    decline_code: fraudulent
    doc_url: https://stripe.com/docs/error-codes/card-declined
    message: Your card was declined.
(...)
message: Your card was declined.
success: false
test: false
authorization: ch_1HEXzZKuuB1fWySnZJUi4MN8
fraud_review:
error_code: card_declined
emv_authorization:
avs_result:
  code:
  message:
  street_match:
  postal_match:
cvv_result:
  code:
  message:

E, [2020-08-10 12:12:43#15269] ERROR -- : Gateway Error
E, [2020-08-10 12:12:43#15269] ERROR -- :   --- !ruby/object:ActiveMerchant::Billing::Response
params:                                                                                                                         
  error:                                                     
    charge: ch_1HEXziKuuB1fWySnrrCmf9UI                      
    code: card_declined                                                                                                         
    decline_code: fraudulent                         
    doc_url: https://stripe.com/docs/error-codes/card-declined
    message: Your card was declined.
    payment_intent:      
      id: pi_1HEXzhKuuB1fWySnAuZL2lcM
      object: payment_intent
      allowed_source_types:
      - card
      amount: 6050
(...)
    type: card_error
message: Your card was declined.
success: false
test: false
(...)
```

sauloperez:
I'm merging although because the failed spec has been long improved since this PR opened. The build in master should then be green.

PR #5753, opened 2020-07-08
Reports rewrite

Matt-Yorkley:
Includes the previous [Reports POC](https://github.com/openfoodfoundation/openfoodnetwork/issues/5576) plus:

- moves new reports code to OrderManagement engine
- adds test coverage
- includes a re-implementation of both Packing reports in the new style
- moves existing packing reports specs into engine and deletes old packing reports code
- integrates the new reports with existing admin reports page
- adds missing translation keys
- implements custom error classes
- refactors various bits
- moves shared controller code into a controller concern (Rails 4 introduces controller concerns :tada:)


Matt-Yorkley:
I updated our codeclimate rubocop version to match our actual rubocop gem version, but it's now counting lots of Layout/LineLength violations in files where we've added exclusions... :man_shrugging:

PR #5756, opened 2020-07-10
Make broken spec fail reliably and set it pending

mkllnk:
#### What? Why?

This spec has been broken for a long time, at least eight months. But it
regularly passed because the search filter is applied with a delay and
in that time the content matches. And once the filter is applied, no
products are shown and the negative matchers pass.

#### What should we test?

No test.

#### Release notes

More accurate tests.

Changelog Category: Fixed


mkllnk:
The bug has been described in #5757.

sauloperez:
This needs rebasing to fix the CI error @mkllnk :warning:

Issue #3584, opened 2019-03-06
Voiding payment does not remove transaction fee


kristinalim:
Voiding either captured or uncaptured payment does not remove transaction fee.

## Description

I'm not sure if the payment method is relevant, but this was observed by @RachL for cash and myself for cheque. 

## Expected Behavior

When you void an existing payment and add another payment, you should only have the transaction fee for the payment that wasn't voided.

## Actual Behavior

When you void an existing payment and add another payment, you end up with two transaction fees.

## Steps to Reproduce

For below, create a cheque or cash payment method for the distributor. Take note of the expected calculation rate.

**As enterprise manager:**

Voiding payment that has been captured:

1. Create an order selecting the payment method earlier.
2. Complete the order without capturing payment. Check that the transaction fee is correct.
3. Capture payment. Check that the transaction fee is still correct.
4. Void the payment.
5. Add another payment. Notice in "Order Details" that there are now two transaction fees charged.

Voiding payment that was not captured:

1. Create another order.
2. Complete the order without capturing payment.
3. Mark the uncaptured payment void.
4. Add another payment to the order. Notice in "Order Details" that there are now two transaction fees charged.

## Context

Discovered by @RachL (and confirmed by myself) while testing #3563.

## Severity

bug-s3: a feature is broken but there is a workaround

You can set the transaction fee to $0.

luisramos0:
This will be related to https://github.com/openfoodfoundation/openfoodnetwork/pull/3559
I'd bet the transaction fee adjustment is being voided as well as the payment but recalculated/re-added afterwards in order.update_payment_fees

PR #5758, opened 2020-07-10
Bring Order Updater from spree_core

luisramos0:
#### What? Why?

Part of #4826 this PR brings Spree::OrderUpdater from spree_core, merges it with OrderUpdater delegator and moves it to a service in OrderManagement engine :tada:

What a wonderful PR :beach_umbrella: 

#### What should we test?
This PR moves code that updates the order whenever something changes so we need to test all the places where an order may be updated, mostly in terms of:
- payment state, payment total and payment fees
- shipment_state and shipping fees
- other adjustments like manual adjustments or enterprise fees
- order totals

In the shopfront when an item is added to the cart and has a fee, that fee needs to be included in the item price as well as reflected in the order total.
In the backoffice, when we change the order in terms of items with fees or adjustments, the values of the fees and the order total should also be updated correctly, for example, adding items with fees, adding manual adjustments, changing shipping method to one with a different fee, etc.

#### Release notes
Changelog Category: Changed
Brought some core Order Management code from Spree to OFN so we can simplify it and improve it


sauloperez:
Unless someone disagrees I'm making the call to this to Test Ready. Given the time it's been sitting there and the nature of the fixes we're doing recently, it's fairly easy that it'll end up with merge conflicts that will bring up the costs of the tech debt.

app/models/spree/order_decorator.rb
```diff
@@ -378,6 +378,13 @@ def address_from_distributor
     address
   end
 
+  # Update attributes of a record in the database without callbacks, validations etc.
+  #   This was originally a ext to ActiveRecord in Spree but only used for Spree::Order
+  def update_attributes_without_callbacks(attributes)
```
mkllnk, line 383: And since Rails 3.1 there is `update_columns` which does the same.
https://apidock.com/rails/ActiveRecord/Persistence/update_columns

filipefurtad0:
Hi @luisramos0,

Thanks for pointing out the main test areas!

- Checked payment state: verified that the states "Balance due", "Paid", "Credit owed", checking that these different status are interchangeable by editing an order. "Payment" state appears on orders which are in the cart state, as before. "Checkout" state turns into "paid" after payment capture.

- shipment state - changes accordingly after shipment (shipped) to "awaiting return", to "returned" after these operations were performed on the respective order.

- shipping fees are applied on checkout and appear correctly on Enterprise Fee Reports

- adjustments like manual adjustments and updating or enterprise fees work as before on the backoffice

- enterprise fees appear on the price in the shopfront, and throughout the shopping process. After checkout it correctly appears in the order total

- performed a sanity check in what concerns order changes in the backoffice, including recalculation of fees, changes on shipment methods, and payment fees - all good (bumped into #3584)

Awesome, this looks good to go!

luisramos0:
Awesome, goof testing!!!

Issue #3243, opened 2018-12-23
Multi language Email issues


luisramos0:
Follow up from #3189, there are currently 6 different issues with multi language emails, this epic represents those issues. When the work is started, we can create individual issues. Some solutions are still not defined, here's a summary of my analysis:

This wiki page summaries all emails: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Emails

### 1. Forgotten password email (case 3 in the wiki)
The email is always sent in English. This is done in devise (a dependency of spree).
Additionally, the email itself is not styled, so we may also need to redo the email template.

### 2. Order shipped email (case 7 in the wiki)
The email is always sent in English. This is done in spree.
Additionally, the email itself is not styled, so we may also need to redo the email template).
To replicate this one you need to go to the bulk order management page and before you see the ship button you need to capture payment.

### 3. Enterprise manager (new account) invitation email and sign up email (case 10.1 and 10.2 in the wiki)
There's no way to know whats the new user locale, the best option here is to user current user locale.

### 4. Subscription confirmation email I (case 12 in the wiki) and then all subs emails (case 13-17)
When creating a subscription for a new customer there's no way of knowing the customer locale. Again, the best option is to use the current user locale. I am not sure but in this case, I think there will be no user created so there will be no place to store the locale.

### 5. Order created in the backoffice (no described in wiki)
Same as in subscriptions, when we create an order in the backoffice for a new customer that has never logged in, there's no way to know what the user locale. Best option is to use current user locale. I am not sure a user record is created in this case...

### 6. GUEST user checkout emails
 Guest user is not registered in the DB so there is no place to store the user locale (I thought of using the session locale but the email is sent on delayedJob which is a different process without access to the session locale).
@maikel commentated: “I think the clearest way to use Delayed Job is to create custom jobs for whatever we want to delay. These jobs can store the current locale of the guest and use it for sending the email.“ 

In these 3 last cases of 1. guest user checkout, 2. order created in the backoffice and 3. subscriptions with new customer, the same problem applies to all order emails (resend confirmation in the backoffice, order cancelation, invoice, etc). All these emails need to be in the user locale, but in some cases, there is no user record to store the locale. This needs to be investigated.

luisramos0:
#3043 was an s3. I keep this one in s3. Specially for the first two cases where "forgot password" and "order shipped" emails are still being sent in English :-/

luisramos0:
[...]

Issue #2910, opened 2018-10-22
Improve order cancellation email


myriamboure:
## Description
If when I cancel an order as a customer, or if a hub manager do it from backoffice, as a customer I receive an email but there are some important info missing:
-  if the order is cancelled, we should add a sentence "if you didn't want to cancel that order, please contact [hub contact email]"
- there is no indication if the order was paid if I will be refunded, or other.

## Expected Behavior
The email I receive is clear, I understand what I have cancelled (content of the order) and I know what action to do if I made a mistake. I know what happens with my payments.

**Template proposal:**
![order cancellation](https://user-images.githubusercontent.com/8135376/51735378-955bb580-2087-11e9-8d62-b5086909eb35.jpg)

With the last sentence we can have 2 options (depending if order has been paid or not):
- "Your order was unpaid so no refund has been made"
- "Your order was paid so Superhub has refunded the full amount"
- And later when we have credits: "Your order was paid so your account has been credited"

## Actual Behavior
![Capture du 2019-05-21 11-50-58](https://user-images.githubusercontent.com/8135376/58086346-be67ff00-7bbe-11e9-92e9-778a230b6d3d.png)

Some info are missing (see above)

## Steps to Reproduce
1. In an OC with a hub that allows to modify order until order closes, make an order.
2. Then go to modify the order and click on "cancel order"
3. Check your mailbox.

## Your Environment

* Version used: V2.1 on Katuma staging
* Browser name and version: Firefox and Chrome
* Operating System and version (desktop or mobile): Ubuntu 18

kirstenalarsen:
I think you're hacking the bug system @myriamboure . . is this a bug or a desired enhancement?

RachL:
@myriamboure shouldn't it be 3 separate issues? Plus I'm not sure they are qualified as bugs...

myriamboure:
I am not sure either but what is clearly a bug to me is the display of enterprise fees details to customers. The rest is not a bug but probably it should be treated at once if someone review this order candellation email, so I was not sure @RachL We could separate in three issues but we should ask a dev first if it's easier to do it all at once or not.

sigmundpetersen:
Just convert this issue to a sub-epic, create 3 new issues and put bug label on the one with the fees?

myriamboure:
[...]

PR #5763, opened 2020-07-14
Bring Mailers from Spree and make order shipped email translatable

### LINKED ISSUE #5686: [ByeBye Spree] Move Mailers to OFN
Move all Mailers to OFN and merge with decorators.

These files:
- core/app/mailers/spree/base_mailer.rb MERGE WITH DECORATOR
- core/app/mailers/spree/order_mailer.rb MERGE WITH DECORATOR
- core/spec/mailers/order_mailer_spec.rb MERGE WITH DECORATOR
- core/app/mailers/spree/shipment_mailer.rb COPY (and improve!)
- core/app/mailers/spree/test_mailer.rb COPY
- core/app/views/spree/shipment_mailer/shipped_email.text.erb COPY
- core/app/views/spree/test_mailer/test_email.text.erb COPY
- core/spec/mailers/shipment_mailer_spec.rb COPY
- core/spec/mailers/test_mailer_spec.rb COPY

### Context
Part of #4826

### Impact and timeline

luisramos0:
#### What? Why?

Closes #5686
and addresses point 2 in #3243

Move all needed mailers from spree_core to OFN.

This was we can now translate test email and order shipped email (this email still needs a better template, but that will be for next time).

#### What should we test?
Here we change the base_mailer so better verify email basics like order confirmation and order cancellation for example. Are they sent correctly? Is the data in the email correct? Is the invoice email sent correctly?
Test the emails that were changed:
- test email - is this email sent when the button pressed in the backoffice mail settings?
- shipped email - is this email sent when order is marked as shipped? Any errors in the old un-styled template?

#### Release notes
Changelog Category: Changed
Brought test and shipment emails from Spree. Order Shipped email is now translatable!


app/views/spree/shipment_mailer/shipped_email.html.haml
```diff
@@ -0,0 +1,19 @@
...
+  = item.variant.product.name
+  = item.variant.options_text
+= "============================================================"
+
+- if @shipment.tracking
+  = Spree.t('shipment_mailer.shipped_email.track_information', tracking: @shipment.tracking)
+- if @shipment.tracking_url
+  = Spree.t('shipment_mailer.shipped_email.track_link', url: @shipment.tracking_url)
+
+= Spree.t('shipment_mailer.shipped_email.thanks')
```
sauloperez: Are you sure switching from .text to .HTML won't break things? I never investigated this but I think .text is used as a fallback text-version of the email for specific old mail servers/clients. Again, no idea about how it works.

luisramos0: Spree emails are all text, and ofn's are html. So I think it's fine, from now on these will also be sent in html. Testing will validate that as well.

I think you are referring to something we _could_ do which is to send html and text emails and the email client would use text as a fall back option. This is also good for deliverability but it's something we are not doing and is out of scope for this PR. Makes sense?

sauloperez: Understood, I thought those text ones were just a fallback and not Spree's only templates :+1: Then, this means that those only-text emails saw Eugeni some time ago were a bug or they were missing an OFN template :see_no_evil:

luisramos0: The emails seen by Eugeni are this shipment email. It was text in spree and after this PR it's html in OFN. But still without the normal OFN styling. That will need to be addressed in another PR.

RachL:
OMG I'm just discovering this PR!!! Being able to translate this email is such a GOOOD news!!

luisramos0:
I know!! :-)

filipefurtad0:
Hi @luisramos0,

Triggered the following emails, before and after the PR:

Test
Order shipped
Confirm Account 
Welcome to OFN
Reset Passwork
Order Confirmation
Order Cancellation



- Test emails

Before

image


After

image




- Order shipped Emails


Before


image


After

image



- Confirm Account Email - OK, before and after

- Welcome to OFN Email - OK, before and after

- Reset Passwork Email - OK, before and after 



- Order Confirmation emails

These seem to be broken after this PR. I checked this both by clicking "Resend" or "Resend Confirmation" in the backoffice, and in both in this case, the user sees:

image


After placing a new order, neither the Hub nor the customer receive confirmation email. This is a silent error, no snail seen. Check Bugsnag errors for details:

https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5f228433c37fb80018f09297


- Order cancellation emails are also broken, after this PR

image



The changes introduced by the PR work fine, but order confirmation/cancellation emails are quite a critical, so perhaps best not to introduce these in the code, right? I'm moving the PR back to In Dev for a second look.



Also, a product question maybe @RachL @lin-d-hop : I'm not sure what's the difference between the two buttons (pic below). "Resend Confirmation" triggers the modal asking for confirmation but other than that, is there a difference in function?

image

RachL:
> Also, a product question maybe @RachL @lin-d-hop : I'm not sure what's the difference between the two buttons (pic below). "Resend Confirmation" triggers the modal asking for confirmation but other than that, is there a difference in function?

Yes I've always wondered as well. 🤷

filipefurtad0:
In that case, maybe we could consider removing one of them, maybe as a papercut - if appropriate.

The "Resend Confirmation" button has a more descriptive naming and already has the associated modal, which is also helpful I think, so we could keep this one. Thoughts?

RachL:
Sounds good :+1: Let's re-test maybe just to see if it is really 100% the same? Maybe you already did?

filipefurtad0:
Yes, I found no difference - the emails looked identical, but I'd wait for feedback from core Devs as well. Perhaps @luisramos0 and/or reviewers can confirm that the code behind these features is the same?

luisramos0:
yes, the code is the same behind both resends :+1:

Thanks for testing Filipe, I cant replicate that error locally. It's a tricky one related to class loading I think because the "missing method" is there and included as part of one of the helper classes... it should work.

I'll have to come back to this later.

filipefurtad0:
Thanks for your feedback @luisramos0.

The snail-errors were reproducible in staging environments (UK and FR). But thinking of it now, I am not entirely sure the PR broke the order confirmation emails as well, or they just got queued after the snail errors. Maybe testing in a different order would result in a different outcome. I'll try this and provide some additional feedback.

Thanks as well for confirming the code on the "Resend" button :+1: 
Opened an Enhancement/Papercut #5850 on this regard.

filipefurtad0:
I did a quick double-checked now:
Order confirmation emails trigger a bugsnag [error](https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5f228433c37fb80018f09297?filters[event.since][0]=30d&filters[error.status][0]=open) as well, and don't arrive the mailbox. This happens also after placing an order immediately after staging the PR, so it appears independent of those errors seen by the user (order cancelling and resend confirmation).

luisramos0:
ok, this was easier than I thought. Yes, there was a bug in the code. There's Spree::CheckoutHelper in Spree that was hidden our CheckoutHelper. I managed to replicate the problem you describe, I pushed a new simple commit (no need for code review) and I restaged the PR. The problem is now gone. This is ready for testing again :+1:

luisramos0:
In staging I got the shipment email like this:
image

So I added some paragraphs and breaklines to the template so it doesnt happen again.
We need to revisit this email template. I think we need a new issue to redo the shipment email template, like this one: https://github.com/openfoodfoundation/openfoodnetwork/issues/2910

luisramos0:
ok, the email now looks like this :+1:

image

RachL:
I've re-tested them all:

- Test : I'm not sure how to get the email. Should I turn one of my accounts as super admin? 
- Order shipped ✅ 
- Confirm Account ✅ 
- Welcome to OFN ✅ 
- Reset Password ✅ 
- Send Invoice ✅ 
- Order Confirmation ✅ 
- Order Cancellation ✅ 

As all email works perfectly, I'm moving this to ready to go!

RachL:
Nevermind: I found how to test the test email :)

image

PR #5766, opened 2020-07-15
5764 Provide credit card brand to Pin Payments

### LINKED ISSUE #5764: PinPayments not supported in OFN v3
## Description

Selecting a PinPayments payment method during checkout results in the error:
> That payment method is unsupported. Please choose another one.

Unfortunately, this enterprise doesn't have any other payment methods configured. We have 15 active payment methods used by 33 enterprises. 5 of those enterprises have only PinPayments configured. All other enterprises have at least on other payment method available.

Only one enterprise has active order cycles: `2876`

Enterprise ids with only PinPayments: `284, 648, 1899,  2220, 2876`

All affected enterprise ids:
```rb
Spree::PaymentMethod.where(type: "Spree::Gateway::Pin").where(active: true).joins(:distributors).uniq.pluck('enterprises.id')
# 56, 86, 95, 111, 112, 121, 123, 134, 140, 141, 189, 220, 259, 265, 283, 284, 319, 397, 502, 520, 524, 624, 648, 666, 738, 1899, 1959, 1960, 2028, 2137, 2141, 2220, 2876
```

## Expected Behavior

PinPayments works during checkout as before. It has been broken in the backend for a while (#2514)

## Actual Behaviour

The checkout fails and an error message is displayed at the top.

## Steps to Reproduce

1. Select a PinPayments payment method during checkout
2.
3.
4.

## Animated Gif/Screenshot

Screenshot from 2020-07-15 10-31-48

## Severity

## Your Environment

* Version used: v3.1.0 on Australian production
* Browser name and version:

## Bugsnag

**RuntimeError** in **checkout#update**
That payment method is unsupported. Please choose another one.

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5f0e3bf28137c100180b4e75?event_id=5f0e3d55005b85c5ac910000&i=gh&m=ci)

## Stacktrace

    app/controllers/checkout_controller.rb:242 - update_failed
    app/controllers/checkout_controller.rb:178 - checkout_workflow
    app/controllers/checkout_controller.rb:52 - update
    app/services/current_order_locker.rb:11 - block in around
    app/services/current_order_locker.rb:24 - block in lock_order_and_variants
    app/services/current_order_locker.rb:22 - lock_order_and_variants
    app/services/current_order_locker.rb:11 - around

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5f0e3bf28137c100180b4e75?event_id=5f0e3d55005b85c5ac910000&i=gh&m=ci)

mkllnk:
#### What? Why?

Closes #5764

Spree changed the way they determine the brand of a credit card. They do it via Javascript and removed the Ruby code. We were missing the card's name and without it a payment method can't determine if it supports that credit card. :bomb: 

#### What should we test?

* Create a payment method with Pin Payments.
* You can obtain a test API key by creating an account at Pin Payments or use the one I saved in Bitwarden.
* Set the server to `test` and check test test mode box.
* Open an order cycle with that enterprise.
* Check out using a test credit card: https://pinpayments.com/developers/api-reference/test-cards
* The error message from the issue should not come up.
* If you use an unsupported card brand like Diners `30000000056030` you should see the error message:
  > That payment method is unsupported. Please choose another one.

#### Release notes

Fixed Pin Payments payment method

Changelog Category: Fixed

The last Spree upgrade caused the bug.

### DIFF
--- a/app/services/checkout/form_data_adapter.rb
+++ b/app/services/checkout/form_data_adapter.rb
@@ -12,6 +12,8 @@ module Checkout
 
       move_payment_source_to_payment_attributes!
 
+      fill_in_card_type
+
       set_amount_in_payments_attributes
 
       construct_saved_card_attributes if @params[:order][:existing_card_id]
@@ -29,6 +31,18 @@ module Checkout
                     payment_source_params = delete_payment_source_params!
 
       @params[:order][:payments_attributes].first[:source_attributes] = payment_source_params
+    end
+
+    def fill_in_card_type
+      payment = params[:order][:payments_attributes]&.first&.dig(:source_attributes)
+
+      return if payment&.dig(:number).blank?
+
+      payment[:cc_type] ||= card_brand(payment[:number])
+    end
+
+    def card_brand(number)
+      ActiveMerchant::Billing::CreditCard.brand?(number)
     end
 
     def delete_payment_source_params!
--- a/app/services/checkout/form_data_adapter.rb
+++ b/app/services/checkout/form_data_adapter.rb
@@ -34,11 +34,16 @@ module Checkout
     end
 
     def fill_in_card_type
-      payment = params[:order][:payments_attributes]&.first&.dig(:source_attributes)
+      return unless payment_source_attributes
 
-      return if payment&.dig(:number).blank?
+      return if payment_source_attributes.dig(:number).blank?
 
-      payment[:cc_type] ||= card_brand(payment[:number])
+      payment_source_attributes[:cc_type] ||= card_brand(payment_source_attributes[:number])
+    end
+
+    def payment_source_attributes
+      @payment_source_attributes ||=
+        params[:order][:payments_attributes]&.first&.dig(:source_attributes)
     end
 
     def card_brand(number)
--- a/app/services/checkout/form_data_adapter.rb
+++ b/app/services/checkout/form_data_adapter.rb
@@ -33,6 +33,11 @@ module Checkout
       @params[:order][:payments_attributes].first[:source_attributes] = payment_source_params
     end
 
+    # Ensures cc_type is always passed to the model by inferring the type when
+    # the frontend didn't provide it. This fixes Pin Payments specifically
+    # although it might be useful for future payment gateways.
+    #
+    # More details:  app/assets/javascripts/darkswarm/services/checkout.js.coffee#L70-L98
     def fill_in_card_type
       return unless payment_source_attributes
 
--- a/spec/services/checkout/form_data_adapter_spec.rb
+++ b/spec/services/checkout/form_data_adapter_spec.rb
@@ -36,6 +36,26 @@ describe Checkout::FormDataAdapter do
         end
       end
 
+      describe "and a credit card is provided" do
+        before do
+          params[:order][:payments_attributes].first[:source_attributes] = {number: "4444333322221111"}
+        end
+
+        it "fills in missing credit card brand" do
+          expect(adapter.params[:order][:payments_attributes].first[:source_attributes][:cc_type]).to eq "visa"
+        end
+
+        it "leaves an existing credit card brand" do
+          params[:order][:payments_attributes].first[:source_attributes][:cc_type] = "test"
+          expect(adapter.params[:order][:payments_attributes].first[:source_attributes][:cc_type]).to eq "test"
+        end
+
+        it "doesn't touch the credit card brand without a number" do
+          params[:order][:payments_attributes].first[:source_attributes][:number] = ""
+          expect(adapter.params[:order][:payments_attributes].first[:source_attributes].key?(:cc_type)).to eq false
+        end
+      end
+
       describe "and existing credit card is provided" do
         before { params[:order][:existing_card_id] = credit_card.id }
 


Matt-Yorkley:
There's some discussion in Slack about deprecating / removing this payment method... it's only used on Aus and possibly only by one enterprise, is that right? And there's potentially a plan for removing it soon?

If it's being deprecated I think we can just go ahead with this fix for now, it's not worth adding new javascript dependencies or spending more dev time on it than we need to...

app/services/checkout/form_data_adapter.rb
```diff
@@ -31,6 +33,18 @@ def move_payment_source_to_payment_attributes!
       @params[:order][:payments_attributes].first[:source_attributes] = payment_source_params
     end
 
+    def fill_in_card_type
+      payment = params[:order][:payments_attributes]&.first&.dig(:source_attributes)
+
+      return if payment&.dig(:number).blank?
```
luisramos0, line 39: if payment is nil this will not return. I think it should be
return unless payment&.dig(:number).present?

app/services/checkout/form_data_adapter.rb
```diff
@@ -31,6 +33,18 @@ def move_payment_source_to_payment_attributes!
       @params[:order][:payments_attributes].first[:source_attributes] = payment_source_params
     end
 
+    def fill_in_card_type
+      payment = params[:order][:payments_attributes]&.first&.dig(:source_attributes)
```
luisramos0, line 37: this is not a payment, these are payment_source_attributes

app/services/checkout/form_data_adapter.rb
luisramos0: I think this is how I'd write this method:

```ruby
def fill_in_card_type
     return unless payment_source_attributes
     
     return if payment_source_attributes.dig(:number).blank?

     payment[:cc_type] ||= card_brand(payment_source_attributes[:number])
end

def payment_source_attributes
    @payment_source_attributes ||= params[:order][:payments_attributes]&.first&.dig(:source_attributes)
end
```

luisramos0:
:+1:

filipefurtad0:
Hi @mkllnk ,

Thank you for your detailed test notes. I went through them step-wise:
- Create a payment method with Pin Payments - OK
- Used the API key saved in Bitwarden - OK
- Set the server to test and check test test mode box - OK
- Open an order cycle with that enterprise - OK
- Check out using a test credit card: https://pinpayments.com/developers/api-reference/test-cards; The error message "That payment method is unsupported. Please choose another one." does not show up - OK. Payments succeeded, where expected.
- Using an unsupported card brand like Diners 30000000056030 triggers the error message: 
image



**Further test cases**

- Verified the security-code length requirements, for different card types:
VISA and Master card require 3 digits
American Express require 4  digirs


- Edge cases on Expiry date:

1) I noticed, that selecting the current month triggers another error other than the "expired card" error. For example:

image

Pressing the Checkout button once more allows a successful payment and checks out normally, leading to the order confirmation.

There is however some sort of "memory-effect": if you trigger an error and then correct it, you may still get the previous error.

2.a) Example: setting valid payment credentials, but setting the expiry date to June triggers:

image

2.b) Correcting June to July renders the same error:

image

I would expect the error shown in case 1). However, the payment succeeds, after retrying, as in 1). So, apparently, the edge case of having an expiry date set to the current month always fails on first attempt, but succeeds on second attempt.


Having this in mind, I tested all the card types and error messages. All are displayed correctly (after cancelling the "memory-effect" pressing Checkout a second time) except Gateway Error, which renders:
Invalid JSON response received from Pin Payments. Please contact support@pin.net.au if you continue to receive this message. (The raw response returned by the API was nil)


**Summary**

- The PR solves the issue - thank you! This is good to go.

- Two minor issues were found:
i) A card with an expiry date set to the current month always fails on first attempt, but succeeds on second attempt.
ii) A memory-effect, in what concerns the triggered errors. Retrying shows the correct error message.
ii) Gateway Errors do not show the correct error message.

Do you thing these should be addressed an deserve opening new issues?

app/services/checkout/form_data_adapter.rb
mkllnk: > if payment is nil this will not return.

Really? I don't understand. `nil&....` returns nil and `nil.blank?` is true.

```rb
nil&.dig(:number).blank?
=> true
```

app/services/checkout/form_data_adapter.rb
```diff
@@ -34,11 +34,16 @@ def move_payment_source_to_payment_attributes!
...
 
-      return if payment&.dig(:number).blank?
+      return if payment_source_attributes.dig(:number).blank?
 
-      payment[:cc_type] ||= card_brand(payment[:number])
+      payment_source_attributes[:cc_type] ||= card_brand(payment_source_attributes[:number])
+    end
+
+    def payment_source_attributes
+      @payment_source_attributes ||=
```
mkllnk, line 50: I tried to avoid ivars here because they are visible throughout the whole class and it's caching the result for longer than needed. It works in this example but you need to double-check potential other uses when refactoring.

I'm happy to leave it as is but I personally try to reduce visibility as much as possible to avoid conflicts and ease refactoring.

mkllnk:
Thanks @filipefurtad0 for the excellent testing!

> i) A card with an expiry date set to the current month always fails on first attempt, but succeeds on second attempt

So that was not due to the memory-effect? I can't imagine how that happens.

> ii) A memory-effect, in what concerns the triggered errors. Retrying shows the correct error message.

I noticed that as well. And then I noticed that a flash came up when loading another page. I'm wondering if messages got queued somehow.

app/services/checkout/form_data_adapter.rb
luisramos0: :-( you are right, I was not aware that the chain would be followed if the conditional navigator & returned nil.
I am sorry about that.
so this method could be:
```
def fill_in_card_type    
     return if payment_source_attributes&.dig(:number).blank?

     payment[:cc_type] ||= card_brand(payment_source_attributes[:number])
end
```
I dont think it's a major difference, otherwise I'd open a PR.

I am sorry I moved this forward and the code is not as you wanted. I did it to get this merged asap. Sorry Maikel.

app/services/checkout/form_data_adapter.rb
luisramos0: Thanks for the feedback!
I see, reducing visibility is important. I thought this improves readability and can also improve performance if we get it re-used in move_payment_source_to_payment_attributes! above.

sauloperez:
> i) A card with an expiry date set to the current month always fails on first attempt, but succeeds on second attempt

while working on bringing in the payment model and the processing logic I noticed somewhere that expiration validation builds a date object by picking the first day of the card's month. This could explain this behaviour. During that month the card will wrongly reported as invalid.

It may make sense to open an issue and investigate a bit more.

sauloperez:
thanks @filipefurtad0 !

app/services/checkout/form_data_adapter.rb
mkllnk: No, no, I'm so happy you took over! We needed to solve this quickly. This bit of code style is not that important compared to the bug we are solving.

luisramos0: :-)

filipefurtad0:
Hi again @mkllnk,

I realize now I missed a relevant test case, which would be to have checked pin payments for orders placed in the backoffice. I have opened an issue in this regard: #5790

mkllnk:
I didn't even try to fix the backoffice because it had been broken for a very long time. See https://github.com/openfoodfoundation/openfoodnetwork/issues/5790#issuecomment-665374249

PR #5770, opened 2020-07-15
Revert "Enhancement/5454 make links to shops under Groups > Producers open in new tab"

luisramos0:
Reverts openfoodfoundation/openfoodnetwork#5601
as discussed in https://github.com/openfoodfoundation/openfoodnetwork/issues/5454#issuecomment-656373253


RachL:
On a group page, producers and shops profiles now are again opening in the same tab.

This is ready to go!

PR #5775, opened 2020-07-16
Make OC advanced settings work by permitting the extra parameter and make the OC edit page work with that option activated

### LINKED ISSUE #5773: Products can't be added to Outgoing O/C (Aus Customer)
## Description

Products can't be added to Order Cycle. When customer clicks on expand arrow to view Products in Outgoing Products tab of Order Cycle, this message appears:

_Loading.._

Expanded product list never loads. Products cannot be added to Order Cycle. 

## Expected Behavior

List of Products available to add to Outgoing should appear. 

## Actual Behaviour

Customer sees message:

_Loading.._

When logged in as Superadmin and viewing this Order Cycle, this error appears:

[missing %{num_of_variants_loaded} value] of 428 Variants Loaded

## Steps to Reproduce

1. Go to any Order Cycle for Baw Baw Food Hub (current order cycle closes 19 July 2020)
2. Go to Outgoing Tab
3. Click on Expand arrow in Products
4. Error (when logged in as Superadmin)
5. Loading... (when logged in as Manager of Enterprise)

## Animated Gif/Screenshot

Screen Shot 2020-07-16 at 11 23 56 AM

Screen Shot 2020-07-16 at 11 36 18 AM

## Workaround
No workaround. Can't add products to Order Cycle (Order Cycle is currently open, but new products need to be added)

luisramos0:
#### What? Why?

Closes #5773 & #5802

Setting the filter in the OC to products from coordinator inventory only was broken due to strong params permissions. That is fixed in this PR.
But the main issue 5773 was when the filter products from coordinator inventory only was active. For some reason in rails 4 the SQL query generation was going wrong and the list of fields in the select part of the SQL included the select from both the visible_for scope in product (which had a distinct in it) and the fields from the included table variants. In this PR I just remove the select from the visible_for scope as it is not needed for this exchange renderer.

#### What should we test?
I can see the problem with the live data on my local machine and I confirm the PR fixes the problem.
For this reason I havent tried to replicate the problem with other data/OC. I did write a auto test that validates the bug so I believe it as simple as activating the "inventory only" option and add some variants to the exchanges.

We should validate the OC page features around adding and removing different types of variants to the exchanges. This combined with activating the "inventory only" setting and validating it works as appropriate: only variants from the coordinators inventory should show in the page to be added to the OC.

#### Release notes
Changelog Category: Fixed
Fixed a bug in the Order Cycles edit page when the option "coordinator inventory only" was activated.

### DIFF
--- a/app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
+++ b/app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
@@ -30,7 +30,9 @@ class Api::Admin::ForOrderCycle::EnterpriseSerializer < ActiveModel::Serializer
   def products_scope
     products_relation = object.supplied_products
     if order_cycle.prefers_product_selection_from_coordinator_inventory_only?
-      products_relation = products_relation.visible_for(order_cycle.coordinator)
+      products_relation = products_relation.
+        visible_for(order_cycle.coordinator).
+        select('DISTINCT spree_products.*')
     end
     products_relation.order(:name)
   end
--- a/app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
+++ b/app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
@@ -34,7 +34,7 @@ class Api::Admin::ForOrderCycle::EnterpriseSerializer < ActiveModel::Serializer
         visible_for(order_cycle.coordinator).
         select('DISTINCT spree_products.*')
     end
-    products_relation.order(:name)
+    products_relation
   end
 
   def products
--- a/spec/services/exchange_products_renderer_spec.rb
+++ b/spec/services/exchange_products_renderer_spec.rb
@@ -7,8 +7,9 @@ describe ExchangeProductsRenderer do
 
   describe "#exchange_products" do
     describe "for an incoming exchange" do
+      let(:exchange) { order_cycle.exchanges.incoming.first }
+
       it "loads products" do
-        exchange = order_cycle.exchanges.incoming.first
         products = renderer.exchange_products(true, exchange.sender)
 
         expect(products.first.supplier.name).to eq exchange.variants.first.product.supplier.name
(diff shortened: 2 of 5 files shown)


app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
```diff
@@ -32,7 +32,6 @@ def products_scope
     if order_cycle.prefers_product_selection_from_coordinator_inventory_only?
       products_relation = products_relation.visible_for(order_cycle.coordinator)
     end
-    products_relation.order(:name)
```
mbudm, line 35: Am attempting to get the tests done to keep this one moving - this line needs to remain AFAICT.

luisramos0: Hi Steve, yes, I thought the implicit return would work but it doesnt. I added it again now. But removed the order(), it's not needed here.

spec/services/exchange_products_renderer_spec.rb
```diff
@@ -7,23 +7,44 @@
...
+        end
+
+        it "loads products from the coordinator inventory" do
+          # Add variant already in the exchange to the coordinator's inventory
+          exchange.variants.first.inventory_items = [create(:inventory_item, enterprise: order_cycle.coordinator)]
+
+          products = renderer.exchange_products(false, exchange.receiver)
+
+          expect(products).to eq [exchange.variants.first.product]
+        end
```
luisramos0, line 46: this test blows up in master just like #5773

mbudm:
I missed that this fixes the steps to create as well and ended up creating #5802 for that bug. So that gives you a 2 issue PR @luisramos0 if you want it.

luisramos0:
ah, yes, that's fixed here as well (it was a very common rails 4 upgrade problem: incomplete list of permitted "strong" params). Thanks Steve.

filipefurtad0:
Hi @luisramos0 ,

Thank you for your help reproducing this issue!

Setting the an OC to "Coordinators Inventory only" breaks the Outgoing Products page. Staging this PR solves this issue. Several tests were done, in this regard, please see my notes in the release-draft testing [here](https://github.com/openfoodfoundation/openfoodnetwork/issues/5805#issuecomment-662694326).

I think this PR could be merged and the patch deployed, and two separate issues open. Awaiting feedback on this.

luisramos0:
I left a comment on the release issue. Let's merge this and create new issues if needed :+1:

Issue #5449, opened 2020-05-18
Partial refund not available (refund amount is greater than unrefunded amount) - Initial/successful partial refund in Stripe not registered in OFN


RachL:
## Description

A big user in France is having this problem on two orders. This user don't want us to try the refund through Stripe directly. I'm a bit stuck on what to do next.

When you try to refund you get a message saying the amount you are trying to refund is greater than the unrefunded amount. Yet the unrefunded amount is a figure that is not known...

Hub is number 922 and they are not on Stripe SCA yet.

## Steps to Reproduce
1. On FR prod orders R287174615 and R288734641 are in credit
2. When hitting the credit button this message is displayed:

![image.png](https://images.zenhubusercontent.com/595e669a61b4ca3b0a737135/cee824a3-51d4-4750-bfcd-04463cf7bcda)

3. Yet both orders seem fine: the amount to refund is lower than what the buyer paid:

![image.png](https://images.zenhubusercontent.com/595e669a61b4ca3b0a737135/b5eba61a-f928-4290-a81e-63054805e95a)

## Workaround
I have no idea if trying on stripe works, they don't want me to test it....

## Severity

Not sure what the severity is....

## Your Environment
* Version used: v2.9.8
* Browser name and version: Firefox
* Operating System and version (desktop or mobile):

kristinalim:
Here are some details for R288734641:

* order `item_total` - 56.66
* order `total` - 48.8
* order `adjustment_total` - -7.86 (consistent with adjustments amount)
* order adjustments `included_tax` - -0.45
* line item adjustments `amount` - 2.95
* line item adjustments `included_tax` - 2.95
* line items `price` sum - 49.89
* payments `amount` sum - 57.51

kristinalim:
@RachL It looks like the Stripe refund succeeded but there was a problem in the OFN side. The Stripe charge has these data, and the difference explains the 5.25 amount:

```
  "amount": 5751,
  "amount_refunded": 5226,
```

The refund consisting of 6 refunds worth 8.71, which was the amount attempted to refund.

kristinalim:
The 8.71 amount comes from payment made minus the adjusted order total.

RachL:
😱 sooo... these people now owe the hub money? Thank you so much for looking into it @kristinalim

kristinalim:
>  sooo... these people now owe the hub money? Thank you so much for looking into it @kristinalim

[...]

Issue #5580, opened 2020-06-10
Backoffice orders with transaction fees


filipefurtad0:
This epic gathers issues related to backoffice placement, refund, adjustment (others?) of orders with transaction fees.

Issue #5574, opened 2020-06-09
[Admin][Order] "Balance due" state after payments with a transaction fee


filipefurtad0:
## Description

The price shown in backoffice orders does not include transaction fees, for the respective payment methods. After payment is made, there is always an amount still to be charged and the order is in a "balance due" state. Attempting to pay for the open amount leads do an additional charge of the fee.

## Expected Behavior

The transaction fee should be included in the total price: it should be updated according to the choices made on Payments section allowing to pay the full price (fee included).

## Actual Behaviour

The transaction fee is not included. Paying for orders with transaction fee always renders the order to be in a  "balance due" state.

## Steps to Reproduce

Admin:
1. Set up a payment method with a fee.
2. Start placing an order in the backoffice, inserting items and customer details.
3. Proceed to Payments and check the total amount to be paid.
4. Choose the previously set payment method and perform the payment. 
5. Check the state of the order: it should be in "Balance due" with an amount corresponding to the value of the transaction fee.
6. Attempt to pay this open amount with that payment method - an additional fee will be charged...

## Animated Gif/Screenshot

Before payment:
![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/3a1c041b-5dfe-4e0f-a1ac-5cf55145be88)

After payment:
![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/e9cf2c8f-192a-4bed-bb4b-2394b02bb1b7)

## Workaround

If you know, in advance, what the value of the transaction fee is, then you can add this value manually, as a "prediction" of what the total cost will be. This allows to pay for the total amount with one payment.

## Severity

bug-s4: it's annoying, but you can use it

## Your Environment

* Version used: v2.10.0
* Browser name and version: Firefox 75
* Operating System and version (desktop or mobile): Desktop Ubuntu 18.04

RachL:
@filipefurtad0 I wonder if this isn't a duplicate of: https://github.com/openfoodfoundation/openfoodnetwork/issues/3490

filipefurtad0:
Hey @RachL ,

I did not see #3490 as a duplicate as it deals with refunds, but it's surely related. Where I really stopped to wonder was this one:
- Payment Fee adjustment after order completion #1168

[...]

Issue #3490, opened 2019-02-14
When crediting an order the transaction fee is applied when it shouldn't


sstead:
## Expected Behavior
I have an order which was paid in full with Stripe. Then I removed an item from the order, putting the order in Credit. I tried to refund this credit using the Credit button. I found that when the credit was applied, the transaction fee that the shop has on Stripe payments was applied to the credit. 

I would not expect a payment method transaction fee to be applied to a credit event.

## Actual Behaviour
The credit event incurred the transaction fee.

## Steps to Reproduce
1. Place an order in a shop and pay with Stripe- where stripe has a payment method fee associated (e.g $5)
2. Go to Orders in the backend and take one item from the order, putting it into credit
3. Go to edit the order and go to payments. Click the credit button.
4. See that the credit was applied but the transaction fee was also applied.

## Animated Gif/Screenshot
![image](https://user-images.githubusercontent.com/10843056/52763133-aa1bdf80-306e-11e9-9153-832326290fd4.png)

## Workaround

The new workaround for this bug is to stop doing partial refunds through OFN or if the partial refund is done through OFN do a negative adjustments and refund through stripe :face_with_head_bandage: 

## Context
I'm trying to do a partial refund from the backend.

## Severity
bug-s3: a feature is broken but there is a workaround

## Your Environment
Product, Aus, Chrome, PC, Desktop

## Possible Fix
Credit event should not carry fee.

sauloperez:
This one is looking at us @luisramos0 :trollface: I wish we could wait until fixing it in v2 but I'm afraid it'll still take some time...

luisramos0:
well, not really, I think we have to ask:
- is this a fundamental feature in ofn and how much is it used today? most probably it's not used and it has been broken for a while
- will this fix in master work in v2 and thus, there will be no rework when merging to v2? I think the answer is no, we would have to fix in master and then think again about it in v2 with all the changes in adjustments/shipments
So, I'd only fix this one only in v2 if no one disagrees.

sauloperez:
That's exactly what I meant. We need to know if it can wait for a single v2-fix or not

kirstenalarsen:
@daniellemoorhead just clearing issues and found this - think is related to your local produce ladies question?

kirstenalarsen:
not sure if we actually need to act on this or not

daniellemoorhead:
> @daniellemoorhead just clearing issues and found this - think is related to your local produce ladies question?

[...]

PR #5780, opened 2020-07-17
Handle credit validation errors

sauloperez:
#### What? Why?

Closes https://github.com/openfoodfoundation/openfoodnetwork/issues/5449

Note this is based on #5806.

I'm not particularly happy about having to come down to the `attr_accessor` trick to skip a callback but this prevents this fix from growing exponentially on LOCs and avoids the risk of making the wrong assumptions. What I like about it is that `skip_source_validation` is explicit enough to pave the way for extracting specific per use-case service objects in the future.

#### What should we test?

We need to cover all payment scenarios:

* Issue a partial credit with each supported payment gateway
* Issue a void operation with each supported payment gateway
* Issue a capture operation with each supported payment gateway
* Place a regular order as a consumer
* Place an order as an admin

When performing these actions as admin, the payment total should be accurate, taking into account any payment fees.

#### Release notes

Handle credit and other payment actions whose credit card is now expired. Stripe kept track of them and we weren't
Changelog Category: Fixed

### DIFF
--- a/app/models/spree/payment.rb
+++ b/app/models/spree/payment.rb
@@ -187,7 +187,7 @@ module Spree
     # Makes newly entered payments invalidate previously entered payments so the most recent payment
     # is used by the gateway.
     def invalidate_old_payments
-      order.payments.with_state('checkout').where("id != ?", id).each do |payment|
+      order.payments.with_state('checkout').where.not(id: id).each do |payment|
         # Using update_column skips validations and so it skips validate_source. As we are just
         # invalidating past payments here, we don't want to validate all of them at this stage.
         payment.update_column(:state, 'invalid')
--- a/app/models/spree/payment.rb
+++ b/app/models/spree/payment.rb
@@ -188,7 +188,6 @@ module Spree
     # is used by the gateway.
     def invalidate_old_payments
       order.payments.with_state('checkout').where("id != ?", id).each do |payment|
-
         # Using update_column skips validations and so it skips validate_source. As we are just
         # invalidating past payments here, we don't want to validate all of them at this stage.
         payment.update_column(:state, 'invalid')
--- a/app/models/spree/payment.rb
+++ b/app/models/spree/payment.rb
@@ -22,7 +22,7 @@ module Spree
 
     has_one :adjustment, as: :source, dependent: :destroy
 
-    before_validation :validate_source
+    validate :validate_source
     before_save :set_unique_identifier
 
     after_save :create_payment_profile, if: :profiles_supported?
--- a/app/models/spree/payment.rb
+++ b/app/models/spree/payment.rb
@@ -191,6 +191,7 @@ module Spree
         # Using update_column skips validations and so it skips validate_source. As we are just
         # invalidating past payments here, we don't want to validate all of them at this stage.
         payment.update_column(:state, 'invalid')
+        payment.ensure_correct_adjustment
       end
     end
 
--- a/app/models/spree/payment.rb
+++ b/app/models/spree/payment.rb
@@ -32,16 +32,12 @@ module Spree
     # invalidate previously entered payments
     after_create :invalidate_old_payments
 
-    # Prevents #validate_source to end up in a chain of validations, from
-    # payment to payment up until the original credit card record.
+    # Skips the validation of the source (for example, credit card) of the payment.
     #
-    # All payments are validated before saving thus, we get nothing by checking
-    # they are valid in the future. Quite the opposite. Stripe is accepting
-    # refunds whose source are cards that were valid when the payment was
-    # placed but are now expired, and we consider them invalid.
-    #
-    # Skipping the source validation when applying a refund avoids this
-    # situation. We trust the payment gateway.
+    # This is used in refunds as the validation of the card can fail but the refund can go through,
+    #    we trust the payment gateway in these cases. For example, Stripe is accepting refunds with
+    #    source cards that were valid when the payment was placed but are now expired, and we
+    #    consider them invalid.
     attr_accessor :skip_source_validation
     attr_accessor :source_attributes
 
@@ -188,10 +184,13 @@ module Spree
       gateway_error e
     end
 
-    # Inherited from Spree, makes newly entered payments invalidate previously
-    # entered payments so the most recent payment is used by the gateway.
+    # Makes newly entered payments invalidate previously entered payments so the most recent payment
+    # is used by the gateway.
     def invalidate_old_payments
       order.payments.with_state('checkout').where("id != ?", id).each do |payment|
+
+        # Using update_column skips validations and so it skips validate_source. As we are just
+        # invalidating past payments here, we don't want to validate all of them at this stage.
         payment.update_column(:state, 'invalid')
       end
     end
--- a/app/models/spree/payment.rb
+++ b/app/models/spree/payment.rb
@@ -32,7 +32,19 @@ module Spree
     # invalidate previously entered payments
     after_create :invalidate_old_payments
 
+    # Prevents #validate_source to end up in a chain of validations, from
+    # payment to payment up until the original credit card record.
+    #
+    # All payments are validated before saving thus, we get nothing by checking
+    # they are valid in the future. Quite the opposite. Stripe is accepting
+    # refunds whose source are cards that were valid when the payment was
+    # placed but are now expired, and we consider them invalid.
+    #
+    # Skipping the source validation when applying a refund avoids this
+    # situation. We trust the payment gateway.
+    attr_accessor :skip_source_validation
     attr_accessor :source_attributes
+
     after_initialize :build_source
 
     scope :from_credit_card, -> { where(source_type: 'Spree::CreditCard') }
@@ -151,7 +163,7 @@ module Spree
     end
 
     def validate_source
-      if source && !source.valid?
+      if source && !skip_source_validation && !source.valid?
         source.errors.each do |field, error|
           field_name = I18n.t("activerecord.attributes.#{source.class.to_s.underscore}.#{field}")
           errors.add(Spree.t(source.class.to_s.demodulize.underscore), "#{field_name} #{error}")
@@ -176,8 +188,12 @@ module Spree
       gateway_error e
     end
 
+    # Inherited from Spree, makes newly entered payments invalidate previously
+    # entered payments so the most recent payment is used by the gateway.
     def invalidate_old_payments
-      order.payments.with_state('checkout').where("id != ?", id).each(&:invalidate!)
+      order.payments.with_state('checkout').where("id != ?", id).each do |payment|
+        payment.update_column(:state, 'invalid')
+      end
     end
 
     def update_order
--- a/spec/models/spree/payment_spec.rb
+++ b/spec/models/spree/payment_spec.rb
@@ -725,7 +725,7 @@ describe Spree::Payment do
       end
     end
 
-    describe "refunding" do
+    describe "refund!" do
       let(:payment) { create(:payment) }
       let(:success) { double(success?: true, authorization: 'abc123') }
       let(:failure) { double(success?: false) }
--- a/spec/models/spree/payment_spec.rb
+++ b/spec/models/spree/payment_spec.rb
@@ -391,13 +391,13 @@ describe Spree::Payment do
 
         context "when response is successful" do
           it "should create an offsetting payment" do
-            Spree::Payment.should_receive(:create!)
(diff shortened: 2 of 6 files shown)


app/models/spree/payment.rb
```diff
@@ -160,6 +160,14 @@ def validate_source
       errors.blank?
     end
 
+    def valid_source?
+      if source.is_a?(Spree::Payment)
+        true
+      else # top-most source
```
luisramos0, line 166: this is very hard for me to understand even if I read the commit message and if know the context inside out...
what if we extract this little method to a little service and add a nice comment to explain?

spec/models/spree/payment_spec.rb
```diff
@@ -1,7 +1,691 @@
...
+        context "when gateway does not match the environment" do
+          it "should raise an exception" do
+            gateway.stub :environment => "foo"
+            expect { payment.credit! }.to raise_error(Spree::Core::GatewayError)
+          end
+        end
+
+        context "when response is successful" do
+          it "should create an offsetting payment" do
+            Spree::Payment.should_receive(:create!)
```
luisramos0, line 390: this syntax can be modernized right? expect .... to receive.

sauloperez: yeah, it should. Good catch.

app/models/spree/payment.rb
sauloperez: I think I can manage with a good method doc. This won't be pretty as we are approaching callbacks nastiness but bear with me.

spec/models/spree/payment_spec.rb
```diff
@@ -401,6 +401,25 @@
...
+          context 'and the source payment card is expired' do
+            let(:card) do
+              Spree::CreditCard.new(month: 12, year: 1995, number: '4111111111111111')
+            end
+
+            let(:successful_response) do
+              ActiveMerchant::Billing::Response.new(true, "Yay!")
+            end
+
+            it 'lets the new payment to be saved' do
```
luisramos0, line 414: I think the correct English is "it 'lets the new payment be saved'" or ""it 'allows the new payment to be saved'""

app/models/spree/payment.rb
```diff
@@ -151,7 +163,7 @@ def revoke_adjustment_eligibility
     end
 
     def validate_source
-      if source && !source.valid?
+      if source && !skip_source_validation && !source.valid?
```
luisramos0, line 166: I know you only added a variable here :-)
I wonder if this is not easier and equal if written like:
```ruby
def validate_source
   return true if !source || skip_source_validation || source.valid?

   source.errors.each do |field, error|
     field_name = I18n.t("activerecord.attributes.#{source.class.to_s.underscore}.#{field}")
     errors.add(Spree.t(source.class.to_s.demodulize.underscore), "#{field_name} #{error}")
    end
    false
end

```

app/models/spree/payment.rb
```diff
@@ -176,8 +188,12 @@ def create_payment_profile
       gateway_error e
     end
 
+    # Inherited from Spree, makes newly entered payments invalidate previously
```
luisramos0, line 191: I dont think we need to tell our story here. I'd remove "Inherited from Spree,".

app/models/spree/payment.rb
```diff
@@ -32,7 +32,19 @@ class Payment < ActiveRecord::Base
     # invalidate previously entered payments
     after_create :invalidate_old_payments
 
+    # Prevents #validate_source to end up in a chain of validations, from
```
luisramos0, line 35: I found this comment quite complicated. Let me try:

First I'd describe what the attribute does and then say where it is used and why:

```ruby
# Skips the validation of the source (for example, credit card) of the payment.
# 
# This is used in refunds as the validation of the card can fail but the refund can go through, we trust the payment 
#    gateway in these cases. For example, Stripe is accepting refunds with source cards that were valid when
#    the payment was placed but are now expired, and we consider them invalid.

```
I think that's it here.

I dont think the chain of validations is related to this refund/skip_source_validation, is it? I think it's related to the invalidation of previous payments that you have wisely changed below from invalidate! to update_column :+1: 
There (line 191), I'd add this comment:
```ruby
# Using update_column skips validations and so it skips validate_source. As we are just invalidating
#     past payments here, we don't want to validate all of them at this stage.
```

app/models/spree/payment.rb
sauloperez: ok, I gave it a try and then I thought. Wait and see if someone mentions it :joy:

app/models/spree/payment.rb
sauloperez: Thanks, that looks better. 

> I dont think the chain of validations is related to this refund/skip_source_validation, is it?

It was happening due to both. Every source was validating its source up until the credit card.

app/models/spree/payment.rb
sauloperez: ok. What's important is the rest.

sauloperez:
@luisramos0 I addressed your comments plus a commit that handles `#refund!`.

app/models/spree/payment.rb
```diff
@@ -176,8 +184,14 @@ def create_payment_profile
       gateway_error e
     end
 
+    # Makes newly entered payments invalidate previously entered payments so the most recent payment
+    # is used by the gateway.
     def invalidate_old_payments
-      order.payments.with_state('checkout').where("id != ?", id).each(&:invalidate!)
+      order.payments.with_state('checkout').where("id != ?", id).each do |payment|
```
Matt-Yorkley, line 190: I think we can start using `#not` in Rails 4 for these "where not equals" AR scopes. Not blocking, just mentioning it...

sauloperez: wooow! I totally forgot Rails 4.0 introduced [#not](https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/WhereChain/not)!

sauloperez:
As expected, chaning from `#invalidate!` to `#update_column` doesn't play with the order total calculation in the build :see_no_evil:

sauloperez:
I thought this would take days but I think I managed to solve it. I also addressed your comment @Matt-Yorkley . ProTip: You need a where because comes out of an association is a ` #ActiveRecord::AssociationRelation` and not a `WhereChain`. I guess the latter substitutes the `ActiveRecord::Relation` in Rails 4.

filipefurtad0:
Hi @sauloperez and @luisramos0,

The PR introduces an error 500 on /admin/enterprises/

Please see:
https://openfoodnetwork.slack.com/archives/CEF14NU3V/p1595523810421200

sauloperez:
Got it. The migration  https://github.com/openfoodfoundation/openfoodnetwork/blob/master/db/migrate/20200616162646_move_all_calculators_outside_the_spree_namespace.rb was applied in that DB but this branch's code still has the calculators within the Spree namespace.

sauloperez:
I forced push to the get the latest changes of https://github.com/openfoodfoundation/openfoodnetwork/pull/5806.

sauloperez:
@filipefurtad0 it's solved now.

filipefurtad0:
Hey @sauloperez ,

I've reversed the order, and numbered the test cases:

i) Place regular orders, as a consumer
ii) Place orders as admin
iii) Issue a partial credit with each supported payment gateway
iv) Issue a void operation with each supported payment gateway
v) Issue a capture operation with each supported payment gateway

There is quite a lot going on here, so let's go, vaya vayita! :-)


Set different enterprise fees, shipping fees, transaction fees. The shop charges GST. Placed 6 orders, with different payment methods in the shopfront. All payment methods have the same transaction fee.

**i) Placing 6 orders, with these payment methods:**

1) Paypal
2) Cash
3) Stripe-SCA (paid with 3D card)
4) Stripe-SCA (paid with regular)
5) Stripe-Connect
6) Pin payments

The value of the orders should be and is the same, as checked on confirmation e-mails and under /orders.


**ii) Placing the same 6 orders, on the backoffice:**

1) Paypal - No Payment Via Admin Backend
2) Cash - OK
3) Stripe-SCA (paid with 3D card) - not supported in the Backoffice
4) Stripe-SCA (paid with regular)  - OK
5) Stripe-Connect - OK
6) Pin payments – not supported in the back office - #5790

Issue “Backoffice orders: "Balance due" state after payments with a transaction fee #5574” was observed; Used the described workaround and obtained the same values as those obtained in shopfront orders.


**iii) Issue a partial credit with each supported payment gateway**

Removed 5 items and recalculated fees. Proceeded to Payments:

1) Paypal - issuing credit is not supported

image

2) Cash - issuing credit is not supported

image

3) Stripe-SCA (paid with 3D card) - Issuing the credit works. Known issue [#3490](When crediting an order the transaction fee is applied when it shouldn't)

image

image

image


Refund appears correctly in Stripe Dashboard:

image


4) Stripe-SCA (paid with regular) - same as 3)
5) Stripe-Connect - same as 3)
6) Pin payments – issuing is not supported:

image




**iv) Issue a void operation with each supported payment gateway:**

1) Paypal - No Payment Via Admin Backend - not supported.
2) Cash - "Voiding payment does not remove transaction fee #3584"
3) Stripe-SCA (paid with 3D card) - If one tries to void the payments on the previously edited order, one gets the errors:

- voiding the refund:

image

- voiding the initial payment:

image


- Voiding a Stripe payment from an order which was not edited works fine.


4) Stripe-SCA (paid with regular) - same as 3)

5) Stripe-Connect - Voiding the initial payment from the order works. Attempting to void the refund, after refunding the whole payment, results in this error:

image


On Stripe dashboard it can be seen, that a partially refunded order goes from this state:

image


Into:

image


6) Voiding Pin payments – is not supported in the backoffice:

image



**v) Issue a capture operation with each supported payment gateway**

I am not 100% sure about this operation @sauloperez . Do you mean this (below) the tick, under /orders? I thought this was only possible for cash payments...

image

2) Cash - Capturing payment works


So in summary:

- I think there is still an issue with refunds on Stripe-SCA. It should be possible to refund the initial amount, without getting the error. This is perhaps worth opening a new issue.

- for Stripe Connect (for which the issue was opened) all looks good.

- The PR touches several previously known issues. See epic [5580](https://github.com/openfoodfoundation/openfoodnetwork/issues/5580). 

I would say this is ready to go, but I'm especially grateful for feedback here. This is quite extensive and touches many corners. Moving to ready to go, but adding the feedback label. Thank you!

luisramos0:
Hey @filipefurtad0 thanks a lot for the testing.
The testing looks good. Re the StripeSCA, I think the refund worked well:
image

I think you tried to void the refund and then void the original payment of an order with a partial refund. I think it's good it failed. After you issue a partial refund, foer example, payed 100 and refunded 50, none of these two movements should be voided. I think all the user can do after this is to issue another partial refund.

I think what also needs to be tested here are the checkout retrial scenarios. I will have a look at #5816 and #5818 so we can continue the conversation there.

So, I will merge this PR now and look at #5816 and #5818 with and without this PR to see if anything has changed :+1: 
Thanks Filipe!

filipefurtad0:
Hey @luisramos0,

Yes, the partial refund works well for both Stripe methods. The subsequent full refund fails for Stripe-SCA (test cases iv) 3 and 4) but works nicely for Stripe-Connect (see test case iv) 5).

If not allowing for a full refund after a partial one is the expected behavior than all is good think, as eventually Stripe-Connect will be discontinued.

I checked anyway for this difference between the two payment methods and it is not introduced by this PR and it does not seem to relate to transaction fees.

luisramos0:
oh, nice finding. I didnt know that would work in StripeConnect. That's basically lost funcionality from StripeConnect to StripeSCA. We should create a specific issue for it and then decide what to do.

The functitonality is "full refund after partial refund", and I think you mean the user presses VOID on the original payment, right?
I wonder if VOID on the partial refund also works on StripeConnect. Voiding a partial refund should work in StripeSCA, I think that could another separate issue to be looked at.
These are straight forward things to do in stripeSCA, when we have the issues, we can decide if we do them now or later.

filipefurtad0:
Hi @luisramos0 ,

> I wonder if VOID on the partial refund also works on StripeConnect. Voiding a partial refund should work in StripeSCA, I think
> that could be another separate issue to be looked at.

Voiding partial refunds do not work in neither Stripe-Connect or -SCA: opened issue #5830

luisramos0:
Hello Filipe! Thanks for confirming that :+1:

sauloperez:
Thanks a lot for this in-depth testing @filipefurtad0 . We are now on a much better position to fix these things :muscle:

PR #5792, opened 2020-07-20
Missing translation key for description, display, tags

romale:
In admin/shipping_methods/new

#### What? Why?

When creating a new shipping method in path "admin/shipping_methods/new" some keys not translated

image

image

image

#### Release notes


romale:
yes, it may well not be a bug. I'm just still on version v2.10.1. I translate the user guide and where errors occur (in my environment), I try to fix it :) sorry

luisramos0:
ok, no problem. We are not currently fixing bugs or releasing in v2 any more so I will close the PR.

But please do submit PRs like this one if you find other problems that may well be in v3 :+1:

PR #5796, opened 2020-07-21
Remove conflicting and duplicate route

### LINKED ISSUE #5771: NoMethodError in spree/user_sessions#new
## Error in OpenFoodNetwork

3 users affected since v3 deployment in AUS, I think it's related to signup/email confirmation workflow gone wrong... needs investigation.

**NoMethodError** in **spree/user_sessions#new**
undefined method `users_url' for #<Spree::UserSessionsController:0x0000555618a3a4d0>
Did you mean?  user_url

[View on Bugsnag](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5f0e4fc053724e0017dc4b89?event_id=5f0ed5dc005b7247dc510000&i=gh&m=ci)

## Stacktrace

    app/controllers/application_controller.rb:25 - redirect_to
    lib/spree/core/controller_helpers/respond_with_decorator.rb:23 - respond_with

[View full stacktrace](https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5f0e4fc053724e0017dc4b89?event_id=5f0ed5dc005b7247dc510000&i=gh&m=ci)

sauloperez:
## What? Why?

Closes #5771

This Spree route conflicts with the one we define:

```
get "/login", to: redirect("/#/login")
```

for whatever reason there are 7 users that managed to hit the Spree one instead of ours when confirming their signup email. It's not clear to me though when this `/login?validation=confirmed` is really hit. The confirmation email link passes a token in the query params and this is not the case.

The idea is that `GET /login` makes the login modal to show up instead of Devise's default behavior (through inheritance) of showing a login form page. OFN was never prepared to handle this as this bug proofs.

#### What should we test?

Confirming a signup email and login should work as usual. I'd try both scenarios.

#### Release notes

Remove duplicate and conflicting Spree route to the login page. We don't have any.
Changelog Category: Fixed

### DIFF
--- a/app/controllers/user_passwords_controller.rb
+++ b/app/controllers/user_passwords_controller.rb
@@ -10,7 +10,7 @@ class UserPasswordsController < Spree::UserPasswordsController
 
     if resource.errors.empty?
       set_flash_message(:success, :send_instructions) if is_navigational_format?
-      respond_with resource, location: spree.login_path
+      respond_with resource, location: main_app.login_path
     else
       respond_to do |format|
         format.html do
--- a/app/controllers/spree/users_controller.rb
+++ b/app/controllers/spree/users_controller.rb
@@ -59,7 +59,7 @@ module Spree
       if @user
         authorize! params[:action].to_sym, @user
       else
-        redirect_to spree.login_path
+        redirect_to main_app.login_path
       end
     end
 
--- a/config/routes/spree.rb
+++ b/config/routes/spree.rb
@@ -15,7 +15,6 @@ Spree::Core::Engine.routes.draw do
   resources :users, :only => [:edit, :update]
 
   devise_scope :spree_user do
-    get '/login' => 'user_sessions#new', :as => :login
     post '/login' => 'user_sessions#create', :as => :create_new_session
     get '/logout' => 'user_sessions#destroy', :as => :logout
     get '/signup' => 'user_registrations#new', :as => :signup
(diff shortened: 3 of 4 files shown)


app/controllers/user_passwords_controller.rb
```diff
@@ -10,7 +10,7 @@ def create
 
     if resource.errors.empty?
       set_flash_message(:success, :send_instructions) if is_navigational_format?
-      respond_with resource, location: spree.login_path
```
luisramos0, line 13: I tried to understand where this login_path takes us now and where the modal is loaded but I couldnt find it...

sauloperez: `login_path` hits `get "/login", to: redirect("/#/login")` and in turn `/#/login` gets handled by the HomeController. Finally, the `#/login` is what the frontend checks to load the modal. I haven't checked where that final step happens but it's a good-old frontend route.

sauloperez:
fixed!

filipefurtad0:
Hey @sauloperez ,

I could not reproduce the issue.
Manually changing the URL from `/login` always redirects to `/#/login`, before and after this PR. Am I missing something?

In any case, tested sign-up, log-in, password recovery functions. All good, both before and after the PR.

I'd say this is ready to go - it would be great to reproduce the bug though.

sauloperez:
Yes, sorry @filipefurtad0 I should have mentioned. This is a very particular issue that only a bunch of users are experiencing and I'm not sure what's causing. What I did was just remove the chance of this happening again. Without duplicate routes there is no way it can happen.

PR #5799, opened 2020-07-21
Add migration to fix calculator preferences

### LINKED ISSUE #5795: Shipping Method and Transaction Fees removed from Enterprise
## Description

Enterprises appear to have had Shipping Method and Payment Method fees removed from their account (noted on 21st July). 

**Specific Example**

Prom Coast Food Collective (2552) had an order cycle that closed on 20th July at 4pm.  There are four orders that were placed between 2.38pm and 3.50pm on 20th July that have a $15 credit owing equal to the shipping method fee for the selected shipping method. 

R421582413
R221736220
R046212123
R752741780

NOTE: it appears that the $15 'credit owing' status was being updated continually in OFN from some time on 21st July. Investigations earlier in the morning only identified 2 and 3 impacted orders, but by 12pm there were 3 impacted orders. 

Prom Coast provided details of Order number R421582413 as follows:
- A customer has placed an order and paid by Stripe. 
- The Stripe payment includes a $15 fee. 
- The order total does not reference this and specifies a $0 shipping method fee
- The Shipping Method that was selected is supposed to have had a $15 fee charged
- Logging in using Admin shows no Shipping Method or Payment Method fees currently for this Enterprise
- There should be fees for this Enterprise as follows ($10 or $15 for shipping methods, Admin Fee on Each Order, 1.75% Fee on Stripe Payments)
- These fees were charged for other customers in this order cycle

**Other Enterprises also impacted ** 
These Enterprises also appear to be missing Fees (to be updated here as more come to light):
Fat Beets Food Hub (2070)
New Life Farm (2026)
North Arm Farm (2331) - Confirmed $1 transaction fee missing

## Steps to Reproduce

- Review Prom Coast Food Collective Orders before and after July 20th/21st
- See Order numbers above with Credit owing
- See that Prom Coast Food Collective Fees are now all set at zero

## Animated Gif/Screenshot

# Workaround

## Severity
S2

-->

## Your Environment

* Version used:
* Browser name and version:

luisramos0:
#### What? Why?

Closes #5795

This will change all adjustments config in spree_preferences to the new calculators without the Spree namespace.

#### What should we test?
Before staging this PR you should replicate the problem, for example, in FR staging now in fredo's farm hub shop the 10eur shipping fee are not applied:
https://staging.openfoodfrance.org/fredo-s-farm-hub/shop
I am pointing this one out because not all fees are broken (all the fees that have been edited manually since v3 release will be working correctly).
You need to test this **with existing fees** on the server, you cant edit any fees otherwise the problem is gone.
After this PR is staged the fees should be applied again.

#### Release notes
Changelog Category: Fixed
Fixed a data migration bug in a recent change to fees calculators where fees were not applied at checkout.

### DIFF
--- a/db/migrate/20200721135726_move_calculators_preferences_outside_spree_namespace.rb
+++ b/db/migrate/20200721135726_move_calculators_preferences_outside_spree_namespace.rb
@@ -1,0 +1,23 @@
+# As we moved the calculators outside the Spree namespace in migration MoveAllCalculatorsOutsideTheSpreeNamespace
+#   We need to move their preferences too (currency, value, etc), otherwise they are not used
+class MoveCalculatorsPreferencesOutsideSpreeNamespace < ActiveRecord::Migration
+  def up
+    replace_preferences_key("/spree/calculator", "/calculator")
+  end
+
+  def down
+    replace_preferences_key("/calculator", "/spree/calculator")
+  end
+
+  private
+
+  def replace_preferences_key(from_pattern, to_pattern)
+    updated_pref_key = "replace( pref.key, '" + from_pattern + "', '" + to_pattern + "')"
+    Spree::Preference.connection.execute(
+      "UPDATE spree_preferences pref SET key = " + updated_pref_key + "
+        WHERE pref.key like '" + from_pattern + "%'
+          AND NOT EXISTS (SELECT 1 FROM spree_preferences existing_pref
+                           WHERE existing_pref.key = " + updated_pref_key + ")"
+    )
+  end
+end
--- a/db/migrate/20200721135726_move_calculators_preferences_outside_spree_namespace.rb
+++ b/db/migrate/20200721135726_move_calculators_preferences_outside_spree_namespace.rb
@@ -19,5 +19,7 @@ class MoveCalculatorsPreferencesOutsideSpreeNamespace < ActiveRecord::Migration
           AND NOT EXISTS (SELECT 1 FROM spree_preferences existing_pref
                            WHERE existing_pref.key = " + updated_pref_key + ")"
     )
+
+    Rails.cache.clear
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20200702112157) do
+ActiveRecord::Schema.define(version: 20200721135726) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"


filipefurtad0:
Hi @luisramos0 ,

I could not reproduce the issue on the Fredo's Farm Hub, in staging-FR. It had a flat percent (10%) on a shipping method (Home delivery), with an invisible tag set. So this shipping method was not appearing in the shopfront. Other pick-up option had zero fees set, and other enterprise fees were applied equally, before and after staging.

I avoided editing these settings explicitly, as editing would probably refresh the DB and show/apply the fees correctly. 

Anyway, all looks good before after staging.

I will leave this PR in test ready for now, and try to find a shop, on another staging server, with shipping fees set, see if I can verify the issue later on.

db/migrate/20200721135726_move_calculators_preferences_outside_spree_namespace.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  private
+
+  def replace_preferences_key(from_pattern, to_pattern)
+    updated_pref_key = "replace( pref.key, '" + from_pattern + "', '" + to_pattern + "')"
+    Spree::Preference.connection.execute(
+      "UPDATE spree_preferences pref SET key = " + updated_pref_key +
+      " WHERE pref.key like '" + from_pattern + "%'" +
+      "   AND NOT EXISTS (SELECT 1 FROM spree_preferences existing_pref
+                          WHERE existing_pref.key = " + updated_pref_key + ")"
```
Matt-Yorkley, line 20: There's blank spaces either side of the string in these keys?

filipefurtad0:
A quick update:

I tried to reproduce the issue as well on an old "untouched" hub in staging-katuma, but still no success: the fees appear in the order total, and are charged as expected.

So I did not notice any hub whose settings were set to zero... This is similar to the issue we observed to Katuma, in production - right? I can't recall having seen any setting misplaced in staging servers, other than missing images. Can we pin-point such a case, in staging?

After review comments are addressed I'd move this to ready to go.

db/migrate/20200721135726_move_calculators_preferences_outside_spree_namespace.rb
luisramos0: why do you ask? no blank spaces, just more characters like the name of the fee (flat_rate for example), that's why we need the %.

luisramos0: maybe you are looking at updated_pref_key and thinking it's a string? updated_pref_key is a call to the replace function and so it can have spaces before and after.

luisramos0:
Hey Filipe, appologies, trying to replicate it myself I just realize it is impossible for your replicate without looking at the DB!
So, in katuma DB I see:
image

and this means that this method should not be zero but 2 eur:
image

and so when I checkout in that shop home delivery should be 2eur not Free:
image

In the same way (so I test this for a payment method as well) because I see this in the DB:
image

I know this payment method should have a flat rate of 1eur and not zero:
image
and on checkout stripe should cost 1eur not be free:

image


I am now going to deploy this PR to katuma staging and verify these fees show up after the deploy.

luisramos0:
ok, after deploying this PR the fees were updated on the DB but the cache was holding the old values, so the issue was not fixed.
I had to connect to the server and run Rails.cache.clear, after that the fix worked, fees were back:
image

I added a commit to clear rails cache after the migration. I dont think this needs code review, it also done in the previous preferences migration, I could have thought about this :face_with_head_bandage: 
I am now going to test the new version of the PR (with the cache clear) on staging UK just to be sure about all this :+1:

luisramos0:
yep, all checked and nice in AUS, I run the same test as described above for an enterprise with existing fees and it all worked after the deployment of this new version. Merging now.

luisramos0:
Just to be clear: this is a release blocker, that's why I merge now. We can address anything that we find out with @Matt-Yorkley s comment above on a separate PR.

filipefurtad0:
Awesome @luisramos0!

Thank you for documenting this, reproducing the issue and double checking it now works - glad you did.

Thanks.

Issue #5654, opened 2020-06-22
POC migration to fast_jsonapi


luisramos0:
### What we should change and why (this is tech debt)
AMS is not mantained anylonger we need to move to something else. Discussion here:
https://community.openfoodnetwork.org/t/alternatives-to-ams/1717/

We need to try fast_jsonapi and see how it would look like. From what I read it will consist of minor adaptions in the serializers and some work to adapt the frontend code to the new jsonapi compatible payloads :tada: 

Ideally, this issue includes more info, or even a second POC, using other alternatives jbuilder or  blueprinter or other.

### Context
Last raised in https://github.com/openfoodfoundation/openfoodnetwork/pull/5638

### Impact and timeline
AMS works with rails 6, but if we know we will need to change why not adapt now.

jeduardo824:
Hey, I can try to move some serializers (I don't know how many) to `fast_jsonapi` as a POC and check if will be simple to move from AMS, what do you guys think?

luisramos0:
yes please, that would be amazing!

jeduardo824:
Ops, I think we will need to wait a little bit more before doing that because `fast_jsonapi` (now renamed to `jsonapi_serializer`) needs Rails 4.2, and we are at 4.1.

luisramos0:
Hello @jeduardo824 
I guess you mean https://github.com/jsonapi-serializer/jsonapi-serializer/  
I think you are right, even fast_jsonapi started with rails 5 and then moved to support rails 4.2 :face_with_head_bandage: 
This is blocked by rails 4.2 upgrade. For what is worth, I have now created the rails 4.2 upgrade issue :-)
https://github.com/openfoodfoundation/openfoodnetwork/issues/5761

luisramos0:
Anyway, thanks a lot, it's really good we have a specific plan in this area :+1:

jeduardo824:
Awesome! I will be happy to help with the 4.1 and 4.2 migration if you guys need anything else (I saw that almost all tickets are assigned, so I don't know if there is anything else to do with the 4.1 migration) and after that, I can move forward with the POC of `jsonapi-serializer`

luisramos0:
:ok_hand: 

If you look at the latest/best 4.1 build on PR #5611 https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-5611/builds/4
You can see other issues to be solved.
Including instances of #5606 with errors like:
[...]

PR #5810, opened 2020-07-22
Improve DFC Provider engine to support version 1.3

### LINKED ISSUE #5809: Improve the DFC API to match the standard evolution
In the version 1.3, the DFC Catalog format was extended in order to add more information into it like Users, Enterprises, Supplied Products, Catalog items and Offers. The data graph root changed to be on an User reference.

The Products API call (/api/dfc_provider/enterprises/default/products.json
) should returns the data in the new format.

## Useful information for inception
The version 1.3 format is described here; https://datafoodconsortium.gitbook.io/dfc-standard-documentation/technical-specification/api-examples.

pacodelaluna:
#### What? Why?

Closes #5809 

I just supplied more data into the DFC Products API. The mapping chosen was:
- DFC Person with OFN User
- DFC Enterprise with OFN Enterprise
- DFC Supplied Product with OFN Product
- DFC Catalog Item with OFN Variant
- DFC Offer with OFN Variant (duplicated mapping as we don't have specific offers yet on OFN)

I decided to use multiple serializers in order to be clearer. The `@id` should be an URI identifier from the `@baseUrl` so I decided to extend the API in order to support it. The controllers will be used later in order to extend the API features, especially with CRUD actions on Supplied Products and Catalog Items.

#### What should we test?
The call on /api/dfc_provider/enterprises/default/products.json should return the JSON in the format described in https://datafoodconsortium.gitbook.io/dfc-standard-documentation/technical-specification/api-examples version 1.3 section.

#### Release notes
- Extend DFC Provider engine in order to support version 1.3

Changelog Category: Changed


engines/dfc_provider/app/serializers/dfc_provider/catalog_item_serializer.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# Serializer used to render a DFC CatalogItemt from an OFN Product
```
luisramos0, line 3: should be CatalogItem instead of CatalogItemt right?

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
```diff
@@ -0,0 +1,24 @@
...
+    def serialized_data
+      {
+        "@id" => "offers/#{@variant.id}",
+        "@type" => "dfc:Offer",
+        "dfc:offeresTo" => {
+          "@type" => "@id",
+          "@id" => "/customerCategoryId1"
+        },
+        "dfc:price" => @variant.price,
+        "dfc:stockLimitation" => @variant.total_on_hand,
```
luisramos0, line 20: the external api is on_hand, total_on_hand is internal and historical from spree.
I am sorry this is not clearly documented anywhere :-(

engines/dfc_provider/app/serializers/dfc_provider/supplied_product_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def serialized_data
+      {
+        "@id" => "/products/#{@product.id}",
+        "dfc:hasUnit" => {
+          "@id" => "/unit/#{@product.unit_description.presence || 'piece'}",
+          "rdfs:label" => "#{@product.unit_description.presence || 'piece'}"
+        },
+        "dfc:quantity" => @product.total_on_hand,
```
luisramos0, line 18: this is legacy spree that we should delete.
the correct way is product.on_hand

luisramos0: btw, I wonder if we should even have this in the system. The variants will most of the times have different sizes or volumes or weights so this number is most of times meaningless.

pacodelaluna:
@luisramos0 Thanks for your feedback, this PR is still in progress, I will fix typos and improve cosmetics. The main issue is about values mapping between DFC and OFN. Keep you updated asap, sorry for the delay, I thought I already commented this ticket.

pacodelaluna:
I have applied the Rubocop suggestions but I have still an issue with JSON serialization. I am using hashes and `to_json` method which is giving too long method, and also not really efficient processing. I need then to choose a new gem for serialization. jsonapi-rb lib (http://jsonapi-rb.org/) looks good, I need to test how I can map JSON-LD format properly into it, there is a new DSL for this gem, looks simple but I need to be sure it is powerful enough for JSON-LD.

luisramos0:
we are aiming for fastjson_api https://community.openfoodnetwork.org/t/alternatives-to-ams/1717/4
but we are blocked until we reach rails 4.2 https://github.com/openfoodfoundation/openfoodnetwork/issues/5654#issuecomment-657583767

Can we keep the slow implementation here and wait for fastjson_api? If you want to add a new serialization gem I think we should agree about it, maybe you want to create a discourse thready about it or continue the discussion in https://community.openfoodnetwork.org/t/alternatives-to-ams/1717/4 ?

luisramos0:
>> and to_json method which is giving too long method

do you want to share the code and the error? I dont think you need to use to_json if you use the AMS serializers.

RachL:
@pacodelaluna I'm not really sure I tested this correctly. But I didn't managed to get it working. 

If you need to stage while I'm away, you can use this doc: https://github.com/openfoodfoundation/openfoodnetwork/wiki/Staging-with-Semaphore :) just make sure no one is using it (the label is not used). You can use the #testing channel on Slack if you have a doubt.

pacodelaluna:
@luisramos0 I am using now ActiveModelSerializer for the JSON serialization, it is working good, I am now working on opening a bit more the API to the new objects created. Hope to complete it today.

pacodelaluna:
I will not add more logic in the PR which will become too big. The main objective is reached, the new data format is implemented.
I will just need an extra deployment on stage to test with DFC POC. And reviews and comments are most welcomed of course.

pacodelaluna:
Tests are failing on Semaphore env, but not on my local env, no getting the issue, digging.

pacodelaluna:
It finally passed! I struggled with a weird issue related to specs on Semaphore. There was an issue with the inheritance at controller side, it was working good locally but not on Semaphore. I had to specify more clearly the inherited class, you can see this in my last commit.

I have also adjust a bit more the logic, I think we are good to go.

engines/dfc_provider/app/serializers/dfc_provider/catalog_item_serializer.rb
```diff
@@ -13,6 +13,8 @@ class CatalogItemSerializer < ActiveModel::Serializer
              serializer: DfcProvider::OfferSerializer,
              key: 'dfc:offeredThrough'
 
+    delegate :sku, to: :object
```
luisramos0, line 16: I dont think you need this. AMS will do this for you. I think you can simply remove this line.

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
```diff
@@ -10,6 +10,8 @@ class OfferSerializer < ActiveModel::Serializer
     attribute :price, key: 'dfc:price'
     attribute :stock_limitation, key: 'dfc:stockLimitation'
 
+    delegate :price, to: :object
```
luisramos0, line 13: same here.

engines/dfc_provider/spec/controllers/dfc_provider/api/enterprises_spec.rb
```diff
@@ -41,6 +41,14 @@
               end
             end
           end
+
+          context 'given with a wrong id' do
+            before { api_get :show, id: 999 }
+
+            it 'returns 404' do
+              expect(response.status).to eq 404
```
luisramos0, line 47: I think you can/should use :not_found instead of 404

engines/dfc_provider/app/controllers/dfc_provider/api/people_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# Controller used to provide the People API for the DFC application
+module DfcProvider
+  module Api
+    class PeopleController < BaseController
```
luisramos0: you can use persons, it's correct english and will be more obvious that it is the controller of the person object :-)

engines/dfc_provider/app/controllers/dfc_provider/api/enterprises_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      private
+
+      def current_enterprise
+        @current_enterprise ||=
+          if params[:id] == 'default'
+            current_user.enterprises.first!
+          else
+            current_user.enterprises.find(params[:id])
+          end
```
luisramos0, line 19: you can avoid this copy pasted code here.

on the base_controller have:
```ruby
def current_enterprise
  @current_enterprise ||=
          if params[enterprise_id_param_name] == 'default'
            current_user.enterprises.first!
          else
            current_user.enterprises.find(params[enterprise_id_param_name])
          end
end

protected

def enterprise_id_param_name
    :enterprise_id
end
```

and then here on the enterprise controller delete current_enterprise and just have:
```ruby
protected

def enterprise_id_param_name
    :id
end
```

luisramos0 (approved):
all right, looking good :ok_hand:

luisramos0:
please dont force push from now on, only add commits, that way I dont need to look at the full PR again, thanks!

engines/dfc_provider/app/controllers/dfc_provider/api/enterprises_controller.rb
pacodelaluna: Yes, I could DRY this, I was not sure about it as it hides a bit the logic, but it is not that complex for sure.

engines/dfc_provider/app/controllers/dfc_provider/api/people_controller.rb
pacodelaluna: Oh ok, didn't know about that.

pacodelaluna:
@luisramos0 I have applied your changes, tell me if it is good, then I can rebase the PR and move next step.

engines/dfc_provider/spec/controllers/dfc_provider/api/enterprises_spec.rb
luisramos0: I meant the symbol :not_found :-) like this:
              expect(response.status).to eq :not_found

luisramos0:
:+1: all good.
I moved this back to code review so another dev can review, no need to rebase imo :+1:

pacodelaluna:
@luisramos0 Should I wait for this extra validation then?

luisramos0:
yes, we must have two reviewers on this PR before it is moved to Test Ready.
But meanwhile you can stage it and test it if you want :+1:

luisramos0:
Hey @pacodelaluna this is in code review and does require a second review before merging.
I think it's better if you leave the merging of PRs in all projects for one of the core devs. Is that ok for you? The column Ready to Go is specifically for that.

Meanwhile, I dont think this is going to cause any trouble in master so we are ok and dont need to revert.
I will ask one of the other core devs to review this. We can make amendments in another PR. I will ping them on slack now.

pacodelaluna:
Hi @luisramos0, ah yes sure, I misunderstood it, sorry about that. Please remove my right for merge and I will use the tag `Ready to go` then. Let me know if it causes any trouble.

engines/dfc_provider/app/controllers/dfc_provider/api/base_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+# Controller used to provide the API products for the DFC application
+module DfcProvider
+  module Api
+    class BaseController < ActionController::Base
+      rescue_from ActiveRecord::RecordNotFound, with: :not_found
+
+      before_action :check_authorization,
+                    :check_user,
+                    :check_enterprise
```
Matt-Yorkley, line 11: Looking at these methods, I'd probably rename this last one to `#set_enterprise`. The other two are checking some validity, this method is setting the `@current_enterprise` instance variable.

engines/dfc_provider/app/controllers/dfc_provider/api/catalog_items_controller.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# Controller used to provide the API products for the DFC application
+module DfcProvider
+  module Api
+    class CatalogItemsController < DfcProvider::Api::BaseController
+      def index
+        render json: current_user, serializer: DfcProvider::PersonSerializer
```
Matt-Yorkley, line 8: In the corresponding spec for the `CatalogItemsController#index` action, the test is checking if a variant is rendered, but in the controller action here it's rendering a user..?

Matt-Yorkley: Shouldn't this action be a list of variants?

engines/dfc_provider/app/controllers/dfc_provider/api/enterprises_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  module Api
+    class EnterprisesController < DfcProvider::Api::BaseController
+      def show
+        render json: current_enterprise, serializer: DfcProvider::EnterpriseSerializer
+      end
+
+      private
+
+      def enterprise_id_param_name
+        :id
```
Matt-Yorkley, line 14: I don't see anywhere where this is used..?

engines/dfc_provider/app/controllers/dfc_provider/api/persons_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      before_action :check_user_accessibility
+
+      def show
+        render json: user, serializer: DfcProvider::PersonSerializer
+      end
+
+      private
+
+      def user
+        @user ||= Spree::User.find(params[:id])
```
Matt-Yorkley, line 18: So this controller is only supposed to allow access to show the current user? I'm not sure we need to fetch the user record again here, it's already have it available via `#current_user`, right?

engines/dfc_provider/app/serializers/dfc_provider/catalog_item_serializer.rb
```diff
@@ -0,0 +1,55 @@
...
+    attribute :type, key: '@type'
+    attribute :references, key: 'dfc:references'
+    attribute :sku, key: 'dfc:sku'
+    attribute :stock_limitation, key: 'dfc:stockLimitation'
+    has_many :offered_through,
+             serializer: DfcProvider::OfferSerializer,
+             key: 'dfc:offeredThrough'
+
+    def id
+      dfc_provider_routes.api_dfc_provider_enterprise_catalog_item_url(
```
Matt-Yorkley, line 17: Some quick tests to check the output of this serializer might be nice, looking at `CatalogItemSerializer#id` and `CatalogItemSerializer#references` in particular.

engines/dfc_provider/app/serializers/dfc_provider/enterprise_serializer.rb
```diff
@@ -0,0 +1,53 @@
...
+    def supplies
+      Spree::Variant.
+        joins(product: :supplier).
+        where('enterprises.id' => object.id)
+    end
+
+    def manages
+      Spree::Variant.
+        joins(product: :supplier).
+        where('enterprises.id' => object.id)
```
Matt-Yorkley, line 44: I see `#supplies` and `#manages` are identical here. Is that right?

Matt-Yorkley: `#supplies` is using the `SuppliedProductSerializer` and `#manages` is using `CatalogItemSerializer`. The catalog items are what's being sold (distributed), right?

engines/dfc_provider/spec/controllers/dfc_provider/api/catalog_items_controller_spec.rb
```diff
@@ -96,4 +99,53 @@
       end
     end
   end
+
+  describe('.show') do
```
Matt-Yorkley, line 103: No need for braces here, and the dot notation isn't used elsewhere in OFN specs. I would use: 
```ruby
describe '#show' do`
```

engines/dfc_provider/spec/controllers/dfc_provider/api/persons_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+          allow_any_instance_of(DfcProvider::AuthorizationControl)
+            .to receive(:process)
+            .and_return(user)
+        end
+
+        context 'given with an accessible id' do
+          before do
+            api_get :show,
+                    id: user.id
+          end
```
Matt-Yorkley, line 27: This can be one line:
```ruby
before { api_get :show, id: user.id }
```

engines/dfc_provider/app/controllers/dfc_provider/api/catalog_items_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      private
+
+      def variant
+        @variant ||=
+          Spree::Variant.
+            joins(product: :supplier).
+            where('enterprises.id' => current_enterprise.id).
+            find(params[:id])
```
Matt-Yorkley, line 22: I think this controller is supposed to show products which are sold/distributed and not just produced/supplied, is that right?

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def offeres_to
+      {
+        '@type' => '@id',
+        '@id' => nil
+      }
+    end
+
+    def stock_limitation
+      object.on_hand
```
Matt-Yorkley, line 29: I'm a bit concerned we're displaying stock levels of variants in lots of places but there's no variant override scoping anywhere...

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Serializer used to render the DFC Offer from an OFN Product
+# into JSON-LD format based on DFC ontology
+module DfcProvider
+  class OfferSerializer < ActiveModel::Serializer
+    attribute :id, key: '@id'
+    attribute :type, key: '@type'
+    attribute :offeres_to, key: 'dfc:offeres_to'
```
Matt-Yorkley, line 9: Maybe this should be "offers"?

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
Matt-Yorkley: I don't fully understand the use cases and context, but I'm also concerned we're showing product lists in general with no reference to order cycles, tags, whether or not the enterprise is considered "active", and various other things that would all modify product lists.

luisramos0: I think here we are not looking at OCs/availability.
We are simply looking at catalog without any OCs/overrides. We will need to include that in Offers at some point!

engines/dfc_provider/app/controllers/dfc_provider/api/catalog_items_controller.rb
pacodelaluna: I agree it is not super clear but the DFC specification is nesting the items catalog into an enterprise, and then the entreprise into an user. I will check out if it is designed by choice or by necessity (semantic specification).

engines/dfc_provider/app/controllers/dfc_provider/api/persons_controller.rb
pacodelaluna: It was refactored few times, and I kept it this way in case an user could access the catalogue of an other user in the future, I wanted to keep the routes as open as possible to avoid later refactoring, always annoying to refactor routes.

engines/dfc_provider/app/serializers/dfc_provider/enterprise_serializer.rb
pacodelaluna: Yes exactly, SuppliedProduct is what you have in your hand, CatalogItem is what you will actually sell (and so linked to offers).

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
pacodelaluna: Right, https://github.com/datafoodconsortium/ontology/blob/master/machine%20readable%20files/business%20model/DFC_BusinessOntology.owl#L1674, perdon my french :)

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
pacodelaluna: We can discuss about data mapping and change it, no issue about that, I needed to return a value and I thought it was the closest fit.

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
pacodelaluna: Ok, the issue was on the DFC doc, https://datafoodconsortium.gitbook.io/dfc-standard-documentation/technical-specification/api-examples, will update it too.

engines/dfc_provider/app/controllers/dfc_provider/api/catalog_items_controller.rb
pacodelaluna: About data mapping, nothing is fixed, I mostly mapped as I could in order to return data for tests. We also need to use the same data source than on the OFN API, we will need to call same services or objects. As @luisramos0 mentioned it, we can't afford to maintain 2 different APIs, DFC API should call OFN logic the same way than the OFN API.

engines/dfc_provider/app/controllers/dfc_provider/api/catalog_items_controller.rb
luisramos0: :-O if you need to keep this as is, it needs a comment explaning it, ok?

engines/dfc_provider/app/serializers/dfc_provider/offer_serializer.rb
luisramos0: yes, this is the closest fit.
As long as we are on catalog looking at products and variants only, there's no problem in ignoring variant overrides.
As soon as you start looking at orders, you must include variant overrides otherwise data will be wrong.
A variant in a line item needs to be scoped to the distributor of the order, otherwise you will be looking at data (stock and price) from the supplier not the distributor.

engines/dfc_provider/app/controllers/dfc_provider/api/catalog_items_controller.rb
pacodelaluna: Added a comment here, https://github.com/openfoodfoundation/openfoodnetwork/pull/6158/commits/2de150f11fa0e61f23daaf0a6b558804749c217c, is it ok?

luisramos0: that's acceptable yes :+1: 
But from an API design perspective, I think this is wrong. This Person object with affiliates inside (and items inside) should be on the /persons endpoint, for me this would be the index action on the PersonsController, not the CatalogItems endpoint.

Issue #5813, opened 2020-07-23
Delay loading products in O/C Outgoing Tab when clicking on Expand All / down arrow


emilyjeanrogers:
## Description
The fix to Issue [#5773](https://github.com/openfoodfoundation/openfoodnetwork/issues/5773) has an outstanding problem. There is a delay while the name of the distributor appears, before a customer can click on Expand All to open the product list in the Outgoing Tab of Order Cycle Management. If customer clicks on the down arrow or Expand All link, an error is triggered.

## Expected Behavior
It shouldn't be possible to attempt to expand the product list while it is not available to load 

## Actual Behaviour

![Screen Shot 2020-07-23 at 11 39 15 AM](https://user-images.githubusercontent.com/47620442/88245089-2fcaa280-ccd9-11ea-9ddd-4e1e5c7e1072.png)

## Steps to Reproduce

- Go to an open Order Cycle 
- Go to Outgoing Tab
- Click on Expand arrow in Products before name of Distributor loads
- Error 

See Issue [#5773](https://github.com/openfoodfoundation/openfoodnetwork/issues/5773) for historical discussion.

## Workaround
Wait until distributor name loads

## Severity
S4

PR #5822, opened 2020-07-24
Performance improvement of the outgoing products page when editing OCs

luisramos0:
#### What? Why?

Testing #5775 in production we realized loading the enterprise is taking a very long time (15secs for the enterprise experiencing the bug, it's an OC with 400+ products coming from 20+ suppliers). I looked at the code I realized I could improve performance a little. See commit message.

This improvement will mitigate the problem in #5813, but wont solve it.

#### What should we test?
Like in #5775 verify the OC page, specifically outgoing exchanges, is still working with the option "coordinator inventory only". It should just be faster, I wonder if that can be seen before and after the PR :-)

#### Release notes
Changelog Category: Changed
Performance improvement of the outgoing products page when editing OCs with large numbers of products.

### DIFF
--- a/app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
+++ b/app/serializers/api/admin/for_order_cycle/enterprise_serializer.rb
@@ -31,8 +31,7 @@ class Api::Admin::ForOrderCycle::EnterpriseSerializer < ActiveModel::Serializer
     products_relation = object.supplied_products
     if order_cycle.prefers_product_selection_from_coordinator_inventory_only?
       products_relation = products_relation.
-        visible_for(order_cycle.coordinator).
-        select('DISTINCT spree_products.*')
+        visible_for(order_cycle.coordinator)
     end
     products_relation
   end


filipefurtad0:
Hey @luisramos0 ,

Thank you for this fix!
But it looks like this PR has not been pushed to Semaphore yet, I find no link to stage this. Could you please have a look? Thanks.

sauloperez:
for some reason, it refused to run the build. You might need to push force @luisramos0 . Manually retrying fails too

Screenshot from 2020-07-31 09-15-09

luisramos0:
ok, force pushed, the build is there now :+1:

filipefurtad0:
Hi @luisramos0,

I checked if the option "coordinator inventory only" is still working and tried to chase down faster page loads, for an OC with ~300 variants (id=203888). I had a few runs, on pressing "Expand All" and "Load All Variants". Also had a look at one with around ~700 variants (id=202632).

Page loads did appear slightly faster, and I did not observe any issues related to #5813 .

Looks good to go - thanks for improving this.

Issue #5818, opened 2020-07-23
Transaction fees charged several times, after using an invalid 3D-card and retrying with valid cards


filipefurtad0:
## Description

Attempting to pay with Stripe-SCA and the card:

4000008260003178 - _This card requires authentication for one-time payments. All payments will be declined with an insufficient_funds failure code even after being successfully authenticated or previously set up._ 

displays the insufficient funds error as expected, but breaks the checkout process. Further attempts to pay with valid cards do not work and, in case a transaction fee is set, this value is charged several times (as many as one attempts).

## Expected Behavior

After using an invalid card, paying with a valid card should be possible. Transaction fees should only be charged once.

## Actual Behaviour

Transaction fees are charged several times, after checkout is broken by using an "insufficient funds" 3D card.

## Steps to Reproduce

Admin
1. Set up Stripe-SCA and add a transaction fee to this payment method

Customer
2. Place items in the cart, and proceed to check out
3. Choose Stripe SCA payment method an insert the card: 4000008260003178 - a warning "Your card has insufficient funds" should be visible
4. Now attempt to pay with the valid card 4000002760003184 - you will see the same error, although this is a valid card.
5. If you re-attempt, you will see a different error "Invalid payment state"
6. Now press F5 - Refresh the page until the error dissapears
7. Insert a valid card
8. Check the values in the order confirmation

## Animated Gif/Screenshot

After first payment attempt, with **invalid** card 4000008260003178:
![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/ec4b2c69-bc8e-49ac-a907-4d061c3e1aca)

After second payment attempt, with **valid** card 4000002760003184:
![image](https://user-images.githubusercontent.com/49817236/88377427-0ec88700-cd97-11ea-90a7-02ea3780518b.png)

Order confirmation:

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/1298e354-ac6d-4ae8-87bd-e3436c1b4f66)

Order State:

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/a1f36ab4-1a67-404b-a970-7029db752415)

Payments:

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/9de33d46-c0b1-4e5e-b022-640c5b3f5675)

Stripe Dashboard:

[...]

Issue #5785, opened 2020-07-19
Stripe not accepting multiple credit cards, can not process order


chezaorchard:
## Description
Customer attempting to check out - completes credit card details for stripe payment - details are not accepted. Customer attempts this with three different credit cards, from two different issuing banks, all return errors. See below errors:

ING debit card - your card has been declined.
ING debit card - your card does not support this type of purchase.
ANZ credit card - your card has been declined.

Customer was unable to process order and contacted hub.

NB. Notes from Customer

![image](https://user-images.githubusercontent.com/2911176/87873182-b3973d00-ca02-11ea-89be-0a0dce2a69e3.png)

## Expected Behavior
Customer credit card details should have been accepted and transaction processed without error. 

## Actual Behaviour
Customer could not check out, order was unable to be made.

## Steps to Reproduce
Not able to reproduce

## Workaround
Hub manually processed the transaction for the customer.

## Severity
bug-s1: a critical feature is broken: checkout, payments, signup, login

## Your Environment

* Version used:
* Browser name and version: Safari
* Operating System and version (desktop or mobile): Desktop

luisramos0:
This could be a problem with the stripe account of the hub? Is the hub able to receive payments from other customers?

kirstenalarsen:
yes @luisramos0  the hub is receiving payments from other customers. They are having a lot of occurrences of #5774 as well - every time there is a failed payment. 

I have just had a look and the two affected customers had emails meluke@skymesh.com.au and jenniebrown769@gmail.com - there are completed orders for them in the system as the hub put them through (using their own credit card). The customers were unable to complete their own orders

I cannot find any record of these customers in Stripe i.e. I can't see the failed payments. I'm not sure if it is possible that OFN didn't get as far as talking to Stripe?. The payment method is Stripe
![image](https://user-images.githubusercontent.com/2911176/87872935-51d5d380-ca00-11ea-9749-285116470992.png)

They are both regular customers - unlikely to be major user error

For one of them (Jennie) we can see all these orders that she has not completed - I don't know how many of them were tried. When I search stripe for those order numbers I get nothing - no record of them in Stripe
![image](https://user-images.githubusercontent.com/2911176/87873113-fa386780-ca01-11ea-96ff-bd34817b3626.png)

Where there are errors of failed payments it looks like most of the errors are, which I think is unusual 
[...]

PR #5824, opened 2020-07-24
Reset order state to cart in case the stripe SCA authorization step fails

### LINKED ISSUE #5816: Stripe-SCA payment do not include ship/transaction fees (on a second payment attempt)
## Description

PR #5798 introduced changes to address the bug described in issue #5774. This relates to incorrect payment amounts processed by Stripe payments, after a first failed attempt, due to an invalid card. These incorrect payment amounts relate to transaction and shipment fees, set by the shop.

It was verified that this PR solves the issue for Stripe-Connect but that the issue persists with Stripe-SCA, namely it what concerns enterprise fees - which is the scope of the present issue.

## Expected Behavior

After a failed payment attempt, a successful payment should cover the total value of the order, including fees.

## Actual Behaviour

After a failed payment attempt, a successful payment does not cover the total value of the order: transaction fees remain marked as "Balance Due (€xx)".

## Steps to Reproduce

Admin
1. Set a transaction for Stripe-SCA and a shipment fee. Open an OC with these payment and shipment methods.

Customer
2. Insert items in the cart, from the above mentioned OC, proceeding to checkout.
3. Choose Stripe-SCA as payment method and Insert the invalid card 4000000000000069 (Charge is declined with an expired_card code)
4. Insert a valid test card (with or without 3D-authentication)
5. Check the total value of your order on the order confirmation

Admin
6. Compare the order total with:
i) Stripe Dashboard
ii) /orders page

Verify the values don't match with the order total. In i) and ii) it should be visible that:

Total order amount = charged amount + transaction fee

Depending on the card used, a failed payment may appear as well under:
iii) /admin/orders/Rxxxxxxxxxxx/payments

## Animated Gif/Screenshot

- Stripe Dashboard:

image

- /orders page (Admin):

image

## Workaround

No workaround

## Severity

## Your Environment

* Version used: Observed while testing PR #5798, which should integrate v3.2.0
* Browser name and version: Firefox 78
* Operating System and version (desktop or mobile): Desktop Ubuntu 20.04

## Possible Fix

Perhaps related to #5818 .

luisramos0:
#### What? Why?

Closes #5816 and #5818
I am putting these two fixes together in one PR because they are not too big but mostly so we can test these things together, they are different but closely related.

Regarding #5816
The handler of any gateway exceptions is used for the StripeSCA authorization step. Whenever there's an error on checkout we need to reset the order to cart state.
We were not doing that before and that was causing the order to stay in the payment state after an authorization error. Because the order was in the payment state, the callback to update order totals was not executed on the second attempt and that caused not all fees to be collected #5816

This is a generic improvement so it will make the checkout code more resilient for some other cases as well.

Regarding #5818
Here we are adding an alternative to [the infamous after_rollback callback called persist_invalid for the payment model](https://github.com/openfoodfoundation/spree/blob/0b0c422369c82b6dd7e7cb627a24e3a9fca19a6c/core/app/models/spree/payment.rb#L35).
On checkout, in the situation where we are [handling a redirect from stripe](https://github.com/openfoodfoundation/openfoodnetwork/blob/add7bb489f52e73b6763ad32c033cf401c57fac7/app/controllers/checkout_controller.rb#L159) and capturing the payment, the payment is correctly set to **failed state** but the order workflow state machine rolls it back to **pending state** causing #5818. Here we are adding some logic to the OrderWorkflow service to persist the payments after a failure so we get the data for debugging.
Related to this there's a good refactoring with code moved from checkout controller to the OrderWorkflow service.

Additionally, while investigating I thought I'd have to patch the checkout workflow code, in the end I didnt but I ended up bringing the checkout.rb file from spree_core to OFN. No changes to it, just rubocop and transpec.
 
#### What should we test?
We need to verify the issue is solved.
The code changed is related to any errors that happen on the stripe side so we can test any other error scenario on stripe and verify that fixing it will process the payment in stripe with the correct amount. For example, we can check if this will also work if we use a card that needs external authorization and we fail the validation.

I just verified that #5816 is also present when we use a card like 4000 0027 6000 3184 and press "fail authentication". So:
- on first attempt use 4000 0027 6000 3184 and press "FAIL authentication"
- on second attempt use 4000 0027 6000 3184 and press "COMPLETE authentication"
This test case should fail before this PR and pass after this PR :+1:

#### Release notes
Changelog Category: Fixed
Fixes StripeSCA problem where the second attempt to make a payment would fail even the card and details were valid.

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -32,8 +32,6 @@ class CheckoutController < Spree::StoreController
 
   helper 'spree/orders'
 
-  rescue_from Spree::Core::GatewayError, with: :rescue_from_spree_gateway_error
-
   def edit
     return handle_redirect_from_stripe if valid_payment_intent_provided?
 
@@ -41,6 +39,8 @@ class CheckoutController < Spree::StoreController
     # a version of paypal that uses this controller, and more specifically
     # the #action_failed method, then we can remove this call
     OrderCheckoutRestart.new(@order).call
+  rescue Spree::Core::GatewayError => e
+    rescue_from_spree_gateway_error(e)
   end
 
   def update
@@ -51,8 +51,6 @@ class CheckoutController < Spree::StoreController
 
     checkout_workflow(params_adapter.shipping_method_id)
   rescue Spree::Core::GatewayError => e
-    # This rescue is not replaceable by the generic rescue_from above because otherwise the rescue
-    #   below (StandardError) would catch the GatewayError and report it as a generic error
     rescue_from_spree_gateway_error(e)
   rescue StandardError => e
     flash[:error] = I18n.t("checkout.failed")
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -254,6 +254,9 @@ class CheckoutController < Spree::StoreController
   end
 
   def rescue_from_spree_gateway_error(error)
+    Bugsnag.notify(error)
+    checkout_failed
+
     flash[:error] = t(:spree_gateway_error_flash_for_checkout, error: error.message)
     # This can also happen during the edit action
     #   but the response needed here is the same as when the update action fails
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -255,10 +255,9 @@ class CheckoutController < Spree::StoreController
 
   def rescue_from_spree_gateway_error(error)
     flash[:error] = t(:spree_gateway_error_flash_for_checkout, error: error.message)
-    respond_to do |format|
-      format.html { render :edit }
-      format.json { render json: { flash: flash.to_hash }, status: :bad_request }
-    end
+    # This can also happen during the edit action
+    #   but the response needed here is the same as when the update action fails
+    update_failed_response
   end
 
   def permitted_params
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -254,13 +254,10 @@ class CheckoutController < Spree::StoreController
   end
 
   def rescue_from_spree_gateway_error(error)
-    Bugsnag.notify(error)
-    checkout_failed
-
     flash[:error] = t(:spree_gateway_error_flash_for_checkout, error: error.message)
     # This can also happen during the edit action
-    #   but the response needed here is the same as when the update action fails
-    update_failed_response
+    #   but the actions and response needed are exactly the same as when the update action fails
+    update_failed(error)
   end
 
   def permitted_params
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -175,9 +175,7 @@ class CheckoutController < Spree::StoreController
         return if redirect_to_payment_gateway
       end
 
-      @order.select_shipping_method(shipping_method_id) if @order.state == "delivery"
-
-      next if OrderWorkflow.new(@order).next
+      next if OrderWorkflow.new(@order).next({ shipping_method_id: shipping_method_id })
 
       return update_failed
     end
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -39,13 +39,13 @@ class CheckoutController < Spree::StoreController
 
     # This is only required because of spree_paypal_express. If we implement
     # a version of paypal that uses this controller, and more specifically
-    # the #update_failed method, then we can remove this call
+    # the #action_failed method, then we can remove this call
     OrderCheckoutRestart.new(@order).call
   end
 
   def update
     params_adapter = Checkout::FormDataAdapter.new(permitted_params, @order, spree_current_user)
-    return update_failed unless @order.update(params_adapter.params[:order])
+    return action_failed unless @order.update(params_adapter.params[:order])
 
     fire_event('spree.checkout.update')
 
@@ -54,7 +54,7 @@ class CheckoutController < Spree::StoreController
     rescue_from_spree_gateway_error(e)
   rescue StandardError => e
     flash[:error] = I18n.t("checkout.failed")
-    update_failed(e)
+    action_failed(e)
   end
 
   # Clears the cached order. Required for #current_order to return a new order
@@ -152,7 +152,6 @@ class CheckoutController < Spree::StoreController
       checkout_succeeded
       redirect_to(order_path(@order)) && return
     else
-      flash[:error] = order_error
       checkout_failed
     end
   end
@@ -165,7 +164,7 @@ class CheckoutController < Spree::StoreController
 
       next if OrderWorkflow.new(@order).next({ shipping_method_id: shipping_method_id })
 
-      return update_failed
+      return action_failed
     end
 
     update_response
@@ -193,7 +192,7 @@ class CheckoutController < Spree::StoreController
       checkout_succeeded
       update_succeeded_response
     else
-      update_failed(RuntimeError.new("Order not complete after the checkout workflow"))
+      action_failed(RuntimeError.new("Order not complete after the checkout workflow"))
     end
   end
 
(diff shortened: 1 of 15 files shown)


app/controllers/checkout_controller.rb
```diff
@@ -261,6 +253,13 @@ def update_failed_response
     end
   end
 
+  def rescue_from_spree_gateway_error(error)
+    flash[:error] = t(:spree_gateway_error_flash_for_checkout, error: error.message)
+    # This can also happen during the edit action
+    #   but the actions and response needed are exactly the same as when the update action fails
+    update_failed(error)
+  end
```
luisramos0, line 261: the key difference here is the call to checkout_failed in update_failed that executes PostCheckoutActions that will reset the order state to cart :+1:

sauloperez:
looks like we had an infinite loop in the build.

luisramos0:
ah, yeah, I missed that. I'll have a look now,. Thanks!

luisramos0:
I tried to fix the build but this after_rollback callback breaks the specs because the specs always rollback the transactions...

I saw that Spree also removes this callback in v2.3.

So I tried to create a better solution. It was hard but I think I do have a slightly better workaround for this.

mkllnk:
The specs need more work, moving back to in-dev.

mkllnk:
I fixed all the obvious specs but there are still some failing:
```
rspec ./spec/features/consumer/shopping/checkout_spec.rb[1:1:4:2:2:5:1:1]
rspec ./spec/features/consumer/shopping/checkout_spec.rb[1:1:4:2:2:5:2:1]
rspec ./spec/features/consumer/shopping/shopping_spec.rb:414
```

The last one passed locally and may be flaky but the first two are definitely broken by this pull request. The credit card secrets are not passed onto the payment gateway and without valid credit card number it fails.

mkllnk:
Reloading the payment is a problem:
https://github.com/openfoodfoundation/openfoodnetwork/blob/07005594ff09a688f8a1af0a5a6afe4c3d1325cc/app/services/order_workflow.rb#L67

While the records look exactly the same, I guess that reloading gets rid of in-memory data like the credit card secrets.

mkllnk:
@luisramos0 I'm handing the keyboard back to you. Let me know what you think. I guess we can go ahead with this fix and create a tech debt issue. Or we investigate fixing the callback. You may have a better overview of that and it may make more sense to tackle that in bye-bye-spree.

app/controllers/checkout_controller.rb
```diff
@@ -261,6 +242,13 @@ def update_failed_response
     end
   end
 
+  def rescue_from_spree_gateway_error(error)
+    flash[:error] = t(:spree_gateway_error_flash_for_checkout, error: error.message)
+    # This can also happen during the edit action
```
sauloperez, line 247: IMO we should stick to the `rescue_from` defined at line 35 for both `#edit` and `#update` or the rescue block but having separate approaches makes things harder. This PR is a good chance.

app/services/order_workflow.rb
```diff
@@ -35,9 +43,31 @@ def advance_order!(options)
...
+
   def after_transition_hook(options)
     if order.state == "delivery"
       order.select_shipping_method(options[:shipping_method_id]) if options[:shipping_method_id]
     end
+
+    persist_all_payments if order.state == "payment"
   end
+
+  # When a payment fails, the order state machine rollbacks all transactions
```
sauloperez, line 62: just thinking out loud. If that's not what we want we'll need to split up that single transaction at some point so that a rollback while persisting payment doesn't affect the order changes. Definitely not today. Removing that callback is a lot already.

luisramos0: yeah, I'd love to do that, but I think it's state_machine that does the rollback by default if the workflow step returns false.
btw, that's why I brought Checkout.rb from spree, because I was trying to do exactly that. But I dont know enough about the state_machine gem... the other problem is that we may want the rollback to happen for other objects other than the payments...

app/controllers/checkout_controller.rb
luisramos0: I removed this comment and renamed the update_failed method to action_failed. I think that makes it more clear. It's just a generic method, not specific to edit or update.

But this takes me to another detail of this method.
The specific rescue in the update action that calls this method was added by Maikel here:
https://github.com/openfoodfoundation/openfoodnetwork/commit/bd67d63b63cca3c889755d95fcf1dd94a16ae9e0
Maybe it's a mistake because this checkout controller already had and still has the generic:
 `rescue_from Spree::Core::GatewayError, with: :rescue_from_spree_gateway_error`
Maybe we can just remove this rescue from the update action? I think we can. I am just asking because this is such a critical part of the checkout process.

sauloperez: > Maybe we can just remove this rescue from the update action?

I believe we can, otherwise, Rails would be broken

app/services/order_workflow.rb
sauloperez: yes, it feels like heart surgery. Overall, I'm a bit concerned about our current status where things are partially implemented in the Spree order state machine, plus patched on our checkout controller and our services. I guess I just need to believe in us while we simplify our customizations up until the day we can question the state machine's implementation.

app/controllers/checkout_controller.rb
luisramos0: ah, I think I see why we need this now! If I remove the first rescue the second more generic rescue will be triggered and the rescue_from rule will not be used as we want. This is why Maikel added it I am pretty sure, I think I'll add a comment to explain this :+1:

app/services/order_workflow.rb
luisramos0: yeah, I see what you mean, but we are improving!

app/controllers/checkout_controller.rb
luisramos0: I ended up removing the rescue_from line and adding the rescue to the edit action. I think this makes it very simple:
- edit rescues GatewayError
- update rescues GatewayError and StandardError

luisramos0:
hey @mkllnk nice async pair programming! Thank you!

I think your fix works! I am not sure the persisted failed payments will end up with all the info but it doesnt matter for now, first checkout will now work and second, we will have spree_log_entries which is the most important.

I added a few more simple commits.
I think this is ready to move forward.

app/services/order_workflow.rb
```diff
@@ -64,8 +64,8 @@ def after_transition_hook(options)
   def persist_all_payments
     order.payments.each do |payment|
       original_payment_state = payment.state
-      if original_payment_state != payment.reload.state
-        payment.update(state: original_payment_state)
+      if original_payment_state != Spree::Payment.find(payment.id).state
+        payment.reload.update(state: original_payment_state)
```
mkllnk, line 68: I'm still a bit confused what's happening here. Do we need a comment explaining that? My theory is:

- During checkout the order is going to state `payment` and a payment object is created.
- The order should go from state `payment` to `complete` but on error the state machine stays in `payment`.
- The payment object's normal states are: `checkout`, `pending`, `processing`, `completed`
- Error states are: `void` and `invalid`.
- A failed checkout causes a database rollback which reverts the payment state from `void` or `invalid` to `checkout` or `pending` (I don't know which).
- Despite the rollback, our in-memory payment object still has the failed state.
- Checking for a state change in the database is a sign of a rollback and we override the database with our payment state.
- In theory, if something else updated the payment record at the same time, like a webhook, we could accidentally override that. There is a potential race condition here but we are most likely fine because of the checkout sequence and synchronisation.

Do you agree with this?

What else could happen? I guess a payment could complete but something else fails and we override the payment state to mark it as completed. That's a good thing.

I reckon that the payment should happen outside of the state machine transition to not be affected by the rollback. The state machine transition just verifies that there is enough money for the order. Out of scope here.

luisramos0: I dont agree about you being confused :-)
I think what's happening is exactly what you have described :ok_hand: and that was what I thought was happening :+1: 
I used some of your description to add code comments. I think it's now better explained, thanks!

filipefurtad0:
Hey @luisramos0 and reviewers,


- About #5816 

Reproduced issue and re-ran the procedure, after staging - for both Stripe-SCA and Connect (which was already working before). This looks good, transaction fee is now correctly charged, after failing the payment once, with card ...0069. Amounts match in stripe-dashboard and on Payments (while editing the order):

image

image

Also verified we now get a warning on [Bugsnag/Slack](https://openfoodnetwork.slack.com/archives/CEF14NU3V/p1596213668446700) on this type of failed payments - which is really awesome, surely helpful as well for issues such as #5785 ! This works for both stripe methods.

- About #5818

Reproduced the issue before staging. True, it seems to break as well by manually failing the 3D authentication process. So, after this PR the process goes as expected, after the first failed attempt:

image

No additional transaction fees are added, for each failed attempt - I tried several other invalid cards, such as 4000008260003178 (insufficient funds) or 4000008400001629 (card_declined), verifying that these errors appear in Bugsnag and in stripe's Dashboard, correctly:

image


Both issues are solved - and we've got data on errors now. Awesome work :muscle: 
Good to go.

luisramos0:
This is great news. I will include this in this release 3.2.1 making it fully ready for the stripeSCA rollout :tada:

lin-d-hop:
@filipefurtad0 
Did you test this when taking payments in the backend?
On release testing I tried placing an order with an SCA card in the backend and it failed.
I think this needs a new issue.

PR #5833, opened 2020-07-27
Add unit tests for order.charge_shipping_and_payment_fees!

### LINKED ISSUE #5812: Add specs for change in PR 5798
We need to add specs for PR 5798 to validate specific cases with orders with multiple payments.

### Context
This was left TODO in 5798 so we could include the fix in the release. 

### Impact and timeline
This is important but I am not sure we will be able to replicate the issue. We can probably add specs to test the code itself (pending_payments vs payments).

luisramos0:
#### What? Why?

Closes #5812

I ended up deciding to only add a couple of unit tests validating that only pending_payments are updated. I think this is enough testing here.

#### What should we test?
Green build is enough.

#### Release notes
Changelog Category: Added
Added tests to cover some code that was added in a previous hotfix #5798

### DIFF
--- a/spec/models/spree/order_spec.rb
+++ b/spec/models/spree/order_spec.rb
@@ -855,12 +855,39 @@ describe Spree::Order do
         order.state = 'delivery' # payment's previous state
 
         allow(order).to receive(:payment_required?) { true }
-        allow(order).to receive(:charge_shipping_and_payment_fees!)
       end
 
-      it 'calls charge_shipping_and_payment_fees!' do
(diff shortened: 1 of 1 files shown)


sauloperez:
I'm merging this for the sake of the pipe's throughput. IMO it's simple enough.

sauloperez:
weird, it doesn't move on from `Checking for ability to merge automatically…` and I can't merge.

luisramos0:
Maybe another branch was being merged or something. I can merge now :+1:

PR #5851, opened 2020-07-31
Open Street Map tile provider updates

cillian:
#### What? Why?

Before these changes the _Open Street Map Provider Options_ setting wasn't being parsed correctly so you could only configure the default OSM tile provider i.e. `OpenStreetMap.Mapnik` whose usage policy probably doesn't support use in production. This fix lets you configure other tile providers correctly.

Also the leaflet-providers JS library has been updated to fix MapBox tile provider, add new tile providers and remove some no longer supported.

If this needs any changes let me know.

#### What should we test?

To test this [bug fix](https://github.com/openfoodfoundation/openfoodnetwork/commit/3c23952fd62d53c762cb2bbce9af96a2c413084c) you probably need to configure an OSM tile provider that requires an API key (https://github.com/leaflet-extras/leaflet-providers#providers-requiring-registration I can send a key if needs be)

1. Go to Admin > Configuration > Content > Map settings
2. Fill in 'MapBox' in the Open Street Map Provider Name field.
3. Fill in the Open Street Map Provider Options field with `{ "accessToken": "insert-access-token-here" }`

#### Release notes

Fix parsing of custom OSM tile provider options and bump leafletjs-providers to 1.10.2 to fix MapBox tile provider.

Changelog Category: Fixed

### DIFF
--- a/app/serializers/api/open_street_map_config_serializer.rb
+++ b/app/serializers/api/open_street_map_config_serializer.rb
@@ -15,7 +15,7 @@ module Api
     end
 
     def open_street_map_provider_options
-      ContentConfig.open_street_map_provider_options.to_json
+      ContentConfig.open_street_map_provider_options
     end
   end
 end
--- a/vendor/assets/javascripts/leaflet-providers.js
+++ b/vendor/assets/javascripts/leaflet-providers.js
@@ -219,66 +219,16 @@
 			}
 		},
 		CyclOSM: {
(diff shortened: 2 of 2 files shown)


filipefurtad0:
Hey @cillian ,

Thanks for this PR and the detailed instructions! Below, for comparison:

Mapnik, before staging this PR
image

Mapbox, after staging this PR and setting a correct (in this case, public) key:
image

It works nicely - thank you for this improvement!
Ready to go.

Issue #3714, opened 2019-04-09
[Customer->Create] Country field cannot be searched


lin-d-hop:
## Description

When I try to add a customer in v2 I cannot type into the Country field search box. Meaning that I have to scroll right down to select the appropriate country. Not an issue in v1.

## Expected Behavior

I can type into the search box to find the appropriate country quickly

## Actual Behaviour

I cannot type into the search box.

## Steps to Reproduce

1. New Customer
2. Add details
3. Try to search for a country
4. Can't type into the box

## Animated Gif/Screenshot

![Screenshot from 2019-04-09 13-50-46](https://user-images.githubusercontent.com/8669848/55801845-dda74300-5ace-11e9-8bda-88b3b0d72109.png)

## Severity

S5
## Your Environment

* Version used:
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): Ubuntu Desktop
* OFN Platform instance where you discovered the bug, and which version of the software they are using.  V2

PR #5866, opened 2020-08-06
[Bye bye Spree] Bring models country, zone and zone_member from spree_core

luisramos0:
#### What? Why?
Continues #4826
- core/app/models/spree/zone_member.rb COPY
- core/app/models/spree/zone.rb COPY
- spec/models/spree/zone_spec.rb COPY
-  core/app/models/spree/state.rb COPY
- spec/models/spree/state_spec.rb COPY
- core/app/models/spree/country.rb COPY
- spec/models/spree/country_spec.rb COPY

#### What should we test?
- Backoffice Configuration - verify all features related to creating editing and deleting Countries, States and Zones are working as before.
- Checkout Zone - the checkout zone of an instance is defined in the config files, when you know what zxone is in the instance you are testing, you can verify that the countries of that particular zone are the countries that show up in the checkout page to be used on addresses. (I think that if the checkout_zone is set to be state base zone, all countries in the database become available on the checkout page)
- Checkout addresses - verify country and state selection on checkout addresses is working well
- Addresses in the Backoffice - verify country and state selectors are working in the backofice: order customer details, and customers and enterprises
- Shipping methods - double check shipping methods zone filter is not used on checkout as before this PR, make a shipping method available in a zone that is not the instance checkout zone, verify you can still checkout using that shipping method.

Any other part of the app with should check that is related to countries, states and zones?

#### Release notes
Changelog Category: Changed
Brought code needed in OFN related to countries, states and zones from Spree so that we can make OFN independent of Spree.


app/models/spree/country.rb
```diff
@@ -6,7 +8,7 @@ class Country < ActiveRecord::Base
 
     def self.states_required_by_country_id
       states_required = Hash.new(true)
-      all.each { |country| states_required[country.id.to_s]= country.states_required }
+      all.find_each { |country| states_required[country.id.to_s] = country.states_required }
```
sauloperez: :ok_hand:

spec/models/spree/country_spec.rb
```diff
@@ -1,9 +1,15 @@
...
 describe Spree::Country do
   it "can find all countries group by states required" do
-    country_states_required= Spree::Country.create({:name => "Canada", :iso_name => "CAN", :states_required => true})
-    country_states_not_required= Spree::Country.create({:name => "France", :iso_name => "FR", :states_required => false})
+    country_states_required = Spree::Country.create({ name: "Canada",
+                                                      iso_name: "CAN",
+                                                      states_required: true })
+    country_states_not_required = Spree::Country.create({ name: "France",
+                                                          iso_name: "FR",
+                                                          states_required: false })
```
sauloperez: curly braces are actually not required, right?

sauloperez:
There are a couple of issues at the bottom in Code Climate that could easily be addressed.

spec/models/spree/country_spec.rb
luisramos0: this was testing dead code, it's removed.

luisramos0:
I fixed the last two warnings.

sauloperez (approved):

filipefurtad0:
I see branch-conflicts warning. Since this is quite an extensive test, perhaps it's best to wait till these are solved. I'll add the feedback-needed label.

sigmundpetersen:
Has this been tested @sauloperez ?

sauloperez:
Nope, my bad :see_no_evil: @filipefurtad0 is aware and he'll test it in master

luisramos0:
@filipefurtad0 has this been tested at some point?

filipefurtad0:
Hey @luisramos0 ,
A few hours ago, I have staged master-branch which includes this PR, and was checking this as of now.
I'll post my testing notes when I am done.

filipefurtad0:
Checked the test-cases you pointed out and GST zone settings.

- Backoffice Configuration, namely creating, editing and deleting States and Zones work as before. Countries cannot be added/deleted/edited through the back office. This happens through the config file `countries.yml` file, right?

- Checkout Zone - created new zone and edited the config/application.yml file + restarted unicorn: this changes successfully the checkout zone of the instance. If this newly created zone is deleted and the set value for checkout_zone in application.yml has no correspondence then I could observe that all countries in the database become available on the checkout page.

- Checkout addresses - verified that country and state selection on checkout addresses works well

- Addresses in the Backoffice - verifed that country and state selectors are working in the backofice: order customer details, and customers and enterprises. Observed bug #3714.

- Shipping methods - double check shipping methods zone filter is not used on checkout as before this PR.
i) checkou_zone is set as Europe
ii) made a "bike delivery" only available for US
iii) verified that checking out is still possible using this shipping method


- checked that GST/VAT still applies correctly to the defined zone: verfified that it is still possible to include/exclude taxes from the check out zone

All here and for deployment of release tomorrow @luisramos0 :+1:

luisramos0:
Countries cannot be added/deleted/edited through the back office. Correct. :+1:

Nice!! Thanks Filipe!

Issue #5784, opened 2020-07-17
Error 422 when placing backoffice orders with more items than those available on hand


filipefurtad0:
## Description

One of the changes introduced by PR #5406 is to prevent stock levels from reaching negative values through backoffice orders. After these changes, creating an order in the backoffice with more items than those in stock is not possible; attempting generates the error on Bugnsnag:
https://app.bugsnag.com/open-food-france/open-food-france-prod/errors/5f11cd5d9f0dd90017a69e3d?event_id=5f11cd5d005b6e2ca49f0000&i=sk&m=nw

See also pic below.

## Expected Behavior

Placing the order should succeed, but with reduced quantity to the max. available stock.

## Actual Behaviour

Attempting to complete a backoffice order with more items than those available in stock fails and displays the message "Error saving payment".

## Steps to Reproduce

Opened two browser sessions, both logged in as admin.

Session A: Set the stock level of a product to limited stock with e.g. 4.
Session B: Create an order, Add less than total stock left to cart, e.g. 3.
Session A: Update the stock level of the product to be less than the quantity in cart, e.g. 2.
Session B: Attempt to complete the order: this will fail and display the message "Error saving payment".

## Animated Gif/Screenshot

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/02fa83fc-ef07-46ff-8c2b-a96a04ff54af)

## Workaround

After seeing the error, refresh the page and insert the a nr. of items smaller or equal to that available in stock.

## Severity

bug-s4

## Your Environment

* Version used: v.3.1.0
* Browser name and version: Chromium 83
* Operating System and version (desktop or mobile): Desktop/GalliumOS

RachL:
@filipefurtad0 @luisramos0 I'm not sure I understand the expected behavior. Why would an admin be allowed to go higher than stock? Most of the users I talk to, don't know their stock info. So the actual behavior described here is something they asking for. 
Also, another workaround for the admin is to simply raise the stock level. So I really don't think this is an s2?

filipefurtad0:
Hi @RachL ,

After PR #5406 it is not possible to exceed stock with backoffice orders, this is one of the changes introduced. After this PR the actual behavior when attempting to exceed available stock in the BO is displaying a payment error message - not informative about stock limitation - and generating an error which turns out in bugsnag as well.

I agree it would be useful for users to know about stock limitations in the backoffice, wouldn't this be a feature request? 

[...]

PR #5867, opened 2020-08-06
[Bye bye Spree] Bring models stock_location and stock_movement from spree_core

luisramos0:
#### What? Why?
Continues #4826
- core/app/models/spree/stock_location.rb MERGED WITH DECORATOR
- spec/models/spree/stock_location_spec.rb COPY
- core/app/models/spree/stock_movement.rb COPY
- spec/models/spree/stock_movement_spec.rb COPY

#### What should we test?
- verify stock management is working correctly. I think a simple stock test is enough here, make a variant on_hand and another on_demand, check them out and verify the stock level is updated correctly. Verify you cant buy a out of stock variant. Also, these files are not used in variant overrides, so it's mostly related to variants that are not in inventory, but we can also make a simple check that inventory stock management is working correctly.
- may we can verify the case we add out of stock variants to an OC, they should be in the OC but not visible in the shopfront, correct?

I am not sure what else to verify here, any other point we should check here related to stock management?

#### Release notes
Changelog Category: Changed

### DIFF
--- a/spec/models/spree/stock_location_spec.rb
+++ b/spec/models/spree/stock_location_spec.rb
@@ -1,0 +1,205 @@
+require 'spec_helper'
+
+module Spree
+  describe StockLocation do
+    subject { create(:stock_location_with_items, backorderable_default: true) }
+    let(:stock_item) { subject.stock_items.order(:id).first }
+    let(:variant) { stock_item.variant }
+
+    it 'creates stock_items for all variants' do
+      subject.stock_items.count.should eq Variant.count
+    end
+
+    context "handling stock items" do
+      let!(:variant) { create(:variant) }
+
+      context "given a variant" do
+        subject { StockLocation.create(name: "testing", propagate_all_variants: false) }
+
+        context "set up" do
+          it "creates stock item" do
+            subject.should_receive(:propagate_variant)
+            subject.set_up_stock_item(variant)
+          end
+
+          context "stock item exists" do
+            let!(:stock_item) { subject.propagate_variant(variant) }
+
+            it "returns existing stock item" do
+              subject.set_up_stock_item(variant).should == stock_item
+            end
+          end
+        end
+
+        context "propagate variants" do
+          let(:stock_item) { subject.propagate_variant(variant) }
+
+          it "creates a new stock item" do
+            expect {
+              subject.propagate_variant(variant)
+            }.to change{ StockItem.count }.by(1)
+          end
+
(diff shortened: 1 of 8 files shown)


spec/models/spree/stock_location_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    context "handling stock items" do
+      let!(:variant) { create(:variant) }
+
+      context "given a variant" do
+        context "propagate all variants" do
+          subject { StockLocation.new(name: "testing") }
+
+          before { subject.propagate_all_variants = true }
```
sauloperez, line 22: This no longer needed, right?

luisramos0: ah, yes, thanks! removed in a new commit :+1:

sauloperez:
@luisramos0 this needs rebasing :recycle:

luisramos0:
:+1:

sauloperez:
there are issues now re factories.

luisramos0:
Thanks for the heads up! Some missing spree factories.

sauloperez:
it's been almost a month this PR got opened. Given it's probably one of the simplest I'm moving it to Test Ready.

filipefurtad0:
Hey @luisramos0 ,

Thanks for pointing out the main areas this PR may have impact on.

I've performed the following tests around stock management.

i) as admin, I've set a variants on_hand and on_demand
i.a) as a customer, verified they behave as expected in the shopfront (not able to buy over the set stock level). Verified this as well as an admin, for orders placed in the backoffice.

This is fine for the customer/shopfront, and works as well for the back office.

i.b) checked the cases above, with variant overrides imposed by inventory settings.
All good here.

ii) checked the case in which a shop has an open OC from which all variants are out of stock.
In this case, the shop does not appear as closed, but rather with no items available, like this (before and after the PR):

image

I don't think this is a but, but I'm not sure this is the ideal case here: how informative is it really to show an empty shop to the user?  


iii) checked the app behaves as before on some edge cases, like out of stock, during shopping journey, specifically, on /checkout
All good, redirection to the cart page takes place, and correction of the stock to a lower value (+ clicking Update) allows the customer to checkout.

This is ready to go Luis - one step closer to say Bye Bye to Spree.

Not introduced by this PR:
On test case i.a) I found another way to cause bug #5784 and place orders above the available stock. Reporting there.

luisramos0:
:tada: thanks Filipe!

Issue #5252, opened 2020-04-17
Newly created and renamed taxons are not showing as shop filters


tschumilas:
## Description
I recently needed to add new taxons as super-admin in configuration.  In some cases,  I renamed existing taxons (they were called 'z' because they were placeholders for future French translation) and in other cases I created new taxons.  I created 6 taxons in total  and they all show up correctly in category dropdown lists.  They all imported properly with product import that I did for a farm with a large list.  They all appear properly in products.  But they do not show up as filters in the shop.  I just tried to create another new taxon as a test -  and same thing.  It seems that renamed or newly created taxons are not showing as shop filters

## Expected Behavior
Taxons that are renamed  or newly created,  and are attached to products as categories,   should show in shop filters

## Actual Behaviour
Renamed and newly created taxons are not showing as shop filters. 

## Steps to Reproduce
1. Create a new taxon,  and attach it to a product.   or Rename an existing taxon and attach it to a product.
2. Create an OC with that product.
3. See that the  new taxon does not appear as a filter.

The taxons filter correctly,  but their names are not matching what is in the category list.

## Your Environment

* Version used:
* Browser name and version:
* Operating System and version (desktop or mobile):

tschumilas:
Note -  will get back to this with more -   but just wanted to put it there.  Something very confusing is happening with taxons on OFN-CAN

luisramos0:
About your comment: /shops is not shop list, it is the list of shops. I think this issue is about the product list, right?

I cant replicate this, if I change the taxon name, I see it in the filters in the shopfront and can filter by that taxon:
![image](https://user-images.githubusercontent.com/1640378/79688346-a7ee7b80-8245-11ea-98e1-888e5ca4307d.png)
![image](https://user-images.githubusercontent.com/1640378/79688347-ade45c80-8245-11ea-9b8c-1586d49c588a.png)

Why do you say you cant delete taxons? Locally, I can delete taxons that are not being used anywhere.

Can you please proviode specific example and screenshots?

tschumilas:
I wonder if the reason I can't delete is because I have no way to see if that taxon is not being used.  Maybe it was used 3 year ago and is in the database.   I'll put up some screenshots -  I can't figure out why the taxons I renamed are not showing as filters.   But a quick question -  re: deleting taxons -   not in current use,  or never used?   If a product has been deleted  -  and its taxon is used nowhere else -  can the taxon be deleted?

luisramos0:
[...]

PR #5868, opened 2020-08-06
[Bye bye Spree] Bring models taxon and taxonomy from spree_core

luisramos0:
#### What? Why?
Continues #4826
- core/app/models/spree/classification.rb MERGED WITH DECORATOR
- core/app/models/spree/taxonomy.rb COPY
- spec/models/spree/taxonomy_spec.rb COPY
- core/app/models/spree/taxon.rb MERGED WITH DECORATOR
- spec/models/spree/taxon_spec.rb MERGED WITH OFN VERSION

#### What should we test?
- Backoffice Taxonomies feature - make sure you can manage taxonomies and taxons in them as before.
- Make sure the changes in the taxonomy are reflected in the product categories both when editing the product and also when buying it in the shopfront.
- Verify the filters by category in the shopfront are working

I am not sure what else to test related to taxonomies. Any other point we should check here?

#### Release notes
Changelog Category: Changed


sauloperez:
There are a few easy to fix Code Climate issues. Other than that :+1:

luisramos0:
I am not sure about the "Useless assignment to variable - name"... so I dont change it.

I fixed the long line :+1:

sauloperez:
I'm moving it to test ready because it's a fairly simple merge and it looks :+1:

filipefurtad0:
Hi @luisramos0,

Checked the following:

i) Verified previously existing Taxons were working as correctly
ii) as Superadmin, created a new Taxon/Class. Created Subclasses.
iii) as Admin - created a new products with the newly created taxons - these appear alphabetically, when editing/creating a product. Placed them on the shopfront.
iv) as User - visited the shopfront and searched for the new product (using the search field) and filtered the results using the "Filter by", using the respective newly created taxon
v) Deleting Taxons is possible, as long as they are not attributed to a product.


Not introduced by this PR:

Error observed on test case ii) but no unusual behavior was seen on the app:

https://openfoodnetwork.slack.com/archives/G010VN35QU9/p1600101520000100


Error observed on test case v) if Taxons are attributed to a product:

https://openfoodnetwork.slack.com/archives/G010VN35QU9/p1600103192000600

image
 
I could not verify issue #5252 after staging this PR, so maybe this solves it? Would that make sense? I'll comment on that issue.

This is good to go! :tada:

filipefurtad0:
I verified #5252 now: it only affects main classes. Renaming sub-classes works both in back- and frontoffice.

luisramos0:
Nice, thanks Filipe. We could create an issue for that JS error (last_rollback), there's some rollback logic in the JS that maybe we dont need.

re #5252 :+1:

filipefurtad0:
Hey @luisramos0, thanks. Opened tech-debt #6030 concerning the removal of the rollback logic and issue #6033 on the observed errors while deleting taxons (bug-s4 I'd say).

Issue #5731, opened 2020-07-06
Can't delete variant images


filipefurtad0:
## Description

After setting an image for a product and all its variants it's possible to set specific images to individual variants. This, however, cannot be undone: it's not possible to delete variant images after these are set. It's only possible to delete the main product image.

After apparently successfully deleting the variants' image, refreshing the page shows them again. They remain visible in the /cart page, but not in /shop. Attempting to edit them show the error:
![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/e59f5df9-dea5-4c38-98da-fbcd9947f2bb)

These are apparently silent errors, no Bugsnag notification.

## Expected Behavior

It should be possible to delete images associated with variants, and set the picture of the main product to them.

## Actual Behaviour

The images remain associated with the variant in the /cart page, despite attempts to delete them in the backoffice.

## Steps to Reproduce

1. Set an image to a product and all its variants
2. Set an image to a specific variant
3. Verify that the desired changes occur in the shopfront: images to specific variants should not be visible, only the main image.
4. Insert different variants in the cart. 
5. Verify that the set images appear for the corresponding variants in the /cart page: here the image set for individual variants should be visible.
6. Delete the image previously set to an individual variant.
7. Press F5. Verify the bug: despite deletion, the image will reappear for that variant, and will be visible in the cart page.
8. Try to edit the deleted image.
9. The "Image not found" error should be visible

## Animated Gif/Screenshot

The gif below shows steps 6. to 9.:

![ghost_image_variants.gif](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/2be3f026-f149-4f50-9412-c9bd96386d8c)

## Workaround

I suppose deleting the variant and starting over would work.

## Severity

bug-s3: a feature is broken but there is a workaround

## Your Environment

* Version used: Observed in v2.1 and in v3.0.2
* Browser name and version: Firefox 
* Operating System and version (desktop or mobile): Desktop Ubuntu 18.04

luisramos0:
As I suggest in #5732, the easy fix to this and #5732 is to remove the possibility to upload images to specific variants.

luisramos0:
I think this is an S4.

filipefurtad0:
Thank you for the feedback on these issues @luisramos0!
I agree with your proposed solution for this and #5732. Downgrading to bug-s4.

RachL:
this is on the roadmap for 2026 and will be dealt with on new issues

PR #5869, opened 2020-08-06
[Bye bye Spree] Bring models image and asset from spree_core

luisramos0:
#### What? Why?
Continues #4826
- core/app/helpers/spree/admin/images_helper.rb COPY
- core/app/models/spree/image.rb MERGE WITH DECORATOR
- core/app/models/spree/asset.rb COPY
- spec/models/spree/asset_spec.rb COPY

#### What should we test?
- Verify that images are being displayed before and after this PR on the server you are testing.
- Backoffice image upload - verify you can upload an image in the bulk product edit page and in the product images page
- Verify the images display correctly in the shopfront and in the backoffice product list

I am not sure what else we need to verify. Any other point we should check here?

#### Release notes
Changelog Category: Changed


app/models/spree/image.rb
```diff
@@ -46,5 +46,27 @@ def no_attachment_errors
...
+
+      Hash[styles_a]
+    end
+
+    def self.reformat_styles
+      Spree::Image.attachment_definitions[:attachment][:styles] =
+        format_styles(Spree::Image.attachment_definitions[:attachment][:styles])
+    end
+
+    reformat_styles
```
sauloperez, line 74: I think this would be better off in an initializer. Not only is weird in here but I bet it gets unnecessarily evaluated more than once, at least in dev.

app/models/spree/image.rb
```diff
@@ -0,0 +1,76 @@
...
+    supports_s3 :attachment
+
+    Spree::Image.attachment_definitions[:attachment][:styles] =
+      ActiveSupport::JSON.decode(Spree::Config[:attachment_styles]).symbolize_keys!
+    Spree::Image.attachment_definitions[:attachment][:path] = Spree::Config[:attachment_path]
+    Spree::Image.attachment_definitions[:attachment][:url] = Spree::Config[:attachment_url]
+    Spree::Image.attachment_definitions[:attachment][:default_url] =
+      Spree::Config[:attachment_default_url]
+    Spree::Image.attachment_definitions[:attachment][:default_style] =
+      Spree::Config[:attachment_default_style]
```
sauloperez, line 30: :clap: :clap: now we can finally DRY ImageSettingsController#update_paperclip_settigns!

luisramos0: image settings controller will die here: https://github.com/openfoodfoundation/openfoodnetwork/pull/5729/files

app/models/spree/image.rb
luisramos0: this is existing code https://github.com/openfoodfoundation/openfoodnetwork/blob/7cb49ac9a8206845c935de4fe1fb38e6b197716d/app/models/spree/image_decorator.rb#L22

I dont want to fix all the spree/decorators mess in these PRs to reduce risk. Are you ok if I create a tech debt issue for this one?

luisramos0:
rebased to resolve conflicts

sauloperez (approved):

app/models/spree/image.rb
luisramos0: ok, thanks.
here: https://github.com/openfoodfoundation/openfoodnetwork/issues/5973

filipefurtad0:
Hey @luisramos0 ,

The spec `asset_spec.rb ` seems to be failing on the CI. Would you say it's ok to proceed with testing the PR? 
Thanks!

luisramos0:
well spotted Filipe, there was a problem after the rebased. I fixed it now. I dont think we need to review this simple change.
Ready for Testing I'd say.

sigmundpetersen:
The failure on the last build is a flaky one which I think we can ignore.

filipefurtad0:
Hey @luisramos0,

Thank you! I have a question about spotting flaky specs. The build was failing again after your fix, but due to a different spec,` rspec ./spec/features/admin/variant_overrides_spec.rb:373 # ` which I believe to have seen failing before - so maybe this is a flaky spec, and no reason for alarm? What's the best way to be sure? After updating master on my local system, I ran this spec several times locally and it always passed - but I realize now that this PR has not been merged. Any advice on this would be great, thanks.

About the PR, I checked the images in staging-UK and before and after staging:

- emails
- group pages
- maps, specifically shop/group pointers
- products (front- and backoffice). Image upload works as well - tested both from bulk product edit or while editing the product itself; changes are reflected in the shopfront. Still can't delete product/variant images # [5731](https://github.com/openfoodfoundation/openfoodnetwork/issues/5731)
- hub profiles
- several places in the homepage
- settings seem to be retained, also the credentials for S3 amazon server. On this respect, enabling S3 for images works well (as before the PR). I noticed, that after enabling S3-images images seem broken: one needs to refresh the page a couple of times and it all works after that. I don't think this relates to the PR.


All as before, ready to go!

PS: I just wrapping up and saw you message @sigmundpetersen - thank you for confirming this! :+1:

filipefurtad0:
I was about to remove the staging label but realized I forgot to add it in the last hour. I'm posting the slack-link as to document that this PR was staged in staging-uk:
https://openfoodnetwork.slack.com/archives/CEF14NU3V/p1599034343019700

sigmundpetersen:
About spotting flaky specs, one way is to re-run the build several times and if the spec in question is only failing sometimes it is a flaky one. If it is a failure related to the PR it usually fails consistently.

luisramos0:
Yes, this one is flaky.
you can always press
image

PR #5870, opened 2020-08-06
[Bye bye Spree] Bring models app_configuration and preference from spree_core

luisramos0:
#### What? Why?
Continues #4826
- core/app/models/spree/preferences/configuration.rb COPY
- core/spec/models/spree/preferences/configuration_spec.rb COPY
- core/app/models/spree/preferences/preferable.rb COPY
- core/spec/models/spree/preferences/preferable_spec.rb COPY
- core/app/models/spree/preferences/preferable_class_methods.rb COPY
- core/app/models/spree/app_configuration.rb MERGED WITH DECORATOR
- spec/models/spree/app_configuration_spec.rb COPY
- core/app/models/spree/preference.rb COPY
- spec/models/spree/preference_spec.rb COPY

#### What should we test?
- verify that the app configuration is the same before and after the PR is deployed - you can check stripe config, fees on some ship and payment method calculators, general config and maybe some content config entries.
- after deploying, make sure you can change some of these configurations and they are reflected in the app
- maybe we should verify both stripe and paypal before and after the deploy of this PR.

#### Release notes
Changelog Category: Changed


Matt-Yorkley (approved):

filipefurtad0:
Hey @luisramos0 ,

Based on your suggestions, I thought I'd split this into two types of tests, on Stripe-Connect/SCA and Paypal:
1) settings and function of existing payment methods
2) settings and function of newly created payment methods


**1) testing settings and function of existing payment methods**

Before Staging the PR:
i) I considered a Hub with the previously set payment methods, made some changes on the settings and took notice of all the settings (transaction fees, description, etc.)
ii) placed three orders with using the these payment methods, taking notice of the emails and order totals

After Staging the PR

iii) checked whether all settings were kept - OK - settings after staging the PR are identical with those previously set
iv) repeated the same orders placed on the point ii) above and compared the order totals/fees with point ii) above - OK - orders display the same (and correct) order totals/fees


**2) settings and function of newly created payment methods**

After Staging the PR

i) payment methods Stripe-Connect/SCA and Paypal were created on an "fresh" enterprise, with different transaction fees, using different calculators

Verified that:
- connection with Stripe works
- Paypal API credentials are taken as valid
- settings are taking effect on the app

ii) placed three orders through the shopfront, with different products for each payment method
- simple card worked well for Stripe Connect
- 3D auth- card worked for StripeSCA
- payment with Paypal proceeded with no issues
- verified and took notice of the charged values for enterprise fees and order totals


iii) placed the same orders, for each payment method (except Paypal), in the backoffice with different products
- simple card worked well for Stripe Connect and SCA
- compared the order values with those obtained in the shopfront - all good.

I found no issues on these tests. This looks good to go :tada:

Matt-Yorkley:
#mergeallthethings

luisramos0:
wonderful!!! thanks Filipe!!!

PR #5877, opened 2020-08-07
Test BugsnagJS error catching

Matt-Yorkley:
WIP


app/assets/javascripts/darkswarm/services/checkout.js.coffee
```diff
@@ -24,7 +24,8 @@ Darkswarm.factory 'Checkout', ($injector, CurrentOrder, ShippingMethods, StripeE
           try
             @loadFlash(error: t("checkout.failed")) # inform the user about the unexpected error
           finally
-            throw error # generate a BugsnagJS alert
+            throw error
+            Bugsnag.notify(error)
```
mkllnk, line 28: I think this line is unreachable. It needs to be the other way round:
```suggestion
            Bugsnag.notify(error)
            throw error
```

Matt-Yorkley:
I've updated this a bit to include javascript unit tests for the bugsnag calls.

Matt-Yorkley:
It's a tricky one to test. I think it's a `dev-test`. 

- Put a bugsnag API key in `application.yml` with `BUGSNAG_JS_KEY:`
- Temporarily change line 8 in `app/views/layouts/_bugsnag_js.html.haml` so the `releaseStage` is set to `staging`
- Create some exceptions to test the response; an error in `checkout_controller.rb` is the simple one and then: a handled error in `app/assets/javascripts/darkswarm/services/checkout.js.coffee` around line 30 and an unhandled error around line 26...

PR #5878, opened 2020-08-07
[Bye bye Spree] Bring models shipping_method, shipping_rates, address and shipping_category from spree_core

luisramos0:
#### What? Why?
Continues #4826
- core/app/models/spree/address.rb MERGE WITH DECORATOR
- spec/models/spree/address_spec.rb keep in parallel to addresses_spec
- core/app/models/spree/shipping_calculator.rb - NO COPY :heavy_check_mark: 
- core/app/models/spree/shipping_category.rb MERGE WITH DECORATOR
- spec/models/spree/shipping_category_spec.rb COPY
- core/app/models/spree/shipping_method_category.rb COPY
- core/app/models/spree/shipping_method.rb MERGE WITH DECORATOR
- spec/models/spree/shipping_method_spec.rb MERGE WITH OFN VERSION
- core/app/models/spree/shipping_rate.rb COPY
- spec/models/spree/shipping_rate_spec.rb COPY

#### What should we test?
We need to verify if everything related to shipping methods and shipping rates and addresses is still working :warning: 
- make sure shipping methods keep config and behaviour before and after deploy
- shipping_method - test the backoffice for to configure shipping methods and checkout with a specific shipping_method
- verify you can checkout even if product.shipping_categories and shipping_method.shipping_categories dont match
- verify shipping rates are correctly calculated, I'd checkout and then tryu to change the shipping method on the backoffice
- address - we need to verify we can edit addresses across the backoffice: orders, customers and enterprises. I had a difficult time fixing tests related to addresses without state. Is this even possible in OFN to use addresses without defining the state? I think address always requires the state in OFN. We need to verify that.
- address on checkout - similar to tests in #5866 we need to verify the address forms on the checkout page are working correctly.

#### Release notes
Changelog Category: Changed


sauloperez (approved):

sauloperez:
I would fix the Code Climate issue `Redundant curly braces around a hash parameter.` others can wait IMO

luisramos0:
:+1: fixed

Matt-Yorkley (approved):

filipefurtad0:
Hey @luisramos0 ,

Went through the test cases you pointed out.

I've set up different shipping methods prior to staging this PR. Then, deployed the PR and verified that: 

- shipping methods keep config and behavior before and after deploy - OK

- configuring shipping (new) methods work. Checked out with a specific shipping_method - OK 

- verify you can checkout even if product.shipping_categories and shipping_method.shipping_categories dont match - OK

"Sepia" product was set to shipping_category "Frozen"; "boat delivery" shipping_method was set to "Default; Vacuum". Checkout worked fine. 

- verify shipping rates are correctly calculated. 

Shipping rates are applied correctly after checking out. Changing shipping method in the backoffice works well; Clicking "Update and recalculate fees" renders the correct shipping fee, adjusting the order total.


- address - verified that editing addresses across the backoffice (Shipping and Billing) address in the backoffice - in Orders, and Customers. Also, edited Enterprises addresses and verified the changes are seen in the map:

Different address
image

Same address
image



> Is this even possible in OFN to use addresses without defining the state? I think address always requires the state in OFN. We need to verify that.

I don't think this is possible. I all cases, I found a State to be a mandatory field:

image

image

image


- address on checkout - verify the address forms on the checkout page are working correctly - OK

I can't think of anything else to test here.

This looks good to go.

luisramos0:
awesome, thanks!

PR #5879, opened 2020-08-07
[Bye bye Spree] Bring models payment_method, credit_card and gateway from spree_core

### LINKED ISSUE #5649: Error while adding a payment method
## Description

While testing the release v3.0.0 an error was observed when adding a payment method. Snail was visible and this bugnsag error seems related:

https://app.bugsnag.com/yaycode/openfoodnetwork-uk/errors/5ec1a91652841000189980a2?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[app.release_stage][0][value]=staging&filters[app.release_stage][0][type]=eq

Reloading the page worked fine, and it was not possible to reproduce, or collect more information from the console. This was observed twice during a short time period, while setting up Stripe-SCA and Paypal payment methods.

## Expected Behavior

Payment method is added, when pressing Update.

## Actual Behaviour

Pressing Update resulted in error (visible snail) and an apparently related Bugnsag message above.

## Steps to Reproduce

1. In the admin dashboard go to Payment methods
2. Click add a new payment method, and fill out all the information
3. Press update
4. -> Snail

## Workaround

Refreshing the page worked. After that, all was normal.

## Severity

## Your Environment

* Version used: v3.0.0
* Operating System and version (desktop or mobile): Desktop/Ubuntu 18.04

## Possible Fix

A possible regression?

luisramos0:
#### What? Why?

Continues #4826
This can potentially close #5649

- core/app/models/spree/credit_card.rb MERGE WITH DECORATOR
- spec/models/spree/credit_card_spec.rb MERGE WITH OFN VERSION
- core/app/models/spree/gateway.rb MERGE WITH DECORATOR
- spec/models/spree/gateway_spec.rb COPY
- core/app/models/spree/payment_method.rb MERGE WITH DECORATOR
- spec/models/spree/payment_method_spec.rb MERGE WITH OFN VERSION
- core/app/models/spree/payment_method/check.rb COPY
- core/app/models/spree/gateway/bogus.rb COPY
- core/app/models/spree/gateway/bogus_simple.rb COPY

#### What should we test?
We need to verify if everything related to payments and payment methods is still working :warning: 
- make sure payment methods keep config and behaviour before and after deploy
- test the backoffice for to configure payment methods and checkout with **every payment method type**
- verify payment fees are correctly calculated
- verify payment collection in the backoffice **with every payment type**

#### Release notes
Changelog Category: Changed


filipefurtad0:
A quick note to remind that this PR may solve #5649 as well - something to double check while testing, although I could not reproduce this bug.

app/models/spree/payment_method.rb
```diff
@@ -70,14 +68,6 @@ def payment_source_class
       raise 'You must implement payment_source_class method for this gateway.'
     end
 
-    def self.available(display_on = 'both')
```
mkllnk: I think that this is still used, for example:

https://github.com/openfoodfoundation/openfoodnetwork/blob/49a60374e6f4c48202b1899b09b7c4a42fa0a027/app/models/enterprise.rb#L111

https://github.com/openfoodfoundation/openfoodnetwork/blob/49a60374e6f4c48202b1899b09b7c4a42fa0a027/app/models/enterprise.rb#L343

```
$ git grep  "\.available\>" -- app/helpers app/models lib
app/helpers/enterprises_helper.rb:    payment_methods = current_distributor.payment_methods.available(:front_end).all
app/helpers/order_cycles_helper.rb:      if options[:shipping_and_payment_methods] && (e.shipping_methods.empty? || e.payment_methods.available.empty?)
app/helpers/order_cycles_helper.rb:        if e.shipping_methods.empty? && e.payment_methods.available.empty?
app/helpers/order_cycles_helper.rb:        elsif e.payment_methods.available.empty?
app/models/calculator/flexi_rate.rb:    def self.available?(_object)
app/models/enterprise.rb:      merge(Spree::PaymentMethod.available).
app/models/enterprise.rb:    shipping_methods.any? && payment_methods.available.any?
lib/open_food_network/enterprise_issue_validator.rb:      @enterprise.payment_methods.available.any?
```

mkllnk (approved):
Ah, I see now. Great work!

sauloperez:
:tada: yay! I was so looking forward to this bit lately!

sauloperez (approved):

filipefurtad0:
Hey @luisramos0 

Thanks for pointing out the main areas where the code migration may have impact. Below, all "payment methods" means:

- Stripe SCA
- Stripe Connect
- Paypal
- Pin Payments
- Cash



Checking the topics you mention, I:

i) made sure payment methods kept their config and behavior before and after deploy. I had all payment methods available on OFN set up before staging, and checked them after deploying this PR. Settings were kept.

ii) tested the backoffice for configuring payment methods
I've set all payment methods from scratch - all good.

iii) ...and checked-out with every payment method type.
Stripe-SCA - OK
Stripe-Connect - OK
Paypal - OK
Pin-Payments - OK
cash - OK

iv) verified payment fees were correctly calculated
I've set a different flat percent on each of them and checked whether each order (from the previous step) had been charged with the correct transaction fee - all good.

v) verified payment collection in the backoffice, for every payment type
Placed orders in the backoffice for all supported payment methods in the BO
Stripe-SCA - OK
Stripe-Connect - OK
Paypal - not supported
Pin-Payments - not supported
cash - OK

Applied fees were correct in all cases.

vi) I had a particular look for errors while creating the various payment methods - no signs of #5649 

vii) checked /order_management/reports/bulk_coop: payments appear and transaction fees appear correctly in the report.

This looks good!
Let's merge.


PS: Please note that I did not test the MIGS payment method - I do not have credentials, never tested this one, to be honest.  Perhaps it would be a good time to assess how many users of this payment method we currently have? I found some discussion about this on Slack, but no conclusion. I think the best approach would be to open an issue to remove MIGS and gather data/feedback from ptoduct/instance managers. Thoughts?

luisramos0:
I am very happy to see this go through, thanks for the long tests Filipe!

Migs looks dead, I see some payments in AU from 2014 and a couple from 2018. None in UK and 1 failed payment in France from May 2020 (probably a test or an error).

PR #5883, opened 2020-08-07
[Bye bye Spree] Bring models tax_rate, tax_categories, adjustments and calculator from spree_core

luisramos0:
#### What? Why?
Continues #4826

- core/app/models/spree/tax_category.rb COPY
- spec/models/spree/tax_category_spec.rb COPY
- core/app/models/spree/tax_rate.rb MERGE WITH DECORATOR
- spec/models/spree/tax_rate_spec.rb MERGE WITH OFN VERSION
- core/app/models/spree/adjustment.rb MERGE WITH DECORATOR
- spec/models/spree/adjustment_spec.rb MERGE WITH OFN VERSION
- core/app/models/spree/calculator.rb MERGE WITH DECORATOR

#### What should we test?
We need to verify if everything related to taxes in OFN as well as do some basic checkes related to adjustments and calculators:
- re calculators, I'd verify enterprise fees with a specific calculator and make sure it's working as usual
- I'd test the backoffice configuration of tax categories and tax_rates. We should probably verify if the settings are being used correctly in the system (to be honest I dont know much about this area maybe it's a good opportunity to write some test cases and see where we are in terms of tax settings)

#### Release notes
Changelog Category: Changed


sauloperez (approved):

sauloperez:
I'm moving it to test ready because it's a fairly simple merge and it looks :+1:

filipefurtad0:
Hi @luisramos0,

I had a look at some specs around these models and the related features, and also related to the calculators for enterprise fees.

Regarding calculators in enterprise fees, it was verified that the different enterprise fees work for each specific calculator, making sure this was working as before:
- created all types of enterprise fees available.
- placed an order before and after the PR.
- compared the obtained values - identical, so all good.

In terms of specs, I found `enterprise_fee_spec.rb` which tests the basic operations of the /enterprise_fees page:
- listing, creating editing, deleting fees
- also checks whether a given enterprise can only perform these actions on the fees its fees (and not others, on the instance)

Since this seemed covered I just did a brief smoke test on CRUD operations on this page.


Backoffice configuration of tax categories and tax_rates

Checked that tax settings work as before, by comparing placed orders before and after the PR, for previously defined settings.

I found `tax_rate_spec.rb` but did not find any spec under /features, to test CRUD operations, using the Capybara/Chromedriver. So, maybe this would be a nice to have? A similar spec to `zones_spec.rb` and `enterprise_fee_spec.rb`, in which the UI actions are tested would perhaps be valuable - maybe worth opening an issue and adding to the Auto-tests column. What do you think? Also #4750 relates to taxes, but rather to reports.

I manually tested that:
- create, read, update and delete work as before, and have the expected result on the shopfront.

- Checked tax settings work correctly with the associated zones, i.e. removing or adding tax to checkout according the country settings

- Checked it's possible to associate a tax category to a product, individually

Looks good, moving to ready to go.

luisramos0:
nice Filipe.
It's great you can read the specs and see what is covered and not :ok_hand: that's a major advance. yes, I think it's a good idea to open issues to add auto specs to those pages, awesome :clap:

filipefurtad0:
Thanks for the support Luis!
I dug a bit more and did found some specs addressing these pages, but I think these could use some improvement - opened issue #6016.

Issue #5708, opened 2020-06-30
Bugsnag when printing of alternative-invoice of orders in cart state


filipefurtad0:
## Description

Bulk printing of orders in cart state generates errors 422 (seen in the console) and Bugsnag errors, across staging servers (both v3 and v2.1). The user sees an endless-spinner and several GET requests are sent.

In staging-ES:

![image](https://user-images.githubusercontent.com/49817236/86139994-4752a900-bae8-11ea-82e8-d650333cddd7.png)

https://app.bugsnag.com/katuma/katuma/errors/5ef35f60d643330019edb2c6?filters[event.since][0]=30d&filters[error.status][0]=open
```
ActionView::Template::ErrorBulkInvoiceService#start_pdf_job_without_delay
undefined method `to_date' for nil:NilClass Did you mean?  to_a18 seconds agoJun 30th, 14:40:35 UTC | staging |  -- | -- | --
```

In staging-AU

![image](https://user-images.githubusercontent.com/49817236/85574134-6952b400-b62e-11ea-8ea7-cfa303ed7366.png)

https://app.bugsnag.com/yaycode/openfoodnetwork/errors/5ef367c30c9a67001715b9dc?event_id=5ef367c3005b7b5abac50000&i=sk&m=nw

```
ActionView::Template::ErrorBulkInvoiceService#start_pdf_job_without_delay
undefined method `full_address' for nil:NilClass
```

- Observed while testing - https://github.com/openfoodfoundation/openfoodnetwork/issues/5561#issuecomment-648855149
- Reproduced by @luisramos0 - see : https://openfoodnetwork.slack.com/archives/C015WAZHF7C/p1593503633228200?thread_ts=1593436667.217800&cid=C015WAZHF7C

## Expected Behavior

Hide the print invoice button when showing incomplete orders.

## Actual Behaviour

Send a bulk print job to cart state orders triggers (many) errors above and leaves the user with an endless spinner.

## Steps to Reproduce

1. Login as superadmin
2. Go to /orders and untick the "Only show complete orders" box
3. Press Filter Results
4. Select an order in the Cart state
5. Open the browser console and press Print Invoices

## Workaround

You need to press escape or F5 to exit the endless-spinner.

## Severity

This is probably an edge case, but generates a lot of errors...
bug-s3: a feature is broken but there is a workaround

## Your Environment

* Version used: v2.1 and v3.
* Browser name and version: Firefox 77
[...]

Issue #5505, opened 2020-05-26
Error 500 - Customer viewing an edited order from which all items were removed


filipefurtad0:
## Description

After placement and payment capture of an order - but before shipment - the hub manager can edit the order and change the number of items, or remove all of them. If this is done and the customer attempts to view the order an error 500 become visible to the user, i.e., in /orders/Rxxxxxxxx. Looking at /transactions shows an error in the column Shipping status (pics below).

## Expected Behavior

Clicking on the edited order should not show the snail and:
- perhaps lead show the order with zero items, and zero amounts, with the corresponding Outstanding balance (would this be a paper cut?)
- the error message under Shipping Status should be improved rather than display the error message

## Actual Behaviour

- error 500 when visiting the edited order (from which nr. of items is now zero)
- [missing "en.spree.shipment_states.null" translation] error appears

## Steps to Reproduce

1. Customer - Place an order
2. Admin - edit the order, removing all items
3. Customer - go to /orders and click on the order -> snail 500
4. Customer - go to /transactions and observe the error 

## Animated Gif/Screenshot

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/2dc2c318-f945-42be-b6f1-30b75f616b36)

![image.png](https://images.zenhubusercontent.com/5d761b049dde650001eaf32f/1cc95a3f-1589-4d52-b1c5-ff5796f37ec7)

## Workaround

Deleting an order, instead of reducing it's numbers to zero.

## Severity

Placing as bug-s4. An argument for upgrading this to s3 would be that the user sees snail.

## Your Environment

* Version used: v2.9.9
* Browser name and version: Firefox 75
* Operating System and version (desktop or mobile): Desktop

filipefurtad0:
I think we can close this, as the only path to generating empty orders is by removing all items from orders, which cancels them automatically, and resuming the order.

However, [resumed orders don't appear to the customer](https://github.com/openfoodfoundation/openfoodnetwork/issues/7282), so as of now, this bug does not occur.

Issue #5500, opened 2020-05-26
[API] Orders endpoint only works in the browser and only after user login


luisramos0:
## Description
API authentication is not working in the api/orders api. IT only works if you login to the app on your browser and access the endpoint through that session.
USing api token and curl wont work.

## Steps to Reproduce

1.
2.
3.
4.

## Your Environment

* Version used:
* Browser name and version:
* Operating System and version (desktop or mobile):

PR #5887, opened 2020-08-08
[Bye bye Spree] Bring models order, line_item and other related from spree_core 

luisramos0:
#### What? Why?
Also included in this PR models: order_contents, inventory_unit, order_inventory, return_authorization, state_changes and tokenized_permission.
The best part of this PR is the test coverage, the order model is actually the best covered model in the Spree codebase :+1:

Continues #4826
- core/app/models/spree/inventory_unit.rb COPY
- core/app/models/spree/line_item.rb MERGE WITH DECORATOR
- core/app/models/spree/order_contents.rb COPY
- core/app/models/spree/order_inventory.rb COPY
- core/app/models/spree/order.rb MERGE WITH DECORATOR
- core/app/models/spree/return_authorization.rb COPY
- core/app/models/spree/state_change.rb COPY
- core/app/models/spree/tokenized_permission.rb COPY
- spec/models/spree/order_contents_spec.rb COPY
- spec/models/spree/order_inventory_spec.rb COPY
- spec/models/spree/order_spec.rb MERGE WITH OFN VERSION
- spec/models/spree/inventory_unit_spec.rb COPY
- spec/models/spree/line_item_spec.rb MERGE WITH OFN VERSION
- spec/models/spree/return_authorization_spec.rb COPY
- core/spec/models/spree/order/address_spec.rb MAYBE COPY
- core/spec/models/spree/order/adjustments_spec.rb MAYBE COPY
- core/spec/models/spree/order/callbacks_spec.rb COPY
- core/spec/models/spree/order/payment_spec.rb MAYBE COPY
- core/spec/models/spree/order/state_machine_spec.rb MAYBE COPY
- core/spec/models/spree/order/tax_spec.rb MAYBE COPY
- core/spec/models/spree/order/updating_spec.rb MAYBE COPY

#### What should we test?
We need to verify if everything related to orders and line_items (yes, everything basically) in OFN is working well.
This is a little heart surgery to OFN. We do need to trust the build here and do some good verifications:

#### Release notes
Changelog Category: Changed


app/models/spree/order.rb
```diff
@@ -720,7 +720,7 @@ def remove_variant(variant)
     end
 
     def cap_quantity_at_stock!
-      line_items.includes(variant: :stock_items).all.find_each(&:cap_quantity_at_stock!)
+      line_items.includes(variant: :stock_items).all.each(&:cap_quantity_at_stock!)
```
sauloperez, line 723: have you tried removing `all` and sticking to `find_each`?

sauloperez (commented):
O M G. I'd like to go again through the merge od the decorators because there's a lot going on but it's looking good. There's so much :hankey: :see_no_evil:

app/models/spree/order.rb
luisramos0: I will try now and watch the build :+1:

luisramos0:
yes :+1: I basically copy pasted things [here](https://github.com/openfoodfoundation/openfoodnetwork/pull/5887/commits/ba17768661904b766fe75f9544e7231547e0f32d) but this commit does deserve a detailed review by someone going through the original decorator and spree file and check every change here.

luisramos0:
rebased to resolve conflicts.

app/models/spree/order.rb
luisramos0: done :+1:  it works.

Matt-Yorkley (approved):

app/models/spree/order.rb
```diff
@@ -0,0 +1,915 @@
...
+
+    checkout_flow do
+      go_to_state :address
+      go_to_state :delivery
+      go_to_state :payment, if: ->(order) {
+        order.update_totals
+        order.payment_required?
+      }
+      go_to_state :complete
+      remove_transition from: :delivery, to: :confirm
```
sauloperez, line 27: now that we removed the :confirm state this line can be removed.

sauloperez (changes requested):
The only bit from the decorator I couldn't find is https://github.com/openfoodfoundation/openfoodnetwork/blob/427d1aaeff51234b8995351f555f1cfd7bb9bb2d/app/models/spree/order_decorator.rb#L32-L40. Where did that go?

app/models/spree/order.rb
luisramos0: good catch :+1: done

luisramos0:
That email related code was a workaround to disablethe default spree email validator. We no longer get it from spree and so we dont need it. We just need our simple validator :+1:

sauloperez (approved):
:clap: I'd be happy to approve it on Code Climate too. There are a few styling related issues in there that can wait IMO. It'd be too much for this PR.

sauloperez:
I made the call and approved it in Code Climate so we merge this ASAP.

filipefurtad0:
Allright, let's test the "OFN heart surgery" :-)
I'm happy for any advice, so before I start here I'd just ask if there is something specific to test - it looks like you were about to add something on the "what should we test" section above @luisramos0? Thank you!

luisramos0:
Hello Filipe. There's nothing specific about this PR. It's changing the files (although the code should be exactly the same) of the orders so it is really touching everything related to orders and stock levels.

I can mention a few things:
- order stock at cart level - add and remove items and see stock limits are respected. This can be done with and without overrides.
- verify stock levels are respected while editing orders in the backoffice - one edge case we can verify is to cancel an order and verify the stock is returned
- return authorizations - we should verify that these are working as before (to be honest I don't know if this is working properly before this PR)
- testing a fee with a weight calculator would be good and verify the final weight volume in the BOM works as expected
- we need some sanity checks in reports
These are just some ideas. I am sure you will find other good test cases to run here :+1:

filipefurtad0:
Hey @luisramos0,

As a first approach and as discussed, I followed what you suggested for test cases. I staged your PR in staging-ES and master branch on staging-UK: I found this to be the best way to compare my observations on each test case. So, if not mentioned otherwise, everything discussed below was confirmed as well in the latest release.

To summarize: so far, I found no differences whatsoever in behavior, between the two builds :tada: 

Let's see the test cases in more detail

**Stock Levels**
1) order stock at cart level - add and remove items and see stock limits are respected. This can be done with and without overrides.

Proceeded as follows:

i) admin: set a limited stock level for the supplier

customer:
ii) attempt to add more items to the cart than available -> modal displays stock limit
ii) added items to cart (below or equal to stock) -> stock levels do not change
iii) proceeded to checkout -> stock levels do not change
iv) checks out -> stock levels are reduced as expected

Repeated this after overriding the variant (i.e., after adding item to inventory and overriding stock level settings). Works as well.


 
**Stock Levels / BO**
2) verify stock levels are respected while editing orders in the backoffice - one edge case we can verify is to cancel an order and verify the stock is returned

Cancelling orders in the backoffice restores stock, to the inventory stock (if a variant override is set) or to the stock set by the producer (if no override is set).

While doing this, bumped into this issue of being able to place orders with stock levels higher than available. I'm pretty certain to have observed this bug before but I can't seem to find the issue. #5784 is related. Anyway, I found out, that the BO allows for values above the stock-set limit, which in turn may lead to these theese scenarios - see below - in which the invoice displays a different number of items and price charged than the actual order: 

image

image

image

Basically, to reproduce this, one needs to set the max number of items available in the order, and then, edit the backoffice-cart, one by one. I guess this deserves an issue.


Return Authorizations
3) we should verify that these are working as before

Tested return authorizations in the current v.3.2.10 - and it works! Also checked it on this PR, and found the same behavior: returning X items, gives back X items back to the stock.

However, I think I may have found a pre-existing bug here: if overrides are set for the returned product, then, the returned items go to the producer stock, and not to the inventory. This is not the case for cancelled orders, nor for backoffice order adjustments, in which the returned stock (be it due to cancellation or order adjustment) gets added back to the producer stock only if the variant is not overridden. Is this the expected behavior? I found nothing on this topic in the user guide. This might be something to as tomorrow at instance-managers.

**Fees and BOM**
4) testing a fee with a weight calculator would be good and verify the final weight volume in the BOM works as expected

This is a really interesting test case. To approach this, I placed an order with a previously set enterprise fee (per weight), and reduced the weight of the order (instead of lowering the number of items) in the bulk order management settings. This seems to work well. Here is the result, if one edits the order again after this: 

image


**Reports**
5) we need some sanity checks in reports

I verified all reports are displayed/rendered correctly. I took a particular look at order related ones, and found no issues. I took particular attention to whether newly placed orders are appearing correctly. 

**Subscriptions**
6) this is a particular type of order, and as such, I felt is should be checked, at least a basic sanity check. Since we had troubles with the sync/proxy orders lately, I opted for checking this manually instead of triggering the proxy order in the rails console.

Found no issues. The order was placed, the payment was processed, all (4) expected emails arrived. All good here. 

**Order-Related S3 Issues on the Pipe** - only checked after staging this PR
7) 

#2551 - reproduced
#4203 - could not reproduce as it is currently not possible to delete products from product list #5947 (reproduced)
#5505 - reproduced
#4186 - could not reproduce, due to the workaround introduced by PR#5253
#4220 - could not reproduce this
#4406 - could not reproduce this, seems to work fine? Will need to have a second look at this one.
#4582 - reproduced
#5236 - production test, did not reproduce
#5367 - reproduced
#5500 - did not test this one
#5708 - reproduced
#6029 - reproduced, related to #5367. I think the title may be need to be changed to "deactivates" instead of "deletes". Deleting (used) shipping methods is currently not possible #6135 (reproduced)
#6101 - reproduced

So, in summary, I found no difference in behavior between this PR and the current master. I think this heart surgery went very well :-D Looks ready to go!!

luisramos0:
Amazing work Filipe!! Specially the part where you try to reproduce all related bugs!

sauloperez:
This is definitely happening! if the order model is in our repo we're nearly there :tada: !

PR #5888, opened 2020-08-09
Allow US units on products/variants

andrewpbrett:
#### What? Why?

No issue has been created yet, but see [Discourse](https://community.openfoodnetwork.org/t/hubs-managers-can-choose-the-adapted-weight-and-measure-units-for-their-shops-given-their-own-local-situation/1289/11) for background.

The change addresses the issue that for many farms in the United States, not having an easy way to sell products by the pound/ounce is enough to dissuade them from using OFN. There is a currently a workaround to use "items" and a custom string but this makes it much easier to use. 

The Discourse post from @kirstenalarsen discusses four points; the first two are fairly straightforward and addressed by this PR. 

Point 3 is certainly possible, and it would be great if enterprises/instances could hide certain units if they never use them so that the dropdown has fewer options. I think it'd take a fair amount of refactoring, however, and possibly making a model out of Scale or Unit, but I don't know a ton about how instances are configured yet and how that config trickles down and is read in the codebase.

Point 4 seems like it might add more complexity than it's worth, however, and I agree that it's beyond "minimum viable feature".

Other items to address: 

As @luisramos0 noted further down the Discourse thread, it would be **really nice** to do some refactoring here... we have: 
1) two places where the units are defined, one in the javascript and one in ruby
2) similarly, two places where we have logic for the autoconversion of the units given a variant unit_value
3) lots of magic numbers floating around now (453.6 shows up a lot of different places, including tests, now and fat-fingering it somewhere would break things)

We could consider adding additional non-metric units for volume (pints, quarts, gallons, liquid ounces); this would strengthen the case/need for an instance/enterprise to be able to choose which units are available. 

I noticed that, for example, if a variant's unit_value is '1.0' (1g) and you change the product's variant_unit_scale from '1.0' (g) to '453.6' (lb), the variant's unit_value gets converted... this was surprising to me when I did it but I suppose it makes sense. And it's probably rare enough that someone would want to change the units on a product once they can create the product with the desired units in the first place. As an aside, it's taken me a while to get the hang of the admin interface for products/variants/units... if it's not on the design roadmap yet, I have some ideas for making it more clear! 

#### What should we test?
The test suite is (obviously) green as of 5680cc9e17308588b3347519e42cbc71d3f1e693 but there are still tests I'd like to add, particularly around the new javascript functions to test for bad/unexpected input from other files. 

We should also do some extensive sanity/smoke tests, particularly around shipping calculations, and likely add some automated tests to the suite there too. 

For example, @luisramos0 asked: "do shipping methods with weight calculators work for orders where the weight was changed in the bulk order management? Example, if I order two items of 0.25kg and then the hub manager uses Bulk Order Management to say the final weight was 400g, the shipping fee with cost 10eur per kilo will have to update from 5eur to to 4eur. This must work across units and scales." I **think** they're working properly, but I'm a little new to the codebase to know what working properly really looks like yet. 

#### Release notes
"Pounds and ounces are now available as units for products and variants"

Changelog Category: Added

Unrelated

#### Discourse thread
https://community.openfoodnetwork.org/t/hubs-managers-can-choose-the-adapted-weight-and-measure-units-for-their-shops-given-their-own-local-situation/1289/11

#### Dependencies
None that I know of.

#### Documentation updates
Probably! Will investigate and add soon.

### DIFF
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -4,15 +4,9 @@ angular.module("admin.products").factory "VariantUnitManager", ->
       'weight':
         1.0: 'g'
         1000.0: 'kg'
-        1000000.0: 'T',
-        # This appears to be what needs to be set in order for
-        # products to have a mass value stored in the database
-        # when they are created. However, it does not appear to
-        # change the existing product(s), so if the scale value
-        # is changed, a data migration may be necessary to make sure
-        # the proper unit X to grams actually works.
-        # TODO: ^^^ Delete this
-        453.592: 'lb'
+        1000000.0: 'T'
+        453.6: 'lb'
+        28.34952: 'oz'
       'volume':
         0.001: 'mL'
         1.0: 'L'
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -48,7 +48,10 @@ angular.module("admin.products").factory "VariantUnitManager", ->
         unitScales[0]
 
     @getUnitName: (scale, unitType) ->
-      @units[unitType][scale]['name']
+      if @units[unitType][scale]
+        @units[unitType][scale]['name']
+      else
+        ''
 
     @unitScales: (unitType) ->
       (parseFloat(scale) for scale in Object.keys(@units[unitType])).sort (a, b) ->
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -4,7 +4,15 @@ angular.module("admin.products").factory "VariantUnitManager", ->
       'weight':
         1.0: 'g'
         1000.0: 'kg'
-        1000000.0: 'T'
+        1000000.0: 'T',
+        # This appears to be what needs to be set in order for
+        # products to have a mass value stored in the database
+        # when they are created. However, it does not appear to
+        # change the existing product(s), so if the scale value
+        # is changed, a data migration may be necessary to make sure
+        # the proper unit X to grams actually works.
+        # TODO: ^^^ Delete this
+        453.592: 'lb'
       'volume':
         0.001: 'mL'
         1.0: 'L'
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -54,4 +54,7 @@ angular.module("admin.products").factory "VariantUnitManager", ->
       (parseFloat(scale) for scale in Object.keys(@units[unitType])).sort (a, b) ->
          a - b
 
-    @compatibleUnitScales
+    @compatibleUnitScales: (scale, unitType) ->
+      scaleSystem = @units[unitType][scale]['system']
+      (parseFloat(scale) for scale, scaleInfo of @units[unitType] when scaleInfo['system'] == scaleSystem).sort (a, b) ->
+         a - b
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -14,7 +14,7 @@ angular.module("admin.products").factory "VariantUnitManager", ->
         453.6:
           name: 'lb'
           system: 'imperial'
-        28.34952:
+        28.35:
           name: 'oz'
           system: 'imperial'
       'volume':
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -1,19 +1,35 @@
 angular.module("admin.products").factory "VariantUnitManager", ->
   class VariantUnitManager
-    @unitNames:
+    @units:
       'weight':
-        1.0: 'g'
-        1000.0: 'kg'
-        1000000.0: 'T'
-        453.6: 'lb'
-        28.34952: 'oz'
+        1.0:
+          name: 'g'
+          system: 'metric'
+        1000.0:
+          name: 'kg'
+          system: 'metric'
+        1000000.0:
+          name: 'T'
+          system: 'metric'
+        453.6:
+          name: 'lb'
+          system: 'imperial'
+        28.34952:
+          name: 'oz'
+          system: 'imperial'
       'volume':
-        0.001: 'mL'
-        1.0: 'L'
-        1000.0: 'kL'
+        0.001:
+          name: 'mL'
+          system: 'metric'
+        1.0:
+          name: 'L'
+          system: 'metric'
+        1000.0:
+          name: 'kL'
+          system: 'metric'
 
     @variantUnitOptions: ->
-      options = for unit_type, scale_with_name of @unitNames
+      options = for unit_type, _ of @units
         for scale in @unitScales(unit_type)
           name = @getUnitName(scale, unit_type)
           ["#{I18n.t(unit_type)} (#{name})", "#{unit_type}_#{scale}"]
@@ -32,8 +48,10 @@ angular.module("admin.products").factory "VariantUnitManager", ->
         unitScales[0]
 
     @getUnitName: (scale, unitType) ->
-      @unitNames[unitType][scale]
+      @units[unitType][scale]['name']
 
     @unitScales: (unitType) ->
-      (parseFloat(scale) for scale in Object.keys(@unitNames[unitType])).sort (a, b) ->
+      (parseFloat(scale) for scale in Object.keys(@units[unitType])).sort (a, b) ->
          a - b
+
+    @compatibleUnitScales
--- a/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
+++ b/app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
@@ -35,4 +35,5 @@ angular.module("admin.products").factory "VariantUnitManager", ->
       @unitNames[unitType][scale]
 
     @unitScales: (unitType) ->
-      (parseFloat(scale) for scale in Object.keys(@unitNames[unitType])).sort()
+      (parseFloat(scale) for scale in Object.keys(@unitNames[unitType])).sort (a, b) ->
+         a - b
--- a/app/models/product_import/unit_converter.rb
+++ b/app/models/product_import/unit_converter.rb
@@ -32,6 +32,10 @@ module ProductImport
       {
         'g' => { scale: 1, unit: 'weight' },
         'kg' => { scale: 1000, unit: 'weight' },
+        # We have _no idea_ what this is doing. It has units?
+        # And it maybe is connected to something related to shipping?
+        # TODO: DELETE THIS ^^^
+        'lb' => { scale: 453.592, unit: 'weight' },
         't' => { scale: 1_000_000, unit: 'weight' },
         'ml' => { scale: 0.001, unit: 'volume' },
         'l' => { scale: 1, unit: 'volume' },
--- a/app/models/product_import/unit_converter.rb
+++ b/app/models/product_import/unit_converter.rb
@@ -32,7 +32,7 @@ module ProductImport
       {
         'g' => { scale: 1, unit: 'weight' },
         'kg' => { scale: 1000, unit: 'weight' },
-        'oz' => { scale: 28.34952, unit: 'weight' },
+        'oz' => { scale: 28.35, unit: 'weight' },
         'lb' => { scale: 453.6, unit: 'weight' },
         't' => { scale: 1_000_000, unit: 'weight' },
         'ml' => { scale: 0.001, unit: 'volume' },
--- a/app/models/product_import/unit_converter.rb
+++ b/app/models/product_import/unit_converter.rb
@@ -32,10 +32,8 @@ module ProductImport
       {
         'g' => { scale: 1, unit: 'weight' },
         'kg' => { scale: 1000, unit: 'weight' },
-        # We have _no idea_ what this is doing. It has units?
-        # And it maybe is connected to something related to shipping?
-        # TODO: DELETE THIS ^^^
-        'lb' => { scale: 453.592, unit: 'weight' },
+        'oz' => { scale: 28.34952, unit: 'weight' },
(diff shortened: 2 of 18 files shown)


andrewpbrett:
Hrrrm, that test was passing on my [build](https://semaphoreci.com/andrewpbrett/openfoodnetwork/branches/imperial-auto-units/builds/5).... is there any chance that's a spurious failure?

Matt-Yorkley:
> is there any chance that's a spurious failure?

I restarted the build, and it's passing now :+1:

app/assets/javascripts/admin/products/services/variant_unit_manager.js.coffee
```diff
@@ -30,7 +48,16 @@ angular.module("admin.products").factory "VariantUnitManager", ->
...
     @getUnitName: (scale, unitType) ->
-      @unitNames[unitType][scale]
+      if @units[unitType][scale]
+        @units[unitType][scale]['name']
+      else
+        ''
 
     @unitScales: (unitType) ->
-      (parseFloat(scale) for scale in Object.keys(@unitNames[unitType])).sort()
+      (parseFloat(scale) for scale in Object.keys(@units[unitType])).sort (a, b) ->
```
Matt-Yorkley, line 57: Nice. I had to look this up, and apparently javascript `sort()` doesn't play nicely with numbers by default unless the compare function is given explicitly, as you've done here. :+1:

andrewpbrett: Haha yeah I had to do the same. Silly javascript.

andrewpbrett:
I put a couple tests in there to start; I'm still sort of sussing out the conventions around testing and what we want to write as unit tests... feedback welcome in that department :)

andrewpbrett:
On the bulk order management page, all adjustments to the weight of a line item currently have to be done in grams. It would be nice if the weights here were scaled in the same way that the line item's variant is scaled, to avoid having to convert things manually to grams. I don't know how much this functionality is used though, or if it's worth holding off on releasing the ability to have US units on products/variants until we can also have US units on this screen.

app/models/product_import/unit_converter.rb
```diff
@@ -32,6 +32,8 @@ def unit_scales
       {
         'g' => { scale: 1, unit: 'weight' },
         'kg' => { scale: 1000, unit: 'weight' },
+        'oz' => { scale: 28.35, unit: 'weight' },
+        'lb' => { scale: 453.6, unit: 'weight' },
```
luisramos0, line 36: this code is repeated from  units option_value_namer#scale_for_unit_value
this should be extracted to a service called Units that would contain these data structured with these values and strings.

lib/open_food_network/option_value_namer.rb
```diff
@@ -63,17 +62,34 @@ def option_value_value_unit_scaled
...
-
-      unit
+      scales = units[@variant.product.variant_unit]
+      product_scale = @variant.product.variant_unit_scale
+      product_scale_system = scales[product_scale.to_f]['system']
+      largest_unit = scales.select { |scale, unit_info|
+        unit_info['system'] == product_scale_system &&
+          @variant.unit_value / scale >= 1
+      }.max
+      largest_unit = units[@variant.product.variant_unit].first if largest_unit.nil?
```
luisramos0, line 90: this could be
   `   largest_unit = scales.first if largest_unit.nil?`

luisramos0: I made this method easier to read. please see my PR.

luisramos0 (approved):
Hi Andy, I have created a PR on top of your branch, I fixed some rubocop issues as I was going through the code:
https://github.com/andrewpbrett/openfoodnetwork/pull/1/files
I have also moved option_value_namer and variant_and_line_item_naming to app/services where they belong. There's a lot to clean up in this space, this is just a minor improvement so we keep the boy scout rule and do a little refactoring in every change. 

I m going to approve this PR as is (even better if you merge my PR into your branch :-)).
Anyway, imo this is good enough :tada: I am approving it.

app/models/product_import/unit_converter.rb
luisramos0: this is an extra improvement on top of my PR. We can skip this for now.

andrewpbrett: I was thinking the same, we could really use a refactoring into a single place for Units to live, and ideally have that be available to both the front end javascript/coffeescript files as well as the rails backend... but agreed it can be a separate effort.

lib/open_food_network/option_value_namer.rb
andrewpbrett: Thanks, I saw that in the PR, it looks good.

andrewpbrett:
Hi Luis, I merged your PR, thanks for the improvements. I'm going to work this week on the shipping calculator being in pounds and hopefully some of the other code cleanliness improvements we've talked about above, but I agree they can be in a separate branch and we can start having people test these changes as they are 🎉

jaycmb:
Regarding the 3 "problems" that shall be adressed in future PRs, from the conversation with @luisramos0 

1- Enterprise Fees: the weight calculator does not work correctly for enterprise fees with products in lbs

→ Hypothesis is that a weight-based enterprise fee is an rather less used option (vs flatrate etc), so a solution to this seems to be lowest priority  

2 - Shipping: weight calculator does not adjust to imperial
→ For evaluating how important it would be to implement a solution on this one, we would like to find out the share of shipment vs pickup option -> Metabase will help

3- BulkOrderManagement: final weight volume in grams - if we use lbs, in the BOM on the
final weight volume field the user will see the correct weight value in
grams and the user will be able to change this value (in grams) to
another value in grams and this is correctly picked by calculators

This it is quite confusing from a user perspective. Bc even if calculating correctly now, for a consistent flow would be necessary to have BOM displayed in the same measure unit as chosen for the product.
For assessing the importance though would be good to know how often this BOM is used?
Also wondering how often it is the case that users would have orders with different measure units, i.e. some products ordered in imperial, some in metric (for determining how the information should be managed in that column)

andrewpbrett:
Thanks @jaycmb - I had very similar thoughts on the three points! #3 seems to be the most important but also the most difficult of the three. What is Metabase (that you said would help with #2)?

jaycmb:
@andrewpbrett Metabase is not ready yet but we are working on this. I added that specific quetsion to the list of things we would like to get clarity on. Once Metabase is setup this should be easy to find out

andrewpbrett:
@luisramos0 I tried to reproduce the enterprise fee issue. Here's what I tried:

- Create an enterprise fee of $1/kg and add it to the Fredo's Farm Hub OC
- Change the price of a product (carrots) to be $3/lb
- When I view Fredo's store front, I see that the price of carrots is $3.45 - a price of $3 and a fee of $0.45. 

Screen Shot 2020-09-14 at 2 48 43 PM

That seems like it's correct to me (1lb = ~ 0.45kg). Were you seeing something else? You said you thought it was mistaking pounds for grams. 

If the calculation is correct, then I would agree that it's a pretty low priority, since there is a workaround and I doubt many enterprises are charging fees by weight. But we would have to look at the data to be sure.

sauloperez (approved):
This is awesome! thanks @andrewpbrett and @luisramos0 :clap: !

LGTM although we could add specific unit tests for `variant_and_line_item_naming.rb`. I'd make me feel safer if we ever refactor this logic. I'm happy to do that in a separate PR as according to Simplecov the methods we touched get covered already.

andrewpbrett:
Thanks for the review @sauloperez! And let me know if you want to collaborate on the refactoring PR.

filipefurtad0:
Hi @andrewpbrett ,

Thanks for pointing me out to the discourse post - very helpful indeed. I went through it and performed the following test cases:

i) expect lb and oz units to appear in 

- the product page (/admin/products), when creating a new product
- the product page (/admin/products), when bulk-editing an existing product
- when editing an order cycle

image

image

- in the shopfront:

image

- order confirmation emails

image


ii) expect unit conversion to work correctly, in bulk product edit page

From a previously created product in (1) Kg 
image

Interchanging from kg to lb and to oz works:
image

image

Converting between lb and oz works as well, ex:

image
 
iii) expect product import to accept new units in the field 

image

image

This currently fails for **oz** and **lb** - I verified it works correctly for **kg** and **g** currently, so I'll open a new issue on this regard.


iv) expect to be able to place orders containing items with units of mixed measurement systems

Works fine and units appear displayed correctly.

v) expect **enterprise fee weight calculator** to work correctly with lb and oz

Verified that enterprise fee weight calculator only includes the kg unit:

image

kg product: 1kg
oz product: 16 oz
lb product: 2.2 lb

image

All good here, I think. What did you find not to be working before @luisramos0 ?   

vi) expect **shipping fee weight calculator** to work correctly with lb and oz

Verified that shipping fee weight calculator only includes the kg unit
image

[9.9792 kg ( = 2.2 lb) + 4.536 kg ( = 16 oz) ] x 2 UK-pounds per kg = 31,0304 

image

Discrepancy in values comes from rounding.  

Altogether:

image


vii) edit orders on **Bulk Order Management** - changing the amount in grams and recalculating fees on the orders to y

The Weight/Volume column shows the amount in grams: 
 
image

After reducing all the items to half:

image

And editing the order one sees:

image

There is a credit owed, respective to the reduction in the amounts in both the enterprise fees and the shipment fees. 


Asides from the import issue - test case iii) - I think this is good to go.
I would propose to merge this, and solve this issue in a separate PR.
Thoughts @luisramos0 @RachL ?

Ounces (oz) and pounds (lb) are introduced in OFN! :tada: 
Thank you so much Andy.

luisramos0:
Thanks Filipe.
re enterprise fees, maybe my test was not correct or maybe it was fixed meanwhile? Have you tried using a different fee value other than 1eur?

filipefurtad0:
You are right @luisramos0:
I didn't, and thinking of it, 1 is quite an unwise value to perform a unique test. Thanks for pointing this out.

I've changed the enterprise fee to 2.5 UK-pounds/kg. It looks correct as well:

image

luisramos0:
awesome, I tested again with sample data locally and enterprise fees are working! I am not sure what happened, it doesnt matter :+1: 

There's a PR for the shipping and payment methods calculators to support pounds so, it will soon be just BOM on the list of problematic things for imperial units :tada:

jaycmb:
Just to be sure that I understood correctly:

Both enterprise fee weight calculator and shipping weight only includes the kg unit (so I as as a enterprise user can **not** choose an imperial unit for enterprise fee or shipping weight, for products I created with imperial units), **but** the units are converted so the amounts are calculated correctly?

And has been an issue created already for BOM?

luisramos0:
Hello, @jaycmb that depends if you are talking before or after https://github.com/openfoodfoundation/openfoodnetwork/pull/5998
@andrewpbrett have you tested Enterprise fees on 5998, will we have enterprise fees that can be set in lbs as pat of 5998?

I think we need a new issue for the BOM case.

andrewpbrett:
Yes, I think you understand it correctly @jaycmb - this PR doesn't address those. However, as part of #5998 enterprise fees can be set in pounds. 

(I did just notice that the UI for setting them up looks a bit off, I'll submit a change to address this):
Screen Shot 2020-09-21 at 7 40 05 AM

I'll create an issue for BOM 👍

jaycmb:
Another concern that came up again in today´s product curation meeting:
might be confusing for users to see a long list of units to choose from.
Could we have some kind of "simple" option (select, toggle..) for enterprises to choose if they want to have imperial, metric or both?

luisramos0:
yes, that's been discussed before :+1: I am not sure where... 
we can have that toggle, we would need an issue for that.
My main concern is what happens if the user selects imperial, creates products with imperial units and then switches to metric? The simplest solution would be, when user switches to metric, to keep using imperial for the products that have imperial units but only show metric units on the drop downs to select units?

andrewpbrett:
That's what I was envisioning as well. I was picturing long term having each enterprise able to choose from a list of units available (I was thinking a multi-tick-box UI, maybe even separated by imperial/metric). Then a shop could even choose to ignore units they never use, like grams or ounces. And also an instance could have in its configuration the default choices for a new shop on that instance. I think the behavior Luis describes is correct for what should happen if a product is in a unit that's no longer in the list of units available - namely, nothing should happen until the user explicitly changes them. 

Here's the issue I created for it: https://github.com/openfoodfoundation/openfoodnetwork/issues/6082

luisramos0:
ok :+1: sounds good, but I think this is out of scope for the original wishlist item: https://community.openfoodnetwork.org/t/hubs-managers-can-choose-the-adapted-weight-and-measure-units-for-their-shops-given-their-own-local-situation/1289

I'd say that instead of an issue #6082 it could be a new wishlist item on discourse to define how we are going to build this extra config and also ask how important is it/prioritize: it needs some inception I think.

andrewpbrett:
Got it; I'll close the issue for now and move its contents here: https://community.openfoodnetwork.org/t/allow-enterprises-to-choose-what-available-units-for-products-variants/2060

jaycmb:
@luisramos0 not sure if I understand the concern: 

"what happens if the user selects imperial, creates products with imperial units and then switches to metric?" 
I envisioned in the most simple solution - that is to solve the potential problem of confused users -  the select should only affect the number of unit options **displayed** when creating new products, i.e. if enterprise wants to reduce number of options, they won´t be displayed in their unit size list (nor weight calculator & BOM), but it does not affect products that have been created already, i.e. they won´t be **converted** and just stay as they are

But that raises another questions, since we now discuss basically 2 different features:

1. Allow user to hide / display options imperial vs metric or both when creating products
2. Allow user to convert created products from one unit type to other

Was 2) tested and is this possible currently?


@luisramos0 point on #6082 being a new wishlist item: for a more sophisticated solution yes, that needs some inception work. 
But for now, simple solution has been "prioritized" in the product curation meeting.
So if @andrewpbrett could work on the Multi-Select: 
- Imperial 
- Metric
to solve 1) 

(Think of for example a user in France, who won´t ever sell in oz just doesn´t wanna be bothered by those weird imperial units ;) (S) he should be able to hide them)

andrewpbrett:
Agreed that something like a multi-select is the first priority. 

For your question about conversion: Right now if I have a product variant that is 1 kg and change the product's units to pounds, it will automatically change the variant to be ~2.2 lb. However, as @tschumilas [points out](https://community.openfoodnetwork.org/t/allow-enterprises-to-choose-what-available-units-for-products-variants/2060/2?u=apb), there's some "going further" work we could do here to encourage/allow different shops that might have different units to share a hub and have all their units show up the same without each farm having to recalculate.

tschumilas:
Seriously -  I had NO idea on this -  if I enter a product in pounds, and I change that product's units to grams,   it will auto-calculate?  OMG -  I can't wait to try this.  Mega feature for Canada!!!!

PR #5889, opened 2020-08-09
Suppress Selenium warnings in test log output.

Matt-Yorkley:
Committing this spec helper suggestion from @mkllnk as agreed in Slack, so we can finally return to having usable test output when running feature specs in dev :tada:

I think this will help with the Rails 4.1 upgrade too.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -181,6 +181,10 @@ RSpec.configure do |config|
 
   config.include JsonSpec::Helpers
 
+  # Suppress Selenium deprecation warnings. Stops a flood of pointless warnings filling the
+  # test output. We can remove this in the future after upgrading Rails, Rack, and Capybara.
+  Selenium::WebDriver.logger.level = :error
+
   # Profiling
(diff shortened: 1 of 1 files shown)


luisramos0:
:+1:

spec/spec_helper.rb
```diff
@@ -181,6 +181,10 @@ def restart_driver
 
   config.include JsonSpec::Helpers
 
+  # Suppress Selenium deprecation warnings. Stops a flood of pointless warnings filling the
+  # test output. We can remove this in the future after upgrading Rails, Rack, and Capybara.
+  Selenium::WebDriver.logger.level = :error
```
mkllnk, line 186:
```suggestion
  if Rails::VERSION::MAJOR == 4 && Rails::VERSION::MINOR == 0
    Selenium::WebDriver.logger.level = :error
  else
    ActiveSupport::Deprecation.warn(
      "Suppressing Selenium deprecation warnings is not needed any more."
    )
  end
```

PR #5890, opened 2020-08-10
fix method to use .map{ }.sum to calculate total fees correctly

### LINKED ISSUE #5632: [Rails 4.1] Fix spec/models/spree/adjustment_spec.rb (4 broken specs)
Broken specs in this build
https://semaphoreci.com/openfoodfoundation/openfoodnetwork-2/branches/pull-request-5611/builds/2

With this error:
```
 17) Spree::Adjustment recording included tax EnterpriseFee adjustments when enterprise fees have a fixed tax_category when enterprise fees are taxed per-item when the tax rate does not include the tax in the price records the tax on TaxRate adjustment on the order
      Failure/Error:
        per_item_fees_total = order.line_items.sum do |line_item|
          calculator.per_item_enterprise_fee_applicators_for(line_item.variant)
            .select { |applicator|
              (!applicator.enterprise_fee.inherits_tax_category && applicator.enterprise_fee.tax_category == rate.tax_category) ||
                (applicator.enterprise_fee.inherits_tax_category && line_item.product.tax_category == rate.tax_category)
            }
            .sum { |applicator| applicator.enterprise_fee.compute_amount(line_item) }
        end
      
      ArgumentError:
        wrong number of arguments (given 1, expected 2..3)
      # ./app/models/spree/calculator/default_tax_decorator.rb:19:in `compute_order'
      # ./app/models/spree/tax_rate_decorator.rb:15:in `adjust_with_included_tax'
      # ./spec/models/spree/adjustment_spec.rb:195:in `block (7 levels) in <module:Spree>'
```

Replacing sum with map and sum, fixes the specs:
image

This needs to be done after #5613 is merged to avoid conflicts.

jeduardo824:
#### What? Why?

Closes #5632 

The method to calculate the total fees per item was not working because the sum method changed on Rails 4.1. I changed it to work properly using `.map{ }.sum` instead of just `.sum`.

#### Release notes

Fix method `per_items_fees_total` on `app/models/calculator/default_tax.rb` to sum fees of the items correctly.

Changelog Category: Fixed

### DIFF
--- a/app/models/calculator/default_tax.rb
+++ b/app/models/calculator/default_tax.rb
@@ -49,11 +49,11 @@ module Calculator
     # Finds relevant fees for each line_item,
     #   calculates the tax on them, and returns the total tax
     def per_item_fees_total(order, calculator)
-      order.line_items.sum do |line_item|
+      order.line_items.map do |line_item|
         calculator.per_item_enterprise_fee_applicators_for(line_item.variant)
           .select { |applicator| applicable_rate?(applicator, line_item) }
           .sum { |applicator| applicator.enterprise_fee.compute_amount(line_item) }
-      end
+      end.sum
     end
 
     def applicable_rate?(applicator, line_item)


Matt-Yorkley:
@luisramos0 I've had a look through the API to try to get some answers...

The deprecation message we're seeing when using `#sum` on AR objects is added in `4.0`, but it only applies to calls that pass a block, eg: `products.sum(&:prices)`. The method can be given a symbol of the column name and still works as intended (it's not removed in Rails 4.1, and is still present in Rails 6.0). So in some cases we can do: `products.sum(:prices)` instead, but with the caveat that the argument must be an attribute of the object and not a method. I think in these cases the performance will be better, as it's designed to work with the relation to efficiently pull only the attributes that are needed when the query executes (it's actually `ActiveRecord::Calculations#sum` being called in this case, otherwise it's `Array#sum` if we use `#to_a` first). See [here](https://apidock.com/rails/v4.0.2/ActiveRecord/Calculations/sum) and [here](https://apidock.com/rails/v4.2.1/ActiveRecord/Calculations/sum) for reference.

In various places in the Rails source code itself where a method must be called on an AR relation, `#to_a` is called before the method. It's pretty ubiquitous. See this commit message in Rails 6: https://github.com/rails/rails/commit/b644964b2b555798fc4b94d384b98438db863b3f that basically says "we're already calling `#to_a` on everything anyway, so lets make AR objects work directly with `Enumerable` methods for simplicity". So I'd assume we should prefer `#to_a` over `#map` in these cases.

**TLDR version:**

- if we're summing an attribute of the objects, we should remove the `&` and do: `relation.sum(:attribute)`. This will perform better, but can't be used in all cases.
- if we're summing something that involves a method call on the objects, we should use `relation.to_a.sum(&:method)`.

luisramos0:
awesome, Matt. thanks for the investigation.
I was aware of the first part, #5611 only changes the calls that are not DB columns :+1: 
It's great we have the approach explained :ok_hand:
@jeduardo824 you can either close this PR in favor of #5611 or it would be greatly appreciated if you could pick up commits and related changes from #5611 into this PR (I'll have to break 5611 in separate PRs anyway).

jeduardo824:
I can pick up the commits, no problems @luisramos0

luisramos0:
Hey @jeduardo824 meanwhile I have already done so and the PRs are now ready for review: #5919 and #5921 

This PR can now be closed. Thanks for your help!

jeduardo824:
👍

PR #5899, opened 2020-08-12
Update dashboard link to point to enterprise shop

andrewpbrett:
#### What? Why?

The link to "your profile" on the admin dashboard currently points to the full map for your OFN instance, without zooming to the shop. I believe there's some work going on to revise the way the map works; this is a band-aid fix to at least show the user the public view of their shop. 

#### What should we test?
Visit the admin dashboard for a single enterprise user and click on Your Profile Live

#### Release notes
Bug fix: the link to Your Profile on the admin dashboard now takes you to your shop's profile, instead of the full map. 

Changelog Category: Fixed

None

#### Discourse thread
https://community.openfoodnetwork.org/t/admin-link-to-profile-store-goes-to-the-full-map-instead-of-the-store/2010

#### Dependencies
None

#### Documentation updates
None

### DIFF
--- a/app/views/spree/admin/overview/single_enterprise_dashboard.html.haml
+++ b/app/views/spree/admin/overview/single_enterprise_dashboard.html.haml
@@ -37,10 +37,8 @@
   .alpha.seven.columns.dashboard_item.single-ent#map
     .header
       %h3
-        %span.icon-map-marker
+        %span.icon-user
         = t "your_profil_live"
-      %p
-        = t "on_ofn_map"
     .list
       %a.button.bottom{href: main_app.enterprise_shop_url(@enterprise), target: '_blank'}
         = t "see"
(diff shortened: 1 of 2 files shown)


app/views/spree/admin/overview/single_enterprise_dashboard.html.haml
```diff
@@ -37,13 +37,10 @@
   .alpha.seven.columns.dashboard_item.single-ent#map
     .header
       %h3
-        %span.icon-map-marker
+        %span.icon-user
         = t "your_profil_live"
-      %p
-        = t "on_ofn_map"
```
Matt-Yorkley, line 43: If we're deleting this line, the key should be removed from `config/locales/en/yml` at the same time, as it's not used anywhere else in the app

andrewpbrett: Good catch, done

filipefurtad0:
Hey @andrewpbrett !

Thanks for this fix,

Previously, the admin was being redirected (somewhere) to the map, but after this PR we can see that the hyperlink appears, when hovering the cursor over it:

image


And redirection to the respective shop occurs, when clicking.

All good, ready to go :tada:

Issue #5897, opened 2020-08-12
Stripe payment fails but no record in OFN


emilyjeanrogers:
## Description
A regular customer of an Aus Enterprise has had a credit card declined multiple times when attempting to place orders over two order cycles, but there are no complete or incomplete orders in OFN for the order cycles in question and no failed or blocked transactions that didn't complete in the Enterprise's Stripe Account

## Expected Behavior
The customer should be able to process a credit card payment.
There would be a record of an incomplete order in OFN. 
There would possibly be a record of a failed transaction in the Enterprise's Stripe account.

## Actual Behaviour

Customer ID #4981 of Prom Coast Food Collective (#1965) has reported a declined credit card when attempting to place orders for the **3rd Sunday August** Order Cycle and the **4th Sunday July** Order Cycle. 

The customer first attempted to checkout on Friday July 17th and was unable to use their credit card. The customer provided screen grabs of the products she had selected in the checkout and details of the credit cards (3 in total) that she tried and the error messages she received:

ING debit card (Mel): your card has been declined (the card that has been used successfully previously)
ING debit card (Luke): your card does not support this type of purchase
ANZ credit card (Mel):  your card has been declined.

At that time the Enterprise placed her order manually, using their own credit card, but the customer still can't place orders. They have used the first ING credit card successfully in the past on OFN. 

The second time the customer attempted to check out was during the 3rd Sunday August Order Cycle.
- There is no complete or incomplete order in OFN for that order cycle, 
- There are no failed or blocked transactions that didn't complete in the Prom Coast Stripe Account

When comparing order numbers for this order cycle to order numbers in Stripe, the following Order number doesn't return any results in OFN - R674447571. This customer for this order tried to pay 8 times before they gave up.

![Screen Shot 2020-08-11 at 5.25.02 pm.png](https://images.zenhubusercontent.com/5ee2cad245bd18f8b0807374/3d76fd3d-1fad-4014-8a96-2381dd624b45)
![Screen Shot 2020-08-11 at 5.24.49 pm.png](https://images.zenhubusercontent.com/5ee2cad245bd18f8b0807374/008f6ac5-7d0c-4e90-8fb4-8fbe778de903)

Another customer was also unable to place orders for the same enterprise on August 3rd. This was logged as a separate Issue https://github.com/openfoodfoundation/openfoodnetwork/issues/5862

## Steps to Reproduce
Unable to Reproduce

## Workaround
The customer cannot place orders.

## Severity
[...]

Issue #5862, opened 2020-08-05
Multiple Customer entries in Stripe but Order was not processed


emilyjeanrogers:
## Description
There are multiple entries in Aus Stripe Connect account for a Customer who tried to create an order on OFN. Customer says they had trouble processing payment and were unable to place an order with Prom Coast Food Collective #1965

## Expected Behavior
Order created, payment processed

## Actual Behaviour
Customer made 7 attempts using one credit card and 1 attempt using another credit card between 9.28am and 9.39am on August 3rd. No fraud report in Stripe.

There is an incomplete order record for this transaction on OFN, with a zero amount

## Steps to Reproduce
Unable to reproduce

## Animated Gif/Screenshot

Customer listed 8 times in Stripe Connect:
![Screen Shot 2020-08-05 at 10 12 03 AM](https://user-images.githubusercontent.com/47620442/89361165-e2edbf80-d70d-11ea-9144-0453fd935113.png)

Customer order is incomplete, in Cart state, with a zero balance:
![Screen Shot 2020-08-05 at 11 03 13 AM](https://user-images.githubusercontent.com/47620442/89361182-f0a34500-d70d-11ea-9d1e-a474c616fd9e.png)

Order is empty:
![Screen Shot 2020-08-05 at 11 19 29 AM](https://user-images.githubusercontent.com/47620442/89361234-1deff300-d70e-11ea-9d51-b6a930b68ab8.png)

No results in User Report on OFN:
![Screen Shot 2020-08-05 at 10 41 01 AM](https://user-images.githubusercontent.com/47620442/89361275-411aa280-d70e-11ea-9716-8edfb0c8866a.png)

## Workaround
The hub can manually create an order for this customer. 
We can manually monitor the Stripe Connect account to identify future occurrences

## Severity

S2

https://github.com/openfoodfoundation/openfoodnetwork/wiki/Bug-severity
-->

## Your Environment

* Version used:
* Browser name and version:
* Operating System and version (desktop or mobile):

RachL:
@luisramos0 @kirstenalarsen could this be similar to https://github.com/openfoodfoundation/openfoodnetwork/issues/5785 ?

kirstenalarsen:
I don't know @RachL - I suspect somehow it is. I didn't monitor Stripe to see if my attempts on that day were picked up in the customer list. 

[...]

Issue #5080, opened 2020-03-26
User can not connect or create account


RachL:
## Description
I got several users in the past day with the same patern:

They go to the platform (recurrent user or new one) and they see this:

![image](https://user-images.githubusercontent.com/6525576/77627620-77b60480-6f47-11ea-8175-4e10d21f6a66.png)

The red button does not work.

A Ctrl+R fixes the display but the red button still does not work for them. This is the console of a user on the latest Chrome:

![image](https://user-images.githubusercontent.com/6525576/77627689-8ac8d480-6f47-11ea-86ef-0ea9f5267b3f.png)

I'm putting s2 even if there is a workaround because I'm afraid we are loosing new customers without knowing it (not everyone contact us)

## Steps to Reproduce
??

## Workaround

I unblocked user so far in 2 ways:
- using their phone if they are customers
- downloading another browser if they are an enterprise

## Severity
bug-s2: a non-critical feature is broken, no workaround

## Your Environment
* Version used: v2.8
* Browser name and version: Chrome
* Operating System and version (desktop or mobile): Desktop

lin-d-hop:
UK have also had a report of a user not being able to sign up. Clicking the red button did nothing.

The user reported using the most up to date version of Firefox (they were tech saavy enough to check their version).

Definitely an s2.

Matt-Yorkley:
Ok, so we're checking one issue related to browser extensions injecting javascript in a suspicious way for two French users, but we have some reports from UK and AUS as well.

I can't replicate this in any browsers myself.

This could be caused at any time by our third-party javascript dependencies failing to load for any reason, namely the Google maps and Stripe javascript we load from external sources. If there's an issue there (poor network connection on the user side, some availability issue with connecting to those servers, etc) then the above detailed issue will definitely happen, the rest of our javascript will break. I have seen this once or twice in `dev` environment when my internet connection was cutting out.

luisramos0:
did you notice the first js error is a stripe error Matt? Content-Security-Policy
It doesnt look like a network speed issue, does it?
I see some pages where people add a CSP meta to the html to allow for this cases (unsafe-inline).
I wonder why is this happening only for some people? could be browser and its version?

on another line, one issue coming out of this could be "make the website work even if maps and stripe URLs are totally down". should be an easy one and increase reliability of the webapp I think.

kirstenalarsen:
[...]

PR #5905, opened 2020-08-14
Load previously lost decorator code and don't delete orders

mkllnk:
#### What? Why?

While investigating #5897 I found some code that could accidentally delete an order. But then I found that the code wasn't loaded and could not have caused a disappearing order.

I continued to investigate the issue of unloaded code because it could be responsible of many random issues. The deeper I dug, the clearer I could see that the unloaded code was mostly insignificant. Nonetheless, I addressed the loading issue and removed the code that could delete orders.

The non-loading was mostly no issue because our code didn't change Spree's logic. The only exception are the changes from 20f610fbee4f6a9e22bb77e3baa7fc6cb684a628 which were lost since v3 and now brought back.

##### Parallel browsing

* As guest user, go to a shop and add items to the cart.
* Open a different browser (or a private window or a different device) and do the same there with different items in your cart.
* The two carts should be persisted and independent.
* Log in with the same user in both browser sessions.
* Both sessions should keep their individual orders.
* Go through checkout with the first browser session.
* The first browser should now show an empty cart again.
* The second browser should still have its old cart.

#### Release notes


lib/open_food_network/model_class_from_controller_name.rb
```diff
@@ -1,8 +0,0 @@
-module OpenFoodNetwork
-  module ModelClassFromControllerName
-    # Equivalent to CanCan's "authorize_resource :class => false" (see https://github.com/ryanb/cancan/blob/master/lib/cancan/controller_resource.rb#L146)
-    def model_class
-      self.class.to_s.sub("Controller", "").underscore.split('/').last.singularize.to_sym
-    end
-  end
-end
```
luisramos0, line 8: :heart:

spec/lib/spree/core/environment_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Spree::Core::Environment do
+  # Our version doesn't add any features we could test.
+  # So we just check that our file is loaded correctly.
```
luisramos0, line 7: :+1:
Nice one.
I think I moved the file to remove the stock_splitters attribute.

app/controllers/base_controller.rb
```diff
@@ -1,3 +1,7 @@
+require 'spree/core/controller_helpers'
```
luisramos0, line 1: I dont think we need this line and line 9 below.
the 4 helpers in ControllerHelpers are included individually.

luisramos0:
Note: on all investigations I have done related to failed payments I only found missing entries on spree_payments. spree_orders were always there. So:
- option 1 - the payments issues I investigated are different from  #5897 where the order is missing
- option 2 - the order.destroy that we are removing here could, in some cases, destroy the payments but not the order?
I'd say option 1 is more likely and this is a different problem from what we have done in #5080 for example.

luisramos0:
I'd mark this as closing #5897 :muscle: 
I'd merge this, close #5897 and then reopen 5897 if we do have any other occurrences.

luisramos0:
I think this could also have an impact on #5862 
I think we need to get this PR in the next release, ideally merge before this Thursday.

mkllnk:
Thanks for your early review. I'll continue my work today. I should have mentioned on Friday that I left it as draft because I didn't verify all my claims. I didn't go through the testing steps myself, they are just guesses from what I saw in the code. At the weekend, I also remembered that some controllers did load the helper file (CartController and Spree::OrdersController). So the deletion of old carts could have happened when logging in on the cart page, maybe? I'll try to replicate.

luisramos0:
ok :+1:
Will you manage to finish this tomorrow?
I think this PR, as is, can fix the bug and is harmless so I think we should go ahead with it in this weeks release :+1:

luisramos0:
Hey @mkllnk I think now I remember in delivery train meeting you suggesting I pick this one up so you can work on mobile?
I am correct? 
If so, what exactly do you think still needs to be done here? Can we just move this to Test Ready?

mkllnk:
I needed to verify the test instructions and found that one part of it was invalid. I added another two commits, addressing your feedback, @luisramos0, and removing the dead code. There is more I can clean up but I will do that in a separate PR.

mkllnk:
@luisramos0 I added two more commits in a separate PR: #5927 If you think it's better to include them here and we can still make it for the release, that would be awesome but I didn't want to put anything in the way of this PR getting into the release.

lib/spree/core/controller_helpers/order.rb
```diff
@@ -68,18 +68,9 @@ def associate_user
...
         end
 
-        # Recover incomplete orders from other sessions after logging in.
+        # Load current order and create a new one if necessary.
         def set_current_order
-          return unless (user = spree_current_user)
-
-          last_incomplete_order = user.last_incomplete_spree_order
-
-          if session[:order_id].nil? && last_incomplete_order
```
luisramos0, line 78: I am not sure about this statement:
"Every page load creates a cart order if none is present."

I thought this method was the place where that was happening, right?
I thought that this conditional was recovering existing carts if no order is identified in the session.
By doing this I think we are just increasing the number of past carts for the users and even maybe loosing resetting the cart between sessions.

My test would be: login to OFN and add something to the cart.
Then, open an icognito tab and login again, do you see your cart as it was?

luisramos0: Yeah, my hypothesis is confirmed by my test.
I run this test above in master and with this PR and in master the cart is preserved into the incognito tab after login, with this PR, after login in the incognito tab, the cart is empty.
We need to revert this change and keep this conditional.

luisramos0: I reverted the last commit so we can move this PR forward now.

Matt-Yorkley: Sounds good to me

Matt-Yorkley (approved):

lib/spree/core/controller_helpers/order.rb
luisramos0: ok, awesome, moving to Test Ready now.

mkllnk: Okay, great. I tried to test that and I couldn't find a way to recover an old cart. Can you share your steps?

luisramos0: just these as described above:
- login to OFN and add something to the cart.
- Then, open an icognito tab and login again, do you see your cart as it was?

filipefurtad0:
Hi @mkllnk ,

Thanks for this investigation and the detailed testing-steps!

Since master branch appears to be broken (as discussed here), I have staged the build used for release testing (v3.2.3), before testing this PR. The images below show session 1 on (the left) and session 2 (on the right side). I followed your procedure:


i) As guest user, went to a shop and added 5 items to the cart.
ii) Repeat the step above, on a different browser/private window. (I found no to play a role, if they are the same items or not)

At this point, both guest sessions have 5 items each. Proceeded:

iii) Logged in with the same user in both sessions

Here it became clear, that the condition
"- Both sessions should keep their individual orders." - is not verified.

The second session, now contains 10 items, the items from session 1 plus its own:

image


iv) At this point, attempting to checkout on session 1 generates the error:

image

So,
- "The first browser should now show an empty cart again." - is not verified
-  "The second browser should still have its old cart." - it still does, but it contains the items from session 1 as well.

v) checking out with session 2 (which contains 10 items) is possible.

After staging the PR, following steps i) to iii), leads to the same result as before:

image

It's possible to check what's in the cart for both sessions - session 2 (right) contains the items from session 1 as well. Also, and as before, it's not possible to check out from session 1; This is only possible for session 2.

So, I could not verify the test cases and I am not sure the PR introduces a benefit. Thoughts? (adding feedback label)



Another issue found, before and after the PR, and therefore unrelated:

After step iii), if instead of attempting to checkout one attempts to:
- go back to the cart on session 1 -> a redirection to the homepage occurs
- after this, trying to checkout on session 2 is not possible, and the error "Checkout Failed. Let us know so we can process your order" is seen.

At this point, the user has items in the cart, but cannot checkout. He cannot visit /cart page - gets redirected to the homepage
Trying to visit another shop leads to the grumpy cat:

image

Logging out and back in does not solve the issue, so basically, this account becomes blocked, in terms of customer function. This will need an issue, as it seems like a point of no return, for the user.


Respective bugsnag error:
https://openfoodnetwork.slack.com/archives/CEF14NU3V/p1597934757001500

luisramos0:
Hmmm, the interesting thing is that all this is before you stage this PR!

This PR does remove the logic of merging orders so my question is: what happens to this test case after the PR. I think it will be better! Can you please try?

luisramos0:
"The two carts should be persisted and independent." this is not correct without the last commit I reverted.
When you log in to the new session your cart should be there as it was on the other session.

filipefurtad0:
Hey @luisramos0,

I ran the tests before and after the PR, and got the same results... Also the issue found applies to before/after (I double checked this).

I also tried this:
- login to OFN and add something to the cart.
- Then, open an icognito tab and login again, do you see your cart as it was? -> Yup, this worked! :+1:

PR #5906, opened 2020-08-14
Fix flash error issues in checkout requests

### LINKED ISSUE #5895: Failed checkout shows incorrect error message
## Description

During checkout the payment can fail for several reasons, for example:

* Invalid card number.
* Invalid expiry date.
* Invalid CVC.
* Insufficient funds.
* ...

The first failure is shown correctly, but for subsequent failures the checkout page shows a previous error message.

## Expected Behavior

The error message shown tells the reason for the most recent failure.

## Actual Behaviour

The error message is often related to previous payment failures.

## Steps to Reproduce

There are two symptoms of this issue:

### Error showing twice

1. Choose a shop, add an item to the cart and fill in the checkout page.
2. Enter an invalid card number: `4444 3333 2222 1111`
3. Enter a valid expiry date and a random cvc: `123`
4. Click on checkout and observer the error message.
5. Click on the main OFN logo to go back to the homepage and observe the same error message.

### Wrong error on the checkout page

1. Choose a shop, add an item to the cart and fill in the checkout page.
2. Enter an invalid card number: `4000000000000002`
3. Enter a valid expiry date and a random cvc: `123`
4. Click on checkout and observer the error message.
5. Enter a card number for incorrect cvc: `4000000000000127`
6. Observe the previous error message "invalid card" instead of "invalid cvc".

## Animated Gif/Screenshot

memory_effect_Stripe_connect

## Workaround

Load another page before checking out again to see the correct error. Otherwise, just check out with a valid card. :wink: 

## Severity

## Your Environment

* Version used: v3.2.1

## Possible Fix

It's likely a problem with the flash error message. The flash is either not cleared or set twice.

Matt-Yorkley:
#### What? Why?

Closes #5895

Flash error messages created during checkout submission were being persisted in the flash session in unexpected ways, e.g. showing a checkout error message again after navigating away from the checkout page, or showing a message twice.

This was due to saving the errors in the flash session in XHR requests. The session is usually cleared after the subsequent action picks up the flash message, but here we needed to clear it after the current action, as the error is shown immediately via Angular on the checkout page (and shouldn't be persisted after that).

#### What should we test?

Using different test cards with Stripe Connect that should give different errors shows the flash errors as expected.

Submitting the checkout with a test card that generates an error and then navigating to the home page does not re-display the flash error.

#### Release notes

Improved flash message display on checkout

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -235,6 +235,7 @@ class CheckoutController < Spree::StoreController
         render :edit
       end
       format.json do
+        xhr_flash_errors
         render json: { errors: @order.errors, flash: flash.to_hash }.to_json, status: :bad_request
       end
     end
@@ -247,5 +248,12 @@ class CheckoutController < Spree::StoreController
 
   def permitted_params
     PermittedAttributes::Checkout.new(params).call
+  end
+
+  def xhr_flash_errors
+    # Marks flash errors for deletion after the current action has completed.
+    # This ensures flash errors generated during XHR requests are not persisted in the
+    # session for longer than expected.
+    flash.discard(:error)
   end
 end


Matt-Yorkley:
I've tried writing controller tests for this, but the flash object doesn't seem to behave as expected. We need to check what it looks like after the action has finished, but it doesn't seem to be updated. I've tested it in the browser and it's working as intended.

app/controllers/checkout_controller.rb
```diff
@@ -248,4 +249,11 @@ def rescue_from_spree_gateway_error(error)
   def permitted_params
     PermittedAttributes::Checkout.new(params).call
   end
+
+  def xhr_flash_errors
```
luisramos0, line 253: discard_flash_errors is a better description of what the method does I think.

Matt-Yorkley: Done.

PR #5907, opened 2020-08-15
Data cleanup

### LINKED ISSUE #2247: Delete old sessions from the database

### LINKED ISSUE #3608: Clean up useless data from live DBs to improve db backups speed and potentially app performance

Matt-Yorkley:
#### What? Why?

Closes #3608
Closes #2247

- Puts our databases on a much-needed diet by clearing out old data from the `spree_state_changes`, `spree_log_entries`, and `sessions` tables.
- Extracts `#remove_transient_data` from `TruncateData` and moves it to a separate rake task.
- Adds an entry in `config/schedule.rb` to run the task once per month.

#### What should we test?

Running `rake ofn:data:remove_transient_data` clears transient data as expected. Tailing the log shows an entry like this after running the task:

Screenshot from 2020-08-15 11-30-04

#### Release notes

Added scheduled task to clear old transient data from the database

### DIFF
--- a/lib/tasks/data/remove_transient_data.rb
+++ b/lib/tasks/data/remove_transient_data.rb
@@ -10,7 +10,7 @@ class RemoveTransientData
   def call
     Rails.logger.info("#{self.class.name}: processing")
 
-    Spree::StateChange.delete_all("created_at < '#{retention_period}'")
+    Spree::StateChange.where("created_at < ?", retention_period).delete_all
     Spree::LogEntry.delete_all("created_at < '#{retention_period}'")
     Session.delete_all("updated_at < '#{retention_period}'")
   end
--- a/lib/tasks/data/remove_transient_data.rb
+++ b/lib/tasks/data/remove_transient_data.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class RemoveTransientData
+  # This model lets us operate on the sessions DB table using ActiveRecord's
+  # methods within the scope of this service. This relies on the AR's
+  # convention where a Session model maps to a sessions table.
+  class Session < ActiveRecord::Base
+  end
+
+  def call
+    Rails.logger.info("RemoveTransientData: processing")
+
+    Spree::StateChange.delete_all("created_at < '#{1.month.ago.to_date}'")
+    Spree::LogEntry.delete_all("created_at < '#{1.month.ago.to_date}'")
+    Session.delete_all("created_at < '#{2.weeks.ago.to_date}'")
+  end
+end
--- a/lib/tasks/data/remove_transient_data.rb
+++ b/lib/tasks/data/remove_transient_data.rb
@@ -11,8 +11,8 @@ class RemoveTransientData
     Rails.logger.info("#{self.class.name}: processing")
 
     Spree::StateChange.where("created_at < ?", retention_period).delete_all
-    Spree::LogEntry.delete_all("created_at < '#{retention_period}'")
-    Session.delete_all("updated_at < '#{retention_period}'")
+    Spree::LogEntry.where("created_at < ?", retention_period).delete_all
+    Session.where("updated_at < ?", retention_period).delete_all
   end
 
   private
--- a/lib/tasks/data/remove_transient_data.rb
+++ b/lib/tasks/data/remove_transient_data.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class RemoveTransientData
+  RETENTION_PERIOD = 2.months.ago.to_date
+
   # This model lets us operate on the sessions DB table using ActiveRecord's
   # methods within the scope of this service. This relies on the AR's
   # convention where a Session model maps to a sessions table.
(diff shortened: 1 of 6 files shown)


lib/tasks/data/remove_transient_data.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    Spree::StateChange.delete_all("created_at < '#{retention_period}'")
+    Spree::LogEntry.delete_all("created_at < '#{retention_period}'")
+    Session.delete_all("updated_at < '#{retention_period}'")
+  end
+
+  private
+
+  def retention_period
+    2.months.ago.to_date
```
mkllnk, line 21: Can we start with six months, please? Sometimes it takes a while until a problem is recognised, documented and worked on.

lib/tasks/data/remove_transient_data.rb
```diff
@@ -0,0 +1,23 @@
...
+  # This model lets us operate on the sessions DB table using ActiveRecord's
+  # methods within the scope of this service. This relies on the AR's
+  # convention where a Session model maps to a sessions table.
+  class Session < ActiveRecord::Base
+  end
+
+  def call
+    Rails.logger.info("RemoveTransientData: processing")
+
+    Spree::StateChange.delete_all("created_at < '#{retention_period}'")
```
mkllnk, line 13:
```suggestion
    Spree::StateChange.where("created_at < ?", retention_period).delete_all
```

lib/tasks/data/remove_transient_data.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class RemoveTransientData
+  # This model lets us operate on the sessions DB table using ActiveRecord's
+  # methods within the scope of this service. This relies on the AR's
+  # convention where a Session model maps to a sessions table.
+  class Session < ActiveRecord::Base
+  end
+
+  def call
+    Rails.logger.info("RemoveTransientData: processing")
```
sauloperez, line 11:
```suggestion
    Rails.logger.info("#{self.class.name}: processing")
```
:ok_hand:

lib/tasks/data/remove_transient_data.rb
sauloperez: Given `Spree::StateChange` has no callbacks, I would prefer getting the performance boost of `delete_all`. Instantiating a potentially large list of models will have its cost.

lib/tasks/data/remove_transient_data.rb
sauloperez: we can start with 6 and see what's the impact on the system.

lib/tasks/data/remove_transient_data.rb
mkllnk: I'm still suggesting to use `delete_all`. The `where` clause is just a bit neater.

sauloperez: right, I got confused with the `where` and thought it would instantiate the records. Yes, it won't.

sauloperez:
I addressed all comments. Once green, we can give it a go in staging and hopefully have it ready soon so we stop dealing with AU's disk manually.

luisramos0 (approved):
Looks good, go!

mkllnk (approved):

sauloperez:
I deployed and the `crontab` lists the new entry:

```
# Begin Whenever generated tasks for: ofn at: 2020-09-09 07:48:03 +0000
0 0 1 * * /bin/bash -l -c 'cd /home/openfoodnetwork/apps/openfoodnetwork/current && RAILS_ENV=staging bundle exec rake ofn:data:remove_transient_data --silent'
```

sauloperez:
Alright, this definitely works although it doesn't log the SQL statements. I think it's totally fine as we already see a log line when it gets executed.

To be extra sure I wasn't getting confused with dates, I changed the code to `RETENTION_PERIOD = 1.day.ago.to_date` and it wiped out these three DB tables. Ready to go :tada: !

PR #5912, opened 2020-08-17
Terms and Conditions - upload PDF in the Enterprise Business Details and read them on Checkout

### LINKED ISSUE #5512: Hubs can upload Terms and Conditions PDF in backoffice and shoppers see a link and text on checkout
## Description

The very simplest iteration of this feature.

In Enterprise Settings -> Business Details hubs can upload a PDF containing their Terms and Conditions.

ToSfile
_Image 1: Terms and Conditions upload when no file has been uploaded_
ToSfile2
_Image 2: Terms and Conditions upload when file has been uploaded_

On Checkout, shoppers will be shown some text a link to this PDF. This link will open in a new Tab.
checkout
_Image 3: Checkout with the Terms and Conditions link available_

- The Hub can upload a PDF in Enterprise Settings -> Business Details. It should be clear to the user that the upload has been successful (eg the green banner)
- This PDF will be saved on OFN servers in a public location.
- When a shopper is checking out from this Hub they will see the above text and a link to the PDF that is viewable from the browser. 
- The link will open in a new tab.
- If the Hub has not uploaded a Terms and Conditions document then no message will be seen at Checkout. The checkout screen will not change for shoppers.
- If the Hub attempts to upload a document in a non-approved format they will see an error. I am not sure about what file types Paperclip gem will easily support so I leave this to developer discretion. PDF is minimum and sufficient. 
_"This file format is not allowed. Please select a PDF, ??"._

### Testing
1. Go to Enterprise Settings -> Business Details for an enterprise. Upload a PDF. Save the changes.
2. Go to Checkout for this enterprise. See the text "By placing this order you agree to the Terms of Service" above the 'Place order Now' button. The words Terms of Service should link to the same URL you entered in Step 1. The link should open in a new tab.
3. Repeat the process in a different language to confirm all fields have translation keys.
4. Go to Enterprise Settings -> Business Details for the same enterprise. Remove the URL. Save the changes.
5. Go to checkout for this enterprise. You should not see the text "By placing this order you agree to the Terms of Service" above the 'Place order Now' button. 
6. Attempt to upload a file in a non-approved format (devs to confirm which formats we can accept) and see an error message.

luisramos0:
#### What? Why?

Closes #5512

We use paperclip to manage an attachment called terms_and_conditions on the Enterprise model just like we manage promo images.
I used this Pr to make a little clean up of the checkout_spec (so I could add the terms and conditions spec but also preparing for the new stripe specs that are coming as part of https://github.com/openfoodfoundation/openfoodnetwork/issues/4639).

This is how it looks:
image

image

image

with the removal option:
image

#### What should we test?
Go to Enterprise Business Details and upload a pdf file. Verify you can click on the uploaded file name after clicking on "Update" and that will open the file.
Go to checkout and verify you can access the latest file uploaded in the Business Details.

#### Release notes
Changelog Category: Added
Enterprise managers can now upload a PDF file with their terms and conditions and users will be able to read the PDF file from the checkout page.


db/schema.rb
```diff
@@ -507,8 +511,8 @@
     t.integer  "source_id"
     t.string   "source_type"
     t.text     "details"
-    t.datetime "created_at",  null: false
-    t.datetime "updated_at",  null: false
```
sauloperez, line 511: :warning: this is not a formatting change like the ones above. Could you take it out? although it's weird that your schema ended up like this :thinking:

spec/features/admin/enterprises/terms_and_conditions_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+        go_to_business_details
+        expect(page).to have_selector("a[href*='logo-black.pdf']")
+      end
+
+      after do
+        # Delete fake PDFs
+        FileUtils.rm_f(white_pdf_file_name)
+        FileUtils.rm_f(black_pdf_file_name)
+      end
```
sauloperez, line 61: I find the [around hook](https://relishapp.com/rspec/rspec-core/v/3-9/docs/hooks/around-hooks) better suited for these things as the before and after logic goes together, so we don't risk messing with one or the other in the future.

spec/features/consumer/shopping/checkout_spec.rb
```diff
@@ -93,365 +115,339 @@
...
 
-        it "informs about previous orders" do
-          expect(page).to have_content("You have an order for this order cycle already.")
-        end
+      let(:response_mock) { { id: "ch_1234", object: "charge", amount: 2000 } }
+
+      before do
+        allow(Stripe).to receive(:api_key) { "sk_test_12345" }
+        allow(Stripe).to receive(:publishable_key) { "some_key" }
+        Spree::Config.set(stripe_connect_enabled: true)
```
sauloperez: if we're not restoring the state and this is not affecting any other specs, we could as well make this purposeful and explicit and set it globally in the spec_helper.

spec/support/request/checkout_helper.rb
```diff
@@ -38,16 +38,16 @@ def fill_out_billing_address
     end
   end
 
-  def fill_out_form
-    choose free_shipping.name
-    choose check_without_fee.name
+  def fill_out_form(shipping_method_name, payment_method_name, save_default_addresses = false)
```
sauloperez: a keyboard argument for `save_default_addresses` would shed some light on what this flag does when seeing it in use. It saves a jump to the method definition.

spec/features/consumer/shopping/checkout_spec.rb
```diff
@@ -154,52 +154,6 @@
...
-
-        # default card is selected, form element is not shown
-        expect(page).to have_no_selector "#card-element.StripeElement"
-        expect(page).to have_select 'selected_card', selected: "Visa x-1111 Exp:01/2025"
-
-        # allows checkout
-        place_order
-        expect(page).to have_content "Your order has been processed successfully"
-      end
-    end
```
sauloperez: :clap:

sauloperez:
Re Code Climate, I think we could get rid of the rule `Each selector in a comma sequence should be on its own single line`. It improves git diffs but I don't see any other benefit.

sauloperez:
Pretty good job @luisramos0 , specially with the deep cleanup you did! :tada:

db/schema.rb
luisramos0: oops! I rebased the PR and remove this :+1:

spec/features/admin/enterprises/terms_and_conditions_spec.rb
luisramos0: yeah, agree. thanks for the suggestion. done in new commit.

spec/support/request/checkout_helper.rb
luisramos0: done :+1:

spec/features/consumer/shopping/checkout_spec.rb
luisramos0: the default is false, I dont think we need to set it in spec_helper.
I added a commit to add restore original value in all specs that change this value.
I think this will make everything faster as stripe will always be disabled except when explicitly enabled.

luisramos0: please have a look at the new commit.

luisramos0:
@sauloperez all points addressed, please have a look.

luisramos0:
I actually like that rule, it's more clear and makes it very clear the different between spaces and commas.

sauloperez (approved):
Let's please fix the Code Climate issue `Avoid qualifying class selectors with an element.` :pray:

luisramos0:
That's completely unrelated to this PR Pau. I'd have to understand what that css is doing and where, remove the span and test if the avaialbe/unavailable still works.

sauloperez:
fair enough

spec/features/admin/enterprises/terms_and_conditions_spec.rb
```diff
@@ -22,30 +23,32 @@ def go_to_business_details
...
-      let(:black_pdf_file_name) { Rails.root.join("app", "assets", "images", "logo-black.pdf") }
+      let(:white_pdf_file_name) { Rails.root.join("app/assets/images/logo-white.pdf") }
+      let(:black_pdf_file_name) { Rails.root.join("app/assets/images/logo-black.pdf") }
 
       before do
         # Create fake PDFs from PNG images
-        FileUtils.cp(Rails.root.join("app", "assets", "images", "logo-white.png"), white_pdf_file_name)
-        FileUtils.cp(Rails.root.join("app", "assets", "images", "logo-black.png"), black_pdf_file_name)
-  
+        FileUtils.cp(Rails.root.join("app/assets/images/logo-white.png"), white_pdf_file_name)
```
mkllnk, line 31: Rubocop complaint about this? I thought the first version would be better...

app/views/checkout/_terms_and_conditions.html.haml
```diff
@@ -0,0 +1,2 @@
+%p
+  = t('.message_html', terms_and_conditions_link: link_to( t( '.link_text' ), current_order.distributor.terms_and_conditions.url, target: '_blank')) if current_order.distributor.terms_and_conditions.present?
```
mkllnk, line 2: Uh, this is a long line, doesn't fit in the Github view.

Anyway, I read that `present?` always returns true here. You need to use `file?`. Don't use `exists?` because that queries Amazon to see if it's really there.

spec/features/consumer/shopping/checkout_paypal_spec.rb
```diff
@@ -44,7 +44,8 @@
   describe "as a guest" do
     it "fails with an error message" do
       visit checkout_path
-      complete_the_form
+      checkout_as_guest
+      fill_out_form(free_shipping.name, paypal.name, false)
```
mkllnk: I don't find this helper very clear, especially the last parameter doesn't have a name. Maybe use a keyword argument here?
https://thoughtbot.com/blog/ruby-2-keyword-arguments

spec/features/consumer/shopping/checkout_spec.rb
```diff
@@ -117,9 +117,13 @@
         end
       end
 
-      it "it doesn't tell about previous orders" do
+      it "doesn't tell about previous orders" do
         expect(page).to have_no_content("You have an order for this order cycle already.")
       end
+
+      it "doesn't show link to terms and conditions" do
+        expect(page).to have_no_link("Terms of Service")
```
mkllnk, line 125: Hm, this seems to prove that I'm wrong about `attachment.present?`...

mkllnk (approved):
The pull request contains a few detours now but the end result looks good.

I learned that some of my knowledge is outdated. This doesn't seem to be true any more:
https://makandracards.com/makandra/1419-check-whether-a-paperclip-attachment-exists

spec/features/admin/enterprises/terms_and_conditions_spec.rb
luisramos0: yes, rubocop asked for this one :+1:

app/views/checkout/_terms_and_conditions.html.haml
luisramos0: :+1: done

spec/features/consumer/shopping/checkout_paypal_spec.rb
luisramos0: that's what Pau suggested, you probably saw it on the last commit.

app/views/admin/enterprises/form/_business_details.html.haml
```diff
@@ -31,3 +31,16 @@
     = f.label :invoice_text, t('.invoice_text')
   .omega.eight.columns
     = f.text_area :invoice_text, style: "width: 100%; height: 100px;"
+
+.row
+  .alpha.three.columns
+    = f.label :terms_and_conditions, t('.terms_and_conditions')
+
+  .omega.eight.columns
+    %a{ href: '{{ Enterprise.terms_and_conditions }}', ng: { if: 'Enterprise.terms_and_conditions' } }
```
andresgutgon, line 40: Why this is an `if` in javascript? Maybe you could check in Ruby/Haml and skip this work to browser when there are not terms and conditions uploaded

app/views/checkout/_terms_and_conditions.html.haml
```diff
@@ -0,0 +1,2 @@
+%p.small
```
andresgutgon, line 1: Do not render this empty `p.small` tag if not necessary. You could put the `if` before it :)

luisramos0: Hi Andres! Welcome to OFN! It's great to have your input here.
I did move the if above in the follow up PR :+1: 
https://github.com/openfoodfoundation/openfoodnetwork/pull/5994/files#diff-b38e245a26b89f5ee71d78eadee38163R2

The p.small is necessary for formatting.

filipefurtad0:
Hi @luisramos0 and @lin-d-hop ,

Thanks to both, for the testing notes and detailed acceptance criteria!

Tested following steps/scenarios:

- Under Enterprise Settings -> Business Details, checked that UI now displays the introduced field. Note the missing translations on the pic below (1): "Terms and Conditions", "Browse", "No File Selected"

image

- Upload a PDF, saving changes.
This works well, for a .pdf file (see below for other file types). Missing translations on the pics below as well (2) - "Remove File

image

- As a customer, on Checkout (for this enterprise) the text "By placing this order you agree to the Terms of Service" above the 'Place order Now' button now appears. Missing here as well (3):

image

- The words Terms of Service link to the same URL as entered before. The link opens in a new tab.
This works fine.

- Going to Enterprise Settings -> Business Details for the same enterprise, remove the URL and saving the changes
All good.

- After removal, the text "By placing this order you agree to the Terms of Service" above the 'Place order Now' button is also missing for customers on checkout.
  
- Repeating the process in different languages.
Went through the above described steps in French and English. Issues will be opened regarding the missing translations.

-  Attempting to upload a file in a non-approved format should display an error message.
I assumed only the .pdf file format is allowed. Attempted to upload a .txt file, the following error is displayed, which shows a missing translation as well (4):

image

After this, if one goes back to -> Business Details, it can be seen that the file did get uploaded:

image

Clicking on the file generates error:

image

Bugsnag
https://openfoodnetwork.slack.com/archives/CEF14NU3V/p1599742532044400


- Added different T&C files, on different enterprises. Made sure these are visible while checking out from the respective enterprises.
All good.




**Summary**

The PR does what is proposed in issue #5512 - so I'd move it to ready to go.

It does introduce a bug in the backoffice, which is reproduced by trying to:
- upload a .txt file
- attempting to view it

I'll open a new issue for this one. Maybe an S3?

It introduces the already mentioned missing translations (1) (2) (3) and (4).

luisramos0:
thanks Filipe.
The missing translations is not a problem. Do you understand the process? The new keys are added as part of the PR but the keys will only get to transifex once the PR is merged. So, all new keys introduced in a PR will be missing when testing the PR.
@lin-d-hop can you please decide about the txt file problem? I'd merge this PR as is and fix the problem in a separate issue. It's not really an S2 but I think we should include it and fix it right way. Should probably be a quick one to fix.

filipefurtad0:
Thanks for pointing that out @luisramos0,
I'll check them after merging again.

In the meantime I've opened issue #6011 should we decide to go ahead with this PR for this week's release.

luisramos0:
"I'll check the translations after merging again."
when the PR is merged, the keys will get to transifex, no point in testing again.
After the keys land on transifex, translators need to do their job and only when a Transifex PR is merged or a release is prepared, thanslations will be in master.
Here's the doc:
https://github.com/openfoodfoundation/openfoodnetwork/wiki/Internationalisation-(i18n)

luisramos0:
I'll take the decision and merge this now.
I think this is a valid MVP, I think #6011 is a minor detail.

RachL:
@luisramos0 @filipefurtad0 but it's important to notice that this detail was in the acceptance criteria of the issue. So unless @lin-d-hop thinks otherwise I think #6011 should be done as part of the T&Cs epic and not as part of a prioritized s3.

luisramos0:
thanks @RachL for the feedback. sure, I agree, I moved 6011 to the TCs column.

PR #5918, opened 2020-08-18
Do not add stripe JS code to layout if stripe is disabled at instance level

luisramos0:
#### What? Why?

This closes the stripe part of #5078 
the stripe js code is not loaded if stripe is deactivated in the instance.

#### What should we test?
We need to verify that stripe payments can be made after de-activating and re-activating stripe again.
Go to Configuration > Stripe and de-active stripe.
Go to the frontoffice, any page, and verify the stripe JS code is not loaded:
image
This tag should not be there when stripe is de-activated.
Go to Configuration > Stripe and re-activate stripe.
Go to the frontoffice, verify the stripe JS code is loaded and that you can make a Stripe payment.

#### Release notes
Changelog Category: Changed
Stripe javascript code is only loaded if stripe is enabled in the instance.

### DIFF
--- a/app/views/layouts/darkswarm.html.haml
+++ b/app/views/layouts/darkswarm.html.haml
@@ -38,9 +38,12 @@
     %loading
 
     = render "layouts/bugsnag_js"
-    %script{:src => "https://js.stripe.com/v3/", :type => "text/javascript"}
+
+    - if Spree::Config.stripe_connect_enabled
+      %script{:src => "https://js.stripe.com/v3/", :type => "text/javascript"}
     - if !ContentConfig.open_street_map_enabled
       %script{src: "//maps.googleapis.com/maps/api/js?libraries=places,geometry#{ ENV['GOOGLE_MAPS_API_KEY'] ? '&key=' + ENV['GOOGLE_MAPS_API_KEY'] : ''} "}
+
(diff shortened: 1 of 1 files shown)


Matt-Yorkley (approved):

filipefurtad0:
Hi @luisramos0 ,

The pic below was taken before staging this PR, with Stripe-connect disabled in Configuration > Stripe Connect. We can see that the Stripe javascript code is there, despite being disabled:

image


After staging the PR, we can verify the code is not being loaded:

image


Re-enabling Stripe-Connect in Configuration allows the Stripe JS code to be loaded, as seen below:

image

I've additionally performed a payment with Stripe just to make sure everything is working well - all good.

Thanks Luis - Ready to go! :tada:

sauloperez:
I think this would also speed up things if we set `stripe_connect_enabled` only on the required specs :ok_hand:

luisramos0:
I think this will speed up specs already because the default is false @sauloperez

sauloperez:
what I'm seeing is that there are specs that don't restore its state back to false.

luisramos0:
:+1: I have fixed that in https://github.com/openfoodfoundation/openfoodnetwork/pull/5912

Issue #5607, opened 2020-06-15
[Rails 4.1] Mutator methods called on Relation


luisramos0:
Where we use mutator methods, we need to convert the relation to an array:
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#mutator-methods-called-on-relation

This is already done in some commits here but it's not complete:
https://github.com/openfoodfoundation/openfoodnetwork/compare/rails-4-1?expand=1

andresgutgon:
Hi @luisramos0 I'm a bit idiot and worked on this issue without realising you already had a PR open 🙈  Sorry for that. 

Any way I open PR because I think I walked all the places where mutator methods were used and I also fixed a couple of unrelated issues:

1. [A broken query in Rails 4.1.x](https://github.com/openfoodfoundation/openfoodnetwork/pull/6063#discussion_r491688453)
2. And this open issue [ActiveSupport::Concern::MultipleIncludedBlocks #6023
](https://github.com/openfoodfoundation/openfoodnetwork/issues/6023)

Feel free to close my PR or use whatever you think can be helpfult. 

Next time I'll filter [issues without assignee for Rails 4-1 epic](https://github.com/openfoodfoundation/openfoodnetwork/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+%5BRails+4.1%5D) 😅

andresgutgon:
Also note that my branch is pointing to `rails-4-1` branch. Not sure how you're working here. What's the idea make mergeable each issue into master or put everything on `rails-4-1`, do QA and merge into master that branch?

luisramos0:
ah, I now see these message :+1: 
If possible we make rails 4.1 changes in master, if not possible we work on the rails 4.1 branch.
I think your PR is good for master.

I think the easiest way is for me to take over your PR yes, I'll get from it what's not in my PR and create a separate PR against master, ok? Thanks a lot for your contribution btw!

luisramos0:
All broken specs are fixed here. Closing.

PR #5919, opened 2020-08-18
[Rails 4.1, works in master] Convert relations with to_a so that they work in rails 4.1  

luisramos0:
#### What? Why?

This addresses a lot of points in https://github.com/openfoodfoundation/openfoodnetwork/issues/5607 
All this should work in master and fix a lot of broken specs in rails 4.1.

#### What should we test?
I think we can trust the build here. We could verify all the little different scenarios that this PR changes with to_a but I dont think it's worth the trouble.

#### Release notes
Changelog Category: Changed
Adapted code to work in rails 4.1

### DIFF
--- a/app/models/model_set.rb
+++ b/app/models/model_set.rb
@@ -50,6 +50,7 @@ class ModelSet
     # Remove all elements to be deleted from collection and return them
     # Allows us to render @model_set.collection without deleted elements
     deleted = []
+    @collection = collection.to_a
     collection.delete_if { |e| deleted << e if @delete_if.andand.call(e.attributes) }
     deleted
   end
--- a/app/views/spree/orders/_bought.html.haml
+++ b/app/views/spree/orders/_bought.html.haml
@@ -3,7 +3,7 @@
     %td.toggle-bought{ colspan: 2, ng: { click: 'showBought=!showBought' } }
       %h5.brick
         %i{ ng: { class: "{ 'ofn-i_007-caret-right': !showBought, 'ofn-i_005-caret-down': showBought}"} }
-        = t(:orders_bought_items_notice, count: @order.finalised_line_items.sum(&:quantity))
+        = t(:orders_bought_items_notice, count: @order.finalised_line_items.to_a.sum(&:quantity))
     %td.text-right{ colspan: 3 }
       %a.edit-finalised.button.radius.expand.small{ href: changeable_orders_link_path, ng: { class: "{secondary: !showBought, primary: showBought}" } }
         = t(:orders_bought_edit_button)
--- a/lib/open_food_network/packing_report.rb
+++ b/lib/open_food_network/packing_report.rb
@@ -61,7 +61,7 @@ module OpenFoodNetwork
                               proc { |_line_items| "" },
                               proc { |_line_items| I18n.t('admin.reports.total_items') },
                               proc { |_line_items| "" },
-                              proc { |line_items| line_items.sum(&:quantity) },
+                              proc { |line_items| line_items.to_a.sum(&:quantity) },
                               proc { |_line_items| "" }] },
           { group_by: proc { |line_item| line_item.product.supplier },
             sort_by: proc { |supplier| supplier.name } },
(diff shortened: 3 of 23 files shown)


app/models/model_set.rb
```diff
@@ -50,6 +50,7 @@ def collection_to_delete
     # Remove all elements to be deleted from collection and return them
     # Allows us to render @model_set.collection without deleted elements
     deleted = []
+    @collection = collection.to_a
     collection.delete_if { |e| deleted << e if @delete_if.andand.call(e.attributes) }
```
Matt-Yorkley, line 54: Isn't `collection` referencing the same object as `@collection` here? Can we just do `collection.to_a.delete_if...` without the extra assignment in line 53?

app/models/spree/line_item_decorator.rb
```diff
@@ -109,7 +109,7 @@ def price_with_adjustments
 
     line_item_adjustments = OrderAdjustmentsFetcher.new(order).line_item_adjustments(self)
 
-    (price + line_item_adjustments.sum(&:amount) / quantity).round(2)
+    (price + line_item_adjustments.to_a.sum(&:amount) / quantity).round(2)
```
Matt-Yorkley, line 112: Isn't `amount` an attribute on adjustment records? Can this move to #5921 as `line_item_adjustments.sum(:amount)`?

app/views/spree/orders/_bought.html.haml
```diff
@@ -3,7 +3,7 @@
     %td.toggle-bought{ colspan: 2, ng: { click: 'showBought=!showBought' } }
       %h5.brick
         %i{ ng: { class: "{ 'ofn-i_007-caret-right': !showBought, 'ofn-i_005-caret-down': showBought}"} }
-        = t(:orders_bought_items_notice, count: @order.finalised_line_items.sum(&:quantity))
+        = t(:orders_bought_items_notice, count: @order.finalised_line_items.to_a.sum(&:quantity))
```
Matt-Yorkley, line 6: This can be `finalised_line_items.sum(:quantity)` right?

lib/open_food_network/packing_report.rb
```diff
@@ -61,7 +61,7 @@ def rules
                               proc { |_line_items| "" },
                               proc { |_line_items| I18n.t('admin.reports.total_items') },
                               proc { |_line_items| "" },
-                              proc { |line_items| line_items.sum(&:quantity) },
+                              proc { |line_items| line_items.to_a.sum(&:quantity) },
```
Matt-Yorkley, line 64: `line_items.sum(:quantity)`?

app/models/model_set.rb
luisramos0: I dont think so, we need to convert the attribute to_a, if we do what you suggest the delete_if will only work on the converted array, not the attribute.

app/views/spree/orders/_bought.html.haml
luisramos0: tests do break with that, in terms of prod code may it would work but there are quite some tests that need this. That's why I went this way...

lib/open_food_network/packing_report.rb
luisramos0: same as above, tests where line_items is given as an array...

Matt-Yorkley (approved):

jhsu802701:
What will it take to merge this pull request?  Is there anything I can do to speed up the process?  I HATE getting spammed by the thousands of deprecation messages, because they obscure what's going on and make the troubleshooting process FAR more difficult.

sigmundpetersen:
Hey @jhsu802701, the process is described here https://github.com/openfoodfoundation/openfoodnetwork/wiki/Pipeline-development-process, specifically under the heading 'Processes'.

luisramos0:
I rebased to resolve conflicts.

app/models/model_set.rb
andrewpbrett: I got bitten by this a few weeks ago (I couldn't find the PR but Luis and I had a similar conversation). I was passing `collection.to_a` into a method and the method was operating on the converted array. So it makes sense that we need the extra line.

andrewpbrett (approved):
It doesn't look like there's anything too complicated going on, so if the build is green I'd say it's good to merge 👍

luisramos0:
bang! I am happily count your review as a second review on this one (I have been waiting for this one for a long time) but we better validate that with the rest of the devs :-)

Issue #5609, opened 2020-06-15
[Rails 4.1] Different accessor code required if model attribute is a DB column or not 


luisramos0:
I am not sure what's the cause of this but we need to make this change in quite a few places I think:
https://github.com/openfoodfoundation/openfoodnetwork/issues/5386#issuecomment-630877549

luisramos0:
All related broken specs are fixed, closing.

PR #5921, opened 2020-08-18
[Rails 4.1] Adapt calls to model attributes depending on if they are db columns or not

luisramos0:
#### What? Why?

This addresses quite a few points related to https://github.com/openfoodfoundation/openfoodnetwork/issues/5609

#### What should we test?
More green specs.
After this PR and https://github.com/openfoodfoundation/openfoodnetwork/pull/5919 is merged to master and rails-4-1, we can start tracing the exact number of broken specs in each PR #finalcountdown


Matt-Yorkley (approved):
Nice

Matt-Yorkley:
Merging, as it's not going into `master` and has been sat here for a month. Also it's pretty uncontroversial and looks good :+1:

PR #5927, opened 2020-08-19
Simplify order loading and avoid unnecessary queries

mkllnk:
#### What? Why?

Related to #5905. This PR contains those commits as well. Only the last two are new.

While working on a previous PR, I saw some more scope for simplifying the code. I branched for a separate PR because I didn't want to delay the previous PR.

#### What should we test?

* Create a shop with inventory (variant overrides).
* Override prices and check them in the shopfront.
* Shop and checkout as guest.
* Shop and checkout as logged in user.

#### Release notes

Changed: Slightly improved performance on all pages.

Changelog Category: Changed


lib/spree/core/controller_helpers/order.rb
```diff
@@ -68,11 +67,6 @@ def associate_user
           session[:guest_token] = nil
         end
 
-        # Load current order and create a new one if necessary.
-        def set_current_order
-          current_order(true) if spree_current_user
```
luisramos0: I am not sure about this change. Looks good but current_order param default is false, a call to current_order will not create the order if it's not there. This was creating the order if it is not there all the time.
I think we may have cases that need this.

We could maybe optimize current_order a bit more instead?
It is already defensive with:
return @current_order if @current_order

but we can improve the scoping part: I think this is loading all the hubs overrides every time you run current_order :scream_cat: 
https://github.com/openfoodfoundation/openfoodnetwork/blob/9771fd7f13549bad7b3465ae091dbd054d8f20bc/lib/spree/core/controller_helpers/order.rb#L20-L24

If I am not mistaking memoizing the scoper here would make a huge difference.

mkllnk: > I think this is loading all the hubs overrides every time you run current_order 

I thought that as well and wanted to fix it but when I tested, the generated query contains `NULL` as distributor and has no results. Still not good to have a useless query.

mkllnk: > I am not sure about this change. Looks good but current_order param default is false, a call to current_order will not create the order if it's not there. This was creating the order if it is not there all the time.
I think we may have cases that need this.

I wasn't 100% sure about this either. That's why I opened a separate pull request. My sanity checks seemed to show that everything works but it would need good testing. I think that this is worth doing. How do we verify that we don't break anything? All automated tests pass....

mkllnk:
I will open a new pull request with the remaining useful commit.

PR #5929, opened 2020-08-19
[Bye bye spree] Bring Spree engine.rb from spree core

luisramos0:
Bring engine-rb from spree_core, the core of the core, almost last step.

core/lib/spree/core/engine.rb COPY most of it.

Build is broken with a problem with a dependency to Spree::Preferences::Preferable in engine.rb
We can wait for #5870 where preferable is brought from spree_core, or we can leave engine.rb for the last step when spree_core is removed.


luisramos0:
this is now in #5978

PR #5938, opened 2020-08-21
Fix the bulk coop report date and distributor filters.

### LINKED ISSUE #5837: Date selector and Distributor ID filters don't work on Bulk Co-op report
## Description

Date selector and Distributor ID filters currently don't work on Bulk Co-op report. Displayed results are the same, independently of choice of distributor or date.

## Expected Behavior

It should be possible to filter results on the Bulk Co-op report by date or distributor ID.

## Actual Behaviour

None of these filters work.
## Steps to Reproduce

Best observed as superadmin
1. Go to Reports and Bulk- Coop; choose any sub-report type.
2. Leave the date selector and Distributor ID fields blank, and press "Generate Report"
3. Set a date range and/or a Distributor ID
4. Press "Generate Report" and compare the query results those obtained in point 2 -> they should be identical.  

## Workaround

If the report does not break, the work around is to filter results manually, or in the CSV file

## Severity

Considering the workaround this would be an bug-s3. However, for some hubs, perhaps it is easy to generate a Error 504 - Gateway Timeout, which would just break the report; in these cases, this would be an bug-s2. So, I think maybe a prioritized bug-s3?

## Your Environment

* Version used: v3.1.3
* Browser name and version: Firefox 78
* Operating System and version (desktop or mobile): Desktop Ubuntu 20.04

## Possible Fix

Together with poor performance - which was already considerably improved with PR #5815 - this could an additional underlying cause to the bug reported in #5814.

Related to #4747 and to #5032.

Possible v3-regression.

cillian:
#### What? Why?

Closes #5837 

Before the date and distributor filters would have no effect. This is because the BulkCoopReport is still generated using an older style method, and isn't generated using the newer method like in the EnterpriseFeeSummaryReport. This older style report expects to receive a `q` parameter but it actually received the newer style `report` parameter so the filters were not being applied.

This keeps the newer style report params but converts them, after they are authorised as safe, into the older style in the controller.

There is a test already for checking that the parameters are authorised correctly [here](https://github.com/openfoodfoundation/openfoodnetwork/blob/master/engines/order_management/spec/controllers/order_management/reports/bulk_coop_controller_spec.rb#L51).

#### What should we test?

1. Go to /order_management/reports/bulk_coop/new
2. Generate report with no filters, then filter it by date and distributor to see the difference.

#### Release notes

Fix the bulk coop report date and distributor filters.

Changelog Category: Fixed

### DIFF
--- a/engines/order_management/app/controllers/order_management/reports/bulk_coop_controller.rb
+++ b/engines/order_management/app/controllers/order_management/reports/bulk_coop_controller.rb
@@ -46,8 +46,8 @@ module OrderManagement
             completed_at_gt: params[:report][:start_at],
             completed_at_lt: params[:report][:end_at],
             distributor_id_in: params[:report][:distributor_ids],
-            report_type: params[:report][:report_type]
-          }
+          },
+          report_type: params[:report][:report_type]
         }
       end
 
--- a/engines/order_management/app/controllers/order_management/reports/bulk_coop_controller.rb
+++ b/engines/order_management/app/controllers/order_management/reports/bulk_coop_controller.rb
@@ -13,7 +13,8 @@ module OrderManagement
 
         @report_parameters.authorize!(@permissions)
 
-        @report = report_klass::ReportService.new(@permissions, params[:report], spree_current_user)
+        @report = report_klass::ReportService.new(@permissions, legacy_format_report_params,
+                                                  spree_current_user)
         renderer.render(self)
       rescue ::Reports::Authorizer::ParameterNotAllowedError => e
         flash[:error] = e.message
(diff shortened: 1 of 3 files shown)


sauloperez (approved):
:tada:

filipefurtad0:
Hi @cillian ,

I had a look at the Bulk Coop report, before and after the PR. As described in the issue, the date and distributor filters were not working, for the different report types:

image

This is now working, **after this PR**:

image


However, in the pics above, it can be seen that there are change introduced on the columns. I checked for each report types,a found that **before this PR**, the columns were:

image

image

image

image


**After this PR** the columns which appear on all report types are:

image

So, the filters now work fine, but something seems to be breaking the way the reports are generated. I think this may hide relevant information to the users, so maybe better not to send into production. Could you have a second look at this PR? Thank you!

Moving back do InDev.

cillian:
@filipefurtad0 thanks, sorry I missed that. That should be fixed now, I also added extra tests.

luisramos0:
:+1: I think one re-review is enough. moving back to test ready.

filipefurtad0:
Hey @cillian ,

This looks great now:

image


image


image


image

CSV files are created accordingly.

Thank you so much :+1: 
Moving this to Ready to Go.

PR #5941, opened 2020-08-22
[Bye bye Spree] Make OFN independent of spec helpers under core/lib/testing_support

luisramos0:
#### What? Why?

Here we move some required helpers from spree_core: testing_support/i18n, preferences helper and some methods from capabara_ext into WebHelper.

This makes OFN independent of all helpers under core/lib/testing_support
The only remaining needed files there are the factories, will be addressed in a separate PR.

#### What should we test?
Only test code changed, a green build is enough.

#### Release notes
Changelog Category: Changed
Updated test code to be independent of spree test helpers so we can make OFN independent of Spree.


spec/support/request/web_helper.rb
```diff
@@ -95,16 +95,29 @@ def wait_until(secs = nil)
...
   end
 
+  def targetted_select2(value, options)
+    # find select2 element and click it
+    find(options[:from]).find('a').click
+    select_select2_result(value)
+  end
+
+  def select_select2_result(value)
+    sleep(1)
```
sauloperez, line 115: :see_no_evil: can you try removing this and use `have_content` or similar if needed?

spec/support/request/web_helper.rb
```diff
@@ -145,6 +170,28 @@ def select_select2_result_async(value)
...
+
+    if label.nil?
+      raise "Could not find label by text #{text}"
+    end
+
+    label
+  end
+
+  def find_label(text)
+    first(:xpath, "//label[text()[contains(.,'#{text}')]]")
```
sauloperez, line 192: If we used `find` instead of `first` I bet the while loop above won't be needed. `find` is one of those Capybara waiting methods.

sauloperez (commented):
It'd be worth taking a look at those things I pointed out. It might be complexity that makes tests slow and we don't need it.

spec/support/request/web_helper.rb
luisramos0: I am just bringing this from Spree in this PR.
I tried removing this but it breaks the specs.
have_content is not appropriate, this method is actually selecting the select2 result with mouseup.
It's dodgy but it's one of those things we need first to bring to our side, and fix in a second step. There will be a few like these in the bye bye spree PRs.

spec/support/request/web_helper.rb
luisramos0: I had a second look, I managed to delete this code, have a look at the new commit.

spec/support/request/web_helper.rb
sauloperez: could you create an issue for it so we don't lose track?

sauloperez (approved):

sauloperez:
This is pretty simple and hardly matter of discussion so I'm moving forward and merging for the sake of the pipe.

Matt-Yorkley (approved):

spec/support/request/web_helper.rb
luisramos0: yes, done: https://github.com/openfoodfoundation/openfoodnetwork/issues/5972
this could save us a few minutes of build time :+1:

sauloperez: you said MINUTES :scream:

luisramos0: yes, it's called dozens of times in the specs :+1:  I moved it to the top of "tech debt prioritized" now :-)

PR #5943, opened 2020-08-23
[Bye bye Spree] Make OFN independent of all spree spec factories

luisramos0:
#### What? Why?
Relates to #4826 
Move all necessary factories from spree_core to ofn.

#### What should we test?
A green build is enough here, only test code changed.

#### Release notes
Changelog Category: Changed
Brought test support code from spree so that we can make OFN independent of Spree.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -160,10 +160,3 @@ FactoryBot.define do
     stripe_publishable_key "xyz456"
   end
 end
-
-FactoryBot.modify do
-  factory :shipping_category, class: Spree::ShippingCategory do
-    initialize_with { DefaultShippingCategory.find_or_create }
-    transient { name 'Default' }
-  end
-end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,6 +1,9 @@
 require 'ffaker'
 
 FactoryBot.define do
+  sequence(:random_description) { Faker::Lorem.paragraphs(1 + Kernel.rand(5)).join("\n") }
+  sequence(:random_email)       { Faker::Internet.email }
+
   factory :classification, class: Spree::Classification do
   end
 
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -162,17 +162,6 @@ FactoryBot.define do
 end
 
 FactoryBot.modify do
-  factory :payment do
-    transient do
-      distributor {
-        order.distributor ||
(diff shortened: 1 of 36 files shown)


spec/factories.rb
```diff
@@ -1,22 +1,10 @@
 require 'ffaker'
-require 'spree/testing_support/factories'
```
luisramos0, line 2: this is the magic of this PR: no spree factories used.

Matt-Yorkley (approved):

spec/factories/variant_factory.rb
```diff
@@ -1,46 +1,66 @@
...
-        producer { product.supplier }
-        coordinator { create(:distributor_enterprise) }
-        distributor { create(:distributor_enterprise) }
-        incoming_exchange_fees { [] }
-        outgoing_exchange_fees { [] }
+        on_demand { false }
+        on_hand { 5 }
       end
 
+      # on_hand 5
```
sauloperez, line 25: :fire:

spec/models/spree/stock_item_spec.rb
```diff
@@ -3,7 +3,15 @@
...
 RSpec.describe Spree::StockItem do
-  let(:stock_location) { create(:stock_location_with_items) }
+  let(:stock_location) { create(:stock_location) }
+
+  before do
+    product_1 = create(:product)
+    product_2 = create(:product)
+
+    stock_location.stock_items.where(:variant_id => product_1.master.id).first.adjust_count_on_hand(10)
+    stock_location.stock_items.where(:variant_id => product_2.master.id).first.adjust_count_on_hand(20)
```
sauloperez, line 13: we can replace the hash rocket while we are at it :pray:

sauloperez (changes requested):
:clap:

spec/factories/variant_factory.rb
luisramos0: :+1:

spec/models/spree/stock_item_spec.rb
luisramos0: well spotted :+1:

luisramos0:
:+1: I ended up running rubocop -a on spec/factories (last commit). Let's see if the build is ok.

sauloperez (approved):

